Issue #9022, opened 2018-10-11
InputObjectType inference broken for JSON-only types


stefanprobst:
## Description

Adding a field in `setFieldsOnGraphQLNodeType` with `GraphQLJSON` only crashes the query compiler.

### Steps to reproduce

```js
// gatsby-node.js

const {
  GraphQLObjectType,
  GraphQLJSON,
  GraphQLBoolean,
} = require('gatsby/graphql')

const CustomType = new GraphQLObjectType({
  name: 'CustomType',
  fields: {
    first: { type: GraphQLJSON },
    second: { type: GraphQLJSON },
    // Adding this dummy field prevents the crash
    // _repair: { type: GraphQLBoolean },
  },
})

exports.setFieldsOnGraphQLNodeType = ({ type }) => {
  if (type.name !== 'MarkdownRemark') {
    return {}
  }
  return {
    customfield: {
      type: CustomType,
      resolve: () => ({
        first: { hello: 'world' },
        second: { hello: 'world' },
      }),
    },
  }
}
```

The crash happens because [graphql type validation](https://github.com/graphql/graphql-js/blob/f4dee2803c22f7e5fde913b9743cc6de345bb871/src/type/validate.js#L532) expects at least one field, but none get added for `GraphQLJSON` in `infer-graphql-input-fields-from-fields` because JSON is not listed in [`scalarFilterMap`](https://github.com/gatsbyjs/gatsby/blob/05a07c481ea1bf2b9517a44b083e049173e21ea0/packages/gatsby/src/schema/infer-graphql-input-fields-from-fields.js#L84).

### Possible solutions

As a workaround, adding a dummy non-JSON field prevents the crash (see above).
Adding something like the following to `scalarFilterMap` does as well, but not sure if this is the correct way to go about this:
```js
JSON: {
    eq: { type: GraphQLJSON }
},
```

pieh:
We have 2 problems here:
 1. Crash when there is no valid properties - which we should fix by skipping fields like that. This one is critical and can happen in other scenarios too. I think here we can create simplified reproduction using hello-world starter that creates single node with one of the field being empty object:
```js
exports.sourceNodes = ({ actions }) => {
  actions.createNode({
    id: `foo`,
    parent: null,
    children: [],
    internal: {
      type: `CrashingNode`
      contentDigest: `bar`,
    },
    fieldThatWillCauseCrash: {}
  })
}
```
[...]

PR #9057, opened 2018-10-12
fix: check for empty fields in InputFilter inference

stefanprobst:
Check for empty fields in InputFilter inference to prevent crash in InputObjectType validation. This simply does in `convertToInputFilter` what is already done in  `convertToInputType`.

See: #9022


pieh:
It would be great to add test for that so we can catch any regressions

stefanprobst:
Test added. In case we decide to add input filters for JSON, these would have to be changed as well.

PR #9061, opened 2018-10-12
refactor(ci): move cypress tests to e2e-tests directory; update circleci config

DSchau:
This PR does some slight refactoring to our project directory and some other tweaks. Changes include:

- Move cypress tests from integration-tests/* to e2e-tests
- Document how to add tests / run them / etc.
    - This should probably be a tutorial/doc somewhere on gatsbyjs.org too
- Tweaks to CircleCI renaming integration -> e2e and running correct scripts
- Changes name of "ignore_pattern" to "trigger_pattern" for clarity
- Only run e2e-tests if change to packages/ _or_ the exact e2e test
    - This could be contentious, but I think it's a worthwhile change

I'm not 100% sure all the tests will pass, but we should be _close_

cc @ahinrichs


DSchau:
Heads up, there are some unrelated changes (e.g. prettified files) that made their way into this commit.

I think it's because that www/ hook fired on the _staged_ file changes, e.g. moving the integration-tests tests to e2e-tests, therefore marking them as changed. Not a big deal, actually a nice change but makes the diff noisy.

ahinrichs:
I also agree on the changes. Definitivly better to have e2e and integration tests separated.

Will go and use this layout for my tests. I will open another PR if I'm happy with it.

PR #8992, opened 2018-10-10
chore(*): Use new createContentDigest helper

samscha:
**Note**: this is my first pr to Gatsby so if I'm missing something (best practices included), please do tell me about them and I'll update this/future pr's.

Added createContentDigest helper for gatsby-transformer-remark and gatsby-source-filesystem.

I opened this pr right now mainly because I wanted to make sure I was adding the createContentDigest helper to these two plugins correctly and not breaking anything (before going on to other plugins).

If it's more convenient, and I have verification I correctly added the createContentDigest helper to these two plugins, this pr can be closed and I can open a new pr later with more plugins that have the createContentDigest helper added.

Thanks!

from
#8805

PR #9072, opened 2018-10-12
[NO-MERGE] Enhancing createContentDigest tests -- requesting help

samscha:
So I've been trying to update tests to properly utilize `createContentDigest` (right now only for gatsby-transformer-csv).

I've found a way to do this **but** it requires a pretty lengthy `import`. It also requires an `import` _outside_ of the plugin's root dir.. which I would rather not do (and is probably not going to pass all CI tests).

I'm still pretty new to Gatsby, but what I would want to do is something like have every plugin auto-import the gatsby/src/utils so I can use them in any test file. This may very well exist and I've just missed it these past two days (doh!).

Thanks for reading and any help is appreciated!

Sam

P.S. the snapshot is directly copy-pasta'd from the pre-contentCreateDigest snapshot ([this blob](https://github.com/gatsbyjs/gatsby/blob/1fb19f9ad16618acdac7eda33d295d8ceba7f393/packages/gatsby-transformer-csv/src/__tests__/__snapshots__/gatsby-node.js.snap))

from
#8992


pieh:
I'm thinking if we should add export to `gatsby` package with utils that can be used (primarily) for testing - something like:
``const { createContentDigest } = require(`gatsby/utils`)``
and then add `gatsby` as devDependency to packages that make use of those utils - this would also allow 3rd party plugins use same tools as plugins in our monorepo. Other utility would be `createNodeId` (this one is a bit more complicated, as we generate that utility function per package)

What do you think about that?

samscha:
P.S. but I definitely would be interested in looking at how to implement this after I finish #8992 !

DSchau:
@samscha sorry about this--this kinda fell through the cracks!

Would you be interested in [signing up for a pairing session](https://www.gatsbyjs.org/contributing/pair-programming/) and we can get this a little closer to getting merged?

samscha:
yes.. 😅 

I definitely would be interested in the pairing session. I'll sign up for that!

KyleAMathews:
This PR was superseded by a newer one.

PR #9075, opened 2018-10-12
docs(www): Adding a SEO component

### LINKED ISSUE #4921: [v2] Document how to create a "SEO" component using <StaticQuery>
This is a common problem people run into — it's much more straightforward in v2 with `<StaticQuery>` so let's improve the docs for this.

LekoArts:
Fixes #4921

So I was at a [hacktoberfest event](https://hacktoberfestffm.de/) this evening and this is what I came up with!

I also saw that the `add-page-metadata` document is already finished but not in the sidebar.

### DIFF
--- a/docs/docs/add-seo-component.md
+++ b/docs/docs/add-seo-component.md
@@ -2,8 +2,7 @@
 title: "Adding a SEO component"
 ---
 
-Every website on the web has basic _meta-tags_ like the title, favicon or description of the page in their `<head>` element. This information gets displayed in the browser and is used when someone, e.g. shares your website on Twitter. You can give your users and these websites additional data to embed your website with more data — and that's where this guide for a SEO component comes in. At the end you'll have a component you can place in your layout file and
-have rich previews for other clients, smartphone users, and search engines.
+Every site on the web has basic _meta-tags_ like the title, favicon or description of the page in their `<head>` element. This information gets displayed in the browser and is used when someone, e.g. shares your website on Twitter. You can give your users and these websites additional data to embed your website with more data — and that's where this guide for a SEO component comes in. At the end you'll have a component you can place in your layout file and have rich previews for other clients, smartphone users, and search engines.
 
 _Note: This component will use StaticQuery. If you're unfamiliar with that, have a look at the [StaticQuery documentation](/docs/static-query/). You also have to have `react-helmet` installed for which you can have a look at [this document](/docs/add-page-metadata)._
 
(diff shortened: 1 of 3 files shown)


docs/docs/add-seo-component.md
```diff
@@ -0,0 +1,199 @@
...
+
+SEO.propTypes = {
+  title: PropTypes.string,
+  description: PropTypes.string,
+  image: PropTypes.string,
+  pathname: PropTypes.string,
+  article: PropTypes.bool,
+}
+
+SEO.defaultProps = {
```
DSchau: I kinda like the idea of _not_ including these. Not a strong reservation though!

DSchau: Or at least not including the ones that are null!

LeKoArts: Why so?

DSchau: undefined will be handled the same way as null, i.e. falsy.

What's the benefit? I can understand defaulting something to an _actual_ value, but not sure I get setting to null!

LeKoArts: I don't have strong feelings about whether using them or not, so whatever the majority thinks. I *personally* use it this way.

docs/docs/add-seo-component.md
```diff
@@ -0,0 +1,199 @@
...
+  description: null,
+  image: null,
+  pathname: null,
+  article: false,
+}
+```
+
+## Examples
+
+You could also put the Facebook and Twitter meta-tags into their own components, add custom favicons you placed in your `static` folder, and add JSONLD data. To see how that works you can have a look at these two examples:
```
jlengstorf: Probably worth linking to more info on JSON-LD here, since most people probably don't know what that is.

It also might be better to talk about schema.org instead of JSON-LD, since schema.org _uses_ JSON-LD to read SEO data, but JSON-LD is a spec that can be used for things other than SEO.

docs/docs/add-seo-component.md
LeKoArts: @dschau Ah, didn’t think of, but my eslint will complain if I don’t set defaultProps when the other props are not `.isRequired`, so technically this would be the more correct way I guess

calcsam:
We may also want to link this from https://www.gatsbyjs.org/docs/seo (and link to that page from this one). 

This can probably happen in a follow-up PR though -- I know @amberleyromo is in and out the next couple days and @shannonbux is travelling, so with everyone's eyes on this I'm going to go ahead and merge. 

Thanks @LekoArts!

docs/docs/add-seo-component.md
```diff
@@ -0,0 +1,209 @@
...
+      const seo = {
+        title: title || defaultTitle,
+        description: description || defaultDescription,
+        image: `${siteUrl}${image || defaultImage}`,
+        url: `${siteUrl}${pathname || "/"}`,
+      }
+
+      return (
+        <>
+          <Helmet>
```
spences10: Should the `titleTemplate` be included here?

```js
<Helmet title={seo.title} titleTemplate={titleTemplate}>
```

Issue #9005, opened 2018-10-10
Dedup deploy docs


yeskunall:
## Summary

There's a lot of duplication in the deploy section of Gatsby docs. For reference, please lookie at these links:

[/docs/deploy-gatsby/#heroku](https://www.gatsbyjs.org/docs/deploy-gatsby/#heroku)
[/docs/deploying-to-heroku](https://www.gatsbyjs.org/docs/deploying-to-heroku)

[/docs/deploy-gatsby/#now](https://www.gatsbyjs.org/docs/deploy-gatsby/#now)
[](https://www.gatsbyjs.org/docs/deploying-to-now)

[/docs/deploy-gatsby/#aerobatic](https://www.gatsbyjs.org/docs/deploy-gatsby/#aerobatic)
[/docs/deploying-to-aerobatic](https://www.gatsbyjs.org/docs/deploying-to-aerobatic)

### Motivation

Not really a follow-up of #8973, but the [feedback](https://github.com/gatsbyjs/gatsby/pull/8973#discussion_r224195654) @DSchau gave me made me look into this further, and I see that this issue exists in a few more places. Since we're aiming to reduce duplication, this would be a nice start to do so!

LMK your thoughts! 💖

kakadiadarpan:
@yeskunall this would be awesome! In my opinion, we should keep all the deployment documentation in a single place and that would be at [/docs/deploy-gatsby](https://www.gatsbyjs.org/docs/deploy-gatsby) and then have all the services listed there.

`/docs/deploy-gatsby#deployment_service`

@DSchau what do you think?

yeskunall:
Hey @kakadiadarpan 👋🏽


Sorry, I had a midterm today. Just got out of it.

I'm not too sure I understand what you mean by that. Gatsby already has `/docs/deploy-gatsby#deployment_service` for all of its services at [/docs/deploy-gatsby](https://www.gatsbyjs.org/docs/deploy-gatsby). What I was suggesting was to remove the duplication seen in for example, let's say, [/docs/deploy-gatsby#now](https://www.gatsbyjs.org/docs/deploy-gatsby/#now) and [/docs/deloying-to-now](https://www.gatsbyjs.org/docs/deploying-to-now), as seen in #8973. I hope I'm making sense here. If not, please bear with me &mdash; still recovering from the midterm 😅

kakadiadarpan:
[...]

PR #9079, opened 2018-10-13
Delete duplicate

namhsuya:
Signed-off-by: namhsuya <ayushmaan.banerjee@gmail.com>

deleted: /docs/deploying-to-heroku
duplicate: /docs/deploy-gatsby/#heroku


namhsuya:
Referenced from: [https://github.com/gatsbyjs/gatsby/issues/9005](#9005 )

DSchau:
@namhsuya thank so much for the PR!

I'm going to close this if that's alright, because I think the strategy we want to follow is going to be slightly different.

Kunall (@yeskunall) said it better than I can, so check out [his reply)(https://github.com/gatsbyjs/gatsby/issues/9005#issuecomment-429523418). In effect, we'd like to keep the "duplicates" and reserve them for adding additional information that isn't present in the deploy doc. This will keep the deploy doc light and eas(ier) to maintain, while also providing an avenue to add additional, specific content to the individual deploy documents. Thanks again!

amberleyromo:
I believe @shannonbux's motivation also included SEO on this! Easier for people to find when searching.

PR #9094, opened 2018-10-14
Fix typos and grammar on CONTRIBUTING_ES.md

imagentleman:
Noticed that there are actually some spanish translations of the docs. So, took a quick look a fixed a couple of typos and a little bit of the grammar.

A few thoughts:
The copy seems to be already outdated compared to the english original (https://www.gatsbyjs.org/docs/how-to-contribute/). Not sure what's the strategy here to keep everything in sync, and how are these translations being used. Are they supposed to be read here on github? Or are they gonna be published eventually on the website (gatsby.org/es or something like that).

There are no docs about how to contribute to the translation efforts (not just how to make the changes, but what tone and language specifics to you use, so things are written consistently and in the most universally accepted way). Translating is hard to get right.

Are spanish native speakers reviewing the spanish prs? Based on this pr (https://github.com/gatsbyjs/gatsby/pull/9063 they are not : S).

Hopefully the translations can take off. Just feels a little bit... like it's missing some direction.
Anyway, keep the good work. Cheers!

### DIFF
--- a/translations/es/CONTRIBUTING_ES.md
+++ b/translations/es/CONTRIBUTING_ES.md
@@ -1,12 +1,12 @@
 ---
-título: Cómo contribuir
+título: ¿Cómo contribuir?
 ---
 
 ## Contribución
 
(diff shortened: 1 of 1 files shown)


pieh:
IMO we just need to remove those, they will always be outdated and having them in cause more harm than good. We can't even actually review translations (other than checking google translate results) :(

bsypher:
Argentino.

Issue #7399, opened 2018-08-17
gatsby-plugin-typography next not live reloading changes to config file. 


mattmillerart:
## Description

Using `"gatsby-plugin-typography": "next"` with a config file in `utils/typography.js`, after saving changes these changes are not reflected in browser. The user must stop the `gatsby develop` process and restart to see the changes reflected. 

Interestingly, using `gatsby next` but using the older `"gatsby-plugin-typography": "^1.7.19",` plugin in package.json, things work as expected and no incompatibility issues seem to exist from using a 1.x module with 2.x

### Steps to reproduce

- Install `"gatsby-plugin-typography": "next"`
- add `utils/typography.js` file, mine has the following: 

```
import Typography from 'typography';
import './../fonts/DDCHardware-Regular.otf';
import './../fonts/DDCHardware-Condensed.otf';
import './../fonts/DDCHardware-Compressed.otf';

const typography = new Typography({
  title: 'Makerfaire',
  scaleRatio: 8,
  headerFontFamily: ['DDCHardware-Regular'],
  bodyFontFamily: ['Open Sans'],
  overrideStyles: ({ adjustFontSizeTo, rhythm }, options, styles) => ({
    h1: {
      textShadow: `2px 2px 0 red`
    },
    'h1 > span': {
      fontSize: rhythm(3)
    }
  })
})

export default typography
```
- run `gatsby develop`
- make a change in the above config file and save. 

### Expected result

Browser should live reload with changes

### Actual result

No live reload, no changes. User must stop and restart `gatsby develop` to see changes.

### Environment

```

  System:
    OS: macOS High Sierra 10.13.4
    CPU: x64 Intel(R) Core(TM) i5-4258U CPU @ 2.40GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 8.11.4 - /usr/local/bin/node
    npm: 5.6.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 68.0.3440.106
    Firefox: 27.0
    Safari: 11.1
  npmPackages:
    gatsby: next => 2.0.0-beta.110
    gatsby-plugin-manifest: next => 2.0.2-beta.6
    gatsby-plugin-offline: next => 2.0.0-beta.9
    gatsby-plugin-react-helmet: next => 3.0.0-beta.4
    gatsby-plugin-sass: next => 2.0.0-beta.10
    gatsby-plugin-typography: ^1.7.19 => 1.7.19
[...]

PR #9099, opened 2018-10-14
Add hotreload of typography.js to tutorial

### LINKED ISSUE #9095: Changes to Typography.js aren't applied without server restart
## Description

When following tutorial 2, [line 240](https://github.com/gatsbyjs/gatsby/blame/master/docs/tutorial/part-two/index.md#L240) implies we should be able to make a change to typography.js configuration and see the changes immediately. 

This has not been working for us when following tutorial 2. 

Additioanlly, we've tried "gatsby new " with the default starter, disabling layout.css and then configuring typography.js but the problem persists. 

If we restart the gatsby server the new changes take effect. 

Notably, if we look at the console output of the development server, nothing is emmited when we alter typography.js (where we configure typography). Perhaps it is being ignored by some watcher ? 
Also, if we hotswap themes instead of hand crafted styled, the changes **do** take effect

### Steps to reproduce
Follow the [second part of the tutorial](https://www.gatsbyjs.org/tutorial/part-two) upto [here](https://www.gatsbyjs.org/tutorial/part-two/#component-css). 

Clear steps describing how to reproduce the issue. Please please please link to a demo project if possible, this makes your issue _much_ easier to diagnose (seriously).

### Expected result

When changing
```js
src/utils/typography.js
import Typography from "typography"

const typography = new Typography({ baseFontSize: "18px" })

export default typography
```
to 

```js
src/utils/typography.js
import Typography from "typography"

const typography = new Typography({ baseFontSize: "24px" })

export default typography
```

Fonts should get bigger
### Actual result
Everything stays the same

### Environment

  System:
    OS: Linux 4.15 Ubuntu 16.04.3 LTS (Xenial Xerus)
    CPU: x64 Intel(R) Core(TM) i7-6500U CPU @ 2.50GHz
    Shell: 4.3.48 - /bin/bash
  Binaries:
    Node: 10.10.0 - /usr/local/bin/node
  Browsers:
    Chrome: 62.0.3202.94
    Firefox: 62.0.3
  npmPackages:
    gatsby: ^2.0.19 => 2.0.19 
    gatsby-plugin-manifest: ^2.0.5 => 2.0.5 
    gatsby-plugin-offline: ^2.0.5 => 2.0.5 
    gatsby-plugin-typography: ^2.2.0 => 2.2.0 
  npmGlobalPackages:
    gatsby-cli: 2.4.2

talolard:
Per #9095, the way the tutorial is set up, hot reloading of typography.js doesn't work without the extra lines I added (credit to @LeKoArts ). Newcomers like myself expect that it will hotreload because the tutorial implies it does, and also, I think, it's the "Gatsby  way"
  Learn more about contributing: https://www.gatsbyjs.org/docs/how-to-contribute/

closes #9095

### DIFF
--- a/docs/tutorial/part-two/index.md
+++ b/docs/tutorial/part-two/index.md
@@ -52,6 +52,12 @@ const typography = new Typography({
   ],
   bodyFontFamily: ["Georgia", "serif"],
 })
+
+// Hot reload typography in development.
+if (process.env.NODE_ENV !== 'production') {
+  typography.injectStyles();
+}
+
 ```
 
 ## Gatsby plugins
@@ -215,6 +221,10 @@ In your site, create a new directory at `src/utils`. In that directory, create a
 import Typography from "typography"
 
 const typography = new Typography({ baseFontSize: "18px" })
+// Hot reload typography in development.
+if (process.env.NODE_ENV !== 'production') {
+  typography.injectStyles();
+}
 
 export default typography
 ```
@@ -259,6 +269,10 @@ import Typography from "typography"
 import bootstrapTheme from "typography-theme-bootstrap"
 
 const typography = new Typography(bootstrapTheme)
+// Hot reload typography in development.
+if (process.env.NODE_ENV !== 'production') {
+  typography.injectStyles();
+}
 
 export default typography
 ```


pieh:
Could this be added straight to `gatsby-plugin-typography` so users don't need to worry about that?

ahinrichs:
Hot reload was included in this plugin up to #5219. If you revert this commit it works again without adding the extra lines to plugin config.

ahinrichs:
Yes. As it was, still is in v1 and should be. I'm not sure whether @KyleAMathews removed this by intent or by accident. Maybe there are some side effects of doing it in plugins `gatsby-browser.js` in v2 I'm not aware of. It seems to work equally either way: adding it to `src/utils/typography.js` or in plugins `gatsby-browser.js`.

dpyzo0o:
Just to mention, if you only export typography from typography.js, you still don't get hot reload working with the extra lines. You have to export something(rhythm, scale for example) and import them somewhere in you code to make the hot reloading work.

shannonbux:
@talolard and @ahinrichs is this PR still active or was it addressed in #7399?

LekoArts:
The change is not needed anymore :)
https://github.com/gatsbyjs/gatsby/pull/10545

PR #9100, opened 2018-10-14
feat(gatsby-transformer-sharp): allow screenshot placeholder

pieh:
Developing and building gatsbyjs.org is quite painful because of all the screenshots that are needed to be fetched and processed. It also doesn't allow us to get deploy previews for PRs against `www`. This PR give option to use `GATSBY_SCREENSHOT_PLACEHOLDER=true gatsby build/develop` and skip fetching and processing of all screenshots which mean that:
 a) local development can be done faster if screenshots dev is not done for showcase related areas
 b) we can get deploy previews from netlify again for PRs (would need to update netlify site to use `GATSBY_SCREENSHOT_PLACEHOLDER` env var

Here's how site showcase looks like when using placeholder from this PR:
localhost_9000_showcase_

Initial gatsby develop using that was ready to use in ~5 minutes on my local machine (there are still lot of images to process in blog and docs) as opposed to 15-20 minutes if we would need to download and process screenshots.


packages/gatsby-transformer-screenshot/src/gatsby-node.js
```diff
@@ -118,32 +117,45 @@ const createScreenshotNode = async ({
...
-      createNodeId,
-    })
+    let fileNode, expires
+    if (USE_PLACEHOLDER_IMAGE) {
+      const getPlaceholderFileNode = require(`./placeholder-file-node`)
+      fileNode = await getPlaceholderFileNode({
+        createNode,
+        createNodeId,
+      })
+      expires = new Date(2999, 1, 1).getTime()
```
pieh: hope that's enough :O

packages/gatsby-transformer-screenshot/src/placeholder-file-node.js
```diff
@@ -0,0 +1,19 @@
+const path = require(`path`)
+const { createFileNode } = require(`gatsby-source-filesystem/create-file-node`)
```
pieh: this could be cleaner if we would export that from `gatsby-source-filesystem` - probably it would need other name as it doesn't really create nodes - just prepare node object

pieh:
I think i've used this https://dummyimage.com/ to generate placeholder - will just adjust colors to use Gatsby color palette ;)

pieh:
Updated placeholder, still not great but probably looks better (used grayish background so screenshot cards stand out a little from white background:
localhost_8000_showcase_

It's just a image file packed in screenshot package, so it can always be updated in the future.

pieh:
TO-DO here: update `www/readme` to show how to make use of that - potentially we could add `.env.development` with that env var set in the future if this proves to be working well for people

KyleAMathews:
In the future, we should move screenshots to a graphql service that we can stitch in. That way images could be hosted on an external service and never need downloaded or processed locally.

packages/gatsby-transformer-screenshot/README.md
```diff
@@ -83,3 +83,19 @@ First, you will need to [create a S3 bucket](https://docs.aws.amazon.com/AmazonS
...
 To set up the HTTP interface, you will need to use AWS API Gateway. Create a new API, create a new resource under `/`, select "Configure as proxy resource", and leave all the settings with their defaults. Create a method on the new resource, selecting "Lambda Function Proxy" as the integration type, and fill in the details of your lambda.
+
+## Placeholder image
+
+If your site pulls a lot of screenshots it might be beneficial to use placeholder image instead of downloading and processing all the screenshots. It will help with data sourcing and query running times.
+
+You can use placeholder image by setting `GATSBY_SCREENSHOT_PLACEHOLDER` environment variable when running `gatsby develop`:
+
+```shell
+GATSBY_SCREENSHOT_PLACEHOLDER=1 gatsby develop
```
DSchau: Can we use `GATSBY_SCREENSHOT_PLACEHOLDER=true`? I'm not a huge shell scripter, but that seems to be clearer?

packages/gatsby-transformer-screenshot/README.md
```diff
@@ -83,3 +83,19 @@ First, you will need to [create a S3 bucket](https://docs.aws.amazon.com/AmazonS
...
+```
+
+or by adding it to `.env.development` file in root of your project:
+
+```shell:title=.env.development
```
DSchau: This title syntax isn't supported here :( Wish it was everywhere, haha.

packages/gatsby-transformer-screenshot/README.md
pieh: yeah, I'll switch to that - seems cleaner indeed. It just need to be truthy

packages/gatsby-transformer-screenshot/README.md
pieh: README for packages is shown in plugin library on gatsbyjs.org:
screen shot 2018-10-15 at 22 17 16

PR #9126, opened 2018-10-15
feat(gatsby-cli): add a clean command to wipe out local dirs

### LINKED ISSUE #6755: Add a snapshot-and-reset option to the CLI
When things go weird with Gatsby, it's often required to delete the cache and start with a clean slate: `rm -rf node_modules .cache public && yarn && yarn develop`

It's been requested to add a feature to clear the cache to the CLI, but there's been resistance to this because we're unable to fix the cache if issues never get reported.

In a discussion with @calcsam, @pieh, and @KyleAMathews, @pieh suggested a "reset, but with a snapshot sent to the OSS team" option.

This solves two problems:

1. Developers aren't blocked by a bad cache
2. The maintainers receive information that helps fix cache issues

## How it could work:

For now, I'm (semi-seriously) calling this command `gatsby nuke`. It does the following:

- takes an anonymous snapshot of the dev environment
- asks an optional question, e.g. “What happened that made a nuke necessary?”
- sends anonymized data to a public log for maintainers to review
- deletes `.cache`, `public`, `node_modules`
- runs `yarn` or `npm install`

## Questions to answer:

1. What details should be included in the snapshot?
    - `envinfo`
    - ...what else?
2. How is the snapshot structured?
3. How do we anonymize data?
    - This will depend on what data we need to include.
4. Where are snapshots logged?
5. What should the command be called?
    - `gatsby nuke`
    - `gatsby clean`
    - `gatsby cache-reset`
    - ...something else?
6. What optional question(s) are most helpful to maintainers?
    - What happened that made nuking the cache necessary?

DSchau:
Fixes #6755

This PR adds a `gatsby clean` command as a last resort when things have gone awry. We'll need to specifically document use cases when it could make sense to run this, or just emphasize that it's only to be used in certain cases.

Things left to do:

1. Improve logging (it's fine now)
1. Integrate prompt for more info
1. Integrate envinfo for env info
1. Integrate opt-out/opt-in analytics to gatsby CLI
1. Store the prompt/envinfo for later investigation
    - I don't know _where_ to put this. Do we have a good pattern for storing data like this?


packages/gatsby/src/commands/clean.js
```diff
@@ -0,0 +1,51 @@
...
+  try {
+    execSync(`${YARN_COMMAND} --version`, { stdio: `ignore` })
+    return true
+  } catch (e) {
+    return false
+  }
+}
+
+// Executes `npm install` or `yarn install` in rootPath.
+const install = async ({ directory: rootPath, report }) => {
```
DSchau: I should encapsulate/re-use this, since it's just copy pasted.

packages/gatsby/src/commands/clean.js
```diff
@@ -0,0 +1,51 @@
...
+
+  report.info(`Deleting ${directories.join(", ")}`)
+
+  await Promise.all(
+    directories.map(dir => fs.remove(path.join(directory, dir)))
+  )
+
+  report.info(`Successfully deleted local directories`)
+
+  await install(args)
```
DSchau: Add a `no-install` argument, maybe?

DSchau:
@jlengstorf if you have some time, could you check this out and chime in on any of the questions! I think this could be some nice functionality for debugging, helping people, etc.

packages/gatsby-cli/src/index.js
```diff
@@ -17,9 +17,9 @@ const updateNotifier = require(`update-notifier`)
 // Check if update is available
 updateNotifier({ pkg }).notify()
 
-if (verDigit < 4) {
+if (verDigit < 6) {
   report.panic(
-    `Gatsby 1.0+ requires node.js v4 or higher (you have ${version}). \n` +
+    `Gatsby 1.0+ requires node.js v6 or higher (you have ${version}). \n` +
```
DSchau: Should change this to use the version of Gatsby too.

pieh: Yhm, this is tricky, because `gatsby-cli` works with both gatsby 1 and gatsby 2 - I feel like this check should be done in gatsby core and not here ;) But I guess we set babel to target node 6 to build `gatsby-cli` now, so maybe that's just moot point

DSchau:
Per @pieh perhaps tarball the .cache directory so we can investigate further.

Send this off to a REST API (or something) that we're in control of

packages/gatsby-cli/src/index.js
jlengstorf: We could also do a breaking change for the CLI and tell people on Gatsby V1 not to upgrade to the next major version. That way the CLI can support _only_ V2.

packages/gatsby-cli/README.md
```diff
@@ -16,68 +16,79 @@ package installed locally.
...
+| :--------------: | ---------------------------------------------------------------------------------------- | :---------: |
+|  `-H`, `--host`  | Set host.                                                                                | `localhost` |
+|  `-p`, `--port`  | Set port.                                                                                |   `8000`    |
+|  `-o`, `--open`  | Open the site in your (default) browser for you                                          |             |
+| `--prefix-paths` | Serve site with link paths prefixed (if built with pathPrefix in your gatsby-config.js). |   `false`   |
 
-### Info
+### `clean`
+
+At the root of a Gatsby site run `gatsby clean` to wipe out `node_modules`, cache, and the `public` directory. This is useful as a last resort when your local project seems to have many issues.
```
jlengstorf: This would maybe benefit from more of an explanation on _why_ it's a last resort. Are there other things people should try first? Where can they go for more info? How can they help improve the cache so this isn't necessary?

I don't know if we have answers, but it'd be great to include whatever we can here.

KyleAMathews:
Are we adding the snapshot functionality here?

packages/gatsby-cli/README.md
DSchau: Great suggestion!

DSchau:
@KyleAMathews I didn't add it, because I didn't know _where_ it should go, whether we have an API in place, etc. I think the functionality is worth introducing regardless, but it'd be helpful to have some debug info to help us troubleshoot things in the future, so I'm open to releasing it as is or building in some type of API call.

jlengstorf:
My vote would be to release this and add a follow-up issue to design the API for sending snapshots on clean. Maybe we can add that to the new GraphQL API I just launched? See https://github.com/gatsbyjs/api.gatsbyjs.org/pull/11 for the new code; we could add a mutation in there to accept the snapshot and log it somewhere.

DSchau:
@pieh you wanna give this a once over and maybe we can get this merged today?

packages/gatsby-cli/src/index.js
DSchau: @pieh and I discussed this, few things:

- We should improve this (in a separate PR)
- Gatsby cli (v1.1.58 at least) doesn't work with Node 4 anyways, so this change is reasonable
  - yurnalist is not compatible

I'm in favor of the breaking change, and also discussed perhaps documenting using npm scripts, e.g. `npm run develop`, `npm run build`, etc. which will use the project local version of Gatsby instead of global. That said, there are some concerns there, e.g. if a community starter doesn't include those scripts.

So more to do here, but relatively tangential to this PR!

packages/gatsby-cli/README.md
```diff
@@ -16,44 +16,44 @@ package installed locally.
...
-|     Option      | Description                                     |
-| :-------------: | ----------------------------------------------- |
-| `-H`, `--host`  | Set host. Defaults to localhost                 |
-| `-p`, `--port`  | Set port. Defaults to 8000                      |
-| `-o`, `--open`  | Open the site in your (default) browser for you |
-| `-S`, `--https` | Use HTTPS                                       |
+|     Option      | Description                                     |   Default   |
+| :-------------: | ----------------------------------------------- | :---------: |
+| `-H`, `--host`  | Set host.                                       | `localhost` |
+| `-p`, `--port`  | Set port.                                       |   `8000`    |
```
wardpeet: not sure but you probably want this to be aligned with `localhost`

wardpeet:
@DSchau @jlengstorf @KyleAMathews It's just personal feedback, so don't read this as blocking.

I would introduce 2 commands `gatsby clean` which cleans the .cache and public folder. `gatsby nuke` will do the same as clean but also removes the node_modules folder. I'm not sure if we have to remove node_modules that often to fix bugs.

Npm/yarn install can be slow, depending on the hardware and OS so that's why i'm a bit hesitant.

it might be a good idea for another PR to ask users to capture data, error reporting, ...

For lighthouse we ask you: `May we anonymously report runtime exceptions to improve the tool over time?` We use this data to capture it to getsentry. (sorry for always bringing up lighthouse :p)

packages/gatsby/src/commands/clean.js
```diff
@@ -0,0 +1,50 @@
...
+
+// Executes `npm install` or `yarn install` in rootPath.
+const install = async ({ directory: rootPath, report, useYarn }) => {
+  const prevDir = process.cwd()
+
+  report.info(`Installing packages...`)
+  process.chdir(rootPath)
+
+  try {
+    let cmd = useYarn ? spawn(YARN_COMMAND) : spawn(`npm install`)
```
wardpeet: is it worth abstracting spawn as a function? (calling execa directly seems less complex)
```
let cmd = useYarn ? execa(YARN_COMMAND) : execa(`npm`, [`install``])
```

packages/gatsby-cli/README.md
DSchau: What did you mean here? This is from prettier, don't have a ton of control over alignment here!

packages/gatsby/src/commands/clean.js
```diff
@@ -0,0 +1,40 @@
+const fs = require(`fs-extra`)
+const execa = require(`execa`)
+const path = require(`path`)
+
+const YARN_COMMAND = `yarnpkg`
+
+const install = async ({ directory: rootPath, report, useYarn }) => {
```
ZYSzys: > I should encapsulate/re-use this, since it's just copy pasted.

Yeah, maybe we should extract it as a helper since it's almost  the same as `init-starter.js` ?
https://github.com/gatsbyjs/gatsby/blob/2032147f23d40fe8d986fd33996a8c5362e6c75d/packages/gatsby-cli/src/init-starter.js#L30-L43

packages/gatsby-cli/README.md
wardpeet: seems like it got fixed, it seems like localhost was not aligned properly but now it is

packages/gatsby/src/commands/clean.js
wardpeet: might be a good idea

DSchau: The issue is that if we make changes, we need to ensure those changes are available in both gatsby/gatsby-cli. For instance, if we make this change and encapsulate a reusable helper--we are then coupling some version of gatsby with some version of gatsby-cli (depending on where we place the helper).

If you look through some of the commits, this is something @pieh noted, so it led to a small re-factor.

That said, I agree--it sucks duplicating this, but it's a little safer for now. When we're ready to make a breaking change, I think there are a lot of improvements we can make to the CLI that

ZYSzys: Is it feasible to put the `install` function into `create-cli.js`, then pass it down ?
https://github.com/gatsbyjs/gatsby/blob/35a0a800ff744022e7aa7cdb4fd894db0b070e3c/packages/gatsby-cli/src/create-cli.js#L94

```diff
- let args = { ...argv, ...siteInfo, useYarn }
+ let args = { ...argv, ...siteInfo, useYarn, install }
```

DSchau: Yes - except that's a bit strange and I honestly prefer duplicating here.

Also then this breaks if gatsby-cli is < the version where this feature was introduced (which is a pretty common use case?)

ZYSzys: Let's get this PR move forward👍.

DSchau:
@wardpeet want to give this another once over? I scaled it back quite a bit--think it's ready and generally applicable now.

@marcysutton I'll look at docs to see how to best document this feature (troubleshooting?) but let me know if you have any ideas.

rgiese:
Just wanted to voice my support for @wardpeet 's comment above re: separating wiping .cache/public from also wiping node_modules. I've often needed to delete .cache (e.g. when messing with webpack config, gatsby-node.js, etc.) but I've never had to delete node_modules.

I think building a command into Gatsby that wipes Gatsby's state (.cache, public) makes a lot of sense. Building a command into Gatsby that wipes someone else's (npm/yarn's) state feels... less clean?

DSchau:
@rgiese that behavior you describe is exactly how it works :)

I reverted the node_modules deletion in 32a952b

marcysutton:
@DSchau for docs the "Debugging" section seems like a logical place. But I'm wondering if maybe the section should be renamed "Troubleshooting Gatsby" so we can add things like `gatsby clean` in there and make it sound less like Gatsby is full of bugs and more about troubleshooting project state. @shannonbux do you have any opinions on a section rename? Or is there a better spot for it? https://www.gatsbyjs.org/docs/debugging/

shannonbux:
@marcysutton I don't have an opinion; I would just call it whatever seems like the most common / understandable way developers describe the tasks involved in the section. The reason we called it debugging was mostly because other platforms have a "debugging" section, so it is standard (doesn't mean it's the best name, though :) 

Here are relevant blog posts about where the current organization of the docs came from:
https://www.gatsbyjs.org/blog/2018-06-26-card-sort-results/
https://www.gatsbyjs.org/blog/2018-07-31-docs-redesign/

DSchau:
@shannonbux @marcysutton see f24aa24. Happy to make any changes, I put it together pretty quickly.

packages/gatsby/src/commands/clean.js
```diff
@@ -0,0 +1,16 @@
+const fs = require(`fs-extra`)
+const path = require(`path`)
+
+module.exports = async function clean(args) {
+  const { directory, report } = args
+
+  const directories = [`.cache`, `public`, `node_modules`]
```
sidharthachatterjee: Shouldn't this exclude `node_modules`?
```suggestion
  const directories = [`.cache`, `public`]
```

packages/gatsby-cli/src/create-cli.js
```diff
@@ -247,6 +247,12 @@ function buildLocalCommands(cli, isLocalSite) {
     },
   })
 
+  cli.command({
+    command: `clean`,
+    desc: `Wipe the local gatsby environment when something has gone wrong`,
```
sidharthachatterjee:
```suggestion
    desc: `Wipe the local gatsby environment including built assets and cache`,
```

Probably makes sense to make this description independent of the typical use case?

sidharthachatterjee:
Successfully published:
 - gatsby-cli@2.4.11
 - gatsby@2.1.8

Issue #7289, opened 2018-08-13
[v2] Pages render briefly then turn blank in non-local environments


jrskerritt:
## Description

I'm running into a strange issue where my pages render as expected for a brief moment, but then the content is removed from the DOM within a few milliseconds. This only happens in our CI environment; locally I can test pages fine and they render without the flashing/flickering.

On debugging, it looks like the parent node of all of the gatsby content is being removed for some reason after the initial render:

```
<div id="___gatsby">
    <div style="outline:none" tabindex="-1" role="group">     <-- this gets removed
        <!-- all page content -->
    </div>
</div>
```

`gatsby build` runs as expected without errors.

I've tried changing versions of various packages (gatsby, plugins, webpack) to see if it resolves, with no luck. Also tried commenting out several sections of React code and plugins to try to identify what might be causing this, without any luck.

Any idea what could cause gatsby to remove its content from the mount node?

### Environment

System:
    OS: Windows 10
    CPU: x64 Intel(R) Core(TM) i7-5820K CPU @ 3.30GHz
  Binaries:
    Yarn: 1.7.0 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.3.0 - C:\Program Files\nodejs\npm.CMD
  npmPackages:
    gatsby: 2.0.0-beta.98 => 2.0.0-beta.98
    gatsby-module-loader: 2.0.0-alpha.3 => 2.0.0-alpha.3
    gatsby-plugin-google-tagmanager: 2.0.0-beta.3 => 2.0.0-beta.3
    gatsby-plugin-react-helmet: 3.0.0-beta.4 => 3.0.0-beta.4
    gatsby-plugin-sass: 2.0.0-beta.10 => 2.0.0-beta.10
    gatsby-source-filesystem: 2.0.1-beta.10 => 2.0.1-beta.10
    gatsby-transformer-json: 2.1.1-beta.5 => 2.1.1-beta.5
    webpack: 4.16.5

### File contents (if changed)

gatsby-config.js:
```
module.exports = {
  pathPrefix: 'path/prefix',
  plugins: [
    'gatsby-plugin-sass',
    'gatsby-plugin-react-helmet',
    'gatsby-transformer-json',
    // gtm plugin, filesystem plugins
```

package.json:
```
"scripts": {
  // other scripts ...
  // This is what gets run on TeamCity
  "build": "npm test && npm run teamcity-lint && gatsby build --prefix-paths",
[...]

PR #9131, opened 2018-10-15
Allow configuration of asset file output location

jrskerritt:
See #7289

This PR adds a configuration for the output location of asset files (any JS, CSS, JSON, and source maps created by `gatsby build`). Check out the documentation I added in `asset-path.md` to learn how to use it.

tl;dr: where previously the `/public/static` and `/public/static/d` directories were being assumed, I updated to check for the `assetPath` value in `gatsby-config.js`. It behaves very similarly to `pathPrefix`.

## To Maintainers:
Definitely take a close look at the changes I made to the plugins to make sure I didn't accidentally break anything! I did my best, but they are plugins I haven't used before.


pieh:
What is the relation between assetPath and pathPrefix - I see You used `assetPath || pathPrefix` so it seems like if `pathPrefix` is used, then `assetPath` should contain it, otherwise it will not work correctly?

jrskerritt:
@pieh Correct -- my idea was to make `assetPath` an absolute path that was separate from `pathPrefix` but I suppose it might make more sense to concatenate them if both are configured. What do you think?

pieh:
Ultimately we would also want to have e2e cypress tests for this like ones we have for path-prefix

KyleAMathews:
Yeah... this feels like a feature that'd be hard to support for relatively little benefit. Wouldn't it be easier for you to change how you deploy your sites than for us to change a core part of gatsby's internals?

jrskerritt:
I understand why you'd be uncomfortable adding this. If stuff breaks later on, you guys would be the ones fixing it, not me. But unfortunately it's not possible for me to change how our site is deployed or how the urls are structured.

Without this feature, our only options would likely be either to maintain an internal fork of Gatsby or move away from Gatsby entirely. Both of these I'd rather not do 🙂

If there's anything else I could do to help potentially make this less fragile and give you guys more confidence, let me know. Happy to help.

DSchau:
@jrskerritt I'm not _all_ that familiar with Cloudfront, but I'd have to think this is a solvable problem outside of changes to Gatsby.

If we presume all _files_ are in S3 (e.g. have an extension) is it possible to create some type of rule that whitelists those? I'd think routes (e.g. /research/US/whatever) would implicitly fall to the S3 bucket, because that's how public/index.html is working, I'd assume?

Another (hackier) solution would be to deploy Gatsby at a unique path, e.g. `/research/US/static` and then do a path re-write on that directory when serving, which _could_ be possible.

I'd think both of those are totally possible, and would solve the problem in a way that's cleaner for us.

Anything I'm missing--apart from knowing whether either of those are possible? Like I said, not an AWS guru by any means!

jrskerritt:
@DSchau - I did some research into your suggestions.

In theory, we could forward only files with desired extensions under `/research` to our S3 origin, but my concern is that this could cause issues if other `.js` or `.css` or whatever files end up (or currently exist) in the same route that aren't in S3. A lot of our website I'd consider "legacy" and so I have no idea what else may exist under `/research`. So I'm not confident that this would work for us.

Rewriting _might_ be possible with lambda functions but that seems like overkill to me.

I can see that you guys aren't willing to budge on this feature so I'm just going to maintain my own fork. It's not ideal, but it's the next best path forward for me and my team.

DSchau:
> A lot of our website I'd consider "legacy" and so I have no idea what else may exist under /research

For sure! It sounded like it was largely API driven, so I figured whitelisting files with extensions seemed like a fair approach!

> Rewriting might be possible with lambda functions but that seems like overkill to me.

It honestly seems like a pretty tenable solution, but totally understand! I'd rather we were able to come to some compromise, but like I said, not an AWS expert so it's hard for me to know best practices from that context.

We appreciate the PR, and sorry we couldn't come to a consensus on this!

Issue #8858, opened 2018-10-06
Inferno v6 build succeed with gatsby


simonjoom:
Hello this is not an issue but an experience,
Just to say i succeed to make working gatsby and inferno together.

The only thing in dev inferno cannot use hot-reload but everything work well with some little hack in core. (just some alias and turn around hot-reload in webpack.conf)
the react16 context work well, just fragment is still not well supported
I can share my experience.
:)

KyleAMathews:
Hey nifty! You want to create and share a repo with all this working? That'd be fun to see!

simonjoom:
i can share sure. The thing is my repo is little big, i need to do one light type hello-world inferno

simonjoom:
Just done here. I made as simple as possible
https://github.com/simonjoom/gatsby-starter-inferno-master
Description to install and use in readme.
"npm run inferno" to start running inferno mode development
"npm run develop" run normal gatsby mode
"npm run build" build with inferno (no react)

simonjoom:
Hello now Fragment is supported in last release of inferno-v6-rc3

simonjoom:
Hello , ok now the stable version inferno-v6 is out )) correctly tested 
it's live with my gatsby repo published here: https://www.ski-courchevel.deals  ))

Issue #8545, opened 2018-09-26
[gatsby-plugin-preact] Uncaught TypeError: Object(...) is not a function


rwieruch:
## Description

I wanted to use gatsby-plugin-preact but I ran into an issue. So I created a minimal issue where it happens too.

### Steps to reproduce

Take the [gatsby-starter-blog](https://github.com/gatsbyjs/gatsby-starter-blog) (or I guess any other project), do the setup, and install [gatsby-plugin-react](https://www.gatsbyjs.org/packages/gatsby-plugin-preact/). [Here is a minimal application for this problem](https://github.com/rwieruch/gatsby-starter-blog-with-preact).

### Expected result

The application should work.

### Actual result

I get the following stack trace:

```
Users/mydspr/Developer/Repos/gatsby-blog/node_modules/@reach/router/es/index.js:61 Uncaught TypeError: Object(...) is not a function
    at createNamedContext (Users/mydspr/Developer/Repos/gatsby-blog/node_modules/@reach/router/es/index.js:61)
    at eval (Users/mydspr/Developer/Repos/gatsby-blog/node_modules/@reach/router/es/index.js:69)
    at Module../node_modules/@reach/router/es/index.js (commons.js:1176)
    at __webpack_require__ (commons.js:725)
    at fn (commons.js:102)
    at eval (Users/mydspr/Developer/Repos/gatsby-blog/node_modules/gatsby-link/index.js:23)
    at Object../node_modules/gatsby-link/index.js (commons.js:2608)
    at __webpack_require__ (commons.js:725)
    at fn (commons.js:102)
    at Module.eval (Users/mydspr/Developer/Repos/gatsby-blog/.cache/gatsby-browser-entry.js:8)
```

kakadiadarpan:
@rwieruch Thanks for providing us with the reproduction repo. I can confirm that `gatsby-plugin-preact` is breaking with Gatsby v2.

Gatsby uses `@reach/router` for routing, and it seems like it isn't compatible with `preact`. Check this [codesandbox](https://codesandbox.io/s/9jpv93q9nr) where only preact and `@reach/router` are being used.

reach/router#9

rwieruch:
@kakadiadarpan thanks for your response and providing a solution. Would you say this is the solution to go for or will the gatsby-preact-plugin get an update eventually and this is only a workaround?

kakadiadarpan:
[...]

PR #9138, opened 2018-10-16
Add new Gatsby plugin for Inferno support

DonnieWest:
Created as a proof of concept for #8858

You can see others who have had success with getting Inferno V6 running on Gatsby and I've also had some luck. Figured I'd essentially fork the Preact plugin and open this as a place for discussion. I'm willing to make any edits requested :)


packages/gatsby-plugin-inferno/package.json
```diff
@@ -0,0 +1,34 @@
...
+  "homepage": "https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-inferno#readme",
+  "keywords": [
+    "gatsby",
+    "gatsby-plugin",
+    "inferno"
+  ],
+  "license": "MIT",
+  "main": "index.js",
+  "peerDependencies": {
+    "gatsby": ">2.0.0-alpha"
```
pieh: We would need `inferno` and  `inferno-compat` as peerDeps too

packages/gatsby-plugin-inferno/.gitignore
```diff
@@ -0,0 +1,2 @@
+/gatsby-node.js
+/gatsby-browser.js
```
pieh: I think it should be something like:
```
/*.js
!index.js
yarn.lock
```
We don't want any `gatsby-` built file in git (including `gatsby-node`)

packages/gatsby-plugin-inferno/src/gatsby-node.js
```diff
@@ -0,0 +1,15 @@
+exports.onCreateWebpackConfig = ({ stage, actions }) => {
+  // Requiring the server version of React-dom is hardcoded right now
+  // in the development server. So we'll just avoid loading Inferno there
```
pieh: Was this copied from `preact` plugin or this is something new? I'm not sure what's the problem this check is workarounding

DonnieWest:
@simonjoom I'm getting around the hot reloading issue by just excluding Inferno during development like the Preact plugin does. It's definitely not an ideal solution, but I'd prefer to maintain consistency with how other plugins do things unless strictly necessary.

This was all I personally needed to do in order to switch to Inferno and seems pretty consistent with how [their docs tell you to migrate](https://github.com/infernojs/inferno/tree/master/packages/inferno-compat#usage-with-webpack) - unless I'm missing something, we shouldn't have to do anything else. For instance, switching the pragma isn't needed because Inferno provides `React.createElement`, right?

packages/gatsby-plugin-inferno/.gitignore
DonnieWest: I'll try to get to this tonight - thanks :)

packages/gatsby-plugin-inferno/package.json
DonnieWest: Would something like this work?

```
  "peerDependencies": {
    "inferno-compat": "6.x",
    "gatsby": ">2.0.0-alpha"
},
```

It's only Inferno 6 that works at this point

packages/gatsby-plugin-inferno/src/gatsby-node.js
pieh: `develop-html` is only for generating things for basic `index.html` used in develop mode. There is separate `develop` stage that actually produce js bundle, so not sure if this changes anything for hot-reload

pieh:
Just a heads up - `preact` plugin has some problems currently https://github.com/gatsbyjs/gatsby/issues/8545, so I wouldn't take how that one is doing as sure thing forward.

simonjoom:
> @simonjoom I'm getting around the hot reloading issue by just excluding Inferno during development like the Preact plugin does. It's definitely not an ideal solution, but I'd prefer to maintain consistency with how other plugins do things unless strictly necessary.
> 
> This was all I personally needed to do in order to switch to Inferno and seems pretty consistent with how [their docs tell you to migrate](https://github.com/infernojs/inferno/tree/master/packages/inferno-compat#usage-with-webpack) - unless I'm missing something, we shouldn't have to do anything else. For instance, switching the pragma isn't needed because Inferno provides `React.createElement`, right?
> 
> Edit: Maybe I'm not actually fixing the hot reloading issue. I'll verify tonight and get back

Hello
About development i think it s good to have an inferno development structure without hotreload

First because it's avoid you to deploy everytime your project to be able to reduce bug involving from inferno. 
Second to have a development more near "production" but keeping the HMR and sourcemap that is great. As well have a development mode with no react working (only inferno) is great to test some library like react-transition-group (It's working even with inferno)

so for me it's why i create a new  variable process.env.NODE_ENV=="inferno" 

In my case i use
" npm run develop" who give me exactly the same than before hacking with gastby 

"npm run inferno" give me my test production case  so process.env.NODE_ENV=="inferno"  here (hot reloaded avoided with this boolean) HMR working and no minifyjs

"npm run build" produce output with inferno inside without react + minifyjs

DonnieWest:
@simonjoom as mentioned, I agree it's not the ideal situation. This at least serves my personal use case and I don't think the loss of HMR is worth it in development

I'll let someone else (preferably from the Gatsby team) decide which route to go

pieh:
What is the problem with inferno and HMR - it straight doesn't work or it's buggy?

simonjoom:
> What is the problem with inferno and HMR - it straight doesn't work or it's buggy?

With inferno:
HMR is working 
the webpack-hot-middleware i mean work well and refresh the page after mods 

But it's the react-hot-loader who cannot work with. because react-hot-loader was done for react and still they did not release a inferno-hot-loader to replace it

pieh:
Ah, that makes sense - yeah then we should just refresh page in there is no equiavalent to `react-hot-loader` for inferno. But definitely should use inferno for development to keep it as close to production as possible

DonnieWest:
@simonjoom @pieh sounds good then. I'll release the code here as my own package since this is all I need and let @simonjoom take this effort from here :)

wardpeet:
@DonnieWest @simonjoom I'll close this one as if I read correct @DonnieWest is not going to pursuit this on the gatsby repo. @simonjoom  feel free to create a new PR with your own code.

You can also just publish it on your own and add it to our lovely plugin directory.
https://www.gatsbyjs.org/docs/submit-to-plugin-library/

Issue #9145, opened 2018-10-16
"gatsby-plugin-offline@2.0.6" is exporting a variable named "onPostPrefetc hPathname" which isn't an API


anilbhanushali:
I am following the tutorial to add offline support https://www.gatsbyjs.org/tutorial/part-eight/#add-offline-support

While building site using `gatsby build` I am getting below error. As mentioned in error, there is no exported method called "onPostPrefetchPathname" in browser-apis. may be this was renamed in version 2.0

```
Your plugins must export known APIs from their gatsby-browser.js.The following exports aren't APIs. Perhaps you made a typo or your plugin is outdated?

See https://www.gatsbyjs.org/docs/browser-apis/ for the list of Gatsby Browser APIs

- The plugin "gatsby-plugin-offline@2.0.6" is exporting a variable named "onPostPrefetc
hPathname" which isn't an API.

Perhaps you meant to export "onPrefetchPathname"?
```
Below is the environment info
```
System:
    OS: macOS 10.14
    CPU: x64 Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 10.11.0 - ~/.nvm/versions/node/v10.11.0/bin/node
    npm: 6.4.1 - ~/.nvm/versions/node/v10.11.0/bin/npm  Browsers:
    Chrome: 69.0.3497.100
    Firefox: 61.0.1
    Safari: 12.0
  npmPackages:
    gatsby: ^2.0.0 => 2.0.0
    gatsby-plugin-emotion: ^2.0.5 => 2.0.5
    gatsby-plugin-manifest: ^2.0.5 => 2.0.5
    gatsby-plugin-offline: ^2.0.6 => 2.0.6
    gatsby-plugin-react-helmet: ^3.0.0 => 3.0.0
    gatsby-plugin-typography: ^2.2.0 => 2.2.0
    gatsby-source-filesystem: ^2.0.3 => 2.0.3
    gatsby-transformer-remark: ^2.1.7 => 2.1.7
```

gaaabor:
I have exactly the same, i just wanted to post the issue.

GuyKh:
Updated to latest (2.0.24) in this PR

arnriu:
Still have this very same issue after updating.
My infos:

```
System:
    OS: macOS 10.14
    CPU: x64 Intel(R) Core(TM) i7-3635QM CPU @ 2.40GHz
[...]

PR #9150, opened 2018-10-16
Update Gatsby version to latest to solve issue 9145

GuyKh:
Update Gatsby version to latest to solve issue #9145

### DIFF
--- a/packages/gatsby-plugin-offline/package.json
+++ b/packages/gatsby-plugin-offline/package.json
@@ -28,7 +28,7 @@
   "license": "MIT",
   "main": "index.js",
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


packages/gatsby-plugin-offline/package.json
```diff
@@ -28,7 +28,7 @@
   "license": "MIT",
   "main": "index.js",
   "peerDependencies": {
-    "gatsby": ">2.0.0-beta"
+    "gatsby": ">=2.0.24"
```
pieh: It should be `2.0.20` (that's where this API was added)

GuyKh:
@pieh - ready :)

PR #9153, opened 2018-10-16
Ensure that File type is created first

stefanprobst:
This ensures that the File type is created before others.

I'm not sure if this change is acceptable, the motivation is to be able to import with `const FileType = require('gatsby/dist/schema/types/type-file').getType()` in `setFieldsOnGraphQLNodeType`. Without this change this would always return `null` as [`setFileNodeRootType`](https://github.com/gatsbyjs/gatsby/blob/fed6d35f991abadca5e4ab335597d0c045d882da/packages/gatsby/src/schema/build-node-types.js#L227) had not been called yet.

### DIFF
--- a/packages/gatsby/src/schema/build-node-types.js
+++ b/packages/gatsby/src/schema/build-node-types.js
@@ -229,7 +229,7 @@ module.exports = async ({ parentSpan }) => {
   }
 
   // Create node types and node fields for nodes that have a resolve function.
-  // Ensure that File nodes are created first, so FileType can be imported
+  // Ensure that File type is created first, so it can be imported
   // in setFieldsOnGraphQLNodeType.
   const { File: fileNodes, ...otherTypes } = types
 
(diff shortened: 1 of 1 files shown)


pieh:
Our whole special casing for `File` type in gatsby core is "weird" to say the least ;) 

I think what we could do is have some type of graphql type manager, so we can ask for any graphql type and either return it if it was already created or create it if it wasn't yet - this would hopefully make it less problematic - example of how this could be used in `setFieldsOnGraphQLNodeType`:

```js
exports.setFieldsOnGraphQLNodeType = ({ type, getGQLType }) => {
  if (type.name === `MarkdownRemark`) {
    return {
      newField: {
        // this would either return already created `File` type or create it if it wasn't created yet
        type: getGQLType(`File`),
        resolve: source => {
          // ... some logic to return file node
        }
      }
    }
  }

  // by default return empty object
  return {}
}
```

stefanprobst:
> some type of graphql type manager

Sounds good! Is this something being worked on already? If not, any pointers where best to start? Thanks!

PR #9160, opened 2018-10-16
fix(ci): use correct cache path per circleci documentation

DSchau:
This PR does the following things:

- Switches to an official cypress docker image (using node 10)
    - This should be a little lighter weight and more specific to Cypress
- persists the ~/.cache folder, which will include cypress + yarn caches
    - This should speed up install and subsequent run times, if yarn.lock doesn't change

Thanks @lekoarts for the tip!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,7 @@ aliases:
       name: Save node modules cache
       key: yarn-cypress-cache-{{ checksum "yarn.lock" }}
       paths:
-        - ~/.cache # yarn + cypress cache
+        - ~/.cache
 
   attach_to_bootstrap: &attach_to_bootstrap
     attach_workspace:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -9,28 +9,26 @@ executors:
...
       keys:
-        - node-modules-{{ checksum "yarn.lock" }}
+        - yarn-cypress-cache-{{ checksum "yarn.lock" }}
 
   install_node_modules: &install_node_modules
     run:
       name: Install node modules
-      command: |
-        yarn
+      command: yarn
```
DSchau: @LekoArts thoughts on using `--frozen-lockfile` here? Think there's any harm?

LekoArts: I think it doesn't hurt, right?
Arguments like [these](https://github.com/yarnpkg/website/issues/700#issuecomment-340993837) seem reasonable - but I didn't use it too long myself so don't know how it behaves at a larger scale.

PR #9167, opened 2018-10-16
[WIP]: ci: add a size limit task that checks the size of gatsby packages

DSchau:
This PR adds some prelim./basic bundle size information and CI integration for tracking on bundle sizes.

The way it works is that it uses [size-limit](https://github.com/ai/size-limit) to get the gzipped sizes of various packages. Currently it's opt-in, so any package that includes a `.size-limit.js` in `packages/{NAME_OF_PACKAGE}` will be available for tracking.

There are some issues (so far), mainly that I wasn't able to get the integration with webpack working with `gatsby/cache-dir`, and so I'm using a naive approach of grabbing the file size of all js files in the cache-dir, i.e. we're not including @reach/router, react, or any dependency of gatsby in that calculation.

Wanted to get some 👀 on this approach though.

### DIFF
--- a/packages/gatsby/.size-limit.js
+++ b/packages/gatsby/.size-limit.js
@@ -1,0 +1,7 @@
+module.exports = [
+  {
+    path: `cache-dir/*.js`,
+    webpack: false,
+    limit: `25 KB`,
+  },
+]
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,6 +103,7 @@ jobs:
     executor: node
     steps:
       - checkout
+      - run: ./scripts/assert-changed-files.sh "packages/*"
       - <<: *restore_node_modules
       - <<: *install_node_modules
       - <<: *attach_to_bootstrap
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,6 +99,15 @@ jobs:
       - <<: *persist_node_modules
       - run: yarn lint
 
+  bundle-size:
+    executor: node
+    steps:
+      - checkout
+      - <<: *restore_node_modules
+      - <<: *install_node_modules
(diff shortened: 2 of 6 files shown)


packages/gatsby/.size-limit.js
```diff
@@ -0,0 +1,7 @@
+module.exports = [
+  {
+    path: `cache-dir/*.js`,
+    webpack: false,
```
DSchau: So I _almost_ solved this, but ultimately kept running into babel issues. The approach I tried was:

- add a `.size-limit.config.js` webpack config file
- the webpack config file adds the babel loader with our preset
- add a `config: './.size-limit.config'` property to `.size-limit.js`
- failure...

Fails with

```
 ERROR  Size Limit can't resolve
        @babel/runtime/helpers/defineProperty
        in /Users/dschau/Code/Work/gatsby/gatsby/packages/gatsby/cache-dir
```

DSchau: Here's `.size-limit.config.js` for posterity

```js
const path = require(`path`)
const { module: browserModule } = require(`./package.json`)
const preset = require(`./babel.config`)

module.exports = {
  entry: path.join(__dirname, browserModule),
  module: {
    rules: [
      {
        test: /\.js$/,
        use: [
          {
            loader: require.resolve(`babel-loader`),
            options: preset
          }
        ]
      }
    ]
  }
}
```

DSchau:
There's _probably_ more I could do here, e.g. add build size warnings for examples, but I think this is a decent start.

DSchau:
Some type of testing mechanism of the bundled assets would be valuable, but I don't believe this is the best mechanism for doing so.

Closing!

PR #9171, opened 2018-10-16
cleanup www packages

pieh:
yup - again


pieh:
ops, wrong repo - don't mind me!

PR #9182, opened 2018-10-17
Used `path` to concat url in serializer

AileenCGN:
In case of using a trailing slash as the `siteUrl`, the `serialize` method returns a URL that contains double slashes. Using `path.join()` to concat the `siteUrl` with the `path` of the current node will prevent this.

### DIFF
--- a/packages/gatsby-plugin-sitemap/src/internals.js
+++ b/packages/gatsby-plugin-sitemap/src/internals.js
@@ -1,6 +1,7 @@
 import fs from "fs"
 import pify from "pify"
 import minimatch from "minimatch"
+import path from "path"
 
 const withoutTrailingSlash = path =>
   path === `/` ? path : path.replace(/\/$/, ``)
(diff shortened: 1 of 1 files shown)


packages/gatsby-plugin-sitemap/src/internals.js
```diff
@@ -59,7 +60,7 @@ export const defaultOptions = {
   serialize: ({ site, allSitePage }) =>
     allSitePage.edges.map(edge => {
       return {
-        url: site.siteMetadata.siteUrl + edge.node.path,
+        url: path.join(site.siteMetadata.siteUrl, edge.node.path),
```
pieh: This might yield wierd results on windows where path separator is `\`

AileenCGN: Damn! You're right. Didn't think of the windows part. In that case, my change doesn't make sense ofc. Sorry for the noise 😬

PR #9183, opened 2018-10-17
feat(www): add hubspot form to homepage again

fk:
Along the way:

* Use „Futura PT“ for form signup message
* Remove Hubspot form bottom margin

@KyleAMathews commented:

> Can we get this to work without loading all the hubspot js? We had this form then removed it because it was slowing our performance.

I frankly am a little stumped on how to approach Kyle's request. Maybe someone else can help?
/cc @gatsbyjs/website

### DIFF
--- a/www/src/components/email-capture-form.js
+++ b/www/src/components/email-capture-form.js
@@ -202,7 +202,9 @@ class EmailCaptureForm extends React.Component {
             }}
           >
             {this.state.successMessage ? (
-              <p>{this.state.successMessage}</p>
+              <div
+                dangerouslySetInnerHTML={{ __html: this.state.successMessage }}
+              />
             ) : (
               <Form
                 onSuccess={this.onSuccess}
--- a/www/src/components/email-capture-form.js
+++ b/www/src/components/email-capture-form.js
@@ -1,5 +1,5 @@
 import React from "react"
-import { rhythm } from "../utils/typography"
+import { rhythm, options } from "../utils/typography"
 import { colors } from "../utils/presets"
 import HubspotForm from "./hubspot-form"
 
@@ -11,6 +11,7 @@ class EmailCaptureForm extends React.Component {
       <div
         css={{
           borderTop: `2px solid ${colors.lilac}`,
+          fontFamily: options.headerFontFamily.join(`,`),
           marginTop: rhythm(3),
           paddingTop: `${rhythm(1)}`,
           ...overrideCSS,
--- a/www/src/components/email-capture-form.js
+++ b/www/src/components/email-capture-form.js
@@ -6,7 +6,7 @@ import { formInput } from "../utils/form-styles"
 import { buttonStyles } from "../utils/styles"
 
 const Label = props => (
-  <label>
+  <label htmlFor="newsletter-signup-email">
     {props.children}
     {props.isRequired && (
       <span
@@ -24,6 +24,7 @@ const Label = props => (
 
 const SingleLineInput = React.forwardRef((props, ref) => (
   <input
+    id="newsletter-signup-email"
     ref={ref}
     css={{
       ...formInput,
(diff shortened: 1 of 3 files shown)


fk:
What I managed to understand is that we are using

* https://github.com/escaladesports/react-hubspot-form
* @ThatOtherPerson added a [PR](https://github.com/escaladesports/react-hubspot-form/pull/14) for `react-hubspot-form` that `defer`s the Hubspot JS that got released as `react-hubspot-form@1.3.5` – which we are using on .org `www` (https://github.com/gatsbyjs/gatsby/blob/6ad15ab7b54d76975a7266be90b32de2d5ff9905/www/package.json#L68)
* `react-hubspot-form` loads `//js.hsforms.net/forms/v2.js` (https://github.com/escaladesports/react-hubspot-form/blob/a9168da6348e07d75860d600690054a8a318f074/src/index.js#L47)

ThatOtherPerson:
@fk I might have missed some context on this, but it should be possible to create a HubSpot plugin for Gatsby that renders HubSpot forms at build time.

DSchau:
@ThatOtherPerson it would still inline the Hubspot JS right? I think that's the main issue, it's something like 200-300Kb or something if I recall correctly!

So injecting at build time wouldn't really help us, but perhaps I'm off base too. If you could explain further, that'd be awesome!

fk:
@pieh is the MVP! 🎉 Thanks so much Michal! 💜

DSchau:
> @pieh is the MVP! 🎉

He really is! Nice work @pieh!

www/src/components/email-capture-form.js
```diff
@@ -1,9 +1,182 @@
...
 import { rhythm, options } from "../utils/typography"
-import { colors } from "../utils/presets"
-import HubspotForm from "./hubspot-form"
+import presets, { colors } from "../utils/presets"
+import hex2rgba from "hex2rgba"
+import { formInput } from "../utils/form-styles"
+import { buttonStyles } from "../utils/styles"
+
+const Label = props => (
+  <label>
```
pieh: I forgot to use `for` prop here

www/src/components/email-capture-form.js
```diff
@@ -1,9 +1,182 @@
...
+      <form onSubmit={this.onSubmit}>
+        <div
+          css={{
+            paddingBottom: `20px`,
+          }}
+        >
+          <Label isRequired for="email">
+            Email
+          </Label>
+          <SingleLineInput
```
pieh: and here I forgot to set `id` (so that `for` prop would actually work)

www/src/components/email-capture-form.js
DSchau: htmlFor, I forget it almost every time 🙃

www/src/components/email-capture-form.js
fk: 👍 🙏  Will fix!

www/src/components/email-capture-form.js
pieh: heh, I've initially used `for` and of course got complaints about syntax problem so as any dev I just deleted that and forgot about it :D

www/src/components/email-capture-form.js
```diff
@@ -1,28 +1,220 @@
...
-import { colors } from "../utils/presets"
-import HubspotForm from "./hubspot-form"
+import { rhythm, options } from "../utils/typography"
+import presets, { colors } from "../utils/presets"
+import hex2rgba from "hex2rgba"
+import { formInput } from "../utils/form-styles"
+import { buttonStyles } from "../utils/styles"
+
+const Label = props => (
+  <label htmlFor={props.for}>
```
DSchau: Does React complain about this in dev mode? Didn't know if `for` may be a reserved prop on _any_ component!

fk:
Took me two attempts 🙄, but should be good now for another look.

Issue #9159, opened 2018-10-16
Search is broken on v1 documentation site


dj:
## Description

When I search the v1 docs site, I get results for pages that only exist on v2. If I click them I get taken to an error page.

### Steps to reproduce

1. Visit https://v1.gatsbyjs.org/docs/
2. Focus search input, type in `createPage`, click first result
<img width="749" alt="screen shot 2018-10-16 at 11 55 20 am" src="https://user-images.githubusercontent.com/2610172/47029849-81c0ba80-d13a-11e8-9e04-a356f84ca96a.png">
3. You are taken to this page: https://v1.gatsbyjs.org/docs/page-creation/#oncreatepage-api

### Expected result

Results for v2 site should not show up when searching on v1 site.

### Actual result

New page loads in weird state, loading icon never goes away and my macbook pro's fans turn on. Can't immediately tell what is going on by looking at network tab.

### Environment

Able to reproduce here:
https://v1.gatsbyjs.org/docs/

Firefox 62.0.3 (64-bit)
macOS High Sierra Version 10.13.6

siddhant1:
Should we redirect them to the v2 docs or should we block these searches?

Yurickh:
I would say that searching from within the v1 docs should only yield results from the v1 docs, and vice-versa.

Yurickh:
My guess here is that we'll need to create a new index in [docsearch](https://community.algolia.com/docsearch/) for the v1 docs, but I'm not sure on how to.

fk:
@pieh I guess I know as much about this as you (probably less ;-)), but what @Yurickh commented makes sense to me.

I think adding a new config file for v1.gatsbyjs.org with the config file from https://github.com/algolia/docsearch-configs/blob/b0c02ab108e53d1e8fe7d17ba26a6546b466104e/configs/gatsbyjs.json — which should have the correct setup/selectors for v1 should work.

👋@s-pace I saw you did the last updates to `[gatsbyjs.json](https://github.com/algolia/docsearch-configs/commits/master/configs/gatsbyjs.json)`, maybe you can help us? 🙏

s-pace:
👋 
We can continue to use the same index and add a `facetFilter` in order to reduce the scope of the search to either v1 or v2. Your analytics will be more relevant.

Could you add a meta tag within your website in order to highlight the version of your documenation? It should looks this way:

```html
<meta name="docsearch:version" content="1.0">
```

Then ping me and I will update your index

[...]

PR #9185, opened 2018-10-17
feat(www): add docsearch:version meta tag (v1)

pieh:
add meta tag so docsearch can index both v1 and v2 websites

see https://github.com/gatsbyjs/gatsby/issues/9159#issuecomment-430604230


pieh:
V1 is not actively maintained so probably not much value in spending time one it

Issue #8916, opened 2018-10-08
[gatsby-image] add placeholderClass prop to gatsby-image


baukevanderlaan:
## Summary

Inspired by #5745, I think a class prop on the placeholder image would be a useful addition. Would you welcome a PR for this feature?

### Basic example

```
<Img
  fluid={data.childImageSharp.fluid}
  placeholderClass="placeholder"
/>
```

### Motivation

Instead of directly setting styles on the placeholder image using `placeholderStyles`, this offers the possibility to set styles based on class using old school stylesheets.

kakadiadarpan:
@baukevanderlaan Sure! You will need to make change [here](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-image/src/index.js) and [README.md](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-image/README.md)

baukevanderlaan:
Thanks for the lead, I plan to do this next week. I suspect it's a minor thing to implement but I need to get to grips with the gatsby-dev-cli first.

kakadiadarpan:
@baukevanderlaan I forgot to provide the link for [Contribution guide](https://www.gatsbyjs.org/docs/how-to-contribute/) earlier.

Please reach out to us if you need any help. 🙂

baukevanderlaan:
Added in #9187

PR #9187, opened 2018-10-17
Feature: placeholderClass prop for gatsby-image

baukevanderlaan:
Adds the ability to provide a class prop to the placeholder element in gatsby-image as described in #8916.

### DIFF
--- a/packages/gatsby-image/src/index.js
+++ b/packages/gatsby-image/src/index.js
@@ -198,6 +198,7 @@ class Image extends React.Component {
       style = {},
       imgStyle = {},
       placeholderStyle = {},
+      placeholderClass,
       fluid,
       fixed,
       backgroundColor,
@@ -249,6 +250,7 @@ class Image extends React.Component {
                 title={title}
                 src={image.base64}
                 style={imagePlaceholderStyle}
+                className={placeholderClass}
               />
             )}
 
@@ -259,6 +261,7 @@ class Image extends React.Component {
                 title={title}
                 src={image.tracedSVG}
                 style={imagePlaceholderStyle}
+                className={placeholderClass}
               />
             )}
 
@@ -346,6 +349,7 @@ class Image extends React.Component {
               title={title}
               src={image.base64}
               style={imagePlaceholderStyle}
+              className={placeholderClass}
             />
           )}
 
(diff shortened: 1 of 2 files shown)


pieh:
Maybe `placeholderClassName` to be more "reacty"?

packages/gatsby-image/src/index.js
```diff
@@ -249,6 +250,7 @@ class Image extends React.Component {
                 title={title}
                 src={image.base64}
                 style={imagePlaceholderStyle}
+                className={placeholderClass}
```
pieh: Could we prep placeholder props early - because `alt`, `title`, `style` and now `className` seems to be repeated 4 times and we could prep 
```js
const placeholderImageProps = {
	title, 
	alt: !this.state.isVisible ? alt : ``},
	style: imagePlaceholderStyle,
	className: placeholderClass // or other prop name as I suggested
}
```

and then we could use something like
```js
<Img src={image.base64} {...placeholderImageProps} />
```

pieh:
It also would be great to add test coverage for that (and placeholderStyle too)

baukevanderlaan:
> Maybe `placeholderClassName` to be more "reacty"?

Agreed!

> Could we prep placeholder props early - because `alt`, `title`, `style` and now `className` seems to be repeated 4 times

Agreed!

> It also would be great to add test coverage for that (and placeholderStyle too)

I'm not yet familiar with testing, but I can give it a try. Does that belong within the scope of this PR or should it have its own PR?

baukevanderlaan:
Hi @pieh, I'm stuck adding test coverage; I can't find a way to query the placeholder image. Also, the placeholder image doesn't seem to be included in the snapshot. But like I said before, I don't have any experience with testing, so maybe someone can point me in the right direction?

pieh:
So it seems like input used to create `Image` component instance doesn't contain any values that could be used to create placeholder:

https://github.com/gatsbyjs/gatsby/blob/9463bd42a851861ebab326fc90abc507692053be/packages/gatsby-image/src/__tests__/index.js#L8-L22

don't have either `base64` or `tracedSVG` property. We should add one of those there - this should be enough, so snapshot actually contains placeholders.

baukevanderlaan:
@pieh I added tests for the placeholder image. Would you like to review?

PR #9197, opened 2018-10-18
feat(www): custom hubspot submission for newsletter

pieh:
so this is very hacky but seems to work:

povpu6vt3a

I was using my throwaway form to not mess with submissions + it didn't have "SFDC Campaign" - it should be fine (if docs at https://developers.hubspot.com/docs/methods/forms/submit_form_ajax are correct). 

And also disabled default html input verification for this gif - with type="email" + required attr, field validation should be handled by browsers (with hubspot server validation being fallback)

This doesn't make use of HubSpot usertoken - but it seems we currently don't use anyway, so there shouldn't be regression in this area (please check hubspot docs linked above for explanation of the token)

**Note** this PR is against @fk branch that re-adds newsletter form to homepage ( https://github.com/gatsbyjs/gatsby/pull/9183 )

/cc @fk @calcsam

### DIFF
--- a/www/src/components/email-capture-form.js
+++ b/www/src/components/email-capture-form.js
@@ -145,9 +145,8 @@ class Form extends React.Component {
           </Label>
           <SingleLineInput
             name="email"
-            type="text"
-            // type="email"
-            // required
+            type="email"
+            required
             autoComplete="email"
             ref={this.email}
           />
--- a/www/src/components/email-capture-form.js
+++ b/www/src/components/email-capture-form.js
@@ -1,9 +1,183 @@
 import React from "react"
 import { rhythm, options } from "../utils/typography"
-import { colors } from "../utils/presets"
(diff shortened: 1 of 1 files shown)


www/src/components/email-capture-form.js
```diff
@@ -1,9 +1,182 @@
...
+        pageUri: window.location.href,
+        pageName: document.title,
+      },
+    }
+
+    if (sfdcCampaignId) {
+      data.context.sfdcCampaignId = sfdcCampaignId
+    }
+
+    const xhr = new XMLHttpRequest()
```
DSchau: image

pieh: :D give me some fetch love and tell world IE is not a thing anymore to not throw in polyfills :)

fk:
🙏 !!!

PR #9120, opened 2018-10-15
add: docs for storybook

kkemple:
Added docs for using Storybook with Gatsby!

Issue #8821, opened 2018-10-05
gatsby `main` references es-module


travi:
I'm not sure that calling this a "bug" is necessarily correct, but it seemed like the best category to have the conversation around this. This probably comes down to project conventions, but I'm hoping we can also align to community conventions.

I've been dragging my feet getting upgraded to v2, but thanks to a helpful nudge from @DSchau we're making progress. However, after [upgrading the components](https://github.com/dsmjs/components/tree/gatsby-v2) (separate repo from the [site](https://github.com/dsmjs/site)), the previously working unit tests start erroring.

## Description

<details>
  <summary>stack trace</summary>

```
.../dsmjs/components/node_modules/gatsby/cache-dir/gatsby-browser-entry.js:1
(function (exports, require, module, __filename, __dirname) { import React from "react"
                                                                     ^^^^^

SyntaxError: Unexpected identifier
    at new Script (vm.js:73:7)
    at createScript (vm.js:245:10)
    at Object.runInThisContext (vm.js:297:10)
    at Module._compile (internal/modules/cjs/loader.js:657:28)
    at Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Object.require.extensions.(anonymous function) [as .js] (.../dsmjs/components/node_modules/babel-register/lib/node.js:152:7)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at Object.<anonymous> (.../dsmjs/components/node_modules/gatsby-link/index.js:25:15)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Object.require.extensions.(anonymous function) [as .js] (.../dsmjs/components/node_modules/babel-register/lib/node.js:152:7)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:20:18)
[...]

PR #6183, opened 2018-06-27
[wip] revert compling cache-dir and remove non standard code from cache-dir

pieh:
#5637 might be to aggressive to transpile it - right now `@babel/preset-env` doesn't know it needs to polyfill `Promise` used in `cache-dir`, which breaks runtime in ie 11 ( #6103 )

This attempt just removes flow types and move cache-dir to same place it was before.

closes #6103 
closes #6102

PR #9201, opened 2018-10-18
WIP: added suggestion for rollup config

travi:
for the gatsby package

this is a suggestion to build a common-js `main` bundle and an es-module `module` bundle using rollup. there are a few issues that will need to be worked out, but wanted to get some feedback about whether this is an approach that the team would be interested in first. 

for context, this is related to https://github.com/gatsbyjs/gatsby/issues/8821 and [a follow-up about jsx being present in the `module`](https://github.com/gatsbyjs/gatsby/pull/9120/files#r226161511)

### DIFF
--- a/.babel-preset.js
+++ b/.babel-preset.js
@@ -1,7 +1,7 @@
 const r = m => require.resolve(m)
 
 function preset(context, options = {}) {
-  const { browser = false, debug = false } = options
+  const { browser = false, debug = false, modules = 'commonjs' } = options
   const { NODE_ENV, BABEL_ENV } = process.env
 
   const PRODUCTION = (BABEL_ENV || NODE_ENV) === "production"
@@ -31,7 +31,7 @@ function preset(context, options = {}) {
             debug: !!debug,
             useBuiltIns: "entry",
             shippedProposals: true,
-            modules: "commonjs",
+            modules,
           },
           browser ? browserConfig : nodeConfig
         ),
--- a/packages/gatsby/package.json
+++ b/packages/gatsby/package.json
@@ -164,7 +164,7 @@
   ],
   "license": "MIT",
   "main": "cache-dir/commonjs/gatsby-browser-entry.js",
-  "module": "cache-dir/gatsby-browser-entry.js",
+  "module": "cache-dir/es/gatsby-browser-entry.js",
   "peerDependencies": {
     "react": "^16.4.2",
     "react-dom": "^16.4.2"
@@ -181,6 +181,7 @@
     "build:internal-plugins": "copyfiles -u 1 src/internal-plugins/**/package.json dist",
     "build:rawfiles": "copyfiles -u 1 src/internal-plugins/**/raw_* dist",
     "build:cjs": "babel cache-dir --out-dir cache-dir/commonjs --ignore **/__tests__",
+    "build:es": "BABEL_ENV=es babel cache-dir --out-dir cache-dir/es --ignore **/__tests__",
     "build:src": "babel src --out-dir dist --source-maps --ignore **/gatsby-cli.js,**/raw_*,**/__tests__",
     "build:rollup": "rollup -c",
     "clean-test-bundles": "find test/ -type f -name bundle.js* -exec rm -rf {} +",
--- a/packages/gatsby/package.json
+++ b/packages/gatsby/package.json
@@ -138,7 +138,10 @@
     "cross-env": "^5.1.4",
     "lerna": "^2.9.0",
     "nyc": "^7.0.0",
-    "rimraf": "^2.6.1"
+    "rimraf": "^2.6.1",
+    "rollup": "^0.66.6",
+    "rollup-plugin-auto-external": "^2.0.0",
+    "rollup-plugin-babel": "^4.0.3"
   },
   "engines": {
     "node": ">6.0.0"
@@ -179,6 +182,7 @@
     "build:rawfiles": "copyfiles -u 1 src/internal-plugins/**/raw_* dist",
     "build:cjs": "babel cache-dir --out-dir cache-dir/commonjs --ignore **/__tests__",
     "build:src": "babel src --out-dir dist --source-maps --ignore **/gatsby-cli.js,**/raw_*,**/__tests__",
+    "build:rollup": "rollup -c",
     "clean-test-bundles": "find test/ -type f -name bundle.js* -exec rm -rf {} +",
     "prepare": "cross-env NODE_ENV=production npm run build",
     "test-coverage": "node_modules/.bin/nyc --reporter=lcov --reporter=text npm test",
--- a/packages/gatsby/rollup.config.js
+++ b/packages/gatsby/rollup.config.js
@@ -1,0 +1,18 @@
+import babel from 'rollup-plugin-babel';
+import autoExternal from 'rollup-plugin-auto-external';
+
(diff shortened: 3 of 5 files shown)


packages/gatsby/package.json
```diff
@@ -179,6 +182,7 @@
     "build:rawfiles": "copyfiles -u 1 src/internal-plugins/**/raw_* dist",
     "build:cjs": "babel cache-dir --out-dir cache-dir/commonjs --ignore **/__tests__",
     "build:src": "babel src --out-dir dist --source-maps --ignore **/gatsby-cli.js,**/raw_*,**/__tests__",
+    "build:rollup": "rollup -c",
```
travi: if this is integrated, the `build:cjs` script above could be removed

packages/gatsby/rollup.config.js
```diff
@@ -0,0 +1,18 @@
...
+  plugins: [
+    autoExternal(),
+    babel({
+      presets: [[require(`path`).join(__dirname, `..`, `..`, `.babel-preset.js`), {modules: false}]],
+      runtimeHelpers: true,
+      exclude: ['./node_modules/**']
+    })
+  ],
+  output: [
+    {file: 'lib/index.cjs.js', format: 'cjs', sourcemap: true},
```
travi: the files produced here are the ones that should be referenced as `main` and `module` in the `package.json`. 

the directory and file names can certainly be adjusted to fit your conventions, but i started with these to demonstrate the mechanics to start with.

.babel-preset.js
```diff
@@ -1,7 +1,7 @@
 const r = m => require.resolve(m)
 
 function preset(context, options = {}) {
-  const { browser = false, debug = false } = options
+  const { browser = false, debug = false, modules = 'commonjs' } = options
```
travi: the output of the transpilation during the rollup build needs to preserve the es-module syntax so that rollup can properly do the rest of its processing. the default retains the previous behavior, but this allows overriding how the modules are handled.

packages/gatsby/package.json
```diff
@@ -138,7 +138,10 @@
     "cross-env": "^5.1.4",
     "lerna": "^2.9.0",
     "nyc": "^7.0.0",
-    "rimraf": "^2.6.1"
+    "rimraf": "^2.6.1",
+    "rollup": "^0.66.6",
```
DSchau: 😈

packages/gatsby/rollup.config.js
```diff
@@ -0,0 +1,18 @@
+import babel from 'rollup-plugin-babel';
+import autoExternal from 'rollup-plugin-auto-external';
+
+export default {
+  input: 'cache-dir/gatsby-browser-entry.js',
```
DSchau: This will fail, unfortunately. We _want_ the files to remain unbundled so that we can conditionally _require_ certain files.

For instance, if we take [`cache-dir/public-page-renderer`](https://github.com/gatsbyjs/gatsby/blob/7405585f5a98231c3391dda18d8425300f9dcde8/packages/gatsby/cache-dir/public-page-renderer.js) we can see the various conditional require statements.

Bundling this for the browser honestly just wouldn't work, I don't think, because it'll trace those dependencies, exclude conditional requires that we'd want in certain phases, and then create a (smaller, but not as functional) bundle file.

travi:
> This fails (for me?)
> > [!] Error: 'default' is not exported by cache-dir/public-page-renderer.js
https://rollupjs.org/guide/en#error-name-is-not-exported-by-module-
cache-dir/gatsby-browser-entry.js (10:7)

yeah, thats one of the issues i mentioned, but could have provided more detail about. that is a solvable issue, but didnt get too invasive with my changes before discussing details a bit more

DSchau:
@travi it's just using a [commonjs shim right](https://github.com/rollup/rollup-plugin-commonjs)? In my minimal experience with rollup, I think that's how I fixed it

packages/gatsby/rollup.config.js
travi: fair point. i'm not familiar enough with the approach, so this is some of the detail i was hoping to tease out. there could possibly still be a way to accomplish as a bundle (or multiple), but that could be more of an investment than its worth at this point. 

maybe the better approach would be simple transpilation then, but with `modules` set to `false` for `@babel/env`, similar to what was done for the common js version.

packages/gatsby/package.json
travi: 🤘

travi:
> it's just using a commonjs shim right?

if i were to fix this, i think i'd actually change it to conditionally export the es-module way. however, to your comment above, that might not be appropriate in this context, since it sounds like that conditional needs to be evaluated at runtime rather than build time.

travi:
i'm happy to update this to just do straight transpilation without bundling, list is done for the common-js version for `main`. 

the update to the preset will still be needed, but we'll need to pass config to the babel cli to use the config with `modules` set to `false`. it's not clear to me what the best way to accomplish this would be, though, since the `.babelrc` is already configured to use the preset with `modules` set to `commonjs`.

i'm open to thoughts since my familiarity with the raw babel-cli is rusty after using babel through either rollup or webpack for a while now.

DSchau:
@travi what would you say are using the benefits of rollup here if we may not actually _want_ a bundle? I think using babel and transpiling JSX is a happy medium, most likely. Just want to weigh any possible performance/file-size increase of this approach.

(and we need to get some type of performance measurements in place so we can track this at the CI/PR level; I started on this here -> gatsbyjs/gatsby#9167)

travi:
sorry, i probably wasn't clear enough, but based on what you highlighted about the dynamic imports needing to be possible at runtime, i don't think rollup is a good fit here. i agree that using babel-cli directly to transpile to some common denominator is the best bet for now. 

i think using the same preset as is used for the common-js version is a pretty good starting point, but with `modules` set to `false` (meaning don't transpile the module syntax). you could potentially decide to flip a few more flags, but its probably not necessary.

the piece that i'm not confident about with my babel-cli rustiness is how to use the same preset with different options for two independent builds from the same directory context. after thinking about it a little more since i posted last night, i did at least remember that `BABEL_ENV` at least used to be available to enable a level of dynamic config in a `.babelrc`. if that still exists, we could probably make it work with that approach, but i'm open to alternatives if you know of a better approach.

pieh:
I've tried naive approach of using same babel config we use for browser packages like `gatsby-link` or `gatsby-image` before but had to revert that - https://github.com/gatsbyjs/gatsby/pull/6183 . I'm not sure what's good approach here, but if we could only transform jsx without transpiling it to es5 that would work

DSchau:
@travi sorry - just circling back to this after quite a while.

> i see the preset has been pulled out to a separate package, so if we agree to go this route, i'll create a new PR to enable that preset to take the modules option

Let's go this route. I think that's a meaningful change that improves the behavior of that preset (it's now in babel-preset-gatsby-package), because I think that is generally valuable and I know other packages could use that behavior.

In the interim--I think there are a few intractable issues with this PR (specifically that the way gatsby is transpiled now isn't super conducive to bundling with something like Rollup). If there are smaller changes we can provide that fix some of the functionality you're seeing with importing this package in tests/storybook/etc. let's go that route, but for now, I don't think we want to introduce a bundler step.

As such--going to close this out. Please let me know if I can help with anything else, and thanks for the PR!

PR #8343, opened 2018-09-20
feat(gatsby-plugin-manifest): add cache busting to icon url

github0013:
Browsers cache favicons, and it's hard to update them without a random parameter in the favicon link.

PR #9217, opened 2018-10-18
docs(gatsby-plugin-manifest): link to plugin-offline

CanRau:
I thought it would be a nice addition for easier discoverability

Off topic question related to `gatsby-plugin-manifest` wouldn't it be good to add some kind of cache busting to the generated icons?


pieh:
I wonder if this link will work on NPM and/or gatsbyjs.org

> Off topic question related to `gatsby-plugin-manifest` wouldn't it be good to add some kind of cache busting to the generated icons?

There is WIP on that https://github.com/gatsbyjs/gatsby/pull/8343 (unfortunate PR name - it won't actually be random)

CanRau:
You're right haven't thought about npm, so then better link to npm or absolute github link?

thanks for the link!

pieh:
I would say use absolute link to gatsby site - https://www.gatsbyjs.org/packages/gatsby-plugin-offline

CanRau:
great, done 👍

PR #9227, opened 2018-10-18
fix(package.json): update homepage to point at the package

travi:
i followed a greenkeeper link, looking for details in the changelog, and was taken to the project root rather than the specific package that was updated in my project. this change seems to align to how the other packages are configured and would result in following links related to the npm package landing at the expected details.

### DIFF
--- a/packages/gatsby/package.json
+++ b/packages/gatsby/package.json
@@ -149,7 +149,7 @@
     "graphql.js",
     "index.d.ts"
   ],
-  "homepage": "https://github.com/gatsbyjs/gatsby#readme",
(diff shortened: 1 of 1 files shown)


packages/gatsby/package.json
```diff
@@ -149,7 +149,7 @@
     "graphql.js",
     "index.d.ts"
   ],
-  "homepage": "https://github.com/gatsbyjs/gatsby#readme",
+  "homepage": "https://github.com/gatsbyjs/gatsby/tree/master/packages#readme",
```
pieh:
```suggestion
  "homepage": "https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby#readme",
```

PR #9230, opened 2018-10-18
fix(www): search not initializing after client-side routing

### LINKED ISSUE #8567: [www] Search stops working after first usage.
## Description
Search function on https://www.gatsbyjs.org/ stops working after your first search.

### Steps to reproduce

1. Visit https://www.gatsbyjs.org/
2. Search for something (in the docs?) and click it
3. try to search again using the searchbar in the header

### Expected result

New search term gets picked up, does it's thing and Algolia populates the results.

### Actual result

Only autosuggestions of previous searches appear (also, not lined up with the searchbox vertically)

### Environment
Windows 10 Home build 1803
Google Chrome Versie 69.0.3497.100 (Officiële build) (64-bits)

fk:
Fixes #8567

### DIFF
--- a/www/src/components/search-form.js
+++ b/www/src/components/search-form.js
@@ -273,6 +273,10 @@ class SearchForm extends Component {
     navigate(`${a.pathname}${a.hash}`)
   }
   init() {
+    if (this.algoliaInitialized) {
+      return
+    }
+
     window.addEventListener(
       `autocomplete:selected`,
       this.autocompleteSelected,
@@ -290,6 +294,7 @@ class SearchForm extends Component {
         keyboardShortcuts: [`s`],
       },
     })
+    this.algoliaInitialized = true
   }
   componentDidMount() {
     if (
@@ -307,7 +312,6 @@ class SearchForm extends Component {
     }
   }
   componentWillUnmount() {
-    loadedJs = false
     window.removeEventListener(
       `autocomplete:selected`,
       this.autocompleteSelected,
@@ -315,12 +319,15 @@ class SearchForm extends Component {
     )
   }
   loadAlgoliaJS() {
-    !loadedJs &&
+    if (!loadedJs) {
       loadJS().then(a => {
         loadedJs = true
         window.docsearch = a.default
         this.init()
       })
+    } else {
+      this.init()
+    }
   }
   render() {
     const { focussed } = this.state
--- a/www/src/components/search-form.js
+++ b/www/src/components/search-form.js
@@ -306,6 +306,14 @@ class SearchForm extends Component {
       document.head.appendChild(link)
     }
   }
+  componentWillUnmount() {
+    loadedJs = false
+    window.removeEventListener(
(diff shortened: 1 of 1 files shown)


fk:
Not sure if this is a good/proper solution. Still babystepping through lifecycles. 👶 👣

www/src/components/search-form.js
```diff
@@ -306,6 +306,14 @@ class SearchForm extends Component {
       document.head.appendChild(link)
     }
   }
+  componentWillUnmount() {
+    loadedJs = false
```
KyleAMathews: This doesn't need undone as `loadedJs` is local to the component so will get removed when the component is removed.

pieh:
https://github.com/gatsbyjs/gatsby/blob/870a64775c7e1e5b899ad88d5e191f67fbdfeac0/www/src/components/search-form.js#L309-L316

Shouldn't be something like

```js
loadAlgoliaJS() { 
  if (!loadedJs) {
    loadJS().then(a => { 
      loadedJs = true 
      window.docsearch = a.default 
      this.init() 
    }) 
  } else {
    this.init() 
  }
} 
```

and 

https://github.com/gatsbyjs/gatsby/blob/2067b6ec248ba446d63cd5d0b007b25b25d3596d/www/src/components/search-form.js#L275-L293

be like:

```js
init() {
  if (this.algoliaInitialized) {
    return
  }
 
  window.addEventListener( 
    `autocomplete:selected`, 
    this.autocompleteSelected, 
    true 
  ) 
  window.docsearch({ 
    apiKey: `71af1f9c4bd947f0252e17051df13f9c`, 
    indexName: `gatsbyjs`, 
    inputSelector: `#doc-search`, 
    debug: false, 
    autocompleteOptions: { 
      openOnFocus: true, 
      autoselect: true, 
      hint: false, 
      keyboardShortcuts: [`s`], 
    }, 
  }) 
  this.algoliaInitialized = true
} 
```

www/src/components/search-form.js
pieh: `loadedJs ` is in module scope and not component state/field so it will be preserved between remounts I think

fk: It doesn't work without setting loadedJs back to false. I strongly guess that's because it's declared outside of the component? https://github.com/gatsbyjs/gatsby/blob/2067b6ec248ba446d63cd5d0b007b25b25d3596d/www/src/components/search-form.js#L11

fk: Oh @pieh thank you for commenting, didn't refresh page. Yeah that sounds a little more informed than my guessplaining :P

KyleAMathews: oops, sorry for the sloppy review 😬

fk: 🤗

fk:
@pieh Updated, `CanIHaz` one more 👀? 🙏

fk:
Thank YOU! <3

PR #9231, opened 2018-10-19
feat(gatsby-source-contentful): add options validation and more detailed error messages

pieh:
Missing required options:
Screenshot 2019-04-19 at 02 34 44

Wrong types of option fields / empty string as value / unknown option:
Screenshot 2019-04-19 at 02 40 25

Depending on response error codes from Contentful API, some (hopefully) helpful troubleshooting + masking secrets to not accidently leak it (i.e. in public CIs)

Wrong `spaceId` (404 error code from API):
Screenshot 2019-04-19 at 02 42 31

Wrong `accessToken` (401 error code from API):
Screenshot 2019-04-19 at 02 43 23


packages/gatsby-source-contentful/src/plugin-options.js
```diff
@@ -0,0 +1,32 @@
+const defaultOptions = {
+  host: `cdn.contentful.com`,
+
+  // default is always master
+  environment: `master`,
+}
+
+const validateOptions = options => {
```
DSchau: Why not use joi or yup here? I've wanted to introduce something like that!

I use it here -> https://github.com/DSchau/gatsby-source-s3/blob/master/src/plugin-options.js

packages/gatsby-source-contentful/src/fetch.js
```diff
@@ -25,14 +28,25 @@ module.exports = async ({ spaceId, syncToken, ...options }) => {
...
+      if (e.response.status === 404) {
+        // host and space used to generate url
+        details = `Endpoint not found. Check if host and space settings are correct`
+      } else if (e.response.status === 401) {
+        // authorization error
+        details = `Authorization error. Check if accessToken and environment is correct`
+      }
+    }
+    const errorMessage = `Accessing your Contentful space failed.
+Try running setting GATSBY_CONTENTFUL_OFFLINE=true to see if we can serve from cache.
```
DSchau:
```suggestion
Try setting GATSBY_CONTENTFUL_OFFLINE=true to see if we can serve from cache.
```

packages/gatsby-source-contentful/src/plugin-options.js
```diff
@@ -0,0 +1,32 @@
...
+  ;[`spaceId`, `accessToken`].forEach(option => {
+    if (!options[option]) {
+      errors.push(`Missing ${option}`)
+    }
+  })
+
+  return errors
+}
+
+const formatOptionsSummary = options => {
```
DSchau: Test would be nice here, just so it clarifies what you're doing here. Looking like you're spacing things out nicely and formatting the options in a table, basically?

DSchau: Also this would be a kind of cool package (that probably already exists). Basically some way to show what was passed, defaults, etc. It'd probably look nice as a table or something too!

Khaledgarbaya:
This is really good, Thanks @pieh 👍

packages/gatsby-source-contentful/src/plugin-options.js
pieh: I didn't want to pull extra dependencies just for checking 2 required fields but `joi` will work here as we already use it with gatsby (plus it will catch typos in fieldnames)

packages/gatsby-source-contentful/src/plugin-options.js
pieh: Yeah, I'll look into using something like https://www.npmjs.com/package/columnify and maybe implement this in reporter so it can be reused in other plugins

packages/gatsby-source-contentful/src/gatsby-node.js
```diff
@@ -58,11 +59,19 @@ exports.sourceNodes = async (
     return
   }
 
+  options = { ...options, ...defaultOptions }
```
pieh: wooops I would break custom `host` and `environment` here and override it everytime with defaults

packages/gatsby-source-contentful/src/plugin-options.js
```diff
@@ -0,0 +1,40 @@
...
+}
+
+const optionsSchema = Joi.object().keys({
+  accessToken: Joi.string()
+    .required()
+    .empty(),
+  spaceId: Joi.string()
+    .required()
+    .empty(),
+  host: Joi.string()
```
Khaledgarbaya: I think host can be optional here. if nothing is passed to the SDK will assume it's `cdn.contentful.com`

packages/gatsby-source-contentful/src/plugin-options.js
```diff
@@ -0,0 +1,40 @@
...
+    .required()
+    .empty(),
+  environment: Joi.string()
```
Khaledgarbaya: Same for `environment` it can be optional

packages/gatsby-source-contentful/src/plugin-options.js
pieh: It is in practice optional, because I merge defaultOptions with passed options before validating - but I probably should shuffle code around so the definition of shape of options is more clear

pieh: so I will look into doing that

pieh:
Reviving this PR after long time of inactivity (check updated description to see how it currently looks like)

packages/gatsby-source-contentful/src/plugin-options.js
pieh: Addressed after refactoring

PR #9232, opened 2018-10-19
Fail hard only for .jsx?|tsx? if default export is missing

ChristopherBiscardi:
fixes https://github.com/gatsbyjs/gatsby/pull/9031#issuecomment-431209901


KyleAMathews:
Can't we just do a different check for mdx? https://github.com/gatsbyjs/gatsby/pull/9031#issuecomment-431211566

We do want to fail hard for normal react components.

ChristopherBiscardi:
added a restriction to make the check apply to "javascript-ish" file extensions, which in this case means `/jsx?|tsx?/`

closes https://github.com/ChristopherBiscardi/gatsby-mdx/issues/178

andrewnester:
Thanks @ChristopherBiscardi ! Appreciate your help :)

Issue #6958, opened 2018-08-02
[v2] ReferenceError: exports is not defined in gatsby-browser


lightstrike:
I hit a strange error in `gatsby-browser.js` when working on a v2 upgrade (running `2.0.0-beta.65`):

```
Uncaught ReferenceError: exports is not defined
    at eval (var/www/gatsby/gatsby-browser.js:14)
    at Module../gatsby-browser.js (commons.js:1027)
    at __webpack_require__ (commons.js:725)
    at fn (commons.js:102)
    at eval (var/www/gatsby/.cache/api-runner-browser-plugins.js:2)
    at Object../.cache/api-runner-browser-plugins.js (commons.js:804)
    at __webpack_require__ (commons.js:725)
    at fn (commons.js:102)
    at eval (var/www/gatsby/.cache/api-runner-browser.js:3)
    at Object../.cache/api-runner-browser.js (commons.js:815)
```

`exports` is set up normally (I cross-checked with the [www file](https://github.com/gatsbyjs/gatsby/blob/master/www/gatsby-browser.js)):
```js
/**
 * Implement Gatsby's Browser APIs in this file.
 *
 * See: https://www.gatsbyjs.org/docs/browser-apis/
 */
import Cookies from 'universal-cookie';

/*
 * Set `siteVersion` cookie that is used for release tracking and split testing.
 */
exports.onClientEntry = () => {
    const cookies = new Cookies()
    cookies.set('siteVersion', process.env.BRANCH, { maxAge: 604800, path: '/' })
}
```

I was able to refactor as follows and everything seems to work as intended.

```js
export function onClientEntry() {
    const cookies = new Cookies();
    cookies.set('siteVersion', process.env.BRANCH, { maxAge: 604800, path: '/' });
}
```

Any ideas what might cause this?

lightstrike:
@Chuloo Thanks for clarifying! :confetti_ball: Would be great to add that to the Browser API docs.

jiv-e:
This is also referenced here: https://github.com/gatsbyjs/gatsby/issues/4452#issuecomment-403456868

Issue #4452, opened 2018-03-08
[v2] gatsby-browser.js errors at runtime: `exports is not defined`


lettertwo:
## Description

When a `gatsby-browser.js` file is loaded at runtime, it can throw an error under the right build circumstances. As far as I can tell, the circumstances seem to be:

- [@babel/preset-env] is configured not to transpile modules (`modules: false`)
- [@babel/plugin-transform-runtime] is being used
- `gatsby-browser.js` contains some code that gets transformed by the runtime transform plugin

### Steps to reproduce

1. Add a `gatsby-browser.js` file to the project with some contents
that could trigger the babel-runtime transform (see file contents below)
2. Run `gatsby build && gatsby serve`
3. Open the project in the browser to see the error in the console (see below)

### Expected result

should render 'hello world!'

### Actual result

Runtime error is thrown:
```
Uncaught ReferenceError: exports is not defined
    at Object../gatsby-browser.js (gatsby-browser.js:3)
    at __webpack_require__ (bootstrap 3868e9d1f8a80626a53e:54)
    at Object../.cache/api-runner-browser.js (api-runner-browser.js:2)
    at __webpack_require__ (bootstrap 3868e9d1f8a80626a53e:54)
    at Object../.cache/production-app.js (commons-1e491a978878eec435ee.js:777)
    at __webpack_require__ (bootstrap 3868e9d1f8a80626a53e:54)
    at webpackJsonpCallback (bootstrap 3868e9d1f8a80626a53e:25)
    at app-773b1833abbf62850cc1.js:1
```

### Environment

* Gatsby version (`npm list gatsby`): 2.0.0-alpha.10
* Node.js version: 8.9.3
* Operating System: OSX 10.13.3

### File contents (if changed):

`gatsby-browser.js`: 
```javascript
const React = require('react');
// Object spread seems to trigger babel runtime transform
exports.replaceComponentRenderer = ({props, loader}) =>
  React.createElement('h1', {...props, loader, children: 'hello world!'});
```

[@babel/preset-env]: https://babeljs.io/docs/plugins/preset-env/
[@babel/plugin-transform-runtime]: https://babeljs.io/docs/plugins/transform-runtime/

lettertwo:
Further findings:

- Using es6 import (`import React from 'react'`) will also cause the runtime error 
- Using es6 export (`export const replaceComponentRender = ...`) causes Gatsby not to be able to find the API hook.

[...]

PR #9239, opened 2018-10-19
Change exports to export const in api-browser-docs

jiv-e:
If you import something in gatsby-browser.js and use 'exports' syntax it will cause the following error:
https://github.com/gatsbyjs/gatsby/issues/4452
https://github.com/gatsbyjs/gatsby/issues/6958

It seems that just replacing the syntax with 'export const' fixes this. I don't know about the details, but at the moment the docs are broken. Just wanted to make this pull request because this fixed my issues. I haven't tested this very well. I have just done this in develop mode.

I hope you can take this forward.

### DIFF
--- a/packages/gatsby/src/utils/api-browser-docs.js
+++ b/packages/gatsby/src/utils/api-browser-docs.js
@@ -1,7 +1,7 @@
 /**
  * Called when the Gatsby browser runtime first starts.
  * @example
- * exports.onClientEntry = () => {
+ * export const onClientEntry = () => {
  *   console.log("We've started!")
  *   callAnalyticsAPI()
  * }
@@ -11,7 +11,7 @@ exports.onClientEntry = true
 /**
  * Called when the initial (but not subsequent) render of Gatsby App is done on the client.
  * @example
- * exports.onInitialClientRender = () => {
+ * export const onInitialClientRender = () => {
  *   console.log("ReactDOM.render has executed")
  * }
  */
@@ -23,7 +23,7 @@ exports.onInitialClientRender = true
  * @param {object} $0.location A location object
  * @param {object} $0.action The "action" that caused the route change
  * @example
- * exports.onPreRouteUpdate = ({ location }) => {
+ * exports const onPreRouteUpdate = ({ location }) => {
  *   console.log("Gatsby started to change location", location.pathname)
  * }
  */
@@ -35,7 +35,7 @@ exports.onPreRouteUpdate = true
  * @param {object} $0.location A location object
  * @param {object} $0.action The "action" that caused the route change
  * @example
- * exports.onRouteUpdateDelayed = () => {
+ * export const onRouteUpdateDelayed = () => {
  *   console.log("We can show loading indicator now")
  * }
  */
@@ -47,7 +47,7 @@ exports.onRouteUpdateDelayed = true
  * @param {object} $0.location A location object
  * @param {object} $0.action The "action" that caused the route change
  * @example
- * exports.onRouteUpdate = ({ location }) => {
+ * export const onRouteUpdate = ({ location }) => {
  *   console.log('new pathname', location.pathname)
  *
  *   // Track pageview with google analytics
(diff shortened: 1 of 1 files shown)


packages/gatsby/src/utils/api-browser-docs.js
```diff
@@ -23,7 +23,7 @@ exports.onInitialClientRender = true
  * @param {object} $0.location A location object
  * @param {object} $0.action The "action" that caused the route change
  * @example
- * exports.onPreRouteUpdate = ({ location }) => {
+ * exports const onPreRouteUpdate = ({ location }) => {
```
yeskunall:
```suggestion
 * export const onPreRouteUpdate = ({ location }) => {
```

packages/gatsby/src/utils/api-browser-docs.js
```diff
@@ -77,7 +77,7 @@ exports.onRouteUpdate = true
  * scroll to, `false` to not update the scroll position, or `true` for the
  * default behavior.
  * @example
- * exports.shouldUpdateScroll = ({
+ * exports const shouldUpdateScroll = ({
```
yeskunall:
```suggestion
 * export const shouldUpdateScroll = ({
```

stefanprobst:
I don't think the docs are wrong here: what is not supported is *mixing* ES and CommonJS module systems, i.e. if you use `import` you have to use `export`, if you use `require` you have to use `exports`.

jiv-e:
Thanks, @yeskunall! :)

@stefanprobst Is there some reason to use CommonJS in gatsby-browser.js? It would be much simpler to just default to one option in the documentation. Maybe give some separate example to use CommonJS and explain the benefits of doing it.

stefanprobst:
> Is there some reason to use CommonJS in gatsby-browser.js? It would be much simpler to just default to one option in the documentation.

It definitely would, AFAIU the issue is that ES modules support in NodeJS (which powers the API runner), is still experimental. For now I think it's best to use ES modules syntax in your React components (webpack handles those), and CommonJS in anything Node-related.

jiv-e:
@stefanprobst I guess that gatsby-browser.js is only run on the client. Is this true?

stefanprobst:
Yes the browser APIs are run in a client context. But they are [assembled in `api-runner-browser-plugins` during the bootstrap phase](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/cache-dir/api-runner-browser-plugins.js), you can check what it looks like in your project in the `.cache` folder.

jiv-e:
@stefanprobst So to reframe the question. The assembling works also with ES syntax. Is there some reason to use CommonJS in this context, which is not Node-related?

jiv-e:
@stefanprobst So to reframe the question. The assembling works also with ES syntax. Is there some reason to use CommonJS in this context, which is not Node-related?

jiv-e:
@stefanprobst To reframe the question. The assembling works also with ES syntax. Is there some reason to use CommonJS in this context, which seems not to be Node-related?

LekoArts:
I agree with @stefanprobst here and would refrain from changing the docs (Hence closing the PR).

@pieh @DSchau What do you think?

LekoArts:
@pieh @DSchau BUMP

pieh:
Yeah, let's close it - original issues are handled by displaying message that clearly show where is the problem as well as link to more details:
```
error This plugin file is using both CommonJS and ES6 module systems together which we don't support.
You'll need to edit the file to use just one or the other.

plugin: /Users/misiek/test/mixed-es-and-commonjs/gatsby-browser.js

This didn't cause a problem in Gatsby v1 so you might want to review the migration doc for this:
https://gatsby.app/no-mixed-modules
```

jiv-e:
Thank you for giving your time!

I still think that it's not consistent or very beginner friendly to have both syntaxes in the examples. It can trip you up.

Because you can't use them both, why not just consistently use only one in the examples? I don't personally care which one.

But it's your call, so I'll shut up now. Thanks for a great work with Gatsby!! ❤️

tcjcodes:
+1 to above, as a beginner (and someone coming back to Gatsby after awhile) it took some digging and clicking on nested issues to get me here. Definitely converting it to the ES6 module syntax in the docs would've been much easier than having to remember the quirks of a framework /:

Also worth mentioning that I didn't get that descriptive error mentioned above, I only got a failure on render (vs in the terminal) with this:

```
Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object.

Check the render method of `PageRenderer`.
```

PR #9194, opened 2018-10-17
Add detail and clarity around Environment Vars

Brondahl:
The existing documentation was all correct, but ambiguous and confusing in places, with a weird mix of assumed and non-assumed knowledge. It was also very unclear how Additional Environments were working.

PR #9240, opened 2018-10-19
Copy EnvVar docs improvements back to v1 Docs

Brondahl:
See PR #9194 .
v2 Docs were still unchanged from v1, so all the same information and statements still hold, thus we should apply the Doc improvements to v1 docs too.

### DIFF
--- a/docs/docs/environment-variables.md
+++ b/docs/docs/environment-variables.md
@@ -2,34 +2,76 @@
 title: "Environment Variables"
 ---
 
-You can easily provide environment variables to your site.
+## Environments and Environment Variables
 
-For JavaScript loaded into the browser, just add a `.env.development` and/or
-`.env.production` file in your root folder for development or production builds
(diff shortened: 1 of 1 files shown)


Brondahl:
Re: CircleCI error - I assume that CircleCI just doesn't work on the v1 branch atm? Surely a doc change can't have broken that, right?

Brondahl:
@m-allanson Could you do the honours?

Brondahl:
@m-allanson Could you do the honours?

Brondahl:
@m-allanson Could you do the honours?

Brondahl:
@m-allanson Could you do the honours?

pieh:
> Re: CircleCI error - I assume that CircleCI just doesn't work on the v1 branch atm? Surely a doc change can't have broken that, right?

Yeah, we no longer actively develop v1 and switched to circleCI recently and just didn't update config in v1 branch.

PR #9245, opened 2018-10-19
Update gatsby-plugin-netlify to allow caching JS files via HTTP headers

### LINKED ISSUE #9151: Documentation about caching javascripts assets needs an update
## Description
Documentation about caching javascript is outdated:

https://www.gatsbyjs.org/docs/caching/#javascript

> Other files e.g. JavaScript files are not (yet) cachable. Gatsby v1 is using webpack 1 which doesn’t make it possible to produce paths tied directly to the content of the file. Gatsby v2 will ship with webpack 4 which will make possible long-term caching of our JavaScript files.

> The cache-control header should be cache-control: public, max-age=0, must-revalidate

### Steps to reproduce

mmmh, visit the documentation link https://www.gatsbyjs.org/docs/caching/#javascript

### Expected result

have the documentation about caching javascript assets for gastsby v2

### Actual result

The documentation describe what to do for gatsby v1

### Environment

/

Thanks !

tanvibhakta:
This PR closes #9151 

Adds a line to apply recommended caching behavior on all /src/ files in a gatsby project served on netlify, increasing cached content and therefore performance. This also makes the plugin consistent with the documentation.

### DIFF
--- a/packages/gatsby-plugin-netlify/src/constants.js
+++ b/packages/gatsby-plugin-netlify/src/constants.js
@@ -27,6 +27,7 @@ export const SECURITY_HEADERS = {
 export const CACHING_HEADERS = {
   "/static/*": [`Cache-Control: public, max-age=31536000, immutable`],
   "/*\.js": [`Cache-Control: public, max-age=31536000, immutable`],
+  "/workbox*/*\.js": [`Cache-Control: public, max-age=31536000, immutable`],
 }
 
 export const LINK_REGEX = /^(Link: <\/)(.+)(>;.+)/
--- a/packages/gatsby-plugin-netlify/src/constants.js
+++ b/packages/gatsby-plugin-netlify/src/constants.js
@@ -26,6 +26,7 @@ export const SECURITY_HEADERS = {
 
 export const CACHING_HEADERS = {
   "/static/*": [`Cache-Control: public, max-age=31536000, immutable`],
+  "/src/*": [`Cache-Control: public, max-age=0, no-cache`],
 }
 
 export const LINK_REGEX = /^(Link: <\/)(.+)(>;.+)/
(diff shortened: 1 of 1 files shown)


pieh:
this is bit confusing - we will not have `src` directory in built site, can You elaborate a bit on how this works? I thought it should use paths inside `public` directory that get served by netlify hosting

tanvibhakta:
You are right, and I completely overlooked that. Will fix and re-push!

pieh:
There is edge case that will be problematic with just caching all `.js` files - there might be scenario where there are `.js` files that are not generated by gatsby webpack in output directory and we don't want to cache them permanently.

We will need to read webpack manifest, extract `.js` (and I think `.css` files too) and we will want to cache those.

> Unsure why the 'no-useless-escape' is being activated, without the escape charecter non-js files with 'js' in the name are also cached.

`eslint` isn't aware that this will be used as Regular Expressions, so we would need to opt out of that check in those lines. But we will also probably will need to drop that regex and target specific files because of my top part of my comment.

packages/gatsby-plugin-netlify/src/constants.js
```diff
@@ -26,6 +26,8 @@ export const SECURITY_HEADERS = {
 
 export const CACHING_HEADERS = {
   "/static/*": [`Cache-Control: public, max-age=31536000, immutable`],
+  "/*\.js": [`Cache-Control: public, max-age=31536000, immutable`],
```
DaleJefferson: Why are `.js` files not in the static folder?

chadfawcett:
@pieh Would the list of assets found at `assetsByChunkName` in the `public/webpack.stats.json` file cover all the `js` and `css` assets that should be cached indefinitely?

wardpeet:
Would love to get this one in! @tanvibhakta if you feel like processing this feedback that would be great. It's probably better to open up a new branch from master and try @pieh's feedback.

PR #8763, opened 2018-10-03
[www/starters] filter out invalid entries and set max number to show

amberleyromo:
Closes #8757

PR #9250, opened 2018-10-19
[starter library] WIP Filter for v2 starters by default

### LINKED ISSUE #7900: starter library - set filter default to Gatsby v2
@LeKoArts in issue #7785 

> I commented somewhere else that I'd propose to use a filter for v1 and v2 on the sidebar and set v2 by default. This way people could upgrade by themselves and the creators have some time to upgrade. After some time the v1 starters would be removed + the filter option.

chmac:
## Untested

I can't get the Gatsby site to run locally. The thumbnail generation takes forever and I haven't figured out the GitHub personal token thing yet.

## Changes

I believe that this will fix #7900 by setting the Gatsby v2 filter automatically when the starter showcase loads.

## Potential issues

* This might intersect weirdly with urls like `/?s=lumen` (a search query).

I'm hoping this PR will casue netlify to create a build preview that I can test.

### DIFF
--- a/www/src/views/starter-library/index.js
+++ b/www/src/views/starter-library/index.js
@@ -45,6 +45,6 @@ class StarterLibraryPage extends Component {
   }
 }
 
-export default RRSM({ s: ``, c: [], d: [], v: [], sort: `recent` })(
+export default RRSM({ s: ``, c: [], d: [], v: [`2`], sort: `recent` })(
   StarterLibraryPage
 )


chmac:
Hmm, I guess no netlify preview. Unfortunately somebody who has a working local install might have to test this. :-(

amberleyromo:
The thumbnail generation will take a long time on the first go. One (hacky) way to circumvent that currently is to shorten the list in `sites.yml` (and not commit it).

If you want to work on/test the starter library, yeah you'll need to set up the github token.

amberleyromo:
Otherwise, once the bug currently being addressed over in #8763 is resolved, I can take a look at running it myself and follow back up here!

chmac:
@amberleyromo Haha, cutting the starters list is a neat trick. I'll try that now! Thanks. :-)

amberleyromo:
@chmac i'd do `sites.yml` (cut the sites list, not the starters list) if you're working on the starter library! Since you don't need the site showcase to build, but you do want the full starter library.

pieh:
> I can't get the Gatsby site to run locally. The thumbnail generation takes forever and I haven't figured out the GitHub personal token thing yet.

https://github.com/gatsbyjs/gatsby/tree/master/www#screenshots-placeholder - we probably need better way of letting people know about this - not sure how yet

chmac:
I do now have `www` running locally, thanks for the help with that.

@amberleyromo The idea of setting `v2` as a filter in `filtered-starters` makes sense. Unfortunately, my experiments thus far have not worked. Here's what I think is happening:

* RRSM gets loaded
  - Reads state from URL
* FilteredStarters gets loaded
  - Is given the state from the URL as props by RRSM
 
At this point, if I push a change to the state in `<FilteredStarters` either in `constructor()`, `UNSAFE_componentWillMount()` or `componentDidMount()` then I believe this happens:

* The `setUrlState()` calls up to RRSM
* RRSM calls `navigate()` which updates the URL with the new filter values

However, because all of this happens in a *single* render cycle, the `<FilteredStarters` component does not get new props. The URL has changed, but the component still sees the original value.

If I wrap my call to `setUrlState()` inside a `setTimeout(fn, 0)` call, then the value does change. However, the whole page has rendered first, then gets re-rendered with the new filter value applied. It's a "content flash". :-(

Any suggestions on how to resolve this? Or maybe a different approach that would make more sense?

amberleyromo:
@chmac hmmm. I think it does need to happen in RRSM after all. Have you made any headway, or do you want me to dig into it further?

chmac:
@amberleyromo No further progress yet. From my perspective, I'm not sure if I follow the logic of hiding filters once they are no longer available in the results. If the `v1` and `v2` filters were always visible, then this might not be an issue. We could just set `v2` as the default value for the filter, and the `v1` checkbox would still be available as an option.

Otherwise, I guess RRSM has to become stateful? My go-to strategy for that would be [recompose](https://github.com/acdlite/recompose), but that probably doesn't make sense here if the library is not already in use. Maybe it's a job better suited for you than me?

LekoArts:
@amberleyromo @chmac Can this PR be closed since https://github.com/gatsbyjs/gatsby/pull/9739 has been merged?

PR #9252, opened 2018-10-19
Ignore all local env files, not just dev

chmac:
To test building locally, I added a  `.env.production` file with my GitHub token. It doesn't make sense to have that file in git. So I broadened the env file ignore from `.env.development` to `.env.*`.

### DIFF
--- a/www/.gitignore
+++ b/www/.gitignore
@@ -33,4 +33,4 @@ src/cache
 /package-lock.json
 
 # Env variables
-.env.development
+.env.*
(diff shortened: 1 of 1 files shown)


www/.gitignore
```diff
@@ -33,4 +33,4 @@ src/cache
 /package-lock.json
 
 # Env variables
-.env.development
+.env.*
```
DSchau: Maybe `.env*` so it'll catch the base case, too? e.g. `.env`

chmac: Great idea, pushing that change now.

PR #9253, opened 2018-10-19
Opensearch support

c0b41:


pieh:
I'm not familiar with opensearch standard - can you elaborate a little on what it does? 

We (potentially) would like to support this on algolia site-wide search and not just plugin search

c0b41:
@pieh more info https://www.algolia.com/doc/tutorials/search-ui/ux-patterns/search-from-the-url-bar/ 
also i will wait the site-wide search

pieh:
So we would need dedicated search page for that I guess, it doesn't seem like it would work well with current search text input in header

c0b41:
@pieh yeah, currently not work site-wide search

PR #9254, opened 2018-10-19
Updates caching docs since v2 has been released

macs91:


DSchau:
@macs91 thanks so much for the PR!

Sorry for the delay in getting this merged, but we've since had a PR come in that I think clarifies this concept a little more cleanly. #9347

As such, I'm going to close out this PR, but I hope you're able to comment and leave some feedback on the above PR _and_ / or we see another PR from you soon!

Thanks for contributing to Gatsby! 💜

PR #9255, opened 2018-10-19
fix: handle node api-api error gracefully

jquense:
I keep getting weird errors and it was because my gatsby-node.js had issues, but the build was continuing assuming there was a `result`


packages/gatsby/src/utils/api-runner-node.js
```diff
@@ -243,13 +242,8 @@ module.exports = async (api, args = {}, pluginSource) =>
       )
     })
       .catch(err => {
```
pieh: it seems like this catch should be for individidual plugin API call and not for `Promise.mapSeries`. It makes using `pluginName` a bit weird there (fact that it needs to be declared early) plus maybe it would allow for easier recovery in develop?

pieh: (and also returning `null` would actually make sense there)

jquense:
Thanks! Sorry for the poor formatting I meant to fix it and got pulled away :/

PR #9268, opened 2018-10-20
Changing stateful components to Stateless

KirankumarAmbati:


KirankumarAmbati:
**CC:** @KyleAMathews

LekoArts:
Why should a statless component be more performant? The reasonings behind your changes are needed from you.

KirankumarAmbati:
Hi @LekoArts,

Thanks for asking !

I changed the pull request title since my changes may not improve the performance drastically.

The reason for these changes are:

1. The code becomes more concise with Functional components.
2. It is preferred to go for Functional Component when we do not use component state or Life Cycle methods.
3. Already some of the components are coded as Functional Components when there is no state. These changes convert the above components to Functional as well to maintain consistent coding style.

Please let me know if I have to make any changes ! Thanks !!

pieh:
> 2\. It is preferred to go for Functional Component when we do not use component state or Life Cycle methods.

Why is it preferred? I don't mind either way, just curious.

mingaldrichgan:
Coincidentally, [React 16.6 was just released](https://reactjs.org/blog/2018/10/23/react-v-16-6.html), introducing [`React.memo`](https://reactjs.org/docs/react-api.html#reactmemo). Quoting from the [Motivation section of the RFC](https://github.com/reactjs/rfcs/blob/gaearon-patch-1/text/0000-memo.md#motivation):

> Today in React, function components and `PureComponent` provide two different kinds of optimizations that can't be unified.
>
> Function components let us avoid constructing a class instance. This benefits the initial render. Function components also tend to minify better than classes, which helps reduce the bundle size.

⋮

> By having `memo` as a first-class API in React itself, we can remove the need to make a choice between these optimizations. It makes it easy to memoize the output of function components without introducing other extra costs.

KirankumarAmbati:
Thanks for the review ! @DSchau and @fk 

I made changes thinking that the usage of Functional components is a good practice. That's what I have read in many articles.😛 :smiley_cat: But as you have mentioned, it might be a preference, and preference is more of a choice.


One thing I would like to bring to your notice is that,

In latest version of React ie. 16.6, React.memo() function has been introduced, which can be used with Functional components to make them optimized by avoiding unnecessary updates to the component.

May be in those lines, if my changes help Gatsby, please accept them. If not, please close the merge request. :grimacing: Thank you ! 🙂

KirankumarAmbati:
I just noticed your comment ! @mingaldrichgan 😃

Thank you for providing the information on React.memo() That might help us in deciding what is good for our Library. 🙂 ( It would also help me in learning things and best practices 😊 )

PR #9269, opened 2018-10-20
fix(gatsby-plugin-netlify-cms): automatic redirect

c0b41:
ref https://github.com/netlify/netlify-cms/issues/1456 , https://github.com/netlify/netlify-cms/issues/683

### DIFF
--- a/packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
+++ b/packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
@@ -146,3 +146,13 @@ exports.onCreateWebpackConfig = (
     webpack(config).run()
   }
 }
+
+exports.onPostBuild = ({ actions }, { publicPath = `admin` }) => {
+  const { createRedirect } = actions
+
+  createRedirect({
+    fromPath: `/${publicPath}`,
+    toPath: `/${publicPath}/`,
+    isPermanent: true,
+  })
+}


zslabs:
Would this potentially have anything to do with our `/admin/` route showing "too many redirects" now? On the latest version and just started seeing this. Not happening locally (in develop or prod) but am seeing this on Netlify prod.

zslabs:
Downgrading to `3.0.4` resolved the issue; so this PR may need a second look to see if it should be reverted. I created an issue that is referenced above.

erquhart:
I was hoping someone on the Gatsby side would confirm the bit about the redirect being exact before merging. This should be rolled back, agreed.

erquhart:
Commented further in #9584, still thinking this should be rolled back to avoid regressions.

borisirota:
In Netlify docs [trailing-slash](https://www.netlify.com/docs/redirects/#trailing-slash) section mentioned that:
```
# This rule will cause an infinite redirect
# because the paths are effectively the same
/blog/remove-my-slashes/   /blog/remove-my-slashes  301!
```
so this PR is the reason for "too many redirects" on Netlify prod.

pieh:
`gatsby-plugin-netlify-cms@3.0.6` was published with this PR reverted, thanks @zslabs for opening PR

PR #9278, opened 2018-10-21
Edit sites.yml. Edit Caddy smells like trees

podabed:

### DIFF
--- a/docs/sites.yml
+++ b/docs/sites.yml
@@ -533,16 +533,21 @@
     - Web Development
     - Technology
 - title: Caddy Smells Like Trees
-  main_url: "https://caddysmellsliketrees.ru/"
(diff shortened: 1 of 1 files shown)


docs/sites.yml
```diff
@@ -533,16 +533,21 @@
...
-  featured: false
   description: >-
     We play soul-searching songs for every day. They are merging in our forests
     in such a way that it is difficult to separate them from each other, and
     between them bellow bold deer poems.
   categories:
     - Music
+    - Band
+    - Gallery
+    - Lyrics
```
pieh: Band and Lyrics categories would create new ones with single entry in it - can we just add "Gallery" here to not create new single entry categories?

podabed: OK.

PR #9291, opened 2018-10-22
feat(www): better social sharing support

### LINKED ISSUE #9288: Incorrect picture when sharing Gatsby's site
## Description

This issue might be specific to Facebook Messenger, but might be a larger issues: if you share [Gatsby's website](https://gatsbyjs.org) in Facebook Messenger, you see what I believe is @calcsam's face! 😆

### Steps to reproduce

Just share [this](https://gatsbyjs.org) link in Facebook Messenger.

### Expected result

What should happen?

Should have seen Gatsby's logo, or something other than @calcsam (no offence, Sam! 😅)

### Actual result

image

### Environment

I'm running Chrome 69.0.3497.100, if that helps!

Seldszar:
Fixes #9288 

This pull request fixes image issues when sharing pages, like explained in the quoted one.

Another good example is when we use the Sharing Debugger on the features page, it shows the Wordpress logo (https://developers.facebook.com/tools/debug/sharing/?q=https%3A%2F%2Fwww.gatsbyjs.org%2Ffeatures%2F).


www/src/components/layout.js
```diff
@@ -151,6 +152,7 @@ class DefaultLayout extends React.Component {
           <meta name="twitter:site" content="@gatsbyjs" />
           <meta name="og:type" content="website" />
           <meta name="og:site_name" content="GatsbyJS" />
+          <meta name="og:image" content={gatsbyIcon} />
```
pieh: this will be relative path (something like `/static/some-hashed-name.png`) and I think og:image need absolute path right?

Seldszar: You're totally right!

www/src/components/layout.js
```diff
@@ -151,6 +152,7 @@ class DefaultLayout extends React.Component {
           <meta name="twitter:site" content="@gatsbyjs" />
           <meta name="og:type" content="website" />
           <meta name="og:site_name" content="GatsbyJS" />
+          <meta name="og:image" content={`https://gatsbyjs.org${gatsbyIcon}`} />
```
KyleAMathews: Please leave off the gatsbyjs.org so this will work on test deploys or locally (also fwiw, it's www.gatsbyjs.org).

pieh: It originally was relative path, but I think that path need to be absolute - didn't test it just searched for issues on the web and found few reports that relative paths don't work (granted it was from 2 years ago)

Seldszar:
As @mikelax suggested (https://github.com/gatsbyjs/gatsby/issues/9288#issuecomment-431915095), I'll include more metadata for better sharing support and edit the title in order to reflect the new direction.

www/src/components/site-metadata.js
```diff
@@ -0,0 +1,44 @@
...
+          }
+        }
+      }
+    `}
+    render={({
+      site: {
+        siteMetadata: { siteUrl, title },
+      },
+    }) => (
+      <Helmet defaultTitle={title} titleTemplate={`%s · ${title}`}>
```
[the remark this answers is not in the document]
Seldszar: Will do right away!

DSchau:
@mikelax that would be great!

PR #9292, opened 2018-10-22
fix(www): make DebounceInput value externally controlled

### LINKED ISSUE #9244: [starter-library] Empty search field on "Reset all filters"
## Summary

When clicking the "Reset all Filters" button the starter showcase, the "Search starters" input should be emptied.

### Basic example

* Go to the starter showcase https://www.gatsbyjs.org/starters/
* Enter `lumen` into the "Search filters" input (top right, left of "Submit a Starter")
* Click the "Reset all Filters" button (top left, grey)

Expected result:

* The "Search filters" input should be cleared

Actual result:

* The search query is no longer applied
* But the text remains in the box

### Motivation

Minor UI bug.

Wattenberger:
Fixes #9244 
Makes `DebounceInput`'s value externally controlled, which allows the parent component to change its value - eg. now we're clearing it on reset.

### DIFF
--- a/www/src/components/debounce-input.js
+++ b/www/src/components/debounce-input.js
@@ -22,7 +22,7 @@ class DebounceInput extends Component {
     this.setInputValue(this.props.value)
   }
 
-  componentDidUpdate(prevProps, prevState, shouldFireRouteUpdate) {
+  componentDidUpdate(prevProps) {
     if (prevProps.value != this.props.value)
       this.setInputValue(this.props.value)
   }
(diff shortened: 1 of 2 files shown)


www/src/components/debounce-input.js
```diff
@@ -5,35 +5,52 @@ import PropTypes from "prop-types"
...
+    inputValue: ``,
   }
 
-  onChangeText = e => {
-    this.setState({ value: e.target.value })
+  componentDidMount() {
+    this.setInputValue(this.props.value)
+  }
+
+  componentDidUpdate(prevProps, prevState, shouldFireRouteUpdate) {
```
pieh: `prevState` and `shouldFireRouteUpdate` are unused - can we remove them?

Wattenberger: sure thing!

PR #9302, opened 2018-10-22
update open graph meta tags attributes from name to property

greybax:
Fix for issue #9288 
Replaced "name=og:<some-name>" -> "property=og<some-name>."

### DIFF
--- a/www/src/templates/template-docs-markdown.js
+++ b/www/src/templates/template-docs-markdown.js
@@ -68,10 +68,12 @@ class DocsTemplate extends React.Component {
         <Helmet>
           <title>{page.frontmatter.title}</title>
           <meta name="description" content={page.excerpt} />
-          <meta name="og:description" content={page.excerpt} />
+          <meta property="og:description" content={page.excerpt} />
+          <meta property="og:title" content={page.frontmatter.title} />
+          <meta property="og:type" content="article" />
+          <meta property="og:image" content={`../../logo.svg`} />
+          <meta property="og:image:alt" content="Gatsby Logo" />
           <meta name="twitter:description" content={page.excerpt} />
-          <meta name="og:title" content={page.frontmatter.title} />
-          <meta name="og:type" content="article" />
           <meta name="twitter.label1" content="Reading time" />
           <meta name="twitter:data1" content={`${page.timeToRead} min read`} />
         </Helmet>
--- a/www/src/templates/template-docs-markdown.js
+++ b/www/src/templates/template-docs-markdown.js
@@ -71,8 +71,6 @@ class DocsTemplate extends React.Component {
           <meta property="og:description" content={page.excerpt} />
           <meta property="og:title" content={page.frontmatter.title} />
           <meta property="og:type" content="article" />
-          <meta property="og:image" content={`../../logo.svg`} />
-          <meta property="og:image:alt" content="Gatsby Logo" />
           <meta name="twitter:description" content={page.excerpt} />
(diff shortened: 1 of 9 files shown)


pieh:
Hmm, is `name` not working?

greybax:
If you are using `og` prefix they should be with `property` meta tag.
Ex:
`name="title"`
or
`property="og:title"`

www/src/templates/template-starter-page.js
```diff
@@ -60,9 +60,10 @@ class StarterTemplate extends React.Component {
             <Helmet>
               <title>{`${repoName}: Gatsby Starter`}</title>
               <meta
-                name="og:image"
-                content={screenshotFile.childImageSharp.fluid.src}
+                property="og:image"
+                content={`../logo.svg`}
```
DSchau: Why this change?

www/src/templates/template-docs-markdown.js
```diff
@@ -68,10 +68,12 @@ class DocsTemplate extends React.Component {
         <Helmet>
           <title>{page.frontmatter.title}</title>
           <meta name="description" content={page.excerpt} />
-          <meta name="og:description" content={page.excerpt} />
+          <meta property="og:description" content={page.excerpt} />
+          <meta property="og:title" content={page.frontmatter.title} />
+          <meta property="og:type" content="article" />
+          <meta property="og:image" content={`../../logo.svg`} />
```
DSchau: Same here.. ? I'm not 100% this will be picked up, either.

www/src/views/starter-library/index.js
```diff
@@ -27,15 +27,17 @@ class StarterLibraryPage extends Component {
...
             content={`Gatsby Starters: ${filtersApplied}`}
           />
           <meta
             name="twitter:description"
             content={`Gatsby Starters: ${filtersApplied}`}
           />
-          <meta name="og:title" content={filtersApplied} />
-          <meta name="og:type" content="article" />
+          <meta property="og:title" content={filtersApplied} />
+          <meta property="og:image" content={`../../logo.svg`} />
```
DSchau: Yeah, so I'm pretty sure this won't work. This isn't _require_'d so this is going to remain a relative path. This would presume that two levels above `https://www.gatsbyjs.org/starters` would contain a logo.svg, which isn't the case, correct?

www/src/templates/template-starter-page.js
fk: We have https://www.gatsbyjs.org/monogram.svg in https://github.com/gatsbyjs/gatsby/tree/master/www/static if that helps? Also could put a `PNG` or `JPG` version of the monogram or full logo (including the "Gatsby" wordmark) in there?

fk: …but I agree with Dustin, we should pass on the screenshot of the starter here?!

www/src/templates/template-docs-markdown.js
greybax: I supposed that this relative path will be resolved in accordance with project structure. Based on file where I've made this change..

www/src/templates/template-starter-page.js
greybax: reverted back to `screenshotFile.childImageSharp.fluid.src`

www/src/views/starter-library/index.js
greybax: Not sure which one we need to use here..

Issue #9249, opened 2018-10-19
[starter-library] Starter slugs are not unique


chmac:
## Description

The slug for each starter is the GitHub repo stub (the part after the username). There are currently 2 starters in the list called `gatsby-starter`. Those two starters have the same URL on the starter showcase.

### Steps to reproduce

* Filter starters by "Cosmic JS" https://www.gatsbyjs.org/starters/?c=Cosmic%20JS
* Click on the `cosmicjs/gatsby-starter` starter (the first in the list)

### Expected result

You should now see the cosmicjs starter.

### Actual result

You're looking at the  fabien0102 starter instead.

The problem lies here:

https://github.com/gatsbyjs/gatsby/blob/b5cb862f686dca3cf261ce72ede453ae800efc5c/www/gatsby-node.js#L447-L453

Using `repoStub` as the slug is not unique.

pieh:
Seems like adding `owner` to `slug` does make sense - but it will mess existing links - so we would need to create redirects as well

@amberleyromo do you have other ideas that we could use (busy day today for starter showcase :) )

jbranchaud:
I'd be happy to work on this.

My first thought would be to include the `owner` as part of the slug (as @pieh suggested). If, for any reason, you didn't want to have the `owner` name in the URL, another possibility is to use an md5 hash of the `owner` and `repoStub`.

amberleyromo:
@pieh @jbranchaud I agree that the most straightforward way would be to add owner to slug.

@jbranchaud thanks for offering to take this up!

jbranchaud:
I've submitted a PR to add `owner` to the slug. Does something also need to be done about redirecting what will now be broken URLs?

amberleyromo:
Closed in #9304 (+ subsequent fix  #10539)

PR #9304, opened 2018-10-23
Add owner to starter showcase URL slugs

jbranchaud:
Issue #9249


amberleyromo:
@pieh but this will create redirects for all new starters added moving forward, as well? we'd also need a way to resolve when there are two repos of the same name (the issue that prompted this improvement)

amberleyromo:
@pieh bump?

pieh:
I thought I replied earlier :/ Sorry about that.

> but this will create redirects for all new starters added moving forward, as well?

Didn't think about that and agree, we probably shouldn't. So We would need to create static list of already existing staters that need redirects.

> we'd also need a way to resolve when there are two repos of the same name (the issue that prompted this improvement)

We can't really do much about old/current URLs for those. I think redirects will need to redirect to same repo that gatsbyjs.org currently show. And the slug change will take care of future instances of that.

amberleyromo:
@jbranchaud see https://github.com/gatsbyjs/gatsby/pull/9304#issuecomment-433530667

chmac:
What about adding a generic redirect like `/starters/slug/` to `/starters/?s=slug`. I think the starter URLs will change from `/starters/slug/` to `/starters/user/slug/` so ideally we could safely redirect with a regex like `^/starters/[\w-]+/$` that would not affect the new starter URLs.

jbranchaud:
From what I can tell, the [`createRedirect`](https://www.gatsbyjs.org/docs/actions/#createRedirect) function doesn't support regex. I think we have to go the route of working from a static list of project slugs. Given that list, I think the suggestion of redirecting to `/starters?s=slug` is a great idea because it helps put the user in a context where they are more likely to see either of the starters they were looking for.

Any suggestions on how I can get a listing of starter projects? It's not clear to me where the set of starters is being seeded from.

chmac:
@jbranchaud The data is at [`docs/starters.yml`](
https://github.com/gatsbyjs/gatsby/blob/master/docs/starters.yml).

pieh:
Updated PR with static redirect map (to not create new redirects in the future)

the 2 starters that are currently problematic are:
- `gatsby-starter` (from `cosmicjs` and `fabien0102` users/orgs)
- `gatsby-starter-typescript` (from `goblindegook` and `haysclark` users/orgs)

pieh:
Thanks @jbranchaud for implementation! And sorry it took so long to actually merge :( - we lost track on this one.

PR #9306, opened 2018-10-23
Update gatsby-core-philosophy.md

jdemieville:
Core philosophy addition


shannonbux:
Hm, I think this is going to be a bigger project that will involve core team input! @jdemieville I'm going to close this and create a new issue where I mention you and the core team to make sure we write a doc that includes everyone's ideas!

Thanks for getting this kicked off!

PR #9307, opened 2018-10-23
Update unstructured data doc

amberleyromo:
Update unstructured data doc

Todo:
- link to blog post when it goes up


KyleAMathews:
Because she also edited the yaml file for the docs nav. Perhaps we should move those files to the docs folder so it doesn't trigger the check.

PR #9312, opened 2018-10-23
feat(www): Ecosystem page starting point & style lint setup [WIP]

greglobinski:
* upgrade `gatsby-transformer-screenshot` to use `GATSBY_SCREENSHOT_PLACEHOLDER=true`
* install and setup `stylelint`
* mute stylelint errors in non Ecosystem's components
* add ecosystem icon
* add ecosystem page
* refactor layout.js -> layout/layout-defaul.js
* refactor banneer.js -> layout/top-bannner.js
* create layout/page-heading.js

After the discussion on Slack I changed styles in styled component way. With that we get Contextual class names in development and linting for CSS code. But still if the decision is that `css` prop is the only way we do styling I can revert the changes. 

I moved the layout component and its sub components (however not all yet) into a subfolder `components/layout`. The new layout does not implement all features the current layout has, only these which are necessary for now [WIP], but I believe it's a good idea to create layout-default.js which can be use in most pages and also other layouts for special cases like for example for Showcase page with modal feature.

The refactored components (layout, top-bar) are for now applied only to Ecosystem page, I will apply them to the rest of the pages gradually together with other changes.  

screenshot from 2018-10-23 10-45-15
screenshot from 2018-10-23 10-45-37


greglobinski:
I'm closing it to revert styling to `css` prop.

www/gatsby-config.js
```diff
@@ -69,7 +69,9 @@ module.exports = {
         showSpinner: false,
       },
     },
-    `gatsby-plugin-emotion`,
+    {
+      resolve: `gatsby-plugin-emotion`,
+    },
```
jlengstorf: Not a problem, but I'm curious where this change came from.

www/src/assets/mobile-nav-icons.js
```diff
@@ -3,5 +3,13 @@ import DocsIcon from "!raw-loader!./docs.svg"
 import TutorialIcon from "!raw-loader!./tutorial.svg"
 import PluginsIcon from "!raw-loader!./plugins.svg"
 import ShowcaseIcon from "!raw-loader!./showcase.svg"
+import EcosystemIcon from "!raw-loader!./ecosystem.svg"
```
jlengstorf: Do we still need the `raw-loader` for SVGs? I've been importing them without it in my V2 sites and everything works as expected. (This may be out-of-scope for this PR, but if we can eliminate extra config like this, I'd like to put it on the roadmap.)

jlengstorf: In fact, this appears to have worked since V1 without the `raw-loader`: https://github.com/jlengstorf/lengstorf.com/blob/9ef7786073c1f586b91b967c67f9706102c77ee1/src/pages/about.js#L6

www/src/components/layout/content-container.js
```diff
@@ -0,0 +1,39 @@
...
+        ? rhythm(presets.maxWidthWithSidebar)
+        : rhythm(presets.maxWidth),
+      margin: `0 auto`,
+      padding: `${rhythm(1.5)} ${rhythm(options.blockMarginBottom)}`,
+      paddingBottom: rhythm(3.5),
+
+      [presets.Tablet]: {
+        paddingBottom: rhythm(1.5),
+      },
+    }}
```
jlengstorf: With the exception of `maxWidth`, it looks like all of this can move up into the initial style declarations.

www/src/components/layout/content-container.js
```diff
@@ -0,0 +1,39 @@
...
+import { css } from "react-emotion"
+
+import presets from "../../utils/presets"
+
+import { rhythm, options } from "../../utils/typography"
+
+const contentContainer = css`
+  background: red;
+  position: relative;
+`
```
jlengstorf: In this case, it's preferable to use `styled('div')` to clean up the component definition, e.g. we can scan for `<ContentContainer>{/*...*/</ContentContainer>` vs. `<div ...>{/*...*/}</div>`.

www/src/components/layout/layout-default.js
```diff
@@ -0,0 +1,98 @@
...
+  height: 1px;
+  width: 1px;
+  margin: -1px;
+  padding: 0;
+  overflow: hidden;
+  position: absolute;
+  z-index: 100;
+  font-size: 0.85rem;
+
+  &:focus {
```
jlengstorf: Not a problem, but `emotion` doesn't require the `&` unless you're trying to override specificity.

www/src/components/layout/page-heading.js
```diff
@@ -0,0 +1,88 @@
...
+  "& .svg-fill-gatsby": { fill: colors.lavender },
+  "& .svg-fill-brightest": { fill: `#fff` },
+  "& .svg-fill-accent": { fill: colors.lavender },
+  "& .svg-stroke-gatsby": { stroke: colors.lavender },
+  "& .svg-fill-gradient-accent-white-top": { fill: `transparent` },
+  "& .svg-fill-gradient-accent-white-45deg": { fill: `transparent` },
+  "& .svg-fill-gradient-accent-white-bottom": { fill: `#fff` },
+  "& .svg-fill-gradient-purple": { fill: colors.lavender },
+  "& .svg-stroke-gradient-purple": { stroke: colors.lavender },
+  "& .svg-fill-wisteria": { fill: `transparent` },
```
jlengstorf: Same here — no need for the `&` unless you have specificity problems.

www/src/components/layout/page-heading.js
```diff
@@ -0,0 +1,88 @@
...
+
+  [presets.Tablet]: {
+    left: 0,
+    position: "fixed",
+    padding: 0,
+    top: `calc(${presets.bannerHeight} + ${presets.headerHeight})`,
+  },
+})
+
+const H1 = styled.h1({
```
jlengstorf: We should standardize on a way of declaring these styled components.

My preference is very strongly toward writing real CSS unless there's a good reason not to do so. I'm happy to be convinced otherwise, but I'd like to see this shifted over to the tagged template literal declaration:

```js
const Component = styled('div')`
	margin-right: 10px;
`;
```

If possible, let's avoid the JS object style declarations altogether.

jlengstorf: Additionally, we should make a call between `styled.div` and `styled('div')`. I don't actually care which we use as long as it's consistent. My preference is for `styled('div')` if no one else feels strongly about this.

www/src/components/layout/top-banner.js
```diff
@@ -0,0 +1,71 @@
...
+
+import presets, { colors } from "../../utils/presets"
+import { rhythm, scale, options } from "../../utils/typography"
+
+const horizontalPadding = rhythm(1 / 2)
+const backgroundColor = props =>
+  props.background ? props.background : colors.gatsby
+
+const TopBannerContainer = styled("div")(props => ({
+  backgroundColor: backgroundColor(props),
```
jlengstorf: To keep these using plain CSS whenever possible, another way to write this would be:

```js
const TopBannerContainer = styled('div')`
	background-color: ${props => backgroundColor(props)};
	height: ${presets.bannerHeight};
	/* ... */
`;
```

This is definitely a nitpick, but my intention is to improve consistency — CSS is challenging as-is, so I'd like to avoid asking maintainers to remember when to use `margin-right` vs. `marginRight`, etc.

www/gatsby-config.js
greglobinski: I experimented with config, in the end removed it but do not revert to the original syntax

greglobinski: It occurred that the default settings is exactly what I need. I will revert it to the one-line sytax

www/src/assets/mobile-nav-icons.js
greglobinski: Yeah, I did not think only conformed to the existing code.

www/src/components/layout/content-container.js
greglobinski: Actually, the content-container.js is unused now, it's job does now the `Content` component in `layout-default.js` I will remove the  file altogether. Sorry for the mistake.

www/src/components/layout/content-container.js
greglobinski: As above, also in my next PR I will unify all Emotion components to one syntax, 
`styled("div")`
In the PR I use also `styled.div(` and that's inconsistency

www/src/components/layout/layout-default.js
greglobinski: OK

www/src/components/layout/page-heading.js
greglobinski: OK

www/src/components/layout/page-heading.js
greglobinski: 1. I agree, I also much more like the real CSS code with the tagged template literals, I used the objects, because nowadays styles are in objects, so after the discussion on Slack I decided to stay as close to the existing style as I can. But I gave my +1 for moving to strings.

2. https://github.com/gatsbyjs/gatsby/pull/9312#discussion_r227462113 Yes, I already did it (on my local)

www/src/components/layout/top-banner.js
greglobinski: +1

greglobinski: @jlengstorf , so now, I have to commit fixes to the PR, am I right?

fk: @greglobinski Yes, just add commits to the PR branch!

www/src/assets/mobile-nav-icons.js
greglobinski: @jlengstorf I've checked it out. It works on your site because you are using svg as  `src` of  `img`, we on www use it as an inline `svg`, and we still need the loader, without it we get `data:image...`

greglobinski:
@amberleyromo  It's a valid point. 

Could you please explain what do you mean by 'standalone project'? I have to update the layout, but sure I can add the new functionality to the existing one. I agree that, what I thought will be a plus point (we well know which pages use the new layout and which do not) could be the source of issues, as you pointed out.  

The topic is even more important than the layout case. I working now on the button component, we need the new look for the ecosystem page and at the same time we do not want to change the buttons on the working pages. So I can create a new-button or add conditional styling to the existing one. For now I did something like this
https://github.com/greglobinski/gatsby/blob/gl/ecosystem/content/www/src/components/shared/new-button.js
but for sure it's a temporary solution, the Button component needs a full refactoring to the new way of styling, with incorporating its styles into the component I think. Now the Button imports its own styles from utlis/styles. What's the motivation to keep the styles outside the component?

greglobinski:
@amberleyromo For sure, I'm open to any suggestions how to manage with the issue.

greglobinski:
@amberleyromo I withdraw everything about the button, now I see that @flo just introduced a new Rounded button and the Squared one will still be used. Am I right, @Flo? Do you plan to use the buttons in both styles Squared and Rounded with the new design?

amberleyromo:
https://github.com/gatsbyjs/gatsby/pull/9312#issuecomment-432693828

> Could you please explain for me what do you mean by 'standalone project'? I have to update the layout, but sure I can add the new functionality to the existing one. I agree that, what I thought will be a plus point (we well know which pages use the new layout and which do not) could be the source of issues, as you pointed out.

@greglobinski sure! What i mean is that creating a landing page is a "standalone" project, in the sense that it's a contained project -- it's a single page. Since this is the initial step in a larger project, what might make sense is:

- if it's required to make this a separate layout (as opposed to updating the existing layout in a non-breaking way), go ahead and make a specialized layout for this landing page.
- once more of this redesign work is done (even past this first phase with the ecosystem landing page) we can see what makes sense as a refactor to pull everything together.
- I do think what could make sense right now is pulling the existing `layout.js`, `banner.js` etc into a "layout" subdirectory within components.

I do not think it makes sense to make a brand new `layout` component the `default`. I think what we have should be appropriately refactored. If we do later decide a larger refactor / reorg effort needs to happen, that should be tackled on its own, imo. But it's out of the scope of creating a new landing page.

jlengstorf:
Jumping in to say that I agree with @amberleyromo — containing the scope to the smallest possible footprint is going to give us the best shot at a clean codebase. Ideally, this will look and feel so nice that we'll prioritize a site refactor to update the rest of the codebase as well. 😄

greglobinski:
@jlengstorf Should I do anything else besides committing fixes, what I've just done. Should I mark your review comments as resolved? I've already stared to do it, but it's probably not my role :). I'm not sure.

greglobinski:
@amberleyromo I've renamed the layout-default.js to layout-with-heading.js https://github.com/gatsbyjs/gatsby/pull/9312/commits/5b0cc9d6f7a2db62300872e3872d5b4f179eb849

greglobinski:
Thank you @jlengstorf 
So, who will merge the PR?

amberleyromo:
@greglobinski the recent updates still have a second version of a banner. Was this intended to be changed as well?

greglobinski:
@amberleyromo It's refactored for the new way of styling. I also moved the `{children}` of the banner to the component, the component is used only in layout, so I decided that in this way I will clear up the 'new' layout component. But sure, I can use the old one if you say so.

amberleyromo:
@greglobinski totally understand 👍 it's more about not having these duplicate components. A few further thoughts:

- I don't see a reason for this banner to live under a `/layouts` subdir in `/components`
- If we're going to essentially refactor the `banner.js` component, I'd rather see the existing component be refactored and integrated in a non-breaking way with the rest of the site. (Rather than having two separate banner instances in different parts of the site)

greglobinski:
@amberleyromo OK, I understand, I will use the original banner. Thank you

greglobinski:
@amberleyromo I refactored the original Banner in a non-breaking way to the new way of styling and to be used with default `children` content. Now there is only one Banner component for all pages.

PR #9313, opened 2018-10-23
[remark-autolink-headers] Add option to maintain case

siddharthkp:
The remark-autolink-headers plugin creates lowercase ids by default. Adding an option to maintain the case for markdown headers

Use case: You want the url to have the same case as the function name.

example: https://reactjs.org/docs/react-component.html# **componentDidMount**

- [x] Added optional configuration to the API
- [x] Added a test
- [x] Added documentation in README


pieh:
Thanks @siddharthkp!

Issue #6775, opened 2018-07-26
[www/showcase] Placement of the "Visit Site" button on modals


berraknil:
The current placement of the "visit site" button makes it harder to read (because it is on top of the image)

![image](https://user-images.githubusercontent.com/6593585/43264490-06ff63a4-90ef-11e8-8006-fbcf9d506617.png)

This might be a better place with the rest of the buttons/links for Source etc?

![image](https://user-images.githubusercontent.com/6593585/43264595-3c389bb2-90ef-11e8-81f8-50bb9c4dd028.png)

@fk what do you think?

fk:
👍 I agree—that part of the design wasn't very well planned!

A community member brought this up a while ago in #6098, too. There I followed up with a quick mockup of how a "realigned" header could look: https://github.com/gatsbyjs/gatsby/issues/6098#issuecomment-400325261

The scope of the aforementioned issue is the "Site Showcase", however we also use the same design to display the detail page of a "Starter" in the "Starter Showcase". The "Starter" detail header has slightly different requirements ("Deploy to Netlify" button, and potentially a couple more) and I still need to consolidate those.

fk:
Hey Berrak, forgot to update you on this: I posted a first stab at trying to consolidate header designs of site and starter showcase in https://github.com/gatsbyjs/gatsby/issues/6710#issuecomment-409296874. Would love to hear your thoughts if you have the time!

KyleAMathews:
@davad thanks! @fk do you think this is "good enough" now?

fk:
Late to the party, yep this is good to go … and already merged I see :/!

Thanks @davad, that brings us much closer to https://github.com/gatsbyjs/gatsby/issues/6710#issuecomment-417380873 and in line with what we do for the [starter showcase detail view](https://www.gatsbyjs.org/starters/gatsby-starter-blog)! 🙏 🤗 

Closed via  #9320. ✌️

PR #9320, opened 2018-10-23
Move the 'Visit site' and 'share' buttons above the showcase screenshot

davad:
This PR moves the the "Visit site" and "share" buttons so that they don't obscure the screenshot.

**Before**:
screen shot 2018-10-23 at 12 00 59 pm
screen shot 2018-10-23 at 12 01 17 pm

**After**:
screen shot 2018-10-23 at 11 59 04 am
screen shot 2018-10-23 at 11 59 32 am


fk:
Fixes #6775

PR #9328, opened 2018-10-23
Fixed a tiny grammar error

JoshyB:
Located in tutorial 4. Just changed 'is' to 'are'.

### DIFF
--- a/docs/tutorial/part-four/index.md
+++ b/docs/tutorial/part-four/index.md
@@ -28,7 +28,7 @@ sites.
 
 What is data?
 
-A very computer science-y answer would be: data is things like `"strings"`,
+A very computer science-y answer would be: data are things like `"strings"`,
 integers (`42`), objects (`{ pizza: true }`), etc.
 
 For the purpose of working in Gatsby, however, a more useful answer is


docs/tutorial/part-four/index.md
```diff
@@ -28,7 +28,7 @@ sites.
 
 What is data?
 
-A very computer science-y answer would be: data is things like `"strings"`,
+A very computer science-y answer would be: data are things like `"strings"`,
```
DSchau: I'm not actually _sure_ that this is right.

@amberleyromo @shannonbux wanna chime in here?

JoshyB: I believe "is" _is_ erroneous because in this context data is referring to a plural example. But the more I run it through my head I think maybe a more correct way would be to say "data types are". Not that it's a big deal or anything. :)

JoshyB: I apologize! I'm incorrect and I just learned something. Because "data" is a noncount noun, singular is correct!

DSchau: Nothing to apologize for!! It didn't _feel_ right to me, so thank you so much for checking!

I look forward to future PRs and thanks again for double checking!

Mind if I close?

DSchau: Or you got it 😉 Thanks!

shannonbux: Well this discussion inspired me to just read this whole article: https://blog.oxforddictionaries.com/2014/04/29/data-media-two-tricky-latin-plurals-handle/. Fascinating stuff. The Oxford English Dictionary is my favorite because they let usage inform what is considered "correct". So correctness is more what sounds right, rather than coming up with "rules".

Looks like data being a noncount noun (and singular) is the most common usage, although in some fields, like law, it's often treated as plural. In latin, it was the plural of datum (which hardly anyone uses now).

Ok now I'm going to stop nerding out lol.

PR #9329, opened 2018-10-23
feat(gatsby-source-wordpress): create site metadata node

### LINKED ISSUE #8051: [gatsby-source-wordpress] How to get blog title via GraphQL query when using gatsby-source-wordpress
I'd like to receive the blog title of my wordpress blog via a graphql query to use it in my gatsby page, but I can't figure how to do it. [A GET Request to /wp-json on the REST-API gives this information](https://wordpress.stackexchange.com/questions/223919/get-blog-title-with-rest-v2) pretty easily, but how do I query for it with GraphQL using the plugin?

openmedi:
This fixes #8051.


pieh:
We could follow up with updating docs on this - but functionality will be there at least

Issue #9204, opened 2018-10-18
Gatsby image sharp traced SVG does not respect forced aspect ratio


nakedible:
## Description

When using gatsby-image and gatsby-plugin-sharp to produce traced SVG images for image loading, and specifying both `maxWidth` and `maxHeight` for the resulting image, the traced SVG image is from the full (uncropped) source image, where as the actual generated image is cropped to aspect ratio with the crop location defined by `cropFocus`. The end result is very jarring where the traces do not match at all with the actual image when loaded.

Instead the traced SVG should match the cropped actual resulting image, meaning that the image should be cropped to correct aspect ratio first, and then traced.

### Steps to reproduce

Following GraphQL query:

```
childImageSharp {
              fluid(
                maxWidth: 600
                maxHeight: 600
              ) {
                ...GatsbyImageSharpFluid_tracedSVG
              }
}
```

And usage:

```
<Img fluid={childImageSharp.fluid} />
```

### Expected result

Traced SVG should match the image that is to be loaded.

### Actual result

Traced SVG is from uncropped image, meaning all lines are at different places.

### Environment

Cannot get gatsby info because of #8502.

nakedible:
Doing some additional testing, this seems to be broken in a few ways. The `healOptions` handler seems really weird, only converting some values from strings to integers, filling in hardcoded `width` even when both `maxHeight` and `maxWidth` are present. In addition to that, the fluid handler does a lot more processing for width and height and other things before passing on those things to resize (through the job queue). Even small changes in the image pipeline can change the `cropFocus` attention or entropy strategy result.

The only reasonable way that I can see this being fixed would be to first do the cropping and resizing to correct size in the job queue, and then trace the exact generated image without any resize or crop - which constitutes of somewhat of a rewrite of the current code.

kakadiadarpan:
@nakedible can you provide a reproduction repo for this? That would make it much easier to diagnose this issue.

As you are not able to provide the result of `gatsby info --clipboard`, can you share your `package.json` here?

fk:
I can confirm that traced SVGs do not match the cropped images. :(
Really sorry you're running into this @nakedible and thanks for opening up this issue! 🙏

fk:
Looked into this for a couple minutes, and want to clarify that this only is happening for `fluid` queries, not for `fixed` ones.

nakedible:
No worries, it's nice to get actual bugs found and fixed. I added a comment about `healOptions` to the pull.

PR #9337, opened 2018-10-24
fix(gatsby-plugin-sharp): tracedSVG for `fluid` images should respect forced aspect ratio

fk:
Ref. #9204, not sure if this is the right way to go…

- fixes memoization
- fixes `examples/using-gatsby-image` CSS that allows seeing the tracedSVG on hover on https://using-gatsby-image.gatsbyjs.org/traced-svg/
- split the [traced-svg](https://using-gatsby-image.gatsbyjs.org/traced-svg/) gallery into two—cropped and uncropped—to demonstrate the fix


examples/using-gatsby-image/src/pages/traced-svg.js
```diff
@@ -57,8 +57,9 @@ const UnsplashMasonry = edges => (
             "& .gatsby-image-wrapper:hover": {
               "& div + img": {
                 opacity: `1 !important`,
+                transition: `none !important`,
               },
-              "& img + img": {
+              "& img + picture > img": {
```
fk: Above brings back this (right image is hovered): 
image

nakedible:
Not necessarily to fix in this pull, but `healOptions` is kinda freaky:

- If the incoming options does not have `height` or `width`, then `width: 400` is added, even if the object has `maxHeight` and/or `maxWidth`.
- If the incoming options does not have `maxHeight` or `maxWidth`, then `maxWidth: 800` is added, even if the object has `height` and/or `width`.
- If the incoming options has `width` set as string, it is converted from string to number. Same for `height`. But if it has both, then *only* `width` is converted to number, but `height` remains as string.
- Same as above for `maxWidth` and `maxHeight`.

This pull sort of works around that by overwriting `width` and `height` from `maxWidth` and `maxHeight`, but still leaving in both. For example, if none of the width/height/maxWidth/maxHeight are specified, the resulting options object will have `width: 400` and `maxWidth: 800`.

pieh:
@nakedible agree, it would be great to split healOptions because right now it tries to "heal" sanitize input for few different functions (eg. `fluid` which is where maxWidth/maxHeight is used and `fixed` or `resize` where width/height) is used, but it's not in scope of this PR.

fk:
I can't exactly say why, but I'm hesitant to merge this…

fk:
Hey @nakedible, I wanted to let you know that @pieh and me took another look at this PR last week and it seemed while this approach is working on 80% of the cropped images, it has some slight offset issues with the other 20%. :-/

The good news is that @pieh worked his magic and did a PoC for a proper fix that will hopefully replace this PR soon.
Please have a little more patience. 🙏

wardpeet:
@pieh or @fk any progress? should we drop this Pr for now?

Issue #8400, opened 2018-09-21
Research running queries in parallel across cores


Moocar:
## Summary

Suggestion for speeding up query execution using multiple cores.

### The Problem(s)

The Query execution phase of a Gatsby build is often the slowest for a variety of reasons:

- Plugins can add fields to types that have custom resolvers (`gatsby-source-filesystem`)
  - Resolvers might perform slow side effects like http requests or fs operations (e.g source-filessystem `publicURL()` resolver)
  - plugin field values are lazy. So when supplying args to a graphql query, we must resolve all nodes of their type (more below)
- There is usually a query per page. So sites with lots of pages have many queries to execute
- sift query algorithm is naiive and iterates through entire node collection until result is found (TODO: Create issue to research using in-memory DB)

## Solution

Run queries across multiple cores. Queries do not depend on each other, so there is potential to run them in parallel. In a sense, we already do this by using `better-queues's` concurrency option. But this only achieves async concurrency. We're still locked to a single core because node.js is single threaded. But libraries like `jest-worker` make spinning up extra node processes easier.

The solution would involve creating a number of jest workers before we start consuming from the `query-queue.js`. Each worker would have its own copy of the redux `nodes` namespace and so in theory would have everything it needed to execute queries. Then, the implementation of `query-runner.js` would be handled by these jest workers who would be given batches of queries to execute. Very similar to how HTML generation works now.

### Cached plugin field resolver's values

Fields provided by plugins during `setFieldsOnGraphQLNodeType` must be resolved before the query can be run. Since this would be infeasibly slow to run for every query, aggressive caching is performed. Since we can't share memory, then each worker will have to keep its own cache. Another option would be to use a centralized cache in the main process, or in something like redis. The three caches used in `run-sift.js` are:

- `resolvedNodesCache` - All fully resolved nodes. key = (type, nodes.length, query filter args)
- `enhancedNodeCache` :: A single fully resolved (recursively finished) node. key = (node, query filter args)
- `enhancedNodesPromiseCache` :: A current unfinished promise that's still executing (since it's not in enhancedNodeCache yet). key = (node, query filter args)

### Increased Memory Footprint

For large sites, the `nodes` redux namespace is the largest. Each time we start a worker, we have to copy it to the new process. We could very quickly run out of memory. And it will only get worse the bigger the site is, which annoyingly is the exact use case we're trying optimize for. 

[...]

Issue #7251, opened 2018-08-11
How to query all posts by an author when you have multiple authors per post


gurtjun:
## Summary

I'm working on a blog where one blogpost can have multiple authors. I got this working with [mappings between my nodes](https://next.gatsbyjs.org/docs/gatsby-config/#mapping-node-types). I can query for a markdown file and get all the linked authors. 

Now I'm working on a page for each author that displays all the posts he/she contributed to. How can this be done? Below you can find the query I'm currently using, but that one's not working (`author.js`).

I setup an example in [this repo](https://github.com/gurtjun/gatsby-multiple-authors).
### File contents
`gatsby-config.js`: 
```
module.exports = {
  ...
  mapping: {
    "MarkdownRemark.frontmatter.authors": `AuthorYaml`,
  },
  ...
}
```
`gatsby-node.js`: 
```
const path = require(`path`)
const { createFilePath } = require(`gatsby-source-filesystem`)

exports.onCreateNode = ({ node, getNode, actions }) => {
  const { createNodeField } = actions
  if (node.internal.type === `AuthorYaml`) {
    const slug = createFilePath({ node, getNode, basePath: `data` })
    createNodeField({
      node,
      name: `slug`,
      value: slug,
    })
  }
}

exports.createPages = ({ graphql, actions }) => {
  const { createPage } = actions
  return new Promise((resolve, reject) => {
    graphql(`
        allAuthorYaml {
          edges {
            node {
              id
              fields {
                slug
              }
            }
          }
        }
      }
    `).then(result => {
      result.data.allAuthorYaml.edges.forEach(({ node }) => {
        createPage({
          path: node.fields.slug,
          component: path.resolve(`./src/templates/author.js`),
          context: {
            // Data passed to context is available
            // in page queries as GraphQL variables.
            slug: node.fields.slug,
            id: node.id
          },
        })
      })
      resolve()
    })
  })
}
```

`src/pages/blog/example-blog-post.md`
```
---
title: "Example Blog Post"
date: "2018-08-11"
authors: [john_doe, jane_doe]
---
```

`src/data/author.yaml`
```
- id: john_doe
  name: John Doe
- id: jane_doe
  name: Jane Doe
```

`src/templates/author.js`
```
...
export const query = graphql`
  query($id: String!) {
    allMarkdownRemark(filter: { frontmatter: { authors: { in: [$id] } } } ) {
      edges {
        node {
          ...
          html
          ...
        }
      }
    }
  }
`
```

[...]

PR #9338, opened 2018-10-24
[WIP]: feat(gatsby): use lokijs for in-memory database

Moocar:
**Draft. Don't merge!! Feedback requested. Questions down the bottom**

## Summary

One of the most expensive bootstrap phases in Gatsby for large sites is the GraphQL querying phase. There are two aspects to this:

1. Custom field resolvers (e.g sharp). Will hopefully be improved by running queries in parallel (#8400)
2. Queries are brute force and don't use indexing

This PR addresses 2, querying.

Gatsby stores all its nodes in a redux namespace (`nodes`). To query, it uses [sift](https://github.com/crcn/sift.js/tree/master). Sift doesn't have any concepts of indexing, so it performs a brute force search by iterating over all nodes until it finds a match (or matches). The same applies for sorting. This is very slow on large sites.

This PR speeds up query performance for **10k page sites by about 2x and 20k sites by about 5x**. See [recording.md](https://github.com/gatsbyjs/gatsby/blob/187adb95712ab8317824c731e35ecc2d3ca92c80/benchmarks/query/recording.md) for results.

## Loki

The main change is to replace the redux `nodes` namespace with [loki.js](http://lokijs.org/#/). And then at query time, use loki to query instead of sift. In redux, nodes are stored in a giant array, whereas with loki, I have divided them up into collections per type of node (`internal.type`). This is because almost everywhere in Gatsby, we want to query for the nodes of a type, and almost never by all nodes in the system. This means that plugins that currently filter through all nodes looking for those of a certain type can call `getNodesByType()` which as O(1) operation.

### Differences

Like sift, loki's query syntax matches mongo. There is one major difference though. Sift can query nested fields using datastructures, e.g `{ foo: {bar: {eq: "1"} } }`, whereas loki uses dotted string paths, e.g `{ "foo.bar": {eq: "1"} }`. 

### Nested indexes

Loki didn't support indexes on nested (dotted) fields. I've addressed this in https://github.com/techfort/LokiJS/pull/718. Hopefully it will be merged soon. If not, we can create a fork for now and publish

### Persistence

Loki is an in-memory DB but supports file serialization. It supports the same autosave functionality that we've built into [redux/index.js:saveState()](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/redux/index.js#L67) but is supposedly even faster. If we move  `pages`, `dataDependencies` etc into loki we should be able to remove most of this functionality.

Another added benefit is that **saves only occur if data is dirty**.

## Query Benchmark

Added a benchmark project to test query performance.

## Questions

- I've called the loki create/update/delete functions from inside the nodes reducer. Any thoughts on that? I can't think of anything "wrong" with it, but it feels icky. I'm tempted to instead bypass redux entirely and call the `db/index.js` directly from `actions.js` and then emit a `CREATE_NOTE` event.
- Assuming there is agreement on this PR, should we move `pages`, `componentDataDependencies` and other redux namespaces into loki instead? Benefits would include dirty flagged persistence, and potential querying speed up 
- I've included a flag `useQueryIndex` which can be set on the query context. This tells loki to create indexes for all fields in the query. This allows users to create indexes where applicable, which for most sites, will only be on node groups > 1000 pages or so. But the other option is that we do something more intelligent to auto set indexes. E.g flag groups of nodes that are > 1000, and if a query occurs against them, then index the query's fields.
- All nodes inserted into loki have a `$loki` field. And nodes are mutable, so that field will leak into the rest of Gatsby. It hasn't been an issue so far, but I'm curious if folks think it will be a problem. We could remove the field after query results are returned, but we'd have to clone the whole object which would be a performance penalty.

Note, the tests are failing. I'm working on fixing them up now but wanted to get the PR up now.


benchmarks/query/recording.md
```diff
@@ -0,0 +1,88 @@
+## Summary
+
+Key findings:
+
+- loki without indexes is overall slightly faster than master, except when there are many types
+- loki with indexes is about 2x faster on sites with 10k pages, and 5x faster with 20k pages. But is ever so slightly slower when those pages are split across 100 types.
+
+Overall, loki is a big win for sites with lots of pages of the same type. For smaller sites, the difference is negligible.
```
DSchau: A caching PR I recently worked on had similar findings, which I think is totally reasonable. If it's mostly the same for smaller sites but it makes Gatsby much more scalable I think that's a big win.

I honestly wonder if using lokijs _consistently_ (e.g. for cache and internal Gatsby utilities) would be worthwhile. I'd lean towards yes!

packages/gatsby-source-contentful/src/gatsby-node.js
```diff
@@ -216,17 +216,16 @@ exports.sourceNodes = async (
...
 
   const CACHE_DIR = path.resolve(
     `${program.directory}/.cache/contentful/assets/`
   )
   await fs.ensureDir(CACHE_DIR)
 
-  const nodes = getNodes()
-
-  if (!nodes.some(n => n.internal.type === `ContentfulAsset`)) {
+  const nodes = getNodesByType(`ContentfulAsset`)
```
DSchau: I'll look through the implementation of this as I go through, but would this be worth returning an empty array or something so we can resolve the empty case more easily?

packages/gatsby/src/bootstrap/index.js
```diff
@@ -211,6 +212,25 @@ module.exports = async (args: BootstrapArgs) => {
...
+  })
+  activity.start()
+  const dbSaveFile = `${program.directory}/.cache/loki/loki.db`
+  try {
+    await db.start({
+      saveFile: dbSaveFile,
+    })
+  } catch (e) {
+    report.error(
+      `Error starting DB. Perhaps try deleting ${path.dirname(dbSaveFile)}`
```
DSchau: 🤔 probably worth just doing it automatically, or at least prompting?

packages/gatsby/src/bootstrap/load-plugins/load.js
```diff
@@ -6,6 +6,7 @@ const crypto = require(`crypto`)
 const glob = require(`glob`)
 const { store } = require(`../../redux`)
 const existsSync = require(`fs-exists-cached`).sync
+const createNodeId = require(`../../utils/create-node-id`)
```
DSchau: @pieh we need a gatsby-utils _stat_ :) Something one of us can work on soon!

packages/gatsby/src/db/index.js
```diff
@@ -0,0 +1,276 @@
...
+function getDb() {
+  return db
+}
+
+/**
+ * Deletes all collections.
+ */
+function deleteAllCollections() {
+  if (db) {
+    _.forEach(db.listCollections(), collInfo => {
```
DSchau: Might be worth benchmarking _.forEach vs. native forEach (in general, not just here!). I tend to prefer to defer to native array methods, but I may be unique in that regard.

I think that it's a safer bet to presume they'll be optimized very efficiently, if they're not already!

packages/gatsby/src/db/index.js
```diff
@@ -0,0 +1,276 @@
...
+ * Updates a node in the DB
+ *
+ * @param {Object} node The new node information. Will be merged over
+ * the old one (shallow merge)
+ * @param {Object} oldNode The old node to merge the new node
+ * over. Optional. If not supplied, the old node is found by querying
+ * by node.id
+ */
+function updateNode(node, oldNode) {
+  invariant(node.internal, `node has no "internal" field`)
```
DSchau: I like the use of invariant here (and elsewhere!)

packages/gatsby/src/db/index.js
```diff
@@ -0,0 +1,276 @@
...
+
+  let coll = db.getCollection(type)
+  if (!coll) {
+    invariant(coll, `${type} collection doesn't exist. When trying to update`)
+  }
+
+  if (!oldNode) {
+    oldNode = getNode(node.id)
+  }
+  const updateNode = _.merge(oldNode, node)
```
DSchau: Is this doing anything an Object.assign wouldn't do? (again, sorry, I just don't always love lodash!)

packages/gatsby/src/db/index.js
```diff
@@ -0,0 +1,276 @@
...
+    return undefined
+  }
+}
+
+/**
+ * Returns all nodes of a type (where typeName == node.internal.type)
+ */
+function getNodesByType(typeName) {
+  const coll = db.getCollection(typeName)
+  if (!coll) return null
```
DSchau: Yeah, maybe just return empty array here? Makes it easier for a consumer of this "API" to handle the empty and null cases the same way.

packages/gatsby/src/schema/run-query.js
```diff
@@ -0,0 +1,56 @@
...
+const { pluginFieldTracking } = require(`./plugin-fields`)
+
+function hasPluginFields(queryArgs) {
+  return _.some(queryArgs.filter, (v, fieldName) =>
+    pluginFieldTracking.has(fieldName)
+  )
+}
+
+function chooseQueryEngine(queryArgs) {
+  if (hasPluginFields(queryArgs)) {
```
pieh: This will definitely need to be more robust - our filter schema support is not really that great but even now we can filter by linked node with ___NODE convention (which need to use custom resolver to access underlying node)

benchmarks/query/recording.md
DSchau: The fs fallback is interesting, since I used a similar approach for the cache! Basically I kept 100-200 items in memory, and then it fell back to the filesystem as its cache, which seems to work pretty well for large sites.

packages/gatsby-source-contentful/src/gatsby-node.js
pieh: Yeah it would save us all of those checks below

benchmarks/query/recording.md
KyleAMathews: @Moocar did you measure how much memory is used in these different benchmarks?

@DSchau same question?

DSchau: @KyleAMathews not explicitly (although I can), but I used the upwards bounds of when it ran _out_ of memory as the basis for which I considered that. It is dependent on machine setup, size of items being cached, etc. too.

packages/gatsby/src/db/index.js
pieh: should we actually merge oldNode with new one? This might cause weird effects, like deleting some fields from node will actually not be deleted in db?

benchmarks/query/recording.md
Moocar: @DSchau Thanks for all the feedback! For caching, using loki would be a step up from using pure arrays, but an even bigger win would probably be to use a real caching library that can also provide persistence. Once that has LRU, paging to disk and other cache-specific semantics, assuming one exists! (I haven't looked)

Moocar: @KyleAMathews Memory usage should be less since we're actually able to remove some caches. But I'll double check. What's your favorite memory profiling tool?

packages/gatsby/src/bootstrap/index.js
Moocar: My thoughts too! Glad you agree. I'll implement it

packages/gatsby/src/db/index.js
Moocar: Cool, I'm actually a relative newcomer to js so any tips like this are greatly appreciated. I may have gone a bit lodash mad in this PR as I come from a functional background. Is this the native for each you mean? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

packages/gatsby/src/db/index.js
Moocar: Yeah, my approach with invariants like this is to be quite liberal in their application, and then perhaps remove them in the future if this code becomes very stable.

packages/gatsby-source-contentful/src/gatsby-node.js
Moocar: Yep! Good call. I'll return the empty collection

packages/gatsby/src/db/index.js
Moocar: addressed in other comment

packages/gatsby/src/schema/run-query.js
Moocar: Great catch. I didn't notice the linked field example. Can you think of any other types of fields that might need to be resolved too? How about mapping fields?

When @KyleAMathews  and I were discussing this PR, we decided that the vast majority of sites would not be using pluginFields. I assume they also aren't using linked fields. But I could be wrong in that assumption

packages/gatsby/src/db/index.js
Moocar: @pieh Great catch again. I just looked at the redux implementation and all it does is overwrite the existing node in redux. Not sure why I did a merge here. I'll change to overwrite.

Moocar:
@pieh I love the idea of feature flagging this. I'm kicking myself for not writing all this code with that in mind. I'll have a think about how hard it would be to do that

packages/gatsby/src/schema/run-query.js
pieh: We currently don't support filtering on mapped fields (that's what I meant that support isn't great), but it's something that we would definitely would want to be able to do (here's one of the issues where people mention they can't really filter by mapped fields - https://github.com/gatsbyjs/gatsby/issues/7251, I have WIP branch for this - https://github.com/pieh/gatsby/commit/ccd110c0074ec6f7e802447a766b079146432a8a).

For now only ___NODE fields would be problem. Worth looking into date or local File fields which do have custom resolvers for output, but we don't create correct filter schema for those right now. 

When we create schema type fields with custom resolvers we should track those fields (similar to how you track fields from plugins)

packages/gatsby/src/schema/build-node-types.js
```diff
@@ -160,6 +175,17 @@ module.exports = async ({ parentSpan }) => {
...
 
+    const inferredFieldNames = _.keys(
+      inferredInputFieldsFromPlugins.inferredFields
+    )
+
+    // Track which fields are supplied by plugins. This will determine
+    // if we can query by pure data or whether we have to resolve the
+    // fields first. See `./run-query.js`
+    _.forEach(inferredFieldNames, fieldName => {
+      pluginFieldTracking.add(fieldName)
```
pieh: I think we should namespace fieldNames - if one node type uses resolver for field `foo`, we don't want to revert to `sift` if we query other type and use field with same `foo` name

packages/gatsby/src/schema/query-loki.js
```diff
@@ -0,0 +1,194 @@
...
+  const gqlArgs = JSON.parse(JSON.stringify(rawGqlArgs))
+
+  const lokiArgs = convertArgs(gqlArgs)
+
+  const coll = getDb().getCollection(gqlType.name)
+
+  // Allow page creators to specify that they want indexes
+  // automatically created for their pages.
+  if (context.useQueryIndex) {
+    ensureIndexes(coll, lokiArgs)
```
pieh: How expensive this is when indexes were already applied? We definitely will run same query (with different context) when we use same template for multiple pages

This has also potential to keep unneeded indexes - for example when we modify filters (remove some fields from filter), but not sure if it's worth keeping track of that

benchmarks/query/recording.md
KyleAMathews: That's what I assume too but it's nice to know for sure.

There's probably something on npm but you can also add a setInterval in the code and log out the memory usage every .1 seconds or whatever. You could also log out different events too e.g. when query running starts etc. so you could visualize things easily.

packages/gatsby/src/schema/query-loki.js
Moocar: @pieh loki handles the case of the index already existing so there's no performance penalty.

But yes, this will keep old indexes around if filters are modified. My solution here for applying queries is very basic, but I think there's a lot more we could do here over time (see original PR description). My suggestion would be to keep as is, but iterate on this based on feedback

packages/gatsby/src/schema/run-query.js
Moocar: I hate the "resolve all the things" approach of run-sift, but I get why it exists. It's a shame, because the second we have to resolve all nodes, we lose any ability to perform index based lookups. 

For plugin filter fields, I think this will always be the case since their resolvers can do *anything*. But I wonder if there's an elegant solution for linked and mapped nodes. If we were using a graph database, then we would already have the link setup. 

One way or another, we'll need to find a proper solution eventually, because I imagine many sites will want to take advantage of linked and mapped filtering.

packages/gatsby/src/db/index.js
DSchau: Silly question, is db.listCollections() an _object_ or an _array_? I presumed array, but on second glance, it could actually be an object?

If it's the latter, then this _.forEach may actually be preferable!

I'm about to board a plane, but I'll add some more feedback on this tomorrow. Nice work!!

Moocar: Sorry, forgot to reply. It's an array. I've been using `for (const foo of someArray)` and I'm definitely preferring it

pieh:
I'll close this PR - as I think everything from here was already merged in more incremental chunks in past couple weeks

Issue #6883, opened 2018-07-31
gatsby-source-dropbox plugin


xuopled:
## Summary

Hello, I was looking for some plugin to fetch content from Dropbox but I did not find any. Is there a reason ? I found [this issue](https://github.com/gatsbyjs/gatsby/issues/324#issuecomment-225398868) where @KyleAMathews speaking about fetching content from Dropbox 2 years ago. Maybe there is some example somewhere

### Basic example

I want to build a web app using my sport activities as content. So I connect my services with [Tapirrik](https://tapiriik.com/) and export all my activities to dropbox. So It could be great to fetch content directly to my dropbox repo

LeKoArts:
> Is there a reason ?

No one created a source plugin yet, easy as that 😅
You could create it if you don’t want to wait for someone else to do it:
https://next.gatsbyjs.org/docs/source-plugin-tutorial/

xuopled:
Thank you for the link

oorestisime:
this looks fun. i ll work on a dropbox source plugin.

oorestisime:
Hey, so i ve worked out a plugin here https://github.com/oorestisime/gatsby-source-dropbox
it is not published yet since i am waiting on the PR i referenced here. You could install this along with the plugin-source-filesystem of the PR to test it :) 

After talking a bit on discord i dont think it makes sense to PR the whole plugin in the main gatsby repo so i am closing the issue. Feel free to open issues / propose features on the other repo.
For the moment i am tracking the release issues here https://github.com/oorestisime/gatsby-source-dropbox/issues/1

oorestisime:
@xuopled it is now released

PR #9349, opened 2018-10-24
Implement two fallbacks for getting extension when creating remote file

oorestisime:
I was working on a dropbox source plugin (#6883) and encountered two issues. Sometimes APIs provide temporary download files which in case of dropbox look like https://dl.dropboxusercontent.com/apitl/1/AAD2_49EPcISoEH8iSevIIxSHWAKbTkggAYuCFa87DzSyDm5MAa0qQPy0DCbIgOBcnMDwoHVF5uUKrxWr20ErtUr4kAkwG1ZA9aArmpoj_G01ce-wf8aT1YLkwbYq6Cm0yApE0fzFcewG3CpVdr3yPmKbOareDzmaB6POuPbnAjDeUkHfc3Cv7jjHGm_Wgfl-LlN62-2bM1qfiftRPm9yW0--gjDbfdUECHDRzqQvWB1PD8ShRNn5b9e04vOVW6nTjUqrlsRDOQ6NfJOulW9R2B2bt2JsmOobqQ_eIFVVrDId8eZsWwyHCqK17dQLY5_ZiO9a0ag4lIavN9JCkw9TVdx

So parsing the url and the path can not get extension which is normal. So fallbacks are:

1) download the file and try and guess the extension
2) let the user provide the extension

The 2 was needed because for example file-type can not detect markdown files.

Maybe i can factor out the guessing part in utils and try and try and provide some unittests, but i am not sure which file i could use for testing purposes since in the repo there are only .js and .md files which are not detectable. (Add a small picture in tests?)

### DIFF
--- a/packages/gatsby-source-filesystem/src/create-remote-file-node.js
+++ b/packages/gatsby-source-filesystem/src/create-remote-file-node.js
@@ -4,6 +4,8 @@ const crypto = require(`crypto`)
 const path = require(`path`)
 const { isWebUri } = require(`valid-url`)
 const Queue = require(`better-queue`)
+const readChunk = require(`read-chunk`)
+const fileType = require(`file-type`)
 
 const { createFileNode } = require(`./create-file-node`)
 const { getRemoteFileExtension } = require(`./utils`)
@@ -128,10 +130,9 @@ async function pushToQueue(task, cb) {
  * @param  {String}   url
  * @param  {Headers}  headers
  * @param  {String}   tmpFilename
- * @param  {String}   filename
  * @return {Promise<Object>}  Resolves with the [http Result Object]{@link https://nodejs.org/api/http.html#http_class_http_serverresponse}
  */
-const requestRemoteNode = (url, headers, tmpFilename, filename) =>
+const requestRemoteNode = (url, headers, tmpFilename) =>
   new Promise((resolve, reject) => {
     const responseStream = got.stream(url, {
       ...headers,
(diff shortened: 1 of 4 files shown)


packages/gatsby-source-filesystem/src/create-remote-file-node.js
```diff
@@ -196,22 +198,31 @@ async function processRemoteNode({
...
-      filename
     )
     // Save the response headers for future requests.
-    cache.set(cacheId(url), response.headers)
-
+    await cache.set(cacheId(url), response.headers)
+
+    // If the user did not provide an extension and we couldn't get one from remote file, try and guess one
+    if (ext === ``) {
+      const buffer = readChunk.sync(tmpFilename, 0, 4100)
```
DSchau: Where's 4100 coming from? What's the significance?

If we have a "magic number" let's encapsulate it like so, e.g.

```js
const MAX_BUFFER_SIZE = 4100

// later

const buffer = readChunk.sync(tmpFilename, 0, MAX_BUFFER_SIZE)
```

pieh: I think we should use constant provided by `file-type` lib (`fileType.minimumBytes`):
```
const readChunk = require('read-chunk');
const fileType = require('file-type');
 
const buffer = readChunk.sync('unicorn.png', 0, fileType.minimumBytes);
 
fileType(buffer);
//=> {ext: 'png', mime: 'image/png'}
```

oorestisime: heheh it was added 4 days ago :D i admit i just c/p the part with the buffer size its true that i had to avoid the magic number.

i ll correct this and mention the changes in the docs.

oorestisime:
I ve just corrected the magic number and i tried to add a few words in the readme file for the documentation.

PR #9351, opened 2018-10-24
Create README-TR.md

aytekk:
Turkish version of readme.md


pieh:
Thanks for PR, but we will not be able to maintain translations of our docs and it will get outdated very soon.

PR #9361, opened 2018-10-24
fix: bump webpack-guess version

DSchau:
This PR fixes the error gatsby-plugin-guess-js was throwing: `TypeError: ts.createNodeArray is not a function`
    
Bumping to 0.1.6 locally (with `resolutions` field in www/package.json) appeared to fix the issue.
    
Note: this is discussed (and fixed!) in guess-js/guess#76


DSchau:
UGH... weird merge. Will fix.

PR #9363, opened 2018-10-24
fix: upgrade guess-webpack version to avoid error

DSchau:
This PR fixes the error gatsby-plugin-guess-js was throwing: `TypeError: ts.createNodeArray is not a function`

Bumping to 0.1.6 locally (with `resolutions` field in www/package.json)
appeared to fix the issue.

Note: this is discussed (and fixed!) in guess-js/guess#76

### DIFF
--- a/packages/gatsby-plugin-guess-js/package.json
+++ b/packages/gatsby-plugin-guess-js/package.json
@@ -28,7 +28,7 @@
   "license": "MIT",
   "dependencies": {
     "@babel/runtime": "^7.0.0",
-    "guess-webpack": "~0.1.3"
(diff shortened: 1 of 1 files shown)


pieh:
hmm `0.1.6` should be caught by `~0.1.3` ( https://jubianchi.github.io/~0.1.3/0.1.6 ), I guess we can still merge this so it's easier to update if someone has older version?

DSchau:
I was using `yarn why` to diagnose this, and when I checked I had 0.1.3 _because_ of this plugin. Re-running yarn and no longer the case, so.... closing!

PR #9366, opened 2018-10-24
chore(gatsby): upgrade dependencies

alexandernanberg:
This PR upgrades a bunch of dependencies of the `gatsby` package. There are still some outdated ones but I think this is a good start! From what I can tell we were not affected by any of the breaking changes (mostly node version bumps).

I haven't run `yarn upgrade` to update all the packages in the lockfile (can add that to the PR if you'd like)

Although the tests are passing –  we probably want to do some testing with "real" projects, such as the starter kits.


pieh:
Did you test if it works with updated dependencies? `gatsby-dev` doesn't handle dependency changes so one way to do that would be use `npm/yarn pack` to create tarball of gatsby package with changed dependencies and install that tarball in test site

pieh:
@alexandernanberg I tried this and get same result when using `yarn add` but it works with `npm install` (for some reason) - I didn't investigate yet why it's happening :(

packages/gatsby/package.json
```diff
@@ -182,7 +180,6 @@
     "clean-test-bundles": "find test/ -type f -name bundle.js* -exec rm -rf {} +",
     "prebuild": "rimraf dist && rimraf cache-dir/commonjs",
     "prepare": "cross-env NODE_ENV=production npm run build",
-    "test-coverage": "node_modules/.bin/nyc --reporter=lcov --reporter=text npm test",
```
alexandernanberg: This isn't used right? The command itself doesn't do anything

alexandernanberg:
@pieh Ah I see, npm worked for me as well. I guess there is some problems with `pack` + `workspaces`, because I think `pack` needs a `yarn.lock` file to work.

I've tested in `gatsby-starter-blog`, `gatsby-starter-default` and did not experience any problems. Do we want to test in some larger projects also?

alexandernanberg:
@pieh Merged with master and I've now tested with a somewhat larger project I'm currently working on and didn't run into any problems.

Issue #9362, opened 2018-10-24
Gatsby V2 can't handle the generation of BIG sites.


rametta:
## Description

Gatsby V2 can't handle the generation of very big sites. We build around 2000 pages (however, the data is those pages is pretty big, e-commerce...), and the build step ends up failing.
We narrowed it down to this line of code: https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/redux/index.js#L82
stringify seems to crash with error:
```
const dataStr = JSON.stringify(data, null, 2)
                     ^
RangeError: Invalid string length
    at JSON.stringify (<anonymous>)
    at Object.<anonymous> (/Users/xxx/Code/stringy-tests/index.js:17:22)
    at Module._compile (module.js:652:30)
    at Object.Module._extensions..js (module.js:663:10)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
    at Function.Module.runMain (module.js:693:10)
    at startup (bootstrap_node.js:188:16)
    at bootstrap_node.js:609:3
error Command failed with exit code 1.
```
Note that this error comes from the reproducible snippet here: https://codepen.io/rametta/pen/xyQpRb

### Steps to reproduce
Use this Codepen to reproduce the JSON.stringify crashing: https://codepen.io/rametta/pen/xyQpRb
Open the codepen, change the 5 to 500000 then open the devtools console to see the error message.
We we're able to build our website by using the `streamed JSON.stringify` technique instead. PR coming soon.

### Expected result
The site should build for extra large JSON objects going through JSON.stringify()
### Actual result
JSON.stringify can't handle it and crashes.
### Environment
```
  System:
    OS: macOS High Sierra 10.13.6
    CPU: x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 8.11.1 - /usr/local/bin/node
    Yarn: 1.6.0 - /usr/local/bin/yarn
    npm: 5.6.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 69.0.3497.100
    Firefox: 52.0.2
    Safari: 12.0
  npmPackages:
    gatsby: ^2.0.5 => 2.0.9 
    gatsby-image: ^2.0.5 => 2.0.10 
    gatsby-plugin-accessibilityjs: 1.0.3 => 1.0.3 
[...]

PR #9370, opened 2018-10-24
feat(gatsby): use json-stream-stringify to serialize redux state

rametta:
This PR changes the way the internal gatsby store is saved to a disk to account for sites that have very large payloads.

Closes issue #9362


KyleAMathews:
We got a fair number of bug reports about circular references in the past so definitely need to handle that.

rametta:
@DSchau Yes definitely, I tested with my local build and with the old way was getting the error mentioned in the issue i opened (#9362) And now with this change, it works with any size file. (The file my build was generating was around 500mb)

pieh:
Here's another json streaming utility https://www.npmjs.com/package/bfj that has some handling for circular references

DSchau:
@pieh even better 👌 Thanks for that link!

KyleAMathews:
Hmmm https://www.npmjs.com/package/bfj#is-it-fast

rametta:
@KyleAMathews I agree that this is not ideal to slow down the build, but I think it's worth it. Here is why:
- This operation is only called maximum 2 times per build
- A slightly slower build is better than no build at all
- Small to medium sites will barely be affected
- BFJ is async so it does not block the event loop like the current implementation
- BFJ also safeguards against circular references
- Will allow bigger sites to be built

Here are my findings after experimenting with different stringify-ing methods

50,000 Records (~134 MB file)

| Method | Time to complete |
| ------ | ---------------- |
| native JSON.stringify | 1.10 seconds |
| json-stringify-safe | 1.19 seconds |
| bfj | 6.9 seconds |

100,000 Records (~251 MB file)

| Method | Time to complete |
| ------ | ---------------- |
| native JSON.stringify | 2.02 seconds |
| json-stringify-safe | 2.22 seconds |
| bfj | 13.36 seconds |

500,000 Records (~1.24 GB file)

| Method | Time to complete |
| ------ | ---------------- |
| native JSON.stringify | **CRASH** |
| json-stringify-safe | **CRASH** |
| bfj | 62 seconds |

rametta:
> `json-stream-stringify` is slower than `bfj` or than `json-stringify-safe` ?

Both

DSchau:
@rametta would you be able to run performance benchmarks with a sample site (e.g. something in [benchmarks](benchmarks)) to ensure this doesn't introduce a performance regression!

rametta:
@DSchau Here are my results with the create pages benchmark.
The results show that not much has changed in terms of build time, which is good.

(After every test, I deleted .cache and public folders manually)

`yarn build` Create Pages Benchmark

| Branch | Pages | Build 1 | Build 2 | Build 3 | AVG |
| ------ | ----- | ---- | ---- | ---- | ---- |
| v2.0.31 | 5,000 | 8.52 | 7.44 | 7.04 | 7.66 |
| v2.0.31 | 10,000 | 12.15 | 9.94 | 10.82 | 10.97 |
| v2.0.31 | 50,000 | 47.30 | 39.86 | 37.85 | 41.67 |
| v2.0.31 | 100,000 | 86.01 | 81.78 | 78.20 | 81.99 |
| This PR | 5,000 | 7.42 | 7.84 | 7.69 | 7.65 |
| This PR | 10,000 | 11.67 | 9.56 | 11.20 | 10.81 |
| This PR | 50,000 | 45.55 | 35.13 | 36.01 | 38.89 |
| This PR | 100,000 | 80.09 | 80.77 | 81.82 | 80.89 |

KyleAMathews:
Sorry for not being more explicit. Can you use the markdown benchmark instead? The create pages benchmark doesn't add anything basically to the cache so it's not a good test for this.

KyleAMathews:
E.g. I just tried a 50k markdown site and it OOMed when it tried to write out the cache file.

rametta:
@KyleAMathews No problem, so I ran the markdown benchmark a few times, and realized I would have to make each node created have content that was VERY large, so i replaced the `content` property of every node made to have a very large snippet of lorum ipsum that would have to be written to .cache/redux-state.json. So even trying to create 5k pages with the old way i would run into the error of:
```
info bootstrap finished - 82.756 s

error Invalid string length


  RangeError: Invalid string length

  - JSON.stringify

  - stringify.js:5 stringify
    [test_stream]/[json-stringify-safe]/stringify.js:5:15

  - index.js:90 saveState
    [test_stream]/[gatsby]/dist/redux/index.js:90:23

  - index.js:118 emitter.on
    [test_stream]/[gatsby]/dist/redux/index.js:118:5

  - mitt.js:1
    [test_stream]/[mitt]/dist/mitt.js:1:268

  - Array.map
```

Then when I switched to use my PR, it would build no problem. (Granted I had to provide 12GB of memory to my node command, but at least it mimics the project I am working on now that lead me to find this issue) When my build passed I had a redux-state.json file of about 990MB, whereas the current gatsby would crash anywhere around 500MB file size because that seems to be the limit of JSON.stringify or safe-stringify implementation.

KyleAMathews:
Awesome! And also no perf regressions?

rametta:
I don't think so, if there is, I don't know how to find them

DSchau:
@rametta one neat trick I built into that markdown pages benchmark is that you can conditionally (with environment variables) change the amount of content for each page :)

So here's how that works:

- I expose a `NUM_ROWS` variable, which will basically create a huge markdown table, which tends to lead to a huge markdown AST
- So running, e.g. `NUM_PAGES=1500 NUM_ROWS=100 yarn build` will build out 1500 pages with a fair amount of content

Feel free to increase these upward bounds to see if we're still in the clear as far as serialized redux state!

KyleAMathews:
>So this will test the cache but not redux state. Maybe it's worth building in something additional to test the node creation process that follows a similar approach?

Not sure what you mean by this?

KyleAMathews:
The raw node content is in the redux state (though not the AST as you mention)

DSchau:
> So this will test the cache but not redux state. Maybe it's worth building in something additional to test the node creation process that follows a similar approach?

@KyleAMathews yeah, whoops! I was thinking only the cache was touched here, but updating that env var will increase the size of the generated node too, so it'll actually test both. The other info is accurate :)

rametta:
> * So running, e.g. `NUM_PAGES=1500 NUM_ROWS=100 yarn build` will build out 1500 pages with a fair amount of content

Ah okay, that's good to know, thanks!

KyleAMathews:
Even without extra rows, the markdown benchmark OOMed somewhere before 50k pages as I mentioned above https://github.com/gatsbyjs/gatsby/pull/9370#issuecomment-433636329.

Just tested this PR w/ the same benchmark and 100k pages and it finished just fine! Super exciting!

```bash
Kyles-MacBook-Pro:markdown (stringify-stream $)$ NUM_PAGES=100000 gatsby build
success open and validate gatsby-configs — 0.006 s
success load plugins — 0.109 s
success onPreInit — 0.715 s
success delete html and css files from previous builds — 0.004 s
success initialize cache — 22.712 s
success copy gatsby files — 0.058 s
success onPreBootstrap — 0.012 s
success source and transform nodes — 82.279 s
success building schema — 5.159 s
success createPages — 29.534 s
success createPagesStatefully — 0.017 s
success onPreExtractQueries — 0.000 s
success update schema — 12.211 s
success extract queries from components — 0.195 s
success run graphql queries — 507.226 s — 100001/100001 197.16 queries/second
success write out page data — 1.296 s
success write out redirect data — 0.001 s
success onPostBootstrap — 0.353 s

info bootstrap finished - 664.473 s

success Building production JavaScript and CSS bundles — 25.580 s
success Building static HTML for pages — 35.440 s — 100001/100001 3060.76 pages/second
info Done building in 726.477 sec
```

pieh:
Hey, I've noticed few problems with this implementation:
 - in some scenarios when build part is really fast (i.e. rendering just 1 static html) `gatsby build` will finish before save completes leaving with `redux-state.json` in unfinished state: 
screen shot 2018-11-04 at 20 13 49
 - there can also be scenarios where we try to do multiple writes in parallel (we don't return promise and even then `lodash.debounce` doesn't wait for promise I think?)

What I think we would need to do is:
a. write to tmp file and rename it to `redux-state` on completion (to not leave with `redux-state` in incomplete state - for example if exit `gatsby develop` mid-writing)
b. in `commands/build` actually wait for state saving to finish
c. investigate better debouncing that would queue next write instead of running in parallel

pieh:
I did some work on above and noticed another problem - the way this handles cyclic refs (or rather repeated refs for same object) is cool (see `gatsbyDependencies` or `miscDependencies` which are subsets of `allDependencies` array): https://gist.github.com/pieh/5c2a52d4744bb20678db07934172f4d6
but problem is we don't recreate those references when building state on next run resulting in different weird data and in consequence different schema and failing queries - this seems to mentioned in [`json-stream-stringify` docs](json-stream-stringify):
> To restore cyclical structures; use [Crockfords Retrocycle method](https://github.com/douglascrockford/JSON-js) on the parsed object (not included in this module).

So I think we should (at least temporarily) revert this and fix those issues before applying this change again, because right now cache is basically broken

pieh:
PR to revert this: https://github.com/gatsbyjs/gatsby/pull/9896

ivanoats:
What’s the best way to incorporate these changes without forking Gatsby? A P.R. To make it an option for large sites?

PR #9377, opened 2018-10-25
feat(using-gatsby-image): refresh design, switch to emotion

fk:
Didn't have teh good internets for a couple of hours today, but a folder of unsplash.com images and just had touched on `using-gatsby-image` yesterday … so I decided to make some good use of the mostly offline dev time and revisit https://using-gatsby-image.gatsbyjs.org/.

Here's a video: https://www.dropbox.com/s/axq2zp8lna9y71y/using-gatsby-image.mov?dl=0

image

image

bildschirmfoto 2018-10-25 um 02 42 36

bildschirmfoto 2018-10-25 um 02 43 45

- refresh'n'Gatsby-fy 💜UI and layout
  - noteworthy: hopefully improve homepage UX on mobile
- move from `glamor` to `emotion`
- refactor most things to [`styled`](https://emotion.sh/docs/styled) components
- switch to system font stack; images and video show [Larsseit](https://www.myfonts.com/fonts/typedynamic/larsseit/) for nav and headlines which I've been playing around locally
- refactor layout and page components a little
- fix active navigation state (adjust to `@reach/router`)
- throw in `header`, `nav`, and `main`
- add Gatsby monogram linking to https://www.gatsbyjs.org/ 
- add "Docs" link to main navigation (pointing to https://www.gatsbyjs.org/packages/gatsby-image/)
- fancy bullet points for the homepage "Out of the box it…" list
- "Documentation" block at the bottom of every page
  - add link to https://www.gatsbyjs.org/docs/using-gatsby-image/
- replace a couple of great unsplash.com images with a bunch of other great unsplash.com images ;)
- fetch all but one image from unsplash.com, removing a ton of data from the repo thanks to @DSchau (#9605) 🎉 

Todo:

- [x] Fix image credits (``)
- [x] Optimize image query `width`/`maxWidth`
- [x] Refactor CSS to template literals…

/cc @gatsbyjs/website


examples/using-gatsby-image/src/components/image-gallery.js
```diff
@@ -0,0 +1,103 @@
...
+        margin: `0 auto`,
+        [presets.Mobile]: {
+          columnCount: 2,
+        },
+        [presets.Tablet]: {
+          columnCount: 3,
+        },
+        "@media screen and (min-width: 1400px)": {
+          columnCount: 4,
+        },
```
jlengstorf: For consistency, can we put everything into `presets` or pull everything out of it?

examples/using-gatsby-image/src/components/floating-image.js
```diff
@@ -0,0 +1,40 @@
...
+          "@media (min-width: 500px)": {
+            display: `none`,
+          },
+        },
+      }}
+      fixed={imageMobile}
+    />
+    <Img
+      style={{ display: `inherit` }}
+      css={{
```
jlengstorf: We had a discussion about styling that landed on using the styled component approach for Emotion, meaning we'll want to refactor this to be:

```jsx
const GatsbyImage = styled(Img)`
	margin-bottom: ${rhythm(options.blockMarginBottom * 2)};
	/* ... */
`;
```

examples/using-gatsby-image/src/components/image-gallery.js
```diff
@@ -0,0 +1,103 @@
...
+            border: `4px solid transparent`,
+            breakInside: `avoid`,
+            position: `relative`,
+            [presets.Mobile]: {
+              borderWidth: 8,
+            },
+            [presets.Desktop]: {
+              borderWidth: 12,
+            },
+            "& img": {
```
jlengstorf: Using the styled component approach will make this easier, too:

```js
const Component = styled('div')`
	background-color: red;

	img {
		border-radius: 2px;
	}
`;
```

examples/using-gatsby-image/src/components/image-gallery.js
fk: Oh, missed that one (and probably more) … 👍

examples/using-gatsby-image/src/components/floating-image.js
fk: Yeah, followed that and already did use `styled` for stuff I had to touch, but not everywhere else (yet). Happy to do so consistently!

fk:
> Maybe we can reduce the raw images down to whatever the maximum size is to avoid the impact on the overall repo size?

👍

examples/using-gatsby-image/src/components/floating-image.js
jlengstorf: Ah, I thought this was all new code.

I'll let @amberleyromo make the call about how much we should refactor. My vote would be to just knock it out while we're in here, but she's the final word on this.

amberleyromo: @fk I would say if you feel it won't take too long and you feel good about it, go ahead and do it. Would be nice to have it as a good standalone example of the code style, since www will be switched over incrementally.

If not, don't stress, we can decide where it fits into priorities.

examples/using-gatsby-image/src/components/image-gallery.js
KyleAMathews: I wouldn't call that "easier", just different

pieh:
> Also, a lot of these new images are _huge_ — I'm not sure we want to put an additional 30MB of images into the repo for people to download. Maybe we can reduce the raw images down to whatever the maximum size is to avoid the impact on the overall repo size?

How about creating unsplash source plugin? Or even just hardcoding urls to download images from unsplash for this example site?

examples/using-gatsby-image/src/components/image-gallery.js
jlengstorf: Sorry — "easier" is in reference to the syntax being less complex, but agreed that it's just different.

examples/using-gatsby-image/src/components/floating-image.js
```diff
@@ -0,0 +1,40 @@
+import React from "react"
+import Img from "gatsby-image"
+
+import { rhythm, options } from "../utils/typography"
+
+const FloatingImage = ({ imageMobile, imageDesktop }) => (
+  <React.Fragment>
+    <Img
+      style={{ display: `inherit` }}
```
DSchau: nit: but this may as well go in the css prop (or as the styled component approach, I think both work well)

examples/using-gatsby-image/src/components/image-gallery.js
```diff
@@ -0,0 +1,103 @@
...
+import numeral from "numeral"
+
+import { options, scale } from "../utils/typography"
+import presets from "../utils/presets"
+
+const UnsplashMasonry = edges => (
+  <div
+    css={{
+      background: `white`,
+      margin: `0 -`,
```
DSchau: Is this valid?

examples/using-gatsby-image/src/components/navigation.js
```diff
@@ -1,84 +1,199 @@
...
+          "&:hover": {
+            backgroundPosition: `100% 0 !important`,
+            color: `initial !important`,
+            WebkitTextFillColor: `initial !important`,
+          },
+        }),
+      }
+    : { className: css({ "&&": { ...linkStyle } }) }
+
+const Header = styled(`header`)({
```
DSchau: another nit, but I'd probably always prefer the shorthand, e.g. `styled.header` vs. the function call.

More of a personal preference here, but I think it's _probably_ worth following that.

examples/using-gatsby-image/src/utils/colors.js
```diff
@@ -0,0 +1,17 @@
+const colors = {
```
DSchau: I like this a lot!

examples/using-gatsby-image/src/utils/presets.js
```diff
@@ -0,0 +1,29 @@
...
+module.exports = {
+  colors,
+  Mobile: `@media (min-width: 400px)`,
+  Phablet: `@media (min-width: 550px)`,
+  Tablet: `@media (min-width: 750px)`,
+  Desktop: `@media (min-width: 1000px)`,
+  Hd: `@media (min-width: 1200px)`,
+  radius: 2,
+  radiusLg: 4,
+  animation: {
```
DSchau: Stealing this for the future!

examples/using-gatsby-image/src/components/navigation.js
amberleyromo: @jlengstorf this is why we need a style definition i think. pretty sure elsewhere people have said the ('...') style is preferred.

examples/using-gatsby-image/src/components/image-gallery.js
fk: Huh…good catch!

examples/using-gatsby-image/src/components/navigation.js
jlengstorf: Yeah, @DSchau we talked about this in another thread and had started using `styled('header')` as a standard. We can change that if there's a strong opinion either way — we just want to keep it consistent everywhere.

examples/using-gatsby-image/src/components/image-gallery.js
fk: Done!

examples/using-gatsby-image/src/components/image-gallery.js
fk: When we decided on `styled`, did we also decide against [object styles](https://emotion.sh/docs/object-styles)?

examples/using-gatsby-image/src/components/floating-image.js
fk: @jlengstorf Not all new, just a little realign :D (also FWIW not all images are new, I tried to balance out old/new amount, and plan on still removing a bunch more)

@amberleyromo I'll give it a shot because I also want to use `styled` a little more. Btw.…already got annoyed a couple times _having_ to name things again ;)

examples/using-gatsby-image/src/components/image-gallery.js
jlengstorf: @fk yeah, we decided to prefer tagged template literals. Keep things as CSS-ey as we can to keep the APIs familiar to devs who maybe haven't done as much CSS-in-JS without giving up the benefits of CSS-in-JS.

examples/using-gatsby-image/src/components/image-gallery.js
```diff
@@ -9,7 +9,6 @@ const UnsplashMasonry = edges => (
   <div
     css={{
       background: `white`,
-      margin: `0 -`,
```
jlengstorf: 🤣

examples/using-gatsby-image/src/components/floating-image.js
fk: `gatsby-image` sets a couple of inline styles on its outer container and allows overriding via its `style` prop. One of these inline styles is `display: inline-block;`.

In this case we have two images, one for mobile and one for desktop, and toggle their visibility with `display`, so we want to override the `gatsby-image` default to make our own styles work.

KyleAMathews:
We could write a little extension to source-airtable that uses createRemoteFileNode to pull down the actual image.

KyleAMathews:
Extension in the site's gatsby-node.js and onCreateNode

examples/using-gatsby-image/src/components/navigation.js
greglobinski: @amberleyromo  @DSchau  @jlengstorf  @fk 

I'm not sure if there is a way to config `babel-plugin-emotion` to use shorthand syntax with components, for now if I try  to use shorthand syntax like this (on www)

```
import { Link } from "gatsby"

const BlockLink = styled.Link`
  border-radius: ${presets.radiusLg}px;
  box-shadow: 0 1px 8px rgba(0, 0, 0, 0.2);
  display: flex;
  height: 100%;
  padding: ${rhythm(3 / 4)};

...
```
instead of `styled(Link)` I got the error, 
```
Uncaught Error: You're trying to use the styled shorthand without babel-plugin-emotion.
Please install and setup babel-plugin-emotion or use the function call syntax(`styled('Link')` instead of `styled.Link`)
```
so for consistency the full syntax is preferred,  we can think of shorthand syntax if it could be used with components not only with primitives

examples/using-gatsby-image/src/components/image-gallery.js
greglobinski: @fk First of all the page looks great :heart: 



I support @jlengstorf with using emotion full syntax with tagged template literals https://github.com/gatsbyjs/gatsby/pull/9377#discussion_r229269948
```
styled("div")`
...
`
```
There are a couple of ways to compose styles.
1. The `css` way

```
const baseStyle = css`
  flex-grow: 1;
  font-size: 0.85rem;
  margin: 0;
  color: ${colors.gray.lightCopy};
  padding: ${rhythm(0.5)} 0 0;
`
const Description = styled("p")`
  ${baseStyle};
  color: red;
```
2. Re-styling components way
```
const Description = styled("p")`
  flex-grow: 1;
  font-size: 0.85rem;
  margin: 0;
  color: ${colors.gray.lightCopy};
  padding: ${rhythm(0.5)} 0 0;
`

const DescriptionBig = styled(DescriptionBase)`
  font-size: 2rem;
`
```
3. merging css classes with `cx` method
https://emotion.sh/docs/cx

4. And of course using dynamic styles
https://emotion.sh/docs/composition#composing-dynamic-styles

I think that we should prefer the fourth and the second method. 

/cc @amberleyromo @jlengstorf @DSchau

KyleAMathews: You can make styles as strings work but it feels like a big step back imo

greglobinski: @KyleAMathews Sure, it could be that I do not spot some use cases when the objects work better.

jlengstorf: I think there are edge cases where styles work better, but overall moving to a JS abstraction over another language is a barrier for contributors who already know CSS.

My preference is to avoid logic in styles, and instead to have base styles that get extended where necessary:

```
const Button = styled('button')`
  background: red;
  border-radius: 0.5rem;
  color: white;
  font-size: 1.5rem;
`;

const BigButton = styled(Button)`
  font-size: 2.5rem;
`;
```

This makes it extremely clear that `BigButton` is an extension of `Button`, and doesn't require keeping how both JS and CSS work in your brain at the same time. 

Conversely, the JS version looks less like styles and (for my brain) is significantly harder to parse whether it's logic for the UI or logic for the styles:

```
const Button = styled('button')(props => ({
  background: 'red',
  borderRadius: '0.5rem',
  color: 'white',
  fontSize: props.isBig ? '2.5rem' : '1.5rem',
}));
```

It's also less clear _which_ button is being used in the second case, because the styles are now tied to the props and the result is hidden.

It's definitely going to come down to preference at the end of the day, but I feel pretty strongly that JS styles don't create enough benefit to justify the extra cognitive overhead.

examples/using-gatsby-image/src/components/navigation.js
fk: 👌Sticking to function call.

examples/using-gatsby-image/src/utils/presets.js
fk: Tried to improve this a bit!

PR #9379, opened 2018-10-25
query-benchmark

Moocar:
Part of WIP PR #9338. Submitting this as its own PR to reduce the size of the other PR and since it's useful on its own. The recording.md won't make sense without the other PR, but I think that's ok.

### DIFF
--- a/benchmarks/query/bin/runQueryTiming.sh
+++ b/benchmarks/query/bin/runQueryTiming.sh
@@ -1,0 +1,8 @@
+# Run the build (after purging .cache) and output the amount of time
+# taken by the query execution phase
+#
+# run with `bin/runQueryTiming.sh`
+
+output=$(rm -rf .cache && gatsby build | grep "run graphql queries")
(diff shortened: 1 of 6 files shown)


benchmarks/query/bin/runQueryTiming.sh
```diff
@@ -0,0 +1,8 @@
+# Run the build (after purging .cache) and output the amount of time
+# taken by the query execution phase
+#
+# run with `bin/runQueryTiming.sh`
+
+output=$(rm -rf .cache && gatsby build | grep "run graphql queries")
+echo $output | cut -d' ' -f 6
+
```
pieh: this (kind of) limits where we can use this benchmark (cough - windows) :) but that's just utility to grab time for running queries, so I wouldn't say that it is a blocker

Moocar: It's a sign of the times when windows doesn't even enter my head. I'll leave it in as an optional useful tool for non windows users.

PR #9386, opened 2018-10-25
feat(gatsby-source-wordpress): normalize baseUrl

gustavpursche:
I'm not very good at reading documentation and didn't see, that `baseUrl` should not contain a trailing slash nor the protocol. The error, which is thrown because of that mistake is quite cryptic, so this PR adds a little of normalization to `baseUrl`:

- remove trailing slashes
- remove protocol

A different approach might be, to print a proper error, which tells me about the mistake.


pieh:
If we can automatically fix this / recover, let's go with that instead of showing errors to the users - will review in a sec

PR #9387, opened 2018-10-25
remove asterisk to note active doc

LukeSchlangen:
It seems like the asterisk in the `.yaml` file is noting which docs should be gray to indicate that they are not complete. The `Adding analytics` document seems complete, so I removed the asterisk.


LukeSchlangen:
You're welcome!

PR #9390, opened 2018-10-25
feat(examples): Update emotion

LekoArts:
- Update the dependencies
- Change all imports to `react-emotion`
- Fix eslint errors (e.g. PropTypes)
- Remove stray backticks
- Only use injectGlobal once (not like in the PrismJS styles which is very chaotic)

I couldn't find something online if [this](https://www.styled-components.com/docs/api#deprecated-injectglobal) also applies to emotion, but I think it's a good change nevertheless:

> We do not encourage the use of this. Try to use it once per app at most, if you must, contained in a single file.

### DIFF
--- a/examples/using-emotion-prismjs/src/components/layout.js
+++ b/examples/using-emotion-prismjs/src/components/layout.js
@@ -19,7 +19,7 @@ const link = css`
   color: inherit;
 `
 
-class Layout extends React.Component {
+class PureLayout extends React.Component {
   render() {
     const HeadingTag = this.props.isIndex ? `h1` : `h3`
     const { siteTitle, pageTitle } = this.props
@@ -35,7 +35,6 @@ class Layout extends React.Component {
           />
           <meta name="referrer" content="origin" />
         </Helmet>
-        ` `
         <div className={indexContainer}>
           <HeadingTag>
             <Link className={link} to={`/`}>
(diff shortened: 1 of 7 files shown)


ThomOrlo:
> I couldn't find something online if [this](https://www.styled-components.com/docs/api#deprecated-injectglobal) also applies to emotion, but I think it's a good change nevertheless:

@DSchau same as styled-components 4, with Emotion 10 :
`import { Global, css } from '@emotion/core'`

https://github.com/emotion-js/emotion/issues/637
https://next.emotion.sh/docs/globals

LekoArts:
I was referring to the quote of injectGlobal 😅
Not emotion 10

examples/using-emotion-prismjs/src/components/layout.js
```diff
@@ -19,7 +19,7 @@ const link = css`
   color: inherit;
 `
 
-class Layout extends React.Component {
+class PureLayout extends React.Component {
```
DSchau: Why the change?

LekoArts: eslint wanted an explicite name (not `export default props`) and since I can't name them both `Layout` I gave this one another name.

DSchau: I guess I would've just called it BaseLayout or something! Pure implies (to me!) something like a PureComponent i.e. a component that doesn't re-render when its props are shallowly equal.

LekoArts: And ai chose Pure because it’s without GraphQL 😀 But I don’t care really

DSchau: Yeah, not a big deal either way, _especially_ because this is an example. Let's just merge :) Thank you!

Issue #9393, opened 2018-10-25
[site showcase] change initial load from 9 sites to 12


amberleyromo:
Currently on first load of [`/showcase/`](https://www.gatsbyjs.org/showcase) we load nine sites. This produces empty space at some resolutions. Changing the initial number loaded to 12 will be smoother at more resolutions.

outofgamut:
also had a PR in progress for this

amberleyromo:
Yep, marking as claimed. Thanks @abilash13!

amberleyromo:
Ah, ok @outofgamut.

@abilash13 could you check out the rest of the hacktoberfest issues? I'll be adding new Ecosystem ones periodically as well.

Apologies, I'll make it more clear that the issue should be claimed before a PR is opened. Y'all are too quick on the draw..!

abilash13:
Ah no worries. Thanks!

amberleyromo:
Thank you @abilash13! 🙏🏼

amberleyromo:
Closed in #9395

PR #9395, opened 2018-10-25
fix: display 12 sites in filtered sites grid

outofgamut:
- fix: display 12 sites in featured site list
- [#9393](https://github.com/gatsbyjs/gatsby/issues/9393)


outofgamut:
got it! making the change right now.

PR #9400, opened 2018-10-25
feat(www): Filter posts by date

### LINKED ISSUE #9298: Filter posts in www by date
Currently all blog posts on `www` are displayed. As a result, when we calendar posts for the future, sometimes we need to hold merging them until the future, and then we forget....

In order to avoid this problem, we want to be able to merge posts right away, but have them not show up in lists until they are published. 

As a result: 
* Blog posts should continue to page be created as pages, no matter if in the past or future...
* ...but they should only show up in the home page, /blog page, and /tag pages if the date is not in the future. 
* Filtering should be server-side (probably can't be client-side due to pagination unfortunately) and future should be calculated as GMT future (to be friendly to European audience).

jgierer12:
Closes #9298


kbariotis:
@jgierer12 Great work! Allow me to add a comment.

I see that you added a new frontmatter property to distinguish the posts that can be displayed and those can't. I was actually wondering if there is a simpler way to avoid having to have an extra property. We could filter the blog posts array based on their published date, which exists as a property already. That should happen in two different places, one in `gatsby-node.js` to calculate the pagination/tags etc based on the number of the blog posts prior to today and also in each of these components to filter those blog posts and not show them at all. The rest can stay as are now. 

What do you think?

pieh:
Did you check if all current blog posts will have `released` field set to true? In gatsby we use `moment` library to parse date strings and in here You used `Date` class to do it and there might be some differences

www/gatsby-node.js
```diff
@@ -231,9 +232,13 @@ exports.createPages = ({ graphql, actions }) => {
         return undefined
       })
 
+      const releasedBlogPosts = blogPosts.filter(post =>
+        _.get(post, `node.fields.released`)
+      )
```
pieh: I think we should use same graphql filter as in `www/src/templates/template-blog-list.js` so filtering happens inside query. We would need to split our `allMarkdownRemark` queries earlier into 2 separate ones - one for blog posts and one for the rests. Otherwise the `next` / `previous` might be unreliable

jgierer12: I preferred this approach so we don't have to duplicate these lines: https://github.com/gatsbyjs/gatsby/blob/f06f6c37698ccdd91b0f550fa30ab55f1d47cae9/www/gatsby-node.js#L251-L266

The `next` problem was addressed in 81a1e50.

If you would prefer to separate the posts in the query anyways, I'll do that of course.

jgierer12:
My bad, I didn't know `moment` was the preferred date library in this project. I did verify that all existing posts are marked as released. But to be sure, I will change it to `moment`.

pieh:
> My bad, I didn't know `moment` was the preferred date library in this project. I did verify that all existing posts are marked as released. But to be sure, I will change it to `moment`.

It's not "preferred" - it's just what is used in gatsby itself ( https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/schema/types/type-date.js ), so we definitely want to support same formats

www/gatsby-node.js
pieh: No, that's fine 👍 - it produces same output

PR #9408, opened 2018-10-25
Add unstructured data example

amberleyromo:
This adds the [unstructured data site example](https://github.com/jlengstorf/gatsby-with-unstructured-data/tree/master) built by @jlengstorf into the `examples` dir in the gatsby repo.


amberleyromo:
What about as a local plugin example?

jlengstorf:
Hmm... I feel like the comparison between these examples is actually the most critical part. Maybe @gatsbyjs/core can weigh in here? Do we have any prior work that's comparing two approaches in Gatsby? Any ideas on how best to show that off?

amberleyromo:
Making it two sites with mirrored content with different approaches could accomplish the same thing well. In the "branch" solution, we still have to message in the README to 1) look at the branch and 2) what it's accomplishing. 

With a "local plugin" example, we could show the gatsby-layer example, and also have it be its own standalone example -- still supported by messaging in both READMEs (same method as currently)

jlengstorf:
@amberleyromo Okay, I'm sold. Sounds good!

amberleyromo:
@jlengstorf Added the other example and modified both README files 👍

examples/using-local-plugins/plugins/gatsby-source-pokeapi/gatsby-node.js
```diff
@@ -0,0 +1,56 @@
+const axios = require("axios")
```
DSchau: Complete and total nit-pick, but did we change something where strings use double quotes or something now? 🤷‍♂️

examples/using-local-plugins/plugins/gatsby-source-pokeapi/gatsby-node.js
```diff
@@ -0,0 +1,56 @@
...
+const createNodeHelpers = require("gatsby-node-helpers").default
+
+const get = endpoint => axios.get(`https://pokeapi.co/api/v2${endpoint}`)
+
+const getPokemonData = names =>
+  Promise.all(
+    names.map(async name => {
+      const { data: pokemon } = await get(`/pokemon/${name}`)
+      const abilities = await Promise.all(
```
DSchau: Minor suggestion!

```suggestion
      const abilities = await Promise.all(
      
      pokemon.abilities.map(({ ability: { name: abilityName } }) =>
          get(`/ability/${abilityName}`)
            .then(res => res.data)
        )
      )
```

PR #9412, opened 2018-10-26
new gatsby default starter homepage

muhaimincs:


pieh:
This doesn't seem right - we are showing hello-world starter in this tutorial (which didn't change).

shannonbux:
Yes @muhaimincs, the `hello-world` starter looks the same. Just the `default-starter` changed.

If you think it's worth adding images to the `hello-world` starter, want to open a new issue for that?

Issue #9373, opened 2018-10-24
Exclude pages from having Google Tag Manager (GTM) code added


SimonIMG:
## Summary

The [gatsby-plugin-google-tagmanager](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-google-tagmanager) plugin adds GTM code to all pages of a Gatsby site. Is there any way to exclude a page from having the GTM snippet applied to it?

### Basic example

Adding a blank page on a website so a service like [Uptime Robot](https://uptimerobot.com/) can check if our site is live (as opposed to pinging the homepage).

Instead of a custom .txt or .html file added to the public folder that could get wiped out during a `gatsby build`, I created a blank Gatsby page in _/src/pages/_ to be generated. It is literally just this...

```js
import React from 'react'

const CheckIfIAmLive = () => <div>Live</div>

export default CheckIfIAmLive
```

Yes, the page could be excluded from Google Analytics, but it would be great if there was a feature or a way to exclude this page from featuring the GTM code in the source and unnecessarily adding all the tags to the page.

### Motivation

I can think of multiple applications for excluding GTM from pages other than website uptime monitoring: admin areas, custom promo pages, easter eggs.

kakadiadarpan:
@SimonIMG Currently `gatsby-plugin-google-tagmanager` does not provide an option of excluding a page from having the GTM snippet applied to it.

But like you said it would be really useful.

The required changes needs to be done in [this file](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-google-tagmanager/src/gatsby-ssr.js). `onRenderBody` function gets a parameter called `pathname`, we can provide an option in `gatsby-plugin-google-tagmanager` to accept a list of excluded paths which we can compare with the pathname and exclude a page from having the GTM snippet.

Would you like to work on this and raise a PR?

piotrkwiecinski:
I can give it a try on the weekend if you could wait.

SimonIMG:
My JS skills are pretty noob level at the moment (trawling through Scott Tolinksi/Web Bos tutorials to up my JS and React skills :) ).

But I rate myself pretty good at testing, UX and documentation. More than happy to wait and happy to pitch into updating docs.

pieh:
@SimonIMG You should be able to that in GTM console. GTM works differently than GA and excluding GTM script can cause weird behaviours. See discussion in #9413

kakadiadarpan:
[...]

PR #9413, opened 2018-10-26
Exclude pages from having GTM snippet applied to it

estrada9166:
#9373

### DIFF
--- a/packages/gatsby-plugin-google-tagmanager/src/gatsby-ssr.js
+++ b/packages/gatsby-plugin-google-tagmanager/src/gatsby-ssr.js
@@ -2,13 +2,24 @@ import React from "react"
 import { oneLine, stripIndent } from "common-tags"
 
 exports.onRenderBody = (
-  { setHeadComponents, setPreBodyComponents },
+  { setHeadComponents, pathname, setPreBodyComponents },
   pluginOptions
 ) => {
   if (
     process.env.NODE_ENV === `production` ||
     pluginOptions.includeInDevelopment
   ) {
+    if (typeof pluginOptions.exclude !== `undefined`) {
+      const minimatch = require(`minimatch`)
+      const excludedPaths = pluginOptions.exclude
+      // loop all the excluded paths to validate if the pattern match with the 
+      // actual pathname
+      for (let i = 0; i < excludedPaths.length; i++) {
+        const exclude = excludedPaths[i]
+        if (minimatch(pathname, exclude)) return
+      }
+    }
+
     const environmentParamStr =
       pluginOptions.gtmAuth && pluginOptions.gtmPreview
         ? oneLine`
--- a/packages/gatsby-plugin-google-tagmanager/src/gatsby-ssr.js
+++ b/packages/gatsby-plugin-google-tagmanager/src/gatsby-ssr.js
@@ -1,6 +1,6 @@
 import React from "react"
 import { oneLine, stripIndent } from "common-tags"
-const minimatch = require(`minimatch`)
+import minimatch from 'minimatch'
 
 exports.onRenderBody = (
   { setHeadComponents, pathname, setPreBodyComponents },
--- a/packages/gatsby-plugin-google-tagmanager/src/gatsby-ssr.js
+++ b/packages/gatsby-plugin-google-tagmanager/src/gatsby-ssr.js
@@ -1,5 +1,6 @@
 import React from "react"
 import { oneLine, stripIndent } from "common-tags"
+const minimatch = require(`minimatch`)
 
 exports.onRenderBody = (
   { setHeadComponents, pathname, setPreBodyComponents },
@@ -10,13 +11,9 @@ exports.onRenderBody = (
     pluginOptions.includeInDevelopment
   ) {
     if (typeof pluginOptions.exclude !== `undefined`) {
-      const minimatch = require(`minimatch`)
-      const excludedPaths = pluginOptions.exclude
-      // loop all the excluded paths to validate if the pattern match with the 
-      // actual pathname
-      for (let i = 0; i < excludedPaths.length; i++) {
-        const exclude = excludedPaths[i]
-        if (minimatch(pathname, exclude)) return
+      if (pluginOptions.exclude.some(excludedPath => minimatch(pathname, excludedPath))) {
+        // bail early if current path is excluded
+        return
       }
     }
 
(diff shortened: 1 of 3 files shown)


packages/gatsby-plugin-google-tagmanager/src/gatsby-ssr.js
```diff
@@ -2,13 +2,24 @@ import React from "react"
...
-  { setHeadComponents, setPreBodyComponents },
+  { setHeadComponents, pathname, setPreBodyComponents },
   pluginOptions
 ) => {
   if (
     process.env.NODE_ENV === `production` ||
     pluginOptions.includeInDevelopment
   ) {
+    if (typeof pluginOptions.exclude !== `undefined`) {
+      const minimatch = require(`minimatch`)
```
pieh: Please import at the top of file

packages/gatsby-plugin-google-tagmanager/src/gatsby-ssr.js
```diff
@@ -2,13 +2,24 @@ import React from "react"
...
+      const excludedPaths = pluginOptions.exclude
+      // loop all the excluded paths to validate if the pattern match with the 
+      // actual pathname
+      for (let i = 0; i < excludedPaths.length; i++) {
+        const exclude = excludedPaths[i]
+        if (minimatch(pathname, exclude)) return
+      }
```
pieh: maybe something like:
```js
if (pluginOptions.exclude.some(excludedPath => minimatch(pathname, excludedPath))) {
  // bail early if current path is excluded
  return
}
```

estrada9166:
@pieh That's a valid point! I made the changes based on [this comment](https://github.com/gatsbyjs/gatsby/issues/9373#issuecomment-432938534) what do you think?

pieh:
Was just reading on tag manager because I didn't used it yet and it seems like any configuration should be done on tag manager side and not in plugin here. With this approach if someone starts with excluded site and navigate to not excluded site you won't get events about it. Can you blacklist paths in google tagmanager console (or whatever it is called)?

pieh:
> Is there any way to exclude a page from having the GTM snippet applied to it?

So this PR definitely does this. But because of answer below I'm not sure if doing this will show result that you would expect by doing this.

> The question is.. why it is going to exclude the events if `starts with excluded site and navigate to not excluded site`?

When Gatsby javascript bundle is loaded your page become Single Page App and when you click on links you don't load html from other pages (you load js resources). That's why you would never load GTM script if you would start with page that excludes GTM

estrada9166:
@pieh Thanks for answering!! Now I get it... but my last question (I think so...) I'm using this package (with the new changes) locally and exclude `/`... so I run `npm run build`(`gatsby build`) and then `npm run serve`(`gatsby serve`)... I visit `http://localhost:9000/` and check the source code and there is no script for GTM and then I click the link to `about` and check the source code again and the is the script for GTM....

What I'm doing wrong?

Thanks!!

pieh:
Do you use `gatsby-link` or `<a>` tags to navigate?

estrada9166:
I use `gatsby-link`

estrada9166:
@pieh you can check it [here](http://far-debt.surge.sh/) I just deploy it with surge

estrada9166:
also, I just added a `console.log()` inside the `if` and another one outside it... and this was the result when I build
```
> gatsby-starter-hello-world@ build /Users/alejandroestrada/Documents/Code/Learn/gatsby/tutorial-part-four
> gatsby build

success open and validate gatsby-configs — 0.058 s
success load plugins — 0.160 s
success onPreInit — 1.090 s
success delete html and css files from previous builds— 0.030 s
success initialize cache — 0.004 s
success copy gatsby files — 0.077 s
success onPreBootstrap — 0.008 s
success source and transform nodes — 0.062 s
success building schema — 0.197 s
success createPages — 0.038 s
success createPagesStatefully — 0.014 s
success onPreExtractQueries — 0.000 s
success update schema — 0.113 s
success extract queries from components — 0.052 s
success run graphql queries — 0.016 s — 3/3 208.86 queries/second
success write out page data — 0.004 s
success write out redirect data — 0.001 s
success onPostBootstrap — 0.001 s

info bootstrap finished - 5.639 s

success Building production JavaScript and CSS bundles— 3.698 s
⠁ Building static HTML for pagesIncluded file: -> /offline-plugin-app-shell-fallback/
⠂ Building static HTML for pagesIncluded file: -> /sweet-pandas-eating-sweets/
Included file: -> /pandas-and-bananas/
Included file: -> /about/
Excluded file: -> /
{ allFile:
   { edges:
      [ [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object] ] } }
Included file: -> /my-files/
success Building static HTML for pages — 0.645 s — 6/625.60 pages/second
Generated public/sw.js, which will precache 10 files, totaling 265448 bytes.
info Done building in 10.08 sec
```

pieh:
So I just checked your link - if you start on homepage (`/` -> excluded) and click `about` (which shouldn't be excluded), the tagmanager will not show up.

If you start on about (`/about` -> included) and click on "Pandas Eating Lots" which will navigate to homepage- tagmanager will be there and send events (if history trigger is setup in tagmanager).

So this is exactly what I mean - excluding script from rendered html is not solution here. You need to configure tag manager in google console

estrada9166:
@pieh sorry my misunderstood.. but if I start on `/` and the visit `about` and check the source code I can see the GTM script

pieh:
how are you checking source code? with devtools? or with "view source code"?

pieh:
if you check with "view source code" you will see static html - but this is only used on initial visit and we don't use it when you switch pages

estrada9166:
@pieh [here](https://www.useloom.com/share/c22ef7a8a66e4918ac012d71d597baab) is a quick video of how I'm doing it!

pieh:
Yes - you are checking `public/index.html` and `public/about/index.html` content this way and tag manager will be in `public/about/index.html` - but this is not how gatsby work when you switch pages - you need to use devtools to check actual DOM, because when you go from `/` to `/about` using gatsby link - you don't actually load `public/about/index.html` file.

estrada9166:
ohhh... @pieh I appreciate all your help and comments!! Thank you so much! I think we can close this PR and clarify on the issue that GTM should be handled on the google console instead of here!!

PR #9416, opened 2018-10-26
feat(gatsby): Add nodes db module

Moocar:
In preparation for #9338, I've added a new nodes db module that can be implemented by an appropriate db nodes backend. Which is only redux as of now. The loki PR will eventually be implemented so that it can be featured flagged on/off using this functionality.

I also added `getNodesByType` since it makes sense even without loki.


packages/gatsby/src/db/nodes.js
```diff
@@ -0,0 +1,16 @@
...
+  case `loki`:
+    nodesDb = require(`./loki/nodes`)
+    break
+  default:
+    throw new Error(
+      `Unsupported DB nodes backend (value of env var GATSBY_DB_NODES)`
+    )
+}
+
+module.exports = nodesDb
```
pieh: ❤️

pieh:
Seems like untintentional changes to `packages/gatsby-remark-embed-snippet/src/__tests__/__snapshots__/index.js.snap` were commited?

Moocar:
@pieh The `gatsby-remark-embed-snippet` test was failing on my machine. Updating the snapshot fixed it on mine, but breaks it on the CI machine. It's quite strange. I've reverted the snapshot so we can merge this. I'll try and figure out why it's breaking on my machine. From what I can see, the markup is getting ordered differently.

Moocar:
@pieh thanks for updating the package versions!

PR #9438, opened 2018-10-26
Better pull request templates

filips123:
This add different pull request templates for features and bugs.
Developers could now provide more details about PRs using some guidelines.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -12,3 +12,6 @@
 
   Learn more about contributing: https://www.gatsbyjs.org/docs/how-to-contribute/
 -->
+
+<!-- For fixing bugs use https://github.com/gatsbyjs/gatsby/compare/?template=bugs.md -->
+<!-- For pull requesting new features, improvements and changes use https://github.com/gatsbyjs/gatsby/compare/?template=features.md -->
--- a/.github/PULL_REQUEST_TEMPLATE/bugs.md
+++ b/.github/PULL_REQUEST_TEMPLATE/bugs.md
@@ -1,0 +1,35 @@
+<!--
+  Q. Which branch should I use for my pull request?
+  A. Use `master` branch (probably).
+
+  Q. Which branch if my change is a bug fix for Gatsby v1?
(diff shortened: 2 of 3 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -12,3 +12,6 @@
 
   Learn more about contributing: https://www.gatsbyjs.org/docs/how-to-contribute/
 -->
+
+<!-- For fixing bugs use https://github.com/gatsbyjs/gatsby/compare/?template=bugs.md -->
```
yeskunall: Feel like this will be missed by a lot of people 😅

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -12,3 +12,6 @@
...
+<!-- For pull requesting new features, improvements and changes use https://github.com/gatsbyjs/gatsby/compare/?template=features.md -->
```
yeskunall: Same with this one.

.github/PULL_REQUEST_TEMPLATE.md
filips123: I know... 
How can I make it more visible?

filips123: Is it possible to have multiple pull request templates with select menu like for issues?

m-allanson:
Thanks @filips123 👍 

Doing a bit of research on this, it seems like GitHub doesn't provide a way for someone to choose a 'type' of pull request, so these templates probably won't be seen by most people. 

I'm going to close this - for these templates to be useful there needs to be an easier way to view them, which GitHub doesn't provide at the moment. Maybe we can revisit this if GitHub changes how they handle PR templates?

However, if there's information that we should ask for _every_ PR, it would make sense to add that to the PR template. e.g. maybe it's useful to add _Description_ and _Related issue_ headings to the default template?

Would you like to open an issue to discuss making changes to the default PR template?

filips123:
I created issue #10470 for that.

I can edit my PR to make them more visible, maybe using header or what do you think it will be the best, and you can then merge it. Unfortunately, I can't make more thing until GitHub supports that.

PR #9441, opened 2018-10-26
Add file name emphasis to docs

### LINKED ISSUE #8393: Implement file name emphasis in docs
## Summary

At the top of code snippets, the file name where they ought to be pasted should be emphasized.

### Basic example

Here is the PR that implemented the functionality: #7867
Now we just need to go through the docs and the tutorial (and possibly official plugin and starter READmes?) and use the functionality wherever there is a code snippet that is meant to be pasted into a specific file.

### Motivation

In a workshop last May, I noticed that many people paste code snippets into the incorrect file. This way of emphasizing file names will hopefully help everyone paste code into the correct place.

### Next steps
If anyone wants to also tackle #5030, that would be an ideal companion issue to tackle along with this one.

Utwo:
Resolve #8393

### DIFF
--- a/docs/docs/add-a-service-worker.md
+++ b/docs/docs/add-a-service-worker.md
@@ -20,8 +20,7 @@ We recommend using this plugin together with the [manifest plugin](https://www.n
 
 Add this plugin to your `gatsby-config.js`
 
-```javascript
-// In your gatsby-config.js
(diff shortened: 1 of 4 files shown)


shannonbux:
Thank you @utwo! Super valuable :)

PR #9445, opened 2018-10-26
Update add-a-manifest-file.md

dhruvikn:


yeskunall:
@shannonbux I think the intended change was "Check" to "check", but the removal of one dash (-) was probably accidental 😅

Issue #4021, opened 2018-02-13
gatsby-image browser support


TuckerWhitehouse:
## Description

`gatsby-image` uses `object-fit/object-position` which are [not supported in IE](https://caniuse.com/#feat=object-fit).
https://www.gatsbyjs.org/docs/browser-support/ states support for IE 9+.
Some discussion about the issue and a workaround can be seen here: https://github.com/gatsbyjs/gatsby/issues/2470#issuecomment-338394370

### Expected behavior

a) Either a note on `gatsby-image` that includes a "blessed" solution for IE,
b) an `object-fit/object-position` polyfill in `gatsby-image`.

I'd be happy to open a pull request for either of these options (not sure what including the polyfill directly would look like), but maybe adding the note about [object-fit-images](https://npm.im/object-fit-images) would be enough?

KyleAMathews:
Oh hmmm yeah — we should definitely polyfill gatsby-image for IE. object-fit-images looks great from a quick look.

thebarty:
plus one - got the same problem

ooloth:
Hey all! I also noticed this issue. Here's how you can fix it:

```sh
npm i intersection-observer object-fit-images
```

```js
// gatsby-browser.js

exports.onClientEntry = () => {
  // NOTE: Don't polyfill Promise here (Gatsby already includes a Promise polyfill)

  // IntersectionObserver polyfill for gatsby-image (Safari, IE)
  if (typeof window.IntersectionObserver === `undefined`) {
    require(`intersection-observer`)
    console.log(`👍 IntersectionObserver is polyfilled`)
  }

  // Object-fit/Object-position polyfill for gatsby-image (IE)
  const testImg = document.createElement(`img`)
  if (
    typeof testImg.style.objectFit === `undefined` ||
    typeof testImg.style.objectPosition === `undefined`
  ) {
    require(`object-fit-images`)()
    console.log(`👍 Object-fit/Object-position are polyfilled`)
  }
}

```

The `intersection-observer` polyfill works out of the box.

However, the 'object-fit-images` polyfill requires that you add the following CSS to your image's styles:

```css
font-family: "object-fit: VALUE_HERE; object-position: VALUE_HERE"
```

You can either add this font-family declaration manually to each image, or create a wrapper for `gatsby-image` like the following:

```js

// Img.js (a gatsby-image wrapper)

const Img = props => {
  // Construct font-family declaration for object-fit-images
[...]

PR #9451, opened 2018-10-26
Dynamically add IO and object-fit/object-position polyfills to gatsby-image (only when needed)

ooloth:
This PR addresses [#4021](https://github.com/gatsbyjs/gatsby/issues/4021) by:

1. **Removed:** ~~Dynamically implementing the `intersection-observer` polyfill to enable lazy loading (only in browsers that need it, currently Safari and IE)~~
2. Dynamically implementing the `object-fit-images` polyfill to enable the `object-fit` and `object-position` CSS properties (only in browsers that need it, currently IE)

It also updates the `gatsby-image` README with the following changes in usage:

1. **Edited:** These new polyfills are optional. To enable them, `gatsby-image` must be added to `gatsby-config` with the `loadPolyfills` option set to `true`. This will dynamically import the polyfills in browsers that need them. (If no polyfills are wanted, `gatsby-image` can still be used as before.)
2. **Removed:** ~~The new `objectFit` and `objectPosition` props must be used to pass these CSS declarations to the `` element  (instead of `imgStyle`) or the `object-fit-images` polyfill won't work. This is necessary because of the way `object-fit-images` duplicates these property declarations in a custom `font-family` declaration that triggers the polyfill.~~

Let me know if I missed anything!


ooloth:
Hmm...sorry about this: a lot of my local changes to `index.js` are missing from this commit...

I'll try to update before anyone spends time reviewing!

ooloth:
Ah, I see! I mistakenly updated `packages/gatsby-image/index.js` but not `packages/gatsby-image/src/index.js`.

Can anyone advise which of the following is required?

1. Update both files
2. Update only `src/index.js`

I'll fix the PR from there.

ooloth:
Any advice about how to clear these failing tests? Some appear to be unrelated to the gatsby-image files I've altered and a couple are snapshots I'm not sure how to fix...

Any tips would be appreciated!

KyleAMathews:
How big are the polyfills?

ooloth:
Let me know if you'd be more comfortable if I released a separate plugin with this functionality (e.g. `gatsby-image-with-polyfills`) rather than baking these polyfills directly into `gatsby-image`.

I think the community would find it simpler to have the polyfills baked in (since most devs will notice the Safari IO bug eventually). However, if you'd rather keep the file size minimal and the API and docs uncluttered by polyfills, I totally get it! 

I'm happy to release the fix to the community in whichever form you prefer.

packages/gatsby-image/README.md
```diff
@@ -60,7 +60,11 @@ effect as well as lazy loading of images further down the screen.
 
 ## Install
 
-`npm install --save gatsby-image`
+`npm install --save gatsby-image intersection-observer object-fit-images`
```
pieh: if those are in `dependencies` in package.json people don't need to install them manually

packages/gatsby-image/gatsby-browser.js
```diff
@@ -0,0 +1,17 @@
...
+    import(`intersection-observer`)
+  }
+
+  // Object-fit/Object-position polyfill for gatsby-image (IE)
+  const testImg = document.createElement(`img`)
+  if (
+    typeof testImg.style.objectFit === `undefined` ||
+    typeof testImg.style.objectPosition === `undefined`
+  ) {
+    import(`object-fit-images`)()
```
pieh: I don't think this is correct? Shouldn't it be something like:
```js
import(`object-fit-images`).then(ObjectFitImages => ObjectFitImages())
```
?

packages/gatsby-image/src/index.js
```diff
@@ -74,21 +74,31 @@ const noscriptImg = props => {
...
   const alt = props.alt ? `alt="${props.alt}" ` : `alt="" ` // required attribute
   const width = props.width ? `width="${props.width}" ` : ``
   const height = props.height ? `height="${props.height}" ` : ``
   const opacity = props.opacity ? props.opacity : `1`
   const transitionDelay = props.transitionDelay ? props.transitionDelay : `0.5s`
-  return (`<picture>${srcSetWebp}${srcSet}<img ${width}${height}${src}${alt}${title}style="position:absolute;top:0;left:0;transition:opacity 0.5s;transition-delay:${transitionDelay};opacity:${opacity};width:100%;height:100%;object-fit:cover;object-position:center"/></picture>`)
+
+  // needed for object-fit-images polyfill:
+  var objectFit = props.objectFit ? props.objectFit : `cover`
+  var objectPosition = props.objectPosition ? props.objectPosition : `50% 50%`
```
pieh: default props should handle fallbacks, also please don't use vars

ooloth: No problem! I was just mirroring the code already in that area. Do you want me to refactor this whole section to use default props and remove vars?

packages/gatsby-image/gatsby-browser.js
ooloth: On my projects, it works as written, but I'm happy to implement the rewrite if it's more solid.

packages/gatsby-image/README.md
ooloth: Excellent! I wasn't sure.

packages/gatsby-image/gatsby-browser.js
ooloth: > But with IntersectionObserver there might be timing issue - https://www.npmjs.com/package/intersection-observer: Be sure to include the polyfill prior to referencing it anywhere in your JavaScript code we might try to reference it earlier.

It's been working for me to load the IO polyfill in `gatsby-browser, but if you know of an earlier place we can load it, that would be great!

ooloth:
@pieh I'm totally onboard with adding a `loadPolyfills` option that lets users opt in/out of adding these polyfills. I'm just not sure how to access the value of this option in `src/index.js` so I can turn the `font-family` declaration on/off... 

I've checked out the [plugin authoring docs](https://www.gatsbyjs.org/docs/plugins/) and can't seem to find the answer. Can you point me in the right direction?

packages/gatsby-image/README.md
```diff
@@ -274,6 +280,8 @@ prop. e.g. `<Img fluid={fluid} />`
 | `className`        | `string` / `object` | Passed to the wrapper element. Object is needed to support Glamor's css prop                                                |
 | `style`            | `object`            | Spread into the default styles of the wrapper element                                                                       |
 | `imgStyle`         | `object`            | Spread into the default styles of the actual `img` element                                                                  |
+| `objectFit`        | `string`            | Spread into the default styles of the actual `img` element in a way that polyfills the behavior for IE. Defaults to `cover`. |
+| `objectPosition`   | `string`            | Spread into the default styles of the actual `img` element in a way that polyfills the behavior for IE. Defaults to `50% 50%`. |
```
pieh: Do we need to mention IE here? having those options actually allow non-IE browsers to set those css properties, correct?

packages/gatsby-image/src/index.js
```diff
@@ -197,26 +211,35 @@ class Image extends React.Component {
...
       ...imgStyle,
       ...placeholderStyle,
     }
 
     const imageStyle = {
       opacity: this.state.imgLoaded || this.state.fadeIn === false ? 1 : 0,
       transition: this.state.fadeIn === true ? `opacity 0.5s` : `none`,
+      objectFit: objectFit,
+      objectPosition: objectPosition,
+      fontFamily: `"object-fit: ${objectFit}; object-position: ${objectPosition}"`,
```
pieh: because this is duplicated for both `imagePlaceholderStyle` and `imageStyle`, we should just add them to `imgStyle` object:

`imgStyle.objectFit = objectFit` before creating `imagePlaceholderStyle` object

packages/gatsby-image/src/index.js
```diff
@@ -464,6 +487,8 @@ Image.propTypes = {
   critical: PropTypes.bool,
   style: PropTypes.object,
   imgStyle: PropTypes.object,
+  objectFit: PropTypes.string, // needs to be separate for object-fit-image polyfill's font-family declaration
+  objectPosition: PropTypes.string, // needs to be separate for object-fit-image polyfill's font-family declaration
```
pieh: I don't think that comment is needed here as well - this is not only for polyfills

pieh:
@ooloth about `loadPolyfills` option - we would need to create `src/gatsby-node.js` file with something like:
```js
exports.onCreateWebpackConfig = ({ actions, plugins }, { loadPolyfills = false }) => {
  actions.setWebpackConfig({
    plugins: [
      plugins.define({
        GATSBY_IMAGE_LOAD_POLYFILLS: JSON.stringify(loadPolyfills),
      }),
    ],
  })
}
```

and later in `gatsby-browser.js` and `index.js` use:
```js
if (process.env.GATSBY_IMAGE_LOAD_POLYFILLS) {
  // do things if we need to load and use polyfills
}
```

packages/gatsby-image/README.md
ooloth: Totally! I’ll remove the specific references to IE.

packages/gatsby-image/src/index.js
ooloth: Nice! Will do.

packages/gatsby-image/src/index.js
ooloth: I’m wondering if devs might think they can just pass these properties to imgStyle (which they can, but then the polyfill will silently fail)...

Any tips for making it clear that the obj declarations need to be passed outside of imgStyle to enable polyfilling?

ooloth:
@pieh Thank you for the plugin-option-passing snippets!

I’ll update the PR based on these notes tomorrow afternoon.

ooloth:
@pieh Any idea why `process.env.GATSBY_IMAGE_LOAD_POLYFILLS` would be `undefined` using the code snippets above?

I'm able to log the value of `loadPolyfills` in `gatsby-node`, but I'm not able to retrieve it from `process.env.GATSBY_IMAGE_LOAD_POLYFILLS` in `gatsby-browser` or `src/index` (it logs as `undefined`).

packages/gatsby-image/src/index.js
ooloth: I reworded the comment to make it clearer to future devs why these style properties are passed separately. Let me know if you still think it should be removed, though, and I'll be happy to remove it.

pieh: oh, right - this can be extracted from `imgStyle` - we could/should just reuse that if user decides to overwrite it

pieh: and use values from there (if set) for `font-family` weird IE hack :)

pieh:
I will take a look - let me know if you are good with me pushing to your branch? We also need to resolve some conflicts it seems

pieh:
Ah, it should be:
```diff
exports.onCreateWebpackConfig = ({ actions, plugins }, { loadPolyfills = false }) => {
  actions.setWebpackConfig({
    plugins: [
      plugins.define({
-       GATSBY_IMAGE_LOAD_POLYFILLS: JSON.stringify(loadPolyfills),
+       "process.env.GATSBY_IMAGE_LOAD_POLYFILLS": JSON.stringify(loadPolyfills),
      }),
    ],
  })
}
```

ooloth:
This seems to be working as expected now:

1. `gatsby-image` still works as before without adding it to `gatsby-config` (no polyfills)
2. It works the same way when adding it to `gatsby-config` with no options set (no polyfills)
3. To enable polyfills, add `gatsby-image` to `gatsby-config` and set `loadPolyfills` to `true`.

pieh:
> * `gatsby-image` still works as before without adding it to `gatsby-config` (no polyfills)
> * It works the same way when adding it to `gatsby-config` with no options set (no polyfills)

The end result seems the same and work the same but there is subtle difference:
if we add `gatsby-image` to `gatsby-config` and `DefinePlugin` is used - webpack will actually remove parts of the code that are inside the if statements (so slightly slower bundle size) - see notes in `Usage` section of DefinePlugin docs - https://webpack.js.org/plugins/define-plugin/#usage. But this isn't that big of a deal because without adding it to `gatsby-config`, `gatsby-browser` isn't included in bundle, only additions in `src/index.js` would be added to bundle size which are very small :)

ooloth:
Ah I see! 

Should the README be updated to consistently recommend adding `gatsby-image` to `gatsby-config`? 

For completeness, I mentioned it can be used without this step, but since that unnecessarily enlarges the bundle size, I'd be happy to remove it.

packages/gatsby-image/src/index.js
```diff
@@ -74,40 +74,65 @@ const noscriptImg = props => {
...
+
+  // needed for object-fit-images polyfill:
+  const objectFit = props.objectFit
+  const objectPosition = props.objectPosition
+
+  if (process.env.GATSBY_IMAGE_LOAD_POLYFILLS) {
+    return `<picture>${srcSetWebp}${srcSet}<img ${width}${height}${src}${alt}${title}style="position:absolute;top:0;left:0;transition:opacity 0.5s;transition-delay:${transitionDelay};opacity:${opacity};width:100%;height:100%;object-fit:${objectFit};object-position:${objectPosition};font-family:'object-fit:${objectFit};object-position:${objectPosition};'"/></picture>`
+  } else {
+    return `<picture>${srcSetWebp}${srcSet}<img ${width}${height}${src}${alt}${title}style="position:absolute;top:0;left:0;transition:opacity 0.5s;transition-delay:${transitionDelay};opacity:${opacity};width:100%;height:100%;object-fit:${objectFit};object-position:${objectPosition};"/></picture>`
+  }
```
pieh: I think we don't need to repeat almost same thing 2 times. How about this:

```diff
-  if (process.env.GATSBY_IMAGE_LOAD_POLYFILLS) {
-    return `<picture>${srcSetWebp}${srcSet}<img ${width}${height}${src}${alt}${title}style="position:absolute;top:0;left:0;transition:opacity 0.5s;transition-delay:${transitionDelay};opacity:${opacity};width:100%;height:100%;object-fit:${objectFit};object-position:${objectPosition};font-family:'object-fit:${objectFit};object-position:${objectPosition};'"/></picture>`
-  } else {
-    return `<picture>${srcSetWebp}${srcSet}<img ${width}${height}${src}${alt}${title}style="position:absolute;top:0;left:0;transition:opacity 0.5s;transition-delay:${transitionDelay};opacity:${opacity};width:100%;height:100%;object-fit:${objectFit};object-position:${objectPosition};"/></picture>`
-  }
+  let styleStr = `position:absolute;top:0;left:0;transition:opacity 0.5s;transition-delay:${transitionDelay};opacity:${opacity};width:100%;height:100%;object-fit:${objectFit};object-position:${objectPosition};`
+  if (process.env.GATSBY_IMAGE_LOAD_POLYFILLS) {
+    styleStr += `font-family:'object-fit:${objectFit};object-position:${objectPosition};';`
+  }
+  return `<picture>${srcSetWebp}${srcSet}<img ${width}${height}${src}${alt}${title}style="${styleStr}"/></picture>`
```

pieh:
> For completeness, I mentioned it can be used without this step, but since that unnecessarily enlarges the bundle size, I'd be happy to remove it.

I think this is fine as is, but I'll get more eyes on this ;) @DSchau what do you think on approach here?

ooloth:
A thought - since we are now capturing the `object-fit/position` values from the `imgStyle` prop, should I remove the new `objectFit` and `objectPosition` props? 

At first, I thought I needed them to capture these values for the polyfill, but I'm realizing that just using `imgStyle` is a simpler API and would work just as well.

sebastienfi:
> A thought - since we are now capturing the object-fit/position values from the imgStyle prop, should I remove the new objectFit and objectPosition props?

I think so 👍

packages/gatsby-image/src/index.js
ooloth: Love it.

pieh:
I thought I replied, sorry @ooloth (I want reply via email for github notifications ever again!)

And yeah, lets avoid duplication of props.

ooloth:
@pieh No worries! Let me know if anything else needs to be tweaked.

ooloth:
Btw, I'm not sure how to fix the unsuccessful checks blocking this merge...

If anyone else knows what to do, please let me know.

packages/gatsby-image/README.md
```diff
@@ -94,6 +94,34 @@ module.exports = {
 }
 ```
 
+### Dynamically including polyfills
+
+You can use `gatsby-image` without adding it to `gatsby-config.js` (it Just Works™). However, lazy loading (e.g. blur-up) may not work in browsers lacking support for `IntersectionObserver` (e.g. Safari and IE) and `object-fit`/`object-position` may not work in browsers lacking support for these CSS properties (e.g. IE). 
+
+To support these features in all browsers, add `gatsby-image` to `gatsby-config.js` and set the `loadPolyfills` option to true. This tells `gatsby-image` to dynamically import the [intersection-observer](https://github.com/w3c/IntersectionObserver#readme) and [object-fit-images](https://github.com/bfred-it/object-fit-images) polyfills in browsers that need them. (These polyfills will not be loaded in browsers that already support these features.)
```
DSchau: What do you think of _just_ loading object-fit-images here? It seems like intersection-observer is more "app level" polyfill whereas object-fit-images is actually needed and makes a ton of sense for an image component to provide opt-in support for that polyfill--additionally because this component _needs_ something extra to make the polyfill useful.

Primarily I mention this because gatsby-link also uses an intersection observer, and we sorta treat that as "progressively enhanced," i.e. pages are prefetched in certain conditions _and_ if the API is available. I think doing something similar here makes sense, but I'm open to your opinion here!

packages/gatsby-image/gatsby-browser.js
```diff
@@ -0,0 +1,19 @@
+"use strict"
+
+exports.onClientEntry = () => {
+  if (process.env.GATSBY_IMAGE_LOAD_POLYFILLS) {
+    // IntersectionObserver polyfill for gatsby-image (Safari, IE)
+    if (typeof window.IntersectionObserver === `undefined`) {
```
DSchau: I think we should remove this in favor of documenting _how_ to polyfill app-level features like this.

packages/gatsby-image/src/__tests__/__snapshots__/index.js.snap
```diff
@@ -29,13 +29,13 @@ exports[`<Img /> should render fixed size images 1`] = `
         alt="Alt text for the image"
         height="100"
         src="test_image.jpg"
-        style="position: absolute; top: 0px; left: 0px; width: 100%; height: 100%; opacity: 1;"
+        style="position: absolute; top: 0px; left: 0px; width: 100%; height: 100%; font-family: "object-fit: cover; object-position: 50% 50%"; opacity: 1;"
```
DSchau: Show this weirdness (necessary for the polyfill, I know!) make its way into the unit tests? I wouldn't have thought the opt-in behavior would have impacted this.

ooloth: That's my bad! I don't know anything about testing, but saw that tests were failing, so fumbled around a bit trying to make the markup match the updated version.

I'm totally happy to drop these changes and I defer to you about how to handle the test failures below. 🤷‍♂️

pieh: you can run `yarn jest -u` to update snapshots in your repository and commit those changes

packages/gatsby-image/gatsby-browser.js
ooloth: Sure! I get what you mean: IntersectionObserver has a number of use cases besides `gatsby-image`, so better to make an app-wide decision about whether to support it. 

Nothing breaks if Safari doesn't blur up nicely.

packages/gatsby-image/README.md
ooloth: Sounds good!

I agree: object-fit/position really *only* affects images, so it makes sense to tie that polyfill to `gatsby-image` (since it's the only component likely to use it). It also affects the alignment of the image, which is crucial in some cases. 

On the other hand, IO is app-level and only affects the `gatsby-image` loading effect, which I agree is a nice-to-have.

Should we explain how to dynamically polyfill IO via `gatsby-browser` in this part of the README? I'm torn since I don't want the preamble to get too long, but I also remember how long it took me to notice, research and solve the `gatsby-image` polyfilling issues, so it would be nice to spare future devs that work.

DSchau: Those are are all really good points! I think what makes the most sense is a doc on polyfilling + gatsby, i.e. how to dynamically load polyfills based on certain conditions. It could even cover using something like [polyfill.io](https://polyfill.io/v2/docs/).

So... short answer: I think we take out the polyfill for now, and then link to a stub, perhaps? I like the idea of linking to more general purpose, reusable documentation rather than re-inventing the wheel as far as documenting things goes!

packages/gatsby-image/src/__tests__/__snapshots__/index.js.snap
ooloth: Thank you @pieh!

packages/gatsby-image/gatsby-browser.js
```diff
@@ -0,0 +1,14 @@
+"use strict"
```
DSchau: Whoops - forgot to comment on this last time. Should this file be in src instead? Or not committed to version control?

ooloth: I'm not sure how it works, actually! This is my first time working on a plugin...

It contains the code that imports the polyfill, so would it still work without committing it? I'm happy to move it or change whatever you recommend. 👍

packages/gatsby-image/README.md
```diff
@@ -94,6 +94,34 @@ module.exports = {
 }
 ```
 
+### Dynamically polyfilling object-fit/object-position
```
wardpeet: `### Add support for object-fit/object-position to older browsers`
might be a better title

packages/gatsby-image/gatsby-browser.js
```diff
@@ -0,0 +1,14 @@
...
+
+exports.onClientEntry = () => {
+  if (process.env.GATSBY_IMAGE_LOAD_POLYFILLS) {
+    // Add object-fit/position polyfill if browser needs it
+    const testImg = document.createElement(`img`)
+    if (
+      typeof testImg.style.objectFit === `undefined` ||
+      typeof testImg.style.objectPosition === `undefined`
+    ) {
+      import(`object-fit-images`).then(ObjectFitImages => ObjectFitImages())
```
JustinTRoss: This line results in an error for me. I'm not sure the promise understands to use `default` for objects resolved from modules. The following should work:

```
import('object-fit-images').then(({ default: ObjectFitImages }) =>
    ObjectFitImages()
);
```

ooloth:
I'm closing this PR in favour of #12681, which resulted from further discussions in #4021.

If anyone has questions or concerns about this, please let me know.

PR #9237, opened 2018-10-19
feat(gatsby-plugin-less): add support for overriding the default options of `css-loader`

yeskunall:
Yay! 💖

<hr />
Closes #9142.

Issue #5458, opened 2018-05-17
[v2] css modules class names with dashes / camelCased locals


pieh:
Currently in v2 css modules there is undocumented breaking change - dashed class names will be camelCased ( https://github.com/gatsbyjs/gatsby/blob/c0e503d0df7cefb3b072e85155e2ce853e48fb53/packages/gatsby/src/utils/webpack-utils.js#L201 )

Omitting that setting or setting it to false will result in same behaviour as in v1. Setting it to true or 'dashes' would return both dashed and camelCased version of it - might be worth using that?

jquense:
`dashes` is the most accepting setting, but it leads to larger bundles since most css classes are duplicated in the locals export (the camel and dash cases). I went with the `dashesOnly` setting a bit out of personal convenience, but i do think it's probably the community wide "best practice" (for lack of a better word) and fairly uncontroversial, saving on bytes while also allowing idiomatic casing for css and js

pieh:
It makes sense - in this case we would "just" need to update https://github.com/gatsbyjs/gatsby/blob/v2/Breaking%20Changes.md so it's documented. 

I'll leave this issue opened just to get more input on the matter from other folks.

pieh:
I was just thinking about it - would it make sense to revert to v1 default and create plugin to easily overwrite css module rule options? Something that walks webpack config object rules and if loader is `css-loader` and `options.modules === true` replace/merge options with what user specify?

m-allanson:
Maybe let's go the other way around? `dashesOnly` seems like a good option and v2 is the time to make breaking changes. I'll add a section to the migration guide.

dustinhorton:
@pieh did you happen to write something to allow for v1 behavior? have multiple projects where this change is causing some pain.

HamidReazaNikoonia:
what about pt-5 , pt5 not working

dustinhorton:
@HamidReazaNikoonia what part 5?

acalderongarcia:
> @HamidReazaNikoonia what part 5?

pt-5 padding-top

dustinhorton:
@acalderongarcia have you added this to your `gatsby-config.js`?

```javascript
module.exports = {
  plugins: [
    {
      resolve: 'gatsby-plugin-sass',
      options: {
        cssLoaderOptions: {
          camelCase: false,
        },
      },
    },
  ],
}
```

PR #9462, opened 2018-10-27
feat(gatsby-plugin-sass): Accept css-loader options

dustinhorton:
v2 changed how dashes/hyphens in class names handled by css modules were treated: https://github.com/gatsbyjs/gatsby/issues/5458. This was burdensome as it required changing all `className`s that utilized dashes.

This allows overriding that change, almost entirely by copying prior art from @yeskunall & @pieh (https://github.com/gatsbyjs/gatsby/pull/9237).

### DIFF
--- a/docs/docs/migrating-from-v1-to-v2.md
+++ b/docs/docs/migrating-from-v1-to-v2.md
@@ -674,7 +674,9 @@ export default ({ children }) => (
 )
 ```
 
-To restore the default from Gatsby v1, see the [plugin docs](/packages/gatsby-plugin-sass/README.md#how-to-use).
+The Gatsby v1 behavior can be restored by passing in [CSS Loader options](https://github.com/webpack-contrib/css-loader#options) when configuring [gatsby-plugin-sass](/packages/gatsby-plugin-sass/README.md#how-to-use) or [gatsby-plugin-less](/packages/gatsby-plugin-less/README.md#how-to-use).
+
+To restore the default from Gatsby v1 when using, see the [plugin docs](/packages/gatsby-plugin-sass/README.md#how-to-use).
 
 ### Update Jest configuration
 
--- a/docs/docs/migrating-from-v1-to-v2.md
+++ b/docs/docs/migrating-from-v1-to-v2.md
@@ -674,9 +674,78 @@ export default ({ children }) => (
 )
 ```
 
-The Gatsby v1 behavior can be restored by passing in [CSS Loader options](https://github.com/webpack-contrib/css-loader#options) when configuring [gatsby-plugin-sass](/packages/gatsby-plugin-sass/README.md#how-to-use) or [gatsby-plugin-less](/packages/gatsby-plugin-less/README.md#how-to-use).
+The Gatsby v1 behavior can be restored by adjusting [CSS Loader options](https://github.com/webpack-contrib/css-loader#options).
 
-To restore the default from Gatsby v1 when using, see the [plugin docs](/packages/gatsby-plugin-sass/README.md#how-to-use).
+For vanilla CSS without a preprocessor:
+
+```javascript
+const cssLoaderRe = /\/css-loader\//
+const targetFile = `.module.css`
+
+const processRule = rule => {
+  if (rule.oneOf) {
+    return {
+      ...rule,
+      oneOf: rule.oneOf.map(processRule),
+    }
+  }
+
+  if (!rule.test.test(targetFile)) {
+    return rule
+  }
+
+  if (Array.isArray(rule.use)) {
+    return {
+      ...rule,
+      use: rule.use.map(use => {
+        if (!cssLoaderRe.test(use.loader)) {
+          return use
+        }
+
+        // adjust css-loader options
+        return {
+          ...use,
+          options: {
+            ...use.options,
+            camelCase: false,
+          },
+        }
+      }),
+    }
+  }
+
+  return rule
+}
+
+exports.onCreateWebpackConfig = ({ getConfig, actions }) => {
+  const config = getConfig()
+
+  const newConfig = {
+    ...config,
+    module: {
+      ...config.module,
+      rules: config.module.rules.map(processRule),
+    },
+  }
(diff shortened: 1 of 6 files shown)


yeskunall:
I'm sorry, but other than README changes, I don't see why you have the same stuff incorporated from #9237.

docs/docs/migrating-from-v1-to-v2.md
```diff
@@ -674,6 +674,8 @@ export default ({ children }) => (
 )
 ```
 
+To restore the default from Gatsby v1, see the [plugin docs](/packages/gatsby-plugin-sass/README.md#how-to-use).
```
pieh: In this section there is talk about css modules (not sass specific version of it) - we should make it clear. Like if you are using less - see less plugin docs, if sass - see sass plugins docs. I will prepare code snippet for restoring v1 default for vanilla css modules

dustinhorton: Good call—updated (will need to tweak the doc further when you make it possible with vanilla).

dustinhorton:
@pieh Need anything else from me?

packages/gatsby-plugin-less/README.md
```diff
@@ -47,6 +47,8 @@ plugins: [
 ]
 ```
 
+*This particular change would restore Gatsby v1 CSS modules behavior.*
```
DSchau: Do you think this is important to note?

packages/gatsby-plugin-sass/src/__tests__/__snapshots__/gatsby-node.js.snap
```diff
@@ -12,29 +12,20 @@ exports[`gatsby-plugin-sass Stage: build-html / No options 1`] = `
...
                   "test": /\\\\\\.module\\\\\\.s\\(a\\|c\\)ss\\$/,
                   "use": Array [
-                    "miniCssExtract",
                     "css({\\"modules\\":true,\\"importLoaders\\":2})",
                     "postcss({})",
                     Object {
                       "loader": "/resolved/path/sass-loader",
                       "options": Object {
-                        "sourceMap": true,
+                        "sourceMap": false,
```
DSchau: Should this be false?

packages/gatsby-plugin-sass/src/__tests__/__snapshots__/gatsby-node.js.snap
```diff
@@ -12,29 +12,20 @@ exports[`gatsby-plugin-sass Stage: build-html / No options 1`] = `
                 Object {
                   "test": /\\\\\\.module\\\\\\.s\\(a\\|c\\)ss\\$/,
                   "use": Array [
-                    "miniCssExtract",
```
DSchau: This seems kinda concerning.. ? I'm not 100% sure these snapshots are completely useful, but it seems like they're catching something valuable here!

packages/gatsby-plugin-less/README.md
DSchau: (and it's becoming more and more unimportant as more people use Gatsby v2!; so I'm inclined to say just scrap this line!)

packages/gatsby-plugin-sass/src/__tests__/__snapshots__/gatsby-node.js.snap
pieh: it is correct - this is for `build-html` stage where we don't extract css

packages/gatsby-plugin-sass/src/__tests__/__snapshots__/gatsby-node.js.snap
pieh: this comes from gatsby core - where css sourceMaps are being produced only for `develop` (https://github.com/gatsbyjs/gatsby/blob/3dbec2c11dfd98c5a949a40ee193195785d7d3f3/packages/gatsby/src/utils/webpack-utils.js#L195), so if this should be changed it should be in separate PR

packages/gatsby-plugin-less/README.md
dustinhorton: Happy to scrap. If so, I should probably elaborate within get migration doc.

docs/docs/migrating-from-v1-to-v2.md
pieh: So here's snippet that can be used for vanilla css modules in gatsby-node.js file that restores v1 behaviour:

```js
const cssLoaderRe = /\/css-loader\//
const targetFile = `.module.css`

const processRule = rule => {
  if (rule.oneOf) {
    return {
      ...rule,
      oneOf: rule.oneOf.map(processRule),
    }
  }

  if (!rule.test.test(targetFile)) {
    return rule
  }

  if (Array.isArray(rule.use)) {
    return {
      ...rule,
      use: rule.use.map(use => {
        if (!cssLoaderRe.test(use.loader)) {
          return use
        }

        // adjust css-loader options
        return {
          ...use,
          options: {
            ...use.options,
            camelCase: false,
          },
        }
      }),
    }
  }

  return rule
}

exports.onCreateWebpackConfig = ({ getConfig, actions }) => {
  const config = getConfig()

  const newConfig = {
    ...config,
    module: {
      ...config.module,
      rules: config.module.rules.map(processRule),
    },
  }
  actions.replaceWebpackConfig(newConfig)
}
```

packages/gatsby-plugin-less/README.md
pieh: Yup, I'd move this to migration guide

docs/docs/migrating-from-v1-to-v2.md
```diff
@@ -674,6 +674,79 @@ export default ({ children }) => (
 )
 ```
 
+The Gatsby v1 behavior can be restored by adjusting [CSS Loader options](https://github.com/webpack-contrib/css-loader#options).
+
+For vanilla CSS without a preprocessor:
+
+```javascript
```
pieh:
```suggestion
```javascript:title=gatsby-node.js
```
this will show title bar indicating where to put this

PR #9469, opened 2018-10-27
Search

mssha:
sitemap.xml is not only a way of displaying all the related urls of a website


pieh:
I'm not sure if information like this should live in example site README. This would be more useful in general docs and also more details would be required

DSchau:
Yeah, as @pieh said I don't think this is a change that improves the documentation here, unfortunately, but we really do appreciate the PR!

We absolutely want to have you as a contributor, so please take a look at [our open issues][open-issues] for ideas, and please reach out to us on Twitter at @gatsbyjs with questions.

Issue #9476, opened 2018-10-27
[CI] Circle CI not triggering unit tests if last commit does not modifies `packages/*|.circleci/*`


phacks:
## Description

A Pull Request may contain several commits, with some of them modifying files in `packages/` or `.circleci/`, and the last one modifying e.g. `examples/` or the root `README.md`.

Such a Pull Request should trigger unit tests (because it modifies a file in `packages/` or `.circleci/`, per #8269 by @DSchau), but it currently does not.

### Steps to reproduce

I cooked up a sample Pull Request to verify the problem at #9475.

### Expected result

Unit tests should be ran on CircleCI.

### Actual result

Unit tests did not run on CircleCI.

### Environment

Problem does not appear locally, only on Circle CI.

phacks:
It looks like replacing

```
FILES_COUNT="$(git diff-tree --no-commit-id --name-only -r $CIRCLE_SHA1 | grep -E "$GREP_PATTERN" | wc -l)"
```

by

```
FILES_COUNT="$(git diff-tree --no-commit-id --name-only -r $CIRCLE_BRANCH master | grep -E "$GREP_PATTERN" | wc -l)"
```

might do the trick.

However, it will most probably trigger unit tests for any Pull Requests against `v1`, independently of the contents of the Pull Request. I was not able to find a Circle env var for “the branch you are pull requesting against”…

What do you all think?

phacks:
If you wish to go with that solution, I opened #9478 that can be merged :)

PR #9475, opened 2018-10-27
Test circle ci unit tests triggering correctly

phacks:
I’m having doubts about the `assert-changed-files` script triggering (or not) unit tests in CircleCI:

Specifically, this line handles the “check” whether a file in `packages/` or in `circleci/` was modified in the Pull Request:

```
FILES_COUNT="$(git diff-tree --no-commit-id --name-only -r $CIRCLE_SHA1 | grep -E "$GREP_PATTERN" | wc -l)"
```

I wonder whether a PR in the following form would trigger the unit tests:

```
commit1 <- changes a file in `packages/`
commit2 <- changes the root `README.md` file
```

This PR has no other point other than testing this hypothesis.

PR #9477, opened 2018-10-27
Update circle ci unit test trigger

phacks:
Follow-up on #9475 and #9476, trying a new trigger that would take in account the whole branch in the Pull Request, instead of the last commit only.

If unit tests are actually ran, it means that the script update worked 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,8 @@
 
 Gatsby is a modern framework for blazing fast websites.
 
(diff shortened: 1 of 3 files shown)


phacks:
Aaaand tests were not run :(

phacks:
Closing the Pull Request as it was only for testing purposes and should not be merged.

PR #9478, opened 2018-10-27
Trigger CI unit tests if *any* commit changes `packages/` or `.circleci/`

### LINKED ISSUE #9476: [CI] Circle CI not triggering unit tests if last commit does not modifies `packages/*|.circleci/*`

phacks:
Trigger CI unit tests if *any* commit changes `packages/` or `.circleci/`, currently unit tests are triggered only if the last commit changes those folders.

Fixes #9476.

### DIFF
--- a/scripts/assert-changed-files.sh
+++ b/scripts/assert-changed-files.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 GREP_PATTERN=$1
 
-FILES_COUNT="$(git diff-tree --no-commit-id --name-only -r $CIRCLE_SHA1 | grep -E "$GREP_PATTERN" | wc -l)"
+FILES_COUNT="$(git diff-tree --no-commit-id --name-only -r $CIRCLE_BRANCH origin/master | grep -E "$GREP_PATTERN" | wc -l)"
 
 if [ $FILES_COUNT -eq 0 ]; then
   echo "0 files matching '$GREP_PATTERN'; exiting and marking successful."


DSchau:
@pieh you think we should just merge this and try it out? We can always revert if we're hitting upper limits of CI time.

phacks:
@DSchau Thanks for explaining, it’s clearer for me now :)

However I think it might be a bit dangerous to keep it that way: imagine the following PR pushed in one time :

- `commit aaaaaa <== updates packages/gatsby-image, unfortunately breaks unit tests`
- `commit bbbbbb <== updates examples/using-gatsby-image`

If you take the PR as a whole CI should break, however today it does not. It might give a sense of false security for the PR author and reviewer.

PR #9462 is a good example: original PR updated `packages/gatsby-plugin-sass` and `packages/gatsby-plugin-less` and initially ended with a [commit](https://github.com/gatsbyjs/gatsby/pull/9462/commits/2fd04c53cd33135fea714247e75607ac64a731dc) changing the migration guide in `docs/`. In that case, unit tests would not be ran in the current setup.

Does that help?

DSchau:
@phacks that's the trade off. False positives! Let's just merged this and try it out :)

Issue #4624, opened 2018-03-20
Traced SVG for gatsby-remark-images


vojtechruz:
I am using Traced SVG option from gatsby-image for my non-article images and I love it.

However for my markdown images, there is only available blur-up effect, which means image loading is inconsistent. It would be nice to have traced svg also for gatsby-remark-images.

pieh:
Hey @vojtechruz , You can create PR adding support for more options to markdown images. Relevant code is here - https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-remark-images/src/index.js

humphreybc:
@pieh If you could explain the required changes at a high level, that would be quite helpful. Then I could take a stab at this. I looked at the source for `gatsby-image` to see how traced SVGs are done there, and it seems to call `image.tracedSVG`, which comes from the Sharp processing library via `gatsby-transformer-sharp`, I think. Hopefully it's as simple as that for `gatsby-remark-images`. Would love to get some more detail from you.

pieh:
@humphreybc In addition to importing and using `sizes` function from `gatsby-plugin-sharp` You would need to import `traceSVG` function, call it and adjust returned `rawHTML` to make use of result of `traceSVG` function.

KyleAMathews:
Due to the high volume of issues, we're closing out older ones without recent activity. Please open a new issue if you need help!

hackhat:
Hi there, any news about this?

Issue #4510, opened 2018-03-13
gatsby-remark-images default settings turns off visibility of the 'blur up' effect


greglobinski:
## Description

If I use `gatsby-remark-images` with default settings as it is described in docs 
```
{
    resolve: `gatsby-transformer-remark`,
    options: {
      plugins: [
        {
          resolve: `gatsby-remark-images`,
          options: {
            maxWidth: 590,
          },
        },
      ],
    },
  }
```
The 'blur up' effect is invisible. It works but is invisible. It is like that because of the plugin's defaults
```
  const defaults = {
    maxWidth: 650,
    wrapperStyle: ``,
    backgroundColor: `white`,
    linkImagesToOriginal: true,
    pathPrefix,
  }
```
The default `backgroundColor` is set to `white` and because of that  
```box-shadow: inset 0px 0px 0px 400px ${options.backgroundColor};```
is added to `img.gatsby-resp-image-image` which covers the 'blured' background of `span.gatsby-resp-image-background-image` and makes it invisible.

To see the 'blur up' effect I have to add to `gatsby-transformer-remark` setup options 
```backgroundColor: "transparent"```

with that the `box-shadow` is transparent and the 'blur up' effect is visible.

### Steps to reproduce

* Install `gatsby-remark-images` with default options as it is described in docs on website with images in remark files. Or clone https://github.com/greglobinski/gatsby-styled-blog-starter
* open the built website in Chorme with Network set to 'Slow 3G', observe the pictures loading
* then add `backgroundColor: "transparent"` to plugin setup options in `gatsby-config.js`
* rebuild website and check pictures loading again

### Expected result
I think that the plugin with default settings should work as it is described in docs "Using the "blur up" technique popularized by Medium and Facebook where a small 20px wide version of the image is shown as a placeholder until the actual image is downloaded."

### Actual result
But to see this default behavior I have to add special parameter `backgroundColor: "transparent"`

### Possible solutions

1. Set `backgroundColor` in the plugin's `defaults` to `transparent` and add info to the docs that if you set `backgroundColor` in the plugin's setup options the 'blur up' effect will not be visible. (PR to plugin code and docs)
or 
2. Add `backgroundColor: "transparent"` to the plugin's setup options in docs with info that ``backgroundColor` parameter different that `transparent` will make 'blur up' effect invisible. (PR only to docs)

I think that the first solution is the better one.

If you confirm, I can make PR.

KyleAMathews:
[...]

PR #7800, opened 2018-09-01
feat(gatsby-remark-images): make images blur up

jamesadarich:
Hey guys,

This is a proposal for adding functionality to mimic the blur up effect in `gatsby-image` for `gatsby-remark-images`. This could be nicer I think but I can't see an obvious way to do that without a great deal of refactoring around how the remark plugins are integrating (but hopefully you guys can guide to an awesome solution :) )

The pros and cons of this solution, I believe, are as follow;

**Pros**
* in keeping with current code (should be minimal impact change)
* works for images with transparencies
* no need to calculate ratio

**Cons**
* inline script - not great for CSP but I've had some thoughts about resolutions for this in general (but this is a separate piece)
* reliant on class names staying the same (but I did a minor refactor to help avoid this being a problem)
* not really testable (I have tested against a local gatsby solution and it seems to work cross browser :) )

Let me know what you guys think about the solution and any ideas you guys have

PR #9490, opened 2018-10-28
feat(gatsby-remark-images): Add option to use tracedSVG instead of blur up effect

### LINKED ISSUE #9414: Traced SVG for gatsby-remark-images
Reopen https://github.com/gatsbyjs/gatsby/issues/4624

I am using Traced SVG option from gatsby-image for my non-article images and I love it.

However for my markdown images, there is only available blur-up effect, which means image loading is inconsistent. It would be nice to have traced svg also for gatsby-remark-ima

tlent:
Closes #9414, #4624 

Unfortunately due to #4510 the effect cannot currently be seen unless `{ backgroundColor: "transparent" }` is set in the gatsby-remark-images options. I will try looking into that issue soon.

Please let me know if any changes need to be made. This is my first contribution to Gatsby.


tlent:
#7800 seems like it will be fixing the issue with the effect not being visible unless `{backgroundColor: "transparent" }` is set.

If that PR is merged `placeholderImageData` should be put inside of ``. The double quote from the style tag was what was forcing the use of `&quot;` but with the `img` tag I believe it will no longer be necessary. Single quotes around placeholderImageData were not an option due to [the way the traceSVG SVG data is encoded](https://codepen.io/tigt/post/optimizing-svgs-in-data-uris).

tlent:
I have now merged master including the changes from #7800. The issue where `{ backgroundColor: "transparent" }` was required to show the preview image has been fixed, so this PR is now working perfectly. 

#7800 was changed back to using `<span>` so my previous comment about changing the quotes no longer applies.

This PR should be ready to merge now.

packages/gatsby-remark-images/src/index.js
```diff
@@ -185,9 +213,7 @@ module.exports = (
   >
     <span
       class="${imageBackgroundClass}"
-      style="padding-bottom: ${ratio}; position: relative; bottom: 0; left: 0; background-image: url('${
-      fluidResult.base64
-    }'); background-size: cover; display: block;"
+      style="padding-bottom: ${ratio}; position: relative; bottom: 0; left: 0; background-image: url(&quot;${placeholderImageData}&quot;); background-size: cover; display: block;"
```
floo51: `&quot;` I don't think you can use a html entity here.

tlent: This is the only way I have found to use the [optimized SVG](https://codepen.io/tigt/post/optimizing-svgs-in-data-uris) that `traceSVG` returns in the style attribute. Based on [this stackoverflow post](https://stackoverflow.com/questions/5352947/how-to-escape-quotes-in-inline-styles/5352973#5352973) and [the HTML spec](https://www.w3.org/TR/html4/intro/sgmltut.html#h-3.2.2) I believe the HTML entity is valid anywhere in an HTML document. It has worked as far as I have seen in my testing.

Using double quotes around the SVG data conflicts with the style attribute's quotes. Using single quotes around the SVG data conflicts with the single quotes in the optimized SVG encoding. Using single quotes for the style attribute as a whole with double quotes around the data also conflicts with the single quotes in the optimized SVG encoding. Using no quotes around the SVG data does not work (I believe because of spaces in the data). I have not thought of any other possible solution but I am open to suggestions.

floo51: Thanks for the deep explanation.

What about escaping singles quotes in the SVG image?

```
const escapeSingleQuotes = (value) => value.replace(/'/g, '\\\'');
```
```
<div style="
    background-image: url('${escapeSingleQuotes(placeholderImageData)}');
    background-size: cover;
"
></div>
```

tlent: Thanks for the suggestion. It definitely is a cleaner solution. I just pushed a commit along these lines. Apparently webpack or something else is optimizing it along the way, so to make things even better there was no increase in the number of bytes sent to the browser even with the escaped quotes.

humphreybc:
Would be great to get this merged!

baobabKoodaa:
I would love to see this approved as well!

baobabKoodaa:
I didn't want to wait for this to get merged and I didn't want to fork Gatsby just for this, so I created a [custom component with rehype-react](https://using-remark.gatsbyjs.org/custom-components/). It does this in a rather hacky way, but it works, and it allows me to customize my Markdown images in any way I want. In case anybody else wants to use it: https://github.com/baobabKoodaa/blog/blob/master/src/components/Article/ReImage.js

sidharthachatterjee:
Published in gatsby-remark-images@3.0.5

baobabKoodaa:
Works nicely!

PR #8427, opened 2018-09-21
Add git hooks and simplify scripts

lipis:
- [x] Add git hooks to make sure the files are formatted before commiting
- [x] Simplify `format*` scripts
- [x] Drop prettier-eslint (is just a wrapper for ESLint and Prettier that we are using already)

----
### The actual formats will happen in follow up PRs

- [Format preview for `format:code`](https://github.com/gatsbyjs/gatsby/pull/8623/files)
- [Format preview for `format:other`](https://github.com/gatsbyjs/gatsby/pull/8622/files)

PR #9496, opened 2018-10-28
chore: update formatting

alexandernanberg:
Basically just run `yarn format`

We might want to add some automatic formatting when you commit, like `husky` with `lint-staged`. I'd be down for creating a PR for that if that's something you want to add.


DSchau:
@alexandernanberg thanks so much for this PR! We agree that automating this (so that we don't just _need_ a format PR every few weeks/days) would be advantageous.

You can follow along with that here -> #8427 

Going to just close this one out for the reason above! I think we should formalize the automation strategy, and then we won't even need a PR like this; that said, it's hugely appreciated and thanks for doing it!

Until next time 👋 💜

Issue #9436, opened 2018-10-26
Gtag plugin doesn't work in dev (Probably just need to document so explicitly) 


talolard:
Hi, 
We just installed the new gtag plugin and were a bit flustered until we realized it doesn't work in development. I assume it is expected behavior, it certainly feels correct in retrospect. However, it would be great to mention it in the docs. 
@tylerbuchea, can you confirm it is expected behavior ? If so, I'm happy to update docs accordingly. 
Cheers

tylerbuchea:
Right it is the expected behavior. But mentioning that in the docs would save some frustration.

Thanks for your help 👊

tylerbuchea:
This issue was resolved by ac9cbb7

PR #9497, opened 2018-10-28
docs(gatsby-plugin-google-gtag): clarify that this plugin is inactive in dev

talolard:
In reference to #9436 @tylerbuchea


talolard:
As the confused party here I disagree. 
I'ma  dumb user who just wants google analytics and I have no idea what else gtag does. When I was testing this  wanted to see myself  in the analytics dashboard and was confused. 
I think I am the lowest common denominator of people who would be using this plugin, and more sophisticated users would infer that other gtag features follow the same pattern. 

In other words, "This plugin is automatically disabled in development, and enabled in production." doesn't tell me why I'm not seeing myself in google analytics

pieh:
> user who just wants google analytics and I have no idea what else gtag does

But if people are using gtag plugin instead of GA plugin that's on users to know what's the difference (I think). Unless we somewhere recommend that for GA, people should use gtag plugin (I'm not aware of that).

Also I'm not really sure - but can You use gtag without connecting it to GA? If so this would be even more weird reading reference to GA if you aren't even using it with gtag

pieh:
Also I don't oppose mentioning GA here - but main point should be that gtag is disabled in development. It can expanded as example that if gtag is connected to GA it will not send traffic data to GA

tylerbuchea:
I found another Gatsby plugin that also only works in production mode and I really like how they worded their description for this behavior. I thought maybe we could copy them if everyone is cool with it?

Also, it's nice for the consumers of these plugins if we have some consistent messaging across the different README's for similar functionality.

The explanation in the plugin I'm referencing (https://www.gatsbyjs.org/packages/gatsby-plugin-feed/) reads like this:

> NOTE: This plugin only generates the /rss.xml file when run in production mode! To test your feed, run: gatsby build && gatsby serve.

So I figured we could update ours to read:

> NOTE: This plugin only works in production mode! To test your Global Site Tag is installed and firing events correctly run: gatsby build && gatsby serve.

I hear what you're saying though @talolard. Gtag is new and confusing and it should probably be briefly explained in the README what it is and why you should use it. Which I think @pieh suggested when I opened the original PR but I didn't get around to implementing his suggestion.

So my plan is that we add this NOTE in to describe just the prod vs dev environment differences and then I'm going to open a new PR that puts a description of what services gtag is capable of handling at the very top of the README so it will be clear that this plugin is for  Google Analytics, Google Ads, Campaign Manager, Display & Video 360, and Search Ads 360.

I'll include this link to the official Google docs on gtag as well so consumers can read more in-depth if they'd like https://developers.google.com/gtagjs/

Does that sound good to everyone? Can we rally behind that 🎉

tylerbuchea:
Also, I think the note should be moved to the top of the README just below the subtext so there is no missing it.

PR #9498, opened 2018-10-28
Fix grammar and punctuation

imronra:
- 'Building' to 'The building'
- 'in the past' to 'in the past,'

### DIFF
--- a/docs/tutorial/part-one/index.md
+++ b/docs/tutorial/part-one/index.md
@@ -364,7 +364,7 @@ Next, build your site by running the following command in the terminal at the ro
 gatsby build
 ```
 
-The Building should take 15-30 seconds. Once the build is finished, it's interesting to take a look at the files that the `gatsby build` command just prepared to deploy.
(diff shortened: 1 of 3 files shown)


docs/tutorial/part-one/index.md
```diff
@@ -364,7 +364,7 @@ Next, build your site by running the following command in the terminal at the ro
 gatsby build
 ```
 
-Building should take 15-30 seconds. Once the build is finished, it's interesting to take a look at the files that the `gatsby build` command just prepared to deploy.
+The Building should take 15-30 seconds. Once the build is finished, it's interesting to take a look at the files that the `gatsby build` command just prepared to deploy.
```
yeskunall:
```suggestion
The build should take 15-30 seconds. Once the build is finished, it's interesting to take a look at the files that the `gatsby build` command just prepared to deploy.
```

I think that is more appropriate.

shannonbux: Yes, @yeskunall's suggestion is correct!

PR #9505, opened 2018-10-28
Update README.md

jsoningram:
Need to install dependencies before running the local server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,7 @@ You can get a new Gatsby site up and running on your local dev environment in 5 
 
     ```sh
     cd my-blazing-fast-site/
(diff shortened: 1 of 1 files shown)


oorestisime:
Hm gatsby new installs all dependencies needed for the first run. I would say this PR is invalid

pieh:
`gatsby new` handles dependency installation, so this is not needed.

jsoningram:
Running `gatsby new my-blazing-fast-site` using v2.4.3 creates the directory `my-blazing-fast-site` along with a lock file but no `node_modules`. Perhaps this is an issue instead?

PR #9508, opened 2018-10-28
feat(gatsby): Move connection out of sift

Moocar:
**Builds on #9416 . Merge that one before this**

In additional to performing the queries, `run-sift.js` also currently converts the results into connections (if connection query). But as part of #9338, we will be using `loki` to perform queries too. So we would need to duplicate the connection logic in both query engines.

This PR pulls the connection logic out of run-sift and into `build-node-connections` which makes sense to me as the query engine doesn't need to know about connections. Only about whether it should return one result or many (thus the `firstOnly` flag).

Another benefit is that the query engine no longer needs to understand page dependencies.


pieh:
merged #9416, can you update PR seems like our squashing on merge desynced this branch :(

Moocar:
@pieh I'm definitely a fan of that new "Suggested change" feature :D. Should be good now

PR #9515, opened 2018-10-29
Update LICENSE

vyaspranjal33:
updated year 2015 -> 2018 in LICENSE

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2015 Gatsbyjs
+Copyright (c) 2018 Gatsbyjs
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


DSchau:
Hi @vyaspranjal33!

Thanks so much for the PR. As far as the license relates, as best as I can tell, the _start_ date is the most important. As Gatsby was created in 2015, I think this is accurate and should remain. As such, gonna close this out!

For further info, see [this article](https://softwareengineering.stackexchange.com/questions/227812/why-do-we-need-years-in-a-software-license), particularly this section:

> The start date is the important part, really. The end date doesn't matter as much, since it's recognized that sometimes things go for more than a year without being updated. Even if your datestamp goes beyond what the law would specify is the term of your copyright, the law wins

Thanks for the PR, and we hope to see more from you in the future! 💜

vyaspranjal33:
@DSchau  please see #9521

PR #9516, opened 2018-10-29
Grammatical fix: "set up" is two words (except when it's a noun)

Keavon:


hugovk:
@pranshuchittora No, the link says it should be "set up".

"[they are] all set up and waiting for you to start building"

is similar to:

> Set up
> This is the verb ‘to set up’.
>
> You must set up your computer.
> To set up your stall now would be most wise.
> He was told lies and was completely set up.

PR #9522, opened 2018-10-29
Improvements to tutorial part zero

tlent:
This PR:

- Simplifies the headings and makes them more clear and consistent
- Clarifies some of the sentences I thought might be confusing
- Improves the summary at the end to cover everything done
- Expands the Git section to better explain how to install it. I added links to the Atlassian guide for installing Git because I found they had the best explanations. The linked guides include a check that git is working so I did not include that.
- Adds an optional link to learn more about Git and Github to the Git section and removes a later link serving the same purpose


amberleyromo:
I'll leave this to @shannonbux to double-check the headings, but I think these changes make sense 👍

docs/tutorial/part-zero/index.md
```diff
@@ -55,13 +57,21 @@ npm has three distinct components: the npm website, the npm registry, and the np
...
+
+The steps to download and install Git depend on your operating system. Follow the guide for your system:
 
-Git is a free and open source distributed version control system designed to handle everything from small to very large projects with speed and efficiency. When you install a Gatsby "starter" site, Gatsby uses Git behind the scenes to download and install the required files for your starter.
+- [Install Git on macOS](https://www.atlassian.com/git/tutorials/install-git#mac-os-x)
+- [Install Git on Windows](https://www.atlassian.com/git/tutorials/install-git#windows)
+- [Install Git on Linux](https://www.atlassian.com/git/tutorials/install-git#linux)
 
-If your system does not have Git installed, install it from the [**Git downloads page**](https://git-scm.com/downloads).
+> 💡 You will not need to use Git yourself to work with Gatsby, but it is a very useful tool to know about as a developer. If you are interested in learning more about version control, Git, and Github, check out Github's [Git Handbook](https://guides.github.com/introduction/git-handbook/).
```
yeskunall: If we're going to make this change, then we also need to mention that git is used everywhere else when cloning gatsby-starters.

docs/tutorial/part-zero/index.md
```diff
@@ -55,13 +57,21 @@ npm has three distinct components: the npm website, the npm registry, and the np
...
 
-## Install the Gatsby CLI
+## The Gatsby CLI
```
yeskunall: The "the" here is unnecessary. What I mean is, it adds no meaning. The Gatsby CLI would make sense in a world where this is the __only__ Gatsby CLI ever. 😆

www/src/data/sidebars/tutorial-links.yaml
```diff
@@ -8,16 +8,16 @@
...
-        - title: Install Node.js
-          link: /tutorial/part-zero/#install-nodejs
-        - title: Familiarize with npm
-          link: /tutorial/part-zero/#familiarize-with-npm
-        - title: Install Git
-          link: /tutorial/part-zero/#install-git
-        - title: Install the Gatsby CLI
-          link: /tutorial/part-zero/#install-the-gatsby-cli
+        - title: The command line
+          link: /tutorial/part-zero/#the-command-line
```
yeskunall: The hyperlinks seem more disconnected to the doc/top we're talking about 🤔

docs/tutorial/part-zero/index.md
shannonbux: Yes, could you please clarify when Git is *not* necessary? It is necessary for many things people do with Gatsby, such as using starters and contributing to Gatsby.

docs/tutorial/part-zero/index.md
shannonbux: I agree!

docs/tutorial/part-zero/index.md
```diff
@@ -92,11 +102,11 @@ npm install --global gatsby-cli
 
 If successfully installed, running `gatsby --version` should return a version number, and running `gatsby --help` will show different commands available to you using the `gatsby-cli` tool.
 
-## Create a site
+## Create a Gatsby site
 
-Now you can use the gatsby-cli tool to create your first Gatsby site. Using the tool, you can use “starters”, or partially built sites with some default configuration, to help you get moving faster on creating a certain type of site. The “Hello World” starter you’ll be using here is a starter with the bare essentials needed for a [Gatsby](/) site.
+Now you are ready to use the Gatsby CLI tool to create your first Gatsby site. Using the Gatsby CLI you can download “starters” (partially built sites with some default configuration already done) to help you get moving faster on creating a certain type of site. The “Hello World” starter you’ll be using here is a starter with the bare essentials needed for a Gatsby site.
```
shannonbux: This is a bit wordy. Try "Now you are ready to use Gatsby CLI to create your first Gatsby site. Using the tool, you can download..."

And then remove the phrase "already done." Default configuration implies the same meaning.

docs/tutorial/part-zero/index.md
```diff
@@ -92,11 +102,11 @@ npm install --global gatsby-cli
...
 
-### ✋ Create a Gatsby site
+### ✋ Create the site
```
shannonbux: Hm, I like the previous heading more because it's clearer

docs/tutorial/part-zero/index.md
tlent: Good point this sentence was not very clear. Hopefully the new version is better. The reason I have included it is to make sure nobody reading gets overwhelmed or too distracted by learning Git or all of its commands that the handbook talks about

docs/tutorial/part-zero/index.md
tlent: You are right I have changed this title and other references to it to make it more clear that it is not the only Gatsby CLI

docs/tutorial/part-zero/index.md
tlent: You are right I have made these changes in the new version

tlent:
Thank you for the feedback! I have reverted the header changes to have verbs again and replied to the other suggestions above.

docs/tutorial/part-zero/index.md
```diff
@@ -17,17 +17,19 @@ It’s not necessary to be an expert with these already — if you’re not, don
...
 
-## Familiarize with command line
+## Familiarize with the command line
 
 The command line is a text-based interface used to run commands on your computer. (You’ll also often see it referred to as the terminal. In this tutorial we’ll use both interchangeably). It’s a lot like using the Finder on a Mac, or Explorer on Windows. Finder and Explorer are examples of graphical user interfaces (GUI). The command line is a powerful, text-based way to interact with your computer.
 
+### ✋ Familiarize yourself with the command line
```
calcsam: this should probably not be here?

yeskunall: I think they are following suite with the rest of the doc.

shannonbux:
I edited the headers to streamline them and removed the emojis in headers for now. We could add them back in later.

PR #9532, opened 2018-10-29
Add usehooks.com

elrumordelaluz:


gragland:
@DSchau Wooo awesome! And thanks for submitting @elrumordelaluz!

elrumordelaluz:
@gragland love the site and the idea, looking forward for those daily `hooks`!

PR #9537, opened 2018-10-29
Fix(www): Containter component may have class: 'undefined'.

emtei:
Set `className` default value to be an empty string, so `undefined` won't appear in HTML code if `className` isn't passed in props.


pieh:
Is this (class="undefined") happening anywhere on gatsbyjs.org right now?

emtei:
@pieh yes, i spotted it here:
https://www.gatsbyjs.org/docs/schema-generation/

image

PR #9538, opened 2018-10-29
docs: add section on multiple layouts to gatsby-layout-plugin docs

msachi:
Adding a section on how to use multiple layouts to gatsby-layout-plugin docs. Might be helpful to anyone trying to build a site with a more complex layout system (and who still needs to use or prefers the layout philosophy used in Gatsby v1).


pieh:
I would move this to `Troubleshooting` section. This definitely need to be below v1 vs v2 explanation

msachi:
No worries, have updated.

PR #9549, opened 2018-10-30
chore: move run-sift

Moocar:
**Builds on #9508. Merge that first**

Another refactoring PR before loki hits. `run-sift` has existed in schema up until now. This PR moves it to under the `redux` folder. The idea is that `schema` deals with the creation of a schema, but the resolver is code that performs a query against a database. This will make even more sense with loki.

I've also moved `node-tracking` from schema to db in anticipation that this will eventually be stored in the DB as well.


packages/gatsby/src/schema/build-node-types.js
```diff
@@ -192,33 +190,20 @@ module.exports = async ({ parentSpan }) => {
...
-          let latestNodes
-          if (
-            process.env.NODE_ENV === `production` &&
-            nodesCache.has(typeName)
-          ) {
-            latestNodes = nodesCache.get(typeName)
-          } else {
-            latestNodes = getNodesByType(typeName)
-            nodesCache.set(typeName, latestNodes)
-          }
```
pieh: can we apply this caching to `getNodesByType` utility function so it doesn't filter all nodes every time in production builds?

Moocar:
I'm hesitant to add caching inside the getNodesByType implementation. What if the function is called halfway through sourceNodes when new nodes are still being added?

I think it still makes sense adding it during run-sift until loki is implemented. Since the redux version is quite slow.

pieh:
> I'm hesitant to add caching inside the getNodesByType implementation. What if the function is called halfway through sourceNodes when new nodes are still being added?

That's good point! We could do cache invalidation when we create/update/delete node of given type, but this is out of the scope for this PR.

pieh:
Was just testing locally, and this actually improve performance a little bit as the caching that You re-added in `run-sift` yesterday isn't currently used for connection queries ;)

Issue #3998, opened 2018-02-13
Gatsby Wordpress PLugin returns an error - No Gatsby Nodes


shansmith01:
Hi. I am running into issues using the WordPress plugin on a self-hosted WordPress website using basic authentication.

On running gatsby build I get an error and no nodes generate. Details. below.

Note I have tried this on a couple of different wordpress sites, I have also tried with the gatsby wordpress starter and always run into the same issueç

### Environment

Gatsby version: 1.1.39
Node.js version: v9.4.0
Operating System: MacOS

### File contents (if changed):

`gatsby-config.js`: 

```
module.exports = {
  siteMetadata: {
    title: `A sample site using gatsby-source-wordpress`,
    subtitle: `Data fetched from a site hosted on wordpress.com`,
  },
  plugins: [
    // https://public-api.wordpress.com/wp/v2/sites/gatsbyjsexamplewordpress.wordpress.com/pages/
    /*
     * Gatsby's data processing layer begins with “source”
     * plugins. Here the site sources its data from Wordpress.
     */
    {
      resolve: `gatsby-source-wordpress`,
      options: {
        /*
        * The base URL of the Wordpress site without the trailingslash and the protocol. This is required.
        * Example : 'gatsbyjswpexample.wordpress.com' or 'www.example-site.com'
        */
        baseUrl: `nomadsworld.com`,
        // The protocol. This can be http or https.
        protocol: `https`,
        // Indicates whether the site is hosted on wordpress.com.
        // If false, then the asumption is made that the site is self hosted.
        // If true, then the plugin will source its content on wordpress.com using the JSON REST API V2.
        // If your site is hosted on wordpress.org, then set this to false.
        hostingWPCOM: false,
        // If useACF is true, then the source plugin will try to import the Wordpress ACF Plugin contents.
        // This feature is untested for sites hosted on Wordpress.com
        useACF: true,

        auth: {
 htaccess_user: "HIDDEN",
 htaccess_pass: "HIDDEN",
 htaccess_sendImmediately: false,
},
  verboseOutput: true
        
      },
   
    },
  ],
}
```

`package.json`: 

```
{
  "name": "gatsby-starter-default",
  "description": "Gatsby default starter",
  "version": "1.0.0",
  "author": "Kyle Mathews <mathews.kyle@gmail.com>",
  "dependencies": {
    "gatsby": "^1.9.193",
    "gatsby-link": "^1.6.36",
[...]

PR #9558, opened 2018-10-30
Feat(gatsby-source-drupal): Added rateLimit to prevent ECONNREFUSED

timhagn:
I ran into a similar Problem like the already closed issue #3998 for gatsby-source-wordpress
(hosting provider using a DDOS prevention module, in this case mod_qos for apache),
leading to "mod_qos(010): access denied, QS_LocRequestLimit*" errors on the
server and resulting in **ECONNREFUSED** or **socket closed** failures for 
`gatsby develop` & `gatsby build` as well as the often cited
`warning The gatsby-source-drupal plugin has generated no Gatsby nodes. Do you need it?`.

I therefore added an option `rateLimit` and replaced the calls to `axios.get()` with an 
`axios.create()` Client call and added a `rateLimiter` request interceptor.

The tests are running through, only the coverage fell from 90% line to 89% (I didn't know how
to handle the `setTimeout()` call).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,7 @@ node_modules/
 
 # IDE specific
 .idea/
+*.iml
 .vscode/
 *.sw*
 
--- a/packages/gatsby-source-drupal/README.md
+++ b/packages/gatsby-source-drupal/README.md
@@ -13,6 +13,9 @@ https://using-drupal.gatsbyjs.org/
 jsonapi used by your Drupal instance. The default value is `jsonapi`, which has
 been used since jsonapi version `8.x-1.0-alpha4`.
 
+`rateLimit` Option allows setting a request rate limit if Drupal sits on a 
+server with DDOS prevention enabled.
+
 ## Install
 
 `npm install --save gatsby-source-drupal`
@@ -27,6 +30,7 @@ plugins: [
     options: {
       baseUrl: `https://live-contentacms.pantheonsite.io/`,
       apiBase: `api`, // optional, defaults to `jsonapi`
+      rateLimit: 1000, // optional, ignored by default
     },
   },
 ]
--- a/packages/gatsby-source-drupal/src/__tests__/index.js
+++ b/packages/gatsby-source-drupal/src/__tests__/index.js
@@ -23,7 +23,7 @@ jest.mock(`axios`, () => {
         },
         interceptors: {
           request: {
-            use: fn => fn,
+            use: fn => fn(),
           },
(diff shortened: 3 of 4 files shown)


pieh:
This is looking cool - one thing that might be more user friendly:

- don't require explicit `rateLimit` setting
- create abstraction around `axios` that would catch `ECONNREFUSED` errors and then apply rate limiting automatically - do you think it would be feasible to do?

timhagn:
Thanks for the comment, although I don't fully understand what you mean by "explicit `rateLimit` setting",
it's optional by default (and doesn't get used if none is given), see [README.md](packages/gatsby-source-drupal/README.md).
For your second remark I have to confess, that this would go beyond my (current) knowledge of `axios`,
but as deep as I went in the sources, it seems that `axios` is generally treating any post 400 statuses as errors and the 403 returns an `undefined` `error.response` Object. Would be great if the rate limiting could be done automatically, though...

pieh:
With explicit `rateLimit` setting I meant if code could automatically apply rateLimiting (when getting `ECONNREFUSED`), then setting wouldn't be needed at all as this would be hidden from user. (I understand it's optional in current form :) ).

timhagn:
> With explicit `rateLimit` setting I meant if code could automatically apply rateLimiting (when getting `ECONNREFUSED`), then setting wouldn't be needed at all as this would be hidden from user. (I understand it's optional in current form :) ).

Ah, OK, so both refer to your second remark ^^. Yeah, would be great, but there I'm sadly out of my depth at the moment (just did this PR, as I had exactly the mentioned problem and from SO it seemed like I wasn't the only one ; ).

dubcanada:
There is axios-retry and other libraries that catch certain HTTP codes and retry the connection. Could use one of those to do this?

timhagn:
@dubcanada Thanks for the suggestion, axios-retry looks promising with exponentialDelay().
Gonna look into it : ).
Don't know which other modules you thought about, though, most I looked into on npmjs depend on axios's error-handling (at least the ones dependent on ^0.18.0)...

timhagn:
Hi @wardpeet,
after @KyleAMathews asked around on Twitter for a way to solve this issue for "Gatsby Core",
I put this matter on hold. But if there's still interest I'd be happy to find some time to look into it further,
after rebasing onto the current master of the Gatsby repo.

wardpeet:
ok i'll be closing until this issue comes up again. Thank you for your time!

PR #9564, opened 2018-10-30
Extend the docs on `sourceNodes`

chmac:
Expanding the docs to be a bit more verbose and explain in more depth what this API does. Based on a discussion with @pieh.

### DIFF
--- a/packages/gatsby/src/utils/api-node-docs.js
+++ b/packages/gatsby/src/utils/api-node-docs.js
@@ -78,7 +78,12 @@ exports.createPages = true
 exports.createPagesStatefully = true
 
 /**
- * Extension point to tell plugins to source nodes.
+ * Extension point to tell plugins to source nodes. This API is called during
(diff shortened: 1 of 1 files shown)


packages/gatsby/src/utils/api-node-docs.js
```diff
@@ -78,7 +78,12 @@ exports.createPages = true
 exports.createPagesStatefully = true
 
 /**
- * Extension point to tell plugins to source nodes.
+ * Extension point to tell plugins to source nodes. This API is called during
+ * the Gatsby bootstrap sequence. Source plugins use this hook to create nodes.
+ * This API is called exactly once per plugin (and once for your site's
+ * `gatsby-config.js` file). If you define this hook in `gatsby-config.js` it
```
DSchau: Do you mean in `gatsby-node.js`?

chmac: Oooh, nice catch. Yep, that’ll be a mistake, will fix and push an update.

PR #9590, opened 2018-10-31
Cloning

Bolsterland:
#About cloning...

If I were to clone myself 5 times and the cloner does not keep track of which is the original, when I wake up, how will I know if I'm the real one?

## A theoretical concept

| CloneID  | WakeUpLocation  | Status |
| ---    | ---      | ---        |
| Z1       | Laboratory 3579  | Thinks he's the original purely based on wake-up location|
| Z2       | Sunway Library   | Confused, being surrounded by biology books and cloning research disoriented subject|
| Z3       | Zhe Jing's Bed   | Thought nothing of the cloning procedure, dismissed it as a dream, in denial |
| Z4       | Disneyland       | Excited, forgot about cloning, fully-utilized VIP pass to have fun|
| Z5       | Deserted Island in the Pacific | Scared and panicked, fully-aware of cloning procedure and cursed his luck, acceptance of death = True |


m-allanson:
Hey @Bolsterland, it's not clear from your PR description what change you're trying to do here. The PR seems to have pulled in a lot of unrelated commits. Could you re-open this with a clearer description? Thanks!

PR #7943, opened 2018-09-06
[v2] Don't match the canonical path for 404 pages

davidbailey00:
Fixes #7939

PR #9595, opened 2018-10-31
fix(gatsby): detect additional 404 page locations when performing canonical path redirects

davidbailey00:
Fixes the problem reported by @lipis in #7943

I've also reordered and commented all the checks so it makes more sense to someone unfamiliar


lipis:
👍

Issue #6649, opened 2018-07-21
Don't use `css` prop in components to pass in styles


wuweiweiwu:
## Description

Components in the website like [Container](https://github.com/gatsbyjs/gatsby/blob/master/www/src/components/container.js) with a prop `css`. When you actually try to specify it, it conflicts with the `glamor` css attribute and the prop isnt passed properly. 

Use something like `containerCss` shown  in [`email-capture-form`](https://github.com/gatsbyjs/gatsby/blob/master/www/src/components/email-capture-form.js#L88)

KyleAMathews:
We are using Glamor so that component is working as expected. Or is there something else going on here?

wuweiweiwu:
Some components use like Container use the css prop to pass in additional style props to the root element that has the glamor css attribute. But when I was doing development locally, it seems to always pass an empty object. I'm thinking that the css prop name conflicts with the glamor css prop.

trevorlitsey:
Hi! Would love to help out on this if it still needs resolving. Looks like all that needs to be done is change the `<Container />` `css` prop to `containerCss`? Didn't see any other react components with a css prop from a cursory search.

m-allanson:
Hey @trevorlitsey I missed your comment earlier, apologies! It turns out there was a bit of tidying up that could happen here - check out @jamezrin's PR for details.

> We are using Glamor so that component is working as expected. Or is there something else going on here?

@KyleAMathews I think the [createElement helper](https://github.com/threepointone/glamor/blob/master/docs/createElement.md) is preventing the `css` prop from being passed through? I confirmed that it wasn't being passed through but didn't do too much digging beyond that.

KyleAMathews:
@m-allanson yeah, looks that's right. Curious why I thought it was working before in the past 🤔

PR #9596, opened 2018-10-31
Add `overflowY: scroll` to filtered starters sidebar

### LINKED ISSUE #9592: gatsby.org starter-pack sidebar doesn't have a scroll bar (in Firefox)
## Description

The starter pack website, once filtered, doesn't have a scroll bar to allow me to scroll though the filters. This means that only after I reach the bottom of the main content does the sidebar content scroll. 

### Steps to reproduce

Go to [this link](https://www.gatsbyjs.org/starters?c=i18n) and scroll on the sidebar. 

### Expected result

When my cursor is above the sidebar, and I turn the wheel on my mouse the sidebar should scroll. Alternatively, I should be shown a scrollbar so I can drag my mouse to where i want it to be.

### Actual result

There is no scrollbar in the sidebar. When it is active and I am scrolling, the bottom of the content is reached before the sidebar starts scrolling.

### Environment

I looked at the website on a Macbook Pro on Chrome and Firefox, this bug is only present on all Firefox browsers (Regular, dev edition, nightly)

hasparus:
Closes #9592.

### DIFF
--- a/www/src/views/shared/sidebar.js
+++ b/www/src/views/shared/sidebar.js
@@ -2,8 +2,8 @@ import React from "react"
 import MdFilterList from "react-icons/lib/md/filter-list"
 import styles from "../shared/styles"
 
-export const SidebarContainer = ({ children, overrideCSS }) => (
-  <div css={[styles.sidebarContainer, styles.sticky, overrideCSS]}>
+export const SidebarContainer = ({ children, className }) => (
+  <div className={className} css={[styles.sidebarContainer, styles.sticky]}>
     {children}
   </div>
 )
--- a/www/src/views/shared/sidebar.js
+++ b/www/src/views/shared/sidebar.js
@@ -3,9 +3,7 @@ import MdFilterList from "react-icons/lib/md/filter-list"
 import styles from "../shared/styles"
 
 export const SidebarContainer = ({ children, style }) => (
-  <div css={[styles.sidebarContainer, styles.sticky]} style={style}>
-    {children}
-  </div>
+  <div css={[styles.sidebarContainer, styles.sticky, style]}>{children}</div>
 )
 
 export const SidebarBody = ({ children }) => (
--- a/www/src/views/shared/sidebar.js
+++ b/www/src/views/shared/sidebar.js
@@ -2,8 +2,10 @@ import React from "react"
 import MdFilterList from "react-icons/lib/md/filter-list"
 import styles from "../shared/styles"
 
-export const SidebarContainer = ({ children, style }) => (
-  <div css={[styles.sidebarContainer, styles.sticky, style]}>{children}</div>
+export const SidebarContainer = ({ children, overrideCSS }) => (
+  <div css={[styles.sidebarContainer, styles.sticky, overrideCSS]}>
+    {children}
+  </div>
 )
 
 export const SidebarBody = ({ children }) => (
--- a/www/src/views/shared/sidebar.js
+++ b/www/src/views/shared/sidebar.js
@@ -2,8 +2,10 @@ import React from "react"
 import MdFilterList from "react-icons/lib/md/filter-list"
 import styles from "../shared/styles"
 
-export const SidebarContainer = ({ children }) => (
-  <div css={[styles.sidebarContainer, styles.sticky]}>{children}</div>
+export const SidebarContainer = ({ children, style }) => (
+  <div css={[styles.sidebarContainer, styles.sticky]} style={style}>
+    {children}
+  </div>
 )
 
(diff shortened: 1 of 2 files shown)


www/src/views/shared/sidebar.js
```diff
@@ -2,8 +2,8 @@ import React from "react"
 import MdFilterList from "react-icons/lib/md/filter-list"
 import styles from "../shared/styles"
 
-export const SidebarContainer = ({ children }) => (
-  <div css={[styles.sidebarContainer, styles.sticky]}>{children}</div>
+export const SidebarContainer = ({ children, style }) => (
+  <div css={[styles.sidebarContainer, styles.sticky, style]}>{children}</div>
```
hasparus: @pranshuchittora, this `style` is actually passed to css prop on a div below (is it `glamor` BTW? I can see there's `gatsby-plugin-emotion` installed, but it looks like `glamor`), so it's as inline as 
https://github.com/gatsbyjs/gatsby/blob/abe65803b0239a2aa811f73e1bdf9c18b626bd2a/www/src/views/starter-library/filtered-starters.js#L107
few lines below.

I couldn't name it `css` as in https://github.com/gatsbyjs/gatsby/issues/6649, so I've used `style`, but now I think the prop should probably be named `overrideCSS`  (as in Button) for consistence.

Or do you mean I should add this to `styles.sidebarContainer`? It would affect other pages though.

hasparus:
@pranshuchittora, this `style` is actually passed to css prop on a div below (is it `glamor` BTW? I can see there's `gatsby-plugin-emotion` installed, but it looks like `glamor`), so it's as inline as 
https://github.com/gatsbyjs/gatsby/blob/abe65803b0239a2aa811f73e1bdf9c18b626bd2a/www/src/views/starter-library/filtered-starters.js#L107


I couldn't name it `css` ( https://github.com/gatsbyjs/gatsby/issues/6649 ), so I've used `style`, but now I think the prop should probably be named `overrideCSS`  (as in Button) for consistence.

Or do you mean I should add this to `styles.sidebarContainer` so it affects other SidebarContainers too?

fk:
@hasparus It indeed now is emotion and its "[object styles](https://emotion.sh/docs/object-styles)" (which made transitioning from `glamor` to `emotion` quite easy).

www/src/views/shared/sidebar.js
```diff
@@ -2,8 +2,10 @@ import React from "react"
 import MdFilterList from "react-icons/lib/md/filter-list"
 import styles from "../shared/styles"
 
-export const SidebarContainer = ({ children }) => (
-  <div css={[styles.sidebarContainer, styles.sticky]}>{children}</div>
+export const SidebarContainer = ({ children, overrideCSS }) => (
+  <div css={[styles.sidebarContainer, styles.sticky, overrideCSS]}>
```
DSchau: I think we clean this up a little bit. Can we try something like the following:

```jsx
export const SidebarContainer = ({ children, className }) => (
  <div className={className} css={[styles.sidebarContainer, styles.sticky]}>{children}</div>
```

(and then see below for more info on how to _use_)

www/src/views/starter-library/filtered-starters.js
```diff
@@ -101,7 +101,7 @@ export default class FilteredStarterLibrary extends Component {
 
     return (
       <section className="showcase" css={{ display: `flex` }}>
-        <SidebarContainer>
+        <SidebarContainer overrideCSS={{ overflowY: "auto" }}>
```
DSchau:
```suggestion
        <SidebarContainer css={{ overflowY: "auto" }}>
```

#6649 doesn't seem relevant anymore because we've moved off of glamor?

www/src/views/shared/sidebar.js
DSchau: (also just for clarification here, the way CSS in JS libraries--like emotion!--work here is that they inject a className, so consuming a class name is a good way to make a component externally stylable with the CSS prop or `styled(SidebarContainer)()`)

www/src/views/starter-library/filtered-starters.js
hasparus: I believe the problem OP of #6649 had is still relevant, because `css` is [syntactic sugar](https://emotion.sh/docs/css#css-prop) for `className={css(...)}`, so we can't pass it like a prop without thinking it's actually `className`.

I'm gonna refactor it right now to use `css` and `className`.

DSchau: I'm not really sure of your point here, but it looks like you went this route?

PR #9599, opened 2018-10-31
Updated License

PankajzYadav:
Updated License to 2018


pieh:
It doesn't have to be updated - see https://github.com/gatsbyjs/gatsby/pull/9597

Issue #9539, opened 2018-10-29
[gatsby-source-filesystem] createRemoteFileNode: Keep original name


LekoArts:
## Summary

Since this issue (https://github.com/gatsbyjs/gatsby/issues/6232) was fixed by @Sekhmet only one thing is left: Also keep the original file name in the downloaded image (as per old issue https://github.com/gatsbyjs/gatsby/issues/3132). Other requests: https://github.com/angeloashmore/gatsby-source-prismic/issues/40

@Sekhmet do you want to tackle that issue, too?

A result should be, e.g.:
- filepath: 
`/static/f28264fbe0c2cf6d02c3b8adc54b717b/18b00/e10dce13ec246d93bf6e57ede0d98ce8.jpg`
- new filepath: 
`/static/f28264fbe0c2cf6d02c3b8adc54b717b/18b00/e10dce13ec246d93bf6e57ede0d98ce8/this-is-the-super-name-here.jpg`

hbish:
I'll have a go at this.

jkjustjoshing:
What's the practicality of allowing the hash to be customizable too (or fully remove for non-duplicate filenames?)?

I'm moving a site over that has a bunch of PDFs on it. I'm using [gatsby-source-directus](https://www.gatsbyjs.org/packages/gatsby-source-directus/) to pull down the PDFs from my Directus server, which uses `createRemoteFileNode()` to add the files to Gatsby. 

Our current PDF urls look something like this - https://www.example.com/static/Financial_Disclosure_Form.pdf - and the users frequently will ⌘+S to save the PDFs. Right now, this causes the default filename to be the MD5 hash of the file, not `Financial_Disclosure_Form.pdf`. 

It looks like PR #9777 will make it so the filename is `79054025255fb1a26e4bc422aef54eb4-Financial_Disclosure_Form.pdf`, which is also not acceptable to me. Would it be possible to do one (or all) of the following?
1. Put the hash at the end - `Financial_Disclosure_Form-79054025255fb1a26e4bc422aef54eb4.pdf`
2. Allow for a shorter hash length (and throw an exception if there is a conflict) - `Financial_Disclosure_Form-79054.pdf`
3. Allow for no hash if there is no filename conflict (how easy will it be to detect filename conflicts vs one file requested twice?)
[...]

PR #9601, opened 2018-10-31
keep original name of remote files along with hash

hbish:
This PR is to enhance `createRemoteFileNode` to retain the filename of the remote file. Additional details in #9539 and [https://github.com/angeloashmore/gatsby-source-prismic/issues/40](https://github.com/angeloashmore/gatsby-source-prismic/issues/40)

### DIFF
--- a/packages/gatsby-source-filesystem/src/utils.js
+++ b/packages/gatsby-source-filesystem/src/utils.js
@@ -13,3 +13,16 @@ const Url = require(`url`)
 export function getRemoteFileExtension(url) {
   return path.parse(Url.parse(url).pathname).ext
 }
+
+/**
+ * getRemoteFileName
+ * --
+ * Parses remote url to retrieve remote file name
+ *
+ *
+ * @param  {String}          url
+ * @return {String}          filename
+ */
+export function getRemoteFileName(url) {
+  return path.parse(Url.parse(url).pathname).name
+}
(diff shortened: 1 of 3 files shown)


packages/gatsby-source-filesystem/src/utils.js
```diff
@@ -13,3 +13,16 @@ const Url = require(`url`)
...
+ * getRemoteFileName
+ * --
+ * Parses remote url to retrieve remote file name
+ *
+ *
+ * @param  {String}          url
+ * @return {String}          filename
+ */
+export function getRemoteFileName(url) {
+  return path.parse(Url.parse(url).pathname).name
```
DSchau: Since this block is pretty much directly shared between `getRemoteFileExtension` and this function, maybe encapsulate the following code into a shared function so as to not duplicate work--and later, if we need to tweak, it'll be an easy swap in _one_ place rather than a few, e.g.

```js
function getParsedPath(url) {
  return path.parse(Url.parse(url).pathname)
}

export function getRemoteFileExtension(url) {
  return getParsedPath(url).ext
}

export function getRemoteFileName(url) {
  return getParsedPath(url).name
}
```

hbish: Fixed. Thanks for the review.

DSchau:
@hbish whoops, looks like we have a merge conflict. If you can fix that up we can get this merged; thank you!

hbish:
@DSchau No issues, I've fixed the merge conflict.

PR #9605, opened 2018-10-31
feat(examples): add a remote-images plugin so as to not introduce bloat

DSchau:
Note: this PR requires #9349 to be merged first

I _think_ this is an improved mechanism for sourcing remote data, so we don't have to have quite a large payload with all those local images!

Wanted to get some 👀 on it!


DSchau:
Also note: there are some downsides to this approach, namely:

- Requires network connection--and quite a good one, the images are _huge_
- Introduces a dependency on Unsplash being up for site to function--but this is probably _OK_

DSchau:
> Hmm, maybe we should keep one or two images in the repo to show both ways—filesystem with "real" local image and downloaded images?

Love that idea! We'll keep one or two of the smaller ones around and then source most other things remotely.

> Unsplash became an institution

It does make sense 👍

DSchau:
@fk this should be good to go now! Kept one image (the one on the index route) but otherwise, everything else is sourced from remote data from Unsplash!

fk:
<3 <3 <3 Gotta wrap my PR up finally :/ :/ :/ … template literals here I come ;-P

examples/using-gatsby-image/src/pages/traced-svg.js
```diff
@@ -40,51 +40,61 @@ export default TracedSVG
...
+          }
         }
       }
     }
-    fullWidthImage: file(relativePath: { regex: "/chuttersnap-1103171/" }) {
-      childImageSharp {
-        fluid(maxWidth: 600) {
-          ...GatsbyImageSharpFluid_tracedSVG
+    fullWidthImage: unsplashImagesYaml(title: { eq: "City from above" }) {
+      credit
```
DSchau: Also @fk I don't know if we _have_ to credit, but we have it available in the YAML now 🙃

fk: 👍 👍 🙏 Yeah that's great! We don't have to credit (https://unsplash.com/license), but I'd like to do so somehow (a little more than putting the credits in the image title). Then again, it's a demo. I think we don't even have page `<title>`s…

fk:
> we should probably merge this

👌Will do!

Managed to refactor a little more to use template literals last night, but didn't get through. Also gotta say I didn't really enjoy that — it just doesn't feel right (anymore?! … maybe the thought "but it gets parsed somehow somewhere _not by the browser_ into something anyway" adds a little objectivity?!) 😅

DSchau:
@fk I totally agree. I guess the metric that we should keep examples approachable _also_ applies here. But yeah, it's kind of tiresome and feels backward to me to use template literals too.

I wonder (and hope/wish) if there's some automated tooling that will do it for us? Closest I found which doesn't _quite_ work -> https://github.com/lachlanjc/objectify-css

KyleAMathews:
Or we could just not use template literals for styles.  Even if you've never seen styles as objects, they're similar enough that people can parse them with a bit of work.

KyleAMathews:
Our whole MO of gatsby is having opinions and teaching people about them. I'm very sure that styles as objects will be a significant part of the future of CSS on the web. It's very prevalent. Also react native defaults to it https://facebook.github.io/react-native/docs/style

PR #9606, opened 2018-10-31
feat(www): add 'lead' animation to the Ecosystem horizontal scrollers

greglobinski:
* ~install `intersection-observer` polyfill~
* refactor `ecosystem-board.js` to `class` component to manage IntersectionObserver
* setup `IntersectionObserver` to observe horizontal scrollers
* add style class with initial 'lead' animation

Video preview at Slack

### DIFF
--- a/www/src/components/ecosystem/ecosystem-board.js
+++ b/www/src/components/ecosystem/ecosystem-board.js
@@ -88,7 +88,6 @@ class EcosystemBoard extends Component {
       icons: { plugins: PluginsIcon, starters: StartersIcon },
       starters,
       plugins,
-      background,
     } = this.props
 
     return (
--- a/www/src/components/ecosystem/ecosystem-board.js
+++ b/www/src/components/ecosystem/ecosystem-board.js
@@ -53,7 +53,7 @@ class EcosystemBoard extends Component {
       if (entry.intersectionRatio > 0) {
         setTimeout(
           () => this.turnOnLeadScroll({ target, duration: 1000, distance: 20 }),
-          500
+          250
         )
         this.observer.unobserve(target)
       }
--- a/www/src/components/ecosystem/ecosystem-board.js
+++ b/www/src/components/ecosystem/ecosystem-board.js
@@ -1,4 +1,4 @@
-import React from "react"
+import React, { Component } from "react"
 import PropTypes from "prop-types"
 import styled from "react-emotion"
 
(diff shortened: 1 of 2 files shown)


www/src/components/ecosystem/ecosystem-board.js
```diff
@@ -20,48 +24,92 @@ const EcosystemBoardRoot = styled("div")`
...
+  }
+
+  setupObserver = () => {
+    const options = { rootMargin: "0px", threshold: [1] }
+    this.observer = new IntersectionObserver(this.handleIntersect, options)
+    this.observerTargets = Array.from(
+      document.querySelectorAll(".featuredItems")
+    )
+
+    for (const target of this.observerTargets) {
```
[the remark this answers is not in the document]
DSchau: http://kangax.github.io/compat-table/es6/#for..of_loops

www/src/components/ecosystem/ecosystem-board.js
```diff
@@ -20,48 +24,92 @@ const EcosystemBoardRoot = styled("div")`
...
+
+  componentWillUnmount() {
+    for (const target of this.observerTargets) {
+      this.observer.unobserve(target)
+    }
+  }
+
+  setupObserver = () => {
+    const options = { rootMargin: "0px", threshold: [1] }
+    this.observer = new IntersectionObserver(this.handleIntersect, options)
```
DSchau: If we just guard against this not existing, i.e. something like the below, I think this is worth doing!

```jsx
if (typeof window.IntersectionObserver !== 'undefined') {
  // regular setup
} else {
  this.observer = {};
  this.observerTargets = [];
}
```

www/src/components/ecosystem/ecosystem-board.js
```diff
@@ -20,48 +24,92 @@ const EcosystemBoardRoot = styled("div")`
...
+    this.observerTargets = Array.from(
+      document.querySelectorAll(".featuredItems")
```
DSchau: Wish there were a better way to do this, but this seems _OK_!

www/src/components/ecosystem/ecosystem-board.js
```diff
@@ -1,4 +1,8 @@
-import React from "react"
+if (typeof window !== "undefined") {
```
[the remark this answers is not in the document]
greglobinski: Without the polyfill the effect will not work on IOS, but sure we can

www/src/components/ecosystem/ecosystem-board.js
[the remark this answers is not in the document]
greglobinski: :+1:

www/src/components/ecosystem/ecosystem-board.js
```diff
@@ -20,48 +24,92 @@ const EcosystemBoardRoot = styled("div")`
...
-        links={[{ label: "Browse Resources", to: "/docs/awesome-gatsby/" }]}
-      />
-    </EcosystemBoardRoot>
-  )
+class EcosystemBoard extends Component {
+  observer
+  observerTargets = []
+
+  componentDidMount() {
+    if (typeof window !== `undefined`) {
```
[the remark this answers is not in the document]
greglobinski: With polyfill it's needed, without that we get error on build "WebpackError: ReferenceError: window is not defined"

www/src/components/ecosystem/ecosystem-board.js
greglobinski: sure, but maybe in componentDidMount
```
  componentDidMount() {
    if (typeof window.IntersectionObserver !== "undefined") {
      this.setupObserver()
    }
  }
```

DSchau: even better; yeah!

greglobinski:
@DSchau @amberleyromo The longer I think of it I'm more and more sure that I should strip out the intersection-observer and do it the old way, by listening to scroll event and comparing the offsets of page and elements. What  do you think? Did I make mistake?

DSchau:
@greglobinski it's totally doable, but personally, I prefer to treat a feature like this, which isn't strictly speaking _necessary_, as an enhancement. If it doesn't work in Safari, that's not _great_ but I don't think it's necessarily a huge issue nor a significantly diminished experience.

Additionally, I think manually wiring up intersection observer like behavior (with scroll listeners) would work, but I just think it would introduce some code that I personally wouldn't really want to maintain. I think the value of keeping the code pretty clean and concise with IntersectionObserver is worth the trade off of it not working in _every_ browser.

www/src/components/ecosystem/ecosystem-board.js
pieh: Potentially we could use `React.createRef` for here for `featuredItems` and pass it and use forwardRef down the road. Or move IntersectionObserver code to `ecosystem-featured-items.js`?

greglobinski: Moving IntersectionOBserver to `ecosystem-featured-items.js` will result in two instances of it. Is not more efficient to have only one ?

greglobinski:
@amberleyromo  /

* remove all "fake scroll"  staff (JS and CSS code manging animation styles) 
* implement real scrolling with ease function
* add some delay (250ms) before launching animation for better performance

www/src/components/ecosystem/ecosystem-board.js
pieh: I'm not sure honestly. We do create new IntersectionObserver instance for each `gatsby-image` instance, so maybe this is not a problem? (Or is a problem and we should look to fix that in gatsby-image)

greglobinski: Honestly, that's beyond my knowledge

amberleyromo: I think it (as is) is a reasonable way to handle this. If we have need to change it later, we can revisit.

Issue #9607, opened 2018-10-31
Troubles adding tags to the blog


d-ivashchuk:
## Summary
I am trying to add tags to my blog as in the following [tutorial](https://www.gatsbyjs.org/docs/adding-tags-and-categories-to-blog-posts/) but the path is not there in frontmatter. As far as I understand I shall add it manually to each blog, if so then tutorial lacks that in the example `.md` file.

Please correct me if I am wrong!

Thanks for help!

nayres:
Hey! What I would do at this point is make sure you are restarting your local server ```gatsby develop``` in order for those fields to be updated. Could you add some context to your request by showing the code involved?

d-ivashchuk:
Yeah I have some experience with Gatsby and I can pick up everything by using graphql. The main questions is about `path`, shall it be added to `frontmatter` manually or its something going behind the scenes in Gatsby. 
And the main thing that concerns me now is the example page, where there is following content of `.md`:

```
---
title: "A Trip To the Zoo"
tags: ["animals", "Chicago", "zoos"]
---

I went to the zoo today. It was terrible.
```
So you can see there is clearly no `path` in frontmatter here. I was able to manually put the path there and everything works, but less experienced people might assume its something that is there by default. If that's not a case and I am just not getting it, that's cool. Otherwise I would suggest updating example.

PR #9608, opened 2018-10-31
Added missing path field to frontmatter #9607

nayres:
https://github.com/gatsbyjs/gatsby/issues/9607

### DIFF
--- a/docs/docs/adding-tags-and-categories-to-blog-posts.md
+++ b/docs/docs/adding-tags-and-categories-to-blog-posts.md
@@ -31,6 +31,7 @@ Fields can be strings, numbers, or arrays. Since a post can usually have many ta
 
 ```md
 ---
+path: "/blog/a-trip-to-the-zoo"
 title: "A Trip To the Zoo"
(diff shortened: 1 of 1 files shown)


docs/docs/adding-tags-and-categories-to-blog-posts.md
```diff
@@ -142,8 +144,8 @@ export const pageQuery = graphql`
       edges {
         node {
           frontmatter {
-            title
             path
+            tags
```
pieh: this seems wrong - we use `title` in this example and don't use `tags`, I think this part shouldn't be changed

nayres: I just review the docs, and replicated issue #9607 . I retract my PR. Thanks @pieh .

DSchau:
Seems like this isn't needed per @nayres 

> I just review the docs, and replicated issue #9607 . I retract my PR. Thanks @pieh 

Closing, but please feel free to re-open if I missed something here. Thanks for the PR!

PR #9486, opened 2018-10-27
docs: delete docs/themes.md

pranshuchittora:
Docs about the themes were not clear.
:memo:  :+1:

PR #9613, opened 2018-10-31
Improves Page Not Found UX

pranshuchittora:
Recently I made some changes to the docs #9486 which included deleting one file. But that doc page has already been indexed by the search engines & when opening shows :point_down: 

screenshot from 2018-11-01 02-04-04

But there is not message or information for the user about what happened & what do to next.
So I made some changes to improve UX :+1:

### DIFF
--- a/www/src/pages/404.js
+++ b/www/src/pages/404.js
@@ -1,7 +1,7 @@
 import React from "react"
 import Container from "../components/container"
 import Layout from "../components/layout"
-
+import { Link } from "gatsby"
 class FourOhFour extends React.Component {
   render() {
     return (
--- a/www/src/pages/404.js
+++ b/www/src/pages/404.js
@@ -8,6 +8,12 @@ class FourOhFour extends React.Component {
       <Layout location={this.props.location}>
         <Container>
           <h1>Page not found</h1>
+          <p>
+            Oops! The page you are looking for has been removed or relocated.
+          </p>
+          <Link to="./">
+            <p>Go Back</p>
+          </Link>
         </Container>
       </Layout>
     )
--- a/www/src/pages/404.js
+++ b/www/src/pages/404.js
@@ -12,7 +12,7 @@ class FourOhFour extends React.Component {
           <p>
             Oops! The page you are looking for has been removed or relocated.
           </p>
-          <Link to="./">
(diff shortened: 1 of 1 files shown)


www/src/pages/404.js
```diff
@@ -8,6 +8,12 @@ class FourOhFour extends React.Component {
       <Layout location={this.props.location}>
         <Container>
           <h1>Page not found</h1>
+          <p>
+            Oops! The page you are looking for has been removed or relocated.
+          </p>
+          <a href="../">
```
LekoArts: When linking to internal pages you should use gatsby-link. Also the `to` prop then would be `"/"` not `../"`

www/src/pages/404.js
```diff
@@ -1,13 +1,19 @@
...
 class FourOhFour extends React.Component {
   render() {
     return (
       <Layout location={this.props.location}>
         <Container>
           <h1>Page not found</h1>
+          <p>
+            Oops! The page you are looking for has been removed or relocated.
+          </p>
+          <Link to="./">
```
DSchau:
```suggestion
          <Link to="/">
```

www/src/pages/404.js
```diff
@@ -1,13 +1,19 @@
 import React from "react"
 import Container from "../components/container"
 import Layout from "../components/layout"
-
+import { Link } from "gatsby"
 class FourOhFour extends React.Component {
```
DSchau:
```suggestion

class FourOhFour extends React.Component {
```

(adding a space!)

www/src/pages/404.js
DSchau: @pranshuchittora just need to accommodate this one, and I can get this merged!

www/src/pages/404.js
```diff
@@ -1,13 +1,20 @@
...
   render() {
     return (
       <Layout location={this.props.location}>
         <Container>
           <h1>Page not found</h1>
+          <p>
+            Oops! The page you are looking for has been removed or relocated.
+          </p>
+          <Link to="/">
+            <p>Go Back</p>
```
KyleAMathews: Why is this "go back" not "go to frontpage" or something equivalent? This doesn't take you back anywhere

PR #9618, opened 2018-10-31
add tutorial for gatsby+contentful+netlify

cwlsn:
Hey folks,

Just adding an article I wrote under the tutorial section. It's about creating a starter hooked up to Contentful, then publishing to Netlify. The starter from the article is already in the starter library!


shannonbux:
Awesome! This is popular trio of technologies :)

PR #9622, opened 2018-11-01
Remove eslint-plugin-flow-vars

alexandernanberg:
Removes [`eslint-plugin-flow-vars`](https://github.com/zertosh/eslint-plugin-flow-vars) since it's been deprecated and integrated in `eslint-plugin-flowtype`


examples/using-gatsby-source-graphql/package.json
```diff
@@ -22,7 +22,6 @@
     "eslint": "^4.19.1",
     "eslint-config-google": "^0.9.1",
     "eslint-config-prettier": "^2.9.0",
-    "eslint-plugin-flow-vars": "^0.5.0",
```
DSchau: I'm really not sure why _any_ of these eslint dependencies are in this example, since it doesn't seem to be used whatsoever, but oh well!

PR #9623, opened 2018-11-01
Remove background, box-shadow and border-bottom from author image (9593)

### LINKED ISSUE #9593: Image as link has horizontal line in background
## Description

The authors image in a blogpost is a link. It inherits link tag stylings which draws a border in the background. 

### Steps to reproduce

Open a [blogpost](https://www.gatsbyjs.org/blog/2018-10-16-why-mobile-performance-is-crucial/) and hover over the authors image.

This is how it looks.
bildschirmfoto 2018-10-31 um 08 45 28

### Expected result

No borders in background for images in link tags.

### Actual result

Border in background.

### Environment

production.

gurtjun:
Fixes #9593

### DIFF
--- a/www/src/templates/template-blog-post.js
+++ b/www/src/templates/template-blog-post.js
@@ -149,11 +149,13 @@ class BlogPostTemplate extends React.Component {
                 <Link
                   to={post.frontmatter.author.fields.slug}
                   css={{
-                    borderBottom: `0 !important`,
-                    boxShadow: `none !important`,
-                    "&:hover": {
-                      background: `none !important`,
-                    },
+                    width:
+                      post.frontmatter.author.avatar.childImageSharp.fixed
+                        .width,
+                    height:
+                      post.frontmatter.author.avatar.childImageSharp.fixed
+                        .height,
+                    borderRadius: `100%`,
                   }}
                 >
                   <Img
--- a/www/src/templates/template-blog-post.js
+++ b/www/src/templates/template-blog-post.js
@@ -149,13 +149,13 @@ class BlogPostTemplate extends React.Component {
                 <Link
                   to={post.frontmatter.author.fields.slug}
                   css={{
-                    width:
-                      post.frontmatter.author.avatar.childImageSharp.fixed
-                        .width,
-                    height:
-                      post.frontmatter.author.avatar.childImageSharp.fixed
-                        .height,
-                    borderRadius: `100%`,
+                    "&&": {
+                      borderBottom: 0,
+                      boxShadow: `none`,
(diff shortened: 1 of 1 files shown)


www/src/templates/template-blog-post.js
```diff
@@ -146,7 +146,16 @@ class BlogPostTemplate extends React.Component {
                   flex: `0 0 auto`,
                 }}
               >
-                <Link to={post.frontmatter.author.fields.slug}>
+                <Link
+                  to={post.frontmatter.author.fields.slug}
+                  css={{
+                    borderBottom: `0 !important`,
```
DSchau: Do all of these _need_ `!important`? I wouldn't have thought so, they shouldn't need extra specificity here.

gurtjun: Without the `!important` it's overridden by other CSS. Not sure how we can set the emotion-js inline CSS as default.

DSchau: It's here -> https://github.com/gatsbyjs/gatsby/blob/master/www/src/utils/typography.js#L169

So (annoyingly) what's happening here is that .main-body a is _more specific_ than what emotion is doing, which will be like `.asdf-1234 { borderBottom: 0; }`

Can we solve this problem without using `!important`? One way you _should_ be able to do it is hoist the css up a level, and just use something like

```js
css={{
  a: {
    borderBottom: 0,
    boxShadow: 'none',
    ':hover': {
      background: 'none'
    }
  }
}}
```

DSchau: I guess `!important` is fine too. It just sorta negates some of the benefit of using CSS in JS in the first place, hah.

gurtjun: Another solution might be using the size of the avatar image for the link. I updated the code, no `!important` needed anymore. 
If this is ok? 😄

fk: Most of the times something like https://github.com/gatsbyjs/gatsby/blob/fc81ecaa4e4b2b0b0b94dfb9337aad5866569726/www/src/components/package-readme.js#L34-L49 should work fine, e.g. 

```javascript
css={{
  "&&": {
  ...styles where specificity needs to be increased
}}
´´´

See https://github.com/emotion-js/emotion/issues/724#issue-332088759 for another example + generated `className`.

DSchau: Nice trick @fk!! @gurtjun would you want to try that?

gurtjun: Good catch @fk! That seems to work. Used the ampersands in f66f114

PR #9633, opened 2018-11-01
example-using-remark reject Promise on error

benjaminabel:

### DIFF
--- a/examples/using-remark/gatsby-node.js
+++ b/examples/using-remark/gatsby-node.js
@@ -33,7 +33,7 @@ exports.createPages = ({ graphql, actions }) => {
       `
     ).then(result => {
       if (result.errors) {
-        console.log(result.errors)
+        return reject(result.errors)
       }
 
       // Create blog posts pages.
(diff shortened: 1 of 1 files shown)


DSchau:
@benjaminabel can you explain why this change is needed? To me, it looks like the promise will reject now, and it won't be clear _why_ something is failing?

I agree it should probably reject, though.

benjaminabel:
@DSchau I thought that including the `result.errors` in the rejection message, would work. However if you think it should reject and log errors separately it could be done like this.
```js
console.error(result.errors)
reject()
```
But i think it is redundant.

DSchau:
@benadam11 [here's an example](https://runkit.com/dschau/5bdb5f86be2d95001241ac4a)

Also if anything, we should log and then `return reject` in order to stop the control flow. You can see in the example above that the console.log is still triggered.

benjaminabel:
You're right, should return reject, I found this in [gatsby-node](https://github.com/gatsbyjs/gatsby/blob/master/www/gatsby-node.js#L239) of offical site www. I edit the PR.

PR #9641, opened 2018-11-01
Edited for punctuation and style guide

Janiceilene:
Mostly edited punctuation. Also, changed an abbreviation to match the style guide.

### DIFF
--- a/docs/tutorial/part-zero/index.md
+++ b/docs/tutorial/part-zero/index.md
@@ -7,23 +7,23 @@ Before you start to code, you’ll need to familiarize with some core web techno
 
 ## Overview of core technologies
 
-It’s not necessary to be an expert with these already — if you’re not, don’t worry! You’ll pick up a lot through the course of this tutorial series; These are some of the main web technologies you’ll use when building a Gatsby site:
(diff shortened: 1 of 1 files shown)


shannonbux:
Hm, I just realized this document uses different ways of saying "Gatsby CLI" "gatsby-cli" etc., and shifts between command line and CLI. We ought to stick with a consistent name for Gatsby CLI and then after providing the acronym CLI once, stick with it. Want to tackle that @Janiceilene ?

Janiceilene:
@shannonbux I noticed that, but didn't want to over edit or step on anyone's toes. Is the acronym one full explanation per doc or section? Would it be helpful to go through and edit for things like concise language?

shannonbux:
Hi @Janiceilene! Thanks! Yes, very helpful! I thought through the Gatsby Style Guide super carefully and then get really excited whenever people start implementing it.

As far as how frequently to explain the acronym, I would say that for the tutorial, just once is good enough, since we are optimizing the experience for people who go through the whole tutorial.

For the docs, we can assume that everyone has gone through the tutorial and refrain from re-explaining any acronyms that have already been explained in the tutorial.

Issue #9640, opened 2018-11-01
Not able to start dev server in /www 


pranshuchittora:
Today I tried to run the instance of gatsby website locally.
It stucks at `source and transform nodes` :disappointed: 
Pls refer this screenshot :point_down: 
![screenshot from 2018-11-02 01-25-09](https://user-images.githubusercontent.com/32242596/47877617-516b6280-de42-11e8-8d64-a77d6fde7781.png)

DSchau:
@pranshuchittora yeah, it's a little unfortunate. In general, the site showcase (and taking screenshots) is the bulk of the time spent there. It'd eventually complete, just takes quite a while.

Can you [try this setting](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-transformer-screenshot#placeholder-image)? The general idea is you'd create `.env.development` and then add `GATSBY_SCREENSHOT_PLACEHOLDER=true`

DSchau:
Sounds like that worked, so closing out!

pranshuchittora:
Thanks @DSchau it working :+1: 
![screenshot from 2018-11-02 02-30-19](https://user-images.githubusercontent.com/32242596/47880466-5849a380-de49-11e8-811b-36d2d6b247d9.png)

pranshuchittora:
Consider these updates in the docs #9643  :+1:

PR #9642, opened 2018-11-01
Updates how-to-contribute docs

pranshuchittora:
Recently I tried to run the instance of gatsby's website locally. 
By using `gatsby develop` in /www, unfortunately, it got stuck. Refer #9640 for more info.
Thanks to @DSchau he suggested trying `GATSBY_SCREENSHOT_PLACEHOLDER=true` to run the dev server. But for a new contributor, it's hard for him to figure this out. 

So I made some changes in the `how-to-contribute` docs.

### DIFF
--- a/www/src/pages/404.js
+++ b/www/src/pages/404.js
@@ -1,7 +1,7 @@
 import React from "react"
 import Container from "../components/container"
 import Layout from "../components/layout"
-
+import { Link } from "gatsby"
 class FourOhFour extends React.Component {
   render() {
     return (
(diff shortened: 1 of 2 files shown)


www/src/pages/404.js
```diff
@@ -1,13 +1,19 @@
 import React from "react"
 import Container from "../components/container"
 import Layout from "../components/layout"
-
+import { Link } from "gatsby"
```
DSchau: 🤔 Don't think we want this file!

PR #9644, opened 2018-11-01
[WIP]: feat: add script and circleci job to sync examples to read-only repos

DSchau:
WIP WIP WIP WIP

This simple script will check the most recent commit's file changes
(i.e. on master push) vs. the previous state of the repo. It then
iterates over those files (if they match a given pattern, e.g.
examples/*) and will push to those repo clones. We can re-use this script later if we add starters/* to this monorepo.

This will give us some nice benefits, e.g. tracking on forks/stars/etc.
for our examples, starters, etc. while keeping the code in our monorepo,
but also providing a means for others to access those repos outside of
the monorepo

cc @amberleyromo @pieh @KyleAMathews

### DIFF
--- a/scripts/sync-repos.sh
+++ b/scripts/sync-repos.sh
@@ -1,17 +1,17 @@
 #!/bin/bash
-FOLDER_PATTERN="${1:-examples}" # grab first argument, otherwise use examples folder
+BASE_FOLDER="${1:-examples}" # grab first argument, otherwise use examples folder
 
 CURRENT_COMMIT="$(git log --format="%H" -n 1)"
 CURRENT_COMMIT_MESSAGE="$(git log --format="%s" -n 1)"
 LAST_COMMIT="$(git log --format="%H" -n 2 | tail -1)"
 
-CHANGED_FILES="$(git diff-tree --no-commit-id --name-only -r $CURRENT_COMMIT $LAST_COMMIT | grep -E "$FOLDER_PATTERN/*")"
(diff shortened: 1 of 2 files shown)


scripts/sync-repos.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+FOLDER_PATTERN="${1:-examples}" # grab first argument, otherwise use examples folder
+
+CURRENT_COMMIT="$(git log --format="%H" -n 1)"
+CURRENT_COMMIT_MESSAGE="$(git log --format="%s" -n 1)"
+LAST_COMMIT="$(git log --format="%H" -n 2 | tail -1)"
+
+CHANGED_FILES="$(git diff-tree --no-commit-id --name-only -r $CURRENT_COMMIT $LAST_COMMIT | grep -E "$FOLDER_PATTERN/*")"
+
+# TODO: figure out secure way to push to github
```
DSchau: and _create_ the repos

scripts/sync-repos.sh
```diff
@@ -0,0 +1,22 @@
...
+for file in $CHANGED_FILES
+do
+  FOLDER="$(echo $file | cut -d'/' -f2)"
+  cd examples/$FOLDER
+
+  git init
```
DSchau: Might not need this 🤷‍♂️ I'll test this further

DSchau:
I’ll have to tweak this so that I only init and push once per folder. In other words, if an example had multiple file changes, this would init and push multiple times

DSchau:
Closing this. I'll follow an approach (but in this monorepo) like dschau/starters

Issue #9375, opened 2018-10-24
Tutorial: Add list of official plugins and official starters to end of tutorial


shannonbux:
## Summary

We would like to link to a filtered list of
- official plugins
- official starters

at the end of the main tutorial.

### Motivation

These links would be a great segue into helping new Gatsby users continue to explore and find plugins and starters that fit their use case.

the plugin library and starter library will have to include filters before we can easily implement this.
^ @amberleyromo

Soundug:
Hi! Can I take this up? 
Also, which plugins should i be adding to the list. Any reference?

shannonbux:
@Soundug 

Official Plugins: https://github.com/gatsbyjs/gatsby/tree/master/packages
Official Starters: I can't remember? @jlengstorf?

jlengstorf:
These are our official starters:
- https://github.com/gatsbyjs/gatsby-starter-default
- https://github.com/gatsbyjs/gatsby-starter-blog
- https://github.com/gatsbyjs/gatsby-starter-hello-world

jlengstorf:
Good call, @oorestisime! Thanks!

PR #9646, opened 2018-11-01
Tutorial update

Yash-Handa:
Hi @jlengstorf and @shannonbux,
I added **Official Plugins and Starters section** to the main Tutorial Page.
In accordance with the issue #9375.
I hope it might help : )


amberleyromo:
@shannonbux i think this is a fine way to handle it until we build the filter into the plugin library and starter library

Issue #9559, opened 2018-10-30
Upgrading gatsby from 2.0.31 to 2.0.34 causes a Typerror


jonleopard:
Not sure what's causing this, it may be a plugin but I'm not sure where to even start.

## Description

Upgrading gatsby from 2.0.31 to 2.0.34 causes a typerror:
```
$ gatsby develop                                                                                                                                               [75/2399]
success open and validate gatsby-configs — 0.011 s
error value must be an array of bytes


  TypeError: value must be an array of bytes

  - v35.js:29 generateUUID
    [jonleopard.com]/[uuid]/lib/v35.js:29:38

  - create-node-id.js:16 createNodeId
    [jonleopard.com]/[gatsby]/dist/utils/create-node-id.js:16:10

  - load.js:131 processPlugin
    [jonleopard.com]/[gatsby]/dist/bootstrap/load-plugins/load.js:131:13

  - load.js:148 config.plugins.forEach.plugin
    [jonleopard.com]/[gatsby]/dist/bootstrap/load-plugins/load.js:148:20

  - Array.forEach

  - load.js:147 module.exports
    [jonleopard.com]/[gatsby]/dist/bootstrap/load-plugins/load.js:147:20

  - index.js:56
    [jonleopard.com]/[gatsby]/dist/bootstrap/load-plugins/index.js:56:21

  - Generator.next

  - debuggability.js:313 Promise._execute
    [jonleopard.com]/[bluebird]/js/release/debuggability.js:313:9

  - promise.js:483 Promise._resolveFromExecutor
    [jonleopard.com]/[bluebird]/js/release/promise.js:483:18

  - promise.js:79 new Promise
    [jonleopard.com]/[bluebird]/js/release/promise.js:79:10

  - index.js:96
    [jonleopard.com]/[gatsby]/dist/bootstrap/load-plugins/index.js:96:17


error UNHANDLED REJECTION


  TypeError: value must be an array of bytes

  - v35.js:29 generateUUID
    [jonleopard.com]/[uuid]/lib/v35.js:29:38

  - create-node-id.js:16 createNodeId
    [jonleopard.com]/[gatsby]/dist/utils/create-node-id.js:16:10

  - load.js:131 processPlugin
    [jonleopard.com]/[gatsby]/dist/bootstrap/load-plugins/load.js:131:13

  - load.js:148 config.plugins.forEach.plugin
[...]

PR #9647, opened 2018-11-01
fix: load plugin object style without options object

zvinless:
Fixes issue at the end of #9559, e.g.:
```javascript
module.exports = {
    plugins: [
        {
            resolve: 'gatsby-plugin-react-helmet'
        }
    ]
}
```
where the plugin is loaded using object syntax but no `options` object is defined.

This wasn't caught by the unit test because the issue was occurring after the early out when it's the `___TEST___` plugin.

### DIFF
--- a/packages/gatsby/src/bootstrap/load-plugins/load.js
+++ b/packages/gatsby/src/bootstrap/load-plugins/load.js
@@ -102,9 +102,11 @@ module.exports = (config = {}) => {
         },
       }
     } else {
+      plugin.options = plugin.options || {}
+
       // Plugins can have plugins.
       const subplugins = []
-      if (plugin.options && plugin.options.plugins) {
+      if (plugin.options.plugins) {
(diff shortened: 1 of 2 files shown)


packages/gatsby/src/bootstrap/load-plugins/load.js
```diff
@@ -130,12 +132,7 @@ module.exports = (config = {}) => {
         // Make sure key is unique to plugin options. E.g there could
         // be multiple source-filesystem plugins, with different names
         // (docs, blogs).
-        id: createNodeId(
-          plugin.options
-            ? plugin.name + JSON.stringify(plugin.options)
```
zvinless: Is `plugin.name` undefined here? Was it intended to use `info.name` or `plugin.resolve`?

pieh: I think it should be `info.name`

PR #9657, opened 2018-11-02
fix(gatsby-remark-autolink-headers): use shouldUpdateScroll api and not onRouteUpdate

### LINKED ISSUE #9638: [www] Tutorial sidebar links don't correctly jump to referenced element
* Open https://www.gatsbyjs.org/tutorial/part-one/#using-gatsby-starters — viewport correctly scrolls to "Using Gatsby starters" headline ✅ 
* Now click the "Using Gatsby starters" link in the sidebar — notice that the viewport scrolls down a few pixels, moving the "Using Gatsby starters" headline below the header/banner 🔴 
* Click any other link in the Tutorial sidebar and observe the same behavior — links pointing to anchors on the same page do not seem to behave correctly
* Exactly the same behavior can be observed on e.g. https://reactjs.org/tutorial/tutorial.html#help-im-stuck

/cc @pieh

pieh:
right now autolink-headers scroll is battling with scrollbehaviour and which should handle scroll to `#elementId`. This allow `autolink-headers` plugin to hook into scrollbehaviour to override default behaviour.

Potential problems: `shouldUpdateScroll` is API that handles only one instance of this hook, so using it in autolink-headers plugin will discard results of this hook in user code (if it's used there)

closes #9638

### DIFF
--- a/packages/gatsby-remark-autolink-headers/src/gatsby-browser.js
+++ b/packages/gatsby-remark-autolink-headers/src/gatsby-browser.js
@@ -1,7 +1,7 @@
 let offsetY = 0
 
 const getTargetOffset = hash => {
-  const id = window.decodeURI(window.location.hash.replace(`#`, ``))
+  const id = window.decodeURI(hash.replace(`#`, ``))
   if (id !== ``) {
     const element = document.getElementById(id)
     if (element) {


packages/gatsby-remark-autolink-headers/src/gatsby-browser.js
```diff
@@ -1,35 +1,30 @@
...
-  setTimeout(() => {
-    const hash = window.decodeURI(window.location.hash.replace(`#`, ``))
-    if (hash !== ``) {
-      const element = document.getElementById(hash)
-      if (element) {
-        const offset = element.offsetTop
-        window.scrollTo(0, offset - offsetY)
-      }
+const getTargetOffset = hash => {
+  const id = window.decodeURI(window.location.hash.replace(`#`, ``))
```
alexandernanberg: Might be missing something – but shouldn't this be `hash` instead of `window.location.hash` here?

DSchau: Seems reasonable! window.location.hash is already being passed here, so seems like you're right!

pieh: Ah, thanks!

PR #9659, opened 2018-11-02
chore: add starters to monorepo

DSchau:
This PR adds starters to gatsbyjs/gatsby repo. The general idea is that
these are going to be the official, de facto source, and we'll change
the various repos (e.g. gatsbyjs/gatsby-starter-default) to be read only
clones that are synced with this repo.

Related to #9644 (_this_ PR should probably be merged first, and then we
can try it out with the starters)


starters/gatsby-starter-blog/.travis.yml
```diff
@@ -0,0 +1,15 @@
+language: node_js
```
DSchau: 🤔 I'd like to retain this functionality. We can follow a similar approach with CircleCI here, and only run tests if these files were changed, but it just adds a bit of complexity.

starters/gatsby-starter-blog/package.json
```diff
@@ -0,0 +1,64 @@
+{
+  "name": "gatsby-starter-blog",
+  "description": "Starter Gatsby Blog",
+  "version": "1.0.0",
+  "author": "Kyle Mathews <mathews.kyle@gmail.com>",
+  "bugs": {
+    "url": "https://github.com/gatsbyjs/gatsby-starter-blog/issues"
+  },
+  "dependencies": {
+    "gatsby": "^2.0.19",
```
DSchau: Not really sure of what our approach is for keeping these up to date, but we'll need to make sure it still works with the monorepo approach.

DSchau:
I want to think more about this 🤔 

Closing for now.

PR #9517, opened 2018-10-29
draft: introducing gatsby themes

ChristopherBiscardi:
Introductory blog post for initial theming support

PR #9663, opened 2018-11-02
Change theme spec to match plugin spec

ChristopherBiscardi:
This changes `__experimentalThemes` from the following

```
__experimentalThemes: [['theme-name', optionsObject]]
```

to match the plugins spec

```
__experimentalThemes: [{
  resolve: 'theme-name',
  options: {}
}]
```

* This is to reduce the number of user-visible config styles.
* There will be a future change to be fully compatible with the [load plugins](https://github.com/gatsbyjs/gatsby/blob/390803c8f679e53d398d40b9535c1a8784380a48/packages/gatsby/src/bootstrap/load-plugins/load.js#L37-L85) functionality (defines a `themes/` directory for local themes and handles absolute paths, etc). This change is meant to match the plugins API spec superficially so that the blogpost #9517 doesn't have syntax incompatible with the future.

### DIFF
--- a/packages/gatsby/src/bootstrap/index.js
+++ b/packages/gatsby/src/bootstrap/index.js
@@ -81,7 +81,7 @@ module.exports = async (args: BootstrapArgs) => {
     const themesConfig = await Promise.mapSeries(
       config.__experimentalThemes,
       async plugin => {
-        const resolve = plugin.resolve || plugin
+        const themeName = plugin.resolve || plugin
         const themeConfig = plugin.options || {}
         const theme = await preferDefault(
           getConfigFile(themeName, `gatsby-config`)
--- a/packages/gatsby/src/bootstrap/index.js
+++ b/packages/gatsby/src/bootstrap/index.js
@@ -80,7 +80,7 @@ module.exports = async (args: BootstrapArgs) => {
   if (config && config.__experimentalThemes) {
     const themesConfig = await Promise.mapSeries(
       config.__experimentalThemes,
-      async ([themeName, themeConfig]) => {
+      async ({ resolve: themeName, options: themeConfig }) => {
         const theme = await preferDefault(
           getConfigFile(themeName, `gatsby-config`)
         )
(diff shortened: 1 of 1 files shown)


ChristopherBiscardi:
@DSchau I totally agree. This is intentionally half-way so that the blogpost can go out with this syntax. I have the full abstraction in some different work locally that I'm going to PR along with the Component Shadowing (which is functionality that it makes more sense to be committed with)

DSchau:
@ChristopherBiscardi I think there's some value in getting this out there (with this syntax) but I'd also rather have the plugin API fully supported, just because I can see many being confused that only one API format is supported in themes.

What do you think? Should we close this and wait until the full plugin API is supported? Also I am happy to assist if you have some work in a branch somewhere, too!

packages/gatsby/src/bootstrap/index.js
```diff
@@ -80,7 +80,7 @@ module.exports = async (args: BootstrapArgs) => {
   if (config && config.__experimentalThemes) {
     const themesConfig = await Promise.mapSeries(
       config.__experimentalThemes,
-      async ([themeName, themeConfig]) => {
+      async ({ resolve: themeName, options: themeConfig }) => {
```
DSchau: Why don't we just resolve this in a "hacky" way (for now!) and do something like this?

```js
if (config && config.__experimentalThemes) {
  const themesConfig = await Promise.mapSeries(
    config.__experimentalThemes,
    async plugin => {
      const pluginName = plugin.resolve || plugin
      const themeConfig = plugin.options || {}
    }
  )
}
```

then we can totally avoid the issue of API idiosyncrasies, right?

ChristopherBiscardi: I updated the code to have this effect

DSchau:
@ChristopherBiscardi merged (with a little tweak). Thanks!

ChristopherBiscardi:
👍 good catch

PR #9664, opened 2018-11-02
docs: removed duplicated 'to' from the docs

Parikshit-Hooda:
removed extra 'to' from the docs


DSchau:
@Parikshit-Hooda looks like you used master as a base for this PR, but then opened it up against v1. Do you want this to be v1, or vs. master?

pieh:
> Do you want this to be v1, or vs. master?

It should be `master` - it's for unstructured data page which is not in v1 branch

Parikshit-Hooda:
Holy smokes! Thanks for the org invite and free swag. I am already in love.

PR #9669, opened 2018-11-02
fix(gatsby): correctly pick up browserslist overrides

### LINKED ISSUE #7003: [v2] IE11 throws Objects are not valid as a React child
## Description

When visiting sites using v2 in IE11 React throws an error

### Steps to reproduce

Visit https://next.gatsbyjs.org in IE11 and check the console

### Expected result

No errors

### Actual result

React throws this error;
```
Objects are not valid as a React child (found: object with keys {$$typeof, type, key, ref, props, _owner}).
```

### Environment

```
  System:
    OS: macOS High Sierra 10.13.6
    CPU: x64 Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 8.11.3 - ~/.nvm/versions/node/v8.11.3/bin/node
    Yarn: 1.9.2 - /usr/local/bin/yarn
    npm: 5.6.0 - ~/.nvm/versions/node/v8.11.3/bin/npm
  Browsers:
    Chrome: 68.0.3440.84
    Firefox: 61.0.1
    Safari: 11.1.2
  npmPackages:
    gatsby: ^2.0.0-beta.64 => 2.0.0-beta.67 
    gatsby-plugin-catch-links: ^2.0.2-beta.5 => 2.0.2-beta.5 
    gatsby-plugin-feed: next => 2.0.0-beta.4 
    gatsby-plugin-glamor: next => 2.0.0-beta.3 
    gatsby-plugin-google-analytics: next => 2.0.0-beta.3 
    gatsby-plugin-manifest: next => 2.0.2-beta.3 
    gatsby-plugin-netlify: next => 2.0.0-beta.5 
    gatsby-plugin-nprogress: next => 2.0.0-beta.4 
    gatsby-plugin-react-helmet: next => 3.0.0-beta.4 
    gatsby-plugin-sharp: next => 2.0.0-beta.7 
    gatsby-plugin-twitter: next => 2.0.0-beta.3 
    gatsby-remark-autolink-headers: next => 2.0.0-beta.5 
    gatsby-remark-code-repls: next => 2.0.0-beta.4 
    gatsby-remark-copy-linked-files: next => 2.0.0-beta.3 
    gatsby-remark-embed-snippet: next => 3.0.0-beta.3 
    gatsby-remark-images: next => 2.0.1-beta.9 
    gatsby-remark-prismjs: next => 3.0.0-beta.5 
    gatsby-remark-responsive-iframe: next => 2.0.0-beta.3 
    gatsby-remark-smartypants: next => 2.0.0-beta.3 
    gatsby-source-filesystem: next => 2.0.1-beta.10 
    gatsby-transformer-remark: next => 2.1.1-beta.5 
    gatsby-transformer-sharp: next => 2.1.1-beta.6 

```

`gatsby-config.js`: N/A 
`package.json`: N/A 
`gatsby-node.js`: N/A 
`gatsby-browser.js`: N/A 
`gatsby-ssr.js`: N/A

DSchau:
Just to get some 👀 on this early, this is somewhat working. It's fairly challenging to test this, so I've been using [`gatsby-plugin-webpack-bundle-analyzer`](https://www.npmjs.com/package/gatsby-plugin-webpack-bundle-analyzer) and inspecting the size of the bundled assets. Currently, can't seem to get it to change with the browserslist query, but it _is_ picked up (correctly) and added to `.cache/babelState.json`.

More to do here 🙃 

Begins to fix #7003, in that enabling browserslist functionality should clean up IE specific stuff.


packages/babel-preset-gatsby/src/index.js
```diff
@@ -25,16 +25,14 @@ function preset(context, options = {}) {
         node: `current`,
       }
     } else {
-      targets = {
-        browsers: pluginBabelConfig.browserslist,
```
DSchau: > Note: this will be removed in later version in favor of just setting "targets" to a query directly.

https://babeljs.io/docs/en/babel-preset-env#targetsbrowsers

packages/gatsby-cli/src/create-cli.js
```diff
@@ -20,14 +22,14 @@ function buildLocalCommands(cli, isLocalSite) {
...
-      : [`>0.25%`, `not dead`]
+      : [`> 0.25%`, `not dead`]
 
   let siteInfo = { directory, browserslist: DEFAULT_BROWSERS }
   const useYarn = existsSync(path.join(directory, `yarn.lock`))
   if (isLocalSite) {
     const json = require(path.join(directory, `package.json`))
     siteInfo.sitePackageJson = json
-    siteInfo.browserslist = json.browserslist || siteInfo.browserslist
+    siteInfo.browserslist = getBrowsersList(directory, DEFAULT_BROWSERS)
```
DSchau: This seems kind of weird to have the set up working like this, i.e. gatsby-cli _sets up_ some program stuff, and then other plugins use it. Might try to improve this.

KyleAMathews: Yeah, can't remember why gatsby-cli does this

DSchau:
Alright, so this PR is ready, although I plan to make the improvements to not requiring browserslist in gatsby-cli--which seems weird.

After testing, this change does work like we'd expect, but it appears there's at least one weird thing.

We use `useBuiltIns: 'usage'`, which means that babel will intelligently detect what polyfills it needs to load from JS, which works great for our source code. You can add something in the src/ directory, e.g. `new Set()` and then use `webpack-bundle-analyzer` to see the collections utility from core-js get added to the bundle. This will be fixed and work like we'd expect in this PR.

Currently the only check we were making was with the users package.json `browserslist` key, rather than the order of checks that browserslist actually uses, e.g. env var, browserslistrc file, package.json key, etc.

However - the gotcha is that we don't parse node_modules with babel. This means that useBuiltIns can't know what polyfills we could need to include for e.g. some third party dep. It may be worth doing that in the future, but it doesn't seem like other projects are doing that quite yet. Might be nice to make this opt-in with a plugin or something in the future.

DSchau:
Just need to fix the tests 😅 Doing that now.

KyleAMathews:
CRA v2 now compiles node_modules https://github.com/facebook/create-react-app/issues/1125

DSchau:
@KyleAMathews I knew I saw something about that! That's probably slightly out of scope for this PR, but that could be worth following a similar approach to how they've done it!

KyleAMathews:
Sounds good!

DSchau:
Also @souporserious this _may_ fix the issue you were running into in development mode! It's kinda hard to validate Windows stuff 😅

Any interest in checking it out?

pieh:
screen shot 2018-11-15 at 00 13 42
IE still won't work with `gatsby develop` - it fails in `react-hot-loader` currently, but this PR does fix `browserlist`, so let's merge it!

PR #9670, opened 2018-11-02
chore(prettier): upgrade prettier to v1.16

alexandernanberg:
Upgrades prettier to the latest version (1.16.1)


pieh:
Also azure CI seems to fail - it might be that prettier did change something about newlines (LF vs CRLF) strategy (I didn't read up on changes they did in 1.15 yet)

alexandernanberg:
@pieh Completely missed that they added an option for [newlines (endOfLine)](https://prettier.io/docs/en/options.html#end-of-line). Looks like that solved the issue, not sure what's up with the e2e test tho

pieh:
> @pieh Completely missed that they added an option for [newlines (endOfLine)](https://prettier.io/docs/en/options.html#end-of-line). Looks like that solved the issue, not sure what's up with the e2e test tho

I didn't check with prettier 1.15, but prettier seems to to do sometimes weird things on 1.14 - it will occasionally change CRLF to LF and then fail because those endlines do not match `¯\_(ツ)_/¯`. I think easiest way around is what You did with forcing LF in `.gitattributes` but it potentially can cause some problems for windows devs for their working directory (that's why we were applying more targeted rules instead of slapping forced LF for all text files). Ultimately we will need to do this, so I'm fine with applying it here.

alexandernanberg:
Yeah I think that will be solved with 1.15 because you can force the line endings to LF. I think it used to guess what line ending is correct. I see, what kind of problems can it cause tho? I got the impression that it should work fine.

From the prettier site
 > Ask Windows users to run git config core.autocrlf false before working on your repo so that git did not convert LF to CRLF on checkout. Alternatively, you can add * text=auto eol=lf to the repo's .gitattributes file to achieve this.

alexandernanberg:
@pieh Is there something currently blocking this or is it good to go? 😄

.gitattributes
```diff
@@ -1,4 +1 @@
-**/*.js.snap text eol=lf
-**/__testfixtures__/** text eol=lf
-**/__tests__/fixtures/** text eol=lf
-**/*.md text eol=lf
\ No newline at end of file
+* text=auto eol=lf
```
DSchau: @pieh what do you think of this? I kinda wanted to do this from the beginning, but I can't remember if there was a reason we were so selective?

.prettierrc
```diff
@@ -1,6 +1,7 @@
 {
-    "semi": false,
-    "singleQuote": false,
-    "tabWidth": 2,
-    "trailingComma": "es5"
+  "endOfLine": "lf",
```
DSchau: 👍

docs/blog/2017-11-09-why-i-created-my-blog-with-gatsby-and-contentful/index.md
```diff
@@ -57,7 +57,7 @@ If you are a little bit more tech savvy and have experience provisioning
 software in a Linux distribution, you might prefer to rent a droplet from
 Digital Ocean and host the blog yourself. This, in turn, will allow you to have
 greater control over your site. However, you will also to pay for that droplet,
-even if it's the cheaper $5/month one.
+even if it's the cheaper \$5/month one.
```
DSchau: A bug in the new version of prettier apparently? We'd maybe _not_ upgrade if this is the case--until it's fixed.

docs/docs/unit-testing.md
```diff
@@ -32,22 +32,23 @@ npm install --save-dev jest babel-jest react-test-renderer identity-obj-proxy 'b
 Because Gatsby handles its own Babel configuration, you will need to manually
 tell Jest to use `babel-jest`. The easiest way to do this is to add a `jest.config.js`. You can set up some useful defaults at the same time:
 
-```json:title=jest.config.js
+```js:title=jest.config.js
```
DSchau: Not sure if you caught this--I'd be shocked if prettier did?--but... you're right!

docs/blog/2017-11-09-why-i-created-my-blog-with-gatsby-and-contentful/index.md
DSchau: Although actually it seems to be displayed exactly as is (e.g. as if it wasn't escaped). I wonder _why_ this change was made 🤔

.gitattributes
pieh: This is the main (and I think only) reason - https://help.github.com/articles/dealing-with-line-endings/#refreshing-a-repository-after-changing-line-endings, but we will need to do it eventually, so might as well rip band-aid off now

DSchau: That sucks so much, ha. But yeah, this is clearer and more maintainable going forward I suppose.

pieh:
> @pieh Is there something currently blocking this or is it good to go? 😄

I don't think anything is blocking other than time needed to check if everything works correctly with those changes (i.e. if escaped `$` are displaying correctly on built site) - this doesn't seem to fix any problem in particular (other than some packages not being up to date), so I wasn't prioritising it.

docs/blog/2017-11-09-why-i-created-my-blog-with-gatsby-and-contentful/index.md
alexandernanberg: Yeah I'm not 100% if this was intentional from prettier or not, can't find anything in the release notes about this other than https://prettier.io/blog/2018/11/07/1.15.0.html#preserve-math-syntax-5050-5220-by-ikatyang. Relevant issue https://github.com/prettier/prettier/issues/5483

docs/docs/unit-testing.md
alexandernanberg: Yeah it was me 😛 Prettier would add parentheses around the block otherwise, eg. 
```(module.exports = {...})```

alexandernanberg:
> OK so personally this looks good. I do think your comment re: removing prettier (and changes) from this PR is a good one.
Would you be up for that, and then we can maybe do a subsequent PR just discussing the prettier changes?

I think it would be easier to keep the prettier updates here and discard the other stuff and create another PR for that. Almost everything in this PR is related to prettier anyways 😛  

> I don't think anything is blocking other than time needed to check if everything works correctly with those changes (i.e. if escaped $ are displaying correctly on built site) - this doesn't seem to fix any problem in particular (other than some packages not being up to date), so I wasn't prioritising it.

Cool, yeah we might wanna hold on until we can confirm that escaped `$` are not a bug also

lipis:
What's the status with this one?

alexandernanberg:
@gatsbyjs/core Would be nice to get this merged (if there isn't anything you want me to change) so we don't have to keep syncing it with master 😅

PR #9672, opened 2018-11-03
fix comment of setFieldsOnGraphQLNodeType

Takumon:


pieh:
@pranshuchittora This is unrelated - sometimes test will fail because of network or other CI issues unrelated to changes in repo

PR #9678, opened 2018-11-03
fix(www): hide feedback and edit on github for remote packages 

hbish:
For plugins not hosted in gatsby/package, the 2 features (feedback and edit in github) in the footer does not work as intended. 

This PR adds a toggle to disable those feature when no page metadata is available.

screen shot 2018-11-03 at 8 24 14 pm


shannonbux:
@pieh I think this is a question for @amberleyromo! My vote is that we get rid of those things for the plugin library entirely and instead introduce other kinds of sorting and filtering to help ensure quality. We've also talked about having a feedback form on the plugin library in general.

amberleyromo:
> @pieh I think this is a question for @amberleyromo! My vote is that we get rid of those things for the plugin library entirely and instead introduce other kinds of sorting and filtering to help ensure quality. We've also talked about having a feedback form on the plugin library in general.

@pieh @shannonbux sorry, somehow I had missed this (thanks @LekoArts). Shannon, I agree we can just get rid of this entirely for the plugin library. #10784 is adding the Unbird feedback form to all plugin package pages.

LekoArts:
Since https://github.com/gatsbyjs/gatsby/pull/10784 is merged, we can get this merged, too, right?

amberleyromo:
@LekoArts yep. i just pushed some changes. if you could re-review:

- removed the feedback thumbs up / thumbs down entirely (Unbird replaces this)
- conflict resolution adding Unbird to remote package template (after persisting sidebar #10590 was merged)
- existing fix for only showing "edit on GitHub" CTA for internal pages is intact

hbish:
Thanks for fixing this @amberleyromo, I wasn't aware there was a plan to replace all the feedback features with unbird. 👍

PR #9679, opened 2018-11-03
fix(gatsby-plugin-offline): Serve the offline shell for short URLs + use no-cors for external resources

### LINKED ISSUE #8145: Offline plugin sometimes has a fetch error
## Description

Some combinations of plugins can cause console errors in the browser when first visiting a site.

Using `gatsby-plugin-offline` with `gatsby-plugin-google-analytics` and a plugin that does something like:

```js
exports.onRenderBody = ({ setHeadComponents }) => {
    setHeadComponents([<script />])
}
```

will trigger the following browser error when first visiting the site:

```
Access to fetch at 'https://www.google-analytics.com/analytics.js' from origin 'http://localhost:9001' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
(index):1 Uncaught (in promise) TypeError: Failed to fetch
```

Reloading the page will load the page without errors. Removing the service worker and reloading the page will display the error again.

In short: 

- offline plugin + analytics plugin = no error
- offline plugin + analytics plugin + another script = analytics fetch error

### Steps to reproduce

Repo: https://github.com/m-allanson/offline-analytics
Demo video: https://www.youtube.com/watch?v=zPzEq7aMGfA

- clone the repo
- `yarn`
- `yarn run build`
- `gatsby serve`
- browse to `http://localhost:9000`
- in Chrome open the application tab of the developer tools
- ensure _unregister service workers_ is checked
- click _clear site data_
- reload the page
- see console errors in the browser

### Expected result

No console error

### Actual result

Console error on first visit

### Environment

```
  System:
    OS: macOS High Sierra 10.13.6
    CPU: x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.9.0 - ~/.nvm/versions/node/v10.9.0/bin/node
    Yarn: 1.9.4 - /usr/local/bin/yarn
    npm: 6.2.0 - ~/.nvm/versions/node/v10.9.0/bin/npm
  Browsers:
    Firefox: 62.0
    Safari: 11.1.2
  npmPackages:
    gatsby: next => 2.0.0-rc.24
    gatsby-plugin-google-analytics: next => 2.0.0-rc.2
    gatsby-plugin-offline: next => 2.0.0-rc.7
  npmGlobalPackages:
    gatsby-dev-cli: 2.0.0-rc.5
    gatsby: 2.0.0-rc.24
```

davidbailey00:
- Fixes the fallback whitelist regex so that short pages match (e.g. `/`), so that these are served offline correctly
- Uses `no-cors` mode when caching external resources, so that errors don't appear if the target doesn't allow CORS requests
- Allow insecure external resources to be cached for testing on `localhost`

Fixes #8145

### DIFF
--- a/packages/gatsby-plugin-offline/src/gatsby-node.js
+++ b/packages/gatsby-plugin-offline/src/gatsby-node.js
@@ -88,7 +88,7 @@ exports.onPostBuild = (args, pluginOptions) => {
     // URLs and not any files hosted on the site.
     //
     // Regex based on http://stackoverflow.com/a/18017805
-    navigateFallbackWhitelist: [/^[^?]*([^.?]{5}|\.html)(\?.*)?$/],
+    navigateFallbackWhitelist: [/^([^.?]*|[^?]*\.([^.?]{5,}|html))(\?.*)?$/],
     navigateFallbackBlacklist: [/\?(.+&)?no-cache=1$/],
     cacheId: `gatsby-plugin-offline`,
     // Don't cache-bust JS or CSS files, and anything in the static directory
@@ -101,7 +101,7 @@ exports.onPostBuild = (args, pluginOptions) => {
       },
       {
         // Use the Network First handler for external resources
-        urlPattern: /^https:/,
+        urlPattern: /^https?:/,
         handler: `networkFirst`,
       },
     ],
--- a/packages/gatsby-plugin-offline/src/sw-append.js
+++ b/packages/gatsby-plugin-offline/src/sw-append.js
@@ -9,13 +9,19 @@ self.addEventListener(`message`, event => {
     event.waitUntil(
       caches.open(cacheName).then(cache =>
         Promise.all(
-          resources.map(resource =>
-            cache.add(resource).catch(e => {
-              // ignore TypeErrors - these are usually due to
-              // external resources which don't allow CORS
-              if (!(e instanceof TypeError)) throw e
-            })
-          )
+          resources.map(resource => {
+            let request
+
+            // Some external resources don't allow
+            // CORS so get an opaque response
+            if (resource.match(/^https?:/)) {
+              request = fetch(resource, { mode: `no-cors` })
+            } else {
+              request = fetch(resource)
+            }
+
+            return request.then(response => cache.put(resource, response))
+          })
         )
       )
     )


packages/gatsby-plugin-offline/src/gatsby-node.js
```diff
@@ -88,7 +88,7 @@ exports.onPostBuild = (args, pluginOptions) => {
     // URLs and not any files hosted on the site.
     //
     // Regex based on http://stackoverflow.com/a/18017805
-    navigateFallbackWhitelist: [/^[^?]*([^.?]{5}|\.html)(\?.*)?$/],
+    navigateFallbackWhitelist: [/^([^.?]*|[^?]*\.([^.?]{5,}|html))(\?.*)?$/],
```
DSchau: This is getting hard to follow 😱 (not your fault!)

So _originally_ this was for capturing a file name without an extension, so if we break this apart

- start of string
- PARTS
  - `[^.?]*` matches _anything_ without a dot, e.g. `/asdf`
  - `[^?]*\.([^.?]{5,}|html)` matches also html files?
  - `(\?.*)` matches query params
- end of string

Did I encapsulate that OK?

packages/gatsby-plugin-offline/src/gatsby-node.js
```diff
@@ -101,7 +101,7 @@ exports.onPostBuild = (args, pluginOptions) => {
       },
       {
         // Use the Network First handler for external resources
-        urlPattern: /^https:/,
+        urlPattern: /^https?:/,
```
DSchau: Is this a best practice? This seems significant in that we'll cache non-secure resources now, too.

DSchau: Ah, got it!

> Allow insecure external resources to be cached for testing on localhost

Is there a risk here of content being cached in localhost and leading to confusion? I'd almost prefer it kept _not working_ in localhost, but perhaps that's a naive statement?

packages/gatsby-plugin-offline/src/sw-append.js
```diff
@@ -9,13 +9,19 @@ self.addEventListener(`message`, event => {
...
         Promise.all(
-          resources.map(resource =>
-            cache.add(resource).catch(e => {
-              // ignore TypeErrors - these are usually due to
-              // external resources which don't allow CORS
-              if (!(e instanceof TypeError)) throw e
-            })
-          )
+          resources.map(resource => {
+            let request
```
DSchau: Few questions here:

- Do we still want to throw on (certain?) errors? I suppose the promise will just reject?
- This seems weird that we've totally changed the approach, i.e. we're fetching everything now?
- This seems to fix things like google-analytics. Were you able to validate that?

Just asking these to make sure we're on the same page and I understand things!

davidbailey00: > Do we still want to throw on (certain?) errors? I suppose the promise will just reject?

If there are any odd errors then yes, the promise will reject the same way as before

> This seems weird that we've totally changed the approach, i.e. we're fetching everything now?

We were already fetching everything with the old code, just never with `no-cors`, so sometimes there would be console errors

> This seems to fix things like google-analytics. Were you able to validate that?

I haven't tested Google Analytics explicitly but yes, it should be fixed since no-cors means it can be cached. I've tested other external resources though to check it's okay.

DSchau: Maybe we validate with that one just to validate we've fixed the issue this is fixing?

Once that's done, I think this will be good to go and merge 🎉

DSchau:
Successfully published:
 - gatsby-plugin-offline@2.0.12

packages/gatsby-plugin-offline/src/sw-append.js
wconnorwalsh: This is not a best practice and will quickly lead to improper caching of resources. 

From [Workbox's own documentation](https://developers.google.com/web/tools/workbox/guides/storage-quota#beware_of_opaque_responses):

> # Beware of opaque responses! 
> A common source of unexpectedly high quota usage is due to runtime caching of opaque responses, which is to say, cross-origin responses to requests made without CORS enabled.
>  
> Browsers automatically inflate the quota impact of those opaque responses as a security consideration. In Chrome, for instance, even an opaque response of a few kilobytes will end up contributing around 7 megabytes towards your quota usage.

This should probably be enabled on a case-by-case basis using the plugin's options or not at all. If the external domain allows CORS, the resource should flag it using the HTML `crossorigin` attribute.

DSchau: @wconnorwalsh awesome; thank you for the info! This change also (could have) illustrated exactly what you reference here:

> A common source of unexpectedly high quota usage is due to runtime caching of opaque responses

as we've seen some recent weirdness with gatsbyjs.org.

I'm mostly out today, would you be interested in reverting this change and making this opt-in rather than enabled by default? If not, we'll get to it as soon as we can!

davidbailey00: cc @wconnorwalsh @DSchau please note that only the changes to `sw-append.js` should be reverted - the others are unrelated!

davidbailey00: I've looked into this a bit more and going to create a PR shortly to revert this + fix a few more things.

PR #9680, opened 2018-11-03
fix(gatsby-plugin-netlify): do not cache service worker file on Netlify

matthieuauger:
Related to https://www.netlify.com/blog/2018/06/28/5-pro-tips-and-plugins-for-optimizing-your-gatsby---netlify-site/#4-get-your-service-workers-um-working

Recommandation is not to cache service worker file : https://github.com/w3c/ServiceWorker/issues/893

### DIFF
--- a/packages/gatsby-plugin-netlify/src/constants.js
+++ b/packages/gatsby-plugin-netlify/src/constants.js
@@ -26,6 +26,7 @@ export const SECURITY_HEADERS = {
 
 export const CACHING_HEADERS = {
   "/static/*": [`Cache-Control: public, max-age=31536000, immutable`],
+  "/sw.js": [`Cache-Control: no-cache`],
 }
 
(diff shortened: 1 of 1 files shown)


DaleJefferson:
Is this PR compatible with https://github.com/gatsbyjs/gatsby/pull/9245

matthieuauger:
@DaleJefferson Yes, if the line provided by @tanvibhakta is added after mine, it will

pieh:
@matthieuauger, @DaleJefferson That PR need changes - we won't be targetting all js files - we will read manifest from webpack and use assets paths generated by webpack (which will have hash for cache busting)

pieh:
Published `gatsby-plugin-netlify@2.0.5`

matthieuauger:
@pieh Really no need to be sorry, thanks for what you do!

About you last comment any hints about how we should perform this so I could have a look?

Thanks

PR #9695, opened 2018-11-04
chore(gatsby): Remove react-hot-loader/patch entry

alexandernanberg:
The patch entry should no longer be necessary (after v4 I think), so I should be safe to remove it.

https://github.com/gaearon/react-hot-loader#no-patch-required

### DIFF
--- a/packages/gatsby/src/utils/webpack.config.js
+++ b/packages/gatsby/src/utils/webpack.config.js
@@ -144,7 +144,6 @@ module.exports = async (
       case `develop`:
         return {
           commons: [
-            require.resolve(`react-hot-loader/patch`),
             `${require.resolve(
               `webpack-hot-middleware/client`
             )}?path=${getHmrPath()}`,


alexandernanberg:
Hmm I have a weird issue with `gatsby-dev-cli` – after I run `gatsby-dev` the permissions of the moved files change and I can't execute them.

So when running `yarn develop` I get this error

```
yarn run v1.12.1
$ gatsby develop
/bin/sh: /Users/alna/dev/gatsby-starter-default/node_modules/.bin/gatsby: Permission denied
error Command failed with exit code 126.
```

Anyway, works when I set the correct permission manually with `chmod`.

In my tests it looks like `react-hot-loader` isn't working without this change either. Because if I add a stateful component the state should persist when I make changes to the file – right?

alexandernanberg:
@DSchau Okay, so I can confirm that this works exactly the same way as before on my machine using `gatsby-dev`. So this should be safe to merge

PR #9699, opened 2018-11-04
feat(www): Re-Organize tags on starters.yml

LekoArts:
- Added "Headless CMS" to entries appropriately
- Removed minor tags that create extra noise (Recompose, Prettier, Lint-Staged, TSLint, Tags) and removed them or put them into "Linting". Especially considering that almost all starters have ESLint and Prettier. "Linting" would mean for me that this starter is especially focused on that or has something fancy
- Removed invalid Styling tag

### DIFF
--- a/docs/starters.yml
+++ b/docs/starters.yml
@@ -52,6 +52,7 @@
   tags:
     - Blog
     - Contentful
+    - Headless CMS
   features:
     - Based on the Gatsby Starter Blog
     - Includes Contentful Delivery API for production build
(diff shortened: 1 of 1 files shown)


docs/starters.yml
```diff
@@ -1075,7 +1084,7 @@
   description: Blog Template X Contentful, Twitter and Facebook style
   tags:
     - Blog
-    - Tags
```
pieh: I think intention here is that starters has support for post tags?

LekoArts: Mhhh, no one used that even though the starter has it. You could also call that categories, so what to choose here then?

pieh: I probably miss-understood intention (I thought this was one of invalid tags). On that I have no opinion

LekoArts:
@shannonbux ^

Issue #4459, opened 2018-03-09
gatsby-transformer-remark not transforming excerpts


vojtechruz:
I am using gatsby-transformer-remark with the following configuraton:
```
{
      resolve: `gatsby-transformer-remark`,
      options: {
        excerpt_separator: `<!--more-->`,
        plugins: [
          {
            resolve: `gatsby-remark-images`,
            options: {
              maxWidth: 590,
            },
          },
          {
            resolve: `gatsby-remark-responsive-iframe`,
            options: {
              wrapperStyle: `margin-bottom: 1.0725rem`,
            },
          },
          'gatsby-remark-prismjs',
          'gatsby-remark-copy-linked-files',
        ],
      },
    },
```
According to this post https://using-remark.gatsbyjs.org/excerpts/ I am querying excerpts from my articles like this:
```

export const pageQuery = graphql`
  query IndexQuery {
    site {
      siteMetadata {
        title
      }
    }
    allMarkdownRemark(sort: { fields: [frontmatter___date], order: DESC }) {
      edges {
        node {
          excerpt
          fields {
            slug
          }
          frontmatter {
            date(formatString: "DD MMMM, YYYY")
            title
            tags
            featuredImage {
                childImageSharp{
                    sizes(maxWidth: 200) {
                        ...GatsbyImageSharpSizes_tracedSVG
                    }
                }
            }
          }
        }
      }
    }
  }
`
```
My article in markdown uses a separater defined in config above:
```

---
title: Break the Java Generics Naming Convention?
date: "2016-04-30T22:12:03.284Z"
tags: ['Java']
path: '/break-java-generics-naming-convention'
featuredImage: './break-java-generics-maning-convention.jpg'
---

Choosing **descriptive**, *intent-revealing* names is one of the key principles of writing clean code, which is easy to read and understand and usually requires little to no comments. According to the official convention, generic type parameter names are supposed to be just one capital letter. Should you break this convention to make your names more descriptive?
<!--more-->
```

---

However, the excerpt I am getting is plain text and not transformed from markdown:

![clipboard02](https://user-images.githubusercontent.com/7038612/37195493-57200fc6-2373-11e8-9ace-a9e502a7028e.jpg)

---
Is this intended behavior? How can I get excerpt which is already formatted? Thank you.

---
    "gatsby": "^1.9.223",
    "gatsby-transformer-remark": "^1.7.34",

Node: v9.5.0
NPM: 5.6.0

[...]

PR #5586, opened 2018-05-29
Bugfix 4459

arlobelshee:
Resolves bug #4459 by, when an excerpt is present on the incoming node, parsing it exactly like the node's content.

Implemented this by extracting the functions involved in the parsing from the caching behavior around them, then calling them twice (once for each part). The lowest-level function (which actually does the remark parsing to an AST, asynchronously) is triggered on both parts as soon as the code asks for either. It parses the two in parallel, storing the pair of results in the cache.

The outer functions (`getHTML()` and friends) are simply duplicated, with the excerpt HTML stored in the cache just like the content HTML. By duplicated, I mean extracted, parameterized by wrapper, and then the wrapper is duplicated for the other behavior. This minimizes code duplication.

Note on my commit messages: each commit is very small. The first character indicates my attempted risk level / kind of transform for that commit. I did make a couple errors along the way; I think I caught them.

The risk level codes I use are:

* r = disciplined refactoring. Low risk.
* a = autoformat done by a tool. Low risk.
* !!! = my intent was to refactor, but I don't have a recipe for this, so it is much riskier.
* F = work on the feature. Intentionally changes behavior, so there is more risk.
* B = bugfix. I made a mistake, caught it, and fixed it. Refers to where the mistake was made. Bugfixes are inherently risky themselves.

Finally, I am totally unfamiliar with this code and have no idea how good your test coverage is. I did not extend it. So I believe everything works, but I can't know how confident to be. Some contributor, please help verify!

PR #9716, opened 2018-11-05
Fix/4459/excerpt formatting ignored

nadrane:
This is the beginnings of a fix for https://github.com/gatsbyjs/gatsby/issues/4459

Right now, the fix works when excerpt_separator is provided. I have not yet implemented a fix for when the excerpt is generated using the `pruneLength` setting, though I have thoughts on how to solve it. My plan to is do a pre-order traversal over the html AST, accumulating nodes until the pruneLength is reached.

I would love some preliminary feedback to get a sense of how the contributors feel about this PR so far. Here are some of my initial thoughts:

1. The work in the `excerpt` resolve function is a little hacky. We are directly invoking a variety of functions get our HTML. I'm curious if anyone has a more streamlined approach for this.

2. We are invoking `getMarkdownAST` (previously getAST) twice, once for the excerpt and once for the entire page. I'm worried that this is going to break non-idempotent existing plugins since they will now run twice.

3. I tore apart the `getAST` function a bit and think I did a good job simplifying it. The big change is that we don't generate multiple promises anymore. i think this goes a long way towards improving readability.

I'd love to hear people's thoughts.


pieh:
Did you check https://github.com/gatsbyjs/gatsby/pull/5586 which tackles same problem?

> 2\. We are invoking `getMarkdownAST` (previously getAST) twice, once for the excerpt and once for the entire page. I'm worried that this is going to break non-idempotent existing plugins since they will now run twice.

There's no way around it I think. Unless excerpt separator is visible in markdown AST, then we could use same AST and break manually on separator (but I don't think it's there)

> 3\. I tore apart the `getAST` function a bit and think I did a good job simplifying it. The big change is that we don't generate multiple promises anymore. i think this goes a long way towards improving readability.

🙏 At some point I would like to have `cache.wrap` that would exactly what new version of `getAST` does but hidden from end user (kind of like `_.memoize`), so you can wrap your async function in cache.

nadrane:
> There's no way around it I think. Unless excerpt separator is visible in markdown AST, then we could use same AST and break manually on separator (but I don't think it's there)

I can look to see if the separator is in the AST...

One alternative is that we could just not run plugins on the extract. This is an unfortunate limitation, though it's definitely better than the current situation and might be better than potentially breaking all/half of the plugins that use the `gatsby-transformer-remark` API. How would you feel about this approach?

nadrane:
So the comment does make its way into the markdown AST. With that said, editing it and everything that follows it out is not trivial. At first I thought it would be simple, but as I looked at the markdown AST data structure a bit closer, I realized that it's storing position relative to the original text. Here is an example:

```js
node { type: 'root',
        children:
         [ { type: 'paragraph', children: [Array], position: [Object] },
           { type: 'html', value: '<!-- end -->', position: [Object] },
           { type: 'paragraph', children: [Array], position: [Object] },
           { type: 'paragraph', children: [Array], position: [Object] } ],
        position:
         { start: { line: 1, column: 1, offset: 0 },
           end: { line: 6, column: 5, offset: 1013 } } }
```

You can see that the AST stores the offsets for the beginning and ending of every node. If I were to remove a subset of nodes from the AST, I would need to update all these positions. Unfortunately, as far as I can tell, there aren't any helper utilities available online to make these AST updates for me.

I think I'm going to continue with the current approach and rely on calling `getAST` twice.

nadrane:
This PR is mostly ready aside from updating some snapshots. The code should be pretty much finalized. The tests need a little help.

Let me give you an overview of what I've done

1.  I was able to successfully parse the markdown AST for the `excerpt_separator`, allowing us to avoid calling `getAST` twice and risking calling plugins multiple times. I did this by placing the `excerpt_separator` on the markdownNode. I couldn’t figure out another way to make it accessible to the `extend-node-type` file. If there is a better strategy, I’d love to learn about it.

2. I had to write quite a bit of code to walk the html AST. Given the amount of caching similar prior code does, I got the impression that it is particularly performance sensitive. I’m doing quite a bit of tree traversal, sometimes invoking the same function multiple times. This could be optimized with some simple memoization if you think that would be valuable. My suspicion is it’s probably fine without it, though (unless maybe an excerpt were particularly long and full of many nodes).

3. I wasn’t sure how to get the `gatsby-node` tests passing again after changing the way excerpts are calculating when an `excerpt_separator` is present. These tests are clearly not calling the `extend-node-type` code, where the `excerpt_separator` is now handled, causing 2 tests to fail. I would love some guidance on this. Or, if it’s ok, I could also just remove the tests, though it’s not clear to me if we would then be sacrificing some degree of test coverage

4. The changes are slightly breaking. Even if a user previously never used markdown with their excerpts, we at the very least wrap ever excerpt in a `<p>` tag. This could definitely break styling on existing websites. I don't really see another option, however

Looking forward to hearing your thoughts

pieh:
> I couldn’t figure out another way to make it accessible to the `extend-node-type` file. If there is a better strategy, I’d love to learn about it.

You should have access to it via `pluginOptions.excerpt_separator` (to not copy it to every markdown node)

> 3\. I wasn’t sure how to get the `gatsby-node` tests passing again after changing the way excerpts are calculating when an `excerpt_separator` is present. These tests are clearly not calling the `extend-node-type` code, where the `excerpt_separator` is now handled, causing 2 tests to fail. I would love some guidance on this. Or, if it’s ok, I could also just remove the tests, though it’s not clear to me if we would then be sacrificing some degree of test coverage

> 4\. The changes are slightly breaking. Even if a user previously never used markdown with their excerpts, we at the very least wrap ever excerpt in a `<p>` tag. This could definitely break styling on existing websites. I don't really see another option, however

Yeah, this would be breaking change so would need major version bump (if merged as-is). Will think if we should do intermediate step of keeping old field (and deprecate it) and add new field (i.e `excertptHtml` with your code behind it). There's lot of content using `<p>${markdownRemark.frontmatter.excerpt}</p>` around and with this change we would need to use `dangerouslySetInnerHTML`, so transition period would be needed here.

nadrane:
> You should have access to it via pluginOptions.excerpt_separator (to not copy it to every markdown node)

I was able to access the `excerpt_separator` using `pluginOptions` in the `on-node-create` file but not in the `extend-node-type` file. In the latter file it was empty. It sounds like you're saying this is not the expected behavior.

> Yeah, this would be breaking change so would need major version bump (if merged as-is). Will think if we should do intermediate step of keeping old field (and deprecate it) and add new field (i.e excertptHtml with your code behind it). There's lot of content using <p>${markdownRemark.frontmatter.excerpt}</p> around and with this change we would need to use dangerouslySetInnerHTML, so transition period would be needed here.

Let me know how you'd like to approach it. I'm fine creating a new GraphQL field in the interim.

pieh:
> I was able to access the `excerpt_separator` using `pluginOptions` in the `on-node-create` file but not in the `extend-node-type` file. In the latter file it was empty. It sounds like you're saying this is not the expected behavior.

was it happening in jests tests or when you were running gatsby with your changes applied?

nadrane:
> was it happening in jests tests or when you were running gatsby with your changes applied?

Yeah it was happening through Jest tests. Sounds like they aren't set up quite correctly?

pieh:
> Yeah it was happening through Jest tests. Sounds like they aren't set up quite correctly?

Seems like it - https://github.com/gatsbyjs/gatsby/blob/76037bfb7866ea7f0664404b33ecd5ab8729267e/packages/gatsby-transformer-remark/src/__tests__/extend-node.js#L34-L36 here we would need to pass `pluginOptions` (which includes `excerpt_separator`)

nadrane:
Ah, that makes sense. Thank you.

As far as a final decision on this PR goes, are you settled on creating a `excerptHTML` field and leaving the existing `excerpt` field as-is? If so, I've got a plane ride in a few hours and will probably do the work then.

nadrane:
@pieh Hey, just checking in, what are you thinking for a final decision on this one?

nadrane:
> Use something like remove-markdown to remove markdown characters
This should (🤞) not really be a breaking change, and users should just see a benefit from it; unless they were relying on that field being markdown, which seems odd

I feel like this solution is a definite improvement, but I don't think it aligns well with expected behavior. As a user, I would expect excerpts to render my markdown as html if they included markdown. The big advantage of this approach, in my opinion, is that its simplicity. With that said, we're already 90% of the way there towards the next two solutions.

> Expose a separate field (excerptHtml) and keep excerpt as is
Expose arguments, e.g. format: 'plain' (this should be the default) and format: 'html'

I agree with you that the latter approach is better. I feel like the former clutters the API, and if we wanted to migrate to an html-only excerpt at a later date, we would be stuck with an field name (excerptHtml) whose name is a bit of an anachronism. 

Is the ultimate goal going to be to phase out support for the plain text option? I honestly don't see a huge use case for it. If so, with whatever option we choose, I could insert a deprecation warning if the user elects to return plain text.

nadrane:
it sounds like you have a wayyy better understanding of the use cases of this field than I do! How are excerpts used for meta tags or SEO, just out of curiosity?

So it sounds like we're in alignment to go with option 3, where we have an `excerpt` field with a format option. I assume we should keep the default as plaintext because, as you explained, there's no clear winner between plain vs. html and for backwards compatibility reasons.

I'll modify the PR with these changes in mind! looking forward to getting this merged!

DSchau:
@nadrane here's an example I've used in the past :) https://github.com/DSchau/blog/blob/master/src/templates/blog-post.js#L32-L34

DSchau:
@nadrane I'm looking forward to getting this merged too! Thanks for being patient and working with us on this, I think this is a great change 🎉

DSchau:
@nadrane awesome! As far as testing the changes locally - you better believe it!

Check out [gatsby-dev-cli](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-dev-cli), specifically check out [these instructions](https://www.gatsbyjs.org/docs/how-to-contribute/#using-yarn) for getting up and running with it. Let us know if you need any help!

packages/gatsby-transformer-remark/src/extend-node-type.js
```diff
@@ -373,8 +389,59 @@ module.exports = (
             type: GraphQLBoolean,
             defaultValue: false,
           },
+          format: {
+            type: GraphQLString,
```
DSchau: Do we want to use an [enum type](https://graphql.org/graphql-js/type/#graphqlenumtype) here? It seems like only `plain` and `html` are options, and designating string could lead to subtle bugs (e.g. typos or something!).

nadrane:
alright, I think this is finally ready for review! Thanks for the gatsby-dev-cli tips. Worked like a charm

DSchau:
@nadrane pulling this down to test right now. Awesome job thus far 👌

nadrane:
@DSchau not sure how that happened. I fixed it :)

DSchau:
```
Successfully published:
 - gatsby-image@2.0.21
 - **gatsby-transformer-remark@2.1.13**
 - gatsby@2.0.56
```

Issue #9705, opened 2018-11-05
Client-only routes: catch-all matchPath has higher "specificity" than static paths


juliansthl:
## Description

If I specify a wildcard matchPath for the root path on my index page `matchPath: '/*'`, the index page is always served, ignoring any "static" paths.

### Steps to reproduce

**[Demo on CodeSandbox](https://codesandbox.io/s/rmz76130xm?module=%2Fgatsby-node.js)**

Based on gatsby-starter-default, I'm adding this to **gatsby-node.js**:

```
const path = require('path');

exports.onCreatePage = async ({ page, actions }) => {
  const { createPage, deletePage } = actions;

  return new Promise(resolve => {
    const oldPage = { ...page };

    if (page.path === '/') {
      
      // Replace index
      deletePage(oldPage);

      createPage({
        ...page,
        matchPath: '/*',
      });

      // Create sub page
      createPage({
        path: `/sub/`,
        matchPath: `/sub/*`,
        component: path.resolve(`./src/templates/sub.js`),
      });
    }

    resolve();
  });
};
```

### Expected result

- When browsing to `/` or `/foo` I would expect to see the index page **_./src/pages/index.js_**
- When browsing to `/sub/` or `/sub/foo` I would expect to see the sub page template **_./src/templates/sub.js_**

I'm quite sure that this worked in the past and [this comment](https://github.com/gatsbyjs/gatsby/issues/8640#issuecomment-425874258) from @pieh makes me believe this is the way the matching is supposed to work.

### Actual result

gatsby always serves the index page **_./src/pages/index.js_**

### Environment

```
System:
    OS: macOS 10.14
    CPU: x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
    Shell: 5.5.1 - /usr/local/bin/zsh
  Binaries:
    Node: 10.6.0 - ~/.nvm/versions/node/v10.6.0/bin/node
    Yarn: 1.9.4 - /usr/local/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v10.6.0/bin/npm
  Browsers:
    Chrome: 70.0.3538.77
    Firefox: 62.0
    Safari: 12.0
  npmPackages:
    gatsby: ^2.0.19 => 2.0.38
    gatsby-image: ^2.0.15 => 2.0.19
    gatsby-plugin-manifest: ^2.0.5 => 2.0.7
    gatsby-plugin-offline: ^2.0.5 => 2.0.11
[...]

PR #9719, opened 2018-11-05
fix: adjust page order to make nested matchPaths work

juliansthl:
As per [#9705](https://github.com/gatsbyjs/gatsby/issues/9705), this changes the order of the pages in pages-writer.js in order to account for the matchPath specificity and therefore make "nested" matchPaths work.

The proposed fix `sortByNumeric` is the most performant solution I could come up with:
[Benchmarks](https://runkit.com/juliansthl/5be08b8f2513440012001807)

Thanks a lot to @pieh for all the help!

(this is my first PR, please let me know if I have to adjust anything)

### DIFF
--- a/packages/gatsby/src/internal-plugins/query-runner/pages-writer.js
+++ b/packages/gatsby/src/internal-plugins/query-runner/pages-writer.js
@@ -36,9 +36,11 @@ const writePages = async () => {
   })
 
   pagesData = _(pagesData)
-    // Ensure pages keep the same sorting through builds
-    // and sort pages with matchPath to end so explicit routes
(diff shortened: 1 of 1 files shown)


pieh:
I think we need to adjust comment right above the change to explain what it does and why:
so pages without `pathPrefix` should be first, and then pages with `pathPrefix` are sorted so more specific patterns are before less specific patterns and we infer specificity from number of path segments

pieh:
Code LGTM, was nice and quick work on this! :)

juliansthl:
Adjusted the comment as suggested. Thanks again!!

PR #9629, opened 2018-11-01
fix: ensure babel-preset-gatsby can be used with unit tests

DSchau:
This PR does a few things, namely:

- Fixes a few things with the test stage, e.g. uses commonjs modules for
test running, uses the current node version for @babel/preset-env, etc.
- Fixes up docs slightly for the unit testing guide
- Fixes #9614

PR #9720, opened 2018-11-05
fix(gatsby-plugin-page-creator): ensure that __tests__ directory is actually ignored

DSchau:
This (simple) PR fixes a bug I was noticing in #9629, specifically that
__tests__/*.js files _were_ being included in the bundle.

This is because the isTestFile function in gatsby-plugin-page-creator is
passed the file name, not the file name _with_ the folder (__tests__ in
this case).

So, this PR does the following:

- Adds a fallback check to check the base dir (which could be __tests__)
- (Also) ignores typescript files as part of this check, since this isn't readily apparent _how_ it would be configurable to ignore them currently

Note: I could've solved this by also passing the raw path (not system path parsed) but wanted to avoid any Windows idiosyncrasies _just in case_

### DIFF
--- a/packages/gatsby-plugin-page-creator/src/__tests__/gatsby-node.js
+++ b/packages/gatsby-plugin-page-creator/src/__tests__/gatsby-node.js
@@ -92,13 +92,14 @@ describe(`JavaScript page creator`, () => {
     const validFiles = [{ path: `page.js` }, { path: `page.jsx` }]
 
     const testFiles = validFiles.concat([
-      { path: `__tests__/something.test.js` },
-      { path: `__tests__/something.test.tsx` },
-      { path: `__tests__/something.js` },
-      { path: `foo.spec.js` },
-      { path: `foo.spec.tsx` },
-      { path: `bar.test.js` },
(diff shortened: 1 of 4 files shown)


packages/gatsby-plugin-page-creator/src/__tests__/gatsby-node.js
```diff
@@ -10,28 +10,26 @@ describe(`JavaScript page creator`, () => {
       { path: `somedir/dir2/test1.js` },
     ]
 
-    expect(validFiles.filter(file => validatePath(file.path)).length).toEqual(validFiles.length)
+    expect(validFiles.filter(file => validatePath(file.path))).toEqual(
```
DSchau: Also removed all the .length checks. It's a lot safer to validate the actual array shape rather than the length of the array, even though there weren't any changes required here!

e.g. if one of the validFiles was failing _and_ one of the invalid files was passing, we could still get the expected length

pieh:
e2e tests failing because of new dependency not being handled, so merging this

PR #9730, opened 2018-11-06
fix minor typo in docs

sw-yx:


sw-yx:
now can i have my socks :)

shannonbux:
Thanks @sw-yx! Nice edit and hope the socks are awesome :)

PR #9734, opened 2018-11-06
fix: use relative HMR path

palerdot:
Construct webpack hmr path relative to the request origin path to prevent url construction issues when accessing from local ip or any other ip that may be mapped to the dev server.

This PR makes following changes to `getHmrPath()` method in `packages/gatsby/src/utils/webpack.config.js`

- Always use the relative path `/`
- Getting rid of manual url construction with protocol, hosts, ports etc as the relative path will automatically take care of this construction


packages/gatsby/src/utils/webpack.config.js
```diff
@@ -71,21 +71,11 @@ module.exports = async (
...
-    let hmrBasePath = `${program.ssl ? `https` : `http`}://${
-      program.host
-    }:${webpackPort}/`
-
+    // ref: https://github.com/gatsbyjs/gatsby/issues/8348
+    // construct the webpack url relative to where the request is coming from
+    const hmrBasePath = `/`
     const hmrSuffix = `__webpack_hmr&reload=true&overlay=false`
 
-    if (process.env.GATSBY_WEBPACK_PUBLICPATH) {
```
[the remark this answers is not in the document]
drubin: What about something like just doing 

`const hmrBasePath = process.env.GATSBY_WEBPACK_PUBLICPATH || `/``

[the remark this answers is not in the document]
palerdot: @drubin Agreed. `const hmrBasePath = process.env.GATSBY_WEBPACK_PUBLICPATH || '/'` is more safer and configurable if needed. I will update the PR with this change. Meanwhile, if there are any other changes to be made please let us know.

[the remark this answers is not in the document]
palerdot: @drubin updated PR with changes. Let me know if there is anything else I can do from my side.

pieh:
Hey, could you list scenarios that you tested this and it works? We definitely need to verify any other scenarios to be as sure as possible that this won't break for some cases

packages/gatsby/src/utils/webpack.config.js
```diff
@@ -71,21 +71,11 @@ module.exports = async (
...
-      const pubPath = process.env.GATSBY_WEBPACK_PUBLICPATH
-      if (pubPath.substr(-1) === `/`) {
-        hmrBasePath = pubPath
-      } else {
-        hmrBasePath = `${pubPath}/`
-      }
-    }
-
```
pieh: please don't remove that part - we still want to ensure path ends with `/` I think

drubin:
@pieh This should work in all situations where you host Gatsby on the root of the domain/IP. If you do access it via a Nginx/proxy on a sub URL you will need to use the `GATSBY_WEBPACK_PUBLICPATH` which you would have to do anyway.

I think the previous implementation made a wrong assumption that the by using the same IP address that the webpack was bound to be the same as the public-facing URL (in most cases this would be the same but not always).

All this means is that the URL for `hmr` that is generated will not include the domain/port/proxy. Since this is requested from the page its self the borwser will infer where to fetch it from (i.e the same place the page came from which seems like the sanest default).

For example, if your dev ENV sites inside a docker image or behind an HTTP proxy (for ssl/name based routing). 

is HMR used for anything other than the browser fetching it?

packages/gatsby/src/utils/webpack.config.js
palerdot: @pieh updated PR with the changes. I tested with following commands
- `GATSBY_WEBPACK_PUBLICPATH=/ gatsby develop -H 0.0.0.0` - setting the public path
- `gatsby develop -H 0.0.0.0` - without setting the public path

In both the cases, I was able to access the gatsby developement server running inside my vagrant machine

palerdot: @pieh @drubin I guess there is no difference between first and second command when it comes to testing this piece of code as thing are falling back to `/` anyway. One think that was not tested is testing with nginx as reverse proxy. I don't have that setup. In any case, what exactly to test with an nginx setup? let's say `mysite.local/gatsby` is proxied to 'localhost:8000', then what should be set in public path?

drubin:
@pieh Yes, you are right I tried quickly searching back in the history 

https://github.com/gatsbyjs/gatsby/blob/3dbc3eb92505caad731060f56d7a98ed66c6c03f/packages/gatsby/src/utils/webpack.config.js#L92 

I found it an older implementation  

``${require.resolve(`webpack-hot-middleware/client`)}?path=http://${program.host}:${webpackPort}/__webpack_hmr``

Seems like it was loading from another plugin and that plugin was passing it in as a query param. That doesn't seem to be the case? 

But of course without actually knowing why it was there in the first place its hard to figure out if it still passes that use case.

packages/gatsby/src/utils/webpack.config.js
palerdot: @drubin Thanks for the clarification. As of now, I have not tested with an nginx setup. I will update here in case I'm able to test with an nginx setup.

pieh: if you need to set custom `GATSBY_WEBPACK_PUBLICPATH` to point to `/gatsby` then there is no sense in testing this as it wouldn't work with current setup as well

pieh:
I tried searching known issues and didn't find any so I think changing to relative is worth trying - if there will be issue reported we can revert this

KyleAMathews:
I don't remember why the code works as it does either. It's possible an earlier version of webpack/rhl needed things like this and then we didn't update.

packages/gatsby/src/utils/webpack.config.js
palerdot: @pieh we can test if the ending `/` part and the default `GATSBY_WEBPACK_PUBLICPATH` takes effect instead of fallback relative path `/` I guess? I'm not sure how else this part can be tested. I don't find any relevant test cases for this piece of code too. But I'm not familiar with codebase and might have missed something.

pieh: > Do you mean existing logic with manual url construction also might not work when public path set to `/gatsby`?

Yes - right now we construct url to be `http(s)://host:port/` so if user visits via `mysite.local/gatsby` then this wouldn't work and user would need to set `GATSBY_WEBPACK_PUBLICPATH` anyway, and this PR doesn't remove `GATSBY_WEBPACK_PUBLICPATH` so nothing changes with this PR

drubin:
@palerdot thanks for the help fixing this.

btw I am not actually on the Gatsby project, I just happened to have the same problem you did. :)

palerdot:
@pieh @drubin Thanks to you guys for the pointers you gave me. I was personally facing this issue as my gatsby server was inside my vagrant machine and I cannot access it from my host because of hmr url problems. Hopefully, this fix helps people with similar setup.

palerdot:
@pieh @drubin @KyleAMathews Is info on `GATSBY_WEBPACK_PUBLICPATH` mentioned anywhere in the docs? I have not seen this environment variable anywhere in the docs. Maybe I didn't look well enough. It should be mentioned somewhere in the docs, so that people with different setup like vagrant or nginx reverse proxy should be able to get up and running without much sweat.

KyleAMathews:
@thebigredgeek was there a reason you picked an ENV variable for the publicpath instead of a command flag? Running `gatsby develop --publicpath /hi-folks/` seems better.

PR #9739, opened 2018-11-06
[starter library] Filter for v2 starters by default

### LINKED ISSUE #7900: starter library - set filter default to Gatsby v2

amberleyromo:
Closes #7900.

Prior work #9250.

CC @chmac


www/src/components/debounce-input.js
```diff
@@ -28,9 +28,10 @@ class DebounceInput extends Component {
 
   render() {
     const { value } = this.state
+    const { initialValue, ...inputAttrs } = this.props
```
amberleyromo: this fixed an error where something was accidentally getting provided to a DOM element that shouldn't.

amberleyromo:
It looks like because I branched off @chmac's working branch from #9250, when I merged master it ran some formatting.

amberleyromo:
@gatsbyjs/website can has review?

www/src/utils/reach-router-state-manager.js
```diff
@@ -1,33 +1,58 @@
...
+    defaultSearchState: {},
+  }
+
+  setUrlState = newState => {
+    const finalState = { ...this.state, ...newState }
+    // update RSSM state
+    this.setState({ ...finalState })
+
+    // sync url to RSSM
     Object.keys(finalState).forEach(function(k) {
```
DSchau: Maybe use reduce here? Something like:

```jsx
// note: there may be more complexity here
// but it seems easier to handle the truthy case and include, rather than
// falsy case(s) and exclude
const params = Object.keys(finalState).reduce((merged, key) => {
  if (finalState[key]) {
    merged[key] = finalState[key]
  }
  return merged
}, {})

return navigate(`${location.pathname}?${queryString.stringify(params)}`)
```

www/src/utils/reach-router-state-manager.js
```diff
@@ -1,33 +1,58 @@
-import React from "react"
+import React, { Component, Fragment } from "react"
 import queryString from "query-string"
 import { navigate } from "@reach/router"
 
-// manage your state entirely within the router, so that it's copiable
-// https://gist.github.com/sw-yx/efd9ee71669413bca6a895d87e30742f
+const emptySearchState = { s: ``, c: [], d: [], v: [], sort: `recent` }
```
DSchau: I don't think this is new in this PR, but why the single letter params here? It's harder to understand what's going on here!

amberleyromo: Yeah, this isn't new to the PR. I'll leave it for now, but I def see your point.

www/src/utils/reach-router-state-manager.js
amberleyromo: I like it, done. The previous check eliminated teh "sort" param, so I added that, so it'll be consistent with what was before. If we decide to make sort robust later, we can take out / tweak it then.

PR #9740, opened 2018-11-06
Add some options to render function

haschema:
I set `throwOnError` to `false` and `strict` to `warn`. Gatsby can ignore katex math formula syntax error when `develop` or `build`

### DIFF
--- a/packages/gatsby-remark-katex/src/index.js
+++ b/packages/gatsby-remark-katex/src/index.js
@@ -7,6 +7,9 @@ module.exports = ({ markdownAST }) => {
     node.type = `html`
     node.value = katex.renderToString(node.value, {
       displayMode: false,
+      throwOnError: false,
+      errorColor: "#cc0000",
+      strict: "warn",
     })
   })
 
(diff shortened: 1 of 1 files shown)


pieh:
> Can you make this plugin be able to config in `gatsby-config.js`? It'd be better.

Yes, please. You can change
```diff
-module.exports = ({ markdownAST })
+module.exports = ({ markdownAST }, pluginOptions = {})
```

and then spread `pluginOptions`:
```js
katex.renderToString(node.value, {
      displayMode: true,
      ...pluginOptions,
    })
```

then you will be able to pass your configuration to plugin:
```
plugins: [
  {
    options: {
      plugins: [
		{
          resolve: `gatsby-remark-katex`,
		  options: {
		    throwOnError: false,
		    errorColor: `#cc0000`,
		    strict: `warn`,
          },
		},
      ],
    },
  },
],
```

haschema:
I found that `throwOnError` had some bugs in KaTeX 0.9.0 ([here](https://github.com/Khan/KaTeX/pull/1169)). The plugin should update KaTeX version to 0.10.0

pieh:
Maybe we should move move `katex` to peerDependency and then users could decide what version they are using. We are doing with some plugins already. We would need to  update installation notes in README then

LekoArts:
Hi @haschema 👋 Do you have time to address the comments from `pieh`? Namely adding the pluginOptions and moving katex to peerDependency (+ changing install instructions on README)

PR #9741, opened 2018-11-06
Update index.js

haschema:


DSchau:
Actually going to close in favor of #9740

PR #9750, opened 2018-11-06
chore: manual eslint fixes

pieh:
this fixes problems that `eslint --fix` can't autofix

### DIFF
--- a/packages/gatsby/cache-dir/load-directly-or-404.js
+++ b/packages/gatsby/cache-dir/load-directly-or-404.js
@@ -36,7 +36,10 @@ export function getRedirectUrl(path) {
 export default function(resources, path, replaceOnSuccess = false) {
   return new Promise((resolve, reject) => {
     const url = getRedirectUrl(path)
-    if (!url) return reject(url)
+    if (!url) {
+      reject(url)
+      return
+    }
 
     // Always navigate directly if a custom 404 page doesn't exist.
     if (!resources) {
--- a/scripts/add-npm-owner/index.js
+++ b/scripts/add-npm-owner/index.js
@@ -1,4 +1,6 @@
 #!/usr/bin/env node
+const util = require(`util`)
+const exec = util.promisify(require(`child_process`).exec)
 const argv = require(`yargs`)
   .command(`$0 <user>`, `Add new owner to gatsby packages`)
   .help().argv
--- a/www/src/views/community/community-header.js
+++ b/www/src/views/community/community-header.js
@@ -6,33 +6,33 @@ import Checkmark from "./check.svg"
 import Button from "../../components/button"
 import ArrowForwardIcon from "react-icons/lib/md/arrow-forward"
 
-const CommunityHeaderLink = ({ linkTo, children }) => (
-  <li
-    css={{
-      display: `flex`,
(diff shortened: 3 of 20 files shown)


pieh:
@lipis `lint` is green! :D still some warning, but I left them for future to not delay PR

www/src/views/community/community-header.js
```diff
@@ -6,33 +6,33 @@ import Checkmark from "./check.svg"
...
+//     >
+//       {children}
+//     </Link>
+//   </li>
+// )
 
 class CommunityHeader extends Component {
   render() {
-    const { forHire, hiring, submissionText } = this.props
+    const { /*forHire, hiring,*/ submissionText } = this.props
```
pieh: decided to comment out "unused vars" as rest of the code that was using that is commented - presumably can be used in future?

www/src/views/starter-library/starter-list.js
```diff
@@ -171,6 +170,7 @@ const StartersList = ({ urlState, starters, count, sortRecent }) => {
       </div>
     )
   }
+  return <React.Fragment />
```
pieh: not sure here if proper - there are `if`s check before that so I think we need to return react element here?

packages/gatsby/cache-dir/load-directly-or-404.js
```diff
@@ -36,7 +36,10 @@ export function getRedirectUrl(path) {
 export default function(resources, path, replaceOnSuccess = false) {
   return new Promise((resolve, reject) => {
     const url = getRedirectUrl(path)
-    if (!url) return reject(url)
+    if (!url) {
+      reject(url)
+      return
```
DSchau: This may as well just return the rejection, right?

scripts/add-npm-owner/index.js
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env node
+const util = require(`util`)
+const exec = util.promisify(require(`child_process`).exec)
```
DSchau: This should fix some things 🎉

www/src/views/starter-library/starter-list.js
DSchau: I'd maybe tweak this, e.g.

```jsx
if (!count) {
  return null; // or empty state from above?
}

return (

); // regular flow 
```

amberleyromo: i think we could just return null

packages/gatsby/cache-dir/load-directly-or-404.js
pieh: having them separate fixes inconsistent return  eslint rule

DSchau: Thought it might be because of the inline return 🤷‍♂️ Yeah, this is fine either way!

www/src/views/starter-library/starter-list.js
pieh: changed to `null`

www/src/views/community/community-header.js
lipis: commented out code should be removed in general from master.

PR #9758, opened 2018-11-06
feat(gatsby-image): Whitelist `crossorigin` prop for pass through to `img` tag

wconnorwalsh:
The `crossorigin` prop is occasionally needed on the root `img` tag to enable CORS requests. 

This PR passes that prop from `Image` to `img`, just as it already passes the `alt` and `title` props.


wardpeet:
lets revisit this in gatsby v3 where we can do lots of optimisations.

packages/gatsby-image/src/__tests__/index.js
```diff
@@ -55,12 +56,13 @@ describe(`<Img />`, () => {
...
   })
 
-  it(`should have correct src, title and alt attributes`, () => {
+  it(`should have correct src, title, alt, and crossOrigin attributes`, () => {
     const imageTag = setup().querySelector(`picture img`)
     expect(imageTag.getAttribute(`src`)).toEqual(`test_image.jpg`)
     expect(imageTag.getAttribute(`srcSet`)).toEqual(`some srcSet`)
     expect(imageTag.getAttribute(`title`)).toEqual(`Title for the image`)
     expect(imageTag.getAttribute(`alt`)).toEqual(`Alt text for the image`)
+    expect(imageTag.getAttribute(`crossOrigin`)).toEqual(`anonymous`)
```
wardpeet: i'll have to validate this still before merging, pretty sure it needs to be `crossorigin` instead of `crossOrigin`

wardpeet: nvm getAttribute is case insensitive

wardpeet:
@wconnorwalsh thank you for your patience! This went way to long! thanks for adding this feature!

PR #9767, opened 2018-11-07
Feat/feature webinar on banner

amberleyromo:
Feature upcoming webinar on banner.


fk:
> Right now it doesn't scale down nicely 

You mean the overflow scroll?

> whatever quick changes you want.

Fixed quote marks and chose to remove the IMO redundant "fallback" content in `banner.js`.

PR #9768, opened 2018-11-07
fix(source-contentful): fix argument arity

macklinu:
This PR fixes an issue with `gatsby-source-contentful` where the internal `prepareStructuredTextNode()` was trying to call an undefined function.

You can see here that the function defines 5 expected arguments:

https://github.com/gatsbyjs/gatsby/blob/19f3d5a70cc857229e1360ecef486dbbba0093fc/packages/gatsby-source-contentful/src/normalize.js#L196-L221

The 4th argument is `createNode` (which is unused). And the only usage I could find of this function receives 4 arguments (with `createNodeId` being the last argument):

https://github.com/gatsbyjs/gatsby/blob/19f3d5a70cc857229e1360ecef486dbbba0093fc/packages/gatsby-source-contentful/src/normalize.js#L415-L420

### DIFF
--- a/packages/gatsby-source-contentful/src/normalize.js
+++ b/packages/gatsby-source-contentful/src/normalize.js
@@ -193,13 +193,7 @@ function prepareTextNode(node, key, text, createNode, createNodeId) {
   return textNode
 }
 
-function prepareStructuredTextNode(
-  node,
-  key,
-  content,
-  createNode,
-  createNodeId
-) {
+function prepareStructuredTextNode(node, key, content, createNodeId) {
   const str = JSON.stringify(content)
   const structuredTextNode = {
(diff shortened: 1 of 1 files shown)


macklinu:
I'm working with Rich Text nodes in a Contentful space, and there isn't a unit test for this currently that I can tell in the `normalize.js` file, so I'm testing changes locally now. Will push up another commit or comment if something isn't working with this PR.

macklinu:
Confirmed by modifying the `gatsby-source-contentful` package in my project's node_modules directory by removing the `createNode` argument, and I am now seeing my rich text models in the GraphQL inspector after running `gatsby develop`. 👍 

image

PR #9781, opened 2018-11-07
fix: linting check on azure/appveyor

pieh:
seems like current (bit whacky) patterns don't play nice on windows for some reason, let's try different ones

note - there is currently unrelated lint error in `www` so linting should fail - but it should have 1 error instead of thousands of errors on azure/appveyor - with this https://github.com/gatsbyjs/gatsby/pull/9780 should be clear


.prettierignore
```diff
@@ -16,9 +16,10 @@ www/public
 **/.cache
 
 # ignore built packages
-packages/*/*.js
-packages/gatsby/cache-dir/commonjs
-packages/gatsby/dist
+packages/**/*.js
+!packages/gatsby/cache-dir/**/*.js
+!packages/*/src/**/*.js
+packages/gatsby/cache-dir/commonjs/**/*.js
```
DSchau: ugh, this sucks :( necessarily evil I suppose.

.gitattributes
```diff
@@ -1,3 +1,4 @@
 **/*.js.snap text eol=lf
 **/__testfixtures__/** text eol=lf
 **/__tests__/fixtures/** text eol=lf
+**/*.md text eol=lf
```
DSchau: This looks good; was there a reason we didn't just add this for all files?

Issue #9747, opened 2018-11-06
Add Content / Copy for .org Homepage - Ecosystem section


amberleyromo:
## Goal

As we're going through the .org redesign, we'll be updating the copy on each of the pages. We want to make sure that we have input and agreement on all sections of homepage copy.

## Next steps

We need to finalize the copy for the Ecosystem homepage section: https://github.com/gatsbyjs/gatsby/issues/9709

@shannonbux @lindaleebumblebee notable questions/needs include:
- The intro under "Plugins & Starters" doesn't exist yet
- The plugins / starters / external resources blurbs mirror https://www.gatsbyjs.org/ecosystem -- is there any reason you'd want them to be different, or is that fine.

### Needed copy
- [x] Main intro copy
  - With over 527 plugins, 86 starters, guides on how to build your own plugins and starters, and dozens of external tutorials and podcasts, the Gatsby ecosystem has tools to help you build your Gatsby projects.

### Approvals
- [x] @lindaleebumblebee 
- [ ] @shannonbux

shannonbux:
I think the copy looks great. 

One question:
Upon first glance, I couldn't remember the difference between "plugin docs" and "plugin authoring." 

For the homepage overall intro, what about:

With over 527 plugins, 86 starters, guides on how to build your *own* plugins and starters, and dozens of external tutorials and podcasts, the Gatsby ecosystem has tools to help you build your Gatsby projects.

amberleyromo:
What about "Using plugins" and "Creating plugins" instead? 

Moving the intro description to the top-level comment.

shannonbux:
oh yeah, I like that!

lindaleebumblebee:
Let's see how the text looks when it's in context (in the design). It might be a bit long. If so, I suggest this:
With over 527 plugins, 86 starters, and dozens of tutorials, the Gatsby ecosystem has tools to help you build your Gatsby projects.

amberleyromo:
Thanks to you both! 

I'm going to mark this issue closed / consider copy final for the initial homepage PR.

fk:
🙏

PR #9783, opened 2018-11-07
feat(www): add Ecosystem section to Homepage

greglobinski:
* add featured items to `IndexRoute`'s pageContext
* add new version of `Button`//
* add `className` props to Ecosystem components for restyling
* refactor `ecosystem/ecosystem-featured-items.js` for restyling 
* create universal `HomepageSection` component
* create `HomepageEcosystem` component with all content of Ecosystem section
* change GraphQL query for `IndexRoute` (+ refactor data destructing )
* add new color `Lemon` to the pallete, the featured items cards colors used in the Ecosystem section are defined in the code. (@fk would you have the also in the palette?)


greglobinski:
* Please take a close look at the restyled/reused Emotion components in https://github.com/gatsbyjs/gatsby/pull/9783/files#diff-623d03a8cfeefafaac1d88a99eb22dd2  derived from `/ecosystem/*` components in the context of our emerging 'way of styling'

* the 'fake' scroll effect is not implemented, it should probably wait till https://github.com/gatsbyjs/gatsby/pull/9606 will be merged

amberleyromo:
@greglobinski revisit the card hover state?

screen shot 2018-11-07 at 10 00 44 am

greglobinski:
@amberleyromo , yeah I think that I fixed it but seams that not,

amberleyromo:
A few content changes, as well (from #9747)
- [ ] Change the intro text (below Plugins & Starters) to:

> With over 527 plugins, 86 starters, and dozens of tutorials, the Gatsby ecosystem has tools to help you build your Gatsby projects.

(I agree with @lindaleebumblebee, i think it needs to be shorter, let's go with this one)

- [ ] Change the text of two of the links under `Plugins` section
  - [ ] "Plugin Docs" => "Using Plugins"
  - [ ] "Plugin Authoring" => "Creating Plugins"
- [ ] For consistency, change the text of the link under `Starters` section also
  - [ ] "Starter Docs" => "Using Starters"

greglobinski:
@amberleyromo 

* hover bug fixed
* intro & label texts changed

Should I also make a PR with the changes for the link labels on the Ecosystem page?

amberleyromo:
This looks great to me!

I'd say go ahead and make the content changes to the ecosystem landing page as part of this PR -- it's not a huge change. Thanks for thinking about it.

I'm good to go on this personally, but would also love @DSchau's eye also on your review request re: styling... style.

> Please take a close look at the restyled/reused Emotion components in https://github.com/gatsbyjs/gatsby/pull/9783/files#diff-623d03a8cfeefafaac1d88a99eb22dd2 derived from /ecosystem/* components in the context of our emerging 'way of styling'

greglobinski:
@amberleyromo 

* move scrollers intersectionObserver to an outer module, so both pages Index and Ecosystem use the same code `www/src/utils/scrollers-observer.js`
* refactor  `featuredItems` observer to `scrollerWithLead` observer, it's ready now to work with another scrollers which will appear in coming Homepage sections (e.g. Blog)
* change buttons labels on Ecosystem page for consistency with Homepage Ecosystem section

www/src/components/homepage/homepage-ecosystem.js
```diff
@@ -0,0 +1,226 @@
...
+    overflow-x: auto;
+  }
+
+  ${presets.Hd} {
+    margin-right: 3rem;
+  }
+`
+
+const FeaturedItemsList = styled(EcosystemFeaturedItemsListBase)`
+  width: ${props => `calc(320px * ${props.numberOfItems})`};
```
DSchau: I don't think this is anything new in this PR, but could you explain why we couldn't just use overflow scrolling here, instead of dynamically setting the width of the container element?

I noticed this above, as well (e.g. `85vw * numberOfItems`)

www/src/components/homepage/homepage-ecosystem.js
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    :after {
+      bottom: 0;
+      content: "${props => props.item.type}";
+      color: ${props =>
+        props.item.type === `Starter` ? ` #0a75c2` : `#9e6100`};
+      font-family: ${options.headerFontFamily.join(`,`)};
+      font-size: 0.8rem;
+      left: 0;
```
DSchau: Total nit, but I find this hard to follow. I usually group properties related to positioning, e.g.

```css
position: absolute;
bottom: 0;
left: 0;

/* Rest of properties */
```

www/src/pages/index.js
```diff
@@ -19,10 +19,79 @@ import FuturaParagraph from "../components/futura-paragraph"
...
+  componentDidMount() {
+    setupScrollersObserver()
+  }
+
+  componentWillUnmount() {
+    unobserveScrollers()
+  }
+
+  combineEcosystemFeaturedItems = ({ starters, plugins }) => {
+    const combinedItems = []
```
DSchau: Just a suggestion:

```js
return new Array(3).fill(undefined).reduce((merged, _, index) => {
  return merged.concat([
    starters[index],
    plugins[index]
  ])
}, [])
```

www/src/utils/scrollers-observer.js
```diff
@@ -0,0 +1,60 @@
+let observer
+let scrollers = []
+
+export const SCROLLER_CLASSNAME = `scrollerWithLead`
+
+export const setupScrollersObserver = () => {
```
DSchau: I like abstracting this for re-use, a lot. Nice work!

www/src/pages/index.js
DSchau: Oh and maybe do something like this, too?

```jsx
combineEcosystemFeaturedItems = ({ starters, plugins }, numFeatured = 3) => {
 // use numFeatured instead of "3"
```

www/src/components/homepage/homepage-ecosystem.js
greglobinski: I do not remember :) Seams that you are right I can set the with of items not the container. I will try.

www/src/components/homepage/homepage-ecosystem.js
greglobinski: I usually order properties alphabetically, but I can adjust

www/src/pages/index.js
greglobinski: :+1:

www/src/components/homepage/homepage-ecosystem.js
fk: Just saw this in the mail and thought "Greg maybe had been following what I've been doing on `www`". I also order alphabetically…when I think of it that is 😉(VS Code plugins doing that which I checked have a hard time with multiple lines).

I still often find myself keeping `position`-related stuff together initially, and when I get to reorder them I still often shrug, but in general I find alphabetic ordering to have it's advantages, and an OK "least common denominator". 🤐 😸

fk: > I find this hard to follow.

Def. is!

www/src/components/homepage/homepage-ecosystem.js
greglobinski: @DSchau 
I figured out why I set the `width` of the container :) Without it I am not able to get a nice looking right-padding at the end of the scroller like below. Without it the scroller scrolls only to the moment the last item is full visible but without any space between it and the screen's border, adding margin to the last item or right padding to the container doesn't work then, so because of that I set the width of the container.  

screenshot from 2018-11-09 13-53-59

fk: Ah yeah, overflow and whitespace. Should we try and get around that by adding another wrapper container and applying the right padding to that? I think we do something similar in `banner.js`.

greglobinski: @fk I don't  think so, the screen shows how the things look like now and I think that's OK, but I can't  get that look without setting `width` for container. Sure we can add additional inner container, but why if we get it working without it.

DSchau: @greglobinski just to be clear here, you're setting width because a wrapping div adds additional markup? I think I prefer a wrapping div here, that seems a little nicer than explicitly setting width!

greglobinski: @DSchau I'm setting `width` for the container to be able to keep padding-right of it. I assume that it could be possible also by adding inner container with needed padding as @fk suggested. If you and Flo prefer that way I will adjust.

greglobinski: @DSchau In any case I have to explicit set `width` although for container or for its children

www/src/components/homepage/homepage-ecosystem.js
jlengstorf: I always go alphabetical, too. I think there's a good argument for contextual grouping, too, but that assumes we all think of the same rules in the same context. 🤷‍♂️

fk:
@greglobinski Sorry for not seeing this earlier.

> add new color Lemon to the pallete, the featured items cards colors used in the Ecosystem section are defined in the code. (@fk would you have the also in the palette?)

Yay 🍋! ;-)

Yes, I would add those featured item card colors to `colors`, too. I like to think this gives us a better starting point for eventually refactoring `colors` — I think we should abandon the "named" color model sooner or later, and move to base color and shades similar to what Material Design does or is [outlined here](https://refactoringui.com/previews/building-your-color-palette/), or a simpler and tighter definition like http://www.carbondesignsystem.com/guidelines/color/swatches; in any case, abandon names like `lavender` or `wisteria`.

greglobinski:
> I think we should abandon the "named" color model sooner or later,
I agree with that, but for now what names I should use to add the Starter/Plugins colors to the palette?
 @fk

www/src/components/homepage/homepage-ecosystem.js
greglobinski: @DSchau I'm working now on the Blog section, there is also a scroller. I'm implementing it without explicitly setting a `width` of the container only of the children (plus a spacer element as a last child of the container to get the proper padding at the end). When I will finish I plan to refactor style of all scrollers to the separate/imported component for re-use, so then also this scroller will be refactored in the way you suggest.

www/src/utils/colors.js
```diff
@@ -5,6 +5,7 @@ const colors = {
   // @see https://www.figma.com/file/J6IYJEtdRmwJQOrcZ2DfvxDD/Gatsby
   gatsby: `#663399`, // was #744c9e
   gatsbyDark: `#442266`,
+  lemon: `#ffDF37`,
```
fk: Nit: Let's stay consistent here and use lowercase.

www/src/components/homepage/homepage-ecosystem.js
fk: > I always prefer less HTML markup

Generally signing that!
In this case though, IMO clarity wins — as far as I understand the code, by defining a container width based on the width of its items (+ _x_ for the whitespace to the right of the last card I suppose), we are creating an unnecessary dependency among the two.

Also it seems like we do _not_ need to add an additional container here to set the right padding/whitespace for the `overflow` scroll. I even think `FeaturedItemsList` in `homepage-ecosystem` already defines that right padding…

Hmmm, actually…instead of trying to explain my thinking, let me see if I can come up with a quick PoC patch or something.

fk: So yeah, AFAIU we don't need an additional wrapper here, PoC:

```diff
diff --git a/www/src/components/ecosystem/ecosystem-featured-items.js b/www/src/components/ecosystem/ecosystem-featured-items.js
index 60178cf16..4d7b8913f 100644
--- a/www/src/components/ecosystem/ecosystem-featured-items.js
+++ b/www/src/components/ecosystem/ecosystem-featured-items.js
@@ -29,7 +29,6 @@ export const ListBase = styled(`ul`)`
   list-style: none;
   margin: 0;
   padding: 0 calc(${rhythm(options.blockMarginBottom)} - 5px) 4px;
-  width: ${props => `calc(85vw * ${props.numberOfItems})`};
 `
 
 const List = styled(ListBase)`
diff --git a/www/src/components/homepage/homepage-ecosystem.js b/www/src/components/homepage/homepage-ecosystem.js
index b5480a9f4..baa66d02f 100644
--- a/www/src/components/homepage/homepage-ecosystem.js
+++ b/www/src/components/homepage/homepage-ecosystem.js
@@ -83,9 +83,9 @@ const FeaturedItems = styled(EcosystemFeaturedItemsRootBase)`
 
 const FeaturedItemsList = styled(EcosystemFeaturedItemsListBase)`
   padding: 0 calc(${rhythm(options.blockMarginBottom)} - 7px) 0;
+  display: inline-flex;
 
   ${presets.Tablet} {
-    width: ${props => `calc(320px * ${props.numberOfItems})`};
   }
 
   ${presets.Desktop} {
@@ -97,6 +97,7 @@ const FeaturedItemsList = styled(EcosystemFeaturedItemsListBase)`
 
 const FeaturedItem = styled(EcosystemFeaturedItem)`
   margin: 0 6px 6px 0;
+  width: 320px;
 
   ${presets.Tablet} {
     border-bottom: none;
```

fk: Oops, sorry @greglobinski, only saw your comment https://github.com/gatsbyjs/gatsby/pull/9783#discussion_r233024842 now. 🙏

SlowFlo. 🐌 🤕 🤧

greglobinski: @fk Just removing the two lines and adding the another two works for you? Although I missed something or... I don't know :). I have a meeting in a minute but I will check that after it. If that works I will commit the changes :+1:

fk: 👌Yeah, was looking good to me and even didn't seem to change the > Tablet (or Desktop, can't remember anymore already) behavior of cards and container. Then again I'm only half concious… ;-/

greglobinski: @fk Thank you very much 🙏 The knowledge I was missing is `display: inline-flex`. I was fighting with `display: flex`, after changing to `inline` everything works as it should.

fk: 🙌 👍

PR #9792, opened 2018-11-07
feat: pass pathname to replaceRenderer and onPreRenderHTML SSR APIs

jafaircl:
For some situations (e.g. implementing AMP pages), a user may need to conditionally remove or add elements in the head or body of a page. Normally, this could be done when the page or node is created. But, styles and scripts may be added later in the lifecycle. The `onRenderBody` SSR API provides a `pathname` argument that could be used but does not allow the removal of existing elements. Using `replaceHeadComponents` or `replaceBodyComponents` in `onPreRenderHTML` would seem to be ideal. But, this API doesn't expose the `pathname` argument.

The following example for combining inline style elements in the head of the document is not currently possible but would be if pathname were exposed.

```js
import React from 'react'
import Helmet from 'react-helmet'

export const onPreRenderHTML = (
  { getHeadComponents, replaceHeadComponents, pathname },
  pluginOptions
) => {
  const headComponents = getHeadComponents()
  const styles = headComponents.reduce((str, x) => {
    if (x.type === 'style') {
      str += x.props.dangerouslySetInnerHTML.__html
    }
    return str
  }, '')
  const regex = new RegExp(pluginOptions.pathRegex || '\/amp\/')
  if (regex.test(pathname)) {
    replaceHeadComponents([
      <Helmet>
        <style id="amp-custom">
          {styles}
        </style>
      </Helmet>,
      ...headComponents.filter(x => x.type !== 'style'),
    ])
  }
}
```
I'm not sure if this needs to be submitted as an RFC or not. It does slightly modify an API but it seems very minor.

### DIFF
--- a/packages/gatsby/cache-dir/static-entry.js
+++ b/packages/gatsby/cache-dir/static-entry.js
@@ -180,6 +180,8 @@ export default (pagePath, callback) => {
     setPreBodyComponents,
     setPostBodyComponents,
     setBodyProps,
+    pathname: pagePath,
+    pathPrefix: __PATH_PREFIX__,
   })
 
   // If no one stepped up, we'll handle it.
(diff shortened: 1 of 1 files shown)


pieh:
I think it's perfectly reasonable thing to do

Issue #9766, opened 2018-11-07
README of Unit Testing section refers to something that doesn't exist


mjkrause:
## Description

Check line _Back to the Jest config, you can see the next option is testRegex. This is the pattern telling Jest which files contain tests. The pattern above matches any .js file inside a `__tests__` directory,..._ in gatsby/docs/docs/unit-testing.md.
There is no variable named `testRegex` in the above example of `jest.config.js`. I have seen somewhere below in that file, though. In any case, that reference doesn't seem to be correct.

### Steps to reproduce

see description

### Expected result

Remove that reference, or put `testRegex = ...` into the example.

kakadiadarpan:
@mjkrause the best way of going through the Gatsby documentation would be through www.gatsbyjs.org, the links would not work as expected in the Github's README.md

For unit testing, please check the following link on gatsbyjs.org: https://www.gatsbyjs.org/docs/unit-testing/

LekoArts:
No, you’re misunderstanding something here @kakadiadarpan
The text mentions a testRegex that was there before but @dschau removed that from the config file with his last commit

kakadiadarpan:
I can see the `testRegex` in the jest config file. Please check the attached screenshot below:

<img width="835" alt="screenshot 2018-11-07 at 1 15 18 pm" src="https://user-images.githubusercontent.com/10805593/48117477-5ccdec00-e28f-11e8-9c07-ac5b30642bfd.png">

DSchau:
Yeah, whoops! Easy fix to remove and re-arrange that paragraph. I did remove it because it wasn't necessary for JS files, but is still necessary for Typescript.

Someone want to tackle this? If not, one of us will get to it as soon as we can!

RobinCsl:
I believe this issue can be closed now, since my PR was merged.
cc @DSchau @LekoArts @shannonbux

PR #9795, opened 2018-11-07
Update unit-testing.md

RobinCsl:
Fixes issue #9766

### DIFF
--- a/docs/docs/unit-testing.md
+++ b/docs/docs/unit-testing.md
@@ -60,32 +60,28 @@ const babelOptions = {
 module.exports = require("babel-jest").createTransformer(babelOptions)
 ```
 
-Back to the Jest config, you can see the next option is `testRegex`. This is the
-pattern telling Jest which files contain tests. The pattern above matches any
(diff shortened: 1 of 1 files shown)


shannonbux:
@LekoArts and @DSchau want to do a brief review here and then merge?

RobinCsl:
@shannonbux Do you have any example in the current Gatsby docs of what you mean? Do you mean that, for instance, the TypeScript section is somewhat optional for most people so that fact should be clearer?

shannonbux:
Hi @RobinCsl in a second read through, I think what I realized is that the "setting up your environment" section is very long and it would be nice to have level 3 headers to break it up by options. Just an idea if you're interested in working on something like that!

RobinCsl:
Hi @shannonbux, so for instance, you would see all of the main sections of `Unit testing` in the side navigation bar. Do I understand your idea correctly? If that's the case, I could look into that!
screenshot 2018-11-08 at 09 31 20

LekoArts:
She means that you use `<h3>` headings before the different options. The thing you screencaptured is the sidebar and I don't think it would make sense to further split this up there.

PR #9800, opened 2018-11-08
chore: Upgrade Prettier to 1.15.1 version

lipis:


pieh:
This will be handled by https://github.com/gatsbyjs/gatsby/pull/9670, so let's not duplicate that

pieh:
also Azure pipelines (`unit_tests_windows` which also run lint) seems to have lot of problems with this :/

PR #9803, opened 2018-11-08
chore: Refactor build node connections

Moocar:
Another refactoring! The plus side of all these is more readable code (hopefully), but they're mostly to ease the loki integration. This one renames a lot of the vars/functions in `build-node-connections` so they're consistent with the actual graphql types and field class names as per https://graphql.org/graphql-js/type/#graphqlobjecttype

Another benefit of this refactor is the ability to create types (connections in this case) one by one instead of having to do them as a whole batch. This will be eventually needed for incremental builds

Update: I slipped a refactor to `build-node-types` in as well.


packages/gatsby/src/schema/__tests__/build-node-types-test.js
```diff
@@ -15,6 +24,19 @@ describe(`build-node-types`, () => {
   }
 
   beforeEach(async () => {
+    const apiRunnerResponse = [
```
pieh: I think we should use `createPageDependency.mockClear()` - without it last test `should create page dependency` will not fail because page dependency for single `child` (see my other comment in `build-node-types`) was created in earlier test

pieh: It also be great if test would check if all kinds of node "links" on RootNodeType create page dependency:
 - `parent`
 - `children`
 - `childX` (with single child nodes of given type X)
 - `childrenX` (with multiple child nodes of given type X)

Moocar: Great catch with that mockClear. I added the extra tests as well. Should be good to go

packages/gatsby/src/schema/build-node-types.js
```diff
@@ -29,170 +29,234 @@ type TypeMap = {
...
+function makeChildResolver(typeName) {
+  return node => node.children.map(getNode).find(nodeIsOfType(typeName))
+}
+
+function buildChildFieldConfigMap(typeName, nodeObjectType) {
+  const fieldName = _.camelCase(`child ${typeName}`)
+  const fieldConfig = {
+    type: nodeObjectType,
+    description: `The child of this node of type ${typeName}`,
+    resolve: makeChildResolver(typeName),
```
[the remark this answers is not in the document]
Moocar: thanks! Fixed

PR #9814, opened 2018-11-08
feat(gatsby-transformer-remark): add options for tableOfContents

### LINKED ISSUE #11452: [gatsby-transformer-remark] TOC depth?
## Summary
Is it possible to set the table of contents depth with gatsby-transformer-remark?

### Environment (if relevant)

  System:
    OS: macOS 10.14.2
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.15.0 - /usr/local/bin/node
  Languages:
  Browsers:
    Chrome: 72.0.3626.81
    Firefox: 61.0.2
    Safari: 12.0.2
  npmPackages:
    gatsby: ^2.0.76 => 2.0.99 
    gatsby-image: ^2.0.20 => 2.0.28 
    gatsby-plugin-manifest: ^2.0.9 => 2.0.14 
    gatsby-plugin-offline: ^2.0.16 => 2.0.21 
    gatsby-plugin-sharp: ^2.0.14 => 2.0.18 
    gatsby-remark-toc: ^1.2.0 => 1.2.0 
    gatsby-source-filesystem: ^2.0.17 => 2.0.18 
    gatsby-transformer-remark: ^2.2.0 => 2.2.1 
    gatsby-transformer-sharp: ^2.1.8 => 2.1.12 
  npmGlobalPackages:
    gatsby-cli: 2.4.8

`gatsby-config.js`: N/A 
`package.json`: N/A 
`gatsby-node.js`: N/A 
`gatsby-browser.js`: N/A 
`gatsby-ssr.js`: N/A

byzanth:
Added the ability to pass options to mdast-util-toc from gatsby-transformer-remark.

closes #11452


packages/gatsby-transformer-remark/README.md
```diff
@@ -81,6 +86,22 @@ Using the following GraphQL query you'll be able to get the table of contents
...
+
+```javascript
+// In your gatsby-config.js
+plugins: [
+  {
+    resolve: `gatsby-transformer-remark`,
+    options: {
+      heading: null,
+      maxDepth: 6,
+      tight: null,
```
pieh: can we group those under `toc` or `tableOfContents` object?

packages/gatsby-transformer-remark/src/extend-node-type.js
```diff
@@ -242,13 +250,29 @@ module.exports = (
...
 
-    async function getTableOfContents(markdownNode, pathToSlugField) {
+    async function getTableOfContents(
+      markdownNode,
+      pathToSlugField,
+      maxDepth,
+      heading,
+      tight
+    ) {
       const cachedToc = await cache.get(tableOfContentsCacheKey(markdownNode))
```
pieh: cache key would need to contain options, otherwise if you query same markdown node with different toc options it will return same output even if options are different

packages/gatsby-transformer-remark/src/extend-node-type.js
```diff
@@ -242,13 +250,29 @@ module.exports = (
...
+      const appliedTocOptions = tocOptions
+      if (maxDepth && _.isNumber(maxDepth) && _.inRange(maxDepth, 0, 7)) {
```
pieh: with `maxDepth` set defaultValue to 6 in graphql field definition - this will always be used over pluginOptions

pieh: you would need to remove `defaultValue` from graphql definition or use pluginOptions to set `defaultValue` there instead of hardcoding values

packages/gatsby-transformer-remark/src/extend-node-type.js
```diff
@@ -435,9 +459,27 @@ module.exports = (
...
+            defaultValue: null,
+          },
+          tight: {
+            type: GraphQLBoolean,
+            defaultValue: false,
+          },
         },
-        resolve(markdownNode, { pathToSlugField }) {
-          return getTableOfContents(markdownNode, pathToSlugField)
+        resolve(markdownNode, { pathToSlugField, maxDepth, heading, tight }) {
```
pieh: I would just pass `args` (without destructuring it here) as is to `getTableOfContents` function and destructure it there

packages/gatsby-transformer-remark/src/extend-node-type.js
```diff
@@ -242,13 +250,29 @@ module.exports = (
...
+    ) {
       const cachedToc = await cache.get(tableOfContentsCacheKey(markdownNode))
+      const appliedTocOptions = tocOptions
+      if (maxDepth && _.isNumber(maxDepth) && _.inRange(maxDepth, 0, 7)) {
+        appliedTocOptions.maxDepth = maxDepth
+      }
+      if (heading && _.isString(heading)) {
+        appliedTocOptions.heading = heading
+      }
+      if (tight && _.isBoolean(tight)) {
```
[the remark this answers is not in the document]
byzanth: Values via GraphQL seem fine, but you can configure them via gatsby-config.js as well (and enter wrong values). I would rather leave that check in, just to be shure nothing breaks.

packages/gatsby-transformer-remark/src/extend-node-type.js
byzanth: I hopefully fixed this with the new commit.

packages/gatsby-transformer-remark/src/extend-node-type.js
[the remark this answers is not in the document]
oorestisime: oh i missed the config part

packages/gatsby-transformer-remark/README.md
byzanth: fixed in new commit

packages/gatsby-transformer-remark/src/extend-node-type.js
byzanth: fixed in new commit

packages/gatsby-transformer-remark/src/extend-node-type.js
byzanth: fixed in new commit

packages/gatsby-transformer-remark/README.md
pieh: I mean in options that user sets:
```
plugins: [
  {
    options: {
	  tableOfContents: {
        heading: null,
        maxDepth: 6,
        tight: null,
      }
     },
  },
]
```

packages/gatsby-transformer-remark/src/extend-node-type.js
```diff
@@ -26,6 +26,7 @@ const stripPosition = require(`unist-util-remove-position`)
 const hastReparseRaw = require(`hast-util-raw`)
 
 let fileNodes
+let tocOptionsCacheStr = ``
```
pieh: this won't work - queries are executed in parallel - it's not safe to set global context - you would need to change `tableOfContentsCacheKey` function to accept second param (tocOptions):

```js
const tableOfContentsCacheKey = (node, tocOptions) =>
  `transformer-remark-markdown-toc-${
    node.internal.contentDigest
  }-${pluginsCacheStr}-${pathPrefixCacheStr}`
  }-${pluginsCacheStr}-${pathPrefixCacheStr}-${JSON.stringifu(tocOptions)}`
```

packages/gatsby-transformer-remark/src/extend-node-type.js
pieh:
```suggestion
        resolve(markdownNode, args) {
```

packages/gatsby-transformer-remark/src/extend-node-type.js
```diff
@@ -435,9 +470,23 @@ module.exports = (
...
         },
-        resolve(markdownNode, { pathToSlugField }) {
-          return getTableOfContents(markdownNode, pathToSlugField)
+        resolve(markdownNode, { pathToSlugField, maxDepth, heading, tight }) {
+          return getTableOfContents(markdownNode, {
+            pathToSlugField: pathToSlugField,
+            maxDepth: maxDepth,
+            heading: heading,
+            tight: tight,
+          })
```
pieh:
```diff
-return getTableOfContents(markdownNode, {
-            pathToSlugField: pathToSlugField,
-            maxDepth: maxDepth,
-            heading: heading,
-            tight: tight,
-          })
+return getTableOfContents(markdownNode, args)
```

packages/gatsby-transformer-remark/src/extend-node-type.js
```diff
@@ -242,27 +251,53 @@ module.exports = (
       }
     }
 
-    async function getTableOfContents(markdownNode, pathToSlugField) {
+    async function getTableOfContents(markdownNode, gqlTocOptions) {
+      // fetch defaults
+      let appliedTocOptions = tocOptions
```
pieh:
```suggestion
      let appliedTocOptions = { ...tocOptions, ... gqlTocOptions }
```
and you won't need all those `if` clauses below

packages/gatsby-transformer-remark/src/extend-node-type.js
```diff
@@ -242,27 +251,53 @@ module.exports = (
...
+      }
+      if (gqlTocOptions.pathToSlugField) {
+        appliedTocOptions.pathToSlugField = gqlTocOptions.pathToSlugField
+      }
+      // reset cache string
+      tocOptionsCacheStr = ``
+      // define cache string
+      Object.keys(appliedTocOptions).map(function(key, index) {
+        tocOptionsCacheStr += key + index
+      })
```
pieh: check `tableOfContentsCacheKey` comment

packages/gatsby-transformer-remark/README.md
```diff
@@ -68,7 +68,29 @@ Using the following GraphQL query you'll be able to get the table of contents
...
+{
+  allMarkdownRemark {
+    edges {
+      node {
+        html
+        tableOfContents(
+          pathToSlugField: "frontmatter.path"
+          heading: "only show toc from this heading onwards"
+          maxDepth: 2
+          tight: true
```
pieh: After reading https://www.npmjs.com/package/mdast-util-toc I still don't really understand what `heading` and `tight` options do - can you show example of using it? Or maybe add test that would show using it?

packages/gatsby-transformer-remark/src/extend-node-type.js
```diff
@@ -44,10 +44,18 @@ const headingsCacheKey = node =>
...
   }-${pluginsCacheStr}-${pathPrefixCacheStr}`
-const tableOfContentsCacheKey = node =>
-  `transformer-remark-markdown-toc-${
+const tableOfContentsCacheKey = function(node, appliedTocOptions) {
+  // define toc options cache string as empty
+  let tocOptionsCacheStr = ``
+  // put applied toc options into toc options cache string
+  Object.keys(appliedTocOptions).map(function(key, index) {
+    tocOptionsCacheStr += key + index
+  })
```
pieh: You can make generating of `tocOptionsCacheStr` more compact by just using `const tocOptionsCacheStr = JSON.stringify(appliedTocOptions)` - and compact it even more

packages/gatsby-transformer-remark/README.md
byzanth: After testing a bit, here are my insights to these two options: "heading" searches for a string in the toc headings and renders the toc from that heading onwards. Alas, I think the implementation of this feature seems to be flawed, because the toc is rendered without the given heading. The "tight" option seems to do nothing at all. As these options should be fixed in mdast-util-toc, I would omit additional tests here.

packages/gatsby-transformer-remark/src/extend-node-type.js
byzanth: changed in new commit

packages/gatsby-transformer-remark/README.md
byzanth: Latest insights: "tight" influences "loose" and "spread" attributes in generated toc JSON object. The plugin that uses these attributes for html generation is "hast-util-to-html" and it toggles space after quoted attributes. Default setting for tight is false, so a space is added after quoted attributes by default (which is the desired behaviour). I'll update the readme.

byzanth: ok, I removed tight option completely, as the desired changes in html output seem to get lost somewhere along the way. Also added test for heading option.

LekoArts:
@pieh Anything else left that needs to be done? 😊

byzanth:
Approve, merge, get the t-shirt ;)

pieh:
Thanks for your work on this @byzanth and sorry it took so long to get this merged!

PR #9819, opened 2018-11-08
docs(gatsby-plugin-google-gtag): fix opt out snippet

valse:


pieh:
Ah, you are right - function should probably be named `gtagOptout` but changing function might be breaking change :/

PR #9829, opened 2018-11-09
fix(gatsby-plugin-catch-links): handle SVGAnimatedString href values

### LINKED ISSUE #9816: The links in the "Behind the scenes" documentation does not work
## Description

Tested in **Firefox 64** and **Chrome 70**.

The links in the generated graph of the Gatsby architecture does not work properly.

When hovering over a node (e.g. **source nodes**) the link seams to be interpreted correctly by the browser, but by clicking the node you get routed to this path [0] instead of [1] which was to be expected.
It does however work when you open the link in a new tab (by holding down the ctrl key while clicking the links).

[0] **/docs/behind-the-scenes/[object%20SVGAnimatedString]**
[1] **/docs/node-creation/**

### Steps to reproduce

1. Navigate to the [behind the scenes](https://www.gatsbyjs.org/docs/behind-the-scenes/) documentation
2. Press one of the links in the graph (e.g. the **source nodes** node).

### Expected result

You get routed to this page: https://www.gatsbyjs.org/docs/node-creation/

### Actual result

You end up on a page with the title **Page not found** and that the path in the browser window is this: https://www.gatsbyjs.org/docs/behind-the-scenes/[object%20SVGAnimatedString].

### Environment

This happens in the live documentation hosted on gatsbyjs.org

Bassetts:
Updates `gatsby-plugin-catch-links` to handle SVG anchors.

Fixes #9816

### DIFF
--- a/packages/gatsby-plugin-catch-links/src/catch-links.js
+++ b/packages/gatsby-plugin-catch-links/src/catch-links.js
@@ -119,6 +119,9 @@ export const routeThroughBrowserOrApp = hrefHandler => event => {
   // in React. Creating a new anchor element to ensure host value is present
   const destination = document.createElement(`a`)
   destination.href = clickedAnchor.href
+  if (clickedAnchor.href instanceof SVGAnimatedString) {
+    destination.href = clickedAnchor.href.animVal
+  }
 
   // In IE, the default port is included in the anchor host but excluded from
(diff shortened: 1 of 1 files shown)


DSchau:
```
Successfully published:
 - gatsby-plugin-catch-links@2.0.8
```

Issue #9834, opened 2018-11-09
[gatsby-source-drupal] Images URI / URL handling changes in JSON API 2.x


dcorb:
## Description

With Gatsby 2.x and latest Drupal module JSON API 2.x version, it's not possible anymore to take advantage of `gatsby-transformer-sharp` images query
``` 
JSON API 1.x on Drupal <8.5.x
data.attributes.url
JSON API 1.x on Drupal 8.5.x
data.attributes.url
OR
data.attributes.uri.url
JSON API 2.x on Drupal >=8.5.x
data.attributes.uri.url
```
From https://www.drupal.org/node/2982209 

### Steps to reproduce

Have a Drupal site with JSON API 2.x and a content type with image fields.
Configure gatsby-source-drupal, and install the "sharp" related plugins.
Run `gatsby develop` and inspect the schema, (try to query images). You won't see 'localFile' 

### Expected result

My expectation is to use the same style of image query like the drupal example, but I didn't see "localFile" field
https://github.com/gatsbyjs/gatsby/blob/master/examples/using-drupal/src/templates/recipe.js

```
relationships {
        category {
          name
        }
        image {
          relationships {
            imageFile {
              localFile {
                childImageSharp {
                  fluid(maxWidth: 470, maxHeight: 353) {
                    ...GatsbyImageSharpFluid
                  }
                }
              }
            }
          }
        }
      }
```

### Actual result

No localFile field in the query.

### Environment

``` 
  System:
    OS: macOS 10.14
    CPU: x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
    Shell: 2.7.1 - /usr/local/bin/fish
  Binaries:
    Node: 10.9.0 - /usr/local/bin/node
    Yarn: 1.9.4 - /usr/local/bin/yarn
    npm: 6.4.1 - /usr/local/bin/npm
  Browsers:
    Chrome: 70.0.3538.77
    Firefox: 63.0
    Safari: 12.0
  npmPackages:
    gatsby: ^2.0.37 => 2.0.37 
    gatsby-image: ^2.0.19 => 2.0.19 
    gatsby-plugin-favicon: ^3.1.4 => 3.1.4 
    gatsby-plugin-manifest: ^2.0.7 => 2.0.7 
    gatsby-plugin-offline: ^2.0.11 => 2.0.11 
    gatsby-plugin-react-helmet: ^3.0.1 => 3.0.1 
[...]

PR #9835, opened 2018-11-09
feat(gatsby-source-drupal): add support  for Drupal JSON API 2.x File URI format

dcorb:
Description of the issue
https://github.com/gatsbyjs/gatsby/issues/9834

(If this approach is correct, I could add a test for this case)


packages/gatsby-source-drupal/src/gatsby-node.js
```diff
@@ -187,8 +187,13 @@ exports.sourceNodes = async (
         node.internal.type === `file__file`
       ) {
         try {
+          let fileUrl = node.url
+          if (typeof node.uri === `object`) {
+            // Support JSON API 2.x file URI format https://www.drupal.org/node/2982209
+            fileUrl = node.url.uri
```
kakadiadarpan: Shoudn't this be like this:
```suggestion
            fileUrl = node.uri.url
```

dcorb:
Thanks @kakadiadarpan . Suggestion applied

pieh:
This looks good. Adding test would be great - we would need to add new entry in this fixture https://github.com/gatsbyjs/gatsby/blob/4db5cdcce9b78db65454935e322f3d61d1c82179/packages/gatsby-source-drupal/src/__tests__/fixtures/file.json that uses new URI format

And then in actual test mock `createRemoteFileNode` and check if it was called with expected urls (we don't have actual testing if files are being downloaded right now even with previous format)

dcorb:
👍

PR #9839, opened 2018-11-09
fix: development proxy throws unhandled exceptions

### LINKED ISSUE #6771: Development proxy throws unhandled exceptions
## Description

Development proxy request errors cause `gatsby` to crash due to an unhandled exception.

It would be nice to keep the `gatsby develop` process alive to test site failover

### Steps to reproduce

Any `gatsby` site with with a development proxy provided in `gatsby-config.js`.

If the proxy url provided is set to fail (proxy server not running, invalid url), `gatsby develop` will crash when a request is made to the proxy.

### Expected result

Exceptions should be handled and reported to the developer. `gatsby develop` process remains alive.

### Actual result
`gatsby develop` process exits on an unhandled exception. 

### Environment
`gatsby info --clipboard` not available in v1?

Shouldn't be required

`gatsby-config.js`: N/A 
`package.json`: N/A 
`gatsby-node.js`: N/A 
`gatsby-browser.js`: N/A 
`gatsby-ssr.js`: N/A 

### Notes

I have tried wrapping `req.pipe(request(proxiedUrl)).pipe(res)` in a `try...catch`, but the error slips through.

ValeraS:
Closes #6771

### DIFF
--- a/packages/gatsby/src/commands/develop.js
+++ b/packages/gatsby/src/commands/develop.js
@@ -158,7 +158,11 @@ async function startServer(program) {
       req
         .pipe(
           request(proxiedUrl).on(`error`, err => {
-            report.error(err)
+            const message = `Error when trying to proxy request "${
+              req.originalUrl
+            }" to "${proxiedUrl}"`
+
+            report.error(message, err)
             res.status(500).end()
           })
         )
--- a/packages/gatsby/src/commands/develop.js
+++ b/packages/gatsby/src/commands/develop.js
@@ -155,7 +155,7 @@ async function startServer(program) {
     const { prefix, url } = proxy
     app.use(`${prefix}/*`, (req, res) => {
       const proxiedUrl = url + req.originalUrl
-      req.pipe(request(proxiedUrl)).pipe(res)
+      req.pipe(request(proxiedUrl).on(`error`, err => report(err))).pipe(res)
     })
   }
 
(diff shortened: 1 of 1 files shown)


packages/gatsby/src/commands/develop.js
```diff
@@ -155,7 +155,7 @@ async function startServer(program) {
     const { prefix, url } = proxy
     app.use(`${prefix}/*`, (req, res) => {
       const proxiedUrl = url + req.originalUrl
-      req.pipe(request(proxiedUrl)).pipe(res)
+      req.pipe(request(proxiedUrl).on(`error`, err => report(err))).pipe(res)
```
pieh: shouldn't this be `report.error(err)`?

pieh:
Also example repo would be helpful here so I could validate this change locally

packages/gatsby/src/commands/develop.js
ValeraS: yes it should, sorry

ValeraS:
With repo, it is harder, I've tested it on this repo https://github.com/freeCodeCamp/freeCodeCamp

ValeraS:
repo for validation https://github.com/ValeraS/gatsby-development-proxy

pieh:
Ok, tested and works as expected - one thing I would add is a little more context in error handler so it's easier for users to figure out what is happening, ie:
```js
const message = `Error when trying to proxy request "${
  req.originalUrl
}" to "${proxiedUrl}"`

report.error(message, err)
res.status(500).end()
```
screen shot 2018-11-09 at 20 35 02

ValeraS:
Totally agree, added

PR #9842, opened 2018-11-09
[gatsby-plugin-google-gtag] Disable default pageview tracking

valse:
The default behavior of the `config` snippet is to send a pageview automatically; this is in contrast with the `exclude` option and the plugin tracks, on page load, even the excluded pages.

This PR solve also the double pageview hits on page load too: the first is made by the default behavior and the second one by the `onRouteUpdate` event.

### DIFF
--- a/packages/gatsby-plugin-google-gtag/src/gatsby-ssr.js
+++ b/packages/gatsby-plugin-google-gtag/src/gatsby-ssr.js
@@ -6,6 +6,8 @@ exports.onRenderBody = (
   pluginOptions
 ) => {
   if (process.env.NODE_ENV !== `production`) return null
+  
+  pluginOptions.gtagConfig["send_page_view"] = false
 
   const firstTrackingId =
     pluginOptions.trackingIds && pluginOptions.trackingIds.length
@@ -49,9 +51,7 @@ exports.onRenderBody = (
         ${pluginOptions.trackingIds
           .map(
             trackingId =>
-              `gtag('config', '${trackingId}', ${JSON.stringify(
-                pluginOptions.gtagConfig || {}
-              )});`
+              `gtag('config', '${trackingId}', ${JSON.stringify(pluginOptions.gtagConfig)});`
           )
           .join(``)}
       }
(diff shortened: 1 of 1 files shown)


packages/gatsby-plugin-google-gtag/src/gatsby-ssr.js
```diff
@@ -6,6 +6,8 @@ exports.onRenderBody = (
   pluginOptions
 ) => {
   if (process.env.NODE_ENV !== `production`) return null
+  
+  pluginOptions.gtagConfig["send_page_view"] = false
```
pieh: Can you add comment explaining why this is used? To prevent duplicate `pageview` events being emitted on initial load of page, because we send `pageview` in `onRouteUpdate` in `gatsby-browser.js` which also handles excluded pages config)

pieh: I'm not sure right now if `gtagConfig` is required by plugin - but can there be scenario when this is `undefined` so maybe some sanitization to ensure that `gtagConfig` is an object:
```js
const gtagConfig = pluginOptions.gtagConfig || {}
```
and then use that new object?

pieh:
published `gatsby-plugin-google-gtag@1.0.6`

ardalis:
How do I override this so it actually sends page views?
My realtime analytics has dropped to 0 since migrating my blog to gatsby from wordpress and using this plugin.

I tried the obvious (in gatsby-config.js):
```
        gtagConfig: {
          anonymize_ip: false,
          cookie_expires: 0,
          send_page_view: true
        },
```
but it seems to be ignored. Help?

valse:
Hi you don't need to override it because it's a bugfix: do you add your tracking id on the `trackingIds` array option?

Are you checking it on production? The tracking is disabled on development mode. 

Do you have disabled any ad blocks: sometimes you forget to disable them and you don't understand why the analytics doesn't work 😅

ardalis:
I ended up using the google tag manager and got it working just fine.

https://www.youtube.com/watch?v=__PctBcNZ-c
This video helped.

Good point about ad blockers, though. :)

PR #9846, opened 2018-11-09
[gatsby-plugin-google-gtag] Add resolveEnv option

valse:
Add new `resolveEnv` option function to change the default enviroment variable resolution taken from `process.env.NODE_ENV`.
This could be useful for Netlify branch or preview deploy for example.


valse:
This plugin must work only in production environment state so the @pieh suggestion is the best solution.

PR #9848, opened 2018-11-09
Add basicauth support to gatsby-source-drupal

gatis:
I have a business case where we need to build gatsby page from drupal webiste that is protected by basicauth (Basic access authentication). Since axios already supports it I slightly modified the gatsby-source-drupal code to enable it. Please see if you can implement this feature as proposed in this PR or similar manner.


pieh:
I think this make sense - only thing is that
```
auth: {
  username: 'thisismyvery',
  password: 'simplepassword'
}
```
might seem like these are credentials to drupal admin - might be worth to use more descriptive fields like `htaccess_username` or rename `auth` to `basicAuth`? Another request would be to not put plain string in your example of usage - please replace them with `process.env.BASIC_AUTH_USERNAME` or something like this (we don't want to "encourage" users of using secrets as plain text).

Issue #9858, opened 2018-11-10
All parentheses and nested calc() being ignored in CSS


akselinurmio:
## Description

In CSS nested calc() functions must be treated as simple parentheses.<sup>[MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/calc#Notes)</sup> In Gatsby, however all the parentheses are currently being dropped in the build process (`gatsby build`).

The issue persists even when the `--no-uglify` flag is used on `gatsby build`. Issue doesn't occur on development (`gatsby develop`).

### Steps to reproduce

I wrote a [small Gatsby project here](https://github.com/akselinurmio/gatsby-parentheses-bug) that reproduces the issue. I have [deployed the repo here](https://gatsby-parentheses-bug.netlify.com/).

Here's an outline of the reproducing process:

1. Add `calc()` or calculation in parentheses `()` in some `calc()` function in one of your CSS files.
2. Run `gatsby build`.
3. Inspect the production CSS and see that your parentheses or nested calc() are gone (and your beautiful layout too).

### Expected result

Parentheses in `calc()` functions represent alter the computation order of the calculations.

### Actual result

Parentheses are dropped in Gatsby's production build process.

### Environment

```
  System:
    OS: macOS 10.14.1
    CPU: (4) x64 Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.13.0 - ~/.nvm/versions/node/v10.13.0/bin/node
    npm: 6.4.1 - ~/.nvm/versions/node/v10.13.0/bin/npm
  Browsers:
    Chrome: 70.0.3538.77
    Firefox: 63.0.1
    Safari: 12.0.1
  npmPackages:
    gatsby: ^2.0.44 => 2.0.44 
```

leimonio:
It seems there is an issue with the minification using `optimize-css-assets-webpack-plugin`.

akselinurmio:
No, this is not stale.

marlonmarcello:
Definitely not stale. I just stumbled upon this same issue.
Did you find a work around @akselinurmio ?

gtsop:
@marlonmarcello 

After a longer than usual debugging session, I found this man's legendary post, enjoy:

https://www.eduard.work/adjust-css-minification-in-gatsby/

I'm reposting his solution here in the interest of future-proofing this resource:

```
exports.onCreateWebpackConfig = ({
[...]

PR #9879, opened 2018-11-12
Update css minification configuration

leimonio:
This PR is for fixing issue https://github.com/gatsbyjs/gatsby/issues/9858

Problem
---
`optimize-css-assets-webpack-plugin` is using `cssnano` (and `postcss` to that extend) for optimised css minification. `postcss-calc` used with the default cssnano preset is quite greedy in reducing the `calc()` usage.

Discussion
---
Nowadays `calc()` support is quite good with most of the browsers, with partial support for IE9.
Gatsby support for IE9+ is on the edge for not allowing such `calc` transformations as part of the `optimize-css-assets-webpack-plugin` configuration.

Solution
---
There are 2 possible options for solving this issue:
1. Default disable `calc` optimisations from `postcss-calc`
2. Expose an api for passing options configuring `optimize-css-assets-webpack-plugin` usage.

### DIFF
--- a/packages/gatsby/src/utils/webpack-utils.js
+++ b/packages/gatsby/src/utils/webpack-utils.js
@@ -460,7 +460,18 @@ module.exports = async ({
       ...options,
     })
 
-  plugins.minifyCss = (options = {}) => new OptimizeCssAssetsPlugin(options)
+  plugins.minifyCss = (options = {}) =>
+    new OptimizeCssAssetsPlugin({
+      cssProcessorOptions: {
+        preset: [
+          `default`,
+          {
+            calc: false,
+          },
+        ],
+      },
+      ...options,
+    })
 
   /**
    * Extracts css requires into a single file;


pieh:
I'm not sure how to approach this honestly - can `browserlist` be leveraged here to configure `optimize-css-assets-webpack-plugin`?

m-allanson:
@leimonio thanks for tracking this over to https://github.com/postcss/postcss-calc/issues/64 💯 Once that's fixed, then this should no longer be a problem here?

In the meantime, it sounds like cssnano should automatically pick up a browserslist config. Can you work around this issue by adding a `browserslist` key to your package.json file or a `.browserslistrc` file to the root of your project?

leimonio:
@m-allanson yes exactly, I think this PR is no more relevant so I'm going to close it. Once the above issus on `postcss-calc` has been fixed we can also close the related issue. I'll check the `browserlist`.

leimonio:
@m-allanson @pieh `browserlist` doesn't affect the output of the `calc` transformation. This is because `postccs-calc` won't take into consideration this list on how to transform calculations.

PR #8624, opened 2018-09-28
Pinned dependencies in package.json

lipis:
There was no consistency with pinned/unpinned dependencies and I think it's better to pin then, to have less surprises between runs.

PR #9890, opened 2018-11-12
Pin all versions in package.json

lipis:
To have fewer surprises when installing in in different places.


LekoArts:
Kyle already said no to this?
https://github.com/gatsbyjs/gatsby/pull/8624#issuecomment-425559827

DSchau:
As @LekoArts and @KyleAMathews said, we're not going to merge this unfortunately. If you can provide concrete reasons for the benefit of this--or how it's broken something for you--we'd be more amenable to perhaps merging. As it stands, I don't think this is something we _want_ to merge.

Thanks for the PR-and all the other PRs you've posted recently!

PR #9892, opened 2018-11-12
fix: enable ref forwarding with forwardRef

DSchau:
Note: I think we should minimally make this more than a patch release per
[React docs](https://reactjs.org/docs/forwarding-refs.html#note-for-component-library-maintainers)

This will make the Link component work more like many would expect
(innerRef is confusing!), but it does come at the cost of requiring
React 16.3. I think it's a reasonable trade off, however.

cc @LekoArts I validated this with a sample repo, but want to pull this down and make sure it fixes your issue, as well?


packages/gatsby-link/src/index.js
```diff
@@ -97,7 +97,6 @@ class GatsbyLink extends React.Component {
       /* eslint-disable no-unused-vars */
       activeClassName: $activeClassName,
       activeStyle: $activeStyle,
-      ref: $ref,
```
DSchau: @LekoArts this fixes the bug re: accessing ref.

DSchau:
@pieh I'll check this out locally, but I think this is a valuable change we should get in.

I don't think it is actually a breaking change if we already require >= 16.3 in Gatsby.

DSchau:
Few things:

- This does break React@<16.3.0 support, but since we have a [peerDep of 16.4.1](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/package.json#L161-L164), that's probably OK
- [Here's a reproduction](https://github.com/DSchau/gatsby-9892) showing both `ref` and `innerRef` working; so this isn't a breaking change if anyone was using innerRef, either
    - Note: this will break both page-2 and page-3 links (on purpose) because I'm just preventing the default on click with an event listener from the ref

DSchau:
@KyleAMathews yeah - I'm thinking once some of the newer, useful stuff launches, e.g. [Suspense with SSR](https://reactjs.org/blog/2018/10/23/react-v-16-6.html#reactlazy-code-splitting-with-suspense), Hooks, etc. that could be a good time to bump the peerDep.

In effect--if it makes the code we maintain smaller, cleaner, etc. that's a good time to make that change.

What do you think?

KyleAMathews:
Yeah, the next thing that should really impact us is suspense as we could replace a lot of our preloading/prefetching/page transition code with it.

Hooks are great but I don't think there's anywhere we'd use them in the Gatsby runtime?

wardpeet:
@DSchau are we good to merge this?

PR looks allright to me

Any updates you want to give @LekoArts about issue testing?

LekoArts:
I think when I tested it back then everything worked 🤔

DSchau:
```
Successfully published:
 - gatsby-image@2.0.26
 - gatsby-link@2.0.8
 - gatsby@2.0.91
```

PR #9898, opened 2018-11-13
Create clearer pathways in docs

amberleyromo:
@shannonbux I began making some changes based on our recent interview and resulting conversation.

## minor
- clarified why VSCode is recommended.

### docs vs tutorial
The conversation we had about being unsure where you are (tutorial vs. docs), and not being sure where to start led me down this pathway. It went a bit further than anticipated, but I think it makes a great deal more sense.
- pulled the "getting started" language out of the "docs" landing page, and into a dedicated doc.
- rewrote the "docs" landing page to:
  1) show distinctly where the user is.
  2) make the decision for what's appropriate for them less ambiguous (tutorial, quick start, recipes, etc).
  3) define the different major sections of the docs.

screen shot 2018-11-12 at 6 54 53 pm

Comparison of docs landing page with tutorial landing page:

screen shot 2018-11-12 at 7 03 39 pm

Compare the above with current org:

screen shot 2018-11-12 at 7 04 57 pm

### GraphQL reference

The GraphQL reference doc was previously indexed under API Reference. It makes more sense in the GraphQL section of the guides. I did some reorganizing along those lines:

screen shot 2018-11-12 at 4 21 38 pm

(The addition here is "Understanding GraphQL syntax", which links to the reference guide).

ETA: One point on metrics too -- splitting this Docs landing page into these different pathways will give us a better sense of what people are choosing to do, rather than automatically landing on Quick Start guide for docs.


www/src/data/sidebars/doc-links.yaml
```diff
@@ -1,9 +1,8 @@
-- title: Quick Start
+- title: Documentation
+  link: /docs/
   items:
-    - title: Getting Started
-      link: /docs/
-    - title: Starter Library
-      link: /starters/
```
pieh: Do we have Starter Library linked anywhere else?

amberleyromo: Planning to merge the ecosystem homepage update today -- the .org redesign will take care of linking the starter library, plugin library, etc.

^That piece of work doesn't account for changing the nav yet though. So perhaps should leave this here, until it's finished.

Another option would be to have another category section in docs "Ecosystem", that mirrors the new Ecosystem landing page -- link out to plugin library, starter library, etc etc.

screen shot 2018-11-13 at 9 45 48 am

Actually this would be great too, because we could reference what they are, top-level, on the new docs landing page.

@shannonbux what do you think?

docs/tutorial/index.md
```diff
@@ -4,7 +4,7 @@ title: Gatsby.js Tutorial
 
 Welcome to Gatsby! We’re glad you’re here. The goal of this tutorial is to guide you through setting up and deploying your first Gatsby site using a starter template. As we walk through that process, we’ll introduce some more general web development topics, and go over the underlying structure of a Gatsby site.
 
-The full tutorial is intended to be as accessible as possible to people without much web development experience (yet!) — no need to be an expert. If you prefer to jump straight to code, feel free to skip the step-by-step tutorial and see the [quick start](/docs/) page.
+> The full tutorial is intended to be as accessible as possible to people without much web development experience (yet!) — no need to be an expert. If you prefer to jump straight to code, feel free to skip the step-by-step tutorial and see the [quick start](/docs/) page.
```
DSchau: 👍

We should consider styling blockquotes (with a certain class? all?) like [reactjs.org does them](https://reactjs.org/tutorial/tutorial.html). I quite like them!

screen shot 2018-11-13 at 10 12 02 am

www/src/pages/docs/index.js
```diff
@@ -18,95 +18,45 @@ class IndexRoute extends React.Component {
...
-              <li>
-                <code>gatsby build</code> — Gatsby will perform an optimized
-                production build for your site generating static HTML and
-                per-route JavaScript code bundles.
-              </li>
-              <li>
-                <code>gatsby serve</code> — Gatsby starts a local HTML server
-                for testing your built site.
-              </li>
+                <Link to="/tutorial/">Tutorial</Link>: The tutorial is written to be as accessible as possible to people without much web development experience.</li>
```
DSchau: These are all quite a bit nicer. Nice work!

Issue #4977, opened 2018-04-14
GraphQL query - sort nulls last


larryhudson:
Hi everyone, I'm so impressed with Gatsby! I'm just having an issue where I'm sorting Contentful items by date (descending), but date isn't a required field. Items without a date (returning a null value) are at the top of the list. Is there an option for having 'nulls last' when sorting?

Here's the code for my query:

```
query categoryItemsQuery($slug: String!) {
  allContentfulPortfolioItem(
    filter: { categories: { slug: { eq: $slug } } },
    sort: {fields: [date], order: DESC}) {
    edges {
      node {
        id
      }
    }
  }
}
```

Sorry if this is the wrong place to post this! Thanks for all your hard work.

pieh:
You have currently 2 options:

- make 2 separate queries (1 for sorted items with date and second for items without date) and concatenate them in your page component:
```gql
query categoryItemsQuery($slug: String!) {
  withDate:allContentfulPortfolioItem(
    filter: { categories: { slug: { eq: $slug } }, date: { ne: null } },
    sort: {fields: [date], order: DESC}) {
    edges {
      node {
        id
      }
    }
  }
  withoutDate:allContentfulPortfolioItem(
    filter: { categories: { slug: { eq: $slug } }, date: { eq: null } }) {
    edges {
      node {
        id
      }
    }
  }
}
```
and then in your component use `this.props.data.withDate` and `this.props.data.withoutDate`

- or add `sortDate` field to your Portfolio Item node in `gatsby-node.js`:
```
exports.onCreateNode = ({ node, boundActionCreators }) => {
  const { createNodeField } = boundActionCreators
  if (node.internal.type === `NameOfType`) { // name of type you can get in graphiql in docs section or by ctrl+clicking on `node` field in your query
    createNodeField({
      node,
      name: `sortDate`,
      value: node.date || "1970-01-01", // either date if node has this field or very early date that should be before any of the date in content
    })
  }
};
```
and then sort by `fields__sortDate`

All the examples are written without testing, so this might not work out of the box - but hope approaches are readable

KyleAMathews:
It seems reasonable too to change our sorting so nodes without the sort field always are at the end.

KyleAMathews:
@larryhudson you could add this at https://github.com/gatsbyjs/gatsby/blob/e4bbeb3b57c27a29217ca43e29c5dd520d034771/packages/gatsby/src/schema/run-sift.js#L166-L173

[...]

PR #9919, opened 2018-11-14
feat(Gatsby): add lokijs nodes db implementation

Moocar:
This PR adds a feature flag `GATSBY_DB_NODES` that can be used to change the storage engine for the gatsby data layer (`nodes`). 

- `redux` (default) which uses the existing redux implementation, and `sift` for querying. Or, you can use 
- `loki` which uses the loki in-memory database to store and query.

This PR re-implements functionality in #9338, but with all tests passing and addressing previous feedback. It should also be easier to review since it builds on several refactorings. 

Some things to note:

1. I [submitted a PR to lokijs](https://github.com/techfort/LokiJS/pull/718) which still hasn't been merged, though the author says he'll start working on it soon. Therefore, in in interim, I've published [@moocar/lokijs](https://www.npmjs.com/package/@moocar/lokijs). 
1. I haven't implemented auto indexing of query fields yet. I'll attack that next.
1. I suggest we ask the community to try out the feature flag once it's merged to get feedback. All the tests pass, but this is such a big change that we'll want to test it gradually
1. While loki uses the same mongo query language as sift, they do have different behavior. Most of my time on this PR was spent ensuring that loki behaves **exactly** like sift. See [db/loki/nodes-query.js](https://github.com/gatsbyjs/gatsby/blob/cddbe893a4ce638babb1cbe5e5da4c13b6f5e57d/packages/gatsby/src/db/loki/nodes-query.js). But there's a chance a few edge cases have slipped through the cracks.
1. the feature flag works with the tests too `GATSBY_DB_NODES=loki yarn test`. We should perhaps look into running this on all PRs


Moocar:
Looks like there are some CI test errors. Closing while I fix them

Moocar:
Fixed

packages/gatsby/src/db/loki/index.js
```diff
@@ -0,0 +1,120 @@
...
+      autoload: true,
+      autoloadCallback: err => {
+        if (err) {
+          reject(err)
+        } else {
+          resolve()
+        }
+      },
+      autosave: true,
+      autosaveInterval: 1000,
```
pieh: This might be potentially problematic with saving part of state in current way and part of it autosaving (potentially leading to inconsistent state). I don't think this should be a blocker as this will be behind experimental env var (just something worth pointing out).

packages/gatsby/src/db/loki/index.js
```diff
@@ -0,0 +1,120 @@
...
+    }
+    db = new loki(saveFile, dbOptions)
```
pieh: Did you experiment with using different persistent adapters? For larger sites it might be beneficial to use `LokiFsStructuredAdapter`

packages/gatsby/src/db/loki/index.js
```diff
@@ -0,0 +1,120 @@
...
+  })
+}
+
+async function startInMemory() {
+  // Use uuid purely for a random name
+  db = new loki(uuidv4())
+}
```
pieh: this is just for unit tests?

packages/gatsby/src/db/loki/nodes.js
```diff
@@ -1,12 +1,354 @@
...
 
+/////////////////////////////////////////////////////////////////////
+// Reducer
+/////////////////////////////////////////////////////////////////////
+
+function reducer(state = new Map(), action) {
+  switch (action.type) {
+    case `DELETE_CACHE`:
+      deleteAll()
+      return new Map()
```
pieh: does it make sense to create `new Map()`? we should be able to set state just to `null` and return same state (or `null`) again so garbage collector doesn't need to additional work that we won't be using?

packages/gatsby/src/db/loki/index.js
Moocar: Yeah, normally I would add this in a `__tests__` file, but it's so small and easier to use in a production namespace. I can move it out of prod code if you want though.

packages/gatsby/src/db/loki/index.js
Moocar: I did actually! `LokiFsStructuredAdapter` was the first I tried. Unfortunately I ran into a few situations in testing where some of the .db files were missing so loki wouldn't start up. Might be one to try again another day. In the meantime, the default file adapter seems just as fast as the redux implementation

packages/gatsby/src/db/loki/index.js
Moocar: Good call. I know that we can trigger loki saves manually, so it might be possible to hook it up to the redux save process so they're synchronized. I'll add a TODO

packages/gatsby/src/db/loki/nodes.js
Moocar: 👍 fixed.

Moocar:
@pieh Thanks for the feedback. Ready for another review.

packages/gatsby/src/db/loki/nodes-query.js
```diff
@@ -231,6 +242,24 @@ function toSortFields(sortArgs) {
   ])
 }
 
+// Every time we run a query, we increment a counter for each of its
+// fields, so that we can determine which fields are used the
+// most. Any time a field is seen more than `FIELD_INDEX_THRESHOLD`
+// times, we create a loki index so that future queries with that
+// field will execute faster.
```
pieh: I think this is fine. Something to think about in the future is that we can analyze queries (through query AST) when we are extracting / compiling them and then we can know what fields are used for filtering. We will know how many times we will run queries (number of pages using same template/query). Not sure if it's worth extra engineering to do that tho.

packages/gatsby/src/db/__tests__/nodes.js
```diff
@@ -0,0 +1,347 @@
+const { actions } = require(`../../redux/actions`)
+const { getNode, getNodes } = require(`../nodes`)
+const { store } = require(`../../redux`)
+require(`./fixtures/ensure-loki`)()
+
+describe(`stuff`, () => {
```
pieh: `stuff` :D 

What's the reason for splitting part of `redux/__tests__/nodes.js` here? I don't quite understand that part

packages/gatsby/src/schema/lazy-fields.js
```diff
@@ -0,0 +1,42 @@
...
+      const field = gqlType.getFields()[fieldName]
+      const { elemMatch } = value
+      if (field.type instanceof GraphQLList && elemMatch) {
+        return _.some(elemMatch, (value, fieldName) => {
+          // This is disgusting, but works
+          const innerType = field.type.ofType.getFields()[fieldName].type.ofType
+          return contains({ [fieldName]: value }, innerType)
+        })
+      }
+      return false
```
pieh: doesn't `GraphQLObject` need to be handled similar way as List?

if we have node:
```
{
  id: `foo`,
  [...parent/children/internals]
  thisWillBeNested: {
    nestedFieldThatWillHaveResolver___NODE: `other_node_id`
  }
}
```
will this work correctly?

packages/gatsby/src/db/loki/nodes.js
```diff
@@ -1,12 +1,354 @@
...
+    invariant(coll, `${type} collection doesn't exist. When trying to update`)
+  }
+
+  if (oldNode) {
+    const lokiKeys = new Set([`$loki`, `meta`, `id`])
+    _.forEach(oldNode, (v, k) => {
+      if (!lokiKeys.has(k)) {
+        delete oldNode[k]
+      }
+    })
```
pieh: you can do:
```js
oldNode = _.omit(oldNode, [`$loki`, `meta`, `id`])
```

packages/gatsby/src/schema/infer-graphql-input-fields.js
```diff
@@ -210,6 +210,8 @@ function inferGraphQLInputFields({
 const EXCLUDE_KEYS = {
   parent: 1,
   children: 1,
+  $loki: 1,
+  meta: 1,
```
pieh: I realize that `meta` is loki generated field, but adds some limitation we previously didn't have - is it possible to swap to different fieldname? something more "obscure" like `$meta` that is less likely to be used by nodes?

packages/gatsby/src/db/loki/nodes-query.js
Moocar: Yeah, I thought about this approach too but didn't want to get lost in the weeds of query ASTs. I definitely think it's the right approach in the future. Thankfully the CPU overhead of the above is very minimal.

packages/gatsby/src/db/__tests__/nodes.js
Moocar: Haha, can you tell I spent many hours in testing land? :P. I'll fix this up.

As for splitting `redux/__tests__/nodes.js`, half those tests were testing the `actions.js` functions as pure functions. I.e checking that the action and reducer result in specific states. The tests I moved into `db/__tests__/nodes.js` are testing all the way down to the database layer, regardless of the db implementation. This allows us to reuse as many tests as possible for both redux and loki.

packages/gatsby/src/db/loki/index.js
```diff
@@ -0,0 +1,120 @@
...
+    name: `gatsby:nodeMeta`,
+    options: {
+      unique: [`id`],
+      indices: [`id`],
+    },
+  },
+  // The list of all node type collections. Each object has keys
+  // `type` and `collName` so you can quickly look up the collection
+  // name for a node type.
+  // e.g { type: `myType`, collName: `gatsby:nodeType:myType` }
```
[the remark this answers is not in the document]
Moocar: Technically, we could get away with something like this:

```
const coll = db.getCollection(nodesDb.makeTypeCollName(typeName))
```

But this ties the name of the node type collection to the `makeTypeCollName` function. What if that function produces a different result in future, but the original collection still exists in the db? There also might be a small performance cost since we would need to calculate the string name for each lookup, vs a fast in-memory db lookup.

packages/gatsby/src/schema/infer-graphql-input-fields.js
Moocar: Great catch. I'll see what the options are in loki

packages/gatsby/src/schema/infer-graphql-type.js
```diff
@@ -270,6 +277,8 @@ function inferFromFieldName(value, selector, types): GraphQLFieldConfig<*, *> {
       }
     } else {
       type = fields[0].nodeObjectType
+      lazyFields.add(type.name, fieldName)
+      lazyFields.add(typeName, fieldName)
```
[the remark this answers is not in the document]
Moocar: I blame TDD. Fixed now along with lazy-fields.js in general.

packages/gatsby/src/schema/lazy-fields.js
Moocar: Another good catch. Lazy-fields looks cleaner now.

packages/gatsby/src/db/loki/nodes.js
Moocar: N/A anymore (see below comment)

packages/gatsby/src/db/loki/nodes.js
```diff
@@ -1,12 +1,354 @@
...
+    Object.assign(oldNode, node)
```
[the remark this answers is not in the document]
Moocar: So, I went down this road because loki doesn't have the concept of `upsert`, i.e if you try and create a node which already has a `$loki` field, or update a field that doesn't, loki will error. Whereas in the redux reducer, we simply overwrite the value for the `node.id` every time. But I just realized I can get around this in `actions.createNode()` by deleting the node itself as well as all its children before we re-insert. See [latest commit](https://github.com/gatsbyjs/gatsby/pull/9919/commits/711af45ede9e997a7b6c7d532c6b404014318401#diff-f61af2b51837961d211c7ea1a4b91837R599). Can you foresee any problems with this approach?

packages/gatsby/src/db/loki/nodes.js
```diff
@@ -1,12 +1,354 @@
...
+  // collName mapping, but there won't be any nodes in the type
+  // collection. Need to create a transaction around this
+  return nodeTypeColl.insert(node)
+}
+
+/**
+ * Updates a node in the DB. This works by removing all fields db
+ * fields ($loki, meta, and id) and then using `Object.assign` to copy
+ * all fields in `node` over `oldNode`. Therefore node should contain
+ * ALL fields and not just changes.
```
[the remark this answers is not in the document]
Moocar: see above

Moocar:
@pieh feedback addressed. On my TODO list as of now:

- [ ] See if we can tell `loki` to rename the `meta` field
- [ ] See if loki can handle null sorting so it's in parity with sift
- [ ] See if we can sync redux and loki saves to avoid inconsistent persistent state
- [ ] Creating field indexes at query compile time by analyzing query AST

pieh:
> * Creating field indexes at query compile time by analyzing query AST

Don't spend too much time on it, there would need to be a lot more logic there to do (and that time can be better spend on more important things):
 - getting root field from AST
 - connecting root field name to actual node type - so extra lookup would be needed when creating schema to point `markdownRemark` and `allMarkdownRemark` to `MarkdownRemark` node type
 - converting graphql args AST to something we could consume easier
 - don't remember if fragments are resolved or not - but potentially would need to resolve/expand fragments

packages/gatsby/src/redux/actions.js
```diff
@@ -596,6 +596,9 @@ actions.createNode = (
           actions.deleteNode({ node: getNode(n) })
         )
       }
+      if (deleteAction) {
+        deleteAction.push(actions.deleteNode({ node: oldNode }))
+      }
```
pieh: can we pass `oldNode` to `CREATE_NODE` action like you had in previous version and move deleting node to `loki` specific code? It feels that we shouldn't cover lack of upsert functionality here

Moocar:
@pieh Ready for another review. The latest changes are:

1. Fixed a [bug in loki](https://github.com/techfort/LokiJS/pull/722) (that I introduced) where nested field indexes weren't being sorted correctly. I've bumped `@moocar/lokijs` to reference the new code until it gets merged and released.
1. Loki queries now use the same rank system as sift when sorting `undefined` and `null` values. I.e `undefined` > `null` > everything else. I think we should consider changing this in `v3` as per #4977. 
1. Gatsby only allows the sort order for the first sort field to be specified. So secondary sorting fields use are all ascending. I've changed our loki code to mirror this. I suggest that in `v3`, we change the sorting interface to allow specification of orders on a per sort field basis.
1. Further refined the loki implementation for `updateNode`. Now, the `oldNode` is passed in the action and loki will then delete the node and recreate it. This leaves the redux code exactly the same
1. Removed `meta` field from all loki nodes. It turns out this is only used for `ttl` and `changes` functionality, which we're not using. If we do decide to use those, we can propose a change to loki to be able to change the field name.

When run on `www`, the two implementations produce identical query results now, except for the `graphviz` weirdness. To test it, make sure you're using the changes in #10085. 

I'm getting pretty sick of loki coding, so I think I'll address the redux/loki autosave syncing till another day.

pieh:
> I'm getting pretty sick of loki coding, so I think I'll address the redux/loki autosave syncing another day.

👍 

Will take another look today! Thanks for all that work!

pieh:
I hope you're not sick of loki to take a look at proposed changes against your branch in https://github.com/Moocar/gatsby/pull/1 - that's the last thing I would like before merging this in (finally!)

Moocar:
Thanks for that PR @pieh. Definitely better.

Moocar:
@pieh ok, I couldn't help myself. I added synchronized db saves. It's in a pull request against my original branch so you can see what's going on and provide feedback if you like: https://github.com/Moocar/gatsby/pull/2

Moocar:
I just compared memory usage between loki and sift implementations on `www`. The below results are taken immediately after Bootstrap has finished. I only ran each a few times, but they were consistent. As expected, Loki gives us slightly lower memory usage since we don't need to cache as much.

### Redux/sift

|  (index)  |  Values  |
|-----------|----------|
|    rss    | '677.82' |
| heapTotal | '471.63' |
| heapUsed  | '259.07' |
| external  | '87.18'  |

### Loki

|  (index)  |  Values  |
|-----------|----------|
|    rss    | '612.18' |
| heapTotal | '424.78' |
| heapUsed  | '266.92' |
| external  | '84.51'  |

PR #9923, opened 2018-11-14
fix(gatsby-plugin-offline): fix certain resources being cached excessively

### LINKED ISSUE #9415: 1668MB cache storage in chrome
image

image

image

image

image

Even if I delete all these items the cache doesn't go down.

image

I went to "\AppData\Local\Google\Chrome\User Data\***\Cache" and checked the whole folder doesn't increase on refresh and is not even the size of the storage chrome show. It might be a chrome bug?

config:

    module.exports = {
      siteMetadata: {
        title: '***',
        siteUrl: `https://www.***.com/`,
      },
      plugins: [
        `gatsby-transformer-sharp`,
        `gatsby-plugin-sharp`,
        {
          resolve: `gatsby-plugin-sitemap`,
          options: {
            exclude: [
              '/internal/*',
            ]
          }
        },
        `gatsby-plugin-catch-links`,
        `gatsby-plugin-emotion`,
        `gatsby-plugin-typescript`,
        'gatsby-plugin-react-helmet',
        {
          options: {
            name: 'gatsby-starter-default',
            short_name: 'starter',
            start_url: '/',
            background_color: '#000000',
            theme_color: '#000000',
            display: 'minimal-ui',
            icon: 'src/content/logo/square.png', // This path is relative to the root of the site.
          },
        },
        'gatsby-plugin-offline',
        {
          resolve: `gatsby-source-filesystem`,
          options: {
            path: `${__dirname}/src/content/pages`,
            name: 'markdown-pages',
          },
        },
        {
          resolve: `gatsby-source-filesystem`,
          options: {
            path: `${__dirname}/src/content/images`,
            name: 'images',
          },
        },
        {
          options: {
            plugins: [
              {
                resolve: `gatsby-remark-images`,
                options: {
                  // It's important to specify the maxWidth (in pixels) of
                  // the content container as this plugin uses this as the
                  // base for generating different widths of each image.
                  maxWidth: 1200,
                  linkImagesToOriginal: false,
                  backgroundColor: 'transparent',
                  withWebp: true,
                },
              },
            ],
          },
        },
        {
          resolve: `gatsby-plugin-google-analytics`,
          options: {
            trackingId: "****",
            // Puts tracking script in the head instead of the body
            head: false,
            // Setting this parameter is optional
            anonymize: false,
            // Setting this parameter is also optional
            respectDNT: false ,
            // Avoids sending pageview hits from custom paths
            exclude: [],
            // Any additional create only fields (optional)
            sampleRate: 100,
            siteSpeedSampleRate: 100,
            cookieDomain: "www.***.com",
          },
        },
      ],
    }

## Description

1668MB cache storage

Looks like every refresh adds 10MB. Doesn't look like is based on some kind of time (the more you wait the more it adds). Just on every refresh I get more 10-20MB of cache.

### Steps to reproduce

I don't know

### Expected result

certainly less than 1.6GB...

### Actual result

1668MB cache storage

### Environment

  System:
    OS: Windows 10
    CPU: x64 Intel(R) Core(TM) i7-5600U CPU @ 2.60GHz
  Binaries:
    Yarn: 1.9.4 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 5.5.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 42.17134.1.0
  npmPackages:
    gatsby: ^2.0.19 => 2.0.21
    gatsby-image: ^2.0.17 => 2.0.17
    gatsby-plugin-catch-links: ^2.0.4 => 2.0.4
    gatsby-plugin-emotion: ^2.0.5 => 2.0.5
    gatsby-plugin-google-analytics: ^2.0.6 => 2.0.6
    gatsby-plugin-manifest: ^2.0.6 => 2.0.6
    gatsby-plugin-offline: ^2.0.5 => 2.0.6
    gatsby-plugin-sharp: ^2.0.8 => 2.0.8
    gatsby-plugin-sitemap: ^2.0.1 => 2.0.1
    gatsby-plugin-typescript: ^2.0.0 => 2.0.0
    gatsby-plugin-typography: ^2.2.0 => 2.2.0
    gatsby-remark-images: ^2.0.4 => 2.0.4
    gatsby-source-filesystem: ^2.0.3 => 2.0.3
    gatsby-transformer-remark: ^2.1.7 => 2.1.7
    gatsby-transformer-sharp: ^2.1.5 => 2.1.5

davidbailey00:
Fixes #9415

This leverages Workbox's automatic caching by creating a `<link>` prefetch element for each resource. This means we don't have to worry about whether the resource can be fetched with CORS or not, since the browser handles it auto. It also changes the runtime caching RegExps so that only paths with certain extensions are cached, but 3rd party resources with these extensions can now be cached.


packages/gatsby-plugin-offline/src/sw-append.js
```diff
@@ -1,29 +1 @@
...
-            }
-
-            return request.then(response => cache.put(resource, response))
-          })
-        )
-      )
-    )
-  }
-})
+// noop
```
DSchau: If we're no longer using this, do we even need this file at all?

Additionally, we probably can remove the appending stuff, as well.

https://github.com/gatsbyjs/gatsby/blob/e02f9d54c05721fd96a8597954fc844d31386b8e/packages/gatsby-plugin-offline/src/gatsby-node.js#L125

packages/gatsby-plugin-offline/src/gatsby-node.js
```diff
@@ -96,13 +96,13 @@ exports.onPostBuild = (args, pluginOptions) => {
...
-        urlPattern: /\.(?:png|jpg|jpeg|webp|svg|gif|tiff|js|woff|woff2|json|css)$/,
+        urlPattern: /^https?:.*\.(?:png|jpg|jpeg|webp|svg|gif|tiff|js|woff|woff2|json|css)$/,
         handler: `staleWhileRevalidate`,
       },
       {
-        // Use the Network First handler for external resources
-        urlPattern: /^https?:/,
-        handler: `networkFirst`,
+        // Google Fonts CSS (doesn't end in .css so we need to specify it)
+        urlPattern: /^https?:\/\/fonts\.googleapis\.com\/css/,
```
DSchau: Maybe just use a regex like in this recipe? https://gist.github.com/addyosmani/0e1cfeeccad94edc2f0985a15adefe54#cache-all-google-fonts-requests-up-to-a-maximum-of-30-cache-entries

Also from what I understand, staleWhileRevalidate will request both cache/network each time, and then update the cache on success.

It seems reasonable that something like google fonts can hit the cache first, then the network if that fails? What do you think?

packages/gatsby-plugin-offline/src/gatsby-browser.js
```diff
@@ -40,8 +39,14 @@ exports.onServiceWorkerActive = ({
     )
   )
 
-  serviceWorker.active.postMessage({
-    api: `gatsby-runtime-cache`,
-    resources: [...resources, ...prefetchedResources],
+  const resources = [...headerResources, ...prefetchedResources]
+  resources.forEach(resource => {
+    // Create a prefetch link for each resource, so Workbox runtime-caches them
+    const link = document.createElement(`link`)
```
DSchau: Isn't it possible that we're duplicating a prefetch here? i.e. `head > link[href]` could match

```
<link rel="prefetch" href="/static/d/368/path---showcase-c-41-46f-ykWCDEvUZCguGaUYZsNI5k1Io.json">
```

and we'd create another prefetch link. Is that OK?

packages/gatsby-plugin-offline/src/gatsby-browser.js
```diff
@@ -40,8 +39,14 @@ exports.onServiceWorkerActive = ({
...
+    link.rel = `prefetch`
+    link.href = resource
+
+    document.head.appendChild(link)
```
DSchau: Should the order of these (append vs. onload) be swapped? I'd think we'd want to register the onload/remove behavior _before_ adding it. If nothing else, seems to be more intuitive that way, even if both work.

packages/gatsby-plugin-offline/src/gatsby-node.js
davidbailey00: `staleWhileRevalidate` in the Workbox config is equivalent to the `CacheFirst` strategy when calling Workbox APIs directly, so the current config actually does what you've described (except for the max 30 entries). `woff2` and other font files from Google (or any domain) are covered by the first regex, which matches common extensions (and now works for external resources due to the added `^https?:.*` at the start).

Also, since the linked code calls Workbox APIs directly, we'd have to move it to `sw-append.js`, which means it can't be customised by the user - so I think it's best to leave it user configurable. Basically Workbox generates similar code by reading the config - `sw-append.js` is for extra stuff which we can't do with the config.

packages/gatsby-plugin-offline/src/gatsby-browser.js
davidbailey00: Yes, this is expected behaviour - we need to "prefetch" after the SW has activated, so that Workbox can cache the responses.

packages/gatsby-plugin-offline/src/gatsby-browser.js
davidbailey00: These can be reordered if you like - although it should work fine as-is, since the browser doesn't start fetching until it's appended. Would you like me to swap the order?

DSchau: Sure, just for clarity! More of a nit than substantive, but I think it's a bit clearer.

PR #9924, opened 2018-11-14
Documentation fix: Execute trailing slash stripping function, rather than assign it.

stevetweeddale:
The trailing slash stripping plugin [executes the function](https://github.com/gatsbyjs/gatsby/blob/f9714a5646523bc728586e3692190507ad68e6c1/packages/gatsby-plugin-remove-trailing-slashes/src/gatsby-node.js#L9) rather than just assigning it, so I'd imagine that's what the example code should be doing too.

### DIFF
--- a/docs/docs/creating-and-modifying-pages.md
+++ b/docs/docs/creating-and-modifying-pages.md
@@ -114,6 +114,8 @@ _Note: There's also a plugin that will remove all trailing slashes from pages au
 [gatsby-plugin-remove-trailing-slashes](/packages/gatsby-plugin-remove-trailing-slashes/)_.
 
 ```javascript:title=gatsby-node.js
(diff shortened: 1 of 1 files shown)


docs/docs/creating-and-modifying-pages.md
```diff
@@ -121,7 +121,7 @@ exports.onCreatePage = ({ page, actions }) => {
   return new Promise(resolve => {
     const oldPage = Object.assign({}, page)
     // Remove trailing slash unless page is /
-    page.path = _path => (_path === `/` ? _path : _path.replace(/\/$/, ``))
```
DSchau: If we're going to go this route, why not create a function (e.g. `const replacePath = _path => // yadda yadda`) and then use it like the link you provide? That seems _much_ clearer!

What do you think?

PR #9926, opened 2018-11-14
readme: add total npm downloads badge

aadibajpai:
The total figure is much higher (6 million) than the monthly rate and quite a lot more impactful. So, it should make for a better listing on the readme :)


KyleAMathews:
The total number is nice! :-)

But current velocity is more interesting to us as that's a stronger indication of current usage — which is what people really care about. Also monthly downloads is a common thing to share so it's easier to compare us to other projects this way.

aadibajpai:
What do you think about both, then? That whopping 6 million figure packs a punch! @KyleAMathews

KyleAMathews:
sure :-)

aadibajpai:
Should be okay now :)

Issue #9880, opened 2018-11-12
Add an `auto-config` section to the debugging doc


m-allanson:
## Summary

Add a new section to the [debugging the build process doc](https://www.gatsbyjs.org/docs/debugging-the-build-process/).

Maybe also restructure the document so the sections go from 'simplest' to 'most complex', like this:

- VS Code debugger (automatic config) _<-- this is the new section_
- VS Code debugger (manual config)
- Chrome DevTools for Node

### Basic example

If you use VS Code and its integrated terminal, you can configure it to automatically create the debug config for you:

- Enable the VS Code setting `Debug > Node: Auto Attach`
- Using VS Code's integrated terminal run `node --inspect node_modules/.bin/gatsby develop` instead of `gatsby develop`
- debug!

These steps are based on the following article: https://itnext.io/the-absolute-easiest-way-to-debug-node-js-with-vscode-2e02ef5b1bad

### Motivation

This reduces the friction of setting up a debug environment, which can make it easier to track down build process errors.

axhon:
I've taken a stab at this. I will make a PR momentarily.

axhon:
The fix to this has been merge #9937

PR #9937, opened 2018-11-15
Adds auto-config option in  debugging process docs

axhon:
Addresses #9880


pieh:
Hey, thanks for tackling #9880!

Few things:
 - need to adjust this small list of content at the beginning (re-order, add auto-config item and adjust link to vs code section):
   ```
   In this guide you will learn how to debug some code using:

   - [Chrome DevTools for Node](#chrome-devtools-for-node)
   - [VS Code debugger](#vs-code-debugger)
   ```
 - There need to be some edits to "Finishing thoughts on DevTools" section - it was prepared with having chrome dev tools as first section in mind - especially this:
   > You can successfully debug your code using Chrome DevTools but using it isn't really that convenient. There are a lot of steps you need to do manually every time you want to use debugger, so in the next section you'll learn how to use the built-in debugging capabilities of VS Code.
   >
   > "Why did we go through all those steps only to find out that there are better options?" you might ask. That's a great question and here are couple of reasons:

axhon:
I've added the edits

PR #9946, opened 2018-11-15
chore(gatsby-remark-katex): move katex to peer dep and update version

hexrcs:
[Katex 0.10.0](https://github.com/KaTeX/KaTeX/releases/tag/v0.10.0) introduced tons of new features and fixes. I also updated the testing snapshots. :)


packages/gatsby-remark-katex/package.json
```diff
@@ -8,7 +8,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.0.0",
-    "katex": "^0.9.0",
+    "katex": "^0.10.0",
```
DSchau: Let's make this a peerDependency like we do in other plugins, e.g. `gatsby-remark-prismjs`, `gatsby-plugin-emotion`, etc.

That way users can install whatever version they'd like, rather than mandating changes like these. I'd recommend something like

```json
"peerDependencies": {
  "katex": ">=0.10.0 <1.0.0"
}
```

(<1.0.0 versions are weird with semver, e.g. `^0.10.0` does not satisfy `0.11.0`)

packages/gatsby-remark-katex/src/__tests__/__snapshots__/index.js.snap
```diff
@@ -32,7 +32,7 @@ Object {
...
...thdefault\\">b</span><span class=\\"msupsub\\"><span class=\\"vlist-t\\"><span class=\\"vlist-r\\"><span class=\\"vlist\\" style=\\"height:0.8141079999999999em;\\"><span style=\\"top:-3.063em;margin-right:0.05em;\\"><span class=\\"pstrut\\" style=\\"height:2.7em;\\"></span><span class=\\"sizing reset-size6 size3 mtight\\"><span class=\\"mord mtight\\">2</span></span></span></span></span></span></span></span><span class=\\"mspace\\" style=\\"margin-right:0.2777777777777778em;\\"></span><span class=\\"mrel\\">=</span><span class=\\"mspace\\" style=\\"margin-right:0.2777777777777778em;\\"></span></span><span class=\\"base\\"><span class=\\"strut\\" style=\\"height:0.8141079999999999em;vertical-align:0em;\\"></span><span class=\\"mord\\"><span class=\\"mord mathdefault\\">c</span><span class=\\"msupsub\\"><span class=\\"vlist-t\\"><span class=\\"vlist-r\\"><span class=\\"vlist\\" style=\\"height:0.8141079999999999em;\\"><span style=\\"top:-3.063em;margin-right:0.05em;\\"><span class=\\"pstrut\\" style=\\"height:2.7em;\\"></span><span class=\\"sizing reset-size6 size3 mtight\\"><span class=\\"mord mtight\\">2</span></span></span></span></span></span></span></span></span></span></span>",
```
DSchau: I'm a little confused why these tests _were_ passing if they needed updated? Any idea here?

pieh: unit tests will install update dependencies properly, so this seems like output of `katex@0.10.0` changed from `katex@0.9.0`

DSchau: ah got it. Yeah should've assumed that 🙃

packages/gatsby-remark-katex/package.json
hexrcs: @DSchau Thanks for the tips! Should I also move `unist-util-visit` and `remark-math` to `peerDependencies`?

DSchau: Just Katex for now, please!

hexrcs: @DSchau  Thanks for the clarification! I am not very familiar with dealing with peerDependencies, I see if I `yarn` it when peer depended, the reference to `katex` inside `yarn.lock` is removed. Is this expected?

Also should I rename this PR? I see many PRs here have names like `feat(Gatsby): `, `chores(Gatsby): ` etc., but couldn't find a full description of the naming convention used. Should this PR be called `update(gatsby-remark-katex)` or similar? :)

DSchau: @hexrcs thanks for noticing! We usually do that come merge time, but I went ahead and did it for you. The general idea is that the general _spirit_ of the PR (e.g. new feature, bug fix, chore, etc.) is to be used, so we can tie into conventional changelog stuff!

> I am not very familiar with dealing with peerDependencies, I see if I yarn it when peer depended, the reference to katex inside yarn.lock is removed. Is this expected?

Yup! So you'll want to do something like `yarn add katex --peer` which appears to be what you did. That should work just fine!

hexrcs: @DSchau Awesome, thanks! Will keep that in mind next time I contribute 😛

Updated this PR 😃

DSchau: We should actually _document_ it :) Because we'd love it if contributors would do that themselves, so seriously thanks so much for noticing and being aware of it 💜 

If you're interested in taking a stab at that document that would be 💯 too! It should probably be documented [here](https://www.gatsbyjs.org/docs/how-to-contribute/), and then maybe in the pull request template, as well.

No pressure at all though, we're already so glad you contributed!

hexrcs: > If you're interested in taking a stab at that document that would be 💯 too! It should probably be documented [here](https://www.gatsbyjs.org/docs/how-to-contribute/), and then maybe in the pull request template, as well.

Sure I'll look in to that! Glad I could help, I love Gatsby! 😄

DSchau:
Oh, one extra thing, we'll want to document that `katex` now needs to be installed alongside this plugin in the readme! Want to take a stab at that?

For an example, check out [`gatsby-remark-prismjs`](https://www.gatsbyjs.org/packages/gatsby-remark-prismjs/)

hexrcs:
@DSchau Yep that's indeed necessary now. Should I do that in this PR or create another one?

DSchau:
@hexrcs in this PR, please! Also while you're at it, could you add `katex` as a devDep (same semver pattern please!) so that the unit tests pass?

hexrcs:
@DSchau Sure! Just noticed bunch of red testing results here 😅

DSchau:
Thanks so much for the PR! Going to release this now!

PR #9952, opened 2018-11-15
ci: update config to exclude docs/blog test running

DSchau:
This will also hide the status check. The general idea is that any PR
whose branch starts with docs/, docs, blog/, or blog (e.g.
docs-some-update, docs/some-update, etc.) will not run e2e tests. This
seems pretty safe, and it's effectively opt-in. Unfortunately, we can't
really enforce this standard, so it's more relying on the community to
open up correct branch names, but if nothing else, it's valuable for us
:shrug:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,8 +44,8 @@ aliases:
     filters:
       branches:
         ignore:
-          - /docs(\/)?.*/
-          - /blog(\/)?.*/
+          - /docs.+/
+          - /blog.+/
 
   test_template: &test_template
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,6 +40,13 @@ aliases:
         ignore:
           - master
 
+  ignore_docs: &ignore_docs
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -40,6 +40,13 @@ aliases:
         ignore:
           - master
 
+  ignore_docs: &ignore_docs
+    filters:
+      branches:
+        ignore:
+          - /docs(\/)?.*/
```
pieh: NIT: is `(\/)?` needed if it's followed by `.*` anyway?

pieh:
> Should we document this too?

I'd say we should merge and see if this works as expected first and then adjust https://www.gatsbyjs.org/docs/how-to-contribute/ (contributing to documentation/blog sections)

.circleci/config.yml
DSchau: What about `docs.+`? It seems unlikely we need to account for raw `docs`, `blog`, etc. anyways.

PR #9954, opened 2018-11-15
Red highlight

KingofLegendary:
Random red highlighting on line 164 through line 203. The " ` " symbol may have done it or the " ' "symbol. made line 164 simpler without the use of the " ' " single quote symbol.


pieh:
I don't quite understand what do You mean with red highlighting?

This seems ok?
screen shot 2018-11-16 at 03 02 37

PR #9960, opened 2018-11-15
Add notice about StaticQuery only working with JSX

Siilwyn:
cc @DSchau


docs/docs/static-query.md
```diff
@@ -91,3 +91,4 @@ StaticQuery can do most of the things that page query can, including fragments.
 
 - page queries can accept variables (via `pageContext`) but can only be added to _page_ components
 - StaticQuery does not accept variables (hence the name "static"), but can be used in _any_ component, including pages
+- StaticQuery only works when called in combination with JSX templating
```
DSchau:
```suggestion
- StaticQuery does not work with raw React.createElement calls; please use JSX, e.g. `<StaticQuery />`
```

PR #9963, opened 2018-11-16
docs: Add "Babel Plugin Macros" guide

jgierer12:
Some thoughts:
- Rename to `Babel Macros`? This guide is more about the macros themselves than the plugin.
- Elaborate on the fact that the plugin is included with Gatsby, and has to be installed manually when using macros in a non-Gatsby project?
- Add instructions for some common macros? ([preval](https://www.npmjs.com/package/preval.macro), [codegen](https://www.npmjs.com/package/codegen.macro), [tagged-translations](https://www.npmjs.com/package/tagged-translations), ...)
This might get too long, and is pretty well described in the docs of each macro.

### DIFF
--- a/docs/docs/babel-plugin-macros.md
+++ b/docs/docs/babel-plugin-macros.md
@@ -2,7 +2,61 @@
 title: Babel Plugin Macros
 ---
 
-This is a stub. Help our community expand it.
+Gatsby includes a powerful new way of applying compile-time code
+transformations:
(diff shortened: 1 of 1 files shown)


KyleAMathews:
Reads great! I don't think it's necessary to make clear it's not installed by default in other frameworks as this is an advanced feature so most people reading it would be able to figure this out, probably.

I'll merge this now but if you'd like to do a follow-up PR listing some other popular plugins, that'd be great!

PR #9985, opened 2018-11-16
Use actions provided, don't assume Gatsby module location. Fixes #9984.

### LINKED ISSUE #9984: [gatsby-plugin-sharp] Requires incorrect gatsby module in mono-repo
## Description

We run Gatsby inside a Docker container to build our web site. Recently we had added a fair number of images and noticed a strange issue where our automed e2e test was failing due to images not being found (404 error). Investigation of the build confirmed that the relevant folders has been generated in `public/static` but were empty.

This was strange because it worked when being built locally on a developer and all images were produced. However, when running on our CI server, this was failing to produce all the images.

Given that the build was happening in a Docker container we assumed this must be some sort of race condition that occurs depending on the physical hardware used. Investigation confirmed that the CI server was taking longer to process each images and hence the overall build was completed before all images had been created.

This was strange as I was aware that the bootstrap process was meant to wait until all outstanding jobs were completed.

After some debugging, I found the root of the problem. `gatsby-plugin-sharp` acquires the bound action creators via this line:

```
const { boundActionCreators } = require(`gatsby/dist/redux/actions`)
``` 

We run Gatsby iniside a monorepo. The plugin is correctly sending the `CREATE_JOB` and `END_JOB` actions but these were going to a totally different Redux store due to the way Yarn had hoisted the dependencies! The require will find the "closest" `gatsby` which might not be the `gatsby` module executing.

Hence the correct store was never receiving the actions and the code in the bootstrap process to check for active jobs would pass even though the plugin was still working.

## Corrective Action

`gatsby-plugin-sharp` should use the `actions` that are passed to the `gatsby-node` API file. This will ensure the plugin is always referring to the correct store.

I will submit a pull request with this change. I've mainly created this issue for reference purposes in case it helps anyone else.

## Other considerations

I've not had time to check but any other plugins that pull out the `boundActionCreators` from ` require('gatsby/dist/redux/actions')` would also suffer from this bug of referencing the wrong Redux store which could lead to all sorts of weird behaviour.

martin-css:
The gatsby module resolve by `require` might not be the `gatsby` module that is executing. This can occur in mono-repo configurations.

This will prevent the Redux store from receiving the job actions and the bootstrap process will complete without waiting for all images to be processed. This can cause the build to finish without all images being generated.

### DIFF
--- a/packages/gatsby-plugin-sharp/src/gatsby-node.js
+++ b/packages/gatsby-plugin-sharp/src/gatsby-node.js
@@ -1,6 +1,6 @@
 const { setBoundActionCreators } = require('./index')
 
-exports.preInit = ({ actions }) => {
+exports.onPreInit = ({ actions }) => {
   setBoundActionCreators(actions)
 }
 
--- a/packages/gatsby-plugin-sharp/src/gatsby-node.js
+++ b/packages/gatsby-plugin-sharp/src/gatsby-node.js
@@ -1,3 +1,9 @@
+const { setBoundActionCreators } = require('./index')
+
+exports.preInit = ({ actions }) => {
+  setBoundActionCreators(actions)
+}
+
 // TODO
 // exports.formatJobMessage = jobs => {
 // return {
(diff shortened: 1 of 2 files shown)


martin-css:
Will tidy up to single commit.

PR #9997, opened 2018-11-17
Add our tech blog to sites.yml

Destroy666x:
As the title says.

### DIFF
--- a/docs/sites.yml
+++ b/docs/sites.yml
@@ -3391,3 +3391,19 @@
   built_by: Henry Kehlmann
   built_by_url: https://github.com/madhenry/
   featured: false
+- title: Boldare Tech Blog
+  url: https://tech.boldare.com/
+  main_url: https://tech.boldare.com/
+  source_url: https://github.com/xsolve-pl/boldare-tech
+  description: >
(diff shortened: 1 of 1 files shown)


KyleAMathews:
@Destroy666x can you fix the git conflict?

pieh:
@Destroy666x this site is heavily based on https://github.com/greglobinski/gatsby-starter-personal-blog - it's ok to use it of course as starter is open sourced, but site showcase entries should be original work - after all, You are claiming that https://www.boldare.com/ built site, which is not fully correct here (as big part of look and feel of the site was built by @greglobinski)

DSchau:
@Destroy666x I'm going to close this out for the reasons that @pieh specified.

We appreciate you using Gatsby, but we want to make sure that the original creators get the credit they deserve, and adding non-original showcase sites just blurs that line.

Please feel free to re-open or discuss further if you feel differently though. Thank you!

Destroy666x:
@DSchau @pieh your choice, but your reasoning doesn't make much sense to me. Nothing is really "original" here because everything's based on Gatsby, yet people don't claim Gatsby built the site either because that's simply invalid. And proper credits for the heavily modified base are given where the creator left them - in the footer.

All you did here was discourage people from using starters to make things happen faster because then they won't be able to share their site no matter what the idea is.

PR #9989, opened 2018-11-16
Added amazing DexWallet to the gatsby showcase

aesthytik:

PR #10002, opened 2018-11-17
Add DexWallet to featured section on showcase

aesthytik:


LekoArts:
As Kyle said in the old PR (https://github.com/gatsbyjs/gatsby/pull/9989): The Gatsby team will decide about that, you shouldn’t set that yourself.

LekoArts:
Please read the note.

However I find this section really misleading. @amberleyromo Why would a user need to set this and not the team?

PR #10010, opened 2018-11-18
fix dead links

nodox:


nodox:
@m-allanson @pieh quick fix for some dead links.

PR #10013, opened 2018-11-18
What files to add to the repo / add to the .gitignore???

debboutr:
This isn't a useful addition to this page, but it would be good to include what files need to be checked into the git repo, I'd imagine we can throw the node-modules and possibly others into the .gitignore. I've also noticed the the `/src` files and others don't get pushed to the github.io site. Is the idea that you would have another repo that would manage all of your commits while developing and then the push to github.io would just be the `/public` folder?? Sorry that this is more of a question than an addition, but I think this would be good to add here. Thanks!

If I set up the way it is stated in the docs, I wouldn't then be able to clone that `<reponame>.github.io` and build from there so, what is the best way to manage that?


LekoArts:
Please create an issue for questions in the future and not a PR. You normally push the content to another branch. Have a look at the doc:
https://www.gatsbyjs.org/docs/how-gatsby-works-with-github-pages/

PR #10022, opened 2018-11-19
docs(gatsby-plugin-manifest): Change "minimal-ui" to "standalone"

LekoArts:
https://developers.google.com/web/fundamentals/web-app-manifest/#display

> In order to show the Add to Home Screen Prompt, display must be set to **standalone**.

I've always automatically used `standalone` (and it works) but just now realized that the README/docs use the wrong property.


gil--:
Not sure I agree with this change since it sets the default for all sites to standalone. On iOS, standalone has numerous issues which make it almost unusable:

- No state persistence so if you leave the PWA and then come back, whatever page or data was present is gone and you start at the beginning.
- Opening a link to an external page pops you back to Safari
- Many sites don't have the required UI to operated standalone (missing back button, etc.) and on iOS, swipe gestures from Safari don't work so you could end up stuck.

I'm of the opinion that it should be kept with standard UI as default until PWA standalone support on Safari improves.

LekoArts:
> Not sure I agree with this change since it sets the default for all sites to standalone

It’s just the documentation. People can still use other values.

> On iOS, standalone has numerous issues which make it almost unusable

That might be right, but you also have these issues with `minimal-ui` (according to my research on the internet). Apple is just bad at their implementation.

Since Gatsby claims to be a PWA generator, this needs to be changed to `standalone` to fulfill the requirements.

LekoArts:
@pieh @dschau Please review :)

LekoArts:
> We need to be careful about defaults we propose, because a lot of people will copy & paste it as-is without much thinking leaving them to hit problems they don't know that they can hit.

I know that many will copy/paste this. But the manifest plugin (+ offline) promise you to have a 100% PWA. And for that you need to have the [Add to Homescreen](https://developers.google.com/web/fundamentals/web-app-manifest/#display) note. And IMO this should be the priority here, to fulfill the PWA specs. To quote from [this issue](https://github.com/w3c/manifest/issues/673):

> More generally, standalone mode explicitly says `Opens the web application to look and feel like a standalone native application`. It's not the role of the manifest spec to enforce that implementing platforms break their own UX norms. iOS never shows a back button in system-level UI: native applications must handle their own back and forward, so a PWA that works well on that platform should be reasonably expected to do the same. It's the developer's prerogative to implement their app in a way that fits the native application model on each platform.

With iOS you already know that there's no back button, it doesn't come to any surprise.

Of course we can / should add a note to the README / tutorial that `standalone` has this behavior with links to the Google documentation.

pieh:
For visibility:
We discussed a bit with @LekoArts - "transcript":

@pieh:
> but with this being changed in multiple places it is getting hard to be sure that user will be aware of implication of the `display` setting
and people that are aware of implications, will probably set proper setting themselves
> > But the manifest plugin (+ offline) promise you to have a 100% PWA. And for that you need to have the Add to Homescreen note.
>
> Where this comes from? Part that to have 100% PWA you need to have “add to homescreen” note (I’m not disagreeing, just trying to get context I lack :slightly_smiling_face: )

@LekoArts:
> > Where this comes from? Part that to have 100% PWA you need to have “add to homescreen” note
>
> Google's site (https://developers.google.com/web/progressive-web-apps/) and all other pages writing about PWA say, that you get an App-like experience and also this "add to homescreen" prompt. The checklist also wants that: https://developers.google.com/web/progressive-web-apps/checklist#metadata-provided-for-add-to-home-screen

@pieh:
> what about something like this:
> ```
> display: `standalone`, // this will enable "add to homescreen" prompt and disable browser UI (including back button), see <link to either our docs with more detailed stuff or any 3rd party page that explains it well enough> 
> ```

@LekoArts:
> Yeah, I'm fine with that. Currently that only would need to be included in three files. I'd probably would link to this page: https://developers.google.com/web/fundamentals/web-app-manifest/#display

LekoArts:
CI fails for whatever reason 🤔

DSchau:
This seems like a legitimate failure. We need to merge in master. Doing that now

> https://registry.yarnpkg.com/flatmap-stream/-/flatmap-stream-0.1.0.tgz: Extracting tar content of undefined failed, the file appears to be corrupt: "Unexpected end of data"

packages/gatsby-source-faker/README.md
```diff
@@ -1,37 +1,37 @@
...
-      },
-      count: 3, // how many fake objects you need
-      type: "NameData", // Name of the graphql query node
-    },
-  },
-]
-```
-
-Example: [Using Faker](https://github.com/gatsbyjs/gatsby/tree/master/examples/using-faker)
+## gatsby-source-faker
```
DSchau: I am not honestly sure why this file was updated.

LekoArts: Also happened to me locally. Don't know why :(

DSchau: i.e. after this change, right? Or you mean it happened to you earlier?

LekoArts: Earlier, when I checked out different branches (don't know the exact steps to reproduce it)

DSchau: I ran `dos2unix` on this file so _maybe_ it's fixed now 🤞 

But again, I don't really know the issue so 🙃

shannonbux:
Relevant part of the style guide: 

> There is an infinite number of tasks that the documentation and tutorials could include, so they should not attempt to teach every way of doing a task. Usually, they should teach the following:
> - the most fundamental way possible to do a task with Gatsby
> - the most common way of doing a task
> - the best practice way of doing a task

taken from a PR https://github.com/gatsbyjs/gatsby/pull/9894/files


also want feedback on that to merge it!

LekoArts:
What sorta feedback? For me this would be the „best practice“. And in regards to PWA probably also the most common way as you want your app not to start in your standard browser.

shannonbux:
@LekoArts feedback on whether the PR I linked to is an improvement to the style guide :)

And as far as whether this change is a best practice, I'll leave that up to more technical folks!

LekoArts:
@shannonbux Uhm, the PR seems good, but I agree with Dustin on the "order". Then I'd put the best practice at the top...

@pieh Do you feel the PR is now read for merge?

pieh:
Yeah, let's merge this - this is more of a doc change, so I adjust PR title here

PR #10027, opened 2018-11-19
WIP to add plug and play functionality

Ikana:
It couldn’t finish the last task in the given 3 hours, but this is what i did

* I read the yarn plug and play white paper

* From there I configured the development environment for Gatsby with the [guide provided]().

* There I started to modify the Yargs command for new

* As per the white paper on the new flow if the flag --use-pnp I inject `
{
  "installConfig": {
"pnp": true }
}` into the package.json of the downloaded starter

Here I came into some problems, in order to plug and play to work, all the modules used directly by a package need to be in it’s package.json. Though this is best practice you can get away with it, in traditional installations, because yarn flattens all the dependencies, and if one of the packages you already installed, has that module that was not provided in the package.json, as it’s dependency, then that module would be inside node_modules and available for use.

Here is a list of the packages that gatsby uses but does not explicitly defined in the package.json

* "browserslist", "3.2.8"
* "fs-exists-cached", "1.0.0"
* "babel-traverse", "6.26.0"
* "address", "1.0.3"

Then the module `relay-compiler@1.5.0` had the wrong value, for the field main in the package.json, for that version it has to be either `index.js` or `lib/RelayCompilerPublic.js` it has been corrected in newer versions of `relay-compiler` but upgrading to those versions makes some tests inside gatsby fail. I found this bug reading this part of the [node documentation ](https://nodejs.org/api/modules.html#modules_all_together).

Also the `graphql-skip-limit` module has the same problem as gatsby meaning it fails to explicitly define some modules in the package.json. They may be more.

To make this work we would need to

* Update our relay-compiler
* Correct the failing tests and update the code to use the new version of the relay-compiler
* Comb through the dependencies of gatsby and modify them to explicitly specify their dependencies
* Make the needed pull requests


DSchau:
@Ikana going to close this out for now. Thanks!

PR #10030, opened 2018-11-19
fix: avoid leaking defined env vars when trying to access not defined env vars

### LINKED ISSUE #10021: Security: all env vars in production bundle
## Description

While inspecting production bundles of a gatsby site, I saw that all environment variables used during the build process are included in a bundle js file. This can also include sensitive data (e.g WordPress htaccess passwords) when used via a .env file.

### Steps to reproduce

- Include .env files as documented on https://www.gatsbyjs.org/docs/environment-variables/ 
- Run production build
- Inspect js bundle `app-*.js` for env vars

### Expected result

Only env vars starting with `GATSBY_` (and maybe others necessary for the build - which ones?) should be included in production bundles, as stated on https://www.gatsbyjs.org/docs/environment-variables/

### Possible solution

I would guess that the problem is this line in the `webpack.config.js`: https://github.com/gatsbyjs/gatsby/blob/1858365bc306c164784bf164b0b5a148001feec0/packages/gatsby/src/utils/webpack.config.js#L70
Maybe we could white-list necessary env vars (including those starting with `GATSBY_ `) and ignore all others. Would love to hear your thoughts on this.

pieh:
fixes #10021

With current approach when `process.env.NOT_EXISTING_ENV_VAR` is used then bundle will contain:
`Object({ NODE_ENV: "production", ...otherEnvVars }).NOT_EXISTING_ENV_VAR` - with this approach it will produce: `{}.NOT_EXISTING_ENV_VAR`, while still being able to access all defined vars.

This is important because not only user code can be responsible for leaks - all imported npm packages can cause this as well.

Limitation - users won't be able to iterate over `process.env` object but this was exactly what was causing potential leak.


elskwid:
@pieh thank you for the security fix. I don't know the best place to make mention of this but the change will seriously impact anyone that is destructuring environment variables.

```
const { FOO, BAR } = process.env
```

Will result in undefined values for both.

brimtown:
Is there a reason that this was not communicated as a breaking change? It's surprising to have an app break after a patch release 🙁 .

jonniebigodes:
@brimtown this was brought to the attention of the team and from the flow of commits and the issue associated to it, needed to be patched with some urgency to prevent further damage to the people using gatsby.

DSchau:
@brimtown you're definitely right, though. This probably should have been communicated and pushed out as something beyond a patch release. Minimally, we should have marked it as a breaking change so it shows up in `CHANGELOG.md`.

We're sorry for the confusion this has caused, and we'll attempt to improve the documentation and communication on these types of changes going forward!

brimtown:
Thanks for the response. Appreciate the hard work yall are doing!

elskwid:
@DSchau what @brimtown said. 😄 Thanks!

PR #9401, opened 2018-10-25
fix(gatsby-plugin-jss): use separate SheetsRegistry for each page

dweirich:
Not sure about the branch for this one. Just a small patch based off of cpboyd's solution for this issue: https://github.com/gatsbyjs/gatsby/issues/7716

closes #7716

PR #10045, opened 2018-11-20
fix: update peerDependency, plugins rely on extra arg being passed in API hooks

pieh:
Those plugins rely on changes introduced in https://github.com/gatsbyjs/gatsby/pull/9401 which was released in `gatsby@2.0.32`


DSchau:
Successfully published:
 - gatsby-plugin-jss@2.0.4
 - gatsby-plugin-styled-components@3.0.3

PR #10062, opened 2018-11-21
chore(gatsby): upgrade autoprefixer

alexandernanberg:
Upgrade autoprefixer to the latest major version. [Release notes](https://github.com/postcss/autoprefixer/releases/tag/9.0.0).


packages/gatsby/package.json
```diff
@@ -17,7 +17,7 @@
     "@babel/runtime": "^7.0.0",
     "@babel/traverse": "^7.0.0",
     "@reach/router": "^1.1.1",
-    "autoprefixer": "^8.6.5",
+    "autoprefixer": "^9.0.0",
```
leachjustin18: The latest version is 9.3.1: https://www.npmjs.com/package/autoprefixer I'm curious why we are specifying 9.0.0 here if we are updating it?

leachjustin18: Thank you for the reply.  I leave it up to you on the exact version.

wardpeet: @alexandernanberg could you bump `^9.3.1` it doesn't matter if it's unnecessary or not but it gives us a sign what version we actually tested.

Issue #9977, opened 2018-11-16
[gatsby-plugin-manifest] Add option to remove theme-color meta tag


GriffinJohnston:
## Summary

gatsby-plugin-manifest inserts a theme-color meta tag into the site head. This means that you can't use React Helmet to set the theme_color, which is a problem for sites with a dark mode/theme-switcher/etc that want to programmatically change color schemes. 

We could either remove the meta tag insertion entirely, since it's redundant with the theme_color value in the manifest, or — if the meta tag is required for some corner-cases I'm not aware of — add an option to turn off this behavior.

I'm just diving into Gatsby, and I've been super impressed so far. Thanks everyone for the hard work on this awesome project!

DSchau:
This would be a great first PR! Would you be interested in making the changes?

Whether you are or not, I can walk through _how_ someone would make the changes.

## Feature Request: Allow disabling of theme-color from meta-tag

In some scenarios, it may make sense to disable the automatic meta `theme-color` injection, i.e. in scenarios where the application is themed and the theme-color tag will need to update.

## What's involved

- Adding an option `disableThemeColor` (feel free to tweak this) which is defaulted to `true`
- Detecting this condition, and if false, _not_ adding the meta tag

## Technical Detail

1. Add some default options [here](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-manifest/src/gatsby-ssr.js#L5-L6)
1. Detect condition (if false) and conditionally do not add if false [here](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-manifest/src/gatsby-ssr.js#L23-L34)
    - Note, Array.prototype.concat is helpful here, e.g. ['red'].concat(true ? ['blue'] : []).concat(false ? ['green'] : []) will create `['red', 'blue']`
1. Add some unit tests validating the functionality (note: jest.fn() is super helpful here, e.g. a spy!)

GriffinJohnston:
Thanks for the quick reply! I'll put together a PR when I get a moment.

DZuz14:
Didn't mean to steal your thunder here @GriffinJohnston, I can rescind this PR if you were indeed currently working on this.

GriffinJohnston:
@DZuz14 Not at all! I got really busy soon after posting this and never got around to it. Glad you jumped on it!

pieh:
[...]

PR #10069, opened 2018-11-21
feat(gatsby-plugin-manifest): don't output `theme-color` if it's not defiened

DZuz14:
Not really sure how I go about submitting a PR here. This addition references issue https://github.com/gatsbyjs/gatsby/issues/9977

Any guidance would be greatly appreciated.


packages/gatsby-plugin-manifest/src/gatsby-ssr.js
```diff
@@ -20,18 +20,30 @@ exports.onRenderBody = ({ setHeadComponents }, pluginOptions) => {
...
+        ? [
+            <meta
+              key={`gatsby-plugin-manifest-meta`}
+              name="theme-color"
+              content={pluginOptions.theme_color}
+            />,
+          ]
+        : []
+    )
+  )
```
pieh: Can we prep headComponents array ahead of time to avoid concat with ternary operator inside it?
```js
const headComponents = [
  <link
    key={`gatsby-plugin-manifest-link`}
    rel="manifest"
    href={withPrefix(`/manifest.webmanifest`)}
  />,
]

if (pluginOptions.theme_color) {
  headComponents.push(
    <meta
      key={`gatsby-plugin-manifest-meta`}
      name="theme-color"
      content={pluginOptions.theme_color}
    />
  )
}

setHeadComponents(headComponents)

DZuz14: Yeah, I thought I got rid of those changes to that ReadMe. I am not even sure how they got changed to be honest. Dog stepping on keyboard perhaps lol.

DZuz14: Is it alright that I created the headComponents array at the top of the function, and changed the other conditionals to just add to this array as we step through the function? I think this makes it pretty legible and easy to understand whats going on. I ran the tests after this change and everything still seems to work just fine.

pieh: Oh, yeah, that's fine too 👍

pieh: Hmm, actually this README changes (that your dog committed 🤣 ) seems like something very weird is happening and I get them locally too, I'll just merge with those - hopefully this will fix problems lol

pieh:
published `gatsby-plugin-manifest@2.0.10`

PR #10071, opened 2018-11-21
Update functions to the es6 style

jamesgeorge007:
> Update `es5` functions to the `es6` way which is a more concise form of representation.

### DIFF
--- a/packages/gatsby/src/commands/develop.js
+++ b/packages/gatsby/src/commands/develop.js
@@ -287,7 +287,7 @@ module.exports = async (program: any) => {
     })
   })
 
-  function prepareUrls(protocol, host, port) {
+  const prepareUrls = (protocol, host, port) => {
     const formatUrl = hostname =>
       url.format({
         protocol,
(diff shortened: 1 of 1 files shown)


pieh:
It actually adds more characters than removes, so I don't know about "more concise form of representation" ;)

Personally I'm against changes just for sake of changes that don't solve anything, but if you can prove that this change is valuable I can change my mind.

jamesgeorge007:
@pieh Actually most of the methods used in the same file is an arrow function. I thought why not changing those wrapper functions to the same as well. What do you think?
Anyways arrow functions have significant advantages over the traditional functions 👍

pieh:
> Anyways arrow functions have significant advantages over the traditional functions 👍

What advantages that would apply in this scenario?

jamesgeorge007:
Most of the functions defined within the respective files are arrow functions except for a few of those wrapping them. There is a practice to use the same syntax everywhere within a file right? (consistency in code) 
Just thoughts 🤔

pieh:
For real Code Consistency we would need to use something like https://www.npmjs.com/package/eslint-plugin-prefer-arrow to enforce this rule everywhere and not just in single file. But I'm not sure if tradeoff between value and inconvenience of this rule is worth it. 

I'll close this PR for now, because it doesn't fix anything and all files are passing linting checks, so this PR just changes things for the sake of change

PR #10085, opened 2018-11-22
add secondary fields to www query sorts

Moocar:
While working on loki (#9919), I've run into lots of fun edge cases to do with sorting. One is that some connection queries specify a sort order, but some returned fields have the same value for the sort field. The behavior here is undefined, but by default sift will order by the order they were added to Gatsby. While that's not a terrible default, it makes testing sift and loki compatibility hard, since loki has undefined ordering for this case.

So this PR adds a specific secondary sort key to various `www` queries, mainly for the purpose of testing sift/loki compatibility, but also because it makes sense.

### DIFF
--- a/www/src/pages/index.js
+++ b/www/src/pages/index.js
@@ -301,7 +301,7 @@ export const pageQuery = graphql`
       }
     }
     allMarkdownRemark(
-      sort: { order: DESC, fields: [frontmatter___date] }
+      sort: { order: DESC, fields: [frontmatter___date, fields___slug] }
       limit: 3
       filter: {
         frontmatter: { draft: { ne: true } }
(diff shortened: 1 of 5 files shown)


www/src/pages/index.js
```diff
@@ -319,6 +319,7 @@ export const pageQuery = graphql`
       filter: {
         fields: { starterShowcase: { slug: { in: $featuredStarters } } }
       }
+      sort: { fields: [fields___starterShowcase___slug] }
```
pieh: @greglobinski this will sort starters for "ecosystem" section on homepage - can this cause unwanted change?

pieh: Just merged ;)

PR #10093, opened 2018-11-22
fix(gatsby-remark-images-contentful)

t2ca:
This is a minor fix to the breaking changes that happened in gatsby-remark-images@3.0.0

### BREAKING CHANGES

- **gatsby-remark-images:** html markup has been changed, high resolution image (img.gatsby-resp-image-image) is no longer nested inside span with low resolution placeholder (span.gatsby-resp-image-wrapper) - it's sibling of that span now. This might affect any custom styling that is applied to inline images

### DIFF
--- a/packages/gatsby-remark-images-contentful/src/index.js
+++ b/packages/gatsby-remark-images-contentful/src/index.js
@@ -99,7 +99,7 @@ module.exports = async (
       style="padding-bottom: ${ratio}; position: relative; bottom: 0; left: 0; background-image: url('${
       responsiveSizesResult.base64
     }'); background-size: cover; display: block;"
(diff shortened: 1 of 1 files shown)


pieh:
I'm thinking if better way than maintaining those 2 plugins side by side would be to add way to `gatsby-remark-images` to handle contentful images. Maybe not straight builtin, but as subplugin maybe? (subplugin in subplugin alert)

wardpeet:
@t2ca thanks for the effort! I haven't found any bugs about this or I've missed these (sorry if I did).

If we really need this I suggest creating an issue were we can discuss how we should fix this as we don't like breaking changes :)

PR #10127, opened 2018-11-26
[gatsby-plugin-google-analytics] Add option for enabling/disabling tracking

### LINKED ISSUE #9796: [gatsby-plugin-google-analytics] Log events when running `gatsby develop`
### Summary
Allow Analytics tracking in development mode, behind a gatsby-config.js flag

### Basic example
One possible solution would be adding a flag in the plugin config called `enabled`. If true, log analytics events. If false, do not log them. The default value could be `process.env.NODE_ENV === 'production'` to not change the default behavior.
Another solution could be a config flag called `onlyInProduction`, with a default of `true`. If the user sets it to `false` then analytics get logged in development mode.

### Motivation
Right now, `gatsby-plugin-google-analytics` only tracks when `env.NODE_ENV === 'production'`. However, I want to test out some analytics stuff while developing, and it's a pain to run `gatsby build` and start a static file server every time I make a change. 
I'm filtering all `localhost` entries out of my analytics already, so extra noise coming from my development environment isn't a problem.

jkjustjoshing:
Fixes #9796 

I tried pulling out the `isEnabled` logic into a separate file to include in `gatsby-ssr.js` and `gatsby-browser.js`, but I couldn't easily get the `require()` to compile properly for running in the browser. I opted to just repeat the logic. If anyone has any suggestions on how to get that compilation working I'm happy to throw it into the PR.

### DIFF
--- a/packages/gatsby-plugin-google-analytics/README.md
+++ b/packages/gatsby-plugin-google-analytics/README.md
@@ -24,6 +24,8 @@ module.exports = {
         respectDNT: true,
         // Avoids sending pageview hits from custom paths
         exclude: ["/preview/**", "/do-not-track/me/too/"],
+        // Enable tracking
+        enabled: "auto",
         // Enables Google Optimize using your container Id
         optimizeId: "YOUR_GOOGLE_OPTIMIZE_TRACKING_ID",
         // Any additional create only fields (optional)
@@ -87,6 +89,10 @@ If you enable this optional option, Google Analytics will not be loaded at all f
 ## The "exclude" option
 
 If you need to exclude any path from the tracking system, you can add it (one or more) to this optional array as glob expressions.
+
+## The "enabled" option
+
+By default, tracking is disabled in development mode and enabled in production. If you want to fully disable/enable tracking in both environments, set this to `true` or `false`. To track in production but not development, omit the option or set to `'auto'`.
 
 ## The "optimizeId" option
 
--- a/packages/gatsby-plugin-google-analytics/src/gatsby-browser.js
+++ b/packages/gatsby-plugin-google-analytics/src/gatsby-browser.js
@@ -1,6 +1,20 @@
-exports.onRouteUpdate = function({ location }) {
-  // Don't track while developing.
-  if (process.env.NODE_ENV === `production` && typeof ga === `function`) {
+exports.onRouteUpdate = function({ location }, pluginOptions) {
+  let { enabled } = pluginOptions
+  let isEnabled
+  if (typeof enabled === `boolean`) {
+    isEnabled = enabled
+  } else if (enabled && enabled !== `auto`) {
+    // If user specifies any value that is truthy but NOT the string `auto` it's invalid
+    throw new Error(
+      `[gatsby-plugin-google-analytics] Valid options for 'enabled' flag: true, false, or 'auto'`
+    )
+  } else {
+    // Default value, if `enabled` option is not specified
+    isEnabled = process.env.NODE_ENV === `production`
+  }
+
+  // Don't track while developing, or follow the `enabled` option value
+  if (isEnabled && typeof ga === `function`) {
     if (
       location &&
       typeof window.excludeGAPaths !== `undefined` &&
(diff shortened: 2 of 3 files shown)


LekoArts:
I don’t think that’s a good idea. There’s the effort that `develop` and `build` have the same behavior but that’s not the case. The best environment to test it is still with the build files. But this option encourages exactly that.

jkjustjoshing:
@LenkoArts Could you elaborate? Why is it not the case that `develop` and `build` will have the same behavior? What is the disadvantage to being able to test analytics in `develop` (right now it's impossible to verify, for example, that manual logging an event works while developing)?

The harm I could see this causing is polluting analytics logs with events from while developing. My project uses 2 safeguards against that: 1) I set my analytics key as a rubbish value if the environment isnt production and 2) my analytics dashboard filters out `localhost` domain events. Are there any ways you can think of to encourage the use of those safeguards for users who want to enable logging in `develop`?

jkjustjoshing:
@LekoArts can you let me know what your concerns are for this? Thanks!

LekoArts:
> Why is it not the case that develop and build will have the same behavior

Because one it running a hot-reloading webpack server, the other one just serves the production ready files. Due to the hot-reloading everything is in the cache and some things (e.g. page loading) behave differently than on the build. In the development server you (as far as I know) load the page twice (at least that's what I saw).

All this leads to a false sense of security, e.g. regarding your analytics. Other plugins also deactivate themselves because of the above reasons in development mode.

LekoArts:
@jkjustjoshing Hey, wanted to circle back here and ask what the status is :)

packages/gatsby-plugin-google-analytics/README.md
```diff
@@ -88,6 +90,10 @@ If you enable this optional option, Google Analytics will not be loaded at all f
 
 If you need to exclude any path from the tracking system, you can add it (one or more) to this optional array as glob expressions.
 
+## The "enabled" option
+
+By default, tracking is disabled in development mode and enabled in production. If you want to fully disable/enable tracking in both environments, set this to `true` or `false`. To track in production but not development, omit the option or set to `'auto'`.
```
pieh: It would be good idea to add warning/note about using this in development (polluting analytics log) and provide sample setup that can be used to test this without polluting logs for production

packages/gatsby-plugin-google-analytics/src/gatsby-browser.js
```diff
@@ -1,6 +1,20 @@
...
+    isEnabled = enabled
+  } else if (enabled && enabled !== `auto`) {
+    // If user specifies any value that is truthy but NOT the string `auto` it's invalid
+    throw new Error(
+      `[gatsby-plugin-google-analytics] Valid options for 'enabled' flag: true, false, or 'auto'`
+    )
+  } else {
+    // Default value, if `enabled` option is not specified
+    isEnabled = process.env.NODE_ENV === `production`
+  }
```
pieh: determining if it's disabled or enabled can be done in build time (and also not be duplicated in `gatsby-ssr`). One way to do it, is replicate setup from https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-layout/src/gatsby-node.js:
 - validate and "compute" `isEnabled` flag in `onPreInit`
 - pass computed `isEnabled` to `DefinePlugin` in `onCreateWebpackConfig` (for example as `GATSBY_GOOGLE_ANALYICS_ENABLED`
 - use it in both gatsby-ssr / gatsby-browser: ``if (GATSBY_GOOGLE_ANALYICS_ENABLED && typeof ga === `function`) {``

jkjustjoshing:
@LekoArts sorry I haven't had time to update this recently. Hopefully I will have time soon, but if you need to close this to clean things up I'm happy to make a new PR when I have time

wardpeet:
Thanks so much for opening this PR! It seems like you're not picking this up right away so I'm closing this one until you or someone else might want to work on this one.

If this is not the case, or if we can help further--please don’t hesitate to reach out or comment on this pull request, and we’d love to take another look.

Thanks for using Gatsby 💪

alexblack:
Did this go anywhere?  I'd love to get GA working in dev/test mode, so we can build and test analytics (we have separate dev and prod GA profiles)

PR #10137, opened 2018-11-26
docs(gatsby-link): add example of passing state

GarrettJMU:
Update docs to show how state is passed through the location object.

### DIFF
--- a/docs/docs/gatsby-link.md
+++ b/docs/docs/gatsby-link.md
@@ -93,8 +93,6 @@ render () {
 }
 ```
 
-
-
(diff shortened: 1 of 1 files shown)


GarrettJMU:
@jgierer12 updated the copy. Good call, much better read now.

jgierer12:
Thank you! Please also download [Prettier](https://prettier.io/) and use it to format the file, that will get rid of the failing CI too 😉

GarrettJMU:
@jgierer12 should pass now.

PR #10128, opened 2018-11-26
Adding test report tasks to Azure pipelines

vagisha-nidhi:
**What kind of change does this PR introduce?**

This PR is related to the issue #10103 
The PR helps to publish test results for reporting in Azure Pipelines.

**Summary**

Azure Pipelines provides excellent reporting capabilities that can be leveraged by this PR.
An example report is shown in the image below.
image
A screenshot for code coverage : 
image

**Does this PR introduce a breaking change?**
No

PR #10140, opened 2018-11-26
ci: do not report coverage unless in CI

DSchau:
In #10128, we added test report tasks to Azure pipelines. Unfortunately,
this adds some additional overhead and coverage type stuff in _local_
development that we probably don't want.

This PR addresses this issue, and conditionally detects whether running
in CI. If running in CI (e.g. Azure Pipelines, Circle, etc.) it will add
the additional coverage reporter and coverage functionality.

Note: this is currently blocked by watson/ci-info#22


DSchau:
This was fixed upstream (watson/ci-info/pull/23) and released as ci-info@2.0.0.

However - I think @pieh fixed this perfectly well, e.g. we pass an environment variable (GENERATE_JEST_REPORT=true) which means we don't have to go this approach.

I'll close this, but it's possible we may want to integrate `ci-info` in the future.

PR #10146, opened 2018-11-26
feat(gatsby): provide a mechanism for plugins to use a named cache instance

### LINKED ISSUE #8788: cache instance for subplugins
After implementing separate cache for each plugin, there is issue with cache instance for subplugins:
- subplugins can implement their own API hooks and there they get cache instance for that subplugin
- way parent plugins interact with subplugins is most often implementing most of subplugin functionality in something like `index.js` in this is called by parent plugin and often we pass cache from parent plugin to subplugin - for example in `gatsby-transformer-remark`: https://github.com/gatsbyjs/gatsby/blob/37489fabcc7aa78b5c97078c8aba80731a167799/packages/gatsby-transformer-remark/src/extend-node-type.js#L186-L203

Because of that different cache instances are used in subplugins `gatsby-node.js` and `index.js` files.

You can check https://github.com/pieh/gatsby-subplugins-cache reproduction (most important thing is in https://github.com/pieh/gatsby-subplugins-cache/tree/master/plugins/gatsby-remark-test ).

I think we would need to pass cache manager to plugins as well so parent plugins can pass correct cache instances to subplugins

/cc @DSchau @raae

DSchau:
Fixes #8788

This PR provides a mechanism for any plugin to call a `getCache` helper, which takes a single argument (name) and returns or creates a new cache instance for that named instance.

This is particularly applicable to cases where a plugin has sub-plugins (e.g. gatsby-transformer-remark). In this scenario, we have a cache mismatch, essentially the cache provided by Gatsby is named to the plugins name, whereas the cache provided by gatsby-transformer-remark is named `gatsby-transformer-remark`

This PR fixes this imbalance. Few questions/comments:

- I think it might make sense to override the cache passed, e.g. `cache: getCache(plugin.name)` for sub plugins. We can still inject getCache so if the parent cache is needed (rarely?) this will still function, as expected
- I'll need to bump the peer dependency of gatsby-transformer-remark to the next version of gatsby at merge time

### DIFF
--- a/packages/gatsby-transformer-remark/src/extend-node-type.js
+++ b/packages/gatsby-transformer-remark/src/extend-node-type.js
@@ -62,7 +62,7 @@ const withPathPrefix = (url, pathPrefix) =>
 const ASTPromiseMap = new Map()
 
 module.exports = (
-  { type, store, pathPrefix, getNode, getNodesByType, cache, reporter },
+  { type, pathPrefix, getNode, getNodesByType, cache, reporter, ...rest },
   pluginOptions
 ) => {
   if (type.name !== `MarkdownRemark`) {
@@ -133,6 +133,7 @@ module.exports = (
                     getNode,
                     reporter,
                     cache,
+                    ...rest,
                   },
                   plugin.pluginOptions
                 )
@@ -201,6 +202,7 @@ module.exports = (
                       pathPrefix,
                       reporter,
                       cache,
+                      ...rest,
                     },
                     plugin.pluginOptions
                   )
--- a/packages/gatsby-transformer-remark/src/on-node-create.js
+++ b/packages/gatsby-transformer-remark/src/on-node-create.js
@@ -3,7 +3,7 @@ const crypto = require(`crypto`)
 const _ = require(`lodash`)
 
 module.exports = async function onCreateNode(
-  { node, getNode, loadNodeContent, actions, createNodeId, reporter },
+  { node, loadNodeContent, actions, createNodeId, reporter },
   pluginOptions
 ) {
   const { createNode, createParentChildLink } = actions
--- a/packages/gatsby/src/utils/__tests__/get-cache.js
+++ b/packages/gatsby/src/utils/__tests__/get-cache.js
@@ -1,0 +1,21 @@
+const getCache = require(`../get-cache`)
+
+const CACHE_KEY = `__test__`
+
+test(`it returns a new cache instance`, () => {
+  const cache = getCache(CACHE_KEY)
+
+  expect(cache.get).toEqual(expect.any(Function))
+  expect(cache.set).toEqual(expect.any(Function))
+})
+
+test(`it retrieves already created cache instance`, async () => {
+  const key = `some-value`
+  const value = [`a`, `b`, `c`]
+  const cache = getCache(CACHE_KEY)
+  await cache.set(key, value)
+
+  const other = getCache(CACHE_KEY)
+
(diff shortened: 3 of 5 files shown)


packages/gatsby-transformer-remark/src/on-node-create.js
```diff
@@ -3,7 +3,7 @@ const crypto = require(`crypto`)
 const _ = require(`lodash`)
 
 module.exports = async function onCreateNode(
-  { node, getNode, loadNodeContent, actions, createNodeId, reporter },
```
DSchau: Wasn't used

packages/gatsby-transformer-remark/src/extend-node-type.js
```diff
@@ -133,6 +133,7 @@ module.exports = (
                     getNode,
                     reporter,
                     cache,
+                    ...rest,
```
pieh: I wonder if doing something like:
```
cache: getCache(plugin.id)
```
to pass individual cache for subplugin woudln't make sense, so subplugins don't need to use `getCache` directly, and would automatically get same `cache` instance as in their `gatsby-node` API hooks? 

Questions:
 - Is there scenario where it is actually desirable that subplugin use parent plugin cache?
I can't think if reason for this of top of my head.
 - Will this break any subplugin?
Cache key generating functions ( https://github.com/gatsbyjs/gatsby/blob/6a8c59385d57b256d839fd861f3cc6be0e262cfc/packages/gatsby-transformer-remark/src/extend-node-type.js#L31-L50 ) are not exported, so shouldn't be considered as public API, so this shouldn't be considered as breaking change.

pieh: Ultimately I would like to prep helper functions/objects in core, so parent plugin could pass prepped helpers, but this is out of scope for this PR I think

DSchau:
@pieh I provided a relatively janky e2e test to validate this functionality in 307b49e. Not in love with it, but it is a nice way to validate this--open to suggestions or just removing it.

One thought I had here is that this is actually a breaking change for gatsby-transformer-remark as currently authored. `getCache` will not be passed to plugins in versions of gatsby < _the version published here_, so I think I'd propose the following:

- Rip out `gatsby-transformer-remark` stuff from this PR
- Publish this PR, largely just exposing the API to plugins
- Bump to a minor (or major?) release of gatsby-transformer-remark
- Add gatsby@_the version published here_ as a peerDependency

Anything I'm missing? Sound reasonable?

pieh:
> @pieh I provided a relatively janky e2e test to validate this functionality in [307b49e](https://github.com/gatsbyjs/gatsby/commit/307b49ee37d55f17364cfaaa6931f0e594e71cca). Not in love with it, but it is a nice way to validate this--open to suggestions or just removing it.

the `><` replacement might seem a bit whacky - but it works, so all is good.

As for `peerDeps` bump - splitting PR seems reasonable. Don't tell anyone but when I was bit lazy in situation like this before I was manually bumping peerDeps to version that will yet be released just before merging and publishing instead of splitting PR.

> Alternatively, I can write a little utility that will guard against `getCache` not existing, and fall back to the parent cache, e.g.

I think we should do both and even possibly warn user to update gatsby version if `getCache` doesn't exist.

DSchau:
@pieh lol I almost proposed _guessing_ what the peerDep would be 😆

I'll get this ship shape and split them apart and add in a little utility.

DSchau:
@pieh want to take another look at this today?

PR #10156, opened 2018-11-27
Made the 'create a sitemap' link active rather than the stub link colour

AnushPatel:
Got rid of the * as the link now isn't a stub


AnushPatel:
Not very good at describing as shown by the title :man_facepalming:

Issue #10087, opened 2018-11-22
Update V1 to V2 - Error: Cannot find module 'node-sass'


MaralS:
## Description

When I try to to "yarn develop", I'm getting this error.
```ERROR  Failed to compile with 14 errors                                                                    13:17:52

 error  in ./src/components/molecules/navigation/navigation.scss

Module build failed (from ./node_modules/sass-loader/lib/loader.js):
Error: Cannot find module 'node-sass'
    at Function.Module._resolveFilename (module.js:547:15)
    at Function.Module._load (module.js:474:25)
    at Module.require (module.js:596:17)
    at require (D:\Maral Sabbagh\Documents\Sites\maralsabbagh\node_modules\v8-compile-cache\v8-compile-cache.js:159:
20)
    at Object.sassLoader (D:\Maral Sabbagh\Documents\Sites\maralsabbagh\node_modules\sass-loader\lib\loader.js:46:72
)

 @ ./src/components/molecules/navigation/navigation.scss 2:14-506 21:1-42:3 22:19-511
 @ ./src/components/molecules/navigation/navigation.js
 @ ./src/components/molecules/navigation/index.js
 @ ./src/components/molecules/index.js
 @ ./src/pages/contact.js
 @ ./.cache/sync-requires.js
 @ ./.cache/app.js
 @ multi (webpack)-hot-middleware/client.js?path=/__webpack_hmr&reload=true&overlay=false D://Maral Sabbagh//Documents//Sites//maralsabbagh//.cache//app``` 

I tried to upgrade every gatsby dependencies, I followed the guideline. I don't know why I have this error.
Every help will be appreciated.

pieh:
Did you install `node-sass` - it was moved to peer dependency - https://www.gatsbyjs.org/packages/gatsby-plugin-sass/#v200

MaralS:
@pieh I installed it without the --save-dev flag, that's why I had numerous issues. I installed as a dev dependency and it works now thanks :)

PR #10159, opened 2018-11-27
feat(gatsby-plugin-sass): Support Dart SASS

sibiraj-s:
Feature: Support Dart SASS.

Hi,

Gatsby uses `sass-loader` which supports both `node-sass` and `dart-sass`. This PR enables users to use sass of his choice.

**Motivation**

In the current implementation `gatsby-plugin-sass` sends options directly to `sass-loader` so the following usage should work 

```js
 options: {
   implementation: require("sass")
  }
``` 

but it throws some error, I am not sure of it yet. 

<details><summary><b>Error Message</b> </summary>

```bash
error Maximum call stack size exceeded


  RangeError: Maximum call stack size exceeded

  - lodash.js:11714 isLength
    [gatsby-hfc-website]/[lodash]/lodash.js:11714:22

  - lodash.js:11334 isArrayLike
    [gatsby-hfc-website]/[lodash]/lodash.js:11334:31

  - lodash.js:13308 keys
    [gatsby-hfc-website]/[lodash]/lodash.js:13308:14

  - lodash.js:4906
    [gatsby-hfc-website]/[lodash]/lodash.js:4906:21

  - lodash.js:2996 baseForOwn
    [gatsby-hfc-website]/[lodash]/lodash.js:2996:24

  - lodash.js:4880
    [gatsby-hfc-website]/[lodash]/lodash.js:4880:18

  - lodash.js:2846 baseEvery
    [gatsby-hfc-website]/[lodash]/lodash.js:2846:7

  - lodash.js:9133 Function.every
    [gatsby-hfc-website]/[lodash]/lodash.js:9133:14

  - data-tree-utils.js:32 isEmptyObjectOrArray
    [gatsby-hfc-website]/[gatsby]/dist/schema/data-tree-utils.js:32:14

  - data-tree-utils.js:36 _.every
    [gatsby-hfc-website]/[gatsby]/dist/schema/data-tree-utils.js:36:16

  - lodash.js:2847
    [gatsby-hfc-website]/[lodash]/lodash.js:2847:20

  - lodash.js:4911
    [gatsby-hfc-website]/[lodash]/lodash.js:4911:15

  - lodash.js:2996 baseForOwn
    [gatsby-hfc-website]/[lodash]/lodash.js:2996:24

  - lodash.js:4880
    [gatsby-hfc-website]/[lodash]/lodash.js:4880:18

  - lodash.js:2846 baseEvery
    [gatsby-hfc-website]/[lodash]/lodash.js:2846:7

  - lodash.js:9133 Function.every
    [gatsby-hfc-website]/[lodash]/lodash.js:9133:14


error UNHANDLED REJECTION


  RangeError: Maximum call stack size exceeded

  - lodash.js:11714 isLength
    [gatsby-hfc-website]/[lodash]/lodash.js:11714:22

  - lodash.js:11334 isArrayLike
    [gatsby-hfc-website]/[lodash]/lodash.js:11334:31

  - lodash.js:13308 keys
    [gatsby-hfc-website]/[lodash]/lodash.js:13308:14

  - lodash.js:4906
    [gatsby-hfc-website]/[lodash]/lodash.js:4906:21

  - lodash.js:2996 baseForOwn
    [gatsby-hfc-website]/[lodash]/lodash.js:2996:24

  - lodash.js:4880
    [gatsby-hfc-website]/[lodash]/lodash.js:4880:18

  - lodash.js:2846 baseEvery
    [gatsby-hfc-website]/[lodash]/lodash.js:2846:7

  - lodash.js:9133 Function.every
    [gatsby-hfc-website]/[lodash]/lodash.js:9133:14

  - data-tree-utils.js:32 isEmptyObjectOrArray
    [gatsby-hfc-website]/[gatsby]/dist/schema/data-tree-utils.js:32:14

  - data-tree-utils.js:36 _.every
    [gatsby-hfc-website]/[gatsby]/dist/schema/data-tree-utils.js:36:16

  - lodash.js:2847
    [gatsby-hfc-website]/[lodash]/lodash.js:2847:20

  - lodash.js:4911
    [gatsby-hfc-website]/[lodash]/lodash.js:4911:15

  - lodash.js:2996 baseForOwn
    [gatsby-hfc-website]/[lodash]/lodash.js:2996:24

  - lodash.js:4880
    [gatsby-hfc-website]/[lodash]/lodash.js:4880:18

  - lodash.js:2846 baseEvery
    [gatsby-hfc-website]/[lodash]/lodash.js:2846:7

  - lodash.js:9133 Function.every
    [gatsby-hfc-website]/[lodash]/lodash.js:9133:14
```
</details>

<br />
But the following works,

```js
const sass = require('sass')

options: {
   implementation: {
     info: sass.info,
     render: sass.render
    }
}
``` 

Hence the PR to simplify the process.

_NOTE: `node-sass` is removed from peerDependencies to let user to use sass of their choice_

### DIFF
--- a/packages/gatsby-plugin-sass/src/gatsby-node.js
+++ b/packages/gatsby-plugin-sass/src/gatsby-node.js
@@ -2,11 +2,15 @@ import resolve from "./resolve"
 
 exports.onCreateWebpackConfig = (
   { actions, stage, rules, plugins, loaders },
-  { cssLoaderOptions = {}, postCssPlugins, ...sassOptions }
+  { cssLoaderOptions = {}, postCssPlugins, implementation, ...sassOptions }
 ) => {
   const { setWebpackConfig } = actions
   const PRODUCTION = stage !== `develop`
   const isSSR = stage.includes(`html`)
+
+  if (implementation && implementation === `dart-sass`) {
+    sassOptions.implementation = require(`sass`)
+  }
 
   const sassLoader = {
     loader: resolve(`sass-loader`),
--- a/packages/gatsby-plugin-sass/README.md
+++ b/packages/gatsby-plugin-sass/README.md
@@ -8,15 +8,33 @@ Provides drop-in support for SASS/SCSS stylesheets
 
 ## How to use
 
-1.  Include the plugin in your `gatsby-config.js` file.
-2.  Write your stylesheets in SASS/SCSS and require or import them as normal.
+1. Include the plugin in your `gatsby-config.js` file.
+2. Write your stylesheets in SASS/SCSS and require or import them as normal.
 
 ```javascript
 // in gatsby-config.js
 plugins: [`gatsby-plugin-sass`]
 ```
 
-If you need to pass options to Sass use the plugins options, see [node-sass](https://github.com/sass/node-sass)
+By Default `node-sass` is used. To use `dart-sass`.
+
+```bash
+npm i -D sass
+```
+
+```javascript
+// in gatsby-config.js
+plugins: [
+  {
+    resolve: `gatsby-plugin-sass`,
+    options: {
+      implementation: "dart-sass",
+    },
+  },
+]
+```
+
+If you need to pass options to Sass use the plugins options, see [node-sass](https://github.com/sass/node-sass)/[dart-sass](https://github.com/sass/dart-sass) docs
 for all available options.
 
 ```javascript
@@ -26,6 +44,7 @@ plugins: [
     resolve: `gatsby-plugin-sass`,
     options: {
       includePaths: ["absolute/path/a", "absolute/path/b"],
+      ...
     },
   },
 ]
--- a/packages/gatsby-plugin-sass/package.json
+++ b/packages/gatsby-plugin-sass/package.json
@@ -26,8 +26,7 @@
   "license": "MIT",
   "main": "index.js",
   "peerDependencies": {
-    "gatsby": ">2.0.0-alpha",
-    "node-sass": "^4.9.0"
+    "gatsby": "^2.0.0"
   },
   "readme": "README.md",
   "repository": "https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-sass",


sibiraj-s:
Also, I updated the peerDependency `gatsby` from alpha to stable version

pieh:
about `Maximum call stack size exceeded` error: this will happen when we try to construct graphql schema (gatsby-config is used there too) and there are some cyclical references. It might be worth exploring ways to short circuit traversing objects if we hit loop there.

What are pros/cons of using `dart-sass` - maybe we should just switch to it by default?

sibiraj-s:
When it comes to feature they doesn't differ much. It is just a JS implementation of `sass`.

http://sass.logdown.com/posts/1022316-announcing-dart-sass

painless maintenance too. In my case I often switch node versions in local environment with nvm. Every time when I do that I have to rebuild `node-sass` with current node. with `dart-sass` there is no need to do such.

sibiraj-s:
> about Maximum call stack size exceeded error: this will happen when we try to construct graphql schema (gatsby-config is used there too) and there are some cyclical references. It might be worth exploring ways to short circuit traversing objects if we hit loop there.

Okay. Can you Point where the schema is created.

packages/gatsby-plugin-sass/src/gatsby-node.js
```diff
@@ -2,12 +2,16 @@ import resolve from "./resolve"
...
   { actions, stage, rules, plugins, loaders },
-  { cssLoaderOptions = {}, postCssPlugins, ...sassOptions }
+  { cssLoaderOptions = {}, postCssPlugins, implementation, ...sassOptions }
 ) => {
   const { setWebpackConfig } = actions
   const PRODUCTION = stage !== `develop`
   const isSSR = stage.includes(`html`)
 
+  if (implementation && implementation === `dart-sass`) {
+    sassOptions.implementation = require(`sass`)
```
pieh: We should do some sanity checks here - if user didn't install `sass` package, we probably should display warning - something like this :
```suggestion
    try {
      sassOptions.implementation = require(`sass`)
    } catch (e) {
      reporter.warn(`"sass" package isn't installed. Reverting to "node-sass".`)
    }
```
(`reporter` is passed as field in first arg, so you would need to destructure `reporter` in line 4)

pieh:
I'm not sure about removing `node-sass` from `peerDependencies`. I think we should still keep it, as it will be used 99% of the time.

packages/gatsby-plugin-sass/src/gatsby-node.js
sibiraj-s: will update now. I added that early but node.js by default will throw module not found exception and hence i ignored it.

sibiraj-s:
> I'm not sure about removing node-sass from peerDependencies.

This is how sass-loader is implemented.

When the module is not installed `sass-loader` will throw some meaningful error. Keeping that will just give `warning that node-sass is not installed` for the users who uses dart-sass too.

packages/gatsby-plugin-sass/src/gatsby-node.js
sibiraj-s: > Reverting to "node-sass". 

If we say so then, `node-sass` should be in dependency. 
Also, Instead of warning. We should throw errors. Since No package is installed build will break

pieh:
> Okay. Can you Point where the schema is created.

This isn't exactly graphql schema creation (yet) - it's preparation for it. It get caught in infinite loop in this file: https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/schema/data-tree-utils.js where we traverse objects to create "pristine example" of node shape (which later on will be used to actually construct schema). Problem is our traversing object doesn't handle cyclical references like this:

```js
const obj = {
  test: 5
}

obj.cyclical = obj

// when traversing object we will get stuck on visiting nested `cyclical` fields 
// (obj.cyclical.cyclical.cyclical...) eventually resulting in maximum call stack size 
// exceeded error
```

sibiraj-s:
is that a known error @pieh

sibiraj-s:
@pieh  I have updated the pr. once #10177 is merged. everything should just work fine. Just the documentation is updated.

Also, node-sass is removed from peerDependency, this is based on how sass-loader works https://github.com/webpack-contrib/sass-loader. Since `sass-loader` is a direct dependency of the plugin. The errors will be handled by sass-loader itself

pieh:
> Also, node-sass is removed from peerDependency, this is based on how sass-loader works [webpack-contrib/sass-loader](https://github.com/webpack-contrib/sass-loader). Since `sass-loader` is a direct dependency of the plugin. The errors will be handled by sass-loader itself

But `sass-loader` doesn't install `node-sass` (see https://github.com/gatsbyjs/gatsby/issues/10087 what it looks like when `node-sass` is not installed). And also this is a way to note version that is supported by plugin.

sibiraj-s:
thats the point in my mind when I removed it in first place. as mentioned in `sass-loader`

> it requires you to install either Node Sass or Dart Sass on your own

I could update the same in our docs too or else I could add peer back. What should I do?

----

Also, I forgot to mention that there is an open PR for setting dart-sass by default instead of node-sass in sass-loader, due to

> Dart Sass is the primary implementation of Sass, which means it gets new features before any other implementation.

sibiraj-s:
@pieh I have reverted the node-sass dependency back

unmem:
```javascript
// in gatsby-config.js
plugins: [
  {
    resolve: `gatsby-plugin-sass`,
    options: {
      implementation: require("sass"),  // should be "dart-sass" instead, otherwise error
    },
  },
]
```

PR #10160, opened 2018-11-27
ci: produce jest xml reports only in azure pipelines

pieh:
temporary alternative to #10140 (until `ci-info` supports azure pipelines)

/cc @DSchau


jest.config.js
```diff
@@ -31,7 +31,7 @@ module.exports = {
   moduleNameMapper: {
     "^highlight.js$": `<rootDir>/node_modules/highlight.js/lib/index.js`,
   },
-  collectCoverage: true,
+  collectCoverage: !!process.env.GENERATE_JEST_REPORT,
```
DSchau: Nit, but could we hoist this up to something like

```js
const useCoverage = !!process.env.GENERATE_JEST_REPORT

// use elsewhere
```

pieh: 👍 done (at least I hope I did what you meant ;) )

DSchau: exactly :)

PR #10170, opened 2018-11-27
fix(www): Update "View all" component

LekoArts:
@greglobinski @amberleyromo 
Currently the second "View all posts" links to the homepage instead of the blog


greglobinski:
Thank you @LekoArts I missed that

PR #10177, opened 2018-11-28
fix(gatsby): skip functions when inferring schema

sibiraj-s:
Hi,

This is a fix for issue as discussed here https://github.com/gatsbyjs/gatsby/pull/10159#issuecomment-442045709.

prevent loop if the object is of type function


sibiraj-s:
@pieh Can you check this.

PR #10186, opened 2018-11-28
Specify react-emotion to ^9.0.0

lindenquan:
latest version is 10.0.0, which will throw error

### DIFF
--- a/docs/tutorial/part-four/index.md
+++ b/docs/tutorial/part-four/index.md
@@ -96,7 +96,7 @@ Then install some other needed dependencies at the root of the project. You'll u
 "Kirkham", and you'll try out a CSS-in-JS library, ["Emotion"](https://emotion.sh/):
 
 ```shell
-npm install --save gatsby-plugin-typography typography react-typography typography-theme-kirkham gatsby-plugin-emotion emotion react-emotion emotion-server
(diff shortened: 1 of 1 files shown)


docs/tutorial/part-four/index.md
```diff
@@ -96,7 +96,7 @@ Then install some other needed dependencies at the root of the project. You'll u
 "Kirkham", and you'll try out a CSS-in-JS library, ["Emotion"](https://emotion.sh/):
 
 ```shell
-npm install --save gatsby-plugin-typography typography react-typography typography-theme-kirkham gatsby-plugin-emotion emotion react-emotion emotion-server
+npm install --save gatsby-plugin-typography typography react-typography typography-theme-kirkham gatsby-plugin-emotion emotion react-emotion@^9.0.0 emotion-server
```
DSchau: Should all of these be pined to 9.0.0? i.e. emotion, emotion-server, etc.

I'd think that is safer.

lindenquan: gatsby-plugin-emotion shows warnings saying it only supports emotion v7, v8, v9.
It would be better to change versions for all emotion modules.
Thanks

DSchau: Sorry - I don't really understand what you mean here? That's exactly what I'm saying I think!

lindenquan: I'm not good at expressing. If you want, you can go through https://www.gatsbyjs.org/tutorial/part-four/  , then you will find it won't work.

Issue #10168, opened 2018-11-27
Create pullquote styles for .org site


amberleyromo:
Pulled out of #9847.

Points of consideration:
- multiple styles (quote style, highlight style) -- as you've suggested, @greglobinski 
- implement using [custom components](https://using-remark.gatsbyjs.org/custom-components/)

![pull-quote-v1](https://user-images.githubusercontent.com/3461087/49109205-bfdbee80-f24f-11e8-91b6-55e0c35cefc7.png)

***

![pull-quote-v2](https://user-images.githubusercontent.com/3461087/49109212-c4a0a280-f24f-11e8-9983-7bfe19522614.png)

greglobinski:
New variations of the designs 

1. a quotation mark to indicate it's a quote added

![screenshot from 2018-11-28 23-21-50](https://user-images.githubusercontent.com/32480082/49186278-729d7100-f364-11e8-893e-5f0c3160a149.png)

2. the 'invisible' yellow color removed

![screenshot from 2018-11-28 23-22-04](https://user-images.githubusercontent.com/32480082/49186279-729d7100-f364-11e8-876f-393713598377.png)

@lindaleebumblebee

lindaleebumblebee:
Thank you. Both are great, but now I like the first one the best (with the box and 'stars') :)

lindaleebumblebee:
Love it! Looks great.

amberleyromo:
Closed in #10190

PR #10190, opened 2018-11-28
feat(www): create Pullquote custom component

greglobinski:
Implement https://github.com/gatsbyjs/gatsby/issues/10168 

- crate custom component `www/src/components/shared/pullquote.js`
- add it to the `www/src/templates/template-blog-post.js`

To Do
- [x] add styles

Note:
the same way as for other custom components (e.g. `HubspotForm`) React in development mode throws a warning

`Warning: validateDOMNesting(...): <blockquote> cannot appear as a descendant of <p>.`

screenshot from 2018-11-28 22-48-01


greglobinski:
@DSchau That's right

```
# Meetups and More

Meetups are also a great way to connect with the Gatsby community. Several community-organized Gatsby meetups have already been popping up across the globe and we’ll be launching a more formal program to support community-organized meetups with free swag and resources (details will be coming soon). In the meantime, if you’re organizing a Gatsby meetup or event and need some support, just [fill out the form here](https://airtable.com/shrpwc99yogJm9sfI) and someone will get back with you to work out the details.

<pullquote>We’ll also be planning more Gatsby Days after the New Year. We did a twitter poll a few weeks ago to gauge interest in locations for Gatsby Days and London gathered over 250 votes!
</pullquote>

And if you’re in the San Francisco bay area, join our [San Francisco meetup](https://www.meetup.com/meetup-group-cHIRVLfX/) to connect with Gatsbyjs community members. These will be more casual gatherings where we talk about front-end web development, the modern web tech stack, and Gatsby. We'll have our first meetup after the new year (in January 2019) so stay tuned!
```

greglobinski:
@DSchau I'm sorry not exactly, I'm using a custom component name `<pullquote>`
```<pullquote>We’ll also be planning more Gatsby Days after the New Year. We did a twitter poll a few weeks ago to gauge interest in locations for Gatsby Days and London gathered over 250 votes!
</pullquote>```

**Comment by greglobinski** [comment]
@DSchau Shoud I rename the custom component's name to `blockquote`?

**Comment by DSchau** [comment]
@greglobinski no! pullquote is better, I misread the component name!

File: www/src/templates/template-blog-post.js [hunk 1]
```diff
@@ -12,17 +12,35 @@ import Container from "../components/container"
...
   createElement: React.createElement,
   components: {
     "hubspot-form": HubspotForm,
     "date-chart": Chart,
+    pullquote: Pullquote,
   },
 }).Compiler
 
 class BlogPostTemplate extends React.Component {
+  componentDidMount = () => {
```
DSchau: This seems not quite ideal. Can't you just re-use `nth-child` here?

Here's an [example](https://codesandbox.io/s/p32pp5xy4j)

greglobinski: I would love to avoid adding classes but for me (now) it's not possible.

#### 1. 
In the example there are no other elements in the container than  `<p><span>text</span></p>`.
```
<div>
  <p><span>text</span></p>
  <p><span>text</span></p>
  <p><span>text</span></p>
</div>
```
in the Gatsby blog we have inside the container also other elements, not only these with `Pullquotes`

```
<div>
  <p>text</p>
  <p>text</p>
  <p><blockquote>text</blockquote></p>
  <p>text</p>
  <p><blockquote>text</blockquote></p>
  <p>text</p>
</div>
```
And we have no idea where the items `<p><blockquote>text</blockquote></p>` will be placed, on which position, so we can't base styling on `p:nth-child(${index + 1}n)`

#### 2. 
We also can't use `:nth-of-type`, which I had in mind when I proposed this 'style variations' feature. Because every of our `<blockquote>` element is always on the first position, it's always the first child of type inside the containing `<p>` tag.

#### 3. 
Assume that we can remove the `<p>` tags containing our `<blockquote>` elements. I would love to have it that way, but I have no idea how to make it (at least for now), probably some logic in `gatsby-transformer-remark` could make it. But even if we would have cleaner markdown like these

```
<div>
  <p>text</p>
  <p>text</p>
  <blockquote>text</blockquote>
  <p>text</p>
  <blockquote>text</blockquote>
  <p>text</p>
</div>
```  
we can't omit the class names, because we cannot touch existing `<blockquote>` elements created with markdown blockquote formatting  `> text...`. So we have to differentiate the `Pullquote` blockquotes from the  generic ones.   

#### 4.
I'm sorry, I'm not sure what do you mean by that 
> If worse came to worse, I almost think a local variable tracking component mounts (and assigning a class) would perhaps be improved, as well.

Could you please explicate it a little bit for me?

DSchau: > in the Gatsby blog we have inside the container also other elements, not only these with Pullquotes

I'm not sure this matters! I've updated my example to show a similar structure. https://codesandbox.io/s/p32pp5xy4j

> Could you please explicate it a little bit for me?

I mean something like this. If the idea is that we want to track instances/occurrences, we can use a local variable (a side-effect) and track it like so in [this example](https://codesandbox.io/s/vmnr47mx57). Note that this is _global_ so it tracks number of invocations of the component. So it wouldn't necessarily be the same every page, e.g. if you went to a page with one pullquote, navigating to a second page would start on the 2nd style.

This _could_ be worse, but I want to give you some options, because I think my "spidey sense" is going off a little bit with the current approach! Totally fine with it as is, just want to make sure we at least vet some other options!

greglobinski: > I'm not sure this matters! I've

Actually the updated example is less clear than the original was :). But  they seam to  work only because you generate the list of paragraph, so there are repeating patterns. 

I've simplified the example https://codesandbox.io/s/1op0l5535q

screenshot from 2018-11-30 06-40-25


As you can see, the first and the second 'Blockquotes' are the same. That's because `p:nth-child` counts every paragraph not only these with the blocquotes inside and as it's presented in the example there are edge cases when two or more following Pullquotes will have the same styles.

>  If the idea is that we want to track instances/occurrences, we can use a local variable (a side-effect)

I like it :+1: 

> navigating to a second page would start on the 2nd style. 

It's not a problem. What matters is that every following Pullquote will have different style, doesn't matter how many generic `<p>` tags will be between it and the previous one. I can refactor the code this way 

> I want to give you some options, [...] just want to make sure we at least vet some other options!

I like it too :100:

greglobinski:
Update

- refactor variant class names logic according to @DSchau suggestion with 'a local variable (a side-effect)'

www/src/templates/template-blog-post.js
greglobinski: @DSchau I refactored code, it's look much cleaner now. Thank you  Dustin! :1st_place_medal:

PR #10196, opened 2018-11-29
fix(gatsby-transformer-remark): render html attributes in html excerpt

### LINKED ISSUE #10194: anchor tags not clickable in excerpts
## Description

Anchor tags are not clickable in markdown excerpts formatted as html. More specifically, the `href` attributes aren't making their way over

This

### Steps to reproduce

make a graphql request like this:
```graphql
{
  allMarkdownRemark {
    edges {
      node {
        excerpt(format: HTML)
      }
    }
  }
}
```
Look at the results and notice that the href attribute of anchor tags is missing

### Expected result

I expect anchor tags to have hrefs

This is a bug in the PR I just did: #9716. I've got a pretty strong suspicion that the bug is [here](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-transformer-remark/src/hast-processing.js#L6)

I strongly suspect there is an attribute that we are failing to duplicate. Shouldn't be too hard to fix

nadrane:
fix #10194

### DIFF
--- a/packages/gatsby-transformer-remark/src/__tests__/extend-node.js
+++ b/packages/gatsby-transformer-remark/src/__tests__/extend-node.js
@@ -245,7 +245,7 @@ title: "my little pony"
 date: "2017-09-18T23:19:51.246Z"
 ---
 
-Where oh [*where*](nick.com) **_is_** ![pony.png](that pony)?`,
+Where oh [*where*](nick.com) **_is_** ![that pony](pony.png)?`,
     `excerpt(format: HTML)
     frontmatter {
         title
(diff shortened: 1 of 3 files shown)


packages/gatsby-transformer-remark/src/__tests__/extend-node.js
```diff
@@ -245,7 +245,7 @@ title: "my little pony"
 date: "2017-09-18T23:19:51.246Z"
 ---
 
-Where oh [*where*](nick.com) **_is_** that pony?`,
+Where oh [*where*](nick.com) **_is_** ![pony.png](that pony)?`,
```
LekoArts: Did you do the switch of filename and alt tag on purpose (pony)?

nadrane: hah, I did not. Good catch!

i'll fix it right away

nadrane: fixed it :)

DSchau:
Thanks once more with this PR! Also check out gatsbyjs/gatsby#10202 if you haven't already 😄

PR #10201, opened 2018-11-29
Correct link to source plugins

Simply007:
The link to the source plugins led to the repository file system that was used for storing source plugins. Currently, source plugins have its own repositories so I defined the link to lead to the plugin library.


Simply007:
My fault - this page is meant to be to build the documentation:
* https://www.gatsbyjs.org/docs/headless-cms/

PR #10204, opened 2018-11-29
feat(examples/using-ghost)

d-cs:
* created minimal working example of using `gatsby-source-ghost`

Inspired by the `using-wordpress` example and the fact that Ghost can be used as a data source I decided to give it a whirl. Please don't be alarmed by the commited field `clientSecret` in `gatsby-config.js`, the client secrets are actually public in Ghost front ends and API security is managed by trusted domains.

The aim of this example was to demonstrate how to use the API to render typical Ghost blog pages, implementing queries such as "Get all posts for a given tag", etc.

### DIFF
--- a/examples/using-ghost/.eslintignore
+++ b/examples/using-ghost/.eslintignore
@@ -1,0 +1,1 @@
+node_modules/
--- a/examples/using-ghost/.gitignore
+++ b/examples/using-ghost/.gitignore
@@ -1,0 +1,10 @@
+# Build Path of Test Fixtures
+.gatsby-context.js
+.DS_Store
+public/
+build/
+.cache/
+.netlify
+
+# IDE specific
+.vscode/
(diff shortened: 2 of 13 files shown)


LekoArts:
Hi!
The example is looking great 👍 
However, I don't think it's a good idea to add the example here as the plugin is not maintained here. We had a few examples that ultimately we had to remove because the plugin wasn't actively maintained anymore.
You could reach out to the folks of Ghost if the want to add an example or you create a Gatsby starter project 😃

d-cs:
@LekoArts thank you! It's my first attempt at contributing to open source. That's a really good point you raise, I'm currently working on a Gatsby starter project now :-) 

I'll close this now.

LekoArts:
@d-cs Great! You can add your starter like that then: https://www.gatsbyjs.org/docs/submit-to-starter-library/

d-cs:
@LekoArts I've created a PR for the starter [here](https://github.com/gatsbyjs/gatsby/pull/10219), referencing it here for trace-ability.

PR #10208, opened 2018-11-29
Add grooovinger.com to showcase

mgrubinger:
I'd like to add my portfolio website to the gatsby showcase.

Thanks for all the great work you're doing! 💪

Thank you.

### DIFF
--- a/docs/sites.yml
+++ b/docs/sites.yml
@@ -3611,3 +3611,14 @@
     - Education
   built_by: Michael Uloth
   built_by_url: "https://www.michaeluloth.com"
+- title: Grooovinger
+  url: https://www.grooovinger.com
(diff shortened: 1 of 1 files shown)


docs/sites.yml
```diff
@@ -3611,3 +3611,15 @@
     - Education
   built_by: Michael Uloth
   built_by_url: "https://www.michaeluloth.com"
+- title: Grooovinger
```
DSchau: Looks like the linter is failing here. I think if you check this out locally, this should work to fix the error:

```bash
npx prettier --write docs/sites.yml
```

mgrubinger: sorry, I missed the new line. Thanks!

PR #10209, opened 2018-11-29
fix(gatsby-remark-prismjs): prevent additional blank line from appearing

DSchau:
This PR fixes a small bug where an extra line character is rendered as the generated HTML contains a structure like `</span>\n</span>`. This new line character causes a blank line to be inserted into the resultant highlighted code.

I've validated that this doesn't seem to cause any regressions with the [kitchen sink](https://gatsby-line-highlighting-kitchen-sink.netlify.com/).

*Changes in this PR*

screen shot 2018-11-29 at 1 59 50 pm

*Published version*

screen shot 2018-11-29 at 2 01 56 pm


packages/gatsby-remark-prismjs/src/__tests__/__snapshots__/highlight-line-range.js.snap
```diff
@@ -1,87 +1,50 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`highlighting a line range highlight-line highlights line 1`] = `
-"<span class=\\"gatsby-highlight-code-line\\">  return \\"hello world\\"
-</span>"
-`;
+exports[`highlighting a line range highlight-line highlights line 1`] = `"<span class=\\"gatsby-highlight-code-line\\">  return \\"hello world\\"</span>"`;
```
DSchau: These diffs are a little messy, but it just removes the hard line breaks from the diffs. This will be handled via HTML start/end tags anyways, so this should be just fine.

packages/gatsby-remark-prismjs/src/highlight-line-range.js
```diff
@@ -47,7 +47,7 @@ const stripComment = line =>
   )
 
 const wrap = line =>
-  [`<span class="gatsby-highlight-code-line">`, `${line}\n`, `</span>`].join(``)
+  [`<span class="gatsby-highlight-code-line">`, `${line}`, `</span>`].join(``)
```
DSchau: The fix. One character 🆗

PR #10210, opened 2018-11-29
Add option to keep metadata in files processed by `gatsby-plugin-sharp`

FineWolf:
## Work Done
- Added an option to keep metadata (ICC profile, EXIF information) from files processed by `gatsby-plugin-sharp`. The old behavior remains as default.
- `gatsby-plugin-sharp` was one of the only plugins who didn't support passing configuration thru `gatsby-config.js`. The only way to enable MozJpeg was via an environment variable. This also has been normalized while maintaining old behavior as default.

## Reasoning
I'm currently rebuilding my website with `gatsby`. While I was working on my portfolio section, I noticed that many of my pictures were rendering differently in Firefox vs Chrome due to the lack of ICC profile attached to the image (untagged images in most browsers render using the Monitor's RGB profile; Chrome seems to assume the pictures are using sRGB IEC61966-2.1). Also, I tend to like to keep my Copyright EXIF information on my images.

This PR (with no breaking changes to old behavior) allows us to specify if we want to strip metadata in the images or not.

### DIFF
--- a/packages/gatsby-plugin-sharp/README.md
+++ b/packages/gatsby-plugin-sharp/README.md
@@ -22,7 +22,15 @@ images. By default it uses a quality setting of [50-75].
 
 ```javascript
 // In your gatsby-config.js
-plugins: [`gatsby-plugin-sharp`]
+plugins: [
+  {
+    resolve: `gatsby-plugin-sharp`,
+    options: {
+      useMozJpeg: process.env.GATSBY_JPEG_ENCODER === `MOZJPEG`,
+      stripMetadata: true,
+    },
+  },
+]
 ```
 
 ## Methods
@@ -242,11 +250,20 @@ You can opt-in to use [MozJPEG][16] for jpeg-encoding. MozJPEG provides even
 better image compression than the default encoder used in `gatsby-plugin-sharp`.
 However, when using MozJPEG the build time of your Gatsby project will increase
 significantly.
-To enable MozJPEG set the [environment variable](/docs/environment-variables/#environment-variables):
+To enable MozJPEG, you can set the `useMozJpeg` plugin option to `true` in
+`gatsby-config.js` or set
+the [environment variable](/docs/environment-variables/#environment-variables):
 
 ```shell
 GATSBY_JPEG_ENCODER=MOZJPEG
 ```
+
+### EXIF and ICC metadata
+
+By default, `gatsby-plugin-sharp` strips all EXIF, ICC and other metadata
+present in your source file. If you wish to preserve this metadata, you can
+set the `stripMetadata` plugin option to `false` in
+`gatsby-config.js`.
 
 [1]: https://alistapart.com/article/finessing-fecolormatrix
 [2]: http://blog.72lions.com/blog/2015/7/7/duotone-in-js
--- a/packages/gatsby-plugin-sharp/README.md
+++ b/packages/gatsby-plugin-sharp/README.md
@@ -265,14 +265,14 @@ GATSBY_JPEG_ENCODER=MOZJPEG
 ### EXIF and ICC metadata
 
 By default, `gatsby-plugin-sharp` strips all EXIF, ICC and other metadata
-present in your source file. This is the recommended default for most
-situations as it leads to smaller file sizes.
+present in your source file. This is the recommended default as it leads to
+smaller file sizes.
 
 However, in situations where you wish to preserve EXIF metadata or ICC profiles
-(for example if you are building a photography portfolio and which to conserve
-the color profile of the photos you've exported from Adobe Lightroom or Phase
-One's Capture One), you can set the `stripMetadata` plugin option to `false` in
-`gatsby-config.js`.
+(example: you are building a photography portfolio and wish to conserve
+the color profile or the copyright information of the photos you've exported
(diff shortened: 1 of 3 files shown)


packages/gatsby-plugin-sharp/src/index.js
```diff
@@ -124,7 +138,14 @@ const processFile = (file, jobs, cb, reporter) => {
 
   let pipeline
   try {
-    pipeline = sharp(file).rotate()
+    pipeline = sharp(file)
+
+    // Keep Metadata
+    if (pluginOptions.stripMetadata === false) {
```
DSchau:
```suggestion
    if (!pluginOptions.stripMetadata) {
```

packages/gatsby-plugin-sharp/src/index.js
```diff
@@ -41,6 +41,22 @@ exports.setBoundActionCreators = actions => {
...
 }
 
+/// Plugin options are loaded onPreInit in gatsby-node
+let pluginOptions = {
+  useMozJpeg: process.env.GATSBY_JPEG_ENCODER === `MOZJPEG`,
+  stripMetadata: true,
+}
+exports.setPluginOptions = options => {
+  // Typechecks
+  if (typeof options.useMozJpeg === `boolean`) {
```
DSchau:
```js
const pluginDefaults = {
  useMozJpeg: process.env.GATSBY_JPEG_ENCODER === `MOZJPEG`,
  stripMetadata: true,
}

exports.setPluginOptions = opts => {
  const options = Object.assign({}, pluginOptions, opts);
}
```

packages/gatsby-plugin-sharp/README.md
```diff
@@ -242,12 +250,21 @@ You can opt-in to use [MozJPEG][16] for jpeg-encoding. MozJPEG provides even
...
+the [environment variable](/docs/environment-variables/#environment-variables):
 
 ```shell
 GATSBY_JPEG_ENCODER=MOZJPEG
 ```
 
+### EXIF and ICC metadata
+
+By default, `gatsby-plugin-sharp` strips all EXIF, ICC and other metadata
+present in your source file. If you wish to preserve this metadata, you can
```
DSchau: Is it helpful to document _why_ you'd want to keep this here?

packages/gatsby-plugin-sharp/README.md
```diff
@@ -22,7 +22,15 @@ images. By default it uses a quality setting of [50-75].
 
 ```javascript
 // In your gatsby-config.js
-plugins: [`gatsby-plugin-sharp`]
+plugins: [
+  {
+    resolve: `gatsby-plugin-sharp`,
+    options: {
+      useMozJpeg: process.env.GATSBY_JPEG_ENCODER === `MOZJPEG`,
```
DSchau: I think you're documenting defaults here, but this seems kinda weird. It seems to be sorta contrary to the docs, because here we're using a plugin option _and_ an environment variable, when the README specifies we only need one or the other.

FineWolf: I was debating whether or not to put a pure Boolean or not.... I feel that if I add a pure Boolean in the documentation, the code should be updated so that `GATSBY_JPEG_ENCODER` should always override the options supplied.

I thought that documenting it this way led to less confusion in terms of backwards compatibility and functionality.

FineWolf: I was debating whether or not to put a pure Boolean or not.... I feel that if I add a pure Boolean in the documentation, the code should be updated so that `GATSBY_JPEG_ENCODER` should always override the options supplied.

I thought that documenting it this way led to less confusion in terms of backwards compatibility and functionality.

packages/gatsby-plugin-sharp/src/index.js
DSchau: You might be using the wrong language if you want strict type-checking :)

I really wouldn't worry too much about it. I do think you raise a good point--I've wanted to implement an API using something like `yup` or `joi` for quite a while, but just haven't gotten around to doing so yet. That being said, if we think of the "false positive" case here (e.g. a string is passed, or undefined, etc.) I think we're pretty isolated from any errors since we're just checking for truthy/falsy rather than using these values in a meaningful way--e.g. passing them to something that needs more type safety.

packages/gatsby-plugin-sharp/README.md
FineWolf: I guess it doesn't hurt to add that information now that I think about it.

FineWolf:
Changes have been made as requested.

FineWolf:
Changes have been made as requested. (cc. @DSchau )

FineWolf:
Made a quick change to ensure that our setting is passed on to Pngquant when running imagemin. `imagemin-mozjpeg` and `imagemin-webp` do not strip metadata.

FineWolf:
@DSchau any news on this PR? Right now I'm relying on a fork for my builds.

DSchau:
@FineWolf thanks for the ping. I'll give this another once over, and probably get it merged in. Thanks for your patience!

DSchau:
Successfully published:
 - gatsby-plugin-sharp@2.0.15

PR #10212, opened 2018-11-29
Sync db autosaves

Moocar:
This PR orchestrates the saving of db state for both loki and redux into one place. Previously, loki would use its `autosave` feature and save to disk every 1000ms. Whereas now, it uses the same style as `gatsby/redux` where a debounced save is triggered for each new event. 

Note that both database save asynchronously (redux's write to disk is async) so this isn't a perfect "Stop the world" synchronized save. However it's better than it was before.

### DIFF
--- a/packages/gatsby/src/bootstrap/index.js
+++ b/packages/gatsby/src/bootstrap/index.js
@@ -21,6 +21,7 @@ const getConfigFile = require(`./get-config-file`)
 const tracer = require(`opentracing`).globalTracer()
 const preferDefault = require(`./prefer-default`)
 const nodeTracking = require(`../db/node-tracking`)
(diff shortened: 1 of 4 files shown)


Moocar:
@DSchau Wow, that was some messed up history. I went for the nuclear option since it's not much code. Should be good to review now.

pieh:
> Note that both database save asynchronously (redux's write to disk is async) so this isn't a perfect "Stop the world" synchronized save. However it's better than it was before.

Yeah, let's get this in - later on we can do some save queue instead of debouncing or something like that, so we create task that will wait for all "partial" state saves before resolving. But this is good for now - small incremental changes are much nicer to review and reason with

PR #10215, opened 2018-11-30
feat(gatsby-plugin-google-analytics): add TS typings for OutboundLink component

tdharmon:
This PR adds typings to `gatsby-plugin-google-analytics`.


packages/gatsby-plugin-google-analytics/index.d.ts
```diff
@@ -0,0 +1,8 @@
+  import * as React from "react";
+
+  interface OutboundLinkProps {
+    onClick?: (event: React.MouseEvent<HTMLAnchorElement>) => void;
```
tdharmon: I omitted `href` and `target` from these props because I'm extending the default `HTMLAnchorElement` props, which include those

PR #10222, opened 2018-11-30
Update Testing Components Installation Section

jsjoeio:
This PR changes the prose related to setting up your `setup-test-env.js` file. Because it's assumed the user followed the unit-testing section, they should have a `jest.config.js` file. If they do, they need to add the `setupTestFrameworkScriptFile` to the jest config, rather than the `package.json` otherwise, their `setup-test-env.js` will **not** run correctly.

I've added the instructions for `jest.config.js` but also kept the original instructions related to adding it to the `package.json`

### DIFF
--- a/docs/docs/testing-react-components.md
+++ b/docs/docs/testing-react-components.md
@@ -33,7 +33,6 @@ Lastly you need to tell Jest where to find this file. Open your `jest.config.js`
 
 ```js:title=jest.config.js
 module.exports = {
-  ...
   'setupTestFrameworkScriptFile': '<rootDir>/setup-test-env.js'
 }
 ```
--- a/docs/docs/testing-react-components.md
+++ b/docs/docs/testing-react-components.md
@@ -38,13 +38,6 @@ module.exports = {
 }
 ```
 
-*If you don't have a `jest.config.js` file, you need to instead add this to your `package.json` like so:*
-```json:title=package.json
-"jest": {
-  "setupTestFrameworkScriptFile": "<rootDir>/setup-test-env.js"
-}
-```
-
 ## Usage
 
 Let's create a little example test using the newly added library. If you haven't done already read the [unit testing guide](/docs/unit-testing) — essentially you'll use `react-testing-library` instead of `react-test-renderer` now. There are a lot of options when it comes to selectors, this example chooses `getByTestId` here. It also utilizes `toHaveTextContent` from `jest-dom`:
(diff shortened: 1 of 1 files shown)


docs/docs/testing-react-components.md
```diff
@@ -29,8 +29,16 @@ import "react-testing-library/cleanup-after-each"
 
 This file gets run automatically by Jest before every test and therefore you don't need to add the imports to every single test file.
 
-Lastly you need to tell Jest where to find this file. Open your `package.json` and add this entry to your `"jest"` section:
+Lastly you need to tell Jest where to find this file. Assuming you followed the [Unit testing](/docs/unit-testing), open your `jest.config.js` and add this entry to the bottom after 'setupFiles':
```
LekoArts: I'd cut the "Assuming" part as this note is already at the top of the guide.

```suggestion
Lastly you need to tell Jest where to find this file. Open your `jest.config.js` and add this entry to the bottom after 'setupFiles':
```

docs/docs/testing-react-components.md
```diff
@@ -29,8 +29,16 @@ import "react-testing-library/cleanup-after-each"
...
 
+```js:title=jest.config.js
+module.exports = {
+  ...
```
LekoArts: I'd cut the `...` so that people don't actually copy/paste that

```suggestion
```

docs/docs/testing-react-components.md
```diff
@@ -29,8 +29,16 @@ import "react-testing-library/cleanup-after-each"
...
+  'setupTestFrameworkScriptFile': '<rootDir>/setup-test-env.js'
+}
+```
+
+*If you don't have a `jest.config.js` file, you need to instead add this to your `package.json` like so:*
```
LekoArts: In my opinion you can also cut the old part. It's not relevant anymore and it could confuse people

DSchau: Agreed! We should always show the simplest/clearest path. Showing multiple just leads to confusion, e.g. why would I use a file vs. package.json?

jsjoeio: Good point! Let me cut that.

jsjoeio:
@LekoArts @DSchau - I made the changes you both suggested! Should look good now.

LekoArts:
Yeah, only linting is failing right now. Please lint the file in your local copy with Prettier, e.g. with:

```bash
npx prettier --write docs/docs/testing-react-components.md
```

or

```bash
npm install -g prettier
```

and then

```bash
prettier --write docs/docs/testing-react-components.md
```

PR #10223, opened 2018-11-30
Add LXDX.co to Showcase

coreyward:
Adding [lxdx.co](https://www.lxdx.co/) to the Gatsby Site Showcase.


LekoArts:
Seems like you need to pull the latest changes. Other than that seems good.

PR #10226, opened 2018-11-30
feat(gatsby-plugin-emotion): update plugin for emotion v10

wKovacs64:
# :tada: Emotion v10 is here! :tada: 

- [X] Update dependencies
- [X] Update critical style handling
- [X] Add Babel config for `jsx` pragma
- [X] Update plugin install instructions
- [X] Update Gatsby tutorial (part 4)
- [X] Update the "using emotion" example

### DIFF
--- a/examples/using-emotion/src/pages/index.js
+++ b/examples/using-emotion/src/pages/index.js
@@ -59,9 +59,9 @@ const IndexPage = () => (
       </h1>
       <p
         // Styling "inline" via css prop and object styles
-        css={css({
+        css={{
           color: `#bd93f9`,
-        })}
+        }}
       >
         <a
           href="https://www.gatsbyjs.org/packages/gatsby-plugin-emotion/"
--- a/examples/using-emotion/src/pages/index.js
+++ b/examples/using-emotion/src/pages/index.js
@@ -1,31 +1,22 @@
 import React, { Fragment } from "react"
 import Helmet from "react-helmet"
-import styled, { css, injectGlobal } from "react-emotion"
-// You can import everything from "react-emotion"
+import styled from "@emotion/styled"
+import { css, Global } from "@emotion/core"
 
 // Emotion supports different styling options, all of which are supported by gatsby-plugin-emotion out of the box
 
-injectGlobal`
+// Create styles for the Global component
+const globalStyles = css`
   * {
     margin: 0;
     padding: 0;
     box-sizing: border-box;
   }
-  html, body {
-    font-family: -apple-system,
-      BlinkMacSystemFont,
-      "Segoe UI",
-      "Roboto",
-      "Roboto Light",
-      "Oxygen",
-      "Ubuntu",
-      "Cantarell",
-      "Fira Sans",
-      "Droid Sans",
-      "Helvetica Neue",
-      sans-serif,
-      "Apple Color Emoji",
-      "Segoe UI Emoji",
+  html,
+  body {
+    font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto",
+      "Roboto Light", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans",
+      "Helvetica Neue", sans-serif, "Apple Color Emoji", "Segoe UI Emoji",
       "Segoe UI Symbol";
   }
 `
@@ -41,7 +32,7 @@ const Wrapper = styled.section`
   width: 100vw;
 `
 
-// Using css with template literal
+// Using css with a template literal
 const title = css`
   font-size: 1.5em;
   color: #ff79c6;
@@ -52,11 +43,6 @@ const title = css`
   }
 `
 
-// Using css with object
-const subtitle = css({
-  color: `#bd93f9`,
-})
-
 const IndexPage = () => (
   <Fragment>
     <Helmet>
@@ -64,16 +50,22 @@ const IndexPage = () => (
       <meta name="description" content="Gatsby example site using Emotion" />
       <meta name="referrer" content="origin" />
     </Helmet>
+    <Global styles={globalStyles} />
     <Wrapper>
-      <h1 className={title}>
+      <h1 css={title}>
         Hello World, this is my first component styled with
         {` `}
         <a href="https://emotion.sh/">emotion</a>!
       </h1>
-      <p className={subtitle}>
+      <p
+        // Styling "inline" via css prop and object styles
+        css={css({
+          color: `#bd93f9`,
+        })}
+      >
         <a
           href="https://www.gatsbyjs.org/packages/gatsby-plugin-emotion/"
-          // Styling “inline” with css prop
+          // Styling “inline” via css prop and a template literal
           css={css`
             color: inherit;
           `}
--- a/packages/gatsby-plugin-emotion/src/gatsby-browser.js
+++ b/packages/gatsby-plugin-emotion/src/gatsby-browser.js
@@ -1,3 +1,3 @@
-import { wrapElement } from "./wrap-element"
+const { wrapElement } = require(`./wrap-element`)
 
 exports.wrapRootElement = ({ element }) => wrapElement(element)
--- a/packages/gatsby-plugin-emotion/src/gatsby-browser.js
+++ b/packages/gatsby-plugin-emotion/src/gatsby-browser.js
@@ -1,11 +1,3 @@
-/* globals window */
-import { hydrate } from "emotion"
+import { wrapElement } from "./wrap-element"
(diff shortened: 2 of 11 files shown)


packages/gatsby-plugin-emotion/src/gatsby-node.js
```diff
@@ -1,11 +1,30 @@
...
+    },
+  })
+}
+
+exports.onCreateWebpackConfig = ({ actions }) => {
+  actions.setWebpackConfig({
+    plugins: [
+      new webpack.ProvidePlugin({
+        Emotion: `@emotion/core`,
+      }),
```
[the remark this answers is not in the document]
wKovacs64: I tried that initially (prior to e04ba5307929a5f09e00e17f27eddbbffa48e707) but couldn't get it to work. See anything wrong here?

https://github.com/gatsbyjs/gatsby/blob/28cbda70bcf8fe36e621677ff3d24f8b7dc2d8cc/packages/gatsby-plugin-emotion/src/gatsby-node.js

Maybe we should just wait for your preset?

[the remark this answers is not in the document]
wKovacs64: How's this look?

https://github.com/gatsbyjs/gatsby/blob/c4f7cc867a1327208cfcab297407d256556e742c/packages/gatsby-plugin-emotion/src/gatsby-node.js

DSchau:
> So, this seems to work for me but it isn't backwards compatible... what's the best way to deal with that

The peerDep handles it right? Requiring v10 of packages as peerDependencies should cover our bases there.

wKovacs64:
> The peerDep handles it right? Requiring v10 of packages as peerDependencies should cover our bases there.

Cool with me if peerDep bumps are good enough, just wasn't sure since bumping this plugin version by a minor or patch would convey non-breaking changes but an emotion v9 consumer's build will explode because the plugin now imports from `@emotion/core`. SSR critical styles are handled differently too, so those would be broken as well (if they got that far by bringing in the new peerDeps). I might be overthinking it, though. 😅

DSchau:
> since bumping this plugin version

Oh yeah. This would be a major version bump for this plugin! Misunderstood you there.

wKovacs64:
> Oh yeah. This would be a major version bump for this plugin! Misunderstood you there.

Cool. I bumped it to `10.0.0` to match emotion, but let me know if you/the team aren't a fan of skipping semver majors like that and I'll drop it to `3.0.0`.

DSchau:
@wKovacs64 I'm so sorry, that's my fault on that one!

If you could, could you revert that commit, e.g.

```bash
git revert f05f5a2
```

and then push that up to this branch. We actually use `lerna` to manage version bumping, so in this example, we'll manually bump this plugin up to the next major version!

wKovacs64:
@DSchau Sure, no problem. Will that handle bumping the `using-emotion` example project, too?

DSchau:
> Will that handle bumping the using-emotion example project, too

Possibly, but we can just leave that for now!

wKovacs64:
Word.

wKovacs64:
> Before we merge, we'll want to edit the description

Will whoever merges do that or do you want me to edit the first commit (currently 28cbda70) and rewrite the PR history?

examples/using-emotion/src/pages/index.js
```diff
@@ -52,28 +43,29 @@ const title = css`
...
-      <h1 className={title}>
+      <h1 css={title}>
         Hello World, this is my first component styled with
         {` `}
         <a href="https://emotion.sh/">emotion</a>!
       </h1>
-      <p className={subtitle}>
+      <p
+        // Styling "inline" via css prop and object styles
+        css={css({
```
mitchellhamilton: You don't need a css call here, you can pass the style object directly to the css prop.

wKovacs64: Removed in 1fc2cee7, thanks.

packages/gatsby-plugin-emotion/src/gatsby-browser.js
```diff
@@ -1,11 +1,3 @@
-/* globals window */
-import { hydrate } from "emotion"
+const { wrapElement } = require(`./wrap-element`)
 
-exports.onClientEntry = () => {
```
DSchau: Just about ready to merge this. Is this no longer required? As best as I could tell, the example works, but not sure if this was previously doing something important!

packages/gatsby-plugin-emotion/src/gatsby-ssr.js
```diff
@@ -1,23 +1,25 @@
...
 
 exports.replaceRenderer = ({
   bodyComponent,
   replaceBodyHTMLString,
   setHeadComponents,
 }) => {
-  const { html, ids, css } = extractCritical(renderToString(bodyComponent))
-
-  const criticalStyle = <style dangerouslySetInnerHTML={{ __html: css }} />
-  const criticalIds = (
```
DSchau: Same thing here.

packages/gatsby-plugin-emotion/src/gatsby-ssr.js
```diff
@@ -1,23 +1,25 @@
-import React from "react"
-import { renderToString } from "react-dom/server"
-import { extractCritical } from "emotion-server"
+const React = require(`react`)
```
DSchau: Should we use an es module here? I prefer to go that way if possible.

It also means some slight refactors to `gatsby-browser.js`

wKovacs64: I'm cool either way, just wanted to be consistent and some were using CJS so I went with that. Say the word and they'll all be ESM! 🙂

DSchau: Let's do ESM, largely because it _was_ ESM! Bit of a nit, for sure--so thanks for addressing this!

packages/gatsby-plugin-emotion/src/gatsby-browser.js
wKovacs64: v10 changed how critical style extraction works and I _believe_ we don't have to do anything here. @mitchellhamilton can confirm, but it seems to work.

packages/gatsby-plugin-emotion/src/gatsby-ssr.js
wKovacs64: v10 uses the `data-emotion-css` attribute for this now, rather than a string on the `window` object, as far as I understand it. Might need Mitchell's blessing here too, though.

packages/gatsby-plugin-emotion/src/gatsby-browser.js
mitchellhamilton: Yes, this is correct.

DSchau: Appreciate you both for chiming in :) Thank you!

@wKovacs64 lets make that ESM change, and I'll get this merged in. It'll be released as `gatsby-plugin-emotion@3.0.0`

packages/gatsby-plugin-emotion/src/gatsby-ssr.js
wKovacs64: The imports were ESM but the exports were CJS. Do you want it that way, or make both imports and exports ESM?

packages/gatsby-plugin-emotion/src/gatsby-ssr.js
mitchellhamilton: This is correct.

packages/gatsby-plugin-emotion/src/gatsby-ssr.js
DSchau: Yeah - that's weird!

Consistent! So if using CommonJS, use CommonJS for _all_ the things, if ESM, ESM for _all_ the things.

So e.g. `export const replaceRenderer` is the way to go here!

wKovacs64: Agreed. Incoming.

DSchau:
@wKovacs64 I'll take a look at this today and get it merged in. Thanks to both you and @mitchellhamilton!

DSchau:
```
Successfully published:
 - gatsby-plugin-emotion@3.0.0
```

wKovacs64:
Doesn't look like lerna updated the example, I'll send another PR for that.

DSchau:
@wKovacs64 could you provide some more info there?

DSchau:
@wKovacs64 ah, yep 😢 Good catch on the missed breaking change. My fault!

PR #10234, opened 2018-12-01
feat(gatsby): Add type checks to createNodeId

LekoArts:
Before the change it was possible to enter all sorts of parameters. Also Numbers threw a weird error. Both fixed now.

### DIFF
--- a/packages/gatsby/src/utils/create-node-id.js
+++ b/packages/gatsby/src/utils/create-node-id.js
@@ -15,7 +15,7 @@ function createNodeId(id, namespace) {
   if (typeof id === `number`) {
     id = id.toString()
   } else if (typeof id !== `string`) {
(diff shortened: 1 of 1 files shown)


packages/gatsby/src/utils/create-node-id.js
```diff
@@ -11,6 +12,12 @@ const seedConstant = `638f7a53-c567-4eca-8fc1-b23efb1cfb2b`
  * @return {String} - UUID
  */
 function createNodeId(id, namespace) {
+  if (typeof id === `number`) {
+    id = id.toString()
+  } else if (typeof id !== `string`) {
+    report.panic(`Parameter passed to createNodeId must be a String or Number`)
```
pieh: maybe we could display actual type that was passed to give more context?
```suggestion
    report.panic(`Parameter passed to createNodeId must be a String or Number (got ${typeof id})`)
```

Issue #10242, opened 2018-12-02
[Site showcase] Visit site buttons are not mobile responsive


daydream05:
When you visit a site showcase the "visit site" buttons are blocking the "source code" button as seen here.

<img width="374" alt="screen shot 2018-12-02 at 1 43 47 pm" src="https://user-images.githubusercontent.com/11723413/49343513-9567a800-f638-11e8-982a-5885b4579009.png">

Based on this design by @fk this is how it's supposed to look

https://projects.invisionapp.com/share/T9H9LRCG8Z5#/screens/292717378

<img width="362" alt="screen shot 2018-12-02 at 1 47 41 pm" src="https://user-images.githubusercontent.com/11723413/49343523-df508e00-f638-11e8-8e72-977b78fea088.png">

daydream05:
This should be a quick fix. I'll be happy to do a PR

daydream05:
Please no one else fix this. This is probably the only PR i can do 😂

wardpeet:
cleaning up the PR and getting it merged

wardpeet:
@daydream05 thank you for starting on this PR. We finally got it merged!

PR #10249, opened 2018-12-03
fix(www): Fix showcase visit site link overlap

daydream05:
It fixes this issue https://github.com/gatsbyjs/gatsby/issues/10242

- Only moved the top position so it’s not blocking the github button
- Just used arbitrary 8px
- Kept the original layout when it goes back to desktop

site showcase details fix


LekoArts:
@fk Are you okay with that design decision?

wardpeet:
I don't have a strong opinion on it but I don't having it on the next line without overlapping the image.

I'm expanding the question of @LekoArts to @gatsbyjs/ecosystem.
> Are you okay with that design decision?

jlengstorf:
Is there an a11y issue if we drop the label text for smaller screens and just show the "visit link" icon? Theoretically this would allow keeping all the buttons on the same row.

LekoArts:
As long as the icon isn't hidden (with `aria-hidden=true` and `focusable=false`) and has an appropriate `aria-label` that should be fine I guess.

fk:
🙏 + 👍 to @wardpeet's comment and what Jason is suggesting — we have the site link right below the URL anyway.

wardpeet:
@gatsbyjs/ecosystem @fk I updated the PR it now does the following:

nog github button:
showcase-nogithub-small
with github button:
showcase-withgithub-small

large:
showcase-big

PR #10255, opened 2018-12-03
Update building-apps-with-gatsby.md

### LINKED ISSUE #10227: gatsby-plugin-create-client-paths remounts the root on every page transition
## Description

`gatsby-plugin-create-client-paths`  is remounting the root component on every single page transition for the client-only path. This means that if you have auth check in your top-level root, your authentication check will run on each new page instead of just once like it would with a normal React SPA (e.g. one built with create-react-app)

### Expected result

The root component should only mount once.

### Actual result

The root (with routes) is remounted on every single transition.

jaredpalmer:
Add docs about avoiding request waterfalls with client routes by overriding component renderer `key` prop (https://github.com/gatsbyjs/gatsby/issues/10227#issuecomment-443543549)

Closes #10227

### DIFF
--- a/docs/docs/building-apps-with-gatsby.md
+++ b/docs/docs/building-apps-with-gatsby.md
@@ -53,3 +53,18 @@ exports.onCreatePage = async ({ page, actions }) => {
 > [gatsby-plugin-create-client-paths](/packages/gatsby-plugin-create-client-paths/).
 
 Check out the ["simple auth" example site](https://github.com/gatsbyjs/gatsby/blob/master/examples/simple-auth/README.md) for a demo implementing user authentication and restricted client-only routes.
+
+### Known Issues
+
+Some Gatsby plugins manipulate the DOM prior to mounting (e.g. [`gatsby-plugin-twitter`](/packages/gatsby-plugin-twitter/).  In order to support these plugins, Gatsby adds a top level `key` which causes the root to remount each page transition (even those to client only routes). For some hybrid apps, this behavior causes unecessary overfetching/request waterfalls. To avoid this, you can add the following to your site's `gatsby-browser.js`
+
+```js
+import React from "react"
+
+export const replaceComponentRenderer = ({ props }) => (
+  <props.pageResources.component key={props.pageResources.page.path} {...props} />
+)
+```
+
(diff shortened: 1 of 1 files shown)


docs/docs/building-apps-with-gatsby.md
```diff
@@ -53,3 +53,18 @@ exports.onCreatePage = async ({ page, actions }) => {
 > [gatsby-plugin-create-client-paths](/packages/gatsby-plugin-create-client-paths/).
 
 Check out the ["simple auth" example site](https://github.com/gatsbyjs/gatsby/blob/master/examples/simple-auth/README.md) for a demo implementing user authentication and restricted client-only routes.
+
+### Known Issues
```
jaredpalmer: Probably needs a better heading.

docs/docs/building-apps-with-gatsby.md
```diff
@@ -53,3 +53,18 @@ exports.onCreatePage = async ({ page, actions }) => {
...
+
+Some Gatsby plugins manipulate the DOM prior to mounting (e.g. [`gatsby-plugin-twitter`](/packages/gatsby-plugin-twitter/).  In order to support these plugins, Gatsby adds a top level `key` which causes the root to remount each page transition (even those to client only routes). For some hybrid apps, this behavior causes unecessary overfetching/request waterfalls. To avoid this, you can add the following to your site's `gatsby-browser.js`
+
+```js
```
LekoArts: You can add a title to the code blocks 👍 

```suggestion
```js:title=gatsby-browser.js
```

jaredpalmer:
sorry i'm on my phone. can you run prettier?

pieh:
After chat with @KyleAMathews we decided to fix/change this for client-only-paths ( https://github.com/gatsbyjs/gatsby/pull/10261 ) + add some tests to catch any regressions in future.

We will need to update docs on this, but message from this PR will need to be inverted - but that's on us. Thanks for jumping on adding this "gotcha" to docs while we were still debating how to approach the problem.

LekoArts:
@pieh Since https://github.com/gatsbyjs/gatsby/pull/10261 is merged we can close this PR? 😊

DSchau:
Yeah - good call @LekoArts.

@jaredpalmer thanks for this PR, but since we fixed the underlying issue, that's probably preferable over documenting something that isn't necessarily needed.

If I'm off-base, please feel free to re-open and we can chat further!

Thanks for using Gatsby 💜

PR #10257, opened 2018-12-03
feat(gatsby): configure physical cores, logical_cores or fixed number

### LINKED ISSUE #11727: --max-workers flag for gatsby build
## Summary

Currently, Gatsby [automatically selects](https://github.com/gatsbyjs/gatsby/blob/22c41fb4595a5c57c97cd014044d8d772d781809/packages/gatsby/src/utils/html-renderer-queue.js#L4) a level of parallelism (based on the reported number of available CPUs) for the render phase of `gatsby build`. In some scenarios, though, `gatsby build` can actually run faster if the amount of parallelism is limited or even set to 1.

One such scenario may be encountered on cloud/container environments that have [virtualized CPU](https://www.hyve.com/what-is-a-vmware-vcpu/) resources, where it is difficult (or impossible) to discover exactly how many vCPU or 'cores' are available for parallelizing the build.

On [CircleCI](https://circleci.com/docs/2.0/configuration-reference/#resource_class), a typical CI server will have 2 vCPUs available, but when Gatsby asks, it will be told there are _18_. As a result, the build may start parallelizing many more render tasks than the environment can handle, paradoxically causing the overall build time to grow significantly.

As  a real-world example, here are some samples of a Gatsby build of ~2400 pages on CircleCI with [`numWorkers`](https://github.com/gatsbyjs/gatsby/blob/22c41fb4595a5c57c97cd014044d8d772d781809/packages/gatsby/src/utils/html-renderer-queue.js#L4) modified to particular values:

| `numWorkers` | HTML done in |
|     :---:    |     :---:    |
| 1 | ~58s |
| 2 | ~20s |
| 4 | ~14s |
| 8 | ~11s  |
| 16 | __server timeout!__ |
| \<reported cpu count\> | __server timeout!__ |

> reported cpu count is `18`, and the server timeout limit is `10m`.

### Basic example

Usage would be simple:

```shell
gatsby build --max-workers=2
```

When defined, this value would be used instead of programmatically determining how many workers to use.

The [Jest](https://jestjs.io/docs/en/cli#maxworkers-num-string) version of this feature also supports percentage values, like:

```shell
gatsby build --max-workers=50%
```

Which might be a nice addition for situations where the number of cores is variable.

### Motivation

This feature is inspired by the fact that Gatsby uses [jest-worker](https://github.com/gatsbyjs/gatsby/blob/22c41fb4595a5c57c97cd014044d8d772d781809/packages/gatsby/src/utils/html-renderer-queue.js#L3) to achieve its parallelism, and by the fact that [Jest has similar features](https://jestjs.io/docs/en/troubleshooting.html#tests-are-extremely-slow-on-docker-and-or-continuous-integration-ci-server) that are recommended for controlling the parallel characteristics to improve performance in CI environments.

dominicfallows:
Gatsby 2 now utilises multi-core builds using jest-worker. By default Gatsby creates a pool of workers equal to the number of physical cores on your machine, see [build-html.js](https://www.gatsbyjs.org/docs/html-generation/#build-htmljs).

In some scenarios it may be appropriate to tell Gatsby to use a different method to calculate the number of worker pools. 

For example, if you are running a Cloud server (like AWS EC2) your DevOps engineers may want to control the number of worker pools to improve the efficiency of server resource usage.

The proposal here is to accept an options env var `GATSBY_CPU_COUNT` to change the method that Gatsby uses to calculate the number of worker pools.

closes #11727


KyleAMathews:
So the scenario is you're running builds on a server with other work also going on and you want to limit the number of cores Gatsby uses?

dominicfallows:
@KyleAMathews either that yes, or the opposite, sometimes we want to allow logical CPU count (more cores). For example, our AWS instances are dedicated to this process, and therefore we want to be able to push them to their limits and speed up our builds.

dominicfallows:
Ok, sounds interesting. I'll have a dig around and look for the `sharp` handling to see how that could work alongside this. 

In future, would you see a global CPU limit/count using an env var as I've drafted, or another config option? etc.

dominicfallows:
@KyleAMathews ok, so I've made a first attempt at aligning sharp usage with proposed Gatsby cpuCount usage. Along the right lines based on what you were thinking?

KyleAMathews:
Could you post build logs from using Physical vs. Logical CPU counts?

If that generally speeds up builds — we should just change our default to use the logical CPU count.

I'd rather not add config if we don't need to.

Could you try running some of the benchmarks w/ both physical/logical and see how that changes things? https://github.com/gatsbyjs/gatsby/tree/master/benchmarks

docs/docs/multi-core-builds.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Warning
+
+You could negatively impact performance if you use this variable incorrectly. The default Gatsby setting (no env variable or `physical_cores`) is the safest option.
+
+### Setup
+
+Set the `GATSBY_CPU_COUNT` environment variable whilst running the `gatsby build` command.
+
+`GATSBY_CPU_COUNT=physical_cores` - (default) calculate the number worker pools based on the number of physical CPU cores on your machine.
```
shannonbux: looks like it ought to read "number of" in each of these lines. Is that right?

docs/docs/multi-core-builds.md
```diff
@@ -0,0 +1,33 @@
...
+
+`GATSBY_CPU_COUNT=2` - calculate the number worker pools based on a definite number.
+
+### More information
+
+Understanding how processors work is complex and out of scope for this documentation.
+
+In summary, some processors use 'Simultaneous Multithreading (SMT)' (sometimes known as 'Hyper-Threading') which is the process of a CPU splitting each of its physical cores into virtual/logical cores.
+
+This _can_ help some workloads to increase performance and allow each physical core to run two streams of work at once. However, sometimes latency can be increased as these logical cores share the same physical CPU core, more memory is required to assign to each worker in the pools and more time is needed spawn worker processes.
```
shannonbux: I got lost in the last sentence. I'm not sure exactly how to rewrite it. It's technically a run-on sentence.

www/src/data/sidebars/doc-links.yaml
```diff
@@ -38,6 +38,8 @@
           link: /docs/path-prefix/
         - title: How Gatsby works with GitHub pages
           link: /docs/how-gatsby-works-with-github-pages/
+        - title: Multi-core builds
+          link: /docs/multi-core-builds
```
shannonbux: needs a trailing slash

docs/docs/multi-core-builds.md
```diff
@@ -0,0 +1,33 @@
+---
+title: Multi-core builds
+---
+
+Gatsby now performs the static HTML generation phase of the overall [Page HTML Generation](/docs/html-generation/) process using multi-core parallel pools of workers. This helps speed up builds by distributing build generation tasks across multiple cores on your machine.
+
+By default Gatsby creates a pool of workers equal to the number of physical cores on your machine, see [build-html.js](/docs/html-generation/#build-htmljs)
```
shannonbux: Punctuation tip: Introductory phrases need commas after them in order to mimic natural speech. For example
> By default, Gatsby...
> In some scenarios, it may...

Commas around nonessential / parenthetical phrases help to mimic the way people speak. Example:
> **For example**, if you are running a Cloud server (like AWS EC2), your DevOps engineers may

Would you please update commas in this doc?

mik-laj:
Should the main file be called `cpu-count`? I am afraid that this name can be misleading. I am afraid that few people run the program in a multi-CPU environment. However, everyone runs the program in a multi-core environment. The better name will be core-count, but it is still not correct.

Valid name for me is `parallelism`.

Reference: https://learn-gevent-socketio.readthedocs.io/en/latest/general_concepts.html#what-s-the-difference-between-concurrency-and-parallelism

packages/gatsby/src/utils/cpu-count.js
```diff
@@ -0,0 +1,46 @@
...
+    const cpuCountArg =
+      parseInt(process.env.GATSBY_CPU_COUNT, 10) || process.env.GATSBY_CPU_COUNT
+
+    switch (typeof cpuCountArg) {
+      case `string`:
+        // Leave at Default CPU count if cpuCountArg === `physical_cores`
+
+        // CPU count === logical CPU count or default
+        if (cpuCountArg === `logical_cores`) {
+          cpuCount = logical_cores || cpuCount
```
wardpeet: shouldn't we throw? it should always return a value if not I think something is wrong 😄

wardpeet:
pinging @KyleAMathews & @pieh wdyt?

I think this is great, especially on VMs.

KyleAMathews:
Haven't looked at the PR but 👍 to the concept

wardpeet:
@dominicfallows did you have some benchmarks? I'm a bit hesitant to add this, it's opt-in but for example sharp will probably have a negative impact as it probably best to use cores instead of threads as it has so many heavy computations.

I've found an issue at the Parcel repo and it is not moving to logical cores:
https://github.com/parcel-bundler/parcel/issues/1554#issuecomment-397852006

packages/gatsby/src/utils/cpu-count.js
```diff
@@ -0,0 +1,46 @@
...
+
+  if (alwaysReturnADefault) {
+    // Default CPU count === physical CPU count,
+    // or default to 1 if we can't detect
+    cpuCount = physical_cores || 1
+  }
+
+  if (typeof process.env.GATSBY_CPU_COUNT !== `undefined`) {
+    const cpuCountArg =
+      parseInt(process.env.GATSBY_CPU_COUNT, 10) || process.env.GATSBY_CPU_COUNT
```
wardpeet:
```suggestion
      Number(process.env.GATSBY_CPU_COUNT) || process.env.GATSBY_CPU_COUNT
```
with number you don't need to add the radix arg.

docs/docs/multi-core-builds.md
dominicfallows: Updated :)

docs/docs/multi-core-builds.md
dominicfallows: Updated, does it read better? :)

www/src/data/sidebars/doc-links.yaml
dominicfallows: Done

docs/docs/multi-core-builds.md
dominicfallows: Updated :)

dominicfallows:
@mik-laj updated the file name to `cpu-core-count.js` and helper function to `cpuCoreCount` which does seem more accurate, thanks :)

packages/gatsby/src/utils/cpu-count.js
dominicfallows: Agreed, I've updated

packages/gatsby/src/utils/cpu-count.js
dominicfallows: Updated :)

dominicfallows:
@wardpeet RE: https://github.com/gatsbyjs/gatsby/pull/10257#issuecomment-468238584 

Benchmarks are proving quite difficult to show using the existing benchmark examples. The improvements are so dependant on the infrastructure setup and setup of the app, hence why I went for an 'opt-in' approach. 

I could create a new benchmark example though, that would go to show the type of app setup and infrastructure where I'm seeing improvements (cloud containers and instances, for example), would that help?

wardpeet:
I guess you're using this already on AWS? If so maybe share the numbers you're encountering? Or maybe a shared blog post or document where this really shows physical vs logical on the cloud infra 😄

> I also wondered about a different approach, somehow having this calculation as a plugin. It would still require changes to the core code (to allow plugins to override the CPU core count calculation), but would move away from needing an env var.

I don't think we want to expose such an API to the world 😄. For now I think env var is good enough and we need to figure out this a bit more for other things as well like experiments so we're probably going to revisit this for Gatsby 3.

packages/gatsby/src/utils/cpu-core-count.js
```diff
@@ -0,0 +1,58 @@
...
+        Number(process.env.GATSBY_CPU_COUNT) || process.env.GATSBY_CPU_COUNT
+
+      switch (typeof coreCountArg) {
+        case `string`:
+          // Leave at Default CPU count if coreCountArg === `physical_cores`
+
+          // CPU count === logical CPU count
+          // throw error if we have a problem counting logical cores
+          if (coreCountArg === `logical_cores`) {
+            coreCount = logical_cores
```
wardpeet:
```suggestion
            coreCount = require(`os`).cpus().length
```
this way we only run cpus when needed. It probably doesn't matter much but 🤷‍♂️

dominicfallows: Good call, updated

paulca99:
Hi, I'm the dev ops tech guy dealing with the codebuild images  running this stuff.  To be blunt, Doms "Logical cores" change knocked nearly 30% off our AWS Codebuild timings. If it's a concern to you, make it a variable... use logical .   vs .    use physical.... give users the option.

paulca99:
it's been 3 months since Dom, discovered this and you're still debating things. if I were you I'd have added a switch, then debated the output later once you had some user results to judge it on.

wardpeet:
I was going to merge this when tests are passing.

@paulca99 Sorry but we have to deal with a lot of PRs and issues. Because of the holidays, it slipped our mind. There is nothing wrong to ask for validation and extra information before merging even if it's opt-in. Extra code could lead to bugs or build errors.

paulca99:
Sorry @wardpeet I didn't notice there were tests failing.

dominicfallows:
@wardpeet Hey, I've updated a test in the `gatsby-plugin-manifest` package and tests now look good. Let me know what else I can share and I'll do it asap.

sidharthachatterjee:
Published in 

gatsby@2.1.20
gatsby-plugin-sharp@2.0.24
gatsby-plugin-manifest@2.0.21

PR #10261, opened 2018-12-03
fix(gatsby): don't remount matchPath pages

### LINKED ISSUE #10227: gatsby-plugin-create-client-paths remounts the root on every page transition

pieh:
Actual fix is one line change - most of this PR is cypress testing to catch any behaviour changes in future PRs

fixes #10227


packages/gatsby/cache-dir/page-renderer.js
```diff
@@ -20,7 +20,7 @@ class PageRenderer extends React.Component {
       replacementElement ||
       createElement(this.props.pageResources.component, {
         ...props,
-        key: this.props.location.pathname,
+        key: this.props.pageResources.page.path,
```
pieh: actual fix

e2e-tests/production-runtime/cypress/integration/lifecycle-methods.js
```diff
@@ -0,0 +1,41 @@
+describe(`Production build tests`, () => {
+  it(`should remount when navigating to different template`, () => {
+    cy.visit(`/`).waitForAPI(`onRouteUpdate`)
+
+    cy.getTestElement(`page2`)
+      .click()
+      .waitForAPI(`onRouteUpdate`)
+
+    cy.window().then(win => {
+      cy.wrap(win.___PageComponentLifecycleCallsLog).snapshot()
```
pieh: using snapshot testing, which is quite verbose and might be hard to read - it might make sense to not use snapshots here and use more programmatic assertions

KyleAMathews: One way to test this would be count renders in local state and add that to the DOM and assert against it

packages/gatsby/cache-dir/page-renderer.js
KyleAMathews: 🎉

e2e-tests/production-runtime/src/pages/index.js
```diff
@@ -2,22 +2,40 @@ import React from 'react'
...
-    <Link to="/page-2/">
-      <span data-testid="page2">Go to page 2</span>
-    </Link>
-    <Link to="/page-3/" data-testid="404">
-      To non-existent page
-    </Link>
-    <Link to="/long-page/" data-testid="long-page">
-      To long page
-    </Link>
+    <pre data-testid="dom-marker">{pageContext.DOMMarker || `index`}</pre>
```
DSchau: This is pretty slick. Nice work!

jaredpalmer:
Thanks for the turnaround on this one! Was pulling my hair out on the undefined.fm's upcoming job board on this. You all rock!

PR #10263, opened 2018-12-03
Remove react-emotion from tutorial

garrowp:
Emotion removed react-emotion package. Not sure which version to put this in but part 4 of the tutorial will be fixed.

### DIFF
--- a/docs/tutorial/part-four/index.md
+++ b/docs/tutorial/part-four/index.md
@@ -96,14 +96,14 @@ Then install some other needed dependencies at the root of the project. You'll u
 "Kirkham", and you'll try out a CSS-in-JS library, ["Emotion"](https://emotion.sh/):
 
 ```shell
-npm install --save gatsby-plugin-typography typography react-typography typography-theme-kirkham gatsby-plugin-emotion emotion react-emotion emotion-server
+npm install --save gatsby-plugin-typography typography react-typography typography-theme-kirkham gatsby-plugin-emotion emotion emotion-server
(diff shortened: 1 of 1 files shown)


DSchau:
@KyleAMathews sorry missed this, but yeah - there are some changes in emotion 10 that make things slightly more complicated--at least at first glance.

I need to dive in a bit more before I fully propose that though.

@LekoArts and I have also talked about it briefly. If the changes in emotion 10 seem substantive enough, styled-components could be a little more beginner friendly, and more applicable to a tutorial.

BoyWithSilverWings:
The sheer number of plugins that are being installed in the tutorial could be a bit confusing for a beginner (`@emotion/core emotion/styled emotion gatsby-plugin-emotion emotion-server` and much more available in docs) Switching over the tutorial to `styled-components` should be nice and easy for the tutorial purpose.

shannonbux:
So for now, @garrowp it sounds like it's best to stick with Emotion v9 and make that clear. Do you mind changing this PR to account for that change?

Then, the suggestions to either use emotion v10 or styled components ought to be put into another issue. @BoyWithSilverWings and/or @DSchau, want to make that issue?

DSchau:
Going to close this out.

The complexity argument is certainly noted here, but I think that's a separate discussion. Happy to have it, just don't think it's applicable here.

As we've made some changes to the tutorial to account for Emotion 10, we can safely close this out.

See #10307 and #10306 for further context.

PR #10265, opened 2018-12-04
Docs - Add clarification to custom domains on github pages

drewatk:
## Changes
- Changed the folder in which the `CNAME` file should be copied to. Github wants it in the root of the folder you deploy
- Added an example deploy script that copies `CNAME` from `src/` on every deploy


pieh:
Current description is correct. Gatsby will copy all files from `static` to `public` directory on builds ( https://www.gatsbyjs.org/docs/static-folder/ )

PR #10276, opened 2018-12-04
Service workers note

danielcooke1996:
- Adding a note that informs readers that the service workers will only register on a production build.
- I was confused and I had to dig into the source code to find out why they weren't working on develop

### DIFF
--- a/docs/docs/add-offline-support-with-a-service-worker.md
+++ b/docs/docs/add-offline-support-with-a-service-worker.md
@@ -41,6 +41,8 @@ Add this plugin to your `gatsby-config.js`
 
 That's all you need to add offline support to your Gatsby site.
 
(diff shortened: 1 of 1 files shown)


docs/docs/add-offline-support-with-a-service-worker.md
```diff
@@ -41,6 +41,8 @@ Add this plugin to your `gatsby-config.js`
 
 That's all you need to add offline support to your Gatsby site.
 
+Note: You must create a production build using `gatsby build` for the service workers to register.
```
pieh: Maybe something like (just to avoid words like `must` but still convey the message)
```suggestion
Note: Service worker registers only in production builds (`gatsby build`).
```

PR #10280, opened 2018-12-04
fix(gatsby-source-filesystem): allow empty password for basic auth in createRemoteFileNode

brxck:
This allows createRemoteFileNode authentication with either user or password. Currently auth params with falsy keys are ignored, but not all authentication schemes require both user and password (e.g. [Stripe API](https://stripe.com/docs/api/authentication?lang=curl)).

### DIFF
--- a/packages/gatsby-source-filesystem/src/create-remote-file-node.js
+++ b/packages/gatsby-source-filesystem/src/create-remote-file-node.js
@@ -188,7 +188,7 @@ async function processRemoteNode({
 
   // Add htaccess authentication if passed in. This isn't particularly
   // extensible. We should define a proper API that we validate.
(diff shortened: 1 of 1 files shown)


pieh:
This is pretty weird (on the Stripe API part, heh). Did you verify that this change allow to pull from files Stripe API?

brxck:
Yes, after this change it's possible to access the authenticated portions of the Stripe API by specifying only a username. 

Actually, I found out that Stripe discards any password you provide, so the password so it's currently possible to

brxck:
Yes, after this change it's possible to access the authenticated portions of the Stripe API by specifying only a username. 

Actually, I found out that Stripe discards any password you provide, so the password so it's currently possible to

PR #10291, opened 2018-12-04
fix(gatsby-plugin-emotion): allow for React.Fragment shorthand syntax

wKovacs64:
This change resolves the following error thrown by `@babel/plugin-transform-react-jsx` when using the React.Fragment shorthand syntax (`<> </>`) in conjunction with emotion v10 and `gatsby-plugin-emotion` v3:

```
Error: transform-react-jsx: pragma has been set but pragmafrag has not been set
```

### DIFF
--- a/packages/gatsby-plugin-emotion/src/gatsby-node.js
+++ b/packages/gatsby-plugin-emotion/src/gatsby-node.js
@@ -14,6 +14,7 @@ export const onCreateBabelConfig = ({ actions }, pluginOptions) => {
     name: `@babel/plugin-transform-react-jsx`,
     options: {
       pragma: pragmaName,
+      pragmaFrag: `React.Fragment`,
     },
   })
 


wKovacs64:
The error went away and it started working when I made this change locally. 🤷‍♂️ I mentioned in the Emotion RFC for their upcoming Babel preset and Mitchell [included it there](https://github.com/emotion-js/emotion/pull/1049/commits/1c3b376058f7de26ffe058e8019d660636106cfb#diff-7da1638584fa6fbbfe12fadbf1ff2d9eR29).

janicduplessis:
I've also hit this issue when upgrading and can confirm that it fixes it. Kind of weird that babel forces you to specify both instead of keeping the default value.

t2ca:
Since the update to V10, the code below stopped working.  Any ideas how to fix this?

```
const getNavItemStyles = ({ isCurrent }) =>
  isCurrent
    ? {
        className: css({
          ...styles.link.default,
          ...styles.link.active,
        }),
      }
    : {
        className: css({
          ...styles.link.default,
        }),
      }

const MobileNavItem = ({ linkTo, label, icon }) => (
  <Link to={linkTo} getProps={getNavItemStyles}>
    <span>{icon}</span>
    <div>{label}</div>
  </Link>
)
```

<a aria-current="page" class="[object Object]" href="/">

DSchau:
@t2ca that's due to a change in emotion. className isn't produced from that css call any longer.

See [this document](https://github.com/emotion-js/emotion/blob/master/docs/migrating-to-emotion-10.md) for more info.

t2ca:
Thank you @DSchau, I was able to find a solution!

DSchau:
```
Successfully published:
 - gatsby-plugin-emotion@3.0.1
```

PR #10296, opened 2018-12-05
reorder imports in tutorial part 2

jimt:
The `import` statements need to precede the `console.log` to avoid the error:
```Import in body of module; reorder to top  import/first```
Either the line numbers of the insertion(s) need to change or simply show desired top of the file.

Further fix to issue #6897

### CODE REVIEW

**Comment by jimt** [comment]
The current instructions in the tutorial do not work for me.  This appears to be the same issue as #6897 but the solution there only fixed one of the occurrences in the tutorial.

`import` statements must be at the top of the file. There can be no intervening `console.log` statement.  Following the tutorial instructions to _add the following on lines 2 and 3_ you have (assuming lines in a file are 1-indexed):
```
import React from "react"
import styles from "./about-css-modules.module.css"
console.log(styles)

```
which will fail to compile, producing the error:
```Import in body of module; reorder to top  import/first```

By changing the line numbers you could append the lines after the `import Container [...]` but then you would not [parallel the lines in the next step of the tutorial](https://github.com/gatsbyjs/gatsby/edit/master/docs/tutorial/part-two/index.md#L439-L443).

Showing the top of the file eliminates confusion about whether numbering is 0 or 1-indexed, and parallels the later example.

File: docs/tutorial/part-two/index.md [hunk 1]
```diff
@@ -407,14 +407,17 @@ Paste the following into the file:
 }
 ```
 
-Now import that file into the `about-css-modules.js` page you created earlier, by adding the following on lines 2 and 3.
-(The `console.log(styles)` code logs the resulting import so you can see what the processed file looks like).
+Now import that file into the `about-css-modules.js` page you created earlier, by editing the first few lines
+of the file to look like:
 
 ```javascript:title=src/pages/about-css-modules.js
```
shannonbux: @DSchau actually I do think this is an improvement. I haven't come up with an official pattern for how to visualize changes to a file; however, in almost every other instance, we simply show the same file with the new lines highlighted.

So actually, @jimt, do you mind updating this PR to include the lines that are new in the file? Here's an example of what goes at the top of the code block
`javascript{2,5}:title=src/pages/about-css-modules.js`

shannonbux: Also this corrects the error, which was that the `console.log(styles)` is not added to line 3, but rather line 5.

jimt: Thanks for pointing out the highlighting method.

PR #10301, opened 2018-12-05
Enable RichText for all users

Khaledgarbaya:
Now that RichText is in beta and more stable we can remove the opt-in flag to make it available for all users


Khaledgarbaya:
@pieh Done

PR #10305, opened 2018-12-05
Add transitive dependency for PnP

Ikana:
Hope you don't think I'm spamming PR, it's just when I fix transitive dependency error another one turns up 😅

Maybe this need to be peerDependency??


pieh:
`graphql` package is quite is problematic - we must ensure there is only 1 `graphql` package instance in `node_modules` (what PnP is not used obviously), but yeah moving it `peerDependencies` might actually best course of action here. `graphql-skip-limit` is dependency of `gatsby` and so is `graphql`, so this should work.

PR #10306, opened 2018-12-05
Update index.md

wesleylhandy:
Update imports statements to reflect `emotion/core` instead of `react-emotion` and the commensurate use of `css` property rather than `className`. Also added an explanatory note on the `graphql` sort statement that there are three underscores, instead of two, which the eyes might only see.


DSchau:
@shannonbux #10263 is basically invalid and I should probably close it out. It was created when we hadn't released the new plugin and added compatibility with emotion 10.

PR #10307, opened 2018-12-05
Update index.md

wesleylhandy:
updated import of `emotion/core` in place of `react-emotion` and made commensurate change of `className` prop to `css.`


wesleylhandy:
Note: this is for a different tutorial page than previous commit

PR #10310, opened 2018-12-05
chore: add starters to monorepo

DSchau:
This PR adds in starters to the monorepo. Longer term, I believe we'll merge starters/examples since they're very much linked.

Future implementation improvements:

- Run a `build` script to validate that the starter(s) can be built
  - I planned to use `lerna run --scope gatsby-starter-*` for this, but if there's another approach, happy to try that
- Implement yarn workspaces/lerna so we can get automatic versioning in the starters (I have this built out at https://github.com/dschau/starters)
- Author `e2e` tests vs. the starters so that we can ensure we're not breaking anything
  - These would follow a similar approach to the current e2e tests

Blocked by

gatsbyjs/gatsby-starter-blog#139 gatsbyjs/gatsby-starter-default#120 gatsbyjs/gatsby-starter-hello-world#32


.circleci/config.yml
```diff
@@ -133,6 +133,20 @@ jobs:
       - <<: *attach_to_bootstrap
       - run: yarn test:integration
 
+  starters_publish:
```
DSchau: I'll add back the other jobs, but this is helpful to validate this approach and get feedback more quickly.

See [CircleCI](https://circleci.com/gh/DSchau/gatsby/48) for an example run.

KyleAMathews:
Oh cool! Yeah run Cypress tests and then remove them before syncing with the read-only repos.

scripts/clone-folder.sh
```diff
@@ -0,0 +1,18 @@
...
+BASE=$(pwd)
+
+for folder in $FOLDER/*; do
+  [ -d "$folder" ] || continue # only directories
+  cd $BASE
+
+  NAME=$(cat $folder/package.json | jq -r '.name')
+  CLONE_DIR="__${NAME}__clone__"
+
+  git clone https://$GITHUB_TOKEN@github.com/dschau/$NAME.git $CLONE_DIR
```
DSchau: Before this is merged, we will need to create a GITHUB_TOKEN environment variable in CircleCI which has `repo` permission.

scripts/clone-folder.sh
```diff
@@ -0,0 +1,18 @@
...
+  cp -r $folder/. $CLONE_DIR
+  cd $CLONE_DIR
+  git add .
+  git commit --message "$(git log -1 --pretty=%B)"
```
DSchau: This uses the _current_ commit's message as the message from "GatsbyJS Bot".

jonniebigodes:
@DSchau if you don't mind my intrusion, but feel free to remove my pull request on the hello world starter. And when all the starters will be converged and consolidated in one place, and if you and @KyleAMathews won't mind, i would like to go through them and start updating the gatsby dependency initially and then moving onto the plugins. This could sound like a monumental task, but i don't mind, more even to prevent issues like [this](https://github.com/gatsbyjs/gatsby/issues/3183#issuecomment-439692575) and [this](https://github.com/gatsbyjs/gatsby/issues/10404) one that popped early this week from happening. Feel free to provide some feedback on what i've said.

KyleAMathews:
@jonniebigodes one thing we could do as part of this move is to revisit how we maintain the package-lock.json files e.g. regenerate them every time we sync to their repos.

DSchau:
More work to do here:

The clone and copy files approach won't work; how do we handle deleted files?

Follow along at dschau/gatsbyjs#5, trying a new approach. I think I'll have to merge origin/master and that should do the trick.

sidharthachatterjee:
Bot to close all new pull requests to starter repos is ready at https://github.com/gatsbyjs/peril-gatsbyjs/pull/64

starters/README-template.md
```diff
@@ -0,0 +1,97 @@
...
+
+9.  **`LICENSE`**: Gatsby is licensed under the MIT license.
+
+10. **`package-lock.json`** (See `package.json` below, first). This is an automatically generated file based on the exact versions of your npm dependencies that were installed for your project. **(You won’t change this file directly).**
+
+11. **`package.json`**: A manifest file for Node.js projects, which includes things like metadata (the project’s name, author, etc). This manifest is how npm knows which packages to install for your project.
+
+12. **`README.md`**: A text file containing useful reference information about your project.
+
+13. **`yarn.lock`**: [Yarn](https://yarnpkg.com/) is a package manager alternative to npm. You can use either yarn or npm, though all of the Gatsby docs reference npm. This file serves essentially the same purpose as `package-lock.json`, just for a different package management system.
```
sidharthachatterjee: We don't include `yarn.lock` in starters anymore so might want to remove this point

DSchau: ✅ Done!

PR #10312, opened 2018-12-05
Rename `community` to `creators`

amberleyromo:
Currently `/community` redirects to `/docs/community/`. This PR changes it to be namespaced under `creators`.

Since this wasn't formally launched (and I still can't track down what happened there), I don't see a need to do redirects. Thoughts?


DSchau:
> Since this wasn't formally launched (and I still can't track down what happened there), I don't see a need to do redirects. Thoughts?

Yeah - I wouldn't add a redirect! Think it's OK to just go ahead and make the change.

DSchau:
@amberleyromo is this good to go?

amberleyromo:
@DSchau tis. just resolved conflicts 👍

PR #10315, opened 2018-12-05
fix(gatsby-plugin-emotion): promote @babel/core to a real dependency

wKovacs64:
This gets rid of the following warnings when using `gatsby-plugin-emotion` v3:

```
warning "gatsby-plugin-emotion > @babel/plugin-transform-react-jsx > @babel/plugin-syntax-jsx@7.0.0" has unmet peer dependency "@babel/core@^7.0.0-0".
warning "gatsby-plugin-emotion > @babel/plugin-transform-react-jsx@7.1.6" has unmet peer dependency "@babel/core@^7.0.0-0".
```

`@babel/core` is already present due to many other packages depending on it (including Gatsby itself), but some users may not be aware of this and the warnings may cause them unnecessary concern.

### DIFF
--- a/packages/gatsby-plugin-emotion/package.json
+++ b/packages/gatsby-plugin-emotion/package.json
@@ -7,6 +7,7 @@
     "url": "https://github.com/gatsbyjs/gatsby/issues"
   },
   "dependencies": {
+    "@babel/core": "^7.0.0",
     "@babel/plugin-transform-react-jsx": "^7.1.6",
     "@babel/runtime": "^7.0.0",
     "babel-plugin-emotion": "^10.0.0",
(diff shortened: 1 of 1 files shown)


wKovacs64:
Yeah, I wasn't sure about it either. 🤔 Just trying to make usage as clean and easy as possible for the users.

`@babel/plugin-transform-react-jsx` lists `@babel/core` as a peerDep, so technically I think using it means you have to install `@babel/core`, which would not happen _in isolation_ without this change. It _does_ happen, coincidentally, because so many other things install it and because most likely very few people are installing with a production-only flag (skipping devDeps). But theoretically I think it's possible, and using this plugin would fail if they did so (as it stands currently), which I think is why the warning is there.

Are you aware of any downsides to making this change?

wKovacs64:
Works for me, thanks for taking a look. 👍

PR #10329, opened 2018-12-06
fix(gatsby-plugin-offline): gracefully degrade if appshell isn't precached

pieh:
It's not clear what cause app shell to not be succesfully precached and this is hot fix to fix serving blank page with following error in console:
`The FetchEvent for "<url>" resulted in a network error response: an object that was not a Response was passed to respondWith().`

Ideally we would try to push that offline appshell to precache in this case. Fix in this PR will only work when people are online

/cc @davidbailey00

### DIFF
--- a/packages/gatsby-plugin-offline/src/sw-append.js
+++ b/packages/gatsby-plugin-offline/src/sw-append.js
@@ -3,13 +3,7 @@
 importScripts(`idb-keyval-iife.min.js`)
 const WHITELIST_KEY = `custom-navigation-whitelist`
 
-let failedToGrabAppShell = false
-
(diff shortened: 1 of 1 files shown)


pieh:
So I introduced a bit of `then` monster, it would be a lot nicer with async/await there, but for that we would need to adjust build process for package which is out of the scope of this PR.

Changes I've made:
If app shell is not in precache (for whatever reason) - try to fetch it. If we fetch successfully - put it in precache for any future requests so we can use happy path again. If we can't fetch app shell - fallback to fetching actual html content of the page.

Huge thanks to @davidbailey00 for help on this one!

PR #10335, opened 2018-12-06
Add support for queryParams in gatsby-source-wordpress

u12206050:
Add support to specify a queryParams object on the options of gatsby-source-wordpress plugin.
Query params are added to each entity request to Wordpress allowing to filter on specific fields such as lang.
```
  {
    ...
    // Only get posts & pages that are english
    queryParams: {
      lang: 'en',
    },
  }
```


packages/gatsby-source-wordpress/src/fetch.js
```diff
@@ -341,11 +346,14 @@ async function getPages(
     let result = []
 
     const getOptions = page => {
+      // Do not filter tags & categories
+      let qs = url.match(/(tags|categories)/ig) ? {} : _queryParams
```
pieh: Can you give context on this?

u12206050: Since tags and categories could be referenced across various posts even different languages filtering them could cause the gatsby-source-wordpress plugin to fail when it normalizes ids of tags&categories within posts and pages.
It could be updated to also be an option to decide exactly which routes queryParams should not be included on.

pieh: I could see language filtering not being only use case for `queryParams`. It also depends on which multilanguage plugin (or even what settings of plugin) you use.

I would need to lookup what are potential use cases that users would want to use queryParams other than language filtering. What I'm thinking right now that `queryParams` could accept functions:
```js
// in gatsby-config
{
  resolve: `gatsby-source-wordpress`,
  options: {
    [...]
    queryParams: ({ url }) => url.match(/(tags|categories)/ig) ? {} : _queryParams
  }
}
```

u12206050: Agreed that is much more flexible.

u12206050:
What is the status of this? Bound for release or?

simonjoom:
hi, thks i need that,
i found a miss in the file fetch.js

    _queryParams is not added in list of params

getOptions take _queryParams but _queryParams is not in params.

      const getOptions = page => {
       let qs = _queryParams
      if (typeof _queryParams === 'function') {
        qs = _queryParams({ url })
      }
      if (!qs || typeof qs !== 'object') qs = {}

sidharthachatterjee:
@pieh Is this good to go?

pieh:
I created detailed issue about implementing this feature ( https://github.com/gatsbyjs/gatsby/issues/17943 ) and will be closing this PR for now.

PR #10342, opened 2018-12-07
react-docgen: add multiple same doclet values as array

jquense:

### DIFF
--- a/packages/gatsby-transformer-react-docgen/src/Doclets.js
+++ b/packages/gatsby-transformer-react-docgen/src/Doclets.js
@@ -1,6 +1,5 @@
 const DOCLET_PATTERN = /^@(\w+)(?:$|\s((?:[^](?!^@\w))*))/gim
 
-const { hasOwnProperty: has } = Object.prototype
 let cleanDocletValue = str => {
   str = str.trim()
   if (str.endsWith(`}`) && str.startsWith(`{`)) str = str.slice(1, -1)
@@ -26,15 +25,15 @@ export const cleanDoclets = desc => {
  * Updated to strip \r from the end of doclets
  */
 const getDoclets = str => {
-  let doclets = Object.create(null)
+  let doclets = []
   let match = DOCLET_PATTERN.exec(str)
   let val
 
   for (; match; match = DOCLET_PATTERN.exec(str)) {
     val = match[2] ? match[2].replace(/\r$/, ``) : true
-    doclets[match[1]] = val
+    const key = match[1]
+    doclets.push({ tag: key, value: val })
   }
-
   return doclets
 }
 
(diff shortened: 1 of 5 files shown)


packages/gatsby-transformer-react-docgen/src/Doclets.js
```diff
@@ -32,7 +32,8 @@ const getDoclets = str => {
 
   for (; match; match = DOCLET_PATTERN.exec(str)) {
     val = match[2] ? match[2].replace(/\r$/, ``) : true
-    doclets[match[1]] = val
+    const key = match[1]
+    doclets[key] = doclets[key] ? [].concat(doclets[key], val) : val
```
pieh: Will this potentially cause type conflicts during schema inferring? For given node type, that field can be either single value or array

jquense: it shouldn't, these are typed as JSON now-a-days

pieh: screenshot 2018-12-07 at 17 43 45
So we still show false warning - seems like the ignore thingy stopped working (or didn't work for deep paths). I'll take a look on that.

I don't really like putting more handling of data shape on the users (sometimes property will be single, sometimes array), but unless we do something like define which properties should be arrays globally there's not much we can do about it I guess

pieh: Agh, we broke it recently - will fix and add some tests to catch regressions, so as far as this PR goes, I think it's good to go. But I do think we need to bump major version here as this will potentially break some sites

jquense: yes major bump sounds good 👍

pieh:
Would you be fine if I would make plugin option to declare which doclets should be arrays and then just force those to be arrays?

I'm pretty conflicted with shape of the data being so fluid (sometimes `property` would be string, sometimes array of strings), so it's easy for builds to break after adding/removing properties - it's all on the end dev to properly guard for that.

I was thinking about making list of doclets that should always be arrays, but there doesn't seems to be any real source that would be definitive for that - so it potentially would require updates (so more breaking changes and more major version bumps which is very silly).

jquense:
I think having an out of band option to tell which doclets are multiple isn't ideal. This field is already an unstructured and untyped scalar. I agree that that it's not also not ideal that it will change shape as you add and remove properties but that's easy enough for the consumer to guard against. 

Another approach, but still a breaking change could be to make `doclets` an array of something like `{ title: 'property', value: 'foo' }` so you can have as many of a type as you want. This is what `documentationjs` does in it's AST i believe

pieh:
Neither options is really great here, but I think latest suggestion to just use array for everything seems like most predictable of all of them. Let's just do that and do major version bump

jquense:
done 👍

pieh:
Merged and publish

@jquense please verify that changelog entry I added is correct - https://github.com/gatsbyjs/gatsby/blob/f1e539c5dc9075ee55e3770f61c9c08c779eaa21/packages/gatsby-transformer-react-docgen/CHANGELOG.md

PR #10350, opened 2018-12-07
fix(docs): ecosystem section in docs sidebar

amberleyromo:
screen shot 2018-12-07 at 10 34 35 am

- better consistency with tutorial sidebar (intro landing page under "introduction")
- added ecosystem section which links out to starter library, plugin library, and awesome gatsby (external) resources.

### DIFF
--- a/www/src/pages/docs/index.js
+++ b/www/src/pages/docs/index.js
@@ -50,6 +50,10 @@ class IndexRoute extends React.Component {
                   and more.
                 </li>
                 <li>
+                  <strong>Ecosystem</strong>: Check out libraries for Gatsby
+                  starters and plugins, as well as external community resources.
+                </li>
+                <li>
                   <strong>API Reference</strong>: Learn more about Gatsby APIs
                   and configuration.
                 </li>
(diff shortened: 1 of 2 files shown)


www/src/pages/docs/index.js
```diff
@@ -49,6 +49,10 @@ class IndexRoute extends React.Component {
                   around building with Gatsby, like sourcing data, deployment,
                   and more.
                 </li>
+                <li>
+                  <strong>Ecosystem</strong>: Check out libraries for Gatsby
```
DSchau: Not really new here, but is there a reason these aren't linked?

DSchau: I suppose because they're linked in the sidebar? But I'm not sure I see a big issue with doubling up here!

amberleyromo: It's because very few of these section headers are linked to actual pages (so there's nothing to link to).

amberleyromo: The point of including this list in the docs introduction landing page is to introduce the section areas more explicitly (people have a better idea of what they're getting).

DSchau: Got it; that makes sense. Thanks!

PR #10353, opened 2018-12-07
docs: new starter features is required

jaredh159:
Tried to submit a new starter using the template provided in the docs, and got this error:

So, it appears that `features` is in fact required. Updating the template accordingly.


DSchau:
Thanks for updating this!

PR #10220, opened 2018-11-30
feat: update builtin ESLint to v5

stefanprobst:
Update ESLint to v5

PR #10356, opened 2018-12-07
Allow newer `eslint-config-react-app` versions

jgierer12:
Since `eslint-config-react-app@3.0.0-next.66cc7a90` has `eslint^4.1.1` as a peer dependency, npm shows an unmet peer dependency warning if `eslint@5.0.0` or higher is installed.

I don't see a reason to stay on this specific version of the config, so I propose we allow newer versions of `eslint-config-react-app`, which in turn allow `eslint@5`.

### DIFF
--- a/packages/gatsby/package.json
+++ b/packages/gatsby/package.json
@@ -48,7 +48,7 @@
     "domready": "^1.0.8",
     "dotenv": "^4.0.0",
     "eslint": "^4.19.1",
-    "eslint-config-react-app": "3.0.0-next.66cc7a90",
+    "eslint-config-react-app": "^3.0.0-next.66cc7a90",
     "eslint-loader": "^2.0.0",
     "eslint-plugin-flowtype": "^2.46.1",
     "eslint-plugin-graphql": "^2.0.0",


packages/gatsby/package.json
```diff
@@ -48,7 +48,7 @@
     "domready": "^1.0.8",
     "dotenv": "^4.0.0",
     "eslint": "^4.19.1",
```
DSchau: Should we take this opportunity to bump eslint to 5.0.0? Worth investigating to see if our lint setup still passes upgrading both?

packages/gatsby/package.json
```diff
@@ -48,7 +48,7 @@
...
-    "eslint-config-react-app": "3.0.0-next.66cc7a90",
+    "eslint-config-react-app": "^3.0.0-next.66cc7a90",
```
DSchau: May as well just call this `^3.0.0` right?

DSchau: (although they're semantically equivalent, it's probably fine to move to a major version)

pieh:
Related - https://github.com/gatsbyjs/gatsby/pull/10220 - I didn't have time to look into that PR yet, but I think that it is more complete

jgierer12:
Oh, thanks @pieh I didn't find that one when searching for duplicates. I'll close this in favor of #10220

PR #10357, opened 2018-12-08
promise anti-pattern in tutorial-part-7

pobch:
I think in the tutorial part 7, there is [an anti-pattern of Promise.](https://stackoverflow.com/questions/23803743/what-is-the-explicit-promise-construction-antipattern-and-how-do-i-avoid-it) So, I try to fix it in this PR. Moreover, I think there is an alternative solution using `async/await` that can give us the same result.
```javascript
// with async/await
exports.createPages = async ({ graphql, actions }) => {
  const result = await graphql(`
    {
      allMarkdownRemark {
        edges {
          node {
            fields {
              slug
            }
          }
        }
      }
    }
  `)
  result.data.allMarkdownRemark.edges.forEach(({ node }) => {
    createPage({
      path: node.fields.slug,
      component: path.resolve(`./src/templates/blog-post.js`),
      context: {
        // Data passed to context is available
        // in page queries as GraphQL variables.
        slug: node.fields.slug,
      },
    })
  })
}
```

### DIFF
--- a/docs/tutorial/part-seven/index.md
+++ b/docs/tutorial/part-seven/index.md
@@ -177,23 +177,20 @@ exports.onCreateNode = ({ node, getNode, actions }) => {
 
 // highlight-start
 exports.createPages = ({ graphql, actions }) => {
-  return new Promise((resolve, reject) => {
-    graphql(`
-      {
-        allMarkdownRemark {
-          edges {
-            node {
-              fields {
-                slug
-              }
+  return graphql(`
+    {
+      allMarkdownRemark {
+        edges {
+          node {
+            fields {
+              slug
             }
           }
         }
       }
-    `).then(result => {
-      console.log(JSON.stringify(result, null, 4))
-      resolve()
-    })
+    }
+  `).then(result => {
+    console.log(JSON.stringify(result, null, 4))
   })
 }
 // highlight-end
@@ -254,35 +251,32 @@ exports.onCreateNode = ({ node, getNode, actions }) => {
 
 exports.createPages = ({ graphql, actions }) => {
   const { createPage } = actions // highlight-line
-  return new Promise((resolve, reject) => {
-    graphql(`
-      {
-        allMarkdownRemark {
-          edges {
-            node {
-              fields {
-                slug
-              }
+  return graphql(`
+    {
+      allMarkdownRemark {
+        edges {
+          node {
+            fields {
+              slug
             }
           }
         }
       }
-    `).then(result => {
-      // highlight-start
-      result.data.allMarkdownRemark.edges.forEach(({ node }) => {
-        createPage({
-          path: node.fields.slug,
-          component: path.resolve(`./src/templates/blog-post.js`),
-          context: {
-            // Data passed to context is available
-            // in page queries as GraphQL variables.
-            slug: node.fields.slug,
-          },
-        })
+    }
+  `).then(result => {
+    // highlight-start
+    result.data.allMarkdownRemark.edges.forEach(({ node }) => {
+      createPage({
+        path: node.fields.slug,
+        component: path.resolve(`./src/templates/blog-post.js`),
+        context: {
+          // Data passed to context is available
+          // in page queries as GraphQL variables.
+          slug: node.fields.slug,
+        },
       })
-      // highlight-end
-      resolve()
     })
+    // highlight-end
   })
 }
 ```


pieh:
> Moreover, I think there is an alternative solution using `async/await` that can give us the same result.

That would be best, but we still support Node 6, which doesn't have support for async/await. This is also reason that in our docs we don't use `npx` for `gatsby-cli`. We will probably bump minimal Node version to 8 in next major gatsby verion and then we will be able to convert the async code to async/await which I do agree is much nicer.

As for used anti-pattern - this is tough call: reason we use it there is that it very explicitly show that we return Promise in API hook - remember, we are not making any assumptions about expertise level of people reading the tutorial (to the point where we explain a bit what command line is in part 0). We could apply your changes, but then would need to explain that `graphql` function return promise (which would divert a bit from overall flow of that part of tutorial).

docs/tutorial/part-seven/index.md
```diff
@@ -177,23 +177,20 @@ exports.onCreateNode = ({ node, getNode, actions }) => {
...
-  return new Promise((resolve, reject) => {
-    graphql(`
-      {
-        allMarkdownRemark {
-          edges {
-            node {
-              fields {
-                slug
-              }
+  return graphql(`
```
DSchau: After talking to @pieh, I think we are on board with this. I agree that it is kinda weird to show returning a new Promise when we already have a promise returned from the graphql call. We also think that we want to keep these examples as explicit as possible, while balancing the "copy-pastability" of the examples, i.e. keeping them as real world as possible.

As such, let's maybe just add a comment before the graphql invocation that illustrates that the graphql call returns a Promise? e.g.

```suggestion
  // the graphql function call returns a Promise
  // see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise for more info
  return graphql(`
```

docs/tutorial/part-seven/index.md
```diff
@@ -254,35 +251,32 @@ exports.onCreateNode = ({ node, getNode, actions }) => {
...
```
DSchau: Same idea down here. I don't have as strong of an opinion as to whether we need it in both places.

pobch:
Should we also refactor the example on [this page](https://www.gatsbyjs.org/docs/node-apis/#createPages)? I think passing a Promise as `resolve()` argument is not an anti-pattern, but we can make it more readable.

The current code example on the page:
```javascript
    const blogPostTemplate = path.resolve(`src/templates/blog-post.js`)
    // Query for markdown nodes to use in creating pages.
    resolve(
      graphql(
        `
      {
        allMarkdownRemark(limit: 1000) {
          edges {
            node {
              fields {
                slug
              }
            }
          }
        }
      }
    `
      ).then(result => {
        if (result.errors) {
          reject(result.errors)
        }

        // Create blog post pages.
        result.data.allMarkdownRemark.edges.forEach(edge => {
            createPage({
              path: `${edge.node.fields.slug}`, // required
              component: blogPostTemplate,
              context: {
                // Add optional context data. Data can be used as
                // arguments to the page GraphQL query.
                //
                // The page "path" is always available as a GraphQL
                // argument.
              },
            })
        })

        return
      })
    )
  })
}
```

My suggestion:
```javascript
  const blogPostTemplate = path.resolve(`src/templates/blog-post.js`)
  return (
    // Query for markdown nodes to use in creating pages.
    graphql(`
      {
        allMarkdownRemark(limit: 1000) {
          edges {
            node {
              fields {
                slug
              }
            }
          }
        }
      }
    `).then(result => {
      if (result.errors) {
        throw result.errors
      }

      // Create blog post pages.
      result.data.allMarkdownRemark.edges.forEach(edge => {
        createPage({
          path: `${edge.node.fields.slug}`, // required
          component: blogPostTemplate,
          context: {
            // Add optional context data. Data can be used as
            // arguments to the page GraphQL query.
            //
            // The page "path" is always available as a GraphQL
            // argument.
          },
        })
      })
    })
  )
}
```

shannonbux:
Deferring to @pieh and @DSchau on technical accuracy. Just a reminder that in the tutorial we try to focus on teaching:
- make sure whatever we teach works with software versions we support (including node, npm, etc) AND is a best practice, if possible
- if the best practice way to do something does not work with software versions we support, then include best practice as a special note, but focus on teaching something all tutorial readers will be able to accomplish

Also a reminder that our audience is front-end devs who could be beginners, and are interested in JavaScript and React, and know enough to make it through the tutorial.

KyleAMathews:
Seems reasonable to me. Agree no reason to create a promise given GraphQL already makes one.

PR #10368, opened 2018-12-09
Fix Starter Library URL and update copy

rgoldfinger:
Previously the "starter list" linked to a page that wasn't found. This fixes it and updates the copy to reflect the current name.


shannonbux:
Thank you @rgoldfinger!

PR #10380, opened 2018-12-10
Add satispay.com to showcase

adnsio:
I'd like to add satispay.com to showcase list.

### DIFF
--- a/docs/sites.yml
+++ b/docs/sites.yml
@@ -3723,3 +3723,14 @@
     - Business
     - Food
     - Beverage
+- title: Satispay
+  url: https://www.satispay.com
+  main_url: https://www.satispay.com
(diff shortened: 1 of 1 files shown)


docs/sites.yml
```diff
@@ -3723,3 +3723,14 @@
     - Business
     - Food
     - Beverage
+- title: Satispay
+  url: https://www.satispay.com
+  main_url: https://www.satispay.com
+  categories:
+    - Business
+    - Finance
+    - Fintech
```
LekoArts: Please remove this tag, as it's a duplication of `Finance`

```suggestion
```

adnsio: Done 👍

PR #10385, opened 2018-12-10
feat(gatsby-cli): Add `gatsby issue` command

### LINKED ISSUE #10095: Open issues on gatsbyjs/gatsby repo via CLI
## Summary

Add a command to the `gatsby` CLI that allows interactively opening bug reports, feature requests and questions via this repo's issues (as mentioned in #6729). I would love to work on this myself, but wanted to open an issue first to check if no one else is already working on this, and if everyone agrees with my approach.

### Basic example

<details>
<summary>Show example interaction</summary>

```console
$ gatsby issue
? Do you want to open a Bug Report, Feature Request, or Question?
> Bug Report 🐞
  Feature Request 💡
  Question 🤔
```

```console
$ gatsby issue
? Do you want to open a Bug Report, Feature Request, or Question? Bug Report 🐞
? What is the title of your bug report: Example issue
```

```console
$ gatsby issue
? Do you want to open a Bug Report, Feature Request, or Question? Bug Report 🐞
? What is the title of your bug report? Example issue
! We found 3 issues that seem similar to yours:
- Some issue (https://github.com/gatsbyjs/gatsby/issues/1)
- Another issue (https://github.com/gatsbyjs/gatsby/issues/2)
- Third issue (https://github.com/gatsbyjs/gatsby/issues/3)
? Does one of the above match the issue you're seeing? (Y/n) n
```

```console
$ gatsby issue
? Do you want to open a Bug Report, Feature Request, or Question? Bug Report 🐞
? What is the title of your bug report? Example issue
? Describe the issue you're seeing: ...
```

.....

```console
$ gatsby issue
? Do you want to open a Bug Report, Feature Request, or Question? Bug Report 🐞
? What is the title of your bug report? Example issue
? Describe the issue you're seeing
? Clear steps describing how to reproduce the issue
? Link to a demo project
? Actual result: ...
```

```console
$ gatsby issue
✔️ All done! We opened a page in your browser where you can double-check all your
information and submit the issue.

If the page did not open automatically, please open the following URL in your browser:
https://github.com/gatsbyjs/gatsby/issues/new?...
```

</details>

### Motivation

This will allow seamlessly opening an issue as soon as the user discovers it, without switching context away from the development environment. It guides users through the required details step by step, ensuring that they don't forget any important information. Additionally, this can automatically add environment data (`gatsby info`) as well as recent error logs.

### Implementation

The new command will use [Inquirer.js](https://github.com/SBoudrias/Inquirer.js) to create an interactive issue prompt. Depending on the chosen issue type, different questions will be asked. We might also be able to list potential duplicate issues by searching for the specified title via GitHub's API.

In case the user selected the "Bug Report" issue type, we collect environment information (`gatsby info`) as well as recent logs (e.g. `~/.npm/_logs/01234-debug.log`) and add them to the issue body in a `<details>` tag.

Finally, we generate the issue URL with [`new-github-issue-url`](https://github.com/sindresorhus/new-github-issue-url) and try to open it in the user's browser with [`opn`](https://github.com/sindresorhus/opn).

### Challenges

- I have never worked with Inquirer.js yet. The API seems quite friendly though so I think I'll manage it.
- I'm not sure yet how to determine the log files which should be included - any idea would be greatly appreciated!
- Don't know if my idea of finding duplicates via GitHub's API will work out. Perhaps simply searching for the title will not be good enough, or we will get rate limited too quickly.

jgierer12:
Closes #10095 

Currently, only envinfo (same as `gatsby info`) and content of  Gatsby config files is collected. All data is simply appended as plain text to the issue body.

## Todo:
- [x] Add base command and templates
- [x] Collect Gatsby config files
- [ ] Collect `.cache` contents
- [ ] Collect logs
- [ ] Remove sensitive data (env variables) from collected data
- [ ] Docs

## Known problems:
- [ ] URL printed to the console doesn't contain line breaks
- [ ] URL can get very long with large config files. Perhaps use a link shortener?
- [ ] When we collect more data, it will probably be necessary to upload it somewhere as a tarball instead of adding it as plain text


jgierer12:
Seems like the CI hasn't updated `prettier` to [1.15](https://prettier.io/blog/2018/11/07/1.15.0.html#flatten-else-branch-for-nested-ternaries-5039-by-suchipi-5272-by-duailibe-5333-by-ikatyang) yet. Should I format it with 1.14 or leave it as is?

DSchau:
> Seems like the CI hasn't updated prettier to 1.15 yet. Should I format it with 1.14 or leave it as is?

You should always use what's in the repo, rather than the globally installed version of prettier. This resolves that idiosyncrasy--otherwise we'd have "out of scope" changes in a PR, e.g. updating prettier to get lint tests passing.

packages/gatsby-cli/src/issue/telemetry.js
```diff
@@ -0,0 +1,81 @@
...
+const mdCodeBlock = (content, lang) =>
+  `
+\`\`\`${lang}
+${content}
+\`\`\`
+  `.trim()
+
+module.exports = answers => {
+  if (!answers.envinfoConfirmation && !answers.filesConfirmation) {
+    return new Promise(resolve => resolve(``))
```
DSchau:
```suggestion
    return Promise.resolve(``)
```

packages/gatsby-cli/src/issue/telemetry.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  const promises = []
```
DSchau: nit: but I like to call this `let` since it's a signal that this will be mutated--even if it's not re-assigned.

Feel free to keep this as is though :)

packages/gatsby-cli/src/issue/telemetry.js
```diff
@@ -0,0 +1,81 @@
...
+
+  if (answers.filesConfirmation) {
+    const files = [
+      `gatsby-browser.js`,
+      `gatsby-config.js`,
+      `gatsby-node.js`,
+      `gatsby-ssr.js`,
+    ]
+
+    files.forEach(fileName => {
```
DSchau: We can do this with a `.map` a little more cleanly, I think

```js
promises = promises.concat(
  files.map(file => (
    // the contents you already had in the forEach, i.e. what was pushed
  )
)
```

packages/gatsby-cli/src/issue/telemetry.js
```diff
@@ -0,0 +1,81 @@
...
+      promises.push(
+        fs
+          .readFile(path.join(process.cwd(), fileName))
+          .then((result, error) => {
```
DSchau: The error argument passed to the then handler is a callback function containing the error, not a simple value. See [this MDN document](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then#Syntax)

I think it's probably a bit cleaner to handle the success case (with a `then`) and the failure case with a `.catch`. It will also simplify the `const content = ` creation there, since that's a bit complex to parse. 

What do you think?

packages/gatsby-cli/src/issue/telemetry.js
```diff
@@ -0,0 +1,81 @@
+const path = require(`path`)
+const fs = require(`fs-extra`)
+
+const envInfo = require(`../env-info.js`)
+
+const mdCodeBlock = (content, lang) =>
+  `
+\`\`\`${lang}
+${content}
```
DSchau: I've solved this problem differently before, e.g.

```js
const mdCodeBlock = (content, lang) => [
  '```' + lang,
  content,
  '```'
].join('\n').trim()
```

packages/gatsby-cli/src/issue/index.js
```diff
@@ -0,0 +1,42 @@
...
+  type: `list`,
+  name: `issueType`,
+  message: `Do you want to open a Bug Report, Feature Request, or Question?`,
+  choices: Object.entries(issueTypes).map(([value, { prettyName: name }]) => {
+    return { value, name }
+  }),
+}
+
+module.exports = () => {
+  inquirer.prompt(issueTypeQuestion).then(({ issueType }) => {
```
DSchau: You could clean this up a little with `async/await`

```js
module.exports = async function createIssue() {
  const { issueType } = await inquirer.prompt(issueTypeQuestion);
  const answers = await inquirer.prompt(issueTypes[issueType].questions);

  // etc.
}
```

packages/gatsby-cli/src/issue/telemetry.js
DSchau: (also this was used once elsewhere at least, so take another look at 
any `.then((value, error)` patterns, as I'd recommend refactoring to explicit `.then(value)` and `.catch(error)` patterns.

jgierer12: Oops, not sure how i missed that when testing :sweat_smile: 

I will use `.then`/`.catch`, thanks.

pieh:
> That's the problem: The prettier version in the repo (`^1.14.3`) _does_ resolve to `1.15.3`, but it seems to still be at `1.14.x` on the CI (based on the error it reports).

`yarn.lock` should handle that and lock `prettier` to version specified there

packages/gatsby-cli/src/issue/telemetry.js
jgierer12: Since eslint wants all strings to be template strings anyways, I figured my solution would be cleaner than

```js
const mdCodeBlock = (content, lang) => [
  `\`\`\`` + lang,
  content,
  `\`\`\``
].join(`\n`).trim()
```

But of course I can also use yours, and maybe turn off the eslint rule for that function.

jgierer12:
Thanks @pieh, using `yarn` instead of `npm` worked.

packages/gatsby-cli/src/issue/telemetry.js
DSchau: I agree with you here. Keep what you had!

KyleAMathews:
@jgierer12 do you want to keep working on this? This isn't a top priority for us as it seems a lot of issues are being created already and this might need a decent bit of polish to get useful. Since it's been sitting for a while, I'm going to close it but feel free to open a new one. I'd aim for a simpler PR to start with that just adds env info as that'll be easier to get in.

jgierer12:
I'm currently a bit busy and don't have a lot of time for open source, so I won't be able to work on this for a while. I'd be happy to continue once I have a bit more free time though!

PR #10389, opened 2018-12-10
fix(www): avoid querying for no-cache=1

Haroenv:
overall this could also have been done by parsing & stringifying with qs, but this should work too.

I haven't been able to try this out yet (I didn't have the patience to wait for all the images to transform ahah), but I don't think this can break anything.

Changes done:

1. small refactors to make code a little shorter
2. replace `no-cache=1` from the query before searching. 

I noticed when seeing a lot of `o-cache=1` hits on Algolia analytics

### DIFF
--- a/www/src/components/plugin-searchbar-body.js
+++ b/www/src/components/plugin-searchbar-body.js
@@ -174,10 +174,6 @@ injectGlobal`
 
 // Search shows a list of "hits", and is a child of the PluginSearchBar component
 class Search extends Component {
-  constructor(props, context) {
-    super(props)
-  }
-
   render() {
     return (
       <div
@@ -253,7 +249,7 @@ class Search extends Component {
                 <Result
                   hit={result.hit}
                   pathname={this.props.pathname}
-                  search={this.props.searchState}
+                  query={this.props.query}
                 />
               )}
             />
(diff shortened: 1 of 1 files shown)


pieh:
> I haven't been able to try this out yet (I didn't have the patience to wait for all the images to transform ahah), but I don't think this can break anything.

You can run `GATSBY_SCREENSHOT_PLACEHOLDER=true gatsby develop` which will skip pulling screenshots for site/starter showcases (which is like 75% of all the images on gatsbyjs.org)

pieh:
We did remove `no-cache` stuff recently (moved to different implementation) - do you still get when browsing gatsbyjs.org?

Haroenv:
I don't know how the URL gets there, but I see recent calls to that query in Algolia: `o-cache=1` is the #1 query without results on the index over the last week still. Also last day still.

btk:
@pieh I come across this problem while coming from google.

Seems like google has indexed some links with this query. After the new implementation its not relevant but, this pr might be helpful.

screenshot at dec 12 03-50-49

Haroenv:
I could make a small implementation that serialises to `?q=xxx`. Which browsers are supposed to be supported in this part of the site, is `new Map(new URLSearchParams(location.search))` acceptable (only latest browsers), or would a more compatible (i.e. using qs or manual parsing) way be preferred

pieh:
`react@16` requires Map and Set, so it would work without extra polyfills (because they are already polifilled), but I don't think we need to do anything more complex here. Did quick tests and works as expected, so I'll merge as-is.

PR #10395, opened 2018-12-10
feat(gatsby-remark-copy-linked-files): add support for video elements with `src` attribute

janmonschke:
This adds support for video elements that only have a `src` attribute.

```html
<video src="myvideo.mp4"></video>
```

### DIFF
--- a/packages/gatsby-remark-copy-linked-files/src/index.js
+++ b/packages/gatsby-remark-copy-linked-files/src/index.js
@@ -234,7 +234,7 @@ module.exports = (
 
     // Handle video tags.
     const videoRefs = []
-    $(`video source`).each(function() {
+    $(`video source, video`).each(function() {
       try {
         if (isRelativeUrl($(this).attr(`src`))) {
           videoRefs.push($(this))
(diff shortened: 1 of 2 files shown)


packages/gatsby-remark-copy-linked-files/src/index.js
```diff
@@ -234,7 +234,7 @@ module.exports = (
 
     // Handle video tags.
     const videoRefs = []
-    $(`video source`).each(function() {
+    $(`video source, video`).each(function() {
```
pieh: How about:
```suggestion
    $(`video source[src], video[src]`).each(function() {
```
to target elements that actually have `src` attribute?

With `video source, video` selector we would push a lot of `undefined` to `videoRefs` array:

```
<video>
  <source type="video/mp4" src="video.mp4"></source>
</video>
```
outer `<video>` would be targeted by selector, but it doesn't have src so it doesn't make sense to push it to array, right?

janmonschke: Ah yes, that's a great idea. I looked at the code again and saw that there is quite some duplication in that file. Would you be open if I refactor that file a bit?

pieh: Absolutely open to that! After you mentioned it I just did quick browse of entire file (instead of just zooming in on area your changes) and see that the handling of video, audio and anchor tags is essentially copy'n'pasted and could be DRYed a bit.

Issue #10373, opened 2018-12-09
Document stub node "Client data fetching" with relevant information


tusharjoshi:
Provide relevant documentation to change this stub into actual document for the section [Client data fetching](https://github.com/gatsbyjs/gatsby/blob/master/docs/docs/client-data-fetching.md)

Located at [Site URL here] (https://www.gatsbyjs.org/docs/client-data-fetching/#client-data-fetching)

Recently there is one [tweet](https://twitter.com/IAmKale/status/1071108771684921344) regarding the state of documentation for this section.

tusharjoshi:
I need information about the context of Client data fetching which should be documented on this page.  Any pointers bout youtube videos, blog posts, code comments or any place where this functionality is discussed will help.  

Is this page title referring to fetching data after the site is built and deployed?  Or this is at the time of building the site.  This node is under section "Use Cases for Gatsby" hence not sure, any information about these from the developers will help me proceed.

btk:
I would say the title is not so clear, If I wanted to write something under that title. I probably wouldn't. The stubs should explain the context of that page with at least one paragraph I think. 

It does not look very inviting to edit or contribute to be honest.

tusharjoshi:
Currently 6 nodes under the section [Use cases with Gatsby](https://www.gatsbyjs.org/docs/building-apps-with-gatsby/) are stubs.  If this title is unclear then we should remove that stub or update it so someone can contribute to make it complete.

I am very new to Gatsby and cannot comment on what shall be the correct title, need help from Gatsby veterans here.

amberleyromo:
Hi @tusharjoshi -- great observation, thank you! We're always trying to expand and cover as much as we can in docs. In this case, we're removing some of these stubs from the sidebar, until we can better define the question we need to answer.

Can you go into more specifics about the problem you were trying to solve, that led you to this doc page?

tusharjoshi:
Hi @amberleyromo Recently there is one [tweet here](https://twitter.com/IAmKale/status/1071108771684921344) regarding the state of documentation for this section.  As I watch @gatsbyjs tweeter feed I came across that tweet and wanted to provide solution by contributing here.

This is [24pullrequest project](https://24pullrequests.com/) month and I keep searching for small contributions which can be done in opensource projects.  Thanks.

shannonbux:
[...]

PR #10397, opened 2018-12-11
Docs: Rm stubs from sidebar

amberleyromo:
@shannonbux @DSchau.

Just rms some of the incomplete stubs we discussed from sidebar. Shannon, I also removed some from under Images. As discussed, left the actual stub pages intact.

### DIFF
--- a/www/src/data/sidebars/doc-links.yaml
+++ b/www/src/data/sidebars/doc-links.yaml
@@ -189,20 +189,8 @@
     - title: Use cases with Gatsby (including apps)
       link: /docs/building-apps-with-gatsby/
       items:
-        - title: Client data fetching*
-          link: /docs/client-data-fetching/
         - title: Building a site with authentication
           link: /docs/building-a-site-with-authentication/
-        - title: Building an e-commerce site*
-          link: /docs/building-an-e-commerce-site/
-        - title: Building a site with asynchronous data*
-          link: /docs/building-a-site-with-asynchronous-data/
-        - title: Building a portfolio*
-          link: /docs/building-a-portfolio/
-        - title: Building a blog*
-          link: /docs/building-a-blog/
-        - title: Building an image-heavy site*
-          link: /docs/building-an-image-heavy-site/
     - title: Adding website functionality
       link: /docs/adding-website-functionality/
       items:
--- a/www/src/data/sidebars/doc-links.yaml
+++ b/www/src/data/sidebars/doc-links.yaml
@@ -62,14 +62,6 @@
           link: /docs/adding-images-fonts-files/
         - title: Static folder
           link: /docs/static-folder/
-        - title: Dropping images into static folders*
-          link: /docs/dropping-images-into-static-folders/
-        - title: Importing single files*
-          link: /docs/importing-single-files/
-        - title: Importing media content*
-          link: /docs/importing-media-content/
-        - title: gatsby-source-filesystem programmatic import*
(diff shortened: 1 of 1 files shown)


KyleAMathews:
Perhaps we could have a page in each section (or in its index page) that lists docs pages we'd like to add, links to their markdown files, and talks about the process for adding additional documentation?

KyleAMathews:
A lot of the stubs _have_ led to contributions so just removing them altogether seems sub-optimal. Still expose them but make it clear they're not interesting.

I don't want us to overreact to one grumpy tweet.

amberleyromo:
@KyleAMathews this is not removing all of the stubs. It's just removing some of them from being linked in the sidebar. We had a separate conversation in Slack that touched led to the conclusion that if we want a contributor to pick up a stub, it needs to be more tightly scoped / described (a valid request). In the meantime, before @shannonbux can revisit, we decided to remove some of them from the sidebar.

amberleyromo:
IMO we're not overreacting to a grumpy tweet, we discussed the context around a piece of valid feedback (that happened to be in the form of a grumpy tweet 😉) and decided how we wanted to approach it.

KyleAMathews:
Sorry for not being clear about my critique — my point is that issues in PRs have a lot less visibility than when they're exposed somehow within the docs. Issues get 100x (more or less) less visitors than the website.

Exposing prospective docs pages in the sidebar a) can lead to new contributors and b) lets people know that Gatsby's docs are actively growing and improving.

I totally agree that issues are a good thing as we found with Oktoberfest as it makes it easier for someone to claim work just don't want to lose value from the other reasons we've had stub docs.

amberleyromo:
Sure, that's a valid critique. I was not sure of @shannonbux's intent with some of these docs, and she doesn't remember making some of these stubs; Therefore she could not speak to the context/intent behind them. Combined with the fact that the docs should reflect her research findings, I thought this was a fair way to proceed (if she doesn't remember making them, and they're inadequately scoped, why leave them there?) I removed the parts under images as well, because I previously had a similar conversation about that section during Hacktoberfest.

The main thought I'll leave here is that I think any doc that's up as a stub should be formatted such that A) the intent is clear B) it's scoped for a contributor to easily pick up (if they choose).

https://github.com/gatsbyjs/gatsby/issues/10373#issuecomment-445603250
> I would say the title is not so clear, If I wanted to write something under that title, I probably wouldn't. The stubs should explain the context of that page with at least one paragraph I think.
> 
> It does not look very inviting to edit or contribute to be honest.

I do agree with this feedback.

Overall tl;dr:  I'll leave this up to @shannonbux since it's her area.

shannonbux:
@KyleAMathews I'm fairly certain that in this whole year, only the testing section of stubs has ever been picked up by contributors outside of the content migration project, which we managed during Hacktoberfest (I could be wrong, would need to measure this). So, we reasoned that we need to replicate the Hackborfest style of issues for each stub article in order for them to every get finished. I don't care too much whether the stub articles stay up right now or not. Perhaps we could put them back up after we've linked to the issue that describes how to fill out the stub article?

KyleAMathews:
I'm not overly concerned about how we do things but some way of indicating to people planned additions and asking for help to write them is important. Perhaps even just having one page that lists docs we'd like written would work.

It'd be cool if there's a way for people to vote as well which we could do once we add GitHub logins.

shannonbux:
Yeah that sounds awesome @kyleamathews. I'm working on issues for all stub
articles. For now, will put them in an overview issue or epic to see how
that goes.

PR #10406, opened 2018-12-11
fix(docs): Environment Variables Examples

LekoArts:
I followed the guide yesterday and stumbled across the issue that the guide is for using env vars in the browser is wrong. After prefixing the vars both in the `.env.*` and frontend files it worked, e.g.:
https://github.com/LekoArts/gatsby-source-tmdb-example/blob/master/src/components/DetailedView.jsx#L264

I added a little example to the client-side part, moved the comments to code block titles and expanded the example while fixing it.

### DIFF
--- a/docs/docs/environment-variables.md
+++ b/docs/docs/environment-variables.md
@@ -81,6 +81,8 @@ Now the variables are available on `process.env` as usual.
 
 ## Example
 
+Please note that you shouldn't commit `.env.*` files to your source control and rather use options given by your CD provider (e.g.  Netlify with its [build environment variables](https://www.netlify.com/docs/continuous-deployment/#build-environment-variables)).
+
 ```shell:title=.env.development
 GATSBY_API_URL=https://dev.example.com/api
 API_KEY=927349872349798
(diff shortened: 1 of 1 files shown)


docs/docs/environment-variables.md
```diff
@@ -77,31 +81,45 @@ Now the variables are available on `process.env` as usual.
...
   return (
     <div>
-      <img src={`${process.env.API_URL}/logo.png`} alt="Logo" />
+      <img src={`${process.env.GATSBY_API_URL}/logo.png`} alt="Logo" />
     </div>
   )
 }
 ```
 
+`API_KEY` will be available to your site (Server-side) as `process.env.API_KEY`:
```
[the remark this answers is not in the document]
pieh: ☝️ is actually verified by https://github.com/gatsbyjs/gatsby/blob/0aade74a1862f7be9875e8f3d84eb049de132c10/e2e-tests/production-runtime/.env.production / https://github.com/gatsbyjs/gatsby/blob/0aade74a1862f7be9875e8f3d84eb049de132c10/e2e-tests/production-runtime/cypress/integration/production.js#L75-L81

docs/docs/environment-variables.md
```diff
@@ -47,6 +47,10 @@ In addition to these Project Environment Variables defined in `.env.*` files, yo
 OS Env Vars. OS Env Vars which are prefixed with `GATSBY_` will become available in
 browser JavaScript.
 
+```shell:title=.env.*
```
shannonbux: oh, does the name of this file actually include an asterisk?

LekoArts: The asterisk means that all names are possible (it's a common way to write it like that), that's why it's already used in line 46.

LekoArts:
@pieh Rewrote the paragraph

docs/docs/environment-variables.md
DSchau: Yeah, @LekoArts I think this is reasonable.

@shannonbux for clarification, we typically use `.env.development` and `.env.production` since those credentials could (and often do) change between environments.

PR #10412, opened 2018-12-11
add sentryio plugin for Gatsby

mhadaily:
Why? 
There is already a plugin for Sentry, however, it seems maintainer doesn't support the plugin anymore and the community needs the latest version as well as better support.
I have already contributed here https://github.com/octalmage/gatsby-plugin-sentry/pull/11 , however, based on suggestions, It would be nice if we can have this plugin under Gatsby repo. 

How?
I have created sentryio plugin which supports the latest version.


pieh:
If original maintainer doesn't support plugin (for whatever reason), best course of action would be if he could pass ownership of the package to someone that actively uses the plugin and that's first option I'd like to see explored here. 

@octalmage what are your thoughts on this?

I hesitate to bring yet another plugin here to maintain - 1) PR feedback loop already takes longer than we would like and adding 2) I can't speak for others by I have very limited experience with Sentry - at very least we would need some kind of test site to verify changes to the plugin.

octalmage:
I am still supporting the plugin, just when I have time. I’ve merged the PR.

octalmage:
I’ll leave this open for another matinainer to decided what to do.

pieh:
> I am still supporting the plugin, just when I have time. I’ve merged the PR.

That's awesome to hear! We definitely do know that maintenance is time consuming, so I really appreciate handling that PR 👍  

> I’ll leave this open for another matinainer to decide what to do.

As things are resolved, there is no point and maintaining 2 versions of same plugin, so I think it's safe to close the PR.

PR #10416, opened 2018-12-11
fix(gatsby-plugin-offline): Remove slash from `/static` + move `sw-append` out of `src`

### LINKED ISSUE #10036: gatsby-plugin-offline: SW isn't always updated on each build, resulting in inconsistent state
This is really hard to reproduce but I'm googling gatsby enough that I see this a couple times a week.
When opening a gatsbyjs.org page from google in a new tab, sometimes (rarely) the full page is blank and remains blank until a manual browser refresh.
I cleared my service worker and full site cache on the site last week thinking it was just a leftover early v2 bug but I'm still seeing it this week.

Screenshot:
screen shot 2018-11-19 at 6 12 48 pm

davidbailey00:
This prevents static resources incorrectly being revisioned by Workbox and prevents Babel compiling `sw-append.js`.

### DIFF
--- a/packages/gatsby-plugin-offline/.gitignore
+++ b/packages/gatsby-plugin-offline/.gitignore
@@ -1,3 +1,4 @@
 /*.js
 !index.js
+!sw-append.js
 yarn.lock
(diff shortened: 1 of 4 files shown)


pieh:
using `/static` is causing some weirdness with Workbox? Is there easy way to validate this?

davidbailey00:
> using /static is causing some weirdness with Workbox? Is there easy way to validate this?

e.g. look at https://www.gatsbyjs.org/sw.js - see that files with URLs beginning with "static" have revisions. This shouldn't happen because their filenames are hashed anyway.

davidbailey00:
Note that Workbox only matches same-origin resources if the regex doesn't match the beginning of the URL, so we won't accidentally cause any 3rd party URLs to be cached by doing this.

PR #10417, opened 2018-12-11
docs(gatsby-plugin-offline): Show how to handle a SW update

### LINKED ISSUE #9087: Document how to detect SW updates and alert users
I am using the gatsby-plugin-offline plugin and would like to know if there is a way to handle updates.

I would like to display a message if a new version is available like 
https://medium.com/progressive-web-apps/pwa-create-a-new-update-available-notification-using-service-workers-18be9168d717

At the moment I need to hard reload the site in order to see changes. 

How to handle this?

Thanks in advance!

davidbailey00:
Closes #9087

I also deleted `/docs/add-offline-support/` since it just redirects to `/docs/add-offline-support-with-a-service-worker/` anyway, and hasn't been updated in a while.


packages/gatsby/src/utils/api-browser-docs.js
```diff
@@ -206,6 +206,10 @@ exports.onServiceWorkerInstalled = true
  * Inform plugins of when a service worker has an update available.
  * @param {object} $0
  * @param {object} $0.serviceWorker The service worker instance.
+ * @example
+ * exports.onServiceWorkerUpdateFound = () => {
+ *   window.alert(`An updated service worker was found.`)
```
DSchau: We generally want to keep these examples relatively useful (e.g. users will probably directly copy and paste this).

Is this the best example we could show? I'm almost inclined to suggest reverting it back to the way it was, but minimally I think perhaps a prompt like you've shown earlier is more effective?

docs/docs/add-offline-support-with-a-service-worker.md
```diff
@@ -27,22 +27,39 @@ Add this plugin to your `gatsby-config.js`
...
 }
 ```
 
 That's all you need to add offline support to your Gatsby site.
 
 Note: Service worker registers only in production builds (`gatsby build`).
 
+### Displaying a message when a service worker updates
+
+To display a custom message once your service worker finds an update, you can use the [onServiceWorkerUpdateFound](/docs/browser-apis/#onServiceWorkerUpdateFound) browser API in your `gatsby-browser.js` file. The following code will display a prompt asking the user whether they would like to refresh the page when an update is found:
```
DSchau:
```suggestion
To display a custom message once your service worker finds an update, you can use the [`onServiceWorkerUpdateFound`](/docs/browser-apis/#onServiceWorkerUpdateFound) browser API in your `gatsby-browser.js` file. The following code will display a prompt asking the user whether they would like to refresh the page when an update is found:
```

shannonbux:
@davidbailey00 looks great! Thanks for noticing the redundancy too; I thought I had fixed that and clearly hadn't! Awesome!

PR #10419, opened 2018-12-11
fix(gatsby-image): TypeScript - Mark more props as optional

juliettepretot:
It seems like these should not be a mandatory. At the moment when building a TypeScript gatsby site, TypeScript forces the user to pass a value.


juliettepretot:
@pieh Excellent point, updated :)

PR #10432, opened 2018-12-12
feat(gatsby-plugin-offline): reload when missing resources and SW was updated + add "onServiceWorkerUpdateReady" API

### LINKED ISSUE #10036: gatsby-plugin-offline: SW isn't always updated on each build, resulting in inconsistent state

davidbailey00:
Fixes #10036


davidbailey00:
@pieh (summary of what I said on Slack): the current API `onServiceWorkerUpdateFound` runs even for initial install - see `register-service-worker.js`. I think it could be beneficial to have an API which users can employ e.g. to show a prompt only when an update is found, but not for initial install.

davidbailey00:
@pieh this site https://t.co/GT7MCm13KT is a good example of why we need a separate API - upon first visit it says "this site has updated"

packages/gatsby-plugin-offline/sw-append.js
```diff
@@ -12,22 +12,7 @@ const navigationRoute = new workbox.routing.NavigationRoute(({ event }) => {
...
-              )
-            } else {
-              return fetch(event.request)
-            }
-          })
-        }
-
-        return cachedResponse
-      })
+      return caches.match(offlineShell, { cacheName })
```
pieh: I don't think it's safe to remove that yet. This handles stuff before runtime actually run, so we can't really rely that any changes we make in runtime are actually in effect (not to mention that people can use "missmatched" versions of gatsby and offline plugin in this could "brick" their sites still in those still not figured out scenarios)

packages/gatsby/cache-dir/register-service-worker.js
```diff
@@ -23,8 +23,14 @@ if (
...
                 // have been added to the cache.
-                // We set a flag so Gatsby Link knows to refresh the page on next navigation attempt
-                window.GATSBY_SW_UPDATED = true
+                // We call the onServiceWorkerUpdateReady API so users can show update prompts.
+                apiRunner(`onServiceWorkerUpdateReady`, { serviceWorker: reg })
+
+                // If resources failed for the current page, reload.
+                if (window.___failedResources) {
+                  console.log(`resources failed, SW updated - reloading`)
+                  window.reload()
```
pieh: shouldn't this be:
```suggestion
                  window.location.reload()
```
?

wardpeet: oh yeah it should!

packages/gatsby-plugin-offline/package.json
```diff
@@ -34,8 +35,10 @@
   },
   "repository": "https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-offline",
   "scripts": {
-    "build": "babel src --out-dir . --ignore **/__tests__",
+    "build": "npm run build:src && npm run build:sw-append",
+    "build:src": "babel src --out-dir . --ignore **/__tests__",
```
pieh: do we need to ignore `sw-append.js` here? for builds it probably doesn't matter - it will be overwritten by cpx in next step, but while watching, babel will try to compile it?

packages/gatsby/cache-dir/register-service-worker.js
```diff
@@ -23,8 +23,14 @@ if (
...
                 // At this point, the old content will have been purged and the fresh content will
                 // have been added to the cache.
-                // We set a flag so Gatsby Link knows to refresh the page on next navigation attempt
-                window.GATSBY_SW_UPDATED = true
+                // We call the onServiceWorkerUpdateReady API so users can show update prompts.
+                apiRunner(`onServiceWorkerUpdateReady`, { serviceWorker: reg })
+
+                // If resources failed for the current page, reload.
+                if (window.___failedResources) {
+                  console.log(`resources failed, SW updated - reloading`)
```
pieh: There might be scenarios where resources don't fail (keeping `public` from previous builds and deploying both new and old files), but we would still want to reload on SW update?

packages/gatsby/cache-dir/ensure-resources.js
```diff
@@ -129,6 +129,8 @@ class EnsureResources extends React.Component {
 
   render() {
     if (!this.hasResources(this.state.pageResources) && isInitialRender) {
+      window.___failedResources = true
```
pieh: there are also other places to set this flag - for example - https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/cache-dir/loader.js#L55-L60

packages/gatsby-plugin-offline/package.json
davidbailey00: ah yep, that's a good point - need to fix that

packages/gatsby/cache-dir/register-service-worker.js
davidbailey00: I think the idea is we should only reload if necessary to prevent breaking user flow - the new API allows users to force a reload on update if they want to do so.

packages/gatsby/cache-dir/ensure-resources.js
davidbailey00: I don't think we should set the flag anywhere else - resources may fail from some other part, but ultimately it only matters if they fail from `EnsureResources` since this is the "final step" before actually displaying a page. Does that make sense or is there still a need to set it elsewhere?

packages/gatsby/cache-dir/register-service-worker.js
davidbailey00: yes, thanks for spotting! I made this same mistake a while ago with the docs for handling SW updates, oops :smile:

packages/gatsby/cache-dir/ensure-resources.js
pieh: Right, this make sense in this context. We will reload on navigation if page resources failed to fetch

pieh:
Was testing this over the weekend and it seems like nice improvement - especially part that reloads page if there are missing resources and service worker did update
screenshot 2019-01-28 at 13 20 04

pieh:
Just need to bump `peerDependency` in offline plugin as we do changes in both gatsby and offline plugin here, but I will handle that

sgnl:
Has anyone experieced this "missing recourses" error on a gatsby site without PWA or using `gatsby-plugin-offline`?

davidbailey00:
@sgnl Please could you open an issue describing this behaviour if you're still having it? Thanks :slightly_smiling_face: Just a heads up, sometimes it can be caused by adblockers so if you're using one, try disabling it and see if you get the same issue.

PR #10433, opened 2018-12-12
Use correct function in example code

nikoladev:
`window.prompt()` doesn't return a boolean value, `window.confirm()` does :)

See https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm


LekoArts:
Hi 👋 
Thanks for the PR! However David already opened up a PR for that:
https://github.com/gatsbyjs/gatsby/pull/10431

Therefore I'm going to close this

PR #10434, opened 2018-12-12
Fix type of dependency to dev

enzoferey:
As it is documented on the preset READMe ([`babel-preset-gatsby`](https://github.com/gatsbyjs/gatsby/tree/master/packages/babel-preset-gatsby)), it should be installed as a devDependency.

### DIFF
--- a/docs/docs/babel.md
+++ b/docs/docs/babel.md
@@ -25,7 +25,7 @@ Gatsby ships with a default .babelrc setup that should work for most sites. If y
 to add custom Babel presets or plugins, you can create your own `.babelrc` at the root of your site, import [`babel-preset-gatsby`](https://github.com/gatsbyjs/gatsby/tree/master/packages/babel-preset-gatsby), and overwrite the `target` option.
 
 ```bash
-npm install --save babel-preset-gatsby
+npm install --save-dev babel-preset-gatsby
 ```
 
 ```json5:title=.babelrc


pieh:
I don't think this should as `dev` dependency - if someone would `npm install --production`, it wouldn't be installed. It would probably still work because `babel-preset-gatsby` would be installed by `gatsby` package, so it would work as transitive dependency.

enzoferey:
As far as I know, all the babel related stuff if always installed as a devDependency (except `@babel/plugin-transform-runtime`).

How different is `babel-preset-gatsby` than any other babel preset ? Looking at its source it just reuses a set of other presets and plugins with some specific configs...

DSchau:
@enzoferey there are some considerations here. I generally agree with you that it _probably_ makes sense to make this a devDep, but as @pieh notes there are some edge cases, particularly around the `production` flag. We want to minimize the amount of those occurring for users, because that's just never a fun thing to deal with!

That being said--let's merge this, because I do think this is _technically_ more correct. `babel` and other build-time tooling is generally considered a devDependency and is not required for your production, runtime code.

enzoferey:
Great @DSchau ! For me it was more a story of finding out it's proper placement, it doesn't make sense to label it as a dependency on the website and as a devDepedency on Github 😄 

I understand that it could cause hard to spot errors, but aren't other babel presets/plugins marked as devDependencies in the same position ? 

I mean, if instead `babel-preset-gatsby` it's just `babel-preset-react`, it would be missing as well if you install with the production flag, so your React code wouldn't get transpiled and everything would crash.

I had never really thought about it, but I guess dependencies in deployment servers are usually installed without production flag and the production flag is then passed during build time. When would you do it the other way around ?

PR #10440, opened 2018-12-12
feat(gatsby-plugin-manifest): add option to remove the "theme color" meta tag

ivorpad:
This fixes https://github.com/gatsbyjs/gatsby/issues/9977

Added an option to remove the "theme color" meta tag from the head.

* `theme_color_in_head` is optional, so if it's not set then the meta tag is added.
* if `theme_color_in_head` is set to false, then the meta tag is not added
* if `theme_color_in_head` is set to true, then the meta tag is added

### DIFF
--- a/packages/gatsby-plugin-manifest/src/gatsby-ssr.js
+++ b/packages/gatsby-plugin-manifest/src/gatsby-ssr.js
@@ -31,16 +31,23 @@ exports.onRenderBody = ({ setHeadComponents }, pluginOptions) => {
       href={withPrefix(`/manifest.webmanifest`)}
     />
   )
-
   // The user has an option to opt out of the theme_color meta tag being inserted into the head.
   if (pluginOptions.theme_color) {
-    headComponents.push(
-      <meta
-        key={`gatsby-plugin-manifest-meta`}
-        name="theme-color"
-        content={pluginOptions.theme_color}
-      />
+    let theme_color = !Object.keys(pluginOptions).includes(
+      `theme_color_in_head`
     )
+      ? pluginOptions.theme_color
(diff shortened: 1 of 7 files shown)


packages/gatsby-plugin-manifest/src/gatsby-ssr.js
```diff
@@ -31,16 +31,23 @@ exports.onRenderBody = ({ setHeadComponents }, pluginOptions) => {
...
   if (pluginOptions.theme_color) {
-    headComponents.push(
-      <meta
-        key={`gatsby-plugin-manifest-meta`}
-        name="theme-color"
-        content={pluginOptions.theme_color}
-      />
+    let theme_color = !Object.keys(pluginOptions).includes(
+      `theme_color_in_head`
     )
```
pieh: this can be:
```
const insertMetaTag = `theme_color_in_head` in pluginOptions 
  ? pluginOptions.theme_color_in_head
  : true
```
We already know that `pluginOptions.theme_color` is truthy, because we are in `if (pluginOptions.theme_color) {`

pieh:
We would also want to add the option in here - https://github.com/gatsbyjs/gatsby/blob/444b7274a83ac675e19350b81f63671fb928ccc4/packages/gatsby-plugin-manifest/src/__tests__/gatsby-node.js#L70-L74 to make sure that this field doesn't get added to web manifest file

ivorpad:
Done, Michal. If there's anything wrong in the added docs then I'd be more than happy to give it another round.

pieh:
I'll move the documentation part to actual plugin and will merge this

Issue #10016, opened 2018-11-18
Static query loads forever when inside a fragment <></>


TylerBarnes:
I've seen this on a couple projects now since I use the fragment syntax pretty often. Each time I do it I forget that the fragment is causing the issue but once I delete it or even replace it with a div the static query loads.

I don't have a demo repo to show and it's a fairly minor bug but it should be easy to reproduce.

These work:
```jsx
<StaticQuery
      query={graphql`
...
`}
render={...}
/>
```
```jsx
<div>
<h1>Component title</h1>
<StaticQuery
      query={graphql`
...
`}
render={...}
/>
</div>
```

This doesn't work and results in the Static query loading message:
```jsx
<>
<h1>Component title</h1>
<StaticQuery
      query={graphql`
...
`}
render={...}
/>
</>
```

jonniebigodes:
Instead of the shorthand version try it like this:
```javascript
<React.Fragment>
    <h1>component Title</h1>
    <StaticQuery
      query={graphql`
        ...
      `}
      render={...}
    />
</React.Fragment>
```
I recreated the issue with the shorthand version and it shows h1 and StaticQuery(loading), but if i use React.Fragment it shows the h1 and also what's inside the render.

victor-cordova:
Hello I've created a PR to address this issue. Please take a look when possible https://github.com/gatsbyjs/gatsby/pull/10443. The explanation of the issue is also there.

PR #10443, opened 2018-12-12
fix(static-query): Use @babel/traverse for traversing algorithm

victor-cordova:
Hello :wave: , this PR fixes https://github.com/gatsbyjs/gatsby/issues/10016.
## Summary
After digging deep into the rabbit hole of the traversal algorithm in `/packages/gatsby/src/internal-plugins/query-runners/file-parser.js`, I found out that the traversing algorithm wasn't taking into account the ast nodes of type JSXFragment. Internally, this is because considered in the babel-types version used by babel-traverse v6.26. Therefore, I am using now the more updated @babel/traverse which is a dependency this package already had.

## Test code
I created a simple `gatsby new gastsby-test` and created a index with the following code
```javascript
import React from 'react'
import { StaticQuery, graphql } from 'gatsby'

import Layout from '../components/layout'

const query = graphql`
  query {
    site {
      siteMetadata {
        title
      }
    }
  }
`;

const IndexPage = () => (
  <Layout>
      <>
        <h1>Hello world!</h1>
        <StaticQuery query={query} render={() => (<div>Hello world</div>)}/>
      </>
  </Layout>
)

export default IndexPage
```

Then, tested it out with before the changes (displaying loading forever) and after the changes (displays correct text). It works fine and tests are still passing :smile: 

Let me know if you have any comments
Cheers

### DIFF
--- a/packages/gatsby/package.json
+++ b/packages/gatsby/package.json
@@ -27,7 +27,6 @@
     "babel-plugin-dynamic-import-node": "^1.2.0",
     "babel-plugin-remove-graphql-queries": "^2.5.2",
     "babel-preset-gatsby": "^0.1.6",
-    "babel-traverse": "6.26.0",
     "better-queue": "^3.8.6",
     "bluebird": "^3.5.0",
     "browserslist": "3.2.8",
--- a/packages/gatsby/src/bootstrap/resolve-module-exports.js
+++ b/packages/gatsby/src/bootstrap/resolve-module-exports.js
@@ -1,6 +1,6 @@
 // @flow
 const fs = require(`fs`)
-const traverse = require(`babel-traverse`).default
+const traverse = require(`@babel/traverse`).default
 const get = require(`lodash/get`)
 const { codeFrameColumns } = require(`@babel/code-frame`)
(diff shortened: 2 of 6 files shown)


jgierer12:
Thanks! We should probably also change the dependency in `package.json` and `yarn.lock`

pieh:
We could also update https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/bootstrap/resolve-module-exports.js (only other place where we import `babel-traverse`). If this is drop-in replacement hopefully this is just matter of adjusting import.

> Thanks! We should probably also change the dependency in `package.json` and `yarn.lock`

We already have `@babel/traverse@^7` in dependencies. So if above is done, then we would need just to remove `babel-traverse@^6` from deps and update lock file.

One thing that could be helpful is to adjust (or add new test) for checking correctness of StaticQuery extraction if its inside `<>`

pieh:
And btw - Awesome job on figuring this one out - it looks like really fringe issue to debug

victor-cordova:
Thanks @pieh :) I'll make those changes and update the PR shortly.

KyleAMathews:
Woah! Thanks for reaching deep and finding this! 🙏

PR #10451, opened 2018-12-13
Typo

HAKRSINGH:
"Keep up with the latest things Gatsby!" didn't seem right. Edited it to Keep up with the latest things about Gatsby!


m-allanson:
Thanks @HAKRSINGH - I think your edit is more technically correct than the current language. 

However I don't think "latest things Gatsby" is _incorrect_. It's slightly more casual version of the phrase. I'm going to close this for now, but also cc @gatsbyjs/docs in case anyone there thinks this is a change we should make.

Thanks again for opening this! Fixing up small errors is a really valuable way of making sure Gatsby's site and docs are high quality.

HAKRSINGH:
Adding "about" gives it more meaning. Just saying. @pieh

PR #10453, opened 2018-12-13
fix(www) Fix query for plugin links always ?=undefined

CarlRosell:
Was just browsing the website and found that #10389 PR changed the prop passed to the Result component to query but did not update the Result component itself.

https://github.com/gatsbyjs/gatsby/pull/10389/files#diff-86b9922590ad1703ea1c4ae10baab77cR252


pieh:
Ops, my verification wasn't thorough enough - I'll quickly validate this locally and hopefully merge - thanks for jumping on this!

PR #10455, opened 2018-12-13
Upgrade `react-hot-loader` to 4.6

### LINKED ISSUE #9489: gatsby-develop | useState from react hooks doesn't work in development environment
## Description

When running dev server, the state from react hooks is not working without any errors. It works, however, in production after compiling the files.

### Steps to reproduce

1. Pull down https://github.com/aamorozov/gatsby-react-hooks;
2. Run `yarn && gatsby develop`;
3. When dev server is running, click on the button - the state is not getting updated here;
4. Run `gatsby build && gatsby serve`
5. When prod server is running, click on the button - the state is being updated correctly.
### Expected result

The state from react-hooks should work in both dev/prod builds. 

### Actual result

The state from react-hooks is working only in production build.

### Environment

```
System:
    OS: macOS 10.14
    CPU: x64 Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 11.0.0 - /usr/local/bin/node
    Yarn: 1.10.1 - /usr/local/bin/yarn
  Browsers:
    Chrome: 69.0.3497.100
    Firefox: 62.0.3
    Safari: 12.0
  npmPackages:
    gatsby: ^2.0.19 => 2.0.28
    gatsby-plugin-flow: ^1.0.2 => 1.0.2
    gatsby-plugin-jsxstyle: ^0.0.3 => 0.0.3
    gatsby-plugin-manifest: ^2.0.5 => 2.0.5
    gatsby-plugin-offline: ^2.0.5 => 2.0.7
  npmGlobalPackages:
    gatsby-cli: 2.4.3
```

jgierer12:
Fixes #9489 

~~The docs also [recommend](https://github.com/gaearon/react-hot-loader#react--dom) (but don't require) using a patched `react-dom`, but as I understand it that would override the user's `react-dom` version we have as a peer dependency so I left it out.~~ :arrow_right: https://github.com/gatsbyjs/gatsby/pull/10455/commits/3894f0a65c1684880e8723e9c6476a042e9c9cc3

/cc @wKovacs64

### DIFF
--- a/packages/gatsby/cache-dir/app.js
+++ b/packages/gatsby/cache-dir/app.js
@@ -17,7 +17,6 @@ setApiRunnerForLoader(apiRunner)
 // necessary for hot-reloading of react hooks
 setConfig({
   ignoreSFC: true,
-  pureRender: true,
 })
 
 // Let the site/plugins run code very early.
--- a/packages/gatsby/cache-dir/app.js
+++ b/packages/gatsby/cache-dir/app.js
@@ -57,7 +57,7 @@ apiRunnerAsync(`onClientEntry`).then(() => {
   loader.addDevRequires(syncRequires)
 
   loader.getResourcesForPathname(window.location.pathname).then(() => {
-    let Root = hot(module)(preferDefault(require(`./root`)))
+    let Root = hot(preferDefault(require(`./root`)))
     domReady(() => {
       renderer(<Root />, rootElement, () => {
         apiRunner(`onInitialClientRender`)
--- a/packages/gatsby/cache-dir/app.js
+++ b/packages/gatsby/cache-dir/app.js
@@ -2,6 +2,7 @@ import React from "react"
 import ReactDOM from "react-dom"
 import domReady from "domready"
 import { setConfig } from "react-hot-loader"
+import { hot } from "react-hot-loader/root"
 
 import socketIo from "./socketIo"
 import emitter from "./emitter"
@@ -18,6 +19,9 @@ setConfig({
   ignoreSFC: true,
   pureRender: true,
 })
+
+const preferDefault = m => (m && m.default) || m
+let Root = hot(preferDefault(require(`./root`)))
 
 // Let the site/plugins run code very early.
 apiRunnerAsync(`onClientEntry`).then(() => {
@@ -57,7 +61,6 @@ apiRunnerAsync(`onClientEntry`).then(() => {
   loader.addDevRequires(syncRequires)
 
   loader.getResourcesForPathname(window.location.pathname).then(() => {
-    let Root = preferDefault(require(`./root`))
     domReady(() => {
       renderer(<Root />, rootElement, () => {
         apiRunner(`onInitialClientRender`)
@@ -65,8 +68,6 @@ apiRunnerAsync(`onClientEntry`).then(() => {
     })
   })
 })
-
-const preferDefault = m => (m && m.default) || m
 
 function supportsServiceWorkers(location, navigator) {
   if (location.hostname === `localhost` || location.protocol === `https:`) {
--- a/packages/gatsby/cache-dir/app.js
+++ b/packages/gatsby/cache-dir/app.js
@@ -2,7 +2,6 @@ import React from "react"
 import ReactDOM from "react-dom"
 import domReady from "domready"
 import { setConfig } from "react-hot-loader"
-import { hot } from "react-hot-loader/root"
 
 import socketIo from "./socketIo"
 import emitter from "./emitter"
@@ -57,7 +56,7 @@ apiRunnerAsync(`onClientEntry`).then(() => {
   loader.addDevRequires(syncRequires)
 
   loader.getResourcesForPathname(window.location.pathname).then(() => {
-    let Root = hot(preferDefault(require(`./root`)))
+    let Root = preferDefault(require(`./root`))
     domReady(() => {
       renderer(<Root />, rootElement, () => {
         apiRunner(`onInitialClientRender`)
--- a/packages/gatsby/cache-dir/app.js
+++ b/packages/gatsby/cache-dir/app.js
@@ -16,6 +16,7 @@ setApiRunnerForLoader(apiRunner)
 // necessary for hot-reloading of react hooks
 setConfig({
   ignoreSFC: true,
+  pureRender: true,
 })
 
 // Let the site/plugins run code very early.
--- a/packages/gatsby/cache-dir/app.js
+++ b/packages/gatsby/cache-dir/app.js
@@ -1,8 +1,6 @@
 import React from "react"
 import ReactDOM from "react-dom"
 import domReady from "domready"
-import { setConfig } from "react-hot-loader"
-import { hot } from "react-hot-loader/root"
 
 import socketIo from "./socketIo"
 import emitter from "./emitter"
@@ -13,15 +11,6 @@ import pages from "./pages.json"
 
 window.___emitter = emitter
 setApiRunnerForLoader(apiRunner)
-
-// necessary for hot-reloading of react hooks
-setConfig({
-  ignoreSFC: true,
-  pureRender: true,
-})
-
-const preferDefault = m => (m && m.default) || m
-let Root = hot(preferDefault(require(`./root`)))
 
 // Let the site/plugins run code very early.
 apiRunnerAsync(`onClientEntry`).then(() => {
@@ -61,6 +50,8 @@ apiRunnerAsync(`onClientEntry`).then(() => {
   loader.addDevRequires(syncRequires)
 
   loader.getResourcesForPathname(window.location.pathname).then(() => {
+    const preferDefault = m => (m && m.default) || m
+    let Root = preferDefault(require(`./root`))
     domReady(() => {
       renderer(<Root />, rootElement, () => {
         apiRunner(`onInitialClientRender`)
--- a/packages/gatsby/cache-dir/app.js
+++ b/packages/gatsby/cache-dir/app.js
@@ -1,7 +1,8 @@
 import React from "react"
 import ReactDOM from "react-dom"
 import domReady from "domready"
-import { hot, setConfig } from "react-hot-loader"
+import { setConfig } from "react-hot-loader"
+import { hot } from "react-hot-loader/root"
 
 import socketIo from "./socketIo"
 import emitter from "./emitter"
--- a/packages/gatsby/src/internal-plugins/query-runner/pages-writer.js
+++ b/packages/gatsby/src/internal-plugins/query-runner/pages-writer.js
@@ -72,15 +72,17 @@ const writePages = async () => {
   components = _.uniqBy(components, c => c.componentChunkName)
 
   // Create file with sync requires of components/json files.
-  let syncRequires = `// prefer default export if available
+  let syncRequires = `const { hot } = require("react-hot-loader/root")
+
+// prefer default export if available
 const preferDefault = m => m && m.default || m
 \n\n`
   syncRequires += `exports.components = {\n${components
     .map(
       c =>
-        `  "${c.componentChunkName}": preferDefault(require("${joinPath(
+        `  "${c.componentChunkName}": hot(preferDefault(require("${joinPath(
           c.component
-        )}"))`
+        )}")))`
     )
     .join(`,\n`)}
 }\n\n`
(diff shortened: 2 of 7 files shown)


wKovacs64:
I took the Medium post and [Anton's comment](https://github.com/gatsbyjs/gatsby/issues/9489#issuecomment-446940087) to mean we wouldn't need to use `setConfig` anymore.

jgierer12:
I think we still need `ignoreSFC`. But you are right, we can definitely remove `pureRender` as that is enabled by default now.

DSchau:
Hmm... so this is looking good.

When I test with my [hooks-example](https://github.com/DSchau/gatsby-react-hooks), I'm getting

> Unhandled Rejection (TypeError): Cannot call a class as a function

```jsx
loader.getResourcesForPathname(window.location.pathname).then(() => {
  let Root = hot(module)(preferDefault(require(`./root`))) // this line
  domReady(() => {
    renderer(<Root />, rootElement, () => {
      apiRunner(`onInitialClientRender`)
    })
  })
})
```

I'll investigate a little further.

jgierer12:
Does `flow-typed/npm/react-hot-loader_vx.x.x.js` have to be updated in some way? I'm not familiar with Flow, so not sure if it should be updated automatically, manually, or not at all.

wKovacs64:
I'm not able to test it myself at the moment, but it would be cool if this helps fix hooks at the top (pages) level. (Or is that more of a Gatsby internals issue?)

jgierer12:
@DSchau hmm, I am getting the same error now in one of my test projects.

@wKovacs64 yes it now also works at the top level (when it works at all :sweat_smile: )

DSchau:
The (partial) fix is to change how we are marking the root component as hot. In `src/packages/gatsby/cache-dir/app.js`

we need to change line 60 from

```jsx
let Root = hot(module)(preferDefault(require(`./root`)))
```

to

```jsx
let Root = hot(preferDefault(require(`./root`)))
```

See [the README of react-hot-loader](https://github.com/gaearon/react-hot-loader#getting-started) for more info!

DSchau:
With the fix above (☝️) this gets us most of the way there.

Unfortunately, I seem to get this error which _could_ be unrelated to this PR but will need to be fixed before we can merge this. Debugging that now.

screen shot 2018-12-13 at 10 51 29 am

jgierer12:
@DSchau I can also reproduce that, I thought it was related to the missing `react-dom` patch but even with that patch I still get the same error. I will investigate further.

However, I did find a way to use the patch with a user-provided `react-dom` version by including RHL's webpack plugin.

jgierer12:
Fixed - the new `hot` API expects the component to be created in the parent module of where `hot` is called. We had it in the same module.

I will now test it with a few more projects, but the one where the two errors occurred (https://github.com/jgierer12/gatsby-api-proposal-demo) finally works now :+1:

jgierer12:
@DSchau what are your thoughts about https://github.com/gatsbyjs/gatsby/pull/10455#issuecomment-447037043?

DSchau:
Not sure which change caused it, but we're actually not able to hot reload _at all_ with the most recent changes. Appears to simply refresh the page upon any content changes.

hot-reloading

DSchau:
@jgierer12 re: 

> Does flow-typed/npm/react-hot-loader_vx.x.x.js have to be updated in some way? I'm not familiar with Flow, so not sure if it should be updated automatically, manually, or not at all.

Keep it as is for now. It's relatively tangential to the PR and it appears that _none_ of those files have been updated for the last two years 🙃

DSchau:
Also I'm not sure we _want_ to implement the webpack loader as we've done in 3894f0a. I think we should prefer the babel loader, which we currently use.

This _was_ working with the babel loader, and there's some _spooky_ language about using the webpack plugin with class plugins.

> We recommend to use babel plugin, but there are situations when you are unable to use it. Then - try webpack loader (as seen in v3), but remember - it is not compatible with class-based components, but help with TypeScript or spreading "cold API" to all node_modules.

jgierer12:
I'll have to call it a day for now, here are my findings in case you want to continue debugging:

*  A weird thing I noticed is that, when placing `import "react-hot-loader/root"` at the top of `packages/gatsby/cache-dir/app.js`, hot reloading works but the `hot update was not successful` banner is back.
* Hot reloading seems to work fine, when the page component is either a class component or an arrow function but not a `function()`.

packages/gatsby/cache-dir/app.js
```diff
@@ -57,7 +57,7 @@ apiRunnerAsync(`onClientEntry`).then(() => {
   loader.addDevRequires(syncRequires)
 
   loader.getResourcesForPathname(window.location.pathname).then(() => {
-    let Root = hot(module)(preferDefault(require(`./root`)))
+    let Root = preferDefault(require(`./root`))
```
theKashey: New `/root/hot` API will not stand asynchronously between the first part and the second.

theKashey:
`hot update was not successful` would be shown when "first" part (inside /root) was called, but second (the actual `hot`) was not.
It's the same `hot(module)(Component)`, just called in two steps.

Plus - @jgierer12 - `pureRender` is __not__ enabled by default.

jgierer12:
> `pureRender` is **not** enabled by default

Yes, I noticed later on that I confused `pureRender` with `pureSFC` :see_no_evil: that's why I reenabled it in https://github.com/gatsbyjs/gatsby/pull/10455/commits/24cb4d710e206263dad2a08fc160e238df8d0ed7

packages/gatsby/cache-dir/app.js
jgierer12: Do you mean that `hot` should stay in this module? i.e.
```js
let Root = hot(preferDefault(require(`./root`)))
```

That was the reason for the original `hot update was not successful` error

jgierer12:
> It's the same `hot(module)(Component)`, just called in two steps.

Thank you @theKashey , I realized that the first step (`hot(module)`) gets already called when `require`ing `react-hot-loader/root`! That helps a lot, now just to figure out how to call it synchronously.

jgierer12:
@DSchau Could you give this another look? It now works again on my end.

packages/gatsby/cache-dir/app.js
theKashey: `hot` or `import 'react-hot-loader/root` are calling `module.hot.accept` for the current module, thus setting self-acceptance.
It's expected that some action would take a place on module hot update ____immediately____ and in a synchronous mode. Using `hot` inside indirectly called function are breaking some expectations - "module" could not accept it, as long as the "real action" is not yet called.

Here - if you want to make `./root` hot - you have to make IT hot. And that's was done.

The second moment is about HMR and an update propagation.
When you change a file webpack _bubbles_ update to parent, unless parent will accept the change, then webpack will update parent and everything below it.
We had WFT-level issues if "parent" is the module with `react-dom/render` - everything got wiped and regenerated, making reconciliation impossible (`Symbols` could be updated for example). 

This is not quite clear from RHL documentation, sorry. The difference is between `self-accept` in v4, and `child-accept`(ie _module.hot.accept('./containers/App'_) in v3.

Here you might want to accept only `./root`, but would accept _everything_.

packages/gatsby/cache-dir/app.js
```diff
@@ -19,6 +20,9 @@ setConfig({
   pureRender: true,
 })
 
+const preferDefault = m => (m && m.default) || m
+let Root = hot(preferDefault(require(`./root`)))
```
theKashey: So - as I mentioned in another comment - `hot` here would setup __current__ module as self-accepted, wiping and updating __everything__ _below_ it every time.
- react-dom
- socketIo
- emitter

It would be better to move it to `./root` or have another file, imported above this one with some important imports. That would link "important" modules with another "parent branch", and protect them during hot update.

Sound recommendation - have `hot` below(in module terms) `react-dom/render` and `redux/createStore` to maintain their values on update. You have to stop event propagation before it bubbles to the places you should keep between updates.

jgierer12: @theKashey thanks again, I now moved it into `.cache/sync-requires.js` so only page components are hot loaded.

@DSchau This is not the most beautiful solution, but it was the best I could think of without applying `hot` further up. An alternative could be to map over `syncRequires` in `app.js` and call `hot` there instead of writing it into the generated `sync-requires.js`.

DSchau:
Checking this out now! Thanks for the work on this both of you, it's appreciated!

packages/gatsby/src/internal-plugins/query-runner/pages-writer.js
```diff
@@ -72,15 +72,17 @@ const writePages = async () => {
   components = _.uniqBy(components, c => c.componentChunkName)
 
   // Create file with sync requires of components/json files.
-  let syncRequires = `// prefer default export if available
+  let syncRequires = `const { hot } = require("react-hot-loader/root")
```
DSchau: All of this looks great--and I was able to validate that this does fix a number of issues with hot reloading + hooks, namely:

- Hot reloading of page components
- Hot reloading does not show an error on launch
- Hot reloading (still) works with class components, markdown files, etc.

However -- I don't think this is the best place to implement this hot reloading functionality. react-hot-loader is basically a noop in non-development mode (e.g. NODE_ENV=development), but I don't love the idea of putting this in front of every page in every environment.

@pieh do you have a better idea as to where this hot reloading functionality should be implemented?

jgierer12: I agree, I originally had it in `app.js` (https://github.com/gatsbyjs/gatsby/pull/10455/commits/f0883f1385171b180fa11320cf95715f506f9c62) which also worked fine, but I moved it into the generated `sync-requires.js` as per https://github.com/gatsbyjs/gatsby/pull/10455#discussion_r241576619. 

Aren't the `syncRequires` only loaded in development mode anyways and replaced by `asyncRequires` in prod?

DSchau: @jgierer12 good catch! Yup!

https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/cache-dir/production-app.js#L15

pieh: > Aren't the `syncRequires` only loaded in development mode anyways and replaced by `asyncRequires` in prod?

Yes, they are currently - but we will be looking to split dev bundles similar to what we do for production - which should result in faster recompilations and also better experience for hosted/not-local preview servers. Not yet sure how would that affect hot reloads.

DSchau:
Taking one final look at this--but think we're good to go and merge this in. Will report back!

DSchau:
```
Successfully published:
 - gatsby-transformer-remark@2.1.16
 - gatsby@2.0.69
```

`gatsby@2.0.69` should contain this updated hot reloading. Thanks all for the help and 👀 on this one! Well done!

DSchau:
I think this caused a fairly large regression in `develop` mode. Fortunately, it doesn't impact the built application, but this is still cause for concern. We'll want to get a fix out tomorrow or possibly revert this commit.

See #10528 for more info. Seems to be related to anonymous arrow functions being used as page components.

jgierer12:
@theKashey could you elaborate on what the implications are when calling `hot` 'further up' like in https://github.com/gatsbyjs/gatsby/pull/10455/commits/be3d53e8055448af21157efe2580a1176eff3230?

I think this would be the best/only other solution without changing the underlying component structure.

theKashey:
The core of RHL is a "proxy" - it knows that instead of OldComponent it should use NewComponent, and that is done via _top level variables_ registration. Basically UID for a Component is `variable name` + `module name`.
The same stuff was hidden inside `hot` - `displayName` + `module name`. As long as it was the same for all used components - they all become the same.

And I have no idea, why it did work before.

google-mac:
React hooks are not functional with these versions:

Is there additional work to utilize the new React features in a new Gatsby project created by the CLI?

- react@16.7.0-alpha.2
- react-dom@16.7.0-alpha.2
- gatsby@2.0.69
- gatsby-transformer-remark@2.1.16

google-mac:
@DSchau I just got it working in a different test. I believe there are more requirements than what I mentioned above. I looked through your repo and updated the dependencies to match for the other plugins. So it is more than React (and React-DOM), and Gatsby. Thanks for the help!

theKashey:
Right now `Gatsby` would not be hot-reloadable without a "patch" applied, due to `ignoreSFC` option set(to support hooks).

There are two ways to fix it:
- apply the patch (yarn add react-dom@npm:@hot-loader/react-dom)
- disable this option

PR #10460, opened 2018-12-13
Set maxWidth of fluid image to avoid them stretching to match div width

oorestisime:
So gatsby-image when on fluid mode sets maxWidth to be 100% for the div. This is a bit problematic since images smaller than the div will be stretched. Here's an example https://gatsby-remark-rehype-images.netlify.com/custom-component/ (further down there is an image).

The problem here though is that this change modifies default behaviour people might have relied upon. Images no longer taking maxWidth are left aligned which might mess with people designs. 
Maybe its best to document this as a breaking change. Open for suggestions.

FTR here's how the presentationWidth is calculated on plugin-sharp

```
let presentationWidth, presentationHeight
  if (fixedDimension === `maxWidth`) {
    presentationWidth = Math.min(
      options.maxWidth,
      Math.round(width / pixelRatio)
    )
    presentationHeight = Math.round(presentationWidth * (height / width))
  } else {
    presentationHeight = Math.min(
      options.maxHeight,
      Math.round(height / pixelRatio)
    )
    presentationWidth = Math.round(presentationHeight * (width / height))
  }
```

This happens more or less in the same way for `fixed` images. Here's the styles applied to the div

```
const divStyle = {
        position: `relative`,
        overflow: `hidden`,
        display: `inline-block`,
        width: image.width,
        height: image.height,
        ...style,
      }
```

### DIFF
--- a/packages/gatsby-image/src/index.js
+++ b/packages/gatsby-image/src/index.js
@@ -253,14 +253,17 @@ class Image extends React.Component {
     if (fluid) {
       const image = fluid
 
+      const divStyle = {
+        position: `relative`,
+        overflow: `hidden`,
+        maxWidth: image.presentationWidth,
+        ...style,
+      }
+
       return (
         <Tag
           className={`${className ? className : ``} gatsby-image-wrapper`}
-          style={{
-            position: `relative`,
-            overflow: `hidden`,
-            ...style,
-          }}
+          style={divStyle}
           ref={this.handleRef}
           key={`fluid-${JSON.stringify(image.srcSet)}`}
         >


LekoArts:
https://www.gatsbyjs.org/packages/gatsby-image/#two-types-of-responsive-images
> Images that stretch across a fluid container

The `maxWidth` parameter is not (at least I understand it like that) about the `max-width` of the wrapping container or something, it's about what you think what the maximum width of the images generated should be. `max-width: 100%` is necessary for this fluid behavior that's described there.

> This is a bit problematic since images smaller than the div will be stretched.

Then your source and/or CSS is not aligned with what you can work with. And you probably should use `fixed` query then.

All in all I don't think that the current behavior should be changed, especially because so many people will rely on that.

pieh:
I've been thinking about this and maybe easier way about it is using extra component to wrap `gatsby-image` and document that in docs:

Didn't test - but something like this:
```js
import Image from `gatsby-image`

const NotStretchedImage = props => {
  let normalizedProps = props
  if (props.fluid && props.fluid.presentationWidth) {
    normalizedProps = {
      ...props,
      style: {
        ...(props.style || {}),
        maxWidth: props.fluid.presentationWidth
      },
    }
  }
  return <Image {...normalizedProps} />
}
```

oorestisime:
Yeah this is the second solution i was thinking! i will wait a bit more to make the change in case somebody has something else to contribute here!

LekoArts:
> Well the behavior wouldn't change for images that are of bigger or equal width as the div enclosing them.

It would break things. I use `maxWidth: 1920` for my header images but stretch it above that on purpose. I just don't want to use source images above 1920.

I like @pieh solution, as it is a documentation change and not a breaking change for everyone.

oorestisime:
So I've discussed this with Pieh and i am closing this PR in profit of another one that just updates the documentation to mention the issue and provide an example of an easy solution as suggested by Pieh above!!

PR #10469, opened 2018-12-14
Upgrade cli dependency yurnalist

0x80:
I have updated yurnalist to use the code and dependencies from the yarn master branch (1.13.0), and then further upgraded all dependencies to their latest version, and stripped some that were not used.

This will solve some things like missing emojis. Also the output from yurnalist can now be muted (for example in CI) via env var `YURNALIST_SILENT` just like yarn has `YARN_SILENT`.

For more changes see: https://github.com/0x80/yurnalist/blob/master/CHANGELOG.md

### DIFF
--- a/packages/gatsby-cli/package.json
+++ b/packages/gatsby-cli/package.json
@@ -29,7 +29,7 @@
     "stack-trace": "^0.0.10",
     "update-notifier": "^2.3.0",
     "yargs": "^11.1.0",
-    "yurnalist": "^0.2.1"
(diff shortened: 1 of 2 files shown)


0x80:
BTW I'm not sure, but I think CI output silencing might be handled automatically by `yurnalist`. This update included detection functions like `is-ci`. In any case the behavior should be the same as `yarn`.

DSchau:
```
Successfully published:
 - gatsby-cli@2.4.7
 - gatsby@2.0.72

```

Issue #8805, opened 2018-10-04
feat: update gatsby plugins to use new createContentDigest helper


DSchau:
## Summary

We [recently added](https://github.com/gatsbyjs/gatsby/pull/8687) a `createContentDigest` helper which is exposed to any gatsby plugin. This internalizes the unique key creation, which changes based on changing content.

We've added this to several packages, and we should now add it to any/all packages.

### Basic example

Any plugin that implements the following code snippet (or close to it) can be refactored with this new helper:

```js
const crypto = require(`crypto`)

module.exports = ({ createNode }) => {
  const obj = { whatever: true }

  // this is the key one!
  const contentDigest = crypto
  .createHash(`md5`)
  .update(JSON.stringify(obj))
  .digest(`hex`)

  createNode() // use content digest
}
```

can be replaced with

```js
const { createContentDigest } = require('gatsby-core-utils');


module.exports = ({ createNode }) => {
  const obj = { whatever: true }

  // this is the key one!
  const contentDigest = createContentDigest(obj)

  createNode() // use content digest
}
```

### Motivation

Cleaning up code and keeping it DRY 💪

### Additional Info

Any plugin that implements this new API should also bump it's `peerDependency` on gatsby, e.g.

```json
{
  "peerDependencies": {
    "gatsby": ">2.0.15"
  }
}
```

samscha:
Thank you for the reference. I'll do some testing on my own for some of the most popular plugins and report back / open a pr!

amberleyromo:
@DSchau Maybe it would be useful to break this apart into different issues for the different plugins? Or at least a listing of todos in this issue description?

frinyvonnick:
Here is the list of plugins that seems not using `createContentDigest` as suggested @amberleyromo :

- [ ] [gatsby-source-hacker-news](https://github.com/gatsbyjs/gatsby/blob/eeef674e03c6e9fbdda0fea4821914a1861584fa/packages/gatsby-source-hacker-news/src/gatsby-node.js)
- [ ] [gatsby-transformer-excel](https://github.com/gatsbyjs/gatsby/blob/95c1eac260007a145e7c306bfb672a1e1e67c39c/packages/gatsby-transformer-excel/src/gatsby-node.js)
[...]

PR #10471, opened 2018-12-14
refactor(gatsby): update internal-data-bridge plugin to use createContentDigest

smurrayatwork:
This updates gatsby's internal-data-bridge plugin to use createContentDigest, as requested by #8805.

First time gatsby contributor here! I was unsure as to how to add tests to this particular plugin, as I didn't see any for the other internal plugins either.


pieh:
Tests for this are tricky, especially one for updating `Site` nodes. I don't love it, but let's merge You PR as-is. I verified it works locally.

Issue #8195, opened 2018-09-15
Gatsby v2, frontmatter image, relative path, netlify-cms


grdp:
Hi, I'm trying to update an old website from gatsby v1 to gatsby v2. The gatsby-node.js file contain this code below to make working an absolute path of an image to a relative path in the frontmatter of a markdown (because of imageSharp and netlify-cms). Anyone has been confronted to the same issue?

```
exports.onCreateNode = ({ node, actions, getNode, getNodes }) => {
  const { createNodeField, createParentChildLink } = actions;

  if (node.internal.type === `MarkdownRemark`) {
    const value = createFilePath({ node, getNode });
    createNodeField({
      name: `slug`,
      node,
      value
    });

    if (typeof node.frontmatter.image === 'string') {
      const pathToFile = path
        .join(__dirname, 'static', node.frontmatter.image)
        .split(path.sep)
        .join('/');

      const fileNode = getNodes().find(n => n.absolutePath === pathToFile);

      if (fileNode != null) {
        const imageSharpNodeId = fileNode.children.find(n =>
          n.endsWith('>> ImageSharp')
        );
        const imageSharpNode = getNodes().find(n => n.id === imageSharpNodeId);

        createParentChildLink({ parent: node, child: imageSharpNode });
      }
    }
  }
};
```

kmorf:
gatsby-node.js
```exports.onCreateNode = ({ node, actions, getNode }) => {
  const { createNodeField } = actions
  const { frontmatter } = node
  if (frontmatter) {
    const { image } = frontmatter
    if (image) {
      if (image.indexOf('/img') === 0) {
        frontmatter.image = path.relative(
          path.dirname(node.fileAbsolutePath),
          path.join(__dirname, '/static/', image)
        )
      }
    }
  }

  if (node.internal.type === `MarkdownRemark`) {
    const value = createFilePath({ node, getNode })
    createNodeField({
      name: `slug`,
      node,
      value,
    })
  }
} 
```
config.yml
```
media_folder: static/img
public_folder: /img
```

grdp:
Great @kmorf ! It seems working. Thanks a lot!

PR #10478, opened 2018-12-14
[RFC] Add standard frontmatter transforms to netlify cms plugin

0x80:
When trying to work with the Netlify CMS plugin, I ran into an issue with images like #8195. I found it odd that I had to manually transform the image paths to relative paths to make them compatible with	the gatsby-transformer-sharp plugin. 

I think it's safe to assume that anyone working with the Netlify CMS will want to process their images for resizing and such, otherwise you would be missing a big part of what makes Gatsby perform great.

In this PR I have pulled the code for transforming images and some other standard tasks like creating a slug into the plugin code.

I have added two config options for the plugin:

* `contentFolder`: This would be the root folder where all of your Netlify content lives, which defaults to "content".
* `imageFields`: A list of frontmatter fields that hold paths to images. These fields will be transformed to relative paths.

Now that I think of it, `imageFields` should maybe be renamed to `assetFields` or `mediaFields` since I assume that you'd want to same transformation if you use other types of assets in the CMS.

Of course the docs would need to be updated also, but first I'm curious to get some feedback and verify if this is all a good idea to begin with.

With these changes you can start using the CMS without writing additional client code in `gatsby-node.js`, which is I think how a plugin should work ideally.

### DIFF
--- a/packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
+++ b/packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
@@ -77,7 +77,7 @@ exports.onCreateNode = (
       createNodeField({
         node,
         name: `slug`,
-        value: slug || ``,
+        value: slug,
       })
 
       // Used to determine a page layout.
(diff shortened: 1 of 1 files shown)


0x80:
I am unsure how to proceed on this. If it is considered a good idea I can wrap things up and document the changes in the plugin readme.

But I would like to have some sort of "go" before I do that. 

@erquhart ?

0x80:
@erquhart Thanks for the heads up. I wasn't aware of the history of relative images in Netlify. I will read up on it 👍

KyleAMathews:
Looks like this is going to be fixed in Netlify CMS so I'll close this here. Thanks!

erquhart:
Relative image path support is now in beta as of Netlify CMS 2.9.8-beta.2 - try it out:

```
npm update netlify-cms-app@beta
```

Then add `media_folder_relative: true` to your Netlify CMS config.

[Docs here](https://www.netlifycms.org/docs/beta-features/#relative-image-paths).

PR #10485, opened 2018-12-15
Improved some grammar

T1duS:


aadibajpai:
I don't really think that those texts need a fix, they seem fine as-is.

HAKRSINGH:
These updates are not required. I'm closing this for now.

Thanks again for opening this! Fixing up small errors is a really valuable way of making sure Gatsby's site and docs are high quality.

LekoArts:
@HAKRSINGH Please don’t close PRs prematurely and instead let members of the team (in this case docs) decide whether it’s a good PR or not.

Issue #10477, opened 2018-12-14
[gatsby-link] | Add onMouseLeave prop


aamorozov:
## Summary

Currently, it's possible to pass `onMouseEnter` but not `onMouseLeave`. Would be nice to add this option. I will work on the pr tonight. 
https://github.com/gatsbyjs/gatsby/blob/0514b7d2532c60c09efc42567d9178ef0396ee27/packages/gatsby-link/src/index.js#L116

### Basic example

```

class EnhancedLink extends Component {
    state={active: false}
    handleMouseEnter = () => this.setState({active: true}) 
    handleMouseLeave = () => this.setState({active: false}) 
    render() {
      return (
        // Link here is essentially a wrapper around gatsby link with an identification of external/internal link
        <Link
            href={`${url}`}
            style={{
                 background: this.state.active ? 'green' : 'yellow'
            }}
            onClick={this.handleClick}
            onMouseEnter={this.handleMouseEnter} // currently possible 
            onMouseLeave={this.handleMouseLeave} // not currently possible
     )}
}
```

### Motivation

I think it will be beneficial to the end user to be able to handle both states directly from the props that are being passed to Link component, versus creating additional wrappers around gatsby link.

jgierer12:
As I said in #10489, this should already work - see [this demo](https://condescending-boyd-451435.netlify.com/) ([Source](https://github.com/jgierer12/gatsby-pr-10489-demo)).

Can you share a repo where your issue can be reproduced?

aamorozov:
@jgierer12 
It does, not sure how I missed it's working without providing explicit prop 🤦‍♂️ .

PR #10489, opened 2018-12-15
[gatsby-link] - adding onMouseLeave prop

aamorozov:
Adding a `onMouseLeave` prop to gatsby-link component to be able to handle on/off hover states. 

Reference issue #10477


jgierer12:
Shouldn't this already be passed with the `...rest`?

jgierer12:
It is, see https://condescending-boyd-451435.netlify.com/ / https://github.com/jgierer12/gatsby-pr-10489-demo

Closing this, as it already works. If you have problems passing an `onMouseLeave` handler, we can continue investigating in your original issue #10477.

PR #10492, opened 2018-12-15
docs: add note about using client-side environment variables 

sondrele:
I recently spent some time scratching my head why my environment variables where no longer available after I updated gatsby.

I couldn't find anything in the documentation of why they were undefined, but after some digging I came across #10030 , and it all made sense. 

Just thought I'd add an extra note of why object destructing of `process.env` doesn't work, I hope what I'm saying is correct. All feedback's appreciated :)

### DIFF
--- a/docs/docs/environment-variables.md
+++ b/docs/docs/environment-variables.md
@@ -93,7 +93,8 @@ GATSBY_API_URL=https://example.com/api
 API_KEY=927349872349798
 ```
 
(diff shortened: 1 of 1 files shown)


shannonbux:
@m-allanson assigned you here because maybe this is related to gatsby v2 (I could be wrong :)

sondrele:
Awesome, thank you @shannonbux !

pieh:
Taking this from @m-allanson plate, because I introduced change that is documented by this PR.

pieh:
Thanks @sondrele for docs clarification on this!

PR #10495, opened 2018-12-16
docs(gatsby-image): mention how to avoid stretched images in fluid

oorestisime:
Following up on PR #10460 with a docs improvement. 
I am not sure there are other places this needs to be mentioned.

### DIFF
--- a/packages/gatsby-image/README.md
+++ b/packages/gatsby-image/README.md
@@ -288,7 +288,7 @@ const NonStretchedImage = props => {
 **Note:** The `GatsbyImageSharpFluid` fragment does not include `presentationWidth`.
 You will need to add it in your graphql query as is shown in the following snippet:
 
-```
+```graphql
 {
   childImageSharp {
     fluid(maxWidth: 500, quality: 100) {
--- a/packages/gatsby-image/README.md
+++ b/packages/gatsby-image/README.md
@@ -154,7 +154,7 @@ object called `fluid`.
 As mentioned previously, images using the _fluid_ type are stretched to
 match the container's width. In the case where the image's width is smaller than the available viewport, the image will stretch to match the container, potentially leading to unwanted problems and worsened image quality.
 
-To counter this edge case one could wrap the _Img_ component in order to set a better, for that case, maxWidth:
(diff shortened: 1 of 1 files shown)


packages/gatsby-image/README.md
```diff
@@ -149,6 +149,31 @@ child object of an image called `fixed` — which you can see in the sample
 component above. For the second type, you do a similar query but for a child
 object called `fluid`.
 
+### Avoiding stretched images using the fluid type
+
+As mentioned previously, images using the _fluid_ type are stretched to
+match the container's width. In the case where the image's width is smaller than the available viewport, the image will stretch to match the container, potentially leading to unwanted problems and worsened image quality.
+
+To counter this edge case one could wrap the _Img_ component in order to set a better, for that case, maxWidth:
```
LekoArts:
```suggestion
To counter this edge case one could wrap the _Img_ component in order to set a better, for that case, `maxWidth`:
```

packages/gatsby-image/README.md
```diff
@@ -149,6 +149,31 @@ child object of an image called `fixed` — which you can see in the sample
...
+
+```js
```
LekoArts:
```suggestion
```jsx
```

pieh:
It probably should be mentioned that default fragments for fluid queries don't have `presentationWidth` field in them and people would need to adjust their queries for that. So maybe we could move this to `"Fluid" queries` or `Some other stuff to be aware of` sections?

oorestisime:
Would it make sense to include it in the fragments? Having used the workaround a few times it would make life a bit easier and i am not sure i see a reason not to.

packages/gatsby-image/README.md
```diff
@@ -260,6 +260,45 @@ prop. e.g. `<Img fluid={fluid} />`
...
+  }
+
+  return <Img {...normalizedProps} />
+}
+```
+
+**Note:** The `GatsbyImageSharpFluid` fragment does not include `presentationWidth`.
+You will need to add it in your graphql query as is shown in the following snippet:
+
+```
```
LekoArts:
````suggestion
```graphql
````

LekoArts:
@pieh Would be good to go IMO

Issue #10498, opened 2018-12-17
Code parsed incorrectly in excerpts


nadrane:
## Description
When excerpts are queried from `gatsby-transformer-remark` using the `format: HTML` option, the resulting HTML is not parsed correctly.

The system is returning something like this:

```
<p>We have an Elasticsearch backed search interface ... <code class="language-text">Joe</code>...  that the user cares about.</p>
```

The problem is that the `code` tag is being processed as a text node rather than parsed as part of the HTML syntax tree. This is related to work I did adding the `format: HTML` option in the first place. I'll try to figure it out.

### Steps to reproduce

Make a GraphQL request like the following:
```graphql
  {
    allMarkdownRemark {
          excerpt(format: HTML)
}
```

nadrane:
hmm after a little investigation, I think it might be unrelated to the `gatsby-transformer-remark` package. I think it might actually be a problem is `prismjs`... I'm going to evaluate that next...

nadrane:
Alright, just confirmed that my suspicion was correct. The issue appears when `gatsby-remark-prismjs` is enabled

nadrane:
It turns out that when we were converted the HTML AST to an HTML string, we were not allowing for raw html input nodes. Basically, an raw HTML node was escaped. The attached PR changes this behavior. Note that this modified behavior aligns with how the resolver for the entire HTML document works

PR #10499, opened 2018-12-17
feat(gatsby-transformer-remark): support raw HTML in markdown excerpts

nadrane:
fix issue #10498 

The first commit was to fix some tests that were just totally not working. We weren't correctly passing around all of our pluginOptions.

The second commit added a test that validated that `code` tags worked at all. Mostly a sanity check for me. Doesn't add a ton of value overall.

The third commit the fix.

One thing to note here is that I don't actually have a test for this issue. Off the top of my head, I'm not 100% sure how to test it. Maybe inserting raw HTML into the excerpt and expecting it to not get escaped would work?

EDIT: Just added a test that makes sure we can render raw html in excerpts and replaced the less useful test from before.

### DIFF
--- a/packages/gatsby-transformer-remark/src/__tests__/__snapshots__/extend-node.js.snap
+++ b/packages/gatsby-transformer-remark/src/__tests__/__snapshots__/extend-node.js.snap
@@ -54,15 +54,6 @@ Object {
 }
 `;
 
-exports[`Excerpt is generated correctly from schema given an html format and an excerpt_separator, it correctly parses code tags 1`] = `
-Object {
-  "excerpt": "<p>Where is my <code>pony</code> named leo?</p>",
-  "frontmatter": Object {
-    "title": "my little pony",
-  },
-}
-`;
-
 exports[`Excerpt is generated correctly from schema given an html format, it correctly maps nested markdown to html 1`] = `
 Object {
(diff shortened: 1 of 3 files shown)


nadrane:
@DSchau Any chance you be my reviewer again? :)

I think this should be a pretty straight-forward one

DSchau:
Will do! I’ll check this out tomorrow morning!

DSchau:
@nadrane ugh, sorry--didn't get to this today.

One thing I thought would be helpful is updating `examples/using-remark` with this new behavior. Any interest in doing that? Tangential to this PR, but it'll "integration" test it as well as providing a helpful resource to some.

DSchau:
@nadrane exactly, yeah! Not a blocker for this PR by any means, but it's a nice thing to show/document!

DSchau:
> BTW. First I wanted to fix this error myself, only later I started to look for issue for related problem and therefore I have a proposal: @DSchau Have you thought about creating labels for plugins, for example plugins/transform-remark? Surely, the issues reported would be faster to find.

Sure--this seems reasonable! I'll make a note of it and circle back with the team to see if it makes sense. Thanks!

As far as this PR goes--last I recall it was ready to go. An example (e.g. using-remark) would be _nice to have_ but not necessary as I said earlier!

For now--let's just get this merged, and we can add that example at some later date. Thanks!

PR #10500, opened 2018-12-17
Emotion update

ChristopherBiscardi:
* Updated gatsby-plugin-emotion to support v10 and automatic SSR.
  - Used new emotion css-prop preset and removed custom jsx pragma, fragment and plugin handling
  - Removed backward compatibility cache and old SSR methods, which if still desired from a project perspective should probably be a separate transitionary tag or package (happy to add that if desired)
* Updated relevant tutorials

* Did not update the testing css-in-js tutorial, which should also be updated but I didn't get a chance to run through the earlier parts to update it yet.


packages/gatsby-plugin-emotion/src/gatsby-node.js
```diff
@@ -1,27 +1,7 @@
 export const onCreateBabelConfig = ({ actions }, pluginOptions) => {
-  const pragmaName = `___EmotionJSX`
-
-  actions.setBabelPlugin({
-    name: `babel-plugin-jsx-pragmatic`,
+  actions.setBabelPreset({
```
DSchau: This should have probably been what we did from the beginning.

Doing some cursory reading--this doesn't require us to set the `jsx` pragma, correct? I'm not sure I like that approach, if so.

DSchau: (although this wasn't out yet--so heh, can't really fault us for that)

ChristopherBiscardi: heh yeah, the nice "single preset" approach wasn't out yet when the original emotion changes were made to gatsby plugin related stuff. It was still the piecemeal approach. 

Making this change *does not* require anyone to manually use the jsx pragma. The pragma is applied universally automatically. You can see [in the preset](https://github.com/emotion-js/emotion/blob/b966abd80cc4817cd27f2e4e6c7994cb2faf9043/packages/babel-preset-css-prop/src/index.js) that it's basically the same as what we were doing before by providing `jsx` and setting the pragma (with the React.Fragment settings, etc) and the core emotion plugin.

DSchau:
@ChristopherBiscardi pulling this down now and checking it out 👌 Let's get this merged in today if all good.

Do you think it's OK to do this as a patch release? I think it was sort of a mistake to bump the previous to major and keep some of the legacy stuff intact.

ChristopherBiscardi:
@DSchau If someone updated to v3 and also kept using the old APIs then they might have an issue if this goes out as a patch. I'd do a minor at least, but since npm installs at a [minor upgrade by default](https://docs.npmjs.com/about-semantic-versioning#using-semantic-versioning-to-specify-update-types-your-package-can-accept) (`^` versions) it depends on how many people we think have installed v3 in the last two weeks (v3 went out on 12/4). The tutorials I changed were still using old APIs that would've required the compatibility cache (like `injectGlobal`), so anyone that installed `gatsby-plugin-emotion` in the last two weeks and also followed the `creating-global-styles` tutorial would break.

I'm of the same opinion that bumping to v3 with the legacy support was probably a mistake, but it seems like the safest move is bump to v4 now that it's happened.

DSchau:
> I'm of the same opinion that bumping to v3 with the legacy support was probably a mistake, but it seems like the safest move is bump to v4 now that it's happened.

Would agree. Let's call this 4.0.0.

DSchau:
```
Successfully published:
 - gatsby-plugin-emotion@4.0.0
```

ryanditjia:
I’m not sure why 4.0.0’s npm installation still has the old version, as if this PR isn’t taken into account. It still has `wrap-element.js` and all.

Did `npm info gatsby-plugin-emotion` and got the download URL: `wget https://registry.npmjs.org/gatsby-plugin-emotion/-/gatsby-plugin-emotion-4.0.0.tgz` to confirm this.

Could you check what went wrong with the publish?

Thanks.

DSchau:
@ryanditjia ugh. Yeah - we don't have a clean step (or use the `files` array in package.json) so some legacy artifacts made their way in.

I'll PR this quickly and get 4.0.1 up.

PR #10514, opened 2018-12-17
Update sites.yml

rcaracaus:

### DIFF
--- a/docs/sites.yml
+++ b/docs/sites.yml
@@ -3779,3 +3779,16 @@
     - Blog
   built_by: mottox2
   built_by_url: "https://mottox2.com"
+- title: Surfing Nosara
+  main_url: "https://www.sufingnosara.com"
+  url: "https://www.sufingnosara.com"
+  description: Real estate, vacation, and surf report hub for Nosara, Costa Rica
+  featured: false
+  categories:
+    - Surf
+    - Business
+    - Blog
+    - Gallery
+    - Marketing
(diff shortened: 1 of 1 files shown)


pieh:
Hey, it seems that site is down - https://downforeveryoneorjustme.com/www.sufingnosara.com

rcaracaus:
Thank God for peer review. Typo. Haha. Please have another go when you get a chance. @pieh my bad!

rcaracaus:
@pieh @LekoArts a few things...

1) The category "Surf" already exists, that is what I chose it, now the categories "Surf" and "Surfing" exist, I agree "Surfing" makes more sense as a category, but we should probably change the other website.

2) There seems to be caching issues with Service Worker. When I went to https://www.gatsbyjs.org/showcase/ even after hard refresh in a new tab, I was not shown the updated showcase (with the new website in the listing or filters). In order to see SurfingNosara.com in the showcase I actually had to open a new incognito window, which I imagine created a new service worker instance. I actually had to remove service worker on SurfingNosara because of this issue, is there a way to cachebust the service worker on every Netlify deploy, or potentially tell Service Worker to disregard certain sections, or perhaps only cachebust certain sections so users don't get served outdated content?

LekoArts:
@rcaracaus Thanks for the notice, I addressed 1) with this PR:
https://github.com/gatsbyjs/gatsby/pull/10541

Issue #10483, opened 2018-12-15
[gatsby-transformer-screenshot] Adding an option for a fullpage screenshot


daydream05:
## Summary

Add a config option that would allow users to specify if they would like a full page screenshot of the given website. 

### Basic example
```jsx
// in your gatsby-config.js
module.exports = {
  plugins: [
    {
      resolve: `gatsby-transformer-screenshot`,
      options: {
        nodeTypes: [`StartersYaml`, `WhateverType`],
        fullPage: true,
      },
    },
  ],
}
```

### Motivation

Sites such as Dribbble or Behance sometimes include attachments that are full page design of a site. It would be nice to be able to do the same but in reverse, website first then the the design comes next.

Looking at the source code, I think it's as simple as modifying gatsby node and allowing the lambda function to accept fullPage parameters and passing it to puppeeteer
```js
  // inside lambda/index.js
  const screenshot = await page.screenshot({ fullPage: request.fullPage || false })
```

sidharthachatterjee:
Thank you for opening this @daydream05. Just took a look at the source, it is a simple fix indeed. 

We'd love to accept a PR for this. Would you be interested? 🙂

apvarun:
I can sent a PR for this right away!

sidharthachatterjee:
Thank you for adding support for this @apvarun 

@daydream05 Look out for the next release of `gatsby-transformer-screenshot`

daydream05:
Sweet! Amazing stuff @apvarun

PR #10517, opened 2018-12-17
Added fullpage option for puppeteer API call

apvarun:
Puppeteer's API already supports the `fullPage` API to take a screenshot of an entire page. This pull request proxies the parameter via the `gatsby-transformer-screenshot` to puppeteer.
Issue reference: #10483

### DIFF
--- a/packages/gatsby-transformer-screenshot/lambda/index.js
+++ b/packages/gatsby-transformer-screenshot/lambda/index.js
@@ -26,9 +26,11 @@ exports.handler = async (event, context, callback) => {
   const width = request.width || 1024
   const height = request.height || 768
 
+  const fullPage = request.fullPage || false
+
   const browser = await setup.getBrowser()
   exports
-    .run(browser, url, width, height)
+    .run(browser, url, width, height, fullPage)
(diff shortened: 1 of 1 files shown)


packages/gatsby-transformer-screenshot/lambda/index.js
```diff
@@ -37,13 +39,13 @@ exports.handler = async (event, context, callback) => {
     })
 }
 
-exports.run = async (browser, url, width, height) => {
+exports.run = async (browser, url, width, height, fullPage) => {
   console.log(`Invoked: ${url} (${width}x${height})`)
```
m-allanson: What do you think about adding fullPage to the log output? 

```
console.log(`Invoked: ${url} (${width}x${height}) fullPage:${fullPage}`)
```

Would it be useful or just unnecessary noise?

apvarun: I did notice that. Didn't add fullPage to it because it is not the part of keyBase, which is the important part of the function that would be useful debugging purpose.

PR #10529, opened 2018-12-18
added key to the container div

deltaskelta:


DSchau:
@deltaskelta could you explain why this PR is necessary and what it's solving? Thanks!

deltaskelta:
react requires the keys be present in the outermost containing component of the map iterator or else it throws warnings. 

screenshot 2018-12-18 11 45 59

PR #10536, opened 2018-12-18
fix(gatsby): add Safari 10 support to default terser options

### LINKED ISSUE #10513: Gatsby uglify not compatible with safari 10
## Description

Everthing works fine with my application expect that the finished build throws errors in
safari mobile <= 10. 
`SyntaxError: Cannot declare a const variable twice: 'e'`

if i skip the uglification with `gatsby build --no-uglify` there are no errors. 

Is there anyway to tell the uglifier to respect safari 10 rules like that?
```
    uglify: {
      uglifyOptions: {
        mangle: { safari10: true }
      }
    }
```

the only way that i can ship the app right now is without uglification which is not really a nice solution.

### Environment

```
  System:
    OS: macOS High Sierra 10.13.2
    CPU: (4) x64 Intel(R) Core(TM) i5-7267U CPU @ 3.10GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
  Browsers:
    Chrome: 71.0.3578.98
    Firefox: 63.0
    Safari: 11.0.2
  npmPackages:
    gatsby: ^2.0.63 => 2.0.63
    gatsby-link: ^2.0.7 => 2.0.7
    gatsby-plugin-catch-links: ^2.0.9 => 2.0.9
    gatsby-plugin-feed: ^2.0.11 => 2.0.11
    gatsby-plugin-google-analytics: ^2.0.8 => 2.0.8
    gatsby-plugin-manifest: ^2.0.11 => 2.0.11
    gatsby-plugin-nprogress: ^2.0.7 => 2.0.7
    gatsby-plugin-offline: ^2.0.18 => 2.0.18
    gatsby-plugin-sharp: ^2.0.14 => 2.0.14
    gatsby-remark-copy-linked-files: ^2.0.7 => 2.0.7
    gatsby-remark-images: ^3.0.1 => 3.0.1
    gatsby-remark-prismjs: ^3.1.4 => 3.1.4
    gatsby-remark-responsive-iframe: ^2.0.7 => 2.0.7
    gatsby-source-wordpress: ^3.0.19 => 3.0.19
    gatsby-transformer-remark: ^2.1.15 => 2.1.15
  npmGlobalPackages:
    gatsby-cli: 2.4.5
```

sidharthachatterjee:
Parcel sets `safari10` to `true` it by default

https://github.com/parcel-bundler/parcel/blob/75310e15f6fa3c8a0ce6593ada1bcca00240ea54/packages/core/parcel-bundler/src/transforms/terser.js#L13

fixes #10513 mentions that site breaks in Safari 10

### DIFF
--- a/packages/gatsby/src/utils/webpack-utils.js
+++ b/packages/gatsby/src/utils/webpack-utils.js
@@ -446,6 +446,7 @@ module.exports = async ({
       sourceMap: true,
       terserOptions: {
         ie8: false,
+        safari10: true,
         parse: {
           ecma: 8,
         },
--- a/packages/gatsby/src/utils/webpack-utils.js
+++ b/packages/gatsby/src/utils/webpack-utils.js
@@ -446,7 +446,9 @@ module.exports = async ({
       sourceMap: true,
       terserOptions: {
         ie8: false,
-        safari10: true,
+        mangle: {
+          safari10: true,
+        },
         parse: {
           ecma: 8,
         },


m-allanson:
I think this is a good fix, thanks @sidharthachatterjee 👍 

The overall options to choose from, I think, are:

- enable safari 10 support by default (done in this PR)
- update Gatsby so it’ll read in options from a terser config file (if it exists)
- create a Gatsby plugin that allows you to configure terser via your `gatsby-config.js` file

Other things to consider:

- https://caniuse.com/usage-table Says that Safari 10 usage is at 0.688% globally, which will only go down over time.
- If we enable Safari 10 support by default, removing it at a later date becomes a breaking change (maybe this is fine).
- Gatsby itself doesn't have any code that breaks in Safari 10 when minified, whether you see an error depends on your particular site.
- Having to track down errors like this in your own site can be really frustrating and time-consuming.

On balance I think this is probably the best fix as it prioritises _no errors_ and a lack of frustration ahead of configurability and bundle size. I'd be interested to hear what @dustin or @pieh think?

sidharthachatterjee:
Create React App also appears to set 

```
mangle: {
  safari10: true,
},

```
https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/config/webpack.config.js#L210

sidharthachatterjee:
Moved the option under `mangle` 

@pieh tested this with the starter and mentions in https://github.com/gatsbyjs/gatsby/issues/10513#issuecomment-448233421 that the resulting size was the same (presumably because no Safari 10 specific mangling was necessary)

Can also confirm that this fixes #10513

dustin:
I have no opinion on this

KyleAMathews:
This is exactly the sort of think browserlist is for. We should just read the site's current settings and use that to set the option.

PR #10540, opened 2018-12-18
support all of chokidar's path options for source-filesystem

jquense:
It seems expensive to set up lots of individual watchers when i want to be very specific about which files to source, Chokidar already supports this, we just need to let it


packages/gatsby-source-filesystem/src/create-file-node.js
```diff
@@ -7,21 +7,14 @@ const prettyBytes = require(`pretty-bytes`)
...
   const parsedSlashed = path.parse(slashed)
   const slashedFile = {
     ...parsedSlashed,
     absolutePath: slashed,
     // Useful for limiting graphql query with certain parent directory
-    relativeDirectory: path.relative(
-      pluginOptions.path || process.cwd(),
-      parsedSlashed.dir
-    ),
+    relativeDirectory: path.relative(process.cwd(), parsedSlashed.dir),
```
jquense: i'm unsure if this breaks anything but the conditional here was always true so this is now a path that wasn't hit before

pieh: yeah, this is definitely breaking change (at very least e2e tests for gatsby-image rely on current behaviour - well for the `relativePath` which was changed similarly).

currently `process.cwd` path is hit/used by `createRemoteFileNode` utility, where there are no pluginOptions (just for context on why it's here)

jquense: Ah ok, I'm not really sure I understand what the old behavior was doing/what it was for so I'm not sure if we can work around it. The simplest back compatible change tho would be to keep the old behavior if options.path is a single non glab path. 

Again tho I don't really understand what this is doing so maybe that still breaks a bunch of usage for folks that opt into this newer API path?

pieh: for the most part it was just setting `relative(Directory|Path)` fields as path relative to one passed in plugin options:

if we had:
```
{
  resolve: `gatsby-source-filesystem`,
  options: {
    name: `content`,
    path: `${__dirname}/src/content/`,
  }
}
```

then `src/content/blog/blogpost_1/index.md` would have `relativePath` as `blog/blogpost_1/index.md` (instead of path relative to project root).

we definitely can't work it out if `path` passed to pluginOptions is not straight path to directory, but is some kind of glob pattern, but I do feel like current behaviour is weird - especially if you have multiple instances of filesystem plugin, each pointing to different directory - then relativePath in global context is really hard to reason with

jquense: what is the value in the relative path here? It seems like it avoids situations where it's sourced from outside the `cwd`? I'd agree on the face of it it's weird that "relativePath" not being anchored always to the site root seems confusing. Maybe there is a better way to handle whatever usecase this is enabling?

sidharthachatterjee:
@jquense Thank you so much for opening this and we're so sorry for dropping the ball on this pull request!

We were just discussing this and here are the observations we're able to make:

Pros
- Being able to support all of the path formats that chokidar currently supports (we only support actual paths at the moment)

Cons
- Supporting this will likely be a breaking change because currently nodes depend on relative paths for  file resolution

Based on the observations made above, it doesn't seem like supporting these path formats is worth a new breaking version. As a result, we will be closing this for now! However, we would love to merge this in if we can transparently implement this without a change in behaviour. 

Thank you so much for your continued contributions and we look forward to many more 💜 💪

jquense:
fair enough, I will say tho this is an annoying pain when making themes, if I want to programmatically source stuff. In many cases to get this working we're adding tons of individual plugin instances.

PR #10541, opened 2018-12-18
docs(showcase): Update two tags

LekoArts:
Surf -> Surfing
React-Native -> React Native

### DIFF
--- a/docs/sites.yml
+++ b/docs/sites.yml
@@ -3568,7 +3568,7 @@
   description: >
     Welcome to Portugal's best kept secret and be amazed with our nature. Here you can explore, surf, taste the world's best gastronomy and wine, feel the North Canyon's power with the biggest waves in the world and so many other amazing things. Find us, discover yourself!
   categories:
-    - Surf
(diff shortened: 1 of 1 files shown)


docs/sites.yml
```diff
@@ -3568,7 +3568,7 @@
   description: >
     Welcome to Portugal's best kept secret and be amazed with our nature. Here you can explore, surf, taste the world's best gastronomy and wine, feel the North Canyon's power with the biggest waves in the world and so many other amazing things. Find us, discover yourself!
   categories:
-    - Surf
+    - Surfing
```
DSchau: We probably want to keep these as broad as possible, I'd probably just remove this.
```suggestion
```

DSchau: lol - yeah it's hard to honestly merge these sometimes because it's challenging to know what tags already exist and what are valid tags.

Surfing definitely is a one-off though, so let's def. remove that. I think the other tags are fine, e.g. Travel, Tourism, etc. seem reasonable.

PR #10546, opened 2018-12-18
Add pullquote citation

amberleyromo:
Adds a basic source citation to blog pullquotes, if supplied.

screen shot 2018-12-18 at 12 34 08 pm

### DIFF
--- a/www/src/components/shared/pullquote.js
+++ b/www/src/components/shared/pullquote.js
@@ -37,6 +37,14 @@ const PullquoteRoot = styled(`blockquote`)`
   }
 `
 
+const Citation = styled(`cite`)`
+  display: block;
+  text-align: right;
+  font-weight: normal;
+  font-style: italic;
+  margin-top: 1rem;
+`
+
 const QuotationMark = styled(`span`)`
   display: flex;
   left: 2.5rem;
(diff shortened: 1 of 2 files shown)


www/src/components/shared/pullquote.js
```diff
@@ -155,13 +163,14 @@ const Star = styled(`span`)`
...
 
-const Pullquote = ({ children }) => {
+const Pullquote = ({ citation, children }) => {
   instancesCounter += 1
   const className = `variant${variants[instancesCounter % variants.length]}`
 
   return (
     <PullquoteRoot className={className}>
       {children}
+      {citation && <Citation>- {citation}</Citation>}
```
DSchau: nit: but should this be an emdash or endash maybe?

```suggestion
       {citation && <Citation>&mdash;{citation}</Citation>}
```

amberleyromo:
Hmm. @shannonbux do you have any opinion on convention for source citation? (see @DSchau's note above). 

Personally I think it's situational. It probably makes sense to have anything from the Gatsby team be noted as such.

www/src/components/shared/pullquote.js
```diff
@@ -37,6 +37,14 @@ const PullquoteRoot = styled(`blockquote`)`
   }
 `
 
+const Citation = styled(`cite`)`
+  display: block;
+  text-align: right;
```
greglobinski: A nit. The alphabetical order of props, so `text-align` goes at the end. 

I also do not care in what order the properties like `font-weight` and `font-style` goes. Only the first character counts for me.

amberleyromo: Lol, done. Thanks for enforcing.

shannonbux:
This looks great. I do think the company would be nice to know!

Issue #10480, opened 2018-12-14
Building fails under incredibly specific circumstances


dstaley:
## Description

When the following conditions are true, Gatsby is unable to build the site, reporting GraphQL validation failures:

1. Use the latest version of `gatsby-remark-images` (`v3.0.1`).
1. Use the `gatsby-paginate` plugin with the suggested template.
1. Include the field `htmlAst` in the `createPages` query.
1. Have a markdown file containing specific content.

### Steps to reproduce

```bash
git clone https://github.com/dstaley/gatsby-starter-blog && cd gatsby-starter-blog
npm i && npm start
```

### Expected result

Gatsby builds the site and begins serving.

### Actual result

```
success open and validate gatsby-configs — 0.008 s
success load plugins — 0.286 s
success onPreInit — 1.662 s
success delete html and css files from previous builds — 0.195 s
success initialize cache — 0.012 s
success copy gatsby files — 0.090 s
success onPreBootstrap — 0.008 s
success source and transform nodes — 0.090 s
success building schema — 0.400 s
success createPages — 0.170 s
success createPagesStatefully — 0.035 s
success onPreExtractQueries — 0.004 s
success update schema — 0.178 s
error Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but "" does not.

Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but "" does not.

Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but "" does not.


  Error: Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but "" does not.
  Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but "" does not.
  Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but "" does not.

  - query-compiler.js:145
    [gatsby-md]/[gatsby]/dist/internal-plugins/query-runner/query-compiler.js:14    5:44

  - Generator.next

  - debuggability.js:313 Promise._execute
    [gatsby-md]/[bluebird]/js/release/debuggability.js:313:9

  - promise.js:483 Promise._resolveFromExecutor
    [gatsby-md]/[bluebird]/js/release/promise.js:483:18

  - promise.js:79 new Promise
    [gatsby-md]/[bluebird]/js/release/promise.js:79:10

  - query-compiler.js:213 Runner.write
[...]

PR #10557, opened 2018-12-19
fix(gatsby): Ignore fields with key set to empty string in getExampleValues

sidharthachatterjee:
Unfortunately, an Object property with an empty string as the key is valid JavaScript. This however breaks schema generation because `graphql` asserts that every key has a [valid](https://github.com/graphql/graphql-js/blob/958eb961513de1147c19508d66f97920c26e99e5/src/type/validate.js#L220) name. 

### Example scenario

[hast-util-raw](https://github.com/syntax-tree/hast-util-raw) seems to produce an AST with `"": ""` (which breaks build) in #10480 

This fix makes schema generation more defensive by ignoring all instances of `key === ""`

### DIFF
--- a/packages/gatsby/src/schema/data-tree-utils.js
+++ b/packages/gatsby/src/schema/data-tree-utils.js
@@ -244,6 +244,9 @@ const extractFieldExamples = (
   const example = {}
   for (let key of allKeys) {
     if (ignoreFields.includes(key)) continue
+    // Unfortunately an empty string as an Object key is perfectly valid JavaScript
+    // We ignore these because our schema generation requires keys to have names
+    if (key === ``) continue
     const nextSelector = selector ? `${selector}.${key}` : key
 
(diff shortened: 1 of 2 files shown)


packages/gatsby/src/schema/data-tree-utils.js
```diff
@@ -244,6 +244,9 @@ const extractFieldExamples = (
   const example = {}
   for (let key of allKeys) {
     if (ignoreFields.includes(key)) continue
+    // Unfortunately an empty string as an Object key is perfectly valid JavaScript
+    // We ignore these because our schema generation requires keys to have names
```
pieh: it's requirement of `graphql` (not ours :) )

sidharthachatterjee: Updated the comment

PR #10561, opened 2018-12-19
fix: don't change default node_modules resolution

jquense:
I feel like i fixed this before, but couldn't find anything :P 

we can't do this. what happens is that module resolution _always_ checks the root node_modules before any other `node_modules`, this essentially breaks any dependency that shares a root dep of a different version

```
/node_modules
 - lodash@v4
 - some-pkg
   - node_modules
     - lodash@v3
```

Given the above structure `some-pkg` will resolve lodash@4 *not* lodash@3. when there are incompatible versions like this it causes extremely hard to debug issues, because the dep tree on disk is correct.


jquense:
awesome, glad this didn't break anything :P

PR #10565, opened 2018-12-19
fix(gatsby): allow environment variables to be replaced per environment

### LINKED ISSUE #10563: Environment variables are not captured utilizing NODE_ENV in Webpack
## Description

[Environment variables](https://www.gatsbyjs.org/docs/environment-variables/) (captured in `.env.ENVIRONMENT`, e.g. `.env.staging`) do not get exposed correctly to the code that gets bundled with Webpack, e.g. the client-side, non-Node code.

We can see this by walking through the build process:

1. Add a script, `build:staging`

    ```json
    {
      "scripts": {
        "build:staging": "cross-env NODE_ENV=staging npm run build"
      }
    }
    ```

1. `gatsby build` is kicked off via the CLI [here](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-cli/src/create-cli.js#L177)
    - This _explicitly_ sets NODE_ENV=production, therefore ignoring any overrides from the user
1. Environment variables are sourced via Webpack [here](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/utils/webpack.config.js#L40)
    - This will _always_ be production; it's not possible to override this since it's set and overwritten early in the build via the CLI

### Steps to reproduce

1. [Clone this repo](https://github.com/dschau/gatsby-env-vars-example)
1. Run `yarn` or `npm install`
1. Run `yarn build:staging`
1. The environment variable `API_KEY` in `.env.staging` should be present in the built code

### Expected result

The environment variable `API_KEY` from `.env.staging` should be present in `public/index.html`

### Actual result

The environment variable `API_KEY` from `.env.production` is used.

## The Fix

I don't believe we can overwrite `NODE_ENV` as `production` is checked fairly often in various plugins/APIs/etc. Using `process.env.NODE_ENV = process.env.NODE_ENV || 'production'` in `gatsby-cli` seems to break the build.

As such - I think there are two things that we should do here to remediate the issue.

1. Tweak [this code](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/utils/webpack.config.js#L40) to check for an environment override _outside_ of `NODE_ENV`
    - Perhaps `ACTIVE_ENV` as we [document here](https://www.gatsbyjs.org/docs/environment-variables/#example-1)
1. Tweak any documentation to make it clear that this new environment variable must be used if you want to use a separate .env file for a "production" build

I don't love the indirection and extra steps here--but not sure I see a better solution off hand. Opinions more than welcome here!

cc @xjamundx

DSchau:
## Description

This PR introduces a mechanism to source environment variables from an `ACTIVE_ENV` environment variable override. If the value is not set, the flow will be the same, e.g. `.env.production` will be used in `gatsby build` and `gatsby serve` and `.env.development` will be used in `gatsby develop`.

However, if an override variable `ACTIVE_ENV` is set, this will be used to source the `.env` file. For example, given `ACTIVE_ENV=staging gatsby build`, gatsby will attempt to source the `.env.staging` file rather than the the `.env.production` file.

## Further info

I'm not really sold on `ACTIVE_ENV` so suggestions welcome. I don't really like the idea of exposing some other non-standard variable, but think this seems more maintainable and better than the alternatives (which I've outlined below in the related issue)

## Related Issues

Fixes #10563


packages/gatsby/src/utils/__tests__/webpack.config.js
```diff
@@ -0,0 +1,115 @@
...
+          {}
+        )
+      )
+    )
+  })
+})
+
+describe(`environment variables`, () => {
+  afterEach(() => {
+    delete process.env.ACTIVE_ENV
```
DSchau: TIL that setting (and then unsetting via `process.env.SOME_ENV = undefined`) still makes that value available as a string, e.g.

```js
process.env.ACTIVE_ENV = `whatever`

// later

process.env.ACTIVE_ENV = undefined

expect(process.env.ACTIVE_ENV).toBe(`undefined`) // true
```

m-allanson:
Maybe call this GATSBY_ACTIVE_ENV?

wardpeet:
@DSchau this looks ready to go. Could you fix the merge conflict ^^

DSchau:
@wardpeet let's do it :) Thanks!

DSchau:
Successfully published:
 - gatsby@2.1.33

joefiorini:
Coming to this late, there are a couple issues open now because this PR causes `GATSBY_ACTIVE_ENV` to override `NODE_ENV`, which I can't imagine was intentional. I think it's a perfectly valid pattern to have two separate environment variables, one that names the environment in which the code lives and the other indicates the mode it should be built with. The latter is what we typically use `NODE_ENV` for, but the former doesn't really have a standard name yet. I worked on a project that heavily used Kubernetes and we called it `ENVIRONMENT_NAME`. `ACTIVE_ENV` works too. Either way, I think that the two should remain separate rather having one override the other. Anyone opposed to a followup PR changing this so that `GATSBY_ACTIVE_ENV` does not override `NODE_ENV` when it gets passed to `DefinePlugin`?

wardpeet:
seems like a bug indeed! fixing

PR #10578, opened 2018-12-20
Fix gatsby-plugin-react-helmet for v6.0.0

Ehesp:
## Description

In version 6.0.0 of react-helmet, the default export has been removed. There is now only a named export, e.g.

```
import { Helmet } from 'react-helmet';
```

In v5.x.x, both a default export and named are exported. Change is here:

https://github.com/nfl/react-helmet/commit/20ea3858f503ec521ad5b1facdac5816a0019317#diff-cc54072daf5278847980b841520c8fffL287

This is a backwards compatible change. v5 has an annoying bug (https://github.com/nfl/react-helmet/issues/373) which v6 seems to fix.

### DIFF
--- a/docs/blog/2017-07-19-creating-a-blog-with-gatsby/index.md
+++ b/docs/blog/2017-07-19-creating-a-blog-with-gatsby/index.md
@@ -260,7 +260,7 @@ We'll want to create the file `src/templates/blog-post.js` (please create the
 
 ```javascript:title=src/templates/blog-post.js
 import React from "react"
-import Helmet from "react-helmet"
+import { Helmet } from "react-helmet"
 
 // import '../css/blog-post.css'; // make it pretty!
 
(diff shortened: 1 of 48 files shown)


DSchau:
I can't actually seem to get this to successfully build with `gatsby-default-starter`, the changes in this PR, and then `react-helmet@6` (currently just a beta?).

I'll investigate this a bit further.

DSchau:
Yeah - I can't seem to get this to build whatsoever.

Did you have any luck? Simply updating to react-helmet@6.0.0-beta seems to break the build, even with this plugin's changes. In fact, it fails to fail before it even invokes the SSR APIs.

Ehesp:
Hey sorry was in a rush yesterday. Yeah 6.0 is in beta but there was a very annoying issue it fixed. Thanks for sorting 👍

PR #10582, opened 2018-12-20
Update create transformer plugin doc

amberleyromo:
cc @KyleAMathews, updated with an example.

### DIFF
--- a/docs/docs/create-transformer-plugin.md
+++ b/docs/docs/create-transformer-plugin.md
@@ -10,7 +10,10 @@ and "transformer" plugins.
 - **Transformer** plugins "transform" data provided by source plugins into new
   nodes and/or node fields.
 
- This doc focuses on transformer plugins.
+The purpose of this doc is to:
+
+1. Define what a Gatsby transformer plugin is, and
+2. Walk through a simplified reimplementation of an existing plugin, to demonstrate how to create a transformer plugin.
 
 ## What do transformer plugins do?
 
@@ -18,7 +21,7 @@ Transformer plugins "transform" data of one type into another type. You'll often
 
 This loose coupling between the data source and transformer plugins allow Gatsby developers to quickly assemble complex data transformation pipelines with little work.
 
-## What does the code look like?
+## How do you create a transformer plugin?
 
 Just like a source plugin, a transformer plugin is a normal NPM package. It has a `package.json` file with optional dependencies as well as a `gatsby-node.js` file where you implement Gatsby's Node.js APIs.
 
--- a/docs/docs/create-transformer-plugin.md
+++ b/docs/docs/create-transformer-plugin.md
@@ -39,7 +39,7 @@ As an example, let's partially rebuild a simplified `gatsby-transformer-yaml` di
 
 ### Make sure the data is sourced
 
-First, we make sure the file is sourced, using `gatsby-source-filesystem` in `gatsby-config.js`:
+First, in `gatsby-config.js`, use the `gatsby-source-filesystem` plugin to create File nodes.
 
 ```javascript:title=gatsby-config.js
 module.exports = {
@@ -54,7 +54,7 @@ module.exports = {
 }
 ```
 
-Now if we query `allFile`...
+These are exposed in your graphql schema which you can query:
 
 ```graphql
 query {
@@ -73,7 +73,7 @@ query {
 }
 ```
 
-... We see we have the sourced file.
+Now you have a `File` node to work with:
 
 ```json
 {
@@ -99,7 +99,7 @@ query {
 
 ### Transform nodes of type `text/yaml`
 
-Now we can transform the sourced data by hooking into the `onCreateNode` API in `gatsby-node.js`.
+Now, transform the newly created `File` nodes by hooking into the `onCreateNode` API in `gatsby-node.js`.
 
 If you're following along in an example project, install the following packages:
 
--- a/docs/docs/create-transformer-plugin.md
+++ b/docs/docs/create-transformer-plugin.md
@@ -16,11 +16,7 @@ and "transformer" plugins.
 
 Transformer plugins "transform" data of one type into another type. You'll often use both source plugins and transformer plugins in your Gatsby sites.
 
- This loose coupling between the data source and transformer plugins allow Gatsby developers to quickly assemble complex data transformation pipelines with little work.
-
-## Transformer plugin example
-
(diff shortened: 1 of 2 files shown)


docs/docs/create-transformer-plugin.md
```diff
@@ -2,36 +2,273 @@
...
 
-The [`gatsby-transformer-remark`](/packages/gatsby-transformer-remark/) plugin
-"transforms" markdown nodes (fetched for example from the filesystem source plugin) into html and provides additional node details such as for example the `timeToRead` the `tableOfContents` and the `excerpt`.
+The [`gatsby-transformer-remark`](/packages/gatsby-transformer-remark/) plugin "transforms" markdown nodes (fetched for example from the filesystem source plugin) into html and provides additional node details such as for example the `timeToRead` the `tableOfContents` and the `excerpt`.
 
 ## What do transformer plugins do?
 
-Each transformer plugin is responsible for "enhancing" the existing source nodes they
-are provided to provide additional information. Therefore, you'll often use both source plugins and transformer plugins in your Gatsby sites.
+Each transformer plugin is responsible for "enhancing" the existing source nodes they are provided to provide additional information. Therefore, you'll often use both source plugins and transformer plugins in your Gatsby sites.
```
KyleAMathews: Enhancing isn't the right word. Transformers take data and we'll, "transform" it into a different type. Typically from something not JavaScript into JavaScript. Or something not html into html. Transformers can enhance data but that's an unusual case.

amberleyromo: Edited and rearranged a bit.

docs/docs/create-transformer-plugin.md
```diff
@@ -2,36 +2,272 @@
...
+
+In our updated `gatsby-node.js`, we'll then iterate through the parsed YAML content, using the helper function to transform each into a new node:
+
+```javascript:title=gatsby-node.js
+const jsYaml = require(`js-yaml`)
+const _ = require(`lodash`)
+
+async function onCreateNode({
+  node,
+  actions, // highlight-line
```
DSchau: This highlight-line stuff is _so_ much nicer. Thanks for using it 👍

docs/docs/create-transformer-plugin.md
```diff
@@ -2,36 +2,272 @@
...
+- name: Jane Doe
+  bio: Developer based in Somewhere, USA
+- name: John Smith
+  bio: Developer based in Maintown, USA
+```
+
+### Make sure the data is sourced
 
-Your `gatsby-node.js` should look something like:
+First, we make sure the file is sourced, using `gatsby-source-filesystem` in `gatsby-config.js`:
```
shannonbux: I'm thinking "you" could still be an advantageous point of view for these docs for consistency and clarity (to focus on the reader as the agent/actor who does the actions).

docs/docs/create-transformer-plugin.md
shannonbux: I agree, and I haven't even used it yet! It just seems way easier than squinting my eyes and hoping I identified the correct numbers of lines to highlight haha

docs/docs/create-transformer-plugin.md
```diff
@@ -2,36 +2,272 @@
 title: Create a Transformer Plugin
 ---
 
-Transformer plugins are responsible for "transforming" data provided by
-[source plugins](/docs/create-source-plugin) into new nodes and/or node fields.
+There are two types of plugins that work within Gatsby's data system, "source"
+and "transformer" plugins.
 
-## Transformer plugin example
```
shannonbux: Oh, forgot to mention that I wasn't clear on this header and section -- is this example provided because I'm supposed to imitate it? Knowing why this example is included would help clarify why it's introduced here

docs/docs/create-transformer-plugin.md
```diff
@@ -2,36 +2,272 @@
...
+Transformer plugins "transform" data of one type into another type. You'll often use both source plugins and transformer plugins in your Gatsby sites.
+
+This loose coupling between the data source and transformer plugins allow Gatsby site builders to quickly assemble complex data transformation pipelines with little work.
+
+## Transformer plugin example
 
-This loose coupling between the data source and transformer plugins allow Gatsby
-site builders to quickly assemble complex data transformation pipelines with
-little work.
+The [`gatsby-transformer-remark`](/packages/gatsby-transformer-remark/) plugin "transforms" markdown nodes (fetched for example from the filesystem source plugin) into html and provides additional node details such as for example the `timeToRead` the `tableOfContents` and the `excerpt`.
```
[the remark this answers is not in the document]
KyleAMathews: Perhaps "typically provided by the `gatsby-source-filesystem` plugin"

docs/docs/create-transformer-plugin.md
KyleAMathews: 👍

KyleAMathews: Perhaps instead of saying "sourced" we make the process explicit here e.g. `gatsby-source-filesystem` creates `File` nodes which are exposed in our graphql schema which we can query. That'll help clarify that when we transform the File nodes into another node type, those too will automatically be exposed through GraphQL.

docs/docs/create-transformer-plugin.md
```diff
@@ -2,36 +2,272 @@
...
+        }
+      ]
+    }
+  }
+}
+```
+
+### Transform nodes of type `text/yaml`
+
+Now we can transform the sourced data by hooking into the `onCreateNode` API in `gatsby-node.js`.
```
KyleAMathews: instead of "sourced data" — use Gatsby terminology "transform the newly created `File` nodes"

docs/docs/create-transformer-plugin.md
amberleyromo: Probably makes sense to remove it. We're defining what transformer plugins are before moving into the "create" content, but an example is probably not necessary. The leading section page intros this.

amberleyromo:
@shannonbux at the moment, it's following the same sort of pattern as the existing "create source plugin" doc.

I can make a couple wording tweaks to make the intro more clear (re: your comment about purpose) but the doc:

1) includes an overview definition of what transformer plugins are
2) explains what a transformer plugin is
3) walks through a reimplementation of an existing transformer plugin.

calcsam:
Took a look on Shannon's behalf and looks good! Going to merge this!

PR #10585, opened 2018-12-20
Fix broken Gatsby Starters link

sloreti:
## Description
Updates broken Gatsby Starters link in "Adding Markdown Pages" docs to new route.


amberleyromo:
Nice catch! How about we link to it with the markdown category prefiltered? https://www.gatsbyjs.org/starters?c=Markdown

sloreti:
Even better. Updated.

PR #10601, opened 2018-12-21
Make emotion available for import in project

andrewtdinh:
Include instruction to install `emotion` so that it's available for import in the project.  The command was inadvertently left out of master


andrewtdinh:
Hi @DSchau,

Yeah, I do see it in part four now that you mention it.  I wonder how I missed it.  Sorry!

PR #10604, opened 2018-12-21
Update content in starter READMEs

sidharthachatterjee:
Updated content in all READMEs by generating them from README-template

- Removed unnecessary point about `yarn.lock` (since we don't include them in starters anymore)


sidharthachatterjee:
Looks like @DSchau is generating README files or something

DSchau:
Oh - yeah, so we don’t need to do this. The READMEs are automatically generated from the readme-template.

What we should do however is wipe out the README content in each of the starters—apart from the comments for auto generated content, which is how markdown magic works its... magic!

sidharthachatterjee:
@DSchau Updated this to remove unnecessary markdown in each README

sidharthachatterjee:
@DSchau Re-added generated markdown in the repo since people might want to read a starter's README in GitHub

PR #10605, opened 2018-12-21
fix(gatsby-plugin-sitemap): add missing dependency pify and minimatch

MartinHelmut:
## Description

The npm module `pify` is required inside the `gatsby-plugin-sitemap` but not added as a dependency. It is implicitly required by the installation of the dependencies of `gatsby-plugin-feed`, therefore nothing fails until a strict module handling is used, e.g. by pnpm.

Therefore I added it to the `package.json` of `gatsby-plugin-sitemap` to fix this.

### DIFF
--- a/packages/gatsby-plugin-sitemap/package.json
+++ b/packages/gatsby-plugin-sitemap/package.json
@@ -8,6 +8,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.0.0",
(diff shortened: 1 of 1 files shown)


pieh:
Looks good - could you add `minimatch` as well?

MartinHelmut:
Ohh `minimatch` is missing as well? Okay, I will check that and add another commit 🙂

PR #10612, opened 2018-12-21
node/npm screenshot version update

NoriSte:
## Description

A colleague of mine has been led to the error from the version shown in the old screenshot. With the versions of the old screenshot, the installation fails

## Related Issues

I fixed all my own the issue but if you'd like to have a detailed description... below you can find the steps to reproduce the issue that make me submit this PR

$ node --version
v8.11.1 # it happens with v8.11.3 too
$ mkdir gastby
$ cd gatsby
$ npx gatsby new gatsby-site
$ cd gatsby-site
$ yarn gatsby develop

```bash
success open and validate gatsby-configs — 0.006 s
success load plugins — 0.163 s
success onPreInit — 0.479 s
success delete html and css files from previous builds — 0.078 s
success initialize cache — 0.007 s
success copy gatsby files — 0.045 s
success onPreBootstrap — 0.003 s
success source and transform nodes — 0.046 s
success building schema — 0.236 s
success createPages — 0.000 s
success createPagesStatefully — 0.036 s
success onPreExtractQueries — 0.005 s
success update schema — 0.116 s
success extract queries from components — 0.104 s
error UNHANDLED EXCEPTION


  TypeError: queue.off is not a function
  
  - page-query-runner.js:162 Queue.onDrain
    [gatsby-site]/[gatsby]/dist/internal-plugins/query-runner/page-query-runner.js:162:13
  
  - queue.js:469 Queue._drained
    [gatsby-site]/[better-queue]/lib/queue.js:469:8
  
  - queue.js:505 Queue._processNextIfAllowed
    [gatsby-site]/[better-queue]/lib/queue.js:505:14
  
  - queue.js:489 Timeout._onTimeout
    [gatsby-site]/[better-queue]/lib/queue.js:489:12
  

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

### Environment

```bash
System:
    OS: macOS High Sierra 10.13.6
    CPU: (8) x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 8.11.1 - /usr/local/bin/node
    Yarn: 1.9.2 - /usr/local/bin/yarn
    npm: 5.6.0 - /usr/local/bin/npm
  npmPackages:
    gatsby: ^2.0.53 => 2.0.74 
    gatsby-image: ^2.0.20 => 2.0.25 
[...]

PR #10593, opened 2018-12-21
fix(gatsby): fix eventemitter leak in page query runner

rexxars:
## Description

The query queue can fire the `drain` event multiple times, and the page query runner listens for the drain event and resolves a promise once it does. When this happens, it doesn't currently clear the resolve event handler from the queue emitter.

This PR fixes that by simply unsubscribing once the drain event is fired.

I think this may properly fix #5510

PR #10613, opened 2018-12-21
fix(query-runner): replace .off with .removeListener

wKovacs64:
## Description

This fixes a bug introduced in #10593 by replacing the `.off` call with `.removeListener`. `.off` was introduced in Node v10.0.0 as an alias for `.removeListener` (https://github.com/nodejs/node/pull/17156).

## Related Issues

Related to #10612


DSchau:
Going to wait for the Appveyor to build--that should be _proof_ that it's working!

DSchau:
```
Successfully published:
 - gatsby@2.0.75
```

PR #10624, opened 2018-12-22
fix(gatsby-source-wordpress): adjust how endpoint urls are constructed to fix fetching for wordpress.com hosted sites and proxied urls

### LINKED ISSUE #10427: [wordpress-source-plugin] Self Tag is not returning proper URL
## Description
I have two apps Using Gatsby JS and the Wordpress-Source-Plugin. One of them is working just fine authenticating and loading the proper WP routes into the GraphQL Schema. This one was created a year or more ago. I also have a newer Wordpress blog that I am using as a data source, however when Authenticating this one I am seeing issues in the Schema. The default route here:
 https://public-api.wordpress.com/wp/v2/sites/[WORDPRESS-BLOG]
Normally returns data like this:
```json
{
  "namespace": "wp\/v2",
  "routes": {
    "\/wp\/v2\/sites\/WORDPRESSBLOG\/posts": {
      "namespace": "wp\/v2",
       ...
      "_links": {
        "self": "https:\/\/public-api.wordpress.com\/wp\/v2\/sites\/WORDPRESSBLOG\/posts"
      }
   ...
}
```

However, in my case, on the newly created Wordpress blog, the default indexing path is returning `https://public-api.wordpress.com/` for EVERY Self property.

This could be a bug on Wordpress's Side, but the fix for me for now was to change the `getRawEntityType` function in `fetch.js` to this:

```javascript
const getRawEntityType = route => route.substring(route.lastIndexOf(`/`) + 1, route.length);
```

While also passing in the path Key to that function on line 615 again in `fetch.js`
```javascript
const entityType = getRawEntityType(key) // Excluding the "technical" API Routes
```
I can provide a PR with this fix if it is valid, but I could just be missing something simple.
### Steps to reproduce
1. Create a new Wordpress Blog, file an API key for it to be able to query things like Media Routes
2. Add the `wpcom_app_clientSecret`, `wpcom_app_clientId`, `wpcom_user`, `wpcom_pass` to `gatsby-config.js` under the Wordpress plugin.
2. Run `gatsby develop` 

### Expected result

The `gatsby-source-wordpress` plugin should pull all data from the routes provided by ` https://public-api.wordpress.com/wp/v2/sites/[WORDPRESS-BLOG]`

### Actual result

Gatsby loads all of the routes from the same URL, if the self key is not correctly representing the key.

### Environment

```
  System:
    OS: Windows 10
    CPU: (4) x64 Intel(R) Core(TM) i7-6500U CPU @ 2.50GHz
  Binaries:
    Yarn: 1.5.1 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.4.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 42.17134.1.0
  npmPackages:
    gatsby: ^2.0.66 => 2.0.66
    gatsby-image: ^2.0.20 => 2.0.20
    gatsby-plugin-offline: ^2.0.15 => 2.0.19
    gatsby-plugin-react-helmet: ^3.0.2 => 3.0.2
    gatsby-plugin-sharp: ^2.0.12 => 2.0.12
    gatsby-plugin-typography: ^2.2.1 => 2.2.1
    gatsby-source-filesystem: ^2.0.12 => 2.0.12
    gatsby-source-wordpress: ^3.0.20 => 3.0.20
    gatsby-transformer-sharp: ^2.1.8 => 2.1.8
```

lukemorton:
## Description

As reported in #10427 some WordPress.com sites return a `_links.$.self` with a fullpath `https://public-api.wordpress.com/wp/v2/sites/$site/users/me` and others without `https://public-api.wordpress.com/`. This means it is not a reliable way of removing the `baseUrl` from the `fullUrl`.

Instead we pass the route key in as the `fullPath` rather than the `fullUrl`. We then extract the `basePath` from the `baseUrl` before finally removing the `basePath` from the `fullPath`.

We also need to extract the raw entity type from the `fullPath` rather than route.

We also need to build the `fullUrl` from `baseUrl` and `fullPath` so I introduced the `buildFullUrl` function to do just that.

I’ve included tests for both WordPress.com and WordPress.org.

## Related Issues

Fixes #10427.


pieh:
This is pretty wild on wp.com side (maybe it was changed in .org / self-hosted too in recent wp version?). Do you perhaps have wp.com I could use to verify everything works correctly with proposed changes

pieh:
Status update - there are currently some issues with non-wp.com sites with proposed changes (missing `/wp-json` path fragment in front of endpoint urls), but this whole issue with `_links.$.self` pointing to `https://public-api.wordpress.com/` for every endpoint (actually all links point to same url) does seem like a bug on wordpress.com side of things.

There is support ticket https://en.forums.wordpress.com/topic/rest-api-returning-different-self-links/ created by PR author, but it wasn't responded yet, so I'd like to hold on a bit with taking action on this. For anyone else hitting the issue - in meantime You can use @lukemorton's fork of the plugin https://www.npmjs.com/package/@madetech/gatsby-source-wordpress that contains changes proposed in this PR.

cardiv:
I can confirm this is not a bug on WordPress.com – I thought so too in the beginning.

But after testing with Postman and removing `_accessToken` from the `axios` header, all responses are good.

See #11021 for more details.

zilahir:
it does not seem that @lukemorton's PR working either :( 

```
⠁ source and transform nodesRoute discovered : /wp/v2
error Plugin @madetech/gatsby-source-wordpress returned an error
  ReferenceError: URL is not defined
```

zilahir:
i investigatet this a bit further, after adding the missing `const URL = require('url').URL;`, it seems, that  the `checkRouteList` always returning false, so the condition `validRoute = inWhiteList && !inBlackList;` will *never* evaluate to true, if the `includedRoutes` defined as the following (from the `documentation`: 

```
includedRoutes: [
          "**/*/*/categories",
          "**/*/*/posts",
          "**/*/*/pages",
          "**/*/*/media",
          "**/*/*/tags",
          "**/*/*/taxonomies",
          "**/*/*/users",
          "/wp/v2/*/*"
        ],
```

kharrop:
Is this currently being worked on? I ended up resolving the issues by removing instance of _accessToken from the fetch.js file.

zilahir:
> Is this currently being worked on? I ended up resolving the issues by removing instance of _accessToken from the fetch.js file.

what exactly you removed? The issue on my side was resolved aswell, but that was crealy an issue with the `includedRoutes`

kharrop:
@zilahir I was referencing https://github.com/gatsbyjs/gatsby/issues/11021#issuecomment-453766094

kharrop:
> But after testing with Postman and removing `_accessToken` from the `axios` header, all responses are good.
> 
> See #11021 for more details.

Sorry, I don't understand how to remove _accessToken; could you please elaborate?

kharrop:
Just pulled in fetch.js and it resolved all issues with pulling content from wp.com. Is there a reason why this hasn't been merged in yet?

packages/gatsby-source-wordpress/src/__tests__/fetch.js
```diff
@@ -24,16 +24,16 @@ describe(`Fetching`, () => {
...
-    const fullUrl = fetch.buildFullUrl(baseUrl, key)
+    const fullUrl = fetch.buildFullUrl(baseUrl, key, true)
     expect(fullUrl).toEqual(
       `https://public-api.wordpress.com/wp/v2/sites/example.wordpress.com/posts`
     )
   })
 
   it(`builds full URL correctly for wordpress.org`, () => {
-    const key = `/wp-json/wp/v2/posts`
+    const key = `/wp/v2/posts`
```
pieh: this is how key actually looks like for .org sites

PR #10631, opened 2018-12-23
Change class names configurations

vadimshvetsov:
## Description

In [`gatsby-remark-custom-blocks`](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-remark-custom-blocks) docs outputed class names are wrong, because if we set this:
```js
options: {
  blocks: {
    danger: {
      classes: "custom-block-danger"
    // ...
}
```
we will get classes in html: `custom-block custom-block-danger` and not `custom-block-danger` only as in docs.

This PR fix the confusion

### DIFF
--- a/packages/gatsby-remark-custom-blocks/README.md
+++ b/packages/gatsby-remark-custom-blocks/README.md
@@ -33,10 +33,10 @@ plugins: [
           options: {
             blocks: {
               danger: {
-                classes: "custom-block-danger",
+                classes: "danger",
               },
               info: {
-                classes: "custom-block-info",
+                classes: "info",
                 title: "optional",
               },
             },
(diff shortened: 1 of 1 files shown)


pieh:
Maybe we could do this other way around - change the output in docs instead of config? Just quick scanning the example it can be ambiguous if the `info`/`danger` class name come from `classes` field or from block name with proposed changes. What do you think?

vadimshvetsov:
If this way is more preferable I can change it. But I think that class `custom-block danger` looks better and less verbose then `custom-block custom-block-danger` thats why I was changed config section.

pieh:
It's no big deal either way. I'm happy to merge it as-is and will just do that now. It certainly is wrong currently :)

PR #10642, opened 2018-12-24
Add Built with Gatsby footer to blog and default starter

sidharthachatterjee:
Left this out of the hello world starter because it is (and ought to be) as minimal as possible. We might want to style the text a little better before merging this in.


KyleAMathews:
What's it look like now?

sidharthachatterjee:
@KyleAMathews 
screenshot 2018-12-24 21 31 08
screenshot 2018-12-24 21 32 27

PR #10650, opened 2018-12-25
[gatsby-plugin-sharp] Url Encoding Image File Names

huntercaron:
Uses `encodeURI()` to fix an issue with image files that contain spaces in the name breaking srcset.

Client was adding images to the CMS backend with spaces in the name, which were breaking the srcset image links. This encodes the file names with the appropriate `%20` and such.

(if there is a better place to put it please suggest!)

## Before
beforeerror
beforeinspect

## After
afterinspect

Thanks!


huntercaron:
Hi @LekoArts , thanks for the help!

I've never written a test before 😛but it was a cool & useful thing to learn! 
It handles quotes 'n such now and I cleaned it up a bit.

Lemme know if it looks good to go!

huntercaron:
hmm not sure whats going on re: these checks. `lint:other` is working file locally :/

huntercaron:
Thanks @pieh!

PR #10666, opened 2018-12-26
Make docs sidebar more browsable

calcsam:
This PR:
* ~Rearranges "Deploying and hosting" to where it would logically go in the progression of site development (ie, the end)~
* Combines the "Use cases" and "website functionality" sections to simplify things
* Retitles sections to tell a story of website development & improve consistency ("Headless CMS" => "Sourcing from Headless CMS").


KyleAMathews:
The deployment section is at the top because a) that's one of the first things people evaluate when looking at a new tool and b) it's also generally one of the first things people do when starting a project is setup a deployment pipeline.

/cc @shannon_bux for evaluating the rest.

PR #10667, opened 2018-12-26
Update debugging docs organization

calcsam:
@pieh @m-allanson I've rearranged and retitled these sections to map to how people will likely logically think about debugging (in development / in production).

Thoughts?

### DIFF
--- a/www/src/data/sidebars/doc-links.yaml
+++ b/www/src/data/sidebars/doc-links.yaml
@@ -178,13 +178,13 @@
     - title: Debugging
       link: /docs/debugging/
       items:
-        - title: Debugging HTML builds
+        - title: Debugging in development
+          link: /docs/debugging-the-build-process/
+        - title: Debugging (builds) in production
           link: /docs/debugging-html-builds/
         - title: Debugging Replace Renderer API
           link: /docs/debugging-replace-renderer-api/
-        - title: Debugging the build process
-          link: /docs/debugging-the-build-process/
-        - title: Trace Gatsby builds
+        - title: Tracing Gatsby builds
           link: /docs/performance-tracing/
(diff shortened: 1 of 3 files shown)


www/src/data/sidebars/doc-links.yaml
```diff
@@ -178,14 +178,17 @@
...
-        - title: Debugging HTML builds
-          link: /docs/debugging-html-builds/
-        - title: Debugging Replace Renderer API
-          link: /docs/debugging-replace-renderer-api/
-        - title: Debugging the build process
+        - title: Debugging in development
           link: /docs/debugging-the-build-process/
-        - title: Trace Gatsby builds
-          link: /docs/performance-tracing/
+        - title: Debugging (builds) in production
```
pieh: This is tough to name correctly and concisely - it targets errors happening during server side rendering (so when actually saving .html files) specifically

calcsam: So it will happen in production but not development, right? If there are other classes of errors that only happen in production, we can add them to this document.

How can we best emphasize "these are errors that only happen in prod, not development"? Maybe "Debugging production builds"?

m-allanson: The first doc is a general guide on 'how to debug the build process', the second doc is specific advice on fixing 'common' types of build errors. How about these as the titles?

- How to debug your Gatsby builds
- Fixing common `gatsby build` errors

I don't love using the word 'common' to reference the frequency of errors :) but couldn't come up with anything better.

LekoArts: What's the status here? :)

calcsam:
Let's close this.

Issue #10662, opened 2018-12-26
Gatsby Version >= 2.0.24 breaks Storybook support


NathHorrigan:
## Description

When following the gatsby docs for storybook support (https://www.gatsbyjs.org/docs/visual-testing-with-storybook/) on any version >= 2.0.24 you get the following fatal error:
```
ERROR in ./node_modules/gatsby/cache-dir/commonjs/public-page-renderer-dev.js
Module not found: Error: Can't resolve './pages.json' in '.....PROJECT..../node_modules/gatsby/cache-dir/commonjs'
 @ ./node_modules/gatsby/cache-dir/commonjs/public-page-renderer-dev.js 12:36-59
 @ ./node_modules/gatsby/cache-dir/commonjs/public-page-renderer.js
 @ ./node_modules/gatsby/cache-dir/commonjs/gatsby-browser-entry.js
 @ ./src/components/header/index.js
 @ ./src/components/header/header.stories.js
 @ ./src sync .stories.js$
 @ ./.storybook/config.js
 @ multi ./node_modules/@storybook/core/dist/server/common/polyfills.js ./node_modules/@storybook/core/dist/server/preview/globals.js ./.storybook/config.js (webpack)-hot-middleware/client.js?reload=true
```

### Steps to reproduce

- Create fresh Gatsby project on version >= 2.0.24
- Follow Gatsby guide: https://www.gatsbyjs.org/docs/visual-testing-with-storybook/
- Error Occurs

### Expected result

Storybook should compile & serve

### Actual result

Fatal Crash: See Above

### Environment

```
System:
    OS: macOS 10.14
    CPU: x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 10.13.0 - ~/.nvm/versions/node/v10.13.0/bin/node
    Yarn: 1.12.3 - /usr/local/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v10.13.0/bin/npm
  Browsers:
    Safari: 12.0
  npmPackages:
    gatsby: 2.0.24 => 2.0.24
    gatsby-image: ^2.0.20 => 2.0.25
    gatsby-plugin-manifest: ^2.0.9 => 2.0.12
    gatsby-plugin-offline: ^2.0.16 => 2.0.20
    gatsby-plugin-react-helmet: ^3.0.2 => 3.0.5
[...]

Issue #10668, opened 2018-12-26
Importing Link from gatsby breaks Storybook


sidharthachatterjee:
Storybook breaks (https://github.com/gatsbyjs/gatsby/issues/10662) if a component includes `import { Link } from 'gatsby'` 

This was working fine before https://github.com/gatsbyjs/gatsby/pull/9123 because 
- main was an es6 build 
- webpack would treeshake when Link was imported

Now that main is a commonjs build and the ES6 build is at module, it breaks since
- the commonjs entry point at `gatsby-browser-entry.js` requires `public-page-renderer` 
- which tries to require `pages.json`
- which doesn't exist because since presumably a gatsby build hasn't been run yet

`public-page-renderer.js` looks like 
```
if (process.env.BUILD_STAGE === `develop`) {
  module.exports = preferDefault(require(`./public-page-renderer-dev`))
} else if (process.env.BUILD_STAGE === `build-javascript`) {
  module.exports = preferDefault(require(`./public-page-renderer-prod`))
} else {
  module.exports = () => null
}
```

I assumed that this would work since `BUILD_STAGE` is not set but it seems webpack wants to resolve both calls to `require`

sidharthachatterjee:
For now I've added a quick fix to the documentation at https://github.com/gatsbyjs/gatsby/pull/10669

stephanschubert:
I applied the fix but now I get this error:

```
WARNING in ./node_modules/gatsby/cache-dir/page-renderer.js 41:24-33
"export 'apiRunner' was not found in './api-runner-browser'
 @ ./node_modules/gatsby/cache-dir/public-page-renderer-prod.js
 @ ./node_modules/gatsby/cache-dir/public-page-renderer.js
 @ ./node_modules/gatsby/cache-dir/gatsby-browser-entry.js
```

DSchau:
@stephanschubert could you provide a reproduction? Sorry you're experiencing that!

stephanschubert:
@DSchau A full reproduction is hard (custom flavored monorepo setup here) but I could narrow the issue down:

```js
  config.module.rules[0].use[0].options.presets = [
    require.resolve("@babel/preset-react"),
    [
      require.resolve("@babel/preset-env"),
      {
        shippedProposals: true,
        useBuiltIns: "entry", // <-- was "usage"
      },
    ],
  ];
```

Switching from `usage` to `entry` resolves it but I don't quite understand why. Do you have more insight here?

[...]

PR #9123, opened 2018-10-15
fix(gatsby): add main field pointing to transpiled commonjs entry

DSchau:
Fixes #8821

This PR adds a main field corresponding to the transpiled/commonjs equivalent of gatsby cache-dir. It does this by transpiling the whole folder, and adding `main` with the cjs entry point. `module` is added, which is still used by Webpack to resolve the es2015+ source with import/export/etc.

Largely, this is to increase ease of unit testing gatsby without a transpile step, and also matches semantically what the main field _should_ produce (commonjs) whereas `module` can point to the es entry point.

To test this, I created the [following repo](https://github.com/DSchau/gatsby-cjs-test) and validated that CJS main field is resolved appropriately. Additionally, I added some logging to ensure that the es build is still used for build/develop lifecycles.

cc @travi

PR #10669, opened 2018-12-26
fix(www): update storybook docs to prefer Gatsby ES6 module in webpack config

sidharthachatterjee:
The steps in the current documentation for using Storybook with Gatsby (https://www.gatsbyjs.org/docs/visual-testing-with-storybook/) don't work because importing Link from gatsby in any component breaks. 

This is because several internals get required as a result of requiring anything from the default commonjs build (which would have been tree shaken before https://github.com/gatsbyjs/gatsby/pull/9123) and eventually require a non-existent `pages.json`. 

More details are at https://github.com/gatsbyjs/gatsby/issues/10668

This pull request updates the documentation with a quick fix (setting resolve.mainFields in the storybook webpack config override thereby _preferring_ es6 gatsby over commonjs) 

Fixes https://github.com/gatsbyjs/gatsby/issues/10662

### DIFF
--- a/docs/docs/visual-testing-with-storybook.md
+++ b/docs/docs/visual-testing-with-storybook.md
@@ -85,6 +85,9 @@ module.exports = (baseConfig, env, defaultConfig) => {
     require.resolve("@babel/plugin-proposal-class-properties"),
   ]
 
+  // Prefer Gatsby ES6 entrypoint (module) over commonjs (main) entrypoint
+  defaultConfig.resolve.mainFields = ["browser", "module", "main"]
+
   return defaultConfig
 }
 ```


docs/docs/visual-testing-with-storybook.md
```diff
@@ -85,6 +85,9 @@ module.exports = (baseConfig, env, defaultConfig) => {
     require.resolve("@babel/plugin-proposal-class-properties"),
   ]
 
+  // Prefer Gatsby ES6 entrypoint (module) over commonjs (main) entrypoint
+  defaultConfig.resolve.mainFields = ["browser", "module", "main"]
```
DSchau: I'm pretty surprised this isn't the way this already works.

[Webpack works this way](https://webpack.js.org/configuration/resolve/#resolve-mainfields)--and Storybook uses Webpack afaik, so this is _odd_.

sidharthachatterjee: @DSchau The _defaultConfig_ from storybook has `mainFields: ['browser', 'main', 'module'],`

I think it gets set at https://github.com/storybooks/storybook/blob/0ef3f40d678fb42ca74505e99cc3ad2449c3bfdf/lib/core/src/server/preview/iframe-webpack.config.js#L84

kkemple: I wonder how this will affect other modules outside of gatsby, with that said I think we should  make the change to docs still

sidharthachatterjee:
Thanks @kkemple 

We can probably remove this too once we fix the underlying issue at https://github.com/gatsbyjs/gatsby/issues/10668

PR #10674, opened 2018-12-27
Fixes authentication tutorial to run without bugs

francesmx:
## Description
- Adds react-helmet as a pre-requisite (the Gatsby plugin will not work without this, and the build fails saying 'module not found')
- Removes reference to a header component which is not provided (layout.js)
- Removes reference to a layout.css file which is not provided (layout.js)

## Related Issues
I searched for related issues but wasn't able to find any.

### DIFF
--- a/docs/docs/authentication-tutorial.md
+++ b/docs/docs/authentication-tutorial.md
@@ -8,6 +8,11 @@ Sometimes, you need to create a site with gated content, available only to authe
 
 You should have already configured your environment to be able to use the `gatsby-cli`. A good starting point is the [main tutorial](/tutorial).
 
+You'll also need to have `react-helmet` installed.
+```shell
+npm install --save react-helmet
+```
+
 ## Security notice
 
 In production, you should use a tested and robust solution to handle the authentication. [Auth0](https://www.auth0.com), [Firebase](https://firebase.google.com), and [Passport.js](http://passportjs.org) are good examples. This tutorial will only cover the authentication workflow, but you should take the security of your app as seriously as possible.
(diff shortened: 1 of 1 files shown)


KyleAMathews:
Hey — thanks for the fixes!

Not sure the react-helmet change is needed — the tutorial is based on the default starter which does have react-helmet already — what was the error you ran into? https://github.com/gatsbyjs/gatsby-starter-default/blob/370cf8ab038f2b3654385fce982e87b1fb603d3a/package.json#L19

francesmx:
Ahhh I'm sorry, looking through my terminal history, I can see that I created the project using the hello world starter rather than the default starter, which was why I ran into all of these issues. I just created a new one using the proper starter and I can see everything is there. Sorry!

francesmx:
I'm not sure why the build failed there, but in any case I'd probably revert this one. Sorry peeps! Next time I'll try and make a more useful contribution!

pieh:
@francesmx Don't worry!

I actually do think we should probably use `hello-world` starter as a base for this tutorial. We recently did some changes to default starter and now the part about adding code to existing page is outdated - the `src/components/layout.js` referenced in tutorial is different than currently used one in default starter https://github.com/gatsbyjs/gatsby-starter-default/blob/master/src/components/layout.js

KyleAMathews:
Yeah no worries! This all part of the process :-)

I agree that the hello world starter is a better base for the tutorial as it won't change.

francesmx:
Yay glad it was of some use after all! Seems a sensible change!

Issue #10620, opened 2018-12-22
"gatsby new" fails on eslint-loader


bjpcjp:
## Description

$gatsby new <appname> consistently fails to compile. 

### Steps to reproduce

$gatsby new home

### Expected result

successful compilation

### Actual result

Log from command line follows:

```
Module Error (from ./node_modules/eslint-loader/index.js):

/home/bjpcjp/projects/code/js/gatsby/home/src/components/image.js
  21:11  error  Cannot query field "childImageSharp" on type "File"  graphql/template-strings

✖ 1 problem (1 error, 0 warnings)

 @ ./src/pages/index.js 10:0-40 22:25-30
 @ ./.cache/sync-requires.js
 @ ./.cache/app.js
 @ multi (webpack)-hot-middleware/client.js?path=/__webpack_hmr&reload=true&overlay=false ./.cache/app

✖ ｢wdm｣: 
ERROR in ./src/components/image.js
Module Error (from ./node_modules/eslint-loader/index.js):

/home/bjpcjp/projects/code/js/gatsby/home/src/components/image.js
  21:11  error  Cannot query field "childImageSharp" on type "File"  graphql/template-strings

✖ 1 problem (1 error, 0 warnings)

 @ ./src/pages/index.js 10:0-40 22:25-30
 @ ./.cache/sync-requires.js
 @ ./.cache/app.js
 @ multi (webpack)-hot-middleware/client.js?path=/__webpack_hmr&reload=true&overlay=false ./.cache/app
ℹ ｢wdm｣: Failed to compile.

```
### Environment

```
  System:
    OS: Linux 4.4 Ubuntu 14.04.5 LTS, Trusty Tahr
    CPU: (8) x64 Intel(R) Core(TM) i7-4710MQ CPU @ 2.50GHz
    Shell: 4.3.11 - /bin/bash
  Binaries:
    Node: 10.6.0 - ~/.nvm/versions/node/v10.6.0/bin/node
    Yarn: 1.10.1 - /usr/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v10.6.0/bin/npm
  Browsers:
    Firefox: 59.0
  npmPackages:
    gatsby: ^2.0.53 => 2.0.75 
    gatsby-image: ^2.0.20 => 2.0.25 
    gatsby-plugin-manifest: ^2.0.9 => 2.0.12 
    gatsby-plugin-offline: ^2.0.16 => 2.0.20 
[...]

PR #10677, opened 2018-12-27
fix(gatsby-plugin-sharp): bail early if sharp isn't working

sidharthachatterjee:
We've seen several issues where builds break because `gatsby-plugin-sharp` fails for some reason. 

Typically because `sharp` didn't build correctly during install like in https://github.com/gatsbyjs/gatsby/issues/10620 most likely due to an incorrect version of python

From https://github.com/nodejs/node-gyp
> python (v2.7 recommended, v3.x.x is not supported)

This PR adds a try-catch around requiring `sharp` and bails early if it fails

### DIFF
--- a/packages/gatsby-plugin-sharp/src/gatsby-node.js
+++ b/packages/gatsby-plugin-sharp/src/gatsby-node.js
@@ -2,11 +2,15 @@ exports.onPreInit = ({ actions, reporter }, pluginOptions) => {
   try {
     require(`sharp`)
   } catch (error) {
-    // Bail early if sharp isn't working
-    // TODO: Add link to docs
+    // Bail early if sharp isn't available
     reporter.panic(
-      `
-      "sharp" doesn't seem to have been built or installed correctly
+      reporter.stripIndent`
+        The dependency "sharp" does not seem to have been built or installed correctly.
+
+        - Try to reinstall packages and look for errors during installation
+        - Consult "sharp" installation page at http://sharp.pixelplumbing.com/en/stable/install/
+        
+        If neither of the above work, please open an issue in https://github.com/gatsbyjs/gatsby/issues
       `
     )
   }
--- a/packages/gatsby-plugin-sharp/src/gatsby-node.js
+++ b/packages/gatsby-plugin-sharp/src/gatsby-node.js
@@ -1,20 +1,4 @@
 exports.onPreInit = ({ actions, reporter }, pluginOptions) => {
-  try {
-    require(`sharp`)
-  } catch (error) {
-    // Bail early if sharp isn't available
-    reporter.panic(
-      reporter.stripIndent`
-        The dependency "sharp" does not seem to have been built or installed correctly.
-
-        - Try to reinstall packages and look for errors during installation
-        - Consult "sharp" installation page at http://sharp.pixelplumbing.com/en/stable/install/
-        
-        If neither of the above work, please open an issue in https://github.com/gatsbyjs/gatsby/issues
-      `
-    )
-  }
-
   const { setBoundActionCreators, setPluginOptions } = require(`./index`)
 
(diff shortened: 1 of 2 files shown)


sidharthachatterjee:
Deleting `node_modules/sharp/build/Release/sharp.node` and running `gatsby develop` fails with an error message
screenshot 2018-12-27 20 13 33
Need to 
- [ ] add link to a docs page explaining how to resolve this on different platforms

pieh:
I think we should also provide some options here:
 - try to reinstall packages and look for error during installation
 - consult `sharp` installation page:  http://sharp.pixelplumbing.com/en/stable/install/

packages/gatsby-plugin-sharp/src/gatsby-node.js
```diff
@@ -1,6 +1,18 @@
-const { setBoundActionCreators, setPluginOptions } = require(`./index`)
+exports.onPreInit = ({ actions, reporter }, pluginOptions) => {
+  try {
+    require(`sharp`)
+  } catch (error) {
+    // Bail early if sharp isn't working
+    // TODO: Add link to docs
+    reporter.panic(
+      `
+      "sharp" doesn't seem to have been built or installed correctly
```
DSchau: Can we add some additional info here?

This doesn't seem like it'd be any _more_ useful to a user as they go through debugging why this error is triggered?

e.g. 

```js
reporter.panic(
  reporter.stripIndent`
    The dependency "sharp" does not seem to have been built or installed correctly.
    Check out this guide for more info: https://gatsby.app/sharp-installation-issues
  `
)
```

DSchau: (also note the link doesn't--and probably shouldn't--be something we maintain)

polarathene:
In my case on alpine linux, building libvips for sharp was problematic. It's only available as a package in the edge repos, and compatibility to use that from stable release repos is unreliable(some existing alpine gatsby images ran into this issue with recent builds). Instead the alpine images are better off being built on either edge base or stable with node 8 or 10 where the sharp npm package supports pulling in binaries for many platforms/distro instead of trying to build from source. If a user decides to use current latest node v11 they'll probably miss out on the benefit of the binary support on v8 or v10, could be worth pointing out(the sharp npm package repo points this out though in their docs).

Still, in some cases I've run into the problem of running `gatsby develop` and even with the `sharp.node` file in a place an error states it could not locate the file, I was able to resolve it by adding `bindings.js` package to my package.json. It's a helper for finding bindings, might be worth mentioning too.

Only other issue I've run into is the supporting packages for jpeg/webp/png, but I think this is specific to alpine again and I need to look into it. I recall some errors, specifically I think mozjpeg failed to install, it built correctly but couldn't run autoconf(or similar filename, can't recall), if it's meant to be a package alpine doesn't seem to have it. When using local images and requesting fluid with webp fragment in graphql, my source jpeg files caused errors due to not being able to use mozjpeg. When I get time I'll investigate that further(if alpine specific will contribute that to the gatsby docker repo where a gatsby-cli docker container is in a current PR).

sidharthachatterjee:
Added better error messaging for this and asserted that it works fine if `sharp.node` isn't available due to a bad install. Wasn't able to reproduce it on Ubuntu 14.04 though (everything worked fine). 

Nonetheless, this should be good to go for now.

sidharthachatterjee:
This renders the following
```
success open and validate gatsby-configs — 0.046 s
success load plugins — 1.688 s
⠁
      The dependency "sharp" does not seem to have been built or installed correctly.

      - Try to reinstall packages and look for errors during installation
      - Consult "sharp" installation page at http://sharp.pixelplumbing.com/en/stable/install/

      If neither of the above work, please open an issue in https://github.com/gatsbyjs/gatsby/issues
```

Issue #9511, opened 2018-10-29
Errors on first install (childImageSharp/gatsby-transformer-sharp)


cmkshaw:
Hi! I did a quick command-F for `childImageSharp`/ `gatsby-transformer-sharp` in existing issues but didn't see anything for first-time installers at a glance. 

## Description

I've followed the first three steps of your [setup guide](https://www.gatsbyjs.org/docs/) and am stalled with errors.

### Steps to reproduce

Ran: 
`npm install --global gatsby-cli`
`gatsby new gatsby-site https://github.com/gatsbyjs/gatsby-starter-default`
`cd gatsby-site`
`gatsby develop` 

### Expected result

Gatsby will start a hot-reloading development environment accessible at localhost:8000

### Actual result

```
~/documents/repos/gatsby-site $gatsby develop
success open and validate gatsby-configs — 0.007 s
success load plugins — 0.182 s
error Plugin gatsby-transformer-sharp returned an error


  Error: Cannot find module '../build/Release/sharp.node'
  
  - v8-compile-cache.js:159 require
    [gatsby-site]/[v8-compile-cache]/v8-compile-cache.js:159:20
 
```
... stuff
```
success onPreInit — 0.266 s
success delete html and css files from previous builds — 0.010 s
success initialize cache — 0.005 s
success copy gatsby files — 0.090 s
error Plugin gatsby-transformer-sharp returned an error


  Error: Cannot find module '../build/Release/sharp.node'
```

Multiple instances of:
```
error Plugin gatsby-transformer-sharp returned an error
```
finally,
```
ERROR in ./src/components/image.js
Module Error (from ./node_modules/eslint-loader/index.js):

/Users/carolineshaw/Documents/Repos/gatsby-site/src/components/image.js
  21:11  error  Cannot query field "childImageSharp" on type "File"  graphql/template-strings

✖ 1 problem (1 error, 0 warnings)

 @ ./src/pages/index.js 12:0-40 49:25-30
 @ ./.cache/sync-requires.js
 @ ./.cache/app.js
 @ multi ./node_modules/react-hot-loader/patch.js (webpack)-hot-middleware/client.js?path=http://localhost:8000/__webpack_hmr&reload=true&overlay=false ./.cache/app
ℹ ｢wdm｣: Failed to compile.

```
### Environment

```
 System:
    OS: macOS High Sierra 10.13.6
    CPU: x64 Intel(R) Core(TM) i5-5287U CPU @ 2.90GHz
[...]

PR #10681, opened 2018-12-27
chore(gatsby-cli): add Python to envinfo settings

sidharthachatterjee:
This adds Python to output of `gatsby info`

`gatsby-plugin-sharp` is very widely used and is included in the default starter. `sharp` uses `nodegyp` which needs `python` and according to its documentation

> v2.7 recommended, v3.x.x is not supported

We've seen several issues where building breaks due to `sharp` not working like https://github.com/gatsbyjs/gatsby/issues/10620 and https://github.com/gatsbyjs/gatsby/issues/9511

Knowing the python version in the env helps in these cases

### DIFF
--- a/packages/gatsby-cli/src/create-cli.js
+++ b/packages/gatsby-cli/src/create-cli.js
@@ -227,6 +227,7 @@ function buildLocalCommands(cli, isLocalSite) {
             System: [`OS`, `CPU`, `Shell`],
             Binaries: [`Node`, `npm`, `Yarn`],
             Browsers: [`Chrome`, `Edge`, `Firefox`, `Safari`],
+            Languages: [`Python`],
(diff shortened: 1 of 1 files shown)


DSchau:
@siddharthkp could you paste a sample output of this command and the additional info it includes?

sidharthachatterjee:
@DSchau That's not me 😛 (Hi Sid 👋)

sidharthachatterjee:
This is the updated output (Languages is the added section)
screenshot 2018-12-27 21 35 22

DSchau:
@sidharthachatterjee Github's mention UI is seriously so unintuitive 🙃 _Why_ would you not show the person who opened the PR _first_ when I type `@sid` 🙄

Thanks for correcting me!

Issue #5919, opened 2018-06-15
gatsby-source-contentful - mirroring images locally?


tylermenezes:
I've looked through the documentation and some of the code for a while, but I can't figure this out: is there a way to download images from Contentful during the build process?

It looks like there was a `cache-image.js` file added as part of #5659 but as far as I can tell it only uses that cache to generate SVGs.

For context, because ctfassets.net resizes/crops/compresses images on cache-miss, requests are sometimes two orders of magnitude slower than images we host ourselves.

KyleAMathews:
Not at the moment. That'd definitely be an interesting option to add. And yeah, we do download small images for generating svgs or generating 20x20 base64 images for the blur-up effect for gatsby-image.

tylermenezes:
@KyleAMathews If I add an option to the plugin, would you consider merging it?

KyleAMathews:
Yeah definitely. /cc @Khaledgarbaya

macandcheese:
+1 ... It would be great to mirror image assets from Contentful in a local folder during build process, and use relative paths to those vs. the Contentful CDN paths. Additionally having these images locally would mean being able to commit them to a Git repo vs. relying upon the Contentful CDN.

Khaledgarbaya:
> For context, because ctfassets.net resizes/crops/compresses images on cache-miss, requests are sometimes two orders of magnitude slower than images we host ourselves.

Can you share some insights on this?

> If I add an option to the plugin, would you consider merging it?

I think this can be a good idea to support offline builds but it should be configurable, but at the same time, it may introduce more complexity to the plugin like how to make sure that you are always getting the right images.

kennedyrose:
+1 We've built offline kiosks that are identical to our websites in the past. Which wouldn't be possible if using assets from Contentful.

polarathene:
@Khaledgarbaya I have made a PR for this. I think I had the same concern about getting the right images(as in if the asset on Contentful has been updated, how to compare with the locally stored/cached version). 

I should probably also add some logic to delete locally cached assets if their node does not exist anymore when running `gatsby-source-contentful` again, but I'm not sure how to go about that.

[...]

Issue #9233, opened 2018-10-19
[CI]: e2e path prefix test seems to intermittently fail


DSchau:
## Description

In CircleCI, the e2e path prefix test seems to fail intermittently. Let's fix it!

### Steps to reproduce

Check out a PR, e.g. https://github.com/gatsbyjs/gatsby/pull/9227

### Expected result

The e2e tests should _all_ pass unless actual regression was introduced in current PR

### Actual result

```
CypressError: Timed out retrying: expected 'blank' to equal '/blog/'
```

m-allanson:
Not stale gatsbot, not stale.

PR #10682, opened 2018-12-27
feat(gatsby-source-contentful): Support storing assets locally

### LINKED ISSUE #10587: contenful images to local folders
When building a site with data/images from contentful, is there a way to copy images to local folder and generate links to local copies?

polarathene:
Downloads and caches Contentful Assets to the local filesystem. Useful for reduced data usage or projects where you want the assets copied locally with builds for deploying without links to Contentful CDN.

Resolves #10587 #5919

### DIFF
--- a/packages/gatsby-source-contentful/README.md
+++ b/packages/gatsby-source-contentful/README.md
@@ -52,6 +52,70 @@ module.exports = {
 }
 ```
 
+### Download assets for static distribution
+
+Downloads and caches Contentful Assets to the local filesystem. Useful for reduced data usage in development or projects where you want the assets copied locally with builds for deploying without links to Contentful's CDN.
+
+Enable this feature with the `downloadLocal: true` option.
+
+```javascript
+// In your gatsby-config.js
+module.exports = {
+  plugins: [
+    {
+      resolve: `gatsby-source-contentful`,
+      options: {
+        spaceId: `your_space_id`,
+        // Learn about environment variables: https://gatsby.app/env-vars
+        accessToken: process.env.CONTENTFUL_ACCESS_TOKEN,
+        downloadLocal: true,
+      },
+    },
+  ],
+}
+```
+
+Query a `ContentfulAsset`'s `localFile` field in GraphQL to gain access to the common fields of the `gatsby-source-filesystem` `File` node. This is not a Contentful node, so usage for `gatsby-image` is different:
+
+```GraphQL
+graphql`
+  query MyQuery {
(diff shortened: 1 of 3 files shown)


polarathene:
`ci/circleci: e2e_tests_path-prefix` fails with:

```
 Production pathPrefix
    ✓ returns 200 on base route (529ms)
    navigation
      ✓ prefixes link with /blog (225ms)
      ✓ can navigate to secondary page (343ms)
      ✓ can navigate back from secondary page (303ms)
      1) can go back


  4 passing (12s)
  1 failing
```

I'm not sure what's causing that or how to fix it. Navigating my own project works fine. Any advice?

DSchau:
@polarathene don't worry about that e2e failure--it's unrelated.

I'll create an issue now so that we can track on it.

DSchau:
Whoops--I already _did_ create that issue.

See gatsbyjs/gatsby#9233, but once more, unrelated to any changes here!

polarathene:
@DSchau @sidharthachatterjee Would either of you be happy to review and merge this then if the failing e2e test is nothing to worry about? :)

I'm not sure if it will work well if an asset is modified on Contentful, it might not invalidate the cache and either not update the asset locally or create a new file copying over old data while the cache hasn't been cleared? (I have mentioned this in the README.md)

The wordpress plugin uses a modified metadata value to handle this, Contentful isn't providing one for ContentfulAsset afaik, there is `internal.contentDigest` but I'm not sure if that's useful.

wardpeet:
@polarathene just to make sure we won't have any regressions in the future could you add a few tests to test your cache invalidation code (https://github.com/gatsbyjs/gatsby/pull/10682/files#diff-caa2171e85ea277bf3075bd98f17c16eR51)?

polarathene:
@wardpeet I could try, but I've not really written much tests before. 

It's been a while since I wrote this, I vaguely remember referencing some existing implementations(although I'm not sure that they had any testing in place either). Any advice(link to a similar test could suffice).

wardpeet:
It's more or less testing your download function to make sure it's only called when options.localDownload is set and if nodes are called correctly.

I have only found one package that tests sourceNodes https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-source-drupal/src/__tests__/index.js so I understand that this won't help you much as contentful doesn't have a lot of tests either. I don't want to burden you with this so if you're not up for it, no biggy.

wardpeet:
Tested this on www.gatsbyjs.com and seems to work. Looks identical to https://www.gatsbyjs.org/docs/image-tutorial/ 👍

PR #10684, opened 2018-12-27
fix(gatsby): avoid full page refresh when navigating to non-existant page

pieh:
Currently in development there is refresh when navigating to not existing page.

This fixes regression caused by https://github.com/gatsbyjs/gatsby/pull/10507


ibrahimgorashi98:
Thanks pro
I'm so happy for this

PR #10687, opened 2018-12-27
keep dataPath as low-case

berezh:
## Description

Ther server require to keep URL as low case


KyleAMathews:
Which server requires this file to be lowercase?

DSchau:
@berezh thanks for this PR! Until you provide that additional data, I'm going to close this out.

That being said--we'd love to have you as a contributor, so if you can provide that detail we requested we can take a further look at this. Additionally, if you have some bandwidth, we have a number of [open-issues][open-issues] that could be fun to tackle.

Thanks again--and please reach out to any of us if we can help further!

PR #10690, opened 2018-12-27
docs(authentication-tutorial): use hello-world stater instead of default

pieh:
This changes starter used in authentication tutorial to hello-world.

Reasons:
 - default starter changes quite often (especially lately), so tutorial parts referencing code of the starter are getting outdated and confusing or would require often manual sync to reflect latest state of the default starter code
 - default starter has parts that are not focus point of this tutorial and generate unneeded distractions (`react-helmet`, `StaticQuery`, `siteMetadata`)

Ref comments in https://github.com/gatsbyjs/gatsby/pull/10674

---edit:
look of the site after completing tutorial:
screenshot 2018-12-27 at 21 10 23
screenshot 2018-12-27 at 21 10 57
screenshot 2018-12-27 at 21 11 09
screenshot 2018-12-27 at 21 11 17

tutorial page preview (lame in image - just to more easily verify highlighting if someone want to):
localhost_8001_docs_authentication-tutorial_

PR #10691, opened 2018-12-27
chore: standardize on lowercase/dasherized file names

DSchau:
## Description

I've seen some tweets that indicate that _sometimes_ there can be confusion as to whether to use uppercase or lowercase conventions for React component file names.

This isn't specific to Gatsby, but we can resolve some of this by documenting a consistent format, and for that I'd recommend that we consistently document lowercase with a dash separating terms (dasherized, e.g. NavBar is `nav-bar.js`). This is what we use in the starters (e.g. src/components/header.js instead of src/components/Header.js) outside of the blog starter--which has been normalized in this PR.

I've also done a simple find and replace for usage of these (e.g. src/components/Bio.js -> src/components/bio.js) so that the documentation matches the usage. There are additional uppercase components (e.g. src/components/SEO.js) but I left those as is, since the usage of them where referenced is also uppercase.

## Related Issues

Sorta related to gatsbyjs/gatsby#10671 as well as my comment [here](https://github.com/gatsbyjs/gatsby/pull/10690#discussion_r244221748)


DSchau:
Also note I removed some misc. files from the starter blog (e.g. .eslint.js, .travis.yml, etc.)

polarathene:
>  lowercase with a dash separating terms (dasherized, e.g. NavBar is nav-bar.js)

This is called [kebab-case](https://medium.com/@pddivine/string-case-styles-camel-pascal-snake-and-kebab-case-981407998841)

DSchau:
@polarathene lodash (and probably jQuery, underscore, and other libraries) popularized the kebab-case term, but it's been called a few different things (including dash-case).

https://stackoverflow.com/a/17820138

But yeah--noted. Thanks!

DSchau:
@pieh mentioned a possibly helpful plugin to automate this out.

https://github.com/selaux/eslint-plugin-filenames

Worth checking out as part of this PR

DSchau:
@ZYSzys re: uppercase, yeah, I honestly don't really care what we use as long as it's consistent. I tend to use kebab-case (dash-case) in my personal projects, but as long as we are consistent in a style I vote that.

Also note, that we don't really use Airbnb style-guide in Gatsby whatsoever, although it certainly is ubiquitous!

LekoArts:
What's the status here? 😊 Do only the merge conflicts need to be resolved?

DSchau:
@LekoArts it should be ready now. Sorry for the delay!

PR #10694, opened 2018-12-28
fix(docs): tweak s3/cloudfront doc to actually use s3 and cloudfront

jariz:
## Description

Make deployment to S3 tutorial a bit simpler & add mention of gatsby-plugin-s3 to the caching docs.
Also adds a mention on how to integrate [serverless framework](https://serverless.com) later on if you wish to do so.

## Questions/Thoughts

Maybe removing amplify entirely from the docs is a bit extreme, however, here are a couple reasons why I think the current tutorial is not a good approach for gatsby projects:

- Redirects are lost (!)
- Client only routes will 404 (!!)
- Gatsby's [recommended caching rules](https://www.gatsbyjs.org/docs/caching/#caching) are not applied.
- Amplify CLI requires a lot of unnecessary setup steps that are identical between gatsby projects.
- Amplify CLI is very hard to use with CI environments (that aren't their own 🙄):
    - Amplify's CLI does not respect environment variables(?!), making it harder to use with a CI environment that isn't amplify.
    - Amplify's 'deploy' command has a prompt that can't be disabled making it impossible to use with CI environments.
- Serverless is a lot more established than amplify is.

I think people with this PR are getting a setup process that is a lot easier and more importantly, optimises a whole lot more, which kind of fits in line with the whole idea of gatsby in general.

Having said that, if people feel strongly on keeping amplify in the docs somehow, I'm all open for suggestions.

### DIFF
--- a/docs/docs/caching.md
+++ b/docs/docs/caching.md
@@ -33,3 +33,4 @@ How you setup your caching depends on how you host your site. We encourage peopl
 The following plugins have been created:
 
 - [gatsby-plugin-netlify](/packages/gatsby-plugin-netlify/)
+- [gatsby-plugin-s3](https://github.com/jariz/gatsby-plugin-s3)
--- a/docs/docs/deploying-to-s3-cloudfront.md
+++ b/docs/docs/deploying-to-s3-cloudfront.md
@@ -2,9 +2,8 @@
 title: Deploying to S3/Cloudfront
 ---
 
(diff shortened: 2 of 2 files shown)


jariz:
@DSchau Good point.
I kinda just linked to instructions on the `gatsby-plugin-s3` repo, but I def. could expand on it a bit more.
Will look into it this weekend.

jariz:
@DSchau should be ready to go now!

Issue #10607, opened 2018-12-21
Autogenerate lock files for starters


sidharthachatterjee:
Now that starters are in this repo courtesy of @DSchau, @pieh and I were discussing how we ought to autogenerate lock files for them since manually updating dependencies is a chore. 

`package.json` for all starters have all dependencies specified with the `^` selector so it's only the lock files that needs updating.

sidharthachatterjee:
@DSchau just made a great point about how we don't supply a `yarn.lock` and `yarn` users therefore don't get deterministic installs

KyleAMathews:
For context, the only reason we include a lock file at all (we didn't for a while) is the npm has the adorable behavior that if you install dependencies with yarn but then install a package with npm, npm not seeing a package-lock.json file will assume that everything in node_modules isn't needed and then delete those and only install the one new package.

We used to get tons of issues from people complaining that gatsby didn't work after a while when going through the tutorial or trying it out.

So deterministic installs isn't the goal, the goal is helping npm avoid deleting all someone's packages.

What would also help is if we let people choose to use NPM or yarn as then "gatsby new" could choose one or the other and then we could avoid lock files altogether.

JacobBlomgren:
@sidharthachatterjee any updates on this? I was just about to file an issue about `Gatsby new`, where yarn will print 

>  package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.

which is related to this. As I see it, there are two solutions to this, either remove the lock file that is not used during `Gatsby new`, or only ship the default starter with package-lock and then do `yarn import` as mentioned above. I’d be happy to help! This has also been reported here: https://github.com/gatsbyjs/gatsby/issues/4514#issuecomment-458729956

sidharthachatterjee:
Sorry for missing this comment @JacobBlomgren 

I've tagged you in https://github.com/gatsbyjs/gatsby/issues/13210 with a potential solution in case you're still interested in taking a stab at this!

sidharthachatterjee:
We now include both `package-lock.json` and `yarn.lock` in official starters (like in https://github.com/gatsbyjs/gatsby-starter-default) and delete the unused one based on the user's selected package manager.

PR #10696, opened 2018-12-28
feat(ci): add a validation step for starters

DSchau:
## Description

This PR will (start) to add some increased validation and checks for starters. In particular, it will generate a `yarn.lock` file in each starter (and add this to the read-only clone), as well as run the `build` script to validate a successful build.

During a *pull request*, it will:

- Run `npm install`
- Run `npm run build`

This will catch issues with a dependency change causing a failure, but it _will not_ (currently) catch something like a failure introduced in the _current_ changes in the PR (e.g. a change to Gatsby like in gatsbyjs/gatsby#10593). Fixing both of these are worth exploring, I just don't want to slow down the build _too_ much.

During a merge to master, it will:

- Run `npm install`
- Run `npm run build`
- Run `yarn import` (generate a yarn.lock file based on the contents of `package-lock.json`)
- Sync to gatsbyjs/gatsby-starter-NAME

My main concern with the merge to master is that it's possible to fail and not know it fails. We do merge to master enough that it's not a huge concern and we'd catch it the next pull request, but worth noting.

## Related Issues

Begins to address #10607

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -194,5 +194,5 @@ workflows:
   version: 2
   build-test:
     jobs:
-      - ? starters:validate
+      - ? starters_validate
         <<: *ignore_master
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -194,5 +194,5 @@ workflows:
   version: 2
   build-test:
     jobs:
-      - ? starters_validate
-        <<: *ignore_master
+      - starters_validate:
+          <<: *ignore_master
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -185,7 +185,6 @@ jobs:
       - <<: *install_node_modules
       - <<: *persist_cache
       - run: yarn markdown
-      - run: sudo apt-get update && sudo apt-get install jq # jq is helpful for parsing json
       - run: git config --global user.name "GatsbyJS Bot"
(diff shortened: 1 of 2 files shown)


DSchau:
Also - I want to think more on this approach.

This doesn’t seem extremely scalable, and I want to avoid having to resort to editing shell scripts and adding additional (possibly slow) checks for every new functionality we want to validate.

It’ll work—but I’d like to see if there’s something cleaner and perhaps simpler.

DSchau:
So after thinking more on this, it may make sense to just make starters part of lerna/workspaces—then we can run whatever arbitrary tasks we want in the repo pretty cleanly.

@pieh what this PR has works, but could possibly be improved. What do you think?

pieh:
I think this looks good to test. One thing that potentially worries me is the lock generation: If I read this right - it will (re)generate lock files on master commits. There is slight problem - lerna pushes publish commits before it push packages to npm registry, so we might be "version" behind with the lock files if we install deps before publish is complete?

DSchau:
@pieh your comment there is if we swap to the lerna/workspaces approach, correct? As it works now, we're just basing the `yarn.lock` generation off of what exists in `package-lock.json` (which currently isn't auto-generated whatsoever, unfortunately).

Issue #10232, opened 2018-12-01
gatsby-plugin-sharp GraphQL fields are missing


vhfmag:
## Description

Even though I have up to date versions of `gatsby-plugin-sharp` and `gatsby-transformer-sharp` installed, I'm unable to query for `allImageSharp` or `imageSharp`. I receive the following error message: `Cannot query field "allImageSharp" on type "Query"`.
![image](https://user-images.githubusercontent.com/357835/49328590-3f78fe80-f56b-11e8-8a88-0c59e403eed7.png)

I tried disabling every non-essential Gatsby plugin and went as far back is git history as https://gitlab.com/vhfmag/vhfmag.gitlab.io/tree/31099d57637122877361486463a60e769e1ae5be (so I ruled out sri, netlify, netlify-cache, mdx, brotli and zopfli plugins), but still couldn't query those fields.

### Steps to reproduce

I couldn't reproduce it using a starter, so I will link to the [project's repo](https://gitlab.com/vhfmag/vhfmag.gitlab.io), its [`package.json`](https://gitlab.com/vhfmag/vhfmag.gitlab.io/blob/master/package.json) and its [`gatsby-config.js`](https://gitlab.com/vhfmag/vhfmag.gitlab.io/blob/master/gatsby-config.js). I'm not asking anyone to debug my site, just linking in case it helps to identify the core issue.

### Expected result

Querying `{ allImageSharp { edges { node { id } } } }` should work.

### Actual result

It fails with the following message: `Cannot query field "allImageSharp" on type "Query"`

### Environment

```
  System:
    OS: Linux 4.18 Ubuntu 18.10 (Cosmic Cuttlefish)
    CPU: (8) x64 Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
    Shell: 5.5.1 - /usr/bin/zsh
  Binaries:
    Node: 8.12.0 - ~/.nvm/versions/node/v8.12.0/bin/node
    Yarn: 1.12.3 - ~/.nvm/versions/node/v8.12.0/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v8.12.0/bin/npm
  Browsers:
    Chrome: 70.0.3538.110
    Firefox: 63.0.3
  npmGlobalPackages:
    gatsby-cli: 2.4.6
```

andrico1234:
Fixed the issue.

[...]

PR #10715, opened 2018-12-29
feat(gatsby-transformer-sharp): Add note about source plugin

LekoArts:
Fixes https://github.com/gatsbyjs/gatsby/issues/10232 as per Slack discussion


andrico1234:
Thanks for this 🙏🙏

PR #10728, opened 2018-12-30
Update sourcing from databases

calcsam:
Add in links to packages referencing how to source data from databases


LekoArts:
Dunno if Hasura would also fit in here, since it’s a database at the end, too.
https://github.com/hasura/graphql-engine/tree/master/community/boilerplates/gatsby-postgres-graphql

pieh:
@LekoArts Let's merge as-is for now. Just reading that link, Hasura seems like it need more setup, so potentially would need more content than just link to it(?)

PR #10732, opened 2018-12-30
feat(gatsby): use V8.serialize instead of JSON.stringify if available

stefanprobst:
Use `v8.serialize` instead of `json-stringify-safe` for persisting redux state for a small performance gain (from ~250ms to ~150ms in the markdown benchmark). Useful?
NOTE: Could only be merged once we switch minimum Node version to 8.0.


KyleAMathews:
Oh nifty. Didn't know about this. We could check the node version being used and use one or the other.

pieh:
or just straight check if `v8.serialize` and `v8.deserialize` exists, and fallback to current method if they don't. @stefanprobst, did you measure memory usage by any chance with your changes (`json-stringify-safe` can hog so much memory sometimes)?

stefanprobst:
Some memory usage numbers: markdown benchmark is at 73%. Also did a benchmark with a 100.000 entries Map (10 run average):

*v8*
time: 466ms
ΔheapUsed: 3 MB
Δexternal: 207 MB

*json-stringify-safe*
time: 1577ms
ΔheapUsed: 444 MB
Δexternal: 112 MB

Two questions:
* why `json-stringify-safe` and not `JSON.stringify`?
* any reason for `writeFile` not `writeFileSync` in the event callback?

DSchau:
@stefanprobst whoa! That is looking encouraging! Nice work!

pieh:
> * why `json-stringify-safe` and not `JSON.stringify`?

"Like JSON.stringify, but doesn't throw on circular references." - there were issues with those in the past, so this package was taking care of them.

> * any reason for `writeFile` not `writeFileSync` in the event callback?

We shouldn't be blocking main thread here. Are there any advantage to using sync version?

stefanprobst:
> circular references

Where are they coming from? I'm asking because we cannot recreate the references with JSON.parse? And also because easiest would be to just use `jest-serialize`.

> any advantage to using sync version?

From the docs:[ It is unsafe to use fs.write() multiple times on the same file without waiting for the callback.](https://nodejs.org/api/fs.html#fs_fs_write_fd_string_position_encoding_callback)

stefanprobst:
Has anyone had a chance to test this?

pieh:
@stefanprobst We didn't.

This will need to wait at least a week, because most of us will be offline next week - so we wouldn't be able to deal with any potential bugs it might introduce

stefanprobst:
One issue: `v8.serialize` throws when it encounters a function, e.g. [like here](https://github.com/gatsbyjs/gatsby/blob/3a52625c7b0048067e7e0f925672a95660c16529/examples/using-gatsby-image/gatsby-config.js#L23).

ivanoats:
This looks really promising. We could use it on large (18K+ pages) site that is not building because our cache is too big (can't serialize redux index saveSate) [here](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/redux/index.js#L82). 

How could we help with this?

stefanprobst:
> How could we help with this?

@ivanoats Would be awesome if you could test this with your site and report back your findings!

stefanprobst:
Not sure how to proceed with this, but another idea is to break the node store out of app state, i.e. keep using `json-stringify-safe` for everything but the `nodes` redux namespace, where we could switch serialization method. We would have to separate out `nodes` anyway once we fully make the Loki switch.

ivanoats:
We (@rsstdd and I) tested this just now. Our site is no longer crashing on build but it looks like many files are missing from the build. Some images, JSON files, and JS files. 
image

I think this is a good direction to go in, not sure how best to help with @stefanprobst 's code right now. It looks like he was looking for feedback from other contributors.

stefanprobst:
@ivanoats Thanks so much for testing this! To quickly clarify: on your site this works without issues with `gatsby develop` - but with `gatsby build` you get missing files? Is this with a cleared `.cache` folder or without?

rsstdd:
@stefanprobst We were able to build ~30k pages after removing `.cache` and `public` for both `gatsby develop` and `gatsby build`. 

Thank you for your work!

stefanprobst:
@rsstdd That's great!! Do you still see issues with a populated `.cache` folder? Because that would indicate that we are losing something somewhere in serialization=>deserialization.

DanielRuf:
The serializer is currently dxperimdntal and the stability is 1. Generally I would not recommend to use this in production or in general as this could change at any time in the future (Node 11, 12, ...).

DanielRuf:
The serializer is currently experimental and the stability is 1. Generally I would not recommend to use this in production or in general as this could change at any time in the future (Node 11, 12, ...).

DanielRuf:
The Serialization API is currently experimental and the stability is 1. Generally I would not recommend to use this in production or in general as this could change at any time in the future (Node 11, 12, ...).

https://nodejs.org/api/v8.html#v8_serialization_api

stefanprobst:
@DanielRuf Well it has been added in Node 8 and is e.g. used by `jest`, so I would not exactly call this unstable.

DanielRuf:
Well, Jest supports `node >= 6` and uses fallbacks for good reasons.

https://github.com/facebook/jest/blob/master/packages/jest-serializer/src/index.ts

Still, I would not recommend this without proper fallbacks.

stefanprobst:
`jest` considers `v8.serialize` stable enough to use it by default when it is available (Node 8+), and falls back to `JSON.stringify` when it is not. This is also what this PR does. Anyway, this is more to gather some more real world data if using `v8` would help with the oom issues some people are experiencing in `saveState`.

wardpeet:
@rsstdd thanks for testing & @stefanprobst thanks for working on this! I'll ping the @gatsbyjs/core to see if we can get this merged.

stefanprobst:
@wardpeet Cool! One open issue is how to deal with nodes that contain functions, e.g. `gatsby-plugin-feed` which allows functions in plugin options, which will be part of its `SitePlugin` node.

DanielRuf:
> to deal with nodes that contain functions

For this I would have recommended https://github.com/yahoo/serialize-javascript

KyleAMathews:
We recreate SitePlugin nodes each time so this isn't an issue. Arguably we wouldn't persist them at all.

Perhaps we should document that only simple values (ints, strings, objects, arrays) are allowed in nodes?

rsstdd:
@stefanprobst The `.cache` is populated correctly and we've had no issues over the course of the week.

stefanprobst:
@KyleAMathews 

> Perhaps we should document that only simple values (ints, strings, objects, arrays) are allowed in nodes?

:+1: Question is if this would be a breaking change, i.e. can we do this in v2? I haven't really looked systematically if there are plugins other than `gatsby-plugin-feed` where this is an issue. Do you know of others?

@DanielRuf 

> For this I would have recommended https://github.com/yahoo/serialize-javascript

I'm not familiar with this library, but it looks like it does not do deserialization at all but recommends `eval` for that?

KyleAMathews:
>👍 Question is if this would be a breaking change, i.e. can we do this in v2?

Kinda. Since these _already_ weren't being serialized, they were already broken.

packages/gatsby/src/db/node-tracking.js
```diff
@@ -15,30 +15,71 @@ const getRootNodeId = node => rootNodeMap.get(node)
...
  * @param {string} nodeId Id of node that contains data passed in first parameter
+ * @param {boolean} sanitize Wether to strip objects of unuspported and not serializable fields
+ * @param {string} [ignore] Fieldname that doesn't need to be tracked and sanitized
+ *
  */
-const addRootNodeToInlineObject = (data, nodeId) => {
-  if (_.isPlainObject(data) || _.isArray(data)) {
-    _.each(data, o => addRootNodeToInlineObject(o, nodeId))
-    rootNodeMap.set(data, nodeId)
+const addRootNodeToInlineObject = (data, nodeId, sanitize, isNode = false) => {
```
wardpeet: if possible I would refactor this function into 2 functions, one that sanitizes and another one that checks the nodeTracking, I think it will improve readability

pieh: definitely possible, but this would make it so deep object traversal need to be done 2 times instead of one :S

pieh: and most of complexity is coming here from sanitization/normalization code and not node tracking (which truly is just just:
```
    if (!isNode) {
      rootNodeMap.set(data, nodeId)
    }
```
part (other than traversal code)

pieh:
I just finished verifying `v8.serialize` changes - sites from site showcase that actually builds after fresh clone also builds with v8.serialize.
From 52 sites that I build:
- 39 were just OK in both of them
- 10 failed baseline build ( didn’t build with just cloning, installing deps and running gatsby build ) - so I skipped those.
- 1 failed … because `The engine "node" is incompatible with this module. Expected version "8.11.x". Got "10.9.0"` during deps installation - skiped that one as well
- 2 initially failed with v8.serialize, but passed baseline test correctly, but after investigation those were related to plugins breaking (one was with `gatsby-remark-relative-images` - 3rd party package, and the other one was `gatsby-mdx` that is fixed in `rc` channel - https://github.com/ChristopherBiscardi/gatsby-mdx/issues/342 ) - after fixing/adjusting deps for those, v8.serialize built as well.

So I'm feeling fairly confident in going forward with this change without doing any feature flags

packages/gatsby/src/db/node-tracking.js
```diff
@@ -15,30 +15,71 @@ const getRootNodeId = node => rootNodeMap.get(node)
...
+      }
+      returnData[key] = addRootNodeToInlineObject(o, nodeId, sanitize)
+
+      if (returnData[key] !== o) {
+        anyFieldChanged = true
+      }
+    })
+
+    if (anyFieldChanged) {
+      if (isPlainObject) {
```
wardpeet:
```suggestion
      data = omitUndefined(returnData);
```

```js
/**
 * @param {Object} data
 * @param {Object} data without undefined values
 */
function omitUndefined(data) {
  const isPlainObject = _.isPlainObject(data)
  if (isPlainObject) {
    return _.pickBy(returnData, p => p !== undefined)
  }

  return returnData.filter(p => p !== undefined)
}

File: packages/gatsby/src/db/node-tracking.js [hunk 3]
```diff
@@ -15,30 +15,71 @@ const getRootNodeId = node => rootNodeMap.get(node)
...
+    // don't need to track node itself
+    if (!isNode) {
+      rootNodeMap.set(data, nodeId)
+    }
+
+    // arrays and plain objects are supported - no need to to sanitize
+    return data
   }
+
+  if (sanitize && data !== null) {
```
wardpeet: `typeof null === 'object'` so we don't have to do a check for !== null
```suggestion
  if (sanitize && !isTypeSupported(data) {
    return undefined;
  }
```

```js
/**
 * @param {*} data
 * @return {boolean}
 */
function isTypeSupported(data) {
  const type = typeof data
  const isSupported =
    type === `number` ||
    type === `string` ||
    type === `boolean` ||
    data instanceof Date

   return isSupported;
}
```

pieh:
I forgot to mention this before. I published this under dist-tag, so if anyone want to try - install `gatsby@v8-serialize`.

pieh:
Here's also some chart showing memory usage improvements - https://lucid-galileo-3c3f54.netlify.com/

Things to check - in first graph (using JSON.stringify) there are huge memory spikes when we persist state (marked by green areas), and eventually the process crashes (around ~84s when memory usage "flatlines"). Second chart (when using those changes) memory barely spikes when persisting state

packages/gatsby/src/redux/index.js
```diff
@@ -1,64 +1,39 @@
...
-  const obj = {}
-  for (let [key, value] of map) {
-    obj[key] = value
-  }
-  return obj
-}
-
-// Read from cache the old node data.
+// Read old node data from cache.
 let initialState = {}
```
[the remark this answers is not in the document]
pieh: addressed in https://github.com/gatsbyjs/gatsby/pull/10732/commits/0641283662bbb60ba9c77041f3d9621bc4629118

packages/gatsby/src/db/node-tracking.js
pieh: addressed in https://github.com/gatsbyjs/gatsby/pull/10732/commits/1905f2ee5207bf8662fe908cdf8639d89ff9bcbf

packages/gatsby/src/db/node-tracking.js
pieh: addressed in https://github.com/gatsbyjs/gatsby/pull/10732/commits/843497eb3d19dca879935c6cdbe48c6eb2b38ad3 and https://github.com/gatsbyjs/gatsby/pull/10732/commits/95dc28fa7fa065b17535f78cd056abff04fec949

PR #10739, opened 2018-12-31
fix(gatsby): add error for missing config in theme

jbolda:
## Description
Adding another error to throw before `nearMatch` if you forget to add a `gatsby-config.js` to your theme. It avoids a cryptic error that `nearMatch` would otherwise throw.

## Related Issues
None

### DIFF
--- a/packages/gatsby/src/bootstrap/get-config-file.js
+++ b/packages/gatsby/src/bootstrap/get-config-file.js
@@ -25,6 +25,11 @@ module.exports = async function getConfigFile(
   try {
     configModule = require(configPath)
   } catch (err) {
+    if (!configModule) {
+      // If you forget the gatsby-config.js in a theme, then nearMatch throws an error as it is looking in your local files.
+      // This error is more for theme creators and the following will catch issues with local config.
+      report.panic(`It appears you may be missing the gatsby-config.js at ${rootDir}. Please create one here.`, err)
+    }
     const nearMatch = await fs.readdir(rootDir).then(files =>
       files.find(file => {
         const fileName = file.split(rootDir).pop()
--- a/packages/gatsby/src/bootstrap/get-config-file.js
+++ b/packages/gatsby/src/bootstrap/get-config-file.js
@@ -28,7 +28,10 @@ module.exports = async function getConfigFile(
     if (!configModule) {
       // If you forget the gatsby-config.js in a theme, then nearMatch throws an error as it is looking in your local files.
       // This error is more for theme creators and the following will catch issues with local config.
-      report.panic(`It appears you may be missing the gatsby-config.js at ${rootDir}. Please create one here.`, err)
+      report.panic(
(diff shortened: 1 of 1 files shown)


pieh:
This will also `panic` when main (not theme's) `gatsby-config.js` is missing? If so that's not good `gatsby-config` is optional - ie. https://github.com/gatsbyjs/gatsby-starter-hello-world doesn't have gatsby-config file

jbolda:
Oh, I didn't realize main was optional. Yes, this will `panic` on main as well.

Perhaps we need to make it optional for themes then. ( If that is the intention, then we have a bug, but this example seems to imply that it is required: [ChristopherBiscardi/gatsby-theme-minimal/gatsby-config.js](https://github.com/ChristopherBiscardi/gatsby-theme-minimal/blob/master/gatsby-config.js).) /cc @ChristopherBiscardi because they said they are in the middle of a rewrite.

pieh:
It's optional for "main site" - I don't think it makes much sense for it to be optional for themes (otherwise themes don't do anything).

ChristopherBiscardi:
If you don't have a gatsby-config in your theme's directory then you're really building a plugin because you're not using the theming gatsby-config composition and also not using Component Shadowing. This points at an interesting dynamic that we're sort of sidestepping right now in that plugins and themes are extremely similar. We're sidestepping that conversation because themes are under an experimental flag anyway, so they're "different" for now by nature of that "feature flag".

I'll put the branch up and PR it so that you can follow the rewrite of this area @jbolda so that you aren't in the dark on what I'm doing :)

jbolda:
So my initial expectation was when themes move out of experimental that they would be listed in the plugins array. As it is implemented now, I don't think that is really realistic. @ChristopherBiscardi makes the point that stuck in my mind as well. A theme really is a kind of plugin, just that it has the option to use `gatsby-config.js` as well.

For some context why I initially thought to add this error, I was making a theme and one of the "plugins" uses `gatsby-config.js` to bring in the "core" components. The next "plugin" where I ran into this error is the navigation and comprises just a few components with a StaticQuery or two. Pieces of that may be something you would be interested in using the component shadowing on, so I think it makes sense to keep it more within the context of a theme. Really there is no need for a `gatsby-config.js` though as it is just a handful of React components.

I think it is pretty clear that this specific PR should be closed, but it feels like we should head towards making themes a new second level plugin (e.g. gastby-theme). Assuming that, all of the other top level API files (node, browser, SSR) are all optional. However, also assuming that, we do need to deal with themes slightly different and then we need to be able to assume if it is actually a theme or not somehow... do we enforce the actual naming scheme and rely on that? I think I saw @DSchau head down that route in #10786. I still need to catch up on what others are doing here, but just talking out loud a bit.

Anyways, closed.

DSchau:
@jbolda to be clear--I don't _want_ to enforce that naming standard, that's just the best way I have now (until a @ChristopherBiscardi PR is merged) to _guess_ that a theme is a theme.

jbolda:
It was totally clear @DSchau 👍

PR #10740, opened 2018-12-31
panic on undefined error due to theme's import

jbolda:
## Description
If you mess up your component imports in a theme, resolveComponentPath will return undefined. Catching it and error before it errors on something else and returns a cryptic error.

## Related Issues
None

### DIFF
--- a/packages/gatsby/src/internal-plugins/webpack-theme-component-shadowing/index.js
+++ b/packages/gatsby/src/internal-plugins/webpack-theme-component-shadowing/index.js
@@ -1,4 +1,5 @@
 const path = require(`path`)
+const report = require(`gatsby-cli/lib/reporter`);
 
 module.exports = class GatsbyThemeComponentShadowingResolverPlugin {
   cache = {}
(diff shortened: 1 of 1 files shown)


DSchau:
> Catching it and error before it errors on something else and returns a cryptic error.

Is the error something re: length, e.g.

```
TypeError: Cannot read property 'length' of undefined
  
  - loader.js:426 Function.Module._resolveLookupPaths
    internal/modules/cjs/loader.js:426:15
  
  - loader.js:596 Function.Module._resolveFilename
    internal/modules/cjs/loader.js:596:20
  
  - v8-compile-cache.js:162 Function.require.resolve
    [project]/[v8-compile-cache]/v8-compile-cache.js:162:23
  
  - index.js:40 Resolver.resolver.plugin
    [project]/[gatsby]/dist/internal-plugins/webpack-theme-component-shadowing/i    ndex.js:40:45
```

if so - yeah this is nicer than that cryptic error!

DSchau:
```
Successfully published:
 - gatsby@2.0.81
```

jbolda:
@ChristopherBiscardi ah nice, that was going to be the next thing I was trying to figure out. I got as far as figuring out why it was erroring out. Do you have a branch somewhere that I can check out where you at? I was looking at importing things from the `.cache` folder and that was when I ran into this error.

PR #10762, opened 2019-01-01
docs(tutorial): tweak part seven to use graphql directly

DSchau:
## Description

After watching @kyleshevlin live code his Gatsby powered blog, it's pretty clear that the `return new Promise()` stuff tends to be a little more confusing than needed. This PR tweaks the _tutorial_ to directly return the Promise returned by the `graphql` invocation.

We can (and probably should) tweak this elsewhere, but fixing it in the tutorial is one of the highest value places we can fix for now.

cc @KyleAMathews


KyleAMathews:
Oh, actually this is a duplicate of https://github.com/gatsbyjs/gatsby/pull/10357/

Lemme close this in place of that — I'll get this in + modify the createPages docs.

PR #10765, opened 2019-01-02
Improve the error message when a non-page component exports a query

KyleAMathews:
Generally it's a problem in their gatsby-node.js so we should point
    people there.


DSchau:
```
Successfully published:
 - gatsby@2.0.82
```

Issue #10113, opened 2018-11-24
Algolia guide


janosh:
The information on how to use Algolia to search a Gatsby site is quite sparse at the moment. All I could find was the [Adding search guide](https://www.gatsbyjs.org/docs/adding-search/#use-an-api-based-search-engine) initiated in May 2018 and this [blog post](https://www.gatsbyjs.org/blog/2017-12-06-gatsby-plus-contentful-plus-netlify/#solution-algolia--gatsby) from December 2017. Neither goes into any detail and neither even mentions [`gatsby-plugin-algolia`](https://github.com/algolia/gatsby-plugin-algolia). Basically both refer you to Algolia's [Docsearch](https://community.algolia.com/docsearch) which isn't that helpful since most sites aren't documentation sites.

I implemented a custom Algolia search on several sites ([Ocean artUp](https://ocean-artup.eu), [SbS](https://studenten-bilden-schueler.de), [my own](https://janosh.io)) and contributed to `gatsby-plugin-algolia` so I have some experience on the matter by now. I would be happy to contribute a blog post, extend the Adding search guide or write a new guide. Would any of those be welcomed?

jonniebigodes:
For me it either would be perfect. i've been reading on algolia but still very very illiterate on the matter and i would like to see more on the subject and how to make it work with gatsby.

janosh:
@jonniebigodes Thanks for chiming in. Good to know that there's some interest. Still let's wait and see what the maintainers say. In the meantime, [this blog post](https://janosh.io/blog/gatsby-algolia-search) I wrote on the topic might help you out.

m-allanson:
+1 for `gatsby-plugin-algolia`! I've been using it for a recent project and agree it'd be great to see more info on using Algolia with Gatsby. 

I'm a little out of the loop at the moment, cc'ing @gatsbyjs/docs what would be the best approach here? 

Referencing https://www.gatsbyjs.org/docs/templates/ it sounds like this could be  handled as a tutorial?

Haroenv:
Feel free to ask me any questions as well. The plugin should work well up to some more advanced use cases, but isn't that optimised yet for less network usage on the indexing side, which is why  I still left it in "beta" stage for now.

m-allanson:
@janosh sorry for the delay! If you still have the time, there's now a placeholder article for an Algolia guide. See https://www.gatsbyjs.org/docs/adding-search-with-algolia/

janosh:
[...]

PR #10772, opened 2019-01-02
docs: Add stub articles for search guides

m-allanson:
## Description

Update the existing search guide to act as an overview for more specific guides.

I've also updated the [[guidelist]] handler to remove the hardcoded paths (cc @jlengstorf)

screenshot 2019-01-02 at 14 25 47

## Related Issues

Refs #10113

### DIFF
--- a/www/src/data/sidebars/doc-links.yaml
+++ b/www/src/data/sidebars/doc-links.yaml
@@ -191,6 +191,13 @@
       items:
         - title: Adding search
           link: /docs/adding-search/
+          items:
+            - title: "Adding search with Algolia*"
+              link: /docs/adding-search-with-algolia/
+            - title: "Adding search with elasticlunr*"
+              link: /docs/adding-search-with-elasticlunr/
+            - title: "Adding search with js-search*"
+              link: /docs/adding-search-with-js-search/
         - title: Adding analytics
           link: /docs/adding-analytics/
         - title: Adding authentication
--- a/www/src/data/sidebars/doc-links.yaml
+++ b/www/src/data/sidebars/doc-links.yaml
@@ -196,7 +196,7 @@
               link: /docs/adding-search-with-algolia/
             - title: "Adding search with elasticlunr*"
               link: /docs/adding-search-with-elasticlunr/
-            - title: "Adding search with js-search*"
+            - title: Adding search with js-search*
               link: /docs/adding-search-with-js-search/
         - title: Adding analytics
           link: /docs/adding-analytics/
--- a/www/src/data/sidebars/doc-links.yaml
+++ b/www/src/data/sidebars/doc-links.yaml
@@ -194,7 +194,7 @@
           items:
             - title: "Adding search with Algolia*"
               link: /docs/adding-search-with-algolia/
-            - title: "Adding search with elasticlunr*"
+            - title: Adding search with elasticlunr*
               link: /docs/adding-search-with-elasticlunr/
             - title: Adding search with js-search*
               link: /docs/adding-search-with-js-search/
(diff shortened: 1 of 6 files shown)


jonniebigodes:
@m-allanson i'm working on documentating how to search with reactivesearch. I can expand the document i'm producing for both of elasticunr and js-search also. Algolia for me it's not something i really not that confortable, so i'll refrain from expanding that stub and probably fumble fingering it.

www/src/templates/template-docs-markdown.js
```diff
@@ -15,9 +15,26 @@ import docsHierarchy from "../data/sidebars/doc-links.yaml"
 // Find the guides in the sidebar YAML.
 const guides = docsHierarchy.find(group => group.title === `Guides`).items
 
+// Search through the guides and their children to find the first item that
+// matches the given slug
+const findBySlug = (guides, slug) => {
```
DSchau: This looks better, but I wonder if we could make it less side-effecty with a `.find` and perhaps a flatten? Here's a start:

```js
const flatten = arr => arr.reduce((merged, item) => merged.concat(Array.isArray(item.items) ? flatten(item.items) : item), [])

const findBySlug = (guides, slug) => {
  return flatten(guides)
    .find(item => item.link === slug)
}
```

DSchau: Presuming we have a structure like (for guides):

```js
const guides = [
    {
      link: '/test'
    },
    {
      items: [
        {
          link: '/test2'
        }
      ]
    }
  ]
```

this seems to work just fine!

www/src/data/sidebars/doc-links.yaml
```diff
@@ -191,6 +191,13 @@
       items:
         - title: Adding search
           link: /docs/adding-search/
+          items:
+            - title: "Adding search with Algolia*"
```
LekoArts: The apostrophes `"` aren't used anywhere else, so you should probably don't use them for coherence

jlengstorf:
```suggestion
            - title: Adding search with Algolia*
```

www/src/data/sidebars/doc-links.yaml
```diff
@@ -191,6 +191,13 @@
...
+              link: /docs/adding-search-with-algolia/
+            - title: "Adding search with elasticlunr*"
```
jlengstorf:
```suggestion
            - title: Adding search with elasticlunr*
```

www/src/data/sidebars/doc-links.yaml
```diff
@@ -191,6 +191,13 @@
...
+              link: /docs/adding-search-with-elasticlunr/
+            - title: "Adding search with js-search*"
```
jlengstorf:
```suggestion
            - title: Adding search with js-search*
```

www/src/data/sidebars/doc-links.yaml
m-allanson: Thanks, I could have sworn I copied the quote style from elsewhere in the document, but clearly I was just making things up 🤔

www/src/templates/template-docs-markdown.js
m-allanson: I had a little play around with this but I don't think flatten is right here, as we want to grab the child items from the item with the matching slug. I rejigged some other bits while playing around with this though, thank you 👍

jonniebigodes:
ok, so with this merge now my pull request #11030 is basically throwing merge conflicts with what is now the master. If you don't mind, a little feedback would be appreciated on this matter

m-allanson:
Sorry I missed your comment earlier, I've fixed up the merge conflicts on your branch.

PR #10777, opened 2019-01-02
docs(gatsby): improve createPages example

KyleAMathews:

### DIFF
--- a/packages/gatsby/src/utils/api-node-docs.js
+++ b/packages/gatsby/src/utils/api-node-docs.js
@@ -12,49 +12,51 @@ exports.resolvableExtensions = true
  *
  * See also [the documentation for the action `createPage`](/docs/actions/#createPage).
  * @example
+ * const path = require(`path`)
+ *
  * exports.createPages = ({ graphql, actions }) => {
  *   const { createPage } = actions
- *   return new Promise((resolve, reject) => {
- *     const blogPostTemplate = path.resolve(`src/templates/blog-post.js`)
- *     // Query for markdown nodes to use in creating pages.
- *     resolve(
- *       graphql(
- *         `
- *       {
- *         allMarkdownRemark(limit: 1000) {
- *           edges {
- *             node {
- *               fields {
- *                 slug
- *               }
+ *   const blogPostTemplate = path.resolve(`src/templates/blog-post.js`)
+ *   // Query for markdown nodes to use in creating pages.
+ *   // You can query for whatever data you want to create pages for e.g.
+ *   // products, portfolio items, landing pages, etc.
+ *   graphql(`
+ *     {
+ *       allMarkdownRemark(limit: 1000) {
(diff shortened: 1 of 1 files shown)


packages/gatsby/src/utils/api-node-docs.js
```diff
@@ -12,49 +12,51 @@ exports.resolvableExtensions = true
...
- *                 //
- *                 // The page "path" is always available as a GraphQL
- *                 // argument.
- *               },
- *             })
- *         })
+ *     }
+ *   `).then(result => {
+ *     if (result.errors) {
+ *       reject(result.errors)
```
DSchau:
```suggestion
  *       return Promise.reject(result.errors)
```

packages/gatsby/src/utils/api-node-docs.js
```diff
@@ -12,49 +12,51 @@ exports.resolvableExtensions = true
...
- *           edges {
- *             node {
- *               fields {
- *                 slug
- *               }
+ *   const blogPostTemplate = path.resolve(`src/templates/blog-post.js`)
+ *   // Query for markdown nodes to use in creating pages.
+ *   // You can query for whatever data you want to create pages for e.g.
+ *   // products, portfolio items, landing pages, etc.
+ *   graphql(`
```
DSchau: Do we want to return here? It doesn't really matter either way, but it doesn't _hurt_

packages/gatsby/src/utils/api-node-docs.js
```diff
@@ -12,49 +12,51 @@ exports.resolvableExtensions = true
...
+ *           // arguments to the page GraphQL query.
+ *           //
+ *           // The page "path" is always available as a GraphQL
+ *           // argument.
+ *         },
  *       })
- *     )
+ *     })
+ *
+ *     return
```
DSchau: We can probably just drop this? The flow ends here anyways, particularly if we return the `graphql` call above.

KyleAMathews:
This is a similar change to @pobch's in https://github.com/gatsbyjs/gatsby/pull/10357

packages/gatsby/src/utils/api-node-docs.js
KyleAMathews: 👍

packages/gatsby/src/utils/api-node-docs.js
KyleAMathews: Oh this should be a `throw`, yikes, didn't read this very closely.

KyleAMathews:
Hmm now that I think about it, in the past I've avoided throwing because that meant you couldn't get to graphiql to try fixing the error.

I wonder if we should automatically catch graphql errors internally and show a nice clean error with a link to open the query as is in graphiql and never return graphql errors.

/cc @gatsbyjs/core

PR #10781, opened 2019-01-02
Improve createPages of blog starter

mottox2:
## Description
Remove extra promise in `starters/blog/gatsby-node.js`.

## Related Issues
The same fixes is made for these pull requests.

https://github.com/gatsbyjs/gatsby/pull/10357
https://github.com/gatsbyjs/gatsby/pull/10777

### DIFF
--- a/starters/blog/gatsby-node.js
+++ b/starters/blog/gatsby-node.js
@@ -4,55 +4,51 @@ const { createFilePath } = require(`gatsby-source-filesystem`)
 exports.createPages = ({ graphql, actions }) => {
   const { createPage } = actions
 
-  return new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


starters/blog/gatsby-node.js
```diff
@@ -4,55 +4,51 @@ const { createFilePath } = require(`gatsby-source-filesystem`)
...
-      ).then(result => {
-        if (result.errors) {
-          console.log(result.errors)
-          reject(result.errors)
         }
+      }
+    `
+  ).then(result => {
+    if (result.errors) {
+      console.log(result.errors)
```
DSchau: We probably don't need to log here since the console will (already) output the thrown error, correct?

mottox2: Thanks. The log is unnecessary so I remove it.

PR #10782, opened 2019-01-02
rename unstructured data to using gatsby without graphql

amberleyromo:
cc @jlengstorf


shannonbux:
Looks great! Wondering if changing the example site and/or blogpost would make any sense: https://github.com/gatsbyjs/gatsby/blob/5379143062440aaa9ab5813fa137e5dcf518b58d/examples/using-unstructured-data/README.md

PR #10784, opened 2019-01-02
feat(www): add plugin feedback form to all plugin pages

amberleyromo:
(not just main lib view).

related: #9678

### DIFF
--- a/www/src/templates/template-docs-local-packages.js
+++ b/www/src/templates/template-docs-local-packages.js
@@ -72,7 +72,7 @@ class DocsLocalPackagesTemplate extends React.Component {
           />
           <Unbird
             dataSetId="5c1ac24b4a828a169b6c235c"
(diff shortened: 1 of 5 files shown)


www/src/templates/template-docs-local-packages.js
```diff
@@ -69,6 +70,11 @@ class DocsLocalPackagesTemplate extends React.Component {
                 : npmPackageNotFound.lastPublisher
             }
           />
+          <Unbird
+            dataSetId="5c1ac24b4a828a169b6c235c"
```
wardpeet: do we want to have these values hardcoded or do we want to put this in .env folder and consume it?

wardpeet: it was more a suggestion, I would be cool with keeping it but I love the change!

amberleyromo: it was a good idea, and itched the back of my mind when I did this originally :)

PR #10786, opened 2019-01-03
fix(gatsby): correctly inject static query in theme components

### LINKED ISSUE #10785: [themes]: Static Query is not resolved/replaced in gatsby theme instances
## Description

When using a `StaticQuery` in a gatsby theme, the query is not replaced and you are left with `Loading (StaticQuery)`.

### Steps to reproduce

See [this repo](https://github.com/DSchau/gatsby-theme-static-query-repro) for a reproduction.

### Expected result

The StaticQuery in `src/components/header.js` should be replaced with the siteTitle.

### Actual result

The StaticQuery is replaced with `Loading (StaticQuery)`

DSchau:
## Description

This PR (correctly) replaces static query content in themes. It does this with a probably naive algorithm, where it checks the plugins array, and if a plugins path matches (gatsby-theme-), that plugin will be added to the query compiler list. It also slightly cleans up some of the code in the query compiler and adds a few tests

Notes:
  - I validated with this repo -> https://github.com/DSchau/gatsby-theme-static-query-repro 
  - It may not be reasonable to impose `gatsby-theme-*` constraint for all themes, so another solution may be preferable
  - Can't really tie into webpack (e.g. to detect a custom loader targeting node_modules/whatever-custom-component) because we use webpack-merge to persist into redux state, which gets translated to an empty object; although I would like to find some solution to non-theme packages in node_modules, as well as themes not named gatsby-theme-*
  
Feedback welcome, cc @ChristopherBiscardi @jbolda

## Related Issues

Fixes #10785

### DIFF
--- a/packages/gatsby/src/internal-plugins/query-runner/query-compiler.js
+++ b/packages/gatsby/src/internal-plugins/query-runner/query-compiler.js
@@ -64,15 +64,24 @@ const validationRules = [
 let lastRunHadErrors = null
 const overlayErrorID = `graphql-compiler`
 
+const resolveThemes = (base, plugins = []) =>
+  plugins.reduce((merged, plugin) => {
+    if (plugin.resolve.includes(`gatsby-theme-`)) {
+      merged.push(path.relative(base, plugin.resolve))
+    }
+    return merged
+  }, [])
+
 class Runner {
-  baseDir: string
+  base: string
+  additional: string[]
   schema: GraphQLSchema
   errors: string[]
   fragmentsDir: string
 
-  constructor(baseDir: string, fragmentsDir: string, schema: GraphQLSchema) {
-    this.baseDir = baseDir
-    this.fragmentsDir = fragmentsDir
+  constructor(base: string, additional: string[], schema: GraphQLSchema) {
+    this.base = base
+    this.additional = additional
     this.schema = schema
   }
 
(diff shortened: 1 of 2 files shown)


packages/gatsby/src/internal-plugins/query-runner/query-compiler.js
```diff
@@ -91,14 +100,20 @@ class Runner {
...
-    let files = glob.sync(`${this.fragmentsDir}/**/*.+(t|j)s?(x)`, {
-      nodir: true,
-    })
-    files = files.concat(
-      glob.sync(`${this.baseDir}/**/*.+(t|j)s?(x)`, { nodir: true })
-    )
+    const filesRegex = `/**/*.+(t|j)s?(x)`
+    let files = [`${this.base}/src`, `${this.base}/.cache/fragments`]
+      .concat(
+        this.additional.map(additional => path.join(this.base, additional))
```
DSchau: This presumes all the same directory (e.g. node_modules/gatsby-theme, or plugins/gatsby-theme)... not sure if/how monorepos would work, so that's a wrinkle to consider.

DSchau: Sounds like an easy fix--thanks!

DSchau:
Need to fix the Windows pathing stuff (as usual...)--addressing that now.

packages/gatsby/src/internal-plugins/query-runner/query-compiler.js
```diff
@@ -217,16 +233,13 @@ class Runner {
     return compiledNodes
   }
 }
-export { Runner }
+export { Runner, resolveThemes }
 
 export default async function compile(): Promise<Map<string, RootQuery>> {
-  const { program, schema } = store.getState()
+  // TODO: swap plugins to themes
```
DSchau: cc @ChristopherBiscardi once #10787 lands I can swap to `themes` from the redux state.

I don't really consider it a blocker for this PR though. We can get this one merged and then tweak later, removing the `gatsby-theme` test which will be 👌

ChristopherBiscardi: sgtm

ChristopherBiscardi:
Really excited to see this in. You rock @DSchau 

I can refactor the themes call in #10787

DSchau:
@ChristopherBiscardi I'll merge this first thing tomorrow AM and get a build out with this tweak!

jbolda:
Cheers @DSchau, works a treat!

PR #10787, opened 2019-01-03
feat(gatsby): extend core theming composition API to be recursive

ChristopherBiscardi:
## Description

Getting this up to reference from other PRs as ongoing work.

* extend the core theming composition API to be recursive, meaning themes can have parents and children.
* add theming to the redux store for usage in plugins like component shadowing
* [ ] update component shadowing for multiple themes

Related examples PR, showing the usage of the additions: https://github.com/ChristopherBiscardi/gatsby-theme-examples/pull/13

### DIFF
--- a/packages/gatsby/src/bootstrap/load-themes/index.js
+++ b/packages/gatsby/src/bootstrap/load-themes/index.js
@@ -1,0 +1,71 @@
+const path = require("path")
+const mergeGatsbyConfig = require(`../../utils/merge-gatsby-config`)
+const Promise = require(`bluebird`)
+const _ = require("lodash")
+const debug = require("debug")("gatsby:load-themes")
+const preferDefault = require(`../prefer-default`)
+const getConfigFile = require(`../get-config-file`)
+
+const resolveTheme = async themeSpec => {
+  const themeName = themeSpec.resolve || themeSpec
+  const themeConfig = themeSpec.options || {}
+  const themeDir = path.dirname(require.resolve(themeName))
+  const theme = await preferDefault(getConfigFile(themeDir, `gatsby-config`))
+  // if theme is a function, call it with the themeConfig
+  let themeConfigObj = theme
+  if (_.isFunction(theme)) {
+    themeConfigObj = theme(themeSpec.options || {})
+  }
+  return { themeName, themeConfigObj, themeSpec }
+}
+
+// single iteration of a recursive function that resolve parent themes
+const processTheme = ({ themeName, themeConfigObj, themeSpec }) => {
+  // gatsby themes don't have to specify a gatsby-config.js (they might only use gatsby-node, etc)
+  if (themeConfigObj && themeConfigObj.__experimentalThemes) {
+    return Promise.mapSeries(
+      themeConfigObj.__experimentalThemes,
+      async spec => {
+        const themeObj = await resolveTheme(spec)
+        return processTheme(themeObj)
+      }
+    ).then(arr => {
+      console.log(arr)
+      return arr.concat([{ themeName, themeConfigObj, themeSpec }])
+    })
+  } else {
+    return [{ themeName, themeConfigObj, themeSpec }]
+  }
+}
+
+module.exports = async config => {
+  let themes = []
+  return await Promise.mapSeries(
+    config.__experimentalThemes,
+    async themeSpec => {
+      const themeObj = await resolveTheme(themeSpec)
+      return processTheme(themeObj)
+    }
+  )
+    .then(arr => {
+      const flatThemes = _.flattenDeep(arr)
+      debug(flatThemes)
+      themes = flatThemes
+      return flatThemes
+    })
+    .mapSeries(({ themeName, themeConfigObj = {}, themeSpec }) => {
+      return {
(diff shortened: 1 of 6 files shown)


ChristopherBiscardi:
Tagging @jbolda @DSchau for visibility since your recent PRs have been theme related.

packages/gatsby/src/redux/reducers/index.js
```diff
@@ -41,4 +41,5 @@ module.exports = {
   babelrc: require(`./babelrc`),
   jsonDataPaths: require(`./json-data-paths`),
   thirdPartySchemas: require(`./thirdPartySchemas`),
+  themes: require(`./themes`),
```
DSchau: This would definitely solve (one of) the issues I was running into in #10786. If there's a JSON structure I can use that contains the known path to themes--that's all I need!

Still want to think about whether it might make sense to offer some mechanism _outside_ of themes (e.g. thinking like a custom component that uses StaticQuery or something), but that's not as crucial and could be a little fringe-y.

packages/gatsby/src/internal-plugins/webpack-theme-component-shadowing/gatsby-node.js
```diff
@@ -4,14 +4,14 @@ exports.onCreateWebpackConfig = (
   { store, stage, getConfig, rules, loaders, actions },
   pluginOptions
 ) => {
-  const { config, program } = store.getState()
+  const { program, themes } = store.getState()
 
-  if (config.__experimentalThemes) {
+  if (themes) {
```
DSchau: Themes will actually be defined here, even if it's not used right? I think we want:

```suggestion
   if (themes.themes) {
```

packages/gatsby/src/bootstrap/load-themes/index.js
```diff
@@ -0,0 +1,68 @@
...
+      }
+    ).then(arr => arr.concat([{ themeName, themeConfigObj, themeSpec }]))
+  } else {
+    return [{ themeName, themeConfigObj, themeSpec }]
+  }
+}
+
+module.exports = async config => {
+  let themes = []
+  return await Promise.mapSeries(
```
DSchau: Maybe tweak this to be consecutive await calls?

e.g. I _think_ this is the same effect:

```js
module.exports = async config => {
  const themes = await Promise.mapSeries(
    config.__experimentalThemes,
    async themeSpec => {
      const themeObj = await resolveTheme(themeSpec)
      return processTheme(themeObj)
    }
  )
    .then(arr => {
      const flatThemes = _.flattenDeep(arr)
      debug(flatThemes)
      return flatThemes
    })

  return themes
    .mapSeries(({ themeName, themeConfigObj = {}, themeSpec }) => {
      return {
        ...themeConfigObj,
        plugins: [
          ...(themeConfigObj.plugins || []),
          // theme plugin is last so it's gatsby-node, etc can override it's declared plugins, like a normal site.
          { resolve: themeName, options: themeSpec.options || {} },
        ],
      }
    })
    .reduce(mergeGatsbyConfig, {})
    .then(newConfig => ({
      config: mergeGatsbyConfig(newConfig, config),
      themes,
    }))
}
```

ChristopherBiscardi: I definitely intend to clean this part of the code up. I need to add a bunch of comments too before this gets merged.

ChristopherBiscardi:
@DSchau eep ok, thanks for the heads up. I usually rebase instead of merging master in for my personal workflow.

packages/gatsby/src/redux/reducers/index.js
jbolda: A thought coming off of #10739: I had mentioned I believed that themes shouldn't need to supply a gatsby-config.js. I am going to walk back on that a bit. What if we consider the recursive composability of configs and the component shadowing as two separate pieces. Really, the configs could be useful within plugins as well (and I think you were coming to this as well @DSchau ). The component shadowing really seems to be specifically theme based.

Presuming themes will move into the main plugins array, we need to be able to identify them somehow. What if a theme needs to declare itself as a theme within the config (effectively requiring a config for themes)? Essentially it would be opting into the component shadowing as config composability is could be used by any plugin.

Thoughts?

ChristopherBiscardi: Mostly I think this conversation (What are the differences between themes and plugins and should they be defined in the same array) is best left to the time at which we start considering "Themes 1.0". There's plenty that will happen between now and then and IMO it's too early to optimize this. I'm intentionally kicking that conversation down the road as I believe there will be *more* functionality added to themes. It's far easier to later merge them into the same array if it turns out that's the right path than it is to split them up after we've merged them (which could result in more community turmoil and a longer deprecation process).

ChristopherBiscardi:
This is *functionally* complete. 

I want to do the following before merging:

- [ ] Deal with formatting/lint/etc
- [ ] Comment code and generally make it "easier to understand" for future people

ChristopherBiscardi:
@DSchau happy to write up and PR an e2e test. Where's the best place to look for other examples I can copy?

PR #10788, opened 2019-01-03
Fix for How to Use gatsby-config.js example in README

richhorace:
## Description
Fix for How to Use in README for gatsby-transformer-csv.  

- The example csv file has headers 
- To get the example to work gatesby-config.js should be change to:
    {
      resolve: `gatsby-transformer-csv`,
      options: {
        noheader: **false**,
      }

### DIFF
--- a/packages/gatsby-transformer-csv/README.md
+++ b/packages/gatsby-transformer-csv/README.md
@@ -46,7 +46,7 @@ module.exports = {
     {
       resolve: `gatsby-transformer-csv`,
       options: {
-        noheader: true,
+        noheader: false,
       },
     },
   ];


wardpeet:
Noheader is needed to tell the csv parser to omit the first line as it's a header line. The name is a bit unfortunate as it might indicate that the file has noheader.

> noheader:Indicating csv data has no header row and first row is data row. Default is false. See header row
you can read up on all parameters on the [official repo](https://github.com/Keyang/node-csvtojson#parameters)

I'm closing, if you have issues feel free to open an issue first.

richhorace:
Sorry for the not being clear.  This PR to fix documentation in the example in the index.md.  GraphQL doesn't work. 

If noheader: true, GraphQL query fails. 

noheader-true

If noheader: false, GraphQL query works. 

noheader-false

PR #10803, opened 2019-01-03
feat(examples): GraphQL reference

LekoArts:
In preparation for my changes to the `GraphQL reference` doc I created this example. It'll be used for querying the `___graphql` URL on a codesandbox. It has all features that gatsbyjs.org also has, just on a smaller scale.


LekoArts:
Better? 😊

DSchau:
@lekoarts you better believe it! Just gotta fix the linter and we’ll be good to go!

PR #10807, opened 2019-01-03
fix(examples): GraphQL reference codesandbox issue

LekoArts:
Codesandbox apparently fails on that

image

### DIFF
--- a/examples/graphql-reference/package.json
+++ b/examples/graphql-reference/package.json
@@ -5,7 +5,7 @@
   "version": "0.1.0",
   "author": "LekoArts <hello@lekoarts.de>",
   "dependencies": {
-    "gatsby": "^2.0.76",
+    "gatsby": "^2.0.83",
     "gatsby-image": "^2.0.22",
     "gatsby-plugin-react-helmet": "^3.0.0",
     "gatsby-plugin-sharp": "^2.0.6",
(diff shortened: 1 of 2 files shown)


pieh:
> Actually... this should be fixed by [a4621c6](https://github.com/gatsbyjs/gatsby/commit/a4621c6f578a4bb5d43f4d20ee765042e1b72d46) and the PR #10220 correct?

Yeah, I hit this when I was using blog starter after deleting `.eslintrc` file there, so I added that to common eslint config

LekoArts:
Codesandbox doesn't have access to that file (I think)

pieh:
It wasn't published yet I think

LekoArts:
Ah, I see. However I think Codesandbox will only restart the dev server (instance) if something changed on the repo. So if we don't change the PATH_PREFIX we'd need to change something else to trigger a new run.

DSchau:
> However I think Codesandbox will only restart the dev server (instance) if something changed on the repo

We'd need to upgrade to the version of gatsby that contains this change--so maybe we do that in this PR?

LekoArts:
Didn't work out as expected, `.83` didn't have the fix in it.

DSchau:
@LekoArts I'm not sure the change has been published yet.

The `eslint-config.js` does _not_ contain the additional global.

pieh:
I just published .84 with the change

PR #10810, opened 2019-01-03
feat(docs): Update GraphQL reference

LekoArts:
Since the old GraphQL examples were often offline, I switched the examples from gatsbyjs.org to [graphql-reference](https://github.com/gatsbyjs/gatsby/tree/master/examples/graphql-reference). The `___graphql` / GraphiQL is now hosted on Codesandbox:
https://codesandbox.io/s/github/gatsbyjs/gatsby/tree/master/examples/graphql-reference

It auto-updates to the contents of the repository. Since the [URL stays the same](https://user-images.githubusercontent.com/16143594/50649475-420db000-0f7e-11e9-8e24-df520033f9a1.png) we can always use that instance.

You can see a preview of the changes I made here:
https://img.lekoarts.de/graphql_reference_preview.png

Furthermore I solved the TODOs and added more examples. I used GraphiQL's introspection (Docs at the right hand side) to find more `format`'s but didn't find any.

### DIFF
--- a/docs/docs/graphql-reference.md
+++ b/docs/docs/graphql-reference.md
@@ -36,13 +36,13 @@ There are several ways to reduce the number of results from a query. Here `total
 
 Skip over a number of results. In this query `skip` is used to omit the first 3 results.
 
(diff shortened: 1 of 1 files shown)


DSchau:
Thanks for doing this @LekoArts! I think this is a big improvement--plus I love using CodeSandbox!

Looks to be working perfectly 👌

shannonbux:
Wow, cool! If I understand correctly, this means we won't have to update
the iframes? I seem to remember @m-allanson needing to do that every so
often. I'm always a fan of reducing maintenance burden!! Thanks!

DSchau:
@shannonbux correct! If we want to change the query (or break the query 😱) we'll have to update--but these will automatically update when a change occurs in [graphql-reference-example](https://github.com/gatsbyjs/gatsby/tree/master/examples/graphql-reference)

Pretty slick!

shannonbux:
Awesome. This is great.

PR #10819, opened 2019-01-03
Indicate code sample is JS

jwworth:
## Description

Changes the code to indicate that it is JavaScript. This doesn't seem to make a different on Github, but changing it will help clarify the filetype on the docs webpage.


KyleAMathews:
Thanks for the fix!

PR #10822, opened 2019-01-04
Update index.md

lukebennett88:
## Description

It looks like the number of plugins has increased by quite a lot since this was written, changed the word 'dozens' to 'hundreds' to reflect this.


shannonbux:
@lukebennett88 haha great catch! Yes, it's been a while since we hit at least 100 plugins :)

Issue #10328, opened 2018-12-06
Document ENABLE_GATSBY_REFRESH_ENDPOINT


m-allanson:
## Summary

Setting an env var `ENABLE_GATSBY_REFRESH_ENDPOINT=true` while running `gatsby develop` will configure the develop server to re-source data when it receives a POST request to `/__refresh`.

This is a really useful feature when working with content while developing your site. The only documentation for this at the moment lives in PR and issue comments. This information should live in the docs of gatsbyjs.org.

### Basic example

Useful info available from:

- https://www.youtube.com/watch?v=8uvnAbMMVhU
- https://github.com/gatsbyjs/gatsby/pull/3054#issuecomment-352858732

Implementation: https://github.com/gatsbyjs/gatsby/blob/2f1b4879b8676e5c4a0d9df791db31bf8a7cb85d/packages/gatsby/src/commands/develop.js#L117-L128

Possibly in other issues / PRs in this repo.

### Motivation

This feature is too useful to remain undocumented 😄

RichardBronosky:
This ticket and https://github.com/gatsbyjs/gatsby/pull/10366/commits and https://github.com/gatsbyjs/gatsby/projects/2 are still the top Google hits. Can we please add links to the real documentation to this ticket and any others discussing it? I still can't find it.

drake-smith:
I understand how this refresh token is working and have tried it out on my local machine, and using localhost. However, I wonder how this can be implemented into a preview system for a non-developer, like a client that needs to preview content before publishing. Especially on Contentful, to set up the POST webhook, I need a non-localhost URL. I could use ngrok, but again, that only works in a limited scope, and wouldn't be a permanent and publicly-accessible URL for others to use.

RichardBronosky:
@drake-smith  You can pass `--host=0.0.0.0` to develop to get it to listen on non-local interfaces.

Charlesdoiron:
Hi @drake-smith i really need the same online preview for my gatsby contentfull website.
Do you find something ?

thanks a lot !

PR #10829, opened 2019-01-04
docs(env-variables): Add cURL example for refresh endpoint

riddla:
## Description

I added a cURL example for the refresh endpoint e.g. http://localhost:8000/__refresh.

## Related Issues

Related to #10328.


riddla:
Thanks for the addition. Makes totally sense.

riddla:
@DSchau, I was wondering if a small helper component like the following would/could make sense, also for others:

``` javascript
import React from 'react'

let style = {
  // ...
}

const RefreshContent = ({ children }) => (
  <button onClick={refresh} style={style}>
    Refresh content
  </button>
)

function refresh() {
  console.log('refreshing content')
  var oReq = new XMLHttpRequest()
  oReq.open('POST', 'http://localhost:8000/__refresh')
  oReq.send()
}

export default RefreshContent
```

As I'm new to the Gatsby ecosystem I don't know if those kind of (development) helper components  already exist, where I could put them etc.
Maybe you have some insights?

PR #10835, opened 2019-01-04
fix(gatsby): test plugin name to handle symlinks, rather than path

DSchau:
## Description

This simple PR augments the behavior to test the plugin's name, rather
than the plugin's resolved path, in the case that it could be
symlinked/linked. This seems to match the behavior we're expecting, and
_again_ will be cleaned up when #10787 lands.

## Related Issues

Related to #10787, #10786

### DIFF
--- a/packages/gatsby/src/internal-plugins/query-runner/query-compiler.js
+++ b/packages/gatsby/src/internal-plugins/query-runner/query-compiler.js
@@ -66,7 +66,7 @@ const overlayErrorID = `graphql-compiler`
 
 const resolveThemes = (plugins = []) =>
   plugins.reduce((merged, plugin) => {
-    if (plugin.resolve.includes(`gatsby-theme-`)) {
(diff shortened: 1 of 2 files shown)


packages/gatsby/src/internal-plugins/query-runner/query-compiler.js
```diff
@@ -66,7 +66,7 @@ const overlayErrorID = `graphql-compiler`
 
 const resolveThemes = (plugins = []) =>
   plugins.reduce((merged, plugin) => {
-    if (plugin.resolve.includes(`gatsby-theme-`)) {
+    if (plugin.name.includes(`gatsby-theme-`)) {
```
DSchau: plugin.resolve is the symlinked path when running `yarn link`. This change will work with symlinked modules and "regular" (e.g. modules in node_modules) alike.

PR #10838, opened 2019-01-04
fix(ci): do not run (unnecessary) install step for starters validation

DSchau:
## Description

Will slightly speed up the install/setup step for validating starters

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -172,9 +172,6 @@ jobs:
     steps:
       - checkout
       - run: ./scripts/assert-changed-files.sh "starters/*|.circleci/*"
-      - <<: *restore_cache
-      - <<: *install_node_modules
-      - <<: *persist_cache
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -188,7 +185,7 @@ jobs:
       - run: sudo apt-get update && sudo apt-get install jq # jq is helpful for parsing json
       - run: git config --global user.name "GatsbyJS Bot"
       - run: git config --global user.email "admin@gatsbyjs.com"
-      - run: sh ./scripts/clone-folder.sh starters
+      - run: sh ./scripts/clone-and-validate.sh starters true
```
DSchau: Swapped this logic (which is cleaner, I think).

pieh: 👍 didn't want to nitpick that in original PR :)

PR #10845, opened 2019-01-05
fix: allow plugins to use gatsby-browser.js to load global styles, etc.

KyleAMathews:


KyleAMathews:
Does anyone remember why we made this distinction?

I can't imagine someone adding something to a plugin's gatsby-browser.js without intending it to be loaded. It's frequently useful for the default gatsby-browser.js and there's certainly use cases for plugins as well. In any case, it's surprising to me there's a difference (and I was the one that added it).

e2e-tests/production-runtime/cypress/integration/global-styles.js
```diff
@@ -0,0 +1,19 @@
+const rgbBlack = `rgb(0, 0, 0, 0.8)` // css jQuery method returns rgb
```
DSchau: One note here: this is the value in _my_ browser, but I wonder if there's a better way to test this? Basically I want to assert that the text is colored, and it's not colored the _default_.

pieh: Can we test exactly for colors defined in .css files?

DSchau: Yeah - except the browser tweaks them, ha. So rgb(255, 0, 0) just adds an alpha channel I believe. Let me test that!

pieh: alternatively we could use other css properties - like `z-index` which has exact integer values (no different formats, no `px` etc)?

Issue #6933, opened 2018-08-01
gatsby-contentful-source queries for all elements of a given type return duplicates


tmergulhao:
Running on: Netlify
Using: Node v8.11.3
Gatsby: 1.9.277

In some instances or page queries that ask for all entries of a given Contentful entry type I get duplicate entries. Such a query should have a return:

```graphql
allContentfulLearnIssue(limit: 5, sort: { fields: [releaseDate], order: DESC }) {
  edges{
    issue: node {
      number
    }
  }
}
```
```javascript
allContentfulLearnIssue: {
  edges: [
    { issue: { number: 2 } },
    { issue: { number: 3 } },
  ]
}
```

Instead I get:

```javascript
allContentfulLearnIssue: {
  edges: [
    { issue: { number: 3 } },
    { issue: { number: 2 } },
    { issue: { number: 3 } },
    { issue: { number: 3 } },
    { issue: { number: 2 } }
  ]
}
```

The funniest thing about this issue is that it is not reproducible using the Contentful GraphiQL interface, but it is using the Gatsby GraphiQL interface.

How may I help identify what is the potential source of the issue. Do you guys know what is the source file in which I could backtrack this issue?

  System:
    OS: macOS 10.14
    CPU: x64 Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 8.11.3 - /usr/local/bin/node
    Yarn: 1.7.0 - /usr/local/bin/yarn
    npm: 6.2.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 67.0.3396.99
    Safari: 12.0
  npmPackages:
    gatsby: ^1.9.166 => 1.9.277
    gatsby-link: ^1.6.46 => 1.6.46
    gatsby-plugin-postcss-sass: ^1.0.22 => 1.0.22
    gatsby-plugin-react-helmet: ^2.0.11 => 2.0.11
    gatsby-source-contentful: ^1.3.54 => 1.3.54
  npmGlobalPackages:
    gatsby-cli: 1.1.58
    gatsby: 1.9.266

tmergulhao:
When Contentful pulls the data, all localizations will be pulled. Therefore, if you have a localization active, it will duplicate the entries.

Narrow the search by filtering the query with `(filter : { node_locale: { eq: "en-US" }})`

kakadiadarpan:
[...]

PR #10851, opened 2019-01-05
docs(gatsby-source-contentful): add note about node_locale filtering

tunap:
## Description

Adds a caveat to gatsby-source-contentful README.md.

## Related Issues

Addresses #6933

### DIFF
--- a/packages/gatsby-source-contentful/README.md
+++ b/packages/gatsby-source-contentful/README.md
@@ -177,7 +177,7 @@ When Contentful pulls the data, all localizations will be pulled. Therefore, if 
 
 ```graphql
 {
-  allContentfulCaseStudy(filter : { node_locale: { eq: "en-US" }}) {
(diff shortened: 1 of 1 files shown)


pieh:
Thanks for jumping to help document this @tunap!

I think we should create mini-section after `#### A note about LongText fields` - it should be then easier to spot this - something like:

````markdown
#### Duplicated entries

When Contentful pulls the data, all localizations will be pulled. Therefore, if you have a localization active, it will duplicate the entries. Narrow the search by filtering the query with `node_locale` filter:

```graphql
{
  allContentfulCaseStudy(filter : { node_locale: { eq: "en-US" }}) {
    edges {
      node {
        id
        slug
        title
        subtitle
        body {
          body
        }
      }
    }
  }
}
```

What do you think?

tunap:
Thanks for the suggestion! I have made the requested changes.

PR #10854, opened 2019-01-05
Suggest developer uses Incognito Mode for auditing

jwworth:
## Description

As I followed this tutorial, I found that my Chrome plugins, (ad blockers, etc.) interfered with the audit. There's no way to know which plugins a user may have installed, so I think it's best to audit the Gatsby app from a clean slate.


pieh:
This sounds reasonable - I think it would be good idea to explain why we recommend Incognito mode (that some browser extensions might affect audit results)

jwworth:
Great idea. I can make that adjustment!

jwworth:
Thanks for the feedback, @pieh; I've added some context.

Issue #10800, opened 2019-01-03
Add missing publishedAt entries to blog posts


m-allanson:
## Description

Some blog posts on gatsbyjs.org have a `canonicalLink`, but no `publishedAt` key in their frontmatter. This leads to the blog post header saying _(originally published at )_, instead of _(originally published at example.com)_.

A quick fix would be to update the blog posts that have a `canonicalLink` but no `publishedAt` text to also include `publishedAt`.

A better fix would be to update [`template-blog-post.js`](https://github.com/gatsbyjs/gatsby/blob/2f4ee947a9c4fcc9491819c62ce3ece1caa56c45/www/src/templates/template-blog-post.js#L210-L220) to automatically display a `publishedAt` value if one isn't supplied. As an example, a `canonicalLink` like `https://www.contentful.com/blog/2018/02/28/contentful-gatsby-video-tutorials/` should be displayed as `contentful.com`.

### Steps to reproduce

[Visit this example post](https://www.gatsbyjs.org/blog/2018-11-03-building-an-accessible-lightbox/)
Check the author credit box at the top of the post:

![screenshot 2019-01-03 at 12 56 56](https://user-images.githubusercontent.com/381801/50638861-217e2f80-0f57-11e9-8f41-6e6a31166eb2.png)

[Here's the markdown file for the above url](https://github.com/gatsbyjs/gatsby/blob/2f4ee947a9c4fcc9491819c62ce3ece1caa56c45/docs/blog/2018-11-03-building-an-accessible-lightbox/index.md). Note that there are other posts where this happens too.

### Expected result

It should say something like _(originally published at 416serg.me)_

### Actual result

It says _(originally published at )_

### Environment

gatsbyjs.org

Nilomiranda:
Wouldn't be an easier fix if we check if `publishedAt` returns something, and if not, omit the "originally published at" ?

ivorpad:
I can give this a try and it should be fairly easy to pull off. What about subdomains?:

```
* subdomain.domain.com
* www.domain.com
* ww2.sub.domain.com
```

Should `publishedAt` always be in the form of `example.com`?

m-allanson:
> Wouldn't be an easier fix if we check if publishedAt returns something, and if not, omit the "originally published at" ?

This is what happens if there's no `canonicalLink`. But if there's a `canonicalLink` without a `publishedAt` we can assume that the author wanted to display a link.

> What about subdomains?

[...]

PR #10855, opened 2019-01-05
fix(www): fallback to hostname of canonicalLink if publishedAt is not specified

ivorpad:
Fixes https://github.com/gatsbyjs/gatsby/issues/10800

### DIFF
--- a/www/gatsby-node.js
+++ b/www/gatsby-node.js
@@ -462,7 +462,7 @@ exports.onCreateNode = ({ node, actions, getNode, reporter }) => {
         slug = `/${parsedFilePath.dir}/`
       }
 
-      // Set released status for blog posts.
+      // Set released status and `published at` for blog posts.
       if (_.includes(parsedFilePath.dir, `blog`)) {
         let released = false
         const date = _.get(node, `frontmatter.date`)
@@ -470,6 +470,17 @@ exports.onCreateNode = ({ node, actions, getNode, reporter }) => {
           released = moment().isSameOrAfter(moment.utc(date))
         }
         createNodeField({ node, name: `released`, value: released })
+
+        const canonicalLink = _.get(node, `frontmatter.canonicalLink`)
+        let publishedAt = _.get(node, `frontmatter.publishedAt`)
+
+        createNodeField({
+          node,
+          name: `publishedAt`,
+          value: canonicalLink
+            ? publishedAt || url.parse(canonicalLink).hostname
+            : null,
+        })
       }
     }
     // Add slugs for package READMEs.
--- a/www/gatsby-node.js
+++ b/www/gatsby-node.js
@@ -472,7 +472,7 @@ exports.onCreateNode = ({ node, actions, getNode, reporter }) => {
         createNodeField({ node, name: `released`, value: released })
 
         const canonicalLink = _.get(node, `frontmatter.canonicalLink`)
-        let publishedAt = _.get(node, `frontmatter.publishedAt`)
+        const publishedAt = _.get(node, `frontmatter.publishedAt`)
 
         createNodeField({
           node,
--- a/www/src/templates/template-blog-post.js
+++ b/www/src/templates/template-blog-post.js
@@ -213,9 +213,7 @@ class BlogPostTemplate extends React.Component {
                       (originally published at
                       {` `}
                       <a href={post.frontmatter.canonicalLink}>
-                        {!post.frontmatter.publishedAt
-                          ? new URL(post.frontmatter.canonicalLink).hostname
-                          : post.frontmatter.publishedAt}
+                        {this.props.data.markdownRemark.fields.publishedAt}
                       </a>
                       )
                     </span>
@@ -346,6 +344,7 @@ export const pageQuery = graphql`
       timeToRead
       fields {
         slug
+        publishedAt
       }
       frontmatter {
         title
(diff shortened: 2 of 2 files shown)


ivorpad:
@m-allanson suggested [template-blog-post.js](https://github.com/gatsbyjs/gatsby/blob/2f4ee947a9c4fcc9491819c62ce3ece1caa56c45/www/src/templates/template-blog-post.js#L210-L220) but I'm happy to revisit it.

ivorpad:
@jgierer12 @m-allanson I have a solution set in place to do it at build time, but I'm not quite sure about it as I have to pass `publishedAt` via `context` however this will result in deriving `publishedAt` from `canonicalLink` and remove the need to query it.

jgierer12:
Sorry, I should have been more clear about this. What I had in mind is to use the [`onCreateNode`](https://www.gatsbyjs.org/docs/node-apis/#onCreateNode) API and its `createNodeField` action to create a new `publishedAt` field with the correct value. Then you can change the query in the template to get `publishedAt` from `fields` rather than `frontmatter`.

---

@m-allanson what are your thoughts on this? Should we keep this as is (i.e. set the fallback `publishedAt` at runtime) or do it at build time?

www/src/templates/template-blog-post.js
```diff
@@ -213,7 +213,9 @@ class BlogPostTemplate extends React.Component {
                       (originally published at
                       {` `}
                       <a href={post.frontmatter.canonicalLink}>
-                        {post.frontmatter.publishedAt}
+                        {!post.frontmatter.publishedAt
+                          ? new URL(post.frontmatter.canonicalLink).hostname
```
pieh: This is not supported by, IE11 without polyfill. I like the suggestion from @jgierer12 about doing the processing in `gatsby-node`, as then we will workaround this annoyance by using that in Node env and not in browser ;)

ivorpad: @pieh Yup and `gatsby-node.js` imports the `url` module already. Alright, I'll start working on it.

ivorpad:
@jgierer12 and @pieh,

Let me know if that's sufficient.

docs/blog/author.yaml
```diff
@@ -203,4 +203,4 @@
 - id: Ondrej Polesny
   bio: Full-stack developer, speaker and trainer with passion for anything new and intriguing. Opensource contributor. Dev evangelist @ Kentico Cloud.
   avatar: avatars/ondrej-polesny.jpg
-  twitter: "@ondrabus"
\ No newline at end of file
+  twitter: "@ondrabus"
```
jgierer12: Please also remove this change

ivorpad: Will do.

www/gatsby-node.js
```diff
@@ -462,14 +462,25 @@ exports.onCreateNode = ({ node, actions, getNode, reporter }) => {
...
         const date = _.get(node, `frontmatter.date`)
         if (date) {
           released = moment().isSameOrAfter(moment.utc(date))
         }
         createNodeField({ node, name: `released`, value: released })
+
+        const canonicalLink = _.get(node, `frontmatter.canonicalLink`)
+        const publishedAt = _.get(node, `frontmatter.publishedAt`)
+
+        if (canonicalLink && !publishedAt) {
```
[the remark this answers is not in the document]
ivorpad: Why remove the `if` statement? I need to check if `canonicalLink` is truthy. 

What about?

```javascript
canonicalLink &&
  createNodeField({
    node,
    name: `publishedAt`,
    value: publishedAt || url.parse(canonicalLink).hostname,
  })
```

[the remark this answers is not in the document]
jgierer12: I'm pretty sure Gatsby will complain when querying for a non-existant field so need to make sure we always create a field when it will be queried.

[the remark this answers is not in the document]
ivorpad: Thanks for your suggestions, this what I came up with: 

```javascript
const canonicalLink = _.get(node, `frontmatter.canonicalLink`)
let publishedAt = _.get(node, `frontmatter.publishedAt`)

createNodeField({
  node,
  name: `publishedAt`,
  value: canonicalLink
    ? publishedAt || url.parse(canonicalLink).hostname 
    : null,
})
```

ivorpad:
Thanks for your suggestions Jonas, gotta love these feedbacks. Pushed the changes, let me know if that's enough.

www/gatsby-node.js
```diff
@@ -462,14 +462,25 @@ exports.onCreateNode = ({ node, actions, getNode, reporter }) => {
...
       if (_.includes(parsedFilePath.dir, `blog`)) {
         let released = false
         const date = _.get(node, `frontmatter.date`)
         if (date) {
           released = moment().isSameOrAfter(moment.utc(date))
         }
         createNodeField({ node, name: `released`, value: released })
+
+        const canonicalLink = _.get(node, `frontmatter.canonicalLink`)
+        let publishedAt = _.get(node, `frontmatter.publishedAt`)
```
jgierer12: This can be a `const`. Other than that this looks good to me :+1:
Thanks for implementing the feedback so quickly!

@pieh do you have anything else?

ivorpad: Oops, totally missed that. Thanks.

PR #4138, opened 2018-02-20
Add starter repo for i18n from Contentful CMS.

mccrodp:
It took me quite a while to get this working, so much so that I think it's useful for others. I focused on creating a starter repo on it before starting my own work. So, just making the docs update below.

---

* [gatsby-starter-contentful-i18n](https://github.com/mccrodp/gatsby-starter-contentful-i18n) [(demo)](https://gatsby-starter-contentful-i18n.netlify.com/)

  Features:
  * Localization (Multilanguage)
  * Dynamic content from Contentful CMS
  * Integrates [i18n plugin starter](https://github.com/angeloocana/gatsby-starter-default-i18n) and [using-contentful](https://github.com/gatsbyjs/gatsby/tree/master/examples/using-contentful) repos

---

It basically integrates the community `i18n` plugin with `using-contentful` repo. i.e. - uses the `en-US` and `de` content from Using Contenful example and displays either of the content depending on lang selected using the lang selector.

I also have a i18n with transformer remark plugin, but that's perhaps better in the i18n plugin repo itself rather than here. Not sure though, if you want here, I can clean it up w.r.t naming, code comments, repo name, live demo on netlify, etc. https://github.com/mccrodp/gatsby-transformer-remark-i18n

PR #10858, opened 2019-01-06
fix(starter-showcase): added description field to mccrodp's starter repo

mccrodp:
## Description

Just a simple description field update on [my starter repo](https://github.com/mccrodp/gatsby-starter-contentful-i18n) added previously.

## Related Issues

Originally added for Gatsby v1 here: https://github.com/gatsbyjs/gatsby/pull/4138, but I just upgraded to v2 and checked the starters.md. Thanks for adding that! 🙇

### DIFF
--- a/docs/starters.yml
+++ b/docs/starters.yml
@@ -248,7 +248,7 @@
     - Responsive Design, optimized for Mobile devices
 - url: https://gatsby-starter-contentful-i18n.netlify.com/
   repo: https://github.com/mccrodp/gatsby-starter-contentful-i18n
-  description: n/a
+  description: Contentful starter repo with added i18n support and language switcher
   tags:
     - i18n
     - Contentful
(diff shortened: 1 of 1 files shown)


mccrodp:
Actually, it seems you pull in the desc from the repo, that's better, **this can be closed**, but might be good to add that as a comment the `starters.md` code snippet in the docs? https://www.gatsbyjs.org/docs/submit-to-starter-library/#steps

pieh:
I didn't even know we read package.json description, but it does seems a bit inconsistent:
 - on listing view we show package.json description ( https://www.gatsbyjs.org/starters?s=gatsby-starter-contentful-i18n&v=2 )
 - but on detail view we show description from .yml file - "N/A" in this case ( https://www.gatsbyjs.org/starters/mccrodp/gatsby-starter-contentful-i18n/ )

Might be worth to look into consolidating those? Not sure if it was intentional or not.

/cc @amberleyromo

mccrodp:
Nice find, @pieh. It also seems that the description pulled and displayed for my starter is not the current one. I updated it 4 days ago, so not sure how those are cached either.

pieh:
I'm not sure either about caching here - I lack lot of context on how starter showcase works, but let's merge PR in meantime, so your starter have proper description ;)

PR #10859, opened 2019-01-06
updated gatsby-plugin-typography to accept user defined absolute paths

Ronva:
## Description

This is an update to allow gatsby-plugin-typography to accept an absolute path, addressing an issue by @ChristopherBiscardi. It is needed for the typography util to work with Gatsby themes.

## Related Issues

Addresses [this issue](https://github.com/ChristopherBiscardi/gatsby-theme-examples/pull/5#issue-231077246)

### DIFF
--- a/packages/gatsby-plugin-typography/src/gatsby-node.js
+++ b/packages/gatsby-plugin-typography/src/gatsby-node.js
@@ -9,11 +9,12 @@ exports.onPreBootstrap = ({ store }, pluginOptions) => {
 
   let module
   if (pluginOptions.pathToConfigModule) {
-    module = `module.exports = require("${path.join(
-      program.directory,
-      pluginOptions.pathToConfigModule
-    )}")`
-    if (os.platform() == `win32`) {
+    module = `module.exports = require("${
+      path.isAbsolute(pluginOptions.pathToConfigModule)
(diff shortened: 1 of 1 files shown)


packages/gatsby-plugin-typography/src/gatsby-node.js
```diff
@@ -22,7 +23,7 @@ const typography = new Typography()
 module.exports = typography`
   }
 
-  const dir = `${__dirname}/.cache`
+  const dir = `${program.directory}/.cache`
```
KyleAMathews: This actually is what we want. We want to write the file into the same directory as where the gatsby-ssr.js file is so it can require it.

KyleAMathews: I'll push a commit to revert this.

KyleAMathews: The problem is ssr files don't know where the root of the site is at.

Ronva: Christopher was trying to help me figure out why it was failing and that line was something he thought might be causing it. Thanks for taking a look at this!

KyleAMathews:
Thanks! Published in gatsby-plugin-typography@2.2.5

PR #10868, opened 2019-01-06
feat(gatsby-cli): Initialize the newly cloned repository as git

caulagi:
## Description

Initialize the newly cloned repository as a git repo. This is useful because a lot of the tooling (shell prompts, grep (via ripgrep), etc) honor `.gitignore` rules, when the directory has `.git`. I add the following two lines at the end `gatsby new` output - 

```bash
info Initialising git in test
Initialized empty Git repository in /private/tmp/test/.git/
```


sidharthachatterjee:
Thank you so much for this @caulagi 

Have you considered creating the initial commit as well?

sidharthachatterjee:
@pieh also makes a great point about creating a `.gitignore` file if it doesn't exist (if a starter didn't include it by mistake) and ensuring it contains .cache, public and node_modules

Would you like to add these to pull request?

caulagi:
Thanks. I added the functionality to create a minimal `.gitignore` if it is missing.

Creating an initial commit should have a switch perhaps, so we can disable it? In that case, it would be a separate PR.

sidharthachatterjee:
I think we should make an initial commit by default. The Angular CLI,  `ember-cli` and `create-react-app` do this. Do we want to add a flag to disable this, @pieh? I personally think we can add a flag later if enough people want to be able to opt out. 

One more thing (sorry I should've mentioned earlier!): we want to check if the folder is already a mercurial or git repo before calling `git init` 

Presence of .git and .hg should be enough, I think

pieh:
> One more thing (sorry I should've mentioned earlier!): we want to check if the folder is already a mercurial or git repo before calling `git init`
> 
> Presence of .git and .hg should be enough, I think

Just the way we scaffold starters is by cloning starter git repository and purging `.git` directory (to get rid of git history associated with starter development), so `.git` won't be there and `.hg` shouldn't be there? unless someone commits it to git repo?

sidharthachatterjee:
> so .git won't be there and .hg shouldn't be there? unless someone commits it to git repo?

Oh yeah! I was thinking of sites created within a monorepo but that is trickier to test for

caulagi:
I added an initial git commit. The last few lines now look like this -

```bash
info Initialising git in g6
Initialized empty Git repository in /private/tmp/nn/g6/.git/
info Creating minimal .gitignore in g6
info Create initial git commit in g6
```

sidharthachatterjee:
LGTM

PR #10887, opened 2019-01-07
fix(gatsby-source-wordpress): use correct glob pattern paths for routes

### LINKED ISSUE #10871: [gatsby-source-wordpress] Error in includedRoutes documentation ?
Hi,

I try to use `includedRoutes` option of `gatsby-source-wordpress` plugin but could not make it work following the documentation here : https://www.gatsbyjs.org/packages/gatsby-source-wordpress/.

I added these lines in my `gatsby-config.json` file  :

```
includedRoutes: [
    "/*/*/categories",
    "/*/*/posts",
    "/*/*/pages",
    "/*/*/tags",
    "/*/*/taxonomies",
    "/*/*/users",
],
```

But when I run `gastby develop` and enable verbose output I have such unexpected messages :

```
Route discovered : /wp/v2/posts
Excluded route: not whitelisted
```

I debugged the plugin and found that the following method in `fetch.js` file : 

```
function checkRouteList(routePath, routeList) {
  return routeList.some(route => minimatch(routePath, route));
}
```

As you can see that this method uses the route path including base url (for example `https://mywebsite.com/wp-json/v2/posts`) and this doesn't match `/*/*/posts`.

I had to change `gatsby-config.js` this way to make it work :

```
includedRoutes: [
    "**/*/*/categories",
    "**/*/*/posts",
    "**/*/*/pages",
    "**/*/*/tags",
    "**/*/*/taxonomies",
    "**/*/*/users",
],
```

Is this an error in the documentation ? Or is there something I didn't understand ? (I am surprised I found nothing about this problem)

gurpreet-hanjra:
As per observations, glob patterns are fixed.

## Related Issues

closes #10871

### DIFF
--- a/packages/gatsby-source-wordpress/README.md
+++ b/packages/gatsby-source-wordpress/README.md
@@ -126,7 +126,7 @@ module.exports = {
           "**/*/*/users",
         ],
         // Blacklisted routes using glob patterns
-        excludedRoutes: ["/*/*/posts/1456"],
+        excludedRoutes: ["**/*/*/posts/1456"],
         // use a custom normalizer which is applied after the built-in ones.
         normalizer: function({ entities }) {
           return entities
--- a/packages/gatsby-source-wordpress/README.md
+++ b/packages/gatsby-source-wordpress/README.md
@@ -117,13 +117,13 @@ module.exports = {
         // all routes that begin with `yoast` from fetch.
         // Whitelisted routes using glob patterns
         includedRoutes: [
-          "/*/*/categories",
-          "/*/*/posts",
-          "/*/*/pages",
-          "/*/*/media",
-          "/*/*/tags",
-          "/*/*/taxonomies",
-          "/*/*/users",
+          "**/*/*/categories",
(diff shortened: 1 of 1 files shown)


sidharthachatterjee:
LGTM, thanks @gurpreet-hanjra!

packages/gatsby-source-wordpress/README.md
```diff
@@ -117,13 +117,13 @@ module.exports = {
...
+          "**/*/*/categories",
+          "**/*/*/posts",
+          "**/*/*/pages",
+          "**/*/*/media",
+          "**/*/*/tags",
+          "**/*/*/taxonomies",
+          "**/*/*/users",
         ],
         // Blacklisted routes using glob patterns
         excludedRoutes: ["/*/*/posts/1456"],
```
sidharthachatterjee: Could you update this as well?

gurpreet-hanjra: Done. Thank you for pointing out.

Issue #10852, opened 2019-01-05
CSS Modules BEM structure


marcobiedermann:
## Description

When trying to use BEM structure or any other structure using hyphens in CSS modules, you get `undefined` as css class output.

### Steps to reproduce

style.module.css

```css
.block { … }
.block--modifier { … }
```

index.jsx

```jsx
import styles from './style.module.css';

const Component = () => (
  <div className={styles['block--modifier']}>
    …
  </div>
);
```

### Expected result

Expect class name in HTML to be processed through CSS modules

```html
<div class="style-module--block--modifier--SOME_HASH">…</div>
```

### Actual result

```html
<div class="undefined">…</div>
```

### Environment

```

  System:
    OS: macOS 10.14.1
    CPU: (8) x64 Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 11.3.0 - /usr/local/bin/node
    npm: 6.5.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 71.0.3578.98
    Safari: 12.0.1
  npmPackages:
    gatsby: ^2.0.64 => 2.0.64 
    gatsby-image: ^2.0.25 => 2.0.25 
    gatsby-plugin-google-analytics: ^2.0.8 => 2.0.8 
    gatsby-plugin-manifest: ^2.0.11 => 2.0.11 
    gatsby-plugin-offline: ^2.0.19 => 2.0.19 
    gatsby-plugin-react-helmet: ^3.0.4 => 3.0.4 
    gatsby-plugin-sharp: ^2.0.16 => 2.0.16 
    gatsby-plugin-sitemap: ^2.0.3 => 2.0.3 
    gatsby-source-filesystem: ^2.0.11 => 2.0.11 
    gatsby-transformer-remark: ^2.1.15 => 2.1.15 
    gatsby-transformer-sharp: ^2.1.9 => 2.1.9 

```

marcobiedermann:
Note:
I just figured out that Gatsby strips of all hyphens when using the class in JS.

Expected:

```jsx
<div className={styles['block--modifier']}>
```

Actual:

```jsx
<div className={styles['blockmodifier']}>
```

marcobiedermann:
@sidharthachatterjee 

At I see, thank you very much for pointing me to the webpack config

PR #10888, opened 2019-01-07
docs(www): Add note about BEM with CSS Modules

sidharthachatterjee:
Adds a note mentioning that we camelize class names with dashes and its effect on in context of BEM

Fixes https://github.com/gatsbyjs/gatsby/issues/10852

### DIFF
--- a/docs/tutorial/part-two/index.md
+++ b/docs/tutorial/part-two/index.md
@@ -490,6 +490,14 @@ The finished page should now look like:
 
 ![css-modules-final](css-modules-final.png)
 
+### CSS Modules and BEM
+
+If you're used to writing CSS with the [BEM](http://getbem.com/) methodology, a quick thing to note is that Gatsby camelizes class names that contain dashes.
+
+For example: `block--modifier` turns into `blockModifier`.
+
+We do this to make access in JavaScript easier.
+
 ### Other CSS options
 
 Gatsby supports almost every possible styling option (if there isn't a plugin


sidharthachatterjee:
@DSchau I definitely agree that CSS Modules + BEM shouldn't be used together! I could edit this into a more generic note about dashes being camelized but I'd guess the most _likely_ scenario a user would hit this in when trying to name classes using BEM

Maybe a good middle ground is making the paragraph generic and then adding another note below—that one ought to keep this in mind if using BEM with CSS Modules (while we certainly discourage it)

DSchau:
> I'd guess the most likely scenario a user would hit this in when trying to name classes using BEM

What if we just _showed_ (rather than told) an example using dash-case/kebab-case, e.g. something like:

```css
.user-avatar {
}
```

import styles from './whatever.css'

// styles.userAvatar

Then it becomes apparent going through the tutorial, rather than needing to be something explicitly called out? I'm not really sold on the idea--but I think I like that better than a separate section. What do you think?

KyleAMathews:
We don't want to add more explanations than necessary to the tutorial. There's a _lot_ of people going through it that have no idea what BEM so any clarification of BEM will likely confuse more than help.

The place for detailed clarifications like this would be a docs page on CSS Modules and not in the tutorial.

DSchau:
@sidharthachatterjee going to close this per @KyleAMathews's comment above.

Feel free to re-open _basically this same_ PR on the CSS Modules doc instead :)

PR #10892, opened 2019-01-07
fix(gatsby-transformer-remark): correctly pass cache to sub plugins

DSchau:
## Description

Separate some changes from #10146, specifically so we can make this a non-breaking change if the right version of Gatsby isn't specified.

Note: I will add a warning/log (or error?) if the API doesn't exist so the user can update.

## Related Issues

Related to #10146

### DIFF
--- a/e2e-tests/development-runtime/plugins/gatsby-remark-subcache/index.js
+++ b/e2e-tests/development-runtime/plugins/gatsby-remark-subcache/index.js
@@ -1,0 +1,11 @@
+const visit = require(`unist-util-visit`)
+const { id } = require(`./constants`)
+
+module.exports = function remarkPlugin({ cache, markdownAST }) {
+  visit(markdownAST, `html`, async node => {
+    if (node.value.match(id)) {
+      const value = await cache.get(id)
+      node.value = node.value.replace(/></, () => `>${value}<`)
+    }
+  })
+}
--- a/packages/gatsby-transformer-remark/src/extend-node-type.js
+++ b/packages/gatsby-transformer-remark/src/extend-node-type.js
@@ -67,7 +67,16 @@ const withPathPrefix = (url, pathPrefix) =>
 const ASTPromiseMap = new Map()
 
 module.exports = (
-  { type, store, pathPrefix, getNode, getNodesByType, cache, reporter },
+  {
+    type,
+    pathPrefix,
+    getNode,
+    getNodesByType,
+    cache,
+    getCache,
+    reporter,
+    ...rest
+  },
   pluginOptions
 ) => {
   if (type.name !== `MarkdownRemark`) {
@@ -150,7 +159,9 @@ module.exports = (
               files: fileNodes,
               getNode,
               reporter,
-              cache,
+              cache: getCache(plugin.name),
+              getCache,
+              ...rest,
             },
             plugin.pluginOptions
           )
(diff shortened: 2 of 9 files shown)


packages/gatsby-transformer-remark/src/extend-node-type.js
```diff
@@ -58,6 +58,26 @@ const tableOfContentsCacheKey = node =>
 const withPathPrefix = (url, pathPrefix) =>
   (pathPrefix + url).replace(/\/\//, `/`)
 
+// TODO: remove this check with next major release
+// the getCache feature was implemented in gatsby@2.0.88
+const safeGetCache = ({ getCache, cache, reporter }) => {
```
DSchau: @pieh this warning looks like:

```shell
success onPreBootstrap — 0.004 s
success source and transform nodes — 0.072 s
success building schema — 0.175 s
success createPages — 0.030 s
success createPagesStatefully — 0.021 s
success onPreExtractQueries — 0.003 s
success update schema — 0.102 s
success extract queries from components — 0.119 s
warning It looks like you're using a slightly outdated version of gatsby with gatsby-transformer-remark.
To update, run 'npm install gatsby@^2.0.86 --save' or 'yarn add gatsby@^2.0.86'
```

I thought about using `onPreBootstrap` but I like how we can remove this code all in one spot (easier refactoring). Up for whatever though.

DSchau: Also it'd be cool to think more broadly about how we introduce non-backwards compatible API changes, and if we could improve this.

For now this will work fine.

e2e-tests/development-runtime/package.json
```diff
@@ -4,7 +4,7 @@
   "version": "1.0.0",
   "author": "Dustin Schau <dustin@gatsbyjs.com>",
   "dependencies": {
-    "gatsby": "^2.0.71",
+    "gatsby": "^2.0.88",
```
DSchau: These don't really matter. Just for local stuff!

e2e-tests/development-runtime/plugins/gatsby-remark-subcache/index.js
```diff
@@ -5,7 +5,7 @@ module.exports = function remarkPlugin({ cache, markdownAST }) {
   visit(markdownAST, `html`, async node => {
     if (node.value.match(id)) {
       const value = await cache.get(id)
-      node.value = node.value.replace(/%SUBCACHE_VALUE%/, () => `>${value}<`)
+      node.value = node.value.replace(/%SUBCACHE_VALUE%/, value)
```
DSchau: 60% less jank ™️

packages/gatsby-transformer-remark/src/extend-node-type.js
```diff
@@ -59,23 +59,11 @@ const withPathPrefix = (url, pathPrefix) =>
...
-        warned = true
-        reporter.warn(reporter.stripIndent`
-          It looks like you're using a slightly outdated version of gatsby with gatsby-transformer-remark.
-          To update, run 'npm install gatsby@^${GATSBY_VERSION_WITH_GET_CACHE} --save' or 'yarn add gatsby@^${GATSBY_VERSION_WITH_GET_CACHE}'
-        `)
-      }
-      return cache
-    }
-    return getCache(id)
+const safeGetCache = ({ getCache, cache }) => id => {
```
DSchau: Note for posterity, we _already_ warn on conflicting peerDeps, so a second warning would be sorta redundant.

PR #10780, opened 2019-01-02
docs(blog): add seo and gatsby blog post

DSchau:
## Description

This PR adds a blog post re: SEO + Gatsby, with a particular focus on the new SEO component we've made available in the default and blog starters.

I'm open to tweaking this however (or making a guide?), but didn't want to overwrite anything existing content and it does seem to make sense as a blog post.

As always, feedback _more_ than welcome, and huge shout-out to @fk and @LekoArts for lots of prior art, here!

## Preview

localhost_8000_blog_2019-01-04-seo-and-gatsby_

PR #10894, opened 2019-01-07
feat(starters): pass through rest of the props, and support children

DSchau:
## Description

With some feedback from @khalwat I think it's prudent to enable some additional features in the SEO component. I want to balance _smart defaults_ with something that's also extensible, and I think this does this pretty well.

Specifically, this enables the (small) features like:

- Ability to pass an array of `link` items
- Pass through any additional props directly to the `<Helmet />` component
- Expose the `children` prop for an alternative API, which some could prefer

Ideally - I'd be able to show how to make images absolute (prefixing with siteUrl), link rel="canonical", etc. but I also want to strike the balance between not offering too opinionated of a component out of the box. We want something relatively simple that can be built upon, and I think building those features in makes too many decisions that are best left to the developer.

For instance, to add a `link rel="canonical"` it'd be something like the below:

```jsx
-function SEO({ children, description, lang, meta, keywords, title, ...rest }) {
+function SEO({ canonical, children, description, lang, link, meta, keywords, title, ...rest }) {
   return (
     <StaticQuery
       query={detailsQuery}
       render={data => {
         const metaDescription =
           description || data.site.siteMetadata.description
+        const siteUrl = data.site.siteMetadata.siteUrl
         return (
           <Helmet
             htmlAttributes={{
@@ -17,6 +18,16 @@ function SEO({ children, description, lang, meta, keywords, title, ...rest }) {
             }}
             title={title}
             titleTemplate={`%s | ${data.site.siteMetadata.title}`}
+            link={
+              []
+                .concat(canonical ? [
+                  {
+                    rel: 'canonical',
+                    href: `${siteUrl}/${canonical}`
+                  }
+                ] : [])
+                .concat(link)
+            }
             meta={[
               {
                 name: `description`,
@@ -72,11 +83,13 @@ function SEO({ children, description, lang, meta, keywords, title, ...rest }) {
 
 SEO.defaultProps = {
   lang: `en`,
+  link: [],
   meta: [],
   keywords: [],
 }
 
 SEO.propTypes = {
+  canonical: PropTypes.string,
   children: PropTypes.node,
   description: PropTypes.string,
   lang: PropTypes.string,
@@ -93,6 +106,7 @@ const detailsQuery = graphql`
       siteMetadata {
         author
         description
+        siteUrl
         title
       }
     }

```

My approach here will be I'll show _how_ to extend this component in #10780 so that it's clear how we can build upon this solid base.

## Related Issues

Related to #10780.

### DIFF
--- a/starters/blog/src/components/seo.js
+++ b/starters/blog/src/components/seo.js
@@ -94,7 +94,7 @@ const detailsQuery = graphql`
     site {
       siteMetadata {
         author
-        description
+        description(pruneLength: 160)
         title
       }
     }
--- a/starters/blog/src/components/seo.js
+++ b/starters/blog/src/components/seo.js
@@ -73,6 +73,7 @@ function SEO({ children, description, lang, meta, keywords, title, ...rest }) {
 SEO.defaultProps = {
   lang: `en`,
   meta: [],
+  link: [],
   keywords: [],
 }
 
@@ -80,6 +81,7 @@ SEO.propTypes = {
   children: PropTypes.node,
   description: PropTypes.string,
   lang: PropTypes.string,
+  link: PropTypes.array,
   meta: PropTypes.array,
   keywords: PropTypes.arrayOf(PropTypes.string),
   title: PropTypes.string.isRequired,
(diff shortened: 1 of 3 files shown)


KyleAMathews:
Just a thought — have you considered turning the SEO component into its own package — `gatsby-seo`? Then we could reuse this more easily across starters + people could pull it into their own sites. SEO seems like something that's pretty consistent across sites. Releasing this new component alongside your blog post would make for a nice "release post".

DSchau:
Let's close this. I'd rather introduce the SEO component instead of going this route.

PR #10898, opened 2019-01-07
Added Documentation for using Tailwind with Gatsby

jakedohm:
## Description

I added a document with a walkthrough of how to get Tailwind CSS up and running with Gatsby via PostCSS. Based on: https://dev.to/jakedohm_34/using-tailwind-with-gatsby-js-10fj

Let me know if anything is missing, or any corrections need to be made!

### DIFF
--- a/docs/docs/tailwind-css.md
+++ b/docs/docs/tailwind-css.md
@@ -6,11 +6,33 @@ Tailwind is a utility-first CSS framework for rapidly building custom user inter
 
 This guide assumes that you have a Gatsby project set up. If you need to set up a project, head to the [**Quick Start guide**](https://www.gatsbyjs.org/docs), then come back.
 
-### Installation and Configuration
+### Overview
 
-Tailwind uses PostCSS under the hood, so you will need PostCSS set up before we install Tailwind.
+There are two ways you can use Tailwind with Gatsby:
+1. Standard: Use PostCSS to generate Tailwind classes, then you can apply those classes using `className`.
+2. CSS-in-JS: Integrate Tailwind classes into Styled Components.
 
-1.  Install the Gatsby plugin [**gatsby-plugin-postcss**](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-postcss).
+You have to install and configure Tailwind for both of these methods, so this guide will walk through that step first, then you can follow the instructions for either PostCSS or CSS-in-JS.
+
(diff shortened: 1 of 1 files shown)


docs/docs/tailwind-css.md
```diff
@@ -0,0 +1,57 @@
...
+
+You can now use the `@tailwind` directives to add Tailwind's utilites, preflight, and components into your CSS. You can also use `@apply` and all of Tailwind's other directives and functions!
+
+To learn more about how to use Tailwind in your CSS, visit the [Tailwind Documentation](https://tailwindcss.com/docs/installation#3-use-tailwind-in-your-css)
+
+
+### Other resources
+
+- [Introduction to postcss](https://www.smashingmagazine.com/2015/12/introduction-to-postcss/)
+- [Tailwind Documentation](https://tailwindcss.com/)
```
LekoArts: Linking to starters could be handy!

```suggestion
- [Tailwind Documentation](https://tailwindcss.com/)
- [Gatsby starters that use Tailwind](https://www.gatsbyjs.org/starters/?c=Styling%3ATailwind&v=2)
```

hagnerd:
Also food for thought, I would recommend adding in how to use it with PurgeCSS since Tailwind generates so much CSS and the end user likely will only use a small portion of what was generated.

jakedohm:
I think I'm going to re-structure the article like so:

Tailwind Installation & Configuration
- Installing Tailwind
- Adding config file

Using Tailwind
- PostCSS
    - Installation/Usage steps
- CSS-in-JS
    - Installation/Usage steps

Since the you have to install and configure Tailwind for _both_ methods, I think it makes sense to only do that stuff in one place, and *then* show them how to implement either method. Does that sound good to you @LekoArts?

@hagnerd I'll also add instructions for how to enable PurgeCSS!

LekoArts:
Yeah, sounds good!

calcsam:
Hey @jakedohm, wanted to check where you were on this? More and more people are asking about Gatsby and Tailwind so would be great to get this in if you're able to finish it up!

jakedohm:
@calcsam @LekoArts hi again! I've updated the docs in my latest commit to show steps for usage with both PostCSS (standard) and CSS-in-JS.

I think this should be pretty close, but feedback is welcome!

calcsam:
This looks great, thanks @jakedohm! I'll add a follow-up PR adding this to the docs hierarchy!

PR #7872, opened 2018-09-04
chore(jest): remove redundancies from jest.config

azdanov:
Repeat of https://github.com/gatsbyjs/gatsby/pull/7434.

PR #10913, opened 2019-01-08
Update babel-jest

stefanprobst:
#7872 updated `jest` to 23.5, but kept `babel-jest` at 22.4.3


stefanprobst:
Hmm, this breaks linting horribly. Needs investigation - seems ok as long as yarn.lock is not removed.

stefanprobst:
Ok because of prettier 1.14 => 11.5. Should 1.14 be pinned?

stefanprobst:
Fixed by #11293

Issue #9139, opened 2018-10-16
Google analytics tracks pages, but not always the path and the title are from the same page. Sometimes a visitor visits "/" but the title of another url. 


hackhat:
## Description

Google analytics tracks pages, but not always the path and the title are from the same page. Sometimes a visitor visits "/" but the title of another url. 

Looks like the url is correct but the title is from the previous page I've seen. I'm using `import { Link } from 'gatsby'` for the links.

### Steps to reproduce

Created default app:

Add Google analytics script:

    {
      resolve: `gatsby-plugin-google-analytics`,
      options: {
        trackingId: "UA-12420899-9",
        // Puts tracking script in the head instead of the body
        head: false,
        // Setting this parameter is optional
        anonymize: false,
        // Setting this parameter is also optional
        respectDNT: false ,
        // Avoids sending pageview hits from custom paths
        exclude: [],
        // Any additional create only fields (optional)
        sampleRate: 100,
        siteSpeedSampleRate: 100,
        cookieDomain: "www.freereikicourse.com",
      },
    },

Add helmet in your page template:

        <Helmet
          title={frontmatter.title}
          meta={[
            { name: 'description', content: frontmatter.description },
            { name: 'keywords', content: frontmatter.keywords },
          ]}
        >
          <html lang="en"/>
        </Helmet>

### Expected result

All pages should have the path matching with the title.

### Actual result

Google analytics tracks pages, but not always the path and the title are from the same page. Sometimes a visitor visits "/" but the title of another url. 

### Environment

gatsby info --clipboard:

        System:
            OS: Windows 10
            CPU: x64 Intel(R) Core(TM) i7-5600U CPU @ 2.60GHz
          Binaries:
            Yarn: 1.9.4 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
            npm: 5.5.1 - C:\Program Files\nodejs\npm.CMD
          Browsers:
            Edge: 42.17134.1.0
          npmPackages:
            gatsby: ^2.0.19 => 2.0.21
            gatsby-plugin-catch-links: ^2.0.4 => 2.0.4
            gatsby-plugin-emotion: ^2.0.5 => 2.0.5
            gatsby-plugin-google-analytics: ^2.0.6 => 2.0.6
            gatsby-plugin-manifest: ^2.0.5 => 2.0.5
            gatsby-plugin-offline: ^2.0.5 => 2.0.6
[...]

PR #3362, opened 2017-12-29
[gatsby-plugin-google-analytics]Wait for the title update

ntwcklng:
As Kyle suggested, we should wrap the Google-Analytics function into a `setTimeout` so it waits until the title is updated.

see #2478

Issue #2478, opened 2017-10-16
[gatsby-plugin-react-helmet] and [gatsby-plugin-google-analytics] don't play well with Title changes


jbrantly:
When using React Helmet and the Google Analytics plugin together, it seems that the route change handler causing GA to fire occurs before React Helmet has a chance to update the `<title>`. This results in GA data often having the *previous* page title for any given URL. 

This can be seen in the existing gatsbyjs.org site by navigating to the homepage, then to the "Community" page, and back and forth. The Community page has a different title than the other pages, and if you check the Network tab for the GA tracker you can see that the title is always the previous page.

KyleAMathews:
Yeah that makes sense!

Perhaps we just wrap the GA update in a setTimeout to let react helmet finish before running?

How's that sound? Want to put together a PR?

tylercrosse:
I haven't dug too deeply into this but using [nfl's react metrics](https://github.com/nfl/react-metrics) might be an option

LeKoArts:
@KyleAMathews Would also be interested in a solution? So you propose to set a timeout here?
https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-google-analytics/src/gatsby-browser.js

KyleAMathews:
@LeKoArts yeah exactly. Just set a timeout of 0 and that *should* be enough to make GA wait for the title update to finish.

Chuloo:
There hasn't been any activity on this issue recently. Due to the high number of incoming GitHub notifications, we have to clean some of the old issues as many of them have already been resolved with the latest updates or explained in a previous issue.

Feel free to [open a new one](https://github.com/gatsbyjs/gatsby/issues/new/choose) if you still experience this problem or a similar one! 👍

PR #10917, opened 2019-01-08
fix(google-analytics): fix pageview timing issue by delaying it

jorgenblindheim:
There is still a problem with title and path not being correct while navigating across a page.

As referenced in: #9139
Solution made in:  #2478 and #3362

For some reason the previous solution is being overwritten


pieh:
> For some reason the previous solution is being overwritten

It might be that context on why setTimeout was used was missing - can you add comment there to not why this is needed?

drake-smith:
@jorgenblindheim @pieh  Checking in here. I am experiencing the same issue (page title fired off to GA is the old page's title because GA updates before React-Helmet does). I've simply installed `gatsby-plugin-google-analytics` and added my Tracking ID to gatsby-config.

 Will this PR be merged in? This is a pretty fundamental bug that i think should be addressed as it essentially sending inaccurate info to Analytics.

wardpeet:
@drake-smith i'll be checking this out.

PR #10919, opened 2019-01-08
fix(gatsby-transformer-remark): remove unused _PARENT field from frontmatter

stefanprobst:
`gatsby-transformer-remark` puts a `_PARENT` field on frontmatter, that does not seem to be used. Maybe from a time before node-tracking?


pieh:
Ah, that's dead code that ~1 year ago was needed to resolve paths to files in frontmatter to File nodes.

Nice catch.

PR #10920, opened 2019-01-08
chore: use null not "SOURCE" as parent of top level nodes

stefanprobst:
Currently top nodes are represented with either `parent: "SOURCE"` or `parent: "___SOURCE___"`, which is not a valid node id. This PR proposes to just use `null` instead.


KyleAMathews:
Ah nice! You're finding all my old design decisions where things changed but I didn't update all the code 😅 Thanks!

Issue #2114, opened 2017-09-14
Consider generating multiple bundles in favor of modern browsers


kripod:
According to [this article](https://philipwalton.com/articles/deploying-es2015-code-in-production-today/), modern browsers may profit from bundles loaded with support for the ES2015 module syntax. Using the methods described in the article may also reduce server load because of the reduced bundle sizes.

KyleAMathews:
Huh super interesting! Would love to see someone set this up on gatsby! Should be doable as a plugin.

Btw, the main benefit appears to be that you don't have to polyfill for newer browsers so the code loaded is significantly smaller.

jquense:
That article seems confused (or I am)...the main jist is to not compile modules via babel. The problem with that tho is that you don't get a bundle out that has uncompiled es modules, you get a compiled webpack bundle with _no_ modules. Babel won't compile the modules but webpack *will*. So you can use the "modern" babel config with any browser, in fact the v2 gatsby branch (which uses webpack2) already does this.

jquense:
ahhhh I get it sorry i'm the dense one. The point is just that you can simplify feature detection by checking `script module` tags?

KyleAMathews:
@davidtheclark over at Batfish is working on a PR for a similar feature https://github.com/mapbox/batfish/pull/188

This technique seems *really* promising.

We'd need to see how much time it adds to the build but if it's not that much more, I'd definitely consider turning it on by default. If the extra build time hurts enough, we could make it optional.

robwierzbowski:
@jquense I'm really into this approach. To summarize my understanding:

- Transpiled code is often larger and less performant than es6+ code.
- `script type="module"` is a good "cut the mustard" line for a lot of es6+ features. Remember [cut the mustard](http://responsivenews.co.uk/post/18948466399/cutting-the-mustard) from maybe 5-6 years ago?
- We should still bundle our modules
- We can create a bundle with much less transpiling for modern browsers, and use  `script type="module"` as a sneaky feature detect to load a minimally transpiled bundle on newer browsers, and a more transpiled / es5 bundle on older browsers.

KyleAMathews:
Yup! And per our parallel polyfill discussion, the  `script type="module"` would need pretty minimal polyfills and be very cheap to deliver them over CDN.

davidtheclark:
[...]

PR #10921, opened 2019-01-08
[WIP] feat(gatsby-seo): introduce new Gatsby search engine optimization component

DSchau:
## Description

This PR introduces a new, reusable component we're calling `gatsby-seo`. This is a thin wrapper around react-helmet that sets up some sane defaults using a user's `gatsby-config.js`.

The ultimate goal is to introduce this into the starters, and continue to iterate on ideas established within this component. In much the same way that `gatsby-image` is ubiquitous and useful, I'd like to establish this component as a useful component that can be used in any gatsby application.

## Related Issues

Related to #10780. Addresses [a comment](https://github.com/gatsbyjs/gatsby/pull/10894#issuecomment-452024978) in #10894.


DSchau:
Few thoughts:

- I could probably add in `image` capability. `image` should be a single item with src/height/width properties
  - This gets a little more complicated with `aspect` ratios (e.g. for Twitter, Facebook, etc.) so didn't want to get too opinionated
- I want to keep this with sensible defaults, while also being super extensible. I _think_ this strikes that balance, but if there are areas where we can improve the defaults with relatively innocuous opinions, all for it. (@khalwat -- if you have any time, I would love any thoughts you have here!)

packages/gatsby/src/internal-plugins/query-runner/query-compiler.js
```diff
@@ -104,6 +104,7 @@ class Runner {
     let files = [
       path.join(this.base, `src`),
       path.join(this.base, `.cache`, `fragments`),
+      path.join(this.base, `node_modules`, `gatsby-seo`),
```
DSchau: We will need this to avoid `Loading StaticQuery...` issues.

Should I conditionally check whether this directory exists? It will just ignore it as far as I know if it's empty--so not a huge concern.

DSchau:
This currently fails to build/extract the query, which I think is because of the way babel transpiles queries, leading to the following error:

> Error: It appears like Gatsby is misconfigured. Gatsby related `graphql` calls are supposed to only be evaluated at compile time, and then compiled away,. Unfortunately, something went wrong and the query was left in the compiled code.

```js
var _gatsby = require("gatsby");
var _taggedTemplateLiteralLoose2 = _interopRequireDefault(require("@babel/runtime/helpers/taggedTemplateLiteralLoose"));

function _templateObject() {
  var data = (0, _taggedTemplateLiteralLoose2.default)(["\n  query DefaultSEOQuery {\n    site {\n      siteMetadata {\n        title\n        description\n        author\n      }\n    }\n  }\n"]);

  _templateObject = function _templateObject() {
    return data;
  };

  return data;
}

var detailsQuery = (0, _gatsby.graphql)(_templateObject());
```

Will investigate!

packages/gatsby-seo/src/index.js
```diff
@@ -0,0 +1,114 @@
...
+                property: `og:type`,
+                content: `website`,
+              },
+              {
+                name: `twitter:card`,
+                content: `summary`,
+              },
+              {
+                name: `twitter:creator`,
+                content: data.site.siteMetadata.author,
```
gregoralbrecht: Hey 👋 I'm not sure if you actually want feedback on this already, so please excuse me if this is too early for me to comment or not my place at all (I'm a new maintainer).

I love the idea of a `<SEO>` component because it seems like a lot of the starters use something like this already! However, I'm not sure about the defaults for the Twitter meta tags. [According to the docs](https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/markup.html), `twitter:creator` is only used in combination with `twitter:card` of type `summary`, so the way you have it right now isn't correct, according to their docs. Also, the username on Twitter is usually different than what people would expect to put in `siteMetadata.author`, so maybe it should be a separate field?

DSchau: Happy for feedback, whenever! If you have anything else--feel free to comment on whatever!

> twitter:creator is only used in combination with twitter:card of type summary

Isn't it used that way, currently? The username is/could be wrong (e.g. author name instead of twitter handle)--I was going to tweak that, but didn't want to introduce a breaking change (e.g. you have to augment your gatsby-config.js to take advantage of this component) for anyone picking this up.

packages/gatsby/src/internal-plugins/query-runner/query-compiler.js
pieh: I really don't like targeting specific packages in gatsby core to extract queries. This is something that ultimately will need to be solved to enable query extraction (and nicer fragments extraction as what plugins currently do to register fragments is copying files with fragment definitions to arbitrary `.cache/fragments` location)

DSchau: Totally agree. I'll think on a better way to do this.

packages/gatsby-seo/src/index.js
gregoralbrecht: Oh my bad! I've been switching between projects where I had a similar but slightly issue for too long today, sorry. I understand your concern with the breaking change.

t2ca:
Fantastic work Dustin, have you considered adding a prop for schema.org

I have something like this in my project.

```
<Helmet>
  <script type="application/ld+json">
     {JSON.stringify(schemaOrgJSONLD)}
  </script>
<Helmet>

  const schemaOrgJSONLD = [
    {
      '@context': 'http://schema.org/',
      '@type': 'Review',
      itemReviewed: {
        '@type': 'Thing',
        name: variable
      },
      author: {
        '@type': 'Person',
        name: variable
      },
      reviewRating: {
        '@type': 'Rating',
        ratingValue: variable,
        bestRating: 100,
      },
    },
  ]
```

DSchau:
This is probably still worth doing--but requires a few, non-trivial things to implement to unblock (e.g. transpiling node_modules with babel) which won't be done in a reasonable amount of time to enable this.

Some similar work is also required in #2114, so let's close this until this is unblocked. I also think this could be a fairly decent use of a theme, e.g. `gatsby-theme-seo`, which _are_ transpiled and which would ensure that `StaticQuery` is extracted in the reusable component. I think it would end up looking something like:

```js
import React from 'react'
import SEO from 'gatsby-theme-seo/src/components/seo'

// use the component
```

khalwat:
I published this article recently that might be of interest as well: [Annotated JSON-LD Structured Data Examples](https://nystudio107.com/blog/annotated-json-ld-structured-data-examples#dev-mode-fm-episode-page)

JSON-LD Structured Data can be a little convoluted at first blush.

PR #10924, opened 2019-01-08
fix(gatsby-transformer-remark): don't convert Date objects

stefanprobst:
`gatsby-transformer-remark` parses frontmatter with `gray-matter` and then iterates through frontmatter fields again to transform Dates into strings.

This PR proposes to only allow types from the YAML 1.2 spec, which has no `timestamp` type anymore (the 1.1 spec has). This way, dates will automatically be parsed as strings.

For the different `schema` options that the yaml parser accepts [see here](https://github.com/nodeca/js-yaml#safeload-string---options-).

### DIFF
--- a/packages/gatsby-transformer-remark/src/on-node-create.js
+++ b/packages/gatsby-transformer-remark/src/on-node-create.js
@@ -1,6 +1,5 @@
 const grayMatter = require(`gray-matter`)
 const crypto = require(`crypto`)
-const yaml = require(`js-yaml`)
 
 module.exports = async function onCreateNode(
   { node, loadNodeContent, actions, createNodeId, reporter },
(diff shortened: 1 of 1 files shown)


pieh:
I actually wonder if this is still problem (fact that Date object are potentially used). I remember fixing inconsistencies when Date object was used, then cached in `.cache/redux-state.json` and read from in next run - it might be just unnecessary code and we don't need to specifically declare schema without that don't support Date objects here?

stefanprobst:
Ah indeed, just checked and it works fine because of [`isMixOfDateObjectsAndDateStrings`](https://github.com/gatsbyjs/gatsby/blob/86fc9dcf2e5cacf90be147a50065235702079f3f/packages/gatsby/src/schema/data-tree-utils.js#L63).

freiksenet:
@stefanprobst @pieh Is this PR still relevant?

pieh:
Yeah, it's still worth to get in, will just validate that change now so at least 2 sets of eyes checked it.

zslabs:
I have a date field within my project that has some written as strings, and others not (have content creators doing both unfortunately); and this PR appears to be breaking the sorting functionality I'm using.

PR #10927, opened 2019-01-08
chore: fix forestry yml files from #10925

DSchau:
## Description

#10925 had some lint issues, this (should) fix them.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -29,3 +29,6 @@ infrastructure
 
 # coverage
 coverage
+
+# forestry files
+.forestry/**/*


.forestry/front_matter/templates/blog.yml
```diff
@@ -1,35 +1,34 @@
----
 label: blog
 hide_body: false
 is_partial: false
 fields:
-- name: title
```
DSchau: @jpatters our linter/prettier seems to think this is invalid.

Will forestry still work without these separators? Otherwise, we'll need to add the .forestry folder to our ignored patterns list!

pieh: I'd say let's add `.forestry` to `.prettierignore` at least for now so it doesn't block us. In later PR we could un-ignore it and format it

jpatters: Forestry will be happy as long as it is valid yaml
```
fields:
- name: title
```
and 
```
fields:
  - name: title
```

are the same thing. (i assume that is likely the complaint)

Issue #10901, opened 2019-01-07
[tutorial] Rewrite the styles section of the tutorial for clarity, focus on Gatsby functionality


jlengstorf:
Currently, part 2 of the tutorial focuses on installing and configuring Typography.js more than showing people how to add global and component styles in Gatsby.

To improve this, let's rewrite this section of the tutorial to focus on Gatsby exclusively, and move third-party scripts and plugins to a "further reading" section.

## Proposed Outline

- [x] Introduction
- [ ] Global Styles using a standard `*.css` stylesheet imported into `gatsby-browser.js`
- [ ] Component CSS using CSS Modules
- [ ] A brief intro to CSS-in-JS with links to additional docs/plugins
- [ ] Gatsby style plugins and additional resources (e.g. Typography.js, Sass, Bootstrap)

My thinking is that this follows a "least complex first" approach:

1. If I’m a developer coming across from plain HTML, CSS, and JavaScript dev, I’m used to dropping a `<link rel="stylesheet" href="main.css" />` into the head of my document and adding styles to `main.css` as plain ol' CSS.

    By focusing on Gatsby’s equivalent of a global stylesheet, we don’t introduce too much magic here: we need to explain why we can't just drop a `link` into `Helmet`, but otherwise it's a global stylesheet like any other.

2. If I'm interested in component styling, but not familiar with CSS-in-JS, CSS Modules are a really accessible way to write plain CSS while getting extra benefits.

3. If I read on, I can learn more about taking CSS Modules to the next level with CSS-in-JS.

4. Going further, I can learn about Gatsby plugins for various third-party styling options with links to their docs.

Based on feedback we’ve seen, I think this will help smooth the process of learning how CSS works in Gatsby.

cc: @amberleyromo @shannonbux

amberleyromo:
@jlengstorf Going through old issues -- I think I closed this out with https://github.com/gatsbyjs/gatsby/pull/10929

jlengstorf:
You sure did!

PR #10929, opened 2019-01-08
[tutorial] Rewrite the styles section

amberleyromo:
A la #10901. WIP.

## Revised outline:
- Intro
- Global styles using a standard `*.css` stylesheet imported into `gatsby-browser.js` (mention shared layouts)
- Component CSS using CSS Modules
- A brief intro to CSS-in-JS with links to additional docs/plugins
- Gatsby style plugins and additional resources (e.g. Typography.js, Sass, Bootstrap)

## Questions:
- @jlengstorf: I've been debating. The way part two was structured before served to seamlessly introduce the concept of plugins. I'm wondering if perhaps we reintegrate Typography.js as the example plugin, so that that part of the tutorial is intact. Otherwise, I think the first mention of plugins is part four (data), maybe.

## Todo:
- [x] Update the screenshots (on approval of the new content flow)
~- [ ] Add a glossary of terms (I think this section needs it)~
- [x] Migrate typography.js / plugin intro to beginning of part three

## Later PR:
- [ ] Ensure that what's been removed of the Typography.js content isn't lost -- needs to be migrated / worked in to the Typography.js guide
- [ ] This prompts a reorg/revisitation of some style docs, particularly component css and css modules


docs/tutorial/part-two/index.md
```diff
@@ -3,91 +3,45 @@ title: Introduction to Using CSS in Gatsby
...
 
-Start by creating a new site. At this point, it probably makes sense to close the terminal windows you used to build tutorial-part-one so that you don't accidentally start building tutorial-part-two in the wrong place. If you don't close tutorial-part-one prior to building tutorial-part-two, you will see that tutorial-part-two appears at localhost:8001 instead of localhost:8000.
-
-Just like in part one, open a new terminal window and run the following commands to create a new Gatsby site in a directory called `tutorial-part-two`. Then, change to this new directory:
+Open a new terminal window, create a new "hello world" gatsby site, and start the development server:
 
 ```shell
 gatsby new tutorial-part-two https://github.com/gatsbyjs/gatsby-starter-hello-world
 cd tutorial-part-two
+gatsby develop
```
jlengstorf: I don't think we want to start it just yet, do we? They'll need to restart for `gatsby-browser.js` to pick up the changes.

docs/tutorial/part-two/index.md
```diff
@@ -96,238 +50,102 @@ This creates a new site with the following structure.
...
+mkdir styles
+cd styles
+touch global.css
 ```
 
-Next, in your code editor, create a file at the root of your project folder named `gatsby-config.js`.
-This is where you add plugins along with other site configuration.
-
-Copy the following into `gatsby-config.js`
+> Note: You'll need to either stop or restart the running development server, or open a second terminal window. Also, feel free to create these directories and files using your code editor, if you'd prefer.
```
jlengstorf: If we don't have them start the server, this can be dropped.

docs/tutorial/part-two/index.md
```diff
@@ -96,238 +50,102 @@ This creates a new site with the following structure.
...
+├── src
+│   └── pages
+│       └── index.js
+│   └── styles
+│       └── global.css
+├── gatsby-browser.js
 ```
 
-![typography-bootstrap](typography-bootstrap.png)
+> 💡 What is `gatsby-browser.js`? Don't worry about this too much for now -- For now, know that `gatsby-browser.js` is one of a handful of special files that Gatsby looks for, and uses (if they exist). Here, the naming of the file **is** important.
```
jlengstorf: Should we link to [browser APIs](https://www.gatsbyjs.org/docs/browser-apis/) here in case they want to dig deeper?

docs/tutorial/part-two/index.md
```diff
@@ -96,238 +50,102 @@ This creates a new site with the following structure.
...
 ```
 
-![typography-lawton](typography-lawton.png)
-
-_Challenge:_ Typography.js has more than 30 themes!
-[Try them live](http://kyleamathews.github.io/typography.js) or check out
-[the complete list](https://github.com/KyleAMathews/typography.js#published-typographyjs-themes) and try installing one on your current Gatsby site.
-
-## Component CSS
+> Note: You can use Node.js `require` or `import` syntax. If you're not sure of the difference, just pick one.
```
jlengstorf: Added a little bit of extra info and made a recommendation instead of putting the choice on the reader.

```suggestion
> Note: Both CommonJS (`require`) and ES Module (`import`) syntax work here. If you’re not sure which to choose, we use `import` most of the time.
```

docs/tutorial/part-two/index.md
```diff
@@ -96,238 +50,102 @@ This creates a new site with the following structure.
...
 
-Gatsby has a wealth of options available for styling components. In this tutorial, you'll explore
-one very popular method: CSS Modules.
+If you take a look another look at your "hello world" project site, you should see your global css changes applied.
```
jlengstorf:
```suggestion
If you take a look another look at your "hello world" project site, you should see your global CSS changes applied.
```

jlengstorf: Also: should we include an image for posterity?

docs/tutorial/part-two/index.md
```diff
@@ -486,16 +297,33 @@ export default () => (
 )
 ```
 
+> Tip: Generally, if you use a component in multiple places on a site, it should be in its own module file in the `components` directory. But, if it's used only in one file, create it inline.
```
jlengstorf: I'm not sure we need this tip. Might spark some bikeshedding, and could also add to overwhelm ("I have to think about _that_, too?!").

docs/tutorial/part-two/index.md
```diff
@@ -486,16 +297,33 @@ export default () => (
...
 
 ![css-modules-final](css-modules-final.png)
 
+### CSS-in-JS
+
+CSS-in-JS is a component-oriented styling approach. Most generally, it is a pattern where [CSS is composed inline using JavaScript](https://reactjs.org/docs/faq-styling.html#what-is-css-in-js).
+
+#### Using CSS-in-JS with Gatsby
+
+There are many different CSS-in-JS libraries, and many of them have Gatsby plugins already. We won't cover an example of CSS-in-JS in this initial tutorial, but we encourage you to [explore](/docs/styling/) what the ecosystem has to offer. There are mini-tutorials for two libraries in particular, [Glamor](/docs/glamor/) and [Styled Components](/docs/styled-components/).
```
jlengstorf: We need an Emotion tutorial. ☹️ 👩‍🎤

amberleyromo: I was thinking this as I was doing this rewrite. I'm on a roll, I might go ahead and do it.

That reminds me, something on the todo i need to add is some tweaks to some of the related style docs.

docs/tutorial/part-two/index.md
amberleyromo: I relocated it, it was already there. Personally I think it's useful -- if someone is like "wait why is it in the same file? i haven't seen that in other places!"

what we could do is remove the "suggestion" element, and just say you can do it multiple ways.

docs/tutorial/part-two/index.md
amberleyromo: We absolutely should -- I didn't want to do the screenshots until the content was sewn up (see top-level PR comment)

docs/tutorial/part-two/index.md
amberleyromo: I debated. I can go either way. I didn't think that was a productive rabbit hole at this juncture, but why not open the hatch.

docs/tutorial/part-two/index.md
amberleyromo: It's not that -- This has to do with the continuance from part one.

docs/tutorial/part-two/index.md
amberleyromo: So... I also thought that was the case, but I tested it, and it doesn't need a restart.

amberleyromo:
@jlengstorf I fully agree in theory. The question becomes, where is the optimal place to put it, then? My thinking is that a very simple example of installing and configuring typography.js actually makes a lot more sense now -- it's in the context of site styles, and is a toe-dip into plugins.

docs/tutorial/part-two/index.md
amberleyromo: Yes, we do. But the dev server from the part one could still be running. we're trying to avoid that. Really this note should be moved to where the new starter is cloned.

amberleyromo:
@jlengstorf alright, i ran through part two and three to make sure everything flowed. this is in a good spot.

PR #10932, opened 2019-01-08
chore(docs): add client data fetching to sidebar

amberleyromo:
@shannonbux: 

- adds client data fetching to docs sidebar
- some cleanup for sidebar text consistency

### DIFF
--- a/www/src/data/sidebars/doc-links.yaml
+++ b/www/src/data/sidebars/doc-links.yaml
@@ -296,6 +296,8 @@
       link: /docs/prpl-pattern/
     - title: Querying data with GraphQL
       link: /docs/querying-with-graphql/
+    - title: Client data fetching
+      link: /docs/client-data-fetching/
 - title: Behind the Scenes
   link: /docs/behind-the-scenes/
   items:
--- a/www/src/data/sidebars/doc-links.yaml
+++ b/www/src/data/sidebars/doc-links.yaml
@@ -223,23 +223,23 @@
     - title: Improving performance
       link: /docs/performance/
       items:
-        - title: Progressive web app (PWA)
+        - title: Supporting progressive web apps (PWAs)
           link: /docs/progressive-web-app/
         - title: Caching
           link: /docs/caching/
-        - title: Local HTTPS
+        - title: Using local HTTPS
           link: /docs/local-https/
-        - title: Audit with Lighthouse
+        - title: Auditing with lighthouse
           link: /docs/audit-with-lighthouse/
-        - title: Add a Manifest File
+        - title: Addding a manifest file
           link: /docs/add-a-manifest-file/
-        - title: Add Offline Support with a Service Worker
+        - title: Addding offline support with a service worker
           link: /docs/add-offline-support-with-a-service-worker
-        - title: Add Page Metadata
+        - title: Adding page metadata
           link: /docs/add-page-metadata/
-        - title: Search Engine Optimization (SEO)
+        - title: Optimizing for search engines (SEO)
           link: /docs/seo/
-        - title: Optimize Prefetching with Guess.js*
+        - title: Optimizing prefetching with Guess.js*
           link: /docs/optimize-prefetching-with-guessjs/
 - title: Ecosystem
   link: /ecosystem/
--- a/www/src/data/sidebars/doc-links.yaml
+++ b/www/src/data/sidebars/doc-links.yaml
@@ -34,24 +34,24 @@
           link: /docs/deploying-to-gitlab-pages/
         - title: Hosting on Netlify
           link: /docs/hosting-on-netlify
-        - title: Path Prefix
+        - title: Adding a path prefix
           link: /docs/path-prefix/
         - title: How Gatsby works with GitHub pages
           link: /docs/how-gatsby-works-with-github-pages/
     - title: Custom configuration
       link: /docs/customization/
       items:
-        - title: Babel.js
+        - title: Customizing Babel.js config
           link: /docs/babel/
-        - title: Babel plugin macros
+        - title: Using Babel.js plugin macros
           link: /docs/babel-plugin-macros/
-        - title: Custom webpack config
+        - title: Customizing webpack config
           link: /docs/add-custom-webpack-config/
         - title: Customizing html.js
           link: /docs/custom-html/
-        - title: Environment variables
+        - title: Using environment variables
           link: /docs/environment-variables/
-        - title: ESLint
+        - title: Using ESLint
           link: /docs/eslint/
         - title: Proxying API requests
           link: /docs/api-proxy/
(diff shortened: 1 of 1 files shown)


www/src/data/sidebars/doc-links.yaml
```diff
@@ -296,6 +294,8 @@
       link: /docs/prpl-pattern/
     - title: Querying data with GraphQL
       link: /docs/querying-with-graphql/
+    - title: Client data fetching
```
ZYSzys:
```suggestion
    - title: Fetching client data
```
IMHO, should this line also be consistent ?

www/src/data/sidebars/doc-links.yaml
```diff
@@ -209,10 +209,6 @@
           link: /docs/creating-a-sitemap/
         - title: Linking between pages
           link: /docs/linking-between-pages/
-        - title: Making your site accessible
-          link: /docs/making-your-site-accessible
-        - title: Routing
-          link: /docs/routing/
```
paulmelnikow: Was Routing removed deliberately?

www/src/data/sidebars/doc-links.yaml
```diff
@@ -34,24 +34,24 @@
...
         - title: How Gatsby works with GitHub pages
           link: /docs/how-gatsby-works-with-github-pages/
     - title: Custom configuration
       link: /docs/customization/
       items:
-        - title: Babel.js
+        - title: Customizing Babel.js config
           link: /docs/babel/
-        - title: Babel plugin macros
+        - title: Using Babel.js plugin macros
```
paulmelnikow: Think **Babel** is the correct styling, rather than **Babel.js**.

www/src/data/sidebars/doc-links.yaml
shannonbux: +1

www/src/data/sidebars/doc-links.yaml
```diff
@@ -222,26 +218,28 @@
...
+        - title: Using local HTTPS
           link: /docs/local-https/
-        - title: Audit with Lighthouse
+        - title: Auditing with lighthouse
           link: /docs/audit-with-lighthouse/
-        - title: Add a Manifest File
+        - title: Addding a manifest file
           link: /docs/add-a-manifest-file/
-        - title: Add Offline Support with a Service Worker
+        - title: Addding offline support with a service worker
```
shannonbux: delete one "d"

www/src/data/sidebars/doc-links.yaml
```diff
@@ -296,6 +294,8 @@
       link: /docs/prpl-pattern/
     - title: Querying data with GraphQL
       link: /docs/querying-with-graphql/
+    - title: Fetching client data
```
shannonbux: What is the more common way of saying this phrase? "fetching client data" or "client data fetching"? It isn't a huge deal, though I think it'd be nice if the title and URL match. Just don't know which one is more common usage. In a cursory google search, looks like "data fetching" is common. This would slightly break the visual pattern of having an "-ing" verb at the beginning of every guide, though technically, it would still grammatically match that pattern, because "client data fetching" is just a _kind_ of fetching. The "client" and "data" are modifiers to "fetching"...if that makes any sense at all!

www/src/data/sidebars/doc-links.yaml
```diff
@@ -144,21 +144,21 @@
     - title: Styling your site
       link: /docs/styling/
       items:
-        - title: Layout Components
+        - title: Using layout components
```
marcysutton: I would title case this and "Using Component CSS", otherwise this looks great! Love the active verb changes.

www/src/data/sidebars/doc-links.yaml
marcysutton: Should the subsequent words after 'Adding' be title cased? (consistent with my comment above)

marcysutton:
> [Verbs] starting with "using" or "adding" since those verbs aren't super interesting to me and seem to lengthen the title without making it that much more specific

Overall I really like the title changes to be more active. I agree with Shannon that it can get a little repetitive, yet consistency is nice...maybe it could be aided with a thesaurus (something I do frequently when I find I'm repeating myself). I think the more important thing, though, is that the titles are less vague so the reader is led on more of a journey and not assumed to know what something like "Babel macros" are...

www/src/data/sidebars/doc-links.yaml
amberleyromo: I would tend to agree, and this was my original change, but it's wholly inconsistent with the rest of the sidebar https://github.com/gatsbyjs/gatsby/pull/10932#discussion_r247081938

amberleyromo:
@marcysutton 	please feel free to take this over and push/merge whatever changes you want here! 👌

marcysutton:
@amberleyromo I realize my comment wasn't super actionable....I'll think about overall sidebar format some more for future changes. But I personally think this PR is already useful and we should go ahead with it. I see two unresolved things at the moment:
- Was `/docs/routing/` an intentional removal?
- Some extra `dd`s and title casing stuff. I'm happy to fix these and add a commit!

amberleyromo:
@marcysutton `/docs/routing/` was an intentional removal, but after looking at it i'm not sure why I didn't move the sub-items? And now I'm further re-evaluating.

The idea was (I think there was a convo on Slack) that the "linking between pages" doc is essentially the "routing" doc. The stemmed from the question "what would we want to be written in the "routing" stub"?

So, I would either leave the top-level routing item, and flesh out the stub (imo no top-level docs should be left as stubs), or move the current routing sub-items under "linking between pages".

Argument for leaving both: depending on the audience, "linking between pages" and "routing" are probably separately useful entry points. (beginners vs more experienced folks).

tl;dr -- up to you!

www/src/data/sidebars/doc-links.yaml
shannonbux: Hm yeah. I like the original change, not sure who has the final call here. Probably @marcysutton!

marcysutton:
How about moving the routing links under "Linking Between Pages"? That way the subitems can go underneath it, including `@reach/router in Gatsby*` which will provide the routing signal to more advanced users? I can take on a @reach/router doc when I get back from vacation.

docs links open to Linking Between Pages

LekoArts:
Hey @marcysutton 😊 
Friendly reminder, what's the status on this? Thanks for looking into it!

marcysutton:
Gah I totally dropped the ball on this one–I had made changes locally and never resolved this PR. Now I get to deal with some major doc-links conflicts! 😜 Working on those now so we can close this out. Thanks for the reminder!

marcysutton:
ok, so I tried pushing to the fork and access was denied...so I'm going to close this PR and open a new one. Thanks for getting it started, @amberleyromo!

PR #10933, opened 2019-01-08
[WIP]: feat(gatsby): add support for asset prefixing

### LINKED ISSUE #2335: Ability to deploy non-html files to separate domain
Doing some search, there was an initial issue which was closed by the OP: #1369.

However, the work I'm presently doing means the gatsby page will sit behind a company wide domain with the assets ideally sitting behind our static domain. Think... `mywebiste.com/gatsby-page` should reference the non-html files that sit at `mystaticcontent.com`. (so something like `mystaticcontent.com/gatsby-page/js-file.js`).

I understand that from your point of view it should all just sit in 1 domain, but we have a very large website with a lot of applications and routing being fronted by a load balancer.

At present, it isn't feasible as well because the URL we want to use for the present page we have build is a top level URL, such as `mywebsite.com/gatsby-page`, but we can't really serve the assets from the top level domain as well.

We did look at the gatsby documentation which has path prefix (so not allowing us to put it in a different domain altogether, but at least trialing putting the assets somewhere else), which puts the whole project under a certain path, but this also does it for react router. We added two rules into our load balancer, one for the URL we want the HTML page to resolve to, and another for the path prefix resolving to the public folder. But when viewing the page in the browser, react-router kicks in and doesn't understand the URL and send you off to the 404 page defined with gatsby.

We also looked into using modifyWebpackConfig and setting the public path just for webpack, but this doesn't seem to be picked up as part of `headComponents` in the html.jsx file (as it is still setting the link refs to `/some-js-file.js`.

### LINKED ISSUE #8833: Question: Putting Assets in a Different Folder
## Summary

I'm a growth engineer @DataCamp, and I'm running an experiment that involves rewriting our home page in Gatsby to measure gains in performance and conversion metrics.

We plan to reverse proxy the homepage on datacamp.com to split 50% of traffic between our current application and our new Gatsby application.

The problem we're running into is that every JS, CSS, and other file that Gatsby generates would also need to be reverse proxied in order for this to work.

So far as I've been able to see, Gatsby does not provide a mechanism to store such files in a separate directory (which would make it easy to reverse proxy all of the traffic there).

Does Gatsby (or a plugin) provide a way to store all of these files in a single directory?

Alternatively, does Gatsby provide a way to prefix calls to these assets if they are being hosted elsewhere (i.e. on a separate subdomain)?

## Relevant information

Potentially relevant past issues, PRs, and plugins: #6346, https://github.com/madecomfy/gatsby-plugin-asset-path

### LINKED ISSUE #9513: Limit `pathPrefix` changes to assets only
## Summary
- We're passing the `pathPrefix` option primarily to access assets at the specified location on our server. The only issue is that this will also redirect to the path-prefixed version of the URL, which is not what we want. 

## Relevant information
- Basically, we want to serve up all assets from the `pathPrefix`ed version of the site, but not have it affect the routing. For example (assuming the `pathPrefix` were set to `/assets`):
  - User navigates to `www.example.com/page-1`
  - Internal routing (?) updates the URL to `www.example.com/assets/page-1` for all requested routes.
- I tried a couple different options, such as just overwriting the webpack config using `gatsby-node.js` to make the `publicPath` match the prefix, but I'm getting strange behavior there.
- I think I've narrowed the issue down to the `basepath` variable for the Reach Router, which I need to have different behavior depending on the environment.
- Our site uses standard Gatsby structure with all needed routes included in the `/pages` directory.

- Is there a way to update the `basepath` variable using gatsby config?
- Is there any other method for adding a prefix to just assets but maintaining the internal routing?
- Is there a way to create a custom `Router` to control that functionality directly?

Let me know if additional information would be helpful. It's mostly just general questions that I haven't been able to track down any solutions for yet. Thanks!

DSchau:
## Description

Alright - so this small (but currently manageable) PR is a somewhat WIP PR for asset prefixing.

### As a developer _using_ this functionality

The general functionality as a developer implementing this functionality would be something like the below

1. The developer would update `gatsby-config.js` to add an `assetPrefix` (and possibly `pathPrefix`--they work together) property, e.g.
    ```js
    module.exports = {
      assetPrefix: 'https://cdn.dustinschau.com',
      // the rest
    }
    ```
1. The developer would use the (existing) flag `--prefix-paths` with `gatsby build`
    - Is this confusing? I don't think so, but... maybe.
1. The developer would implement the `onPostBuild` API in `gatsby-node.js` like so:
    ```js
    exports.onPostBuild = async function onPostBuild({ assets }) {
      await Promise.all(
        assets.map(asset => someFunctionToUploadAsset(asset))
      )
      // or perhaps you just move these assets to top level assets folder and upload from there as part of some CLI (we're not oppinionated here)
    }
    ```

In theory, all of this will _just work_. All assets (that Webpack + Gatsby) know about are tracked in the `assets` array passed to `onPostBuild`. These are appropriately prefixed because we've joined pathPrefix + assetPrefix into one unified concept, so there shouldn't be that much additional complexity added. As a developer adding this functionality, you would need to:

### As a developer _tapping into_ asset path / path prefix

One of the struggles we've had with these types of changes in the past is that we increase the surface area and we introduce subtle bugs because plugin authors forget to tap into this functionality. This isn't necessarily solved in this PR--but I do think we can improve upon this and make this more maintainable. Specifically, I introduce a few new API methods we pass to plugins that I think is easier to use, specifically `withPathPrefix` (this will be used by anything implementing )

Rather than manually joining a path like the below:

```js
const prefixedSrc =
    options.pathPrefix + `/static/${digestDirPrefix}` + encodedImgSrc
```

we will use the exposed `withAssetPrefix` helper (because this is an asset) like so:

```js
const prefixedSrc = withAssetPrefix(`/static`, digestDirPrefix, encodedImageSrc)
```

this will be automatically bound to the assetPrefix (and pathPrefix, if specified) so this _should_ be less fragile in the future. We just need to document this appropriately.

## Things Left to Validate

- [ ] Find areas where we may need browser vs. Node path strings (e.g. path.join vs. joining on `/` character)
  - This may lead to a slightly different API
- [x] Validate that it makes sense to wipe out `__PATH_PREFIX__` with `${assetPrefix}${pathPrefix}`
- [x] Validate I haven't hopelessly broken path prefix (e2e tests 👍)
- [ ] Write e2e test validating this functionality (😅)
- [ ] This doesn't address dynamic pathing (e.g. `__webpack_public_path__` global). Should it?
- [ ] (Intentionally) I haven't added this API to any plugins yet. I want to introduce the API then implement the API _in_ plugins
  - Beyond bumping peerDep, we should think of ways to introduce this with as little churn as possible
- [ ] Update documentation for this new functionality (this will actually be a little bit challenging to document, e.g. when to use one vs. the other)

## Related Issues

- (Possibly) fixes #9513 (you just wouldn't use pathPrefix)
- (Possibly) fixes #8833 (you would just copy to a folder and use pathPrefix or assetPrefix)
- Fixes #2335
  - The assets are still built into the same folder (`/public`), but now there's a way to change wherever those assets eventually end up

cc @xjamundx


packages/gatsby/src/redux/reducers/config.js
```diff
@@ -40,6 +40,12 @@ module.exports = (state = {}, action) => {
         action = _.set(action, [`payload`, `pathPrefix`], ``)
       }
 
+      // If pathPrefix isn't set, set it to an empty string
+      // to avoid it showing up as undefined elsewhere.
+      if (!_.has(action, [`payload`, `assetPrefix`])) {
```
DSchau: I don't really think this is needed, but will validate.

packages/gatsby/src/utils/__tests__/__snapshots__/asset-path-registry.js.snap
```diff
@@ -0,0 +1,95 @@
...
+  "42-d5c0588bf92953044654.js.map",
+  "component---src-pages-showcase-js-710ac9cbe8f04ccc5629.js",
+  "component---src-pages-showcase-js-710ac9cbe8f04ccc5629.js.map",
+  "39-72be356c49b30af6ec24.js",
+  "39-72be356c49b30af6ec24.js.map",
+  "component---src-pages-starters-js-5cea20767c8766bdcda4.js",
+  "component---src-pages-starters-js-5cea20767c8766bdcda4.js.map",
+  "component---src-pages-ecosystem-js-633f4f4caa8ef50afcbe.js",
+  "component---src-pages-ecosystem-js-633f4f4caa8ef50afcbe.js.map",
+  "component---src-pages-index-js-0a2a452fc816778a6329.js",
```
DSchau: This is currently _only_ the assets webpack knows about. It will be _all_ assets registered with `withAssetPrefix` once plugins implement this API.

packages/gatsby/src/utils/api-runner-node.js
```diff
@@ -98,6 +106,7 @@ const runAPI = (plugin, api, args) => {
       {
         ...args,
         pathPrefix,
+        assetPrefix,
```
DSchau: This is provided merely for convenience; I think the `withAssetPrefix` and `withPathPrefix` should be used instead.

packages/gatsby/src/utils/path.js
```diff
@@ -10,6 +12,34 @@ export function joinPath(...paths) {
...
+function withBasePath(basePath) {
   return (...paths) => joinPath(basePath, ...paths)
 }
+
+function normalizePath(...parts) {
+  const normalized = parts
+    .filter(part => part && part.length > 0)
+    .map(part => part.replace(/^\/+/, ``).replace(/\/+$/, ``))
+
+  return ((parts[0] || ``).match(/^http/) ? [] : [``])
```
DSchau: jank.

DSchau:
All the tests are failing 😬 Will address!

Daniel15:
> I haven't done much of any manual testing on this yet, primarily because I don't have an easy way to set up some type of CDN/host where I can test this out

No need to actually use a real CDN... You could just configure your preferred web server (Nginx, Apache, whatever) with two separate sites 👍

DSchau:
@Daniel15 good point! I could probably just create a repo with the static assets too (e.g. a separate Netlify domain!).

I'll check that out now.

Daniel15:
Yeah, a separate Netlify or Github Pages site would work too.

DSchau:
Progress!

Here's an example _sorta_ working hosted on Netlify.

The HTML assets are hosted on https://gatsby-asset-prefix.netlify.com, whereas the other assets (currently JavaScript, CSS, etc.) are hosted on https://gatsby-asset-prefix-cdn.netlify.com.

Some issues to iron out--namely that pathPrefix can't be used _always_ so we need a way to expose both and use appropriately.

DSchau:
Have to fix a number of things (e.g. important things like _tests_) but it's working, including with images!

https://gatsby-asset-prefix.netlify.com/

KyleAMathews:
Woot!

DSchau:
No idea why the CircleCI server seems to be down/not running. Checking it out.

packages/gatsby/src/joi-schemas/joi.js
```diff
@@ -4,6 +4,8 @@ export const gatsbyConfigSchema = Joi.object().keys({
   __experimentalThemes: Joi.array(),
   polyfill: Joi.boolean(),
   siteMetadata: Joi.object(),
+  // TODO: validate as URL
```
DSchau: TODONE

DSchau:
I'd like to circle back to this tomorrow. Beyond the checklist in the first issue, there is at least one outstanding issue.

How do we introduce this breaking change? This introduces a _new_ API in Gatsby core which is used by plugins. We do currently have a mechanism in Gatsby core, where if a plugin specifies a peerDep on gatsby _greater than_ the current version, we throw a warning to the console (but do not fail).

So we could introduce this in two parts:

- In this PR, introduce the necessary core APIs
- Once merged, publish this (probably as a minor release?)
- In a separate PR (or PRs), _use_ these exposed APIs and bump the peerDep to a version of Gatsby that contains the APIs
  - We could major bump all of these plugins since it's technically breaking, too

I do worry this could cause some churn, so I want to think if there are ways we could avoid this and ensure users have the best possible user experience with this new feature.

More to think about!

KyleAMathews:
Could the plugins check for the existence of the new API and only use it then? We obviously want to avoid major version bumps whenever possible.

DSchau:
Not sure if it's a new change or not, but the e2e test failures seem to be failing for a _real_ reason. Debugging.

GuillaumeAmat:
Hello there, do you know how much work/time that feature needs to be done?

cc @DSchau

trickydisco78:
This is great. Good work. Will try testing it myself

DSchau:
@trickydisco78 some things I've noticed (or intentionally broken!):

- I've broken up the image stuff (gatsby-plugin-sharp / gatsby-transformer-sharp) changes into a different PR
  - This feature is a minor bump for Gatsby, so I want to get the version of gatsby and _then_ update plugins peerDep so they can get the changes
- Seems like there were some offline bugs, will investigate further

Otherwise, this seems to work pretty well! See https://gatsby-asset-prefix.netlify.com/ for a demo, and in particular, the repo will probably be helpful: https://github.com/dschau/gatsby-asset-prefix

gniquil:
@DSchau I have a question regarding how to use this.

Currently many generated fragments for lazy loading and etc. are directly placed in the root of the public folder. Presumably, since they have a unique hash, they can be cached for 1-year (set in Cache-Control header). However at the same time, any files inside the "static" folder are also merged into the public folder, and presumably we don't want these to be cached forever. So now, there's no simple way to tell which files inside public can be cached or not. 

Thus Is it possible to "consolidate" all assets to a single directory (that we can optionally change) so that we can have a simple rule (in our proxy/CDN/s3 upload) for cache control? 

Let me know if I should open up a separate issue. However, I would imagine this to be pretty important for those who will be serving their assets through CDN?

DSchau:
@gniquil pretty tangential - would you mind opening an issue for that?

DSchau:
👋 everyone! Went a different approach on this one, so closing it out.

Please feel free to follow along at #12128.

PR #10940, opened 2019-01-09
test(production-e2e): Add offline tests

davidbailey00:
Changes:

- Introduce `CYPRESS_SUPPORT` env variable for whether to enable Cypress API hooks, and when running tests, set this variable
- Use `TEST_PLUGIN_OFFLINE` env variable to determine whether or not to install the offline plugin when testing
- Add some additional tests which check the SW is installed correctly and perform a few things which I'd been manually testing
  - Existing `production-runtime` tests also now run both without and with `gatsby-plugin-offline`
- Update CircleCI config to allow running custom test commands
- Disable `chromeWebSecurity` for `production-runtime` tests because it doesn't work with service workers :(
- Delete some Cypress files accidentally added in the wrong place


davidbailey00:
Gatsbygram tests failed due to a segfault, not due to code changes here

Issue #10747, opened 2019-01-01
BUG: gatsby-plugin-sharp does not respect forced format depending on input format


polarathene:
## Description

I've discovered a bug which appears to be due to the `sharp` package, [raised an issue there](https://github.com/lovell/sharp/issues/1528), this is more of a tracker to make sure `gatsby-plugin-sharp` gets the update when fixed(I think `"sharp": "^0.21.0"` in the package json will ensure any fixes are made available, eg `0.21.02`?) .

If you input a JPEG image and use `toFormat: PNG` in graphql, it will be ignored as the chaining approach presently has a bug where the input formats mimetype will override any prior formats in the pipeline chain, even if they have `force: true` in their format options set.

### Steps to reproduce

Use the default starter with a jpg image instead, try force/convert to png with `toFormat: PNG`, notice that with fixed/fluid with `gatsby-image` results in png format like desired in the `<picture>` element, but the `base64` placeholder is actually jpeg.

### Expected result

The base64 should be png format.

### Actual result

As the pipeline called `.jpeg()` after `.png()` and the input type is `jpeg`, it ignored `.png()`'s `force: true` option, resulting in a base64 in jpeg format.

Rearranging the pipeline order provides the expected result(but then breaks it another way due to actual bug).

### Environment

Gatsby 2.0.78 in an Alpine Docker container.

polarathene:
Fixed in sharp 0.21.02

PR #10944, opened 2019-01-09
feat(gatsby-image): Placeholder Improvements

### LINKED ISSUE #7025: gatsby-remark-images-contentful isn't doing the "blur up" effect, just fades in
## Description

As per the `README.md` on [gatsby-remark-images-contentful](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-remark-images-contentful) it states: 

> Using the "blur up" technique popularized by Medium and Facebook where a small 20px wide version of the image is shown as a placeholder until the actual image is downloaded.

However, when I try to use it, it just fades in the image 🤷🏼‍♂️

### Steps to reproduce

1. Install https://github.com/ryanwiemer/gatsby-starter-gcn
2. Change `heroImage` part of the query inside `src/templates/post.js` to:

```
heroImage {
  title
  sizes(maxWidth: 1800) {
    # Removing the '_noBase64' adds the 20x20 placeholder for blur up
    ...GatsbyContentfulSizes_withWebp 
  }
  ogimg: resize(width: 1800) {  
    src
    width
    height
  }
}
```

### Expected result

When the post page is loading, the blur up effect should take place on the hero image

### Actual result

The hero image still just fades in.

### Environment

```

  System:
    OS: macOS 10.14
    CPU: x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 8.9.3 - ~/.nvm/versions/node/v8.9.3/bin/node
    Yarn: 1.9.4 - /usr/local/bin/yarn
    npm: 5.5.1 - ~/.nvm/versions/node/v8.9.3/bin/npm
  Browsers:
    Chrome: 67.0.3396.99
    Safari: 12.0
  npmPackages:
    gatsby: ^1.9.248 => 1.9.260 
    gatsby-image: ^1.0.45 => 1.0.51 
    gatsby-link: ^1.6.40 => 1.6.44 
    gatsby-plugin-canonical-urls: ^1.0.16 => 1.0.18 
    gatsby-plugin-feed: ^1.3.21 => 1.3.22 
    gatsby-plugin-manifest: ^1.0.21 => 1.0.21 
    gatsby-plugin-netlify: ^1.0.19 => 1.0.19 
    gatsby-plugin-nprogress: ^1.0.14 => 1.0.14 
    gatsby-plugin-offline: ^1.0.16 => 1.0.16 
    gatsby-plugin-react-helmet: ^2.0.10 => 2.0.11 
    gatsby-plugin-segment: ^1.0.2 => 1.0.2 
    gatsby-plugin-sitemap: ^1.2.23 => 1.2.23 
    gatsby-plugin-styled-components: ^2.0.11 => 2.0.11 
    gatsby-remark-images-contentful: ^1.0.2 => 1.0.2 
    gatsby-remark-prismjs: ^1.2.24 => 1.2.24 
    gatsby-source-contentful: ^1.3.50 => 1.3.51 
    gatsby-source-filesystem: ^1.5.29 => 1.5.35 
    gatsby-transformer-remark: ^1.7.39 => 1.7.41 
  npmGlobalPackages:
    gatsby-cli: 1.1.58
```

`gatsby-config.js`: N/A 
`package.json`: N/A 
`gatsby-node.js`: N/A 
`gatsby-browser.js`: N/A 
`gatsby-ssr.js`: N/A

### LINKED ISSUE #9857: [gatsby-plugin-sharp] custom base64 width and height of 1px to get 'average color' placeholder
## Summary

Hello, I am using `gatsby-image` to display images from a WordPress source. I am trying to manipulate the base64 placeholder to display an 'average' color of the source image.

## Relevant information

My first idea was to request a base64 with a width and height of 1px, so that the image library reduces it to 1px of the average color. I think, at this time, there is no way to specify the size of the base64 generated when calling a fluid or fixed image fragment. It looks like it is set to a [hardcoded 20px](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-sharp/src/index.js#L642-L643).

I found a workaround by requesting a image of 1px separate from my fluid/fixed image query.

My graphql query looks like this :

```
...
edges {
  node {
    image {
      localFile {
        childImageSharp {
          placeholder: fixed(width: 1, height: 1) {
            ...GatsbyImageSharpFixed_noBase64
          }
          fluid: fluid(maxWidth: 800, quality: 100) {
            ...GatsbyImageSharpFluid_noBase64
          }
        }
      }
    }
  }
}
```

And when I call `gatsby-image` `` element, I add the src image of my 'placeholder' as the 'base64' key of my fluid image, like this :

```
  // Short for the fluid image
  const fluid = node.image.localFile.childImageSharp.fluid

  // If placeholder object exist, add its src to fluid's base64 key
  if (node.image.localFile.childImageSharp.placeholder) {
    fluid.base64 = node.image.localFile.childImageSharp.placeholder.src
    // will look like : fluid.base64 = '/static/hash__.jpg'
  }

  // gatsby-image use the 1px src as the placeholder
  return (
    <Img fluid={fluid}/>
  )
```

It works fine, but I would much prefer to have the option to specify parameters for the base64 to my fluid or fixed image fragemnt.

I am writing here in case I might be wrong and the option is already there. If not, and if the idea of having 'average color' placeholder sounds interesting or simply have the option of custom width or height, I would love to try and submit a PR.

Thanks!

### LINKED ISSUE #10705: Feature: Configurable width/size for base64 placeholder with webp support and backgroundColor fallback
## Summary

Configurable width/size for placeholder image in `gatsby-image`.

Webp support for base64 placeholder image element, and compatibility with the backgroundColor element to use as fallback for those that don't have Webp(due to inlining base64, less benefit to support multiple image formats, 2019 a good portion of users should have webp access).

### Basic example

A global option in gatsby-config.js for width/size and maybe a prop for the image component for overriding? Nothing else really changes API wise, you just get the ability to also use webp format for the base64 placeholder image, and compatibility/fallback with backgroundColor element.

### Motivation

Configurable width/size for placeholder image in `gatsby-image`. 20px vs 40px can make quite a bit of difference in the preview aesthetic. Facebook if I'm not mistaken uses around 40px and Medium 30px.

I'd also like this base64 inlined data to be webp, I notice it's jpeg(and I think can also be png if input is png?). I take it webp is not supported there as the placeholder does not use the `<picture>` element? Firefox presently doesn't have support for webp, but it's current beta and nightly both do so it's users will be able to benefit from webp as well from 2019. There's also a polyfill now.

To better handle users who browsers do not support webp, I would like to also use the backgroundColor placeholder as a fallback. Presently this is not possible as the backgroundColor element is after/above the image placeholder element... An alternative would be to use background-color style on the image element? The element order could be rearranged, but the styles conflict a little, image element has a 500ms opacity transition while the backgroundColor element has a 350ms transition delay(no transition duration on opacity, just a delay before it switches).

Custom styles could be passed in to resolve the placeholder issues, but I think some/all of those could be addressed by updating the source directly?

I'd like to contribute all of the above features/changes if anyone else thinks they'd be value in doing so.

polarathene:
3 features: 

- Support base64 widths other than `20px`. Closes #9857
- Support forcing base64 image format.
- Support base64 with WEBP format.

1 fix:

- Swap the order of `backgroundColor` and `base64` placeholder elements allowing both to work together. Resolves #7025 

2 refactors:

- Replace some magic numbers with variables for clarity.
- Share a common variable for placeholder transition delays(and fix typo?).

~~Forcing formats(base64 or current support) [doesn't work reliably currently](https://github.com/gatsbyjs/gatsby/issues/10747). Bug has been fixed in `sharp` and available when the `0.21.2` release is published.~~ Available!

## Related Issues

Closes #10705 (My own issue requesting these changes and feedback to proceed)


polarathene:
Could this PR receive some feedback/review please?

polarathene:
Would a demo project to visually see the changes help at all? I could set one up?

I'd appreciate receiving feedback on this and getting a decision to merge or reject as the longer this PR stays unattended the more conflicts I'll have to resolve(which I'd rather not).

packages/gatsby-image/src/index.js
```diff
@@ -295,13 +286,23 @@ class Image extends React.Component {
...
-                transitionDelay: `0.35s`,
+                transitionDelay: initialDelay,
                 right: 0,
                 left: 0,
               }}
             />
           )}
 
+          {/* Show the blurry base64 image. */}
+          {image.base64 && (
```
wardpeet: what's the reason for moving it down?

packages/gatsby-image/src/index.js
```diff
@@ -295,13 +286,23 @@ class Image extends React.Component {
                 top: 0,
                 bottom: 0,
                 opacity: !this.state.imgLoaded ? 1 : 0,
-                transitionDelay: `0.35s`,
+                transitionDelay: initialDelay,
```
wardpeet: this used to be `0.35s` now it's `0.25s` unsure if it matters (I don't think) I'll check it out.

packages/gatsby-image/src/index.js
```diff
@@ -379,12 +370,22 @@ class Image extends React.Component {
...
                 width: image.width,
                 opacity: !this.state.imgLoaded ? 1 : 0,
-                transitionDelay: `0.25s`,
+                transitionDelay: initialDelay,
                 height: image.height,
               }}
             />
           )}
 
+          {/* Show the blurry base64 image. */}
```
wardpeet: what's the reason for moving it down?

packages/gatsby-plugin-sharp/src/index.js
```diff
@@ -570,9 +585,10 @@ async function fluid({ file, args = {}, reporter, cache }) {
   }
 
   const { width, height, density, format } = metadata
+  const density_72PPI = 72 // Standard digital image pixel density
```
wardpeet:
```suggestion
  const defaultImagePPI = 72 // Standard digital image pixel density
```

packages/gatsby-plugin-sharp/src/index.js
```diff
@@ -570,9 +585,10 @@ async function fluid({ file, args = {}, reporter, cache }) {
...
   const pixelRatio =
     options.sizeByPixelDensity && typeof density === `number` && density > 0
-      ? density / 72
+      ? density / density_72PPI
```
wardpeet:
```suggestion
      ? density / defaultImagePPI
```

packages/gatsby-plugin-sharp/src/index.js
```diff
@@ -672,13 +688,14 @@ async function fluid({ file, args = {}, reporter, cache }) {
     })
   })
 
-  const base64Width = 20
+  const base64Width = options.base64Width || defaultBase64Width()
```
wardpeet: can't you just use `options.base64Width`? It should be always set because of the healOptions function

wardpeet:
* Support base64 with WEBP format.

I'm not 100% sold on this one but the user decides so I don't see a big issue with it

packages/gatsby-image/src/index.js
polarathene: Check the commit history, there are explanations :) In this case, I noted that there was a discrepancy which was also causing a visual transition bug for me in a slider component. I assumed this was probably a typo and that the delays are meant to be in sync?

packages/gatsby-image/src/index.js
polarathene: As the PR and commits should explain, it was to fix the layer ordering with `backgroundColor` placeholder, so that it can be used as a fallback with an image placeholder. Presently the image would be covered by the `backgroundColor` placeholder.

packages/gatsby-image/src/index.js
polarathene: As the PR and commits should explain, it was to fix the layer ordering with `backgroundColor` placeholder, so that it can be used as a fallback with an image placeholder. Presently the image would be covered by the `backgroundColor` placeholder.

packages/gatsby-plugin-sharp/src/index.js
polarathene: I am pretty sure I required it here but it's been a while since I wrote this so I'm not sure what situation required it.(**EDIT:** Answered)

Looking through the commits, this const was originally hardcoded with a value of 20. I then had `options.base64Width || 20` before moving the default value into a variable, and finally a function(to support a config option value in `gatsby.config.js`, otherwise falling back to the default value of 20).

`options.base64Width` is an optional param that can be set on the node/graphql function, should you want to use a different base64 width than the default value of 20. If you want to override the default globally via `gatsby.config.js`, `defaultBase64Width()` will return that value instead of 20.

At line 806, `base64Args` for the `fixed()` variant, `width` param is replaced with the `options.base64Width` if provided.

> It should be always set because of the healOptions function

Went over the code, the `fluid()` and `fixed()` functions initialize the `options` object at the start with no default params passed in. Only `generateBase64()` does this, where it originally passed in the hard coded default width of 20, my PR replaced that with `defaultBase64Width()` accordingly. That method is called by both fluid and fixed methods when they call `base64()` which first checks for a cached base64 data else it creates a new base64 image assigning the default width.

Fluid needs to know the width in advance for computing the height, this was the case prior to my PR as well.

Fixed derives it's height from the width during generation of base64 image afaik, I added a width param to it's base64 args only if a custom width was requested, otherwise it's provided a default width during base64 generation as mentioned above with `generateBase64`.

---

**TL;DR:** 
No, `options.base64Width` isn't handled by `healOptions()`, it's passed in from a param in the graphql query. 

For `fluid()` the width must be known to derive the `base64Height`, prior to this PR it was a fixed value, 20; now it can also be a configurable global value in `gatsby.config.js` too.

For `fixed()`, it's only interested in the graphql query param if available, otherwise it gets a default width from the 2nd `healOptions()` at a later stage in `generateBase64()`(assuming no cached version exists).

---

Should `options.base64Width` be changed to `args.base64Width` to avoid this confusion with `healOptions()`? Should any of what I've said be added as comments to document what is going on within the code more clearly?

polarathene:
> No one likes to touch plugin-sharp so I think that's why we were so hesitant to review this one. 

Might help to split into multiple files? It's getting a bit large in size.  

I can understand the hesitance, I just spent a fair bit of time trying to answer one of your review questions and had to grok the flow of logic again to understand my small changes and if I misunderstood anything based on that question.

> I'm not 100% sold on this one but the user decides so I don't see a big issue with it

It's not default iirc, Chrome and Firefox(since earlier this year) have WEBP support now and this PR allows for using `backgroundColor` as a fallback for browsers that don't. WEBP allows for smaller file sizes or squeezing in more pixels with larger base64 placeholders, I figure that'd be a nice addition to support as opt-in :)

---

Where to from here? If this PR looks good to merge, I'll fix the conflicts and look into updating related docs?

packages/gatsby-image/src/index.js
polarathene: The `transitionDelay` of `0.35s` was added by @KyleAMathews some time ago, via [this commit](https://github.com/gatsbyjs/gatsby/commit/76cda50d75dd9ea8950b74b8ebc97bc0380d2250) according to git blame. It was batched into the same commit as many other changes, so it's purpose isn't clear.

I think it's safe to share the `0.25s` value placeholder uses for transition delay?

polarathene: Investigating #12254 and working on a PR for that, I've since come to believe the `0,25s` value is never actually used now, but `bgcolor` should use `0,5s`. `initialDelay` loses it's purpose, but could be renamed to the duration/delay these components share(all for the shared fade-in transition).

polarathene:
> Works and this might fix the bad blur issues as well as you can now choose your base64 width & height.

@wardpeet Just to clarify, this PR only lets you customize the base64 width(height is still inferred).

Thanks for taking the time to review and merge :)

PR #10945, opened 2019-01-09
feat(gatsby): Add noscript tag to default-html

LekoArts:
Since Google's lighthouse check complains about not having a `<noscript>` tag outside the app shell (because with service workers it doesn't turn them off...) I added this styleable noscript tag + a little note on service workers. Feedback on the note is welcome.

The note should convey the message that Gatsby also works without JavaScript but is better with it enabled + the service worker bug (while having JS disabled)

Also fixed the headings of the doc page


packages/gatsby/cache-dir/default-html.js
```diff
@@ -16,6 +16,11 @@ export default class HTML extends React.Component {
         </head>
         <body {...this.props.bodyAttributes}>
           {this.props.preBodyComponents}
+          <noscript key={`noscript`} id={`gatsby-noscript`}>
+            This site works best with JavaScript enabled. If this site uses
```
DSchau: Let's _just_ do this:

```
This site works best with JavaScript enabled.
```

The second note sorta indicates that we _expect_ issues and while we've seen issues, that's not what we want to convey!

LekoArts: Alright! Will also update the snapshots

LekoArts: Done

DSchau:
Looks like this works just dandy 🎉 

There is one consideration here... for sites that don't register an offline plugin, the content will load (as expected) but they're greeted with the following message. I think this is OK, but just wanted to raise in case we have any ideas for improvement here.

relaxed-mcclintock-f34bbd netlify com_
screen shot 2019-01-09 at 10 32 09 am
screen shot 2019-01-09 at 10 34 21 am

LekoArts:
> I think this is OK, but just wanted to raise in case we have any ideas for improvement here.

I think this is fine though. With the `id` people can style (or even hide) it 👍

LekoArts:
I personally think it's okay. I'd just assume that people see the id and then think of styling it 🙈

DSchau:
Successfully published:
 - gatsby@2.0.90

kripod:
Being able to localize (or even disable) the string would be a nice little addition for international users.

PR #10955, opened 2019-01-09
fix: update calls to sourceNodes

jlengstorf:
- we still had v1 syntax in here
- `createNodeId` was destructured improperly
- removed unused var

### DIFF
--- a/packages/gatsby-source-shopify/src/gatsby-node.js
+++ b/packages/gatsby-source-shopify/src/gatsby-node.js
@@ -23,11 +23,7 @@ import {
 } from "./queries"
 
 export const sourceNodes = async (
-  {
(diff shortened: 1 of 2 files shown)


jlengstorf:
@nikhilag could we get you to deprecate the forked `gatsby-source-shopify2` with a message pointing back to this package? I think there's some confusion around this plugin, and — hopefully — this PR brings us to a stable enough place to officially point everyone to this package.

Let me know if there's anything from your fork that we're missing in this one.

Thanks!

jlengstorf:
@DSchau seems to be working properly. I'll get a PR up for the store to verify.

nikhilag:
@jlengstorf Sounds like a good idea after this change! Will deprecate my plugin today.

PR #10956, opened 2019-01-09
[gatsby-docs] Use correct (absolute) link

gurpreet-hanjra:
Add an absolute link to the tutorial. Currently, it re-directs to a relative path /tutorial/ to 404 on Github. Replaced it with https://www.gatsbyjs.org/tutorial/


gurpreet-hanjra:
Thank you so much @DSchau Makes sense!

PR #10960, opened 2019-01-10
feat(gatsby-plugin-page-creator): add validatePath and createPath options

leonardodino:
## Description

adds custom validation and url creation to `gatsby-plugin-page-creator`.

allows for filtering files (eg: drafts) and changing url (eg: nesting or translating) while keeping the filesystem clean.

#### example added in `README.md`

```javascript
{
  resolve: `gatsby-plugin-page-creator`,
  options: {
    path: `${__dirname}/src/articles`,
    createPath: (basePath, filePath) => {
      // Creates a custom URL based on filename
      return `/articles/${slugify(basename(filePath))}/`
    },
    validatePath: (relativePath, isValid) => {
      // Filter page creation based on criteria,
      // the second argument is the default validatePath function
      if (basename(relativePath).startsWith('draft-')) return false
      return isValid(relativePath)
    },
  },
},
```

### DIFF
--- a/packages/gatsby-plugin-page-creator/README.md
+++ b/packages/gatsby-plugin-page-creator/README.md
@@ -43,6 +43,25 @@ module.exports = {
         path: `${__dirname}/src/settings/pages`,
       },
     },
+    // You can also provide "createPath" and "validatePath" functions
+    // to override default behaviour, if needed
+    {
+      resolve: `gatsby-plugin-page-creator`,
+      options: {
+        path: `${__dirname}/src/articles`,
+        createPath: (basePath, filePath) => {
+          // Creates a custom URL based on filename
+          return `/articles/${slugify(basename(filePath))}/`
+        },
+        validatePath: (relativePath, isValid) => {
+          // Filter page creation based on criteria,
+          // the second argument is the default validatePath function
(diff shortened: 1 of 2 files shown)


KyleAMathews:
Hey! Sorry for not reviewing this until now! 😅We somehow missed it.

I'm not sure however that we want to do this. This plugin is intentionally pretty simple. If people want custom logic for page creation — we generally suggest people use custom logic in their gatsby-node.js.

What we _could_ do though is expose the watching logic from gatsby-page-creator as a logic so you could use this in your gatsby-node.js perhaps like this:

```js
const { watchDirectoryForPages } = require(`gatsby-plugin-page-creator`) // or a different package name

exports.createPagesStatefully = ({ actions: { createPage }}) => {
  watchDirectoryForPages(`./a-custom-page-directory`)
    .on(`newPage`, (path) => {
      // do custom validation & manipulate the path
      createPage(/* data */)
   })
}
```

/cc @gatsbyjs/core

KyleAMathews:
Moved my idea to a separate issue https://github.com/gatsbyjs/gatsby/issues/13606

@leonardodino would love your help creating this library!

PR #10962, opened 2019-01-10
docs(gatsby-source-wordpress): adapted section about self-signed certificates

crstnio:
## Description

To keep `gatsby-node.js` clean and tidy, `NODE_TLS_REJECT_UNAUTHORIZED` should be moved to `.env.development` – where all the other environment variables for development should be too.

Additionally, added a hint about caution for setting this in production.

### DIFF
--- a/packages/gatsby-source-wordpress/README.md
+++ b/packages/gatsby-source-wordpress/README.md
@@ -913,11 +913,13 @@ In order to resolve this, you can manually change the `post_parent` value of the
 
 When running locally, or in other situations that may involve self-signed certificates, you may run into the error: `The request failed with error code "DEPTH_ZERO_SELF_SIGNED_CERT"`.
 
-To solve this, you can disable Node.js' rejection of unauthorized certificates by adding the following to `gatsby-node.js`:
+To solve this, you can disable Node.js' rejection of unauthorized certificates by adding the following to `.env.development`:
 
-```javascript
-process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0"
 ```
+NODE_TLS_REJECT_UNAUTHORIZED=0
+```
+
+**CAUTION:** This should never be set in production. Always ensure that you disable `NODE_TLS_REJECT_UNAUTHORIZED` in development with `gatsby develop` only.
 
 [dotenv]: https://github.com/motdotla/dotenv
 [envvars]: https://www.gatsbyjs.org/docs/environment-variables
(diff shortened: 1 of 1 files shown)


pieh:
It might be worth adding that users need to add

```
require("dotenv").config({
  path: `.env.${process.env.NODE_ENV}`,
})
```

in gatsby-config or gatsby-node, for `.env` files to be loaded? (or link to https://www.gatsbyjs.org/docs/environment-variables/#server-side-nodejs )

cardiv:
Interestingly not. I assume since we do not use it directly in the code – the environment variables are accessible in Node, but they don't get exposed to the config files **by default**. As I understand it.

pieh:
> Interestingly not. I assume since we do not use it directly in the code – the environment variables are accessible in Node, but they don't get exposed to the config files **by default**. As I understand it.

Huh, that's pretty interesting. Do you know of quick way to verify this?

cardiv:
I have a local site running with _Local by Flywheel_. It sets up local WordPress sites with almost a single click and has a feature to trust certificates – they get self-signed.

Then I try `gatsby develop` without adding `NODE_TLS_REJECT_UNAUTHORIZED` to `.env.development`. Error: `Cannot read property 'allWordpressPage' of undefined`

After that, I add the _env_ and all works as expected.

pieh:
> I have a local site running with _Local by Flywheel_.

Oh cool - I used this year or 2 ago - will set this up and verify that way

pieh:
Yeah, so I checked it and it doesn't work without using:
```
require("dotenv").config({
  path: `.env.${process.env.NODE_ENV}`,
})
```

cardiv:
Dang! I forgot my gatsby-config.js using that to substitute `WORDPRESS_BASEURL` as env. 🤦‍♂️

So `.env.development` works for me out of the box. Let me add that quickly.

LekoArts:
@pieh LGTM. Anything you want changed? Otherwise I'd merge

wardpeet:
@LekoArts 

> Yeah, so I checked it and it doesn't work without using:
> 
> ```
> require("dotenv").config({
>   path: `.env.${process.env.NODE_ENV}`,
> })
> ```

is something that needs to change or rephrase like run NODE_TLS_REJECT_UNAUTHORIZED=1 npm run develop

LekoArts:
> Please note that you need to add `dotenv`, as mentioned earlier, to expose environment variables in your gatsby-config.js or gatsby-node.js files.

IMO this is sufficient as the earlier mentioned dotenv is already linking to our docs?

Issue #10815, opened 2019-01-03
dev: run image processing on demand


pieh:
## Summary

Don't process all images in develop mode and only do it when it's needed (requested by browser).

### Motivation

Booting up development server on large sites can take several minutes, because we wait for:
 - graphql query results
 - and **image processing** to finish (which usually is more time/cpu intensive task of the two).

Those are needed to produce complete production builds, but we don't need all the images for development. We only need images that user is looking at and we can lazy generate them when browser requests them.

### Details

This can be done by adjusting `gatsby-plugin-sharp` alone:
 - In development, instead of creating and running image processing job as soon as possible (during query running), push the job entry into some kind of in-memory store - we will need all the information needed to perform job and also way to find job that produces given file.
 - We will need to persist lazy jobs data. Because image processing would be lazy we have no guarantee that all images would be produced and we need to keep track of them.
 - Hook into dev express server and intercept requests for image files that are not yet produced -> trigger job processing to get that file -> respond with freshly produced image.

For reference: this is split from #7348 (running queries on demand).

/cc @wardpeet

wardpeet:
i've put it to blocked as lot of efforts is already done but we need some kind of internal invalidation api so we know which nodes are valid.

cc @pieh 😛

wardpeet:
Going to close this in favor of https://github.com/gatsbyjs/gatsby/discussions/7348

Issue #8125, opened 2018-09-13
OOM issues with markdown content


eLod:
## Description

gatsby build (and develop) dies with oom with ~800 somewhat heavy markdown content, only used plugins are fs-source and remark.

### Steps to reproduce

See repo at https://github.com/eLod/gatsby-bench

cc @pieh

pieh:
Also for ref https://github.com/gatsbyjs/gatsby/issues/6656

Issue #6656, opened 2018-07-22
[v2] Consider different method of persisting nodes data


pieh:
Currently we serialize entire nodes map every time we save redux state to `.cache/redux-state.json` - this gets problematic when we hit scenarios with large amount of nodes - it might lead to high memory footprint and longer than necessary I/O when we save state.

Potential solution would be switching to sqlite ( f.e. https://github.com/JoshuaWise/better-sqlite3 ) or some noSQL document store which could handle incremental in more performant manner. We can listen to `CREATE_NODE` and `DELETE_NODE` actions and incrementally update persisted data without serializing entire nodes map. Because this would be used just for persistence - I think we really only need performant key-value (id->node) store that handles incremental updates nicely.

Initial list of things to consider:
- performance comparisons
- make sure that this won't cause problems with persistent data consistency - we would have 3 different methods of persisting data:
  - saving redux-state.json (same as now minus nodes data) that uses `_.debounce` (as of right now after bootstrap in `gatsby develop`)
  - nodes specific method
  - key-value cache store (saved to `.cache/cache/db.json)

Ref: https://github.com/gatsbyjs/gatsby/issues/6611

KyleAMathews:
It'd be nice to make this pluggable. Most sites are fine with the existing serialization but a simple API could be built to allow for plugins to take over persisting changes.

It'd be nice too then to persist individual changes instead of the entire data tree each time as that's much faster obviously.

VindulaF:
It would nice to have this in, I am building an prototype using Gatsby for my company's static website and we have a Contentful space with about ~9000 entries excluding assets and 8 active locales. 

Thanks to  #6611 I can build the website without facing OOM issues but however it takes far too long at `source and transform nodes` it eats up a lot of time in the build process even when building from cache.  To be precise here are the numbers `source and transform nodes — 365.245 s` which is slow for a ~300 page static website. 

Right now the time it takes is a deal breaker for us but I love the flexibility that Gatsby offers hoping to see this and would love to help in testing this PR against our static website.

DSchau:
Let's re-open this--worth considering this again.

We have gatsbyjs/gatsby#10732, which may help, but worth considering if there are better approaches.

Some things I was thinking:

- Persist _separate_ parts of the redux state (and assemble on re-launch?)
    - Would need to benchmark to see how much slower this is
    - Some slowdown is acceptable (e.g. 5-10%?) if it means Gatsby is scalable to sites of 50K+ pages easily
[...]

Issue #6611, opened 2018-07-20
[v2] out-of-memory error with large amount of nodes


pieh:
With sufficently large amount of data heavy nodes Gatsby v2 will fail with `JavaScript heap out of memory` (without increasing heap size).

<details><summary>Terminal output</summary>

```
⢀ source and transform nodescontentTypes fetched 42
Updated entries  6908
Deleted entries  0
Updated assets  2800
Deleted assets  0
Fetch Contentful data: 108808.495ms
success source and transform nodes — 116.323 s

<--- Last few GCs --->

[60520:0x103800000]   132309 ms: Mark-sweep 1422.2 (1564.9) -> 1421.8 (1564.9) MB, 138.1 / 0.0 ms  allocation failure GC in old space requested
[60520:0x103800000]   132557 ms: Mark-sweep 1421.8 (1564.9) -> 1421.8 (1523.9) MB, 247.6 / 0.7 ms  last resort GC in old space requested
[60520:0x103800000]   132726 ms: Mark-sweep 1421.8 (1523.9) -> 1421.8 (1519.4) MB, 168.9 / 0.0 ms  last resort GC in old space requested


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0xf62d8aa57c1 <JSObject>
    1: mapToObject(aka mapToObject) [/Users/mike/dev/gatsbyinc/project/node_modules/gatsby/dist/redux/index.js:~32] [pc=0xb78e9d0547a](this=0xf625f7822d1 <undefined>,map=0xf622b48d6c1 <Map map = 0xf62f42048d9>)
    2: /* anonymous */ [/Users/mike/dev/gatsbyinc/project/node_modules/gatsby/dist/redux/index.js:103] [bytecode=0xf6295584df1 offset=79](this=0xf62dea0c211 <JSGlobal Object>,state...

FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
 1: node::Abort() [/Users/mike/.nvm/versions/node/v8.11.1/bin/node]
 2: node::FatalException(v8::Isolate*, v8::Local<v8::Value>, v8::Local<v8::Message>) [/Users/mike/.nvm/versions/node/v8.11.1/bin/node]
 3: v8::internal::V8::FatalProcessOutOfMemory(char const*, bool) [/Users/mike/.nvm/versions/node/v8.11.1/bin/node]
```
</details>

[...]

PR #10964, opened 2019-01-10
 refactor(gatsby-plugin-sharp): split single file into more maintainable chunks 

wardpeet:
## Description
This PR changes how the gatsby-plugin-sharp plugin handles images. Instead of processing our image queue right away we wait processing depending on your environment.

### Process
Inside queueImageResizing we are pushing every job into a memory queue and we don't care 
about processing here at all. Inside gatsby-node we're distinguishing develop and build.  

onPostBootstrap we're going to save our jobs into a cache. This cache is needed when we switch from develop to build mode. If no data changed we won't execute any queries which means no images are scheduled.

When we run `gatsby develop` we will use the `onCreateDevServer` hook to add an extra middelware to the express instance. Each file that doesn't exist yet gets fowarded to this route. We look into our queue if we know this url, if we do we process it and send the file over. If not we ignore and let express handle it.

`gatsby build` looks at our cached queue to see if it needs to do any processing. It loops over the queue and waits until all files are processed. We do this inside `onPostBuild` to make sure it's only ran in build mode.

## Updates
- Added jest-serializer-path to fix snapshots absolute paths
- Updated snapshots as sharps arguments got changed to reduce extension & toFormat checks
- Added integration tests to see if caching works between develop & build
- Added unit test to test image on demand.

## Related Issues
#10815


packages/gatsby-plugin-sharp/src/gatsby-node.js
```diff
@@ -1,10 +1,84 @@
...
+}
+
+/**
+ * Execute all unprocessed images on gatsby build
+ */
+let promises = []
+exports.onPreBuild = async ({ cache, boundActionCreators }, pluginOptions) => {
+  const cachedQueue = await getQueueFromCache(cache)
+
+  let promises = []
```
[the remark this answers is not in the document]
DSchau: If nothing else--it's sorta confusing, so I'd maybe take another look at this!

packages/gatsby-plugin-sharp/src/gatsby-node.js
```diff
@@ -1,10 +1,84 @@
...
+  Promise.all(promises)
+    .then(() => cache.set(`queue`, []))
+    .catch(({ err, message }) => {
+      reportError(message || err.message, err, reporter)
+    })
+
+/**
+ * Build images on the fly when they are requested by the browser
+ */
+exports.onCreateDevServer = ({ app, cache }, pluginOptions) => {
```
DSchau: This is slick!

packages/gatsby-plugin-sharp/.gitignore
```diff
@@ -2,3 +2,6 @@
 /node_modules
 /gatsby-node.js
 /duotone.js
```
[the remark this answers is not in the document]
wardpeet: wasn't sure how to handle this so I just kept adding but good idea!

packages/gatsby-plugin-sharp/src/gatsby-node.js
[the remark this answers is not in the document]
wardpeet: oh damn, yeah refactoring...

packages/gatsby-plugin-sharp/src/gatsby-node.js
KyleAMathews: nifty!

Could we name this `onCreateDevelopmentServer`? I'm strongly allergic to shorten names for APIs :-)

Is the changes to develop.js not pushed yet? Plus the API docs changes?

packages/gatsby-plugin-sharp/src/gatsby-node.js
```diff
@@ -1,10 +1,84 @@
...
+
+exports.onPreInit = async ({ actions }, pluginOptions) => {
   setBoundActionCreators(actions)
   setPluginOptions(pluginOptions)
 }
 
+/**
+ * save queue to the cache
+ */
+exports.onPostBootstrap = async ({ cache }) => {
```
pieh: in `develop` we can run queries after bootstrap (especially when working with files and editing them while `develop` is running) - it's good idea to wait for this event for initial save - but after that we would need to re-save queue

wardpeet: Any suggestions? I Probably need to save it when dev server is created, just before [line 99](https://github.com/gatsbyjs/gatsby/pull/10964/files#diff-f60f2b266f94ca7cde8437d8451cfe39R66).

If queries change is there any mechanism to know what nodes are being removed because I probably want to remove unused images.

packages/gatsby-plugin-sharp/src/gatsby-node.js
pieh: Oh, this is not new API (we have `onCreateDevServer` API for a while). So this is my bad for adding this API with shorten name :( mea culpa

packages/gatsby-plugin-sharp/src/gatsby-node.js
```diff
@@ -1,10 +1,93 @@
...
+    .then(() => cache.set(`queue`, []))
+    .catch(({ err, message }) => {
+      reportError(message || err.message, err, reporter)
+    })
+
+/**
+ * Build images on the fly when they are requested by the browser
+ */
+exports.onCreateDevServer = async ({ app, cache, compiler }, pluginOptions) => {
+  compiler.hooks.done.tap(`gatsby-plugin-sharp`, () => {
```
wardpeet: In need to hook in on the compilation lifcycle to know whenever a new build is done.

wardpeet:
So what I haven't fixed yet is removing images that should not be processed because a page or query changed.

packages/gatsby-plugin-sharp/src/gatsby-node.js
```diff
@@ -1,10 +1,93 @@
...
+    }
+
+    const job = queue.get(req.originalUrl)
+
+    // wait until the file has been processed and saved to disk
+    await Promise.all(processFile(job.file.absolutePath, [job], pluginOptions))
+    queue.delete(req.originalUrl)
+
+    // JSON.stringify doesn't work on an Map so we need to convert it to an array
+    cache.set(`queue`, Array.from(queue))
```
wardpeet: I should use ` saveQueueToCache `

packages/gatsby-plugin-sharp/src/gatsby-node.js
wardpeet: peer-review with @m-allanson 

> lets go for emitter.on('QUERY_QUEUE_DRAINED') instead of exposing compiler

wardpeet:
should be good to go

packages/gatsby-plugin-sharp/src/index.js
```diff
@@ -464,13 +176,29 @@ function queueImageResizing({ file, args = {}, reporter }) {
...
+
   return {
     src: prefixedSrc,
     absolutePath: filePath,
     width,
     height,
     aspectRatio,
-    finishedPromise,
+    // finishedPromise is needed to not break our API (https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-transformer-sqip/src/extend-node-type.js#L115)
+    finishedPromise: () => {
```
pieh: Will this actually work? Maybe we should make lazy image generation more opt-in? So `gatsby-transformer-sharp` and `gatsby-remark-images` specifically opt-in into it with something like extra `lazy` flag? and keep current behaviour of scheduling job asap if caller didn't opt into it?

wardpeet:
> But there is also one unsolved issue:
some images can be scheduled to be processed, but later on should be invalidated that currently doesn't happen - for most part this would just mean extra unneded work, but there is edge cases where this would cause straight errors - for example image was scheduled, but later on source image was deleted and now this scheduled task have no way of finishing and we would actually throw errors about trying to read not-existing files? Initially we can just add some sanity checks to ensure input file exists before trying to process it I think, but later on I think we need to keep track of queries to remove stale tasks	

Is it ok if I do file test in this PR and do the query stuff in a follow up PR.

packages/gatsby-plugin-sharp/src/processFile.js
```diff
@@ -0,0 +1,182 @@
...
+const debug = require(`debug`)(`gatsby:gatsby-plugin-sharp`)
+const duotone = require(`./duotone`)
+const imagemin = require(`imagemin`)
+const imageminMozjpeg = require(`imagemin-mozjpeg`)
+const imageminPngquant = require(`imagemin-pngquant`)
+const imageminWebp = require(`imagemin-webp`)
+
+const writeFileAsync = (file, buffer) =>
+  new Promise((resovle, reject) => {
+    fs.writeFile(file, buffer, err => {
```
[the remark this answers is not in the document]
wardpeet: sweet i'll inline, it's better that way :)

packages/gatsby-plugin-sharp/src/index.js
```diff
@@ -464,13 +176,29 @@ function queueImageResizing({ file, args = {}, reporter }) {
   const prefixedSrc =
     options.pathPrefix + `/static/${digestDirPrefix}` + encodedImgSrc
 
+  // Create job and add it to the queue, the queue will be processed inside gatsby-node.js
+  const job = {
+    file,
```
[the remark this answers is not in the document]
wardpeet: good point, I wasn't thinking of this, i'll normalize it.

packages/gatsby-plugin-sharp/src/processFile.js
```diff
@@ -0,0 +1,182 @@
...
+      ((transformFile.extension === `jpg` && args.toFormat === ``) ||
+        (transformFile.extension === `jpeg` && args.toFormat === ``) ||
+        args.toFormat === `jpg`)
+    ) {
+      await compressJpg(clonedPipeline, outputPath, args)
+      return transform
+    }
+
+    if (
+      (file.extension === `webp` && args.toFormat === ``) ||
```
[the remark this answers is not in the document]
wardpeet: yes 🤦‍♂️

pieh:
> Is it ok if I do file test in this PR and do the query stuff in a follow up PR.

👍 from me - it will be much easier to review

packages/gatsby-plugin-sharp/src/index.js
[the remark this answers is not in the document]
wardpeet: I can even remove it as only absolutePath is needed from file (if not mistaken) need to check to be sure

packages/gatsby-plugin-sharp/src/index.js
wardpeet: I'm checking if I don't break but I actually like this approach as no one needs to opt-in for speed. Most plugins have enough with just queuing a job.

I probably write an  RFC about cleaning this plugin up, it will need some changes to core to expose reporter and createJob from a require.

wardpeet: @pieh I added a test to demonstrate that this is working. I also tested it with the examples/using-sqip where this is used and it is working

wardpeet:
Some updates:
- Added jest-serializer-path to fix snapshots absolute paths
- Updated snapshots as sharps arguments got changed to reduce extension & toFormat checks
- Added integration tests to see if caching works between develop & build
- Added unit test to test image on demand.

I'm pretty confident of shipping this when all tests are green. I've tested in on www and some examples and I did not run into any issues. Big issues take some time to get resized but aren't blocking for the user 👍

wardpeet:
when this get approved, let me know so I can merge it myself and write a proper commit message :)

jlengstorf:
This is so exciting!

pieh:
@m-allanson yes! I hope to do this today

jlengstorf:
We need a shipit squirrel emoji reaction.

packages/gatsby-plugin-sharp/src/gatsby-node.js
```diff
@@ -1,10 +1,96 @@
...
+
+  // merge both queues
+  for (const [key, job] of cachedQueue) {
+    if (!queue.has(key)) {
+      queue.set(key, job)
+    }
+  }
+
+  // JSON.stringify doesn't work on an Map so we need to convert it to an array
+  return cache.set(`queue`, Array.from(queue))
```
pieh: This is only thing that could potentially be blocker.

Just to get it out of the way - I suggested using `cache` initially, so this is my fault.

Since we started looking into cache issues, I started paying more attention to potential problems with it. Problem with using `cache` instead of redux state here would be that we potentially can have inconsistent state. 

Processing queue is linked to query running (not exclusively but major usage of it is). State that triggers query running is stored in redux state (there are other triggers, but things we persist is stored there). Scenario when processing queue fails to persist (to `cache`), but redux state persist correctly can lead to situations when images that are supposed to be created during production build, aren't, and build completes successfully (no errors thrown).

Long story short - we can't store related chunks of data in different persistence engines. We need to use redux state here. 

Luckily we have easy way to do that through [setPluginStatus](https://www.gatsbyjs.org/docs/actions/#setPluginStatus) action:
```
exports.whateverAPI = async ({ actions }) => {
  actions.setPluginStatus({ queue: Array.from(queue) })
}
``` 

Reading it is not as nice, but still OK-ish:
```js
let getPluginState = null
exports.onPreBootstrap = ({ store }) => {
  getPluginState = () => {
    const pluginStatus = store.getState().status.plugins[`gatsby-plugin-sharp`]
    if (pluginStatus && pluginStatus.queue) {
      return new Map(pluginStatus.queue)
    }
    return new Map()
  }
}
```
(ideally we add helper function for this to gatsby, but I would suggest using this to be backward compatible as there aren't any breaking changes here)

So all this can potentially increase problems with persisting redux state (higher memory usage / long writes), which is not great :/ Some issues for reference - https://github.com/gatsbyjs/gatsby/issues/8125 https://github.com/gatsbyjs/gatsby/issues/6611 https://github.com/gatsbyjs/gatsby/issues/6656

So I hate to even suggest this given how great feature this is - but should we give a way to opt-out of it? (if other folks agree with my essay that is - I would very happy to be proven wrong about inconsistency issue - it happened before)

www/README.md
```diff
@@ -39,15 +39,3 @@ GATSBY_FEEDBACK_KEY_STARTERLIB=ADD_KEY
...
-
-If you are not working on starter or site showcase, it might be beneficial to use a placeholder image instead of actual screenshots. It will skip downloading screenshots and generating responsive images for all screenshots and replace them with a placeholder image.
-
-Add the following env variable to your `.env.development` file to enable placeholder behaviour:
-
-```
-GATSBY_SCREENSHOT_PLACEHOLDER=true
-```
-
-For more information checkout [`gatsby-transformer-screenshot` docs](http://www.gatsbyjs.org/packages/gatsby-transformer-screenshot#placeholder-image).
```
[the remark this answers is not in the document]
wardpeet: sweet i'll revert

packages/gatsby-plugin-sharp/src/index.js
```diff
@@ -464,19 +185,37 @@ function queueImageResizing({ file, args = {}, reporter }) {
...
   return {
     src: prefixedSrc,
     absolutePath: filePath,
     width,
     height,
     aspectRatio,
-    finishedPromise,
+    // finishedPromise is needed to not break our API (https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-transformer-sqip/src/extend-node-type.js#L115)
+    finishedPromise: {
+      then: (resolve, reject) => {
```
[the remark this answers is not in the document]
wardpeet: use new Promise where you can :) I once read that this worked with async/await just unsure where I found it, tried to google to put a reference here but 🤷‍♂️

packages/gatsby-plugin-sharp/src/gatsby-node.js
wardpeet: @pieh thanks a lot! I'll see what I can do with it. I'll add a new pluginOption `lazyImageGenaration` (if you have a better please tell me). Which is set by default on true.

It's sad to hear that redux state persisting might cause memory issues. We maybe should try to move to something else or be smarter about our store, like don't load it upfront unless it's asked for and clear objects when they aren't accessed anymore

packages/gatsby-transformer-toml/src/__tests__/__snapshots__/gatsby-node.js.snap
```diff
@@ -81,7 +81,7 @@ Array [
     test_array = [ \\"] \\", \\" # \\"]      # ]
     test_array2 = [ \\"Test #11 ]proved that\\", \\"Experiment #9 was a success\\" ]
     another_test_string = \\" Same thing, but with a string #\\"
-    harder_test_string = \\" And when \\\\\\"'s are in the string, along with # \\\\\\"\\"
+    harder_test_string = \\" And when /\\"'s are in the string, along with # /\\"\\"
```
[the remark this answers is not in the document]
wardpeet: yeah it is 🤔 it's probably a bug as they shouldn't touch it but it does it consistently so it's not breaking. I'll make a bug on their repo and perhaps fix it

packages/gatsby-plugin-sharp/src/gatsby-node.js
wardpeet: I did the changes, unsure if this is what you meant 😛 @pieh

pieh: Memory issues mostly are related to saving state to file - and not saving part but serializing ( https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/redux/index.js#L91 ).

Just the way `stringify` (either `JSON.stringify` or one from `json-stringify-safe`) works, is that it creates many intermediatte strings which add up in memory and not being garbage collected quick enough by Node's V8 engine.

We tried some streaming alternatives but this wasn't really working for us (at least our implementation of using it). See https://github.com/gatsbyjs/gatsby/pull/9896 for revert of it and linked comment in PR description that summarise issue we had with it.

packages/gatsby-transformer-toml/src/__tests__/__snapshots__/gatsby-node.js.snap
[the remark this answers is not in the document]
pieh: I don't mind if this is consistent (and unit tests pass on both circleCI and azure so we should be fine). Was just curious where this change came from

KyleAMathews:
I don't think this will add that much more data to the cache right? Seems minor relative to the benefits we'll get.

We need to figure out more scalable ways of storing the cache but given that this is a relatively small % of the cache data, it won't cause more trouble to people already having trouble.

pieh:
for gatsbyjs.org on cold cache it will add ~12k entries with shape like this https://github.com/gatsbyjs/gatsby/blob/03d4e7cbd2dbdf3adf576c202c39ef8f9b024a3d/packages/gatsby-plugin-sharp/src/__tests__/__snapshots__/index.js.snap#L78-L95

I'm doing actual testing right now - yesterday I was doing code review only. So will report `redux-state.json` before/after comparison.

wardpeet:
@DSchau thanks! I hope we can get more of these integration tests running to test some core pieces.

@pieh if it helps I can omit default values from the args to make the cache smaller. I could do the same for inputPath & outputPath (strip the program root, where possible). I added an escape hatch called `lazyImageGeneration` that you can set to false if we want to opt out. I should add it to the README.md

pieh:
>@pieh if it helps I can omit default values from the args to make the cache smaller.

If it's relatively straight forward to do - let's strip as much as we can. But with changes you did to not persist entire File node earlier we are pretty good already (so this is nice to have, not a blocker)

I'm still checking and validating this, but initial checks redux-state.json for gatsbyjs.org (with screenshot placeholders - so not trully production build, but still ~2400 image thumbnails) increased from `26314316` to `30325343` bytes so ~15% (that's after running gatsby develop and not visiting site, so all images processing tasks are in cache).

Moocar:
@wardpeet Thanks for the heads up on this PR. I'm working on making the graphql resolvers run in other processes. To do this, those resolvers can't rely on accessing the redux store directly, since it runs in another process. This was fine, except that `plugin-sharp` manually requires the redux store in order to call the `setJob` bound action creators. When running in the main process, that's fine, but when it's running in another process, it actually ends up creating a redux store per process, which means the setJob calls are entirely uncoordinated.

I've been teasing that out so that the job API functions are provided to plugin-sharp instead of it having to require them. This PR makes it even harder to run plugin-sharp in another process because it introduces more globals and dependence on the gatbsy-node APIs. But the benefits are obviously massive so I recommend merging this PR, and I'll have a think about whether it's possible to make this run in parallel resolvers land.

Moocar:
I think it might be possible to make parallel resolvers work here. Calls to `fluid`, `fixed` etc don't cause side effects (apart from reading image size from disk and generating base64) until they add the processing job to the queue. So theoretically, we could change `queueImageResizing` so that if it's running in another process, it instead makes a call to the main process where it puts the job into the in memory queue. Otherwise it runs as normal.

I'll wait till this is merged, and then have more of a think about it.

wardpeet:
@pieh I could run this to remove default ones:
https://repl.it/@wardpeet/omit-defaults

I could even remove lodash if needed. WDYT? should I sneak it in and than we're good to go?

wardpeet:
@pieh found an issue where we could remove images from disk between develop & build. At the moment we fail hard 😄. The idea is to do a simple file_exists on all files that are inside the cached version and ignore those errors. We will still fail hard if our current queue has errors.

Any issues with this approach? I would love to get this merged 😄

pieh:
I was thinking yesterday and today about this and this approach also have problems:
- use default starter
- run gatsby develop (don't open browser) - image in https://github.com/gatsbyjs/gatsby-starter-default/blob/master/src/components/image.js won't be processed but job will be cached
- delete "src/images/gatsby-astronaut.png"

With silencing error build succeed, but we have missing assets for that image. I think build should fail.

I don't have good approach yet. We might need for gatsby to expose more information to `gatsby-plugin-sharp` to properly invalidate stale jobs

wardpeet:
Turned lazy loading off so we can merge this without having impact 👍 

cc @pieh

sidharthachatterjee:
Published in gatsby-plugin-sharp@2.0.26

KyleAMathews:
Just a thought — if image processing jobs are marked as a dependency of a query — then when we re-run the query, we would delete all the old jobs associated with it (and recreate them of course when the query runs). This would solve I believe the stale job problem because deleting a file between dev and build would dirty the query making it re-run.

muescha:
any plans to turn the `lazyImageGeneration` on?

wesbos:
Please turn this on :) My build takes ~8 mins on a really fast macbook pro before I can even see my site.

Issue #10948, opened 2019-01-09
Show progress for downloading remote files


KyleAMathews:
Many plugins rely on [`createRemoteFileNode` from `gatsby-source-filesystem`](https://www.gatsbyjs.org/packages/gatsby-source-filesystem/?=#createremotefilenode) to download files from remote systems. E.g. gatsby-source-wordpress downloads all the files from the wordpress instance so you can include them in your site and do data transformations e.g. image thumbnail creation.

This is great but some systems have many files which can take a long time to download. This downloading can cause confusion because the Gatsby bootstrap just stalls out for no apparent reason.

To start improving this, the first thing we need to do is to show a progress bar & current status for downloads.

Related issue about problems this causes with wordpress https://github.com/gatsbyjs/gatsby/issues/6654

polarathene:
Also related: [#6125](https://github.com/gatsbyjs/gatsby/issues/6125#issuecomment-452585023)

Over 11 minutes on `source and transform nodes` for me with one of the example projects(gatsby-image that pulls 130MB remote images from unsplash). Before being aware of the large remote content being downloaded I thought the build process had become stuck.

My suggestion in the linked issue was to consider a `(sourced_nodes/total_nodes)` feedback counter for the step with a timer delay before showing some hints to why the process might be taking long to the developer(especially useful and perhaps easier to implement if knowing the total nodes/remote content in advance is not an easy feature to provide). Linked issue contains some example messages/tips that could cycle.

oorestisime:
I would love to give this a go since i use that api quite often. I am assigning it to myself to avoid double work.

PR #10979, opened 2019-01-10
[gatsby-source-filesystem] Add progress bar

oorestisime:
Fix: #10948

## Description
Add a progress bar while downloading remote content, so that user gets a live feedback instead of thinking build is stuck or so

### DIFF
--- a/packages/gatsby-source-filesystem/src/create-remote-file-node.js
+++ b/packages/gatsby-source-filesystem/src/create-remote-file-node.js
@@ -6,10 +6,19 @@ const { isWebUri } = require(`valid-url`)
 const Queue = require(`better-queue`)
 const readChunk = require(`read-chunk`)
 const fileType = require(`file-type`)
+const ProgressBar = require(`progress`)
 
 const { createFileNode } = require(`./create-file-node`)
 const { getRemoteFileExtension, getRemoteFileName } = require(`./utils`)
 const cacheId = url => `create-remote-file-node-${url}`
+
+const bar = new ProgressBar(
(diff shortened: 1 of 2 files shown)


packages/gatsby-source-filesystem/src/create-remote-file-node.js
```diff
@@ -6,11 +6,20 @@ const { isWebUri } = require(`valid-url`)
...
 const readChunk = require(`read-chunk`)
 const fileType = require(`file-type`)
+const ProgressBar = require(`progress`)
 
 const { createFileNode } = require(`./create-file-node`)
 const { getRemoteFileExtension, getRemoteFileName } = require(`./utils`)
 const cacheId = url => `create-remote-file-node-${url}`
 
+const bar = new ProgressBar(
+  `Downloading remote content [:bar] :current/:total :elapsed secs :percent`,
```
KyleAMathews: Could you change content to "files"

KyleAMathews:
Woot it works!

screenshot 2019-01-11 10 37 28


One problem with this right now is it blinks back and forth with the the updates from the "source and transform nodes" log line — we need to add a way to core to display multiple things at once and clean up stuff like this.

But given this is a frequently confusing thing for folks — wondering why the Gatsby bootstrap stalls — let's merge this in and then fix up our console output later.

KyleAMathews:
This is what it looks like — definitely better than not having it at all but it'd be nice if we abstracted outputting updates to the console so two different processes aren't fighting with each other.

This also happens with images.

overlapping-outputs

oorestisime:
Thanks for the quick review and merge :) is there an issue tracking the problem you mention with the blinking or should i open one?

PR #10980, opened 2019-01-10
fix(gatsby-source-filesystem): report when downloading fails

oorestisime:
## Description
While working on #10979 i came across this. 
The queue was receiving the error as the result argument.
i Added a warn there in order to avoid failing silently

### DIFF
--- a/packages/gatsby-source-filesystem/src/create-remote-file-node.js
+++ b/packages/gatsby-source-filesystem/src/create-remote-file-node.js
@@ -211,47 +211,38 @@ async function processRemoteNode({
   const tmpFilename = createFilePath(pluginCacheDir, `tmp-${digest}`, ext)
 
   // Fetch the file.
-  try {
-    const response = await requestRemoteNode(url, headers, tmpFilename)
-    // Save the response headers for future requests.
-    await cache.set(cacheId(url), response.headers)
+  const response = await requestRemoteNode(url, headers, tmpFilename)
+  // Save the response headers for future requests.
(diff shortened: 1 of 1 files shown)


packages/gatsby-source-filesystem/src/create-remote-file-node.js
```diff
@@ -114,7 +114,8 @@ async function pushToQueue(task, cb) {
     const node = await processRemoteNode(task)
     return cb(null, node)
   } catch (e) {
-    return cb(null, e)
+    console.warn(`Failed to process remote content ${task.url}`)
+    return cb(e)
```
pieh: This code path seems to never be hit, because of try/catch in `processRemoteNode` function ( https://github.com/gatsbyjs/gatsby/blob/03dc6c97b7ea56ac28de5fd563f3769c74e6e16f/packages/gatsby-source-filesystem/src/create-remote-file-node.js#L262-L265 ).

To me it seems we should remove try/catch there (or rethrow error, but given we don't do anything with error there I think removing is fine)

pieh:
Looking good. In the future we will want to move to use `reporter.warn`, but this is in general problematic right now because we don't pass reporter to `createRemoteFileNode` so for now `console.warn` is fine-ish

oorestisime:
should i open an issue for that?

pieh:
@oorestisime We don't have issue for not using `console.(log|warn|error)` yet, so let's not create issue specifically for it yet.

Issue #10978, opened 2019-01-10
Warn in development when people use `<Link>` for external links


KyleAMathews:
People often don't realize that `<Link>` can't be used with external links.

We should detect this and warn people in the console and point them to https://www.gatsbyjs.org/docs/gatsby-link/#use-link-only-for-internal-links

Probably we could use the same regex there to test links and print out a warning in the console.

PR #10981, opened 2019-01-10
[gatsby-link] warn when <link> is used for external links

oorestisime:
Fix: #10978

### DIFF
--- a/packages/gatsby-link/src/__tests__/index.js
+++ b/packages/gatsby-link/src/__tests__/index.js
@@ -86,9 +86,10 @@ describe(`<Link />`, () => {
     }
 
     it(`accepts to as a string`, () => {
+      jest.spyOn(global.console, `warn`)
       const location = `/courses?sort=name`
       const { link } = setup({ linkProps: { to: location } })
-
+      expect(console.warn).not.toBeCalled()
       expect(link.getAttribute(`href`)).toEqual(location)
     })
 
@@ -97,6 +98,13 @@ describe(`<Link />`, () => {
       const location = `/courses?sort=name`
       const { link } = setup({ linkProps: { to: location }, pathPrefix })
       expect(link.getAttribute(`href`)).toEqual(`${pathPrefix}${location}`)
+    })
+
+    it(`warns when not internal`, () => {
+      jest.spyOn(global.console, `warn`)
+      const to = `https://gatsby.org`
+      setup({ linkProps: { to } })
+      expect(console.warn).toBeCalled()
     })
   })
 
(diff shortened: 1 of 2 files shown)


packages/gatsby-link/src/__tests__/index.js
```diff
@@ -86,9 +86,10 @@ describe(`<Link />`, () => {
     }
 
     it(`accepts to as a string`, () => {
+      jest.spyOn(global.console, `warn`)
```
DSchau: Nit: but I'd keep this test separate from this functionality.

In other words--I'd probably add a test testing that `does not warn when internal link` which just tests the console.warn functionality.

packages/gatsby-link/src/index.js
```diff
@@ -104,6 +104,12 @@ class GatsbyLink extends React.Component {
       ...rest
     } = this.props
 
+    if (process.env.NODE_ENV !== `production` && !/^\/(?!\/)/.test(to)) {
```
DSchau:
```suggestion
const LOCAL_URL = /^\/(?!\/)/
    if (process.env.NODE_ENV !== `production` && !LOCAL_URL.test(to)) {
```

packages/gatsby-link/src/__tests__/index.js
oorestisime: Heh yeah my bored personnality took over :D Fixed

packages/gatsby-link/src/index.js
```diff
@@ -104,6 +104,13 @@ class GatsbyLink extends React.Component {
       ...rest
     } = this.props
 
+    const LOCAL_URL = /^\/(?!\/)/
+    if (process.env.NODE_ENV !== `production` && !LOCAL_URL.test(to)) {
+      console.warn(
+        `Use the Link component only for internal links. See: https://gatsby.app/internal-links`
```
pieh: Should we add `to` prop value to warning message? It might be tedious to find instance of `Link` that uses external link without extra context. It still might not be great when `to` prop is generated dynamically, but at least you could search for it in DOM explorer to pinpoint where this is coming from

oorestisime: Yeah didn't think of that! 
Not sure how to word this.  Something like: `External link ${to} was detected in a Link component. Use the Link component only for internal links. See: https://gatsby.app/internal-links`

DSchau: I am good with both 1) the comment, and 2) the proposed solution :) Thanks @oorestisime!

Issue #10950, opened 2019-01-09
Site search is linking to headers when not needed


shannonbux:
## Description & Steps to Reproduce

1. I typed in "api" into the search bar on .org
2. clicked on the title "Debugging replace renderer api"
3. it took me to:
https://www.gatsbyjs.org/docs/debugging-replace-renderer-api/#debugging-replace-renderer-api

Seems strange that it took me to the title instead of the overall page, since they have the same name...the URL just looks messy.

### Expected result

It should have taken me to https://www.gatsbyjs.org/docs/debugging-replace-renderer-api/

### Actual result

It took me to https://www.gatsbyjs.org/docs/debugging-replace-renderer-api/#debugging-replace-renderer-api

### Environment

Chrome browser

arjunnn:
Hi, may I submit a PR for this? It would check if the fragment part is the same as the slug and remove it.

sidharthachatterjee:
Sure @arjunnn 

We'd love to receive a pull request fixing this! You can add the check in https://github.com/gatsbyjs/gatsby/blob/2684055ba669737c983769159e6726cc4685acb0/www/src/components/search-form.js#L275

Thanks!

glauberfc:
Hi, @sidharthachatterjee, I trying to solve this, but I have some questions:

1. What is `navigate(${a.pathname}${a.hash}) ` for? I thought that it redirect the user when it clicked on a search's result link, but I commented it and the page continued to be redirect to link.

2. Where the search result link is created? I noted that it already have a link and I don't know how update it

glauberfc:
I tried to solve this issue on this way:

![image](https://user-images.githubusercontent.com/15311858/51000826-f5d9e700-150c-11e9-87e1-370f91949f10.png)

For example on the url **https://www.gatsbyjs.org/docs/debugging-replace-renderer-api/#debugging-replace-renderer-api**:
1. I extract the hash **#debugging-replace-renderer-api** and remove the **#** and save on const **hashString**;
2. I extract the pathname **/docs/debugging-replace-renderer-api/** and remove **/docs/** and the final **/** and save on const **sourceString**;
3. In this example if **hashString** and **sourceString** be equals, the page will be redirected to **/docs/debugging-replace-renderer-api/** else it will be redirected to **/docs/debugging-replace-renderer-api/#debugging-replace-renderer-api**

arjunnn:
Hi @sidharthachatterjee, thank you for the pointers, I just submitted a PR.

[...]

PR #10986, opened 2019-01-11
fix(www): add check to compare and remove hash with slug when navigating to search results

arjunnn:
## Description

[www] fix messy URLs, in cases where the slug and fragment of the URL are the same. 

Example: https://www.gatsbyjs.org/docs/debugging-replace-renderer-api/#debugging-replace-renderer-api becomes https://www.gatsbyjs.org/docs/debugging-replace-renderer-api/

## Related Issues

Addresses #10950


arjunnn:
Thank you so much @pieh :smiley: 
:purple_heart: Gatsby

PR #10994, opened 2019-01-11
style(www): improve style of code block language tabs

Raincal:
## Description
- adjust the style of the type tabs
- fix highlighting on https://www.gatsbyjs.org/docs/node-apis/

### before:
before

### after:
after


DSchau:
@Raincal this is looking great! Why the change to `data-language` attributes and targeting them specifically?

www/src/utils/typography.js
```diff
@@ -108,102 +108,95 @@ const _options = {
...
-        textAlign: `right`,
         color: `#444`,
         fontWeight: `700`,
         letterSpacing: `0.8px`,
         textTransform: `uppercase`,
         borderRadius: `0 0 5px 5px`,
         background: `#ddd`,
       },
-      ".gatsby-highlight pre[class='language-javascript']::before": {
+      ".gatsby-highlight[data-language='javascript']::before": {
```
Raincal: @DSchau After this change, the type tab will disappear if the `data-language` is not included.

DSchau: cc @amberleyromo is this behavior we'd even want?

@Raincal this makes this functionality opt-in correct? I think we want this on all of our language blocks.

Raincal: In .md files, the data-language attribute will be added by gatsby-remark-prismjs.
https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-remark-prismjs/src/index.js

DSchau: Right - but is there a technical reason to make it opt-in? We already sort of have that functionality if we wanted it, because we (granularly) would just remove the style for that type of code block we don't want highlighted.

In other words--just trying to understand why the change in behavior was made! Thanks for understanding!

Raincal: @DSchau You are right. It's not necessary to use `data-language`.

LekoArts:
@fk Could you please have a look at the merge conflict (the typography.js file) -- It has to do with your design tokens. I don't have context on this.

fk:
@LekoArts Done.

@Raincal Thank you for this PR, apologies for only getting aware of it now!
I took the liberty to fix the conflict (f78cc22), update your PR to make use of the new design tokens in 6dabb7c, and fix the padding/intendion for larger screens (a92b874). Hope that's fine with you!

DSchau:
Thanks so much to **both** @Raincal and @fk for getting this out the door.

Appreciate it--think this is a nice improvement!

www/src/utils/typography.js
fk: @Raincal Still catching up on things (unfortunately after we merged :/).
I just checked the codesandbox demo — thanks for putting that together! 🙏 

So to summarize:

- this PR makes `pre` the container that provides the scrollbar for overflowing content (as opposed to `.gatsby-highlight`)
- Mobile Safari doesn't respect `position: absolute` in overflowing content, so if we attach the language badges to its `before` pseudo element, they will scroll with the content on Mobile Safari (see @Raincal's demo)
- this is why @Raincal's original PR attached the badge to `.gatsby-highlight` instead, but that change was reverted in a453b22 after the thread here

@DSchau Looking at https://github.com/gatsbyjs/gatsby/blob/794fb4be4c47744379d575f68021987bec4d74fb/packages/gatsby-remark-prismjs/src/index.js#L69-L78 and also https://github.com/gatsbyjs/gatsby/blob/794fb4be4c47744379d575f68021987bec4d74fb/packages/gatsby-remark-prismjs/src/index.js#L39-L42, I'm thinking we should always have the same values for `data-language` on `.gatsby-highlight` and `class` on `.gatsby-hightlight pre`, and a default value of `text` … so I think @Raincal's original approach would not have caused any change irt to "opt-in/out"?

Maybe I'm not seeing the forest before the trees, but I think the change suggested by @Raincal should not alter code badges in any way other than fixing the "scrolling away" problem not only for desktop browsers, but also for Mobile Safari.

@Raincal Would love if you wanted to chime in if you find the time!

PR #10996, opened 2019-01-11
Fix plugin search result selected style

### LINKED ISSUE #10951: Plugin library shows two plugins selected
## Description & steps to reproduce

At https://www.gatsbyjs.org/plugins/
1. I typed in "shopify"
2. I clicked on `gatsby-source-shopify2`
3. I couldn't tell which READme file was getting pulled up on the right, partially because the titles of both `gatsby-source-shopify2` and `gatsby-source-shopify` are the same, and partially because the sidebar looked like both plugins were selected:

screen shot 2019-01-08 at 8 09 26 pm

### Expected result

Just one plugin should be able to be selected at a time

### Actual result

Both of the top 2 plugins look selected.

### Environment

Chrome browser

tsriram:
There's a bug in the way we determine the selected package in plugins page. It was just checking if the selected package's name is part of the pathname. So if you select `gatsby-source-shopify2` or `gatsby-source-shopify-storefront`, `gatsby-source-shopify` will also be highlighted (screenshot below)

*Before*
image

This PR fixes this issue. This also works with scoped packages.

*After*
image

Closes #10951


LekoArts:
19_01_8c400cs2

Works!

Fixed the linting.

www/src/components/plugin-searchbar-body.js
```diff
@@ -303,9 +303,11 @@ class Search extends Component {
 // the result component is fed into the InfiniteHits component
 const Result = ({ hit, pathname, query }) => {
   // Example:
-  // pathname = `/plugins/gatsby-link/` || `/plugins/@comsoc/gatsby-mdast-copy-linked-files`
+  // pathname = `/packages/gatsby-link/` || `/packages/@comsoc/gatsby-mdast-copy-linked-files`
   //  hit.name = `gatsby-link` || `@comsoc/gatsby-mdast-copy-linked-files`
-  const selected = pathname.includes(hit.name)
+  const packagesRegex = new RegExp(`/packages/`, `g`)
+  const packageName = pathname.replace(packagesRegex, ``).slice(0, -1)
+  const selected = packageName === hit.name
```
wardpeet:
```suggestion
  const selected = new RegExp(`^/packages/${hit.name}/?$`).test(pathname)
```
we could replace it with above regex. IMO it's clearer than the 3 lines.

tsriram: @wardpeet I like this too :+1:

PR #10997, opened 2019-01-11
fix(gatsby-source-wordpress): If given route is relative url then replace it with absolute url

imanhodjaev:
## Description

Use cases exist where wordpress is configured via plugins to serve relative urls and wordpress plugins skips those relative routes. To support this use case I added detection if route is a relative url using `startsWith` and if it is so then build an absolute url instead.

Thanks.

### DIFF
--- a/packages/gatsby-source-wordpress/src/fetch.js
+++ b/packages/gatsby-source-wordpress/src/fetch.js
@@ -131,6 +131,7 @@ Mama Route URL: ${url}
     let validRoutes = getValidRoutes({
       allRoutes,
       url,
+      baseUrl,
       _verbose,
       _useACF,
       _acfOptionPageIds,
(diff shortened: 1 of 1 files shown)


cardiv:
Hey @imanhodjaev,

thank you for the contribution!

Looks good to me, but a review from someone with write access is still required to merge.

Apart from this: I do not understand why someone would like to have relative urls in WordPress, since absolute urls are [recommended](https://core.trac.wordpress.org/ticket/17048#comment:2).

imanhodjaev:
@cardiv thanks for reviewing it. We have been using following setup 

1. Gatsby + WP plugin,
2. Our domain bound to Netlify,
3. Serving only content which available once gatsby built website,
4. Using redirects on Netlify to proxy all unknown requests to our WP website.

I hope it clarifies your questions.

imanhodjaev:
So we decided to go the recommended way and I will close this PR.
Thank you very much for your time and efforts @cardiv.

PR #11008, opened 2019-01-11
chore(gatsby): update Sift to version 7

stefanprobst:
Update Sift to v7 which contains some important fixes like querying for Dates (which we will want to use in the future).


wardpeet:
you probably want to change the require here too.
https://github.com/gatsbyjs/gatsby/blob/4bcca2a529aae89707c64e09b5aa0abaa19cf35a/packages/gatsby/src/db/loki/nodes-query.js#L4

stefanprobst:
@wardpeet Thx! Fixed. Tests did pass before as well though.

pieh:
> @wardpeet Thx! Fixed. Tests did pass before as well though.

This code path isn't tested currently and is behind feature flag, we will be looking to run tests for that (either subset of all tests or even all tests), but it's not done yet.

KyleAMathews:
The plan is to replace sift with loki.js (already in behind a feature flag) — so probably doesn't make much sense to touch sift anymore.

pieh:
> AFAIK loki.js can only be used for querying when we don't have to manually call resolvers before, no? I thought this is what the `lazy-fields` Map was for.

Yes, we still need to fallback to `sift` in case non-default field resolver need to be used to filter or sort nodes

KyleAMathews:
>Yes, we still need to fallback to sift in case non-default field resolver need to be used to filter or sort nodes

Is this a permanent restriction? Maintaining two separate query systems isn't ideal of course :-)

pieh:
> Is this a permanent restriction? Maintaining two separate query systems isn't ideal of course :-)

I think we can move fully to loki - we still would need to resolve nodes as we do now, but we can put those resolved nodes into semi-temporary loki collections, which should let us drop sift completely. @Moocar am I missing something that would block it?

Moocar:
@pieh @KyleAMathews I'm pretty sure we can remove `run-sift.js` entirely. See my proposed solution [here](https://github.com/gatsbyjs/gatsby/issues/11184)

stefanprobst:
> Could you expand on what you mean by supporting querying on arrays?

@Moocar I have filtering and sorting on arrays of scalars or arrays of objects working [here](https://github.com/stefanprobst/gatsby/tree/rfc-schema-refactor). For example `allType(filter: { arrayOfObjects: { intField: { in: [1,2,3] } }) { id }` or `allType(sort:{fields: [arrayOfObjects___intField] }) { id }`

stefanprobst:
@Moocar Sorry I have to take this back, what I have is for arrays of objects, but actually *not* the behavior of `elemMatch`... I only now understood that elemMatch checks that *all* query criteria match on *one* object. This wasn't clear to me maybe also because the current tests only check one prop - where there is no difference. In any case, I do feel that `elemMatch` feels not very GraphQL-y but the alternative would probably be to allow query args not only on the root Query type fields, which would be a bigger change.

stefanprobst:
Should we update Sift, or close this?

PR #11009, opened 2019-01-11
fix(gatsby-source-contentful): Fix check for fluid and fixed size fliters

sidharthachatterjee:
Fixes three bugs
1. `resolveFixed` was incorrectly filtering out a size equal to the original size when creating `srcset` (https://github.com/gatsbyjs/gatsby/issues/10987—reproduction in https://github.com/sidharthachatterjee/contentful-playground)
2. `resolveFluid` was incorrectly filtering out a size equal to the original size when creating `srcset`
3. `resolveFluid` would add original size even if it was already in the sizes 

Adds two tests and updates snapshots to assert this behaviour


sidharthachatterjee:
@DSchau Added tests and fixed another unrelated but colocated bug

DSchau:
Seems like this fixes the bug (and maybe another!) so excellent work.

Let's get this merged and released.

cc @sarahbethfederman

DSchau:
```
gatsby-source-contentful@2.0.26
```

Issue #10316, opened 2018-12-05
CSS Modules (Gatsby v2) - New css styles not hot reloading. Requires a page refresh 


jonlow:
## Description

When adding a new class to a css module, hot reload does not inject the css module class and no styles are shown. Manually refreshing the page will show the new css module class and the correct styles.

Once the page has been manually refreshed after creating a new class, any subsequent changes to the class will hot reload correctly.

Screencast of the issue https://www.useloom.com/share/a1061fb2d4344942af1f9efa4870d965

Note: I've checked gatsby v1 and new css modules display correctly without a page refresh

### Steps to reproduce

I can confirm this is an issue using the latest:

- Gatsby Starter Default https://github.com/gatsbyjs/gatsby-starter-default
- Using CSS Modules example https://github.com/gatsbyjs/gatsby/tree/master/examples/using-css-modules

### Expected result

New css module classes should hot reload and display on the page (after saving) without a manual page refresh

### Actual result

No styles show up on the page. A manual page refresh is required

### Environment

  System:
    OS: macOS High Sierra 10.13.6
    CPU: (4) x64 Intel(R) Core(TM) i7-4650U CPU @ 1.70GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 8.11.3 - /usr/local/bin/node
    Yarn: 1.12.3 - ~/.yarn/bin/yarn
    npm: 4.2.0 - ~/.npm-packages/bin/npm
  Browsers:
    Chrome: 70.0.3538.110
    Firefox: 63.0.3
    Safari: 12.0.1
  npmPackages:
    gatsby: ^2.0.53 => 2.0.63 
    gatsby-image: ^2.0.20 => 2.0.22 
    gatsby-plugin-manifest: ^2.0.9 => 2.0.11 
    gatsby-plugin-offline: ^2.0.16 => 2.0.18 
    gatsby-plugin-react-helmet: ^3.0.2 => 3.0.4 
    gatsby-plugin-sharp: ^2.0.14 => 2.0.14 
    gatsby-source-filesystem: ^2.0.8 => 2.0.10 
    gatsby-transformer-sharp: ^2.1.8 => 2.1.9 
  npmGlobalPackages:
    gatsby-cli: 2.4.6

jonlow:
Screencast of the issue https://www.useloom.com/share/a1061fb2d4344942af1f9efa4870d965

As you can see the title styles don't apply unless I do a manual refresh

[...]

PR #11032, opened 2019-01-13
fix(core): Disable HMR for CSS modules

jonlow:
## Description

CSS modules will not work correctly with style-loader when HMR is enabled (HMR was introduced in gatsby v2). For new styles to appear on a page, a full page refresh is required for css modules to work. Otherwise, an error is thrown. This PR disables HMR for css-modules as suggested by https://github.com/webpack-contrib/style-loader/issues/320 so that saving a `.module.scss` file will refresh the page and show any new styles. 

Here's a screencast showing the issue before the fix https://www.useloom.com/share/a1061fb2d4344942af1f9efa4870d965

## Related Issues

https://github.com/gatsbyjs/gatsby/issues/10316

### DIFF
--- a/packages/gatsby-plugin-less/src/gatsby-node.js
+++ b/packages/gatsby-plugin-less/src/gatsby-node.js
@@ -30,7 +30,7 @@ exports.onCreateWebpackConfig = (
   const lessRuleModules = {
     test: /\.module\.less$/,
     use: [
-      !isSSR && loaders.miniCssExtract(),
+      !isSSR && loaders.miniCssExtract({ hmr: false }),
       loaders.css({ ...cssLoaderOptions, modules: true, importLoaders: 2 }),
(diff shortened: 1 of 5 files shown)


KyleAMathews:
This is modifying the sass rules? I assume you meant to modify all of the CSS module rules?

jonlow:
@KyleAMathews I'm only modifying the SASS CSS modules (eg `components.module.scss`), since this code currently lives in gatsby-plugin-sass. I figured that anything specific to SASS should live in this plugin.

If we want to add this fix for regular CSS modules (no SASS), then shouldn't that fix go into the main Gatsby package? (maybe `utils/webpack-utils.js`)

wardpeet:
@jonlow sorry for keeping you waiting, i'll be checking this today! 

It would be really helpful if you do it for all cssmodules plugins like less, ...

PR #11038, opened 2019-01-13
fix(gatsby-remark-prismjs): use aliased name for highlighters

chitoku-k:
## Description

When having a configuration with an aliased name like below, `gatsby-remark-prismjs` does not handle them properly.

```js
{
  resolve: 'gatsby-transformer-remark',
  options: {
    plugins: [
      {
        resolve: 'gatsby-remark-prismjs',
        options: {
          aliases: {
            xaml: 'xml',
          },
        },
      },
    ],
  },
}
```

Currently, the highlighter uses the original name (`xaml`) but the code block is wrapped with the aliased name (`xml`).

The documentation reads:

```js
// This lets you set up language aliases.  For example,
// setting this to '{ sh: "bash" }' will let you use
// the language "sh" which will highlight using the
// bash highlighter.
```

which does not match the current behavior.

Thanks.

### DIFF
--- a/packages/gatsby-remark-prismjs/src/__tests__/__snapshots__/index.js.snap
+++ b/packages/gatsby-remark-prismjs/src/__tests__/__snapshots__/index.js.snap
@@ -154,6 +154,47 @@ Object {
 }
 `;
 
+exports[`remark prism plugin generates a <pre> tag with aliases that do not exist 1`] = `
+Object {
+  "children": Array [
+    Object {
+      "lang": "foobar",
+      "position": Position {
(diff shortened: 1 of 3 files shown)


calcsam:
Thanks @chitoku-k! Can you add a test for this case that passes one way and fails the other way?

chitoku-k:
@calcsam I added a test that ensures it doesn’t highlight an unmatched alias. Does that make sense?

chitoku-k:
Glad to hear! Thanks for getting this merged.

PR #11043, opened 2019-01-13
Add missing break statements

DanielRuf:
## Description
A switch case entry should have always a `break` statement to ensure that the logical flow is also stopped there.

### DIFF
--- a/packages/gatsby-source-contentful/fetch-data.js
+++ b/packages/gatsby-source-contentful/fetch-data.js
@@ -68,6 +68,7 @@ module.exports = function () {
             // someone develop a contentful site even if not connected to the internet.
             // For prod builds though always fail if we can't get the latest data.
             process.exit(1);
+            break;
 
           case 18:
             currentSyncData = void 0;
@@ -87,6 +88,7 @@ module.exports = function () {
 
             console.log("error fetching contentful data", _context.t1);
             process.exit(1);
+            break;
 
           case 30:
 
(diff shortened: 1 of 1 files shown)


packages/gatsby-source-contentful/fetch-data.js
```diff
@@ -68,6 +68,7 @@ module.exports = function () {
             // someone develop a contentful site even if not connected to the internet.
             // For prod builds though always fail if we can't get the latest data.
             process.exit(1);
+            break;
```
ZYSzys: This line would never be executed since process exited before, right ?

packages/gatsby-source-contentful/fetch-data.js
```diff
@@ -87,6 +88,7 @@ module.exports = function () {
 
             console.log("error fetching contentful data", _context.t1);
             process.exit(1);
+            break;
```
ZYSzys: Ditto

packages/gatsby-source-contentful/fetch-data.js
DanielRuf: Not really. Our lint check told us that there should be always a break or return.

wardpeet: @DanielRuf @ZYSzys is right, when running process.exit the process stops which means break isn't needed. It might be best practise to be consistent with breaks. A better way is to use process.exitCode = 1; and let the node process gracefully stop but atm this isn't possible with the current setup.

PR #11044, opened 2019-01-13
chore: remove redundant code

DanielRuf:
## Description
This PR removes some unneeded checks and removes an unreachable code in an `else` which is never reached.


DanielRuf:
> Production pathPrefix navigation can go back:
>     CypressError: Timed out retrying: expected 'blank' to equal '/blog/'

Is this related to my changes or just a flaky test?

DSchau:
@DanielRuf see #9233. Unfortunately it fails very intermittently, and it's a bit hard to trace down.

If you have any interest--we'd love a second set of 👀 to take a look!

PR #11048, opened 2019-01-13
fix: reevaluate page query when page context is modified

### LINKED ISSUE #6097: Update page context
Hi,

I created a system that creates pages automatically via json files.

In `createPages` i have a graphql query to find all pages i want to create and respective json content
But when i change the json files the page context is not updated.
I even tried to call `deletePage` and `createPage` again with the new context and the page is not getting updated.

It seems like the `public/static/d/path--index-**hash**.json` is not being updated, so even restarting the dev server do not work, i have to delete the files in `public/static/d` to be able to get the new context.

I am using gatsby v2, but this was already giving me the same trouble before

RomanGotsiy:
## Description

When page context is updated by `createPage` the path is not marked as dirty and page query is not reevaluated resulting in the page not updating in develop mode.

## Related Issues

Fixes #6097

### DIFF
--- a/packages/gatsby/src/internal-plugins/query-runner/page-query-runner.js
+++ b/packages/gatsby/src/internal-plugins/query-runner/page-query-runner.js
@@ -15,7 +15,6 @@ const queue = require(`./query-queue`)
 const { store, emitter } = require(`../../redux`)
 
 let queuedDirtyActions = []
-let queuedDirtyPathnames = []
 
 let active = false
 let running = false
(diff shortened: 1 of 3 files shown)


packages/gatsby/src/internal-plugins/query-runner/page-query-runner.js
```diff
@@ -65,6 +70,12 @@ emitter.on(`DELETE_NODE`, action => {
   queuedDirtyActions.push({ payload: action.payload })
 })
 
+emitter.on(`CREATE_PAGE`, action => {
+  if (action.contextModified) {
+    queuedDirtyPathnames.push(action.payload.path)
```
pieh: I think we can just reuse `runQueriesForPathnamesQueue` Map instead of creating new variable? We don't track why exactly query for pathname is dirty, so it should be safe to do

RomanGotsiy: Good suggestion! Thanks!
Addressed

Issue #11037, opened 2019-01-13
[Regression] createRemoteFileNode is not able to guess filename from temporary urls


oorestisime:
(Sorry for not respecting the template)

So #9777 introduced a regression when the It tries to guess the remote file name on urls that do not contain the filename. Such urls are temporary download urls. I guess this is acceptable

**But** when those urls are somewhat long (>255 characters) then it will try to save the file under that name and fail since it exceeds the maximum characters allowed in a filename (255 on ext4).

**Proposal:** same thing with the extension, the `createRemoteFileNode` could take an optional parameter that enforces the name of the file.

If this is ok I am willing to make the necessary changes here.

(How i came here https://github.com/oorestisime/gatsby-source-dropbox/issues/2)

hbish:
I think it makes sense to have an override as there are lots of different use case of this utility method and there is no way to cater for each and every use case. Thanks for proposing this @oorestisime

PR #11054, opened 2019-01-13
feat(gatsby-source-filesystem): add optional name parameter to createRemoteFileNode

oorestisime:
Fix: #11037

## Description

Take an optional name parameter to fix issue with temporary urls having no "guessable" name

### DIFF
--- a/packages/gatsby-source-filesystem/README.md
+++ b/packages/gatsby-source-filesystem/README.md
@@ -227,9 +227,9 @@ exports.downloadMediaFiles = ({
 
 The file node can then be queried using GraphQL. See an example of this in the [gatsby-source-wordpress README](/packages/gatsby-source-wordpress/#image-processing) where downloaded images are queried using [gatsby-transformer-sharp](/packages/gatsby-transformer-sharp/) to use in the component [gatsby-image](/packages/gatsby-image/).
 
-#### Retrieving the remote file extension
+#### Retrieving the remote file name and extension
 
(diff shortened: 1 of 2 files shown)


oorestisime:
> Thank you for this, @oorestisime
> 
> Would you like to also add a test or two to assert this behaviour and ensure we don't see a regression again? slightly_smiling_face

I would really like to do so but for this particular package it is always hard, at least for me. There are a lot of mocks to do and basically i am not sure how do to do so (plus there are not many in place to get inspired). If you have an example in mind i can work on it probably.

sidharthachatterjee:
@oorestisime So in `packages/gatsby-source-filesystem/src/__tests__/create-remote-file-node.js` you could add a test asserting `overrides name if passed in` 

In this you would pass in a `name` and a url and assert that createNode got called with the right one

Check out the documentation at https://jestjs.io/docs/en/mock-functions for how to listen on arguments a mock fn is called with

oorestisime:
But until now these tests were not entering here https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-source-filesystem/src/create-remote-file-node.js#L180-L262

and there i need to mock `requestRemoteNode` etc. what if i change `getRemoteFileName` to take as parameters a url and optionnaly a name. so that i could indeed unit-test only that part?

DSchau:
```
Successfully published:
 - gatsby-source-filesystem@2.0.17
```

PR #11087, opened 2019-01-15
fix(gatsby-source-contentful): checking if entryItemFieldValue[0] exists

brosapinto:
## Description

On the _source-contentful plugin_, an error is being thrown in the `createContentTypeNodes` function when `entryItemFieldValue` is an empty array. This PR adds a condition to check whether it's safe to access that prop or not, to prevent that exception.

/cc @Khaledgarbaya @KyleAMathews @pieh

### DIFF
--- a/packages/gatsby-source-contentful/src/normalize.js
+++ b/packages/gatsby-source-contentful/src/normalize.js
@@ -297,6 +297,7 @@ exports.createContentTypeNodes = ({
           const entryItemFieldValue = entryItemFields[entryItemFieldKey]
           if (Array.isArray(entryItemFieldValue)) {
             if (
+              entryItemFieldValue[0] &&
               entryItemFieldValue[0].sys &&
               entryItemFieldValue[0].sys.type &&
               entryItemFieldValue[0].sys.id
--- a/packages/gatsby-source-contentful/src/__tests__/__snapshots__/normalize.js.snap
+++ b/packages/gatsby-source-contentful/src/__tests__/__snapshots__/normalize.js.snap
@@ -362,9 +362,7 @@ Our Lemnos products are made carefully by our craftsmen finely honed skillful te
           },
         },
         "phone": Object {
-          "en-US": Array [
-            "+1 212 260 2269",
-          ],
+          "en-US": Array [],
         },
         "website": Object {
           "en-US": "http://www.lemnos.jp/en/",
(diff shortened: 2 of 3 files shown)


packages/gatsby-source-contentful/src/normalize.js
```diff
@@ -297,6 +297,7 @@ exports.createContentTypeNodes = ({
           const entryItemFieldValue = entryItemFields[entryItemFieldKey]
           if (Array.isArray(entryItemFieldValue)) {
             if (
+              entryItemFieldValue[0] &&
```
Khaledgarbaya: What do you think about moving this check a bit up?

`if (Array.isArray(entryItemFieldValue) && entryItemFieldValue.length > 0) {`

wardpeet: This sounds great! you can even make it shorter by omitting the `> 0`

`if (Array.isArray(entryItemFieldValue) && entryItemFieldValue.length) {`

wardpeet:
@brosapinto would it be possible to add a unit test for it so we won't regress 😄

packages/gatsby-source-contentful/src/normalize.js
brosapinto: Yeah, I did that initially. But if `entryItemFieldValue` is indeed an array, does it make sense to jump to the `else if` below?

I was going for the smallest change possible, otherwise maybe we could make this block a little DRYier:

```js
const value = Array.isArray(entryItemFieldValue)
  ? entryItemFieldValue
  : [entryItemFieldValue]

value
  .filter(v => _.has(v, 'sys.id'))
  // Don't create link to an unresolvable field.
  .filter(v => resolvable.has(v.sys.id)) 
  .forEach(v => {
    if (!foreignReferenceMap[v.sys.id]) {
      foreignReferenceMap[v.sys.id] = []
    }

    foreignReferenceMap[v.sys.id].push({
      name: `${contentTypeItemId}___NODE`,
      id: entryItem.sys.id,
    })
  })
```

brosapinto:
Sure @wardpeet, I'll try adding some assertions 👍

packages/gatsby-source-contentful/src/normalize.js
Khaledgarbaya: Actually, I am wrong @brosapinto, sorry about that I totally missed the else if

brosapinto:
@Khaledgarbaya and @wardpeet, I changed the mock data to include an example where `entryItemFieldValue` can be an empty array, which blows up before this fix. What do you guys think?

Khaledgarbaya:
Hey @brosapinto,
How about adding the two cases of an array with data and an empty array, this way we cover both cases

brosapinto:
`data.json` already has examples of array with data and now i've added/changed an entry to simulate an empty array, @Khaledgarbaya. So, I think it covers both cases? What do you suggest?

wardpeet:
if @Khaledgarbaya approves than this one is good to go. Thank you for fixing!

PR #11088, opened 2019-01-15
Fix sort field compatibility with `graphql-tools`

### LINKED ISSUE #10709: Make gatsby compatible with graphql-tools last version
I just found an issue when graphql-tools is updated on last version 
gatsby do not compile
Error:
Variable "$_v0_sort" got invalid value for desc -> do you mean DESC ?

graphql-tools need uppercase

to resolve this i changed in dist/schema/create-sort-field.js
 value 'asc' to 'ASC'
and value 'desc' to 'DESC'

now compile good
```
module.exports = function createSortField(typeName, fieldNames) {
....
  return {
....
          name: _.camelCase(`${typeName} sortOrder`),
          defaultValue: `ASC`,
          type: new GraphQLEnumType({
            name: _.camelCase(`${typeName} sortOrderValues`),
            values: {
              ASC: {
                value: `ASC`
              },
              DESC: {
                value: `DESC`
              }
...}
```

itaisteinherz:
## Description

This uppercases the sort field's sort enum values.

## Related Issues

This fixes #10709 (or should, at least 😅).

### DIFF
--- a/packages/gatsby/src/schema/create-sort-field.js
+++ b/packages/gatsby/src/schema/create-sort-field.js
@@ -37,8 +37,8 @@ module.exports = function createSortField(
           type: new GraphQLEnumType({
             name: _.camelCase(`${typeName} sortOrderValues`),
             values: {
-              ASC: { value: `asc` },
-              DESC: { value: `desc` },
+              ASC: { value: `ASC` },
+              DESC: { value: `DESC` },
             },
           }),
         },


wardpeet:
@itaisteinherz thank you for this PR but I'm pretty sure it's not the root problem as this is a graphql Enum which just forces specific values.

PR #7062, opened 2018-08-06
plugin checker initial commit

ovhemert:
Initial plugin checker for issue #5211

It reports found plugins to "plugins.json".
Blacklist property can be manually set to avoid processing again.
Notifying => see issue

Issue #5211, opened 2018-05-01
Add script to find Gatsby plugins w/o the "gatsby-plugin" keyword and add comment to repo suggesting they add it


KyleAMathews:
Spin off from discussion in https://github.com/gatsbyjs/gatsby/pull/5180

@calcsam noticed that there's several dozen more Gatsby plugins that aren't showing up in the plugin search because they lack the `gatsby-plugin` keyword in their package.json.

We should write a script which would search for NPM plugins that mention `gatsby-source`, `gatsby-plugin`, and `gatsby-transformer` in the text and then compare that our existing keyword search (also excluding deprecated packages) and find those packages without our keywords. This will still include a number of false positives so the script should include a way to "blacklist" spurious results (and save these in the repo so they won't show up in the future) — probably with https://www.npmjs.com/package/inquirer#checkbox---type-checkbox — and then the remaining packages, the script will automatically create a comment in the github repo suggesting they add the `gatsby-plugin` keyword and linking to the docs on this.

cricrio:
I'll be happy to work on it.

alexluong:
It's been a while ago. Are you still working on this @cricrio? I'd love to help if you stop or need help.

cricrio:
Hi! I started to work on it but I didn't find the time to finish. There is what I have done.
I am using the npms api to find the package containing the keywords in the description but not in the keyword section.
If you want to have look on the api docs: https://api-docs.npms.io/

```js
const axios = require(`axios`);

const keywords = ['gatsby-plugin', 'gatsby-source', 'gatsby-transformer'];

function filterDuplicate([head, ...rest], result) {
	if (!head) {
		return result;
	} else {
		result.push(head);
		return filterDuplicate(rest.filter((r) => r.package.name != head.package.name), result);
	}
}

function mergeResults(results) {
	return results.reduce((t, r) => [...t, ...r], []);
}

const promises = keywords.map((keyword) =>
	axios
	.get(`https://api.npms.io/v2/search?q=${keyword}+keywords:-gatsby-plugin+not:deprecated&size=250`)
	.then((responses) => responses.data.results)
);

exports.fetchRepository = Promise.all(promises)
	.then(mergeResults)
	.then((r) => {
		console.log(r.length);
		return r;
	})
	.then((results) => filterDuplicate(results, []))
```

alexluong:
@cricrio Thanks for the head start. I'll try and tackle this.

KyleAMathews:
[...]

PR #11093, opened 2019-01-16
feat(gatsby-plugin-checker): add more filters to filter out bad packages

reese:
## Description

Updating the gatsby-plugin-checker. It now filters out packages that lack a repository, have bad repo links, incorrect name formats, or lack READMEs. It also created a README in the folder to explain functionality and future features (such as automated notifications).

## Related Issues

This PR stems from issue #5211 . Note that the script still doesn't automatically create issues yet. As mentioned after PR #7062, this requires a Github API key, so the best solution would likely be to put the key in an ENV variable and then run it locally. I'll leave this for a later PR.


scripts/gatsby-plugin-checker/index.js
```diff
@@ -64,6 +64,60 @@ const filterNotNotified = (packages, plugins) => {
   return packages.filter(pkg => notified.indexOf(pkg.name) < 0)
 }
 
+const removePackagesWithoutRepository = packages =>
+  packages.filter(p => hasRepository(p))
```
wardpeet: There is no need to create another function as it's only used once.

```suggestion
  packages.filter(package => !!package.links.repository)
```

scripts/gatsby-plugin-checker/index.js
```diff
@@ -64,6 +64,60 @@ const filterNotNotified = (packages, plugins) => {
...
+
+const hasRepository = packageToCheck => {
+  if (packageToCheck.links.repository) {
+    return true
+  }
+  return false
+}
+
+const removePackagesWithBadRepoLinks = async packages => {
```
wardpeet: I'm not sure if we should remove these packages. The plugins still work and are on npm. It's not required to have a github repo.

scripts/gatsby-plugin-checker/index.js
```diff
@@ -64,6 +64,60 @@ const filterNotNotified = (packages, plugins) => {
...
+  if (response.statusCode == 404) return false
+  return true
+}
+
+const removeBadNameFormats = packages => packages.filter(p => hasGoodName(p))
+
+const hasGoodName = pkg => {
+  const name = pkg.name
+
+  if (name.indexOf(`/`) !== -1) {
```
wardpeet: Let's rewrite this function to reduce its complexity (less nested ifs)

```js
const isScopedPackage = name.startsWith('@');
if (!isScopedPackage) {
  return startsWithAllowedPrefix(name);
}

const nameWithoutScope = name.slice(0, name.indexOf('/')
return startsWithAllowedPrefix(nameWithoutScope)
```

scripts/gatsby-plugin-checker/index.js
```diff
@@ -64,6 +64,60 @@ const filterNotNotified = (packages, plugins) => {
...
+    const nameWithoutScope = name.split(`/`, 2)[1]
+    if (startsWithAllowedPrefix(nameWithoutScope)) return true
+  } else {
+    if (startsWithAllowedPrefix(name)) return true
+  }
+  return false
+}
+
+const startsWithAllowedPrefix = name => {
```
wardpeet:
```js
const startsWithAllowedPrefix = name => keywords.some(keyword => name.startsWith(keyword))
```

scripts/gatsby-plugin-checker/index.js
```diff
@@ -64,6 +64,60 @@ const filterNotNotified = (packages, plugins) => {
...
+    if (name.indexOf(keyword) === 0) isGoodName = true
+  })
+  return isGoodName
+}
+
+const removePackagesWithoutReadme = packages =>
+  packages.filter(p => hasReadMe(p))
+
+const hasReadMe = pkg => {
+  if (pkg.links.homepage || pkg.readme) return true
```
wardpeet: we should try to do some extra work here to check if README.md is found in the repository if it has one like yarn does: https://github.com/yarnpkg/website/blob/master/js/src/lib/Details/index.js#L122-L131 we can just check for a 404 like you do with the git repository

LekoArts:
Hi @reese 	👋 	
Do you have time to address Ward's comments? Would be awesome! Thanks 😊

reese:
Yes, I'll get back to these once I get a chance! Sorry for the delay, hopefully will get to it in the next few days.

scripts/gatsby-plugin-checker/index.js
reese: I just basically did a check for a 404 by checking if it was in the root folder of the project, not sure if there's a better way to do it, but I figure pretty much everyone puts the README in the root directory.

wardpeet:
@reese I'm so sorry to let you do all this work but I think it was all for nothing. Initially I thought we were using this on www but we're not using this package anywhere. We use gatsby-source-npm-package-search.

Maybe you're using it and have a use case you want to share?

reese:
@wardpeet This was a script that was requested in Issue #5211. The goal initially was to have this script available for manual usage. The idea is that since not all repos mention the `gatsby-plugin` keyword in their `package.json`, this script would be a way to collect those repos so that people could reach out to the maintainers and have them add the keyword if they wanted to appear on the Gatsby site. Eventually this would all be automated. That said, this issue might be stale by now, the original issue was closed so there may not be a need for it any more.

wardpeet:
thanks @reese I was just going to circle back to you. I talked with @pieh and he said the same thing as you mentioned here in the comment so forgive me 😉. I'm testing it and let's merge this afterwards! 😄

Issue #8714, opened 2018-10-02
[feat]: add gatsby-source-sql


DSchau:
## Summary

It'd be awesome to be able to (at build time!) inject content from a SQL database, just like we do for other source types, e.g. filesystem, Contentful, etc.

### Motivation

- People may have useful data stored away in SQL that could be exposed to enable helpful things, e.g. documentation, useful data, etc.
- The plugin will need to connect to a SQL database, and then pre-load queries (e.g. pass the SQL queries directly to the plugin?)

### Further Detail

The general idea is that the source plugin will need to be able to connect to a SQL database (MySQL, MSSQL, ... ?; happy to start small and grow it in the future!) and query that data to inject into the GraphQL content layer for further processing via Gatsby.

Consider [mysql](https://github.com/mysqljs/mysql), [mssql](https://www.npmjs.com/package/mssql) for various means to inject data

See other source plugins for more info on how to build this out, e.g. [gatsby-source-contentful](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-source-contentful), [gatsby-source-wordpress](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-source-wordpress), etc.

daubaris:
me me me! Could you also direct on where to start? :)

DSchau:
@daubaris 👋 Absolutely! So I'd recommend checking out one of the source plugins I linked, and kinda re-tracing from there!

So high level, this is what this will look like:

- Create/copy a directory in packages/gatsby-source-sql
- Implement the appropriate Node APIs (probably `sourceNodes`)
- Open up a PR (early/WIP is fine, just mark it as such!) and we'll be happy to give some feedback

If you're interested, we'd love to help out with a pairing session, as well. Sign up for those [here](https://calendly.com/gatsbyjs/pair-programming)

And let me know if I can help with this further :)

garetmckinley:
If I were to suggest sqlite support as well, would that be lumped into this plugin or it's own thing?

DSchau:
@garetmckinley good question! I'm not sure how to best resolve that complexity to be honest with you.

_Ideally_ there'd be one shared sql plugin, but the more I think about it, I'm not sure how tenable that is because the _engine_ will have to change between MySQL, mssql, SQLLite, etc.

So we may be able to introduce some shared helpers, but I think this plugin may take additional plugins, and those plugins should implement some API perhaps?

DSchau:
[...]

PR #11100, opened 2019-01-16
feat(gatsby-source-sql): add new plugin to handle sql as a source

mrfunnyshoes:
## Description

Add a gatsby source plugin to get data from SQL databases.

## Related Issues

Addresses #8714


wardpeet:
@mrfunnyshoes I have no idea how we missed this. Sorry for the delay! 🙏.

Is there like any demo database we can use to test this on? If not I'll just create one my own.

sidharthachatterjee:
Hey @mrfunnyshoes 

First of all, this is really great work! Thank you for taking the time to build this and we apologise for not being able to take a look at this earlier. 

While this is certainly a super useful source plugin, whether this should _live_ in the gatsby repo or not is a subject that we've discussed. The conclusion we came to is that we would prefer to not include this in the core repo because:
- It's a large and complex plugin that will be hard to maintain 
- There are simpler alternatives to GraphQL-izing SQL databases (like Prisma and Hasura for instance)

Because of the above, I'll be closing this for now but urge you to publish this in your own repo for everyone that would undeniably find this useful. 

Please know that we really value having you as a contributor and look forward to having you on board for more PRs in the future! Thank you once again 💪 💜

muescha:
FYI:
if someone search for this - the plugin is here available:

https://github.com/edgarrmondragon/gatsby-source-sql

PR #11112, opened 2019-01-16
fix(plugin-google-gtag): make pluginConfig optional

### LINKED ISSUE #11111: [gatsby-plugin-google-gtag] no pluginConfig causes crashing
## Description

The gtag plugin crashes if `pluginConfig` is not defined, even though its fields are in turn documented as optional. The code carefully treats all of the fields as optional too ([exclude](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-google-gtag/src/gatsby-ssr.js#L23), [head](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-google-gtag/src/gatsby-ssr.js#L30)). And there is definitely a use case for the plugin without any of them, so to avoid unnecessary config bloat, it should be optional. Probably just an oversight.

### Steps to reproduce

The following `gatsby-config` makes it crash:
```javascript
module.exports = {
  plugins: [
    {
      resolve: `gatsby-plugin-google-gtag`,
      options: {
        trackingIds: [`GA_TRACKING_ID`],
      },
    },
  ],
}
```

### Expected result

It shouldn't crash as all the options are optional.

### Actual result

TypeError: Cannot read property 'exclude' of undefined
at line 20, or [23](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-google-gtag/src/gatsby-ssr.js#L23) non-transpiled.

### Environment

  System:
    OS: macOS 10.14.2
    CPU: (8) x64 Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.15.0 - ~/.nvm/versions/node/v10.15.0/bin/node
    Yarn: 1.13.0 - ~/.nvm/versions/node/v10.15.0/bin/yarn
    npm: 6.5.0 - ~/.nvm/versions/node/v10.15.0/bin/npm
  Languages:
    Python: 3.6.7 - /anaconda3/bin/python
  Browsers:
    Chrome: 71.0.3578.98
    Firefox: 64.0.2
    Safari: 12.0.2
  npmPackages:
    gatsby: ^2.0.76 => 2.0.91 
    gatsby-image: ^2.0.20 => 2.0.26 
    gatsby-plugin-google-gtag: ^1.0.8 => 1.0.8 
    gatsby-plugin-manifest: ^2.0.9 => 2.0.13 
    gatsby-plugin-offline: ^2.0.16 => 2.0.21 
    gatsby-plugin-sharp: ^2.0.14 => 2.0.17 
    gatsby-source-filesystem: ^2.0.8 => 2.0.16 
    gatsby-transformer-sharp: ^2.1.8 => 2.1.10 
  npmGlobalPackages:
    gatsby-cli: 2.4.8

JacobBlomgren:
## Description

Make `pluginConfig` optional.

## Related Issues

Fixes #11111


JacobBlomgren:
@wardpeet sure, test added!

JacobBlomgren:
Aaand another commit because I screwed up the logic to run the plugin in prod and test, but not dev :flushed: The test now passes on this branch but fails on master as expected for me 😄

PR #11117, opened 2019-01-16
fix(showcase): Remove Stitch Fix from featured sites

allypalanzi:
## Description

Stitch Fix is no longer using GatsbyJS

### DIFF
--- a/docs/sites.yml
+++ b/docs/sites.yml
@@ -6,17 +6,6 @@
   categories:
     - Web Development
     - Featured
-- title: Stitch Fix
-  main_url: "https://www.stitchfix.com/"
-  url: "https://www.stitchfix.com/"
-  description: >
-    Stitch Fix is an online styling service that delivers a truly personalized
-    shopping experience.
-  categories:
-    - eCommerce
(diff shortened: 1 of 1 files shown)


allypalanzi:
I'm not sure, I can't find it!

barrymcgee:
@allypalanzi Yeah, me either.. it does seem to reflect your homepage exactly as it is right now though which makes me think it might be generated on the fly?

wardpeet:
@allypalanzi mind sharing with us in DM or email why you switched so we can improve our product based on your feedback

barrymcgee:
@wardpeet Do you mind explaining where the screenshot is?

LekoArts:
@barrymcgee https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-transformer-screenshot creates the initial image, Sharp transforms the image then: https://github.com/gatsbyjs/gatsby/blob/master/www/src/pages/showcase.js#L29

barrymcgee:
@LekoArts Thank you 🙏🏻

PR #11126, opened 2019-01-17
docs(gatsby-source-wordpress): Replace slug with pathname of link

crstnio:
## Description

The WordPress slug does not include url fragments that are not set per post.

For example:

- The permalink settings of WordPress could add `/blog/` before the slug for blog posts.
- WPML prepends the whole path with the current language like `/en/` in _folder configuration_.

This modified `gatsby-node.js` example ensures that the WordPress url structure gets preserved.
Not doing so would confuse non-technical editors that wonder why the displayed permalink in WordPress does not work.


riddla:
@cardiv, why you create an `URL` first and then use the `pathname` of it? In my tests `edge.node.link` is always the same as the `pathname` ...

cardiv:
Not in my case. If you fetch the data with Postman, there is always an absolute link in the response.

If Gatsby removes the domain using the `searchAndReplaceContentUrls` option, it does not work for me because I rewrite it in WordPress to match the Gatsby site domain.

Well, `URL` should be rewritten to handle relative paths too. I didn't expect `link` could be relative.

riddla:
Good point! I totally forgot that I added a relative path plugin at some point ... 🤦‍♂️
Thank you for the clarification :)

riddla:
PS: That kind of Wordpress plugin could „save“ your step when you rewrite the URLs in Wordpress.

cardiv:
I do that mainly for permalinks so that an editor can click on the permalink for a WordPress post/page and will be directed to the Gatsby equivalent.

But for the REST API, relative paths are much better. I agree with you.

riddla:
Oh, the editor topic makes totally sense. 
I’m curious (and sorry for hijacking this PR): did you try to get a preview functionality up and running? I know it is possible for frameworks like NextJS etc. that are backed by a Node/Express server in production.

wardpeet:
Hey cardiv! thanks for creating this PR! Any chance you could give us a reproduction? It's hard to keep up with wordpress, contentful & any other headless cms so helping us understand is always a big PLUS.

cardiv:
@wardpeet I will try to find some spare time for this.

No problem @riddla. Not much going on here.

Yes, I worked on a preview. But it is highly experimental and not tested. I created an API endpoint for that. The workflow is as follows:

Editor clicks on normal WordPress preview button during post edit. This button is altered with `preview` (post id) and `token` GET parameters. I wrote some kind of wrapper function, that takes the `node` object before it gets passed to page rendering in Gatsby.

This function looks for those parameters and sends them back to WordPress (including the `node` object). The endpoint validates the token (gets refreshed every time an editor opens the post to edit or expires in one day), fetches the requested post after validation and sends the altered `node` object back to Gatsby.

In theory, this could work for existing pages. But I have not thought about new pages at the moment.

wardpeet:
I read the pr thoroughly and i'm unsure why we use slug instead of the link inside wordpress. So I need some other people to comment on this PR.

cc @KyleAMathews @gatsbyjs/ecosystem

riddla:
@wardpeet, the change from slug to link worker for my case also. I found this PR after I implemented it myself. 

It is especially needed for subpages like „about/team“. The slug always remains „team“. Only the link seems to contain parent structures, if given.

packages/gatsby-source-wordpress/README.md
```diff
@@ -773,9 +773,11 @@ exports.createPages = ({ graphql, actions }) => {
         // Create Page pages.
         const pageTemplate = path.resolve("./src/templates/page.js")
         // We want to create a detailed page for each
-        // page node. We'll just use the WordPress Slug for the slug.
+        // page node. We'll use the WordPress link for the slug to preserve url structure.
         // The Page ID is prefixed with 'PAGE_'
         _.each(result.data.allWordpressPage.edges, edge => {
+          const link = new URL(edge.node.link)
```
pieh: we need to add `link` field to graphql query few lines above (and few lines below) and can also delete `slug` field from query

pieh:
This looks good - aparat from small change (adjusting graphql queries) this looks good to go.

I'm thinking that this is common feature that people would want to use, so we should also consider adding `pathname` field directly in `gatsby-source-wordpress` plugin so users don't need to process `link` field themselves (but let's just update queries here and merge this)

riddla:
@cardiv: do you have time for the proposed changes? Otherwise I could help ...

cardiv:
I don't have much time now. Your PR is very much appreciated, thanks! Will update this docs change when the PR gets merged.

wardpeet:
@riddla mind fixing up the readme to match your latest PR https://github.com/gatsbyjs/gatsby/pull/11775? I synced with master but unsure what to change.

riddla:
@wardpeet: Thank you for the reminder. I altered the README.

PR #2710, opened 2017-11-01
Update SitePage schema before onPreExtractQueries

fusepilot:
Addresses #2685.

It looks like there was an attempt to fix this problem before judging by these lines:
https://github.com/gatsbyjs/gatsby/blob/3691d65bccc0ccb5ca487c36adf13416ee7993cc/packages/gatsby/src/bootstrap/index.js#L363-L367

This PR simply moves that second `require('../schema')()` call to before `onPreExtractQueries`. And that seems to fix it.

Of course, I tried this *after* attempting to update the old schema in this [branch](https://github.com/fusepilot/gatsby/tree/topics/site-page-querying). So thats started in case I've missed something and we need that in between "SitePage schema only" update.

Issue #2685, opened 2017-10-31
GraphQL Error on pages that query conditionally added SitePage fields


fusepilot:
Repro: https://github.com/fusepilot/gatsby-page-query-conditional-field-repro
Possibly Related: https://github.com/gatsbyjs/gatsby/issues/1718

I'm adding a custom field using createNodeField to certain pages of my site. I get a compile time GraphQL error unless I add the field to *all* pages. Everything works fine at runtime.

````
GraphQL Error Unknown field `fields` on type `SitePage`

  file: /Users/michaeldelaney/Downloads/gatsby-starter-default-master 2/src/pages/index.js

   1 |
   2 |   query IndexQuery {
   3 |     allSitePage {
   4 |       edges {
   5 |         node {
>  6 |           fields {
     |           ^
   7 |             foo
   8 |           }
   9 |         }
  10 |       }
  11 |     }
  12 |   }
  13 |
````

The relevant code:

````
// gatsby-node.js

// define a field on certain pages
function conditionalCreateNodeField({ node, createNodeField }) {
  if (node.path == '/page-a/') {
    createNodeField({
      node,
      name: 'foo',
      value: 'bar a',
    })
  }
}

// define a field on all pages
function unconditionalCreateNodeField({ node, createNodeField }) {
  createNodeField({
    node,
    name: 'foo',
    value: 'bar',
  })
}

exports.onCreateNode = async ({ node, boundActionCreators }) => {
  const { createNodeField } = boundActionCreators

  if (node.internal.type == 'SitePage') {
    conditionalCreateNodeField({ node, createNodeField }) // throws GraphQL Error
    // unconditionalCreateNodeField({ node, createNodeField }) // no error
  }
}
````

````
// index.js

export const pageQuery = graphql`
  query IndexQuery {
    allSitePage {
      edges {
        node {
          fields {
            foo
          }
        }
      }
    }
  }
`
````

KyleAMathews:
In Gatsby v1, `SitePages` wasn't intended to be queried as part of building a site — It was only added as a debugging tool — e.g. a quick way to introspect on some core data. Since launching v1 there's been several people giving feedback about why this would be super useful to have e.g. for auto-generating menus, etc.

But there needs to be some internal changes first for this to work namely regenerating the GraphQL schema for pages after pages are added (which happens after the first time the GraphQL schema is generated).

If this is something you'd like to take on, would love to get you going on a PR adding this!

fusepilot:
[...]

PR #1268, opened 2017-06-27
[1.0] Add to pages docs on debugging

KyleAMathews:

PR #11131, opened 2019-01-17
Remove schema updating from bootstrap

stefanprobst:
Currently we infer a GraphQL schema twice during bootstrap. This PR proposes to remove schema updating.

# Reasoning

Schema updating was originally intended for debugging (#1268), and later in #2710 to solve a very specific problem (#2685): adding custom fields to `SitePage` nodes with `createNodeField`, depending on whether a `path` condition is met. Things to note:
* custom fields added unconditionally to SitePage end up in the schema without the need for schema updating
* adding custom fields wrapped in a conditional like below did not work, because a page with that `path` does not yet exist in the store when the schema is created, so there is nothing to infer.
```js
exports.onCreateNode = ({ actions, node }) => {
  if (node.internal.type === `SitePage`) {
    if (node.path === `/path-2/`) {
      const { createNodeField } = actions
      createNodeField({
        node,
        name: `foo`,
        value: `bar`,
      })
    }
  }
}
```

While infering the schema again after all pages have been created does solve this problem, the far less expensive option would have been to always create the field, but fill in the value conditionally. The following works without the need for schema updating:
```js
exports.onCreateNode = ({ actions, node }) => {
  if (node.internal.type === `SitePage`) {
    const { createNodeField } = actions
    createNodeField({
      node,
      name: `foo`,
      value: node.path === `/path-2/` ? `bar` : '',
    })
  }
}
```

Other than this usecase, is there any other reason for schema updating? Fields on page `context` would be missing, but they are available as query variables anyway.

# Issues

Because schema creation is not run twice, type name postfixes will differ. I tried to adjust the counting in [`create-type-name`](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/schema/create-type-name.js), which unfortunately did not work consistently. For a default starter, take a look at the schema diff [here](https://pastebin.com/Ag8kw6cS).

### DIFF
--- a/packages/gatsby/src/bootstrap/index.js
+++ b/packages/gatsby/src/bootstrap/index.js
@@ -386,6 +386,8 @@ module.exports = async (args: BootstrapArgs) => {
   await require(`../schema`).build({ parentSpan: activity.span })
   activity.end()
 
+  require(`../schema/type-conflict-reporter`).printConflicts()
+
   // Collect resolvable extensions and attach to program.
   const extensions = [`.mjs`, `.js`, `.jsx`, `.wasm`, `.json`]
   // Change to this being an action and plugins implement `onPreBootstrap`
@@ -440,16 +442,6 @@ module.exports = async (args: BootstrapArgs) => {
   activity.start()
   await apiRunnerNode(`onPreExtractQueries`, { parentSpan: activity.span })
   activity.end()
-
-  // Update Schema for SitePage.
-  activity = report.activityTimer(`update schema`, {
-    parentSpan: bootstrapSpan,
-  })
-  activity.start()
-  await require(`../schema`).build({ parentSpan: activity.span })
-  activity.end()
-
-  require(`../schema/type-conflict-reporter`).printConflicts()
 
   // Extract queries
   activity = report.activityTimer(`extract queries from components`, {


KyleAMathews:
People use the page schema for creating menus etc. So they'll add data via pageContext and then query the resulting schema. So we could remove support for this but it would be a breaking change.

freiksenet:
@KyleAMathews any examples in the wild where this feature is used? I want to think about possible workarounds.

pieh:
Even if there will be workarounds (and I'm sure there will be), we can't make breaking changes like that (at least during v2).

Can you add some context on why you want to remove it?

KyleAMathews:
I don't know for sure what sites use this just that people talk about it. One example could be you add metadata to each page e.g. it's section and page title and then when generating a navigation, you query for pages within a section and sort by title.

stefanprobst:
SitePage *is* actually already created in the first schema build step (and also includes all fields added with createNodeField), what is missing is anything that was added to page context in createPages. I was not aware that people used this other than from the query variables in page queries, thanks @KyleAMathews for the clarification!

freiksenet:
Right, so stuff is passed to page context and *that* is inferred as GraphQL types.

Is there any reason to actually query pageContext from GraphQL? I know it's a breaking change, but I feel maybe we can change it to be a list of key-value pairs and thus eliminate the need of regenerating the schema.

freiksenet:
My main concern here is that we are getting a "bag of fields" type for that context, because practically every page (or realistically every page kind), will have their own context, but one would be able to query for all of those fields for any of the pages. It should either be an actual dict type (so list of key/val pairs) or a union of all possible page contexts.

freiksenet:
Ok, so this would need to wait to Gatsby 3 cause this is a breaking change. I'll summarize motivations here, it can become an RFC.

# Why is this important?

The gatsby build structure can be divided into 4 steps. 

1. Collect nodes
2. Create schema
3. Generate pages
4. Render pages

Currently 3) can feed into 2) by defining additional page context. This means that we have a "cycle" in our logic, with pages optionally affecting the schema. Not only it's confusing directly (schema before 3 and after 3 is different), it also causes weird naming issues (all things being someGeneratedType_2 because we generated it twice). In addition this makes doing incremental and/or distributed builds harder, because schema is tied to page generation. Thus ideally we would have a directly data flow, with pages not affecting the schema. (well technically we have pages feed into the nodes when they are created, but we can consider it to be a special case).

# How to minimize changes?

Currently you can query page context through all the possible fields every page context has. So if you added fields `foo` and `bar` to different pages, your page query would be smth like that:

```graphql
{
  allSitePage {
    edges {
      node {
        context {
          foo
          bar
        }
      }
    }
  }
}
```

I propose to instead have page context in GraphQL be a list of `key` and `value` fields. The query would look as follows:

```graphql
{
  allSitePage {
    edges {
      node {
        context {
          key
          value
        }
      }
    }
  }
}
```

This minimizes the changes that user have to make to adopt their website.

freiksenet:
Create RFC here https://github.com/gatsbyjs/rfcs/pull/26

freiksenet:
Closing this because of schema refactor.

Issue #10912, opened 2019-01-08
gatsby build does not call componentWillMount()


cpboyd:
## Description

The redux state for the current page doesn't seem to be cleared from a prior rendering or updated by `componentWillMount()` before being rendered.

Possibly due to asynchronous rendering?

### Steps to reproduce

1. Have a site with multiple pages using a single, global Redux in `wrapRootElement()`
2. Modify a redux state parameter utilized by layout in `componentWillMount()`
3. Layout will represent the prior rendered page's state.

### Expected result

State changes implemented in `componentWillMount()` should take effect BEFORE the page is rendered.

Ideally, each page seems like it should be rendered from the initial redux state as well without any potential prior path issues.

### Actual result

Pages render with incorrect redux state.

### Environment

```
  System:
    OS: macOS 10.14.2
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 11.2.0 - /usr/local/bin/node
    Yarn: 1.12.3 - /usr/local/bin/yarn
    npm: 6.5.0 - /usr/local/bin/npm
  Languages:
    Python: 2.7.10 - /usr/bin/python
  Browsers:
    Chrome: 71.0.3578.98
    Safari: 12.0.2
  npmPackages:
    gatsby: ^2.0.40 => 2.0.85 
    gatsby-image: ^2.0.19 => 2.0.25 
    gatsby-plugin-google-analytics: ^2.0.7 => 2.0.8 
    gatsby-plugin-layout: ^1.0.8 => 1.0.11 
    gatsby-plugin-manifest: ^2.0.8 => 2.0.13 
    gatsby-plugin-offline: ^2.0.17 => 2.0.21 
    gatsby-plugin-react-helmet: ^3.0.1 => 3.0.5 
    gatsby-plugin-remove-serviceworker: ^1.0.0 => 1.0.0 
    gatsby-plugin-sharp: ^2.0.12 => 2.0.17 
    gatsby-plugin-sitemap: ^2.0.2 => 2.0.4 
    gatsby-plugin-typescript: ^2.0.1 => 2.0.3 
    gatsby-remark-embed-video: ^1.4.0 => 1.7.0 
    gatsby-remark-external-links: ^0.0.4 => 0.0.4 
    gatsby-remark-images: ^2.0.6 => 2.0.6 
[...]

PR #11134, opened 2019-01-17
fix(using-redux): create a fresh store for each SSR page

cpboyd:
## Description

As I'd mentioned in #10912, the current [using-redux](https://github.com/gatsbyjs/gatsby/tree/master/examples/using-redux) example creates a static store that persists across multiple SSR pages.

It seems that the **expected** behavior should likely be that each page's initial load is from a fresh store, in order to match the behavior of the browser rendering.

## Related Issues

Related to #10912

### DIFF
--- a/examples/using-redux/wrap-with-provider.js
+++ b/examples/using-redux/wrap-with-provider.js
@@ -7,5 +7,5 @@ import createStore from "./src/state/createStore"
 export default ({ element }) => {
   // Create a fresh store for each SSR page:
   const store = createStore()
-  return (<Provider store={store}>{element}</Provider>)
+  return <Provider store={store}>{element}</Provider>
 }
--- a/examples/using-redux/wrap-with-provider.js
+++ b/examples/using-redux/wrap-with-provider.js
@@ -5,7 +5,9 @@ import createStore from "./src/state/createStore"
 
 // eslint-disable-next-line react/display-name,react/prop-types
 export default ({ element }) => {
-  // Create a fresh store for each SSR page:
+  // Instantiating store in `wrapRootElement` handler ensures:
+  //  - there is fresh store for each SSR page
+  //  - it will be called only once in browser, when React mounts
   const store = createStore()
   return <Provider store={store}>{element}</Provider>
 }
--- a/examples/using-redux/wrap-with-provider.js
+++ b/examples/using-redux/wrap-with-provider.js
@@ -3,7 +3,9 @@ import { Provider } from "react-redux"
 
 import createStore from "./src/state/createStore"
 
-const store = createStore()
-
 // eslint-disable-next-line react/display-name,react/prop-types
-export default ({ element }) => <Provider store={store}>{element}</Provider>
+export default ({ element }) => {
+  // Create a fresh store for each SSR page:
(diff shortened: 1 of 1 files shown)


wardpeet:
Could you explain:

> It seems that the expected behavior should likely be that each page's initial load is from a fresh store, in order to match the behavior of the browser rendering.

Creating a new store for each page is something you don't really want when gatsby re-hydrates your page as the redux store should be shared across pages.

cpboyd:
@wardpeet Yes, the redux store should be shared across pages once the page has been loaded, but not for **every initial rendering of every single page** during `gatsby build`.

Currently, as mentioned in #10912, each page gets rendered retaining the prior state since the store is created static and thus retained between each initial rendering in the `gatsby build` process.

cpboyd:
From #10912, based on the "using-redux" example from this repo:
https://github.com/cpboyd/gatsby-redux-bug

With `gatsby develop`, all pages initially display a `1` if they're the first page loaded.

However, `gatsby build` results in the following being rendered (before Javascript browser rendering takes over):

0. A
1. B
2. C
3. Home

Note: I'd initially tried using `componentWillMount()`, which apparently doesn't get called during SSR in React.  Moving the increment to `render()` instead should exhibit similar behavior.

cpboyd:
A few new remarks:
Redux's help indicates SSR **should** create a new store for each page, but the client is static:
https://redux.js.org/recipes/server-rendering#serverjs-1

As a result, maybe we should just go back to having separate functions for `gatsby-browser.js` (which would match the prior `wrapWithProvider` and `gatsby-ssr.js` which would create a new store for each rendered page.

Further, even placing redux state changes in the `constructor()` doesn't necessarily result in the state change being dispatched **BEFORE rendering** occurs.

As a result, it might be worth updating the example to show pages with differing initial states.

e.g., I'm currently using:
```javascript
const getStateFor = (pathname)  => {
  if (!pathname) {
    return {};
  }
...<computation here>
  return {
    ...<computed values>
  }
}

const createStore = (pathname) => {
  const pathState = getStateFor(pathname);
  return reduxCreateStore(reducer,
    {
      ...initialState,
      ...pathState
    }, composeWithDevTools(applyMiddleware()));
}
```

pieh:
the `wrapRootElement` is called just once in gatsby-browser (just before initial render), so this change is safe to do, but we could adjust comment there a bit, to make it clear that it will cause fresh store instance for each SSR page, but it will only be called once in browser

examples/using-redux/wrap-with-provider.js
```diff
@@ -3,7 +3,9 @@ import { Provider } from "react-redux"
 
 import createStore from "./src/state/createStore"
 
-const store = createStore()
-
 // eslint-disable-next-line react/display-name,react/prop-types
-export default ({ element }) => <Provider store={store}>{element}</Provider>
+export default ({ element }) => {
+  // Create a fresh store for each SSR page:
```
pieh: maybe this would work?
```suggestion
  // Instantiating store in `wrapRootElement` handler ensures:
  //  - there is fresh store for each SSR page
  //  - it will be called only once in browser, when React mounts
```

PR #11141, opened 2019-01-18
Cache resolved nodes during bootstrap

stefanprobst:
Currently, we only cache resolved nodes in a production build. It *should* be safe to also cache in develop during bootstrap, since no new nodes should be created during the run-queries phase.


pieh:
There probably are scenarios where some plugins can insert/modify/delete nodes - i.e. gatsby-source-filesystem that adds and removes nodes on its own and it can actually happen mid query running - we don't handle those scenarios that well currently, but it's not safe assumption to make that nodes won't be added between end of sourceNodes and bootstrap finishing

pieh:
good thing to do would be listen to `CREATE_NODE`/`DELETE_NODE` events and invalidate caches for given node types

stefanprobst:
Makes sense, thanks!

PR #11145, opened 2019-01-18
chore(*): add directory details to the package.json of all packages

greysteil:
Specifying the directory as part of the `repository` field in a `package.json` allows npm and third party tools to provide better support when working with monorepos. For example, it allows them to correctly construct a commit diff for a specific package.

This format was accepted by npm in https://github.com/npm/rfcs/pull/19.

### DIFF
--- a/packages/gatsby-cli/package.json
+++ b/packages/gatsby-cli/package.json
@@ -56,7 +56,11 @@
   ],
   "license": "MIT",
   "main": "lib/index.js",
-  "repository": "https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-cli",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/gatsbyjs/gatsby.git",
+    "directory": "packages/gatsby-cli"
+  },
   "scripts": {
     "build": "babel src --out-dir lib --ignore **/__tests__",
     "prepare": "cross-env NODE_ENV=production npm run build",
--- a/packages/babel-plugin-remove-graphql-queries/package.json
+++ b/packages/babel-plugin-remove-graphql-queries/package.json
@@ -14,6 +14,11 @@
     "gatsby": "^2.0.0",
     "graphql": "^14.1.1"
   },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/gatsbyjs/gatsby.git",
+    "directory": "packages/babel-plugin-remove-graphql-queries"
+  },
   "scripts": {
     "build": "babel src --out-dir . --ignore **/__tests__",
     "prepare": "cross-env NODE_ENV=production npm run build",
--- a/packages/babel-preset-gatsby-package/package.json
+++ b/packages/babel-preset-gatsby-package/package.json
@@ -11,6 +11,11 @@
     "@babel/preset-flow": "^7.0.0",
     "@babel/preset-react": "^7.0.0"
   },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/gatsbyjs/gatsby.git",
+    "directory": "packages/babel-preset-gatsby-package"
+  },
   "license": "MIT",
   "main": "index.js"
 }
--- a/packages/babel-preset-gatsby/package.json
+++ b/packages/babel-preset-gatsby/package.json
@@ -15,6 +15,11 @@
   },
   "license": "MIT",
   "main": "index.js",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/gatsbyjs/gatsby.git",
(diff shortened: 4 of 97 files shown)


wardpeet:
Hey @greysteil. I'm guessing this is not backwards compatible with yarn & older npm versions? If so we're not able to merge this at this time as we don't accept breaking changes.

greysteil:
@wardpeet what do you mean by older versions? I'm not sure how this change could break any `yarn` or `npm` versions, and it's already been adopted by the folks at react, blueprint and jest without any problems.

As I understand it, currently the repository links here aren't usable directly by git, so they *may* currently be breaking very old versions of `npm` and `yarn` (although I doubt it).

packages/gatsby-cli/package.json
```diff
@@ -46,7 +46,11 @@
   ],
   "license": "MIT",
   "main": "lib/index.js",
-  "repository": "https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-cli",
```
DSchau: So as @wardpeet noted, this is something we'll want to do (eventually?), but I do think this is _maybe_ a worse experience in two scenarios:

1. Browsing on npmjs.org and using the repository link, e.g. [gatsby-cli](https://www.npmjs.com/package/gatsby-cli)
1. Using `npm repo <repo-name>`, e.g. `npm repo gatsby-cli`

In both of these scenarios, the root of the repo will open, rather than the specific package.

greysteil: Agreed - until npm improves their links to support the new format properly the hack is probably a better experience.

Maybe worth leaving this open, but with a comment that support for https://github.com/npm/cli/pull/140, and an implementation of the npmjs.org change (I believe that repo is private), would be awesome.

m-allanson:
As @greysteil [noted](
https://github.com/gatsbyjs/gatsby/pull/11145#discussion_r), let's merge this once https://github.com/npm/cli/pull/140 is in, and npmjs.com supports the new directory field.

greysteil:
Just realised that we could unblock this by updating all of the URLs in it to be the same as you currently have (and keeping the directory details too). Would you be 👍 if I make that change?

DSchau:
@greysteil would that actually work?

Maybe I'm misunderstanding, but given `gatsby-source-filesystem` (as an example), the URL would have to be `https://github.com/gatsbyjs/gatsby.git` and the directory would then be `packages/gatsby-source-filesystem` right?

But... the directory keyword wouldn't work with the [existing URL](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-source-filesystem/package.json#L41), right?

Happy to be corrected here--more a clarifying question so I understand than anything else! Thanks!

greysteil:
@DSchau - depends how it's implemented. It wouldn't be that hard to split the directory out of existing URLs (we have logic to do that in Dependabot [here](https://github.com/dependabot/dependabot-core/blob/master/common/lib/dependabot/source.rb#L5-L24) already that I could port to npm).

Thinking about it, though, I think it probably makes more sense to work. I'm hoping to have PRs for npm and yarn in the next couple of days, so this should be mergeable as-is before to long.

sidharthachatterjee:
Looks like https://github.com/npm/cli/pull/140 is merged in. Time to rebase this and merge it in!

sidharthachatterjee:
Rebased, fixed conflicts and added directory details to the missing packages as well!

sidharthachatterjee:
Waiting on https://npmjs.com to actually work with the new repository field

KyleAMathews:
@sidharthachatterjee it won't hurt anything right to merge this now? It's a new field so will just be meaningless data until new NPM clients and npmjs.com support.

sidharthachatterjee:
@KyleAMathews Not a new field, it's the existing `repository` field but with a nested structure as opposed to a plain `String`

Current behaviour:

https://www.npmjs.com/package/gatsby-cli points to https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-cli

Behaviour after this (until https://npmjs.com updates this):
https://www.npmjs.com/package/gatsby-cli points to https://github.com/gatsbyjs/gatsby

Also will affect `npm repo package-name`

KyleAMathews:
Hmmm ok — how do we know when npmjs.com supports this?

sidharthachatterjee:
I guess we keep clicking on the link in https://www.npmjs.com/package/react-cache till the day it links to https://github.com/facebook/react/tree/master/packages/react-cache 😅

freiksenet:
Still not working on npm :D

moonmeister:
replaced by #15477

greysteil:
No worries - glad to have this done! 😄

PR #11146, opened 2019-01-18
refactor: convert default-html to a function component

alexandernanberg:
## Description

Refactor the `<HTML />` component into a function component instead of a class component.

## Related Issues
 N/A


pieh:
> Why do you think that this is necessary?

There's not a ton of benefit, but functional components result in smaller output (classes are more verbose). So for any stateless components (until hooks are stable) it's just micro optimisation.

alexandernanberg:
Yeah as @pieh said – there isn't really

PR #11148, opened 2019-01-18
[showcase] Add Ashley Thouret and The AZOOR Society

ooloth:
Adds two recent Gatsby sites to the Showcase: an opera singer and a health-related charity.


ooloth:
@LekoArts Thanks! I'm aware of that warning and confused by it since all references to `cdnjs.com` in the codebase are to `https://cdnjs.cloudflare.com`. 

I've triple-checked, so I'm not sure why this warning persists. 🤷‍♂️

PR #11149, opened 2019-01-18
fix(gatsby): Resolve all fields before querying with Sift

stefanprobst:
Fix `extractFieldsToSift` silently omitting fields that need to be resolved before querying. See:
```js
const _ = require('lodash')
const diff = require('jest-diff')

const filter = {
  foo: { eq: true },
  bar: {
    foo: { eq: true },
    bar: {
      foo: { eq: true },
      bar: {
        foo: { eq: true },
      },
    },
  },
  baz: { eq: true },
  qux: {
    foo: { eq: true },
  },
}

const expected = {
  foo: true,
  bar: {
    foo: true,
    bar: {
      foo: true,
      bar: {
        foo: true,
      },
    },
  },
  baz: true,
  qux: {
    foo: true,
  },
}

function extractFieldsToSift(prekey, key, preobj, obj, val) {
  if (_.isPlainObject(val)) {
    _.forEach(val, (v, k) => {
      if (k === `elemMatch`) {
        extractFieldsToSift(prekey, key, preobj, obj, v)
        return
      }
      preobj[prekey] = obj
      extractFieldsToSift(key, k, obj, {}, v)
    })
  } else {
    preobj[prekey] = true
  }
}

const fieldsToSift = {}
_.each(filter, (v, k) => {
  extractFieldsToSift(``, k, {}, fieldsToSift, v)
})

console.log(diff(fieldsToSift, expected))

```

### DIFF
--- a/packages/gatsby/src/redux/__tests__/run-sift.js
+++ b/packages/gatsby/src/redux/__tests__/run-sift.js
@@ -25,6 +25,19 @@ const mockNodes = [
     id: `id_3`,
     string: `baz`,
   },
+  {
+    id: `id_4`,
+    string: `qux`,
+    first: {
+      willBeResolved: `willBeResolved`,
+      second: {
+        willBeResolved: `willBeResolved`,
+        third: {
+          foo: `foo`,
+        },
+      },
+    },
+  },
 ]
 
 jest.mock(`../../db/nodes`, () => {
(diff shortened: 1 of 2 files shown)


stefanprobst:
Oh, Node 6 does not have `Object.entries`. Will fix tomorrow.

freiksenet:
Thanks!

PR #11163, opened 2019-01-19
Fix JSON in `.babelrc` example.

peterp:
Thank you for this project!

## Description

The example `.babelrc` configuration was not valid JSON, I've converted the keys to strings.


jgierer12:
`.babelrc` is JSON5 which allows for unquoted keys.

peterp:
Colour me enlightened.

PR #11164, opened 2019-01-19
Pass-through query from graphql function

jgierer12:
I have a use case where I need to pass through the value of the `graphql` function's query for use in node. Specifically, I want to define a fragment that I can use both in page queries and in `gatsby-node.js`:

<details>
<summary>Details</summary>

```js
// src/fragments/foo.js

import { graphql } from "gatsby";

export const query = graphql`
  fragment Foo on Bar {
    // ...
  }
`;
```

```js
// src/pages/index.js

import { graphql } from "gatsby";

export const query = graphql`
  query {
    ...Foo
  }
`;
```

```js
// gatsby-node.js

require(`@babel/register`);
const fooFragment = require(`./src/fragments/foo`);

exports.createPages = async ({ actions, graphql }) => {
  // ...

  const result = await graphql(`
    ${fooFragment.query}
    
    query {
      ...Foo
    }
  `)

  // ...
};
```

</details>
<br />

Since the `graphql` function throws an error when actually called, regardless of environment, I can't use it in `gatsby-node.js`. This PR changes this so that you can opt-in via an env variable to let the `graphql` function simply pass through its query instead of throwing. This enables the use case described above as well as potential use cases of having access to the query in the browser.

### DIFF
--- a/packages/gatsby/cache-dir/gatsby-browser-entry.js
+++ b/packages/gatsby/cache-dir/gatsby-browser-entry.js
@@ -36,7 +36,11 @@ StaticQuery.propTypes = {
   children: PropTypes.func,
 }
 
-function graphql() {
+function graphql(query) {
+  if (process.env.GATSBY_GRAPHQL_PASSTHROUGH) {
+    return query
+  }
+
   throw new Error(
     `It appears like Gatsby is misconfigured. Gatsby related \`graphql\` calls ` +
       `are supposed to only be evaluated at compile time, and then compiled away,. ` +


pieh:
Would auto injecting fragments to `gatsby-node` graphql queries (same as we do in page queries / static queries) solve the problem? I really don't like proposed change (different fragments handling based on env)

jgierer12:
Yes in my case that would also do the trick. But I believe passing through the query has other use cases that won't be solved by that. And my proposal has nothing to do with fragments, it simply passes through any string parameter of the `graphql` function when it's called, instead of throwing an error. And since it's opt-in through an env variable it doesn't break anything.

DSchau:
@jgierer12 that may very well be the case, but I think @pieh's comment here was a good one.

While this behavior isn't a breaking change, it is another code path to maintain--albeit a small one, and we'd generally like to minimize the amounts of these small tweaks to support some not necessarily general-purpose use cases.

Could you provide more insight into this statement?

> But I believe passing through the query has other use cases that won't be solved by that

Personally - I'd prefer we go the route of making fragments available in `gatsby-node.js` (even if that's a bit more work) because it feels more correct and less like a band-aid to fix a semi-related problem.

jgierer12:
> But I believe passing through the query has other use cases that won't be solved by that

I'm primarily thinking about 'isomorphic queries' here, queries that can be used during both build time and run time without having to write them twice. I came across this wile experimenting with running a Gatsby server as a live GraphQL API that lets you filter and search the content dynamically (you may remember I posted a RFC about this a while ago).

One of the biggest annoyances with that is that you have to write every query twice. A quick code example:

```js
import React from "react";
import { graphql } from "gatsby";
import { request } from "graphql-request";

const query = graphql`
  query($foo: String) {
    myData(foo: $foo) {
      something
    }
  }
`;

// Exactly the same as `query`!
// But since `gatsby#graphql` doesn't pass through the contents of the
// query, we need to duplicate it.
const rtQuery = `
  query($foo: String) {
    myData(foo: $foo) {
      something
    }
  }
`;

export default function MyComponent({ data: initialData, pageContext }) {
  const [data, setData] = React.useState(initialData);
  const [foo, setFoo] = React.useState(pageContext.foo);

  const requestData = async () => {
    const newData = await request(`https://example.com`, rtQuery, { foo });
    setData(newData);
  };

  return (
    <>
      <input type="text" value={foo} onChange={e => setName(e.target.value)} />
      <button onClick={requestData}>Request Data</button>
      <pre>{JSON.stringify(data)}</pre>
    </>
  );
}
```

As you can imagine this can get very lenghty very quickly. With this PR, we could just pass `query` to `request` instead of having to duplicate it.

I personally see a huge potential in the possibility of an 'isomorphic' Gatsby env, where the data is available at build time AND run time. Being able to make fully dynamic apps while losing none of the advantages of prerendering. It would open up Gatsby for an entire new kind of apps, ones where data querying based on run time parameters is needed.

But for that to work without a bad Developer Experience, something like this change needs to happen.

I also know that this is far from the perfect solution. Of course, the best thing would be to write the query as a string, and being able to pass it to gatsby's `graphql` as a function parameter (which as you know is a much requested feature but of course it's is a very difficult thing to implement). In the mean time I think something like this behind a env flag would suffice.

> While this behavior isn't a breaking change, it is another code path to maintain--albeit a small one, and we'd generally like to minimize the amounts of these small tweaks to support some not necessarily general-purpose use cases.

That is completely reasonable, and I understand if you reject this PR. But I think that it would be a step into the right direction to make this isomorphic Gatsby environment more viable.

jgierer12:
By the way, if you want to see a semi-production example of such an app, I am building https://anthem-gear.com ([Source](https://github.com/jgierer12/anthem-gear)). Search is currently the only dynamic aspect but I'm working on filtering as I speak. Note that I also had to make two separate queries for the fragments (`src/fragments/equipment.js`, `src/fragments/equipment-runtime.js`) and listing pages (`src/pages/list.js`, `src/templates/list.js`).

pieh:
I wonder if better way wouldn't be adding support for writing queries in x.graphql files and importing them in components and just straight reading x.graphql file content in gatsby-node?

pieh:
The code you pasted with `graphql-request` wouldn't also work with pass through flag - we remove/replace graphql function calls in babel plugin so `query` becomes just identifier - see:

fixures with code
https://github.com/gatsbyjs/gatsby/blob/a82682c01ae223752a64239124599a51436ad1a4/packages/babel-plugin-remove-graphql-queries/src/__tests__/index.js#L190-L204

become:
https://github.com/gatsbyjs/gatsby/blob/master/packages/babel-plugin-remove-graphql-queries/src/__tests__/__snapshots__/index.js.snap#L114-L123

queries are getting removed

jgierer12:
Hmm, that's actually right. In that case I will close this PR and think about other ways to solve this problem. I will update the RFC with my thoughts.

I still think your original suggestion of auto injecting fragments to `gatsby-node` graphql queries would be a good thing to implement, and solve the original reason I opened the PR. I might work on this too, although I'm not 100% sure where to start with that.

PR #11166, opened 2019-01-19
fix(gatsby-transformer-sharp): Pick extension from file prop

stefanprobst:
The resolvers for `srcSetWebp` and `srcWebp` check the extension on `image` (which is the ImageSharp node that has no `extension` prop) instead of `file` (which is the parent File node).


freiksenet:
Thanks!

Issue #11014, opened 2019-01-12
gatsby-plugin-google-tagmanager and Google Search Console Verification


stmcallister:
Has anyone had issues getting Google Search Console to verify ownership of your site with Google Tag Manager using `gatsby-plugin-google-tagmanager`? 

We're using the plugin, and our analytics are flowing wonderfully. But, I'm being told by the person who is trying to add our site in Google Search Console that GTM is not verifying. The only difference between the gatsby plugin code and the code they want me to use is this:  `'https://www.googletagmanager.com/gtm.js?id='+i+dl+''` and `'https://www.googletagmanager.com/gtm.js?id='+i+dl`. The first one as an extra `+''` at the end...which isn't functionally different. 

So, has anyone else using the plugin had issues verifying their sites?

cardiv:
Hey @stmcallister,

thank you for raising this issue.

Does the requested changes to GTM code work adding the site to Search Console?

If so, would you mind creating a PR for that?

maxlapides:
I tested the idea of removing the `+''` and it unfortunately didn't resolve the issue for me.

stmcallister:
Do you get any kind of error message?

maxlapides:
"We could not find any Google Tag Manager container ID on the index page of your site.”

sidharthachatterjee:
@stmcallister @maxlapides 

`gatsby-plugin-google-tagmanager` adds the script recommended in https://developers.google.com/tag-manager/quickstart at https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-google-tagmanager/src/gatsby-ssr.js

Can you please link to a minimal reproduction of this if possible?

stmcallister:
Turns out the culprit is the `iframe` code snippet just inside `<body>`.  The plugin is generating line breaks to make the `iframe` code more readable, like so:

`<iframe
  src="https://www.googletagmanager.com/ns.html?id=GTM-xxxxxx"
  height="0"
  width="0"
  style="display: none; visibility: hidden"
></iframe>`

This does not allow Google Search Console to verify ownership of the site. However, if you remove the line breaks, and make the `iframe` code look like the code below GSC is happy and verifies site ownership.

`<iframe src="https://www.googletagmanager.com/ns.html?id=GTM-xxxxxx" height="0" width="0" style="display: none; visibility: hidden"></iframe>`

My pull request above removes the line breaks in the iframe code.

wardpeet:
[...]

PR #11169, opened 2019-01-20
fixed format of preBodyComponents to work with Google Search Console …

stmcallister:
When the line breaks are present in the iframe tag, Google Search Console is not able to verify ownership of the site. This pull request removes the line breaks so that the iframe tag is all on a single line.

This is response to issue #11014.

### DIFF
--- a/packages/gatsby-plugin-google-tagmanager/src/gatsby-ssr.js
+++ b/packages/gatsby-plugin-google-tagmanager/src/gatsby-ssr.js
@@ -37,14 +37,12 @@ exports.onRenderBody = (
         key="plugin-google-tagmanager"
         dangerouslySetInnerHTML={{
           __html: stripIndent`
-            <iframe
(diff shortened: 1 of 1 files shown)


packages/gatsby-plugin-google-tagmanager/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "gatsby-plugin-google-tagmanager",
   "description": "Gatsby plugin to add google tagmanager onto a site",
-  "version": "2.0.7",
+  "version": "2.0.8",
```
pieh: Please revert that. Package version is being bumped automatically during our publishing process

DSchau:
Thanks @m-allanson for the review :)

```
Successfully published:
 - gatsby-plugin-google-tagmanager@2.0.8
 - gatsby@2.0.96
```

PR #11171, opened 2019-01-20
feat(gatsby): add support for ExportSpecifier in API files

### LINKED ISSUE #10553: Is `export { ... } from` supported in plugins?
I saw the following export declarations in a plugin [here](https://github.com/hupe1980/gatsby-i18n/blob/master/packages/gatsby-plugin-lingui/src/gatsby-node.js#L1-L6):

```
export {
  onPreBootstrap,
  onCreateNode,
  onPreExtractQueries,
  onCreatePage,
} from '@wapps/gatsby-i18n/plugin';
```

But is this case covered in [`bootstrap/resolve-module-exports.js`](https://github.com/gatsbyjs/gatsby/blob/cdddb29f2f1eb805da7eafa0bc8dadbdb4899561/packages/gatsby/src/bootstrap/resolve-module-exports.js#L65)?

I couldn't get it work in my `gatsby-ssr.js` when using:

```
export { wrapPageElement } from "./wrap-page-element";
export { wrapRootElement } from "./wrap-root-element";
```

**EDIT:** Surprisingly these exports in my `gatsby-node.js` worked (despite not getting picked up in `load-plugins/validate.js` [here](https://github.com/gatsbyjs/gatsby/blob/cdddb29f2f1eb805da7eafa0bc8dadbdb4899561/packages/gatsby/src/bootstrap/load-plugins/validate.js#L128-L130))

```
export { onCreateBabelConfig } from "./on-create-babel-config";
export { onCreatePage } from "./on-create-page";
```

harryparkdotio:
## Description

This PR allows named exports to be used in
- `gatsby-node`
- `gatsby-ssr`
- `gatsby-browser`

### *example*

**`some-random-file.js`**
```js
export const wrapRootElement = ({ element }) => element
```

**`gatsby-ssr.js`**
```js
export { wrapRootElement } from './some-random-file'
```

## Related Issues

Fixes #10553

### DIFF
--- a/packages/gatsby/src/bootstrap/__tests__/resolve-module-exports.js
+++ b/packages/gatsby/src/bootstrap/__tests__/resolve-module-exports.js
@@ -111,6 +111,10 @@ describe(`Resolve module exports`, () => {
       exports.__esModule = true;
       exports.foo = '';
     `,
+    "/export/named": `const foo = ''; export { foo };`,
+    "/export/named/from": `export { Component } from 'react';`,
+    "/export/named/as": `const foo = ''; export { foo as bar };`,
+    "/export/named/multiple": `const foo = ''; const bar = ''; const baz = ''; export { foo, bar, baz };`,
   }
 
   beforeEach(() => {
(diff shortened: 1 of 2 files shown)


pieh:
This is awesome! 

Not sure about `gatsby-node` here given that's Node - I guess there is a way to make it work, but ESM won't work out of the box with Node. That's not to take away of this improvement! It's still awesome and gatsby-browser and -ssr works out of the box!

Issue #11120, opened 2019-01-17
Use Playground instead of GraphiQl


Saifadin:
## Summary

Gatsby should use [Prisma GraphQl Playground](https://github.com/prisma/graphql-playground) instead of GraphiQl for its Query Browser.

### Motivation

After using the Prisma Playground for the last months, I can not go back to GraphiQl. To just list the pros/cons and after being motivated by @KyleAMathews on twitter to create an issue. I will try to be objective and focus on the benefits for Gatsby

#### Pros
- 🗂Tabs to save prior queries
- 📖Query Browser easier to navigate back and forth
- ⚙️Customisation Options (for the editor)
- ⚛️Maintained by one of the biggest GraphQl providers for React and recommended by Apollo
- 🔏Tracing
- 🌑Dark Mode 

#### Cons
- New URL needs to be implemented, maybe only with a major version upgrade or both in parallel (/___playground)
- People got used to GraphiQl

If you have any own Pros or Cons, please add them here. Maybe they will help the Gatsby make a more educated decision. If good ones are mentioned, I might edit the above list.

**EDIT:** I went ahead and created [this plugin](https://github.com/Saifadin/gatsby-plugin-playground) that will add a route at `/___playground` with the Prisma Playground. 😄

polarathene:
I haven't tried the playground, is it that much different to get used to from GraphiQL as your cons suggest? Does it not support something you can do with GraphiQL, or work/behave in a very different way to do similar tasks?

glauberfc:
I think that's a good change

Huvik:
Hey :), it would be nice to have `graphql-playground` bundled with Gatsby. But first of all `playground` is made by [Prisma](https://github.com/prisma). And if you want some new feature inside [playground](https://github.com/prisma/graphql-playground/) please create an issue. Playground supports theming and there could be even `Gatsby` theme, so it would fit more to whole Gatsby family.

Saifadin:
@Huvik you are totally right, we are using it with Apollo and mixed them up haha
I will fix that.

sidharthachatterjee:
@Saifadin I could see this being a plugin!

Should be easy to set up a new route in express [onCreateDevServer](https://www.gatsbyjs.org/docs/node-apis/#onCreateDevServer)

Saifadin:
@sidharthachatterjee that is also a possible approach, that can be used in the meanwhile, as I think this would be at least minor version jump or even a major one.
[...]

PR #11193, opened 2019-01-21
feat(gatsby): GraphQL playground

Saifadin:
## Description

A PR that adds the Prisma Playground to the Gatsby Build process alongside GraphiQL.

Example Image

## Related Issues

#11120 

## TO-Dos

- [x] Install `graphql-playground-middleware-express`
- [x] Add `/___playground` route
- [x] Add Gatsby Theme
- [x] Add to documentation


Saifadin:
@sidharthachatterjee @Huvik here it is, will still add the Theme and the documentation. My colleague is making a contrast-rich theme design, that I will add, so it can be tweaked.

Saifadin:
@LekoArts yeah I was thinking the same, that is why I kept GraphiQL on `/___graphql` and added `/___playground` or where you referring to the console.log? That one should mention both of them I think.

Saifadin:
Maybe like this?
```js
    console.log()
    console.log(`Or use Playgound, to explore your site's data and schema`)
    console.log()
    console.log(`  ${urls.localUrlForTerminal}___playground`)
```

LekoArts:
No, what I meant is: Do a complete switch over to Playground but keep the same URL. That wouldn't be a breaking change IMO as just the interface changes. You can remove GraphiQL. At least that is what @freiksenet said.

Saifadin:
Ahh got it, will do that 👍🏽

Saifadin:
@LekoArts done

Saifadin:
Added a Custom Theme for Gatsby, what do you guys think of this? We tried to stick to the Gatsby colors.

screenshot 2019-01-21 at 18 16 17

Huvik:
@Saifadin It looks nice :) You can even change editor color scheme. And I see there is an old playground inside. There is right now hardcoded bad version of the playground, but I will remove it in the new release so there is always up to date version.

Huvik:
@Saifadin Yea, there is space for improvements, but basically code editor follows basic `Codemirror` theme, so any theme working with code mirror should works. If its remapped to object. But I agree right now there is a mess I should fix it asap.

Saifadin:
@Huvik Glad to help, if you can point me to a list of main issues :)

packages/gatsby/src/commands/develop.js
```diff
@@ -361,7 +400,7 @@ module.exports = async (program: any) => {
 
     console.log()
     console.log(
-      `View GraphiQL, an in-browser IDE, to explore your site's data and schema`
+      `View the Playground, an in-browser IDE, to explore your site's data and schema`
```
glennreyes:
```suggestion
      `View the GraphQL Playground, an in-browser IDE, to explore your site's data and schema`
```

packages/gatsby/src/commands/develop.js
```diff
@@ -92,13 +94,50 @@ async function startServer(program) {
       heartbeat: 10 * 1000,
     })
   )
-  app.use(
+  app.post(
     `/___graphql`,
     graphqlHTTP({
       schema: store.getState().schema,
-      graphiql: true,
+      graphiql: false,
```
glennreyes:
```suggestion
```

packages/gatsby/src/commands/develop.js
Saifadin: 👍

packages/gatsby/src/commands/develop.js
Saifadin: Was not sure what the default was, but this is cleaner :)

sidharthachatterjee:
> Added a Custom Theme for Gatsby, what do you guys think of this? We tried to stick to the Gatsby colors.
> 
> screenshot 2019-01-21 at 18 16 17
> 
> Another option would be this a little bit darker color scheme
> 
> screenshot 2019-01-21 at 18 19 33

I think accessibility might be an issue here 

cc @fk

Saifadin:
@sidharthachatterjee can you elaborate?
The Colors in this Theme are all AAA from a color contrast.

Would love to hear your concerns in more detail.

m-allanson:
I haven't had a chance to try this out yet, but it looks great! 

I'm not 100% sold on straight up replacing GraphiQL with Playground, at least not in a minor release. I also think including GraphiQL _and_ Playground would be confusing. Could we do a gradual rollout, first by enabling Playground via an env var? 

This would let you do `GATSBY_GRAPHQL_IDE=playground gatsby develop` to use Playground right now (well, not _right now_ but as soon as this PR is released). 

Then we could switch Playground to be the default in the next major release of Gatsby, before finally removing GraphiQL in a later version. I think this would give us a nice balance of enabling the new goodness for people that want it while letting us work through any issues _and_ avoiding disrupting people's existing workflows.

Here's an example where we've [done that for the node db that Gatsby uses internally](https://github.com/gatsbyjs/gatsby/pull/9919/files/c0e9fadb6a54cd8c32fe4df3db6346cfffc7880b#diff-8c8b888e741f1c31e97d4bf05894f50d). 

### Theming

Theme choices are always going to be somewhat subjective, for this PR I'd prefer to stick with the default theme. We can then use a follow up PR to work out a better theme, as the theme is not directly related to making this PR mergeable. This is also a nice benefit of enabling Playground behind a feature flag to start with - while it's feature flagged we can let people try it out and provide feedback, and don't need to worry too much about changing things until it's un-feature-flagged.

Personally I'd like a theme that's very similar to the existing GraphiQL theme. There are a _lot_ of guides and tutorials about Gatsby that have GraphiQL screenshots in them, anything we can do to avoid them looking outdated is going to be an overall win for the community. But I'd be happy to be convinced that a light theme is a bad idea :)

Saifadin:
Concerning the gradual release, I kind of agree. It is too big a change for a minor release. A sudden change might be surprising, especially for people not closely following the news etc, so adding it behind an environment variables is a great way.

Basically these steps:
1. Opt-In through environment variable and mentioning this possibility
2. Deprecating GraphiQL in the next major release, while giving an opt-in possibility to it and updating docs with new images
3. Next major release after that, remove it

### Theming
If we use a gradual release, then I would put the custom theme aside for now and use the default.

Concerning light vs dark theme I think it subjective, but we can add a note in the docs, that it is changeable in the editor settings in the playground, by changing `"editor.theme": "dark"` to `"editor.theme": "light"`.
If it is behind a feature flag, adding a dark theme is I think less of a problem. The moment we make it the default option, it might be necessary to make this discussion in detail.

So for now I am removing the custom theme from this PR, but I will be waiting for other feedback from the other Reviewers to the feature-flag discussion 😄

LekoArts:
> So for now I am removing the custom theme from this PR, but I will be waiting for other feedback from the other Reviewers to the feature-flag discussion

Yeah, sorry for the confusion. What Mike said is probably the best solution to this

fk:
Great to see this happening @Saifadin!

Re: the theme — I love that you put in the work for a dark Gatsby theme (and took care about accessibility; I briefly checked the (blurry) JPGs and only saw `AA` contrast values for the two or three things I measured)!

That said, @m-allanson's comments regarding theming and the general plan to move forward make a lot of sense to me — there's enough work to take care of in terms of UX which does _not_ concern the visual appearance. It makes sense to me to move incrementally — get the foundation done first, and then add the "theme" icing-on-the-cake in a separate step. This unblocks this PR from discussions regarding the latter.

I'd personally love to have both light and dark themes available.
Your working regarding the dark theme is a great foundation to build upon and I like to think it wasn't made in vain!

Saifadin:
Thanks a lot @fk ^__^

When we created the theme in Sketch it showed us `AAA`, but we would need to tweak it when we work on the theme in the future, for which I will create a second PR after this one is accepted.

Like I said, light and dark themes exist out of the box, which could also be indicated by an ENV variable maybe?

@m-allanson and the rest: If there are no **big** blockers, I will start implementing an env based solution today.

fk:
@Saifadin 🤗 

>  I briefly checked the (blurry) JPGs and only saw AA contrast values for the two or three things I measured)

That sentence didn't come out correctly, and I forgot to mention that I think that I see `AAA` is due to the blurryness of the image. Sorry for making it sound like I wanted to say "the `AAA` contrast values you talked about aren't there" — that really wasn't my intention at all! 🙏 

Super cool to hear that you went through the effort of using Sketch to design the theme!

Saifadin:
All good ^^
Not me, but my UI designer colleague did it together with me. He is ready to rework it in the later PR.^^

packages/gatsby/src/commands/develop.js
```diff
@@ -92,13 +94,19 @@ async function startServer(program) {
       heartbeat: 10 * 1000,
     })
   )
-  app.use(
+  app.post(
```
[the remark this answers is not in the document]
Saifadin: I was not aware of that, good catch.
But then I am confused about how I can use the same Url for two `get`s. Once for playground UI and for the `GET` requests.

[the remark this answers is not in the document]
Saifadin: Can you have another look, it should work now

Saifadin:
Added to the documentation

What is the status here?

Huvik:
@Saifadin There is a new version of the playground, could you please upgrade middleware version? 😇

pieh:
> What is the status here?

Need some time to verify the change locally. Code looks good.

Saifadin:
@Huvik done :)
@pieh Got it 👍

docs/docs/using-graphql-playground.md
```diff
@@ -0,0 +1,23 @@
...
+## What is The Prisma Playground?
+
+The Playground is a way for you to interact with the data your sources and plugins add as schemas. You will be interacting with this data a lot and the Playground will help you greatly with exploring this data.
+
+## Accessing the Playground
+
+As this is an experimental feature, you need to access it by adding an environmental variable, you need to add `GATSBY_GRAPHQL_IDE` to your start script, like this:
+
+```
+"start": "GATSBY_GRAPHQL_IDE=playground npm run develop",
```
pieh: maybe we should add this to `develop` script (as this script actually assume that `develop` is a script)?
```suggestion
"develop": "GATSBY_GRAPHQL_IDE=playground gatsby develop",
```
?

pieh: But it's not a blocker I think, we can iterate on docs. But feature itself was tested and work so I'll go ahead and merge and publish

Saifadin: Good catch, committed that

docs/docs/using-graphql-playground.md
```diff
@@ -0,0 +1,23 @@
...
+
+In this page we want to introduce you to an alternative to the current IDE for your GraphQl queries: The [Prisma Playground](https://github.com/prisma/graphql-playground).
+
+## What is The Prisma Playground?
+
+The Playground is a way for you to interact with the data your sources and plugins add as schemas. You will be interacting with this data a lot and the Playground will help you greatly with exploring this data.
+
+## Accessing the Playground
+
+As this is an experimental feature, you need to access it by adding an environmental variable, you need to add `GATSBY_GRAPHQL_IDE` to your start script, like this:
```
pieh: I guess we also need to adjust this (given you merged my previous suggestion):
```suggestion
As this is an experimental feature, you need to access it by adding an environmental variable, you need to add `GATSBY_GRAPHQL_IDE` to your `develop` script, like this:
```

Saifadin: Very true, done!

PR #11202, opened 2019-01-22
fix(gatsby): Fix input filter tests

stefanprobst:
The input filter InputObjectType is on a `filter` field. This adjusts the tests to check the correct type


freiksenet:
Thanks!

PR #11207, opened 2019-01-22
fix(docs): change bound-action-creators links into actions links

jbolda:
## Description

The panic still points to the old page. Updating it to point to the new page.


wardpeet:
do you mind changing all `/docs/bound-action-creators` into `/docs/actions`

jbolda:
It would be a while before I get to it. I just did this one quick through the website UI.

PR #11210, opened 2019-01-22
docs: Document internal resource handling + SWs

davidbailey00:


docs/docs/resource-handling-and-service-workers.md
```diff
@@ -0,0 +1,71 @@
...
+
+Here is how the `EnsureResources` component handles each of these scenarios:
+
+1. On non-initial renders, reload upon missing resources so that the browser can load the server's 404 page
+
+2. On non-initial renders, reload upon missing resources so that the browser can load the latest page
+
+   On initial renders, flag the page as failed and throw an error to prevent rendering a blank page (static HTML will suffice), then reload it once the service worker updates
+
+3. On initial renders, throw an error to prevent rendering a blank page (static HTML will suffice)
```
m-allanson: Should this line be removed in favour of the previous line?

davidbailey00: This applies to 2 and 3 which is why it's duplicated - I thought it would be clearest to write it twice so that I can keep the numbered list format, although maybe it'll be clearer that it's intentional if I swap the order of the points under number 2.

PR #11213, opened 2019-01-22
feat(gatsby): add server instance on onCreateDevServer

pedronauck:
## Description

This pull request just adds the `server` instance on `onCreateDevServer` hook. This can be helpful if you need to reuse the server to create something (eg. another WebSocket instance).

### DIFF
--- a/packages/gatsby/src/commands/develop.js
+++ b/packages/gatsby/src/commands/develop.js
@@ -171,8 +171,6 @@ async function startServer(program) {
     })
   }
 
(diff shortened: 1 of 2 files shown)


wardpeet:
@pedronauck Why do you need the server and not the express instance?

DSchau:
Hey @pedronauck circling back to this--is this still something that should be merged in?

If so--could you clarify why exactly this is valuable?

Thanks!

DSchau:
Going to close this out--if this is something we need to build in, happy to re-open this, just let us know!

Thanks!

PR #11218, opened 2019-01-22
feat(gatsby-remark-prismjs): Add class when using highlighted lines

JasonEtco:
## Description

Hiya! This PR adds an additional class to `.gatsby-highlight` if the code block has any highlighted lines. This is necessary for styling like this:

image

With the given code snippet:

    ```docker{3}
    # Every Action needs a Dockerfile
    FROM alpine
    RUN apk add --no-cache bash curl jq
    COPY comment /usr/bin/comment
    ENTRYPOINT ["comment"]
    ```

The key point here is that I'm able to style the block but still target the highlighted lines - here's the CSS, but my actual design choice isn't relevant:

```css
.gatsby-highlight.gatsby-highlight-with-lines > pre > code,
.gatsby-highlight.gatsby-highlight-with-lines
  > pre
  > code
  > *:not(.gatsby-highlight-code-line) {
  color: #959da5 !important;
}

.gatsby-highlight-code-line {
  color: #393a34;
  display: block;
}
```

This will hopefully allow users to be a lot less restricted in terms of styling; currently only the highlighted lines can be styled specially, but now the whole block 😊

Let me know if I can clarify anything, or if this isn't an idea y'all would like to act on. Happy to chat through it 🙌

### DIFF
--- a/packages/gatsby-remark-prismjs/src/__tests__/__snapshots__/index.js.snap
+++ b/packages/gatsby-remark-prismjs/src/__tests__/__snapshots__/index.js.snap
@@ -195,6 +195,50 @@ Object {
 }
 `;
 
+exports[`remark prism plugin generates a <pre> tag with highlighted lines 1`] = `
+Object {
+  "children": Array [
+    Object {
+      "lang": "js{2}",
+      "position": Position {
+        "end": Object {
+          "column": 4,
+          "line": 5,
+          "offset": 27,
+        },
+        "indent": Array [
+          1,
+          1,
+          1,
+          1,
+        ],
+        "start": Object {
+          "column": 1,
+          "line": 1,
+          "offset": 0,
+        },
+      },
+      "type": "html",
+      "value": "<div class=\\"gatsby-highlight has-highlighted-lines\\" data-language=\\"js\\"><pre class=\\"language-js\\"><code class=\\"language-js\\"><span class=\\"token comment\\">// 1</span>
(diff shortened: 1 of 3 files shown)


JasonEtco:
It looks like the linter is failing - but I'm not really sure why. Any chance someone can dig in?

KyleAMathews:
Kinda curious what your use case for this is?

KyleAMathews:
Perhaps make the class name "has-highlighted-lines"?

JasonEtco:
> Could you also please add tests to validate this behavior?

Ah yes will do, thanks for the link @LekoArts!

@KyleAMathews - my use-case is purely cosmetic. In using the highlighted lines, I found that styling the lines themselves was too restricted, because the colorful lines took visual focus, regardless of the addition of highlight-specific styles. With this class, I'm able to say "everything in here should be grayed out _except_ highlighted lines" - this lets focus be drawn entirely to those lines, while still showing the context code. Hope that makes sense!

KyleAMathews:
Cool idea!

m-allanson:
Thanks @JasonEtco 👍

JasonEtco:
Rad, thanks folks ✨ What does the release schedule look like? Should I be keeping my eye out for this in a next release, or will it be soon? Don't mean to rush you, just trying to decide if I should publish my own namespaced version for now.

m-allanson:
It should be available now in `gatsby-remark-prismjs@3.2.2` 🎉

JasonEtco:
Y'all are the best ❤️

Issue #10884, opened 2019-01-07
Scroll position flashes when page is reloaded


d3lm:
## Description

I have built a [small site](https://dazzling-bose-d84fff.netlify.com/) with Gatsby and recently upgraded to v2. Now I am facing two issues that I cannot figure out and I have been reading plenty of Github issues already. For your information the site is deployed on Netlify and the **code** can be found [here](https://github.com/yoga-gabi-goeller/website).

**Problem 1: Scroll position flashes when page is reloaded**

This is a very obvious issue and when ever the page is reloaded, it scrolls to the top for a short period of time and then jumps back to its original scoll position.

**Problem 2: Font and content flashes**

I am using the `gatsby-plugin-google-font` to load a font. Even though this adds a link tag to the head, which means the font should be loaded before the content is being rendered, I can see my text flashing. For a short period of time I can see the font changing.

The same goes for the content sometimes. For a short period of time I can see the styles not being loaded but the content is already rendered. I am using bootstrap grid for example. I found a very similar [issue](https://github.com/gatsbyjs/gatsby/issues/5667) but it's still open and no solution was successful for me.

### Steps to reproduce

**Problem 1:** Open the [page])(https://dazzling-bose-d84fff.netlify.com/) scroll down a little bit (or a few sections, doesn't matter) and reload the page. You'll notice that the scroll jumps. I have the feeling that the scroll position is restored once the JS kicks in, but this is a vague guess.

**Problem 2:** Open the [page])(https://dazzling-bose-d84fff.netlify.com/) and hard reload the page a couple of times. You'll notice that the font is flashing and you can also sometimes see unstyled content.

### Expected result

**Problem 1:** I expect the page scroll not to jump, at least this was the case with Gatsby v1. It worked properly and there was no flashes at all. It shouldn't be the solution to stay on v1.

**Problem2:** I expect the fonts to be loaded before the content is rendered. Is the link tag maybe dynamically (with JS) added to the head? The same goes for the styles. I'd love to load the styles before the content is rendered to avoid seeing unstyled content.

### Actual result

See problem description.

### Environment

My local setup is the following:

```
System:
  OS: macOS 10.14
  CPU: (4) x64 Intel(R) Core(TM) i5-4258U CPU @ 2.40GHz
  Shell: 3.2.57 - /bin/bash
Binaries:
  Node: 11.4.0 - /usr/local/bin/node
  Yarn: 1.12.3 - /usr/local/bin/yarn
[...]

PR #11224, opened 2019-01-23
fix(gatsby-react-router-scroll): keep scroll positions in tact on reload

wardpeet:
## Description
This should fix scroll restoration on refresh. The current solution relies on javascript to do the restoration of scroll which is great for Single Page Apps because no SSR is done. The node package scroll-behaviour sets the scrollRestoration to "manual" so we force it to be "auto" so the browser can do it's thing.

https://react-gatsby-scroll.surge.sh is a demo page of the react gatsby site. When running it on a good laptop, you won't see much difference but when you throttle your cpu you'll see the difference.

image

## Related Issues
https://github.com/gatsbyjs/gatsby/issues/10884

### DIFF
--- a/packages/gatsby-react-router-scroll/src/ScrollBehaviorContext.js
+++ b/packages/gatsby-react-router-scroll/src/ScrollBehaviorContext.js
@@ -33,18 +33,6 @@ class ScrollContext extends React.Component {
     }
   }
 
-  componentDidMount() {
-    /**
-     * scrollBehavior sets history.scrollRestoration to "manual" which means we have to take care of scrolling ourselves for pages.
-     * It works great in a client side SPA as the browser can't jump to the correct content because it's not there yet.
-     * Gatsby can because it's awesome (ships SSR html) so we don't want to disable scrollRestoration
-     * see #7454
-     */
-    if (`scrollRestoration` in window.history) {
-      window.history.scrollRestoration = `auto`
-    }
-  }
-
   componentDidUpdate(prevProps) {
     const { location } = this.props
     const prevLocation = prevProps.location
--- a/packages/gatsby-react-router-scroll/src/ScrollBehaviorContext.js
+++ b/packages/gatsby-react-router-scroll/src/ScrollBehaviorContext.js
@@ -25,14 +25,22 @@ class ScrollContext extends React.Component {
       getCurrentLocation: () => this.props.location,
       shouldUpdateScroll: this.shouldUpdateScroll,
     })
-
-    this.scrollBehavior.updateScroll(null, this.getRouterProps())
   }
 
   getChildContext() {
     return {
       scrollBehavior: this,
     }
+  }
+
+  componentDidMount() {
+    /**
+     * scrollBehavior sets history.scrollRestoration to "manual" which means we have to take care of scrolling ourselves for pages.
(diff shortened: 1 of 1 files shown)


KyleAMathews:
`scrollRestoration: auto` isn't supporter on edge.

Also, can't we just _not_ change the scroll position on the initial render?

wardpeet:
By default scrollRestoration is enabled on edge (default behaviour even without the supported property) but it boils down to being consistent if a browser does not support it we don't jump on reload. As long as we are consistent on the platform we are good to go.

We could disable scroll on initial render when the user has scrolled. If `pageY > 0` we know they scrolled so it's probably a good idea not to jump. If we move to native restoration, it happens much faster than js which means users have less time to scroll or get frustrated so I actually like this approach better than checking if we scroll.

KyleAMathews:
Ooops, mis-spoke earlier — `manual` isn't what's supported in edge.

>We could disable scroll on initial render when the user has scrolled.

That's what your removing the line in the constructor did `this.scrollBehavior.updateScroll(null, this.getRouterProps())` — we didn't need to try restoring the scroll on the initial mount.

We don't need to set scrollBehavior to auto to restore the browsers default behavior on load because until Gatsby's runtime is loaded, that _is_ the behavior :-D

KyleAMathews:
So if you remove the `componentDidMount` additions, this PR should be ready to go!

wardpeet:
I see, this means we will never jump on initial load. That's the behaviour we want? If so, that's so much easier 😄 I thought we wanted jumps

KyleAMathews:
Haha yeah, this is a pretty simple bug. We never should have had that line there restoring scroll on startup but since most of the time it doesn't cause trouble, it's stayed around until now.

duongthienlee:
It looks like this fix doesn't complete work with react-transition-group. I am following the code from https://github.com/gatsbyjs/gatsby/tree/master/examples/using-page-transitions/ to archieve page transition.
If i remove react-transition-group code, it works, otherwise, the bug still appear. Here is my code
            
       import React from 'react'
       import {TransitionGroup, Transition as ReactTransition}  from 'react-transition-group'
       const timeout = 300
       const getTransitionStyles = {
        entering: {
            position: `absolute`,
            opacity: 0
        },
        entered: {
            position: `relative`,
            transition: `all ease-in ${timeout}ms`,
            opacity: 1,
            transform: 'translateY(0px)'
        },
        exiting: {
            transition: `all  ease-in ${timeout}ms`,
            opacity: 0,
            transform: 'translateY(100px)'
        },
    }
    
    const Transition = (props) => {
        const { children, location } = props
        const currentKey = location.pathname.split('/')[1] || '/'
        return (
            <TransitionGroup className="page-content">
                <ReactTransition
                    key={currentKey}
                    timeout={{
                        enter: timeout,
                        exit: timeout,
                    }}
    
                >
                    {status => (
                        <div
                            style={{
                                ...getTransitionStyles[status],
                            }}
                        >
                            {children}
                        </div>
                    )}
                </ReactTransition>
            </TransitionGroup>
        )
    }
    
    export default Transition 
You can check my website where the bug happens : https://thienwinchester.netlify.com.
Scroll down the page a little bit then navigate to other pages, the page seems scroll to top first then navigate to other pages.

Version : "react-transition-group": "^2.2.1",
 "gatsby": "^2.4.3",

Issue #11222, opened 2019-01-23
[gatsby-plugin-typography] omitGoogleFont not working on development server since v2.2.3


MitchellCash:
## Description

Since `v2.2.3` of `gatsby-plugin-typography` the `omitGoogleFont` option no longer works on the development server.

When running a build it does the right thing and the font is not fetched from Google Fonts, but on the development server fetching the font is still injected into the `<head>` even with `omitGoogleFont: true`.

There is only one change between `v2.2.2..v2.2.3` and it is #10545. This pull request also happens to relate to Google Font loading so presumably this is where the bug was introduced from.

### Steps to reproduce

1. Test on a Gatsby site with `gatsby-plugin-typography`, `react-typography` and `typography` installed
2. I will also use the `typography-theme-ocean-beach` theme for testing purposes as it relies on the Google Roboto font
3. Create a `src/utils/typography.js` file:

```js
import Typography from 'typography'
import oceanBeachTheme from 'typography-theme-ocean-beach'

const typography = new Typography(oceanBeachTheme)

export default typography
```

3. Create a `gatsby-config.js` file:

```js
module.exports = {
  plugins: [
    {
      resolve: "gatsby-plugin-typography",
      options: {
        pathToConfigModule: "src/utils/typography",
        omitGoogleFont: true
      }
    }
  ]
}
```

4. Run `gatsby develop`
5. Jump into the browser developer tools and take note that in the `<head>` it is still fetching the fonts from Google Font:
6. Now run `gatsby build && gatsby serve`
7. Note the correct result and that in the `<head>` it does not fetch anything form Google Fonts

### Expected result

When using the option `omitGoogleFont: true` and deploying the development site with `gatsby develop` that it should not fetch the fonts from Google Fonts.

### Actual result

Even when using`omitGoogleFont: true` the development site still fetches the font from Google Fonts.

### Environment

```bash
System:
  OS: macOS 10.14.2
  CPU: (4) x64 Intel(R) Core(TM) i5-4278U CPU @ 2.60GHz
  Shell: 5.0.0 - /usr/local/bin/bash
Binaries:
  Node: 10.15.0 - /usr/local/opt/node@10/bin/node
  Yarn: 1.13.0 - /usr/local/bin/yarn
  npm: 6.4.1 - /usr/local/opt/node@10/bin/npm
Languages:
  Python: 2.7.15 - /Users/mitchellcash/.pyenv/shims/python
Browsers:
[...]

PR #11226, opened 2019-01-23
fix(gatsby-plugin-typography): omitGoogleFont in "develop"

LekoArts:
Fixes https://github.com/gatsbyjs/gatsby/issues/11222

Regression from the earlier PR that added the `gatsby-browser` again. TIL you can pass `pluginOptions` there, too 😋


wardpeet:
could you add a unit test? ^^ Nothing else to add

LekoArts:
How would I do that?

wardpeet:
it was hard to actually add so I did it myself.

Issue #11219, opened 2019-01-22
SVGs not rendering on iOS with "gatsby develop"


lannonbr:
## Description

When running gatsby in development mode and exposing it to the local network with `-H 0.0.0.0` and opening a site that has SVGs, the image doesn't appear and if it has alt-text, the alt text appears.

That said, when built and served, the SVG does render on iOS.

### Steps to reproduce

Here's a reproducible build repo: https://github.com/lannonbr/gatsby-svg-bug

As said in the readme there, run `gatsby develop -H 0.0.0.0` and see the bug and then build the site and serve it and see that the image appears on iOS.

### Environment

```
  System:
    OS: macOS 10.14.2
    CPU: (4) x64 Intel(R) Core(TM) i7-5557U CPU @ 3.10GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 11.6.0 - /usr/local/bin/node
    Yarn: 1.12.3 - ~/.yarn/bin/yarn
    npm: 6.5.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 71.0.3578.98
    Firefox: 64.0.2
    Safari: 12.0.2
  npmPackages:
    gatsby: ^2.0.76 => 2.0.76 
    gatsby-image: ^2.0.20 => 2.0.25 
    gatsby-plugin-manifest: ^2.0.9 => 2.0.12 
    gatsby-plugin-offline: ^2.0.16 => 2.0.20 
    gatsby-plugin-react-helmet: ^3.0.2 => 3.0.5 
    gatsby-plugin-sharp: ^2.0.14 => 2.0.16 
    gatsby-source-filesystem: ^2.0.8 => 2.0.12 
    gatsby-transformer-sharp: ^2.1.8 => 2.1.9 
  npmGlobalPackages:
    gatsby-cli: 2.4.6
```

As well, I am testing this on an iPhone and iPad on iOS 12.1.2

sidharthachatterjee:
This is happening because the url for the SVG is set to `http://0.0.0.0/<path>` which won't resolve correctly on any other device on the network. This should happen on any device other than the machine serving `gatsby develop`

![screenshot 2019-01-23 19 04 21](https://user-images.githubusercontent.com/7701981/51610057-c3cf7880-1f41-11e9-8a68-ad26c526652a.png)

sidharthachatterjee:
On build, it seems to set a relative url for this

We should probably set relative urls for `develop` as well

[...]

Issue #5801, opened 2018-06-08
Testing Site with Gatsby Develop from Local LAN


kimfucious:
## Summary

I've searched as best I could, but I could not find a simple way to enable a "stock" Gatsby install to support testing a site, running on a local server from another device on the same local network.

BrowserSync does a great job with this.  With webpack dev server, I would normally, edit the webpack config, changing localhost to an IP address.

I've checked for existing plugins as suggested, before trying a custom webpack config, but to no avail.  I could open up the hood to do this, but before I do, I thought I'd ask:

**What's the best (i.e. generally agreed upon) way to enable a Gatsby site for local development testing from other devices (i.e. mobiles) on the same local network?**

> :point_up:  I got a clever suggestion to use `gatsby develop -H 192.168.1.100 -p 8000` from @ryanditjia on #5733, which does work; however, **HMR seems to stop auto refreshing changes** when using this.

Thanks in advance for your suggestions!

## Relevant information
Using gatsby-default-starter

### Environment (if relevant)
* Gatsby version (`npm list gatsby`): 1.9.269
* gatsby-cli version (`gatsby --version`): 1.1.57

mquandalle:
We had a discussion about this on the Discord chat yesterday, and apparently binding only the local IP is a design decision. My opinion is also that Gatsby should behave like other build tools (create-react-app or Meteor) and have a development server that is network accessible. 

<details>
  <summary>Chat details</summary>

_httpteapot_:
Does gatsby develop (v1 or v2) can bind a port so that its accessible from the local network using $MY_COMPUTER_IP:8000?
Like create-react-app for instance

_ghardin137_
only during development or with "gatsby serve"

_httpteapot_
I'm interested in that feature during development, but gatsby develop doesn't open a network accessible port
(unless the problem is something else)

_LekoArts_
get your local ip with e.g. devip (npm package) and then use -H $YOURIP$ -p $YOURPORT$

_ghardin137_
by default it listens on 0.0.0.0
which would be anything
so it should be accessible to the network

_httpteapot_
Hum so it's supposed to work, because I've tried on multiple networks on multiple occasions and I never was able to connect to the development server from a different device on my network

_ghardin137_
i can try it real fast but i'm pretty sure it works

_httpteapot_
and it was always working on create-react-app and others tools

_LekoArts_
It also never worked for me. That's why I use the -H flag

_httpteapot_
I'm on linux

_LekoArts_
windows 10 here

_ghardin137_
[...]

PR #11227, opened 2019-01-23
fix(gatsby): Ensure publicPath is always relative for gatsby develop

sidharthachatterjee:
In `gatsby develop` we were setting `publicPath` for `webpack-dev-server` to an absolute url  _constructed_ from the program host, port and protocol

This is unnecessary and causes HMR and static links to break when viewed on a separate device over a network like in this screenshot below
screenshot 2019-01-23 19 04 21
This pull request fixes this by making the publicPath _always_ relative 

Fixes https://github.com/gatsbyjs/gatsby/issues/11219 and https://github.com/gatsbyjs/gatsby/issues/5801

### DIFF
--- a/packages/gatsby/src/utils/webpack.config.js
+++ b/packages/gatsby/src/utils/webpack.config.js
@@ -107,11 +107,7 @@ module.exports = async (
           // Add /* filename */ comments to generated require()s in the output.
           pathinfo: true,
           // Point sourcemap entries to original disk location (format as URL on Windows)
-          publicPath:
-            process.env.GATSBY_WEBPACK_PUBLICPATH ||
-            `${program.ssl ? `https` : `http`}://${
-              program.host
-            }:${webpackPort}/`,
+          publicPath: process.env.GATSBY_WEBPACK_PUBLICPATH || `/`,
           devtoolModuleFilenameTemplate: info =>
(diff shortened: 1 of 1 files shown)


packages/gatsby/src/utils/webpack.config.js
```diff
@@ -107,11 +107,7 @@ module.exports = async (
           // Add /* filename */ comments to generated require()s in the output.
           pathinfo: true,
           // Point sourcemap entries to original disk location (format as URL on Windows)
-          publicPath:
-            process.env.GATSBY_WEBPACK_PUBLICPATH ||
-            `${program.ssl ? `https` : `http`}://${
-              program.host
-            }:${webpackPort}/`,
+          publicPath: process.env.GATSBY_WEBPACK_PUBLICPATH || `/`,
```
DSchau: Wouldn't we still want to handle the ssl stuff? Seems like this would be a breaking change for that behavior.

If nothing else--would you be able to validate this hasn't broken that behavior?

sidharthachatterjee: Just tested, works fine!

> Keep in mind that the automatic certificates issued with the --https flag are explicitly issued to localhost and will only be accepted there. Using it together with the --host option will likely result in browser warnings.

This behaviour documented in https://www.gatsbyjs.org/docs/local-https/ remains the same

PR #11228, opened 2019-01-23
fix: env var name change for build

### LINKED ISSUE #11143: [www] Plugins feedback button is not working
## Description

When i click on the mail icon in the plugins page i get

screenshot_2019-01-18 gatsby-source-instagram

### Expected result
Not sure :D

amberleyromo:
@DSchau ready for when you make that change.

Fixes #11143

### DIFF
--- a/www/README.md
+++ b/www/README.md
@@ -34,8 +34,8 @@ Several areas of the site include feedback widgets (currently starter library an
 If you have access to the keys, add them like so:
 
 ```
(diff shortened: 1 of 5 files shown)


sidharthachatterjee:
Fixes https://github.com/gatsbyjs/gatsby/issues/11143

DSchau:
@sidharthachatterjee thanks for tagging that in! I don't _know_ if commenting like that will also close, so I went ahead and added that to the description as well--which I know does close the issue!

PR #11230, opened 2019-01-23
fix(www): Prefix unbird env keys with GATSBY_

sidharthachatterjee:
Potentially fixes UNBIRD keys which are not in an .env file and therefore need to be prefixed

- [ ] Needs to be tested with Preview

Fixes https://github.com/gatsbyjs/gatsby/issues/11143


DSchau:
Closing as duplicate of #11228. Thanks though!

PR #11234, opened 2019-01-23
chore(examples): update deps and tweak style for styled-components example

xxczaki:
## Description

- Updated all dependencies
- Beautified example page, by styling the `<a>` tag and aligning text to the center :smile:


examples/using-styled-components/package.json
```diff
@@ -5,22 +5,22 @@
...
-    "gatsby-plugin-google-analytics": "^2.0.5",
-    "gatsby-plugin-offline": "^2.0.5",
-    "gatsby-plugin-react-helmet": "^3.0.0",
-    "gatsby-plugin-styled-components": "^3.0.0",
-    "gatsby-plugin-typography": "^2.2.0",
-    "lodash": "^4.17.10",
-    "react": "^16.4.0",
-    "react-dom": "^16.4.0",
+    "babel-plugin-styled-components": "^1.10.0",
+    "gatsby": "^2.0.91",
```
DSchau: Because we don't use lock files for these examples--these are synonymous. e.g. `^2.0.0` will resolve to the latest version of gatsby, which in this case is now `2.0.93`.

So while this change is appreciated--it's not necessary! This same comment applies to most of these changes!

xxczaki: Oh, ok. Didn't know about that actually :smile:

examples/using-styled-components/src/pages/index.js
```diff
@@ -16,8 +16,16 @@ const Title = styled.h1`
   color: palevioletred;
 `
 
+// Create an URL component that'll render an <a> tag with some styles
+const Url = styled.a`
```
DSchau: Perhaps we just call this a `Link`? Or `ExternalLink` for clarity?

(the [URL global](https://developer.mozilla.org/en-US/docs/Web/API/URL)--while capitalized differently--can also be slightly confusing here!)

xxczaki: I think I will call it `ExternalLink` so that no one will confuse it with Gatsby's [Link](https://www.gatsbyjs.org/docs/gatsby-link/)

PR #11253, opened 2019-01-24
Add Packrs official website to gatsby showcase

aesthytik:

### DIFF
--- a/docs/sites.yml
+++ b/docs/sites.yml
@@ -4062,3 +4062,15 @@
   built_by: Gaël Billon
   built_by_url: "https://gaelbillon.com"
   featured: false
+- title: Packrs
+  main_url: "https://www.packrs.co/"
+  url: "https://www.packrs.co/"
+  description: Packrs is a local parcel delivery platform, A one spot destination for all your daily needs on a single click. Just pick the locations and get your deliveries packred!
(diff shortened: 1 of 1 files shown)


docs/sites.yml
```diff
@@ -4062,3 +4062,15 @@
   built_by: Gaël Billon
   built_by_url: "https://gaelbillon.com"
   featured: false
+- title: Packrs
+  url: "https://www.packrs.co/"
+  main_url: "https://www.packrs.co/"
+  description: >
+    Packrs is a local delivery platform, one spot for all your daily requirements. On a single tap get everything you need at your doorstep.
+  categories:
+    - React Native
```
DSchau: Let's keep these as categories of the _site_ not the app! Otherwise this is fairly confusing :)

DSchau:
```suggestion
```

aesthytik: @DSchau the website is a landing page for an app build with react native. So, category is suited right?

DSchau: The categories should be representative of the app built with Gatsby, not the underlying product. Does that make sense?

aesthytik: understood @DSchau thanks!

PR #11254, opened 2019-01-24
chore(docs): update headless CMS ordering

calcsam:
* Added note about ordering
* Reordered links to reflect usage
* Used gatsby-link
* Added links to guides and starters


calcsam:
I think ultimately we'll want a custom table component here, but this is a start.

PR #11257, opened 2019-01-24
Update lerna

m-allanson:
## Description

Update Lerna to the latest version

It also seems we had an unused devdep for an older version of Lerna. I've removed it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "jest": "^23.5.0",
     "jest-cli": "^23.5.0",
     "jest-junit": "^5.0.0",
-    "lerna": "^3.3.0",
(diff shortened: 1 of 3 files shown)


m-allanson:
> My only comment is that I vaguely recall @pieh noting he ran into an issue when we've updated this in the past.

Hmm I vaguely remember that too 🤔 @pieh do you remember what issues you ran into?

(Thanks for fixing the title!)

pieh:
Gotta be honest here and say that I don't remember any specifics. I do think it's worth a try - if stuff break, we can either try to fix it if it's quick-fix or just revert this upgrade?

PR #11265, opened 2019-01-24
feat(gatsby-transformer-react-docgen): allow parsing typescript files

jquense:
Typescript has no mime type :/


packages/gatsby-transformer-react-docgen/src/on-node-create.js
```diff
@@ -10,6 +10,19 @@ const digest = str =>
 const propsId = (parentId, name) => `${parentId}--ComponentProp-${name}`
 const descId = parentId => `${parentId}--ComponentDescription`
 
+function canParse(node) {
+  return (
+    node &&
```
DSchau: I don't think this check is really needed!

```suggestion
```

packages/gatsby-transformer-react-docgen/src/on-node-create.js
```diff
@@ -10,6 +10,19 @@ const digest = str =>
...
+    (node.internal.mediaType === `application/javascript` ||
+      // Typescript doesn't really have a mime type and .ts files are a media file :/
+      node.internal.mediaType === `application/typescript` ||
```
DSchau: Is there a reason we need _all_ of these checks? I would've thought there's one condition we could reliably use, i.e.

```js
return (
  node.internal.mediaType === `application/javascript` ||
  node.internal.mediaType === `application/typescript`
)
```

packages/gatsby-transformer-react-docgen/src/on-node-create.js
jquense: ya it more out of an abundance of caution i can remove if you like?

DSchau: It's probably OK. Let's just merge this, thank you!

DSchau:
Published as [`gatsby-transformer-react-docgen@3.0.3`](https://github.com/gatsbyjs/gatsby/blob/8830d47dd72e049e007ffeaeb966ec1632afd857/packages/gatsby-transformer-react-docgen/CHANGELOG.md#303-2019-01-29)

jquense:
🎉

thchia:
@jquense I'm getting the following error when attempting to parse a TS file:

```bash
error There was a problem parsing component metadata for file: "layout.tsx"

   8 | interface LayoutProps {
   9 |   children?: React.ReactNode
> 10 |   foo: string
     |  ^
  11 | }
  12 |
  13 | /**


  SyntaxError: Unexpected token (10:2)

- index.js:3834 Object.raise
    [gatsby-ts-docgen]/[@babel]/parser/lib/index.js:3834:17

  - index.js:5142 Object.unexpected
    [gatsby-ts-docgen]/[@babel]/parser/lib/index.js:5142:16

  - index.js:1542 Object.flowObjectTypeSemicolon
    [gatsby-ts-docgen]/[@babel]/parser/lib/index.js:1542:12

...
```

Seems to be related to the `flowObjectTypeSemicolon` check during the parse operation. AFAIK, omitting the semicolon is valid TS.

jquense:
@thchia yeah at the moment full ts support isn't there, this adds the ability to even see .ts/.tsx files. We still need better upstream react-docgen support, e.g. https://github.com/reactjs/react-docgen/pull/320

thchia:
@jquense would adding https://github.com/styleguidist/react-docgen-typescript be of any help? I'm playing around now trying to switch parsers based on the file type.

jquense:
I'd really like to _not_ deal with that, and instead just snag the proptypes handler from there and use it with the original react-docgen.

thchia:
Do you mean not deal with adding an extra dependency? Because it looks like the snaggable bits form the bulk of the code anyway. Either way I'm with you on not having multiple parser instances.

jquense:
it's a few things, not having the extra dep but also not having to deal with inconsistencies and differences between the two tools.

Issue #11268, opened 2019-01-24
Enable support for Webp in gatsby-remark-images-contentful


AlbericTrancart:
## Summary

As of now, `gatsby-remark-contentful-images` doesn't have a `withWebp` option like `gatsby-remark-images`.

### Basic example

This config:

```
{
  resolve: `gatsby-transformer-remark`,
  options: {
    plugins: [
      {
        resolve: `gatsby-remark-images-contentful`,
        options: {
          maxWidth: 1000,
          withWebp: true
        }
      }
    ]
  }
},
```

Would add srcsets to webp images with a fallback on the classic srcset for unsupported browsers.

### Motivation

This could provide a substantial speed improvement on supported browsers.

I volunteer for coding this feature :)

AlbericTrancart:
Hey! [This issue has a PR ready and waiting for approval.](https://github.com/gatsbyjs/gatsby/pull/11273)

PR #11273, opened 2019-01-25
feat(gatsby-remark-images-contentful): enable webp sources

AlbericTrancart:
## Description

Add a `withWebp` option to include a webp source in parsed images from contentful.

## Related Issues

Addresses #11268.


KyleAMathews:
Is there any reason to not just make this the default? Seems like everyone should have webp enabled :-)

AlbericTrancart:
@KyleAMathews I didn't enabled it by default because it isn't in `gatsby-remark-images` so I thought there should be a good reason, but I will make it the default then :+1: 

@wardpeet what kind of structure are you thinking of? Something like this? (I will implement this TODO rightaway if possible :wink: )

```
srcSet: {
  original: ...,
  webp: ...
}
```

wardpeet:
hey @AlbericTrancart it's super cool that you want to take this on right away but it would be a breaking change so we have to wait for gatsby v3. Also I haven't thought about it yet 😄.

AlbericTrancart:
Updated!

KyleAMathews:
@wardpeet this wouldn't be a breaking change actually because it's hidden from the user — nobody's code would break — their sites would just get faster :-)

KyleAMathews:
@AlbericTrancart the only reason we don't include it by default in gatsby-remark-images is that generating webp thumbnails is a _lot_ more expensive than jpg. So we want people to opt into webp so they realize the cost they're taking on (as not everyone wants to tradeoff longer build times for smaller images).

But with a hosted service like Contentful where webp images are generated on demand, it makes a lot more sense to default to webp.

AlbericTrancart:
@KyleAMathews @wardpeet any news on merging this PR? :)

AlbericTrancart:
@wardpeet PR rebased to fix a merge conflict on the README.

ryanwiemer:
Any update on this? I would love to see the gatsby-remark-images-contentful plugin get some improvements.

AlbericTrancart:
@wardpeet @KyleAMathews ping to merge this PR :)

sidharthachatterjee:
Published in gatsby-remark-images-contentful@2.0.9

Issue #11281, opened 2019-01-25
[gatsby-plugin-netlify-cms] Plugin's webpack build process overrides Gatsby's build process


asistapl:
## Description

After adding `gatsby-plugin-netlify-cms` to gatsby project generated html files doesn't include scripts and are running without javascript. `chunk-map.json` file is overridden with chunks generated by plugin's webpack build and looks like this: `{"cms":["/cms.js"]}`.

### Steps to reproduce

https://github.com/asistapl/gatsby-plugin-netlify-cms-bug

I created that repository with these steps:

1. `gatsby new gatsby-plugin-netlify-cms-bug`
2. `cd gatsby-plugin-netlify-cms-bug`
3. `yarn add netlify-cms gatsby-plugin-netlify-cms`
4. Added plugin to `gatsby-config.js`
5. Updated all dependencies.

To see a bug follow these steps:
1. `git clone https://github.com/asistapl/gatsby-plugin-netlify-cms-bug`
2. `cd gatsby-plugin-netlify-cms-bug`
3. `yarn install`
4. `gatsby build` - It's a first build and it sometimes work at that moment
5. Run `gatsby build` again - now bug should be visible, because gatsby is winning a race thanks to cache.
6. To see a bug run `gatsby serve` and check if scripts are running
    or open `chunk-map.json` file and see that it includes only chunks generated by the plugin.

I tried to:
1. Update all dependencies in `package.json` and even gatsby-cli.
2. Move plugin at the top of array in `gatsby-config.js`.

It didn't fixed the problem.

### Expected result

Plugin shouldn't change Gatsby's build output.

### Actual result

Plugin overrode Gatsby's build output.

### Environment

```
  System:
    OS: macOS 10.14.2
    CPU: (8) x64 Intel(R) Core(TM) i7-7920HQ CPU @ 3.10GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.9.0 - /usr/local/bin/node
    Yarn: 1.12.3 - /usr/local/bin/yarn
    npm: 6.2.0 - /usr/local/bin/npm
  Languages:
    Python: 2.7.10 - /usr/bin/python
  Browsers:
    Chrome: 71.0.3578.98
    Firefox: 64.0.2
    Safari: 12.0.2
  npmPackages:
    gatsby: ^2.0.98 => 2.0.98 
    gatsby-image: ^2.0.28 => 2.0.28 
[...]

PR #11288, opened 2019-01-25
fix(gatsby-plugin-netlify-cms): exclude gatsby-webpack-stats-extractor from being called

sidharthachatterjee:
The Netlify CMS plugin runs its own instance of webpack and uses Gatsby's plugins as a base webpack config

Unfortunately our previously inlined (fixed in https://github.com/gatsbyjs/gatsby/pull/11287) gatsby-webpack-stats-extractor gets called and overwrites `chunk-map.json` as seen in https://github.com/gatsbyjs/gatsby/issues/11281 and reproducible in https://github.com/asistapl/gatsby-plugin-netlify-cms-bug

As a result of chunk-map.json being overwritten by the plugin's version (which only contains cms.js) the entire app breaks because static-entry.js can't find the right chunks

This PR adds `GatsbyWebpackStatsExtractor` to the list of ignored plugins in this plugin's `onCreateWebpackConfig`

Fixes https://github.com/gatsbyjs/gatsby/issues/11281

This PR depends on https://github.com/gatsbyjs/gatsby/pull/11287 and needs that to be merged in first

### DIFF
--- a/packages/gatsby-plugin-netlify-cms/package.json
+++ b/packages/gatsby-plugin-netlify-cms/package.json
@@ -34,7 +34,7 @@
   "license": "MIT",
   "main": "index.js",
   "peerDependencies": {
-    "gatsby": "^2.0.0-beta.29",
+    "gatsby": "^2.0.101",
     "netlify-cms": "^2.0.6"
   },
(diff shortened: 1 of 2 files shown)


pieh:
I've just merged and published  #11287. Will update peerDep

PR #11293, opened 2019-01-25
chore: upgrade to jest 24, update docs to reflect changes

DSchau:
## Description

This PR upgrades our root Jest dependency to Jest 24 (released today!), and also does the following:

- Tweaks the docs/guides for unit testing for Jest 24
- Uses the default starter instead of blog starter for testing examples (we should make this a convention going forward, it just makes things easier)
- Tweaks the using-jest example for Jest 24 support

## Related Issues

Fixes #11267

PR #11302, opened 2019-01-25
fix(docs): update key name for jest@24 support

jsjoeio:
## Description

Jest recently added a "Deprecation Warning" saying 
> Option "setupTestFrameworkScriptFile" was replaced by configuration "setupFilesAfterEnv", which supports multiple paths.

This PR fixes the docs to match the newly updated configuration.

screen shot 2019-01-25 at 3 34 06 pm

### DIFF
--- a/docs/docs/testing-react-components.md
+++ b/docs/docs/testing-react-components.md
@@ -33,7 +33,7 @@ Lastly you need to tell Jest where to find this file. Open your `jest.config.js`
 
 ```js:title=jest.config.js
 module.exports = {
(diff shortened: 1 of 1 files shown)


LekoArts:
Hi, thanks for the PR. However, as I understand it it's still in beta release? https://github.com/facebook/jest/pull/7119#issuecomment-456205302
So reading that I don't think we should merge that just yet.

DSchau:
@LekoArts it's available in Jest@24--which released last week!

I've updated the documentation and examples, and the PR was just merged in #11293!

This is actually in a separate file that I missed--so let's get this merged.

Thanks Joe!

Issue #11298, opened 2019-01-25
Swap image source with gatsby-image


wmcheung:
A quick question, is it possible to swap images using gatsby-image?

The query I'm using in my Image component is:

```
query {
              allImageSharp {
                edges {
                  node {
                    ... on ImageSharp {
                      fluid(maxWidth: 1250) {
                        ...GatsbyImageSharpFluid,
                        originalName
                      }
                    }
                  }
                }
              }
            }
```

This looks for the file name in my Image component which is sent to the gatbsy-image.
```
const image = data.allImageSharp.edges.find( (e) => {
              if (e.node.fluid.originalName === this.props.src) {
                return e.node;
              }
            });

<Img fluid={image.node.fluid} />
```

What should happen is if I create an onClick and I choose to change it into `image-2.png`, it should show the new image. The image gets replaced but it remains to show the blurred version.

Here is a GIF of what is happening https://cl.ly/c2729110e408

Much appreciated for your help!

jonniebigodes:
@wmcheung this is related to #8707 . I've been testing out and you can get a "false positive" running on development mode. You can change it at will. The real issue is when you start a production build. I'm testing to see if i can solve that issue if yo ucan, be on the lookout for that one and see if it helps

wmcheung:
@jonniebigodes issue #11303 solved my problem.

CarlRosell:
I will reopen this one as the PR is not merged yet and this is still a problem

PR #11303, opened 2019-01-25
fix(gatsby-image): use the cache to tell if the image was already loaded

CarlRosell:
## Description

I found a weird issue while browsing the gatsby store earlier today. When I reloaded the product detail page the images were not loaded correctly.
It turns out that the order of the IntersectionObserver and the image load event is not deterministic. By using the image cache to tell if the image was loaded or not, the call order will no longer be an issue.

Before:

kapture 2019-01-25 at 23 31 32

After:

kapture 2019-01-25 at 23 36 00

## Related Issues
Fixes: #11298

### DIFF
--- a/packages/gatsby-image/src/index.js
+++ b/packages/gatsby-image/src/index.js
@@ -189,14 +189,15 @@ class Image extends React.Component {
   handleRef(ref) {
     if (this.state.IOSupported && ref) {
       listenToIntersections(ref, () => {
+        const imageInCache = inImageCache(this.props)
         if (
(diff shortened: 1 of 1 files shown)


CarlRosell:
This issue might be related #11298

wmcheung:
@CarlRosell thanks for referencing to my issue. I cloned your project and tested it by replacing the gatsby-image index file and it fixed my problem.

Before:
https://cl.ly/c2729110e408

After
https://cl.ly/8cfc535ec836

PR #11304, opened 2019-01-26
feat(gatsby-plugin-page-creator): add ignore support via plugin options, fix test file ignores on Windows

### LINKED ISSUE #11168: Add ignore option to gatsby-plugin-page-creator
## Summary

I would like to give my client the option of choosing, via a config file, whether or not to include blog pages on their generated site.

Example project:
```
src/
    content/
        blog/
            post.md              <--- this becomes http://mysite.com/blog/post
        other/
            other.md
    pages/
        blog.js                    <--- this is http://mysite.com/blog
        other.js
```

The first part, dynamically selecting which content files to turn into pages, was easy.

The second part, dynamically selecting which `pages/*.js` files to process, eludes me, and is especially annoying when building results in a GraphQL error due to the blog content not existing for the blog index page.

## Relevant information

I looked into [`gatsby-plugin-page-creator`](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-page-creator) since it is responsible for creating `/pages/*.js` pages, but unfortunately it doesn't seem to support the `ignore` option that `gatsby-source-filesystem` uses.

## Request

Could an ignore option be added to [`gatsby-plugin-page-creator`](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-page-creator)? Or is there another way to programmatically decide which `/pages` files to create?

Warped2713:
## Description

This PR modifies the [gatsby-plugin-page-creator](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-page-creator) plugin to allow for custom exclusion of page files. It uses the patterns and options of [micromatch](https://www.npmjs.com/package/micromatch) which is what the plugin already uses for excluding various test files.

There were also two adjustments to the plugin, detailed below.

### Changes specific to this feature

- Adds a new option to [gatsby-plugin-page-creator](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-page-creator), `ignore: { patterns: [], options: {} }` which gets passed to [micromatch.any](https://www.npmjs.com/package/micromatch#any) to filter the files in the given pages directory path
- Updates the README with example usage for the ignore option
- Adds 4 unit tests
  - Does not filter without an ignore object
  - Does not filter without ignore.patterns
  - Filters with ignore.patterns
  - Filters with ignore.patterns AND ignore.options (nocase for case insensitive)
  - `yarn jest "packages/gatsby-plugin-page-creator"` results in 13/13 passed tests, 2/2 passed snapshots

### Additional changes

1. Apply `slash()` to `pagesDirectory` (line 51) and to files added by `chokidar` (line 63) to avoid weirdness in `systemPath.posix.relative(pagesDirectory, filePath)` when pageDirectory and filePath use different slash conventions
    1. Without the change to line 51, the resulting path (on Windows) looked like `../../../../C:/user/path/repo/src/pages/file.js` which fails the micromatch pattern of `**/`
    2. Without adding `path = slash(path)` in between line 61-62, files were added twice (on Windows): once with forward slashes by glob (line 55) and once with back slashes by chokidar (line 62)

https://github.com/gatsbyjs/gatsby/blob/026d2a956296cb01936bcf45d2be1066dd844d00/packages/gatsby-plugin-page-creator/src/gatsby-node.js#L51-L66

2. Update the README to note that the default file exclusions also exclude `test.js` and `spec.js` (and .ts, jsx, etc) meaning that if you make a `pages/test.js`, it will be ignored by the page creator due to line 11 (unchanged) in validate-path.js

https://github.com/gatsbyjs/gatsby/blob/026d2a956296cb01936bcf45d2be1066dd844d00/packages/gatsby-plugin-page-creator/src/validate-path.js#L11

## Related Issues

Fixes #11168


Warped2713:
Okay, sure, I'll go ahead and adjust for that.

Warped2713:
This PR is ready for review

packages/gatsby-plugin-page-creator/README.md
```diff
@@ -15,6 +15,10 @@ With this plugin, _any_ file that lives in the `src/pages` folder (or subfolders
 - `_*`
 - `.*`
 
+NOTE: also excludes `test.js` and `spec.js` (or .ts, .jsx, .tsx)
```
pieh: This should be fixed in https://github.com/gatsbyjs/gatsby/pull/11862 so I think it's safe to drop this line

packages/gatsby-plugin-page-creator/src/__tests__/gatsby-node.js
```diff
@@ -1,5 +1,6 @@
 const validatePath = require(`../validate-path`)
 const createPath = require(`../create-path`)
+const ignorePath = require(`../ignore-path`)
```
pieh: Can we move those tests to separate test file (`ignore-path.js`)? Added tests check `ignorePath` specifically - so moving them to separate file should make it cleaner?

packages/gatsby-plugin-page-creator/README.md
```diff
@@ -46,3 +50,57 @@ module.exports = {
...
+
+### Ignoring Specific Files
+
+#### Shorthand
+
+```javascript
+// The following example will disable the `/blog` index page
+// Note: This will only stop the creation of the `/blog` page!
+// To disable both the `/blog` and `/blog/post-slug` pages,
+// make a config.js which defines an `ignorePages` array
```
pieh: This is very valid note, but I think it comes too early (before showing basic example of `ignore` options in the plugin.

I think we need this note, but it should probably be placed after `gatsby-config` code snippet (?)

Warped2713: Sure. Looking at this again, I wonder if the entire note is necessary, given that it details a specific implementation. Maybe the following would be more appropriate?

```md
**NOTE**: The above code snippet will only stop the creation of the `/blog` page. 
This plugin is only responsible for creating pages from React components. 
It does not affect the generation of pages from content/data files 
(see [gatsby-source-filesystem](https://www.gatsbyjs.org/packages/gatsby-source-filesystem/)).
```

packages/gatsby-plugin-page-creator/src/__tests__/gatsby-node.js
Warped2713: Done!

packages/gatsby-plugin-page-creator/README.md
Warped2713: Done!

packages/gatsby-plugin-page-creator/README.md
Warped2713: @pieh do you prefer the original note, or the new one regarding the scope of `ignore` for `gatsby-plugin-page-creator`?

pieh: I think new one is better - we don't need to show opinionated way of doing global ignore that is used both by this plugin and in gatsby-node.

What about:
```
This plugin does not affect the programmatically generated pages with [`createPages` API](https://www.gatsbyjs.org/docs/node-apis/#createPages)
```

Warped2713: Yeah, that's much clearer.

slorber:
Hi,

Just want to point out that it may be worth to change default setup as well.

I'm using Apollo with dynamic data and using apollo codegen to generate page graphql TS types. I get the following output:

image



I suggest to auto exclude __generated__ folders + any .ts file. .ts files are unlikely to export React components because .tsx extension is required

Warped2713:
@slorber I am concerned that auto-excluding files according to a specific pipeline might cause trouble for users with a different pipeline (ex: someone whose `__generated__` folder contains files that do export React components). Might this be a discussion worthy of its own issue?

@pieh I just realized that I was supposed to 'accept' your requested changes and not implement them myself, and now your requests are `outdated`. What are the next steps for getting this PR merged?

slorber:
sure @Warped2713 you are probably right, just wanted to report my usecase somewhere, but I would be fine when this feature gets merged, in the meantime I'm using the flatOutput option of codegen which avoid colocating generated defs to their comp

paulmelnikow:
Thanks so much for this work @Warped2713!

paulmelnikow:
badges/shields#3208 is about a related bug that PR fixes 🙌, which prevents the ignore patterns from working correctly on Windows. That issue is fixed by this change:

> Apply `slash()` to `pagesDirectory` (line 51) and to files added by `chokidar` (line 63) to avoid weirdness in `systemPath.posix.relative(pagesDirectory, filePath)` when pageDirectory and filePath use different slash conventions
> 1. Without the change to line 51, the resulting path (on Windows) looked like `../../../../C:/user/path/repo/src/pages/file.js` which fails the micromatch pattern of `**/`
> 2. Without adding `path = slash(path)` in between line 61-62, files were added twice (on Windows): once with forward slashes by glob (line 55) and once with back slashes by chokidar (line 62)

@Warped2713 As I'm understanding this more deeply in the context of the discussion at badges/shields#3208, I'm wondering if it would be clearer to pass around paths relative to `pagesDirectory`, instead of complete paths. Both `chokidar` and `glob` accept `cwd` options which allow setting the base directory for searching. I'd assume the paths emitted are then also relative to that path (though I'm not totally sure).

It seems like working with `cwd`s would remove the need to call `systemPath.posix.relative`. This would make `validatePath` and `ignorePath` a bit more obvious, and it seems to me would also fix the Windows issue, as I suspect `systemPath.posix.relative` is what is introducing the leading `../../../`'s in Windows. Adding calls to `slash()` works to strip those off, though it seems cleaner to avoid putting them on in the first place 😁

Warped2713:
@paulmelnikow I agree that the current code feels rather messy. I'll take a closer look tonight. 

But, if I remember correctly, it is `const pagesDirectory = systemPath.posix.join(pagesPath)` which introduces the `../` and not `systemPath.posix.relative` (which gets `pagesDirectory` as input).

That same `pagesDirectory` is part of what gets passed to glob/chokidar to find the file paths, and this `pagesDirectory` is also passed to createPath, ignorePath, and validatePath.

paulmelnikow:
Ah, you're right, I mistyped that. Will fix.

The more I think about it, the clearer it seems that `systemPath.posix.join` is the problematic call. It wants to join multiple posix paths, which makes it totally wrong to hand a Windows path to it and expect the right result.

I think it's possible that `pagesPath` could be used directly as the `cwd` for both `glob` and `chokidar`, and that each of these would then return paths relative to that.

If the path needs to be processed prior to handing off to either of those, maybe `systemPath.posix.abspath` (which I suggested in the other thread), `systemPath.abspath`, or `systemPath.resolve` would do a better job.

If `chokidar` and `glob` return paths without the cwd prefix (as I hope they do by default or can be configured to do), then the rest of the validation functions can work in relative posix paths which seems more straightforward.

Warped2713:
@paulmelnikow Okay I looked into it, and actually your original comment was right. The `../` is added by `systemPath.posix.relative`. Interestingly, it adds a `../` for every forward slash (plus ending `/`) in the directory path, which means `slash(directoryPath)` would result in more `../` than a directory path with mixed `/` and `\` or only `\`

As for `systemPath.posix`, it would probably make sense to let Node choose between `path.posix` and `path.win32` based on `process.platform`. I didn't see any `abspath` in the path docs, but using `systemPath.resolve(process.cwd(), pagesPath)` seems to work nicely.

So with that decided, here's what I'm changing:
- set `pagesDirectory` to`systemPath.resolve(process.cwd(), pagesPath)` and change all `systemPath.posix` to simply `systemPath` except for in `create-path.js` where we want Unix-style output
- remove `${pagesDirectory}/` from `pagesGlob` so it is simply `**/*.{${exts}}`
- add `{cwd: pagesPath}` as options to `glob()` and `chokidar.watch()`
- pass the relative `filePath` to `validatePath()` and `ignorePath()`, instead of `systemPath.posix.relative(pagesDirectory, filePath)`
- set all component paths to `systemPath.posix.join(pagesDirectory, filePath)` since `filePath` is now relative to `pagesPath` and component paths must be absolute. The equality comparison for chokidar's unlink event should now compare `page.component` to this updated component path instead of `filePath` (which is now relative).
- `slash()` should still be applied to the relative `filePath` found by chokidar, during both the add and unlink events, since these paths are compared to those collected by glob (which uses `/` regardless of system). This also helps maintain the expectation that `create-path` outputs Unix paths
- simplify `create-path.js` to use the relative `filePath` instead of `path.posix.relative(slash(basePath), slash(filePath))`
- Update `create-path` unit test to reflect the fact that given `filePath` is always relative. The "It should deal with windows paths" test is no longer necessary as a result.

packages/gatsby-plugin-page-creator/src/gatsby-node.js
```diff
@@ -48,48 +49,56 @@ exports.createPagesStatefully = async (
...
 
-  const pagesDirectory = systemPath.posix.join(pagesPath)
-  const pagesGlob = `${pagesDirectory}/**/*.{${exts}}`
+  const pagesDirectory = systemPath.resolve(process.cwd(), pagesPath)
+  const pagesGlob = `**/*.{${exts}}`
 
   // Get initial list of files.
-  let files = await glob(pagesGlob)
-  files.forEach(file => _createPage(file, pagesDirectory, createPage))
+  let files = await glob(pagesGlob, { cwd: pagesPath })
```
wardpeet: could you move to `micromatch` here as well? or move to glob everwhere. It doesn't make much sense dependency wise to have 2 glob deps 😄

packages/gatsby-plugin-page-creator/src/gatsby-node.js
```diff
@@ -48,48 +49,56 @@ exports.createPagesStatefully = async (
...
+  files.forEach(file => _createPage(file, pagesDirectory, createPage, ignore))
 
   // Listen for new component pages to be added or removed.
   chokidar
-    .watch(pagesGlob)
+    .watch(pagesGlob, { cwd: pagesPath })
```
wardpeet: this is great 👍

packages/gatsby-plugin-page-creator/src/gatsby-node.js
```diff
@@ -48,48 +49,56 @@ exports.createPagesStatefully = async (
...
+        _createPage(path, pagesDirectory, createPage, ignore)
         files.push(path)
       }
     })
     .on(`unlink`, path => {
       path = slash(path)
       // Delete the page for the now deleted component.
       store.getState().pages.forEach(page => {
-        if (page.component === path) {
+        const componentPath = systemPath.join(pagesDirectory, path)
```
wardpeet: I'm not sure but should we go for posix.join? I'm unsure if it's needed here
```suggestion
        const componentPath = systemPath.posix.join(pagesDirectory, path)
```

packages/gatsby-plugin-page-creator/src/gatsby-node.js
```diff
@@ -48,48 +49,56 @@ exports.createPagesStatefully = async (
...
     return
   }
 
   // Create page object
+  const createdPath = createPath(filePath)
   const page = {
-    path: createPath(pagesDirectory, filePath),
-    component: filePath,
+    path: createdPath,
+    component: systemPath.join(pagesDirectory, filePath),
```
wardpeet: I'm not sure but should we go for posix.join? I'm unsure if it's needed here

```suggestion
    component: systemPath.posix.join(pagesDirectory, filePath),
```

packages/gatsby-plugin-page-creator/src/gatsby-node.js
```diff
@@ -48,48 +49,56 @@ exports.createPagesStatefully = async (
     )
   }
 
-  const pagesDirectory = systemPath.posix.join(pagesPath)
-  const pagesGlob = `${pagesDirectory}/**/*.{${exts}}`
+  const pagesDirectory = systemPath.resolve(process.cwd(), pagesPath)
```
wardpeet: should we move to posix here?

packages/gatsby-plugin-page-creator/src/gatsby-node.js
paulmelnikow: `micromatch` is a replacement for `minimatch`; it doesn't include the filesystem parts of `glob`. [`glob-fs`](https://www.npmjs.com/package/glob-fs) exists based on `micromatch` but according to [the github page](https://github.com/micromatch/glob-fs) it's experimental and doesn't work on Windows.

Perhaps declaring a dependency on `minimatch` instead of `micromatch` in the other places would be a good way to go.

packages/gatsby-plugin-page-creator/src/create-path.js
```diff
@@ -1,11 +1,9 @@
 // @flow
 import parsePath from "parse-filepath"
```
paulmelnikow: Our of curiosity, is there any reason `parse-filepath` is being used in place of the native `path.parse`? It seems this code would not be bundled.

If it would be helpful, I could follow up with a PR to remove that dependency.

packages/gatsby-plugin-page-creator/src/gatsby-node.js
Warped2713: The [createPages API](https://www.gatsbyjs.org/docs/actions/#createPage) example doesn't use posix for resolving component path, so I figure it's best to let Node use the native OS syntax for finding the component files.

```js
createPage({
  path: `/my-sweet-new-page/`,
  component: path.resolve(`./src/templates/my-sweet-new-page.js`),
  // The context is passed as props to the component as well
  // as into the component's GraphQL query.
  context: {
    id: `123456`,
  },
})
```

packages/gatsby-plugin-page-creator/src/gatsby-node.js
Warped2713: The [createPages API](https://www.gatsbyjs.org/docs/actions/#createPage) example doesn't use posix for resolving component path, so I figure it's best to let Node use the native OS syntax for finding the component files.

```js
createPage({
  path: `/my-sweet-new-page/`,
  component: path.resolve(`./src/templates/my-sweet-new-page.js`),
  // The context is passed as props to the component as well
  // as into the component's GraphQL query.
  context: {
    id: `123456`,
  },
})
```

packages/gatsby-plugin-page-creator/src/gatsby-node.js
Warped2713: Since `pagesDirectory` is only being used is to form the component path in createPages, it should be consistent and use the same OS path implementation as the previous two comments.

paulmelnikow: Since `process.cwd()` and `pagesPath` are system-dependent paths, I suspect `systemPath.resolve` is right.

packages/gatsby-plugin-page-creator/src/gatsby-node.js
paulmelnikow: Am I right that `pagesDirectory` is a system-dependent path (with `\` on Windows), while `filePath` is posix? In that case `systemPath.join` seems correct, as it would interpret both `\` and `/` as a path separators.

paulmelnikow:
Oops, @Warped2713 you were faster. You posted your comments while I was drafting.

packages/gatsby-plugin-page-creator/src/gatsby-node.js
Warped2713: Hmm... chokidar uses micromatch (via anymatch), but as paulmelnikow says, micromatch doesn't provide any filesystem utility so we'd need to initialize the files array some other way.

**gatsby-source-filesystem** handles the initialization by using chokidar with a state machine (xstate). Would you prefer that over a glob dependency? It'd be a bit of a refactor.

https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-source-filesystem/src/gatsby-node.js

packages/gatsby-plugin-page-creator/src/create-path.js
Warped2713: It doesn't look like any of the additional `parse-filepath` utilities are being used, so it makes sense to remove this dependency in favor of the native `path.parse`.

Go ahead 👍

paulmelnikow: Cool, if @wardpeet agrees I'll do that after this gets merged.

paulmelnikow:
@Warped2713 Is it worth updating the title to reflect that it also properly ignores tests on windows? (i.e. the bug from badges/shields#3208)

Warped2713:
@paulmelnikow Sure. It's a bit long now, but probably easier for others to find.

packages/gatsby-plugin-page-creator/src/gatsby-node.js
wardpeet: TIL @paulmelnikow you can ignore this 👍

@Warped2713 we could go for chokidar to listen to pages but probably not worth it unless this one is getting slow. Chokidar is much more memory hungry. We might want to change this in the future.

packages/gatsby-plugin-page-creator/src/create-path.js
wardpeet: seems right, parse-filepath is a polyfill and parse.path is supported in all versions we support so please remove 👍

wardpeet:
I'm going to give this a spin on windows & linux to see if everything works. Code looks good, thank you for your patience

wardpeet:
Works like a charm on windows and linux! thanks for all your patience and explaining me everything I need to know to get this merged.

paulmelnikow:
Thank you both so much!

PR #11313, opened 2019-01-26
Add mention of options allowed

FabioRosado:
Hello, 

I had to look at how to add google fonts to my project so I thought it could be good to include these in the plugin documentation.

Not sure if this will be a good idea for others, but it might save some clicks and tabs 😄 

Thank you for the time

### DIFF
--- a/docs/docs/typography-js.md
+++ b/docs/docs/typography-js.md
@@ -67,6 +67,16 @@ export default typography
 
 Font sizes of all elements in Typography.js grow and shrink in relation to the `baseFontSize` defined above. Try playing around with this value and see the visual difference it can make to your website.
 
+`Typography` allows you to specify different kinds of options:
+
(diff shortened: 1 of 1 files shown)


LekoArts:
Hi, thanks for the PR! However I think that the line `A full list of options that can be specified when defining a new typography can be found at [Typography.js](https://kyleamathews.github.io/typography.js/).` is enough as we then don't have to keep the options in sync with the typography plugin. What's the reasoning behind your PR except clicking this link?

FabioRosado:
I had to do a bit of search on how to make the Google fonts working and what sort of options available.

I decided to add these options to the documentation because the typography page is quite long and if you are using a mobile device the design tool that allows you to try the typography is quite big and makes it hard to read the text.

I do understand that this could make it harder to keep the documentation up to date, so I understand if you don't think this is a good idea 😄

Issue #11275, opened 2019-01-25
cache.set does not resolve with value being cached


mgmolisani:
## Description

When using the cache in plugins, calling cache.set always resolves to `undefined` when it appears the intention was to have it resolve with what was stored (pretty standard practice).

### Steps to reproduce

Attempt to access the return value of `await cache.set('a', 'b')` in your gatsby-node.js file inside of an api. Mine was in the resolve of a new field in `exports.setFieldsOnGraphQLNodeType`.

### Expected result

For the above code, the expected response is `'b'`

based on the content in [cache.js](https://github.com/gatsbyjs/gatsby/blob/e1f1a9864f9b88ca33710f7b5fa837f5153a786d/packages/gatsby/src/utils/cache.js#L47):

``` javascript
  set(key, value, args = {}) {
    return new Promise(resolve => {
      this.cache.set(key, value, args, (_, res) => resolve(res))
    })
  }
```
it should resolve with the response of what was stored. It appears the cache-manager package callback argument suggests a resolve value exists (see: [cache-manager](https://github.com/BryanDonovan/node-cache-manager/blob/deeddaf4276b8da5cab4283f19955341ec85baff/lib/multi_caching.js#L197)). This is what is used under the hood for cache.set but one level deeper into the 'async' package (see: [async](https://github.com/gatsbyjs/gatsby/blob/e1f1a9864f9b88ca33710f7b5fa837f5153a786d/packages/gatsby/src/utils/cache.js#L47)), we see the third argument callback only gets passed an error or nothing, never a second result argument. I think aync.each is like Array.forEach where it has no returned array unlike map functions.

So my best guess here, is 'cache-manager' promised a result, but under the hood, it's callback is never defining that result since aync.each provides no args or err to that callback hence it bubbles back up to us as `undefined`. The Gatsby tests for this only check if the response is thenable, which it does return a Promise that just resolves to `undefined`.

### Actual result

Always `undefined`. All packages that implement cache await the response to make sure something gets stored with the correct key but then return what they passed to the cache instead of a response from the cache which is dangerous if there was ever a mismatch as subsequent cache.get could all be different than the initial value.

### Environment

Got errors running this. I definitely have more than just Edge installed.... I'm not a monster.
```
  System:
    OS: Windows 10
    CPU: (4) x64 Intel(R) Core(TM) i5-6300U CPU @ 2.40GHz
  Binaries:
    npm: 6.5.0 - C:\Program Files\nodejs\npm.CMD
[...]

PR #11327, opened 2019-01-27
feat(gatsby): update cache.set to resolve with set value

mgmolisani:
## Description

Cache set method will now resolve to the stored value. Both cache set and get now also reject on cache errors.

Also, I have never written Jest tests so this was a pretty exciting learning experience. Mocking was not very intuitive, especially due to the resolve function being in the callback of the get and set cache manager package. We also don't actually talk to the cache so the get tests feel a little wonky. Also, I do not know the shape of the error object sent from the cache manager or what actually can cause it so I mocked its existence with `!undefined` which is obviously just `true` but I wanted it to read as something that was defined and not just true. Is there a better/conventional way to say `err` is defined? Would `{}` be a better solution? Interested if anyone has a better testing strategy than just forcing the error argument to exist or not.

## Related Issues

Addresses #11275


mgmolisani:
General Jest question after continuing to think of Jest best practices. For the code where I need the `err` and `res` variables set dynamically per test, would it be better to declare regular variables and clean them up with an afterEach? Or if calling them as mocked functions like they are now, is it better practice to use `jest.fn().mockReturnValue()` when we are using them like variables and not functions since the body of the function is always just the returned value? Really deep diving into Jest but it doesn't seem like mocking variables isn't explained in the docs (likely because you can literally just use variables or callback mocked functions to return a value as stated above). What do other people think?

packages/gatsby/src/utils/__tests__/cache.js
```diff
@@ -98,4 +105,36 @@ describe(`cache`, () => {
...
+    it(`resolves to the value it cached`, () => {
+      const cache = getCache()
+
+      return expect(cache.set(`a`, `b`)).resolves.toBe(`b`)
+    })
+
+    it(`rejects on caching error`, () => {
+      const cache = getCache()
+
+      mockErrorValue.mockImplementationOnce(() => !undefined)
```
DSchau: Could we use `mockReturnValueOnce`?

mgmolisani: Yes, we can. After further learning and playing with Jest, I tested with this exact suggestion and mockReturnValueOnce would be much cleaner in the case of both variables declaring both initially as just `jest.fn()`s.

Very new to this whole comment system and PRs. Should I just commit the change on my branch with the suggested change?

DSchau:
@mgmolisani yep! Just update with a new commit (as many as you want actually!). Come merge time, we'll just squash all those commits down into one nice, clean commit message :)

mgmolisani:
@DSchau Updated with suggestions (sorry if this is basically just a duplicate notification).

packages/gatsby/src/utils/cache.js
```diff
@@ -40,13 +40,25 @@ class Cache {
 
   get(key) {
     return new Promise(resolve => {
-      this.cache.get(key, (_, res) => resolve(res))
+      this.cache.get(key, (err, res) => {
+        if (err) {
+          throw new Error(
+            `The value cached with key: ${key} could not be retrieved.`
```
pieh: will this throw if value is not yet cached? if so this would be change in behaviour

wardpeet: it's a breaking change indeed, if it's not in the cache we resolve to null or undefined probably.

mgmolisani: On another note, `set` would have the same issue where it would now have to be caught. Do we want to handle that differently than the current `throw` solution. 

(I apologize if my initial pull request seemed to have no regard for breaking changes. I originally wasn't planning to handle the errors and haven't ever really had to work on in production code. I appreciate all the careful reviews.)

pieh: Don't worry @mgamsjager - I'm just trying to gather all the facts here and make sure we don't introduce breaking changes by accident ;)

mgmolisani:
Updated. Now resolves to `undefined` on errors. Is this the behavior that is desired?

(Also, as a git contributing noob, I didn't know that I should pull the official master branch into my topic branch to sync with the prod build. Caused a bunch of test failures. Is this something that is worth a pull request to the docs for?)

PR #11328, opened 2019-01-27
chore(gatsby-transformer-remark): bump underscore.string version

### LINKED ISSUE #10958: underscore.string < 3.3.5: Moderate vulnerability for gatsby-transformer-remark
## Description

npm audit found 1 moderate severity vulnerability

### Actual result

```
yarn audit v1.12.3
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ moderate      │ Regular Expression Denial of Service                         │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ underscore.string                                            │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >=3.3.5                                                      │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ gatsby-transformer-remark                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ gatsby-transformer-remark > underscore.string                │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://nodesecurity.io/advisories/745                       │
└───────────────┴──────────────────────────────────────────────────────────────┘
1 vulnerabilities found - Packages audited: xxxxx
Severity: 1 Moderate

```
### Environment

```
  System:
    OS: macOS High Sierra 10.13.6
    CPU: (8) x64 Intel(R) Core(TM) i7-4980HQ CPU @ 2.80GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 11.1.0 - /usr/local/bin/node
  Languages:
    Python: 3.7.1 - /Users/mthomas/.pyenv/shims/python
  Browsers:
    Chrome: 71.0.3578.98
    Firefox: 64.0
    Safari: 12.0.2
  npmPackages:
    gatsby: 2.0.90 => 2.0.90 
    gatsby-plugin-react-helmet: ^3.0.2 => 3.0.2 
    gatsby-plugin-sharp: ^2.0.13 => 2.0.13 
    gatsby-plugin-typography: ^2.2.1 => 2.2.1 
    gatsby-remark-autolink-headers: ^2.0.11 => 2.0.11 
    gatsby-remark-images: ^3.0.0 => 3.0.0 
    gatsby-remark-prismjs: ^3.1.0 => 3.1.0 
    gatsby-source-filesystem: ^2.0.9 => 2.0.9 
  npmGlobalPackages:
    gatsby-cli: 2.4.8
```

saranshkataria:
## Description

Removed underscore.string from gatsby-transformer-remark and replaced it with lodash truncate

## Related Issues
Fixes #10958


packages/gatsby-transformer-remark/src/extend-node-type.js
```diff
@@ -453,11 +452,11 @@ module.exports = (
...
               pruneLength - (unprunedExcerpt.length - lastTextNode.value.length)
             if (!truncate) {
-              lastTextNode.value = prune(
-                lastTextNode.value,
-                amountToPruneLastNode,
-                `…`
-              )
+              lastTextNode.value = _.truncate(lastTextNode.value, {
+                length: amountToPruneLastNode,
+                separator: /,?\.* +/,
```
DSchau: Is this the behavior of `prune`, or why exactly do we need this? Also let's be explicit with the space, e.g. `\s` 

Also a note that I generally like to encapsulate this as a var if we're going to re-use it a few times, e.g. if we hoist to near the top of the file like

```js
const EXCERPT_SEPARATOR = /,?\.*\s+/
```

DSchau: Right - but this could theoretically be a breaking change then, right?

saranshkataria: Without the separator, it definitely was a breaking change. With it in place, I tried to mimic it to existing prune implementation. Existing snapshot tests are passing with this. There might be some boundary conditions that I might have missed, can't be 100% sure about that.

pieh:
Can we just bump version of `underscore.string`? It's fixed and even now only way to have old version installed is we lock file locks it down (because version range selector allow installing updated version)?

saranshkataria:
I tried to make the functionality similar to that of prune, but was not able to. Reverted the changes and bumped up the underscore version for now to fix the original issue. Will give removing underscore another shot sometime later.

Issue #11198, opened 2019-01-21
Build fails after with "Unexpected token" in `async-requires` after upgrade to 2.0.84


dlindahl:
## Description

Running `rm -rf .cache && rm -rf public && gatsby-build` works fine under 2.0.83. After upgrading to 2.0.84, an error is thrown, aborting the build.

### Steps to reproduce

The only I am taking is running `gatsby build`. After downgrading back to 2.0.83, the issue disappears.

### Expected result

The build should complete successfully

### Actual result

The build quits with the following error:

```
success onPostBootstrap — 0.202 s

info bootstrap finished - 6.171 s


error Generating JavaScript bundles failed


  Error: ./.cache/async-requires.js 8:11
  Module parse failed: Unexpected token (8:11)
  You may need an appropriate loader to handle this file type.
  | exports.components = {
  |   "component---node-modules-gatsby-plugin-offline-app-shell-js": function componentNodeModulesGatsbyPluginOfflineAppShellJs() {
  >     return import("/Users/dereklindahl/Work/APP/node_modules/gatsby-plugin-offline/app-shell.js"
  |     /* webpackChunkName: "component---node-modules-gatsby-plugin-offline-app-shell-js" */
  |     );
   @ ./.cache/production-app.js 18:0-45 21:23-36 26:23-36
```

### Environment

```
npx gatsby info -- --clipboard                 

  System:
    OS: macOS High Sierra 10.13.6
    CPU: (4) x64 Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
    Shell: 5.4.2 - /usr/local/bin/zsh
  Binaries:
    Node: 10.14.0 - ~/.nodenv/versions/10.14.0/bin/node
    Yarn: 1.12.3 - /usr/local/bin/yarn
    npm: 6.4.1 - ~/.nodenv/versions/10.14.0/bin/npm
  Languages:
    Python: 2.7.10 - /usr/bin/python
  Browsers:
    Chrome: 71.0.3578.98
    Firefox: 63.0.3
    Safari: 12.0.2
  npmPackages:
    gatsby: 2.0.84 => 2.0.84 
    gatsby-image: ^2.0.25 => 2.0.25 
    gatsby-mdx: ^0.2.0 => 0.2.0 
    gatsby-plugin-canonical-urls: ^2.0.8 => 2.0.8 
    gatsby-plugin-google-analytics: ^2.0.8 => 2.0.8 
[...]

PR #11331, opened 2019-01-27
fix: use require instead if the ESM import statement

DanielRuf:
## Description

This fixes an issue with the `build` script.

In the dev mode it seems we have the babel transform plugin for the engine and dynamic require / import.

But the build does not and we have to use `require`. In fact the code in this file uses `exports` but not the `require`. I guess this was overseen.

We should use the CJS way to use them (require) instead of the ESM one (import).

https://github.com/gatsbyjs/gatsby/blame/master/packages/gatsby/src/internal-plugins/query-runner/pages-writer.js#L97

https://github.com/gatsbyjs/gatsby/blame/master/packages/gatsby/src/internal-plugins/query-runner/pages-writer.js#L104

## Related Issues

https://github.com/gatsbyjs/gatsby/issues/11198

### DIFF
--- a/packages/gatsby/src/internal-plugins/query-runner/pages-writer.js
+++ b/packages/gatsby/src/internal-plugins/query-runner/pages-writer.js
@@ -94,14 +94,14 @@ const preferDefault = m => m && m.default || m
   asyncRequires += `exports.components = {\n${components
     .map(
       c =>
-        `  "${c.componentChunkName}": () => import("${joinPath(
+        `  "${c.componentChunkName}": () => require("${joinPath(
           c.component
         )}" /* webpackChunkName: "${c.componentChunkName}" */)`
     )
     .join(`,\n`)}
 }\n\n`
 
-  asyncRequires += `exports.data = () => import(/* webpackChunkName: "pages-manifest" */ "${joinPath(
+  asyncRequires += `exports.data = () => require(/* webpackChunkName: "pages-manifest" */ "${joinPath(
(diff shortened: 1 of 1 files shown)


DanielRuf:
Hm, does my change break the tests? It seems so.

pieh:
It seems like You are changing dynamic import into static one here. What about doing this other way and making this entirely ESM:
`exports.data` => `export const data`?
It's also suprising that it break things for some people, but not for others :/

pieh:
Do you have example project that is failing currently that I could investigate?

DanielRuf:
@pieh see https://github.com/gatsbyjs/gatsby/issues/11198#issuecomment-457905564 for a repository for testing the problem. All previous gatsby releases have this issue so it might be because of another dependency.

DanielRuf:
It seems a new install does not havr this issue. I'll do a diff between the working and not working setup.

DanielRuf:
> `exports.data` => `export const data`?

This does not change anything, the needed runtime transforms are still not applied.

pieh:
Yeah, I cloned the reproduction and had same problem when installing with `npm` - but after wiping lock file and reinstalling node_modules with `npm` problem went away, so to me it seems like some dependency issues

DanielRuf:
Seems so. But so far we do not know the exact cause. I'm closing the PR.

DanielRuf:
It seems it still happens with npm, but not Yarn.

PR #11339, opened 2019-01-27
add company website starter

kmelve:
## Description

This PR adds Sanity.io’s company website starter to the starter library.


docs/starters.yml
```diff
@@ -1710,3 +1710,19 @@
     - Registration form
     - Signup form
     - User sign in
+- url: https://example-company-website-gatsby-sanity-combo.netlify.com/
+  repo: https://github.com/sanity-io/example-company-website-gatsby-sanity-combo
+  description: This examples combines Gatsby site generation with Sanity.io content management in a neat company website.
+  tags:
+    - sanity.io
```
LekoArts: All over the website you only say "Sanity", so why "sanity.io" here?

docs/starters.yml
```diff
@@ -1710,3 +1710,19 @@
...
+    - Headless CMS
+    - Company website
```
LekoArts: Please remove that tag, it's not specific enough

```suggestion
```

docs/starters.yml
kmelve: We mostly use `sanity.io` in contexts outside the website to avoid puns/ambiguity, and make it clearer. If you think it should be just `Sanity` that's ok though.

LekoArts: I want to keep it coherent that's why I asked :)
https://github.com/gatsbyjs/gatsby/pull/11098
This will also add "Sanity" and the starters using prismic.io also use "Prismic". Are you fine with choosing "Sanity" here?

PR #11340, opened 2019-01-27
feat(gatsby-plugin-google-analytics): Pass event object to the user's onClick callback in OutboundLink

asistapl:
## Description
Pass event object to the user's onClick callback


dance2die:
Thank you guys. This is indeed working now~

I had a trouble of having to add event listeners for ouboundlink generated anchors but now it works perfect.

https://github.com/dance2die/landing-page-v3-gatsby/blob/master/src/components/util/index.js#L5

PR #11365, opened 2019-01-28
docs(gatsby-plugin-netlify-cms): remove old `stylesPath` option from example code

asistapl:


asistapl:
@pieh Is that comment now still accurate?

asistapl:
Btw. This code `CMS.registerPreviewStyle('cms.css')` generates error when not importing any css files. I'm using emotion and I'm seeing error like this in NetlifyCMS's edit page:

"Refused to apply style from 'https://_mysite_.netlify.com/admin/cms.css' because its MIME type ('text/html') is not a supported stylesheet MIME type, and strict MIME checking is enabled."

Not a big deal but maybe we could leave that style registration to the user or change behavior of that option?

pieh:
> Not a big deal but maybe we could leave that style registration to the user or change behavior of that option?

I don't have enough context on `netlify-cms` plugin, but it does seem like sane thing to do to check is `cms.css` is there before trying to register it. Just not sure how to do it

pieh:
Ok, so it's a bit annoying but at least it's not breaking. Let's merge this PR as-is, as the missing .css is not strictly related problem

asistapl:
I think so too, it would probably require to write a custom webpack plugin just to know if that line should be included.
Thanks!

PR #11378, opened 2019-01-29
Fix loki regex operator on nested fields

Moocar:
While building a test site, I noticed that loki doesn't have the same behavior as sift for regex on nested fields. Sift first checks to see if a nested field exists before running the regex. Whereas loki happily runs the regex on `undefined`. And unfortunately in js, `/.*/.test(undefined)` returns true.

To test, run the tests with the env var `GATSBY_DB_NODES=loki`. E.g

```
GATSBY_DB_NODES=loki yarn test
```


freiksenet:
Sorry about that. I actually fixed it when I noticed it locally, but didn't think it was actively used. Thanks!

Issue #4976, opened 2018-04-13
GA Plugin and GDPR


kbariotis:
## Description

Google's new [EU user consent policy](https://www.google.com/about/company/consentstaging.html) states that Google's products, such as Google Analytics must have user's consent in order to work in a page. That means that a website must take the user's consent and then enable the Google Analytics plugin. With that in mind, I would like to see some sort of a toggle to enable/disable the Gatsby Google Analytics plugin on demand.

### Expected result

I would like to be able to have Google Analytics plugin disabled by default. Then ask the user's consent and only then enable it.

Here are some references on GA tracking code:

1) https://developers.google.com/analytics/devguides/collection/gtagjs/user-opt-out
2) https://developers.google.com/analytics/devguides/collection/gtagjs/#disable_page_view_tracking

sebastienfi:
To know if the user is browsing from EU, we'd have to use an IP Geolocation service. I don't know any that's free.
As an easy workaround, we could detect the timezone of the user and from there extrapolate it's location. 
@kbariotis would you like to submit a PR doing that ?

kbariotis:
No @sebastienfi I don't think this is Gatsby's responsibility at all. This must be left at the owner's choice. Gatsby should just give control to the owner whether the plugin is enabled or not and allow to enable it at run time, rather than built time.

jurosh:
Btw. using google analytics is OK (even in Europe) if you anonymize IPs and also disable cross session (bad) cookies.

With this config tracking should be fine:
```
gtag('config', 'UA-...', {
      'anonymize_ip': true,
      'cookie_expires': 0
  });
```
Ofc. you loose cross-session user tracking feature of analytics, but it might be good for many developers who jsut want to do basic tracking and don't care about GDPR.

Currently plugin allows to anonymize IP, but doesn't allow (cannot see option) to enable session only cookies - `cookie_expires`.

As I understand (don't take as lawyer advice) whole GDPR (no consent) compliant code might be:
```
<script async src="https://www.googletagmanager.com/gtag/js?id=UA-XYZ"></script>
<script>
  window.dataLayer = window.dataLayer || [];
  function gtag(){dataLayer.push(arguments);}
  gtag('js', new Date());
  gtag('config', 'UA-XYZ', {
      'anonymize_ip': true,
      'cookie_expires': 0
  });
</script>
```

discordianfish:
[...]

PR #11379, opened 2019-01-29
feat(gatsby-plugin-google-tagmanager): GDPR consent + default options

crstnio:
## Description

As of the GDPR it is a best practice to gather the user agreement on setting cookies and/or enabling tracking scripts. It's the developers responsibility to enable tracking/cookies on user consent and to remove cookies if the user declines afterwards.

But, we should give the developers options to make that possible and their jobs easier.

This PR allows two more options:

- `gdpr` to disable Google Tag Manager on page load. Set to `false` by default.
- `gdprConsent` for a global variable (inside `window`) that holds the users consent. Results in `window['gdprConsent']` by default.

`gdpr` could be set by an environment variable, that is set depending on the users location (EU or not) and controlled by the ADN.

Again, it is the developers responsibility to gather the consent and set `window[gdprConsent]`. This value should come from a cookie and I plan to develop a plugin/component that enables the user to agree, sets both the cookie and variable, executes optional functions and renders blocked children.

The code is now wrapped inside a `gtmCode()` function that is needed to enable Google Tag Manager without reloading the page.

### Object destructuring

This feature introduces default options by destructuring the options object. It seemed appropriate to add this with the GDPR feature.

The default value of `includeInDevelopment` (README) is now set too.

## Related Issues

@sebastienfi [uses GA](https://github.com/gatsbyjs/gatsby/issues/4976#issuecomment-381344919) (comment) in a similar way, but with a `window.gdpr-agreed` variable.


wardpeet:
Thanks for creating this PR, I'm just not sure this is a good fix for gdpr & googletagmanager.

Googletag manager doesn't break GDPR regulations. It's what googletag is configured to do, that might have issues with GPDR. Adding these gdpr variables in my opinion isn't a correct fix and might make this plugin a bit more complicated to use.

The correct fix is to actually push something on your DataLayer that tells gtm to load or not load scripts.

If this is hard to do I'm ok with creating a shouldLoad option that has a function to skip loading which is more general approach than just checking a global variable.

What do you think? (also want to involve @gatsby/core into this discussion)

cardiv:
Hm, I tried the approach of using `dataLayer` events that trigger tags in GTM.

To control this you would need to add a script (for browser) + the `dataLayer` (ssr) before GTM is loaded. Time for another plugin or build this right into `gatsby-plugin-google-tagmanager`?

wardpeet:
Yes, that's true. I just don't think `GDPR` should be part of GTM. What do you think of creating a shouldLoad option where you can pass in a function?

cardiv:
The `shouldLoad` option would only be evaluated at build time, right?

Since the user needs to agree in the browser, this would be no solution I think.

As you said, `dataLayer` seems the way to go. But that requires the ability to push before GTM both at build time and in the browser.

We could add an option that takes a function which will be injected before the GTM code and allows for early `dataLayer` actions. Maybe the GTM code could be blocked until `dataLayer` is ready.

wardpeet:
I'll make 2 examples one with dataLayer and one with the shouldLoad function. I was thinking a function that's ran inside the browser.

wardpeet:
@cardiv would something like this work?

packages/gatsby-plugin-google-tagmanager/README.md
```diff
@@ -12,14 +12,19 @@ Easily add Google Tagmanager to your Gatsby site.
...
+    resolve: "gatsby-plugin-google-tagmanager",
     options: {
       id: "YOUR_GOOGLE_TAGMANAGER_ID",
 
       // Include GTM in development.
       // Defaults to false meaning GTM will only be loaded in production.
       includeInDevelopment: false,
 
+      // datalayer to be set before GTM is loaded
+      // should be a stringified object or object
```
cardiv: `generateDefaultDataLayer` checks for a function too. So not only objects are supported, right?

packages/gatsby-plugin-google-tagmanager/src/gatsby-ssr.js
```diff
@@ -1,46 +1,70 @@
...
+  })(window,document,'script','dataLayer', '${id}');`
+
+const generateGTMIframe = ({ id, environmentParamStr }) =>
+  oneLine`<iframe src="https://www.googletagmanager.com/ns.html?id=${id}${environmentParamStr}" height="0" width="0" style="display: none; visibility: hidden"></iframe>`
+
+const generateDefaultDataLayer = (dataLayer, reporter) => {
+  let result = `window.dataLayer = window.dataLayer || [];`
+
+  if (typeof dataLayer === `function`) {
+    result += `window.dataLayer.push((${dataLayer})()));`
```
cardiv: Aren't there way too much brackets at the end? Or am I missing something here?

cardiv:
Nice work @wardpeet, thank you! I am a bit confused about two code parts but this could handle some more use cases.

packages/gatsby-plugin-google-tagmanager/src/gatsby-ssr.js
wardpeet: good catch 1 to many 😄

packages/gatsby-plugin-google-tagmanager/README.md
wardpeet: thanks will have to update this comment, it indeed allows a function and an object :) stringified object isn't possible anymore.

wardpeet:
@cardiv could you test if it works now. I updated the test so we catch my errors 👍

wardpeet:
@cardiv ping :) I would love to get this merged.

freiksenet:
There is some cleanup that needs to be done before merging. @wardpeet will address this as soon as he has time. Thank you for your patience!

PR #11393, opened 2019-01-29
chore(www): bump gatsby version

pieh:
just couple of bump to be more up to date


pieh:
our build-runner will not update dependencies if package.json doesn't change - so this is just hint for build-runner to actually update deps

pieh:
alternatively we can kill the pod, which will have similiar effect

Issue #10438, opened 2018-12-12
Remaining steps for launching loki.js support


KyleAMathews:
- [ ] enable for gatsbyjs.org builds
- [ ] Create more tests for edge cases on queries & add loki.js to CI to make sure it's passing all snapshots
- [ ] incremental rollout to folks? Enable for 5% of users and then 10% and then 20%, etc. 

/cc @Moocar @pieh

Moocar:
@KyleAMathews How would an incremental rollout work? How would we enable it on a subset of users? I was thinking we could mention it on the gatsby gazette so that friendly users could try it out first.

@pieh wrote a great tool that compares query output between sift and loki. I think we should be able to extend this functionality to compare example sites during the CI phase. I can add this to my TODO

KyleAMathews:
It'd be automatic. We'd hard code the percentage of the rollout and on bootstrap, if the site wasn't already assigned to one or the other, we'd do Math.random and use that to put them in them in the Loki experiment or redux.

We should definitely push for more friendlies to try it first.

Moocar:
One other thing is that we're using [@moocar/lokijs](https://www.npmjs.com/package/@moocar/lokijs), as the changes to loki haven't been released yet. I'll ping the author and see what the timeline looks like.

KyleAMathews:
@thebigredgeek could you take care of adding the environment variable to the gatsbyjs.org build? The env var is `GATSBY_DB_NODES=loki`

KyleAMathews:
Not stale

PR #11403, opened 2019-01-29
chore(gatsby): Run loki builds on circle ci

Moocar:
## Description

Enable loki feature flag during CI. This also adds a script that runs a site in both redux and loki mode and then compares the query results to make sure they're identical. Props to @pieh for the initial implementation.

This is my first time editing the CI pipeline. Let me know if I can do anything else to test etc.

## Related Issues

- #10438 - Remaining steps for launching loki.js support


pieh:
with some hacking it finally do *something*, now need to figure out why it can't find redux-state.json (probably because it doesn't persist state before finishing build - will take a look more in depth tomorrow)

Moocar:
@pieh thanks for bashing your head on a wall on my behalf :)

pieh:
I thought about this a bit over night, and here's some thoughts:
 - the fact that we can finish build without actually persisting redux-state is pretty bad - need to add waiting for the autosave before exiting build process (need to research how much time to build step this adds)
 - using gatsbygram (or any other example site) is not ideal, because we most certainly don't use all type of queries there - I will try to create e2e site that will run queries in more controlled environment (probably use gatsby-node's createPages) - so it's more test like and so we don't have to rely on `redux-state` for asserting correctness

pieh:
Here's what I had in mind with not using gatsbygram - https://github.com/gatsbyjs/gatsby/pull/11502 - I added just very basic queries and intientionally broke custom comparators to show how it fail when query results don't match

Moocar:
@pieh I love the idea of testing query results inside a running process vs having to deal with the filesystem. But I'm not a fan of creating some arbitrary data/queries and testing those at the integration level, because it's testing very similar things to [run-query.js](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/schema/__tests__/run-query.js) tests, just with a lot of extra processing. 

In a perfect world, I would use generative testing (e.g [testcheck-js](https://github.com/leebyron/testcheck-js)) to test loki/sift comparison. We would generate random data and random queries, run them against loki/sift and ensure the results were identical. In theory, we could use [gql-generator](https://github.com/modelo/gql-generator) to create the queries. Creating the data should be relatively easy using test-check generators. 

But that approach still won't be able to search the entire problem space (and will be very inefficient). That's where testing actual sites has value. We could in theory run a comparison test on hundreds of different gatsby sites.

pieh:
You are of course right - having to define arbitrary queries is main problem here. It's just gatsbygram really has just 3 queries, so this is very close to my arbitrary queries - 
- https://github.com/gatsbyjs/gatsby/blob/afa4dade603cb5ac3387843ef2ea2a90848ac487/examples/gatsbygram/src/pages/index.js#L203-L223
- https://github.com/gatsbyjs/gatsby/blob/afa4dade603cb5ac3387843ef2ea2a90848ac487/examples/gatsbygram/src/templates/post-page.js#L42-L47
- https://github.com/gatsbyjs/gatsby/blob/afa4dade603cb5ac3387843ef2ea2a90848ac487/examples/gatsbygram/gatsby-node.js#L25-L33

So this isn't very comprehensive test site as well :(

I'm intrigued by https://github.com/modelo/gql-generator - I'll experiment with it a bit - if we use this instead of arbitrary queries I do think we could get away with my approach (if I actually create data / config to that will result in schema that uses all available resolvers in gatsby), will post update if I can make it work in reasonable amount of time ;)

> We could in theory run a comparison test on hundreds of different gatsby sites.

Ideally, yeah - but it's not feasible yet I think

Moocar:
@pieh Nice! I imagine the queries won't make "sense" (no sane developer would write them), but as long as the outputs of sift and loki queries are consistent, that's all we care about.

When paired with a quickcheck library like testcheck-js, we'll be able to say "just run it 100 or 1000 times" and tell us which cases fail.

pieh:
Just to move things forward - because `compare_loki_sift-gatsbygram` is blocked by gatsby potentially finishing before persisting state, let's just merge running `GATSBY_DB_NODES=loki yarn jest -w 1` for now - I'll do separate PR for that. I'll work on fixing up `redux-state` stuff (will do this at the end of the week, as I have work already queued up for at lest 2 days)

Moocar:
@pieh Sounds good. I'll remove the other code from this PR for now. It's still in commit history at https://github.com/gatsbyjs/gatsby/commit/af052e6c2c64207d351ecabe22e1c0a43b7424ce.

pieh:
Thanks @Moocar - I was swamped with discussions today and didn't do much of things I wanted to do ;(

Moocar:
@pieh done. I created a new branch for `compare-loki-sift` at https://github.com/Moocar/gatsby/tree/compare-loki-sift. I can definitely try and get the redux-state stuff working. I might just need your help learning about how to test on CI.

pieh:
Don't worry about failing e2e tests - this is because this was branched before some of graphql packages changes ;)

PR #11410, opened 2019-01-30
fix(blog): Re-arrange content after code was moved between snippets

m-muhsin:
Hey @DSchau! I read the blog post once again from localhost after those changes. I felt we need to change the place of certain paragraphs after the code within those two snippets was moved between each other.


wardpeet:
I summon the @gatsbyjs/docs team to give this a review ^^

pieh:
Thanks @m-muhsin for continuously improving on the blog post!

PR #11414, opened 2019-01-30
feat(gatsby-plugin-manifest): make favicon link tag optional

### LINKED ISSUE #11308: [plugin-manifest] Add option to opt-out of generating favicon (vs. pwa icon)
## Summary

Currently `gatsby-plugin-manifest` automatically generates and links to a favicon as well as PWA style icons when in automatic or hybrid mode, and there's no way to opt-out of this behaviour without going full manual mode. 

The plugin should expose an opt-out flag for this, as it already does for other problematic on-by-default features (eg: theme-color tags). Especially since a) favicons are arguably out of scope for a PWA manifest-style featureset anyway, and b) it's trivial for users to include their own single favicon asset.

### Basic example

Easiest way to accomplish this is to add another optional config parameter to the plugin 

```js
module.exports = {
  plugins: [
    {
      options: {
        name: `GatsbyJS`,
	...
        icon: `src/images/icon.png`,
	include_favicon: true // Defaults to true to be non-breaking
      }
    }
  ]
}
```

### Motivation

It's very common to want very different assets for a favicon and PWA icon — eg: a small transparent icon for the favicon, and a more intricate icon with a background for the PWA manifest. While you can achieve this using 'manual' mode, that's a frustrating workaround for a fairly large use-case.

gurpreet-hanjra:
Related issue - https://github.com/gatsbyjs/gatsby/issues/11308

## Description

Makes favicon link tag optional using plugin option `include_favicon`

### DIFF
--- a/packages/gatsby-plugin-manifest/README.md
+++ b/packages/gatsby-plugin-manifest/README.md
@@ -233,9 +233,9 @@ module.exports = {
 }
 ```
 
-## Include `favicon` link tag
+## Exclude `favicon` link tag
 
-Inserts `<link rel="shortcut icon" href="/favicon.png" />` link tag to html output. You can set `include_favicon` plugin option to `true` to opt-in of this behaviour.
+Excludes `<link rel="shortcut icon" href="/favicon.png" />` link tag to html output. You can set `include_favicon` plugin option to `false` to opt-out of this behaviour.
 
 ```javascript:title=gatsby-config.js
 module.exports = {
@@ -251,7 +251,7 @@ module.exports = {
         display: `standalone`,
         icon: `src/images/icon.png`, // This path is relative to the root of the site.
         theme_color_in_head: false, // This will avoid adding theme-color meta tag.
-        include_favicon: true, // Include favicon
+        include_favicon: false, // This will exclude favicon link tag
       },
     },
   ],
--- a/packages/gatsby-plugin-manifest/README.md
+++ b/packages/gatsby-plugin-manifest/README.md
@@ -50,6 +50,7 @@ module.exports = {
         // see https://developers.google.com/web/fundamentals/web-app-manifest/#display
         display: `standalone`,
         icon: `src/images/icon.png`, // This path is relative to the root of the site.
+        include_favicon: true, // Include favicon
       },
     },
   ],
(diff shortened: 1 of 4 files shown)


packages/gatsby-plugin-manifest/README.md
```diff
@@ -50,6 +50,7 @@ module.exports = {
         // see https://developers.google.com/web/fundamentals/web-app-manifest/#display
         display: `standalone`,
         icon: `src/images/icon.png`, // This path is relative to the root of the site.
+        include_favicon: true, // Include favicon
```
sidharthachatterjee: I think option keys ought to be camelCased for consistency
```suggestion
        includeFavicon: true, // Include favicon
```

packages/gatsby-plugin-manifest/src/gatsby-ssr.js
```diff
@@ -10,8 +10,8 @@ exports.onRenderBody = ({ setHeadComponents }, pluginOptions) => {
   const legacy =
     typeof pluginOptions.legacy !== `undefined` ? pluginOptions.legacy : true
 
-  // If icons were generated, also add a favicon link.
-  if (pluginOptions.icon) {
+  // The user has an option to opt out of the favicon link tag being inserted into the head.
+  if (pluginOptions.icon && pluginOptions.include_favicon) {
```
sidharthachatterjee: We should _default_ to `true` so that this isn't a breaking change

packages/gatsby-plugin-manifest/README.md
gurpreet-hanjra: Hello @sidharthachatterjee Thank you. All other option keys are mentioned in snakeCase in documentation - https://www.gatsbyjs.org/packages/gatsby-plugin-manifest/?=manife#automatic-mode I would rather refactor all to camelCase or leave this one as is. Please suggest.

packages/gatsby-plugin-manifest/src/gatsby-ssr.js
gurpreet-hanjra: Done. Silly of me.

packages/gatsby-plugin-manifest/src/__tests__/gatsby-ssr.js
```diff
@@ -36,7 +35,25 @@ describe(`gatsby-plugin-manifest`, () => {
...
+    })
+    expect(headComponents).toMatchSnapshot()
+  })
+
+  it(`Adds link favicon tag if "include_favicon" is set to true`, () => {
+    onRenderBody(ssrArgs, { icon: true, include_favicon: true })
+    expect(headComponents).toMatchSnapshot()
+  })
+
+  it(`Adds link favicon if "include_favicon" option as is not provided `, () => {
```
wardpeet: Small nit on spelling:
```suggestion
  it(`Adds link favicon if "include_favicon" option is not provided`, () => {
```

packages/gatsby-plugin-manifest/src/gatsby-ssr.js
```diff
@@ -10,11 +10,17 @@ exports.onRenderBody = ({ setHeadComponents }, pluginOptions) => {
   const legacy =
     typeof pluginOptions.legacy !== `undefined` ? pluginOptions.legacy : true
 
-  // If icons were generated, also add a favicon link.
+  // The user has an option to opt out of the favicon link tag being inserted into the head.
   if (pluginOptions.icon) {
     let favicon = icons && icons.length ? icons[0].src : null
 
-    if (favicon) {
+    let insertFaviconLinkTag = Object.keys(pluginOptions).includes(
```
wardpeet: if you look at line 10 we check if a property exists a little bit easier. The way you do it looks a bit more work.
```suggestion
    const insertFaviconLinkTag = typeof pluginOptions.include_favicon !== `undefined` ? pluginOptions.include_favicon : true
```

packages/gatsby-plugin-manifest/README.md
wardpeet: I agree with @gurpreet-hanjra better to keep it consistent. Might be something we want to consider for v3

gurpreet-hanjra:
@wardpeet @sidharthachatterjee review comments are fixed. Please review.

PR #11418, opened 2019-01-30
Adding our agency to the creators list

PolGuixe:
## Description
Just added our agency to the creators' list ;)


PolGuixe:
Not sure why the lint and unit test windows fail.

PR #11420, opened 2019-01-30
ci(azure-pipelines): remove optional field that seems to cause problems

pieh:
just testing - seems to azure pipelines seems to fail recently because `parameters` is unrecognized :O

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -23,7 +23,6 @@ steps:
     condition: succeededOrFailed()
     inputs:
       testResultsFiles: "**junit*.xml"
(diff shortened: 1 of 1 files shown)


pieh:
As far as I can tell it's not used anywhere - here's last non-fail run https://dev.azure.com/gatsbyjs/gatsby/_build/results?buildId=5375&view=logs and I couldn't find reference to "TestRun" there. Docs I found say that this field is optional so ¯\_(ツ)_/¯

pieh:
And yeah, I'd say let's merge and see if this helps - we can revert if we will notice different issues with this

PR #11425, opened 2019-01-31
feat(gatsby-source-wordpress): add jwt_base_path option

rotexhawk:
## Description

Removes hardcoded JWT token path in gatsby-source-wordpress and passes it in auth options, to support other Wordpress JWT plugins. 

Test Passed and the functionality works locally.


wardpeet:
@rotexhawk Thanks for this PR. Could you perhaps create an issue too with some more information about which plugins and why this is necessary.

Code looks great!

rotexhawk:
I added this issue: https://github.com/gatsbyjs/gatsby/issues/11488

PR #11439, opened 2019-01-31
chore(docs): reword CSS in JS docs for clarity

gurpreet-hanjra:
## Description

Gives clear info about installation of libraries, helpers and gatsby plugins. Adds review comments from emotion docs.

### DIFF
--- a/docs/docs/glamor.md
+++ b/docs/docs/glamor.md
@@ -2,7 +2,7 @@
 title: Glamor
 ---
 
-In this guide, we'll walk through setting up a site with the CSS-in-JS library [Glamor](https://github.com/threepointone/glamor).
+In this guide, you will learn setting up a site with the CSS-in-JS library [Glamor](https://github.com/threepointone/glamor).
 
 Glamor is not actively maintained, the maintainer recommends using [Emotion](/docs/emotion).
 
@@ -19,7 +19,7 @@ First, open a new terminal window and run the following to create a new site:
 gatsby new glamor-tutorial https://github.com/gatsbyjs/gatsby-starter-hello-world
 ```
 
-Second, we’ll install the necessary dependencies for Glamor and Gatsby.
+Second, install the necessary dependencies for Glamor and Gatsby.
 
 ```shell
 npm install --save gatsby-plugin-glamor glamor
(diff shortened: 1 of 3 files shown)


docs/docs/glamor.md
```diff
@@ -19,7 +19,7 @@ First, open a new terminal window and run the following to create a new site:
 gatsby new glamor-tutorial https://github.com/gatsbyjs/gatsby-starter-hello-world
 ```
 
-Second, install the Gatsby plugin for Glamor.
+Second, we’ll install the necessary dependencies for Glamor and Gatsby.
```
marcysutton: Remove `we'll` here, since the docs are intended for the person reading. You can learn more about this in the Gatsby style guide, which recommends using the more personal "you" for both the docs and tutorial (or in this case, omitting it): https://www.gatsbyjs.org/docs/gatsby-style-guide/#use-you-as-the-pronoun

I can see there are other uses of "we'll" up above in this doc; if you feel like rewording them to fit in with the style guide that would be awesome...but it would be totally fine to limit your changes to this line only.

docs/docs/styled-components.md
```diff
@@ -20,7 +20,7 @@ First, open a new terminal window and run the following to create a new site:
 gatsby new styled-components-tutorial https://github.com/gatsbyjs/gatsby-starter-hello-world
 ```
 
-Next, we'll install the necessary dependencies for `styled-components`, including the Gatsby plugin.
+Second, we'll install the necessary dependencies for `styled-components`, including the Gatsby plugin.
```
marcysutton: Since this line is being updated, can you remove `we'll` here as well? Even though it was in there already :)

docs/docs/glamor.md
gurpreet-hanjra: Thank you so much @marcysutton for pointing me to the style guide. I have made the suggested changes!

docs/docs/styled-components.md
gurpreet-hanjra: Removed `we will` from all the styling docs. :)

gurpreet-hanjra:
Hello, @marcysutton please review :)

PR #11444, opened 2019-01-31
docs(tutorial): Updated div's style to be consistent with the previous one

tsvetlin:
## Description
In the tutorial, the Layout header style was changed, but it was not highlighted, as changed, so I fixed that.

### DIFF
--- a/docs/tutorial/part-three/index.md
+++ b/docs/tutorial/part-three/index.md
@@ -232,8 +232,8 @@ const ListLink = props => (
 // highlight-end
 
 export default ({ children }) => (
+  {/* highlight-start */}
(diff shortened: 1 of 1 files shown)


docs/tutorial/part-three/index.md
```diff
@@ -232,8 +232,8 @@ const ListLink = props => (
 // highlight-end
 
 export default ({ children }) => (
+  {/* highlight-start */}
   <div style={{ margin: `0 auto`, maxWidth: 650, padding: `1.25rem 1rem` }}>
```
pieh: Maybe we should update styling here to be same as in previous code snippet instead?

PR #11448, opened 2019-01-31
fix(gatsby): check type when querying by ID

pieh:
When querying/filter by id our special case doesn't check for type and can return nodes of wrong type ( https://github.com/gatsbyjs/gatsby/blob/2f79efeddd090c1d834625fedc011cf3b4d30b4b/packages/gatsby/src/redux/run-sift.js#L287-L295 )

This adds validation that return node is of correct type.

This fixes `Expected value of type "X" but got: [object Object].` errors when using queries like:
```
    wordpressWpTalk(id: { eq: $id }) {
      title
      content
    }
    wordpressPage(id: { eq: $id }) {
      title
      content
    }
```
where both fields would try to return same node even if it should only return node for one of them (one that matches the type)

### DIFF
--- a/packages/gatsby/src/redux/__tests__/run-sift.js
+++ b/packages/gatsby/src/redux/__tests__/run-sift.js
@@ -17,18 +17,30 @@ const mockNodes = [
   {
     id: `id_1`,
     string: `foo`,
+    internal: {
+      type: `notTest`,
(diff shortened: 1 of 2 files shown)


KyleAMathews:
Nice fix!

PR #11454, opened 2019-01-31
fix(core): empty cache if delete fails [docker]

### LINKED ISSUE #11097: .cache not deleted when mounted in docker
## Description

When running gatsby in a container with a mounted .cache folder, deletion of the .cache folder fails.

### Pre-requisites:
 - git
 - docker

### Steps to reproduce

1. Clone my [dockerized gatsby-default-starter](https://github.com/edgarnansen/gatsby-docker-mounted-cache)
2. Run `docker build -t gatsby-bug-mounted-cache-folder .`
This might take a few minutes the first time.
3. Run `docker run -v mounted-public-folder:/app/public -v mounted-cache-folder:/app/.cache --rm gatsby-bug-mounted-cache-folder`

### Expected result

Gatsby deletes the `.cache` folder.

### Actual result

The folder cannot be deleted.

### Environment
```
  System:
    OS: Windows 10
    CPU: (4) x64 Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz
  Binaries:
    Yarn: 1.12.3 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.5.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 42.17134.1.0
  npmPackages:
    gatsby: ^2.0.76 => 2.0.76
    gatsby-plugin-offline: ^2.0.16 => 2.0.20
    gatsby-plugin-sharp: ^2.0.14 => 2.0.16
    gatsby-source-filesystem: ^2.0.8 => 2.0.12
    gatsby-transformer-sharp: ^2.1.8 => 2.1.9
```
`Docker version 18.09.1, build 4c52b90`

### Possible solutions
The problematic code can be found [here](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/bootstrap/index.js#L200).
We could either replace `fs.remove` with `fs.emptyDir`, or try to `fs.emptyDir` if an `EBUSY` error is thrown.

edgarnansen:
## Description
Attempts to empty the `.cache/` directory if deleting it fails. Deletion might fail if for example the directory is a mountpoint for a docker volume.

Works locally™

## Related Issues
This fixes #11097

### DIFF
--- a/packages/gatsby/src/bootstrap/index.js
+++ b/packages/gatsby/src/bootstrap/index.js
@@ -175,10 +175,10 @@ module.exports = async (args: BootstrapArgs) => {
       data
     `)
   }
-
+  const cacheDirectory = `cacheDirectory`
   if (!oldPluginsHash || pluginsHash !== oldPluginsHash) {
     try {
-      await fs.remove(`${program.directory}/.cache`)
+      await fs.remove(cacheDirectory)
     } catch (e) {
       report.error(`Failed to remove .cache files.`, e)
     }
@@ -197,7 +197,7 @@ module.exports = async (args: BootstrapArgs) => {
 
   // Now that we know the .cache directory is safe, initialize the cache
   // directory.
-  await fs.ensureDir(`${program.directory}/.cache`)
+  await fs.ensureDir(cacheDirectory)
 
   // Ensure the public/static directory
   await fs.ensureDir(`${program.directory}/public/static`)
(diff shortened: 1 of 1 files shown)


packages/gatsby/src/bootstrap/index.js
```diff
@@ -175,10 +175,12 @@ module.exports = async (args: BootstrapArgs) => {
       data
     `)
   }
-
+  const cacheDirectory = `cacheDirectory`
```
pieh: I think this should be 
```suggestion
  const cacheDirectory = `${program.directory}/.cache`
```
?

edgarnansen: Uhhh... Yeah 😓

PR #11465, opened 2019-02-01
docs: Add codesandbox sortorder example to graphql-reference

stefanprobst:
Add codesandbox for specifying sort order per sort field.
Will the codesandbox auto-update when `examples/graphql-reference/package.json` is changed?


freiksenet:
@LekoArts how can one upgrade sandbox?

PR #11480, opened 2019-02-01
feat(gatsby): allow schema customization

freiksenet:
Rendered [release blog post](https://github.com/gatsbyjs/gatsby/blob/3f2c4a015b93bc9444c8ab95f6020e05145a5e16/docs/blog/2019-03-18-releasing-new-schema-customization/index.md)

Is in runnable state right now. Try it:

```
yarn add gatsby@schema-customization
```

Report all bugs in #12272.


freiksenet:
My next plans:

- Confirm www can build
- Add addTypeDefs and addResolvers apis
- Put this under a flag
- Port all tests
- Make an RFC and try to get feedback
- Get it merged

Longer term stuff / nice to haves

- API to add directives
- Better resolver writing experience
- Clean up of inference
- Schema directives to selectively enable/disable inference for types

packages/gatsby/src/bootstrap/index.js
```diff
@@ -385,7 +386,16 @@ module.exports = async (args: BootstrapArgs) => {
...
   const graphqlRunner = (query, context = {}) => {
     const schema = store.getState().schema
-    return graphql(schema, query, context, context, context)
+    return graphql(
+      schema,
+      query,
+      context,
+      {
+        ...context,
+        nodeModel: nodeStore,
```
stefanprobst: If we expose the store directly, should users register page dependencies with getNodeAndSavePathDependency?

packages/gatsby/src/db/nodes.js
```diff
@@ -1,8 +1,21 @@
...
+
+interface NodeStore {
+  getNodes: () => Array<any>;
+  getNode: (id: string) => any | undefined;
+  getNodesByType: (type: string) => Array<any>;
+  getTypes: () => Array<string>;
+  hasNodeChanged: (id: string, digest: string) => boolean;
+  getNodeAndSavePathDependency: (id: string, path: string) => any | undefined;
+  // XXX(freiksenet): types
+  runQuery: (...args: any) => any | undefined;
```
stefanprobst: We can use the current runQuery, but we should think about extracting some of the changes I did, e.g. resolving nodes for sort, group, and distinct fields; also not relevant right now but when we add addTypeDefs: handling NonNull fields

packages/gatsby/src/schema/infer/example-value.js
```diff
@@ -0,0 +1,165 @@
+const _ = require(`lodash`)
+const { reportConflict } = require(`./type-conflict-reporter`)
```
stefanprobst: The current TypeConflictReporter exports a class, and the method is `addConflict` not `reportConflict`

packages/gatsby/src/schema/infer/example-value.js
```diff
@@ -0,0 +1,165 @@
...
+          entriesByType.map(entry => entry.type),
+          arrayWrappers
+        )
+      ) {
+        // FIXME: is a mix of date *objects* and strings problematic?
+        // I.e. when date objects will be treated as strings,
+        // are they automatically stringified? Probably yes, because
+        // `Date` has `toString` method. In current master, we return
+        // InvalidValue.
+        value = `String`
```
stefanprobst: This is actually a change in behavior: we currently infer a mix of Date objects and strings as INVALID_VALUE

packages/gatsby/src/schema/infer/infer.js
```diff
@@ -0,0 +1,302 @@
...
+  // (ii) hinders reusing types.
+  if (linkedTypes.length > 1) {
+    const typeName = linkedTypes.sort().join(``) + `Union`
+    schemaComposer.getOrCreateUTC(typeName, utc => {
+      const types = linkedTypes.map(typeName =>
+        schemaComposer.getOrCreateTC(typeName)
+      )
+      utc.setTypes(types)
+      types.forEach(type => {
+        utc.addTypeResolver(
```
stefanprobst: In graphql-compose 5.8 we now have `utc.setResolveType(node => node.internal.type)`

packages/gatsby/src/schema/infer/infer.js
```diff
@@ -0,0 +1,302 @@
...
+      return `Boolean`
+    case `number`:
+      return is32BitInteger(value) ? `Int` : `Float`
+    case `string`:
+      if (isDate(value)) {
+        return dateResolver
+      }
+      // FIXME: The weird thing is that we are trying to infer a File,
+      // but cannot assume that a source plugin for File nodes is actually present.
+      if (schemaComposer.has(`File`) && isFile(nodeStore, selector, value)) {
```
stefanprobst: We can remove the `schemaComposer.has('File')` check here now (see above)

packages/gatsby/src/schema/resolvers.js
```diff
@@ -0,0 +1,244 @@
...
+  const operator = Array.isArray(fieldValue) ? oneOf : equals
+  args.filter = by.split(`.`).reduceRight((acc, key, i, { length }) => {
+    return {
+      [key]: i === length - 1 ? operator(acc) : acc,
+    }
+  }, fieldValue)
+
+  const returnType = getNullableType(info.returnType)
+  const type = getNamedType(returnType)
+  const possibleTypes = isAbstractType(type)
```
stefanprobst: I think we should move all of this out of `link` into `find` (see my -v2 branch)

packages/gatsby/src/schema/resolvers.js
```diff
@@ -0,0 +1,244 @@
...
+// FIXME: Handle array of arrays
+// Maybe TODO: should we check fieldValue *and* info.returnType?
+const link = ({ by, from }) => async (source, args, context, info) => {
+  const fieldValue = source && source[from || info.fieldName]
+
+  if (fieldValue == null || _.isObject(fieldValue)) return fieldValue
+  if (
+    Array.isArray(fieldValue) &&
+    // TODO: Do we have to look with fieldValue.some(v => isObject(v))?
+    (fieldValue[0] == null || _.isObject(fieldValue[0]))
```
stefanprobst: Here (and probably elsewhere): my `isObject` translates to `_.isPlainObject`

packages/gatsby/src/schema/resolvers.js
```diff
@@ -0,0 +1,244 @@
...
+  const returnType = getNullableType(info.returnType)
+  const type = getNamedType(returnType)
+  const possibleTypes = isAbstractType(type)
+    ? info.schema.getPossibleTypes(type)
+    : [type]
+
+  if (returnType instanceof GraphQLList) {
+    const results = await Promise.all(
+      possibleTypes.map(type =>
+        findMany(type.name)({
```
stefanprobst: Now that `link` is in the same file and does not use the `findMany` wrapped with `withPageDependencies` we need to do this ourselves

packages/gatsby/src/schema/resolvers.js
```diff
@@ -0,0 +1,244 @@
...
+          context,
+          info,
+        })
+      )
+    )
+    return results.reduce((acc, r) => acc.concat(r))
+  } else {
+    let result
+    for (const type of possibleTypes) {
+      result = await findOne(type.name)({
```
stefanprobst: withPageDependencies

packages/gatsby/src/schema/schema.js
```diff
@@ -0,0 +1,317 @@
...
+            nodes: nodeStore.getNodesByType(typeName),
+          },
+          traceId: `initial-setFieldsOnGraphQLNodeType`,
+          parentSpan: parentSpan,
+        })
+        if (fields) {
+          // NOTE: `setFieldsOnGraphQLNodeType` only allows setting
+          // nested fields with a path as property name, i.e.
+          // `{ 'frontmatter.published': 'Boolean' }`, but not in the form
+          // `{ frontmatter: { published: 'Boolean' }}`
```
stefanprobst: Should be documented as new behavior

packages/gatsby/src/schema/schema.js
```diff
@@ -0,0 +1,317 @@
...
+//           tc.addFields({ [fieldName]: fieldConfig })
+//         }
+//       })
+//     }
+//   })
+// }
+
+const addResolvers = ({ schemaComposer, typeComposer }) => {
+  const typeName = typeComposer.getTypeName()
+  const inputTypeComposer = toInputObjectType(
```
stefanprobst: Note that currently we exclude `parent` and `children` fields from the InputObjectType

packages/gatsby/src/schema/schema.js
```diff
@@ -0,0 +1,317 @@
...
+  typeComposer.addResolver({
+    name: `findOne`,
+    type: typeComposer,
+    args: {
+      ...FilterInputTC.getFields(),
+    },
+    resolve: findOne(typeName),
+  })
+  typeComposer.addResolver({
+    name: `findMany`,
```
stefanprobst: I don't think we currently use the non-paginated version

packages/gatsby/src/schema/types/NodeInterface.js
```diff
@@ -0,0 +1,76 @@
+const getOrCreateNodeInterface = schemaComposer => {
+  // TODO: why is `mediaType` on Internal? Applies only to File!?
+  // Is `fieldOwners` actually set anywhere? And is it supposed to be an array,
+  // as the Joi schema claims, or an object, which is the default in `redux/actions.js`
+  const InternalTC = schemaComposer.getOrCreateTC(`Internal`, tc => {
+    tc.addFields({
+      content: `String`,
+      contentDigest: `String`,
+      description: `String`,
+      fieldOwners: [`String`],
```
stefanprobst: It is actually an object

packages/gatsby/src/schema/types/NodeInterface.js
```diff
@@ -0,0 +1,76 @@
...
+  InternalTC.getITC()
+
+  const NodeInterfaceTC = schemaComposer.getOrCreateIFTC(`Node`, tc => {
+    tc.setDescription(`Node Interface`)
+    tc.addFields({
+      id: `String!`,
+      parent: {
+        type: `Node`,
+        resolve: async (source, args, context, info) =>
+          context.nodeModel.getNode(source.parent),
```
stefanprobst: withPageDependencies

packages/gatsby/src/schema/types/NodeInterface.js
```diff
@@ -0,0 +1,76 @@
...
+      },
+      children: {
+        type: `[Node]!`,
+        resolve: async (source, args, context, info) =>
+          context.nodeModel.getNodes(source.children),
```
stefanprobst: withPageDependencies

packages/gatsby/src/schema/types/NodeInterface.js
```diff
@@ -0,0 +1,76 @@
...
+      contentDigest: `String`,
+      description: `String`,
+      fieldOwners: [`String`],
+      ignoreType: `Boolean`,
+      mediaType: `String`,
+      owner: `String`,
+      type: `String!`,
+    })
+  })
+  InternalTC.getITC()
```
stefanprobst: Probably not needed now that we use toInputObjectType directly (same with `NodeInterface.getITC()`) - We called this here to register a non-prefixed name (`Nodenput` and not `MarkdownParentNodeInput` - but we should try to remove prefixing upstream)

packages/gatsby/src/schema/types/NodeInterface.js
```diff
@@ -0,0 +1,76 @@
...
+  })
+  NodeInterfaceTC.getITC()
+
+  return NodeInterfaceTC
+}
+
+const addNodeInterface = ({ schemaComposer, typeComposer }) => {
+  const NodeInterfaceTC = getOrCreateNodeInterface(schemaComposer)
+  typeComposer.addInterface(NodeInterfaceTC)
+  NodeInterfaceTC.addTypeResolver(
```
stefanprobst: With graphql-compose 5.8 we now have `setResolveType(node => node.internal.type)`

packages/gatsby/src/schema/types/NodeInterface.js
```diff
@@ -0,0 +1,76 @@
...
+    typeComposer,
+    node => node.internal.type === typeComposer.getTypeName()
+  )
+  addNodeInterfaceFields({ schemaComposer, typeComposer })
+}
+
+const addNodeInterfaceFields = ({ schemaComposer, typeComposer }) => {
+  const NodeInterfaceTC = getOrCreateNodeInterface(schemaComposer)
+  typeComposer.addFields(NodeInterfaceTC.getFields())
+  NodeInterfaceTC.addTypeResolver(
```
stefanprobst: Like above

stefanprobst: Although we maybe also want to add `typeComposer.isTypeOf` to the types that implement the Node interface?? Might be beneficial in the future when we allow custom interfaces so the correct types are resolved automatically?

packages/gatsby/src/schema/schema.js
```diff
@@ -0,0 +1,317 @@
...
+    const fields = QueryTC.getFields()
+    // TODO: Wrap field resolvers to include projected fields in the
+    // selection set.
+    schemaComposer.Query.addFields(fields)
+
+    // Explicitly add the third-party schema's types, so they can be targeted
+    // in `addResolvers` API.
+    const rootTypeName = Object.values(fields)[0].type.name
+    const types = schema.getTypeMap()
+    Object.entries(types).forEach(
```
stefanprobst: Node 6 support (Object.entries)

packages/gatsby/src/schema/schema.js
```diff
@@ -0,0 +1,317 @@
...
+  const nodes = nodeStore.getNodesByType(typeName)
+
+  const hasChildrenByType = nodes.reduce((acc, node) => {
+    const children = node.children.map(nodeStore.getNode)
+    const childrenCountByType = children.reduce((acc, child) => {
+      const { type } = child.internal
+      acc[type] = acc[type] + 1 || 1
+      return acc
+    }, {})
+    Object.entries(childrenCountByType).forEach(([type, count]) => {
```
stefanprobst: Node 6 support (Object.entries)

packages/gatsby/src/schema/infer/is-file.js
```diff
@@ -0,0 +1,85 @@
...
+    .some(node => node.absolutePath === filePath)
+  return filePathExists
+}
+
+module.exports = {
+  isFile,
+}
+
+const getFirstValueAt = (node, selector) => {
+  let value = _.get(node, selector)
```
stefanprobst: I did not use `_.get` but `getValueAtSelector` because lodash does not work with arrays of objects - we can make certain assumptions in a graphql context that a more general solution like lodash cannot: that arrays are lists with uniform types.

packages/gatsby/src/bootstrap/index.js
stefanprobst: We should also think about if we want to provide something for getting nodes of Interface and Union types, i.e. those that don't map 1:1 to node types in the store?

packages/gatsby/src/db/nodes.js
stefanprobst: Also: when manually calling resolvers in [`run-sift`](https://github.com/gatsbyjs/gatsby/blob/9168d1f912017d3078decd7ea72f7a123ccbdbe3/packages/gatsby/src/redux/run-sift.js#L104), we should put the nodeModel on context as well

stefanprobst: And: if we use something like `getValueAtSelector` instead of `_.get` for [sorting in `run-sift`](https://github.com/gatsbyjs/gatsby/blob/9168d1f912017d3078decd7ea72f7a123ccbdbe3/packages/gatsby/src/redux/run-sift.js#L256), we can sort on properties on objects in arrays - this would be either a bugfix or breaking change

packages/gatsby/src/schema/infer/infer.js
```diff
@@ -0,0 +1,302 @@
...
+        const resolver = fieldConfig.resolve || defaultFieldResolver
+        fieldConfig.resolve = (source, args, context, info) =>
+          resolver(source, args, context, {
+            ...info,
+            fieldName: unsanitizedKey,
+          })
+      }
+    }
+
+    if (typeComposer.hasField(key)) {
```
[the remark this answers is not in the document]
freiksenet: I want to add resolvers to fields even if they are defined in typeDefs, if they don't have a resolver.

packages/gatsby/src/schema/infer/infer.js
freiksenet: Yeah, same for interfaces. I've actually asked maintainer to add that :D I'll fix that.

packages/gatsby/src/schema/resolvers.js
freiksenet: This code in general causes lots of problems, I think I want to clean it up. I'll look at it tomorrow. I'll check out v2 stuff too.

packages/gatsby/src/schema/types/NodeInterface.js
freiksenet: I'd rather use typeResolver than isTypeOf - isTypeOf is called for every possible type of the interface whenever an object is checked, thus it will be a O(N) checks for N types.

freiksenet:
@stefanprobst Thanks a lot for review!

packages/gatsby/src/schema/infer/infer.js
```diff
@@ -0,0 +1,302 @@
...
+  // create a GraphQLUnionType. Note that we don't support the case where
+  // scalar fields link to different types. Similarly, an array of objects
+  // with foreign-key fields will produce union types if those foreign-key
+  // fields are arrays, but not if they are scalars. See the tests for an example.
+  // FIXME: The naming of union types is a breaking change. In current master,
+  // the type name includes the key, which is (i) potentially not unique, and
+  // (ii) hinders reusing types.
+  if (linkedTypes.length > 1) {
+    const typeName = linkedTypes.sort().join(``) + `Union`
+    schemaComposer.getOrCreateUTC(typeName, utc => {
```
stefanprobst: `type = sc.getOrCreatUTC`

examples/using-type-definitions/gatsby-node.js
```diff
@@ -0,0 +1,82 @@
...
+    GraphCMS_Asset: {
+      // Instead of a resolve function it is possible to provide a
+      // field config for a *new* field.
+      // This allows specifying a return type. The values of sibling fields
+      // are available from `source` if they are in the selection set.
+      // Be aware that without other fields in the selection set this will
+      // throw if the parent is NonNull, even if this resolver returns a value.
+      // This is because the delegated schema does not know about this field.
+      imageFile: {
+        type: `File`,
```
KyleAMathews: Nifty! So this will "just work" with the existing data layer e.g. you'll be able to transform images now w/ gatsby-transformer-sharp?

examples/using-type-definitions/gatsby-node.js
```diff
@@ -0,0 +1,82 @@
+exports.addTypeDefs = ({ addTypeDefs }) => {
```
KyleAMathews: BTW, we use `create` rather than `add` as our operator for making a new thing https://www.gatsbyjs.org/docs/api-specification/#operators

packages/gatsby/src/schema/infer/is-file.js
```diff
@@ -0,0 +1,85 @@
...
+
+const getFilePath = (nodeStore, field, relativePath) => {
+  const [typeName, ...selector] = field.split(`.`)
+
+  if (typeName === `File`) return null
+
+  const looksLikeFile =
+    !path.isAbsolute(relativePath) &&
+    mime.getType(relativePath) !== null &&
+    // FIXME: Do we need all of this?
```
KyleAMathews: URLs are prone to looking like files so we just kept adding more and more tests

examples/using-type-definitions/gatsby-node.js
stefanprobst: @KyleAMathews Yes you'll get a `childImageSharp` field for free, so it will just work. Note however, that the details may change a bit, what's reflected in the `using-type-defs` example is still my prototype. E.g. @freiksenet has suggested using `fragments` not `projection`.

packages/gatsby/src/schema/schema.js
```diff
@@ -72,12 +73,17 @@ const updateSchemaComposer = async ({
   schemaComposer.addDirective(InferDirective)
   schemaComposer.addDirective(DontInferDirective)
   await addTypeDefs({ schemaComposer, parentSpan, typeDefs })
-  await addInferredTypes({ schemaComposer, nodeStore, parentSpan })
-  await addSetFieldsOnGraphQLNodeTypeFields({
+  nodeStore.getTypes().forEach(typeName => {
```
stefanprobst: This is so people don't have to use `implements Node` in addTypeDefs, right?

packages/gatsby/src/schema/schema.js
```diff
@@ -176,28 +182,35 @@ const addThirdPartySchemas = ({
...
+  const intermediateSchema = schemaComposer.buildSchema()
+  const addResolvers = resolvers => {
+    Object.keys(resolvers).forEach(typeName => {
+      const fields = resolvers[typeName]
+      if (schemaComposer.has(typeName)) {
+        const tc = schemaComposer.getTC(typeName)
+        Object.keys(fields).forEach(fieldName => {
+          const fieldConfig = fields[fieldName]
+          if (tc.hasField(fieldName)) {
+            tc.extendField(fieldName, fieldConfig)
```
stefanprobst: When extending a field, how do we expect people to get to the original field resolver (e.g. when they just want to wrap it). Should they get it from the intermediateSchema?

stefanprobst: One other thing to think about is if we allow changing field types, and adding new fields, they are not automatically reflected in the InputObjectType

packages/gatsby/src/schema/infer/infer.js
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    while (arrays--) {
+      fieldConfig = { ...fieldConfig, type: [fieldConfig.type] }
+    }
+
+    // Proxy resolver to unsanitized fieldName in case it contained invalid characters
+    if (key !== unsanitizedKey) {
+      // Don't create a field with the sanitized key if a field with that name already exists
+      if (!(exampleObject[key] == null && !typeComposer.hasField(key))) {
```
stefanprobst: Should we keep the `invariant` here? I.e. for the case when the sanitized field name already exists on the type.

freiksenet:
Okey, current status and next steps.

1. Add type defs and addResolvers works, however default resolvers aren't applied to new fields in type defs. Figuring out if it should be automatic, explicit or semi automatic.

Eg give inferred types

```gql
# inferred type def
type Post implements Node {
  title: String
  time: Date
}

# add type defs 
type Post implements Node {
  title: String
  time: Date
  newTime: Date
}
```

What should be the logic of applying args and resolvers to defined and inferred fields? I think the following logic can apply:

* If field is inferred and not defined, create our default args/resolvers
* If field is inferred and defined, add our default args/resolvers if field doesn't have args or @dontInfer directive.
* If field is not inferred and defined, don't add default resolvers. Maybe provide tools to add default format args or smth.

2. Readd all the tests

3. Adjust starter to match up-to-date code

4. Put stuff under flag (or should we not?)

5. Make an RFC

packages/gatsby/src/schema/infer/example-value.js
freiksenet: I don't see a point of treating mix of dates/strings differently than other conflicting values and we don't have concept of INVALID_VALUE anymore, so I'll move it to normal conflicts.

freiksenet: @pieh could you comment if this would be a correct behaviour?

packages/gatsby/src/schema/infer/index.js
```diff
@@ -0,0 +1,57 @@
...
+  })
+
+  addInferredFields({ schemaComposer, typeComposer, nodeStore, exampleValue })
+  return typeComposer
+}
+
+const addInferredTypes = ({ schemaComposer, nodeStore, parentSpan }) => {
+  const typeNames = nodeStore.getTypes()
+
+  // Infer File first so all the links to it would work
```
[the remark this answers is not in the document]
freiksenet: Duly noted, let's do it in second stage when we move all the plugins to use this.

packages/gatsby/src/schema/infer/infer.js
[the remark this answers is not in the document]
freiksenet: This actually doesn't work as it should, I'll fix it.

packages/gatsby/src/schema/resolvers.js
freiksenet: @stefanprobst while you are doing resolver cleanup, maybe link can also be cleaned up.

packages/gatsby/src/schema/schema.js
```diff
@@ -0,0 +1,317 @@
+const _ = require(`lodash`)
+const { toInputObjectType } = require(`graphql-compose`)
+const apiRunner = require(`../utils/api-runner-node`)
+const GraphQLDate = require(`./types/Date`)
```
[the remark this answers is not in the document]
freiksenet: Duly noted, let's change it at later stage.

packages/gatsby/src/schema/schema.js
freiksenet: Is this additive or breaking?

packages/gatsby/src/schema/schema.js
freiksenet: You mean in master?

packages/gatsby/src/schema/types/NodeInterface.js
freiksenet: What do you mean by "remove prefixing upstream"?

packages/gatsby/src/schema/schema.js
```diff
@@ -259,50 +259,60 @@ const addConvenienceChildrenFields = ({
...
-  }, {})
+  const childNodesByType = groupChildNodesByType({ nodeStore, nodes })
 
-  Object.keys(hasChildrenByType).forEach(typeName => {
-    const hasChildren = hasChildrenByType[typeName]
-    const fieldName = (hasChildren ? `children` : `child`) + typeName
-    let resolver
-    if (hasChildren) {
-      resolver = (source, args, context, info) => {
-        const result = context.nodeModel.getNodes(source.children)
```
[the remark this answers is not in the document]
freiksenet: TBH, something was broken but I couldn't figure out what exactly :(. Reverting to original impl made it work.

packages/gatsby/src/schema/schema.js
freiksenet: It's so that types exist when `setFieldsOnGraphQLNodeTypeFields` is called.

freiksenet: `implements Node` is mandatory.

packages/gatsby/src/schema/schema.js
freiksenet: We probably should only allow changing resolver in case field is defined, that is a good point.

packages/gatsby/src/schema/infer/infer.js
freiksenet: I've removed invariant because I think some behavior in www expected it to just be ignored, but I'm not 100% sure now.

packages/gatsby/src/schema/schema.js
stefanprobst: Should be additive

packages/gatsby/src/schema/schema.js
stefanprobst: Yes in master. But we now can handle the circular structure of NodeInput

packages/gatsby/src/schema/types/NodeInterface.js
stefanprobst: Just not adding the type name [here](https://github.com/graphql-compose/graphql-compose/blob/364dcdd2c483ba81d99dfd6c784823f6555eaea4/src/utils/toInputObjectType.js#L105)

packages/gatsby/src/schema/infer/infer.js
stefanprobst: Might have been #11578

freiksenet: I'll try.

packages/gatsby/src/schema/types/NodeInterface.js
freiksenet: One thing I was thinking about is introducing concept of a "TypeSetComposer" which is an abstraction on top of TypeComporers, InputTypeComposers etc and allows defining dependencies and transformations. Eg we can then write own getITC that knows how to not prefix. This would also solve the issue with clearing the ITC cache.

stefanprobst: I'll have to think a bit about this, but not sure if we need to do much more than we already have: can we just call `tc.removeInputTypeComposer(); tc.getITC()` [here](https://github.com/gatsbyjs/gatsby/blob/9a59b474bb2e663f4a16f1ca53f469c6cf41c32b/packages/gatsby/src/schema/schema.js#L246)? What other issues would TypeSetComposer address?

freiksenet: I thought it might also have all the other type composers linked, like filter, sort etc. So we could clear all the generated types in one go when original type changes. Could be useful for things like live schema reload.

packages/gatsby/src/schema/infer/example-value.js
freiksenet: OK, we could just remove it and it passes tests, so I assume this behaviour works correctly.

packages/gatsby/src/schema/infer/example-value.js
```diff
@@ -0,0 +1,143 @@
...
+  const exampleValue = getExampleObject({
+    nodes,
+    prefix: typeName,
+    typeConflictReporter,
+    ignoreFields,
+  })
+  return exampleValue
+}
+
+const INVALID_VALUE = Symbol(`INVALID_VALUE`)
```
stefanprobst: Are we using this anywhere?

freiksenet: No, thanks.

packages/gatsby/src/db/nodes.js
freiksenet: We should extract the changes that resolve fields so that we can have unified logic for it. Ideally stuff in what we call "nodeStorage" shouldn't care about GraphQL Types. But let's do it after releasing this.

stefanprobst: Sounds good. I'll remove the resolving for sort fields that I apparently sneaked in (oops).

packages/gatsby/src/schema/types/node-interface.js
```diff
@@ -0,0 +1,67 @@
...
+      owner: `String`,
+      type: `String!`,
+    })
+  })
+  InternalTC.getITC()
+
+  const NodeInterfaceTC = schemaComposer.getOrCreateIFTC(`Node`, tc => {
+    tc.setDescription(`Node Interface`)
+    tc.addFields({
+      id: `String!`,
```
stefanprobst: Why make the `id` field `String`, not `ID`? [It is ID now as well.](    debugger
)

packages/gatsby/src/schema/infer/add-inferred-fields.js
```diff
@@ -0,0 +1,321 @@
...
+        if (lists === arrays && fieldType instanceof GraphQLObjectType) {
+          addInferredFieldsImpl({
+            schemaComposer,
+            typeComposer: typeComposer.getFieldTC(key),
+            exampleObject: value,
+            typeMapping,
+            nodeStore,
+            prefix: selector,
+            depth: depth + 1,
+            addDefaultResolvers: true,
```
stefanprobst: Why is addDefaultResolvers true here?

packages/gatsby/src/schema/types/node-interface.js
freiksenet: Link doesn't work. I had www fail when it was ID, there were some String variables that were passing in.

packages/gatsby/src/schema/infer/add-inferred-fields.js
freiksenet: Fixed to inherit it from parent.

stefanprobst: We probably also want `addDefaultResolvers` in `getFieldConfig`.

packages/gatsby/src/schema/types/node-interface.js
stefanprobst: Hmm that's weird

packages/gatsby/src/schema/node-model.js
```diff
@@ -0,0 +1,178 @@
...
+      this.createPageDependency
+    )
+  }
+
+  getTypes() {
+    return this.nodeStore.getTypes()
+  }
+
+  findRootNodeAncestor(...args) {
+    return this.nodeStore.findRootNodeAncestor(...args)
```
stefanprobst: Is `findRootNodeAncestor` on the nodeStore now?

packages/gatsby/src/schema/schema.js
```diff
@@ -175,7 +175,7 @@ const addThirdPartySchemas = ({
     const rootTypeName = fields[Object.keys(fields)[0]].type.name
     const types = schema.getTypeMap()
     Object.keys(types).forEach(typeName => {
-      if (typeName.startsWith(rootTypeName)) {
+      if (!typeName === rootTypeName) {
```
stefanprobst: Why change this? Third party schemas have all their types prefixed with the name of the type of the only rootquery field. We want to grab those, but not others like the builtin scalars etc.

freiksenet: Good point about built-in scalars.

Third party schemas aren't prefixed, the plugin prefixes them. So we shouldn't bind plugin implementation to the core schema merging.

packages/gatsby/src/schema/node-model.js
freiksenet: Good find.

packages/gatsby/src/schema/schema.js
```diff
@@ -0,0 +1,350 @@
...
+              if (fieldConfig.args) {
+                newConfig.args = fieldConfig.args
+              }
+              if (fieldConfig.resolve) {
+                newConfig.resolve = fieldConfig.resolve
+              }
+              tc.extendField(fieldName, newConfig)
+            } else if (fieldTypeName) {
+              report.warn(
+                `\`addResolvers\` passed resolvers for field \`${typeName}.${fieldName}\` with type ${fieldTypeName}. Such field with type ${originalTypeName} already exists on the type. Use \`addTypeDefs\` to override type fields.`
```
stefanprobst: Is it `createTypes` now?

packages/gatsby/src/schema/schema.js
```diff
@@ -0,0 +1,350 @@
...
+              )
+            }
+          } else {
+            tc.addFields({ [fieldName]: fieldConfig })
+          }
+        })
+      } else {
+        report.warn(
+          `\`addResolvers\` passed resolvers for type \`${typeName}\` that doesn't exist in the schema. Use \`addTypeDefs\` to add the type before adding resolvers.`
```
stefanprobst: Is it `createTypes` now?

packages/gatsby/graphql.js
```diff
@@ -1,6 +1,6 @@
 "use strict"
 
 const graphql = require(`graphql`)
-const GraphQLJSON = require(`graphql-type-json`)
+const { GraphQLJSON } = require(`graphql-compose`)
 
 module.exports = Object.assign({}, graphql, { GraphQLJSON })
```
stefanprobst: Should we also export `GraphQLDate` here?

packages/gatsby/src/schema/infer/add-inferred-fields.js
```diff
@@ -0,0 +1,354 @@
...
+        schemaComposer.getOrCreateTC(typeName)
+      )
+      utc.setTypes(types)
+      utc.setResolveType(node => node.internal.type)
+    })
+  } else {
+    type = linkedTypes[0]
+  }
+
+  return { type, resolve: link({ by: foreignKey || `id`, from: key }) }
```
stefanprobst: `from: key` is not necessary because we have the proxy resolvers

examples/using-type-definitions/gatsby-node.js
```diff
@@ -0,0 +1,93 @@
...
+            (post.authors || []).some(author => author === fieldValue)
+          )
+        },
+      },
+    },
+    BlogJson: {
+      // Add a resolver to a field defined with `createTypes`.
+      authors: {
+        resolve(source, args, context, info) {
+          const emails = source[info.fieldName]
```
stefanprobst: Not sure how best to approach the issue where `source` changes when the field is in both the input filter and selection set. I had added a helper for this, but removed it again, because it seemed to make the example less straightforward. I feel a resolvers-centered approach, and one with mongo-like query syntax, are sometimes a bit orthogonal. Nevertheless, we need to show some kind of best practice solution here.

packages/gatsby/src/schema/schema.js
```diff
@@ -0,0 +1,362 @@
...
+      addedTypes.forEach(type => {
+        const { GraphQLInterfaceType, GraphQLUnionType } = require(`graphql`)
+        let typeComposer
+        if (type instanceof GraphQLInterfaceType) {
+          typeComposer = schemaComposer.getOrCreateIFTC(type.name)
+        } else if (type instanceof GraphQLUnionType) {
+          typeComposer = schemaComposer.getOrCreateUTC(type.name)
+        }
+        if (typeComposer) {
+          typeComposer.setResolveType(node => node.internal.type)
```
stefanprobst: This might not be the best place to do this -- I feel we now have multiple places where we do some kind of initial processing of types? But we can take a second look at this later.
I added `addSchemaMustHaveType` because in the tests the custom abstract types from SDL would be stripped from the built schema otherwise.
And `setResolveType` is similarly needed -- but this means that we only allow custom abstract types that combine types that implement Node. Is that restriction ok? Otherwise we'd have to provide some other way to add a type resolver, but I'm not sureI'd want that.

packages/gatsby/src/schema/node-model.js
```diff
@@ -0,0 +1,207 @@
...
+      result,
+      pageDependencies,
+      this.createPageDependency
+    )
+  }
+
+  async runQuery(args, pageDependencies) {
+    const { query, firstOnly, type } = args || {}
+
+    // We don't support querying union types (yet?), because the combined types
```
stefanprobst: We should rethink not supporting union types in `runQuery` later, as part of refactoring resolving nodes and run-sift (e.g. we cannot call `getFields()` on a union type there). Note that `getAllNodes` etc. _do_ support both unions and interfaces.

packages/gatsby/src/schema/__tests__/node-model.js
```diff
@@ -0,0 +1,446 @@
...
+      const firstOnly = false
+      const result = await nodeModel.runQuery({ query, firstOnly, type })
+      expect(result.length).toBe(2)
+      expect(result[0].id).toBe(`file1`)
+      expect(result[1].id).toBe(`file3`)
+    })
+  })
+
+  describe(`findRootNodeAncestor`, () => {
+    it(`returns an object's top most ancestor node`, () => {
```
stefanprobst: I'm not 100% clear why finding the top most ancestor is default, as opposed to finding an object's parent node (which is what you'd need most of the time)?

packages/gatsby/src/schema/__tests__/node-model.js
```diff
@@ -0,0 +1,446 @@
...
+        path: `/`,
+        nodeId: `post1`,
+      })
+      expect(createPageDependency).toHaveBeenCalledWith({
+        path: `/`,
+        nodeId: `post3`,
+      })
+    })
+
+    it(`creates page dependencies with connection type`, async () => {
```
stefanprobst: Also not 100% clear when we need to add a `connection` type page dependency.

packages/gatsby/src/schema/schema.js
```diff
@@ -0,0 +1,362 @@
...
+        const tc = schemaComposer.getTC(typeName)
+        Object.keys(fields).forEach(fieldName => {
+          const fieldConfig = fields[fieldName]
+          if (tc.hasField(fieldName)) {
+            const originalTypeName = tc.getFieldType(fieldName).toString()
+            const fieldTypeName =
+              fieldConfig.type && fieldConfig.type.toString()
+            if (
+              !fieldTypeName ||
+              tc.getFieldType(fieldName) === fieldConfig.type.toString()
```
stefanprobst: I think for third-party types, we *should* allow changing the type, because those cannot use `createTypes` action. Could be as simple as setting a `type.isThirdPartyType` flag when adding them in `addThirdPartySchemas`

packages/gatsby/src/schema/schema.js
```diff
@@ -0,0 +1,362 @@
...
+            ) {
+              const newConfig = {}
+              if (fieldConfig.args) {
+                newConfig.args = fieldConfig.args
+              }
+              if (fieldConfig.resolve) {
+                newConfig.resolve = fieldConfig.resolve
```
stefanprobst: I think I had left a comment about this already, but can't find it atm -- It would be good to make the original resolver available on `info` or smth for easy wrapping

packages/gatsby/src/schema/schema.js
stefanprobst: Yep that works

packages/gatsby/src/schema/schema.js
```diff
@@ -0,0 +1,362 @@
...
+  parentSpan,
+}) =>
+  Promise.all(
+    Array.from(schemaComposer.values()).map(async tc => {
+      if (
+        tc instanceof schemaComposer.TypeComposer &&
+        tc.hasInterface(`Node`)
+      ) {
+        const typeName = tc.getTypeName()
+        const [fields] = await apiRunner(`setFieldsOnGraphQLNodeType`, {
```
stefanprobst: My mistake - I think the array returned by `setFields...` can have more than one entry. I'l doublecheck and fix this.

pieh: `apiRunner` will return array because multiple plugins can implement same API - I.e. both plugin A and plugin B extends `MarkdownRemark` type, but they will be separate items in returned array.

packages/gatsby/src/schema/infer/get-infer-config.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  return inferConfig
+}
+
+module.exports = getInferConfig
+
+const getNoDefaultResolvers = directive => {
+  const noDefaultResolvers = directive.arguments.find(
+    ({ name }) => name.value === `noDefaultResolvers`
```
stefanprobst: Do we want this to be `addDefaultResolvers` or `noDefaultResolvers`?

packages/gatsby/src/schema/infer/get-infer-config.js
```diff
@@ -0,0 +1,49 @@
...
+  )
+  if (noDefaultResolvers) {
+    if (noDefaultResolvers.value.kind === Kind.BOOLEAN_NODE) {
```
stefanprobst: I think it's [`Kind.BOOLEAN`](https://github.com/graphql/graphql-js/blob/4116e2fc4fe36688f683258388f4a2d52076d199/src/language/kinds.js#L35)?

examples/using-type-definitions/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+## 🚀 Quick start
+
+1.  **Create a Gatsby site.**
+
+    Use the Gatsby CLI to create a new site, specifying the default starter.
+
+    ```sh
+    # create a new Gatsby site using the default starter
+    npx gatsby new my-default-starter https://github.com/gatsbyjs/gatsby-starter-default
```
muescha:
```suggestion
    gatsby new my-default-starter https://github.com/gatsbyjs/gatsby-starter-default
```

examples/using-type-definitions/src/components/layout.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { StaticQuery, graphql } from 'gatsby'
+
+import Header from './header'
+import './layout.css'
+
+const Layout = ({ children }) => (
+  <StaticQuery
```
muescha: better use here the new `useStaticQuery`? -> https://www.gatsbyjs.org/docs/use-static-query/

examples/using-type-definitions/src/components/seo.js
```diff
@@ -0,0 +1,96 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import Helmet from 'react-helmet'
+import { StaticQuery, graphql } from 'gatsby'
```
muescha: better use here the new `useStaticQuery`? -> https://www.gatsbyjs.org/docs/use-static-query/

examples/using-type-definitions/README.md
stefanprobst: Thanks!

examples/using-type-definitions/src/components/layout.js
stefanprobst: The example was bootstrapped from a default starter, so we're just using what is the default there.

examples/using-type-definitions/src/components/seo.js
stefanprobst: see above

docs/blog/2019-03-18-releasing-new-schema-customization/index.md
```diff
@@ -0,0 +1,131 @@
...
+        return authors.map(author => author.name)
+      },
+    },
+  },
+})
+```
+
+## The Type Builder API
+
+While `createTypes` accepts `graphql-js` types along with a [Schema Definition Language (SDL)](https://www.prisma.io/blog/graphql-sdl-schema-definition-language-6755bcb9ce51) string, we've also added an option to use `graphql-js` types so that users could create types with resolvers. However, `graphql-js` is somewhat verbose and it can be hard to refer to types that don't yet exist or don't exist in a current scope. Therefore, we decided to add another programmatic API that combines brevity of SDL with flexibility of `graphql-js`.
```
stefanprobst:
```suggestion
While `createTypes` accepts GraphQL types as a [Schema Definition Language (SDL)](https://graphql.org/learn/schema/) string, we've also added an option to use `graphql-js` types so that users could create types with resolvers. However, `graphql-js` is somewhat verbose and it can be hard to refer to types that don't yet exist or don't exist in a current scope. Therefore, we decided to add another programmatic API that combines brevity of SDL with flexibility of `graphql-js`.
```

docs/blog/2019-03-18-releasing-new-schema-customization/index.md
```diff
@@ -0,0 +1,131 @@
...
+    }),
+  ])
+}
+```
+
+# Potential for Breaking Changes
+
+We have tried to avoid any breaking changes in this refactor of the underlying GraphQL layer, testing it in notable Gatsby sites and ensuring all tests were passing. However, there are areas where we needed to introduce more stable naming, and in these instances it _could_ be possible that a breaking change was introduced if you were relying on this undocumented API.
+
+Specifically, before this refactor Gatsby type names weren't stable. They could have names like `frontmatter_2` because of some quirks in our schema generation. Now the types names are **stable** and **defined**. For a `Node`, it's always a Pascal Camel Cased name of the `Node` type (for example, `AllMarkdownRemark`). For an inline object, it's the name of the node plus the name of the field, again Pascal Camel Cased. So `frontmatter_2` would be available as `MarkdownRemarkFrontmatter` now. If you've had fragments referring to some types by their old names, you may need to change it to new names, e.g.:
```
stefanprobst:
```suggestion
Specifically, before this refactor Gatsby type names weren't stable. They could have names like `frontmatter_2` because of some quirks in our schema generation. Now the types names are **stable** and **defined**. For a `Node`, it's always a Pascal Camel Cased name of the `Node` type (for example, `MarkdownRemark`). For an inline object, it's the name of the node plus the name of the field, again Pascal Camel Cased. So `frontmatter_2` would be available as `MarkdownRemarkFrontmatter` now. If you've had fragments referring to some types by their old names, you may need to change it to new names, e.g.:
```

PR #11484, opened 2019-02-01
feat(gatsby-plugin-netlify): Allow status codes in redirects (#11255) 

### LINKED ISSUE #11255: gatsby-plugin-netlify does not support rewrites, or setting a different status code
## Summary

The netlify plugin allows you to add redirects programmatically. It does differentiate between permanent and non-permanent redirects (status code 301 or 302) but does not allow any other values.

Netlify redirect rules are as follows:

```
FROM TO STATUS_CODE[!] [key1=value1] [key2=value2]
```
The plugin allows you to set all of these except the status code:

 - `FROM` is set with `fromPath`
 - `TO` is set with `toPath`
 - The exclamation mark is set with `force`
 - Key value pairs are also supported

But the only way to control the status code is to get it to switch from 301 to 302 by setting `isPermanent` to `true`.

Given that this is a netlify-specific plugin, and Netlify supports plenty of status codes, I feel there should be a way to handle this through the plugin. The most common use-cases are:

 - Set status code 200 for rewrites (instead of redirects)
 - Set status code 404 for custom 404 pages

### Basic example

Redirect creation is handled [here](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-netlify/src/create-redirects.js#L17) as such:

```js
redirects = redirects.map(redirect => {
    const {
      fromPath,
      isPermanent,
      redirectInBrowser, // eslint-disable-line no-unused-vars
      force,
      toPath,
      ...rest
    } = redirect

    let status = isPermanent ? `301` : `302`

    if (force) status = status.concat(`!`)

    // The order of the first 3 parameters is significant.
    // The order for rest params (key-value pairs) is arbitrary.
    const pieces = [fromPath, toPath, status]
```

I would propose to add a `statusCode` that (if it's set) would take precendce over the `isPermanent` setting. Something like this:

```js
redirects = redirects.map(redirect => {
    const {
      fromPath,
      isPermanent,
      redirectInBrowser, // eslint-disable-line no-unused-vars
      force,
      toPath,
      statusCode,
      ...rest
    } = redirect

    let status = isPermanent ? `301` : `302`
    if (statusCode) status = statusCode

    if (force) status = status.concat(`!`)

    // The order of the first 3 parameters is significant.
    // The order for rest params (key-value pairs) is arbitrary.
    const pieces = [fromPath, toPath, status]
```

### Motivation

If `statusCode` defaults to `FALSE` this wouldn't break anything for existing users, yet allow people to use this plugin to fully leverage the Netlify redirects/rewrites system.

Not to mention that unless I'm missing something, this is a rather small change.

joostdecock:
## Description

This addresses #11255. A request to implement support for status
codes other that 301/302 for this plugin, as Netlify also supports
other status codes.

@sidharthachatterjee suggested I submitted a PR
with my proposed changes, and this is that PR.

It adds one option to the `createRedirect` call which is `statusCode`.
You can use it to manually set the status code. If it's not set, the
status code will be determined by the `isPermanent` option as before
(`301` if `true` or `302` if not (the default) ).

I've also updated the README to document this new feature.

## Related Issues

Fixes #11255

### DIFF
--- a/packages/gatsby-plugin-netlify/README.md
+++ b/packages/gatsby-plugin-netlify/README.md
@@ -113,10 +113,9 @@ You can create redirects using the [`createRedirect`](https://www.gatsbyjs.org/d
 
 In addition to the options provided by the Gatsby API, you can pass these options specific to this plugin:
 
-| Attribute | Description                                                                                                                                                                                                        |
-| --------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| Attribute | Description |
+| --------- | ----------- |
 | `force`   | Overrides existing content in the path. This is particularly useful for domain alias redirects. See [the Netlify documentation for this option](https://www.netlify.com/docs/redirects/#structured-configuration). |
-
 | `statusCode` | Overrides the HTTP status code whih is `302` by default, or `301` when [`isPermanent`](https://www.gatsbyjs.org/docs/actions/#createRedirect) is `true`. Since Netlify supports more status codes, this allows you to set a different status code. Like `200` for a rewrite, or `404` for a custom error page. See [the Netlify documentation for this option](https://www.netlify.com/docs/redirects/#http-status-codes). |
 
 An example:
--- a/packages/gatsby-plugin-netlify/README.md
+++ b/packages/gatsby-plugin-netlify/README.md
@@ -116,7 +116,7 @@ In addition to the options provided by the Gatsby API, you can pass these option
 | Attribute | Description |
 | --------- | ----------- |
 | `force`   | Overrides existing content in the path. This is particularly useful for domain alias redirects. See [the Netlify documentation for this option](https://www.netlify.com/docs/redirects/#structured-configuration). |
-| `statusCode` | Overrides the HTTP status code whih is `302` by default, or `301` when [`isPermanent`](https://www.gatsbyjs.org/docs/actions/#createRedirect) is `true`. Since Netlify supports more status codes, this allows you to set a different status code. Like `200` for a rewrite, or `404` for a custom error page. See [the Netlify documentation for this option](https://www.netlify.com/docs/redirects/#http-status-codes). |
+| `statusCode` | Overrides the HTTP status code which is set to `302`  by default or `301` when [`isPermanent`](https://www.gatsbyjs.org/docs/actions/#createRedirect) is `true`. Since Netlify supports custom status codes, you can set one here. For example, `200` for rewrites, or `404` for a custom error page. See [the Netlify documentation for this option](https://www.netlify.com/docs/redirects/#http-status-codes). |
 
 An example:
 
--- a/packages/gatsby-plugin-netlify/README.md
+++ b/packages/gatsby-plugin-netlify/README.md
@@ -123,6 +123,7 @@ An example:
 ```javascript
 createRedirect({ fromPath: "/old-url", toPath: "/new-url", isPermanent: true })
 createRedirect({ fromPath: "/url", toPath: "/zn-CH/url", Language: "zn" })
+createRedirect({ fromPath: "/url_that_is/not_pretty", toPath: "/pretty/url", statusCode: 200 })
 ```
 
 You can also create a `_redirects` file in the `static` folder for the same effect. Any programmatically created redirects will be appended to the file.
--- a/packages/gatsby-plugin-netlify/README.md
+++ b/packages/gatsby-plugin-netlify/README.md
@@ -113,17 +113,21 @@ You can create redirects using the [`createRedirect`](https://www.gatsbyjs.org/d
 
 In addition to the options provided by the Gatsby API, you can pass these options specific to this plugin:
 
-| Attribute | Description |
-| --------- | ----------- |
-| `force`   | Overrides existing content in the path. This is particularly useful for domain alias redirects. See [the Netlify documentation for this option](https://www.netlify.com/docs/redirects/#structured-configuration). |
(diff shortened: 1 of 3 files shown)


sidharthachatterjee:
Thank you so much, @joostdecock 

Could you please rebase over master once? I think Prettier behaviour changed from the last version and that's why we're seeing the linting errors

packages/gatsby-plugin-netlify/README.md
```diff
@@ -111,15 +111,20 @@ You can validate the `_headers` config through the
 
 You can create redirects using the [`createRedirect`](https://www.gatsbyjs.org/docs/actions/#createRedirect) action.
 
+In addition to the options provided by the Gatsby API, you can pass these options specific to this plugin:
+
+| Attribute | Description |
+| --------- | ----------- |
+| `force`   | Overrides existing content in the path. This is particularly useful for domain alias redirects. See [the Netlify documentation for this option](https://www.netlify.com/docs/redirects/#structured-configuration). |
+| `statusCode` | Overrides the HTTP status code whih is `302` by default, or `301` when [`isPermanent`](https://www.gatsbyjs.org/docs/actions/#createRedirect) is `true`. Since Netlify supports more status codes, this allows you to set a different status code. Like `200` for a rewrite, or `404` for a custom error page. See [the Netlify documentation for this option](https://www.netlify.com/docs/redirects/#http-status-codes). |
```
sidharthachatterjee: What do you think about adding an example of the new option?

sidharthachatterjee:
```suggestion
| `statusCode` | Overrides the HTTP status code which is set to `302`  by default or `301` when [`isPermanent`](https://www.gatsbyjs.org/docs/actions/#createRedirect) is `true`. Since Netlify supports custom status codes, you can set one here. For example, `200` for rewrites, or `404` for a custom error page. See [the Netlify documentation for this option](https://www.netlify.com/docs/redirects/#http-status-codes). |
```
Small typo and some rephrasing

joostdecock: Example added. Thanks for the suggestion.

joostdecock:
> Could you please rebase over master once? I think Prettier behaviour changed from the last version and that's why we're seeing the linting errors

My bad. I hand-edited the table markup to make it *prettier* but it seems prettier didn't agree.

packages/gatsby-plugin-netlify/src/create-redirects.js
```diff
@@ -21,10 +21,12 @@ export default async function writeRedirectsFile(
       redirectInBrowser, // eslint-disable-line no-unused-vars
       force,
       toPath,
+      statusCode, // eslint-disable-line no-unused-vars
```
pieh: is the `eslint-disable-line no-unused-vars` needed here?

KyleAMathews:
We'd need to update the docs too https://www.gatsbyjs.org/docs/actions/#createRedirect

joostdecock:
> We'd need to update the docs too https://www.gatsbyjs.org/docs/actions/#createRedirect

I assumed this documented the Gatsby `createRedirect` API? This adds one option (`statusCode`) which is specific to the plugin. But the plugin already has the `force` option, which is also specific to the plugin, and not documented in the Gatsby API.

Which is why I figured this should only be documented in the plugin README. If you disagree, I'd be happy to take a stab at updating the documentation, but I assumed it was intentionally not included in the Gatsby API docs.

If you can advice on the best course of action, I'd be happy to work on it.

packages/gatsby-plugin-netlify/src/create-redirects.js
joostdecock: I've removed it.

joostdecock: In a very unsurprising development, turns out you were right :+1:

joostdecock:
> If you can advice on the best course of action, I'd be happy to work on it.

As this has gone cold a bit, I figured I'd let you know that I'm still happy to update the documentation. But before doing so, someone would need to decide whether it's needed or not (see [my earlier question](https://github.com/gatsbyjs/gatsby/pull/11484#issuecomment-459954268)).

sidharthachatterjee:
I agree with @joostdecock that updating the createRedirect docs shouldn't be necessary since this only affects the netlify-plugin

pieh:
My gut feel right now is to merge as-is and we can discuss adding `statusCode` to gatsby's `createRedirect` docs in the meantime to not block this feature for `gatsby-plugin-netlify`, but will wait at least a day in case folks feel otherwise

KyleAMathews:
A plugin that requires a field creates a defacto standard. If we're going to add support to the plug-in, we should document it.

jariz:
Thanks for the mention @pieh.  
Considering we have multiple plugins being able to handle these statuscodes (such as plugin-netlify & plugin-s3), I think it's very vital to document it. (why not? it's backwards compatible anyway)   
It is being said that plugin-netlify will be the only one implementing this, but obviously I'm also interested in implementing this myself if it gives more freedom to the users, and so will any future plugin maintainers likely.

pieh:
@jariz You obviously can implement this as well, but my worry is that there are other plugins (that I'm not aware of) that implement that feature.

But I think You raised very valid point that we should document it in `createRedirect` docs, because this is the place where plugin maintainer will get their information - I just got convinced :)

@joostdecock We are in alignment here that we should document this option in https://www.gatsbyjs.org/docs/actions/#createRedirect

joostdecock:
> We are in alignment here that we should document this option in https://www.gatsbyjs.org/docs/actions/#createRedirect

That's great. I took a first stab and updated the (inline) docs. In've added the `statusCode` option (that I added in the PR) but also the `force` option that was already in the Netlify plugin.

@jariz I've added your S3 plugin as an example, but it would be good if you could see of the way I phrazed things works out for you.

Thanks in advance for your feedback.

jariz:
@joostdecock lgtm!

PR #11492, opened 2019-02-01
feat(docs): link stub to github issue

### LINKED ISSUE #11342: [docs] Map stubs to existing issues on GitHub
## Summary

Throughout the docs, we have a links in the sidebar that are greyed out noting an article that doesn't exist yet but would be useful to have. When you visit one, it currently has a notice saying 

### Basic example

For instance, the [How to run a Gatsby workshop](https://www.gatsbyjs.org/docs/how-to-run-a-gatsby-workshop/) page is a current stub, but we also have a issue open for such: #8847.

We could possibly add a field to the frontmatter in the stub file to link the github issue and then render it out somehow when generated.

```markdown
---
title: How to run a Gatsby workshop
issue: https://github.com/gatsbyjs/gatsby/issues/8847
---
```

Then, when a user visits the stub, we could guide them to the GitHub issue for said stub.

### Motivation

I believe adding links to the specific issues would possibly increase the experience for helping draw contributors. With this, they would know how to flesh out the page based upon the style guide as well as some specific guidance about the topic from the community / Gatsby team.

Not all stubs have matching issues, but for the ones that do, this could make the flow of seeing a stub to wanting to help flesh out the stub smoother.

isaac-martin:
## Description
- Adds a field to docs frontmatter to link to an issue on GitHub
- If this field is present it will generate a link to the issue on GitHub
- Updates documentation on removing the link to issue when a `stub` is converted to a `doc`

## Related Issues
Closes #11342

Not sure the style of the link is the best way to display this. Happy to have someone provide contribution to this.

### DIFF
--- a/docs/docs/how-to-contribute.md
+++ b/docs/docs/how-to-contribute.md
@@ -81,7 +81,19 @@ Check the GitHub repo for issues labeled with ["documentation" and "good first i
    > Note: If your issue and/or PR doesn't meet the above contribution criteria, it may receive a comment reminding you to do so. If, after two weeks, these updates haven't been made, your issue and/or PR may be closed, which helps us triage issues and PRs efficiently. You can request that it be reopened if and when you are ready to make the updates required.
 3. GitHub then allows you to commit the change and raise a PR right in the UI. This is the _easiest_ way you can contribute to the project!
 
-If you wrote a new document that was previously a stub, update `www/src/data/sidebars/doc-links.yaml` accordingly by removing the asterisk behind the document's title:
+If you wrote a new document that was previously a stub, there is two things you need to update.
+
+1. Remove the frontmatter that links to the issue
+
+```diff:title=docs/docs/example-doc.md
+  ...
+    title: Example Document
+- - issue: https://github.com/gatsbyjs/gatsby/issues/00000
++ -
+  ...
+```
+
+2. Edit `www/src/data/sidebars/doc-links.yaml` by removing the asterisk behind the document's title:
 
 ```diff:title=www/src/data/sidebars/doc-links.yaml
   ...
--- a/docs/docs/how-to-contribute.md
+++ b/docs/docs/how-to-contribute.md
@@ -81,7 +81,7 @@ Check the GitHub repo for issues labeled with ["documentation" and "good first i
    > Note: If your issue and/or PR doesn't meet the above contribution criteria, it may receive a comment reminding you to do so. If, after two weeks, these updates haven't been made, your issue and/or PR may be closed, which helps us triage issues and PRs efficiently. You can request that it be reopened if and when you are ready to make the updates required.
 3. GitHub then allows you to commit the change and raise a PR right in the UI. This is the _easiest_ way you can contribute to the project!
 
-If you wrote a new document that was previously a stub, there is two things you need to update.
+If you wrote a new document that was previously a stub, there are two things you need to update.
 
 1. Remove the frontmatter that links to the issue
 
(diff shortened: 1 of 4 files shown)


pieh:
This is such an awesome idea!

/cc @gatsbyjs/docs - need some 👀 to verify wording in contribution instructions and on link to the related issue

Also copying screenshot provided by @isaac-martin in #11492 to show how stubs look like with this change:
image

docs/docs/how-to-contribute.md
```diff
@@ -81,7 +81,19 @@ Check the GitHub repo for issues labeled with ["documentation" and "good first i
    > Note: If your issue and/or PR doesn't meet the above contribution criteria, it may receive a comment reminding you to do so. If, after two weeks, these updates haven't been made, your issue and/or PR may be closed, which helps us triage issues and PRs efficiently. You can request that it be reopened if and when you are ready to make the updates required.
 3. GitHub then allows you to commit the change and raise a PR right in the UI. This is the _easiest_ way you can contribute to the project!
 
-If you wrote a new document that was previously a stub, update `www/src/data/sidebars/doc-links.yaml` accordingly by removing the asterisk behind the document's title:
+If you wrote a new document that was previously a stub, there is two things you need to update.
```
jlengstorf:
```suggestion
If you wrote a new document that was previously a stub, there are two things you need to update.
```

marcysutton:
This is great, thanks for working on it! I have two questions:
- are there any tests for this part of the site that could be updated (I'm new here)?
- is the next step to update all stub frontmatter to link to corresponding Github issues?

isaac-martin:
- Not sure re tests. It's a pretty minimal change and nothing failed so I guess that's good news. 
- yep need to update front matter. I don't mind doing that was just waiting for this to be merged so we wouldn't have to change to much if this wasn't the convention

KyleAMathews:
Love this idea!

There's no particular reason to add this via frontmatter. Since we have to edit every stub anyways, could we just directly change the markdown? This way we can modify the stubs easily e.g. add a suggested list of things to cover etc.

isaac-martin:
Mmm that's a valid point Kyle, however I guess keeping all details (eg suggested list of things to cover) on the issue lead to more opportunities for discussion and collaboration? 

Also then making a new stub is just copying a current stub and updating title and issue?

And finally having it the way it is ☝️ensures consistent wording and design and behavior without whoever is merging stubs having to manually check / request changes. 

What's everyone else's thoughts on this - happy to take whatever feedback on board, don't want to come in here, make one PR and say my way or the highway.

KyleAMathews:
The downside is they now can't vary.

Also, only one person has created stubs to date so it's relatively easy to keep things consistent.

This isn't a big deal either way though so you can also just ignore my comments 😂

jlengstorf:
I like the frontmatter idea. It keeps stubs simple and avoids editing markup in copy-pasta. It probably doesn't make a huge difference in the long run, so I don't see a need to change what's already implemented, personally.

shannonbux:
Ok awesome. So what are the final instructions that need to be given to contributors and others who write or update stub articles?

isaac-martin:
Hey @shannonbux [I updated contributing docs](https://github.com/gatsbyjs/gatsby/pull/11492/files#diff-3408697fdc772fa5e3a7f4c7efc99faf)

However on second look there may need to be some other info added in any docs about contributing a new stub? This only takes into account converting an existing stub to a 'completed' doc.

marcysutton:
There is a page on Docs Templates (which will likely move to the new Contributing section)...perhaps we could add another one on "Creating a stub"? https://www.gatsbyjs.org/docs/templates/

shannonbux:
I think that seems like the right place, @marcysutton, although I haven't tested how easy-to-find those templates are. Since all the templates show different markdown front matter examples, it seems like the right place! One interesting note is that for those unfamiliar with markdown, the phrase "front matter" might not make sense, so showing people examples is important!

marcysutton:
An update: I've added a new "how to make a stub" page in the contributing docs, which will be in PR-form later today.

PR #11493, opened 2019-02-02
docs(gatsby-plugin-catch-links): clarify this handle all links (and not just markdown)

### LINKED ISSUE #11485: docs(gatsby-plugin-catch-links): Make it more clear that not only markdown links are affected
In the readme, only markdown is mentioned. But does this plugin work with every link inserted through `dangerouslySetInnerHtml` or equivalent?

Example: I fetch a WYSIWYG text with internal links from WordPress and that is inserted as HTML.

If not, how could I replace HTML links with the `gatsby-link` component?

jordanoverbye:
This issue hopefully fixes #11485


packages/gatsby-plugin-catch-links/README.md
```diff
@@ -1,24 +1,33 @@
...
 ## How to use
 
 ```javascript
 // In your gatsby-config.js
 plugins: [`gatsby-plugin-catch-links`]
 ```
+
+## Examples
+
+Check out the [_Using Remark_ example](https://github.com/gatsbyjs/gatsby/tree/master/examples/using-remark) to see this plugin in action.
```
pieh: Maybe we could also link to https://using-remark.gatsbyjs.org/copy-linked-files-intercepting-local-links/#intercepting-local-links to showcase this in action?

jordanoverbye: Sure, thats a good idea @pieh - I've just pushed another commit.

jordanoverbye: Hey @pieh, are there any other changes you would like done?

packages/gatsby-plugin-catch-links/README.md
```diff
@@ -1,24 +1,33 @@
...
-`npm install --save gatsby-plugin-catch-links`
+```
+npm install --save gatsby-plugin-catch-links
+```
+
+or
+
+```
+yarn add gatsby-plugin-catch-links
+```
```
pieh: we don't use `yarn` in our docs for users (just convention - we just think that if someone is using `yarn`, they will be able to translate `npm install` into `yarn`, so let's just remove "or" part

packages/gatsby-plugin-catch-links/README.md
pieh: Hey, sorry for radio silence - I just added small comment

packages/gatsby-plugin-catch-links/README.md
jordanoverbye: @pieh no problem, just pushed another commit to fix this.

Issue #8697, opened 2018-10-02
Starter for creating source plugins


ajayns:
## Summary

I've been reading through the [Source Plugin Tutorial](https://www.gatsbyjs.org/docs/source-plugin-tutorial/). It's great, I love the narrative approach to building a plugin, but the problem is when it comes to the part where it should explain how to publish on npm, it lacks a bit and is not very helpful. 

### Basic example

I can recommend two things to improve the docs here: 
One would be to add a short section on how a source-plugin published to npm would look like and how to convert the current plugin for publishing (maybe even test with npm link)
Another things would be a Starter for source plugins it could have babel compilation, testing and such already setup for the ease of the user.

### Motivation

The primary goal is that, the user can get to code as soon as possible and understand how things work in the simplest of the ways. A starter for source plugins would mean you could understand what's happening by taking a glance at the structure and running it and also it makes kickstarting source-plugin development super fast.

kakadiadarpan:
@ajayns I agree with you about the `Starter for source plugins`, that would be amazing to have.

Could you share more information about the following:
>how a source-plugin published to npm would look like?
>how to convert the current plugin for publishing?

ajayns:
I'm not actually sure of those questions as I haven't read anything that recommends anything regarding them, but recently I've worked on a source plugin for Plone and will share a couple of things I figured then.

> How a source plugin published to npm would look like

I'm unaware of any structure recommended by the Gatsby community so the setup I used was a `gatsby-node.js` that's transpiled by Babel (this enables support for latest ES6+ features), another js file for helper functions (functions are exported here, making it easy for tests to be written), Jest for testing. This we came up with looking at a couple of plugins that were popular. Take a look at what I'm talking about: https://github.com/collective/gatsby-source-plone/
As I mentioned, I'm not sure about what's recommended, this is just what works for us, please do let me know if I'm wrong and I can improve on this.

> How to convert the current plugin for publishing?

This answer would be pretty easy by just mentioning what all files a gatsby-source plugin would include: which primarily is the `gatsby-node.js`. An example structure of plugin with explanation for use of each file should be enough to figure out the rest.

pieh:
I think skeleton/scaffold for plugins is good idea - i could see at least 2 templates - for source and transform plugins. For packages in gatsby repo we have plop template but this relies on gatsby repo babel setup, so for standalone plugins we would need something different.

[...]

PR #11500, opened 2019-02-02
[WIP] Plugin Starter

moonmeister:
## Description

@jlengstorf and I have been chatting about this for a while...so I sat down and took a first stab at creating a starter for plugins. This template supplies it's own build, test, deploy scripts and dependencies (meaning it's assumed a plugin using this template is not in the gatsby monorepo).

## TODO

- The docs still need a lot of work. 
- Confirm various configs are ok
- Confirm various ignore files are ok
- I've made some initial doc assumptions that there would be a `gatsby new plugin` CLI interface to initialize this. I wouldn't expect this to be part of this PR and this can be changed as the @gatsbyjs/docs team sees fit

## Related Issues

Related to #8697


starters/plugin/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "gatsby-plugin-starter",
```
jlengstorf: This naming is probably going to be a problem, because the `gatsby-plugin-*` pattern is used for full-on plugins.

I don't have a strong opinion about this, but maybe `gatsby-template-plugin`?

moonmeister:
@jlengstorf I'm not understanf the practical difference between a template and a starter. Can you elaborate on how that would change this PR and/or a developers workflow?

jlengstorf:
@moonmeister in my mind, a "starter", in Gatsbyland, is an independent repo that can be run standalone, unmodified. The CLI for starters clones that repo, installs dependencies, and strips out the `.git` stuff.

A "template", to my mind, would be something used by e.g. [plop](https://www.npmjs.com/package/plop) that is _not_ standalone and would look something like this (for a `package.json`:

```json
{
  "name": "{{kebabCase name}}",
  ...
}
```

The trade-offs are, in my opinion:

✅ Pros of templates | ❌ Cons of templates
---- | ----
More deliberate, which leads to a smaller number of higher-quality plugin templates. | A little more cumbersome to maintain, which can make maintenance a little more of a hassle.
Files are customized out of the gate via CLI prompts (e.g. the `name` is already set) | Not editing the files by hand might make it feel more mysterious to use/edit files?
Keeping all plugin templates in one place makes it easier to manage upgrades over time. | All plugin templates will need to live and be maintained in the core gatsby repo.
Templates can include functionality on an opt-in basis (e.g. "do you need to import images from a third-party source?") | 

I feel like someone like @m-allanson, @pieh, or @DSchau  is probably better suited to weigh these trade-offs, but since we're starting from near-zero on this CLI it's probably worth discussing before we get too far down any path.

shannonbux:
Really interesting ideas here! I don't think the docs team would make the
final call here, at least how the responsibilities currently stand. It'd be
the https://github.com/orgs/gatsbyjs/teams/starters team if this ends up
being a starter, or...not sure who if it ends up being a template!
@jlengstorf do you know?

moonmeister:
@jlengstorf Thanks for that explanation. I agree, templates make sense. We'll need to make sure we account for whether the plugin will be in the Gatsby monorepo or not.

jlengstorf:
@shannonbux I think @gatsbyjs/core will probably have final approval here.

@moonmeister If we go with templates, these would probably live inside the CLI codebase.

moonmeister:
@jlengstorf clarification: I didn't mean whether the template itself lives inside the monorepo...that I'd expect. But whether the plugin being created will...the resulting files/config will look different depending on whether the plugin lives in the monorepo or outside in it's own repo.

KyleAMathews:
We already have a plugin creator for plugins here: https://github.com/gatsbyjs/gatsby/tree/master/plop-templates/package

jlengstorf:
@KyleAMathews where is that called from? Is there a CLI command I'm not aware of?

@moonmeister I think the goal should be to create files in such a way that they're publishable to npm. However, I _think_ we can make sure that the same template will work both locally and as a package.

pieh:
@jlengstorf you can use `yarn plop` in our repo, which will ask you few questions ( defined in https://github.com/gatsbyjs/gatsby/blob/master/plopfile.js ) and will scaffold package or example. I'm not sure how easy it would be to package it with `gatsby-cli` and make sure we avoid duplication

jlengstorf:
@pieh Ah, that makes sense. Thanks!

> I'm not sure how easy it would be to package it with `gatsby-cli` and make sure we avoid duplication

Can you clarify what you mean by duplication?

pieh:
> > I'm not sure how easy it would be to package it with `gatsby-cli` and make sure we avoid duplication
> 
> Can you clarify what you mean by duplication?

I just meant that we could straight reuse our plop templates for `gatsby-cli` to scaffold new plugin (which potentially would mean that we would have copy of templates in `plop-templates` and in `gatsby-cli`.

But after few moments, I realized we could just switch to scaffold new packages here (which we do very rarely) using `gatsby-cli` as well - so that comment can be discarded.

sidharthachatterjee:
@moonmeister Thank you so much for opening this and for all the work you've put into this!

The @gatsbyjs/core team was just discussing this and I think we all agree with the points made by @jlengstorf in https://github.com/gatsbyjs/gatsby/pull/11500#pullrequestreview-202700934

More specifically, we think that maintaining multiple such starters for every source plugin will add a lot of overhead and templates definitely seem like the way going forward! I think we ought to continue the discussion around them in https://github.com/gatsbyjs/gatsby/issues/8697 and thank you again for driving this forward and getting us started. 

Let's close this for now but please feel free to publish this as a starter in a separate repository! Thank you so much for your continued contributions and we look forward to many more 💜 💪

PR #11501, opened 2019-02-02
feat(www): use blog post preview component on tags pages

### LINKED ISSUE #11491: Sort tagged blog post by date + add author avatar/name
E.g. on pages like https://www.gatsbyjs.org/blog/tags/themes

screenshot 2019-02-01 14 56 15

https://github.com/gatsbyjs/gatsby/blob/master/www/src/templates/tags.js

Saifadin:
## Improvements to Tags Page
Adding more details to the tags Page. To accomplish that, I reused the `BlogPostPreviewItem` component that is already used by the `contributor`-template.

This shoud fix #11491

It looks like this now:

screenshot 2019-02-02 at 10 45 39

### DIFF
--- a/www/package.json
+++ b/www/package.json
@@ -93,7 +93,7 @@
   "scripts": {
     "build": "gatsby build",
     "deploy": "gatsby build --prefix-paths && gh-pages -d public",
-    "develop": "gatsby develop",
+    "develop": "GATSBY_GRAPHQL_IDE=playground gatsby develop",
     "start": "npm run develop",
     "serve": "gatsby serve",
     "test": "echo \"Error: no test specified\" && exit 1",
--- a/www/src/templates/tags.js
+++ b/www/src/templates/tags.js
@@ -1,9 +1,10 @@
 import React from "react"
-import PropTypes from "prop-types"
 import { Link, graphql } from "gatsby"
 
+import BlogPostPreviewItem from "../components/blog-post-preview-item"
 import Container from "../components/container"
 import Layout from "../components/layout"
+import { rhythm } from "../utils/typography"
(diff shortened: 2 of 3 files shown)


www/src/templates/tags.js
```diff
@@ -16,52 +17,26 @@ const Tags = ({ pageContext, data, location }) => {
...
-}
-
 export default Tags
 
 export const pageQuery = graphql`
   query($tag: String) {
     allMarkdownRemark(
       limit: 2000
-      sort: { fields: [frontmatter___date, fields___slug], order: DESC }
+      sort: { fields: [frontmatter___date], order: DESC }
```
pieh: Why remove this sort field? This makes sure we don't have potentially inconsistencies between builds when there are multiple blog posts from same day/date (sometimes one post can be before the other and in next build it might change - depending on order of node creation).

pieh: (I know it's pretty fringe issue, but please, let's not remove it)

www/package.json
```diff
@@ -93,7 +93,7 @@
   "scripts": {
     "build": "gatsby build",
     "deploy": "gatsby build --prefix-paths && gh-pages -d public",
-    "develop": "gatsby develop",
+    "develop": "GATSBY_GRAPHQL_IDE=playground gatsby develop",
```
pieh: hahah 👍

pieh: I don't mind having this in - especially if we want to move forward with playground, using it for `gatsbyjs.org` dev seems actually good idea

Saifadin:
Just checking in, any more changes wanted here?

PR #11517, opened 2019-02-03
fix(gatsby): Fix Loki query operators special casing

stefanprobst:
There are subtle differences between Loki and Sift query operators that we are trying to account for. This fixes some of those:
* Sift allows querying arrays with `$eq` and `$ne`, which translates to `$contains` and `$containsNone` in Loki
* The special case for `$nin` added `false`, which should be `undefined`. This also needed removing cloning the query args with JSON.parse/stringify, because undefined=>null, but it does not seem to serve a purpose anymore anyway.
* The fix for `{ $ne: true }` now handles nesting levels > 2, e.g. ` { "foo.bar.baz": { $ne: true } }`
* The conversion to dotted query fields was producing `{ "": {} }` for an empty object.

### DIFF
--- a/packages/gatsby/src/db/loki/nodes-query.js
+++ b/packages/gatsby/src/db/loki/nodes-query.js
@@ -117,6 +117,18 @@ function toMongoArgs(gqlFilter, lastFieldType) {
         const mm = new Minimatch(v)
         mongoArgs[`$regex`] = mm.makeRe()
       } else if (
+        k === `eq` &&
+        lastFieldType &&
+        lastFieldType.constructor.name === `GraphQLList`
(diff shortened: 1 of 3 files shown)


freiksenet:
CC @Moocar

Moocar:
@stefanprobst thanks so much! This looks great. The only thing I'm unsure of is the removal of `JSON.parse/stringify` as that was there before my time. But seeing as its removal doesn't break any tests, I think you're right in removing it.

Moocar:
@freiksenet looks like I don't have merging rights, so I'll leave it up to you to merge (I approve).

PR #11521, opened 2019-02-03
chore(benchmarks): update versions of dependencies

stefanprobst:
Update `gatsby` deps in benchmarks and integration tests (`next` seems to still refer to the RC)


benchmarks/markdown/package.json
```diff
@@ -8,8 +8,8 @@
     "serve": "gatsby serve"
   },
   "dependencies": {
-    "gatsby": "^2.0.9",
-    "gatsby-transformer-remark": "^2.1.4",
+    "gatsby": "latest",
```
KyleAMathews: Instead of "latest", could you change these to the actual latest versions? I'd rather be explicit about versions.

KyleAMathews:
Thanks for noticing this and putting up a fix!

benchmarks/markdown/package.json
stefanprobst: Done!

e2e-tests/development-runtime/package.json
```diff
@@ -4,18 +4,18 @@
...
+    "gatsby-plugin-offline": "^2.0.22",
+    "gatsby-plugin-react-helmet": "^3.0.6",
+    "gatsby-plugin-sharp": "^2.0.20",
+    "gatsby-source-filesystem": "^2.0.20",
+    "gatsby-transformer-remark": "^2.2.3",
+    "gatsby-transformer-sharp": "^2.1.13",
     "prop-types": "^15.6.2",
-    "react": "16.7.0-alpha.2",
-    "react-dom": "16.7.0-alpha.2",
+    "react": "16.7.0",
```
DSchau: We actually want to bump this to 16.8.0

```suggestion
    "react": "16.8.0",
```

(we rely on hooks, so need to have that!)

e2e-tests/development-runtime/package.json
```diff
@@ -4,18 +4,18 @@
...
+    "react-dom": "16.7.0",
```
DSchau:
```suggestion
    "react-dom": "16.8.0",
```

same here!

e2e-tests/development-runtime/package.json
DSchau: (for the e2e-tests/development-runtime tests, at least!)

PR #11525, opened 2019-02-03
feat(gatsby-dev-cli): add verdaccio support

pieh:
This adds verdaccio (self run/hosted NPM registry) support to `gatsby-dev-cli`. There is no extra configuration required - it's all handled internally by `gatsby-dev-cli`.

## Goals:
 - most important - handle dependency changes
 - remove extra manual work when using it - remove the need to install dependencies before using gatsby-dev (i.e instead of running yarn you just run gatsby-dev and it handles dependency installation and also changes from monorepo). If deps are installed and dependency didn't change - it will not add extra step of installing deps in local site

## The way it works:
 - on initial monorepo package file globbing it will examine `package.json` files:
   - if package is already installed locally in example site it will compare dependencies in that package. If new dependency was added, or dependency changed version selector/range - it will queue publishing for that package.
   - if package is not yet installed - it will check unkpg.com to see if we need to publish dev version of package or can use package from public NPM repostiroy
 - when package is queued for publishing - it checks if any other gatsby package depend on it and add them to publishing queue (i.e. dependency change in `gatsby-plugin-page-creator` will cause `gatsby` package to be published as well)
 - once final list of packages to be published is done, it start publishing process:
   - if there are any packages that need to be published, it starts temporary verdaccio server:
     - server is configured to:
       - allow anonymous publishes (so user don't have to login to that local registry)
       - set up storage that is set to be `gatsby-dev-cli/verdaccio`
     - everytime server starts it has clean slate (it clears its storage
   - it temporarily add `.npmrc` to package to allow using `npm publish` without being logged in to that local npm repository ( see https://github.com/verdaccio/verdaccio/issues/212#issuecomment-308578500 - this comment was made almost 2 years ago - right now `yarn` also require to be logged in). This tricks `npm` to think it's logged in by applying dummy access token
   - it temporarirly edit `package.json` and adjust package version and version selector for any dependencies that are in packages to be published list (this is needed for deps of deps to install correctly)
   - it publishes that package to local registry with `gatsby-dev` dist-tag
   - it reverts package.json changes and removes temporary `.npmrc` file
   - it installs modified "top-level" packages (so if you changed `gatsby-cli` - it will run `yarn add gatsby@gatsby-dev`
     - because we adjusted version selectors - this will correctly install modified `gatsby-cli@gatsby-dev` too
     - potential issue - this changes `package.json`/`yarn.lock` in local site - I potentially could restore it, but I think this would cause more troubles because `yarn` uses those files as source of truth, so leaving it modified seems like better option

## Limitations:
 - right now it only handle dependency changes on initial scan and doesn't work later on (it will output notices to terminal, but because `gatsby-dev-cli` is "fire-and-forget" it's unlikely that people will see them quickly. This is potentially solvable but would add more work and delay this.

## Changes:
 - Copying files for initial scan is postponed to after `ready` event is emitted and (if any) packages are installed. That's because now it also potentially install packages from public NPM, so I need to make sure that files are copied after that step


packages/gatsby-dev-cli/src/verdaccio.js
```diff
@@ -0,0 +1,137 @@
...
+  version,
+}) => {
+  // <ake sure verdaccio is running
+  await startServer()
+
+  // npm unpublish
+  // because version doesn't change when developing locally verdaccio wouldn't
+  // let republish package with same version - so as quick workaround we
+  // unpublish package.
+  // Ideally we could check if we need to unpublish: i.e. compare deps of
```
juanpicado: If you force publish `npm publish --force` the version is replaced. Perhaps that helps.

pieh: ohhh, cool, thanks for the tip! But there is another issue that will render this a bit irrelevant - let me write up follow up comment to document what I mean

pieh:
My initial implementation that publish to local repository with same package version has couple of issues when handling cases of nested deps (i.e. `gatsby` depending on `gatsby-cli`, `gatsby-link` etc or `gatsby-source-drupal` depending on `gatsby-source-filesystem`):

When we publish modified version of nested dependency with same version, and then try to instal "top level" package `yarn` will be smart and don't reinstall deps of top level package (because they are installed)

So in next round what I will do is actually use `lerna` to publish `canary` version to local repository and then specifically install those `canary` versions in example site.

This brings some side effect of changing versions in monorepo packages so I will also need to make sure I store package.json content in memory and restore it after publishing to not introduce unwanted changes to package.json

packages/gatsby-dev-cli/src/verdaccio.js
```diff
@@ -0,0 +1,137 @@
...
+    [`unpublish`, `${packageName}@${version}`, `--registry=${registryUrl}`],
+  ]
+
+  console.log(
+    `Trying to unpublish ${packageName}@${version} to local registry, in case it was published before`
+  )
+  try {
+    await promisifiedSpawn(unpublishCmd)
+    console.log(`Unpublished ${packageName}@${version} from local registry`)
+  } catch {
```
DSchau: Do we want to log the error here? Is there an error?

pieh: There is an error because I made `promisifiedSpawn` to reject with exit code if it's non-zero and unpublish command will exit with non-zero if package doesn't exist. But as mentioned above - I will actually need to scratch this code because it doesn't handle deps of deps (i.e. if I change dep in `gatsby-cli` and re-install `gatsby` package - it won't reinstall `gatsby-cli` - because version didn't change) - I will use `lerna` to publish canary release to local registry and later on install that canary - yet to fully verify, but it "should work" (TM)

pieh: I'm not sure if you publish `canary` 2 times in a row (with same code) it will result in same "hash" or not - TBD

packages/gatsby-cli/package.json
```diff
@@ -26,7 +26,7 @@
     "opentracing": "^0.14.3",
     "pretty-error": "^2.1.1",
     "resolve-cwd": "^2.0.0",
-    "source-map": "^0.5.7",
+    "got": "^7.1.0",
```
pieh: just testing if deps changes cause publishing/installing - will be removed

packages/gatsby-dev-cli/src/local-npm-registry/check-deps-changes.js
```diff
@@ -0,0 +1,89 @@
...
+
+    const diff2 = difference(
+      localPKGjson.dependencies,
+      monorepoPKGjson.dependencies
+    )
+
+    let needPublishing = false
+    const depChangeLog = _.uniq(Object.keys({ ...diff, ...diff2 }))
+      .reduce((acc, key) => {
+        if (
```
pieh: need to add handling of temporary file changes when I set dependency to `gatsby-dev` dist-tag - as this cause random and confusing console output that you can't publish/install things

packages/gatsby-dev-cli/src/local-npm-registry/verdaccio.js
```diff
@@ -0,0 +1,171 @@
...
+      packageName,
+      packagesToPublish,
+      root,
+      versionPostFix,
+    })
+  }
+
+  // TO-DO install packages locally
+  // need to install intersection of `packagesToPublish` and `packages`
+  // (packages filtered out will be deps of packages to install)
```
pieh: this will cause changes to package.json and yarn.lock in local site (yarn doesn't support `--no-save` - https://github.com/yarnpkg/yarn/issues/1743). I tend to lean toward documenting this behaviour instead of hacking things (storing content of those files and restoring it after installation) as this could cause some weirdness with yarn as it does use those files as source of truth

pieh:
There is one thing that I need to do here actually.

My goal here is to remove the need to install dependencies before using `gatsby-dev` (i.e instead of running `yarn` you just run `gatsby-dev` and it handles dependency installation and also changes from monorepo).

What I need to add is to run `yarn` if deps are not installed in local site and there is no dependency changes in any `gatsby` package

pieh:
If anyone want to give it a try - I published `gatsby-dev-cli@dev-verdaccio` (temporary dist-tag)

packages/gatsby-dev-cli/src/local-npm-registry/check-deps-changes.js
```diff
@@ -0,0 +1,190 @@
...
+
+  let localPKGjson
+  let packageNotInstalled = false
+  try {
+    localPKGjson = JSON.parse(fs.readFileSync(newPath))
+  } catch {
+    packageNotInstalled = true
+    // there is no local package - so we still need to install deps
+    // this is nice because devs won't need to do initial package installation - we can handle this.
+    if (isInitialScan) {
```
wardpeet: we could rewrite this one into the following. Might be a bit better to read
```
  if (!isInitialScan) {
    console.log(
      `'${packageName}' doesn't seem to be installed. Restart gatsby-dev to publish it`
    )
    return {
      didDepsChanged: false,
      packageNotInstalled,
    }
  }

  // if package is not installed, we will do http GET request to
  // unkpg to check if dependency in package published in public
  // npm repository are different

  // this allow us to not publish to local repository
  // and save some time/work
  try {
    localPKGjson = await new Promise((resolve, reject) => {
      request(`https://unpkg.com/${packageName}/package.json`, function (
        error,
        response,
        body
      ) {
        if (response && response.statusCode === 200) {
          return resolve(JSON.parse(body))
        }

        return reject(error)
      })
    })
  } catch {
    console.log(
      `'${packageName}' doesn't seem to be installed and is not published on NPM.`
    )
    return {
      didDepsChanged: true,
      packageNotInstalled,
    }
  }
```

packages/gatsby-cli/package.json
```diff
@@ -13,6 +13,7 @@
     "@babel/code-frame": "^7.0.0",
     "@babel/runtime": "^7.0.0",
     "bluebird": "^3.5.0",
+    "cat-me": "1.0.3",
```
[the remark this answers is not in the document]
pieh: Sid said the same! :D 

This is just to test if it works correctly - and hey - we have cats now in CircleCi output for e2e tests :)

packages/gatsby-dev-cli/src/watch.js
```diff
@@ -3,15 +3,75 @@ const _ = require(`lodash`)
...
+/*
+ * non-existant packages break on('ready')
+ * See: https://github.com/paulmillr/chokidar/issues/449
+ */
+function watch(root, packages, { scanOnce, quiet, monoRepoPackages }) {
+  let afterPackageInstallation = false
+  let queuedCopies = []
+
+  const realCopyPath = ({ oldPath, newPath, quiet, resolve, reject }) => {
+    console.log(`wat`)
```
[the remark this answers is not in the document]
pieh: lol, thanks!

packages/gatsby-dev-cli/src/local-npm-registry/utils.js
```diff
@@ -0,0 +1,28 @@
+const child_process = require(`child_process`)
+const path = require(`path`)
+
+exports.promisifiedSpawn = ([cmd, args = [], spawnArgs = {}]) =>
```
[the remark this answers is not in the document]
pieh: probably would be cleaner - will check it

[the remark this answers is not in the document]
pieh: I replaced with `exaca`, but I kept `promisifiedSpawn` to handle some common defaults

packages/gatsby-dev-cli/src/watch.js
```diff
@@ -24,29 +84,68 @@ const copyPath = (oldPath, newPath, quiet) =>
...
+  }
 
   chokidar
     .watch(watchers, {
       ignored: [filePath => _.some(ignored, reg => reg.test(filePath))],
     })
-    .on(`all`, (event, filePath) => {
+    .on(`all`, async (event, filePath) => {
       const watchEvents = [`change`, `add`]
       if (_.includes(watchEvents, event)) {
```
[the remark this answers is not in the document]
pieh: 👍 done - our revolt against bracket hell has started

packages/gatsby-dev-cli/src/watch.js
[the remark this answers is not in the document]
pieh: bye, "wat" - you will be missed

packages/gatsby-dev-cli/src/watch.js
```diff
@@ -3,15 +3,75 @@ const _ = require(`lodash`)
...
+  packages.forEach(p => {
+    const pkgJson = require(path.join(root, `packages`, p, `package.json`))
+
+    const fromMonoRepo = _.intersection(
+      Object.keys({ ...pkgJson.dependencies }),
+      monoRepoPackages
+    )
+
+    fromMonoRepo.forEach(pkgName => {
+      if (!depTree[pkgName]) {
```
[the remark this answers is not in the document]
pieh: I went even further to make it concise :)

packages/gatsby-dev-cli/src/local-npm-registry/check-deps-changes.js
pieh: done!

packages/gatsby-dev-cli/src/index.js
```diff
@@ -73,15 +73,14 @@ let packages = Object.keys(
...
+  packages = monoRepoPackages
 } else {
-  const { dependencies } = JSON.parse(
-    fs.readFileSync(path.join(gatsbyLocation, `packages/gatsby/package.json`))
-  )
-  packages = packages
-    .concat(Object.keys(dependencies))
-    .filter(p => p.startsWith(`gatsby`))
+  // intersect dependencies with monoRepoPackags to get list of packages to watch
+  packages = _.intersection(monoRepoPackages, packages)
```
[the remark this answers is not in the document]
pieh: I still look into packages to find dependencies of dependencies - but I do this in `watch.js` and it's not limited to deps of `gatsby` package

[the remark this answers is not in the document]
wardpeet: this makes it more verbose that's all :)

packages/gatsby-dev-cli/src/local-npm-registry/check-deps-changes.js
```diff
@@ -0,0 +1,189 @@
...
+    const diff2 = difference(
+      localPKGjson.dependencies,
+      monorepoPKGjson.dependencies
+    )
+
+    let needPublishing = false
+    let isPublishing = false
+    const depChangeLog = _.uniq(Object.keys({ ...diff, ...diff2 }))
+      .reduce((acc, key) => {
+        if (monorepoPKGjson.dependencies[key] === `gatsby-dev `) {
```
wardpeet: remove space
```suggestion
        if (monorepoPKGjson.dependencies[key] === `gatsby-dev`) {
```

packages/gatsby-dev-cli/src/local-npm-registry/verdaccio.js
```diff
@@ -0,0 +1,235 @@
...
+    {
+      cwd: pathToPackage,
+    },
+  ]
+
+  console.log(
+    `Publishing ${packageName}@${newPackageVersion} to local registry`
+  )
+  try {
+    await promisifiedSpawn(publishCmd)
```
wardpeet: whenever I kill the process between publishing the npmrc & package.json aren't cleanup

packages/gatsby-dev-cli/src/watch.js
```diff
@@ -3,15 +3,71 @@ const _ = require(`lodash`)
...
-  new Promise((resolve, reject) => {
+const traversePackagesDeps = ({
+  root,
+  packages,
+  monoRepoPackages,
+  seenPackages = [...packages],
+  depTree = {},
+}) => {
+  packages.forEach(p => {
+    const pkgJson = require(path.join(root, `packages`, p, `package.json`))
```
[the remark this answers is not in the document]
pieh: Nice catch - I will add sanity checks and warning message!

packages/gatsby-dev-cli/src/watch.js
```diff
@@ -3,15 +3,71 @@ const _ = require(`lodash`)
...
+
+    // only traverse not yet seen packages to avoid infinite loops
+    const newPackages = _.difference(fromMonoRepo, seenPackages)
+
+    if (newPackages.length) {
+      newPackages.forEach(depFromMonorepo => {
+        seenPackages.push(depFromMonorepo)
+      })
+
+      traversePackagesDeps({
```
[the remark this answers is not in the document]
pieh: `seenPackages` is used to generate `newPackages`:
```
// only traverse not yet seen packages to avoid infinite loops
const newPackages = _.difference(fromMonoRepo, seenPackages)
```

`depTree` is used later (outside of this function) to see what other potential packages would need to be published - i.e. if we change dependency in `gatsby-cli` - checking `depTree` will know that `gatsby` package need to be published as well.

Does this explain it? I should add some comments there.

packages/gatsby-dev-cli/src/local-npm-registry/check-deps-changes.js
```diff
@@ -0,0 +1,189 @@
...
+      return {
+        didDepsChanged: false,
+      }
+    }
+  }
+
+  let localPKGjson
+  let packageNotInstalled = false
+  try {
+    localPKGjson = JSON.parse(fs.readFileSync(newPath))
```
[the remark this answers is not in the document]
pieh: `JSON.parse` accept buffers (or anything that has `toString` method), but I will add this to avoid confusion

packages/gatsby-dev-cli/src/local-npm-registry/check-deps-changes.js
```diff
@@ -0,0 +1,189 @@
...
+    // and save some time/work
+    try {
+      localPKGjson = await new Promise((resolve, reject) => {
+        request(`https://unpkg.com/${packageName}/package.json`, function(
+          error,
+          response,
+          body
+        ) {
+          if (response && response.statusCode === 200) {
+            resolve(JSON.parse(body))
```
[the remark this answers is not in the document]
pieh: This will cause eslint error ( https://eslint.org/docs/rules/consistent-return ) unless I do the same for `reject` - is it worth it?

packages/gatsby-dev-cli/src/local-npm-registry/check-deps-changes.js
pieh: oops

packages/gatsby-dev-cli/src/local-npm-registry/verdaccio.js
pieh: Right, I probably should use something like https://stackoverflow.com/questions/14031763/doing-a-cleanup-action-just-before-node-js-exits to do cleanup if it's in middle of publishing - dunno if it will work tho

packages/gatsby-cli/src/reporter/index.js
```diff
@@ -22,6 +23,9 @@ type ActivityArgs = {
  */
 
 module.exports = Object.assign(reporter, {
+  cat() {
+    console.log(showMeARandomASCIICat())
```
sidharthachatterjee: Can we please not remove this? 😛

pieh: using `cat-me` was a mistake 😂

packages/gatsby-dev-cli/src/utils/get-dependant-packages.js
```diff
@@ -0,0 +1,25 @@
...
+const getDependantPackages = ({
+  packageName,
+  depTree,
+  packagesToPublish = new Set(),
+}) => {
+  packagesToPublish.add(packageName)
+  const dependants = depTree[packageName]
+  if (dependants) {
+    dependants.forEach(dependant =>
+      getDependantPackages({
```
wardpeet: This can cause an infinite loop. You probably want something like this:
```js
const alreadySeen = packagesToPublish.has(packageName);
packagesToPublish.add(packageName)
  const dependants = depTree[packageName]
  if (dependants && !alreadySeen) {
```

packages/gatsby-dev-cli/src/utils/__tests__/traverse-package-deps.js
```diff
@@ -1,7 +1,9 @@
+const path = require(`path`)
+
 const { traversePackagesDeps } = require(`../traverse-package-deps`)
 
 jest.doMock(
```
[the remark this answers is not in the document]
pieh: ah, I was messing with mocks to allow to mock non-existing modules until I found `virtual` option, this is probably not needed (tho I don't know the difference between `mock` and `doMock` :P)

wardpeet:
@pieh mind removing the cat stuff so we can ship this? 😸

packages/gatsby-dev-cli/src/utils/check-deps-changes.js
```diff
@@ -0,0 +1,188 @@
...
+      }
+    }
+
+    // if package is not installed, we will do http GET request to
+    // unkpg to check if dependency in package published in public
+    // npm repository are different
+
+    // this allow us to not publish to local repository
+    // and save some time/work
+    try {
```
sidharthachatterjee: This is great!

packages/gatsby-dev-cli/src/utils/check-deps-changes.js
```diff
@@ -0,0 +1,188 @@
...
+    localPKGjson.dependencies
+  )
+
+  if (!areDepsEqual) {
+    const diff = difference(
+      monorepoPKGjson.dependencies,
+      localPKGjson.dependencies
+    )
+
+    const diff2 = difference(
```
sidharthachatterjee: This is awesome 😄

KyleAMathews:
💥awesome work @pieh! This has been a TODO nearly forever :-D Will be huge help

Issue #11526, opened 2019-02-03
Gatsby Source Wordpress crashes with Woocommerce Products


rotexhawk:
Gatsby Source Plugin Crashes with Woocommerce products because the following function tries to normalize product categories the same way it does for posts. The problem is that product categories are not ids but instead they are objects so `t => tags.find(tObj => t === tObj.wordpress_id).id` `t never equals tObj.wordpress.id`.  and we get `error, can't get id of undefined`

```javascript
exports.mapPostsToTagsCategories = entities => {
  const tags = entities.filter(e => e.__type === `wordpress__TAG`)
  const categories = entities.filter(e => e.__type === `wordpress__CATEGORY`)

  return entities.map(e => {

    let entityHasTags = e.tags && Array.isArray(e.tags) && e.tags.length
    if (tags.length && entityHasTags) {
      e.tags___NODE = e.tags.map(
       // Crashes because, posts.tags = [Int], while product.tags = [Obj]
        t => tags.find(tObj => t === tObj.wordpress_id).id
      )
      delete e.tags
    }

    let entityHasCategories =
      e.categories && Array.isArray(e.categories) && e.categories.length
    if (categories.length && entityHasCategories) {
      e.categories___NODE = e.categories.map(
      // Crashes because, posts.categories = [Int], while product.categories = [Obj]
        c => categories.find(cObj => c === cObj.wordpress_id).id
      )
      delete e.categories
    }

    return e
  })
}
```
**Product** 
```javascript
 {
        "id": 3161,
        "name": "Clear Turkish Teardrop Lamp",
        "slug": "clear-turkish-teardrop-lamp",
        "permalink": "https://www.example.com/product/clear-turkish-teardrop-lamp/",
        "description": "",
        "short_description": "",
        "sku": "P-A-L2H",
        "price": "181.245",
        "regular_price": "181.245",
        "sale_price": "",
        "categories": [
            {
                "id": 82,
                "name": "Wall Lamps",
                "slug": "wall-lamps"
            }
        ],
        }
    },
```

Post
```javascript
  {
        "id": 3322,
        "date": "2019-02-01T18:10:01",
        "date_gmt": "2019-02-01T18:10:01",
        "guid": {
            "rendered": "https://www.turkishlampwholesaler.com/?p=3322"
        },
        "modified": "2019-02-01T18:10:01",
        "modified_gmt": "2019-02-01T18:10:01",
        "slug": "test",
        "status": "publish",
        "type": "post",
[...]

PR #11527, opened 2019-02-03
fix(gatsby-source-wordpress): handle woocommerce categories and tags

rotexhawk:
Gatsby Source Plugin Crashes with Woocommerce products because `normalize.mapPostsToTagsCategories` function tries to normalize product categories the same way that it does for posts. The problem is that product categories are not ids but instead they are objects so t => tags.find(tObj => t === tObj.wordpress_id).id t never equals tObj.wordpress.id. and we get error, can't get id of undefined

Fixes: https://github.com/gatsbyjs/gatsby/issues/11526

All the tests passed locally and the build passed on a local site with woocommerce installed.


wardpeet:
@rotexhawk mind sharing an reproduction (private message is fine too) or write a test for it so we can validate

rotexhawk:
@wardpeet I don't know if this is what you meant, but I added a sample product and couple of `wc_categories` and updated the snapshot. Thanks.

pieh:
I'll try to verify this locally

pieh:
Ok, verified that this fixes the woocommerce issues.

I'm actually curious how you made it work @rotexhawk in more production ready env - I had serious problems with auth for WooCommerce and resorted to this hack:
```php
add_filter( 'woocommerce_rest_check_permissions', 'my_woocommerce_rest_check_permissions', 90, 4 );

function my_woocommerce_rest_check_permissions( $permission, $context, $object_id, $post_type  ){
  return true;
}
```
That essentially disable auth/permission check for WooCommerce endpoint, but obviosuly wouldn't advise anyone to use it (as they could edit your product or order data heh)

rotexhawk:
Thanks guys for taking care of this so quickly. @pieh first I used https://wordpress.org/plugins/jwt-authentication-for-wp-rest-api/ but it only worked partially. Some endpoints still returned 403. 
So I switched to basic auth and everything worked. https://github.com/WP-API/Basic-Auth

PR #11531, opened 2019-02-04
fix(gatsby-transformer-asciidoc): abstract the source of asciidoc transformer

jtanguy:
## Description

This PR changes the way asciidoc documents are loaded, via the other method `.load`, which takes the actual content retrieved by the relevant gatsby source plugin.

The current plugin relies on the `loadFile` method on, which does load
from the filesystem, breaking gatsby's source abstraction.

## Related Issues

None


pieh:
Thanks! Implementation for transformer isn't ideal because we do check for parent's node `extension` field ( https://github.com/gatsbyjs/gatsby/blob/64d4c1ea6b40340d33c304768da772b7bd5a81f6/packages/gatsby-transformer-asciidoc/src/gatsby-node.js#L15 ) - mostly because mediaType is not set for anything `asciidoc` related. We probably could add hacky check for not existing `text/asciidoc` mediaType but let's ship your changes first.

Issue #5836, opened 2018-06-11
Warn when creating multiple pages with the same path


pgegenfurtner:
Edit: see feature description at https://github.com/gatsbyjs/gatsby/issues/5836#issuecomment-396903475

## Description
I have two page templates which get filled with markdown content. On one of the templates I get an error during `gatsby build` that `props.data` is undefined. In develop it just works.

This is the query in the page template:

```
export const query = graphql`
    query LegalQuery($slug: String!) {
        markdownRemark(frontmatter: {slug: {eq: $slug}}) {
            id
            frontmatter {
              title
              slug
              headerImage
            }
            html
        }
    }
`;
```

Console error:
```
  30 |
  31 | function LegalPage(props) {
> 32 |   const html = props.data.markdownRemark.html;
     |                           ^
  33 |   const meta = props.data.markdownRemark.frontmatter;
  34 |
  35 |   return (


  WebpackError: Cannot read property 'markdownRemark' of undefined
```

### Environment
  System:
    OS: macOS High Sierra 10.13.5
    CPU: x64 Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 8.9.4 - ~/.nvm/versions/node/v8.9.4/bin/node
    Yarn: 1.3.2 - /usr/local/bin/yarn
    npm: 5.6.0 - ~/.nvm/versions/node/v8.9.4/bin/npm
  Browsers:
    Chrome: 67.0.3396.79
    Firefox: 59.0.2
    Safari: 11.1.1
  npmPackages:
    gatsby: ^1.9.270 => 1.9.270 
    gatsby-image: ^1.0.52 => 1.0.52 
    gatsby-link: ^1.6.44 => 1.6.44 
    gatsby-plugin-google-analytics: ^1.0.31 => 1.0.31 
    gatsby-plugin-netlify: ^1.0.21 => 1.0.21 
    gatsby-plugin-react-helmet: ^2.0.11 => 2.0.11 
    gatsby-plugin-styled-components: ^2.0.11 => 2.0.11 
    gatsby-plugin-typography: ^1.7.18 => 1.7.18 
    gatsby-source-filesystem: ^1.5.38 => 1.5.38 
    gatsby-transformer-excel: ^1.0.8 => 1.0.8 
    gatsby-transformer-json: ^1.0.19 => 1.0.19 
[...]

PR #8897, opened 2018-10-07
Warn when creating a page on an existing path

oorestisime:
Closes: #5836

I've spent a few hours trying to find out why a blog post was taking the contents of another one. Hope i ve put the warning in the best place.

PR #11532, opened 2019-02-04
Add warning when overriding page path

tsvetlin:
## Description
There was an already existing pull request for it: [https://github.com/gatsbyjs/gatsby/pull/8897](https://github.com/gatsbyjs/gatsby/pull/8897)

It reuses the same logic, but also updates the documentation.

## Related Issues
Closes: #5836


docs/tutorial/part-three/index.md
```diff
@@ -232,7 +232,7 @@ const ListLink = props => (
 // highlight-end
 
 export default ({ children }) => (
-  <div style={{ margin: `0 auto`, maxWidth: 650, padding: `1.25rem 1rem` }}>
+  <div style={{ margin: `3rem auto`, maxWidth: 650, padding: `0 1rem` }}>
```
pieh: Ops, I thought we branched out of master, apparently we did something wrong then, but let's remove this change from this PR

tsvetlin: I missed to switch the branch. Sorry. I'll fix that.

tsvetlin: Updated the pull request.

pieh:
We should also `deletePage` in this plugin https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-create-client-paths/src/gatsby-node.js#L32 (as otherwise it would trigger this new warning message)

tsvetlin:
> We should also `deletePage` in this plugin https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-create-client-paths/src/gatsby-node.js#L32 (as otherwise it would trigger this new warning message)

I'll fix this too. Thank you for pointing it out!

tsvetlin:
I fixed this too.

pieh:
I was thinking a bit about this and I think we would show a lot of warnings in `gatsby develop` (when we rerun `createPages` API hook). I don't have good solution yet, will need to think more about it

LekoArts:
Hey @pieh 👋 	Any updates on that? :)

sidharthachatterjee:
Thank you so much for opening this pull request @tsvetlin 👍 

This is definitely a valid issue (I've reopened the linked issue for now) but this change will result in too many false positives (most plugins update pages at the moment with multiple calls to `createPage` and _do not_ call `deletePage` before they do)

We think the false positives outweigh the benefit of this warning so let's close this till we figure out how to best handle this! If you like, we can continue this discussion on the issue at https://github.com/gatsbyjs/gatsby/issues/5836

Thank you so much nonetheless for opening this! We value your contributions and would love to see you around 💜 💪

PR #11538, opened 2019-02-04
chore(www): Increase inline code background visibility

### LINKED ISSUE #11537: Documentation code background
## Summary

When reading the docks, I saw that the code sections inline with other text have a different font and background. The background color is '#fdfaf6', it's close to white, thus hardly visible. I have trouble isolating it from the text. 

### Basic example

https://www.gatsbyjs.org/docs/how-to-contribute/#contributing-to-the-repo
At this link, the shell commands should be copied from the page. These code snippets are on the same line as the text, but since the code blocks background is barely visible, it's easy to miss a command.

Current background:
screenshot 2019-02-04 at 14 22 06

Darker background:
screenshot 2019-02-04 at 14 22 38

I think the background should be darker, so it would be more readable. for example '#ccc'

### Motivation

For me it's really hard to see which part of the text is code.  Maybe other people out there have also trouble with that. I'd happily change the css.
The background property should be changed of the code css class.

fk:
Before/after:

image

Fixes #11537 — thanks to @tsvetlin for reporting!

### DIFF
--- a/www/src/utils/colors.js
+++ b/www/src/utils/colors.js
@@ -36,7 +36,8 @@ const colors = {
     superLight: gray(96, 270),
   },
   code: {
-    bg: `#fdfaf6`, // colors.a[0] #fcf6f0
+    bgInline: `#fbf2e9`,
+    bg: `#fdfaf6`,
     border: `#faede5`,
     text: `#866c5b`,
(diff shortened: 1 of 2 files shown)


tsvetlin:
It would close #11537

Issue #8103, opened 2018-09-13
Content migration overview issue


shannonbux:
## Summary

Lots of content in our blog, plugin and starter READmes, and other external resources have valuable content we could pull into the docs.

1. If you'd like to write one of these docs, pick one from the list below
2. Then click on this URL to check what existing content can be repurposed in the article you're writing: https://airtable.com/shrU9erBeq1lhDy1Z
4. Then, open a PR with your article and mention this Issue #8103 in the PR description!

Note -- These docs may be in one of three states:
1. No doc exists yet (no corresponding file/page exists)
2. A stub article exists (the file is created and needs to be filled out)
3. A doc exists, but there is additional content we've identified that could possibly be used to make it better.

### Sourcing
- [x] Sourcing from Netlify CMS
- [x] Sourcing from Drupal* #8740
- [x] Sourcing from WordPress* #8742
- [x] Sourcing from Contentful* #8741 
- [x] Sourcing from the filesystem* #8730

### Deploying
- [x] Deploying to S3/Cloudfront
- [x] Deploying to Netlify
- [x] Deploying to Aerobatic #8736
- [x] Deploying to Heroku #8721 (PR)
- [x] Deploying to Now #8737
- [x] Deploying to GitLab Pages #8738

### Hosting
- [x] Deploying & Hosting (overview)
- [x] Hosting on Netlify #8844 

### Styling
- [x] CSS Modules*
- [x] Component CSS* #8726
- [x] Creating global styles
- [x] Typography.js*
- [x] PostCSS* #8839

### Building pages
~~Building a page with a GraphQL query~~ ([Duplicative. Already exists](https://www.gatsbyjs.org/docs/page-query/))
~~Rendering sidebar navigation dynamically~~ (New doc, not migration)
- [x] Creating and modifying pages
- [x] Creating slugs for pages* #8727
- [x] Layout components #8845
- [ ] Programmatically create pages from data* #8841 (PR 
under review)

### Performance + Optimization
- [x] using gatsby-image to prevent image bloat
- [x] Progressive web app (PWA)
- [x] Performance (overview)
- [x] Lighthouse audit*
- [x] Caching
- [x] Using Service Workers With Gatsby #8849 
- [ ] Optimizing plugins with Guess.js* #8846 (PR in progess)
~~Optimizations that Gatsby makes for you (**hold, needs clarification; missing in airtable**)~~

### Overviews
[...]

PR #11539, opened 2019-02-04
chore(docs): link stubs to pertinent issues

isaac-martin:
## Description

Update stubs to add issue links. Follow up to PR #11492 
There is also a lot of stubs that do not have there own issues, but are referenced in #8103 - Should I add that link to any that apply?

## Related Issues
#11342

### DIFF
--- a/docs/docs/build-caching.md
+++ b/docs/docs/build-caching.md
@@ -1,5 +1,6 @@
 ---
 title: Build Caching
+issue: https://github.com/gatsbyjs/gatsby/issues/8103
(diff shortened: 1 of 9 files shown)


lannonbr:
In reference to the stubs referenced in #8103, I think it would be good to reference them given some of the stubs are on hold for particular reasons so rather than seeing a stub and wanting to jump onto it, you can see if it is ready to be fleshed out or not at the moment.

isaac-martin:
Will update.

Issue #11508, opened 2019-02-02
TypeError: Cannot read property 'minify' of undefined at minify with terser plugin


podabed:
## Description

Production build doesn't build, but dev build is good

### Steps to reproduce

Just run gatsby build

P. S. This is after update
Part of yarn.lock before update while everything's works (good prod build without any errors)

```
terser@^3.8.1:
  version "3.14.1"
  resolved "https://registry.yarnpkg.com/terser/-/terser-3.14.1.tgz#cc4764014af570bc79c79742358bd46926018a32"
  integrity sha512-NSo3E99QDbYSMeJaEk9YW2lTg3qS9V0aKGlb+PlOrei1X02r1wSBHCNX/O+yeTRFSWPKPIGj6MqvvdqV4rnVGw==
  dependencies:
    commander "~2.17.1"
    source-map "~0.6.1"
    source-map-support "~0.5.6"
```

After update we have this (errors in "Environment" section below)

```
terser@^3.8.1:
  version "3.16.0"
  resolved "https://registry.yarnpkg.com/terser/-/terser-3.16.0.tgz#04028e6e5da461d91691cedd75fa53a17f2f20d9"
  integrity sha512-Ua8BhyibmsQBFXDZZ3Es7GASB2yFrQJr0jgAlZK1FBLbFarrHoCuMHPCro5MbX4jidcaFGiV+uTc3wxodmGjUg==
  dependencies:
    commander "~2.17.1"
    source-map "~0.6.1"
    source-map-support "~0.5.9"
```

### Expected result

Prod build

### Actual result

Prod build with error on js stage

### Environment

`gatsby info --clipboard`
``` ubuntu@ubuntu  ~/pet/podabed.org   sourceV2 ●✚  gatsby info --clipboard

  System:
    OS: Linux 4.4 Ubuntu 16.04.5 LTS (Xenial Xerus)
    CPU: (4) x64 AMD Athlon(tm) X4 860K Quad Core Processor
    Shell: 5.1.1 - /usr/bin/zsh
  Binaries:
    Node: 10.15.1 - ~/.nvm/versions/node/v10.15.1/bin/node
    Yarn: 1.13.0 - ~/.nvm/versions/node/v10.15.1/bin/yarn
    npm: 6.7.0 - ~/.nvm/versions/node/v10.15.1/bin/npm
  Languages:
    Python: 2.7.12 - /usr/bin/python
  Browsers:
    Chrome: 72.0.3626.81
[...]

PR #11542, opened 2019-02-04
fix(gatsby): update terser-webpack-plugin to avoid webpack error

sidharthachatterjee:
`terser-webpack-plugin` broke two days ago because a minor version bump to `terser` had a breaking change

This was fixed in https://github.com/terser-js/terser/commit/03b08760fcac9ee3f633d7c8e82af417c47db4e7 and released in 

This PR updates dependencies and lock files for starters so that no one accidentally installs the breaking version again

Related to https://github.com/gatsbyjs/gatsby/issues/11508

### DIFF
--- a/starters/blog/package-lock.json
+++ b/starters/blog/package-lock.json
@@ -891,7 +891,7 @@
     },
     "@types/debug": {
       "version": "0.0.29",
-      "resolved": "https://registry.npmjs.org/@types/debug/-/debug-0.0.29.tgz",
+      "resolved": "http://registry.npmjs.org/@types/debug/-/debug-0.0.29.tgz",
       "integrity": "sha1-oeUUrfvZLwOiJLpU1pMRHb8fN1Q="
     },
     "@types/events": {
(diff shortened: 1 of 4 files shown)


starters/blog/package-lock.json
```diff
@@ -891,7 +891,7 @@
     },
     "@types/debug": {
       "version": "0.0.29",
-      "resolved": "https://registry.npmjs.org/@types/debug/-/debug-0.0.29.tgz",
+      "resolved": "http://registry.npmjs.org/@types/debug/-/debug-0.0.29.tgz",
```
DSchau: Hmm.. wonder why this change?

sidharthachatterjee:
diff for starters is broken because of an npm issue 

https://npm.community/t/npm-install-downgrading-resolved-packages-from-https-to-http-registry-in-package-lock-json/1818/14

Recommendation is to manually update http back to https, let me do that

DSchau:
@sidharthachatterjee if you want the fix instead -> https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285/56

pieh:
Should we just bump `terser-webpack-plugin` here -> publish and let renovate handle bumps in starters?

sidharthachatterjee:
@pieh That seems like a better idea, reverting changes to starter locks

starters/blog/package-lock.json
sidharthachatterjee: Only updating gatsby dependency now and letting Renovate handle updates to starters

sidharthachatterjee:
We're good now, @pieh

DSchau:
Published as [`gatsby@2.0.112`](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/CHANGELOG.md#20112-2019-02-04)

PR #11545, opened 2019-02-04
fix(gatsby): correctly replace exported queries for StaticQuery

### LINKED ISSUE #11477: [bug]: StaticQuery fails when `export` is used to export query
## Description

When a component with a `StaticQuery` exports its query, e.g.

```jsx
import React from 'react'
import { StaticQuery, graphql } from 'gatsby'

export default function Header() {
  return (
    <StaticQuery
      query={titleQuery}
      render={data => (
        <header><h1>{data.site.siteMetadata.title}</h1></header>
      )}
    />
  )
}

export const titleQuery = graphql`
  {
    site {
      siteMetadata {
        title
      }
    }
  }
`

```

Building this code will lead to the query data _not_ being injected into the component and will lead to a display of `Loading (StaticQuery)`

### Steps to reproduce

- Use the above code, _or_ check out [this repo](https://github.com/marcysutton/gatsby-site/commit/7ae73eb46dbd008494b7610d7ae98ca6116ee014#diff-76a7ceef21a0a9f68cb5ca6205d6ac57R86)

### Expected result

Query should be injected and data should be available

### Actual result

Query is not stripped, data is not available

### Fix

Not sure what the _exact_ fix is, but it almost certainly some type of AST parsing issue in [this file](https://github.com/gatsbyjs/gatsby/blob/master/packages/babel-plugin-remove-graphql-queries/src/index.js)

DSchau:
## Description

This PR fixes the static query extraction (rather the import statement)
failing when the query is `export`ed. The issue was that the AST was
expected to be at a certain depth (e.g. the .parentPath stuff), and we
need to go up _one_ more level to get the correct element we expected,
i.e. the component.

I also added an e2e test to guard against this behavior in the future.

## Related Issues

Fixes #11477


packages/babel-plugin-remove-graphql-queries/src/index.js
```diff
@@ -76,12 +83,12 @@ function removeImport(tag) {
     else importPath.remove()
   }
   if (importPath.isObjectProperty()) {
-    if (parent.node.properties.length === 1)
-      importPath.findParent(p => p.isVariableDeclaration())?.remove()
```
DSchau: These were causing eslint to go wild. I _lightly_ refactored this and used a helper.

DSchau:
I used [this example](https://github.com/DSchau/gatsby-static-query-example) to validate this PR. Doesn't seem to cause any regressions as best as I can tell, just fixes the bug.

DSchau:
@KyleAMathews I think this is _one_ of them. The other is the one that @tkh44 @jlengstorf seem to be intermittently hitting, i.e. something like `cannot find relative path '../../public/static/d/asdfasdfs.json'`.

I think that's a monorepo/workspaces bug though, not necessarily something with StaticQuery. Still need to research it more though.

DSchau:
```
Successfully published:
 - babel-plugin-remove-graphql-queries@2.5.5
 - gatsby@2.0.114
```

Issue #11548, opened 2019-02-04
[docs]: move to async/await when appropriate


DSchau:
## Description

Now that the [Node 8 RFC has been merged](https://github.com/gatsbyjs/rfcs/pull/24) we can perform some changes to our documentation with functionality available in Node 8 and NPM 5.

Specifically, one of these is **using async/await in our documentation.**

## What should happen

All documentation examples, blog posts, tutorials, etc. should be migrated to using `async` and `await` **with a focus on simplifying and removing explicit use of a returned `Promise`**. For example, the following example is an excellent illustration of a code example that is not incredibly clear and is error-prone. Migrating to async/await drastically simplifies this

```js
exports.createPages = function createPages({ graphql }) {
  return new Promise((resolve, reject) => {
    graphql(`{
      site {
        siteMetadata {
          titel
        }
      }
    }`)
      .then(result => {
        if (result.errors) {
          return reject(result.errors)
        }

        // do something with result.data
        resolve()
      })
  })
}
```

with a fixed example like:

```js
exports.createPages = async function createPages({ graphql }) {
  const result = await graphql(`
    {
      site {
        siteMetadata {
          titel
        }
      }
    }
  `)
    .then(response => {
      if (response.errors) {
        throw new Error(response.errors)
      }
      return response
    })

  // do something with result | result.data
 
}
```

## What should _not_ happen

Migrating Promise based code in existing examples (e.g. a component with a componentDidMount lifecycle method) that is not substantially improved with async await should _not_ be added to this fix.

```jsx
import React, { Component } from 'react'

export default class IndexPage extends Component {
  componentDidMount() {
    fetch(`${process.env.API_URL}/comments`)
      .then(res => res.json())
      .then(comments => {
        this.setState({
          comments
        })
      })
  }
}
```

### Why?

We want to keep the barrier to entry low. Adding the `async` keyword does not improve comprehension in _every_ scenario. Particularly, sometimes it can be advantage to make it clear that something is _using_ a Promise in our documentation, and `async` and `await` could obfuscate this in certain scenarios.

## Examples

The [`adding-pagination`](https://github.com/gatsbyjs/gatsby/blob/master/docs/docs/adding-pagination.md#example) docs example is a great illustration of using this anti-pattern that can be cleaned up with `async` and `await`.

```diff
diff --git a/bad.js b/good.js
index 1238f0c..f1e931f 100644
--- a/bad.js
+++ b/good.js
@@ -1,53 +1,51 @@
[...]

Issue #11547, opened 2019-02-04
[docs]: move off of global installs and use package scripts


DSchau:
## Description

Now that the [Node 8 RFC has been merged](https://github.com/gatsbyjs/rfcs/pull/24) we can perform some changes to our documentation with functionality available in Node 8 and NPM 5.

Specifically, one of these is **moving off of global installs and migrating to `npx`.**

## What should happen

All instances of `npm install -g gatsby-cli` followed up by `gatsby <some-command>` should be replaced and unified with `npx gatsby <some-command>`.

Coupled with this issue is that any commands that are "bare" (e.g. `gatsby <some-command>`) should use the script name in `package.json`. For instance, given `gatsby develop`, that should be changed to `npm run develop`.

Both of these will isolate us from some issues re: CLI version and project version _and_ simplify set-up instructions.

This should be done in a single PR, if possible, but if multiple PRs are necessary--that's OK too!

## Examples

For instance, given the document [`docs/docs/authentication-tutorial.md`](https://github.com/gatsbyjs/gatsby/blob/master/docs/docs/authentication-tutorial.md#building-your-gatsby-app), we should upgrade like so:

```diff
- gatsby new gatsby-auth gatsbyjs/gatsby-starter-hello-world
- cd gatsby-auth
+ npx gatsby new gatsby-auth gatsbyjs/gatsby-starter-hello-world
```

This type of change should be done to any articles, tutorials, and even blog posts in the top-level `docs` directory.

isaac-martin:
@DSchau I'll start doing some of these tomorrow. Keep a look out for a PR, may not have time to get a PR for all of them

joshdcuneo:
@isaac-martin @DSchau sorry, should have mentioned I have been working on this. Admittedly it will be my first PR so no guarantees. 

Submitting the PR now, I think I got all the references except in the translations. There were a few places where explanations needed to change as well as the code and that was a bit beyond me in another language.

isaac-martin:
Thanks for the contribution @joshdcuneo

PR #11549, opened 2019-02-04
fix(docs): explicitly call out minimum node version (node 8)

DSchau:
## Description

Due to the [Node 8 RFC being merged](https://github.com/gatsbyjs/rfcs/pull/24), we can safely document that our minimum supported version is Node 8--even if technically it's Node 6 until we make an explicit breaking change in Gatsby 3.

This is done for a few reasons, specifically:

- Node 6 will be deprecated in a few months
- Generally, most of our users are going to be on Node 8
- Doing this as a gradual, technically non-breaking change offers a relatively seamless upgrade path
- Documenting "best practices" and a minimum supported version will encourage people to update to this minimum version, so again, we can ease the migration later

## Related Issues

Related to #11548 and #11547

### DIFF
--- a/docs/docs/gatsby-style-guide.md
+++ b/docs/docs/gatsby-style-guide.md
@@ -417,10 +417,6 @@ wording.
 
 When Gatsby commits to support a specific version of software (e.g. Node 8 and up), this is reflected in documentation. Gatsby documentation should be usable by all people on supported software, which means we don't introduce any commands or practices that can't be used by people on versions we've committed to support. In rare circumstances, we'll consider mentioning a newly introduced command or practice as side notes.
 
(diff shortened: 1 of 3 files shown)


docs/docs/gatsby-style-guide.md
```diff
@@ -417,10 +417,6 @@ wording.
 
 When Gatsby commits to support a specific version of software (e.g. Node 8 and up), this is reflected in documentation. Gatsby documentation should be usable by all people on supported software, which means we don't introduce any commands or practices that can't be used by people on versions we've committed to support. In rare circumstances, we'll consider mentioning a newly introduced command or practice as side notes.
 
-For example, npm 5.2.0 (which comes with Node 8) introduced a command called `npx` that is not available for versions of Node below 8. Since Gatsby recommends Node 8 and up but still technically supports Node 6 until the next major release, documentation should point out the requirements for `npx`:
```
DSchau: I just outright removed this. Thoughts?

PR #11554, opened 2019-02-05
feat(starters): add gatsby-starter-santa-fe

osogrizz:


osogrizz:
> Thanks for this. Looks great!

Thanks for that. I'm a total noob at this. Do I need to do anything else or does this mean my starter will be added to the starter library?

Issue #8080, opened 2018-09-12
[v2] `gatsby serve`bugs


jquense:
Running the latest v2 rc:

- ~~`ctrl +c` doesn't exit the server~~ this seems to be just me
- client paths aren't handled and result in a 404, might need the history fallback middleware
- keeps trying to send a non-existent 404 page on any client route'

maaaybe more, still checking

kakadiadarpan:
Hi @jquense, can you please provide more information on this? Also, I'm able to successfully exit the server using `ctrl + c`.

jquense:
@kakadiadarpan you're right it seems like something is funky in my env that it's not exiting. The other two issues are still relevant tho. I can hopefully throw a fix together for it

jquense:
Still relevant I believe

DSchau:
```
Successfully published:
 - gatsby@2.0.116
```

PR #11566, opened 2019-02-05
fix(gatsby-cli): use host option if it is passed

zauni:
## Description

The serve command didn't use the "--host" option and just used localhost.
It got deleted in commit https://github.com/gatsbyjs/gatsby/commit/bff0cd377359ad1432a1da042349793096cb8d37

## Related Issues

Related to #8080

### DIFF
--- a/packages/gatsby/src/commands/serve.js
+++ b/packages/gatsby/src/commands/serve.js
@@ -8,7 +8,7 @@ const preferDefault = require(`../bootstrap/prefer-default`)
 const chalk = require(`chalk`)
 
 module.exports = async program => {
-  let { prefixPaths, port, open } = program
+  let { prefixPaths, port, open, host } = program
   port = typeof port === `string` ? parseInt(port, 10) : port
 
   const config = await preferDefault(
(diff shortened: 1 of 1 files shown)


DSchau:
```
Successfully published:
 - gatsby@2.0.115
```

seidtgeist:
The default was listening to `0.0.0.0`, not `localhost`. Only the log output said localhost.

I am running `gatsby serve` in production behind a CDN, and heroku started complaining about the app not listening to the specified port, however, it was listening on the right port but on the wrong host.

According to the documentation, these methods are identical:

- [express app.listen()](http://expressjs.com/en/api.html#app.listen)
- [node core HTTP server.listen()](https://nodejs.org/api/http.html#http_server_listen)
- [node core net server.listen()](https://nodejs.org/api/net.html#net_server_listen_port_host_backlog_callback)

The default is this:

> If `host` is omitted, the server will accept connections on the unspecified IPv6 address (::) when IPv6 is available, or the unspecified IPv4 address (0.0.0.0) otherwise.

I would suggest changing the default back to what it was in order to avoid problems for people upgrading to this version. Happy to help if you agree.

PR #11580, opened 2019-02-05
fix(docs): add example for pageContext usage within a page

guayom:
## Description

This page explains very well that you can pass `context` to a page automatically created, but it's missing an example of how to use it in a page via `pageContext`. I think it is important to explicitly explain it because the name of the variable is different to the prop name.
I added an example.

## Related Issues
I didn't find any related issues.

### DIFF
--- a/docs/docs/creating-and-modifying-pages.md
+++ b/docs/docs/creating-and-modifying-pages.md
@@ -152,3 +152,19 @@ exports.onCreatePage = ({ page, actions }) => {
   })
 }
 ```
+
+On your pages and templates, you can access your context via the prop `pageContext` like this:
+
+```
+import React from 'react'
+
+const Page = ({ pageContext }) => {
+  return (
+    <div>
+      {pageContext.house}
+    </div>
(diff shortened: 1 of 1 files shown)


docs/docs/creating-and-modifying-pages.md
```diff
@@ -152,3 +152,19 @@ exports.onCreatePage = ({ page, actions }) => {
   })
 }
 ```
+
+On your pages and templates, you can access your context via the prop `pageContext` like this:
+
+```
```
DSchau:
```suggestion
```jsx
```

guayom:
:S Sorry, I didn't understand that part. Should I modify a file or run that command?

guayom:
Oh! Gotcha! I did the change from the UI, but now I see that the next time it's best to download the repo and run the linter before commiting. Thanks for your help and for working on this amazing software!

DSchau:
Now you helped too :) Thanks for _your_ help, we appreciate it!

PR #11581, opened 2019-02-05
fix(gatsby): use history fallback to display client-only routes in serve

### LINKED ISSUE #8080: [v2] `gatsby serve`bugs

DSchau:
## Description

As best as I can tell--this fixes a few issues identified with
client-only routes, without causing a regression in existing
functionality.

What I validated:

- Client only routes work (and work on initial navigation, e.g. given
dynamic path /app/1234, it works when navigated to and on initial route)
- Existing routes created via SSR work
- 404 routes work (and direct 404.html route)
  - implicit 404, e.g. `/asdfasdfasdf` loads 404 page
  - explicit 404, e.g. `/404` loads 404 page

cc @jquense 

## Related Issues

Fixes #8080

### DIFF
--- a/packages/gatsby/package.json
+++ b/packages/gatsby/package.json
@@ -56,6 +56,7 @@
     "eslint-plugin-react": "^7.8.2",
     "express": "^4.16.3",
     "express-graphql": "^0.6.12",
+    "express-history-api-fallback": "^2.2.1",
     "fast-levenshtein": "~2.0.4",
     "file-loader": "^1.1.11",
     "flat": "^4.0.0",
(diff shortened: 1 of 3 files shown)


DSchau:
Checking out the failing production runtime test!

DSchau:
The tests pass locally - must be some CI weirdness :/ I'll check this out in a bit, and try to fix the flaky gatsby-image test too.

PR #11588, opened 2019-02-06
feat(gatsby): add useStaticQuery hook

sidharthachatterjee:
Adds a new [React Hook](https://reactjs.org/docs/hooks-intro.html) called `useStaticQuery`

### Known Limitations
- Only a single instance of useStaticQuery per file is supported at the moment

### Todo
- [x]  Add typings
- [x]  Add hook implementation in browser entry
- [x]  Add support for variables in queries
- [x]  Add support for inlined queries
- [x]  Add production babel transform
- [x]  Add tests for production babel transform
- [x]  Add conditional to check that hook is imported from gatsby and improve call expression checks
- [x]  Add proper warning  if `useContext` is undefined
- [x]  Add warning  if query var is undefined?
- [x]  Remove import for useStaticQuery in babel transform
- [x]  Remove query variable in babel transform
- [x]  Add end to end tests for new hook
- [x]  Fix TypeScript typings

This is good to go now! Documentation and support for multiple instances will be added in later.


packages/babel-plugin-remove-graphql-queries/src/__tests__/__snapshots__/index.js.snap
```diff
@@ -54,6 +54,17 @@ export default (() => React.createElement(StaticQuery, {
 }));"
 `;
 
+exports[`Transforms queries defined in own variable in useStaticQuery 1`] = `
+"import staticQueryData from \\"public/static/d/2626356014.json\\";
+import React from 'react';
+import { useStaticQuery } from 'gatsby';
```
sidharthachatterjee: Should remove this `import` in the babel transform

packages/babel-plugin-remove-graphql-queries/src/__tests__/__snapshots__/index.js.snap
```diff
@@ -54,6 +54,17 @@ export default (() => React.createElement(StaticQuery, {
...
+export default (() => {
+  const query = \\"2626356014\\";
```
sidharthachatterjee: Need to remove query as well in babel transform in case of queries declared in their own variables

packages/babel-plugin-remove-graphql-queries/src/index.js
```diff
@@ -170,6 +170,43 @@ export default function({ types: t }) {
           },
         }
 
+        const nestedHookVisitor = {
+          CallExpression(path2) {
+            if (
+              [`production`, `test`].includes(process.env.NODE_ENV) &&
+              path2.node.callee.name === `useStaticQuery`
```
sidharthachatterjee: Check for import from `gatsby`

packages/gatsby/cache-dir/gatsby-browser-entry.js
```diff
@@ -29,6 +29,15 @@ const StaticQuery = props => (
   </StaticQueryContext.Consumer>
 )
 
+const useStaticQuery = query => {
```
sidharthachatterjee: Add warning if `useContext` is undefined

packages/gatsby/index.d.ts
```diff
@@ -20,4 +20,6 @@ export interface StaticQueryProps {
 
 export class StaticQuery extends React.Component<StaticQueryProps> {}
 
+export const useStaticQuery: (query: any) => void
```
sidharthachatterjee: Could probably improve typings here since `query` needs to be graphql tagged template

pieh: It also should return `Object` (definitely not `void`, as this would probably make TS complain when using result)?

packages/babel-plugin-remove-graphql-queries/src/__tests__/index.js
```diff
@@ -9,6 +9,37 @@ function matchesSnapshot(query) {
...
+    const query = graphql\`{site { siteMetadata { title }}}\`
+    const siteTitle = useStaticQuery(query)
+    
+    return (
+      <h1>{siteTitle.site.siteMetadata.title}</h1>
+    )
+  }
+  `)
+})
+
```
pieh: Can we maybe add "kitchen sink" example where `StaticQuery`, page query and `useStaticQuery` is used in single file? It currently wouldn't work in gatsby as-is, but here we just test transformation correctness and not if this will work later on with gatsby

packages/gatsby/index.d.ts
sidharthachatterjee: Ah, missed that! It can be `null` or obj though

packages/babel-plugin-remove-graphql-queries/src/__tests__/index.js
sidharthachatterjee: Yeah! There’s value in at least asserting that the transform is good to go. I’ll add one!

sidharthachatterjee:
I’ve currently set the default return for `useStaticQuery` to be `null`

Not sure what the pattern is here with hooks  yet but `null` would mean the user would need to do null checks themselves which the render prop component does for them. 

What do you think of this @pieh?

packages/babel-plugin-remove-graphql-queries/src/__tests__/index.js
```diff
@@ -9,6 +9,53 @@ function matchesSnapshot(query) {
   expect(code).toMatchSnapshot()
 }
 
+it(`Transforms queries in useStaticQuery`, () => {
+  matchesSnapshot(`
```
DSchau: Not related to this PR, but want to jot it down so we don't forget...

I don't actually think these tests are _all that_ valuable. The tests should validate something more meaningful, e.g. we should get the result of a transform, and then validate that e.g. an import statement appears and that import statement matches the hash of the data that's replaced.

It's harder to trace changes with these snapshots, and I don't think they're all that helpful.

packages/babel-plugin-remove-graphql-queries/src/__tests__/index.js
```diff
@@ -9,6 +9,53 @@ function matchesSnapshot(query) {
...
+  `)
+})
+
+it(`Transforms queries defined in own variable in useStaticQuery`, () => {
+  matchesSnapshot(`
+  import React from 'react'
+  import { graphql, useStaticQuery } from 'gatsby'
+
+  export default () => {
+    const query = graphql\`{site { siteMetadata { title }}}\`
```
DSchau: Generally these are used below/outside of the invoking function. Could we update this snapshot like so?

For example:

```jsx
import React from 'react'
import { graphql, useStaticQuery } from 'gatsby'

export default function Whatever() {
  const data = useStaticQuery(siteQuery)

  return <h1>{data.site.siteMetadata.title}</h1>
}

export const siteQuery = graphql`
  {
    site {
      siteMetadata {
        title
      }
    }
  }
`
```

DSchau: (this will also validate the `export` still works)

packages/gatsby/cache-dir/gatsby-browser-entry.js
```diff
@@ -1,4 +1,4 @@
-import React from "react"
+import React, { useContext } from "react"
```
DSchau: This isn't a functional change necessarily, but maybe just _use_ `React.useContext` instead? We can check for the existence of that rather easily, e.g.

```jsx
if (typeof React.useContext !== `function` && process.env.NODE_ENV === `development`) {
  // warn
}
```

packages/gatsby/cache-dir/gatsby-browser-entry.js
```diff
@@ -29,6 +29,15 @@ const StaticQuery = props => (
   </StaticQueryContext.Consumer>
 )
 
+const useStaticQuery = query => {
+  const context = useContext(StaticQueryContext)
+  if (context[query] && context[query].data) {
+    return context[query].data
+  } else {
```
DSchau: nit: else not required

packages/gatsby/index.d.ts
DSchau: @sidharthachatterjee good call! I think I'd just do

```ts
export const useStaticQuery: (query: any) => null | object
```

packages/gatsby/cache-dir/gatsby-browser-entry.js
DSchau: Also one note here... I think it's worth validating that these checks are stripped from our code in production (e.g. with our uglifier).

packages/babel-plugin-remove-graphql-queries/src/__tests__/__snapshots__/index.js.snap
```diff
@@ -67,6 +87,34 @@ export default (() => React.createElement(StaticQuery, {
...
+"import staticQueryData from \\"public/static/d/2626356014.json\\";
+import React from 'react';
+import { useStaticQuery } from 'gatsby';
+export default (() => {
+  const siteTitle = staticQueryData.data;
+  return React.createElement(\\"h1\\", null, siteTitle.site.siteMetadata.title);
+});"
+`;
+
+exports[`Works correctly with the kitchen sink 1`] = `
```
sidharthachatterjee: @pieh So this is _pretty_ broken but not sure because of _which_ transform

What should the correct result be here?

packages/gatsby/index.d.ts
sidharthachatterjee: Fixed the return here

I'll fix the input for both `StaticQuery` and `useStaticQuery` in a separate PR

It looks like they should be `TemplateStringsArray` but don't want to touch StaticQuery in this PR

packages/babel-plugin-remove-graphql-queries/src/__tests__/index.js
```diff
@@ -9,6 +9,30 @@ function matchesSnapshot(query) {
...
+  import React from 'react'
+  import { graphql, useStaticQuery, StaticQuery } from 'gatsby'
+
+  export default () => {
+    const query = graphql\`{site { siteMetadata { title }}}\`
+    const siteDescription = useStaticQuery(query)
+
+    return (
+      <StaticQuery
+        query={graphql\`{site { siteMetadata { title }}}\`}
```
DSchau: Maybe make this query different? If the content of the query is the hash, it's probably a good idea that we ensure the hash changes.

packages/gatsby/cache-dir/gatsby-browser-entry.js
pieh: This is tricky, because I would want this error to print out during `built-html` stage (when we write to static html files), but not actually end up in client js bundle (just micro optim to save few bytes)

packages/babel-plugin-remove-graphql-queries/src/__tests__/index.js
sidharthachatterjee: Oh, yeah. Missed that.

packages/babel-plugin-remove-graphql-queries/src/__tests__/__snapshots__/index.js.snap
DSchau: So a correct result (I think) would be something like (largely pseudo code)

```jsx
import staticQueryData1 from 'public/static/d/unique-hash-1.json'
import staticQueryData2 from 'public/static/d/unique-hash-2.json'
import React from 'react'
import { useStaticQuery, StaticQuery } from 'gatsby'

export default () => {
  const query = 'unique-hash-1'
  const siteDescription = staticQueryData1.data

  return (
    React.createElement(StaticQuery, { /* yadda */}, {
      data: staticQueryData2
    })
  )
}
```

packages/gatsby/cache-dir/gatsby-browser-entry.js
sidharthachatterjee: @pieh This will never end up in the bundle because I'll be stripping away the import and replacing it with static data

packages/babel-plugin-remove-graphql-queries/src/__tests__/index.js
DSchau: Also want to be clear here that we can isolate this PRs changes to _just_ useStaticQuery. I wouldn't necessarily say you have to make a ton of other improvements _unless_ they're necessary for the implementation.

I do like the kitchen sink test though (even if technically "broken"). One thing we can do is use `test.todo` introduced in Jest 24, and mark it as a todo for future implementation.

packages/gatsby/cache-dir/gatsby-browser-entry.js
sidharthachatterjee: Removed it but added it back now because 
- linter wants consistent returns
- generally more readable _imo_

DSchau: That should still be a consistent return though? Does ESLint really fail on something like this?

```js
const useStaticQuery = condition => {
  if (condition) {
    return { data: {} }
  }

  return null
}
```

That _is_ a consistent return! But - more a question, _not_ required for this PR. Keep it as is!

packages/gatsby/cache-dir/gatsby-browser-entry.js
sidharthachatterjee: All imports to useStaticQuery are now removed in build so this should be good

pieh: Ah, you are completely right - so this check would need to be done in babel plugin? (yuck - mixed concerns)

packages/babel-plugin-remove-graphql-queries/src/__tests__/__snapshots__/index.js.snap
sidharthachatterjee: @DSchau Any reason you can think of to keep these declarations?

packages/gatsby/cache-dir/gatsby-browser-entry.js
pieh: Actually, I guess it would need to be duplicated - in case there are 3rd party tools (Storybook etc) that want to use `gatsby-browser-entry` outside of `gatsby` "realm"?

packages/babel-plugin-remove-graphql-queries/src/__tests__/__snapshots__/index.js.snap
DSchau: Yeah - I was actually going to comment on this. Again--I'd say it's sort of outside the scope of this PR, but they seem... useless, right?

sidharthachatterjee: Definitely useless for both useStaticQuery as well as StaticQuery. I'll go ahead and remove in a later PR!

Definitely some cleanup to be done

packages/gatsby/cache-dir/gatsby-browser-entry.js
```diff
@@ -29,6 +29,24 @@ const StaticQuery = props => (
   </StaticQueryContext.Consumer>
 )
 
+const useStaticQuery = query => {
+  if (
+    typeof React.useContext !== `function` &&
+    process.env.NODE_ENV === `development`
+  ) {
+    throw new Error(
+      `You're likely using a version of React that doesn't support Hooks` +
```
sidharthachatterjee: @m-allanson Can you help me write a better error message here please?

packages/gatsby/cache-dir/gatsby-browser-entry.js
```diff
@@ -29,6 +29,24 @@ const StaticQuery = props => (
...
+        `Please update to React 16.8 or later. `
```
DSchau:
```suggestion
        `Please update to React 16.8.0 or later.`
```

packages/babel-plugin-remove-graphql-queries/src/__tests__/__snapshots__/index.js.snap
sidharthachatterjee: This is a micro optimization and done now

But it only works for instances of `import` and not `require` or object accessors on a top level require of gatsby

I think it's okay to mark this as resolved and fix later since it won't break anything but I'd love to hear what you have to say @pieh @DSchau

packages/babel-plugin-remove-graphql-queries/src/__tests__/index.js
sidharthachatterjee: Added a test for this!

sidharthachatterjee:
Okay, I was mistaken. Destructuring _just works_ @jlengstorf

DSchau:
@jlengstorf @sidharthachatterjee I used [this repo](https://github.com/DSchau/gatsby-static-query-example) to validate that

https://github.com/DSchau/gatsby-static-query-example/blob/65424c5b1b6604cba23ce29ac43f1a086083b6d7/src/components/seo.js

sidharthachatterjee:
TL; DR: Destructuring and any expression starting with `const` is fine!

To quickly elaborate on why I thought it wouldn't work — when parsing and transforming this AST, we _always_ create a [VariableDeclaration](https://babeljs.io/docs/en/babel-types#variabledeclaration) of type `const` with the original identifier (LHS) and assign it to the data object (RHS). 

`const data = useStaticQuery(query)`
`const { site } = useStaticQuery(query)`

These two are both _still_ instances of VariableDeclaration. So they work!

This however, will not — because it isn't a declaration!
`let data`
`data = useStaticQuery(query)`

One more edge case I can think of is that we _always_ rewrite to const so if someone wants to mutate down the line, that will throw. Let me see if I can fix this.

lannonbr:
@sidharthachatterjee I think the edge case you showed in the bottom of declaring it at one spot and then initializing somewhere else may be alright given the use case I do that usually is when I initialize a variable in a conditional and hooks can't be put in conditionals as of right now.

packages/babel-plugin-remove-graphql-queries/src/__tests__/index.js
sidharthachatterjee: `it.todo` is awesome! Done!

sidharthachatterjee:
@lannonbr Makes sense. 

I've fixed the transform to _preserve_ variable types nonetheless because this could be a common use case —
```
let data = useStaticQuery(query)
data = data.filter(...)
data = data.map(...)
```

sidharthachatterjee:
@pieh Maybe we want to do a minor bump for this?

DSchau:
@sidharthachatterjee there's really nothing breaking here--unless we've broken existing behavior (and I'll rectify that with an e2e test validating in a few).

So this is more of like an opt-in feature, because the developer will have to intentionally import the `useStaticQuery` helper (e.g. this isn't existing code that could break). The only negative side effect of this would be that they then will see a warning in their console to bump react/react-dom if they haven't already.

packages/gatsby/cache-dir/gatsby-browser-entry.js
sidharthachatterjee: Oh, gotcha! Done!

sidharthachatterjee:
@DSchau Agree. I figured we'd take this opportunity to do a minor bump because this is a new feature. I think @pieh had some of the changelog and version bumping fixed to follow semver _better_ so I figured _maybe_ we could when we release this.

pieh:
@sidharthachatterjee the version bump fixes (and other lerna/changelog related tasks) are not implemented in monorepo yet - we definitely could release this is minor manually tho if this is what folks want (just need little extra effort than normal publish)

e2e-tests/production-runtime/cypress/integration/static-query.js
```diff
@@ -0,0 +1,49 @@
...
+    it(`works with variable query`, () => {
+      cy.getTestElement(`use-static-query-variable`)
+        .invoke(`text`)
+        .should(`not.contain`, `Loading`)
+    })
+
+    it(`works with exported variable query`, () => {
+      cy.getTestElement(`use-static-query-exported`)
+        .invoke(`text`)
+        .should(`not.contain`, `Loading`)
```
DSchau: Silly assertion, but I don't really want to hard code a bunch of data for these tests either. Thoughts?

DSchau:
Added a bunch of e2e tests. This should serve as a baseline for tracking on regressions in the future _and_ it helps validate the contents of this PR in a more real-world environment than in a snapshot test with Jest.

sidharthachatterjee:
End to end tests looks awesome, @DSchau 

Thank you!

e2e-tests/production-runtime/cypress/integration/static-query.js
sidharthachatterjee: Yeah, I don't like the idea of hard coding data as well. However, this assertion will pass even if functionality breaks and `null` gets rendered for all data.

DSchau: So a breaking version of this is merely rendering a null?

DSchau: If so—I can just guard against that and output “broken” or something if data comes back null

sidharthachatterjee: It'll actually throw at the moment in `render` because we don't have a `null` check. But yeah, guarding against and asserting that it is not “broken“ seems like a good idea!

e2e-tests/development-runtime/src/components/static-query/use-static-query/hot.js
```diff
@@ -0,0 +1,22 @@
...
+    }
+  `)
+  return (
+    <p {...props}>
+      {data.site.siteMetadata.title}: {data.site.siteMetadata.author}
+    </p>
+  )
+}
+
+export default SupaHotFire
```
sidharthachatterjee: 😂

sidharthachatterjee:
- Added a warning for query variable not found (in case it is imported) like the one for StaticQuery 
- Removed useless query variable declaration in transform!

vadistic:
Hello, looks great!

I wanted jut to plug my one typescript related suggestion.

Please use generic to specify query type.

This is convention i.e in `apollo-client` (https://www.apollographql.com/docs/react/recipes/static-typing.html#typing-components) or `apollo-client-hooks` (https://github.com/trojanowski/react-apollo-hooks). Or even vanilla react hooks api.

Lot of us folks are using generators like `graphql-code-generator` or `apollo` to generate typings for their queries and generic is easy provide it.

I'm already missing it in StaticQuery Component and i'm retyping it in all of my projects

```ts
// gatsby.d.ts
  export type RenderCallback<Data> = (data: Data) => React.ReactNode

  export interface StaticQueryProps<Data> {
    query: any
    render?: RenderCallback<Data>
    children?: RenderCallback<Data>
  }

  export class StaticQuery<Data> extends React.Component<StaticQueryProps<Data>> {}
```

So I can do:
```tsx
const Example: React.FC = () => (
  <StaticQuery<MyQuery> query={MY_QUERY}>
    {(data) => (<p>{data.nicelyTypedProperty}</p>)}
    </StaticQuery>
)
```

This PR and new API is a great opportunity to do it right without breaking changes. 

So I suggest that instead of:

```ts
export const useStaticQuery: (query: any) => null | object
```

Do:

```ts
  export const useStaticQuery: <TData = object | null>(query: any) => TData

```

So as a user I can easily provide type for data using my codegen'ed typings:

```tsx
const Example: React.FC = () => {
  const data = useStaticQuery<MyQuery>(MY_QUERY)
  return (
  <p>{data.nicelyTypedProperty}</p>)
)}

```

btw. can it really return just null from a valid query?
not `{queriedField: null ??}` or Error - query is empty/  query field not found on schema?

Hope you'll take it under consideration, cheers :smile:

DSchau:
> btw. can it really return just null from a valid query?

No! That's what gets returned from an invalid/broken query apparently.

vadistic:
Thanks for a reply, I wasn't sure if maybe it's just a relay thing.

So, it's your decision is whether to provide null as fail safe for instance of broken, non-custom-typed static queries in typescript. 

I would suggest not typing for failed/error state (especially since it's compilation, not any useful run-time validation) - it seems misleading to me.

Ok, I'm here just to lobby for generics, please consider it :upside_down_face:

pieh:
> Ok, I'm here just to lobby for generics, please consider it 🙃

We are not really TS users and most of our definition are community contributions so we appreciate suggestion. Just one question - reading this sets default generic so by using your suggestion it would default to `object | null` type (so same behaviour as in PR if user doesn't provide generic)?

vadistic:
@pieh Yes, no difference. It's `generic parameter default`, works as you would guess from js.

It's the absolutely the most popular way to do it, here's similar line for `react-apollo`

https://github.com/apollographql/react-apollo/blob/master/src/Query.tsx#L99

```ts
                         // \/ exactly the same
export default class Query<TData = any, TVariables = OperationVariables> extends React.Component<
  QueryProps<TData, TVariables>
> 
... // hundered lines
    return data;
}
```

Also react typings use generic, see? :smile: 
```ts
// not providing default param, because S is inferred form initial state type
function useState<S>(initialState: S | (() => S)): [S, Dispatch<SetStateAction<S>>];

// but it's common pattern to use this generic to i.e override context type during initialisation
const MyContext = <MycontextType>React.createContext({} as any)

```

Btw. Not providing second generic field for variables will even suggest ts user that static query cannot accept variables :raised_hand_with_fingers_splayed: 

---

My second suggestion about this `null` type is to use this version
```ts
  export const useStaticQuery: <TData = object >(query: any) => TData
```

Because it forces to do null type check/ guards
```ts
const Example = () => {
const data = useStaticQuery(QUERY)

if(!data)
  // noop
  return null
}

// only now inferred type of data = object

...
return (...)
}
```

And since valid query should never return null, you would force users to add for completely unnecessary 4-5 lines of boilerplate for every (untyped) `useStaticQuery` instance.

packages/gatsby/cache-dir/gatsby-browser-entry.js
DSchau: I'm pretty OK with our error message here! Think it's OK.

Maybe I'd add to update React _and_ ReactDOM? It could also be useful _how_ to update both?

sidharthachatterjee: Added the point about ReactDOM! Instructions for update will have to take into account npm or yarn etc so leaving that out

pieh:
@vadistic We were doing some checks and yeah, generics are definitely way to go. Remaining question is default type for query results - it seems like using `object`  isn't ideal either:
```
error TS2339: Property 'foo' does not exist on type 'object'.

64     const test = data.foo
                         ~~~
```
this in practice forces users to either provide their type definition or use `any` - maybe we should just default to `any`?

pieh:
Current blocker:
There is potential timing issue in `gatsby develop` and we could render page before static query results are available in StaticQueryContext. This might be unlikely but definitely possible. For test I added timeout before emitting static query results in https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/utils/websocket-manager.js#L132-L138 which resulted in this:

kapture 2019-02-08 at 18 58 44
For `<StaticQuery>` component this is acceptable because we render Loading state, but we can't do the same for `useStaticQuery` hook. This can only happen on initial render (or if there are query errors, but then solution is to fix queries) and can cause clientside errors if null checks for query results aren't added. 

- [ ] @pieh to open separate issue to discuss potential solutions.

vadistic:
@pieh good point, skipped my mind :)

Query response will always be string indexed object extending scalar GraphQL types or arrays of those. We could provide index signatures for the first level but yeah, just go with `any`.

As in my link above - `react-apollo` does that (350k+ downloads weekly + it's typescript project), so I'm pretty confident it can be considered as standard practice.

packages/gatsby/src/internal-plugins/query-runner/query-compiler.js
```diff
@@ -217,6 +218,18 @@ class Runner {
...
       }
+
+      if (
+        query.isHook &&
+        process.env.NODE_ENV === `production` &&
+        typeof require(`react`).useContext !== `function`
+      ) {
+        report.panicOnBuild(
+          `You're likely using a version of React that doesn't support Hooks.
+Please update to React 16.8.0 or later to use the useStaticQuery hook.`
```
wardpeet: you probably want to put in the same error message as above?
https://github.com/gatsbyjs/gatsby/pull/11588/files#diff-4bc662f92e2fed3fcf4fa02501ad6d3dR39

sidharthachatterjee: Great catch! Fixed.

sidharthachatterjee:
Documentation for this is added in https://github.com/gatsbyjs/gatsby/pull/11741

PR #11593, opened 2019-02-06
chore(*): Upgrade babel-eslint

alexandernanberg:
## Description

Upgrade `babel-eslint` to v10

## Related Issues

N/A


alexandernanberg:
Ah sure, I understand! 

For this particular one it was mostly to get rid of the babel 7 beta dependencies

PR #11595, opened 2019-02-06
feat(gatsby-source-wordpress): Add yoast premium redirects in know entitites

maru3l:
## Description
Integrate Yoast redirects in know Entities and add documentation on how to implement Yoast with gatsby-source-wordpress


pieh:
@maru3l can You provide URL to example wordpress instance with premium yoast plugin installed/activated? Either publicly here or via DM on [discord](https://gatsby.app/discord) or twitter (@mipiechowiak) 

I have no way of testing this, because I don't have this premium plugin on any of sites I worked on.

maru3l:
@pieh This is my server in poduction. You can use it for your own test.

`https://wp-api.picki.ca/wp-json/`

pieh:
@maru3l The response for https://wp-api.picki.ca/wp-json/yoast/v1/redirects shows
```
[
  {
    "origin": "assurance-vie-sans-examen-medicale",
    "url": "assurance-vie-sans-examen-medical",
    "type": 301,
    "format": "plain"
  }
]
```

Which is an array, indicating that there could be more than one redirect. But proposed changes would only work for single redirect (because `id` is missing in that response, the plugin can't create multiple distinct nodes).

I think we would need to generate fake ids before we call https://github.com/gatsbyjs/gatsby/blob/66f6db07e67799ac02b93c1c40fff37dc1ec3af4/packages/gatsby-source-wordpress/src/gatsby-node.js#L95-L96
and don't edit `excludeUnknownEntities` function at all - for example:

in normalize.js
```js
exports.generateFakeWordpressId = entities => entities.map(
  e => {
    if (e.__type === `wordpress__yoast_redirects`) {
      // This doesn't have to be real ID - we just use `wordpress_id` as unique 
      // identifier within entity type. Redirect URLs should be unique
      // so we can use it
      e.wordpress_id = e.url
    }
    return e
  })
```

And then call this before `excludeUnknownEntities` so those entities are not discarded

maru3l:
@pieh Thanks for the idea. I've removed the exception in `excludeUnknownEntities` and add a `generateFakeWordpressId` in normalizers.

The only thing a made differant is the value you were suggestion for the ID.

I made a combination of the origin URL and the destination URL because we can have the scenario where someone remove a page and point the URL to a existing page who was already redirect to the same destination.

packages/gatsby-source-wordpress/README.md
```diff
@@ -600,6 +606,52 @@ Full example:
...
+}
+```
+
+### Query Yoast Redirects
+
+**_only work with Yoast Premium_**
+
+```graphql
+{
+  wordpressYoastRedirects {
```
pieh: I would suggest using

```graphql
{
  allWordpressYoastRedirects {
    edges {
      node {
        origin
        url
        type
        format
      }
    }
  }
}
```

to get all redirects and not just one

maru3l: Voilà 😁

sidharthachatterjee:
Published in gatsby-source-wordpress@3.0.52

PR #11605, opened 2019-02-06
fix(blog): update aws amplify blog post with latest info/screenshots

dabit3:
## Description
This PR updates the existing blog post to be current with the newer version of Amplify & to use the Amplify Console.

## Related Issues
n/a


dabit3:
@DSchau ok definitely will keep that in mind! Thanks for your help :)

PR #11610, opened 2019-02-06
fix(gatsby): add history fallback for client-only routes

DSchau:
## Description

Quick fix for an issue introduced in #11581 that is causing e2e failures in production_runtime.

The main issue is that the history fallback is not smart enough to understand the underlying page/routing structure of a Gatsby application and is over-eager in handling 404 routes. In other words--it consumes _meaningful_ 404 errors that we use in our e2e tests.

~The longer term fix that @pieh proposed is to tie into `.cache/pages.json` and intelligently match on client-only routes (using `matchPath`) which would check the `req.url` against the matchPath pattern. If it matches, it should serve the `index.html` route, otherwise it should fallback to 404 behavior.~

Implemented in f999abb

### DIFF
--- a/packages/gatsby/package.json
+++ b/packages/gatsby/package.json
@@ -56,7 +56,6 @@
     "eslint-plugin-react": "^7.8.2",
     "express": "^4.16.3",
     "express-graphql": "^0.6.12",
-    "express-history-api-fallback": "^2.2.1",
     "fast-levenshtein": "~2.0.4",
     "file-loader": "^1.1.11",
     "flat": "^4.0.0",
--- a/packages/gatsby/package.json
+++ b/packages/gatsby/package.json
@@ -86,6 +86,7 @@
     "lokijs": "^1.5.6",
     "md5": "^2.2.1",
     "md5-file": "^3.1.1",
+    "micromatch": "^3.1.10",
     "mime": "^2.2.0",
     "mini-css-extract-plugin": "^0.4.0",
     "mitt": "^1.1.2",
--- a/packages/gatsby/package.json
+++ b/packages/gatsby/package.json
@@ -86,7 +86,6 @@
     "lokijs": "^1.5.6",
     "md5": "^2.2.1",
     "md5-file": "^3.1.1",
-    "micromatch": "^3.1.10",
     "mime": "^2.2.0",
     "mini-css-extract-plugin": "^0.4.0",
     "mitt": "^1.1.2",
--- a/packages/gatsby/src/commands/serve.js
+++ b/packages/gatsby/src/commands/serve.js
@@ -8,7 +8,7 @@ const express = require(`express`)
 const getConfigFile = require(`../bootstrap/get-config-file`)
 const preferDefault = require(`../bootstrap/prefer-default`)
 const chalk = require(`chalk`)
-const mm = require(`micromatch`)
+const { match: reachMatch } = require(`@reach/router/lib/utils`)
 
 const getPages = directory =>
   fs
@@ -22,7 +22,7 @@ const historyRouter = (pages, options) => {
     .map(page => page.matchPath)
   return (req, res, next) => {
     const { url } = req
-    if (clientOnlyRoutes.some(route => mm.isMatch(url, route))) {
+    if (clientOnlyRoutes.some(route => reachMatch(route, url) !== null)) {
       if (req.accepts(`html`)) {
         return res.sendFile(`index.html`, options, err => {
           if (err) {
@@ -56,10 +56,9 @@ module.exports = async program => {
   router.use(historyRouter(pages, { root }))
   router.use((req, res, next) => {
     if (req.accepts(`html`)) {
-      res.status(404).sendFile(`404.html`, { root })
-    } else {
-      next()
+      return res.status(404).sendFile(`404.html`, { root })
     }
+    return next()
   })
   app.use(pathPrefix, router)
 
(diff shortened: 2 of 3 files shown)


DSchau:
Wanted to get this out before I head out for the day. f999abb contains a simple history mechanism where if the page `matchPath` matches the requested route, it will serve the index route.

If no matches are found--will fall through to existing 404 page (which matches existing functionality).

Marking this a WIP, because I haven't tested it a ton. Will do more tomorrow.

DSchau:
@jquense if you have some time--could you validate with a client-only path example of yours?

I reached out to a few people on Discord and don't have a ton of examples. I validated it works with some of our basic ones, but want to ensure it also handles the more "real world" examples.

packages/gatsby/src/commands/serve.js
```diff
@@ -1,13 +1,39 @@
...
+    .catch(() => [])
+
+const historyRouter = (pages, options) => {
+  const clientOnlyRoutes = pages
+    .filter(page => page.matchPath)
+    .map(page => page.matchPath)
+  return (req, res, next) => {
+    const { url } = req
+    if (clientOnlyRoutes.some(route => reachMatch(route, url) !== null)) {
+      if (req.accepts(`html`)) {
```
pieh: maybe we could move this check up so we don't even iterate over routes if it's not "html" request?

packages/gatsby/src/commands/serve.js
```diff
@@ -1,13 +1,39 @@
...
 const chalk = require(`chalk`)
+const { match: reachMatch } = require(`@reach/router/lib/utils`)
+
+const getPages = directory =>
+  fs
+    .readFile(path.join(directory, `.cache`, `pages.json`))
+    .then(contents => JSON.parse(contents))
+    .catch(() => [])
+
+const historyRouter = (pages, options) => {
```
pieh: NIT: maybe we could name it `clientOnlyPathsRouter`/`matchPathRouter` or something more descriptive to what it actually do - I find `historyRouter` name to not match what it actually does

DSchau: Agreed! It's the convention from the existing solution (e.g. https://www.npmjs.com/package/express-history-api-fallback) but yeah--this is nicer.

packages/gatsby/src/commands/serve.js
DSchau: Yup!

packages/gatsby/src/commands/serve.js
pieh: Yeah, I totally understand - I even started writing that I assume it comes from currently used package name (which I also find confusing :P)

packages/gatsby/package.json
```diff
@@ -56,7 +56,6 @@
     "eslint-plugin-react": "^7.8.2",
     "express": "^4.16.3",
     "express-graphql": "^0.6.12",
-    "express-history-api-fallback": "^2.2.1",
```
DSchau: bye felicia

pieh:
Totally not a blocker - do you think it's worth adding `serve` tests to `production-runtime` e2e tests? We have client-only-path there https://github.com/gatsbyjs/gatsby/blob/master/e2e-tests/production-runtime/src/pages/client-only-paths.js already so it potentially could be just matter of verifying:
```js
cy.visit(`/client-only-paths/param`).waitForAPI(`onRouteUpdate`)
cy.getTestElement(`dom-marker`).contains(`param`)

```

DSchau:
@pieh I think that's a great idea! I'll do this today!

jquense:
@DSchau this looks great! I'll try and validate on our complex app with client routes today

pieh:
Just circling back to my comment about e2e - I guess it wouldn't work because if we didn't handle client-only paths in serve, then we would first render 404 page and when app mounts we would re-render client-only-path, so this test would cause just false sense of correctness. Unless there's way to disable JS in cypress for single visit

packages/gatsby/src/commands/serve.js
```diff
@@ -1,13 +1,39 @@
...
+
+const clientOnlyPathsRouter = (pages, options) => {
+  const clientOnlyRoutes = pages
+    .filter(page => page.matchPath)
+    .map(page => page.matchPath)
+  return (req, res, next) => {
+    const { url } = req
+    if (req.accepts(`html`)) {
+      if (clientOnlyRoutes.some(route => reachMatch(route, url) !== null)) {
+        return res.sendFile(`index.html`, options, err => {
```
pieh: I think we need to serve potentially other files than `index.html`

if non-index page has client only paths:

```js
createPage({
  path: '/nested',
  matchPath: '/nested/*',
  ...restOfArgs
})
```
we would want to serve `nested/index.html` for `/nested/*` paths

wardpeet:
```
onst clientOnlyPathsRouter = (pages, options) => {
  console.log(pages)
  const clientOnlyRoutes = pages
    .filter(page => page.matchPath)
    .map(page => ({
      path: page.path,
      matchPath: page.matchPath,
    }))
  return (req, res, next) => {
    const { url } = req
    if (req.accepts(`html`)) {
      const route = clientOnlyRoutes.find(
        route => reachMatch(route.matchPath, url) !== null
      )
      if (route) {
        return res.sendFile(`.${route.path}index.html`, options, err => {
          if (err) {
            next()
          }
        })
      }
    }
    return next()
  }
}
```

something like this might work, path is `/` for default routes and `/nested/` for the nested once. I tried it with examples/simple-auth

PR #11612, opened 2019-02-07
fix(www): a11y tweaks

amberleyromo:
couple low-hanging fruit a11y tweaks:
- upped text contrast on mobile nav
- unique id attributes


DSchau:
@marcysutton is there any automated a11y checking we could do to catch regressions at PR time? Could be a nice CI check for www/.

marcysutton:
We could add axe-core/axe-webdriverjs to the test suite, it would be run as part of either unit or integration tests and could stop a build via CI. Here are some slides from a talk I did on it: https://marcysutton.github.io/a11y-and-ci/#/28

axe-puppeteer is out now too, in case you don't want to use Selenium Webdriver: https://github.com/dequelabs/axe-puppeteer

PR #11626, opened 2019-02-07
feat(gatsby-plugin-offline): Allow appending custom scripts to sw.js

davidbailey00:


packages/gatsby-plugin-offline/src/gatsby-node.js
```diff
@@ -125,8 +124,13 @@ exports.onPostBuild = (args, pluginOptions) => {
       const swAppend = fs
         .readFileSync(`${__dirname}/sw-append.js`, `utf8`)
         .replace(/%pathPrefix%/g, pathPrefix)
-
       fs.appendFileSync(`public/sw.js`, `\n` + swAppend)
+
+      if (pluginOptions.injectScript) {
```
sidharthachatterjee: I think this is a breaking change?

packages/gatsby-plugin-offline/README.md
```diff
@@ -19,13 +19,66 @@ in the service worker.
...
+      },
+      injectScript: `src/custom-sw-code.js`,
+    },
+  },
+]
+```
+
+## Upgrading from 2.0.x
+
+To upgrade from 2.0.x to 2.1.x, move any existing options into the `workboxConfig` option. If you haven't specified any options, you have nothing to do.
```
pieh: we would need to do major version bump for changing shape of plugin options (it's breaking change)

packages/gatsby-plugin-offline/README.md
```diff
@@ -19,13 +19,66 @@ in the service worker.
...
+## Available options
 
-When adding this plugin to your `gatsby-config.js`, you can pass in options to
-override the default [Workbox](https://developers.google.com/web/tools/workbox/modules/workbox-build) config.
+As of `gatsby-plugin-offline` 2.1.0, the following options are available:
 
-The default config is as follows. Warning: you can break the offline support by
-changing these options, so tread carefully.
+- `workboxConfig` allows you to override the default Workbox options - see [Overriding Workbox configuration](#overriding-workbox-configuration)
+- `injectScript` lets you specify a file to be injected into the end of the generated service worker (`sw.js`)
```
m-allanson: Could you add an example here? Something that demonstrates why you might use this option.

davidbailey00:
@pieh should be ready to merge now!

packages/gatsby-plugin-offline/src/gatsby-node.js
```diff
@@ -125,8 +124,13 @@ exports.onPostBuild = (args, pluginOptions) => {
...
+        const userAppend = fs.readFileSync(pluginOptions.injectScript, `utf8`)
+        fs.appendFileSync(`public/sw.js`, `\n` + userAppend)
```
sidharthachatterjee: What do you think of wrapping this in a try catch here and printing a friendly error if the file doesn't exist? Definitely not a blocker for this PR though.

packages/gatsby-plugin-offline/src/gatsby-node.js
```diff
@@ -105,11 +105,10 @@ exports.onPostBuild = (args, pluginOptions) => {
     clientsClaim: true,
   }
 
-  // pluginOptions.plugins is assigned automatically when the user hasn't
-  // specified custom options - Workbox throws an error with unsupported
-  // parameters, so delete it.
-  delete pluginOptions.plugins
```
sidharthachatterjee: Just noticed this. Do we have to handle it if pluginOptions.workboxConfig .plugins exists?

sidharthachatterjee:
Left a few comments but overall looks great, @davidbailey00! 👍

packages/gatsby-plugin-offline/src/gatsby-node.js
pieh: The `plugins` is injected by `gatsby` as a field on `pluginOptions`. If we are moving workbox build config to be subfield - the normalization of workboxConfig is no longer needed

packages/gatsby-plugin-offline/src/gatsby-node.js
davidbailey00: Agreed, I'll implement this change

davidbailey00: ok, just done this!

evanwinter: Hi there, sorry for intruding, but I'm wondering about the status of this (not super familiar with the code review interface on GH, but it appears to have passed review?). 

Anybody know if this will be implemented soon, or if it's still waiting on something? Or is there anything I can help with to move it along? 

I'm currently working on a project that would benefit greatly from being able to extend functionality of the plugin's service worker!

Thanks in advance 🙂

pieh:
I was just checking some docs - and was wondering if we could just document how to use [`importScripts`](https://developers.google.com/web/tools/workbox/modules/workbox-build#generateSW-importScripts) workbox config options instead of implementing our custom one?

m-allanson:
@davidbailey00 if it's possible to use workbox's `importScripts`, let's close this PR and create a separate issue.

KyleAMathews:
It was introduced in Workbox 4.3

KyleAMathews:
So we can't use it since we're on v3 still.

m-allanson:
Would it be worth updating to workbox 4 instead of doing this?

davidbailey00:
Hmm we could use the Workbox option - it wouldn't just be a one-stop operation though, since we would need to insert the content hash into the filename for any custom scripts this way and then modify the filenames in the config to match, so the first version is not cached forever. The approach taken in this PR instead inserts the custom code into the `sw.js` file itself. I think either way would work fine, so perhaps it would be best to upgrade Workbox and use the new option after all rather than an API breaking change?

pieh:
> Hmm we could use the Workbox option - it wouldn't just be a one-stop operation though, since we would need to insert the content hash into the filename for any custom scripts this way and then modify the filenames in the config to match, so the first version is not cached forever

Maybe we could do this on user behalf if `importScripts` was used (i.e. rename file to attach content hash and send renamed filename to workbox)? Not sure if this wouldn't be confusing tho. Or even just completely took over `importScripts` option and do what is currently done in this PR and strip that field from options passed to workbox?

m-allanson:
Moving this out of "In Progress" for now. We can circle back to it.

evanwinter:
Hey all, I'm sure this is not the place to ask this question, but I've had a hard time getting an answer.  I really need this feature for a project like, now - and I was going to take a crack at implementing what I've seen in this PR on a local copy of the plugin.  However I cannot for the life of me get a local copy to work -- it recognizes the local plugin but doesn't execute the code in it.  Is there anything unique about how Gatsby executes the code in this plugin?

I also tried publishing my local copy to NPM, so I could pull it in that way, but I get a babel error and am unable to figure out why that would be... 

Sorry again, just hoping to get past this roadblock and I figure those involved with this ticket would know the most.

packages/gatsby-plugin-offline/README.md
```diff
@@ -19,13 +19,89 @@ in the service worker.
...
-## Overriding options
+## Available options
 
-When adding this plugin to your `gatsby-config.js`, you can pass in options (via the `options` key) to
-override the default [Workbox](https://developers.google.com/web/tools/workbox/modules/workbox-build) config.
+As of `gatsby-plugin-offline` 3.0.0, the following options are available:
 
-The default config is as follows. Warning: You can break the offline support by
-changing these options, so tread carefully.
+- `appendScript` lets you specify a file to be injected into the end of the generated service worker (`sw.js`). For example:
```
sidharthachatterjee:
```suggestion
- `appendScript` lets you specify a file to be appended at the end of the generated service worker (`sw.js`). For example:
```

sidharthachatterjee:
Need to fix these 

```
The 'dontCacheBustUrlsMatching' option has been renamed to 'dontCacheBustURLsMatching'. Please update your config. 'dontCacheBustUrlsMatching' is now deprecated and will be removed in a future release of Workbox.
The 'modifyUrlPrefix' option has been renamed to 'modifyURLPrefix'. Please update your config. 'modifyUrlPrefix' is now deprecated and will be removed in a future release of Workbox.
Specifying 'cacheFirst'' in a 'runtimeCaching[].handler' option is deprecated. Please update your config to use 'CacheFirst' instead. In v4 Workbox strategies are now classes instead of functions.
Specifying 'networkFirst'' in a 'runtimeCaching[].handler' option is deprecated. Please update your config to use 'NetworkFirst' instead. In v4 Workbox strategies are now classes instead of functions.
Specifying 'staleWhileRevalidate'' in a 'runtimeCaching[].handler' option is deprecated. Please update your config to use 'StaleWhileRevalidate' instead. In v4 Workbox strategies are now classes instead of functions.
Specifying 'staleWhileRevalidate'' in a 'runtimeCaching[].handler' option is deprecated. Please update your config to use 'StaleWhileRevalidate' instead. In v4 Workbox strategies are now classes instead of functions.
```

sidharthachatterjee:
This is good to go and we will publish it on Monday!

sidharthachatterjee:
Running this on org and posting a build url here in a bit

sidharthachatterjee:
Looks like this works great on https://gatsby-org-offline3-test.netlify.com/

davidbailey00:
Published in `gatsby-plugin-offline@3.0.0` :tada:

PR #11630, opened 2019-02-07
 feat(gatsby-source-contentful): Expose JSON AST on Contentful Rich Text Node

sbezludny:
## Description

PR exposes JSON AST on Contentful Rich Text Node; added a deprecation note for `nodeType`.

### DIFF
--- a/packages/gatsby-source-contentful/src/extend-node-type.js
+++ b/packages/gatsby-source-contentful/src/extend-node-type.js
@@ -459,7 +459,7 @@ const fluidNodeType = ({ name, getTracedSVG }) => {
 }
 
 exports.extendNodeType = ({ type, store }) => {
-  if (type.name.endsWith(`RichTextNode`)) {
+  if (type.name.match(/contentful.*RichTextNode/)) {
     return {
       json: {
         type: GraphQLJSON,
--- a/packages/gatsby-source-contentful/src/extend-node-type.js
+++ b/packages/gatsby-source-contentful/src/extend-node-type.js
@@ -5,6 +5,7 @@ const {
   GraphQLString,
   GraphQLInt,
   GraphQLFloat,
+  GraphQLJSON,
 } = require(`gatsby/graphql`)
 const qs = require(`qs`)
 const base64Img = require(`base64-img`)
@@ -458,6 +459,18 @@ const fluidNodeType = ({ name, getTracedSVG }) => {
 }
 
 exports.extendNodeType = ({ type, store }) => {
+  if (type.name.endsWith(`RichTextNode`)) {
+    return {
+      json: {
+        type: GraphQLJSON,
+        resolve: (source, fieldArgs) => {
+          const contentJSON = JSON.parse(source.internal.content)
+          return contentJSON
+        },
+      },
+    }
+  }
+
   if (type.name !== `ContentfulAsset`) {
     return {}
   }
--- a/packages/gatsby-source-contentful/src/extend-node-type.js
+++ b/packages/gatsby-source-contentful/src/extend-node-type.js
@@ -461,6 +461,10 @@ const fluidNodeType = ({ name, getTracedSVG }) => {
 exports.extendNodeType = ({ type, store }) => {
   if (type.name.match(/contentful.*RichTextNode/)) {
     return {
+      nodeType: {
+        type: GraphQLString,
+        deprecationReason: `This field is deprecated, please use 'json' instead.`,
+      },
       json: {
         type: GraphQLJSON,
         resolve: (source, fieldArgs) => {
--- a/packages/gatsby-source-contentful/src/normalize.js
+++ b/packages/gatsby-source-contentful/src/normalize.js
@@ -193,9 +193,9 @@ function prepareTextNode(node, key, text, createNodeId) {
   return textNode
 }
 
-function prepareStructuredTextNode(node, key, content, createNodeId) {
+function prepareRichTextNode(node, key, content, createNodeId) {
   const str = stringify(content)
-  const structuredTextNode = {
+  const richTextNode = {
     ...content,
     id: createNodeId(`${node.id}${key}RichTextNode`),
     parent: node.id,
@@ -209,9 +209,9 @@ function prepareStructuredTextNode(node, key, content, createNodeId) {
     },
   }
 
-  node.children = node.children.concat([structuredTextNode.id])
+  node.children = node.children.concat([richTextNode.id])
 
-  return structuredTextNode
+  return richTextNode
 }
 function prepareJSONNode(node, key, content, createNodeId, i = ``) {
   const str = JSON.stringify(content)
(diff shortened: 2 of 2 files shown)


sbezludny:
Thanks for the feedback @zcei

Regarding breaking changes:
1. you're right, and I will mark commit as breaking because we're no longer spreading properties `(...content)`
2. I believe accessing `internal.type` is still possible.

> Also: do you want to expose the JSON AST or the String representation of the JSON AST?

The intention was to expose JSON, **not**  String representation of the JSON AST

zcei:
> The intention was to expose JSON, **not** String representation of the JSON AST

Any reason then to stringify it to put it as a string onto the `internal.content` only to then parse it to an object in the resolver? (Maybe it's some Gatsby convention I'm not familiar with?)

packages/gatsby-source-contentful/src/normalize.js
```diff
@@ -193,14 +193,12 @@ function prepareTextNode(node, key, text, createNodeId) {
   return textNode
 }
 
-function prepareStructuredTextNode(node, key, content, createNodeId) {
+function prepareRichTextNode(node, key, content, createNodeId) {
   const str = stringify(content)
-  const structuredTextNode = {
-    ...content,
```
Khaledgarbaya: I would suggest to first deprecate this field first and name the other one ast. WDYT?
That's the beauty of GraphQL ❤️

mgmolisani:
@sbezludny I was just scanning Gatsby PRs and came across this. Is this based off my work in [this gist](
https://gist.github.com/mgmolisani/aafbf9d0ce6ee341df1046fa1fae1cb5#file-gatsby-node-richtextast-js) which I posted on an issue for the new React renderer we are working on for Contentful rich text? I assume it is since we have been working on this closely with you so I just have a few questions on this.

1. Why only check the tail end of the node type.name? Wouldn’t it be better to check the beginning and end to make sure it is a Contentful node and not some other clashing *RichTextNode? My understanding was this function gets called for all nodes and not just Contentful nodes but that may be wrong.
2. Not caching the value? Seems like a really simple improvement.

Do you mind if I make these changes and get some credit on this PR? I was planning to do this anyway but just waiting on the rich text React renderer to get reviewed and merged.

sbezludny:
> Any reason then to stringify it to put it as a string onto the `internal.content`

@zcei it was implemented this way to have the consistency with JSON field. // cc: @Khaledgarbaya

sbezludny:
@mgmolisani 
thanks for your awesome feedback

> 1. Why only check the tail end of the node type.name?

We've incorporated your suggestion and condition is more sophisticated:

https://github.com/gatsbyjs/gatsby/pull/11630/files#diff-dc42475a9f0ed45d6e1bf554ec9958a2R463

> 2\. Not caching the value?

I'd suggest creating a separate PR for that.

pieh:
Just commenting here, that I didn't forgot about this - last week I was on issues duty so I wasn't really handling any PRs. This is on my TO-DO list for Monday

pieh:
There seems to be some github hiccup and I can't do inline comments, so will try to use top level comment:
 - the `data` field in `extendNodeType` - where this is coming from? I've setup contentful space with rich text field, but I'm not getting this field (using current master version - without changes from this PR) - I would assume this is from the name of actual RichText field? I do have other fields (`content` and `nodeType`)

There is also removal of `[key]: str` which I think wasn't particularly useful but is breaking change.

This does clean up stuff up a lot with usage of `GraphQLJSON` so I really want this in. I think we do need to bump major version with these changes, because they are breaking. This is not a blocker - just a note.

pieh:
example of query that worked with master, but broke after using changes:

screenshot 2019-02-18 at 19 59 08

packages/gatsby-source-contentful/src/extend-node-type.js
```diff
@@ -458,6 +460,30 @@ const fluidNodeType = ({ name, getTracedSVG }) => {
...
+    return {
+      data: {
+        type: GraphQLJSON,
+        deprecationReason: `This field is deprecated, please use 'json' instead.`,
+      },
+      nodeType: {
+        type: GraphQLString,
+        deprecationReason: `This field is deprecated, please use 'json' instead.`,
+      },
+      content: {
```
Khaledgarbaya: @sbezludny deprecating the content field broke existing queries mainly because `type: new GraphQLList(GraphQLJSON)` I think.
check @pieh message

pieh: It's a shame that we can't "just" add `deprecationReason` (right now) and keep inferred type. So I think we need to drop deprecation notice here

sbezludny:
> * the `data` field in `extendNodeType` - where this is coming from? * I've setup contentful space with rich text field, but I'm not getting this field

The data is part of the `content` object:

see: 
https://github.com/gatsbyjs/gatsby/pull/11630/files#diff-0f3299a5d7eb6ef91bdf3a22e5059e0fR199

but is not queryable because it is always an empty object.

packages/gatsby-source-contentful/src/extend-node-type.js
sbezludny: thanks for the suggestion @pieh I'll update the PR.

pieh:
> The data is part of the `content` object:

Interesting - what do I need to make it not empty then? I would assume I need to do something specific in my contentful space? In any case setting it as `GraphQLJSON` would also be breaking change then (similar to `content`)

sbezludny:
> Interesting - what do I need to make it not empty then?

ATM, there is no way to have anything there, it exists for consistency reasons -- every node in the Rich Text Document has `.data` property to hold arbitrary data associated with it.

packages/gatsby-source-contentful/src/normalize.js
```diff
@@ -193,14 +193,13 @@ function prepareTextNode(node, key, text, createNodeId) {
...
-function prepareStructuredTextNode(node, key, content, createNodeId) {
+function prepareRichTextNode(node, key, content, createNodeId) {
   const str = stringify(content)
-  const structuredTextNode = {
+  const richTextNode = {
     ...content,
     id: createNodeId(`${node.id}${key}RichTextNode`),
     parent: node.id,
     children: [],
-    [key]: str,
```
pieh: Could this be potentially considered as breaking change? Is there harm it keeping (just for backward compat if someone for some reason used it). If `key` is the same as field defined in `extendNodeType` then `extendNodeType` will win, so there shouldn't be harm in keeping it

sbezludny: You're right, this is a breaking change; however, we can leave it outside of the scope of this PR and introduce this breaking change later. Would you recommend reverting this change?

sbezludny:
Hey @pieh, could you take another look?

packages/gatsby-source-contentful/src/normalize.js
pieh: Sorry for late reply - if leaving this in doesn't mess anything up, then I would recommend it. If it's possible to avoid breaking change, we should always do it so users don't need to do extra work migrate their code.

pieh:
Checking again after changes, but I think this will be good to go. And will remove `breaking change` label

PR #11635, opened 2019-02-07
fix accessibility of linked headers plugin

marcysutton:
## Description

The plugin for auto-linking Markdown headers is super slick, but it had some accessibility issues. Specifically:

- With `icon` being truthy, the plugin introduces an anchor tag wrapping an SVG icon inside of the heading.
- The link receives `aria-hidden="true"` but isn't disabled from the keyboard with `tabindex="-1"`, so it effectively becomes a "ghost control" in screen readers since the accessibility information is stripped out.

Instead of hiding them, this PR makes those links keyboard and screen reader accessible so they can be used by more people.

- Added an `aria-label` to the link using a version of the header slug with spaces instead of dashes and the word "permalink" at the end making it clear what these links are for. 
- Added `focusable: false` to the default SVG so it won't receive focus in some assistive technologies (since it's not supposed to be interactive).

## To-do

I could use some help with the tests, if we want some test coverage for this. There wasn't any precedent for testing the links, only header IDs and icon children.

### DIFF
--- a/packages/gatsby-remark-autolink-headers/src/index.js
+++ b/packages/gatsby-remark-autolink-headers/src/index.js
@@ -10,7 +10,7 @@ function patch(context, key, value) {
   return context[key]
 }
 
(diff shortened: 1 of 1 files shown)


DSchau:
```
Successfully published:
 - gatsby-remark-autolink-headers@2.0.14
```

KyleAMathews:
Nice!

PR #11640, opened 2019-02-08
fix(gatsby): pin webpack to fix unexpected token error during build

paulmelnikow:
## Description

This addresses #11198.

There's a regression bug webpack/webpack#8656 affecting 4.29.0 through at least 4.29.3 which causes errors like this during `gatsby build`.

  Error: ./.cache/async-requires.js 8:11
  Module parse failed: Unexpected token (8:11)
  You may need an appropriate loader to handle this file type.
  | exports.components = {
  |   "component---frontend-components-main-js": function componentFrontendCompo  nentsMainJs() {
  >     return import("/home/circleci/project/frontend/components/main.js"
  |     /* webpackChunkName: "component---frontend-components-main-js" */
  |     );
   @ ./.cache/production-app.js 18:0-45 21:23-36 26:23-36

## Related Issues

- Addresses #11198
- Related to webpack/webpack#8656

### DIFF
--- a/packages/gatsby/package.json
+++ b/packages/gatsby/package.json
@@ -120,7 +120,7 @@
     "url-loader": "^1.0.1",
     "uuid": "^3.1.0",
     "v8-compile-cache": "^1.1.0",
-    "webpack": "^4.12.0",
+    "webpack": "~4.28.2",
     "webpack-dev-middleware": "^3.0.1",
     "webpack-dev-server": "^3.1.14",
     "webpack-hot-middleware": "^2.21.0",
--- a/packages/gatsby/package.json
+++ b/packages/gatsby/package.json
@@ -120,7 +120,7 @@
     "url-loader": "^1.0.1",
     "uuid": "^3.1.0",
     "v8-compile-cache": "^1.1.0",
-    "webpack": "~4.28.2",
+    "webpack": "~4.28.4",
     "webpack-dev-middleware": "^3.0.1",
     "webpack-dev-server": "^3.1.14",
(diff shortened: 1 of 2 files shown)


packages/gatsby/package.json
```diff
@@ -120,7 +120,7 @@
     "url-loader": "^1.0.1",
     "uuid": "^3.1.0",
     "v8-compile-cache": "^1.1.0",
-    "webpack": "^4.12.0",
+    "webpack": "~4.28.2",
```
DSchau: The linked issue suggests `4.28.4`? https://github.com/webpack/webpack/issues/8656#issuecomment-456153381

```suggestion
    "webpack": "~4.28.4",
```

Is this not accurate?

paulmelnikow: Ah yes, I think I snagged from 4.28.2 from the top post. Since this is using a tilde match, the specific patch version doesn’t matter, though may as well use the latest as of today (which is 4.28.4). You can see in the lockfile that’s what’s being pulled in.

DSchau: The lockfile is what _we_ pull in for our monorepo, not really what would get exposed to developers using gatsby! But yeah, `~4.28.2` should resolve to `4.28.4` so I think we're fine here.

paulmelnikow: Yea, I understand.

I feel like pinning the patch version is short-sighted. The tilde match seeeems like it’ll be okay.

DSchau:
```
Successfully published:
 - gatsby@2.0.118
```

PR #11642, opened 2019-02-08
docs(starters): remove imigliori.vivigratis.com (project cancelled)

valse:


valse:
Yes it was mine and the project has been cancelled... Thanks

Issue #11574, opened 2019-02-05
chore(www): Gatsby Showcase site categories sticking out from the page.


tsvetlin:
## Description

When I tried to scroll down to the bottom, to check out all the sections, it is not working properly. The sidebar with the categories is cut down, thus many categories on the bottom of the screen are unaccesible this way. If I resize my window to be smaller, even more categories are missing. 

### Steps to reproduce

If you haven't selected a category yet, you can't scroll down to the end, if you select one, you can scroll down to the end..

If you resize the window to be smaller, when no category is selected, the categories stick out of the screen, and will be inaccessible. You can see the scrollbar is at the end of the page.

Inspector on the right shows that many more elements are out of the screen.

![screenshot 2019-02-05 at 18 19 29 2](https://user-images.githubusercontent.com/15650896/52291771-6f7dbd00-2973-11e9-8bba-cc23d3ad0e1b.png)

### Expected result

What should happen?
I'd like to be able to scroll to the bottom of the categories, to check them out and filter.

![screenshot 2019-02-05 at 18 27 13 2](https://user-images.githubusercontent.com/15650896/52291915-c8e5ec00-2973-11e9-9c76-8cf0c8053896.png)

### Actual result

What happened?
Bunch of categories are missing on the bottom of the page.

tsvetlin:
I'd like to fix this issue.

DSchau:
I can't really replicate this--am I missing something? I'm using Chrome (latest), are you on a different version? Did I not follow the reproduction steps correctly?

[Link to video](https://gfycat.com/ShockingHarshClumber)

tsvetlin:
I had Chrome Version 72.0.3626.81 (Official Build) (64-bit), it was buggy there (no scrollbar, and content was out of the screen), what my screenshots indicate.

On the newest version of Chrome Version 72.0.3626.91 (Official Build) (64-bit) it seems to be working as expected.

But on Mozilla its still not working as designed. There is no scrollbar inside the Categories. My PR also fixes that.

Moreover, theres a className which value is undefined, it fixes also that.

![image](https://user-images.githubusercontent.com/15650896/52500750-0601e680-2bdf-11e9-84e0-23ca89c3c1df.png)

tsvetlin:
Maybe it was a browser issue and Firefox will release a new version and it will be fine.

wardpeet:
Fixed by #12449

PR #11644, opened 2019-02-08
fix(www): Showcase categories css fix.

tsvetlin:
## Description

I found an issue, on the Showcase site, categories were invisible on some browsers. 
Details of the issue are in #11574

## Related Issues

 Link to the issue that is fixed by this PR (if there is one)
Fixes: #11574


DSchau:
Thanks for this!

@tsvetlin I commented on the original issue--could you provide a clearer reproduction? Unsure how to reproduce the original issue, so unsure if we _need_ the fix! Thanks!

wardpeet:
Thanks so much for the PR!

It seems like #12449 fixes this PR more correctly and is indeed a firefox issue.

PR #11646, opened 2019-02-08
Improve handling of internal links with anchor

### LINKED ISSUE #11645: feat(gatsby-link): Improve handling of internal links with anchor
Currently the `Link` component of the `gatsby-link` package will check if the given target URL is internal and will print a warning when in development environment/mode, but the used regex doesn't handle anchors correctly resulting in (multiple) warnings being printed on each page change.

To prevent this console logging noise the regex can be extended to include the `#` character to handle anchors for valid internal links:

```diff
-/^\/(?!\/)/
+/^[/#](?!\/)/
```

###### Example

```js
import React from "react"
import { Link } from "gatsby"

const MyLink = props => <Link to="/#my-anchor" {...props}>My Anchor</Link>;

export default MyLink;
```

Given the example code above a warning message will be printed to the console:

```raw
External link ${to} was detected in a Link component. Use the Link component only for internal links. See: https://gatsby.app/internal-links
```

Using the improved regex will allow such links since they are valid internal links and a common use case.

### Motivation

Internal anchor links are a common use case (especially for Markdown/MDX).
Even when the `Link` component works as expected when used with an anchor link before such an improvement, the large amount of warnings in the console (in development mode) are noisy and could unsettle beginners/newcomers and might lead them to the decision that their code contains bugs/problems.

arcticicestudio:
> Resolves #11645

### DIFF
--- a/packages/gatsby-link/src/__tests__/index.js
+++ b/packages/gatsby-link/src/__tests__/index.js
@@ -105,6 +105,13 @@ describe(`<Link />`, () => {
       expect(console.warn).not.toBeCalled()
     })
 
+    it(`does not warn when internal with anchor`, () => {
+      jest.spyOn(global.console, `warn`)
+      const to = `/#local`
+      setup({ linkProps: { to } })
+      expect(console.warn).not.toBeCalled()
+    })
+
     it(`warns when not internal`, () => {
       jest.spyOn(global.console, `warn`)
       const to = `https://gatsby.org`
--- a/packages/gatsby-link/src/index.js
+++ b/packages/gatsby-link/src/index.js
@@ -106,7 +106,7 @@ class GatsbyLink extends React.Component {
       ...rest
     } = this.props
 
-    const LOCAL_URL = /^\/(?!\/)/
+    const LOCAL_URL = /^[/#](?!\/)/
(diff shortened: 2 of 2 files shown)


arcticicestudio:
> With this solution, this would not throw, and functionality like an `IntersectionObserver`, low priority fetching (for a link/page that doesn't exist), etc. would be enabled--and I think we want to avoid this.

This is indeed an important point I haven't thought about!
Extending the warning message is the better solution in this case. My thoughts were wrapped too much into a custom wrapper component from one of my projects where I tried to create all-in-one solution for internal and external links instead of simply using `<a>` :upside_down_face: 

I'll update the PR to tweak the warning message instead and will also extend the docs to include some information about how to handle anchor links.

DSchau:
@arcticicestudio you're the best; thank you for being open to addressing that!

DSchau:
@arcticicestudio let's get this closed out, as it won't be merged in its current state.

If you have a few minutes later this week--would appreciate that PR addressing this in the docs! Not necessarily _as_ concerned with the code change, but if you want to include something that will tweak the error, that could be OK too!

Thanks for this PR! Please feel free to reply back if any of us can help with anything further!

PR #11654, opened 2019-02-08
fix(blog): remove all date strings

DSchau:
## Description

Quick fix to standardize on date format strings, which appears to have been broken with #10924


www/package.json
```diff
@@ -48,7 +48,7 @@
     "gatsby-source-npm-package-search": "^2.0.0",
     "gatsby-transformer-csv": "^2.0.0",
     "gatsby-transformer-documentationjs": "^2.0.0",
-    "gatsby-transformer-remark": "^2.1.8",
```
DSchau: I could have also just pinned this--but may as well standardize while we're at it.

Issue #11368, opened 2019-01-28
Issues with sort fields


stefanprobst:
Some issues with sort fields:

* fields linking to other nodes with the `___NODE` naming convention are included as is in the sort fields enum. This means (i) we cannot sort by fields on the linked node; (ii) we also cannot sort by the foreign-key value on the ___NODE field, because we [replace triple underscores with dots](https://github.com/gatsbyjs/gatsby/blob/ff3f14929348015bc5f6967427c61f9974b5b169/packages/gatsby/src/redux/run-sift.js#L255). We have to do this because we unnecessarily store the underscored field names as enum values. ([here](https://github.com/gatsbyjs/gatsby/blob/ff3f14929348015bc5f6967427c61f9974b5b169/packages/gatsby/src/schema/data-tree-utils.js#L347) and [here](https://github.com/gatsbyjs/gatsby/blob/ff3f14929348015bc5f6967427c61f9974b5b169/packages/gatsby/src/schema/infer-graphql-input-fields-from-fields.js#L193))
* sorting on arrays of objects is not possible. when collecting sort fields we [assume that arrays are arrays of scalars](https://github.com/gatsbyjs/gatsby/blob/ff3f14929348015bc5f6967427c61f9974b5b169/packages/gatsby/src/schema/infer-graphql-input-fields-from-fields.js#L188). See also [this test](https://github.com/gatsbyjs/gatsby/blob/ff3f14929348015bc5f6967427c61f9974b5b169/packages/gatsby/src/schema/__tests__/infer-graphql-input-type-from-fields-test.js#L377) for an example why this is unexpected (we cannot sort on objects without also specifying a property).
* <s>although we include fields added with the `setFieldsOnGraphQLNodeType` API in the sort fields enum, we cannot sort on those fields, because [we don't call field resolvers on sort fields before querying](https://github.com/gatsbyjs/gatsby/blob/70c3f32bb52c07f9690b4c011f4a9ad1564af5c3/packages/gatsby/src/redux/run-sift.js#L305) (only for `filter` fields </s> Fixed; the same issue is probably true for `distinct` and `group` `field` arg)

sidharthachatterjee:
cc @freiksenet

stefanprobst:
Note that all of this is working as expected in the schema refactor branch - I noticed those issues when porting over the old tests, and noted them here so there is something public to refer to.

zslabs:
[...]

PR #11656, opened 2019-02-08
fix(gatsby): Resolve nodes for sorting and stringify Date on sort field

stefanprobst:
Related: #11368, #11654 

This PR stringifies Dates on sort fields to correctly sort when we have a mix of Date objects and date strings.
It also adds sort fields to `fieldsToSift`, i.e. fields where the field resolver should be called before querying. This allows e.g. sorting on linked fields.


stefanprobst:
Sorry, I'll fix this tomorrow

stefanprobst:
Ok, this is working, but not yet with Loki.

stefanprobst:
Unfortunately, this has become quite invasive. Probably better for now to just [do this](https://github.com/gatsbyjs/gatsby/compare/master...stefanprobst:stringify-dates-for-sort), which is a small change -- but won't work for Loki???

DSchau:
@stefanprobst do we want to close this? We've fixed the underlying issue (by reverting the regression), but it would be nice to implement this functionality in a clean way--if we can.

stefanprobst:
Yes let's close this for now. I'll try to tackle our sorting issues once the schema refactor is in.

PR #11659, opened 2019-02-08
Add node model

stefanprobst:
This is for #11480

Questions:
* should nodeModel methods be called with `context`, or with `{ path: context: path }`?
* `getNodesByType` and `getNodeByType` handle GraphQL abstract types, what about `runQuery`?
* what else to put on nodeModel?
* `runQuery` signature (e.g. GraphQL type or typeName as arg)?
* why do we have `TypeConflictReporter` (instead of logging conflict right away)?

### DIFF
--- a/packages/gatsby/src/db/__tests__/node-tracking-test.js
+++ b/packages/gatsby/src/db/__tests__/node-tracking-test.js
@@ -4,7 +4,6 @@ const {
 } = require(`../../redux/actions`)
 const { getNode } = require(`../../db/nodes`)
 const { findRootNodeAncestor, trackDbNodes } = require(`../node-tracking`)
-const nodeTypes = require(`../../schema/build-node-types`)
 const { run: runQuery } = require(`../nodes-query`)
 require(`./fixtures/ensure-loki`)()
 
@@ -18,7 +17,7 @@ function makeNode() {
     },
     inlineArray: [1, 2, 3],
     internal: {
-      type: `TestNode`,
+      type: `Test`,
       contentDigest: `digest1`,
       owner: `test`,
     },
@@ -43,7 +42,7 @@ describe(`track root nodes`, () => {
         },
         inlineArray: [1, 2, 3],
         internal: {
-          type: `TestNode`,
+          type: `Test`,
           contentDigest: `digest2`,
         },
       },
@@ -89,7 +88,19 @@ describe(`track root nodes`, () => {
     let type
 
     beforeAll(async () => {
-      type = (await nodeTypes.buildAll({})).testNode.nodeObjectType
+      const { SchemaComposer } = require(`graphql-compose`)
+      const { addInferredFields } = require(`../../schema/infer/infer`)
+      const { getExampleValue } = require(`../../schema/infer/example-value`)
+
+      const sc = new SchemaComposer()
+      const typeName = `Test`
+      const tc = sc.createTC(typeName)
+      addInferredFields({
+        schemaComposer: sc,
+        typeComposer: tc,
+        exampleValue: getExampleValue({ nodes: [makeNode()], typeName }),
+      })
+      type = tc.getType()
     })
 
     it(`Tracks objects when running query without filter`, async () => {
--- a/packages/gatsby/src/schema/node-model.js
+++ b/packages/gatsby/src/schema/node-model.js
@@ -1,3 +1,4 @@
+const _ = require(`lodash`)
 const {
   getNodes,
   getNode,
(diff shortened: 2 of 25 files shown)


freiksenet:
> why do we have `TypeConflictReporter` (instead of logging conflict right away)?

I think so we can log them all at the end in one place.

packages/gatsby/src/db/__tests__/node-tracking-test.js
```diff
@@ -89,7 +88,19 @@ describe(`track root nodes`, () => {
     let type
 
     beforeAll(async () => {
-      type = (await nodeTypes.buildAll({})).testNode.nodeObjectType
+      const { SchemaComposer } = require(`graphql-compose`)
```
freiksenet: Considering this is a such common pattern, is our API a bit wrong? Ideally we would be able to call `addInferredType({ schemaComposer, typeName: 'Test', nodeStore: fakeNodeStore, })`, what is preventing that from happening?

packages/gatsby/src/schema/types/NodeInterface.js
```diff
@@ -22,13 +21,19 @@ const getOrCreateNodeInterface = schemaComposer => {
...
+          return context.nodeModel.getNode(source.parent, { path })
+        },
       },
       children: {
         type: `[Node]!`,
-        resolve: async (source, args, context, info) =>
-          source.children.map(context.nodeModel.getNode),
+        resolve: (source, args, context, info) => {
+          const { path } = context
+          return source.children.map(id =>
```
freiksenet: The fact that we use this pattern several times, means we need `getNodes` as a model method.

packages/gatsby/src/schema/node-model.js
```diff
@@ -0,0 +1,99 @@
...
+  return result
+}
+
+const nodeModel = {
+  findRootNodeAncestor,
+  getNode: withPageDependencies(getNodeById),
+  getNodeByType: withPageDependencies(getNodeByGQLTypeName),
+  getNodes: withPageDependencies(getNodes),
+  getNodesByType: withPageDependencies(getNodesByGQLTypeName),
+  // TODO: What should this return? Schema types? Types in the store?
```
freiksenet: We can't really return schema types, so it should return type names.

packages/gatsby/src/schema/node-model.js
```diff
@@ -0,0 +1,99 @@
...
+  // hasNodeChanged,
+  // getNodeAndSavePathDependency,
+  // TODO: Rethink the signature of runQuery!
+  runQuery,
+} = require(`../db/nodes`)
+const { findRootNodeAncestor } = require(`../db/node-tracking`)
+const createPageDependency = require(`../redux/actions/add-page-dependency`)
+const { store } = require(`../redux`)
+
+const withPageDependencies = fn => async (args, pageDependencies) => {
```
freiksenet: Really cool.

packages/gatsby/src/schema/schema.js
```diff
@@ -285,10 +272,14 @@ function createChildrenField(typeName) {
...
-          .map(context.nodeModel.getNode)
-          .filter(nodeIsOfType(typeName))
+      async resolve(source, args, context) {
+        const { path } = context
+        const result = await Promise.all(
+          source.children.map(id =>
+            context.nodeModel.getNodeByType({ id, type: typeName }, { path })
+          )
+        )
+        return result.filter(Boolean)
```
freiksenet: Why can it have non-true values? getNodeByType with typename should filter the stuff.

packages/gatsby/src/schema/schema.js
```diff
@@ -298,10 +289,14 @@ function createChildField(typeName) {
...
-          .map(context.nodeModel.getNode)
-          .find(nodeIsOfType(typeName))
+      async resolve(source, args, context) {
+        const { path } = context
+        const result = await Promise.all(
+          source.children.map(id =>
+            context.nodeModel.getNodeByType({ id, type: typeName }, { path })
+          )
+        )
+        return result.find(Boolean)
```
freiksenet: Ditto as above.

packages/gatsby/src/schema/types/Date.js
```diff
@@ -61,7 +61,34 @@ export function isDate(value) {
...
+  } else if (difference) {
+    return moment().diff(
+      moment.utc(normalizedDate, ISO_8601_FORMAT, true).locale(locale),
+      difference
+    )
+  }
+  return normalizedDate
+}
+
+export const dateResolver = {
```
freiksenet: :+1:

freiksenet:
> `getNodesByType` and `getNodeByType` handle GraphQL abstract types, what about `runQuery`?
> * `runQuery` signature (e.g. GraphQL type or typeName as arg)?

Probably should have same behaviour. So probably graphql type.

> * should nodeModel methods be called with `context`, or with `{ path: context: path }`?

Let's have all arguments explicit, so we don't pass context or info objects directly to model layer.

>  * what else to put on nodeModel?

getNodes that gets multiple nodes by ids.

freiksenet:
Really happy about this stuff in general, great work. :+1:

packages/gatsby/src/db/__tests__/node-tracking-test.js
stefanprobst: Oh this would work perfectly, if we provide a nodeStore

PR #11670, opened 2019-02-09
feat(starters): add starter magicsoup.io

simonfranzen:
## Features

Starter includes Optimized images, SEO friendly, Google Webmaster Tools, Responsive UI, Static content with JSON and Markdown, Convert Markdown to StyledComponents, Webfonts, SSR, Testing


simonfranzen:
Hey, I have no idea why it fails. I had some problems with characters in wrong encoding, but every validator is telling me that my yml is valid. The error is something with `gatsby-starter-blog` but I dont have any dependency with that. 

Thank you

docs/starters.yml
```diff
@@ -1829,3 +1829,21 @@
...
+- url: https://gatsby.magicsoup.io/
+  repo: https://github.com/magicsoup-io/gatsby-starter-magicsoup
+  description: A production ready gatsby starter using magicsoup.io
+  tags:
+    - SEO
+    - Markdown
+    - Testing
+    - Styling:Styled-Components
+  features:
+    - Optimized images! No more large, unoptimized images which dramatically slow down your site. The starter uses gatsby-image for optimized image loading.
```
DSchau: Can we drastically simplify these features? If you compare to some other starters--they are generally just the first part of these features, e.g.

```suggestion
    - Optimized images
```

Not as fun, I know--but generally more consistent!

simonfranzen: Yes for sure. I will mention which starter or how tos I have implemented here and cut it down.

simonfranzen: I removed the information about the images, because it's gatsby standard I would think.

simonfranzen: The info about SSR is obsolet as well I think

PR #7710, opened 2018-08-29
[sites-showcase] Adding target for external links

mattbag:
Personally I am used to showcase sites links to open a new tab (definitely for github pages that I might open and read later), also the icon on visit site is suggesting new tab ;)
I skipped the url under the title, and added a space between the icon and 'visit'.

Also note that upon closing this window, the sidebar filters are not retained. I think it's a proper bug.

PR #11671, opened 2019-02-10
chore(www): add target blank to showcase url

Saifadin:
It is quite annoying to go through the showcases and the link not opening a new tab.

This PR adds `target="_blank"` to the Showcase Details Modal.

### DIFF
--- a/www/src/components/showcase-details.js
+++ b/www/src/components/showcase-details.js
@@ -12,7 +12,7 @@ import Layout from "../components/layout"
 import ShareMenu from "../components/share-menu"
 
 import MdArrowUpward from "react-icons/lib/md/arrow-upward"
-import MdLaunch from "react-icons/lib/md/launch"
+import MdLink from "react-icons/lib/md/link"
 import FeaturedIcon from "../assets/featured-detailpage-featuredicon.svg"
 import FeatherIcon from "../assets/showcase-feather.svg"
 import GithubIcon from "react-icons/lib/go/mark-github"
@@ -473,7 +473,7 @@ const ShowcaseDetails = ({ parent, data, isModal, categories }) => (
                         },
                       }}
                     >
-                      <MdLaunch
+                      <MdLink
                         style={{
                           verticalAlign: `sub`,
                         }}


LekoArts:
It was declined in the past
https://github.com/gatsbyjs/gatsby/pull/7710

IMO that’s the right way to do it. You can choose to open it in a new window if you want :)

Saifadin:
In my opinion either we remove the icon which on other places on the page, indicate, that something opens in a new page or we make it target blank otherwise it is bad UX :/

wardpeet:
cc @gatsbyjs/docs @fk

DSchau:
@Saifadin as @LekoArts noted, we've had this exact type of PR in the past and our *excellent* UI and UX designer @fk mentioned that it's nice behavior to defer to the user on behavior here. In other words, if the user wants to open in a new window they can, and if the user wants to to open in a new tab they can right click or "center click" to open in a new tab.

I think that's behavior we want to continue using, and it's internally consistent with links throughout GatsbyJS.org.

As such--I think we should close this out but we do very much appreciate this PR and the time you spent creating it. Hope to see more from you soon!

Thanks again!

marcysutton:
I actually agree that the icon is a little confusing if it opens in the same window. That icon is typically used to indicate "opens in a new window"...perhaps we could use a link icon instead (and open a new issue)?

visit site link button

fk:
> I actually agree that the icon is a little confusing if it opens in the same window. That icon is typically used to indicate "opens in a new window"...

@marcysutton (and everybody) I agree!
So much for being excellent *runs and hides* 😅 😉

DSchau:
Well then let's re-open :) Thanks all!

Should we merge this @fk @marcysutton?

fk:
Yes — thanks (and apologies) everybody! 🙏 

I'm curious if my opinion in regards to "don't force a decision on the user" still holds in 2019 in terms of accessibility (and general UX)? I admit that it's an idealistic one, and agree that we are currently making the "average" user do _more_ to reach the (most likely) desired outcome.

Any accessibility concerns with `_blank` in 2019 @marcysutton?

KyleAMathews:
I still hate _blank :-)

KyleAMathews:
Very few people use it so 🤷‍♂️

marcysutton:
I see two options here:

1. Add `target=_blank` to the link, which also requires adding a `title` or `aria-describedby` reference with the text "opens in a new window" to keep screen reader users from making an unexpected jump in their browser. I wish AT would handle that for us programmatically, but they don't. It's kind of a pain and the results on AUX (accessible user experience) are mixed.

2. Leave the markup as a regular link and change the icon instead. That would be way simpler in my opinion. I think a "link" SVG icon could be dropped in pretty easily here and the UX would be less confusing. Is there an icon system in use where we could find something like that? That would probably mean closing this PR and opening a new issue to start that work.

Saifadin:
It sounds like to me, that the most would rather want to change icon not add `target="blank"`.

I can do that, as soon as we have a decision ;)

wardpeet:
I vote for version 2 👍 I use my keyboard and mouse to open up in new tabs

Saifadin:
Updated the PR to change `MdLaunch` to `MdLink` and remove the `target="_blank"` on the link.

marcysutton:
@Saifadin thanks again for working on this! If you're interested, I found another instance of the "Visit" link that indicates it opens in a new window (but it doesn't), this time on the starters. Would make for a very helpful PR. https://www.gatsbyjs.org/starters/gatsbyjs/gatsby-starter-blog/

Saifadin:
On it ;)

DanielRuf:
In general `target=_blank` is bad UX and if we use it we should also add `rel="noopener noreferrer"`.

PR #11682, opened 2019-02-10
feat(createRemoteFileNode): allow passing headers to request

iFlameing:
## Description
This allows a user to pass header to create-remote-file-node for fetching data which is in private mode in cms.

thanks to @datakurre who also helped me in creating this pr.

### DIFF
--- a/packages/gatsby-source-filesystem/src/create-remote-file-node.js
+++ b/packages/gatsby-source-filesystem/src/create-remote-file-node.js
@@ -186,7 +186,7 @@ async function processRemoteNode({
   createNode,
   parentNodeId,
   auth = {},
-  headers = {},
+  httpHeaders = {},
   createNodeId,
   ext,
   name,
@@ -202,7 +202,7 @@ async function processRemoteNode({
   // See if there's response headers for this url
   // from a previous request.
   const cachedHeaders = await cache.get(cacheId(url))
-  const headers = {}
+  const headers = { httpHeaders }
   if (cachedHeaders && cachedHeaders.etag) {
     headers[`If-None-Match`] = cachedHeaders.etag
   }
@@ -315,7 +315,7 @@ module.exports = ({
   createNode,
   parentNodeId = null,
   auth = {},
-  headers = {},
+  httpHeaders = {},
   createNodeId,
   ext = null,
   name = null,
@@ -361,7 +361,7 @@ module.exports = ({
     parentNodeId,
     createNodeId,
     auth,
-    headers,
+    httpHeaders,
     ext,
     name,
   })
--- a/packages/gatsby-source-filesystem/src/create-remote-file-node.js
+++ b/packages/gatsby-source-filesystem/src/create-remote-file-node.js
@@ -202,7 +202,7 @@ async function processRemoteNode({
   // See if there's response headers for this url
   // from a previous request.
   const cachedHeaders = await cache.get(cacheId(url))
-  const headers = { httpHeaders }
+  const headers = { ...httpHeaders }
   if (cachedHeaders && cachedHeaders.etag) {
     headers[`If-None-Match`] = cachedHeaders.etag
   }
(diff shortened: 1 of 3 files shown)


iFlameing:
@pieh can you review this?

packages/gatsby-source-filesystem/src/create-remote-file-node.js
```diff
@@ -199,7 +200,7 @@ async function processRemoteNode({
   // See if there's response headers for this url
   // from a previous request.
   const cachedHeaders = await cache.get(cacheId(url))
-  const headers = {}
+  headers = { headers: { ...headers } }
```
pieh: can we maybe rename top level - because this line is really confusing when I first read it:
```suggestion
  const requestArgs = { headers: { ...headers } }
```
or something like that? (and of course adjust when needed afterwards)

this actually also made me question if our current code for `If-None-Match` header do anything it all given that it's passed in top level for `got` request (and not in `headers`) - but this is definitely not in scope of this PR

pieh: also - do we need to create extra object copy of headers object (`{...headers}`)? 

wouldn't:
```
const requestArgs = { headers }
```
work?

iFlameing: I think we can do  it. I will modify my changes.

iFlameing: yes, current code for If-none-match not caching header correctly.

iFlameing: I modify the code as you suggested. can you review this now

datakurre: @pieh But with just  { headers } isn’t there danger that the code will later start mutating the headers causing unexpected side-effects and it will be up to caller to ensure that it only gives copy of its configured headers when calling.

iFlameing:
@pieh I think it needs review once again.

iFlameing:
@pieh can you review this? this is quiet simple.

datakurre:
@iFlameing Unfortunately, you need to rebase this.

It also seems that passing headers forward from `requestRemoteNode` has now been fixed in master, so renaming headers to requestArgs is no longer required.

packages/gatsby-source-filesystem/src/create-remote-file-node.js
```diff
@@ -184,6 +184,7 @@ async function processRemoteNode({
   cache,
   createNode,
   auth = {},
+  headers = {},
```
wardpeet: should we call it `customHttpHeaders`, `httpHeaders` or `requestHeaders`

iFlameing: why not should I changed it to httpHeaders, because I think it is the most appropriate. what do you think?

iFlameing:
@datakurre once I get a reply from @wardpeet I will rebase this pr and do the necessary changes as you said.

packages/gatsby-source-filesystem/src/create-remote-file-node.js
pieh: Sorry for late reply - I think `httpHeaders` seems good!

iFlameing: @pieh np! I renamed it as you suggested :)

iFlameing:
@datakurre I rebased the pr and modified the changes as you suggested. Please take a look!

iFlameing:
@pieh, please! review this once again.

pieh:
I think this looks good, but I didn't actually tested as I don't have anything that require some specific headers to download files - so to merge I need some example to test against

iFlameing:
@pieh We already tested this for downloading private images from Plone Source. Here is the link of issue where we tested it https://github.com/collective/gatsby-source-plone/issues/162.

wardpeet:
i'll test with a simple glitch site.

sidharthachatterjee:
Published in `gatsby-source-filesystem@2.0.29`

PR #11693, opened 2019-02-11
www: add analytics tags to the plugin search

Haroenv:
## Description

This will make the analytics more relevant, when we can filter for only the Gatsby results

### DIFF
--- a/www/src/components/plugin-searchbar-body.js
+++ b/www/src/components/plugin-searchbar-body.js
@@ -196,7 +196,7 @@ class Search extends Component {
           <SearchBox translations={{ placeholder: `Search Gatsby Library` }} />
 
           <div css={{ display: `none` }}>
-            <Configure analyticsTags={['gatsby-plugins']} />
+            <Configure analyticsTags={[`gatsby-plugins`]} />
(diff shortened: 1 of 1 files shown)


www/src/components/plugin-searchbar-body.js
```diff
@@ -195,6 +195,7 @@ class Search extends Component {
           <SearchBox translations={{ placeholder: `Search Gatsby Library` }} />
 
           <div css={{ display: `none` }}>
+            <Configure analyticsTags={['gatsby-plugins']} />
```
pieh: `Configure` is not defined anywhere so we probably are missing some import for it?

Haroenv: ah good call, I guess I was sleeping when I made the PR

www/src/components/plugin-searchbar-body.js
```diff
@@ -195,6 +196,7 @@ class Search extends Component {
...
```
Haroenv:
```suggestion
            <Configure analyticsTags={[`gatsby-plugins`]} />
```

PR #11697, opened 2019-02-11
docs(creating-and-modifying-pages): remove unnecessary promise

jkjustjoshing:
I also added a "note" indicating that you can still use this function asynchronously. Instead of (or in addition to) this, we might want to add a true async example.


jkjustjoshing:
Change made after this Twitter exchange https://twitter.com/gatsbyjs/status/1095068494075326465

m-allanson:
Oops, the listing failed - I'll update this quickly.

PR #11698, opened 2019-02-11
feat(gatsby-source-mongodb): mongodb to ver 3, added ability to pass entire connection string

wesleylhandy:
## Questions/Concerns
This PR requires testing. Also, are there any other features of more recent versions of MongoDB or the mongodb npm package that should be incorporated?

## Description

This change allows the use of connections strings compatible with MongoDB 3.6+ by passing in the entire connection string provided by a service such as MongoDB Atlas or mLab. The default configuration is retained for legacy versions and for local defaults.This also updates the mongodb npm package to version 3+.

## Related Issues

None, just my personal need to connect to higher version DB than could connect via the current plugin.


packages/gatsby-source-mongodb/src/gatsby-node.js
```diff
@@ -21,12 +21,15 @@ exports.sourceNodes = (
...
+  const connectionURL = pluginOptions.connectionString
+    ? `${pluginOptions.connectionString}/${dbName}${connectionExtraParams}`
+    : `mongodb://${authUrlPart}${serverOptions.address}:${
+        serverOptions.port
+      }/${dbName}${connectionExtraParams}`
 
   return MongoClient.connect(connectionURL)
-    .then(db => {
+    .then(client => {
+      let db = client.db(dbName)
```
DSchau: nit: but if this is not re-assigned, I generally use a `const`

```suggestion
      const db = client.db(dbName)
```

DSchau: Awesome :) I've added a `status: blocked` label to this so that we can go through that session before getting this merged and validated.

Thanks for doing that!

wesleylhandy:
@DSchau I added a few more tests for mapping.js. I've got to think through ways to create tests for the mongodb connection, but I updated the code based on our session yesterday. Everything should be in working order. Is there anything else you'd like to see on this before merging?

wesleylhandy:
**Screenshot of build via `gatsby-dev`** 
successful_build

**Screenshot of edges imported via plugin to graphql**
successful_graphql_import

wardpeet:
@wesleylhandy how breaking is this?

https://github.com/mongodb/node-mongodb-native/blob/master/HISTORY.md#breaking-changes-1
https://github.com/mongodb/node-mongodb-native/blob/master/HISTORY.md#breaking-changes

wesleylhandy:
@wardpeet The changes shouldn't break what I've implemented. I've updated the call to connect to reflect the expected changes should a user either pass in the entire connection string (which, honestly, makes the most sense, will be precise, and work across platforms) or uses the plugin options to build the string.

So below, we call a new instance of the MongoClient, passing url and options object. If we wanted to allow, we can add more options to the plugin to allow greater customization of these options (https://mongodb.github.io/node-mongodb-native/3.0/api/MongoClient.html). Currently, I'm just passing in the `useNewUrlParser` to allow parsing of the various URI connection strings. All other options passed to the plugin are being appended to the connection string as query parameters.

```
const mongoClient = new MongoClient(connectionURL, { useNewUrlParser: true })
  return mongoClient
    .connect()
    .then(client => {
```

In regards to collection and aggregation, currently, the plugin only makes a general `find` on each collection, returning all the documents within. We would need to develop a more advanced plugin to allow custom queries that would involve aggregation and the various pipeline methods. But I wonder if such advanced queries would just be handled by graphql after the documents have been imported. Thoughts?

```
let collection = db.collection(collectionName)
let cursor = collection.find()
```

jorishermans:
First of all you put the collections into gatsby and graphql. Later on you can program your queries with gatsby and graphql. Having queries already in this plugin will make this plugin more complicated, need to see usecases for that and then test them if they are not already possible with this approach.

Maybe we can also add this as a configuration option, like you are saying { useNewUrlParser: true }. Makes sense, so you open it a little bit more.

Or as an optional feature, so that you can set this or we are setting it for the end-user of the plugin.

wesleylhandy:
@jorishermans I agree. No need to add complex queries at the import stage.

Here's an interesting discussion on the addition of `{ useNewUrlParser: true}`: https://stackoverflow.com/questions/50448272/avoid-current-url-string-parser-is-deprecated-warning-by-setting-usenewurlpars

Using this, in conjunction with my proposed update should not break old connection strings but it should enable forward compatibility and allow passing certain connection strings that currently aren't supported.

If you or others think, I can add another optional parameter whereby connection options like `useNewUrlParser` can be passed separately from the other connection parameters. Technically, there are dozens of other options that could be passed: https://mongodb.github.io/node-mongodb-native/3.0/api/MongoClient.html#.connect

Should I refactor to allow these to be passed as an optional parameter to the plugin?

jorishermans:
It is always nice to make it a little bit more open.

You can refactor it to allow these to be passed as an optional parameter to the plugin for the connectionString, maybe call it connectionStringOptions and then use as default { useNewUrlParser: true}

So Yes wo

wesleylhandy:
@jorishermans I updated to allow passing in options and added more to README with links to documentation to distinguish between what has been termed as `extraParams` and the new additional parameter `clientOptions`. I chose `clientOptions` to distinguish it from the connection string parameters.

I'm not sure why the circleci tests are failing. I have only updated code within this plugin as can be seen from the commits.

wesleylhandy:
@jorishermans @DSchau Hey guys, so I fetched the most recent version of gatsby and merged with this branch, well at least I attempted to. I would love some help getting circleci tests passing.

jorishermans:
It looks good. Last thing for me is to test the example mongoDB website, try to get myself some time this weekend!

jorishermans:
Great new @wardpeet thanks for verifying that for me!

wesleylhandy:
@wardpeet No worries. I know when moving to a new version that has explicitly stated was a breaking change, it's better to be careful. Plus, I learned a ton through this process and look forward to making more contributions to gatsby in the future!

wardpeet:
We would love that @wesleylhandy ❤️

PR #11706, opened 2019-02-12
Improve model layer

stefanprobst:
This is for #11480 

TODO:
- [ ] Handle abstract types in `runQuery`
- [ ] Resolve nodes for `group`/`distinct` `field` arg
- [ ] Use `findRootNodeAncestor` in `is-file` (now that we use the original `node-tracking`)

### DIFF
--- a/packages/gatsby/src/schema/infer/add-inferred-fields.js
+++ b/packages/gatsby/src/schema/infer/add-inferred-fields.js
@@ -7,6 +7,7 @@ const {
 } = require(`graphql`)
 const invariant = require(`invariant`)
 
+const { addNodeInterface } = require(`../types/NodeInterface`)
 const { isFile } = require(`./is-file`)
 const { link, fileByPath } = require(`../resolvers`)
 const { isDate, dateResolver } = require(`../types/Date`)
@@ -212,7 +213,9 @@ const getFieldConfigFromFieldNameConvention = (
     const typeName = linkedTypes.sort().join(``) + `Union`
     type = schemaComposer.getOrCreateUTC(typeName, utc => {
       const types = linkedTypes.map(typeName =>
-        schemaComposer.getOrCreateTC(typeName)
+        schemaComposer.getOrCreateTC(typeName, tc => {
+          addNodeInterface({ schemaComposer, typeComposer: tc })
+        })
       )
       utc.setTypes(types)
       utc.setResolveType(node => node.internal.type)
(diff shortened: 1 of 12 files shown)


packages/gatsby/src/schema/infer/add-inferred-fields.js
```diff
@@ -212,7 +213,9 @@ const getFieldConfigFromFieldNameConvention = (
     const typeName = linkedTypes.sort().join(``) + `Union`
     type = schemaComposer.getOrCreateUTC(typeName, utc => {
       const types = linkedTypes.map(typeName =>
-        schemaComposer.getOrCreateTC(typeName)
+        schemaComposer.getOrCreateTC(typeName, tc => {
```
freiksenet: Why? All node types should be created before inference.

freiksenet:
Some general comments about model layer:

1. Let's make it independent from the state, so that it never calls getState from within the function. One way to do is to make it something stateful, like a class and pass it schema and node access (nodeStorage) on construction. This would also simplify creating tests.
2. We can then create "FakeNodeStorage" and pass it both to model and stuff like inferrence, then we can cut all the code that injects stuff into store.
3. Ideally we'd have several integration level tests that test the whole pipeline. They should work by injecting stuff into global store, constructing the schema and running operations on the schema.

freiksenet:
I think I'll do the above thing, hopefully we won't get too many conflicts.

freiksenet:
I've copied the tests from this branch to main branch.

freiksenet:
> I think I'll do the above thing, hopefully we won't get too many conflicts.

Didn't have time to do anything substantial, all code pushed up.

packages/gatsby/src/schema/infer/add-inferred-fields.js
stefanprobst: Mostly for testing (when we don't use the whole pipeline). I guess doesn't hurt to have it there

stefanprobst:
Just quickly do you mean something like in the last commit? Could of course be made into a class (but just so I understand correctly).
As for the tests: I 100% agree!

freiksenet:
Yeah like that. Class is better so there isn't a global state.

packages/gatsby/src/schema/infer/add-inferred-fields.js
freiksenet: Ok, so I think I've added that addNodeInterface thing as a hack so I could distinguish between typeDefs that erroneously don't have implements. I guess it should actually be in https://github.com/gatsbyjs/gatsby/blob/d45c97c31e198aceb5c9f39928e907339a59fa8c/packages/gatsby/src/schema/infer/index.js

freiksenet: I'll make changes to fix that. Also I fill add "createGatsbySchemaComposer" function that adds default types and directives, that we can use in tests too.

freiksenet:
I've copied model code from here and now closing.

PR #11708, opened 2019-02-12
Fix code example for custom normalizers

stefan-meier:
This PR fixes two errors in the custom normalizers example.


riddla:
@stefan-meier, that's funny: I stumbled upon the doubled return yesterday and wanted to send a PR today :)

stefan-meier:
@riddla Well, great minds think alike ;)

PR #11713, opened 2019-02-12
Update sites.yml

uxuimike:


m-allanson:
Thanks @uxuimike! Heads up that I get a _Your connection is not private_ error when visiting your site. It looks like the site is configured with a Netlify SSL certificate, instead of one for your domain. If you're hosting on Netlify you should be able to get that fixed up in the Netlify admin panel.

uxuimike:
Thanks for the heads up on the certificate.  The 'www' was the problem.  Should be fixed now.

PR #11716, opened 2019-02-12
fix(gatsby-transformer-remark): restore behavior of serializing date-like fields to string

DSchau:
## Description

This PR reverts behavior in gatsby-transformer-remark so that date-like fields are serialized to string. This fixes a bug that some were experiencing where content types with a date string were sorted differently than content types with a date object, e.g. this change will revert the behavior so that the below two posts are sorted _the same_

```md
---
date: 2019-01-01
---
```

```md
---
date: "2019-01-01"
---
```

Note that I do think this was a good change, so we'll want to consider perhaps reverting this and baking in better support in the resolvers for sorting these date fields in the future. But for now--we should revert to avoid breaking people's apps.

## Related Issues

Fixes regression in #10924


packages/gatsby-transformer-remark/src/__tests__/on-node-create.js
```diff
@@ -13,8 +13,11 @@ const {
   inferObjectStructureFromNodes,
 } = require(`../../../gatsby/src/schema/infer-graphql-type`)
 
-describe(`Process markdown content correctly`, () => {
-  const node = {
+let node
+let actions
+let createNodeId
+beforeEach(() => {
```
DSchau: Some light refactoring here to make this easier to use in the future

packages/gatsby-transformer-remark/src/on-node-create.js
```diff
@@ -50,12 +60,16 @@ module.exports = async function onCreateNode(
 
     createNode(markdownNode)
     createParentChildLink({ parent: node, child: markdownNode })
+
+    return markdownNode
```
DSchau: Just to make easier to test--the functionality is already accomplished with createNode and createParentChildLink

DSchau:
Also let's hold off on merging until the Azure pipeline/windows test passes. I've seen some weirdness with dates before--so want to make sure we're good to go there!

packages/gatsby-transformer-remark/src/on-node-create.js
m-allanson: Is it worth mentioning this as a comment?

DSchau: @m-allanson I did mention it later (e.g. `// eslint`) but _probably_.

I'll add that before I merge!

packages/gatsby-transformer-remark/src/__tests__/on-node-create.js
```diff
@@ -93,13 +88,37 @@ Sed bibendum sem iaculis, pellentesque leo sed, imperdiet ante. Sed consequat ma
...
+        ).toBeTruthy()
+        expect(actions.createNode.mock.calls[0][0].excerpt).not.toEqual(0)
+        expect(actions.createParentChildLink.mock.calls).toMatchSnapshot()
+        expect(actions.createNode).toHaveBeenCalledTimes(1)
+        expect(actions.createParentChildLink).toHaveBeenCalledTimes(1)
+      })
+    })
+  })
+
+  describe(`date formatting`, () => {
```
pieh: Is it worth mentioning here (maybe in comment) that this is only needed because there is bug in gatsby core that doesn't really allow sorting mix of date object and strings?

packages/gatsby-transformer-remark/src/__tests__/on-node-create.js
```diff
@@ -23,9 +26,13 @@ describe(`Process markdown content correctly`, () => {
     },
   }
 
-  // Make some fake functions its expecting.
-  const loadNodeContent = node => Promise.resolve(node.content)
+  actions = { createNode: jest.fn(), createParentChildLink: jest.fn() }
```
pieh: can we do:

```
let markdownNode
beforeEach(() => { markdownNode = null})
createNode: jest.fn(markdownNode => createNode = node)
```

to avoid those returns in source? `return {}` is very confusing to read (without looking at tests)

PR #11728, opened 2019-02-13
chore(docs): clarify that Gatsby can be hosted anywhere

### LINKED ISSUE #11721: docs: clarify gatsby site can be hosted anywhere
Recently I saw question:

> Hi, I have a question about deployment.  Can Gatsby be run on any server?  I can only see guides for deploying to AWS, Netlify, Surge etc.  Thanks

that is very reasonable thing to ask when considering new tech stack. 

We don't explicitly say in docs (["Deploying and hosting"](https://www.gatsbyjs.org/docs/deploying-and-hosting/)) or home page ("How Gatsby works" diagram) that you can deploy your site literally anywhere. We do have specific guides for selection of hosting providers, which might seem like there is some specific integration needed.

We should edit copy in ["Deploying and hosting"](https://www.gatsbyjs.org/docs/deploying-and-hosting/) that hosting provider guide list is not exhaustive and that you can deploy anywhere (potentially linking to https://www.gatsbyjs.org/docs/caching/ to make sure people also setup appropriate caching rules for best performance of their sites)

saranshkataria:
fixes #11721

### DIFF
--- a/docs/docs/deploying-and-hosting.md
+++ b/docs/docs/deploying-and-hosting.md
@@ -5,8 +5,8 @@ overview: true
 
 Getting your shiny new Gatsby site onto the cloud is probably the first thing you will want to do now that it's built! Also, give yourself a pat on the back real quick for creating something new!
 
+When you run a Gatsby build, the output is static content: HTML, CSS, JavaScript, and images. This content is incredibly easy and affordable to host with any provider. We have several guides below for hosting with particular providers, but this list is by no means exhaustive. Whether you're deploying with AWS, Netlify, or something in between, switching to static content makes deploying a trivial concern. The only thing that youo need to care about are the [caching rules](https://www.gatsbyjs.org/docs/caching/) for the best performance of your site.
+
 In this section you'll find answers to questions about dealing with app routes that aren't at the root of your domain with `path prefixes`, as well as how to deploy to countless hosting solutions such as [Netlify](https://www.netlify.com/), [AWS Amplify Console](https://console.amplify.aws), and [GitHub Pages](https://pages.github.com/). As well as what's necessary to make sure you're fully prepared for a deploy! 🚀
 
 [[guidelist]]
-
-**Note:** Gatsby deployment is not limited to the above mentioned list and you can deploy anywhere you wish to as long as you setup appropriate [caching rules](https://www.gatsbyjs.org/docs/caching/) for best performance of your site.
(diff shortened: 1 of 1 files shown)


docs/docs/deploying-and-hosting.md
```diff
@@ -8,3 +8,5 @@ Getting your shiny new Gatsby site onto the cloud is probably the first thing yo
 In this section you'll find answers to questions about dealing with app routes that aren't at the root of your domain with `path prefixes`, as well as how to deploy to countless hosting solutions such as [Netlify](https://www.netlify.com/), [AWS Amplify Console](https://console.amplify.aws), and [GitHub Pages](https://pages.github.com/). As well as what's necessary to make sure you're fully prepared for a deploy! 🚀
 
 [[guidelist]]
+
+**Note:** Gatsby deployment is not limited to the above mentioned list and you can deploy anywhere you wish to as long as you setup appropriate [caching rules](https://www.gatsbyjs.org/docs/caching/) for best performance of your site.
```
DSchau: Can we make this clearer by putting this up top? I think it's important enough to be explicit here!

saranshkataria: Mentioning that the list down below is incomplete without the list in context seems a bit odd to me I guess

saranshkataria:
> Additionally, looks like some extra commits came along for the ride 🙃

Not sure how that happened. Though those have been merged in but I will create a cleaner PR having just the documentation change

DSchau:
@saranshkataria 

```
git reset HEAD~3
git stash
git checkout c9f5d75 docs/docs/deploying-and-hosting.md
git push origin +hosting-documentation-change
```

_should_ do the trick!

docs/docs/deploying-and-hosting.md
DSchau: Right - so you wouldn't move this _exact_ content up, you'd move a similar disclaimer up to the top! i.e. something like (feel free to re-word this, wrote this quickly)

```
The result of Gatsby's build process is static content: HTML, CSS, JavaScript, images, and more that power your great web application. This content that is produced by this build process is incredibly easy and affordable to host with any provider. We have several guides below for hosting with particular providers, but this list is by no means exhaustive. Whether you're deploying with AWS, Netlify, or something in between--switching to static content makes deploying a trivial concern.
```

saranshkataria:
@DSchau thanks, that worked! Not sure why the tests are failing though

PR #11740, opened 2019-02-13
fix(gatsby): use route path to serve _exact_ page in client routing

DSchau:
## Description

Small fix to #11610 to route based on the matched sub-path. For example, given a route `/app/client/*` this will serve from the `public/app/client/index.html` path, rather than the root-level `index.html`.


sidharthachatterjee:
Successfully published:
 - gatsby@2.1.7

PR #11741, opened 2019-02-13
docs(gatsby): Add documentation for useStaticQuery

sidharthachatterjee:
Adds documentation for upcoming hook `useStaticQuery` (https://github.com/gatsbyjs/gatsby/pull/11588)

### DIFF
--- a/docs/docs/use-static-query.md
+++ b/docs/docs/use-static-query.md
@@ -2,7 +2,7 @@
 title: Querying data in components with the useStaticQuery hook
 ---
 
-Gatsby v2.1.0 introduces `useStaticQuery`, a new [React Hook](https://reactjs.org/docs/hooks-intro.html) that. Just like the `StaticQuery` component, it allows your React components to retrieve data via a GraphQL query. But with the simplicity, shareability and elegance of Hooks.
+Gatsby v2.1.0 introduces `useStaticQuery`, a new Gatsby feature that provides the ability to use a [React Hook](https://reactjs.org/docs/hooks-intro.html) to query with GraphQL at _build time_. Just like the `StaticQuery` component, it allows your React components to retrieve data via a GraphQL query that will be parsed, evaluated, and the resulting data will be injected into the component. However--`useStaticQuery` uses a hook rather than a render prop!
 
 In this guide, we'll walk through an example using `useStaticQuery`. If you're not familiar with static queries in Gatsby, you might want to check out [the difference between a static query and a page query](/docs/static-query/#how-staticquery-differs-from-page-query).
 
--- a/docs/docs/use-static-query.md
+++ b/docs/docs/use-static-query.md
@@ -1,0 +1,47 @@
+---
+title: Querying Data in Components with the useStaticQuery hook
+---
+
+Gatsby v2.1.0 introduces `useStaticQuery`, a new [React Hook](https://reactjs.org/docs/hooks-intro.html) that. Just like the `StaticQuery` component, it allows your React components to retrieve data via a GraphQL query. But with the simplicity, shareability and elegance of Hooks.
+
+In this guide, we'll walk through an example using `useStaticQuery`. If you're not familiar with static queries in Gatsby, you might want to check out [the difference between a static query and a page query](/docs/static-query/#how-staticquery-differs-from-page-query).
+
+## How to use `useStaticQuery` in components
+
+> 💡 You'll need React and ReactDOM 16.8.0 or later to use `useStaticQuery`.
+
+`useStaticQuery` is a React Hook. All the [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html) apply.
+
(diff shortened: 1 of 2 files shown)


docs/docs/use-static-query.md
```diff
@@ -0,0 +1,47 @@
+---
+title: Querying Data in Components with the useStaticQuery hook
+---
+
+Gatsby v2.1.0 introduces `useStaticQuery`, a new [React Hook](https://reactjs.org/docs/hooks-intro.html) that. Just like the `StaticQuery` component, it allows your React components to retrieve data via a GraphQL query. But with the simplicity, shareability and elegance of Hooks.
```
DSchau:
```suggestion
Gatsby v2.1.0 introduces `useStaticQuery`, a new Gatsby feature that provides the ability to use a [React Hook](https://reactjs.org/docs/hooks-intro.html) to query with GraphQL at _build time_. Just like the `StaticQuery` component, it allows your React components to retrieve data via a GraphQL query that will be parsed, evaluated, and the resulting data will be injected into the component. However--`useStaticQuery` uses a hook rather than a render prop!
```

DSchau: Also maybe link to StaticQuery docs when you mention it?

docs/docs/use-static-query.md
```diff
@@ -0,0 +1,47 @@
...
+
+In this guide, we'll walk through an example using `useStaticQuery`. If you're not familiar with static queries in Gatsby, you might want to check out [the difference between a static query and a page query](/docs/static-query/#how-staticquery-differs-from-page-query).
+
+## How to use `useStaticQuery` in components
+
+> 💡 You'll need React and ReactDOM 16.8.0 or later to use `useStaticQuery`.
```
DSchau:
```suggestion
> 💡 You'll need React and ReactDOM 16.8.0 or later to use `useStaticQuery`.
>
> 📦 `npm install react@^16.8.0 react-dom@16.8.0`
```

docs/docs/use-static-query.md
```diff
@@ -0,0 +1,47 @@
...
+
+`useStaticQuery` is a React Hook. All the [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html) apply.
+
+It takes your `query` and returns your data. That's it!
```
DSchau:
```suggestion
It takes your GraphQL query and injects the retrieved data. That's it!
```

docs/docs/use-static-query.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Basic example
+
+Let's create our familiar `Header` component:
```
DSchau:
```suggestion
Let's create a `Header` component that queries for the site title from `gatsby-config.js`:
```

docs/docs/use-static-query.md
```diff
@@ -0,0 +1,47 @@
...
+
+  return (
+    <header>
+      <h1>{data.site.siteMetadata.title}</h1>
+    </header>
+  )
+}
+```
+
+## Known Limitations
```
DSchau: Do we want to add the timing issue?

sidharthachatterjee: Not sure since it's a general issue that could affect page queries as well (in the fringe case). I feel that mentioning it might cause more confusion than adding value at the moment. What do you think?

sidharthachatterjee: Also will never affect production, only development and I've added an error in the hook for development 

https://github.com/gatsbyjs/gatsby/blob/496641aa3ef33202dc23bafbb46d56b3b473af20/packages/gatsby/cache-dir/gatsby-browser-entry.js#L47

DSchau: > Not sure since it's a general issue that could affect page queries as well (in the fringe case). I feel that mentioning it might cause more confusion than adding value at the moment. What do you think?

Probably agree. Let's keep this as is.

docs/docs/use-static-query.md
```diff
@@ -0,0 +1,51 @@
...
+
+Just like the [StaticQuery](/docs/static-query/) component, it allows your React components to retrieve data via a GraphQL query that will be parsed, evaluated, and injected into the component. However, `useStaticQuery` is a hook rather than a component that takes a render prop!
+
+In this guide, we'll walk through an example using `useStaticQuery`. If you're not familiar with static queries in Gatsby, you might want to check out [the difference between a static query and a page query](/docs/static-query/#how-staticquery-differs-from-page-query).
+
+## How to use useStaticQuery in components
+
+> 💡 You'll need React and ReactDOM 16.8.0 or later to use `useStaticQuery`.
+>
+> 📦 `npm install react@^16.8.0 react-dom@16.8.0`
```
pieh:
```suggestion
> 📦 `npm install react@^16.8.0 react-dom@^16.8.0`
```
or just drop versions all together - it should update versions (both in package.json and node_modules)

PR #11743, opened 2019-02-13
fix(gatsby-remark-images): ensure query string is ignored when detecting images

anotherjames:
## Description

Referencing an image in markdown like this:
```
![Alt text](../path/to/image.png?raw=true)
```
Does not load the image.

This PR uses `query-string` to parse the URL out of the path.

## Related Issues

(No issue opened yet.)

This is my first PR, I hope it's ok!

### DIFF
--- a/packages/gatsby-remark-images/src/index.js
+++ b/packages/gatsby-remark-images/src/index.js
@@ -6,6 +6,7 @@ const {
 const visitWithParents = require(`unist-util-visit-parents`)
 const getDefinitions = require(`mdast-util-definitions`)
 const path = require(`path`)
+const queryString = require(`query-string`)
 const isRelativeUrl = require(`is-relative-url`)
 const _ = require(`lodash`)
 const { fluid } = require(`gatsby-plugin-sharp`)
@@ -80,7 +81,9 @@ module.exports = (
     const parentNode = getNode(markdownNode.parent)
     let imagePath
     if (parentNode && parentNode.dir) {
-      imagePath = slash(path.join(parentNode.dir, node.url))
+      imagePath = slash(
+        path.join(parentNode.dir, queryString.parseUrl(node.url).url)
+      )
     } else {
       return null
     }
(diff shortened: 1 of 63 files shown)


packages/gatsby-remark-images/src/index.js
```diff
@@ -113,7 +116,7 @@ module.exports = (
     const presentationWidth = fluidResult.presentationWidth
 
     // Generate default alt tag
-    const srcSplit = node.url.split(`/`)
+    const srcSplit = queryString.parseUrl(node.url).url.split(`/`)
```
DSchau: Not specific to your PR (this behavior already existed!), but this seems like a good opportunity for a helper, perhaps something like?

```js
const getImageInfo = uri => {
  const { url, query } = queryString.parseUrl(uri)
  return {
    ext: path.extname(url),
    url,
    query
  }
}
```

[Here's an example](https://runkit.com/dschau/5c6483a2bbdc5c00133343cf)

anotherjames: I've added a helper in 185f8ca. The filename without the extension is pulled out for a default alt tag, which could be added to the helper. But that's the only place that's done, so I thought best to leave that part unless it's needed elsewhere. I'm happy to move it into the helper too if wanted though :-)

I'll look into adding a test next.

anotherjames:
I don't understand why those tests have failed, the failures don't look relevant to me? Anyone that does, please help, thanks! :-)

DSchau:
@anotherjames we had a little hiccup with one of our dependencies this morning--all fixed now! I just merged upstream/master into your branch and we should be good to go as far as tests passing now!

packages/gatsby-remark-images/src/index.js
```diff
@@ -328,7 +338,7 @@ module.exports = (
                 return resolve()
               }
 
-              const fileType = formattedImgTag.url.slice(-3)
+              const fileType = getImageInfo(formattedImgTag.url).ext
```
DSchau: `path.extname` returns a file extension with a dot. In other words, given `const file = 'asdfasdf.png'`

- `path.extname(file) === '.png'`
- `file.slice(-3) === 'png'`

Wouldn't this then break the subsequent checks?

anotherjames: Oh, yes, sorry!

anotherjames: That should now be resolved by d10a826. Thanks so much for you guidance throughout this PR!

DSchau:
Published as [`gatsby-remark-images`@3.0.4](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-remark-images/CHANGELOG.md#304-2019-02-15)

Thanks for the contribution!

anotherjames:
Thank you for the wonderfully warm welcome! 😃

PR #11751, opened 2019-02-14
feat(starters): Add font smoothing

jahlherapath:
This PR adds font smoothing via `layout.css` to the default gatsby starter theme.

### DIFF
--- a/starters/default/src/components/layout.css
+++ b/starters/default/src/components/layout.css
@@ -5,6 +5,8 @@ html {
 }
 body {
   margin: 0;
+  -webkit-font-smoothing: antialiased;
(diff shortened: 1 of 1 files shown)


jahlherapath:
Thanks for the feedback! @DSchau 

1. Yeah it basically makes the text more legible and a good inclusion for sites with heaps of content. No issues between platforms.
2. I could add to the other starters but there are no default styles (mostly inline).

wardpeet:
@DSchau windows will just ignore it as they don't support font smoothing. A workaround is using SVGs for windows but it can gives you weird issues as it's less flexible as woff, ...

By adding this it will only make osx a bit smoother and I'm all for it! 😄

PR #11752, opened 2019-02-14
fix(gatsby-source-shopify): fix query pagination

### LINKED ISSUE #11724: [gatsby-source-shopify] Products limited to 250
## Description

Shopify restricts node edges to 250 per query, this project seems to have missed this feature so only the first 250 products are returned.

Issue is brought up here: https://github.com/angeloashmore/gatsby-source-shopify/issues/1#issuecomment-426396621
## Steps to reproduce

- Hook gatsby-source-shopify into a store with over 250 products, attempt to find the data.
- Alternatively, create a collection, attach products 251+ to that collection, gatsby will fail to build as it is unable to link (find) the products in that collection

## Expected result

All products should be indexed

##  Actual result

```Error: Invariant Violation: Encountered an error trying to infer a GraphQL type for: "products___NODE". There is no corresponding node with the id field matching:```

## Environment

```
  System:
    OS: macOS High Sierra 10.13.6
    CPU: (4) x64 Intel(R) Core(TM) i5-5250U CPU @ 1.60GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 11.6.0 - /usr/local/bin/node
    npm: 6.5.0 - /usr/local/bin/npm
  Languages:
  Browsers:
    Chrome: 71.0.3578.98
    Firefox: 62.0
    Safari: 12.0.3
  npmPackages:
    gatsby: ^2.0.115 => 2.0.115
    gatsby-link: ^2.0.10 => 2.0.10
    gatsby-plugin-google-analytics: ^2.0.6 => 2.0.13
    gatsby-plugin-layout: ^1.0.12 => 1.0.12
    gatsby-plugin-sass: ^2.0.10 => 2.0.10
    gatsby-plugin-sharp: ^2.0.20 => 2.0.20
    gatsby-source-filesystem: ^2.0.20 => 2.0.20
    gatsby-source-shopify: ^2.0.13 => 2.0.13
    gatsby-transformer-sharp: ^2.1.13 => 2.1.13
  npmGlobalPackages:
    gatsby-cli: 2.4.9```

pieh:
Main fix here is that `get` was using `false` argument (for unknown reason) in place for data object, so the condition was never true.

Just in case - issue that this PR is fixing can be reproduced by using https://github.com/gatsbyjs/store.gatsbyjs.org and monkey-patching https://github.com/gatsbyjs/gatsby/blob/67c0131a0dbaac4d9b535197f70bcbde0f37f49c/packages/gatsby-source-shopify/src/lib.js#L41 to default to for example 5.

250 default limit is shopify's maximum (using more would cause query to error out)

fixes #11724

### DIFF
--- a/packages/gatsby-source-shopify/src/lib.js
+++ b/packages/gatsby-source-shopify/src/lib.js
@@ -39,8 +39,8 @@ export const queryAll = async (
   path,
   query,
   first = 250,
-  after,
-  aggregatedResponse
+  after = null,
+  aggregatedResponse = null
 ) => {
   const data = await queryOnce(client, query, first, after)
 
@@ -51,8 +51,8 @@ export const queryAll = async (
     ? aggregatedResponse.concat(nodes)
     : nodes
 
-  if (get([...path, `pageInfo`, `hasNextPage`], false, data))
-    return queryAll(
+  if (get([...path, `pageInfo`, `hasNextPage`], data)) {
+    return await queryAll(
       client,
       path,
       query,
@@ -60,6 +60,7 @@ export const queryAll = async (
       last(edges).cursor,
       aggregatedResponse
     )
+  }
 
   return aggregatedResponse
 }
--- a/packages/gatsby-source-shopify/src/lib.js
+++ b/packages/gatsby-source-shopify/src/lib.js
@@ -52,7 +52,7 @@ export const queryAll = async (
     : nodes
 
   if (get([...path, `pageInfo`, `hasNextPage`], data)) {
-    return await queryAll(
+    return queryAll(
       client,
       path,
       query,


packages/gatsby-source-shopify/src/lib.js
```diff
@@ -51,15 +51,16 @@ export const queryAll = async (
     ? aggregatedResponse.concat(nodes)
     : nodes
 
-  if (get([...path, `pageInfo`, `hasNextPage`], false, data))
-    return queryAll(
+  if (get([...path, `pageInfo`, `hasNextPage`], data)) {
+    return await queryAll(
```
DSchau: nit: I think these are equivalent (await vs return await)

pieh: hah, you are right - I initially replaced return with await, only then to add return back thinking to limit changes that don't matter and forgetting I added await :D

sidharthachatterjee:
Verified locally, thank you @pieh 👍

PR #11753, opened 2019-02-14
feat(www): Marking official plugins in Plugin Library

### LINKED ISSUE #11082: Plugin Library: Mark plugins that are officially maintained by Gatsby
## Summary

It would be helpful to have on plugins that are maintained by Gatsby to have something on each plugin page so devs can know what plugins are official plugins compared to developed by the community. So something like the gatsby-image and other official plugins can be easily noticed as being products being built internally by Gatsby

lannonbr:
## Description

I've made a indicator in the plugin library to show the user if a plugin is official plugin that is located in the packages/ folder in the Gatsby monorepo. In the sidebar there is a Gatsby logo as suggestion by @shannonbux to notify if it is an official plugin. As well, I'm adding a component in the main section to also state it a bit more clearly that plugins with the Gatsby logo are official plugins.

screen shot 2019-02-22 at 7 48 12 pm

**TODO:**

- [x] figure out design for component on main section
- [x] Try fixing bug where it appears in the main section for every single page no matter if a plugin is official or 3rd party

## Related Issues

resolves #11082


www/src/components/package-readme.js
```diff
@@ -51,6 +52,26 @@ const PackageReadMe = props => {
         )}
       </div>
 
+      <div
```
DSchau: Don't you just need to perform similar logic here and check whether `githubUrl` contains gatsbyjs/gatsby?

shannonbux:
Looking good! I think adding it to the right side of the page is a great addition too. @fk might be curious to see this!

wardpeet:
yeah, I'm curious what @fk has to say.

I would put the Gatsby icon somewhere on the same level as the download count, now it's really overwhelming but might on purpose 😃

lannonbr:
> yeah, I'm curious what @fk has to say.
> 
> I would put the Gatsby icon somewhere on the same level as the download count, now it's really overwhelming but might on purpose 😃

Yeah I am thinking that may be good as well because then it will have a fairly consistent placement where currently it isn't always aligned and sometimes feels a bit off as seen in the sidebar for `gatsby-plugin-removing-trailing-slashes`

fk:
How about something along the lines of this:

image

image

lannonbr:
That looks great @fk. Less obtrusive but still can get the point across. I'll look into implementing this later today

www/src/components/package-readme.js
lannonbr: Done

lannonbr:
And ready for another look

DSchau:
@lannonbr looks like we have a merge conflict. Could you resolve when you get a chance?

Once done, we can get this merged in!

Thank you!

lannonbr:
just fixed the merge conflict

www/src/components/plugin-searchbar-body.js
```diff
@@ -420,6 +423,21 @@ const Result = ({ hit, pathname, query }) => {
             fontSize: rhythm(4 / 7),
           }}
         >
+          {hit.repository &&
+            hit.repository.url.startsWith(
```
DSchau: Actually - one note here--startsWith [will break compatibility with IE11](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith), correct?

We can include a polyfill--but it'd probably be just as easy to do something like:

```js
hit.repository.url.indexOf('https://github.com/gatsbyjs/gatsby') === 0

// or

hit.repository.url.test(/^https:\/\/github.com/gatsbyjs/gatsby/)
```

neither are as succinct, but that's probably OK here!

lannonbr: Just swapped startsWith with indexOf in both locations I do this check

www/src/components/plugin-searchbar-body.js
```diff
@@ -420,6 +423,20 @@ const Result = ({ hit, pathname, query }) => {
...
+          {hit.repository &&
+            hit.repository.url.indexOf(`https://github.com/gatsbyjs/gatsby`) ===
+              0 && (
+              <img
+                src={GatsbyIcon}
+                css={{
+                  width: 20,
+                  marginBottom: 0,
+                  marginRight: 5,
+                  filter: selected ? false : `grayscale(100%) opacity(50%)`,
```
wardpeet:
```suggestion
                  filter: selected ? false : `grayscale(100%)`,
                  opacity: selected ? false : `50%`,
```
filter is "expensive" I'm unsure if this actually improves anything

lannonbr: Further, would it be better to create a separate SVG for the logo in grayscale rather than applying a CSS filter?

DSchau: @wardpeet it should probably be fine here though, right? I presumed that was more a case of images that are very large--is that off base?

lannonbr:
@m-allanson cleaned up the typography and now if any plugin has an `@` at the beginning indicating it is namespaced, it will not be labelled as an official plugin

fk:
Hey @lannonbr — the conflict is my bad, I just merged #12126.
If it's OK for you, I'm happy to resolve it?

lannonbr:
> Hey @lannonbr — the conflict is my bad, I just merged #12126.
> If it's OK for you, I'm happy to resolve it?

Yeah it's good on my end so go ahead to resolve the conflict

fk:
👋 Hey again Benjamin!
I jumped ahead and did a little more than only resolving the merge conflict. 😬 

I really hope you don't mind 🙏🙏 — my fingers just itched. 😅 
Here's how it looks now:

image

I chose to put the "Official Plugin" label in front of the GitHub link (that now has a label thanks to @kbrgl's work from #12126), reduced the Monogram size, and added very basic responsiveness for the „Plugin header items“ („Official Plugin“ label, GitHub source link, „Starters using this“).

My vote is :shipit: — we can add a Monogram SVG component and think about order and improving things for small screens in another iteration.

lannonbr:
Looks good. Thanks for the help 😄

fk:
@lannonbr Sure! Happy to assist! 🤗
But you are the MVP, so _thank you!_ 🙏

m-allanson:
Looking great, thanks everyone :)

PR #11758, opened 2019-02-14
fix #11006 ServiceWorker registration causes a quick blank page on website loading/refresh

### LINKED ISSUE #11006: [gatsby-plugin-offline] ServiceWorker registration causes a quick blank page on website loading/refresh
## Description

In a project using the gatsby-offline-plugin, in production (gatsby build), when you access or refresh a page, you'll get a quick blank page instead of the styled HTML from the React's render. Disabling the ServiceWorker from Chrome's DevTools (bypass for network) fixes this.

### Steps to reproduce

1. Install gatsby-offline-plugin with vanilla options in a project with a home page that has a colored background.
2. Build.
3. Serve your build and access the root url of the website.

### Expected result

Not having any blank page rendering as it is when you disable the ServiceWorker.

### Actual result

A blank page appears before the page starts rendering.

### My thoughts on the problem

I think that the registration of the ServiceWorker is blocking the React's render and that delaying it after everything is loaded should fix this. Related: https://stackoverflow.com/questions/47637126/why-my-service-worker-block-the-render 

### Environment

**System**:
    OS: Linux 4.20 Arch Linux undefined
    CPU: (4) x64 Intel(R) Core(TM) i5-6200U CPU @ 2.30GHz
    Shell: 4.4.23 - /bin/bash
**Binaries**:
    Node: 11.6.0 - /usr/bin/node
    npm: 6.5.0 - /usr/bin/npm
**Languages**:
    Python: 3.7.2 - /usr/bin/python
**Browsers**:
    Firefox: 64.0
**npmGlobalPackages**:
    gatsby-cli: 2.4.8

Pong420:
## Description

Since plugin fallback to `app-shell` ( the blank page ) each time so just add a checking before.

## Related Issues

Fixes #11006

### DIFF
--- a/packages/gatsby-plugin-offline/src/sw-append.js
+++ b/packages/gatsby-plugin-offline/src/sw-append.js
@@ -12,6 +12,10 @@ const navigationRoute = new workbox.routing.NavigationRoute(({ event }) => {
       const offlineShell = `%pathPrefix%/offline-plugin-app-shell-fallback/index.html`
       const cacheName = workbox.core.cacheNames.precache
 
+      if (navigator.onLine) {
+        return fetch(event.request)
+      }
+
       return caches.match(offlineShell, { cacheName }).then(cachedResponse => {
         if (cachedResponse) return cachedResponse
 


davidbailey00:
Thanks for this PR :smile: We're super happy to have you contributing to Gatsby!

However, I feel like it might not be best to merge this currently, since it would have a detrimental effect on performance. I'm currently working on a PR which approaches this issue differently, by attempting to serve the page's content as part of the offline shell within the HTML, even when offline (see #11965).

For now I'll be closing this PR in favour of the mentioned one, however please feel welcome to contribute to this PR!

Issue #11733, opened 2019-02-13
[www] Showcase filter not responding to forward/backward buttons


universse:
Hi. The filters on https://www.gatsbyjs.org/showcase aren't responding to forward/backward navigation buttons. There seems to be a bug in components/url-query.

Should be

```
this.unlisten = globalHistory.listen(({ location: { search }) => {
      this.getDerivedStateFromQuery(search)
})
```

instead of

```
this.unlisten = globalHistory.listen(({ search }) => {
      this.getDerivedStateFromQuery(search)
})
```

danedavid:
@universse There doesn't seem to be a trouble with the site showcase modal previous & next buttons.

universse:
Sorry, guess I wasn't clear enough. When you click on sidebar filters both the URL and the Sites section will update. When you click the browser back/forward buttons (not the modal) after clicking a few sidebar filters, the URL will update but the Sites section does not.

danedavid:
Guess this has been fixed in #11770. Hence closing.

PR #11770, opened 2019-02-14
fix(docs): Fix showcase navigation issue when navigating from browser buttons.

gurpreet-hanjra:
## Description

Fix showcase navigation issue when navigating from browser buttons.
Currently its getting search from props (instead of props.location) it throws error on live site

url-query.js:27 Uncaught TypeError: Cannot read property 'replace' of undefined
    at t.getDerivedStateFromQuery (url-query.js:27)
    at url-query.js:18
    at r (history.js:38)

## Related Issues

https://github.com/gatsbyjs/gatsby/issues/11733


m-allanson:
I think the test failures can be fixed by merging the latest code in from the `master` branch.

gurpreet-hanjra:
Thank you, tests are passing now.

PR #11773, opened 2019-02-14
fix(www): Plugin library scrollbar issue

fk:
Don’t set `display` CSS prop to empty string — ends up being interpreted as `false` by the browser 😉 😅. Instead we now explicitly return `false` so that `emotion` doesn't overwrite the `display` prop.

image

### DIFF
--- a/www/src/components/page-with-plugin-searchbar.js
+++ b/www/src/components/page-with-plugin-searchbar.js
@@ -34,7 +34,7 @@ const widthLarge = rhythm(16)
 const styles = {
   sidebar: {
     height: `calc(100vh - ${presets.headerHeight})`,
-    width: `100vw`,
+    width: `100%`,
     zIndex: 1,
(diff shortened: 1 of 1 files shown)


www/src/components/page-with-plugin-searchbar.js
```diff
@@ -20,7 +20,7 @@ const PageWithPluginSearchBar = ({ isPluginsIndex, location, children }) => (
       css={{
         ...styles.content,
         // mobile: hide README on gatsbyjs.org/plugins index page
-        display: `${isPluginsIndex && `none`}`,
+        display: isPluginsIndex ? `none` : false,
```
DSchau: _Could_ also solve like so, both work though 🤷‍♂️ 

```suggestion
        ...(isPluginsIndex && { display: 'none' })
```

fk: 💯

PR #11775, opened 2019-02-14
feat(gatsby-source-wordpress): Add "path" field to post/page/custom post type entities

riddla:
## Description

Picked up from the documentation change over at #11126, this makes a `path` field available within the WordPress entities.

## Usage example ##

``` javascript
// [...]
graphql(
  `
    {
      allWordpressPage {
        edges {
          node {
            id
            link
            path
            status
            template
          }
        }
      }
    }
  `
).then(result => {
  if (result.errors) {
    console.log(result.errors)
    reject(result.errors)
  }

  const pageTemplate = path.resolve(`./src/templates/page.js`)

    createPage({
      path: edge.node.path,
      component: slash(pageTemplate),
      context: {
        id: edge.node.id,
      },
    })
  })
});
// [...]
```

## Related Issues

Related to #11126


packages/gatsby-source-wordpress/src/normalize.js
```diff
@@ -608,3 +608,16 @@ exports.createNodesFromEntities = ({ entities, createNode }) => {
     })
   })
 }
+
+exports.createUrlPathsFromLinks = entities =>
+  entities.map(e => {
+    if (e.__type === `wordpress__PAGE` || e.__type === `wordpress__POST`) {
```
pieh: This won't handle custom post types. I wonder if we instead could check:
```js
if (e.link && !e.path) 
```
to check if link is on entity and just sanity check to not override path field if it already exists (someone or some plugin enriching REST endpoint on wordpress side)

riddla: Thanks for your feedback. Also the sanity check makes totally sense.

I was conservative regarding the types, as the current Wordpress project I use for testing only has normal pages and posts. 
But I should have a deeper look into the example data stemming from the official Gatsby demo wordpress site. 

I’ll try it tomorrow :)

riddla: @pieh, I added your suggestion.

cardiv:
Thanks for this PR @riddla! Looks good, I will update #11126 when this gets merged.

packages/gatsby-source-wordpress/src/normalize.js
```diff
@@ -608,3 +608,16 @@ exports.createNodesFromEntities = ({ entities, createNode }) => {
     })
   })
 }
+
+exports.createUrlPathsFromLinks = entities =>
+  entities.map(e => {
+    if (e.link && !e.path) {
+      try {
+        const link = new URL(e.link)
```
cardiv: I had some trouble with Netlify refusing the build because `URL` is not imported in the file. I am not sure if this is required here for production builds. Maybe a Node version change is required – Netlify uses version 8 by default. Switching to version 10 of Node solved most of the past issues i had.

pieh: Yeah, I think we need to import this.

We do this in `gatsby-source-drupl` - https://github.com/gatsbyjs/gatsby/blob/747943cdc9bcffcf21df5c6ac83dd2cc9df16713/packages/gatsby-source-drupal/src/gatsby-node.js#L5

pieh: It should be available in Node 6+ (probably earlier too, but right now we care about 6+)

riddla: @pieh, thanks for the hint. I added the missing import.

sidharthachatterjee:
@riddla Could you please fix the linting issue? Looks great otherwise!

riddla:
@sidharthachatterjee, we should be good now regarding ESLint.

riddla:
@pieh, @sidharthachatterjee, we should be good here?

pieh:
Awesome!
screenshot 2019-02-26 at 14 10 32

creating paths that matches Wordpress URLs will be much easier now!

PR #11777, opened 2019-02-14
fix(gatsby): Stopped queueing further calls to onCreatePage after a page is deleted

### LINKED ISSUE #11173: `deletePage` action is unexpectedly asynchronous
## Description

The issue seems to be that calling `deletePage` in two different `onCreatePage` hooks is unexpectedly asynchronous, causing the [DELETE_NODE reducer](https://github.com/gatsbyjs/gatsby/blob/fc2ff3503ac1ed32a4b7677b12edf4c4ea4fd050/packages/gatsby/src/redux/reducers/nodes.js#L16) to receive 

```js
{ type: 'DELETE_NODE', plugin: undefined, payload: undefined }
```

This causes

image

### Steps to reproduce

Here's a full description with a repro repo: https://github.com/ChristopherBiscardi/gatsby-mdx/issues/159

I could probably get this example even smaller if it needs to be, but [gatsby-mdx's onCreatePage](https://github.com/ChristopherBiscardi/gatsby-mdx/blob/b1657ae41bd04db4beea814a09cfb0cf23c1b774/packages/gatsby-mdx/gatsby/on-create-page.js) uses `deletePage` and so does the user's own `gatsby-node.js`.

### Expected result

* Async code should not overflow into createPagesStatefully lifecycle
* Using `deletePage` should not cause other hooks to fail

### Actual result

```
TypeError: Can not read property 'id' of undefined
```

---

##### More info

It seems like changing 

```js
emitter.on(`DELETE_PAGE`, action => {
  const nodeId = createPageId(action.payload.path)
  const node = getNode(nodeId)
  boundActionCreators.deleteNode({ node })
})
```

to use await

```js
emitter.on(`DELETE_PAGE`, async action => {
  const nodeId = await createPageId(action.payload.path)
  const node = await getNode(nodeId)
  boundActionCreators.deleteNode({ node })
})
```

fixes this issue [in the data bridge's gatsby-node](https://github.com/gatsbyjs/gatsby/blob/fc2ff3503ac1ed32a4b7677b12edf4c4ea4fd050/packages/gatsby/src/internal-plugins/internal-data-bridge/gatsby-node.js#L154-L159). I don't know why as I'm not familiar enough with the internal data bridge.

lannonbr:
## Description

Whenever a `DELETE_PAGE` action is called from a `onCreatePage` API hook, we turn on a flag that stops the rest of the queued `onCreatePage` calls for the partiular page we are working with given we can't try deleting a page that was already deleted. 

## Related Issues

Fixes #11173


packages/gatsby/src/utils/api-runner-node.js
```diff
@@ -229,7 +230,26 @@ module.exports = async (api, args = {}, pluginSource) =>
       apisRunningByTraceId.set(apiRunInstance.traceId, 1)
     }
 
+    let stopQueuedApiRuns = false
+    let unlisten = null
```
lannonbr: We think `unlisten` is  not the clearest variable name to describe the function to unsubscribe from the action we run below. Does anyone have any suggestion for a better variable name?

KyleAMathews:
We should do the same thing for deleting nodes.

lannonbr:
@KyleAMathews would you suggest doing the same in this PR or branch to another PR to stop the queued calls when we delete nodes?

As well, When we were working on this yesterday, @pieh commented that there may be some issues with doing this for deleting nodes. @pieh, could you clarify what you described to me about this again here?

pieh:
> As well, When we were working on this yesterday, @pieh commented that there may be some issues with doing this for deleting nodes. @pieh, could you clarify what you described to me about this again here?

So what I commented on is that we have node owners checks, so I wasn't sure if it's possible to actually delete nodes created by other plugins.

Another reason for not adding it initially was that I wanted to put up PR first for comments. I'm not really fond if putting `onCreatePage` node specific logic in mostly API agnostic `api-runner-node`, but we also didn't want to add any abstraction to make it nicer initially, just so people can comment on this approach of "cancelling" scheduled api calls, before making it "nicer" and applying to more cases

pieh:
> So what I commented on is that we have node owners checks, so I wasn't sure if it's possible to actually delete nodes created by other plugins.

Just checking https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/redux/actions.js#L317-L359 it seems we don't have any checks for owners for node deletion (we do have them for node creation).

Question is should we add those checks? If it's ok to delete nodes created by others plugins then we should adjust changes in this PR to cover `onCreateNode` as well - in this case I definitely would welcome some abstraction.

packages/gatsby/src/utils/api-runner-node.js
pieh: I'm about to work on PR that would also need callback like this and was thinking if this should be named `onAPIRunComplete`?

I will also think I will introduce some abstraction - in this separate PR and then we can coordinate those

lannonbr: So would you like me to rename the callback now or refactor it later?

pieh: I actually needed to make changes a bit elsewhere, so I won't be touching same code. I think it makes sense to rename it here - if that name makes sense to you?

lannonbr: Okay yeah the name fits. I'll push the change

KyleAMathews:
hmm yeah — actually I agree we should just block plugins from deleting nodes created by other plugins. That's our ownership policy, that you can only affect nodes you create.

pieh:
> hmm yeah — actually I agree we should just block plugins from deleting nodes created by other plugins. That's our ownership policy, that you can only affect nodes you create.

This is separate issue and should be handled in different PR (created issue to track it - https://github.com/gatsbyjs/gatsby/issues/11930), so I'll go ahead, fix merge conflict and merge this one

pieh:
Verified again locally. Thanks @lannonbr for jumping in on this one!

Issue #11680, opened 2019-02-10
Disable javascript


yepes:
Hi everyone

I have created a single page webpage that doesn't provide any user interaction. 
I have all the content in markdown files, and I use gatsby to generate a static, just one page website.
All I need is the html content and the images. Everything is working, but gatsby is adding some javascript files that i wont never use.

How can I avoid that?

mherchel:
Gatsby is a static site generator based on React. One of it's purposes is to prefetch additional pages. 

I honestly think that a one-page non-JS site is outside of Gatsby's use-case. For a one-pager, why not use static HTML?

yepes:
Yep, maybe that's right. I just discovered gatsby the other day and had to do something simple and test it. Just wondering if that would be possible. 

I know about other "older" static sites generator, I just wanted to try something new

pieh:
We don't support this out of the box, but if you feel advanteruous I think this would doable using [`onPreRenderHTML`](https://www.gatsbyjs.org/docs/ssr-apis/#onPreRenderHTML) hook in `gatsby-ssr.js` - especially for js bundles you would want to use `getHeadComponents` and filter out `js` ones (or just all script tags, and use `replaceHeadComponents` with filtered list of components. Might need to check "pre-body components" and "post-body components"

KyleAMathews:
People build tons of 1 pager sites with Gatsby. Just like larger sites, it's far more productive to use Gatsby with ready-designed development environment and plugins, and optimized production builds vs doing it by hand. Also one page sites rarely stay one page. Also it's rare too for sites to not eventually need JavaScript.

Back to the subject of removing the js script tags. It'd be nice if there was a plugin which followed the algorithm @pieh described. It could work for all pages or just specified pages.

LekoArts:
> I honestly think that a one-page non-JS site is outside of Gatsby's use-case. For a one-pager, why not use static HTML?

One-Pager are a perfect usecase. I saw so many websites with 30MB of images loading... You get that optimizations for free then

yepes:
@pieh thanks, i will take a look at it and try to do something this week.

@KyleAMathews yeah, that would be awesome. But i'm getting a grasp of React and this is my first project using gatsby, so I don't think I can do that yet

itmayziii:
I wrote a Gatsby plugin that removes Gatsby generated javascript for those that find this issue in the future

https://www.npmjs.com/package/gatsby-plugin-no-javascript

PR #11782, opened 2019-02-14
[WIP] plugin for removing javascript

jordanoverbye:
The idea for this PR came from #11680. I think this would be really cool for single page sites.

TODO
- [ ] Add options to enable/disable on certain pages
- [ ] Filter out gatsby-script-loader
- [ ] Figure out what to do when the website uses gatsby-image. Should we support this?
- [ ] Add documentation

This is just the initial start - any help would be appreciated.


DSchau:
Hello @jordanoverbye!

Thanks for this PR! I think this is going to be a pretty challenging one to implement, and I'm not really sure we'd want to maintain this in our central monorepo (although we'd be happy to document/link to this plugin when a site is truly static).

The central issues here are that:

1. Some features (you noted images, but also links) require the React runtime to function effectively or most effectively
  - e.g. Link content is preloaded on entry into viewport, prefetched onMouseOver
  - Images would (as you note) only show their base64 representation if using gatsby-image
1. The hybrid approach (e.g. statically generated and then enhanced with JS) is a core functionality of Gatsby; not sure we'd _want_ to spend much time maintaining this

With this in mind--it probably makes sense to keep working on this plugin, but outside the context of the Gatsby monorepo--perhaps under your own Github account. However--since this could be a valuable contribution to some in the ecosystem, we'd love to help you out with some of your questions with a [pair programming session](https://www.gatsbyjs.org/docs/pair-programming/). Hopefully that works for you--and thanks again for the PR!

PR #11556, opened 2019-02-05
docs: remove global install instructions and tweak with package scripts

joshdcuneo:
## Description

Changes all docs references to global install of Gatsby CLI to use npx instead.

Changes all docs references to using Gatsby CLI commands to either use npx (eg. npx gatsby --help) or npm scripts (eg. npm run develop).

First PR so let me know if I should do anything differently!

## Related Issues

Fixes #11547

Issue #6896, opened 2018-07-31
Floating Table of Contents for longer docs (ToC)


shannonbux:
@calcsam idea: Quip writing app does this in a clean and clear way

<img width="1046" alt="screen shot 2018-07-31 at 10 09 07 am" src="https://user-images.githubusercontent.com/32467162/43475210-ddcc0f00-94a9-11e8-9107-2379134df175.png">

shannonbux:
Hi @fk I put this in the Docs redesign phase 3 EPIC, open to discussion on how to prioritize it!

fk:
👍

johncmunson:
@shannonbux you might be interested in checking out this [Gatsby starter](https://gatsby-clean-blog.netlify.com/farms-around-the-world/) I've been working on which implements a floating TOC like you've described. I can't remember exactly why it wasn't sufficient for my use case, but my implementation does not make use @DSchau's [TOC plugin](https://github.com/DSchau/gatsby-remark-toc)

shannonbux:
Hi @johncmunson, cool starter! I don't see the floating TOC yet - am I missing something?

Also, just realizing this issue could be split into two
1. Implement TOC in gatsbyjs.org
2. Write a doc that explains how to implement TOC on any site

Finally, have you submitted your starter to the starter library yet?

shannonbux:
@johncmunson that would be excellent. @fk Do you think there would be enough whitespace in docs to implement something like this? This isn't necessarily a top priority, unless @johncmunson you want to work on it!

fk:
@shannonbux Yep, I think starting from at around 1300px viewport width we should have enough whitespace.

johncmunson:
Sounds pretty fun, I'll take a look this week and see what I come up with!

shannonbux:
sounds great!

fk:
Not stale!

lannonbr:
@shannonbux, @marcysutton and I brainstormed on this today and we came up with some ideas on how we could implement this.

On large displays, we have room on the right side of the page to have a sticky section on the right side of the screen that we could have to notify people of their current place in the current document. Below I show how we will take the table of contents out of the normal flow of the document and into a sidebar-like area:

![Large Display TOC](https://user-images.githubusercontent.com/3685876/56622781-72e22580-65ff-11e9-950f-e4628ce7628c.png)

As well, on mobile, we could keep the TOC in the flow of the document but move it up to right under the main heading of the page:

[...]

PR #11787, opened 2019-02-15
docs: add new doc on why Gatsby uses GraphQL

jlengstorf:
This is a new doc that walks through the process of creating increasingly more complex pages in an attempt to better illustrate why GraphQL is so useful in Gatsby.

It starts with hard-coded components, moves on to context, and ends with complex pages that have images — after showing the “hard way” of passing everything through `context` and still ending up with unoptimized images, etc., it then shows how GraphQL can solve for many complexity issues that would otherwise be really hard to manage as a site scales.

I wanted to get initial feedback on this now, but the plan is to add video lessons to this that will walk through building these pages.

@marcysutton @KyleAMathews @gatsbyjs/core @shannonbux @amberleyromo if I missed anything or otherwise need to revise the code, please let me know so I can address it _before_ I make videos. 😅 

Thanks!


docs/docs/why-gatsby-uses-graphql.md
```diff
@@ -0,0 +1,360 @@
...
+2. The image paths are absolute from the _built_ site, not the source code, which makes it confusing to know how to find them from the JSON.
+3. The images are unoptimized, and any optimization you do would have to be manual.
+4. To create a preview listing of all products, we’d need to pass _all_ of the product info in `context`, which will get unweildy as the number of products increases.
+5. It’s not very obvious where data is coming from in the templates that render the pages, so updating the data might be confusing later.
+
+To overcome these limitations, Gatsby introduces GraphQL as a data management layer.
+
+## Create pages using GraphQL
+
+There’s a bit more up-front setup required to get data into GraphQL, but the benefits often far outweigh the cost.
```
DSchau: Maybe a bit stronger here? We can/should strongly encourage GraphQL.

```suggestion
There’s a bit more up-front setup required to get data into GraphQL, but the benefits far outweigh the cost.
```

docs/docs/why-gatsby-uses-graphql.md
```diff
@@ -0,0 +1,360 @@
...
+}
+```
+
+To check that this worked, let’s use the GraphQL Playground, which is available during development, by running:
+
+```
+GATSBY_GRAPHQL_IDE=playground yarn develop
+```
+
+> **NOTE:** The `GATSBY_GRAPHQL_IDE=playground` part of this command is optional. Adding it enables the GraphQL Playground instead of GraphiQL, which is an older interface for exploring GraphQL.
```
DSchau: I would probably just have used the default here. It's nice to just document the happy path.

(But also I agree, I so much prefer the playground)

DSchau: Also I'm aware this would require some updates to screenshots and other stuff, so this can totally stay too!

docs/docs/why-gatsby-uses-graphql.md
```diff
@@ -0,0 +1,360 @@
...
+    </div>
+  )
+}
+
+export default Product
+```
+
+A few notes about this file:
+
+1. The result of the query is added to the template component as the `data` prop.
```
DSchau:
```suggestion
1. The result of the query is injected to the template component as the `data` prop.
```

docs/docs/why-gatsby-uses-graphql.md
```diff
@@ -0,0 +1,360 @@
...
+
+Save this file, run `gatsby develop`, then open `localhost:8000/gql/purple-hat/`:
+
+![Lazy loaded image of an angry cat wearing the purple hat.](./images/why-gql-images.gif)
+
+The image is now optimized and lazy loaded.
+
+After the initial setup, loading data with GraphQL is fairly similar to directly loading JSON, but it provides extra benefits like automatically optimizing images and keeping the data loading in the same place where it’s used.
+
+It’s certainly not required, and in simple use cases it might not be worth the effort, but in many sites and apps, GraphQL will vastly simplify the process of building and optimizing your pages.
```
DSchau: I'd maybe again just tweak this to be stronger. We can and should hold that GraphQL is best practice for structuring Gatsby applications.

```suggestion
GraphQL is certainly not required--as shown earlier--but the benefits of adopting GraphQL are obvious and numerous. GraphQL will generally simplify the process of building and optimizing your pages, and so it is and will remain a best practice in structuring and writing your next, great Gatsby application.
```

docs/docs/why-gatsby-uses-graphql.md
```diff
@@ -0,0 +1,360 @@
...
+  <section>
+    <h1>{pageContext.title}</h1>
+    <div dangerouslySetInnerHTML={{ __html: pageContext.content }} />
+  </section>
+)
+
+export default WithContext
+```
+
+Start the development server with `gatsby develop` and visit `localhost:8000/page-with-context/` to see the created page:
```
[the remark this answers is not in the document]
jlengstorf: Works for me. I'll update!

docs/docs/why-gatsby-uses-graphql.md
```diff
@@ -0,0 +1,360 @@
...
+    />
+    <p>{pageContext.price}</p>
+    <div dangerouslySetInnerHTML={{ __html: pageContext.description }} />
+  </div>
+)
+
+export default Product
+```
+
+Run `gatsby develop`, then open `localhost:8000/product/space-socks/` to see one of the generated products.
```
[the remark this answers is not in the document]
jlengstorf: As long as every starter supports this command, I'm on board with it. I'm going to do one big update for all instances (already ran the find-and-replace), so I'm not going to merge this suggestion, though.

docs/docs/why-gatsby-uses-graphql.md
jlengstorf: Yeah, I could go either way. I think there are definitely simple use cases where GraphQL isn’t needed, and we shouldn’t necessarily push folks toward GraphQL if they don't _need_ it. That being said, we can (and should) have a point of view, so I'm happy to make this a bit more strongly worded.

docs/docs/why-gatsby-uses-graphql.md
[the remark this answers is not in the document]
DSchau: > As long as every starter supports this command, I'm on board with it

They do! They also support `npm start` which is maybe a bit simpler, too.

> I'm going to do one big update for all instances (already ran the find-and-replace), so I'm not going to merge this suggestion, though.

I actually already did this! Some instances were intentionally left because they're referring to the command itself rather than _using_ the command. See gatsbyjs/gatsby#11556

docs/docs/why-gatsby-uses-graphql.md
jlengstorf: (I actually did this because I thought the goal was to make the Playground the default after we’re sure it's not introducing any regressions and I wanted to avoid updating all the screenshots.) 😅

DSchau: I like your style 😎 It's a lot better, in my opinion.

docs/docs/why-gatsby-uses-graphql.md
jlengstorf: I went back and forth on some of this wording. My intention is to avoid any computer science-ey phrasing to make sure we don't lose people on syntax. I know that's a losing battle, but here we are. 😄

DSchau: Leave it :) It _feels like_ we generally use injected (so that's why I suggested) but maybe that term is in itself confusing?

docs/docs/why-gatsby-uses-graphql.md
jlengstorf: Made some tweaks here, but I like it!

docs/docs/why-gatsby-uses-graphql.md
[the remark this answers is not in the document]
jlengstorf: > I actually already did this!

I meant in this article only — just pushed that update. 😄

docs/docs/why-gatsby-uses-graphql.md
jlengstorf: I think that it's one of those things that you don't know until you know. It's hard for me to be sure whether that's me being paranoid or if it's a valid concern, though.

docs/docs/why-gatsby-uses-graphql.md
```diff
@@ -0,0 +1,362 @@
...
+      title: "We Don’t Need No Stinkin’ GraphQL!",
+      content: "<p>This is page content.</p><p>No GraphQL required!</p>",
+    },
+  })
+}
+```
+
+The `context` property accepts an object, and we can pass in any data we want the page to be able to access.
+
+> **NOTE:** There are a few reserved names that _cannot_ be used in `context`. They are: `path`, `matchPath`, `component`, `componentChunkName`, `pluginCreator___NODE`, and `pluginCreatorId`.
```
KyleAMathews: do we document this anywhere else?

jlengstorf: Not that I'm aware of — the docs are pretty light on information about `context` in general. @pieh pointed me to the source code, which is where I grabbed these. https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/redux/actions.js#L134-L141

jlengstorf:
@shannonbux I don't know if I have a strong opinion either way on that. I do think this is designed to make a structured argument, and I wouldn't want the ToC to imply that we're necessarily calling some of these examples "ready-to-use code samples". But it's also nice to be able to navigate between sections.

I think I'd really just like to get the tutorial-style sidebar nav working on all docs pages, personally. 😄

docs/docs/why-gatsby-uses-graphql.md
```diff
@@ -0,0 +1,362 @@
...
+2. Image paths in `data/products.json` can be relative to the JSON file.
+3. Gatsby can automatically optimize images for faster loading and better user experience.
+4. We no longer need to pass all product data into `context` when creating pages.
+5. Data is loaded using GraphQL in the components where it’s used, making it much easier to see where data comes from and how to change it.
+
+### Add the necessary plugins to load data into GraphQL
+
+In order to load the product and image data into GraphQL, we need to add a few [Gatsby plugins](/plugins/). Namely, we need plugins to:
+
+- Load the JSON file into GraphQL ([`gatsby-source-filesystem`](/packages/gatsby-source-filesystem/))
```
stefanprobst: I find the terminology a bit confusing here: what is a GraphQL node? What does it mean when data is loaded into GraphQL? Can we say something like "loaded into Gatsby's internal data store, that can be queried with GraphQL"?

jlengstorf: That makes sense, yeah. I'll clean it up.

docs/docs/why-gatsby-uses-graphql.md
```diff
@@ -0,0 +1,360 @@
...
+For example, if you had the following component:
+
+```jsx:title=src/templates/no-data.js
+import React from "react"
+
+const NoData = () => (
+  <section>
+    <h1>This Page Was Created Programmatically</h1>
+    <p>
+      No data was required to create this page — it’s just a React component!
```
marcysutton: This text is cut off in the code example. Not a huge deal, but I thought I'd pass that along!

docs/docs/why-gatsby-uses-graphql.md
```diff
@@ -0,0 +1,360 @@
...
+
+When Gatsby creates pages, it includes a prop called `pageContext` and sets its value to `context`, so we can access any of the values in our component:
+
+```jsx:title=src/templates/with-context.js
+import React from "react"
+
+const WithContext = ({ pageContext }) => (
+  <section>
+    <h1>{pageContext.title}</h1>
+    <div dangerouslySetInnerHTML={{ __html: pageContext.content }} />
```
marcysutton: A Senior Web Developer, international public speaker and winner of an O'Reilly Web Platform Award in 2016, Marcy is passionate about accessibility and test automation.

marcysutton: Is there really no better standard than `dangerouslySetInnerHTML`? Can we recommend `ReactHtmlParser`, at least? https://www.npmjs.com/package/react-html-parser

docs/docs/why-gatsby-uses-graphql.md
marcysutton: I LOVE this post! It makes a great case for Gatsby and GraphQL, while introducing incremental steps with the added benefit of helping users accomplish different kinds of tasks (maybe context works for other small scenarios?). Nice work!

docs/docs/why-gatsby-uses-graphql.md
jlengstorf: Cut off? Or scrolling?

docs/docs/why-gatsby-uses-graphql.md
jlengstorf: I'm hesitant to introduce another package at this point — I'd almost prefer to drop the HTML altogether before adding additional complexity like that.

We tend to expect that the headless CMS (or the user) is paying attention to innerHTML stuff, but I can see how that's probably making a few assumptions. 😅 

Does `react-html-parser` protect against any attack vectors? Or does it only avoid using `dangerouslySetInnerHTML`? I couldn't see anything about additional safety on the package page (I skimmed, though).

docs/docs/why-gatsby-uses-graphql.md
jlengstorf: Thanks!

docs/docs/why-gatsby-uses-graphql.md
marcysutton: Understood about not wanting to add another package, and one that hasn't been thoroughly vetted. This is obviously a bigger discussion than this one post...so I'm happy to punt on it and discuss elsewhere. But it seems odd to me that the React docs say this: 
> So, you can set HTML directly from React, but you have to type out dangerouslySetInnerHTML and pass an object with a __html key, to remind yourself that it’s dangerous

And then we all use it anyway. ¯\_(ツ)_/¯

docs/docs/why-gatsby-uses-graphql.md
marcysutton: I meant scrolling, sorry!

calcsam:
@shannonbux it would probably make more sense to have a TOC generated programmatically from headings for all docs pages rather than individually on a per-page basis.

shannonbux:
@jlengstorf makes sense to me, and @calcsam that could be cool as long as perhaps there is an opt-out ability for really short docs (like ones that fit on a page), because in that case, the ToC might take more room than its worth. I don't think it's a top priority yet.

marcysutton:
> a TOC generated programmatically from headings for all docs pages

That sounds like a great reusable component! I know I would have used it in the new contributing section (where there are some hard-coded). Opt-in would be the easiest transition, in my opinion, so we could play around with heuristics before automatically inserting a TOC into every docs page (`n` headings? Doc text length? Include subheadings?). I can file an issue for a reusable component so we can keep this post moving.

calcsam:
@marcysutton optimizing heuristics and an opt-in approach seems like a great idea!

Check out https://www.gatsbyjs.org/packages/gatsby-transformer-remark/#getting-table-of-contents if you haven't seen it already --

calcsam:
Also we were talking about this before in this issue: https://github.com/gatsbyjs/gatsby/issues/6896

docs/docs/why-gatsby-uses-graphql.md
jlengstorf: I think it’s more just to remind people that we need to be sure we know what the HTML is. But I agree, it'd be great not to do that.

KyleAMathews:
It's pretty common to support a short code like `[toc]` to add a table of contents at that point. Agree that always adding table of contents isn't what we want as they're only useful on long docs. Though yeah, we could also add a heuristic that e.g there's 1500+ words and > N headers.

docs/docs/why-gatsby-uses-graphql.md
KyleAMathews: It's only dangerous for user generated input — when data comes from a CMS or other place you control about it's fine.

jlengstorf:
Waiting on the egghead videos to be published. Once they’re live this will be good to merge.

jlengstorf:
This is ready! Can I get a final review/merge? cc @marcysutton

docs/docs/why-gatsby-uses-graphql.md
marcysutton: It could we good to document those elsewhere, it would work well on this page: https://www.gatsbyjs.org/docs/creating-and-modifying-pages/#pass-context-to-pages

I'll create an issue and PR for it.

jlengstorf:
@marcysutton ha. Yes. 😅  Let me do that really quickly.

marcysutton:
Perhaps you could link to your new doc from here too? https://www.gatsbyjs.org/docs/querying-with-graphql/#why-is-graphql-so-cool

jlengstorf:
@marcysutton added a few links + updated the sidebar. I _think_ we're good to go?

muescha:
Section: Add the necessary plugins to load data into GraphQL

In the egghead video you use yarn to add the plugins and yarn develop

In the doc later npm to install, but later yarn develop

It should always use one of it and also doc and video should be equal

muescha:
A site node: it takes me a lot of time to find out what to use and when: npm or yarn

npm as user of Gatsby and yarn as developer?

Suggestion: There should be also a doc page about the differences and when to use what

muescha:
Section: Add the necessary plugins to load data into GraphQL

The video contains more info than the text of this section. The text ends with only `slug` in query And the rest is missing as text

LekoArts:
> Suggestion: There should be also a doc page about the differences and when to use what

I think that’s out of scope of the Gatsby docs. One could google that to find the differences. And you can use both anytime, there are no „rules“ for when to use what :)

muescha:
I mean in context of gatsby. If you new to npm then it is a complicated to decide. 

But for example: if i use `yarn` as a user then I never can do `gatsby develop` because this uses `npm`

LekoArts:
If you’re a npm user then you don’t have to switch except when you want to contribute to the gatsby monorepo (as it uses yarn workspaces). So only displaying npm commands everywhere is fine as it’s interchangeable with yarn.

No, you misunderstand something there. gatsby develop is available if you installed the Gatsby CLI. yarn develop and npm run develop are available if the package.json has a script called develop.

muescha:
maybe we  extract this npm/yarn discussion into a new issue? how we can move comments?

Issue #10049, opened 2018-11-20
Smart source-plugin bulk fetching remote file assets


calcsam:
(from a conversation with @KyleAMathews )

**Problem**: People are  frustrated by long build times driven by downloading data from Wordpress

**Cause**: Say gatsby-source-wordpress hits a Wordpress site and starts downloading 10,000 images. This creates problems on a number of levels. Often the receiving Wordpress site can’t deal with so many open connections. It overwhelms the Wordpress site, the server starts running out of memory, and the download slows down or fails altogether

**Proposed solution**: This is a two-part solution -- (1) fundamental technical improvement (2) better UX for users.

- We need more robust downloading of files (for createRemoteFileNode). 
	- We need a file throttler / ConnectionHandler  to manage connections to the remote Wordpress server.
	- It would be powered by an algorithm deciding how many connections to open. The algorithm would do this by track bytes / second downloaded, and continuing to add open connections until the bytes/second rate stalls out. 
	- There’s potential other, simpler, options like “don’t have more than X open connections”
	- Download manager would add new files, make sure that remote server isn’t getting overloaded.
	- Exponential backoffs.
- Provide visibility of download process to user
	- Showing file download count // download progress for remote files -- “we’re downloading 250 of 500 files, eta 5:00” -- reduce user frustration

**Similarity to other systems**:

- @KyleAMathews has tweeted out looking at this problem, and looked for packages that handle connections in Wordpress side. So far, this search has come up blank. 
- Similar problem to the what rsync does -- we’re trying to sync files through smart management of server 
- Also similar to the strategy a lot of browsers use -- TCP --  get initial byte, doubling download speed, exponential backoff, fixed max # of simultaneous open connections….

TylerBarnes:
This plugin handles transferring media files from wp site to wp site. 
https://github.com/wp-sync-db/wp-sync-db-media-files
It might be useful for getting some ideas. I've used it and it works well

muescha:
This download manager should also include an cache and some strategies how not redownload media assets when nothing is changed.

muescha:
Is there also a stragegy only to download the used media assets (when maybe more assets are saved in wordpress but only some used in the used articles)?

cedricdelpoux:
Any news about this issue?

PR #11788, opened 2019-02-15
Add create-remote-file-node-factory

teimurjan:
## Description

I've had a problem with [gatsby-source-prismic](https://github.com/angeloashmore/gatsby-source-prismic) during downloading a big amount of files. The problem was caused by `Request timeout error` raised inside the function `createRemoteFileNode` in *gatsby-source-filesystem*. After a day of debugging, I have found that the problem is in  [200 concurrent tasks](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-source-filesystem/src/create-remote-file-node.js#L102) in *gatsby-source-filesystem*. So I have created a new exported function called `createRemoteFileNodeFactory` which allow to choose your own options for maximum concurrent tasks and also for disabling progress logging(as during the downloading about 1000 files it's pretty noisy)


pieh:
@teimurjan what's the concurrent task limit you used? Maybe we should just lower the default non configurable one?

teimurjan:
@pieh, you know in different situations it’s needed to have different tasks number. In my case it was 5. I don’t know whether it was Prismic DDoS security or my laptop couldn’t run all well. So I think it would be be nice to have a way to control it by yourself

pieh:
We have idea of making file downloader smarter - https://github.com/gatsbyjs/gatsby/issues/10049 so it could auto adjust concurrent connection limit (based on error rates), add auto-retry, etc. but this is not implemented yet.

I'm bit worried about adding this API like this, it would potentially limit us in the future (having to maintain this api, even if it doesn't really fit new design). But perhaps in this case if we finally implement the above idea, we could deprecate this and make this factory really only return passthrough function to new implementation (so no new download managers effectively would be used, but API would still be usable).

Will need to think a bit about this. But if we decide to move forward with factory approach, I think we would want to swap default `createRemoteFileNode` to use that factory (to avoid code duplication):

in `create-remote-file-node.js`
```js
const createRemoteFileNodeFactory = require(`./create-remote-file-node-factory`)
module.exports = createRemoteFileNodeFactory()
```

teimurjan:
I added it with duplication as I didn't want to make a breaking change. So okay, will be waiting for your decision

KyleAMathews:
@teimurjan this a super interesting approach!

It'd be nice to make things a bit more automatic though — we try hard to avoid config whenever possible.

What do you think about the idea of automatically scaling up and down the number of concurrent downloads?

teimurjan:
It's a really good idea!

LekoArts:
Hey @teimurjan 	👋 	Would be really cool to have this feature merged in, do you have time? 😊 	Also, @pieh & @KyleAMathews 	any further comments from you?

teimurjan:
@LekoArts Hey! I've pushed it. Check it out 😁

packages/gatsby-source-filesystem/src/create-remote-file-node-factory.js
```diff
@@ -0,0 +1,262 @@
+jest.mock(`fs-extra`, () => {
```
DSchau: Little confused here--this should be in __tests__ right? e.g.

`packages/gatsby-source-filesystem/src/__test__/create-remote-file-node-factory.js`

packages/gatsby-source-filesystem/src/index.js
```diff
@@ -5,6 +5,7 @@ function loadNodeContent(fileNode) {
 }
 
 exports.createFilePath = require(`./create-file-path`)
-exports.createRemoteFileNode = require(`./create-remote-file-node`)
+exports.createRemoteFileNodeFactory = require(`./create-remote-file-node-factory`)
+exports.createRemoteFileNode = exports.createRemoteFileNodeFactory()
```
DSchau: Nit, but would probably prefer something like this:

```js
const createRemoteFileNodeFactory = require('`./create-remote-file-node-factory`)

// later

exports.createRemoteFileNodeFactory = createRemoteFileNodeFactory
exports.createRemoteFileNode = createRemoteFileNodeFactory()
```

DSchau:
Going to close this per my comments [here](https://github.com/gatsbyjs/gatsby/pull/11788#pullrequestreview-244002449).

That being said, want to make two things super clear:

1. Thanks so much for this PR! We appreciate all the work you've done thus far.
2. We hope you'd be willing to build upon what was done here and try your hand at implementing #10049

Again -- thanks so much!

PR #11795, opened 2019-02-15
fix(gatsby-source-filesystem): Let plugins set parent when creating File nodes with createRemoteFileNode

KyleAMathews:
Not doing this is causing problems as a) when the server is restarted, if the parent of a remote File node is loaded from the cache, the linked File node won't be recreated which means it will be garbage collected and b) if a parent node is deleted, the linked File node won't also be deleted.

This PR is experimental still. Will be putting together a quick test case to make sure this fixes things.


KyleAMathews:
Validated this code change with this gatsby-node.js:

```javascript
const { createRemoteFileNode } = require(`gatsby-source-filesystem`);

exports.sourceNodes = ({ actions, createNodeId }) => {
  const { createNode } = actions;
  createNode({
    id: createNodeId(`test`),
    cool: `yeah`,
    url: `https://t3.ftcdn.net/jpg/00/92/53/56/240_F_92535664_IvFsQeHjBzfE6sD4VHdO8u5OHUSc6yHF.jpg`,
    internal: {
      type: `TESTING`,
      contentDigest: `1`,
    },
  });
};

exports.onCreateNode = async ({
  node,
  actions,
  store,
  cache,
  createNodeId,
}) => {
  const { createNode, createNodeField } = actions;
  if (node.internal.type === `TESTING`) {
    console.log({ node, actions, createNodeId });
    const fileNode = await createRemoteFileNode({
      url: node.url,
      store,
      cache,
      createNode,
      createNodeId,
      parentNodeId: node.id,
    });
    console.log({ fileNode });
    createNodeField({
      node,
      name: `urlFile___NODE`,
      value: fileNode.id,
    });
  }
};
```

KyleAMathews:
@gatsbyjs/core bump

pieh:
I think this is good - we need to document that people need to use `parentNodeId` when using `createRemoteFileNode` in `onCreateNode` API. 

Should we also add warning for nodes creaated in `onCreateNode` that don't have parent set?

sidharthachatterjee:
Is there value in adding the above code that you validated against as an end to end test?

KyleAMathews:
>Is there value in adding the above code that you validated against as an end to end test?

Probably not — this is just normal node stuff which is tested elsewhere — the problem was just that we didn't let plugins set themselves as the parent when using `createRemoteFileNode`.

@pieh yeah, will document & add warning.

sidharthachatterjee:
@KyleAMathews Alright, testing and verifying this locally

KyleAMathews:
Hmm actually their is no way to warn since there's no way to know which plugin called `createRemoteFileNode`. I think the main thing to do is to file issues in various plugins that use this helper.

pieh:
> Hmm actually their is no way to warn since there's no way to know which plugin called `createRemoteFileNode`. I think the main thing to do is to file issues in various plugins that use this helper.

So there is this trick I've used for `createPage` warning - we can create ad hoc stack-trace and then get first frame that is not originating from `gatsby-source-filestem` to get actual file (and location in file) where this was called

sidharthachatterjee:
Successfully published:
 ```- gatsby-remark-graphviz@1.0.7
 - gatsby-source-drupal@3.0.24
 - gatsby-source-filesystem@2.0.21
 - gatsby-source-shopify@2.0.15
 - gatsby-source-wordpress@3.0.37
 - gatsby-transformer-screenshot@2.0.12
 - gatsby@2.1.11```

KyleAMathews:
>So there is this trick I've used for `createPage` warning - we can create ad hoc stack-trace and then get first frame that is not originating from `gatsby-source-filestem` to get actual file (and location in file) where this was called

Ooo that's cool — I'll create an issue for it.

PR #8092, opened 2018-09-12
Add assets to entries sorted according to their dep graph.

jquense:
This ~~steals~~ _borrows_ the chunk sorting algo from html-plugin. At the moment assets are included in sort of a random order which doesn't work for css especially.  There is a bunch here that i'm a little iffy on but in my tests it seems to work...

PR #11800, opened 2019-02-15
fix(gatsby): create a single CSS file to avoid code-splitting issues

### LINKED ISSUE #11072: Stop code splitting CSS?
We get many many big reports for v2 where people report weird CSS problem where depending on how they load the site and the order they visit pages, they get inconsitent styling.

These seem all related to splitting CSS as that means styles load in different orders changing the priority of rules.

We had far fewer problems in v1 without CSS splitting.

This tweet is also a good evidence for the problem.

>Importing components (using .css files) in different orders, whether you're code-splitting or not, can make the order of CSS rules change. It's worse for code-splitting because the order can also change at runtime if user loads screens/components in different orders.
https://twitter.com/necolas/status/1084944571056803840?s=19

CSS splitting can be useful for large sites. But given all the trouble it causes, it seems we should make it opt-in so people know what they're dealing with.

KyleAMathews:
Fixes #11072

### DIFF
--- a/packages/gatsby/src/utils/webpack.config.js
+++ b/packages/gatsby/src/utils/webpack.config.js
@@ -391,6 +391,18 @@ module.exports = async (
       },
       splitChunks: {
         name: false,
+        cacheGroups: {
+          // Only create one CSS file to avoid
+          // problems with code-split CSS loading in different orders
+          // causing inconsistent/non-determanistic styling
+          // See https://github.com/gatsbyjs/gatsby/issues/11072
+          styles: {
+            name: `styles`,
+            test: /\.css$/,
+            chunks: `all`,
+            enforce: true,
+          },
+        },
       },
       minimizer: [
         // TODO: maybe this option should be noMinimize?


wardpeet:
Might be nice to add a link to the [webpack docs in code](https://webpack.js.org/plugins/mini-css-extract-plugin/#extracting-all-css-in-a-single-file). Would be handy to have a nice way to turn it on and off because I think css codesplitting doesn't break when cssmodules is used so if we could like add it as a option to our loaders that would be awesome.

Maybe not worth the extra code. Code looks 👍

KyleAMathews:
Things can still break with CSS Modules (that's what Twitter uses and they still had a lot of trouble) but yeah, it'd be harder. 

We could definitely document how to re-enable CSS splitting — let's see if anyone asks for it first though.

KyleAMathews:
Publishing!

dustinhorton:
I use CSS modules and it’s definitely still been an issue—so much so that I’ve not been using gatsby-link as a workaround. Really excited to try this.

wardpeet:
oh yeah specificity of course 🤦‍♂️. Yeah it can still break

dustinhorton:
@KyleAMathews This didn't fix the issue for me. Still think what I mentioned @ https://github.com/gatsbyjs/gatsby/pull/8092#issuecomment-443570882 is right—it's the preloading & injecting CSS onto the page that's the issue.

KyleAMathews:
Yes please share your code — not sure what CSS would be preloaded anymore? The point of this PR was that we're not code-splitting CSS anymore.

dustinhorton:
https://www.classchirp.com/blog

There's a link near the top that's prefixed with 'hover me', and text right about it that will increase with size.

```
// package.json

[…]
"gatsby": "^2.1.4",
"gatsby-link": "^2.0.10",
[…]
```

Here's the network tab showing CSS being preloaded on that hover:

screen shot 2019-02-16 at 7 45 21 am

KyleAMathews:
Do you have a lock file? Doesn't seem like you've upgraded.

dustinhorton:
I’m on mobile until early tomorrow and seemingly can’t view it on GitHub because the diff is too large. Will update then.

dustinhorton:
@KyleAMathews From my yarn.lock:

```
gatsby@^2.1.4:
  version "2.1.4"
  resolved "https://registry.yarnpkg.com/gatsby/-/gatsby-2.1.4.tgz#a69ee6212d37326cb93238a5cf21d3237d7a6f16"
  dependencies:
    "@babel/code-frame" "^7.0.0"
    "@babel/core" "^7.0.0"
    "@babel/parser" "^7.0.0"
    "@babel/polyfill" "^7.0.0"
    "@babel/runtime" "^7.0.0"
    "@babel/traverse" "^7.0.0"
    "@gatsbyjs/relay-compiler" "2.0.0-printer-fix.2"
    "@reach/router" "^1.1.1"
    address "1.0.3"
    autoprefixer "^9.4.3"
    babel-core "7.0.0-bridge.0"
    babel-eslint "^9.0.0"
    babel-loader "^8.0.0"
    babel-plugin-add-module-exports "^0.2.1"
    babel-plugin-dynamic-import-node "^1.2.0"
    babel-plugin-remove-graphql-queries "^2.6.0"
    babel-preset-gatsby "^0.1.7"
    better-opn "0.1.4"
    better-queue "^3.8.6"
    bluebird "^3.5.0"
    browserslist "3.2.8"
    cache-manager "^2.9.0"
    cache-manager-fs-hash "^0.0.6"
    chalk "^2.3.2"
    chokidar "^2.0.2"
    common-tags "^1.4.0"
    compression "^1.7.3"
    convert-hrtime "^2.0.0"
    copyfiles "^1.2.0"
    core-js "^2.5.0"
    css-loader "^1.0.0"
    debug "^3.1.0"
    del "^3.0.0"
    detect-port "^1.2.1"
    devcert-san "^0.3.3"
    domready "^1.0.8"
    dotenv "^4.0.0"
    eslint "^5.6.0"
    eslint-config-react-app "^3.0.0"
    eslint-loader "^2.1.0"
    eslint-plugin-flowtype "^2.46.1"
    eslint-plugin-graphql "^2.0.0"
    eslint-plugin-import "^2.9.0"
    eslint-plugin-jsx-a11y "^6.0.3"
    eslint-plugin-react "^7.8.2"
    event-source-polyfill "^1.0.5"
    express "^4.16.3"
    express-graphql "^0.6.12"
    fast-levenshtein "~2.0.4"
    file-loader "^1.1.11"
    flat "^4.0.0"
    friendly-errors-webpack-plugin "^1.6.1"
    fs-exists-cached "1.0.0"
    fs-extra "^5.0.0"
    gatsby-cli "^2.4.10"
    gatsby-link "^2.0.10"
    gatsby-plugin-page-creator "^2.0.6"
    gatsby-react-router-scroll "^2.0.4"
    glob "^7.1.1"
    graphql "^14.1.1"
    graphql-playground-middleware-express "^1.7.10"
    graphql-relay "^0.6.0"
    graphql-skip-limit "^2.0.5"
    graphql-tools "^3.0.4"
    graphql-type-json "^0.2.1"
    hash-mod "^0.0.5"
    invariant "^2.2.4"
    is-relative "^1.0.0"
    is-relative-url "^2.0.0"
    jest-worker "^23.2.0"
    joi "12.x.x"
    json-loader "^0.5.7"
    json-stringify-safe "^5.0.1"
    kebab-hash "^0.1.2"
    lodash "^4.17.10"
    lokijs "^1.5.6"
    md5 "^2.2.1"
    md5-file "^3.1.1"
    mime "^2.2.0"
    mini-css-extract-plugin "^0.4.0"
    mitt "^1.1.2"
    mkdirp "^0.5.1"
    moment "^2.21.0"
    name-all-modules-plugin "^1.0.1"
    normalize-path "^2.1.1"
    null-loader "^0.1.1"
    opentracing "^0.14.3"
    optimize-css-assets-webpack-plugin "^5.0.1"
    physical-cpu-count "^2.0.0"
    postcss-flexbugs-fixes "^3.0.0"
    postcss-loader "^2.1.3"
    raw-loader "^0.5.1"
    react-dev-utils "^4.2.1"
    react-error-overlay "^3.0.0"
    react-hot-loader "^4.6.2"
    redux "^4.0.0"
    request "^2.85.0"
    semver "^5.6.0"
    shallow-compare "^1.2.2"
    sift "^5.1.0"
    signal-exit "^3.0.2"
    slash "^1.0.0"
    socket.io "^2.0.3"
    string-similarity "^1.2.0"
    style-loader "^0.21.0"
    terser-webpack-plugin "^1.2.2"
    type-of "^2.0.1"
    url-loader "^1.0.1"
    uuid "^3.1.0"
    v8-compile-cache "^1.1.0"
    webpack "~4.28.4"
    webpack-dev-middleware "^3.0.1"
    webpack-dev-server "^3.1.14"
    webpack-hot-middleware "^2.21.0"
    webpack-merge "^4.1.0"
    webpack-stats-plugin "^0.1.5"
    yaml-loader "^0.5.0"
```

pieh:
It definitely looks like css is split still https://www.classchirp.com/webpack.stats.json.

KyleAMathews:
Can you share your site's code?

dustinhorton:
@KyleAMathews I've just added you as a collaborator. Branch is `gatsby-link`. Let me know if there's another way you'd prefer I share the code.

KyleAMathews:
Confirmed the bug — both on @dustinhorton's site + on using-css-modules + using-sass. Looks like it's specific to gatsby-plugin-sass as other types of css are being combined correctly.

KyleAMathews:
This fixes things https://github.com/gatsbyjs/gatsby/pull/11927

PR #11804, opened 2019-02-15
chore(gatsby-source-wordpress): Correct glob pattern examples

riddla:
## Description

Adds a test for a route blog pattern that is working both on custom hosted WordPress API paths and on API paths from sites  that are hosted on wordpress.com.

## Related Issues

This is picking up the discussion on the routes documentation within #10871 and the according pull request #10887.


packages/gatsby-source-wordpress/src/__tests__/routePatternExamples.js
```diff
@@ -0,0 +1,17 @@
+const minimatch = require(`minimatch`)
+
+describe.only(`Glob patterns in the README`, () => {
```
DSchau:
```suggestion
describe(`Glob patterns in the README`, () => {
```

DSchau: Also generally, not really sure what these tests are testing. It looks like we're basically testing minimatch?

riddla: As minimatch is used within `checkRouteList` I just wanted to (programmatically) show that the suggested patterns would work on both types of WordPress installations. Also as a reference point for futures issues/questions regarding the exclude/include patterns. If more examples/bugs come up, we could extend the tests.

riddla: And thank you for spotting the `.only`. In other projects I used to have a pre-commit hook for this, spotting `.only` within tests ... ;)

I corrected it.

DSchau: @riddla cool! But generally - we want to test _our own_ code, otherwise updating our code won't make this test fail (ever).

So we would want to actually write some assertions that validate `checkRouteList` and import that function, or a function that invokes it, in order to test this more effectively.

riddla: Yeah, you're right. After all it's just about the examples in the Readme and not the `checkRouteList` itself. I removed the test and just added the information that it carried to the commit message. Should be sufficient for future reference, if this topic ever comes up again in some issue.

riddla:
@DSchau would you mind having another look?

PR #11808, opened 2019-02-15
feat(create-gatsby): add a simple CLI forwarder to use with npm init

DSchau:
## Description

This PR introduces an alternate means of initializing a Gatsby project using `npm init`. Why?

- `npm init` requires a convention of `create-<command-name>`, e.g. `create-react-app` can be used with `npm init react-app name-of-my-app`
- `npm init gatsby new my-gatsby-app` seems slightly nicer than `npx gatsby new my-gatsby-app`
  - Note: I can also tweak this to be `npm init gatsby your-blog gatsbyjs/gatsby-starter-blog`, but wanted to keep it flexible in case/when we augment the CLI with plugin capabilities (alternatively, maybe we create npm init gatsby-plugin?)

Consider this an experimental PR. Is this easier to teach? Use? Thoughts and opinions welcome!

cc @marcysutton @jlengstorf

### DIFF
--- a/packages/create-gatsby/.gitignore
+++ b/packages/create-gatsby/.gitignore
@@ -1,0 +1,2 @@
+/lib
+/node_modules
--- a/packages/create-gatsby/README.md
+++ b/packages/create-gatsby/README.md
@@ -1,0 +1,17 @@
+# create-gatsby
+
+A utility that invokes [`gatsby-cli`][gatsby-cli] (that's it!). This exists simply to forward any arguments and commands to gatsby-cli in a specific context, e.g. when invoked with `npm init gatsby`
+
+This should _never_ be used in place of `gatsby-cli`, but rather should be used only in conjunction with `npm init` (see [Usage below](#usage)).
+
+## Usage
+
+All options, commands, etc. should be referenced from [`gatsby-cli`][gatsby-cli], however the most general use case will be creating a new gatsby project, like so:
+
+```shell
+npm init gatsby new my-great-app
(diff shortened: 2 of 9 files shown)


DSchau:
Oh also there's an example of this available via

```shell
npm init @dschau/gatsby new project-name
```

KyleAMathews:
I'd never heard of `npm init` which doesn't make it seem popular. Having two methods to do the same thing seems redundant rather than helpful? Is there anything new this gives us?

DSchau:
> I'd never heard of `npm init` which doesn't make it seem popular.

Yeah, this was one of my concerns. I don't know how popular the feature is. It's documented on create-react-app--but perhaps not mentioned as much because it requires npm 6? https://facebook.github.io/create-react-app/docs/getting-started#npm

> Having two methods to do the same thing seems redundant rather than helpful?

This is what I was worried about. I don't like having two avenues to do the same thing--so we'd have to be consistent in documentation.

> Is there anything new this gives us?

Possibly easier syntax for beginners? `npm init` seems to be clearer in intent than `npx`. I also like the idea of eventually introducing `npm init gatsby-plugin plugin-name` which this would pair nicely with.

Otherwise, nada. That's why I marked it draft/experimental, so thanks for the feedback!

DSchau:
Let's close. We should unify on one path (like we have with `npx`) rather than introducing a split here.

That said--I think this is worth revisiting once/if the `npm init` command becomes more prevalent.

PR #11811, opened 2019-02-15
docs(www): use static query blog post

sidharthachatterjee:
New blog post announcing the new `useStaticQuery` hook


docs/blog/2019-02-20-introducing-use-static-query/index.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Say hello to useStaticQuery
+
+`useStaticQuery` is a hook that takes a GraphQL query and returns your data. That's it.
+
+**No more Render Props necessary to use a Static Query**
+
+This simplifies accessing data in your components and also keeps your component tree shallow.
+
+Let's quickly check out a simple example. Here's a simple Header component, first using `StaticQuery` and then with `useStaticQuery`.
```
DSchau: Nit, but it's nice to not use "simple"

```suggestion
Let's quickly check out an example. Here's a sample Header component, first using `StaticQuery` and then with `useStaticQuery`. Comparing in this way makes the benefits of `useStaticQuery` extremely evident.
```

marcysutton: If possible, it would be good to change usage of the word "simple" to something less assuming. How about "check out an example", and  "Here's a basic Header component"?

docs/blog/2019-02-20-introducing-use-static-query/index.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+### After
+
+```jsx
+import React from "react"
+import { useStaticQuery, graphql } from "gatsby"
+
+const Header = () => {
+  const { site } = useStaticQuery(
```
marcysutton: Would it be overkill to highlight the changes in these examples? Like lines 33, 42 and 44 in the "before" example, and 61, 73 and 75 in the after example (at least in the version I'm looking at).

docs/blog/2019-02-20-introducing-use-static-query/index.md
```diff
@@ -0,0 +1,143 @@
...
+          }
+        }
+      }
+    `
+  )
+  return site.siteMetadata
+}
+```
+
+Now everytime you need the site title, you can use this custom hook. And in case you change the structure of that data, you only need to change the query once!
```
marcysutton: Can you also show an example of this being consumed? I think that would help!

docs/blog/2019-02-20-introducing-use-static-query/index.md
```diff
@@ -0,0 +1,143 @@
...
+
+![excited-gif](./images/excited.gif)
+
+We're really excited about this pattern and everything it enables. With our upcoming [themes][themes] feature, a theme could export custom hooks of its own and consumers could access data without writing a single query.
```
marcysutton: Is this `themes` link accurate?

docs/blog/2019-02-20-introducing-use-static-query/index.md
```diff
@@ -0,0 +1,139 @@
...
+
+# Where can I get this?
+
+<blockquote class="twitter-tweet"><p lang="en" dir="ltr">useStaticQuery hook snuck into <a href="https://twitter.com/gatsbyjs?ref_src=twsrc%5Etfw">@gatsbyjs</a> pretty quick 😍 <a href="https://t.co/PbzTuzk6hb">https://t.co/PbzTuzk6hb</a> waaay nicer API than using the &lt;StaticQuery&gt; component. <a href="https://t.co/6Mf4sSF5C3">pic.twitter.com/6Mf4sSF5C3</a></p>&mdash; Jacob Impson (@jacobimpson) <a href="https://twitter.com/jacobimpson/status/1095930703504584706?ref_src=twsrc%5Etfw">February 14, 2019</a></blockquote> <script async src="https://platform.twitter.com/widgets.js" charset="utf-8"></script>
+
+The sharp reader (Jacob is sharp, be like Jacob) would've noticed that we _stealthily_ released `useStaticQuery` last week.
```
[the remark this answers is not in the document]
sidharthachatterjee: I was actually breaking the fourth wall by acknowledging the current reader of this post. Unless it's super weird and not obvious, let's keep it? 😛 

Replaced sharp with astute for that dramatic old school narrative style ✍️

docs/blog/author.yaml
```diff
@@ -234,3 +234,7 @@
   bio: React and GraphQL enthusiast. Meetup organizer. Mentor. Blogs at dslemay.com
   avatar: avatars/daniel-lemay.jpg
   twitter: "@dslemay"
+- id: Sidhartha Chatterjee
```
marcysutton: Great post Sid! 🤘

docs/blog/2019-02-20-introducing-use-static-query/index.md
sidharthachatterjee: It links to the blog post by @jlengstorf at /blog/2019-02-11-gatsby-themes-livestream-and-example/ 

Not sure what our canonical go to themes link should be at the moment. Suggestions?

docs/blog/2019-02-20-introducing-use-static-query/index.md
sidharthachatterjee: This is a great point! 

> Let's quickly check out a basic example. Here's a typical Header component,

What do you think of this?

docs/blog/2019-02-20-introducing-use-static-query/index.md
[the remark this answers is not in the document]
sidharthachatterjee: @marcysutton Thoughts on this?

docs/blog/2019-02-20-introducing-use-static-query/index.md
marcysutton: Ah, I just wanted to make sure it wasn't a broken link! Looks like we haven't really documented themes yet...so the blog post makes sense for now.

docs/blog/2019-02-20-introducing-use-static-query/index.md
sidharthachatterjee: Definitely not overkill and I see the value in this! Done.

docs/blog/2019-02-20-introducing-use-static-query/index.md
sidharthachatterjee: Great catch! Done.

docs/blog/2019-02-20-introducing-use-static-query/index.md
```diff
@@ -0,0 +1,164 @@
...
+          siteMetadata {
+            title
+          }
+        }
+      }
+    `
+  )
+
+  return (
+    // highlight-next-line
```
marcysutton: I think the `return (` statement needs to be highlighted here...? (I may have gotten a line number wrong)

docs/blog/2019-02-20-introducing-use-static-query/index.md
```diff
@@ -0,0 +1,164 @@
...
+  return site.siteMetadata
+}
+```
+
+Now everytime you need the site title, you can use this custom hook.
+
+```jsx
+const Header = () => {
+  // highlight-next-line
+  const { title } = useSiteMetadata()
```
marcysutton: 🆒 👍

docs/blog/2019-02-20-introducing-use-static-query/index.md
sidharthachatterjee: I highlighted the next line since the value of the return changes but not the return statement. Not sure though. Here is what it looks like. 
screenshot 2019-02-21 05 09 27
screenshot 2019-02-21 05 09 33

docs/blog/2019-02-20-introducing-use-static-query/index.md
marcysutton: That works! :shipit:

docs/blog/2019-02-20-introducing-use-static-query/index.md
marcysutton: Oh yeah, you're totally right–I see it now! You can keep this as-is. :)

PR #11822, opened 2019-02-16
Fix: make portfolio search case insensitive

### LINKED ISSUE #11820: Starter search should be case insensitive
I think that the starter search should be case insensitive. I believe it is better also because on mobile devices the first letter is capitalized by default.

I can fix it!

Search with first letter capitalized: image

Search with first letter non capitalized: image

NeverMendel:
## Description
Make portfolio search case insensitive.

## Related Issues

Fix #11820


wardpeet:
Looks good to me, awaiting feedback of @gatsbyjs/ecosystem to make sure it's something we want but I believe we do

KyleAMathews:
Let's do it 👍 Thanks for the fix!

PR #11824, opened 2019-02-16
Docs cache-control: added complete example + replaced "immutable" flag

baobabKoodaa:
The "immutable" flag is not supported by Chrome, so it shouldn't be used (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control)

I also added a complete working example of headers.


KyleAMathews:
Thanks for the PR. We can't accept this as is as "must-revalidate" is the exact opposite of immutable. Also immutable is supported by chrome https://stackoverflow.com/a/41958007/182702

KyleAMathews:
Also, this is a general document so documentation specific to a single vendor wouldn't be appropriate here. There's a plugin for Netlify that sets up headers for folks.

baobabKoodaa:
Okay, sorry, I didn't realize the Mozilla docs were out of date wrt to "immutable" browser support. In any case "must-revalidate" with max age 1 year is pretty close to "immutable" since the resources can be delivered for 1 year before revalidating.

Issue #10844, opened 2019-01-05
Pages not deleted on node update/delete


rexxars:
## Description

When pages are created from nodes, and you use the `createPageDependency()` action to declare a dependency on the node, when the node is deleted (after the first build), no page rebuilding is triggered. The change is visible once something else triggers a page rebuild, but the page is not removed. 

A similar thing happens when you update a node which causes a new page to be built: the node is updated, but the page building fails with this error:

```
Error loading a result for the page query in "/updated-title-of-post-3". Query was not run and no cached result was found.
Page not found /updated-title-of-post-3 
```

### Steps to reproduce

I've create a sample repo where you can easily reproduce both bugs. See https://github.com/rexxars/gatsby-node-update-bugs

### Expected result

I had expected that the page would be marked as dirty and thus removed. In the case of the updated node, I would have expected the new page to work as intended, while the old page would be removed.

### Actual result

As outlined in the description, the pages never seem to be removed, and new pages builds with errors.

### Environment

```
  System:
    OS: macOS 10.14.2
    CPU: (8) x64 Intel(R) Core(TM) i7-7920HQ CPU @ 3.10GHz
    Shell: 4.4.12 - /usr/local/bin/bash
  Binaries:
    Node: 10.8.0 - ~/.nvm/versions/node/v10.8.0/bin/node
    Yarn: 1.12.3 - ~/.nvm/versions/node/v10.8.0/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v10.8.0/bin/npm
  Browsers:
    Chrome: 71.0.3578.98
    Firefox: 64.0
    Safari: 12.0.2
  npmPackages:
    gatsby: ^2.0.76 => 2.0.85 
  npmGlobalPackages:
    gatsby-cli: 2.4.7
```

sidharthachatterjee:
Thank you for opening this @rexxars 

> no page rebuilding is triggered. The change is visible once something else triggers a page rebuild, but the page is not removed.

Just so I understand correctly, when you say _the page is not removed_, you mean the assets still exist in `public`?

rexxars:
1. You have add a node through `createNode()` of type `blogPost`.
2. Gatsby calls `createPages()`
  * A query is run to fetch all blog posts
  * A page is created from the given blog post
  * `createPageDependency()` binds the page and the blog post node
3. At some point later, the blog post node is deleted through `deleteNode()`
[...]

PR #7388, opened 2018-08-16
don't run debouncedCreatePages on DELETE_NODE

pieh:
Fixes #7300 

Problem with running `debouncedCreatePages ` is it assumes we can finish all processing in 100ms which is not always the case. In #7300 it was causing to run `createPages` before `slug` was attached, so query inside component would never find actual node because it would try to find node with `slug === null`

here's my debugy output from before the change:
```
info changed file at D:\dev\i7300\src\post\tile-layouts.mdx
deleting createMdxNode 32a0c299-deff-5812-894a-6df20be75e2d >>> Mdx
deleting createMdxNode finish 32a0c299-deff-5812-894a-6df20be75e2d >>> Mdx
here's debouncedCreatePages is called --> DELETE_NODE create pages
createMdxNode 32a0c299-deff-5812-894a-6df20be75e2d >>> Mdx
createMdxNode finish 32a0c299-deff-5812-894a-6df20be75e2d >>> Mdx
createPages (slug isn't added, we are passing null in context -> query then can't find node with null slug)
onCreateNode 32a0c299-deff-5812-894a-6df20be75e2d >>> Mdx /post/tile-layouts/
createMdxNode ADD_FIELD_TO_NODE finish 32a0c299-deff-5812-894a-6df20be75e2d >>> Mdx
running query for /post/tile-layouts
running query for /
running query for Finished /post/tile-layouts
running query for Finished /
```
after the change (we run `createPages` after all processing was done):
```
info changed file at D:\dev\i7300\src\post\tile-layouts.mdx
deleting createMdxNode 32a0c299-deff-5812-894a-6df20be75e2d >>> Mdx
deleting createMdxNode finish 32a0c299-deff-5812-894a-6df20be75e2d >>> Mdx
createMdxNode 32a0c299-deff-5812-894a-6df20be75e2d >>> Mdx
createMdxNode finish 32a0c299-deff-5812-894a-6df20be75e2d >>> Mdx
onCreateNode 32a0c299-deff-5812-894a-6df20be75e2d >>> Mdx /post/tile-layouts/
createMdxNode ADD_FIELD_TO_NODE finish 32a0c299-deff-5812-894a-6df20be75e2d >>> Mdx
API_RUNNING_QUEUE_EMPTY create pages
createPages
running query for /post/tile-layouts
running query for /
running query for Finished /post/tile-layouts
running query for Finished /
```

PR #7375, opened 2018-08-16
fix deletePage wiping query text for all other pages that uses same template

pieh:
`gatsby-plugin-remove-trailing-slashes` plugin deletes and recreates pages and it's causing problems because we currently clear the query text when we deletePage

this adds checking if we have any pages using template before clearing query text

fixes #7314

Issue #7300, opened 2018-08-14
[v2] GraphQL race condition


geddski:
## Description
Gatsby compiler is working fine but then browser throws an error because the GraphQL query data is null. I'm using `gatsby-mdx` and originally thought it was an [issue there](https://github.com/ChristopherBiscardi/gatsby-mdx/issues/57).

This error happens 90% of the time when I edit a markdown/mdx file. A simple page refresh doesn't fix it. The only way to fix the page again is to *change* the `gatsby-config.js` and restart `gatsby develop`. This is making switching to Gatsby very painful.

### Steps to reproduce
- create a starter Gatsby v2 project
- get it working with `gatsby-mdx`
- put a ton of images/gifs/videos into your static folder (not sure if that's what's causing this but worth mentioning as I have in my blog project)
- make an `.mdx` post that contains a React component
- edit said post
- wish you were never born

### Environment

 System:
    OS: macOS High Sierra 10.13.4
    CPU: x64 Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz
    Shell: 2.5.0 - /usr/local/bin/fish
  Binaries:
    Node: 10.8.0 - /usr/local/bin/node
    Yarn: 1.9.4 - /usr/local/bin/yarn
    npm: 6.2.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 68.0.3440.106
    Firefox: 61.0.1
    Safari: 11.1
  npmPackages:
    gatsby: next => 2.0.0-beta.99
    gatsby-mdx: 0.0.8 => 0.0.8
    gatsby-plugin-react-helmet: next => 3.0.0-beta.4
    gatsby-plugin-styled-components: next => 3.0.0-beta.3
    gatsby-source-filesystem: next => 2.0.1-beta.10
    gatsby-transformer-remark: next => 2.1.1-beta.6
  npmGlobalPackages:
    gatsby-cli: 2.0.0-beta.13

### File contents

`gatsby-config.js`:
```js
module.exports = {
  siteMetadata: {
    title: 'Geddski',
  },
  plugins: [
    'gatsby-plugin-react-helmet',
    'gatsby-plugin-styled-components',
    {
      resolve: `gatsby-source-filesystem`,
      options: {
        name: `src`,
        path: `${__dirname}/src/`,
      },
    },
    {
      resolve: 'gatsby-mdx',
      options: {
        decks: []
      },
    },
  ],
}
```

`package.json`:
```json
{
[...]

PR #11831, opened 2019-02-17
fix: check for dirty pages when nodes are deleted (so queries are ru-run and data is removed from pages)

KyleAMathews:
Bug reported @ https://twitter.com/gregwhitworth/status/1096882065943883777?s=19

### DIFF
--- a/packages/gatsby/src/bootstrap/page-hot-reloader.js
+++ b/packages/gatsby/src/bootstrap/page-hot-reloader.js
@@ -46,7 +46,7 @@ const runCreatePages = async () => {
     })
     .map(p => p.id)
 
-  const timestamp = new Date().toJSON()
+  const timestamp = Date.now()
 
   await apiRunnerNode(`createPages`, {
     graphql,
--- a/packages/gatsby/src/bootstrap/page-hot-reloader.js
+++ b/packages/gatsby/src/bootstrap/page-hot-reloader.js
@@ -16,7 +16,7 @@ emitter.on(`CREATE_NODE`, action => {
 emitter.on(`DELETE_NODE`, action => {
   if (action.payload.internal.type !== `SitePage`) {
     pagesDirty = true
-    apiRunnerNode(`I_AM_SO_FAKE`)
+    apiRunnerNode(`FAKE_API_CALL`)
   }
 })
 
--- a/packages/gatsby/src/bootstrap/page-hot-reloader.js
+++ b/packages/gatsby/src/bootstrap/page-hot-reloader.js
@@ -16,6 +16,10 @@ emitter.on(`CREATE_NODE`, action => {
 emitter.on(`DELETE_NODE`, action => {
   if (action.payload.internal.type !== `SitePage`) {
     pagesDirty = true
+    // Make a fake API call to trigger `API_RUNNING_QUEUE_EMPTY` being called.
+    // We don't want to call runCreatePages here as there might be work in
+    // progress. So this is a safe way to make sure runCreatePages gets called
+    // at a safe time.
     apiRunnerNode(`FAKE_API_CALL`)
   }
 })
--- a/packages/gatsby/src/bootstrap/page-hot-reloader.js
+++ b/packages/gatsby/src/bootstrap/page-hot-reloader.js
@@ -16,6 +16,11 @@ emitter.on(`CREATE_NODE`, action => {
 emitter.on(`DELETE_NODE`, action => {
   if (action.payload.internal.type !== `SitePage`) {
     pagesDirty = true
+    // We call runCreatePages directly as
+    // there often isn't API calls associated with deleting nodes
+    // (especially with stateful source plugins like gatsby-source-filesystem)
+    // so API_RUNNING_QUEUE_EMPTY won't be invoked.
+    debouncedRunCreatePages()
   }
 })
 
@@ -66,6 +71,7 @@ const runCreatePages = async () => {
 
   emitter.emit(`CREATE_PAGE_END`)
 }
+const debouncedRunCreatePages = _.debounce(runCreatePages, 150)
 
 module.exports = graphqlRunner => {
   graphql = graphqlRunner
--- a/packages/gatsby/src/bootstrap/page-hot-reloader.js
+++ b/packages/gatsby/src/bootstrap/page-hot-reloader.js
@@ -16,16 +16,7 @@ emitter.on(`CREATE_NODE`, action => {
 emitter.on(`DELETE_NODE`, action => {
   if (action.payload.internal.type !== `SitePage`) {
     pagesDirty = true
-    // We call runCreatePages directly as
-    // there often isn't API calls associated with deleting nodes
-    // (especially with stateful source plugins like gatsby-source-filesystem)
-    // so API_RUNNING_QUEUE_EMPTY won't be invoked.
-    //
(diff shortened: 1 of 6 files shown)


KyleAMathews:
Ha wait. @pieh reverted this exact code last August https://github.com/gatsbyjs/gatsby/pull/7388 🤔

KyleAMathews:
API_RUNNER_QUEUE_EMPTY is definitely not being called when deleting the markdown file.

KyleAMathews:
The problem with https://github.com/gatsbyjs/gatsby/pull/7388 was it assumed that all DELETE_NODEs would lead to API calls — which in simple node deletions it doesn't.

How to detect though when _just_ a delete happens and nothing else.

KyleAMathews:
So funny solution but perhaps a good idea is we could add a `onDeleteNode` API hook — which would mean `API_RUNNING_QUEUE_EMPTY` would now always be called 😎

KyleAMathews:
I think I work around #7300 by checking that the DELETE_NODE node type is `File` as files are only deleted by `gatsby-source-filesystem` when a file is actually deleted. This wouldn't lead to any sort of infinite loop as if the file was immediately recreated, then `runCreatePages` wouldn't get called again until the new nodes are finished being created.

The problem with this is that then we'd need to add support for each new source plugin that deletes nodes. So `onDeleteNode` seems like a strong solution.

pieh:
This is problem with any stateful source plugin - in our monorepo only filesystem behaves like that - but sanity plugin (with live data refresh feature enabled) that doesn't rely on `sourceNodes` API hook for deleting nodes has exactly same problem.

This only happens for node deletion, because node updates trigger `onCreateNode` API to run which then will emit `API_RUNNER_QUEUE_EMPTY` (this doesn't happen for node deletion).

We were discussing and pairing on this with @sidharthachatterjee earlier and my idea was to "fake" API running after `DELETE_NODE` so we could just `API_RUNNER_QUEUE_EMPTY` message. This would make sure we don't trigger creating pages / query running if `deleteNode` was called during API run.

KyleAMathews:
What did you think of the `onDeleteNode` idea?

pieh:
I didn't want to add new public API that there is probably no practical use for. But both adding `onDeleteNode` and faking API run would achieve same result - only question is do we want this new public API?

KyleAMathews:
👍 yeah, it seems potentially useful but until we have concrete use cases for onDeleteNode, there's no point in adding. A special fake invocation of api-runner sounds like a great solution and can be swapped out easily for a real call if we ever add onDeleteNode.

pieh:
I think this will also fix https://github.com/gatsbyjs/gatsby/issues/10844 (which to me seems like same issue) - will validate now

pieh:
After tests with https://github.com/rexxars/gatsby-node-update-bugs it fixes the node deletion not causing rerunning createPages / rerunning queries

But I uncovered bug with page-hot-reloader not cleaning up not re-created pages - https://github.com/gatsbyjs/gatsby/blob/a420eb40451fcf7652432b53387dab0a27b64472/packages/gatsby/src/bootstrap/page-hot-reloader.js#L60 timestamp check is not working because it compares number to date string - solution is to use `Date.now()` in https://github.com/gatsbyjs/gatsby/blob/a420eb40451fcf7652432b53387dab0a27b64472/packages/gatsby/src/bootstrap/page-hot-reloader.js#L48

I'm checking out Bug #2 from the repro repo description right now - might we can kill all of those with one stone

KyleAMathews:
Great catch on updatedAt! And yikes 😱 types would be nice in core. No idea why I used `Date.now()` for page.createdAt. Doing a quick look around, it looks like 85% of dates are ISO8601 and the remaining 15% are randomly unix timestamps.

KyleAMathews:
Oh and verified that pages are added/removed on the dev 404 page — which wasn't updating smoothly for some reason.

KyleAMathews:
Oh and verified that pages are added/removed on the dev 404 page — which wasn't updating smoothly for some reason.

KyleAMathews:
Looking at @rexxars' (awesome!) repro — it looks like this fixes part of things — still seeing a problem where the title updates which causes the slug to change — for some reason the query for the new/changed page isn't run until I save the page component. I recall seeing an issue around changes to `pageContext` not triggering re-runs of GraphQL. Investigating...

KyleAMathews:
https://github.com/gatsbyjs/gatsby/pull/7375#issuecomment-464724935

KyleAMathews:
Let's get this in with these fixes. Working on more fixes for https://github.com/rexxars/gatsby-node-update-bugs in https://github.com/gatsbyjs/gatsby/pull/11897

packages/gatsby/src/bootstrap/page-hot-reloader.js
```diff
@@ -45,7 +46,7 @@ const runCreatePages = async () => {
     })
     .map(p => p.id)
 
-  const timestamp = new Date().toJSON()
+  const timestamp = Date.now()
```
wardpeet: `new Date().toJSON()` => "2019-02-20T10:47:41.468Z"
`Date.now()` => 1550659670910

I don't know how page.updatedAt looks so no idea if this is good or bad 😄

packages/gatsby/src/utils/api-runner-node.js
```diff
@@ -157,7 +157,7 @@ module.exports = async (api, args = {}, pluginSource) =>
     })
 
     // Check that the API is documented.
-    if (!apiList[api]) {
+    if (!apiList[api] && api !== `FAKE_API_CALL`) {
```
wardpeet: I can't see `FAKE_API_CALL` if you could point me in the right direction that would be awesome

KyleAMathews: Oh lemme add a comment on this. It's a bit weird :-D

packages/gatsby/src/bootstrap/page-hot-reloader.js
KyleAMathews: Yeah... for some reason I made a few types (like page) use unix timestamps instead of iso8601 :-(

So this check was failing before.

packages/gatsby/src/bootstrap/page-hot-reloader.js
```diff
@@ -16,6 +16,7 @@ emitter.on(`CREATE_NODE`, action => {
 emitter.on(`DELETE_NODE`, action => {
   if (action.payload.internal.type !== `SitePage`) {
     pagesDirty = true
+    apiRunnerNode(`FAKE_API_CALL`)
```
wardpeet: I missed this one

packages/gatsby/src/utils/api-runner-node.js
wardpeet: I missed https://github.com/gatsbyjs/gatsby/pull/11831/files#diff-deed9a141e906a882e5f3b2d2c4e21aeR19 my bad :) it makes sense

KyleAMathews:
Thanks for the review!

KyleAMathews:
Do you mean we _need_ to fix that warning? Like the new warning is popping up while developing a site?

KyleAMathews:
And you create a new page?

wardpeet:
no it's fixed :) Else I wouldn't approve 😄 

When removing a blog post and when I try to rollback the blogpost I got an error. Your fix also fixes this.

Issue #11424, opened 2019-01-30
Plugins search a11y


JacobBlomgren:
First of all, the website is pretty nice a11y-wise, but I noted a couple of things with the plugins search. 

1. The plugin search doesn't support keyboard navigation (up and down arrow keys), even though the main (nav) search does. As I understand it, it's Algolia that provides that search bar, but I think this behavior should be unified. At least I expected it to be. Github and others have this too and I'd argue that it's convenient for everyone. It should be noted though that the plugin search has good tab focus behavior, so it is usable already. This is more of an improvement.
2. While I am at it, there are some minor things that can be polished: 
  * Neither the search submit button nor the reset button show visible focus. Also these aren't there in the main search, so maybe this should be unified too?
  * There is visual structural info in each search result that is lost, i.e. header and description. Each result is a link, and putting a heading element inside a link is a bit off, admittedly. However, it's valid html and I would argue for it.
  * Finally, the visual information of the downloads icon is lost and a screenreader just reads “314.6k”. I would prefer a screen reader only “314600 downloads”, if we're ok with that a11y pattern.

PR #11836, opened 2019-02-17
docs(www): plugin search a11y improvements

JacobBlomgren:
Fixes some a11y issues with the plugin search raised in #11424, which got @marcysutton’s thumps up.

However, I abandoned the main issue, keyboard navigation, because after much investigation, I realized that all searches with keyboard navigation I could find are drop down searches. Likewise, any react solutions, e.g. Downshift, assume that the search is a drop down or so called combo box. So I got really hesitant, maybe it shouldn’t have keyboard nav after all… On top of that I couldn't get Downshift (the de facto react solution for this) to play nicely with the `InfiniteHits` from instantsearch because it needs an index to keep track of the highlighted result. So basically, I got demoralized. If anyone has any feedback that’d be much appreciated!


marcysutton:
Hey @JacobBlomgren, thanks for working on this! By "keyboard navigation", do you mean using the arrow keys instead of the TAB key to navigate through the list of plugins? Because they use anchor tags, making that change would require some significant refactoring. They do require tabbing all the way through to get past that list, though, so one quick fix would be to add a skip link that takes you past the plugins list when activated. You could show it on focus, like the "Skip to Main content" link at the top (but absolutely positioned at the top of the plugin search list).

One other thing I noticed was there's a visually hidden submit button after "Search Gatsby Library" that acts as an extra invisible tab stop for sighted keyboard users. Perhaps this button could be shown on focus, or removed entirely? It doesn't seem to do much as it stands right now.  Here's the code I used to find it, by pasting in the browser console:

```document.body.addEventListener('focusin', function(e){ console.log(e.target); })```

Gatsby plugin search open in Chrome devtools, with a focusin listener in the console

If you're interested, a third improvement would be to add focus management for the list of plugins. Currently when you activate one of those links with the ENTER key, your focus remains in the same place. It would help keyboard ergonomics to take the user into the plugin content they selected (focusing on the h1, i.e. `h1[id=gatsby-source-filesystem]`. Does that make sense?

JacobBlomgren:
I detailed my reasoning in #11424 but yes, arrow keys is what I meant. Since the main search in the nav has it, I expected it here too, essentially. But the skip link you suggested sounds like an easy win to start with. The search button focus (and “clear search” that will appear once you start typing) is fixed with this PR already :) And the third improvement is also something I could tackle!

JacobBlomgren:
@marcysutton I added the skip link. I got stuck though with the focus management because I don’t know how to give a `h1` focus. As I understand it, this would also be hard because reach router focuses the body after clicking a link, as `document.activeElement` will tell you. But maybe that’s achievable by doing something on mounting. Otherwise I think the PR is finished!

marcysutton:
@JacobBlomgren I'll file an issue to look into focus management in general, since after doing some testing I'm not sure the current behavior is what we want. 

Thanks for working on this!

PR #11844, opened 2019-02-17
docs(gatsby): Removing `pathname` of `wrapRootElement` SSR (jsdocs)

NickCis:
## Description

This PR removes the property `pathname` in the docs to the `wrapRootElement` function [as it isn't passed](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/cache-dir/static-entry.js#L145)


wardpeet:
Sweet @NickCis this is some forgotten cleanup.

PR #11855, opened 2019-02-18
[schema] Fix list inference

stefanprobst:
Don't re-wrap with arrays earlier, so we can compare `arrays===lists` when tc.hasField(key)


freiksenet:
Merged this directly.

PR #11739, opened 2019-02-13
docs(gatsby-source-wordpress): Remove lodash + bluebird + fix Promise hell

DylanTackoor:
## Description

This commit:
- Removes lodash in favor of Array.forEach
- Bluebird in favor of the native Promise object
- Avoids nested thenables by using Promise.all()

I've been using this code in my builds and it's been working fine for whatever that worth :)

PR #11868, opened 2019-02-18
fix(gatsby-source-wordpress): Sync sample w/docs

DylanTackoor:
## Description

Updates example site gatsby-node.js to match documentation.

## Related Issues

Previously implemented in #11739.

### DIFF
--- a/examples/using-wordpress/gatsby-node.js
+++ b/examples/using-wordpress/gatsby-node.js
@@ -1,5 +1,3 @@
-const _ = require(`lodash`)
-const Promise = require(`bluebird`)
 const path = require(`path`)
 const slash = require(`slash`)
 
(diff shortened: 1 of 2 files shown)


examples/using-wordpress/gatsby-node.js
```diff
@@ -1,108 +1,87 @@
...
+  // queries against the local WordPress graphql schema. Think of
+  // it like the site has a built-in database constructed
+  // from the fetched data that you can run queries against.
+  const result = await graphql(`
+    {
+      allWordpressPage {
+        edges {
+          node {
+            id
+            link
```
DSchau: This should be slug right? _Or_ if it is link it should match the iteration in the forEach later when the node content is sourced--currently we're grabbing slug.

examples/using-wordpress/gatsby-node.js
```diff
@@ -1,108 +1,87 @@
...
-      .then(result => {
-        if (result.errors) {
-          console.log(result.errors)
-          reject(result.errors)
+      }
+      allWordpressPost {
+        edges {
+          node {
+            id
+            link
```
DSchau: Same here.

DylanTackoor:
@DSchau anything else this needs?

examples/using-wordpress/gatsby-node.js
```diff
@@ -1,108 +1,87 @@
...
-            edges {
-              node {
-                id
-                slug
-                status
-                template
-              }
-            }
+  // The “graphql” function allows us to run arbitrary
+  // queries against the local WordPress graphql schema. Think of
```
DSchau:
```suggestion
  // queries against the local Gatsby GraphQL schema. Think of
```

examples/using-wordpress/gatsby-node.js
```diff
@@ -1,108 +1,87 @@
...
-            })
-          })
-          resolve()
-        })
-      })
-    // ==== END POSTS ====
+  // Access query results via object destructuring
+  const { allWordpressPage, allWordpressPost } = result.data
+
+  // Create Page pages.
```
DSchau:
```suggestion
```

Not really needed--the code is sorta self-documenting here, right?

DylanTackoor:
Done :)

examples/using-wordpress/gatsby-node.js
wardpeet: @DylanTackoor this one is still here

DylanTackoor: @wardpeet Whoops! Got it!

sidharthachatterjee:
Published in gatsby-source-wordpress@3.0.42

PR #11874, opened 2019-02-18
Added gatsby-transformer-yaml info to Documentation about markdown

ArcanisCz:
Came to this page from https://www.gatsbyjs.org/docs/sourcing-from-netlify-cms/#processing-netlify-cms-output-with-gatsby and was confused why gatsby-transformer-remark cannot parse my files. After while i discovered, it parses only .md files, but files from Netlify CMS are .yaml, for which i need different plugin!

I dont know how exactly incorporate that link into documentation, but it definitely needs to mention gatsby-transformer-yaml somehow.

### DIFF
--- a/docs/docs/adding-markdown-pages.md
+++ b/docs/docs/adding-markdown-pages.md
@@ -41,7 +41,7 @@ Now that we've "sourced" the markdown files from the filesystem, we can now "tra
 
 ### Transforming markdown — `gatsby-transformer-remark`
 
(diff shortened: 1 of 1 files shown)


KyleAMathews:
Thanks for the PR!

I don't think we want to make this change however. This page is only about adding support for markdown. Mentioning that other transforms exist for other file types feels a bit off-topic and non sequitur-ish.

PR #11880, opened 2019-02-18
[schema] Add node model tests

stefanprobst:
This is for #11480 

* Add node model tests
* Minor fixes
* Also minor stuff for abstract types (=> comment inline)

The failing unit test in CI is because we need #11857

### DIFF
--- a/packages/gatsby/src/schema/__tests__/node-model.js
+++ b/packages/gatsby/src/schema/__tests__/node-model.js
@@ -1,0 +1,446 @@
+const { store } = require(`../../redux`)
+const nodeStore = require(`../../db/nodes`)
+require(`../../db/__tests__/fixtures/ensure-loki`)()
+const { LocalNodeModel } = require(`../node-model`)
+const { build } = require(`..`)
+
+const nodes = [
+  {
+    id: `person1`,
+    parent: null,
+    children: [],
+    internal: { type: `Author` },
+    name: `Person1`,
+    email: `person1@example.com`,
+  },
+  {
+    id: `person2`,
+    parent: null,
+    children: [],
+    internal: { type: `Contributor` },
+    name: `Person2`,
+    email: `person2@example.com`,
+  },
+  {
+    id: `person3`,
+    parent: null,
+    children: [],
+    internal: { type: `Author` },
+    name: `Person3`,
+    email: null,
+  },
+  {
+    id: `post1`,
+    parent: `file1`,
+    children: [],
+    internal: { type: `Post` },
+    frontmatter: {
+      authors: [`person1`],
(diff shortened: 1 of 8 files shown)


packages/gatsby/src/schema/node-model.js
```diff
@@ -89,8 +106,18 @@ class LocalNodeModel {
   async runQuery(args, pageDependencies) {
     const { query, firstOnly, type } = args || {}
 
+    // We don't support querying union types (yet?), because the combined types
+    // need not have any fields in common.
```
stefanprobst: We can rethink not supporting union types in `runQuery` later, but it should be part of refactoring resolving nodes and run-sift (e.g. we cannot call `getFields()` on a union type there). Note that `getAllNodes` etc. *do* support both unions and interfaces.

packages/gatsby/src/schema/schema.js
```diff
@@ -117,9 +117,21 @@ const processTypeComposer = async ({
 
 const addTypes = ({ schemaComposer, types, parentSpan }) => {
   types.forEach(typeOrTypeDef => {
-    console.log(typeOrTypeDef)
     if (typeof typeOrTypeDef === `string`) {
-      schemaComposer.addTypeDefs(typeOrTypeDef)
+      const addedTypes = schemaComposer.addTypeDefs(typeOrTypeDef)
+      addedTypes.forEach(type => {
+        const { GraphQLInterfaceType, GraphQLUnionType } = require(`graphql`)
```
stefanprobst: This might not be the best place to do this -- I feel we now have multiple places where we do some kind of initial processing of types? But we can take a second look at this later.
I added `addSchemaMustHaveType` because in the tests the custom abstract types from SDL would be stripped from the built schema otherwise.
And `setResolveType` is similarly needed -- but this means that we only allow custom abstract types that combine types that implement Node. Is that restriction ok? Otherwise we'd have to provide some other way to add a type resolver, but I'm not sureI'd want that.

packages/gatsby/src/schema/__tests__/node-model.js
```diff
@@ -0,0 +1,446 @@
...
+      const firstOnly = false
+      const result = await nodeModel.runQuery({ query, firstOnly, type })
+      expect(result.length).toBe(2)
+      expect(result[0].id).toBe(`file1`)
+      expect(result[1].id).toBe(`file3`)
+    })
+  })
+
+  describe(`findRootNodeAncestor`, () => {
+    it(`returns an object's top most ancestor node`, () => {
```
stefanprobst: I'm not 100% clear why finding the top most ancestor is default, as opposed to finding an object's parent node (which is what you'd need most of the time)?

PR #11883, opened 2019-02-19
feat(gatsby): show warning if createPage was called after createPages API finished

pieh:
This adds warning if `createPage` was called after gatsby *thought* that createPages already completed.

This can happen most commonly when calling async functions and not returning promise / using `done` callback - for example:

```js
const path = require(`path`);

exports.createPages = ({ actions, graphql }) => {
  actions.createPage({
    path: `/on-time`,
    component: path.resolve(`./src/templates/index.js`)
  });

  graphql(`
    {
      site {
        siteMetadata {
          title
        }
      }
    }
  `).then(() => {
    actions.createPage({
      path: `/way-too-late`,
      component: path.resolve(`./src/templates/index.js`)
    });
  });
};
```
Promise is not returned - and it's very easy to miss and currently this can lead to very confusing errors like:

Scenario 1: `createPage` was called after we written out `pages.json`:
```
error Building static HTML failed for path "/some-path"

See our docs page on debugging HTML builds for help https://gatsby.app/debug-html

  114 | 
  115 |   let dataAndContext = {}
> 116 |   if (page.jsonName in dataPaths) {
      |            ^
  117 |     const pathToJsonData = `../public/` + dataPaths[page.jsonName]
  118 |     try {
  119 |       dataAndContext = JSON.parse(


  WebpackError: TypeError: Cannot read property 'jsonName' of undefined
```

Scenario 2: `createPage` was called after we run queries:
```
error Building static HTML failed for path "/some-path"

See our docs page on debugging HTML builds for help https://gatsby.app/debug-html

  15 | import {get} from 'lodash'
  16 | 
> 17 | export default function Post({data: {site, mdx}}) {
     |                                      ^
  18 |   const {
  19 |     isWriting,
  20 |     author = config.author,


  WebpackError: TypeError: Cannot read property 'site' of undefined
```

This change adds warning like this:
screenshot 2019-02-19 at 03 53 06

TO-DO:
 - [ ] Write docs for https://gatsby.app/sync-actions showing common problems and how to fix them.

AFAIK we currently only have `Async plugins` section in https://www.gatsbyjs.org/docs/node-apis/ that explain this, but as with any docs - it's very easy for users to miss this.

### DIFF
--- a/docs/docs/debugging-async-lifecycles.md
+++ b/docs/docs/debugging-async-lifecycles.md
@@ -4,7 +4,7 @@ title: Debugging Asynchronous Lifecycle Methods
 
 Various lifecycle methods (see: [Gatsby Node APIs](/docs/node-apis/)) within Gatsby are presumed to be asynchronous. In other words, these methods can _eventually_ resolve to a value and this value is a [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise). We wait for the `Promise` to resolve, and then mark the lifecycle method as completed when it does.
 
-In the context of Gatsby, this means that if you are invoking async functionality (e.g. data requests, `graphql` calls, etc.) and not correctly returning the Promise an internal issue can arise where the result of those call(s) happens _after_ the lifecycle method has already been marked as completed. Let's consider an example:
+In the context of Gatsby, this means that if you are invoking asynchronous functionality (e.g. data requests, `graphql` calls, etc.) and not correctly returning the Promise an internal issue can arise where the result of those call(s) happens _after_ the lifecycle method has already been marked as completed. Let's consider an example:
 
 ```js:title=gatsby-node.js
 exports.createPages = async function({ actions, graphql }) {
@@ -35,7 +35,7 @@ exports.createPages = async function({ actions, graphql }) {
 }
 ```
 
-Can you spot the error? In this case, an asyncronous action (`graphql`) was invoked but this async action was neither `return`ed nor `await`ed from `createPages`. This means that the lifecycle method will be marked as complete before it's actually completed--which leads to missing data errors and other hard-to-debug errors.
+Can you spot the error? In this case, an asynchronous action (`graphql`) was invoked but this asynchronous action was neither `return`ed nor `await`ed from `createPages`. This means that the lifecycle method will be marked as complete before it's actually completed--which leads to missing data errors and other hard-to-debug errors.
 
 The fix is surprisingly simple--just one line to change!
 
--- a/docs/docs/debugging-async-lifecycles.md
+++ b/docs/docs/debugging-async-lifecycles.md
@@ -1,0 +1,95 @@
+---
+title: Debugging Asynchronous Lifecycles
+---
+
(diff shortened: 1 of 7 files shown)


packages/gatsby/src/utils/api-runner-node.js
```diff
@@ -94,6 +96,44 @@ const runAPI = (plugin, api, args) => {
 
     const cache = getCache(plugin.name)
 
+    // Ideally this would be more abstracted and applied to more situations, but right now
+    // this can be potentially breaking so targeting `createPages` API and `createPage` action
+    let apiFinished = false
+    if (api === `createPages`) {
+      let alreadyDisplayed = false
+      const createPageAction = doubleBoundActionCreators.createPage
+      doubleBoundActionCreators.createPage = (...args) => {
```
pieh: trippleBound!

www/src/pages/docs/node-apis.js
```diff
@@ -34,7 +34,12 @@ class NodeAPIDocs extends React.Component {
             calling remote APIs, etc.) you must either return a promise or use
             the callback passed to the 3rd argument. Gatsby needs to know when
             plugins are finished as some APIs, to work correctly, require
-            previous APIs to be complete first.
+            previous APIs to be complete first. See{` `}
```
DSchau: Tangential to this--@pieh what do you think if we remove the callback example in the node-apis page?

packages/gatsby/src/utils/api-runner-node.js
```diff
@@ -125,13 +165,18 @@ const runAPI = (plugin, api, args) => {
...
           callback(err, val)
+          apiFinished = true
         }
         gatsbyNode[api](...apiCallArgs, cb)
       })
     } else {
       const result = gatsbyNode[api](...apiCallArgs)
       pluginSpan.finish()
-      return Promise.resolve(result)
+      const resultPromise = Promise.resolve(result)
```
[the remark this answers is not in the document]
pieh: Yeah, will adjust - remnant of initial code that I didn't clean up

docs/docs/debugging-async-lifecycles.md
```diff
@@ -0,0 +1,95 @@
+---
+title: Debugging Asynchronous Lifecycle Methods
+---
+
+Various lifecycle methods (see: [Gatsby Node APIs](/docs/node-apis/)) within Gatsby are presumed to be asynchronous. In other words, these methods can _eventually_ resolve to a value and this value is a [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise). We wait for the `Promise` to resolve, and then mark the lifecycle method as completed when it does.
```
pieh: This can be misleading - we don't presume that APIs are async, but we allow them to be - and if they are they need to return Promise.

`createPages` can be sync:
```js
exports.createPages = ({ actions }) => {
  const data = require('./some-data.json')
  data.forEach(datum => actions.createPage({ path: datum.path, component: <some_component_path> }))
}
```
is correct and will work just fine.

DSchau: I don't think this is misleading necessarily. They _are_ presumed to be async.

DSchau: How should I tweak the language though?

pieh: I think we should drop my comment to ship this and iterate if needed.

PR #11884, opened 2019-02-19
feat(gatsby-plugin-sharp): add defaultQuality option 

### LINKED ISSUE #4873: Add an option to configure default image quality for ImageSharp queries
Currently, the default quality of ImageSharp-generated images are too low for most of the projects I participate in. I would like to configure a global default for ImageSharp query parameters, because it may become tedious to always specify `quality: 92` in every single image query.

Issue #3837 is related, but it's about specifying a default locale for Moment.js date queries.

sbardian:
## Description
Enable passing a `defaultQuality` option to `gatsby-plugin-sharp`.  Should also work with `gatsby-transformer-sharp`.  Could easily be expanded to allow passing other defaults, but only pulling out `defaultQuality` in case I'm way off base on the approach/missing something/etc. 

Thanks!

## Related Issues
Fixes #4873

### DIFF
--- a/packages/gatsby-plugin-sharp/README.md
+++ b/packages/gatsby-plugin-sharp/README.md
@@ -32,6 +32,7 @@ plugins: [
     options: {
       useMozJpeg: false,
       stripMetadata: true,
+      defaultQuality: 75,
     },
   },
 ]
@@ -248,6 +249,10 @@ fixed(
   tracedSVG
 }
 ```
+
+### Setting a default quality
+
+You can set a default quality by passing the `defaultQuality` option to the plugin.
 
 ### Using MozJPEG
 
--- a/packages/gatsby-plugin-sharp/README.md
+++ b/packages/gatsby-plugin-sharp/README.md
@@ -252,7 +252,7 @@ fixed(
 
 ### Setting a default quality
 
-You can set a default quality by passing the `defaultQuality` option to the plugin.
+You can pass a default image quality to `sharp` by setting the `defaultQuality` option.
 
 ### Using MozJPEG
 
--- a/packages/gatsby-plugin-sharp/src/index.js
+++ b/packages/gatsby-plugin-sharp/src/index.js
@@ -96,12 +96,8 @@ const generalArgs = {
   sizeByPixelDensity: false,
 }
 
-const healOptions = (pluginOptions, args, defaultArgs) => {
-  const quality =
-    pluginOptions.defaultQuality && args.quality === 50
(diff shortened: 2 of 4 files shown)


packages/gatsby-plugin-sharp/src/index.js
```diff
@@ -95,8 +96,12 @@ const generalArgs = {
   sizeByPixelDensity: false,
 }
 
-const healOptions = (args, defaultArgs) => {
-  let options = _.defaults({}, args, defaultArgs, generalArgs)
+const healOptions = (pluginOptions, args, defaultArgs) => {
+  const quality =
+    pluginOptions.defaultQuality && args.quality === 50
```
sbardian: Don't really like this, but because `gatsby-transformer-sharp` has a default `arg` of `50`, you will never get the `defaultQuality` option applied.  I could update the PR to remove the default `arg` from `gatsby-transformer-sharp`.  Then this would be simplified to just: 

```javascript
let options = _.defaults({}, {...pluginOptions}, args, defaultArgs, generalArgs);
```

sidharthachatterjee:
Looks great overall @sbardian 

Thank you!

packages/gatsby-plugin-sharp/README.md
```diff
@@ -249,6 +250,10 @@ fixed(
 }
 ```
 
+### Setting a default quality
+
+You can set a default quality by passing the `defaultQuality` option to the plugin.
```
sidharthachatterjee:
```suggestion
You can pass a default image quality to `sharp` by setting the `defaultQuality` option.
```

sbardian:
> Just added a small comment. Also, worth adding some tests?
> 
>     * the default is set correctly when nothing is passed in
> 
>     * the quality is overridden correctly when the option is passed in
> 
> 
> What do you think?

As it is, quality is not returned by any of the exposed functions (base64, fluid, fixed, queueImageResizing, getImageSize).  So, the snapshot tests are as good as any I could do now, and I don't think they tell us anything about quality?  I suppose we could add `quality` as a returned value to each?

sidharthachatterjee:
@sbardian Yeah, you're right. I suppose we can add robust tests for this in later. Just cloning this and testing it out locally!

sidharthachatterjee:
Successfully published:
 - gatsby-plugin-sharp@2.0.21
 - gatsby-source-contentful@2.0.30
 - gatsby-transformer-sharp@2.1.14
 - gatsby-transformer-sqip@2.0.17

PR #11893, opened 2019-02-19
fix(www): Change tag of one starter

LekoArts:
`bs-css` is kinda a one-off tag, others were grouped under Styling:CSS-in-JS in the past @sidharthachatterjee


sidharthachatterjee:
Thank you, @LekoArts 👍

PR #4555, opened 2018-03-16
[v2] improve how we load GraphQL query results in development/production (aka Ludicrous Mode)

m-allanson:
The beginnings of a custom JSON loader for query results. In `develop` mode this pushes query results straight to the browser via websockets:

kapture 2018-03-15 at 21 10 55

PR #11897, opened 2019-02-19
feat(gatsby): Move page component state & side effect handling to xstate

KyleAMathews:
Internally within Gatsby, some of our most complex logic is around creating and managing page components — there's quite a few states a page component can be in e.g. while bootstrapping, extracting queries, running queries for its pages, etc.

The code for handling this has been some of the most annoying and bug prone code to work in.

XState has been on list of things to try in core for quite a while as it's perfect for these sorts of things. @m-allanson did a small PR adding it for ludicrious mode last year https://github.com/gatsbyjs/gatsby/pull/4555 and it's been problem free there.

### Visualization
Screenshot 2019-03-22 18 49 42

## TODOs
- [x] extract machine to separate directory
- [x] add tests for critical flows

## Follow up questions/work
- Model individual pages
  - ~Fix bug #2b in https://github.com/rexxars/gatsby-node-update-bugs the old page is now deleted due to https://github.com/gatsbyjs/gatsby/pull/11831 — to fix running the query again, we need to model new pages~
  - Fix this bug with `pageContext` changing not triggering rebuilds
  - Track errors per page
  - pages run their own queries instead of the parent pageComponent handling that — pageComponent tells them the query has changed and they then message the query runner to re-run their query.
- how to model static queries? It looks like they're getting run every time their component changes.
- would xstate be helpful for data-driven state changes? Presumably as a data change could happen at the same time as a query is being extracted. So we should then send action when a page is dirtied.
- it seems like handling error reporting in machines would work really well
- how do we want to make private actions truly private? Move to different file? Mark with underscore and filter out from passing to plugins?


packages/gatsby/src/internal-plugins/query-runner/graphql-errors.js
```diff
@@ -55,21 +55,22 @@ function formatError(message: string, filePath: string, codeFrame: string) {
 }
 
 function extractError(error: Error): { message: string, docName: string } {
-  const docRegex = /Invariant Violation: (RelayParser|GraphQLParser): (.*). Source: document `(.*)` file:/g
```
KyleAMathews: This regex broke when we moved to Relay Compiler v2.

packages/gatsby/src/redux/actions.js
```diff
@@ -1142,6 +1142,65 @@ actions.addThirdPartySchema = (
   }
 }
 
+/**
+ *
+ * Report that a query has been extracted from a component. Used by
+ * query-compilier.js.
+ *
+ * @param {Object} $0
```
KyleAMathews: I want to use actions more for internal communication but don't want them all to show up here (https://www.gatsbyjs.org/docs/actions/)

Thinking we could start marking actions as `private` to get around this?

packages/gatsby/src/redux/reducers/components.js
```diff
@@ -1,29 +1,164 @@
...
+      }),
+      setBootstrapFinished: assign({
+        isInBootstrap: false,
+      }),
+    },
+  }
+)
+
+const services = new Map()
+let programStatus = `BOOTSTRAPPING`
```
wardpeet: we probably want to move to an enum.

packages/gatsby/src/internal-plugins/internal-data-bridge/gatsby-node.js
```diff
@@ -157,3 +157,7 @@ emitter.on(`DELETE_PAGE`, action => {
   const node = getNode(nodeId)
   boundActionCreators.deleteNode({ node })
 })
+
+emitter.on(`BOOTSTRAP_FINISHED`, () =>
+  boundActionCreators.setProgramStatus(`BOOTSTRAP_FINISHED`)
```
stefanprobst: Should we move this [here](https://github.com/gatsbyjs/gatsby/blob/0df55f61587fe7f86e48bba5599d3fdc5d59b19b/packages/gatsby/src/bootstrap/index.js#L523)?

packages/gatsby/src/internal-plugins/query-runner/page-query-runner.js
```diff
@@ -42,13 +42,14 @@ const runQueries = async () => {
 
   // Find ids without data dependencies (i.e. no queries have been run for
   // them before) and run them.
-  const cleanIds = findIdsWithoutDataDependencies()
```
KyleAMathews: probably need to restore this? For now

packages/gatsby/src/internal-plugins/internal-data-bridge/gatsby-node.js
KyleAMathews: Yeah, that'd make more sense 👍

KyleAMathews:
@gatsbyjs/core this is ready for review

KyleAMathews:
All the test errors should be from xstate dependency not existing

packages/gatsby/src/redux/actions.js
stefanprobst: I'm all for using actions more (and less event emitting). Can we move them to the `actions` folder that already exists (but only has the page dependencies stuff), and export a subset that can be used in node APIs?

packages/gatsby/src/internal-plugins/query-runner/page-query-runner.js
```diff
@@ -116,10 +112,7 @@ const findIdsWithoutDataDependencies = () => {
   // Get list of paths not already tracked and run the queries for these
   // paths.
   const notTrackedIds = _.difference(
-    [
-      ...Array.from(state.pages.values(), p => p.path),
-      ...[...state.staticQueryComponents.values()].map(c => c.jsonName),
-    ],
+    [...[...state.staticQueryComponents.values()].map(c => c.jsonName)],
```
wardpeet: do we want to put this on multiple lines into some temporary variables? It might be more readable.

packages/gatsby/src/redux/reducers/components.js
```diff
@@ -1,29 +1,128 @@
...
+          // an Action call which Redux squawks about.
+          setTimeout(() => {
+            runQueryForPage(action.payload.path)
+          }, 0)
+        }
+      }
+
+      state.set(
+        action.payload.componentPath,
+        _.merge(
```
wardpeet:
```suggestion
        Object.assign(
```

packages/gatsby/src/redux/reducers/components.js
```diff
@@ -1,29 +1,128 @@
...
+    case `PAGE_COMPONENT_CHANGED`:
+    case `QUERY_EXTRACTION_BABEL_ERROR`:
+    case `QUERY_EXTRACTION_GRAPHQL_ERROR`: {
+      action.payload.componentPath = normalize(action.payload.componentPath)
+      const service = services.get(action.payload.componentPath)
+      service.send({
+        type: action.type,
+        ...action.payload,
       })
+      services.set(action.payload.componentPath, service)
```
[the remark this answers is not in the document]
wardpeet: it's not, you get a reference of the instance anyway so you don't need to reset it.

packages/gatsby/src/redux/actions.js
wardpeet: I like just using pick :)

packages/gatsby/src/internal-plugins/query-runner/query-watcher.js
```diff
@@ -132,22 +101,42 @@ const handleQuery = (
...
+        // show a warning about having a query in a non-page component.
+      } else if (isFirstRun && !snapshot.components.has(component)) {
         report.warn(
           `The GraphQL query in the non-page component "${component}" will not be run.`
         )
         queriesWillNotRun = true
       }
     })
 
+    // TODO restore this logic for page components.
```
wardpeet: will this cause a regression? (should we address it in this PR?)

packages/gatsby/src/redux/reducers/components.js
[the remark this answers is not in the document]
KyleAMathews: 👍 ok right, I'll remove them

packages/gatsby/src/internal-plugins/query-runner/query-watcher.js
KyleAMathews: Oh actually already did this.

KyleAMathews:
Tested (mostly) on gatsbyjs.org.

packages/gatsby/src/redux/machines/page-component.js
```diff
@@ -0,0 +1,143 @@
+const {
+  Machine,
+  actions: { assign },
```
wardpeet:
```suggestion
  assign,
```
docs write to directly[ import from xstate](https://xstate.js.org/docs/guides/context.html#updating-context-with-assign)

It's the same thing so does not make a big difference
https://github.com/davidkpiano/xstate/blob/master/src/index.ts#L21-L46

packages/gatsby/src/internal-plugins/query-runner/page-query-runner.js
```diff
@@ -45,11 +49,25 @@ const runQueries = async () => {
...
-  ])
+  let pathnamesToRun = _.uniq([...dirtyIds, ...cleanIds])
+
+  // If this is the initial run, remove pathnames from `runQueriesForPathnamesQueue`
+  // if they're also not in the dirtyIds or cleanIds.
+  //
+  // We do this because the page component reducer/machine always
+  // adds pages to runQueriesForPathnamesQueue but during bootstrap
+  // we may not want to run those page queries if their data hasn't
+  // changed since the last time we ran Gatsby.
```
pieh: This introduces regression:
`gatsby build` -> edit query -> `gatsby build` - we will not run query, but query did change

KyleAMathews: Fixed here https://github.com/gatsbyjs/gatsby/pull/11897/commits/5100484904723f30e2d4816f15645be822ac9f42

Thanks for identifying this!

pieh:
the `NEW_PAGE_CREATED` event was never emitted after https://github.com/gatsbyjs/gatsby/pull/11897/commits/5100484904723f30e2d4816f15645be822ac9f42 (so queries for new pages in https://github.com/rexxars/gatsby-node-update-bugs#bug-2---updating-nodes was potentially not run - it might have been run by "accident" if something triggered query running and clean id logic would pick this up)

I've re-added emitting `NEW_PAGE_CREATED` event (the setPage action handles not actually querying it before bootstrap "finish"). I've also adjusted existing tests + added 2 to cover `NEW_PAGE_CREATED` handlers

DSchau:
@pieh is this ready for another set of 👀? I can test this out soon, if so.

pieh:
Yes, please!

KyleAMathews:
@dschau did you have a chance to test?

DSchau:
@KyleAMathews didn't. I'll pull this down and validate tomorrow!

KyleAMathews:
Thanks!

KyleAMathews:
I had another plane ride earlier so thought went through this PR again and thought up a manual test plan for all behaviors this code should handle correctly. These would be good to start turning into formal e2e tests as well.  Please suggest more tests I've missed!

I got through 2.v before my plane landed.

---

1. Compile during bootstrap
    1. compile all queries from clean cache run
    2. compile changed page queries after restart
    3. No queries are compiled until the appropriate place in bootstrap
2. Run queries during bootstrap
    1. Run all queries when there’s a clean cache
    2. Run all queries for pages with changed data & pages with changed queries & queries we don’t have any tracked data for
    3. No queries are run until the appropriate place in the bootstrap
    4. _[Not working for static queries — they're run twice during bootstrap and everytime queries are compiled thereafter]_ No queries are re-run (unless data/query changes) after the initial point in the bootstrap (implying that we know for sure at that point all queries which need run)
    5. _[not working if a page query isn’t run]_ Each page is tracked inside the page context
3. Compile during development
    1. Queries should be re-compiled whenever a JS file inside src/ changes
    2. Queries should re-update 
4. Run queries during development
    1. When the query changes
    2. When data relied on by a query changes
    3. When a new page added
    4. When an existing page is deleted and then re-added (the old page data dependencies should be deleted so adding the new page runs everything fresh)
5. Error states handling (how do we help people resolve errors/warnings)
    1. JS & GraphQL errors in a JS file block their page queries from running
            1. Once the error is resolved, if the query has changed, queries are run again. If it’s a JS error, queries aren’t re-run

pieh:
> Run queries during development
> 3. When a new page added

I would add sub points:
  - when new page using already existing template is added
  - when new page using new template is added

There are different code paths for those

DSchau:
@KyleAMathews @pieh didn't have as much time today as I would've liked, but got through some of these. Definitely a few issues thus far that need fixing.

I'll look into actually _fixing_ these tomorrow afternoon.

Here's what I found thus far:

## Issues

4.iv) Deleting a page seems to hard fail

```
TypeError: expected a string

- index.js:12 normalizePath
[xstate-example]/[normalize-path]/index.js:12:11
  - components.js:131 module.exports
    [xstate-example]/[gatsby]/dist/redux/reducers/components.js:131:40
  
  - redux.js:451 combination
    [xstate-example]/[redux]/lib/redux.js:451:29
  
  - redux.js:211 dispatch
    [xstate-example]/[redux]/lib/redux.js:211:22
  
  - index.js:78 action
    [xstate-example]/[gatsby]/dist/redux/index.js:78:91
  
  - redux.js:468 
    [xstate-example]/[redux]/lib/redux.js:468:12
  
  - api-runner-node.js:55 doubleBoundActionCreators.(anonymous function).args
    [xstate-example]/[gatsby]/dist/utils/api-runner-node.js:55:13
```

5. Recovering from an Error
    - Used `throw new Error` in a function body
    - Hot reload failed
    - Removed throw new error, error still appears

## Misc. / Possibly Unrelated

- StaticQuery cannot be changed and hot reloaded (or really used at all in the development server)
    > Error: The result of this StaticQuery could not be fetched.
- Hot reloading seems to log things at least twice?
    - This may be unrelated to this PR
	- Add a console.log in a page/template/component and it logs (at least) twice

pieh:
> 4.iv) Deleting a page seems to hard fail
> 
> * Add a page (e.g. pages/page-2.js)
> * Delete the page

Should be fixed in 36bda2e

> * StaticQuery cannot be changed and hot reloaded (or really used at all in the development server)
>   > Error: The result of this StaticQuery could not be fetched.

This is bigger one - with approach change in here with emitting `PAGE_COMPONENT_CHANGE` action and not calling to extract directly we essentially we will never trigger query runs for static queries changes (edits/adding new static queries or removing old ones). I'm not exactly sure why query extraction was moved to be called by page-component machine and why it needs to be tracked there. I will open PR against this branch so it can be reviewed separately

KyleAMathews:
Gotta another plane ride so I can work on the static query stuff.

KyleAMathews:
Checked and https://github.com/rexxars/gatsby-node-update-bugs are also fixed

jlengstorf:
Niiiice. Let me know when you've got an alpha for this and I'll run my sites against it. Super stoked to see this!

ChristopherBiscardi:
same. Happy to run the mdx examples and such against it too.

KyleAMathews:
Sweet thanks! Release `gatsby@2.1.5-alpha.72`

jlengstorf:
Got an error while building:
```
error Plugin gatsby-source-airtable returned an error


  TypeError: #<Promise> is not iterable
  
  - Function.all
  
  - gatsby-node.js:165 Object.exports.sourceNodes
    [lengstorf.com]/[gatsby-source-airtable]/gatsby-node.js:165:18
  
  - next_tick.js:81 processTicksAndRejections
    internal/process/next_tick.js:81:5
```

Offending code was here: https://github.com/jbolda/gatsby-source-airtable/blob/af5d379989edaae54c8eee2f81ee1256202faebc/gatsby-node.js#L165-L167

```
$ gatsby info

  System:
    OS: macOS 10.14.3
    CPU: x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 11.12.0 - /usr/local/bin/node
    Yarn: 1.15.2 - /usr/local/bin/yarn
    npm: 6.7.0 - /usr/local/bin/npm
  Languages:
  Browsers:
    Chrome: 72.0.3626.121
    Firefox: 63.0
    Safari: 12.0.3
  npmPackages:
    gatsby: 2.1.5-alpha.72 => 2.1.5-alpha.72+cedc761b3 
    gatsby-plugin-algolia: ^0.3.0 => 0.3.0 
    gatsby-plugin-amplitude-analytics: ^1.3.0 => 1.5.0 
    gatsby-plugin-manifest: ^2.0.6 => 2.0.24 
    gatsby-plugin-netlify: ^2.0.12 => 2.0.13 
    gatsby-plugin-offline: ^2.0.9 => 2.0.25 
    gatsby-plugin-remove-serviceworker: ^1.0.0 => 1.0.0 
    gatsby-plugin-sitemap: ^2.0.9 => 2.0.10 
    gatsby-source-airtable: ^2.0.2 => 2.0.5 
    gatsby-theme-jason-blog: ^1.0.1 => 1.0.2 
  npmGlobalPackages:
    gatsby-cli: 1.1.58
    gatsby-dev-cli: 2.0.0-alpha.80a21f04
```

Moocar:
Looking forward to this one. I've been working on a follow up PR that eliminates some of the eventing and global state by making the query runner queue constructable. I'll create a PR once this is one is merged. https://github.com/gatsbyjs/gatsby/compare/xstate-components...Moocar:query-runner-refactor?expand=1

KyleAMathews:
Tests passed — manually tested this on a number of sites + my manual checklist above.

Merging and publishing.

I'll create an issue for turning the checklist into automated tests.

DSchau:
@KyleAMathews 🎉 awesome news!

KyleAMathews:
```
Successfully published:
 - gatsby@2.2.12
```

PR #11909, opened 2019-02-19
docs(gatsby-link): Added some information regarding the limitations with the link package

Minivera:
## Description

The `<Link>` component and the `navigate` function do not allow the user to replace to a hash-link nor a query parameter while `@reach/router` allows it. This PR adds a small section in the Link API docs to let users know of possible workarounds this limitation.

This was discussed in this PR: https://github.com/gatsbyjs/gatsby/pull/11625

### DIFF
--- a/docs/docs/gatsby-link.md
+++ b/docs/docs/gatsby-link.md
@@ -357,7 +357,7 @@ You can similarly check for file downloads:
 
 [egghead]: https://egghead.io/playlists/use-gatsby-s-link-component-to-improve-site-performance-and-simplify-site-development-7ed3ddfe
 
-## Limitations when appending query parameters or hash links programmatically
+## Recommendations for programmatic, in-app navigation
 
 Neither `<Link>` not `navigate` can be used for in-route navigation with a hash or query parameter. If you need this behavior, you should either use an anchor tag or import the `@reach/router` package to make use of its `navigate` function, like so:
 
--- a/docs/docs/gatsby-link.md
+++ b/docs/docs/gatsby-link.md
@@ -359,7 +359,7 @@ You can similarly check for file downloads:
 
 ## Limitations when appending query parameters or hash links programmatically
 
-Neither `<Link>` not `navigate` can be used to add query parameters or send your user to a hash link due to their use of `withPrefix`. If you need this behavior, you should either use an anchor tag or load the `@reach/router` package to make use of its `navigate` function.
+Neither `<Link>` not `navigate` can be used for in-route navigation with a hash or query parameter. If you need this behavior, you should either use an anchor tag or import the `@reach/router` package to make use of its `navigate` function, like so:
 
 ```jsx
 import { navigate } from '@reach/router';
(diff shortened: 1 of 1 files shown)


docs/docs/gatsby-link.md
```diff
@@ -356,3 +356,19 @@ You can similarly check for file downloads:
 ```
 
 [egghead]: https://egghead.io/playlists/use-gatsby-s-link-component-to-improve-site-performance-and-simplify-site-development-7ed3ddfe
+
+## Limitations when appending query parameters or hash links programmatically
```
DSchau: General question - what about query parameters makes them untenable with the `Link` component?

This seems like a perfectly reasonable use-case, e.g.

```jsx
<Link to="/user?id=1234">Go to 1234 user</Link>
```

Is there more nuance here? Is it query param navigation on the _same_ page that doesn't work?

Minivera: Yes, it wouldn't work if you did `<Link to="?id=1234">Go to 1234 user</Link>`. For example, if I built a Tab control that used the query parameter for the current tab.

DSchau: Let's maybe dial back the language a bit then? 

Maybe something like:

```suggestion
## Recommendations for programmatic, in-app navigation
```

docs/docs/gatsby-link.md
```diff
@@ -356,3 +356,19 @@ You can similarly check for file downloads:
...
+
+Neither `<Link>` not `navigate` can be used to add query parameters or send your user to a hash link due to their use of `withPrefix`. If you need this behavior, you should either use an anchor tag or load the `@reach/router` package to make use of its `navigate` function.
```
DSchau:
```suggestion
Neither `<Link>` not `navigate` can be used for in-route navigation with a hash or query parameter. If you need this behavior, you should either use an anchor tag or import the `@reach/router` package to make use of its `navigate` function, like so:
```

Minivera:
@DSchau are you okay with approving the PR?

docs/docs/gatsby-link.md
```diff
@@ -356,3 +356,19 @@ You can similarly check for file downloads:
 ```
 
 [egghead]: https://egghead.io/playlists/use-gatsby-s-link-component-to-improve-site-performance-and-simplify-site-development-7ed3ddfe
+
+## Recommendations for programmatic, in-app navigation
+
+Neither `<Link>` not `navigate` can be used for in-route navigation with a hash or query parameter. If you need this behavior, you should either use an anchor tag or import the `@reach/router` package to make use of its `navigate` function, like so:
```
DSchau:
```suggestion
Neither `<Link>` nor `navigate` can be used for in-route navigation with a hash or query parameter. If you need this behavior, you should either use an anchor tag or import the `@reach/router` package--which Gatsby already depends upon--to make use of its `navigate` function, like so:
```

DSchau:
Left one comment, otherwise I think this is largely good to go!

Minivera:
Good catch, my eyes were reading `nor` instead of `not`

PR #11919, opened 2019-02-20
feat(themes): add gatsby-theme-blog

ChristopherBiscardi:
Initial themes are going to be used for e2e tests in a future PR so
that we can use them as we add features and fix bugs


ChristopherBiscardi:
@DSchau these happen to be published already, so I need to transfer the npm packages to the gatsby user. There shouldn't be any issues with adding them to lerna now (some people pull and play with them when testing out themes, but it's not a lot).

PR #11920, opened 2019-02-20
fix(gatsby-image): Fix issue where Safari downloads multiple resolutions

tbrannam:
## Description

Safari browser (Safari 12 / Mac OS) downloads multiple assets for a single `Img` tag.

The observed behavior is that the attributes of the `` tag are queued early in the parsing of the document.

Resolved by removing the default (non-webp) `<source>` and placing those value back into the fallback `` tag.  Additionally the order of the attributes of `` are important as well.   If `src` attributes precedes `srcset` in the `` Safari’s pre-parser will enqueue both images to download.

Finally - added a note regarding a dependency on IntersectionObserver for lazy-loading

Testing Fail-case:  https://using-gatsby-image.gatsbyjs.org/prefer-webp/ note that each image has two downloaded image sizes (also note that the markup for the cactus image is referenced twice, thus 4 images are downloaded).

### DIFF
--- a/e2e-tests/gatsby-image/src/pages/fixed.js
+++ b/e2e-tests/gatsby-image/src/pages/fixed.js
@@ -24,7 +24,7 @@ export const pageQuery = graphql`
     fruitsFixed: file(relativePath: { eq: "citrus-fruits.jpg" }) {
       childImageSharp {
         fixed(width: 500) {
-          ...GatsbyImageSharpFixed
+          ...GatsbyImageSharpFixed_withWebp
         }
       }
     }
--- a/e2e-tests/gatsby-image/src/pages/fluid.js
+++ b/e2e-tests/gatsby-image/src/pages/fluid.js
@@ -24,7 +24,7 @@ export const pageQuery = graphql`
     fruitsFluid: file(relativePath: { eq: "citrus-fruits.jpg" }) {
       childImageSharp {
         fluid(maxWidth: 500) {
-          ...GatsbyImageSharpFluid
+          ...GatsbyImageSharpFluid_withWebp
         }
       }
     }
--- a/packages/gatsby-image/README.md
+++ b/packages/gatsby-image/README.md
@@ -348,3 +348,5 @@ You will need to add it in your graphql query as is shown in the following snipp
 - Gifs can't be resized the same way as pngs and jpegs, unfortunately—if you try
   to use a gif with `gatsby-image`, it won't work. For now, the best workaround is
   to [import the gif directly](/docs/adding-images-fonts-files).
+- Lazy loading behavior is dependent on `IntersectionObserver` which is not available
+  in some fairly common browsers including Safari and IE. A polyfill is recommended.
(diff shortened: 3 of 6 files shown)


sidharthachatterjee:
@tbrannam Looks good but want to quickly clarify

I can't seem to reproduce this issue on https://using-gatsby-image.gatsbyjs.org/prefer-webp/

Maybe I'm misunderstanding? 
screenshot 2019-02-21 02 15 16

tbrannam:
Can you tell me the version of Safari and MacOS you are running?
________________________________

sidharthachatterjee:
Away from computer at the moment. 

Safari 12 on macOS Mojave

Can tell you exact minor versions a little later

tbrannam:
I'm on High Sierra still but using Safari 12.0.3.  I've spot checked Mojave on a colleagues machine and was able to reproduce the same.   Additionally we have reproduced this same issue on Safari - iPhone X 12.1.2
 
screen shot 2019-02-20 at 5 15 47 pm

sidharthachatterjee:
So strange. I still can't reproduce this. Check out versions in the screenshot. 
screenshot 2019-02-21 04 11 14

tbrannam:
This is disappointing - I cannot see a reason why we are getting different results.  I have run out of test devices to try to narrow it down. 

I've also tried in 'private window' as well as well as bypass cache in the inspector to try to vary the behavior - but they all seemed to give me the same results.

sidharthachatterjee:
@tbrannam Yeah, really strange. 

Let's keep this open and see if the others can reproduce in the mean time. cc @DSchau @pieh 

Also, is there value in merging in this anyway irrespective of this particular bug you're seeing?

pieh:
I reproduced this locally with very weird findings - for some reason I saw 2 different sizes being downloaded ... only if safari window was in certain width range - somewhere between 720 and 760 px wide (below and above that it was just grabbing single "pug" image) - it might happen in more scenarios, but I was happy enough I could reliably reproduce that and didn't search for more. Also verified that the changes in the PR fixes it, but didn't yet look through the code changes here.

tbrannam:
@pieh great callout

```
<picture>
<source 
 srcSet="/static/3b314577d3b8dd16b97a6b23c068b374/91f71/7b214c41dead6dff0f496c55d52769d2.jpg 180w,
/static/3b314577d3b8dd16b97a6b23c068b374/a4889/7b214c41dead6dff0f496c55d52769d2.jpg 360w,
/static/3b314577d3b8dd16b97a6b23c068b374/044eb/7b214c41dead6dff0f496c55d52769d2.jpg 720w,
/static/3b314577d3b8dd16b97a6b23c068b374/bd259/7b214c41dead6dff0f496c55d52769d2.jpg 1080w,
......" sizes="(max-width: 720px) 100vw, 720px" />
<img src="/static/3b314577d3b8dd16b97a6b23c068b374/044eb/7b214c41dead6dff0f496c55d52769d2.jpg"  ..... />
</picture>
```
`/static/3b314577d3b8dd16b97a6b23c068b374/044eb/7b214c41dead6dff0f496c55d52769d2.jpg` is the url for `` as well as the of `<source srcset>` 

So the failcase would only occur if the viewport fell outside that breakpoint.

sidharthachatterjee:
@tbrannam That makes sense. Merging this in.

sidharthachatterjee:
Published in gatsby-image@2.0.30

PR #11895, opened 2019-02-19
chore(gatsby-plugin-page-creator): upgrade chokidar@2.1.2 to fix ReDoS vulnerability #11847

yogeshkotadiya:
## Description
Upgraded chokidar to @2.1.2 to fix Regular Expression Denial of Service (ReDoS).

## Related Issues
Fixes #11847

PR #11922, opened 2019-02-20
chore(gatsby-plugin-emotion): allow newer versions of babel-preset-css-prop

### LINKED ISSUE #11915: gatsby-plugin-emotion throws an error if a component only contains fragment JSX
## Description

gatsby-plugin-emotion throws an error if it encounters a component that only contains fragment syntax and no other JSX. This is due to [an upstream issue in `babel-preset-css-prop` that has since been fixed](https://github.com/emotion-js/emotion/issues/1147).

Would you accept a PR to bump the dependency version? Not sure if the dependency is pinned to 10.0.5 for a specific reason or if it's just that nobody's had good reason to update yet :)

### Steps to reproduce

* Create a Gatsby project from the hello world starter.
* Add `gatsby-plugin-emotion` to the project.
* Render the following component:

```js
import React from "react";

export default function Component(props) {
    return <></>;
}
```

### Expected result

The component should be rendered, and the Emotion Babel transforms should be applied.

### Actual result

A `ReferenceError: ___EmotionJSX is not defined` is thrown.

I was able to work around the issue by manually inserting the missing pragma import into the file:

```js
import { jsx as ___EmotionJSX } from "@emotion/core";
```

Or by amending the component to use `React.Fragment` instead of the shorthand syntax.

### Environment
  System:
    OS: Windows 10
    CPU: (12) x64 AMD Ryzen 5 1600 Six-Core Processor
  Binaries:
    npm: 6.7.0 - C:\Program Files\nodejs\npm.CMD
  Languages:
    Python: 3.7.2 - C:\Users\caven\AppData\Local\Programs\Python\Python37-32\python.EXE
  Browsers:
    Edge: 42.17134.1.0
  npmPackages:
    gatsby: ^2.1.9 => 2.1.9
    gatsby-plugin-catch-links: ^2.0.11 => 2.0.11
    gatsby-plugin-emotion: ^4.0.3 => 4.0.3
    gatsby-source-filesystem: ^2.0.20 => 2.0.20
    gatsby-transformer-remark: ^2.2.5 => 2.2.5

(`npx gatsby info --clipboard` crashes at that point for me, not sure what's happening there... let me know if you need more info!)

17cupsofcoffee:
## Description

This PR updates the `package.json` of gatsby-plugin-emotion to allow newer versions of babel-preset-css-prop to be used.

## Related Issues

This fixes #11915, as newer versions of babel-preset-css-prop have resolved that issue.

### DIFF
--- a/packages/gatsby-plugin-emotion/package.json
+++ b/packages/gatsby-plugin-emotion/package.json
@@ -8,7 +8,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.0.0",
-    "@emotion/babel-preset-css-prop": "10.0.5"
+    "@emotion/babel-preset-css-prop": "^10.0.5"
   },
   "devDependencies": {
     "@babel/cli": "^7.0.0",


17cupsofcoffee:
Not sure why one of the E2E tests is failing - it doesn't depend on gatsby-plugin-emotion, so I think it's just CI being weird...

wKovacs64:
Cypress on Circle is dodgy, I just re-ran it and it passed. Of course, now it's failing for another unrelated reason, but... oh well. ¯\\\_(ツ)\_/¯

DSchau:
@wKovacs64 the starters:validate task is actually failing (and has been for a few days now).

Follow along gatsbyjs/gatsby#11895 for more info

wKovacs64:
@DSchau Yep, I'm aware. 🙂

DSchau:
@wKovacs64 whoops - I didn't know they were failing because the restart fixed it (so they were all green, besides the starters:validate).

I'll take another look at that tomorrow--that seems to be a pesky reoccurring one.

Thanks for clarifying!

DSchau:
```
Successfully published:
 - gatsby-plugin-emotion@4.0.4
```

🎉 Thanks so much for this PR!

wKovacs64:
@DSchau All good, the timing definitely muddied that scenario. :)

Issue #6921, opened 2018-08-01
Inconsistent / incomplete internal Node data structure


lightstrike:
I'm exploring how to do some data transformation on Gatsby nodes using the `onCreateNode` hook for the `gatsby-source-wordpress` plugin. I'm finding that the [internal Node data structure documented](https://next.gatsbyjs.org/docs/node-interface/) is different from what I'm observing: 

```js
{
  type: 'SitePage',
  contentDigest: '5e06d7f4cf28e1f0498cb2fa52539086',
  description: 'Your site\'s "gatsby-node.js"',
  owner: 'internal-data-bridge'
}
```

I'm specifically trying to transform WordPress media nodes. I believe the internal type should be `wordpress__wp_media` for these nodes but only nodes with type `SitePage` are being logged within `onCreateNode`.

I'm trying this with `gatsby: 2.0.0-beta.20` and `gatsby-source-wordpress:  3.0.0-beta.5`

KyleAMathews:
Are you sure any WordPress nodes are being created?

lightstrike:
Yes they [are created here](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-source-wordpress/src/normalize.js#L572), which is called within [`sourceNodes`](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-source-wordpress/src/gatsby-node.js#L154) in `gatsby-node`.

I can query all the WordPress nodes within graphQL, they aren't surfacing within `onCreateNode` though.

lightstrike:
Looking at this again, internal structure is good, so I'm closing :smile:

PR #11925, opened 2019-02-20
feat(gatsby-plugin-sharp): Remove sharp SIMD instructions

### LINKED ISSUE #6291: Errors when processing large(ish) numbers of images with gatsby-plugin-sharp
## Description

I'm processing around 1500 images with gatsby-plugin-sharp/gatsby-transformer-sharp and GraphQL, to display with gatsby-image as per [this page in the docs](https://www.gatsbyjs.org/packages/gatsby-image/).

For smaller numbers of images I have had no issues, but now that I'm processing 1500 images the build randomly fails with the following error:

```
(sharp:104): GLib-GObject-WARNING **: 09:35:11.293: gtype.c:4265: type id '0' is invalid

(sharp:104): GLib-GObject-WARNING **: 09:35:11.293: can't peek value table for type '<invalid>' which is not currently referenced

(sharp:104): GLib-GObject-WARNING **: 09:35:11.293: gvalue.c:188: cannot initialize GValue with type '(null)', this type has no GTypeValueTable implementation

(sharp:104): GLib-GObject-CRITICAL **: 09:35:11.293: g_value_transform: assertion 'G_IS_VALUE (src_value)' failed
Received 'segmentation fault' signal
```

This happens during the "Generating image thumbnails" step in the build process and it occurs apparently randomly (sometimes 10% of the way through the images, sometimes at 80%, sometimes not at all). Therefore I do not believe it is caused by a "faulty image".

niklasravnsborg:
## Description

On larger builds where the gatsby-plugin-sharp is used to transform large amounts of images, the build sometimes fails in an operation ‘close to the metal‘. This change increases the build time a little bit but it should prevent the issue. Further testing is required, to ensure the error doesn't occur again.

## Related Issues

This should fix #6291.

### DIFF
--- a/packages/gatsby-plugin-sharp/src/index.js
+++ b/packages/gatsby-plugin-sharp/src/index.js
@@ -56,12 +56,6 @@ exports.setPluginOptions = opts => {
 // raw) callback-accepting toBuffer(...) method
 Promise.promisifyAll(sharp.prototype, { multiArgs: true })
 
-// Try to enable the use of SIMD instructions. Seems to provide a smallish
-// speedup on resizing heavy loads (~10%). Sharp disables this feature by
-// default as there's been problems with segfaulting in the past but we'll be
-// adventurous and see what happens with it on.
-sharp.simd(true)
-
 const bar = new ProgressBar(
   `Generating image thumbnails [:bar] :current/:total :elapsed secs :percent`,
   {


KyleAMathews:
What kind of speed differences do you see on builds?

tinder-kyleboss:
I do have some bad news. 😢 

I have a fairly large project that has many images being processed. This project errors via segmentation fault from the `gatsby-plugin-sharp` plugin.

We decided to give this PR a run and we've found that the segmentation faults still exist. The build-time actually remained the same, however. 👍

KyleAMathews:
Btw, we've worked with the sharp author in the past. If you have a site / images that consistently fail, we could contact him to look into things.

tinder-kyleboss:
Hey @KyleAMathews,

That would be really cool. It seems as sites get bigger with more fluid images, we seem to be getting more failures. It's manageable now, but definitely something that is on our 'scalability' radar.

wardpeet:
@KyleAMathews should we contact him? I have no clue if this is a good or bad change 😄

sidharthachatterjee:
@niklasravnsborg I'm conflicted on merging this in because if there isn't any difference in behaviour (segfaults for a particular image that @tinder-kyleboss mentions in either case) then it's better to have the SIMD optimization

Do you want to run a benchmark to check the speed difference? You could run it [www.gatsbyjs.org](https://github.com/gatsbyjs/gatsby/tree/master/www) 

@tinder-kyleboss If it is possible to narrow it down, would you like to open an issue with the particular image(s) that break?

tinder-kyleboss:
My apologies, as I didn't want to hijack this PR. Just wanted to mention that the proposed fix didn't fully fix our segmentation fault issue with gatsby-plugin-sharp. While the issue of segmentation faults might still exist, it's worth mention that is may have lowered the chances of one happening.

My issue is the exact same as #6921. I am following it very closely :) So I will just keep an eye out there.

wardpeet:
@tinder-kyleboss agree to close this one until we pinpoint the issue?

tinder-kyleboss:
@niklasravnsborg, has this PR improved anything with the segmentation faults on your end?

niklasravnsborg:
@tinder-kyleboss I only experience the errors on a large production build on a cloud build machine. I can't test this branch there because I don't have direct shell access.

I agree to close this until we pinpoint the issue.

PR #11927, opened 2019-02-20
Fix(webpack.config.js) Fix test for one CSS chunk to cover more flavors of CSS

KyleAMathews:

### DIFF
--- a/packages/gatsby/src/utils/webpack.config.js
+++ b/packages/gatsby/src/utils/webpack.config.js
@@ -398,7 +398,8 @@ module.exports = async (
           // See https://github.com/gatsbyjs/gatsby/issues/11072
           styles: {
             name: `styles`,
-            test: /\.css$/,
+            // This should cover all our types of CSS.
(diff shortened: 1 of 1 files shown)


packages/gatsby/src/utils/webpack.config.js
```diff
@@ -398,7 +398,8 @@ module.exports = async (
           // See https://github.com/gatsbyjs/gatsby/issues/11072
           styles: {
             name: `styles`,
-            test: /\.css$/,
+            // This should cover all our types of CSS.
+            test: /\.(css|scss|sass|less)$/,
```
wardpeet: you should add `styl` as well.

pieh:
```suggestion
            test: /\.(css|scss|sass|less|styl)$/,
```
We also have stylus.

Some day we could traverse rules and extract `test` from ones that use `mini-css-extract`. But for quick fix this should be fine.

KyleAMathews:
Lots of `styl` fans haha :-P

PR #11940, opened 2019-02-20
Add plugins option to gatsby-transformer-json

sbardian:
## Description

`Enhancement`:  Wanted an option to limit the nodes that `gatsby-transformer-json` performs transformations on.  This would be a performance improvement and limit the amount of unused types and queries polluting your schema when using this plugin.

Thoughts?

### DIFF
--- a/packages/gatsby-transformer-json/README.md
+++ b/packages/gatsby-transformer-json/README.md
@@ -140,6 +140,23 @@ Which would return:
 
 ## Configuration options
 
+**`plugins`** [string[]][optional]
+
+By default, all JSON nodes are transformed. Use the `plugins` array option as an opt-in performance enhancement to only transform nodes created by specific plugins. This will also reduce the amount of unused queries and types in your GraphQL schema.
+
+```javascript
+module.exports = {
+  plugins: [
+    {
+      resolve: `gatsby-transformer-json`,
+      options: {
+        plugins: [`gatsby-source-filesystem`],
+      },
+    },
+  ],
+}
+```
+
 **`typeName`** [string|function][optional]
 
 The default naming convention documented above can be changed with
--- a/packages/gatsby-transformer-json/src/__tests__/gatsby-node.js
+++ b/packages/gatsby-transformer-json/src/__tests__/gatsby-node.js
@@ -208,4 +208,59 @@ describe(`Process JSON nodes correctly`, () => {
       expect(createParentChildLink).toHaveBeenCalledTimes(1)
     })
   })
+
+  it(`process node that is listed in plugins pluginOption`, async () => {
+    const data = { id: `foo`, blue: true, funny: `yup` }
+    const node = {
+      ...baseNonFileNode,
+      content: JSON.stringify(data),
+      internal: {
+        ...baseNonFileNode.internal,
+        owner: `gatsby-source-filesystem`,
+      },
+    }
+    const createNode = jest.fn()
+    const createParentChildLink = jest.fn()
+    const loadNodeContent = jest.fn()
+    loadNodeContent.mockReturnValue(JSON.stringify(data))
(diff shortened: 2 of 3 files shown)


pieh:
Just to check - do you want to ignore `gatsby-source-filesystem`?

sbardian:
I was using `gatsby-source-sanity` and `gatsby-transformer-json` was running on each of the node types from sanity, so I would end up with I think 4 extra types/queries in my schema *JSON for each sanity node type.  I was using `gatsby-transformer-json` with `gatsby-source-filesystem` along side `gatsby-source-sanity`, so was looking to limit the transformer-json to just `gatsby-source-filesystem` and ignore `gatsby-source-sanity`.  Hope that makes sense?

sbardian:
I guess a better answer would be, I want to specify what plugin(s) `gatsby-transformer-json` should transform the nodes json.

pieh:
I understand the use case, but I'm not sure we should implement this in transformer plugin itself - this seems like something that should be bumped to be handled by gatsby core, so same could be applied to all kinds of transformer plugins. I'm not sure yet how this could look like tho - that's longer discussion (probably needs RFC because it does add new dynamic to data model).

If this is particularly bothering you I would suggest forking `gatsby-transformer-json` and adding it is [local plugin](https://www.gatsbyjs.org/docs/plugin-authoring/#local-plugins).

sbardian:
No, only noticed because `gatsby-source-sanity` wasn't setting `content` or `loadNodeContent()` on their nodes, causing an error when `gatsby-transformer-json` would hit the sanity nodes (I submitted a PR to `gatsby-source-sanity` for this), and it gave me the idea to be able to narrow what plugins `gatsby-transformer-json` transformed on.

pieh:
if that's the case we should add handling of empty node content better - as in, not error and just skip. Does it error in `gatsby-transformer-json` or somewhere in `gatsby`?

sbardian:
Ah, not a bad idea.  Here is the error I believe I was getting. 
https://github.com/gatsbyjs/gatsby/blob/bf042a04d95b33ee0d2e4675181587f890b72240/packages/gatsby/src/db/nodes.js#L40

Would be happy to attempt another PR if you would like, you think just bailing out is the best approach?

sbardian:
Side note, `gatsby-source-sanity` has stopped returning:
```javascript
mediaType: 'application/json',
```
on their nodes, so the `gatsby-transformer-json` is no longer trying to transform it.  Could still hit other nodes with a `mediaType` of `application/json` though.

sbardian:
hey @pieh thought I would bump this one and see if you needed/wanted me to do anything else as far as RFC to discuss the option of handling this in `gatsby core`? This specific issue was cleared up for me with the update to `gatsby-source-sanity`, but the underlying problem still exists and could present itself in many plugins. Or we can just close this one? 😁

wardpeet:
Like @pieh mentioned this is not something we want to introduce in our transformers but rather have a mechanism in core to actually deal with this.

I'm wondering if the new GraphQL api can help here. @freiksenet @stefanprobst

freiksenet:
@wardpeet Unfortunately that's not going to be helped by schema refactoring. This all would happen in node processing which happens before schema is created. I think @pieh has more context about what API should be, but I'd imagine all transformer plugin configs will accept a filter option that lets you select which nodes it's applied too.

wardpeet:
I agree we should probably create an issue instead and from that issue, we can probably create an RFC. I'll be closing this PR for this reason. @sbardian do you mind creating an issue and describing the problem so we can start the discussion with the community.

sbardian:
See issue #14081.  Wasn't sure if I was stating things very well, please feel free to edit/correct any misunderstandings on my part/etc.

PR #11941, opened 2019-02-20
feat(docs): adding an RSS Feed

marcysutton:
## Description

When I set up an RSS feed for my website, I found some critical information was missing from the `gatsby-plugin-feed` README. I also didn't see a doc for creating an RSS feed, so I added one. That's what this PR does, in addition to some title case updates to `doc-links.yml` while I was in there.


docs/docs/adding-an-rss-feed.md
```diff
@@ -0,0 +1,157 @@
...
+
+```sh
+npm install --save gatsby-plugin-feed
+```
+
+## How to use [gatsby-plugin-feed](/packages/gatsby-plugin-feed/)
+
+Once installation is complete, you can now add this plugin to your site's config file, like so:
+
+```json:title=gatsby-config.js
```
DSchau:
```suggestion
```js:title=gatsby-config.js
```

You'll want to tweak the code snippet too!

docs/docs/adding-an-rss-feed.md
```diff
@@ -0,0 +1,157 @@
...
+(module.exports = {
+  "siteMetadata": {
+    "siteUrl": `https://www.example.com`
+  },
+  "plugins": [`gatsby-plugin-feed`]
+})
+```
+
+To complete the feed setup, you need to expose a GraphQL entry for our content called `fields.slug` by modifying `gatsby-node.js`. Start with the following code, noting the reference to `MarkdownRemark` content. For content sources other than Markdown, you will want to modify it:
```
DSchau:
```suggestion
To complete the feed setup, you need to expose a GraphQL entry for our content called `fields.slug` by modifying `gatsby-node.js`. Start with the following code, noting the reference to `MarkdownRemark` content. For content types other than Markdown, you will want to tweak the condition:
```

Also maybe an example? e.g. content types other than Markdown like Contentful, Wordpress, etc.

docs/docs/adding-an-rss-feed.md
```diff
@@ -0,0 +1,157 @@
...
+    },
+```
+
+This snippet contains a custom `gatsby-plugin-feed` setup in `gatsby-config.js` to query metadata for your site, like its `title` and `siteUrl`. It also includes a `feeds` array with at least one object containing a GraphQL query and `serialize` method, which allows you to output a custom RSS feed structure. In this example, the RSS content comes from Markdown files sourced from your site, and queried with the key `allMarkdownRemark` and its associated filters and fields.
+
+The `output` field in your feed object allows you to customize the filename for your RSS feed, and `title` for the name of your site's RSS feed.
+
+> To see your feed in action, run `gatsby build && gatsby serve` and you can then inspect the content and URLs in your RSS file at `http://localhost:9000/rss.xml`.
+
+## Additional customization for content slugs
```
DSchau: This seems a little bit tangential? (and the regex is a little gnarly 😈)

packages/gatsby-plugin-feed/README.md
```diff
@@ -22,6 +22,25 @@ plugins: [
 ]
 ```
 
+To complete the feed setup, you need to expose a GraphQL entry for our content called `fields.slug` by modifying `gatsby-node.js`. Start with the following code, noting the reference to `MarkdownRemark` content. For content sources other than Markdown, you will want to modify it:
```
DSchau: What do you think about linking to the newly established doc you're creating here?

Generally - this makes it seem like you _have_ to create a slug, when rather you could just customize the query to match whatever data structure your posts require? (e.g. path from wordpress, something else from drupal, etc.)

docs/docs/adding-an-rss-feed.md
marcysutton: I don't think my setup is that unusual, and it seems like something worth mentioning so people don't have to figure it out on their own? I'm open to regex wizardry suggestions 😄

packages/gatsby-plugin-feed/README.md
marcysutton: I can link to the new doc, for sure. The examples given in the plugin README don't work on their own without more details, though. Perhaps I can reword it to be more open-ended?

DSchau: Yeah, really good point! Our README's should always show a working case, even if it's slightly trivial. This one is kind of hard because the developer will _have_ to implement this section depending on how they're sourcing the data.

But yeah - I like your idea of re-wording maybe. As in, "Here's an example of how you could implement this plugin with Markdown, but for other sources, you will need a way to uniquely identify content--typically a URL or slug."

or something!

docs/docs/adding-an-rss-feed.md
DSchau: Generally - I think if we document how to create a slug (which we have!), then the rest will sort of follow.

So if someone knew they wanted to remove dates from their slugs--since they already know _how_ to create a slug, that part would be pretty understandable?

Not sure--I can see how there's value in documenting extra cases, but I also don't know how common this is. Guessing you're doing this for your site?

KyleAMathews:
The plugin should be fixed so "fields.slug" isn't required. You should be able to specify that directly. Lots of people will have the URL elsewhere.

LekoArts:
Just as an idea: What about i18n and RSS feeds?

marcysutton:
@LekoArts I love the idea! Since there isn't much on i18n currently, I think that addition could come later after this doc is merged.

marcysutton:
Ok, this PR is ready! I had to fix some permalinks in my own feed to get a full understanding of what was going on. In short, I have custom `path` entries in content frontmatter to match my old site 
 URLs, and I needed to use those instead of `createFilePath`. That and use `my-blog-post/index.md` instead of `my-blog-post/2019-01-04-my-blog-post.md` in named content directories. All good now 😜 

For the post, I kept the customization details simple based on your feedback and added an "ask us for help!" callout in case anyone gets stuck. Given Wordpress generates RSS automatically, we should make this as easy to figure out as possible.

I'll file an issue to look into i18n for RSS. Thanks for the suggestions!

docs/docs/adding-an-rss-feed.md
```diff
@@ -0,0 +1,135 @@
...
+Your content might not fit neatly into the blog-starter scenario, for various reasons like:
+
+- Your content isn't in Markdown so the plugin doesn't know about it
+- Your Markdown files have dates in the filenames, for which the slug URLs cause 404s
+
+The good news is you can accommodate these scenarios and more in `gatsby-config.js` and `gatsby-node.js`.
+
+To customize the default feed schema (a.k.a. structure) output by the plugin to work with your website's content, you can start with the following code:
+
+```json:title=gatsby.config.js
```
DSchau: This is js!

Would generally recommend doing something like:

````md
```js:title=gatsby-config.js
module.exports = {
 plugins: [
   // your code block
 ]
}
```
````

just so we can show it as real-world as possible!

File: docs/docs/adding-an-rss-feed.md [hunk 6]
```diff
@@ -0,0 +1,135 @@
...
+    },
+```
+
+This snippet contains a custom `gatsby-plugin-feed` setup in `gatsby-config.js` to query metadata for your site, like its `title` and `siteUrl`. It also includes a `feeds` array with at least one object containing a GraphQL query and `serialize` method, which allows you to output a custom RSS feed structure. In this example, the RSS content comes from Markdown files sourced from your site, and queried with the key `allMarkdownRemark` and its associated filters and fields.
+
+The `output` field in your feed object allows you to customize the filename for your RSS feed, and `title` for the name of your site's RSS feed.
+
+To see your feed in action, run `gatsby build && gatsby serve` and you can then inspect the content and URLs in your RSS file at `http://localhost:9000/rss.xml`.
+
+> NOTE: if your blog has custom permalinks, such as links with or without dates in them, you may need to [customize `gatsby-node.js`](https://github.com/gatsbyjs/gatsby-starter-blog/blob/master/gatsby-node.js#L57) to output the correct URLs in your RSS feed. [Get in touch with us](/contributing/how-to-contribute/) if you need any help!
```
DSchau: Like this pattern a lot!

docs/docs/adding-an-rss-feed.md
marcysutton: Done!

docs/docs/adding-an-rss-feed.md
marcysutton: Seriously, when I was learning to build websites I would have loved this. I felt bad always pestering my one friend who knew how everything worked...having a whole community available would have been amazing.

docs/docs/adding-an-rss-feed.md
```diff
@@ -0,0 +1,139 @@
...
+      resolve: `gatsby-plugin-feed`,
+      options: {
+        query: `
+          {
+            site {
+              siteMetadata {
+                title
+                description
+                siteUrl
+                site_url: siteUrl
```
muescha: any explanation why this additional `site_url` is needed?

PR #11945, opened 2019-02-20
fix(tests): delay going back in path-prefix E2E test

wKovacs64:
## Description

This is an attempt at fixing the `e2e_tests_path-prefix` CI job which often fails but will succeed upon simply running the test again. I believe the test is trying to call `cy.go('back')` before everything has settled, so the assertion following that instruction times out waiting for something that is never going to arrive (just an educated guess, though). This change waits for an expected element to exist in the DOM prior to calling `cy.go('back')` to give poor Cypress time to catch its breath.


wardpeet:
Thanks for creating this PR. It seems our friendly neighbourhood member @DSchau already fixed this issue. I see that you created this one before him but sadly his PR already got merged.

Thanks for making Gatsby better! 💪 and we hope to see more contributions from you! ❤️

Issue #11137, opened 2019-01-18
gatsby-plugin-netlify-cms errors when components dependent on gatsby modules are imported in cms.js


vai0:
## Summary
I am attempting to register a preview template in `cms.js` with the component I'm also using in Gatsby. However, gatsby-plugin-netlify-cms throws the error below when Gatsby is built. I've narrowed it down to the reason being that gatsby-plugin-netlify-cms' webpack instance cannot process the Link component from the "gatsby" module.

```js
// cms.js
import CMS from "netlify-cms";

import Test from "components/Test";
```

```js
// Test.js
import React from "react";

import { Link } from "gatsby";

export class Test extends React.Component {
    render() {
        return <div><Link to="/hello">Hello</Link></div>;
    }
}
```

Error
```bash
error  in ./node_modules/gatsby/cache-dir/gatsby-browser-entry.js

Module parse failed: Unexpected token (16:2)
You may need an appropriate loader to handle this file type.
|
| const StaticQuery = props => (
>   <StaticQueryContext.Consumer>
|     {staticQueryData => {
|       if (

 @ ./src/components/Link.js 61:6-22
 @ ./src/components/Test.js
 @ ./src/cms/cms.js
 @ multi ./plugins/gatsby-plugin-netlify-cms/cms.js ./src/cms/cms.js
```

Is there an obvious workaround for this? 

**Gatsby version:** 2.0.90
**gatsby-plugin-netlify-cms version:** 3.0.9

sidharthachatterjee:
Interestingly, I'm not able to reproduce this 

https://github.com/sidharthachatterjee/netlify-cms-test seems to build fine 

I do remember fixing this for Storybook though so could you please try the repository I've linked to and reproduce this issue there?

vai0:
@sidharthachatterjee I was able to reproduce it by adding the following into your test repo's `gatsby-plugin-netlify-cms`'s `gatsby-node.js` webpack config variable:

```
var config = (0, _extends2.default)({}, gatsbyConfig, {
  ...
  resolve: {
    modules: [_path.default.resolve(__dirname, "../../src"), "node_modules"]
  },
  ...
});
```

Error:
```
ERROR  Failed to compile with 1 errors                                                        17:35:33

 error  in ./node_modules/gatsby/cache-dir/gatsby-browser-entry.js

Module parse failed: Unexpected token (16:2)
You may need an appropriate loader to handle this file type.
|
| const StaticQuery = props => (
[...]

Issue #11119, opened 2019-01-16
Can modules imported into cms.js for netlify-cms be resolved relative to one specified directory?


vai0:
It seems like imports and the imported module's dependencies in the cms.js module must be all be relative. Is there a workaround to allowing absolute imports, something similar along the lines of webpack config's [resolve](https://webpack.js.org/configuration/resolve/)?

The reason I ask is because I'm importing a React component that is resolving it's own imports relative to `src`.

Essentially, it seems like the `cms.js` isn't respecting my modified webpack config which added a `resolve.modules`.

```
// cms.js
import CMS from "netlify-cms";

import StandardLandingPagePreview from "./components/preview-templates/StandardLandingPagePreview";

CMS.registerPreviewTemplate('lp-standard', StandardLandingPagePreview)
```

```
// StandardLandingPagePreview.js
import React from "react";

import BaseLayout from "components/layouts/BaseLayout";
import StandardLandingPage from "components/StandardLandingPage";

export default class StandardLandingPagePreview extends React.Component {
    render() {
        ...
        );
    }
}

```

This will throw error:
```
 ERROR  Failed to compile with 2 errors                                                                                14:06:49

These dependencies were not found:

* components/StandardLandingPage in ./src/components/preview-templates/StandardLandingPagePreview.js
* components/layouts/BaseLayout in ./src/components/preview-templates/StandardLandingPagePreview.js

To install them, you can run: npm install --save components/StandardLandingPage components/layouts/BaseLayout
```
Here is my gatsby-node.js file:
```
// gatsby-node.js
exports.onCreateWebpackConfig = ({ stage, actions, plugins, loaders }) => {
    let config = {
        resolve: {
            modules: [path.resolve(__dirname, "src"), "node_modules"],
        },
        ...
    };

    actions.setWebpackConfig(config);
};
```

**Applicable Versions:**
 - Netlify CMS version: 2.3.3
 - Git provider: Gitlab
 - OS: Mac OS 10.14

**Additional context**
I am running netlify-cms with GatsbyJS via [gatsby-plugin-netlify-cms](https://www.gatsbyjs.org/packages/gatsby-plugin-netlify-cms/?=gatsby-plugin-netlify)

sidharthachatterjee:
@vai0 The Netlify CMS plugin runs a separate instance of webpack it does spread the gatsby webpack config  but this is probably being run before your code) so that would explain why your code in `gatsby-node.js` isn't working 

[...]

PR #11947, opened 2019-02-21
Add resolvePaths, get webpack instance to recognize react and es6 syntax

vai0:
## Description
Changes:
- Currently gatsby-plugin-netlify-cms' webpack instance cannot recognize React or ES6 class syntax when components are pulled in and registered as templates. This adds react and es6 class syntax support.
- Adds the resolvePaths config option, which gets dumped into the webpack's resolve config. i.e., now referenced modules with absolute path imports work.
- Deal with components that import gatsby components.

## Related Issues
https://github.com/gatsbyjs/gatsby/issues/11119
https://github.com/gatsbyjs/gatsby/issues/11137

### DIFF
--- a/packages/gatsby-plugin-netlify-cms/src/cms.js
+++ b/packages/gatsby-plugin-netlify-cms/src/cms.js
@@ -1,5 +1,27 @@
 import CMS from "netlify-cms"
 
+// The following window and global config settings below were taken from here.
+// https://github.com/gatsbyjs/gatsby/blob/master/docs/docs/visual-testing-with-storybook.md
+// They're required because the netlify-cms runs on a separate webpack config,
+// and outside of Gatsby. This ensures any Gatsby components imported into the
+// CMS works without errors
+
+// highlight-start
+// Gatsby's Link overrides:
+// Gatsby defines a global called ___loader to prevent its method calls from creating console errors you override it here
+global.___loader = {
+  enqueue: () => {},
+  hovering: () => {},
+}
+
+// Gatsby internal mocking to prevent unnecessary errors
+global.__PATH_PREFIX__ = ``
+
+// This is to utilized to override the window.___navigate method Gatsby defines and uses to report what path a Link would be taking us to
+window.___navigate = pathname => {
+  alert(`This would navigate to: ${pathname}`)
+}
+
 /**
  * The stylesheet output from the modules at `modulePath` will be at `cms.css`.
  */
--- a/packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
+++ b/packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
@@ -55,6 +55,7 @@ exports.onCreateWebpackConfig = (
     enableIdentityWidget = true,
     htmlTitle = `Content Manager`,
     manualInit = false,
+    resolvePaths = [],
   }
 ) => {
   if ([`develop`, `build-javascript`].includes(stage)) {
@@ -74,6 +75,9 @@ exports.onCreateWebpackConfig = (
       },
       output: {
         path: path.join(program.directory, `public`, publicPathClean),
+      },
+      resolve: {
+        modules: [...resolvePaths, `node_modules`],
       },
       module: {
         /**
@@ -160,6 +164,30 @@ exports.onCreateWebpackConfig = (
       optimization: {},
       devtool: stage === `develop` ? `cheap-module-source-map` : `source-map`,
     }
+
+    config.module.rules.push({
+      test: /gatsby\/cache-dir.*\.js$/,
+      loader: require.resolve(`babel-loader`),
+      options: {
+        presets: [
+          require.resolve(`@babel/preset-react`),
+          [
+            require.resolve(`@babel/preset-env`),
+            {
+              shippedProposals: true,
+              useBuiltIns: `entry`,
+            },
+          ],
+        ],
+        plugins: [require.resolve(`@babel/plugin-proposal-class-properties`)],
+      },
+    })
+
+    // Transpile Gatsby module because Gatsby includes un-transpiled ES6 code.
+    config.module.rules.exclude = [/node_modules\/(?!(gatsby)\/)/]
+
+    // Prefer Gatsby ES6 entrypoint (module) over commonjs (main) entrypoint
+    config.resolve.mainFields = [`browser`, `module`, `main`]
 
     if (stage === `develop`) {
       webpack(config).watch({}, () => {})
--- a/packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
+++ b/packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
@@ -186,9 +186,6 @@ exports.onCreateWebpackConfig = (
     // Transpile Gatsby module because Gatsby includes un-transpiled ES6 code.
     config.module.rules.exclude = [/node_modules\/(?!(gatsby)\/)/]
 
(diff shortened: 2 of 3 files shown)


packages/gatsby-plugin-netlify-cms/src/cms.js
```diff
@@ -1,5 +1,27 @@
 import CMS from "netlify-cms"
 
+// The following window and global config settings below were taken from here.
```
wardpeet: to me this is a different PR. If I read correctly this will fix support for storybook and has nothing to do with import compnents in cms.js? Am I correct?

packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
```diff
@@ -145,6 +149,30 @@ exports.onCreateWebpackConfig = (
...
+        ],
+        plugins: [require.resolve(`@babel/plugin-proposal-class-properties`)],
+      },
+    })
+
+    // Transpile Gatsby module because Gatsby includes un-transpiled ES6 code.
+    config.module.rules.exclude = [/node_modules\/(?!(gatsby)\/)/]
+
+    // Prefer Gatsby ES6 entrypoint (module) over commonjs (main) entrypoint
+    config.resolve.mainFields = [`browser`, `module`, `main`]
```
wardpeet: if not mistaken it's the default https://webpack.js.org/configuration/resolve/#resolvemainfields

packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
```diff
@@ -145,6 +149,30 @@ exports.onCreateWebpackConfig = (
       devtool: stage === `develop` ? `cheap-module-source-map` : `source-map`,
     }
 
+    config.module.rules.push({
+      test: /gatsby\/cache-dir.*\.js$/,
+      loader: require.resolve(`babel-loader`),
+      options: {
+        presets: [
+          require.resolve(`@babel/preset-react`),
```
wardpeet: you'll have to add this as a `dependency` or `peerdependency`

packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
```diff
@@ -145,6 +149,30 @@ exports.onCreateWebpackConfig = (
...
+          [
+            require.resolve(`@babel/preset-env`),
```
wardpeet: same here

packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
```diff
@@ -145,6 +149,30 @@ exports.onCreateWebpackConfig = (
...
+            {
+              shippedProposals: true,
+              useBuiltIns: `entry`,
+            },
+          ],
+        ],
+        plugins: [require.resolve(`@babel/plugin-proposal-class-properties`)],
```
wardpeet: same here

wardpeet:
I left some comments, I don't know enough of Contenful to actually review this one correctly. I'm wondering if a chlid compilation instead of a whole webpack compilation is a better approach so we don't have to recreate the whole config. I'm not sure child compilations take all config from the parent.

@gatsbyjs/core any ideas?

packages/gatsby-plugin-netlify-cms/src/cms.js
vai0: This is also relevant here because netlify runs it's own webpack instance, and encounters the same build issues as storybook's webpack instance when it encounters gatsby-related components, i.e., <Link>, <StaticQuery>, etc...

packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
vai0: Good point, removing. This setting was directly taken from: https://www.gatsbyjs.org/docs/visual-testing-with-storybook/ Given that storybook runs on webpack 4 and the webpack docs is referencing v4, not exactly sure why this setting is relevant.

BrunoBernardino:
@gatsbyjs/core friendly bump! 😊

m-allanson:
Hey @vai0, are you still seeing this issue when using the newer `gatsby-plugin-netlify-cms` version 4? It was released in https://github.com/gatsbyjs/gatsby/pull/13036. If that fixes the issue then let's go ahead and close this PR, otherwise can you re-apply your changes to the latest version?

Many thanks!

vai0:
@m-allanson suuper busy atm. will try getting to it end of this week! please do ping me again if it doesn't happen by fri

wardpeet:
The latest version of [gatsby-plugin-netlify-cms](https://www.gatsbyjs.org/packages/gatsby-plugin-netlify-cms/?=gatsby-plugin-netlify-cms) fixes this issue. React components can be used inside cms.js from version 2.9.0.

I'll be closing this one. Thanks for all your effort!

vai0:
Just tested this now. The latest version is still missing the ability to resolve absolute paths. will open a separate PR for that

andru:
For anyone coming across this struggling with one problem or another relating to `gatsby-plugin-netlify-cms` not finding your modules - like if you're using `directory-named-webpack-plugin`, or not having the available plugins to process them - like if you're using  if you're importing SVGs as React components with 

The problem is that any webpack config you have in your `gatsby-node.js`  in the project root is loaded _after_ `gatsby-plugin-netlify-cms` has been processed and copied over the gatsby webpack config to build netlify-cms.
 
I've found the easiest workaround is to just move my custom webpack config from the `gatsby-node.js` in the project root into a custom plugin, and add it as the first plugin so that config is available to all other plugins, including `netlify-cms`.

Maybe there's another lifecycle hook to add webpack config before plugins are run but if there is I can't find it.

dustinhorton:
Agreed with @vai0—this isn't fixed.

I'm using 'gatsby-plugin-root-import' so all my absolute imports are failing in Netlify CMS.

@vai0 are you still using what's in your PR?

@andru do you have an example of how you're doing that?

vai0:
@dustinhorton yes, still relying on my own fork for this

dustinhorton:
@vai0 Thanks—having some issues getting this working. By fork you mean what's at https://github.com/vai0/gatsby-plugin-netlify-cms, not this PR, right? Could I trouble you to give an example of how you set `resolvePaths` in `gatsby-config.js`?

vai0:
yes that fork. here's my `gatsby-config.js`
```
...
{
    resolve: "gatsby-plugin-netlify-cms",
    htmlTitle: "Bifrost Content Manager",
    options: {
        enableIdentityWidget: false,
        modulePath: `${__dirname}/src/cms/cms.js`,
        htmlTitle: "Bifrost Admin",
        resolvePaths: [path.resolve("./src")],
    },
},
...
```

andru:
I don't have a public repo to share for my fix, but it's pretty straightforward:

- Make a local plugin ([gatsby docs](https://www.gatsbyjs.org/docs/loading-plugins-from-your-local-plugins-folder/))
- Make sure this plugin is listed before netlify plugins in  `gatsby-config.js`
- Add whatever webpack config netlify needs in the `onCreateWebpackConfig` hook in your plugin. e.g...

```js
exports.onCreateWebpackConfig = ({
  stage,
  getConfig,
  plugins,
  rules,
  loaders,
  actions,
}) => {
  actions.setWebpackConfig({
    // ... your config
  })
}
```

In my case, I'm using `directory-named-webpack-plugin` in my project, so to getso my plugin looks like this: https://jsfiddle.net/norapmh7/

PR #11954, opened 2019-02-21
fix(themes): reverse order of themes checked when shadowing

### LINKED ISSUE #11951: [themes] Component shadowing doesn't resolve properly if multiple themes shadow the same component
## Description

In a theme scenario where there are multiple ancestors, component shadowing appears to only respect the first component (or it stops looking after a certain point).

Example (site uses theme3, theme3 uses theme2, theme2 uses theme1):
```
=> theme1/src/components/foo.js
=> theme2/src/theme1/components/foo.js (actually shows up)
=> theme3/src/theme1/components/foo.js (should show up)
=> site (where develop is run from)
```

### Steps to reproduce

See this repo: https://github.com/jlengstorf/theme-shadowing-bug

### Environment

```
  System:
    OS: macOS 10.14.3
    CPU: (8) x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    npm: 6.4.1 - ~/.nvm/versions/node/v10.14.2/bin/npm
  Languages:
  Browsers:
    Chrome: 72.0.3626.109
    Firefox: 63.0
    Safari: 12.0.3
```

ChristopherBiscardi:
## Description

Component shadowing with a total number of potential shadowings >= 2 (excluding the theme with the component and the site itself, which are absolutely positioned in the order relative to other potential shadow paths) exhibit a bug due to the ordering in which potential theme paths are checked. The order should be reversed. This PR copies the themes array (not strictly necessary, but we shouldn't depend on the assumption that we can safely mutate) then reverses it.

## Related Issues

Fixes #11951

### DIFF
--- a/packages/gatsby/src/internal-plugins/webpack-theme-component-shadowing/index.js
+++ b/packages/gatsby/src/internal-plugins/webpack-theme-component-shadowing/index.js
@@ -66,9 +66,11 @@ module.exports = class GatsbyThemeComponentShadowingResolverPlugin {
         path.join(path.resolve(`.`), `src`, theme),
       ]
         .concat(
-          themes.map(aTheme =>
-            path.join(path.dirname(require.resolve(aTheme)), `src`, theme)
-          )
+          Array.from(themes)
+            .reverse()
+            .map(aTheme =>
+              path.join(path.dirname(require.resolve(aTheme)), `src`, theme)
+            )
         )
         .map(dir => path.join(dir, component))
         .find(possibleComponentPath => {


wardpeet:
I'm unsure if we're already at that state but would be cool if we had some unit tests or integration tests for these functionalities.

DSchau:
Looks good to me!

@wardpeet it's on @ChristopherBiscardi's radar! Should be coming soon!

DSchau:
```
Successfully published:
 - gatsby@2.1.14
```

ChristopherBiscardi:
@wardpeet yep, tests are on my radar. some e2e ones should show up soon to start

PR #11965, opened 2019-02-21
fix(gatsby-plugin-offline): Redesign plugin architecture [draft]

### LINKED ISSUE #11006: [gatsby-plugin-offline] ServiceWorker registration causes a quick blank page on website loading/refresh

davidbailey00:
- Use the IDB to store resource files required for each page's path
- Serve the offline shell if all resources are available; otherwise fetch the page from the network
- Check the correct app-XXX.js file is cached - otherwise use network
- **Serve root HTML for previously-visited pages dynamically, even if the original HTML file has never been downloaded**

Fixes #11006


wardpeet:
@davidbailey00 any updates on this?

davidbailey00:
Hey @wardpeet, after discussing the ideas of this PR with a few people (@DSchau and @Moocar) we've come to the conclusion that it would be better to split this PR up into smaller ones, and that the idea of caching HTML content without downloading the HTML page might not be possible to do reliably (at least at the moment) due to the challenges of getting innerHTML content before any scripts run or any browser extension content is inserted etc. So for now this PR is in "permanent draft" mode.

I've opened up #13502 which includes the navigation handler logic changes without the experimental CSS and HTML capture / insertion into app shell.

davidbailey00:
Closed in favour of #13502

PR #11971, opened 2019-02-21
chore(babel-preset-gatsby): Adds the missing peer dependency on @babel/core

arcanis:
## Description

The Babel plugins have a peer dependency on `@babel/core`, and since `babel-preset-gatsby` doesn't provide it it needs to list it as a peer dependency itself.

### DIFF
--- a/packages/babel-preset-gatsby/package.json
+++ b/packages/babel-preset-gatsby/package.json
@@ -10,6 +10,9 @@
     "@babel/preset-react": "^7.0.0",
     "babel-plugin-macros": "^2.4.2"
   },
+  "peerDependencies": {
+    "@babel/core": "*"
+  },
   "license": "MIT",
   "main": "index.js",
   "scripts": {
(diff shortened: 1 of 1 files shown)


packages/babel-preset-gatsby/package.json
```diff
@@ -10,6 +10,9 @@
     "@babel/preset-react": "^7.0.0",
     "babel-plugin-macros": "^2.4.2"
   },
+  "peerDependencies": {
+    "@babel/core": "*"
```
DSchau: Should we be relatively safer here and pin to `^7.0.0`?

arcanis: Sure! Will update the PR - since the plugins are scoped to `^7.0.0` it should be fine 👍

arcanis:
The fail looks unrelated

DSchau:
@arcanis it is. Looking at it now.

sidharthachatterjee:
Published in babel-preset-gatsby@0.1.8

PR #11972, opened 2019-02-21
chore(gatsby-link): Adds a missing peer dependency on react-dom

arcanis:
## Description

The `@reach/router` package has a peer dependency on `react-dom`, and since `gatsby-link` doesn't provide it it needs to list it as a peer dependency itself.

### DIFF
--- a/packages/gatsby-link/package.json
+++ b/packages/gatsby-link/package.json
@@ -12,6 +12,9 @@
     "@types/reach__router": "^1.0.0",
     "prop-types": "^15.6.1"
   },
+  "peerDependencies": {
+    "react-dom": "*"
+  },
   "devDependencies": {
     "@babel/cli": "^7.0.0",
     "@babel/core": "^7.0.0",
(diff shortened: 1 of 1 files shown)


packages/gatsby-link/package.json
```diff
@@ -12,6 +12,9 @@
     "@types/reach__router": "^1.0.0",
     "prop-types": "^15.6.1"
   },
+  "peerDependencies": {
```
sidharthachatterjee: Want to move this over to the section below? 🙂 There's already `peerDependencies` set at the bottom! Also maybe setting it to map to the react version is a good idea too. So in this case, that would be "react-dom": "^16.3"

DSchau: Nice catch!

Should probably map to the same versions we use in gatsby proper, too.

https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/package.json#L161-L164

arcanis: Updated to `^16.4.2`! 👍

sidharthachatterjee:
Published in gatsby-link@2.0.12

PR #11973, opened 2019-02-21
docs(gatsby-plugin-sitemap): Put important note to top

nextlevelshit:
## Description

The sitemap is only served in production mode. This information is very important, because the first thing I've done was testing it in development mode. Putting this crucial information after the code block could be easily overseen. So, I've put it in a more visible place.


sidharthachatterjee:
@nextlevelshit Seems fair! Could you please run prettier once to fix the linting issue?

PR #11974, opened 2019-02-21
docs(www): Add hooks to static-query docs

KenzoBenzo:
## Description
Referenced from Jason Lengstorf's Gatsby + GraphCMS live-stream webinar.

### DIFF
--- a/docs/docs/static-query.md
+++ b/docs/docs/static-query.md
@@ -44,7 +44,7 @@ export default () => (
 
 Using `StaticQuery`, you can colocate a component with its data. No longer is it required to, say, pass data down from `Layout` to `Header`.
 
-### Hooks Example
+### useStaticQuery
 
 Using the hooks version of static query is very similar. Checkout the documentation of `useStaticQuery` [here!](/docs/use-static-query/)
 
--- a/docs/docs/static-query.md
+++ b/docs/docs/static-query.md
@@ -62,7 +62,6 @@ export default () => {
       }
     }
   `)
-};
 
   return (
     <header>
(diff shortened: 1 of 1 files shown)


wardpeet:
on a side not should we mention `Gatsby version 2.1`, I would just omit it.

marcysutton:
I agree with @DSchau that since there are already two docs, I don't think it makes sense to double up the detailed information. A link from the static-query doc to useStaticQuery would be very helpful, though!

In general, I do think the `Gatsby version 2.1` part is useful, at least for a little while, since useStaticQuery was released in a minor version. But it's already in the useStaticQuery doc, so leaving it out of a mention here would create less work in the future (only one place to update)!

docs/docs/static-query.md
```diff
@@ -44,6 +44,10 @@ export default () => (
 
 Using `StaticQuery`, you can colocate a component with its data. No longer is it required to, say, pass data down from `Layout` to `Header`.
 
+### Hooks Example
```
sidharthachatterjee:
```suggestion
### useStaticQuery
```

docs/docs/static-query.md
```diff
@@ -44,6 +44,10 @@ export default () => (
...
+
+Using the hooks version of static query is very similar. Checkout the documentation of `useStaticQuery` [here!](/docs/use-static-query/)
```
sidharthachatterjee:
```suggestion
There's also a hooks version of StaticQuery called `useStaticQuery`. Check out its documentation [here!](/docs/use-static-query/)
```

PR #11976, opened 2019-02-21
fix(e2e-tests): prevent flaky test runs by waiting for the correct API

### LINKED ISSUE #9233: [CI]: e2e path prefix test seems to intermittently fail

DSchau:
## Description

This fixes the flaky path-prefix test by appropriately invoking the `waitForAPI` helper in various phases of the failing test.

First, a little background:

- This annoying e2e-test fails infrequently, but often enough that people notice it and are confused by it
    - Note: it always fails on the same assertion ("expected blank to equal /blog")
- The test (seems to) always run fine locally, so it seems to be a CI and timing issue

To validate this fix, I `ssh`ed into the CircleCI container, ran the e2e-test _first_ to validate failures (it does fail! see below 👇). It appears to fail once every 10 runs or so. I made _this_ change to the test and then validated that after 30+ runs the e2e-test does not fail.

download

## Related Issues

Fixes #9233

### DIFF
--- a/e2e-tests/path-prefix/cypress/integration/path-prefix.js
+++ b/e2e-tests/path-prefix/cypress/integration/path-prefix.js
@@ -1,4 +1,3 @@
-/* global cy */
 const { pathPrefix } = require(`../../gatsby-config`)
 
 const withTrailingSlash = url => `${url}/`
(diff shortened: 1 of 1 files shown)


DSchau:
Also note: yeah, it seems super weird that this fixes. I think the `cy.go` API is perhaps a little more finicky, so we don't have to waitForAPI in other assertions.

To validate this fix, we'll just first:

1. Ensure this _PR_ passes all tests (of course)
1. SSH into the container (re-run job with SSH), and validate we don't see a failure after running the following script many times

```shell
cd project
./scripts/e2e-test.sh "e2e-tests/path-prefix"
```

DSchau:
> **ci/circleci: e2e_tests_path-prefix ** — Your tests passed on CircleCI!

👀 SSHing to the job now.

DSchau:
🤞 

```
#!bin/sh
for i in {1..50}; do
  ./scripts/e2e-test.sh "e2e-tests/path-prefix"
done

```

Issue #11912, opened 2019-02-19
Refactor tracedSVG calls from transformer-sharp into plugin-sharp


oorestisime:
## Description

Right now tracedSVG support is handled mostly in the transformer plugin rather than the plugin-sharp itself. https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-transformer-sharp/src/extend-node-type.js#L65

It would be better if plugin-sharp takes all the necessary configuration options and creates tracedSVG images when needed.

This would enable third-party plugins to add tracedSVG support without needing to take care of it on themselves

### A bit more context

A bit more context can be found here https://github.com/oorestisime/gatsby-remark-rehype-images/issues/2

## Problem

While the refactoring wouldn't be complicated or hard to do introduce a possible conflict. What happens when newest plugin-sharp is installed but transformer one is on the previous version?
Would the cache take it upon it hence not really create any issues?

pieh:
We talked with @oorestisime on discord a bit about it before he created the issue.

One of potential problem that I initially didn't thought about with this:
SVG tracing is considerably more expensive than base64 placeholder generation. Right now we only do this if user specifically request `tracedSVG` field on `fixed`/`fluid` objects/transformations (either directly or through our fragments).

Checking if `traceSVG` arg actually changes behaviour - because user definitely can do:
```graphql
fluid(maxWidth: 800, traceSVG: { color: "#FF0000" }) {
  ... GatsbyImageSharpFluid
}
```
which doesn't contain `tracedSVG` field and currently wouldn't cause svg tracing - but would if we only checks for args.

But we could definitely provide convience utility that would not cause change for `gatsby-transformer-sharp` by having extra flag in args:

```js
const { fluid } = require(`gatsby-plugin-sharp`)
[...]
fluid({
  ...allUsualArgs,
  tracedSVG: {
    color: "#FF0000"
  },
  generateTracedSVG: true // if this is true and `tracedSVG` options are passed - then generate traced svg
})
```
?

oorestisime:
Yeap sounds fair to me.
and actually pass those flags along from transformer as well?

pieh:
I think transformer would need to stay as-is unless we have access to what subfields are requested at `fluid`/`fixed` resolver level (no idea if we do)

PR #11981, opened 2019-02-22
feat(gatsby-plugin-sharp): Add tracedSVG option in fluid and fixed processors

oorestisime:
Fix: #11912

### DIFF
--- a/packages/gatsby-plugin-sharp/src/__tests__/index.js
+++ b/packages/gatsby-plugin-sharp/src/__tests__/index.js
@@ -319,6 +319,53 @@ describe(`gatsby-plugin-sharp`, () => {
       expect(result).toMatchSnapshot()
     })
   })
+
+  describe(`tracedSVG`, () => {
+    it(`doesn't always run`, async () => {
+      const args = {
+        maxWidth: 100,
+        width: 100,
+        tracedSVG: { color: `#FF0000` },
+      }
+
+      let result = await fixed({
+        file,
+        args,
+      })
+
+      expect(result.tracedSVG).toBeUndefined()
+
+      result = await fluid({
+        file,
+        args,
+      })
+
+      expect(result.tracedSVG).toBeUndefined()
+    })
+
+    it(`runs on demand`, async () => {
+      const args = {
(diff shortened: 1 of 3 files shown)


wardpeet:
Mind adding some tests for it? It's good enough to just test if traceSVG is ran with the correct values. And not ran if it's not needed ^^
https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-sharp/src/__tests__/index.js

oorestisime:
Oups sorry didn't pay attention to tests.

I also kind of avoided duplicate code from my first commit. i am not sure though if it is  explicit enough. or maybe a better function name. let me know

packages/gatsby-plugin-sharp/src/__tests__/index.js
```diff
@@ -319,6 +319,53 @@ describe(`gatsby-plugin-sharp`, () => {
...
+        width: 100,
+        tracedSVG: { color: `#FF0000` },
+      }
+
+      let result = await fixed({
+        file,
+        args,
+      })
+
+      expect(result.tracedSVG).toBeUndefined()
```
sidharthachatterjee: I'd add a jest mock for the `traceSVG` function (it's already exported) and add an assertion that it's not called. What do you think?

oorestisime: Any way i could mock this properly?

thought about

```
const spy = jest.spyOn(rest, `traceSVG`)

    beforeEach(() => {
      spy.mockClear()
    })

    afterAll(() => {
      spy.mockClear()
    })
```

with rest

```
const {
  base64,
  fluid,
  fixed,
  queueImageResizing,
  getImageSize,
  ...rest
} = require(`../`)
```

sidharthachatterjee: This seems trickier than I initially thought

Guess we should merge and iterate on the tests and the code (to make it easier to test) later

sidharthachatterjee:
Published in gatsby-plugin-sharp@2.0.22

PR #11985, opened 2019-02-22
docs: Replace "gatsby develop" with "npm run develop" in second part of tutorial

Daniel15:
I was trying to follow this tutorial, but `gatsby develop` didn't work as I don't have a global version of Gatsby installed. Replaced it with `npm run develop` as that's what the first part of the tutorial uses.

### DIFF
--- a/docs/tutorial/part-two/index.md
+++ b/docs/tutorial/part-two/index.md
@@ -120,7 +120,7 @@ import "./src/styles/global.css"
 3. Start the development server:
 
 ```shell
-gatsby develop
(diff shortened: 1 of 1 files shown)


Daniel15:
I just noticed that other parts of the tutorial mention `gatsby develop` too... I'll do a find-replace across all the tutorial pages tomorrow.

wardpeet:
Sweet! We changed our first tutorial page to npx which is why this is failing.

Thank you for fixing!

PR #11986, opened 2019-02-22
chore(gatsby-source-contentful): Remove stale files

stefanprobst:
The `gatsby-source-contentful` plugin seems to have two unused, transpiled files in its repo. Could someone double check they are indeed not used?


sidharthachatterjee:
Published in gatsby-source-contentful@2.0.31

PR #12001, opened 2019-02-22
feat(gatsby-transformer-react-docgen): use local babel config for react-docgen

jquense:
BREAKING CHANGE: now uses the local babel config with a fallback


jquense:
ping :)

packages/gatsby-transformer-react-docgen/src/__tests__/on-node-create.js
```diff
@@ -114,10 +119,10 @@ describe(`transformer-react-doc-gen: onCreateNode`, () => {
 
   it(`should infer a name`, async () => {
     node.__fixture = `unnamed.js`
-    node.absolutePath = path.join(__dirname, `UnnamedExport`)
+    // node.absolutePath = path.join(__dirname, `UnnamedExport`)
```
DSchau: Is there a reason this is commented out?

DSchau: Should we just remove?

jquense:
ya i'm exploring what is required from the docs, give me a sec i've pushed a version to a scoped package to try out!

jquense:
Ok needed some additional work donw upstream, but thats merged and in 👍 

> If I'm reading correctly, it looks like we now require a cwd and parserOptions passed to the plugin, but we don't document this case.

Not required, only passed through for the sake the tests so that the packages own `.babelrc` doesn't affect the test outcomes.

I've documented the options as well so it's clearer what the behavior is and how to adjust it.

jquense:
ok good to go 👍

jquense:
friendly ping

LekoArts:
I'm also interested in this update, as I'll then be able to use this resolver, right?
https://github.com/Jmeyering/react-docgen-annotation-resolver

Nudge for @DSchau 	👍

jquense:
@LekoArts you can specify a resolver now, so that should already work.

DSchau:
Waiting for the CI checks to pass--then will get this merged in and published as `gatsby-transformer-react-docgen`@`4.0.0`.

DSchau:
Published as [`gatsby-transformer-react-docgen`@`4.0.0`](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-transformer-react-docgen/CHANGELOG.md#400-2019-04-02)

PR #12004, opened 2019-02-22
gatsby-starter-default useStaticQuery over StaticQuery

DylanTackoor:
## Description

Updates gatsby-starter-default [layout.js](https://github.com/gatsbyjs/gatsby/blob/master/starters/default/src/components/layout.js) and [seo.js](https://github.com/gatsbyjs/gatsby/blob/master/starters/default/src/components/seo.js) components to use the new `useStaticQuery` hook.

I think it looks cleaner, thoughts? 🤷‍♂️


sidharthachatterjee:
Thank you @DylanTackoor 

You'll need to update `package.json` and `package-lock.json` to have react and react-dom 16.8+ and gatsby 2.1+ since hooks won't work below that

sidharthachatterjee:
@pieh just made a great point about how our starters serve to enable discovery of various features. Going by that, do you think it would make sense to keep one instance of `StaticQuery` and `useStaticQuery` each?

sidharthachatterjee:
@DylanTackoor You can also update gatsby-starter-blog if you like since they're mostly the same files 🙂

DylanTackoor:
@sidharthachatterjee I'll update dependencies, and having one of each type of query is a great idea. I'll revert one and add in links to StaticQuery and useStaticQuery docs an dod the same for the blog starter 😄

DylanTackoor:
@sidharthachatterjee Done!

PR #12005, opened 2019-02-22
fix(gatsby): normalize component path casing

### LINKED ISSUE #11688: Error: RelayParser: Encountered duplicate defintitions for one or more documents
## Description

Building a blog / portfolio with GatsbyJS and I'm struggling with this issue. I had the website working for the most part but all of a sudden it's stopped working. Really not sure why. The error doesn't make much sense to me. Hoping someone cold help me work out why.

### Steps to reproduce

https://github.com/ChildishhAlbino/childishhalbino.github.io/tree/problem-fixing 

Above is a git repo with the project as I last left it. 
1. run NPM install for node packages
2. Run gatsby develop.
3. Visit localhost/8000

An error screen should appear with the above error.

### Expected result

* Page should load without errors. 
* There should be 3 blog posts with the content from the /posts/ markdown files. 
* The 'about' tab should display 3 'cards' each with information from the /bio/ markdown files.
* The blurb should load with information from the 'blurb.md' file.

There will be warnings. There are some imports that aren't being used but aside from that everything should be fine.

### Actual result

* Page will not load -> returns above error.
* Unsure how else to describe the issue.

### Environment

Run `npx gatsby info --clipboard` in your project directory and paste the output here.

System:
    OS: Windows 10
    CPU: (4) x64 Intel(R) Core(TM) i5-6600K CPU @ 3.50GHz
  Binaries:
    npm: 6.4.1 - D:\Software Development\Languages, Frameworks,
etc\nodeJs\npm.CMD
  Languages:
    Python: 3.7.0 - C:\Users\conno\AppData\Local\Programs\Python\Python37-32\python.EXE
  Browsers:
    Edge: 42.17134.1.0
  npmPackages:
    gatsby: ^2.0.118 => 2.0.118
    gatsby-image: ^2.0.19 => 2.0.29
    gatsby-plugin-buildtime-timezone: ^1.0.3 => 1.0.3
    gatsby-plugin-manifest: ^2.0.7 => 2.0.17
    gatsby-plugin-offline: ^2.0.23 => 2.0.23
    gatsby-plugin-react-helmet: ^3.0.1 => 3.0.6
    gatsby-plugin-sass: ^2.0.2 => 2.0.10
    gatsby-plugin-sharp: ^2.0.11 => 2.0.20
    gatsby-source-filesystem: ^2.0.6 => 2.0.20
    gatsby-transformer-remark: ^2.2.4 => 2.2.4
    gatsby-transformer-sharp: ^2.1.7 => 2.1.13

error The system cannot find the path specified.

  Error: The system cannot find the path specified.

  - envinfo.js:1 Function.e.exports.sync
    [childishhalbino.github.io]/[envinfo]/dist/envinfo.js:1:4970
  - envinfo.js:1 Object.copySync
    [childishhalbino.github.io]/[envinfo]/dist/envinfo.js:1:6694    8

  - envinfo.js:1 Object.t.writeSync.e [as writeSync]
    [childishhalbino.github.io]/[envinfo]/dist/envinfo.js:1:4877    7

  - envinfo.js:1
    [childishhalbino.github.io]/[envinfo]/dist/envinfo.js:1:4693    4

  - envinfo.js:1 Promise.all.then.e
    [childishhalbino.github.io]/[envinfo]/dist/envinfo.js:1:4694    9

  - util.js:16 tryCatcher
    [childishhalbino.github.io]/[bluebird]/js/release/util.js:16    :23

  - promise.js:512 Promise._settlePromiseFromHandler
    [childishhalbino.github.io]/[bluebird]/js/release/promise.js    :512:31

  - promise.js:569 Promise._settlePromise
    [childishhalbino.github.io]/[bluebird]/js/release/promise.js    :569:18

  - promise.js:614 Promise._settlePromise0
    [childishhalbino.github.io]/[bluebird]/js/release/promise.js    :614:10

  - promise.js:694 Promise._settlePromises
    [childishhalbino.github.io]/[bluebird]/js/release/promise.js    :694:18

  - promise.js:638 Promise._fulfill
    [childishhalbino.github.io]/[bluebird]/js/release/promise.js    :638:18

  - promise_array.js:126 PromiseArray._resolve
    [childishhalbino.github.io]/[bluebird]/js/release/promise_ar    ray.js:126:19

  - promise_array.js:144 PromiseArray._promiseFulfilled
    [childishhalbino.github.io]/[bluebird]/js/release/promise_ar    ray.js:144:14

  - promise.js:574 Promise._settlePromise
    [childishhalbino.github.io]/[bluebird]/js/release/promise.js    :574:26

  - promise.js:614 Promise._settlePromise0
    [childishhalbino.github.io]/[bluebird]/js/release/promise.js    :614:10

  - promise.js:694 Promise._settlePromises
    [childishhalbino.github.io]/[bluebird]/js/release/promise.js    :694:18

error UNHANDLED REJECTION

  Error: The system cannot find the path specified.

  - envinfo.js:1 Function.e.exports.sync
    [childishhalbino.github.io]/[envinfo]/dist/envinfo.js:1:4970
  - envinfo.js:1 Object.copySync
    [childishhalbino.github.io]/[envinfo]/dist/envinfo.js:1:6694    8

  - envinfo.js:1 Object.t.writeSync.e [as writeSync]
    [childishhalbino.github.io]/[envinfo]/dist/envinfo.js:1:4877    7

  - envinfo.js:1
    [childishhalbino.github.io]/[envinfo]/dist/envinfo.js:1:4693    4

  - envinfo.js:1 Promise.all.then.e
    [childishhalbino.github.io]/[envinfo]/dist/envinfo.js:1:4694    9

  - util.js:16 tryCatcher
    [childishhalbino.github.io]/[bluebird]/js/release/util.js:16    :23

  - promise.js:512 Promise._settlePromiseFromHandler
    [childishhalbino.github.io]/[bluebird]/js/release/promise.js    :512:31

  - promise.js:569 Promise._settlePromise
    [childishhalbino.github.io]/[bluebird]/js/release/promise.js    :569:18

  - promise.js:614 Promise._settlePromise0
    [childishhalbino.github.io]/[bluebird]/js/release/promise.js    :614:10

  - promise.js:694 Promise._settlePromises
    [childishhalbino.github.io]/[bluebird]/js/release/promise.js    :694:18

  - promise.js:638 Promise._fulfill
    [childishhalbino.github.io]/[bluebird]/js/release/promise.js    :638:18

  - promise_array.js:126 PromiseArray._resolve
    [childishhalbino.github.io]/[bluebird]/js/release/promise_ar    ray.js:126:19

  - promise_array.js:144 PromiseArray._promiseFulfilled
    [childishhalbino.github.io]/[bluebird]/js/release/promise_ar    ray.js:144:14

  - promise.js:574 Promise._settlePromise
    [childishhalbino.github.io]/[bluebird]/js/release/promise.js    :574:26

  - promise.js:614 Promise._settlePromise0
    [childishhalbino.github.io]/[bluebird]/js/release/promise.js    :614:10

  - promise.js:694 Promise._settlePromises
    [childishhalbino.github.io]/[bluebird]/js/release/promise.js    :694:18

pieh:
If user create page with programmatic `createPage` and this path doesn't fully match exact casing that is on system, relay-compiler throws:

```
RelayParser: Encountered duplicate defintitions for one or more documents
```

That's because in our [query-compiler](https://github.com/gatsbyjs/gatsby/blob/93bbe2b0600c24d6c1d79f007cab34e6f68ce18c/packages/gatsby/src/internal-plugins/query-runner/query-compiler.js#L100) we grab every js files from source directory, add page templates and run resulting array through `_.uniq`. If there is case missmatch between auto extract files from `src` directory and page component path, we end up with duplicates, causing same query to be extracted twice.

This adds sanity check, automatically adjust casing if it doesn't match exactly and show warning (as casing differences will cause CI/CD that uses case sensitive file system to fail with "file doesn't exist" error.

Example output:
screenshot 2019-02-22 at 22 18 09

Fixes #11688


packages/gatsby/src/redux/actions.js
```diff
@@ -195,6 +198,39 @@ ${reservedFields.map(f => `  * "${f}"`).join(`\n`)}
...
+              if (letter !== trueComponentPath[index]) {
+                return `^`
+              }
+              return ` `
+            })
+            .join(``)
+
+          report.warn(
+            stripIndent`
+          ${name} created a page with a component path that doesn't match casing of file. This will work locally, but will break on Linux (on most CI/CD pipelines).
```
DSchau:
```suggestion
          ${name} created a page with a component path that doesn't match the casing of the actual file. This may work locally, but will break on systems which are case-sensitive, e.g. most CI/CD pipelines.
```

DSchau: nice syntax highlighting ☝️  haha

packages/gatsby/src/redux/actions.js
```diff
@@ -195,6 +198,39 @@ ${reservedFields.map(f => `  * "${f}"`).join(`\n`)}
...
+      // normalize component path
+      page.component = slash(page.component)
+      // check if path uses correct casing - incorrect casing will
+      // casue issues in query compiler and incosistencies when
+      // developing on Mac or Windows and trying to deploy from
+      // linux CI/CD pipeline
+      const trueComponentPath = slash(truePath(page.component))
+      if (trueComponentPath !== page.component) {
+        if (!hasWarnedForPageComponentInvalidCasing.has(page.component)) {
+          const markers = page.component
```
DSchau: Slick idea here 🎉

packages/gatsby/src/redux/actions.js
DSchau: Do we want to add a gatsby.app (or gatsby.dev) short link for this? Worth documenting?

packages/gatsby/src/redux/actions.js
```diff
@@ -195,6 +198,39 @@ ${reservedFields.map(f => `  * "${f}"`).join(`\n`)}
       console.log(``)
       console.log(page)
       noPageOrComponent = true
+    } else if (page.component) {
+      // normalize component path
+      page.component = slash(page.component)
+      // check if path uses correct casing - incorrect casing will
+      // casue issues in query compiler and incosistencies when
```
DSchau: nit

```suggestion
      // cause issues in query compiler and inconsistencies when
```

packages/gatsby/src/redux/actions.js
pieh: I think this hopefully is more selfexplaining than async `createPages` warning, so I would personally skip creating dedicated page for it (we just try to adjust file casing and pointing where exactly it needs to change)

packages/gatsby/src/redux/actions.js
```diff
@@ -195,6 +198,39 @@ ${reservedFields.map(f => `  * "${f}"`).join(`\n`)}
...
       console.log(page)
       noPageOrComponent = true
+    } else if (page.component) {
+      // normalize component path
+      page.component = slash(page.component)
+      // check if path uses correct casing - incorrect casing will
+      // cause issues in query compiler and inconsistencies when
+      // developing on Mac or Windows and trying to deploy from
+      // linux CI/CD pipeline
+      const trueComponentPath = slash(truePath(page.component))
```
pieh: I was wondering if I should memoize `truePath` as it does use `glob.sync` under the hood so `fs` checking might be excessive

packages/gatsby/src/redux/actions.js
DSchau: Yeah - agreed. I've been thinking it may be a helpful thing to always have a documentation page for errors, but perhaps that's overkill (especially when the error message is self-documenting, as it is here);.

KyleAMathews:
Nice check!

sidharthachatterjee:
Nicely done, @pieh 💯

sidharthachatterjee:
Published in gatsby@2.1.18

rmehner:
🐞 This introduced a breaking bug for us, as true-case-path trips on umlauts.

We had a file in `src/pages` named `röck-döts.js` and it causes the build to break with:

`TypeError: Cannot read property 'replace' of undefined`

I'm still in the process of hunting this down, so far my reproduction case looks like this:

```js
const trueCase = require('true-case-path')

const filePaths = [
  '/Users/rmehner/gatsby-site/src/pages/roeck-doets.js', // this works fine
  '/Users/rmehner/gatsby-site/src/pages/röck-döts.js', // this will yield undefined
]

filePaths.forEach(filePath => {
  console.log(`${filePath}: `, trueCase(filePath))
})
```

(this is on an APFS case-insensitive filesystem on current macOS, might behave differently on something else)

Temporary solution for us is to go back to Gatsby 2.1.17 or to remove the umlauts.

I'll report back once I know more about the root cause.

rmehner:
Okay, found the root cause.

This line: https://github.com/barsh/true-case-path/blob/a1623002cf442ba7b8bb91c6a981cc46c887b85b/index.js#L16

Problem is that since macOS 10.13 the underlying default file system for Macs is no longer HFS+ but APFS (except for non-ssd macs). Apparently they no longer store that as NFD, so it does not need to be normalized.

So as long as this is not fixed upstream, we can't update to a new gatsby version for our gatsby projects (since the projects are in Switzerland, many of them have to deal with Umlauts ;))

Unfortunately the project seems to be not actively maintained, but I've tried my luck here: https://github.com/barsh/true-case-path/issues/9

Let's see where this goes.

DSchau:
@rmehner whoa - sorry about that. Excellent work debugging that behavior and tracing the bug upstream.

We could:

1. Fork and release a fix (if no traction after a day or two on that PR)
1. Introduce some fallback or bail-out if we get undefined for a path

Let's wait a _bit_ to see if we can get the fix upstream, otherwise we can investigate either of the above options.

rmehner:
@DSchau No replies yet. This is somewhat of a gnarly problem, as to properly fix this we'd need to test it on many different file systems. Unless we find another, more compatible approach.

I do have an old mac at home that I could use to test this on HFS+ to make sure it both works on old and new macOS versions, but I have no idea how to do that within an CI environment.

rmehner:
@DSchau there's still no activity on the issue, so I'm not sure what's the best way going forward would be. We're still blocked on upgrading gatsby for that reason.

There's a the V2 PR (https://github.com/barsh/true-case-path/pull/8) that introduces some breaking changes, but fixes the issue (but I can't test this on HFS+ devices).

I'd propose that we wait until mid of next week and if there's still no traction on the other project, I'd fork it and release that version to npm with all attributions and create another PR on gatsby to use this instead of the old version. Or can you think of a better way?

PR #12010, opened 2019-02-23
fix(gatsby): theme component shadow fails when extension is used

jbolda:
## Description
[Imports that include an extension](https://github.com/jbolda/gatsby-theme-bulma/blob/de293e0a417f5f391fef93460a90eb1ae18fcfb0/packages/gatsby-theme-bulma-homepage/src/Hero/HeroTemplateWithArticles.js#L2) will fail component shadowing. Add the extension to `path.basename` that imports with or without will work.

## Related Issues
None


sidharthachatterjee:
Published in gatsby@2.1.18

PR #12016, opened 2019-02-23
Improve the external link checking to prevent false positives behaviour

bartekus:
## Description
This change resolves the issue that was affecting [overreacted.io](https://github.com/gaearon/overreacted.io), and causing false positives warnings in the console.

The implementation uses try-catch guard due to incompatibility with IE, and as such, will fall back to the old way when run on IE.

### DIFF
--- a/packages/gatsby-link/src/index.js
+++ b/packages/gatsby-link/src/index.js
@@ -109,11 +109,22 @@ class GatsbyLink extends React.Component {
       ...rest
     } = this.props
 
-    const LOCAL_URL = /^\/(?!\/)/
-    if (process.env.NODE_ENV !== `production` && !LOCAL_URL.test(to)) {
-      console.warn(
-        `External link ${to} was detected in a Link component. Use the Link component only for internal links. See: https://gatsby.app/internal-links`
-      )
+    if (process.env.NODE_ENV !== `production`) {
+      const LOCAL_URL = /^\/(?!\/)/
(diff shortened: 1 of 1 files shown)


packages/gatsby-link/src/index.js
```diff
@@ -109,11 +109,22 @@ class GatsbyLink extends React.Component {
...
-    if (process.env.NODE_ENV !== `production` && !LOCAL_URL.test(to)) {
-      console.warn(
-        `External link ${to} was detected in a Link component. Use the Link component only for internal links. See: https://gatsby.app/internal-links`
-      )
+    if (process.env.NODE_ENV !== `production`) {
+      const LOCAL_URL = /^\/(?!\/)/
+      const message = `External link ${to} was detected in a Link component. Use the Link component only for internal links. See: https://gatsby.app/internal-links`
+
+      try {
+        const link = new URL(to, window.location)
```
sidharthachatterjee: This is problematic because it'll always revert to the old behaviour (inside the `catch`) during server side rendering since `window` will not be accessible

wardpeet:
Thanks so much for the PR!

Unfortunately, we're not going to merge this at this time. I can't seem to reproduce it or find an issue related to this PR.

As such--we're going to close this out. Feel free to re-open with a linked issue or any information. We want you to know that we certainly want to keep you as a contributor to Gatsby, so we'd encourage you to check out our open issues.

PR #12026, opened 2019-02-24
update slash to 2.0.0

rohintangirala:
## Description

Update slash version to 2.0.0 in examples.


pieh:
Are there any breaking changes between v1 and v2 that could potentially affect us? We also use `slash` in gatsby core package - https://github.com/gatsbyjs/gatsby/blob/fd12550dae58becdc52057d8b43f1eb6960a6883/packages/gatsby/package.json#L114

So this change would cause installing both slash@1 and slash@2 in the examples

Issue #11997, opened 2019-02-22
[gatsby-plugin-sharp] Cache traceSVG calculations


oorestisime:
Now that plugin-sharp supports traceSVG for fluid and fix processing (#11912) it should probably also 
get the cache mechanism to avoid re-calculations.

Right now we memoize the function but it could be better. a similar feature for this was implemented not a long time ago for base64 #6999

PR #9059, opened 2018-10-12
feat(gatsby-plugin-sharp): cache base64 if possible

oorestisime:
Closes: #6999

PR #12044, opened 2019-02-24
fix(*): cache tracedSVG calculations when cache is present

oorestisime:
Fix: #11997

Similar to #9059 but for tracedSVG.

### DIFF
--- a/packages/gatsby-plugin-sharp/src/index.js
+++ b/packages/gatsby-plugin-sharp/src/index.js
@@ -10,7 +10,7 @@ const { scheduleJob } = require(`./scheduler`)
 const { createArgsDigest } = require(`./process-file`)
 const { reportError } = require(`./report-error`)
 const { getPluginOptions, healOptions } = require(`./plugin-options`)
-const { memoizedTraceSVG } = require(`./trace-svg`)
+const { memoizedTraceSVG, notMemoizedtraceSVG } = require(`./trace-svg`)
 
 const imageSizeCache = new Map()
 const getImageSize = file => {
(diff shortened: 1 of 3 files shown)


wardpeet:
sweet, mind adding some tests to make sure your cachekeys are working

oorestisime:
hey, thanks for the feedback! I am not sure i follow here.
Is there a test somewhere doing this? Doesn't sound good to me to export the cache key generator.
But i might be seeing this completely wrong so a few more details would be appreciated!

wardpeet:
oh sorry, more or less I mean if we call base64 twice it is only ran once.

oorestisime:
Heya,

doing so would produce the same result but without actually testing cache keys. I won't be able to pass a cache object in the arguments since there is no way to import the cache as a standalone (i think).
Only solution i see here is mocking the cache object, although it seems that doing so beats the purpose of the test

LekoArts:
Would be cool to have this in! Any follow-up questions/additions from your side @wardpeet ?

oorestisime:
@wardpeet kind reminder :)

wardpeet:
@oorestisime oh my bad. It loosk great, can you fix conflicts and we'll merge! 👍

oorestisime:
Rebased :) please have a close look i hope i did a correct merge, there were a lot of changes

wardpeet:
Patched a few things. I'll be testing this one and getting this in. Thanks for your patience and rebasing it!

oorestisime:
Yeah sorry i did this quickly because i ll be with no pc for a week so i didn't want to get it stalled! thanks for taking over !

PR #12054, opened 2019-02-25
fix(gatsby-source-filesystem): Do not re-download cached files from createRemoteFileNode

### LINKED ISSUE #12053: createRemoteFileNode re-downloads unchanged files (the `If-None-Match` header is not being sent)
## Description

When using `createRemoteFileNode` from `gatsby-source-filesystem`, remote files get downloaded each time `yarn develop` is restarted (or each build) even though the remote files are unchanged and the cache is present.

There are multiple issues in `createRemoteFileNode`.

1. The `headers` option is not being set properly on the `got` HTTP request and, thus, the `If-None-Match` header is not getting sent.
2. When (1) is fixed, you'll find that the cached response header gets overwritten by the 304 response headers and you lose the Etag.  We should only encache response headers from a 200 response.

I'm uncertain how to wrap this up in tests but both patches are trivial and I can submit PR.

### Steps to reproduce

1. Create a new project
2. Add the following to `gatsby-node.js`

```
const { createRemoteFileNode } = require('gatsby-source-filesystem');

exports.sourceNodes = async ({ actions, store, cache, createNodeId }) => {

  fileNode = await createRemoteFileNode({
    url: "https://www.nasa.gov/centers/goddard/images/content/638831main_globe_east_2048.jpg",
    cache,
    store,
    createNode,
    createNodeId
  });
}
```
3. `yarn develop`
4. Observe the file will download and you can check in cache
```
$ ls -alt .cache/gatsby-source-filesystem/0d5a6ba9af7781580b3c59af9e560259/638831main_globe_east_2048.jpg
-rw-r--r--  1 blake  staff  536295 Feb 24 21:23 .cache/gatsby-source-filesystem/0d5a6ba9af7781580b3c59af9e560259/638831main_globe_east_2048.jpg
$ cksum .cache/gatsby-source-filesystem/0d5a6ba9af7781580b3c59af9e560259/638831main_globe_east_2048.jpg
1844973668 536295 .cache/gatsby-source-filesystem/0d5a6ba9af7781580b3c59af9e560259/638831main_globe_east_2048.jpg
```

5. Wait a few minutes and restart `yarn develop`.
6. Check the cached remote files

```
$ ls -alt .cache/gatsby-source-filesystem/0d5a6ba9af7781580b3c59af9e560259/638831main_globe_east_2048.jpg
-rw-r--r--  1 blake  staff  536295 Feb 24 21:24 .cache/gatsby-source-filesystem/0d5a6ba9af7781580b3c59af9e560259/638831main_globe_east_2048.jpg
$ cksum .cache/gatsby-source-filesystem/0d5a6ba9af7781580b3c59af9e560259/638831main_globe_east_2048.jpg
1844973668 536295 .cache/gatsby-source-filesystem/0d5a6ba9af7781580b3c59af9e560259/638831main_globe_east_2048.jpg
```

### Expected result

I expect the cached file to be unchanged after building (i.e., not downloaded again).

The request should send the `If-None-Match` header, the response should be a 304, and the cached file unchanged.  Subsequent restarts should continue to use the cached file.

### Actual result

The file gets re-downloaded.  You can confirm by checking the file mtime.  Secondly, I've confirmed with logging in createRemoteFileNode that the file is being re-downloaded.

### Environment

```

  System:
    OS: macOS 10.14.3
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 11.10.0 - /usr/local/bin/node
    npm: 6.7.0 - /usr/local/bin/npm
  Languages:
    Python: 2.7.13 - /usr/local/bin/python
  Browsers:
    Chrome: 72.0.3626.109
    Firefox: 65.0.1
    Safari: 12.0.3
  npmPackages:
    gatsby: ^2.1.17 => 2.1.17
    gatsby-plugin-manifest: ^2.0.19 => 2.0.19
    gatsby-source-filesystem: ^2.0.22 => 2.0.22
```

skinandbones:
## Description

Fixes multiple issues in `createRemoteFileNode` that is causing the cache mechanism to not work correctly:

1. Pass the headers option in `got()` correctly.
2. Only encache response headers on success responses (so we don't overwrite the Etags – these are not returned in the 304 responses).

## Related Issues

Fixes #12053 

There are some issues reported in `gatsby-source-wordpress` about large downloads on restarts/builds that sound like they could be related.

### DIFF
--- a/packages/gatsby-source-filesystem/src/create-remote-file-node.js
+++ b/packages/gatsby-source-filesystem/src/create-remote-file-node.js
@@ -145,7 +145,7 @@ async function pushToQueue(task, cb) {
 const requestRemoteNode = (url, headers, tmpFilename) =>
   new Promise((resolve, reject) => {
     const responseStream = got.stream(url, {
-      ...headers,
+      headers,
       timeout: 30000,
       retries: 5,
     })
--- a/packages/gatsby-source-filesystem/src/create-remote-file-node.js
+++ b/packages/gatsby-source-filesystem/src/create-remote-file-node.js
@@ -225,8 +225,11 @@ async function processRemoteNode({
 
   // Fetch the file.
   const response = await requestRemoteNode(url, headers, tmpFilename)
-  // Save the response headers for future requests.
-  await cache.set(cacheId(url), response.headers)
+
+  if (response.statusCode == 200) {
+    // Save the response headers for future requests.
+    await cache.set(cacheId(url), response.headers)
+  }
 
   // If the user did not provide an extension and we couldn't get one from remote file, try and guess one
   if (ext === ``) {


packages/gatsby-source-filesystem/src/create-remote-file-node.js
```diff
@@ -145,7 +145,7 @@ async function pushToQueue(task, cb) {
 const requestRemoteNode = (url, headers, tmpFilename) =>
   new Promise((resolve, reject) => {
     const responseStream = got.stream(url, {
-      ...headers,
+      headers,
```
sidharthachatterjee: Nice catch!

packages/gatsby-source-filesystem/src/create-remote-file-node.js
```diff
@@ -225,8 +225,11 @@ async function processRemoteNode({
 
   // Fetch the file.
   const response = await requestRemoteNode(url, headers, tmpFilename)
-  // Save the response headers for future requests.
-  await cache.set(cacheId(url), response.headers)
+
+  if (response.statusCode == 200) {
```
sidharthachatterjee: Do we need to handle any other successful response status codes? Like `201` etc

I noticed that we also check for only `200` later in the file when moving temp data over so I'd guess we should be okay.

sidharthachatterjee:
Merging this in because I think it's good for now with the check for 200. We'll revisit this in case we find that it's not. 

Still a great fix of course! Thank you @skinandbones 👍

PR #12062, opened 2019-02-25
fix(gatsby): Adapt Loki workaround for eq:null

stefanprobst:
The previous Loki workaround for `eq:null` only worked for non-indexed fields. Not 100% sure how loose equality is treated differently on indexed fields, but this at least fixes the tests in the schema refactor branch. For the future: we should always expand `null` to `[null, undefined]`, not only for `eq` queries.


freiksenet:
@Moocar LGTM, sensible for you too?

Moocar:
Yep, LGTM

PR #12081, opened 2019-02-25
fix(gatsby-plugin-manifest): Fix incorrect favicons size bug

### LINKED ISSUE #12051: [gatsby-plugin-manifest] Rectangular canvas svg generates incorrect favicons
## Description

When you provide an svg with rectangular canvas (not square), gatsby-plugin-manifest produces malformed favicons, e.g. 144x130, which then results in an error in the browser.

### Steps to reproduce

Add rectangular svg (see below) to your project and modify gatsby-config.js to point to that svg
Set gatsby-config.js:
```javascript
    {
      options: {
        name: `favicon bug`,
        short_name: `bug`,
        start_url: `/`,
        background_color: `#ffffff`,
        theme_color: `#ca9e67`,
        display: `minimal-ui`,
        icon: `src/images/mountains.svg`, // This path is relative to the root of the site.
        include_favicon: true
      },
    },
```

Save svg to src/images/mountains.svg:
```xml
<?xml version="1.0" encoding="utf-8"?>
<!-- Generator: Adobe Illustrator 23.0.1, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
<svg version="1.1" class="mountains" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
	 viewBox="0 0 47.5 32.9" style="enable-background:new 0 0 47.5 32.9;" xml:space="preserve">
<polygon points="0,32.9 16.5,6.6 18.9,10.5 17.9,11.8 16.5,9.6 2.8,31.3 44.8,31.4 27.7,2.9 15.3,22.9 24.7,22.9 21.3,17.2 22.3,16 
	27.5,24.4 12.4,24.4 27.7,0 47.5,32.9 " style="fill: #CA9E67"/>
</svg>
```

### Expected result

There are two options, either correctly sized favicons should be produced or the build should emit warning or error.

### Actual result

Everything goes smoothly until you find an error in your console

> Error while trying to use the following icon from the Manifest: http://localhost:8000/icons/icon-144x144.png (Resource size is not correct - typo in the Manifest?)

### Environment

Run `gatsby develop` and observe the error in Chrome.

keidrun:
Fix #12051 issue

## Description

Changed the shape of favicons exported to a square when even providing rectangular images

## Related Issues

Fixes #12051

### DIFF
--- a/packages/gatsby-plugin-manifest/src/gatsby-node.js
+++ b/packages/gatsby-plugin-manifest/src/gatsby-node.js
@@ -75,14 +75,14 @@ exports.onPostBootstrap = async ({ reporter }, pluginOptions) => {
 
     let sharpIcon = sharp(icon)
 
-    sharpIcon.metadata().then(metadata => {
-      if (metadata.width !== metadata.height) {
-        reporter.warn(
-          `The icon(${icon}) you provided to 'gatsby-plugin-manifest' is not square.\n` +
-            `The icons we generate will be square and for the best results we recommend you provide a square icon.\n`
-        )
-      }
-    })
+    let metadata = await sharpIcon.metadata()
+
+    if (metadata.width !== metadata.height) {
+      reporter.warn(
+        `The icon(${icon}) you provided to 'gatsby-plugin-manifest' is not square.\n` +
+          `The icons we generate will be square and for the best results we recommend you provide a square icon.\n`
+      )
+    }
 
     //add cache busting
     const cacheMode =
--- a/packages/gatsby-plugin-manifest/src/gatsby-node.js
+++ b/packages/gatsby-plugin-manifest/src/gatsby-node.js
@@ -33,13 +33,17 @@ function generateIcons(icons, srcIcon) {
     // Sharp accept density from 1 to 2400
     const density = Math.min(2400, Math.max(1, size))
     return sharp(srcIcon, { density })
-      .resize(size, size)
+      .resize({
+        width: size,
+        height: size,
+        fit: `contain`,
+        background: { r: 255, g: 255, b: 255, alpha: 0 },
+      })
       .toFile(imgPath)
-      .then(() => {})
   })
 }
 
-exports.onPostBootstrap = async (args, pluginOptions) => {
+exports.onPostBootstrap = async ({ reporter }, pluginOptions) => {
   const { icon, ...manifest } = pluginOptions
 
   // Delete options we won't pass to the manifest.webmanifest.
@@ -68,6 +72,19 @@ exports.onPostBootstrap = async (args, pluginOptions) => {
     if (!doesIconExist(icon)) {
       throw `icon (${icon}) does not exist as defined in gatsby-config.js. Make sure the file exists relative to the root of the site.`
     }
+
+    let sharpIcon = sharp(icon)
+
+    sharpIcon.metadata().then(metadata => {
+      if (metadata.width !== metadata.height) {
+        reporter.warn(
+          `The icon(${icon}) you provided to 'gatsby-plugin-manifest' is not square. \n` +
+            `The icons we generate will be square and for the best results we recommend you provide a square icon. \n` +
+            `We'll do our best with what you provide :)`
+        )
+      }
+    })
+
 
     //add cache busting
     const cacheMode =
--- a/packages/gatsby-plugin-manifest/src/gatsby-node.js
+++ b/packages/gatsby-plugin-manifest/src/gatsby-node.js
@@ -78,9 +78,8 @@ exports.onPostBootstrap = async ({ reporter }, pluginOptions) => {
     sharpIcon.metadata().then(metadata => {
       if (metadata.width !== metadata.height) {
         reporter.warn(
-          `The icon(${icon}) you provided to 'gatsby-plugin-manifest' is not square. \n` +
-            `The icons we generate will be square and for the best results we recommend you provide a square icon. \n` +
-            `We'll do our best with what you provide :)`
+          `The icon(${icon}) you provided to 'gatsby-plugin-manifest' is not square.\n` +
+            `The icons we generate will be square and for the best results we recommend you provide a square icon.\n`
         )
       }
     })
--- a/packages/gatsby-plugin-manifest/src/gatsby-node.js
+++ b/packages/gatsby-plugin-manifest/src/gatsby-node.js
@@ -33,7 +33,7 @@ function generateIcons(icons, srcIcon) {
     // Sharp accept density from 1 to 2400
     const density = Math.min(2400, Math.max(1, size))
     return sharp(srcIcon, { density })
-      .resize(size)
+      .resize(size, size)
       .toFile(imgPath)
       .then(() => {})
   })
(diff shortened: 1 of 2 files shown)


moonmeister:
@DSchau According to the docs if you leave out one dimension (as it is without this change) the second one is assumed to be the same. https://sharp.dimens.io/en/stable/api-resize/#parameters MY guess is there is a bug with SVGs that makes this untrue.

DSchau:
@moonmeister I don't think that's really how I am understanding that document, and just validated with an example.

> Use null or undefined to auto-scale the height to match the width

That means it'll scale both dimensions equally (cover) (e.g. if you had a 50x100 image, if you resized it to 50 height, it'd be 25x50).

[Here's an example](https://github.com/DSchau/sharp-resize-example)

[Output with width and height](https://github.com/DSchau/sharp-resize-example/blob/master/output.png)
[Output with just width](https://github.com/DSchau/sharp-resize-example/blob/master/output-without-width.png)

You can see when specifying both width/height it seems to trim the edges.

moonmeister:
@DSchau Ah, okay, I'm glad you actually tested that. Do we need to change the fit option to fix that? Given the default is `cover` your results make sense. Maybe it should be `contain` or `inside` or `outside`, all seem like they may be what we want.

DSchau:
I _don't_ think we want those! I think we should just warn here, rather than making an opinionated choice for the user.

I suppose we could _warn_ and resize (with one of those options) but either way, I think it should intentionally be more "noisy" here so the user is aware!

moonmeister:
@DSchau I think I'm not understanding why you're concerned about this. I don't mind warning the user if we do something. If a non-square logo is provided it needs to be made square to work...I've never seen a manifest, ios icon, or favicon that wan't square...so I assume this is "expected". Should we warn that they need to provide a square icon or just make it fit into a square with a warning?

DSchau:
@moonmeister sure!

If you accidentally provide a non-square input file (e.g. you take your company's logo.svg, which has dimensions of 100x200), I don't think you'd want behavior of either cutting off or "stretching" your logo to fill the container. If we did default to doing either of those, we should warn when the image is non-square and that it'll be resized.

For the why--branding, brand guidelines, etc. etc.

moonmeister:
@DSchau Okay, Yeah, I think we can scale it (not stretch or cut off) the non-square logo into a square logo (with a warning)...maybe I'm misunderstanding their descriptions of the fit options but that seems like what they do. the Default (cover) cut's off (we don't want this), fill appears to stretch the logo to make it fit, and inside, outside, contain all create an image that's square (based on provided dimensions) with the logo not cut or stretched inside.

DSchau:
contain seems promising! Want to validate with that repo I posted above?

moonmeister:
@DSchau I'm working on some other things atm but if I get time I will

DSchau:
@keidrun any interest in addressing some of the comments above? Specifically:

1. I'm not sure users will want this behavior _automatically_ and a warning may be better suited
1. If there is a way to perform this work programmatically (with a nice appearance, e.g. not stretching), then we can do that--just not sure what's possible here!

Thanks for your patience!

keidrun:
@DSchau No such a good idea. I just think the behavior automatically with stretching like http://tools.dynamicdrive.com/favicon/  is better than only warning for me but if most people think a way of a warning is better, we should choose the latter

moonmeister:
@DSchau Okay, I took a couple minutes to do some experimenting with your test! 

my recommendation would be we do:

```js
.resize({
      width: 128,
      height: 128,
      fit: "contain",
      background: { r: 255, g: 255, b: 255, alpha: 0 }
    })

```

We would also throw a warning in the console letting the dev know this expects a square input but we've done our best with what was provided.

This results in containing the provided image within the set dimensions and instead of having black bars it adds transparent bars. This makes the image "look" unchanged while also forcing it to comply.

moonmeister:
@DSchau Rebased, linted, etc. Let me know what you think.

packages/gatsby-plugin-manifest/src/gatsby-node.js
```diff
@@ -69,6 +73,18 @@ exports.onPostBootstrap = async (args, pluginOptions) => {
...
     }
 
+    let sharpIcon = sharp(icon)
+
+    sharpIcon.metadata().then(metadata => {
+      if (metadata.width !== metadata.height) {
+        reporter.warn(
+          `The icon(${icon}) you provided to 'gatsby-plugin-manifest' is not square. \n` +
+            `The icons we generate will be square and for the best results we recommend you provide a square icon. \n` +
+            `We'll do our best with what you provide :)`
```
wardpeet: I would drop this sentence but maybe @DSchau thinks differently.
```suggestion
            `We'll do our best with what you provide :)`
```

packages/gatsby-plugin-manifest/src/gatsby-node.js
```diff
@@ -69,6 +73,18 @@ exports.onPostBootstrap = async (args, pluginOptions) => {
       throw `icon (${icon}) does not exist as defined in gatsby-config.js. Make sure the file exists relative to the root of the site.`
     }
 
+    let sharpIcon = sharp(icon)
+
+    sharpIcon.metadata().then(metadata => {
+      if (metadata.width !== metadata.height) {
+        reporter.warn(
+          `The icon(${icon}) you provided to 'gatsby-plugin-manifest' is not square. \n` +
```
wardpeet: remove the space on the end before ` \n`
```suggestion
          `The icon(${icon}) you provided to 'gatsby-plugin-manifest' is not square.\n` +
```

packages/gatsby-plugin-manifest/src/gatsby-node.js
```diff
@@ -69,6 +73,18 @@ exports.onPostBootstrap = async (args, pluginOptions) => {
...
+            `The icons we generate will be square and for the best results we recommend you provide a square icon. \n` +
```
wardpeet: remove the space on the end before ` \n`
```suggestion
            `The icons we generate will be square and for the best results we recommend you provide a square icon.\n` +
```

packages/gatsby-plugin-manifest/src/gatsby-node.js
DSchau: Agreed with ya! Let's drop this!

moonmeister: So I'm open to reword this but I feel something is needed. The previous sentences make it sound like something is broken or wrong. But that's not the case, the generation has suceeded. We're just giving them a heads-up. Thoughts?

moonmeister:
I'll get the tests fixed tomorrow. It's late in Perth.

packages/gatsby-plugin-manifest/src/gatsby-node.js
moonmeister: Nevermind...I just re-read it and it's fine without.

moonmeister:
@wardpeet FYI, should be good to go.

wardpeet:
thanks a bunch! merging! Keep up the good work! 💪

PR #12085, opened 2019-02-25
refactor(gatsby-plugin-feed): warn for deprecations, validate options

DSchau:
## Description

gatsby-plugin-feed tends to have slightly confusing behavior.

It's _too_ generalized, and for most non-trivial usages, consumers of the plugin will _want to_ implement custom feeds, queries, and serialize functions.

This PR introduces a few things, specifically:

- Validation of options (I've been bitten before by a missing output)
- Warnings of upcoming deprecations re: _not_ passing a custom feed, as well as not passing a custom serializer, both of which will be required in the next major bump of this plugin

Additionally, it fixes a bug, specifically that the _base_ query was deleted (no idea why?) which leads to an error with multiple feeds:

```shell
error Plugin gatsby-plugin-feed returned an error


  TypeError: Cannot read property 'siteMetadata' of undefined
  
  - internals.js:38 setup
    [rss-feed]/[gatsby-plugin-feed]/internals.js:38:40
```

I propose that we make this a minor version bump for this plugin, once merged.

## Related Issues

We need to wait until #11941 is merged to merge this one (and create the short link!).


packages/gatsby-plugin-feed/README.md
```diff
@@ -8,29 +8,13 @@ Create an RSS feed (or multiple feeds) for your Gatsby site.
 
 ## How to Use
 
-```javascript
```
DSchau: Deleted this, rather than documenting the simple case (which will rarely be used).

packages/gatsby-plugin-feed/README.md
```diff
@@ -86,6 +70,10 @@ plugins: [
 ]
 ```
 
-To see what option keys are valid, see [the itemOptions section](https://www.npmjs.com/package/rss#itemoptions) of the RSS module.
+Each feed must include `output`, `query`, and `title`. Additionally, it is strongly recommended to pass a custom `serialize` function, otherwise an internal serialize function will be used which may not exactly match your particular use case.
```
DSchau: @marcysutton is this documented OK? I think this could maybe be improved, so suggestions more than welcome!

Also probably a good idea to link to your new doc in this README--when it's created/merged!

packages/gatsby-plugin-feed/package.json
```diff
@@ -8,9 +8,8 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.0.0",
+    "fs-extra": "^7.0.1",
```
DSchau: This differs from most other packages (most are on 4.x), but that's probably OK.

packages/gatsby-plugin-feed/README.md
marcysutton: This is great and way more helpful than before!

packages/gatsby-plugin-feed/src/gatsby-node.js
```diff
@@ -1,12 +1,21 @@
...
 const publicPath = `./public`
 
+const warnMessage = (error, behavior) => `
+  gatsby-plugin-feed was initialized in gatsby-config.js without a ${error}.
+  This means that we the plugin will use ${behavior}, which may not match your use case.
+  This behavior will be removed in the next major release of gatsby-plugin-feed.
+  For more info, check out: https://gatsby.app/rss-feed
+`
+
+// TODO: remove in the next major release
```
marcysutton: Should we file an issue for this TODO or does that happen automatically? 🙏

packages/gatsby-plugin-feed/src/plugin-options.js
```diff
@@ -0,0 +1,17 @@
+import Joi from "joi"
```
marcysutton: I never knew Joi was a thing. Cool!

DSchau: It's amazing :) We use it in gatsby core, and it's really nice to warn/error when we don't get the data we expect!

packages/gatsby-plugin-feed/src/gatsby-node.js
DSchau: These are more for internal use--i.e. when we're ready to introduce breaking changes, we'll take that opportunity to refactor all these TODOs and clean up the code!

packages/gatsby-plugin-feed/src/gatsby-node.js
```diff
@@ -19,9 +28,40 @@ const serialize = ({ query: { site, allMarkdownRemark } }) =>
...
+        reporter.stripIndent(
+          warnMessage(
+            `serialize function in a feed`,
+            `the internal serialize function`
+          )
+        )
+      )
+    }
+  } catch (e) {
+    throw new Error(e.details.map(detail => detail.message).join(`\n`))
```
pieh: We should mention that this error comes from config validation. I think gatsby will say it comes from `gatsby-plugin-feed`, but figuring out might be confusing without more context

DSchau: Just prefix the error? e.g. `[Config Validation]: ${detail.message}` or something?

secretfader: I imagine you'll want a format that can be (eventually) adopted by other plugins. For now, I agree it makes sense to prefix with `gatsby-plugin-feed`.

pieh: Sounds good, just something that user isn't greeted with `Error: "query" is required` message and not sure where to go from there

DSchau:
@secretfader thanks! We also added a test (and ensured behavior is good to go!) on an older issue that @pieh dug up.

We now ensure that we don't mutate the base query.

Thanks for your assistance in helping out with this plugin--if there's anything we can do to help out or get you back into using and contributing to Gatsby, please do let us know!

secretfader:
Thanks for your work on this, @DSchau. This plugin, unlike some others, hasn't received much love since the initial release (mandatory API upgrades excepted). It seems to be rising in popularity, too.

I'll likely message you off-thread about opportunities where my skills might intersect with Gatsby once again, but in the meantime I just wanted to give a big :+1:. This is very good work, and I hope it is laying a foundation for future improvements to `gatsby-plugin-feed`.

PR #12087, opened 2019-02-25
docs: add API reference for params passed to gatsby-node APIs

### LINKED ISSUE #4120: document helper functions passed to gatsby-node APIs
## Description

I have a collection of markdown files that I'd like to become blog posts. In the course of trying to figure out how to filter by folder, I came across `getNode()` in various code samples, but I can't find anything about it the docs. ~~It seems to be a part of the `boundActionCreators` inside `onCreateNode`, but I'm not 100% sure of that.~~

Can we add it to the docs? I'm happy to have a bash if somebody can point me the right direction.

EDIT: I realise that `getNode` looks like a sibling of `boundActionCreators` rather than a child of it.

pieh:
This includes changes from https://github.com/gatsbyjs/gatsby/pull/11597 which is needed for nested and also some really unrelated changes (to make preview work on netlify). Actual changes for review are limited to https://github.com/pieh/gatsby/compare/58c6855...pieh:www-doc-preview

.org preview available at https://www-doc-preview--tender-curie-facda8.netlify.com/docs/node-api-helpers/

TO-DO before the merge:
 - [x] change new path to `/docs/node-api-helpers` (from `/docs/node-apis-helpers`)
 - [x] remove temporary commit that allow to build previews on netlify ( https://github.com/gatsbyjs/gatsby/pull/12087/commits/58c6855ab458ee05a3557cec0ac871591f9dbc35 )
 - [x] merge #11597 and remove commits from that PR from this branch

Closes #4120


.prettierrc
```diff
@@ -3,5 +3,13 @@
   "semi": false,
   "singleQuote": false,
   "tabWidth": 2,
-  "trailingComma": "es5"
+  "trailingComma": "es5",
+  "overrides": [
+    {
+      "files": ["**/api-node-helpers-docs.js"],
+      "options": {
+        "semi": true
```
pieh: This is lame - but `documentation` package used to read jsdocs is just working weirdly without semicolons (for more complex cases like this at least)

www/src/pages/docs/node-apis-helpers.js
```diff
@@ -0,0 +1,63 @@
...
+  render() {
+    const docs = sortBy(
+      this.props.data.allDocumentationJs.edges.map(({ node }) => node),
+      docs => docs.name
+    )
+    return (
+      <Layout location={this.props.location} itemList={itemListDocs}>
+        <Container>
+          <Helmet>
+            <title>Node APIs helpers</title>
```
pieh: I am calling this `helpers` - but those are not really `helpers` - I'm not sure what to call them.

KyleAMathews: This sounds fine. To bike shed for a bit, "Utilities for Node APIs" would work too.

KyleAMathews:
Page looks great! Excited to get this out there.

www/src/components/api-reference/doc-block.js
```diff
@@ -0,0 +1,135 @@
...
+`
+
+const Deprecated = ({ definition }) => {
+  if (definition.deprecated && definition.deprecated.childMarkdownRemark) {
+    return (
+      <div
+        css={`
+          p:before {
+            color: #e8bd36;
+            content: "(depracated) ";
```
m-allanson:
```suggestion
            content: "(deprecated) ";
```

www/src/pages/docs/node-apis-helpers.js
```diff
@@ -0,0 +1,63 @@
...
+          </Helmet>
+          <h1 id="gatsby-node-helpers-apis" css={{ marginTop: 0 }}>
+            Gatsby Node APIs helpers
+          </h1>
+          <p>
+            First argument passed to{` `}
```
m-allanson:
```suggestion
            The first argument passed to each of{` `}
```

www/src/pages/docs/node-apis-helpers.js
```diff
@@ -0,0 +1,63 @@
...
+            <Link to="/docs/node-apis/">Gatsby Node APIs</Link> is object
```
m-allanson:
```suggestion
            <Link to="/docs/node-apis/">Gatsby's Node APIs</Link> is an object
```

www/src/pages/docs/node-apis-helpers.js
```diff
@@ -0,0 +1,63 @@
...
+            containing set of utilities.
```
m-allanson:
```suggestion
            containing a set of utilities.
```

packages/gatsby/src/utils/api-node-helpers-docs.js
```diff
@@ -0,0 +1,293 @@
...
+   * Set of utilities to output information to user
+   * @type {GatsbyReporter}
+   */
+  reporter: true,
+
+  /**
+   * Get single node by given ID and creates dependency for given path.
+   * This should be used instead of `getNode` in graphql resolvers to enable
+   * tracking dependencies for query results. If it's not used Gatsby will
+   * not rerun query if node. See
```
m-allanson: missing end of sentence? _if node `<something>`._

.prettierrc
m-allanson: Is it possible to add comments this file? This will look quite mysterious to the next person that finds it.

packages/gatsby/src/utils/api-node-helpers-docs.js
```diff
@@ -0,0 +1,293 @@
...
+  parentSpan: true,
+
+  /**
+   * @callback GatsbyTracing.StartSpan
+   * @param {string} spanName name of the span
+   * @returns {Opentracing.Span}
+   */
+
+  /**
+   * Convience function to start span using API run span as parent.
```
m-allanson:
```suggestion
   * Convenience function to start a span using API run span as parent.
```

Fixed a typo, but I don't understand the sentence. Maybe we could re-phrase it?

packages/gatsby/src/utils/api-node-helpers-docs.js
```diff
@@ -0,0 +1,293 @@
...
+   *   span.finish()
+   * }
+   */
+  startSpan: true,
+};
+
+/** */
+const GatsbyNodeHelpers = {
+  /**
+   * Key-value store used to persist results of time/memory/cpu instensive
```
m-allanson:
```suggestion
   * Key-value store used to persist results of time/memory/cpu intensive
```

packages/gatsby/src/utils/api-node-helpers-docs.js
```diff
@@ -0,0 +1,293 @@
...
+  /**
+   * Get cache instance by name - this should only be used by plugins that
+   * accept subplugins.
+   * @param {string} id Test
+   * @returns {GatsbyCache} See [`cache`](#cache) section for reference.
+   */
+  getCache: true,
+
+  /**
+   * Create stable content digest from string or object. Useful to set
```
m-allanson: How about something like this?

_Create a stable content digest from a string or object, you can use this to set the `internal.contentDigest` field on nodes. Gatsby uses the value of this field to invalidate stale data when your content changes._

packages/gatsby/src/utils/api-node-helpers-docs.js
```diff
@@ -0,0 +1,293 @@
...
+  /**
+   * All action creators wrapped with a dispatch.
+   *
+   * See [`actions`](/docs/actions/) reference.
+   * @type {Actions}
+   */
+  actions: true,
+
+  /**
+   * @param {Node} node
```
m-allanson: Missing description? We can add this in a follow up PR.

packages/gatsby/src/utils/api-node-helpers-docs.js
```diff
@@ -0,0 +1,293 @@
...
+   *     const textContent = await loadNodeContent(node)
+   *     // process textContent and create child nodes
+   *   }
+   * }
+   */
+  loadNodeContent: true,
+
+  /**
+   * Internal redux state used for application state. Do not use, unless you
+   * absolutely must. Store is considered private API and can change with
```
m-allanson:
```suggestion
   * absolutely must. Store is considered a private API and can change with
```

packages/gatsby/src/utils/api-node-helpers-docs.js
```diff
@@ -0,0 +1,293 @@
...
+   * any version.
+   * @type {Redux.Store}
+   */
+  store: true,
+
+  /**
+   * Internal event emitter / listener.  Do not use, unless you absolutely
+   * must. Emitter is considered private API and can change with any version.
```
m-allanson:
```suggestion
   * must. Emitter is considered a private API and can change with any version.
```

packages/gatsby/src/utils/api-node-helpers-docs.js
```diff
@@ -0,0 +1,293 @@
...
+   * @returns {Node[]} Array of nodes.
+   * @example
+   * const markdownNodes = getNodesByType(`MarkdownRemark`)
+   */
+  getNodesByType: true,
+
+  /**
+   * Compares `contentDigest` of cached node with passed value
+   * to determine if node has changed.
+   * if node changed.
```
m-allanson:
```suggestion
   *
```

packages/gatsby/src/utils/api-node-helpers-docs.js
```diff
@@ -0,0 +1,293 @@
...
+   * for more details.
+   * @param {string} ID id of the node.
+   * @param {string} path of the node.
+   * @returns {Node} Single node instance.
+   */
+  getNodeAndSavePathDependency: true,
+
+  /**
+   * Utility function useful to generate globally unique and stable node IDs.
+   * It will generate different IDs for different plugin if they use same
```
m-allanson:
```suggestion
   * It will generate different IDs for different plugins if they use same
```

www/gatsby-node.js
```diff
@@ -213,24 +213,24 @@ exports.createPages = ({ graphql, actions, reporter }) => {
 
   return new Promise((resolve, reject) => {
     const docsTemplate = path.resolve(`src/templates/template-docs-markdown.js`)
-    const blogPostTemplate = path.resolve(`src/templates/template-blog-post.js`)
-    const blogListTemplate = path.resolve(`src/templates/template-blog-list.js`)
-    const tagTemplate = path.resolve(`src/templates/tags.js`)
-    const contributorPageTemplate = path.resolve(
-      `src/templates/template-contributor-page.js`
-    )
+    // const blogPostTemplate = path.resolve(`src/templates/template-blog-post.js`)
```
m-allanson: Leftover debugging/wip comments?

www/src/data/sidebars/doc-links.yaml
```diff
@@ -289,6 +289,8 @@
       link: /docs/gatsby-config/
     - title: Gatsby Node APIs
       link: /docs/node-apis/
+    - title: Gatsby Node Helpers
+      link: /docs/node-apis-helpers/
```
m-allanson:
```suggestion
      link: /docs/node-api-helpers/
```

I think _apis_ can be singular because _helpers_ is plural.

www/src/pages/docs/node-apis-helpers.js
m-allanson: I agree Node API helpers is fine here.

```suggestion
            <title>Node API helpers</title>
```

.prettierrc
pieh: Yes, the normal non-jsdoc comments shouldn't affect jsdoc (but given this `semi` problem that I spent few hours trying to figure out why things don't work as expected I won't guarantee this before trying

m-allanson: Oops sorry - I meant adding a comment to this file to say "`documentation` package used to read jsdocs is just working weirdly without semicolons (for more complex cases like this at least), see https://github.com/gatsbyjs/gatsby/pull/12087"

www/src/components/api-reference/doc-block.js
fk: 👍

www/src/data/sidebars/doc-links.yaml
pieh: I will change this last, as I need to also change page path - I will add this to TO-DO items in PR description

www/gatsby-node.js
pieh: This is part of temporary commit that limits amount of pages, so I can build preview for this PR on netlify ( https://github.com/gatsbyjs/gatsby/pull/12087/commits/58c6855ab458ee05a3557cec0ac871591f9dbc35 ). This commit will be removed before merge

www/src/data/sidebars/doc-links.yaml
pieh: Another question - should this go below `/docs/node-apis/`? Both in sidebar hierarchy and in actual url - `/docs/node-apis/helpers`?

.prettierrc
pieh: oh - yeah, we can should be able to switch `.prettierrc` to `.prettierrc.js`, and then we will be able to add inline comments

packages/gatsby/src/utils/api-node-helpers-docs.js
pieh: I remember now why I skipped this - we have this logic:
https://github.com/gatsbyjs/gatsby/blob/f7a6bb5d9615e3952827a8736039eba9aa372e28/packages/gatsby/src/db/nodes.js#L22-L50

But I didn't find documentation for implementing custom `loadNodeContent` function in plugins (not using it, but providing it in case `internal.content` isn't set). Example of this is in `gatsby-source-filesystem`
https://github.com/gatsbyjs/gatsby/blob/f7a6bb5d9615e3952827a8736039eba9aa372e28/packages/gatsby-source-filesystem/src/index.js#L3-L10

This allows to lazy load content (so we don't read file content from f.e. images, because no transformer actually need that as a string)

I will just use `Get content for a node from the plugin that created it.` so it at least have some description, but this needs to be documented further (probably separate PR)

pieh: Added very basic description in https://github.com/gatsbyjs/gatsby/pull/12087/commits/5d351ea79d9011339689427c6d3f2293d109f02f - need more work, but don't want to block this PR

packages/gatsby/src/utils/api-node-helpers-docs.js
pieh: Reworded in https://github.com/gatsbyjs/gatsby/pull/12087/commits/b4f18e0730dd92a144a8a158c64513d00d1dcda8

packages/gatsby/src/utils/api-node-helpers-docs.js
pieh: Finished sentence in https://github.com/gatsbyjs/gatsby/pull/12087/commits/c13fdef4d8625f7596ed133d9057e1b52a926ad4

.prettierrc
pieh: Added comment in https://github.com/gatsbyjs/gatsby/pull/12087/commits/ee7ede041dde7230114e15fcdd04adb2e3542a76

www/src/data/sidebars/doc-links.yaml
m-allanson: It's probably ok to keep the structure flat here? Maybe we can change it later if we want to look at restructuring the whole of the API Reference section of the docs.

packages/gatsby/src/utils/api-node-helpers-docs.js
m-allanson: Thanks for the example, that makes sense. Hmmmm this is a tricky one - how about:

Start a tracing span. The span will be created as a child of the currently running API span.
This is a convenience wrapper for `` tracing.tracer.startSpan(`span name`, { childOf: tracing.parentSpan}) ``.

www/src/pages/docs/node-apis-helpers.js
```diff
@@ -0,0 +1,63 @@
...
+      docs => docs.name
+    )
+    return (
+      <Layout location={this.props.location} itemList={itemListDocs}>
+        <Container>
+          <Helmet>
+            <title>Node API helpers</title>
+          </Helmet>
+          <h1 id="gatsby-node-helpers-apis" css={{ marginTop: 0 }}>
+            Gatsby Node APIs helpers
```
KyleAMathews: This should be `Gatsby Node API helpers`

packages/gatsby/src/utils/api-node-helpers-docs.js
```diff
@@ -0,0 +1,292 @@
...
+   *   internal: {
+   *     type: `TypeOfNode`,
+   *     contentDigest: createContentDigest(nodeData)
+   *   }
+   * }
+   */
+  createContentDigest: true,
+
+  /**
+   * All action creators wrapped with a dispatch.
```
KyleAMathews: We can probably leave off the "wrapped with a dispatch" bit since generally speaking, we're obscuring that Redux is used under the hood so the "dispatch" bit is an implementation detail and not relevant.

packages/gatsby/src/utils/api-node-helpers-docs.js
```diff
@@ -0,0 +1,292 @@
...
+   *
+   * See [`actions`](/docs/actions/) reference.
+   * @type {Actions}
+   * @deprecated Will be removed in gatsby 3.0. Use [actions](#actions)
+   * instead.
+   */
+  boundActionCreators: true,
+
+  /**
+   * All action creators wrapped with a dispatch.
```
KyleAMathews: same

packages/gatsby/src/utils/api-node-helpers-docs.js
```diff
@@ -0,0 +1,292 @@
...
+  /**
+   * Get cache instance by name - this should only be used by plugins that
+   * accept subplugins.
+   * @param {string} id Test
+   * @returns {GatsbyCache} See [`cache`](#cache) section for reference.
+   */
+  getCache: true,
+
+  /**
+   * Create a stable content digest from a string or object, you can use this
```
KyleAMathews: "you can use the result of this function"

packages/gatsby-transformer-documentationjs/src/gatsby-node.js
```diff
@@ -80,76 +60,166 @@ exports.onCreateNode = async ({
...
-
-        // When documenting destructured parameters, the name
-        // is parent.child where we just want the child.
-        if (param.name.split(`.`).length > 1) {
-          param.name = param.name
-            .split(`.`)
-            .slice(-1)
-            .join(`.`)
-        }
+      const index = documentationJson.findIndex(
```
wardpeet: You're only using the index to use it as `documentationJson[index]` so might be better to use find instead. Unless i've missed something
```suggestion
      const comment = documentationJson.find(
```

packages/gatsby-transformer-documentationjs/src/gatsby-node.js
```diff
@@ -80,76 +60,166 @@ exports.onCreateNode = async ({
...
+      }
 
-          return ret
-        })
+      if (type.type === `NameExpression` && type.name) {
+        type.typeDef___NODE = getNodeIDForType(type.name)
+      }
+    }
+
+    const prepareNodeForDocs = (
```
wardpeet: maybe add a few comments to this function to tell you what you're doing as from reading the code I can't but I have no knowledge of parsing documentation 😄

packages/gatsby/cache-dir/api-ssr-docs.js
```diff
@@ -99,17 +99,17 @@ exports.onRenderBody = true
  * replace head components to be rendered in your `html.js`. This is useful if
  * you need to reorder scripts or styles added by other plugins.
  * @param {Object} $0
- * @param {Array} $0.getHeadComponents Returns the current `headComponents` array.
+ * @param {Array<ReactNode>} $0.getHeadComponents Returns the current `headComponents` array.
```
wardpeet: does ReactNode work? It's not defined anywhere but maybe it's just to give a better understanding to people read it.

on line https://github.com/gatsbyjs/gatsby/pull/12087/files#diff-2ad21b5b5ed90fd9c676b17f828cbea5R16 we were using `@returns {string[]}` there is no difference in parsing but just not sure what syntax we like

scripts/publish-site.sh
```diff
@@ -12,7 +13,7 @@ cd "$1" || exit
 yarn
 
 echo "=== Copying built Gatsby to website."
-gatsby-dev --scan-once
+gatsby-dev --scan-once --packages gatsby gatsby-transformer-documentationjs
```
wardpeet: specific reason to scope to these packages? Might be handy to add a comment

www/package.json
```diff
@@ -90,6 +90,9 @@
   "license": "MIT",
   "main": "n/a",
   "private": true,
+  "resolutions": {
+    "documentation": "^9.0.0"
```
wardpeet: don't forget to remove this one either :smile:

www/src/components/api-reference/doc-block.js
```diff
@@ -0,0 +1,134 @@
...
+  SignatureWrapper,
+  TypeComponent,
+  SignatureElement,
+  isFunctionDef,
+} from "./signature"
+import ReturnBlock from "./returns"
+import { Header } from "./utils"
+import { options, scale } from "../../utils/typography"
+
+const Optional = styled.span`
```
wardpeet: unsure if screen readers read the content field. ping @marcysutton

.prettierrc.js
```diff
@@ -0,0 +1,19 @@
+module.exports = {
+  endOfLine: "lf",
+  semi: false,
+  singleQuote: false,
```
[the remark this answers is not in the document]
jlengstorf: If these aren't explicitly set, the defaults of some IDEs (like VS Code) will be used, so it's probably worth keeping all Prettier options explicitly set.

packages/gatsby-transformer-documentationjs/src/gatsby-node.js
```diff
@@ -17,49 +11,35 @@ const stringifyMarkdownAST = (node = ``) => {
...
-  node,
-  docNodeId,
-  markdownStr,
-  name,
-  actions,
-  createNodeId
-) {
-  const { createNode } = actions
+function prepareDescriptionNode(node, markdownStr, name, helpers) {
+  const { createNodeId, createContentDigest } = helpers
```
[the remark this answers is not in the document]
jlengstorf: With a more complex function signature like this, I kind of like the extra bit of documentation this provides: "ah, okay, I get helpers, and I'm going to use these two helpers."

Not sure it matters in the long run, but I think that extra context is worth the extra line.

scripts/publish-site.sh
pieh: I will remove this change before actual merge (see TO-DO list in PR description) - this is part of temporary commit that allow me to actually build preview on netlify (this + removing bunch of pages in www/gatsby-node.js)

.prettierrc.js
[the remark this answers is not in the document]
pieh: To all the prettierrc changes - I just changed from `json` format that is used currently ( https://github.com/gatsbyjs/gatsby/blob/master/.prettierrc ) to `.js`, so I would be able to add comment before override, why this override is there - there are no config changes other than this override. If there are changes that can be done to config - it should be in separate PR

packages/gatsby-transformer-documentationjs/src/gatsby-node.js
```diff
@@ -17,49 +11,35 @@ const stringifyMarkdownAST = (node = ``) => {
...
-function createDescriptionNode(
-  node,
-  docNodeId,
-  markdownStr,
-  name,
-  actions,
-  createNodeId
-) {
-  const { createNode } = actions
+function prepareDescriptionNode(node, markdownStr, name, helpers) {
```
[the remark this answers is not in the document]
pieh: Because I don't actually use `createNode` here, the name change is to signify that `createNode` is not called here without having to look into function itself. It's something I picked up from `gatsby-source-contentful` and I think that this naming convention does make a lot of sense

packages/gatsby-transformer-documentationjs/src/gatsby-node.js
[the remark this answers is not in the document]
pieh: I think it doesn't really matter that much. I can switch to destructuring in arg list if you feel strongly about it ;)

packages/gatsby-transformer-documentationjs/src/gatsby-node.js
```diff
@@ -17,49 +11,35 @@ const stringifyMarkdownAST = (node = ``) => {
...
  * @param {Object} super this is a super param
  */
-exports.onCreateNode = async ({
-  node,
-  loadNodeContent,
-  actions,
-  createNodeId,
-}) => {
+exports.onCreateNode = async ({ node, actions, ...helpers }) => {
+  const { createNodeId, createContentDigest } = helpers
```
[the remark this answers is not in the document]
pieh: here it makes more actual sense, because I can easily pass `helpers` later to `prepareDescriptionNode` without reconstructing `helpers` object. Fact that I grab `...helpers` just future-proofs the code a bit, that if `prepareDescriptionNode` need access to another helper - it already has it without the need to adjust calls.

packages/gatsby/cache-dir/api-ssr-docs.js
pieh: Those are really just to create API reference at this point. Only real thing that this does is actually display `ReactNode` in https://www-doc-preview--tender-curie-facda8.netlify.com/docs/ssr-apis/#onPreRenderHTML

If this would be used for providing typings than yeah, we probably would need to look into actually importing React types somewhere to make it actually work.

pieh: As for `Array<ReactNode>` or `ReactNode[]` - I don't really have preference and can switch no problem

www/src/components/api-reference/doc-block.js
pieh: That's good point. It probably makes sense to adjust. I'm just not sure what semantic structure actually makes sense to use here.

packages/gatsby-transformer-documentationjs/src/gatsby-node.js
pieh: I also use it as `commentNumber` field (passing it in line 79) - this is added to node (for backward compatibility)

m-allanson:
Is this basically ready to go? @pieh do you want to handle the final TODO list you mentioned in the description and get this merged in?

I think there's huge value in in getting these docs live on .org, then we can follow up with any smaller changes as needed.

packages/gatsby-transformer-documentationjs/src/gatsby-node.js
pieh: Added in https://github.com/gatsbyjs/gatsby/pull/11597/commits/2c055f85ca11ede26095a6510605edab956df20b (there is separate PR for documentation plugin). Commits here are only so netlify preview use those changes

packages/gatsby-transformer-documentationjs/src/gatsby-node.js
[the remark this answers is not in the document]
wardpeet: I don't 😄  just asking 😛

packages/gatsby/cache-dir/api-ssr-docs.js
wardpeet: I like the Array syntax but who am i 😜

.prettierrc.js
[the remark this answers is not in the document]
wardpeet: Ah that's why I didn't see it. Should have looked a bit better. Thanks for clarifying!

www/src/components/api-reference/doc-block.js
marcysutton: Yes they do read it! It becomes part of the computed text for an element node.

www/src/pages/docs/node-api-helpers.js
```diff
@@ -0,0 +1,63 @@
...
+        <Container>
+          <Helmet>
+            <title>Node API helpers</title>
+          </Helmet>
+          <h1 id="gatsby-node-helpers-apis" css={{ marginTop: 0 }}>
+            Gatsby Node API helpers
+          </h1>
+          <p>
+            The first argument passed to each of{` `}
+            <Link to="/docs/node-apis/">Gatsby's Node APIs</Link> is an object
```
marcysutton: It would help to see a code example here to provide more context. How does the "object containing a set of utilities" relate to this list of helpers? Are the helpers passed into Gatsby Node APIs?

pieh: Yes, the first argument passed to Gatsby Node APIs is object containing those "helpers", I will try to clarify that.

I also started to wonder if this should be called `context` instead of `helpers` (might be less confusing?).

marcysutton: Is it the same context that we see elsewhere? If so, then it would be good to be consistent. Otherwise, the naming is less important than showing it in use.

pieh: There plenty of different contexts, and we never used `context` to describe first param in gatsby-node APis :( pages have their context, graphql have their context, so I'll just drop this idea

PR #12096, opened 2019-02-26
docs(gatsby-source-wordpress): add ACF option pages troubleshooting

HurricaneInteractive:
## Description

Added some troubleshooting notes for ACF Option pages after running into issues relating to data not updating or not showing up during setup.


pieh:
Thanks @HurricaneInteractive!

PR #12097, opened 2019-02-26
[schema] Make output fields non-null

stefanprobst:
This is for #11480

### DIFF
--- a/packages/gatsby/src/schema/types/node-interface.js
+++ b/packages/gatsby/src/schema/types/node-interface.js
@@ -1,15 +1,16 @@
 const getOrCreateNodeInterface = schemaComposer => {
   // TODO: why is `mediaType` on Internal? Applies only to File!?
   // `fieldOwners` is an object
+  // Should we drop ignoreType?
   const InternalTC = schemaComposer.getOrCreateTC(`Internal`, tc => {
     tc.addFields({
       content: `String`,
-      contentDigest: `String`,
+      contentDigest: `String!`,
       description: `String`,
       fieldOwners: [`String`],
       ignoreType: `Boolean`,
       mediaType: `String`,
-      owner: `String`,
+      owner: `String!`,
       type: `String!`,
     })
   })
--- a/packages/gatsby/src/schema/types/node-interface.js
+++ b/packages/gatsby/src/schema/types/node-interface.js
@@ -28,7 +28,7 @@ const getOrCreateNodeInterface = schemaComposer => {
         },
       },
       children: {
-        type: `[Node]!`,
+        type: `[Node!]!`,
         resolve: (source, args, context, info) => {
           const { path } = context
           return context.nodeModel.getNodesByIds(
--- a/packages/gatsby/src/schema/types/node-interface.js
+++ b/packages/gatsby/src/schema/types/node-interface.js
@@ -37,7 +37,7 @@ const getOrCreateNodeInterface = schemaComposer => {
           )
         },
       },
-      internal: `Internal`,
+      internal: `Internal!`,
     })
   })
   NodeInterfaceTC.getITC()
--- a/packages/gatsby/src/schema/types/pagination.js
+++ b/packages/gatsby/src/schema/types/pagination.js
@@ -63,14 +63,14 @@ const getPagination = ({ schemaComposer, typeComposer }) => {
   })
   const fields = {
     distinct: {
-      type: [`String`],
+      type: [`String!`],
       args: {
         field: FieldsEnumTC.getTypeNonNull(),
       },
       resolve: distinct,
     },
     group: {
-      type: [getGroup({ schemaComposer, typeComposer })],
+      type: [getGroup({ schemaComposer, typeComposer }).getTypeNonNull()],
       args: {
         skip: `Int`,
         limit: `Int`,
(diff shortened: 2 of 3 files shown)


packages/gatsby/src/schema/types/pagination.js
```diff
@@ -18,26 +18,34 @@ const getEdge = ({ schemaComposer, typeComposer }) => {
...
+  fields,
+  typeName,
+}) => {
+  const paginationTypeComposer = schemaComposer.getOrCreateTC(typeName, tc => {
     tc.addFields({
-      totalCount: `Int`,
-      edges: [getEdge({ schemaComposer, typeComposer })],
-      pageInfo: getPageInfo({ schemaComposer }),
+      totalCount: `Int!`,
+      edges: [getEdge({ schemaComposer, typeComposer }).getTypeNonNull()],
```
freiksenet: Array itself should also be non-null.

freiksenet: While we are at it, could you add `nodes` field as a shortcut to `edges { node }`.

packages/gatsby/src/schema/types/pagination.js
```diff
@@ -69,7 +77,15 @@ const getPagination = ({ schemaComposer, typeComposer }) => {
...
   }
-  return createPagination({ schemaComposer, typeComposer, fields, typeName })
+  const paginationTypeComposer = createPagination({
+    schemaComposer,
+    typeComposer,
+    fields,
+    typeName,
+  })
+  paginationTypeComposer.makeFieldNonNull(`distinct`)
+  paginationTypeComposer.makeFieldNonNull(`group`)
```
freiksenet: Does it apply it to Array too?

packages/gatsby/src/schema/types/pagination.js
stefanprobst: `edges` array is made nonnull further down with `makeFieldNonNull(`edges`)`.

stefanprobst: shortcut field: 100% yes! should be a new PR?

packages/gatsby/src/schema/types/pagination.js
stefanprobst: Yes. This is what we have now:
```graphql
type FileConnection {
  totalCount: Int!
  edges: [FileEdge!]!
  pageInfo: PageInfo!
  distinct(field: FileFieldsEnum!): [String]!
  group(skip: Int, limit: Int, field: FileFieldsEnum!): [FileGroupConnection]!
}

type FileEdge {
  next: File
  node: File!
  previous: File
}

type FileGroupConnection {
  totalCount: Int!
  edges: [FileEdge!]!
  pageInfo: PageInfo!
  field: String!
  fieldValue: String
}
```

stefanprobst: One thing I missed though is making `internal` nonnull on Node interface

packages/gatsby/src/schema/types/pagination.js
freiksenet: Can be the same.

packages/gatsby/src/schema/types/pagination.js
freiksenet: `distinct` should be `[String!]!`, same for `[FileGroupConnection!]!`

stefanprobst: Hm should they? I did this on purpose because they can be empty arrays, for example with:
```graphql
{
  allFile {
    distinct(field:parent___id)
    group(field:parent___id) {
      edges {
        node {
          id
        }
      }
    }
  }
}
```

stefanprobst: I guess you're right though, this *should* show `null` as a `distinct` value.

stefanprobst: So: `distinct` should be `[String]!` (can include null), and `group` should be  `[FileGroupConnection!]!` (with fieldValue being nullable), right?

freiksenet: Yes :+1:

packages/gatsby/src/schema/types/node-interface.js
```diff
@@ -36,7 +37,7 @@ const getOrCreateNodeInterface = schemaComposer => {
           )
         },
       },
-      internal: `Internal`,
+      internal: `Internal!`,
```
freiksenet: Children should also be [Node!]!. We should return an error if one of the children nodes are null.

packages/gatsby/src/schema/types/pagination.js
stefanprobst: Hm, sorry, this still not 100% clear to me:

* `distinct` is clear: we mirror the behavior of master which filters out `null` values, so this can be an empty array -> therefore `[String]!`
* `group` is weird though, because it uses `lodash`'s `groupBy`, which apparently treats `null` and `"null"` (string) the same.
`_.groupBy([{ field: null }, { field: "null" }], 'field')`
will return
`{ null: [ { field: null }, { field: 'null' } ] }`

stefanprobst:
@freiksenet Sorry for the noise, this is good to go now!

Issue #12003, opened 2019-02-22
[www/showcase] Filter and scroll issues


fk:
Clicking any "Category" label on a "Showcase site card" does update the URL, but does not actually set and filter by the clicked category:

* open https://www.gatsbyjs.org/showcase
* scroll down past "Featured Sites" to the list of all Showcase sites (418)
* click "Featured" on the reactjs.org card <img src="https://user-images.githubusercontent.com/21834/53264833-ea1d3b00-36dc-11e9-9b70-07dd3714f52e.png" width="300px" />
* URL updates to https://www.gatsbyjs.org/showcase?filters%5B0%5D=Featured, but
  * category is not set as filter in sidebar
  * list is unfiltered, still showing all 418 Showcase sites

---

Another fun 🙉🙈one:

* click the "View all" link next to "Featured Sites" 
![image](https://user-images.githubusercontent.com/21834/53265041-792a5300-36dd-11e9-9f3d-1586d9d65c14.png)
* watch the viewport being scrolled down to `#showcase`, the URL being updated to https://www.gatsbyjs.org/showcase?filters%5B0%5D=Featured, and then …
* the scroll position is being reset to `0`
* filters are not set (just like described above)

rohintangirala:
Just to clarify, the first issue applies to all the filter categories, not just Featured.

freiksenet:
Thank you for the report! Would you like to make a PR to fix this issue?

rohintangirala:
I'll check it out.

wardpeet:
@rohintangirala I'm actually fixing this one already #11001

wardpeet:
@fk i'm closing this one in favour of #11001 so we don't have 2 people working on the same issue.

rohintangirala:
@fk Alright sounds good. I didn't really get to it anyway.

wardpeet:
Sorry about this. If I knew earlier I would have gave you a few pointers instead

Issue #11001, opened 2019-01-11
[site-showcase] category filters shouldn't jump up


wardpeet:
## Description
It's part of #9048 where we want to scroll down when people visit the page with ?filters in their url.

Sadly we scroll up on each time we select a new category.

### Steps to reproduce

1) go to https://www.gatsbyjs.org/showcase/
2) click on a category filter
3) site jumps back up

### Expected result

It should not scroll up, maybe it should scroll back to the top of the showcase but not to the top of the page
![image](https://user-images.githubusercontent.com/1120926/51036465-81a35000-15ad-11e9-9ceb-8f825a0bf897.png)

### Actual result

![scroll showcase](https://user-images.githubusercontent.com/1120926/51036383-38530080-15ad-11e9-8cdc-39b1ffdd875b.gif)

marcysutton:
I wonder if that's focus management happening with Reach Router, where focus is being set to the top of the page when the view changes? Might be worth looking into. Perhaps focus could go to the lower part of the view if that makes more sense from an interaction standpoint ("N Sites (filtered)" in your animated GIF").

PR #12105, opened 2019-02-26
fix(www): fix scroll when filtering showcase

wardpeet:
## Description
Fixes scrolling when clicking on filters in the showcase. All filters do exactly the same now. I moved state up.

## Related Issues
#11001 
#12003

### DIFF
--- a/www/src/utils/scroll-to-anchor.js
+++ b/www/src/utils/scroll-to-anchor.js
@@ -65,7 +65,9 @@ const scrollToLocation = (element, targetPos, duration, callback) =>
   })
 
 const scrollToAnchor = (target, callback) => event => {
-  event.preventDefault()
+  if (event) {
+    event.preventDefault()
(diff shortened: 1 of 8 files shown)


wardpeet:
So I use replaceState instead of pushState for now but we probably want to fix this as well.

I can probably disable scrolling with `shouldUpdateScroll` api or use native @reach/router's navigate unsure which one is the best for this one.

I did a lot of cleanup as it was hard to follow how everything worked and not everything worked.. 😋

t2ca:
It's working great! Just wondering if it's possible to have the url updated with the categories selected?

t2ca:
Im not sure if its just me, but after i click "Reset all filters", i can't scroll down.

wardpeet:
good catch thank you! I'll fix that right away, it seems to be in an infinite loop.

Issue #12102, opened 2019-02-26
gatsby-plugin-typography - Error whilst following Gatsby tutorial.


JABedford:
## Description

Trying to get up and running with the Typography plugin as per the gatsby tutorial. An error is thrown when I come to run develop.

### Steps to reproduce

`npm install --save gatsby-plugin-typography react-typography typography`

I update my gatsby-config file as per the tutorial.

`gatsby develop`

### Logs

  ```
System:
    OS: macOS 10.14.2
    CPU: (8) x64 Intel(R) Core(TM) i5-8259U CPU @ 2.30GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 8.15.0 - ~/.nvm/versions/node/v8.15.0/bin/node
    npm: 6.4.1 - ~/.nvm/versions/node/v8.15.0/bin/npm
  Languages:
    Python: 2.7.10 - /usr/bin/python
  Browsers:
    Chrome: 72.0.3626.119
    Firefox: 65.0
    Safari: 12.0.2
  npmPackages:
    gatsby: ^2.1.18 => 2.1.18 
    gatsby-plugin-typography: ^2.2.7 => 2.2.7 
  npmGlobalPackages:
    gatsby-cli: 2.4.11
```

```
 Error: ./node_modules/gatsby-plugin-typography/gatsby-ssr.js
  Module not found: Error: Can't resolve './.cache/typography' in '/Users/jbedford/Documents/GitHub/jabedford/blog/node_modul  es/gatsby-plugin-typography'
  resolve './.cache/typography' in '/Users/jbedford/Documents/GitHub/jabedford/blog/node_modules/gatsby-plugin-typography'
    using description file: /Users/jbedford/Documents/GitHub/jabedford/blog/node_modules/gatsby-plugin-typography/package.jso  n (relative path: .)
      using description file: /Users/jbedford/Documents/GitHub/jabedford/blog/node_modules/gatsby-plugin-typography/package.j  son (relative path: ./.cache/typography)
        no extension
          /Users/jbedford/Documents/GitHub/jabedford/blog/node_modules/gatsby-plugin-typography/.cache/typography doesn't exist

```

freiksenet:
Thank you for the report! You need to create a Javascript file for the typography config (default is `src/utils/typography.js`). Documentation should be improved so that it tells one to do it and provide a minimal config example. Do you want to make a PR to fix the docs?

yogeshkotadiya:
[...]

PR #12106, opened 2019-02-26
docs(typography): add some references about typography to tutorial

JABedford:
## Description

Makes the process of creating a `typography.js` file more explicit in the tutorial. Removal of the code tags in the link below to make the link behave/appear better.

## Related Issues

Fixes my own issue of [gatsbyjs#12102](https://github.com/gatsbyjs/gatsby/issues/12102)

### DIFF
--- a/docs/tutorial/part-three/index.md
+++ b/docs/tutorial/part-three/index.md
@@ -50,6 +50,8 @@ npm install --save gatsby-plugin-typography react-typography typography
 
 > Note: Typography.js requires a few additional packages, so those are included in the instructions. Additional requirements like this will be listed in the "install" instructions of each plugin.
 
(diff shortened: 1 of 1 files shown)


KyleAMathews:
Hmm previously we had the actual typography.js file people should use. We don't want people going through the tutorial to need to visit other doc pages as that increases risk beginners will get lost.

Was the typography file lost in your refactor @amberleyromo?

marcysutton:
Ah, good point @KyleAMathews! That would be good to look into.

KyleAMathews:
The tutorial did used to tell you to write out a typography.js config file — the screenshots and some of the language don't make sense without it so I restored the previous flow https://github.com/gatsbyjs/gatsby/pull/12131

marcysutton:
Thank you, @KyleAMathews! 💪

PR #12107, opened 2019-02-26
chore(docs): grammatical tweak to page query docs

thekogmo:
small grammar change

### DIFF
--- a/docs/docs/page-query.md
+++ b/docs/docs/page-query.md
@@ -2,7 +2,7 @@
 title: Querying data in pages with GraphQL
 ---
 
-Gatsby's `graphql` tag enables page components to retrieve data via GraphQL query.
+Gatsby's `graphql` tag enables page components to retrieve data via GraphQL queries.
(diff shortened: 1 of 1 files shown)


docs/docs/page-query.md
```diff
@@ -2,7 +2,7 @@
 title: Querying data in pages with GraphQL
 ---
 
-Gatsby's `graphql` tag enables page components to retrieve data via GraphQL query.
+Gatsby's `graphql` tag enables page components to retrieve data via GraphQL queries.
```
wardpeet:
```suggestion
Gatsby's `graphql` tag enables page components to retrieve data via a GraphQL query.
```

We only support one per component so I guess this is better wording

thekogmo: I agree 👍

thekogmo:
i noticed the label is still "awaiting author's response". 
the requested change has been added, is there anything else to add?

Issue #12079, opened 2019-02-25
gatsby-plugin-feed requires creating slugs for pages


johnreynoldsventure:
This is probably obvious to many, but while adding gatsby-plugin-feed to my site I discovered that you must add slugs to your MarkdownReact nodes.

The documentation for how to do this is very clear: https://www.gatsbyjs.org/docs/creating-slugs-for-pages/

What wasn't clear (to me) was that this was a dependency for gatsby-plugin-feed.

I hope this helps others...

KyleAMathews:
The plugin shouldn't work this way. You should pass into the configuration which field to use does the slug. Would you be interested in working on a PR for this?

johnreynoldsventure:
I'd love to help more Kyle, but I fear I don't know enough - I'm just getting started with gatsby.
I started with the gatsby tutorial, and have been added functionality as I need it.  Likely my problems are due to me missing a step or two along the way rather than anything fundamental.

tomoyukikashiro:
@johnreynoldsventure 

> What wasn't clear (to me) was that this was a dependency for gatsby-plugin-feed.

Your are right. :)  `gatsby-plugin-feed` require `slug` by default. 
=> [here is the code](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-feed/src/gatsby-node.js#L16-L17) where plugin use `slug` to define each page url.

But you can costomize it by passing option into `gatsby-config.js`.

In my example I try to change default site url from `https://example.com/${slug}` to `https://example.com/posts/${slug}` in feed.
=> [here is the code](https://github.com/tomoyukikashiro/gatsby-starter-blog-amp-to-pwa/commit/f69c56666a28bf6dead474b76e095e247998a795).

laurosilvacom:
I'll like to work on a PR @KyleAMathews. Can you point me on the write direction?

VGoose:
I think this should be fixed by updating the docs stating that the default serializer requires a slug field, rather than any change to the plugin.

DSchau:
@VGoose we actually did that recently -> https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-feed#how-to-use

(We don't note what you need to do to use the internal serialize, just that we recommend you write your own!)

VGoose:
@DSchau any reason to not be explicit about the requirements of the default serializer? The issue was opened because it wasn't clear.

PR #12108, opened 2019-02-26
Update gatsby-plugin-feed docs

Bilie:
## Description

Updates `gatsby-plugin-feed` docs to make usage clearer.
Looking forward to your feedback!

## Related Issues

Related to #12079


marcysutton:
Hi, thanks for the contribution! You should also check out https://github.com/gatsbyjs/gatsby/pull/12085, as some of the changes in this PR will be out of date when that gets merged.

There is also a new RSS doc: https://github.com/gatsbyjs/gatsby/blob/master/docs/docs/adding-an-rss-feed.md

Bilie:
Hi Marcy, I saw the related issues and PRs after I opened this one. If you think these changes will be obsolete after the other PRs get merged, then we can close this PR. Thank you for having a look at it!

DSchau:
@bilie I missed your comment that we can close!

Let’s do that—but I do think this could be a super worthwhile follow-up PR if you’re up for it.

Thanks for this contribution!

PR #12114, opened 2019-02-26
chore(site-showcase): add GVD Renovations site

dkrasniy:
Added GVD Renovations Gatsby site to showcase!

### DIFF
--- a/docs/sites.yml
+++ b/docs/sites.yml
@@ -4437,9 +4437,8 @@
   description: >
     GVD Renovations is a home improvement contractor with a well known reputation as a professional, quality contractor in California.
   categories:
-    - Home Improvement
+    - Construction
     - Business
-    - Contractor
   built_by: David Krasniy
   built_by_url: http://dkrasniy.com
   featured: false
--- a/docs/sites.yml
+++ b/docs/sites.yml
@@ -4431,7 +4431,7 @@
     - Featured
   built_by: Algolia
   featured: true
-  - title: GVD Renovations
+- title: GVD Renovations
   url: https://www.gvdrenovations.com/
   main_url: https://www.gvdrenovations.com/
   description: >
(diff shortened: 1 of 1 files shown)


docs/sites.yml
```diff
@@ -4431,3 +4431,15 @@
     - Featured
   built_by: Algolia
   featured: true
+- title: GVD Renovations
+  url: https://www.gvdrenovations.com/
+  main_url: https://www.gvdrenovations.com/
+  description: >
+    GVD Renovations is a home improvement contractor with a well know reputation as professional, quality contractor in California.
```
DSchau:
```suggestion
    GVD Renovations is a home improvement contractor with a well known reputation as a professional, quality contractor in California.
```

docs/sites.yml
```diff
@@ -4431,3 +4431,15 @@
...
+  categories:
+    - Home Improvement
```
DSchau: Could you do me a quick favor and ensure that these categories are more generally applicable?

In other words--if you check out https://www.gatsbyjs.org/showcase/, these categories should have appeared before if at all possible! We want to keep these as "general" as possible, so in this case I'd recommend something like:

```yaml
categories:
  - Business
  - Real Estate
```

and then any other that you think fit! Please let me know if you have any questions on this!

docs/sites.yml
dkrasniy: Sure! Thanks for catching that

docs/sites.yml
dkrasniy: I see. Let's go with 

- Construction
- Business

DSchau: Works for me!

dkrasniy:
Committed the two changes. Let me know if I need to do anything else. Thanks for the quick turnaround!

DSchau:
Thanks for using Gatsby! Merging this in!

PR #12123, opened 2019-02-26
chore(docs): add missing redirects for contributing docs

marcysutton:
## Description

I missed a few redirects in the last edits of the contributing docs, so the How to File an Issue and How to Label an Issue pages are still coming up in Google with their old URLs. 

Note: it would be cool to have an automated check that warns you of moving a page without adding a redirect.


marcysutton:
Woo! Thanks @DSchau

PR #12124, opened 2019-02-26
Add markdown link checking CI task

RDIL:
## Description
The tool will search all markdown files for dead links.  

## Related Issues
N/A


.circleci/config.yml
```diff
@@ -188,6 +188,7 @@ jobs:
       - <<: *install_node_modules
       - <<: *persist_cache
       - run: yarn markdown
+      - run: yarn markdown-link-check
```
DSchau: This seems a little misplaced here as part of validating starters.

Probably want a separate task?

package.json
```diff
@@ -80,6 +81,7 @@
     "lint:flow": "babel-node scripts/flow-check.js",
     "lint:other": "npm run prettier -- --check",
     "markdown": "md-magic --path \"starters/**/*.md\"",
+    "markdown-link-check": "find . -name \*.md -exec markdown-link-check {} \;",
```
DSchau: I think this is invalid JSON. It's failing CircleCI with:

`An unexpected error occurred: "/home/circleci/project/package.json: Unexpected token * in JSON at position 2667". `

RDIL: Fixed

.circleci/config.yml
RDIL: Fixed

DSchau:
@RDIL thanks! So when you add a dependency to the root package.json (or really any package) you also need to update the `yarn.lock` file.

If you run `yarn` in the root of the repo, you should get a changed `yarn.lock` file, and you'll want to commit that to this branch!

RDIL:
@DSchau 
after an hour of killing my internet I have done it!

.circleci/config.yml
```diff
@@ -104,6 +104,12 @@ jobs:
           root: packages
           paths:
             - "*"
+  markdown:
```
DSchau: You'll also want to add this to the jobs below! Sorry - forgot to mention this earlier!

RDIL:
@DSchau so basically JavaScript object notation doesn't seem to support escape characters, this should now be fixed however.

RDIL:
Also had to move it completely to the CI config file.

DSchau:
Alright - thank you so much!

Few issues, here unfortunately. I'm not 100% sure they're solvable, but want to note them so you're aware!

1. The check takes 31 minutes--which will not work for us
    - We could improve this by only running this check on a subset of the files, e.g. the files _staged_ in this PR perhaps
1. The check is failing (check the [CI task output](https://circleci.com/gh/gatsbyjs/gatsby/72787?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)), but is not actually failing the CI task
1. The "failures" seem to be quite a number of false positives
    ```
    [✖] https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-canonical-urls/compare/gatsby-plugin-canonical-urls@2.0.9...gatsby-plugin-canonical-urls@2.0.10
    [✖] https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-canonical-urls/compare/gatsby-plugin-canonical-urls@2.0.8...gatsby-plugin-canonical-urls@2.0.9
    [✖] https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-canonical-urls/compare/gatsby-plugin-canonical-urls@2.0.7...gatsby-plugin-canonical-urls@2.0.8
    [✖] https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-canonical-urls/compare/gatsby-plugin-canonical-urls@2.0.6...gatsby-plugin-canonical-urls@2.0.7

    ```
1. I do not believe this will detect our "custom" links, e.g. we'll want to re-write something like `/docs/path-to-doc/` with a custom checker, e.g. that the file docs/path-to-doc.md _exists_ rather than it's a "live" link (or we could just prepend https://gatsbyjs.org, but this won't work with newly introduced docs)

All of this being said--I think this is a really good idea, just not sure it's quite in a state where we could get this merged! I'd be happy to pair with you on this if you're interested, [more info available here](https://www.gatsbyjs.org/contributing/pair-programming/).

RDIL:
Hmm. Not sure how some of these could be solved.  Would you only want one subdirectory being checked to speed it up?  The failures are a known bug for diffs like those.

RDIL:
Full response to all your points:
1. I limited it to just docs to speed it up
2. Its meant as a static analysis task that passes even if dead links are found because it is a minor check
3. Diffs on GH bug out
4. Agreed
5. True

RDIL:
If you want I can just close this

DSchau:
@RDIL not what I want whatsoever!

Would you be interested in pairing on this? I think this is actually a really valuable PR because we should always focus on automating out "simple" tasks so that we can focus on more complex ones and better use our time.

I'd propose a [pairing session](https://www.gatsbyjs.org/contributing/pair-programming/) so that we can talk through this in a bit more detail, and then get developing the improvements, as well.

If you're not interested--no worries, it just may be quite a bit more work than you're expecting to get this merged in. Happy to help however you prefer, though!

RDIL:
I will sign up

RDIL:
Actually, I'm willing to do the work.  I'm not the best at human interactions.  I'll take the you help option if that's fine with you

DSchau:
@RDIL I don't think this PR will be mergeable, and it's something that would be much more productive to work on together. That being said--totally understand that that's not for everyone.

If you're interested in fixing this up, here are the issues that would need solved and some general advice for how to solve them:

- [ ] False positives cannot happen
    - We could use unist-util-visit, parse a Markdown file into an AST, and then re-write the links (e.g. /docs/whatever would be rewritten to the relative docs folder, and we'd validate that file exists)
    - If you're looking for some prior art here, most packages/gatsby-remark-* generally act upon a markdownAST and mutate/traverse its structure (e.g. we'd parse into a Markdown AST, and then visit all links)
- [ ] A CI check should fail if dead links are detected
    - A CI check that passes with invalid input is not super useful to us--we'd introduce a manual workflow into a process that should be automated
    - This is probably a matter of checking the options for the tool we use, or whether we build it ourselves
- [ ] Output needs to be scannable
    - If we're scanning this at a glance, it needs to be quick to us (or the author of the PR, or anyone!) _why_ the check failed
    - Something like [`prettier --check`](https://prettier.io/docs/en/cli.html#check) would be very helpful here
- [ ] Output needs to be quick
    - We can use `git` here, and get the list of changed files in the commit vs. upstream/master. This will make the scan as quick as possible!

As you noted earlier, it's _OK_ to close this--but that being said, I hope I've given you enough to go on so that you can re-open this PR with some improvements and we'd love to take a look. As always--we'll happily answer any questions and help in any way we can.

Thanks again!

RDIL:
I had a feeling that this would get used.  As I'm not the author of the tool, I can't really do many of these improvements, so I'm not going to take a shot this PR again, and I have nothing more to contribute so farewell.

PR #12126, opened 2019-02-26
fix(www): make GitHub icon above plugin readme clearer

kbrgl:
It was hard to find and isn't very clear.

## Description
Adds 'View plugin on GitHub' next to the GitHub icon, and introduces some spacing to make it easier to find. It wasn't very clear earlier where it goes (view source on GitHub could mean either page source or plugin source.)

Just a minor something I've clashed with a few times 😄 

screen shot 2019-02-27 at 2 59 46 am

### DIFF
--- a/www/src/components/package-readme.js
+++ b/www/src/components/package-readme.js
@@ -42,7 +42,6 @@ const PackageReadMe = props => {
           }}
           href={githubUrl}
           aria-label={`${packageName} source`}
-          title={`View plugin on GitHub`}
         >
           <GithubIcon focusable="false" style={{ verticalAlign: `text-top` }} />&nbsp;&nbsp;<span>View plugin on GitHub</span>
         </a>
--- a/www/src/components/package-readme.js
+++ b/www/src/components/package-readme.js
@@ -27,13 +27,12 @@ const PackageReadMe = props => {
         <a
           css={{
             "&&": {
-              display: githubUrl ? `inline-block` : `none`,
+              display: githubUrl ? `flex` : `none`,
               fontWeight: `normal`,
               border: 0,
               color: colors.gray.calm,
               boxShadow: `none`,
-              display: 'flex',
-              alignItems: 'center',
+              alignItems: `center`,
               "&:hover": {
                 background: `none`,
                 color: colors.gatsby,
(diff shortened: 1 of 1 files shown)


www/src/components/package-readme.js
```diff
@@ -40,9 +42,8 @@ const PackageReadMe = props => {
           }}
           href={githubUrl}
           aria-label={`${packageName} source`}
-          title={`View source on GitHub`}
         >
-          <GithubIcon focusable="false" style={{ verticalAlign: `text-top` }} />
+          <GithubIcon focusable="false" style={{ verticalAlign: `text-top` }} />&nbsp;&nbsp;<span>View plugin on GitHub</span>
```
m-allanson: Can you remove the `&nbsp;&nbsp;` and instead add the spacing with CSS?

www/src/components/package-readme.js
```diff
@@ -40,9 +42,8 @@ const PackageReadMe = props => {
           }}
           href={githubUrl}
           aria-label={`${packageName} source`}
```
m-allanson: Maybe remove this `aria-label` and replace it with an [`aria-labelledby` attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute) that links to the text `<span>View plugin on GitHub</span>`? 

cc @marcysutton is that best practice here?

wardpeet: I'm pretty sure they do the same thing but it's better to not repeat yourself so 👍 for `aria-labelledby`

www/src/components/package-readme.js
kbrgl: Yep, no problem!

www/src/components/package-readme.js
kbrgl: Sure 👍

kbrgl:
@m-allanson the text doesn't fit at 320px, so I've added a media query to hide it at widths of 385px and below. Here's a look at the text at just under 385px:

screenshot_2019-02-27 gatsby-plugin-sharp

And just over:

screenshot_2019-02-27 gatsby-plugin-sharp 1

I've fixed the other issues.

fk:
Hey @kbrgl, thanks so much for this PR — I completely agree with the changes, 💯 to both.

Wondering if we should slightly adjust the "See starters that use this" to "See starters using this"?

In regards to the small screen issues: Looking at #11753 incoming, I think we won't get around breaking/wraping items into rows instead of hiding labels. We also want to unfiy link styles 😅, but IMO that's a thing for another PR.

kbrgl:
Hi, @fk!

Here's how that'll look:

screenshot_2019-02-27 gatsby-plugin-react-helmet

fk:
@kbrgl #12253

www/src/components/package-readme.js
```diff
@@ -27,22 +27,35 @@ const PackageReadMe = props => {
...
               "&:hover": {
                 background: `none`,
                 color: colors.gatsby,
               },
             },
           }}
           href={githubUrl}
-          aria-label={`${packageName} source`}
-          title={`View source on GitHub`}
+          aria-labelledby="github-link-label"
```
marcysutton: For future reference, `aria-labelledby` isn't needed on a link if it contains text content.

PR #12128, opened 2019-02-27
feat(gatsby): add assetPrefix to support deploying assets separate from html

### LINKED ISSUE #2335: Ability to deploy non-html files to separate domain

### LINKED ISSUE #8833: Question: Putting Assets in a Different Folder

DSchau:
## Description

If at first you don't succeed, try, try again 😅 

The first attempt at asset prefixing went _OK_, but I think there was a fair enough amount of complexity that trying something new and hopefully simpler is valuable. Shout out to @pieh for the simpler idea here.

This PR introduces the ability to add `assetPrefix` to your gatsby-config.js. This should be a URL, and it is on the user of this functionality to ensure that assets are copied appropriately to this destination. It works with `pathPrefix`, as well, but again, slightly manual--for now!

As someone using this feature, here's how you would use it:

- Add an `assetPrefix` option to your `gatsby-config.js`
  ```js
  module.exports = {
    assetPrefix: `https://cdn.example.com`
  }
  ```
- Continue to use functionality in your site that you've already been using, e.g. `Link`, `withPrefix`, etc. (no code or plugin changes should be required)
- `__PATH_PREFIX__` will be the joined result of `assetPrefix` and `pathPrefix`

Alright, alright. Now let's build this.

At _build time_, you will:

- Run `gatsby build --prefix-paths` (just like you did with `pathPrefix`) and this will be respected whether using either of pathPrefix or assetPrefix
    - If you're using `pathPrefix`, you'll need to copy the files in public to a nested directory (e.g. if using /blog, will need to create move /public/ to /public/blog and deploy with your hosting provider of choice)
- The `public` directory is _assets_ (HTML, CSS, JS, etc.) so this can be uploaded to wherever, e.g. a CDN
    - Feel free to exclude HTML, but otherwise everything in this folder should _already_ be an asset

This will create a version of your site that has its assets (CSS, JS, images, etc.) prefixed with `${assetPrefix}${pathPrefix || ''}`, so those assets will need to be deployed and available. If so--this works just fine!

See [gatsby-asset-prefix.netlify.com](https://gatsby-asset-prefix.netlify.com/blog) and the [Github Repo](https://github.com/dschau/gatsby-asset-prefix) for an example of this feature in use.

## To do

- [x] Add documentation for assetPrefix
- [ ] Add more tests
    - Not sure best approach of testing this e2e, but it'd be super nice to have some e2e tests for this functionality
- [x] Fix relative path being used with asset prefix
    - [url.resolve doesn't respect relative paths for whatever reason](https://runkit.com/dschau/5c7711b616d1d00012405c7a), so will just switch to a [].join('/') or something
- [x] Fix e2e test failures (they seem legitimate)
- [x] Hard fail (?) when gatsby-plugin-offline is used with this feature; I _do not believe_ you can serve service workers from a separate domain
    - It _may_ be possible to get this working with some custom Workbox stuff, but I think the service worker will still need to be served from the _same_ domain
    - cc @davidbailey00 any insight here? Would highly appreciate it!
- [] Figure out what plugins--especially third party--this could break (generally, pathPrefix is used for assets, but it is also used as a "base path", e.g. for links)

## Related Issues

Fixes #2335, Fixes #8833 (need to validate this one--but pretty sure it does!)


packages/gatsby/src/commands/serve.js
```diff
@@ -65,6 +65,15 @@ module.exports = async program => {
     }
     return next()
   })
+  app.use(function(req, res, next) {
```
DSchau: Could probably use cors here 🤷‍♂️

packages/gatsby/src/joi-schemas/joi.js
```diff
@@ -1,10 +1,18 @@
 const Joi = require(`joi`)
 
+const stripTrailingSlash = chain => chain.replace(/(\w)\//, `$1`)
```
DSchau: This is a nice addition - rather than requiring documentation (e.g. don't add a trailing slash), this will just remove it for us.

It only impacts a string with content, e.g. `/` is left alone.

packages/gatsby/src/joi-schemas/joi.js
```diff
@@ -1,10 +1,18 @@
...
+
 export const gatsbyConfigSchema = Joi.object().keys({
   __experimentalThemes: Joi.array(),
   polyfill: Joi.boolean(),
   siteMetadata: Joi.object(),
-  pathPrefix: Joi.string(),
+  assetPrefix: stripTrailingSlash(Joi.string().uri()),
```
DSchau: This _could possibly_ be a relative URL too--so will investigate as to whether that makes sense.

DSchau:
Will need to add one more thing: a helper to create `publicPath` (what this really is) and pass that as `pathPrefix` to plugins.

packages/gatsby/src/joi-schemas/joi.js
DanielBrierton: Our use case would definitely benefit from using a relative URL.

We're using a Cloudflare worker to serve content from a Gatsby site for certain routes. Our issue right now is that the assets are dumped into the root of `public`. We've worked around this in Gatsby v1 using `--prefix-paths` and having our worker serve Gatsby files for the `prefixPath`.

With Gatsby v2 this is also prefixing our routes, which causes our homepage to redirect to our `prefixPath`. We could certainly work around having provide a full URI, but it seems relative would be easier for us.

After adding `allowRelative: true` here, it seems to work well for our use case.

DSchau: @DanielBrierton did you actually check this feature out--or do you mean that the relative URL validation will work for your use case?

May be helpful if we could get together this week or next and see if this feature works for your use case, and actually get something deployed!

DSchau:
✅ e2e-tests passing in b144bd2

docs/docs/asset-prefix.md
```diff
@@ -0,0 +1,79 @@
+---
+title: Adding an Asset Prefix
+---
+
+Gatsby produces static content that can be hosted _anywhere_ at scale in a cost-effective manner. This static content is comprised of HTML files, JavaScript, CSS, images, and more that power your great, Gatsby application.
+
+There are scenarios in which it may be advantageous or necessary to deploy _assets_ (where an asset is a non-HTML resource, e.g. JavaScript, CSS, etc.) to a separate domain, oftentimes to enable functionality like a CDN or a particular hosting strategy that your company may employ where assets need to be separate from the core application.
```
DSchau:
```suggestion
There are scenarios in which it may be advantageous or necessary to deploy _assets_ (where an asset is a non-HTML resource, e.g. JavaScript, CSS, etc.) to a separate domain. This can oftentimes be used to enable functionality like a CDN or a particular hosting strategy that your company may employ where assets need to be separate from the core application.
```

lol - suggestions work on my own PR 🎉

packages/gatsby/src/joi-schemas/joi.js
DanielBrierton: Yeah, I checked out your PR and made the change to actually test it. I haven't tested it in our production/staging environment, but I'm confident it will work based off how it worked in my local dev env.

Happy to help testing it. We were originally going to make a plugin based on https://github.com/gatsbyjs/gatsby/issues/8833#issuecomment-458147983 (which works too), but an actual supported feature in Gatsby would obviously be much more preferable.

eballeste:
can't wait for this, thanks!

DSchau:
@eballeste if you are willing to test, I can write up some instructions for how to use this PR in its current state with a project of yours. Will help us e2e validate, as well!

GuillaumeAmat:
@DSchau I'd be happy to test your PR on my project 👌

DSchau:
@GuillaumeAmat sweet!

So first thing, check out [the documentation in this PR](https://github.com/gatsbyjs/gatsby/blob/fe0f94a563d5462e51c2afbac00f8f2e9e7f43ad/docs/docs/asset-prefix.md)

Next, presuming you want to deploy this (which makes sense) we'll have to create some tarballs that we can use.

1. Clone the gatsbyjs/gatsby repo
1. Add this repo as a remote, e.g. `git remote add dschau git@github.com:DSchau/gatsby.git`
1. Fetch this branch, `git fetch dschau gatsby/asset-prefix-improved`
1. Run `yarn bootstrap` to install dependencies and build out packages
1. Create tarballs
  - `cd packages/gatsby && yarn pack -f gatsby.tgz`
  - `cd packages/gatsby-link && yarn pack -f gatsby-link.tgz`
1. change directories to your project
1. copy tarballs to your project, e.g. `cp <absolute-path-to-the-gatsby-repo>/packages/gatsby/gatsby.tgz .` && `cp <absolute-path-to-the-gatsby-repo>/packages/gatsby-link/gatsby-link.tgz .`
1. Update package.json with the .tgz files, e.g.
    ```json
    {
      "dependencies": {
        "gatsby": "./gatsby.tgz",
        "gatsby-link": "./gatsby-link.tgz"
      }
    }
    ```
1. Run `yarn` or `npm install`
1. Add the `assetPath` functionality (refer to the documentation)
1. Deploy, and ensure you've followed some approach to deploy to your cdn/separate domain/etc with the assets

Please let me know if you have any questions, or if you run into any issues!

GuillaumeAmat:
Great, I'll test it tomorrow and let you know!

GuillaumeAmat:
@DSchau: I followed the steps and it seems to work perfectly for the imported/required assets. Good job! 👌

Two of my needs are not fulfilled though:

* `withPrefix` seems to not care about the `assetPrefix` config
* The `static` folder assets are not handled

I'm also wondering about the need of `gatsby-link`. Isn't it "deprecated" since Gatsby v2?

DSchau:
@GuillaumeAmat sorry for the delay!

> withPrefix seems to not care about the assetPrefix config

Correct - and this is probably desired behavior. This should be used for internal linking, e.g. to other pages. It _should_ respect pathPrefix, but not necessarily assetPrefix. Thinking through how to best handle this, because this means that `gatsby-plugin-manifest` will need to slightly change--and possibly others.

Is a manifest an asset? Something else? Kind of a misfit.

> The static folder assets are not handled

This is not really my findings. Could you share more detail here?

If you check out https://gatsby-asset-prefix.netlify.com/blog/ (just deployed with latest changes), you can see the /static files are prefixed like you'd expect.

> The icons generated by gatsby-plugin-manifest are not handled either

Yep.

> I'm also wondering about the need of gatsby-link. Isn't it "deprecated" since Gatsby v2?

It's a dependency of gatsby that we export from the top-level package (gatsby) for convenience. It still gets updated and changed fairly frequently!

davidbailey00:
@DSchau agh sorry for the late reply!

Actually SWs can handle files from another domain and cache these - just as normal if using CORS, or an "opaque" response without CORS. So I think it should actually be fine to keep plugin-offline as-is here!

DSchau:
@davidbailey00 sweet!

Yeah - if we can avoid weird conditions where plugins "just work" with this feature, even better.

This should be in a state you can test in, so please let me know if I can help with anything. I have some instructions above for testing, as well.

szimek:
@DSchau I've just tried to build our blog app using the latest version of your branch. I'm also using `pathPrefix` there and I've got `pathPrefix` set to `/en/blog` and `assetPrefix` to `/assets/blog`. 

The problem is that when `assetPrefix` is set to a path and not a domain, the generated paths for asset files are incorrect. The generated paths look e.g. like this:
```
/assets/blog/en/blog/webpack-runtime-29ed2b43b1c8515d6c84.js
```
so the browser is trying to fetch them from e.g. `http://localhost:8081/assets/blog/en/blog/webpack-runtime-29ed2b43b1c8515d6c84.js`, which returns 404.

The issue is that when `assetPrefix` is a path and not a domain, the order should be reversed, i.e. `/<pathPrefix>/<assetPrefix>/<path>`, so in this particular case it should be:
```
/en/blog/assets/blog/webpack-runtime-29ed2b43b1c8515d6c84.js
```

szimek:
I modified our build script and replaced `pathPrefix` with `assetPrefix` and it works almost perfectly! ;)

The only issue I found so far is with `gatsby-plugin-sitemap`, where I'm getting all my paths prefixed with `assetPrefix`, so e.g. when `assetPrefix` is set to `/assets/blog`, I'm getting urls like `https://www.freeletics.com/assets/blog/en/blog/posts/women-of-freeletics/` instead of `https://www.freeletics.com/en/blog/posts/women-of-freeletics/`. Fortunately, I can provide my own serializing function to remove `/assets/blog` string from all generated urls, so it's not a big issue, but it would be great if it just worked.

GuillaumeAmat:
@DSchau It seems to work like a charm!

Like @szimek, my only issue is about a plugin though. The icons generated by `gatsby-plugin-manifest` are not prefixed, but you already know that.

Question is, do you plan to change its behavior? What's your opinion about that? I personally think that it should be handled as every other assets but I may miss something.

DSchau:
@GuillaumeAmat definitely want to handle.

Shipping this with broken plugins (core plugins, no less!) is definitely not something I'm comfortable with.

As I see it, these are the current blockers for this PR:

- [ ] `gatsby serve` does not seem to work correctly with a relative assetPrefix _and_ pathPrefix
- [ ] Some plugins (notably gatsby-plugin-manifest, gatsby-plugin-sitemap, gatsby-mdx, and gatsby-plugin-offline) may be broken with this change, and should be fixed
- [ ] General weirdness around _relative_ asset prefix, as @szimek noted

DSchau:
Extra extra--read all about it!

We're going to publish this next week under a tag (e.g. `gatsby@asset-prefix`) once we've ironed out some of the bugs I mentioned [here](https://github.com/gatsbyjs/gatsby/pull/12128#issuecomment-472068564).

Would appreciate any early adopters to trial it out and then if all goes good--we'll publish this in a minor release of Gatsby after a week or two. Thanks all for your patience!

GuillaumeAmat:
Great to read @DSchau! 🙌

I'd be happy to test and feedback the pre-release.

andrewstaker:
Great work @DSchau ! I'm testing this on our project and haven't run into any issues yet.

brookemckim:
Also testing this out and it looks good. Thank you @DSchau! 

However, it would make it much more flexible if the assetPrefix could contain arbitrary paths after the URL. Using pathPrefix does this, but pathPrefix also changes the behavior of links and redirects. My use case is just wanting to have the assets under a different path.

szimek:
@brookemckim After this feature is merged, I'd like to introduce `assetPath` option, which would be used to specify the path only. `assetPrefix` would then be used to specify the domain. This would easily allow Gatsby to output all files into `./public/${assetPath}` folder, so that you don't have to move/copy them yourself anymore and would make it easier to generate public urls and local paths to asset files. 

Unfortunately, even this feature is going to be accepted, it wouldn't end up in 2.x branch anyway, because it would be a breaking change and it does require a lot of changes in the way not only Gatsby, but also plugins work.

franksmule:
Will be happy to test once the tag is released

DSchau:
> However, it would make it much more flexible if the assetPrefix could contain arbitrary paths after the URL

This would work just fine?

If you set an assetPrefix to `https://cdn.example.com/whatever` that should work as long as the assets are accessible at that URL? I'll write a test to validate, but that should work like you expect.

brookemckim:
> > However, it would make it much more flexible if the assetPrefix could contain arbitrary paths after the URL
> 
> This would work just fine?
> 
> If you set an assetPrefix to `https://cdn.example.com/whatever` that should work as long as the assets are accessible at that URL? I'll write a test to validate, but that should work like you expect.

This fails validation and it says it is not a valid url.

DSchau:
> This fails validation and it says it is not a valid url.

Not sure this is the case! If we look at the [validation](https://github.com/gatsbyjs/gatsby/pull/12128/files#diff-bd517ec758aa05f3e6e1c34b15f65394R9) we can see it's just validating as a URL.

One thing that I did find is that `stripTrailingSlash` was creating an invalid URL--so I've fixed that up and [added a test here](https://github.com/gatsbyjs/gatsby/pull/12128/commits/39403b8ac211d9c70b7fc028772aed37a1970807).

Could you provide any more details as to how this failed validation?

brookemckim:
> > This fails validation and it says it is not a valid url.
> 
> Not sure this is the case! If we look at the [validation](https://github.com/gatsbyjs/gatsby/pull/12128/files#diff-bd517ec758aa05f3e6e1c34b15f65394R9) we can see it's just validating as a URL.
> 
> One thing that I did find is that `stripTrailingSlash` was creating an invalid URL--so I've fixed that up and [added a test here](https://github.com/gatsbyjs/gatsby/pull/12128/commits/39403b8ac211d9c70b7fc028772aed37a1970807).
> 
> Could you provide any more details as to how this failed validation?

Ah, that was probably it. Thanks for doing this! I will give it another go. This will allow me to deploy Gatsby behind an AWS ALB with path based routing.

DSchau:
Just published a few packages with the tag `asset-prefix` that will make testing this feature out super easy.

```
gatsby-link@asset-prefix
gatsby-plugin-catch-links@asset-prefix
gatsby-plugin-feed@asset-prefix
gatsby-plugin-manifest@asset-prefix
gatsby-plugin-offline@asset-prefix
gatsby-plugin-sitemap@asset-prefix
gatsby@asset-prefix
```

If you're interested in testing out this feature, it will look something like:

```
npm install --save gatsby@asset-prefix gatsby-link@asset-prefix # and any additional plugins you're using in the above list!
```

## Known Issues

- [ ] Relative assetPrefix is not usable with pathPrefix
    - I'm not sure how to best resolve this and how common it is
    - Note: also there are no warnings for this yet--but if we decide to build in this limitation, I'll add that
- [ ] I think gatsby-mdx may be broken with this change, will validate in a bit

## Notes

When this feature is published, gatsby-link will not need to be explicitly installed since gatsby depends upon it (and will depend upon the version that contains the assetPrefix feature).

## Wrap-up

Please use this issue to report any feedback you have, good, bad, or otherwise.

Thank you so much!

szimek:
@DSchau Thank you so much! 

Unfortunately, the sitemap plugin still doesn't work correctly with relative `assetPrefix` option, because its value is passed as `pathPrefix` (which is a bit confusing ;)) in `onPostBuild` hook in the plugin and then used to generate urls:
```js
page.node.path = (pathPrefix + page.node.path)...
...
return {
  url: site.siteMetadata.siteUrl + edge.node.path,
  ...
}
```
and I'm ending up with urls like `https://www.freeletics.com/<assetPrefix>/en/blog/posts/women-of-freeletics/`. Fortunately, like I mentioned before, it's possible to provide a custom `serialize` function in the plugin options to fix it.

DanielBrierton:
Not sure why, but I had to install `gatsby-link@asset-prefix` specifically, otherwise I was getting my `href`s prefixed with the `assetPrefix`.

Now that I've installed `gatsby-link` though, it all works well. We're not using any of the other plugins though, for what it's worth.

docs/docs/path-prefix.md
```diff
@@ -2,46 +2,82 @@
...
+      {/* highlight-next-line */}
+      <Link to="page-2">Page 2</Link>
+    </Layout>
+  )
+}
+```
+
+Without doing _anything_ and merely using the `Link` component, this link will be prefixed with our specified `pathPrefix` in `gatsby-config.js`. Woo hoo!
+
+If we want to do programattic/dynamic navigation, totally possible too! We expose a `navigate` helper, and this too automatically handles path prefixing.
```
muescha:
```suggestion
If we want to do programatic/dynamic navigation, totally possible too! We expose a `navigate` helper, and this too automatically handles path prefixing.
```

packages/gatsby-link/src/index.js
```diff
@@ -1,4 +1,4 @@
-/*global __PATH_PREFIX__ */
+/*global __BASE_PATH__ */
```
muescha: path prefix is also global?

```suggestion
/*global __BASE_PATH__ */
/*global __PATH_PREFIX__ */
```

packages/gatsby-link/src/index.js
```diff
@@ -8,7 +8,11 @@ import { parsePath } from "./parse-path"
 export { parsePath }
 
 export function withPrefix(path) {
-  return normalizePath(`${__PATH_PREFIX__}/${path}`)
+  return normalizePath(`${__BASE_PATH__}/${path}`)
+}
+
+export function withAssetPrefix(path) {
+  return `${__PATH_PREFIX__}/${path}`
```
muescha: why is the `withAssetPrefix` not populated with `__ASSET_PREFIX__`

from:
https://github.com/gatsbyjs/gatsby/blob/39403b8ac211d9c70b7fc028772aed37a1970807/packages/gatsby/src/utils/webpack.config.js#L184-L186

muescha: is here no `normalizePath` necessary?

leonardodino:
> I think gatsby-mdx may be broken with this change, will validate in a bit

Altough my use of `gatsby-mdx` is pretty limited and simple,
it's working well on my website with the `assetPrefix`. :tada:

I'm using: `gatsby-mdx@0.4.3` + `@mdx-js/mdx@0.20.3`.
both are pinned at those versions because of patches. 😅

franksmule:
I'm using this branch and its working well, but I've had to patch it as .json files require CORS headers. 
(So I serve .json files relatively and all other files with the prefix)
Is there anyway to disable Gatsby loading these context .json files? as in my case they're all empty anyway.

sami616:
This looks great 👌

szimek:
@DSchau Is there anything I could do to help get this PR merged? 

I just got an invitation for Gatsby Preview (yay!) and I'd like to simplify our quite complicated build process using `assetPrefix` option before trying it out  ;)

DSchau:
@szimek it'll be my top priority early next week! Goal is to have it merged by the Community Maintainers meeting on Wednesday! 🚀

leonfs:
We've been using it for the last 3 weeks, so far we haven't seen any issues.

packages/gatsby-link/src/index.js
DSchau: These aren't really needed anyways. I'll just remove the comment.

DSchau: Also - none of these are used in this file--so this comment was a little misplaced to begin with!

packages/gatsby-link/src/index.js
DSchau: __PATH_PREFIX__ itself is populated with assetPrefix. normalizePath would strip the double slashes, e.g. if assetPrefix is set to `https://cdn.example.org` and `pathPrefix` is set to `/blog`, normalizePath would create a structure like `https:/cdn.example.org/blog/${path}` which is not what we want.

DSchau: And it _is_ prefixed with __ASSET_PREFIX__. We want to use the full path here, e.g. assetPrefix joined with pathPrefix.

packages/gatsby/src/utils/eslint-config.js
```diff
@@ -7,6 +7,7 @@ module.exports = schema => {
       globals: {
         graphql: true,
         __PATH_PREFIX__: true,
+        __BASE_PATH__: true, // this will rarely, if ever, be used by consumers
```
DSchau: For more context here--it's used in the internal linking (e.g. with just pathPrefix) because pathPrefix itself is overloaded.

DSchau:
Alright folks!

This seems ready for primetime. I've gone ahead and re-published all packages behind the tag `asset-prefix`. Unless there are any objections, I'd like to publish this tomorrow before the community maintainers meeting at 8:45 AM PST.

So--if anyone has some time before then, would very much appreciate any final feedback before we get this merged. I've gone ahead and made some opinionated choices (that I think are for the better), specifically:

- assetPrefix and pathPrefix cannot both be relative
    - There's some weirdness here in choosing which goes first, and it's just generally far cleaner to use one or the other, or an absolute assetPrefix with a relative pathPrefix
- If using a version of gatsby <2.4.0, this should not _break_ as it will just fall back to `withPathPrefix` helper if `withAssetPrefix` does not exist.
- all plugins that now rely on withAssetPrefix (from gatsby-link) now have a peerDependency on gatsby of at least 2.4.0 (the version this will be published in!)
- why withAssetPrefix? We need a way to separate withPathPrefix (e.g. typically used for in-app links that contain the formerly "pathPrefix") vs. content that should have assetPrefix and pathPrefix merged (e.g. assets)

packages/gatsby-link/src/__tests__/index.js
```diff
@@ -6,10 +6,10 @@ import {
   createHistory,
   LocationProvider,
 } from "@reach/router"
-import Link, { navigate, push, replace, withPrefix } from "../"
+import Link, { navigate, push, replace, withPrefix, withAssetPrefix } from "../"
 
 afterEach(() => {
-  global.__PATH_PREFIX__ = ``
+  global.__BASE_PATH__ = ``
```
DSchau: __BASE_PATH__ is the new __PATH_PREFIX__. __PATH_PREFIX__ is now overloaded with assetPrefix + pathPrefix.

packages/gatsby-plugin-catch-links/src/__tests__/catch-links.js
```diff
@@ -365,13 +365,13 @@ describe(`pathPrefix is handled if catched link to ${pathPrefix}/article navigat
   })
 
   afterAll(() => {
-    global.__PATH_PREFIX__ = ``
+    global.__BASE_PATH__ = ``
     eventDestroyer()
```
DSchau: 🔥

packages/gatsby/src/joi-schemas/joi.js
```diff
@@ -2,24 +2,58 @@ const Joi = require(`joi`)
...
+    mapping: Joi.object(),
+    plugins: Joi.array(),
+    proxy: Joi.object().keys({
+      prefix: Joi.string().required(),
+      url: Joi.string().required(),
+    }),
+    developMiddleware: Joi.func(),
+  })
+  // throws when both assetPrefix and pathPrefix are defined
+  .when(
```
DSchau: This is actually pretty cool! Gives us some conditional validation for gatsby-config.js. In this case, this will throw when assetPrefix and pathPrefix are both defined and both relative.

DSchau:
For reviewers, I've gone ahead and:

1. Tested these _new_ versions of plugins with older versions of Gatsby (works just fine!)
1. Tested assetPrefix, pathPrefix, and assetPrefix + pathPrefix

So it _seems_ good to go--but this is touching a lot of surface area. I think it likely there may be some small minor edge-cases we'll need to get a patch release out for, but to the best of my understanding, this won't break any one's current workflow!

packages/gatsby/src/joi-schemas/joi.js
DSchau: relativeUrls for assetPrefix work just fine! Albeit with the gotcha that it can't be used in conjunction with a relative pathPrefix, as well!

docs/docs/path-prefix.md
```diff
@@ -2,46 +2,82 @@
...
 
-NOTE: When running the command without the `--prefix-paths` flag, Gatsby ignores
-your `pathPrefix`.
+If this flag is not passed, Gatsby will ignore your `pathPrefix` and build out your site as if it were hosted from the root domain.
+
+### In-app linking
+
+As a developer using this feature, it should be seamless. We provide APIs and libraries to make using this functionality a breeze. Specifically, the [`Link`](/docs/gatsby-link/) component has built-in functionality to handle path prefixing.
+
+For example, if we want to link to our `/page-2` link (but the actual link will be prefixed, e.g. `/blog/page-2`) we don't want to hard code this path prefix in all of our links. We have your back! Merely by using the `Link` component, we will automatically prefix your paths for you. If you later migrate off of `pathPrefix` your links will _still_ work seamlessly.
```
pieh: No huge deal, but should we avoid words like `merely`? (here and later in same docs). How about
```suggestion
For example, if we want to link to our `/page-2` link (but the actual link will be prefixed, e.g. `/blog/page-2`) we don't want to hard code this path prefix in all of our links. We have your back! By using the `Link` component, we will automatically prefix your paths for you. If you later migrate off of `pathPrefix` your links will _still_ work seamlessly.
```

docs/docs/asset-prefix.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Adding to `gatsby-config.js`
+
+```js:title=gatsby-config.js
+module.exports = {
+  assetPrefix: `https://cdn.example.com`,
+}
+```
+
+That was easy! One more step - when we build out this application, we need to add a flag so that Gatsby picks up this option.
```
pieh: Let's avoid words like "easy" :)

packages/gatsby/src/utils/webpack.config.js
```diff
@@ -188,8 +184,10 @@ module.exports = async (
       // optimizations for React) and what the link prefix is (__PATH_PREFIX__).
       plugins.define({
         ...processEnv(stage, `development`),
-        __PATH_PREFIX__: JSON.stringify(
-          program.prefixPaths ? store.getState().config.pathPrefix : ``
+        __BASE_PATH__: JSON.stringify(program.prefixPaths ? pathPrefix : ``),
```
m-allanson: Worth adding a note to explain why `base_path = pathPrefix` and `path_prefix = publicPath`? Pulling from your [earlier comment](https://github.com/gatsbyjs/gatsby/pull/12128/files#r279947544), something like:

>BASE_PATH is the new PATH_PREFIX. PATH_PREFIX is now overloaded with assetPrefix + pathPrefix. More info at  https://github.com/gatsbyjs/gatsby/pull/12128

packages/gatsby/src/utils/webpack.config.js
```diff
@@ -188,8 +184,10 @@ module.exports = async (
...
+        __PATH_PREFIX__: JSON.stringify(program.prefixPaths ? publicPath : ``),
+        __ASSET_PREFIX__: JSON.stringify(
+          program.prefixPaths ? assetPrefix : ``
```
pieh: I'm not sure where to document this, but I feel like it would be useful to define differences between those. To my understanding plugins (for now) should only ever use `__PATH_PREFIX__`? `__ASSET_PREFIX__` seems not used ATM anywhere - would it make sense not to define it at all?

docs/docs/asset-prefix.md
```diff
@@ -0,0 +1,96 @@
...
+```json:title=package.json
+{
+  "scripts": {
+    "build": "gatsby build --prefix-paths",
+    "postbuild": "aws s3 sync public s3://mybucket"
+  }
+}
+```
+
+Now whenever the `build` script is invoked, e.g. `npm run build`, the `postbuild` script will be invoked _after_ the build completes, therefore making our assets available on a _separate_ domain after we have finished building out our appliacation with prefixed assets.
```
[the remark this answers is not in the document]
DSchau: heh - thank you!

packages/gatsby/src/utils/webpack.config.js
DSchau: Correct! `__BASE_PATH__` should be used for what was `__PATH_PREFIX__`, and `__ASSET_PREFIX__` isn't really used. I'll look for a best place to document this!

packages/gatsby/src/utils/get-public-path.js
```diff
@@ -0,0 +1,21 @@
+const trimSlashes = part => part.replace(/(^\/)|(\/$)/g, ``)
+
+const isURL = possibleUrl =>
+  [`http`, `//`].some(expr => possibleUrl.startsWith(expr))
```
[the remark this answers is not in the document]
DSchau: What is this fixing though? It seems like `http` would also catch `https`?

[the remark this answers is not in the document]
muescha: Just with `http` there is also a `httpfolder/` a valid match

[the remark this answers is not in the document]
DSchau: @muescha do you have a practical example of what that exactly means?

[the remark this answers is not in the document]
DSchau: In other words--when would a user run into that? Or are you saying, in theory, this is too leniant?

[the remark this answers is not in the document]
muescha: Yes. There some edge cases where the function returns wrong result. If the function means isUrl then it should be an Url ;)

`http//domain` is true
`httpfolder` is true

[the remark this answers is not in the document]
muescha: Also as I see it right from the code:

It never hit the `//` check

Because a `//something` is always stripped with `stripSlashes` to `/something`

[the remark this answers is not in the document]
DSchau: https://github.com/gatsbyjs/gatsby/blob/5521302fb42edb88a445016aed4445a0c423a56a/packages/gatsby/src/utils/__tests__/get-public-path.js#L58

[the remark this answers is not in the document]
DSchau: But yeah - agreed, we can make this more correct easily.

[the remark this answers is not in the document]
muescha: Ok. `handles double slashes` that's magic ;)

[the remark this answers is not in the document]
muescha: Ok - I see - because the `//` check is never `true` in `isUrl` for this it added later the `/` to the `/something` again to be again an `//something`

DSchau:
_Almost done_ with the e2e tests.

One test is failing for an unknown reason, so I need to fix/figure out why!

```
 1) Production pathPrefix navigation can go back:
     CypressError: cy.then() timed out after waiting '9999ms'.

Your callback function returned a promise which never resolved.

The callback function was:

win => {
    if (!win.___apiHandler) {
      win.___apiHandler = _apiHandler.default.bind(win);
    }

    return _apiHandler.waitForAPI.call(win, api).then(() => subject);
  }
```

DSchau:
Alrighty--should have finished up the last, outstanding bug and test failure! Will get this merged and shipped... today!

DSchau:
Published as gatsby@2.4.0. I bumped the _minor_ version for all packages in this PR, as well.

- gatsby@2.4.0
- gatsby-link@2.1.0 (gatsby@2.4.0 depends upon this, don't need to install manually!)
- gatsby-plugin-feed@2.2.0
- gatsby-plugin-manifest@2.1.0
- gatsby-plugin-offline@2.1.0
- gatsby-plugin-sitemap@2.1.0

Soon enough, the documentation will be available at [/docs/asset-prefix/](https://gatsbyjs.org/docs/asset-prefix/), but you can get a sneak peek [here](https://github.com/gatsbyjs/gatsby/tree/master/docs/docs)

Thanks everyone for all the help testing this and getting it 🚢'd! ❤️

andrewstaker:
Thanks @DSchau and all those who pitched in to make this happen. This feature has saved our team a lot of headaches. 🙌

leonfs:
Thanks @DSchau - Fantastic work.

vv13:
My version is gatsby@2.3.4, then I got an error:
```
  29 |
  30 | function withPrefix(path) {
> 31 |   return normalizePath(__BASE_PATH__ + "/" + path);
     | ^
  32 | }
  33 |
  34 | function withAssetPrefix(path) {


  WebpackError: ReferenceError: __BASE_PATH__ is not defined
```

It make me confused, but fixed in gatsby@2.4.0.

szimek:
@DSchau Huuuge thanks for releasing this feature!

I found one small issue when using relative path. I'm using `gatsby-source-contentful` and `gatsby-transformer-remark` plugins and all links with relative paths in markdown are prefixed with value of `assetPrefix`. I guess it's because of `--path-prefix` option, causing `gatsby-transformer-remark` to add `pathPrefix` to all relative links.

For now I fixed it with the following local plugin:
```js
const visit = require('unist-util-visit');

// pathPrefix has value of assetPrefix here
module.exports = ({ markdownAST, pathPrefix }) => {
  visit(markdownAST, 'link', (node) => {
    if (node.url.startsWith(pathPrefix)) {
      node.url = node.url.replace(pathPrefix, '');
    }
  });

  return markdownAST;
};
```

leonfs:
Hi @DSchau.

We moved to version 2.4.2, and we are getting the following exceptions on build:

```
Error: ./.cache/gatsby-browser-entry.js
  Module not found: Error: Can't resolve 'gatsby-link' in ...

Error: ./.cache/navigation.js
  Module not found: Error: Can't resolve 'gatsby-link' in ...

```

You've mentioned in https://github.com/gatsbyjs/gatsby/pull/12128#issuecomment-488796099 that the `gatsby-link` package should not be installed manually. But those 2 files are in the `.cache` folder and `node` will try to resolve the package at the root node_modules folder, it won't be able to infer that it's within the gatsby package in `node_modules`.

I can make it work if I install the package but I was just wondering if this has been an oversight due to the test repo using the resolutions feature, it could have slept through without noticing.

Is anyone else experiencing this problem?

Issue #12080, opened 2019-02-25
Feature request: use file hashes to download the same asset only once


baobabKoodaa:
## Summary

If a static file is downloaded and cached in the browser, it shouldn't be downloaded again. The browser should just serve it from cache. Currently this is not working as it should, because separate GraphQL queries generate separate sets of assets - even when the output files are exactly same. This means that the exact same file will be generated into different paths, so the browser has to download the same file from multiple paths.

### More information

I've put up a live demo at https://stoic-raman-037869.netlify.com/

When I go to the front page, my browser fetches [this image](https://stoic-raman-037869.netlify.com/static/be21c7dd0d11c74c18530fc39aefa922/4b90c/mailboxes.webp). When I click on the image and go to the article page, the browser again fetches the exact same image from a [different path](https://stoic-raman-037869.netlify.com/static/be21c7dd0d11c74c18530fc39aefa922/6a184/mailboxes.webp).

These are the exact same file:

```
/public/static/be21c7dd0d11c74c18530fc39aefa922/4b90c$ sha256sum mailboxes.webp 
cc5c80f43d8a4e21f2488f2356ecf9a3adaeeb1b722b19dbe4de6e4f4eb98cf4  mailboxes.webp
/public/static/be21c7dd0d11c74c18530fc39aefa922/6a184$ sha256sum mailboxes.webp 
cc5c80f43d8a4e21f2488f2356ecf9a3adaeeb1b722b19dbe4de6e4f4eb98cf4  mailboxes.webp

```

An easy way to fix this issue would be to place all generated static files into the same folder and make the file hash a part of the filename. This way the second generation of this image would simply overwrite the first generation. And any page which needs this image would reference the same file, allowing the browser to cache it efficiently and avoid downloading the same asset multiple times.

In addition: I'm all ears if anybody knows how I can make this work for me now without forking Gatsby?

freiksenet:
We generate file names in `gatsby-plugin-sharp`, it contains file internal digest (which is the same for your images) and sorted picture resize argument digest (which is different). Unless arguments that are used to process this image are different, then this is a bug somewhere. Could you link to the source of the blog so I can investigate more?

baobabKoodaa:
It looks like it's not possible for me to set the exact same arguments in both places.

Here is my `gatsby-config.js` for "how to transform `<re-img>` images inside markdown files":

```
resolve: `gatsby-remark-rehype-images`,
options: {
              tag: 're-img',
[...]

Issue #11837, opened 2019-02-17
gastby-plugin-sharp: constantly regenerating same images on subsequent query runs in dev mode


pieh:
## Description

We use `fs-exists-cached` (memoized `fs.lstat(sync)`) in `gatsby-plugin-sharp` to check if processed file already exists to skip processing. This works in production, but in development cause problem:

If file doesn't exist on initial check, it will cache this -> sharp plugin will generate image. On next query run that would cause sharp plugin processing we use cached check and regenerate same image.

Ideally we would invalidate that cached value for the file, but `fs-exists-cached` doesn't support that currently. We could fork it (it's pretty low weight - https://github.com/isaacs/fs-exists-cached/blob/master/index.js ) and add invalidation mechanism and use that once images are generated.

DSchau:
@pieh this is a reasonably straight-forward fix. Would you be opposed to just swapping to fs.existsSync here, or do you think the performance benefit is worth the cost of forking/maintaining an fs-exists-cached with an invalidation mechanism?

pieh:
Let's just switch. @oorestisime already kind of did that in https://github.com/gatsbyjs/gatsby/pull/12129 when he was working on another `gatsby-plugin-sharp` issue

oorestisime:
FYI I am planning on rebasing my PRs for sharp plugin on current master today or tomorrow. This should be taken care of in on of my PRs.

wardpeet:
@pieh I guess this is fixed?

pieh:
Yes, it should be fixed. Would be good to verify before closing, so let me do that

wardpeet:
Going to close this :)

PR #12129, opened 2019-02-27
fix(gatsby-plugin-sharp): strip non related args when hashing resizing args

oorestisime:
So continuing on the work to make gatsby-plugin-sharp friendly on third-party packages (https://github.com/oorestisime/gatsby-remark-rehype-images/issues/2) we now face this issue https://github.com/gatsbyjs/gatsby/issues/12080

TL;DR images generated from transformer-sharp or and other plugins are the same (same sha256) but with different hash (depends on the available options). The problem is that all the plugins leak their options into the hash generation and even ones that are unrelated to the resizing processing (base64, generateTraceSVG, traceSVG etc)

I've been discussing this with @pieh for the last 2-3 hours :D
Proposed solution was to sanitize the args based on a whitelist, which is implemented on the PR. Which leads to the first question:

- is it ok if after this release _all_ image hashes are changing? 

By doing so it can also potentially help  lazy image processing @wardpeet by getting rid of unrelated options when persisting jobs.

Anyway current state of the PR is that it kind of works, i ve tested this in websites using my plugin and other that not. The issue here is that there are some differences in the number of images created when this modification is on or not.

I also verified that public folder doesn't contain duplicate images:

`find public -regex ".*\.\(jpg\|webp\|png\|jpeg\)" -type f -exec sha256sum {} \; | cut -d ' ' -f 1 |sort| uniq -d | wc -l`

This needs to be a bit more tested (and fix update snapshots etc)


oorestisime:
### Updated metrics:

##### On https://github.com/baobabKoodaa/blog/tree/temp384
Without the plugin 438 images. 66 duplicate images 
With plugin 372 images no duplicate images.

##### On using remark
without the plugin 70 images. no duplicates
with plugin 62 images

I probably hadn't cleaned public before.

oorestisime:
Minor update:

Each run is not deterministic on the number of images. Probably related to #11837
OTOH i hashed each image produced in both runs (for using-remark) and in both times we have the exact same list of sha256 and a total of 54 images.

oorestisime:
Another update:

- total variations seems to be fixed if i replace `exists-sync-cached` with `fs.existsSync` validating suspicions of #11837
```
2 /home/grok/personal/gatsby/examples/using-remark/public/static/584f423221bf98fe367b9e4de67b8e15/75e2a/jay.jpg false
      2 /home/grok/personal/gatsby/examples/using-remark/public/static/d286acdd98800790ea1bbdddbc982962/084f1/daisy.jpg false
      2 /home/grok/personal/gatsby/examples/using-remark/public/static/d286acdd98800790ea1bbdddbc982962/25aa5/daisy.jpg false
      2 /home/grok/personal/gatsby/examples/using-remark/public/static/d286acdd98800790ea1bbdddbc982962/c814f/daisy.jpg false
      3 /home/grok/personal/gatsby/examples/using-remark/public/static/584f423221bf98fe367b9e4de67b8e15/084f1/jay.jpg false
      3 /home/grok/personal/gatsby/examples/using-remark/public/static/584f423221bf98fe367b9e4de67b8e15/25aa5/jay.jpg false
      3 /home/grok/personal/gatsby/examples/using-remark/public/static/584f423221bf98fe367b9e4de67b8e15/c814f/jay.jpg false

oorestisime:
I forced-push this PR rebasing to currnet master. i know jest tests will fail due to not updating snapshots. I am waiting for feedback and will then do it.

PR #12130, opened 2019-02-27
docs(starters): add styled-components starter

blakenoll:
## Description

Modified the default starter to use styled components allowing for a basic way to get up and running with Gatsby and styled components.


docs/starters.yml
```diff
@@ -1931,3 +1931,11 @@
     - Responsive Web Design
     - Auto generated Sidebar
     - Auto generated Anchor
+- url: https://gatsby-starter-styled-components.netlify.com/
+  repo: https://github.com/blakenoll/gatsby-starter-styled-components
+  description: The Gatsby default starter modified to use styled-components
+  tags: 
+    - styled-components
```
wardpeet: We put all these frameworks like emotion in the same bucket

```suggestion
    - Styling:CSS-in-JS
```

wardpeet:
You got a linting error 😄 

You can run `npx prettier --write docs/starters.yml` and then commit that file, that should resolve the linting error.

blakenoll:
@wardpeet Thanks for showing me the fix. I went in and ran prettier and also updated the tag to what you suggested above.

Issue #7484, opened 2018-08-20
[docs][guides] Docs page for REPL


kkemple:
A new `repl` command was added to Gatsby CLI but it is currently undocumented.

V1shvesh:
Hey @kkemple!
Can I be assigned to this issue?

marcysutton:
@V1shvesh sorry for the delay, yes we would love for you to take this one! I've added a stub for the REPL and some stuff in the CLI docs in a new PR, once it is merged you can start there: https://github.com/gatsbyjs/gatsby/pull/12132

V1shvesh:
Awesome! I believe it's merged now.

V1shvesh:
Hey @kkemple. Feeling a lil embarassed to admit that I don't have a clue where to start.

I found the [list of REPL commands](https://www.gatsbyjs.org/docs/gatsby-cli/#repl), but I am unable to understand the motivation for such a feature. Please help.

pieh:
@V1shvesh Original idea came from https://github.com/gatsbyjs/gatsby/issues/6952 and it was implemented in https://github.com/gatsbyjs/gatsby/issues/7484 . For the most part it is to allow examining internal state of Gatsby, without resorting to bunch of `console.log` or debug breakpoints while Gatsby is doing it's processing.

It might be useful during development/debugging of plugins, but it does require knowledge of Gatsby internals. Or using it can help people understand internals better (depending how you look at it).

giacomocerquone:
This is awesome, if someone of you needs help tell me, I'm available to work on the doc!

marcysutton:
@gatsbyjs/core is there anything we need to do to the Gatsby REPL to get it into shape for proper documentation? It is in the [Gatsby CLI doc](https://www.gatsbyjs.org/docs/gatsby-cli/#repl), but this issue is about filling in the [top-level docs page](https://www.gatsbyjs.org/docs/gatsby-repl/).

n8finch:
@marcysutton I'd love to give this a go! It might take me a week or two to get familiar with the commands and such to get some basic documentation together, but then we could go from there.

It seems like I can just break up the commands into different pages, and show examples of how and why you would use each, as well as how this could be combined with the GraphQL explorer to be useful to folks. 

I can start with three or four commands and go from there. Thoughts?

marcysutton:
[...]

PR #12132, opened 2019-02-27
Update to CLI docs, plus a REPL stub

marcysutton:
## Description

When working on some Gatsby materials, I found the CLI docs could be easier to follow by putting each command on its own line. This was especially clear for the REPL section, which didn't even appear to be a command to me. 😜 So I added more details, and a stub to fill out the rest of the docs.

## Related Issues

Related to https://github.com/gatsbyjs/gatsby/issues/7484, which when addressed will add a more complete doc on the REPL.

### DIFF
--- a/docs/docs/gatsby-cli.md
+++ b/docs/docs/gatsby-cli.md
@@ -12,11 +12,11 @@ _We provide similar documentation available with the gatsby-cli [README](https:/
 
 The Gatsby CLI (`gatsby-cli`) is packaged as an executable that can be used globally--in fact, this was previously how we recommended using the CLI.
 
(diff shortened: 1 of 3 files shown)


docs/docs/gatsby-cli.md
```diff
@@ -38,13 +38,14 @@ Run `npx gatsby --help` for full help.
 
 `npx gatsby new gatsby-site`
```
alexandrtovmach: I'm not sure, but if we going to show code on new line maybe better to use code blocks to highlight the syntax:
```
> ```sh
> npx gatsby --help
> ```
```

wardpeet: I like this approach 👍 especially when they are already on their own line. It might take more space than we actually want to.

m-allanson: Generally I agree, but in this case I don't think anything will get highlighted? Let's merge this in and then re-visit once the page is live.

PR #12133, opened 2019-02-27
fix(gatsbygram): update broken image

calcsam:
https://gatsbygram.gatsbyjs.org/ has a broken Avatar image due to Instagram changing permissions. This removes that image so the demo displays correctly


alexandrtovmach:
@calcsam Hello, thanks for your job. Youre removed component from layout, but left that import, could you please fix error:
```
8:8  error  'Avatar' is defined but never used  no-unused-vars
```

m-allanson:
Thanks Sam! 

Note that there's another missing avatar on the post detail page. Fixing that would be a good intro issue, so I'll write that up seperately.

PR #12139, opened 2019-02-27
feat(gatsby): Add normalization for pathPrefix, siteUrl, etc

### LINKED ISSUE #11871: [feat]: add validation or normalization for pathPrefix, siteUrl, etc.
## Summary

We should validate siteMetadata.siteUrl (as a url) and pathPrefix (as a string starting with a forward slash and not ending with one) and warn or normalize when a trailing slash is detected.

### Basic example

See #11869 for an example of this causing issues. In effect:

- Certain keys can not contain a trailing slash
- We should warn when these values _do_ contain a trailing slash, rather than documenting this behavior with something like an inline comment like [`// Note: it must *not* have a trailing slash.`](https://www.gatsbyjs.org/docs/path-prefix/#production-build)

We'll want to update our [schema validation](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/joi-schemas/joi.js#L3-L15) with either a warning or outright normalization of the expected structure. See [this example](https://runkit.com/dschau/5c6ae2aff8150400120f142c) for a way we could normalize--but note that there could be a cleaner way to do this.

### Motivation

We should warn early (or normalize if we _know_ a structure should be a certain way), rather than relying on the user to investigate underlying functionality and read comments.

tterb:
## Description

Updated the schema validation to include normalization of pathPrefix and siteUrl leading/trailing slashes.  
I wasn't able to get clarification on if the intent was to apply the same normalization to the proxy.url and proxy.prefix or any other values, but I can also add normalizations for those as well.  

Additionally, would you like me to remove `// Note: it must *not* have a trailing slash.` from the docs, since these changes will make it unecessary?  

## Related Issues

Fixes #11871


pieh:
We have this normalization code for `pathPrefix`:
https://github.com/gatsbyjs/gatsby/blob/1280a8b202eac32e5bd2d0790eea0ccde856e2ae/packages/gatsby/src/redux/reducers/config.js#L26-L35

This should handle making sure that pathPrefix is in correct shape - it doesn't work?

tterb:
@pieh You're right, that's a good catch.  
I had assumed that pathPrefix wasn't already being handled since it was a primary target within the issue, but I should've investigated further.  
I've updated the schema to remove the unnecessary logic.  
  
Nonetheless, I wasn't able to find any normalization for the proxy.url and proxy.prefix, so if you'd like I can also update the schema to include normalization for these keys and/or update the docs relating to the siteUrl.

tterb:
@pieh Do you have any insight on why the `e2e_tests_gatsbygram` would only fail after the prefixPath checks were removed when it hadn't previously?  

I've taken a look at the trace but it appears to lead to an issue with `gatsby-example-image-gallery`, which is unchanged.

tterb:
@pieh Do you have any insight on why `e2e_tests_gatsbygram` test would fail after removing the unnecessary pathPrefix checks when it had passed prior?
Looking at the errors it looks like it has to do with `gatsby-example-image-gallery`, which has remained unchanged and shouldn't be effected by the changes.

packages/gatsby/src/joi-schemas/joi.js
```diff
@@ -1,10 +1,17 @@
 const Joi = require(`joi`)
 
+const normalizeUrl = url => url.replace(/\/*$/, ``)
```
DSchau: Perhaps consider my method [here](https://github.com/gatsbyjs/gatsby/pull/12128/files#diff-bd517ec758aa05f3e6e1c34b15f65394R3)?

```suggestion
const stripTrailingSlash = chain => chain.replace(/(\w)\/+$/, `$1`)
```

I have a set of unit tests that validate some edge cases, as well (not necessary for this PR).

tterb:
Thanks @DSchau for the suggestion. I've committed your suggested changes and updated the schema calls.

Should `// Note: it must *not* have a trailing slash.` be removed from [the docs](https://www.gatsbyjs.org/docs/path-prefix/#production-build) as a part of this pull request to keep everything up-to-date?

tterb:
@DSchau That's a good call, I hadn't thought of it that way.
It's probably best to leave the note for now and maybe revisit it after the next major release.  
Thanks for all your help!

Issue #11798, opened 2019-02-15
[feat]: warn when processing overly large images with gatsby-plugin-sharp


DSchau:
## Summary

Images tend to be a key determinant in build-times growing exponentially large. In particular, overly large images processed in memory tend to put undue and unnecessary strain on the build process.

If these images are resized and optimized _before_ processing, there are drastic improvements in build time and build performance (e.g. sometimes even resolving out of memory issues).

With this in mind--we should minimally do a few things:

1. Document _how_ to resize overly large images and why doing so is valuable (tracked in https://github.com/gatsbyjs/gatsby/issues/24793)
1. Augment `gatsby-plugin-sharp` to display a warning when overly large images are encountered

UPDATE: This issue now covers only item # 2, warning from the CLI. It has been filed internally as product feedback but we'll keep the issue open to also track it in the open.

### Basic example

We should **first** create a docs page (that can be linked to from gatsby-plugin-sharp warning message) that conveys the _why_ of optimizing images before processing.

Once created, we should then augment `gatsby-plugin-sharp` with the ability to warn on "overly large" images, where overly large images can be defined as: (note: this is fluid, needs some more research)

- Image dimensions larger than a certain threshold, e.g. 2000 pixels wide
- Image overall file size larger than a certain threshold, e.g. 2MB images
  - Note: this is probably even more important than the image width

When an image is processed that meets one or both of the above conditions, we should warn and link to the docs page that will be created as part of this issue. One note: if the requested image processing is itself large (e.g. requesting a 4000 pixel-wide hero image like below) we **should** short circuit and _not_ warn in those cases.

```graphql
{
  file(relativePath: { regex: "/i-am-overly-large-intentionally.jpg/" }) {
    childImageSharp {
      fluid(maxWidth: 4000) {
        ...GatsbyImageSharpFluid
      }
    }
  }
}
```

### Motivation

Consider an example from an awesome community member @oorestisime. He's working on a Gatsby blog, and has a large number of (large) images, some 3000 pixels wide and several megabytes large.

These images are resized using Gatsby's excellent image-processing capabilities with a max-width of 1000 pixels. As such, optimizing the images to be closer to the maximum size of images makes a ton of sense and the quality difference is negligible.

When building in CircleCI _without_ optimization and the overly large images, the build actually [times out](https://circleci.com/gh/oorestisime/oasome/7).

[...]

PR #12144, opened 2019-02-27
Warn users when sharp is used for large images

oorestisime:
Contributes: #11798

This ia draft because:

- [ ] i am not sure how to generate shortened link
- [ ]  Should we update the documentation to talk about the thresholds and how to configure them?
- [ ] I think it is a little hacky to export the warn function but at the same time it enables me to unit-test this easily. Or maybe I should put it in a different file ?
- [ ] i need to test this on using-remark or similar to see if the output is too much or ..

### DIFF
--- a/packages/gatsby-plugin-sharp/src/__tests__/__snapshots__/index.js.snap
+++ b/packages/gatsby-plugin-sharp/src/__tests__/__snapshots__/index.js.snap
@@ -74,6 +74,26 @@ Object {
 }
 `;
 
+exports[`gatsby-plugin-sharp optimization warnings warns when dimension exceeds 1`] = `
+Array [
+  Array [
+    "
+Image test exceeds the warning threshold for image optimization. Read more on image optimization link!
+",
+  ],
+]
+`;
+
+exports[`gatsby-plugin-sharp optimization warnings warns when size exceeds 1`] = `
+Array [
+  Array [
+    "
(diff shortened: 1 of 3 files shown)


pieh:
> * I think it is a little hacky to export the warn function but at the same time it enables me to unit-test this easily. Or maybe I should put it in a different file ?

Normally it would be fine to export, but given that those exports are user facing, then yeah probably moving it to some `utils` (or something like that) file would be good idea (to not export internals to users).

> [ ] i need to test this on using-remark or similar to see if the output is too much or ..

What we could do is track if we already displayed message for image and not show it again for same source file - maybe also add some tracking that if there are over 10 images that cause this warning to stop showing this in console? and say "X more"?

LekoArts:
Hey @oorestisime and @pieh 👋 Any suggestions on what to do with this PR? If you don't have the time at the moment we could close this or edit this PR to a warning in the docs first?

oorestisime:
hey @LekoArts thanks for pinging me here! I somehow forgot this! I ll try and work on this sometime next week!

freiksenet:
Thank you very much for working on this @oorestisime! 

As this PR hasn't had much activity lately, we are going to close for now, as part of our regular stale PR cleanup. However, we feel this PR is valuable for Gatsby, so feel free to reopen it when you have time to update it.

Thank you very much for your contribution! :purple_heart: :muscle:

PR #12039, opened 2019-02-24
chore(docs): Revert "fix(docs): Update "gatsby" commands to "npm run""

DSchau:
Reverts gatsbyjs/gatsby#12025

PR #12150, opened 2019-02-27
update README - use npm script instead of CLI

fkanout:
## Description

In the **_Get Up and Running in 5 Minutes_** README section, in the 2nd step we should use `npm run develop` instead of `gatsby develop` (getting error message `command not found` which is normal).
Because in the previous step, we used `npx`, assuming that gatsby is not installed globally on the machine.

## Related Issues
None


wardpeet:
This seems like a duplicate of #12039. We're reverting back to global installs so give us a few days to get it all sorted out.

Issue #10245, opened 2018-12-03
Implementing yarn PnP


Ikana:
In order to implement [yarn plug and play](https://github.com/yarnpkg/rfcs/pull/101) with Gatsby some tasks need to be completed first

- [x] Explicitly define any dependencies and not relying  on transitive dependencies to fill in the gaps
- [x] Updating relay-compiler to a compliant with the PnP spec
- [ ] Combing thought Gatsby’s dependencies and making them spec compliant as well 
- [ ] adding a flag on gatsby-cli to trigger the PnP flow
- [ ] refactoring the gatsby-cli to work with out node_modules

Even if the task is practically unattainable (specially from the third step) the other steps are intrinsically worth it.

DSchau:
So glad you opened this! I have the bulk of the functionality built out in gatsbyjs/gatsby#9897 so we can iterate further on any core changes required there _once_ the issues you've identified are solved 😅

Ikana:
No problem @DSchau !
When I was testing this originally, updating relay compiler from 1.5.0 to 1.7.0, showed some test failing but not any more 😝, but I would like to discuss the ramifications of updating that package.

My theory is that specifying the transitive dependencies fixed it.

vjpr:
@DSchau Did you run into any issues relating to the way Gatsby copies all files in `cache-dir` to `.cache`, as this approach relies on hoisting.

I just got Gatsby working with pnpm which requires dependencies to be strictly defined here https://github.com/pnpm/pnpm/issues/991#issuecomment-446330885. Just wondering if you had a similar issue.

vjpr:
Just added an entry to webpack `resolve.modules` config as mentioned in comment link above. No changes to babel.

arcanis:
Hey! I'm playing a bit with Gatsby and I think I could help with this issue (I'm also Yarn's maintainer and PnP's author, which helps)! 🙂 

So far the main issue I have is related to the `.cache` folder: the files seem properly copied from the packages into the cache folder, but since the cache folder isn't part of the dependency tree PnP has no way to know which dependencies it should be able to require.

I'll try to figure out a way to solve this, but the problem is that it's not really a PnP bug: even under the regular install strategy there's nothing that guarantees that the dependencies that will be required by those files will be the one that you expect - hoisting makes it kinda work, but under some circumstances might end up requiring the wrong versions.

DSchau:
@arcanis wow - well this would be amazing :)

[...]

PR #12161, opened 2019-02-27
chore: bump fs-extra to 7.0.0

arcanis:
## Description

This change bumps `fs-extra` to `7.0.0` (which was already used in the repo elsewhere, like `gatsby-plugin-sharp`). It would probably be a good idea to bump (and synchronize) all the `fs-extra`, but one fire at a time :))

## Related Issues

#10245

### DIFF
--- a/packages/gatsby-transformer-sharp/package.json
+++ b/packages/gatsby-transformer-sharp/package.json
@@ -9,7 +9,7 @@
   "dependencies": {
     "@babel/runtime": "^7.0.0",
     "bluebird": "^3.5.0",
-    "fs-extra": "^7.0.1",
+    "fs-extra": "^7.0.0",
     "potrace": "^2.1.1",
     "probe-image-size": "^4.0.0",
     "sharp": "^0.21.3"
--- a/packages/gatsby-transformer-sharp/package.json
+++ b/packages/gatsby-transformer-sharp/package.json
@@ -9,7 +9,7 @@
   "dependencies": {
     "@babel/runtime": "^7.0.0",
     "bluebird": "^3.5.0",
-    "fs-extra": "^4.0.2",
+    "fs-extra": "^7.0.1",
     "potrace": "^2.1.1",
     "probe-image-size": "^4.0.0",
     "sharp": "^0.21.3"


wardpeet:
could you look for any breaking changes between 4 & 7? it's a 3 version jump so we want to make sure that we are compatible.

arcanis:
There are some changes in the `copy` mechanism but it seems mostly bugfixes in this case:

https://github.com/jprichardson/node-fs-extra/blob/master/CHANGELOG.md#700--2018-07-16

The `gatsby-transformer-sharp` package doesn't make much use of the advanced capabilities of `fs-extra` (it only uses the simplest form of `copy` and `existsSync`), so changes to options and such don't matter in this case.

arcanis:
The main change would be "Drop support for Node.js versions 4, 5, & 7", but since it's already being used by other packages I guess you already have a 8+ requirement.

DSchau:
@arcanis that could be a _little_ early to make that change! Unfortunately, while Node 8 is our minimum recommended, we have until this point avoided making breaking changes to Node 6 capability in the _current version_ of Gatsby (v2).

We can certainly make this change, but generally - is there a benefit to making this change that we need to do? Otherwise, I'd be inclined to perhaps unify on fs-extra@4.x.x for v2, and then we can bump to 7.0.0 for Gatsby v3 when we'd like to drop Node 6.

@pieh @wardpeet what are your thoughts, here?

arcanis:
> We can certainly make this change, but generally - is there a benefit to making this change that we need to do?

In my case the problem is that 4.x's `fsx.copy` is currently using `fs.open` in order to copy the files from a location to another. It causes issues when used together with the zip layer, because zipped files cannot be abstracted away at the moment (there's some subtle issues to implement transparent support, because `open` is meant to return a file descriptor).

Starting from the 5.x `fsx.copy` now uses `fs.copyFile` when available, which completely solves the problem since it's easy to support it transparently.

After looking in more details, maybe I don't need 7.x right now - I would have to double check but maybe the 5.x would be fine given what I mentioned, and would preserve the compatibility with Node 6.

DSchau:
Also @arcanis [`gatsby-source-filesystem`](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-sharp/package.json#L14) already uses this, so we may have already unintentionally broken Node 6 support with that package 🙃

arcanis:
Ahah yeah that's why I originally picked `^7.0.0` since it was used elsewhere 😄 What do you want me to do? Use `^5.0.0` or keep it at `^7.0.0`?

pieh:
Let's test this with `yarn` resolutions forcing fs-extra to 7, as @arcanis mentioned we use really simple `copy` for single files and `existsSync` from it

pieh:
`fs-extra` still supports node 6 ... just not node 7 (wat): https://github.com/jprichardson/node-fs-extra/blob/master/package.json#L6

arcanis:
Eh, I completely missed that 6 wasn't in the list of "Drop support for Node.js versions 4, 5, & 7". Confusing 😅

pieh:
With test case that uses all `fs-extra` calls from `gatsby-transformer-sharp`
```
➜  pr12161 git:(master) ✗ node -v
v6.14.4
➜  pr12161 git:(master) ✗ yarn why fs-extra
yarn why v1.10.1
warning ../package.json: No license field
[1/4] 🤔  Why do we have the module "fs-extra"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "fs-extra@7.0.1"
info Reasons this module exists
   - "gatsby-transformer-sharp" depends on it
   - Hoisted from "gatsby-transformer-sharp#fs-extra"
   - Hoisted from "gatsby-source-filesystem#fs-extra"
   - Hoisted from "gatsby#fs-extra"
   - Hoisted from "gatsby-plugin-sharp#fs-extra"
   - Hoisted from "gatsby#gatsby-cli#fs-extra"
   - Hoisted from "gatsby-plugin-offline#workbox-build#fs-extra"
info Disk size without dependencies: "264KB"
info Disk size with unique dependencies: "352KB"
info Disk size with transitive dependencies: "352KB"
info Number of shared dependencies: 3
```
It works on node 6 properly, so I'll go ahead and merge this

pieh:
Published `gatsby-transformer-sharp@2.1.15`

PR #12162, opened 2019-02-27
chore(gatsby): adds missing prop-types dependency

arcanis:
## Description

The `gatsby` package is missing `prop-types` in its dependencies (used by `cache-dir/default-html.js`).

## Related Issues

#10245

### DIFF
--- a/packages/gatsby/package.json
+++ b/packages/gatsby/package.json
@@ -101,6 +101,7 @@
     "physical-cpu-count": "^2.0.0",
     "postcss-flexbugs-fixes": "^3.0.0",
     "postcss-loader": "^2.1.3",
+    "prop-types": "^15.6.1",
     "raw-loader": "^0.5.1",
     "react-dev-utils": "^4.2.1",
     "react-error-overlay": "^3.0.0",


wardpeet:
Weird that we never had a bug report about this. I don't think it does any harm on bundle size but I'm going to check for sure

DSchau:
@wardpeet it's usually included at the starter level, e.g. https://github.com/gatsbyjs/gatsby-starter-default/blob/master/package.json#L16

Agreed - though, this probably should have been included! I can't imagine it'd have an impact on bundle size, but glad you're checking it out to ensure!

arcanis:
It doesn't add anything, a lot of other packages were depending on `prop-types` as well (such as `react`, `react-hot-loader`, `gatsby-images`, `react-helmet`, ...). Package managers will just merge them together.

@DSchau Speaking of which, what exactly is meant to contain `.cache`? Is it a strict copy of `gatsby/cache-dir`? If so why not directly use this directory instead of copying it? Is there any file generated in the cache that could import a package that's not a dependency of `gatsby`?

sidharthachatterjee:
Published in gatsby@2.1.20

PR #12171, opened 2019-02-28
chore(docs): "behaviour" to "behavior"

bryndyment:
docs update: `behaviour` -> `behavior`

## Related Issues

(none)

### DIFF
--- a/docs/docs/environment-variables.md
+++ b/docs/docs/environment-variables.md
@@ -4,7 +4,7 @@ title: "Environment Variables"
 
 ## Environments and Environment Variables
 
-You can provide environment variables to your site to customise its behaviour in different environments.
+You can provide environment variables to your site to customise its behavior in different environments.
 
 Note that we need to distinguish in this discussion between variables which have been defined in
 special places in order to be used in different deployment environments, and true OS-level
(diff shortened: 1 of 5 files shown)


wardpeet:
Hey! I don't think we're going to accept this one as there is nothing wrong with `behaviour`. (correct me if I'm wrong but this is what google said)

> Behavior is the preferred spelling in American English. Behaviour is preferred everywhere else. Other than the spelling, there is no difference between the two words.

https://www.grammar.com/behavior_vs._behaviour

yogeshkotadiya:
> Hey! I don't think we're going to accept this one as there is nothing wrong with `behaviour`. (correct me if I'm wrong but this is what google said)
> 
> > Behavior is the preferred spelling in American English. Behaviour is preferred everywhere else. Other than the spelling, there is no difference between the two words.
> 
> [grammar.com/behavior_vs._behaviour](https://www.grammar.com/behavior_vs._behaviour)

There is none, its like `color` vs `colour`.

bryndyment:
True, they are both correct depending upon your country.  We spell it `behaviour` in Canada, for instance.

There are 59 instances of `behavior` on the site and 5 of `behaviour`, and it's an American company, so I think it makes sense to standardize on the American spelling.

What do you think?

wardpeet:
I think it's going to be hard to standardize on this one and keep an 👁 on following releases.

I'm not going to make the call so pinging @gatsbyjs/docs

Issue #12014, opened 2019-02-23
[gatsby-link] TypeError useRef and Link component


ollelauribostrom:
## Description
I'm trying to use the new useRef hook to create a reference to a Link component. This results in an error saying `TypeError: this.props.innerRef is not a function`.

### Steps to reproduce
```js
  const linkRef = useRef(null);
  return (
    <Link
      ref={linkRef}
      to="https://github.com/gatsbyjs/gatsby-starter-hello-world"
    >
      Hello world!
    </Link>
  );
```
Full example project [here](https://github.com/ollelauribostrom/gatsby-link-useref). Same exact result if passing the `innerRef` props instead.

### Expected result
I expect the mutable ref object to reference the rendered `<a>`-tag

### Actual result
Error thrown from [here](https://github.com/gatsbyjs/gatsby/blob/fd12550dae58becdc52057d8b43f1eb6960a6883/packages/gatsby-link/src/index.js#L71).
```
Uncaught TypeError: this.props.innerRef is not a function
    at GatsbyLink.handleRef (index.js:117)
    at GatsbyLink.handleRef (react-hot-loader.development.js:652)
    at commitAttachRef (react-dom.development.js:17291)
    at commitAllLifeCycles (react-dom.development.js:18537)
    at HTMLUnknownElement.callCallback (react-dom.development.js:149)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:199)
    at invokeGuardedCallback (react-dom.development.js:256)
    at commitRoot (react-dom.development.js:18744)
    at react-dom.development.js:20214
    at Object.unstable_runWithPriority (scheduler.development.js:255)
    at completeRoot (react-dom.development.js:20213)
    at performWorkOnRoot (react-dom.development.js:20142)
    at performWork (react-dom.development.js:20050)
    at performSyncWork (react-dom.development.js:20024)
    at requestWork (react-dom.development.js:19893)
    at scheduleWork (react-dom.development.js:19707)
    at scheduleRootUpdate (react-dom.development.js:20368)
    at updateContainerAtExpirationTime (react-dom.development.js:20396)
    at updateContainer (react-dom.development.js:20453)
    at ReactRoot../node_modules/react-dom/cjs/react-dom.development.js.ReactRoot.render (react-dom.development.js:20749)
    at react-dom.development.js:20886
    at unbatchedUpdates (react-dom.development.js:20255)
    at legacyRenderSubtreeIntoContainer (react-dom.development.js:20882)
[...]

PR #12174, opened 2019-02-28
feat(gatsby-link): support RefObject

ollelauribostrom:
This PR adds support for passing a RefObject as a ref to the Link component. In other words, now something like this should be supported (previously resulted in a TypeError):

```js
  const linkRef = useRef(null);
  return (
    <Link
      ref={linkRef}
      to="https://github.com/gatsbyjs/gatsby-starter-hello-world"
    >
      Hello world!
    </Link>
  );
```

## Related Issues
#12014

### DIFF
--- a/packages/gatsby-link/src/index.js
+++ b/packages/gatsby-link/src/index.js
@@ -69,7 +69,9 @@ class GatsbyLink extends React.Component {
   }
 
   handleRef(ref) {
-    if (this.props.innerRef) {
+    if (this.props.innerRef && this.props.innerRef.hasOwnProperty(`current`)) {
+      this.props.innerRef.current = ref
+    } else if (this.props.innerRef) {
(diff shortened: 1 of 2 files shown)


packages/gatsby-link/src/index.js
```diff
@@ -161,7 +163,10 @@ class GatsbyLink extends React.Component {
 
 GatsbyLink.propTypes = {
   ...NavLinkPropTypes,
-  innerRef: PropTypes.func,
+  innerRef: PropTypes.oneOfType([
```
wardpeet: we could remove this proptype as it shouldn't be mandatory.

ollelauribostrom: Thanks for having a look, removed it! 🙂

sidharthachatterjee:
Published in gatsby-link@2.0.14

PR #12177, opened 2019-02-28
test(babel-preset-gatsby-package): add test for production mode

ollelauribostrom:
## Description
Adding a test case for `babel-preset-gatsby-package` to make sure that the correct presets and plugins are used in production.

### DIFF
--- a/packages/babel-preset-gatsby-package/__tests__/index.js
+++ b/packages/babel-preset-gatsby-package/__tests__/index.js
@@ -148,3 +148,51 @@ it(`Specifies proper presets and plugins in debug browser mode`, () => {
     ),
   ])
 })
+
+it(`Specifies proper presets and plugins in production browser mode`, () => {
+  const currentBabelEnv = process.env.BABEL_ENV
+  let presets
+  let plugins
+
+  try {
+    process.env.BABEL_ENV = `production`
+    const config = preset(null, { browser: true })
+    presets = config.presets
+    plugins = config.plugins
+  } finally {
+    process.env.BABEL_ENV = currentBabelEnv
+  }
+
+  expect(presets).toEqual([
+    [
+      expect.stringContaining(path.join(`@babel`, `preset-env`)),
+      {
+        debug: false,
+        loose: true,
+        modules: `commonjs`,
+        shippedProposals: true,
+        targets: {
+          browsers: [`last 4 versions`, `safari >= 7`, `ie >= 9`],
+        },
+        useBuiltIns: false,
+      },
+    ],
+    [
+      expect.stringContaining(path.join(`@babel`, `preset-react`)),
+      { development: false },
+    ],
+    expect.stringContaining(path.join(`@babel`, `preset-flow`)),
+  ])
+  expect(plugins).toEqual([
+    expect.stringContaining(
+      path.join(`@babel`, `plugin-proposal-class-properties`)
+    ),
+    expect.stringContaining(
+      path.join(`@babel`, `plugin-proposal-optional-chaining`)
+    ),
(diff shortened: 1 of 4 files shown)


packages/babel-preset-gatsby-package/__tests__/index.js
```diff
@@ -148,3 +148,51 @@ it(`Specifies proper presets and plugins in debug browser mode`, () => {
...
   ])
 })
+
+it(`Specifies proper presets and plugins in production browser mode`, () => {
+  const currentBabelEnv = process.env.BABEL_ENV
+  let presets
+  let plugins
+
+  try {
+    process.env.BABEL_ENV = `production`
```
DSchau: Interesting way to write this!

Why would this ever fail?

packages/babel-preset-gatsby-package/__tests__/index.js
```diff
@@ -148,3 +148,51 @@ it(`Specifies proper presets and plugins in debug browser mode`, () => {
...
+        useBuiltIns: false,
+      },
+    ],
+    [
+      expect.stringContaining(path.join(`@babel`, `preset-react`)),
+      { development: false },
+    ],
+    expect.stringContaining(path.join(`@babel`, `preset-flow`)),
+  ])
+  expect(plugins).toEqual([
```
DSchau: Let’s make this a separate test, eg 

1) sets correct plugins
2) sets correct presets

ollelauribostrom:
@DSchau Did some updates to this. By moving `const r = m => require.resolve(m)` to a separate file we can mock the resolving in our tests and use snapshots instead of `expect.stringContaining`. Should make stuff a bit more readable. 🙂

packages/babel-preset-gatsby-package/__tests__/index.js
ollelauribostrom: I dont think setting the BABEL_ENV could actually fail (moved this out of the try-finally now). The try-finally itself is just a guard so that if the `preset()` call would throw, we dont end up running subsequent tests with `BABEL_ENV = 'production'` 🙂

DSchau: Ah - yeah, so I'd recommend just doing something like:

```
let env
beforeEach(() => {
  env = process.env.BABEL_ENV
  process.env.BABEL_ENV = 'production'
})

afterAll(() => {
  process.env.BABEL_ENV = env
})
```

sidharthachatterjee:
Thank you, @ollelauribostrom 👍 

Do you want make the quick changes that @DSchau recommended?

ollelauribostrom:
@sidharthachatterjee Sorry didn't notice @DSchau last comment. Just a quick thought, wouldn't this: 
```js
let env
beforeEach(() => {
  env = process.env.BABEL_ENV
  process.env.BABEL_ENV = 'production'
})

afterAll(() => {
  process.env.BABEL_ENV = env
})
```  
mean that the BABEL_ENV would always be set to 'production' after running the tests. It would look like this: 
1. We start out with BABEL_ENV = 'test'
2. Before the first test we set env = 'test', BABEL_ENV = 'production'
3. Before the second test we set env = 'production', BABEL_ENV = 'production'
4. We reset BABEL_ENV = env, but now env is set to 'production'

It would also mean that we would run all test cases in production mode, which I don't think we want? We could extract the tests for production mode to its own describe block, and do 
```js
describe('production', () => {
  let env;
  beforeAll(() => {
    env = process.env.BABEL_ENV
    process.env.BABEL_ENV = 'production'
  })
  //..some test cases
  afterAll(() => {
    process.env.BABEL_ENV = env
  })
})
``` 

What are your thoughts?

PR #12178, opened 2019-02-28
Add ga for main CTAs on gatsbyjs.org

jamo:
Add ga instrumentation for the main CTAs on the gatsbyjs.org

### DIFF
--- a/www/src/components/button.js
+++ b/www/src/components/button.js
@@ -19,6 +19,7 @@ const Button = ({
   tiny,
   secondary,
   ondark,
+  tracking,
   ...rest
 }) => {
   const Tag = components[tag || `link`]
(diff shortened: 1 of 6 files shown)


wardpeet:
@jamo does it make sense to make https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-google-analytics/src/index.js more flexible so you don't have to "copy" so much code.

www/src/components/button.js
```diff
@@ -41,8 +42,42 @@ const Button = ({
...
+    if (props.target && props.target.toLowerCase() !== `_self`) {
+      redirect = false
+    }
+
+    if (tracking && window.ga) {
+      window.ga(`send`, `event`, {
+        eventCategory: `Outbound Link`,
+        eventAction: `click`,
+        eventLabel: `${tracking} - ${props.to || props.href}`,
+        transport: redirect ? `beacon` : ``,
```
wardpeet: why wouldn't you use `beacon` for everything?

jamo: As per https://developers.google.com/analytics/devguides/collection/analyticsjs/sending-hits and https://caniuse.com/#feat=beacon I'm in the understanding that it's currently best to let the lib pick the transport - except for the case when we are redirecring and a plain old `fetch` is expected not to work.

jamo:
We can't use the `OutboundLink` as it's a `<a>` alternative. But we need the same behavior for a couple of `<Link>`s.
That said, I'm not sure if there is still a better way around it.

www/src/components/button.js
wardpeet: > By default, analytics.js picks the HTTP method and transport mechanism with which to optimally send hits. The three options are 'image' (using an Image object), 'xhr' (using an XMLHttpRequest object), or 'beacon' using the new navigator.sendBeacon method.

sweet! thought if you set it to beacon GA would use beacon and fallback but it's not super clear so 👍

wardpeet:
that's what I meant with flexible like choose your own Tag :D for example

`<OutboundLink as="a" tracking="whatever" />`
`<OutboundLink as={Link} tracking="whatever" />`
`<OutboundLink as={Button} tracking="whatever" />`

jamo:
I see. Refactoring the OutboundLink might require more refactorings on gastbyjs.org side as well. I'll let someone more with more react and gatsbyjs.org knowledge decide :)

PR #12180, opened 2019-02-28
feat(gatsby-source-drupal): added Request Headers and GET Request Params options

aposudevsky:
## Description

Adds possibility to customize request by adding Request Headers or GET params to url.


pieh:
This seems reasonable. I think that in the future we will want to allow some control over params (i.e. different params for some endpoints), but this would be a matter of making `params` accept functions.

DSchau:
Successfully published: `gatsby-source-drupal`@`3.1.0`

Thanks!

PR #12185, opened 2019-02-28
fix(www): add gatsby-remark-embed-video to www/package.json

janosh:
- enables markdown video embedding from sources such as YouTube, Vimeo, etc.
- fixes non-functional embed at the very end of the [Adding search with algolia](https://www.gatsbyjs.org/docs/adding-search-with-algolia)


janosh:
@DSchau Oops, good catch. Just added that.

Issue #12154, opened 2019-02-27
Command name in help output is wrong.


ronaldborman:
## Description

When I run 'npx gatsby --help' (as suggested by https://www.gatsbyjs.org/tutorial/part-zero/) the help text displays the wrong command name:
> Usage: gatsby.js <command> [options]
> 
> Commands:
>   gatsby.js develop...
> 

When I then install the cli globally the help text output (with or without the use of npx) changes to:
>  Usage: index.js <command> [options]
> 
> Commands:
>   index.js develop...
> 

### Steps to reproduce

run 'npx gatsby --help'
run 'npm install -g gatsby-cli'
run 'gatsby --help'

### Expected result

The help text should contain the command name 'gatsby' (without the extension .js).

### Actual result

The help text contains the command name 'gatsby.js' (when npx is used) or 'index.js' (after the global installation of gatsby-cli).

### Environment

Node 8.9.4, npm 5.10.0, gatsby 2.4.11.

yogeshkotadiya:
I can't reproduce this, working as expected.
Same as #5838
I'll take a look into this.

yogeshkotadiya:
Merged in #12186

PR #12186, opened 2019-02-28
fix(gatsby-cli):  Fixed incorrect scriptname in gatsby-cli

yogeshkotadiya:
## Description
Resolves incorrect script name in output when using `gatsby --help` command.
Also removed `yargs` configuration from package.json because of this warning
`Configuring yargs through package.json is deprecated and will be removed in the next major release, please use the JS API instead.`

### TO-DO

- [x] Resolve any Breaking changes due to a semver major upgrade.

## Packages upgraded
yargs 11.1.0 --> ~~13.2.1~~ --> 12.0.5
## Related Issues
Addresses #12154

### DIFF
--- a/packages/gatsby-cli/package.json
+++ b/packages/gatsby-cli/package.json
@@ -51,5 +51,8 @@
     "build": "babel src --out-dir lib --ignore **/__tests__",
     "prepare": "cross-env NODE_ENV=production npm run build",
     "watch": "babel -w src --out-dir lib --ignore **/__tests__"
+  },
+  "yargs": {
+    "boolean-negation": false
   }
 }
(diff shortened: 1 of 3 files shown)


packages/gatsby-cli/package.json
```diff
@@ -28,7 +28,7 @@
     "source-map": "^0.5.7",
     "stack-trace": "^0.0.10",
     "update-notifier": "^2.3.0",
-    "yargs": "^11.1.0",
+    "yargs": "12.0.5",
```
sidharthachatterjee: Since this is also a major semver upgrade, are there any breaking changes we should resolve? Also, should we add a ^ as opposed to pinning the version?

sidharthachatterjee:
```suggestion
    "yargs": "^11.1.0",
```
Would probably be better to not change the version at all in this PR, what do you think?

sidharthachatterjee: Just saw that scriptName is 12+

sidharthachatterjee:
Thank you so much, @yogeshkotadiya 👍

sidharthachatterjee:
Published in gatsby-cli@2.4.14

PR #12190, opened 2019-02-28
[schema] Add schema query tests, inference fixes

stefanprobst:
This is for #11480 

Adds:
* more query tests
* test for merging inferred field types
* some test todos
* added JSdoc

Fixes:
* ensure schema composer has Node interface
* make `noDefaultResolvers: true` mean `addDefaultResolvers: false`
* rebuild with `SitePage` `context` now works correctly. A more general solution for on-demand schema rebuild would still require a bit more work.
* merging nested types in inference: because we now guarantee priority when we have conflicting field names, we have to deal with temporary TypeComposers on nested fields. Also ensure that we honor array depth; and get the nested TC's name from an existing TC, not from createSelector.
* make sorting work (lowercase sort order enum) -- this has triggered another Loki mismatch (it is the same issue with `null` values on indexed fields: the kitchen sink query sorts DESC on `likes`, which sorts `nulls` to the front, and those screw things up. It is good that we know this, and we should find a solution, but it's an edge case. The easiest "fix" for now would be to adjust sort order in kitchen sink query from DESC to ASC and update the snapshot. EDIT: I now did just that.

### DIFF
--- a/examples/using-type-definitions/gatsby-config.js
+++ b/examples/using-type-definitions/gatsby-config.js
@@ -22,14 +22,5 @@ module.exports = {
         path: `./src/data/`,
       },
     },
-    {
-      resolve: `gatsby-source-graphql`,
-      options: {
-        fieldName: `cms`,
-        url: `https://api-euwest.graphcms.com/v1/cjjr1at6d0xb801c3scjrm0l0/master`,
-        typeName: `GraphCMS`,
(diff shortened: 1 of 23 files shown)


freiksenet:
LGTM, do you want to add TODO tests on this branch or separately? Anyway, you can merge it if you want.

PR #12192, opened 2019-02-28
test(gatsby-plugin-typescript): add tests for ts plugin

ollelauribostrom:
## Description
Adding tests for `gatsby-plugin-typescript`. Test cases verify that:
- The correct resolvable extensions are exposed
- onCreateBabelConfig sets the correct babel config
- onCreateWebpackConfig sets the correct webpack config


sidharthachatterjee:
Published in gatsby-plugin-typescript@2.0.10

PR #12193, opened 2019-02-28
test(gatsby-plugin-twitter): add tests for twitter plugin

ollelauribostrom:
## Description
Adding tests for `gatsby-plugin-twitter`.

#### Summary of changes
- Adding test cases to verifiy that onRouteUpdate injects the twitter script when it's necessary.
- Minor refactor to make the injected script easier to read + deleting the unused `location` parameter.

### DIFF
--- a/packages/gatsby-plugin-twitter/src/__tests__/gatsby-browser.js
+++ b/packages/gatsby-plugin-twitter/src/__tests__/gatsby-browser.js
@@ -1,0 +1,28 @@
+const { onRouteUpdate } = require(`../gatsby-browser`)
+
+describe(`gatsby-plugin-twitter`, () => {
+  describe(`onRouteUpdate`, () => {
+    it(`injects the twitter script if it hasn't already been injected`, () => {
+      const querySelector = jest.spyOn(document, `querySelector`)
+      const createElement = jest.spyOn(document, `createElement`)
+      const getElementsByTagName = jest.spyOn(document, `getElementsByTagName`)
+      const element = {}
+      querySelector.mockReturnValue(true)
+      createElement.mockReturnValueOnce(element)
+      getElementsByTagName.mockReturnValue([{ appendChild: jest.fn() }])
+      global.twttr = { widgets: { load: jest.fn() } }
+      onRouteUpdate()
+      onRouteUpdate()
+      expect(element).toMatchSnapshot()
+      expect(createElement).toHaveBeenCalledTimes(1)
+      expect(global.twttr.widgets.load).toHaveBeenCalledTimes(2)
+    })
+    it(`only injects the twitter script if there is an embedded tweet`, () => {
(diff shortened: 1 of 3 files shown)


packages/gatsby-plugin-twitter/src/__tests__/gatsby-browser.js
```diff
@@ -0,0 +1,28 @@
+const { onRouteUpdate } = require(`../gatsby-browser`)
+
+describe(`gatsby-plugin-twitter`, () => {
+  describe(`onRouteUpdate`, () => {
+    it(`injects the twitter script if it hasn't already been injected`, () => {
```
DSchau: One note here, not really a blocker, but I probably would've written two tests here.

```
it(`adds the twitter script`, () => {})
it(`only adds the twitter script once`, () => {})
```

just to keep the test a little more encapsulated.

ollelauribostrom: Thanks for looking at this, will keep that in mind 🙂

sidharthachatterjee:
Published in gatsby-plugin-twitter@2.0.10

jsanchez034:
@ollelauribostrom @sidharthachatterjee  the newly formatted `window.twttr` code is missing semi-colons causing twitter embeds no to render..
```
    window.twttr = (function(d, s, id) {
      var js,
        fjs = d.getElementsByTagName(s)[0],
        t = window.twttr || {}
      if (d.getElementById(id)) return t
      js = d.createElement(s)
      js.id = id
      js.src = "https://platform.twitter.com/widgets.js"
      fjs.parentNode.insertBefore(js, fjs)
      t._e = []
      t.ready = function(f) {
        t._e.push(f)
      }
      return t
    })(document, "script", "twitter-wjs")
```
I don't have everything setup to run Gatsby locally and create a PR. I could create a PR with the fix later today but I figured I would notifiy you guys because you might be able to create a PR before I do.

DSchau:
Yikes - yeah. This is outside of Gatsby, but [this CodeSandbox](https://codesandbox.io/s/xojkonnvrq) illustrates the issue/fix.

If we change `innerText` to `innerHTML` that should do the trick.

jsanchez034:
Ah I see but shouldn't the embedded javascript code have proper javascript syntax with semi-colons as well?

detikpw:
For temporary fix in my development I add some semicolon same as @jsanchez034 seggestion, but I check the script that injected add tag <br> in each line, then I try to change `innerText`  => `innerHTML` it works without add semicolon, and there's no more <br> tags in each line :+1:

ollelauribostrom:
Sorry about breaking this, I just formatted the injected script in line with the rest of the gatsby code base without thinking too much about it. Added back the semi-colors in #12296

jsanchez034:
Awesome, thanks for the quick fix @ollelauribostrom! 😃

PR #12200, opened 2019-03-01
fix(gatsby-remark-images): override all default styling with wrapperStyle

florinpop17:
Addressing issue [12199](gatsbyjs#12199) where the default values couldn't be overwritten.


sidharthachatterjee:
Published in gatsby-remark-images@3.0.9

PR #12211, opened 2019-03-01
test(gatsby-plugin-preact): add tests for preact plugin

ollelauribostrom:
## Description
Adding tests for `gatsby-plugin-preact`. Test cases verify that:
- onCreateWebpackConfig sets the correct webpack config


packages/gatsby-plugin-preact/src/__tests__/gatsby-node.js
```diff
@@ -0,0 +1,11 @@
+const { onCreateWebpackConfig } = require(`../gatsby-node`)
+
+describe(`gatsby-plugin-preact`, () => {
+  it(`sets the correct webpack config`, () => {
```
DSchau: Let's split these out.

```js
it(`invokes setWebpackConfig with correct config when not develop-html`, () => {
  const setWebpackConfig = jest.fn()
  onCreateWebpackConfig({ stage: `develop`, actions: { setWebpackConfig } })

  expect(setWebpackConfig).toHaveBeenCalledTimes(1)
  // I like to be more explicit here, but up to you!
  expect(setWebpackConfig).toHaveBeenCalledWith({
    resolve: {
      alias: {
        create-react-class: '// etc',
      }
    }
  })
})

it(`does not invoke setWebpackConfig when stage is develop-html`, () => {
  const setWebpackConfig = jest.fn()
  onCreateWebpackConfig({ stage: `develop-html`, actions: { setWebpackConfig } })

  expect(setWebpackConfig).not.toHaveBeenCalled()
})
```

ollelauribostrom: Thanks for having a look, I've now split these out  🙂

PR #12212, opened 2019-03-01
test(gatsby-plugin-nprogress): add tests for nprogress plugin

ollelauribostrom:
## Description
Adding tests for `gatsby-plugin-nprogress`. 

#### Summary of changes
- Verifies that onClientEntry calls NProgress.configure with the correct styles
- Verifies that onRouteUpdateDelayed calls NProgress.start
- Verifies that onRouteUpdate calls NProgress.done
- Minor refactor, `export const` instead of `exports.` to keep this consistent throughout the package

### DIFF
--- a/packages/gatsby-plugin-nprogress/src/__tests__/gatsby-browser.js
+++ b/packages/gatsby-plugin-nprogress/src/__tests__/gatsby-browser.js
@@ -1,11 +1,11 @@
+jest.mock(`nprogress`)
+
 import NProgress from "nprogress"
 import {
   onClientEntry,
   onRouteUpdateDelayed,
   onRouteUpdate,
 } from "../gatsby-browser"
-
-jest.mock(`nprogress`)
 
 describe(`gatsby-plugin-nprogress`, () => {
   describe(`onClientEntry`, () => {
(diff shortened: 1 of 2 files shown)


packages/gatsby-plugin-nprogress/src/__tests__/gatsby-browser.js
```diff
@@ -0,0 +1,40 @@
+import NProgress from "nprogress"
+import {
+  onClientEntry,
+  onRouteUpdateDelayed,
+  onRouteUpdate,
+} from "../gatsby-browser"
+
+jest.mock(`nprogress`)
```
DSchau: Can we hoist this to the top (I think jest does this anyways, but it's usually clearer for the future "reader" that this module is mocked if the mocks are first)

In other words, this should be the first line of this file.

packages/gatsby-plugin-nprogress/src/__tests__/gatsby-browser.js
```diff
@@ -0,0 +1,40 @@
...
+  describe(`onClientEntry`, () => {
+    it(`calls NProgress.configure with the correct styles`, () => {
+      const element = {}
+      const createElement = jest.spyOn(document, `createElement`)
+      const appendChild = jest.spyOn(document.head, `appendChild`)
+      createElement.mockReturnValue(element)
+      appendChild.mockReturnValue(true)
+      NProgress.configure = jest.fn()
+      onClientEntry(null, { showSpinner: false })
+      expect(NProgress.configure.mock.calls).toMatchSnapshot()
```
DSchau: Let's tweak this. I think it's probably better to just assert a few things separately, e.g.

```js
expect(NProgress.configure).toHaveBeenCalledWith({}) // defined options here, or expect.any(Object) if you just want to assert it was called
```

```
expect(element.id).toBe(`nprogress-styles `)
expect(element.innerHTML).toMatchSnapshot()
```

ollelauribostrom:
@DSchau Thanks for reviewing this, I've done some updates according to your comments.

PR #12232, opened 2019-03-01
docs(sites): Add sherpalo.com to showcase.

laradevitt:
Updated docs/sites.yml to include sherpalo.com in the showcase. Thanks!


laradevitt:
Hmm. Looks like there is something odd with the DNS (you must be seeing the old site). This wasn't the case a few hours ago. Better put this on hold while I inquire with the provider. Sorry about that!

wardpeet:
just ping us when back available. A good choice it to use something like https://tools.pingdom.com/ to test dns resolutions

laradevitt:
@wardpeet Will do, thanks (we use dig). DNS records were switched over successfully last week but there was a change in providers over the weekend and something happened there. Hopefully will be resolved soon!

sidharthachatterjee:
Seeing the new one now. Looks great! Thank you for adding it, @laradevitt 👍

PR #12236, opened 2019-03-02
Move diagram above card and change explanatory text

KyleAMathews:
We couldn't fit the newly written text at the top pending a redesign of
the masthead but since we also wanted to move the diagram to the top, we
replaced its subheader with our new one.

screenshot 2019-03-01 18 16 36


KyleAMathews:
The main downside to this is that the diagram looks a bit funny with the white space on either side on larger screens — but this seems good enough to get in for now so merging!

PR #12243, opened 2019-03-02
fix(core): Only delete html/css files during prod builds

KyleAMathews:

### DIFF
--- a/packages/gatsby/src/bootstrap/index.js
+++ b/packages/gatsby/src/bootstrap/index.js
@@ -119,22 +119,24 @@ module.exports = async (args: BootstrapArgs) => {
   await apiRunnerNode(`onPreInit`, { parentSpan: activity.span })
   activity.end()
 
-  // Delete html and css files from the public directory as we don't want
+  // During builds, delete html and css files from the public directory as we don't want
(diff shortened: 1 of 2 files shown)


wardpeet:
Couldn't this enable more edge cases on develop? Like ghost pages that should have been deleted but aren't because invalidation didn't kick in. It's not a bug with this part of the code base but more a general issue we're having. Xstate refactoring will fix most if not all bugs we're seeing on this front.

Not against merging just double checking ^^

pieh:
@wardpeet has very good point here - we fallback to serving `index.html` in express app running in develop, but first we do serve existing static assets, so we can serve stale html pages with this on first render:

screenshot 2019-03-04 at 23 17 32

This is solvable. Easiest option would be to blacklist `.html` files in `express.static`, but it doesn't expose this option easily

KyleAMathews:
oh hmmm... yeah great point @wardpeet. We used to have code to ignore .html files — but that was removed at some point. It looks like we can just set `index` to false so I'll add that.

KyleAMathews:
Pushed up commit that seems to fix things with testing locally.

PR #12245, opened 2019-03-02
chore(docs): migrate back to global installs

### LINKED ISSUE #12195: Revert instructing people to use npx and npm scripts

Things to revert:

- [ ] GitHub issue templates
- [ ] getting started doc
- [ ] recipes
- [ ] tutorial

ryancharris:
## Description
This PR reverts all references to `npx` in the codebase that occur outside of the `docs/tutorial/` subdirectory.

A follow up PR from @DSchau will address the references in the tutorial docs.

## Related Issues
- Fixes #12195 
- Reverts some of the work done as part of [#11556](https://github.com/gatsbyjs/gatsby/pull/11556)

### DIFF
--- a/docs/docs/gatsby-cli.md
+++ b/docs/docs/gatsby-cli.md
@@ -28,6 +28,8 @@ Now we have the `develop` script available to be used which will use our package
 
 ## How to use
 
+The Gatsby CLI is available via [npm](https://www.npmjs.com/) and should be installed globally by running `npm install -g gatsby-cli` to use it locally.
+
 Run `gatsby --help` for full help.
 
 ### `new`
--- a/docs/docs/gatsby-cli.md
+++ b/docs/docs/gatsby-cli.md
@@ -28,15 +28,11 @@ Now we have the `develop` script available to be used which will use our package
 
 ## How to use
 
-By using [npx](https://www.npmjs.com/package/npx) you can utilize the Gatsby command line tool without installing it globally. This lets you scaffold a project with just one command!
-
-For example you can run `npx gatsby new your-site-name` to generate a new Gatsby site.
-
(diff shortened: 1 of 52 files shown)


docs/docs/gatsby-cli.md
```diff
@@ -28,15 +28,11 @@ Now we have the `develop` script available to be used which will use our package
 
 ## How to use
 
-By using [npx](https://www.npmjs.com/package/npx) you can utilize the Gatsby command line tool without installing it globally. This lets you scaffold a project with just one command!
-
-For example you can run `npx gatsby new your-site-name` to generate a new Gatsby site.
-
-Run `npx gatsby --help` for full help.
+Run `gatsby --help` for full help.
```
DSchau: This should reference globally installing it first and foremost!

packages/gatsby-cli/README.md
```diff
@@ -23,15 +23,13 @@ We can now use the `develop` script tied to our app's version of Gatsby, rather
 
 ## How to use
 
-Rather than installing the Gatsby CLI tool locally you can run it using [npx](https://www.npmjs.com/package/npx)
-
-To see the commands available to run `npx gatsby --help`.
+Run `gatsby --help` for full help.
```
DSchau: Same here, e.g. reference that it needs to be globally installed and _how_ to globally install.

docs/docs/gatsby-cli.md
ryancharris: @DSchau You got it. Will update 🙌

packages/gatsby-cli/README.md
ryancharris: Same here

ryancharris:
@DSchau Hit your notes in 3163c74

PR #12251, opened 2019-03-03
Added Beautiful Buttons to Readme

KartSriv:


lachlanjc:
I like the idea here, but I don't think these images fit with the rest of the README—the typeface is different, and the buttons need manual sizing for higher-resolution displays.

KartSriv:
I can use the same typeface you use and render these buttons in SVG. Can I get some information about Color Code, Typeface, Size Preference, and other changes recommended? Thanks for advice! - Karthik Srivijay

wardpeet:
I love the idea as well but it takes a lot of real estate on the screen. Unsure how many people actually need this as http://www.gatsbyjs.org is pretty straight forward to get to the docs.

sidharthachatterjee:
@fk What do you think of this?

sidharthachatterjee:
Also just want to ensure this is accessible, paging @marcysutton

KartSriv:
@sidharthachatterjee @wardpeet @lachlanjc I can make it better by fixing the issues stated by @lachlanjc  but I need to know whether if you like this idea. I don't wanna lose my effort for a closed pull request. Thanks, Everyone! - Karthik Srivijay

KartSriv:
Thanks to everyone for the compliment and it would be nice if someone could provide Gatsby's Logo in .svg so that I can render the buttons in true SVG from photoshop. This, in turn, will give full resolution and no distortion no matter what the end user's Desktop Resolution is. Thanks!

yogeshkotadiya:
@KartSriv [Gatsby Logo svg](https://raw.githubusercontent.com/gatsbyjs/gatsby/master/www/src/monogram.svg)

KartSriv:
Thanks, @yogeshkotadiya! Just waiting for everyone confirmation.

LekoArts:
I don't think that this is a good idea. First priority for a README is to be accessible, not to be pretty at all cost. "Form follows function".

The README is embedded on multiple sites (e.g. [npm](https://www.npmjs.com/package/gatsby)) and they all have a different layout. Such buttons will just disturb the layout and you'll be greeted with a lot of buttons at first glance. And if you make them smaller they have the same size as the text, so imo you then should use text.

I think this doesn't add value to the README sadly, sorry.

Kosai106:
And just a side note here: The images are not all the same size.

KartSriv:
@LekoArts ` if you make them smaller they have the same size as the text` - I can make the buttons smaller.  `if we want to add another item: Where would we get the file to create that?` - I can give you the .psd file for new buttons. `The README is embedded on multiple sites (e.g. npm) and they all have a different layout. Such buttons will just disturb the layout` - It kinda makes sense but we can still use pure HTML+CSS buttons so what about that?

sidharthachatterjee:
@KartSriv Thanks so much for the PR!

Unfortunately, we're not going to merge this at this time. As stated above we want to make sure our README is 100% accessible and showing correctly on all platforms.

PR #12253, opened 2019-03-03
fix(www): Edge case layout issue, consistent design for plugin detail header links

fk:
Follows up on #12126.

Consolidates link styles for „Edit this page on GitHub“, „View plugin on GitHub“, and „See starters using this [plugin]“:

* add `linkStyles` to `src/utils/styles.js`, and use them for the aforementioned links (I really don’t want to do this, but instead create a component … let’s move incrementally and shiiiip ;-))
* reducing font-size and slightly updating the link copy from „See starters that use this“ removed the need for hiding „View plugin on GitHub“ on mobile (tested on 320px wide viewport/~iPhone 5)
* also resolves the edge case issue shown in https://github.com/gatsbyjs/gatsby/pull/12126#pullrequestreview-209893262

image

### DIFF
--- a/www/src/components/markdown-page-footer.js
+++ b/www/src/components/markdown-page-footer.js
@@ -2,8 +2,8 @@ import React from "react"
 import { graphql } from "gatsby"
 import EditIcon from "react-icons/lib/md/create"
 
-import { rhythm, scale } from "../utils/typography"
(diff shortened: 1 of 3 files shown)


fk:
Sorry in advance for self-approving this. :-/

PR #12262, opened 2019-03-03
fix(gatsby): Run onPostBootstrap only once after bootstrap

stefanprobst:
I'm not 100% sure what the intended behavior is, but currently we run `onPostBootstrap` every time the job queue is emptied. I *think* we only want to run `onPostBootstrap` once after bootstrap, so this PR unregisters the event handler.


KyleAMathews:
Yikes yeah. Only one call is necessary 😅 I guess not very many people use this hook.

sidharthachatterjee:
Published in gatsby@2.1.20

PR #2588, opened 2017-10-22
Add warning when node-creating plugins create no nodes

djm:
#### If applied this PR will..

Solve #2212. 

1) Add a warning when node-creating plugins create no nodes.
2) Introduce a way to store "plugin" <-> "Gatsby APIs implemented" data in the store. 

#### Why is this change is being made?

The warning serves two purposes: a) to notify those that are using the plugin that it hasn't created any nodes and can likely be removed, and b) to reduce the confusion caused by a GraphQL query for a specific node erroring as no nodes have been created yet. See #2212 for more.

As part of this work, two key changes have been made to the redux store:

a) Each plugin now has a `nodeAPIs` key whose value retains an array of strings denoting the names of which Gatsby APIs the plugin implements.

b) At the root of the store, a new key exists named `apiToPlugins`has been created which stores a simple key -> val map from the name of a Gatsby API, to an array containing the string names of all plugins which implement that API.

a) is helpful when needing to know what APIs a given plugin implements and b) is helpful when code requires a quick look up of which plugins implement a specific API - we pay the cost once at creation time rather than iterating through the plugin objects whenever we need to know.

#### Testing

The plugin -> implemented APIs store changes are covered by the existing load-plugins tests and therefore I've updated the snapshots (after checking they were correct). However, the implemented API -> plugin storage in the store is not covered with tests yet - I'm happy to add that if you can offer some guidance on how you'd like to test it? None of the other `store.dispatch` from load-plugins.js are tested, and I think with the way they are now they would require a mock store, which I'm not sure about introducing before you say yes :) Let me know.

Thanks Kyle!

PR #12263, opened 2019-03-03
fix(gatsby): Only use one redux namespace for plugins

stefanprobst:
Currently we have three redux namespaces for plugins (`flattenedPlugins`, `plugins`, `apiToPlugins`). This PR reduces this to just `flattenedPlugins`.
Also: use `flattenedPlugins.nodeAPIs` in `api-runner-node` and `page-hot-reloader`.


wardpeet:
this is cool! Thanks for the cleanup, I still need to test this out. Removing these namespaces does this mean it's a breaking change? I haven't looked really at it.

stefanprobst:
I don't think this was public API (?), so should be good.
`apiToPlugins` was introduced in #2588; and `plugins` is used once in `page-hot-reloader`. Both can easily just use `flattenedPlugins`.

stefanprobst:
Looks like the one usage in `page-hot-reloader` also got removed recently in #12671

stefanprobst:
@wardpeet Have you had a chance to look at this? Thanks!

stefanprobst:
Anything missing from this?

wardpeet:
doing it tomorrow!

wardpeet:
Tomorrow took a while but i've tested it and read through the code and it looks 👌 

Thanks!

stefanprobst:
Cool thanks!

sidharthachatterjee:
Published in gatsby@2.3.5

PR #12273, opened 2019-03-04
Remove npm/npx commands - Tutorial

### LINKED ISSUE #12195: Revert instructing people to use npx and npm scripts

JABedford:
## Description

Removal of `npm` and `npx` commands from the tutorial. 

## Related Issues

Fixes #12195


JABedford:
Agreed. Will that do the trick?

PR #12277, opened 2019-03-04
chore(gatsby-cypress): make our cypress config public

wardpeet:
## Description
- [x] rename cypress-gatsby => gatsby-cypress to keep it in sync with other packages #branding 
- [x] create Redirect for cypress-gatsby as somehow it's indexed by google
- [x] make gatsby-cypress public

follow up is to actually change relative imports into package imports


www/gatsby-node.js
```diff
@@ -540,6 +540,13 @@ exports.createPages = ({ graphql, actions, reporter }) => {
         }
       })
 
+      // redirecting cypress-gatsby => gatsby-cypress
+      createRedirect({
```
wardpeet: when searching gatsby cypress you get https://www.gatsbyjs.org/packages/cypress-gatsby/ so I've made a redirect as well.

PR #12292, opened 2019-03-04
chore(gatsby-plugin-google-analytics): export class to avoid `declare module` error

matthewrfindley:
## Description

Running `tsc` against a project using `gatsby-plugin-google-analytics` throws an error.

``` bash
node_modules/gatsby-plugin-google-analytics/index.d.ts:7:1 - error TS1046: A 'declare' modifier is required for a top level declaration in a .d.ts file.
```

Exporting the class or wrapping in a module would fix the issue.


matthewrfindley:
I have re-run `tsc` against the changes with no errors.

DSchau:
Published as `gatsby-plugin-google-analytics@2.0.16`

PR #12293, opened 2019-03-04
fix(docs): Fix broken import for GatsbyLink

johno:
The import shouldn't alias GatsbyLink to Link since it's being used directly as GatsbyLink in the example and conflicts with the local component name.


johno:
I had this backwards and then checked and my code exploded. My bad 🤦‍♂️

Issue #9911, opened 2018-11-13
Inconsistent CSS Styling Differs Between gatsby develop and build - Why?


systemlevel:
develop:
<img width="282" alt="images" src="https://user-images.githubusercontent.com/894983/48446821-90a98580-e757-11e8-9c47-f1fbd0c3a2c0.png">

build:
<img width="265" alt="images-1" src="https://user-images.githubusercontent.com/894983/48446838-9ef7a180-e757-11e8-9ab8-1379c9e17054.png">

I don't even know where to start on this.  I've cleared the .cache in the project folder to no avail.  

Also, it's interesting to note that some styling was, in fact, maintained during the build.  But other styling associated with the default fonts were not.

My bigger question is why does this occur? It would seem like the build should be the same as develop?

Thank you,

lundgren2:
I've had similar behaviour before. Have you tried to both remove .cache and node_modules, the lockfile and run npm install again? 

Otherwise, what plugins are you using? are you running the latest version of Gatsby? You can also use e.g. Chrome DevTools to inspect the elements and see what css-classes used both under the "Styles" and "Computed" tabs.

lundgren2:
Hi! I think it worked after I removed .cache and node_modules, but it was a long time ago.

I see you're using gatsby-plugin-offline, try to disable it and have a look at #5734

kakadiadarpan:
@systemlevel can you check if the CSS applied in `develop` and `build` is in a different order? I think it might be related to #9733

systemlevel:
@kakadiadarpan I can confirm that I am experiencing the same thing.  The CSS styles are loading in different order.

<img width="399" alt="screen shot 2018-11-14 at 11 57 18 am" src="https://user-images.githubusercontent.com/894983/48505865-48e13780-e805-11e8-9904-d36e123e6359.png">

<img width="400" alt="screen shot 2018-11-14 at 11 58 58 am" src="https://user-images.githubusercontent.com/894983/48505879-4d0d5500-e805-11e8-8b87-ac88a0f4fa17.png">

You'll notice a slightly difference CSS order between develop and build.

I'm using:

`import styled from 'styled-components';`

and I'm importing my CSS in the following way:

```
import 'bootstrap/dist/css/bootstrap.css';
import 'antd/dist/antd.css';
```

It could be similar to how #9733 is importing?

It sounds like there is a possible merge request to try and resolve this issue.  If that is the case is there an update on that and is it known if that resolved the issue?

[...]

Issue #6302, opened 2018-07-05
[v2] CSS (via Sass) inlined in different locations between develop and build


jbolda:
## Description

Using `typography.js` and `gatsby-plugin-sass` (and associated libraries of each), the css is "blobbed" after `typography` in develop and inlined before `typography` in build. This means the look of the website changes from develop to build. Images below (obviously more drastic with more css being moved around).

develop:
![develop](https://user-images.githubusercontent.com/2019387/42297744-35fe17d8-7fc7-11e8-98b8-0e8ab2e0da3c.PNG)

build:
![build](https://user-images.githubusercontent.com/2019387/42297750-3b2942fa-7fc7-11e8-8dda-a1da67fed82e.PNG)

### Steps to reproduce

[Super basic repro available here.](https://github.com/jbolda/gatsby-v2-inline-css-bug)

### Expected result

The look website end result in v1 matched what I am seeing in develop in v2 so I would expect that `typography` would be listed in the `<head>` before the inlined css.

### Actual result

The opposite of expected.

### Environment
Current environment below, but tested with both yarn and npm on Windows locally. I noticed the issue after a Netlify deploy preview (so Linux and Yarn).

```
  System:
    OS: Windows 10
    CPU: x64 Intel(R) Core(TM) i7 CPU         930  @ 2.80GHz
  Binaries:
    Yarn: 1.7.0 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.0.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 42.17134.1.0
  npmPackages:
    gatsby: next => 2.0.0-beta.17
    gatsby-plugin-sass: next => 2.0.0-beta.3
    gatsby-plugin-typography: next => 2.2.0-beta.2
```

### File contents (if changed)

See repro repo. Essentially `my-hello-world` with both plugins added.

octalmage:
I’m experiencing the same issue on my site. Ultimately what I had to do to resolve it was to create a custom html.js and move the Typography.js styles to the top. 

I’m not sure how this could be resolved in Gatsby core. Should Gatsby core manage the sort order of head elements? Should their be a priority for setHeadCompoments? It’s kind of an ugly problem.

octalmage:
There's probably a better way to handle this, I just wanted it fixed. Here's the commit where I fixed the issue: 

[...]

Issue #1994, opened 2017-09-01
CSS Ordering in Gatsby


brsanthu:
I have been trying to setup gatsby build with bootstrap but not getting the things as expected. Here is what I'm looking to achieve.

1. Load bootstrap
2. Load open sans font using typeface-open-sans module
3. Load application's global styles
4. Load layout's styles
5. Load component styles

Tried putting import in html.js, didn't work
Tried putting import in layouts/index.jsx, didn't work

I get the exact order opposite of what I'm looking for (that is loading comps styles, layout styles, fontface and then finally bootstrap).

Any help is much appreciated.

KyleAMathews:
A reproduction site would be nice!

hew:
Using sass? I've noticed that webpack will inject style imports into the head in seemingly random orders. Since Sass mostly depends on the order (resets, etc), it's problematic. I have not yet found a workaround. I'll update if I do.

brsanthu:
Yes, we are using Saas.

hew:
Yeah check out [this issue](https://github.com/webpack/webpack/issues/215). According to Sokra, the order is random during development, which isn't good.

hew:
To be fair, this seems to be not a Gatsby-specific issue. More a webpack one.

speedDeveloper:
I am having the same problem, any suggestions yet?

304NotModified:
Yes, please create a small sample application to reproduce the problem ;)

speedDeveloper:
Thank you @304NotModified, the repo is here https://github.com/speedDeveloper/philippczernitzki.me
When you start the gatsby blog, you see that font-awesome overrides styled-components styles.
For example the icons below the banner should have an increased font-size. It's not working because font-awesome is loaded after styled-components stylesheet.

![image](https://user-images.githubusercontent.com/10871302/33949711-b7fff1d0-e02a-11e7-9e72-52478e746f66.png)

KyleAMathews:
You need to modify the default order of loading CSS — the default is CSS is loaded *after* css-in-js. Copy `.cache/default-html.js` to `src/html.js` and then switch the order of these two lines https://github.com/gatsbyjs/gatsby/blob/99e3249438f4bf23267d68c538a42c3a1aba04b5/packages/gatsby/cache-dir/default-html.js#L32-L33

speedDeveloper:
Hey thank you very much. But sadly it doesn't work for my project. Font-Awesome still overrides the css of styled-components.
[...]

PR #12295, opened 2019-03-05
fix(gatsby-plugin-typography): sort typography style tag first to avoid overwriting issues

KyleAMathews:
Fixes https://github.com/gatsbyjs/gatsby/pull/12295
Fixes https://github.com/gatsbyjs/gatsby/issues/6302
Fixes https://github.com/gatsbyjs/gatsby/issues/9911
Fixes https://github.com/gatsbyjs/gatsby/issues/1994


DSchau:
Looks good to me--added some unit tests to validate behavior.

Merging and publishing!

screen shot 2019-03-04 at 7 13 35 pm

DSchau:
Successfully published `gatsby-plugin-typography@2.2.8`

KyleAMathews:
🎉

octalmage:
woohoo, glad to see this make it in! Thanks!

jbolda:
I am still seeing some things inlined differently. I can reproduce it in the [examples on the theme](https://github.com/jbolda/gatsby-theme-bulma) I have thrown together. Particularly noticeable with the Links being light green vs blue on the /articles/ page.

Krizzu:
I've come across the same issue today.
My issue was that my `CSS` files were added to `Head` before `typography.js` injection, hence few styles were overridden. 

So I've fixed it by overriding `injectStyles` method from `typography` instance. 

```js
// override method before exporting typography
typography.injectStyles = function() {
  if (typeof document !== 'undefined') {
    // Replace existing
    const styleNode = document.getElementById('typography.js');
    if (styleNode !== null) {
      styleNode.innerHTML = this.toString();
    } else {
      const node = document.createElement('style');
      node.id = 'typography.js';

      // note: use typography context 
      node.innerHTML = typography.toString();

	  // note here: I use prepend, not appendChild
      document.head.prepend(node);
    }
  }
};
```

KyleAMathews:
@jbolda @Krizzu I assume this is development? @DSchau has a PR open for the same change as @Krizzu at https://github.com/KyleAMathews/typography.js/pull/201

jbolda:
@KyleAMathews yep, thanks for the heads up!

DSchau:
@jbolda any interest in validating the changes in that PR with your repo?

You can just build out and copy over files, or use `npm|yarn link` with the PR Kyle linked above.

Would help us assert that it's actually the fix--although it seems pretty obvious that is the fix!

KyleAMathews:
@DSchau's PR with the `injectStyles` fix is merged and published. Upgrade to `typography@0.16.19`

robert-claypool:
Now that `<style id="typography.js">` is forced to the top of `<head>`, I don't know how to include my own reset CSS file, or any other file that I want evaluated before typography.js CSS.

[Ejecting](https://www.gatsbyjs.org/docs/custom-html/) `default-html.js` to `src/html.js`, and adding `<link rel="stylesheet" href="reset.css" />` is now useless since there is no way to keep that reset tag above typography.js CSS. A reset should be the baseline on which typography.js and other custom CSS override.

robert-claypool:
This PR also pushes `<meta charset="utf-8">` down, but UTF-8 must be completely within the first 1024 bytes of the document, see https://github.com/h5bp/html5-boilerplate/blob/7.0.1/dist/doc/html.md#the-order-of-the-title-and-meta-tags

KyleAMathews:
@robert-claypool the expectation is that you'd use the normalize.css file Typography.js includes. Is there something special about your reset? Perhaps we could add support for custom resets to Typography.js. The assumption is that Typography.js _is_ your reset — so you wouldn't need anything more than that.

The second issue should definitely be fixed. Would you like to give it a go at fixing it? You'd need to edit the sort so that it puts `meta charset` above typography.js https://github.com/gatsbyjs/gatsby/blob/632f0356d2c75f91dfbd7cb6853499c17050f074/packages/gatsby-plugin-typography/src/gatsby-ssr.js#L26-L32

PR #12296, opened 2019-03-05
fix(gatsby-plugin-twitter): add back semi-colons to injected script

ollelauribostrom:
## Description
Adds back the semi-colors to the injected script that was removed in #12193. As described [here](https://github.com/gatsbyjs/gatsby/pull/12193#issuecomment-469418343) this was causing the twitter embeds not to render.


DSchau:
Published as `gatsby-plugin-twitter`@2.0.11. Thanks for the quick turnaround!

Issue #12225, opened 2019-03-01
gatsby develop can only serve 5 pages at a time


jeffrafter:
## Description

When running `gatsby develop` everything works perfectly and fast for the first 5 pages I open (in multiple tabs). Subsequent pages opened in new tabs hang indefinitely. When I close one of the previously opened tabs the last pages load immediately. I've tried to see if this is related to similar issues including (https://github.com/gatsbyjs/gatsby/issues/12143, https://github.com/gatsbyjs/gatsby/issues/11747 https://github.com/gatsbyjs/gatsby/pull/10257 and https://github.com/gatsbyjs/gatsby/issues/11727) but it seems different. 

![gatsby-closing-tabs](https://user-images.githubusercontent.com/4064/53659141-9c8a5c00-3c0f-11e9-8ce9-f58779694a0b.gif)

### Steps to reproduce

Start `gatsby develop`, open a bunch of tabs simultaneously.

### Expected result

All of them load, HMR is ready.

### Actual result

Only the first four load, the others hang.

### Environment

```
  System:
    OS: macOS 10.14.2
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.13.0 - ~/.nvm/versions/node/v10.13.0/bin/node
    Yarn: 1.12.3 - ~/.nvm/versions/node/v10.13.0/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v10.13.0/bin/npm
  Languages:
    Python: 2.7.15 - /usr/local/bin/python
  Browsers:
    Chrome: 72.0.3626.119
    Firefox: 65.0.1
    Safari: 12.0.2
  npmPackages:
    gatsby: ^2.0.85 => 2.1.17 
    gatsby-image: ^2.0.30 => 2.0.30 
    gatsby-plugin-canonical-urls: ^2.0.5 => 2.0.10 
    gatsby-plugin-feed: ^2.0.13 => 2.0.13 
    gatsby-plugin-google-analytics: ^2.0.14 => 2.0.14 
    gatsby-plugin-manifest: ^2.0.19 => 2.0.19 
    gatsby-plugin-react-helmet: ^3.0.0 => 3.0.6 
    gatsby-plugin-sharp: ^2.0.5 => 2.0.22 
    gatsby-plugin-styled-components: ^3.0.5 => 3.0.6 
[...]

PR #12302, opened 2019-03-05
Use SPDY for local development

jeffrafter:
## Description

This PR switches local development to use `SPDY` as opposed to HTTPs. This removes the connection limitations imposed by HTTP/1 and allows more than 5 simultaneous tabs open with hot module reloading.

## Related Issues

Fixes: https://github.com/gatsbyjs/gatsby/issues/12225

### DIFF
--- a/packages/gatsby/src/commands/develop.js
+++ b/packages/gatsby/src/commands/develop.js
@@ -200,7 +200,13 @@ async function startServer(program) {
 
   // If a SSL cert exists in program, use it with `createServer`.
   if (program.ssl) {
-    server = require(`https`).createServer(program.ssl, app)
+    server = require(`spdy`).createServer(
+      {
+        key: program.ssl.key,
+        cert: program.ssl.cert,
+      },
+      app
+    )
   }
   websocketManager.init({ server, directory: program.directory })
   const socket = websocketManager.getSocket()


adamclifford:
Spdy is now a deprecated protocol and will not be supported going forward. 
https://en.wikipedia.org/wiki/SPDY

The new HTTP/2 standard is based upon SPDY and is implemented in node since V8 

Here is an article about switching to HTTP/2
https://medium.com/the-node-js-collection/node-js-can-http-2-push-b491894e1bb1

I hope that helps.

sidharthachatterjee:
Thanks so much for the PR @jeffrafter 

Unfortunately, we're not going to merge this at this time because
- spdy is deprecated
- requiring https for local development is unintuitive and also a breaking change

jeffrafter:
> requiring https for local development is unintuitive and also a breaking change

To be clear, the enhanced capability is only added if `--https` otherwise everything continues to work normally. 

That said, reading more, I agree this should be closed as SPDY's deprecation is coming to node directly as well. The path to HTTP/2 (using express) looks to be a ways off. Two alternate approaches I could try is modifying the `develop` script to allow for injecting a server (decoupling SPDY from core Gatsby) or altering the HMR mechanism to a fallback that would allow for more connections. Is it worth it to see those in code?

kitze:
@jeffrafter did you maybe find a solution for this? Even a plugin or a node_modules patch could be useful.

PR #12303, opened 2019-03-05
fix(gatsby-transformer-remark): Fix unreturned Promise warning

sidharthachatterjee:


sidharthachatterjee:
@DSchau Does `cache.set` return a promise? Makes sense to return it then

DSchau:
It does, yeah! It should return the resolved value, but since we were returning null, shouldn’t really matter what we return!

sidharthachatterjee:
Yeah, makes sense. Changing that!

PR #12304, opened 2019-03-05
fix(gatsby-starter-blog): Return null in then in gatsby-node

sidharthachatterjee:
Fixes a warning from Bluebird about a promise being created and not returned


sidharthachatterjee:
@DSchau Yup, removes the warning. Node 6 doesn't support async/await. I'm not sure if we transpile gatsby-node

DSchau:
Our minimum recommended is Node 8, per this [RFC](https://github.com/gatsbyjs/rfcs/pull/24)

Also we don’t transpile them.

PR #12310, opened 2019-03-05
docs(www): Fix 'Plugin Library' link

Digetal:
## Description
It looks like there is a redirect in place from `/packages/` to `/plugins/`. This works for full page loads, but the link in this page still points to `/packages/` which results in a 404 via client side routing.

This change points the 'Plugin Library' link to `/plugins/`


Digetal:
No problem @sidharthachatterjee. Thanks for the approval.

I notice you changed the title. Are there any guidelines published anywhere for naming conventions so I can stick to them?

PR #12316, opened 2019-03-05
fix(www): set item count to 0 when undefined in plugin searchbar

### LINKED ISSUE #12294: gatsbyjs.org plugin search breaks with some values
## Description

Plugin search on www.gatsbyjs.org break page with some search phrases 

### Steps to reproduce

Just open this in a browser:
https://www.gatsbyjs.org/plugins/?=len
or
https://www.gatsbyjs.org/plugins/?=lenr
or
https://www.gatsbyjs.org/plugins/?=lern

### Expected result

0 to many search results should be visible.

### Actual result

Empty white page and error in console: `TypeError: Cannot read property 'toLocaleString' of undefined`

### Environment

Website. The same result on chrome and firefox (Kubuntu).

imbhargav5:
## Description

This PR fixes a bug that originates from RefinementList component when item.count is undefined.

## Related Issues

 Fixes #12294

### DIFF
--- a/www/src/components/plugin-searchbar-body.js
+++ b/www/src/components/plugin-searchbar-body.js
@@ -232,6 +232,14 @@ class Search extends Component {
             <Configure analyticsTags={[`gatsby-plugins`]} />
             <RefinementList
               attributeName="keywords"
+              transformItems={items =>
+                items.map(({ count, ...item }) => {
+                  return {
+                    ...item,
+                    count: count || 0,
+                  }
+                })
+              }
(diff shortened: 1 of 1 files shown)


pieh:
One thing to comment here - this seems like bug in `react-instantsearch/dom` - this solves the crash but there other things breaking (or rather not working properly)

I'll merge this to fix the crash, but we should look into updating our `react-instantsearch` version (from 4 to 5) to see if this was fixed upstream (and potentially create issue on their repo to track it)

imbhargav5:
YES!!!!!!!!!!!!!!!!!!!!!!!!

PR #12325, opened 2019-03-05
ci: add build-www task (and lightly refactor scripts)

DSchau:
## Description

Adds an approval task for building out www. When manually triggered, will build out the www/ folder so we can check for errors.

Note: this will be an extremely slow running task (probably ~25 minutes or more) so will want to use this sparingly on critical PRs.


wardpeet:
This is sooooooo awesome!!!

DSchau:
Woo hoo! Works now, now let's see if it actually builds out and how long it takes.

example

wardpeet:
Next step is to add an preview in gatsbot ^^

DSchau:
These environment variables are driving me insane 😩

We have GITHUB_API_TOKEN exposed via CircleCI to the container, but this doesn't seem to be making its way into gatsby-node.js. Working on fixing this.

DSchau:
Update here - this is blocked until we can find a clean way to expose `GITHUB_API_TOKEN` to forked containers.

DSchau:
One step forward, one step back. I removed the hard failing on `GITHUB_API_TOKEN` failing to exist in a CI environment, but the [build still fails](https://circleci.com/gh/gatsbyjs/gatsby/80169). Looks to be some type of image issue--running a formatter may just fix, so let's see.

wardpeet:
@DSchau I don't think it will ever be possible. Travis-ci has the same issue, the only possible way is to have a proxy doing it for us that contains the api token.

DSchau:
[Cool failure pngquant](https://circleci.com/gh/gatsbyjs/gatsby/83156)

jlengstorf:
@DSchau looks like a different one blew up this time. 😕

DSchau:
@jlengstorf in 100 hours I will get all of these 🙃

DSchau:
This is currently blocked by this _fun_ error:

> Error: spawn E2BIG

Which is apparently thrown by `terser-webpack-plugin` which uses a library `worker-farm` under the hood.

I think it's the same issue mentioned/solved in #10257, specifically that CircleCI reports os.cpus() for the machine rather than the container (32 vs 2), e.g. https://discuss.circleci.com/t/memory-problems-with-jest-and-workers/10297

Trying with `--no-uglify` 🤞

PR #12336, opened 2019-03-05
fix(gatsby): Vendor express-static so we can avoid loading html files from public during development

KyleAMathews:
Fixes https://github.com/gatsbyjs/gatsby/pull/12243#pullrequestreview-210895624

Works locally for reactjs.org which is a site that uses paths like
`/something/about-us.html`

screenshot 2019-03-05 14 18 11


packages/gatsby/src/commands/develop-static.js
```diff
@@ -0,0 +1,117 @@
...
+      res.end()
+      return null
+    }
+
+    var forwardError = !fallthrough
+    var originalUrl = parseUrl.original(req)
+    var path = parseUrl(req).pathname
+    const parsedPath = sysPath.parse(path)
+
+    // Ignore html files.
```
KyleAMathews: This is what I added

KyleAMathews:
We have some development server tests? This would be easy to test. Write out a html file to public and then request it and a regular path from the Dev server and verify the two html files are the same.

wardpeet:
How do we feel about the following code?
This way we don't need to keep serve-static in sync, we just do the html check ourselves. rest is handled by static-serve.

```js
const sysPath = require(`path`)
const express = require(`express`)
const parseUrl = require(`parseurl`)

/**
 * Module exports.
 * @public
 */

module.exports = serveStatic

/**
 * @param {string} root
 * @param {object} [options]
 * @return {function}
 * @public
 */

function serveStatic(root, options) {
  const expressStatic = express.static(root, options)

  return function(req, res, next) {
    if ([`GET`, `HEAD`].includes(req.method)) {
      const path = parseUrl(req).pathname
      const parsedPath = sysPath.parse(path)
      if ([`.htm`, `.html`].includes(parsedPath.ext)) {
        return next()
      }
    }

    return expressStatic(req, res, next)
  }
}
```

DSchau:
@kyleamathews @wardpeet that seems like a really good solution! It’s nice to not have to patch and maintain things—so if that works the same we should definitely go that route!

KyleAMathews:
Yeah that'd be much better 👌

KyleAMathews:
Want to just take over the PR?

sidharthachatterjee:
Went ahead and made Ward's changes!

DSchau:
Successfully published:
 - gatsby@2.1.27

PR #12344, opened 2019-03-06
feat(gatsby): chunk page bundles on develop

### LINKED ISSUE #8342: Enable split chunk plugin in development
I disabled it for no good reason and webpack maintainers say it's faster for webpack to do HMR with chunks + it'd be faster to load dev sites if you don't have to load all code up front.

The main work to do I think is to make both dev and prod use async-requires.js from .cache as that'll automatically create splits in development. Also update loading resources for development so similar to production.

This will speedup loading the development site as there'll be far less JS loaded initially for larger sites. E.g. gatsbyjs.com loads 8mbs of JS right now.

wardpeet:
## Description
This an attempt to make async requires work inside develop. For now, I used react loadable so  I didn't have to change to much and have a playing field. Hot-reloading is our biggest issue here. The benefit of moving everything to async-requires is to have on require story instead of 2 and have faster recompilation. (haven't actually measured it for .org)

React loadable gives us the benefit of having an SSR & client side story without much work. We could swap loadable with React.lazy on the client and when suspense land on the server as well so we don't add extra bundle cost. Loadable adds 4.75kb to the bundle:
image

I'm pretty confident we could do it without loadable as we do it for production. The biggest problem is HMR which is a problem with loadable as well. react-hot-loader needs to be on the export level and not on the import level when doing async components. We have 2 options. We write a wrapper for each page that we save inside cache and use as a component path instead of the original page component (haven't tested this but should work). The other option is to let webpack augment the source of the entry file with the react-hot-loader package.

I'm in favour of the .cache strategy as it might be easier to follow than some webpack voodoo.

# Testing
if you like to test this, you'll need to wrap hot calls in the pages itself.
```jsx
import React from "react"
import { Link } from "gatsby"
import { hot } from 'react-hot-loader/root';

import Image from "../components/image"
import SEO from "../components/seo"

const IndexPage = () => (
  <Layout>
    <SEO title="Home" keywords={[`gatsby`, `application`, `react`]} />
    <h1>Hi people</h1>
    <p>Welcome to your new Gatsby site.</p>
    <p>Now go build something great.</p>
    <div style={{ maxWidth: `300px`, marginBottom: `1.45rem` }}>
      <Image />
    </div>
    <Link to="/page-2/">Go to page 2</Link>
  </Layout>
)

export default hot(IndexPage)
```

## Related Issues

Fixes #8342

### DIFF
--- a/packages/gatsby/cache-dir/app.js
+++ b/packages/gatsby/cache-dir/app.js
@@ -6,8 +6,9 @@ import socketIo from "./socketIo"
 import emitter from "./emitter"
 import { apiRunner, apiRunnerAsync } from "./api-runner-browser"
 import loader, { setApiRunnerForLoader, postInitialRenderWork } from "./loader"
-import syncRequires from "./sync-requires"
-import pages from "./pages.json"
+// import syncRequires from "./sync-requires"
+import asyncRequires from "./async-requires"
+import { dataPaths, pages } from "./data.json"
 
 window.___emitter = emitter
 setApiRunnerForLoader(apiRunner)
@@ -47,7 +48,8 @@ apiRunnerAsync(`onClientEntry`).then(() => {
   )[0]
 
   loader.addPagesArray(pages)
-  loader.addDevRequires(syncRequires)
+  loader.addDataPaths(dataPaths)
+  loader.addRequires(asyncRequires)
 
   loader.getResourcesForPathname(window.location.pathname).then(() => {
     const preferDefault = m => (m && m.default) || m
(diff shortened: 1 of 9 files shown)


pieh:
We could try using https://www.npmjs.com/package/react-hot-loader-loader to apply `hot` wrapper for page template files, so `(a)sync-requires` could be something like this:

```js
"${c.componentChunkName}": () => import("${is_develop ? `react-hot-loader-loader!` : ``)${path_to_component_file}")
```

Next steps would be to adjust https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/cache-dir/loader.js
to straight remove all all/most code path differences between dev and prod builds (we still need different handling for fetching query result data, so this need some research) and all imports for `sync-requires` (other than one in `static-entry` I think)

DSchau:
Possibly consider just shipping this with react-loadable--and then look at the custom dev 404 page later.

DSchau:
Blocked until #13004 lands.

Issue #12327, opened 2019-03-05
gatsby-remark-code-repls do not support scoped dependencies (started by @ character)


krzysztofpniak:
## Description

At the moment `gatsby-remark-code-repls` plugin do not support scoped npm packages (started by @ character). It expects only one @ character for name - version separation.

### Steps to reproduce
1. git clone https://github.com/reactjs/reactjs.org.git
2. cd reactjs.org
3. run `yarn`
4. go to file: `gatsby-config.js`
5. find `dependencies` array and add `@material-ui/core` to that array
6. go to file: `content/docs/hello-world.md`
7. replace `[](codepen://hello-world)` by `[](codesandbox://hello-world)`
8. run project by: `npm run dev`
9. go to your browser: http://localhost:8000/docs/hello-world.html and click link `Try it on CodePen`

### Expected result

code on the Code Sandbox should have @material-ui/core correctly attached (`"@material-ui/core": "latest"`)

### Actual result

if you one package.json you'll see "": "material-ui/core" in dependencies.

I have patched it on my locak copy. I can prepare pull request with the patch.

### Environment

 System:
    OS: macOS 10.14.2
    CPU: x64 Intel(R) Core(TM) i5-5287U CPU @ 2.90GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 8.12.0 - ~/.nvm/versions/node/v8.12.0/bin/node
    Yarn: 1.12.3 - /usr/local/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v8.12.0/bin/npm
  Browsers:
    Chrome: 72.0.3626.119
    Firefox: 64.0
    Safari: 12.0.2
  npmPackages:
    gatsby: ^2.0.0 => 2.0.21 
    gatsby-plugin-catch-links: ^2.0.0 => 2.0.4 
    gatsby-plugin-feed: ^2.0.0 => 2.0.8 
    gatsby-plugin-glamor: ^2.0.0 => 2.0.5 
    gatsby-plugin-google-analytics: ^2.0.0 => 2.0.6 
    gatsby-plugin-manifest: ^2.0.0 => 2.0.5 
    gatsby-plugin-netlify: ^2.0.0 => 2.0.1 
    gatsby-plugin-nprogress: ^2.0.0 => 2.0.5 
    gatsby-plugin-react-helmet: ^3.0.0 => 3.0.0 
[...]

PR #12347, opened 2019-03-06
fix(gatsby-remark-code-repls): handle scoped packages in dependencies…

krzysztofpniak:
… (#12327)
## Description

It allows dependencies with scope like @material-ui/core in the plugin gatsby-remark-code-repls config. That dependencies are used with codesandbox intgration.

## Related Issues

Fixes: #12327


packages/gatsby-remark-code-repls/src/index.js
```diff
@@ -134,12 +134,10 @@ module.exports = (
             "package.json": {
               content: {
                 dependencies: dependencies.reduce((map, dependency) => {
-                  if (dependency.includes(`@`)) {
-                    const [name, version] = dependency.split(`@`)
-                    map[name] = version
-                  } else {
-                    map[dependency] = `latest`
-                  }
+                  const [, name, , version] = /^(@?.+?)(@(.+))?$/.exec(
```
DSchau: Do you think you could find a cleaner or more self-documenting solution here?

A bit of a nit, but this is really hard to follow.

krzysztofpniak: What about:
```javascript
if (dependency.includes(`@`)) {
  const splitIndex = dependency.findLastIndex('@')
  const name = dependency.substr(0, splitIndex)
  const version = dependency.substr(splitIndex + 1)
  map[name] = version
} else {
  map[dependency] = `latest`
}
```
If you are fine with this version, do you still want a unit test for it? If so, I think its better to extract it to function to a separate file and make some test for it. Anyways, I think after adding unit test I should also cover most complicated cases, where we have two or more @ signs and at least one of them is not at position 0. That test will force me to complicate the code. But I'm leaving this decision up to you.

DSchau: @krzysztofpniak yeah, I'm more aligned with that version, but that also doesn't really work in its current form.

Given:

```js
const getNameAndVersion = (dependency, map) => {
  if (dependency.includes(`@`)) {
    const splitIndex = dependency.findLastIndex('@')
    const name = dependency.substr(0, splitIndex)
    const version = dependency.substr(splitIndex + 1)
    map[name] = version
  } else {
    map[dependency] = `latest`
  }
  return map
}
```

`getNameAndVersion('react')` `{ react: 'latest' }`
`getNameAndVersion('react@16.0.0')` // dependency.findLastIndex is not a function at getNameAndVersion

I'm definitely in favor of breaking out and adding a few unit tests to assert functionality here.

krzysztofpniak: Sorry for the delay. Now it is ok, I also added unit tests.

pieh: I know it's late - but maybe this would be most self documenting approach - https://www.npmjs.com/package/npm-package-arg ;)

krzysztofpniak: @pieh using external libraries for such a small things increases node_modules size. We live in the days when people create libraries for everything but IMHO this is not the way. Of course if library is widely used in project it makes sense. In this situation this is just one call of one simple function.

DSchau: @krzysztofpniak we are sympathetic to that--but using those small little packages also tends to have a lot of other benefits, e.g.

- We don't need to test the code
- We get any fixes (or new features, updates, etc.) for free just by using the package

So for Node.js code--we really recommend using packages if they are available and can be used.

DSchau: Would you be up for making that change? I'd recommend using that library--great idea @pieh

krzysztofpniak: Of course. Done.

wardpeet:
it looks great! thanks a bunch! Do you have time to create a small test for this one to make sure this one is fixed.

krzysztofpniak:
@wardpeet, how this test should look like? I'm a little bit confused. I created tests 8 days ago, then @pieh mentioned `npm-package-arg` and @DSchau [wrote](https://github.com/gatsbyjs/gatsby/pull/12347#discussion_r267024326), that this way tests are not required. So I removed tests I created before.

pieh:
@krzysztofpniak This is mostly for integration - to verify usage of that package (and not package itself) works, let me look up details, push some minimal test and get this merged finally

Issue #6591, opened 2018-07-19
Add guess.js to gatsbyjs.org


KyleAMathews:
https://github.com/guess-js/guess is a key part of the narrative for v2 so we need to make sure it's running well on gatsbyjs.org

/cc @mgechev and @addyosmani

KyleAMathews:
https://github.com/gatsbyjs/gatsby/issues/9572

m-allanson:
We still want to do this.

lannonbr:
Closing this as it has been implemented in #12351

Issue #9572, opened 2018-10-30
Regression in the Guess.js plugin for Gatsby


mgechev:
## Description

Currently, the Guess.js plugin for Gatsby does not prefetch the resources associated with the pages. It only invokes `getResourcesForPathname` but does not perform any actions.

It looks like the regression was introduced by [this PR](https://github.com/gatsbyjs/gatsby/commit/2605aa0ec74a79fa56bc0eae8d101356c25c38de#diff-79243c75f482266054a48c78b300dc5b).

Downgrading the Guess.js plugin to 1.0.0 resolves the issue.

### Steps to reproduce

Install the Guess.js plugin version 1.0.2 and introduce the plugin in `gatsby-config.js`.

### Expected result

The plugin should prefetch assets associated with pages predicted from Guess.js

### Environment

Gatsby version 2.0.31
Guess.js plugin version 1.0.2

kakadiadarpan:
@mgechev can you provide a reproduction repo for this?

mgechev:
You can reproduce the issue by unpinning the plugin version in this project https://github.com/guess-js/guess-js.github.io. Alternatively, you can introduce the latest version of the Guess.js plugin to Gatsby.com and you should get the same behavior.

PR #12351, opened 2019-03-06
feat(guess): add guess.js to gatsbyjs.org

wardpeet:
## Description
Enabling guess.js to gatsbyjs.org. It only works when you enable the service account env variable as you need special gatsby access to get to it so it's disabled by default.

When building the site we need to get the key.pem and service account from somewhere. Service account can be stored inside an env variable. The key file I'm not sure.

## Related Issues
#6591

### DIFF
--- a/www/README.md
+++ b/www/README.md
@@ -40,6 +40,17 @@ GATSBY_FEEDBACK_KEY_STARTERLIB=ADD_KEY
 
 If there's a problem with the feedback widgets, please open an issue in the repo.
 
+### Enabling guess.js
+
+Guess.js is disabled by default and can be enabled by setting `ANALYTICS_SERVICE_ACCOUNT` and `ANALYTICS_SERVICE_ACCOUNT_KEY` env variables. These variables need to have access to the gatsbyjs.org analytics.
+
+If you have access to the keys, add them like so:
+
+```
+ANALYTICS_SERVICE_ACCOUNT="service account@email.com"
+ANALYTICS_SERVICE_ACCOUNT_KEY="PEM KEY VALUE"
+```
+
 ## Running slow build? (Screenshots placeholder)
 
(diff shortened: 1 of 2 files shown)


www/README.md
```diff
@@ -40,6 +40,18 @@ GATSBY_FEEDBACK_KEY_STARTERLIB=ADD_KEY
 
 If there's a problem with the feedback widgets, please open an issue in the repo.
 
+### Working with guess.js
+
+To enable Guess.js you'll need to have access to gatsbyjs.org analytics service account. Guess needs a key file and a service account.
```
yogeshkotadiya: You mean `Google analytics account?` Right.

www/gatsby-config.js
```diff
@@ -2,6 +2,32 @@ require(`dotenv`).config({
   path: `.env.${process.env.NODE_ENV}`,
 })
 
+const GA = {
+  identifier: `UA-93349937-5`,
```
muescha: suggest set this also as `process.env`

wardpeet: thanks but there is no real benefit to add it as an env var.

wardpeet:
Guess is only added on production so best to copy your .env.development over to .env.production and remove the GITHUB_TOKEN error code.

DSchau:
@wardpeet is this ready to merge? We'll need to add some environment variables to appropriate containers, correct?

wardpeet:
@DSchau yes, you'll have to add these 2 env vars. If you don't, it will still build but without guess support so we don't need to expose those secrets in forks, ....

When you run `gatsby build` & `gatsby serve` you'll have `__GUESS__` on the window object.

mgechev:
Hey everyone, thanks for integrating Guess.js with gatsbyjs.org ❤️ ! Sometime back I noticed a regression in the Guess.js plugin for Gatsby https://github.com/gatsbyjs/gatsby/issues/9572.

Did you have a chance to verify that the prefetching works as expected and the regression has been resolved?

wardpeet:
@mgechev let me check it out, pretty sure it works.

Issue #12343, opened 2019-03-06
Issues with large page numbers (>60k)


me4502:
## Description

I've been setting up a content surfacing system using GatsbyJS, and we're encountering a fair few issues with the amount of pages we have. I've made a few changes as suggested in Discord, and done a fair bit of investigating into the cause of the slowdowns.

The following symptoms have been noticed:

1. `npm run develop` is significantly slower than `npm run build` (50 minutes vs 2 minutes)
2. Slowdown occurs during the "running graphql queries" step, but before the text has shown
3. Some machines encounter an "invalid instruction" crash after the "info bootstrap finished" text (http://paste.enginehub.org/tb7s6C)
4. It appears to be running a query per page, despite the fact that the pages have no queries. There appears to already be an issue for this (https://github.com/gatsbyjs/gatsby/issues/12216)

A few notes about our setup:

- All data is passed to pages/templates via the context, rather than a per-page query (As per discord recommendation)
- I can provide the source code to the Gatsby team privately if required

What we've discovered:

- The major slowdowns appear to be related to the queue library, and a QuickSort that gets run on insertion of elements. From what we can tell, the "Running graphql queries" step is actually running a list of tasks that may not be related to graphql, so it may be worth renaming this.
- As for why it significantly slows down in develop, we discovered that in build the priority function is deleted from the queue. We've noticed the same speedups by setting the priority of non-active paths to `undefined`, which skips sorting for them entirely. (https://github.com/diamondio/better-queue-memory/blob/cff881f2074ff0508bcb6e932bda0b92977d3d2b/index.js#L48)
- Halving the page number takes the time from 50 minutes to 10 minutes, so it's not a linear slowdown.

### Steps to reproduce

Using the source code that I can provide privately:

1. Run `npm run build`, notice the time it takes to run, and then crashes Node.
2. Run `npm run develop`, notice that it takes significantly longer, with the same crash.

### Expected result

Gatsby should be able to handle this quantity of pages, as there are multiple sources that state they're running ~10 million with little to no issue.

### Actual result

Gatsby struggles at these page numbers.

### Environment

  System:
    OS: macOS 10.14.2
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.15.3 - ~/.nvm/versions/node/v10.15.3/bin/node
[...]

PR #12365, opened 2019-03-06
fix(develop): don't sort the task queue when adding low-priority tasks to the end

me4502:
## Description

This change prevents the entire task queue being sorted every time a low-priority task is added to the queue. This takes a run of `npm run develop` with 60k pages from 50 minutes to 1.5 minutes. There should be no behaviour change here, as these were already the lowest priority and will be added to the end of a FIFO queue. High priority tasks still sort, which means they will still be correctly moved to the front.

## Related Issues

Fixes one aspect of #12343


pieh:
This seems very valid fix (or rather workaround) for excessive sorting done by default memory store used by `better-queue` ( https://github.com/diamondio/better-queue-memory/blob/master/index.js#L41-L57 ) - as pointed in description - it just pushes task to tail of queue and avoid excessive sorting

me4502:
I think this should basically fix it for any size site also, as I assume the activePaths refers to currently open pages in the web browser that are being handled by web hooks? So it'll only sort for the number of different pages open in tabs. Which means during the initial build process it'll never actually do a sort, only on hot-reload.

pieh:
Ok, I just checked and setting priority to `undefined` will break prioritization completely, because comparator will always always be `false` when you compare any value with undefined:

```
10 < undefined // false
10 > undefined // false
10 == undefined // false
```

As you mentioned we still don't need priorities for initial query running so instead of this change we could return `undefined` during initial query running:
Priority could look like:
```js
priority: (job, cb) => {
	if (runningInitialQueries) {
      return cb(null, undefined)
    }
    const activePaths = Array.from(websocketManager.activePaths.values())
    if (job.id && activePaths.includes(job.id)) {
      cb(null, 10)
    } else {
      cb(null, undefined)
    }
  },
```
`runningInitialQueries` would be initially false, and we could call from https://github.com/gatsbyjs/gatsby/blob/1ef70f474dab51c44f92a001ba2d88e1ba318245/packages/gatsby/src/internal-plugins/query-runner/page-query-runner.js#L27-L35 after `await runQueries() ` that we finished initial queries and to toggle `runningInitialQueries` to false and enable prioritization after that

pieh:
Other - probably mess lessy solution would be to create custom better-queue store implementation that instead of sorting would have priority queue and normal queue (because we only have really 2 priorities) and wouldn't do any sorting just always get from prio queue first and only after that pop tasks from normal queue.

For reference - https://github.com/diamondio/better-queue-memory/blob/master/index.js this is currently used store implementation so ~100 loc which could be used as a base

me4502:
One potential idea I've had rather than having two queues, would be to have a custom MemoryStore inside Gatsby that handles "high/low priority" as a boolean in putTask, and then just unshifts / pushes to the queue based on which is given. Would this work? No sorting, and it's a smaller change as it doesn't require setting up a second queue and changing all the task queue length etc code.

me4502:
I've implemented a solution that is based off your suggestion, however I made it support an arbitrary number of priorities rather than just 2.

pieh:
> however I made it support an arbitrary number of priorities rather than just 2.

Nice!

Let me take that for a test run

pieh:
Ok, it seems like it has some problems. If You don't mind I'll apply some changes and few tests

me4502:
Looks good, thanks. Didn’t think about the ability to change priorities etc after the task had been made.

KyleAMathews:
Nice work! Better living through algorithms 👌

me4502:
Would it be possible to please get this reviewed soonish? 😃

pieh:
Published `gatsby@2.2.2`

PR #12366, opened 2019-03-07
fix(gatsby-plugin-canonical-urls): fix typo in "How to use" section (http -> https)

bryndyment:
https://www.gatsbyjs.org/packages/gatsby-plugin-canonical-urls/

## Description

`siteUrl` includes `https` but the sample generated tag has `http` … looks like a minor typo

## Related Issues

(none)

### DIFF
--- a/packages/gatsby-plugin-canonical-urls/README.md
+++ b/packages/gatsby-plugin-canonical-urls/README.md
@@ -28,5 +28,5 @@ With the above configuration, the plugin will add to the head of every HTML page
 a `rel=canonical` e.g.
 
 ```html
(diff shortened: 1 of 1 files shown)


bryndyment:
Hi!  Actually, the thing I was pointing out was the fact that *part* of the example uses `http` and *another part* of the example uses `https`.

They should match, regardless of whether it's `http` or `https`.

This PR just makes sure that both parts of the example are matching.

I believe this was just a typo/oversight.

wardpeet:
Good gatch, it's indeed an oversight 👍 thanks for fixing!

PR #12369, opened 2019-03-07
fix gatsby showcase file

hesamkaveh:
Recently added my weblog to gatsby showcase, but have an extra title that didn't use


hesamkaveh:
it will show on gastsby [showcase](https://www.gatsbyjs.org/showcase/?filters%5B0%5D=WordPress)?

DSchau:
@hesamkaveh it will, yeah! We need to give it a little time to build out, but it should be up in 20-30 minutes.

PR #12372, opened 2019-03-07
chore(blog): add missing parenthesis to code example

sondrele:
## Description

Copied the code example and noticed that a closing parenthesis was missing.


pieh:
Oh, so weird, Prettier doesn't like that, but it seems right :O

pieh:
> Oh, so weird, Prettier doesn't like that, but it seems right :O

Ok, this was just because embedded code block had different formatting issues ;) running format (which I just did) should fix this (hopefully)

Issue #12078, opened 2019-02-25
gatsby-cli error semantics `Did you mean <command>?` doesn't correctly convey command options given


anthonytranDev:
## Description
When `gatsby y develop` has been typed. The resulting stdout suggests `Did you mean new ?`, more precisely `new` being very different from `develop`. Doesn't accurately depict the command typed in

### Steps to reproduce
Demo Project - not required
Start up a new project gatsby and type in `gatsby y develop`

### Expected result
`Did you mean develop?`

### Actual result
`Did you mean new?`

### Solution Suggestions
Include the following criteria
1. Increase the accuracy of string search with a **list** of suggestions

```
// Example based on npm
$ npm istall

// Result
Did you mean one of these?
    install
    uninstall
    star
```
2. Notify developer that the command is wrong and throw error

```
// Example of incorrect yarn command entered
$ yarn a upgrade

// result
yarn run v1.13.0
error Command "a" not found.
info Visit https://yarnpkg.com/en/docs/cli/run for documentat
ion about this command.
```

### Environment
```
  System:
    OS: macOS Mojave 10.14.3
    CPU: x64 Intel(R) Core(TM) i5-7567U CPU @ 3.50GHz
    Shell: 3.2.57(1)-release - /bin/bash
  Binaries:
    Node: v10.14.2 - /Users/anthonytran/.nvm/versions/node/v10.14.2/bin/node
    npx: v6.5.0 - /Users/anthonytran/.nvm/versions/node/v10.14.2/bin/npx
    Yarn: 1.13.0 - /usr/local/bin/yarn
    npm: 6.5.0 - /Users/anthonytran/.nvm/node_modules/.bin
    gatsby: 2.4.8 - /Users/anthonytran/.config/yarn/global/node_modules/gatsby-cli
  Browsers:
    Chrome: 72.0.3626.109
  npmPackages:
     dependencies
      gatsby: ^2.1.16,
      gatsby-image: ^2.0.29,
      gatsby-plugin-manifest: ^2.0.18,
      gatsby-plugin-offline: ^2.0.24,
      gatsby-plugin-react-helmet: ^3.0.6,
      gatsby-plugin-sharp: ^2.0.21,
      gatsby-source-filesystem: ^2.0.22,
      gatsby-transformer-sharp: ^2.1.14,
      prop-types: ^15.7.2,
      react: ^16.8.3,
      react-dom: ^16.8.3,
      react-helmet: ^5.2.0
```
[...]

PR #12392, opened 2019-03-08
feat(gatsby-cli) better suggestions

anthonytranDev:
## Description
Gatsby-cli will give more accurate suggestion(s) upon fail. 

## Related Issues
This is #12078

### DIFF
--- a/packages/gatsby-cli/src/create-cli.js
+++ b/packages/gatsby-cli/src/create-cli.js
@@ -279,17 +279,6 @@ function isLocalGatsbySite() {
   return inGatsbySite
 }
 
-function getOnFailSuggestion(arg, commands) {
-  let suggestion
-  if (!arg) {
-    suggestion = `Pass --help to see all available commands and options.`
-    return suggestion
-  } else {
-    const suggestion = didYouMean(arg, commands) || ``
-    return suggestion
-  }
-}
-
 module.exports = argv => {
   let cli = yargs()
   let isLocalSite = isLocalGatsbySite()
@@ -325,7 +314,7 @@ module.exports = argv => {
       ),
     })
     .wrap(cli.terminalWidth())
-    .demandCommand(1)
+    .demandCommand(1, `Pass --help to see all available commands and options.`)
     .strict()
     .fail((msg, err, yargs) => {
       const availableCommands = yargs.getCommands().map(commandDescription => {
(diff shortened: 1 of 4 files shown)


packages/gatsby-cli/src/create-cli.js
```diff
@@ -315,7 +316,17 @@ module.exports = argv => {
...
+      const availableCommands = yargs.getCommands().map(commandDescription => {
+        const [command] = commandDescription
+        return command.split(` `)[0]
+      })
+      const arg = argv.slice(2)[0]
+      const suggestion = arg ? didYouMean(arg, availableCommands) : ``
+
+      cli.showHelp()
+      report.log(suggestion)
+      report.log(msg)
```
sidharthachatterjee: Should be log the `msg` as well? 

```
~/Code/gatsby/www feat/gatsby-cli-suggestions*
❯ ./node_modules/gatsby-cli/lib/index.js ne .
Usage: node_modules/gatsby-cli/lib/index.js <command> [options]

Commands:
  develop                   Start development server. Watches files, rebuilds, and hot reloads if something changes
  build                     Build a Gatsby project.
  serve                     Serve previously built Gatsby site.
  info                      Get environment information for debugging and issue reporting
  clean                     Wipe the local gatsby environment including built assets and cache
  repl                      Get a node repl with context of Gatsby environment, see (add docs link here)
  new [rootPath] [starter]  Create new Gatsby project.

Options:
  --verbose      Turn on verbose output                                                                                                                                       [boolean] [default: false]
  --no-color     Turn off the color in output                                                                                                                                 [boolean] [default: false]
  -h, --help     Show help                                                                                                                                                                     [boolean]
  -v, --version  Show version number                                                                                                                                                           [boolean]


Did you mean this?
     new

Unknown arguments: ne, .
```

In this case, maybe we can do without the `Unknown arguments` bit. What do you think?

sidharthachatterjee: Seems like this is good to have for future proofing as correctly pointed out by @pieh

sidharthachatterjee:
Published in gatsby-cli@2.4.14

packages/gatsby-cli/src/create-cli.js
anthonytranDev: Just to let you know, I decided to log the `msg` as `Unknown arguments: <args>` as it clearly states the mistakes the user has made.

This will help Newcomers and Junior Developers like myself take up Gatsby faster. 

The ethos here is copy a **Rust Lang** level of stderr, as to makes it so obvious, that it makes **Gatsby** easier to adopt

Nonetheless I've been thinking of some improvements, that can be made to make the `Unknown arguments: ` part more readable and aesthically pleasing to Gatsby Commnity

I've already got some ideas, plus it would be nice to hear back on what you guys think 😄

PR #12400, opened 2019-03-08
fix(gatsby-cypress): Run files through babel

LekoArts:
Cypress was complaining about `import` statements. So with this PR the files get "babelfied".
Also added a note about the needed env var 👍


sidharthachatterjee:
Published in gatsby-cypress@0.1.5

PR #12401, opened 2019-03-08
[schema] Add code frame to SDL parsing error

stefanprobst:
This is for #11480 

* Added code frame to SDL parsing error message.
* Adjusted using-type-definitions example to also import typedefs from external .gql file.
* Consistent quotes in error messages.


freiksenet:
Awesome stuff.

PR #12402, opened 2019-03-08
Update steps in README for creating a gatsby site

topaztee:
Running ```gatsby develop``` won't work without first installing the npm packages, Currently this step is missing in the README.


sidharthachatterjee:
@topaztee gatsby new will install dependencies for you so it shouldn't be necessary to run `npm install` 🙂

marcysutton:
I agree, @sidharthachatterjee, I don't think we should make this change because that's what `gatsby-new` does. A better change would be to note that's where the installation is happening. Something like:
> 2.  Install a new Gatsby site including dependencies: `gatsby new hello-world https://github.com/gatsbyjs/gatsby-starter-hello-world`. (_Note: Depending on your download speed, the amount of time this takes will vary. For brevity's sake, the gif below was paused during part of the install_).

topaztee:
never mind. was running an older node version

PR #12403, opened 2019-03-08
fix(scroll): use location.key for scroll behaviours

### LINKED ISSUE #12390: Improve location key for scroll-behaviors so it's not purely derived from the path
https://github.com/taion/scroll-behavior/issues/135#issuecomment-468962628

hentielouw:
## Description

This is a simple fix to use `location.key` for saving scroll state, if it's available.

Based on the suggestions found in this discussion: https://github.com/taion/scroll-behavior/issues/135#issuecomment-468962628

## Related Issues

Fixes #12390.


KyleAMathews:
What does `locationKey` look like when you browse around?

Also I'm assuming you verified the bug I posted doesn't happen any longer?

KyleAMathews:
And thanks for jumping on this!

hentielouw:
I've tested it locally and looks like it's working as expected. I haven't added any automated tests since the package doesn't seem to have any tests at the moment - happy to add tests if you think that's needed, but might need a hand!

To illustrate how `locationKey` changes, I've added a `console.log` for `locationKey` and set up a test site (Gatsby Default Starter with some Lorem Ipsum). I took the screenshot below after navigating using "Go to page 2" and "Go back to the homepage" links a few times (without using "back"). I then used browser "back" to go to previous pages.

On first load the `locationKey` value is `initial`. Scrolling up and down logs the same value multiple times. On each navigation `locationKey` changes to a new integer value. Pressing "back" uses previous key values.

Hope that makes sense! Gets a bit muddy talking about this scrolling/navigation behaviour.

screenshot 2019-03-08 at 17 25 41

pieh:
I've added tests for scroll behavior simulating example scenario with some changes (from https://github.com/taion/scroll-behavior/issues/135#issuecomment-468962628 )
```
visit /
PUSH to /long-page
Scroll to middle-of-the-page
PUSH to /
PUSH to /long-page
Scroll to bottom-of-the-page
PUSH to /
POP to /long-page
  assert path is /long page
  assert we are scrolled to bottom-of-the-page
  assert we are NOT scrolled to middle-of-the-page
POP to /
POP to /long-page
  assert path is /long page
  assert we are scrolled to middle-of-the-page
  assert we are NOT scrolled to bottom-of-the-page
```

Which looks like this in cypress logs - hopefully in works in CI correctly:
Screenshot 2019-03-28 at 02 59 29

t2ca:
Looking for some help, this update stopped the following code from working on my project.

```
exports.shouldUpdateScroll = () => {
  if (window.__navigatingToLink === true) {
    return [0, 0]
  }
  return true
}

exports.onRouteUpdate = () => {
  window.__navigatingToLink = false
}
```

I added this code to gatsby-browser.js to fix an issue I was having, but since this update, this code no longer works for me. Would love some help, thanks!

KyleAMathews:
What was the issue you were having and what was the above code supposed to do and how does it behave now? Can you open a new issue with answers to those questions? Thanks! That'll help us help you.

t2ca:
Thanks Kyle for taking the time to respond, i opened a new issue https://github.com/gatsbyjs/gatsby/issues/12997

narration-sd:
Please see where the fix rolled in at 2.0.7 is breaking importantly expected 2.0.6 behavior, so that we need this regression optionable-out, thanks: https://github.com/gatsbyjs/gatsby/issues/12997#issuecomment-480454997

PR #12414, opened 2019-03-08
chore(docs): changed from `npm run` to `gatsby` in doc quick start

muescha:
## Description

changed from `npm run` to `gatsby`

## Related Issues

Addresses #12195


DSchau:
Would you have any interest in addressing your comment here?

> but there much more places when i run a search for npm run over the full repo

If there are more places, it would be great if we could get them into this PR. Thanks!

muescha:
sorry - no time today and on weekend :(

PR #12425, opened 2019-03-08
chore(www): Remove non-Gatsby site from the showcase

m-allanson:
## Description

This url isn't a Gatsby site :(

### DIFF
--- a/docs/sites.yml
+++ b/docs/sites.yml
@@ -1188,12 +1188,6 @@
   categories:
     - Entertainment
     - News
-- title: Cool Hunting
-  main_url: "http://www.coolhunting.com/"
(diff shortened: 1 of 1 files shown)


jlengstorf:
@m-allanson this could be an interesting e2e sort of test: loop through showcase sites and check for Gatsby headers?

m-allanson:
Yeah! I was thinking it could also run as a CI task when adding new sites.

This particular site was the result of [this Spectrum thread](https://spectrum.chat/gatsby-js/general/is-coolhunting-com-a-gatsby-site~11570eef-2686-4f80-be0b-f2d59612238a), but there could be others that have snuck in to the showcase.

PR #12445, opened 2019-03-10
chore(docs): tweak unit testing to remove invalid props from Link

appleJax:
## Description
I followed the Unit-Testing docs to set up jest with Gatsby, and I ran into the following error related to Gatsby's `Link` component:
```
React does not recognize the `activeStyle` prop on a DOM element.
```

## Changes
Removing the `activeStyle` prop from the `Link` mock fixed the issue for me. Seems like this should be the default setup. I was unable to find any discussion of this issue on the internet, sorry if there's something I'm missing.

### DIFF
--- a/docs/docs/unit-testing.md
+++ b/docs/docs/unit-testing.md
@@ -132,12 +132,14 @@ const gatsby = jest.requireActual("gatsby")
 module.exports = {
   ...gatsby,
   graphql: jest.fn(),
-  Link: jest.fn().mockImplementation(({ activeClassName, activeStyle, to, ...rest }) =>
-    React.createElement("a", {
-      ...rest,
-      href: to,
-    })
-  ),
+  Link: jest
+    .fn()
+    .mockImplementation(({ activeClassName, activeStyle, to, ...rest }) =>
+      React.createElement("a", {
+        ...rest,
+        href: to,
+      })
+    ),
   StaticQuery: jest.fn(),
 }
 ```
--- a/docs/docs/unit-testing.md
+++ b/docs/docs/unit-testing.md
@@ -132,7 +132,7 @@ const gatsby = jest.requireActual("gatsby")
 module.exports = {
   ...gatsby,
   graphql: jest.fn(),
-  Link: jest.fn().mockImplementation(({ to, ...rest }) =>
+  Link: jest.fn().mockImplementation(({ activeClassName, activeStyle, to, ...rest }) =>
     React.createElement("a", {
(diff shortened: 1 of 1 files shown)


docs/docs/unit-testing.md
```diff
@@ -132,12 +132,14 @@ const gatsby = jest.requireActual("gatsby")
...
   graphql: jest.fn(),
-  Link: jest.fn().mockImplementation(({ to, ...rest }) =>
-    React.createElement("a", {
-      ...rest,
-      href: to,
-    })
-  ),
+  Link: jest
+    .fn()
+    .mockImplementation(({ activeClassName, activeStyle, to, ...rest }) =>
```
DSchau: Does it make sense to keep this extension? Not all props valid for gatsby-link are valid for a link, so maybe we just grab the `to` prop?

Is there a case where we'd want additional?

appleJax:
@DSchau I think we should keep forwarding `...rest` because users may want to use `title`, `id`, `className`, `data-testid`, or other valid custom attributes.

I added a new commit that removes additional invalid props:
- `getProps`
- `innerRef`
- `ref`
- `replace` (causes an error if using the boolean shorthand `<Link replace />`)

appleJax:
@DSchau what are your thoughts? Do these changes seem reasonable? ⬆️

appleJax:
@pieh Hey Michal, how've you been? Can you take a look at this PR?

DSchau:
@appleJax taking another look now! Thanks for the reminder!

docs/docs/unit-testing.md
```diff
@@ -132,12 +132,24 @@ const gatsby = jest.requireActual("gatsby")
...
-    React.createElement("a", {
-      ...rest,
-      href: to,
-    })
-  ),
+  Link: jest
+    .fn()
+    .mockImplementation(
+      ({
+        activeClassName,
```
DSchau:
```suggestion
        // these props are invalid for an `a` tag
        activeClassName,
```

appleJax:
@DSchau thanks for taking a look! That's a great idea to have an official `gatsby-test-utils` package.

There's one failing test that seems unrelated to this PR. Is there anything I can do to resolve this?

DSchau:
Thanks for this PR!

Sorry about the CI failure here--we'll get that fixed up as soon as possible!

appleJax:
@DSchau awesome! Thank you, I appreciate it! 🎉

PR #12456, opened 2019-03-10
docs: Replace npm run develop with gatsby develop

jonathanconway:
This is necessary for Netlify to run locally.
See: https://github.com/netlify-templates/gatsby-starter-netlify-cms#access-locally

### DIFF
--- a/docs/docs/sourcing-from-netlify-cms.md
+++ b/docs/docs/sourcing-from-netlify-cms.md
@@ -77,7 +77,7 @@ collections:
       - { name: body, label: Body, widget: markdown }
 ```
 
-Then in your terminal run `npm run build` to build the site then `npm run develop` to start the
+Then in your terminal run `gatsby develop` to start the
 Gatsby development server. Once the server is running, it will print the address to open for
 viewing. It's typically `localhost:8000`. Open that in a browser and you should see the text
 "Hello World" in the top left corner. Now navigate to `/admin/` - so if your site is at
--- a/docs/docs/sourcing-from-netlify-cms.md
+++ b/docs/docs/sourcing-from-netlify-cms.md
@@ -77,11 +77,11 @@ collections:
       - { name: body, label: Body, widget: markdown }
 ```
 
-Then in your terminal run `npm run develop` to start the Gatsby development server. Once the server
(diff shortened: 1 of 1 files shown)


sidharthachatterjee:
@jonathanconway It isn't necessary to build the site to be able to run `develop`

Am I missing something?

polarathene:
You've mixed things up a bit. It should stay as it is presently.

What you want to do is instead link to the link you've given in your PR about running it locally. Or mention instead of `gatsby develop`, you now need to run `gatsby build` followed by `gatsby serve`. Your changes imply that anyone reading should be developing for Netlify CMS locally. `gatsby develop` also will invalidate the prior build afaik?(since you suggest that instead of `serve`).

docs/docs/sourcing-from-netlify-cms.md
```diff
@@ -77,11 +77,11 @@ collections:
       - { name: body, label: Body, widget: markdown }
 ```
 
-Then in your terminal run `npm run develop` to start the Gatsby development server. Once the server
-is running, it will print the address to open for viewing. It's typically `localhost:8000`. Open that
-in a browser and you should see the text "Hello World" in the top left corner. Now navigate to
-`/admin/` - so if your site is at `localhost:8000`, go to `localhost:8000/admin/`. **The trailing
-slash is required!**
+Then in your terminal run `npm run build` to build the site then `npm run develop` to start the
```
DSchau:
```suggestion
Then in your terminal run `gatsby develop` to start the
```

polarathene: The PR is trying to add documentation for working with Netlify CMS locally(vs a remote CMS with local Gatsby development if I understand correctly. However the author has mistakenly mixed up build/serve calls as build/develop (which afaik invalidates the build output?).

It should be left as it was previously, with a comment linking to the docs the author provided here, or otherwise mention the build/serve workflow when using Netlify CMS locally? (I might have misunderstood the linked docs, I don't use Netlify CMS, but the instructions clearly make a distinction between develop and build/serve).

LekoArts:
Hey @jonathanconway 	👋 	
Do you have time to address the comments made here in this PR?

jonathanconway:
Sorry for the delay; will address these soon; thanks for everyone’s feedback.

PR #12461, opened 2019-03-10
fix(gatsby): Emit BOOTSTRAP_FINISHED when bootstrap finishes

stefanprobst:
Currently we are only emitting `BOOTSTRAP_FINISHED` when the job queue is empty when bootstrap finishes, but not in the Promise.

EDIT: Should be emitted before or after `onPostBootstrap` API?

Also: might be a "good first issue" to reduce code duplication in this part a bit?


freiksenet:
I feel it should be before `onPostBootstrap` because it's "post bootstrap".

stefanprobst:
I agree - it's just that in master we emit *after* `onPostBootstrap`, so I wasn't sure if that's on purpose

sidharthachatterjee:
Published in gatsby@2.1.28

PR #12264, opened 2019-03-03
[wip] feat(www): Spring cleaning, episode 1: Drop „Spectral“ in favor of native sans-serif; moar 

fk:
I like Spectral. I suggested [Spectral to replace TeX Gyre Schola](https://github.com/gatsbyjs/gatsby/pull/1765). But something never sat right with it with me — it runs a little too wide, and IMO renders a little to dark on certain devices and OSes, among other things.

I've also grown a little tired of the serif/Futura-PT-for-inline-links combination; IMO it adds unnecessary visual noise esp. on small screens.

I def. will miss Spectral Italic, but I propose dropping Spectral in favor of the native __sans__-serif font stack. I think Spectral served us well, but not as a `bodyFontFamily`. We have a lot of "UI" needs on gatsbyjs.org, and a sans-serif default just makes sense IMO. We can work towards scoping CSS for long-form content and make it easier to bring back bespoke styles for this type of content, e.g. a (better) serif.

We also mix in Futura PT on i.e. https://www.gatsbyjs.org/features/ and the homepage for regular copy. That's fine per se, but in context things feel messy. Under the hood things really are quite messy, too — _we accumulated a lot of local styling for things that should be defined globally_. This makes it harder for everyone "doing the right thing", to review contributions, and ultimately building and maintaining a consistent UI.

So aside from dropping Spectral and making the required adjustments to do so — which would have caused quite a bit of noise/diff on its own — this PR aims to make a start at cleaning those things up, too, and establishing a basic set of design tokens for spacing and typography.
Thus this PR is a big, ugly blob, and I really don't want to ask anyone to really review the code. But `pliiiis` "Just trust me!" 😅— there's a lot of good things in here! __Nothing will explode.__ There will be follow up PRs, and I'm of course here to address any regression ASAP.  

Here's whats in the bag … roughly:

- simplify link styles — no more Futura PT, no more bold, no more three, but one pixel underline; `colors.lilac` instead of dark gray
- use `monospaceFontStack` for the PrismJS code highlight block badges
- only scale up `baseFontSize` once, at `presets.Xxl`, to 18px — no more 17 and 21px, and adjust for that — `presets.scale` _not_ included yet
- lighten `colors.gray.copy` a tiny bit
- add `presets.space`: it's coupled to` typography.js` (which might be a bit silly), and given `baseLineHeight = 1.5` and `baseFontSize = 16px` it provides the following spacing units:
  - `0`
  - `0.25rem` / `4px`
  - `0.5rem` / `8px`
[...]

PR #12464, opened 2019-03-10
feat(www): Spring cleaning episode 2

fk:
Picking up where #12264 left off:

- `presets.scale` _almost_ everywhere
- less `boxShadow: none` and `:hover { background: "none" }` overrides for links — we don’t need to overwrite this style anymore
- add `presets.letterSpacings` and use them
- add `colors.white` and use it
- import `radii` directly instead of doing `presets.radii[x]`
- rename `presets.animation` to `presets.transition`
- import `transition` directly instead of doing `presets.transition[x]`
- add `presets.shadows.cardActive`
- more `presets.space` refactoring
- remove Futura PT `specimen_files`, demo HTML, and „How to use web fonts“
- first improvements to the „Starters“ search input styles
- simplify things here and there 😉


fk:
Another huge changeset which I really don't want to ask anyone to review — so, as in #12264, allow me to ask you "Just trust me!" 😅 
__Nothing will explode.__

Another final rinse — taking care of a few spacing leftovers, buttons, box-shadows, and a bit of consolidation for "Showcase" and "Starters", and the transition to design tokens should be done! 🤞

fk:
Thank you Dustin! 🤗
This is good to merge, builds fine locally and is ready for production.

The bit you quoted unfortunately isn‘t in this one yet — it will be the upcoming, hopefully last „spring cleaning“ bit. Gotta admit I thought I‘d be faster 🤷‍♂️😶

Issue #12254, opened 2019-03-03
If gatsby-image already has an item in cache, don't show the placeholder


KyleAMathews:
You can see this happening on this blog post. Click to it and back several times. The blurred up base64 image is shown every time.

https://www.gatsbyjs.org/blog/2019-03-01-localization-with-gatsby-and-sanity/

polarathene:
By "in cache" is this referring to the browser cache for images or the lookup object/dict `imageCache` in `gatsby-image`? If the latter, that state would not persist would it? 

Also note the `src` value used as a key is not equivalent to the image elements `currentSrc` that the browser sets(eg for webp when supported), not that it matters for a cache key(it does if you want to check browser cache, but making the relevant ref available when cache is checked in the components constructor isn't an option afaik?

If ignoring the `currentSrc` issue above, the browser cache could be checked and work with single image format sources?

polarathene:
I have spent time looking into resolving this. ~~Considered Cache API and types of persistent storage, realizing they are not suitable(invalid if viewport width has changed requiring a different source size).~~

~~I came across the `currentSrc`  check for cache I mentioned earlier, and I've got that working(either via componentDidUpdate, or an additional ref to the `picture` element so the `currentSrc` can be properly checked/queried once the `picture` element is mounted/visible.~~

## A) Prior browser cache check logic via Promise

Previously, I was doing a more thorough check of the browser cache via a promise that would wait about 10ms to check the browser cache(0ms for in-memory cache, but disk-cache on SSD with my system would require up to 10ms, on slower systems or disk access, that could be even longer making it less reliable:

```JS
// A function outside of `Image` component
const isImageCached = (currentSrc) => {
  return new Promise((resolve, reject) => {
    // empty string if not in the browser cache
    if (currentSrc.length > 0) {
        const image = new window.Image();
        image.src = currentSrc;

        const checkBrowserCache = () => {
          const isCached = image.complete
          image.src = ""

          resolve(isCached)
        }

        // Image should be loaded via browser cache within 10ms
        setTimeout(checkBrowserCache, 10);
    } else { 
      resolve(false) 
    } 
  })
}

// Within `handleRef()`:
const img = this.imageRef.current
isImageCached(img.currentSrc).then(isCached => 
    this.setState({ imgCached: isCached })
)
```

[...]

PR #8350, opened 2018-09-20
[gatsby-image] Restore placeholder image transition

stefanprobst:
In #7083 the css `transition` was moved [from Img](https://github.com/gatsbyjs/gatsby/blob/bcf3af57f3b6438424b712ce1785a6a423b937da/packages/gatsby-image/src/index.js#L117) to [imageStyle](https://github.com/gatsbyjs/gatsby/blob/5ffb1307c17bef802026d5992e1ad000ff654f11/packages/gatsby-image/src/index.js#L224), leaving [imagePlaceholderStyle with a transitionDelay](https://github.com/gatsbyjs/gatsby/blob/5ffb1307c17bef802026d5992e1ad000ff654f11/packages/gatsby-image/src/index.js#L217), but without an actual transition.
Although I don't follow everything that changed with that commit, I think that wasn't on purpose. @tbrannam can you confirm?

PR #4639, opened 2018-03-21
Omit undefined attributes from noscript images

ooloth:
To solve issue #4317, this sets the default values of the `<noscript>` image's `height` and `width` attributes to null (rather than `""`). 

The current `width`/`height` defaults of `""` fail HTML validation because these attributes (if present) must be non-negative integers. Omitting these non-required attributes when no width/height is provided (by setting them to null) solves the validation issue.

PR #4105, opened 2018-02-17
Make Gatsby Image more configurable

tsriram:
Closes #3561

PR #12468, opened 2019-03-11
feat(gatsby-image): don't fadein image when already loaded "browser-cache"

### LINKED ISSUE #10937: [gatsby-image] Transparent PNGs always fade-in
## Description

When using a PNG with a transparent background, there's a blurred fade-in effect even with `fadeIn` set to `false`.

### Steps to reproduce

Add a PNG with an alpha channel to the page using ``.

### Expected result

There should be no fading when the page loads or when navigating between pages.

### Actual result

A blurred version of the image is visible for a short period of time and then fades out.

Here's an example with a version of the `gatsby-astronaut.png` image that's had the background made transparent. This shows what the fade looks like when the page is initialized (just navigating between cached pages here).

untitled

### Environment
```
  System:
    OS: macOS Sierra 10.12.6
    CPU: (8) x64 Intel(R) Core(TM) i7-4850HQ CPU @ 2.30GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 8.10.0 - ~/.nvm/versions/node/v8.10.0/bin/node
    Yarn: 1.12.1 - /usr/local/bin/yarn
    npm: 5.6.0 - ~/.nvm/versions/node/v8.10.0/bin/npm
  Browsers:
    Chrome: 71.0.3578.98
    Safari: 12.0.2
  npmPackages:
    gatsby: ^2.0.76 => 2.0.76 
    gatsby-plugin-offline: ^2.0.16 => 2.0.20 
    gatsby-plugin-sharp: ^2.0.14 => 2.0.16 
    gatsby-source-filesystem: ^2.0.8 => 2.0.12 
    gatsby-transformer-sharp: ^2.1.8 => 2.1.9 
```

polarathene:
## Description

Placeholder transition CSS continues to work as normal, only when the image exists in the browser cache are the transition styles removed. This is mainly to address/reduce the undesirable transition for transparent images as #12254 describes.

Additionally: 

- base64 and bgcolor placeholder delays `0.25s` and `0.35s` respectively have been replaced with the `0.5s` duration of the fadein transition.
- base64 transition duration was removed. It's not something that would be witnessed unless using transparent images, in which case it only prolongs the visibility after the main image has transitioned in.

See my comment in the referenced issue for some considerations/variations or further reasoning for the changes made.

`imgLoaded` and `imgCached` could be renamed to `isImgLoaded` `isImgCached`?
`delayHide` maybe should be `delayHideStyle`?

## Related Issues

Fixes: #12254, Fixes #10937


polarathene:
## noscript, removed transition and transition-delay, dropped width/height props in "fixed" image section

noscript version migrated sometime ago via [this PR](https://github.com/tsriram/gatsby/commit/ffe2372a79c266d98f8113a066a642d5e3b5da95)(states due to errors but no citation of what those were, edit: [here it is](https://github.com/gatsbyjs/gatsby/pull/3122) and direct link to [issue/workaround comment](https://github.com/facebook/react/issues/11423#issuecomment-341751071), **[As of late 2018 it is fixed](https://github.com/facebook/react/issues/11423#issuecomment-419370442)**, but only for React 16.5+, [if you use 16.4, you don't get the error but lose lazy loading](https://github.com/facebook/react/issues/11423#issuecomment-417052048).. ). So `gatsby-image` could go back to supporting noscript without a string for the image, probably more maintainable, just need to ensure users are aware to use React 16.5+, **thoughts?**

Previously it used the `` component, which at the time had `opacity` and `transitionDelay` props(although the noscript version never used them afaik), those were brought over into the `noscriptImg` function during the change from using the `` component to returning a regular `` tag string. And then [another PR to address an issue with that new approach](https://github.com/gatsbyjs/gatsby/pull/4639/files) changing the function to resemble what we have presently.

The code for "Fixed" images , also passes in width/height, this is taken from the `image` object(fluid/fixed prop types) width/height fields, although this object is passed in as props anyhow via spread syntax which overwrites the prior width/height props (**just like custom styles could override this PR feature from working correctly, even if not related to opacity transition**).


## CSS Transition on placeholder unintended?

The CSS transition duration for opacity was removed from the default `` styles in [this commit](https://github.com/gatsbyjs/gatsby/commit/5ffb1307c17bef802026d5992e1ad000ff654f11#diff-5941a73c5d794b120294a0429b6351c1R224), it ensured that the transition was only applied to the actual image to fade-in, while the placeholder would delay changing it's opacity until after the fade-in animation completed... however, someone thought this omission of the transition on the placeholder style was an error, but no actual issue/reason justified it being added back in the [PR](https://github.com/gatsbyjs/gatsby/pull/8350). 

Transition delay also seems to have varied. While it's `0.35s` for for `bgcolor`, [it used to also be that for base64 placeholder](https://github.com/gatsbyjs/gatsby/pull/4105/files#diff-5941a73c5d794b120294a0429b6351c1L353), both placeholders have mixed delays at [this commit](https://github.com/tsriram/gatsby/commit/76cda50d75dd9ea8950b74b8ebc97bc0380d2250), just noticed that `bgcolor` has different delay values for fixed/fluid in master currently... I think it's safe to say this was a typo that moved around.

polarathene:
Well this failure is confusing. I had run the tests locally and updated them noting in the relevant commit that it seemed odd the error hadn't been raised before this PR, and now the CI is unhappy with the snapshot(while local is fine with the exact same code).

```
yarn run jest ./packages/gatsby-image
yarn run v1.13.0
$ jest ./packages/gatsby-image
 PASS  packages/gatsby-image/src/__tests__/index.js
  <Img />
    ✓ should render fixed size images (31ms)
    ✓ should render fluid images (6ms)
    ✓ should have correct src, title and alt attributes (21ms)
    ✓ should have correct placeholder src, title, style and class attributes (6ms)
    ✓ should call onLoad and onError image events (14ms)

Test Suites: 1 passed, 1 total
Tests:       5 passed, 5 total
Snapshots:   2 passed, 2 total
Time:        1.072s
Ran all test suites matching /.\/packages\/gatsby-image/i.
Done in 1.68s.
```

cat ./packages/gatsby-image/src/__tests__/index.js:

```JS
// `display: inline`, oddly doesn't show up in the snapshot either?
const setup = (fluid = false, onLoad = () => {}, onError = () => {}) => {
  const { container } = render(
    <Img
      backgroundColor
      className={`fixedImage`}
      style={{ display: `inline` }}
      title={`Title for the image`}
      alt={`Alt text for the image`}
      {...fluid && { fluid: fluidShapeMock }}
      {...!fluid && { fixed: fixedShapeMock }}
      onLoad={onLoad}
      onError={onError}
      itemProp={`item-prop-for-the-image`}
      placeholderStyle={{ color: `red` }}
      placeholderClassName={`placeholder`}
    />
  )

  return container
}
```
cat ./packages/gatsby-image/src/index.js:

```JS
// Img component has the styles that are now in the snapshot and CI should recognize the last 3 lines of style
const Img = React.forwardRef((props, ref) => {
  const { sizes, srcSet, src, style, onLoad, onError, ...otherProps } = props

  return (
    <img
      sizes={sizes}
      srcSet={srcSet}
      src={src}
      {...otherProps}
      onLoad={onLoad}
      onError={onError}
      ref={ref}
      style={{
        position: `absolute`,
        top: 0,
        left: 0,
        width: `100%`,
        height: `100%`,
        objectFit: `cover`,
        objectPosition: `center`,
        ...style,
      }}
    />
  )
})
```

My snapshot was updated by running `yarn run jest -u` like `yarn test` suggests upon failures. I thought I had updated everything correctly and since it "worked on my machine", pushed it. Any ideas why the CI is acting differently?(some sort of cache or different environment?)

polarathene:
`this.state.fadeIn === false` could be added for negating the placeholder delay too? Probably isn't a case where you have `fadeIn` disabled and want to delay the placeholder opacity toggle? That'd cater to #10937 

Still need advice about the failing snapshot test.

polarathene:
@wardpeet Would you be able to review/merge this please? **TL;DR** version is:

Addresses both linked issues:
- Avoid image transition if cached.
- Respect `fadeIn=false` to avoid undesirable transitions with transparent images due to placeholder transition style.

Three main changes(each commit provides more details):
- Check if image is available in the browser cache. (it was more complex but this seems to work fine)
- Removed some dead code related to supporting noscript (we could also revert the change to string approach if React >=16.5 can be enforced)
- Refactor/organize style logic:
  - Move some conditionals out into variables with clear names for what they're checking for.
  - Remove the `0.25s` duration as it was a typo some time ago, and instead
  - Use a shared duration variable to avoid the typo issue in future
  - Placeholder style no longer has a transition, only a delay for opacity toggle. (Doesn't appear to be a good reason to keep an opacity transition effect).

wardpeet:
looking at this right now! thanks for you patience and great comments!

packages/gatsby-image/src/index.js
```diff
@@ -216,7 +217,14 @@ class Image extends React.Component {
           this.props.onStartLoad({ wasCached: imageInCache })
         }
 
-        this.setState({ isVisible: true, imgLoaded: imageInCache })
+        // imgCached and imgLoaded must update after isVisible,
+        // more details in this comments related commit
```
wardpeet: it's probably a good way to put the comment here as well just so we don't forget it in the long run.

packages/gatsby-image/src/index.js
```diff
@@ -250,24 +258,30 @@ class Image extends React.Component {
       itemProp,
     } = convertProps(this.props)
 
+    const shouldDisplay = this.state.imgLoaded || this.state.fadeIn === false
```
wardpeet: not sure but maybe shouldLazyLoad is a better name? not sure if lazyload is the right term here

packages/gatsby-image/src/index.js
```diff
@@ -250,24 +258,30 @@ class Image extends React.Component {
...
+      opacity: shouldDisplay ? 1 : 0,
+      transition: shouldFadeIn ? `opacity ${durationFadeIn}` : `none`,
+      ...imgStyle,
+    }
+
     const bgColor =
       typeof backgroundColor === `boolean` ? `lightgray` : backgroundColor
 
-    const initialDelay = `0.25s`
+    const delayHide = {
```
wardpeet: I would go for you suggestion
```suggestion
    const delayHideStyle = {
```

packages/gatsby-image/src/index.js
polarathene: I think the condition for that boolean gives the context/clarity what it is for, it was just more appropriate to summarize it as a `should` variable where it was used so the code for the style was still easy to read when the logic would otherwise get a bit long/messy.

If `shouldDisplay` doesn't feel right, how about `shouldReveal`? The variable is describing when the image is ready to display by toggling the opacity, a `shouldLazyLoad` seems to describe the wrong relationship, you could try `isLazyLoaded` or `isReady`(since ultimately it's a true value you're after, there isn't really a distinction to lazy-loading or not.

packages/gatsby-image/src/index.js
polarathene: Well, that was the point of the comment there, so it could be looked up via git blame, no one should have a reason to remove the comment knowing it's giving a reason for why the related code is how it is.

I'll update it to not rely on that, it's just longer, which I thought would be preferred to avoid.

packages/gatsby-image/src/index.js
polarathene: Sure!

polarathene:
The Windows unit test is failing unrelated to the contents of this PR:

```
> prettier "**/*.{md,css,scss,yaml,yml}" "--check"

Checking formatting...
docs\blog\2019-03-29-interview-with-david-eads\index.md
Code style issues found in the above file(s). Forgot to run Prettier?
```

packages/gatsby-image/src/index.js
wardpeet: 👍 for shouldReveal! let me change that and merge afterwards!

sidharthachatterjee:
Published in gatsby-image@2.0.37

PR #12469, opened 2019-03-11
feat(remark): Add frontmatter defaults

c0d3d:
Allow users to configure defaults for frontmatter entries. Default
values will be set in the node's frontmatter.

## Description

Added new option to supply default values inside a markdown's frontmatter.

An example use case for this (and the reason I've made it) is to guard page creations for unfinished posts. I want to add an optional `hidden` field, and set it to true for hidden posts. I also don't want to put `hidden: false` in every other post. This solves that issue by allowing me to supply a default of false to that `hidden` field.

Obviously this could be done using the `onCreateNode` API, but I think that this would be useful for others.

### DIFF
--- a/packages/gatsby-transformer-remark/README.md
+++ b/packages/gatsby-transformer-remark/README.md
@@ -22,7 +22,14 @@ plugins: [
       pedantic: true,
       // GitHub Flavored Markdown mode (default: true)
       gfm: true,
-      // Plugins configs
+      // Supply default values for specific frontmatter fields (default: {})
+      frontmatterDefaults: {
+        // Supply default key-value pair for your frontmatters.
+        // The following example would set 'hidden' to false on any
+        // frontmatter that didn't already have a value for 'hidden'
+        hidden: false
+      },
+      // Plugin's configs
       plugins: [],
     },
   },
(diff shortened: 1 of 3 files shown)


c0d3d:
@KyleAMathews, I think you might be the owner on this one. 
If not, please let me know. Not sure who to tag here ... :P

Also not sure on the PR workflow ...

freiksenet:
Hi!

You can solve this problem with our new alpha feature - [schema customization](https://github.com/gatsbyjs/gatsby/issues/12272). You would now need to use alpha version, but we will release it to release channel soon. 

With schema customization you can provide both type definitions that will work even if data isn't consistent and resolvers that can output the default data.

```js
exports.sourceNodes = ({ actions, schema }) => {
  const { createTypes } = actions

  createTypes([
    schema.buildObjectType({
      name: `MarkdownRemarkFrontmatter`,
      fields: {
        // only provide types
        description: "String",
        // provide default value
        title: {
         type: "String!",
         resolve: (parent) => {
            if (parent.title) {
              return parent.title
            } else {
              return "(Untitled)"
            }
         },
      },
      interfaces: [`Node`],
    }),
  ])
}
```

I hope this helps :)

c0d3d:
@freiksenet Thanks!

PR #12470, opened 2019-03-11
Update/fix to "Creating Tags Pages" documentation

shinytoyrobots:
## Description

This updates the markup provided in this documentation (and one minor text tweak to clarify markup change) so that it is not dependent on a GraphQL query to frontmatter.path. This query fails, and @jlengstorf mentioned that this was a bug. I was working through this problem from the gatsby-starter-blog codebase.

Replaced calls to `path` with `slug` and updated the code examples provided (having worked through relevant issues and have tags & categories pages working at https://github.com/shinytoyrobots/shinytoyrobots-dotcom/pull/5).

In addition removed line `filter: { frontmatter: { published: { ne: false } } }` from GraphQL query in tutorial section on creating a Tags index page. This only threw errors and compile failure for me, and this specific documentation page does not appear to be dependent on any other previous tutorial which has required a `published` field, so it seemed unnecessary that it be there.

### DIFF
--- a/docs/docs/adding-tags-and-categories-to-blog-posts.md
+++ b/docs/docs/adding-tags-and-categories-to-blog-posts.md
@@ -54,8 +54,7 @@ Try running the following query in Graph<em>i</em>QL (`localhost:8000/___graphql
   ) {
     edges {
       node {
-        fields
-        {
+        fields {
           slug
         }
         frontmatter {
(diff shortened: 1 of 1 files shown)


pieh:
Thanks!

It wasn't totally wrong - it depended on users having `path` field in their frontmatter (but it defenitely didn't mentiond it). But given that we use `fields.slug` almost everywhere - let's merge this - it should be less confusing.

PR #12472, opened 2019-03-11
fix(gatsby): improve missing page component error message

### LINKED ISSUE #12447: The plugin "theme" created a page with a component that doesn't exist (`path.resolve` doesn't work while `require.resolve` works)
## Description

I am using Gatsby Themes & my folder structure is like 👇

```
packages/
   theme/
      src/
        blog/
        template/
           blog-post.js
site/
    src/
       blog/
```

And in `gatsby-node.js`, I have `createPages` lifecycle method which takes in `component`.

The `component` works only if I use `require.resolve(`./src/templates/blog-post.js`)`

And does not work if I use `path.resolve(`./src/templates/blog-post.js`)`

This is because I don't have `./src/templates/blog-post.js` in `site` directory but I do have it in `theme` directory.

I think this should be mentioned in the docs since the error was confusing me for a while but I watched the live stream by Jason about Gatsby Themes & remembered to replace `path` with `require` & it worked.

For context, the error was `The plugin "theme" created a page with a component that doesn't exist`. Maybe provide a `path` to the component so it's easier to figure it out :)

ChristopherBiscardi:
## Description

Adds the component to the error message so when people `require.resolve` instead of `path.resolve` they have some information with which to debug the location of the missing component.

## Related Issues

Fixes #12447

### DIFF
--- a/packages/gatsby/src/redux/actions.js
+++ b/packages/gatsby/src/redux/actions.js
@@ -192,7 +192,9 @@ ${reservedFields.map(f => `  * "${f}"`).join(`\n`)}
   // component paths.
   if (process.env.NODE_ENV !== `test`) {
     if (!fileExistsSync(page.component)) {
-      const message = `${name} created a page with a component that doesn't exist`
+      const message = `${name} created a page with a component that doesn't exist. Missing component is ${
+        page.component
+      }`
       console.log(``)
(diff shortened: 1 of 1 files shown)


sidharthachatterjee:
Published in gatsby@2.1.28

PR #12473, opened 2019-03-11
fix(gatsby): correct bootstrap emit order

freiksenet:
## Related Issues
#12461

### DIFF
--- a/packages/gatsby/src/bootstrap/index.js
+++ b/packages/gatsby/src/bootstrap/index.js
@@ -488,49 +488,19 @@ module.exports = async (args: BootstrapArgs) => {
 
   let onEndJob
 
-  const checkJobsDone = _.debounce(resolve => {
+  const checkJobsDone = _.debounce(async resolve => {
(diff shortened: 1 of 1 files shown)


packages/gatsby/src/bootstrap/index.js
```diff
@@ -539,3 +509,20 @@ module.exports = async (args: BootstrapArgs) => {
     })
   }
 }
+
+const finishBootstrap = async bootstrapSpan => {
+  report.log(``)
```
wardpeet: I would move this under `onPostBootstrap`. The emit is used internally to save our redux state which should happen when bootstrap is completely finished. onPostBootstrap is the last hook that users can fix/update anything they want. For gatsby this is part of the bootstrap process so it should be after from a user perspective it doesn't make much sense as it's 'postBootstrap'.

Does this make sense?

sidharthachatterjee:
Published in gatsby@2.1.28

Issue #12446, opened 2019-03-10
Bug: Gatsby netlify-cms can't access admin screen on latest version


Mustaflex:
## Summary
Hi guys, I need some help. I'am new to gatsby and the static world. I was following the gatsby/netlify-cms tutorial and I'm stuck early on. 
When I run "npm start"  everything  runs correctly, I get the message that netlify-cms is running on localhost:8000/admin/ and the admin files are properly generated in  the public/admin folder.
but when I hit the url, I get a 404.
I can only access the admin screen if I run gatsby build and gatsby serve.
running on  (windows) the latest version of gatsby,  gatsby-plugin-netlify-cms, netlify-cms.
Thank you
### File contents (if changed)

`gatsby-config.js`: 
```
backend:
  name: test-repo

media_folder: static/assets
public_folder: assets

collections:
  - name: blog
    label: Blog
    folder: blog
    create: true
    fields:
      - { name: path, label: Path }
      - { name: date, label: Date, widget: date }
      - { name: title, label: Title }
      - { name: body, label: Body, widget: markdown }
```

`package.json`: 
```
   "gatsby": "^2.1.27",
    "gatsby-plugin-netlify-cms": "^3.0.12",
    "netlify-cms": "^2.6.0",
    "react": "^16.8.4",
    "react-dom": "^16.8.4"`

```

Mustaflex:
Update

I tested with this npm packages:
```
 "gatsby": "2.0.0",
    "gatsby-plugin-netlify-cms": "^3.0.12",
    "netlify-cms": "^2.6.0",
    "react": "^16.8.4",
    "react-dom": "^16.8.4"
```

and It worked, I can access the admin page when running in dev mode.

erikdstock:
Can confirm this just happened to me while updating dependencies - 2.1.22 works :) Thank you @Mustaflex

KyleAMathews:
We recently disallowed loading html during development https://github.com/gatsbyjs/gatsby/pull/12243

Netlify CMS is the only tool I know of with this pattern of loading another app during development. Which we didn't think about while making this change :-/

Options are plugins spinning up another server on a different port for serving their apps or we could add an API for plugins to set routes like this as exceptions.

pieh:
> Options are plugins spinning up another server on a different port for serving their apps or we could add an API for plugins to set routes like this as exceptions.

[...]

PR #12474, opened 2019-03-11
fix(gatsby-plugin-netlify-cms): Add listener for /admin

sidharthachatterjee:
Fixes https://github.com/gatsbyjs/gatsby/issues/12446

### DIFF
--- a/packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
+++ b/packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
@@ -39,7 +39,7 @@ exports.onCreateDevServer = ({ app, store }) => {
       path.join(program.directory, `public/admin/index.html`),
       err => {
         if (err) {
-          res.status(500).end()
+          res.status(500).end(err.message)
         }
       }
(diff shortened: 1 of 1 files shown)


packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
```diff
@@ -32,6 +32,20 @@ function deepMap(obj, fn) {
...
 }
 
+exports.onCreateDevServer = ({ app, store }) => {
+  const { program } = store.getState()
+  app.get(`/admin`, function(req, res) {
+    res.sendFile(
+      path.join(program.directory, `public/admin/index.html`),
+      err => {
+        if (err) {
+          res.status(500).end()
```
wardpeet: perhaps expose the err here?
```suggestion
          res.status(500).end(err.message)
```

we could also do check if the file exists before using sendFile and warn or error the user in console.

sidharthachatterjee:
Published in gatsby-plugin-netlify-cms@3.0.13

PR #12482, opened 2019-03-11
test(gatsby-plugin-flow): add tests for flow plugin

ollelauribostrom:
## Description
Adding tests for `gatsby-plugin-flow`. 

#### Summary of changes
- Adding a test case to verify that onCreateBabelConfig sets the correct babel preset.
- Minor refactor: use `export const` instead of `exports.` and use arrow functions.


pieh:
The `ci/circleci: starters_validate` seems like flake, let me restart it just so checks are nice and green

PR #12495, opened 2019-03-11
feat(gatsby-link): adds support for partiallyActive=true to Link

### LINKED ISSUE #7208: Add option to gatsby-link for matching active when on child routes
## Summary

Would be great to mirror reach router's `isCurrent` and `isPartiallyCurrent` options to `gatsby-link`s `activeStyle` and `activeClassname` props. Basically I'd want to be able to designate a Link as active when any child routes are also active, without having to bypass gatbsy-link and use `isPartiallyCurrent` directly. 

### Basic example

Hesitant to suggest just tacking on another prop as lazy DX, but that would probably make the most sense. Something like this:

```js
<Link ... activeStyle={{...}} greedyActive={true} />
```

Would default to false, so non-breaking change that doesn't really add much cognitive overhead for people.

### Motivation

Common use-case is top-level nav for any nested content (eg: a blog), where you'd often want to designate the nav item active (eg: 'Blog') when on a child path (eg: `/blog/some-post`).

If this isn't added, would at least be worth documenting how to achieve by accessing react router APIs directly in gatsby-link docs, as I imagine it's a common pattern.

arcanis:
## Description

Implements a new option in the `<Link/>` component: `partiallyActive={bool}`. If true, the route will activate its `activeStyle` and `activeClassName` parameter even if only partially active.

I've put the settings on the component rather than the route, as there might be places where you want this settings to be on and others where you want it to be off. Also it was simpler 🙂 

I've also moved `@reach/router` into the peerDependencies (from the regular dependencies) since it's a stateful package (due to the context ref) and you don't want to run the risk of having version conflicts. It shouldn't be a problem because the `gatsby` package already provides `@reach/router`.

## Related Issues

Fixes #7208

### DIFF
--- a/docs/docs/gatsby-link.md
+++ b/docs/docs/gatsby-link.md
@@ -80,28 +80,27 @@ const SiteNavigation = () => (
 
 ### Show active styles for partially matched and parent links
 
-<iframe title="Screencast on egghead of how to style partially matched links in Gatsby." class="egghead-video" width=600 height=348 src="https://egghead.io/lessons/egghead-customize-styles-for-partially-matched-urls-with-gatsby-s-link-component/embed" />
-
-Video hosted on [egghead.io][egghead].
-
-The `activeStyle` or `activeClassName` prop are only set on a `<Link>` component if the current URL matches its `to` prop _exactly_. Sometimes, we may want to style a `<Link>` as active even if it partially matches the current URL. For example:
+By default the `activeStyle` and `activeClassName` props will only be set on a `<Link>` component if the current URL matches its `to` prop _exactly_. Sometimes, we may want to style a `<Link>` as active even if it partially matches the current URL. For example:
 
 - We may want `/blog/hello-world` to match `<Link to="/blog">`
 - Or `/gatsby-link/#passing-state-through-link-and-navigate` to match `<Link to="/gatsby-link">`
 
-In instances like these, we can use [@reach/router's](https://reach.tech/router/api/Link) `getProps` API to set active styles as follows:
+In instances like these, just add the `partiallyActive` prop to your `<Link>` component and the style will also be applied even if the `to` prop only is a partial match:
 
 ```jsx
 import React from "react"
 import { Link } from "gatsby"
 
-const partiallyActive = className => ({ isPartiallyCurrent }) => ({
-  className: className + (isPartiallyCurrent ? ` active` : ``),
-})
-
-const PartiallyActiveLink = ({ className, ...rest }) => (
-  <Link getProps={partiallyActive(className)} {...rest} />
-)
+const Header = <>
+  <Link
+    to="/articles/"
+    activeStyle={{ color: "red" }}
+    {/* highlight-next-line */}
+    partiallyActive={true}
+  >
+    Articles
+  </Link>
+</>;
 ```
 
 ### Pass state as props to the linked page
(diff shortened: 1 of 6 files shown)


arcanis:
Here's the cherry! I won't have the time to make an egghead video, though 😄

DSchau:
```
Successfully published:
 - gatsby-link@2.0.16
 - gatsby@2.1.31
```

dustinhorton:
curious about that gatsby-link version—I'm on latest Gatsby (`2.1.35`), and `partiallyActive` is not working, and if I look at my lockfile, I'm only on gatsby-link `2.0.6`.

Are we not supposed to import gastby-link via Gatsby anymore?

lmcjt37:
@dustinhorton try running `npm update -S gatsby` this updated my Gatsby package and also included the addition of `partiallyActive`. Fixed my issue where the prop didn't exist.

dustinhorton:
@lmcjt37 what's the `-S` flag? not seeing it @ https://docs.npmjs.com/cli/update.html

lmcjt37:
@dustinhorton sorry my fault, that's second nature to add the --save flag. `npm update gatsby` should update your package. But if you're on 2.1.35 then you should be able to use the `partiallyActive` prop now?

dustinhorton:
Now on `2.2.9`, and it has gatsby-link `2.0.16` but `partiallyActive` still isn't working for me.

Tried with both `activeClassName` and `activeStyle` props.

lmcjt37:
@dustinhorton Do you have an example of how you're using Gatsby-Link? Maybe I can help identify the issue

dustinhorton:
@lmcjt37 Thanks for the offer—caused me to take _another_ fresh look and I realized the error in my implementation (just some leftover destructuring from when I was handling it myself).

lmcjt37:
@dustinhorton Awesome! Glad you managed to fix it 👍

imshuffling:
Thanks for doing this.

As I'm bringing my navigation in dynamically I had to add a conditional to the activeClassName to stop the homepage menu item being rendered as active.

Leaving this here as it might help someone.

```                
<Link partiallyActive={true} activeClassName={slug === '/' ? null : 'active'}>
   {item}
</Link>

**Comment by gpspake** [comment]
I run in to a similar issue where the home `/` link is always active when partially active is set to true.

**Comment by gpspake** [comment]
I ran in to a similar issue where the home `/` link is always active when `partiallyActive` is set to true but my navigation is kinda fancy so I had a slightly different fix. Mine ended up looking like
```jsx
<Link
  key={page.name}
  className={navigationStyles.navLink}
  to={page.link}
  partiallyActive={true}
  activeStyle={page.link === '/' && pathname !== '/' ? {} : { background: "#0084B4" }}
  >
  {page.name}
</Link>
```
It aint pretty but it works...

Here's the navigation component where I used it, for context. I also had to import `Location`:
```jsx
import React from 'react'
import { Link, StaticQuery } from "gatsby"
import navigationStyles from "./navigation.module.scss"
import { Location } from '@reach/router';

const Navigation = props => {
  const { menuLinks } = props.siteMetadata
  const { pathname } = props.location

  return (
    <div className={navigationStyles.siteNavigation}>
      <nav >
        {menuLinks.map(page => (
          <Link
            key={page.name}
            className={navigationStyles.navLink}
            to={page.link}
            partiallyActive={true}
            activeStyle={page.link === '/' && pathname !== '/' ? {} : { background: "#0084B4" }}
          >
            {page.name}
          </Link>
        ))}
      </nav>
    </div>
  )
}

export default () => (
  <Location>
    {locationProps => <StaticQuery
      query={graphql`
      query {
        site {
          siteMetadata {
            menuLinks {
              name
              link
            }
          }
        }
      }
    `}
      render={data => <Navigation {...locationProps} siteMetadata={data.site.siteMetadata} />}
    />}
  </Location>
)
```
Thanks for the tip, @imshuffling

PR #12496, opened 2019-03-11
chore(docs): fix grammatical error in docs index page

ananta:


marcysutton:
Yep, that reasoning makes sense to me!

Issue #11696, opened 2019-02-11
gatsby-source-drupal - allow json api filter parameters


apmsooner:
jsonapi module allows query parameters to filter list like: /jsonapi?filter[some_field]=some_value

It would be nice option if gatsby-source-drupal could accommodate this so one drupal site could store data for a number of gatsby sites and the gatsby sites could just get the data they need. An example would be a filtering nodes by a taxonomy term so it might mean gatsby processing 100 nodes instead of 10,000.

pieh:
How would that work - should this apply same query params to all requests?

Right now, we hit main endpoint to get list of available endpoints and then make requests for those - should all of those requests have same query params? Would this potentially cause using taxonomy term filter for endpoint where it doesn't make sense (i.e. on content type that doesn't use this taxonomy) and potentially "nulling" the output of it? (I don't have enough context on drupal jsonapi module)

Just for testing the idea you could "monkey patch" `node_modules/gatsby-source-drupal/gatsby-node.js` file and find `axios.get(<some_path>, { auth: basicAuth })` (should be 2 of them) and add `params` there:

`axios.get(<some_path>, { auth: basicAuth, params: { "filter[some_field]": "some_value } })`

and see what happens

apmsooner:
Yeah, I was thinking in gatsby-config.js in line with how you wrote out the example being able to specify:
```
resolve: `gatsby-source-drupal`,
      options: {
        baseUrl: `http://somesite.com/`,
        apiBase: `jsonapi`,
        parameters: {},
        basicAuth: {},
      },
```
If the filter field isn't present or the value isn't found, yeah... null the results. I'll look into trying what you described. Seems like it should be doable and would really speed up the build process when dealing with a backend site containing alot of data.

apmsooner:
Oh... wait now i know what you are saying cause it would affect ALL endpoints within the jsonapi path. I saw another issue request to ask for multiple json api endpoints. I think that would have to happen first make this doable.

pieh:
Just to check we are on the same page I'll walk you through how currently drupal plugin works:

 1. First we hit `http://domain.example/jsonapi` which give us list of dedicated endpoints for each content types / taxonomies and so on - response looks something like this: https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-source-drupal/src/__tests__/fixtures/jsonapi.json
[...]

PR #12508, opened 2019-03-12
feat(gatsby-source-drupal): Add ability to use JSON API filters when querying

eojthebrave:
## Description

Adds the ability to use JSON API filters when querying Drupal. This can be configured per JSON API collection in your gatsby-config.js file. This allows a developer to selectively reduce the amount of data that needs to be pulled from Drupal. It potentially reduces build time for large data sets. And allows developers to ignore data they don't need.

Example configuration:

```javascript
module.exports = {
  plugins: [
    {
      resolve: `gatsby-source-drupal`,
      options: {
        baseUrl: `https://live-contentacms.pantheonsite.io/`,
        apiBase: `api`,
        filters: {
          // collection : filter
          recipe: "filter[tags.name][value]=British",
        },
      },
    },
  ],
}
```

Would result in the "recipe" collection being queried like so `https://live-contentacms.pantheonsite.io/api/recipes?filter[tags.name][value]=British` while leaving all other collections unfiltered.

## Related Issues

Fixes https://github.com/gatsbyjs/gatsby/issues/11696 based on suggestions made by @apmsooner.

### DIFF
--- a/packages/gatsby-source-drupal/README.md
+++ b/packages/gatsby-source-drupal/README.md
@@ -34,6 +34,46 @@ module.exports = {
 }
 ```
 
+### Filters
+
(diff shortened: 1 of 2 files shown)


sidharthachatterjee:
Published in gatsby-source-drupal@3.0.28

PR #4813, opened 2018-04-03
add combined date/time formats for ISO 8601 that allow for local time

grantly7:
Currently **local time** ISO 8601 formats are not identified as a "date" and thus aren't eligible for the formatString() argument in Gatsby's GraphQL queries. This allows the use of both Local Time and UTC dates and resolves issue #4789

A brief recap from ISO 8601 [Wiki](https://en.wikipedia.org/wiki/ISO_8601):

> Time zones in ISO 8601 are represented as **local time** (with the location unspecified), as UTC, or as an offset from UTC.

> If the time is in UTC, add a Z directly after the time without a space. Z is the zone designator for the zero UTC offset.

Includes the local date/time combination formats (no time zone designator on the end) within the ISO_8601_FORMAT const. This way we ensure local date/time combinations can be received / manipulated by the .utc() method as well.

PR #12511, opened 2019-03-12
Add microsecond support for ISO 8601 datetime formats

himynameistimli:
## Description
When attempting to use `formatString` to format dates in GraphQL, we get "Invalid date", which I believe is due to the precision format of seconds.

From what I've seen in other issues/PRs, it seems to be related to the list of formats in type-date.js file. It seems to only have ISO formats with millisecond precision (`YYYY-MM-DDTHHmmss.SSS`). We're using a python backend api for page generation and their default precision is in microseconds (`YYYY-MM-DDTHHmmss.SSSSSS`).

(ref: https://docs.python.org/3.7/library/datetime.html)

The proposed fix is to add microseconds support for UTC and local time.

## Related Issues

Related to #4813

### DIFF
--- a/packages/gatsby/src/schema/types/type-date.js
+++ b/packages/gatsby/src/schema/types/type-date.js
@@ -22,6 +22,8 @@ const ISO_8601_FORMAT = [
   `YYYY-MM-DDTHHmmss`,
   `YYYY-MM-DDTHH:mm:ss.SSS`,
   `YYYY-MM-DDTHHmmss.SSS`,
+  `YYYY-MM-DDTHH:mm:ss.SSSSSSS`,
+  `YYYY-MM-DDTHHmmss.SSSSSSS`,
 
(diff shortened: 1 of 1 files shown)


sidharthachatterjee:
This looks great to me, @himynameistimli 

Worth adding a few tests?

freiksenet:
We are working on big changes to GraphQL in https://github.com/gatsbyjs/gatsby/pull/11480. We plan to merge it next week. I wonder if you could either wait until next week and make a new PR against master or just make the PR for the schema refactor branch. Otherwise there will be merge conflicts that might result in this code getting lost.

Thanks!

himynameistimli:
Alright let me PR it against the schema refactor branch for now.

himynameistimli:
Closing this PR, see https://github.com/gatsbyjs/gatsby/pull/12533

PR #12515, opened 2019-03-12
chore(showcase): add seomonitor.com

alentodorov:
Built with GatsbyJS, hosted on netlify and with contentful as an CMS.


wardpeet:
seems like you have a linting issue

PR #12516, opened 2019-03-12
feat(publish): clear built files before building package and packing

### LINKED ISSUE #12491: [fix]: run git clean before publishing
## Description

To avoid issues like #12489, we should bake in support for running `git clean -xfd` before publishing.

This will clean all gitignored files and give a clean slate so we don't have to worry about issues like these popping up again.

## Implementation

We should edit [the `bootstrap` script](https://github.com/gatsbyjs/gatsby/blob/master/package.json#L67) to run `git clean -xfd` before bootstrapping, so that you are always given a clean slate. The publish workflow is currently:

- `yarn bootstrap`
- `yarn run publish`

So baking it into the bootstrap step makes sense. It should be done in the `prebootstrap` script, so that it can be done before actual bootstrap tasks.

Longer term - it'd be nice to move to a CI publish workflow, but we're not quite there yet!

### Steps to reproduce

Publish with an unclean package (e.g. gatsby-plugin-emotion with extraneous files)

### Expected result

Files that should be gitignore'd should be ignored

### Actual result

Files are not gitignored, and misc. files can make their way into a publish

pieh:
This adds extra step before publishing to clear any previously built files.

Initially I added prompt to confirm file deletion just in case script need to be adjusted and problems can be caught

Example output:
https://gist.github.com/pieh/1ebb938eb300aaac882491bd94a3506f

Btw - this also catch some weirdness in our package build scripts - like `graphql-skip-limit` packs both `src` and `dist`, or testfixtures being packed in `gatsby-codemods`

Closes #12491


scripts/clear-package-dir.js
```diff
@@ -0,0 +1,137 @@
...
+    execSync(
+      `node_modules/.bin/lerna changed --json --loglevel=silent`
+    ).toString()
+  )
+  const filesToDelete = _.flatten(changed.map(getListOfFilesToClear))
+
+  if (!argv[`dry-run`] && filesToDelete.length > 0) {
+    if (
+      argv[`force`] ||
+      PromptUtilities.confirm(`Are you sure you want to delete those files?`)
```
DSchau: This returns a Promise! (ran into this locally, doesn't actually wait for me to prompt and deletes the files).

Annoyingly, we could make this whole thing async--not sure if there's a better solution, e.g.

```js
async function clearPackgeDir() {
 // all this logic, with the one await block
}

clearPackageDir()
```

pieh: ops, yeah, will wrap entire thing in async function, seems like easiest solution here

pieh:
I'll close this is this has potential for problems and comment what I think is more straight forward solution in issue

pieh:
Argh, my bad, solution I thought of is not really possible (just output built files to `lib`/`dist` and rimraf it on builds - as gatsby-X.js files need to be in root of package), so reopening this...

wardpeet:
Any updates on this @pieh?

pieh:
No updates on this yet, let me sync this branch and address Dustin's comments. And should be good to go I think

pieh:
All right, updated the PR and is ready for review

Issue #9828, opened 2018-11-08
gatsby-plugin-netlify v2 excludes some link headers


tbrannam:
## Description

Since migrating to Gatsby V2 - it appears that fewer JS files are eligible to be be considered for preloading when generating the `_headers` file.

### Steps to reproduce

Add gatsby-plugin-netlify to gatsby-config, no options are necessary.

Build project and view `public/_headers`

Compare results of `/` with the <link> headers in `public/index.html`

In Gatsby V1 - more JS bundles matched - so fewer v2 JS files are being http2/pushed

### Expected result

A higher number of JS bundles should match the Link Header filter.

```
        <link as="script" rel="preload" href="/1-6b41147d931e1c49c090.js"/>
        <link as="script" rel="preload" href="/component---src-pages-index-js-9b3d4efb1774bd71ce96.js"/>
        <link as="script" rel="preload" href="/app-c4e3b0537e6949a7f629.js"/>
        <link as="script" rel="preload" href="/webpack-runtime-6a1e34ce8603dc702023.js"/>

```

### Actual result

Typically only two javascript files are linked.  In Gatsby V2 - I would also suggest that /static/*.json files should also be eligible (  `<link rel="preload" href="/static/d/688/path---index-6a9-DFSIFoYuSMh60BvyPO97Ulcr4c.json" as="fetch"/>` )

example:
```
/
  Link: </webpack-runtime-c05d24c23b76e621b476.js>; rel=preload; as=script
  Link: </1-6b41147d931e1c49c090.js>; rel=preload; as=script
```

### Environment

```

  System:
    OS: macOS High Sierra 10.13.6
    CPU: x64 Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.9.0 - ~/.nvm/versions/node/v10.9.0/bin/node
    Yarn: 1.9.4 - /usr/local/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v10.9.0/bin/npm
  Browsers:
    Chrome: 70.0.3538.77
    Firefox: 62.0
    Safari: 12.0.1
  npmPackages:
    gatsby: ^2.0.31 => 2.0.32 
    gatsby-cli: ^2.4.2 => 2.4.3 
    gatsby-image: ^2.0.13 => 2.0.16 
    gatsby-plugin-google-tagmanager: ^2.0.5 => 2.0.5 
[...]

PR #12521, opened 2019-03-12
fix(gatsby-plugin-netlify): add all .js files from webpack.stats.json to _headers

mihaiblaga89:
## Description

The issue was that only the first file was taken from each key from webpack.stats.json. Modified it to get all of them and render them accordingly.

## Related Issues

Fixes to #9828


lablancas:
Looking forward to getting this fix. Thanks @mihaiblaga89

pieh:
Hey @mihaiblaga89! Thanks for your patience with us on this.

Here's example `_headers` output with this change:
```
/
  Link: </webpack-runtime-38690a1d9ceb63d36823.js>; rel=preload; as=script
  Link: </styles.1025963f4f2ec7abbad4.css>; rel=preload; as=script
  Link: </styles-565f081c8374bbda155f.js>; rel=preload; as=script
  Link: </app-143317de0d2e0a4686a0.js>; rel=preload; as=script
  Link: </0-84a0c17fb8865c06616b.js>; rel=preload; as=script
  Link: </component---src-pages-index-js-34bdaacb8d3687c4c6c6.js>; rel=preload; as=script
```
It definitely fixes the problem of grabbing just first chunk. But there is something to potentially adjust:
- I think we need to filter out `css` because we do inline it, so preloading css file shouldn't be necessary (I need to verify this). If we do need it in the end we would need to adjust `as=script` for css files, but this is TBD if it's needed.

mihaiblaga89:
@pieh Sorry for the late reply, was off the grid for a bit. Thank you too for the filtering, you're right, css is inline.

PR #12523, opened 2019-03-12
feat(www): Spring cleaning episode 3

fk:
Picking up where #12464 left off:

- fix letter-spacing for homepage newsletter `<Name>`("THE GATSBY NEWSLETTER")
- fix a couple `key` issues for `/features/`
- homepage blog post card hover now matches `/blog/`
- expand `presets.shadows` and consolidate various `box-shadow` styles (blog post cards, "Showcase" modal, "Showcase" and "Starters" list items, "Docs/Tutorial/Features" sidebar, probably more)
- slightly bump social icon size in main navigation
- add `<Avatar>` component and use it
- adjust "blog post card" avatar image size to 32px
- consolidate "blog post card“ GraphQL query, extend and reuse fragment from `<BlogPostPreviewItem>` for `homepage-blog-post`
- bits of cleanup for the „Starters“ details CSS
- add arrow icon to the „Curious yet?“ Sections „Get Started“ button, and make the latter `large`
- position `.gatsby-highlight`s language badge via `space` and consistently across breakpoints and changing parent container padding
- minor adjustments to the Homepage diagram
- more `presets.space`
- fix .gatsby-remark-responsive-iframe margin bottom


fk:
This is good to go — thought I'd keep it below 50 files this time… 😅 🤐 
I thought I'd get where I want this to be in three runs, but I have to admit the trilogy needs a part four. :/

_But_ we're _close_ — only a couple linering `px` and `rem`s, and typography/spacing things to straighten out after this one — plus some "easy" refactoring, like putting all breakpoints in `presets` in an object of their own, and figuring out the best way on how to use `presets.space` without requiring `rhythm`.

fk:
GTG! ✌️ 

6ec2b340131d484aaa59e1e98d99bb82

www/src/components/docsearch-content.js
```diff
@@ -4,7 +4,7 @@ export default ({ children }) => (
   // todo settle on one className to facilitate Algolia DocSearch;
   // `post` and `post-body` are only in use for the blog:
   // https://github.com/algolia/docsearch-configs/blob/6c2fd2d71302e3ead9a0a83dd44dca02ae0701e3/configs/gatsbyjs.json#L66-L77
-  <main id={`reach-skip-nav`} className={`docSearch-content post post-body`}>
+  <main id={`reach-skip-nav`} className={`docSearch-content post`}>
```
DSchau: Just so I'm aware, this is the json we're using now via Algolia right?

https://github.com/algolia/docsearch-configs/blob/6c2fd2d71302e3ead9a0a83dd44dca02ae0701e3/configs/gatsbyjs.json

Maybe we update the comment?

www/src/components/evaluation-table.js
```diff
@@ -116,9 +116,9 @@ class EvaluationTable extends Component {
                       <SectionHeaderBottom
                         display={row.node.Subcategory}
                         category={row.node.Subcategory}
-                        key={`section-header-bottom-${i}`}
+                        key={`section-header-${s}-bottom-${i}`}
```
DSchau: Fixing a bunch of errors in the console right? Thank you for doing this!

fk: 👍 Yeah, quite a few :) — btw. would have loved to resolve the last remaining warning for "Features", which is the product of this https://github.com/gatsbyjs/gatsby/blob/1257a225de0bb3be8314ecc5684d1c6635d3ee05/www/src/components/evaluation-table.js#L93-L102 along the way, but gave up after 2 minutes ("illegal nesting: `tbody` in `div`" or something).

Either the scrolling lib mentioned there is still not playing nicely with anchors in tables (@calcsam that was you who wrote that comment, right?), or something with the sidebar "scrollspy" doesn't work correctly. Didn't give moving the `id` around actually, maybe that solves it — I think it's currently attached to a `td`…

www/src/components/docsearch-content.js
fk: Oh yeah, I can see how that's a bit misleading. Great catch! 🙏 

In #12464 I moved both `post` and `post-body` classnames to `docsearch-content` to make a first step to consolidate that stuff, write that comment, and keep track of stuff in one place.

When I was looking at the inline-CSS-in-docsearch-results issue today, I noticed that I had forgotten to actually remove `post-body` in `template-blog-post` — this is why I removed it in `docsearch-content` again in this PR.

Will push an update in a bit sorting this out!

I will give https://community.algolia.com/docsearch/config-file.html a good read tomorrow and add a PR over at https://github.com/algolia/docsearch-configs that finally resolves this: I want to have one selector (`.docSearch-content`) for that stuff (unless you stop me 😉).

fk: Updated in 4b831a9!

PR #12526, opened 2019-03-12
chore(docs): add useStaticQuery to docs for mocking Gatsby

dslemay:
## Description

Adds `useStaticQuery` to the docs for mocking out Gatsby related to unit testing.


dslemay:
That sounds really intriguing and curious to see the discussion on it. Experimenting with some higher level integration tests which may have multiple sub components that are using StaticQuery. I also wonder if that may be a place of either increased documentation and/or related to this package idea.

PR #12530, opened 2019-03-12
feat(www): tagline on plugin library landing page

calcsam:
This is an idea for a tagline for the plugin page to illustrate how versatile the Gatsby plugin ecosystem is.

What do y'all think? @KyleAMathews @fk @marcysutton @amberleyromo?

If we want to ship this we should:

- [ ] turn it off (set to "anything") if there's anything in the search bar to avoid it being distracting
- [x] see if we want to pretty up the design at all (tweak the line spacing on 3rd line, hover on the links)

Screen Recording 2019-03-12 at 03 43 PM


calcsam:
Alternate design, a bit more lightweight (second commit). Instead of `<` and `>` we probably want slightly larger, clickable chevron icons that overlap both lines. But you get the idea.

Screen Recording 2019-03-12 at 03 58 PM

www/src/components/rotater.js
```diff
@@ -0,0 +1,112 @@
...
+    })
+  }
+  
+  incrementItem() {
+    this.setState({ 
+      item: ((this.state.item + 1) % this.props.items.length)
+    })
+  }
+  componentDidMount() {
+    const intervalId = setInterval(() => this.incrementItem(), 5000);
```
wardpeet: I'm wondering if we should work with onAnimationEnd event instead of an interval? If we keep an interval I would suggest to use setTimeout instead as setInterval might not be 5s over time and setTimeout always will be.

marcysutton:
One thing I'm wondering is if there's a way to introduce a stop button so the rotating text doesn't distract people. We could use [prefers-reduced-motion](https://webkit.org/blog/7551/responsive-design-for-motion/) in CSS or JS to respect user preferences in supported browsers...but an additional icon button would help, too.

calcsam:
these seem like good ideas! right now clicking the arrows will stop the rotation. 

@fk offered to take this over to prettify this and get it merged in!

fk:
@wardpeet 

> This looks great! It looks a bit janky to me so I'm thinking of making the widths of the dynamic parts fixed so you get some extra spacing if it's smaller but it won't jump around. Another option which takes a bit more time is to use translates to move the text in front. I can do that in a followup.

💯, https://github.com/braposo/react-text-loop/tree/master/ -like behavior would be great.
To avoid the extra spacing, maybe we could tweak the copy to something that allows us to put an equivalent to the current "Need" on its own line … e.g.

```
Looking for
responsive images?
```

or something?

💯 to everything that @marcysutton said, too — maybe if we don't want to invest in play/pause just hide the whole thing if the user `prefers-reduced-motion`?

wardpeet:
We could go for react-text-loop to get this one going. I'm not the biggest fan as it's quite big https://bundlephobia.com/result?p=react-text-loop@2.0.1 
image 20kb is pretty big in js world

marcysutton:
Whatever we do end up using, I'd love to review it for accessibility prior to shipping as rotating things usually need a little help.

fk:
First measures to get an initial version ready to ship:

- added icons and screenreader text ("Previous", "Next") to the control buttons, turned them into actual `<button>`s (so they are focusable), and updated their position:
  > Instead of < and > we probably want slightly larger, clickable chevron icons that overlap both lines
- added animation of the alternating text's `width`

dee04465-88d7-40c0-b271-04cb5fb32b27

fk:
Simpler:

aac1e22d-5cd2-40c6-812f-e2e456bcebb3

fk:
Wooah thanks @DSchau for the review! Right in time for @wardpeet and me pairing to push this over the finish line. Btw. what's your favorite — minimal or "welled"?

DSchau:
@fk I like the minimal--presuming that's the one without a background and a slight grey border?

But they both look great 💜

fk:
"minimal" is this one:

aac1e22d-5cd2-40c6-812f-e2e456bcebb3

And you are in favor of that one:

 dee04465-88d7-40c0-b271-04cb5fb32b27

DSchau:
Just to be 100% clear I like this one better :) But I defer to you--both are great!

Image

www/src/components/rotator.js
```diff
@@ -142,16 +141,14 @@ class Rotator extends Component {
...
               ref={this.sliderContainer}
+              aria-live={this.intervalId ? `off` : `polite`}
             >
               {!enableSlider ? (
                 <>{text}</>
               ) : (
-                <Slider
-                  items={[text]}
-                  color={`#000`}
-                  getSize={size => this.setState({ size })}
```
fk: Argh! Thanks for catching that @wardpeet! 🙏

fk:
@wardpeet and myself (with the help of @DSchau 🙏) settled on the "minimal" variant, and think it's good to ship as a first version:

<p align=center></p>

Summing up, we modified @calcsam's first shot to

- animate the alternating text's `width` to make the transition less "jumpy"
- play nice on smaller screens (in context of the current slider content)
- update control buttons UI and position according to @calcsam's comment
    > Instead of < and > we probably want slightly larger, clickable chevron icons that overlap both lines
- hopefully be accessible enough to ship — @marcysutton could you please take a look and let us know if we missed anything?
  - turn control buttons into actual `<button>`s (so they are focusable), add icons, screenreader text ("Previous", "Next"), and `aria-controls` attributes
  - honor `prefers-reduced-motion` — if set, we completely disable fade and `width` animations
  - `aria-live` set to either `off` or `polite`, depending on `prefers-reduced-motion`

We didn't look into disabling the slider if there's anything in the search bar because we didn't find it too distracting.

/cc @calcsam

calcsam:
Holy smokes, this looks beautiful @fk!

sidharthachatterjee:
Just leaving to mention that this looks so beautiful! Love it. Great work Flo!

DSchau:
Heads up - this broke SSR because we weren't guarding window. Will PR a quick fix in a sec!

fk:
🙏 @DSchau! (and @Sid, and @calcsam, and @marcysutton, and @wardpeet!)

wardpeet:
Ugh my bad sorry 🙏

PR #12531, opened 2019-03-12
fix(gatsby): properly support --no-color for pretty-error

pieh:
pretty errors currently don't honor `--no-color` switch, because we were calling `prettyError.withoutColors()` on instance from global `gatsby-cli` installation.

This change sets `FORCE_COLOR=0` env var (which I picked up from https://github.com/chalk/chalk#chalksupportscolor , so hopefully it's kind of standard - didn't do much research on that I admit) and add support for that in our `pretty-error` setup


packages/gatsby-cli/src/create-cli.js
```diff
@@ -295,6 +299,7 @@ module.exports = argv => {
       global: true,
     })
     .option(`no-color`, {
+      alias: `no-colors`,
```
pieh: I can remove this, but seems like `chalk` automatically support that as well, so this is just QoL addition (not part of bug fix)

pieh:
Update on this after doing more checking:
We also use webpack plugin to provide "friendly messages", which uses chalk@^1 ... and chalk@^1 will only check if `FORCE_COLOR` is set and will force colors even if it's set to `0` - I opened PR there https://github.com/geowarin/friendly-errors-webpack-plugin/pull/86 to bump chalk version

We could use different env var in meantime, but another problem is that `webpack-dev-middleware` also have colored output ... but it uses `ansi-colors` (alternative to `chalk`), which only checks `FORCE_COLOR` env var - so it trully is whack-a-mole ... unless we provide custom logger to webpack-dev-middleware or maybe just disable logs from that because webpack-dev-middleware logs are those lines:
```
ℹ ｢wdm｣: 
ℹ ｢wdm｣: Compiled successfully.

[...]

ℹ ｢wdm｣: Compiling...
```
which seems unnecessary, given that we use `friendly-errors-webpack-plugin`?

pieh:
On the left - with changes from this PR, on the right - current master (so errors are colorized and output from webpack-dev-middleware - `ℹ` is colorized in `ℹ ｢wdm｣: Compiled successfully.`
Screenshot 2019-03-14 at 23 35 08

packages/gatsby-cli/src/create-cli.js
```diff
@@ -83,6 +83,10 @@ function buildLocalCommands(cli, isLocalSite) {
   function getCommandHandler(command, handler) {
     return argv => {
       report.setVerbose(!!argv.verbose)
+      if (argv.noColor) {
+        process.env.FORCE_COLOR = `0`
```
DSchau: Small comment--but we usually add comments here, since this a Chalk documented variable right?

e.g.

```suggestion
        // disables colors in chalk
        // see: https://www.npmjs.com/package/chalk#chalksupportscolor
        process.env.FORCE_COLOR = `0`
```

packages/gatsby-cli/src/create-cli.js
DSchau: I think this is a good idea!

packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
```diff
@@ -4,7 +4,7 @@ import webpack from "webpack"
 import HtmlWebpackPlugin from "html-webpack-plugin"
 import HtmlWebpackExcludeAssetsPlugin from "html-webpack-exclude-assets-plugin"
 import MiniCssExtractPlugin from "mini-css-extract-plugin"
-import FriendlyErrorsPlugin from "friendly-errors-webpack-plugin"
+import FriendlyErrorsPlugin from "@pieh/friendly-errors-webpack-plugin"
```
DSchau:
```suggestion
// TODO: swap back when https://github.com/geowarin/friendly-errors-webpack-plugin/pull/86 lands
import FriendlyErrorsPlugin from "@pieh/friendly-errors-webpack-plugin"
```

packages/gatsby-cli/src/create-cli.js
pieh: This disables colors in `chalk@2` and in `ansi-colors` - it's problematic in `chalk@1` (which is why I needed to fork `friendly-errors-webpack-plugin` package)

PR #11520, opened 2019-02-03
Adds support for more date/time formats

miktirr:
## Description

Support for `YYYY-MM-DD HH:mm:ss Z` could be useful since this is the format [Jekyll uses](https://jekyllrb.com/docs/front-matter/#predefined-variables-for-posts) in the frontmatter. Supporting the same format would make it easier to move from Jekyll to Gatsby.

It’s not perfectly clear to me if using a space instead of the uppercase T is valid in ISO 8601 format (strictly speaking probably not), but at least [Moment.js regards these as valid date strings](https://momentjs.com/docs/#/parsing/string/) and handles them in the same way as those with the T.

PR #12533, opened 2019-03-13
feat(gatsby): add support for microsecond and alternative iso datetime formats

himynameistimli:
I've pulled this PR against this branch per @freiksenet's request

## Original Description

When attempting to use `formatString` to format dates in GraphQL, we get "Invalid date", which I believe is due to the precision format of seconds.

From what I've seen in other issues/PRs, it seems to be related to the list of formats in type-date.js file. It seems to only have ISO formats with millisecond precision (`YYYY-MM-DDTHHmmss.SSS`). We're using a python backend api for page generation and their default precision is in microseconds (`YYYY-MM-DDTHHmmss.SSSSSS`).

I've gone and added additional support for `YYYY-MM-DDTHHmmss.SSSSSS` and `YYYY-MM-DDTHHmmss.SSSSSSZ` as well as support for `YYYY-MM-DD HHmmss.SSSSSS` and `YYYY-MM-DD HHmmss.SSSSSSZ`

## Notes 
Per @sidharthachatterjee I tried to add some tests. Please forgive, I'm not so familiar with tests in JS, so I tried to follow from elsewhere to put together the tests. If anyone has any advice or suggestions on how I can improve them, please let me know.

One issue that I've found is after adding the support for microsecond, not only do microseconds validate, but nanoseconds (`YYYY-MM-DDTHHmmss.SSSSSSSSS`) and any precision validates (e.g. `2018-08-31T23:25:16.0123456789999+02:00`) as well. I'm not sure how to interpret it, but I've just skipped tests that pass when in theory should fail. I've also commented out explicit nanosecond precision support until the tests can pass.

## Related Issues

Previous PR #12511
Previous related issue #4813
Should also address #11520 @miktirr plz comment


packages/gatsby/src/schema/types/__tests__/date.js
```diff
@@ -0,0 +1,453 @@
...
+const { isDate } = require(`../date`)
+
+// Timestamps grabbed from https://github.com/moment/moment/blob/2e2a5b35439665d4b0200143d808a7c26d6cd30f/src/test/moment/is_valid.js
+
+describe(`isDate`, () => {
+  beforeEach(async () => {
+    // pass
+  })
+
+  it(`supports valid ISO 8601 datetimes`, async () => {
```
harryparkdotio: maybe refactoring these into `it.each` style would be beneficial for test run purposes, since when using `.each` all are run, even if any fail, rather than bailing early

ie.

```js
it.each([
  '1970',
  '2019',
  '1970-01',
  ...
])('should return true for %s', dateString => {
  expect(isDate(dateString)).toBeTruthy();
})
```

packages/gatsby/src/schema/types/__tests__/date.js
```diff
@@ -0,0 +1,453 @@
...
+      `2012-04-01T00:00:00-05:00`,
+      `2012-11-12T00:00:00+01:00`,
+    ]
+    let i
+    for (i = 0; i < tests.length; i++) {
+      expect(isDate(tests[i])).toEqual(true)
+    }
+  })
+
+  it(`supports ISO datetimes without T`, async () => {
```
harryparkdotio: none of these tests are actually using `await` or returning a `Promise`, so you shouldn't use an `async` function

wardpeet:
@himynameistimli thanks so much for recreating this PR on our graphql refactoring branch! We'll have a look shortly.

packages/gatsby/src/schema/types/__tests__/date.js
himynameistimli: Thanks @harryparkdotio 	for the advice, went ahead and modified the tests according to what you mentioned.

freiksenet:
I've rebased in against master and made this target it, because we  merged schema refactoring. I'll merge as soon as this passes CI.

Thanks a lot!

PR #12535, opened 2019-03-13
docs: add lerna badge in README

armujahid:
gatsby is using [lerna](https://lernajs.io/) that's why I think we should add lerna badge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,9 @@
   <a href="https://twitter.com/intent/follow?screen_name=gatsbyjs">
     <img src="https://img.shields.io/twitter/follow/gatsbyjs.svg?label=Follow%20@gatsbyjs" alt="Follow @gatsbyjs" />
   </a>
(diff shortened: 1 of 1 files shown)


wardpeet:
Thanks so much for opening this PR! We don't want to clutter our README with badges of tools we use and want to keep it to a minimal. We only like to show gatsby related badges like npm version, download count, ...

We might want to add more tools to https://github.com/gatsbyjs/gatsby#-thanks-to-our-contributors-and-sponsors. Maybe create an issue for it so we can move the discussion there?

Thanks for using Gatsby 💪

armujahid:
@wardpeet lerna is already mentioned in  https://github.com/gatsbyjs/gatsby#a-note-on-how-this-repository-is-organized so I don't think we should add it again in https://github.com/gatsbyjs/gatsby#-thanks-to-our-contributors-and-sponsors

PR #12543, opened 2019-03-13
fix(gatsby-remark-responsive-iframe): use html node rather than unknown node type

johno:
## Description

The gatsby-remark-responsive-images plugin will now directly
insert an HTML node rather than HAST data attributes. This
should have no effect because we're already embedding HTML
as part of the compilation process.

It will break any potential plugin that might be depending
on the existence of these data attributes, but as far as I
can tell it isn't being used (especially considering the
node type is marked as "unknown").

This makes it easier for other projects like gatsby-mdx
to support existing gatsby-remark-* plugins because then
we can watch for HTML nodes injected into the MDAST
and transform them to JSX.

## Related:

- Downstream issue: https://github.com/ChristopherBiscardi/gatsby-mdx/issues/183
- HTML injection: https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-transformer-remark/src/extend-node-type.js#L489-L492

### DIFF
--- a/packages/gatsby-remark-responsive-iframe/src/index.js
+++ b/packages/gatsby-remark-responsive-iframe/src/index.js
@@ -13,6 +13,8 @@ const isUnitlessNumber = n => {
 const isUnitlessOrPixelNumber = n =>
   n && (isUnitlessNumber(n) || isPixelNumber(n))
 
+const needsSemicolon = s => !/;$/.test(s)
+
 // Aspect ratio can only be determined if both width and height are unitless or
 // pixel values. Any other values mean the responsive wrapper is not applied.
 const acceptedDimensions = (width, height) =>
@@ -32,9 +34,20 @@ module.exports = ({ markdownAST }, pluginOptions = {}) =>
         const height = iframe.attr(`height`)
 
         if (acceptedDimensions(width, height)) {
+          const existingStyle = $(`iframe`).attr(`style`) // Other plugins might set border: 0
+          // so we make sure that we maintain those existing styles. If other styles like height or
+          // width are already defined they will be overridden anyway.
+
+          let fullStyle = ``
+          if (existingStyle && needsSemicolon(existingStyle)) {
+            fullStyle = `${existingStyle};`
+          } else if (existingStyle) {
+            fullStyle = existingStyle
+          }
+
           $(`iframe, object`).attr(
             `style`,
-            `
+            `${fullStyle}
             position: absolute;
             top: 0;
             left: 0;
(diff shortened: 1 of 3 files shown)


packages/gatsby-remark-responsive-iframe/src/index.js
```diff
@@ -32,9 +34,18 @@ module.exports = ({ markdownAST }, pluginOptions = {}) =>
         const height = iframe.attr(`height`)
 
         if (acceptedDimensions(width, height)) {
+          const existingStyle = $(`iframe`).attr(`style`) // Other plugins might set border: 0
+          // so we make sure that we maintain those existing styles. If other styles like height or
+          // width are already defined they will be overridden anyway.
+
+          let fullStyle = ``
+          if (existingStyle && needsSemicolon(existingStyle)) {
```
jlengstorf: Does this mean the `existingStyle` is ignored if it doesn't need a semicolon?

johno: Nice catch, I've amended my PR to address this and added a test case.

packages/gatsby-remark-responsive-iframe/src/index.js
```diff
@@ -13,6 +13,8 @@ const isUnitlessNumber = n => {
 const isUnitlessOrPixelNumber = n =>
   n && (isUnitlessNumber(n) || isPixelNumber(n))
 
+const needsSemicolon = s => !/;$/.test(s)
```
DSchau: Slightly nit:

```suggestion
const needsSemicolon = str => !str.endsWith(';')
```

packages/gatsby-remark-responsive-iframe/src/index.js
```diff
@@ -61,17 +74,8 @@ module.exports = ({ markdownAST }, pluginOptions = {}) =>
           </div>
           `
 
-          node.data = {
-            hChildren: [{ type: `raw`, value: rawHTML }],
-          }
-          // Set type to unknown so mdast-util-to-hast will treat this node as a
-          // div not an iframe — it gets quite confused otherwise.
-          node.type = `unknown`
```
DSchau: @pieh do you have any insight into why this was done?

I validated the new approach with a sample project, and it works just fine setting the type to `html`.

Just entirely unsure if this would break other plugins. Doesn't look like we have other plugins depending on this in _our_ repo, at least.

pieh: Sorry for late reply - I have absolutely no context on this. Maybe some older version of `mdast-util-to-hast` had problems with this? I vaguely remember that we did bump `mdast-util-to-hast` at some point, so it might fix the issue?

DSchau: No problem! I assumed as much--because I validated it worked like we expect with this new approach too, so hopefully no issues here!

DSchau:
Published as [`gatsby-remark-responsive-iframe`@`2.1.0`](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-remark-responsive-iframe/CHANGELOG.md#210-2019-03-14)

PR #12548, opened 2019-03-13
chore(www): add top margin to Return value

sbardian:
## Description

After adding `pluginOptions` as a parameter I noticed `Return value` was right up against `pluginOptions` parameter.  `Return value` did not have the top margin that `Example` has. 

## Related Issues

N/A

### DIFF
--- a/www/src/components/function-list.js
+++ b/www/src/components/function-list.js
@@ -88,7 +88,7 @@ export default ({ functions }) => (
         )}
         {node.returns && node.returns.length > 0 && (
           <div>
-            <h4>Return value</h4>
+            <h4 css={{ marginTop: rhythm(space[5]) }}>Return value</h4>
             {node.returns.map(ret => (
               <div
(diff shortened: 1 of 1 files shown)


sbardian:
See PR #12547 about adding `pluginOptions` to docs.

pieh:
Hey, I will merge https://github.com/gatsbyjs/gatsby/pull/12087 shortly that changes API reference components, but I will handle transplanting your change after that merge

pieh:
Hey, with new api reference components - this is how it looks like (without any styling changes from this PR):
Screenshot 2019-03-14 at 00 53 44

This seems ok to me as-is - let me know if you still would want to apply changes here

sbardian:
Nope, looks good to me too!

pieh:
> Nope, looks good to me too!

All right, so I'll close this one and we can handle #12547 without this being a blocker.

Issue #12549, opened 2019-03-13
[gatsby-plugin-typography] fails to compile when headerComponents include null


danielkov:
## Description

After upgrading from `v1` to `v2` following the official guide I am witnessing the following issue:

```
error UNHANDLED REJECTION


  TypeError: Cannot read property 'key' of null

  - render-page.js:1997
    /Users/danielkov/projects/project/public/render-page.js:1997:11

  - Array.sort

  - render-page.js:1996 Object../node_modules/gatsby-plugin-typography/gatsby-ssr.js.exports.onPreRenderHTML
    /Users/danielkov/projects/project/public/render-page.js:1996:18

  - render-page.js:196
    /Users/danielkov/projects/project/public/render-page.js:196:36

  - Array.map

  - render-page.js:191 ./.cache/api-runner-ssr.js.module.exports
    /Users/danielkov/projects/project/public/render-page.js:191:25

  - render-page.js:597 Module.default
    /Users/danielkov/projects/project/public/render-page.js:597:57

  - worker.js:35 Promise
    [project]/[gatsby]/dist/utils/worker.js:35:36

  - debuggability.js:303 Promise._execute
    [project]/[gatsby]/[bluebird]/js/release/debuggability.js:303:9

  - promise.js:483 Promise._resolveFromExecutor
    [project]/[gatsby]/[bluebird]/js/release/promise.js:483:18

  - promise.js:79 new Promise
    [project]/[gatsby]/[bluebird]/js/release/promise.js:79:10

  - worker.js:31 Promise.map.path
    [project]/[gatsby]/dist/utils/worker.js:31:37

  - util.js:16 tryCatcher
    [project]/[gatsby]/[bluebird]/js/release/util.js:16:23

  - map.js:61 MappingPromiseArray._promiseFulfilled
    [project]/[gatsby]/[bluebird]/js/release/map.js:61:38

  - promise_array.js:114 MappingPromiseArray.PromiseArray._iterate
    [project]/[gatsby]/[bluebird]/js/release/promise_array.js:114:31

  - promise_array.js:78 MappingPromiseArray.init
    [project]/[gatsby]/[bluebird]/js/release/promise_array.js:78:10

```

I realise this is because in the following method:

```js
exports.onPreRenderHTML = function (_ref2) {
  var getHeadComponents = _ref2.getHeadComponents,
      replaceHeadComponents = _ref2.replaceHeadComponents;
  var headComponents = getHeadComponents();
[...]

PR #12555, opened 2019-03-13
fix(gatsby): filter null values in headComponents, preBodyComponents and postBodyComponents

### LINKED ISSUE #12553: [gatsby] headComponents can be empty (null or empty array) values
## Description

This is related to #12549 and as discussed in the comments, sanitizing `headComponents` would ensure that plugins that do similar transformations as the one in that issue can assume that all elements are actual components.

### Steps to reproduce

As of now it is possible to call `setHeadComponents` or `replaceHeadComponents` with any values and there are no sanity checks in place.

### Expected result

Empty values (null, empty array) should be filtered out.

### Actual result

Everything gets added.

### Environment

Not environment specific.

danielkov:
## Description

> This is related to #12549 and as discussed in the comments, sanitizing headComponents would ensure that plugins that do similar transformations as the one in that issue can assume that all elements are actual components.
> -- <cite>#12553</cite>

## Related Issues

Fixes #12553


packages/gatsby/cache-dir/__tests__/static-entry.js
```diff
@@ -61,6 +61,48 @@ const reverseHeadersPlugin = {
...
+    },
+  },
+}
+
+const checkNonEmptyHeadersPlugin = {
+  plugin: {
+    onPreRenderHTML: ({ getHeadComponents }) => {
+      const headComponents = getHeadComponents()
+      expect(headComponents.includes(null)).toBeFalsy()
+      expect(headComponents.find(val => Array.isArray(val) && val.length === 0))
```
pieh: this seems to miss assertion? the `expect()` should be falsy here as well, right?

packages/gatsby/cache-dir/__tests__/static-entry.js
```diff
@@ -147,6 +189,42 @@ describe(`static-entry`, () => {
     })
   })
 
+  test(`onPreRenderHTML can filter out null value`, done => {
+    global.plugins = [setNullHeaderPlugin, checkNonEmptyHeadersPlugin]
+
+    StaticEntry(`/about/`, (_, html) => {
+      expect(html).toMatchSnapshot()
```
pieh: do we need snapshot testing here if we assert head components directly? I would like avoid busy snapshots like that that, especially that I think this snapshot would pass right now without any changes in `static-entry` file, right?

pieh:
It would be great to apply that to `PreBodyComponents` and `PostBodyComponents` as well (would be nice to structure test to not repeat mocked plugins / tests for each of those

packages/gatsby/cache-dir/__tests__/static-entry.js
danielkov: Good catch! It's a midnight-coding error on my part.

packages/gatsby/cache-dir/__tests__/static-entry.js
danielkov: As it stands they do seem to be redundant. I will remove them.

danielkov:
> It would be great to apply that to PreBodyComponents and PostBodyComponents as well

I will do that. I've scanned the official plugins and none of them seems to rely on the current behaviour. However I will browse through some of the popular plugins outside of this repo, to make sure it doesn't regress community packages.

packages/gatsby/cache-dir/static-entry.js
```diff
@@ -49,6 +49,16 @@ const getPage = path => pagesObjectMap.get(path)
 
 const createElement = React.createElement
 
+const sanitizeHeadComponents = components => {
```
wardpeet: you could rewrite this function to something like

```js
const sanitizeHeadComponents = components => {
  const componentsAsArray = Array.isArray(components) ? components : [components];

  return componentsAsArray.filter(Boolean);
};
```
The filter Boolean removes all falsy (undefined, null, fasle) values. I don't think you can have nested arrays.

danielkov: As of right now in my particular example I can get the following, just using Helmet:

```js
[
  [
    {
      $$typeof: Symbol(react.element),
      type: "title",
      key: "",
      ref: null,
      props: [Object],
      _owner: null,
      _store: {}
    }
  ],
  [],
  [],
  [],
  [],
  [],
  null,
  {
    $$typeof: Symbol(react.element),
    type: "link",
    key: "gatsby-plugin-feed-0",
    ref: null,
    props: {
      rel: "alternate",
      type: "application/rss+xml",
      title: undefined,
      href: "/rss.xml"
    },
    _owner: null,
    _store: {}
  },
  // Other component definitions
];
```

They're all valid React Children, so it makes sense. I wanted to use this as a starting point for sanitizing the input.

wardpeet: ok, I missed what you were trying to do. Thanks for clarifying.

packages/gatsby/cache-dir/static-entry.js
```diff
@@ -49,6 +49,16 @@ const getPage = path => pagesObjectMap.get(path)
...
+  if (Array.isArray(components)) {
+    // remove falsy items
+    return components.filter(val => (Array.isArray(val) ? val.length > 0 : val))
+  } else {
```
wardpeet: you can drop this else because of the return

Issue #11667, opened 2019-02-09
request: add warnings when remark prism language not found


cdaringe:
# problem

- typescript was not syntax highlighting, even though supported via prism.
- no errors or warnings are presented even when a failure is occurring

# reproduction

- #1917 explains the failure mode in depth, where i reached the failure by using `ts` after the triple backtick, vs typescript

# request

in the `gatsby-remark-prism`, add helpful logging rather than silently bailing:

```js
  if (!Prism.languages[language]) {
    try {
      console.log(`loading language: ${language}`)
      loadPrismLanguage(language);
    } catch (e) {
      // Language wasn't loaded so let's bail.
      // ^^ REQUEST, don't bail so fast!  log.warn(`unable to find language ${language} for highlighting. applying generic code block`)
      if (language === `none`) {
        return code; // Don't escape if set to none.
      } else {
        return _.escape(code);
      }
    }
  }
```

pieh:
For reference - this is location of code referenced by @cdaringe - https://github.com/gatsbyjs/gatsby/blob/9fdf931f62e1e9e0ea2bebc536cdf780afee2c58/packages/gatsby-remark-prismjs/src/highlight-code.js#L9-L20

I think it makes sense to only warn if can't load language and is different than `none` (as then user opts in to not add highlighting).

This feature has potential of being very spammy so we might want to track if warning was already displayed for `language` before showing it.

connollyf:
Hi, I'd like to try to tackle this issue. Is there a style guide for printing error messages?

cdaringe:
hush hush @gatsbot, the issue is alive and valid

arturfsousa:
Hey guys, I've sent a PR to add the warning #12565. Let me know if you need any changes.

Thank you 😀

PR #12565, opened 2019-03-14
chore(gatsby-remark-prismjs): add warnings when language not found

arturfsousa:
## Description

Adding a warning to the `gatsby-remark-prismjs` plugin when language is not found. The warn pops up only once per language to avoid spamming messages as @pieh commented in the issue #11667.

## Related Issues

Fixes: #11667

### DIFF
--- a/packages/gatsby-remark-prismjs/src/highlight-code.js
+++ b/packages/gatsby-remark-prismjs/src/highlight-code.js
@@ -3,7 +3,7 @@ const _ = require(`lodash`)
 
 const loadPrismLanguage = require(`./load-prism-language`)
 const handleDirectives = require(`./directives`)
-const notSupportedLanguages = []
+const unsupportedLanguages = new Set()
 
 module.exports = (language, code, lineNumbersHighlight = []) => {
   // (Try to) load languages on demand.
@@ -15,12 +15,12 @@ module.exports = (language, code, lineNumbersHighlight = []) => {
       if (language === `none`) {
         return code // Don't escape if set to none.
       } else {
-        if (!notSupportedLanguages.includes(language)) {
+        if (!unsupportedLanguages.has(language)) {
           console.warn(
             `unable to find prism language '${language}' for highlighting.`,
             `applying generic code block`
           )
-          notSupportedLanguages.push(language)
+          unsupportedLanguages.add(language)
         }
         return _.escape(code)
       }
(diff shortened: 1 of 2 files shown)


packages/gatsby-remark-prismjs/src/highlight-code.js
```diff
@@ -3,6 +3,7 @@ const _ = require(`lodash`)
 
 const loadPrismLanguage = require(`./load-prism-language`)
 const handleDirectives = require(`./directives`)
+const notSupportedLanguages = []
```
DSchau: How about we use a Set here?

```suggestion
const unsupportedLanguages = new Set()
```

packages/gatsby-remark-prismjs/src/highlight-code.js
```diff
@@ -14,6 +15,13 @@ module.exports = (language, code, lineNumbersHighlight = []) => {
       if (language === `none`) {
         return code // Don't escape if set to none.
       } else {
+        if (!notSupportedLanguages.includes(language)) {
```
DSchau:
```suggestion
        if (!unsupportedLanguages.has(language)) {
```

packages/gatsby-remark-prismjs/src/highlight-code.js
```diff
@@ -14,6 +15,13 @@ module.exports = (language, code, lineNumbersHighlight = []) => {
...
+          console.warn(
+            `unable to find prism language '${language}' for highlighting.`,
+            `applying generic code block`
+          )
+          notSupportedLanguages.push(language)
```
DSchau:
```suggestion
          unsupportedLanguages.add(language)
```

DSchau: Also maybe we lowercase here? e.g. we wouldn't want to warn for JavaScript vs. javascript.

packages/gatsby-remark-prismjs/src/highlight-code.js
arturfsousa: Nice catch!

packages/gatsby-remark-prismjs/src/highlight-code.js
arturfsousa: Good idea. I will do it.

arturfsousa: Done @DSchau

PR #12575, opened 2019-03-14
feat(www): Spring cleaning episode 4

fk:
(Yawn … I know :/) Picking up where #12523 left off, this:

- groups all breakpoint constants from `presets` in an object `breakpoints` 🎉 😉— e.g. `presets.Lg` becomes `presets.breakpoints.Lg` — doing this not only makes things a little clearer, but also gets us a step closer to discuss things like `styled-system`
- switches the naming scheme around — e.g. `breakpoints.Lg` becomes `breakpoints.lg` and vice versa — this way, the most used `breakpoints` (ready to be used as a mobile-first media query in CSS) are available lowercase
- groups `presets.headerHeight`, `presets.bannerHeight`, etc. in `dimensions`


fk:
<3 🙏

PR #12582, opened 2019-03-14
feat(create-remote-file-node): add `url` field to downloaded file nodes

DarioDiem:


pieh:
`gatsby-source-filesystem@2.0.26` was published with this change

PR #12586, opened 2019-03-14
let shadowed components import parents

ChristopherBiscardi:
## Description

If we have a theme file and a shadow of that file

    
    gatsby-theme-a/src/components/bio.js
    site/src/gatsby-theme-a/components/bio.js
    

the site's bio.js can not import the theme's bio.js. Now it can.

### DIFF
--- a/packages/gatsby/src/internal-plugins/webpack-theme-component-shadowing/index.js
+++ b/packages/gatsby/src/internal-plugins/webpack-theme-component-shadowing/index.js
@@ -35,6 +35,21 @@ module.exports = class GatsbyThemeComponentShadowingResolverPlugin {
       // get the location of the component relative to src/
       const [, component] = request.path.split(path.join(theme, `src`))
 
+      const issuerExtension = path.extname(request.context.issuer)
+
+      if (
+        request.context.issuer
+          .slice(0, -issuerExtension.length)
+          .endsWith(component)
+      ) {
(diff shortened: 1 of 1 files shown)


DSchau:
@jlengstorf 😱 going forward, no big deal in doing this--but if we can it's ideal to update the commit message so we get the CHANGELOG updated appropriately. For this:

`fix(gatsby): for themes, let shadowed components import parents`

jlengstorf:
@DSchau dammit, didn't realize I needed to scope it. Sorry.

DSchau:
@jlengstorf doesn't really _hurt_ anything, just it's a miss on the CHANGELOG! We can update it manually too, no big deal at all!

jlengstorf:
I _did_ add the `feat:` tag (https://github.com/gatsbyjs/gatsby/commit/368c9e0f16a5f6be8b22f27720209f1155d00373). Does that mean it shows up... somewhere? Or does it only work when fully scoped?

DSchau:
@jlengstorf oh interesting, yeah it looks like not including the scope just updates gatsby -> https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/CHANGELOG.md#features

Which is technically what we want here!

jlengstorf:
Hooray! I accidentally did the right thing! 😅

PR #12596, opened 2019-03-15
chore(showcase): Adds www.projectival.de

sascha53:
Adds https://www.projectival.de/


docs/sites.yml
```diff
@@ -4954,17 +4954,6 @@
   built_by: d:code:it
   built_by_url: https://dcodeit.com
   featured: false
-- title: SupremeKun - Street Fashion Marketplace and News
```
DSchau: Why is this removed?

lannonbr: It looks like this is not a Gatsby Site anymore. I'd say the removal would better to be in a separate PR rather than pushed into this PR

DSchau: I should've validated 🤦‍♂️

I'll send a revert

DSchau: Done in #12597

PR #12604, opened 2019-03-15
fix(gatsby): extract queries from themes not starting with "gatsby-theme-*" name

grantglidewell:
## Description

Changes in reference to the comment on line# 239 in `query-compiler.js`
mentioned here: https://github.com/gatsbyjs/gatsby/pull/10786
`resolveThemes` function references themes in the redux store instead of plugins.

## Related Issues

Related to #10786
Addresses https://github.com/ChristopherBiscardi/gatsby-theme-examples/issues/14


packages/gatsby/src/internal-plugins/query-runner/query-compiler.js
```diff
@@ -62,11 +62,9 @@ const validationRules = [
 let lastRunHadErrors = null
 const overlayErrorID = `graphql-compiler`
 
-const resolveThemes = (plugins = []) =>
-  plugins.reduce((merged, plugin) => {
-    if (plugin.name.includes(`gatsby-theme-`)) {
-      merged.push(plugin.resolve)
-    }
+const resolveThemes = (themes = []) =>
+  themes.reduce((merged, plugin) => {
```
pieh: It seems like `themes` in redux state is not array,
but it's object that contains subfield `themes` ( https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/redux/reducers/themes.js )

see https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/webpack-theme-component-shadowing/gatsby-node.js#L14 for example where `themes.themes.map` is used

I'm not sure we need it to be object like that to be honest, but just to get it going, let's adjust here
```suggestion
  themes.themes.reduce((merged, plugin) => {
```

grantglidewell: the payload passed into the store is `themes.themes` which should be an array. However, my recent fix addressing this is still unsuccessful.

PR #12607, opened 2019-03-15
feat(gatsby-plugin-twitter): load widget for all Twitter embeds

tmadeira:
Besides `.twitter-tweet`, there are other elements that Twitter widget renders: timelines, follow and share buttons - as mentioned in [Twitter developer docs](https://developer.twitter.com/en/docs/twitter-for-websites/overview.html).

This small change loads the widget script if _any_ of these elements is present on the page, not only `.twitter-tweet` elements.

### DIFF
--- a/packages/gatsby-plugin-twitter/src/gatsby-browser.js
+++ b/packages/gatsby-plugin-twitter/src/gatsby-browser.js
@@ -27,10 +27,12 @@ const injectTwitterScript = () => {
 
 let injectedTwitterScript = false
 
-const embedClasses = `.twitter-tweet,
-  .twitter-timeline,
-  .twitter-follow-button,
-  .twitter-share-button`
+const embedClasses = [
+  `.twitter-tweet`,
+  `.twitter-timeline`,
+  `.twitter-follow-button`,
+  `.twitter-share-button`
+].join(`,`)
 
 exports.onRouteUpdate = () => {
   // If there's an embedded element, lazy-load the twitter script (if it hasn't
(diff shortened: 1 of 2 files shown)


packages/gatsby-plugin-twitter/src/gatsby-browser.js
```diff
@@ -27,10 +27,15 @@ const injectTwitterScript = () => {
 
 let injectedTwitterScript = false
 
+const embedClasses = `.twitter-tweet,
+  .twitter-timeline,
+  .twitter-follow-button,
+  .twitter-share-button`
```
pieh: I wonder if 
```js
const embedClasses = [
  `.twitter-tweet`,
  `.twitter-timeline`,
  `.twitter-follow-button`,
  `.twitter-share-button`
].join(`,`)
```
would be more readable?

On initial check I didn't notice `.twitter-tweet` being still there because it wasn't aligned with rest of selectors

tmadeira: Sure! Fixed that.

pieh:
Published `gatsby-plugin-twitter@2.0.13`

alvincrespo:
@pieh @tmadeira Do you know if we should probably be including this right before </body>? I ask because I can't seem to get the nicely formatted widget to be styled correctly on my blog here: https://blog.alvincrespo.com/til-episode-4/

The PR where I did this work is here: https://github.com/alvincrespo/blog/pull/46

I'm unsure what might be the issue. Any help would be awesome!

alvincrespo:
oops, it's coming through now - guessing something on Twitter's side.

PR #12615, opened 2019-03-16
[schema] Update docs for new documentationjs version

stefanprobst:
This is for #11480 

Adapt API docs for documentationjs update (#12087)

What is not yet working:
* arrays in mixed type params, i.e. while `@param {string}` works and `@param {string[]}` works, this fails: `@param {(string | string[])}`
* `@returns {Promise<Node[]>}`

### DIFF
--- a/packages/gatsby/src/redux/actions.js
+++ b/packages/gatsby/src/redux/actions.js
@@ -1173,14 +1173,16 @@ actions.addThirdPartySchema = (
   }
 }
 
+import type GatsbyGraphQLType from "../schema/types/type-builders"
 /**
  * Add type definitions to the GraphQL schema.
  *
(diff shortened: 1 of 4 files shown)


pieh:
> * arrays in mixed type params, i.e. while `@param {string}` works and `@param {string[]}` works, this fails: `@param {(string | string[])}`
> * `@returns {Promise<Node[]>}`

This is mostly because right now we rely on inferring and if we didn't have any jsdocs that used this particular mix of expressions, then we couldn't edit queries/fragment to query for nested types (as this would break queries). I want to add typedefs to documentation plugin to be able to query those fields.

stefanprobst:
@pieh Adjusting the `DocumentationTypeFragment` like [this](https://github.com/gatsbyjs/gatsby/compare/master...stefanprobst:api-docs-support-array-types) seems to work for both the Promise return and the array type params.

PR #12620, opened 2019-03-16
only check for update when not on CI

mathieudutour:
## Description

When running Gatsby on a CI, the prompt to update it is annoying. Moreover, if it fails (because CI env are weird sometimes), the entire command with:

```
error There was a problem loading the local build command. Gatsby may not be installed. Perhaps you need to run "npm install"?

┌────────────────────────────────────────────────────────────────┐
│ gatsby-cli update check failed │
│ Try running with sudo or get access │
│ to the local update config store via │
│ sudo chown -R $USER:$(id -gn $USER) /home/sbx_user1051/.config │
└────────────────────────────────────────────────────────────────┘
```

which is a really weird error message, especially considering the beginning of it.

This PR makes sure that the check for updates is skipped on CIs

### DIFF
--- a/packages/gatsby-cli/package.json
+++ b/packages/gatsby-cli/package.json
@@ -21,6 +21,7 @@
     "fs-exists-cached": "^1.0.0",
     "fs-extra": "^4.0.1",
     "hosted-git-info": "^2.6.0",
+    "is-ci": "^2.0.0",
     "lodash": "^4.17.10",
     "meant": "^1.0.1",
     "opentracing": "^0.14.3",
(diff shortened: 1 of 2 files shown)


wardpeet:
Thank you for creating this Pull Request. If you have any troubles with getting these message on ci I would refer you to https://github.com/yeoman/update-notifier/.

Update-notifier is also using the [is-ci package](https://github.com/yeoman/update-notifier/blob/master/package.json#L40).

It seems like I've provided enough information for those reasons I'm closing this PR. 

If this is not the case, or if we can help further--please don’t hesitate to reach out or comment on this PR, and we’d love to take another look.

Thanks for using Gatsby 💪

mathieudutour:
Ah interesting, I thought that was the case but then my CI failed because of that and I noticed `npm` is using `is-ci` manually as well: https://github.com/npm/cli/blob/86f5bdb91f7a5971953a5171d32d6eeda6a2e972/bin/npm-cli.js#L85.

Must have been something I didn't configure well

PR #12603, opened 2019-03-15
feat(www): Spring cleaning episode 5

fk:
Attention, noooooooooob Friday stuff … potential public humiliation stuff 😅🥳

Since we want to define `presets.space` in relation to Typography.js’ `rhythm/baseLineHeight`, but doing so directly in `presets`  would have introduced a circular dependency among `utils/presets` and `utils/typography` (from what I _think_ I understand 🤓), up to now we had to put values from `presets.space` through `rhythm` for each usage in components:

```jsx
import { space } from "../../utils/presets"
import { rhythm } from "../../utils/typography"

<div css={{
  padding: `${rhythm(space[7])} ${rhythm(space[4])} ${rhythm(space[4])}`
}} />
```

This caused a lot of unnecessary noise, and was an ugly derivation from the norm for that one group of design tokens in `presets` … and AFAICT smells a lot.

This PR cleans that up — as good as I can. I would appreciate any input and am happy to adjust this, crossing fingers that this PR at least shows where I want things to go:

```jsx
import { space } from "../../utils/presets"

<div css={{
  padding: `${space[7]} ${space[4]} ${space[4]}`
}} />
```

TL; DR:

- minimize direct usage of Typography.js’ `scale` and `rhythm`
- maximize usage of `presets` tokens for consistency, and make that easy and as concise as possible

—

What's in the bag:

- split up design tokens in `utils/presets` and move them to `utils/tokens/*` — theme-style FTW?!
- move `utils/colors` to `utils/tokens/colors`
- refactor to remove circular dependency among `utils/typography` and `utils.presets` — this enables us to use Typography.js’ `rhythm` in `utils.presets` so we can avoid manual „composition“ of `space` values in components
- move `paddingLeft` helper function used in `components/sidebar/accordion` and `components/sidebar/section-title` to `utils/sidebar/indention`
- adjust changes from #12087 to use `space` design tokens instead of Typography.js' `rhythm`
- move font stack definitions to tokens, and use `presets.fonts` in components instead of `typography.options` and `join(`,`)
- realigned/uncluttered docs/tutorial/features, plugins sidebars:
  image . 
  image
  - opt out of styling scrollbars in WebKit in favor of user settings on Mac OS X (where users can pick among always show scrollbars, show them depending on mouse/trackpad, show them while scrolling — our WebKit scrollbar CSS interfered with the latter two), and a perfectly fine scrollbar on Windows that users know and recognize
  - move padding to anchor/title — allows for full-width background/hover, etc.
[...]

PR #12622, opened 2019-03-16
Add a nav link to the Gatsby Starters

Jared-Dev:
## Description

I consistently have to do a search via a search engine to get to the
starters. Using the Algolia search ends up taking me to the docs
which will finally get me to the starters. I find the starters to be a
huge benefit to Gatsby and think having a nav link directly to them
would be a benefit to people starting out with Gatsby (or that just
want an easier way to get to them)!


www/src/utils/presets.js
```diff
@@ -9,8 +9,8 @@ module.exports = {
     sm: `@media (min-width: 550px)`,
     Md: `(min-width: 750px)`,
     md: `@media (min-width: 750px)`,
-    Lg: `(min-width: 1000px)`,
-    lg: `@media (min-width: 1000px)`,
+    Lg: `(min-width: 1050px)`,
```
sidharthachatterjee:
```suggestion
    Lg: `(min-width: 1000px)`,
```

Can we avoid changing the breakpoint?

www/src/utils/presets.js
```diff
@@ -9,8 +9,8 @@ module.exports = {
...
+    lg: `@media (min-width: 1050px)`,
```
sidharthachatterjee:
```suggestion
    lg: `@media (min-width: 1000px)`,
```

www/src/utils/presets.js
Jared-Dev: You can, but it causes a horizontal scroll-bar.

fk:
Good call @Jared-Dev!

FWIW, we discussed grouping "Plugins" and "Starters" in a new (already existing) landing page "Ecosystem" (https://www.gatsbyjs.org/ecosystem/) last year, which could replace the current "Plugins" item in the main navigation, and would resolve the space issue — but I can see how it helps having both "Plugins" and "Starters" in the main navigation. I'll bring that question up with the team again!

www/src/utils/presets.js
fk: I agree with @sidharthachatterjee — we shouldn't adjust the breakpoint constant, but instead opt-out of using it to define when the mobile navigation is hidden/shown, and instead use a bespoke media query (`@media (min-width: [whatever value we actually need]`).

fk: On another note, I worked on #12603 over the weekend, which (among other stuff) changes how the main navigation items look, which gains us a bit of space:

image

@sidharthachatterjee #12603 is ready to go — maybe we merge that, and then check how things look here with the "Starters" item added? Happy to merge `master` here.

Jared-Dev: Dropping the uppercase does seem to fix the spacing issue even with the breakpoint reset to `1000px` and leaving starters in place.

Jared-Dev:
Not sure how I missed that! 

I didn't even realize Starters existed until @kkemple introduced them to me (I was looking for a TypeScript based setup). I personally would prefer a direct link in the main nav! If you guys have an A/B framework in place however it could be a really interesting test to see how the two options get used though (Plugins + Starters vs. Ecosystem).

KyleAMathews:
I'd rather not add more items to the top nav — there's quite a few there already.

An alternative proposal https://github.com/gatsbyjs/gatsby/issues/12767

LekoArts:
@KyleAMathews Can we close this PR as #12767 was fixed?

KyleAMathews:
Yup! Thanks @LekoArts for following up on this and for others who worked on fixing this issue.

Issue #12619, opened 2019-03-16
Accessible name of gatsbyjs.com link is unclear


dengeist:
## Description

The gatsbyjs.com link (visible label `.com`) is a little confusing. 

### Steps to reproduce

1. Go to www.gatsbyjs.org (or any path in the .org site)
2. Use a screen-reader to listen to the `.com` nav link
3. OR observe the label as a sighted keyboard/mouse user.

### Expected result

Whether visually or aurally, the visible label of the link should clearly tell users its destination.

### Actual result

The destination is unclear. Which `.com` does the link go to? Users _might_ be able to infer from context that, if they are on `gatsbyjs.org`, they will be taken to `gatsbyjs.com`, but clarity here will help cognitively disabled or distracted users. Also, speech dictation users may ask their tools to "click on `.com`" and be confused when nothing happens.

Using `title` to further describe the link is helpful, but:  
- the delay in announcing the title could cause screen-reader users to miss that extra info
- mouse users may not know they can hover to see a title 
- sighted keyboard users can never see the title.

### Suggestion

Unsure! Maybe use `gatsbyjs.com` as the visible label of the link, and use a title to optionally explain further if necessary.

PR #12625, opened 2019-03-17
Adds aria-label to gatsbyjs.com element

papistan:
Addresses issue #12619  Accessible name of gatsbyjs.com link is unclear

## Description

Adds aria-label
## Related Issues

Addresses issue #12619

### DIFF
--- a/www/src/components/navigation.js
+++ b/www/src/components/navigation.js
@@ -162,6 +162,7 @@ const Navigation = ({ pathname }) => {
           <SocialNavItem
             href="https://www.gatsbyjs.com"
             title="gatsbyjs.com"
+            aria-label="gatsbyjs.com"
             overrideCSS={{ paddingRight: 0, fontSize: presets.scale[1] }}
           >
             .com


wardpeet:
shouldn't we add it to all SocialNavItems?

cc @marcysutton

papistan:
Debatable, since for this issue screen readers take text within elements first - therefor for this element reading ".com" instead of reaching for the title like it would in the twitter or github element. For this, case an aria-label is needed. For others is a separate issue.

dengeist:
Thank you so much for tackling my issue, @papistan! I think I explained myself poorly in my ticket, so I wanted to chime in again here.

I should have said that the **visible label** should be `gatsbyjs.com`. Changing the ARIA label will help screen-reader users, but sighted users might still be a bit confused. The visible text `.com` doesn't really explain _what_ website the link goes to. And, for that matter, speech-dictation users may dictate "Click on `.com`" and be confused when nothing happens. Pursuant to [WCAG 2.1 criterion 2.5.3 (Label In Name)](https://www.w3.org/TR/WCAG21/#label-in-name) the accessible name should probably start with or match the visible label.

I'm guessing that the site originally went with `.com` because it fits nicely at all screen sizes, and changing the visible label would require working with designers so the new visible label looks good everywhere, so I appreciate you and the community taking this on. Thanks again, and sorry for the confusion I caused. :)

www/src/components/navigation.js
```diff
@@ -162,6 +162,7 @@ const Navigation = ({ pathname }) => {
           <SocialNavItem
             href="https://www.gatsbyjs.com"
             title="gatsbyjs.com"
+            aria-label="gatsbyjs.com"
             overrideCSS={{ paddingRight: 0, fontSize: presets.scale[1] }}
           >
             .com
```
muescha: @dengeist - that was you suggestion

```suggestion
             Gatsbyjs.com
```

DanielRuf: Afaik the `title` is already used then. It is either label or title used.

marcysutton:
As discussed here, there are accessibility issues with the visible `.com` text differing from both the link's accessible name (provided by `aria-label` in this PR) and description (the existing `title` solution).

Because the header is so limited on space at small viewports, we are also limited on what we can do to fix it. We're discussing an alternative approach like moving that link to a new footer where we could avoid hacks and use the full link text of `gatsbyjs.com` (and we could add other things, like a link to a new accessibility page). So I think closing this PR is a good idea for now since we aren't quite ready to make the more substantial change. Thanks so much for taking the initiative to fix it; I should have more information for you soon.

PR #12626, opened 2019-03-17
Refactor Gatsby Default Starter – SEO page to include og:image and og:share + improve readability

thomaswangio:
## Description

- Added `keywords`, `image` and `url` to  `siteMetadata` in `gatsby-config`
- Added `og:image` and `og:url` to `seo.js` to have a better boilerplate for sharing on Twitter and FB
- Reorganized and declared variables `metaTitle`, `metaAuthor`, `metaKeywords`, `metaImage`, and `metaUrl` to make sure every page will have the these meta tags even if not passed as props to the component. This improves the readability of this file.
- Added corresponding propTypes

Screen Shot 2019-03-17 at 2 41 04 AM
Screen Shot 2019-03-17 at 2 41 38 AM
Screen Shot 2019-03-17 at 2 41 54 AM


thomaswangio:
Note: I made sure to change the `og:url` in `gatsby-config` to https://gatsby-starter-default-demo.netlify.com, which is from the official Gatsby starters page in the [docs](https://www.gatsbyjs.org/docs/starters/#starters). Also, the image url is a dropbox file, but also included in `src/images/gatsby-share.png`.

starters/default/gatsby-config.js
```diff
@@ -3,6 +3,9 @@ module.exports = {
     title: `Gatsby Default Starter`,
     description: `Kick off your next, great Gatsby project with this default starter. This barebones starter ships with the main Gatsby configuration files you might need.`,
     author: `@gatsbyjs`,
+    keywords: [`gatsby`, `application`, `react`],
+    image: `https://www.dropbox.com/s/bhwnfkuwewq03sl/gatsby-share.png?raw=1`,
```
muescha: Image should be in GitHub. Not at dropbox

thomaswangio: @muescha done, just tested with https://raw.githubusercontent.com/gatsbyjs/gatsby/master/starters/default/src/images/gatsby-astronaut.png

After merge, the new raw image link (https://raw.githubusercontent.com/gatsbyjs/gatsby/master/starters/default/src/images/gatsby-share.png) will be pulled from the official gatsby repo.

starters/default/gatsby-config.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
     description: `Kick off your next, great Gatsby project with this default starter. This barebones starter ships with the main Gatsby configuration files you might need.`,
     author: `@gatsbyjs`,
     keywords: [`gatsby`, `application`, `react`],
-    image: `https://www.dropbox.com/s/bhwnfkuwewq03sl/gatsby-share.png?raw=1`,
+    image: `https://raw.githubusercontent.com/gatsbyjs/gatsby/master/starters/default/src/images/gatsby-share.png`,
```
muescha: the image is checked into the repo and should be in the local folder, then a relative path would be enough

thomaswangio: @muescha Yes, but Open Graph doesn't support relative URLs - it's `content` not `src`, and only takes a "string of Unicode characters forming a valid URL having the http or https scheme"

Reference: https://stackoverflow.com/questions/9858577/open-graph-can-resolve-relative-url

thomaswangio: Let me know if it's alright to merge. Or if you have an alternative suggestion that would work.

muescha: but this looks not right. it has every time to download a file from github which already exists in the folder. someone change later the pic in the folder and not understand why it not changed in the app

maybe someone other can help with this case?

note:
on the blog starter the fixed images are places in folder `assets` 
and in the `bio.js` the pic is referenced in a query - so it should be possible to reference the local image

thomaswangio:
Hey @muescha! Thanks for the feedback. I've got something working! (using your suggestion). There is an issue with an updated `package.lock.json` so I created a new PR https://github.com/gatsbyjs/gatsby/pull/12740.

PR #12647, opened 2019-03-18
fix(gatsby-transformer-remark): always include the root node of AST

chitoku-k:
## Description

Fixes a bug in `cloneTreeUntil` function in `gatsby-transformer-remark`, where it only returns the last child node. This function is expected to clone a `root` node, but the current implementation forgets to revert the variable back to the `root` node. Reassigning `clonedRoot = newNode` (originally this is the argument `root`) should fix this. By resolving this bug, it appears to correctly returns the entire root node of AST in `excerptAst` of the markdown node, instead of returning the partial result.

## Related Issues

Related to #11237


DSchau:
```
Successfully published:
 - gatsby-transformer-remark@2.3.7
```

PR #12648, opened 2019-03-18
chore(*): tweak changelog

DSchau:
## Description

Per some discussion on Twitter, the root-level CHANGELOG is oftentimes used and confusion arises because it's out of date.

Let's tweak to just re-direct to the Gatsby package and provide some explanatory information.


pieh:
Can we move old changelog (that is removed here) to the bottom of gatsby package changelog, just so it's not lost? (no re-formatting - just moving content)

KyleAMathews:
Yeah, that's what I was thinking

DSchau:
@KyleAMathews @pieh yeah, great point. Updated!

PR #12659, opened 2019-03-19
chore: Dead code removal

stefanprobst:
[This](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/utils/checkpoints-promise.js) does not seem to be used anywhere anymore (neither does BOOTSTRAP_STAGE)?

Also: do we want to keep [this](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/utils/profile.js)?


pieh:
> Also: do we want to keep [this](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/utils/profile.js)?

I do uncomment and use this occasionally, so I would keep that in.

PR #12669, opened 2019-03-19
docs(gh-pages) Update how-gatsby-works-with-github-pages.md

Hermanya:
Add `.nojekyll` and `--dotfiles`


pieh:
Can you add some context on why this is needed or what does it fix?

pieh:
Also I'm not sure if `touch` will work on Windows if someone doesn't have bash commands installed

PR #12671, opened 2019-03-19
fix(gatsby): keep track of pages created by stateful createPages after edits

### LINKED ISSUE #12143: [V2] Hot reload doesn't work on pages deleted and created by onCreatePage
## Description

Hot reload fails when updating pages deleted and created by `deletePage` and `createPage` inside `onCreatePage`. It renders correctly after `npm start` but it fails to hot reload. Updating any other type of file works correctly.

It actually updates correctly but it attempts to refresh the page after 1 second and then it fails.

The idea is to replace pages inside `src/pages` by their i18n version. So `/` is deleted and replaced by a component which redirects to `/es/` and `/en/`. The relevant code is:

```
exports.onCreatePage = async ({ page, actions }) => {
  const { createPage, deletePage } = actions;

  if (page.path.includes('404') || page.context.locale) {
    return;
  }

  const redirect = path.resolve('src/i18n/redirect.js');

  const redirectPage = {
    ...page,
    component: redirect,
    context: {
      languages,
      locale: '',
      routed: false,
      redirectPage: page.path,
    }
  };

  deletePage(page);
  createPage(redirectPage);

  languages.forEach(({ value }) => {
    const localePage = {
      ...page,
      originalPath: page.path,
      path: `/${value}${page.path}`,
      context: {
        languages,
        locale: value,
        routed: true,
        originalPath: page.path,
      }
    };

    createPage(localePage);
  });
};
```

### Steps to reproduce

1. Clone https://github.com/tcerdaITBA/SurvivalRecipes
2. `git checkout a5dff2c4bf3d4fdcc2a05c59294734832a421d3b`
3. `npm install`
4. `npm start`
5. Navigate to `localhost:8000`
6. Modify `src/pages/index.jsx`

### Expected result

It should render the updated changes without refreshing the page.

### Actual result

404-development page is rendered and restarting application is required to fix.

### Environment

```
  System:
    OS: macOS 10.14.3
    CPU: (4) x64 Intel(R) Core(TM) i5-4250U CPU @ 1.30GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 11.5.0 - ~/.nvm/versions/node/v11.5.0/bin/node
    npm: 6.7.0 - ~/.nvm/versions/node/v11.5.0/bin/npm
  Languages:
    Python: 3.6.4 - /usr/local/bin/python
  Browsers:
    Firefox: 65.0.1
    Safari: 12.0.3
  npmPackages:
    gatsby: ^2.1.18 => 2.1.18 
    gatsby-plugin-favicon: ^3.1.5 => 3.1.5 
    gatsby-plugin-react-helmet: ^3.0.7 => 3.0.7 
    gatsby-plugin-styled-components: ^3.0.6 => 3.0.6 
    gatsby-source-filesystem: ^2.0.23 => 2.0.23 
    gatsby-transformer-remark: ^2.2.0 => 2.2.0 
  npmGlobalPackages:
    gatsby-cli: 2.4.8
```

### Additional Information

This started to happen since Gatsby 2.1.10, related to #11831.

I have already tried fixing the React hot load warning but the error persisted.

Thank you very much for your help.

pieh:
If in `gatsby-node.js` we add
```js
exports.onCreatePage = ({ actions, page }) => {
  actions.createPage(page)
}
```

Next time any node update, pages created by `gatsby-plugin-page-creator` will be deleted. This is regression introduced by fix in https://github.com/gatsbyjs/gatsby/pull/11831 (that fixed the problem of pages never getting deleted)

The problem is that when we do `createPage` by plugin/site that isn't implementing `createPagesStatefully` it will be marked to be deleted.

This change keeps track if original API was `createPages` or `createPagesStatefully` by using `traceId` instead of directly checking what plugin created the page in the end.

Fixes #12143


khades:
Easiest way to break development environment with this regression is to use gatsby-plugin-remove-trailing-slashes.

khades:
It was just released, and it works like charm, thank you.

PR #12672, opened 2019-03-19
feat(gatsby-plugin-netlify-cms): support multiple cms module paths

sondrele:
## Description
This change allows for multiple cms module paths to be imported
and compiled into the netlify cms applicationm generated by this plugin.

It's generally not necessary to need more than one cms module entry in
a simple gatsby-site, but it's convenient to support it in combination with
`gatsby themes`. This makes it possible to create a custom theme that e.g.
sets up the cms and provides it's own custom cms module, while also allowing
theme-consumers to provide their own cms modules with additional config.

Note that the previous config option `modulePath` has been renamed to `modulePaths`, while the plugin is still backwards compatible.
This is maybe a little weird, and personally I would prefer to bump the major version of the plugin and remove the reference to the old `modulePath` config option instead.
I haven't done that since I'm not sure whether this is a wanted feature or not, but would be happy to do so (or any other feedback regarding this feature/API change).

### DIFF
--- a/packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
+++ b/packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
@@ -52,9 +52,10 @@ exports.onCreateWebpackConfig = (
         cms: [
           manualInit && `${__dirname}/cms-manual-init.js`,
           `${__dirname}/cms.js`,
-          modulePath,
           enableIdentityWidget && `${__dirname}/cms-identity.js`,
-        ].filter(p => p),
+        ]
+          .concat(modulePath)
+          .filter(p => p),
       },
       output: {
         path: path.join(program.directory, `public`, publicPathClean),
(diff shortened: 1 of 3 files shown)


packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
```diff
@@ -36,6 +36,7 @@ exports.onCreateWebpackConfig = (
   { store, stage, getConfig, plugins, pathPrefix },
   {
     modulePath,
+    modulePaths = modulePath,
```
pieh: I'm not sure if it's worth to change the field name here (probably could be more confusing for people comparing their current config to documentation when they see that they use `modulePath` and not `modulePaths` but it still works? I would think just adding type and potentially adding note that it can accept single path or array of paths would be good

sondrele: Cool! I've kept the name instead of renaming

PR #12673, opened 2019-03-19
fix: darken link color on homepage diagram

amberleyromo:
@fk I think some of the recent homepage changes resulted in insufficient contrast on the links in the diagram.

I've darkened it a shade for sufficient contrast.

### DIFF
--- a/www/src/components/diagram.js
+++ b/www/src/components/diagram.js
@@ -176,7 +176,16 @@ const ItemDescription = ({ children }) => (
 )
 
 const ItemDescriptionLink = ({ to, children }) => (
-  <Link to={to}>{children}</Link>
+  <Link
+    css={{
+      "&&": {
+        color: colors.gatsbyDark,
+      },
+    }}
+    to={to}
+  >
+    {children}
+  </Link>
 )
 
 const Gatsby = () => (


johno:
Thanks @amberleyromo!!!

> Wondering if @johno and @jxnblk might have built something along styled-system already?

@jxnblk built [colorable](https://github.com/jxnblk/colorable) which enumerates all possible combinations with sufficient contrast and I built [get-contrast](https://github.com/johno/get-contrast) quite a while back which are related, but we don't have anything that truly automates that kind of linting right now in `styled-system`. It's a bit tricky because you don't necessarily know from a Text component what background you're rendering against for testing the contrast in `styled-system`.

We can use the tokens to document which color combinations are acceptable, but it's prolly better to go with [full-fledged a11y testing in CI](https://marcysutton.github.io/a11y-and-ci).

marcysutton:
There is also this resource from Alice Boxhall at Google that could come in handy if we do want to create something (ratio computation and color suggestions): https://github.com/A11yance/a11y-color

fk:
Thanks @johno and @marcysutton!

So much for "thinking" out loud :D — we _do not_ have anything without the whole rendering context. Hands down for CI a11y testing! ✅

PR #12675, opened 2019-03-19
feat(theme): Implement MDX blog theme

johno:
This implements a basic MDX-based blog theme. I'm opening up a PR now to get what's currently implemented into master. I'll be following with a second PR later this week to update this implementation using a data model for blog posts and authors.

---

cc/ @jxnblk @ChristopherBiscardi @jlengstorf


themes/gatsby-theme-blog-mdx/gatsby-config.js
```diff
@@ -0,0 +1,75 @@
+const path = require(`path`)
+
+module.exports = ({ defaultLayouts = {} } = {}) => {
+  const themeLayouts = {
+    posts: require.resolve(`./src/templates/post`),
+    default: require.resolve(`./src/components/layout`),
```
ChristopherBiscardi: since default applies to all MDX files ever, we probably shouldn't use it in themes.

themes/gatsby-theme-blog-mdx/gatsby-node.js
```diff
@@ -0,0 +1,155 @@
...
+      path,
+      context: node,
+      component: Post,
+    })
+  })
+
+  // Create post list pages
+  const posts = mdxPages.edges
+  const numPages = Math.ceil(posts.length / postsPerPage)
+  Array.from({ length: numPages }).forEach((_, i) => {
```
ChristopherBiscardi: can we get rid of pagination by default? Could always offer a helper function.

themes/gatsby-theme-blog-mdx/src/tokens/typography.js
```diff
@@ -0,0 +1 @@
+export { default } from "typography-theme-bootstrap"
```
ChristopherBiscardi: long-term we're going to have to figure out a solution for the "global" typography styles I think.

ChristopherBiscardi:
looks good to me :D 

New data stuff is in now, so it'll be great to apply that to the parent theme and use that for both children

themes/gatsby-theme-blog-mdx/src/components/layout.js
```diff
@@ -0,0 +1,12 @@
+import React from "react"
+import { useTypography, TypographyProvider } from "typography-system"
+
+export default ({ children }) => {
+  const typography = useTypography()
+
+  return (
+    <TypographyProvider theme={typography}>
```
jxnblk: Isn't this context provided with the `MDXThemeProvider`? I also feel like generally `typography-system` is pretty unstable at the moment. It currently uses React's `useEffect` to add the Google font for example

johno: Just pushed a commit that removes this file entirely since it's no longer being used. Should I opt for `@styled-system/typography` in this theme? I'd like to steer towards something more composition-based to ensure as much flexibility as possible.

PR #11361, opened 2019-01-28
feat(docs): Add docs for gatsby-mdx

johno:
At some point in the future we will be bringing [gatsby-mdx](https://github.com/ChristopherBiscardi/gatsby-mdx) into the Gatsby monorepo as an officially supported plugin. As we work on improving/reorganizing/expanding docs for gatbsy-mdx v1 we decided to begin integrating into the .org site since that will be the eventual location for the library.

I've arbitrarily gone with adding MDX section to the Guides portion of the sidebar, though I'm not sure if this is the best location, or if it should even live as a grouping and instead have relevant docs dispersed throughout other sections.

screen shot 2019-01-28 at 10 13 54 am

I think @shannonbux, @jlengstorf, and @marcysutton might have opinions/ideas on this?

## :books: What's included?

• **Introduction:** Outlining what MDX is and why it might be useful for your project
• **Getting started:** Instructions for using a starter and incorporating gatsby-mdx into an existing project
• **Writing MDX:** Brief guide on how to write an MDX document (a bit duplicated from official MDX docs)
• **Programmatically creating pages:** Long form guide showing how to source custom MDX content, create pages, and use templates/layouts
• **Customizing components**: Short form guide on customizing rendering for different MD elements with components
• **Using plugins:** Short form guide on how to using `gatsby-remark-*` plugins with MDX and remark plugins directly

I've opted for writing a full introduction under the assumption that there will be a lot of folks that will be introduced to MDX from the sidebar itself. This way we have a gentle introduction that introduces MDX and then links to the official docs for further reading and exploration.

## ☁️ What's missing?

I think it'd be useful to add in another long-form doc about using composition in MDX to achieve a lot of the features that are currently added in via plugins. Things like:

• Table of contents
• Syntax highlighting
• Coding playgrounds/live editors

This is important because the current practice of reaching for a plugin can be essentially replaced with composition in MDX. However, this then introduces a bit of a problem with MD/remark-focused docs because these approaches both make sense for their contexts, but conflict with each other.

Perhaps it might also be worthwhile to introduce an official MDX starter as well?

## 🤔 Considerations

- The programmatic pages doc has some overlap with [Programmatically create pages from data](https://www.gatsbyjs.org/docs/programmatically-create-pages-from-data/). Should we merge these somehow?
[...]

PR #12679, opened 2019-03-19
Add section to gatsby-remark-smartypants docs about gatsby-mdx

johngrasty:
## Description

The differences in setup with `gatsby-mdx` are not currently mentioned in the docs. This would save a a some effort in tracking down the differences in setup. This might be especially important with the big push I've noticed that you are doing to get more thorough mdx docs available on the Gatsby site. I attempted to _briefly_ address the two main differences, namely the use of `gatsbyRemarkPlugins` in place of `plugins` and the need to use the longer syntax even if you aren't setting any options.

It's your project, but obviously feel free to modify this PR however best suits your needs, or go about it an a completely different way, but I do think it would be helpful for the information to be along with the plugin since it is used with both remark and mdx. Thanks!

### DIFF
--- a/packages/gatsby-remark-smartypants/README.md
+++ b/packages/gatsby-remark-smartypants/README.md
@@ -45,3 +45,26 @@ plugins: [
   },
 ]
 ```
+
+### Use and Options with `gatsby-mdx`
+
+The configuration is slightly different when you use it with `gatsby-mdx`. Here is an example like above with the same `oldschool` options, but notice that you write `gatsbyRemarkPlugins` in place of `plugins` and you have to use the longer resolve form even if you don't have any options.
+
+```javascript
+// In your gatsby-config.js
+plugins: [
+  {
+    resolve: "gatsby-mdx",
+    options: {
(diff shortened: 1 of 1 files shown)


pieh:
I do think this should be documented in `gatsby-mdx` (if it isn't already) and not in each remark subplugin.

/cc @ChristopherBiscardi @johno

johngrasty:
@pieh That seems reasonable, and it is [documented there ](url
https://www.gatsbyjs.org/packages/gatsby-mdx/?=mdx#gatsby-remark-plugins).

In that case maybe it would be worth considering a short one sentence mention and a link to the `gatsby-mdx` docs at the end of the subplugins. Otherwise this is an inherent stumbling block to mdx adoption, as one has to always remember that the subplugin directions are not the one source of truth.

It's all the same to me though, as I have now figured it out. 😀

ChristopherBiscardi:
also https://github.com/gatsbyjs/gatsby/pull/11361 is going to be merged soon, so it will be documented on the gatsby site as well as the gatsby-mdx plugin README on the plugins page

PR #12681, opened 2019-03-19
feat(gatsby-image): Add gatsby-image/withIEPolyfill export with object-fit/position support

### LINKED ISSUE #4021: gatsby-image browser support

ooloth:
## Description

This PR will allow users who want `object-fit/object-position` support in IE to simply `import Img from "gatsby-image/withIEPolyfill"` instead of importing directly from `"gatsby-image"`.

Please feel free to review the PR and let me know if the implementation can be improved.

The `/withIEPolyfill` version of `gatsby-image` currently does the following:

1. Checks if the browser supports the `object-fit/position` CSS properties.
  a. If yes, no polyfill is loaded.
  b. If no, the `object-fit-images` polyfill is imported and called.
2. Wraps `gatsby-image` in a component that passes the new `objectFit` and `objectPosition` prop values to the polyfill implementation (which requires a weird `font-family` hack).

### New props

To make the implementation simple, I've added new `objectFit` and `objectPosition` props that will needed to be used to pass the corresponding values to the polyfill. This is to avoid a scenario where a user attempts to set these values in a way we can't pass to the polyfill (e.g. via an external CSS stylesheet) and doesn't understand why the polyfill isn't working. Let me know if this API can be improved.

### Loading polyfill in /withIEPolyfill/index.js

I opted to load the `object-fit-images` polyfill directly in `withIEPolyfill/index.js` rather than in `gatsby-browser.js`. If anyone knows of a better way to approach this, please let me know and feel free to make improvements.

### Polyfill repo has been archived

I was surprised to see that the `object-fit-images` [repo](https://github.com/bfred-it/object-fit-images) has recently been archived by its owner. It still works, but if anyone is concerned about this or knows of a reliable alternative polyfill, please let me know.

### Docs to be updated once implementation is finalized

Once we've settled on the API for this `/withIEPolyfills` component, a brief explanation will need to be added to the `gatsby-image` docs. I held off on adding one for now in case we end up changing the API.

## Related Issues

Fixes #4021.


packages/gatsby-image/withIEPolyfill/index.js
```diff
@@ -0,0 +1,56 @@
+"use strict";
```
wardpeet: you should add this file to gitignore as we don't commit compiled sources.

packages/gatsby-image/src/withIEPolyfill/index.js
```diff
@@ -0,0 +1,40 @@
...
+import Image from "../index"
+
+// Dynamically load polyfill for object-fit/object-position if needed (e.g. in IE)
+const testImg = document.createElement(`img`)
+if (
+  typeof testImg.style.objectFit === `undefined` ||
+  typeof testImg.style.objectPosition === `undefined`
+) {
+  import(`object-fit-images`).then(({ default: ObjectFitImages }) =>
+    ObjectFitImages()
```
wardpeet: I can read `This will fix all the images on the page and also all the images added later (auto mode).` on the github page but I can't see how they actually achieve this. It looks like they listen for onload but nothing after that.

I would go for componentDidMount and if possible createRef so we can only trigger it on the current image so we don't do to much work. If you need an example of what I'm thinking feel free to ask.

wardpeet:
this is great!

packages/gatsby-image/withIEPolyfill/index.js
ooloth: Will do!

I notice that the compiled version of the main image element (`packages/gatsby-image/index.js`) doesn't appear to be included in the `.gitignore` file... Should I add it as well, or is this wrapper component handled differently?

packages/gatsby-image/src/withIEPolyfill/index.js
ooloth: Sounds good! I assume the ref will need to be attached to the actual `` element rendered by the `<Image />` component?

Since the `` element isn't exposed in this wrapper component, any tips on how to drill the ref down to it?

wardpeet:
```js
const imageRef = React.createRef();

componentDidMount() {
ObjectFitImages(imageRef.current.querySelector('img'))
}
```

something like this might work?

packages/gatsby-image/withIEPolyfill/index.js
wardpeet: sorry for giving you half baked answers.

`/*.js` is only handling direct children and because `withIEPolyfill/index.js` is inside a folder it's not being handled. You can change it into
```
/**/*.js
!/src/**/*.js
```
to also check subdirectories but keep the src folder in version control but for your usecase it's porbably enough to just disable withIEPolyfill like `/withIEPolyfill/*.js`

packages/gatsby-image/src/withIEPolyfill/index.js
ooloth: Of course! Silly me. 🤦‍♂️ On it.

packages/gatsby-image/withIEPolyfill/index.js
ooloth: I see! Thanks so much for that walkthrough.

packages/gatsby-image/src/withIEPolyfill/index.js
```diff
@@ -0,0 +1,52 @@
+import React, { Component, createRef } from "react"
+import PropTypes from "prop-types"
+import Image from "../index"
+
+class ImageWithIEPolyfill extends Component {
+  constructor(props) {
```
wardpeet: you could go for `imageRef = createRef()` and drop the constructor, babel transpiles this correctly.

wardpeet:
I updated the code a bit to accept an external ref as well. I'm going to test this that it doesn't break anything and get this merged tomorrow! 🎉 Thanks for all the hard work.

ooloth:
Love it! Thanks for the improvements. 👍

wardpeet:
pushed the latest fix, ref handling is ugly but changing forwardRef inside gatsby-image would be a breaking change.

ooloth:
Sounds great! Thanks for guiding it over the finish line.

lannonbr:
Awesome to see this merged in! 🎉 

Is it possible to append the docs for gatsby-image about this new feature if it hasn't already? @ooloth

ooloth:
Absolutely! 

I just submitted PR #12869, which adds instructions to the `gatsby-image` README. 

Feel free to improve it as needed.

PR #12690, opened 2019-03-20
Add stack trace to error reporting in GraphiQL

stefanprobst:
By default, `graphql-js`'s `formatError` which is also used in Graph_i_QL does not show the `GraphQLError` stack trace which can be quite useful for debugging.


KyleAMathews:
Does this error happen as a result of normal user errors? Is the stack trace useful for them? There's a number of graphql errors that trigger useless stack traces we need to eliminate as they significantly complicate users fixing problems. Just want to make sure we're not adding more noise to the cli.

stefanprobst:
@KyleAMathews The CLI is unchanged, this is only about what will be reported in GraphiQL IDE. It looks like this:
graphiql

KyleAMathews:
Ok perfect!

PR #12691, opened 2019-03-20
chore(gatsby-cli): remove bluebird promise from global namespace

stefanprobst:
We currently set `global.Promise` to `bluebird`. I'm not entirely clear why we do this, but it does add quite a bit of noise to stack traces.


wardpeet:
it was more or less a fix for node < 6 unsure if this breaks anything though (I don't think so as all plugins require bluebird themselves)

PR #12693, opened 2019-03-20
feat(redux): cache truePath lookups for performance

bennetthardwick:
## Description
When creating 50k pages, Gatsby spends a lot of time inside the `truePath` method (which down the stack reads every directory on each of the patch segments). In an application with a small number of components (2 or so), this method doesn't need to be run 50k times. 

To fix this, a small `Map` was added to cache these calls, instead of running them every time.

Using this patch with [gatsby-intense-benchmark](https://github.com/bennetthardwick/gatsby-intense-benchmark), the `createPages` step of the build goes from about 60 seconds to 20 - 30 seconds.


packages/gatsby/src/redux/actions.js
```diff
@@ -86,6 +86,7 @@ const pascalCase = _.flow(
 )
 const hasWarnedForPageComponentInvalidContext = new Set()
 const hasWarnedForPageComponentInvalidCasing = new Set()
+const pageComponentCache = new Map()
```
sidharthachatterjee: Want to replace this with a WeakMap? WeakMaps can be cleaned up by GC so great for caches

DSchau: This is a great idea! We should also see if this approach improves performance in other areas where we're using Map (where it makes sense to use a WeakMap!).

bennetthardwick: Unfortunately according to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap#Description) primitive values can't be as keys / values for weak maps, only object references.

packages/gatsby/src/redux/actions.js
```diff
@@ -201,37 +202,48 @@ ${reservedFields.map(f => `  * "${f}"`).join(`\n`)}
...
-        `
-          )
-          hasWarnedForPageComponentInvalidCasing.add(page.component)
+      // check if we've processed this component path
+      // before, before running the expensive "truePath"
+      // operation
+      if (pageComponentCache.has(page.component)) {
+        page.component = pageComponentCache.get(page.component)
+      } else {
+        const originalPageComponet = page.component
```
DSchau:
```suggestion
        const originalPageComponent = page.component
```

packages/gatsby/src/redux/actions.js
```diff
@@ -201,37 +202,48 @@ ${reservedFields.map(f => `  * "${f}"`).join(`\n`)}
...
+                                 ${markers}
+          `
+            )
+            hasWarnedForPageComponentInvalidCasing.add(page.component)
+          }
+
+          page.component = trueComponentPath
         }
-        page.component = trueComponentPath
+        pageComponentCache.set(originalPageComponet, page.component)
```
DSchau:
```suggestion
        pageComponentCache.set(originalPageComponent, page.component)
```

packages/gatsby/src/redux/actions.js
pieh: This map won't ever be huge (as this just map template paths), so I don't worry about memory/gc here

bennetthardwick:
> Quick nit: re a typo, otherwise this looks good.

Thanks @DSchau! I can't believe I missed those.

> With your comment on WeakMap only supporting objects, we could just _use_ an object here--or we can just keep the original Map.
> 
> I'm not sure there'd be a hugely meaningful difference.

Reading more in to Maps, it seems the difference between them and objects is that the keys are kept in the order they were inserted. Since this isn't needed it's probably unneeded overhead (albeit marginal). I'll update it to use an object instead. :+1:

Issue #12692, opened 2019-03-20
New Gatsby type inference is slow on 60k pages


bennetthardwick:
## Description

After the `onPreExtractQueries` step of the build process, Gatsby gets when looking through 60k or so pages. Specifically, the `isDate` method (which is run on every string to check whether it's a string or a date string), is taking up the most time. These methods are all being run from the [example-value.js](https://github.com/gatsbyjs/gatsby/blob/07e69be2038de51bdd829bb8298cf2e728d49007/packages/gatsby/src/schema/infer/example-value.js).

### Steps to reproduce

1. Clone [gatsby-intense-benchmark](https://github.com/bennetthardwick/gatsby-intense-benchmark)
2. Run:
```sh
node --inspect-brk node_modules/.bin/gatsby develop
```
3. Open chrome / chromium, navigate to `chrome://inspect` and start the debugger.
4. When the build process finishes `onPreExtractQueries`, wait for a few minutes and then pause the debugger (chances are it will be in the right spot)
5. Alternatively run the profiler for a few minutes and inspect the chart

The build will be stuck in this state for 30+ minutes (I haven't had a successful build).

### Expected result

Build completes in a reasonable time.

### Actual result

Build never completes.

### Environment

  System:
    OS: Linux 4.15 Ubuntu 18.04.2 LTS (Bionic Beaver)
    CPU: (4) x64 Intel(R) Core(TM) i5-4300U CPU @ 1.90GHz
    Shell: 5.4.2 - /usr/bin/zsh
  Binaries:
    Node: 10.15.0 - ~/.nvm/versions/node/v10.15.0/bin/node
    Yarn: 1.13.0 - ~/.nvm/versions/node/v10.15.0/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v10.15.0/bin/npm
  Languages:
    Python: 2.7.15 - /usr/bin/python
  npmPackages:
    gatsby: ^2.2.1 => 2.2.1 
    gatsby-image: ^2.0.34 => 2.0.34 
    gatsby-plugin-manifest: ^2.0.24 => 2.0.24 
    gatsby-plugin-offline: ^2.0.25 => 2.0.25 
    gatsby-plugin-react-helmet: ^3.0.10 => 3.0.10 
    gatsby-plugin-sharp: ^2.0.29 => 2.0.29 
    gatsby-source-filesystem: ^2.0.27 => 2.0.27 
[...]

PR #12700, opened 2019-03-20
fix(gatsby): quick check if string looks like a date

wardpeet:
## Description
Checks if strings start with 4 numbers as all ISO 8601 date-time are using YYYY syntax

## Related Issues
#12692

### DIFF
--- a/packages/gatsby/src/schema/types/date.js
+++ b/packages/gatsby/src/schema/types/date.js
@@ -99,8 +99,13 @@ const GraphQLDate = new GraphQLScalarType({
 // Check if this is a date.
 // All the allowed ISO 8601 date-time formats used.
 function isDate(value) {
+  // quick check if value does not look like a date
+  if (typeof value === `number` || !/^\d{4}/.test(value)) {
+    return false
+  }
+
   const momentDate = moment.utc(value, ISO_8601_FORMAT, true)
-  return momentDate.isValid() && typeof value !== `number`
+  return momentDate.isValid()
 }
 
 const formatDate = ({


millette:
Awesome! https://github.com/gatsbyjs/gatsby/issues/12692#issuecomment-475048164

stefanprobst:
I think we should also revert #12533 which increased our supported date formats from 24 to 55!

packages/gatsby/src/schema/types/date.js
```diff
@@ -99,8 +99,13 @@ const GraphQLDate = new GraphQLScalarType({
 // Check if this is a date.
 // All the allowed ISO 8601 date-time formats used.
 function isDate(value) {
+  // quick check if value does not look like a date
+  if (typeof value === `number` || !/^\d{4}/.test(value)) {
```
me4502: I'm not sure an inline regex is the fastest way of determining if the first 4 values are numeric.

You could pre-compile, or alternatively even have a set of possible single digit strings, and iterate over the first 4 and break if it's not in the set.

me4502: According to this benchmark, it seems inline regex or pre-compiled is actually fastest. 

https://jsperf.com/check-whether-string-starts-with-four-numbers

In Chrome it doesn't appear to matter if it's inline or not, in Safari I notice a 2% improvement in pre-compiling.

millette: I tried both ways and the change was marginal, about 600s to build the schemas. Without the patch, it was 8000s. My guess is that the regex is optimized automatically.

wardpeet: @me4502 I used string.slice(0, 4) and checked if it was a number and regex was faster. I'm pretty sure v8 optimizes for it.

@millette at least we got somewhere, not fast enough but faster.

PR #12719, opened 2019-03-21
chore(*): Remove CSS false positives - Use newer JSDom version

polarathene:
## Description

Certain CSS styles have been failing to appear in Jest tests due to an outdated JSDom(and some it's dependencies, one being updated with bugfix in Feb 2019). This was causing false positives for `gatsby-image` tests for a while.

Nothing major, just some CSS props weren't appearing in tests when they technically should have been.

## Details

[My PR](https://github.com/gatsbyjs/gatsby/pull/12468) is failing in the CI but was passing tests locally.

I had been running tests on the `gatsby-image` package on it's own, which uses `react-testing-library: ^5.0.0`, and it failed against the snapshot test. Updated snapshots failed when my PR ran on CircleCI. The main repo root yarn.lock file has `react-testing-library: ^5.2.1` locked down(retrieves 5.2.1 instead of 5.9.0, although is a v6 release now). 

Correcting this didn't help for passing the test suite. [`cssstyle: 1.2.0`](https://github.com/jsakas/CSSStyleDeclaration/pull/74) fixes the [problems with certain styles](https://github.com/jsdom/jsdom/issues/1965#issuecomment-326177053) being [ignored](https://github.com/kentcdodds/react-testing-library/issues/214)(eg. `transitionDelay`, and certain color values for `backgroundColor`, "red" works, but not "lightgray"). That however [needs `jsdom: ^12.0.0`](https://github.com/jsdom/jsdom/commit/ed114650bf2c5340eba3da6295f5fc13c5f63307#diff-b9cfc7f2cdf78a7f4b91a753d10865a2), but [Jest refuses to update as support for Node <8 is dropped](https://github.com/facebook/jest/issues/8016#issuecomment-468673486)..

I've added a workaround until Jest supports a newer major version of JSDom. Using [`jest-environment-jsdom-fourteen`](https://github.com/ianschmitz/jest-environment-jsdom-fourteen) is the [advised approach](https://github.com/facebook/jest/pull/7122#issuecomment-458651329) for the meantime, it'll only impact users running Jest test suite on Node 6, that should be fine for Gatsby? (Node 6 CI test seems to pass)

Only `gatsby-image` tests needed updates. [Node 6 will be EoL in May](https://github.com/nodejs/Release#release-schedule), no idea when Jest will bump JSDom from then, but we could wait a few months if you rather avoid the workaround?

## Related Issues

#12468 (Spotted it while working on this)

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -45,4 +45,5 @@ module.exports = {
   },
   collectCoverageFrom: coverageDirs,
   reporters: [`default`].concat(useCoverage ? `jest-junit` : []),
+  testEnvironment: `jest-environment-jsdom-fourteen`,
 }
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "husky": "1.1.1",
     "jest": "^24.0.0",
     "jest-cli": "^24.0.0",
+    "jest-environment-jsdom-fourteen": "^0.1.0",
     "jest-junit": "^6.1.0",
     "jest-serializer-path": "^0.1.15",
     "lerna": "^3.10.7",
--- a/packages/gatsby-image/src/__tests__/__snapshots__/index.js.snap
+++ b/packages/gatsby-image/src/__tests__/__snapshots__/index.js.snap
@@ -7,14 +7,14 @@ exports[`<Img /> should render fixed size images 1`] = `
     style="position: relative; overflow: hidden; display: inline; width: 100px; height: 100px;"
   >
     <div
-      style="width: 100px; opacity: 1; height: 100px;"
+      style="background-color: lightgray; width: 100px; opacity: 1; transition-delay: 0.25s; height: 100px;"
       title="Title for the image"
     />
     <img
       alt=""
       class="placeholder"
       src="string_of_base64"
-      style="position: absolute; top: 0px; left: 0px; width: 100%; height: 100%; opacity: 1; color: red;"
+      style="position: absolute; top: 0px; left: 0px; width: 100%; height: 100%; object-fit: cover; object-position: center; opacity: 1; transition: opacity 0.5s; transition-delay: 0.25s; color: red;"
       title="Title for the image"
     />
     <picture>
@@ -29,7 +29,7 @@ exports[`<Img /> should render fixed size images 1`] = `
         itemprop="item-prop-for-the-image"
         src="test_image.jpg"
         srcset="some srcSet"
-        style="position: absolute; top: 0px; left: 0px; width: 100%; height: 100%; opacity: 0;"
+        style="position: absolute; top: 0px; left: 0px; width: 100%; height: 100%; object-fit: cover; object-position: center; opacity: 0; transition: opacity 0.5s;"
         title="Title for the image"
         width="100"
       />
@@ -51,14 +51,14 @@ exports[`<Img /> should render fluid images 1`] = `
       style="width: 100%; padding-bottom: 66.66666666666667%;"
     />
     <div
-      style="position: absolute; top: 0px; bottom: 0px; opacity: 1; right: 0px; left: 0px;"
+      style="background-color: lightgray; position: absolute; top: 0px; bottom: 0px; opacity: 1; transition-delay: 0.25s; right: 0px; left: 0px;"
       title="Title for the image"
     />
     <img
       alt=""
       class="placeholder"
       src="string_of_base64"
-      style="position: absolute; top: 0px; left: 0px; width: 100%; height: 100%; opacity: 1; color: red;"
+      style="position: absolute; top: 0px; left: 0px; width: 100%; height: 100%; object-fit: cover; object-position: center; opacity: 1; transition: opacity 0.5s; transition-delay: 0.25s; color: red;"
       title="Title for the image"
     />
     <picture>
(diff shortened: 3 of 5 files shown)


polarathene:
Slightly off-topic. Is the `yarn.lock` ever "refreshed"? Or only updated portions when a PR requires it?

I haven't updated `react-testing-library` in this PR, and I have noticed there's a fair amount of dependencies listed for `yarn outdated`, some of the major versions are multiple versions ahead, and there is some packages listed with mixed major versions, along with some packages that have reached their `1.0.0` release which might be worth switching to. 

Plenty of minor and patch versions too. If tests pass with a `yarn upgrade`(ignores major versions), is that a good indication of stability and sending a PR of an updated `yarn.lock`?

pieh:
> If tests pass with a `yarn upgrade`(ignores major versions), is that a good indication of stability and sending a PR of an updated `yarn.lock`?

Not entirely - dependency changes in packages and e2e/integration tests are currently problematic as `gatsby-dev-cli` doesn't handle them currently (working on adding support for that in https://github.com/gatsbyjs/gatsby/pull/11525 ). Unit tests will be fine (deps changes are covered there).

polarathene:
@pieh Ok so once #11525 is merged, it'd be a good time to look at doing a more blanket update on `yarn.lock`?(not modifying `packages.json` unless updating major versions)

Is that also why all those CI tests are breaking? I can't replicate that error they're all breaking on.

pieh:
Yes, failing e2e/integration tests is symptom of `gatsby-dev-cli` not handling dependency changes: we published new version of gatsby with dependency updates, but git commit that you created branch from uses older version of those deps, so any breaking changes in those packages will cause this weird error.

With verdaccio support (#11525) we will install same dependencies that are specified in your branch, so those tests will actually be reliable and updating package deps will be much easier both for us (to review) and for contributors (to reliably test their changes before opening PRs).

PS. Over the last few days I think "dependencies" is my most used word ;)

wardpeet:
mmh technically we still support node 6 so i'm afraid that our tests might fail in the future

pieh:
Hmm if this change should work in node 8+ then our test suite is wonky because unit tests for node 6 are passing right now

wardpeet:
That's the weird thing 😂 maybe babel trickery?

Op vr 22 mrt. 2019 17:22 schreef Michal Piechowiak <notifications@github.com

pieh:
I just made temporary PR with test that would fail in node 6 - and it does fail ( https://github.com/gatsbyjs/gatsby/pull/12763 ), so I'm fine with dev-only change that technically require node 8+, but actually works :P

But yeah, first need verdaccio stuff to for integration/e2e tests to be reliable

polarathene:
> mmh technically we still support node 6 so i'm afraid that our tests might fail in the future

@wardpeet It was passing node 6 unit tests originally iirc, I know there is some reworking of the CI going on so maybe that's why they're failing now?

Node 6 will be EOL in May, so I wouldn't be too concerned about future support for much longer? Once Jest drops support for 6 and updates JSDom from 11 to 12+ the corrected test for gatsby-image will fail, requiring the same changes because prior to JSDom 12, there was CSS failing to be picked up.

`jest -u` was trusted to generate the correct snapshot and I guess no one realized the js test file was comparing a style that was missing other styles that should have been in the test as a result of that? So that's why the test has been passing fine presently including in node 6. That's a false positive though?

> Hmm if this change should work in node 8+ then our test suite is wonky because unit tests for node 6 are passing right now

@pieh Jest won't upgrade JSDom because they drop node 6 support with newer versions, that doesn't necessarily mean that tests will break on node 6, it could mean that certain codepaths may break. I recall the unit tests were passing for node 6 when I submitted this PR but they no longer are now?(Oh nvm it was canceled)

This PR shouldn't affect users on node 6 too much afaik, it's only related to JSDom for Jest tests. If it was causing problems, removing the jest config line should use the usual package that was being replaced and they should be able to run tests as before. Either or, Node 6 goes EOL in a little over a month and Jest will update to JSDom 12+ at some point then.

---

Happy to have this PR rejected or left open rather than merged if you like. Apart from the updated `gatsby-image` test, it's only intended as a temporary fix until Jest updates JSDom.

If you'd rather not merge, I'll update my linked PR to remove it's updated tests(that are valid when testing `gatsby-image` in isolation due to `react-testing-library` using `jest-dom` for Jest tests, full suite tests with `jest-cli` package enforces the older JSDom though which is why there's that disparity(not sure if that's possibly a stale yarn.lock issue).

wardpeet:
> @wardpeet It was passing node 6 unit tests originally iirc, I know there is some reworking of the CI going on so maybe that's why they're failing now?

i'm not against it but technically we are still supporting node 6 as dropping it would be a breaking change.

If you merge lastest master in this PR you'll see that CI will work again.

polarathene:
> I'm not against it but technically we are still supporting node 6 as dropping it would be a breaking change.

I'm not requesting to drop node 6 support. This doesn't change anything on Gatsby's end for actual users afaik, it's just related to CI / Jest for generating proper snapshots and the related test to actually get the correct CSS that an actual Gatsby site would generate anyhow. 

AFAIK, nothing else is requiring JSDom <12? Nor does this PR appear to cause any breakage in the node 6 test.

> If you merge lastest master in this PR you'll see that CI will work again.

Done. If this PR is merged, the committed workaround should be removed once Jest updates to JSDom 12+ later in the year.

wardpeet:
I've done a comparison with https://github.com/jsdom/jsdom/compare/13.2.0...14.0.0 and I agree with @polarathene lets just go for it and if node 6 starts failing we can lock the version.

PR #12722, opened 2019-03-21
fix(gatsby): added looksLikeADate to check date on schema creation

wardpeet:
## Description
Moment.js is slow at parsing dates so this PR does a lot of extra validation before we fallback to moment. I copied some code from moment.js to parse the ISO formats and do a quick regex myself.

Build Schema on www now is 2s on my machine instead of 60s before.

Checks that are being done:
 1) is it a number?
 2) does the length of the value comply with any of our formats
 3) does the str starts with 4 digites (YYYY)
 4) does the str ends with something that looks like a date
 5) Small regex to see if it matches any of the defined ISO formats
 6) check momentjs

## Related Issues
#12692


packages/gatsby/src/schema/types/date.js
```diff
@@ -96,14 +96,70 @@ const GraphQLDate = new GraphQLScalarType({
...
+ * 6) check momentjs
+ *
+ * @param {string|number} value
+ * @return {boolean}
+ */
 function isDate(value) {
   // quick check if value does not look like a date
-  if (typeof value === `number` || !/^\d{4}/.test(value)) {
+  if (
+    typeof value === `number` ||
```
freiksenet: `== null` check

wardpeet:
ok so I couldn't get the isDate checks to work how it should. Momentjs does checks for validation of dates like 30 february will never be true but 29 might be. So i added an extra function that checks if it looks like a date which is not bullet proof but probably good enough for schema creation.

freiksenet:
Thank you Ward!

packages/gatsby/src/schema/types/__tests__/date.js
```diff
@@ -152,6 +152,132 @@ describe(`isDate`, () => {
   })
 })
 
+describe(`looksLikeADate`, () => {
+  it.each([
```
DSchau: Today I learned of `it.each`. Cool!

wardpeet: me too actually 😛

PR #12726, opened 2019-03-21
fix(www): ensure rotator is ssr-friendly

DSchau:
## Description

Fixes a tiny regression re: SSR rendering introduced in #12530

Validated this by:

1. Ensuring rotater animates (by default!) in built version
1. Ensuring rotater _does not_ animate when reduced motion is enabled in System Preferences - Accessibility
1. Ensuring site can build with `gatsby build` & `gatsby serve`

### DIFF
--- a/www/src/components/rotator.js
+++ b/www/src/components/rotator.js
@@ -31,6 +31,7 @@ const controlButtonStyles = {
 
 class Rotator extends Component {
   state = {
+    shouldAnimate: false,
     item: 0,
     size: {},
   }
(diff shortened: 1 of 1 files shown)


DSchau:
Hey @marcysutton! I also missed that you requested to review #12530--we actually broke the build in that PR (nothing's built out yet 😅), so if you have a chance would love an accessibility review on [this component](https://github.com/gatsbyjs/gatsby/blob/ae57945103802eef17f8ed2504d259294ab80fce/www/src/components/rotator.js) 🙏

marcysutton:
I'll test it locally, we can always add a revision PR! Thanks for looping me in.

PR #12727, opened 2019-03-21
fix(www): prevent filter accordions from disappearing on collapse

### LINKED ISSUE #12517: (www) "Starters" and "Showcase" sidebar accordions "disappear" when collapsing
* visit https://www.gatsbyjs.org/starters/?v=2
* click "Categories" or "Gatsby Dependencies" in the filter sidebar
* see the filters "disappear" instead of collapse
* markup is still there: 
image

amad-person:
## Description

The following filters on the Gatsby website were disappearing from view on collapsing them:
1. **Categories** in Starters
2. **Gatsby Dependencies** in Starters
3. **Categories** in Showcase

All three of these filter accordions use the `Collapsible` component, which takes in a prop called `fixed` to set the `minHeight` and `maxHeight` of the component. If these heights are set, then the filter accordions correctly do not disappear from view.

This PR hence introduces the following changes:

1. Adds the prop `fixed={150}` for **Categories** in Starters
2. Adds the prop `fixed={150}` for **Gatsby Dependencies** in Starters
3. Adds the prop `fixed={300}` for **Categories** in Showcase

Notes: 
- The **Gatsby Version** category in Starters uses `fixed={150}` already, so I have used the same value for the other two categories in Starters. 
- `Categories` in Showcase is the only filter, so I chose a bigger value for it.

Please let me know if any other changes have to be made! Thanks 😄 

## Related Issues

Fixes #12517.

### DIFF
--- a/www/src/views/showcase/collapsible-filter-list.js
+++ b/www/src/views/showcase/collapsible-filter-list.js
@@ -14,7 +14,7 @@ const CollapsibleFilterList = ({
   setFilters,
   heading,
 }) => (
-  <Collapsible heading={heading}>
+  <Collapsible heading={heading} fixed={300}>
     {categoryKeys.map(c => (
       <button
         key={c}
--- a/www/src/views/starter-library/filtered-starters.js
+++ b/www/src/views/starter-library/filtered-starters.js
@@ -134,6 +134,7 @@ export default class FilteredStarterLibrary extends Component {
               setFilters={setFiltersVersion}
             />
             <LHSFilter
+              fixed={150}
               heading="Categories"
               data={Array.from(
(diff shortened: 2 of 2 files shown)


fk:
Didn't get to anything the last time around, will give this another shot now.

fk:
Hey @amad-person! Sorry for being so slow on the updates. :-( 🙏

In the meantime, @jgalat found a fix for the original issue that we just merged #13673.
It avoids setting min- and max-heights, and doesn't exhibit the "available height" issue described in https://github.com/gatsbyjs/gatsby/pull/12727#pullrequestreview-227691405.

Thanks for your work here, it's really appreciated! <3

PR #12735, opened 2019-03-21
fix(themes): check issuer for null

ChristopherBiscardi:
If a theme author sets the following config in their theme
(specifically `modules` with `src`), then the issuer for some requests
will be null.

```js
exports.onCreateWebpackConfig = ({ actions }) => {
  actions.setWebpackConfig({
    resolve: {
      modules: [path.resolve(__dirname, 'src'), 'node_modules'],
      alias: {
        $components: path.resolve(__dirname, 'src/components'),
      },
    },
  })
}
```

If the issuer is null, it can't be a self-reference shadow, so skip
the check that enables shadowed components to import their
shadow-parents.

---

Overriding `modules` is not generally a good idea in themes (if this config is set in two themes then you can get the wrong module depending on the order of theme inclusion), but we can't stop people from doing it and we shouldn't break unnecessarily if they do.

### DIFF
--- a/packages/gatsby/src/internal-plugins/webpack-theme-component-shadowing/index.js
+++ b/packages/gatsby/src/internal-plugins/webpack-theme-component-shadowing/index.js
@@ -35,20 +35,32 @@ module.exports = class GatsbyThemeComponentShadowingResolverPlugin {
       // get the location of the component relative to src/
       const [, component] = request.path.split(path.join(theme, `src`))
 
-      const issuerExtension = path.extname(request.context.issuer)
+      /**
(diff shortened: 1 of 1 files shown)


pieh:
> Overriding `modules` is not generally a good idea in themes (if this config is set in two themes then you can get the wrong module depending on the order of theme inclusion), but we can't stop people from doing it and we shouldn't break unnecessarily if they do.

If this is something that can cause problems when using themes / component shadowing, perhaps we could show warning about so theme authors are aware of it and should not override `modules` configuration? Out of scope for this PR of course.

PR #12737, opened 2019-03-21
Make the content field docs clearer for the createNode action

KyleAMathews:
Make very clear it should only be used in rare cases e.g. when pulling markdown strings from an API.

Also that you _shouldn't_ `JSON.stringify` your node's content here — a pattern which leaked into the wild and keeps coming up.


muescha:
you mean this as example?
https://github.com/gatsbyjs/gatsby/blob/afa0f1e9c1ca0af106d9e9520aaf4950b0430dc7/packages/gatsby-source-contentful/src/extend-node-type.js#L471

KyleAMathews:
@muescha that's an edge case for sure where setting `content` is helpful. But really weird so we wouldn't want to mention it in the docs.

PR #12746, opened 2019-03-22
feat(themes): copy static folders in themes

grantglidewell:
## Description

Copy theme static folders to the public folder

## Related Issues

Addresses https://github.com/ChristopherBiscardi/gatsby-theme-examples/issues/7


grantglidewell:
This implementation prefers the end user's static directory over the theme static directories. If we want a different behavior I would just need to know the priority.

packages/gatsby/src/commands/build.js
```diff
@@ -5,7 +5,7 @@ const buildHTML = require(`./build-html`)
 const buildProductionBundle = require(`./build-javascript`)
 const bootstrap = require(`../bootstrap`)
 const apiRunnerNode = require(`../utils/api-runner-node`)
-const { copyStaticDir } = require(`../utils/get-static-dir`)
+const { copyStaticDirs } = require(`../utils/get-static-dir`)
```
ChristopherBiscardi: @pieh Can we reasonably expect that no one is using this export? It's technically require'able in the gatsby npm module.

pieh: Those are considered internals. Anything that would end up being `gatsby/dist/something` imports should be considered private APIs (with exception of `commands` directory, which are needed for `gatsby-cli` to not have any changes).

So export name change is fine to make (if it makes sense)

ChristopherBiscardi: it makes sense. It's now copying all static/ dirs from themes and the user's site into the build folder.

ChristopherBiscardi:
> This implementation prefers the end user's static directory over the theme static directories. If we want a different behavior I would just need to know the priority.

Yeah we want this behavior. The order of theme inclusion is important and the user's site is always last so that it "wins" if there are any conflicts.

ChristopherBiscardi:
@pieh looks like we still need core sign-off on this

pieh:
Thanks @grantglidewell for implementing this!

PR #12748, opened 2019-03-22
feat(gatsby): Allow awaiting API run triggered by createNode action

stefanprobst:
When the `createNode` action is called, we currently have no way of knowing when subsequent `onCreateNode` API calls have finished. This is e.g. a problem in custom field resolvers, where we cannot safely call `createRemoteFileNode`.

This PR makes `createNode` a thunk which still synchronously dispatches the actions, but returns a Promise of the triggered API run that can optionally be awaited.


packages/gatsby/src/redux/actions.js
```diff
@@ -655,6 +656,24 @@ actions.createNode = (
...
+  if (createNodeAction) {
+    const { payload: node, traceId, parentSpan } = createNodeAction
+    return apiRunnerNode(`onCreateNode`, {
+      node,
+      traceId,
+      parentSpan,
+      traceTags: { nodeId: node.id, nodeType: node.internal.type },
+    })
+  }
+  return undefined
```
wardpeet: couldn't we just drop this line?
```suggestion
```

packages/gatsby/src/redux/actions.js
```diff
@@ -655,6 +656,24 @@ actions.createNode = (
   }
 }
 
+actions.createNode = (...args) => dispatch => {
+  const actions = createNode(...args)
+  dispatch(actions)
+  const createNodeAction = (Array.isArray(actions) ? actions : [actions]).find(
+    action => action.type === `CREATE_NODE`
+  )
+  if (createNodeAction) {
```
wardpeet: this is just a suggestion, I find this a bit easier to read

```js
if (!createNodeAction) {
  return;
}

const { payload: node, traceId, parentSpan } = createNodeAction
return apiRunnerNode(`onCreateNode`, {
  node,
  traceId,
  parentSpan,
  traceTags: { nodeId: node.id, nodeType: node.internal.type },
})
```

packages/gatsby/src/redux/actions.js
stefanprobst: That's there to make the linter happy.

stefanprobst:
> why can't we safely call createRemoteFileNode?

Currently the `createNode` action is [re-emitted as an event](https://github.com/gatsbyjs/gatsby/blob/d5c3351732f5f683202161d83b3708c1de111517/packages/gatsby/src/redux/index.js#L99), which is [caught in `plugin-runner`](https://github.com/gatsbyjs/gatsby/blob/d5c3351732f5f683202161d83b3708c1de111517/packages/gatsby/src/redux/index.js#L99), which calls `api-runner-node`. The event-emitter sort of cuts off the feedback to the callsite.

The problem is that when we call `createRemoteFileNode` in a resolver, and want to e.g. query a `childImageSharp` field on that file node. When `createRemoteFileNode` returns we can be sure that the `File` node has been created (all is synchronous), but we need to await the subsequent `onCreateNode` API call as well that creates the `ImageSharp` child node. Otherwise, the `childImageSharp` field will be `null`.

packages/gatsby/src/redux/actions.js
stefanprobst: Thanks! Done.

stefanprobst:
Anything missing from this?

packages/gatsby/src/redux/actions.js
```diff
@@ -502,7 +503,7 @@ const typeOwners = {}
  *   }
  * })
  */
-actions.createNode = (
+const createNode = (
```
pieh: NIT: it's a bit confusing to have `createNode` and `actions.createNode` be something else - does it make sense to make slight name change?

jsdoc being applied to this and not `actions.createNode` possibly can mess API reference page?

stefanprobst: @pieh hmm that was kind of on purpose
* didn't want to touch the action creator, just wrap it
* what should be documented is the thing that is passed down, which is the bound action creator, so we have to keep the name there; and we have to keep the name for what *is* passed down on the `actions` object
* action docs page looks ok
* that said, we *will* want to revisit this when we more selectively distinguish between public actions and internal actions.

DSchau:
Thinking this looks good--let's get this merged today?

DSchau:
@pieh want to take a look at this soon?

pieh:
> @pieh want to take a look at this soon?

Yes, on it.

pieh:
I just verified using this simple snippet:
```md
---
title: Hello World
date: "2015-05-01T22:12:03.284Z"
remoteURL: "https://images.unsplash.com/photo-1556379092-dca659792591"
---
```

```js
exports.sourceNodes = ({ actions, schema, store, cache, createNodeId }) => {
  actions.createTypes(`
    frontmatter: Frontmatter
  }
  `)
  actions.createTypes(
    schema.buildObjectType({
      name: "Frontmatter",
      fields: {
        remoteURL: {
          type: `File`,
          resolve: async (source, args, context, info) => {
            console.log({ fieldValue })

            const fileNode = await createRemoteFileNode({
              url: fieldValue,
              store,
              cache,
              createNode: actions.createNode,
              createNodeId,
              ext: `.jpg`,
            })

            return fileNode
          },
        },
      },
    })
  )
}
```

that this indeed works as expected. We will need to look at bumping/adjusting query running concurrency later on (because using above will limit downloading files to be just from 4 queries, and this can be bottleneck). Will run few more tests (against our .org and few other sites to verify there isn't any unintentional breakage somewhere, but code LGTM

pieh:
Also additional note - (not a blocker) - right now, fact that `createNode` will return Promise is not documented. I'm not 100% sure where it should be documented - at least to action signature in API reference I think can be adjusted to note that Promise is returned, and when we have actually docs that cover using `await SomethingThatUsesCreateNode` in resolvers (so probably some advanced schema docs),  then we could add cross linking

packages/gatsby/src/redux/index.js
```diff
@@ -32,21 +31,20 @@ const readState = () => {
   return {}
 }
 
-exports.readState = readState
+const multi = ({ dispatch }) => next => action =>
```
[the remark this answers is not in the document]
stefanprobst: sure!

stefanprobst:
> We will need to look at bumping/adjusting query running concurrency later on

yes good point!

stefanprobst:
>  fact that `createNode` will return Promise is not documented

Added short note about the returned Promise

PR #12749, opened 2019-03-22
Remove unsued dependeny, and variable - Axios

prashant-andani:


polarathene:
Odd that it was never used when originally committed either.

PR #12758, opened 2019-03-22
feat(gatsby): add anonymous telemetry instrumentation to gatsby

jamo:
This PR adds anonymous telemetry instrumentation as outlined in the [telemetry RFC](https://github.com/gatsbyjs/rfcs/pull/33)

This implementation buffers all telemetry events to a local buffer file, from which events are then sent asynchronously on in a background process.

### DIFF
--- a/docs/docs/telemetry.md
+++ b/docs/docs/telemetry.md
@@ -34,8 +34,8 @@ Specifically, we collect the following information for _all_ telemetry events:
 
 - Timestamp of the occurrence
 - Command invoked (e.g. `build` or `develop`)
-- Unique and unidentifiable machine ID
-- Unique and unidentifiable session ID
+- Gatsby machine ID. This is generated with UUID and stored in global gatsby config at ~/.config/gatsby/config.json.
+- Unique session ID. This is generated on each run with UUID.
 - One-way hash of the current working directory or a hash of the git remote
 - General OS level information (operating system, version, CPU architecture, and whether the command is run inside a CI)
 - Current Gatsby version
@@ -44,6 +44,6 @@ The access to the raw data is highly controlled, and we cannot identify individu
 
 ## What about sensitive data? (e.g. secrets)
 
-We perform additional steps to ensure that secure data (e.g. environment variables used to store secrets for the build process) **do not** make their way into our analytics. We strip logs, error messages, etc. of this sensitive data to ensure we _never_ gain access to this sensitive data.
+We perform additional steps to ensure that secure data (e.g. environment variables used to store secrets for the build process) **do not** make their way into our analytics. [We strip logs, error messages, etc.](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-telemetry/src/sanitize-error.js) of this sensitive data to ensure we _never_ gain access to this sensitive data.
 
 You can view all the information that is sent by Gatsby’s telemetry by setting the environment variable `GATSBY_TELEMETRY_DEBUG`to `1` to print the telemetry data instead of sending it over.
--- a/docs/docs/telemetry.md
+++ b/docs/docs/telemetry.md
@@ -17,7 +17,7 @@ Users may always opt-out from the telemetry with `gatsby telemetry --disable` or
 Since much of Gatsby’s function revolves around community plugins and starters, we want to collect information on usage
 and reliability so that we can ensure a high-quality ecosystem.
 
-This raises a question. How will we use this telemetry data to improve the ecosystem? Some examples are helpful:
+This raises a question: how will we use this telemetry data to improve the ecosystem? Some examples are helpful:
 
 - We will be able to understand which plugins are typically used together. This will enable us to surface this information in our public plugin library and build more relevant starters and tutorials based on this data.
 - We will be able to surface popularity of different starters in the starter showcase.
--- a/docs/docs/telemetry.md
+++ b/docs/docs/telemetry.md
@@ -1,0 +1,49 @@
+---
+title: Telemetry
+---
+
+Gatsby contains a telemetry feature that collects anonymous usage information that is used to help improve Gatsby for all users.
+The Gatsby user base is growing very rapidly. It's important that our small team and the greater community will better understand the usage patterns, so we can best decide how to design future features and prioritize current work.
+
+You will be notified when installing Gatsby and when running it for the first time.
+
+## How to opt-out
+
+Users may always opt-out from the telemetry with `gatsby telemetry --disable` or setting the environment variable `GATSBY_TELEMETRY_DISABLED` to `1`
+
+## Why?
+
+**Anonymous** aggregate user analytics allow us to prioritize fixes and features based on how and when people use Gatsby.
+Since much of Gatsby’s function revolves around community plugins and starters, we want to collect information on usage
+and reliability so that we can ensure a high-quality ecosystem.
+
+This raises a question. How will we use this telemetry data to improve the ecosystem? Some examples are helpful:
+
+- We will be able to understand which plugins are typically used together. This will enable us to surface this information in our public plugin library and build more relevant starters and tutorials based on this data.
+- We will be able to surface popularity of different starters in the starter showcase.
+- We will be able to get more detail on the types of errors users are running into in _every_ build stage (e.g. development, build, etc.). This will let us improve the quality of our tool and better focus our time on solving more common, frustrating issues.
+- We will be able to surface reliability of different plugins and starters, and detect which of these tend to error more frequently. We can use this data to surface quality metrics and improve the quality of our plugins and starters.
+- We will be able to see timings for different build stages to guide us in where we should focus optimization work.
+
+## What do we track?
+
+We track general usage details, including command invocation, build process status updates, performance measurements, and errors.
+We use these metrics to better understand the usage patterns. These metrics will directly allow us to better decide how to design future features and prioritize current work.
+
+Specifically, we collect the following information for _all_ telemetry events:
+
+- Timestamp of the occurrence
+- Command invoked (e.g. `build` or `develop`)
+- Unique and unidentifiable machine ID
+- Unique and unidentifiable session ID
+- One-way hash of the current working directory or a hash of the git remote
+- General OS level information (operating system, version, CPU architecture, and whether the command is run inside a CI)
+- Current Gatsby version
+
+The access to the raw data is highly controlled, and we cannot identify individual users from the dataset. It is anonymized and untraceable back to the user.
+
+## What about sensitive data? (e.g. secrets)
+
+We perform additional steps to ensure that secure data (e.g. environment variables used to store secrets for the build process) **do not** make their way into our analytics. We strip logs, error messages, etc. of this sensitive data to ensure we _never_ gain access to this sensitive data.
+
+You can view all the information that is sent by Gatsby’s telemetry by setting the environment variable `GATSBY_TELEMETRY_DEBUG`to `1` to print the telemetry data instead of sending it over.
--- a/packages/gatsby-cli/src/create-cli.js
+++ b/packages/gatsby-cli/src/create-cli.js
@@ -355,9 +355,11 @@ module.exports = argv => {
             description: `Disable telemetry`,
           }),
 
-      handler: handlerP(({ enable, disable }) =>
-        setTelemetryEnabled(enable || !disable)
-      ),
+      handler: handlerP(({ enable, disable }) => {
+        const enabled = enable || !disable
+        setTelemetryEnabled(enabled)
+        report.log(`Telemetry collection ${enabled ? `enabled` : `disabled`}`)
+      }),
     })
     .wrap(cli.terminalWidth())
     .demandCommand(1, `Pass --help to see all available commands and options.`)
--- a/packages/gatsby-cli/src/create-cli.js
+++ b/packages/gatsby-cli/src/create-cli.js
@@ -5,6 +5,11 @@ const report = require(`./reporter`)
 const didYouMean = require(`./did-you-mean`)
 const envinfo = require(`envinfo`)
 const existsSync = require(`fs-exists-cached`).sync
+const {
+  trackCli,
+  setDefaultTags,
+  setTelemetryEnabled,
+} = require(`gatsby-telemetry`)
 
 const handlerP = fn => (...args) => {
   Promise.resolve(fn(...args)).then(
@@ -40,6 +45,11 @@ function buildLocalCommands(cli, isLocalSite) {
         `gatsby`,
         `package.json`
       ))
+      try {
+        setDefaultTags({ installedGatsbyVersion: packageInfo.version })
+      } catch (e) {
+        // ignore
+      }
       majorVersion = parseInt(packageInfo.version.split(`.`)[0], 10)
     } catch (err) {
       /* ignore */
@@ -311,6 +321,15 @@ module.exports = argv => {
 
   buildLocalCommands(cli, isLocalSite)
 
+  try {
+    const { version } = require(`../package.json`)
+    setDefaultTags({ gatsbyCliVersion: version })
+  } catch (e) {
+    // ignore
+  }
+
+  trackCli(argv)
+
   return cli
     .command({
       command: `new [rootPath] [starter]`,
@@ -320,6 +339,24 @@ module.exports = argv => {
           const initStarter = require(`./init-starter`)
           return initStarter(starter, { rootPath })
         }
+      ),
+    })
+    .command({
+      command: `telemetry`,
+      desc: `Enable or disable Gatsby anonymous analytics collection.`,
+      builder: yargs =>
+        yargs
+          .option(`enable`, {
+            type: `boolean`,
+            description: `Enable telemetry (default)`,
+          })
+          .option(`disable`, {
+            type: `boolean`,
+            description: `Disable telemetry`,
+          }),
+
+      handler: handlerP(({ enable, disable }) =>
+        setTelemetryEnabled(enable || !disable)
       ),
     })
     .wrap(cli.terminalWidth())
--- a/packages/gatsby-cli/src/init-starter.js
+++ b/packages/gatsby-cli/src/init-starter.js
@@ -7,7 +7,7 @@ const sysPath = require(`path`)
 const report = require(`./reporter`)
 const url = require(`url`)
 const existsSync = require(`fs-exists-cached`).sync
-
+const { trackCli, trackError } = require(`gatsby-telemetry`)
 const spawn = (cmd: string, options: any) => {
   const [file, ...args] = cmd.split(/\s+/)
   return execa(file, args, { stdio: `inherit`, ...options })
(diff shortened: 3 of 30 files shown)


packages/gatsby-telemetry/src/send.js
```diff
@@ -0,0 +1,15 @@
+const Telemetry = require(`./telemetry`)
+const instance = new Telemetry()
+
+const flush = _ => {
+  instance
+    .sendEvents()
+    .then(e => {
```
jlengstorf: Do you need this `.then` with a `.catch` below it?

packages/gatsby-telemetry/src/store.js
```diff
@@ -0,0 +1,76 @@
...
+
+    try {
+      renameSync(filePath, newPath)
+    } catch (e) {
+      // ignore
+      return
+    }
+    const contents = readFileSync(newPath, `utf8`)
+
+    // There is still a chance process dies while sending data and some events is lost
```
jlengstorf:
```suggestion
    // There is still a chance process dies while sending data and some events are lost
```

packages/gatsby-telemetry/src/telemetry.js
```diff
@@ -0,0 +1,196 @@
...
+    // Cache the result
+    if (this.trackingEnabled !== undefined) {
+      return this.trackingEnabled
+    }
+    let telemetry = this.store.getConfig(`telemetry`)
+    if (telemetry === undefined || telemetry === null) {
+      console.log(
+        `Gatsby has started collecting anonymous usage analytics to help improve Gatsby for all users.\n` +
+          `If you'd like to opt-out, you can use \`gatsby analytics --disable\`\n` +
+          `To learn more, checkout http://gatsby.dev/analytics`
```
jlengstorf: Should this link be updated to say `https://gatsby.dev/telemetry`?

packages/gatsby-telemetry/src/telemetry.js
```diff
@@ -0,0 +1,196 @@
...
+  isTrackingEnabled() {
+    // Cache the result
+    if (this.trackingEnabled !== undefined) {
+      return this.trackingEnabled
+    }
+    let telemetry = this.store.getConfig(`telemetry`)
+    if (telemetry === undefined || telemetry === null) {
+      console.log(
+        `Gatsby has started collecting anonymous usage analytics to help improve Gatsby for all users.\n` +
+          `If you'd like to opt-out, you can use \`gatsby analytics --disable\`\n` +
```
jlengstorf:
```suggestion
          `If you'd like to opt-out, you can use \`gatsby telemetry --disable\`\n` +
```

docs/docs/telemetry.md
```diff
@@ -0,0 +1,49 @@
...
+
+We track general usage details, including command invocation, build process status updates, performance measurements, and errors.
+We use these metrics to better understand the usage patterns. These metrics will directly allow us to better decide how to design future features and prioritize current work.
+
+Specifically, we collect the following information for _all_ telemetry events:
+
+- Timestamp of the occurrence
+- Command invoked (e.g. `build` or `develop`)
+- Unique and unidentifiable machine ID
+- Unique and unidentifiable session ID
```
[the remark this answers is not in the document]
KyleAMathews: 👍

packages/gatsby-telemetry/src/postinstall.js
```diff
@@ -0,0 +1,5 @@
+console.log(
```
KyleAMathews: This is going to run even if the user has already opted out.

Also, this will run every time the user installs Gatsby — even a normal `npm install gatsby` to upgrade — which seems noisy/excessive.

Perhaps we just include a global counter and show the notice a few times and then stop?

packages/gatsby-telemetry/README.md
```diff
@@ -0,0 +1,35 @@
...
+- Timestamp of the occurrence
+- Command invoked (e.g. `build` or `develop`)
+- Unique and unidentifiable machine ID
+- Unique and unidentifiable session ID
+- One-way hashed current working directory
+- General OS level information (operating system, version, CPU architecture, and whether the command is run inside a CI)
+- Used Gatsby version
+- Errors with PII (e.g., paths) sanitized.
+
+In addition to these events, we may collect additional performance affecting measurements, such as how many pages / queries etc. the build involved.
```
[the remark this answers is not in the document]
KyleAMathews: 👍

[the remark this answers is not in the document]
KyleAMathews: I'd assume this would just include API docs and link to the main doc page for the "why/what"

docs/docs/telemetry.md
```diff
@@ -0,0 +1,49 @@
...
+- One-way hash of the current working directory or a hash of the git remote
+- General OS level information (operating system, version, CPU architecture, and whether the command is run inside a CI)
+- Current Gatsby version
+
+The access to the raw data is highly controlled, and we cannot identify individual users from the dataset. It is anonymized and untraceable back to the user.
+
+## What about sensitive data? (e.g. secrets)
+
+We perform additional steps to ensure that secure data (e.g. environment variables used to store secrets for the build process) **do not** make their way into our analytics. We strip logs, error messages, etc. of this sensitive data to ensure we _never_ gain access to this sensitive data.
```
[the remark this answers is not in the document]
KyleAMathews: e.g. link to the test suite

docs/docs/telemetry.md
```diff
@@ -0,0 +1,49 @@
...
+
+Users may always opt-out from the telemetry with `gatsby telemetry --disable` or setting the environment variable `GATSBY_TELEMETRY_DISABLED` to `1`
+
+## Why?
+
+**Anonymous** aggregate user analytics allow us to prioritize fixes and features based on how and when people use Gatsby.
+Since much of Gatsby’s function revolves around community plugins and starters, we want to collect information on usage
+and reliability so that we can ensure a high-quality ecosystem.
+
+This raises a question. How will we use this telemetry data to improve the ecosystem? Some examples are helpful:
```
KyleAMathews: All these sentences need changed to present tense as when this PR is merged, things will be live.

packages/gatsby-telemetry/src/telemetry.js
pieh: Also this command will fail if user didn't update global `gatsby-cli`, so troubleshooting step (i.e. "if command is not recognized, update "gatsby-cli" with `npm i -g gatsby-cli` command")

packages/gatsby-telemetry/src/index.js
```diff
@@ -0,0 +1,34 @@
+const Telemetry = require(`./telemetry`)
```
pieh: this file seems to be missing from PR?

pieh: oh nvm - my bad

packages/gatsby-telemetry/src/telemetry.js
```diff
@@ -0,0 +1,196 @@
...
+  decorateNextEvent(event, obj) {
+    const cached = this.metadataCache[event] || {}
+    this.metadataCache[event] = Object.assign(cached, obj)
+  }
+
+  decorateAll(tags) {
+    this.defaultTags = Object.assign(this.defaultTags, tags)
+  }
+
+  setAnalyticsEnabled(enabled) {
```
pieh:
```suggestion
  setTelemetryEnabled(enabled) {
```

Seems like it was missed in analytics -> telemetry change (this is called with `setTelemtryEnabled` in https://github.com/gatsbyjs/gatsby/pull/12758/files#diff-54b3fd6936bf2f005bf1de5eb5ea79e0R24

packages/gatsby-telemetry/src/telemetry.js
```diff
@@ -0,0 +1,196 @@
...
+    this.trackingEnabled = enabled
+    this.store.updateConfig({ telemetry: enabled })
```
pieh: Should we just output confirmation to terminal that setting was changed to true/false?

packages/gatsby-telemetry/src/postinstall.js
jamo: @KyleAMathews great point, do we actually want to log this when installing, or should we only log this on the first time `gatsby` command is run?

packages/gatsby-telemetry/src/send.js
jamo: Indeed, `.catch` should suffice.

packages/gatsby-telemetry/src/telemetry.js
jamo:
```suggestion
          `To learn more, checkout http://gatsby.dev/telemetry`
```

packages/gatsby-telemetry/src/postinstall.js
jamo: @KyleAMathews I have now updated this, but I'm not sure if checking and updating the config is the approach we want to take, especially I'm uncertain about updating the config here.

packages/gatsby-telemetry/src/telemetry.js
jamo: Thanks!

packages/gatsby-telemetry/src/postinstall.js
KyleAMathews: yeah, I think just showing when installing the `gatsby-cli` is probably enough

KyleAMathews: We don't show any other postinstall messages there so it should stand out enough to be read.

docs/docs/telemetry.md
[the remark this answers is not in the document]
jamo: Updated, but it doesn't sound as great I'd like it to

packages/gatsby-telemetry/README.md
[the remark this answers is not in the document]
jamo: Indeed, this is superseded with the docs article.

packages/gatsby-telemetry/src/telemetry.js
jamo: Great idea. done

docs/docs/telemetry.md
jamo: or when we finally take these steps.

packages/gatsby-cli/src/create-cli.js
```diff
@@ -40,6 +45,11 @@ function buildLocalCommands(cli, isLocalSite) {
         `gatsby`,
         `package.json`
       ))
+      try {
+        setDefaultTags({ installedGatsbyVersion: packageInfo.version })
+      } catch (e) {
+        // ignore
```
wardpeet: when does this throw an error? unsure if we need to try catch

packages/gatsby-cli/src/init-starter.js
```diff
@@ -147,18 +147,22 @@ module.exports = async (starter: string, options: InitOptions = {}) => {
 
   const urlObject = url.parse(rootPath)
   if (urlObject.protocol && urlObject.host) {
+    trackError(`NEW_PROJECT_NAME_MISSING`)
```
wardpeet: i'm wondering if we should move into an enum or leverage typescript types for these strings. Should we also namespace these like `STARTER_NEW_PROJECT_NAME_MISSING` or would this be overkill as we won't have that many events.

packages/gatsby-telemetry/src/event-storage.js
```diff
@@ -0,0 +1,64 @@
...
+  store = new Store()
+  debugEvents = !!process.env.GATSBY_TELEMETRY_DEBUG
+  disabled = !!process.env.GATSBY_TELEMETRY_OPTOUT
+
+  addEvent(event) {
+    if (this.disabled) {
+      return
+    }
+
+    if (this.debugEvents) {
```
wardpeet: wondering if debugEvents should log and capture it instead of just logging.

packages/gatsby-telemetry/src/event-storage.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  async submitEvents(events) {
+    try {
+      const res = await fetch(`https://analytics.gatsbyjs.com/events`, {
+        method: `POST`,
+        headers: { "content-type": `application/json` },
+        body: JSON.stringify(events),
+      })
+      if (res.ok) {
```
wardpeet: you can omit this if (res.ok is a boolean)

packages/gatsby-telemetry/src/telemetry.js
```diff
@@ -0,0 +1,197 @@
...
+
+module.exports = class AnalyticsTracker {
+  store = new EventStorage()
+  debouncer = {}
+  metadataCache = {}
+  defaultTags = {}
+  osInfo // lazy
+  trackingEnabled // lazy
+  componentVersion
+  sessionId
```
wardpeet:
```suggestion
  sessionId = uuid()
```

packages/gatsby-telemetry/src/telemetry.js
```diff
@@ -0,0 +1,197 @@
...
+  constructor() {
+    this.sessionId = uuid()
```
wardpeet:
```suggestion
```

packages/gatsby-telemetry/src/telemetry.js
```diff
@@ -0,0 +1,197 @@
...
+
+  setTelemetryEnabled(enabled) {
+    this.trackingEnabled = enabled
+    this.store.updateConfig(`telemetry.enabled`, enabled)
+    console.log(`Telemetry collection`, enabled ? `enabled` : `disabled`)
+  }
+
+  async sendEvents() {
+    if (!this.isTrackingEnabled()) {
+      return async _ => {} // so that then/await won't fail
```
wardpeet: Wouldn't this suffice as well? (async seems a bit more complex to read)
```suggestion
      return Promise.resolve()
```

packages/gatsby/src/commands/build.js
```diff
@@ -26,6 +28,11 @@ type BuildArgs = {
 module.exports = async function build(program: BuildArgs) {
   initTracer(program.openTracingConfigFile)
 
+  telemetry.trackCli(`BUILD_START`)
+  signalExit(() => {
```
wardpeet: Cool plugin! this makes it so much easier!

packages/gatsby-telemetry/src/telemetry.js
```diff
@@ -185,6 +185,7 @@ module.exports = class AnalyticsTracker {
   setTelemetryEnabled(enabled) {
     this.trackingEnabled = enabled
     this.store.updateConfig(`telemetry.enabled`, enabled)
+    console.log(`Telemetry collection`, enabled ? `enabled` : `disabled`)
```
jlengstorf: Stray log?

packages/gatsby-telemetry/README.md
```diff
@@ -1,35 +1,33 @@
...
-- One-way hashed current working directory
-- General OS level information (operating system, version, CPU architecture, and whether the command is run inside a CI)
-- Used Gatsby version
-- Errors with PII (e.g., paths) sanitized.
+### setDefaultTags(tags)
 
-In addition to these events, we may collect additional performance affecting measurements, such as how many pages / queries etc. the build involved.
+Set additional tags to be included in all future events.
+
+### decorateEvent(type, tags)
```
jlengstorf: Could we add arg types here? Or a description of what type each arg should be? Not blocking, but probably useful for future maintainers.

packages/gatsby-cli/src/create-cli.js
jamo: We could get around with a couple of iffs too.

packages/gatsby-cli/src/init-starter.js
jamo: These are also mostly freeform events

packages/gatsby-telemetry/src/event-storage.js
jamo: lol, looks like I partially had this fixed already. Fixed now.

docs/docs/telemetry.md
KyleAMathews: I think it's safe to change it to present tense now. As soon as data starts coming in, we're going to be using it for many of these ideas

packages/gatsby-cli/src/create-cli.js
wardpeet: no biggy just can't think of any way this would throw

packages/gatsby-telemetry/src/event-storage.js
jamo: This felt more natural when implementing it, it has the benefit of having ability of introspect the events w/o sending any data. And it's the same behaviour as with homebrew analytics.

packages/gatsby-cli/src/create-cli.js
jamo: yeah, I think it's just me playing extra safe to minimize places where the instrumentation would cause any issues

packages/gatsby-telemetry/src/index.js
```diff
@@ -0,0 +1,34 @@
...
+const flush = require(`./flush`)
+
+const instance = new Telemetry()
+
+process.on(`exit`, flush)
```
wardpeet: do we wanna use the signalExit here too?

packages/gatsby-telemetry/src/store.js
```diff
@@ -0,0 +1,76 @@
...
+  readFileSync,
+  renameSync,
+  existsSync,
+  unlinkSync,
+} = require(`fs`)
+const { ensureDirSync } = require(`fs-extra`)
+const Configstore = require(`configstore`)
+
+module.exports = class Store {
+  config = new Configstore(`gatsby`, {}, { globalConfigPath: true })
```
sidharthachatterjee: We might want to wrap this in a try catch for certain environments when permissions don't allow writing to ~/.config. 

Not blocking and can do later though!

sidharthachatterjee: Actually, never mind. This should be fine since it falls back to os.tmpdir() if no XDG Base Directory is found

https://github.com/yeoman/configstore/blob/0df1ec950d952b1f0dfb39ce22af8e505dffc71a/index.js#L11

packages/gatsby-telemetry/src/store.js
```diff
@@ -0,0 +1,76 @@
...
+    this.config.set(...fields)
+  }
+
+  appendToBuffer(event) {
+    const bufferPath = this.getBufferFilePath()
+    appendFileSync(bufferPath, event, `utf8`)
+  }
+
+  getConfigPath() {
+    const configPath = path.join(homedir(), `.config/gatsby`)
```
sidharthachatterjee: This isn't always guaranteed to be home dir since it falls back to `os.tmpdir()` (see comment above)

Could get path from config.path (https://github.com/yeoman/configstore#path)

jamo: this should be ok for now

packages/gatsby-telemetry/src/postinstall.js
```diff
@@ -0,0 +1,16 @@
+let enabled = false
+try {
+  const Configstore = require(`configstore`)
+  const config = new Configstore(`gatsby`, {}, { globalConfigPath: true })
+  enabled = config.get(`telemetry.enabled`)
+  if (enabled === undefined) {
+    console.log(
+      `Gatsby has started collecting anonymous usage analytics to help improve Gatsby for all users.\n` +
+        `If you'd like to opt-out, you can use \`gatsby telemetry --disable\`\n` +
+        `To learn more, checkout http://gatsby.dev/telemetry`
```
DSchau: this should be https

PR #11992, opened 2019-02-22
chore(gatsby-source-filesystem): Upgrade chokidar to v2

sidharthachatterjee:

PR #12759, opened 2019-03-22
fix(gatsby-source-filesystem): pin  chokidar@2.1.2 to fix unix issues

DSchau:
## Description

Fixes apparent regression introduced in #11992

## Related Issues

Preliminary guess is that something changed in paulmillr/chokidar#791 (specifically, chokidar@2.1.3) which breaks some fs operations/reads on Linux. Doesn't seem to impact MacOS.

### DIFF
--- a/packages/gatsby-plugin-page-creator/package.json
+++ b/packages/gatsby-plugin-page-creator/package.json
@@ -21,7 +21,7 @@
   "dependencies": {
     "@babel/runtime": "^7.0.0",
     "bluebird": "^3.5.0",
-    "chokidar": "^2.1.2",
+    "chokidar": "2.1.2",
     "fs-exists-cached": "^1.0.0",
     "glob": "^7.1.1",
     "lodash": "^4.17.10",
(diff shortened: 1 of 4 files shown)


DSchau:
The e2e tests will _probably_ still fail here--because we're not installing new versions of dependencies in the e2e tests themselves.

DSchau:
Let's close for now--it seems like chokidar may have published a patch version.

DSchau:
e2e tests are _expected_ to fail because we don't update the version of deps (currently).

Let's merge and publish.

DSchau:
```
Successfully published:
 - gatsby-plugin-page-creator@2.0.11
 - gatsby-source-contentful@2.0.42
 - gatsby-source-drupal@3.0.32
 - gatsby-source-filesystem@2.0.28
 - gatsby-source-mongodb@2.0.16
 - gatsby-source-shopify@2.0.22
 - gatsby-source-wordpress@3.0.49
 - gatsby@2.2.8
```

paulmillr:
just Released 2.1.5.

PR #12761, opened 2019-03-22
[WIP]: chore(*): publish es versions of browser packages

DSchau:
## Description

In preparation for modern builds, we need to also publish some of our dependencies as es. We use microbundle to do that.

Currently blocked by microbundle using buble, which isn't able to minify everything as we'd want (e.g. class is still transpiled to function). This is supposed to be merged in next week or so upstream--so we'll wait for that.

Preliminary findings are that this shaves off a few kb (non-gzipped!) compared to the CommonJS transpiled version.

Note: We'll want to publish babel-preset-gatsby in a separate PR _before_ this change.

### DIFF
--- a/packages/babel-preset-gatsby-package/index.js
+++ b/packages/babel-preset-gatsby-package/index.js
@@ -1,12 +1,20 @@
 const r = require(`./resolver`)
 
-function preset(context, options = {}) {
-  const { browser = false, debug = false, nodeVersion = `8.0` } = options
+const modernConfig = {
+  loose: true,
+  targets: {
+    esmodules: true,
+  },
+  useBuiltIns: false,
+}
+
+function preset(_, options = {}) {
+  const { browser = false, debug = false, nodeVersion = `8.0`, modern = process.env.MODERN } = options
   const { NODE_ENV, BABEL_ENV } = process.env
 
   const PRODUCTION = (BABEL_ENV || NODE_ENV) === `production`
 
-  const browserConfig = {
+  const browserConfig = modern ? modernConfig : {
     useBuiltIns: false,
     targets: {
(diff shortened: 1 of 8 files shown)


DSchau:
@wardpeet this is probably good to re-visit now that your microbundle PR was merged, right? Or at least I thought it was merged 🙃

wardpeet:
Definitly! has some babel-preset changes We should install microbundle@next which has the latest updates.

freiksenet:
Is anything else needs to be done here @DSchau?

DSchau:
@freiksenet I'll need to pull it down again to check if microbundle has been updated.

I believe @wardpeet was making some upstream changes to enable Babel (instead of Buble) which will unblock us and we can ship this!

DSchau:
OK, mostly working. Few notes:

- Need to solve withIEPolyfill transpilation in gatsby-image (microbundle —entry fails with `Error: You must set "output.dir" instead of "output.file" when generating multiple chunks` as if it’s trying to parse the dynamic import statement or something; we just want it ignored
- Still (weirdly) transpiling classes, so we’ll want to ensure the `esmodules` target makes sense

Some preliminary numbers (note: these are non-gzipped; the actual numbers are not as drastic)

|Package|Format|Size|
|:-----:|------|:--:|
|`gatsby-image`|`cjs`|`15kb`|
|`gatsby-image`|`mjs`|`8.7kb`|
|`gatsby-link`|`cjs`|`5.8kb`|
|`gatsby-link`|`mjs`|`3.8kb`|

DSchau:
Let's close this -- when we're closer to implementing modern bundles, this will serve as a great reference implementation of one piece that will need to be done to maximize the benefit to end users.

PR #12763, opened 2019-03-22
do not merge - just testing something

pieh:
this is just for testing purposes - I will close this later

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,6 +50,7 @@ aliases:
   test_template: &test_template
     steps:
       - checkout
+      - run: node -v
       - run: ./scripts/assert-changed-files.sh "packages/*|.circleci/*"
       - <<: *restore_cache
       - <<: *install_node_modules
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -50,6 +50,7 @@ aliases:
   test_template: &test_template
     steps:
       - checkout
+      - run: node -v
```
DSchau: 😬 hopefully the executors are working like we'd expect! Good idea.

pieh:
This is just to test something that came up in https://github.com/gatsbyjs/gatsby/pull/12719 - because according to info there - it should fail for node6 (as some module is node8+ only), but it passed there :S

pieh:
It reports expected node version - so not sure why test in PR linked above doesn't fail in some ways if it needs node 8+ stuff - hmm, let's check something more

DSchau:
@pieh that package could be following our approach, haha. e.g. "soft deprecating" but still supporting?

pieh:
That's also possible. Maybe just parts of the package that we use might not require 8+?

pieh:
Ok, so tests that is designed to fail in node 6 and pass in node 8+ works - so our tests doesn't seem to be wonky after all

PR #12775, opened 2019-03-22
fix(gatsby): Allow setting nullability in createResolvers

stefanprobst:
Currently we don't allow overriding the field type in `createResolvers`. This PR loosens this restriction and allows setting nullability.

### DIFF
--- a/packages/gatsby/src/schema/schema.js
+++ b/packages/gatsby/src/schema/schema.js
@@ -245,7 +245,7 @@ const createTypeComposerFromGatsbyType = ({
       return InterfaceTypeComposer.createTemp(type.config, schemaComposer)
     }
     default: {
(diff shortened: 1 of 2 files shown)


packages/gatsby/src/schema/schema.js
```diff
@@ -245,7 +245,7 @@ const createTypeComposerFromGatsbyType = ({
       return InterfaceTypeComposer.createTemp(type.config, schemaComposer)
     }
     default: {
-      console.warn(`Illegal type definition: ${JSON.stringify(type.config)}`)
+      report.warn(`Illegal type definition: ${JSON.stringify(type.config)}`)
```
muescha: if you replace it here. Maybe also replace it in this line to have all on `report`:

`// console.log(printSchema(schema))`

PR #12779, opened 2019-03-23
feat(docs): add using gatsby professionally section

calcsam:
@lindaleebumblebee's excellent posts on the blog recently got us thinking -- why don't we have a whole section in the documentation about how to use Gatsby professionally, explain Gatsby to other stakeholders in a website project, and use Gatsby at scale?


muescha:
How about to place it into a subfolder?

calcsam:
@muescha We've created this as a top-level item since it's a set of guides diving in-depth to a specific topic (similar to the Behind the Scenes docs in this way).

calcsam:
Thanks @DSchau! Yep, ran Prettier before pushing!

Issue #12782, opened 2019-03-23
Plugin option to access renamed dataLayer for gatsby-plugin-google-tagmanager


spwisner:
## Summary

Google Tag Manager allows users to rename the dataLayer in GTM (https://developers.google.com/tag-manager/devguide#renaming).  The current configuration of gatsby-plugin-google-tagmanager does not accommodate a renamed dataLayer and data pushed to a renamed dataLayer is inaccessible.  The proposed feature would include a plugin option (which would be optional) in gatsby-config.js to include the name of a renamed dataLayer, thereby making this data available to GTM.  

### Basic example

I will provide a real life example of why this feature is needed.  I used Gatsby to build landing pages that included an embedded interactive form.  This form has a GTM container configured to push events to a renamed dataLayer (named "bzDataLayer").  My employer renamed the dataLayer to "bzDataLayer" for the embedded application because it is hosted on numerous partner websites and there was concern that failure to do so could cause events from our form to be pushed to the dataLayer of a partner's GTM container and create conflicts.

The landing pages, which was configured in Gatsby to use a different GTM container, cannot access the custom events pushed to the form's "bzDataLayer."  As a result, I had to modify gatsby-plugin-google-tagmanager to make the renamed dataLayer accessible (see associated pull request for details).

### Motivation

Being unable to access custom events in a renamed dataLayer was a blocker for releasing the landing pages.  Although it may be relatively uncommon for developers to rename dataLayer, it is included as an option in the GTM documentation (https://developers.google.com/tag-manager/devguide#renaming).  As a result, it should be accounted for in Gatsby plugin.  Moreover, because of the prevalence and importance of GTM  (it was a blocker for our release), every effort should be made to ensure that gatsby-plugin-google-tagmanager can be configured to address all the needs of users.

sidharthachatterjee:
Published in gatsby-plugin-google-tagmanager@2.0.11

Thank you @spwisner 🎉

PR #12783, opened 2019-03-23
feat(gatsby-plugin-google-tagmanager): Add plugin option for custom dataLayer name

spwisner:
## Description

This pull request provides a plugin option that will make data in a renamed dataLayer accessible (see #12782 for details).  The value assigned to "dataLayerName" is simply the renamed dataLayer (e.g. "bzDataLayer").  Then, any data pushed to a renamed dataLayer (e.g. "bzDataLayer") will also be pushed to dataLayer, thereby making it accessible through a standard GTM configuration.  

## Related Issues

Addresses #12782

### DIFF
--- a/packages/gatsby-plugin-google-tagmanager/src/gatsby-ssr.js
+++ b/packages/gatsby-plugin-google-tagmanager/src/gatsby-ssr.js
@@ -41,11 +41,8 @@ exports.onRenderBody = (
             new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
             j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
             'https://www.googletagmanager.com/gtm.js?id='+i+dl+'${environmentParamStr}';f.parentNode.insertBefore(j,f);
-            })(window,document,'script','${
-              pluginOptions.dataLayerName
-                ? pluginOptions.dataLayerName
-                : `dataLayer`
-            }', '${pluginOptions.id}');`,
+            })(window,document,'script','${pluginOptions.dataLayerName ||
+              `datalayer`}', '${pluginOptions.id}');`,
         }}
       />,
     ])
--- a/packages/gatsby-plugin-google-tagmanager/src/gatsby-ssr.js
+++ b/packages/gatsby-plugin-google-tagmanager/src/gatsby-ssr.js
@@ -19,20 +19,6 @@ exports.onRenderBody = (
         : ``
 
     setHeadComponents([
-      ...(pluginOptions.dataLayerName
-        ? [
-            <script
-              key="plugin-google-tagmanager-dataLayerName"
-              dangerouslySetInnerHTML={{
-                __html: stripIndent`
-                  // pre-populate the GTM dataLayer
-                  var dataLayer = dataLayer || [];
-                  // set the dataLayerName to the existing dataLayer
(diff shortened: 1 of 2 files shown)


packages/gatsby-plugin-google-tagmanager/src/gatsby-ssr.js
```diff
@@ -27,7 +41,11 @@ exports.onRenderBody = (
             new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
             j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
             'https://www.googletagmanager.com/gtm.js?id='+i+dl+'${environmentParamStr}';f.parentNode.insertBefore(j,f);
-            })(window,document,'script','dataLayer', '${pluginOptions.id}');`,
+            })(window,document,'script','${
+              pluginOptions.dataLayerName
+                ? pluginOptions.dataLayerName
+                : `dataLayer`
```
muescha: Maybe shorter? 
```
pluginOptions.dataLayerName || `datalayer`
```

spwisner: Fair enough and I'll make the change.  Thanks for your help.

packages/gatsby-plugin-google-tagmanager/src/gatsby-ssr.js
```diff
@@ -19,6 +19,20 @@ exports.onRenderBody = (
         : ``
 
     setHeadComponents([
+      ...(pluginOptions.dataLayerName
```
sidharthachatterjee: We shouldn't need to do this right?

packages/gatsby-plugin-google-tagmanager/src/gatsby-ssr.js
```diff
@@ -27,7 +41,8 @@ exports.onRenderBody = (
             new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
             j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
             'https://www.googletagmanager.com/gtm.js?id='+i+dl+'${environmentParamStr}';f.parentNode.insertBefore(j,f);
-            })(window,document,'script','dataLayer', '${pluginOptions.id}');`,
+            })(window,document,'script','${pluginOptions.dataLayerName ||
```
sidharthachatterjee: In context of the comment above, this should be enough according to https://developers.google.com/tag-manager/devguide

packages/gatsby-plugin-google-tagmanager/src/gatsby-ssr.js
spwisner: You are correct that it is not required for renaming the dataLayer.  The reason I included this was because of my use case where my embedded React app, which used a different GTM container with the renamed dataLayer, needed to be pushed to the second GTM container configured in Gatsby (which used the default name dataLayer).  However, because most users of the plugin will only have one GTM container and will not encounter this use case, it should be removed.  I'll take this part out now and will update the changes.  Thanks for your help.

packages/gatsby-plugin-google-tagmanager/src/gatsby-ssr.js
spwisner: That's correct. I pushed my changes.

sidharthachatterjee:
Published in gatsby-plugin-google-tagmanager@2.0.11

PR #12791, opened 2019-03-23
chore(docs/unit-testing): remove invalid props from link mock

appleJax:
## Description
I followed the Unit-Testing docs to set up jest with Gatsby, and I ran into the following error related to Gatsby's `Link` component:
```
React does not recognize the `activeStyle` prop on a DOM element.
```

## Changes
Removed the following props (which are invalid for an `a` tag) from the link mock:
- `activeClassName`
- `activeStyle`
- `getProps`
- `innerRef`
- `ref`
- `replace`

## Duplicate of #12445 (that PR was made through the GitHub editing interface and is failing for reasons that are unresolvable in that context)

### DIFF
--- a/docs/docs/unit-testing.md
+++ b/docs/docs/unit-testing.md
@@ -132,11 +132,22 @@ const gatsby = jest.requireActual("gatsby")
 module.exports = {
   ...gatsby,
   graphql: jest.fn(),
-  Link: jest.fn().mockImplementation(({ to, ...rest }) =>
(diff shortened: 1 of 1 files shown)


DSchau:
@appleJax sorry about this! We have an upstream dependency (`js-yaml`) that npm audit marks as invalid.

I'll close this, and merge the other one--since it's clear that that PR didn't actually cause that failure.

Thank you!

PR #12757, opened 2019-03-22
refactor: (gatsby-plugin-manifest) misc code and doc updates

moonmeister:
## Description

Code and Docs updates for `gatsby-plugin-manifest`

- Finished moving to module imports
- Finished migrating off of `bluebird` and onto async/await (order of magnitude speed increase)
- Major doc updates
- Improve logging output for generation

PR #12794, opened 2019-03-23
feat(gatsby-plugin-manifest): add icon_options as an option to support the purpose property

### LINKED ISSUE #12793: [plugin-manifest] Option to generate maskable favicons
## Summary

Currently, [icon masks](https://www.w3.org/TR/appmanifest/#icon-masks) are not supported by gatsby-plugin-favicon, thus, Android releases with circular icons may present the application icon as a rectangle inside a circle:

image

### Basic example

Add a boolean `icon_maskable` option or an object-specifiable `icon_options` override to be merged with each item of the generated or manually-specified `icons` array, e.g.:

```js
// gatsby-config.js
module.exports = {
  plugins: [
    {
      options: {
        // ...
        icon: `src/images/icon.png`,
        icon_options: {
          purpose: `maskable`,
        },
      },
    },
  ],
}
```

### Motivation

PWAs with maskable icons feel more "native" than uncroppable rectangular logos.

kripod:
## Description

Adds an option to specify PWA icon options, e.g. [`purpose`](https://www.w3.org/TR/appmanifest/#purpose-member).

## Related Issues

Resolves #12793

### DIFF
--- a/packages/gatsby-plugin-manifest/README.md
+++ b/packages/gatsby-plugin-manifest/README.md
@@ -209,6 +209,7 @@ module.exports = {
         icon_options: {
           // For all the options available, please see:
           // https://developer.mozilla.org/en-US/docs/Web/Manifest
+          // https://w3c.github.io/manifest/#purpose-member
           purpose: `maskable`,
         },
       },
--- a/packages/gatsby-plugin-manifest/README.md
+++ b/packages/gatsby-plugin-manifest/README.md
@@ -208,7 +208,7 @@ module.exports = {
         icon: `src/images/icon.png`, // This path is relative to the root of the site.
         icon_options: {
           // For all the options available, please see:
-          // https://www.w3.org/TR/appmanifest/#imageresource-and-its-members
+          // https://developer.mozilla.org/en-US/docs/Web/Manifest
           purpose: `maskable`,
         },
       },
--- a/packages/gatsby-plugin-manifest/README.md
+++ b/packages/gatsby-plugin-manifest/README.md
@@ -188,6 +188,35 @@ module.exports = {
 }
 ```
 
+## Custom icon options
+
+In order to specify manifest options merged with each item of the `icons` array, `icon_options` may be used as follows:
+
+```js
+// in gatsby-config.js
+module.exports = {
+  plugins: [
+    {
+      resolve: `gatsby-plugin-manifest`,
+      options: {
+        name: `GatsbyJS`,
+        short_name: `GatsbyJS`,
+        start_url: `/`,
+        background_color: `#f7f0eb`,
+        theme_color: `#a2466c`,
+        display: `standalone`,
+        icon: `src/images/icon.png`, // This path is relative to the root of the site.
+        icon_options: {
+          // For all the options available, please see:
+          // https://www.w3.org/TR/appmanifest/#imageresource-and-its-members
+          purpose: `maskable`,
+        },
+      },
+    },
+  ],
+}
+```
+
 ## Legacy option
 
 iOS 11.3 added support for service workers but not the complete webmanifest spec. Therefore iOS won't recognize the icons defined in the webmanifest and the creation of `apple-touch-icon` links in `<head>` is needed. This plugin creates them by default. If you don't want those icons to be generated you can set the `legacy` option to `false` in plugin configuration:
--- a/packages/gatsby-plugin-manifest/src/gatsby-node.js
+++ b/packages/gatsby-plugin-manifest/src/gatsby-node.js
@@ -54,6 +54,11 @@ exports.onPostBootstrap = async (args, pluginOptions) => {
     manifest.icons = defaultIcons
   }
 
+  // Specify extra options for each icon (if requested).
+  manifest.icons.forEach(icon => {
+    Object.assign(icon, pluginOptions.icon_options)
+  })
+
   // Determine destination path for icons.
   const iconPath = path.join(`public`, path.dirname(manifest.icons[0].src))
 


wardpeet:
It might be handy to add some documentation on how maskable works or when to use it.

packages/gatsby-plugin-manifest/src/gatsby-node.js
```diff
@@ -54,6 +54,11 @@ exports.onPostBootstrap = async (args, pluginOptions) => {
     manifest.icons = defaultIcons
   }
 
+  // Specify extra options for each icon (if requested).
+  manifest.icons.forEach(icon => {
```
sidharthachatterjee: Another nit (feel free to ignore!)

I typically prefer map over forEach since it discourages mutations to Objects (which after many more additions, may manifest in hard to debug errors)

```
manifest.icons = manifest.icons.map(icon => ({
  ...icon,
  ...pluginOptions.icon_options,
}))
```

packages/gatsby-plugin-manifest/README.md
```diff
@@ -188,6 +188,35 @@ module.exports = {
...
+        name: `GatsbyJS`,
+        short_name: `GatsbyJS`,
+        start_url: `/`,
+        background_color: `#f7f0eb`,
+        theme_color: `#a2466c`,
+        display: `standalone`,
+        icon: `src/images/icon.png`, // This path is relative to the root of the site.
+        icon_options: {
+          // For all the options available, please see:
+          // https://www.w3.org/TR/appmanifest/#imageresource-and-its-members
```
sidharthachatterjee: Let's link to MDN instead? Probably simpler than W3
```suggestion
          // https://developer.mozilla.org/en-US/docs/Web/Manifest
```

packages/gatsby-plugin-manifest/README.md
```diff
@@ -188,6 +188,35 @@ module.exports = {
...
+      resolve: `gatsby-plugin-manifest`,
+      options: {
+        name: `GatsbyJS`,
+        short_name: `GatsbyJS`,
+        start_url: `/`,
+        background_color: `#f7f0eb`,
+        theme_color: `#a2466c`,
+        display: `standalone`,
+        icon: `src/images/icon.png`, // This path is relative to the root of the site.
+        icon_options: {
```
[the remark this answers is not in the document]
kripod: Not sure, I just followed the conventions of previously-added custom keys.

packages/gatsby-plugin-manifest/src/gatsby-node.js
kripod: I was thinking about the same approach, but in such a small variable scope, I don't think that mutations may cause any trouble. After all, mutations are also more performant and in this particular case, they even result in shorter code with better expression of intents.

kripod:
Thank you for the review and the suggestions! MDN is a good idea, but unfortunately, it isn't nearly as detailed as W3C in this scenario.

packages/gatsby-plugin-manifest/README.md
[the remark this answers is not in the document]
sidharthachatterjee: That's alright for now then!

sidharthachatterjee:
@kripod Yeah, that's true. Do you think we should add a small comment on how `maskable` works like @wardpeet recommended?

kripod:
@sidharthachatterjee yes, probably, until the docs of MDN get improved

moonmeister:
@kripod I'm all for over documenting. I say we leave MDN in the comment and link w3 in the docs as an additional resource. that should cover everyone's curiosities without re-documenting the manifest spec itself.

packages/gatsby-plugin-manifest/README.md
```diff
@@ -188,6 +188,35 @@ module.exports = {
...
+          // For all the options available, please see:
+          // https://developer.mozilla.org/en-US/docs/Web/Manifest
```
wardpeet:
```suggestion
          // https://developer.mozilla.org/en-US/docs/Web/Manifest
          // https://w3c.github.io/manifest/#purpose-member
```

moonmeister:
Sry guys, but this PR has some breaking bugs. I didn't take a close enough look before it was merged. Namely, if this object isn't specified it'll fail cause the "forEach" method doesn't exist on an undefined object. Second, the array is never deleted from what gets written to the Manifest file. 

I already have a clean up PR going to the Manifest code at #12757, I can add the fixes there. we just can't release without that PR then cause these will break everybody's build.

kripod:
@moonmeister Overriding `icons` with `icon_options` is intentional, as by default, only the most basic options are specified for each icon (e.g. individual size/format). Merging an undefined object is perfectly fine and has a well-specified behavior as far as I know.

moonmeister:
@kripod let me clarify: Currently, if I set my config like this: 
```js
{
  icon_options: {
    purpose: `maskable`,
  },
  icons: [
    {
      src: `icons/icon-48x48.png`,
      sizes: `48x48`,
      type: `image/png`,
      purpose: `all`,
    },
    {
      src: `icons/icon-128x128.png`,
      sizes: `128x128`,
      type: `image/png`,
    },
  ],
}
```

The resulting config will be:
```js
{
  icons: [
    {
      src: `icons/icon-48x48.png`,
      sizes: `48x48`,
      type: `image/png`,
      purpose: `maskable`, // notice despite setting this in the 'icon' array, it has been changed.
    },
    {
      src: `icons/icon-128x128.png`,
      sizes: `128x128`,
      type: `image/png`,
      purpose: `maskable`, //this is applied
    },
  ],
}
```

What I am suggesting is that we make the resulting config be this:

```js
{
  icons: [
    {
      src: `icons/icon-48x48.png`,
      sizes: `48x48`,
      type: `image/png`,
      purpose: `any`, // notice despite setting this in the 'icon_options' object, my setting from the 'icon' array has persisted.
    },
    {
      src: `icons/icon-128x128.png`,
      sizes: `128x128`,
      type: `image/png`,
      purpose: `maskable`, // this is is still applied
    },
  ],
}
```

kripod:
@moonmeister Thank you for the clear example! I agree with your points and support swapping the order of precedence.

wardpeet:
@moonmeister as discussed in the oss meeting it's not a breaking issue. Browsers are super smart and actually do progressive enhancement pretty great. It doesn't really matter if extra stuff is in the manifest so it's not a breaking change :)

PR #12797, opened 2019-03-23
chore(starters): add gatsby-starter-julia

niklasmtj:
## Description
I started with Gatsby a couple weeks ago and redesigned my website with it. I then had the idea to also create it as a starter guide for people who want a minimal blog. So here is my PR with the starter. I hope you like it.

### DIFF
--- a/docs/starters.yml
+++ b/docs/starters.yml
@@ -2082,3 +2082,18 @@
   features:
     - Showcase of portfolio items
     - About me page
+- url: https://gatsby-starter-julia.netlify.com/
+  repo: https://github.com/niklasmtj/gatsby-starter-julia
+  description: A minimal blog starter template built with Gatsby
+  tags:
+    - Markdown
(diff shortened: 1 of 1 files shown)


muescha:
You should add a Readme to Thema repo

niklasmtj:
That's what I thought I did [README](https://github.com/niklasmtj/gatsby-starter-julia/blob/master/README.md). Or do you mean a different one?

It still has a lot of the default README in it but I think the passages are brilliant to understand quickly how GatsbyJS works so I kept them

muescha:
Yes. I mean customize the default readme. Thanks

niklasmtj:
I did. 
Added the explanation of the `/src` directory, a new introduction and a couple things in the text.

I actually think that the README provided by the default starter is really great. And changing it too much wouldn't help to get people going with the starter as fast as possible.

Niklas

PR #12814, opened 2019-03-24
fix(schema): fix `types` for buildUnionType config object

pieh:
using `schema.buildUnionType` results in

```
Error: Union type <type_name} must define one or more member types.
```

This should fix it (TM)

I've also added thunk handling (not sure if needed). This was something I hit when I tried to debug initial error I saw without doing debugging first ;)


stefanprobst:
Thunks should not be necessary since types are provided as strings

PR #12816, opened 2019-03-24
feat(gatsby-source-contentful): provide gql type definitions using content model introspection and schema-customization

pieh:
This is work in progress. It works, but there are few things that need to be cleaned up before merging this.

This works currently (at least with `using-contentful` example) - you can swap currently used Contentful space configuration in `using-contentful/gatsby-config.js` to one that is currently commented out. This space doesn't have any content but all queries still work and you can build a site (even if it does not have any content apart from things that are hardcoded in react components).

## Potential blockers (discussion points)
- parent/children - this is big one. Right now transformers rely on actually having parent nodes. There are some notes in code (search for `// Long string (markdown)`)
- there is currently small issue in Contentful API with single reference fields when you setup validation to restrict possible content types. API currently doesn't pass that information back to the client (it does for multiple references fields). So right now we create unions for those fields (which can be seen by change in one of queries in `using-contentful` example)

I might have forgotten some things - I tried to add pretty detailed comments beside the code.

## Breaking changes:

This also introduce number of breaking changes:
- JSON fields return `GraphQLJSON` type:
   - JSON field was inferred before, but we can't replicate that with free-form json content. I also generally think returning it as JSON type and not object makes much more sense
   - RichText: `json` field inside richText field is preserved (to maintain somewhat compatibility) but all other fields are gone (one that was already deprecated) and others that were inferred (they were pretty useless anyway, as it's really hard to query all the fields from structured content like that)
 - Union type name for reference fields can be different - it's produced from used type names and is using same method as one that inferred union by Gatsby, but because it's not relying on data - there might be extra types there. This can break fragments

TO-DO:
- [x] adjust documentation - probably as last step, as things can still change
- [x] revert temporary changes in using-contentful example's gatsby-config

/cc @Khaledgarbaya @sbezludny @freiksenet @stefanprobst


pieh:
`using-contentful` deploys on netlify:
  - using default space used for example:
    - logs https://app.netlify.com/sites/using-contentful/deploys/5c97fe7241b8a30009b4968d
    - preview https://5c97fe7241b8a30009b4968d--using-contentful.netlify.com/ (same as https://using-contentful.gatsbyjs.org/ )
  - using temporary space with same content model but without any content:
    - logs https://app.netlify.com/sites/using-contentful/deploys/5c9800eda038550008e737e5
    - preview https://5c9800eda038550008e737e5--using-contentful.netlify.com/

Khaledgarbaya:
Hi @pieh,
This looks really Great

>there is currently small issue in Contentful API with single reference fields when you setup validation to restrict possible content types. API currently doesn't pass that information back to the client (it does for multiple references fields). So right now we create unions for those fields (which can be seen by change in one of queries in using-contentful example)

I think we can live with the Union type for now as it's the easiest approach for now. I don't think having the validation for single ref fields will happen anytime soon.

For the complex example I can definetly help with testing on complex examples

sbezludny:
Hi @pieh
> I think straight returning JSON of structured content make sense, as right now you need to query for...

Own [Contentful's GraphQL API](https://www.contentful.com/developers/docs/references/graphql/#/reference/schema-generation/rich-text ) exposes `json` field on rich text types. It is done to separate content from the link resolution (`links` field) which for this plugin is handled by contentful.js SDK.

Other than that, I don't see reasons for keeping this field.

pieh:
> Own [Contentful's GraphQL API](https://www.contentful.com/developers/docs/references/graphql/#/reference/schema-generation/rich-text) exposes `json` field on rich text types. It is done to separate content from the link resolution (`links` field) which for this plugin is handled by contentful.js SDK.

Oh, interesting - I might revisit this to be more inline with how Contentful GraphqQL have RichText fields set up (once we deal with changed that we need to make in `gatsby` package). I see value in having schema generated by `gatsby-source-contentful` and Yours to be as similar as possible (i.e. easier for users to migrate queries from one source to another).

Khaledgarbaya:
Hey @pieh Anything we can help with the unblock this?

pieh:
This is partly blocked by https://github.com/gatsbyjs/gatsby/pull/13028 - after that we will be able to get rid of parts where I was importing gatsby internals (which is not safe):
```
// Ideally we would have nice access to it - I also potentially missed something
// obvious and I don't need to import from gatsby internals (yikes #1)
const { link } = require(`gatsby/dist/schema/resolvers`)
// This probably will be handled by automatically adding fieldArgs/resolver
// to Date fields - for now importing directly from gatsby internals (yikes #2)
const { dateResolver } = require(`gatsby/dist/schema/types/date`)
```

I will also need to dial back unneeded back-references changes and address rich text types changes.

Was there any update on validation field for single references in Contentful API (didn't check in quite a bit)?

Khaledgarbaya:
> Was there any update on validation field for single references in Contentful API (didn't check in quite a bit)?

I don't it will happen anytime soon so we can live with a union to all content types

pieh:
This is getting closer after round of updates today. I updated description of PR. I also published `gatsby-source-contentful@schema-types` for people to try it.

I think this is ready for initial reviews

examples/using-contentful/gatsby-config.js
```diff
@@ -3,11 +3,22 @@ module.exports = {
     title: `Gatsby with Contentful`,
   },
   plugins: [
+    // {
```
Khaledgarbaya: Do we need these comments?

packages/gatsby-source-contentful/src/normalize.js
```diff
@@ -336,11 +295,12 @@ exports.createContentTypeNodes = ({
         }
       })
 
-      // Add reverse linkages if there are any for this node
+      // // Add reverse linkages if there are any for this node
       const foreignReferences = foreignReferenceMap[entryItem.sys.id]
       if (foreignReferences) {
         foreignReferences.forEach(foreignReference => {
           const existingReference = entryItemFields[foreignReference.name]
+          // console.log(`adding backreference`, foreignReference)
```
Khaledgarbaya: I think we can remove this too

packages/gatsby-source-contentful/src/gatsby-node.js
```diff
@@ -93,6 +94,10 @@ exports.sourceNodes = async (
     pluginConfig,
   })
 
+  if (schema && actions.createTypes) {
```
freiksenet: Do we need this check? Maybe we make a major upgrade and drop support for older Gatsby? It's kinda moot anyway, you aren't fixing anything for old versions, are you?

packages/gatsby-source-contentful/src/types.js
```diff
@@ -0,0 +1,406 @@
...
+    const textTypeName = makeTextNodeType(
+      makeTypeName(context.typeName),
+      context.fieldName
+    )
+
+    const textGQLTypeConfig = context.schema.buildObjectType({
+      name: textTypeName,
+      fields: {
+        [context.fieldName]: `String`,
+        // missing - how to properly add childMarkdownRemark field (and other children that
```
freiksenet: So children stuff happens outside normal inference flow, so if you add Remark children to Text types when creating them, it should work.

freiksenet: OK, disregard that.

I think we can add extension "potentialChildren" that allows specifying which children fields to create.

freiksenet: However, that doesn't solve children actually appearing inside the type.

packages/gatsby-source-contentful/src/types.js
```diff
@@ -0,0 +1,406 @@
...
+          extensions: {
+            link: {
+              from: `${context.fieldName}___NODE`,
+            },
+          },
+        }
+      } else if (contentTypes.length === 1) {
+        // this is single type - not sure if this should be special case
+        // (using regular object type)
+        // or should create union with single type?
```
freiksenet: Probably should create a Union with single type, to force a spread inside query. That would mean that if more content types are added, query is still backwards compatible.

packages/gatsby-source-contentful/src/types.js
```diff
@@ -0,0 +1,406 @@
...
+    }
+    return null
+  },
+  // Number - integer
+  Integer: () => `Int`,
+  // Number - decimal
+  Number: () => `Float`,
+  Location: () => makeTypeName(`Location`),
+
+  // TO-DO figure out how to attach fields args / resolver properly.
```
freiksenet: Is this comment moot?

packages/gatsby-source-contentful/src/types.js
```diff
@@ -0,0 +1,406 @@
...
+    return {
+      type: `Date`,
+      extensions: {
+        dateformat: {},
+      },
+    }
+  },
+
+  RichText: (field, context) => {
+    const richTextTypeName = _.camelCase(
```
freiksenet: If we only have json type here, why do we create a type at all? Is it for future use?

packages/gatsby-source-contentful/src/types.js
```diff
@@ -0,0 +1,406 @@
...
+  const typeConfigs = contentTypeItems.reduce((acc, contentType) => {
+    const fieldType = {
+      name: makeTypeName(contentType.name),
+      fields: contentType.fields.reduce(
+        (fields, field) => {
+          const gqlType = fieldTypeToGraphQLType(field, {
+            gqlTypes: extraTypes,
+            typeName: contentType.name,
+            fieldName: field.id,
+            addBackReference: ({ referencingType, referencedType }) => {
```
freiksenet: Cool!

packages/gatsby-source-contentful/src/types.js
Khaledgarbaya: Yes, to allow other transformers to kick in if needed

examples/using-contentful/gatsby-config.js
pieh: Added TO-DO in PR description to remove those changes later on - for now I use this to validate everything works when there is 0 content (schema is the same regardless if there is content or not, so queries do not break)

packages/gatsby-source-contentful/src/normalize.js
pieh: Thanks, yeah, need to remove that stray console.log (relict of debugging)

packages/gatsby-source-contentful/src/gatsby-node.js
pieh: Right, I already bumped peer dependency and marked PR as breaking change (so major version bump will be needed), so I think it's safe to drop this check

packages/gatsby-source-contentful/src/types.js
pieh: This is unsolved problem - not quite sure if it's blocker or not - let my edit PR description. 

My current thinking is that ideally I could add metadata that this node type represents/contains `text/markdown` and transformer plugins could subscribe to that and attach children node type to it.

I don't think we should really specify that children of this type will be specifically `RemarkMarkdown` (we would also need to specify `Mdx` as well) - it's not source plugin job to know what transformers are able to handle their nodes (and we will potentially miss 3rd party or custom transformer plugins this way)

packages/gatsby-source-contentful/src/types.js
pieh: Ah, not removed comment from before type extensions. Will remove

packages/gatsby-source-contentful/src/types.js
pieh: Also https://github.com/gatsbyjs/gatsby/pull/12816#issuecomment-476250725 (and my next comment).

@Khaledgarbaya https://github.com/contentful/gatsby-transformer-contentful-richtext is deprecated and https://www.contentful.com/developers/docs/tutorials/general/rich-text-and-gatsby/ seems like proper/canonical way to work with Rich Text now? So I'm not sure about transformers here. Also my changes actually remove ability to use transformers as JSON is not being created as new node now (but I can change that, if transformers are still wanted)

packages/gatsby-source-contentful/src/types.js
pieh: My biggest worry here is that unions and inline fragments needed to properly query them seem to be common source of confusion for people using graphql for the first time (graphiql doesn't help much - by default people think they can query jsut for `__typename` - but documenting how to query union fields should work here)

pieh:
> Is Markdown thing are blocker/regression?

I should be more clear - those are potential blockers not strict blockers - I wanted input from others on those.

> Do you often get Union created twice issue?

With current contentful, yeah - any reference field will now create union with same types (all the contentful types) due to limitation of Contentful API that doesn't let us know which types can be used in given reference field, as mentioned in PR description ("blocker" no 2)

It's easy enough to track created unions and make sure I don't try to recreate same union type - I wouldn't make this a blocker, just something that potentially Gatsby can make easier for plugin maintainers in the future

sami616:
Hi, i just installed this PR via ```gatsby-source-contentful@schema-types``` and when running develop i see all my fragments being printed to my terminal with the following errror alongside any that use the field ```title```:

```Fields "title" conflict because they return conflicting types String! and String. Use different aliases on the fields to fetch both if this was intentional.```

Have i done something stupid here? i have tried running the same query in ```http://localhost:8000/___graphql``` and all is good.

Thanks!

**Comment by pieh** [comment]
@sami616 This is interesting, I'm not sure when this could happen - did you have any type definitions in your own code (that could this mismatch)?

**Comment by sami616** [comment]
Nope! Real weird, took me a while to figure out where the problem was as the error was  so long

File: packages/gatsby-source-contentful/src/types.js (hunk 5 above)
Comment (KyleAMathews) [reply packages/gatsby-source-contentful/src/types.js]: graphiql explorer actually shows them which is really nice https://github.com/gatsbyjs/gatsby/issues/11602

File: packages/gatsby-source-contentful/src/types.js (hunk 7 above)
Comment (Khaledgarbaya) [reply packages/gatsby-source-contentful/src/types.js]: That's a valid point @pieh go for it

**Comment by sami616** [comment]
@pieh I have come across another thing, apologies if this is known, or if i'm misunderstanding something.

It seems that my query is failing with the bellow error when i have a single reference field.

```"Expected Iterable, but did not find one for field ContentfulNavigationItem.page."```

Here is my query:

```
{
  contentfulPage(slug: {eq: "/"}) {
    nav {
      __typename
      ... on ContentfulNavigation {
        heading 
        items { 
          label
	  # Error here at page field
          page {
            id
            title
            slug
          }
        }
      }
    }
  }
}

# `items` allows multiple entries of type ContentfulNavigationItem
# `page` accepts a single entry of ContentfulPage
```

**Comment by pieh** [comment]
@sami616 I didn't hit this error, so this is not known (to me at least) - are you able to share config for your Contentful space so I can debug this locally - either here or privately (i.e via DM - https://twitter.com/mipiechowiak)?

**Comment by sami616** [comment]
@pieh yes, will do - i have temporarily managed to work around this by setting the ```page``` field to a multi reference, but limiting the number of accepted instances to 1 in the contentful GUI.

**Comment by pieh** [comment]
> i have temporarily managed to work around this by setting the `page` field to a multi reference, but limiting the number of accepted instances to 1 in the contentful GUI.

Interesting.

This also seem like nice workaround for current limitation for single reference not handling picked types in validation options for the field (described in potential blocker no 2 in PR description)

File: packages/gatsby-source-contentful/src/types.js (hunk 6 above)
Comment (pieh) [reply packages/gatsby-source-contentful/src/types.js]: done

File: packages/gatsby-source-contentful/src/normalize.js (hunk 2 above)
Comment (pieh) [reply packages/gatsby-source-contentful/src/normalize.js]: Removed

File: packages/gatsby-source-contentful/src/types.js (hunk 5 above)
Comment (pieh) [reply packages/gatsby-source-contentful/src/types.js]: Yeah, graphiql explorer would indeed nullify my worry here:
Kapture 2019-05-22 at 21 53 14

File: packages/gatsby-source-contentful/src/types.js (hunk 4 above)
Comment (pieh) [reply packages/gatsby-source-contentful/src/types.js]: Update here for anyone interested. We will ship this without solving children/parent/transformer + schema customization completely, but will document how to workaround this (as temporary solution). The workaround depends on https://github.com/gatsbyjs/gatsby/pull/14116 (being able to define most of the type inside contentful plugin and allowing users to add `childMarkdownRemark` or `childMdx` in project's gatsby-node - and that part will be documented)

File: packages/gatsby-source-contentful/src/types.js (hunk 5 above)
Comment (pieh) [reply packages/gatsby-source-contentful/src/types.js]: https://github.com/gatsbyjs/gatsby/pull/14280 is close to be finished, so I will go ahead and change to use Union here so references will always create unions

**Comment by pieh** [comment]
Update on issue @sami616 reported - this is bug in my current implementation where field created for backreference (named `page`) overwrites field created from regular fields in content type (also named `page`). I need to add check to not create backreference field in some cases

**Comment by KyleAMathews** [comment]
>Update on issue @sami616 reported - this is bug in my current implementation where field created for backreference (named page) overwrites field created from regular fields in content type (also named page). I need to add check to not create backreference field in some cases

Perhaps pre/post-fix backreference field names? It's nice to make clear that they're not fields that the someone created in Contentful. I think I did this for the Drupal source plugin though I'm forgetting the convention I used.

**Comment by pieh** [comment]
> Perhaps pre/post-fix backreference field names? It's nice to make clear that they're not fields that the someone created in Contentful. I think I did this for the Drupal source plugin though I'm forgetting the convention I used.

That really could/should start discussion about normalizing how to do backreferences so this is consistent. Drupal plugin does use straight type name for backreferences field (so similar as current contentful, and this PR doesn't change that). But this doesn't actually tell users from which field it was referenced so for more complex cases it doesn't provide much value (I think). Initial idea is to do something like this:

```js
node: {
  backreferences: {
    TypeName: {
      FieldName: [list_of_nodes_that_are_referencing_current_one_using_given_field]
    }
  }
}
```

Currently it looks like this (in both Contentful and Drupal):
```js
node: {
  TypeName: [list_of_nodes_that_are_referencing_current_one_using_any_field]
}
```

sami616:
Hi @pieh i just came across something that i think falls under the blocker no 2 in the PR desc. When a single reference field is set to required - but the linked reference is unpublished we get the following error during build: 

```Cannot return null for non-nullable field [fieldname]```

My workaround of setting the filed to a multi ref field but limiting entries to 1 seems to fix this.

**Comment by pieh** [comment]
> Hi @pieh i just came across something that i think falls under the blocker no 2 in the PR desc. When a single reference field is set to required - but the linked reference is unpublished we get the following error during build:
> 
> `Cannot return null for non-nullable field [fieldname]`
> 
> My workaround of setting the field to a multi ref but limiting entries to 1 seems to fix this.

This also happens when using contentful preview:
non-nullable title

I don't have good solution for this other than ignoring "required" validation and not setting non-null decorator on graphql fields. But this would force users to do lot of null checking in their code.

I do think tho it makes sense that build would fail if referenced entry is not published - if field is required then we should enforce that I think

**Comment by pieh** [comment]
Ok, backref field names are streamlined now - backref field names will contain field name of referencing fields (not quite sure how to say it so it's not confusing)

Next step - docs updates

File: packages/gatsby-source-contentful/README.md [hunk 9]
```diff
@@ -303,6 +295,97 @@ To get **all** the `CaseStudy` nodes with ShortText fields `id`, `slug`, `title`
 
 When querying images you can use the `fixed`, `fluid` or `resize` nodes to get different sizes for the image (for example for using [gatsby-image](https://www.gatsbyjs.org/packages/gatsby-image/)). Their usage is documented at the [gatsby-plugin-sharp](https://www.gatsbyjs.org/packages/gatsby-plugin-sharp/) package. The only difference is that gatsby-source-contentful also allows setting only the `width` parameter for these node types, the height will then automatically be calculated according to the aspect ratio.
 
+## Notes on Contentful Content Models
+
+### References
+
+Contentful reference fields will create [Union type](https://graphql.org/learn/schema/#union-types) fields. To access data [Inline fragments](https://graphql.org/learn/queries/#inline-fragments) need to be used.
```
m-allanson:
```suggestion
Contentful reference fields will create [Union type](https://graphql.org/learn/schema/#union-types) fields. [Inline fragments](https://graphql.org/learn/queries/#inline-fragments) should be used to access the data on these fields.
```

packages/gatsby-source-contentful/README.md
```diff
@@ -303,6 +295,97 @@ To get **all** the `CaseStudy` nodes with ShortText fields `id`, `slug`, `title`
...
+
+For example, if you have `Product` and `Category` Content Models in Contentful space and `Product` has `categories` field that reference `Category` entries, you can get list of categories for product by querying:
```
m-allanson: I think showing the demo query and then explaining the requirements makes it easier to understand this?

```suggestion
See below for a references example where you query for categories on a product:
```

packages/gatsby-source-contentful/README.md
```diff
@@ -303,6 +295,97 @@ To get **all** the `CaseStudy` nodes with ShortText fields `id`, `slug`, `title`
...
+      # We are using inline fragment on Category type
+      # to access category fields
+      ... on ContentfulCategory {
+        title
+      }
+    }
+  }
+}
+```
+
```
m-allanson:
```suggestion
The above example assumes that you have `Product` and `Category` Content Models in your Contentful space,  and that `Product` has a `categories` field that references `Category` entries.

```

packages/gatsby-source-contentful/README.md
```diff
@@ -303,6 +295,97 @@ To get **all** the `CaseStudy` nodes with ShortText fields `id`, `slug`, `title`
...
+### Reverse references
+
+When using reference fields, be aware that this source plugin will automatically create the reverse reference. You do not need to create references on both content types. Name of reverse reference will contain field and type names.
```
m-allanson:
```suggestion
When using reference fields, be aware that this source plugin will automatically create the reverse reference. You do not need to create references on both content types. The name of the reverse reference will contain field and type names.
```

packages/gatsby-source-contentful/README.md
```diff
@@ -303,6 +295,97 @@ To get **all** the `CaseStudy` nodes with ShortText fields `id`, `slug`, `title`
...
+
+For example, if you have `Product` and `Category` Content Models in Contentful space and `Product` has `categories` field that reference `Category` entries, you can get list of all products in given category by quering:
```
m-allanson:
```suggestion
See below for a reverse references example where you query for all products in a given category:
```

packages/gatsby-source-contentful/README.md
```diff
@@ -303,6 +295,97 @@ To get **all** the `CaseStudy` nodes with ShortText fields `id`, `slug`, `title`
...
+  contentfulCategory {
+    title
+    categoriesProduct {
+      name
+      price
+    }
+  }
+}
+```
+
```
m-allanson:
```suggestion
The above example assumes that you have `Product` and `Category` Content Models in your Contentful space, and `Product` has a `categories` field that references `Category` entries.

```

packages/gatsby-source-contentful/README.md
```diff
@@ -303,6 +295,97 @@ To get **all** the `CaseStudy` nodes with ShortText fields `id`, `slug`, `title`
...
+For simplicity, it is easier to put the reference field on the child in child/parent relationships.
+
+### Markdown transformers on Long Text fields
+
+Version 3 of `gatsby-source-contentful` creates GraphQL schema directly from Contentful Content Model and is not inferring fields from data anymore. This means that queries won't break if you don't have any fields in Your contentful space. There is however edge case for transformed data - particularly for Long Text fields that can be transformed with `gatsby-transformer-remark` or `gatsby-mdx`. Contentful plugin is not aware of transformers so you might need to provide hints for gatsby to create `childMarkdownRemark`/`childMdx` fields.
```
m-allanson:
```suggestion
Version 3 of `gatsby-source-contentful` creates GraphQL schema directly from the Contentful Content Model and no longer infers fields from data. 

This means that queries will keep working even if your Contentful space doesn't have any fields in it. 

There is however, an edge case for transformed data - particularly for Long Text fields that can be transformed with `gatsby-transformer-remark` or `gatsby-mdx`. The Contentful plugin is not aware of transformers so you might need to provide hints for Gatsby to create `childMarkdownRemark`/`childMdx` fields.
```

packages/gatsby-source-contentful/README.md
```diff
@@ -303,6 +295,97 @@ To get **all** the `CaseStudy` nodes with ShortText fields `id`, `slug`, `title`
...
+
+Here's example of attaching `childMarkdownRemark`:
```
m-allanson:
```suggestion
Here's an example of attaching `childMarkdownRemark`:
```

packages/gatsby-source-contentful/README.md
```diff
@@ -349,5 +432,51 @@ documentToReactComponents(node.bodyRichText.json, options)
...
+      title
+  -   product {
+  +   categoriesProduct {
+        name
+        price
+      }
+    }
+  }
+  ```
+- Rich text fields now don't infer fields. Only available field is `json`. See [Contentful Rich Text](#contentful-rich-text) section how to use it.
```
m-allanson:
```suggestion
- Rich text fields now don't infer fields. Only available field is `json`. See [Contentful Rich Text](#contentful-rich-text) section on how to use it.
```

examples/using-contentful/gatsby-node.js
```diff
@@ -2,6 +2,43 @@ const _ = require(`lodash`)
 const path = require(`path`)
 const slash = require(`slash`)
 
+exports.sourceNodes = ({ actions, schema }) => {
+  actions.createTypes(`
+    # This is needed if there aren't any MarkdownRemark nodes.
+    type MarkdownRemark implements Node {
```
freiksenet:
```suggestion
    type MarkdownRemark implements Node @infer {
```

pieh:
Update on this:
I've tested this with more real-world scenario (contentful space used for gatsbyjs.com site) than space used for `using-contentful` example and demo spaces I created when developing and it caused severe memory (and perf issues). 

Used space has 23 content types which means that single references create unions of 23 types (because of limitation of Contentful API that doesn't provide validation information for single references). This coupled with fact that Gatsby generates sorting enums 3 levels deep and it creates values for each type in Unions, means there are 23 enums with ~23^3 values each (that's approximation - it is dependent on exact content types/fields)

Memory allocation chart:
Screenshot 2019-07-16 at 01 27 30

I printed out generated schema to .graphql file and this file turned out to be almost 200MB and have 2568743 lines (yes 2.5 million).

It doesn’t seem like PR is inherently wrong or buggy - just sheer amount of types/unions cause this.

In any case this PR is blocked until this is somehow handled - yet to figure out what's the correct way of handling this

pieh:
Also note on above - even if Contentful API provided validation field for single references - this probably wouldn't fully unblock this - because of how fields are created in Contentful UI - setting available types for reference field is pretty deep in field configuration: you need to go to field settings -> validations -> check "Accept only specified entry type" box and select type(s), it means that in many cases those won't be set. So I would mark this as having those would be very nice - but is not core issue here. We do need to limit amount of values that sorting enums have when there are nested union fields in our types

pieh:
Another update - I was wrong in initial hypothesis about Unions (but this is still linked to lack of types for single references).

New hypothesis is, this is caused by reverse/back links:

Any single-value reference right now will create reverse link field in all contentful types as we don't have complete information what type+field can link to.

So I'm thinking right now is to add (hopefully temporary) plugin options to define types that field can reference to. In case there is no information about possible types reference field can link to we would skip creating back references and show warning with rich context explaining the problem and showing how to use this new plugin option (and link to contentful docs on how to define possible types in case of multi-value references)

pieh:
WIP of rich warning I mentioned in previous comment
Screenshot 2019-07-17 at 04 35 07

pieh:
Missing things here right now:
 - docs updates related to `referenceTypes` plugin option added in latest comment
 - tests to verify behaviour (and also catch any regressions in future)

pieh:
I spend more time on this yesterday - mostly trying this in practice in I right now have doubts about decision to convert reference fields that allow single type to create 1-type unions.

While it makes sense to do this, given that at any time you can edit reference field to allow accepting more types - and gatsby changing from GraphQLObject to GraphQLUnion would break queries. But just looking at what queries with references allowing single type look like - it's really messy:

```diff
 export const FeaturesListFragment = graphql`
   fragment FeaturesListFragment on ContentfulSection {
     featuresList {
-      title
-      text {
-        childMarkdownRemark {
-          html
+      ... on ContentfulSection {
+        title
+        text {
+          childMarkdownRemark {
+            html
+          }
         }
-      }
-      keyvisual {
-        name
-      }
-      associatedPage {
-        name
-        slug
-        parentPage {
-          name
-          slug
+
+        keyvisual {
+          ... on ContentfulKeyvisual {
+            name
+          }
+        }
+        associatedPage {
+          ... on ContentfulPage {
+            name
+            slug
+            parentPage {
+              ... on ContentfulPage {
+                name
+                slug
+              }
+            }
+          }
         }
       }
     }
````

So this is one thing that should be discussed again I think. Here's link to earlier comments on this.

File: packages/gatsby-source-contentful/src/types.js (hunk 5 above)
Comment (pieh) [reply packages/gatsby-source-contentful/src/types.js]: I started to think if we should still do 1-type unions here. Was doing more testing on more real project and queries really looks messy with this - see https://github.com/gatsbyjs/gatsby/pull/12816#issuecomment-515762555

**Comment by arnorhs** [comment]
So, any word on this?

**Comment by sidharthachatterjee** [comment]
> Any single-value reference right now will create reverse link field in all contentful types as we don't have complete information what type+field can link to.

@arnorhs Blocked on the above

**Comment by marcuslindfeldt** [comment]
Is there currently any way of explicitly defining the contentful schema for at least some of the resources so that the build does not break when no resources have been created? 

Placeholder content in Contentful is a pretty ugly and fragile workaround as it's very easy to accidentally delete by e.g. an editor which would then break the build.

**Comment by brumm** [comment]
@marcuslindfeldt 

We're in the same boat and are using custom resolvers to provide defaults for missing contentful keys like so:

```js
const mapOfTypesAndAttributesWeNeedToFix = {
  ContentfulQuoteGroupItem: ['authorThumbnail', 'quoteImage'],
  ContentfulPage: ['pageTitle'],
  // ...
}

const resolvers = {}

for (const typeName in mapOfTypesAndAttributesWeNeedToFix) {
  const attributes = mapOfTypesAndAttributesWeNeedToFix[typeName]
  
  for (const attributeName of attributes) {
    resolvers[typeName] = resolvers[typeName] || {}
    
    resolvers[typeName][attribute] = {
      type: 'String',
      resolve(source) {
        if (!source[attribute]) {
          return null
        } else {
          return source[attribute]
        }
      },
    }
  }
}

exports.createResolvers = ({ createResolvers }) => {
  createResolvers(resolvers)
}
```

pieh:
While we are kind of blocked here with merging this one, maybe you could try checking if creating temporarily placeholder context to create schema snapshot with https://www.gatsbyjs.org/packages/gatsby-plugin-schema-snapshot/, and then removing placeholders and relying on generated snapshot would work?

mmellado:
> While we are kind of blocked here with merging this one, maybe you could try checking if creating temporarily placeholder context to create schema snapshot with https://www.gatsbyjs.org/packages/gatsby-plugin-schema-snapshot/, and then removing placeholders and relying on generated snapshot would work?

I can confirm this works. I created a Contentful environment in which I ensured to have at least one entry will all fields for each content type as well as a reference to all possible types of components in every multi relationship field. Generated the schema using the https://www.gatsbyjs.org/packages/gatsby-plugin-schema-snapshot/ plugin, then unpublished all entries for some content types. Build passed ignoring the content types with no entries as they were still generated in the GraphQL schema (they simply return no results). Null checks are still necessary but this at least allows me to have queries in `gatsby-node.js` to generate pages for content types which may not necessarily have entries all the time. Thanks!

kitos:
> I spend more time on this yesterday - mostly trying this in practice in I right now have doubts about decision to convert reference fields that allow single type to create 1-type unions.
> 
> While it makes sense to do this, given that at any time you can edit reference field to allow accepting more types - and gatsby changing from GraphQLObject to GraphQLUnion would break queries. But just looking at what queries with references allowing single type look like - it's really messy:

@pieh what about adding plugin option - `alwaysCastReferenceFieldsToUnions`?

bsgreenb:
Question: will this do anything to allow making Required-> Non_Null?  I'm using graphql simply for a static build and would like to verify my Required fields are present.  

I know there's issue of Preview environments.. I'm fine one-off handling these on the frontend.  I'm building for production and would like to be able to require things with typescript generated from graphql.

pieh:
> Question: will this do anything to allow making Required-> Non_Null? I'm using graphql simply for a static build and would like to verify my Required fields are present.

Current implementation does make Contentful required field as NonNull Graphql fields and I do want to keep it that way for production builds / when using "production" data from Contentful.

I'm not sure how exactly to make it to continue work with Contentful's Preview API yet (as queries would just fail if any of required field is not there). Maybe this is fine for them to not work/update results until all required fields are filled, but the DX when leaving it like that will be awful. One way we could make it work is if we use mock field values for required fields when using preview API (but not in production mode). So empty text field would be something like `<No content in required field "[fieldName]">` (scalar fields are pretty "easy" to mock, reference fields would be harder, but possible).

I guess question is - do we want types to change (probably not) or are we fine with mocked data for required fields to workaround the issue.

bsgreenb:
@pieh What if it was configurable on a per-environment basis what Required meant?  So for the server using preview environment, you could build with those fields being nullable based on a config param?

Would totally be fine with using default fallbacks for preview environment.

packages/gatsby-source-contentful/src/normalize.js
```diff
@@ -2,14 +2,13 @@ const _ = require(`lodash`)
 const crypto = require(`crypto`)
 const stringify = require(`json-stringify-safe`)
 const deepMap = require(`deep-map`)
+const { makeTypeName, makeTextNodeType } = require(`./types`)
 
 const digest = str =>
   crypto
     .createHash(`md5`)
     .update(str)
     .digest(`hex`)
```
muescha: new code style suggestion from:

#8805 Issue: feat: update gatsby plugins to use new createContentDigest helper

```suggestion
const digest = str => createContentDigest(str)
```
and add to imports:
```javascript
import { createContentDigest } from "gatsby-core-utils"
```

or better replace all usages of `digest` with `createContentDigest`

pvdz:
(I've converted this PR into a Draft PR. Although it's been a year since the start, is any progress expected at all?)

bsgreenb:
I want to follow up on this.  The Maybe<> that wraps everything pulled by the graphql code generator requires me to overwrite all my types, and gives me no info about fields that are actually optional or required.

https://gist.github.com/bsgreenb/05202d74c7ef079970b536fad0fc5c88

@pieh was concerned about preview environments, but why dont you just disallow previewing content that doesn't have required fields?  The only fields I require on a post are a title and a body.  I want the build to break if it doesn't have those, and and editor can easily enter them.

Here is what my type overwrites look like.  I don't think this gets prettier (aside from maybe rich text json) because we start using Gatsby's new API.

LekoArts:
@axe312ger @pieh Just to make sure: Is this change also needed with the new source plugin for Contentful and/or will this be part of other changes already?

axe312ger:
@LekoArts still needed and one of the highest ranking issues for us after getting rich text run with complex content models!

rshackleton:
> > I spend more time on this yesterday - mostly trying this in practice in I right now have doubts about decision to convert reference fields that allow single type to create 1-type unions.
> > While it makes sense to do this, given that at any time you can edit reference field to allow accepting more types - and gatsby changing from GraphQLObject to GraphQLUnion would break queries. But just looking at what queries with references allowing single type look like - it's really messy:
> 
> @pieh what about adding plugin option - `alwaysCastReferenceFieldsToUnions`?

Rather than using a union, could we use the base type `ContentfulReference` or `ContentfulEntry` as this supports using inline fragments to query fields for all potential types? This wouldn't suffer from issues with the type name changing.

We used a similar approach for the Kentico Kontent plugin by using the KontentItem type for all relationships.

This approach also works when patching the schema locally:

```js
function customiseHomeModel() {
  const typeDefs = `
    type ContentfulHome implements Node {
      metadata: ContentfulReference @link(by: "id", from: "metadata___NODE")
      sections: [ContentfulReference] @link(by: "id", from: "sections___NODE")
    }
  `;

  return typeDefs;
}
```

In the GraphQL query you can then use an inline fragment to pull any data you need:

```gql
query MyQuery {
  allContentfulHome {
    nodes {
      sections {
        contentful_id
        id
        ... on ContentfulArticle {
          id
          date
          slug
          title
        }
      }
    }
  }
}

```

axe312ger:
Closing this PR in favour of #30855.

Feel free to comment and test the upcoming major version.

PR #12822, opened 2019-03-25
[WIP]: plugin/theme configuration api 

moonmeister:
## Description

Here we go! First working example of how an `onInstall` (name is probably not final, please recommend better name) node API might work for plugins and themes self configuring on install. It's rough, but it's something. 

I've added an example of this API in use to `gatsby-plugin-manifest` for testing purposes.

Any feed back is appreciated.

Thanks to @jlengstorf and @DSchau for their assists on this.

## Testing

```
//on this branch
$ yarn bootstrap
...
$ yarn run watch --scope={gatsby,gatsby-plugin-manifest,gatsby-cli}
```

```
//in a test gatsby site 
$ gatsby-dev --packages gatsby-plugin-manifest gatsby

//in a new shell on the same test site

//add an alias to your shell to the dev cli
$ alias gatsby-l="/path/to/repo/packages/gatsby-cli/lib/index.js"

//dry-run is important to not override the dev copy of 'gatsby-plugin-manifest'
$ gatsby-l plugins add gatsby-plugin-manifest --dry-run
```
## Related Issues

#10294
RFC: https://github.com/gatsbyjs/rfcs/blob/master/text/0007-gatsby-plugin-cli.md


moonmeister:
I'm trying to work on exporting an updated config back to the config file. The trouble I'm having is this is JS and not JSON. Any suggestions on a good way to write JS to a file from an object?

moonmeister:
## Side affects of writing config:

- Comments are lost
- Backticks convert to `"`
- `${__dirname}/src/images` gets output as a complete path.
- if using a prettier config not in a prettier file it won't get picked up by prettier. Not sure how to solve this in the long run.

Generally writing to and from JS file and not JSON might be error prone. The current method is not perfect but works as a proof of concept.

## Other Things to solve:

- [ ] There is little (no real) validation happening on the name of the plugin. We should probably make some assumptions about plugin name conventions. Like like expecting gatsby-[plugin/source/etc?]-<purpose> and requiring override if it doesn't validate.
- [ ] Right now there are a couple assumptions made around what config is returned from the API. We could document this but need to make sure it fails cleanly and handles most reasonable use cases.

## Conclusion
People are welcome to start testing if they want. I haven't built out an example workflow of questions in the manifest plugin that's being used for demo. I'll do that soon.

moonmeister:
# Update

Just did a bunch of work to refactor the code and implemented writing the code to config file using ASTs.

## Gotchas
- Config is currently written to `gatsby-config-test.js` for dev comparison.
- Config is currently set using JS objects. This is converted to an AST for merging. This simplifies the work having to be done when the API is called to not concat a bunch of strings and values together to create a string version of the config for AST conversion. HOWEVER: This doesn't allow plugin authors to set things dynamically (e.g. `${__dirname}`, or `process.env.SOMETHING`) or include `//comments` in what's returned. The former being more important than the latter. This either needs a simplified solution that can handle these scenarios or we'll need to provide and "advanced" option to provide a config string for AST conversion so these things can be done. 

## Functionality
- Cleans config on Removal of plugin
- Add basic `"gatsby-config-manifest"` entry to plugins by default if there is no API cal in plugin.
- Handles adding/removing "Sub-plugins"
- Defines class for PluginConfig and provides defaults. plugin just has to add config or modify defaults on call.

## To Resolve
- [ ] How do we handle plugins like "gatsby-source-filesystem" that are used multiple times?
- [ ] Build out functionality to write `siteMetadata` if plugin needs
- [ ] Build out functionality and logic so existing config persists through reconfig and API caller doesn't have to handle keeping existing. This will also make existing config values available for setting "defaults" to questions by api caller.

CC: @sidharthachatterjee @wardpeet @DSchau

KyleAMathews:
Potential functions for manipulating the site plugins array in gatsby-config.js:

- `createPluginConfig` — would default to overwriting anything there or if boolean set, would add a new instance
- `setPluginConfigField` — overwrites or sets a field
- `deletePluginConfigField`
- `deletePluginConfigInstance` — deletes a plugin instance — defaults to the first instance but can also pass a zero-ordered number for the instance

DSchau:
As in #10294 (see [my reply](https://github.com/gatsbyjs/gatsby/pull/10294#issuecomment-494925097), as well), we aren't quite ready to merge this in as we'd like to establish API-level enhancements in Gatsby Core (for example, a plugin validation API, and likely an API to add a plugin with options to a config, etc.).

As such--let's close this out _for now_ but we absolutely hope you'll be willing in the future to help and offer advice, expertise, and of course _code_ to help us deliver similar feature(s) using this PR as a foundational base for the type of functionalities we'd love to enable.

PR #12824, opened 2019-03-25
fix(schema): Add tests for incorrect list counting

stefanprobst:
~Pulled out from #12806 since that needs more discussion.~ (merged)
Add tests for correct list counting.


sidharthachatterjee:
Published in gatsby@2.2.13

PR #12825, opened 2019-03-25
fix(schema): Don't default to tracking all nodes

freiksenet:
## Description

We accidentaly made `getAllNodes` track everything when called. It's not an expected behavioru, because this function is often used to manually filter nodes. Thus, we disable tracking unless you explicitly ask for it. This PR also adds manual tracking function.

### DIFF
--- a/packages/gatsby/src/schema/node-model.js
+++ b/packages/gatsby/src/schema/node-model.js
@@ -47,6 +47,10 @@ export interface NodeModel {
     pageDependencies?: PageDependencies
   ): Promise<any>;
   getTypes(): Array<string>;
+  trackPageDependencies<nodeOrNodes: Node | Node[]>(
+    result: nodeOrNodes,
+    pageDependencies?: PageDependencies
+  ): nodesOrNodes;
 }
 
 class LocalNodeModel {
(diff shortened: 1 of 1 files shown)


packages/gatsby/src/schema/node-model.js
```diff
@@ -142,11 +140,11 @@ class LocalNodeModel {
...
     }
 
-    return trackPageDependencies(
-      result,
-      { path: this.path, ...pageDependencies },
-      this.createPageDependency
-    )
+    if (pageDependencies) {
+      return this.trackPageDependencies(result, pageDependencies)
+    } else {
```
wardpeet: no need to do an else statement

stefanprobst:
LGTM!

packages/gatsby/src/schema/node-model.js
```diff
@@ -224,6 +218,35 @@ class LocalNodeModel {
...
+    const fullDependencies = {
+      path: this.path,
+      ...(pageDependencies || {}),
+    }
+    const { path, connectionType } = fullDependencies
+    if (path) {
+      if (connectionType) {
+        this.createPageDependency({ path, connection: connectionType })
+      } else {
+        const nodes = Array.isArray(result) ? result : [result]
```
wardpeet: This should work too, don't know which one is more readable. (feel free to ignore)
```suggestion
        const nodes = [].concat(result);
```

wardpeet:
This looks good to me. Just to clarify `getAllNodes` is actually the function that fixes this issue right? Could you also explain a bit what this fixes? Can't seem to see an issue attached or a proper explanation why this is good 😛 Listening to fewer nodes is obviously a plus but want to know if there is more to it.

Issue #12802, opened 2019-03-24
gatsby-plugin-google-analytics should pre-connect to https://www.google-analytics.com


pgilad:
Running `lighthouse-4.2.0` on Gatsby blog, that includes `gatsby-plugin-google-analytics` will result in the following note:

![image](https://user-images.githubusercontent.com/4533329/54877296-30bd9c80-4e25-11e9-9c32-298855dda99e.png)

I ran `lighthouse` from the cli, with Gatsby info:

```bash
% npx gatsby info

  System:
    OS: macOS 10.14.3
    CPU: (4) x64 Intel(R) Core(TM) i7-6567U CPU @ 3.30GHz
    Shell: 3.0.2 - /usr/local/bin/fish
  Binaries:
    Node: 11.11.0 - ~/.nvm/versions/node/v11.11.0/bin/node
    npm: 6.9.0 - ~/.nvm/versions/node/v11.11.0/bin/npm
  Languages:
    Python: 2.7.16 - /usr/local/bin/python
  Browsers:
    Chrome: 73.0.3683.86
    Firefox: 66.0
    Safari: 12.0.3
  npmPackages:
    gatsby: ^2.2.8 => 2.2.8
    gatsby-cli: ^2.4.17 => 2.4.17
    gatsby-image: ^2.0.34 => 2.0.34
    gatsby-plugin-canonical-urls: ^2.0.12 => 2.0.12
    gatsby-plugin-catch-links: ^2.0.13 => 2.0.13
    gatsby-plugin-feed: ^2.1.0 => 2.1.0
    gatsby-plugin-google-analytics: ^2.0.17 => 2.0.17
    gatsby-plugin-manifest: ^2.0.24 => 2.0.24
    gatsby-plugin-netlify: ^2.0.13 => 2.0.13
    gatsby-plugin-offline: ^2.0.25 => 2.0.25
    gatsby-plugin-react-helmet: ^3.0.10 => 3.0.10
    gatsby-plugin-robots-txt: ^1.4.0 => 1.4.0
    gatsby-plugin-sharp: ^2.0.30 => 2.0.30
    gatsby-plugin-sitemap: ^2.0.10 => 2.0.10
    gatsby-plugin-typescript: ^2.0.11 => 2.0.11
    gatsby-remark-autolink-headers: ^2.0.16 => 2.0.16
    gatsby-remark-copy-linked-files: ^2.0.11 => 2.0.11
[...]

PR #12826, opened 2019-03-25
feat(gatsby-plugin-google-analytics): Add preconnect and dns-prefetch

pgilad:
Added `preconnect` and `dns-prefetch` link headers when using the `gatsby-plugin-google-analytics` plugin. It's recommended by the Lighthouse tool, also see the discussion on #12802 

I also didn't like that all code had to be indented for the `process.env.NODE_ENV` check, so I changed it to early return (Sorry about that 😸 )


yogeshkotadiya:
Thanks for this PR :+1:
But can we please add plugin option to `preconnect` GA.
And also adding some test would be nice :)

packages/gatsby-plugin-google-analytics/src/gatsby-ssr.js
```diff
@@ -18,91 +18,105 @@ exports.onRenderBody = (
...
-      })
-    }
+  if (process.env.NODE_ENV !== `production`) {
+    return null
+  }
+
+  // Lighthouse recommends pre-connecting to google analytics
+  setHeadComponents([
+    <link
+        rel="preconnect"
```
wardpeet:
```suggestion
        rel="dns-prefetch preconnect"
```

packages/gatsby-plugin-google-analytics/src/gatsby-ssr.js
```diff
@@ -18,91 +18,105 @@ exports.onRenderBody = (
...
+        key="preconnect-google-analytics"
+        href="https://www.google-analytics.com"
+    />,
+    <link
```
wardpeet: Please remove this option as this is negated in above comment

wardpeet:
> Thanks for this PR 👍
> But can we please add plugin option to `preconnect` GA.

Honestly I don't think this is needed, preconnect has hardly any downside and is done when the browser is idle and has time to do it. Most connections are kept in cache which means that preconnect is probably already done.
@yogeshkotadiya if you have other experiences feel free to add more context why you would like to make this optional.

@pgilad could you check your formatting rules as lots has changed which shouldn't be changed (see linting errors)

pgilad:
Thanks @wardpeet - no auto formatting 😢 ?

sidharthachatterjee:
@pgilad Linting issues fixed thanks to @pieh

packages/gatsby-plugin-google-analytics/src/gatsby-ssr.js
pgilad: Is it supported to specify multiple resource hints?

wardpeet: yeah it's actually best practice to have it this way so the browser will use preconnect if it supports it (all modern browsers) but will fallback to dns-prefetch if it doesnt (ie11)

pgilad: Couldn't find any documentation on that, but I trust you 😜

packages/gatsby-plugin-google-analytics/src/gatsby-ssr.js
```diff
@@ -18,31 +18,43 @@ exports.onRenderBody = (
...
 
-    const gaCreateOptions = {}
-    for (const option in knownOptions) {
-      if (typeof pluginOptions[option] === knownOptions[option]) {
-        gaCreateOptions[option] = pluginOptions[option]
-      }
+  // Lighthouse recommends pre-connecting to google analytics
+  setHeadComponents([
+    <link
+      rel="dns-prefetch preconnect"
```
yogeshkotadiya:
```suggestion
      rel="preconnect dns-prefetch"
```
Put `preconnect` first otherwise, the browser will always use `dns-prefetch` Order matters.

pgilad: Thanks for noticing this!

wardpeet: oh yeah sorry my bad 😄 @yogeshkotadiya thanks for having another pair of eyes

yogeshkotadiya:
> Honestly I don't think this is needed, preconnect has hardly any downside and is done when the browser is idle and has time to do it. Most connections are kept in cache which means that preconnect is probably already done.

@wardpeet 	That's `prefetch` not `preconnect`

`preconnect` initiates early in the browser and since some browsers have limited `preconnect` threads e.g. Chrome can request up to 6 DNS request simultaneously.



Pre-connecting to noncritical resource host can then block other DNS requests.
Also, the browser does not cache `preconnect` to browser cache instead it's stored into `HTTP cache` which expires in around ~2 Min.

wardpeet:
@yogeshkotadiya it depends on how you look at it, the browser keeps the preconnect connection ready for 10s but there is more to it than that. The operating system is caching the dns. Also when using tls 1.3 the handshake will do 1 roundtrip instead of 2 which is also something the operating system will provide. Chrome dedicates 8 threads to dns resolution if not mistaken.

I agree that preconnect shouldn't be thought of lightly but for a resource like analytics it's probably not bad as it mostly already is resolved and will be used in 10s. If you feel really strongly about this feel free to create a PR to disable it but it should be enabled by default.

yogeshkotadiya:
@wardpeet This makes sense, you made things very clear and I learned something also so thanks :+1:
So go ahead and merge it.

PR #12830, opened 2019-03-25
Fix nested inferrence

freiksenet:
## Description

Fixes issue with wordpress described in [here](https://github.com/gatsbyjs/gatsby/issues/12272#issuecomment-476154034). Still can't figure out a reproduction.


freiksenet:
Disregard that.

Issue #9770, opened 2018-11-07
Why does Gatsby unregister all service workers in development?


OrKoN:
I have realized that after spending some hours debugging a not working service worker used for push notifications. This service worker is a feature of the website, and it's not just for caching/offline mode. So I was wondering why does Gatsby unregister all service workers in development? In my opinion, it should only unregister service workers it has installed itself. Otherwise, the behavior is quite harmful, and it's breaking service worker implementations not related to the Gatsby itself making it hard to debug and develop service workers.

Related code: https://github.com/gatsbyjs/gatsby/blob/2605aa0ec74a79fa56bc0eae8d101356c25c38de/packages/gatsby/cache-dir/app.js#L37

btk:
I think you are right, gatsby should tag the service workers that registered by itself or its plugins and unregister these ones on the development, leaving developer the chance to debug their registered service workers.

But keep in mind the development version is not a static output, a javascript application that mimics the output. So it would be tricky to debug service workers on that environment. 

I'm not really educated on how that would go on. I think you should work on a PR if you have time. (I would start with tagging the sw `gatsby-plugin-offline` is registering)

OrKoN:
@btk the interesting part is that I am not using `gatsby-plugin-offline` but it seems that the code to unregister service workers belongs to Gatbsy project itself. Are there any other service workers installed by GatsbyJS? is it possible to move the unregister code to `gatsby-plugin-offline`?

btk:
@OrKoN Afaik there is not, but let's tag inkteam to see if unregister service worker should be moved to the offline plugin.

KyleAMathews:
Hmm yeah — we definitely shouldn't unregister all service workers, just ones from gatsby-plugin-offline. Or perhaps make it configurable.

The reason the code was added is because people will frequently have SWs hanging around which can be super confusing as it blocks your development environment with no clear way to fix it.

prajapati-parth:
This issue makes working on web notifications difficult. Does the same issue exist in Gatsby v2?

KyleAMathews:
@prajapati-parth it does. Would you like to create a PR removing the code that does the unregistration?

prajapati-parth:
Sure. I removed some code from the file mentioned by the author of this issue and seems like many tests are failing. I tried setting up in development by cloning and doing `yarn`. For testing I did `npm test`. Did I miss anything or do we have a getting started for development link?

wardpeet:
[...]

PR #12834, opened 2019-03-25
Removed unregistering of service workers on reload

prajapati-parth:
## Description

Fixed issue https://github.com/gatsbyjs/gatsby/issues/9770

### DIFF
--- a/packages/gatsby/cache-dir/app.js
+++ b/packages/gatsby/cache-dir/app.js
@@ -22,22 +22,6 @@ apiRunnerAsync(`onClientEntry`).then(() => {
     })
   }
 
-  /**
-   * Service Workers are persistent by nature. They stick around,
-   * serving a cached version of the site if they aren't removed.
-   * This is especially frustrating when you need to test the
-   * production build on your local machine.
-   *
-   * Let's unregister the service workers in development, and tidy up a few errors.
-   */
-  if (supportsServiceWorkers(location, navigator)) {
-    navigator.serviceWorker.getRegistrations().then(registrations => {
-      for (let registration of registrations) {
-        registration.unregister()
-      }
-    })
-  }
-
   const rootElement = document.getElementById(`___gatsby`)
 
   const renderer = apiRunner(
@@ -60,10 +44,3 @@ apiRunnerAsync(`onClientEntry`).then(() => {
     })
   })
 })
-
-function supportsServiceWorkers(location, navigator) {
-  if (location.hostname === `localhost` || location.protocol === `https:`) {
-    return `serviceWorker` in navigator
-  }
-  return false
-}


davidbailey00:
> Let's see if we can make this only remove gatsby-plugin-offline instances?

I think in this case it _does_ make sense to remove all service workers.

Firstly, SWs are port-specific, so for example if you're testing a production build on port 9000, the worker won't be removed when testing the development version on port 8000. Therefore it shouldn't interfere with SWs from other HTTP servers as long as they're on different ports.

Secondly, I think if anything, it's more important to remove _non-Gatsby_ service workers, rather than vice versa. Otherwise there would be random things going on outside of our control if another SW is installed from a different non-Gatsby site on the same port.

prajapati-parth:
> Secondly, I think if anything, it's more important to remove non-Gatsby service workers, rather than vice versa.

Isn't the main purpose of this PR to not remove non-Gatsby service workers? I tried developing firebase cloud messaging on a gatsby site and every time I reload the page in development mode, it de-registers the service worker.

davidbailey00:
> Isn't the main purpose of this PR to not remove non-Gatsby service workers? I tried developing firebase cloud messaging on a gatsby site and every time I reload the page in development mode, it de-registers the service worker.

Ah, I understand the reasoning now. In this case I think it does make sense to remove this code and possibly display a warning instead - we can't really tell which SWs a user wants to keep or not so it's best not to guess, but at least if we display a warning in the console then it'll save people time when they're confused about why a site isn't updating on refresh.

@DSchau thoughts on this?

DSchau:
> Isn't the main purpose of this PR to not remove non-Gatsby service workers? I tried developing firebase cloud messaging on a gatsby site and every time I reload the page in development mode, it de-registers the service worker.

This seems like a reasonable use case--to me. I don't think we should remove "user code" that we're not in control of.

> Ah, I understand the reasoning now. In this case I think it does make sense to remove this code and possibly display a warning instead - we can't really tell which SWs a user wants to keep or not so it's best not to guess, but at least if we display a warning in the console then it'll save people time when they're confused about why a site isn't updating on refresh.

Warning/prompt is _OK_ but some outstanding questions are helpful:

1. Is it reasonable to remove the **Gatsby** service worker in development?
    - I think the answer to this is (apparently and historically) yes.
1. Is it reasonable to remove _all_ service workers?
    - I think the answer to this is probably no--especially if it's breaking/preventing a development workflow for a common tool like Firebase.
1. Is there a clean way to remove a Gatsby specific service worker?
    - e.g. filename is _decent_ (sw.js) but some way to uniquely identify would be even better

That being said--the solution should not be to entirely prevent removal of _any_ service workers here (as in this PR)--we should target the Gatsby service worker for removal and leave others alone.

prajapati-parth:
> Is there a clean way to remove a Gatsby specific service worker?
e.g. filename is decent (sw.js) but some way to uniquely identify would be even better

Correct. Because most of the services that ask users to create service workers with a specific name also allows them to configure the use of an existing service worker with different name. Eg. Firebase cloud messaging needs a SW with name firebase-messaging-sw.js but it can also be configured with name sw.js. So we'll definitely need other way for unique identification.

davidbailey00:
I've created a PR at #13716 in favour of this one, which warns about SWs detected but doesn't remove them. In development, we shouldn't have any Gatsby SWs since gatsby-plugin-offline only creates a SW in production mode, so I think detecting the "creator" of a SW might be unnecessary.

As a result I'll be closing this PR, but thanks anyway your contribution! Feel free to open an issue or another PR in the future if you feel anything else is wrong and we'll be sure to have a look :)

PR #12837, opened 2019-03-25
fix(schema): Remove accidental capitalization of lowercase typenames.

freiksenet:
## Description

We accidentally uppercased typenamse when making inline types, causing very weird errors. This reverts to old behaviour.


sidharthachatterjee:
Published in gatsby@2.2.13

PR #12841, opened 2019-03-25
chore(docs): add "--save" flag to script

stufinn:
## Description

Adds a "--save" flag to script for installing modules.

New devs may not remember to include the --save flag when writing the script, or it may remain left out if copied/pasted directly from the docs.

## Related Issues

Couldn't find an existing issue for this.

### DIFF
--- a/docs/docs/using-gatsby-image.md
+++ b/docs/docs/using-gatsby-image.md
@@ -40,7 +40,7 @@ With Gatsby, we can make images way way better.
 1. Install `gatsby-image` and other, necessary dependencies like `gatsby-plugin-sharp` and `gatsby-transformer-sharp`
 
 ```shell
-  npm i --save gatsby-image gatsby-transformer-sharp gatsby-plugin-sharp
+  npm install --save gatsby-image gatsby-transformer-sharp gatsby-plugin-sharp
 ```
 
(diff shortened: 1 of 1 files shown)


stufinn:
@DSchau I wasn't aware of that default behaviour; good to know. The npm install scripts in the official Gatsby tutorial include the --save flag, so I'd argue that for consistency's sake it's worth including. Unless of course the tutorial docs should be updated to reflect that.

docs/docs/using-gatsby-image.md
```diff
@@ -40,7 +40,7 @@ With Gatsby, we can make images way way better.
 1. Install `gatsby-image` and other, necessary dependencies like `gatsby-plugin-sharp` and `gatsby-transformer-sharp`
 
 ```shell
-  npm i gatsby-image gatsby-transformer-sharp gatsby-plugin-sharp
+  npm i --save gatsby-image gatsby-transformer-sharp gatsby-plugin-sharp
```
DSchau:
```suggestion
  npm install --save gatsby-image gatsby-transformer-sharp gatsby-plugin-sharp
```

Sure - it's more explicit, which I suppose is better. If we can make this change we can get this merged!

stufinn:
Thanks @DSchau!

PR #12851, opened 2019-03-26
Use merge=union for yaml files so we have less merge conflicts

KyleAMathews:

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,8 @@
 * text=auto eol=lf
 
+/docs/sites.yml merge=union
+/docs/starters.yml merge=union
+
 # These files are binary and should be left untouched
 *.png binary
 *.jpg binary


.gitattributes
```diff
@@ -1,5 +1,8 @@
 * text=auto eol=lf
 
+/docs/sites.yml merge=union
```
DSchau: Would we ever have yaml files we wouldn't want this behavior to apply to?

pieh: IMO sidebar links should cause merge conflicts that need to be manually resolved, so 👍 for selective merge strategy just for showcase entries

KyleAMathews: Yeah, I was thinking more explicit is better until we play with things a bit. And yeah, good point on sidebar links merge conflicts potentially being problematic.

PR #12852, opened 2019-03-26
fix(gatsby-plugin-sitemap): include path prefix

bennetthardwick:
## Description

The Gatsby sitemap plugin can be configured to split up all of the sites URLs into smaller sitemaps, all being referenced from a main "index" sitemap. Unfortunately, this index sitemap doesn't take the `pathPrefix` option into account and will never link to the correct "child" sitemaps.

As `sitemap.js` doesn't have an option for `pathPrefix`, this change appends the `pathPrefix` to the site's `hostname` when more than one sitemap needs to be created.


bennetthardwick:
@gatsbyjs/core is there any chance someone can take a look at this?

packages/gatsby-plugin-sitemap/src/__tests__/gatsby-node.js
```diff
@@ -176,5 +176,36 @@ describe(`Test plugin sitemap`, async () => {
...
+        `application/xml`
+      )
+      const elements = DOM.querySelectorAll(`sitemap loc`)
+      expect(elements.length).toBe(2)
+      expect(elements[0].innerHTML).toBe(
+        queryResult.data.site.siteMetadata.siteUrl + prefix + `/sitemap-0.xml`
+      )
+      expect(elements[1].innerHTML).toBe(
+        queryResult.data.site.siteMetadata.siteUrl + prefix + `/sitemap-1.xml`
+      )
```
pieh: perhaps we should just test if `createSitemapIndex` is called with `hostname` that includes pathPrefix? This mocks some internals of `sitemap` package that can change, making this test not that maintainable

bennetthardwick: Thanks @pieh, I've updated the test to reflect your feedback.

packages/gatsby-plugin-sitemap/src/__tests__/gatsby-node.js
```diff
@@ -176,5 +177,16 @@ describe(`Test plugin sitemap`, async () => {
...
       expect(contents).toMatchSnapshot()
     })
+    it(`should include path prefix when creating creating index sitemap`, async () => {
+      const sitemapSpy = jest.spyOn(sitemap, `createSitemapIndex`)
+      const options = {
+        sitemapSize: 1,
+      }
+      const prefix = `/test`
+      await onPostBuild({ graphql, pathPrefix: prefix }, options)
+      expect(sitemapSpy.mock.calls[0][0].hostname).toEqual(
```
pieh: NIT: don't need to change this (we do use it in different places as well), but for future, please use
```js
expect(sitemapSpy).toBeCalledWith(
  expect.objectContaining({
    hostname: `${queryResult.data.site.siteMetadata.siteUrl}${prefix}`
  })
)
```
reaching to `.mock.calls[0][0]` works, but is less readable

PR #12855, opened 2019-03-26
Simplify part-zero of tutorial significantly.

baryluk:
There is A LOT of noise in this tutorial. It is off putting.

Remove almost all of it, and make it really really simple. No need to treat people like idiots.

### DIFF
--- a/docs/tutorial/part-zero/index.md
+++ b/docs/tutorial/part-zero/index.md
@@ -3,75 +3,38 @@ title: Set Up Your Development Environment
 typora-copy-images-to: ./
 ---
 
-Before you start building your first Gatsby site, you’ll need to familiarize yourself with some core web technologies and make sure that you have installed all required software tools.
+Before you start building your first Gatsby site, you’ll need to make sure that you have installed all required software tools.
 
 ## Overview of core technologies
 
-It’s not necessary to be an expert with these already — if you’re not, don’t worry! You’ll pick up a lot through the course of this tutorial series. These are some of the main web technologies you’ll use when building a Gatsby site:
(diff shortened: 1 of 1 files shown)


LekoArts:
Hi,

thank you for the PR! No need to call beginners idiots, though. Everyone started at zero and things like installing Git or Node.js can be difficult! We want our tutorial to be accessible to everyone.

That's why we also wrote this into the introduction:

> The full tutorial is intended to be as accessible as possible to people without much web development experience (yet!) — no need to be an expert. If you prefer to jump straight to code, feel free to skip the step-by-step tutorial and see the quick start page.

The linked [quickstart](https://www.gatsbyjs.org/docs/quick-start) is essentially what you proposed to do here. So you certainly have the option to follow the quickstart and begin with **Part one** of the tutorial afterwards 😊 	

As such--we're going to close this out. We want you to know that we certainly want to keep you as a contributor to Gatsby, so we'd encourage you to check out our [open issues][open-issues].

baryluk:
https://www.gatsbyjs.org/docs/quick-start , indeed makes more sense. I didn't notice it (I was just following the first point in docs). I think what could be done better:

Make Get Started on https://www.gatsbyjs.org/ link directly to quick start, not to "Docs". The Docs link at the top of main page is already serving this purpose.

The disclaimer on top of the quick start will already handle the case of people needing more info.

LekoArts:
Thank you for the suggestion! However, I think the docs overview already does a great job at pointing you into the right direction:

> Tutorial: The tutorial is written to be as accessible as possible to people without much web development experience.

> Quick start: The quick start is intended for intermediate to advanced developers who prefer to dig straight in.

Or what is unclear here? 😊 	

> Make Get Started on gatsbyjs.org link directly to quick start, not to "Docs". The Docs link at the top of main page is already serving this purpose.

It's common practice to place a CTA (Call to Action) button to something on the page that is (of course) also in the navigation (so that you can access it from anywhere). Due to the reasons above we chose to link that page.

PR #12859, opened 2019-03-26
fix(gatsby-source-wordpress): fix WP-API-MENUS when reported endpoint url doesn't match base url

STUkh:
## Description

This PR fixes case when wordpress site URL exists behind proxy or in docker-container with it\'s local own address and route.self links give us wrong URL that gatsby-source-wordpress try to fetch (and obviously fails).

## Related Issues
For me it's fixes intercontainer communication, when wordpress hosted inside of docker with internal address http://localhost:8080, but for exposes as http://wordpress:8080 for other containers. Base URL fetched properly by gatsy-source-wordpress, but all internal self links are broken because rely on `localhost` instead of desired `wordpress` base host.

I see it should fixe the same problem described in #10427


pieh:
Are you able to test changes from https://github.com/gatsbyjs/gatsby/pull/10624 ? It touches similar problem and it _might_ fix this issue as well.

pieh:
@STUkh Can you try using `gatsby-source-wordpress@3.0.50`? I merged PR linked above and I do believe it should fix this

STUkh:
@pieh, thanks for answer, will try.

STUkh:
@pieh , now It works better, but fails in menus fetching
https://monosnap.com/file/UdFdxrO3KdFqczxTmRSIsguvCeQqZk

pieh:
@STUkh We are getting close ;)
https://github.com/gatsbyjs/gatsby/blob/632f0356d2c75f91dfbd7cb6853499c17050f074/packages/gatsby-source-wordpress/src/fetch.js#L294-L308

Seems like this references `links.self` so we need to swap just this one. Are you able to share minimal docker-compose (or whatever) setup you are using so I could reproduce it?

STUkh:
@pieh updated PR content because of new PRs was merged. Fixed menus paths. Docker example project to reproduce bug: https://github.com/STUkh/gatsby-wordpress-docker
Not it uses local fixed version of gatsby-plugin-wordpress, so it will be easy to see changes.

PR #12860, opened 2019-03-26
chore(showcase): add Intranet Italia Day - ariadnedigital.it

ariadne-github:
## Description

Add Intranet Italia Day to showcase

### DIFF
--- a/docs/sites.yml
+++ b/docs/sites.yml
@@ -5125,7 +5125,7 @@
   built_by_url: https://twitter.com/Pattyrn4
   featured: false
 - title: Intranet Italia Day
-  main_url: https://www.ariadnedigital.it/en
+  main_url: https://www.intranetitaliaday.it/en
   url: https://www.intranetitaliaday.it/en
   description: >
     The Italian event dedicated to the digital workplace that focuses on planning, governance and company intranet management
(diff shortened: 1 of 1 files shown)


docs/sites.yml
```diff
@@ -5124,3 +5124,15 @@
...
+  url: https://www.intranetitaliaday.it/en
+  description: >
+    The Italian event dedicated to the digital workplace that focuses on planning, governance and company intranet management
+  categories:
+    - Event
+    - Intranet
+    - Conference
+  built_by: Ariadne Digital
+  built_by_url: https://www.ariadnedigital.it
+  featured: false
```
yogeshkotadiya:
```suggestion
  featured: false
```
This might fix linting error.

docs/sites.yml
```diff
@@ -5124,3 +5124,15 @@
   built_by: Pattyrn
   built_by_url: https://twitter.com/Pattyrn4
   featured: false
+- title: Intranet Italia Day
+  main_url: https://www.ariadnedigital.it/en
+  url: https://www.intranetitaliaday.it/en
```
yogeshkotadiya: Which site you want to add,

`https://www.ariadnedigital.it/en`
or
`https://www.intranetitaliaday.it/en`

Because both `main_url` and `url` should have same domain.

ariadne-github: The correct url is https://www.intranetitaliaday.it/en, my mistake. Thanks @yogeshkotadiya, also for spotting the subtle formatting error.

PR #12875, opened 2019-03-26
fix(gatsby-telemetry): don't inherit args from main process

stefanprobst:
Don't inherit args from main process. This is e.g. an issue when starting node with `--inspect-brk`, which would leave the forked process sitting on port 9229.


DSchau:
Thanks @stefanprobst!

```
Successfully published:
 - gatsby-cli@2.5.2
 - gatsby-telemetry@1.0.2
 - gatsby@2.3.2

```

Issue #12398, opened 2019-03-08
[gatsby-transformer-remark] Extra space when link is in excerpt


kylejrp:
## Description

When a markdown file includes a link that would be shown in the plaintext excerpt, an extra space is added after the link.

### Steps to reproduce

1. Clone https://github.com/gatsbyjs/gatsby-starter-blog
2. Modify `content/blog/hello-world/index.md` so that the first paragraph has a link:
  `This is my first post on my new [fake blog](http://example.com)! How exciting!`
3. Start the site with `gatsby develop` and read the excerpt on the Hello World blog post

### Expected result

The excerpt on the page should say:
> This is my first post on my new fake blog!

### Actual result

The excerpt on the page includes a space before the exclamation point:
> This is my first post on my new fake blog !

### Environment
`gatsby info --clipboard` threw an error when I tried to run it, but here's what it spit out:

```
   System:
     OS: Windows 10
     CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
   Binaries:
     npm: 6.4.1 - C:\Program Files\nodejs\npm.CMD
   Browsers:
     Edge: 44.17763.1.0
   npmPackages:
     gatsby: ^2.1.24 => 2.1.24
     gatsby-image: ^2.0.31 => 2.0.31
     gatsby-plugin-feed: ^2.0.14 => 2.0.14
     gatsby-plugin-google-analytics: ^2.0.16 => 2.0.16
     gatsby-plugin-manifest: ^2.0.22 => 2.0.22
     gatsby-plugin-offline: ^2.0.24 => 2.0.24
     gatsby-plugin-react-helmet: ^3.0.8 => 3.0.8
     gatsby-plugin-sharp: ^2.0.25 => 2.0.25
     gatsby-plugin-typography: ^2.2.8 => 2.2.8
     gatsby-remark-copy-linked-files: ^2.0.9 => 2.0.9
     gatsby-remark-images: ^2.0.6 => 2.0.6
     gatsby-remark-prismjs: ^3.2.4 => 3.2.4
     gatsby-remark-responsive-iframe: ^2.0.9 => 2.0.9
     gatsby-remark-smartypants: ^2.0.8 => 2.0.8
     gatsby-source-filesystem: ^2.0.23 => 2.0.23
     gatsby-transformer-remark: ^2.3.1 => 2.3.1
     gatsby-transformer-sharp: ^2.1.15 => 2.1.15

```

[...]

PR #12878, opened 2019-03-26
fix(gatsby-transformer-remark): fix excerpt generation - strip excessive white spaces, extract alt from images

macklinu:
## Description

This is a failing test case to reproduce the issue stated in https://github.com/gatsbyjs/gatsby/issues/12398. Here is a screenshot of test failure output:

image

## Related Issues

Relates to #12398


wardpeet:
@macklinu mind fixing this as well? Thanks for creating the test, if you don't really know how we can take over from here.

macklinu:
@wardpeet I'll take a look at patching this and ping you if I come up with a solution or need to ask for another contributor's help. Thank you!

macklinu:
Would it make sense to use [`strip-markdown`](https://github.com/remarkjs/strip-markdown) as a remark plugin instead of writing our own markdown processor in order to parse out the excerpt content? Or maybe we don't need all the functionality of `strip-markdown` and just need to account for a few more node types? Any implementation suggestions?

https://github.com/gatsbyjs/gatsby/blob/15ef1783620154abefdd5eac9e8d5145ded68d09/packages/gatsby-transformer-remark/src/extend-node-type.js#L412-L449

wardpeet:
strip-markdown strips markdown away, we still want to show the link or bold text right? That plugin just makes the text unstyled.

macklinu:
Ah, I see. I think my failing test assertion is incorrect then. We don't want to necessarily strip markdown, but we just want to preserve markdown as is and not remove any content?

wardpeet:
yes, the issue is that an extra space is added between the link & the ! symbol. It might be something that we have to fix upstream.

frinyvonnick:
Hi @macklinu 👋, 

First thanks for your failing test it really helped me diving in this bug 👏. 

I found out that the function `getExcerpt` in [extend-node-type.js](https://github.com/gatsbyjs/gatsby/pull/12878/commits/f6a6e33388adba13d8612496011e8891e865b146#diff-488d99b036db541ee4a25a6991027b5dR412) wasn't handling `image` and the `link` was classified as an `image`. That was causing the missing parts. The second problem of extra spaces was caused by a join of excerpt parts with a space. 

Sorry to not asking before pushing on your branch 😊

packages/gatsby-transformer-remark/src/__tests__/extend-node.js
```diff
@@ -376,6 +376,27 @@ Where oh [*where*](nick.com) **_is_** ![that pony](pony.png)?`,
...
+
+Where oh [*where*](nick.com) **_is_** ![that pony](pony.png)?`,
+    `excerpt
+      excerptAst
+      frontmatter {
+          title
+      }
+      `,
+    node => {
+      expect(node).toMatchSnapshot()
```
pieh: Do we need this assertion?

packages/gatsby-transformer-remark/src/__tests__/extend-node.js
```diff
@@ -376,6 +376,27 @@ Where oh [*where*](nick.com) **_is_** ![that pony](pony.png)?`,
...
+  bootstrapTest(
+    `excerpt does have missing words and extra spaces`,
+    `---
+title: "my little pony"
+date: "2017-09-18T23:19:51.246Z"
+---
+
+Where oh [*where*](nick.com) **_is_** ![that pony](pony.png)?`,
+    `excerpt
+      excerptAst
```
pieh: I would say this is not needed (if we remove snapshot for this test)

frinyvonnick:
Thanks for your review @pieh 👍

wardpeet:
Sweet! thanks for getting this PR through the door @frinyvonnick! I'll fix the unit test.

frinyvonnick:
Thanks @wardpeet 🙏 !

wardpeet:
pretty sure we can ignore danger here.

sidharthachatterjee:
Published in `gatsby-transformer-remark@2.3.12`

PR #12883, opened 2019-03-26
fix(themes): Handle shared parent theme with component shadowing

johno:
If `gatsby-theme-a` and `gatsby-theme-b` both install and use
`gatsby-theme-c` (as a parent theme) this would cause an error
to be raised when either sibling theme shadowed a component
from `gatsby-theme-c` when it should be ignored.

### DIFF
--- a/packages/gatsby/src/internal-plugins/webpack-theme-component-shadowing/index.js
+++ b/packages/gatsby/src/internal-plugins/webpack-theme-component-shadowing/index.js
@@ -1,6 +1,7 @@
 const path = require(`path`)
 const debug = require(`debug`)(`gatsby:component-shadowing`)
 const fs = require(`fs`)
+const _ = require(`lodash`)
 
 module.exports = class GatsbyThemeComponentShadowingResolverPlugin {
   cache = {}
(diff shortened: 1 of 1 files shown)


packages/gatsby/src/internal-plugins/webpack-theme-component-shadowing/index.js
```diff
@@ -14,10 +15,16 @@ module.exports = class GatsbyThemeComponentShadowingResolverPlugin {
...
       // find out which theme's src/components dir we're requiring from
-      const matchingThemes = this.themes.filter(name =>
+      const allMatchingThemes = this.themes.filter(name =>
         request.path.includes(path.join(name, `src`))
       )
-      // 0 matching themes happens a lot fo rpaths we don't want to handle
+
+      // Two sibling themes can each have the same parent theme, when
+      // this occurs it shouldn't cause an error because when shadowing
+      // because they'll typically resolve to the same module.
```
ChristopherBiscardi: This is a more general problem than the comment suggests, how about:

> The same theme can be included twice in the themes list causing multiple matches. This case should only be counted as a single match for that theme.

PR #12888, opened 2019-03-26
fix(gatsby-telemetry): Ensure quickly running commands exit freely

jamo:
It turns out the events timer caused `gatsby clean` in particular to never exit - unlike other fairly quickly running commands such as help, telemetry and build...

This should fix it

### DIFF
--- a/packages/gatsby-cli/src/create-cli.js
+++ b/packages/gatsby-cli/src/create-cli.js
@@ -10,7 +10,6 @@ const {
   setDefaultTags,
   setTelemetryEnabled,
 } = require(`gatsby-telemetry`)
-console.log(require.resolve(`gatsby-telemetry`))
 
 const handlerP = fn => (...args) => {
   Promise.resolve(fn(...args)).then(
--- a/packages/gatsby-cli/src/create-cli.js
+++ b/packages/gatsby-cli/src/create-cli.js
@@ -10,6 +10,7 @@ const {
   setDefaultTags,
   setTelemetryEnabled,
 } = require(`gatsby-telemetry`)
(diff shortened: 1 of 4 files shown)


packages/gatsby-cli/src/create-cli.js
```diff
@@ -10,6 +10,7 @@ const {
   setDefaultTags,
   setTelemetryEnabled,
 } = require(`gatsby-telemetry`)
+console.log(require.resolve(`gatsby-telemetry`))
```
pieh:
```suggestion
```

packages/gatsby/src/commands/develop.js
```diff
@@ -258,6 +258,7 @@ async function startServer(program) {
 module.exports = async (program: any) => {
   initTracer(program.openTracingConfigFile)
   telemetry.trackCli(`DEVELOP_START`)
+  telemetry.startBackgroudUptate()
```
pieh: is this also needed (or useful) in `gatsby serve` which is second non-self-exiting gatsby command?

packages/gatsby-cli/src/create-cli.js
jamo: So this ended up sneaking in after all, thanks for spotting it

packages/gatsby/src/commands/develop.js
jamo: Indeed, we should add it for that too. 
My Mac is just installing updates so I'll update this tomorrow morning. Or if we want to ship this before it, feel free to edit the pr :)

sidharthachatterjee: Done, shipping now

PR #12898, opened 2019-03-27
feat(gatsby-plugin-less): allow passing less plugins

### LINKED ISSUE #4645: Pass LESS plugins to gatsby-plugin-less
I want to add a LESS specific plugin (https://github.com/seven-phases-max/less-plugin-lists) to gatsby-plugin-less.  

I've installed via NPM and in my gatsby-config.js
```
const LessPluginLists = require('less-plugin-lists')
module.exports = {
  plugins: [
    {
      resolve: `gatsby-plugin-less`,
      options: {
        plugins: [new LessPluginLists()],
      },
    },
}
```

however I get this error

```
error UNHANDLED REJECTION


  Error: Unable to find plugin "undefined"

  - load.js:89 resolvePlugin
    /[gatsby]/dist/bootstrap/load-plugins/load.js:89:11

  - load.js:138 processPlugin
    /[gatsby]/dist/bootstrap/load-plugins/load.js:138:27

  - load.js:121
    /[gatsby]/dist/bootstrap/load-plugins/load.js:121:35

  - Array.forEach

  - load.js:120 processPlugin
    /[gatsby]/dist/bootstrap/load-plugins/load.js:120:40

  - load.js:159
    /[gatsby]/dist/bootstrap/load-plugins/load.js:159:28

  - Array.forEach

  - load.js:158 _callee$
    /[gatsby]/dist/bootstrap/load-plugins/load.js:158:28

  - index.js:61 _callee$
    /[gatsby]/dist/bootstrap/load-plugins/index.js:61:18
```

Is there a way to pass plugins to the LESS compiler ? I've tried other LESS specific plugins in the same manner but I get the same error.

slorber:
Currently we can't declare less loader plugins because plugins can be nested one inside each others, according to https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/bootstrap/load-plugins/load.js#L134

So forwarding the "lessPlugins" attribute to the plugins loader option is a workaround to avoid this problem.

See original issue https://github.com/gatsbyjs/gatsby/issues/4645

But I think Gatsby should rather generalize find a more robust and generalized solution to this conflict between "gatsby plugins with gatsby subplugins" vs "gatsby plugin with plugin taking plugins as params"


packages/gatsby-plugin-less/src/gatsby-node.js
```diff
@@ -13,6 +13,7 @@ exports.onCreateWebpackConfig = (
     options: {
       sourceMap: !PRODUCTION,
       ...lessOptions,
```
slorber: Note it's important to put this AFTER `...lessOptions` because the lessOptions always contain a plugin attribute which would override the  `lessPlugins` attribute

pieh: This makes sense. Could you update plugin's README to show that to use less plugins, the `lessPlugins` option need to be used?

pieh:
I'll fix unit tests after README changes - they only break because there is empty array by default, which is not present in snapshots

slorber:
I'll do the changes next Monday when I come back from holiday. Thanks for
the review

slorber:
@pieh  I've added the documentation in readme

slorber:
@pieh don't forget me please this is a tiny change ;)

pieh:
Hey, it's not forgotten. I'm out of the office this week. Let's get back to
it next week.

pieh:
Thanks for PR (and reminder) @slorber!

wardpeet:
published in `gatsby-plugin-less@2.0.13` 🎉

PR #12899, opened 2019-03-27
fix(gatsby-telemetry): add the missing fs-extra dependency

arcanis:
## Description

The new `gatsby-telemetry` package [uses `fs-extra`](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-telemetry/src/store.js#L10) but doesn't list it in the `package.json` field.

You should use Plug'n'Play 😛

### DIFF
--- a/packages/gatsby-telemetry/package.json
+++ b/packages/gatsby-telemetry/package.json
@@ -13,6 +13,7 @@
     "ci-info": "2.0.0",
     "configstore": "4.0.0",
     "envinfo": "^5.8.1",
+    "fs-extra": "^7.0.1",
     "node-fetch": "2.3.0",
     "resolve-cwd": "^2.0.0",
     "source-map": "^0.5.7",


pieh:
@arcanis Are you aware of any automated tool so we could validate in our CI checks on PR that every used package is declared in dependencies? That would help a lot with whack-a-mole like this

arcanis:
> Are you aware of any automated tool so we could validate in our CI checks on PR that every used package is declared in dependencies?

Yarn v2 is super close from supporting Gatsby (only this PR and https://github.com/yarnpkg/berry/pull/44 are missing), so what I suggest is to add an integration step using the v2 to run `yarn dlx gatsby new gatsby-starter-default`. In this case it's how I found out the problem.

Generally speaking, my advice regarding PnP was semi-serious: I get that it's a heavy fix since there's sometimes a bit of work to port existing setups, but it would have prevented this (on top of the extra benefits it brings to the table like zero-installs).

PR #12900, opened 2019-03-27
chore(docs): fix broken gatsby style guide link

thiskevinwang:
# Description
This fixes a broken `Gatsby Style Guide` link

## In
The ["Adding Page Transitions with gatsby-plugin-page-transitions"](https://www.gatsbyjs.org/docs/adding-page-transitions-with-plugin-page-transitions/) page

`docs/docs/adding-page-transitions-with-plugin-page-transitions.md`

### DIFF
--- a/docs/docs/adding-page-transitions-with-plugin-page-transitions.md
+++ b/docs/docs/adding-page-transitions-with-plugin-page-transitions.md
@@ -8,5 +8,5 @@ This guide will cover the setup and the various ways you can utilize this plugin
 
 This is a stub. Help our community expand it.
 
-Please use the [Gatsby Style Guide](/docs/gatsby-style-guide/) to ensure your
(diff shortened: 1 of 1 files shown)


marcysutton:
Good find! However, there should be a redirect in place for that. @DSchau any idea why this isn't working? https://github.com/gatsbyjs/gatsby/blob/master/www/gatsby-node.js#L115

DSchau:
@marcysutton it is!

If you open in a new tab, then the Netlify redirect will take over, e.g. https://gatsbyjs.org/docs/gatsby-style-guide/ will redirect like you expect.

However - agreed that this is confusing. I think we'll want to use `redirectInBrowser` as [documented in our Node API spec](https://www.gatsbyjs.org/docs/actions/#createRedirect)--probably for _all_ of our links.

PR #12902, opened 2019-03-27
chore(docs): Update programmatically-create-pages-from-data.md

ewalpole:
Changing from Promise/Resolve to Async.

## Description

Tried using the Promise/Resolve example in my code and it was throwing errors. Found your debugging async page and used the example there (very similar) and it worked. For consistency, moved that example to the other page as well. Was there a reason you were using Promise/Resolve in this space?

### DIFF
--- a/docs/docs/programmatically-create-pages-from-data.md
+++ b/docs/docs/programmatically-create-pages-from-data.md
@@ -41,8 +41,8 @@ exports.createPages = async function({ actions, graphql }) {
       const slug = edge.node.fields.slug
       actions.createPage({
         path: slug,
-        component: require.resolve(`./src/templates/post.js`),
-        context: { slug },
+        component: require.resolve(`./src/templates/blog-post.js`),
+        context: { slug: slug },
       })
     })
   })
--- a/docs/docs/programmatically-create-pages-from-data.md
+++ b/docs/docs/programmatically-create-pages-from-data.md
@@ -23,7 +23,6 @@ which is at the core of programmatically creating a page.
 
 ```js:title=gatsby-node.js
 exports.createPages = async function({ actions, graphql }) {
-  // highlight-start
   const { data } = await graphql(`
     allMarkdownRemark {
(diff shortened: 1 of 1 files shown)


docs/docs/programmatically-create-pages-from-data.md
```diff
@@ -21,35 +21,29 @@ access to the
...
-        createPage({
-          path: node.fields.slug,
-          component: path.resolve(`./src/templates/blog-post.js`),
-          context: {
-            slug: node.fields.slug,
-          },
-        })
+    }
+  `).then(res => {
+    res.data.allMarkdownRemark.edges.forEach(edge => {
```
DSchau: Perhaps move this after the await? It tends to clean up the code if we can move off of .then blocks!

e.g.

```
const { data } = await graphql(`
 # query
`)

data.allMarkdownRemark.edges.forEach(edge => {
```

ewalpole: @DSchau I made the changes and tested it, works for me. Added the highlighting as it was in the previous code sample.

docs/docs/programmatically-create-pages-from-data.md
```diff
@@ -21,37 +21,29 @@ access to the
...
-          context: {
-            slug: node.fields.slug,
-          },
-        })
-      })
-      // highlight-end
-      resolve()
+```js:title=gatsby-node.js
+exports.createPages = async function({ actions, graphql }) {
+  // highlight-start
```
DSchau: This should be moved down 👇 (where it was previously!)

ewalpole: @DSchau Now? :D Sorry this is my first docs fix. Still figuring it out.

DSchau: Please don't apologize!! We appreciate you taking the time--and this looks great now.

Thank you!

PR #12907, opened 2019-03-28
fix(gatsby-plugin-manifest): ensure icon_options is stripped

moonmeister:
## Description

Here are those bug fixes, and cleanup for the new `icon_options`. I also added tests.

## Related Issues

#12794


packages/gatsby-plugin-manifest/src/__tests__/gatsby-node.js
```diff
@@ -99,23 +105,22 @@ describe(`Test plugin manifest options`, () => {
     const icon = `pretend/this/exists.png`
     const size = 48
 
+    let pluginSpecificOptions = {
```
wardpeet: This can be a const right?
```suggestion
    const pluginSpecificOptions = {
```

packages/gatsby-plugin-manifest/src/__tests__/gatsby-node.js
```diff
@@ -126,23 +131,15 @@ describe(`Test plugin manifest options`, () => {
   it(`fails on non existing icon`, async () => {
     fs.statSync.mockReturnValueOnce({ isFile: () => false })
 
+    let pluginSpecificOptions = {
```
wardpeet: This can also be a const right?
```suggestion
    const pluginSpecificOptions = {
```

moonmeister:
@wardpeet I'll fix those other lets in my clean up pr

packages/gatsby-plugin-manifest/README.md
```diff
@@ -188,9 +188,11 @@ module.exports = {
 }
 ```
 
-## Custom icon options
+## Iterative icon options
```
DSchau: Calling this is an iterator is kinda weird, no?

I think I'd just re-word this slightly

```suggestion
## Custom icon options
```

packages/gatsby-plugin-manifest/src/__tests__/gatsby-node.js
```diff
@@ -159,6 +156,7 @@ describe(`Test plugin manifest options`, () => {
       plugins: [],
       theme_color_in_head: false,
       cache_busting_mode: `name`,
+      icon_options: {},
```
DSchau: This doesn't _need_ to be passed right? i.e. it's tolerant if this isn't passed.

packages/gatsby-plugin-manifest/src/__tests__/gatsby-node.js
```diff
@@ -209,8 +208,31 @@ describe(`Test plugin manifest options`, () => {
...
+    await onPostBootstrap(
+      { reporter },
+      {
+        ...manifestOptions,
+        ...pluginSpecificOptions,
+      }
+    )
+
+    expect(sharp).toHaveBeenCalledTimes(3)
+    const content = JSON.parse(fs.writeFileSync.mock.calls[0][1])
```
DSchau: Not new here (so this is a nit pick!), but couldn't this be written as:

```js
expect(fs.writeFileSync).toHaveBeenCalledWith(expect.any(String), [
  expect.objectContaining({
    purpose: 'all'
  }),
  expect.objectContaining({
    purpose: 'maskable'
  })
], expect.any(String))
```

Something to consider for the future, but does not need to be implemented here!

packages/gatsby-plugin-manifest/src/gatsby-node.js
```diff
@@ -52,16 +52,22 @@ exports.onPostBootstrap = async ({ reporter }, pluginOptions) => {
...
   // If icons are not manually defined, use the default icon set.
   if (!manifest.icons) {
     manifest.icons = defaultIcons
   }
 
   // Specify extra options for each icon (if requested).
-  manifest.icons.forEach(icon => {
-    Object.assign(icon, pluginOptions.icon_options)
-  })
+  if (pluginOptions.icon_options) {
```
DSchau: Seems cleaner to not mutate here. Nicely done!

packages/gatsby-plugin-manifest/src/__tests__/gatsby-node.js
moonmeister: The whole point of this test is to make sure these extra properties don't show up in what's written to the manifest. I needs to be there so it can be deleted.

packages/gatsby-plugin-manifest/src/__tests__/gatsby-node.js
moonmeister: Probably, I will not claim to be an 'expect' expert.

moonmeister:
I think this is ready. Builds are failling for all PRs right not cause of a CI issue.

packages/gatsby-plugin-manifest/src/__tests__/gatsby-node.js
DSchau: This test should probably be re-written to make that more clear then, e.g. something like

```
const manifestOptions = {
  theme_color_in_head: false,
  cache_busting_mode: `name`
}

const iconOptions = {
  icon_options: {},
  something_else: true
}

// call the method
// assert no extra options were passed
expect(whatever).toEqual(manifestOptions)
```

otherwise it's sorta testing a side effect. This test could fail in the future, and it's not clear _why_ it's failing because it's not clear what it's testing.

DSchau:
```
Successfully published:
 - gatsby-plugin-manifest@2.0.26
```

PR #12909, opened 2019-03-28
Use schema metadata

freiksenet:
## Description

Store info about infer config in schema metadata. This allows specifying infer config for builder types too.

### DIFF
--- a/packages/gatsby/src/redux/reducers/schema-customization.js
+++ b/packages/gatsby/src/redux/reducers/schema-customization.js
@@ -21,9 +21,20 @@ module.exports = (
     case `CREATE_TYPES`: {
       let types
       if (_.isArray(action.payload)) {
-        types = [...state.types, ...action.payload]
+        types = [
+          ...state.types,
+          ...action.payload.map(typeOrTypeDef => {
+            return {
+              typeOrTypeDef,
+              plugin: action.plugin,
+            }
+          }),
+        ]
       } else {
-        types = [...state.types, action.payload]
+        types = [
+          ...state.types,
+          { typeOrTypeDef: action.payload, plugin: action.plugin },
+        ]
       }
       return {
         ...state,
--- a/packages/gatsby/src/schema/schema.js
+++ b/packages/gatsby/src/schema/schema.js
@@ -202,21 +202,23 @@ const processAddedType = ({
   schemaComposer.addSchemaMustHaveType(typeComposer)
 
   typeComposer.setExtension(`createdFrom`, createdFrom)
-  typeComposer.setExtension(`plugin`, plugin)
+  typeComposer.setExtension(`plugin`, plugin ? plugin.name : null)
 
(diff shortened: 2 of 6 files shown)


packages/gatsby/src/redux/reducers/schema-customization.js
```diff
@@ -21,9 +21,20 @@ module.exports = (
     case `CREATE_TYPES`: {
       let types
       if (_.isArray(action.payload)) {
-        types = [...state.types, ...action.payload]
+        types = [
+          ...state.types,
+          ...action.payload.map(typeOrTypeDef => {
+            return {
+              typeOrTypeDef,
+              plugin: action.plugin,
```
stefanprobst: What is the shape of `plugin` here? If it is more than `plugin.name`, we can just pluck off the name.

packages/gatsby/src/schema/schema.js
```diff
@@ -174,6 +200,42 @@ const processAddedType = ({ schemaComposer, type, parentSpan }) => {
...
   schemaComposer.addSchemaMustHaveType(typeComposer)
+
+  typeComposer.setExtension(`createdFrom`, createdFrom)
+  typeComposer.setExtension(`plugin`, plugin)
+
+  if (createdFrom === `sdl`) {
+    if (type.astNode && type.astNode.directives) {
+      type.astNode.directives.forEach(directive => {
+        if (directive.name.value === `infer`) {
+          typeComposer.setExtension(`inferConfig`, {
```
stefanprobst: I guess in this case there is no good way, but going forward I think it would be good to keep the naming of (user-facing) directives and extensions as parallel as possible.

packages/gatsby/src/schema/schema.js
```diff
@@ -174,6 +200,42 @@ const processAddedType = ({ schemaComposer, type, parentSpan }) => {
     }
   }
   schemaComposer.addSchemaMustHaveType(typeComposer)
+
+  typeComposer.setExtension(`createdFrom`, createdFrom)
```
stefanprobst: Just thinking out loud: is there value in grouping certain extensions under `extensions.metadata`? Hmm probably not because we'd always have to deepmerge

freiksenet: Yeah, let's not do that. I don't want to make a huge utility module for that.

packages/gatsby/src/redux/reducers/schema-customization.js
freiksenet: Good idea.

packages/gatsby/src/schema/schema.js
freiksenet: Fixed.

freiksenet: Mostly, addDefaultResolvers is going to be different, but we are deprecating that.

PR #12914, opened 2019-03-28
fix(babel-preset-gatsby): Optimize babel config

alexandernanberg:
## Description

Implements some of the optimizations CRA have done in their babel config

https://github.com/facebook/create-react-app/blob/master/packages/babel-preset-react-app/create.js

## Related Issues
N/A


packages/babel-preset-gatsby/src/index.js
```diff
@@ -62,19 +64,44 @@ module.exports = function preset(_, options = {}) {
...
+        },
+      ],
       [
         resolve(`@babel/plugin-proposal-class-properties`),
         {
           loose: true,
         },
       ],
+      [
+        resolve(`@babel/plugin-proposal-object-rest-spread`),
```
KyleAMathews: Is this needed? AFAIK @babel/preset-env already includes this.

packages/babel-preset-gatsby/src/index.js
```diff
@@ -62,19 +64,44 @@ module.exports = function preset(_, options = {}) {
       ],
     ],
     plugins: [
+      [
+        resolve(`@babel/plugin-transform-destructuring`),
```
KyleAMathews: Can you copy over the comment from CRA as well that this is for performance and link to the issue?

packages/babel-preset-gatsby/src/index.js
```diff
@@ -62,19 +64,44 @@ module.exports = function preset(_, options = {}) {
...
+        },
+      ],
       resolve(`babel-plugin-macros`),
       resolve(`@babel/plugin-syntax-dynamic-import`),
       [
         resolve(`@babel/plugin-transform-runtime`),
         {
           helpers: true,
           regenerator: true,
+          useESModules,
```
KyleAMathews: I don't understand this change — what's the reasoning? It doesn't seem to be something CRA is doing.

packages/babel-preset-gatsby/src/index.js
alexandernanberg: Yeah it's indeed included but AFAIK you need to explicitly include it to set the`useBuiltIns` option. But might not be needed since we have set `loose: true` in the `preset-env`, couldn't find anything about it in the docs though 🤔

packages/babel-preset-gatsby/src/index.js
alexandernanberg: This change is _probably_ unnecessary since we already use `loose` in the `preset-env`, so I'll remove this.

packages/babel-preset-gatsby/src/index.js
alexandernanberg: CRA actually does include this [here](https://github.com/facebook/create-react-app/blob/265c1592dcf3853122dcdd50e1bfc1e43f697fbe/packages/babel-preset-react-app/create.js#L32-L36).

And [here](https://babeljs.io/docs/en/babel-plugin-transform-runtime#useesmodules) is some more information about it. So basically since we use webpack (which can handle es modules) we can turn on `useESModules` to save some bytes.

PR #12917, opened 2019-03-28
fix(gatsby): add the path prefix to the URL in the terminal instructions

janmonschke:
## Description

I noticed that when I ran `gatsby develop --prefix-paths`, the URLs in the terminal instructions did not include the prefixed path. Before it would say

```
(...)
You can now view gatsby-starter-blog in the browser.

  http://localhost:8000/
(...)
```

and with the changes in this PR, it will now say

```
(...)
You can now view gatsby-starter-blog in the browser.

  http://localhost:8000/blog/
(...)
```

I understand that running `gatsby develop` with `--prefix-paths` might not be a common use-case and therefore this fix only has limited impact. 

I also noticed that `lanUrlForConfig` is returned from `prepareUrls` but it is apparently never read anywhere. Would it be okay to remove it from the returned object?

### DIFF
--- a/packages/gatsby/src/commands/develop.js
+++ b/packages/gatsby/src/commands/develop.js
@@ -314,21 +314,28 @@ module.exports = async (program: any) => {
     })
   })
 
-  function prepareUrls(protocol, host, port) {
-    const formatUrl = hostname =>
+  function prepareUrls(protocol, host, port, path) {
+    const formatUrl = (hostname, pathname) =>
       url.format({
         protocol,
         hostname,
         port,
-        pathname: `/`,
+        pathname,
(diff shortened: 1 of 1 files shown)


janmonschke:
Is there a way to re-run the tests on circle-ci? They did not fail due to my changes ;)

KyleAMathews:
This is deliberate as we don't prefix paths during development https://www.gatsbyjs.org/docs/path-prefix/#development

This is potentially something that can be changed as I can definitely see it being confusing!

Sorry for the delay responding to your PR! It slipped through the cracks. Generally we try to respond within a working day :-(

PR #12920, opened 2019-03-28
fix(gatsby-plugin-google-tagmanager): update dataLayer to be camelCased

alexjsdev:
## Description

For the `gatsby-plugin-google-tagmanager` this PR/release: https://github.com/gatsbyjs/gatsby/pull/12783, has introduced a bug where by default the `'dataLayer'` is being set to `'datalayer'` (all lower case). Which means that by default this will not work with GTM. This PR just updates to use the correct form, `dataLayer` as the default. It is possibly the smallest PR I've ever made, but it will prevent a lot of people from being confused why they can't push to the dataLayer... 

## Related Issues
issue i filed for this: 
https://github.com/gatsbyjs/gatsby/issues/12921


alexjsdev:
looks like these tests are failing on `yarn install`... i didn't change anything other than making an "l" => "L" uppercase 😄

spwisner:
This bug was caused by my previous pull request and it should be "dataLayer" instead of "datalayer". Sorry. This LGTM.

DSchau:
Published as gatsby-plugin-google-tagmanager@2.0.12

PR #12927, opened 2019-03-28
fix(gatsby-plugin-sharp): don't write to same temporary time multiple times at a same time

### LINKED ISSUE #8301: gatsby-plugin-sharp tracedSVG crashes build
## Description

Using tracedSVG on a file ending in .jpeg causes gatsby-plugin-sharp to report the following vague error:

```
error UNHANDLED EXCEPTION


  TypeError: Cannot read property 'bitmap' of undefined
  
  - Potrace.js:1000 Potrace._processLoadedImage
    [www]/[potrace]/lib/Potrace.js:1000:35
  
  - Potrace.js:1046 Jimp.<anonymous>
    [www]/[potrace]/lib/Potrace.js:1046:14
  
  - index.js:85 Jimp.throwError
    [www]/[jimp]/index.js:85:44
  
  - index.js:201 ReadFileContext.callback
    [www]/[jimp]/index.js:201:44
```

Of note, Twitter always uses .jpeg for its image suffix, so any images saved from Twitter will crash tracedSVG.

### Steps to reproduce

1. Upload a .jpeg file to Contentful
2. Load the image in GraphQL using `tracedSVG`

### Expected result

- Image is traced into an SVG, **or**
- Gatsby tells me which error could not be traced

### Actual result

Vague error message.

### Environment

```

  [4mSystem:[0m
    OS: Linux 4.13 Ubuntu 17.10 (Artful Aardvark)
    CPU: x64 Intel(R) Core(TM) i7-7600U CPU @ 2.80GHz
    Shell: 5.2 - /usr/bin/zsh
  [4mBinaries:[0m
    Node: 8.11.2 - /usr/bin/node
    npm: 6.4.1 - /usr/bin/npm
  [4mBrowsers:[0m
    Chrome: 67.0.3396.62
    Firefox: 61.0.1
  [4mnpmPackages:[0m
    gatsby: ^2.0.0-rc.28 => 2.0.0-rc.28 
    gatsby-image: ^2.0.0-rc.4 => 2.0.0-rc.4 
    gatsby-plugin-catch-links: ^2.0.2-rc.1 => 2.0.2-rc.1 
    gatsby-plugin-manifest: ^2.0.2-rc.1 => 2.0.2-rc.1 
    gatsby-plugin-react-helmet: ^3.0.0-rc.1 => 3.0.0-rc.1 
    gatsby-plugin-react-next: next => 2.0.1-13 
    gatsby-plugin-remove-trailing-slashes: ^1.0.9 => 1.0.9 
    gatsby-plugin-sass: ^2.0.0-rc.2 => 2.0.0-rc.2 
    gatsby-plugin-sharp: ^2.0.0-rc.7 => 2.0.0-rc.7 
    gatsby-plugin-sitemap: ^2.0.0-rc.2 => 2.0.0-rc.2 
    gatsby-source-contentful: ^2.0.1-rc.9 => 2.0.1-rc.9 
    gatsby-source-filesystem: ^2.0.1-rc.6 => 2.0.1-rc.6 
    gatsby-transformer-sharp: ^2.1.1-rc.3 => 2.1.1-rc.3 
    gatsby-transformer-yaml: ^2.1.1-rc.6 => 2.1.1-rc.6 
  [4mnpmGlobalPackages:[0m
    gatsby-cli: 2.4.1

```

pieh:
Memoization key for tracedSVG was not entirely correct and we were trying to write to temporary file multiple times. This was causing those temporary files to become corrupted in dependencies of `potrace` were crashing while trying to read those corrupted input files (there was a lot of concurrency, so that's why those crashes were very random - sometimes timing was bad and we would end up . This _should_ fix memoization key and avoid writing to same file more than once - there is also extra memoization layer just for that, so if general memoization key for `traceSVG` function fails, there is another memoization that checks just temporary file location.

This was happening most likely when there were copies of same image in multiple places (as memoization key was using `absolutePath` before).

I've moved some code around (mostly because I needed to export some trace SVG related functions for added tests). I'll try to mark places in removed code that did change.

related issue #8301

### DIFF
--- a/packages/gatsby-plugin-sharp/src/index.js
+++ b/packages/gatsby-plugin-sharp/src/index.js
@@ -18,14 +18,18 @@ try {
 }
 
 const sharp = require(`sharp`)
-const crypto = require(`crypto`)
+
 const imageSize = require(`probe-image-size`)
-const { promisify } = require(`bluebird`)
+
 const _ = require(`lodash`)
 const fs = require(`fs-extra`)
 const path = require(`path`)
+
 const { scheduleJob } = require(`./scheduler`)
 const { createArgsDigest } = require(`./process-file`)
+const { reportError } = require(`./report-error`)
+const { getPluginOptions, healOptions } = require(`./plugin-options`)
+const { memoizedTraceSVG } = require(`./trace-svg`)
 
 const imageSizeCache = new Map()
 const getImageSize = file => {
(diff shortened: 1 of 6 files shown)


packages/gatsby-plugin-sharp/src/index.js
```diff
@@ -639,94 +557,10 @@ async function fixed({ file, args = {}, reporter, cache }) {
...
-  )
-
-  return trace(tmpFilePath, optionsSVG)
-    .then(optimize)
-    .then(svgToMiniDataURI)
-}
-
-const memoizedTraceSVG = _.memoize(
-  notMemoizedtraceSVG,
-  ({ file, args }) => `${file.absolutePath}${JSON.stringify(args)}`
```
pieh: This was changed to
```js
({ file, args, fileArgs }) => `${file.internal.contentDigest}${JSON.stringify(args)}${JSON.stringify(fileArgs)}`
```

packages/gatsby-plugin-sharp/src/index.js
```diff
@@ -639,94 +557,10 @@ async function fixed({ file, args = {}, reporter, cache }) {
...
-  }-${crypto
-    .createHash(`md5`)
-    .update(JSON.stringify(fileArgs))
-    .digest(`hex`)}.${file.extension}`
-
-  await new Promise(resolve =>
-    pipeline.toFile(tmpFilePath, (err, info) => {
-      resolve()
-    })
-  )
```
pieh: generation of temporary input file was refactored into separate function `notMemoizedPrepareTraceSVGInputFile` and added memoization on top that use `tmpFilePath` as key

tylermenezes:
(This doesn't fully fix #8301 -- there are a bunch of other reasons you could get that error.)

pieh:
@tylermenezes can you expand on that?

tylermenezes:
Basically "media files potrace supports" is a subset of "media files". I'm still not totally sure why my original test case wasn't supported (maybe it was because it also had a colon in the name), but I troubleshooted this as a fix and it didn't solve it.

This is a useful fix for many people, I just think it's premature to call the original issue really fixed, when the error message is so generic that literally any problem will cause it.

pieh:
> Basically "media files potrace supports" is a subset of "media files". I'm still not totally sure why my original test case wasn't supported (maybe it was because it also had a colon in the name), but I troubleshooted this as a fix and it didn't solve it.

Can you share your reproduction test case? I'm not sure if this is about `.jpeg` extension, but I did try using those even before the fix and it was working, so it might more file specific?

> This is a useful fix for many people, I just think it's premature to call the original issue really fixed, when the error message is so generic that literally any problem will cause it.

Sure, no argument here, I'll change my PR description to not autoclose linked issue, and will post message there so people update and report back.

DSchau:
@tylermenezes just to be clear, you tried this fix out with your example--and still running into errors?

I know we asked in the original thread, but what would be _most_ helpful here is if we could access that reproduction. Feel free to DM me on [twitter](https://twitter.com/schaudustin) if any secret details could be shared. We haven't really had a reproducible example to craft this fix--so anything towards that end will ensure we're able to fix _each_ occurrence of this issue.

Thank you!

tylermenezes:
You're welcome to close it if you're convinced it will fix the issue. I didn't save any images that caused errors so I'm not able to provide the same reproduction I had last time.

marcosilvestroni:
Hi, Maybe It could help to find an other usecase,  I've got the same problem here while I'm running in develop and then try to build the project, and vice versa. I have to clean the directrories to make it works or remove the tracedSVG from my gasby-node.js query while generating pages

AlmogRnD:
Has this been fixed I'm getting the same issue?

fraserisland:
I'm getting this issue when querying for tracedSVG's. I'm getting different errors depending on the file type. 
- PNG's work with no errors. 
- WebP images will crash the builds with: `TypeError: Cannot read property 'bitmap' of undefined  at Potrace._processLoadedImage`. 
- JPEG's throw this error: `"VipsJpeg: Premature end of JPEG file\nVipsJpeg: out of order read at line 224\n"`, however don't crash the build. 

```
"gatsby-source-contentful": "4.4.1",
"gatsby-transformer-sharp": "2.10.1",
"gatsby-plugin-sharp": "2.12.2",
"gatsby": "2.30.3",
"gatsby-image": "2.9.0",
```

Issue #12813, opened 2019-03-24
[themes] component shadowing has unexpected behavior with parent themes


jlengstorf:
I just hit some weird behavior with component shadowing that I _think_ is a bug.

Here's what I'm doing:

1. I have a parent theme
2. I have a child theme that shadows a parent component
3. i have a site that needs to shadow a parent component

My goal was to allow shadowing in the child theme by defining the shadowed component in `src/components` in the child theme, then exporting it from `src/gatsby-theme-a/components/foo.js`:

```jsx
import Foo from '../../components/foo';

export default Foo;
```

My expectation is that the file at `gatsby-theme-b/src/components/foo.js` is shadowable, and that this shadowed component would be picked up by `gatsby-theme-a` in the import/export dance.

This doesn't work as expected, though. Somehow, the import/export dance seems to break shadowing for that component.

I put together a minimal repro showing what's going on. `gatsby-theme-a` is the parent, `gatsby-theme-b` is the child.

https://github.com/jlengstorf/component-shadowing-weirdness

Am I doing something wrong, or is this actually a bug?

ChristopherBiscardi:
It's a bug in the parent-shadow-import.

If
the `issuer`: `theme-b/src/gatsby-theme-a/components/foo.js`
and the required path: `theme-b/src/components/foo.js`
end with the [same suffix](https://github.com/gatsbyjs/gatsby/blob/03b766210aec74d804fcd528899bb3f13f33d490/packages/gatsby/src/internal-plugins/webpack-theme-component-shadowing/index.js#L54) that is the universal component path (that is, the path at which this component would sit in any theme or site base location: `components/foo.js`), we [skip shadowing](https://github.com/gatsbyjs/gatsby/blob/03b766210aec74d804fcd528899bb3f13f33d490/packages/gatsby/src/internal-plugins/webpack-theme-component-shadowing/index.js#L56).

ChristopherBiscardi:
We could make this check more restrictive and only check valid shadow paths instead of the "same suffix" that worked for the PoC. Basically this would consist of constructing the list of potential paths and then running `potentialPaths.includes(issuer)` without the extensions.

jlengstorf:
That seems right to me. This pattern makes it possible to avoid needing to know the parent theme(s) to shadow properly. (E.g. using `gatsby-theme-b` but needing to shadow `gatsby-theme-a`.)

[...]

PR #12930, opened 2019-03-29
fix(themes): Restrict the paths that count as shadowable for an issuer

jlengstorf:
This is an attempt to fix https://github.com/gatsbyjs/gatsby/issues/12813 by:

1. Creating a list of possible shadowing paths
2. Checking if the requested component path matches
3. Skipping the shadowing behavior

**NOTE: Right now this _does not_ work.**

I know I’m doing something wrong with the comparison, but I’m not exactly sure _what_ I’m doing wrong.

### DIFF
--- a/packages/gatsby/src/internal-plugins/webpack-theme-component-shadowing/__tests__/index.js
+++ b/packages/gatsby/src/internal-plugins/webpack-theme-component-shadowing/__tests__/index.js
@@ -1,0 +1,57 @@
+import ShadowingPlugin from "../"
+
+describe(`Component Shadowing`, () => {
+  it(`gets matching themes`, () => {
+    const plugin = new ShadowingPlugin({
+      themes: [`a-theme`, `theme-b`, `gatsby-theme-c`].map(name => {
+        return {
+          themeName: name,
+          themeDir: `/some/place/${name}`,
+        }
+      }),
+    })
+    expect(
+      // simple request path to a theme's component
+      plugin.getMatchingThemesForPath(
+        `/some/place/a-theme/src/components/a-component`
+      )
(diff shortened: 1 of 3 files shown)


ChristopherBiscardi:
I've written code in this PR now, so it should be reviewed by @johno or other @gatsbyjs/themes team members instead of myself.


The code I added introduced a refactor so that I could add more comments and tests, with the goal of making the code more readable in general. It also fixed the original issue as a byproduct of introducing the additional named functions and tests.

jlengstorf:
@ChristopherBiscardi should I pull this out of draft status? Trying to keep the context of what's happening in my head causes my brain to bubble out of my ears, but from what I can tell this looks solved pending a review, right?

ChristopherBiscardi:
@johno is doing some work to understand the PR to put a review up, so yeah we can pull this out of draft status. There's one bug relating to the issuer extension comparison that needs to be fixed.

packages/gatsby/src/internal-plugins/webpack-theme-component-shadowing/__tests__/index.js
```diff
@@ -0,0 +1,57 @@
...
+        return {
+          themeName: name,
+          themeDir: `/some/place/${name}`,
+        }
+      }),
+    })
+    expect(
+      // simple request path to a theme's component
+      plugin.getMatchingThemesForPath(
+        `/some/place/a-theme/src/components/a-component`
```
ChristopherBiscardi: this path needs to be constructed with `path.join`

ChristopherBiscardi: so it works on windows

PR #12931, opened 2019-03-29
fix(query-extraction): handle duplicated graphql query names error

d4rekanguok:
## Description

Previously, when there were duplicated graphql query operation names ( i.e `query SiteTitleQuery { site { ... } }`), gatsby threw an error and carried on:

```
success update schema — 0.015 s
error GraphQL Error There was an error while compiling your site's GraphQL queries.
  Error: RelayParser: Encountered duplicate defintitions for one or more documents: each document must have a unique name. Duplicated documents:
- SiteTitleQuery
success extract queries from components — 0.075 s
```

In the latest release (after [this commit?](https://github.com/gatsbyjs/gatsby/commit/91086d43e9b69ba357f7f7ad6bbf2eb0b8b5fbff)), it'll throw an unhandled exception:

```
success update schema — 0.015 s
error UNHANDLED REJECTION
  TypeError: expected a string
  
  - index.js:12 normalizePath
    [gatsby-test]/[normalize-path]/index.js:12:11
  
  - components.js:99 module.exports
    [gatsby-test]/[gatsby]/dist/redux/reducers/components.js:99:40
  
  - redux.js:451 combination
    [gatsby-test]/[redux]/lib/redux.js:451:29
  
  - redux.js:211 dispatch
    [gatsby-test]/[redux]/lib/redux.js:211:22
  
  - index.js:78 action
    [gatsby-test]/[gatsby]/dist/redux/index.js:78:91
  
  - redux.js:468 Object.queryExtractionGraphQLError
    [gatsby-test]/[redux]/lib/redux.js:468:12
```

This pull request modifies `redux/reducers/components.js` so that `state` is returned for `QUERY_EXTRACTION_GRAPHQL_ERROR` & `QUERY_EXTRACTION_BABEL_ERROR` so the error can be seen on the console again.

## Step to reproduce

Just have duplicate query operation name in a site.


pieh:
Hey, thanks for PR. This approach fixes the crash, but it does introduce some other problems - I will push less workaroundy approach here.

Because this is "global" graphql extraction error and not one tied to specific file, we can't extract file that is responsible for it. So in cases like that we need to send `QUERY_EXTRACTION_GRAPHQL_ERROR` to all services (and not omit sending it at all)

Also /cc @KyleAMathews

wardpeet:
@pieh sweet! do we need some kind of test for this? (or don't we have anything in place yet)

d4rekanguok:
@pieh Thanks for the quick response & explanation of why it wouldn't work!

Sidetracking a little bit, I come across this error when I was trying to understand why query name has to be unique in gatsby, as it doesn't seem to be a requirement for graphql whereelse. 

I discovered that gatsby had already generating a name for unnamed queries, i thought it wouldn't be hard or cost a lot to make the name unique by appending file name to existing query name. So I was wondering, is it by design that query name has to be unique?

pieh:
> @pieh sweet! do we need some kind of test for this? (or don't we have anything in place yet)

Ideally yes, but I don't see a way to do this without mocking half of gatsby :/

@d4rekanguok 

Direct requirement comes from `relay-compiler` which we use to normalize queries and also auto-attach fragments (that's why you can use fragments like https://www.gatsbyjs.org/packages/gatsby-image/#gatsby-transformer-sharp without declaring them inside same query. We probably could handle this case as well, but if someone explicitely define name, then we don't change it (but we do autogenerate names if someone didn't provide name themselves)

DSchau:
Successfully published:
 - gatsby@2.3.11

PR #12936, opened 2019-03-29
refactor(gatsby-remark-code-repls): move config per provider

deblasis:
## Description

Currently it's possible to upload multiple files only by using CodeSandbox.

This PR adds an option that allows adding CSS files to CodePen examples by setting the css property in the payload object.

Ref: https://blog.codepen.io/documentation/api/prefill/

Matching CSS means that if the user is linking to example.js and the option is set to true, if example.css is present in the same directory, its content is sent to CodePen.

The default behaviour is left untouched
If the option is set to true and the matched file doesn't exist, nothing changes
tests included

## Related Issues

This is based upon the suggestions in PR #12110 and it includes the breaking API changes


packages/gatsby-remark-code-repls/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "gatsby-remark-code-repls",
   "description": "Gatsby plugin to auto-generate links to popular REPLs like Babel and Codepen",
-  "version": "2.0.8",
+  "version": "3.0.0",
```
wardpeet: We will make sure we publish a major :)
```suggestion
  "version": "2.0.8",
```

deblasis:
> This is really great! thanks for the refactoring. I'm going to test this one

Awesome! You are welcome, happy to contribute back a little bit :)

deblasis:
Hi @wardpeet , the PR that adds the feature was merged, this one is the API refactoring one. Let's remember to publish a major version since the API changed as suggested 😄

LekoArts:
Hey @wardpeet 	👋 	Seems like this needs final approval, want to have a look? Thanks!

wardpeet:
I was testing this plugin on reactjs.org site and it seemed like it wasn't working great on windows machines so I made it work. Will publish when everything is green on new major

Issue #12938, opened 2019-03-29
Large amount of locales/data enabled in contentful case "JavaScript heap out of memory"


stoltzrobin:
## Summary
Enable the user to filter out unnecessary locales from Contentful.

### Basic example
```
 {
      resolve: `gatsby-source-contentful`,
      options: {
        [... other settings needed]
        localeFilter: locale =>
          locale.code === ACTIVE_LOCALE || locale.code === DEFAULT_LOCALE,
      },
},
```

By passing a filter function we can limit the amount of nodes created in graphQL.

### Motivation

In our case we have over 50 locales in Contentful and over 2500 entries (this numbers are increasing every week as well). This create a lot of nodes in the Contentful plugin, which case the JS VM instance to run our of memory. In our application we only build one locale per build (to improve build speed) and all other nodes from different locales an useless and not needed.

In our application it reduces the JS VM runtime memory around 80%.

wardpeet:
Thanks for filling an issue and fixing it yourself! 💪 that's just bonkers!

Could you provide a bit more info about where it fails exactly? Just making sure it's not a known issue with saving redux state as we have an issue open for sites with lots of pages.

https://github.com/gatsbyjs/gatsby/issues/12343

Would be handy if it's the same issue that you can comment in that thread as well.

thibautRe:
I can provide more info (Robin and I work together, and he's on holiday as of now <sup>I'm super jealous btw! :D</sup> so he might not be able to answer until he's back).

The errors we started to experience randomly are memory issues (not speed), and would lead to our JS runtime running out of memory and throwing a wonderful C++ error. Using `--max-old-space-size` was helping us for local development but were not fixing the issue for Cloud builds - where the machines have limited amount of memory.

The OutOfMemoryError would get thrown randomly - sometimes not at all, and at random times... Most of the time the error would happen between the node sourcing and the graphql schema generation steps. I don't remember us seeing the error during the graphql queries step, or HTML generation step.

Also, we tried using the `lokijs` feature flag but that didn't help.

thibautRe:
Fixed in #12939

PR #12939, opened 2019-03-29
feat(gatsby-source-contentful): add locale filter

stoltzrobin:
## Description
This allows the user to filter out locales that are not needed for the build. For example, if the space in Contentful is rather large with very many locales the javascript memory easiely runs out of memory.

## Related Issues
#12938


pieh:
Is it possible to instruct `contentful` client to fetch data just for given locale? If not this seems good to go.

pieh:
I think I just answered my own question when checking https://www.contentful.com/developers/docs/concepts/sync/

> Because the Sync API retrieves all localized content, it might be better to use the delivery API to retrieve results of a single locale.

packages/gatsby-source-contentful/src/normalize.js
```diff
@@ -244,9 +244,10 @@ exports.createContentTypeNodes = ({
   foreignReferenceMap,
   defaultLocale,
   locales,
+  localeFilter,
 }) => {
   const contentTypeItemId = contentTypeItem.name
-  locales.forEach(locale => {
+  locales.filter(localeFilter || (() => true)).forEach(locale => {
```
pieh: I think we could filter locales earlier - there are few other places where we iterate through locales (for example when creating asset nodes).

stoltzrobin:
@pieh that was my initial though as well but as shown in the link that you posted it's not really the idea of the `.sync()`function. Maybe in the future it should be nice to support to pass queries in the sync call as well

packages/gatsby-source-contentful/src/normalize.js
pieh: Place I'm thinking of is https://github.com/gatsbyjs/gatsby/blob/5a13c3dd3a2be3346218da14ad7d7c924bfc6ba8/packages/gatsby-source-contentful/src/fetch.js#L24. If we filter it there - locale will be filtered before we iterate over them when creating nodes

stoltzrobin: Sounds like a good idea! Will change that!

pieh:
Looking good now, gotta check it out in practice before merging

stoltzrobin:
@pieh nice 👍

thibautRe:
Hello @pieh, just checking if there's anything I can do to help get this approved and merged :)

packages/gatsby-source-contentful/README.md
```diff
@@ -144,6 +144,12 @@ Downloads and caches `ContentfulAsset`'s to the local filesystem. Allows you to
 
 You can pass in any other options available in the [contentful.js SDK](https://github.com/contentful/contentful.js#configuration).
 
+**`localeFilter`** [function][optional] [default: `() => true`]
+
+Possibility to limit how many locales/nodes are created in graphQL. This can limit the memory usage by reducing the amout of nodes created. Useful if you have a large space in contentful and only want to get the data from one selected locale.
+
+For example, to filter locales on only germany `localeFilter: locale => locale.code === 'de-DE'`
```
pieh: I just tried it with space we are using in our example, and was confused why this wasn't working, but apparently code for German localization in that space is just `de`. I think I'll add some console output that lists locale so it's easier for users to specify locale code

pieh: Ok, output is getting messy if I do it
Screenshot 2019-04-16 at 12 32 04

So maybe instead of that (at least before https://github.com/gatsbyjs/gatsby/pull/13089 land) We could do:
```suggestion
For example, to filter locales on only germany `localeFilter: locale => locale.code === 'de-DE'`

List of locales and their codes can be found in Contentful app -> Settings -> Locales
```

thibautRe: That sounds totally fine to me! I actually also got confused when 

(I can't push commits to Robin's PR though I realized, so it's either I fork that PR and create my own with that change, or if you have the possibility to add commits, feel free to add it)

pieh: I'll try to push it directly and merge it.

I checked - in locale settings you can pick both `de` and `de-DE` there - so it depends on which one you chose from dropdown it will use that code. But locale overview also display codes so that should be fine to add here.

pieh:
Thanks @stoltzrobin, @Khaledgarbaya and @thibautRe for help with getting it merge-ready!

PR #12941, opened 2019-03-29
tmp(CircleCi): use temporary libvps dists

pieh:


sidharthachatterjee:
Relevant issue in sharp: https://github.com/lovell/sharp/issues/1632

PR #12942, opened 2019-03-29
fix(www): use a flatten to get all pages

DSchau:
## Description

Due to [this line](https://github.com/gatsbyjs/gatsby/pull/12779/files#diff-355f19a97a68170ef2cd96a7e96c3398R449) we broke the build because the `page.items` compares against `undefined` instead of null.

```
success Building production JavaScript and CSS bundles — 101.484 s


  15 | 
  16 |   pages.forEach(page => {
> 17 |     if (page.items !== undefined) {
     |         ^
  18 |       // Recurse downwards
  19 |       stubs.push(...findStubs(page.items))
  20 |     }


  WebpackError: TypeError: Cannot read property 'forEach' of null
  
  - stub-list.js:17 findStubs
    lib/src/pages/contributing/stub-list.js:17:9
  
  - stub-list.js:20 
    lib/src/pages/contributing/stub-list.js:20:6
  
  
  - stub-list.js:17 findStubs
    lib/src/pages/contributing/stub-list.js:17:9
  
  - stub-list.js:20 
    lib/src/pages/contributing/stub-list.js:20:6
```

This fixes the sidebar issue, and introduces a flatten algorithm that is a little cleaner and should be more fault tolerant in the future.

### DIFF
--- a/www/src/data/sidebars/doc-links.yaml
+++ b/www/src/data/sidebars/doc-links.yaml
@@ -455,7 +455,6 @@
       items:
         - title: As a Freelancer
           link: /docs/gatsby-for-freelancers/
-          items:
         - title: Inside an Agency
           link: /docs/gatsby-for-agencies/
           items:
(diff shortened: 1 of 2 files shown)


fk:
🙏

PR #12962, opened 2019-03-31
fix(docs): Update GraphQL description in Tutorial Part Zero

ozanmuyes:
## Description

Update GraphQL description on "Overview of core technologies" subtitle.

Originally `A query language. A programming language that allows you to pull data into your website.` was written. The current explanation is confusing to me and I think it contradictory; is it a query language or a _programming language_. I'd go with the first one.

Also as of my observations just GraphQL is not sufficient to pull data, there must be an implementation on top it (e.g. Apollo Server).


ozanmuyes:
It was a small correction but happy to help :+1:

PR #12963, opened 2019-03-31
chore(docs): add starter library to docs sidebar under starters

calcsam:
40% of people visiting the starters page go here next, so we should probably add this:

image


calcsam:
And while I'm at it, add Plugin Library too, because 35% of people visiting /docs/plugins go _there_ next...

doc

PR #12973, opened 2019-03-31
♻️Use new createContentDigest helper in gatsby-source-medium, gatsby-transformer-toml and gatsby-source-drupal

charlyx:
## Description

Hi 👋

First of all thanks for your work: I'm using gatsby everyday at work! 👏

This PR makes use of the new createContentDigest helper in the following plugins:

* gatsby-source-medium
* gatsby-transformer-toml
* gatsby-source-drupal

## Related Issues

Related to #8805

### DIFF
--- a/packages/gatsby-source-medium/src/gatsby-node.js
+++ b/packages/gatsby-source-medium/src/gatsby-node.js
@@ -76,7 +76,7 @@ exports.sourceNodes = async (
     resources.map(resource => {
       convertTimestamps(resource)
 
-      const contentDigest = createContentDigest(JSON.stringify(resource))
+      const contentDigest = createContentDigest(resource)
 
       const links =
         resource.type === `Post`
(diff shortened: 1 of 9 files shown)


packages/gatsby-source-medium/src/gatsby-node.js
```diff
@@ -77,10 +76,7 @@ exports.sourceNodes = async (
     resources.map(resource => {
       convertTimestamps(resource)
 
-      const digest = crypto
-        .createHash(`md5`)
-        .update(JSON.stringify(resource))
-        .digest(`hex`)
+      const contentDigest = createContentDigest(JSON.stringify(resource))
```
DSchau: The `createContentDigest` actually uses stringify if it receives a non-string!

So we can tweak this to:

```suggestion
      const contentDigest = createContentDigest(resource)
```

sidharthachatterjee:
Published in 
 - gatsby-source-drupal@3.0.33
 - gatsby-source-medium@2.0.7
 - gatsby-transformer-toml@2.1.8

PR #12960, opened 2019-03-31
feat(docs): add to contributing pages

calcsam:
Add some copy to contributing pages to emphasize how excited we are for everyone to help!

PR #12988, opened 2019-04-01
feat(docs): fix eslint warnings

SheetJSDev:
## Description

PR #12960 introduced two apostrophes, violating rule
`react/no-unescaped-entities`.  A basic `yarn test` reveals:

```
  24:63  warning  HTML entities must be escaped  react/no-unescaped-entities
```

This PR replaces the bare apostrophes with `&apos;`

### DIFF
--- a/www/src/pages/contributing/index.js
+++ b/www/src/pages/contributing/index.js
@@ -21,11 +21,11 @@ class IndexRoute extends React.Component {
               Contributing to Gatsby.js
             </h1>
             <p>
-              Thanks for being interested in contributing! We're so glad you
+              Thanks for for your interest in contributing! We&apos;re glad you
               want to help!
             </p>
             <p>
-              Below you'll find guides on the Gatsby.js community, code of
+              Below you&apos;ll find guides on the Gatsby.js community, code of
               conduct, and how to get started contributing:
             </p>
             <ul>


KyleAMathews:
Is this rule actually needed? I don't entirely understand why it's enabled as from reading the rule docs, only not escaping `><` are problematic. Escaping apostrophes seems more safe than sorry.

SheetJSDev:
Github's JSX linter, for one, treats a bare apostrophe as an error.  see https://github.com/gatsbyjs/gatsby/blob/master/www/src/pages/contributing/index.js#L24 in the Github browser preview window.

It's also part of the airbnb config https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/rules/react.js#L357

KyleAMathews:
AirBnB is the poster-child for obnoxious eslint configs :-P

From the interwebs:
>My hard and fast rule is to disable all eslint rules that prettier leaves alone

https://twitter.com/kylpeacock/status/1112889214054498304

KyleAMathews:
Ugh, this is in our eslint config :-(. Removing...

KyleAMathews:
Removed in https://github.com/gatsbyjs/gatsby/pull/13015

Thanks for adding the fix all the same! This seems like the only lint rule that routinely gets by us — probably because the changes are part of small content changes and people (very reasonably!) don't think it needs linted.

ljharb:
There's no need to be hostile; the rule catches bugs. In the case of apostrophes, straight quotes should *never* appear in prose, since they're typographically incorrect (use curly quotes). If you don't want that protection, disabling the rule is fine.

KyleAMathews:
Sorry @ljharb - wasn't meant to be hostile. Just meant to push back against idea that Airbnb's lint preset is a model we want to emulate. I was trying to be joke-y but yeah, it was a bit much. Keep up the good work.

Issue #12961, opened 2019-03-31
gatsby serve 出现端口被占用问题，能不能改gatsby serve 的端口号呀


tgy3300:
F:\gatsby-site>gatsby serve
error UNHANDLED EXCEPTION
  Error: listen EADDRINUSE 127.0.0.1:9000

anastasialanz:
I am also seeing this issue.

sidharthachatterjee:
Like @anastasialanz correctly mentioned, `EADDRINUSE` is thrown when the port is already in use. This is a great opportunity to auto detect and let the user pick a port like we have for develop so I've gone ahead and added that in. 

Thank you for opening this @tgy3300 and for your response @anastasialanz!

sidharthachatterjee:
Published in gatsby@2.3.5

PR #12989, opened 2019-04-01
feat(gatsby-cli): Detect if something is already running on a port when running gatsby serve

sidharthachatterjee:
Because great developer experience and resolves confusion like in https://github.com/gatsbyjs/gatsby/issues/12961

### DIFF
--- a/packages/gatsby/src/commands/develop.js
+++ b/packages/gatsby/src/commands/develop.js
@@ -32,6 +32,7 @@ const slash = require(`slash`)
 const { initTracer } = require(`../utils/tracer`)
 const apiRunnerNode = require(`../utils/api-runner-node`)
 const telemetry = require(`gatsby-telemetry`)
+const detectPortInUseAndPrompt = require(`../utils/detect-port-in-use-and-prompt`)
 
 // const isInteractive = process.stdout.isTTY
 
@@ -260,7 +261,6 @@ module.exports = async (program: any) => {
   telemetry.trackCli(`DEVELOP_START`)
   telemetry.startBackgroundUpdate()
 
-  const detect = require(`detect-port`)
   const port =
     typeof program.port === `string` ? parseInt(program.port, 10) : program.port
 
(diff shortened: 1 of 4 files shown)


packages/gatsby/src/utils/detect-port-in-use-and-prompt.js
```diff
@@ -0,0 +1,30 @@
+const detect = require(`detect-port`)
+const report = require(`gatsby-cli/lib/reporter`)
+
+// Checks if a port is in use and prompts the user to enter another one
+// Then calls callback with new port
+const detectPortInUseAndPrompt = (port, rlInterface, callback) => {
+  let newPort = port
+
+  detect(port, (err, _port) => {
```
wardpeet: how about making a promise interface?
```
return new Promise((resolve, reject) =>
  detect(port, (err, __port) => {
    ...
    resolve(port);
  })
);
```

packages/gatsby/src/commands/serve.js
```diff
@@ -74,27 +90,29 @@ module.exports = async program => {
...
+          )
+        )
+      }
+    })
+  }
 
-  signalExit(() => {
-    telemetry.trackCli(`SERVE_STOP`)
-    server.close()
+  detectPortInUseAndPrompt(port, rlInterface, newPort => {
```
wardpeet: if we go for a promise interface we can write it as
`port = await detectPortInUseAndPrompt(port, rlInterface, newPort)`

packages/gatsby/src/commands/develop.js
```diff
@@ -286,31 +290,12 @@ module.exports = async (program: any) => {
 
   let compiler
   await new Promise(resolve => {
```
wardpeet: if we had a promise interface this would become

```
program.port = await detectPortInUseAndPrompt(port, rlInterface, newPort)

return startServer(program).then(([c, l]) => { compiler = c })
```

packages/gatsby/src/utils/detect-port-in-use-and-prompt.js
sidharthachatterjee: The function is sync, I just needed to conditionally call a passed in function (the callback) 🤔

wardpeet:
```js
rlInterface.question(question, answer => {
        if (answer.length === 0 || answer.match(/^yes|y$/i)) {
          newPort = _port
        }
        callback(newPort)
      })
```
this part is async as we're waiting for an answer, we block everything else but technically it's async but it's not a big deal I just like promise syntax better than callbacks especially in `develop.js`

packages/gatsby/src/commands/develop.js
```diff
@@ -49,10 +51,13 @@ const rlInterface = rl.createInterface({
 
 // Quit immediately on hearing ctrl-c
 rlInterface.on(`SIGINT`, () => {
-  telemetry.trackCli(`DEVELOP_STOP`)
   process.exit()
 })
 
+onExit(() => {
```
jamo: which of these is run first, on SIGINT or on exit. I.e. will this hook exec from `SIGINT` + `process.exit` from above?

packages/gatsby/src/utils/detect-port-in-use-and-prompt.js
DSchau: I'm fine merging this--but it does show a promise based interface (they show it with a `yield`) in the documentation.

packages/gatsby/src/commands/develop.js
sidharthachatterjee: @jamo SIGINT is run first. And `onExit` gets called with the call to `process.exit`. Tested this and moved the call to telemetry in `onExit` since that gets called gracefully even in case of the process getting killed by kill `<pid>`, for example.

sidharthachatterjee:
Published in gatsby@2.3.5

PR #12995, opened 2019-04-01
feat: pathname in SSR APIs when in develop

neo:
Update the docs for this change: https://github.com/gatsbyjs/gatsby/pull/9792.
It's not a breaking change.
It also adds the pathname to `onRenderBody` and `onPreRenderHTML` in developer.


neo:
Actually I also found that it only works in build but not develop?

pieh:
In develop it will always be `/` (because we don't create actual html pages there other than main one), but it probably makes sense to use it anyway for parity

neo:
good point, I added that because it broke `npm start` when I tried to use `pathname` in the SSR APIs.

actually I was also confused that why SSR was called during `gatsby develop`...

Issue #12980, opened 2019-04-01
Correct custom eslint instructions


KyleAMathews:
If people customize their eslint setup, we should recommend they start from our default setup https://www.npmjs.com/package/eslint-config-react-app and go from there.

Our current docs recommend something entirely different:
https://www.gatsbyjs.org/docs/eslint/#configuring-eslint

KyleAMathews:
This is the file that defines our setup https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/utils/eslint-config.js

KyleAMathews:
@kkemple looked into writing out the schema so then a custom eslint config could use it. @freiksenet @stefanprobst — it'd be great to do that https://github.com/gatsbyjs/gatsby/issues/6167

stefanprobst:
@KyleAMathews this is on the [roadmap](https://github.com/gatsbyjs/gatsby/issues/12510). In the meantime there is [this plugin](https://github.com/NickyMeuleman/gatsby-plugin-extract-schema).

KyleAMathews:
Ok, no worries @stefanprobst 

So for now we should just recommend people use:

```js
{
      globals: {
        graphql: true,
        __PATH_PREFIX__: true,
      },
      extends: `react-app`,
}
```

DSchau:
Would love for someone to pick this up--right now we're displaying a document that is a little bit less focused than we'd like.

## Description

- Update [this document](https://github.com/gatsbyjs/gatsby/blob/master/docs/docs/eslint.md) with the above instructions per @KyleAMathews 
    - Why?
        - This document is doing a bit too many things, e.g. adding prettier
        - We should advocate _simple_ set ups that work out of the box and scale as needed
- (Longer term!) we should expose something like `gatsby-config-eslint` which will be an opinionated set of defaults that are currently [internalized here](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/utils/eslint-config.js)
    - This requires changes to dumping our schema, so a quicker win is to first update the documentation and longer term externalize our config so it can be used by anyone

DSchau:
@lukebrobbs not a dumb question, at all!

Yep, exactly. You'll want to edit [this file](https://github.com/gatsbyjs/gatsby/blob/master/docs/docs/eslint.md) and instructions to be a little simpler, and just install the necessary dependencies, e.g. `eslint-config-react-app`

Feel free to reach out if we can help with anything else!

[...]

PR #12998, opened 2019-04-01
Updated es-lint docs to instruct how to install eslint-config-react-app

lukebrobbs:
## Description

- Updated es-lint docs to recommend using `eslint-config-react-app` 
- Removed extra documentation to make the instructions a little more focused.

## Related Issues
#12980

### DIFF
--- a/docs/docs/eslint.md
+++ b/docs/docs/eslint.md
@@ -28,7 +28,7 @@ touch .eslintrc.js
 
 ### Configuring ESLint
 
-The snippet below mimics the [default ESLint config](https://github.com/gatsbyjs/gatsby/blob/master/.eslintrc.json) Gatsby ships with . We recommend copying the content to your newly created `.eslintrc.js` file and modifying it per your needs. Reference ESLint's [rules documentation](https://eslint.org/docs/rules/) for more options.
+The snippet below mimics the [default ESLint config](https://github.com/gatsbyjs/gatsby/blob/master/.eslintrc.json) Gatsby ships with. We recommend copying the content to your newly created `.eslintrc.js` file and modifying it per your needs. Reference ESLint's [rules documentation](https://eslint.org/docs/rules/) for more options.
 
 ```js:title=.eslintrc.js
 module.exports = {
--- a/docs/docs/eslint.md
+++ b/docs/docs/eslint.md
@@ -4,22 +4,23 @@ title: ESLint
 
 ## Why use ESLint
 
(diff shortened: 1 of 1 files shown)


docs/docs/eslint.md
```diff
@@ -8,17 +8,17 @@ Gatsby ships with Prettier, which is a simple, opinionated code _formatter_. [ES
 
 ## How to use ESLint
 
-Here we will explore an ESLint configuration that acts like Prettier by adhering to [Standard.js](https://standardjs.com) rules. ESLint might seem intimidating at first, however it is aimed at providing a number of configurable options to make your code format fit your style. Run the following commands to remove Prettier and install ESLint.
+Here we will explore an ESLint configuration designed to be used with React applications. ESLint might seem intimidating at first, however it is aimed at providing a number of configurable options to make your code format fit your style. Run the following commands to remove Prettier and install ESLint and necessary dependencies.
```
KyleAMathews: We don't want to remove Prettier so please remove that

docs/docs/eslint.md
```diff
@@ -33,57 +33,14 @@ touch .eslintrc.js
 
 ### Configuring ESLint
 
-We recommend copying our default .eslintrc.js content below to your newly created `.eslintrc.js` file and modifying it per your needs. Reference ESLint's [rules documentation](https://eslint.org/docs/rules/) for more options.
+We recommend copying the following .eslintrc.js content below to your newly created `.eslintrc.js` file and modifying it per your needs. Reference ESLint's [rules documentation](https://eslint.org/docs/rules/) for more options.
```
KyleAMathews: Add a note that this mimics the default eslint config that Gatsby ships with.

dakebl: Might be worth mentioning and linking the shipped ESLint file here too for reference. So that developers are aware that a more complex version exists internally: https://github.com/gatsbyjs/gatsby/blob/master/.eslintrc.json

JDDoesDev:
I know this issue has been closed for a while, and I can open a new one if need be, but I find the language confusing on https://www.gatsbyjs.org/docs/eslint/

In the main section it says:

> You’ll replicate (mostly) the ESLint config Gatsby ships with so you can then add additional presets, plugins, and rules.

but further down the page it says:

> When you include a custom .eslintrc file, Gatsby gives you full control over the ESLint configuration. This means that it will override the built-in eslint-loader and you need to enable any and all rules yourself.

On first read-through, it appears to contradict the first statement and imply that by using the snippet on the page, you're losing out on all the things from the default config.

PR #13000, opened 2019-04-01
chore(*): Use new createContentDigest helper

frinyvonnick:
## Description

Hi,

Thank you for your work, I use Gatsby on multiple projects and it's really nice 😄 !

This PR makes use of the new createContentDigest helper in the following plugins:
- gatsby-transformer-excel
- gatsby-source-hacker-news
- gatsby-source-wikipedia
- gatsby-source-wordpress
- gatsby-transformer-javascript-frontmatter
- gatsby-source-graphql
- gatsby-transformer-hjson
- gatsby-source-npm-package-search
- gatsby-source-lever
- gatsby-transformer-javascript-static-exports
- gatsby-transformer-react-docgen
## Related Issues

Related to #8805


packages/gatsby-source-lever/src/normalize.js
```diff
@@ -142,7 +146,7 @@ exports.createNodesFromEntities = ({ entities, createNode }) => {
       children: [],
       internal: {
         type: `lever`,
-        contentDigest: digest(JSON.stringify(entity)),
+        contentDigest: createContentDigest(entity),
```
muescha: you forget to remove this?:
https://github.com/gatsbyjs/gatsby/blob/807560593b24c1a412e0f9c2ffc79d70d150af21/packages/gatsby-source-lever/src/normalize.js#L14-L18
and change this:
https://github.com/gatsbyjs/gatsby/blob/807560593b24c1a412e0f9c2ffc79d70d150af21/packages/gatsby-source-lever/src/normalize.js#L39

frinyvonnick: I did it on purpose the `createGraphQLNode` function is exported but unused in gatsby codebase. No documentation refers to it and I need to add a parameter to the function to make it work, this is a breaking change. I don't know how to do about it, any advice ?

frinyvonnick: I just saw an [example](https://github.com/gatsbyjs/gatsby/pull/8992/files#diff-1a077ef3f1d488e1041a8fb4003d4f53R8) that imports directly `createContentDigest` that would be a good solution for not making a breaking change. What is the difference between the imported version and the injected version @DSchau ?

frinyvonnick: @LekoArts suggested me to make the change anyway 😄

muescha: also if it is not used: i would change it anyway to complete the Job "Use new createContentDigest" 

and later look if we remove `createGraphQLNode` if is not used anymore

muescha: but wait what the @gatsbyjs/core team suggest as the best way

muescha: note: 

my researches about `createGraphQLNode`

- [search all Github](https://github.com/search?q=createGraphQLNode&type=Code): there are some hits in code, but they always are defined and used in the same file
- [search Google](https://www.google.com/search?q="createGraphQLNode"): only 2 hits - one to github and one is just a SEO spam

wardpeet: We should create the breaking change :) It would be weird if people are using it outside of gatsby.

sidharthachatterjee:
Love these kind of cleanup PRs! Thank you so much and congratulations on your first contribution  @frinyvonnick 🎉 

Published in 
```
 - gatsby-transformer-excel@2.1.11
 - gatsby-source-hacker-news@2.0.11
 - gatsby-source-wikipedia@2.0.6
 - gatsby-source-wordpress@3.0.53
 - gatsby-transformer-javascript-frontmatter@2.0.10
 - gatsby-source-graphql@2.0.16
 - gatsby-transformer-hjson@2.1.8
 - gatsby-source-npm-package-search@2.0.5
 - gatsby-source-lever@2.0.8
 - gatsby-transformer-javascript-static-exports@2.1.7
 - gatsby-transformer-react-docgen@4.0.1
```

SleeplessByte:
FWIW, `gatsby-source-graphql` now started erroring with:

```
TypeError: createContentDigest is not a function
```

frinyvonnick:
Oh I made a mistake I destructured it from [second parameter](https://github.com/frinyvonnick/gatsby/blob/8e2b081ea500bb560eae19dbfb32272866358be4/packages/gatsby-source-graphql/src/gatsby-node.js#L18) instead of first parameter...

SleeplessByte:
I was about to comment exactly that. Until the fix is pushed, fixing the version on `2.0.15` does still work :)

frinyvonnick:
I made a patch here: #13214

PR #13004, opened 2019-04-01
[WIP] Replace global data.json with page-data.json per page

Moocar:
## Description

Removes the global `data.json` file and instead provides a `page-data.json` for each page on a gatsby site. See https://github.com/gatsbyjs/gatsby/issues/11982 for background.

This PR is a draft and should not be merged. I'll be providing a series of smaller PRs instead. Reviewers can always reference this PR to see the full context of the changes. That said, if you're reading this PR and spot any problems, then please do leave some feedback!

## Benefits/Trade offs

**pros**

- no more downloading `data.json` to the browser. For 10'000 page sites, this file was already approaching 500kb compressed.
- navigating to prefetched pages should be faster since there's no `data.json` to search through
- `production-app.js` webpack build no longer references pages or their data. Which will make incremental builds much easier to implement in the future

**cons**

- Requesting a page from the server now requires 2 requests. First `page-data.json`, then the component. Though prefetched pages still do this in the background.
- Since page data isn't immutable anymore, we can't rely on infinite browser caching
- Some potential breaking changes in `loader.getResourceURLsForPathname`

## Sup-PRs

Breaking up these PRs is hard. To make it a bit easier, I'm going to have gatsby saving two sources of data to disk. `data.json` and `static/d/...`, as well as the new `page-data.json`. This will require duplication of memory and disk, so we won't be able to merge to master. But we will be able to gradually introduce new functionality, and once it's ready, delete the old data sources and merge to master. I'm not a fan of long lived non-master branches, but I think it makes sense here.

The long lived branch is [per-page-manifest](https://github.com/gatsbyjs/gatsby/tree/per-page-manifest). The PRs that will merge directly into that are:

**TODO**

- [ ] Bad handling of timeouts etc https://sentry.io/organizations/gatsby-inc-9z/issues/1054501319/?project=1471074&statsPeriod=24h
- [X] page-data.json requested a LOT (resulting in 304s)
  - From my reading, it looks like these are all coming from sentry's `breadcrumbs.js` file.

**TODO but won't block release**

- [ ] https://github.com/gatsbyjs/gatsby/issues/13430
- [ ] Review "behind the scenes" docs. They'll be very out of date now.
- [ ] `public/static/d/....` isn't used except for static query results. So we should ask the user to delete their `public` directory when they run gatsby with this code for the first time? It's not required.
- [ ] [Add prefetch for linked component modules to page HTML](https://github.com/gatsbyjs/gatsby/issues/14262)

Already reviewed and Merged:

[...]

PR #13005, opened 2019-04-01
refactor(core): Merge build-html/develop-html

Moocar:
## Description

While working on https://github.com/gatsbyjs/gatsby/pull/13004 I noticed that `build-html`/`develop-html` are essentially the same thing. So I refactored them to be so.

## Related Issues


packages/gatsby/src/commands/build-html.js
```diff
@@ -1,69 +1,76 @@
...
 const webpackConfig = require(`../utils/webpack.config`)
-const { store } = require(`../redux`)
 const { createErrorFromString } = require(`gatsby-cli/lib/reporter/errors`)
 const renderHTMLQueue = require(`../utils/html-renderer-queue`)
 const telemetry = require(`gatsby-telemetry`)
 
-module.exports = async (program: any, activity: any) => {
+const runWebpack = compilerConfig =>
+  new Promise((resolve, reject) => {
+    webpack(compilerConfig).run((e, stats) => {
```
wardpeet: I like explicit variables like err or error over e
```suggestion
    webpack(compilerConfig).run((err, stats) => {
```

packages/gatsby/src/commands/build-html.js
```diff
@@ -1,69 +1,76 @@
...
+        )
+    throw error
+  }
+  return outputFile
+}
+
+const buildRenderer = async (program, stage) => {
+  const { directory } = program
+  const config = await webpackConfig(program, directory, stage, null)
+  return await doBuildRenderer(program, config)
```
wardpeet: no need to await here as you await in buildPages
```suggestion
  return doBuildRenderer(program, config)
```

Moocar: In general, I like to explicitly `await` on async functions, since it's an extra hint to the reader that `doBuildRenderer` is an async function. As you say, it's very much redundant, but the positive is a little more clarity in the code. I don't feel strongly either way though, so happy to change it.

DSchau: I think both are fine. I write my code like @wardpeet does here--but I also see your point in clarifying intent, which I kinda like!

I wouldn't necessarily feel like you have to take this suggestion--but thanks for being receptive to it!

packages/gatsby/src/commands/build-html.js
Moocar: Addressed in 0ef8272

packages/gatsby/src/commands/build-html.js
wardpeet: just a suggestion, so no need to change :) thanks for responding

DSchau:
Successfully published:
 - gatsby@2.3.11

PR #13007, opened 2019-04-01
refactor(gatsby): expicitly call db saveState and autoSave

Moocar:
## Description

As part of my work on https://github.com/gatsbyjs/gatsby/pull/13004, I've been trying to remove reliance on global events such as `BOOTSTRAP_FINISHED`. There are definitely some good use cases for global events, but I feel in this case, the code is more clear by specifically invoking `db.saveState/autoSave`. Curious on other's thoughts.

## Related Issues

### DIFF
--- a/packages/gatsby/src/db/index.js
+++ b/packages/gatsby/src/db/index.js
@@ -27,33 +27,14 @@ async function saveState() {
 const saveStateDebounced = _.debounce(saveState, 1000)
 
 /**
- * Sets up listeners so that once bootstrap has finished, all
- * databases save their state to disk. If we're in `develop` mode,
- * then any new event triggers a debounced save as well.
+ * Starts listening to redux actions and triggers a database save to
+ * disk upon any action (debounced to every 1 second)
  */
 function startAutosave() {
-  // During development, once bootstrap is finished, persist state on changes.
-  let bootstrapFinished = false
-  if (process.env.gatsby_executing_command === `develop`) {
-    emitter.on(`BOOTSTRAP_FINISHED`, () => {
-      bootstrapFinished = true
-      saveState()
-    })
-    emitter.on(`*`, () => {
-      if (bootstrapFinished) {
-        saveStateDebounced()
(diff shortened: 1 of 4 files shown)


Moocar:
> Just wondering why we're listening on emitter and not on the redux state changing itself

Do you mean by registering a reducer that reacts to any redux action? My guess is that we would have no guarantee that all other reducers had executed before performing the save. Whereas `emitter` is emitted via a [subscription to redux](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/redux/index.js#L99), which is guaranteed to fire after all reducers have finished.

pieh:
I think we need to remove `startAutosave` call from `bootstrap` ( https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/bootstrap/index.js#L26 ) with this?

packages/gatsby/src/db/index.js
```diff
@@ -27,33 +27,14 @@ async function saveState() {
...
-    })
-  }
-
-  // During builds, persist state once bootstrap has finished.
-  if (process.env.gatsby_executing_command === `build`) {
-    emitter.on(`BOOTSTRAP_FINISHED`, () => {
-      saveState()
-    })
-  }
+  emitter.on(`*`, () => saveStateDebounced())
```
pieh: should `startAutosave` also call `saveState` initially? Right now in develop we would start autosaving, but we potentially would never actually save if no action is emitted?

Moocar:
> I think we need to remove `startAutosave` call from `bootstrap` with this?

Good catch. Fixed in 5028776

packages/gatsby/src/db/index.js
Moocar: Another good catch. Fixed in f6a254e

Moocar:
@pieh conflicts resolved

sidharthachatterjee:
Published in gatsby@2.3.14

PR #13008, opened 2019-04-02
refactor(gatsby): move redirects-writer to bootstrap

Moocar:
## Description

In a future sub-PR of https://github.com/gatsbyjs/gatsby/pull/13004, I'll be moving the `query-runner` from `src/internal-plugins/query-runner` to `src/query`. I was surprised to find that `redirects-writer` was part of `query-runner`. I think it makes more sense to be under `bootstrap` so have moved it.

## Related Issues

### DIFF
--- a/packages/gatsby/src/bootstrap/index.js
+++ b/packages/gatsby/src/bootstrap/index.js
@@ -38,9 +38,7 @@ const {
 } = require(`../internal-plugins/query-runner/page-query-runner`)
 const queryQueue = require(`../internal-plugins/query-runner/query-queue`)
 const { writePages } = require(`../internal-plugins/query-runner/pages-writer`)
-const {
-  writeRedirects,
-} = require(`../internal-plugins/query-runner/redirects-writer`)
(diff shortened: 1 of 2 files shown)


wardpeet:
It's probably not worth it to tangle this in your PRs but would it be better to get the `bootstrapFinished` local variable out of redux store?

convert
https://github.com/gatsbyjs/gatsby/blob/61d7e5dc4e38f829b047cd61db0562d316846569/packages/gatsby/src/bootstrap/redirects-writer.js#L41
into 
```js
store.getState().program.status === 'BOOTSTRAP_FINISHED'
```

Moocar:
> would it be better to get the bootstrapFinished local variable out of redux store?

@wardpeet Absolutely, but I'd probably go a step further and explicitly start the debounced writer from `develop.js`. Like I just did for `db.startAutosave()`. I'll add a TODO to me list.

Moocar:
to be clear, I'll refactor `redirects-writer` in a future PR. So this one should be good to merge.

PR #13010, opened 2019-04-02
fix(link): URI encode the to field

me4502:
## Description

Currently if you use a Link tag pointing at a non-URI encoded URL, it won't show as active on the page. This fixes it by URI encoding the 'to' passed into the Link.

### DIFF
--- a/packages/gatsby-link/src/__tests__/index.js
+++ b/packages/gatsby-link/src/__tests__/index.js
@@ -118,6 +118,12 @@ describe(`<Link />`, () => {
       setup({ linkProps: { to } })
       expect(console.warn).toBeCalled()
     })
+
+    it(`supports non-encoded URIs`, () => {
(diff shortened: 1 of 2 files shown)


me4502:
So for some reason I hadn't actually thought of just encoding it in my component.

However I still feel this should be inside Gatsby itself. My main motivation is that it brings the href that the Link component checks against, inline with what the browser sees the href as.

DSchau:
@me4502 going to close this out.

I _do_ think it's probably worth building this in at some point, but as it stands now the potential for breaking changes doesn't seem worth the benefit of the feature (since it can be easily baked into a custom component).

However -- I do think this is something to consider for the future, so let's keep this branch and idea around when we're ready to revisit.

As always--thanks for your work on this feature. We appreciate it more than you know!

PR #13012, opened 2019-04-02
feat(gatsby): write match-paths.json

Moocar:
## Description

Writes `match-paths.json` during pages writing. This file contains a map of `matchPath` -> `path`. It will be used by https://github.com/gatsbyjs/gatsby/pull/13004 when finding pages in the production-app. See https://github.com/gatsbyjs/gatsby/pull/13004/files#diff-06ce5193de7c2c01dd947a5001c147c3R60 for how it's used.

## Related Issues


packages/gatsby/src/redux/reducers/match-paths.js
```diff
@@ -0,0 +1,32 @@
+const _ = require(`lodash`)
+
+module.exports = (state = {}, action) => {
+  switch (action.type) {
+    case `DELETE_CACHE`:
+      return {}
+    case `CREATE_PAGE`: {
+      const page = action.payload
+      if (page.matchPath) {
+        const { oldPage } = action
```
wardpeet: I don't think oldPage is set on the action.

pieh:
In `pages-writer.js` we do iterate over `pages` redux substate anyway, so it might be easier to generate derivative `matchPaths` from it? So we don't have to have 2 reducers reacting to `CREATE_PAGE`/`DELETE_PAGE`?

I.e:
```diff
-  let { program, jsonDataPaths, pages, matchPaths } = store.getState()
+  let { program, jsonDataPaths, pages } = store.getState()
  pages = [...pages.values()]
  const matchPaths = {}

  const pagesComponentDependencies = {}

  // Write out pages.json
  let pagesData = []
  pages.forEach(({ path, matchPath, componentChunkName, jsonName }) => {
    const pageComponentsChunkNames = {
      componentChunkName,
    }

    if (program._[0] === `develop`) {
      pagesComponentDependencies[path] = pageComponentsChunkNames
    }

    pagesData.push({
      ...pageComponentsChunkNames,
      jsonName,
      path,
      matchPath,
    })
+
+   if (matchPath) {
+     matchPaths[matchPath] = path
+   }
  })
```

Moocar:
Apologies, this was a sloppy PR. The `match-paths` reducer was necessary in incremental build land, which this work originated from. But I never went back and de-incrementalized it. @pieh I like your suggestion. It's much clearer. I'll implement that now

Moocar:
Fixed in 966f035

PR #13015, opened 2019-04-02
remove unnecessary eslint rule

KyleAMathews:

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -48,7 +48,6 @@
     "quotes": ["error", "backtick"],
     "react/display-name": "off",
     "react/jsx-key": "warn",
-    "react/no-unescaped-entities": "warn",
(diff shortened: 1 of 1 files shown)


polarathene:
Hi @KyleAMathews , looks like this accidentally changed the warnings into errors in the CI. Doesn't seem like you waited for the tests to complete before merging?

I've brought the rule back in the referencing PR above, but disabled it.

KyleAMathews:
Ooops, sorry — was feeling frustrated at our lint setup — we have way too many presets and rules which makes the whole thing feel complex and unhelpful.

Lemme PR turning it off which is what I guess I had to do.

KyleAMathews:
https://github.com/gatsbyjs/gatsby/pull/13021 —> I'll wait for the tests to finish this time :-D

PR #13017, opened 2019-04-02
Fix Misuse of Commas, Et Al

dhurlburtusa:
Use em dash in place of two hyphens.
Format paragraph to be on one line (apparently the convention).


DSchau:
Same idea here.

Let's encapsulate these changes in #13031. We love these types of PRs (so thank you!) but it's idea if we can group these related changes.

PR #13018, opened 2019-04-02
fix(CI): Get CI passing again after recent PRs slipped through the testing process causing failures

polarathene:
For some reason the test suite [passed for the PR](https://github.com/gatsbyjs/gatsby/pull/12994), Windows Azure Pipeline currently fails in PRs since(but CircleCI isn't picking up on it), however the PR passed as this part of the test suite doesn't appear in the PR test results?

This PR fixes the immediate issue failing other PRs with a `prettier --write` to get the corrections. ~~**It doesn't address the issue with CircleCI not picking up on this when it should**(stale cache?).~~ (Appears to be fixed now, at least with newer PRs throwing lint error as well)

Closes: #13011

PR #13021, opened 2019-04-02
Turn off instead of remove as one of our presets enables it

KyleAMathews:


KyleAMathews:
Oh, this is a duplicate of https://github.com/gatsbyjs/gatsby/pull/13018

PR #13006, opened 2019-04-01
refactor(gatsby): gatsby develop: cleanup port detection

Moocar:
## Description

Purely subjective refactor of port detection in `gatsby develop`. I think it's easier to read now.

## Related Issues

- Sub-PR of https://github.com/gatsbyjs/gatsby/pull/13004

PR #13024, opened 2019-04-02
feat(gatsby): add util.promisify polyfill for older node version

Moocar:
## Description

`util.promisify` was introduced in node v8, but gatsby still supports node v6. It's a useful utility when calling other libraries. Bluebird already supports [Promise.promisify](http://bluebirdjs.com/docs/api/promise.promisify.html), but not everyone is aware of it, and newcomers to gatsby will likely reach for `util.promisify` instead, since it's in the official node documentation.

This PR adds a polyfill for `util.promisify`.

## Related Issues

- enables https://github.com/gatsbyjs/gatsby/pull/13006

### DIFF
--- a/packages/gatsby/package.json
+++ b/packages/gatsby/package.json
@@ -124,6 +124,7 @@
     "true-case-path": "^1.0.3",
     "type-of": "^2.0.1",
     "url-loader": "^1.0.1",
+    "util.promisify": "^1.0.0",
     "uuid": "^3.1.0",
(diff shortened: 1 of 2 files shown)


Moocar:
cc @wardpeet

wardpeet:
I'm a fan of this 🎉

sidharthachatterjee:
Published in gatsby@2.3.9

PR #13025, opened 2019-04-02
feat: Use Loki's elemMatch operator

stefanprobst:
Loki currently falls back on Sift for elemMatch queries. This PR uses Loki's native elemMatch operator that was recently added. Unfortunately it is not clear when a new Loki release will happen, so use a scoped package in the meantime.


freiksenet:
Could you backmerge and resolve yarn conflict (running yarn should do it)?

stefanprobst:
Done!

sidharthachatterjee:
Published in gatsby@2.3.9

Issue #11604, opened 2019-02-06
onInitialClientRender not called on IE10


gluemonkey:
## Description

onInitialClientRender is never called and initial client render does not happen in IE10

### Steps to reproduce

View a basic starter build in IE10 and notice that render is never called within your components.

### Expected result

onInitialClientRender should be called inside of gatsby-browser.js and your React components should render.

### Actual result

onInitialClientRender is not called inside of gatsby-browser.js and your React components do not render.  Pre-rendered version still loads.  Upon navigating and hitting the back button the intended functionality will occur.

### Environment

Standard gatsbyjs build.

### Workaround

Using yarn I am able to override the domready version and this seems to fix the issue
```
"resolutions": {
    "domready": "1.0.7"
  }
```

Please refer to this issue - https://github.com/ded/domready/issues/53

jonniebigodes:
@gluemonkey i've picked up on your issue. And below are the steps i've taken and the results achieved.

- Created a new Gatsby website using the default starter through `gatsby new test_quick_start`
- Updated `gatsby-browser.js` to the following to check if i could get it to trigger:

```javascript
exports.onInitialClientRender=()=>{
    console.log(`onInitialClientRender was fired`);
}
```
- Issued `gatsby develop` and waited to the process to complete.
- Opened edge (as i don't have access to ie10 currently) and the devtools, switched emulation to the following:
![emulation](https://user-images.githubusercontent.com/22988955/52501345-03e85980-2bd8-11e9-84e8-1fdca0ee2e64.png)
- With the emulation activated the site was re-rendered and switching to the console tab i'm presented with this:
![console](https://user-images.githubusercontent.com/22988955/52501390-20849180-2bd8-11e9-9202-eb9e1f98980c.png)

As you can see the site is being rendered and browser api is triggered, No issues whatsoever from my side. It could be that you have something else happening in the api that could be causing the issue.
If the work you're doing is not private or otherwise protected, if you can share a repository with the code it would be great so that it could be further investigated. Otherwise, create a dummy repository with the code that triggers the issue you've experienced.

Feel free to provide further feedback so that this issue can be solved.

sidharthachatterjee:
Based on https://github.com/ded/domready/issues/53 it does seem that IE10 support is broken by domready after 1.0.7 because of some regex matching code that was added

Let me try reproducing this and I'll post what I can find here

franksmule:
Same issue although for me - although 1.0.7 does not work either.
[...]

PR #13027, opened 2019-04-02
chore(gatsby): fix domready package to be compatible with ie10

mikaelkristiansson:
## Description

When a gatsby site is loaded in internet explorer 10 it will not render.
Since gatsby use package domready (not maintained since 4 years back) I decided to fork it and fix the problem.

this is how gatsby call the function:
```javascript
domReady(() => {
      renderer(
        <NewRoot />,
        typeof window !== `undefined`
          ? document.getElementById(`___gatsby`)
          : void 0,
        () => {
          postInitialRenderWork()
          apiRunner(`onInitialClientRender`)
        }
      )
    })
```

but since domready isn't working correct in ie10 the renderer function won't be called.

The problem causing this in domready package was it calls
```javascript
document.documentElement.doScroll
```
in ie10 this returns as a function, so I had to change it to call the function to get the value expected:
```javascript
ie10 ? document.documentElement.doScroll() : document.documentElement.doScroll
```

link to package and git repo:
https://www.npmjs.com/package/domready-loaded
https://github.com/mikaelkristiansson/domready

## Related Issues

Fixes this issue: #11604

### DIFF
--- a/packages/gatsby/cache-dir/app.js
+++ b/packages/gatsby/cache-dir/app.js
@@ -1,6 +1,6 @@
 import React from "react"
 import ReactDOM from "react-dom"
-import domReady from "domready"
+import domReady from "domready-loaded"
 
 import socketIo from "./socketIo"
 import emitter from "./emitter"
--- a/packages/gatsby/cache-dir/app.js
+++ b/packages/gatsby/cache-dir/app.js
@@ -1,6 +1,6 @@
 import React from "react"
 import ReactDOM from "react-dom"
-import domReady from "domready-loaded"
+import domReady from "@mikaelkristiansson/domready"
 
 import socketIo from "./socketIo"
 import emitter from "./emitter"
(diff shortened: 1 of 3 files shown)


wardpeet:
why wouldn't you publish it on your own name `@mikaelkristiansson/domready`? Also you should create a PR for this on his repo so we can switch back to domready later on.

Another approach is just doing it ourselves XD For ie9+ we don't need any doScroll hacks.
```
function domReady(cb => {
  if (document.readyState === 'complete') {
    return cb()
  }

  document.addEventListener('DOMContentLoaded', function listener() {
    document.removeEventListener('DOMContentLoaded', listener);
    cb()
  })
});
```

mikaelkristiansson:
I've created a PR to the original repo:
https://github.com/ded/domready/pull/59
and also updated package to use @mikaelkristiansson/domready

wardpeet:
thanks! Let me also explain why `@mikaelkristiansson/domready` is valuable for us instead of `domready-loaded`. This mostly indicated that this is a temp package derived from domready. With domready-loaded we have to open your github to see that it's a fork, ...

PR #13041, opened 2019-04-02
Add ignore statement

strets123:
## Description

A number of previous issues reference the error:

`
ReferenceError: exports is not defined in api-runner-browser.js
`
If no custom version of gatsby-browser.js has been created then this error still occurs whenever a simple `.babelrc` file is used. I have added the ignore statement to the docs to reflect this so that others do not need to sepnd time searching for this solution or implement a custom api-runner-browser.js.

## Related Issues
https://github.com/gatsbyjs/gatsby/issues/6958

### DIFF
--- a/docs/docs/babel.md
+++ b/docs/docs/babel.md
@@ -23,6 +23,7 @@ browsers.
 
 Gatsby ships with a default .babelrc setup that should work for most sites. If you'd like
 to add custom Babel presets or plugins, you can create your own `.babelrc` at the root of your site, import [`babel-preset-gatsby`](https://github.com/gatsbyjs/gatsby/tree/master/packages/babel-preset-gatsby), and overwrite the `targets` option.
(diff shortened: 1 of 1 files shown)


pieh:
This doesn't seem right. Why should this file be ignored? If it doesn't work for some reason, this need to be fixed and not ignored

strets123:
Fair enough, to work on a fix I have a question. The docs made it sound like there is a default Babel config and that it is equivalent to putting the Gatsby preset in .babelrc it seems that this is not the case. If not, what does is the default config applied to? Only ./src?

sidharthachatterjee:
@strets123 Yes, the default is applied in `packages/gatsby/src/utils/babel-loader-helpers.js` and `packages/gatsby/src/utils/babel-loader.js`

Closing this in the mean time! 

Thank you so much and looking forward to your first contribution @strets123 👍

Issue #10398, opened 2018-12-11
Suggest people install Node.js w/ Homebrew (on macs)?


KyleAMathews:
Something we've heard quite a few times is that installing Node.js on macs can be difficult for people new to the Node world e.g. there's frequently problems with permissions which can be very confusing and a major roadblock.

I always use Homebrew to install dev tools which takes care of things like this.

What do people think about suggesting people use Homebrew for installing Node on Macs for Gatsby?

/cc @gatsbyjs/docs @gatsbyjs/core

sidharthachatterjee:
I prefer https://github.com/creationix/nvm because

- works on Linux and macOS (so might be a better and more general recommendation)
- the ease of switching between versions etc 

However, Homebrew is much simpler and probably a better recommendation for someone new to Node and if Mac only

m-allanson:
Both homebrew and nvm require XCode's Command Line Tools which adds another step to the install. But that is probably simpler than dealing with permissions issues after using the official node installer.

I've done this in the past:

- Install XCode CLI tools
- Install node installer installer (homebrew)
- Install node installer (nvm)
- Install node

Which is great once it's done but does seem a bit silly. You could easily drop one of homebrew or nvm just for getting started.

jonniebigodes:
@KyleAMathews and @m-allanson if you don't mind my two cents, but adding homebrew and Xcode's Command Line Tools, and with that adding another step to the install process, might look like somewhat of a hassle beforehand, but further down the line it will avoid some issues inbound to here. As much as gatsby tends to be a "turn key" solution, as you add it and you're quasi good to go to start hammering the keyboard and pushing out apps, content or what the person is going to use gatsby for.
Without this preventive measure, let's call it that, people will start flooding the issues, asking about what's happening and how to solve it and what is this error. I'm saying this based on recent issues that have spawned here. But again this is just my two cents

shannonbux:
From about 4 usability tests with Mac users and the tutorial, Macs prompt
users to install Xcode when they download VSCode anyway. So there might not
be a way for Mac users to avoid installing Xcode, though being prompted by
their operating system and clicking an "ok" button is a little easier than
installing it themselves.

jonathanprozzi:
[...]

PR #13051, opened 2019-04-02
docs: Install Node with Homebrew

marcysutton:
## Description

This PR is a start to documenting Homebrew & Node install instructions as discussed in #10398. 

One thing I considered but didn't commit was a callout for more specific instructions around global modules, such as mentioned in https://github.com/gatsbyjs/gatsby/issues/10398#issuecomment-466451396.

> To troubleshoot issues with globally installed packages, visit the [extended Node.js docs](/docs/troubleshooting-global-node/).

If we feel an extended doc is necessary, I can work on adding it...but I intentionally kept this simple for now. Feedback is welcome!


docs/tutorial/part-zero/index.md
```diff
@@ -25,15 +25,34 @@ Take a moment to locate and open up the command line interface (CLI) for your co
...
 
-## Install Node.js
+## Install Homebrew for Node.js
+
+To install Gatsby and Node.js, it is recommended to use [Homebrew](https://brew.sh/). A little set-up in the beginning can save you from some headaches later on!
+
+How to install or verify Homebrew on your computer:
+
+1. Open your Terminal
+1. See if Homebrew is installed by running `brew -v`. You should see "Homebrew" and a version number, if so.
```
wardpeet: the `if so` reads a bit strange in my opinion, it looks like something is missing.
Wouldn't it be better to just omit that?

marcysutton: Makes sense to me, I've updated the PR!

KyleAMathews:
This PR has 584 files changed

marcysutton:
Weird, a botched merge from master. I'll try to fix it...

marcysutton:
This PR is fixed up now! I wonder if we should add anything about what to do if you already have Node installed, though. Do we want to add anything about upgrading or uninstalling Node that was set up in other ways?

wardpeet:
I think it is good enough as is, unsure if many people have node installed without knowing 😛

marcysutton:
Ok! I'm going to merge this since it was approved and it doesn't have 584 files changed anymore :)

PR #13053, opened 2019-04-02
fix(gatsby): don't try to statically analyze gatsby-node, require it and examine exports instead

pieh:
https://github.com/gatsbyjs/gatsby/pull/12263 added API filtering in `api-runner-node`. This exposed some issues in valid, but pretty unusual cases when babel transpile source code to something like:
```js
"use strict";

Object.defineProperty(exports, "__esModule", {
  value: true
});
Object.defineProperty(exports, "onPreBootstrap", {
  enumerable: true,
  get: function get() {
    return _plugin.onPreBootstrap;
  }
});
Object.defineProperty(exports, "onCreateNode", {
  enumerable: true,
  get: function get() {
    return _plugin.onCreateNode;
  }
});
```

This is oddly looking, but is valid javascript and we should support that. Making it work with static analisys might be challenging, but for `gatsby-node` APIs we can just require `gatsby-node` file and check export keys. We can't really do that for `gatsby-ssr` and `gatsby-browser` because it can contain code that will just crash in Node.js context (i.e. JSX).

### DIFF
--- a/packages/gatsby/src/bootstrap/resolve-module-exports.js
+++ b/packages/gatsby/src/bootstrap/resolve-module-exports.js
@@ -120,7 +120,7 @@ https://gatsby.dev/no-mixed-modules
  * @param {string} mode
  * @param {function} resolver
  */
-module.exports = (modulePath, { mode = `analisys`, resolver } = {}) => {
+module.exports = (modulePath, { mode = `analysis`, resolver } = {}) => {
   if (mode === `require`) {
     try {
       return Object.keys(require(modulePath))
--- a/packages/gatsby/src/bootstrap/resolve-module-exports.js
+++ b/packages/gatsby/src/bootstrap/resolve-module-exports.js
@@ -6,16 +6,7 @@ const { codeFrameColumns } = require(`@babel/code-frame`)
 const { babelParseToAst } = require(`../utils/babel-parse-to-ast`)
 const report = require(`gatsby-cli/lib/reporter`)
 
-/**
- * Given a `require.resolve()` compatible path pointing to a JS module,
- * return an array listing the names of the module's exports.
- *
- * Returns [] for invalid paths and modules without exports.
- *
- * @param {string} modulePath
- * @param {function} resolver
- */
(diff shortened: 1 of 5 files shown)


packages/gatsby/src/bootstrap/resolve-module-exports.js
```diff
@@ -118,3 +109,25 @@ https://gatsby.dev/no-mixed-modules
...
+ * Given a `require.resolve()` compatible path pointing to a JS module,
+ * return an array listing the names of the module's exports.
+ *
+ * Returns [] for invalid paths and modules without exports.
+ *
+ * @param {string} modulePath
+ * @param {string} mode
+ * @param {function} resolver
+ */
+module.exports = (modulePath, { mode = `analisys`, resolver } = {}) => {
```
DSchau:
```suggestion
module.exports = (modulePath, { mode = `analysis`, resolver } = {}) => {
```

DSchau: This doesn't _really_ matter here--but for correctness 🤷‍♂️

DSchau:
Validated this works e2e with:

```js
module.exports = {
  createPages() {}
}
```

```js
exports.createPages = () => {}
```

and

```js
Object.defineProperty(exports, 'createPages', {
  enumberable: true,
  get() {}
})
```

and all seem to work pretty flawlessly. There are probably things I missed--but I think moving to Node's require isolates us more effectively than parsing the exports ourselves.

packages/gatsby/src/bootstrap/resolve-module-exports.js
pieh: lol 🙈 I want autocorrect in vs code 😂

Moocar:
I ran into this issue as well. My local site's `gatsby-node.js` wasn't being picked up. Ran it with this branch and issue is fixed. 👍

DevSide:
This breaks gatsby-node using require('esm'). Any ideas ?

PR #13056, opened 2019-04-02
chore(docs): Marking stubs in "Using Gatsby Professionally"

lannonbr:
## Description

A series of docs in the "Using Gatsby Professionally" section are still to be fully fleshed out as they have the stub copy text at the bottom of the pages. With such, I added the asterisk to any of these pages that didn't already have it so it will be displayed on the stub list as well as partially greyed out in the docs sidebar.


DSchau:
Ugh - some merge conflicts that I missed. @lannonbr would you mind fixing up?

Otherwise I'm happy to do so as well.

lannonbr:
Should be good to go once CI is done

PR #13058, opened 2019-04-03
fix(gatsby-transformer-documentationjs): handle free floating jsdocs

### LINKED ISSUE #13052: gatsby-transformer-documentationjs creates non-unique ids
`gatsby-transformer-documentationjs` creates nodes that can have non-unique ids, for example for comments like these:

```js
/**
 * This is a free-floating comment
 */

/**
 * This is a function
 */
const fn = () => {}
```

When trying to build such a site with `GATSBY_DB_NODES=loki` this will throw:
```
error Error processing Markdown in node 27c286f3-9f32-5454-9d77-a9265e600a7d:
Trying to update unsynced document. Please save the document first by using insert() or addMany()
```

See repo: https://github.com/stefanprobst/gatsby-bug-docjs

pieh:
This is not great fix, but it is fix for issue described in #13052 . This is happening because `documention` package creates 2 jsdocs for same object. I'm not sure if "free floating" jsdoc is valid or not (in scenarios other than defining type definition / callback), but because this is problematic quick fix is to not create multiple nodes with same `id`

Fixes #13052

### DIFF
--- a/packages/gatsby-transformer-documentationjs/src/__tests__/gatsby-node.js
+++ b/packages/gatsby-transformer-documentationjs/src/__tests__/gatsby-node.js
@@ -117,8 +117,9 @@ describe(`transformer-react-doc-gen: onCreateNode`, () => {
     })
 
     it(`doesn't create multiple nodes with same id`, () => {
-      Object.values(groupBy(createdNodes, `id`)).forEach(nodes =>
-        expect(nodes.length).toBe(1)
+      const groupedById = groupBy(createdNodes, `id`)
+      Object.keys(groupedById).forEach(id =>
+        expect(groupedById[id].length).toBe(1)
       )
     })
   })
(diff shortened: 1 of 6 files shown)


packages/gatsby-transformer-documentationjs/src/__tests__/gatsby-node.js
```diff
@@ -113,9 +115,19 @@ describe(`transformer-react-doc-gen: onCreateNode`, () => {
         })
       )
     })
+
+    it(`doesn't create multiple nodes with same id`, () => {
+      Object.values(groupBy(createdNodes, `id`)).forEach(nodes =>
```
pieh: keep forgetting `Object.values` is not supported in node 6 - will fix

stefanprobst:
Confirmed that www builds with Loki with this.

PR #13059, opened 2019-04-03
fix(gatsby-plugin-manifest): allow multiple icon paths

### LINKED ISSUE #13055: [gatsby-plugin-manifest] Buggy output (manifest file and images)
## Description

1) gatsby-plugin-manifest fails to create images when paths are different.
2) gatsby-plugin-manifest does not delete the `include_favicon` option when creating the manifest file.

### Steps to reproduce

```
# gatsby-config.js
module.exports = {
  ...
  plugins: [
    ...
    {
      options: {
        name: `foo`,
        short_name: `bar`,
        start_url: `/?utm_source=homescreen`,
        background_color: `#111111`,
        display: `minimal-ui`,
        icon: `src/images/gatsby-icon.png`,
        include_favicon: false,
        legacy: false,
        theme_color_in_head: false,
        icons: [
          {
            src: `favicon.ico`,
            sizes: `32x32`,
            type: `image/ico`,
          },
          {
            src: `icons/icon-192x192.png`,
            sizes: `192x192`,
            type: `image/png`,
          },
        ],
      },
    },
    ...
  ],
}
```

```
$ rm -rf public && gatsby build
...
Generating image thumbnails [==============================] 6/6 0.3 secs 100%
error Plugin gatsby-plugin-manifest returned an error

  Error: vips__file_open_write: unable to open file "public/icons/icon-192x192.png" for writing
  unix error: No such file or directory
```

This happens because of https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-manifest/src/gatsby-node.js#L72-L73

For the `include_favicon` issue, this is because that option is not listed at https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-manifest/src/gatsby-node.js#L46-L55

### Expected result

1) The plugin should create the images.
2) The option should not be present in the final output.

### Actual result

See description.

### Environment
  System:
    OS: macOS 10.14.2
    Shell: 3.0.2 - /usr/local/bin/fish
  Binaries:
    Node: 10.15.3 - ~/.asdf/installs/nodejs/10.15.3/bin/node
  npmPackages:
    gatsby: ^2.3.3 => 2.3.3
    gatsby-image: ^2.0.35 => 2.0.35
    gatsby-plugin-manifest: ^2.0.26 => 2.0.26
    gatsby-plugin-offline: ^2.0.25 => 2.0.25
    gatsby-plugin-postcss: ^2.0.7 => 2.0.7
    gatsby-plugin-react-helmet: ^3.0.11 => 3.0.11
    gatsby-plugin-sharp: ^2.0.32 => 2.0.32
    gatsby-source-filesystem: ^2.0.28 => 2.0.28
    gatsby-transformer-sharp: ^2.1.17 => 2.1.17
  npmGlobalPackages:
    gatsby-cli: 2.5.4

me4502:
## Description

gatsby-plugin-manifest fails to create images when paths are different.
~~gatsby-plugin-manifest does not delete the include_favicon option when creating the manifest file.~~

## Related Issues

Fixes #13055


packages/gatsby-plugin-manifest/src/gatsby-node.js
```diff
@@ -74,11 +74,13 @@ exports.onPostBootstrap = async ({ reporter }, pluginOptions) => {
   }
 
   // Determine destination path for icons.
-  const iconPath = path.join(`public`, path.dirname(manifest.icons[0].src))
+  for (let i = 0; i < manifest.icons.length; i++) {
```
DSchau:
```suggestion
  const uniqueIconPaths = manifest.icons.reduce((merged, icon) => {
    return merged.add(path.join(`public`, path.dirname(icon.src))
  }, new Set())
  
  uniqueIconPaths.forEach(iconPath => {
    if (!fs.existsSync(iconPath) {
      fs.mkdirSync(iconPath)
    }
  })
```

me4502:
This used to fix include_favicon, however it was fixed in master so my fix got removed when I rebased.

packages/gatsby-plugin-manifest/src/__tests__/gatsby-node.js
```diff
@@ -105,6 +108,45 @@ describe(`Test plugin manifest options`, () => {
...
+    const firstIconPath = path.join(
+      `public`,
+      path.dirname(`icons/icon-48x48.png`)
+    )
+    const secondIconPath = path.join(
+      `public`,
+      path.dirname(`other-icons/icon-48x48.png`)
+    )
+
+    const calls = fs.mkdirSync.mock.calls
```
DSchau:
```suggestion
    expect(fs.mkdirSync).toHaveBeenNthCalledWith(1, firstIconPath)
    expect(fs.mkdirSync).toHaveBeenNthCalledWith(2, secondIconPath)
```

I haven't tested this--but just a suggestion!

me4502: I had it like that originally, however I modified it to use calls to be inline with the other tests in the file. I can change it back if you'd like

moonmeister: @me4502 My 2 cents: calls are probably the less clean way to do this (I wrote some of these tests...I just didn't no this other option existed). This seems much cleaner and robust. Do as you wish imo. Another PR to clean up all the tests is welcome.

me4502: If it'd make more sense to clean up all tests in a single PR, I can do that instead.

DSchau: It's a fair response to continue with what was _in_ the test/file to begin wit though--so don't want to dissuade from that behavior.

Let's revisit this in a future PR! For now: 🚀

moonmeister: :man_shrugging: I have no preference. See if @DSchau does. My guess is he won't.

me4502:
Thanks, I'll write another PR to cleanup the tests shortly.

PR #13060, opened 2019-04-03
feat(www): don't check for overview frontmatter

pieh:
Fixes `[[guidelist]]` not being replaced in few pages (i.e. in https://www.gatsbyjs.org/docs/winning-over-stakeholders/ )

The code is handling cases when there are no children (tried building locally and it worked just fine).

Follow up to this would be to clear `.md` files that use `overview` and adjust page query. But this is just quick fix.

### DIFF
--- a/www/src/templates/template-docs-markdown.js
+++ b/www/src/templates/template-docs-markdown.js
@@ -32,10 +32,6 @@ const childItemsBySlug = (docsHierarchy, slug) => {
 }
 
 const getPageHTML = page => {
-  if (!page.frontmatter.overview) {
-    return page.html
-  }
-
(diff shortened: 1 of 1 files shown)


jlengstorf:
Does it still show an empty overview if no children are present? If not, then _I guess_ you can touch my stuff. 😅

pieh:
> Does it still show an empty overview if no children are present? If not, then _I guess_ you can touch my stuff. 😅

If the list is empty it _should_ replace `[[guidelist]]` with empty string. Let me try to verify this

pieh:
Yup, it shows nothing if there are no children

PR #13063, opened 2019-04-03
refactor(gatsby): Refactor query calc query ids

Moocar:
## Description

A purely cosmetic refactor. I found the logic in `page-query-runner/runQueries` difficult to get my head around. Hopefully it's a little better now but I'm sure it can still be improved.

## Related Issues

- Sub-PR of https://github.com/gatsbyjs/gatsby/pull/13004
- builds on https://github.com/gatsbyjs/gatsby/pull/13061

PR #13064, opened 2019-04-03
refactor(gatsby): Split static/page queries

Moocar:
## Description

For https://github.com/gatsbyjs/gatsby/pull/13004, we need to run page queries after build-javascript has executed. This PR introduces the ability to split queries into static/page and run them explicitly.

## Related Issues

- builds on https://github.com/gatsbyjs/gatsby/pull/13063

### DIFF
--- a/packages/gatsby/src/bootstrap/index.js
+++ b/packages/gatsby/src/bootstrap/index.js
@@ -461,7 +461,10 @@ module.exports = async (args: BootstrapArgs) => {
     parentSpan: bootstrapSpan,
   })
   activity.start()
-  await pageQueryRunner.processStaticQueries(staticQueryIds, { activity, state: store.getState() })
+  await pageQueryRunner.processStaticQueries(staticQueryIds, {
+    activity,
+    state: store.getState(),
+  })
   activity.end()
 
   activity = report.activityTimer(`run page queries`)
--- a/packages/gatsby/src/bootstrap/index.js
+++ b/packages/gatsby/src/bootstrap/index.js
@@ -31,7 +31,7 @@ process.on(`unhandledRejection`, (reason, p) => {
 })
 
 const { extractQueries } = require(`../query/query-watcher`)
-const { runInitialQueries } = require(`../query/page-query-runner`)
+const pageQueryRunner = require(`../query/page-query-runner`)
 const { writePages } = require(`../query/pages-writer`)
 const { writeRedirects } = require(`./redirects-writer`)
 
@@ -452,12 +452,21 @@ module.exports = async (args: BootstrapArgs) => {
     require(`./page-hot-reloader`)(graphqlRunner)
   }
 
-  // Run queries
-  activity = report.activityTimer(`run graphql queries`, {
+  const queryIds = pageQueryRunner.calcInitialDirtyQueryIds(store.getState())
+  const { staticQueryIds, pageQueryIds } = pageQueryRunner.groupQueryIds(
+    queryIds
+  )
+
+  activity = report.activityTimer(`run static queries`, {
     parentSpan: bootstrapSpan,
   })
   activity.start()
-  await runInitialQueries(activity)
+  await pageQueryRunner.processStaticQueries(staticQueryIds, { activity })
+  activity.end()
+
+  activity = report.activityTimer(`run page queries`)
+  activity.start()
+  await pageQueryRunner.processPageQueries(pageQueryIds, { activity })
   activity.end()
 
   require(`../redux/actions`).boundActionCreators.setProgramStatus(
--- a/packages/gatsby/src/bootstrap/index.js
+++ b/packages/gatsby/src/bootstrap/index.js
@@ -461,7 +461,7 @@ module.exports = async (args: BootstrapArgs) => {
     parentSpan: bootstrapSpan,
   })
   activity.start()
-  await pageQueryRunner.processStaticQueries(staticQueryIds, { activity })
+  await pageQueryRunner.processStaticQueries(staticQueryIds, { activity, state: store.getState() })
   activity.end()
 
   activity = report.activityTimer(`run page queries`)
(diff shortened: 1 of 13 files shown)


Moocar:
I'm noticing that `page-query-runner.js` is becoming a bad name for that file. It's more acting as a one stop shop for anything query related. Except for stuff covered by watcher. I'm tempting to move this to `src/query/index.js`.

KyleAMathews:
Makes sense to me

wardpeet:
@Moocar could you rebase

Moocar:
@wardpeet Done, thanks!

packages/gatsby/src/bootstrap/index.js
```diff
@@ -452,12 +452,21 @@ module.exports = async (args: BootstrapArgs) => {
...
+  const { staticQueryIds, pageQueryIds } = pageQueryRunner.groupQueryIds(
+    queryIds
+  )
+
+  activity = report.activityTimer(`run static queries`, {
     parentSpan: bootstrapSpan,
   })
   activity.start()
-  await runInitialQueries(activity)
+  await pageQueryRunner.processStaticQueries(staticQueryIds, { activity })
```
wardpeet: What's the reason for not passing store.getState() in here?

```suggestion
  await pageQueryRunner.processStaticQueries(staticQueryIds, { activity, state: store.getState() })
```

packages/gatsby/src/query/page-query-runner.js
```diff
@@ -137,57 +136,18 @@ const calcInitialDirtyQueryIds = state => {
...
-
-const runInitialQueries = async activity => {
-  const pathnamesToRun = calcInitialDirtyQueryIds(store.getState())
-  if (pathnamesToRun.length === 0) {
-    return
+/**
+ * groups queryIds by whether they are static or page queries.
+ */
+const groupQueryIds = queryIds => {
+  const grouped = _.groupBy(queryIds, p => p.slice(0, 4) === `sq--`)
```
wardpeet: I know this is bikeshedding but this wasn't super easy to follow (might be me though). Wdyt of making it explicit?
```suggestion
  const grouped = _.groupBy(queryIds, p => p.slice(0, 4) === `sq--` ? 'static' : 'page')
  return {
    staticQueryIds: grouped['static'] || [],
    pageQueryIds: grouped['page'] || [],
  }
```

packages/gatsby/src/query/page-query-runner.js
```diff
@@ -197,9 +157,61 @@ const runInitialQueries = async activity => {
...
+const processStaticQueries = async (queryIds, { state, activity }) => {
+  state = state || store.getState()
+  await processQueries(
+    queryIds.map(id => makeStaticQueryJob(state, id)),
+    activity
+  )
+}
+
+const makePageQueryJob = (state, queryId) => {
+  const page = state.pages.get(queryId)
```
wardpeet: could this be null? there is an falsy check in the old code. Unsure if a query is being marked as dirty on deletion which might make this null.

packages/gatsby/src/bootstrap/index.js
Moocar: Mostly laziness :)

packages/gatsby/src/query/page-query-runner.js
Moocar: This is a great suggestion! It's much better than relying on `true`/`false`

packages/gatsby/src/query/page-query-runner.js
Moocar: Theoretically no. `runQueuedQueries` runs in a single code block. So there is no chance for any other code to run between when it calculates the query IDs, and when it creates the query jobs.

```js
const runQueuedQueries = () => {
  if (listenerQueue) {
    const state = store.getState()
    const { staticQueryIds, pageQueryIds } = groupQueryIds(
      calcDirtyQueryIds(state)
    )
    const queryJobs = [
      ...staticQueryIds.map(id => makeStaticQueryJob(state, id)),
      ...pageQueryIds.map(id => makePageQueryJob(state, id)),
    ]
    listenerQueue.push(queryJobs)
  }
}
```

wardpeet: Ah yeah 🤦‍♂️ 👍

packages/gatsby/src/query/page-query-runner.js
```diff
@@ -137,57 +136,20 @@ const calcInitialDirtyQueryIds = state => {
...
-    return
+/**
+ * groups queryIds by whether they are static or page queries.
+ */
+const groupQueryIds = queryIds => {
+  const grouped = _.groupBy(queryIds, p =>
+    p.slice(0, 4) === `sq--` ? `static` : `page`
+  )
+  return {
+    staticQueryIds: grouped[`static`] || [],
```
wardpeet: mybad, we can do dot notation
```suggestion
    staticQueryIds: grouped.static || [],
```

packages/gatsby/src/query/page-query-runner.js
```diff
@@ -137,57 +136,20 @@ const calcInitialDirtyQueryIds = state => {
...
+    pageQueryIds: grouped[`page`] || [],
```
wardpeet:
```suggestion
    pageQueryIds: grouped.page || [],
```

DSchau:
The integration test failure seems actually meaningful? May want to revisit that to ensure we're not introducing anything breaking here--may even need to tweak the integration test itself related to changes in this PR!

Moocar:
@DSchau Yep. It's legit. I'm looking into it.

Moocar:
The failing integration test was caused by not checking if a page exists before creating a query job for it. There's a case where we create a `SitePage` node for `/dev-404-page/`, but only create the corresponding `Page` during `gatsby develop`. 

A more permanent fix would be to not even create the `/dev-404-page/` SitePage during `gatsby build`, but schema depends on it so we'll have to leave it for another day.

@wardpeet you were totally right in your earlier comment about checking for nulls 🎩

Moocar:
Published in gatsby@2.3.24

Issue #12485, opened 2019-03-11
Building process crashes with message Illegal instruction


stjepangolemac:
## Description

When building large number of pages, building process crashes with message `Illegal instruction (core dumped)` on Linux or `Illegal instruction: 4` on Mac.

### Steps to reproduce

I can't provide any code here, but this happens when building between 4500-4750 pages in the affected project. Building 4500 pages works. This is a huge blocker for us as we need to build between 15000-25000 pages in total.

Previously we had problems with very slow query executions (1-2qps) but managed to dramatically improve that (300-400qps) by moving the queries from pages to `gatsby-node.js`.

During `gatsby build` of 4750 pages RAM usage tops out between 4-6 GB and stays there for some time. After that the process crashes.

I'll paste the command output below along with two files from the profiler.

### Expected result

Build all the pages.

### Actual result

Crashed.

### Environment

I've tried it locally on my Mac, and on AWS instances with up to 16 vcpus and 30 GB of RAM. This is the information from the Mac.

```
  System:
    OS: macOS 10.14.3
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 11.9.0 - ~/n/bin/node
    Yarn: 1.12.1 - /usr/local/bin/yarn
    npm: 6.5.0 - ~/n/bin/npm
  Languages:
    Python: 2.7.10 - /usr/bin/python
  Browsers:
    Chrome: 72.0.3626.121
    Firefox: 65.0
    Safari: 12.0.3
  npmPackages:
    gatsby: ^2.1.13 => 2.1.23 
    gatsby-image: ^2.0.29 => 2.0.31 
    gatsby-plugin-extract-schema: ^0.0.5 => 0.0.5 
    gatsby-plugin-manifest: ^2.0.20 => 2.0.22 
    gatsby-plugin-postcss: ^2.0.2 => 2.0.5 
    gatsby-plugin-react-helmet: ^3.0.2 => 3.0.8 
    gatsby-plugin-remove-trailing-slashes: ^2.0.7 => 2.0.9 
    gatsby-plugin-sass: ^2.0.7 => 2.0.10 
    gatsby-plugin-segment-js: ^3.0.1 => 3.0.1 
    gatsby-plugin-sharp: ^2.0.14 => 2.0.25 
    gatsby-plugin-sitemap: ^2.0.6 => 2.0.6 
[...]

PR #13066, opened 2019-04-03
feat(gatsby): add DANGEROUSLY_DISABLE_OOM

wardpeet:
## Description
This adds a process.env variable called `DANGEROUSLY_DISABLE_OOM` to disable saving the redux state to disk. This helps with going out of memory in the bootstrap phase. The downside is that your build will be much slower as it will always build from a clean dir.

We won't put this variable inside our docs as it's a feature we don't want to highlight or people to use.

## Related Issues
#12485


packages/gatsby/src/redux/index.js
```diff
@@ -73,6 +74,16 @@ const store = Redux.createStore(
 
 // Persist state.
 function saveState() {
+  if (process.env.DANGEROUSLY_DISABLE_OOM) {
+    // meausure how many people are using DANGEROUSLY_DISABLE_OOM in build statements
+    telemetry.decorateEvent(`BUILD_END`, {
```
wardpeet: @jamo wdyt of this? Maybe I should just omit it 😛

jamo: I like this a lot! 
We(I) need to allow this key in the API before we can ship it.

wardpeet: Maybe it's better to omit this for now? So the PR isn't blocked by this?
cc @jamo

KyleAMathews:
Is this intended as a temporary hack? E.g. once v8.seralize stuff gets figured out we remove this?

muescha:
i would add it anyhow to documentation with an big warning to avoid more and more "hidden preferences" or avoid some surprising behavior.

anyhow it would name it somehow what it do and not what it avoid `DANGEROUSLY_DISABLE_SAVING_REDUX_STATE`

pieh:
Yes it's meant to be temporary. Users are blocked on OOM issues and until `v8.serialize` land, there is nothing they can do to unblock themselves (other than monkey patching and commenting out `JSON.stringify` parts in `saveState` function.

@muescha This is intentionally not mentioning redux state. We don't want to encourage using it for the purpose of skipping redux state caching (as then we can never fix cache issues if users don't use it at all). This is only to unblock cases of out of memory issues. This will also be removed once proper fix is merged.

wardpeet:
Yes temporary yes. Let me just remove the analytics and get this merged.

packages/gatsby/src/redux/index.js
jamo: sorry for the delays and a good call.

muescha:
maybe comment in code why its temporary and what it solves ... just to see it later in code reviews and don't forget to remove it later

DSchau:
@wardpeet I'm gonna pick up this PR and add some documentation if that's cool!

PR #13070, opened 2019-04-03
feat(gatsby): Per-page manifest files

davidbailey00:
The goal is to remove `data.json` and use individual page data files, so that we can download these when required rather than downloading a single, potentially large file on initial page load.


wardpeet:
@davidbailey00 could you add a small description to your PR as @DSchau suggested ^^

davidbailey00:
Apologies for leaving this so long, just updated the description now! Having said that I think Anthony's PR / proposal seems more developed so we might want to move forward with that one - #13070

DSchau:
@davidbailey00 sorry that we crossed some wires here!

Let's close this out for the reasons you noted--but definitely consider leveraging your approach here and what you gained in assisting in getting #13004 or future, similar PRs iterated upon and shipped.

Thank you!

Issue #12713, opened 2019-03-21
[examples]: Add "using-i18n" example


LekoArts:
Since more and more people show interest in "How to use i18n with Gatsby" a first step can & should be adding a `using-i18n` example. There are various third-party libaries like [react-i18next](https://react.i18next.com/) or [lingui](https://github.com/lingui/js-lingui) and also already multiple gatsby-plugins like:
- https://github.com/angeloocana/gatsby-plugin-i18n
- https://github.com/ikhudo/gatsby-i18n-plugin
- https://github.com/ikhudo/gatsby-i18n-plugin

In my opinion we should show an example that uses none of these third-party libaries and also no gatsby-plugins because it can all be done with Gatsby's APIs and React itself. Also I think it's nice to show how things work and not hide something behind a plugin.

I have a working example (with a CMS) here: https://github.com/LekoArts/gatsby-starter-prismic-i18n

However, the example should use `gatsby-mdx` and for that one decision is still open: What strategy to use for the blogposts?

- Individual folders, so e.g.: `content/en/topic-1/index.mdx` and `content/de/topic-1/index.mdx`
- Same folders but different filenames, so e.g. `content/topic-1/index.en.mdx` and `content/topic-1/index.de.mdx`

----

Please let me know your thoughts about this `using-i18n` example

lannonbr:
I like the idea of having it without using any 3rd party plugin so rather than abstracting things away, you can focus in on the key architecture of doing internationalization.

With the blogpost strategy, I'm thinking the 2nd option is going to be cleaner in the long run so if you want to create a blog post and then have other people translate it, you can just tell them to duplicate the file in the folder it currently exists and then have them rename the file and translate the text

PR #13077, opened 2019-04-03
feat(examples): Add "using-i18n" example

LekoArts:
Closes https://github.com/gatsbyjs/gatsby/issues/12713

---

Example site that demonstrates how to build Gatsby sites with multiple languages (Internationalization / i18n) without any third-party plugins or packages. Per language a dedicated page is built (so no client-side translations) which is among other things important for SEO.


examples/using-i18n/src/pages/index.js
```diff
@@ -0,0 +1,56 @@
...
+  query Index($locale: String!) {
+    allMdx(
+      filter: { fields: { locale: { eq: $locale } } }
+      sort: { fields: [frontmatter___date], order: DESC }
+    ) {
+      edges {
+        node {
+          frontmatter {
+            title
+            date(formatString: "DD/MM/YYYY")
```
muescha: date should be also i18n for example:
- german: DD.MM.YYYY
- english: MM/DD/YYYY

polarathene: > english: MM/DD/YYYY

That format doesn't apply to all countries where English is the dominant language. Here in New Zealand it's commonly: `DD/MM/YYYY`.

This is more l10n (localisation) specific. You can see a good example of this on the [moment.js homepage](http://momentjs.com/), scroll to the bottom and see the different english locales. They also demonstrate ways to format the date to what you'd like while respecting the locale.

Alternatively something like [`toLocaleDateString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString) might also work? It's a lightweight/native alternative to something like Moment, which follows the official ECMAScript Internationalization API spec.

examples/using-i18n/src/components/layout.js
```diff
@@ -0,0 +1,35 @@
...
+// This e.g. enables the LocalizedLink to function correctly
+// As this component wraps every page (due to the wrapPageElement API) we can be sure to have
+// the locale available everywhere!
+const Layout = ({ children, pageContext: { locale } }) => (
+  <LocaleContext.Provider value={{ locale }}>
+    <div className="global-wrapper">
+      <header className="global-header">
+        <nav>
+          <LocalizedLink to="/" aria-label="Back to home">
+            Homepage
```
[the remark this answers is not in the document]
LekoArts: Can stay the same :)

examples/using-i18n/src/components/layout.js
```diff
@@ -0,0 +1,35 @@
...
+// Use the built-in Context API to make the "locale" available to every component in the tree
+// This e.g. enables the LocalizedLink to function correctly
+// As this component wraps every page (due to the wrapPageElement API) we can be sure to have
+// the locale available everywhere!
+const Layout = ({ children, pageContext: { locale } }) => (
+  <LocaleContext.Provider value={{ locale }}>
+    <div className="global-wrapper">
+      <header className="global-header">
+        <nav>
+          <LocalizedLink to="/" aria-label="Back to home">
```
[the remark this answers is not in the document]
LekoArts: Good idea!

examples/using-i18n/src/pages/index.js
LekoArts: Yeah, I don't really want to get into localization. But I passed the dateFormat as context, too. So both pages now show a different format :)

examples/using-i18n/src/utils/gatsby-node-helpers.js
```diff
@@ -0,0 +1,24 @@
+// Use a little helper function to remove trailing slashes from paths
+exports.removeTrailingSlash = path =>
+  path === `/` ? path : path.replace(/\/$/, ``)
+
+exports.localizedSlug = ({ isDefault, locale, slug }) =>
+  isDefault ? `/${slug}` : `/${locale}/${slug}`
+
+// From lodash:
+// https://github.com/lodash/lodash/blob/750067f42d3aa5f927604ece2c6df0ff2b2e9d72/findKey.js
+exports.findKey = (object, predicate) => {
```
DSchau: Nit: but if this is _only_ used by gatsby-node.js I'd recommend just using lodash.findKey directly.

However -- this shouldn't block this PR.

DSchau: For the _why_ of that statement... it's Node.js code so size isn't a huge concern. We can and should use utilities and libraries that are purpose-built for these types of things. While it may make sense to encapsulate our own utilities within this repo, this also means we can't as easily get features, fixes, etc. that we'd get for free merely by using the utility.

Polcius:
How can I fork the example? I have tried adapting my existing site with the config and it doesn't work, so I would like to fork the example to compare it and debug it. 

My error is in gatsby-node.js but I am at a loss debugging it:

```
TypeError: Cannot read property 'frontmatter' of null

  - gatsby-node.js:108 postList.forEach
    C:/Users/Pol/Desktop/Projects/sindhuka/gatsby-node.js:108:37

  - Array.forEach

  - gatsby-node.js:103 Object.exports.createPages
    C:/Users/Pol/Desktop/Projects/sindhuka/gatsby-node.js:103:12
```
Feel free to take a look: https://github.com/Polcius/sindhuka-serif/tree/multilanguage

Issue #12972, opened 2019-03-31
gatsby-plugin-sharp fluid maxWidth maxHeight misleading


VGoose:
## Description

- maxWidth and maxHeight will control aspect ratio/cropping if both are provided but this isn't documented 
- sizes is an option for `fluid` but not documented
- not clear that by default maxWidth or maxHeight is used to determine the source break points - leads to poorly optimized breakpoints if maxWidth or maxHeight is close to or over original size. 
- when cropped, placeholders remain at original aspect ratio #12008

### Steps to reproduce

```
  gatsby new fluid-test https://github.com/gatsbyjs/gatsby-starter-default
```
in src/components/image.js, call `fluid` like so:
```javascript
 fluid(maxWidth: 500, maxHeight: 400)
```

### Expected result

Expect maxWidth and maxHeight to control the max image size made by fluid.  

### Actual result

- Image gets cropped.
- Placeholder is still 1:1 aspect ratio

### Environment

- Firefox 66 
- Chrome 73

### Proposal 

- add cropping behavior option
- add additional option for cropping, instead of using `maxWidth` and `maxHeight`
- update docs

#11851

missmatsuko:
I think that using `maxWidth` and `maxHeight` to set aspect ratio shouldn't be documented, but rather the behaviour changed and crop or cover open added to address both the situations in https://github.com/gatsbyjs/gatsby/issues/11851#issuecomment-478279744

tremby:
I think you might be misunderstanding the `sizes` attribute. It's expected and desirable that some of the assets Gatsby produces are larger than the `maxWidth` and `maxHeight` parameters.

With those width and height parameters you're supposed to be expressing the largest width and height you're going to display the image at, *in CSS pixels*.

But your browser will choose a source image to make use of as many *physical* pixels as it can. If your display is high resolution (like most current mobile phones, and many current laptops), 

This is why

tremby:
My suggestion would be to add an option for cropping behaviour: cover or contain (and possibly further options for cover, to control how the cropping happens), and for now to set the default to cover, since it sounds like that's the current behaviour.

At the next major version bump, that default can change to contain, which I'd argue is the expected behaviour.

VGoose:
@tremby I didn't know about physical pixels vs CSS pixels before.  This makes a lot more sense now.  

I agree that the real issue now is the misleading nature of maxHeight and maxWidth and what they do with cropping. 

[...]

PR #4379, opened 2018-03-05
fix netlify-cms and sass plugin css extraction

erquhart:
Creates a separate `ExtractTextPlugin` instance for both the Netlify CMS and Sass plugins. We probably need every plugin that uses css extraction to do this.

Fixes #4335.

PR #13078, opened 2019-04-03
feat(*): add options fit and background to image sharp

VGoose:
## Description

gatsby-plugin-sharp
- add `fit` and `background` options to control cropping when using fluid

gatsby-transformer-sharp
- add new `ImageFitType` type and add `fit` and `background` to `fluidNodeType` args.

## Related Issues

Addresses #12972.


packages/gatsby-plugin-sharp/src/__tests__/process-file.js
```diff
@@ -47,12 +49,14 @@ describe(`createArgsDigest`, () => {
     testHashDifferent(`rotate change`, {
       rotate: defaultArgsBaseline.rotate + 1,
     })
-    testHashDifferent(`dutone change`, {
+    testHashDifferent(`duotone change`, {
```
wardpeet: thanks for fixing this typo!

yarn.lock
```diff
@@ -18740,6 +18740,11 @@ shell-escape@^0.2.0:
   resolved "https://registry.yarnpkg.com/shell-escape/-/shell-escape-0.2.0.tgz#68fd025eb0490b4f567a027f0bf22480b5f84133"
   integrity sha1-aP0CXrBJC09WegJ/C/IkgLX4QTM=
 
+shell-escape@^0.2.0:
```
wardpeet: I believe this is a mistake? could you revert this file?

packages/gatsby-plugin-sharp/README.md
```diff
@@ -96,13 +96,21 @@ a base64 image to use as a placeholder) you need to implement the "blur up"
 technique popularized by Medium and Facebook (and also available as a Gatsby
 plugin for Markdown content as gatsby-remark-images).
 
+When both a `maxWidth` and `maxHeight` are provided, the default way by which the image `fit` is `COVER`, with additional available options of `CONTAIN`, and `FILL`.
```
wardpeet: I'm not the best writer but how do you feel about:

```md
When both a `maxWidth` and `maxHeight` are provided, sharp will use `COVER` as a fit strategy by default. This might not be ideal so you can now choose between `COVER`, `CONTAIN` and `FILL` as a fit strategy. To see them in action the [CSS property object-fit](https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit) comes close to its implementation.

#### Note
fit strategies `CONTAIN` and `FILL` will not work when `cropFocus` is assigned to [sharp.strategy][6]. The `cropFocus` option cannot be `ENTROPY` or `ATTENTION`


```

packages/gatsby-plugin-sharp/README.md
```diff
@@ -96,13 +96,21 @@ a base64 image to use as a placeholder) you need to implement the "blur up"
...
+
+`CONTAIN` and `FILL` will not work with `cropFocus` of [sharp.strategy][6]. A `cropFocus` option must be passed that is not `ENTROPY` or `ATTENTION`.
+
+`background` - background colour when using a `fit` of `CONTAIN`, parsed by the [color module][18], defaults to black without transparency.
```
wardpeet: I would omit this line as people probably will look it up anyway in sharp if it throws an error

```suggestion
```

packages/gatsby-plugin-sharp/README.md
VGoose: Yeah this is better, thanks!

yarn.lock
VGoose: Done. I'm not sure why it changed.. possibly [#4379](https://github.com/yarnpkg/yarn/issues/4379)?

VGoose:
Failing ci/circleci: starters_validate
```
npm ERR! code ENOAUDIT
npm ERR! audit Your configured registry (https://registry.npmjs.org/) does not support audit requests.
```

I'm not sure how to fix this, all tests pass locally for me.

tremby:
This unfortunately doesn't quite cover the use case I had in mind.

In "contain" mode, with "maxWidth" and "maxHeight" both set, I was expecting the output images to still be the source image's aspect ratio, not forced to the ratio maxWidth:maxHeight and with a background colour applied.

Example:

- Source images are various sizes and aspects. A particular one is 2000x1000 pixels (2:1) ratio. Another is 1000x2000 pixels (1:2 ratio).
- We want to show them all in a grid of squares (1:1 ratio), but want to see the original aspects within those squares. Those squares are a presentation concern -- my HTML/CSS will deal with that. I just want suitably-sized images.
- We request fluid images with maxHeight 200, maxWidth 200, contain mode.

I expected the output for these particular images to be 200x100 pixels (along with smaller and bigger versions for different size viewports and pixel densities, eg 100x50, 150x75, 300x150, 400x200) and 100x200 pixels (along with smaller and bigger versions): the original aspect ratios, but with the "canonical" size fitting within the 200x200 I specified as maximum.

But with this patch I get two 200x200 images with black bars baked in (along with smaller and bigger versions, all with black bars).

VGoose:
@tremby thanks for the clear example! 

When I was making this PR I left [`inside` and `outside`](http://sharp.dimens.io/en/stable/api-resize/#resize) out because I didn't really understand the use case but now I think you're describing what `inside` is.  I was also trying to keep my first PR small and so less prone to errors. I think this should be easy to add; I'll take a stab at it soon. 

**Note for others who might read this: with `contain`, you can specify the `backgroundColor` and can make it any color or transparent to avoid the default black bars.

tremby:
That's a really useful link.

Yes, what I'm looking for and would have expected for "contain" is what they call "inside". I expect there are valid use cases for both, and I see no reason to break from their terminology. :)

While I can't think of a use case for "outside" off the top of my head, I do imagine one exists!

wardpeet:
@tremby this is still a great addition and another pr could define those other 2 props. @VGoose mind taking this one? 😄

VGoose:
Yep! I'll submit a PR once it gets completed.

PR #13080, opened 2019-04-03
fix(www): Hubspot form UI

fk:
Currently the Hubspot form UI at https://www.gatsbyjs.org/blog/2018-08-01-partner-program/#partner-application-form looks quite a bit broken. This wasn't always the case.

Under the hood we are using https://github.com/escaladesports/react-hubspot-form. It allows passing options to HubSpot (https://github.com/escaladesports/react-hubspot-form#options) via component props, which we at some point successfully used to disable the default HubSpot CSS: https://github.com/gatsbyjs/gatsby/blob/bb43fe0c926f2b5d06fb6e9fa3a057ad5d3a685d/www/src/components/hubspot-form.js#L64

I understand we did t like this to bypass emotions `css` prop. I strongly suppose this doesn't work anymore; it looks like the styling issues are caused by us currently loading the HubSpot default CSS — please compare their default CSS documented at https://designers.hubspot.com/docs/cos/hubspot-form-markup#default-embedded-form-css to what I'm seeing on .org:

image (3)

This PR works around the issue by pulling in the current version of `react-hubspot-form` locally, and setting the `css` option there. I don't know if there's a more elegant way to fix this. Happy to make adjustments!

---

Along the way, I adjusted the hubspot form input styles to match the reference "Homepage Newsletter" input; before/after:

image

image

### DIFF
--- a/www/package.json
+++ b/www/package.json
@@ -69,7 +69,6 @@
     "react-dom": "^16.4.1",
     "react-helmet": "^5.2.0",
     "react-highcharts": "^16.0.2",
-    "react-hubspot-form": "^1.3.5",
     "react-icons": "^2.2.7",
     "react-instantsearch": "^4.5.1",
     "react-modal": "^3.4.4",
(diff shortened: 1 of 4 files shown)


muescha:
maybe as follow up: 

create a issue for an option/parameter in https://github.com/escaladesports/react-hubspot-form/issues so that the inline code can replaced when fixed

PR #13089, opened 2019-04-03
refactor(gatsby-cli): rework cli to use ink

### LINKED ISSUE #11076: [feat]: improve logging by introducing structured logging
I am restructuring this issue after my pair session with @DSchau  to cover the
broader problems facing (discussed while trying to fix this one)
with logging in Gatsby.

To solve the flickering issue it will require either an ugly hack
or more structural changes.

Current problems with logging:

* non standard way to log in gatsby core and in packages. Plugin authors
  are using a mixture of `console.warn`, `console.info`, `console.log`, the
  reporter from gatsby-cli (more on this later), progress bars and possibly other
  methods.
* no way format log output depending on the platform running gatsby build.
  Right now if the build is running on a CI it doesn't support color handling then
  the logs are full of coloring codes. In addition logs disappear depending on
  the stage of the build.
* the reporter being in the gatsby-cli makes it hard for plugin authors and the
  core plugins to use it. It is passed along on most the gatsby node API but then
  it is up to the user to pass it along to more internal APIs such as the sharp
  transformer or the `createRemoteFileNode` which makes the use of the reporter
  in those cases non possible (besides importing in an ugly fashion the file)
* there is no way for plugin authors to set a level of logging such as
  `INFO`, `ERROR` to limit or make more verbose their logs
* while errors are outputted on stdout it could be interesting to log errors in
  a separate file (like yarn-error) that end users could be attach to issues when
  describing failures, rather than searching for the interesting parts in stdout,
  scrolling and c/p the logs around.
* concurrent logs as shown in the image are causing issues

Quoting Kyle 
>  blinks back and forth with the the updates from the "source and transform nodes" log line — we need to add a way to core to display multiple things at once and clean up stuff like this.

example

wardpeet:
## Description
This PR brings the world of CLI & React to Gatsby with
[image](https://github.com/vadimdemedes/ink)

This PR replaces our augmented `yurnalist` reporter with [ink](https://github.com/vadimdemedes/ink). Everything still works the same as before so no changes were made to the reporter api.

One extra thing that has been done is mapping the native console API into the reporter as well so everything is handled by INK.

Gatsby build:
[asciicast](https://asciinema.org/a/uGyjR6109yVr3jrsmdOUu3MLa)

Gatsby develop:
[asciicast](https://asciinema.org/a/dTIkluZabytMQiwUKZKSQwie3)

## Related Issues
Partially fixes #11076

cc @vadimdemedes @sindresorhus if they like to give some pointers


packages/gatsby-cli/package.json
```diff
@@ -22,11 +22,14 @@
...
     "gatsby-telemetry": "^1.0.4",
     "hosted-git-info": "^2.6.0",
+    "ink": "^2.0.5",
+    "ink-spinner": "^3.0.0",
     "lodash": "^4.17.10",
     "meant": "^1.0.1",
     "node-fetch": "2.3.0",
     "opentracing": "^0.14.3",
     "pretty-error": "^2.1.1",
+    "react": "^16.8.4",
```
wardpeet: react is a dependency of cli because ink needs it. Because cli can be installed globally a Peer dependency won't work and we won't use this version in the browser or anything so the user still needs to install it's own version on their site.

packages/gatsby-cli/src/reporter/index.js
```diff
@@ -65,8 +55,8 @@ module.exports = Object.assign(reporter, {
       error = message
       message = error.message
     }
-    base.error.call(this, message)
-    if (error) console.log(errorFormatter.render(error))
+    // base.error.call(this, message)
```
DSchau: I don't really know what this was doing--seems like just calling reporter.error but from _this_ context.

Probably safe to remove presuming things seem like they're backwards compatible.

packages/gatsby-cli/src/reporter/index.js
```diff
@@ -65,8 +55,8 @@ module.exports = Object.assign(reporter, {
...
+    if (error) this.log(errorFormatter.render(error))
```
DSchau: While we're here--we should probably fix this to use the correct code (e.g. .error). This was something that maybe is impacting cloud re: stripping ANSI color codes.

packages/gatsby-cli/package.json
KyleAMathews: React — boldly going where no JS framework has ever gone before 👩‍🚀

packages/gatsby-cli/src/reporter/index.js
```diff
@@ -75,43 +65,25 @@ module.exports = Object.assign(reporter, {
...
-      return `${convertHrtime(elapsed)[`seconds`].toFixed(3)} s`
-    }
+    return inkReporter.createActivity(name, activityArgs)
+  },
+}
 
-    const { parentSpan } = activityArgs
-    const spanArgs = parentSpan ? { childOf: parentSpan } : {}
-    const span = tracer.startSpan(name, spanArgs)
+console.log = reporter.log
```
DSchau: We should _make sure_ we want to overwrite globals like this. It does solve the problem nicely re: plugins opting in to our reporter, though 🤔

packages/gatsby-cli/src/reporter/ink/reporter.js
```diff
@@ -0,0 +1,149 @@
...
+        this.setState(
+          state => {
+            const activities = { ...state.activities }
+            delete activities[name]
+
+            return {
+              activities,
+            }
+          },
+          () => {}
```
stefanprobst: Do we need this?

packages/gatsby-cli/src/reporter/index.js
wardpeet: I actually fixed it immediately when we I was looking at the diff myself

packages/gatsby-cli/src/reporter/index.js
wardpeet: `errorFormatter` should be set with no color when the function is being set:
https://github.com/gatsbyjs/gatsby/pull/13089/files#diff-8a16e03a18503e23944418632510f11bR34

I'll still need to check if it also strips already added colors.

packages/gatsby-cli/src/reporter/index.js
wardpeet: I'm not the biggest fan of this either but INK doesn't play super nicely with other process writing to stdout.

I've looked into the node codebase and they are actually just using util.format so I'm using it too.
https://github.com/gatsbyjs/gatsby/pull/13089/files#diff-5753f39bea69d2159152a45abdca82eaR87

I probably want to add some tests (maybe react snapshots could work here) to see if everything is wired correctly.

packages/gatsby-cli/src/reporter/ink/reporter.js
wardpeet: nope :) just some debugging left :p I removed it.

wardpeet:
@KyleAMathews I added Gifs and links to asciinema

packages/gatsby-cli/src/reporter/index.js
KyleAMathews: yeah, my thinking is that instead of directly replacing console.log with reporter.log we should `util.format` what's passed in and then send it to reporter.log — we're going to overload the reporters so that core & plugins can send in structured data which can then be rendered by a component for richer reporting. So we can replace all the places in core where we're trying to do more sophisticated layout of messages with sending data to a component.

KyleAMathews: we'll centralize design in some standard components which we can continue to iterate on over time + we'll have structured data that a) we can send out as JSON for `--json` output and that we can send to other UIs e.g. the coming Gatsby UI

wardpeet: I think if we look at structured logging so have multiple ways of saving these logs I think it's better to move to winston and have an ink reporter as a plugin. (probably v3)

i'll like the console.log switch, let me update that.

wardpeet:
Note to myself need to check our sources to see what api yurnalist provides and which we use.

for example `reporter.format` is a thing which seems to be an alias to chalk

packages/gatsby-cli/src/reporter/ink/reporter.js
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  setVerbose(isVerbose = true) {
+    this.verbose = isVerbose
+  }
+
+  _addMessage(type, str) {
+    // threat null/undefind as an empty character, it seems like ink can't handle empty str
+    if (!str) {
+      str = `\u2800`
```
vadimdemedes: I think this issue is fixed now, you can use empty string here ;)

vadimdemedes:
Amazing, excited to see this PR ship! 🎈

vadimdemedes:
Let me know if you run into any issues with Ink.

jest.config.js
```diff
@@ -36,6 +36,8 @@ module.exports = {
   transform: { "^.+\\.js$": `<rootDir>/jest-transformer.js` },
   moduleNameMapper: {
     "^highlight.js$": `<rootDir>/node_modules/highlight.js/lib/index.js`,
+    // hosting isn't being done by yarn workspaces so we map it ourselves
```
wardpeet: > 
```suggestion
    // hoisting isn't being done by yarn workspaces so we map it ourselves
```

DSchau:
@wardpeet once the e2e test is fixed (which required upstream changes?) is this ready to go?

If so -- 🎉

wardpeet:
should be ready, only cavaet is that node 8 is inside the engine list of ink so basically we drop support for node 6 on gatsby-cli

millette:
@wardpeet Well, node 6 support is ending at the end of April according to <https://github.com/nodejs/Release#release-schedule>. If the new (ink) cli is released in 2 weeks, it should help to encourage upgrades.

wardpeet:
@millette that's true :) I don't think it's a big deal 😛 but it is still a breaking change 🙄.

packages/gatsby-cli/src/reporter/index.js
```diff
@@ -1,17 +1,15 @@
 // @flow
-
-const { createReporter } = require(`yurnalist`)
+// loads Object.entries polyfill on node 6
+// @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries
+require(`object.entries/auto`)
```
pieh: I think this should be handled by https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-cli/src/index.js#L5 ? We do use `object.entries` in different parts of the code and it works, and it's most often just tests that fail on those (because they don't use `gatsby-cli` as entry point?

wardpeet: I see, should I just put it in the failing tests? Or not bother with node 6 at all?

sw-yx:
i think im a bit behind - what exactly is different about the CLI output here? is it that no random partial artefacts are left behind when another process console.logs? just trying to make sure i really understand.

wardpeet:
@sw-yx sorry for my late response. This is PR that replaces yurnalist with ink so the output itself shouldn't be changed. We will create some follow-up prs that actually changes output.

wardpeet:
yurnalist is used on CI to make sure we pipe to stderr as ink does not which means azure's OOM is fixed 😛

packages/gatsby-cli/src/reporter/reporters/index.js
```diff
@@ -0,0 +1,8 @@
+const semver = require(`semver`)
+const { isCI } = require(`ci-info`)
+
+if (semver.satisfies(process.version, `>=8`) && !isCI) {
```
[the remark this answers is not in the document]
wardpeet: ink works on non-interactive consoles, the reason why we check for isCI, is that ink does not pipe to stderr and yurnalist does.

wardpeet:
> I'm not 100% sure about the reporter design. It feels we'll have to extract lots of that stuff from component when we move to structured logging

I don't really understand this comment. What do you mean with extracts lots of that stuff?

packages/gatsby-cli/src/reporter/reporters/index.js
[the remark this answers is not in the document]
freiksenet: Until v3 I guess.

freiksenet:
My comment is just about general architecture of this. Let's move it out of the context of this PR, I think it's fine as it is otherwise.

packages/gatsby-cli/src/reporter/reporters/index.js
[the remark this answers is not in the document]
pieh: I'm not sure why user would want to specifically use `yurnalist` over `ink`? It might make some sense when we start implementing actual CLI changes, but here's mostly 1:1 implementation with only visible changes being different spinner and not mangled output :)

PR #13095, opened 2019-04-03
[WIP]: feat(docs): add creating an example project to docs

DSchau:
## Description

This PR adds a doc for creating an "Example Project" that we can use for issues/PRs/etc. when an example is helpful. This is solving an issue with pull requests being opened that are very hard to validate (e.g. specific Wordpress setup causing issues), so it'd be nice to point to something to speed up the review process.

This is (and should be?) separate from [Creating a Reproducible Test Case](https://gatsby.dev/reproduction) but I'm happy to tweak this to be additional notes on the existing document, as well. Unsure which is clearer!

Few notes:

- Not convinced this should go in the sidebar (seems confusing?)
- Unsure of terminology: "Example Project" seems OK, but open to something else

### DIFF
--- a/docs/contributing/how-to-make-an-example-project.md
+++ b/docs/contributing/how-to-make-an-example-project.md
@@ -1,0 +1,23 @@
+---
+title: How to Make an Example Project
+---
+
+> Looking for how to create a reproduction for a Gatsby bug? Check out [How to Make a Reproducible Test Case][reproducible-test-case]
+
+## What is an example project?
+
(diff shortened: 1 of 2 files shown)


muescha:
### "test case project" vs "example project" 

after reading the text i don't see a difference between "test case project" and "example project" :(
maybe if its on the same page it can be explained better where is the difference

### a side note about private keys or something would be nice: 
you should explain not to expose private credentials with this example (if its on github) when a user provide a working example.

marcysutton:
I agree with @muescha that it isn't super clear what the difference is. Perhaps the doc should answer what an example project would consist of that a test case wouldn't?

muescha:
and same on test case page also the do and do not

DSchau:
Thanks for the feedback! I think it makes sense to augment the existing document with this additional info, so I'll do that now.

DSchau:
Let's just close this for now--will address later!

PR #13096, opened 2019-04-03
fix(gatsby-cli): Only init git repository if no git repository exists

### LINKED ISSUE #12858: [gatsby-cli] Don't create .git if project is created into existing git repo
I'm on an existing git repo and run `gatsby new mysite`.
Generating a nested .git folder for the gatsby site was confusing me as I couldn't add the newly generated site to git which though it was a git submodule or something...

Follow-up of conversation with @KyleAMathews here https://twitter.com/kylemathews/status/1110276125345775616

mayaarguelles:
## Description

This checks the working directory to see if an existing git repository exists when `gatsby new mysite` to prevent generating nested .git folders. If no .git folder is found backtracking through the working directory's path, a new git repository is initialized.

## Related Issues

FIxes #12858


muescha:
maybe just run (solutions from stack overflow):
- `git status` or 
- `git rev-parse --is-inside-work-tree` or 
- `git -C <path> rev-parse`

packages/gatsby-cli/src/init-starter.js
```diff
@@ -29,6 +29,20 @@ const shouldUseYarn = () => {
 
 // Initialize newly cloned directory as a git repo
 const gitInit = async rootPath => {
+  let curDir = sysPath.normalize(rootPath)
```
wardpeet: I like @muescha's response & also what @sidharthachatterjee had to say

I'm thinking of something like this
```js
const isAlreadyGitRepository = async () => {
  try {
    await spawn(`git rev-parse --is-inside-work-tree`)
    return true;
  } catch (err) {
    return false
  }
}

// Clones starter from URI.
const clone = async (hostInfo: any, rootPath: string) => {
  let url
  // Let people use private repos accessed over SSH.
  if (hostInfo.getDefaultRepresentation() === `sshurl`) {
    url = hostInfo.ssh({ noCommittish: true })
    // Otherwise default to normal git syntax.
  } else {
    url = hostInfo.https({ noCommittish: true, noGitPlus: true })
  }

  const branch = hostInfo.committish ? `-b ${hostInfo.committish}` : ``

  report.info(`Creating new site from git: ${url}`)

  await spawn(`git clone ${branch} ${url} ${rootPath} --single-branch`)

  report.success(`Created starter directory layout`)

  await fs.remove(sysPath.join(rootPath, `.git`))

  const isGit = await isAlreadyGitRepository()
  await install(rootPath)
  if (!isGit) {
    await gitInit(rootPath)
  }
  await maybeCreateGitIgnore(rootPath)
  if (!isGit) {
    await createInitialGitCommit(rootPath, url)
  }
}
```

ashley:
@muescha I've added a condition that checked whether or not the spawn failed. Let me know if it looks good (I'm still new to using promises in this codebase).

ashley:
Hi @sidharthachatterjee, let us know if this looks good or if there's anything else we can add 👍

wardpeet:
@mayaarguelles @ashley what do you think of moving to git commands? If git is not installed we should skip just all together.

ashley:
@wardpeet I think one of my teammates who was working on another PR pushed to the same forked branch so I've moved our changes to another PR #14010 . Is using git commands like `git rev-parse --is-inside-work-tree` what you are proposing?

DSchau:
@ashley we should have resolved the issue that required you to open a separate PR--but happy to review this however it is structured!

Can we close either this one or #14010 so we can unify on one approach? We'll squash and merge in any scenario, so as long as _this_ PR contains the changes you want, we probably don't need the other one.

ashley:
Ah @DSchau thank you for clarifying. I think I got confused because the files changes I saw were from older commits. I've added by our changes from the other PR, and I will close the duplicate PR I made.

muescha:
I am not so happy that it return a `false` when the command throw an exception. Because it can also be an git repo when the command failed.

wardpeet:
@muescha you're right but I think it's good enough. If we get any issues we can fix it later

zzzz4:
Hi!
There might still be a lingering issue:
`gatsby new` **WILL** create a git repo within an existing git repo.

However, `gatsby new my-site` WILL NOT create the git repo, as expected based on this fix.

So the code path when a site name is not provided to `gatsby new` is not behaving as expected...

PR #12842, opened 2019-03-25
feat(www): Move sidebar settings to YAML, add sidebar title "meta" field

fk:
Inspired by #12767. Describing the changes is a bit … complicated, matching the convoluted stuff we are currently doing for our tutorial/docs/features sidebars. Please ask if something is unclear!

---

Prior to this, we added information

- about the sidebar `key` (`docs`, `tutorial`, `contributing`, `features`), used to write and read local storage information about the expanded/collapsed status of sidebar items,
- wether the sidebar has an „Expand/Collapse All“ button („Features“ does not) via the `disableExpandAll` option, and
- wether the sidebars _first level_ accordions should be disabled via the (unfortunately not clearly named) `disableAccordions` option (again to facilitate the "Features" sidebar UI)

to _each_ sidebar item’s information in [`src/utils/sidebar/item-list`](https://github.com/gatsbyjs/gatsby/blob/master/www/src/utils/sidebar/item-list.js). 💩 

This PR moves that information to the root node of each sidebar’s YAML file  in [`src/data/sidebars`](https://github.com/gatsbyjs/gatsby/tree/master/www/src/data/sidebars):

- puts all current sidebar meta/config data into the sidebar’s YAML
- allows a simpler, generic `src/utils/sidebar/item-list`
- adds a **setting for the sidebar title without having to abuse sidebar items** (see the following description, and https://github.com/gatsbyjs/gatsby/issues/12767#issuecomment-475815324 ff.)

—

Prior to this, we also

- nested all navigation items in „Contributing“ and „Tutorial“ under a root item „Contributing“ and „Tutorial (step-by-step)“ respectively, **only to have a headline for the sidebar items** 😿 
- prevented those items from being collapsable by setting `disableAccordions` to `true`

With this PR, we now expect exactly one node for each sidebar YAML.
This node’s 

- `title` is now used to provide the title for the sidebar,
- `disableAccordion` and `disableExpandAll` can be set here
- `items` does now contain the items to be rendered in the sidebar

`title` allows us to drop nesting all navigation items in "Contribution" and "Tutorial", and removing the "Documentation" item in `src/data/sidebars/doc-links.yaml`.

—

We also relied on the „Tutorial“ sections to be subitems in regards to their styling — we didn’t want them to be rendered as small caps items, like all root-level items with subitems were rendered up to now.

[...]

PR #13101, opened 2019-04-03
chore(docs) Bring back small caps titles for /docs/ sidebar first-level accordions

fk:
* back to a more obvious information hierarchy for https://www.gatsbyjs.org/docs/ by bringing back small caps titles for first-level sidebar accordions (originally removed in #12842)
* move "Cheat Sheet" link down below "Commands (Gatsby CLI)"
* link to "Cheat Sheet" from "Commands (Gatsby CLI)" and vice versa
* move "Glossary" stub down a bit
* finally remove the "Ecosystem" section as we have all its links at the top of the docs hierarchy now, and the item itself only links to https://www.gatsbyjs.org/ecosystem/ (which we still have to take a leap for)

image

image


fk:
Thanks Dustin! (-;

shannonbux:
Just getting to this after coming back into town!

I'm wondering if it's possible to make the cheat sheet clickable and just use that as a landing page for the docs? It seems like it's the most up-to-date resource on what Gatsby users want at their fingertips based on @marcysutton's research, so surfacing it in some way would make sense to me. @calcsam what was the reason for putting it below the fold?

marcysutton:
@shannonbux do you mean the Cheat Sheet images? They link to full-size images, and there is PDF link to download as well as a link to the Quick Start documentation.

I didn't get a chance to review this before it was merged unfortunately, but I'm wondering what the reasoning was for keeping Awesome Gatsby up so high in the sidebar (what do the analytics look like for that page?). The [ecosystem page](https://www.gatsbyjs.org/ecosystem/) links to it as "Browse Resources", perhaps we could move it farther down? 

Related to all of this: one of my Learning tasks is to come up with a proposal for a docs IA reorg (things like Advanced Tutorials moving into Tutorial(s), which could give the docs a more limited scope for translations). I'll be sure to loop y'all in when there is something ready to review, so we can pick up from where this PR left off.

Thanks for all your work on this, @fk!

fk:
@marcysutton Sorry for not letting you properly review it — Dustin and me went a little too YOLO here with the quick merge!

> I'm wondering what the reasoning was for keeping Awesome Gatsby up so high in the sidebar

@calcsam was of the same opinion, it was me who worried about mobile users and "changing too much on the fly". We also observed that the whole top part of "Awesome Gatsby" is quite redundant, so I'm absolutely seeing where you're coming from.

I really shouldn't be messing around here, I mainly tried to pick up the feedback in regards to information hierarchy, and adressed the other input from the team in the same PR, which was a mistake IMO — content and UI appearance should have went to separate PRs.

marcysutton:
No worries, this is a great PR! We can make subsequent changes for sure–I think the styling stuff and IA reorg can be separate items, so it makes sense to not hold that work up here.

PR #13107, opened 2019-04-04
chore(www) Align more design tokens with system-ui spec (scale -> fontSizes, dimensions -> sizes)

fk:
Ref. https://system-ui.com/theme/

- rename `scale` to `fontSizes` — this is a no-brainer really (and comes with the benefit of neatly distingushing itself from Typography.js' `scale`)
- rename `dimensions` to `sizes` as currently discussed in https://github.com/system-ui/theme-specification/pull/2

/cc @jxnblk @johno


www/src/components/banner.js
```diff
@@ -2,14 +2,14 @@ import React from "react"
...
-import { colors, space, dimensions, fonts } from "../utils/presets"
+import { colors, space, sizes, fonts } from "../utils/presets"
 
 const horizontalPadding = space[6]
 const backgroundColor = colors.gatsby
 
 const BannerContainer = styled(`aside`)`
   background-color: ${backgroundColor};
-  height: ${dimensions.bannerHeight};
+  height: ${sizes.bannerHeight};
```
jxnblk: Yeah, I definitely think having `sizes` makes a lot of sense

fk: 💯 Next up: Tidying up `breakpoints`, and adding `mediaQueries`!
Thanks so much for that initiative @jxnblk! 🙏

Issue #13073, opened 2019-04-03
[gatsby-source-filesystem] processRemoteNode concurrent requests overload


violy:
## Relevant information

During build, on my local environment, using `gatsby-source-wordpress` to fetch Wordpress `custom_post`, `acf` (with `image` and `gallery`) I observe the `Failed to process remote content` warning during media fetching, warning logged in `gatsby-source-filesystem`, `create-remote-file-node`.

This failure prevents media to be fetched, and the build failed. 
In walking through gatbsy github issues… and found this comment, https://github.com/gatsbyjs/gatsby/issues/5371#issuecomment-390969703 
and fix the warning ! 
only replace `concurrent` value `200` with  `10` in `Queue` managment !

https://github.com/gatsbyjs/gatsby/blob/15cdf2c77e4e5cd1086b87bc9c828128c9022f24/packages/gatsby-source-filesystem/src/create-remote-file-node.js#L80-L92

But i fix this locally… 
**is there a way to configure this _concurrent value_, maybe through a `gatsby-source-filesystem` config options ?** 
the value seems actuallys static… why is this value static ?

thanks.

### Environment

Mac OS, local dev build.

pieh:
It's not feasible to do this via pluginOptions, because it's the other plugins (not `gatsby-source-filesystem`) that are calling `createRemoteFileNode`. All this code really need to be rewritten to handle failures much better, but in meantime perhaps we can use env var to overwrite default limit - for example:

```
concurrent: process.env.GATSBY_CONCURRENT_DOWNLOAD || 200, 
```

and use it either in `.env` files ( https://www.gatsbyjs.org/docs/environment-variables/#server-side-nodejs ) or command line / package.json script ( `GATSBY_CONCURRENT_DOWNLOAD=10 gatsby develop` )

violy:
Ok, it's a quick resolution, and I like it. 

Such _similar_ environment core configurations is listed/documented somewhere ?

violy:
According your resolution, @pieh 
I've made a Pull Request. 

## other statics in `create-remote-file-node.js`

The PR adding 2 others environment const, initially statics, in `create-remote-file-node.js` 

* `GATSBY_DOWNLOAD_TIMEOUT`
* `GATSBY_DOWNLOAD_RETRIES`

assuming previous statics values as default, preventing breaking change. 

## more verbose warning

Also, I improve the `Failed to process remote content` warning, including the error message, after url. 
[...]

PR #13110, opened 2019-04-04
feat(gatsby-source-filesystem): add an environment variable to control concurrent queue size

violy:
related to #13073

### DIFF
--- a/packages/gatsby-source-filesystem/README.md
+++ b/packages/gatsby-source-filesystem/README.md
@@ -62,6 +62,8 @@ They will be added to the following default list:
 ../**/dist/**
 ```
 
+To prevent concurrent requests overload of `processRemoteNode`, you can adjust the `200` default concurrent downloads, with `GATSBY_CONCURRENT_DOWNLOAD` environment variable.
+
 ## How to query
 
 You can query file nodes like the following:
--- a/packages/gatsby-source-filesystem/src/create-remote-file-node.js
+++ b/packages/gatsby-source-filesystem/src/create-remote-file-node.js
@@ -81,7 +81,7 @@ const createFilePath = (directory, filename, ext) =>
 const queue = new Queue(pushToQueue, {
   id: `url`,
   merge: (old, _, cb) => cb(old),
-  concurrent: 200,
+  concurrent: process.env.GATSBY_CONCURRENT_DOWNLOAD || 200,
 })
 
 /**


pieh:
Thanks for changes @violy!

Just curious what were values you used for timeout / retries - maybe better approach would be to change defaults there?

violy:
as I explain on the issue #13073 — during my troubleshooting, I observed this warning, for different reasons (server overload, network / local overload) related to my stack and my work environment, and my successive attempts to valid locally my build. 

During my troubleshooting, I've try to change `timeout` & `retries` values, **but with no effect**. The only success was changing `concurrent` to lower value (`10` instead of `200`, in my case) the original question in my issue / PR. 
But I think to _other case_ where troubleshooting could be painful, access to `const` via `.env` variables or equivalent, is a good way to adjust the stack with no pain. 

When I finally deployed my app, via _netlify_, there was **no problem** (no need to fix with custom fork 😅)! 
So initial default static values may be not so bad!

pieh:
So if changing those didn't have any effect I'm not sure it makes sense to add yet more moving parts. I would prefer to just add `GATSBY_CONCURRENT_DOWNLOAD` and more detailed error message here.

muescha:
the new possible env variables should be anywhere documented:

- GATSBY_CONCURRENT_DOWNLOAD
- GATSBY_DOWNLOAD_TIMEOUT
- GATSBY_DOWNLOAD_RETRIES

violy:
Thank you, according your comment @pieh 
I've update my PR, by removing the variables `GATSBY_DOWNLOAD_TIMEOUT` & `GATSBY_DOWNLOAD_RETRIES` maybe useless.

muescha:
Maybe it would be nice to have an "main entry point" for all "magic numbers" and a default solution how to overwrite it.

LekoArts:
@violy Hi 👋 Do you have time to update your PR (adding information to README) and resolve the conflicts? Would be cool to have this in!

violy:
@LekoArts thanks for your request, 
so I've resolve conflict and add a small word in the README Configuration paragraph and force-push all the change in one commit.

violy:
Any news / reviewer ?

muescha:
This download manager should also include an cache and some strategies how not redownload media assets when nothing is changed.

DSchau:
@muescha I think the issue (#10049) is a better place for that comment; thanks! Otherwise it's very possible we'll lose track of your suggestions.

muescha:
Is there also a stragegy only to download the used media assets (when maybe more assets are saved in wordpress but only some used in the used articles)?

muescha:
Thanks for your hint. I was thinking i am already in the other issue.

KyleAMathews:
Yeah, env variables seem appropriate now as a temporary fix until we can work on #10049

Issue #12138, opened 2019-02-27
The site is empty after going back


whs-dot-hk:
## Description

Describe the issue that you're seeing.

I see nothing, the page is blank.

### Steps to reproduce
1) Visit https://www.gatsbyjs.org/showcase
2) Click on ReactJS
2) Click the purple visit site button
3) Go back

Clear steps describing how to reproduce the issue. Please please please link to a demo project if possible, this makes your issue _much_ easier to diagnose (seriously).

### Expected result

What should happen?

See the original page.

### Actual result

What happened.

Blank

### Environment

Run `npx gatsby info --clipboard` in your project directory and paste the output here.

JacobBlomgren:
Interesting. I can reproduce on my machine. The console shows `TypeError: Cannot read property 'json' of undefined` and this https://github.com/gatsbyjs/gatsby/blob/3971ed4d6cc788b2fdc44d3b197bfe3fe68037ab/packages/gatsby/cache-dir/public-page-renderer-prod.js#L12 seems to be the line that breaks.

whs-dot-hk:
Hi, my error is like this

![screenshot from 2019-02-27 23-07-37](https://user-images.githubusercontent.com/35202375/53500055-92b80a00-3ae4-11e9-8c63-46ecf36e75b0.png)

t2ca:
Im getting the same error and the blank page.

```
TypeError: Cannot read property 'json' of undefined
    at c (public-page-renderer-prod.js:12)
```

whs-dot-hk:
Looks like the findpage function is null

https://github.com/gatsbyjs/gatsby/blob/3971ed4d6cc788b2fdc44d3b197bfe3fe68037ab/packages/gatsby/cache-dir/loader.js#L163

but I can't fire a debugger to confirm

How to debug a built gatsbyjs site?

yogeshkotadiya:
Not just `reactjs.org` site but all the sites.

* Go to [showcase](https://gatsbyjs.org/showcase) Page
* Click on the site title.
* Press Reload button.

JacobBlomgren:
@whs-dot-hk Do you want to take a stab at this? Otherwise I can try.

LaRuaNa:
@JacobBlomgren  I'd also like to give that one a try if you are not already on it :]

whs-dot-hk:
@JacobBlomgren You can

JacobBlomgren:
[...]

PR #13113, opened 2019-04-04
fix(gatsby): bail early if pageResources are not available in public pageRenderer

whs-dot-hk:
See https://github.com/gatsbyjs/gatsby/issues/12138

### DIFF
--- a/packages/gatsby/cache-dir/public-page-renderer-prod.js
+++ b/packages/gatsby/cache-dir/public-page-renderer-prod.js
@@ -6,6 +6,10 @@ import loader from "./loader"
 
 const ProdPageRenderer = ({ location }) => {
   const pageResources = loader.getResourcesForPathnameSync(location.pathname)
+  if (!pageResources) {
+    return null
+  }
+
   return React.createElement(InternalPageRenderer, {
     location,
     pageResources,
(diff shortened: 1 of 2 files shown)


pieh:
This doesn't seem correct - it will render same page component infinite amount of times (as it will render itself recursively)

whs-dot-hk:
@pieh Brilliant.

It looks something like here

https://github.com/gatsbyjs/gatsby/blob/2ba27496107bb6895e6e1371f0b75b5e93a9cafa/packages/gatsby/cache-dir/json-store.js#L87-L89

but I can't fire up the production build with `yarn build` here to confirm

pieh:
@whs-dot-hk we don't use `json-store` in production (only in development). The crashes happen if `pageResources` in `public-page-renderer-prod.js` is null and we try to access `pageResources.json` (in https://github.com/gatsbyjs/gatsby/blob/6dc5dcd5d5b7205d426fc0ffc34d73fd0e3247bd/packages/gatsby/cache-dir/public-page-renderer-prod.js#L16 )

pieh:
I was testing it out, and it works, I just forgot to merge it after I pushed changes and was waiting for tests to come back

whs-dot-hk:
Thanks, one more thing we will have to upgrade the gatsby package version of the site later when there is new release

PR #13122, opened 2019-04-04
chore(gatsby-plugin-page-creator): replace parse-filepath polyfill with native path.parse

paulmelnikow:
## Description

This removes `parse-filepath` which is a polyfill and replaces it with native `path.parse`.

The code using it is not bundled, and Gatsby isn't using any of the additional features of `parse-filepath`.

See discussion: https://github.com/gatsbyjs/gatsby/pull/11304#discussion_r268219015


paulmelnikow:
Hmm, https://circleci.com/gh/gatsbyjs/gatsby/101144 looks like a real failure, though it's rather puzzling to me. Does anyone understand what that test is about?

I'd be happy to add some documentation about why the polyfill is being used, once I understand. 😁

packages/gatsby-plugin-page-creator/src/create-path.js
```diff
@@ -1,9 +1,8 @@
 // @flow
-import parsePath from "parse-filepath"
 import path from "path"
 
 module.exports = (filePath: string): string => {
-  const { dirname, name } = parsePath(filePath)
+  const { dirname, name } = path.parse(filePath)
```
pieh: there is no `dirname` in result of `path.parse` (it will be `undefined`)

paulmelnikow:
Ah, thanks for wrapping this up!

Issue #13119, opened 2019-04-04
Default aliases are not used by gatsby-remark-prismjs


ojeytonwilliams:
## Description

Prismjs language aliases like `shell` for `bash` are not being accepted.

### Steps to reproduce

I put together a [demo project](https://github.com/ojeytonwilliams/gatsby-remark-prismjs-bug).  In that project there's a markdown file with a code block

> \```shell
>   git pull origin master
> \```

which should produce a page, `/shell/`, in which the word 'git' is highlighted

### Expected result

Code blocks labelled with aliases should be highlighted

### Actual result

Those code blocks only receive the default style.  Changing the label to `bash` results in the correct highlighting.

It is possible to create aliases via options passed to the plugin, but I think the existing ones should be picked up by default.

I'm happy to put together a PR for this (I'm 90% sure I know how to solve it), but I figured I'd best put this issue up first.

### Environment

  System:
    OS: Linux 4.18 Ubuntu 18.10 (Cosmic Cuttlefish)
    CPU: (4) x64 Intel(R) Core(TM) i5-7400 CPU @ 3.00GHz
    Shell: 4.4.19 - /bin/bash
  Binaries:
    Node: 11.7.0 - ~/.nvm/versions/node/v11.7.0/bin/node
    Yarn: 1.13.0 - /usr/bin/yarn
    npm: 6.9.0 - ~/.nvm/versions/node/v11.7.0/bin/npm
  Languages:
    Python: 2.7.15+ - /usr/bin/python
  Browsers:
    Chrome: 73.0.3683.86
    Firefox: 66.0.2
  npmPackages:
    gatsby: ^2.3.11 => 2.3.11 
    gatsby-remark-prismjs: ^3.2.6 => 3.2.6 
    gatsby-source-filesystem: ^2.0.28 => 2.0.28 
    gatsby-transformer-remark: ^2.3.8 => 2.3.8 
  npmGlobalPackages:
    gatsby-cli: 2.5.4

pieh:
Does prismjs has list of those aliases we could use to set default ones?

ojeytonwilliams:
There's a list of all their languages on their [site](https://prismjs.com/#supported-languages) and you can find them in [components.js](https://github.com/PrismJS/prism/blob/master/components.js).  I've not been able to find anything like `{shell: "bash", py: "python"...}`, unfortunately.

sidharthachatterjee:
Thank you for pointing this out @ojeytonwilliams 

[...]

PR #13127, opened 2019-04-04
feat(gatsby-remark-prismjs): load languages when referred to by an alias

sidharthachatterjee:
Currently, referring to a language by an alias (defined in https://github.com/PrismJS/prism/tree/master/components) doesn't work unless the language is already loaded. 

This PR adds first class support for aliases so that languages are loaded when they are referred to by their aliases as well. 

Fixes https://github.com/gatsbyjs/gatsby/issues/13119

### DIFF
--- a/packages/gatsby-remark-prismjs/src/__tests__/load-prism-language.js
+++ b/packages/gatsby-remark-prismjs/src/__tests__/load-prism-language.js
@@ -41,4 +41,32 @@ describe(`load prism language`, () => {
     expect(Prism.languages).toHaveProperty(requiredLanguage)
     expect(languagesAfterLoaded.length).toBe(languagesBeforeLoaded.length + 2)
   })
+
+  it(`loads languages when referred by their aliases (set as a String in components)`, () => {
+    const language = `python`
+    const languageAlias = `py`
+    const Prism = require(`prismjs`)
+
+    expect(Prism.languages).not.toHaveProperty(language)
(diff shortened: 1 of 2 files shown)


packages/gatsby-remark-prismjs/src/__tests__/load-prism-language.js
```diff
@@ -41,4 +41,32 @@ describe(`load prism language`, () => {
     expect(Prism.languages).toHaveProperty(requiredLanguage)
     expect(languagesAfterLoaded.length).toBe(languagesBeforeLoaded.length + 2)
   })
+
+  it(`loads languages when referred by their aliases (set as a String in components)`, () => {
```
sidharthachatterjee: Sorry about these test names 🤦‍♂️ 

If you think of better ones, please recommend!

sidharthachatterjee:
So tests are failing because

- yarn is resolving prismjs ^1.14.0 and ^1.15.0 dependencies to 1.15.0 as opposed to 1.16.0 (which is the latest)
- prismjs 1.15.0 doesn't have `py` included as an alias for `python` while 1.16.0 does

DSchau:
Published in gatsby-remark-prismjs@3.2.7

ojeytonwilliams:
Thanks, @sidharthachatterjee and you're very welcome, @DSchau!  It's great you could sort this out so quickly.

PR #13132, opened 2019-04-04
Link to Correct Location in Tutorial

dhurlburtusa:
Part 7 is the correct location for generating pages using data. Maybe it used to be part 4.


KyleAMathews:
It did get changed and we missed this link! Thanks!

PR #13139, opened 2019-04-04
feat(gatsby): Page data without compilation hash

Moocar:

## Description

This PR saves the `page-data.json` during query running. In order to minimize the size of PRs, my strategy is to save the page-data.json along with the normal query results, and then gradually shift functionality over to use `page-data.json` instead of `data.json`. Once all those PRs are merged, we'll be able to go back and delete the static query results, jsonNames, dataPaths, data.json etc.

It does mean that we'll be storing double the amount of query results on disk. Hopefully that's ok in the interim.

Compilation-hash will be added in future PRs.

## Related Issues

### DIFF
--- a/packages/gatsby/src/utils/page-data.js
+++ b/packages/gatsby/src/utils/page-data.js
@@ -1,0 +1,21 @@
+const fs = require(`fs-extra`)
+const path = require(`path`)
+
+const getFilePath = ({ publicDir }, pagePath) => {
+  const fixedPagePath = pagePath === `/` ? `index` : pagePath
+  return path.join(publicDir, `page-data`, fixedPagePath, `page-data.json`)
+}
+
+const write = async ({ publicDir }, page, result) => {
+  const filePath = getFilePath({ publicDir }, page.path)
+  const body = {
+    componentChunkName: page.componentChunkName,
+    path: page.path,
+    result,
+  }
+  await fs.outputFile(filePath, JSON.stringify(body))
+}
+
+module.exports = {
+  write,
+}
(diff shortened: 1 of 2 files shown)


Moocar:
ready for review

packages/gatsby/src/utils/page-data.js
```diff
@@ -0,0 +1,21 @@
...
+}
+
+const write = async ({ publicDir }, page, result) => {
+  const filePath = getFilePath({ publicDir }, page.path)
+  const body = {
+    componentChunkName: page.componentChunkName,
+    path: page.path,
+    result,
+  }
+  await fs.outputFile(filePath, JSON.stringify(body))
```
wardpeet: we should move to V8 serialize here too
https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/redux/persist.js
```js
const useV8 = Boolean(v8.serialize)
const [serialize, deserialize, file] = useV8
  ? [v8.serialize, v8.deserialize, `${process.cwd()}/.cache/redux.state`]
  : [jsonStringify, jsonParse, `${process.cwd()}/.cache/redux-state.json`]
```

It might be good to create a simple serialize utility that does the v8.serialize check.

Feel free to merge, as this is not blocking. Could be a follow up

Moocar: Just catching up on v8.serialize. It looks like it serializes using a non-JSON binary format. But the `page-data.json` needs to be loaded in the browser, so I don't think it will work for this use case.

wardpeet: oh yeah, that's true.

Moocar: huzzah!

PR #13147, opened 2019-04-05
feat(gatsby-source-wordpress): add the ability to set status argument on selected routes

maru3l:
## Description

add the ability to set status argument on selected routes in `gatsby-source-wordpress` plugin.

This will let authenticated user load `Draft` and other status in their query.


LekoArts:
Hi!

In the interest of moving this along most effectively, we strongly encourage our contributors to provide an [example project](https://gatsby.dev/example-project) which we can then use to validate the changes in this PR. This will allow us to quickly and easily ensure that this does indeed address the purported feature or fix in this PR!

An example project:

- can be run locally or credentials have been provided
	- Feel free to message on Twitter, Discord, etc.
- can be used to validate this functionality
	- In other words, the necessary functionality has been enabled/disabled in the project for us to easily validate these changes

💜

maru3l:
A exemple I can use is the website https://www.picki.ca .

On production, I only show the posts with the status `publish`.

On the develop branch `https://develop--picki.netlify.com/`, I will show all the `publish` and `draft` posts

I've made it as a list of routes and status because some route have different status type, like the media who by default une `inherit`. All the list of status by type can be found here: https://developer.wordpress.org/rest-api/reference/posts/

maru3l:
I've integrated it in this commit : https://github.com/maru3l/picki-www/commit/459d5484e140c3c00a50ca4eb799293676f0e3ed

you can see in gatsby/createAdvicesPage.js that I made a condition if the env is in `production\ and the post is in `draft` to not create the page, otherwise, create the page.

KyleAMathews:
So all this does is add another field `status` to the post nodes?

KyleAMathews:
If so, shouldn't we just get all statuses by default?

pieh:
Wordpress doesn't return entries with non `published` status by default (and requires authenticated requests on top of explicit query for all statuses) to return entries.

In any case, I will close this in favour of either https://github.com/gatsbyjs/gatsby/pull/10942 or https://github.com/gatsbyjs/gatsby/pull/10335 as those implement more generic API that allow setting query params (including status)

PR #13158, opened 2019-04-05
feat(gatsby-cli): validate rootPath to follow naming conventions as required

### LINKED ISSUE #13153: Validate directory name <rootPath> provided with gatsby new
## Description

`rootPath` name provided with `gatsby new` should be validated inorder to keep up with npm naming conventions. 
> There is actually an npm package inorder to perform the validation => `validate-npm-package-name`

### Steps to reproduce

Fire in `gatsby new *^#`

### Expected result

A suitable warning such as that the naming is not as per convention.

### Actual result

A directory is being created `<rootPath>` named as `*^#`

### Environment
```
System:
    OS: Linux 4.18 Ubuntu 18.10 (Cosmic Cuttlefish)
    CPU: (4) x64 Intel(R) Core(TM) i5-7200U CPU @ 2.50GHz
    Shell: 4.4.19 - /bin/bash
  Binaries:
    Node: 8.11.4 - /usr/bin/node
  Browsers:
    Chrome: 73.0.3683.75
    Firefox: 66.0.2
  npmGlobalPackages:
    gatsby-cli: 2.5.5
```

jamesgeorge007:
Closes #13153

## Description

Validates `rootPath` provided with `gatsby new` so that it accepts only valid names.


jamesgeorge007:
@pieh What are your thoughts?

wardpeet:
could you post an an example gatsby new command that breaks and one that passes.

pieh:
This will break `gatsby new .` - we can't check `validate-npm-package-name` like that. What we can do is determine final path where starter will be installed and validate that this path is valid - for example using https://www.npmjs.com/package/is-valid-path

jamesgeorge007:
@pieh Made the changes as you suggested :+1:

jamesgeorge007:
@KyleAMathews Thoughts :thinking:

packages/gatsby-cli/src/init-starter.js
```diff
@@ -154,6 +155,13 @@ module.exports = async (starter: string, options: InitOptions = {}) => {
     return
   }
 
+  if (!isValid(rootPath)) {
+    report.panic(
+      `Could not create a project called ${rootPath} because of npm naming restrictions`
```
pieh: I think the message need to be adjusted a bit - for example:
```suggestion
      `Could not create a project in ${rootPath} because it's not valid path`
```

jamesgeorge007: :+1:

jamesgeorge007:
@pieh Hope we're good to go :+1:

PR #13159, opened 2019-04-05
fix(gatsby-remark-prismjs): Update warnings to handle missing language

ojeytonwilliams:
## Motivation

At the moment, the plugin will correctly warn users if they try to use a language that is not supported by PrismJS.  However, the same type of message appears if no language has been supplied.

If their markdown contains either
>\`some code`

or

>\```
  some code block
\```

They get the warning `unable to find prism language 'text' for highlighting.`.  This is potentially confusing since it's unclear, from the outside, why the language `text` was being looked for.  This is particularly true for inline code as that's not often highlighted (at least in my experience).  The appearance of the above message can be seen in [this project](https://github.com/ojeytonwilliams/gatsby-remark-prismjs-bug) at `/shell/`.

## Description

This PR adds distinct error messages for when the user does not specify a language.  It uses the class postfix `-inline-text` to differentiate inline code from code blocks and generate the correct messages.

I did consider suppressing the messages for inline text, since it's pretty common to not try and highlight them. However, I wasn't 100% sure what the best approach was and I erred on the side of more warnings!  If anything seems off, let me know.

### DIFF
--- a/packages/gatsby-remark-prismjs/src/index.js
+++ b/packages/gatsby-remark-prismjs/src/index.js
@@ -79,7 +79,7 @@ module.exports = (
 
   if (!noInlineHighlight) {
     visit(markdownAST, `inlineCode`, node => {
-      let languageName = `text`
+      let languageName = `inline-text`
 
       if (inlineCodeMarker) {
         let [language, restOfValue] = node.value.split(`${inlineCodeMarker}`, 2)
--- a/packages/gatsby-remark-prismjs/src/index.js
+++ b/packages/gatsby-remark-prismjs/src/index.js
@@ -70,7 +70,7 @@ module.exports = (
     + `<div class="${highlightClassName}" data-language="${languageName}">`
     +   `<pre${numLinesStyle} class="${className}${numLinesClass}">`
     +     `<code class="${className}">`
-    +       `${highlightCode(languageName, node.value, highlightLines)}`
+    +       `${highlightCode(languageName, node.value, highlightLines, noInlineHighlight)}`
     +     `</code>`
     +     `${numLinesNumber}`
     +   `</pre>`
@@ -79,7 +79,7 @@ module.exports = (
 
   if (!noInlineHighlight) {
     visit(markdownAST, `inlineCode`, node => {
-      let languageName = `inline-text`
+      let languageName = `text`
 
       if (inlineCodeMarker) {
         let [language, restOfValue] = node.value.split(`${inlineCodeMarker}`, 2)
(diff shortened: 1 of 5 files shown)


ojeytonwilliams:
@DSchau I hope you don't mind me asking you to take a look at this.  It was kind of a follow up to #13119 and I thought I would try my hand at fixing it, rather than just flagging up another problem!

packages/gatsby-remark-prismjs/src/index.js
```diff
@@ -79,7 +79,7 @@ module.exports = (
 
   if (!noInlineHighlight) {
     visit(markdownAST, `inlineCode`, node => {
-      let languageName = `text`
+      let languageName = `inline-text`
```
wardpeet: It's probably a good idea to remove inline-text and replace it with text so we don't make a distinction. If people want to add their own custom language "text" they only have to do it once.

wardpeet: I don't think it's worth it, they actually choose for this behaviour so they are probably smart enough to look at all code blocks. You might want to edit your warning based on the noInlineHighlight flag to include inline-code blocks in your message.

ojeytonwilliams: @wardpeet I made those changes and a bit of extra test coverage to make sure the messages are tied to the noInlineHighlight flag.  I think this is a bit better.

ojeytonwilliams:
@sidharthachatterjee Thanks for sorting that out - I didn't realize it would be a problem.  Also, is there anything I can do to make ci/circleci: starters_validate pass?

sidharthachatterjee:
@ojeytonwilliams Looks like that was an `npm` failure (audit was failing) so unrelated. Publishing this now.

sidharthachatterjee:
Published in `gatsby-remark-prismjs@3.2.8`

ojeytonwilliams:
@sidharthachatterjee Glad to hear it.  Thanks again!

docwhat:
Hey! Thanks for fixing this.  I really appreciate it.

The original bug was mine; I added the `text` type because according to the PrismJS docs if the type wasn't found it would be treated as "generic text".

At the time, there was no warning about invalid types.

When I saw the `text` warning, I knew it was my bug but I couldn't figure out how to intelligently solve it.

I'm glad you figured it out!  The warning looks great and is informative.

ojeytonwilliams:
@docwhat No problem!  I don't think this was too serious, it was only a slightly puzzling warning after all.  Still, I'm happy I was able to make it a little easier to understand.

PR #13162, opened 2019-04-05
chore(gatsby-plugin-typography): cleanup tests

wardpeet:
## Description
It looks like lots of tests are failing because of the react addition in package.json. Seems like gatsby-plugin-typography was the biggest offender. We should add peer-deps to all our packages that have react & react-dom.

This pr fixes at least the flaky tests so the ink pr doesn't fail for no reason.

## Related Issues
#13089

### DIFF
--- a/packages/gatsby-plugin-typography/package.json
+++ b/packages/gatsby-plugin-typography/package.json
@@ -14,6 +14,8 @@
     "@babel/core": "^7.0.0",
     "babel-preset-gatsby-package": "^0.1.4",
     "cross-env": "^5.1.4",
+    "react": "^16.8.6",
+    "react-dom": "^16.8.6",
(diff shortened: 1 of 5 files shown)


wardpeet:
I might be wrong but I think @pieh ran a scrip to find missing deps to be compatible with yarn berry (pnp) which is probably something we want to do in the future.

packages/gatsby-plugin-typography/package.json
```diff
@@ -28,6 +30,8 @@
   "main": "index.js",
   "peerDependencies": {
     "gatsby": "^2.0.0",
+    "react": "^16.4.2",
```
wardpeet: these are equal to gatsby package

Issue #12947, opened 2019-03-29
Using Gatsby Image site uses Lorem Ipsum placeholder text


robphoenix:
## Summary

I noticed the site at https://using-gatsby-image.gatsbyjs.org/ currently uses Lorem Ipsum placeholder text, is there plans to update this to explanations of the different features of gatsby image? If so, is there any direction for what would make a useful PR to replace the current text?

sidharthachatterjee:
Thank you for opening this @robphoenix 

The home page at https://using-gatsby-image.gatsbyjs.org/ seems to have some documentation and links to https://www.gatsbyjs.org/packages/gatsby-image/ but the rest is placeholder text. 

Updating the placeholder text to some explanation of the features used in context will be helpful and I can definitely see the value in it!

Would you like to open a PR? The relevant files are in https://github.com/gatsbyjs/gatsby/tree/master/examples/using-gatsby-image/ and the lorem text is coming from https://github.com/gatsbyjs/gatsby/blob/master/examples/using-gatsby-image/src/components/lorem.js

CC @marcysutton

Issue #4319, opened 2018-03-01
Improve gatsby-image docs


shannonbux:
## Summary

There are several docs about how to use images in Gatsby and there is some confusion about what best practices are.

Here is a list of the docs that need to be edited to be clear and consistent:
https://www.gatsbyjs.org/docs/working-with-images/
https://www.gatsbyjs.org/docs/using-gatsby-image/
https://www.gatsbyjs.org/docs/adding-images-fonts-files/
https://www.gatsbyjs.org/packages/gatsby-image/?=gatsby-image

## Motivation
People get confused about when and how to use `gatsby-image`. Here are some notes from an interviewee about their struggles with `gatsby-image` docs if this helps

"
no mention of css background images.  there is discussion of how Img replaces img but would be great to also address if this can be used for css bg image.

no mention of which props are mandatory or optional

also, the props section... is it meant to be nested under the "sizes" queries section?  or does it apply to resolutions too?

in the example codebase, there seems to be a lot of "noise".  for example remove all unnecessary css.  no need for utils folder. need to really isolate the core teachings and learnings here.

theres no comments in the code.  im forced to hack thru it

remove all the lorem text.  that actually makes things super confusing.  instead replace it with text explaining what is exactly happening.  even me, a native english speaker, spent a moment or two trying to figure what the heck the text was referring to

https://www.gatsbyjs.org/packages/gatsby-image/#some-other-stuff-to-be-aware-of
this makes no sense.  this needs a "why".  why would someone want to set display none in this specific context. maybe its my own ignorance but from what i can tell, its because you want to load different images for different viewports?  if thats the case, then state it!  "if you want to load a different image for a different viewport then you will likely use the display none property.  this means you will also.... etc"

https://www.gatsbyjs.org/packages/gatsby-image/
please link to the example page.

https://www.gatsbyjs.org/packages/gatsby-image/#two-types-of-responsive-images
this section seems contradictory. why cant i have both?  what if i want retina images but also want those images to stretch across a fluid container.  also, are we sure we are using the term "thumbnail" correctly?  i usually associate that with a small image.  

https://www.gatsbyjs.org/packages/gatsby-image/#resolutions-queries
where's the full API?

is gatsby-image dependent on gatsby-transformer-sharp?  that is, do i need to `yarn add` both?  if so, why dont you simply roll up sharp into image?  

and when i look at the using-gatsby-image example, i see no mention of gatsby-image in the gatsby-config file

[...]

PR #13170, opened 2019-04-05
fix(docs): improve Gatsby Image docs

marcysutton:
## Description

As part of the Learning workflow project, this PR introduces some changes to the Gatsby Image docs to make them more clear and easier to follow.

## Related Issues

Closes https://github.com/gatsbyjs/gatsby/issues/4319
Closes https://github.com/gatsbyjs/gatsby/issues/12947


docs/docs/importing-assets-into-files.md
```diff
@@ -0,0 +1,97 @@
...
+
+Webpack finds all relative module references in CSS (they start with `./`) and replaces them with the final paths from the compiled bundle. If you make a typo or accidentally delete an important file, you will see a compilation error, just like when you import a non-existent JavaScript module. The final filenames in the compiled bundle are generated by Webpack from content hashes. If the file content changes in the future, Webpack will give it a different name in production so you don’t need to worry about long-term caching of assets.
+
+If you're using SCSS the imports are relative to the entry SCSS file.
+
+Please be advised that this is also a custom feature of Webpack.
+
+Two alternative ways of handling static assets are described in the next sections.
+
+## Querying for a `File` in GraphQL using gatsby-source-filesystem
```
muescha: maybe add a link to package gatsby-source-filesystem in the following paragraph...

docs/docs/static-folder.md
```diff
@@ -17,25 +17,14 @@ You can create a folder named `static` at the root of your project. Every file y
...
 ### Referencing your static asset
 
-In order to reference assets from the `static` folder in your code, you'll need to [import the `withPrefix` helper function from `gatsby`](/docs/gatsby-link/#prefixed-paths-helper):
+You can reference assets from the `static` folder in your code without anything special required:
 
 ```js
-import { withPrefix } from 'gatsby'
-
 render() {
   // Note: this is an escape hatch and should be used sparingly!
```
muescha: Ist this note still valid?

muescha: Or why we describing an escape hatch instead of an "normal import"

docs/docs/static-folder.md
```diff
@@ -17,25 +17,14 @@ You can create a folder named `static` at the root of your project. Every file y
...
   // Normally we recommend using `import` for getting asset URLs
   // as described in the “Adding Images, Fonts, and Files” page.
```
muescha: I remember this page removed by this pull request? Should we add an link to an replacing page in normal text

marcysutton: Which page?

docs/docs/static-folder.md
marcysutton: The language here is subjective, I chose to keep it in to describe it as something you should try to avoid rather than painting imports as "more normal" (normal usually isn't the correct word, as it means something different to everyone)

marcysutton:
If there aren't any comments on the `using-gatsby-image` docs, I think we should merge this in. We can always make improvements later!

PR #13180, opened 2019-04-06
fix(gatsby): move db.startAutosave after bootstrap in `gatsby develop`

Moocar:
## Description

In `gatsby develop`, we should start db autosaving **after** bootstrap.


sidharthachatterjee:
Published in `gatsby@2.3.15`

PR #13181, opened 2019-04-06
docs: fix broken link and typo in security-for-modern-web-frameworks

Addile:
The link to https://developer.mozilla.org/en-US/docs/Web/HTTP/Public_Key_Pinning had an extra asterisk at the start resulting it in being parsed as a relative link to a non-existent internal page.

Also '`rouge` third party' should be `rogue`.


wardpeet:
Seems like this issue already got resolved as the markdown file also got a different url.

PR #13184, opened 2019-04-07
fix(react-hooks-support): Changed the gatsby browser entry to not bre…

daniloster:
…ak hooks rules

## Description

Added fix to not break with react hooks (useMemo) inside `StaticQuery`. 🎉 
The file `packages/gatsby/cache-dir/gatsby-browser-entry.js` had conditionals in a way that violated the react hooks rules. This PR does not change any behaviour.

## Related Issues

I could not find an existing issue, but I am facing it with my blog project.


wardpeet:
Is it possible to give us an example where this breaks and this PR fixes it?

daniloster:
Sure thing, it happens when using useMemo + Suspense

https://github.com/daniloster/daniloster.github.io/blob/blog/src/templates/PagesTemplate.js

DSchau:
@daniloster quick question then, you're not using StaticQuery in that example--how exactly would this PR fix that?

Do you have an example of where _this PR change_ fixes an issue?

daniloster:
The file previously mentioned has `Page` as wrapper, then, `Page` is the one using the StaticQuery
https://github.com/daniloster/daniloster.github.io/blob/blog/src/components/Page.js

daniloster:
Sorry, I was on the train on my phone. It was difficult to send the whole scenario.
I believe that what we have now is the whole thing. Also, I have sent the incorrect link previously.

### Components
PostsRenderer: `https://github.com/daniloster/daniloster.github.io/blob/blog/src/components/PostsRenderer.js`
Page: `https://github.com/daniloster/daniloster.github.io/blob/blog/src/components/Page.js`
PostsRenderer has Page as a wrapper, then, Page is the one using the StaticQuery.

### Explanation

Indeed, the most external node will be the Page using the StaticQuery, as it wraps everything. 

I believe being following all rules to use hooks. Also, I checked this blog code a few times trying to find a slip/bug and I could not see an issue there (which does not mean that there isn't). Everything seems clear about the component rendering flow.

If someone could point me a way to work with this. It would be nice.

### Error message

The error I am getting is this:
Screen Shot 2019-04-09 at 21 01 08

### Resolution

As it is mentioned in the error, the StaticQuery inside `<project_folder>/.cache/gatsby-browser-entry.js:22` is likely to be breaking a hooks rule.
Used yarn to check duplicity of react as well as mismatch between React and ReactDOM. There were fine.

Based on the code line error, we can see that there is a conditional flow which may be affecting the hooks' lifecycle. (**we have an if/else inside a renderProps patterns in StaticQuery which delegates to StaticQueryContext.Consumer**)

Breaking the StaticQuery into 2 components helped me to make the render more predictable.

daniloster:
Please, ignore the Suspense case initially mentioned. I had this sorted, there were issue and docs that helped with this one.

Looking forward to hearing from you about the `useMemo` inside the `StaticQuery`.

danilosterrapid7:
Hi guys, I put the example. It is my own repo `https://github.com/daniloster/daniloster.github.io/blob/blog` on the blog branch.

sidharthachatterjee:
Published in `gatsby@2.8.8`

PR #13186, opened 2019-04-07
refactor(gatsby): Get list of /dev-404-page/ pages via graphql

Moocar:
## Description

The `gatsby develop` `/dev-404-page/` shows a list of pages on the site. This is currently sourced from `pages.json` which is written during bootstrap. https://github.com/gatsbyjs/gatsby/pull/13004 aims to remove reliance on global files such as `data.json`. This PR changes `/dev-404-page/` to source the list of pages directly from the running `develop` serve via a graphql query, and removes reliance on `pages.json`. 

Note, that this isn't absolutely necessary for https://github.com/gatsbyjs/gatsby/pull/13004, but it does move us closer to a world where the webpack build doesn't need to know about page data.

## Related Issues

### DIFF
--- a/packages/gatsby/cache-dir/app.js
+++ b/packages/gatsby/cache-dir/app.js
@@ -48,8 +48,10 @@ apiRunnerAsync(`onClientEntry`).then(() => {
 
   loader.addPagesArray(pages)
   loader.addDevRequires(syncRequires)
-
-  loader.getResourcesForPathname(window.location.pathname).then(() => {
+  Promise.all([
+    loader.getResourcesForPathname(`/dev-404-page/`),
+    loader.getResourcesForPathname(window.location.pathname),
+  ]).then(() => {
     const preferDefault = m => (m && m.default) || m
     let Root = preferDefault(require(`./root`))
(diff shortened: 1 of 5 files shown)


KyleAMathews:
Not in this PR but something we might want to do is move this to a searchable interface. E.g. show only say 100 pages by default and then people can search using glob or regex syntax.

As it is, this query and page don't scale up to very many pages.

pieh:
This is dev-only so for that we could use some client-side graphql request (i.e. https://www.npmjs.com/package/graphql-request seems really lightweight) against gatsby schema that uses `limit` + some filter.

Moocar:
Yep, I was thinking of limiting the results as I was working on this, but decided it was better left to another day. As it is, the page is only real useful for small sites.

KyleAMathews:
For simple client-side usages of GraphQL, fetch is all you need.

KyleAMathews:
Added an issue to track the follow-up work https://github.com/gatsbyjs/gatsby/issues/13204

Moocar:
Published in gatsby@2.3.24

PR #13197, opened 2019-04-07
fix(gatsby): fix not focusing router wrapper after navigation

pieh:
This is work in progress to assert that focus management is working as expected on navigation.

Right now we are misusing `@reach/router` and don't take full advantage of its focus management feature because we have <Router> with single catch-all route - which means that this route never change and focus is never triggered.

Initially I'll try to push just failing test (to verify correctness), with actual changes following that


e2e-tests/production-runtime/cypress/support/commands.js
```diff
@@ -33,3 +33,7 @@ Cypress.Commands.add(`shouldNotMatchScrollPosition`, id =>
     expect(scrollPosition).not.to.be.closeTo(storedScrollPositions[id], 100)
   })
 )
+
+Cypress.Commands.add(`assertRouterWrapperFocus`, () =>
```
marcysutton: This is dope! 👍

packages/gatsby/src/utils/webpack.config.js
```diff
@@ -343,6 +343,19 @@ module.exports = async (
         PnpWebpackPlugin,
       ],
     }
+
+    const target =
+      stage === `build-html` || stage === `develop-html` ? `node` : `web`
+    if (target === `web`) {
+      // force to use es modules when importing internals of @reach.router
```
marcysutton: Should we open an issue on @reach/router to make any changes?

pieh: We do import internal `match` function (to match those parameterized / wildcarderded paths like `/app/:someParam` or `/app/*`). If `@reach/router` exported that from entry point, we wouldn't need this part, but I'm not sure if Ryan would really want that - as this is lower level function. Worth asking at least ;)

pieh:
finally got unblocked on this, next step here is to adjust `gatsby develop` to use similar technique (right now only production builds are handled)

packages/gatsby/cache-dir/ensure-resources.js
```diff
@@ -131,8 +131,13 @@ class EnsureResources extends React.Component {
...
 
-      // prevent hydrating
-      throw new Error(`Missing resources for ${this.state.location.pathname}`)
+      console.error(`Missing resources for ${this.state.location.pathname}`)
+      return (
+        <div
+          dangerouslySetInnerHTML={{ __html: `` }}
+          suppressHydrationWarning
+        />
+      )
```
pieh: @davidbailey00 I would like sanity check here:
this seems to do similar thing (preventing react from cleaning up `___gatsby` div) and it works in more scenarios than throwing (in case we throw before any parent component "renders" any DOM element, react will wipe `___gatsby` div). It seems like we were "lucky" with this, because <div> was always there coming from <Router> above <EnsureResources>. My change moves <Router> below so no DOM element is above, which was causing issues.

packages/gatsby/src/utils/webpack.config.js
wardpeet: @pieh are you sure we need this? https://github.com/reach/router/blob/master/package.json#L6 should take care of it.

packages/gatsby/cache-dir/ensure-resources.js
wardpeet: this means we will do a full client rehydration if something goes wrong, right? maybe add a comment about it?

wardpeet:
so the fix is actually bringing all routes under the <Router> as separate pages? This way Reach router knows about all paths and can do it's magic tricks?

pieh:
> so the fix is actually bringing all routes under the as separate pages? This way Reach router knows about all paths and can do its magic tricks?

We still have one route, but it's dynamic - see `<PageRenderer>` `path` prop under `<Router>`. This should make sure that `@reach/router` detect that route has actually changed and will properly trigger focusing router's wrapper div (which should tell assistive tech that something happened on navigation)

packages/gatsby/cache-dir/ensure-resources.js
pieh: It's partial hydration - things above the actual page will be hydrated - so if users use `wrapRootElement`/`wrapPageElement` or `gatsby-plugin-layout` those will be actually hydrated, but the content of the page itself (in case template chunk or query results can't be loaded) won't be trully hydrated, but content from static html will be there

This is a neat trick that you can do when using:
```
<div
          dangerouslySetInnerHTML={{ __html: `` }}
          suppressHydrationWarning
        />
```
during hydration, as it won't try to actually change DOM nodes under it.

But yeah, I should add more context there for sure

pieh:
This should be ready to review. Here's demo with voiceover for production build https://www.dropbox.com/s/tbrt2uczih2jolk/Kapture%202019-05-18%20at%2017.21.45.mp4?dl=0 - page content is being announced properly (I think). To live check it, you can use for example https://deploy-preview-13197--using-remark.netlify.com/ which is build with changes from this PR

Notes:
- There is problematic scenario that I didn't fix in development (see skipped tests in development-runtime). Problematic scenario is navigating in development from or to 404 page - this is because 404 in development is pretty special case and we have 2 code paths that it's pretty difficult to properly use router there. This code may also significantly change for page manifest changes so I chose to not spend too much time on it and focus on shipping this as this fixes majority of issues with focus management
- I duplicated some code in production and runtime e2e tests - not sure if extra effort to keep it DRY is worth the time investment? It might be more worth if we have more duplicated tests across those two?

e2e-tests/development-runtime/cypress/support/commands.js
```diff
@@ -10,3 +10,40 @@ Cypress.Commands.add(`lifecycleCallCount`, action =>
         ).length
     )
 )
+
+Cypress.Commands.add(`assertRouterWrapperFocus`, (shouldBeFocused = true) =>
```
marcysutton: These are so awesome!! 🌮

marcysutton:
Thanks again for working on this @pieh! It looks like the keyboard mechanics are working, and it's an improvement from what I'm seeing on the .org site. I also did some screen reader testing, though, and I'm finding in many cases that nothing is announced. Chrome and Voiceover are pretty much the only combination that works; Safari and Voiceover doesn't announce anything much of the time, and it's inconsistent which pages are actually announced. On Windows, I don't get any output from NVDA and Firefox at all (need to download JAWS again to test on my VM).

I'll keep playing with it, but there are two things I'll focus on in trying to get this to work:
- the timing of the focus call as compared to page content loading - it could be that the accessibility tree isn't ready yet when the call happens
- the semantics of the focus wrapper: it has `role="group"`, may need something else to get assistive tech to play ball. (I'll experiment with this)

pieh:
> * the timing of the focus call as compared to page content loading - it could be that the accessibility tree isn't ready yet when the call happens

I will do some testing. Is accessibility tree something I can check from js? For example to make snapshot of it/console.log when the router div is being focused?

> * the semantics of the focus wrapper: it has `role="group"`, may need something else to get assistive tech to play ball. (I'll experiment with this)

This is coming from `@reach/router` - we might be able to override it, but if this attribute is not suitable, it would be better handled upstream I think

pieh:
> * the timing of the focus call as compared to page content loading - it could be that the accessibility tree isn't ready yet when the call happens

I wasn't able to find much about accessibility tree (in sense of programmatic checks), one thing I could verify is that `focus()` on router div wrapper is happening when DOM is updated with new page content, so I was hoping that updating DOM would synchronously update accessibility tree as well.

> Chrome and Voiceover are pretty much the only combination that works; Safari and Voiceover doesn't announce anything much of the time, and it's inconsistent which pages are actually announced. On Windows, I don't get any output from NVDA and Firefox at all (need to download JAWS again to test on my VM).

I was testing just with chrome and voiceover, but I definitely can reproduce issues in Safari. It's pretty weird that it doesn't seem to announce some pages (focusing router div wrapper is consistent, and I have no clue what else can I check here :( ).

> > * the semantics of the focus wrapper: it has `role="group"`, may need something else to get assistive tech to play ball. (I'll experiment with this)
> 
> This is coming from `@reach/router` - we might be able to override it, but if this attribute is not suitable, it would be better handled upstream I think

Update on this, I checked and we should be able to change the role="group" ( https://github.com/reach/router/blob/4cd2e31297f2293ba701a6f1146fb55a29a26c26/src/index.js#L342 ), but I'm out of my depth here

marcysutton:
So I've been testing this some more (and doing prototype testing with users)...and focus management doesn't seem to be working with the keyboard at all in this branch, even with Voiceover turned off. When I click on one of the main nav links (Docs, Tutorials, etc.), focus is still set to the top of the page and I have to start over with the TAB key. I'm not seeing the new wrapper ID in the rendered markup - should that be working? I checked and double checked my git remote + branch matched yours, and cleared the cache...still no go.

Steps to reproduce:
- Go to the www site landing page locally
- Hit the TAB key until you hit "Docs" or "Tutorials"
- After the page refreshes, notice your focus point is back at the top of the page, before the Skip to Main Content link. It should be on the `<main>` wrapper element.

If you're seeing something different in any browser, let me know!

As for the accessibility tree, there aren't ways to interact with it directly...it's entirely dependent on the DOM currently. But I think if we sort out the base keyboard issue, hopefully we'll get it sorted out without hacky workarounds.

pieh:
It seems like something went wrong when building gatsbyjs.org with those changes. Here's link for preview that make use of changes from this PR - https://5cf57f3314a67857eb443b8d--gatsby-org-pieh-previews.netlify.com/

marcysutton:
Ok, I think I see what's going on now....the focus wrapper element taking focus is pretty high up in the DOM tree. When you click on a link like "Docs" or "Tutorials", focus is reset back to that parent DOM element high up the tree, which I've discovered through user testing isn't ideal because it forces keyboard and screen reader users to navigate through the entire page again. (You can reproduce this by clicking on one of those links and hitting TAB–you'll notice that you have to start over from the top.) That's what the live site is doing, essentially.

A better pattern would be to send focus to the _new_ and relevant content, like an element wrapping the main body (in the case of the docs, would be the `.main-body` div that includes the sidebar `<nav>` and `<main>`). The router should know about the content that's changing, is it possible to send focus there instead? In user tests, sending focus to the new content's wrapper element worked okay and I think that's our best bet, since we can't guarantee it will contain a h1-h6 heading we can send focus to instead. Does that make sense?

pieh:
> A better pattern would be to send focus to the _new_ and relevant content, like an element wrapping the main body (in the case of the docs, would be the `.main-body` div that includes the sidebar `<nav>` and `<main>`). The router should know about the content that's changing, is it possible to send focus there instead? In user tests, sending focus to the new content's wrapper element worked okay and I think that's our best bet, since we can't guarantee it will contain a h1-h6 heading we can send focus to instead. Does that make sense?

It's not feasible to do it like that. `.main-body` etc would need to be handled by user code and we don't have way to ensure it exists (also would be breaking change). Also we wouldn't really be able to use `@reach/router` for this I think.

In any case this (to me) seems a bit out of scope for this PR - as purpose of this was to ensure router wrapper consistently get focused on navigation (and waiting for more feature/fixes will really stall this PR which is already pretty old)

KyleAMathews:
A pattern I've seen is adding a react hook for Gatsby to set the focus on what/where makes sense.

pieh:
> A pattern I've seen is adding a react hook for Gatsby to set the focus on what/where makes sense.

We can add this, but we would need to completely disable `@reach/router` focus management and handle this on our own (as with `@reach/router` it can handle just focusing router wrapper element).

In any case, this is really not what I tried to do with this PR and if changes I did don't improve anything I should just close this

pieh:
We chatted with Marcy about this and plan is to merge this PR (after resolving conflicts) and then do research on how could we implement further improvements that would focus just content area.

pieh:
Merge conflicts after `per-page-manifest` merge are resolved and this is ready for review

e2e-tests/production-runtime/src/pages/paths.js
```diff
@@ -7,7 +7,7 @@ const NotFound = () => <h1>Not Found in App</h1>
 
 const AppPage = () => (
   <Router>
-    <App path="/paths/app" />
```
sidharthachatterjee: Sweet!

sidharthachatterjee:
Published in `gatsby@2.13.2`

Issue #11490, opened 2019-02-01
[docs] restructure styling section


robinmetral:
In the [styling section of the docs](https://www.gatsbyjs.org/docs/styling/), there are guides about different styling methods (CSS Modules, styled-components, etc.), followed by a general article about creating global styles.

Since creating global styles varies depending on the styling method, I suggest to split this article up into the relevant guides. For example, [creating Global Styles with Emotion](https://www.gatsbyjs.org/docs/creating-global-styles/#how-to-add-global-styles-in-gatsby-using-css-in-js) would go into the [Emotion guide](https://www.gatsbyjs.org/docs/emotion/).

If you also think this makes sense, I'll submit a PR :tada: 
I'll also add some missing info about creating global styles with styled-components.

marcysutton:
Hi @robinmetral, thanks for the suggestion! I agree with you, as I did not expect docs on global styles in Emotion to be in the general article about global CSS styles. I want to check with @shannonbux, though; do you have any insight into the current structure of the CSS Styling guides? Were those created based in research, and/or can we move things around a bit?

robinmetral:
Thanks for your message @marcysutton!

I'm not sure how the current guides were created, but I don't think it would be a problem of reorganizing them a little :slightly_smiling_face: (in fact wouldn't they be all the more useful?)

Good idea to check with @shannonbux anyway, I'll get started when the confirmation comes in :rocket:

shannonbux:
Hi @marcysutton! The current organization, titles, and content of style
docs mostly existed before I got here, so I'm not sure how they came to be.
The main thing based on research I've done are the general top-level
categories and the basic idea that each doc should be as standalone as
possible, so should cover one workflow or task and then link to related
pages.

marcysutton:
Sorry for the delay on this, I'd say go ahead with the restructuring! I look forward to seeing what you come up with.

robinmetral:
Thanks @marcysutton, I will :slightly_smiling_face:

marcysutton:
Hey @robinmetral, how's it going? There've been some comments about the CSS docs needing attention, and your issue came to mind. https://github.com/gatsbyjs/gatsby/issues/5997

I'm looking at these docs in the current sprint, but I wanted to check with you to see if you're still planning on contributing! Let me know if you want to pair on anything.

robinmetral:
Hi @marcysutton, I've been a bit busy these last few weeks, but I was planning on getting to work on this starting this coming week!

[...]

Issue #5997, opened 2018-06-18
[v2] Document how to use global CSS


mquandalle:
Gatsby v2 removes the global layout and we now have to wrap every component into a normal React component, let's say from `src/components/layout.js`.

From this file I include some stylesheets using something like `import "./design.css"`. The problem is that every time a page is loaded (for instance using `window.___loader.hovering` when hovering a link), the whole CSS file is reloaded (even if it's already there) including linked resources such as images and fonts. This causes a page blink.

What is the Gatsby v2 way, now that the global layout is gone, to include a global stylesheet that isn't re-downloaded when a new page is loaded?

KyleAMathews:
The best place to load global styles now is in your gatsby-browser.js as that's loaded at the root of your site.

We need to document that in the upgrade guide. Please try out this change and if it feels good to everyone, we can close this issue by updating the guide.

mquandalle:
I spent a lot of time trying to solve this apparently simple issue and I didn't think of `gatsby-browser.js`, happy to find there is such a simple solution! But you're right that this definitely needs to be documented.

So apparently adding the `import` line isn't enough for the file to be executed in the browser (and the style to be loaded) and we need to add a stub like `exports.onClientEntry = () => {};`. If the goal is to have `gatsby-browser.js` be a `main.js` for the client maybe this behavior should be changed?

KyleAMathews:
Oh... hmm that's odd. I'll look into this soon-ish then.

KyleAMathews:
Also welcome anyone else who wants to investigate why gatsby-browser.js isn't loading css

chasemccoy:
@KyleAMathews I am also being affected by this. My fonts are loading many times and flashing on load in production. 

@mquandalle did you find a way to get this to work in `gatsby-browser.js`? You mentioned a stub.

chasemccoy:
@mquandalle Actually, I was just able to get this to work. Putting this (and only this) in `gatsby-browser.js` worked for me and loaded all of the appropriate packages and files:

```
require("./src/styles/index.scss");
require("./src/fonts/fonts.css");
require("svgxuse");
```

ahlusar1989:
@KyleAMathews Would you like someone to look into this?

KyleAMathews:
That'd be great!

KyleAMathews:
[...]

PR #13199, opened 2019-04-07
[docs] Restructure styling section

robinmetral:
## Description

See #11490 for context and discussion.

## Related Issues

#11490 
#5997


DSchau:
@robinmetral thanks for opening this! I do think an issue is perhaps a better choice of discussion for this rather than a PR--mostly because I don't feel we're as of yet in a state to begin working on a PR with the information we currently have.

I'm going to close this so we can continue the discussion in #11490, and I'll add some thoughts there relating to what you brought up _here_.

Hope to see this PR very soon, but first I think we need to ensure we're on the same page re: styling!

robinmetral:
Makes sense @DSchau, let's move it over there

PR #13208, opened 2019-04-08
Fix #13202 - Remove render-page.js and render-page.js.map from ./public

### LINKED ISSUE #13202: Why Gatsby 2.3.11 and above started adding render-page.js and render-page.js.map to ./public
## Summary

I noticed that after Gatsby@2.3.11 two new files get added to the ./public folder on build. The 2 new files are `render-page.js` and `render-page.js.map`.
Is there any particular reason they are added there because I think that they are not used by the application after it's been built?

I digged up a little bit and I think 1534fe52 causes this to happen.

Can these 2 new files be safely removed from ./public?

LubomirGeorgiev:
## Description
I created #13202 asking why after version 2.3.11 render-page.js and render-page.js.map are added to ./public. After looking at [this commit](https://github.com/gatsbyjs/gatsby/commit/1534fe529a170d1e4f0ae6013e97ce5255c0c4d0#diff-87b9cc88fc16a731c1acb2ac60723f0aL52) I realized that
```
fs.unlinkSync(outputFile)
fs.unlinkSync(`${outputFile}.map`)
```
Got changed into:
```
await fs.unlink(rendererPath)
await fs.unlink(`${rendererPath}.map`)
```

So this is what prevented the files from being deleted after the build process has finished. To confirm this I console logged the error in the catch block [right here](https://github.com/gatsbyjs/gatsby/blob/1534fe529a170d1e4f0ae6013e97ce5255c0c4d0/packages/gatsby/src/commands/build-html.js#L49-L51) and this is the result.

gatsby-error

After that I reverted it back to `fs.unlinkSync(rendererPath)` without await and it got fixed.

## Related Issues and commits
#13202 and [1534fe52](https://github.com/gatsbyjs/gatsby/commit/1534fe529a170d1e4f0ae6013e97ce5255c0c4d0#diff-87b9cc88fc16a731c1acb2ac60723f0aL52)

### DIFF
--- a/packages/gatsby/src/commands/build-html.js
+++ b/packages/gatsby/src/commands/build-html.js
@@ -44,8 +44,8 @@ const buildRenderer = async (program, stage) => {
 
 const deleteRenderer = async rendererPath => {
   try {
-    await fs.unlink(rendererPath)
-    await fs.unlink(`${rendererPath}.map`)
+    fs.unlinkSync(rendererPath)
+    fs.unlinkSync(`${rendererPath}.map`)
   } catch (e) {
     // This function will fail on Windows with no further consequences.
   }


wardpeet:
Oops seems like @sidharthachatterjee also made a PR #13209.

We would like to keep things async as it has better support on windows so I'll be closing this one in favor of #13209 

LubomirGeorgiev:
@wardpeet yeah I agree that @sidharthachatterjee 's solutions is better.

Issue #13188, opened 2019-04-07
(node:14763) [DEP0013] DeprecationWarning: Calling an asynchronous function without callback is deprecated.


mskian:
Gatsby Version - `2.3.14`

```
$ node --trace-deprecation node_modules/gatsby/dist/bin/gatsby build
```

```
success open and validate gatsby-configs — 0.083 s
success load plugins — 8.263 s
success onPreInit — 0.185 s
success delete html and css files from previous builds — 2.661 s
success initialize cache — 0.008 s
success copy gatsby files — 0.279 s
success onPreBootstrap — 0.024 s
success source and transform nodes — 0.568 s
success building schema — 0.741 s
success createPages — 0.088 s
success createPagesStatefully — 0.041 s
success onPreExtractQueries — 0.000 s
success update schema — 0.057 s
success extract queries from components — 0.204 s
success run graphql queries — 0.020 s — 5/5 264.06 queries/second
success write out page data — 0.003 s
success write out redirect data — 0.001 s
success onPostBootstrap — 0.143 s

info bootstrap finished - 26.843 s

success Building production JavaScript and CSS bundles — 13.609 s
success Building static HTML for pages — 3.749 s — 16/16 13.02 pages/second
(node:14763) [DEP0013] DeprecationWarning: Calling an asynchronous function without callback is deprecated.
    at makeCallback (fs.js:127:12)
    at Object.fs.unlink (fs.js:1049:14)
    at /home/groot/Santhosh/gatsby/hello-gatsby/node_modules/gatsby/dist/commands/build-html.js:72:16
    at Generator.next (<anonymous>)
    at asyncGeneratorStep (/home/groot/Santhosh/gatsby/hello-gatsby/node_modules/@babel/runtime/helpers/asyncToGenerator.js:3:24)
    at _next (/home/groot/Santhosh/gatsby/hello-gatsby/node_modules/@babel/runtime/helpers/asyncToGenerator.js:25:9)
    at /home/groot/Santhosh/gatsby/hello-gatsby/node_modules/@babel/runtime/helpers/asyncToGenerator.js:32:7
    at new Promise (<anonymous>)
    at /home/groot/Santhosh/gatsby/hello-gatsby/node_modules/@babel/runtime/helpers/asyncToGenerator.js:21:12
    at deleteRenderer (/home/groot/Santhosh/gatsby/hello-gatsby/node_modules/gatsby/dist/commands/build-html.js:79:18)
[...]

PR #13209, opened 2019-04-08
chore(gatsby): Get rid of "DeprecationWarning: asynchronous function without callback"

sidharthachatterjee:
Before:
Screenshot 2019-04-08 16 05 50

After:
Screenshot 2019-04-08 16 11 17

Fixes https://github.com/gatsbyjs/gatsby/issues/13188

### DIFF
--- a/packages/gatsby/src/commands/build-html.js
+++ b/packages/gatsby/src/commands/build-html.js
@@ -1,6 +1,8 @@
 /* @flow */
 const webpack = require(`webpack`)
 const fs = require(`fs`)
+const { promisify } = require(`util`)
+const promisifiedUnlink = promisify(fs.unlink)
 
 const webpackConfig = require(`../utils/webpack.config`)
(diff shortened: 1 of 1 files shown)


packages/gatsby/src/commands/build-html.js
```diff
@@ -1,6 +1,8 @@
 /* @flow */
 const webpack = require(`webpack`)
 const fs = require(`fs`)
+const { promisify } = require(`util`)
+const promisifiedUnlink = promisify(fs.unlink)
```
wardpeet: honestly I would move to fs-extra and use remove insetad of the promise dance

sidharthachatterjee: Good point since we already have fs-extra. Done!

sidharthachatterjee:
Published in `gatsby@2.3.15`

PR #13214, opened 2019-04-08
fix(gatsby-source-graphql): Destructure createContentDigest from first parameter

frinyvonnick:
## Description

It fixes a wrong import of createContentDigest that breaks gatsby-source-graphql since 2.0.16

## Related Issues

Mistake made in a refactoring (#13000)


frinyvonnick:
Sorry for making the mistake on first PR 😕

DSchau:
@frinyvonnick re:

> Sorry for making the mistake on first PR 🙏

Not at all, and nothing to apologize for! We should've caught that--sorry that we didn't! I've added some unit tests to catch this type of thing in the future, but thanks for identifying that there's an issue _and_ fixing it.

Thank you!

DSchau:
Successfully published:
 - gatsby-source-graphql@2.0.17

Issue #13201, opened 2019-04-07
Add support for native lazy loading to gatsby-image


KyleAMathews:
https://twitter.com/addyosmani/status/1114777583302799360

We'll want to feature detect this using the code in [Addy's blog post](https://addyosmani.com/blog/lazy-loading/):

```jsx
if ('loading' in HTMLImageElement.prototype) { 
    // Browser supports `loading`, don't setup IntersectionObserver
} else {
   // Use existing IntersectionObserver code
}
```

If lazy-loading exists, then we'll want to use it, if not, we'll continue to use our IntersectionObserver setup.

For images using the `critical` prop, we should use `loading="eager"` and for other images, we should use `loading="lazy"` — this should replicate our existing semantics.

```html
<!-- Lazy-load an offscreen image when the user scrolls near it -->
<img src="unicorn.jpg" loading="lazy" alt=".."/>

<!-- Load an image right away instead of lazy-loading -->
<img src="unicorn.jpg" loading="eager" alt=".."/>
```

addyosmani:
Excited to see the loading attribute being considered. Initially, those mappings for lazy vs eager LGTM.

andrewfairlie:
I would like to have a go at this, it would be my first contribution though so I'm unsure what the procedure is in taking an issue?

sidharthachatterjee:
Hey @andrewfairlie 

Thank you for offering to take this up! We really appreciate it and this would be an awesome first contribution to Gatsby. Would you like to pair on this sometime tomorrow and get this in?

andrewfairlie:
@siddharthkp 👍my schedule's a little wonky this week with a fair bit of travel. If it's cool with you I'll fork this when I have an hour or two and if I don't make much progress could we then arrange a pair session to get me on the right lines?

Really appreciate the offer of pairing, very cool! 👍

KyleAMathews:
@andrewfairlie we want to get this in pretty quickly so if you're not available this week, @sidharthachatterjee might just take it up directly. There's plenty of other issues tagged with https://github.com/gatsbyjs/gatsby/labels/help%20wanted so please help out when you can! 

We also offer pair programming sessions you can sign up for https://www.gatsbyjs.org/contributing/pair-programming/

andrewfairlie:
That makes sense to me. I'll keep an eye on this and see how the problem was solved and will take another issue when I have the time.

This will be a great feature 🤘

polarathene:
[...]

PR #13217, opened 2019-04-08
feat(gatsby-image): Add support for native lazy loading

sidharthachatterjee:
So... Chrome just added support for native lazy loading for images and iframes (https://twitter.com/addyosmani/status/1114777583302799360) 

This PR adds support in `gatsby-image` so that we use native lazy loading when available and fall back to our IntersectionObserver code when not. 

### Test
`yarn add gatsby-image@loading-attribute`

### Notes

I've exposed the `loading` attribute as a prop but just realised that we'll want to set that based on our `critical` prop as @KyleAMathews noted in https://github.com/gatsbyjs/gatsby/issues/13201

Another thing to note is that we _always_ set this for the no script Img tag (rendered during SSR and no js) since it's a progressive enhancement and setting it does not break anything in browsers that are yet to support it

### Todo
- [x] Add some snapshot tests
- [x] Update README and documentation for gatsby-image
- [ ] Blog post
- [x] Deprecate `critical` prop and set `loading` based on it
- [ ] Update using-gatsby examples

### Related

- Chromium: https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/jxiJvQc-gVg
- Addy's blog post: https://addyosmani.com/blog/lazy-loading/
- Fixes https://github.com/gatsbyjs/gatsby/issues/13201


packages/gatsby-image/src/index.js
```diff
@@ -100,11 +100,34 @@ const noscriptImg = props => {
...
 
-  return `<picture>${srcSetWebp}<img ${width}${height}${sizes}${srcSet}${src}${alt}${title}${crossOrigin}style="position:absolute;top:0;left:0;opacity:1;width:100%;height:100%;object-fit:cover;object-position:center"/></picture>`
+  // Since we're in the noscript block for this image (which is rendered during SSR or when js is disabled),
+  // we have no way to "detect" if native lazy loading is supported by the user's browser (at least not today — See 1)
+  // Since this attribute is a progressive enhancement, it won't break a browser with no support
+  // Therefore setting it by default is a good idea.
+
+  // 1. In the future, a Client Hint for this might be available (https://bugs.chromium.org/p/chromium/issues/detail?id=949365)
+
+  const loading = props.loading ? `loading="${props.loading}" ` : ``
```
KyleAMathews: we should default `loading` to lazy since that's our existing behavior

packages/gatsby-image/src/index.js
```diff
@@ -160,6 +185,17 @@ class Image extends React.Component {
...
     }
 
+    // Chrome Canary 75 added native lazy loading support!
+    // https://addyosmani.com/blog/lazy-loading/
+    if (
+      typeof HTMLImageElement !== `undefined` &&
+      `loading` in HTMLImageElement.prototype
+    ) {
+      // Setting isVisible to true to short circuit our IO code and let the browser do its magic
+      isVisible = true
```
KyleAMathews: maybe to make it clear this is now representing two different states it could be changed to `isVisibleOrNativeLazyLoadingSupported` — verbose is effective :-D

packages/gatsby-image/src/index.js
KyleAMathews: Oh, looks like below we do default to `lazy`? In that case, we don't need to check if the prop is set.

packages/gatsby-image/src/index.js
```diff
@@ -160,6 +185,17 @@ class Image extends React.Component {
       IOSupported = true
     }
 
+    // Chrome Canary 75 added native lazy loading support!
+    // https://addyosmani.com/blog/lazy-loading/
+    if (
+      typeof HTMLImageElement !== `undefined` &&
```
polarathene: You could set this immediately after declaring `nativeLazyLoadSupported` (or use the value of this to define that variable).

```JS
// Extracted feature detection to vars to avoid the interleaving logic flipping booleans, especially `isVisible`
// Native lazy-loading support: https://addyosmani.com/blog/lazy-loading/
const hasNativeLazyLoadSupport = (
  typeof HTMLImageElement !== `undefined` &&
  `loading` in HTMLImageElement.prototype
)

const hasIOSupport = (
  typeof window !== `undefined` &&
  window.IntersectionObserver
)

// Renamed `hasNoScript` and `IOSupport` vars to have clearer intents
const isBrowser = typeof window !== `undefined`
const addNoScript = !(props.critical && !props.fadeIn)
const useIOSupport = !props.critical && !seenBefore && hasIOSupport

// Condensed all the prior logic for setting `isVisible`, ordered by override priority
// Probably isn't that readable and would be better off verbose?
const isVisible = (
  props.critical || 
  isBrowser && (hasNativeLazyLoadSupport || !useIOSupport)
)
```

Too terse? I think it's a bit clearer regarding the interplay vs the conditional statements flipping booleans, but extensive short-circuit via logical operators might counter readability benefits, especially for `isVisible`?

packages/gatsby-image/src/index.js
```diff
@@ -207,6 +244,10 @@ class Image extends React.Component {
   }
 
   handleRef(ref) {
+    if (this.state.nativeLazyLoadSupported) {
+      // Bail because the browser natively supports lazy loading
+      return
+    }
```
polarathene: This ignores the logic below that updates the `imgLoaded` and `imgCached` state after the image component is mounted(and thus the ref fires to make the actual image visible(`isVisible`) and begin loading).

I understand that `isVisible` is being set to `true` earlier, but `imgLoaded` needs to be set to reveal the image when it's finished loading in with a fade-in effect(unless marked critical or fade-in disabled). `imgCached` additionally avoids the transition if there's no loading delay.

Perhaps `imgCached` needs to be added to `handleImageLoaded()`? When I implemented `imgCached` it didn't seem necessary but I hadn't tested with other configurations such as with Interaction Observer disabled. This is the only other method it seems `imgLoaded` is set.

packages/gatsby-image/src/index.js
```diff
@@ -100,11 +100,34 @@ const noscriptImg = props => {
...
+    style,
+    onLoad,
+    onError,
+    nativeLazyLoadSupported,
+    loading,
+    ...otherProps
+  } = props
+
+  const loadingAttribute = {
+    ...(nativeLazyLoadSupported && loading),
```
polarathene: Is `nativeLazyLoadSupported` boolean actually useful here?

This doesn't seem like the right approach for handling this single attribute. You're defining a const obj with a potential single prop that if the boolean prop is true, adopts the `loading` var as a key/prop to it's contained string value?

The destructuring approach here and below for this seems to harm readability. Why not instead go with:

```JS
loading={loading || `auto`}
// or conditionally, if the prop isn't specified, the attribute won't be assigned
loading={loading}
```

If the browser supports this attribute, `auto` is applied by default implicitly, there isn't a need to check for native support to assign it, makes no difference.

packages/gatsby-image/src/index.js
polarathene: @KyleAMathews isn't the `isVisible` state here linked to the image components visibility state for it's picture element? I suggested a refactor [here](https://github.com/gatsbyjs/gatsby/pull/13217#discussion_r273377916), but perhaps that's the wrong approach.

packages/gatsby-image/src/index.js
polarathene: If the attribute isn't set, and a browser supports native lazy loading, they default to `auto` which will use `eager` or `lazy` strategies based on distance from viewport afaik?

We can also drop this noscript string approach if min version of React for Gatsby can be 16.5+ which has fixed the issue that required changing to this approach for noscript support.

packages/gatsby-image/src/index.js
polarathene:
```JS
// Visible if `critical` or is in a browser with native lazy-load support, or no IntersectionObserver support available. Not visible/rendered during SSR(unless `critical`), or when lazy-loading with IntersectionObserver in the browser.
const isVisible = (
  props.critical || 
  isBrowser && (hasNativeLazyLoadSupport || !useIOSupport)
)
```

Perhaps the comment clears up any confusion/thinking.. or maybe it's just bad code on my part :sweat_smile:

DSchau:
Fixes #13201

polarathene:
Any feedback to what I gave? If there's nothing wrong with it, I could perhaps make the changes as it seems @sidharthachatterjee is busy/away for a couple weeks now?

If so, how do I contribute to this PR? Or do I need to fork it and send another PR in? Or does @sidharthachatterjee still want to finish this PR?

packages/gatsby-image/src/index.js
```diff
@@ -258,8 +303,27 @@ class Image extends React.Component {
...
       itemProp,
+      critical,
     } = convertProps(this.props)
 
+    let { loading } = convertProps(this.props)
+
+    if (typeof critical === `boolean`) {
+      console.log(
+        `
+        The "critical" option is now deprecated and will be removed in a later version
```
DSchau:
```suggestion
        The "critical" option is now deprecated and will be removed the next major version 
```

packages/gatsby-image/src/index.js
```diff
@@ -258,8 +303,27 @@ class Image extends React.Component {
...
 
+    let { loading } = convertProps(this.props)
+
+    if (typeof critical === `boolean`) {
+      console.log(
+        `
+        The "critical" option is now deprecated and will be removed in a later version 
+        of "gatsby-image"
+
+        Please use the native "loading" attribute instead of critical.
```
DSchau:
```suggestion
        Please use the native "loading" attribute instead of "critical" 
```

packages/gatsby-image/src/index.js
sidharthachatterjee: We're good to go because the code in `handleRef` only handles the IO logic. onLoad for the image is called in `handleImageLoaded` (which will be called irrespective)

packages/gatsby-image/README.md
```diff
@@ -352,10 +352,16 @@ You will need to add it in your graphql query as is shown in the following snipp
...
 | `onError`              | `func`              | A callback that is called when the image fails to load.                                                                     |
 | `Tag`                  | `string`            | Which HTML tag to use for wrapping elements. Defaults to `div`.                                                             |
-| `critical`             | `bool`              | Opt-out of lazy-loading behavior. Defaults to `false`.                                                                      |
-| `objectFit`            | `string`            | Passed to the `object-fit-images` polyfill when importing from `gatsby-image/withIEPolyfill`. Defaults to `cover`.          |
+| `objectFit`            | `string`            | Passed to the `object-fit-images` polyfill when importing from `gatsby-image/withIEPolyfill`. \                             |
 | `objectPosition`       | `string`            | Passed to the `object-fit-images` polyfill when importing from `gatsby-image/withIEPolyfill`. Defaults to `50% 50%`.        |
 
+|
+| |
+| `loading` | `string` | Set the browser's native lazy loading attribute. One of `lazy`, `eager` or `auto`. Defaults to `lazy`. |
```
wardpeet: your markdown table seems to be broken

sidharthachatterjee: Fixed!

sidharthachatterjee:
@pieh You mean a glitch in the Matrix

sidharthachatterjee:
Published in `gatsby-image@2.1.0`

PR #13212, opened 2019-04-08
fix(gatsby-cli): Fix --clipboard on Windows

sidharthachatterjee:
`gatsby info --clipboard` has issues on Windows as documented in https://github.com/gatsbyjs/gatsby/issues/11496

envinfo dropped clipboard support in v6 and removed `clipboardy` so using it directly here. Not sure if it will fix https://github.com/gatsbyjs/gatsby/issues/11496 since no behaviour should've changed 

### Todo
- [ ] Get a reproduction of https://github.com/gatsbyjs/gatsby/issues/11496 from @saranshkataria and open issue in https://github.com/sindresorhus/clipboardy
- [ ] Upgrade to latest envinfo
- [ ] Fix issue upstream in clipboardy

### Related 
https://github.com/tabrindle/envinfo/issues/83 and https://github.com/gatsbyjs/gatsby/issues/11496

### Test

`npm i gatsby-cli@clipboardy`

### Other Changes
- The underline to the sections is gone since we log to console ourselves now (not a big deal in my opinion)
- Added `duplicates` and `showNotFound` options (useful for debugging and cra uses them too) 

### Before
Screenshot 2019-04-08 18 14 21
### After
Screenshot 2019-04-08 18 14 48
(Notice the Not Found for Firefox)

PR #13220, opened 2019-04-08
[DO NOT MERGE] ci: trigger ci runs

DSchau:
## Description

Just to see if the development_runtime test reliably fails.


DSchau:
Yeah - weirdly this passes just fine.

Something is a little weird with #13212 for some reason, as it doesn't seem _random_ but rather persistently fails, e.g. after a restart of the workflow.

PR #13223, opened 2019-04-08
test(www): add unit test base for www code base

### LINKED ISSUE #12545: [www]: Implement testing for www site
## Summary

After discussion during today's Community Meeting, we discussed that it would be good to start working on adding some forms of testing whether it be with Jest or Cypress to test parts of the site as well as the business logic for some pages. This is going to be an open issue to discuss what would be best to implement for the .org site

### Resources

* @DSchau had a example reference that we could use as a base for beginning to add tests to the site: https://github.com/DSchau/gatsby/pull/8
* Kent C Dodds and @jlengstorf did a livestream recently on testing Gatsby Sites: https://www.youtube.com/watch?v=BzRAYt7BHRw

DSchau:
## Description

This PR adds unit tests and some other niceties for www. This will give us at least a somewhat greater degree of confidence in more involved changes in the future.

I used react-testing-library and jest to implement the testing basis. Longer term--it'd be nice to have unit tests _and_ e2e tests for www, but we're currently limited by the size of gatsbyjs.org, currently.

## Related Issues

Fixes #12545


www/package.json
```diff
@@ -57,6 +57,7 @@
     "graphql-request": "1.6.0",
     "gray-percentage": "^2.0.0",
     "hex2rgba": "^0.0.1",
+    "jest-dom": "^3.1.3",
```
DSchau: Whoops! Waiting for the tests to (hopefully) pass then will send a commit to fix this.

marcysutton:
> Longer term--it'd be nice to have unit tests and e2e tests for www, but we're currently limited by the size of gatsbyjs.org

Is that because of images/other assets in source control, or something else?

DSchau:
> Is that because of images/other assets in source control, or something else?

gatsbyjs.org is a _fairly_ large and intensive Gatsby application. An e2e test would need to be artificially limited (e.g. source far less data from starters, showcases, etc.) to be able to be run in a CI task. In other words--building gatsbyjs.org currently times out in deploying to Netlify, so we'd need some way to move around that issue to be able to author an e2e test against the built version of the site.

It's not intractable, just takes a little more time and thought to do it correctly _and_ in a reasonable amount of time.

DSchau:
Going to merge this in, since it's a fairly innocuous change and adds some small, nice benefits so we can ask for a unit test for logic-heavy PRs to www!

Thanks @marcysutton for the review--as always!

Issue #13210, opened 2019-04-08
Remove one of package-lock.json and yarn.lock on gatsby new 


sidharthachatterjee:
Our starters currently include both `package-lock.json` and `yarn.lock` 

This typically leads to some confusion due to yarn's warning (https://github.com/gatsbyjs/gatsby/issues/4514#issuecomment-458729956)

> package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.

And in worst case, inconsistencies between the two, leading to issues like https://github.com/gatsbyjs/gatsby/issues/13194

### Potential solution

It'd be nice to remove both lock files altogether but we can't because of https://github.com/gatsbyjs/gatsby/issues/10607#issuecomment-450290701

We could remove `yarn.lock` from the starter repositories and run `yarn import` if `yarn` is detected (this solves inconsistencies in the starter itself as well as inconsistencies _after_ the project is created if both files are still present like in https://github.com/gatsbyjs/gatsby/issues/13194)

What do you think @DSchau @KyleAMathews?

sidharthachatterjee:
@JacobBlomgren Want to take a stab at this?

sidharthachatterjee:
Just remembered that we added @renovate[bot] and it'll handle updates to both `package-lock.json` and `yarn.lock`. And we only track `package-lock.json` in git, `yarn.lock` is created via `yarn import` in CI thanks to @DSchau 

So all we really need to do is, during `gatsby new`
- Check if `yarn` is installed
- If it is, remove `package-lock.json`
- If it is not, remove `yarn.lock`

We could also prompt, but since we internally use yarn currently if it is available (https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-cli/src/init-starter.js#L16-L28), we should just consistently do that for now

pieh:
> * Check if `yarn` is installed
> * If it is, remove `package-lock.json`
> * If it is not, remove `yarn.lock`

Just to clarify: the check should be "if should use yarn" ( `shouldUseYarn` function that was linked by @sidharthachatterjee in previous comment) and not direct check if `yarn` is installed.

KyleAMathews:
The real long-term solution here is to ask people which client they use on their first `gatsby new` and then use that in the future.

frinyvonnick:
[...]

PR #13225, opened 2019-04-08
feat(gatsby-cli): Remove one of package-lock.json and yarn.lock on gatsby new

frinyvonnick:
## Description

I made a first implementation that asks user with [prompts](https://github.com/terkelg/prompts) which package manager he wants to use.

## Related Issues

Related to issue #13210


packages/gatsby-cli/src/init-starter.js
```diff
@@ -67,8 +68,29 @@ const install = async rootPath => {
   process.chdir(rootPath)
 
   try {
-    let cmd = shouldUseYarn() ? spawn(`yarnpkg`) : spawn(`npm install`)
-    await cmd
+    const npmCmd = `npm install`
+    let response = npmCmd
+    if (shouldUseYarn()) {
```
sidharthachatterjee: So in this it'll only prompt the user is `yarn` is detected and if not, it'll continue with a prompt. Correct?

packages/gatsby-cli/src/init-starter.js
```diff
@@ -67,8 +68,29 @@ const install = async rootPath => {
...
+      const promptsAnswer = await prompts([
+        {
+          type: `select`,
+          name: `package_manager`,
+          message: `Which package manager would you like to use ?`,
+          choices: [
+            { title: `yarn`, value: `yarnpkg` },
+            { title: `npm`, value: npmCmd },
+          ],
+          max: 1,
```
sidharthachatterjee: Let's set the `initial` here to `yarn` by default? 

https://www.npmjs.com/package/prompts#initial

packages/gatsby-cli/src/init-starter.js
```diff
@@ -67,8 +68,29 @@ const install = async rootPath => {
...
+        },
+      ])
+      response = promptsAnswer.package_manager
```
sidharthachatterjee: We need to persist the user's input using https://github.com/yeoman/configstore so that it's a one time prompt.

packages/gatsby-cli/src/init-starter.js
frinyvonnick: Sure !

frinyvonnick:
Thank you for your review @sidharthachatterjee 👍

packages/gatsby-cli/src/init-starter.js
frinyvonnick: I used the package.json for the [configuration](https://github.com/gatsbyjs/gatsby/pull/13225/commits/189e0ec08ce358ec204df2f56b52886c022a15a9#diff-7dc3091639b4f0b32b7271d7eb617ea8R19) was that what you expected ?

packages/gatsby-cli/src/init-starter.js
```diff
@@ -8,11 +9,15 @@ const report = require(`./reporter`)
...
 const existsSync = require(`fs-exists-cached`).sync
 const { trackCli, trackError } = require(`gatsby-telemetry`)
+const prompts = require(`prompts`)
 const spawn = (cmd: string, options: any) => {
   const [file, ...args] = cmd.split(/\s+/)
   return execa(file, args, { stdio: `inherit`, ...options })
 }
 
+const pkg = require(`../package.json`)
+const conf = new Configstore(pkg.name, {})
```
wardpeet: We already have configstore in place in our telemetry package and it's probably ok to reuse that file.
https://github.com/gatsbyjs/gatsby/blob/f5a85af786e2be73afcaf94f8da68f1e08bb42f9/packages/gatsby-telemetry/src/event-storage.js#L19

```suggestion
const conf = new Configstore(`gatsby`, {}, { globalConfigPath: true })
```

we probably want to abstract it away.

yarn.lock
```diff
@@ -6298,7 +6305,7 @@ config-chain@^1.1.11:
     ini "^1.3.4"
     proto-list "~1.2.1"
 
-configstore@4.0.0:
+configstore@4.0.0, configstore@^4.0.0:
```
wardpeet: while you're at it could you update https://github.com/gatsbyjs/gatsby/blob/f5a85af786e2be73afcaf94f8da68f1e08bb42f9/packages/gatsby-telemetry/package.json#L14 to have a ^ `"configstore": "^4.0.0"`

frinyvonnick:
Thank for your review @wardpeet 👍

frinyvonnick:
Can't make e2e tests locally even on master last commit I always have the same tests failing. @wardpeet @sidharthachatterjee do you have some pointer to help me getting rid of this ?

packages/gatsby-cli/src/init-starter.js
```diff
@@ -8,11 +9,14 @@ const report = require(`./reporter`)
 const url = require(`url`)
 const existsSync = require(`fs-exists-cached`).sync
 const { trackCli, trackError } = require(`gatsby-telemetry`)
+const prompts = require(`prompts`)
 const spawn = (cmd: string, options: any) => {
   const [file, ...args] = cmd.split(/\s+/)
   return execa(file, args, { stdio: `inherit`, ...options })
 }
 
+const conf = new Configstore(`gatsby`, {}, { globalConfigPath: true })
```
sidharthachatterjee: @jamo Can you check if this affects our conditional in telemetry?

packages/gatsby-cli/src/init-starter.js
```diff
@@ -67,8 +71,30 @@ const install = async rootPath => {
...
+            { title: `npm`, value: npmCmd },
+          ],
+          initial: 0,
+        },
+      ])
+      response = promptsAnswer.package_manager
+    }
+    conf.set(`package_manager`, response)
+    if (response.includes(`yarn`)) {
+      await spawn(`rm package-lock.json`)
```
jamo: Won't this fail on e.g., windows? 

I suppose we could just use the [`fs`](https://nodejs.org/api/fs.html#fs_fs_unlinksync_path) module to delete the file
```javascript
fs.unlinkSync(`package-lock.json`)
```

packages/gatsby-cli/src/init-starter.js
```diff
@@ -67,8 +71,30 @@ const install = async rootPath => {
   process.chdir(rootPath)
 
   try {
-    let cmd = shouldUseYarn() ? spawn(`yarnpkg`) : spawn(`npm install`)
-    await cmd
+    const npmCmd = `npm install`
+    let response = npmCmd
+    if (shouldUseYarn()) {
+      const promptsAnswer = await prompts([
```
jamo: Will this wait infinitely long for user-input?

packages/gatsby-cli/src/init-starter.js
jamo: It's a bit tricky to verify, but based on one two tests and one simulation co-existence of multiple configstores is fine. And as long as we don't poke the `telemetry` namespace we should be good.

packages/gatsby-cli/src/init-starter.js
frinyvonnick: Yes until user select one choice

frinyvonnick:
Thanks for your review @jamo 👍

packages/gatsby-cli/src/init-starter.js
frinyvonnick: How could I be sure to not make a regression on this ?

jamo: Reading the configstore core and based on manual testing I'm fairly confident that we can safely have multiple configstore instances w/o any issues or regressions

packages/gatsby-cli/src/init-starter.js
wardpeet: This is great! no need to ask if they don't have it installed.

packages/gatsby-cli/src/init-starter.js
```diff
@@ -67,8 +71,30 @@ const install = async rootPath => {
...
+            { title: `npm`, value: npmCmd },
+          ],
+          initial: 0,
+        },
+      ])
+      response = promptsAnswer.package_manager
+    }
+    conf.set(`package_manager`, response)
+    if (response.includes(`yarn`)) {
+      fs.unlinkSync(`package-lock.json`)
```
wardpeet: async removal of files has better support on windows as it retries if it fails on the first try
```suggestion
      await fs.remove(`package-lock.json`)
```

https://github.com/jprichardson/node-fs-extra/blob/master/lib/remove/rimraf.js#L45-L51

packages/gatsby-cli/src/init-starter.js
```diff
@@ -67,8 +71,30 @@ const install = async rootPath => {
...
+    } else {
+      fs.unlinkSync(`yarn.lock`)
```
wardpeet: see comment above
```suggestion
      await fs.remove(`yarn.lock`)
```

wardpeet:
This looks great! just a few remarks on fs.unlinksync but after that this is great!

ivanoats:
How would we resolve this issue regarding using the starters, their own repos, and how they might be updated?

For example, right now I am trying `gatsby new hello-world-test` [old starter repo that is in the docs](https://www.gatsbyjs.org/tutorial/part-zero/#create-a-gatsby-site) and see it still has both yarn and package.json in it's own repo. But the starter does not have `yarn.lock` in this main repo.

I see this is being covered on https://github.com/gatsbyjs/gatsby/issues/16135 so I guess I'm just leaving this comment here to help other people out who might be running into this problem.

PR #13226, opened 2019-04-08
chore(docs): reword module name for building themes

Saifadin:
I tripped over this line, while working with themes and I think making it clear, that the theme name goes here is important.

### DIFF
--- a/docs/docs/themes/api-reference.md
+++ b/docs/docs/themes/api-reference.md
@@ -118,7 +118,7 @@ module.exports = {
     {
       resolve: "gatsby-plugin-compile-es6-packages",
       options: {
-        modules: ["gatsby-theme-developer"],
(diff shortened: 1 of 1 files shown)


docs/docs/themes/api-reference.md
```diff
@@ -118,7 +118,7 @@ module.exports = {
     {
       resolve: "gatsby-plugin-compile-es6-packages",
       options: {
-        modules: ["gatsby-theme-developer"],
+        modules: ["NAME_OF_YOUR_THEME"],
```
DSchau: How about we solve this in a different way?

```suggestion
        // replace with the name of your theme
        // highlight-next-line
        modules: ["gatsby-theme-developer"],
```

Saifadin: Yeah that also works^^

PR #13228, opened 2019-04-08
[www] Show no sidebar on top-level markdown pages

marcysutton:
## Description

The Accessibility Statement makes sense to be top level, so I added the Markdown file to the base `/docs` directory. It was showing the tutorial sidebar due to some code in the docs-markdown template: https://github.com/gatsbyjs/gatsby/blob/master/www/src/templates/template-docs-markdown.js#L59

This PR simply removes the fallback array entry with the Tutorial sidebar. I wasn't sure if there were some tests I should update as well (that would make sense!). It seemed to work fine locally.

### DIFF
--- a/www/src/templates/template-docs-markdown.js
+++ b/www/src/templates/template-docs-markdown.js
@@ -56,7 +56,7 @@ const getDocsData = location => {
     tutorial: itemListTutorial,
   }
 
-  return [urlSegment, itemListLookup[urlSegment] || itemListTutorial]
+  return [urlSegment, itemListLookup[urlSegment]]
 }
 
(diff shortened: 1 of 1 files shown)


fk:
>  I wasn't sure if there were some tests I should update as well (that would make sense!).

There aren't any, but there is … where … ah: #12545 (and you know about that already ;-)).

marcysutton:
Sweet! Perhaps we could add some sidebar tests once that PR is merged. As long as @DSchau doesn't see any issues, it might be good to merge this one so the Accessibility Statement page doesn't have an irrelevant sidebar.

PR #13230, opened 2019-04-08
fix(e2e-tests): run gatsby-dev without --copy-all

pieh:


DSchau:
@pieh I did validate this seems to be the fix for the e2e tests. However -- looks like the integration tests failed (for some reason) in 153dfdf.

I'll have a little time to check this out tomorrow--but I actually think we can/should merge this for the e2e tests.

sidharthachatterjee:
@DSchau Integration tests were fixed in https://github.com/gatsbyjs/gatsby/pull/13251. Merging this in.

Issue #7064, opened 2018-08-06
Investigate compiling `node_modules` code


m-allanson:
## Summary

Investigate:

Copy [CRA's implementation for compiling `node_modules` with babel-preset-env](https://github.com/facebook/create-react-app/pull/3776) which will fix problems like https://github.com/gatsbyjs/gatsby/issues/3780

Taken from #3870

jarreds:
Definitely interested in this. With more modules moving to ES6 and the need to support older browsers (Google bot 👀), this would be wonderful. Currently using old ES5 versions of the problematic modules. Not great.

Even if it's not the default, an easy way to flag compilation of `node_modules` would be great.

m-allanson:
This probably won't be done for the initial v2 release.

ZLevine:
Any idea if this is on the roadmap for the near future?

pgegenfurtner:
Any news on this? With the new theming capabilities it would be a great addition to extract graphql fragments into a theme. Or could this be achieved in any other way?

DSchau:
@pgegenfurtner great point! So how we currently handle this now is we leverage the existing APIs we already offer. Longer term--we'd like to _not_ have to do this (e.g. transpile node_modules) but for now, you can just use `onCreateWebpackConfig` to transpile the current theme.

For an example, check out my [blog-theme](https://github.com/DSchau/gatsby-theme-blog/blob/master/gatsby-node.js#L130).

There's some indirection there, so the general idea is as simple as:

```js
exports.onCreateWebpackConfig = function onCreateWebpackConfig({ actions, loaders }) {
  actions.setWebpackConfig({
    module: {
      rules: [
        {
          test: 'gatsby-theme-your-theme',
          use: [loaders.js()],
        },
      ],
    },
  });
}
```

Definitely your point is well taken that this can be improved.

As far as fragments go--we currently (automatically) inspect fragments in a package named `gatsby-theme-*`, but longer term, anything added to the `themes` (currently experimentalThemes) will automatically be loaded and the fragments and graphql queries will be available!

pgegenfurtner:
@DSchau thank you for your quick response!
Via [this issue](https://github.com/ChristopherBiscardi/gatsby-theme-examples/issues/14) and your [fixing commit](https://github.com/gatsbyjs/gatsby/pull/10835) I found out about the naming convention with `gatsby-theme-*`. It works just like I imagined!

Definitely one of the best new gatsby features.

[...]

PR #13241, opened 2019-04-09
chore(gatsby): add ie polyfill to webpack entry

### LINKED ISSUE #7003: [v2] IE11 throws Objects are not valid as a React child

nihgwu:
## Description

This PR adds polyfill for IE according to the `browserlist` config, the current polyfill strategy (`useBuiltIns: usage`) works well in most cases, but I encountered with errors only on IE in all my sites built with Gatsby, I digged a lot and found that(my guess) it's caused by the polyfill for `Symbol`, the `useBuiltIns: usage` doesn't work as `React` can live without `Symbol` so it's not polyfilled, but then it's polyfilled if the `Symbol` is used in other packages, so we have to add the `Symbol` polyfill before `React` bootstraps, here is my current solution for my sites https://github.com/Autodesk/react-base-table/blob/master/website/gatsby-node.js#L16

## Related Issues

Fixes #7003

### DIFF
--- a/packages/gatsby/src/utils/webpack.config.js
+++ b/packages/gatsby/src/utils/webpack.config.js
@@ -5,6 +5,7 @@ const path = require(`path`)
 const dotenv = require(`dotenv`)
 const FriendlyErrorsWebpackPlugin = require(`@pieh/friendly-errors-webpack-plugin`)
 const PnpWebpackPlugin = require(`pnp-webpack-plugin`)
+const browserslist = require(`browserslist`)
 const { store } = require(`../redux`)
 const { actions } = require(`../redux/actions`)
 const debug = require(`debug`)(`gatsby:webpack-config`)
@@ -28,6 +29,18 @@ module.exports = async (
   webpackPort = 1500
 ) => {
   const directoryPath = withBasePath(directory)
+
+  // Polyfil for IE support
+  const supportedBrowsers = browserslist(program.browserslist)
+  let iePolyfil = false
+  if (
+    supportedBrowsers.includes(`ie 9`) ||
+    supportedBrowsers.includes(`ie 10`)
+  ) {
+    iePolyfil = `react-app-polyfill/ie9`
+  } else if (supportedBrowsers.includes(`ie 11`)) {
+    iePolyfil = `react-app-polyfill/ie11`
+  }
 
   process.env.GATSBY_BUILD_STAGE = suppliedStage
 
@@ -151,12 +164,13 @@ module.exports = async (
       case `develop`:
         return {
           commons: [
+            iePolyfil,
             `event-source-polyfill`,
             `${require.resolve(
               `webpack-hot-middleware/client`
             )}?path=${getHmrPath()}`,
             directoryPath(`.cache/app`),
-          ],
+          ].filter(Boolean),
         }
       case `develop-html`:
         return {
@@ -168,7 +182,9 @@ module.exports = async (
         }
       case `build-javascript`:
         return {
-          app: directoryPath(`.cache/production-app`),
+          app: [iePolyfil, directoryPath(`.cache/production-app`)].filter(
+            Boolean
+          ),
         }
       default:
         throw new Error(`The state requested ${stage} doesn't exist.`)
--- a/packages/gatsby/src/utils/webpack.config.js
+++ b/packages/gatsby/src/utils/webpack.config.js
@@ -30,16 +30,16 @@ module.exports = async (
 ) => {
   const directoryPath = withBasePath(directory)
 
-  // Polyfil for IE support
+  // Polyfill for IE support
   const supportedBrowsers = browserslist(program.browserslist)
-  let iePolyfil = false
+  let iePolyfill = false
   if (
     supportedBrowsers.includes(`ie 9`) ||
     supportedBrowsers.includes(`ie 10`)
   ) {
-    iePolyfil = `react-app-polyfill/ie9`
+    iePolyfill = `react-app-polyfill/ie9`
   } else if (supportedBrowsers.includes(`ie 11`)) {
-    iePolyfil = `react-app-polyfill/ie11`
+    iePolyfill = `react-app-polyfill/ie11`
   }
 
   process.env.GATSBY_BUILD_STAGE = suppliedStage
@@ -164,7 +164,7 @@ module.exports = async (
       case `develop`:
         return {
           commons: [
-            iePolyfil,
+            iePolyfill,
             `event-source-polyfill`,
             `${require.resolve(
               `webpack-hot-middleware/client`
@@ -182,7 +182,7 @@ module.exports = async (
         }
       case `build-javascript`:
         return {
-          app: [iePolyfil, directoryPath(`.cache/production-app`)].filter(
+          app: [iePolyfill, directoryPath(`.cache/production-app`)].filter(
             Boolean
           ),
         }
--- a/docs/docs/browser-support.md
+++ b/docs/docs/browser-support.md
@@ -49,8 +49,6 @@ By default, Gatsby emulates the following config:
 If you only support newer browsers, make sure to specify this in your
 `package.json`. This will often enable you to ship smaller JavaScript files.
 
-## Note about IE < 11
+## Note about IE support
 
-React depends on collection types `Map` and `Set`. While these are not used by Gatsby, Gatsby uses React and you will need to polyfill these if you support older browsers and devices including IE < 11.
-
-Read more about this in [https://reactjs.org/docs/javascript-environment-requirements.html](https://reactjs.org/docs/javascript-environment-requirements.html)
+Gatsby will polyfil minimum requirements for IE support according to the `browserslist` config.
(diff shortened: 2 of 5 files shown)


wardpeet:
This PR looks really great! It's a bit weird that babel-polyfill isn't polyfilling. Let me dig into that first before proceeding with this PR.

nihgwu:
I guess that's because in SSR building, `React` is being processed but `Symbol` is not polyfilled as it's optional, React will use a fallback, then processing the third-party packages, `Symbol` is polyfilled because of `useBuiltIns=true` (I included the third-party packages for `babel-loader`), then you deploy the SSR output, `React` detects mismatch of elements, so I tried to polyfill `Symbol` before `React` and it works like a . charm, and I find there are a lot of this kind of errors like reported in #7003 not only in Gatsby, but most of answers are not really helpful, someone found that add `@babel/polyfill` before the entry solve the problem, but I guess the polyfill for `Symbol` is the key as other polyfills should work with `useBuiltins=true`

packages/gatsby/src/utils/webpack.config.js
```diff
@@ -29,6 +30,18 @@ module.exports = async (
 ) => {
   const directoryPath = withBasePath(directory)
 
+  // Polyfill for IE support
+  const supportedBrowsers = browserslist(program.browserslist)
+  let iePolyfill = false
+  if (
+    supportedBrowsers.includes(`ie 9`) ||
+    supportedBrowsers.includes(`ie 10`)
```
nihgwu: there are `ie-mob 10` and `ie-mob 11`, should we also check for those versions? Personally I don't want to do that

packages/gatsby/src/utils/webpack.config.js
```diff
@@ -29,6 +30,18 @@ module.exports = async (
...
+  ) {
+    iePolyfill = `react-app-polyfill/ie9`
```
nihgwu: Actually we only need to polyfill `Set` `Map` `Symbol` and other polyfills will be done by `babel-loader`, we could create our own polyfill rules instead of using `react-app-polyfill`, but I love the best practice from `create-react-app`

nihgwu:
@wardpeet any chance to review? the change is relative slight but would save those who are struggling with IE support, especially the `Symbol` problem which is hard to identify, thanks

wardpeet:
Hey, sorry for the late response but we rather not make this change. We are more eager to go for https://github.com/gatsbyjs/gatsby/issues/7064. Which will let us babel transform all our plugins which will let babel-polyfill, polyfill all features.

At the moment there is a possible workaround by using https://www.gatsbyjs.org/packages/gatsby-plugin-compile-es6-packages/ and adding react-dom to that list.

We want you to know that we certainly want to keep you as a contributor to Gatsby.

nihgwu:
@wardpeet as I said, even you add those es6 packages to be compiled by babel loader, it won't solve the problem I pointed out, the `Symbol` polyfill issue, I did add those es6 packages in my project https://github.com/Autodesk/react-base-table/blob/master/website/gatsby-node.js#L35-L38, and it still doesn't help because of that

BTW, I'm ranked 21st in the contributors, I contributed a lot before but was busy with my own stuffs then. I just received my swags today(I got the coupons long before, but just used them last week), they are so good 😄

nihgwu:
Let me explain a bit more with IE, I'm using an ES6 package `react-inspector`, so I have to make it transpiled by babel, and that package also uses Symbol and Symbol is polyfilled by babel, but before that no packages use `Symbol` directly as they fallback if Symbol is not available like `var hasSymbol = typeof Symbol === 'function' && Symbol.for;`,  when I haven't used `react-inspector`, everything goes fine, as `Symbol` is not required, but when I navigate to the page used `react-inspector` the page will crash, because in this page, the Symbol is available since it's polyfilled, but `React` is initialized without Symbol support before that, so there are mismatches and throw `Objects are not valid as a React child (found: object with keys {$$typeof, type, key, ref, props, _owner}).`

pieh:
Hey @nihgwu! Did you by any chance check if https://github.com/gatsbyjs/gatsby/pull/14111 solve the problem? Ideally we could just use that and not add extra babel presets

nihgwu:
@pieh see my comment here https://github.com/gatsbyjs/gatsby/pull/13241#issuecomment-484060739, that PR just make all the packages transpiled by Babel, but still doesn't solve the Symbol issue I described

pieh:
> @pieh see my comment here [#13241 (comment)](https://github.com/gatsbyjs/gatsby/pull/13241#issuecomment-484060739), that PR just make all the packages transpiled by Babel, but still doesn't solve the Symbol issue I described

Hmm, I didn't try it, but if all packages would be transpiled (including React) - wouldn't @babel/preset-env pick up that it needs to polyfill Symbol and that would happen before react-inspector and then both React and react-inspector would use that polyfilled Symbol?

wardpeet:
https://github.com/gatsbyjs/gatsby/pull/14111 seems to work for react-inspector. Somehow I can't get your example working

nihgwu:
@wardpeet That PR works for `react-inspector` for sure, but I guess it should works for my site as well I guess, https://babeljs.io/repl#?babili=false&browsers=last%201%20versions&build=&builtIns=usage&spec=false&loose=true&code_lz=G4QwTgBAFiDODKBPAtgIwPYBsIF4IBdEAHAU3QDMIk0tcc8BycgVwDsBjfAS3VYYgBkAqigyYAdOXRgA3EA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.4.5&externalPlugins=%40babel%2Fplugin-proposal-private-methods%407.4.3%2C%40babel%2Fplugin-proposal-class-properties%407.4.0%2C%40babel%2Fplugin-proposal-object-rest-spread%407.4.3, the Symbol will be polyfilled then in a test expression, so it will be with React, I don't have a windows device to test, btw I updated my website a bit, not sure if it's related to your failure building

wardpeet:
I've tested https://github.com/gatsbyjs/gatsby/pull/14111 and got your example to work with the babel-loader PR.

Thanks a bunch for tackling this, sadly i'll be closing as babel-loader is a better approach as it will make our live easier when moving #14289 forward.

nihgwu:
@wardpeet Thanks so much for you for you patients

wardpeet:
@nihgwu Fix got p in `gatsby@2.11.0`.

Thank you, for your patience :wink:

nihgwu:
cool, will try, thanks

PR #13251, opened 2019-04-09
fix(gatsby-dev-cli): Persist verdaccio stuff in os.tmpdir

sidharthachatterjee:
It currently persists in a dir two steps back (where gatsby-dev is) and sometimes that isn't writable because of an operation previously run with `sudo` like in https://circleci.com/gh/gatsbyjs/gatsby/104049?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Note: `require('os').tmpdir()` is good in Node.js 4 and up. So we're fine.


sidharthachatterjee:
Published in `gatsby-dev-cli@2.4.15`

PR #13256, opened 2019-04-09
chore(gatsby-cli): speed up the CLI by using ncc and compiling

### LINKED ISSUE #12953: Use https://zeit.co/blog/ncc to compile gatsby-cli to a single file for faster installs & startups
We want our CLI to be as fast as possible to install and bootup since we expect people to run it all the time. Compiling the package with [ncc](https://zeit.co/blog/ncc) could help a lot with that.

DSchau:
## Description

This PR introduces [ncc](https://github.com/zeit/ncc), which uses Webpack to compile a Node.js folder to a single file which purports to speed up boot-up/execution times drastically.

I saw _reasonable_ improvements in my testing, generally around ~`200ms` improvements in comparing compiled vs. the existing version. I think this would be more drastic on slower machines.

Note: I had to remove `update-notifier` which is failing on [this line](https://github.com/yeoman/update-notifier/blob/master/index.js#L15) via compilation through Webpack. This seems to be very valuable behavior--so we'll need to find a solution to that outside of just outright removal.

### Performance

I ran this script to loop 5 times and compare the timings:

```bash
#!/bin/bash
set -e # fail on errors, what a novel concept
ITERATIONS=5
FOLDER_NAME="test"
declare -a NCC_TIMINGS
declare -a NODE_TIMINGS

rm -rf $FOLDER_NAME
gatsby new $FOLDER_NAME

cd $FOLDER_NAME

for i in `seq 1 ${ITERATIONS}`;
do
  rm -rf .cache public
  echo "NODE ================"
  NODE_TIMINGS[$i]=$(time node ../../gatsby/packages/gatsby-cli/lib/index.js build)
  echo "NCC ================"
  NCC_TIMINGS[$i]=$(time node ../../gatsby/packages/gatsby-cli/dist/index.js build)
  rm -rf .cache public
done    
```

## Related Issues

Closes #12953

Blocked by yeoman/update-notifier#123


KyleAMathews:
Very cool! And yeah, this would be vastly faster on computers with slow hard drives. I've seen people starting gatsby on slow computers and it's easily 2-4x slower than a new Mac.

DSchau:
We'll also need to keep the lib/ directory around for backwards compatibility reasons (thus the _very helpful_ suite of failing e2e tests).

DSchau:
Just need to add the lib/ directory back and this is mostly good to go. Will want to find a workaround for update-notifier, though.

packages/gatsby-cli/package.json
```diff
@@ -21,7 +22,7 @@
     "convert-hrtime": "^2.0.0",
     "core-js": "^2.5.0",
     "envinfo": "^5.8.1",
-    "execa": "^0.8.0",
+    "execa": "^1.0.0",
```
DSchau: Was running into issues--this seems to not have any breaking changes for us!

KyleAMathews:
The latest version of update-notifier only supports Node 8+ https://github.com/yeoman/update-notifier/commit/aafd8a04737efb798925ad1ba321965ec8e7a460

So we still might want to fork an older version until v3.

DSchau:
@KyleAMathews good catch. That was technically already an issue though, right?

e.g. our dependency range was `^2.3.0` which includes the latest releases.

KyleAMathews:
haha yeah — I guess they didn't think dropping support is a breaking change 🤷‍♂ 

Perhaps this is why we have so few node 6 users 🤔

wardpeet:
@DSchau After the merge I can fix update-notifier to be compliant with webpack, so we can remove our own published version. On the node 8 front I think they'll publish to version 3 so we should be good with keeping us at version ^2.0.0

wardpeet:
oh ncc has it's own webpack config we can't really touch... that's a bummer. We could put upgrade-notifier as an external..

wardpeet:
just tested it a few times on windows and it's about 250ms slower on default starter (on my machine). Should we keep lib? As you don't include it in your npm package so we might want to clean it up after ncc is done?

DSchau:
I want to investigate @wardpeet's findings earlier before we merge this, because it should absolutely provide a speed-up in all cases due to decreased start-up cost!

DSchau:
Due to the additional complexity, and the uncertainty around whether this change actually makes a meaningful difference in start-up time (and more broadly speaking, general CLI time), going to close this out. We can revisit in the future (and use this PR as a basis).

PR #13258, opened 2019-04-09
Updated part zero to move optional links to references section

### LINKED ISSUE #12845: Bring tutorial into alignment with style guide section regarding reducing hyperlinks
## Summary

Bring tutorial into alignment with style guide section regarding reducing hyperlinks: https://www.gatsbyjs.org/contributing/gatsby-style-guide/#dont-make-users-think-more-than-is-necessary

All hyperlinks that aren't strictly necessary for beginners to complete the steps of the tutorial should be removed and/or moved to the bottom of the page in an "additional resources" section.

### Basic example

For example, this section has a link to 404 pages and to API docs, neither of which are necessary for beginners to complete the tasks of the tutorial. https://www.gatsbyjs.org/tutorial/part-one/#-using-the-link--component

### Motivation

We are doing this to make sure beginners don't get side-tracked more than is necessary. We want this tutorial to be as powerful as possible. It isn't easy to do this, and will likely require this kind of maintenance to bring it back into alignment with the style guide on a regular basis.

shannonbux:
Partially fixes #12845 

Moves optional learning materials into a References section at the bottom of Part Zero.

It would be cool to have footnotes...maybe. Those can be distracting a bit too, just less distracting.

### DIFF
--- a/docs/tutorial/part-zero/index.md
+++ b/docs/tutorial/part-zero/index.md
@@ -5,25 +5,11 @@ typora-copy-images-to: ./
 
 Before you start building your first Gatsby site, you’ll need to familiarize yourself with some core web technologies and make sure that you have installed all required software tools.
 
-## Overview of core technologies
-
-It’s not necessary to be an expert with these already — if you’re not, don’t worry! You’ll pick up a lot through the course of this tutorial series. These are some of the main web technologies you’ll use when building a Gatsby site:
-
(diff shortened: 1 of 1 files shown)


lannonbr:
Good stuff @shannonbux. I was watching someone go through the tutorial a few weeks back and they already knew web-dev and node stuff so pushing this content down to the bottom I think is a win.

PR #13260, opened 2019-04-09
fix(gatsby): fix infinite calls with circular nodes

### LINKED ISSUE #13259: Infinite recursion calls with circular nodes
## Description

After upgrade to gatsby 2.3.15, causing issue xiaoxinghu/orgajs#24.
Get errors like `RangeError: Maximum call stack size exceeded` when building with plugin [gatsby-tranformer-orga](https://github.com/xiaoxinghu/orgajs)

xiaoxinghu:
## Description

This should be a mistake, causing issues with nodes have `parent` field.

## Related Issues

#13259

### DIFF
--- a/packages/gatsby/src/schema/infer/example-value.js
+++ b/packages/gatsby/src/schema/infer/example-value.js
@@ -106,6 +106,7 @@ const getExampleObject = ({
         nodes: objects,
         prefix: selector,
         typeConflictReporter,
+        ignoreFields,
       })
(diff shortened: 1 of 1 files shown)


stefanprobst:
We don't include `ignoreFields` on deeper nesting levels on purpose, since it is perfectly valid  to e.g. have a `id` or `parent` field on a nested type. I commented on the linked issue as well -- a small reproduction repo would be very helpful to investigate. Thanks!!

PR #13282, opened 2019-04-11
Expose LOCAL_URL as gatsby-link method

Kikobeats:
Reading [gatsby-link](https://www.gatsbyjs.org/docs/gatsby-link/#reminder-use-link-only-for-internal-links) documentation, it says you need to validate links you pass to `GatsbyLink` should be just internal links:

```js
import { Link as GatsbyLink } from "gatsby"

// Since DOM elements <a> cannot receive activeClassName,
// destructure the prop here and pass it only to GatsbyLink
const Link = ({ children, to, activeClassName, ...other }) => {
  // Tailor the following test to your environment.
  // This example assumes that any internal link (intended for Gatsby)
  // will start with exactly one slash, and that anything else is external.
  const internal = /^\/(?!\/)/.test(to)
```

This is done using a regex; however, instead of re-declare the regex, I expose it from the internal module, being the same and exposed as function:

```js
import { Link as GatsbyLink } from "gatsby"
import { isInternalLink } from "gatsby-link"

// Since DOM elements <a> cannot receive activeClassName,
// destructure the prop here and pass it only to GatsbyLink
const Link = ({ children, to, activeClassName, ...other }) => {
  // Tailor the following test to your environment.
  // This example assumes that any internal link (intended for Gatsby)
  // will start with exactly one slash, and that anything else is external.
  const internal = /^\/(?!\/)/.test(to)
  // Use Gatsby Link for internal links, and <a> for others
  if (isInternalLink(to)) {
```

### DIFF
--- a/docs/docs/gatsby-link.md
+++ b/docs/docs/gatsby-link.md
@@ -297,17 +297,13 @@ following may be a good starting point:
 
 ```jsx
 import { Link as GatsbyLink } from "gatsby"
+import { isInternalLink } from "gatsby-link"
 
 // Since DOM elements <a> cannot receive activeClassName,
 // destructure the prop here and pass it only to GatsbyLink
 const Link = ({ children, to, activeClassName, ...other }) => {
-  // Tailor the following test to your environment.
-  // This example assumes that any internal link (intended for Gatsby)
-  // will start with exactly one slash, and that anything else is external.
-  const internal = /^\/(?!\/)/.test(to)
-
   // Use Gatsby Link for internal links, and <a> for others
-  if (internal) {
+  if (isInternalLink(to)) {
     return (
       <GatsbyLink to={to} activeClassName={activeClassName} {...other}>
         {children}
--- a/packages/gatsby-link/index.d.ts
+++ b/packages/gatsby-link/index.d.ts
@@ -17,6 +17,7 @@ export default class GatsbyLink<TState> extends React.Component<
 > {}
 export const navigate: NavigateFn
 export const withPrefix: (path: string) => string
+export const isInternalLink: (to: string) => boolean
 
 // TODO: Remove navigateTo, push & replace for Gatsby v3
 export const push: (to: string) => void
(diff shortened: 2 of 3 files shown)


docs/docs/gatsby-link.md
```diff
@@ -297,17 +297,13 @@ following may be a good starting point:
 
 ```jsx
 import { Link as GatsbyLink } from "gatsby"
```
Kikobeats: I'm not completely sure why `Link` is used here.

why not `import GatsbyLink from 'gatsby-link'`?

pieh: Because `gatsby-link` is internal dependency of gatsby (is not installed by users directly). Importing modules that are not direct dependencies of your packages (site in this case) is bad practice and can lead to very unexpected results (specifically if there would be more than 1 instance of `gatsby-link` installed in `node_modules)

pieh: Because of that - the `isInternalLink` would need to be re-exported from `gatsby` (similiar to how `Link` is re-exported from `gatsby`)

Kikobeats: oh, gotcha.

What do you think expose is using `gatsby.Link`?

```
import { Link as GatsbyLink } from "gatsby"
const {isInternalLink} = GatsbyLink
```

Kikobeats:
I'm blocked until @pieh feedback me :(

sidharthachatterjee:
Hey @Kikobeats 

I think @pieh meant that we'll have to reexport `isInternalLink` from `gatsby` since we want users to import it like

`import { isInternalLink } from gatsby`

(like they do for Link)

pieh:
I was thinking about this more, and extra export from Gatsby doesn't seem to provide much value IMO: 
It's pretty niche so users would only learn about it from that section in docs and if they see that, they might as well use regex that is provided there. 

Adding public APIs is always tough decision, and if users are able to do same thing without extra public API, then we should not add it, as we lock ourselves to support that API for a long time.

Given above we will not be adding `isInternalLink` public API. Thanks again, and we look forward to seeing more PRs from you in the future! :muscle::purple_heart:

PS. I think that documentation there (when it's not known if links are internal or not) should instead propose to use `<a>` tags and [gatsby-plugin-catch-links](https://www.gatsbyjs.org/packages/gatsby-plugin-catch-links/), but that's separate discussion.

PR #13286, opened 2019-04-11
chore(starters): add gatsby-antd-docs

jannikbuschke:


docs/starters.yml
```diff
@@ -2207,3 +2207,17 @@
...
     - TypeScript support
     - Easily work in GatsbyJS and Framer X at the same time
+- url: https://www.jannikbuschke.de/gatsby-antd-docs/
+  repo: https://github.com/jannikbuschke/gatsby-antd-docs
+  description: A template for documenation websites
+  tags:
+    - Documentation
+    - Typescript
+    - Antd
+    - AntDesign
```
pieh: I think `Antd` and `AndDesign` relate to same thing, so I would prefer to chose one - and use `Styling:` prefix in same way we do it for other UI libraries - `Styling:AntDeisgn` seems best (IMO)

PR #13287, opened 2019-04-11
fix(gatsby-image): ensure that currentSrc exists

thebuilder:
Trying to read `currentSrc` before it has been set throws an `Unable to get property 'length' of undefined or null reference` error.
It should be an empty string, but IE11 can return `undefined`.

The issue was introduced by #12468 - it's simply missing an extra guard to ensure the value is defined.

### DIFF
--- a/packages/gatsby-image/src/index.js
+++ b/packages/gatsby-image/src/index.js
@@ -224,7 +224,7 @@ class Image extends React.Component {
         this.setState({ isVisible: true }, () =>
           this.setState({
             imgLoaded: imageInCache,
-            imgCached: this.imageRef.current.currentSrc && this.imageRef.current.currentSrc.length > 0,
+            imgCached: !!this.imageRef.current.currentSrc,
           })
         )
       })
--- a/packages/gatsby-image/src/index.js
+++ b/packages/gatsby-image/src/index.js
@@ -224,7 +224,7 @@ class Image extends React.Component {
         this.setState({ isVisible: true }, () =>
           this.setState({
             imgLoaded: imageInCache,
-            imgCached: this.imageRef.current.currentSrc.length > 0,
+            imgCached: this.imageRef.current.currentSrc && this.imageRef.current.currentSrc.length > 0,
           })
         )
       })


packages/gatsby-image/src/index.js
```diff
@@ -224,7 +224,7 @@ class Image extends React.Component {
         this.setState({ isVisible: true }, () =>
           this.setState({
             imgLoaded: imageInCache,
-            imgCached: this.imageRef.current.currentSrc.length > 0,
+            imgCached: !!this.imageRef.current.currentSrc,
```
pieh: this changes behaviour (`!![]` is true, because empty array is truthy). Adding sanity check seems better here?
```suggestion
            imgCached: this.imageRef.current.currentSrc && this.imageRef.current.currentSrc.length > 0,
```

pieh: I just noticed this is exactly what you had initially ;)

thebuilder: But `currenctSrc` is `DOMString`, not an `Array`. I'm thinking the initial solution was just another way of ensuring it wasn't an empty `string`

thebuilder: Either will work, but the `length` check just seems redundant for a string.

wardpeet:
I'm on ie11 and I don't get this issue as it's wrapped inside IOSupported which is never true on ie11. Nonetheless, it's an ok check

thebuilder:
> I'm on ie11 and I don't get this issue as it's wrapped inside IOSupported which is never true on ie11. Nonetheless, it's an ok check

The error i caught: https://sentry.io/share/issue/da7dc5b5859943e09889b2089f52ee3b/

There's a lot of moving parts, in `gatsby-image` - Most likely the issue is triggered by the stack i'm using it with, combining Contentful with an ObjectFit polyfill.

packages/gatsby-image/src/index.js
polarathene: @thebuilder It had to be a string that was not empty. Sorry for the confusion with how I approached that.

I had unfortunately not tested other cases outside of native IO support with desktop Chrome. I'm pretty sure other paths could cause it to not work correctly, which the new `loading` attribute for native browser lazy-load support may also introduce..

There is a longer more drawn out way to check for a cache, but adds much more delay to be reliable due to Promise / setTimeout to check if the `currentSrc` had been updated since.

Did you confirm that the caching detection was correctly working for when a cache was and was not present? Would it be better to compare to an empty string than the `!!` approach?(which isn't particularly clear at first glance with no comments, had to look up git blame to find this once it caught my eye).

thebuilder: The reason i went with `!!` (and we can add this as a comment to clarify), is that it validates that the value not `undefined`, `null` or empty `""`.

I think this is a fine way of checking if the src is cached. You already get an async delay by using the `IntersectionObserver`, no need to increase it.

I'd personally consider it a graceful fallback if the internal caching isn't working 100% in an old browser like IE, as long as it still loads the image.

polarathene: @thebuilder Ah ok, yeah a comment might be helpful to clarify that. Are there cases where it's not an empty string? Or is that just assumed?

> You already get an async delay by using the IntersectionObserver, no need to increase it.

IO is delaying when the main image element is "mounted"/rendered by flipping `isVisible` to true. Prior to that, the`currentSrc` isn't available. The ref that is assigned, isn't attached to anything until that point.

As mentioned, I believe there are some routes/paths in the code where I haven't tested, I don't have IE on hand either. Not a big loss though, just avoids some transition effect / rendering when the image is in the browsers cache(while the other cache logic is for run-time cache).

Thanks for catching this though and fixing it! :)

thebuilder: It should always be a (empty) string. But, as the issue is about, IE can have the `currentSrc` as `undefined` instead of an empty string.

PR #13289, opened 2019-04-11
fix: Get parent not root File node in resolver

### LINKED ISSUE #13267: File path infer problem
Assuming I have three node

```
A = {
internal: {type: "File"},
children: [B],
parent: null,
absolutePath: '/home/test/a'
}

B = {
internal: {type: "File"},
children: [C],
parent: A
absolutePath: '/home/test/b'
}
C = {
internal: {type: "Custom"}
children: [],
parent: B,
cover: './featured.png'
}
```

I hope C's property (cover) will be inferred to `/home/test/b/featured.png`, but not  `/home/test/a/featured.png`.

Is it possible get the `/home/test/b/featured.png` cover?
or 
have any custom condition function to stop find root node?

stefanprobst:
In the `fileByPath` resolver, we need a node's parent `File` node to resolve relative to absolute paths. For this we use `findRootNodeAncestor`, which will return a node's root node. This can be problematic for (unusual) scenarios like in #13267 where a node has more than one `File` node ancestor. We should just use the closest `File` parent.

Fixes #13267

### DIFF
--- a/packages/gatsby/src/schema/__tests__/queries-file.js
+++ b/packages/gatsby/src/schema/__tests__/queries-file.js
@@ -48,6 +48,30 @@ const nodes = [
     absolutePath: filePath(`test.txt`),
   },
   {
+    id: `file4`,
+    parent: `file5`,
+    children: [`test2`],
+    internal: {
+      type: `File`,
+      contentDigest: `file4`,
+    },
+    name: `parent.txt`,
+    dir: basePath,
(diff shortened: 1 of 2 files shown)


wardpeet:
Could we have a unit test for this one? In kitchen-sink perhaps?

stefanprobst:
Sure, test added!

PR #13294, opened 2019-04-11
fix(cpu-core-count): fallback to generic Node.JS method if system specific call fails

### LINKED ISSUE #12734: gatsby commands fail if "sysctl" isn't on the $PATH
`physical-cpu-core` doesn't try/catch around their system calls.

The package also isn't on GitHub so we can't PR a fix 🤦‍♂️ 

Here's the existing code. I recommend we vendor it with the fix:

```js
const os = require('os')
const childProcess = require('child_process')

function exec (command) {
  const output = childProcess.execSync(command, {encoding: 'utf8'})
  return output
}

let amount
const platform = os.platform()

if (platform === 'linux') {
  const output = exec('lscpu -p | egrep -v "^#" | sort -u -t, -k 2,4 | wc -l')
  amount = parseInt(output.trim(), 10)
} else if (platform === 'darwin') {
  const output = exec('sysctl -n hw.physicalcpu_max')
  amount = parseInt(output.trim(), 10)
} else if (platform === 'windows') {
  const output = exec('WMIC CPU Get NumberOfCores')
  amount = output.split(os.EOL)
    .map(function parse (line) { return parseInt(line) })
    .filter(function numbers (value) { return !isNaN(value) })
    .reduce(function add (sum, number) { return sum + number }, 0)
} else {
  const cores = os.cpus().filter(function (cpu, index) {
    const hasHyperthreading = cpu.model.includes('Intel')
    const isOdd = index % 2 === 1
    return !hasHyperthreading || isOdd
  })
  amount = cores.length
}

module.exports = amount
```

RikoSmith:
## Description
"Vendor"-ed physical-cpu-count package with try-catch envelops.

## Related Issues

Fixes #12734


wardpeet:
Thanks for you PR I cleaned up this PR a little bit to make it a bit nicer to look at. I also went for `it.each`.

RikoSmith:
Thanks for cleanup, @wardpeet ) Do you mean the same procedure as we did in PP session? So I need to add some mock for ````cpu.model.includes(`Intel`)```` inside existing fallback test or do I need to create separate test for it?

wardpeet:
same file but a new it. I'm thinking in the line of
```js
  it(`should check for hyperthreading when intel is the processor`, () => {
    const cProc = require(`child_process`)
    cProc.execSync.mockImplementation(() => {
      throw new Error(`Fail!`)
    })

    os.cpus.mockImplementation(() => [{ model: `Intel` }, { model: `Intel` }])
    mockPlatform(`linux`)
    const pcc = require(`../physical-cpu-count`)
    expect(pcc).toBe(1)
  })
```

RikoSmith:
Ok, I got it) Your test does pass, but this one does not: 
````javascript
it(`should return correct CPU count on Windows`, () => {
    const cProc = require(`child_process`)
    os.cpus.mockImplementation(() => [{ model: `Test` }])
    cProc.execSync.mockImplementation(
      () => `NumberOfCores

    4

    `
    )
    mockPlatform(`win32`)
    const pcc = require(`../physical-cpu-count`)
    expect(pcc).toBe(4)
  })
````
I guess there is a typo? 

Error message:
````javascript
physical-cpu-count › should return correct CPU count on Windows

    expect(received).toBe(expected) // Object.is equality

    Expected: 4
    Received: 0

      47 |     mockPlatform(`win32`)
      48 |     const pcc = require(`../physical-cpu-count`)
    > 49 |     expect(pcc).toBe(4)
         |                 ^
      50 |   })
      51 | 
      52 |   it(`should return fallback CPU count on Windows when childProcess fails`, () => {

      at Object.toBe (packages/gatsby/src/utils/__tests__/physical-cpu-count.js:49:17)
````
I corrected it to be (removed ````NumberOfCores```` and line breaks):
````javascript
it(`should return correct CPU count on Windows`, () => {
    const cProc = require(`child_process`)
    os.cpus.mockImplementation(() => [{ model: `Test` }])
    cProc.execSync.mockImplementation(() => `4`)
    mockPlatform(`win32`)
    const pcc = require(`../physical-cpu-count`)
    expect(pcc).toBe(4)
  })
````
and it worked. Should I leave this way or it supposed to be like in a previous form?

RikoSmith:
I'll push the changes, it will be clearer for inspection

pieh:
> and it worked. Should I leave this way or it supposed to be like in a previous form?

I think ideally we would replicate what `WMIC CPU Get NumberOfCores` command does on windows there, so we know that the function to parse that output actually works. I'll take a look now, but I think what Ward used there should be example output he got locally.

pieh:
Yeah, this is example output of command, so ideally we revert to previous mock and figure out why it fails exactly

Screenshot 2019-04-12 at 22 40 20

RikoSmith:
Yes, now I see it. Apparently the problem was in formatting of output. Thanks @pieh for explanation!  And thanks @wardpeet, @KyleAMathews  for PP sessions, it helped me a lot! I guess we can cancel the next Zoom meeting, the issue is successfully resolved)

PR #13299, opened 2019-04-11
docs: updated CSS links in tutorial part 2

eostrom:
## Description

docs: updated CSS links in tutorial part 2.

- steer users toward new Emotion instead of old Glamor
- add PostCSS plugin link

The [Gatsby Glamor tutorial](https://www.gatsbyjs.org/docs/glamor/) says Glamor is not supported and the (former) maintainer recommends Emotion. Seems prudent to steer new users to the Emotion tutorial instead.

PostCSS is popular and there's a plugin and a tutorial for it, so I added it to the "other options" list.

### DIFF
--- a/docs/tutorial/part-two/index.md
+++ b/docs/tutorial/part-two/index.md
@@ -309,7 +309,7 @@ CSS-in-JS is a component-oriented styling approach. Most generally, it is a patt
 
 #### Using CSS-in-JS with Gatsby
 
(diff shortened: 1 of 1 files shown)


pieh:
I agree it makes sense to recommend Emotion over Glamor. One thing that I think makes sense is to remove Emotion from "other css options" section if we already mention Emotion before

eostrom:
Fair point. I've revised this commit.

PR #13300, opened 2019-04-11
docs(gatsby-plugin-typography): New tutorial link

eostrom:
## Description

Typography.js is now in tutorial part 3, not part 2.

(The previous link was to an anchor within part 2, but this link is to the top of part 3. I chose not to link to [Using Plugins](https://www.gatsbyjs.org/tutorial/part-three/#using-plugins) because it's not explicitly a section about Typography.js, because it's not far into the document, and because I think the preamble provides useful context. But I could have gone either way.)

### DIFF
--- a/packages/gatsby-plugin-typography/README.md
+++ b/packages/gatsby-plugin-typography/README.md
@@ -2,7 +2,7 @@
 
 A Gatsby plugin for utilizing the [Typography](https://kyleamathews.github.io/typography.js/) library with minimal configuration.
 
-See it in action in the [Tutorial](https://www.gatsbyjs.org/tutorial/part-two/#typographyjs)
+See it in action in the [Tutorial](https://www.gatsbyjs.org/tutorial/part-three/)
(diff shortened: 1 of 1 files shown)


pieh:
What about https://www.gatsbyjs.org/tutorial/part-three/#-install-and-configure-gatsby-plugin-typography (which is subsection specifically dedicated to typography plugin)?

eostrom:
I looked at that too. My feeling was that, if I clicked a link to a tutorial, I would want a more explanatory page that I could follow step-by-step. If I just wanted basic installation and usage info, I would stay on the plugin page, which assumes I already have the background required to understand it.

If we link to the top of the tutorial, you get that explanation and those step-by-step instructions. If we link directly to the subsection, it's instructions for adding the plugin to the tutorial project and then customizing styles... but you've already missed the step where you created the project.

That said, it's easy enough to scroll up. The important thing to me is that the link point to the right page. So if you think the subsection is better, I'm happy to revise the PR.

PR #13309, opened 2019-04-11
feat(www): Remove `iframe` borders

fk:
Meh irt calling this a `feat` 😅 
This globally removes the border for `iframe`, making the GraphiQL embeds on https://www.gatsbyjs.org/docs/graphql-reference/ look a bit "nicer"; before/after:

image

image


fk:
Rare moments of clarity!  😅❤️

PR #13314, opened 2019-04-12
fix(gatsby): fix NODE_ENV & GATSBY_ACTIVE_ENV in webpack

### LINKED ISSUE #13273: Keying off GATSBY_ACTIVE_ENV breaks non production env builds
## Description

We have been utilizing GATSBY_ACTIVE_ENV for setting different staging/test environments.  PR [10565](https://github.com/gatsbyjs/gatsby/pull/10565) broke none productions builds for us due to Gatsby using it internally to determine whether to fetch data in .cache/loader.  

This was an easy fix for us as we simply use another variable for determining which environments we point our site too.  
### Steps to reproduce
[Example Repo](https://github.com/jakegibson/gatsbyActiveEnvBug)
With gatsby-plugin-offline installed. 
GATSBY_ACTIVE_ENV=staging gatsby build
gatsby serve

### Expected result
Should see rendered page

### Actual result
Can not read component-bundlename-index.js of undefined in .cache/loader.js  line 379

### Environment
gatsby >= 2.1.33

### Potential solution 
In multiple places .cache/loader splits off process.env.NODE_ENV !== `production`.  Seems like process.env.NODE_ENV !== `development` would make more sense as when we are testing in staging or test it is best practice to have a build that mirrors production.

### LINKED ISSUE #13304: GATSBY_ACTIVE_ENV vs. NODE_ENV – confusion
## Summary

As of #10565 Gatsby supports overriding the default `NODE_ENV` by specifying a `GATSBY_ACTIVE_ENV` environment variable. This is exactly the functionality I wanted when I was figuring out how to load my app's `.env.{env name}` config files into the gatsby build. However, now my build is breaking at runtime (seeing the same error as in #8285) because the webpack define plugin is getting the value of `GATSBY_ACTIVE_ENV` as `process.env.NODE_ENV` instead of the `NODE_ENV` set by calling `gatsby build`. It seems like whatever is set for `GATSBY_ACTIVE_ENV` should not replace the `NODE_ENV` value originally set by gatsby, right?

### Steps to Reproduce

Run `gatsby serve` with `GATSBY_ACTIVE_ENV=staging`, then try to hit one of your pages, you should see an error in `loader.js` saying that `syncRequires.components` is undefined.

### Expected Behavior

When running `gatsby build` or `gatsby serve`, `process.env.NODE_ENV` should be `production`, regardless of the value specified for `GATSBY_ACTIVE_ENV`. This behavior is what I expected based on my interpretation of the docs which say:

> As noted above NODE_ENV is a reserved environment variable in Gatsby as it is needed by the build system to make key optimizations when compiling React and other modules. For this reason it is necessary to make use of a secondary environment variable for additional environment support, and manually make the environment variables available to the client-side code.

It sounds like I should _never_ be able to change `NODE_ENV` no matter what I pass, yet clearly I can.

Happy to open a PR fixing this issue (and maybe cleaning up the docs a bit to clarify the use of `GATSBY_ACTIVE_ENV` & autoloading `.env`). I just want to make sure that the issue I'm noticing is actually a bug and not the intended behavior first.

wardpeet:
## Description
if GATSBY_ACTIVE_ENV is set, it's also used in webpack for NODE_ENV. This breaks our build because we rely on production or development mostly. This change will use `GATSBY_ACTIVE_ENV` to lookup the .env file but keep using NODE_ENV for webpack.

## Related Issues
fixes #13304
fixes #13273

### DIFF
--- a/packages/gatsby/src/utils/__tests__/webpack.config.js
+++ b/packages/gatsby/src/utils/__tests__/webpack.config.js
@@ -101,6 +101,10 @@ describe(`environment variables`, () => {
         expect.stringContaining(`.env.staging`),
         expect.anything()
       )
+      expect(DefinePlugin).toBeCalledTimes(1)
+      expect(DefinePlugin.mock.calls[0][0]).toMatchObject({
+        [`process.env.NODE_ENV`]: JSON.stringify(process.env.NODE_ENV),
+      })
     })
 
     it(`falls back to NODE_ENV`, async () => {
--- a/packages/gatsby/src/utils/webpack.config.js
+++ b/packages/gatsby/src/utils/webpack.config.js
@@ -38,17 +38,22 @@ module.exports = async (
 
   function processEnv(stage, defaultNodeEnv) {
     debug(`Building env for "${stage}"`)
-    const env =
-      process.env.GATSBY_ACTIVE_ENV ||
-      process.env.NODE_ENV ||
-      `${defaultNodeEnv}`
-    const envFile = path.join(process.cwd(), `./.env.${env}`)
+    // node env should be DEVELOPMENT | PRODUCTION as these are commonly used in node land
+    // this variable is used inside webpack
+    const nodeEnv = process.env.NODE_ENV || `${defaultNodeEnv}`
+    // config env is depednant on the env that it's run, this can be anything from staging-production
+    // this allows you to set use different .env environments or conditions in gatsby files
+    const configEnv = process.env.GATSBY_ACTIVE_ENV || nodeEnv
+    const envFile = path.join(process.cwd(), `./.env.${configEnv}`)
     let parsed = {}
     try {
       parsed = dotenv.parse(fs.readFileSync(envFile, { encoding: `utf8` }))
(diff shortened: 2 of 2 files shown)


packages/gatsby/src/utils/__tests__/webpack.config.js
```diff
@@ -101,6 +101,10 @@ describe(`environment variables`, () => {
         expect.stringContaining(`.env.staging`),
         expect.anything()
       )
+      expect(DefinePlugin).toBeCalledTimes(1)
+      expect(DefinePlugin.mock.calls[0][0]).toMatchObject({
+        [`process.env.NODE_ENV`]: JSON.stringify(process.env.NODE_ENV),
+      })
```
pieh: not blocking NIT: this could be
```
expect(DefinePlugin).toBeCalledWith(expect.objectContaining({
  "process.env.NODE_ENV": JSON.stringify(process.env.NODE_ENV),
})
```
(that allows us to not grab jest internals (`.mock.calls[0][0]`) )

PR #13326, opened 2019-04-12
feat(gatsby-dev-cli): Handle websocket upgrades

choffmeister:
## Description

This PR opens up for websocket proxying of the development server. Currently there is only access to the express app, but for websockets to be proxied one needs to be able to register a websocket upgrade handler with the HTTP server itself. With this PR the following will be possible:

```
// gatsby-config.js
const proxyMiddleware = require('http-proxy-middleware')

module.export = {
  developmentMiddleware: (app, server) => {
    const proxy = proxyMiddleware('/api', {
      target: 'http://localhost:8080',
      ws: true,
    });
    app.use(proxy);
    server.on('upgrade', proxy.upgrade);
  }
}
```

This PR is supposed to (and should) be backwards compatible.

### DIFF
--- a/docs/docs/api-proxy.md
+++ b/docs/docs/api-proxy.md
@@ -86,12 +86,12 @@ If you need to proxy websocket connections.
 var proxy = require("http-proxy-middleware")
 module.exports = {
   developMiddleware: (app, server) => {
-    const proxy = proxyMiddleware('/api', {
-      target: 'http://localhost:9000',
+    const proxy = proxyMiddleware("/api", {
+      target: "http://localhost:9000",
       ws: true,
     })
     app.use(proxy)
-    server.on('upgrade', proxy.upgrade)
+    server.on("upgrade", proxy.upgrade)
   },
 }
 ```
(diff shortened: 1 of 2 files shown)


pieh:
Super interesting. I wonder if this could somehow potentially mess with internal websocket that gatsby creates or webpack one, but I guess it's already possible right now.

Need to run few tests on this

KyleAMathews:
Curious what your use case(s) are?

choffmeister:
I have a search field on my Gatsby site and my backend serves the search results as they come as a stream of results, served via a websocket connection.

Though, now that I think about it again: I could also use a streamed HTTP response here and use `fetch()` to have the streaming effect...

wardpeet:
fetch is probably a better solution as you don't need a long lived connection. 👍

KyleAMathews:
You would just want to proxy your own websocket right? Which is already possible? So we just need to add your docs on how to do this?

KyleAMathews:
You can create multiple websocket connections from the browser.

KyleAMathews:
Would love a PR to add docs for how to proxy websocket connections but I'll close this one.

PR #13336, opened 2019-04-13
feat(showcase): add Pantene website

prasadseth:
added a Pantene website which uses gatsby to site showcase

### DIFF
--- a/docs/sites.yml
+++ b/docs/sites.yml
@@ -5372,3 +5372,12 @@
     - Portfolio
     - Personal Website
     - Photography
+- title: Pantene
+  main_url: https://pantene.com
+  url: https://pantene.com
+  featured: false
+  description: >
+    Pantene is a Swiss-created American brand of hair care products owned by Procter & Gamble
+  categories:
+    - Product Website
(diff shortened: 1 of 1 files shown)


prasadseth:
why 2 checks are failing ?

pieh:
@prasadseth The linting check (one that makes sure formatting of markdown / code is consistent) has some issues - I'm just running format on the PR to fix this

docs/sites.yml
```diff
@@ -5372,3 +5372,12 @@
...
     - Photography
+- title: Pantene
+  main_url: https://pantene.com
+  url: https://pantene.com
+  featured: false
+  description: >
+    Pantene is a Swiss-created American brand of hair care products owned by Procter & Gamble
+  categories:
+    - Product Website
+    - Hair care product website
```
pieh: It would be great if we could pick one of already existing categories and not create new ones that would contain single entry

My suggestion would be:
 - Beauty
 - Business

prasadseth: ok I will make changes in the file and create a new PR

PR #13338, opened 2019-04-13
Fix formatting of last row of props table

ShivamGoyal1899:


pieh:
This was already fixed in https://github.com/gatsbyjs/gatsby/pull/13327

Issue #12709, opened 2019-03-21
gatsby-plugin-canonical-urls query params and hash


MarkSalabutin:
Could you let me know why you added `location.hash` and `location.search` to canonical url? I faced with a problem that search engines treat pages `example.com/page` and `example.com/page?search=query` as different, but in practice, they are the same.

sidharthachatterjee:
From what I understand `location.hash` is ignored by most search engines and `location.search` should be included because its values often represent unique pages.

MarkSalabutin:
But what about cases when `location.search` don't represent unique page? For example, `fbclid` which facebook add to every external link. If canonical URL includes that parameter search engine will treat this URL as a different page.

sidharthachatterjee:
@MarkSalabutin That’s a fair. We could add options to configure this, I suppose. Would you like to pick this up?

Ehesp:
@MarkSalabutin @sidharthachatterjee just submitted a PR for this as I also ran into this issue. Would be good to get feedback.

LekoArts:
fixed by https://github.com/gatsbyjs/gatsby/pull/13339

PR #13339, opened 2019-04-13
[gatsby-plugin-canonical-urls] Only append search parameters if required

Ehesp:
[gatsby-plugin-canonical-urls]

cc @sidharthachatterjee @MarkSalabutin

## Description

This is a fairly opinionated PR. I can flip the logic if needed 👍 

Issue https://github.com/gatsbyjs/gatsby/issues/12709 makes a good point about the search params (e.g. `?tag=foobar`) being added to the canonical urls this plugin generates. From my limited SEO knowledge, the search params being included makes this plugin somewhat redundant. 

> A canonical tag (aka "rel canonical") is a way of telling search engines that a specific URL represents the master copy of a page. Using the canonical tag prevents problems caused by identical or "duplicate" content appearing on multiple URLs. Practically speaking, the canonical tag tells search engines which version of a URL you want to appear in search results. https://moz.com/learn/seo/canonicalization

Lets say I have the following page: `/blog`

If I add a tagging system, and have functionality for `/blog?tag=foobar`, search engines will still index this as a separate page, even with this plugin installed. The master reference is in-fact `/blog`.  This may not be the case all of the time, but I can't think of many cases, especially in Gatsby, where you'd want a "duplicate page" to be indexed when search params are different - it'd make more sense to create a totally different page/route.

These changes means search params won't be added by default, but gives the user the ability to set a flag adding them on. As mentioned above, I can flip this logic if you guys think it's too opinionated to do that 😄 

It would be good for someone else to test this, as breaking this functionality can have pretty serious consequences for websites SEO!

## Related Issues

https://github.com/gatsbyjs/gatsby/issues/12709

### DIFF
--- a/packages/gatsby-plugin-canonical-urls/CHANGELOG.md
+++ b/packages/gatsby-plugin-canonical-urls/CHANGELOG.md
@@ -3,6 +3,10 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [2.0.13](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-canonical-urls/compare/gatsby-plugin-canonical-urls@2.0.12...gatsby-plugin-canonical-urls@2.0.13) (2019-03-13)
+
+- **gatsby-plugin-canonical-urls:** URL search params only included if `search` option is `true` 
+
 ## [2.0.12](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-canonical-urls/compare/gatsby-plugin-canonical-urls@2.0.11...gatsby-plugin-canonical-urls@2.0.12) (2019-03-11)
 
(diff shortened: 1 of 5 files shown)


pieh:
Please don't change version and changelog manually - this happen automatically when we publish new release.

Given breaking nature of proposed change (in current shape) - I would vote for introducing option to strip search / query from canonical link (so it's not breaking for anyone who relies on current behavior) and do some research for any cases when having query string is desirable and figure out what's best default behavior - if we will end up deciding that striping search by default is best we would just flip default value for strip option and bump major version then.

Ehesp:
@pieh no problem will revert the changelog/package.json changes. 

That makes sense, I'll revert the logic so the params can be excluded.

Ehesp:
Updated to exclude search params if search is specifically set to `false`.

pieh:
What do you think about renaming `stripSearchParam` to `stripQueryString` as query string is actual name of that part of urls (so technically more correct :P)?

LekoArts:
@pieh changed the name :D

Issue #2426, opened 2017-10-12
gatsby-source-wordpress: Get menus


Unforgiven-wanda:
Hello again,
There was this [issue](https://github.com/gatsbyjs/gatsby/issues/1612) discussing the manner of getting menus and their locations, with @sebastienfi suggesting to use allWordpressWpApiMenusMenusExtended, but the plugin seems to have been updated since (the docs a bit later, which was confusing) because that node no longer exists.
So my question is: Is there, at the moment, any way to fetch the menus of Worpdress? Possibly according to their location? And if not is there an ETA on this feature?

Thank you.

Edit: I have just now noticed that the docs changed. However, with `WP Api Menus` installed on Wordpress, and the latest Gatsby version running, I can't run the basic example of `AllWordpressWpApiMenusMenuLocations`. Getting the error `Cannot query field \"allWordpressWpApiMenusMenusLocation\" on type \"RootQueryType\". Did you mean \"allWordpressWpApiMenusMenus\", \"wordpressWpApiMenusMenus\", or \"allWordpressWpStatuses\"?`

sebastienfi:
Here you go

```gql

allWordpressWpApiMenusMenusItems {
    edges {
      node {
        name
        count
        items {
          order
          title
          url
          wordpress_children {
            wordpress_id
            title
            url
          }
        }
      }
    }
  }

```

Then, in `render()` : 
`const siteMenus = this.props.data.allWordpressWpApiMenusMenusItems.edges`

Does this gives the data you expected ?

sebastienfi:
Closing this issue, please reopen if you feel that more information is needed :)

dmackerman:
It looks like the routes for the Wordpress plugin have changed?

This seems like the route for me:
`/wp-json/wp-api-menus/v2/menus`

braco:
Also getting an error from the query above
```
{
  "errors": [
    {
      "message": "Cannot query field \"allWordpressWpApiMenusMenusItems\" on type \"Query\". Did you mean \"allWordpressWpStatuses\", \"allWordpressWpTaxonomies\", \"allWordpressGutenbergSearch\", \"allWordpressWpTypes\", or \"wordpressWpStatuses\"?",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ]
    }
  ]
```

loganfromlogan:
[...]

PR #13343, opened 2019-04-13
feat(gatsby-source-wordpress): Add support for WP-REST-API V2 Menus

STUkh:
Add support for WP-REST-API V2 Menus

## Description

Add support for WP-REST-API V2 Menus that accept ACF inside menus ( WP-API-MENUS **doesn't** support ACF )
Do not conflicts with original WP-API-MENUS implementation
Can be enabled the same way with asterisk sign
Also includes the fix for media links paths, the same one as was for page URL host ~2 weeks before (VERY helpful for docker builds when _siteURL differs from localhost, but internal links goes there )
```
{
        resolve: 'gatsby-source-wordpress',
        options: {
        // ...
            includedRoutes: [
                '/*/*/categories',
                '/*/*/posts',
                '/*/*/pages',
                '/*/*/menus', // this will allow new menu items
                ...
          ],
       // ...
}
```
Adds `allWordpressMenusMenus` and `allWordpressMenusMenusItems` queries

## Related Issues

Partially solve #2426


LekoArts:
Hi!

In the interest of moving this along most effectively, we strongly encourage our contributors to provide an [example project](https://gatsby.dev/example-project) which we can then use to validate the changes in this PR. This will allow us to quickly and easily ensure that this does indeed address the purported feature or fix in this PR!

An example project:

- can be run locally or credentials have been provided
	- Feel free to message on Twitter, Discord, etc.
- can be used to validate this functionality
	- In other words, the necessary functionality has been enabled/disabled in the project for us to easily validate these changes

💜

packages/gatsby-source-wordpress/src/normalize.js
```diff
@@ -360,12 +362,21 @@ exports.searchReplaceContentUrls = function({
...
   const media = entities.filter(e => e.__type === `wordpress__wp_media`)
 
   return entities.map(e => {
     // Map featured_media to its media node
 
+    // Replace media URL if rely on localhost, when _siteURL isn't
+    if (
+      e.source_url &&
+      e.source_url.indexOf(`localhost`) !== -1 &&
+      _siteURL.indexOf(`localhost`) === -1
```
pieh: This is problematic - checking if localhost is used (IMO) seems to handle very specific usecase (docker).

We also can't switch every domain to use one used for API url, as there are some wordpress plugin that push images to CDNs.

What about this - in part of code that does discover routes, we could try to figure out what replacement is needed (based on passed API url and self link reported by main `/wp-json` endpoint) and use that to perform any url swaps?

STUkh: Agree, better to rethink this part and move to another PR, at least it is not related to main PR idea. But it works pretty well without any issues, anyway it weird situation when _siteUrl goes to custom domain and self links in response from custom domain get us to localhost.

pieh: If we could drop this change from this PR and move it separate one, we could merge the https://wordpress.org/plugins/wp-rest-api-v2-menus/ support faster. I'll go ahead and verify just the menu rest api part

pieh: Also another idea - we have this `searchReplaceContentUrls` plugin option that potentially would cover this case (but not as automatically as solutions discussed here). It would be good idea to test that one out (one problem I see is that search & replace is run after media mapping + downloading (currently), so we would need to move it before those (tho maybe there is reason for doing this way).

STUkh: Good idea. I'll try to do it. But for now i think we can remove this part and if everything else is ok - accept this PR. As improvement it would be good to update docs in source-wp plugin

pieh:
I've checked this locally and for the most part it works nicely, but there are few edge cases where it behaves weirdly - i.e. if you have hierarhical menus - i.e.:
Screenshot 2019-04-18 at 15 09 51

the rest endpoint return `items` as object with `"0"`, `"1"` and `"3"` as keys (not an array, which means that gatsby will create this pretty weird structure in schema:
Screenshot 2019-04-18 at 15 11 17

So we need to add some normalization to always convert `items` (and potentially `child_items`) to arrays, so schema doesn't change based on the shape of your menus

gustavpursche:
@STUkh @pieh Are there any plans to move forward with this PR? Is there anything I can support you with?

pieh:
@gustavpursche We would need to drop changes to pulling images (which are unrelated to WP-REST-API V2 Menus support) and also solve issue I described in https://github.com/gatsbyjs/gatsby/pull/13343#issuecomment-484502128. 

If you are willing to do so, you should be able to push to this PR as you are Maitainer in Gatsby repo using those instructions:

### Prerequisites

- [hub](https://github.com/github/hub) installed on your machine
- hub configured to use the ssh protocol for its commands: `git config --global hub.protocol ssh`
- An open pull request, e.g. `https://github.com/gatsbyjs/gatsby/pull/<prNumber>`
- The username of the PR author `<authorUsername>`

### Steps

- Add the author's remote: `hub fetch <authorUsername>`
- Check out the PR locally: `hub pr checkout <prNumber>`
- Make your changes and commit them
- Push your changes to the author's branch: `git push <authorUsername> <branch>`

### Troubleshooting

#### Allow edits from maintainers

The PR author must enable the [_Allow edits from maintainers_ option](https://blog.github.com/2016-09-07-improving-collaboration-with-forks/) on their pull request.

STUkh:
@pieh , @gustavpursche sorry for such a long delivery. Just removed part of code that discussed and actually should be part of another fix.

Issue #13221, opened 2019-04-08
feat(www): Add link to accessibility statement


marcysutton:
## Summary

PR #13190 adds a new page for an Accessibility Statement, but we decided to ship it without a link on the homepage until it gets some design attention. This issue is a reminder to do that, so people can actually find the accessibility statement! 

My proposed change (now removed) put a centered link at the bottom of the homepage. It would also be a good place to move the `.com` link potentially, as it would create more space at the top of the page and we could spell out "Gatsbyjs.com" like discussed in #12619. 

<img width="1633" alt="Accessibility statement link at bottom of homepage" src="https://user-images.githubusercontent.com/1045233/55742648-0226f080-59e5-11e9-9180-2f6da4e2cc50.png">

Ideally, the footer would exist on all of the .com site so that the links could be reached from any page. But the docs and plugins pages go full-height, and a plain text footer would have to be worked in somehow. @fk would love to get your input on this!

PR #13346, opened 2019-04-14
Blog post on accessibility statement & new footer links

marcysutton:
## Description

This PR introduces a blog post about our new accessibility statement, and also adds a link to that accessibility statement in the docs. I moved the .com link to a new footer component so the Accessibility Statement link wasn't lonely, and also solved an issue where `.com` wasn't super clear (in addition to freeing up space in the header). I took @fk's suggestion of putting the links into the main content area on pages with sidebars, similar to the React.js docs. I think the centered links look decent with a top border for some separation. Thoughts?

homepage footer
docs page with links in main content area

Because of the specific placement in various templates and layouts, I didn't see an easy way to do this in one place....hence the many files. If there's a cleaner way to do it, let me know (I probably missed a few as well but it seemed good enough for now).

## Related Issues

Closes https://github.com/gatsbyjs/gatsby/issues/13221
Closes https://github.com/gatsbyjs/gatsby/issues/12619


KyleAMathews:
👍 to adding a footer but we'd want to fill it out more. E.g. code of conduct, contribution guide, repeat top-level links, etc.

marcysutton:
@kylemathews we can totally do that. Can we add onto it incrementally, though, so it doesn’t hold up this PR (which also contains a blog post I’d like to publish next week)?

greglobinski:
Some suggestions: 

- unify the styles of separators
- match the footer width with the main contentment width
- differentiate the content of footer by aligning text to the right
- differentiate footer links by creating a new link style


Screenshot from 2019-04-15 00-19-20
Screenshot from 2019-04-15 00-19-07

DSchau:
@marcysutton 

> we can totally do that. Can we add onto it incrementally, though, so it doesn’t hold up this PR (which also contains a blog post I’d like to publish next week)?

IANK (I am not Kyle), but... yeah, I'd propose we do this incrementally but I don't want to de-value the importance of the footer and adding the accessibility statement in a visible place.

Perhaps @fk @greglobinski could work on a quick design that we could publish around the same time (in a separate PR) to this blog post next week? If I can get something in Figma I can add the footer, as well.

@fk @greglobinski any bandwidth? If not--no worries at all! Already appreciate you both chiming in with design ideas!

marcysutton:
@DSchau that sounds great! And thanks @greglobinski for the styling ideas, I went ahead and made some of the changes in this PR to keep it moving. Here are some comments:

> match the footer width with the main content width
Sounds reasonable! I made this change anywhere that I could–the only hack I wasn't stoked about was overriding top/bottom padding on the homepage. But I wanted to use the same breakpoints/etc. to match the content above, and using `!important` worked.

> mark out content of the footer by aligning text to the right
The reason I didn't go for right-aligned was for low-vision users who have to zoom the page...they would have to horizontally scroll to the right to find it. Centered looked better than left-aligned though, so i went for the middle.

I also adjusted the blog post date to Thursday the 18th. I'm hoping to publish this week to use some of the momentum we had from announcing it at conferences last week. Also I will be out of network next week so would be unable to field comments/questions.

greglobinski:
@marcysutton 

> The reason I didn't go for right-aligned was for low-vision users who have to zoom the page...they would have to horizontally scroll to the right to find it. 

I get it

@DSchau 

> could work on a quick design

Sure, but for the amount of content as it is now (two simple links) opening a design task seems unclear.

marcysutton:
Ok, I got the feedback in there @fk. I removed the bottom margin on all pages except the homepage. Added the footer to the showcase page, starters was already in there.

As for any other design tasks, I think what would be left would be to address @KyleAMathews' original comment about building the footer out more; the feedback given for this version has been adequately addressed.

Thanks all!

marcysutton:
Going to merge this so we can publish tomorrow! thanks again for the feedback.

PR #13357, opened 2019-04-14
feat(gatsby-transformer-asciidoc): add pathPrefix handling for images, define supported file extensions and add "page-*" attributes from asciidoc file to AsciiDoc node

lask79:
## Description
This PR provides 3 additional features to the asciidoc transformer:

- Handling of pathPrefix for the imagesdir attribute
- Define which file extensions are supported by the transformer (e.g. .adoc and .ad)
- Add all asciidoc attributes starting with "page-" to the AsciiDoc node

### Handling of pathPrefix
The current gatsby-transformer-asciidoc plugin does not handle pathPrefix for the images used in the asciidoc file. It is common in asciidoc to set an imagesdir as an attribute like this:

      resolve: `gatsby-transformer-asciidoc`,
      options: {
        attributes: {
          imagesdir: `/images`,
        },
      }
The default value is /images

The PR checks if a pathPrefix is set and prefixes the defined imagesdir with it before Asciidoctor.js is called.

### Define supported AsciiDoc file types
In the current implementation only .adoc files are supported but there are multiple file extensions used for asciidoc like .ad or .asciidoc. The PR offers a way to define which fileExtensions are used in the project:

      resolve: `gatsby-transformer-asciidoc`,
      options: {
        attributes: {
          imagesdir: `/images`,
        },
        fileExtensions: [`ad`, `adoc`],
      }
The default values are ['adoc','asciidoc']

### Add all asciidoc attributes starting with "page-" to the AsciiDoc node
In other static site generators like middleman or awestruct, which I have used in the passed, it was common that the developer/blogger could define attributes inside the asciidoc file without the need of frontmatter. The common way was to define attributes that started with "page-" like:
`page-tags`, `page-category`, `page-permalink` etc.

This PR takes all of the defined page-* attributes and adds them to the AsciiDoc node and adds them under the pageAttributes value. The prefix "page-" will be removed from the key name. This way those page attributes can be queried via GraphQL and be used for creating the pages like this:

    {
        allAsciidoc {
           nodes {
              pageAttributes {
                  path
              }
           }
        }
    }

### DIFF
--- a/packages/gatsby-transformer-asciidoc/src/gatsby-node.js
+++ b/packages/gatsby-transformer-asciidoc/src/gatsby-node.js
@@ -1,57 +1,63 @@
-"use strict";
+"use strict"
 
-const asciidoc = require(`asciidoctor.js`)();
+const asciidoc = require(`asciidoctor.js`)()
 
-async function onCreateNode({
-  node,
-  actions,
-  pathPrefix,
-  loadNodeContent,
-  createNodeId,
-  reporter,
-  createContentDigest
-}, pluginOptions) {
+async function onCreateNode(
+  {
+    node,
+    actions,
+    pathPrefix,
+    loadNodeContent,
+    createNodeId,
+    reporter,
+    createContentDigest,
+  },
+  pluginOptions
+) {
   var extensionsConfig = pluginOptions.fileExtensions
 
   // make extensions configurable and use adoc and asciidoc as default
-  var supportedExtensions = ( typeof extensionsConfig != 'undefined' && extensionsConfig instanceof Array ) ? extensionsConfig : [`adoc`, `.asciidoc`]
+  var supportedExtensions =
+    typeof extensionsConfig != `undefined` && extensionsConfig instanceof Array
+      ? extensionsConfig
+      : [`adoc`, `.asciidoc`]
   var currentExtension = node.extension
-  var isSupportedExtension = (supportedExtensions.indexOf(currentExtension) > -1);
+  var isSupportedExtension = supportedExtensions.indexOf(currentExtension) > -1
 
   if (!isSupportedExtension) {
-    return;
+    return
   }
 
   // changes the incoming imagesdir option to take the
-  var changedImagesDir = resolveImagesDir(pathPrefix, pluginOptions.attributes.imagesdir)
+  var changedImagesDir = resolveImagesDir(
+    pathPrefix,
+    pluginOptions.attributes.imagesdir
+  )
   pluginOptions.attributes.imagesDir = changedImagesDir
 
-  const {
-    createNode,
-    createParentChildLink
-  } = actions; // Load Asciidoc contents
+  const { createNode, createParentChildLink } = actions // Load Asciidoc contents
 
-  const content = await loadNodeContent(node); // Load Asciidoc file for extracting
+  const content = await loadNodeContent(node) // Load Asciidoc file for extracting
   // https://asciidoctor-docs.netlify.com/asciidoctor.js/processor/extract-api/
   // We use a `let` here as a warning: some operations, like .convert() mutate the document
 
-  let doc = await asciidoc.load(content, pluginOptions);
+  let doc = await asciidoc.load(content, pluginOptions)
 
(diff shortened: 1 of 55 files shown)


packages/gatsby-transformer-asciidoc/src/gatsby-node.js
```diff
@@ -1,34 +1,54 @@
...
 ) {
-  // Filter out non-adoc content
-  if (!node.extension || node.extension !== `adoc`) {
+  var extensionsConfig = pluginOptions.fileExtensions
+
+  // make extensions configurable and use adoc and asciidoc as default
+  var supportedExtensions =
+    typeof extensionsConfig != `undefined` && extensionsConfig instanceof Array
+      ? extensionsConfig
+      : [`adoc`, `.asciidoc`]
```
pieh:
```suggestion
      : [`adoc`, `asciidoc`]
```

packages/gatsby-transformer-asciidoc/src/gatsby-node.js
```diff
@@ -1,34 +1,54 @@
...
     loadNodeContent,
     createNodeId,
     reporter,
     createContentDigest,
   },
   pluginOptions
 ) {
-  // Filter out non-adoc content
-  if (!node.extension || node.extension !== `adoc`) {
+  var extensionsConfig = pluginOptions.fileExtensions
```
pieh:
```suggestion
  const extensionsConfig = pluginOptions.fileExtensions
```

packages/gatsby-transformer-asciidoc/src/gatsby-node.js
```diff
@@ -1,34 +1,54 @@
...
+
+  // make extensions configurable and use adoc and asciidoc as default
+  var supportedExtensions =
```
pieh:
```suggestion
  const supportedExtensions =
```

packages/gatsby-transformer-asciidoc/src/gatsby-node.js
```diff
@@ -1,34 +1,54 @@
...
   }
 
-  const { createNode, createParentChildLink } = actions
-  // Load Asciidoc contents
-  const content = await loadNodeContent(node)
-  // Load Asciidoc file for extracting
+  // changes the incoming imagesdir option to take the
+  var changedImagesDir = resolveImagesDir(
+    pathPrefix,
+    pluginOptions.attributes.imagesdir
```
pieh: `pluginOptions.attributes` will not always be defined, so we need to do some sanity checks to avoid "can't read imagesdir of undefined" errors

packages/gatsby-transformer-asciidoc/src/gatsby-node.js
```diff
@@ -1,34 +1,54 @@
+"use strict"
```
pieh: Please remove this from source file

pieh:
There is also PR bumping `asciidoctor` package major version ( https://github.com/gatsbyjs/gatsby/pull/13368 ). Which I tested against current version of the plugin and it doesn't break anything there. But I'm not sure if it will also work with those changes (need to test it - there doesn't seem to be CHANGELOG for `asciidoctor` package).

And final thing - we will need to update documentation (README.md) and ideally our example ( https://github.com/gatsbyjs/gatsby/tree/master/examples/using-asciidoc ) to use those new features you added:
 - probably not `pathPrefix` handling, but example of using `imagesdir` would be nice
 - page-attributes usage

lask79:
Sure I can update the documentation for this. I also bumped the version locally without any issues.

packages/gatsby-transformer-asciidoc/src/gatsby-node.js
lask79: sure

packages/gatsby-transformer-asciidoc/src/gatsby-node.js
lask79: good point

lask79:
I did all the recommended fixes and updated the documentation. Please have a look.

Issue #10727, opened 2018-12-30
Site Showcase permalink error


shannonbux:
## Description

The last several times I've clicked on "visit a site" from a site detail page and then, after browsing the site, use the back arrow to try and go back to the site showcase, there is just a blank white page staring at me!

### Steps to reproduce

1. Go to gatsbyjs.org/showcase
2. Click on any site
3. Click on "visit site" button
4. Once the site loads, click on back arrow

Voila, there is probably a white page.

### Expected result

You should see the detail page of the site that you just visited.
For example: https://www.gatsbyjs.org/showcase/www.howtographql.com (actually, @fk is this the correct URL I should expect for a permalink?)

### Actual result

A blank white page. Here's the console errors:

<img width="1275" alt="screen shot 2018-12-29 at 9 33 15 pm" src="https://user-images.githubusercontent.com/32467162/50544408-6c8fef80-0bb1-11e9-8e02-fc5438f658df.png">

### Environment

Chrome browser

jgierer12:
I can reproduce this in Chrome, but not Firefox.

sidharthachatterjee:
Breaks on `public-page-renderer-prod.js` because `pageResources` is `undefined`

@pieh Related to ensure resources?

sidharthachatterjee:
This only occurs when visiting a site via the modal, normal page works fine. 

@pieh mentioned that we need to put we put _shouldShowModal_ in history 

Still strange that it works on Firefox though

pieh:
Also for reference: https://github.com/gatsbyjs/gatsby/pull/7501
> Using directly has the caveat that you can't use it
unless the page resources are already loaded. [...] Otherwise it'll break.
which is exact scenario that is happening here

sidharthachatterjee:
Firefox seems to be doing some magic by persisting a snapshot of the page and using that when one hits the back button. Refreshing on Firefox also causes this issue.

pieh:
We have couple of possibilities here:
 - wrap public PageRenderer in EnsureResources - we might hit same problems we hit before with react constantly updating and re-rendering that wasn't really figured out - need to research that.
 - add check if resources exist in public PageRenderer and return `null` if we don't have resources - also would need to adjust `www` code to fallback to non-modal version of showcase details page if background page would be `null`

laradevitt:
Just came across this. Some additional info, FWIW...

[...]

PR #13358, opened 2019-04-14
(www)Showcase:  Set target "_blank" to site link on showcase-details

nephlin7:
## Description
 Set target "_blank" to site links on showcase-details  to keep the user on the gatsby page and prevent this error
image

## Related Issues
Related #10727


marcysutton:
I'd rather get to the bottom of the error than change to external links if possible–there have been a lot of discussions about that already. Is there a way to fix it without the `target="_blank"`?

KyleAMathews:
Yeah, if people are seeing errors clicking back, then this is a core bug to fix.

Can you reliably reproduce this? If so, it's most likely something in gatsby-plugin-offline.

PR #13375, opened 2019-04-15
refactor(gatsby): Move query/page-query-runner to query/index

Moocar:
## Description

In https://github.com/gatsbyjs/gatsby/pull/13064, I mentioned that we should move `page-query-runner.js` to `query/index.js` since it is becoming a general interface for running queries, whether static, or page. This PR makes that happen. 

Note, I'm not sold on requiring `query/index` as `queryUtil` as I have in this PR, but can't think of a better name. Very much open to suggestions.

## Related Issues

- builds on https://github.com/gatsbyjs/gatsby/pull/13064


Moocar:
Published in gatsby@2.3.24

PR #13385, opened 2019-04-16
chore(www): clean commented out code

prashant-andani:
Cleaned commented out code in multiple files


fk:
Thank you @prashant-andani! 🙏

PR #13388, opened 2019-04-16
chore(www): Align more design tokens with system-ui spec (refactor breakpoints to contain raw values, add `presets.mediaQueries` derived from it for convenience)

fk:
Following up on #13107 and https://github.com/gatsbyjs/gatsby/pull/13107#discussion_r272737947, this refactors `www/src/utils/tokens/breakpoints.js` to contain the "raw" breakpoint values, and moves the former `breakpoints` convenience strings (ready to be used as CSS media queries) to `www.src/utils/tokens/media-queries.js` and `presets.mediaQueries`.

Ref. https://system-ui.com/theme/
/cc @jxnblk @johno


fk:
Ah sweet, thank you!

Issue #7313, opened 2018-08-14
Using Gatsby Image with Picture tag?


deadcoder0904:
## Summary

I want to use `gatsby-image` with the `picture` tag. I need to use `picture` tag because I have different images for mobile, tablet & desktop resolutions.

## Relevant information

This is the current way I do it. Is this the right approach?

```js
<picture>
        <source
          media="(max-width: 560px)"
          srcSet={data.mobile.childImageSharp.fluid.src}
        />
        <source
          media="(max-width: 875px)"
          srcSet={data.tablet.childImageSharp.fluid.src}
        />
        <Img src={data.desktop.childImageSharp.fluid.src} />
</picture>
```

Because its used like this in the docs & I don't know if the method above is right

```js
<Img resolutions={data.file.childImageSharp.resolutions} />
```

I need 2 things from what I'm doing:

1. Optimized Images with Traced SVG in Pages
2. Different Images for Mobile, Tablet & Desktop

Is it right what I'm doing or is there another approach?

KyleAMathews:
gatsby-image doesn't support "art direction" images like this as it seemed like it might overly complicate the API so I left it out.

If you can come up with a nice API, I'd love to explore adding it. Otherwise, your approach looks great! Except, make sure to query `srcSet` in addition to src and pass it to the source `srcSet`.

deadcoder0904:
Hey @KyleAMathews the current way I showed above is wrong as I was not using `Img` from `gatsby-image` but a `styled-component` lol 😂

And AFAIK `Img` from `gatsby-image` is not a replacement for `img` from `DOM`. So I don't know how to get this thing work?

Currently, I am doing:

```js
<picture>
        <source
          media="(max-width: 560px)"
          src={data.mobile.childImageSharp.fluid.src}
          srcSet={data.mobile.childImageSharp.fluid.srcSet}
        />
        <source
          media="(max-width: 875px)"
          src={data.tablet.childImageSharp.fluid.src}
          srcSet={data.tablet.childImageSharp.fluid.srcSet}
        />
        <CustomImg fluid={data.desktop.childImageSharp.fluid} />
</picture>
```

where my custom image is a `styled-component` around `Img` from `gatsby-image` like:

```js
const CustomImg = styled(Img)`
  width: 5rem;
`;
```

The issue is I cannot show Mobile or Tablet images because of that. Only Desktop Image shows up even when screen is reduced. What's the workaround?

Issue #4491, opened 2018-03-12
[gatsby-image] Art direction


Yurickh:
## Description

I needed a way to set up gatsby-image to work alongside _art direction_. That is, use different image sources for different display sizes.

Currently `gatsby-image` only renders an `img` element with some div wrappers with a beautiful srcset created from different resolutions from a given image.

My suggestion is to open up to receive more than one image and wrap it up in a `picture` tag, with different `source`s.

I don't know if this is in line with the purpose of the plugin, but would surely be nice to have all the capabilities of the plugin with the possibility of the art direction.

### Expected result

I should be able to render from a list of different images depending on screen size/resolution.

### Actual result

But I'm currently able to choose only one image to render. In order to achieve a similar result, I need to render multiple images and `display: none` based on media queries, which is sub-optimal since I don't want platforms downloading images that they won't even show up.

KyleAMathews:
This would be excellent!

I'm totally open to adding this to gatsby-image. I wonder if it might be simpler in a separate component? I'm not sure — I haven't used art direction ideas on websites before — but it's not a bad thing to have more than one component.

BTW, do browsers really download images that have `display: none`? I thought they didn't.

Yurickh:
> BTW, do browsers really download images that have display: none? I thought they didn't.

AFAIK, they still do it yeah.
From what I've searched, it's not common behaviour to stop downloading images with `display: none`. [This site](https://www.quirksmode.org/css/displayimg.html) is a great way to check if your does.

> I wonder if it might be simpler in a separate component?
It would probably be a fork, or something, since I loved the svg pre-render and the imagesharp api, and really wanted to have that added to the art direction as well.

skube:
Tools like http://www.responsivebreakpoints.com/ are handy for generating the appropriate code and images.

It would be the holy grail if `gatsby-image` could handle art-direction on top on what it already amazingly does.

KyleAMathews:
Due to the high volume of issues, we're closing out older ones without recent activity. Please open a new issue if you need help!

Issue #362, opened 2016-07-22
Add eslint-loader


KyleAMathews:
We ship starters with eslint configs but they're basically useless unless a user has their editor setup to support Eslint. Add eslint-loader would change that as errors would then show up in the console/browser. If we do this however, the eslint config in starters should probably be a lot less opinionated though so we're not pointlessly pushing the arbitrary eslint choices I've made here.

https://github.com/MoOx/eslint-loader

KyleAMathews:
I really like create-react-app's approach here as they define their own eslintrc that's focused on what will cause bugs not styling problems. We should just copy their approach. I think we can even just depend on their configuration directly.

https://github.com/facebookincubator/create-react-app/blob/c5f5b006d83316b3c016adc41b95c28912ecdda8/packages/react-scripts/config/webpack.config.dev.js#L106

https://github.com/facebookincubator/create-react-app/blob/c5f5b006d83316b3c016adc41b95c28912ecdda8/packages/eslint-config-react-app/index.js

karanjthakkar:
@KyleAMathews I think I can pick this up if you just want to shadow the config create-react-app has. Is there anything else you'd want to be considered in this?

KyleAMathews:
Thinking about it more, we might not want to use their exact lint rules as I think they disallow some things that we wouldn't want to disallow.

How about instead of directly depending on their config, you just copy it into our project so we can change things as needed.

f0rr0:
Also consider checking out https://github.com/sindresorhus/xo. It comes with great defaults and the logging is really pretty. There is loader similar to eslint for webpack as well.

MoOx:
Xo is just an eslint config at the end ;)

f0rr0:
@MoOx exactly the config can be dropped in as well if not the entire thing. https://github.com/sindresorhus/eslint-config-xo

KyleAMathews:
XO is very nice. I haven't mentioned it before (was planning on making a new issue but perhaps we can just do it here) but I *really* like how Dan added support for showing lint errors in the browser in an overlay — https://twitter.com/dan_abramov/status/780017771447025672

I think we should copy that. It's much better than just showing them in the browser/terminal console.

f0rr0:
[...]

Issue #155, opened 2016-02-26
Update README example for modifying Webpack to one that isn't now supported in core


KyleAMathews:
It still shows how to extract a `styles.css` file.

PR #13395, opened 2019-04-16
feat(gatsby-image): Add art direction

### LINKED ISSUE #13164: [gatsby-image] Add art direction
## Summary

[Art direction](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#Art_direction) refers to serving different source images at different breakpoints, in order to show users a different crop or image depending on their screen size.

### CSS `display: none` ❌ 

While art direction can be approximated from CSS using media queries and `display: none`, browsers (at least Chrome and Safari) will still load an image that has `display: none`. This is suboptimal from a performance and user-experience perspective.

**Art Direction Loading Example:** https://codesandbox.io/s/lp6746pkw9

### HTML `<picture>` and `<source>` ✅
A better way to accomplish art direction is by using a `<picture>` tag containing a `<source>` element per crop, with an accompanying `media` attribute. This allows the browser to choose the correct source to display depending on the matching `media` string.

With a few modifications, `gatsby-image` can support art direction for an arbitrary number of image crops.

### High-Level Changes Needed
1. Add an optional `media` key to both `FluidObject` and `FixedObject`, to pass in a media query string that corresponds to a specific image for the browser to match on.
1. Allow `gatsby-image` to take `Array<FluidObject>` and  `Array<FixedObject>`, to generate a `<source>` per image crop.

### Basic example

Given a setup like this:
```ts
const MOBILE_QUERY = `(max-width: 767px)`;
const DESKTOP_QUERY = `(min-width: 768px)`;

// images here can be either fluid or fixed
const sources = [
  {
    ...mobileImage.fluid,
    media: MOBILE_QUERY,
  },
  {
    ...desktopImage.fluid,
    media: DESKTOP_QUERY,
  },
];
```
#### Option 1: Introduce new `fluidSources` / `fixedSources` props

```js
<Image
  alt={mobileImage.description}
  fluidSources={sources}
/>
```

Having two new props to support this leads to some potential footguns (what if a user passes both `fluid` and `fluidSources`?), and additional branching logic in `gatsby-image`.

#### Option 2: Augment existing `fluid` / `fixed` props to take an array of images

```js
<Image
  alt={mobileImage.description}
  fluid={sources}
/>
```

This option removes the need to introduce any new props, and would allow the internals of `gatsby-image` to be refactored to always operate in terms of an array of images. Passing a single image here would just be treated as an array of length 1.

#### Option 3 (Not Considered): New `multi-image` Component
If this were a separate component, it would want to have many of the same features as `gatsby-image` does (image cache, lazy loading, base64 / tracedSVG, etc.). This would seem to require a much larger refactor of `gatsby-image` to make these features shareable, and would increase the barrier to adopting art direction incrementally.

### Motivation

This feature has been previously requested (#4491, #7313), with some indication of support from @KyleAMathews .

By having `gatsby-image` handle this directly, users won't need to choose between using all of `gatsby-image`'s many benefits (lazy loading, `critical`, CMS integrations, etc.), and accomplishing art direction.

We needed functionality at Harry's for the new homepage hero on https://www.shopflamingo.com/ (you can resize the browser window to see the image change). We forked `gatsby-image` to add this functionality, and would love to contribute it back to `gatsby-image` directly.

The fork would still need a little bit more work (it's currently only implemented for `fluid` images), but I'm more than happy to open a PR to start that process! Would love any especially love any direction on the proposed APIs.

brimtown:
## Description
This PR adds art direction capabilities to `gatsby-image`.

See issue #13164 for additional background and discussion.

### Decisions Made
1. Introduce new `fixedImages` and `fluidImages` props, instead of overloading the existing `fixed` and `fluid`
1. Only generate additional `<sources>` if there is more than one variant. This keeps the snapshots for the existing tests the same, except for the addition for a single `<picture>` tag for base64 / tracedSVG placeholders.
1. Use the first image's `src` as the key for the image cache, instead of creating new entries in the cache per `source`

I'm open to revisiting any of these, as each one is a tradeoff.

## Related Issues

Fixes #13164, #4491, #7313


packages/gatsby-image/src/__tests__/index.js
```diff
@@ -56,8 +124,18 @@ describe(`<Image />`, () => {
...
+  })
+
+  it(`should render multiple fluid image variants`, () => {
+    const component = setupImages(true)
+    expect(component).toMatchSnapshot()
+  })
+
   it(`should have correct src, title, alt, and crossOrigin attributes`, () => {
-    const imageTag = setup().querySelector(`picture img`)
+    const imageTag = setup().querySelectorAll(`picture img`)[1]
```
brimtown: This was necessary since now there are two `<picture>` tags for images with base64 or tracedSVG placeholders. Feels a little janky to me, and I'd be open to ideas on refactoring.

packages/gatsby-image/src/__tests__/index.js
```diff
@@ -45,6 +87,32 @@ const setup = (fluid = false, onLoad = () => {}, onError = () => {}) => {
   return container
 }
 
+const setupImages = (
```
brimtown: It seemed easier to add this as a separate `setupImages` function, but I could also refactor the original `setup` to handle this case too.

packages/gatsby-image/src/index.js
```diff
@@ -13,6 +13,13 @@ const convertProps = props => {
     delete convertedProps.sizes
   }
 
+  if (convertedProps.fixedImages) {
```
brimtown: This was to avoid refactoring the existing places where we check for or use `fluid` / `fixed`, namely when deciding what to render, and in for the image cache.

packages/gatsby-image/src/__tests__/__snapshots__/index.js.snap
```diff
@@ -54,19 +56,157 @@ exports[`<Image /> should render fluid images 1`] = `
       style="background-color: lightgray; position: absolute; top: 0px; bottom: 0px; opacity: 1; right: 0px; left: 0px; transition-delay: 0.5s;"
       title="Title for the image"
     />
-    <img
-      alt=""
-      class="placeholder"
-      src="string_of_base64"
-      style="position: absolute; top: 0px; left: 0px; width: 100%; height: 100%; object-fit: cover; object-position: center; opacity: 1; transition-delay: 0.5s; color: red;"
+    <picture>
```
brimtown: I think the diff from here to line 91 is a bit misleading; all that changed for the existing snapshots was the addition of a `<picture>` tag around the base64 placeholder.

brimtown:
I've done some manual QA on this by plugging it into a staging build of shopflamingo.com and testing across Chrome and Safari. Is there a better way to set up some shareable test links for this?

Also, would this PR be the place to add additions to the [docs](https://www.gatsbyjs.org/docs/using-gatsby-image/) / [the demo site](https://using-gatsby-image.gatsbyjs.org/)?

brimtown:
I'm also not totally clear on the `starter_validate` test failure, it seems like it may be unrelated to this change?

polarathene:
Review Summary:

**TL;DR:** version, links to parts of the review for further information.

## Questions / Concerns

- **Potential breakage** concern on update for users from the introduction of wrapping `img` elements in `picture`s.
- The placeholder `generate..()` methods are noops for instances not providing variants, while the noscript and image differ by handling non-variant cases, I don't like the inconsistency there, nor with handling of props/attributes.
- Are there any concerns with placeholder styles due to the `picture` element wrapping them? Should they stay on `img` or be moved to `picture`?
- Both `fluid` and `fixed` conditionals in the render have logic depending on `image` props that aren't supporting image variants with different widths/heights/aspectRatios. Discussed [here](https://github.com/gatsbyjs/gatsby/pull/13395#discussion_r277528760)
- Current implementation is creating both sources for the first image as well, but the non-webp source was removed in a prior commit due to Safari bug, will this introduce a regression? Discussed [here](https://github.com/gatsbyjs/gatsby/pull/13395#discussion_r277539433)

## Change suggestions

- Use the generate sources methods to handle the variants only instead of catering to all sources which requires special handling affecting readability and consistency/predictability of how the group of methods behave. Add a clear `&&` guard in the JSX so it's clearly separated as logic for handling multiple sources(helpful for when we refactor and split this file up).
- Drop the copy/override behaviour of `convertProps()` that assigns first `fluidImages/fixedImages` element to `fluid`/`fixed`. 
- Refactor `generateNoscriptSources()` to one of the alternatives suggested in review [here](https://github.com/gatsbyjs/gatsby/pull/13395#discussion_r277513429) (also adjust for additional concerns raised on that link).
- Refactor the start of `noscriptImg()` as suggested in review [here](https://github.com/gatsbyjs/gatsby/pull/13395#discussion_r277521230).
- Refactor `generateImageSources()` to one the suggestion in review [here](https://github.com/gatsbyjs/gatsby/pull/13395#discussion_r277516418).
- Refactor the rendering for the main image (picture) element as suggested in review [here](https://github.com/gatsbyjs/gatsby/pull/13395#discussion_r277518050), note comment to remove a guard.
- Refactor the start of `fluid` and `fixed` render conditionals, to have a base image var like we currently do, and variants as array(with either 1st element included or removed, based on other refactoring), as discussed in review [here](https://github.com/gatsbyjs/gatsby/pull/13395#discussion_r277523835).

## Misc

Not required for the PR to be approved, but related for future work:

- May want to include `type` attribute for all sources? Discussed [here](https://github.com/gatsbyjs/gatsby/pull/13395#discussion_r277536253).

packages/gatsby-image/src/index.js
```diff
@@ -69,6 +76,95 @@ function getIO() {
...
+        media={variant.media}
+        srcSet={variant.srcSetWebp}
+        sizes={variant.sizes}
+      />
+    ) : (
+      undefined
+    )
+
+    const nonWebPSource = (
+      <source
```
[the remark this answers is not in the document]
tbrannam: When converting gatsby-image to use picture tag originally - I reasoned that the last source was the 'most compatible' and very likely the same image format that would be used in the fallback img tag as well.   Seems reasonable to add I suppose - just keeping in mind that it isn't always Jpeg.

packages/gatsby-image/src/index.js
```diff
@@ -69,6 +76,95 @@ function getIO() {
...
+        key={variant.src}
+        media={variant.media}
+        srcSet={variant.srcSet}
+        sizes={variant.sizes}
+      />
+    )
+
+    return [webPSource, nonWebPSource]
```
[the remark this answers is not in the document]
tbrannam: @brimtown and I can take a look at that - to check for regressions.

packages/gatsby-image/src/index.js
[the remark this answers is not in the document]
polarathene: > Seems reasonable to add I suppose - just keeping in mind that it isn't always Jpeg.

The mimetype can probably be added to the metadata object during build time, so that it's just another property that gets used if available? There is an issue with that and sharp code atm, where the mozjpeg support may not process a png input, but output a file with jpg extension that's still png data, without raising an error or warning.

There is a mimetype crate that I linked to on the relevant issue about it that could help avoid that.

packages/gatsby-image/src/__tests__/index.js
polarathene: At the very least the `1` should not be used directly, assign it to a const that identifies what the index is intended to represent.

packages/gatsby-image/src/index.js
```diff
@@ -332,24 +430,24 @@ class Image extends React.Component {
 
           {/* Show the blurry base64 image. */}
           {image.base64 && (
-            <Img src={image.base64} {...placeholderImageProps} />
+            <picture>
+              {generateBase64Sources(imageVariants)}
+              <Img src={image.base64} {...placeholderImageProps} />
```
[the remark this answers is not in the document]
brimtown: From [what I can find](https://stackoverflow.com/questions/25396605/when-targeting-a-picture-element-in-css-should-we-use-img-or-picture-selector) on StackOverflow, people generally recommend applying styles to the `` instead of the `<picture>`, partly to avoid the issue that you raised, and partly because it's ultimately the `` tag that gets rendered, not the `<picture>`. We can think of it as metadata to the browser instead of an element that we need to target with styling, so I think keeping the styles on `` still makes sense here.

And since `<picture>` elements are [inline](https://developer.mozilla.org/en-US/docs/Web/HTML/Inline_elements#Elements), it won't have any impact on layout like a `<div>` would.

packages/gatsby-image/src/index.js
```diff
@@ -69,6 +76,95 @@ function getIO() {
...
+    const webPSource = variant.srcSetWebp ? (
+      <source
+        key={`${variant.src}-webp`}
+        type="image/webp"
+        media={variant.media}
+        srcSet={variant.srcSetWebp}
+        sizes={variant.sizes}
+      />
+    ) : (
+      undefined
```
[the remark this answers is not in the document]
brimtown: Good catch, this should just be `null`.

packages/gatsby-image/src/index.js
```diff
@@ -332,24 +430,24 @@ class Image extends React.Component {
 
           {/* Show the blurry base64 image. */}
           {image.base64 && (
-            <Img src={image.base64} {...placeholderImageProps} />
+            <picture>
```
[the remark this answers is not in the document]
brimtown: Yeah, I'm curious if the internal structure of this package is considered public API. It would be fairly straightforward to switch based on `if (fixedImages || fluidImages)`, but it may look a bit messier.

packages/gatsby-image/src/index.js
```diff
@@ -69,6 +76,95 @@ function getIO() {
   return io
 }
 
+function generateImageSources(imageVariants) {
+  const initialImage = imageVariants[0]
+
+  if (imageVariants.length === 1) {
```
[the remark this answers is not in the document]
brimtown: I was thinking that it felt weird to me to spread the source generation logic across two places. I was choosing to combine all source generation logic into the `generateImageSources` function, instead of keeping the implementations separate based on whether there was art direction required or not.

That said, I actually really like the approach you've proposed, and seeing it written out makes it seem much cleaner than what I had. I'll update it to match.

packages/gatsby-image/src/index.js
```diff
@@ -69,6 +76,95 @@ function getIO() {
...
+    return initialImage.srcSetWebp ? (
+      <source
+        type={`image/webp`}
+        media={initialImage.media}
```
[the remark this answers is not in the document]
brimtown: Yeah it's a good question, I was thinking about an edge case where someone wanted to apply `media` to an image but not with multiple images? I can't imagine a legitimate use case for that though, happy to keep all `media` stuff with the multiple variants cases.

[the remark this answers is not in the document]
polarathene: I'm not against supporting it, it's not supported on the img element though, so wouldn't that require providing both sources for the first image too?

packages/gatsby-image/src/index.js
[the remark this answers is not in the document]
polarathene: > I was thinking that it felt weird to me to spread the source generation logic across two places. 

Yeah, I understand that. It was just not as readable since you had to conditionally handle for 2/4 of those methods for returning a webp source only with the other two methods returning null.

packages/gatsby-image/src/index.js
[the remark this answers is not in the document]
polarathene: If the core devs are ok with it, I think it's fine for the most part. Lock files should avoid the few that are likely to be affected with existing projects.

packages/gatsby-image/src/index.js
```diff
@@ -293,6 +390,7 @@ class Image extends React.Component {
 
     if (fluid) {
       const image = fluid
+      const imageVariants = fluidImages || [fluid]
```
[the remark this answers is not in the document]
brimtown: We do need to add `media` to the initial source when there are image variants, and the order of the sources matters as well.

We could still split them up between the first image and the rest, but we'd need to choose between 1) always rendering a `<source>` for the initial image in case there is a `media` element on it, or 2) conditionally rendering the first `<source>` depending on if there is a `media` element.

1) would increase the rendered HTML size for all consumers regardless of if they were utilizing variants, so I'll work towards accomplishing this with 2).

packages/gatsby-image/src/index.js
```diff
@@ -69,6 +76,95 @@ function getIO() {
...
+  if (imageVariants.length === 1) {
+    return null
+  }
+
+  return imageVariants.map(variant => (
+    <source key={variant.src} media={variant.media} srcSet={variant.base64} />
+  ))
+}
+
+function generateNoscriptSources(imageVariants, sizes) {
```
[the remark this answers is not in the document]
brimtown: I will refactor it to one of the ones you suggested! Good callout on `sizes`.

Regarding what happens if a `media` key is not passed: it will have the same behavior as the plain HTML `<picture>` tag would, which means the browser would just pick the first `<source>` it is capable of rendering. It does put the onus on the user to pass `media`, but no more so than `<picture>` itself does.

packages/gatsby-image/src/index.js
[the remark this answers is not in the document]
polarathene: Bit late on responding to this but always including the extra source for the initial image isn't that big of a deal is it? A lot of it should compress well which should be a standard practice.

Consider the weight added when gzipped which is the most common compression format. It shouldn't be all that high. this also only affects noscript/critical support as lazy loading support uses isVisible before rendering markup for the main image. Placeholders would be affected though I guess.

packages/gatsby-image/src/index.js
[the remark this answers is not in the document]
brimtown: The order matters for these `<sources>`, and we also need to render both the Webp and an additional `<source>` for the first image to make sure that art direction works. So it actually needs to look more like this.

```js
{this.state.isVisible && (
  <picture>
    {image.srcSetWebp && (
      <source
        type={`image/webp`}
        media={image.media}
        srcSet={image.srcSetWebp}
        sizes={image.sizes}
      />
    )}
    {image.media && (
      <source
        media={image.media}
        srcSet={image.srcSet}
        sizes={image.sizes}
      />
    )}
    {imageVariants.length > 0 && generateImageSources(imageVariants)}
    <Img
      alt={alt}
      title={title}
      sizes={image.sizes}
      src={image.src}
      crossOrigin={this.props.crossOrigin}
      srcSet={image.srcSet}
      style={imageStyle}
      ref={this.imageRef}
      onLoad={this.handleImageLoaded}
      onError={this.props.onError}
      itemProp={itemProp}
    />
  </picture>
)}
```

Similarly for the base64 / tracedSVG placeholders:

```js
{image.base64 && (
  <picture>
    {image.media && (
      <source media={image.media} srcSet={image.base64} />
    )}
    {imageVariants && generateBase64Sources(imageVariants)}
    <Img src={image.base64} {...placeholderImageProps} />
  </picture>
)}
```

Does that seem acceptable to you? We could also remove the `image.media?` check from these and just render the first source by default (assuming no regressions).

packages/gatsby-image/src/index.js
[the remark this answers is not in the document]
brimtown: It would, but we need to generate that initial source anyway if the user is using image variants. I'm happy to always generate that initial source regardless of if they're using variants or not (assuming no regressions on Safari). I had just assumed we'd want to keep the current no-`<source>` behavior for people not using variants though.

packages/gatsby-image/src/index.js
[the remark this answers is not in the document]
polarathene: @brimtown Yeah, that looks good. React will ignore the media attribute if the prop isn't available, it's just with the current noscript support via string template that it needs to be handled conditionally.

If there were any issues from this approach, a refactor can be done later. The additional source element should be a non-issue. All the duplicated fields and values should compress fine afaik.

polarathene:
Reviewing the [changes](https://github.com/gatsbyjs/gatsby/compare/master...brimtown:changes) branch commits [6282c75](https://github.com/gatsbyjs/gatsby/commit/6282c752fed09a63233becd9641298ac2a33aeef):`Respond to code review` and [fc12f62](https://github.com/brimtown/gatsby/commit/fc12f62547b14580af95f7bf460891d664711f52):`Use const in tests` as requested for feedback by @brimtown .

[packages/gatsby-image/src/index.js - Line #19](https://github.com/gatsbyjs/gatsby/commit/6282c752fed09a63233becd9641298ac2a33aeef#diff-5941a73c5d794b120294a0429b6351c1R19):

```JS
// Find the source of an image to use as a key in the image cache.
// Use `fixed` or `fluid` if specified, or the first image in
// either `fixedImages` or `fluidImages`
const getImageSrc = props =>
  props.fluid
    ? props.fluid.src
    : props.fixed
    ? props.fixed.src
    : props.fluidImages
    ? props.fluidImages[0].src
    : props.fixedImages[0].src
```

Refactoring the ternary logic from the two locations into a function is good, but this nested ternary approach, not so much. 

Would it be simpler to read if you instead destructure the props and use the `||` operator for selection? Same outcome I believe:

```JS
const getImageData = ({fluid, fixed, fluidImages, fixedImages}) => (
  fluid || fixed || fluidImages[0] || fixedImages[0]
)

const src = getImageData(convertedProps).src
```

or

```JS
const getImageSrc = ({fluid, fixed, fluidImages, fixedImages}) => {
  const data = fluid || fixed || fluidImages[0] || fixedImages[0]
  return data.src
}

const src = getImageSrc(convertedProps)
```

[packages/gatsby-image/src/index.js - Line #155](https://github.com/gatsbyjs/gatsby/commit/6282c752fed09a63233becd9641298ac2a33aeef#diff-5941a73c5d794b120294a0429b6351c1R155):

```JS
  let initialSources = ``
  if (props.srcSetWebp) {
    initialSources += generateNoscriptSource(props, true)
  }
  if (props.media) {
    initialSources += generateNoscriptSource(props)
  }

   const variantSources =
    props.imageVariants.length > 0
      ? generateNoscriptSources(props.imageVariants)
      : ``

   return `<picture>${initialSources}${variantSources}<img ${width}${height}${sizes}${srcSet}${src}${alt}${title}${crossOrigin}style="position:absolute;top:0;left:0;opacity:1;width:100%;height:100%;object-fit:cover;object-position:center"/></picture>`
```

Any reason for splitting `sources` into `initialSources` and `variantSources` here? My concerns earlier for keeping a distinction between this features additions were more related to what you were doing with your `generateSources()` methods. It'd be fine to be consistent here with another if conditional that appends to the same `sources` string. The feature would still have a clear separation.

[packages/gatsby-image/src/index.js - Line #362](https://github.com/gatsbyjs/gatsby/commit/6282c752fed09a63233becd9641298ac2a33aeef#diff-5941a73c5d794b120294a0429b6351c1R362):

```JS
// First image data is supplied to `image`, if an array the rest will go to `variants`
const [image, ...imageVariants] = fluidImages ? fluidImages : [fluid]
```

Slight mistake here as this was due to taking one of my suggestions. `variants` should be `imageVariants` in the comment. You may also want to add the comment for the same line in the `fixed` logic.

---

Other than that, all change requests are resolved. These concerns are still valid:

- `fixed` and `fluid` image conditional blocks both still have [properties that don't adapt accordingly](
https://github.com/gatsbyjs/gatsby/pull/13395/files#r277528760)
- Any regression from this PR that would [affect Safari again](https://github.com/gatsbyjs/gatsby/pull/13395/files#r277539433)?

There is another issue regarding the cache key, but as it also affects the current `gatsby-image` code, it's not a blocker for accepting this PR, I'll raise that as a separate issue.

packages/gatsby-image/src/index.js
```diff
@@ -16,26 +16,32 @@ const convertProps = props => {
   return convertedProps
 }
 
+// Find the source of an image to use as a key in the image cache.
+// Use `fixed` or `fluid` if specified, or the first image in
+// either `fixedImages` or `fluidImages`
+const getImageSrc = props =>
```
brimtown: Not a massive fan of this. This might be cleaner if we overloaded `fluid` and `fixed` to take an Array, instead of introducing new props.

packages/gatsby-image/src/index.js
```diff
@@ -100,7 +152,20 @@ const noscriptImg = props => {
     ? `crossorigin="${props.crossOrigin}" `
     : ``
 
-  return `<picture>${srcSetWebp}<img ${width}${height}${sizes}${srcSet}${src}${alt}${title}${crossOrigin}style="position:absolute;top:0;left:0;opacity:1;width:100%;height:100%;object-fit:cover;object-position:center"/></picture>`
+  let initialSources = ``
```
brimtown: By splitting the logic between "generate source for first image" and "generate source for other images", this section starts to feel really funky. The logic is really the same for all of the sources (generate a source if there is a `media` key, and generate an additional source if there is a `srcSetWebp` key).

brimtown:
Sorry @polarathene , I had missed that you had reviewed this before I integrated those changes into this branch! Yes, those sections you called out were the places I felt like had regressed in the latest part. Will update those further 😄.

For the other two remaining issues:

#### Check for Safari Regressions
I imagine the Safari validation will be fairly straightforward, and I will do that after this latest round of updates.

#### Check how properties adapt in different circumstances
I've been wondering how best to validate some of this. I had planned to make a sort of Gatsby Image Playground that tries out several of these options (both `fixed` and `fluid`, with and without `critical`, with the same and different aspect ratios / heights & widths) and using that as a test bed for some of the behavior here.

I think I need to explore some of those combinations a bit more, and figure out which places we may need to handle, which places we'd want to document, and if there are any combinations that wouldn't be supported. At least for the `critical` case, we wouldn't even know which image was currently selected (since the browser is determining that itself), so there may be some combinations that wouldn't work exactly the same way as in the non-art-directed version.

KyleAMathews:
@brimtown would love to pair with you sometime on this! This is a really exciting PR!

KyleAMathews:
@brimtown an example site showing off all the different combinations would be very helpful as well

polarathene:
@brimtown 

> before I integrated those changes into this branch!

Just to confirm, there were no additional changes since? (You had mentioned you were going to add some more commits to the `changes` branch you linked me to on Discord before merging back to this PR)

> and figure out which places we may need to handle, which places we'd want to document, and if there are any combinations that wouldn't be supported.

It seems my last review link on this didn't properly link to the review like the [prior link](https://github.com/gatsbyjs/gatsby/pull/13395#discussion_r277528760), not sure if that had something to do with the newer commits or not. 

The linked review comment points out these properties tied to the first variant/element as `image`:

`fixed` section
- `divStyle` and `bgColor` both reference `image` width and height 
- `Img`(width/height attributes on the img element doesn't appear to matter?)

`fluid` section
- `image.aspectRatio`  used in the style for a `Tag` element
  ```JS
  paddingBottom: `${100 / image.aspectRatio}%`,
  ```

Those would render incorrectly when using variants with differing width/height such as your example of a tall vertical asset for a portrait orientation / mobile view, vs the wide horizontal asset for desktop / landscape orientation.

For a `fixed` image, the top-level wrapping element will use `divStyle` and if a solid colour placeholder is used, `bgColor` will also be affected. For `fluid` image, the aspect ratio would be invalid. Both cases should result in rendering issues for variants showcased in examples like https://www.shopflamingo.com/ (doesn't showcase the issue, just a variant scenario which would when using this PR).

I'm happy to approve the PR despite that, so long as users are made aware in the docs with a linked issue to be addressed in future.

> At least for the critical case, we wouldn't even know which image was currently selected (since the browser is determining that itself)

We won't know what image is selected until the element begins loading anyway. Similar to the cache key issue, that's too late for these elements. The mentioned properties above tied to `image` will be the same with the current code regardless of the `critical` prop.

There is a hook that can be used when an image starts to load `onStartLoad`. It may be possible to correct their values at that point, or depending on `sizes` and `media` attributes for the image, a JS query to the viewport may be able to match the appropriate variant(I'm not sure if `sizes` is important regarding variants? The browser is also not predictable with the value it chooses from `srcSet`, eg it will prefer higher quality/resolution if available in cache).

packages/gatsby-image/src/index.js
```diff
@@ -291,8 +358,9 @@ class Image extends React.Component {
       className: placeholderClassName,
     }
 
-    if (fluid) {
-      const image = fluid
+    if (fluid || fluidImages) {
+      // First image data is supplied to `image`, if an array the rest will go to `variants`
```
polarathene: Change to:

```JS
// Image variants support. First element to `image`, remaining to `imageVariants`
```

Add comment for the related `fixed` equivalent too.

KyleAMathews:
Generally it's best to keep doc / example site changes with code — both so people looking at the PR in the future can get context for the change in one look and also so that reviewers have a fighting chance to understanding what's going on :-D

--
Kyle Mathews

Blog: http://bricolage.io

Sent via Superhuman ( https://sprh.mn/?vip=mathews.kyle@gmail.com )

packages/gatsby-image/index.d.ts
```diff
@@ -20,13 +21,16 @@ interface FluidObject {
   tracedSVG?: string
   srcWebp?: string
   srcSetWebp?: string
+  media?: string
 }
 
 interface GatsbyImageProps {
   resolutions?: FixedObject
   sizes?: FluidObject
   fixed?: FixedObject
```
wardpeet: how do we feel about moving to the following:
```
sizes?: FluidObject | FluidObject[]
fixed?: FixedObject | FixedObject[]
```

packages/gatsby-image/src/index.js
```diff
@@ -69,6 +75,55 @@ function getIO() {
...
+          sizes={sizes}
+        />
+      )}
+      <source media={media} srcSet={srcSet} sizes={sizes} />
+    </React.Fragment>
+  ))
+}
+
+function generateTracedSVGSources(imageVariants) {
+  return imageVariants.map(variant => (
```
wardpeet: You destructor above so let's destructure here too.
```suggestion
  return imageVariants.map(({ src, media, tracedSVG })=> (
```

packages/gatsby-image/src/index.js
```diff
@@ -69,6 +75,55 @@ function getIO() {
...
+    <source
+      key={variant.src}
+      media={variant.media}
+      srcSet={variant.tracedSVG}
+    />
+  ))
+}
+
+function generateBase64Sources(imageVariants) {
+  return imageVariants.map(variant => (
```
wardpeet: idem ditto about destructuring

packages/gatsby-image/index.d.ts
polarathene: Should `resolutions` and `sizes` support this? My understanding was they were legacy and discouraged.

Or were you suggesting to drop `fixedImages` & `fluidImages`?

brimtown: I talked to @KyleAMathews about this. I don't think either of us had an incredibly strong preference one way or another, but based on how this will be used (and the fact that it's already implemented as such), we felt good about keeping these as separate props.

**Usage example:**
```jsx
const sources = [
  {
    ...mobileImage.fluid,
    media: `(max-width: 767px)`,
  },
  {
    ...desktopImage.fluid,
    media: `(min-width: 768px)`,
  },
];

<Image
  alt={mobileImage.description}
  fluidImages={sources}
/>
```

packages/gatsby-image/src/index.js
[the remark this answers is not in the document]
brimtown: Updated to this, but also with undefined checks for the arrays.

brimtown:
@polarathene @wardpeet I believe I've addressed the outstanding comments! Need to take a closer look at the failing E2E test, might be related to changes this PR makes to the markup (or potentially https://github.com/gatsbyjs/gatsby/pull/14201 as @KyleAMathews pointed out).

Still TODO before this is mergeable:

- [ ] Verify no Safari regressions
- [ ] Add docs to `gatsby-image` README
- [ ] Add example to Using Gatsby Image site
- [ ] Open an issue about art direction with mismatched aspect ratios / widths and heights

polarathene:
LGTM

brimtown:
@polarathene @wardpeet Would love a review on the docs I've added, and the e2e test updates I made. It also looks like a test flaked out, but I don't think I have the ability to re-run it.

#### Using-Gatsby-Image site
I was going to add an "Art Direction" section to https://using-gatsby-image.gatsbyjs.org/, but it looks like it might be easier to accomplish that once this has gone out in a release, so that that app can depend on the newly published version of this package.

#### No Safari Regressions
@tbrannam and I verified no regressions in Safari for this feature. It seems like ordering the attributes on the `<source>` tags as:

1. `media` (when present)
2. `srcSet`
3. `src`

causes Safari to behave as it should. Feel free to verify this for yourselves in the demo site.

#### Demo Site
I've also put up a really simple [site where you can see](https://inspiring-gates-f6a434.netlify.com/) this feature in action. Repo is over at https://github.com/brimtown/gatsby-image-playground if that's helpful.

You can see that for `fluid` images with different aspect ratios, or `fixed` images with different dimensions, that it will choose the sizing of the first image passed in the array. Like we talked about, this feels mergeable even with this, but I'm happy to document this or open a new issue about it.

Barring any other updates from additional review, and where to open an issue / document the different dimensions behavior, this seems ready to go from my end.

packages/gatsby-image/README.md
```diff
@@ -331,31 +331,88 @@ You will need to add it in your graphql query as is shown in the following snipp
 }
 ```
 
+## Art-directing multiple images
+
+If you want to show different images at different breakpoints (also known as [art direction](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#Art_direction), you can define your own array of `fixed` or `fluid` images, along with a `media` key per image, and pass it to `gatsby-image`'s `fixedImages` or `fluidImages` props. The `media` key that is set on an image can be any valid CSS media query.
```
polarathene: > The `media` key that is set on an image can be any valid CSS media query.

The media attribute takes a [media condition](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture#The_media_attribute)(similar to a media query). Small difference that probably doesn't matter mentioning much, it's only with attributes like `sizes` where there's some limitations afaik.

Since the MDN linked docs don't really clarify the difference they're pointing out and I wasn't able to find much information elsewhere, changing `query` to `condition` doesn't need to happen, it'd possibly cause some confusion to users. 

Only mentioning it in case anyone else knows more.

e2e-tests/gatsby-image/cypress/integration/fixed.js
```diff
@@ -24,7 +24,7 @@ describe(`fixed`, () => {
 
   it(`applies 1x/2x`, () => {
     cy.getTestElement(fixedTestId)
-      .find(`picture > source`)
+      .find(`picture:nth-of-type(2) > source`)
```
polarathene: The numbers `1` and `2` in all these tests should probably use a variable to represent the target? Or we could extract the common part of the selector to a descriptive variable name which should suffice:

```JS
const placeholderImage = `picture:nth-of-type(2)`
const primaryImage = `picture:nth-of-type(2)`
// ...
.find(`${primaryImage} > source`)
.find(`${primaryImage} > img`)
// ...
.find(`${placeholderImage} > source`)
.find(`${placeholderImage} > img`)
```

polarathene:
> I've also put up a really simple site where you can see this feature in action.

Thanks for the demo site :) If similar is done for https://using-gatsby-image.gatsbyjs.org/ , it'd be good to see the media conditions and expected matching image width/height up front(the inclusion within the actual image was a nice touch!). 

I did get a bit confused initially due to the size of the image that was loaded, my viewport was around 800-900px width, inspection I first saw `max-width: 1000px` but had misread what it was doing(apart from the fact it was sizes attribute for img) but the `min-width: 768px` media condition was also matching the 2000px width image.

```HTML
<source media="only screen and (min-width: 768px)" srcset="/static/3d845a0608b867aa95f6027e7834cc0a/32ee9/2000x1000.jpg 500w,
/static/3d845a0608b867aa95f6027e7834cc0a/2f7e7/2000x1000.jpg 1000w,
/static/3d845a0608b867aa95f6027e7834cc0a/883ab/2000x1000.jpg 2000w" sizes="(max-width: 2000px) 100vw, 2000px">
```

Actual inspection of above was that the `2000x1000` wasn't representing width/height, it's just the usual output from sharp, whoops :confounded: Probably asking too much to have the rescaled images also labelled(on the image) for the demo :sweat_smile: 

> Like we talked about, this feels mergeable even with this, but I'm happy to document this or open a new issue about it.

Open an issue about it, I think I know a way to resolve it and will try to contribute a fix sometime in June when I have the time to spare. If I can resolve it and get it a fix merged we can probably avoid documenting the issue. 

Native lazy-load support was merged recently with docs and console log stating `critical` prop is deprecated and to use the `loading` attribute instead, but currently that's not the case, `loading` does not presently replace `critical` prop completely..

packages/gatsby-image/README.md
brimtown: Ah, great callout. From what I can find, it seems like an entire `@media ...` expression is referred to as a query, but it's the `...` that is a media condition (or conditions).

It also looks like `media-condition` is a name used in the [formal spec](https://drafts.csswg.org/mediaqueries-4/#evaluating) for media queries.

brimtown:
Will open the issue about different dimensions / aspect ratios against `gatsby-image` as soon as this is merged. @wardpeet would love a re-review at your convenience 😄

packages/gatsby-image/src/index.js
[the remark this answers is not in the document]
wardpeet: oh yeah I agree with @polarathene we need to fix this.

`if (imageVariants) { /* picture logic */ } else { /* old behaviour */ }`

packages/gatsby-image/src/index.js
```diff
@@ -313,8 +376,9 @@ class Image extends React.Component {
       className: placeholderClassName,
     }
 
-    if (fluid) {
-      const image = fluid
+    if (fluid || fluidImages) {
+      // Image variants support. First element to `image`, remaining to `imageVariants`
+      const [image, ...imageVariants] = fluidImages ? fluidImages : [fluid]
```
wardpeet: I wonder if we should do the following to reduce some of the code below.
```
const imageVariants = fluidImages ? fluidImages : [fluid]
const image = imageVariants[0]
```

packages/gatsby-image/src/index.js
```diff
@@ -354,12 +418,20 @@ class Image extends React.Component {
 
           {/* Show the blurry base64 image. */}
           {image.base64 && (
-            <Img src={image.base64} {...placeholderImageProps} />
+            <picture>
+              <source media={image.media} srcSet={image.base64} />
```
wardpeet: if we go for the above imageVariants / image variable strategy we could  omit this line as it's part of imageVariants anyway
```suggestion
{generateBase64Sources(imageVariants)}
```

packages/gatsby-image/src/index.js
```diff
@@ -354,12 +418,20 @@ class Image extends React.Component {
...
+              {imageVariants && generateBase64Sources(imageVariants)}
+              <Img src={image.base64} {...placeholderImageProps} />
+            </picture>
           )}
 
           {/* Show the traced SVG image. */}
           {image.tracedSVG && (
-            <Img src={image.tracedSVG} {...placeholderImageProps} />
+            <picture>
+              <source media={image.media} srcSet={image.tracedSVG} />
```
wardpeet: same as above with base64

packages/gatsby-image/src/index.js
```diff
@@ -368,11 +440,17 @@ class Image extends React.Component {
               {image.srcSetWebp && (
```
wardpeet: same as above with base64

packages/gatsby-image/src/index.js
polarathene: Sorry, how is this reducing the code? You've just split the single line into two lines, this does exactly the same thing:

```JS
const [image, ...imageVariants] = fluidImages ? fluidImages : [fluid]
```

It either takes an array of `fluidImages` or puts the single `fluid` item into an array. The first element of the array is destructured into the const `image`, and the rest into the const `imageVariants`. The comment directly above it should make that clear?

packages/gatsby-image/src/index.js
```diff
@@ -101,6 +107,53 @@ function getIO() {
   return io
 }
 
+function generateImageSources(imageVariants) {
+  return imageVariants.map(({ src, srcSet, srcSetWebp, media, sizes }) => (
+    <React.Fragment key={src}>
```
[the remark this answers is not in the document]
polarathene: I definitely am not a fan of the 2nd suggestion(single loop), and the 1st one with duplicate code would be better handled with a single loop calling the function, I had to read through both to double check what the conditional was for(was something different), then re-read the intention of your reasoning for it(grouping media and non-media).

Do it this way:

```JSX
// Return an array ordered by elements having a media prop, does not use native sort, as a stable sort is not guaranteed
function groupByMedia(imageVariants) {
  const withMedia = []
  const without = []
  imageVariants.forEach((variant) => (variant.media ? withMedia : without).push(variant))

  return  [...withMedia, ...without]
}

function generateImageSources(imageVariants) {
  return groupByMedia(imageVariants).map(({ src, srcSet, srcSetWebp, media, sizes }) => (
    <React.Fragment key={src}>
      {srcSetWebp && (
        <source
          type="image/webp"
          media={media}
          srcSet={srcSetWebp}
          sizes={sizes}
        />
      )}
      <source media={media} srcSet={srcSet} sizes={sizes} />
    </React.Fragment>
  ))
}
```

This way, all other methods that should also implement it, can just use the same method call to get the desired order to operate on.

packages/gatsby-image/index.d.ts
```diff
@@ -20,13 +21,16 @@ export interface FluidObject {
...
```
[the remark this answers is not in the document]
polarathene: Didn't you already bring this up in your earlier review [here](https://github.com/gatsbyjs/gatsby/pull/13395#discussion_r285478901)?

Assuming it is about the same thing, using the current fixed/fluid props to accomodate array inputs too, that was answered on your original review about it. And discussed earlier during implementation.

packages/gatsby-image/src/index.js
```diff
@@ -40,26 +40,32 @@ const convertProps = props => {
   return convertedProps
 }
 
+// Find the source of an image to use as a key in the image cache.
+// Use `fixed` or `fluid` if specified, or the first image in
+// either `fixedImages` or `fluidImages`
+const getImageSrc = ({ fluid, fixed, fluidImages, fixedImages }) => {
```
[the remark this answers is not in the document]
polarathene: This was a function I contributed during review. It does return the image source url, so it's not inaccurate. It just happens to be used only for setting a cache key. Granted the comment explains how it's being used only for cache key purposes, so you could adjust the name if that helps?

It's used at line 60 here:

```JSX
  // Find src
  const src = getImageSrc(convertedProps)
  return imageCache[src] || false
```

Where it's clear that it's retrieving the image src value, but then using it as a key in the `imageCache` object.

packages/gatsby-image/src/index.js
```diff
@@ -101,6 +107,53 @@ function getIO() {
...
+        />
+      )}
+      <source media={media} srcSet={srcSet} sizes={sizes} />
+    </React.Fragment>
+  ))
+}
+
+function generateTracedSVGSources(imageVariants) {
+  return imageVariants.map(({ src, media, tracedSVG }) => (
+    <source key={src} media={media} srcSet={tracedSVG} />
```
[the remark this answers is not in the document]
polarathene: This issue is happening because of the spaces, in `srcset` you provide a list of urls as sources. If no descriptor(w or x unit) is provided, it defaults to 1x or 100w. List is separated by comma, and width descriptors when used by a space after the url/uri before the comma.

So base64 uri is working as there is no spaces, browser understands that fine. With the url encoded inlined xml, it hasn't encoded the spaces to %20, so it gets confused as it's not indicating a valid descriptor. This will work:

```
srcset="data:image/svg+xml,%3Csvg%20xmlns%3D%27http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%27%20width%3D%27120%27%20height%3D%27180%27%3E%3Cpath%20d%3D%27M51%2042l-19%204-2-1c2-2-1-2-3%200-3%202-4%204-1%203l-1%203-3%203c-1%200-3%205-3%208l-1%204c-3%201-5%205-4%209%200%204%201%205%204%205%202%201%201%202-2%201-1%200-2%201-2%203l-4%206c-5%207-3%2016%206%2021l7%202c1-1%205%202%206%204%200%204%203%209%204%2010l3%207c1%206%204%2010%208%2011%201%201%204%202%204%204%203%203%206%201%204-3-1-2-1-5%201-5l3%206c0%203%201%205%202%205%209%202%2014%200%2021-6%206-5%2015-17%2018-22l3-4%203-4%203-4c2%200%205-6%206-13%201-8%200-10-5-17-2-3-3-5%200-5%201%201%201%201%201-1v-2c2-1-2-5-7-8-3-1-3-1-2%202v4l-1%203c1%201%200%202-1%202l-1-2c1-3-1-10-2-10l-1-1-2-5c-2-2-3-5-3-8%200-4-1-6-4-6l-4-1-10-3a514%20514%200%200%201-19%201%27%20fill%3D%27%23d3d3d3%27%20fill-rule%3D%27evenodd%27%2F%3E%3C%2Fsvg%3E"
```

But base64 would be less in length, so you probably want to use that when assigning the value to srcset(it's fine for regular src on img though).

No need for SVG guru :)

packages/gatsby-image/src/index.js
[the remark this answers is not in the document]
polarathene: Github has some weird way of showing updates, it looked like you gave a new comment, rather than replying to an old review comment :confused: 

I can't say how many projects would be impacted by this, but another option is just to bump the major version of the package? (although due to concerns with the last PR I contributed and how this package version management is handled, that's not a great idea)

packages/gatsby-image/src/index.js
polarathene: ? Your suggestion is to remove the source element and add another `generateBase64Sources()` without the `imageVariants` logical guard?

This is the current PR version(since the preview is a bit annoying for context):

```JSX
          {/* Show the blurry base64 image. */}
          {image.base64 && (
            <picture>
              <source media={image.media} srcSet={image.base64} />
              {imageVariants && generateBase64Sources(imageVariants)}
              <Img src={image.base64} {...placeholderImageProps} />
            </picture>
          )}
```

In my earlier review I requested it be separate from variants handling. For more details, see the [related review comment](https://github.com/gatsbyjs/gatsby/pull/13395#discussion_r277518050), also [kind of related](https://github.com/gatsbyjs/gatsby/pull/13395#discussion_r277495092)

Originally there was conditional logic for the 1st element/image in variants, then I made a request to handle it differently. @brimtown mentioned that both sources may be needed and that the order matters.

It's been a while, so I can't recall if there were any other concerns/reasons or the code has evolved since then. It does seem with the current state of code that we can offload all source generation to the methods without any conditional handling?

That aside, if there are no image variants, should an additional source being generated regardless?(I think that was to cater specifically to media prop, if so it could use a logical guard if not moved into generate method)

packages/gatsby-image/src/index.js
polarathene: Oh nvm, you wanted to keep the original `fluidImages` array but mapped to `imageVariants`.

It was like that originally, but was specially handling the first element(see the [outdated code here](https://github.com/gatsbyjs/gatsby/blob/46dc7318d291b299bd79b3d94bf3542e5ab5612b/packages/gatsby-image/src/index.js#L79)), this may no longer be an issue as pointed in a [response to one of your other review comments](https://github.com/gatsbyjs/gatsby/pull/13395#discussion_r291798702).

packages/gatsby-image/src/index.js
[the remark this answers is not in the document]
wardpeet: Sweet @polarathene. You're now part of the svg guru team 😄 lets just encode spaces with %20 everytime. It's less breaking and should be mostly smaller in size.

https://codepen.io/tigt/post/optimizing-svgs-in-data-uris is a good read to see that svg is smaller even with encoding spaces.

packages/gatsby-image/src/index.js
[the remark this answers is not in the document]
wardpeet: This seams much cleaner! Seems like a good fit. If we can presort imageVariants  that would be great like you suggested.

polarathene:
Consolidating the review by @wardpeet + my input.

## Change Requests

### SVG inlining bug

@wardpeet [noticed a problem with tracedSVG](https://github.com/gatsbyjs/gatsby/pull/13395#discussion_r291776267), I've identified it and have a solution below.

I've [raised an issue](https://github.com/tigt/mini-svg-data-uri/issues/9) about the SVG inline issue. If they don't feel their library should accommodate for that or in the meantime, I think all we need to do is convert literal spaces to `%20`, else serve up a larger base64 string.

Just add in `encodeSpaces` method as below to `gatsby-plugin-sharp` [here](https://github.com/gatsbyjs/gatsby/blob/44aad18a396ffc41a29a6bc76cb6e3ae032f9716/packages/gatsby-plugin-sharp/src/trace-svg.js#L108-L123):

```JS
 // ...

 // `srcset` attribute rejects URIs with literal spaces
 const encodeSpaces = (str) => str.replace(/ /gi, `%20`)
  
 return trace(tmpFilePath, optionsSVG) 
   .then(optimize) 
   .then(svgToMiniDataURI) 
   .then(encodeSpaces)
```

### Ensuring mixed media

@wardpeet expressed [concern on the off-chance you had a mix of variants supplied where only some had a media prop](https://github.com/gatsbyjs/gatsby/pull/13395#discussion_r291768635).

```JS
// Return an array ordered by elements having a media prop, does not use native sort, as a stable sort is not guaranteed by all browsers/versions
function groupByMedia(imageVariants) {
  const withMedia = []
  const without = []
  imageVariants.forEach((variant) => (variant.media ? withMedia : without).push(variant))

  return  [...withMedia, ...without]
}
```

### Include first image in `imageVariants` as well

Then it was also suggested to [revert back to keeping the 1st image supplied in the `imageVariants`](https://github.com/gatsbyjs/gatsby/pull/13395#discussion_r291779417) as well. This makes the name less appropriate, but it doesn't appear that we need special handling/separation anymore?

Modifying @wardpeet 's suggestion a little to sort the array by media prop(leaving the generate methods alone):

```JS
const imageVariants = fluidImages ? groupByMedia(fluidImages) : [fluid]
const image = imageVariants[0]
```

### Avoid breaking change for placeholders

@wardpeet responded to the old review comment about introducing a potential breakage, [advising to avoid it via conditional](https://github.com/gatsbyjs/gatsby/pull/13395#discussion_r291779282). We could do this a little differently:

```JSX
// Avoid breakage during v2 prior to wrapping in a picture element
const Placeholder ({src, spreadProps, imageVariants, generateSources}) => {
  const baseImage = <Img src={src} {...spreadProps} />

  return imageVariants.length > 1 && (
    <picture>
      {generateSources(imageVariants)}
      {baseImage}
    </picture>
  ) : baseImage
}

// ...

{/* Show the blurry base64 image. */}
{image.base64 && (
  <Placeholder 
    src={image.base64}
    spreadProps={placeholderImageProps}
    imageVariants={imageVariants}
    generateSources={generateBase64Sources}
  />
)}

{/* Show the traced SVG image. */}
{image.tracedSVG && (
  <Placeholder 
    src={image.tracedSVG}
    spreadProps={placeholderImageProps}
    imageVariants={imageVariants}
    generateSources={generateTracedSVGSources}
  />
)}
```

I've omitted the extra `source`, these was [decided earlier](https://github.com/gatsbyjs/gatsby/pull/13395#discussion_r277748597) to be kept on the basis that a user might add a media attribute without supplying variants for art direction for some unknown reason. 

The single image media attribute edge-case described, is no longer supported. If a user has a need for it, they could raise an issue requesting support, which is fine, we don't know of any legit case for supporting it?

**Tests will probably need an update due to having two placeholder variants(with and without picture)?**

### noscript gets simplified from all this

The noscript support will need to be altered too(previously, we were checking length without checking `imageVariants` existed, missed that bug):

```JS
// ...
let sources = ``
if (props.srcSetWebp) {
  sources += generateNoscriptSource(props, true)
}
if (props.media) {
  sources += generateNoscriptSource(props)
}
if (props.imageVariants.length > 0) {
  sources += generateNoscriptSources(props.imageVariants)
}
```

Simplified to:

```JS
let sources = generateNoscriptSources(props.imageVariants)
```

### Have all sources generated via the method

As evident above, the code has progressed to a point that it should be ok to [get all sources from the relevant methods](https://github.com/gatsbyjs/gatsby/pull/13395#discussion_r291779511).

That just leaves the primary image to drop it's source elements and `imageVariants` guard to just be the generate method above the `Img` tag/component.

### Change a function name

There was a request to [change the name of a function](https://github.com/gatsbyjs/gatsby/pull/13395#discussion_r291758964), it's not used for anything else at present, and it only retrieves the 1st element/image src, so probably makes sense to make the name less generic.

---

## Considerations

@wardpeet questioned [why do we have separate props for this feature vs using existing ones](https://github.com/gatsbyjs/gatsby/pull/13395#discussion_r291757631). It's been discussed a few times(in the original issue and PR), there isn't any strong reasons to prefer either approach, other than the code might be more easier to read/understand when the two types are more distinct, and users may deal with extra confusion(unless you should pass an array of inputs in regardless).

General consensus was to keep it as is.

packages/gatsby-image/src/index.js
wardpeet: I find
```html
<source media={image.media} srcSet={image.base64} />
{imageVariants && generateBase64Sources(imageVariants)} 
```
a bit weird to read because `<source media={image.media} srcSet={image.base64} />` is also a source tag. We can put the guard inside generateBase64Sources and just return null.

packages/gatsby-image/src/index.js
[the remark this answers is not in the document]
polarathene: Oh, you started responding to my earlier comments. I just deleted two older ones that have been better addressed in the larger consolidation post I made, mostly for my benefit, the deleted comments no longer contributed anything, I write enough for others to read as is :sweat_smile:

Thanks for that link :) I had seen a small table comparison by the same author of the `mini-svg-data-uri` libraries npm page. Hadn't gotten around to updating my comment to no longer suggest base64 :sweat_smile:

polarathene:
The issue with SVG URI lib is being addressed upstream over the weekend. So if you wait until that's released, we'll have a fix rather than the mentioned workaround.

wardpeet:
I would like to get "Avoid breaking change for placeholders" and "Ensuring mixed media" fixed. If svg-uri-lib hasn't published a new version yet we can do a quickfix ourselves.

polarathene:
> does this make sense?

What do you mean by fixed? I know that you're listing the sub-headings in my summary post, but I thought it made the change request rather clear?

If I didn't, all @brimtown needs to do is apply [these changes](https://www.diffchecker.com/9vKmPQvx)(diff), [pastebin here](https://pastebin.com/9cGFfBJ6) for easier copying.

That should work, just needs the tests updated to go with it.

> If svg-uri-lib hasn't published a new version yet we can do a quickfix ourselves.

The [linked issue](https://github.com/tigt/mini-svg-data-uri/issues/9) has the author welcoming a PR, else they'll take care of it in the weekend. 

You could apply the quickfix I gave above for it in the meantime, or just raise a bug issue here and close it next week with another PR once `mini-svg-data-uri` releases the update.

wardpeet:
I mean your solutions implemented before merging ^^. I can do it

polarathene:
@wardpeet oh? I thought we have to wait on @brimtown to update it. I'm not sure how I could contribute to the PR more directly.

Let's get the tests updated and merge this? :grinning: 

I can raise the issues about svg srcset and render not supporting variants(outside of picture elements) properly atm.

brimtown:
Applying the linked diff and updating the tests now. I'll hold off on the SVG encoding.

wardpeet:
lets go for the fix ourselves @brimtown if that's possible so we don't have to wait upstream

@polarathene maybe because it's because I'm an admin but I can edit PRs when creators allow.

packages/gatsby-image/src/index.js
```diff
@@ -107,44 +107,60 @@ function getIO() {
...
+        )}
+        <source media={media} srcSet={srcSet} sizes={sizes} />
+      </React.Fragment>
+    )
+  )
 }
 
 function generateTracedSVGSources(imageVariants) {
-  return imageVariants.map(({ src, media, tracedSVG }) => (
+  return groupByMedia(imageVariants).map(({ src, media, tracedSVG }) => (
```
polarathene: This wasn't in my pastebin, the `groupByMedia` call should only be run once in `render()` not in any of these generate methods.

packages/gatsby-image/src/index.js
```diff
@@ -525,38 +538,27 @@ class Image extends React.Component {
...
-                  srcSet={image.srcSetWebp}
-                  sizes={image.sizes}
-                />
-              )}
-              <source
-                media={image.media}
-                srcSet={image.srcSet}
-                sizes={image.sizes}
-              />
               {imageVariants && generateImageSources(imageVariants)}
```
polarathene: The pastebin has the logical guard removed. `imageVariants` always exists now(even if it contains only a single element).

brimtown:
Sorry @polarathene , I was confused by some earlier feedback. Will update this to make it literally the same as the Pastebin.

polarathene:
@brimtown No worries. I tried to help avoid the confusion from the individual discussions with that summary/consolidated post.

When I reviewed your commit, there was at least one thing I spotted that I know I missed(I hadn't tested that pastebin) but you got. `getImageSrcKey` was not updated for `activateCacheForImage`, my bad :grimacing:

wardpeet:
I'll do a proper test after changes and fix a few things if guards are missing.

Thanks @polarathene & @brimtown much appreciate the review & code changes :pray:

brimtown:
Updated on my end 👍

packages/gatsby-image/src/index.js
```diff
@@ -202,7 +202,12 @@ const noscriptImg = props => {
 // Earlier versions of gatsby-image during the 2.x cycle did not wrap
 // the `Img` component in a `picture` element. This maintains compatibility
 // until a breaking change can be introduced in the next major release
-const Placeholder = ({ src, spreadProps, imageVariants, generateSources }) => {
+const Placeholder = ({
+  src,
+  imageVariants,
+  generateSources,
+  ...spreadProps
```
polarathene: If you're going to do this, might as well stick to the usual conventional name of `...rest`? 

`spreadProps` was a prop name to communicate the intent to spread the props to a child. Which I believe is less ambiguous to the reader vs collecting any additional props(some which can be implicit like the children prop(not a concern here since this is all handled internally).

Does the use of two additional spread syntax (one here, and another per instance where the prop name is now removed) provide value in some way? Your commit message didn't really indicate why this change was made, I'm curious.

packages/gatsby-image/src/index.js
```diff
@@ -37,18 +37,17 @@ const convertProps = props => {
     convertedProps.loading = `eager`
   }
 
+  // convert fluid & fixed to arrays so we only have to work with arrays
+  convertedProps.fluid = [].concat(convertedProps.fluid).filter(Boolean)
```
polarathene: @brimtown did [something similar to this originally in `convertProps()`](https://github.com/gatsbyjs/gatsby/pull/13395/files/46dc7318d291b299bd79b3d94bf3542e5ab5612b#diff-5941a73c5d794b120294a0429b6351c1R16) and [I wasn't too fond of it](https://github.com/gatsbyjs/gatsby/pull/13395#discussion_r277495092). Especially now that the method indicates it's intended for handling deprecation transitions. Which I guess can apply here and encourage users to always wrap in an array(or have the related sharp plugin/transformer accommodate).

@brimtown 's version assigned the first element if an array were present, while the code has now changed to expect an array to work with regardless. I would say, an earlier suggestion I made would be more appropriate here(nitpicking), as it's less operations, and a little more clearer to read at a glance what the desired outcome is:

```JS
convertedProps.fluid = Array.isArray(fluid) ? fluid : [fluid]
convertedProps.fixed = Array.isArray(fixed) ? fixed : [fixed]
```

packages/gatsby-image/src/index.js
```diff
@@ -37,18 +37,17 @@ const convertProps = props => {
...
-// either `fixedImages` or `fluidImages`
-const getImageSrcKey = ({ fluid, fixed, fluidImages, fixedImages }) => {
-  const data =
-    fluid ||
-    fixed ||
-    (fluidImages && fluidImages[0]) ||
-    (fixedImages && fixedImages[0])
+// Use `the first image in either `fixed` or `fluid`
+const getImageSrcKey = ({ fluid, fixed }) => {
+  const data = (fluid.length && fluid[0]) || (fixed.length && fixed[0])
```
polarathene: Is the length check necessary? It'll likely fail iirc checking the `length` property if the prop itself is undefined? Or is this because of `convertProps()` that it's safe to always assume an array value present?

I would add that we have both `inImageCache()` and `activateCacheForImage()` both calling this method after first doing:

```JS
const convertedProps = convertProps(props)
```

We might as well just move that duplication into this method and drop the need to pass in props?

packages/gatsby-image/src/index.js
```diff
@@ -395,8 +392,8 @@ class Image extends React.Component {
       className: placeholderClassName,
     }
 
-    if (fluid || fluidImages) {
-      const imageVariants = fluidImages ? groupByMedia(fluidImages) : [fluid]
+    if (fluid.length) {
+      const imageVariants = groupByMedia(fluid)
```
polarathene: I see that you've moved this logic to `convertProps()` but left the `groupByMedia()` method here. My only concern about that is the render methods referencing `imageVariants` will now potentially have a different order of elements than anything referencing fluid/fixed by index directly. Presently, that's only `getImageSrcKey()`. 

I would like to properly address returning the correct key for a variant in future so that `image` will not have the issues it presently will with this feature when a variant differs from properties that `image` is used for instead.

Should `groupByMedia()` be moved to `convertProps()` as well? If there's any concerns for additional overhead, the result could always be memoized.

polarathene: Like earlier, what is the `length` here for? If you want to check this is an array, use `Array.isArray(fluid)`, although it should always be given `convertProps()`?

packages/gatsby-image/src/index.js
```diff
@@ -131,6 +131,14 @@ function groupByMedia(imageVariants) {
     (variant.media ? withMedia : without).push(variant)
   )
 
+  if (without.length > 1 && process.env.NODE_ENV !== `production`) {
+    console.warn(
+      `We've found ${
+        without.length
+      } sources without a media property. You should only have 1.`
```
polarathene: "You should only have 1", isn't clear. Only 1 property with or without media property? It confused me for a little bit at first by what was meant.

Is referring to them as `sources` clear enough? I see on the docs for the PR there's a `sources` variable, so I guess so. `image variants` made a bit more sense to me, since there are multiple sources in the picture element to support additional image formats, and multiple images of different sizes for srcset.

What should a user do with this information? Do they know which source can exclude a media prop/key? Would linking to relevant doc page/header help?

packages/gatsby-image/src/__tests__/index.js
```diff
@@ -167,9 +167,39 @@ describe(`<Image />`, () => {
...
   it(`should have the the "critical" prop set "loading='eager'"`, () => {
+    jest.spyOn(global.console, `log`)
+
     const props = { critical: true }
     const imageTag = setup(false, props).querySelector(`picture img`)
     expect(imageTag.getAttribute(`loading`)).toEqual(`eager`)
+    expect(console.log).toBeCalled()
+  })
+
+  it(`should warn if multiple sources with no media are used.`, () => {
```
polarathene: Would this be clearer? The tests are more related to development of `gatsby-image`, where `source` has varied meaning especially regarding the `picture` element, whereas this warning is more to do with the fluid/fixed input array provided to the component.

"should warn if image variants provided are missing media keys."

packages/gatsby-image/src/index.js
polarathene: Disregard this suggestion. As it is no longer in scope of just fluid or fixed, it can return a `[undefined]` value that'd mess things up.

polarathene:
Review done. Some questions for @wardpeet regarding some commits he made, along with some suggestions.

What is with the [Danger failure](https://github.com/gatsbyjs/gatsby/pull/13395/checks?check_run_id=149762914)? Was it trying to merge on green?

> "triggeredByUsername": "wardpeet"

packages/gatsby-image/src/index.js
wardpeet: oh I just feel this is better XD I didn't see much point of not using {...placeholderImageProps} but i'll revert as i'm probably the only one thinking that.

wardpeet:
No clue about dangerjs. It happend when I pushed. We can just ignore it

packages/gatsby-image/src/index.js
wardpeet: I'll make it more verbose with wrapping it. I don't think `[].concat` is super hard to understand
```
// convert fluid & fixed to arrays so we only have to work with arrays
if (convertedProps.fluid) {
  convertedProps.fluid = [].concat(convertedProps.fluid)
}
if (convertedProps.fixed) {
  convertedProps.fixed = [].concat(convertedProps.fixed)
}
```

packages/gatsby-image/src/index.js
wardpeet: Moved groupByMedia to convertedProps, that's a good suggestion. Even if it might slow down a bit. It's clearer in code. I'll refactor it a little in another PR so negate this effect.

packages/gatsby-image/src/index.js
wardpeet: Thanks!

packages/gatsby-image/src/__tests__/index.js
wardpeet: Renamed it as you mentioned

wardpeet:
I think I have resolved all review comments.

packages/gatsby-image/src/index.js
```diff
@@ -406,7 +410,7 @@ class Image extends React.Component {
     }
 
     if (fluid.length) {
```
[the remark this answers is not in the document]
wardpeet: yeah, tests are failing because of that 😛

packages/gatsby-image/src/index.js
```diff
@@ -38,17 +38,26 @@ const convertProps = props => {
...
+ * Use `the first image in either `fixed` or `fluid`
+ * @param {{fluid: {src: string}[], fixed: {src: string}[]}} args
+ * @return {string}
+ */
 const getImageSrcKey = ({ fluid, fixed }) => {
-  const data = (fluid.length && fluid[0]) || (fixed.length && fixed[0])
-  return data.src
+  const data = (fluid && fluid[0]) || (fixed && fixed[0])
+
+  return data && data.src
```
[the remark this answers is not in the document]
wardpeet: oh yeah lets fail hard 👍

antoinegrelard:
Hi @brimtown , thanks for this !

I just upgraded the package to get access to this and I find myself stuck with a rather annoying issue.

From my understanding, Art direction means we can display different images at different breakpoints, which means we can have different aspect ratios through viewport sizes. The issue I think is the following :

Given this sources array :

```
const fakeImage = {
  aspectRatio: 0.75,
  src: "https://via.placeholder.com/1440x1920",
  srcSet:
    "https://via.placeholder.com/300x400 300w, https://via.placeholder.com/600x800 600w, https://via.placeholder.com/900x1200 900w, https://via.placeholder.com/1440x1920 1440w",
  sizes: "(max-width: 1440px) 100vw, 1440px",
};

const fakeImageBig = {
  aspectRatio: 1.7777777778,
  src: "https://via.placeholder.com/1600x900",
  srcSet:
    "https://via.placeholder.com/300x169 300w, https://via.placeholder.com/600x338 600w, https://via.placeholder.com/1200x675 1200w, https://via.placeholder.com/1600x900 1600w",
  sizes: "(max-width: 1600px) 100vw, 1600px",
};

const sources = [
  fakeImage,
  {
    ...fakeImageBig,
    media: `(min-width: 768px)`,
  },
];
```

I would expect the aspect ratio to be at 0.75 below 768px, and be updated to 1.77778 over, but currently it's just taking the image.aspectRatio, which is fluid[0], therefore not updating the aspect ratio.

Am I just not doing the right thing ?

Thanks!

brimtown:
Hi @antoinegrelard ! Yes, this is a current limitation with the code as is. Since `gatsby-image` uses `style` tags that don't support media queries, we de-scoped it to get this merged.

Regardless, I believe @polarathene had some ideas how to handle from Gatsby itself. I will open the issue on this and tag you in it, and it should be fixable 👍 .

In the meantime, we've worked around this in our own app code with a snippet like this:

```js
// These styles fix a problem in `gatsby-image` where using
// art direction with images that have different aspect ratios
// defaults to using the aspect ratio of the first image passed
// to `fluidSources`. This is hardcoded to work for two images,
// but could be generalized to work for more. Ideally this
// will be fixed upstream in the future.
const ResponsiveImage = styled(GatsbyImage)`
  ${({ sources }) => `
    & > div:first-child {
      padding-bottom: ${100 / sources[0].aspectRatio}% !important;

      ${mediaQueries.DESKTOP
        padding-bottom: ${100 / sources[1].aspectRatio}% !important;
      `}
    }
  `}
`;
```

If you know how many variations you'll need, hardcoding it with something like this works fairly well. You could also map over `sources` and generate these more programmatically, which Gatsby will need to do to solve this directly.

Glad you're using the feature! I think once this gets fixed, we can share it more broadly in a blog post and improve some of the documentation here.

PR #13397, opened 2019-04-16
chore(showcase): add Gold Medal Flour to site showcase

thadroe:

### DIFF
--- a/docs/sites.yml
+++ b/docs/sites.yml
@@ -5366,3 +5366,13 @@
   categories:
     - Open Source
     - Web Development
+- title: Gold Medal Flour
+  main_url: https://www.goldmedalflour.com
+  url: https://www.goldmedalflour.com
+  description: >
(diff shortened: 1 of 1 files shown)


docs/sites.yml
```diff
@@ -5366,3 +5366,13 @@
   categories:
     - Open Source
     - Web Development
+- title: Gold Medal Flour
+  main_url: https://www.goldmedalflour.com
+  url: https://www.goldmedalflour.com
+  description: >
+    Gold Medal Four is a brand of flour products owned by General Mills. The new site was built using Gatsby v2 with data sources from Wordpress and an internal recipe API, and features multifaceted recipe filtering and a modified version of Gatsby Image to support art direction images.
```
DSchau: Hey cool -- we're actually working on this right now. Any insight into that would be super valuable!

https://github.com/gatsbyjs/gatsby/pull/13395

thadroe: Oh sweet! Glad to see it being worked on! I'll check out that PR and see if I can offer up anything. What I did for this might be a little on the crude side, but essentially what I ended up doing is copying modifying the default Gatsby Image component and used styled components, also adding an option for a custom breakpoint prop. I put it up here just now if you want to have a look.

https://github.com/thadroe/art-direction-images-for-gatsby

PR #13418, opened 2019-04-17
fix: usage of pluginOptions leads to TypeError: Cannot read property 'imagesdir' of undefined

lask79:
In the last refactoring a small bug was introduced that leads to an undefined error when the user has no attributes defined in the pluginOptions.

clonedPluginOptions should be used instead.


pieh:
published `gatsby-transformer-asciidoc@1.0.5`, thanks again!

PR #13424, opened 2019-04-17
feat(gatsby-plugin-google-tagmanager): Allow to place the GTM script …

NoriSte:
…at the end of the <body> tag

## Description

I added one more option to the `gatsby-plugin-google-tagmanager` plugin to allow adding the GTM script in the head (by default, as it is now) or append it to the body (to make it non-blocking).

### DIFF
--- a/packages/gatsby-plugin-google-tagmanager/README.md
+++ b/packages/gatsby-plugin-google-tagmanager/README.md
@@ -24,6 +24,11 @@ plugins: [
       gtmAuth: "YOUR_GOOGLE_TAGMANAGER_ENVIROMENT_AUTH_STRING",
       gtmPreview: "YOUR_GOOGLE_TAGMANAGER_ENVIROMENT_PREVIEW_NAME",
       dataLayerName: "YOUR_DATA_LAYER_NAME",
+
+      // Whether to put the GTM script into the <head> (as suggested by Google)
+      // or append it to the <body> (making it non-blocking).
+      // If you activate Google Optimize through Google Tag Manager leave it untouched.
+      placeInHead: true,
     },
   },
 ]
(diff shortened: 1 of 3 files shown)


packages/gatsby-plugin-google-tagmanager/README.md
```diff
@@ -24,6 +24,11 @@ plugins: [
       gtmAuth: "YOUR_GOOGLE_TAGMANAGER_ENVIROMENT_AUTH_STRING",
       gtmPreview: "YOUR_GOOGLE_TAGMANAGER_ENVIROMENT_PREVIEW_NAME",
       dataLayerName: "YOUR_DATA_LAYER_NAME",
+
+      // Whether to put the GTM script into the <head> (as suggested by Google)
+      // or append it to the <body> (making it non-blocking).
+      // If you activate Google Optimize through Google Tag Manager leave it untouched.
+      placeInHead: true,
```
sidharthachatterjee: Since the current default is to place it in HEAD, I would flip this over and add an option to explicitly place this in body (`addTagInBody` or whatever) so that we don't change default behaviour

NoriSte: It makes sense! I've just pushed the change 😊
Thank you @sidharthachatterjee

NoriSte:
Is the NPM release of the plugin automatic? Do you know when it could happen? To know when I can discard my changes and re-download the official one. 😊

sidharthachatterjee:
@NoriSte Releases are manual at the moment and requires one of us from the core team to publish. 

I've reverted this in https://github.com/gatsbyjs/gatsby/pull/13424 because @wardpeet correctly pointed out that since the script is always added `async`, this doesn't really have any impact and hence it's best that we don't add another option. 

Nonetheless, thank you again for your contribution and sorry for the confusion!

NoriSte:
I need some more investigations because I got (obviously as it seems) wrong feedbacks from my first tests. I'll let you know as soon as I find out something more.
Anyway: don't worry and sorry!

Issue #13003, opened 2019-04-01
Roadmap to migrate away from request


DanielRuf:
## Summary
`request` is in maintenance mode now so we should check what we have to do to migrate away from it.

## Relevant information
https://github.com/request/request/issues/3142

KyleAMathews:
There's no specific rush to do this as `request` isn't going to stop working :-)

I did a quick look and we only use `request` once in our code base — https://github.com/gatsbyjs/gatsby/blob/a43288d15831926c6c92192e1c0f50878b6fd798/packages/gatsby/src/commands/develop.js#L193

If someone wants to PR moving to a different library, that seems reasonable.

DanielRuf:
> I did a quick look and we only use `request` once in our code base

I thought there were more places where we still use `request`.

> There's no specific rush to do this as `request` isn't going to stop working :-)

Sure but let's keep an eye on this.

> If someone wants to PR moving to a different library, that seems reasonable.

👍

sindresorhus:
If you're willing to give [`got`](https://github.com/sindresorhus/got) a try, I think you could simply replace `request(` with `got.stream(` in the referenced code.

See: https://github.com/sindresorhus/got/blob/master/migration-guides.md#more-about-streams

wardpeet:
sweet! we're open for PRs but this isn't really urgent.

ryandrew14:
Looked into it a little more, and I think @sindresorhus is right -- `got` seems like the best option. They have a great [comparison](https://www.npmjs.com/package/got#comparison) on their website of how it stacks up against other Node libraries for making requests. I will make a PR for this tonight.

pieh:
Just note on `got` - we still do support node 6 :( and `got@^9` seems to require `node: ">=8.6"`, so if we would switch we would need to use `got@<=8` (which works on `node: ">=4"`). 

The node 6 support is super problematic for us, but we can't drop it in the middle of `gatsby@2` (even if node 6 end of life is generally coming pretty soon - I think end of April 2019).

ryandrew14:
Closed by #13434.

sidharthachatterjee:
Published in `gatsby@2.3.25`

armchairdeity:
> If you're willing to give [`got`](https://github.com/sindresorhus/got) a try, I think you could simply replace `request(` with `got.stream(` in the referenced code.
> 
> See: https://github.com/sindresorhus/got/blob/master/migration-guides.md#more-about-streams

[...]

PR #13434, opened 2019-04-18
chore(gatsby): change HTTP request library from `request` to `got` for future support

ryandrew14:
## Description
This PR changes the HTTP request library used by `gatsby develop` to [`got`](https://www.npmjs.com/package/got#comparison) instead of [`request`](https://www.npmjs.com/package/request). This is based on #13003 which explains that `request` is going into maintenance mode. The PR makes as few changes as possible.

## Related Issues
#13003


ryandrew14:
Hey Kyle, not sure if I needed to tag you on this (just seeing that "core" is automatically added as a reviewer. Sorry about that!

For whoever reviews this, can I ask why this isn't using a Promise API in the first place and is using streams instead? Is this because Promises are a lot of overhead in Node? Thanks!

packages/gatsby/package.json
```diff
@@ -71,6 +71,7 @@
     "gatsby-react-router-scroll": "^2.0.7",
     "gatsby-telemetry": "^1.0.7",
     "glob": "^7.1.1",
+    "got": "^9.6.0",
```
pieh: We can't use `got@^9` because it is Node 8+ and we still support Node 6 (for a little while at least) :(

ryandrew14: Great, thanks for the heads-up! I made a quick commit to downgrade.

ryandrew14:
Thanks @sidharthachatterjee! I'm new to Node development -- is the reason we use streams over promises here just that promises are a lot of overhead in Node?

sidharthachatterjee:
@ryandrew14 Thank you for making the contribution!

> is the reason we use streams over promises here just that promises are a lot of overhead in Node?

I wouldn't worry _too_ much about overhead. In this instance, it looks like we're using streams because piping from `req` and to `resp` seems intuitive and elegant!

https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/commands/develop.js#L213

sidharthachatterjee:
Published in `gatsby@2.3.25`

PR #13435, opened 2019-04-18
chore(showcase): add sals-pals.net to showcase 

stollerm:
## Description
Submitting sals-pals.net for sites.yml. Built with gatsby v2 🚀


docs/sites.yml
```diff
@@ -5361,6 +5361,15 @@
...
   built_by_url: "https://jonzfisher.com"
+- title: Sal's Pals
+  main_url: https://www.sals-pals.net
+  url: https://www.sals-pals.net
+  featured: false
+  description: >
+    Sal's Pals is a professional dog walking and pet sitting service based in Westfield, NJ. New site built with gatsby v2.
+  categories:
+    - Dog walking
+    - Pet sitting
```
pieh: I think those categories (that don't exist right now) are too specific - ideally some already existing one could be reused, or if not at least more general would be created

stollerm:
Ah, just seeing the requested changes now! Thanks @LekoArts @pieh and @wardpeet!

PR #3987, opened 2018-02-12
Add note about plugin options

m-allanson:
I'm not sure this is the best place for this note, but there's not (yet) docs that give a more technical overview of how plugins work.

Refs #3834

Issue #3834, opened 2018-02-03
Pass function in plugin options to gatsby-browser.js


HriBB:
How can we pass functions in plugin options?

`app/gatsby-config.js`
```
{
  resolve: `gatsby-plugin-i18next`,
  options: {
    getLanguage(pathname) {
      return pathname.substr(1,2) === 'en' ? 'en' : 'si'
    },
    config: {
      fallbackLng: 'en',
      ns: [],
      interpolation: {
        escapeValue: false, // code escaping in interpolated text is not needed with react
      },
      react: {
        wait: false,
      },
    },
  },
},
```
`gatsby-plugin-i18next/gatsby-node.js`
```
exports.onCreatePage = ({ page } , pluginOptions) => {
  console.log(pluginOptions);
}
```

Output of console.log
```
{ plugins: [],
  config: 
   { fallbackLng: 'si',
     ns: [],
     interpolation: { escapeValue: false },
     react: { wait: false } } }
```
As you can see, `getLanguage` function is not available in plugin options. Am I missing something obvious here?

HriBB:
Hmm wait, `getLanguage` IS available in `gatsby-node.js`, but not in `gatsby-browser.js`

calcsam:
What are you trying to do exactly? 

Gatsby has a build step where pages are generated, where the code from `gatsby-node` are called. It also runs some code client-side, where the code from `gatsby-browser` is called. `getLanguage` deals with client navigation and path, why are you running it in `gatsby-node`?

HriBB:
I was trying to write a i18n plugin, where user would define a function in the plugin config. This function would be used throughout the plugin code, to determine the language from `pathname`. I tried to do something like this in `gatsby-browser.js`:
```
exports.onClientEntry = (obj, { getLanguage }) => {
  //const language = getLanguage(window.location.pathname) // getLanguage is undefined
  const language = window.location.pathname.substr(1,2) === 'en' ? 'en' : 'si'
  i18next.changeLanguage(language)
  }
}

exports.onRouteUpdate = ({ location }, { getLanguage }) => {
  const language = location.pathname.substr(1,2) === 'en' ? 'en' : 'si'
  if (i18next.language !== language) {
    i18next.changeLanguage(language)
  }
}
```
But it's not possible, because `pluginOptions` in `gatsby-browser.js` is `JSON.stringified` version of the config options and does not contain functions. Might be worth mentioning in the docs. You can close this issue I guess ...

[...]

PR #13440, opened 2019-04-18
docs: Remove note about stringified options

CanRau:
I'm not entirely sure about this, I'm aware of the origin https://github.com/gatsbyjs/gatsby/pull/3987 yet this PR I made proofs the opposite https://github.com/gatsbyjs/gatsby/pull/12060 
Only thing I can think of right now that browser APIs get stringified, couldn't verify this so far, this https://github.com/gatsbyjs/gatsby/issues/3834 is suggesting they get stringified and it's the reason the note was added to the docs in the first place..

Couldn't just leave my confusion alone and thought doing it in a PR might be useful if it's actually not valid anymore so it can be merged directly, otherwise feel free to reject it of course 👍

### DIFF
--- a/docs/docs/plugins.md
+++ b/docs/docs/plugins.md
@@ -78,8 +78,6 @@ module.exports = {
 }
 ```
 
-Note that plugin options will be stringified by Gatsby, so they cannot be functions.
-
 ## Loading plugins from your local plugins folder
 
 Gatsby can also load plugins from the your local website plugins folder which is a folder named `plugins` in the website's root directory.


pieh:
Browser APIs options get JSON.stringified ( https://github.com/gatsbyjs/gatsby/blob/9f0835addebb8a89d3dff6010c6ce8ea2c59bbd5/packages/gatsby/src/bootstrap/index.js#L313 )
and SSR APIs options get JSON.stringified ( https://github.com/gatsbyjs/gatsby/blob/9f0835addebb8a89d3dff6010c6ce8ea2c59bbd5/packages/gatsby/src/bootstrap/index.js#L333 )

Node APIs are not getting stringified because they executed in the same context, so they are just used from memory as they are

CanRau:
Okay, good to know, thanks for clarifying 🙏 
Then the note either stays or could get an update like
> Note that plugin options of `gatsby-ssr.js` and `gatsby-browser.js` get stringified by Gatsby, so they cannot be functions.

But maybe it's not the right place 🤔 I think it's a little bit misleading as they can be in certain cases..
What do you think?

mike1808:
I was trying to pass function for `gatsby-browser` and I faced this issue. I think, it's worth mentioning in the docs that options for ssr and browser are stringified and you cannot pass functions.

abumalick:
I think it is worth mentioning somewhere in the plugin docs too.

I don't know in which section it would fix through

https://www.gatsbyjs.org/docs/creating-plugins/

PR #13449, opened 2019-04-18
docs: update tutorial to useStaticQuery

andrewfairlie:
## Description
A colleague of mine came across this page when learning Gatsby and mentioned that the code looked very messy and a bit hacky.

I introduced him to the useStaticQuery hook which clicked a lot more for him. When I first started learning Gatsby I also found the current code sample using the component inelegant and I wonder how many people drop out of the tutorial at this point because of the perceived code-smell.

I don't know if it's the core team's intention for people to be using useStaticQuery over the original component but if they are, might I suggest we update the tutorial as this pull request suggests? I think it'd be more accessible and also introduces the idea of using hooks in a hopefully easy to understand way into the tutorial which I think is a useful addition.

### DIFF
--- a/docs/tutorial/part-four/index.md
+++ b/docs/tutorial/part-four/index.md
@@ -292,7 +292,7 @@ Page queries live outside of the component definition -- by convention at the en
 ### Use a StaticQuery
 
 [StaticQuery](/docs/static-query/) is a new API introduced in Gatsby v2 that allows non-page components (like our `layout.js` component), to retrieve data via GraphQL queries.
-
+Let's use its newly introduced hook version — [`useStaticQuery`](/docs/use-static-query/)
 Go ahead and make some changes to your `src/components/layout.js` file to use the `useStaticQuery` hook and a `{data.site.siteMetadata.title}` reference that uses this data. When you are done your file looks like this:
 
 ```jsx:title=src/components/layout.js
@@ -303,8 +303,8 @@ import { useStaticQuery, Link, graphql } from "gatsby"
 
 import { rhythm } from "../utils/typography"
 
-// highlight-start
 export default ({ children }) => {
+// highlight-start
   const data = useStaticQuery(
     graphql`
       query {
@@ -316,8 +316,8 @@ export default ({ children }) => {
       }
     `
   )
+  // highlight-end
   return (
-    // highlight-end
     <div
       css={css`
         margin: 0 auto;
@@ -347,10 +347,8 @@ export default ({ children }) => {
       </Link>
       {children}
     </div>
-    // highlight-start
   )
 }
-// highlight-end
 ```
 
 Another success! 🎉
--- a/docs/tutorial/part-four/index.md
+++ b/docs/tutorial/part-four/index.md
@@ -293,16 +293,7 @@ Page queries live outside of the component definition -- by convention at the en
 
 [StaticQuery](/docs/static-query/) is a new API introduced in Gatsby v2 that allows non-page components (like our `layout.js` component), to retrieve data via GraphQL queries.
 
-There are two ways to use StaticQuery:-
-
-1. The [StaticQuery component](/docs/static-query/)
-2. The newer [useStaticQuery hook](/docs/use-static-query/)
-
-In this tutorial you'll be using the newer hook method which is cleaner and more succinct, but you may wish to learn about the component version too.
-
-You need to include hooks within the body of your function component. To facilitate this you change the `export` line to include `{` rather than `(`. Also, within the function you'll use `return()`. When you use `(` on your export line you're effectively using a shortcut to writing `return(` each time. But because you now need to run the hook before you return the component you will need to define it.
-
-Go ahead and make some changes to your `src/components/layout.js` file to run a `useStaticQuery` hook and a `{data.site.siteMetadata.title}` reference that uses this data. When you are done your file looks like this:
+Go ahead and make some changes to your `src/components/layout.js` file to use the `useStaticQuery` hook and a `{data.site.siteMetadata.title}` reference that uses this data. When you are done your file looks like this:
 
 ```jsx:title=src/components/layout.js
 import React from "react"
@@ -362,6 +353,8 @@ export default ({ children }) => {
 }
 {/* highlight-end */}
 ```
+
+You've now added the query before you return the component by changing your function to use `{}` instead of `()` and wrapping what was there before with `return()`.
 
 Another success! 🎉
 
--- a/docs/tutorial/part-four/index.md
+++ b/docs/tutorial/part-four/index.md
@@ -300,7 +300,7 @@ There are two ways to use StaticQuery:-
 
 You'll be using the newer hook method which is cleaner and more succinct, but you may wish to learn about the component version too.
 
-With React, Hooks need to be within the body of the function component. Notice the change of the `export` line to `{` rather than `(` and the introduction of `return()`. This includes what was previously the only thing in our function. This allows you to do additional work before you render your component, in this case that means running our static query.
+You need to include hooks within the body of your function component. To facilitate this you change the `export` line to include `{` rather than `(`. Also, within the function you'll use `return()`. When you use `(` on your export line you're effectively using a shortcut to writing `return(` each time. But because you now need to run the hook before you return the component you will need to write it longhand this time.
 
 Go ahead and make some changes to your `src/components/layout.js` file to run a `useStaticQuery` hook and a `{data.site.siteMetadata.title}` reference that uses this data. When you are done your file looks like this:
 
--- a/docs/tutorial/part-four/index.md
+++ b/docs/tutorial/part-four/index.md
@@ -298,7 +298,7 @@ There are two ways to use StaticQuery:-
 1. The [StaticQuery component](/docs/static-query/)
 2. The newer [useStaticQuery hook](/docs/use-static-query/)
 
-You'll be using the newer hook method which is cleaner and more succinct, but you may wish to learn about the component version too.
+In this tutorial you'll be using the newer hook method which is cleaner and more succinct, but you may wish to learn about the component version too.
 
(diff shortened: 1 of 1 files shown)


andrewfairlie:
Sorry, I didn't run prettier on this. Will do now.

docs/tutorial/part-four/index.md
```diff
@@ -293,64 +293,74 @@ Page queries live outside of the component definition -- by convention at the en
 
 [StaticQuery](/docs/static-query/) is a new API introduced in Gatsby v2 that allows non-page components (like our `layout.js` component), to retrieve data via GraphQL queries.
 
-Go ahead and add a `<StaticQuery />` to your `src/components/layout.js` file and a `{data.site.siteMetadata.title}` reference that uses this data. When you are done your file looks like this:
+There are two ways to use StaticQuery:-
```
DSchau: I think it's really important when writing a tutorial to have an opinionated choice here. In other words--as a reader of this document new to the space, my first thought reading this will be one of confusion. _Why_ would I use one or the other? Why am I being given this choice?

So I'd prefer we're consistent here and only introduce one choice.

@marcysutton what do you think?

andrewfairlie: That makes sense and I agree, good thinking.

I will stand by the thought that presenting `useStaticQuery` over `<StaticQuery>` is a good one though, because it feels neater and easier to understand coming at this new. But that might just be me and my colleague 😄

sidharthachatterjee: Just want to chime in to say that you’ll need React and ReactDOM 16.8.0 or later to use `useStaticQuery`

We might want to mention that to avoid confusion if we switch this tutorial to recommend `useStaticQuery`

andrewfairlie: I figured that because those following the tutorial would (presumably) be on the current version which already includes those it would confuse matters to mention it.

andrewfairlie:
Upon @DSchau's suggestion I've removed reference to `<StaticQuery>` in favour of focusing just on `useStaticQuery`. I think this has helped focus the tutorial a lot.

DSchau:
@marcysutton what do you think about this?

Generally - I think we've given `useStaticQuery` enough time to "bake," and it's fully ready to be used. It seems simpler and more easily understandable than render props (StaticQuery), so I think I'm mostly in favor of the change--but as always, curious what you think!

docs/tutorial/part-four/index.md
```diff
@@ -293,66 +293,69 @@ Page queries live outside of the component definition -- by convention at the en
...
-import { StaticQuery, Link, graphql } from "gatsby"
+import { useStaticQuery, Link, graphql } from "gatsby"
 
 import { rhythm } from "../utils/typography"
 
-export default ({ children }) => (
-  {/* highlight-start */}
-  <StaticQuery
-    query={graphql`
+{/* highlight-start */}
```
DSchau: This is not correct. We want our code snippets to be valid JSX (thus the JSX expression embedding comment syntax) but this portion of the code is vanilla JS, e.g.

```suggestion
// highlight-start
```

docs/tutorial/part-four/index.md
```diff
@@ -293,66 +293,69 @@ Page queries live outside of the component definition -- by convention at the en
...
-            {data.site.siteMetadata.title}{/* highlight-line */}
-          </h3>
-        </Link>
-        <Link
-          to={`/about/`}
+    `
+  )
+
+  return(
+  {/* highlight-end */}
```
DSchau: Maybe move this up a line and then use the comment syntax? e.g.

```
// highlight-end
```

docs/tutorial/part-four/index.md
```diff
@@ -293,66 +293,69 @@ Page queries live outside of the component definition -- by convention at the en
...
+        to={`/about/`}
+        css={css`
+          float: right;
+        `}
+      >
+        About
+      </Link>
+      {children}
+    </div>
+  {/* highlight-start */}
```
DSchau: What's the benefit of this? I think this should be removed (and the closing highlight-end tag too)

docs/tutorial/part-four/index.md
```diff
@@ -293,66 +293,69 @@ Page queries live outside of the component definition -- by convention at the en
...
+  )
+}
+{/* highlight-end */}
 ```
 
+You've now added the query before you return the component by changing your function to use `{}` instead of `()` and wrapping what was there before with `return()`.
```
DSchau: I would maybe just drop this, or at least re-word. It's not clear to me the benefit of this statement and what it's attempting to convey.

docs/tutorial/part-four/index.md
marcysutton: I think it shows the closing parenthesis from the return statement, which makes sense to me. I don't think the closing curly bracket needs to be in there, though.

marcysutton:
I agree that it's simpler, and this is a good change. I migrated a Create React App to Gatsby the other day and found hooks much easier to use! Thumbs up from me 👍

docs/tutorial/part-four/index.md
andrewfairlie: By having it on this line I was trying to highlight the introduction of return( which would be a new addition to a user following the tutorial

docs/tutorial/part-four/index.md
andrewfairlie: Because both are a change to what the user had so far in the tutorial I thought highlighting both would be valuable.

I worry that some readers might focus just on the highlighted changes and if we don’t highlight this change it’ll cause them woe and frustration.

Thoughts?

docs/tutorial/part-four/index.md
andrewfairlie: Yep, reading it again this morning I agree. I’ll remove that today.

docs/tutorial/part-four/index.md
andrewfairlie: Updated in latest push.

docs/tutorial/part-four/index.md
andrewfairlie: I kept the return statement in for the reasons in my previous comment, but I have adjusted it to t he comment syntax and seems to work. Could you clarify if that's correct though?

Builds and passes tests correctly locally so assuming it's a-ok.

andrewfairlie:
I've made some changes to the syntax highlighting and removed the paragraph explaining the use of `return()` because I think it made the tutorial lose its focus.

This is building correctly and passing tests locally, but I'm still unsure whether I've used the correct highlighting syntax.

In this version, I've continued to highlight the opening and closing of `return()` because I think it may be unfamiliar to the first time reader working through this tutorial and if we don't highlight it I think some may miss its inclusion and their code might not work leading to frustration.

sidharthachatterjee:
For visual context, this is what the highlight looks after this change:
Screenshot 2019-04-19 19 03 06

Don't mean to nit pick but I would remove the highlights on the `return` start and end

docs/tutorial/part-four/index.md
```diff
@@ -293,64 +293,64 @@ Page queries live outside of the component definition -- by convention at the en
 
 [StaticQuery](/docs/static-query/) is a new API introduced in Gatsby v2 that allows non-page components (like our `layout.js` component), to retrieve data via GraphQL queries.
```
sidharthachatterjee:
```suggestion
Let's use its newly introduced hook version — [`useStaticQuery`](/docs/use-static-query/)
```

Adding some context so that the jump from `StaticQuery` in the introduction to `useStaticQuery` in the next line is more obvious

docs/tutorial/part-four/index.md
```diff
@@ -293,64 +293,64 @@ Page queries live outside of the component definition -- by convention at the en
...
+        to={`/about/`}
+        css={css`
+          float: right;
+        `}
+      >
+        About
+      </Link>
+      {children}
+    </div>
+    // highlight-start
```
sidharthachatterjee:
```suggestion
```

docs/tutorial/part-four/index.md
```diff
@@ -293,64 +293,64 @@ Page queries live outside of the component definition -- by convention at the en
...
+  )
+}
+// highlight-end
```
sidharthachatterjee:
```suggestion
```

docs/tutorial/part-four/index.md
```diff
@@ -293,64 +293,64 @@ Page queries live outside of the component definition -- by convention at the en
...
-          >
-            {data.site.siteMetadata.title}{/* highlight-line */}
-          </h3>
-        </Link>
-        <Link
-          to={`/about/`}
+    `
+  )
+  return (
+    // highlight-end
```
sidharthachatterjee:
```suggestion
```

docs/tutorial/part-four/index.md
```diff
@@ -293,64 +293,64 @@ Page queries live outside of the component definition -- by convention at the en
...
-import { StaticQuery, Link, graphql } from "gatsby"
+import { useStaticQuery, Link, graphql } from "gatsby"
 
 import { rhythm } from "../utils/typography"
 
-export default ({ children }) => (
-  {/* highlight-start */}
-  <StaticQuery
-    query={graphql`
+// highlight-start
```
sidharthachatterjee:
```suggestion
```

docs/tutorial/part-four/index.md
```diff
@@ -293,64 +293,64 @@ Page queries live outside of the component definition -- by convention at the en
...
+export default ({ children }) => {
+  const data = useStaticQuery(
```
sidharthachatterjee:
```suggestion
// highlight-start
  const data = useStaticQuery(
```

docs/tutorial/part-four/index.md
```diff
@@ -293,64 +293,64 @@ Page queries live outside of the component definition -- by convention at the en
...
-              font-style: normal;
-            `}
-          >
-            {data.site.siteMetadata.title}{/* highlight-line */}
-          </h3>
-        </Link>
-        <Link
-          to={`/about/`}
+    `
+  )
```
sidharthachatterjee:
```suggestion
  )
  // highlight-end
```

andrewfairlie:
@sidharthachatterjee sorry for slow response but thanks for helping :)

PR #13452, opened 2019-04-18
fix(gatsby-plugin-netlify): Add Referrer-Policy to security headers

browniebroke:
## Description

Add 'Referrer-Policy: same-origin' to SECURITY_HEADERS.

This is one of the headers checked by securityheaders.com and currently [it's not included by Gatsby by default](https://securityheaders.com/?q=gatsbyjs.org&followRedirects=on). I was adding it to a few of my projects, but I realised it could probably be added upstream, so here it is.

Looking around, it seems that 'same-origin' is a good default, but I don't feel strongly about it. Would this be welcome?

### DIFF
--- a/packages/gatsby-plugin-netlify/src/constants.js
+++ b/packages/gatsby-plugin-netlify/src/constants.js
@@ -21,6 +21,7 @@ export const SECURITY_HEADERS = {
     `X-Frame-Options: DENY`,
     `X-XSS-Protection: 1; mode=block`,
     `X-Content-Type-Options: nosniff`,
+    `Referrer-Policy: same-origin`,
   ],
 }
 


sidharthachatterjee:
Published in `gatsby-plugin-netlify@2.0.14`

rmehner:
This is actually a breaking change for us and it took me a while to find this.

We use a map image provider who doesn't use an API key but the referrer to regulate access. This stopped working for us, as it's not `same-origin` and we'd need `no-referrer-when-downgrade` at minimum.

It's easy to fix by adjusting the gatsby-config:

```js
{
  resolve: 'gatsby-plugin-netlify',
  options: {
    headers: {
      '/*': [`Referrer-Policy: no-referrer-when-downgrade`],
    },
  },
},
```

Should have checked the release notes more thoroughly before updating the package :)

rmehner:
Turns out my fix is wrong, as it will create the header as:

`referrer-policy: no-referrer-when-downgrade, same-origin`

I'll change it to this:

```js
{
      resolve: 'gatsby-plugin-netlify',
      options: {
        mergeSecurityHeaders: false,
        headers: {
          '/*': [
            `X-Frame-Options: DENY`,
            `X-XSS-Protection: 1; mode=block`,
            `X-Content-Type-Options: nosniff`,
            `Referrer-Policy: no-referrer-when-downgrade`,
          ],
        },
      },
    },```

Issue #12754, opened 2019-03-22
Netlify deploy failing (Invalid filename 'icons/icon-96x96.png?v=a19f3881b4ae7f39c6f8740545bdbfb6915cf4b0'. Deployed filenames cannot contain # or ? characters)


jakeherp:
## Description

Since a couple of days ago it keeps failing to deploy to Netlify with the following error:
`8:20:02 AM: failed during stage 'deploying site': Invalid filename 'icons/icon-96x96.png?v=a19f3881b4ae7f39c6f8740545bdbfb6915cf4b0'. Deployed filenames cannot contain # or ? characters`

I have confirmed this to be a problem with `gatsby-plugin-manifest` by locking its version to 2.0.18. In 2.0.18 it compiles fine.

### Steps to reproduce

Cannot reproduce the issue locally, even using Netlify's own Docker container. Files do not get generated with query strings in my local environment.

### Expected result

Gatsby project gets built and served on Netlify server.

### Actual result

Deploy failing due to query string in icon filename.

moonmeister:
@jakeherp Thanks for the report. I wrote that cache busting code and will see if I can reproduce. 

Is there any more context for that error? That should be the url link provided in the html head, not the actual file name I'm not sure from this error which it's complaining about.

It is possible Netlify doesn't support url queries on files. Cache busting can be set to `name` or `none`  mode to fix this. https://www.gatsbyjs.org/packages/gatsby-plugin-manifest/?=#disabling-or-changing-cache-busting-mode

It might not be needed for netlify: https://www.netlify.com/blog/2015/09/11/instant-cache-invalidation/

moonmeister:
I will note that The gatsbyjs.org site is on Netlify and using query base cache busting with no problem. Though maybe someone already made a work around.

moonmeister:
@jakeherp I just deployed the default gatsby starter no problem. If you could post your env info using `gatsby info --clipboard` and please also post the configuration for `gatsby-plugin-manifest` and if you've added a netlify config via `gatsby-plugin-netlify` or a `netlify.toml` those configs as well. Thanks!

jakeherp:
The full log from Netlify is:

```
1:29:13 PM: Build ready to start
1:29:15 PM: build-image version: bc22f7ea2f97324f8ca25c36353605ddd91e092d
1:29:15 PM: build-image tag: v2.6.15
[...]

PR #13471, opened 2019-04-18
feat(gatsby-plugin-manifest): add i18n, localization

### LINKED ISSUE #10250: [gatsby-plugin-manifest] internationalization (i18n)
I'm quite new to PWAs, manifests & add-to-homescreen stuff and I'm not 100% sure I get everything. I am not even sure where to discuss my thoughts/questions, a RFC seems to big for this, and I'm not into those chats like spectrum and discord, it's annoying for me to have to check them etc so I'm making this a question for now and see where it goes, hopefully that's okay for everyone else 🙏 

## Summary
Provide a way for pages to specify separate manifests for each language

## Relevant information
That would be one possible solution, maybe a bit verbose/repeating, though you could define your shared base options and then just extend them but it gives a lot of flexibility so people can define different icons, colors etc (proof that icon localization is a thing https://github.com/w3c/manifest/issues/676#issuecomment-442434811)
```
{
  options: [
    {
      name: `Full name`,
      short_name: `Name`,
      start_url: `/en/?utm_source=a2hs`,
      background_color: `#fff`,
      theme_color: `red`,
      display: `standalone`,
      icon: `static/icon.png`,
      pattern: `^/en/.*$`,
    },
    {
      name: `Vollständiger Name`,
      short_name: `Name`,
      start_url: `/de/?utm_source=a2hs`,
      background_color: `#fff`,
      theme_color: `rebeccapurple`,
      display: `standalone`,
      icon: `static/icon_de.png`,
      pattern: `^/de/.*$`,
    },
  ],
},
```
I would suggest still allowing a single object in options to not break anything.

I run on a patched version I just threw in my local packages workspace https://github.com/GaiAma/gaiama.org/commit/ae194c204c8e34c85bec074ba4d6e353ce9eb981#diff-0cee23a635d1c08646f35342b6dee83f which doesn't seem to work properly right now, but it did work…

Another possibility might be to just specify multiple plugin declarations and "just" add a `pattern` property or similar like:
```
{
  options: {
    name: `GaiAma`,
    short_name: `GaiAma`,
    start_url: `/en/?utm_source=a2hs`,
    background_color: `#fff`,
    theme_color: `#287482`,
    display: `standalone`,
    icon: `static/gaiama_pictogram.png`,
    pattern: `^/en/.*$`,
  },
},
{
  options: {
    name: `GaiAma`,
    short_name: `GaiAma`,
    start_url: `/de/?utm_source=a2hs`,
    background_color: `#fff`,
    theme_color: `#287482`,
    display: `standalone`,
    icon: `static/gaiama_pictogram.png`,
    pattern: `^/de/.*$`,
  },
},
```
So the pattern would specify in gatsby-ssr which to run/include

I'm not sure if the manifests should be then save in the respective language directory all called `manifest.webmanifest` or put them all in `/` root and name them differently? 

Another question I have is, what about the manual trigger stuff and listening on `beforeinstallprompt` in chrome https://developers.google.com/web/fundamentals/app-install-banners/#trigger this plugin should probably support it sooner than later right?

Let me know what you think about all this, any suggestions/feedback are really appreciated 🙏 
I think I could implement the i18n stuff properly, if it gets accepted and we settled on an implementation as I don't want to keep my local copy up to date.. I might be able to do the trigger stuff, but that could probably enter another PR

CanRau:
## Description

As introduced here https://github.com/gatsbyjs/gatsby/issues/10250 I thought why not just update my fork and push it

So to use i18n you use the new `manifests` property and give it an array with all versions.

```js
// in gatsby-config.js
module.exports = {
  plugins: [
    {
      options: {
        manifests: [
          {
            name: `GatsbyJS`,
            short_name: `GatsbyJS`,
            start_url: `/de/`,
            regex: `^/de/.*`,
            language: `de`,
            background_color: `#f7f0eb`,
            theme_color: `#a2466c`,
            display: `standalone`,
          },
          {
            name: `GatsbyJS`,
            short_name: `GatsbyJS`,
            start_url: `/`,
            regex: `.*`,
            background_color: `#f7f0eb`,
            theme_color: `#a2466c`,
            display: `standalone`,
          },
        ],
      },
    },
  ],
}
```

The `language` prop works as a filename suffix, so the first array entry will write to `manifest_de.webmanifest` and as a default regex (which is not documented so far).

All manifests are written to the root of `/public`

It works so far, I'm not exactly sure about all coding styles or if my tests are sufficient, still not enough into testing (really have to get into it!!).

So please let me know if I can improve anything, I'm keen to learn.

## Related Issues

Closes #10250


KyleAMathews:
Huh very cool! I love the idea of supporting i18n manifests!

Having both `language` and `regex` for selecting the language seems a bit cumbersome. Any reason for having both? Also the key name `regex` isn't overly obvious what it's for. A more verbose key name might be more self-documenting perhaps.

CanRau:
Kay agree on the naming 👍
The regex is to provide various styles of language differentiation but falls back to beginning of path like `/en/path`
In this case it's optional
I thought people might have another setup like `path?lang=en` even though it's usually not recommended as I understand

Any thoughts or ideas? We could provide only the first for now and see what feedback we get to start simple..so regex wouldn't be customizable only language has to be defined..

moonmeister:
@CanRau I love this idea as well. First question, how does the browser know which manifest to use? Second, have you put any thought into implementing this in a way that doesn't break the existing API? If so, is there a reason you didn't choose this method? 

Thanks for sharing!

moonmeister:
Just read #10250 and saw your discussion around non-breaking API changes. Please expand on why you didn't follow that path if possible. 

FYI: I know the manifest plugin currently blows up if you configure it multiple times. I believe this could be fixed but it's been a bit since I realized this so I don't remember what causes it. #12754 for ref

CanRau:
Not sure what is breaking in the proposed API? It's just adding some props, all existing tests pass without a problem. Um, maybe I'm missing something? Just realized the API I proposed in https://github.com/gatsbyjs/gatsby/issues/10250 is indeed a little different, the first proposal wouldn't work because pluginOptions are always an `object` at least at this moment, the second proposal having multiple instances of the plugin seems more complicated, the proposal here in the PR is simpler and would easily allow one to define all manifests in another file. Think it's more complicated to determine which manifest to include in the `<head/>` maybe using gatsby's store could work don't know..

When you set it up as in the [readme example](https://github.com/gatsbyjs/gatsby/blob/3eee6b95844afec3a12f0853ac4a2dc63e3dbe6b/packages/gatsby-plugin-manifest/README.md#i18n--multilang-configuration) it'll generate all manifests for all languages in the root of the project like `/manifest_en.webmanifest` which is happening [here in gatsby-node.js](https://github.com/gatsbyjs/gatsby/blob/3eee6b95844afec3a12f0853ac4a2dc63e3dbe6b/packages/gatsby-plugin-manifest/src/gatsby-node.js#L50-L54) if you've defined the new `manifests` prop as an array it'll create multiple manifests otherwise one, just like before

And [here in gatsby-ssr.js](https://github.com/gatsbyjs/gatsby/blob/3eee6b95844afec3a12f0853ac4a2dc63e3dbe6b/packages/gatsby-plugin-manifest/src/gatsby-ssr.js#L13-L18) it'll pick the right manifest from the `manifests` array based on the `language` and the `regex` which matches against the page `pathname` which looks like `/en/awesome-page` so that it sets the proper headComponent.

moonmeister:
@CanRau Thanks for the detailed explanation. I just saw the `manifests: []` and assumed that would break the existing API. But you've handled that well.  

So you've given examples on how to not duplicate cod useing Assign or rest/spread operators.

moonmeister:
Wrong button...

packages/gatsby-plugin-manifest/src/gatsby-node.js
```diff
@@ -43,8 +43,22 @@ function generateIcons(icons, srcIcon) {
...
 }
 
-exports.onPostBootstrap = async ({ reporter }, pluginOptions) => {
-  const { icon, ...manifest } = pluginOptions
+exports.onPostBootstrap = async ({ reporter }, { manifests, ...manifest }) => {
+  const activity = reporter.activityTimer(`Build manifest and related icons`)
+  activity.start()
+
+  if (Array.isArray(manifests)) {
+    await Promise.all(manifests.map(x => makeManifest(reporter, x)))
```
moonmeister:
```suggestion
    await Promise.all(manifests.map(manifest => makeManifest(reporter, manifest)))
```

packages/gatsby-plugin-manifest/src/gatsby-ssr.js
```diff
@@ -6,7 +6,17 @@ import fs from "fs"
 
 let iconDigest = null
 
-exports.onRenderBody = ({ setHeadComponents }, pluginOptions) => {
+exports.onRenderBody = (
+  { setHeadComponents, pathname = `/` },
+  pluginOptions
+) => {
+  if (Array.isArray(pluginOptions.manifests)) {
+    pluginOptions = pluginOptions.manifests.find(x =>
```
moonmeister:
```suggestion
    pluginOptions = pluginOptions.manifests.find(manifest =>
```

packages/gatsby-plugin-manifest/src/gatsby-ssr.js
```diff
@@ -6,7 +6,17 @@ import fs from "fs"
...
+      RegExp(x.regex || `^/${x.language}/.*`).test(pathname)
```
moonmeister:
```suggestion
      RegExp(manifest.regex || `^/${manifest.language}/.*`).test(pathname)
```

packages/gatsby-plugin-manifest/src/gatsby-ssr.js
```diff
@@ -93,4 +105,5 @@ exports.onRenderBody = ({ setHeadComponents }, pluginOptions) => {
   }
 
   setHeadComponents(headComponents)
+  return true
```
moonmeister: Any reason this was added? it's not needed so not sure it needs to be here.

packages/gatsby-plugin-manifest/src/gatsby-node.js
moonmeister: This just makes it a little more readable.

moonmeister: So currently this would run the icon creation for every manifest. The icons aren't being internationalized. In automatic mode they couldn't be without code modification. I do see a use case for i18n of icons...but for those who aren't creating all these icons over and over seems pointless. We should probably add a caching mechanism to fix this. That can be a separate PR though. no need to solve it here as it technically affects the current plugin regenerating icons on every build currently even if they already exist.

For now, should we be concerned with adding the i18n suffix to icons? This would fix the automatic mode issue of writing all icons to the same files.

packages/gatsby-plugin-manifest/src/gatsby-ssr.js
moonmeister: This just makes it a little more readable.

packages/gatsby-plugin-manifest/src/gatsby-ssr.js
moonmeister: This just makes it a little more readable.

packages/gatsby-plugin-manifest/README.md
```diff
@@ -130,6 +134,79 @@ icons: [
 
 In the manual mode, you are responsible for defining the entire web app manifest and providing the defined icons in the [static](https://www.gatsbyjs.org/docs/static-folder/) folder. Only icons you provide will be available. There is no automatic resizing done for you.
 
+### i18n – Multilang configuration
```
moonmeister:
```suggestion
#### i18n – Multilang configuration
```
This whole section should be placed under "Feature Configuration" heading. This conforms to the progressive nature of the docs flow.

packages/gatsby-plugin-manifest/README.md
```diff
@@ -69,6 +69,10 @@ There are three modes in which icon generation can function: automatic, hybrid,
   - Legacy icon support - yes
   - Cache busting - never
 
+- i18n - Generate separate versions for multiple languages
```
moonmeister: Should be listed as a feature under each mode.

packages/gatsby-plugin-manifest/src/gatsby-ssr.js
CanRau: Added it due to the eslint rule consistent-return complained, because of this line https://github.com/gatsbyjs/gatsby/blob/4844244b2e1579ea2833667a576301fb88e661f8/packages/gatsby-plugin-manifest/src/gatsby-ssr.js#L17

Any suggestion?

moonmeister: Ah, I missed the other return got added. If that need to be there that's fine. Thanks

packages/gatsby-plugin-manifest/README.md
CanRau: Like
```
- Cache busting - yes
- i18n - yes
```
or better `internationalization`?

moonmeister: Maybe `- internationalization - optional` Cause `yes` implies it's the default behavior (or at least that was my intent when I wrote this).

CanRau:
Thank you for your suggestions/comments @moonmeister 🙏 
Any thoughts on @KyleAMathews' request: https://github.com/gatsbyjs/gatsby/pull/13471#issuecomment-485499592
And my proposals https://github.com/gatsbyjs/gatsby/pull/13471#issuecomment-485579061 about config naming, especially the `regex` prop?

moonmeister:
@CanRau I understand your reasoning. My thought is let's keep it simple. If using the path is best practice. then let's just build for that. If someone decides they really want to not use path they could contribute that functionality themselves. This keeps the API simpler and more approachable and only adds complexity if the community needs it. 

If we do decide to keep this let's just make sure it' clearly documented that only one of the options is need. Or we maybe design it so it is a single option. Meaning someone can just type the path or a Regex...then the code determines what it is and uses it appropriately?

CanRau:
> @CanRau Thanks for the detailed explanation. I just saw the manifests: [] and assumed that would break the existing API. But you've handled that well.

No worries, we all have a lot to do, all fine 👊 😉 

> Anything that is duplicated in the `manifests` array should override what's in the top level `options`. This seems like a less drastic change that is easier to convert to if your adding i18n to an existing manifest config.

I like it 👍 going to bed now, integrating it tomorrow 

I'm open to dropping the `regex` altogether, as I think it seems path based seems the preferred by Google, at least over query params..

> Meaning someone can just type the path or a Regex...then the code determines what it is and uses it appropriately?

Would love to know how you'd do that? I mean finding the proper manifest in `gatsby-ssr.js` works already like this, but the name suffix in both, how would you do this? Just out of curiosity 🤔 Cause I think just dropping this customization might be the best for now anyway, to have a smaller and simpler API.

packages/gatsby-plugin-manifest/src/gatsby-node.js
CanRau: Haven't thought about the icon suffix 😨 Can do this tomorrow 👍 I didn't even include the icon in the i18n readme example, could be more complex than maybe necessary, but we could check where the icon has been defined, as shared or language specific and add the suffix only in the latter case.. what do you think? 🤔

moonmeister: That's a great idea. Suggestion: process the "default" first (i.e. what's in the options and not the "manifest" array). Then as you iterate through the `manifest: []` only if icon is uniquely specified in there or if `icons` is specified do you generate new icons...otherwise you just reference the default ones.

Does that make sense? there maybe other things to consider in this implementation. It would also be simpler to just always create unique icons. :man_shrugging:

moonmeister:
> Would love to know how you'd do that? I mean finding the proper manifest in `gatsby-ssr.js` works already like this, but the name suffix in both, how would you do this? Just out of curiosity  Cause I think just dropping this customization might be the best for now anyway, to have a smaller and simpler API.

Then let's just KIS (Keep it simple!) and eliminate regex. Using a single option to be one or the other would probably get complicated...but you could probably write a regex to determine if it was regex :roll_eyes:. It's a bad idea...just ignore me.

packages/gatsby-plugin-manifest/README.md
```diff
@@ -143,25 +142,22 @@ module.exports = {
...
             regex: `^/de/.*`,
             language: `de`,
             background_color: `#f7f0eb`,
             theme_color: `#a2466c`,
-            display: `standalone`,
           },
           {
-            name: `GatsbyJS`,
-            short_name: `GatsbyJS`,
             start_url: `/`,
```
moonmeister: My thought is there's no need to explicitly define the root or default manifest here. This can all be assumed from the root of options and only additional manifests need be defined here.

moonmeister:
Just realized we need to make sure this accounts for the use of [`pathPrefix`](https://www.gatsbyjs.org/docs/path-prefix/)

LekoArts:
Before merging this I want to discuss if this is actually the best approach: https://github.com/gatsbyjs/gatsby/issues/10250#issuecomment-486177665

packages/gatsby-plugin-manifest/README.md
CanRau: Like so
```
{
      options: {
        name: `GatsbyJS`,
        short_name: `Gatsby`,
        display: `standalone`,
        icon: `src/images/icon.png`,
        start_url: `/`,
        background_color: `#663399`,
        theme_color: `#fff`,
        manifests: [
          {
            start_url: `/de/`,
            language: `de`,
            background_color: `#f7f0eb`,
            theme_color: `#a2466c`,
          },
        ],
      },
    },
```
you mean? Maybe we then rename `manifests` to something like `localized_manifests` or `additional_manifests` (probably too vague)

moonmeister: Exactally. I'm torn on naming. Let's make sure to address @lekoarts comments on the issue for now.

CanRau:
> Just realized we need to make sure this accounts for the use of [`pathPrefix`](https://www.gatsbyjs.org/docs/path-prefix/)

@moonmeister  Think it's already using [`withPrefix`](https://www.gatsbyjs.org/docs/gatsby-link/#add-the-path-prefix-to-paths-using-withprefix) in `gatsby-ssr.js` which should still work as I'm pretty sure I didn't break former behavior?!

> Before merging this I want to discuss if this is actually the best approach: [#10250 (comment)](https://github.com/gatsbyjs/gatsby/issues/10250#issuecomment-486177665)

Thanks for bringing this up @LekoArts, can't recall my tests and research exactly.

Not sure how I missed the [`lang`](https://developer.mozilla.org/en-US/docs/Web/Manifest#lang) property of manifests

Think it's interesting Google uses [`manifest.json`](https://developer.chrome.com/webstore/i18n#manifest) whereas [this from RealFaviconGenerator](https://github.com/RealFaviconGenerator/realfavicongenerator/issues/372) [pointing to](https://github.com/RealFaviconGenerator/realfavicongenerator/issues/342) brought me to Mozilla where they [explicitly state](https://developer.mozilla.org/en-US/docs/Web/Manifest#Deploying_a_manifest_with_the_link_tag) it should use the `.webmanifest` extension ^^

Pretty sure I've come to the current naming convention from some Stackoverflow but can't find it right now 🤔 think I tested my implementation on our website and it worked, providing the correct manifest depending on the language in pathname, but just tested Chrome though...

So should we follow the [Chrome Webstore](https://developer.chrome.com/webstore/i18n) suggestion? Not even 💯 % sure it really applies to web-apps?
I mean we could "just" stuff the manifests in the appropriate language subfolder and don't add a suffix

Probably because of [this Google web fundamentals](https://developers.google.com/web/fundamentals/web-app-manifest/) I didn't realized the `lang` property, they're not mentioning it

And then it seems sooner or later people have to implement a [`beforeinstallprompt` event listener](https://developers.google.com/web/fundamentals/app-install-banners/#listen_for_beforeinstallprompt) themselves, any thoughts on this? Should we just explain this in the readme or provide some customizable/importable basic version?

moonmeister:
So, I think we need to figure out what of this is actual spec and what of this is Google. Mozilla docs are likely to be closer to the spec. We should build this to enable any spec that exists. I can do more research for tomorrow, for now it's bed time.

CanRau:
Have a good rest then 😴 👍 

Here's some discussion going on in https://github.com/w3c/manifest/issues/676 but there's more to find [w3c/manifest?q=i18n](https://github.com/w3c/manifest/issues?utf8=%E2%9C%93&q=i18n)

moonmeister:
Curiosity is the enemy of sleep:

https://www.w3.org/TR/appmanifest/#internationalization

My reading of this is "do what works". For static sites, the proposed seems a good method. Maybe just use the built in "lang" instead of "language".

Also, I think event listeners are optional customizations...not required to add to home screen...don't quote me I could be wrong...but at least FF is like this.

CanRau:
> Also, I think event listeners are optional customizations...not required to add to home screen...don't quote me I could be wrong...but at least FF is like this.

Might not be forever optional https://love2dev.com/blog/beforeinstallprompt/ and this
>Success: you may want to wait before showing the prompt to the user, so you don't distract them from what they're doing. For example, if the user is in a check-out flow, or creating their account, let them complete that before interrupting them with the prompt.

[Right below the code block](https://developers.google.com/web/fundamentals/app-install-banners/#notify_the_user_your_app_can_be_installed) sounds reasonable to me

Kay I'm renaming `language` to lang now and include it in the manifest right?
And process the root manifest even if `manifests` array is defined correct?
Did I get it right that it might be interesting to always include `lang`? Should we add it to the docs?
How do we name `manifests`?

moonmeister:
Let's leave the event listeners to another PR but sounds like you could open an issue so we don't forget. 

Yes. 

Yes.

I wouldn't worry about adding it outside the l10n docs

Keep the `.webmanifest` extension as the spec uses this. I don't think it matters but no point in changing it.

CanRau:
Okay then 👍 

> How do we name `manifests`?

Meant how do we name the currently [`manifests` array prop in options](https://github.com/gatsbyjs/gatsby/pull/13471/files#diff-59d121de99abb8bcad2eb74ba715379bR148), or do we just keep it?

moonmeister:
I like `localize`.

CanRau:
> I like `localize`.

👍 great
Just realized that `start_url` could've already been a good matcher 🤦‍♂️ as I'm figuring out the root manifest integration think it's helpful

CanRau:
As I'm not sure if you, @moonmeister, have been notified about my last updates, probably not?! Please let me know cause I still don't know all ins and outs of GH^^

packages/gatsby-plugin-manifest/src/gatsby-node.js
CanRau: Realized the icon suffix is still missing..so sorry for the ping^^ 🙏

CanRau: Hum, any thoughts on it? Right now as we're merging the defaults directly with the localized versions I'd have to probably change the way data gets passed to [makeManifest](https://github.com/gatsbyjs/gatsby/pull/13471/files#diff-4606cc9066942a1129552adc6bd0b87dR54)

> It would also be simpler to just always create unique icons.

Any news on how do you feel about it?

moonmeister: I think providing unique icons for a language is some what rare. Additionally, creating unique icons by default sounds generally wasteful, especially if no one is ever needing this. 

Basically, no matter how we do this we have to track what's been done already. We could do it like I suggested above...or we could check the input file and see if its digest has already been seen and written to disk, if so...then we use default file paths, otherwise we generate new file paths. Then add them the new digest to running log. If the digest is used again we;ll know when can use that same path.

moonmeister:
@CanRau I did some work on the docs but cloning a fresh gatsby repo from your fork is taking ages(it's not you, just slow internet and massive repo)...Here's the patch if you could apply it that'd be great. Let me know if you have any questions or feedback on my changes.

https://gist.githubusercontent.com/moonmeister/b0157ce5bc43390a67f54c1846dfe420/raw/012afa26efadbcf8627e55d1da1c2480ff282904/l10ndocs.patch

`git apply l10ndocs.patch`

CanRau:
Yea I know that too well, I'm in the jungles of Perú so internet does its own thing 😅 that's why I try to pull upstream changes as often as possible into my local gatsby fork so it won't grow too much 😆 

Thanks for the patch + hint how to use it, never applied a git patch before 🙏 

Think I merged all your changes, like them, thanks for your help & time 💜 
Your readme is great 👏 
I'm going to bed now 😴 Have a great week 🥇

moonmeister:
> Do I have to manually close all your suggestions clicking "Resolve conversation"? GitHub is not detecting? Or does that happen after the checks pass? Or any other trick?

If you accepted them through git hub it'd autoresolve...I'm not sure if you manually do the change how it works...probably need to do it manually...I can do it.

CanRau:
Uh, that's what changed we both deleted the .patch file almost simultaneously 🤣

CanRau:
@moonmeister Pinging cause I'm not sure what the next step is? Have I missed anything?

packages/gatsby-plugin-manifest/src/gatsby-node.js
moonmeister: @CanRau Has this been resolved?

CanRau: No, you're right, only thing is that you're now able to define a default icon for all languages, which would still run on every language

So, I'm going to implement

> process the "default" first (i.e. what's in the options and not the "manifest" array). Then as you iterate through the manifest: [] only if icon is uniquely specified in there or if icons is specified do you generate new icons...otherwise you just reference the default ones

then?

moonmeister: Dude, this is so simple and clean I'm slightly embarrassed I didn't think of it...good work. Once this is done I'll do a final review, but atm I think that covers all the issues.

CanRau: Not sure I get your comment but I just quoted your suggestion https://github.com/gatsbyjs/gatsby/pull/13471#discussion_r277966669 😉

CanRau:
Wow, that was a long time, didn't expect that, I'm sorry for just "disappearing" 🤦‍♂ a lot going on here in the jungle right now, then my mom is visiting us and they're cutting electricity for about 15 days from next Saturday it seems 😦 

@moonmeister, or anyone else, just so you know I don't claim anything, if you find time and interest implementing the icon "fix" or anything else please got for it 🙏 
If not, which is great, too, I know I'm not the only one being busy 😅 I'll continue once I get time for coding which might take a couple weeks unfortunately though

Thanks for your patience guys, sorry again for disappearing without any comment ❤️

moonmeister:
@canrau no worries! I've kinda done the same thing lately. Enjoy family time and the quiet peace of no electricity. 🤷‍♂️

freiksenet:
Hi @CanRau!

Sorry for dropping the ball on this PR. I wonder what's the status? It seems to be good to merge apart from the merge conflicts.

Thanks! :purple_heart: :muscle:

moonmeister:
@freiksenet There should be one remaining unresolved conversation above that needs to be fixed, it's a significant bug. I have time to take a look so I'll see if I can get it wrapped and ready.

packages/gatsby-plugin-manifest/src/gatsby-node.js
moonmeister: Okay, this should be resolved. I ended up just requiring the name based cache busting on non-default locales with unique icons (automatic mode only) to keep icons from being overridden.

moonmeister:
@freiksenet Alright, I think I've resolved that issue and added some basic caching to keep duplicate icons from being rebuilt multiple times. The caching made the existing tests do some weird things but I think I have everything passing in a valid way. This *Should* be ready to merge now.

freiksenet:
Thank you @moonmeister @CanRau!

sidharthachatterjee:
Published in `gatsby-plugin-manifest@2.2.1`

packages/gatsby-plugin-manifest/src/gatsby-node.js
```diff
@@ -76,6 +76,10 @@ exports.onPostBootstrap = async ({ reporter }, { localize, ...manifest }) => {
         makeManifest(cache, reporter, {
           ...manifest,
           ...locale,
+          ...{ cache_busting_mode: `name` },
+          /* localization requires unique filenames for output files if a different src Icon is defined.
```
dkthehuman: @moonmeister I'm looking into modifying this plugin to be more extensible and fix some existing bugs like #17255. I'm trying to understand some of the cache busting logic but parts of it are pretty confusing so I was hoping you could explain some parts. Why does localization require name-based cache busting rather than being query-based?

moonmeister: @dkthehuman first the code you show above is outdated, there are more recent commits that only require name based busting when you override the icon for a specific locale.  Without this the icons of the different locales would overwrite each other.  When I added caching to the icon generation i was able to detect if it was a different icon being generated for different locales, but I still needed away to write a different file name. The name based cache busting does just that, so it's required. FYI: cache busting doesn't generate a random hash to bust the cache. It generates a hash based on the source image. meaning you can run the build infinite times and cache busting will always have the same hash in the name or query string.

dkthehuman: Ah, gotcha. Thanks!

PR #13476, opened 2019-04-19
fix: raise Unexpected Errors in resolve-module-exports

ChristopherBiscardi:
On stream today I ran into an issue in the mdx-js repo when trying to upgrade
gatsby-plugin-mdx to 1.0. The issue was that the mdx-js repo happened to have an
example with a package name of `webpack` and `gatsby-plugin-mdx` requires
`webpack` in it's `gatsby-node.js`. As a result, we get a "can not find module
webpack" error, but it gets swallowed by resolve-module-exports when trying to
find and require `gatsby-node`.

This PR checks for "Can not find module `modulePath`" and if the error or module
aren't what we expect, we forward the error to the user.


ChristopherBiscardi:
cc/ @johno @lannonbr who were on the stream and have context on the bug.

PR #13477, opened 2019-04-19
Put gatsbyjs.org on a diet

KyleAMathews:
This drops gatsbyjs.org's app bundle from 664kb to 222kb 😱 

### Switch from lodash to lodash-es

I noticed a number of components were pulling in the full lodash. I
removed that in favor of lodash-es which supports tree-shaking and
swapped out all lodash imports

### Lazy load Algolia's terrible React components

They're huge (~250kb) and kinda annoying and we got to get rid of them
someday but in the meantime, let's not load them unless we actually need
them.

### DIFF
--- a/www/src/layouts/index.js
+++ b/www/src/layouts/index.js
@@ -1,8 +1,17 @@
-import React from "react"
-import PluginLibraryWrapper from "../components/layout/plugin-library-wrapper"
+import React, { useState } from "react"
 
+let PluginLibraryWrapper
 export default props => {
-  if (props.pageContext.layout === `plugins`) {
+  const [loaded, setLoaded] = useState(false)
+
+  const promise = import(`../components/layout/plugin-library-wrapper`)
+  if (props.pageContext.layout === `plugins` && !loaded) {
+    promise.then(pl => {
+      PluginLibraryWrapper = pl.default
+      setLoaded(true)
+    })
+    return null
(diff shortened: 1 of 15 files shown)


www/src/layouts/index.js
```diff
@@ -1,8 +1,17 @@
-import React from "react"
-import PluginLibraryWrapper from "../components/layout/plugin-library-wrapper"
+import React, { useState } from "react"
 
+let PluginLibraryWrapper
 export default props => {
-  if (props.pageContext.layout === `plugins`) {
+  const [loaded, setLoaded] = useState(false)
+
+  const promise = import(`../components/layout/plugin-library-wrapper`)
```
KyleAMathews: Lazy loading with hooks!

DSchau:
@KyleAMathews wonder why the lodash plugin wasn’t working?

https://github.com/gatsbyjs/gatsby/blob/058c1a51516cd4df7be7718bc006d12624f0d014/www/gatsby-config.js#L123

KyleAMathews:
Oh... good point — I was mistaken actually seeing all the lodash imports in the sourcemap — those, as it turned out, were from react-instantsearch — but I didn't figure that out until later.

I still think moving to lodash-es makes more sense — gatsby-plugin-lodash was from Gatsby v1 where we couldn't do treeshaking yet but with treeshaking, lodash-es is probably better. At least less ad-hoc.

I'll remove gatsby-plugin-lodash.

PR #13478, opened 2019-04-19
feat(gatsby-source-wordpress): Add retry and bail out for image download failures

### LINKED ISSUE #10347: gatsby-wordpress-source image localFile null / if i add new Image somewhere on the page another Posts Image localFile becomes null
## Description

I have a several custom post types which have either an repeater for images or a featured image. Until a random lower amount of posts ~20 everything seems fine but than if i add new images to the page like for example in my projects post, the localfile for another post not related to the one becomes null. i could not detect any pattern for that. it seems random and looks like there is just certain space / limit available and if that space is full new localFile items override the old ones. i tried deleting .cache and public and reuploading the images but no effect. 

when i check the api response from wordpress everything is fine and the image sources are all in the api response

i first thought that the problem was when i upload a image which is smaller in its natural dimensions than defined in the graphql query like this.

```
project_images {
          image {
            localFile {
              childImageSharp {
                fixed(width: 940) {
                  src
                  width
                  height
                  originalName
                }
              }
            }
          }
```

e. g i upload an image with 900px width. if i do so localFile will be null. 

I will get this in the terminal during build 
Generating image thumbnails [============================   ] 14/16 0.5 secs 100%
The requested width "940px" for a resolutions field for
the file /Users/arturnagenrauft/dev/si-page/.cache/gatsby-source-filesystem/0a7d60e9c09b0ab319d3d49daa855b4f.jpg
was larger than the actual image width of 900px!
If possible, replace the current image with a larger one.

but i reduced the fixed width so the images are not to big for the required width but i still get this issue

### Expected result

I would expect null for entries which have no images but in this case the thumbnail should still get generated and in the localFile be populated.

### Actual result

`TypeError: Cannot read property 'localFile' of null`

localFile seems to get deleted by another localFile

### Environment

gatsby-config.js

```
const config = require('./data/SiteConfig')

const pathPrefix = config.pathPrefix === '/' ? '' : config.pathPrefix

module.exports = {
  pathPrefix: config.pathPrefix,
  siteMetadata: {
    siteUrl: config.siteUrl + pathPrefix,
    rssMetadata: {
      site_url: config.siteUrl + pathPrefix,
      feed_url: config.siteUrl + pathPrefix + config.siteRss,
      title: config.siteTitle,
      description: config.siteDescription,
      image_url: `${config.siteUrl + pathPrefix}/logos/logo-512.png`,
      author: config.userName,
      copyright: config.copyright
    }
  },
  plugins: [
    'gatsby-plugin-sass',
    'gatsby-plugin-react-helmet',
    //'gatsby-plugin-page-transitions',
    {
      resolve: 'gatsby-source-wordpress',
      options: {
        // The base url to your WP site.
        baseUrl: 'si.mobydigg.de',
        // baseUrl: 'wpgatsby.wtf',
        // WP.com sites set to true, WP.org set to false
        hostingWPCOM: false,
        // The protocol. This can be http or https.
        protocol: 'http',
        // Use 'Advanced Custom Fields' Wordpress plugin
        useACF: true,
        auth: {},
        // Set to true to debug endpoints on 'gatsby build'
        verboseOutput: true,
        includedRoutes: [
          "/*/*/categories",
          "/*/*/posts",
          "/*/*/projects",
          "/*/*/jobs",
          "/*/*/pages",
          "/*/*/media",
          "/*/*/tags",
          "/*/*/taxonomies",
          "/*/*/menus"
        ],
      }
    },
    'gatsby-transformer-sharp',
    'gatsby-plugin-sharp',
    // {
    //   resolve: 'gatsby-plugin-google-analytics',
    //   options: {
    //     trackingId: config.googleAnalyticsID,
    //   },
    // },
    {
      resolve: 'gatsby-plugin-nprogress',
      options: {
        color: config.themeColor
      }
    },
    'gatsby-plugin-catch-links',
    'gatsby-plugin-sitemap',
    {
      resolve: 'gatsby-plugin-manifest',
      options: {
        name: config.siteTitle,
        short_name: config.siteTitle,
        description: config.siteDescription,
        start_url: config.pathPrefix,
        background_color: config.backgroundColor,
        theme_color: config.themeColor,
        display: 'minimal-ui',
        icons: [
          {
            src: '/logos/logo-192x192.png',
            sizes: '192x192',
            type: 'image/png'
          },
          {
            src: '/logos/logo-512x512.png',
            sizes: '512x512',
            type: 'image/png'
          }
        ]
      }
    },
    'gatsby-plugin-offline'
  ]
}
```

  System:
    OS: macOS High Sierra 10.13.2
    CPU: (4) x64 Intel(R) Core(TM) i5-7267U CPU @ 3.10GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
  Browsers:
    Chrome: 71.0.3578.80
    Firefox: 63.0
    Safari: 11.0.2
  npmPackages:
    gatsby: ^2.0.63 => 2.0.63
    gatsby-link: ^2.0.7 => 2.0.7
    gatsby-plugin-catch-links: ^2.0.9 => 2.0.9
    gatsby-plugin-feed: ^2.0.11 => 2.0.11
    gatsby-plugin-google-analytics: ^2.0.8 => 2.0.8
    gatsby-plugin-manifest: ^2.0.11 => 2.0.11
    gatsby-plugin-nprogress: ^2.0.7 => 2.0.7
    gatsby-plugin-offline: ^2.0.18 => 2.0.18
    gatsby-plugin-sharp: ^2.0.14 => 2.0.14
    gatsby-remark-copy-linked-files: ^2.0.7 => 2.0.7
    gatsby-remark-images: ^3.0.1 => 3.0.1
    gatsby-remark-prismjs: ^3.1.4 => 3.1.4
    gatsby-remark-responsive-iframe: ^2.0.7 => 2.0.7
    gatsby-source-wordpress: ^3.0.19 => 3.0.19
    gatsby-transformer-remark: ^2.1.15 => 2.1.15
  npmGlobalPackages:
    gatsby-cli: 2.4.5

kristojorg:
## Description

The fundamental issue is that often in development, an image will fail to download and the resulting `localFile` will be `null`. This sometimes happens to many images, especially with large builds or bad internet. 

This PR uses https://github.com/zeit/async-retry to give `createRemoteFileNode` two chances at properly downloading an image. If both of those fail, it will throw the resulting error (in production). In development it will delete the parent node `wordpress__wp_media` node to try and allow developers to continue working even if with only some images. 

The PR is a draft because I actually do not have a WP demo site to test it on, and failed when trying to get one set up. If someone can point me to a working one that would be great. This also omits tests. 

## Related Issues

Fixes #10347

### DIFF
--- a/packages/gatsby-source-wordpress/package.json
+++ b/packages/gatsby-source-wordpress/package.json
@@ -9,6 +9,7 @@
   "bundledDependencies": [],
   "dependencies": {
     "@babel/runtime": "^7.0.0",
+    "async-retry": "^1.2.3",
     "axios": "^0.18.0",
     "better-queue": "^3.8.6",
     "bluebird": "^3.5.0",
--- a/packages/gatsby-source-wordpress/src/normalize.js
+++ b/packages/gatsby-source-wordpress/src/normalize.js
@@ -2,7 +2,7 @@ const deepMapKeys = require(`deep-map-keys`)
 const _ = require(`lodash`)
 const { createRemoteFileNode } = require(`gatsby-source-filesystem`)
 const { URL } = require(`url`)
-
+const retry = require(`async-retry`)
 const colorized = require(`./output-color`)
 const conflictFieldPrefix = `wordpress_`
 // restrictedNodeFields from here https://www.gatsbyjs.org/docs/node-interface/
@@ -475,6 +475,7 @@ exports.downloadMediaFiles = async ({
   createNode,
   createNodeId,
   touchNode,
+  deleteNode,
   getNode,
   _auth,
 }) =>
@@ -503,15 +504,19 @@ exports.downloadMediaFiles = async ({
         // If we don't have cached data, download the file
         if (!fileNodeID) {
           try {
-            const fileNode = await createRemoteFileNode({
-              url: e.source_url,
-              store,
-              cache,
-              createNode,
-              createNodeId,
-              parentNodeId: e.id,
-              auth: _auth,
-            })
+            const fileNode = await retry(
+              () =>
+                createRemoteFileNode({
+                  url: e.source_url,
+                  store,
+                  cache,
+                  createNode,
+                  createNodeId,
+                  parentNodeId: e.id,
+                  auth: _auth,
+                }),
+              { retries: 2 }
+            )
 
             if (fileNode) {
               fileNodeID = fileNode.id
@@ -522,7 +527,12 @@ exports.downloadMediaFiles = async ({
               })
             }
           } catch (e) {
-            // Ignore
+            if (process.env.NODE_ENV === `production`) {
+              // fail the build due to the error
+              throw new Error(e)
+            }
+            // delete the parent node to try and allow continued development with one less node
+            deleteNode(e)
           }
         }
       }
--- a/yarn.lock
+++ b/yarn.lock
@@ -3796,6 +3796,13 @@ async-limiter@~1.0.0:
   resolved "https://registry.yarnpkg.com/async-limiter/-/async-limiter-1.0.0.tgz#78faed8c3d074ab81f22b4e985d79e8738f720f8"
   integrity sha512-jp/uFnooOiO+L211eZOoSyzpOITMXx1rBITauYykG3BRYPu8h0UcxsPNB04RR5vo4Tyz3+ay17tR6JVf9qzYWg==
 
+async-retry@^1.2.3:
+  version "1.2.3"
+  resolved "https://registry.yarnpkg.com/async-retry/-/async-retry-1.2.3.tgz#a6521f338358d322b1a0012b79030c6f411d1ce0"
+  integrity sha512-tfDb02Th6CE6pJUF2gjW5ZVjsgwlucVXOEQMvEX9JgSJMs9gAX+Nz3xRuJBKuUYjTSYORqvDBORdAQ3LU59g7Q==
+  dependencies:
+    retry "0.12.0"
+
 async@1.5.2, async@^1.4.0, async@^1.5.2:
   version "1.5.2"
   resolved "https://registry.yarnpkg.com/async/-/async-1.5.2.tgz#ec6a61ae56480c0c3cb241c95618e20892f9672a"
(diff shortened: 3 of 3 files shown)


pieh:
It would be great if we support that directly in `createRemoteFileNode` so all consumers of that would benefit from it.

But this is a bit of hammer approach - usually files fail to download is because:
 - a) wordpress (or any other server) doesn't handle lot of concurrent requests
 - b) local machine / connection doesn't handle lot of concurrent requests very well
So retrying adds more fault tolerence, but doesn't exactly solve that if other params (like concurrency doesn't smartly adjust - i.e. lowering concurrency when 50x failures appear to happen).

Btw, we already have some retry mechanism in `createRemoteFileNode` - https://github.com/gatsbyjs/gatsby/blob/133ff1d6135e1fc6142821f927a4c805b64515d9/packages/gatsby-source-filesystem/src/create-remote-file-node.js#L148

But it's possible that this just doesn't work - some verification is needed there.

kristojorg:
@pieh interesting I didn’t realize retries were already happening. I think you’re right it’s about lowering concurrency when failures are happening. 

I solved this in userland because (a) I don’t know all the use cases of createRemoteFileNode and (b) I think different packages might want to have different ways to deal with failure. 

Honestly I think the main issue is that in the documentation for createRemoteFileNode it basically instructs source plugin builders to ignore errors, which is what they’ve been doing and it doesn’t work well.

sidharthachatterjee:
@kristojorg Thank you for your work on this. Like @pieh already mentioned, I think we need to account for two things here:

1. Handling this correctly in `createRemoteFileNode `
1. Ensuring a proper global concurrency rate 

@calcsam detailed a lot of this in https://github.com/gatsbyjs/gatsby/issues/10049

Let's close this in the mean time? Thank you again for your (very valuable) work and we look forward to more PRs from you! 💪 💜

kristojorg:
@sidharthachatterjee I agree there needs to be proper handling in `createRemoteFileNode`, but no matter the situation there, I think consumers of that api need to correctly handle errors. The current documentation suggests errors thrown from there should be ignored, but it seems we should instead stop the build process. Maybe this just requires a doc change?

clintonzhang:
I have run into this problem. Is there a way to workaround null being generated in the mean time?
Loading another placeholder image?

PR #13482, opened 2019-04-19
Fix styles to be sorted on static-entry.js

originerd:
## Description

There was a problem when a user tries to refresh without cache. In this case, style tags in HTML are messed up. To solve it, I added some code to sort style files by name.

- FYI (without this PR)
  1. Refresh without cache
    스크린샷 2019-04-19 오후 3 59 15

  1. Refresh with cache
    image

### DIFF
--- a/packages/gatsby/cache-dir/static-entry.js
+++ b/packages/gatsby/cache-dir/static-entry.js
@@ -22,6 +22,16 @@ const chunkMapping = JSON.parse(
   fs.readFileSync(`${process.cwd()}/public/chunk-map.json`, `utf-8`)
 )
 
+// Compare styles by names
+const compareStyleNames = ({ name: nameA }, { name: nameB }) => {
+  const [numberA, numberB] = [nameA, nameB].map(name =>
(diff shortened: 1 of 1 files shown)


pieh:
@originerd we are creating single css file since https://github.com/gatsbyjs/gatsby/pull/11800 / `gatsby@2.1.3`, so this should not happen (unless you modify webpack config).

There is also a lot more nuance to this problem in general than just sorting by chunk index - see https://github.com/gatsbyjs/gatsby/pull/8092 to see more complete implementation of fixes for the problem. This is pretty hard problem to solve, because this might fix initial css "ordering" but as you navigate and webpack runtime inject more chunks - it will always append them to "end" so they will end up in random position in <head> tag

sidharthachatterjee:
@originerd Thank you for attempting to fix this!

As @pieh mentioned the issue is a little more nuanced (and shouldn't be relevant after https://github.com/gatsbyjs/gatsby/pull/11800) so we're closing this out for the moment. 

Please know that we really appreciate your work on this nonetheless and we're looking forward to having you as a contributor! 💪 💜

PR #13489, opened 2019-04-19
fix(gatsby-remark-images): update to accept alt as title when no title present

### LINKED ISSUE #13448: gatsby-remark-images not show captions in md file
When i write image like this `my image`  in markdown file ,  
 `gatsby-remark-images` not show captions (because source code only detect img.title), 

(PS: I have enabled showCaptions in gatsby-config.js file)

so I have to write like `</img>` instead. 
But when I switched to use `gatsby-mdx` later, 
`</img>`   not worked (I watched output, src not replaced by processed images under static files, maybe a bug?) ,
 
so I have to write back `my image` again,
but captions not showed in this way, 

so can you change `const showCaptions = options.showCaptions && node.title` ([source code](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-remark-images/src/index.js#L244) ) to `const showCaptions = options.showCaptions && (node.title || node.alt)`  or add something else to detect if image is written in `my image` style and add captions to it?

jamessimone:
## Description

Some image sources only allow the setting of the alt image attribute when creating images in markdown. An example of this is Dropbox Paper. It would be nice to use the alt as the title for the image when no title is present and the showCaptions config value is enabled.

## Related Issues

Fixes #13448


sidharthachatterjee:
Went ahead and cleaned up some code I noticed after approving! Hope you don't mind @jamessimone 🙂

jamessimone:
@sidharthachatterjee not at all! Thank you for making this even better 😄

jamessimone:
@sidharthachatterjee I went to add a test for this, but I wanted to ask first if there is a way to only run the tests for gatsby-remark-images. When I ran `yarn` jest `packages/gatsby-remark-images/src/` I got the following error:  `Cannot find module 'gatsby-plugin-sharp' from 'index.js'`

I saw that plugin-sharp is listed as a peer dependency to the project but perhaps the tests can only be run from the root gatsby level due to that? Just wanted to ensure I wasn't doing something wrong there. Thank you to you and @pieh for jumping on this!

pieh:
@jamessimone Sid had exact same issue with jest. I'm not sure why both of you have this issue and I don't ...

But to move things forward - I added some basic tests for `showCaptions` to not block the PR because of those jest problems.

sidharthachatterjee:
Thank you so much @jamessimone and @pieh 🚀

PR #13491, opened 2019-04-19
fix(showcase): creator detail 'worked on' site list multiple lines

riencoertjens:


pieh:
Hey @riencoertjens, could you share before/after screenshot for the changes in `template-creator-details.js` ?

riencoertjens:
before: 
Screenshot 2019-04-19 at 18 59 29

after:
Screenshot 2019-04-19 at 18 58 53

docs/sites.yml
```diff
@@ -5383,3 +5383,87 @@
...
+  featured: false
+- title: nicdougall.com
+  url: https://nicdougall.netlify.com/
+  main_url: https://nicdougall.netlify.com/
+  description: >
+    Athlete website with Netlify CMS for blog content.
+  categories:
+    - Athlete
+    - Blog
+    - Netlify CMS
```
pieh: I think we could skip specifying "Netlify CMS", given that source is not provided, so fact that Nelify CMS is used wouldn't be visible to showcase visitors?

docs/sites.yml
```diff
@@ -5383,3 +5383,87 @@
...
+  featured: false
+- title: het Groeiatelier
+  url: https://www.hetgroeiatelier.be/
+  main_url: https://www.hetgroeiatelier.be/
+  description: >
+    Workspace for talent development and logopedics. onepage site with basic info and small calendar CMS.
+  categories:
+    - logopedics
+    - One Page
+    - Netlify CMS
```
pieh: Similar here

docs/sites.yml
```diff
@@ -5383,3 +5383,87 @@
   built_by: Nikodem Deja
   built_by_url: https://nikodemdeja.pl
   featured: false
+- title: manuvel.be
+  url: https://www.manuvel.be
+  main_url: https://www.manuvel.be
+  description: >
+    Cycling themed café coming this april in Sint Niklaas, Belgium. Has funky css-grid trickery!
+  categories:
+    - Café
```
pieh: Could we perhaps use Coffee here too? I know it's not exact, but given that Coffee is added in other entry we would end up with Café and Coffee categories each with single entry

docs/sites.yml
```diff
@@ -5383,3 +5383,87 @@
...
+    - Blog
+    - Netlify CMS
+  built_by: Web Hart
+  built_by_url: https://www.web-hart.com
+  featured: false
+- title: het Groeiatelier
+  url: https://www.hetgroeiatelier.be/
+  main_url: https://www.hetgroeiatelier.be/
+  description: >
+    Workspace for talent development and logopedics. onepage site with basic info and small calendar CMS.
```
pieh:
```suggestion
    Workspace for talent development and logopedics. One page site with basic info and small calendar CMS.
```

docs/sites.yml
```diff
@@ -5383,3 +5383,87 @@
...
+  categories:
+    - logopedics
```
pieh: This category seems very specific and likely to be one-off - ideally we could use something more generic that would apply to more sites to avoid categories that would hold just single entries

riencoertjens:
those comments all make sense, it was obviously a long day... 
I made those changes, added some source_urls and changed my mind about the my name 😅

riencoertjens:
@pieh have you had a chance to look at this?

pieh:
Sorry about delay @riencoertjens and thanks for patience.

Issue #11930, opened 2019-02-20
deleteNode doesn't respect node type owner


pieh:
We have node type owner checks when creating new nodes - https://github.com/gatsbyjs/gatsby/blob/04a50965af5d320097020a6050a60ed1b2afa8c9/packages/gatsby/src/redux/actions.js#L545-L547

We don't use same checks for deleting nodes (in https://github.com/gatsbyjs/gatsby/blob/04a50965af5d320097020a6050a60ed1b2afa8c9/packages/gatsby/src/redux/actions.js#L317 ), but we should validate those too.

PR #13492, opened 2019-04-19
fix(gatsby): respect node type owner in deleteNode

frinyvonnick:
## Description

It updates deleteNode action so it respects node type owner

## Related Issues

#11930

### DIFF
--- a/packages/gatsby/src/redux/actions.js
+++ b/packages/gatsby/src/redux/actions.js
@@ -403,6 +403,27 @@ actions.deleteNode = (options: any, plugin: Plugin, args: any) => {
   // Always get node from the store, as the node we get as an arg
   // might already have been deleted.
   const node = getNode(id)
+  const pluginName = plugin.name
+
+  if (typeOwners[node.internal.type] !== pluginName)
+    throw new Error(stripIndent`
+        The plugin "${pluginName}" deleted a node of a type owned by another plugin.
+
+        The node type "${node.internal.type}" is owned by "${
+      typeOwners[node.internal.type]
+    }".
+
+        If you copy and pasted code from elsewhere, you'll need to pick a new type name
+        for your new node(s).
+
+        The node object passed to "deleteNode":
+
+        ${JSON.stringify(node, null, 4)}
+
+        The plugin deleting the node:
+
+        ${JSON.stringify(plugin, null, 4)}
+      `)
 
   const createDeleteAction = node => {
     return {
--- a/packages/gatsby/src/redux/actions.js
+++ b/packages/gatsby/src/redux/actions.js
@@ -414,9 +414,6 @@ actions.deleteNode = (options: any, plugin: Plugin, args: any) => {
         typeOwners[node.internal.type]
       }".
 
-          If you copy and pasted code from elsewhere, you'll need to pick a new type name
-          for your new node(s).
-
           The node object passed to "deleteNode":
 
           ${JSON.stringify(node, null, 4)}
--- a/packages/gatsby/src/redux/actions.js
+++ b/packages/gatsby/src/redux/actions.js
@@ -405,7 +405,7 @@ actions.deleteNode = (options: any, plugin: Plugin, args: any) => {
   const node = getNode(id)
   const pluginName = plugin.name
 
-  if (typeOwners[node.internal.type] !== pluginName)
+  if (node && typeOwners[node.internal.type] !== pluginName)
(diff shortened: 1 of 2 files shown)


packages/gatsby/src/redux/actions.js
```diff
@@ -403,6 +403,27 @@ actions.deleteNode = (options: any, plugin: Plugin, args: any) => {
   // Always get node from the store, as the node we get as an arg
   // might already have been deleted.
   const node = getNode(id)
+  const pluginName = plugin.name
```
sidharthachatterjee: Needs to be wrapped with `if (plugin) {` like other instances in the file

frinyvonnick:
Thank you for your review @sidharthachatterjee 👍 I added a little test 😄

packages/gatsby/src/redux/actions.js
```diff
@@ -403,6 +403,29 @@ actions.deleteNode = (options: any, plugin: Plugin, args: any) => {
...
+    if (node && typeOwners[node.internal.type] !== pluginName)
+      throw new Error(stripIndent`
+          The plugin "${pluginName}" deleted a node of a type owned by another plugin.
+
+          The node type "${node.internal.type}" is owned by "${
+        typeOwners[node.internal.type]
+      }".
+
+          If you copy and pasted code from elsewhere, you'll need to pick a new type name
+          for your new node(s).
```
pieh: I know this was copied from `createNode`, but (to me) this note about picking different type name doesn't really work for deleting nodes. Maybe let's just remove lines 417-418

frinyvonnick: Yes seems logic, sorry about that 😊

frinyvonnick:
Thank you for your review @pieh 👍 !

jessepinho:
This is a breaking change. Shouldn't it have bumped the major version number?

sidharthachatterjee:
@jessepinho You're absolutely right. This is a breaking change and it seems that we missed that at the time. 

Our sincere apologies for this. We completely understand how frustrating it can be for breaking changes to creep in in minor or patch versions. 

Can you elaborate on the issue you're facing with respect to this? We'd love to help!

jessepinho:
@sidharthachatterjee thanks for understanding! <3 

The issue I'm facing is that I used `onCreateNode` in my `gatsby-node.js` to delete nodes that I deemed invalid — e.g., Contentful articles that had no `slug` field filled in.

There are two problems with simply filtering these nodes out when using GraphQL queries:
1. I have to remember to include that filter in every GraphQL query across the site.
2. I literally _can't_ filter via GraphQL in certain queries. For example, when querying for _author_ entries in Contentful, and then adding the `article` property to the query to get a list of articles that reference this author, there's no way to filter the `article` list for only those that have a `slug` filled in. I'd have to do it client-side after receiving the results of the query, which results in both more client-side work and a larger dataset returned by the query — both of which hurt performance.

In short, I'd like to have only nodes that I consider "valid" to be included in Gatsby's internal node database.

(I am currently considering a PR to `gatsby-source-contentful` to allow a `filterNodes` option or something, which would be a callback that could have a similar effect to what I'm trying to do with `onCreateNode`. But that's a bit of extra work when I already had a working solution. What's the cache issue that this PR fixes?)

sidharthachatterjee:
@jessepinho Ah, that makes sense. Your local plugin should be able to delete anything. I've opened an issue for this in https://github.com/gatsbyjs/gatsby/issues/14870! 

Thank you for all the details.

sidharthachatterjee:
After speaking to @pieh we've come to the conclusion that local plugins _shouldn't_ actually be able to delete anything. This can cause all sorts of issues because of broken references and cause other plugins that _own_ these nodes to break. 

@pieh will elaborate

pieh:
Deleting nodes that you don't own, is asking for trouble. In your example, if that article you deem invalid, would be linked to from some other content - deleting just the article would cause errors. I understand that this is frustrating, but adding this check wasn't just to make people's life harder.

There is undocumented workaround to "impersonate" owner plugin which will allow you to still do this, even if this is something that probably will never be recommended:
```
actions.deleteNode({ node }, { name: `gatsby-source-contentful` })
```

jessepinho:
OK, hm... Would the same problem happen if I were to do my `filterNodes` PR idea above? i.e., make it possible to filter nodes via a callback _before_ they're created by the plugin?

PR #13497, opened 2019-04-19
refactor(gatsby-cli): update reporter to support different implementations

wardpeet:
## Description
Basically this is just a refactor to allow us to fallback to yurnalist when people are on node 6. We can also fallback on CI as Gatsby Cloud would need this as we don't write to stderr (I think) when using ink.

In this PR it's only using yurnalist to make sure we don't break anything and keep this PR small. Ink will build from this one.

## Related PRs
#13089

### DIFF
--- a/packages/gatsby-cli/src/reporter/index.js
+++ b/packages/gatsby-cli/src/reporter/index.js
@@ -33,8 +33,9 @@ const reporter = {
    * @param {boolean} [isNoColor=false]
    */
   setNoColor(isNoColor = false) {
+    reporterInstance.setColors(isNoColor)
+
     if (isNoColor) {
-      reporterInstance.setColors()
       errorFormatter.withoutColors()
     }
   },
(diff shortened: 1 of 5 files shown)


sidharthachatterjee:
Published in `gatsby-cli@2.5.12`

packages/gatsby-cli/src/reporter/index.js
```diff
@@ -1,50 +1,47 @@
...
 const { stripIndent } = require(`common-tags`)
-const convertHrtime = require(`convert-hrtime`)
+const chalk = require(`chalk`)
+const { trackError } = require(`gatsby-telemetry`)
 const tracer = require(`opentracing`).globalTracer()
 const { getErrorFormatter } = require(`./errors`)
-const { trackError } = require(`gatsby-telemetry`)
-
-const VERBOSE = process.env.gatsby_log_level === `verbose`
+const reporterInstance = require(`./reporters/yurnalist`)
```
muescha: this is thight binding to yurnalist? you change this later to a setting?

sidharthachatterjee: Yup, incrementally!

PR #13500, opened 2019-04-19
docs(gatsby-cypress): add missing deps + more info + examples

jlengstorf:
As part of a [livestream](https://youtu.be/IADSsClWVtA) with @marcysutton, we hit some snags with `gatsby-cypress`. These updates cover what we ran into and corrects the docs accordingly.

### DIFF
--- a/packages/gatsby-cypress/README.md
+++ b/packages/gatsby-cypress/README.md
@@ -30,7 +30,7 @@ Once imported, the following additional commands are available:
   ```js
   cy.getTestElement("btn-to-test").click()
   ```
-  
+
   > **NOTE:** It’s recommended not to use test IDs. Instead, consider using [`cypress-testing-library`](https://github.com/kentcdodds/cypress-testing-library) and relying on `getByText` instead.
 
 - `cy.waitForRouteChange()`: Waits for Gatsby to finish the route change, in
--- a/packages/gatsby-cypress/README.md
+++ b/packages/gatsby-cypress/README.md
@@ -1,6 +1,10 @@
 # gatsby-cypress
 
-This package provides additional [Cypress](https://cypress.io/) commands for [testing Gatsby websites](/docs/end-to-end-testing/). To use these commands, first install the necessary packages:
+This package provides additional [Cypress](https://cypress.io/) commands for [testing Gatsby websites](/docs/end-to-end-testing/).
+
+> **NOTE:** This package is _not_ required to use Gatsby and Cypress together. It only exists to add extra commands for convenience.
+
+To use these commands, first install the necessary packages:
 
 ```bash
 npm install cypress gatsby-cypress start-server-and-test --save-dev
@@ -26,6 +30,8 @@ Once imported, the following additional commands are available:
   ```js
   cy.getTestElement("btn-to-test").click()
   ```
+  
+  > **NOTE:** It’s recommended not to use test IDs. Instead, consider using [`cypress-testing-library`](https://github.com/kentcdodds/cypress-testing-library) and relying on `getByText` instead.
 
 - `cy.waitForRouteChange()`: Waits for Gatsby to finish the route change, in
   order to ensure event handlers are properly setup. Example:
--- a/packages/gatsby-cypress/README.md
+++ b/packages/gatsby-cypress/README.md
@@ -18,13 +18,13 @@ Once imported, the following additional commands are available:
 
 - `cy.getTestElement(selector)`: Selects elements where the `data-testid`
   attribute matches the value of `selector`. Example:
-  
+
   ```jsx
   <button data-testid="btn-to-test">click me</button>
   ```
 
   ```js
-  cy.getTestElement('btn-to-test').click()
+  cy.getTestElement("btn-to-test").click()
   ```
 
 - `cy.waitForRouteChange()`: Waits for Gatsby to finish the route change, in
@@ -62,7 +62,7 @@ Add tests by creating a spec file. We recommend starting with a `cypress/integra
 context("Homepage", () => {
   beforeEach(() => {
     cy.visit(`http://localhost:8000`)
-    
+
     // Wait for React to finish mounting all components before testing.
     cy.wait(100)
   })
--- a/packages/gatsby-cypress/README.md
+++ b/packages/gatsby-cypress/README.md
@@ -3,7 +3,7 @@
 This package provides additional [Cypress](https://cypress.io/) commands for [testing Gatsby websites](/docs/end-to-end-testing/). To use these commands, first install the necessary packages:
 
 ```bash
-npm install cypress gatsby-cypress starter-server-and-test --save-dev
+npm install cypress gatsby-cypress start-server-and-test --save-dev
 ```
 
 ## Add custom Gatsby testing commands
--- a/packages/gatsby-cypress/README.md
+++ b/packages/gatsby-cypress/README.md
@@ -20,7 +20,17 @@ import "gatsby-cypress/commands"
 
 Once imported, the following additional commands are available:
 
-- `cy.getTestElement(selector)`: Selects elements where the `data-testid`
+- `cy.waitForRouteChange()`: Waits for Gatsby to finish the route change, in
+  order to ensure event handlers are properly setup. Example:
+
+  ```js
+  // after navigating to another page via a link
+  cy.waitForRouteChange()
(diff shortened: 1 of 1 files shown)


packages/gatsby-cypress/README.md
```diff
@@ -3,29 +3,28 @@
 This package provides additional [Cypress](https://cypress.io/) commands for [testing Gatsby websites](/docs/end-to-end-testing/). To use these commands, first install the necessary packages:
 
 ```bash
-npm install cypress gatsby-cypress --save-dev
+npm install cypress gatsby-cypress starter-server-and-test --save-dev
```
DSchau:
```suggestion
npm install cypress gatsby-cypress start-server-and-test --save-dev
```

packages/gatsby-cypress/README.md
```diff
@@ -3,29 +3,28 @@
...
-"test": "CYPRESS_SUPPORT=y npm run build && start-server-and-test develop http://localhost:8000 cypress open"
-```
-
 Once imported, the following additional commands are available:
 
 - `cy.getTestElement(selector)`: Selects elements where the `data-testid`
   attribute matches the value of `selector`. Example:
+  
+  ```jsx
+  <button data-testid="btn-to-test">click me</button>
```
DSchau: Nit: but @kentcdodds has expressed some opinions in preferring `getByText` rather than using data-testid. Doesn't need to block this PR though!

kentcdodds: Yes! Please do avoid test IDs! This one should be selected with `getByText(/click me/i)`

packages/gatsby-cypress/README.md
```diff
@@ -37,3 +36,40 @@ Once imported, the following additional commands are available:
...
+    ... other scripts ...
++   "cy:open": "cypress open",
++   "test:dev": "CYPRESS_SUPPORT=y start-server-and-test develop http://localhost:8000 cy:open",
++   "test": "CYPRESS_SUPPORT=y npm run build && start-server-and-test serve http://localhost:9000 cy:open"
+  }
+```
+
+> **NOTE:** `test:dev` runs the site in development mode, which allows you to quickly fix and retest any issues. `test` is better suited for build systems like Circle CI, Travis CI, etc.
+
+## Writing Cypress tests for Gatsby pages
```
LekoArts: Why not link the already existing doc page here?

LekoArts:
I haven't watched the livestream completely but I had the impression that you both had some misconceptions about what this package is. It’s not mandatory for all Gatsby projects and also only adds commands, and not does setup Cypress for you. Did you came to these conclusions later in the stream? If not probably the actual end-to-end doc should be linked early.

packages/gatsby-cypress/README.md
jlengstorf: Should we remove the `getTestElement` API from `gatsby-cypress`? I was just trying to demonstrate what the exported commands do.

packages/gatsby-cypress/README.md
jlengstorf: @LekoArts what's the already existing page? I was trying to add a minimal example so that someone on this README can get tests up and running without needing to look elsewhere.

packages/gatsby-cypress/README.md
DSchau: Worth keeping it around--there are use cases where it's needed (and we're using them all over the place in our e2e tests 😬), but...

We _should_ create a new API method `getByText` or something similar and expose it and recommend it! However--outside the scope of this PR!

jlengstorf:
@LekoArts it would be great to get a better explanation at the top of this README that explains why this package exists. Both @marcysutton and I were under the impression that this was sort of like `gatsby-plugin-emotion`, which handles the stuff we need to do to make Gatsby work properly with a given tool.

packages/gatsby-cypress/README.md
LekoArts: We can just use [cypress-testing-library](https://github.com/kentcdodds/cypress-testing-library) like I wrote here 👍 
https://www.gatsbyjs.org/docs/end-to-end-testing/

packages/gatsby-cypress/README.md
LekoArts: https://www.gatsbyjs.org/docs/end-to-end-testing/

packages/gatsby-cypress/README.md
DSchau: Great - thanks Lennart!

jlengstorf: So in terms of action steps:

1. Mark this method as not recommended
2. Point to https://github.com/kentcdodds/cypress-testing-library for more powerful commands?

Should I remove the example, too? Or is just the notice enough?

packages/gatsby-cypress/README.md
LekoArts: I didn't add an example back when I created this page as the Cypress docs are pretty good. However, I'll need to edit the CI docs on this page to switch to `build` and `serve` on CI (could do that next week)

LekoArts:
Yeah, then there is definitely room for improvement. This only provides commands from which I mainly use the `waitForRouteChange` (like [here](https://github.com/LekoArts/gatsby-starter-portfolio-emma/blob/master/cypress/e2e/home.js)) because Cypress sometimes has problems not "seeing" things after the `click()` with Gatsby.

Hence my note that a comment like "These are only commands. If you want to use/install Cypress head over to the docs" would be great at the top.

packages/gatsby-cypress/README.md
pieh: I honestly think we should not even mention `getTestElement` in `gatsby-cypress` docs - this is really messy helper that we use in our tests (that probably can be rewritten to use `getByText`).

Only useful thing for users in `gatsby-cypress` really is `cy. waitForRouteChange()`:
In your livestream yesterday you needed to add `cy.wait(500)` when testing `gatsby develop` - `cy.waitForRouteChange()` is more reliable equiavalent of that.

The problem was happening because `cy.visit('/')` resolves when "DOM ready" event is emitted - which most often will happen before react application mounts. `cy.waitForRouteChange()` ensures to wait for react application to mount (on initial visit) and page to actually change on navigation before continuing with tests.

jlengstorf: @pieh do you have sample code that shows this working? I ended up doing this:

```js
  beforeEach(() => {
    cy.visit(`http://localhost:8000`)
     // Wait for React to finish mounting all components before testing.
    cy.wait(100)
  })
```

This might be specific to `react-helmet`, but when I use `cy.waitForRouteChange()` the Helmet stuff hasn't happened yet. If I use `cy.wait(100)`, it has.

pieh: It might be specific to `react-helmet`, it does use `requestAnimationFrame` to update `head`/DOM under the hood. For the most part I was referring to part when `cypress-axe` didn't pick anything up in `gatsby develop` (no landmarks rules, etc).

We do use `waitForRouteChange` for all of our e2e tests (for example in https://github.com/gatsbyjs/gatsby/blob/a4f03cfe0d76a6feb7be76ed155a16a225827941/e2e-tests/development-runtime/cypress/integration/navigation/linking.js#L2-L4 ) for both development and production runtime tests and it works there (and remove flakiness of those tests - we were using `cy.wait` but it would intermittently fail in cases when wait time was not enough in CircleCI). But we don't have any tests that would check anything related to `react-helmet` (as it's not part of core `gatsby` package), so probably that's why we didn't hit the issue.

jlengstorf:
Is there anything else to change here? Or is this good to go?

KyleAMathews:
@jlengstorf looks like just replacing the ids with getByText

jlengstorf:
@KyleAMathews done. I left the docs in, but removed all usage that's not in the doc itself + added a big, bold recommendation not to use `getTestElement` anymore.

packages/gatsby-cypress/README.md
pieh: Let's keep example here for now and once https://www.gatsbyjs.org/docs/end-to-end-testing/ is updated with example we can swap this section with "See https://www.gatsbyjs.org/docs/end-to-end-testing/ for example usage" (or something along this lines)

LekoArts: Yeah, I'll work on this today a bit and send in a PR when finished 👍 But shouldn't be a blocker

PR #13502, opened 2019-04-19
fix(gatsby-plugin-offline): Change navigation handler logic

davidbailey00:
The more "stable" part of #11965

Changes:

- Remove page path whitelist
- Whenever a path has finished prefetching, store the name of its resource files in the IDB
- Replace `resetWhitelist` API with `clearPathResources`, which clears the page resource dependencies
- When loading a page from SW, check the correct app bundle is present and all the resources for the page
  - If not, or if the resource filename list isn't found, serve from the network

Rationale: using the path whitelist turned out to be buggy, especially during upgrades when files could be in inconsistent state. Many of the initial bugs were fixed, but some of the fixes are more workarounds than proper solutions which can still result in temporary white screen flashes.

Therefore these changes should be more reliable, since by checking the main bundle is present as well as the required resource files, we can be more sure that the latest essential files are cached rather than just looking at a list.


m-allanson:
@davidbailey00 I'm going to close this for now. Let's come back and re-open it once you're back in (Gatsby) action!

davidbailey00:
<p class="btn">test</p>

packages/gatsby-plugin-offline/src/sw-append.js
```diff
@@ -1,101 +1,47 @@
 /* global importScripts, workbox, idbKeyval */
 
 importScripts(`idb-keyval-iife.min.js`)
-const WHITELIST_KEY = `custom-navigation-whitelist`
+workbox.setConfig({ debug: false })
```
davidbailey00: temp, can be undone

packages/gatsby-cli/src/reporter/reporters/ink/components/develop.js
```diff
@@ -44,7 +44,7 @@ class Develop extends Component {
     })
   }
 
-  componentWillUpdate(nextProps) {
+  UNSAFE_componentWillUpdate(nextProps) {
```
sidharthachatterjee: I think we should defer this change to after we move to React 17 

From the documentation at https://reactjs.org/docs/react-component.html

> That name will continue to work until version 17

davidbailey00:
ayy

image

PR #13504, opened 2019-04-19
chore(docs): Update using-graphql-playground

frankievalentine:
Added missing step to require dotenv package, misc formatting changes

### DIFF
--- a/docs/docs/using-graphql-playground.md
+++ b/docs/docs/using-graphql-playground.md
@@ -12,12 +12,14 @@ GraphQL Playground is a way for you to interact with the data your sources and p
 
 ## Accessing the Playground
 
-To access GraphQL Playground, first require the dotenv package in your `gatsby-config.js` file (see [here](https://www.gatsbyjs.org/docs/environment-variables/)), add a development environmental variable file, and finally add `GATSBY_GRAPHQL_IDE` to your `develop` script in your `package.json`, like this:
+To access GraphQL Playground add `GATSBY_GRAPHQL_IDE` to your `develop` script in your `package.json`, like this:
 
 ```
 "develop": "GATSBY_GRAPHQL_IDE=playground gatsby develop",
 ```
 
-You can then access it, when the development server is running on `https://localhost:8000/___graphql`
+Use `npm run develop` instead of `gatsby develop` and access it when the development server is running on `https://localhost:8000/___graphql`
+
+To still be able to use `gatsby develop` you would require the dotenv package to your gatsby-config.js file and add an environment variable file, typically called `.env.development`. Finally, add the same script above to `.env.development`.
 
 ![An image pointing out where to find the GraphQl schema](images/playground-schema.png)
--- a/docs/docs/using-graphql-playground.md
+++ b/docs/docs/using-graphql-playground.md
@@ -4,15 +4,15 @@ title: Using the GraphQL Playground
 
 ## Intro
 
-In this page we want to introduce you to an alternative to the current IDE for your GraphQL queries: The [Prisma Playground](https://github.com/prisma/graphql-playground).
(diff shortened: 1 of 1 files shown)


pieh:
Hi @frankievalentine! Thanks for PR!

Few notes:
You don't need `dotenv` - running `GATSBY_GRAPHQL_IDE=playground gatsby develop` is enough on most most platforms (not counting Windows), but if you apply it to your scripts in package.json (as currently documented) and run `npm/yarn run develop` - it will work everywhere. 

It's either `dotenv` approach or the steps documented currently. and this changes is making it a bit more confusing because it seems like you need to do all the steps

frankievalentine:
> Hi @frankievalentine! Thanks for PR!
> 
> Few notes:
> You don't need `dotenv` - running `GATSBY_GRAPHQL_IDE=playground gatsby develop` is enough on most most platforms (not counting Windows), but if you apply it to your scripts in package.json (as currently documented) and run `npm/yarn run develop` - it will work everywhere.
> 
> It's either `dotenv` approach or the steps documented currently. and this changes is making it a bit more confusing because it seems like you need to do all the steps

Hi @pieh 
Right so I guess that is where there is some confusion in this doc. Typically, in development, it is known to utilize `gatsby develop` instead of `npm/yarn run develop` so I guess that's where the documentation doesn't state specifically to use 'npm/yarn run develop' instead of `gatsby develop` to utilize Playground.

Would you advise I edit the PR to include both .env methods?

pieh:
> Would you advise I edit the PR to include both .env methods?

I think, this way we would end up having to duplicate those methods in every docs page that uses environment variables. Ideally we could just say to use `GATSBY_GRAPHQL_IDE=playground` variable and link to section in https://www.gatsbyjs.org/docs/environment-variables/ that document those 2 methods so we don't need to duplicate it everywhere

frankievalentine:
> > Would you advise I edit the PR to include both .env methods?
> 
> I think, this way we would end up having to duplicate those methods in every docs page that uses environment variables. Ideally we could just say to use `GATSBY_GRAPHQL_IDE=playground` variable and link to section in https://www.gatsbyjs.org/docs/environment-variables/ that document those 2 methods so we don't need to duplicate it everywhere

In just the case of using Playground, I don't think this causes duplication. How I see it currently, yes, the easiest method is to add the script to your package.json file and use `npm/yarn run develop`. Which, I believe this doc for this PR should include that you use `npm/yarn run develop` versus `gatsby develop`.

But, if you still wish to use `gatsby develop` you would add the 'dotenv' package to your gatsby-config.js file and add the environment variable file.

I think there is a significant difference, not enough to change the process as a whole, but one that other developers should know about.

frankievalentine:
I have updated the README to include the correct instructions for utilizing GraphQL Playground with `npm run develop` and `gatsby develop`.

docs/docs/using-graphql-playground.md
```diff
@@ -4,20 +4,22 @@ title: Using the GraphQL Playground
...
-## What is The Prisma Playground?
+## What is Prisma's GraphQL Playground?
 
-The Playground is a way for you to interact with the data your sources and plugins add as schemas. You will be interacting with this data a lot and the Playground will help you greatly with exploring this data.
+GraphQL Playground is a way for you to interact with the data your sources and plugins add as schemas. You will be interacting with this data a lot and the Playground will help you greatly with exploring this data.
 
 ## Accessing the Playground
 
-As this is an experimental feature, you need to access it by adding an environmental variable, you need to add `GATSBY_GRAPHQL_IDE` to your `develop` script in your `package.json`, like this:
+To access GraphQL Playground add `GATSBY_GRAPHQL_IDE` to your `develop` script in your `package.json`, like this:
```
KyleAMathews: Could you add back the note about this being an experimental feature? That's important as we may remove it in a future update so people shouldn't expect this to stay around.

frankievalentine: Sent a new commit with the note included again!

PR #13510, opened 2019-04-20
feat(docs): Grouping stubs on Stub List

### LINKED ISSUE #13474: [docs] Improve stub-list to group similar stubs
## Summary

Looking at the current stub list, some of the stubs can be confusing on what they mean by looking at the title for them (for instance, some of the Using Gatsby Professionally stubs)

Screen Shot 2019-04-18 at 8 11 45 PM

One way to clean this up could be group based upon a particular level in the sidebars for each stub. As well, we could make some sort of breadcrumbs so each stub shows the point in the docs tree

lannonbr:
## Description

Screenshot_2019-04-19 Stub List GatsbyJS(2)

Rather than having a strict list full of links, I am chunking them based upon the `parentTitle` field in each stub. If said field is undefined, it will go into a group titled "Other"

## Related Issues

Closes #13474

### DIFF
--- a/www/src/pages/contributing/stub-list.js
+++ b/www/src/pages/contributing/stub-list.js
@@ -10,6 +10,7 @@ import {
 import Container from "../../components/container"
 import DocsearchContent from "../../components/docsearch-content"
 import FooterLinks from "../../components/shared/footer-links"
+import { fontSizes, space } from "../../utils/presets"
 
 const findStubs = pages =>
   pages.filter(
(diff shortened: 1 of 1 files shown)


lannonbr:
@fk Could you look at this and give some critique on how the layout is and if there is things that we need to change?

www/src/pages/contributing/stub-list.js
```diff
@@ -28,6 +28,26 @@ class StubListRoute extends React.Component {
       flatten([...itemListContributing.items, ...itemListDocs.items])
     )
 
+    let groupedStubs = {
+      Other: [],
+    }
+
+    stubs.forEach(stub => {
```
lannonbr: Betting there's likely a better way to group these into a map based upon the parentTitle field.

www/src/pages/contributing/stub-list.js
```diff
@@ -48,13 +68,24 @@ class StubListRoute extends React.Component {
...
-              {stubs.map(stub => (
-                <li key={stub.title}>
-                  <Link to={stub.link}>{stub.title.slice(0, -1)}</Link>
-                </li>
-              ))}
-            </ul>
+            <section data-testid="list-of-stubs">
+              {sortedCategories.map(category => {
+                let categoryTitle =
+                  category.slice(-1) === `*` ? category.slice(0, -1) : category
```
lannonbr: categories themselves may be stubs so remove the asterisk if it is there. Is this alright or should we clean this up somewhere else?

lannonbr:
As well, I am fully up to change how we group the various stubs. Willing to hear other ideas if it makes it even cleaner.

lannonbr:
@fk I took your suggestions for the styling and decided on @pieh's idea for the other category to be renamed "Top Level Documentation Pages". I decided to put it as the first item and then sort the rest alphabetically.

Here's the new state of how it looks:

Stub List

In terms of some of the stubs not actually appearing, I am guessing it has to do with the logic in the findStubs method at the top of this file, but I think we should do this in a separate PR just so this one doesn't do more than it was originally meant to do.

Finally, I am unsure on having a link for if the category title itself is a stub given once we fix the missing stubs, they will all be listed at least once in the page.

Otherwise, I think we are good to go unless there is any remaining things to be resolved with this PR that either of you want to look into.

www/src/pages/contributing/stub-list.js
```diff
@@ -28,6 +28,26 @@ class StubListRoute extends React.Component {
       flatten([...itemListContributing.items, ...itemListDocs.items])
     )
 
+    let groupedStubs = {
+      Other: [],
+    }
+
+    stubs.forEach(stub => {
+      if (stub.parentTitle === undefined) {
+        groupedStubs[`Other`].push(stub)
```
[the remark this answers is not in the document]
lannonbr: Done

fk:
> decided on @pieh's idea for the other category to be renamed "Top Level Documentation Pages". I decided to put it as the first item and then sort the rest alphabetically.

💯— I like this much more than "Other", and listing those first makes total sense!

👍 to everything else, and thanks so much for opening #13590! 🙏

PR #13538, opened 2019-04-22
tests: add node 12 to tests

moonmeister:
## Description

Modified the Circle CI config in preparation for the release for Node 12 on April 23rd. 

As the next node LTS we should start testing against this version to support it. IMO this test not be required to test for now but can be enabled to give us info on what's passing and not.  At the latest, we can require these tests to pass on 2019-10-22 when it's made an active LTS. In the mean, time we can do any work needed to get it passing.

We'll need to wait to merge this till CircleCI also supports node 12. Hopefully it won't take them too long.

https://nodejs.org/en/about/releases/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -137,7 +137,7 @@ jobs:
   unit_tests_node12:
     executor:
       name: node
-      image: "12"
+      image: "latest"
     <<: *test_template
 
   unit_tests_www:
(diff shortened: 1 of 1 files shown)


KyleAMathews:
Any word when circleci will add support? If it's uncertain, maybe just close this temporarily until it's ready to go in (when we can reopen).

moonmeister:
@KyleAMathews I can't be be sure but based on their docker hub and other various info I could glean I have a feeling it'll be out pretty quick.

PR #13540, opened 2019-04-22
docs: fix ts-jest testRegex

nevik:
## Description

The previously used testRegex for TS tests with Jest did not work as (I assume) was intended: files inside a `__tests__` folder would only be recognized if they had double file extensions, such as `__tests__/file.ts.ts`.

In some preliminary tests, the new regex in this change should work as intended, catching all ts(x)/js(x) files inside a `__tests__` folder (or any subfolders of it), and all files with extensions of .spec/test.ts(x)/js(x).

For easier validation, I've put this into a little playground where you can edit the Regex and add more test cases on the fly: https://codesandbox.io/embed/3v2z573qrm


nevik:
Thanks for checking and merging this so quickly! :sparkle:

PR #13545, opened 2019-04-22
Fixed broken URL in README.md

polishedwp:
## Description
The demo URL was pointing to `gatsby.org` instead of `netlify.com`.


LekoArts:
Hi, thanks for PR!
I can’t merge this as it’s the right link — we just need to set up the DNS to resolve to that :) Eventually this link will work and I'll see that we get that fixed ASAP

PR #13547, opened 2019-04-22
feat(gatsby-source-filesystem): add ability to specify file destination path

artbycrunk:
## Description

Currently, the `gatsby-source-filesystem` plugin doesn't allow users to specify a folder to output the images found in markdown frontmatter and will `default to public/static`.
```
frontmatter {
    header_image {
        path {
            publicURL
        }
    }
}
```
So from the above query, the source header_image will be copied to /public/static, and a public url will be returned back to the new files location, this is not flexible for cases where a user wants to better organize their static folder.

This PR add a new plugin option `fileDestinationDir` that allows the user to customize the folder structure of the output files. ( Paths need to be relative to and within the public directory )
```
gatsby-config.js
{
     resolve: 'gatsby-source-filesystem',
     options: {
                name: 'posts',
                path: 'content',
                fileDestinationDir: 'static/images/source',
            }
}
```

## Related Issues
Looking to add this functionality to `gatsby-plugin-sharp` as well.

### DIFF
--- a/packages/gatsby-source-filesystem/src/extend-file-node.js
+++ b/packages/gatsby-source-filesystem/src/extend-file-node.js
@@ -2,9 +2,32 @@ const { GraphQLString } = require(`gatsby/graphql`)
 const fs = require(`fs-extra`)
 const path = require(`path`)
 
-module.exports = ({ type, getNodeAndSavePathDependency, pathPrefix = `` }) => {
+module.exports = ({ type, getNodeAndSavePathDependency, pathPrefix = `` }, pluginOptions) => {
   if (type.name !== `File`) {
     return {}
+  }
+
+  const { fileDestinationDir } = pluginOptions
+
+  const newPublicPath = (fileName, fileDestinationDir) => {
+    if (fileDestinationDir) {
+      return path.posix.join(process.cwd(), `public`, fileDestinationDir, fileName)
+    }
+    return path.posix.join(process.cwd(), `public`, `static`, fileName)
+  }
+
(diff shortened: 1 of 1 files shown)


pieh:
> this is not flexible for cases where a user wants to better organize their static folder.

Can you expand on this? Why user would need to organize it / what's the value of that option?

artbycrunk:
> > this is not flexible for cases where a user wants to better organize their static folder.
> 
> Can you expand on this? Why user would need to organize it / what's the value of that option?
Well, there is a long run value in being organized, currently, the files that go into the static folder don't have any real structure, its not easy to know what are the source images files from the other static content. It's an added option to separate the generated content from other static content.

pieh:
> Well, there is a long run value in being organized, currently, the files that go into the static folder don't have any real structure, its not easy to know what are the source images files from the other static content. It's an added option to separate the generated content from other static content.

But why do user need to know that? Also source files should not go into public directory - this is directory for build artificats.

sidharthachatterjee:
Hey @artbycrunk 

While this adds _some_ value, I don't believe the value is high enough versus the complexity and added API surface bloat. Considering such tradeoffs is important for us, especially when adding new options to the public API for gatsby or the core plugins like this one. 

The results of the output folder are currently controlled by the compiler (Gatsby) and ceding that control will make it harder for us to make guarantees over outputs in the future. 

Because of the above, I'm closing this PR. However, please know that we _absolutely_ value having you as a contributor though so please don't feel discouraged by this! We value the time you put into writing this and opening it. 

Issue #12621, opened 2019-03-16
Support JSX nodes in `gatsby-remark-copy-linked-files`


danoc:
## Summary

Hello! A [recent change to MDX](https://github.com/mdx-js/mdx/pull/429) prevents `gatsby-remark-copy-linked-files` from recognizing `<img>`, `<a>`, and other tags in MDX files. This happens because `gatsby-remark-copy-linked-files` is now receiving `jsx` nodes, not `html`.

`gatsby-remark-copy-linked-files` has support for visiting `html` nodes: https://github.com/gatsbyjs/gatsby/blob/3834b2bce5edc60cd5386561498ca29f028c1d30/packages/gatsby-remark-copy-linked-files/src/index.js#L206

Should there also be a `visit` for JSX? If so, how would that work?

This is a follow-up to the original issue I opened in https://github.com/ChristopherBiscardi/gatsby-mdx/issues/311.

### Basic example

N/A

### Motivation

Be able to share `gatsby-remark-*` plugins across `.md` and `.mdx`.

Thanks!

***
cc: @ChristopherBiscardi — Hopefully I explained this correctly!

ChristopherBiscardi:
yep, looks good to me. For some additional context: MDX can be considered two layers for the purpose of this discussion.

* User Interface
* Remark AST Representation

# User Interface

The UI for Remark is Markdown and HTML tags. The UI for MDX is Markdown and JSX. Consider the following valid file in both formats

```markdown
# some title

<img src="./some/place" />
```

## Remark AST

In gatsby-transformer-remark, the initial AST is

```json
{
  "type": "root",
  "children": [
    {
      "type": "heading",
      "depth": 1,
      "children": [
        {
          "type": "text",
          "value": "some title",
        }
      ]
    },
    {
      "type": "html",
      "value": "<img src=\"./some/place\" />"
    }
  ]
}
```

## MDX Remark AST

In gatsby-mdx, the initial AST is

```json
{
  "type": "root",
  "children": [
    {
      "type": "heading",
      "depth": 1,
      "children": [
        {
          "type": "text",
          "value": "some title",
        }
      ]
    },
    {
      "type": "jsx",
      "value": "<img src=\"./some/place\" />"
    }
  ]
}
```

The core issue is the node that swaps between html and jsx, which prevents re-using the gatsby-remark-* plugins that operate on `html` nodes.

```json
{
  "type": "jsx",
  "value": "<img src=\"./some/place\" />"
}
```

[...]

PR #13552, opened 2019-04-22
fix(gatsby-remark-copy-linked-files): Support MDX by visiting JSX nodes

trevorblades:
## Description

This PR addresses an issue where files referenced in `` tags were ignored if the file was being parsed as MDX, since those nodes are of the `jsx` type, instead of the expected `html`. To address this, I changed the `test` argument passed to `visit` from the string `'html'` to an array: `['html', 'jsx']`

The `test` argument expects some input that is compatible with [`is`](https://github.com/syntax-tree/unist-util-is#istest-node-index-parent-context), which includes an array of strings as used in this PR. Here's some API reference for the `test` arg as it pertains to the `visit` function: https://github.com/syntax-tree/unist-util-visit-parents#visittree-test-visitor-reverse

## Related Issues

Fixes https://github.com/gatsbyjs/gatsby/issues/12621

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -24,6 +24,7 @@ module.exports = {
   notify: true,
   verbose: true,
   roots: pkgs,
+  transformIgnorePatterns: [`node_modules/(?!remark-mdx)`],
   modulePathIgnorePatterns: ignoreDirs,
   coveragePathIgnorePatterns: ignoreDirs,
   testPathIgnorePatterns: [
--- a/jest.config.js
+++ b/jest.config.js
@@ -24,7 +24,6 @@ module.exports = {
   notify: true,
   verbose: true,
   roots: pkgs,
-  transformIgnorePatterns: [`node_modules/(?!remark-mdx)`],
   modulePathIgnorePatterns: ignoreDirs,
   coveragePathIgnorePatterns: ignoreDirs,
   testPathIgnorePatterns: [
--- a/packages/gatsby-remark-copy-linked-files/package.json
+++ b/packages/gatsby-remark-copy-linked-files/package.json
@@ -22,7 +22,8 @@
     "babel-preset-gatsby-package": "^0.1.4",
     "cross-env": "^5.1.4",
     "remark": "^10.0.1",
-    "remark-mdx": "^1.0.14"
+    "remark-mdx": "^1.0.14",
+    "semver": "^6.0.0"
   },
   "homepage": "https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-remark-copy-linked-files#readme",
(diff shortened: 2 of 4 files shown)


pieh:
The code change looks very reasonable. It's pretty unfortunate that `remark-mdx` requires node 8+ (failing node6 unit tests)

/cc @ChristopherBiscardi 

Does whole mdx ecosystem require node 8+? If so, I'm thinking just skipping mdx related tests in node 6 instead of trying to fix them

trevorblades:
@pieh yea I noticed that! It's too bad 😢  I just made another commit that adds a line to the root `jest` config so that `remark-mdx` is transformed before tests run, but I would also understand if you didn't want to use that kind of workaround. I'm happy to skip on the JSX test if you would prefer that!

pieh:
I would rather see tests being "honest" and not "forced" to pass: those should not pass in node 6 if package itself doesn't work there without some extra work, so skipping test in node 6 (possibly just with comment that `remark-mdx` doesn't support node 6) is (IMO) clearer approach

pieh:
some guidelines to address skipping test in Node 6 only:
we can use `process.versions.node` and `semver` package to something like this:

```
  // helper somewhere near the top of the file
  const semver = require('semver')
  const testInNode8OrHigher = (...args) => {
    const isNode8OrHigher = semver.satisfies(process.versions.node, ">=8")
    if (isNode8OrHigher) {
      it(...args)
    } else {
      // maybe manipulate test title to note it was skipped because it's not node6
      it.skip(...args)
    }
  }


  // use helper
  testInNode8OrHigher(`can copy JSX images`, async () => {
    const path = `images/sample-image.gif`

    const markdownAST = remark()
      .use(mdx)
      .parse(`<img src="${path}" />`)

    await plugin({ files: getFiles(path), markdownAST, markdownNode, getNode })

    expect(fsExtra.copy).toHaveBeenCalled()
  })
```

trevorblades:
@pieh sounds good. At first, I was thinking of wrapping the test in a `if (semver.gte(...`, but your solution using `skip` looks much better. Thanks!

trevorblades:
It turned out that just importing `remark-mdx` was breaking the tests in node 6, so I moved the `require` statement inside the test, so that it only gets imported in node 8 or greater.

ChristopherBiscardi:
> Does whole mdx ecosystem require node 8+? 

yeah, I believe that's the case. Not really intentionally, that's just how it worked out over time.

@johno does this look good on the `remark-mdx` side to you?

sidharthachatterjee:
Published in `gatsby-remark-copy-linked-files@2.0.12`

Issue #13549, opened 2019-04-22
Alias `--with-prefix` to `--prefix-paths`


jaredsilver:
I don't know about everyone else, but I make this mistake all the time. `--with-prefix` feels like a more natural way to say this.

frinyvonnick:
I made a PR for this 😄

wardpeet:
It's how you would read this 😛 . --prefix-paths actually tells we add prefix to paths, with-prefix does not.

It's probably better to make sure we show you a message like --prefix-paths not found, you probably mean --with-prefix. We have a https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-cli/src/did-you-mean.js but I'm unsure how it actually works

jaredsilver:
It’s not necessarily just paths that are being prefixed now though. It could also be assets.

sidharthachatterjee:
> It’s not necessarily just paths that are being prefixed now though. It could also be assets.

@jaredsilver While I understand that this can be confusing, I don't think this is a valuable addition because:
- This could be solved by better documentation
- Added aliases will need to be maintained till the next breaking version (and best to avoid API surface bloat)

What do you think?

jaredsilver:
Sorry, I should clarify that the way that I read this is "gatsby build with prefix" which makes sense as an English sentence whereas "gatsby build asset prefix" does not. If others don't have this same issue, then it's not really a problem. I'll close the issue.

PR #13565, opened 2019-04-23
feat(gatsby-cli): add alias with-prefix to prefix-paths

frinyvonnick:
## Description

Add an command `with-prefix` that alias `prefix-paths`

## Related Issues

#13549

### DIFF
--- a/packages/gatsby-cli/src/create-cli.js
+++ b/packages/gatsby-cli/src/create-cli.js
@@ -182,6 +182,10 @@ function buildLocalCommands(cli, isLocalSite) {
         default: false,
         describe: `Build site with link paths prefixed (set pathPrefix in your gatsby-config.js).`,
       })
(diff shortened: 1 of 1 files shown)


wardpeet:
See https://github.com/gatsbyjs/gatsby/issues/13549#issuecomment-485804375

frinyvonnick:
Thanks for you review @sidharthachatterjee @wardpeet, I'll wait until the discussion has been resolved 👍 !

frinyvonnick:
I close this since the issue has been too ?

KyleAMathews:
Yeah let's close it. It's an excellent suggestion for a flag name and I wish you'd been around when it got named it a few years ago as that might have won out — but the current name is fine enough and adding aliases causes more confusion than would help.

Perhaps creating a custom npm script would work better for you?

PR #13566, opened 2019-04-23
feat(gatsby-image): Add durationFadeIn

### LINKED ISSUE #12836: gatsby image fadeIn duration
## Summary

Gatsby Image property fadeIn takes a numeric value instead of boolean and sets the duration of the transition to that value.

### Motivation

I am using traced svgs as placeholders. They are very pretty in themselves and are perfect if loading time is actually slow. However, when the actual images load fast, the placeholder is more irritating then helpful. The placeholder is visually so different that it needs to be seen a bit longer to appreciate the effect. I'd love to increase the transition duration via passing a property so that the effect visually works on both slow and fast connections.

frinyvonnick:
## Description

Make `Img` component from `gatsby-image` accept number instead of boolean to customize animation duration.

## Related Issues

Closes #12836

### DIFF
--- a/packages/gatsby-image/src/__tests__/index.js
+++ b/packages/gatsby-image/src/__tests__/index.js
@@ -73,7 +73,7 @@ describe(`<Image />`, () => {
     )
     // No Intersection Observer in JSDOM, so placeholder img will be visible (opacity 1) by default
     expect(placeholderImageTag.getAttribute(`style`)).toEqual(
-      `position: absolute; top: 0px; left: 0px; width: 100%; height: 100%; object-fit: cover; object-position: center; opacity: 1; transition-delay: 0.5s; color: red;`
+      `position: absolute; top: 0px; left: 0px; width: 100%; height: 100%; object-fit: cover; object-position: center; opacity: 1; transition-delay: 500ms; color: red;`
     )
     expect(placeholderImageTag.getAttribute(`class`)).toEqual(`placeholder`)
   })
--- a/packages/gatsby-image/src/__tests__/index.js
+++ b/packages/gatsby-image/src/__tests__/index.js
@@ -23,7 +23,12 @@ const fluidShapeMock = {
   base64: `string_of_base64`,
 }
 
-const setup = (fluid = false, onLoad = () => {}, onError = () => {}) => {
+const setup = (
+  fluid = false,
+  props = {},
+  onLoad = () => {},
+  onError = () => {}
+) => {
   const { container } = render(
     <Image
       backgroundColor
@@ -39,6 +44,7 @@ const setup = (fluid = false, onLoad = () => {}, onError = () => {}) => {
       itemProp={`item-prop-for-the-image`}
       placeholderStyle={{ color: `red` }}
       placeholderClassName={`placeholder`}
+      {...props}
     />
   )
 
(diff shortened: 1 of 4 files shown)


packages/gatsby-image/src/index.js
```diff
@@ -260,9 +260,13 @@ class Image extends React.Component {
       itemProp,
     } = convertProps(this.props)
 
-    const shouldReveal = this.state.imgLoaded || this.state.fadeIn === false
-    const shouldFadeIn = this.state.fadeIn === true && !this.state.imgCached
-    const durationFadeIn = `0.5s`
+    const shouldReveal =
+      this.state.imgLoaded || Boolean(this.state.fadeIn) === false
+    const shouldFadeIn =
+      Boolean(this.state.fadeIn) === true && !this.state.imgCached
```
sidharthachatterjee: Changing the type of an option and relying on coercion feels like a bad idea? Should we add a separate _opt-in_ `prop` called `durationFadeIn` (that defaults to 0.5s)?

KyleAMathews: ooo yeah, I like this better 👍 

They are separate ideas so splitting makes sense + better backwards compatibility

frinyvonnick: Sure, I'll do it !

frinyvonnick:
Thanks for your reviews @KyleAMathews @sidharthachatterjee 👍 I made the changes 😉 !

sidharthachatterjee:
@frinyvonnick Some snapshot tests seem to be failing

```
- position: absolute; top: 0px; left: 0px; width: 100%; height: 100%; object-fit: cover; object-position: center; opacity: 1; transition-delay: 0.5s; color: red;
    + position: absolute; top: 0px; left: 0px; width: 100%; height: 100%; object-fit: cover; object-position: center; opacity: 1; transition-delay: 500px; color: red;
```

packages/gatsby-image/src/__tests__/index.js
```diff
@@ -73,7 +73,7 @@ describe(`<Image />`, () => {
     )
     // No Intersection Observer in JSDOM, so placeholder img will be visible (opacity 1) by default
     expect(placeholderImageTag.getAttribute(`style`)).toEqual(
-      `position: absolute; top: 0px; left: 0px; width: 100%; height: 100%; object-fit: cover; object-position: center; opacity: 1; transition-delay: 0.5s; color: red;`
+      `position: absolute; top: 0px; left: 0px; width: 100%; height: 100%; object-fit: cover; object-position: center; opacity: 1; transition-delay: 500ms; color: red;`
```
sidharthachatterjee: Bit of a nit pick! (Feel free to ignore)

Do we need to update the unit from seconds to milliseconds?

packages/gatsby-image/src/index.js
```diff
@@ -509,6 +510,7 @@ Image.propTypes = {
   fixed: fixedObject,
   fluid: fluidObject,
   fadeIn: PropTypes.bool,
+  durationFadeIn: PropTypes.number,
```
sidharthachatterjee: Worth adding a snapshot test with a value passed in?

packages/gatsby-image/src/__tests__/index.js
frinyvonnick: Usually these kind of animation are under 1s, no ? I found it more readable. I can change it back to seconds :smile: !

sidharthachatterjee: No, you're right! I'm just wary of change unless absolutely required. This wasn't a public API before this PR so `ms` are good!

sidharthachatterjee:
Published in `gatsby-image@2.0.40`

PR #13568, opened 2019-04-23
Revert "chore(*): Use new createContentDigest helper (#8992)"

KyleAMathews:
This reverts commit c91d110807d900bc6b8aeb629ff319fb8e5b9a49 and PR https://github.com/gatsbyjs/gatsby/pull/8992

We can't introduce a new file to the core gatsby package and then also update packages to use it without having a fallback as people often update/install packages without updating core.

### DIFF
--- a/packages/gatsby-source-filesystem/package.json
+++ b/packages/gatsby-source-filesystem/package.json
@@ -36,7 +36,7 @@
   ],
   "license": "MIT",
   "peerDependencies": {
-    "gatsby": "^2.0.15"
+    "gatsby": "^2.0.0"
   },
   "repository": "https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-source-filesystem",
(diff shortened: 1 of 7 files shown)


KyleAMathews:
There was a new file added https://github.com/gatsbyjs/gatsby/pull/8992/files#diff-7b0d4dc1ad15094456fc37848d323bb1

The problem is that `npm` doesn't enforce `peerDependencies` so while there'd be a warning, most people wouldn't see that and not understand why upgrading/installing a package broke things.

KyleAMathews:
Another problem @pieh just pointed out is that the `utils.js` file wasn't published due to https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/package.json#L147-L152

Olanetsoft:
I still get the same error how do I fix it, please?

PR #13579, opened 2019-04-23
fix(docs): Fix setup code for Gatsby theme starter

erictleung:
## Description

The current code block is missing the name of the new Gatsby project and
thus will create an error when you run this code. This commit fixes and
adds a name for the Gatsby starter theme to get started.

## Related Issues

I couldn't find any.


pieh:
Thanks @erictleung!

PR #13028, opened 2019-04-02
feat(schema): Inference controls and improvements

freiksenet:
# Evolution of schema customization

## tl/dr

Try it now with `@schema-customization` tag.

```
yarn add gatsby@schema-customization
```

Use resolver directives/extensions to add default arguments or/and resolvers to fields.

```graphql
type MyType @infer {
  date: Date @dateformat(formatString: "DD MMM", locale: "fi")
  image: File @fileByRelativePath
  authorByEmail: Author @link(by: "email")
}
```

Use `@dontInfer` and enjoy better performance with no inferrence.

As usual, report issues in #12272

## Summary

After about a month of testing schema customization both here and in pre-release we determined a couple of issues. We set up to do this because we wanted to remove uncertainty in people's schemas, when the data changes. However, the original design allowed some uncertainties anyway. In addition, we have made inferrence a more heavy process, trading performance for consistency and didn't really provide a way to opt out of it completely. To summarize:

* Resolvers and arguments of fields like Date and File was determined by inferred data
* There was no easy way to use arguments/resolvers to override the above
* Inferrence was run even when `@dontInfer` flag was on
* There was no way to control inference outside of SDL

Therefore we have some proposed changes to the way we do inferrence and some deprecations.

## noDefaultResolvers and inferrence modes

First of all, we are deprecating `noDefaultResolvers` argument of `infer` and `dontInfer`. We feel it was confusing and very catch all and didn't even add resolvers at the end. We will support `noDefaultResolvers` until version 3, after which `@infer` behaviour will become a default and `noDefaultResolvers` will be removed.

We didn't want to break things, so we keep old default behaviour, even though we think it's not optimal. Add explicit `@infer` and resolver extensions to fields to be future proof.

### Default (deprecated in v3)

Applies with no `@infer` or there is `@infer(noDefaultResolvers: false)` on type.

Type gets all inferred fields added. If type has defined fields of types `Date`, `File` and any other node, and we inferred that they should have resolvers/args, resolvers/args will be added to type with a warning.

### Strict inference (future default in v3)

Applies with `@infer` or `@infer(noDefaultResolvers: true)`.

Type gets all inferred fields added. Existing fields won't automatically get resolvers

### No inferrence 

Applies with `@dontInfer` or `@dontInfer(noDefaultResolvers: true)`.

Inferrence won't run at all. Existing fields won't automatically get resolvers

### No new fields with default resolvers (deprecated, removed in v3)

Applies with `@dontInfer(noDefaultResolvers: false)`

[...]

PR #13591, opened 2019-04-24
[schema] Ensure default by: id for link() resolver

stefanprobst:
This is for #13028

Because `addResolver` directive is generic we don't have defaultValue for options. So better ensure we have default `by: id` arg for link resolver.

### DIFF
--- a/packages/gatsby/src/schema/resolvers.js
+++ b/packages/gatsby/src/schema/resolvers.js
@@ -93,7 +93,7 @@ const paginate = (results = [], { skip = 0, limit }) => {
   }
 }
 
-const link = ({ by, from }) => async (source, args, context, info) => {
+const link = ({ by = `id`, from }) => async (source, args, context, info) => {
   const fieldValue = source && source[from || info.fieldName]
 
   if (fieldValue == null || _.isPlainObject(fieldValue)) return fieldValue


packages/gatsby/src/schema/resolvers.js
```diff
@@ -93,7 +93,7 @@ const paginate = (results = [], { skip = 0, limit }) => {
   }
 }
 
-const link = ({ by, from }) => async (source, args, context, info) => {
+const link = ({ by = `id`, from }) => async (source, args, context, info) => {
```
sidharthachatterjee: I wonder if we can/need to generate a unique id (based on timestamp or something) to set as default?

stefanprobst: @sidharthachatterjee This is about the default node field for foreign-key relations. So that it is possible to say something like this in SDL:
```graphql
type Post implements Node {
  author: AuthorJson @addResolver(type: "link")
  # instead of:
  # author: AuthorJson @addResolver(type: "link", options: { by: "id" })
}
```
The ID itself must already be unique on nodes. Hope that make sense?

sidharthachatterjee: That makes sense! Thank you for the added context, @stefanprobst

Issue #12923, opened 2019-03-28
[feat]: make analytics notice more obvious


DSchau:
## Description

Surfaced in #12922, it seems apparent that the telemetry notice--while helpful--can easily be overlooked as a wall of text from other logging draws more attention.

To remediate, I propose we make the telemetry notice more obvious so that users are aware of the opt-out nature and that we are beginning to track on telemetry data.

_Note: curious about this telemetry data? Check out https://gatsby.dev/telemetry for more info_

## Ideas

I have a few ideas for how to make the notice more apparent:

- Usage of color
    - e.g. chalk.blue, chalk.bold, etc. could be used to separate the notice from other logging
- Usage of padding and/or a tool like [boxen](https://github.com/sindresorhus/boxen)
- Moving the notice to a different phase in Gatsby core, e.g. in Gatsby bootstrap

KyleAMathews:
Just came here to suggest boxen too :-)

DSchau:
@KyleAMathews when we roll out the Ink functionality (cc @wardpeet) there's even an [Ink component, too](https://github.com/sindresorhus/ink-box).

PR #13597, opened 2019-04-24
feat(gatsby-telemetry): Clearer notice through `boxen`

LekoArts:
## Description

image

- Added a blue box around the message
- Added a line-break due to text wrapping: https://github.com/sindresorhus/boxen/issues/16
- Put the notice into its own function for re-usability

## Related Issues

Fixes https://github.com/gatsbyjs/gatsby/issues/12923

The reposition of the note to "after bootstrap" wasn't pursued as (after talking with @jamo) this would mean a bigger rewrite. Due to the discussion about the CLI in general & Ink this idea was postponed.

---

### Purple version

We could also choose any other color, e.g. purple

image

### DIFF
--- a/packages/gatsby-telemetry/src/postinstall.js
+++ b/packages/gatsby-telemetry/src/postinstall.js
@@ -1,4 +1,4 @@
-const notice = require(`./notice`)
+const showAnalyticsNotification = require(`./showAnalyticsNotification`)
 
 let enabled = false
 try {
@@ -7,7 +7,7 @@ try {
   const config = new Configstore(`gatsby`, {}, { globalConfigPath: true })
   enabled = config.get(`telemetry.enabled`)
   if (enabled === undefined && !ci.isCI) {
-    notice()
+    showAnalyticsNotification()
   }
 } catch (e) {
   // ignore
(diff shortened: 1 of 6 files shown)


packages/gatsby-telemetry/src/postinstall.js
```diff
@@ -1,15 +1,13 @@
...
   const Configstore = require(`configstore`)
   const config = new Configstore(`gatsby`, {}, { globalConfigPath: true })
   enabled = config.get(`telemetry.enabled`)
   if (enabled === undefined && !ci.isCI) {
-    console.log(
-      `Gatsby has started collecting anonymous usage analytics to help improve Gatsby for all users.\n` +
-        `If you'd like to opt-out, you can use \`gatsby telemetry --disable\`\n` +
-        `To learn more, checkout http://gatsby.dev/telemetry`
-    )
+    notice()
```
wardpeet: I'm more in favour of renaming the function or putting the text back here and in telemetry.

if we rename it I would go for
```suggestion
    showAnalyticsNotification()
```

LekoArts: Done 👌

sidharthachatterjee:
Published in `gatsby-telemetry@1.0.8`

PR #13598, opened 2019-04-24
fix(docs): Fix flatten to include current item (#13590)

### LINKED ISSUE #13590: [docs] Stub list not collecting all stubs
## Description

#13510 from @fk:

> I noticed we don't currently catch https://www.gatsbyjs.org/docs/best-practices-for-orgs/ (which is a stub, too) — I don't have the time to figure out why this happens, and I'm not sure if there are others we are missing

I believe this is due to some mishap in the findStubs method in the [stub-list.js](https://github.com/gatsbyjs/gatsby/blob/master/www/src/pages/contributing/stub-list.js) file.

jgalat:
## Description

Fixes `flatten` to also include the current item of the array

## Related Issues
Fixes  #13590

### DIFF
--- a/www/src/pages/contributing/stub-list.js
+++ b/www/src/pages/contributing/stub-list.js
@@ -19,7 +19,8 @@ const findStubs = pages =>
 
 const flatten = pages =>
   pages.reduce(
-    (flat, item) => flat.concat(item.items ? flatten(item.items) : item),
+    (flat, item) =>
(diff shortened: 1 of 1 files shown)


fk:
What @pieh said! <3 🤗

DSchau:
Sweet! Thanks for fixing my mistake 🎉

Issue #11934, opened 2019-02-20
React-Hot-Loader: react-🔥-dom patch is not detected. React 16.6+ features may not work.


kripod:
## Description

After updating all the dependencies of my starter project, I noticed the following message in the browser console after running `gatsby develop`:

```
React-Hot-Loader: react-🔥-dom patch is not detected. React 16.6+ features may not work.
```

### Steps to reproduce

1. Clone [gatsby-starter-strict@6c06471](https://github.com/kripod/gatsby-starter-strict/tree/6c06471110b9a821871decf5772330a332894b34)
2. `yarn && yarn develop`

### Expected result

No warnings should be thrown.

### Actual result

A warning is thrown even with a starter project.

### Environment

```
  System:
    OS: Windows 10
    CPU: (4) x64 Intel(R) Core(TM) i5-2500 CPU @ 3.30GHz
  Binaries:
    Yarn: 1.13.0 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.4.1 - C:\Program Files\nodejs\npm.CMD
  npmPackages:
    gatsby: ^2.1.10 => 2.1.10
    gatsby-plugin-manifest: ^2.0.18 => 2.0.18
    gatsby-plugin-offline: ^2.0.24 => 2.0.24
    gatsby-plugin-react-helmet: ^3.0.6 => 3.0.6
    gatsby-plugin-styled-components: ^3.0.6 => 3.0.6
    gatsby-plugin-typescript: ^2.0.8 => 2.0.8
```

t2ca:
I can confirm I have also seen this warning. I was also able to reproduce this warning with gatsby-starter-default.

wardpeet:
We are probably missing https://github.com/gaearon/react-hot-loader/tree/7089062eac273832102c074a368d5af27e23e0b0#webpack-plugin

It's not a big deal atm because reacth fire :fire: is not official yet.

achmadk:
@wardpeet I replace `react-dom` with [@hot-loader/react-dom](https://github.com/gaearon/react-hot-loader/tree/7089062eac273832102c074a368d5af27e23e0b0#react--dom) to remove this warning.

wardpeet:
that works as well! 💪

JustFly1984:
@achmadk @wardpeet  where did you replaced react-dom with 
@hot-loader/react-dom ?
I do not see that in my codebase, it should be somewhere inside the gatsby code

[...]

PR #13610, opened 2019-04-24
fix(gatsby): fixes react-hot-reloader for new react features

wardpeet:
## Description
Fixes HMR for hooks & suspense features with react-🔥-dom. It uses the webpack loader to patch react-dom as react-dom is a peer dependency we can't force users to install `@hot-loader/react-dom` as mentioned in the [README of react-hot-loader](https://github.com/gaearon/react-hot-loader#react--dom).

After:
image

Before:
image

It also actually fixes HMR for components using hooks 😛 not just remove the message.

## Related Issues
#11934

### DIFF
--- a/packages/gatsby/src/utils/webpack.config.js
+++ b/packages/gatsby/src/utils/webpack.config.js
@@ -156,6 +156,7 @@ module.exports = async (
       case `develop`:
         return {
           commons: [
+            `react-hot-loader/patch`,
             `event-source-polyfill`,
             `${require.resolve(
               `webpack-hot-middleware/client`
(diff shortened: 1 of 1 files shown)


sidharthachatterjee:
Published in `gatsby@2.3.31`

tu4mo:
Did this break dev mode in IE11? Getting `'Promise' is undefined` error.

hydrosquall:
This may be related to a break in the `npm run develop` build stage with the Gatsby docs site: https://github.com/gatsbyjs/gatsby/issues/13641

Issue #12924, opened 2019-03-28
Proxying lambda requests with gatsby serve


zslabs:
## Description

`developMiddleware` is lost in CLI `serve` command for Netlify Lambda functions.

### Steps to reproduce

Using `netlify-lambda` and the following rewrite for local access within `gatsby-config.js`, requests are getting through just fine when developing ([from the tutorial](https://www.gatsbyjs.org/docs/api-proxy/)):

```
  developMiddleware: (app) => {
    app.use(
      '/.netlify/functions/',
      proxy({
        target: 'http://localhost:9000',
        pathRewrite: {
          '/.netlify/functions/': '',
        },
      }),
    );
  },
```

However, when I want to use `gatsby serve`, even when attempting to run things like so, the path rewrites are not happening (but are still available on port 9000):

```
"gatsby build && concurrently \"yarn start:lambda\" \"gatsby serve -p 10000\""
```

Overriding port here to not conflict with lambda server. Any thoughts on how we can expose the rewrite for the CLI `serve` command? Thanks!

### Expected result

Proxy API requests just as `gatsby develop` does.

### Actual result

The `:9000` endpoint is still available, but not respecting proxy. I do understand the `serve` is not meant to run as `develop`, but with the current config; it makes local build testing not work correctly.

### Environment

```
  System:
    OS: macOS 10.14.4
    CPU: x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Shell: 3.0.0 - /usr/local/bin/fish
  Binaries:
    Node: 10.15.3 - /usr/local/bin/node
    Yarn: 1.12.1 - ~/.yarn/bin/yarn
    npm: 6.5.0 - ~/.npm-packages/bin/npm
  Languages:
    Python: 2.7.10 - /usr/bin/python
  Browsers:
    Chrome: 73.0.3683.86
    Firefox: 65.0.1
    Safari: 12.1
  npmPackages:
    gatsby: 2.2.11 => 2.2.11
    gatsby-image: 2.0.34 => 2.0.34
    gatsby-plugin-canonical-urls: 2.0.12 => 2.0.12
    gatsby-plugin-catch-links: 2.0.13 => 2.0.13
    gatsby-plugin-feed: 2.1.0 => 2.1.0
    gatsby-plugin-google-tagmanager: 2.0.11 => 2.0.11
    gatsby-plugin-manifest: 2.0.24 => 2.0.24
[...]

PR #13613, opened 2019-04-24
Adding new api serveMiddleware

iFlameing:
## Description

This pr related to #12924. It exposes a new API serveMiddleware which is used for configuring serve port while using proxy.

@wardpeet can you review this :)


sidharthachatterjee:
Hey @iFlameing Thanks so much for opening this!

We just discussed this and think the value this adds versus the maintenance overhead makes adding this not worth it. It's important for us to measure these tradeoffs for any newly proposed APIs (since we'd have to maintain them) and with respect to `serveMiddleware`, it seems like this is a pretty small edge case. 

Since we'd like to keep `serve` as lean as possible, we won't be accepting this at the moment. 

We absolutely value having you as a contributor though and look forward to seeing more PRs from you in the future! 💪 💜

PR #13616, opened 2019-04-25
[feedback-widget] feat: lazy load feedback widget component

pieh:
This is PR against https://github.com/gatsbyjs/gatsby/pull/13550

After running webpagetest:
- current code - https://www.webpagetest.org/result/190424_CA_151d1ef08f83ff5d1da8e8fa119dc778/
- lazy loaded widget - https://www.webpagetest.org/result/190424_YE_957ff7c9ee1556fec219362c60fefa0e/

first interactive seems significantly faster on lazy loaded one and feedback component isn't critical to be immediately shown on initial mount / page transition so I think tradeoff of (slightly) delaying it in favour of faster overall interactivity seems good idea

Preview for this version available at https://5cc0c487ba5e4e79a0a62cf8--gatsby-org-pieh-previews.netlify.com/docs/


KyleAMathews:
Could we lazy load it on hover or click?

KyleAMathews:
onMouseOver actually?

pieh:
> Could we lazy load it on hover or click?
> onMouseOver actually?
Sure, would require some reorganization (as we would want to show button always) and start lazy loading on hover (and focus :) ). Probably would also need some extra handling to queue click events if it wan't loaded yet to not have glitchy experience

pieh:
I can address this tomorrow

pieh:
Preview link after addressing lazy loading on hover/focus - https://5cc1c705443c86d19f93cd46--gatsby-org-pieh-previews.netlify.com/docs/

pieh:
Here's preview of it - lazy loading starts when button is hovered or focused

Kapture 2019-04-25 at 21 02 57

On slow connection (as shown with emulated 2g connection above) it could be useful to indicate somehow that something is happening (if after clicking the button widget doesn't load in 500-1000ms)

KyleAMathews:
Maybe use intersection observer then? A little more head start. It's 23kb so not that much code.

KyleAMathews:
The animations are great btw 😍

pieh:
> The animations are great btw 😍

All hail @greglobinski ! :)

> Maybe use intersection observer then? A little more head start. It's 23kb so not that much code.

Would you want to show button before we load full widget? Because if yes, then we will still have same problem on slower connection - user can click button before widget is ready to be used. If we don't show button, then there would be this weird (IMO) interaction on mobile, when you scroll to the bottom, suddenly button would "randomly" show up (on desktop button is fixed in viewport)

pieh:
In any case - I'll try to repurpose the "working" spinning icon that is used when submitting feedback

wardpeet:
Because the button has a scale, the svg is pretty blurry when transitioning, it also isn't super fluent better to set the width & height of the button to the correct value and scale down instead of scaling up. 🙈 

Perhaps add a spinner to the button or set the button to disabled to show that something is happening. We could also leverage suspense for this to only show this when it's taking long.

marcysutton:
In testing [the preview link](https://5cc1c705443c86d19f93cd46--gatsby-org-pieh-previews.netlify.com/docs/) with the keyboard and Voiceover, I can't seem to trigger the feedback widget to open with the space or enter keys. 

There is also no visible focus style–possibly because of the `display: flex` on the button? I'm not quite sure.

pieh:
> Because the button has a scale, the svg is pretty blurry when transitioning, it also isn't super fluent better to set the width & height of the button to the correct value and scale down instead of scaling up. 🙈

(I think) I didn't change anything related to styling (other than moving few things around), so this potentially would need to be mentioned in main PR - https://github.com/gatsbyjs/gatsby/pull/13550

> Perhaps add a spinner to the button or set the button to disabled to show that something is happening. We could also leverage suspense for this to only show this when it's taking long.

I don't want to "disable" button, as it would loose focusability (I think). Suspense I think is problematic right now for us (SSR), but I can implement this by hand.

> In testing [the preview link](https://5cc1c705443c86d19f93cd46--gatsby-org-pieh-previews.netlify.com/docs/) with the keyboard and Voiceover, I can't seem to trigger the feedback widget to open with the space or enter keys.
> 
> There is also no visible focus style–possibly because of the `display: flex` on the button? I'm not quite sure.

What browser were You using? It seems to work on Chrome for me (both navigation / toggling widget with space/enter and visible focus), but I want to take another look if I can reproduce this

marcysutton:
> What browser were You using?
I used Safari and Chrome. It doesn't work for me in either, even without Voiceover on.

greglobinski:
>  It seems to work on Chrome for me (both navigation / toggling widget with space/enter and visible focus)

For me too, also on FF (Ubuntu) 

> I used Safari and Chrome. It doesn't work for me in either, 

@marcysutton Did it work for you before?

greglobinski:
> the svg is pretty blurry when transitioning,

@wardpeet Are you pointing the `hover` rotating transition of the "close-X" icon? I will fix that, thanks?

pieh:
> What browser were You using? It seems to work on Chrome for me (both navigation / toggling widget with space/enter and visible focus), but I want to take another look if I can reproduce this

I was able to reproduce this in Safari. This happens because the initial button is removed once full widget loads, and focus is not being transferred to toggle button from full widget. It's pretty tough tho, because re-focusing button has visible drop-down animation (suddenly orange "outline" disappear and is animated back in).

I hoped using same `ref` would prevent that, but it doesn't and this is result of react diffing algorithm. TBD what's best way to solve this - either prevent focus outline animation on initial widget render or hoist the button totally, so it never change location in component tree.

pieh:
I forgot to add preview for latest version of it - https://5cc73d8862760198feb5f6c5--gatsby-org-pieh-previews.netlify.com/docs/

It should also fix focus/hover box shadow on toggle button in Safari

www/src/components/feedback-widget/lazy-feedback-widget.js
```diff
@@ -0,0 +1,65 @@
...
+      // if we didn't load widget within 1 second, show loading indicator
+      setWidget(state => {
+        if (!state.Widget) {
+          return {
+            loading: true,
+          }
+        }
+        return state
+      })
+    }, 1000)
```
wardpeet: following rail practices https://developers.google.com/web/fundamentals/performance/rail this should be 100ms.

pieh: I have no strong opinion on this (but I think 100ms might be to short?) - I've used timeout we use already in core https://github.com/gatsbyjs/gatsby/blob/f4e8d618e4cc77624df9b3d1779b2dcac2fdfcc5/packages/gatsby/cache-dir/navigation.js#L75-L82
for plugins like `gatsby-plugin-nprogress` to show loader on page transitions

marcysutton:
What should we do with this PR @pieh?

pieh:
> What should we do with this PR @pieh?

Oh, hmm, if this didn't prove to be problem we can just discard this PR (if we will want to get back to it - I will keep branch around, tho it's pretty outdated at this point)

PR #10897, opened 2019-01-07
chore(gatsby): add TypeScript definitions for config & plugin APIs

JamesMessinger:
## Description

I added TypeScript definitions for the [`gatsby-config.js`](https://www.gatsbyjs.org/docs/gatsby-config/#gatsby-config), [`gatsby-node.js`](https://www.gatsbyjs.org/docs/node-apis/), [`gatsby-browser.js`](https://www.gatsbyjs.org/docs/browser-apis/), and [`gatsby-ssr.js`](https://www.gatsbyjs.org/docs/ssr-apis/) APIs. 

Some of the definitions could use more details (for example, many functions are simply defined as `Function`), but at least every method, parameter, and property are defined, which provides good intellisense and code-completion assistance in tools like VSCode. 😎

## Related Issues
N/A

PR #13619, opened 2019-04-25
chore: add & fix type definitions for gatsby, gatsby-link, gatsby-source-fs

orta:
## Description

Improves the TypeScript type definitions to Gatsby, and also annotates them so that you get inline documentation + API references. Not complete, but a great start. Builds off the work by @JamesMessinger in https://github.com/gatsbyjs/gatsby/pull/10897

### config
Screen Shot 2019-04-24 at 9 46 53 PM

### gatsby-node
Screen Shot 2019-04-24 at 10 09 26 PM

and
Screen Shot 2019-04-24 at 10 10 09 PM

### Link

Screen Shot 2019-04-24 at 10 12 07 PM


packages/gatsby/index.d.ts
```diff
@@ -11,24 +14,859 @@ export {
...
+  tracing: Tracing;
+  [key: string]: unknown;
+}
+
+export interface Actions {
+  /** @see https://www.gatsbyjs.org/docs/actions/#deletePage */
+  deletePage: Function;
+
+  /** @see https://www.gatsbyjs.org/docs/actions/#createPage */
+  createPage(args: CreatePageArgs, option?: ActionOptions): void;
```
wardpeet: CreatePageArgs isn't really correct for this function

We'll need to reuse plugin a few types so probably best to create a type for it
```ts
interface ActionPlugin {
  name: string;
}
```

```suggestion
  createPage(args: {path: string, component: string, context: Record<string, unknown>}, plugin?: ActionPlugin, option?: ActionOptions): void;
```

packages/gatsby/index.d.ts
```diff
@@ -11,24 +14,859 @@ export {
...
+
+  /** @see https://www.gatsbyjs.org/docs/actions/#deletePage */
+  deleteNode(args: DeleteNodeArgs, option?: ActionOptions): void;
```
wardpeet: This one isn't correct either
```suggestion
  deleteNode(options: { node: Node }, plugin?: ActionPlugin, option?: ActionOptions): void;
```

packages/gatsby/index.d.ts
```diff
@@ -11,24 +14,859 @@ export {
...
+
+  /** @see https://www.gatsbyjs.org/docs/actions/#deleteNodes */
+  deleteNodes: Function;
+
+  /** @see https://www.gatsbyjs.org/docs/actions/#createNode */
+  createNode(node: Node, options?: ActionOptions): void;
```
wardpeet:
```suggestion
  createNode(node: Node, plugin?: ActionPlugin, options?: ActionOptions): void;
```

packages/gatsby/index.d.ts
```diff
@@ -11,24 +14,859 @@ export {
...
+
+  /** @see https://www.gatsbyjs.org/docs/actions/#touchNode */
+  touchNode: Function;
+  
+  /** @see https://www.gatsbyjs.org/docs/actions/#createNodeField */
+  createNodeField(args: CreateNodeFieldArgs, options?: ActionOptions): void;
```
wardpeet: I could not find `CreateNodeFieldArgs`
```suggestion
  createNodeField(args: {node: Node, fieldName?: string, fieldValue?: string, name?: string, value: any}, plugin?: ActionPlugin, options?: ActionOptions): void;
```

packages/gatsby/index.d.ts
```diff
@@ -11,24 +14,859 @@ export {
...
+}
+
+export interface Store {
+  dispatch: Function;
+  subscribe: Function;
+  getState: Function;
+  replaceReducer: Function;
+}
+
+export interface Reporter {
```
wardpeet: The reporter got some kind of makeover so these types should suffice

```ts
type logMessageType = (format: string, ...args: any[]) => void

export interface Reporter {
  stripIndent: Function;
  format: object;
  setVerbose(isVerbose: boolean): void;
  setNoColor(isNoColor: boolean): void;
  panic(...args: any[]): void;
  panicOnBuild(...args: any[]): void;
  error(message: string, error: Error): void;
  uptime(prefix: string): void;
  success: logMessageType;
  verbose: logMessageType;
  info: logMessageType;
  warn: logMessageType;
  log: logMessageType;
  activityTimer(name: string, activityArgs: { parentSpan: object }): { start: () => void, 
status(status: string): void, end: () => void, span: object}
}
```

stripIndent uses [common-tags](https://github.com/declandewet/common-tags) & format is using [chalk](https://github.com/chalk/chalk)

orta:
Perfect, I did this to learn the Gatsby API - so I'll make all those tweaks. 

Re: exported interfaces, it can depend a bit - if plugins or other functions end up exposing some of these interfaces then we'll have to include them. I'm definitely not in a position to know which is/isn't eventually exposed. So I'm happy to remove thing you recommend.

packages/gatsby/index.d.ts
orta: For now, it can be simpler, doing the whole `@types` dependency thing is a bit tricky ATM

stefanprobst:
Would be cool to also add
https://github.com/gatsbyjs/gatsby/blob/dcf6be845dbc73f5cda87f596c58afe8ce683eb8/packages/gatsby/src/redux/actions.js#L1335
https://github.com/gatsbyjs/gatsby/blob/dcf6be845dbc73f5cda87f596c58afe8ce683eb8/packages/gatsby/src/utils/api-node-docs.js#L238

wardpeet:
thanks, @stefanprobst I added both APIs.

@orta mind going over the 3 commits I made and fix them if I made some mistakes (I ran prettier on the last one so you should be able to ignore it unless we need to add some different configs). If so I can merge this one! 🎉 if anything is missing we can iterate on this one in a follow-up PR.

packages/gatsby-source-filesystem/index.d.ts
```diff
@@ -0,0 +1,72 @@
...
+  url: string;
+  store: object;
+  cache: object;
+  createNode: Function;
+  createNodeId: Function;
+  auth?: object;
+  ext?: string;
+}
+
+export interface FileSystemNode extends Node {
```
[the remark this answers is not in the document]
orta: I *think* this can be turned into something that extends from the node types import, but this is a good start for now IMO

wardpeet:
published in:
 - gatsby-image@2.0.41
 - gatsby-link@2.0.17
 - gatsby-plugin-feed@2.1.2
 - gatsby-plugin-google-analytics@2.0.19
 - gatsby-plugin-google-gtag@1.0.17
 - gatsby-source-filesystem@2.0.33
 - gatsby@2.3.34

antoinerousseau:
`createRemoteFileNode` returns a `Promise`, not a `FileSystemNode`

PR #13623, opened 2019-04-25
[WIP] [schema] Allow registering custom field extensions

stefanprobst:
This is a WIP for allowing custom GraphQL field extensions. This is how it works:

The `registerFieldExtension` API accepts an extension consisting of a description, extension options, and a `process` function that should return a (partial) field config which will be merged into the existing field config. The extension can decide to wrap or overwrite an existing resolver.

Example:

```js
exports.registerFieldExtension = ({ registerFieldExtension }) => {
  registerFieldExtension(`volume`, {
    description: `Adjust the volume.`,
    args: {
      loud: { type: `Boolean` },
    },
    // Return a new partial field config to extend the previous field config with.
    // Receives extension `args` and the current field config as parameters.
    process(defaults, prevFieldConfig) {
      return {
        type: `String`,
        args: {
          loud: { type: `Boolean` },
        },
        resolve(source, args, context, info) {
          const shouldUpperCase = args.loud != null ? args.loud : defaults.loud
          return shouldUpperCase ? fieldValue.toUpperCase() : fieldValue
        },
      }
    },
  })
}
```

It can then be used either with Gatsby Type Builders, or in SDL, e.g.:
```js
  actions.createTypes(`
    type Site implements Node {
      siteMetadata: SiteMetadata
    }
    type SiteMetadata {
      title: String @volume(loud: true)
    }
  `)
}
```

Would love to get feedback on this!


KartSriv:
Merged

KyleAMathews:
@KartSriv hey, any particular reason you merged that? It was WIP (work in progress) still so shouldn't have been merged. In the future, please don't merge PRs but just leave reviews and allow core maintainers to sign off on the work and do the merging.

stefanprobst:
Yes, this really shouldn't have been merged, as we first should have discussed if this is the way we want to go with extensions. Also, this shouldn't be part of the 2.4 release.

PR #13625, opened 2019-04-25
feat(gatsby): Add details to pagination info

stefanprobst:
Pretty sure I'm embarrassing myself here so I'd be super grateful if someone better at basic maths than I could fill in an easier pagination calculation. Thank you!


sidharthachatterjee:
Published in `gatsby@2.3.30`

KyleAMathews:
Nifty!

PR #13636, opened 2019-04-25
fix(www): Remove Unbird feedback widget

### LINKED ISSUE #13633: [www] remove plugin and starter feedback widget
On https://www.gatsbyjs.org/plugins/, there is a feedback collection widget for the Plugin Library. When I tried logging into Unbird (where the data is going), I found that Gatsby doesn't have a current subscription. The widget is also inaccessible, and uses a non-interactive SVG icon instead of a button element. Given this has gone unnoticed for a while and the plugin library data hasn't been reviewed, we should remove the widget.

plugin library widget

There is a separate effort to add a [feedback widget to the docs](#13550) that we will entirely control and thus won't require paying for a subscription. We could incorporate that into the Plugin Library section if we decide we want to solicit feedback.

polishedwp:
## Description
Fixes #13633 

Removed from `plugin.js` page the `Unbird` component and removed the component from `src/components`. If I can help any more let me know. I will be online this evening.


sidharthachatterjee:
Removed

- other instances in some other files
- in starter library as well 
- section from README

marcysutton:
Thanks guys! It's a little bittersweet to remove, but if needed we can reintroduce something more accessible that we control.

polishedwp:
Thanks for the opportunity to help some with a great project.

PR #13661, opened 2019-04-26
Docs: fix more plugin documentation links

marcysutton:
## Description

I opened https://github.com/gatsbyjs/gatsby/pull/13637 yesterday, and then I noticed some more link references that needed to be updated. I added some contributing documentation for how to fix links when moving things around, too, including redirects and find + replace instructions!

Hopefully this is the last of them. 😅


missmatsuko:
I think this mainly concerns the links in `docs/blog/2019-02-14-behind-the-scenes-q-and-a/index.md`, but wouldn't it be better for accessibility to use the title value as the link's text content rather than having it in a title attribute and using the URL as the link text? 

The whole URL is a lot to take in visually, and it looks like the screen readers that announce the title at all will also read the text content (the full URL):
https://www.powermapper.com/tests/screen-readers/labelling/a-title/

A couple of the links' titles (lines 81 and 125 are the ones I noted) are also set to the URL, so it may read the full URL twice for some screen readers.

marcysutton:
@missmatsuko totally good advice moving forward! If you're interested in submitting a PR after this gets merged, that would be a huge help. I used up the amount of time I could spend on this today before having to switch gears. I appreciate your input!

PR #13663, opened 2019-04-26
[schema] Merge user-defined types with plugin-defined types

stefanprobst:
When defining GraphQL types with `createTypes`, allow users to extend types which have already been defined by plugins, while disallowing plugins to modify types defined by other plugins.

Unfortunately this means that we cannot use upstream `schemaComposer.addTypeDefs` anymore, but there is no overhead in doing this ourselves.

To discuss: How exactly do we want extensions to be merged here?


freiksenet:
Does this check for object defined types too? I don't see where this is done. Also would like to see merge of SDL defined types with object defined types (and with graphql-js types) and vice versa, just to see that it works.

stefanprobst:
> Does this check for buildType defined types too?

Hmm not 100% sure what you mean?

> merge of SDL defined types with object defined types (and with graphql-js types) and vice versa

There are now tests for merging
* SDL into SDL
* typeBuilder into SDL
* SDL into typeBuilder
* typeBuilder into typeBuilder
* SDL into graphql-js
* [graphql-js into SDL](https://github.com/gatsbyjs/gatsby/blob/606def0aacdd2773b604270854800115336251c0/packages/gatsby/src/schema/__tests__/build-schema.js#L509) -- This last one does NOT work in cases where the grapql-js type has a field that references an ObjectType that is already defined in the schema composer - those will be duplicated. I'm a bit hesitant if we should add complexity for fixing this scenario? It's sort of the reason why TypeBuilders are preferable.

freiksenet:
@stefanprobst will you reopen this?

stefanprobst:
i'll make a fresh PR against master (this was against the other branch and i messed up with git)

stefanprobst:
continued in #14116

PR #13678, opened 2019-04-27
Update anchor titles and text for accessibility

missmatsuko:
## Description

I think the use of the title attribute on links has been overused/abused in the docs, and the text content of many links could be improved as well. 

Generally, I don't think the title attribute should be used, and there's a lot of material out there about why we should avoid using it:
- [HTML5 Accessibility Chops: title attribute use and abuse](https://developer.paciellogroup.com/blog/2012/01/html5-accessibility-chops-title-attribute-use-and-abuse/)
- [I thought title text improved accessibility. I was wrong.](https://silktide.com/blog/2013/i-thought-title-text-improved-accessibility-i-was-wrong)
- [The Trials and Tribulations of the Title Attribute](https://www.24a11y.com/2017/the-trials-and-tribulations-of-the-title-attribute/)
- [Don't Rely on the Title Attribute for Accessibility](https://www.mediacurrent.com/blog/dont-rely-title-attribute-accessibility-seo/)

I found many where the title value would be better if used as the anchor text, or the title was duplicating the anchor text or anchor URL. I don't think having the tooltip on hover containing the same text that's already visible or is a full URL is useful. Additionally, [screen readers don't always read aloud the title attribute, and when they do, it's usually in addition to the link text](https://www.powermapper.com/tests/screen-readers/labelling/a-title/). This means that a screen reader might read a huge URL twice when it's used as both the link's text and title!

I'm not sure if maybe I was a bit heavy-handed or maybe light-handed in some areas, but I think this is an improvement overall.

I tried to avoid making significant changes to verbiage.

I left some titles where I thought it wasn't too critical if it weren't read, and not so disruptive if it were read.

I also left some links where the text was the link URL when the URL itself might be important (e.g. localhost URLs) or if it was really short.


freiksenet:
Thank you so much for doing all this work!

PR #13692, opened 2019-04-28
Ensure a consistent schema for gatsby-plugin-documentationjs and add TypeScript support

devongovett:
This adds a schema definition for gatsby-plugin-documentationjs to ensure that it is consistent no matter what files are actually processed. Currently, with gatsby's automatic schema generation, the schema can be missing fields if they are unused by the code that is being parsed and therefore the query would have to change as the code changes. This ensures that the schema will have all fields available no matter what.

Also, [TypeScript support](https://github.com/documentationjs/documentation/pull/1234) was recently added to documentationjs, so this enables `.ts` and `.tsx` files to be processed as well.

### DIFF
--- a/packages/gatsby-transformer-documentationjs/src/gatsby-node.js
+++ b/packages/gatsby-transformer-documentationjs/src/gatsby-node.js
@@ -66,7 +66,7 @@ exports.sourceNodes = ({ actions }) => {
       type: DoctrineType
       default: JSON
       augments: [DocumentationJs]
-      examples: [JSON]
+      examples: [DocumentationJsExample]
       implements: [DocumentationJs]
       params: [DocumentationJs]
       properties: [DocumentationJs]
@@ -75,6 +75,11 @@ exports.sourceNodes = ({ actions }) => {
       todos: [DocumentationJs]
       yields: [DocumentationJs]
       members: DocumentationJsMembers
+    }
+
+    type DocumentationJsExample {
+      caption: String
+      description: String
     }
 
     type DocumentationJsMembers {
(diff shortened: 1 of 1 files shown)


freiksenet:
This is awesome. I'll test it with Gatsby docs website.

packages/gatsby-transformer-documentationjs/src/gatsby-node.js
```diff
@@ -42,6 +42,64 @@ function prepareDescriptionNode(node, markdownStr, name, helpers) {
...
+      alias: String
+      copyright: String
+      author: String
+      license: String
+      since: String
+      lends: String
+      type: DoctrineType
+      default: JSON
+      augments: [DocumentationJs]
+      examples: [JSON]
```
pieh: This (and other `JSON` fields in `DoctrineType`) do change schema that will be generated (which is not a blocker - just wanted to mentioned that we would need to bump major as this will be breaking change)

freiksenet: I feel that if shape of those types is defineable, it should be fully typed. JSON type should only be used in cases where shape of data is impossible to define.

devongovett: Examples probably can, but I specifically made the ones in DoctrineType JSON because they can be circularly typed, and therefore impossible to query with GraphQL...

freiksenet:
Published as `gatsby-transformer-documentationjs@4.0.0` (because of possible breaking change). Thank you @devongovett!

pieh:
I think we still need to do some work here - https://github.com/gatsbyjs/gatsby/blob/616fb24c332b9b32a601ef72c60ca08d42412916/packages/gatsby-transformer-documentationjs/package.json#L11 probably need to be bumped to `^10.1.0` (to actually support Typescript)

devongovett:
Oops, yes. 🤦‍♂️ will make another pr

PR #13693, opened 2019-04-28
[schema] Clarify type conflict warning message

stefanprobst:
We currently show a warning message in case of conflicting field types, even when the field type has been explicitly defined with `createTypes`, which is a bit confusing. This tries to give more context to the warning.

See https://github.com/gatsbyjs/gatsby/issues/12272#issuecomment-479227514 and https://github.com/gatsbyjs/gatsby/pull/13028#issuecomment-487342249


CanRau:
I like it, think it's a bit verbose, yet way better than currently.

Should I just merge it❓I never merged here, yet the big fat green button smiles at me 😄 
If so, is the squash and merge button what you guys do/recommend or what's the proper protocol?

PR #13709, opened 2019-04-29
chore: format - Use Https over http in docs

prashant-andani:
Use Https over http and typo fix


wardpeet:
Thanks for updating these links! :heart:

PR #13716, opened 2019-04-29
chore(gatsby): warn about service workers running in development

davidbailey00:
## Description

Warn about service workers detected on a site running in development, rather than removing them automatically.

## Related Issues

#12834


wardpeet:
I wonder if a console.warn is enough :p Code looks good but I would like some other opinions before merging.

Great job!

PR #13721, opened 2019-04-29
fix(www): Fix urlToSearch to correctly decode the query string

### LINKED ISSUE #13696: [docs] Plugin search input encoding
## Description
When visiting the docs plugin search directly with a pre-existing query in the URL, the search input displays the encoded characters, such as `%20`, instead of in this example, a space ` `.

image

**Note:** this does not appear to impact functionality, just aesthetics.

### Steps to reproduce
* Go to https://www.gatsbyjs.org/plugins/
* Search for `marketing platform` or really any multi-word phrase
* Simply hit refresh or copy the URL with the query and open a new tab with https://www.gatsbyjs.org/plugins/?=marketing%20platform
* See the plugins search input on the left

### Expected result
Would expect the input to show `marketing platform` instead of `marketing%20platform`.

### Actual result
It shows `marketing%20platform`.

### Environment
OSX 10.14.4 (18E226)
Chrome Version 76.0.3778.0 (Official Build) canary (64-bit)

jgalat:
## Description

Fixes urlToSearch to decode the query string correctly.

## Related Issues

Fixes #13696

### DIFF
--- a/www/src/components/plugin-searchbar-body.js
+++ b/www/src/components/plugin-searchbar-body.js
@@ -464,7 +464,10 @@ class PluginSearchBar extends Component {
   urlToSearch = () => {
     if (this.props.location.search) {
       // ignore this automatically added query parameter
-      return this.props.location.search.replace(`no-cache=1`, ``).slice(2)
+      const search = this.props.location.search
+        .replace(`no-cache=1`, ``)
+        .slice(2)
+      return decodeURIComponent(search)
     }
     return ``
   }


jgalat:
Hi @DSchau, thanks for reviewing this

It could be done with `decodeURIComponent`, but you'll also require to manually remove the string `'no-cache'` and the first 2 characters of the string as the original code did. With this approach, you don't have to do that.

KyleAMathews:
We'd rather not add extra dependencies if we don't have to.

jgalat:
Sure. I see your point. I'll update the code right away. Thanks for the review

KyleAMathews:
Thanks!

freiksenet:
Thanks!

PR #13726, opened 2019-04-29
Theme blog core: changing the navigation from a post because of the posts order

rachelcarmena:
## Description

**Problem** (small problem): the list of posts is in reverse order of date. However, the previous and next links from a post follow the order of date (not reverse). I think it's a little confusing. I'm not the only one who experimented the same thing.

**Solution** (small change): swapping the previous and next values. In this way, there will be the same order both in the main list and in a single post.

### DIFF
--- a/themes/gatsby-theme-blog-core/gatsby-node.js
+++ b/themes/gatsby-theme-blog-core/gatsby-node.js
@@ -94,9 +94,9 @@ exports.createPages = ({ graphql, actions }) => {
         const posts = result.data.allMarkdownRemark.edges
 
         posts.forEach((post, index) => {
-          const previous =
+          const next =
             index === posts.length - 1 ? null : posts[index + 1].node
-          const next = index === 0 ? null : posts[index - 1].node
+          const previous = index === 0 ? null : posts[index - 1].node
 
           debug(`creating`, post.node.fields.slug)
           createPage({


ChristopherBiscardi:
This logic is because we're matching the official gatsby starter functionality: https://github.com/gatsbyjs/gatsby/blob/0d82f2803961f1d443a86b46da311cca9a7c1309/starters/blog/gatsby-node.js#L37-L38 If you want to change it you should make the change in the starter first and get approval from @gatsbyjs/core for the change and then we'll follow that with this themes change.

I don't know what @gatsbyjs/core's viewpoint on the functionality change is, so will leave this open until they say something

LekoArts:
I don't think that swapping the values here is a good solution. The problem is that in the theme the order is not explicitly defined:

https://github.com/gatsbyjs/gatsby/blob/616fb24c332b9b32a601ef72c60ca08d42412916/themes/gatsby-theme-blog-core/src/pages/index.js#L32-L50

It also should have the `DESC` like here:

https://github.com/gatsbyjs/gatsby/blob/0d82f2803961f1d443a86b46da311cca9a7c1309/starters/blog/src/pages/index.js#L58-L74

ChristopherBiscardi:
oh word, yeah. I'd definitely accept a PR to fix the above ordering issue in the theme @rachelcarmena

LekoArts:
Hey! Thanks so much for opening this pull request!

I opened up an issue to track the bug you found and my proposed solution. You can take the issue on if you want! https://github.com/gatsbyjs/gatsby/issues/13795

rachelcarmena:
Thanks to you, @ChristopherBiscardi and @LekoArts, for your kindness and consideration. Happy to know that it's useful for you!

About the other alternative, I'm wondering if it can affect to the order of the posts in the main page. I'll follow the created issue to know more about it. I've seen it's already assigned.

Thanks for the info about the options to contribute! I hope to "see" you again :wink:

Issue #13641, opened 2019-04-26
[docs site] Webassembly Error When Building


hydrosquall:
## Description

Documentation site was unable to build locally, encountered while trying to test #13640 

### Steps to reproduce

Following the directions in the [docs contributions](https://github.com/gatsbyjs/gatsby/blob/master/docs/contributing/docs-contributions.md#docs-site-setup-instructions) page

### Expected result

Build completes without error

### Actual result

```
cd www/
yarn install
yarn run develop

$ GATSBY_GRAPHQL_IDE=playground gatsby develop
success open and validate gatsby-configs — 0.011 s
success load plugins — 0.659 s
success onPreInit — 0.007 s
success initialize cache — 0.006 s
success copy gatsby files — 0.041 s
success onPreBootstrap — 0.015 s
success source and transform nodes — 3.982 s
success building schema — 1.366 s
success createPages — 1.387 s
success createPagesStatefully — 0.066 s
success onPreExtractQueries — 0.002 s
success update schema — 0.091 s
success extract queries from components — 0.371 s
success run static queries — 0.030 s — 4/4 136.43 queries/second
success run page queries — 0.061 s — 9/9 149.28 queries/second
success write out page data — 0.033 s
success write out redirect data — 0.001 s
success Build manifest and related icons — 0.176 s
success onPostBootstrap — 0.177 s

info bootstrap finished - 10.044 s

error UNHANDLED REJECTION

  Error: You must provide the URL of lib/mappings.wasm by calling SourceMapConsumer.initialize({ 'lib/mappings.wasm': ... }) before using SourceMapConsumer
  
  - read-wasm.js:8 readWasm
    [www]/[react-hot-loader]/[source-map]/lib/read-wasm.js:8:13
  
  - wasm.js:25 wasm
    [www]/[react-hot-loader]/[source-map]/lib/wasm.js:25:16
  
  - source-map-consumer.js:264 BasicSourceMapConsumer.then.that
    [www]/[react-hot-loader]/[source-map]/lib/source-map-consumer.js:264:14
  
```

### Environment

```

  System:
    OS: macOS High Sierra 10.13.6
    CPU: (12) x64 Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 11.13.0 - /usr/local/bin/node
    Yarn: 1.13.0 - /usr/local/bin/yarn
[...]

PR #13731, opened 2019-04-30
fix(www): Update minimum required Gatsby version to 2.3.33

hydrosquall:
## Description

Sets a new minimum required version for the `www/` (docs) site, it looked appropriate since there is a bot which upgrades all the starters to use the latest Gatsby in https://github.com/gatsbyjs/gatsby/pull/13706 (I'm surprised it doesn't update this folder too).

I confirmed locally that this version does not run into the webassembly issue reported in #13641.

## Related Issues

- Ensures that people don't run into #13641, presently doing a clean yarn install led to reinstalling `2.3.31` on my device

### DIFF
--- a/www/package.json
+++ b/www/package.json
@@ -13,7 +13,7 @@
     "dotenv": "^6.0.0",
     "email-validator": "^1.1.1",
     "fuse.js": "^3.2.0",
-    "gatsby": "^2.3.0",
+    "gatsby": "^2.3.33",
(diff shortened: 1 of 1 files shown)


wardpeet:
@hydrosquall it's not really necessary as you can remove yarn.lock as well to get the latest versions but I agree that it's a good idea to bump www so people don't have to think about this. Good call! ❤️

PR #13738, opened 2019-04-30
[WIP] [schema] Allow registering custom field extensions

stefanprobst:
This is a WIP for allowing custom GraphQL field extensions. This is how it works:

The `registerFieldExtension` API accepts an extension consisting of a description, extension options, and a `process` function that should return a (partial) field config which will be merged into the existing field config. The extension can decide to wrap or overwrite an existing resolver.

Example:

```js
exports.registerFieldExtension = ({ registerFieldExtension }) => {
  registerFieldExtension(`volume`, {
    description: `Adjust the volume.`,
    args: {
      loud: { type: `Boolean` },
    },
    // Return a new partial field config to extend the previous field config with.
    // Receives extension `args` and the current field config as parameters.
    process(defaults, prevFieldConfig) {
      return {
        type: `String`,
        args: {
          loud: { type: `Boolean` },
        },
        resolve(source, args, context, info) {
          const shouldUpperCase = args.loud != null ? args.loud : defaults.loud
          return shouldUpperCase ? fieldValue.toUpperCase() : fieldValue
        },
      }
    },
  })
}
```

It can then be used either with Gatsby Type Builders, or in SDL, e.g.:

```js
  actions.createTypes(`
    type Site implements Node {
      siteMetadata: SiteMetadata
    }
    type SiteMetadata {
      title: String @volume(loud: true)
    }
  `)
}
```

Would love to get feedback on this!

Continued from #13623


freiksenet:
The code example doesn't feel right. I think resolver should get normal arguments of the field and args should come from the process arguments. Is there a typo there?

freiksenet:
So what it should be like is more

```js    process(extensionArgs, prevFieldConfig) {
      return {
        type: `String`,
        args: {
          loud: { type: `Boolean` },
        },
        resolve(source, args, context, info) {
          const shouldUpperCase = args.loud != null ? args.loud : defaults.loud
          return shouldUpperCase ? fieldValue.toUpperCase() : fieldValue
        },
      }
    },
```

freiksenet:
It seems to be correct in code, just not in the example.

packages/gatsby/src/schema/extensions/index.js
```diff
@@ -0,0 +1,210 @@
...
+const processFieldExtensions = ({
+  schemaComposer,
+  typeComposer,
+  parentSpan,
+}) => {
+  typeComposer.getFieldNames().forEach(fieldName => {
+    const extensions = typeComposer.getFieldExtensions(fieldName)
+    Object.keys(extensions)
+      .filter(name => !internalExtensionNames.includes(name))
+      .sort(a => a === `proxyFrom`) // Ensure `proxyFrom` is run last
```
freiksenet: I don't like those special cases. Now it feels extensions should be a list.

packages/gatsby/src/schema/schema.js
```diff
@@ -121,25 +123,29 @@ const processTypeComposer = async ({
...
+        typeComposer,
+        nodeStore,
+        parentSpan,
+      })
+      await addTypeToRootQuery({ schemaComposer, typeComposer, parentSpan })
+    }
   }
 }
 
+const registerExtensions = ({ parentSpan }) =>
```
freiksenet: No reason to have separate API. Let's make it an action like `createTypes`.

packages/gatsby/src/schema/extensions/index.js
```diff
@@ -0,0 +1,210 @@
...
+// Reserved for internal use
+const internalExtensionNames = [
+  `addDefaultResolvers`,
+  `createdFrom`,
+  `directives`,
+  `infer`,
+  `plugin`,
+]
+
+const typeExtensions = {
```
freiksenet: This global state should be in Redux and passed from the top like for other global state.

packages/gatsby/src/schema/extensions/index.js
```diff
@@ -0,0 +1,210 @@
...
+        description:
+          `Type of the resolver. Types available by default are: ` +
+          `"dateformat", "link" and "fileByRelativePath".`,
+      },
+      options: {
+        type: GraphQLJSON,
+        description: `Resolver options. Vary based on resolver type.`,
+      },
+    },
+    process(args, fieldConfig) {
```
freiksenet: :+1: Really like this API.

packages/gatsby/src/schema/extensions/index.js
```diff
@@ -0,0 +1,210 @@
...
+      },
+      options: {
+        type: GraphQLJSON,
+      },
+    },
+  },
+
+  addResolver: {
+    description: `Add a resolver specified by "type" to field.`,
+    args: {
```
freiksenet: We have a bit of a weird discrepancy, where directives do type checking for arguments, but extensions applied directly won't. Not sure if we should solve and if we should then how. One way would be to manually use args to validate.

stefanprobst:
> The code example doesn't feel right.

Hmm, could you point me to what's wrong? The `process` function gets the extension args as the first argument (which is called `defaults` here), and the field resolver gets the field args as the second argument (`args` here) as usual. So in the resolver we use the field arg if we get one, and fall back to the extension default arg if not.

packages/gatsby/src/schema/schema.js
stefanprobst: Let's first quickly discuss how happy we are with `createTypes` being an action? The issue is that actions get passed to all Node APIs but we need it to be called before schema generation, that's why `createTypes` should really only be called in `sourceNodes` even though it is available in other APIs as well. But I'm happy if we make this the general convention to follow.

packages/gatsby/src/schema/extensions/index.js
stefanprobst: Agreed

packages/gatsby/src/schema/extensions/index.js
stefanprobst: Yeah good point! Hmm yes probably need to manually do this for extensions

freiksenet:
@stefanprobst nm my comment about example, I misread everything :D It's actually correct.

packages/gatsby/src/schema/schema.js
freiksenet: Okay, let's do it this way:

1. Add deprecation warnings to actions for createTypes  and third party schema adding.
2. Create a schema customization runner API that is called after source nodes. Have createTypes, thirdPartySchemas and this new stuff available there.
3. Add all further schema customization stuff there.
4. Update source-graphql to use 3 instead of 1.

packages/gatsby/src/schema/extensions/index.js
```diff
@@ -0,0 +1,210 @@
...
+    process(args, fieldConfig) {
+      const { process } = fieldExtensions[args.type] || {}
+      if (typeof process === `function`) {
+        return process(args.options || {}, fieldConfig)
+      }
+      return {}
+    },
+  },
+
+  dateformat: {
```
freiksenet: Not 100% sure we should add them separately.

stefanprobst: I don't think there are disadvantages to adding the shortcuts as well. The advantage otoh is that it allows a more concise way to add functionality to fields. E.g. adding foreign-key relations (which is probably the most common use-case) can be shortened to
```graphql
field: SomeType @link
```
instead of
```graphql
field: SomeType @addResolver(type: "link")
```

packages/gatsby/src/schema/schema.js
stefanprobst: Sounds good! Probably should be a separate PR?

packages/gatsby/src/schema/extensions/index.js
freiksenet: I'm now considering not having addResolver, but having those instead.

The original motivation was that you'll have new resolvers to add, but API would be  the same with this extension api.

packages/gatsby/src/schema/schema.js
```diff
@@ -121,25 +123,29 @@ const processTypeComposer = async ({
...
+  apiRunner(`registerFieldExtension`, {
```
wardpeet:
```suggestion
  apiRunner(`createFieldExtension`, {
```

packages/gatsby/src/schema/schema.js
wardpeet: Separate PR sounds good!

packages/gatsby/src/schema/extensions/index.js
```diff
@@ -0,0 +1,210 @@
...
+          typeComposer.extendField(
+            fieldName,
+            process(extensions[name], prevFieldConfig)
+          )
+        }
+      })
+  })
+}
+
+const registerFieldExtension = (name, extension) => {
```
wardpeet: I feel like the winning condition is a bit buried in these ifs.

```js
const registerFieldExtension = (name, extension) => {
  if (internalExtensionNames.includes(name)) {
    report.error(`The extension ${name} is reserved for internal use.`)
    return
  }
  if (fieldExtensions[name]) {
    report.error(
      `A field extension with the name ${name} is already registered.`
    )
    return
  }

  fieldExtensions[name] = extension
  return
}
```

packages/gatsby/src/schema/extensions/index.js
wardpeet: won't you have the same issue with a list?

packages/gatsby/src/schema/extensions/index.js
```diff
@@ -0,0 +1,210 @@
...
+  typeComposer,
+  parentSpan,
+}) => {
+  typeComposer.getFieldNames().forEach(fieldName => {
+    const extensions = typeComposer.getFieldExtensions(fieldName)
+    Object.keys(extensions)
+      .filter(name => !internalExtensionNames.includes(name))
+      .sort(a => a === `proxyFrom`) // Ensure `proxyFrom` is run last
+      .forEach(name => {
+        const { process } = fieldExtensions[name] || {}
```
wardpeet: A little bit of bikeshed 😄 

what's the reason for the name process? It doesn't feel completely right to me. Process doesn't really tell me what this function does, does extend make more sense?

some other words that might apply: override, merge, apply

packages/gatsby/src/schema/extensions/index.js
```diff
@@ -0,0 +1,210 @@
...
+  defaultFieldResolver,
+} = require(`graphql`)
+const { GraphQLJSON } = require(`graphql-compose`)
+const report = require(`gatsby-cli/lib/reporter`)
+
+const { link, fileByPath } = require(`../resolvers`)
+const { getDateResolver } = require(`../types/date`)
+
+// Reserved for internal use
+const internalExtensionNames = [
```
wardpeet: should we prefix them? As we won't be able to add any more to the list unless we publish a major as this might break people's code.

packages/gatsby/src/schema/schema.js
freiksenet: Yeah, let's merge the thing without customization, then have a separate PR together with Redux changes.

packages/gatsby/src/schema/extensions/index.js
stefanprobst: Yes that's better :+1:

packages/gatsby/src/schema/extensions/index.js
stefanprobst: No particular reason, I'm open to suggestions -- maybe `extend` yes

packages/gatsby/src/schema/extensions/index.js
stefanprobst: Good point

stefanprobst:
I've removed the public `createFieldExtension` API and the `addResolver` extension. Next steps in follow-up PRs after 2.5 release should be:
* [move schema customization actions to one API](https://github.com/gatsbyjs/gatsby/pull/13738#discussion_r279738433) Does this need more discussion? I'm :+1:
* move extensions registry into redux
* add type checking for extension args
* decide on prefixing internal extension names
* decide on a name for `process` function. `extend`?

PR #13741, opened 2019-04-30
chore: add script to add packages to a npm team

wardpeet:
## Description
Attempt to create a script to add packages to a npm team. With teams we shouldn't have to add access rights to each npm packages. We struggled a bit with packages losing owners, hopefully teams fixes this.

There is also bad news, the npm cli doesn't support this so I just print out all packages which I can than use inside the npm website to add all these packages so I don't have to write them manually

this is the output you get when running:
```bash
Will be adding @gatsbyjs:ink-team to packages:
 - babel-preset-gatsby
 - gatsby-codemods
 - gatsby-plugin-sitemap
[
  "babel-preset-gatsby",
  "gatsby-codemods",
  "gatsby-plugin-sitemap",
  "gatsby-plugin-styletron"
]
```


scripts/get-unowned-packages/index.js
```diff
@@ -19,10 +19,18 @@ const getPackagesWithReadWriteAccess = async user => {
...
 }
 
-module.exports = function getUnownedPackages({
+module.exports = async function getUnownedPackages({
   rootPath = path.join(__dirname, `../..`),
   user,
 } = {}) {
+  // infer user from npm whoami
+  // set registry because yarn run hijacks registry
+  if (!user) {
```
wardpeet: I moved this up as we only need to do this check once and not for every package.

PR #13743, opened 2019-04-30
chore(gatsby-cli): Improve Flow annotations for reporter

mpartipilo:
## Description

Added some Flow annotations to describe the API for the CLI reporter


packages/gatsby-cli/src/reporter/types.js
```diff
@@ -0,0 +1,29 @@
...
+  end(): Function,
+  setStatus(s: string): Function,
+  span: Object,
+}
+
+export type ActivityArgs = {
+  parentSpan?: Object,
+}
+
+export interface Reporter {
```
wardpeet: is it possible to mimic these typings? As it's not entirely correct.
https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/index.d.ts#L892-L915

mpartipilo: Sure! I wasn't aware of the TS typings. I'll sync them.

mpartipilo: Maybe later these can be generated with flowgen

Issue #13714, opened 2019-04-29
cant disable telemetry using: export GATSBY_TELEMETRY_DISABLED=1


pivotal-kaveh:
Hi,

according to [telemetry docs](https://www.gatsbyjs.org/docs/telemetry/), one should be able to opt-out by setting ` GATSBY_TELEMETRY_DISABLED` environment variable. 

My build script is something like: 

```
export GATSBY_TELEMETRY_DISABLED=1
gatsby build
```
but this doesn't seem working. I still see the following message during the build:
```
╔════════════════════════════════════════════════════════════════════════╗                                                                                                                                                                                                                                                                                                                                                                                                                                         
║                                                                        ║
║   Gatsby has started collecting anonymous usage analytics              ║
║   to help improve Gatsby for all users.                                ║
║                                                                        ║
║   If you'd like to opt-out, you can use `gatsby telemetry --disable`   ║
║   To learn more, checkout https://gatsby.dev/telemetry                 ║
║                                                                        ║
╚════════════════════════════════════════════════════════════════════════╝
```

sidharthachatterjee:
Thank you @pivotal-kaveh 

Currently, when `GATSBY_TELEMETRY_DISABLED` is set to something truthy, telemetry is **not** sent (as expected) but this message is displayed and I see the potential confusion. 

I've fixed this in https://github.com/gatsbyjs/gatsby/pull/13747

PR #13747, opened 2019-04-30
fix(gatsby-telemetry): Set enabled to false in isTrackingEnabled if GATSBY_TELEMETRY_DISABLED

sidharthachatterjee:
Currently, when `GATSBY_TELEMETRY_DISABLED` is set to something truthy (for example, `1`, `true`), we disable telemetry in `src/event-storage.js`

This PR adds an escape hatch _earlier_ in `src/telemetry.js`, a happy side effect of which is a fix for https://github.com/gatsbyjs/gatsby/issues/13714

### DIFF
--- a/packages/gatsby-telemetry/src/telemetry.js
+++ b/packages/gatsby-telemetry/src/telemetry.js
@@ -153,6 +153,12 @@ module.exports = class AnalyticsTracker {
       enabled = true
       this.store.updateConfig(`telemetry.enabled`, enabled)
     }
+
+    // `GATSBY_TELEMETRY_DISABLED` overrides config value
+    if (this.store.disabled) {
+      enabled = false
+    }
+
     this.trackingEnabled = enabled
     return enabled
   }


packages/gatsby-telemetry/src/telemetry.js
```diff
@@ -153,6 +153,12 @@ module.exports = class AnalyticsTracker {
       enabled = true
       this.store.updateConfig(`telemetry.enabled`, enabled)
     }
+
+    // `GATSBY_TELEMETRY_DISABLED` overrides config value
+    if (this.store.disabled) {
+      enabled = false
+    }
+
```
pieh: should this go before https://github.com/gatsbyjs/gatsby/blob/6c17305abe3ef8a8c2cfc83b9c8d730d818089d4/packages/gatsby-telemetry/src/telemetry.js#L149

sidharthachatterjee: Eh, yeah 🤦‍♂ 

Wait, let me clean this up

packages/gatsby-telemetry/src/telemetry.js
```diff
@@ -153,6 +153,12 @@ module.exports = class AnalyticsTracker {
       enabled = true
       this.store.updateConfig(`telemetry.enabled`, enabled)
     }
+
+    // `GATSBY_TELEMETRY_DISABLED` overrides config value
+    if (this.store.disabled) {
```
wardpeet: we still show showAnalyticsNotification so this might still be confusing for users.

freiksenet:
@sidharthachatterjee Ping? Are we good to merge or do you want to address @wardpeet comment?

PR #13749, opened 2019-04-30
docs(gatsby-plugin-netlify): remove comment that states to put it last

smakosh:
## Description

Removes a comment stating to put netlify plugin as last item on the array of plugins
Depends on: https://github.com/gatsbyjs/gatsby/pull/13750


smakosh:
Can't push any other commit to this branch as it's protected

DSchau:
Thank you!

PR #13751, opened 2019-04-30
fix(gatsby-transformer-documentationjs): bump documentationjs to get typescript support

devongovett:
Followup to #13692. This actually adds typescript support.


DSchau:
Thanks @devongovett! We appreciate it!!

PR #13752, opened 2019-04-30
fix SEO component not working when nested

jmau111:
## Description

seems that SEO component cannot work within Layout component, it needs to be outside most of the time.

### DIFF
--- a/starters/default/src/pages/index.js
+++ b/starters/default/src/pages/index.js
@@ -6,16 +6,18 @@ import Image from "../components/image"
 import SEO from "../components/seo"
 
 const IndexPage = () => (
-  <Layout>
+  <>
     <SEO title="Home" keywords={[`gatsby`, `application`, `react`]} />
-    <h1>Hi people</h1>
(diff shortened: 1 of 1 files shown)


jmau111:
> Could you provide an example of how this was/is broken?

I'm just replicating this right now using the starter, meta title, seo etc only appear when I extract SEO component from Layout component.

LekoArts:

I can't reproduce your described error using the default starter. If you're seeing a bug it's best (most of the times) to open an issue first (see [How to file an issue](https://www.gatsbyjs.org/contributing/how-to-file-an-issue/)) and provide a reproduction repository. At this stage the community can investigate if it's really a bug or a user error. After that a PR can be opened 👍 

Of course there can be exceptions to this workflow but I think it's better to first discuss in issues than in PRs.

jmau111:
@LekoArts Sorry for that, ok I'll do it. I think it's probably related the .cache/ folder, pretty weird, I should have tested it a little bit more.

PR #13753, opened 2019-04-30
feat(showcase) : add peintagone website

opeah:
## Description

Adding Peintagone website to the showcase


DSchau:
@opeah thanks for sharing this in Discord. Indeed--an excellent example of what a Gatsby application can be for an end-user experience. Thank you again!

PR #13755, opened 2019-04-30
fix(types): update TypeScript declarations

### LINKED ISSUE #13754: [TS] TypeScript errors updating gatsby 2.3.33 to 2.3.34
## Description

Upon attempting to update the core gatsby package from 2.3.33 to 2.3.34, I ran into a few TypeScript errors introduced in #13619:

```
node_modules/gatsby/index.d.ts(3,29): error TS7016: Could not find a declaration file for module 'express'. 'gatsby-2334-ts/node_modules/express/index.js' implicitly has an 'any' type.
  Try `npm install @types/express` if it exists or add a new declaration (.d.ts) file containing `declare module 'express';`
node_modules/gatsby/index.d.ts(812,15): error TS2300: Duplicate identifier 'Node'.
node_modules/gatsby/index.d.ts(812,15): error TS7031: Binding element 'Node' implicitly has an 'any' type.
node_modules/gatsby/index.d.ts(812,28): error TS2300: Duplicate identifier 'Node'.
node_modules/gatsby/index.d.ts(812,28): error TS7031: Binding element 'Node' implicitly has an 'any' type.
node_modules/gatsby/index.d.ts(869,3): error TS7010: 'addThirdPartySchema', which lacks return-type annotation, implicitly has an 'any' return type.
node_modules/gatsby/index.d.ts(872,5): error TS1016: A required parameter cannot follow an optional parameter.
node_modules/gatsby/index.d.ts(876,3): error TS7010: 'createTypes', which lacks return-type annotation, implicitly has an 'any' return type.
node_modules/gatsby/index.d.ts(879,5): error TS1016: A required parameter cannot follow an optional parameter.
```

1. We need `@types/express` now. Easy enough to add to my project, but maybe it should be a Gatsby dep/devDep? Not sure.
1. ~Gatsby's `Node` interface conflicts with [`Node` from the TypeScript `dom` library](https://github.com/Microsoft/TypeScript/blob/49d6f61298af7a1b470257d293e160694dcd1416/lib/lib.dom.d.ts#L10585). Not sure how to handle this, maybe the declarations just need wrapped in a namespace? @orta or someone will probably know the right:tm: solution.~ (resolved with point 5 below)
1. `addThirdPartySchema` and `createTypes` just need their return types specified (looks like `void` - easy PR, but wanted to figure out these other errors as well).
1. `addThirdPartySchema` and `createTypes` declarations currently flag the `plugin` parameter as optional and the `traceId` parameter as required, but it looks like in reality `plugin` is required and `traceId` is optional (in [both](https://github.com/gatsbyjs/gatsby/blob/dffb74e2c731c7e2778b179cd0932757ace37c64/packages/gatsby/src/redux/actions.js#L1233) [cases](https://github.com/gatsbyjs/gatsby/blob/dffb74e2c731c7e2778b179cd0932757ace37c64/packages/gatsby/src/redux/actions.js#L1335))? Another easy fix if I'm understanding it correctly.
1. `createParentChildLink` signature is slightly incorrect (renaming while destructuring vs. typing).

(Side note: `"skipLibCheck": true` will silence the errors, but that's just a hack.)

### Steps to reproduce

Create a Gatsby+TS project with `"strict": true` in `tsconfig.json` and check the types with `tsc`.

➡️  [Reproduction repository](https://github.com/wKovacs64/gatsby-2334-ts)

### Expected result

Type checking should pass.

### Actual result

Type checking fails.

### Environment

```
System:
  OS: Windows 10
  CPU: x64 Intel(R) Core(TM) i7-7700 CPU @ 3.60GHz
Binaries:
  Yarn: 1.15.2 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
Languages:
  Python: 2.7.15 - /c/Users/user/.windows-build-tools/python27/python
Browsers:
  Edge: 41.16299.1004.0
npmPackages:
  gatsby: 2.3.34 => 2.3.34 
  gatsby-image: 2.0.41 => 2.0.41 
  gatsby-plugin-manifest: 2.0.29 => 2.0.29 
  gatsby-plugin-offline: 2.0.25 => 2.0.25 
  gatsby-plugin-react-helmet: 3.0.12 => 3.0.12 
  gatsby-plugin-sharp: 2.0.35 => 2.0.35 
  gatsby-source-filesystem: 2.0.33 => 2.0.33 
  gatsby-transformer-sharp: 2.1.18 => 2.1.18 
```

wKovacs64:
## Description

- replace Express.js `Application` type import with `any` to avoid `@types/express` dependency
- update `addThirdPartySchema`, `createTypes`, and `createParentChildLink` signatures
- rename `Node` to `GatsbyNode` to avoid collisions with the [`dom` lib](https://github.com/Microsoft/TypeScript/blob/49d6f61298af7a1b470257d293e160694dcd1416/lib/lib.dom.d.ts#L10585)
  - rename `GatsbyNode` to `GatsbyNodeAPI` to allow the above
  - append the `API` suffix to all the other API interfaces for consistency

Plenty to bikeshed here if we want, but this fixes things for now.

## Related Issues

Fixes #13754

### DIFF
--- a/packages/gatsby-source-filesystem/index.d.ts
+++ b/packages/gatsby-source-filesystem/index.d.ts
@@ -1,4 +1,4 @@
-import { GatsbyNode, Store, Cache } from "gatsby"
+import { Node, Store, Cache } from "gatsby"
 
 /**
  * @see https://www.gatsbyjs.org/packages/gatsby-source-filesystem/?=files#createfilepath
@@ -13,7 +13,7 @@ export function createRemoteFileNode(
 ): FileSystemNode
 
 export interface CreateFilePathArgs {
-  node: GatsbyNode
+  node: Node
   getNode: Function
   basePath?: string
   trailingSlash?: boolean
@@ -35,7 +35,7 @@ export interface CreateRemoteFileNodeArgs {
   name?: string
 }
 
-export interface FileSystemNode extends GatsbyNode {
+export interface FileSystemNode extends Node {
   absolutePath: string
   accessTime: string
   birthTime: Date
--- a/packages/gatsby-source-filesystem/index.d.ts
+++ b/packages/gatsby-source-filesystem/index.d.ts
@@ -1,4 +1,4 @@
-import { Node, Store, Cache } from "gatsby"
+import { GatsbyNode, Store, Cache } from "gatsby"
 
 /**
  * @see https://www.gatsbyjs.org/packages/gatsby-source-filesystem/?=files#createfilepath
@@ -13,7 +13,7 @@ export function createRemoteFileNode(
 ): FileSystemNode
 
 export interface CreateFilePathArgs {
-  node: Node
+  node: GatsbyNode
   getNode: Function
   basePath?: string
   trailingSlash?: boolean
@@ -35,7 +35,7 @@ export interface CreateRemoteFileNodeArgs {
   name?: string
 }
 
-export interface FileSystemNode extends Node {
+export interface FileSystemNode extends GatsbyNode {
   absolutePath: string
   accessTime: string
   birthTime: Date
--- a/packages/gatsby/index.d.ts
+++ b/packages/gatsby/index.d.ts
@@ -70,7 +70,7 @@ export const graphql: (query: TemplateStringsArray) => void
  *
  * @see https://www.gatsbyjs.org/docs/gatsby-config/
  */
-export interface GatsbyConfig {
+export interface GatsbyConfigAPI {
   /** When you want to reuse common pieces of data across the site (for example, your site title), you can store that here. */
   siteMetadata?: Record<string, unknown>
   /** Plugins are Node.js packages that implement Gatsby APIs. The config file accepts an array of plugins. Some plugins may need only to be listed by name, while others may take options. */
@@ -103,7 +103,7 @@ export interface GatsbyConfig {
  *
  * @see https://www.gatsbyjs.org/docs/node-apis/
  */
-export interface GatsbyNode {
+export interface GatsbyNodeAPI {
   /**
    * Tell plugins to add pages. This extension point is called only after the initial
    * sourcing and transformation of nodes plus creation of the GraphQL schema are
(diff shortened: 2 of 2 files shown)


pieh:
> * rename `Node` to `GatsbyNode` to avoid collisions with the [`dom` lib](https://github.com/Microsoft/TypeScript/blob/49d6f61298af7a1b470257d293e160694dcd1416/lib/lib.dom.d.ts#L10585)

I'm not TS expert, but this is surprising that definitions from one package cause problems elsewhere. Are definitions part of some global namespace? Can they be namespaced?

wKovacs64:
> I'm not TS expert, but this is surprising that definitions from one package cause problems elsewhere. Are definitions part of some global namespace? Can they be namespaced?

I misinterpreted the duplicate identifier error initially. The collision with the ambient `Node` declaration from `lib` was due to an erroneous `createParentChildLink` signature which was renaming while destructuring instead of typing the destructured properties. The accidental rename to `Node` was the collision, not the type declaration itself. I've reverted that part of the PR, thanks for bringing it up.

cc @orta

pieh:
Oh wow, that's a bit frustrating but I'm glad we probably can get away without renaming typings / using namespaces. I did some light reading on declaration files and seems like namespacing is actually encouraged ( https://www.typescriptlang.org/docs/handbook/declaration-files/library-structures.html#preventing-name-conflicts ), just not sure if adding it now would be breaking change?

orta:
These imports aren't in a 'global declaration file' - they'd be scoped to a library, normally [inside a library](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-native/index.d.ts#L9530-L9551) you'd use `declare global`:

```ts
declare global {
    interface NodeRequire {
        (id: string): any;
    }

    var require: NodeRequire;
   ...
}
```

So, yeah, I don't think anything needs changing. PR looks 👍

wKovacs64:
Thanks all.

pieh:
`gatsby@2.3.35` was published

PR #13234, opened 2019-04-09
docs: Remark Plugin tutorial

lannonbr:
## Description

This is an advanced tutorial on how to create plugins to modify the Abstract Syntax Tree of markdown files parsed through `gatsby-transformer-remark`. This will serve as a reference on how to add additional functionality by manipulating the AST directly.

## Related Issues

Closes #13222

PR #13758, opened 2019-05-01
feat(docs): Adding AST and Remark to the glossary

lannonbr:
## Description

Upon suggestion by @marcysutton in #13234, I went ahead and added `AST` as well as `Remark` to the Glossary page in the docs.

As well since the page is no longer a stub, I removed the asterisk from the yaml entry.

### DIFF
--- a/docs/docs/glossary.md
+++ b/docs/docs/glossary.md
@@ -8,7 +8,7 @@ When you're new to Gatsby there can be a lot of words that seem alien. This glos
 
 ### AST
 
-Abstract Syntax Tree: A tree representation of the source code that is found during a [compiliation](#compiler) step between two languages. For example, [gatsby-transformer-remark](/packages/gatsby-transformer-remark/) will create an AST from [Markdown](#markdown) to describe a Markdown document in a tree structure and then further generate [HTML](#html) using that tree using the [Remark](#remark) parser.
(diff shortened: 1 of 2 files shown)


docs/docs/glossary.md
```diff
@@ -216,6 +220,10 @@ The process of requesting specific data from somewhere. With Gatsby you normally
 
 A code library (written with [JavaScript](#javascript)) for building user interfaces. It’s the framework that [Gatsby](#gatsby) uses to build pages and structure content.
 
+### Remark
+
+A markdown parser to translate [Markdown](#markdown) to other formats like [HTML](#html) or [React](#react) code.
```
lachlanjc:
```suggestion
A parser to translate [Markdown](#markdown) to other formats like [HTML](#html) or [React](#react) code.
```
Prefer to keep the first usage of a term linked & I think "a parser" is clearer anyway?

docs/docs/glossary.md
```diff
@@ -6,6 +6,10 @@ When you're new to Gatsby there can be a lot of words that seem alien. This glos
 
 ## A
 
+### AST
+
+Abstract Syntax Tree: A tree representation of the source code that is found during a [compiliation](#compiler) step between two languages. For example, [gatsby-transformer-remark](/packages/gatsby-transformer-remark/) will create an AST from [Markdown](#markdown) to describe a Markdown document in a tree structure and then further generate [HTML](#html) using that tree using the [Remark](#remark) parser.
```
DSchau:
```suggestion
Abstract Syntax Tree: A tree representation of the source code that is found during a [compilation](#compiler) step between two languages. For example, [gatsby-transformer-remark](/packages/gatsby-transformer-remark/) will create an AST from [Markdown](#markdown) to describe a Markdown document in a tree structure using the [Remark](#remark) parser.
```

lannonbr: Done

Issue #13699, opened 2019-04-29
Repo Size


moonmeister:
## Summary

Cloning the Gatsby repository is becoming a little absurd due to its size. The git clone transfers 550 MB compressed and uncompressed on disk the repo is **854 MB** ( this takes several minutes to clone even on a decent connection). There has been one attempt to fix this in the past(#6486) though I'm not sure if they rewrote git history to purge the large files. 

Getting this reduced would help everyone but would also encourage contributions from countries and areas where slower connections are the norm. At this point the repository size is not helping.

## Relevant information

There are 92 files in the repo over 1mb. 
Compressed (by git) size: 550 MB
Size on Disk: 850+ MB

```
124K    ./examples/no-plugins
172K    ./examples/using-asciidoc
244K    ./examples/hn
316K    ./examples/simple-auth
136K    ./examples/no-trailing-slashes
164K    ./examples/using-contentful
100K    ./examples/using-glamor
1.7M    ./examples/image-processing
100K    ./examples/feed
84K     ./examples/using-styled-jsx
164K    ./examples/using-typescript
156K    ./examples/sitemap
144K    ./examples/using-faker
72K     ./examples/using-styletron
112K    ./examples/using-stylus
5.9M    ./examples/using-remark-copy-linked-files
176K    ./examples/using-emotion-prismjs
168K    ./examples/using-redirects
188K    ./examples/using-page-transitions
88K     ./examples/using-styled-components
120K    ./examples/using-path-prefix
780K    ./examples/using-gatsby-image
2.5M    ./examples/using-sqip
516K    ./examples/using-shopify
148K    ./examples/using-css-modules
176K    ./examples/using-mobx
104K    ./examples/using-sass
108K    ./examples/using-excel
272K    ./examples/styleguide
468K    ./examples/using-type-definitions
180K    ./examples/using-drupal
172K    ./examples/using-wordpress
1004K   ./examples/using-javascript-transforms
100K    ./examples/using-csv
128K    ./examples/client-only-paths
84K     ./examples/using-emotion
132K    ./examples/using-hjson
544K    ./examples/using-jest
164K    ./examples/using-redux
368K    ./examples/using-i18n
144K    ./examples/using-mongodb
88K     ./examples/using-cxs
88K     ./examples/using-medium
332K    ./examples/graphql-reference
156K    ./examples/using-multiple-providers
96K     ./examples/using-page-loading-indicator
14M     ./examples/gatsbygram
[...]

PR #13762, opened 2019-05-01
chore(dev workflow): add pre-commit hook to compress staged images

moonmeister:
## Description

Adds pre-commit hook to the repo using existing husky config to compress `jpg,png,gif` images that have been added to the repository. 

This [package](https://github.com/tomchentw/imagemin-lint-staged/) uses plugins for compression that are lossless. This is good for quality but means images can remain quite large. 

https://web.dev/use-imagemin-to-compress-images - for list of lossless vs lossy plugins.

## Related Issues

#13699

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "fs-extra": "^7.0.0",
     "glob": "^7.1.1",
     "husky": "1.1.1",
+    "imagemin-lint-staged": "^0.4.0",
     "jest": "^24.0.0",
     "jest-cli": "^24.0.0",
     "jest-environment-jsdom-fourteen": "^0.1.0",
@@ -55,6 +56,10 @@
     ],
     "*.svg": [
       "svgo --pretty --indent=2 --config=svgo.yml --multipass",
+      "git add"
+    ],
+    "*.{png,jpeg,jpg,gif}": [
+      "imagemin-lint-staged",
       "git add"
     ]
   },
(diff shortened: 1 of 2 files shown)


moonmeister:
@DSchau Hadn't thought of doing it that way...I think it introduces a couple side effects that might be more troublesome. The primary goal of this is to is to reduce repo size. If we allow people to commit then ask a bot to checkout, compress, and re-commit images then we're duplicating the copies of images in history and significantly increasing repo size. This could be over come if the bot were to rewrite history with the new version of the image. But this seems more fragile and potentially dangerous then doing it with a commit hook. This said I am inexperienced with this kind of git operation, maybe it'd be fine. 

Also, we're already using SVGO to optimize SVGs in commit hooks, this is effectively the same thing but for JPG/PNG/GIF. In fact, the tool I'm using includes SVGO for SVGs but I haven't included them in the husky file pattern matchers as our config is different. 

All this said, you make a good point, this needs to be robust and not prohibit contributions in anyway. Maybe we figure out some ways to test this and see if it causes any troubles. Thoughts here? The one thing I noticed on REALLY large images(2+MB) it can take a while with little to indicate what's happening.

wardpeet:
I agree with @DSchau, also imagemin needs some dependencies like libpng, libjpg, .... How do we feel about adding a lint rule on ci so we get a notice of this.

sidharthachatterjee:
Agree with @wardpeet and @DSchau here that it would be best to avoid problematic hooks in the commit flow 

Closing this for now but definitely think this could be a scheduled GitHub Action perhaps

Thank you for the work though, Alex! 🥇

PR #13768, opened 2019-05-01
fix(gatsby-transformer-documentationjs): re-add ability to get typeDef field from nested type field

pieh:
few things here that can't really fit in PR title:
- actually bump `documentation` package to be able to make use of added TypeScript support
- adjust peerDeps - we are using features from gatsby@2.2 (createTypes)
- added some missing, not quite accurate type definitions (we are doing some post processing on results of `documentation` package)
- actually resolve `typeDef___NODE` fields recursively (in initial PR we lost ability to get typeDef if it was in one of nested fields (like in elements, applications etc)
- use new version of `gatsby-tranformer-documentaitonsj` on .org (this might be split to separate PR, that could be merged after publishing those changes)

For reference - original changes were added in https://github.com/gatsbyjs/gatsby/pull/13692 . I went with approach that doesn't introduce breaking changes to this one


pieh:
This is not ready for merge, there are some weird quirks it seems:
With changes from this PR - on the left, current .org on the right
Screenshot 2019-05-17 at 02 22 07
Seems like there is missing data here

PR #13769, opened 2019-05-01
feat(www): Update banner link title

fk:
„Register now“ -> „Watch now“
/cc @lindaleebumblebee

### DIFF
--- a/www/src/components/__tests__/banner.js
+++ b/www/src/components/__tests__/banner.js
@@ -6,7 +6,7 @@ import Banner from "../banner"
 test(`it renders an external link`, () => {
   const { getByText } = render(<Banner />)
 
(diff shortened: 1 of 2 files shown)


www/src/components/__tests__/banner.js
```diff
@@ -6,7 +6,7 @@ import Banner from "../banner"
 test(`it renders an external link`, () => {
   const { getByText } = render(<Banner />)
 
-  const link = getByText(`Register now`)
+  const link = getByText(`Watch now`)
```
DSchau: Heh - we should fix this. Don't want to tweak a test every time we want to change the text.

fk: 👍

PR #13799, opened 2019-05-02
fix(types): node APIs typings should be in GatsbyNode interface not in Node

### LINKED ISSUE #13798: Major type issue in v2.3.35
This issue is about the TypeScript types shipped by Gatsby.

## Description

As shown [here](https://github.com/gatsbyjs/gatsby/commit/b41755b66bdc6ad671852ce43ba036b52a9a2472#diff-60a650c8008bf341b6e36303e102c9b1R106), a commit was made that made it into v2.3.35 that changed the `GatsbyNode` type's name to just `Node`. Although this was a breaking change for those depending on the types, I thought it would be a simple thing to update and get our codebase compiling again.

However, I was getting a strange error message when I updated the type (`GatsbyNode`) to the new name (`Node`):

> Type '{ createPages: ((args: CreatePagesArgs & { traceId: "initial-createPages"; }, options?: PluginOptions | undefined, callback?: PluginCallback | undefined) => void) | undefined; }' is missing the following properties from type 'Node': id, parent, children, internal. ts(2739)

After looking further into the issue, I realised that Gatsby already exports a separate type called `Node` (that is related to [this](https://www.gatsbyjs.org/docs/node-interface/) kind of node), which can be seen [here](https://github.com/gatsbyjs/gatsby/blob/b41755b66bdc6ad671852ce43ba036b52a9a2472/packages/gatsby/index.d.ts#L1102-L1140). This is what was causing the above error message, as this `Node` type requires the `id`, `parent`, `children` and `internal` properties to be defined.

The likely reason this mistake was allowed to slip through is because of [Declaration Merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) in TypeScript.

To fix this issue, please revert this name change made to this type (`Node` -> `GatsbyNode`).

pieh:
Fixes regression introduced in #13755 

Fixes #13798 (this also has most context around it)

/cc @wKovacs64

### DIFF
--- a/packages/gatsby/index.d.ts
+++ b/packages/gatsby/index.d.ts
@@ -103,7 +103,7 @@ export interface GatsbyConfig {
  *
  * @see https://www.gatsbyjs.org/docs/node-apis/
  */
-export interface Node {
+export interface GatsbyNode {
   /**
    * Tell plugins to add pages. This extension point is called only after the initial
    * sourcing and transformation of nodes plus creation of the GraphQL schema are


pieh:
Published `gatsby@2.3.36`

Issue #13795, opened 2019-05-02
[themes]: Align index query of gatsby-theme-blog-core with blog starter


LekoArts:
## Description

> The list of posts is in reverse order of date. However, the previous and next links from a post follow the order of date (not reverse). I think it's a little confusing. I'm not the only one who experimented the same thing.

From https://github.com/gatsbyjs/gatsby/pull/13726

## How to fix it

Update the query of the theme so that it has the same functionality as the starter.

https://github.com/gatsbyjs/gatsby/blob/616fb24c332b9b32a601ef72c60ca08d42412916/themes/gatsby-theme-blog-core/src/pages/index.js#L34-L50

It also should have the `DESC` like here:

https://github.com/gatsbyjs/gatsby/blob/0d82f2803961f1d443a86b46da311cca9a7c1309/starters/blog/src/pages/index.js#L58-L74

Validate that the change shows the expected result

## How to get started

Read our Contributing guide to get started: https://www.gatsbyjs.org/contributing/code-contributions/

yudhv:
I'll take this issue if you guys assign it to me....

polishedwp:
@yudhv it has already been merged. 

@LekoArts @ChristopherBiscardi Do we need to close this issue now?

PR #13800, opened 2019-05-02
Fix the post queries to match the query from the blog-starter #13795

polishedwp:
## Description

fix the `allMarkdownRemary` query in the theme to match the starter per #13795. 
I did not remove `node.id` from the query in the theme, because it is referenced on line 23.


ChristopherBiscardi:
Thanks!

PR #13818, opened 2019-05-02
fix(blog): update housecall pro post

calcsam:


DSchau:
@calcsam thanks! Also was waiting for the linter to pass on #13817, but I'll close that now!

PR #13826, opened 2019-05-03
feat(gatsby-plugin-feed): add feedUrl plugin option

leonardodino:
Thanks so much for the work on the [`asset-prefix` PR](https://github.com/gatsbyjs/gatsby/pull/12128)! :tada:

I just found a small setback, this PR solves that.

## Description

**use case**:
My project uses a different url for each build (via the just-released `--path-prefix` option), but the feedUrl should be kept stable, as it's user visible.

The built urls are behind a reverse-proxy, and only appear for stable/hashed resourced. For feed readers, the url must continue the same.

## Related Issues

Related to #12128

### DIFF
--- a/e2e-tests/path-prefix/cypress/integration/asset-prefix.js
+++ b/e2e-tests/path-prefix/cypress/integration/asset-prefix.js
@@ -50,7 +50,7 @@ describe(`assetPrefix`, () => {
     })
     it(`keeps RSS feedUrl intact`, () => {
       cy.get(`head link[type="application/rss+xml"]:last`)
-        .should(`have.attr`, `href`)
+        .should(`have.attr`, `href`, `http://localhost:9000/rss-2.xml`)
         .and(`not.matches`, assetPrefixExpression)
     })
   })
--- a/e2e-tests/path-prefix/cypress/integration/asset-prefix.js
+++ b/e2e-tests/path-prefix/cypress/integration/asset-prefix.js
@@ -46,7 +46,12 @@ describe(`assetPrefix`, () => {
 
   describe(`gatsby-plugin-feed`, () => {
     it(`prefixes RSS feed`, () => {
-      assetPrefixMatcher(cy.get(`head link[type="application/rss+xml"]`))
+      assetPrefixMatcher(cy.get(`head link[type="application/rss+xml"]:first`))
+    })
+    it(`keeps RSS feedUrl intact`, () => {
+      cy.get(`head link[type="application/rss+xml"]:last`)
+        .should(`have.attr`, `href`)
+        .and(`not.matches`, assetPrefixExpression)
     })
   })
 })
--- a/e2e-tests/path-prefix/gatsby-config.js
+++ b/e2e-tests/path-prefix/gatsby-config.js
@@ -58,6 +58,29 @@ module.exports = {
             title: `assetPrefix + pathPrefix RSS Feed`,
             output: `rss.xml`,
           },
+          {
+            query: `
+              {
+                pages: allSitePage {
+                  nodes {
+                    path
+                  }
+                }
+              }
+            `,
+            serialize({ query: { site, pages } }) {
+              return pages.nodes.map(node => {
+                return {
+                  description: `A sample page hello world suh dude`,
+                  date: `10-08-1990`,
+                  url: `${site.siteMetadata.siteUrl}${pathPrefix}${node.path}`,
+                }
+              })
+            },
+            title: `feedUrl RSS Feed`,
+            feedUrl: `http://localhost:9000/rss-2.xml`,
+            output: `rss-2.xml`,
+          },
         ],
       },
     },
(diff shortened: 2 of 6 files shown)


e2e-tests/path-prefix/gatsby-config.js
```diff
@@ -58,6 +58,29 @@ module.exports = {
...
+              return pages.nodes.map(node => {
+                return {
+                  description: `A sample page hello world suh dude`,
+                  date: `10-08-1990`,
+                  url: `${site.siteMetadata.siteUrl}${pathPrefix}${node.path}`,
+                }
+              })
+            },
+            title: `feedUrl RSS Feed`,
+            feedUrl: `http://localhost:9000/rss-2.xml`,
```
DSchau: So in this case, output isn't used whatsoever?

e2e-tests/path-prefix/cypress/integration/asset-prefix.js
```diff
@@ -46,7 +46,12 @@ describe(`assetPrefix`, () => {
 
   describe(`gatsby-plugin-feed`, () => {
     it(`prefixes RSS feed`, () => {
-      assetPrefixMatcher(cy.get(`head link[type="application/rss+xml"]`))
+      assetPrefixMatcher(cy.get(`head link[type="application/rss+xml"]:first`))
+    })
+    it(`keeps RSS feedUrl intact`, () => {
```
DSchau: Nice! Thanks for adding these!

e2e-tests/path-prefix/gatsby-config.js
leonardodino: `output` is still used for saving the file to disk (`gatsby-node.onPostBuild`)

DSchau: For clarity/conciseness then, it'd probably be a good idea to not include the output file in feedUrl, and infer it from the existing output?

leonardodino:
Sorry if it wasn't clear, more precisely, I'm using netlify [`DEPLOY_URL`](https://www.netlify.com/docs/continuous-deployment/#environment-variables) as part of my `asset-prefix`. This variable changes for each build. (:

More precisely: `${DEPLOY_URL}/blog-assets`

e2e-tests/path-prefix/gatsby-config.js
leonardodino: that's fits my usecase.
currently my feedUrl is `https://example.org/${output}`.

I left intentionally as a escape-hatch for even more complex use-cases, where an user can opt-out of the magic.

I can rewrite this portion if needed (: (but I think we should change it to something like `baseUrl` as it won't be the final one in the DOM)

DSchau:
@leonardodino it seems like there'd be better ways to remediate outside of adding in additional options to the plugin, however.

Could you share more detail on your deploy, or even a link to the repository, itself?

As I understand the problem, you're using Netlify's `DEPLOY_URL` environment variable, which will change for each and every deploy as your assetPrefix. Couldn't you just not use that? I'm unclear exactly as to the _why_ of this PR quite yet--so an example would be very helpful!

e2e-tests/path-prefix/gatsby-config.js
DSchau: I think `siteUrl` is typically used for this--right? i.e. `siteMetadata.siteUrl` refers to the deployed URL of your application.

DSchau: (not suggesting you change the name--just an observation!)

leonardodino:
Hey @DSchau I will have to consult whether people feel ok with sharing the complete example.

I agree it's better to not over-complicate core plugins.

We are using some complex netlify rules, serving many "smaller" sites stitched together, the total build time for everything under the `www.` domain would take a good portion of a day.

Due to limitations on netlify's redirect system, especially regarding nested redirects and caching, this was the solution we found, to have files linked on another domain, instead of relying on the `/` path.

LekoArts:
Hi @leonardodino 	👋 	Did you find the time to build an example/collect details on your deploy? Currently I also feel like this is an edge case / quite don't understand the use case unfortunately :/

leonardodino:
Hey @LekoArts, thanks for triaging this issue! (:

Unfortunately I don't have much more info, besides what [I said in the last comment](https://github.com/gatsbyjs/gatsby/pull/13826#issuecomment-489102102).

I pinged the team to see if there's any more details behind the explanation.

---

I believe my case is a bit more general than just a feature for the `feed` plugin.

re-reading one of the @DSchau questions:
> I don't see what this is doing exactly, apart from making it so that assetPrefix isn't really being respected for this asset?

I figured out that a feed is not actually an asset, it's a user-visible URL.

Example: when compressing images we don't care that much about the URL, it will be cited in the HTML for the browser to fetch, it can be long and random.
But, a feed is actually more of a route, that is directly accessed by the feed reader, so it's important that is constant for a website.

leonardodino:
**UPDATE**: You may close this PR now.
I still think it's a valid option, but now there's a way around it. :tada:

I've "hacked" the `match` option in the adjacent PR (#13827) to never match,
so I can emit the `<link>` tag myself, using `react-helmet`.

(only saw that other PR today, it happens that we both opened them only hours apart.)

LekoArts:
@leonardodino 	How about closing this PR (to keep the history of the changes clean) and you could open up a new PR to document your usage in the README?

PR #13827, opened 2019-05-03
feat(gatsby-plugin-feed): add match plugin option

laysent:
## Description

This PR is to propose a way to only include RSS feed in part of website, not all pages of it. `gatsby-plugin-feed` allows to customize the generated feed to only include part of the contents. However, the generated feed will be included in all pages by default, which might not be necessary in some cases.

It will be useful when for example blog is only part of the website, and there are many other pages which are not related to blog posts; or when there are different RSS feeds in website and each belongs to different part.

Adding an extra field `match` in feed array provides the ability to control where feed should be inserted when generating SSR result.

## Related Issues

No issue related

### DIFF
--- a/docs/docs/adding-an-rss-feed.md
+++ b/docs/docs/adding-an-rss-feed.md
@@ -130,6 +130,8 @@ This snippet contains a custom `gatsby-plugin-feed` setup in `gatsby-config.js` 
 
 The `output` field in your feed object allows you to customize the filename for your RSS feed, and `title` for the name of your site's RSS feed.
 
+By default, feed is referenced in every page. You can customize this behavior by providing an extra field `match` of type `string`. This string will be used to build a `RegExp`, and this regular expression will be used to test the `pathname` of current page. Only pages that satisfied the regular expression will have feed reference included.
+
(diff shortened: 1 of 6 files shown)


laysent:
@DSchau Hi, I noticed that this PR is labeled as `awaiting author response`. May I know if there is anything I can do here?

LekoArts:
I think Dustin accidentally chose the wrong label so I think there's nothing needed from you at the moment. The team is out of office right now, so please have remind us of this PR next week. Thanks!

wardpeet:
@laysent is there any reason why you chose a string and not directly went for a regex as an option?

laysent:
@wardpeet I tried RegExp before, however it seems the RegExp option configured in `gatsby-config.js` cannot be passed to `gatsby-plugin-feed`. Looking at `api-runner-ssr.js` file generated under `.cache` folder, I found that RegExp provided in option has been transfered to `{}` there. Thus I then chose to use string instead, as it won't be transformed.

laysent:
@LekoArts Hi, just to remind for this PR

PR #13841, opened 2019-05-03
feat(gatsby-remark-embed-snippet): let gatsby-remark-prismjs do the heavy lifting

timhagn:
## Description

While writing a documentation page for `gbi`, I noticed that line numbering and some other options of `gatsby-remark-prismjs` weren't working here. Stepping through it with Intellij Idea's Debugger confirmed that the [`visit(markdownAST, `code`, node => {...}`](https://github.com/gatsbyjs/gatsby/blob/f4e8d618e4cc77624df9b3d1779b2dcac2fdfcc5/packages/gatsby-remark-prismjs/src/index.js#L22) branch of it was of course never reached, as this plugin already had processed it to HTML. 
So I changed the constructed `node.type` to `code`, just gave the read in file as `node.value` and added `node.lang` to the returned node, which are needed for further processing down the line.
It just seemed unnecessary to duplicate `gatsby-remark-prismjs`'s code-highlighting functionality by creating _nearly_ the same output (and it should prevent any future deviations).

## Related Issues
Fixes the issue I would have opened otherwise ; ).

### DIFF
--- a/packages/gatsby-remark-embed-snippet/README.md
+++ b/packages/gatsby-remark-embed-snippet/README.md
@@ -177,15 +177,6 @@ module.exports = {
           {
             resolve: "gatsby-remark-embed-snippet",
             options: {
-              // Class prefix for <pre> tags containing syntax highlighting;
-              // defaults to 'language-' (eg <pre class="language-js">).
-              // If your site loads Prism into the browser at runtime,
-              // (eg for use with libraries like react-live),
-              // you may use this to prevent Prism from re-processing syntax.
-              // This is an uncommon use-case though;
-              // If you're unsure, it's best to use the default value.
-              classPrefix: "language-",
-
               // Example code links are relative to this dir.
               // eg examples/path/to/file.js
               directory: `${__dirname}/examples/`,
--- a/packages/gatsby-remark-embed-snippet/README.md
+++ b/packages/gatsby-remark-embed-snippet/README.md
@@ -167,7 +167,7 @@ function App() {
 **Important**: This module must appear before `gatsby-remark-prismjs` in your
 plugins array, or the markup will have already been transformed into a code
 block and this plugin will fail to detect it and inline the file.
-For further information about it's options, visit `gatsby-remark-prismjs`'s
+For further information about its options, visit `gatsby-remark-prismjs`'s
 [README](https://www.gatsbyjs.org/packages/gatsby-remark-prismjs/).
 
 ```js
--- a/packages/gatsby-remark-embed-snippet/README.md
+++ b/packages/gatsby-remark-embed-snippet/README.md
@@ -164,7 +164,11 @@ function App() {
 
 ## How to use
 
-**Important**: This module must appear before `gatsby-remark-prismjs` in your plugins array, or the markup will have already been transformed into a code block and this plugin will fail to detect it and inline the file.
+**Important**: This module must appear before `gatsby-remark-prismjs` in your
+plugins array, or the markup will have already been transformed into a code
+block and this plugin will fail to detect it and inline the file.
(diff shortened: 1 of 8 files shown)


timhagn:
One question I forgot above: Do we need the `try ... catch` code block around the `node.` processing any more? As no function or anything outside is called...

packages/gatsby-remark-embed-snippet/README.md
```diff
@@ -164,7 +164,11 @@ function App() {
 
 ## How to use
 
-**Important**: This module must appear before `gatsby-remark-prismjs` in your plugins array, or the markup will have already been transformed into a code block and this plugin will fail to detect it and inline the file.
+**Important**: This module must appear before `gatsby-remark-prismjs` in your
+plugins array, or the markup will have already been transformed into a code
+block and this plugin will fail to detect it and inline the file.
+For further information about it's options, visit `gatsby-remark-prismjs`'s
```
KyleAMathews: its

timhagn: \*miff\* I could pull the "non-native speaker card", but that one's too obvious \*facepalm\*
#corrected

timhagn:
Damnit, tests failing cause of 
```
found 2 low severity vulnerabilities in 26416 scanned packages
run `npm audit fix` to fix 2 of them.
```
: /

timhagn:
Never mind, I just forgot to sync upstream with master and merge it % ).

timhagn:
Ah, haven't read your question pertaining testing it on reactjs.org, @KyleAMathews. Where should I start, just clone their repo?

KyleAMathews:
Yeah clone their repo and then use gatsby-dev-cli to copy over your source (or manually do it). See the instructions for gatsby-dev-cli here: https://www.gatsbyjs.org/contributing/setting-up-your-local-dev-environment/#gatsby-repo-install-instructions

timhagn:
Thanks, though I know my way around `gatsby-dev-cli` ; ). Have looked at its sources and already starred it ^^. Will do, hope it doesn't break in reactjs.org - although, as `gatsby-remark-prismjs` nearly returns the same HTML as this plugin plugin (quite recursive, how much deeper can we get ; ) did, only with `-prismjs`'s options in effect. Gonna try it and report.

timhagn:
\*miff\* reactjs.org seems to depend on the HTML creation of this plugin, line highlighting doesn't work at the moment. Got it working with my documentation site, though (through adding a `{highlight: true}` option object to "language"), so I'm gonna investigate this further and perhaps build a switch for "legacy-behavior". Am keen what @bvaughn has to say, cause I think this plugin should really be "in canon" with the rest of the prismjs chain.

timhagn:
Ah, OK, should have looked at it in the first place - reactjs.org is still using `gatsby-remark-prismjs` at `v3.0.2`, changing that to the current `v3.2.9` correctly highlights lines, without adding any options to the language. So everything seems to work as intended (and reactjs.org could add line numbers ; ).

timhagn:
I accidentally used the automatic rebase instead of merge of IntelliJ Idea while syncing with the latest upstream/master and couldn't revert it (probably should do it on the CLI again). Does this pose a problem and should I open a new (cleaned up) PR - or are there `git` tricks (like squashing or such) which could still rectify this faux pas?

marcysutton:
@timhagn while we do squash PRs, the mixed commit history makes this very hard to review. Can you create a new PR against master limited to your changes so we can review it more effectively?

timhagn:
Hi @marcysutton will do : ).

timhagn:
@marcysutton @KyleAMathews @DSchau Created a new PR from a clean master with all my previous changes over at #13973 : ).

PR #13845, opened 2019-05-03
fix(www): use end of day, use correct iso timestamp

### LINKED ISSUE #13844: Events Page: Missing content after loading on iPhone
## Description

Events Page: Missing content after full page load complete on iPhone

### Steps to reproduce

i open the gatsby [community events](https://www.gatsbyjs.org/contributing/events/) page on iPhone 6S+ (iOS 12.1.3) 
i tested it in Safari app and Chrome app

### Expected result

- i see the full events list

### Actual result

- while loading i see the full events list
- after loading complete the events are empty

DSchau:
## Description

Dates are fun. This seems to do the trick, specifically:

1. Uses a sort with GraphQL instead of client-side JS
1. Uses _end_ of day instead of start of day
    - If there's something obviously wrong or better here--happy to tweak
1. Keep date objects in data store, leave formatting up to React component

This fixes a few issues, namely that a date like `2019-05-03 00:00:00` (`new Date('2019-05-03 00:00:00')` is an invalid date in some browsers (like Safari) and so was filtering _all_ the events to not show once it hits the browser.

## Related Issues

Fixes #13844

### DIFF
--- a/www/src/components/events/__tests__/events.js
+++ b/www/src/components/events/__tests__/events.js
@@ -1,0 +1,44 @@
+import React from "react"
+import { render } from "react-testing-library"
+import { useStaticQuery } from "gatsby"
+
+import Events from "../events"
+
+const toEvent = (date, id) => {
+  return {
+    id,
+    data: {
+      date: date,
+      type: `event`,
+      name: `something something`,
+    },
+  }
+}
+
+const mockEvents = events =>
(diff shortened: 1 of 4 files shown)


DSchau:
Going to merge this, because:

1. I've confirmed it fixes a bug
1. It has unit tests
1. YOLO

🚀

PR #13852, opened 2019-05-04
fix(gatsby-plugin-sharp): ignore sizeByPixelDensity option and add deprecation warning

### LINKED ISSUE #12743: [gatsby-plugin-sharp] Density should be ignored for non-SVG images
We saw a few reports of itty bitty images come in from @joelhooks and @kentcdodds. The `max-width` from `gatsby-remark-images` was coming out with tiny values like `38px`, which, of course, is bad.

Digging into this, @ChristopherBiscardi and I (with an assist from the all-knowing @pieh) traced the issue back to `gatsby-plugin-sharp` and this code:
https://github.com/gatsbyjs/gatsby/blob/74e1063f57916f04b853f7e5aae466ce72a2ff73/packages/gatsby-plugin-sharp/src/index.js#L380-L383

Further sleuthing from Chris found a couple comments from @lovell (https://github.com/lovell/sharp/issues/967#issuecomment-333180519 and https://github.com/lovell/sharp/issues/823#issuecomment-332742955) that say the density option isn't supported for JPG/PNG.

And... since we only process JPG/PNG with `gatsby-plugin-sharp`, we should probably drop the `sizeByPixelDensity` option altogether.

At minimum, we should ignore the option when `format` is non-vector.

For now we've opened PRs for egghead (https://github.com/eggheadio/how-to-egghead/pull/46) and Kent (https://github.com/kentcdodds/kentcdodds.com/pull/108) to disable the option, but we should probably come up with a better solution that "don't use it". 😅

jlengstorf:
fixes #12743

### DIFF
--- a/packages/gatsby-plugin-sharp/README.md
+++ b/packages/gatsby-plugin-sharp/README.md
@@ -107,10 +107,11 @@ fit strategies `CONTAIN` and `FILL` will not work when `cropFocus` is assigned t
 - `maxWidth` (int, default: 800)
 - `maxHeight` (int)
 - `quality` (int, default: 50)
-- `sizeByPixelDensity` (bool, default: false)
 - `srcSetBreakpoints` (array of int, default: [])
 - `fit` (string, default: '[sharp.fit.cover][6]')
 - `background` (string, default: 'rgba(0,0,0,1)')
+- [deprecated] `sizeByPixelDensity` (bool, default: false)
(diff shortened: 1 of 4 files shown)


jlengstorf:
@KyleAMathews ah, good call. Just added a note to the README!

jlengstorf:
Are these test failures due to obsolete snapshots? I didn't see an actual failure, but I also can't seem to run `yarn bootstrap` on my machine right now (libvips errors! yay!) to try this locally.

KyleAMathews:
Yeah, pretty sure "obsolete" means Jest knows the tests aren't run anymore. Since you can't run them locally, you could just try editing the snapshot directly.

PR #13864, opened 2019-05-05
docs: Add variable usage to Gatsby Node API docs

BrunoScheufler:
## Description

Recently I stumbled over a use case where I had to include dynamic variables in my page creation queries for Gatsby. After scanning the existing documentation for an entry, which I couldn't find, I updated the existing Gatsby Node API documentation using a variable for the entry limit and also updated the `allMarkdownRemark` query to use the current frontmatter instead of fields, which is no longer available on this type. 

## Related Issues

n/a


muescha:
maybe update or append also the docs at:

https://www.gatsbyjs.org/docs/creating-and-modifying-pages/

https://www.gatsbyjs.org/docs/node-apis/

BrunoScheufler:
@muescha I'd love to but I think especially for the second page that I already edited the right file as I couldn't find another one with that exact content. I'm not sure how the docs are created for those pages though, so maybe I'm missing something here 🤔

muescha:
I see it is generated

PR #13871, opened 2019-05-05
fix(gatsby): refresh browser if webpack rebuild occurs

Moocar:

## Description

The is the final new piece of functionality in https://github.com/gatsbyjs/gatsby/pull/13004. It covers the case where a user is browsing the site, and a rebuild occurs in the background. In the old `data.json` world, the user would never see the new changes until they manually refreshed, since everything is statically linked. But in the de-globalized approach, the user might click a link after a rebuild occurs and get a page-data.json that is incompatible with the browser's component code.

To avoid this, we save the webpack `stats.hash` which is a hash of the entire webpack build. So whenever any source code in the site changes, there will be a new hash. This is then added to every page-data and html file, so that when the browser loads a new page, it can compare the html's hash to the page's hash, and force a refresh if they're different.

Since the page-datas now rely on the webpack compilation hash, I've moved query running out of bootstrap and into `build.js` so that the page querys can be run _after_ the javascript bundle has been built.

## Related Issues


Moocar:
@KyleAMathews @pieh @wardpeet This is the last big PR in the series. Would love a review when you get a chance.

I'm looking into the `e2e_tests_production_runtime` test. They're all passing on my machine (TM) so not sure what's going on yet.

Moocar:
Ahh, I see now, it's the `yarn test:offline` variant. Thank goodness for these integration tests.

Moocar:
ok, tests fixed. I was stuck by the interdependencies between `1-production.js` and the rest of the cypress tests in `e2e-tests/production-runtime`. Should be working now. @davidbailey00 would you mind having a look at the code around resetting the whitelist in [navigation.js](https://github.com/gatsbyjs/gatsby/pull/13871/files#diff-6747c31f3f017f3a91c43ad8b7762d36R93)? Not sure if this code should be here or in the offline plugin. Also, I know you have some WIP to improve the offline plugin, so feel free to rewrite all this stuff after it's merged.

packages/gatsby/src/utils/page-data.js
```diff
@@ -1,28 +1,54 @@
...
   }
   await fs.outputFile(filePath, JSON.stringify(body))
 }
 
+const updateCompilationHashes = (
+  { publicDir },
+  pagePaths,
+  webpackCompilationHash
+) => {
+  const workerPool = new Worker(require.resolve(`./page-data-worker`), {
```
KyleAMathews: I'm surprised a big that this makes things faster? What was the difference? Node does use a thread pool for FS. Hmmm looking at this https://stackoverflow.com/questions/20346097/does-node-js-use-threads-thread-pool-internally there is limits so for mass updates I can see the limits on fs threads causing trouble.

KyleAMathews:
@Moocar lemme know about my question but otherwise I'll merge this in later.

packages/gatsby/src/utils/page-data.js
Moocar: So, I haven't actually run benchmarks on this, but my thinking is that the JSON deserialization/serialization is computationally expensive, so running that work on multiple workers will result in a performance increase. Especially since no data needs to be passed between processes (except the file path). Basically, this is a CPU optimization, not an I/O one.

KyleAMathews: Yeah — seems reasonable — does this reuse the workers for rendering pages? It'd be nice not to spawn extra Node.js processes if we don't have to.

Moocar: No, it's a different worker pool. Interesting idea on sharing them. The challenge is that we'd have to write a new `worker.js` that references the code for both html rendering and this code, and then export the functions for each. That's totally doable, but might add a bit too much complexity for now. I'd say that if we use jest-worker in any other parts of gatsby, it might make sense to consolidate at that point.

packages/gatsby/cache-dir/__tests__/__snapshots__/static-entry.js.snap
```diff
@@ -6,8 +6,8 @@ exports[`develop-static-entry onPreRenderHTML can be used to replace postBodyCom
...
+exports[`static-entry onPreRenderHTML can be used to replace headComponents 1`] = `"<!DOCTYPE html><html><head><meta charSet=\\"utf-8\\"/><meta http-equiv=\\"x-ua-compatible\\" content=\\"ie=edge\\"/><meta name=\\"viewport\\" content=\\"width=device-width, initial-scale=1, shrink-to-fit=no\\"/><link as=\\"fetch\\" rel=\\"preload\\" href=\\"/page-data/about/page-data.json\\" crossorigin=\\"use-credentials\\"/><style> .style3  </style><style> .style2  </style><style> .style1  </style><meta name=\\"generator\\" content=\\"Gatsby 2.0.0\\"/></head><body><noscript id=\\"gatsby-noscript\\">This app works best with JavaScript enabled.</noscript><div id=\\"___gatsby\\"><div style=\\"outline:none\\" tabindex=\\"-1\\" role=\\"group\\"></div></div><script id=\\"gatsby-script-loader\\">/*<![CDATA[*/window.pagePath=\\"/about/\\";window.webpackCompilationHash=\\"undefined\\";/*]]>*/</script><script id=\\"gatsby-chunk-mapping\\">/*<![CDATA[*/window.___chunkMapping={};/*]]>*/</script></body></html>"`;
```
wardpeet: bikeshed: do we want to mock this value `window.webpackCompilationHash=\\"undefined\\"` just to make sure it's not undefined 🤷‍♂

Moocar: Not a bikeshed at all. Fixed.

Moocar:
Hmm, not sure about that production-runtime error. Can't reproduce it locally.

packages/gatsby/src/utils/page-data.js
wardpeet: this sounds a lot like a "job" api that handles things to a worker pool.

packages/gatsby/cache-dir/navigation.js
```diff
@@ -82,6 +82,28 @@ const navigate = (to, options = {}) => {
...
   loader.loadPage(pathname).then(pageResources => {
+    // If the loaded page has a different compilation hash to the
+    // window, then a rebuild has occurred on the server. Reload.
+    if (process.env.NODE_ENV === `production` && pageResources) {
+      if (
+        pageResources.page.webpackCompilationHash !==
+        window.___webpackCompilationHash
+      ) {
+        // Purge plugin-offline cache
+        if (
```
wardpeet: seems like I didn't save my comment. This feels a bit hacky IMO but good enough to go forward with this one.

IMO it feels like we need to create a gatsby-browser api for this so plugins can handle this on their behalf. Now it looks like that core & offline plugin are intertwined. Maybe just put up a TODO comment?

wardpeet:
@Moocar maybe just merge master? Didn't want to use the bot to make sure I don't break anything.

Circle ci:
> Your plugins must export known APIs from their gatsby-browser.js.
The following exports aren't APIs. Perhaps you made a typo or your plugin is outdated?

> See https://www.gatsbyjs.org/docs/browser-apis/ for the list of Gatsby Browser APIs

> - The plugin "gatsby-plugin-offline@2.1.0" is exporting a variable named "onPostPrefetchPathname" which isn't an API.

packages/gatsby/src/utils/page-data.js
KyleAMathews: @Moocar we'd just have to merge the two files right and have different exports for the two? Am I missing something? Doesn't seem too complex.

Moocar: @KyleAMathews Yep, and then make sure we start and end the worker pool at the right place. More than happy to code this up for this PR.

KyleAMathews: Thanks, let's do that — this will make things more efficient + it'll make it really easy to add the worker pool to other tasks as they come up as we'll have the basic infra in place + since the workers are already spun up, sending task to them is really cheap.

packages/gatsby/cache-dir/navigation.js
KyleAMathews: I don't like intertwining Gatsby and gatsby-plugin-offline. But adding a new API with only one known use case is _also_ problematic. Worse IMO because APIs are forever while hacks can be removed pretty quickly. This seems pretty harmless all in all and if there's ever another use case that comes up for telling people when the webpack compilation hash changes, we could add an API and move this code to gatsby-plugin-offline.

KyleAMathews:
:shipit:

PR #13879, opened 2019-05-06
chore(docs): Clarify language codename in Kentico tutorial

CollierCZ:
## Description

Clarified what language codename was being used after some issues getting started.

### DIFF
--- a/docs/docs/sourcing-from-kentico-cloud.md
+++ b/docs/docs/sourcing-from-kentico-cloud.md
@@ -39,7 +39,7 @@ Next, install the [Kentico Cloud source plugin](https://github.com/Kentico/gatsb
 npm install --save gatsby-source-kentico-cloud
 ```
 
-Once that's done, you need to add the plugin to `gatsby-config.js` (note this uses the language codename from the Sample Project, which would be `default` in a blank project):
+Once that's done, you need to add the plugin to `gatsby-config.js`:
 
 ```javascript
 module.exports = {
@@ -53,8 +53,9 @@ module.exports = {
           deliveryClientConfig: {
             projectId: `<YourProjectID>` // Fill in your Project ID
           },
+          // Please note that with the Sample Project generated above, `en-US` is the default language for the project and this config. For a blank project, this needs to be `default`.
           languageCodenames: [
-                    `en-US` // Or whatever languages you have installed in your project (Project settings -> Localization)
+                    `en-US` // Or the languages in your project (Project settings -> Localization)
           ]
       }
     },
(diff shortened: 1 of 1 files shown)


docs/docs/sourcing-from-kentico-cloud.md
```diff
@@ -39,7 +39,7 @@ Next, install the [Kentico Cloud source plugin](https://github.com/Kentico/gatsb
 npm install --save gatsby-source-kentico-cloud
 ```
 
-Once that's done, you need to add the plugin to `gatsby-config.js`:
+Once that's done, you need to add the plugin to `gatsby-config.js` (note this uses the language codename from the Sample Project, which would be `default` in a blank project):
```
DSchau: I'm not sure this meaningfully clarifies things--could you explain how this is helpful?

In general - it seems like it raises a question, e.g. why would I, as someone who wants to source from Kentico Cloud, need this info?

CollierCZ: I added this info because someone was following the tutorial with a different project than the suggested Sample Project. They did not get their content because they overlooked the need to switch the language codename from `en-US` to `default`. So I tried to add context here to help people see that detail.

LekoArts: Follow-up question: Why is `en-US` used then in this example configuration / tutorial? Shouldn't the language you use in your project automatically be the `default` ?

CollierCZ: In new/blank projects, the language is `default` unless you change it. For this tutorial, we used a Sample Project that already has content it in (to more quickly see the results/data structure). That Sample Project has content in English and Spanish.

LekoArts: Mh, I see. I'd want to move the comment to above line 56 (because I assume you then can delete the languageCodename option?)

CollierCZ: At the moment, the [languageCodenames property is required](https://github.com/Kentico/gatsby-source-kentico-cloud#install-to-your-existing-gatsby-project) and can't be deleted.

LekoArts: You're not mentioning `default` either? Shouldn't the property be optional and if not provided fallback to `default`. Well, that's probably besides the point of this discussion. I'd be fine with moving the comment to above the `languageCodenames` with something like:

> Please note that with using the "Sample Project generator" this config defines "en-US" as its default language. In a blank project this needs to be "default"

CollierCZ: I opened an [issue in the source plugin](https://github.com/Kentico/gatsby-source-kentico-cloud/issues/58). Also created a [pull request for moving the comment](https://github.com/gatsbyjs/gatsby/pull/13899) to this branch (as it's protected).

PR #13882, opened 2019-05-06
use slash2 instead of slash

woodgear:
## Description
origin slash will return origin input if it contains noneasci [code here](https://github.com/sindresorhus/slash/blob/262981c1fca19193a20657ac99690214d9483d0e/index.js#L6),but path could contains NonAscii,
for example slash("foo\\测试") will return "foo\\测试" instead of “foo/测试”
slash2 remove this limit

### DIFF
--- a/packages/gatsby-source-filesystem/src/create-file-node.js
+++ b/packages/gatsby-source-filesystem/src/create-file-node.js
@@ -1,4 +1,4 @@
-const slash = require(`slash`)
+const slash = require(`slash2`)
 const path = require(`path`)
 const fs = require(`fs-extra`)
 const mime = require(`mime`)
--- a/packages/gatsby-source-filesystem/package.json
+++ b/packages/gatsby-source-filesystem/package.json
@@ -19,7 +19,7 @@
     "pretty-bytes": "^4.0.2",
     "progress": "^1.1.8",
     "read-chunk": "^3.0.0",
-    "slash": "^1.0.0",
+    "slash2": "^2.0.0",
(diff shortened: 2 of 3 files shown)


packages/gatsby-source-filesystem/src/create-file-node.js
```diff
@@ -1,4 +1,4 @@
-const slash = require(`slash`)
+const slash = require(`slash2`)
```
poyea: I think we still require `slash`, even for `slash2`.

sidharthachatterjee: The author has forked `slash`, added their fix and published as `slash2`

sidharthachatterjee:
Not sure why tests are failing here

If I understand correctly, this fixes handling of paths with non ASCII characters. Can you try opening a PR to the original `slash` repo? I'd prefer not to have to maintain a fork if this can go upstream!

woodgear:
there is a issue about this,[Not working with Unicode](https://github.com/sindresorhus/slash/issues/5),but the author close that issue

woodgear:
sounds great,i will open a new PR to do that

woodgear:
@freiksenet here is the new PR #14372

ewirth:
I have the same problem with Hungarian unicode letters like: á, ő, ű.
When a url name consists one of them, the slashes are converted to blackslashes, and the page links are broken.

superhawk610:
Replaced by #14372.

PR #13904, opened 2019-05-07
fix(gatsby-plugin-catch-links): Fall back to default browser link handling when resources fail to fetch

### LINKED ISSUE #13903: [gatsby-plugin-catch-links] - Missing resources cause links to not work
## Description

When resources fail to fetch, causing the initial hydration of the app to fail, gatsby-plugin-catch-links continues to call `navigate()` despite nothing being able to handle this. This results in broken links which cannot be used to navigate.

### Steps to reproduce

- Create a project using gatsby-plugin-catch-links, with a page linking to another without using `<Link />` but instead using `<a />`
- Build the site and then remove / rename / block some resource (e.g. anything `component---path-to-component-<hash>.js` seems to work).
- Visit a page which uses this resource and try to navigate from one page to another by clicking the aforementioned `<a />` element.

### Expected result

Despite the app failing to hydrate, the browser should be able to navigate from one page to another, falling back on the default non-SPA behaviour.

### Actual result

The click event is swallowed and the default behaviour prevented. No navigation occurs.

### Environment

```
  System:
    OS: Linux 4.15 Ubuntu 18.04.2 LTS (Bionic Beaver)
    CPU: x64 Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
    Shell: 5.4.2 - /usr/bin/zsh
  Binaries:
    Node: 10.15.3 - /tmp/yarn--1557232001187-0.10878563852082901/node
    Yarn: 1.15.2 - /tmp/yarn--1557232001187-0.10878563852082901/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v10.15.3/bin/npm
  Languages:
    Python: 2.7.15 - /usr/bin/python
  Browsers:
    Chrome: 73.0.3683.103
    Firefox: 66.0.3
  npmPackages:
    gatsby: ^2.1.18 => 2.1.18 
    gatsby-plugin-catch-links: ^2.0.5 => 2.0.5 
    gatsby-plugin-google-analytics: ^2.0.10 => 2.0.10 
    gatsby-plugin-less: ^2.0.11 => 2.0.11 
    gatsby-plugin-sharp: ^2.0.7 => 2.0.7 
    gatsby-plugin-styled-components: ^3.0.0 => 3.0.0 
    gatsby-remark-autolink-headers: ^2.0.7 => 2.0.7 
    gatsby-remark-copy-linked-files: ^2.0.5 => 2.0.5 
    gatsby-remark-images: ^2.0.4 => 2.0.4 
    gatsby-remark-katex: ^3.0.1 => 3.0.1 
    gatsby-remark-prismjs: ^3.0.2 => 3.0.2 
    gatsby-remark-relative-images: ^0.2.0 => 0.2.0 
    gatsby-remark-responsive-iframe: ^2.0.5 => 2.0.5 
    gatsby-source-filesystem: ^2.0.5 => 2.0.5 
    gatsby-transformer-remark: ^2.1.8 => 2.1.8 
```

liamness:
## Description

Essentially disables gatsby-plugin-catch-links if resources fail to fetch, as the resulting runtime error seems to prevent the router from responding to navigation events.

## Related Issues

Fixes #13903 

However would not fix links being broken by some other failure to hydrate or bootstrap the application. I'm also wary that I might be reaching into gatsby's internals with this approach, am open to ideas about how to achieve this fix using only public APIs.

### DIFF
--- a/packages/gatsby-plugin-catch-links/src/__tests__/catch-links.js
+++ b/packages/gatsby-plugin-catch-links/src/__tests__/catch-links.js
@@ -286,7 +286,7 @@ describe(`navigation is routed through gatsby if the destination href`, () => {
       view: window,
     })
 
-    hrefHandler.mockImplementation(() => {
+    setTimeout(() => {
       expect(hrefHandler).toHaveBeenCalledWith(
         `${sameOriginAndTopPath.pathname}`
       )
@@ -313,7 +313,7 @@ describe(`navigation is routed through gatsby if the destination href`, () => {
       view: window,
     })
 
-    hrefHandler.mockImplementation(() => {
+    setTimeout(() => {
       expect(hrefHandler).toHaveBeenCalledWith(
         `${withAnchor.pathname}${withAnchor.hash}`
       )
@@ -340,7 +340,7 @@ describe(`navigation is routed through gatsby if the destination href`, () => {
       view: window,
     })
 
-    hrefHandler.mockImplementation(() => {
+    setTimeout(() => {
       expect(hrefHandler).toHaveBeenCalledWith(
         `${withSearch.pathname}${withSearch.search}${withSearch.hash}`
       )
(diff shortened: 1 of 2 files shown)


packages/gatsby-plugin-catch-links/src/__tests__/catch-links.js
```diff
@@ -340,7 +340,7 @@ describe(`navigation is routed through gatsby if the destination href`, () => {
       view: window,
     })
 
-    hrefHandler.mockImplementation(() => {
+    setTimeout(() => {
```
liamness: Changed these to `setTimeout` as otherwise there would be a discrepancy between this and the new test, which could have explained the difference in behaviour.

liamness:
Hmm. Not getting that failure locally. I'll see if I can figure out what's going on.

liamness:
@sidharthachatterjee I don't really understand why that helped (if `___failedResources` is undefined that would also be falsy?) but thank you

pieh:
> @sidharthachatterjee I don't really understand why that helped (if `___failedResources` is undefined that would also be falsy?) but thank you

This is some weird node 6 quirk, where it seems `delete` (maybe with combination with globals) don't really do what we expect it to do. We've seen this a couple times in our tests failing in similar scenarios in node 6 environment.

liamness:
Yep, definitely a thing facebook/jest#3152

packages/gatsby-plugin-catch-links/src/__tests__/catch-links.js
wardpeet: I moved to addEvenlistener instead, it makes it more verbose when reading the tests. Feel free to tell me to revert if you think otherwise

liamness: I think all we need is something that runs after our handler would have been called, either approach will achieve this so I'm happy. I personally think it's fine for tests to be verbose if the verbosity adds meaning, and a click handler vs setTimeout does add meaning.

PR #13913, opened 2019-05-07
fix(gatsby-remark-responsive-iframe): Support MDX by visiting JSX nodes

johno:
In MDX the nodes that need to be visited are labeled as JSX so this
includes that node type in addition to HTML.

## Related Issues

- https://github.com/mdx-js/mdx/issues/553
- #13552


packages/gatsby-remark-responsive-iframe/src/__tests__/index.js
```diff
@@ -73,6 +83,22 @@ describe(`gatsby-remark-responsive-iframe`, () => {
...
+<iframe url="http://www.example.com/" style="border:0;" width="600px" height="400px"></iframe>
+    `)
+
+    const transformed = await plugin({ markdownAST })
+    const node = find(transformed, function(node) {
+      return node.type === `html`
+    })
+
+    expect(node).toBeDefined()
+    expect(node.value).toMatchSnapshot()
```
pieh: Yikes, it seems like snapshot assertion doesn't work well with `testInNode8OrHigher` (tests fail with `1 snapshot obsolete`)

johno: I'll update the test to not use a snapshot 👍

wardpeet:
Published in `gatsby-remark-responsive-iframe@2.2.1`

PR #13919, opened 2019-05-08
chore(gatsby-plugin-preact): Update to Preact X beta

### LINKED ISSUE #13380: [plugin-preact] Update to use Preact X (preact/compat)
## Description

Preact 8.x is incompatible with reach router and errors out on Gatsby v2 (see #8545). Preact X includes compat for contexts and other React 16 features needed by Gatsby, and has moved the `compat` module to core. This plugin should use that, even though it's currently in alpha, since the current implementation is broken.

Just a matter of updating alias refs from `preact-compat` to `preact/compat` and requiring a peer dep of preact `^10.0.0`.

### Steps to reproduce

Use `gatsby-plugin-preact` with Preact 10 alpha

### Expected result

It should work

### Actual result

It doesn't

seaneking:
## Description

This PR updates `gatsby-plugin-preact` to use the Preact X beta, since Preact 8 is incompatible with Gatsby v2.

## Related Issues

Fixes #13380

### DIFF
--- a/packages/gatsby-plugin-preact/src/gatsby-node.js
+++ b/packages/gatsby-plugin-preact/src/gatsby-node.js
@@ -6,9 +6,8 @@ exports.onCreateWebpackConfig = ({ stage, actions }) => {
     actions.setWebpackConfig({
       resolve: {
         alias: {
-          react: `preact-compat`,
-          "react-dom": `preact-compat`,
-          "create-react-class": `preact-compat/lib/create-react-class`,
+          react: `preact/compat`,
+          "react-dom": `preact/compat`,
         },
       },
     })
--- a/packages/gatsby-plugin-preact/src/gatsby-node.js
+++ b/packages/gatsby-plugin-preact/src/gatsby-node.js
@@ -2,7 +2,7 @@ exports.onCreateWebpackConfig = ({ stage, actions }) => {
   // Requiring the server version of React-dom is hardcoded right now
   // in the development server. So we'll just avoid loading Preact there
   // for now.
-  if (stage !== `develop-html`) {
+  if (stage !== `develop`) {
     actions.setWebpackConfig({
       resolve: {
         alias: {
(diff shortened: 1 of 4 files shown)


seaneking:
On further testing this seems to still be broken in development, assuming because it skips out on the `develop-html` stage. Still an improvement on the current situation (plugin is completely broken), but probably not worth merging until that issue resolved.

seaneking:
Preact was erroring out in development, likely because of the mentioned hardcoded dependency on `react-dom` dev server in code comments. Disabled for entire develop stage for now, as it works in prod builds.

Not ideal, but a vast improvement on the current situation (the plugin doesn't work at all).

packages/gatsby-plugin-preact/src/gatsby-node.js
```diff
@@ -2,13 +2,12 @@ exports.onCreateWebpackConfig = ({ stage, actions }) => {
   // Requiring the server version of React-dom is hardcoded right now
   // in the development server. So we'll just avoid loading Preact there
   // for now.
-  if (stage !== `develop-html`) {
+  if (stage !== `develop`) {
```
wardpeet: This shouldn't be necessary as preact is 100% compatible with react 16. There are some issues with HMR & devtools.

seaneking:
Understand that Preact is still in beta, but the current plugin (which is published in prod with Gatsby V2) is completely broken. So surely merging this is better than the current situation.

If we want to hold off on it, then you should probably deprecate and remove the current plugin from gatsby core.

packages/gatsby-plugin-preact/src/gatsby-node.js
seaneking: Didn't dig too far into it, but Preact definitely breaks in dev with Gatsby. Assumed it was to do with HMR et al

wardpeet:
I feel you! I just think we should hold of a few weeks until hmr is fixed and it works smoothly. I'll fix those issues on the preact repo.

wardpeet:
```
Successfully published:
 - gatsby-plugin-preact@3.0.0
```

PR #13929, opened 2019-05-08
docs(gatsby-source-wordpress): Clarify how to retrieve post meta

### LINKED ISSUE #13883: [docs] gastby-source-wordpress unclear on how to fetch basic post meta
I think I'm missing something here. I see how to fetch ACF data and all when querying posts/pages, but I am not seeing instructions as to how to fetch basic post meta. I searched around and #10240 was the only issue I could find dealing with this but the solution isn't clear?

kmcaloon:
Docs were a little unclear as to how to interact with post meta fields. See #13883.

## Related Issues

Fixes #13883

### DIFF
--- a/packages/gatsby-source-wordpress/README.md
+++ b/packages/gatsby-source-wordpress/README.md
@@ -28,7 +28,7 @@ This module currently pulls the following entities from WordPress:
 - [x] Custom post types (any type you could have declared using WordPress'
       `functions.php`)
 
-Note that all entities and fields must be available via the REST API. This means that if you want to retrieve post meta in your queries, they first must be registered using WordPress' `register_meta()` function with `show_in_rest` set as `true`.  You will then see your registered post meta in your Gatsby GraphQL Schema nested within the `meta` field for associated entities. For more details, see  https://developer.wordpress.org/reference/functions/register_meta/.
+Note that all entities and fields must be available via the REST API. This means that if you want to retrieve post meta in your queries, they first must be registered using WordPress' `register_meta()` function with `show_in_rest` set as `true`. You will then see your registered post meta in your Gatsby GraphQL Schema nested within the `meta` field for associated entities. For more details, see https://developer.wordpress.org/reference/functions/register_meta/.
 
 We welcome PRs adding support for data from other plugins.
 
--- a/packages/gatsby-source-wordpress/README.md
+++ b/packages/gatsby-source-wordpress/README.md
@@ -882,9 +882,9 @@ exports.createPages = async ({ graphql, actions }) => {
 
 ### Missing Fields
 
-If you have custom post types or metadata that are not showing up within the schema, make sure that they are enabled within the REST API. 
+If you have custom post types or metadata that are not showing up within the schema, make sure that they are enabled within the REST API.
 
-- **Custom Meta**  
+- **Custom Meta**
 
   To retrieve custom post meta in your queries, they first must be registered using WordPress' `register_meta()` function with `show_in_rest` set as `true`. You will then see your registered post meta in your Gatsby GraphQL Schema nested within the `meta` field for associated entities. For more details, see https://developer.wordpress.org/reference/functions/register_meta/.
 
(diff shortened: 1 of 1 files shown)


packages/gatsby-source-wordpress/README.md
```diff
@@ -28,6 +28,8 @@ This module currently pulls the following entities from WordPress:
 - [x] Custom post types (any type you could have declared using WordPress'
       `functions.php`)
 
+Note that all entities and fields must be available via the REST API. This means that if you want to retrieve post meta in your queries, they first must be registered using WordPress' `register_meta()` function with `show_in_rest` set as `true`. You will then see your registered post meta in your Gatsby GraphQL Schema nested within the `meta` field for associated entities. For more details, see https://developer.wordpress.org/reference/functions/register_meta/.
```
pieh: This makes a lot of sense. 

One thing I would propose here is to add some title to this, because on first reading it's not clear that this note relates to `meta` field on post/page/custom_post_type entities.

```suggestion
- [x] Post metadata
      Note that all entities and fields must be available via the REST API. This means that if you want to retrieve post metadata in your queries, they first must be registered using WordPress' `register_meta()` function with `show_in_rest` set as `true`. You will then see your registered post meta in your Gatsby GraphQL Schema nested within the `meta` field for associated entities. For more details, see [`register_meta` documentation](https://developer.wordpress.org/reference/functions/register_meta/).
```

kmcaloon: I was wondering the same thing, but wasn't sure where exactly in the docs it belonged.  It also doesn't just apply to custom post meta, as all custom post types must be enabled to show in rest as well. Maybe we should refer people to the [REST API Handbook](https://developer.wordpress.org/rest-api/extending-the-rest-api/adding-rest-api-support-for-custom-content-types)?

kmcaloon: Also, not included within the Handbook, if users want to use access custom meta for custom post types via the rest api they must make sure that the CPT is registered with custom-field support.

```
$args = [
  ...,
  'supports' => [ 'custom-fields']
];
register_post_type( 'custom_post_type', $args );
```

Should we add this in?

pieh: Yeah, it feels like this is missing information that relates to more endpoints / fields. What about something like this:

```md
## WordPress and custom entities

This module currently pulls the following entities from WordPress:

- [x] All entities are supported (posts, pages, tags, categories, media, types,
      users, statuses, taxonomies, site metadata, ...)
- [x] Any new entity should be pulled as long as the IDs are correct.
- [x] [ACF Entities (Advanced Custom Fields)](https://www.advancedcustomfields.com/)
- [x] Custom post types (any type you could have declared using WordPress'
      `functions.php`)
- [x] Post metadata

We welcome PRs adding support for data from other plugins.

Note : If some fields are missing, check [troubleshooting missing fields](#) section.
``` 
And add those notes to troubleshooting section:
 - first verify that fields are in REST responses
 - if they are not, add link to REST API Handbook, and note you did about post meta and `supports` setting on custom post type

I think we should not derail plugin introduction section with heavy troubleshooting notes, so moving it to the bottom + having quick link seems reasonable?

kmcaloon: I really like the idea of having some troubleshooting notes for this.  How about including some of these modifications to the bullets

```
## WordPress and custom entities

This module currently pulls the following entities from WordPress:

- [x] All entities are supported (posts, pages, tags, categories, media, types,
      users, statuses, taxonomies, site metadata, ...)
- [x] Any new entity should be pulled as long as the IDs are correct.
- [x] [ACF Entities (Advanced Custom Fields)](https://www.advancedcustomfields.com/)
- [x] Custom post types (any type you could have registered and enabled in the REST API via  `register_post_type()` )
- [x] Post metadata (any meta fields you could have registered and enabled in the REST API via `register_meta()`)

We welcome PRs adding support for data from other plugins.

Note : If some fields are missing, check [troubleshooting missing fields](#) section.
```

pieh: This sounds good! I just hope mentioning concrete wordpress APIs won't be confusing for people who use plugins to define post types / meta fields (like for example https://wordpress.org/plugins/custom-post-type-ui/ )

pieh: In any case - I still prefer your proposal over what we have now (more concrete information there)

kmcaloon: Sorry I got sidetracked with this. Just committed all suggested changes!

PR #13942, opened 2019-05-09
Change magic numbers to instead use same divisor

baerrach:
## Description

Change the magic numbers in space.js into numbers using the same divisor.

This makes it obviously the relationship between the spaces.


LekoArts:

The change you’ve proposed is not going to be accepted because this change doesn't meaningfully improve docs, fix a bug, or otherwise improve functionality. These tokens are used as an array (so e.g. `space[1]` or `space[3]`) so it doesn't really matter how the spaces are calculated as the user is supposed to use the above way of choosing the right value.

baerrach:
Well there are no docs on this part of gatsby, and the only way to learn about it is to look at the code.

When someone with no knowledge of the system comes in and sees `space[X]` calls they will track down where space is defined and look at the values, but since they do not share a common divisor they dont mean anything.

It wasn't until I did that translation in my head that I could the first 6 values were incrementing in 1/6, and the next 6 incrementing in 2/6. Hence my suggested change, it is immediately obvious what those values are doing.

But I'm not going to argue to hard to get the change accepted.

Issue #12848, opened 2019-03-25
[gatsby-source-drupal] "Failed to process remote content" error on Drupal images


mmafe:
## Summary

Occasionally, during our Gatsby build, right after the line `Starting to fetch data from Drupal`, we will see about 20-25 images which say `Failed to process remote content`. 

In the snippet below, I've change the exact filenames, but kept spaces, dashes, filetypes so you can get a sense of the kind of files that occasionally fail.

Note that some of the files that fail to process are large (4 mb) but others are tiny (80 kb). We've noticed pngs, gifs, and jpgs so far.

Is there a maximum file-size limit that Gatsby will be able to process for remote content? Or is there a file-naming issue with any of the files below (e.g. do we need to avoid dashes, underscores, or spaces)?

Any assistance would be appreciated!

```
Starting to fetch data from Drupal
Failed to process remote content http://site-url/sites/default/files/inline-images/Name_Of_File_S.jpg
Failed to process remote content http://site-url/sites/default/files/2019-03/name-file_0.png
Failed to process remote content http://site-url/sites/default/files/2019-03/Test%20File%20With-Dash_cropped.jpg
Failed to process remote content http://site-url/sites/default/files/nameoffile.gif
Failed to process remote content http://site-url/sites/default/files/Name%20of%20file_2018_1.jpg
Failed to process remote content http://site-url/sites/default/files/NameOfFile.gif
Failed to process remote content http://site-url/sites/default/files/TEST_file2_colour.jpg
Failed to process remote content http://site-url/sites/default/files/Nameoffile.gif
Failed to process remote content http://site-url/sites/default/files/Test_test-file_test-1440x960.jpg
Failed to process remote content http://site-url/sites/default/files/Test%name%20file_2018%20resized%2050.jpg
Failed to process remote content http://site-url/sites/default/files/Test.jpg
Failed to process remote content http://site-url/sites/default/files/Test.png
Failed to process remote content http://site-url/sites/default/files/Test_File_with_Underscores.jpg
Failed to process remote content http://site-url/sites/default/files/testfile_0.png
Failed to process remote content http://site-url/sites/default/files/Test%20File_RETOUCH_S.jpg
Failed to process remote content http://site-url/sites/default/files/testing.gif
Failed to process remote content http://site-url/sites/default/files/only-dashes-and-numbers-292636697.jpg
Failed to process remote content http://site-url/sites/default/files/TestFile.png
[...]

Issue #10339, opened 2018-12-07
gatsby-source-drupal image alt tags


pascal-kubrick:
## Summary

I'm trying to query an image's alt-text / title but it seems like it's not available. I'd love to fork it and fix it myself but can't seem to wrap my head around gatsby-source-drupal/gatsby-node.js (which is where I suppose the changes should be made).

### Basic example

```
{
  nodeHome {
    id

    relationships{
field_slides {
  id
  alt
  localFile {
publicURL
  }
}
    }
  }
}
```

pascal-kubrick:
I've worked around it for now by doing these things;

- replaced all the image fields on the content types with media fields
- disable the alt field on the media types' image field
- manually add an alt text field to the image media type

Not ideal but it'll do for now :)

`{
  nodeHome {
    id
    relationships {
      field_media {
        id
        field_alt
        relationships {
          field_media_image {
            localFile {
              childImageSharp {
                original {
                  src
                }
              }
            }
          }
        }
      }
    }
  }
}
`

bakeruk:
+1 

Same issue!

JSON API has the data for the `alt` tag as meta data, but the source module does not publish this data via GraphQL. It would be great if this could be pushed to the `file__file` relationship or somewhere that easily accessible and semantic.

`...,
"field_cover_image": {
  "data": {
    "type": "file--file",
    "id": "2ab20275-366c-431c-8ec6-d5cc42d5c20e",
    "meta": {
      "alt": "A lovely place where the \"magic\" happens. Toilet close.",
      "title": "",
      "width": 700,
      "height": 580
    }
  },
  "links": {
    "self": {
      "href": "http://site.domain/jsonapi/node/project/243d7fd4-7e55-47c4-a048-ed781b92cc9f/relationships/field_cover_image"
    },
    "related": {
      "href": "http://site.domain/jsonapi/node/project/243d7fd4-7e55-47c4-a048-ed781b92cc9f/field_cover_image"
    }
  }
},
...`

It's not very SEO friendly using this module until this gets fixed.

hehooleehoo:
+ 1 Same issue here...

hehooleehoo:
+1 

Having the same issue here too!

bakeruk:
> Hiya!
> 
> This issue has gone quiet. Spooky quiet. 👻
> 
> We get a lot of issues, so we currently close issues after 30 days of inactivity. It’s been at least 20 days since the last update here.
> 
[...]

PR #13943, opened 2019-05-09
fix(gatsby-source-drupal): limit concurrentFileRequests while downloading remote images

bakeruk:
- Added fix for **gatsby-source-drupal** regarding `createRemoteFileNode` buffer memory exhaustion.
- Added `activity` timers and logs to show steps during a long build process.

## Description

The issue related to this PR is in the "Download all files" logic. The code is written in **parallel** which causes buffer and memory exhaustion issues when `createRemoteFileNode` is called. This is caused by unresolved promises in the call-stack especially when you have substantial pages and media entities.

The use of `for` and `lodash` loops that do not respect resolving promises before continuing on to the next item in the loop is to blame. They currently overwhelm the call-stack with unresolved promises. 

I have simply refactored the code to be **serialised** using the appropriate for-loop functionality and voila!

## Related Issues

#12848 [gatsby-source-drupal] "Failed to process remote content" error on Drupal images

### DIFF
--- a/packages/gatsby-source-drupal/src/gatsby-node.js
+++ b/packages/gatsby-source-drupal/src/gatsby-node.js
@@ -1,3 +1,4 @@
+const report = require(`gatsby-cli/lib/reporter`)
 const axios = require(`axios`)
 const _ = require(`lodash`)
 const { createRemoteFileNode } = require(`gatsby-source-filesystem`)
@@ -9,6 +10,8 @@ exports.sourceNodes = async (
   { baseUrl, apiBase, basicAuth, filters, headers, params }
 ) => {
   const { createNode } = actions
+  const drupalFetchActivity = report.activityTimer(`Fetch data from Drupal`)
+  const downloadingFilesActivity = report.activityTimer(`Remote file download`)
 
   // Default apiBase to `jsonapi`
   apiBase = apiBase || `jsonapi`
@@ -20,7 +23,7 @@ exports.sourceNodes = async (
 
   // Fetch articles.
   // console.time(`fetch Drupal data`)
-  console.log(`Starting to fetch data from Drupal`)
+  console.info(`Starting to fetch data from Drupal`)
 
   // TODO restore this
   // let lastFetched
@@ -31,6 +34,8 @@ exports.sourceNodes = async (
   // lastFetched = store.getState().status.plugins[`gatsby-source-drupal`].status
   // .lastFetched
   // }
+
+  drupalFetchActivity.start()
 
   const data = await axios.get(`${baseUrl}/${apiBase}`, {
     auth: basicAuth,
@@ -93,6 +98,8 @@ exports.sourceNodes = async (
       return result
     })
   )
+
+  drupalFetchActivity.end()
 
   // Make list of all IDs so we can check against that when creating
   // relationships.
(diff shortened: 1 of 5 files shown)


coreyward:
If the issue being experienced is related to concurrency it seems the better solution would be to implement a request pool rather than to abandon concurrency entirely. One potential example: https://github.com/rxaviers/async-pool

bakeruk:
Possibly.

With the PR code, 3000+ images takes 208 seconds.

As no one knows what server/instance this code will be running on or what the memory allocation wilp be. It will be hard to configure the pool sizes to make sure this issue don't arise again.

bakeruk:
For got to say (*getting off a train)...

I'll give concurrency a go with small pool sizes and see if that works 👍🏼

bakeruk:
- Added file request concurrency logic to gatsby-source-drupal to speed build times.
- Added changeable `concurrentFileRequests` option to the plugin and updated `README.md` file

coreyward:
What’s your experience with the async pool? Better timing than the synchronous implementation? Does it still reliably resolve the failing downloads?

bakeruk:
It's really useful! Cheers for bringing it to my attention.

It's very similar to FTP client functionality, which allows 10+ simultaneous transfers at once and continues to the next item once has completed. 

With just the right pool size, it allows just enough promises in the call-stack that it doesn't exhaust the memory but still give you a semi-non blocking runtime.

3000+ images now take ~86 secs to a win if you ask me =)

Will be rolling this async-pool out in other projects of mine.

bakeruk:
@KyleAMathews It would be great if this could get merged in before I take a look at #10339

packages/gatsby-source-drupal/src/gatsby-node.js
```diff
@@ -187,48 +206,62 @@ exports.sourceNodes = async (
...
+          store,
+          cache,
+          createNode,
+          createNodeId,
+          parentNodeId: node.id,
+          auth,
+        })
+      } catch (err) {
+        // Ignore
+        console.error(err)
```
wardpeet: do we want to show an error? We ignored this before. If we want this I recommend using report.error and remove the comment

packages/gatsby-source-drupal/src/gatsby-node.js
```diff
@@ -187,48 +206,62 @@ exports.sourceNodes = async (
...
       }
-    })
-  )
 
-  nodes.forEach(n => createNode(n))
+      // If the fileNode exists set the node ID of the local file
+      if (fileNode && typeof fileNode === `object`) {
```
wardpeet:
```suggestion
      if (fileNode) {
```
do we need to do an object check?

packages/gatsby-source-drupal/src/gatsby-node.js
```diff
@@ -1,26 +1,41 @@
...
 ) => {
   const { createNode } = actions
+  const drupalFetchActivity = report.activityTimer(`Fetch data from Drupal`)
+  const downloadingFilesActivity = report.activityTimer(`Remote file download`)
 
   // Default apiBase to `jsonapi`
   apiBase = apiBase || `jsonapi`
 
+  // Default concurrentFileRequests to `20`
+  concurrentFileRequests = concurrentFileRequests || 20
```
wardpeet: could you explain why 20 is the magic number? 😄

bakeruk: It's dependent on your pipeline's or instance's memory allocation.

20 should be sufficient for all cases. This, of course, can be changed if you know you have sufficient memory to handle the number of promises in the call-stack.

Oddly, in RL testing I found the image size would also affect the success of the runtime if the concurrency pool size was too large.

packages/gatsby-source-drupal/src/gatsby-node.js
bakeruk: It would be nice (if anything was to fail) to have the visibility of what went wrong.

Will change this to `reporter.error`

packages/gatsby-source-drupal/src/gatsby-node.js
```diff
@@ -1,26 +1,41 @@
...
+
   // Touch existing Drupal nodes so Gatsby doesn't garbage collect them.
   // _.values(store.getState().nodes)
   // .filter(n => n.internal.type.slice(0, 8) === `drupal__`)
   // .forEach(n => touchNode({ nodeId: n.id }))
 
   // Fetch articles.
   // console.time(`fetch Drupal data`)
-  console.log(`Starting to fetch data from Drupal`)
+  console.info(`Starting to fetch data from Drupal`)
```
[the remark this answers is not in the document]
bakeruk: Wil do

bakeruk:
Amended PR accordingly from @wardpeet's comments above.

wardpeet:
Updated PR with reporter from node-api instead.

Tested with our drupal example and works fine.

wardpeet:
Published in `gatsby-source-drupal@3.1.5` :tada:

Issue #13910, opened 2019-05-07
Bug: not able to serve built site in macos


unmem:
## Description

Error when running the command `gatsby serve`.

<img width="2506" alt="Screen Shot 2019-05-08 at 12 10 47 AM" src="https://user-images.githubusercontent.com/16096559/57315297-dbfc8980-7125-11e9-91bf-009bedf79a40.png">

### Steps to reproduce
```bash
  cd /tmp
  gatsby new global-styles https://github.com/gatsbyjs/gatsby-starter-hello-world
  cd global-styles
  gatsby build
  gatsby serve
```

### Environment
    System:
      OS: macOS 10.14.4
      CPU: (6) x64 Intel(R) Core(TM) i5-8500 CPU @ 3.00GHz
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 11.13.0 - ~/.nvm/versions/node/v11.13.0/bin/node
      Yarn: 1.15.2 - /usr/local/bin/yarn
      npm: 6.7.0 - ~/.nvm/versions/node/v11.13.0/bin/npm
    Languages:
      Python: 3.7.3 - /Users/peng/.virtualenvs/python-dev/bin/python
    Browsers:
      Chrome: 74.0.3729.131
      Safari: 12.1
    npmPackages:
      gatsby: ^2.4.2 => 2.4.2

cluzier:
You're sure you're in the correct directory? I know in the reproduction steps you do `cd tmp` but is `tmp` really the right place?

unmem:
sorry about missing one step.. you can see the current directory in the screenshot is `global-styles`.

polishedwp:
I can confirm on my MacOS with the Hello-world start, the same result as you have described. However, the Default starter works fine.

thomascorthouts:
I was able to reproduce as well, a quick check learned me that adding the default `gatsby-*.js` files made this one "serve-able" from localhost. 

It seems like an empty `gatsby-config.js` is sufficient to make it work. No immediate idea on the origin of the issue, but just putting it here.

polishedwp:
@thomascorthouts good point. the `hello-world` is a fairly blank starter and it makes since it would not `serve` in is fresh install state. 

I wonder if it would make since to include a `gatsby-config.js` file with the basic `sitemetadata` and no plugins for this starter?

thomascorthouts:
[...]

PR #13948, opened 2019-05-09
fix(gatsby): fallback for non-existing config

thomascorthouts:
This is a proposed way of doing it, I would love to have feedback or a conversation on how to fix/address this in a more Gatsby way.

## Description

I found that with changing from Gatsby `v2.3.36` to `v2.4.2`, a Gatsby website would not be able to be served using `gatsby serve` when there is no `gatsby-config.js` file in the root folder of the project.  This was due to the destructuring on line 72 in the `gatsby/src/commands/serve.js` file, which threw a `TypeError` when the command tried to destructure from `undefined`. 

I tried to solve this error by having a fallback to an empty object. But another (and maybe better) approach could also be that it is clearly documented as a needed file in the structure and to add it to the `gatsby-starter-hello-world` Starter project.

## Related Issues

This issue was addressed in #13910

### DIFF
--- a/packages/gatsby/src/commands/serve.js
+++ b/packages/gatsby/src/commands/serve.js
@@ -69,7 +69,7 @@ module.exports = async program => {
     getConfigFile(program.directory, `gatsby-config`)
   )
 
-  const { pathPrefix: configPathPrefix } = config
+  const { pathPrefix: configPathPrefix } = config || {}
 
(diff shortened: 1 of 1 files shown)


thomascorthouts:
Follow-up question: Would it make sense to have regression tests whether the starters keep on working after updates?

PR #13953, opened 2019-05-09
chore(docs): Update e-commerce tutorial

waleedshkt:
- Removed the legacy betas:['beta_version'] of Stripe Checkout as there is latest production version avalaible shown in the link: https://stripe.com/docs/payments/checkout/migration-from-beta 

- Converted the redirectedToCheckout function to arrow one. It's because otherwise you need to bind the function to the class 'this'. But since in button, the function is being called using 'this' and it looks neat without appending (.bind(this)), so I simple changed the function to arrow one to natively bind to class 'this'

### DIFF
--- a/docs/docs/ecommerce-tutorial/index.md
+++ b/docs/docs/ecommerce-tutorial/index.md
@@ -134,17 +134,15 @@ const buttonStyles = {
   letterSpacing: "1.5px",
 }
 
-const Checkout = class extends React.Component {
+class Checkout extends React.Component {
   // Initialise Stripe.js with your publishable key.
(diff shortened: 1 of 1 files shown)


LekoArts:
> But since in button, the function is being called using 'this' and it looks neat without appending (.bind(this)), so I simple changed the function to arrow one to natively bind to class 'this'

I think you missed the necessary changes to the button, didn't you?

waleedshkt:
@LekoArts, I was telling the alternative option of using 'bind' if not changing the redirectToCheckout function to arrow. But since I proposed the simpler option i.e convert to arrow function therefore no need to change anything in button

PR #13955, opened 2019-05-10
feat(gatsby-plugin-styletron): update to v5, debug mode, useStyletron

tajo:
## Description

- update to the latest v5 styletron-react
- add [debug mode](https://www.styletron.org/api/#browser-debug-mode) when deving
- remove hacky `styletron.css()` export since now you can use a new native hook based API instead:

```js
imporet {css} from 'styletron-react';
const [css] = useStyletron();
```

**This is a breaking change, the plugin should be bumped to v4.*

### DIFF
--- a/packages/gatsby-plugin-styletron/src/gatsby-browser.js
+++ b/packages/gatsby-plugin-styletron/src/gatsby-browser.js
@@ -5,12 +5,16 @@ const { Provider, DebugEngine } = require(`styletron-react`)
 const debug = process.env.NODE_ENV === `production` ? void 0 : new DebugEngine()
 
 // eslint-disable-next-line react/prop-types
-exports.wrapRootElement = ({ element }, options) => (
-  <Provider
-    value={styletron(options).instance}
-    debug={debug}
-    debugAfterHydration
-  >
-    {element}
-  </Provider>
-)
+exports.wrapRootElement = ({ element }, options) => {
+  const enableDebug =
+    options.debug === true || typeof options.debug === `undefined`
+  return (
+    <Provider
+      value={styletron(options).instance}
+      debug={enableDebug ? debug : undefined}
+      debugAfterHydration={enableDebug}
+    >
+      {element}
+    </Provider>
+  )
+}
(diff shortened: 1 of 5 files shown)


packages/gatsby-plugin-styletron/src/gatsby-browser.js
```diff
@@ -1,8 +1,16 @@
 const React = require(`react`)
 const styletron = require(`./index`)
-const { Provider } = require(`styletron-react`)
+const { Provider, DebugEngine } = require(`styletron-react`)
+
+const debug = process.env.NODE_ENV === `production` ? void 0 : new DebugEngine()
```
DSchau: Could/should this also be configurable? e.g. with pluginOptions?

tajo: Yep! You can now disable it:

```js
module.exports = {
  plugins: [
    {
      resolve: `gatsby-plugin-styletron`,
      options: {
        debug: false
      },
    },
  ],
}
```

**Dev ENV:** By default, it's enabled. Can be disabled through `options`.
**Prod ENV:** Always disabled. It adds quite an overhead, so it should never be used in production.

DSchau:
Whoops - one more comment @tajo--do we want to mention that this now requires styletron 5, e.g. like how we mention [gatsby-plugin-emotion requires emotion 10](https://www.gatsbyjs.org/packages/gatsby-plugin-emotion/#gatsby-plugin-emotion)?

tajo:
> Whoops - one more comment @tajo--do we want to mention that this now requires styletron 5, e.g. like how we mention [gatsby-plugin-emotion requires emotion 10](https://www.gatsbyjs.org/packages/gatsby-plugin-emotion/#gatsby-plugin-emotion)?

Added a note. 

Styletron is a depedency of this plugin, so it will get bumped automatically anyway. Emotion is a peerDep.

DSchau:
Published as `gatsby-plugin-styletron@4.0.0`. Thanks for all your help here!

PR #13972, opened 2019-05-11
gatsby-transformer-remark: add ability to query for slugs to markdown headings

### LINKED ISSUE #13925: gatsby-remark-autolink-headers: add slug for markdown headings to GraphQL
Would be great if [`gatsby-remark-autolink-headers`](https://gatsbyjs.org/packages/gatsby-remark-autolink-headers) made the slugs it generates to markdown headings available through GraphQL, i.e. as an additional attribute to the `headings` array generated by [`gatsby-transformer-remark`](https://gatsbyjs.org/packages/gatsby-transformer-remark).

### Basic example

```graphql
{
  allMarkdownRemark {
    edges {
      node {
        headings {
          depth
          value
          slug  // new
        }
      }
    }
  }
}
```

### Motivation

Would make building custom tables of contents [as discussed here](https://github.com/gatsbyjs/gatsby/issues/6896) much easier.

janosh:
## Description

Adds the ability to query for slugs to markdown headings with `gatsby-transformer-remark`. Includes docs and tests.

## Related Issues

Closes #13925.


janosh:
Just noticed that my changes broke some of the existing tests. Will push more commits later...

pieh:
I think something went wrong with brackets here - there is missing closing bracket before `resolve` (L612) and extra bracket that shouldn't be there in L691

janosh:
@pieh Thanks for the tip. I fixed that. However, there remains another issue. I realized after I opened this PR that the `heading.children[0].url` I append to the slug on [line 286](https://github.com/gatsbyjs/gatsby/pull/13972/files#diff-488d99b036db541ee4a25a6991027b5dR286) is only available if `gatsby-remark-autolink-headers` is specified as a plugin for `gatsby-transformer-remark`.

I guess the way to go in this situation is to move the code that makes the `slug` available via GraphQL from `gatsby-transformer-remark` to `gatsby-remark-autolink-headers`? Or is there a canonical way of handling interdependencies between Gatsby packages?

pieh:
> I guess the way to go in this situation is to move the code that makes the `slug` available via GraphQL from `gatsby-transformer-remark` to `gatsby-remark-autolink-headers`? Or is there a canonical way of handling interdependencies between Gatsby packages?

Maybe `gatsby-remark-autolink-headers` could attach new `slug` field to Heading type using schema-customization APIs, will need to check with @stefanprobst. He recently added type extension support, but it might not work when one plugin try to extend other plugin types (afaik for now only user can do that)

stefanprobst:
yes, when using `createTypes` we currently only allow users and the owner plugin to extend existing types. not sure how subplugins fit into this. do they have to follow some naming convention that we can check for?

that said, i think in the case of the remark transformer that does not apply since it still uses `setFieldsOnGraphQLNodeType` API, where we just add whatever fields we get to the type.

there is an escape hatch (adding the field with `createResolvers` to `MarkdownHeading`), which currently works but I'd not want to recommend that as a best practice.

janosh:
> that said, i think in the case of the remark transformer that does not apply since it still uses `setFieldsOnGraphQLNodeType` API, where we just add whatever fields we get to the type.

Is that likely to change in the foreseeable future?

> there is an escape hatch (adding the field with createResolvers to MarkdownHeading), which currently works but I'd not want to recommend that as a best practice.

Could you spell it out for me? What would be the problem with this approach?

janosh:
I'm unsure how to proceed with this PR. If I understood @stefanprobst correctly, I could try to extend the `HeadingType` in `gatsby-transformer-remark` from `gatsby-remark-autolink-headers`? Would I use the [setFieldsOnGraphQLNodeType](https://www.gatsbyjs.org/docs/schema-gql-type/#plugin-fields) API for this?

pieh:
> I'm unsure how to proceed with this PR. If I understood @stefanprobst correctly, I could try to extend the `HeadingType` in `gatsby-transformer-remark` from `gatsby-remark-autolink-headers`? Would I use the [setFieldsOnGraphQLNodeType](https://www.gatsbyjs.org/docs/schema-gql-type/#plugin-fields) API for this?

I think you would need to use `createResolvers`:
so something like in `gatsby-remark-autolink-headers` gatsby-node file

```
exports.createResolvers = ({ createResolvers }) => {
  createResolvers({
    MarkdownHeading: {
      slug: {
        type: `String`,
        resolve(heading) {
          return _FigureOutHowToConstructSlugHere_
        },
      }
    }
  })
}
```
Stefan mentioned he would not recommend this, but really this is only method right now if we want to extend type defined in other plugin

janosh:
@pieh Thanks a lot for that hint! Another question just came up which might be a stupid one. How do I inform `gatsby-transformer-remark` from `gatsby-remark-autolink-headers` that the cache is outdated? I modified the `headingsCacheKey` in `gatsby-transformer-remark/src/extend-node-type.js` to include the `headingsOptions`
```diff
const headingsCacheKey = node =>
  `transformer-remark-markdown-headings-${
    node.internal.contentDigest
-  }-${pluginsCacheStr}-${pathPrefixCacheStr}`
+  }-${pluginsCacheStr}-${JSON.stringify(headingsOptions)}-${pathPrefixCacheStr}`
```
But if I move the options from `gatsby-transformer-remark` to `gatsby-remark-autolink-headers`, I won't be able to access them anymore. I assume `gatsby-transformer-remark` can't access options from subplugins?

On the other hand, moving the cache function as well would mean I'd need access to the entire markdown node in `gatsby-remark-autolink-headers` to read `node.internal.contentDigest`. Seems convoluted.

beaussart:
Any update on this ?

janosh:
@beaussart I'm afraid not.

pieh:
> Another question just came up which might be a stupid one. How do I inform `gatsby-transformer-remark` from `gatsby-remark-autolink-headers` that the cache is outdated?

We would use all passed options to construct cache-key (so regardless of which plugin declared options, `gatsby-transformer-remark` should get all the options)

Stefan did some work on `gatsby-transformer-remark` that will allow us to change type (adding slug from to fields), but we still can't edit args because of use of `setFieldsOnGraphQLNodeType`. I'm not even sure if it's doable if we switch it to schemaCustomization API (without redeclaring all of it in autolink-headers subplugin) 🤔

pieh:
After thinking a bit more about this - if we set `pathToSlugField` and `slugPrefix` args on `slug` field (which will be defined in `autolink-headers`), keep cache-key as it is right now and adjust object returned from `headings` resolver (just resolver, not actual graphql type), so resolver inside `autolink-headers` have everything it needs, I think we can make it work. It will require to update both transformer remark and autolink headers plugins, but IMO that's ok

janosh:
@pieh Sounds good! I’m a little busy atm but feel free to take over this PR if you’ve got it figured out.

pieh:
I'll start today and see if this is actually feasible (not sure if I would be able to finish it tho) - first need to resolve those conflicts :)

janosh:
> first need to resolve those conflicts :)

That's always annoying. Might be easiest to just copy the changes to `gatsby-transformer-remark/README.md` over to a new branch created from the current master and then force push those here.

pieh:
> That's always annoying. Might be easiest to just copy the changes to `gatsby-transformer-remark/README.md` over to a new branch created from the current master and then force push those here.

Yeah, that's what I did locally, I still copied over few changes like using args for cache key (tho maybe this should be in separate pr)

pieh:
I got it working: https://github.com/gatsbyjs/gatsby/compare/master...pieh:janosh-patch-1-autolink-header :
Screenshot 2019-09-27 at 19 27 47

... but as I was looking into it more I started to question wether this should be moved to autolink-headers subplugin actually:

there are other plugins that can set ids on headings - i.e. translations for reactjs.org uses https://github.com/reactjs/reactjs.org/blob/master/plugins/gatsby-remark-header-custom-ids/index.js plugin which allow to declare heading id inside markdown, like:

```
## Agrega React en un minuto {#add-react-in-one-minute}
```
(from https://raw.githubusercontent.com/reactjs/es.reactjs.org/master/content/docs/add-react-to-a-website.md)

Only thing that trully would need to be done here is to read that `id` instead of trying to get url from anchor link that autolink-header generates. The `url` could just be link to page without anchor part if id is not set (granted it wouldn't be very useful, but still correct)

janosh:
> Only thing that truly would need to be done here is to read that id instead of trying to get url from anchor link that autolink-header generates. The url could just be link to page without anchor part if id is not set (granted it wouldn't be very useful, but still correct)

Maybe reserve that for a separate PR?

pieh:
> Maybe reserve that for a separate PR?

I already do this in branch I linked (just need to move it over to that other plugin)

pieh:
What do you think about just adding `id` field (instead of `slug` / `url`) skipping `slug` args all together? Users will be able to query for something like `fields.slug` separately and making assumptions about page path construction is not great idea for general purpose feature.

janosh:
You mean let users generate `fields.slug` themselves in `gatsby-node.js` with any assumptions about page path construction they want to make? Sure thing. Might be a little more work for users but as you say robustness is important for general purpose functionality.

pvdz:
(I've converted this PR into a Draft PR. Although it's been a year, is this PR still expected to be finished?)

janosh:
I think it can be closed unless @pieh would like to push it over the finish line.

pieh:
I probably won't be able to get to it soon, so let's close it.

There will be a paper trail (you can look through closed PRs) if anyone will want to work on this

muescha:
@pieh - would be nice to have a special label for closed PR that can be continued?

pieh:
> @pieh - would be nice to have a special label for closed PR that can be continued?

Maybe? But I would be against - we should track via issues and not PRs (in principle - there is of course nuance to this). Having too many ways/places to track things just make them unmaintainable in my mind and having multiple sources of truth proves to be a problem time and time again (desync, duplication etc)

We do discovery work on overhauling issues and PRs management and I added note about your question

muescha:
Ok

Just asked, because the referenced #13925 was closed

Issue #8122, opened 2018-09-13
Expose an API for manually triggering prefetching for a pathname


schalkventer:
## Summary

Currently there is no way to programmatically trigger the prefetching  done for a pathname when the `<Link>` component is mounted.

### Basic example

```
const fireEvent = () => {
  const id = createNewPerson();
  navigate(`/person?id=${id}`);
}
```

```
<button onClick={fireEvent}>Create Person</button>
```

### Motivation

Developers might want to leverage prefetching when programmatically triggering `navigate`.

kakadiadarpan:
@schalkventer Currently, if anyone hovers over a `Link` component, the resources required for that path are prefetched by default.

@pieh @m-allanson @KyleAMathews any suggestions to leverage prefetching for programmatic navigation?

KyleAMathews:
This seems pretty reasonable to me. 

The API would be something like:

```javascript
import { prefetchPathname } from "gatsby"

prefetchPathname(`/about/`)
```

schalkventer:
@KyleAMathews, if it is as straightforward as the above it seems like a good solution. 👍 

For the sake of clarity, what is the baseline behaviour for preloading via `Link`? Is it as indicated above only once you hover over the link, or is it on the rendering of the view that contains the link?

KyleAMathews:
It's when the link appears in the viewport. We leverage https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API for this.

schalkventer:
Has this been implemented in Gatsby 2? I.e. can it be closed?

devrchancay:
@schalkventer I would think that if it can be closed.

KyleAMathews:
Would love a PR from you for this? Is this something you'd like to work on in the near future? I can provide some hints on how to do it.

schalkventer:
@KyleAMathews For sure!

Do you have any suggestions where I can start looking to find this mechanism in the code?

KyleAMathews:
Great!

The entry module for the browser for `gatsby` is https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/cache-dir/gatsby-browser-entry.js so that's where `prefetchPathname` should be exported from.

The function which does the prefetching is `enqueue` in `loader.js` https://github.com/gatsbyjs/gatsby/blob/0124389f81df542cd0c22fa68fbe980cb04b0120/packages/gatsby/cache-dir/loader.js#L223

[...]

PR #13985, opened 2019-05-12
feat(gatsby): add prefetchPathname to programmatically prefetch (#8122)

ahfarmer:
## Description

Exposes an API for triggering prefetching for a pathname as discussed in #8122


wardpeet:
@ahfarmer this is awesome! Mind adding an e2e-test to make sure this one works the way it should?

pieh:
It would be great if we could also:
 - update TS typings ( https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/index.d.ts ) - it probably would be something like:
   ```ts
   export const prefetchPathname: (path: string) => void
   ```
 - document it - this is tough, as lot of routing docs are stubs now, but we perhaps could add section to https://www.gatsbyjs.org/docs/gatsby-link/#how-to-use-the-navigate-helper-function with programmatic prefetch example

stonehz:
@wardpeet ,
the following PR introduces a warning on the console when you run `gatsby develop` by importing wrong function:
`.cache/gatsby-browser-entry.js`  contains
```js
import { enqueue as prefetchPathname } from "./loader"
```

but `.cache/loader.js` has
```js
export const publicLoader = {
  // Deprecated methods. As far as we're aware, these are only used by
  // core gatsby and the offline plugin, however there's a very small
  // chance they're called by others.
  getResourcesForPathname: rawPath => {
    console.warn(
      `Warning: getResourcesForPathname is deprecated. Use loadPage instead`
    )
    return instance.i.loadPage(rawPath)
  },
  getResourcesForPathnameSync: rawPath => {
    console.warn(
      `Warning: getResourcesForPathnameSync is deprecated. Use loadPageSync instead`
    )
    return instance.i.loadPageSync(rawPath)
  },
  enqueue: rawPath => instance.prefetch(rawPath),

  // Real methods
  getResourceURLsForPathname: rawPath =>
    instance.getResourceURLsForPathname(rawPath),
  loadPage: rawPath => instance.loadPage(rawPath),
  loadPageSync: rawPath => instance.loadPageSync(rawPath),
  prefetch: rawPath => instance.prefetch(rawPath),
  isPageNotFound: rawPath => instance.isPageNotFound(rawPath),
  hovering: rawPath => instance.hovering(rawPath),
}

export default publicLoader
```

PR #13991, opened 2019-05-12
chore(showcase): Add accessible-app.com

marcus-herrmann:
Adding accessible-app.com to the showcase


docs/sites.yml
```diff
@@ -5883,3 +5883,18 @@
   built_by: Pol Milian
   built_by_url: https://github.com/Polcius/
   featured: false
+- title: Accessible App
+  main_url: https://accessible-app.com
+  url: https://accessible-app.com
+  source_url: https://github.com/accessible-app/accessible-app_com
+  description: >
+    Learn how to build inclusive web applications and Single Page Apps in modern JavaScript frameworks. This project collects strategies, links, patterns and plugins for React, Vue and Angular.
+  categories:
```
wardpeet: could you checkout https://www.gatsbyjs.org/showcase/ categories to make sure you use already used categories? Web-app, React, Vue & Angular aren't really the best ones to pick.

marcus-herrmann: Got it! Changed them to established categories.

PR #14024, opened 2019-05-14
chore(gatsby-plugin-typescript): Types install instruction

iwgx:
## Description

When I'm finish did the instruction, I got an error `JSX element class does not support attributes because it does not have a 'props' property.` at react-helmet declaration

That's was because **I need to install @types/react-helmet** first, so I think that's important to mention they need to install type definition also.

screenshot

### DIFF
--- a/packages/gatsby-plugin-typescript/README.md
+++ b/packages/gatsby-plugin-typescript/README.md
@@ -9,7 +9,6 @@ Provides drop-in support for TypeScript and TSX.
 ## How to use
 
 1.  Include the plugin in your `gatsby-config.js` file.
-1.  Install related type definition based on your installed packages.
 1.  Write your components in TSX or TypeScript.
 1.  You're good to go.
 
@@ -21,6 +20,8 @@ module.exports = {
   plugins: [...`gatsby-plugin-typescript`],
 }
 ```
+
+**\*Please note**: If packages don't ship with TypeScript definitions you'll need to manually install those type definitions, e.g. for React. A typical Gatsby project would need: `npm install --save-dev @types/react @types/react-dom @types/node.`\*
 
 ## Options
 
--- a/packages/gatsby-plugin-typescript/README.md
+++ b/packages/gatsby-plugin-typescript/README.md
@@ -9,6 +9,7 @@ Provides drop-in support for TypeScript and TSX.
 ## How to use
 
 1.  Include the plugin in your `gatsby-config.js` file.
+1.  Install related type definition based on your installed packages.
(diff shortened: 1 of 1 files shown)


packages/gatsby-plugin-typescript/README.md
```diff
@@ -9,6 +9,7 @@ Provides drop-in support for TypeScript and TSX.
 ## How to use
 
 1.  Include the plugin in your `gatsby-config.js` file.
+1.  Install related type definition based on your installed packages.
```
DSchau: When we introduce a change like this, we need to weigh whether it's improving comprehension or just introducing potential confusion.

In this case, I'd wager:

1. Someone using this plugin (and more broadly speaking, TypeScript) would generally _already_ know this (?), or
2. If they don't--this doesn't seem like it'd necessarily be helpful to that person?
    - In other words, what action would this cause a reader to take to fix the warning?

So, I'd lean towards just keeping this the way it is before this change. What do you think?

iwgx: Well, I'm new in TypeScript world and maybe you right. But assume people already know this I think a bit harsh(?)

And maybe I'm biased from how create-react-app explicitly said to install type definition (not assume people already know Typescript or not)

https://facebook.github.io/create-react-app/docs/adding-typescript#docsNav

LekoArts: > If they don't know, this can help them to fix the error above.

I'm with Dustin here. The current sentence wouldn't help them and rather cause confusion. It's also not necessary in the context of setting up this plugin (which this README is about).

How about a paragraph below the steps like:

_**Please note:** If packages don't ship with TypeScript definitions you'll need to manually install those type definitions, e.g. for React. A typical Gatsby project would need: `npm install --save-dev @types/react @types/react-dom @types/node`._

---

Because `react-helmet` is too specific IMO. Not everyone is using that.

iwgx: Sure, that's better!

Sorry, it's my bad can't arrange the sentence properly. My main concern is mention that to everyone whose use this plugin. So no problem for the notes above.

And before, I think `react-helmet` is used in all gatsby template so I'm include that there. Thanks for clarification.

LekoArts: Cool :) You can go ahead @iwgx and make the change!

wardpeet: @LekoArts 👍 that sounds awesome. As mentioned above, we aren't as opinionated as create-react-app so it makes sense to be as general as we can.

PR #14057, opened 2019-05-15
docs(gatsby-plugin-sitemap): fix the link to default options

jessepinho:
Always use the SHA! Otherwise, the wrong line gets highlighted ;)

## Description

Used the SHA when linking to the default options, so that the link is truly permanent. And fixed the line numbers to highlight.

### DIFF
--- a/packages/gatsby-plugin-sitemap/README.md
+++ b/packages/gatsby-plugin-sitemap/README.md
@@ -23,7 +23,7 @@ generated sitemap will include all of your site's pages, except the ones you exc
 
 ## Options
 
-The `defaultOptions` [here](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-sitemap/src/internals.js#L34) can be overridden.
+The `defaultOptions` [here](https://github.com/gatsbyjs/gatsby/blob/8291044020254b04eb47b021101e79743794243a/packages/gatsby-plugin-sitemap/src/internals.js#L45-L78) can be overridden.
 
 We _ALWAYS_ exclude the following pages: `/dev-404-page/`,`/404` &`/offline-plugin-app-shell-fallback/`, this cannot be changed.
 


jessepinho:
This was just a documentation change, so the test failures seem unrelated to this PR.

pieh:
Yeah, the test failing are unrelated (we accidentally merged PR that wasn't updated recently and some tests broke)

I think we should just show default options and not link to it. It has same cons as linking to SHA'ed link (potentially not being up to date), but it's easier to read without url hopping.

wardpeet:
Sadly we can't push to this branch.

I wanted to add the following code:
```md
## Options

### `output`: String, default: `/sitemap.xml`

The file path where the output of the sitemap will be saved into. This path is an absolute path starting from your site's root.

### `exclude`: array\<String\>, default: `['/dev-404-page', '/404', '/404.html', '/offline-plugin-app-shell-fallback']`

A list of paths we don't want to include in our sitemap.

We _ALWAYS_ exclude the following pages: `/dev-404-page/`,`/404` & `/offline-plugin-app-shell-fallback/`, this cannot be changed.

### `createLinkInHead`: Boolean, default: `true`

Add `<link rel="sitemap" ="application/xml" href="path to sitemap" />` automatically to your application.

### `serialize`: Function

It gives you control over how every record gets saved to your sitemap. You can change `priority`, `changefreq` or `url`.

### `query`: String

The graphql query we run to get all the necessary data to create your sitemap.This data gets injected into the `serialize` function. By default, we query `allSitePage` and your `siteUrl` from `siteMetadata`.
```

@jessepinho any chance you can update this PR?

jessepinho:
@wardpeet Thanks for adding to this! Weirdly, I can no longer push to/update this branch either. Mind opening a separate PR and tagging this PR in it so we can follow up?

PR #14062, opened 2019-05-15
test(gatsby-source-filesystem): Fix test assertions now that createRemoteFileNode throws

sidharthachatterjee:


antoinerousseau:
how about testing fetch errors like 404? I'm having a hard time catching them, my process crashes with an `UNHANDLED REJECTION` even though I added a `catch` on `createRemoteFileNode`

PR #14071, opened 2019-05-16
[docs] reorganize Tutorials, Rename Docs Sections

marcysutton:
## Description

As part of the [IA Reorganization proposal](https://docs.google.com/document/d/1-tc0on7qUbgYIb_92etuLpW3ETqqswp7g_uOsLsYXaw/edit#), this PR makes some (not all) of the proposed changes to docs a reality. Specifically, I've updated these areas:

- `Tutorial` renamed to `Tutorials`.
- `Docs` > `Guides` are now called `Reference Guides` in the sidebar.
- `Docs` > `API Reference` is now called `Gatsby API` in the sidebar.
- `Docs` > `Advanced Tutorials` are now in `Tutorials`.

URLs were kept the same, except for the Advanced Tutorials pages which moved from `/docs/*` to `/tutorial/*`

Not in this PR: Recipes, Contributing structure changes, and renaming of Behind the Scenes. I'll opt to wait until I can do some more styling work and content research to make sure any changes are actually improvements.


www/src/data/sidebars/doc-links.yaml
```diff
@@ -174,8 +174,8 @@
                   link: /docs/creating-a-transformer-plugin/
                 - title: Submit to Plugin Library
                   link: /contributing/submit-to-plugin-library/
-                - title: Pixabay Source Plugin Tutorial
-                  link: /docs/pixabay-source-plugin-tutorial/
+                - title: Pixabay Image Source Plugin Tutorial
+                  link: /tutorial/pixabay-source-plugin-tutorial/
```
marcysutton: @shannonbux this is worth a discussion around source plugin tutorials. Should they stay here, in the Creating Plugins section of the docs? Advanced Tutorials could then be more consumer-based and less creator-based.

amberleyromo: At one point we discussed splitting up "creator" (creating for Gatsby) oriented tutorials, and "consumer" (using Gatsby things) oriented tutorials, explicitly. @jlengstorf do you remember this?

jlengstorf: Yeah, I think it would be a good idea to move toward two contextual buckets: "Using Gatsby" and "Building for Gatsby".

That way we can have a bucket of "here's how to build plugins/themes/whatever" that is clearly an advanced, "how to customize Gatsby" area of the docs.

The goal is to keep the happy path free of noise, so for someone who just wants to build a site and add some plugins, they shouldn't ever need to see or care about how to _build_ plugins.

www/src/data/sidebars/doc-links.yaml
```diff
@@ -333,7 +333,7 @@
           items:
             - title: Creating Prefixed 404 Pages for Different Languages
               link: /docs/creating-prefixed-404-pages-for-different-languages/
-    - title: API Reference
+    - title: Gatsby API
```
jlengstorf: Should we keep "reference" in here?

marcysutton: I went back and forth on that, and ultimately decided to differentiate from "Reference Guides". But we could change it!

marcysutton:
> One small thought on /docs: Since we're now adding "intermediate", should we also have a "beginner" heading for parts 0-3 (for formatting consistency? "beginner, intermediate, advanced?"

How about `Gatsby fundamentals`?

amberleyromo:
I like that. Just something to keep the flow. It seems weird to anchor the other two sections, and not the first.

marcysutton:
I added a heading to the first set of tutorials and moved the source plugin tutorial back to the docs. We can figure out what to do with "Building with Gatsby" tutorials at a later date.

www/src/data/sidebars/doc-links.yaml
marcysutton: I like that idea! I've moved the source plugin tutorial back to Docs > Plugins > Creating Plugins to limit the scope of this PR, and we can work on Building for Gatsby separately.

www/src/data/sidebars/doc-links.yaml
jlengstorf: I'll defer to you here. It's easy to change later if we see people getting lost.

docs/tutorial/index.md
```diff
@@ -1,22 +1,30 @@
...
 
+## Gatsby fundamentals
+
 0.  [Set Up Your Development Environment](/tutorial/part-zero/): We'll introduce you to core technologies that power Gatsby, and guide you through setting up your development environment.
 1.  [Get to know Gatsby building blocks](/tutorial/part-one/): Starting new projects, developing, and deploying sites.
 1.  [Introduction to using CSS in Gatsby](/tutorial/part-two/): Explore libraries like Typography.js and CSS Modules.
 1.  [Building nested layouts in Gatsby](/tutorial/part-three/): Layouts are sections of your site that are reused across multiple pages like headers and footers.
 
-## Advanced tutorials
+## Intermediate tutorials
```
shannonbux: oh nice, I like these changes. Fwiw, I don't think the 1-3 parts of tutorial are beginner. I agree that they are just fundamental. Part 0 could be considered beginner and moved somewhere else if it is too beginner for the main tutorial. Not sure how many people look at it! I had a huge part in writing it and yet never measured how successful it was...

The only question here is will the /tutorial/ sidebar lead people thru the new "advanced tutorials" as if they are part of the main tutorial? That might work ok, although the flow could be a little strange since they change in tone. Also the bottom of each doc in the main tutorial has a link to the next part of the tutorial, so that consistency would be lost and need work to update

marcysutton: Hey @shannonbux, great points! Here are some screenshots to show how it's set up right now. The Advanced Tutorials sidebar item is still linked from Docs, and on the Tutorials page it's separate from steps 0-8. The Advanced Tutorials are listed on their own page and linked from the Tutorial introduction. I'll take a look at the copy at the end of the main part of the tutorial, and we can make additional improvements after this too (like adding more of a flow–I think this is sort-of a first pass on advanced tutorials, especially with the major changes planned for "Choose your own adventure" tutorials).

Tutorials sidebar

Tutorials page

jlengstorf: This looks great! Do we need an "INTERMEDIATE TUTORIALS" sidebar grouping on the tutorials page to complement the "ADVANCED TUTORIALS" section?

marcysutton: Beginner and intermediate are more of a group, as they go in order from 0-8. Advanced Tutorials has a separate heading in the sidebar because those don't yet follow the same flow...and people wouldn't necessarily do them in a specific order. They are more a la carte!

Given we have a future plan to redesign the tutorials to a "Choose Your Own Adventure" based on skill and experience level as starting points, I would hesitate to put too much effort into changing the existing structure. The biggest thing holding it back is the fact that sidebar grouping would collapse the tutorial items. Is there a way to add styled headers to the sidebar without adding "Gatsby fundamentals" and "Intermediate tutorials" pages? That would be overkill IMO.

marcysutton:
I've updated this to work with the MDX branch. I think we should merge it, and continue to make improvements!

PR #14096, opened 2019-05-17
fix(www): workaround schema quirks

pieh:
this fixes .org builds

it should be temporary and be reverted when this is fixed (if the behaviour change is bug)


pieh:
We try to remember about it ;)

But seriously, I'm creating minimal reproduction and creating issue and will create to-do list there and reverting this will be one of the items

Issue #14085, opened 2019-05-16
with `gatsby new`, users have options to select one of three default starters or visit the starter library


shannonbux:
## Summary

Create-react-app has a helpful flow when starting new projects. Ideas: open starter library or give options of starters in the CLI.

### Basic example

```shell
$ gatsby new
What is your project called? (my-gatsby-project)
```
```shell
Please select a starter or visit https://www.gatsbyjs.org/starters?v=2 to browse more starters: (use arrow keys)
* gatsby-starter-default
* gatsby-starter-blog
* gatsby-starter-hello-world
```

### Motivation

This makes the choice between starters more discoverable and sets a good precedence when we expand a list of ways to start a Gatsby project in the future.

PR #14097, opened 2019-05-17
feat(gatsby-cli): Prompt user for options (no args with gatsby new)

gillkyle:
## Description

This PR would allow the `gatsby new` command with empty args to prompt the user for the path and official starter they'd like to use like this:

gtsby-new

I added a 4th option where they could select use another option which just exits the command and logs the url to the starter showcase as well:

gtsby-new-2

Some things worth discussing:
- should the starter used be the first question? It would allow a short circuit exit if they select option 4 (Use a different starter)
- is their wording that should change?

## Related Issues

Addresses issue #14085


pieh:
Nice!

I would add to the discussion:
 - maybe open browser with starter library url when selecting 4th option (f.e. using `opn` package)
 - maybe show starter selection prompt anytime that starter wasn't specified (so change current behaviour of auto selecting default starter)

As for your question - I would keep default starter as default choice (especially if we would move forward with my 2nd discussion item)

This is minor and not really relevant as this is draft PR, but we already use `prompts` for ... prompts :P so would be good to not use multiple packages for similar thing (wanted to mentioned before you add more prompts :D)

gillkyle:
I think those are both great ideas, just switched `inquirer` for `prompts` 😅 , guess I should've looked through the existing cli commands a little more first.

gillkyle:
This now opens the starter library when an official starter isn't selected. 

I do like @pieh's idea of showing a prompt to select a starter when no starter is supplied, though I think having the default starter being used with just a path defined is useful in places like the [Quick Start](https://www.gatsbyjs.org/docs/quick-start) where `gatsby new gatsby-site` is one less step to be explained. Maybe I'm being nitpicky though. 

With the changes in this PR the behavior of entering commands with other combinations of args aren't changed, but resolves the weird error you get when you run `gatsby new` alone in a folder that isn't empty.

pieh:
> though I think having the default starter being used with just a path defined is useful in places like the [Quick Start](https://www.gatsbyjs.org/docs/quick-start) where `gatsby new gatsby-site` is one less step to be explained. Maybe I'm being nitpicky though.

This is very valid point. Selecting one of starters assume user already have some familiarity with Gatsby. Tho even that could be somewhat resolved with adding more context to selection (something like):
- gatsby-starter-default - good to see what Gatsby is about
- gatsby-starter-blog - directory structure and configured plugins to quickly jumpstart your markdown blog
- gatsby-starter-hello-world - minimal, barebones project, no plugins or packages other than gatsby and react

But this isn't hill I want to die on ;)

gillkyle:
The order of selections for starters right now aligns with their popularity of usage. The default starter is the most popular, but maybe because it is just the default when gatsby new is run without a starter given?

wardpeet:
RIP @pieh XD

This is awesome!

packages/gatsby-cli/src/init-starter.js
```diff
@@ -177,9 +224,21 @@ type InitOptions = {
...
-  const rootPath = options.rootPath || process.cwd()
+  const { starterPath, rootPath, selectedOtherStarter } = await getPaths(
+    starter,
+    options.rootPath
+  )
 
   const urlObject = url.parse(rootPath)
+
+  if (selectedOtherStarter) {
+    report.log(
```
wardpeet: report.info might be a better fit
```suggestion
    report.info(
```

Issue #13427, opened 2019-04-17
Problems building for IE11


philippsiegmund:
## Summary

Hey there! I have problems building our page for IE11.
But there seem to more underlying issues as well and I think it comes from webpack or ES6 in modules or similar.
While starting with 'gatsby develop' I get 'WeakSet' not found, so I started manipulating with .babelrc and core-js, I tried core-js2 and 3, in devDependecies, in dependencies, I tried to change the webpack config, and currently I get the page builded on IE11 without google fonts, without Lottie animations, because gatsby mistakenly taken those as Graph Data.

I also found an open PR, which seems to modify the webpack config for ES6 modules, but currently I am happy for any input!

Best regards,

Philipp!

### Environment (if relevant)

  System:
    OS: macOS 10.14.4
    CPU: (8) x64 Intel(R) Core(TM) i5-8259U CPU @ 2.30GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 10.14.1 - /usr/local/bin/node
    Yarn: 1.15.2 - /usr/local/bin/yarn
    npm: 6.7.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 73.0.3683.103
    Firefox: 66.0.2
    Safari: 12.1
  npmPackages:
    gatsby: ^2.3.23 => 2.3.23
    gatsby-cli: ^2.4.11 => 2.5.8
    gatsby-image: ^2.0.29 => 2.0.39
    gatsby-plugin-copy-files: ^1.0.1 => 1.0.1
    gatsby-plugin-manifest: ^2.0.16 => 2.0.29
    gatsby-plugin-offline: ^2.0.25 => 2.0.25
    gatsby-plugin-react-helmet: ^3.0.5 => 3.0.12
    gatsby-plugin-sharp: ^2.0.19 => 2.0.34
    gatsby-plugin-styled-components: ^3.0.5 => 3.0.7
    gatsby-plugin-web-font-loader: ^1.0.4 => 1.0.4
    gatsby-source-filesystem: ^2.0.19 => 2.0.29
    gatsby-transformer-json: ^2.1.7 => 2.1.11
    gatsby-transformer-remark: ^2.2.3 => 2.3.8
    gatsby-transformer-sharp: ^2.1.13 => 2.1.18

### File contents (if changed)

`gatsby-config.js`:
```
module.exports = {
  siteMetadata: {
    title: `title`,
    description: `description`,
    author: `@magicsoup.io`,
[...]

Issue #12409, opened 2019-03-08
Feature Request: babel-preset-gatsby additional settings


Undistraction:
## Summary

It would be very useful to be able to override:

1. Which browsers are polyfilled when running in development (for example to allow IE11 to be tested when running `gatsby develop`.
2. Which strategy to use in Polyfilling (`usage` vs `entry`). When using third part modules that require polyfilling, Gatsby will not apply the polyfills they need. Switching to `entry` would allow third party modules to be covered.

### Basic example

It currently exposes only a `targets` setting. It could add additional `developmentTargets` and `babelEnvStrategy` settings. 

### Motivation

Currently to change these settings you need to copy the whole of Gatsby's babel config and edit. IMHO it would make more sense to expose these settings as an API for the plugin.

DSchau:
> Which browsers are polyfilled when running in development

I'm not sure this needs to live in babel-preset-env. One really important note here is that we _generally_ want to keep development and build as unified as possible. If we include something in development that is _not_ included in production, I could see that leading to bugs just as frequently as not including polyfills at all.

That being said--development should work in IE 11, so if it doesn't--let's fix that!

> Which strategy to use in Polyfilling (usage vs entry). When using third part modules that require polyfilling, Gatsby will not apply the polyfills they need. Switching to entry would allow third party modules to be covered.

This probably makes sense as a quick remediation and exposing an option (defaulted to current behavior of usage). Something like (in your app) with `.babelrc.js`

```js
module.exports = {
  presets: [
    ['babel-preset-gatsby', {
      useBuiltIns: 'entry'
    }]
  ]
}
```

Would need to validate that these changes don't cause issues in any current flow while also providing a feasible alternative for users with requirements of supporting browsers like IE 11.

Undistraction:
Thanks. As I mention in my reply to https://github.com/gatsbyjs/gatsby/issues/12399 I think Gatsby already treats development differently from prodution.

> targets
> { [string]: number | string }, defaults to { "browsers": ["last 4 versions", "safari >= 7", "ie >= 9"] } in production and { "browsers": ["last 2 versions", "not ie <= 11", "not android 4.4.3"] } in development when targeting the browser and { "node": 6 } in production and { "node": "current" } in development when targeting Node.js.

So unless I'm misunderstanding the docs, this problem can be fixed by using the same browser range for both cases (at least in userland).

DSchau:
Well - let's not make changes quite yet!

[...]

Issue #12399, opened 2019-03-08
Gatsby's Use Of Polyfills and the 'Missing Resources for x' error


Undistraction:
## Description

I've just been through the incredibly painful process of getting a Gatsby site working in IE 11 and thought it was worth writing up the problems I faced due to Gatsby's current architecture. 

We were receiving an error for every page in IE: 

> Can't find resource for ...

There was no other error. The pages loaded fine in all the other browsers we tested, so my hunch was that Gatsby was swallowing some kind of IE-specific error. 

After a lot of digging I eventually tracked the point of failure to Gatsby's [loader](https://github.com/gatsbyjs/gatsby/blob/88866c7d639d62e65c21dad0445e5c861ad9b9d2/packages/gatsby/cache-dir/loader.js), specifically, [this](https://github.com/gatsbyjs/gatsby/blob/88866c7d639d62e65c21dad0445e5c861ad9b9d2/packages/gatsby/cache-dir/loader.js#L112) line.

```
.catch(() => {
   failed = true
})
```
Inspecting the value passed to the `catch` I could see that the error was related to something that needed to be polyfilled. The problem here being that Gatsby was swallowing this useful error and giving me nothing to help me pinpoint the issue. The resource wasn't missing - it was just erroring when the Loader tried to fetch it. It would make a lot of sense to error here with a more useful message, passing on the specific error that the loader had encountered.

This leads me on to the second issue. Gatsby uses `babel-preset-env` to handle polyfills, and by default `useBuiltIns` is set to `usage`. This sounds great on the surface, but masks what I would consider to be a big problem. If any third-party modules need polyfilling, it seems that Gatsby will not add the Polyfills they need. It will add polyfill imports to project-based code, but will not add imports to non-project-based code. This means if a third-part module was published with `useBuiltIns` set to `entry`, it appears that Gatsby will not include the expected `babel-polyfill` import and replace with the necessary polyfill imports. This means that if a third-party library needs a polyfill that hasn't been supplied due usage within the gatsby project, then this polyfill will not be included. In my case it was a dependency of another dependency. 

Combined with the first issue, this was nightmarishly hard to debug. 

[...]

PR #14111, opened 2019-05-17
feat(gatsby): enable babel-loader for all dependencies

wardpeet:
## Description
I ran gatsby through babel-loader to make sure babel-preset-env can adds it's polyfills for all dependencies so we don't break on older browsers. This will also help us for modern builds so our modern build will have fewer polyfills than our classic one.

To give you an idea on how much slower it is. The first run on a big website takes about double so that's something to wonder about.

| **default-starter** | no babel cache | babel cache |
|-----------------------------------------|----------------|-------------|
| transpile node_modules (no babel-cache) | 5,5623s | 2,887s |
| exclude node_modules (no babel-cache) | 3,148s | 2,224s |
| diff | 76% | 29% |

| **gatsbyjs.org** | no babel cache | babel cache |
|-----------------------------------------|----------------|-------------|
| transpile node_modules (no babel-cache) | 31,487s | 14,684s |
| exclude node_modules (no babel-cache) | 15,896s | 10,804s |
| diff | 98% | 36% |

| **default-starter** | runtime 1 | runtime 2 | runtime 3  |
|-----------------------------------------|----------------|-------------|-------------|
| transpile node_modules (no babel-cache) | 5,198s | 5,768s | 5,721s |
| transpile node_modules (babel-cache)    | 2,771s | 2,958s | 2,932s |
| exclude node_modules (no babel-cache) | 2,937s | 3,339s | 3,168s |
| exclude node_modules (babel-cache)    | 2,172s | 2,283s | 2,217s |

| **gatsbyjs.org** | runtime 1 | runtime 2 | runtime 3  |
|-----------------------------------------|----------------|-------------|-------------|
| transpile node_modules (no babel-cache) | 30.013s | 32.846s | 31.602s |
| transpile node_modules (babel-cache)    | 14.761s | 15.519s | 13.771s |
| exclude node_modules (no babel-cache) | 15.686s | 16.385s | 15.617s |
| exclude node_modules (babel-cache)    | 12.336s | 10.532s | 9.545s |

# bundle info
http://wardpeet-filestorage.surge.sh/gatsby-store/pr-14111

## Related Issues
https://github.com/gatsbyjs/gatsby/issues/12409
https://github.com/gatsbyjs/gatsby/issues/12399
https://github.com/gatsbyjs/gatsby/issues/13427
https://github.com/gatsbyjs/gatsby/issues/5553
https://github.com/gatsbyjs/gatsby/issues/7064
https://github.com/gatsbyjs/gatsby/pull/13241

### DIFF
--- a/packages/gatsby/src/utils/webpack-utils.js
+++ b/packages/gatsby/src/utils/webpack-utils.js
@@ -295,32 +295,14 @@ module.exports = async ({
   {
     let js = (options = {}) => {
       return {
-        test: /\.jsx?$/,
+        test: /\.(js|mjs|jsx)$/,
         exclude: /core-js|event-source-polyfill|webpack-hot-middleware\/client/,
+        type: `javascript/auto`,
         use: [loaders.js(options)],
       }
     }
 
     rules.js = js
-  }
-
-  /**
-   * mjs loader:
-   * webpack 4 has issues automatically dealing with
-   * the .mjs extension, thus we need to explicitly
-   * add this rule to use the default webpack js loader
-   */
-  {
-    let mjs = (options = {}) => {
-      return {
-        test: /\.mjs$/,
-        include: /node_modules/,
-        type: `javascript/auto`,
-        ...options,
-      }
-    }
-
-    rules.mjs = mjs
   }
 
   {
--- a/packages/gatsby/src/utils/webpack-utils.js
+++ b/packages/gatsby/src/utils/webpack-utils.js
@@ -296,7 +296,7 @@ module.exports = async ({
     let js = (options = {}) => {
       return {
         test: /\.jsx?$/,
-        exclude: /core-js/,
+        exclude: /core-js|event-source-polyfill|webpack-hot-middleware\/client/,
         use: [loaders.js(options)],
       }
     }
--- a/packages/gatsby/src/utils/webpack-utils.js
+++ b/packages/gatsby/src/utils/webpack-utils.js
@@ -293,10 +293,14 @@ module.exports = async ({
    * JavaScript loader via babel, excludes node_modules
    */
   {
-    let js = (options = {}) => {
+    let js = ({ exclude = [], ...options } = {}) => {
+      const excludeRegex = [
+        `core-js|event-source-polyfill|webpack-hot-middleware/client`,
+      ].concat(exclude)
+
       return {
         test: /\.(js|mjs|jsx)$/,
-        exclude: /core-js|event-source-polyfill|webpack-hot-middleware\/client/,
+        exclude: new RegExp(excludeRegex.join(`|`)),
         type: `javascript/auto`,
         use: [loaders.js(options)],
       }
--- a/packages/gatsby/src/utils/webpack-utils.js
+++ b/packages/gatsby/src/utils/webpack-utils.js
@@ -296,7 +296,7 @@ module.exports = async ({
     let js = (options = {}) => {
       return {
         test: /\.jsx?$/,
-        exclude: vendorRegex,
+        exclude: /core-js/,
         use: [loaders.js(options)],
       }
     }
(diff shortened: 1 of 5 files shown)


packages/gatsby/src/utils/webpack-utils.js
```diff
@@ -292,7 +292,7 @@ module.exports = async ({
     let js = (options = {}) => {
       return {
         test: /\.jsx?$/,
-        exclude: vendorRegex,
+        exclude: /core-js/,
```
wardpeet: Somehow core-js needs to be excluded because babel will try to polyfill core-js with core-js XD also mini-extra-css-plugin doesn't work well without this. This magically fixes it, i'm not 100% sure so the above comments are just speculation 😄 .

I'll add a comment when it's getting out of draft mode.

sidharthachatterjee:
This will finally allow dependencies (themes and others) to have queries! Woohoo 🚀

packages/gatsby/src/utils/webpack-utils.js
sidharthachatterjee: > will try to polyfill core-js with core-js 😂

sidharthachatterjee:
Want to also post diff of bundle size from before and after? Curious how much this adds or removes

sidharthachatterjee:
Can also remove note in https://www.gatsbyjs.org/docs/browser-support/#polyfills

wardpeet:
I've added webpack stats:
exclude node_modules:
http://wardpeet-filestorage.surge.sh/gatsby-store/pr-14111/current-stats.html
http://wardpeet-filestorage.surge.sh/gatsby-store/pr-14111/current-webpack.stats.json

transpile node_modules:
http://wardpeet-filestorage.surge.sh/gatsby-store/pr-14111/node-modules-stats.html
http://wardpeet-filestorage.surge.sh/gatsby-store/pr-14111/node-modules-webpack.stats.json


**NOTE: these are not uglified so i'll make an uglified version too"

wardpeet:
I've edited above comment with new elements! Lets get this in! ^^

KyleAMathews:
💯yeah, let's do this! The extra build time is a bummer but so is sites not working on older browsers.

Does this PR disable transpiling node_modules during development? That'd be something to consider as people are more sensitive to build times when trying to start the develop server vs. doing a production build (which probably is happening on a CI server somewhere)

wardpeet:
I've found a bug on develop where somehow react-hot-loader is causing issues. trying to find out why

wardpeet:
seems like I needed to add event-source-polyfill & webpack-hot-middleware/client as we don't want them to get polyfilled. As they act as HMR polyfills. If we don't exclude we get some core-js loops and errors.

wardpeet:
I haven't tested the latest change, I'll do that tomorrow

wardpeet:
Ok I made a mistake :p I was running all node_modules during develop instead of builds.

For .org it took me 40s less to bootup without babel running on develop so 👍

wardpeet:
Published in `gatsby@2.11.0`

monsieurnebo:
That's great, thanks for this @wardpeet !

jquense:
this is kinda of a breaking change for sites with custom babel configs ya'll

DSchau:
@jquense yikes —- we’re sorry to hear that!

Would you mind opening up an issue and showing us exactly how this is breaking?

jquense:
It wasn't a huge deal for us, but it's not obvious what was happening. There isn't a lot to how it is unfortunately. It's just not safe to apply any babel config to any code.

For instance: `babel-dev-expression` breaks trying to compile `warning`

joealden:
@wardpeet the increased build times are a pretty big issue for us, as our build times were long enough as they were. They have gone from about 6 minutes all the way up to 11 minutes (and rebuilds are triggered quite regularly). As we are paying for compute time per minute spent building, this drives up the costs for CI builds to nearly double.

While I understand why the changes in this PR were introduced, would it not be possible to have an option to turn this off (in `gatsby-config.js` or as a env var)?

wardpeet:
@joealden thanks, i've created an issue about this https://github.com/gatsbyjs/gatsby/issues/15213

PR #14114, opened 2019-05-17
fix(www): Normalize data from airtable on schema level, remove unnessecary hook.

freiksenet:
We are blocked because currently sort doesn't resolve the data. Waiting for @stefanprobst to push his PR.


pieh:
This looks so much better. We can gradually get this in by doing extra sort on data from graphql in page component if this is blocked by sorting on fields with resolvers - so after Stefan's PR is completed we can just remove one sort call

m-allanson:
PR that resolve sort fields: #14423

www/src/hooks/use-community-events.js
```diff
@@ -1,40 +0,0 @@
...
-        }
-      }
-    }
-  `)
-
-  const events = nodes.map(event => {
-    return {
-      id: event.id,
-      ...event.data,
-      date: new Date(`${event.data.date}T23:59:59.999Z`),
```
m-allanson: This was part of an off-by-one fix that was added in https://github.com/gatsbyjs/gatsby/pull/13845. I don't it's  handled by this PR?

wardpeet:
just a question why do you export the fragments as a query constant? https://github.com/gatsbyjs/gatsby/blob/7d8d6e072e55ac1898632718e41fad49cdf3c964/www/src/components/events/events.js#L38

Just wondering as it might conflict with https://www.gatsbyjs.org/docs/page-query/#add-the-graphql-query & static-queries.

pieh:
> just a question why do you export the fragments as a query constant?

This is what https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/query/file-parser.js looks for when looking for GraphQL documents (so same thing as queries). We don't really care about name of export - we only care that something is named export, is `graphql` tagged template literal - later on we process extracted documents and figure out what are queries there and let `relay-compiler` to stitch used fragments

PR #13766, opened 2019-05-01
[docs] Restructure styling section

robinmetral:
## Description

See #11490 for context and discussion. Build a local copy of the branch to better see the changes to the docs structure :slightly_smiling_face: 

## Related Issues

Closes #11490 
#5997

cc @marcysutton

PR #14115, opened 2019-05-17
docs: update example css background color to be more accessible

noahblon:
Updated from lavenderblush to peachpuff which has more contrast against white and should be more accessible.

## Description
As a user with color blindness, I had a hard time picking up lavenderblush.  Peachpuff has more contrast and is still fun!

## Related Issues

None.


marcysutton:
This change was going to get swallowed by upcoming CSS docs updates, so I merged it into https://github.com/gatsbyjs/gatsby/pull/13766/ in https://github.com/gatsbyjs/gatsby/pull/13766/commits/fb29153d642a76612335aa88e5d7742242bb967b

PR #14116, opened 2019-05-17
[schema] Merge user-defined types with plugin-defined types

stefanprobst:
continued from  #13663

When defining GraphQL types with `createTypes`, allow users to extend types which have already been defined by plugins, while disallowing plugins to modify types defined by other plugins.

This works for merging:
*  SDL into SDL
* typeBuilder into SDL
* SDL into typeBuilder
* typeBuilder into typeBuilder
* SDL into graphql-js

BUT NOT FOR:
* graphql-js into SDL (see the skipped test)
This last one does NOT work in cases where the graphql-js type has a field that references an ObjectType that is already defined in the schema composer - those will be duplicated. I'm a bit hesitant if we should add complexity for fixing this scenario? It's sort of the reason why TypeBuilders are preferable in the first place.

### DIFF
--- a/packages/gatsby/src/schema/schema.js
+++ b/packages/gatsby/src/schema/schema.js
@@ -5,6 +5,7 @@ const {
   isIntrospectionType,
   defaultFieldResolver,
   assertValidName,
+  parse,
 } = require(`graphql`)
 const {
   ObjectTypeComposer,
@@ -138,18 +139,25 @@ const processTypeComposer = async ({
 const addTypes = ({ schemaComposer, types, parentSpan }) => {
   types.forEach(({ typeOrTypeDef, plugin }) => {
     if (typeof typeOrTypeDef === `string`) {
-      let addedTypes
+      let parsedTypes
+      const createdFrom = `sdl`
       try {
-        addedTypes = schemaComposer.addTypeDefs(typeOrTypeDef)
+        parsedTypes = parseTypeDefs({
+          typeDefs: typeOrTypeDef,
+          plugin,
+          createdFrom,
+          schemaComposer,
+          parentSpan,
+        })
       } catch (error) {
         reportParsingError(error)
       }
-      addedTypes.forEach(type => {
+      parsedTypes.forEach(type => {
         processAddedType({
           schemaComposer,
           type,
           parentSpan,
-          createdFrom: `sdl`,
+          createdFrom,
           plugin,
         })
       })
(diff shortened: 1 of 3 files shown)


packages/gatsby/src/schema/schema.js
```diff
@@ -591,6 +649,65 @@ const addTypeToRootQuery = ({ schemaComposer, typeComposer }) => {
...
+
+    if (schemaComposer.has(name)) {
+      // We don't check if ast.kind matches composer type, but rely
+      // that this will throw when something is wrong and get
+      // reported by `reportParsingError`.
+
+      // Keep the original type composer around
+      const typeComposer = schemaComposer.get(name)
+      // Only allow user to extend an already existing type
+      if (!plugin || plugin.name === `default-site-plugin`) {
```
freiksenet: Am I missing something or is this only checked for SDL types?  What about plugin defined type builder or graphql-js types?

packages/gatsby/src/schema/schema.js
```diff
@@ -161,26 +169,71 @@ const addTypes = ({ schemaComposer, types, parentSpan }) => {
...
 
+const mergeTypes = ({
+  typeComposer,
+  type,
+  plugin,
+  createdFrom,
+  parentSpan,
+}) => {
+  typeComposer.merge(type)
+  if (
```
freiksenet: Why is it only merged for those types?

packages/gatsby/src/schema/schema.js
stefanprobst: You're right, this was wrong.

packages/gatsby/src/schema/schema.js
stefanprobst: Honestly can't remember, changed to NamedTypeComposer

packages/gatsby/src/schema/schema.js
```diff
@@ -161,26 +171,79 @@ const addTypes = ({ schemaComposer, types, parentSpan }) => {
       })
 
       if (type) {
+        const typeName = type.getTypeName()
+        const createdFrom = `typeBuilder`
+        checkIsAllowedTypeName(typeName)
+        if (schemaComposer.has(typeName)) {
```
pieh: Does this rely on order of added types?

I.e. if `default-site-plugin` adds type in `onPreBootstrap`, and some plugin adds types in `sourceNodes` - will we try to merge plugin type into type from site (which would fail)

stefanprobst: good point, yes, types added with the `createTypes` action are currently processed in the order they were added to the store.

the (documented, but not enforced) convention however is to only call `createTypes` from `sourceNodes` where `default-site-plugin` is guaranteed to run last (i think, actually have to check)

in the future, [the plan (which probably needs a bit more discussion)](https://github.com/gatsbyjs/gatsby/pull/13738#issuecomment-489318413) is to introduce one new Node API as the only place where schema customization stuff (createTypes, addThirdPartySchema, createFieldExtension) is allowed to be called. this would also solve the fact that the `createTypes` action currently can also be called in APIs running after schema generation, which means those will not be processed at all.

for now, we can either check the plugin name [right in the reducer](https://github.com/gatsbyjs/gatsby/blob/e47ed896723aa23979c9c7b1ecbede9312a8913d/packages/gatsby/src/redux/reducers/schema-customization.js#L21) to sort `default-site-plugin` to the end, or even simpler just sort [here](https://github.com/gatsbyjs/gatsby/blob/e47ed896723aa23979c9c7b1ecbede9312a8913d/packages/gatsby/src/schema/index.js#L15).

freiksenet:
I'll test using @pieh's magical tool and merge.

PR #14121, opened 2019-05-17
fix(www): Add og:url meta property to fix LinkedIn shares

### LINKED ISSUE #14093: [bug] [www] gatsbyjs.org links do not have correct destination on LinkedIn social share
This is a bug that impacts all gatsbyjs.org links that are shared on the LinkedIn social network.

## Description

When gatsbyjs.org links are posted on LinkedIn for sharing, only "How Gatsby works" — the title of the home page — appears, and when a miniature preview is clicked on a published LinkedIn post, it leads to the gatsbyjs.org home page.

### Steps to reproduce

1. Share any gatsbyjs.org URL that is not the home page to a post on LinkedIn.
2. Note that the miniature preview points to "How Gatsby works" — the home page — and not the desired link.
3. Note that after the post is published, the link still points to the home page.

### Expected result

The link should work correctly on LinkedIn social share and not lead to the gatsbyjs.org homepage.

### Actual result

Screen Shot 2019-05-16 at 4 58 29 PM

The link instead points to the gatsbyjs.org home page.

### Environment

Not relevant.

Jared-Dev:
## Description

Adds the `og:url` meta tag to the `blog` template to try and resolve the issue. I don't have a way to share a local page to test unfortunately though... However, I can't see any negatives to having this additional data on the page.

## Related Issues

I believe this fixes #14093

### DIFF
--- a/www/src/templates/template-blog-post.js
+++ b/www/src/templates/template-blog-post.js
@@ -37,9 +37,11 @@ const renderAst = new rehypeReact({
 
 class BlogPostTemplate extends React.Component {
   render() {
-    const post = this.props.data.markdownRemark
-    const prev = this.props.pageContext.prev
-    const next = this.props.pageContext.next
+    const {
+      pageContext: { prev, next },
+      data: { markdownRemark: post },
+      location: { href },
+    } = this.props
     const prevNextLinkStyles = {
       "&&": {
         borderBottom: 0,
@@ -73,7 +75,7 @@ class BlogPostTemplate extends React.Component {
         <link rel="canonical" href={post.frontmatter.canonicalLink} />
       )
     }
-
+    console.log(this.props)
     return (
       <Layout location={this.props.location}>
         <Container>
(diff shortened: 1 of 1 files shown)


KyleAMathews:
Canonical link isn't always set btw. It's for when we repost a post from another site and we want Google juice to go there.

Jared-Dev:
Yea, now that you mention that... It likely isn't the url we'd want there anyway. I'll get that updated.

Jared-Dev:
Okay, @KyleAMathews I updated this to use the `href` value off of the `location` value from `props`.

www/src/templates/template-blog-post.js
```diff
@@ -73,7 +75,7 @@ class BlogPostTemplate extends React.Component {
         <link rel="canonical" href={post.frontmatter.canonicalLink} />
       )
     }
-
+    console.log(this.props)
```
KyleAMathews: Could you remove this? :-)

miltonbolonha:
@Jared-Dev @KyleAMathews @sidharthachatterjee , I spended many hour on this and I fixed it by removing og:url and canonical (temporary). 

Now LinkedIn inspector read my images and info's.

The thing is, the virtualDOM and the DOM are not the same for theses url's tags.

PR #14137, opened 2019-05-18
feat(image-sharp): add trim option

cpboyd:
## Description

Add the [@lovell/sharp](https://github.com/lovell/sharp/) `trim` option (and standardize `fit` & `background` from `fluid` to `fixed`).

I was trying different sharp options with a list of video screenshots (currently using `fixed()`), some of which have black bars, and saw that sharp offered a trim option.

Further, the `fit` and `background` settings weren't available on `fixed`.

Note: My original plan was to have it accept either a `boolean` or `int`, with `true` representing the `sharp` default threshold. (currently `10`)

Unfortunately, per https://github.com/graphql/graphql-spec/issues/488, GraphQL doesn't accept union types as inputs currently.  `GraphQLInt` coalesces true to 1 and false to 0 (0 is an invalid threshold value, so it's still OK to use to disable the setting)


packages/gatsby-plugin-sharp/src/index.js
```diff
@@ -150,6 +150,10 @@ async function generateBase64({ file, args, reporter }) {
     return null
   }
 
+  if (options.trim) {
+    pipeline = (typeof options.trim === 'number') ? pipeline.trim(options.trim) : pipeline.trim();
```
cpboyd: The type check isn't needed if using `GraphQLInt`, but I left it in case someone had a better suggestion.

I also considered having separate `trim` (as `GraphQLBoolean`) and `trimThreshold` and trying to coalesce them into a single `trim` option, but that seemed unnecessarily complicated just to retain a default of `10`.

packages/gatsby-transformer-sharp/src/extend-node-type.js
```diff
@@ -154,10 +154,22 @@ const fixedNodeType = ({
...
+      background: {
+        type: GraphQLString,
+        defaultValue: `rgba(0,0,0,1)`,
+      },
       rotate: {
         type: GraphQLInt,
         defaultValue: 0,
       },
+      trim: {
+        type: GraphQLInt,
```
cpboyd: After another look, it seems that the type should be `GraphQLFloat` per [libvps](https://github.com/libvips/libvips/blob/master/libvips/arithmetic/find_trim.c#L56)

packages/gatsby-transformer-sharp/src/extend-node-type.js
```diff
@@ -154,10 +154,22 @@ const fixedNodeType = ({
...
+        type: GraphQLString,
+        defaultValue: `rgba(0,0,0,1)`,
+      },
       rotate: {
         type: GraphQLInt,
         defaultValue: 0,
       },
+      trim: {
+        type: GraphQLFloat,
+        defaultValue: false,
```
freiksenet: Let's instead check that it hasn't been passed. It doesn't feel right to have different default value type than the actual type (actually pretty surprised it even works on GraphQL side)

cpboyd: @freiksenet As I mentioned, GraphQL coalesces false and true to 0 and 1 respectively.

I can just change it to 0, which is effectively the same.

cpboyd:
@freiksenet I made the requested changes, but squashed the prior commits before pushing the new changes which may have interrupted the normal review workflow.

Do I need to do make any other changes?

Issue #13543, opened 2019-04-22
Depreciated packages with yarn upgrade 


polishedwp:
## Description

When updating the packages in a starter ( `yarn upgrade --latest` ), there are several depreciated package error messages presented in the terminal. 

### Steps to reproduce

1. `gatsby new awesome-site`(Just an example, but affects most other starters I have tested)
2. `cd ./awesome-site`
3. `yarn upgrade —latest`- to upgrade to the latest packages from `package.json`.

### Expected result

In a perfect world there will be no errors messages, however, some of these errors messages are based on packages outside the scope of Gatsby development. 

### Actual result

Resulting error messages indicating depreciated packages. 

Gatsby —> [joi](https://www.npmjs.com/package/joi)
This error is directly related to Gatsby and its dependency `joi`:

```
warning gatsby > joi@12.0.0: 
This version has been deprecated in accordance with the hapi support policy (hapi.im/support). 
Please upgrade to the latest version to get the best features, bug fixes, and security patches. 
If you are unable to upgrade at this time, paid support is available for older versions (hapi.im/commercial).
```
#### Rest of the messages are issues with Gatsby dependencies 
Gatsby-plugin-offline  (see Issue  #13525  )

```
warning gatsby-plugin-offline > workbox-build > joi@11.4.0: 
This version has been deprecated in accordance with the hapi support policy (hapi.im/support). Please upgrade to the latest version to get the best features, bug fixes, and security patches. If you are unable to upgrade at this time, paid support is available for older versions (hapi.im/commercial).
```

Additional packages affecting Gatsby-plugin-offline (hoek and topo)

```
warning gatsby-plugin-offline > workbox-build > joi > hoek@4.2.1: 
This version has been deprecated in accordance with the hapi support policy (hapi.im/support). Please upgrade to the latest version to get the best features, bug fixes, and security patches. If you are unable to upgrade at this time, paid support is available for older versions (hapi.im/commercial).

warning gatsby-plugin-offline > workbox-build > joi > topo@2.0.2: 
This version has been deprecated in accordance with the hapi support policy (hapi.im/support). Please upgrade to the latest version to get the best features, bug fixes, and security patches. If you are unable to upgrade at this time, paid support is available for older versions (hapi.im/commercial).
```

It seems like the additional errors below are affected by hoek, which affects joi and topo  

```
warning gatsby > joi > hoek@4.2.1: 
[...]

PR #14146, opened 2019-05-18
chore(gatsby): Upgrade joi dep

millette:
Fixes warnings about joi, hoek and topo,


KyleAMathews:
What were the breaking changes from 14 to 15?

millette:
From https://github.com/hapijs/joi/issues/1776

* Joi.string().email() options changed. checkDNS and errorLevel are no longer supported, minDomainAtoms is replaced with minDomainSegments and tldBlacklist/tldWhitelist are replaced with tlds.allow and tlds.deny;
* Some email addresses considered valid before are no longer valid (specifically those with quoted strings "first last"@example.com and domain literals test@[127.0.0.1]).

polishedwp:
This is related to [13543](
https://github.com/gatsbyjs/gatsby/issues/13543#issuecomment-485870351)

millette:
In this case, all hapijs packages moved to scoped packages, so it's really more an upgrade path then a deprecation.

pieh:
This is weird - this change should break at least some of our tests, because those changes should require also changed imports inside `gatsby` package (for example in https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/joi-schemas/joi.js). It's pretty alarming that our tests passes here. I think those tests pass because other packages depends on `joi` also, so it end up in monorepo `node_modules`

@DSchau Do we need to install monorepo `node_modules` for e2e tests?

millette:
@pleh you're right, I completely forgot to change the require/imports with the new name. And yet the tests pass, strange indeed.

DSchau:
@pieh I think this is sort-of expected, and not sure there's anything we should do here regarding the tests. Also you mentioned e2e tests, you mean unit tests, right? It would slow down the unit tests quite a bit, but theoretically it'd be possible to avoid monorepo resolution and only install packages specific to that package--but I'm not sure this is worth it, generally!

The _why_ of why the unit tests still pass is that joi is hoisted as part of the monorepo (as you note!), so we get something like:

```
➜ yarn why joi
yarn why v1.15.2
[1/4] 🤔  Why do we have the module "joi"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "joi@14.0.0"
info Reasons this module exists
   - "_project_#gatsby-source-contentful" depends on it
   - Hoisted from "_project_#gatsby-source-contentful#joi"
info Disk size without dependencies: "356KB"
info Disk size with unique dependencies: "568KB"
info Disk size with transitive dependencies: "616KB"
info Number of shared dependencies: 4
=> Found "workbox-build#joi@11.4.0"
info This module exists because "_project_#gatsby-plugin-offline#workbox-build" depends on it.
```

So we got lucky here that gatsby-source-contentful depends upon Joi!

pieh:
> @pieh I think this is sort-of expected, and not sure there's anything we should do here regarding the tests. Also you mentioned e2e tests, you mean unit tests, right? It would slow down the unit tests quite a bit, but theoretically it'd be possible to avoid monorepo resolution and only install packages specific to that package--but I'm not sure this is worth it, generally!

No, I really meant e2e tests - unit tests would be quite difficult to workaround (as you mentioned). For e2e tests, those tests should have all deps and devDeps in their package.json (and are not part of yarn workspaces anyway), so avoiding monorepo dependencies should catch cases like that?

pieh:
Update on this - this seems more difficult that I initially thought to adjust our e2e-tests (and also wouldn't work in this case, because not only one of our other packages rely on `joi` - `start-server-and-test` and `workbox-build` both will install it as well)

millette:
Regarding the joi update, should I go ahead and change  gatsby-source-contentful  as well, including the correct require/import statements?

DSchau:
@millette yeah -- seems reasonable!

pieh:
#14694 was merged with all needed changes, so closing this one

PR #12371, opened 2019-03-07
fix(examples): use updated avatar URLs

FrederikBolding:
## Description
Changed avatar URLs to valid Instagram CDN URLs

## Related Issues
Fixes #12147

PR #14154, opened 2019-05-19
feat(examples): re-added gatsbygram avatar

FrederikBolding:
## Description

Re-added the avatar on the main gatsbygram page.

The avatar was removed due to issues in #12133 - but as of #12371 avatars are working again, I re-added the main page avatar.


examples/gatsbygram/src/components/Avatar.js
```diff
@@ -0,0 +1,49 @@
...
+        [`@media (min-width: 460px)`]: {
+          width: rhythm(3),
+          height: rhythm(3),
+        },
+        [`@media (min-width: 525px)`]: {
+          width: rhythm(4),
+          height: rhythm(4),
+        },
+        [`@media (min-width: 600px)`]: {
+          width: rhythm(6),
```
wardpeet: we've used the value `inherit` before, what's the reason for using `rhythm(6)`?

FrederikBolding: As the avatar for some reason is very low resolution, using 'inherit' doesn't look right, instead I chose 'rhythm(6)'.

I can change it back if you like it better the other way.

FrederikBolding: Yeah, we could just remove it - it's simpler that way

FrederikBolding: @wardpeet Just tested it - The image is way to small without the media query on my screen.

wardpeet: mind showing a screenshot?

FrederikBolding: image

image

First one is with my changes, last one is without media query

wardpeet: oh weird, well just go with yours, it's no big deal anyway 😄

PR #7680, opened 2018-08-27
[gatsby-transformer-sharp] Remove the width/maxWidth default values

NindroidX:
As of [this commit](https://github.com/gatsbyjs/gatsby/commit/7cda59b359f4281e67030c777a985c05c6943b92) gatsby-plugin-sharp can now calculate the image's width based on its height but for that the work the default values need to be removed.

I chose to simply remove the values since `gatsby-plugin-sharp` adds default values when both width and height aren't set (and the same goes for maxWidth/maxHeight).

Fixes: https://github.com/gatsbyjs/gatsby/issues/7663

Note: This depends on the changes in [this pr](https://github.com/gatsbyjs/gatsby/pull/7671) to work. Otherwise the width will be set to NaN and Sharp will throw an error.

Issue #6954, opened 2018-08-02
[ gatsby-image ] resolutions( height ) creates a cropped image


schwiet:
## Description

Using `height` with `resolutions` causes cropped image, while `width` does not. Documentation states that using _both_ will cause cropping.

### Steps to reproduce

In my case, using:

```
img {
    childImageSharp {
        resolutions( height: 400 ) {
            ...GatsbyImageSharpResolutions
```
For a source image that is 710px by 800px produces a slightly down-scaled, partially cropped image. In this case, for some reason, the image will only be scaled down until it reaches a width of 400px and then it is cropped, so the image generated by the above query is 400px by 400px with it's height cropped. If I run the same query but with `width: 355`, the result is an image that is 355px by 400px as expected.

### Expected result

The resulting image should be scaled in the same manner as when `width` is used, right? 

### Actual result

An arbitrary scale limit (of 400px in my case) in the width dimension is enforced.

### Environment

```  System:
    OS: macOS Sierra 10.12.2
    CPU: x64 Intel(R) Core(TM) i7-3770S CPU @ 3.10GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 8.11.3 - ~/.nvm/versions/node/v8.11.3/bin/node
    npm: 5.6.0 - ~/.nvm/versions/node/v8.11.3/bin/npm
  Browsers:
    Chrome: 67.0.3396.99
    Firefox: 61.0.1
    Safari: 10.0.2
  npmPackages:
    gatsby: ^1.9.277 => 1.9.277 
    gatsby-image: ^1.0.55 => 1.0.55 
    gatsby-link: ^1.6.46 => 1.6.46 
    gatsby-plugin-catch-links: ^1.0.24 => 1.0.24 
    gatsby-plugin-feed: ^1.3.25 => 1.3.25 
    gatsby-plugin-google-analytics: ^1.0.31 => 1.0.31 
    gatsby-plugin-manifest: ^1.0.27 => 1.0.27 
    gatsby-plugin-nprogress: ^1.0.11 => 1.0.14 
    gatsby-plugin-offline: ^1.0.18 => 1.0.18 
    gatsby-plugin-react-helmet: ^2.0.11 => 2.0.11 
    gatsby-plugin-sharp: ^1.6.48 => 1.6.48 
    gatsby-plugin-sitemap: ^1.2.25 => 1.2.25 
    gatsby-plugin-twitter: ^1.0.16 => 1.0.20 
[...]

Issue #6752, opened 2018-07-25
[gatsby-plugin-sharp] correctly infer width when passing only target height


danielberndt:
If I do 

```graphql
        file(relativePath: {eq: "img"}) {
          childImageSharp {
            fixed(width: 50) {
              ...GatsbyImageSharpFixed
            }
          }
        }
```

it's behaving as expected. The resulting image's height get automatically set according to the aspect ratio.

If on the other hand I want to fix the height:

```graphql
        file(relativePath: {eq: "img"}) {
          childImageSharp {
            fixed(height: 50) {
              ...GatsbyImageSharpFixed
            }
          }
        }
```

unexpected things happen. The width is not inferred by the aspect ratio. Instead it is set to 400 because it's set as a default [here](https://github.com/gatsbyjs/gatsby/blob/af799208eb210bc550caada43b5c1aa777dd82cf/packages/gatsby-plugin-sharp/src/index.js#L65).

For a more consistent behaviour It would be really nice to only use the fallback width of 400 if no height and no width has been set.

KyleAMathews:
Just checking that you're on the latest version of all your plugins? I remember this being an issue but also getting fixed.

danielberndt:
Ah sorry, I should have mentioned: I started a new project today and installed the `@next`version of all plugins:
```
"gatsby": "2.0.0-beta.59",
"gatsby-image": "^2.0.0-beta.7",
"gatsby-plugin-sharp": "^2.0.0-beta.7",
"gatsby-source-filesystem": "^2.0.1-beta.10",
"gatsby-transformer-sharp": "^2.1.1-beta.6",
```

PR #14159, opened 2019-05-19
feat(gatsby-source-contentful): Remove default image transform width

### LINKED ISSUE #14156: [gatsby-source-contentful] Image transforms still have default width
## Description

[This pull request](https://github.com/gatsbyjs/gatsby/pull/7680) removed the default image width for image transforms to enable you to only set a `height` and have the image infer the correct width automatically, just like it does with the height when setting a `width`.

This default still exists and can be removed in gatsby-source-contentful because [this pull request](https://github.com/gatsbyjs/gatsby/pull/6970) for gatsby-plugin-sharp infers the defaults automatically if neither width nor height are given.

### Steps to reproduce

Source an image from contentful and only set the height for the image transform with a query like:

```
query getImageFixedHeight {
  allContentfulImages {
    nodes {
      image {
        fixed(height: 160) {
          width
          height
        }
      }
    }
  }
}
```

### Expected result

`height` should be 160 and `width` should be scaled according to the aspect ratio of the image.

### Actual result

`width` is always 400.

### Environment
  System:
    OS: Windows 10
    CPU: (8) x64 Intel(R) Core(TM) i7-4700MQ CPU @ 2.40GHz
  Binaries:
    Yarn: 1.13.0 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
  Languages:
    Python: 2.7.15 - /c/Python27/python
  Browsers:
    Edge: 42.17134.1.0
  npmPackages:
    gatsby: ^2.4.3 => 2.4.3 
    gatsby-cli: ^2.5.8 => 2.5.13 
    gatsby-image: ^2.0.41 => 2.0.41 
    gatsby-plugin-material-ui: ^2.0.0-alpha.4 => 2.0.0-beta.0 
    gatsby-plugin-sharp: ^2.0.37 => 2.0.37 
    gatsby-plugin-styled-components: ^3.0.7 => 3.0.7 
    gatsby-source-contentful: ^2.0.40 => 2.0.56 
    gatsby-transformer-sharp: ^2.1.19 => 2.1.19

LuudJanssen:
## Description

This PR resolves the width of an image automatically using the aspect ratio when only the height is given in an image transform. Previously, a default width was used resulting in a cropped image.

This PR also creates unit tests to test this behaviour.

## Related Issues

Fixes #14156 
Related to #6752
Related to #6954
Related to [this pull request about removing the default width when using gatsby-transformer-sharp](https://github.com/gatsbyjs/gatsby/pull/7680/files)
Related to [this pull request about calculating the right dimensions / aspect ratio based on the passed arguments when using gatsby-plugin-sharp](https://github.com/gatsbyjs/gatsby/issues/6752)


LuudJanssen:
I have no idea why the end-to-end test is failing. It's failing on the `Gatsby Preview (Updating) "before each" hook for "updates when content is deleted"` because it reached a timeout of 9999ms. Could any of the core contributors help me resolve this?

LekoArts:
I re-run the CI, seems like it was a hiccup. What do you think about adding some notes to the README, e.g. about the defaults for width/height?

LuudJanssen:
@LekoArts, sure! I'll try to add the changes to the documentation as soon as possible!

LuudJanssen:
@LekoArts, I merged the `master` branch into my branch and added some documentation about the `height` parameter. I didn't change the `gatsby-plugin-sharp` documentation, as `gatsby-source-datocms` and `gatsby-source-sanity` don't support setting only the `height` parameter.

Should we file issues to those repositories to fix this difference?

Also, I noticed `gatsby-source-contentful`, `gatsby-plugin-sharp`, `gatsby-transformer-sharp`, `gatsby-source-santiy` and `gatsby-source-datocms` share a lot of code regarding the definition of the additional node types (like `resize`, `fixed`, `fluid`; see [this definition](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-source-contentful/src/extend-node-type.js#L519) for example) and utilties for resolvers (like calculating `width` based on `height` and vice versa). Would it be wise to create some sort of utility library? Or offload most of this work to some other part of Gatsby I don't know about?

LuudJanssen:
Also, @LekoArts, I'm experiencing build issues again.

pieh:
> Also, @LekoArts, I'm experiencing build issues again.

Sorry for that, sometimes development runtime tests are flaky. After restarting them we are good again.

> @LekoArts, I merged the `master` branch into my branch and added some documentation about the `height` parameter. I didn't change the `gatsby-plugin-sharp` documentation, as `gatsby-source-datocms` and `gatsby-source-sanity` don't support setting only the `height` parameter.
> 
> Should we file issues to those repositories to fix this difference?

I think it's worth at least notifying datocms and sanity we did this change and they might apply it too for uniformity

> Also, I noticed `gatsby-source-contentful`, `gatsby-plugin-sharp`, `gatsby-transformer-sharp`, `gatsby-source-santiy` and `gatsby-source-datocms` share a lot of code regarding the definition of the additional node types (like `resize`, `fixed`, `fluid`; see [this definition](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-source-contentful/src/extend-node-type.js#L519) for example) and utilties for resolvers (like calculating `width` based on `height` and vice versa). Would it be wise to create some sort of utility library? Or offload most of this work to some other part of Gatsby I don't know about?

There might be already lib for that. This could be good idea (depending how much that utility would abstract and make it easier to read code), but it's not critical to do

sidharthachatterjee:
Published in `gatsby-source-contentful@2.0.69`

Issue #14178, opened 2019-05-20
[gatsby-source-filesystem] unhandled rejection in createRemoteFileNode even if caught


antoinerousseau:
## Description

When calling `createRemoteFileNode` in `gatsby-node`'s `sourceNodes`, if a remote URL fails to download (e.g. if it's a 404), the whole process crashes with an `UNHANDLED REJECTION` error, even if we `.catch`.

https://github.com/antoinerousseau/gatsby-bug-mvce

### Steps to reproduce

- Create a new Gatsby project (`gatsby new gatsby-bug-mvce`)
- Add these lines to `gatsby-node.js`:
```js
const { createRemoteFileNode } = require("gatsby-source-filesystem")

exports.sourceNodes = ({ actions, store, cache, createNodeId }) => {
  const { createNode } = actions

  return createRemoteFileNode({
    url: "https://httpstat.us/404",
    store,
    cache,
    createNode,
    createNodeId,
  })
  .catch(err => {
    console.log("caught error:", err)
  })
  .then(node => {
    console.log("created node:", node)
  })
}
```
https://github.com/antoinerousseau/gatsby-bug-mvce/commits/master

### Actual result

```
yarn run v1.15.2
$ gatsby build
success open and validate gatsby-configs — 0.004 s
success load plugins — 0.239 s
success onPreInit — 0.004 s
success delete html and css files from previous builds — 0.010 s
success initialize cache — 0.015 s
success copy gatsby files — 0.037 s
success onPreBootstrap — 0.006 s
caught error: failed to process https://httpstat.us/404
HTTPError: Response code 404 (Not Found)
error UNHANDLED REJECTION

  Error: failed to process https://httpstat.us/404
  HTTPError: Response code 404 (Not Found)

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

### Expected result

A caught rejection means it's handled, so the process should display:
```
...
caught error: failed to process https://httpstat.us/404
created node: undefined
```
and continue.

### Environment

```
  System:
    OS: macOS 10.14.4
    CPU: (8) x64 Intel(R) Core(TM) i5-8259U CPU @ 2.30GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 11.14.0 - /usr/local/bin/node
    Yarn: 1.15.2 - /usr/local/bin/yarn
    npm: 6.9.0 - /usr/local/bin/npm
  Languages:
[...]

PR #14180, opened 2019-05-20
fix(gatsby-source-filesystem):  fix unhandled rejection by returning promise chain

wardpeet:
## Description
We weren't updating our promise chain which meant we weren't catching all variations of the promise. When using then or catch a new Promise is created. This one makes sure we are using the same promise.

Sadly I couldn't test if an unrejected promise was happening but I added a test to test for rejection.

### Repro from issue:

- Create a new Gatsby project (`gatsby new gatsby-bug-mvce`)
- Add these lines to `gatsby-node.js`:
```js
const { createRemoteFileNode } = require("gatsby-source-filesystem")

exports.sourceNodes = ({ actions, store, cache, createNodeId }) => {

  return createRemoteFileNode({
    url: "https://httpstat.us/404",
    store,
    cache,
    createNode,
    createNodeId,
  })
  .catch(err => {
    console.log("caught error:", err)
  })
  .then(node => {
    console.log("created node:", node)
  })
}
```
https://github.com/antoinerousseau/gatsby-bug-mvce/commits/master

## Related Issues
#14178

### DIFF
--- a/packages/gatsby-source-filesystem/src/__tests__/create-remote-file-node.js
+++ b/packages/gatsby-source-filesystem/src/__tests__/create-remote-file-node.js
@@ -104,6 +104,11 @@ describe(`create-remote-file-node`, () => {
         pipe: jest.fn(() => gotMock),
         on: jest.fn((mockType, mockCallback) => {
           if (mockType === type) {
+            // got throws on 404/500 so we mimic this behaviour
+            if (response.statusCode === 404) {
+              throw new Error(`Response code 404 (Not Found)`)
+            }
+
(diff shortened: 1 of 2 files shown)


DSchau:
Successfully published as gatsby-source-wordpress@3.0.62

antoinerousseau:
> Successfully published as gatsby-source-wordpress@3.0.62

I guess you meant `gatsby-source-filesystem` ;)

DSchau:
👀 

gatsby-source-filesystem@2.0.37

Correct--thanks for catching :)

PR #14191, opened 2019-05-20
[wip] [do-not-merge] skip installing monorepo deps for e2e tests

pieh:


.circleci/config.yml
```diff
@@ -82,9 +82,6 @@ commands:
     steps:
       - checkout
       - run: ./scripts/assert-changed-files.sh "<< parameters.trigger_pattern >>|<< parameters.test_path >>/*"
-      - <<: *restore_cache
-      - <<: *install_node_modules
```
DSchau: This should actually speed up the tests a little too-- 🤞

pieh:
Ok, this won't be that easy because when dependencies changes and we publish to verdaccio we also need to build package which requires package's devDeps

pieh:
Ok, this doesn't seem worth the effort (at least right now).

PR #14201, opened 2019-05-21
feat(gatsby-plugin-sharp): remove 3x DPi resolutions

### LINKED ISSUE #14190: Remove 3x images from gatsby-plugin-sharp
From Twitter Engineering blog: https://blog.twitter.com/engineering/en_us/topics/infrastructure/2019/capping-image-fidelity-on-ultra-high-resolution-devices.html

>The most modern screens are OLED. These screens boast some really great features like pure blacks, and are marketed as 3x scale. However, nearly no "3x scale" OLED actually has perfect 3x3 pixels per dot on their screen.
>
>This means that most OLED screens that say they are 3x resolution, are actually 3x in the green color, but only 1.5x in the red and blue colors. Showing a 3x resolution image in the app vs a 2x resolution image will be visually the same, though the 3x image takes significantly more data. Even true 3x resolution screens are wasteful as the human eye cannot see that level of detail without something like a magnifying glass.
>
>What does this change look like?
>
>There's no difference that the human eye can see, but will save 38% on data and 32% on latency on the capped image load for this particular example which is reflective of most images that load on Twitter.

So removing support for generating 3x images would both speed up loading images for users & it'd speed up generating the site since it'd be one fewer thumbnail to generate.

This wouldn't be a breaking change since thumbnail generation isn't directly controllable by the users. Thoughts @gatsbyjs/core?

thecodingaviator:
## Description

Removes all the 3x image sizing from gatsby-plugin-sharp as per [this](https://github.com/gatsbyjs/gatsby/issues/14190#issuecomment-494081137) comment.

I've also tried to go through all instances of the 3x image sizing and the code wherever required and tried to remove it. Consequently, changes have been made to the `/docs` too.

## Related Issues

Closes #14190


thecodingaviator:
@DSchau Please help me out on what to do here.

**Context:** This was not tested using `gatsby-dev` because my machine is extremely slow and always crashed on running `npm run watch`.

thecodingaviator:
Okay, I've got the tests to work, and IMO, I've changed all things relating to this plugin. This PR is ready for QA 🎊

thecodingaviator:
**Addnl. info:** I tried testing the build times but my build always gets stuck at `source and transform nodes`. Idk what's up. Since this is my first time working with gatsby, if someone could guide me, that'll be great!

thecodingaviator:
@DSchau I've made one change to the file which I felt was the only one. If you do stumble upon something else I need to change, please leave a review and I'd be glad to make changes :)

e2e-tests/gatsby-image/cypress/integration/fixed.js
```diff
@@ -23,12 +23,12 @@ describe(`fixed`, () => {
   })
 
   // TODO: figure out why these tests are failing
```
DSchau:
```suggestion
```

DSchau: nit (so won't block the PR!) but we should probably remove this, since they are passing 😅

thecodingaviator:
@DSchau , removed. Any idea what’s going wrong in CI?

thecodingaviator:
Sounds awesome!

DSchau:
Published as `gatsby-plugin-sharp@2.1.0` by the way!

PR #14208, opened 2019-05-21
fix(gatsby): improve `externals` webpack configuration to better support yarn PnP

### LINKED ISSUE #12948: The files within `.cache` and `public` should see their dependencies pre-resolved
## Description

The files within the `.cache` and `public` folders (here `public/render-page.js`) contain the bare, unresolved, imports. This causes issues under PnP (because as far as the loader goes the public files belong to the project rather than Gatsby), but also in general (depending on the hoisting, the wrong dependencies will be loaded).

The resolution should instead be pre-resolved, at least partially (relative paths, which would be portable across systems), so that you can guarantee that the files will use exactly the same version as you expect.

### Expected result

```js
exports["lib"] = factory(require("./node_modules/@reach/router/index.js"), require("./node_modules/core-js/modules/es6.object.assign.js"), ...);
```

### Actual result

Example of the `render-page.js`:

```js
exports["lib"] = factory(require("@reach/router"), require("core-js/modules/es6.object.assign"), ...);
```

arcanis:
## Description

This diff. A few notes:

- I looked at what packages were dependencies of Gatsby itself and assumed that they were meant to be pinned (and that all the other ones were meant to be provided by the consumer). I can change this list if it doesn't look right (for example I would have thought that `es6-promise` was meant to be pinned by Gatsby, but it isn't listed in its dependencies so I haven't added it).

- I tested locally (on `^2.2.9`) but there's at least one other PR needed for Gatsby master to unbreak w/ PnP so I can't test it against master at the moment. Hopefully the CI will provide good signal 😃

cc @pieh, since I think you implemented the external list

## Related Issues

Fixes #12948

### DIFF
--- a/packages/gatsby/src/utils/webpack.config.js
+++ b/packages/gatsby/src/utils/webpack.config.js
@@ -451,8 +451,11 @@ module.exports = async (program, directory, suppliedStage) => {
   }
 
   if (stage === `build-html` || stage === `develop-html`) {
-    // Packages we want to externalize to save space
+    // Packages we want to externalize to save some build time
+    // https://github.com/gatsbyjs/gatsby/pull/14208#pullrequestreview-240178728
     const externalList = [
+      `@reach/router/lib/history`,
+      `@reach/router`,
       `common-tags`,
       /^core-js\//,
       `crypto`,
@@ -467,7 +470,7 @@ module.exports = async (program, directory, suppliedStage) => {
       `zlib`,
     ]
 
-    // Packages we want to externalize because the user should provide them
+    // Packages we want to externalize because meant to be user-provided
     const userExternalList = [
       `es6-promise`,
       `minimatch`,
@@ -475,8 +478,6 @@ module.exports = async (program, directory, suppliedStage) => {
       `react-helmet`,
       `react`,
       /^react-dom\//,
-      `@react/router`,
-      `@reach/router/lib/history`,
     ]
 
     const checkItem = (item, request) => {
--- a/packages/gatsby/src/utils/webpack.config.js
+++ b/packages/gatsby/src/utils/webpack.config.js
@@ -451,47 +451,61 @@ module.exports = async (program, directory, suppliedStage) => {
   }
 
   if (stage === `build-html` || stage === `develop-html`) {
+    // Packages we want to externalize to save space
     const externalList = [
-      // match `lodash` and `lodash/foo`
-      // but not things like `lodash-es`
-      `lodash`,
-      /^lodash\//,
-      `react`,
-      /^react-dom\//,
-      `pify`,
-      `@reach/router`,
-      `@reach/router/lib/history`,
       `common-tags`,
+      /^core-js\//,
+      `crypto`,
+      `debug`,
+      `fs`,
+      `https`,
+      `http`,
+      `lodash`,
       `path`,
       `semver`,
-      `react-helmet`,
-      `minimatch`,
-      `fs`,
-      /^core-js\//,
-      `es6-promise`,
-      `crypto`,
+      /^lodash\//,
       `zlib`,
-      `http`,
-      `https`,
-      `debug`,
     ]
+
+    // Packages we want to externalize because the user should provide them
+    const userExternalList = [
+      `es6-promise`,
+      `minimatch`,
+      `pify`,
+      `react-helmet`,
+      `react`,
+      /^react-dom\//,
+      `@react/router`,
+      `@reach/router/lib/history`,
+    ]
+
+    const checkItem = (item, request) => {
+      if (typeof item === `string` && item === request) {
+        return true
+      } else if (item instanceof RegExp && item.test(request)) {
+        return true
+      }
+      return false
+    }
+
+    const isExternal = request => {
+      if (externalList.some(item => checkItem(item, request))) {
+        return `umd ${require.resolve(request)}`
+      }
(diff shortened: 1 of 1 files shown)


packages/gatsby/src/utils/webpack.config.js
```diff
@@ -451,47 +451,61 @@ module.exports = async (program, directory, suppliedStage) => {
   }
 
   if (stage === `build-html` || stage === `develop-html`) {
+    // Packages we want to externalize to save space
```
pieh: Primary goal of using externals is/was to shave some time (not space) when building bundle used to produce static html files (just to provide some context why we have in the first place)

packages/gatsby/src/utils/webpack.config.js
```diff
@@ -451,47 +451,61 @@ module.exports = async (program, directory, suppliedStage) => {
...
+      `https`,
+      `http`,
       `lodash`,
+      `path`,
+      `semver`,
       /^lodash\//,
+      `zlib`,
+    ]
+
+    // Packages we want to externalize because the user should provide them
```
pieh: This is mix of should and could:
- `react` and `react-dom` is expected to be provided by user (marked as gatsby's peerDependencies) - example of "should"
- `react-helmet` is not dep or peerDep of gatsby, but it's used in majority of gatsby projects to handle <head> tags - example of "could"

packages/gatsby/src/utils/webpack.config.js
```diff
@@ -451,47 +451,61 @@ module.exports = async (program, directory, suppliedStage) => {
...
+    const userExternalList = [
+      `es6-promise`,
+      `minimatch`,
+      `pify`,
+      `react-helmet`,
       `react`,
       /^react-dom\//,
-      `pify`,
-      `@reach/router`,
+      `@react/router`,
```
pieh: this should be @reach/router and it also should be in main externalList (not "optional" ) as this is hard dependency in gatsby (at least right now)

arcanis: Oops typo! Fixed 👍

packages/gatsby/src/utils/webpack.config.js
```diff
@@ -451,47 +451,62 @@ module.exports = async (program, directory, suppliedStage) => {
...
+      `es6-promise`,
+      `minimatch`,
+      `pify`,
+      `react-helmet`,
+      `react`,
+      /^react-dom\//,
+    ]
+
+    const checkItem = (item, request) => {
+      if (typeof item === `string` && item === request) {
```
wardpeet: should we refactor to something like?
`return typeof item === `string` && item === request || item instanceof RegExp && item.test(request)`

arcanis: I think the current writing might be more readable (I find complex conditions easier to parse when they're split in semantical blocks)

arcanis:
It changes one thing: a `require.resolve` is now applied to some of the imported files (so instead of being `require('semver')` it will be `require('/path/to/node_modules/semver/index.js')`).

I think it should be relatively safe - the one concern might be storing absolute paths inside a file, but since that only affects the build of `render-page.js` (which is a development build artifact) I assume it's ok.

packages/gatsby/src/utils/webpack.config.js
```diff
@@ -451,47 +451,62 @@ module.exports = async (program, directory, suppliedStage) => {
...
+        return true
+      } else if (item instanceof RegExp && item.test(request)) {
+        return true
+      }
+      return false
+    }
+
+    const isExternal = request => {
+      if (externalList.some(item => checkItem(item, request))) {
+        return `umd ${require.resolve(request)}`
```
arcanis: This is the line that matters (whereas previously we were always writing `umd ${request}`, breaking the dependency encapsulation and causing the generated artifacts to rely on the hoisting to work).

pieh:
Tangential to this PR develop doesn't seem to work with PnP because in https://github.com/gatsbyjs/gatsby/blob/9b7077e438c23e46530beb2c9601022356ba4127/packages/gatsby/src/utils/webpack.config.js#L154-L160 we don't wrap `event-source-polyfill` with `require.resolve` - I'll create separate PR for this shortly

btw, is there quick way to monkey-patch packages to avoid having to pack or publish packages when using PnP with berry? In yarn@1 I could go to local cache and just mess with files there, but here it seems like we get .zip files so that's not really feasible? That's just from quick look

packages/gatsby/src/utils/webpack.config.js
wardpeet: thanks for explaining

packages/gatsby/src/utils/webpack.config.js
wardpeet: make sense :)

arcanis:
Awesome, thanks!

> btw, is there quick way to monkey-patch packages to avoid having to pack or publish packages when using PnP with berry? In yarn@1 I could go to local cache and just mess with files there, but here it seems like we get .zip files so that's not really feasible? That's just from quick look

There are two ways:

- If you're a cowboy you can edit the zip archives in place using Emacs or Vim; both support reading and writing files even when stored within their container.

- Or the safer way is to use `yarn unplug <package-name>`, which will instruct Yarn to materialize the given package into the `.yarn/unplugged/` directory, ready for edition.

PR #14209, opened 2019-05-21
chore(gatsby): adds support for PnP in test-require-error

arcanis:
## Description

The current implementation of `test-require-error` was simply checking the error message to figure out whether the exception thrown was caused by the file not existing. This didn't work under PnP since we throw semantic errors (the message is different depending on the reason why the require call failed).

Instead we should check for the `QUALIFIED_PATH_RESOLUTION_FAILED` code, which means that the dependency check was successful but the resolver was unable to find a file on the disk that would match the specified request.

## Related Issues

Needed to test #14208 w/ PnP

### DIFF
--- a/packages/gatsby/src/utils/test-require-error.js
+++ b/packages/gatsby/src/utils/test-require-error.js
@@ -1,6 +1,12 @@
 // This module is also copied into the .cache directory some modules copied there
 // from cache-dir can also use this module.
 export default (moduleName, err) => {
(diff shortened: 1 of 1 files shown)


DSchau:
@arcanis not _exactly_ sure why this would cause the e2e test to fail, but it _seems_ legitimate. Re-running the tests just to verify no issues.

arcanis:
Everything is green but now it's #14208 that got a timeout - I guess it's an infra fail, right?

Issue #13355, opened 2019-04-14
build(webpack) : Window is not defined


pranshuchittora:
## Description

While using an npm lib called [canvas-confetti](https://www.npmjs.com/package/canvas-confetti),
everything is working awesome during development. While building the prod build using `gatsby build` it showed this error :point_down: 

![Screenshot from 2019-04-12 23-38-03](https://user-images.githubusercontent.com/32242596/56091039-2db74880-5ec7-11e9-93bd-9f74c6831cf6.png)

### Steps to reproduce

Install [canvas confetti](https://www.npmjs.com/package/canvas-confetti) & try to build a static html

### Expected result

It should build successfully

### Actual result

To get a successful build I removed the lib from local import & used a cdn version directly. Just for building.

### Environment
OS : Linux
Node v10.12.0

jonniebigodes:
@pranshuchittora as you may or may not know, Gatsby in development mode is way more permissive and let's you get a way with certain things that you won't when you issue a production build. When you issue a production build almost if not all of the work is done through the server side, more specifically through node and it does not have access to certain apis, like for instance this one, `window` and that's the reason you'll get that error. Bear in mind that most developers when working on a package, most of the time don't take in consideration Gatsby and other server side rendering generators. They push content based on the fact that you'll use it in the client either with CRA(Create React APP), or your own project setup and leave you to handle it. Also the [README](https://github.com/catdad/canvas-confetti#readme) clearly states that:

> You can then require('canvas-confetti'); to use it in your project build. Note: this is a client component, and will not run in Node. You will need to build your project with something like webpack in order to use this.

The operative word here being client.

Alternatively you can use a [ref](https://reactjs.org/docs/refs-and-the-dom.html) with a full blown class component like this:

```javascript
import React, { Component } from "react"

class ConfettiWithRef extends Component {
  constructor(props) {
    super(props)
    this.confettiCanvas = React.createRef()
  }
  componentDidMount() {
    if (typeof window !== "undefined") {
      const confetti = require("canvas-confetti")// es 6 import will not work, will throw error window not defined
      const confettis = document.createElement("canvas")
      this.confettiCanvas.current.appendChild(confettis)
      const myConfetti = confetti.create(confettis, { resize: true })
      myConfetti({ particleCount: 100, spread: 160 })
    }
  }


  render() {
    return (
      <div>
        <h1>confetti with a ref</h1>
        <div ref={this.confettiCanvas} />
      </div>
    )
  }
}

[...]

PR #14211, opened 2019-05-21
chore(docs): Add content for "client side only packages"

mbifulco:
## Description
This stub comes from a tip I came across here: https://github.com/gatsbyjs/gatsby/issues/13355#issuecomment-483265025

On occasion, gatsby developers may run into a need to use a client-side only npm / javascript dependency.  In my case, it was a shopify dependency, but this may come in handy for things like Intercom/Drift/etc, which are dynamic, client-only libraries which _won't_ work during SSR.

## Related Issues
https://github.com/gatsbyjs/gatsby/issues/13355


docs/docs/using-client-side-only-libraries.md
```diff
@@ -0,0 +1,71 @@
+---
+title: Using client side only packages
```
marcysutton:
```suggestion
title: Using Client-Side Only Packages
```
Using title case to match the [Gatsby Style Guide](https://www.gatsbyjs.org/contributing/gatsby-style-guide/).

docs/docs/using-client-side-only-libraries.md
```diff
@@ -0,0 +1,71 @@
...
+issue: https://github.com/gatsbyjs/gatsby/issues/13355
+---
+
+This is a stub. Help our community expand it.
```
marcysutton: This probably shouldn't be a stub, since it has content on it. We could always add a callout at the end for contributions.

docs/docs/using-client-side-only-libraries.md
```diff
@@ -0,0 +1,71 @@
...
+
+Please use the [Gatsby Style Guide](/contributing/gatsby-style-guide/) to ensure your
+pull request gets accepted.
+
+On occasion, you may need to use a function or library that only works client side. This usually is because the library in question accesses something that isn't available during server-side rendering (SSR).
```
marcysutton:
```suggestion
On occasion, you may need to use a function or library that only works client side. This usually is because the library in question accesses something that isn't available during server-side rendering (SSR), like [browser DOM](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model) methods.
```

docs/docs/using-client-side-only-libraries.md
```diff
@@ -0,0 +1,71 @@
...
+Reference error: Window is not Defined
+```
+
+## Workaround 1: Use a different library or approach
+
+Sometimes the simplest approach is to work around the problem. If you can re-implement your component using a plugin which _doesn't_ break SSR, that's probably best.
+
+## Workaround 2: Add client-side package via CDN
+
+In the component where you need it, load the package via CDN using a [`<script />`](https://www.w3schools.com/Tags/tag_script.asp) tag with `react-helmet`. You'll still need to check to see if `window` exists before using the library, so that SSR can still complete.
```
marcysutton:
```suggestion
In the component where you need it, load the package via CDN using a [`<script />`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script) tag with `react-helmet`. You'll still need to check to see if `window` exists before using the library, so that SSR can still complete.
```
I'd recommend MDN over W3schools as a learning resource.

docs/docs/using-client-side-only-libraries.md
```diff
@@ -0,0 +1,71 @@
...
+
+```jsx
+import { Helmet } from 'react-helmet'
+
+
+const MyComponent = (props) => {
+  if (typeof window !== undefined) {
+    // check for external package and use as needed here
```
marcysutton: What would be involved here?

docs/docs/using-client-side-only-libraries.md
```diff
@@ -0,0 +1,71 @@
...
+
+const Loading = () => <LoadingSpinner /> // custom loader visualization component
+
+const LoadableBuyButton = Loadable({
+  loader: () => import("./ShopifyBuyButton"),
+  loading: Loading,
+})
+
+export default LoadableBuyButton
+```
```
marcysutton: You could add a callout at the end for more ideas, like:

```
> **Note:** There are other potential workarounds than those listed here. If you've had success with another method, check out the [contributing docs](/contributing/docs-contributions/) and add yours!
```

docs/docs/using-client-side-only-libraries.md
mbifulco: It depends on the specific use case - any function call on a dependency which talks to `window` would fail.  In the case of shopify, the imported library is made available as something like `window.shopifyUi`, and for Intercom, `window.Intercom`, for example.  Some dependencies call on window under the covers somewhere, like the example in the original post from [this issue](https://github.com/gatsbyjs/gatsby/issues/13355#issue-432952792).  

I wasn't really sure how to flesh out the example because of that - any particular thoughts on what might be best?

docs/docs/using-client-side-only-libraries.md
mbifulco: Cool! I'll get these changes in.

docs/docs/using-client-side-only-libraries.md
mbifulco: Great idea 👍

docs/docs/using-client-side-only-libraries.md
marcysutton: It might be more clear when split into two comments:
```
// check for browser DOM readiness
if (typeof window !== undefined) {
    // use external package as needed here, e.g. `window.externalLibrary.method()`
```
What do you think?

mbifulco:
@marcysutton thanks so much for your help.  I think I've addressed everything here 🎉

LekoArts:
Should we also link to https://www.gatsbyjs.org/docs/debugging-html-builds/ ?

www/src/data/sidebars/doc-links.yaml
```diff
@@ -309,6 +309,8 @@
                   link: /docs/centralizing-your-sites-navigation/
                 - title: Rendering Sidebar Navigation Dynamically*
                   link: /docs/rendering-sidebar-navigation-dynamically/
+            - title: Using Client-Side Only Packages
+            - link: /docs/using-client-side-only-libraries
```
LekoArts: Should we also rename the file to `/docs/using-client-side-only-packages` ?

mbifulco: Good point - I think that makes sense.  I'll get that pushed up.

mbifulco:
Just double checking on this - anything else I can do here?

LekoArts:
You should await @marcysutton 	final review now :)

wardpeet:
I don't think workaround 2 is something we should advocate. using `if (typeof window !== undefined) {`
https://github.com/gatsbyjs/gatsby/pull/14211/files#diff-51d23587c5dad942b2081a4d676d19d7R27

`componentDidMount` or `useEffect` are the correct ways to do this.
https://reactjs.org/docs/integrating-with-other-libraries.html#how-to-approach-the-problem

Thanks for taking this on. This will help a lot of users.

mbifulco:
@wardpeet Great point.  I'll update the docs and get that back in front of y'all for review.

mbifulco:
@wardpeet I just pushed an update for this PR - I'm having trouble coming up with an easy to understand example to use here.  I think it may make sense to remove the code sample from that workaround, and instead rely on React's own docs.  What do you think?

docs/docs/using-client-side-only-packages.md
```diff
@@ -0,0 +1,84 @@
...
+        {/* etc */}
+      </React.Fragment>
+    )
+  }
+}
+```
+
+## Workaround 3: Load client-side dependent components with react-loadable
+
+Install [react loadable](https://github.com/jamiebuilds/react-loadable) and use it as a wrapper for a component that wants to use a client side only package.
```
wardpeet: React is advocating [loadable-components](https://github.com/smooth-code/loadable-components) instead in their docs https://reactjs.org/docs/code-splitting.html#reactlazy.

marcysutton:
I agree with @wardpeet that the code snippet helps. Since it's already in there, I think it's fine to leave as-is. 

We should update the PR to align with the React docs' recommendation of loadable-components, though.

docs/docs/using-client-side-only-packages.md
mbifulco: Good to know! Thanks for pointing that out.  I'll get this updated.

marcysutton:
As a side-note, this doc is so timely! I'm seeing a trend in Github issues:
- https://github.com/gatsbyjs/gatsby/issues/14336
- https://github.com/gatsbyjs/gatsby/issues/14504

mbifulco:
Updated the docs to use `loadable-components`.  Thanks for the tip @marcysutton & @wardpeet!

marcysutton:
Thanks so much for the PR, @mbifulco!

mbifulco:
Ah, fantastic! Thanks everyone for the help!

wardpeet:
Thank you ;) for your amazing work on getting these docs out the door! People will love this.

PR #14215, opened 2019-05-21
Fixing links in awesome-gatsby.md

luizdotpro:
The **https://www.gatsbyjs.org/docs/awesome-gatsby/** leads to this markdown file in Github, but
here some links are broken. I changed finally all the relative links (only a single one was working properly) and the links work now properly.

## Description

Fixed links in the **awesome-gatsby.md**

## Related Issues

The same problem may be present in other docu files, which I may slowly review and eventually send an MR.

### DIFF
--- a/docs/docs/awesome-gatsby.md
+++ b/docs/docs/awesome-gatsby.md
@@ -6,15 +6,15 @@ A curated list of interesting Gatsby community projects and learning resources.
 
 ## Site showcase
 
-See the [site showcase](/showcase/)
+See the [site showcase](https://www.gatsbyjs.org/showcase/)
 
 ## Starters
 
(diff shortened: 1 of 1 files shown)


marcysutton:
@luizdotpro what do you mean by "leads to this markdown file"? The entire Gatsbyjs.org site is sourced on Github, but we don't recommend viewing the content here. We use relative links in the source so they will work when running the site locally; therefore the links probably should stay as-is.

KyleAMathews:
Yeah, we don't encourage people to read the docs on github so all the links are intended for gatsbyjs.org. Thanks for the PR!

https://github.com/gatsbyjs/gatsby/blob/master/docs/README.md

PR #14227, opened 2019-05-22
chore(docs): Files Gatsby Looks for in a Plugin

florantara:
I realized the `main` field with the entry file of the plugin is needed on the `package.json`file in order for the plugin to work if it's installed from `npm`.

I discovered it by looking into the source of other plugins and trying things out, so I thought it'd be nice to have it in the doc.

## Description

* Adding info about the `main` field into the `package.json` line of the docs.

### DIFF
--- a/docs/docs/files-gatsby-looks-for-in-a-plugin.md
+++ b/docs/docs/files-gatsby-looks-for-in-a-plugin.md
@@ -6,7 +6,7 @@ title: Files Gatsby Looks for in a Plugin
 
 All files are optional unless specifically marked as required.
 
-- `package.json` — [required] this can be an empty object (`{}`) for local plugins
+- `package.json` — [required] this can be an empty object (`{}`) for local plugins. If the plugin will be installed from [npm](https://npmjs.com) it needs the `main`[field](https://docs.npmjs.com/creating-node-js-modules#create-the-file-that-will-be-loaded-when-your-module-is-required-by-another-application).
   - `name` is used to identify the plugin when it mutates Gatsby’s GraphQL data structure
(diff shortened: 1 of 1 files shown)


LekoArts:
Hey, thanks so much for the PR! 💪 

Adding information about the `main` field is a good idea as it's indeed required. I wouldn't put it into the first paragraph though but rather place it into the secondary list, like:

---

- `package.json` — [required] this can be an empty object (`{}`) for local plugins.
  - `name` is used to identify the plugin when it mutates Gatsby’s GraphQL data structure
    - if `name` isn’t set, the folder name for the plugin is used
  - [`main`](https://docs.npmjs.com/creating-node-js-modules#create-the-file-that-will-be-loaded-when-your-module-is-required-by-another-application) is the name of the file that will be loaded when your module is required by another application. The default name is `index.js`.
    - When publishing your Gatsby plugin to npm it is recommended to create an empty `index.js` file with the contents `//noop` like shown in this [example plugin](https://github.com/gatsbyjs/gatsby/tree/817a6c14543c73ea8f56c9f93d401b03adb44e9d/packages/gatsby-source-wikipedia).
  - `version` is used to manage the cache — if it changes, the cache is cleared
    - if `version` isn’t set, an MD5 hash of the `gatsby-*` file contents is used to invalidate the cache
    - omitting the `version` field is recommended for local plugins
  - `keywords` is used to make your plugin discoverable
    - plugins published on the npm registry should have `gatsby` and `gatsby-plugin` in the `keywords` field to be added to the [Plugin Library](/packages/)

florantara:
Fabulous, great edits! Thank you guys! 

Team work 💪🏻👊🏻

PR #14233, opened 2019-05-22
fix(gatsby-cli): make ink an optional dependency

wardpeet:
## Description
Fixes gatsby installation for node 6. Optional dependencies aren't breaking installs when they aren't compatible.

You can test it by installing `yarn add gatsby@ink-node-6`. You still need to add a resolutions section to package.json and `gatsby build` will work again. Gatsby develop is broken because of reac-hot-loader.

```
"resolutions": {
    "source-map": "0.6.1"
  },
```

Thanks @pieh for the solution

### DIFF
--- a/packages/gatsby-cli/package.json
+++ b/packages/gatsby-cli/package.json
@@ -26,8 +26,6 @@
     "fs-extra": "^4.0.1",
     "gatsby-telemetry": "^1.0.9",
     "hosted-git-info": "^2.6.0",
-    "ink": "^2.0.5",
-    "ink-spinner": "^3.0.1",
     "is-valid-path": "^0.1.1",
     "lodash": "^4.17.10",
     "meant": "^1.0.1",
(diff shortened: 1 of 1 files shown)


DSchau:
Published as gatsby-cli@2.6.1

thchia:
I might be missing something but won't this cause problems [here](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-cli/src/reporter/reporters/ink/index.js#L2)?

wardpeet:
@tchia It could break if installation goes wrong, that's true but now we break node 6 as ink has the engine field set to >= 8 so node 6 won't be able to install it anymore. Moving it to optional will not complain about this anymore as we fallback to yurnalist. Perhaps we should add an if else if ink is not available. (we probably wait for an issue to be responded)

thchia:
@wardpeet I had an install script that specified the `--no-optional` flag (for reasons) and this change caused the gatsby develop script to break. My understanding is we should consider putting the import/require of ink in a conditional if it's possible that it won't be available. 😃

wardpeet:
I agree we need a better check for this 👍 

what's the reason for using `--no-optional`

thchia:
I work behind a corporate firewall which (amongst other things) prevents things like prebuilt binaries being downloaded from github. Luckily the only dependencies we have that do this are optional, therefore we run our install with that flag.

It’s an edge case and for sure it’s “caused” by my environment, but I still agree that if something is an optional dependency it should not be statically imported.

wardpeet:
I was just curious :) but yes it's something we need to fix

PR #14234, opened 2019-05-22
chore(gatsby): accept multiple localhost addresses

### LINKED ISSUE #14084: accept multiple url variants for graphiql IDE
## Summary

In usability tests, we discovered that when people type the wrong thing in the browser (e.g. “localhost:8000/_ _ graphiql), we could either redirect to "correct" one or just support multiple different cases. This will help them spend less time looking for the right link.

### Basic example
I'm not sure what the best practice is here. Here are some ideas:

Support various number of underscores before graphiql e.g. _graphiql (1 underscore), __graphiql (2 underscores), ___graphiql (3 underscores), ____graphiql (4 underscores)
Support "graphql" and "graphiql"
Other ideas?

### Motivation

We observed that many users (including ourselves!) say "how many  underscores is this again? I can never remember" and "I always misspell Graphiql. How do you even say Graphiql??" haha.

thomascorthouts:
This is a proposal to match multiple versions of a path for the graphiql endpoint. I also _tried_ to create an e2e test to ensure multiple versions of the pattern get you to the graphiql IDE. I put them in a file where I think they could be, but I would love to get feedback on how to improve this or where it should live inside the code.

## Description

When I saw this issue, I could relate to the struggle of thinking about what the exact endpoint was again to play around in GraphiQL. This PR uses the option of express to use a pattern to match the path, this will match with all endpoints with graphql or graphiql with 1 or more leading underscores(`_`)

## Related Issues

Fixes #14084


packages/gatsby/src/commands/develop.js
```diff
@@ -125,7 +125,7 @@ async function startServer(program) {
 
   if (process.env.GATSBY_GRAPHQL_IDE === `playground`) {
     app.get(
-      `/___graphql`,
+      `/_+graphi{0,1}ql`,
```
pieh: Nice! 

Small nitpick:
What about extracting this to const variable that can be used in both route handlers and add some context in comment before that:
```
/**
 * Pattern matching all endpoints with graphql or graphiql with 1 or more leading underscores
 */
const graphqlEndpoint = `/_+graphi{0,1}ql
```

e2e-tests/development-runtime/cypress/integration/functionality/graphql-endpoint.js
```diff
@@ -0,0 +1,10 @@
+const endpoints = [`/___graphql`, `/_graphql`, `/___graphiql`]
+
+describe(`The GraphQL endpoint`, () => {
+  endpoints.forEach(endpoint => {
+    it(`Should appear on ${endpoint}`, () => {
+      cy.visit(endpoint)
+      cy.title().should(`eq`, `GraphiQL`)
+    })
+  })
+})
```
pieh: Awesome!

packages/gatsby/src/commands/develop.js
thomascorthouts: That sounds like a good idea, thanks!

packages/gatsby/src/commands/develop.js
```diff
@@ -123,9 +123,14 @@ async function startServer(program) {
 
   app.use(cors())
 
+  /**
+   * Pattern matching all endpoints with graphql or graphiql with 1 or more leading underscores
+   */
+  const graphqlEndpoint = `/_+graphi{0,1}ql`
```
wardpeet: swap `{0,1}` into `?`. I like this syntax more but I might be the only one :p (so if you don't agree, please ignore)
```suggestion
  const graphqlEndpoint = `/_+graphi?ql`
```

thomascorthouts: Good point, I didn't think about using an optional operator instead of a range.

DSchau:
Published as gatsby@2.6.2--again, thank you!

PR #12514, opened 2019-03-12
refactor(www): switch www to MDX

ChristopherBiscardi:
## Description

This swaps .org to use MDX instead of remark. It is not finished, but me and @johno have ironed out the upstream issues. It should be fair game for anyone to help push over the edge.

PR #14236, opened 2019-05-22
 fix(www): Blog posts not rendering

fk:
We missed refactoring one occurence of `markdownRemark` in the blog post template at https://github.com/gatsbyjs/gatsby/blob/dc331820f937ee01ff08a868028194ab0c9346e7/www/src/templates/template-blog-post.js#L30 when switching `www` to MDX in #12514.

This PR fixes that. ✌️ 
/cc @ChristopherBiscardi @jlengstorf @wardpeet


fk:
😽 🤗

PR #14237, opened 2019-05-22
chore(docs): Explain `setup` field at gatsby-plugin-feed README

### LINKED ISSUE #14136: docs: Add explanation about `setup` field at gatsby-plugin-feed README
## Summary

Add explanation about `setup` field (or, the more appropriate way of setting the top-level fields of feed filed) at `gatsby-plugin-feed` README

### Basic example

N/A

### Motivation

Yesterday, I had a journey as below.

1. Found out my feed file's top-level `link` field was unintentionally set as `https://github.com/dylang/node-rss` ([configuration](https://github.com/heejongahn/blog/blob/8d4c5a8f22196c3dbe4c0b86a9dd165a4b625306/gatsby-config.js#L11-L64))
2. Couldn't find how to set it on both [the package README](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-feed) and [the official doc](https://www.gatsbyjs.org/docs/adding-an-rss-feed/)
3. The plugin README mentions `node-rss`, so after reading its README I found out `site_url` and `feed_url`.
4. Tried [this commit](https://github.com/heejongahn/blog/commit/089decb516c99a323305ae4405fce5306445106b). Still no luck.
5. So I opened the plugin source code, found [this line](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-feed/src/gatsby-node.js#L87), and finally made it work. ([commit](https://github.com/heejongahn/blog/commit/d6518d08f1d5986bd23e2472c4d27d4c1b333426))

I think setting the top-level fields for the feed is quite a common needs, and it would be nice if it is documented. I'm also willing to contribute, so if the team agrees that this documentation might help, please let me know.

heejongahn:
I'm not a native English speaker, so please correct me if there's any awkward sentences/words.

## Description

Add an explanation about `setup` field of `gatsby-plugin-feed`.

## Related Issues

Closes #14136.

### DIFF
--- a/packages/gatsby-plugin-feed/README.md
+++ b/packages/gatsby-plugin-feed/README.md
@@ -78,10 +78,11 @@ Each feed must include `output`, `query`, and `title`. Additionally, it is stron
 
 `match` is an optional configuration, indicating which pages will have feed reference included. The accepted types of `match` are `string` or `undefined`. By default, when `match` is not configured, all pages will have feed reference inserted. If `string` is provided, it will be used to build a RegExp and then to test whether `pathname` of current page satisifed this regular expression. Only pages that satisifed this rule will have feed reference included.
 
-All additional options are passed through to the [`rss`][rss] utillity. For more info on those additional options, [explore the `itemOptions` section of the `rss` package](https://www.npmjs.com/package/rss#itemoptions).
+All additional options are merged with `query.site.siteMetadata`, and processed through `setup` option (which is a function). Return value of `setup` is, in turn, passed as a [feedOptions][rss-feedOptions] argument of the [`rss`][rss] utillity. For instance, you can return object containing `site_url` field from `setup` function to set the top-level `<link>` tag of the feed. For more info on those additional options, [explore the `itemOptions` section of the `rss` package](https://www.npmjs.com/package/rss#itemoptions).
 
 Check out an example of [how you could implement](https://www.gatsbyjs.org/docs/adding-an-rss-feed/) to your own site, with a custom `serialize` function, and additional functionality.
 
 _NOTE: This plugin only generates the `xml` file(s) when run in `production` mode! To test your feed, run: `gatsby build && gatsby serve`._
 
 [rss]: https://www.npmjs.com/package/rss
+[rss-feedOptions]: https://github.com/dylang/node-rss#feedoptions


packages/gatsby-plugin-feed/README.md
```diff
@@ -78,10 +78,11 @@ Each feed must include `output`, `query`, and `title`. Additionally, it is stron
...
 
-All additional options are passed through to the [`rss`][rss] utillity. For more info on those additional options, [explore the `itemOptions` section of the `rss` package](https://www.npmjs.com/package/rss#itemoptions).
+All additional options are merged with `query.site.siteMetadata`, and processed through `setup` option (which is a function). Return value of `setup` is, in turn, passed as a [feedOptions][rss-feedOptions] argument of the [`rss`][rss] utillity. For instance, you can return object containing `site_url` field from `setup` function to set the top-level `<link>` tag of the feed. For more info on those additional options, [explore the `itemOptions` section of the `rss` package](https://www.npmjs.com/package/rss#itemoptions).
```
DSchau: Documentation is a tricky thing. While we don't want others to go through what you experienced in #14136, I also don't necessarily think providing this would clarify to others exactly how this should be used.

Additionally -- we want to keep our documentation concise and approachable, and this change seems to lessen that and could introduce confusion as to _why_ we're telling someone this info they may not need.

marcysutton: It seems like we need an "advanced usage" section possibly. I'm still not quite grokking what the problem was here, though. I'll ask in #14136.

heejongahn:
@DSchau Thanks for the comment and suggestions! I understand and mostly agree with them. 

I'm still down to contribute if you think that adding this kind of info into [this doc](https://github.com/gatsbyjs/gatsby/blob/master/docs/docs/adding-an-rss-feed.md) would be helpful for the project. So in that case, please let me know, I'll try to come up with corresponding changes.

Otherwise – If you think that there would be better way of dealing with this issue (or maybe, it's not that big of an issue in the first place), I respect the gatsby team's opinion on the official docs.

I just wanted to prevent someone going through the same steps I had to go through (which I personally found quite frustrating). Please let me know if there's sth I could do to help, and thanks for the awesome project by the way 😄

DSchau:
@heejongahn few quick notes!

First of all, thanks so much for opening this PR. Can't convey enough how much we love that when someone runs into something, they make an effort to _fix_ it for others. So again, thank you!

However--I think this is a relatively advanced and niche use case, and as in my earlier statement, we want to keep the documentation approachable and focused for general purpose use cases. It's really unfortunate that you ran into this, but I think it'll only lead to confusion for others in documenting this relatively strange feature.

In general, gastby-plugin-feed tends to be a bit confusing and is _ripe_ for an upgrade of the actual code to be more user friendly and clear. From this basis, I'd rather make the concerted effort to make some breaking changes, and revamp the plugin to be much clearer and user-friendly. I think that will pay off much more handsomely than documenting the niche, advanced use cases that really shouldn't exist in the first place (or rather, should be clearer and less "magic" with the setup function).

As such, going to close this out with the intent to make some plugin changes in the future. Again, we appreciate your work here, and hope to see other PRs from you soon!

Thank you!

heejongahn:
Yup, I understand and totally agree with you. I'll be looking forward the updates of the plugin :) Thanks for the considerate comment!

PR #14238, opened 2019-05-22
fix(gatsby-telemetry): enable CI name for unsupported CI & fixed cpu count

jamo:
The ci_name doesn't appear to be a widely adopted standard but will help on a couple of places.

### DIFF
--- a/packages/gatsby-telemetry/src/telemetry.js
+++ b/packages/gatsby-telemetry/src/telemetry.js
@@ -185,10 +185,10 @@ module.exports = class AnalyticsTracker {
       nodeVersion: process.version,
       platform: os.platform(),
       release: os.release(),
-      cpus: cpus && cpus.length > 0 && cpus[0].model,
+      cpus: (cpus && cpus.length > 0 && cpus[0].model) || undefined,
       arch: os.arch(),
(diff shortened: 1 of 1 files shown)


packages/gatsby-telemetry/src/telemetry.js
```diff
@@ -185,10 +185,10 @@ module.exports = class AnalyticsTracker {
       nodeVersion: process.version,
       platform: os.platform(),
       release: os.release(),
-      cpus: cpus && cpus.length > 0 && cpus[0].model,
+      cpus: (cpus && cpus.length > 0 && cpus[0].model) || undefined,
       arch: os.arch(),
-      ci: ci.isCI,
-      ciName: (ci.isCI && ci.name) || undefined,
+      ci: ci.isCI || !!process.env.CI,
```
wardpeet: are you sure this is necessary?
https://github.com/watson/ci-info/blob/master/index.js#L52-L59

jamo: nice catch! 
Updating the PR

sidharthachatterjee:
Published in `gatsby-telemetry@1.0.10`

Issue #12157, opened 2019-02-27
API docs should link to the code definition in github for each function


KyleAMathews:
Right now if you go to an action e.g. https://www.gatsbyjs.org/docs/actions/#createPage there's no link to the actual code definition at https://github.com/gatsbyjs/gatsby/blob/026eaaa171ba04c9c9278e00f597f61b3b09a438/packages/gatsby/src/redux/actions.js#L111

Documentation is great but sometimes you can get more clarity by looking at the code itself.

JacobBlomgren:
@KyleAMathews I can do it! Should it be a text link or the github icon link used in other places in the docs?

KyleAMathews:
@pieh might actually already have handled this in a different PR he was already working on.

pieh:
The UI design is TBD (I didn't work on this part)

For `actions` reference page:

I just added support for providing code location in https://github.com/gatsbyjs/gatsby/pull/11597/commits/c81f6e0d0e3a8df0edb7cb6c785550100f16d6f0 (it's in not yet merged PR #11597)
![screenshot 2019-02-28 at 01 53 05](https://user-images.githubusercontent.com/419821/53581330-a6fc1580-3b7d-11e9-9579-87a2563d90b5.png)

After that is merged we can adjust query in https://github.com/gatsbyjs/gatsby/blob/master/www/src/pages/docs/actions.js#L73-L82 to use `codeLocation` to construct github links that always point to accurate position in github

For Gatsby Node/SSR/Browser API reference pages:

I've done proof of concept parser that finds where we "call" APIs - https://gist.github.com/pieh/ccbf0b2556fe3d19202ddaf61b881937 :
 - there is example output in gist as well - I just ran this on older branch so hightlighted lines are outdated in some cases
 - it doesn't find `gatsby-browser` API calls because I didn't handle ESM imports in my parser implementation
 - this should be turned into gatsby plugin (I think local plugin for .org site), so data can be used on .org site

PR #14240, opened 2019-05-22
feat(www): Add source links to API docs

LekoArts:
Fixes https://github.com/gatsbyjs/gatsby/issues/12157

- Adds a local plugin / transformer. Uses the filesystem and babel to find API calls
- Modifies the actions pages to link to the source
- Accounts for single source + multiple sources
- Uses commit SHA for links

Thanks @pieh for the help <3

---

Previews:

**Actions**

image

**Node API**

image


www/src/pages/docs/ssr-apis.js
```diff
@@ -65,5 +76,21 @@ export const pageQuery = graphql`
         ...DocumentationFragment
       }
     }
+    ssrAPIs: allGatsbyApiCall(filter: { group: { eq: "SSRAPI" } }) {
+      group(field: name) {
```
pieh: It might be worth to put this into fragment as this is duplicated in ssr, browser and node API reference pages (I can help with that)

pieh:
How does it looks on narrow viewports (~mobile)? APIs with long name might wrap this to newline I think (for example https://www.gatsbyjs.org/docs/node-apis/#setFieldsOnGraphQLNodeType )

www/src/pages/docs/ssr-apis.js
LekoArts: Yeah, will put this into a fragment when I've got time tomorrow (not sure)

LekoArts:
I tested it on mobile and hence put the `marginRight` on the first part of the header. Should look fine

image


image

www/src/pages/docs/ssr-apis.js
LekoArts: Done!

PR #14253, opened 2019-05-22
feat(gatsby-remark-autolink-headers): Support {#custom-id} header syntax

yunabe:
## Description

Support custom heading ID syntax in gatsby-remark-autolink-headers (disabled by default)

```
### My Great Heading {#custom-id}
```
https://www.markdownguide.org/extended-syntax/#heading-ids


packages/gatsby-remark-autolink-headers/src/__tests__/index.js
```diff
@@ -182,4 +182,23 @@ describe(`gatsby-remark-autolink-headers`, () => {
...
+# With *Bold* {#cusgom-withbold}
+    `)
+    const enableCustomId = true
+
+    const transformed = plugin({ markdownAST }, { enableCustomId })
+
+    visit(transformed, `heading`, node => {
+      expect(node.data.id).toBeDefined()
+
+      expect(node).toMatchSnapshot()
```
DSchau: Sort of a nit: but generally I'd argue we way overuse snapshots and they make future failures more confusing than they need to be.

Could you rather test for a specific id, e.g. something like

```
const customId = `custom-id`

const markdownAST = remark.parse(`
# Heading With a Custom Id {#${customId}}
`)

// later

visit(transformed, `heading`, node => {
  expect(node.data.id).toBe(customId)
})
```

Obviously you can scale this however you'd like if there's value in testing for multiple instances.

packages/gatsby-remark-autolink-headers/src/index.js
```diff
@@ -20,13 +20,30 @@ module.exports = (
...
 ) => {
   slugs.reset()
 
   visit(markdownAST, `heading`, node => {
-    const slug = slugs.slug(toString(node), maintainCase)
-    const id = removeAccents ? deburr(slug) : slug
+    let id
+    if (enableCustomId && node.children.length > 0) {
+      const last = node.children[node.children.length - 1]
+      const match = /^(.*?)\s*\{#([\w-]+)\}$/.exec(toString(last))
```
DSchau: This seems a little messy. Is this code from elsewhere that's been copy/pasted or rather, can this be optimized/tweaked?

yunabe: Do you mean the regex looks messy or the entire code structure is too complicated?
In the former case, I will write unit tests for the regex.

> Is this code from elsewhere that's been copy/pasted
No, I implemented this from scratch.

DSchau: Just the combination of the regex and the other code there.

Maybe we can implement something like in [reactjs.org](https://github.com/reactjs/reactjs.org/blob/master/plugins/gatsby-remark-header-custom-ids/index.js)

yunabe: I hesitated to look at another repository maintained under a different copyright and I had not looked into reactjs.org version before your comment.

Anyway, the major difference between my PR and the code you mentioned is that

- my PR extracts `#{id}` only from the last node.
- reactjs.org version extract `#{id}` from the entire header node text.

For this markdown example, which I added to the test cases:

```
# Invalid {#this_is_italic}
```

My version does NOT extract a custom ID because `_is_` is italic. reactjs version extracts `#thisisitalic` and does NOT remove `{#this_is_` part from the header text.
I think the behavior of my version is easy to understand for users.

What do you think?

packages/gatsby-remark-autolink-headers/src/__tests__/index.js
yunabe: I agree. I rewrote this test without using snapshots. Does it look better? If you like this, I will rewrite other tests with this style in another PR.

packages/gatsby-remark-autolink-headers/src/__tests__/index.js
```diff
@@ -182,4 +183,42 @@ describe(`gatsby-remark-autolink-headers`, () => {
...
+      {
+        id: `custom-heading-two`,
+        text: `Heading Two`,
+      },
+      {
+        id: `custom-withbold`,
+        text: `With Bold`,
+      },
+      {
+        id: `invalid-thisisitalic`,
```
wardpeet: I'm no markdown expert but why wouldn't I want to get the following?
```json
{
  "id: "this_is_italic",
  "text": "Invalid",
}
```

packages/gatsby-remark-autolink-headers/src/index.js
wardpeet: could you add a comment above this line that tells us what this regex does?
Something like: `// Matches {#<word>} at the end of a string

yunabe: Done

packages/gatsby-remark-autolink-headers/src/__tests__/index.js
yunabe: Because `# Invalid {#this_is_italic}` is parsed as `<h1>Invalid {#this<i>is</i>italic}</h1>` (Three child nodes under <h1>) before this plugin.
The current implementation extracts `{#id}` only from the last child of header nodes.

yunabe:
Added additional tests. Could you take another look?

packages/gatsby-remark-autolink-headers/src/__tests__/index.js
wardpeet: Oh because we convert markdown to html. Maybe that's a mistake, maybe we should use the raw value instead to create the idea. Nonetheless, we can iterate on this in a follow PR when people are complaining about it.

Thanks for the explanation! 💯

wardpeet:
published in `gatsby-remark-autolink-headers@2.0.17` :tada:

yunabe:
Thank you for the reviewing, Ward.
Also, it's great to see my PR was released so quickly. Amazing😄

PR #14254, opened 2019-05-22
fix(www): remove animation property from background in "how gatsby works"  diagram

ocularrhythm:
The vertical animation combined with the horizontal striped animation could trigger vertigo with some users. I myself do not suffer from vertigo, but it even made my eyes jump a little bit. If the contrasting animations served a non-aesthetic purpose I would be hesitant to create a pull request. However, I believe this section looks just as nice without it and is somewhat unnecessary.

Maybe an alternative to keep everyone happy, is to provide a mechanism to pause any animations. The `prefers-reduced-motion` is used in a few places in the site, but that is still not supported in all browsers.

Since I am coming into this project for the first time, I don't know of all of the decisions that have been made previously so forgive me if I'm coming at this incorrectly. 

---

As a side note, I recently begun thinking about things like this after reading this [great blog post highlighting the use of animations or scroll effects](https://alistapart.com/article/designing-safer-web-animation-for-motion-sensitivity/). Here is another good post for those interested in [using responsible motion on websites](https://source.opennews.org/articles/motion-sick/).


pieh:
For context - this removes animation of skewed stripes in the background in "How Gatsby works" diagram on home page.

I think that if this used reduced motion query here, this would be approved unanimously as it does make a lot of sense. Changing this for every user will probably result in longer discussion. But let's see what others have to say

/cc @fk @marcysutton

marcysutton:
Great issue, thanks for opening it! 

I believe this animation actually fails [WCAG success criterion 2.2.2](https://www.w3.org/TR/UNDERSTANDING-WCAG20/time-limits-pause.html) for "pause stop hide" because it lasts longer than 5 seconds with no way to turn it off. I typically recommend pairing `prefers-reduced-motion` (which has decent support now) [with a toggle button](https://codepen.io/marcysutton/pen/yqVVeY) to raise support to include all browsers. But for this instance where the motion really doesn't seem critical to understanding, I would also question the necessity of it to begin with–it would be much simpler to remove it than to add multiple UI affordances.

fk:
Hey @ocularrhythm! 👋 

👍 I voted for dropping these animations before — not only in the diagram, but also for the button hover — but had no other reason than personal preference. If I remember correctly, @KyleAMathews was a little more attached to the animation than me?

@ocularrhythm @marcysutton What about the button hovers, where the same animation is used?

ocularrhythm:
> 
> 
> Great issue to discuss, thanks for submitting it!
> 
> I believe this animation actually fails [WCAG success criterion 2.2.2](https://www.w3.org/TR/UNDERSTANDING-WCAG20/time-limits-pause.html) for "pause stop hide" because it lasts longer than 5 seconds with no way to turn it off. I typically recommend pairing `prefers-reduced-motion` (which has decent support now) [with a toggle button](https://codepen.io/marcysutton/pen/yqVVeY) to raise support to include all browsers. But for this instance where the motion really doesn't seem critical to understanding, I would also question the necessity of it to begin with–it would be much simpler to remove it than to add multiple UI affordances.

I was hesitant to quote the WCAG guidelines, as that particular criterion is specifically talking about "moving, blinking, scrolling, or auto-updating information". Since the animation is a background, I didn't really consider it to qualify as "information."

@fk I actually didn't notice the button hovers until now. They seem quite subtle so don't seem necessary to remove entirely. They should definitely be addressed with the `prefers-reduced-motion` query though I would think.

marcysutton:
> Since the animation is a background, I didn't really consider it to qualify as "information."

I tend to focus on user impact, and a moving background has an impact regardless of whether it's in the foreground or background! As for button interactions, I think `prefers-reduced-motion` for those would be great in a separate PR.

fk:
🙏 Thank you @ocularrhythm, and @pieh for getting this in!

> As for button interactions, I think prefers-reduced-motion for those would be great in a separate PR.

Seconding that!

PR #14266, opened 2019-05-23
chore(starters): add gatsby-agency-portfolio gatsby-localization-app-starter gatsby-docs gatsby-ecommerce-website

JacobKnaack:
## Description

Adding 4 Cosmic JS starters to the starter.yml file


docs/starters.yml
```diff
@@ -2619,3 +2619,52 @@
...
     - Web App Manifest
+- url: https://cosmicjs.com/apps/gatsby-agency-portfolio/demo
+  repo: https://github.com/cosmicjs/gatsby-agency-portfolio
+  description: Static Webpage for displaying your agencies skills and past work.  Implements 4 sections for displaying information about your company, A home page, information about services, projects, and the people in your organization.
+  tags:
+    - Landing Page
+    - Blog
+    - Portfolio
+    - Cosmic JS
+    - React
```
pieh: Can we drop "React" tag? No starter used that yet, but it would apply to all of them and it might be confusing that some starters have it and some don't

docs/starters.yml
```diff
@@ -2619,3 +2619,52 @@
...
+    - Node JS
```
pieh: Same here I think

pieh: (and both react and node JS comments apply to all starters)

docs/starters.yml
```diff
@@ -2619,3 +2619,52 @@
...
+  features:
+    - Gatsby localization starter app
+- url: https://cosmicjs.com/apps/gatsby-docs/demo
+  repo: https://github.com/cosmicjs/gatsby-docs-app
+  description: Be able to view and create documentation using Gatsby and Cosmic JS. Leveraging the speed and high powered APIs of the Gatsby framework and the simplicity and scalability of Cosmic JS.
+  tags:
+    - React
+    - Node JS
+    - Cosmic JS
+    - Docs
```
pieh:
```suggestion
    - Docs
    + Documentation
```
Documentation is already used, so let's not create new tag

docs/starters.yml
```diff
@@ -2619,3 +2619,52 @@
...
+    - Projects
+    - People
+- url: https://cosmicjs.com/apps/gatsby-localization-app-starter/demo
+  repo: https://github.com/cosmicjs/gatsby-localization-app-starter
+  description: A localized Gatsby starter application powered by Cosmic JS.
+  tags:
+    - React
+    - Node JS
+    - Cosmic JS
+    - Localization
```
pieh:
```suggestion
    - Localization
    + i18n
```
(we already have tag like that and I think it makes sense here?)

docs/starters.yml
```diff
@@ -2619,3 +2619,52 @@
...
+  features:
+    - manage docs in static web file format for zippy delivery
+- url: https://cosmicjs.com/apps/gatsby-ecommerce-website/demo
+  repo: https://github.com/a9kitkumar/Gatsby-Ecommerce
+  description: A localized Gatsby starter application powered by Cosmic JS.
+  tags:
+    - React
+    - Node JS
+    - Cosmic JS
+    - Ecommerce
```
pieh:
```suggestion
    - eCommerce
```

docs/starters.yml
JacobKnaack: Cool, removed React tags.

docs/starters.yml
JacobKnaack: Removed node js tags as well.

docs/starters.yml
JacobKnaack: Updated localization tag.

docs/starters.yml
JacobKnaack: Updated Docs to Documentation.

docs/starters.yml
JacobKnaack: Updated Ecommerce to eCommerce.

PR #14269, opened 2019-05-23
chode(docs): updated quick links

Athul-CA:
## Description
Updated most of tge quick-link bugs I could find and did a small change in linking to Prismic's Plugin Docs

### DIFF
--- a/docs/docs/sourcing-from-prismic.md
+++ b/docs/docs/sourcing-from-prismic.md
@@ -56,7 +56,7 @@ API_KEY=paste-your-secret-access-token-here-wou7evoh0eexuf
 
 _Note: If you want to locally build your project you'll also have to create a `.env.production` file with the same content._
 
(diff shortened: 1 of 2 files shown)


docs/docs/sourcing-from-prismic.md
```diff
@@ -56,7 +56,7 @@ API_KEY=paste-your-secret-access-token-here-wou7evoh0eexuf
...
+Now you need to configure the plugin (See all [available options](https://github.com/angeloashmore/gatsby-source-prismic/tree/master/README.MD)). The `repositoryName` is the name you have entered at the creation of the repository (you'll also find it as the subdomain in the URL). The `linkResolver` function is used to process links in your content. Fields with rich text formatting or links to internal content use this function to generate the correct link URL. The document node, field key (i.e. API ID), and field value are provided to the function. This allows you to use different [link resolver logic](https://prismic.io/docs/javascript/query-the-api/link-resolving) for each field if necessary.
```
pieh: How about using `https://www.npmjs.com/package/gatsby-source-prismic#how-to-use` link? One you added also doesn't work btw because it should be lowercase `.md` :)

Athul-CA: 🤦‍♂️🤦‍♂️🤦‍♂️Dumb me...     But I'll update the link with the npm url

Athul-CA:
@pieh updated it

Issue #6438, opened 2018-07-13
Add resolve-url-loader to gatsby-plugin-sass


cezarneaga:
## Summary

In sass files that use paths to images, fonts, svg... we need to have the resolve-url loader so that webpack can find the files

### Basic example

```
.header-banner.green {
      background: map-get($group-accelerate, start)
        url('../../../images/new/crowdfunding/banner.png') 45px 30px no-repeat;
    }
    .header-banner {
      color: $white;
      margin: -2px 0 30px 0;
      border-radius: $global-radius;
      padding: 30px;
      background: map-get($foundation-palette, primary)
        url('../../../images/new/crowdfunding/banner.png') 45px 30px no-repeat;
      ::selection {
        background: $white;
        color: $black;
      }
```
this code doesn't compile to css without it.

### Motivation
Common stuff like this should work out of the box.
What do you think?

thanks
tagging: @jquense  :)

jquense:
hmm i've never run into this.. I'm using the latest sass plugin right now and it has no problem with the files.

Can you maybe put together a very small repo demonstrating the issue? and what version of everything you are running?

cezarneaga:
sure. will do and get back to you.

cezarneaga:
Hi @jquense,

here is a sample repo: https://github.com/cezarneaga/sass-example i made cloning the default starter and adding sass plugin.
am i doing something wrong?

cezarneaga:
@jquense did you try the repo?

fardarter:
This needs to be on front page docs. Just took two hours from me.

jquense:
@cezarneaga sorry I've been busy :/ I'll take look, when you say urls need to be relative, do mean like background images? I think you're seeing the normal sass import behavior, which inlines the file in the file your importing. So it'd make sense that important would be relative to the entry in that case

I usually use Sass with css modules where there is no "entry" file since you import each file in the component you want to use the styles with. Or in the normal sass usage, I tend to have the font/image path as a variable. 

I'm not sure if changing the default behavior (if that's what this is) is a good idea as a default but maybe yeah adding note about potential pitfalls is good.

Edit: this pitfalls is documented on the sass-loader page! https://github.com/webpack-contrib/sass-loader/blob/master/README.md#problems-with-url

KyleAMathews:
Yeah, it suggests adding https://github.com/bholloway/resolve-url-loader which perhaps we could do.

jquense:
[...]

PR #14272, opened 2019-05-23
feat(gatsby-plugin-sass): add option to enable resolve-url-loader

### LINKED ISSUE #7776: How to use resolve-url-loader with gatsby-plugin-sass in v2?
In v1, I could do this to make relative `url()` in Sass work:

```js
// gatsby-node.js

exports.modifyWebpackConfig = function (config, stage) {
  if (config.config._loaders.sass.config.loaders) {
    config.config._loaders.sass.config.loaders.splice(config.config._loaders.sass.config.loaders.length - 1, 0, 'resolve-url-loader');
  }
  return config;
}
```

It's ugly but inserts the resolve-url-loader as the previous-to-last item in the list of style loaders.

How to achieve the same in v2?

carlosruana:
## Description

Currently plugin just resolves url() paths relative to the main sass entry file.
resolve-url-loader adds the ability to use paths relative to the file where the url() is being used.

By default if not configured in the gatsby-plugin-sass, the loader is not being used, but you can activate it and also configure resolve-url-loader plugin with the options it provides.

Usage has been also documented in the README.md of the gatsby-plugin-sass

## Related Issues

Fixes #7776, #6438


wardpeet:
Cool 😎 ! Any reason why we shouldn't enable this by default? Any possible breaking scenarios?

carlosruana:
> Cool 😎 ! Any reason why we shouldn't enable this by default? Any possible breaking scenarios?

Well, sass and sass-loader do not provide url rewriting by default, they recommend using resolve-url-loader to solve that. So, the expected behavior is using paths in url() relative to the entry file and I would not change defaults as it can confuse plugin users. It would break also projects using images already created with those paths relatives to the entry file and not relative to the file where the image is loaded itself.

Activating it will benefit though users trying to generate documentation from external projects which uses relative url() to the file without having to modify their projects code (that is mainly the reason I need it). Also users generating new documentation with Gatsby from scratch will benefit by being able to add short relative paths as url(./image_name.svg) and not having to remember the main entry file.

slang25:
Some anecdotal justification for this change:

I've just picked up gatsbyjs for the first time for a side project (cause it looks great!), and have started to migrate my site to it. This is the first issue I faced with gatsby and I wasn't ready to hack around the problem, and until seeing this PR I was about to drop it and try a different framework altogether!

PR #14276, opened 2019-05-23
feat(themes): gatsby official themes

amberleyromo:
_Context: I started messing around with themes awhile ago. As part of that started porting Dan Abramov's [overreacted.io](https://overreacted.io/) (which was built off the gatsby blog starter) to a theme. I modified my own blog site from overreacted, and I've seen others do the same-- I thought it would be helpful as a theme! We've decided to port it over into the monorepo and take it across the finish line as the gatsby blog theme._

There's still much to do:

- User can install the blog theme, and have a fully functional blog site with default text placeholders a default Gatsbyesque color palette (without customizing anything except installing the theme).
- User can customize the following site details:
  - Site title
  - Author name
  - Site description
  - Site URL
  - Footer content
  - Bio content
  - Blog posts
- User by default has dark mode / light mode functionality
- User can opt-in to a variety of predefined, accessible color palettes (_"nice-to-have" for initial publish, non-blocking_
  - Goal to have at least one alternative palette as an example
  - Have a full discussion on config-based color palettes
- User can opt to use a fully custom color palette. (shadowing)
- User can easily select their own header/body text typefaces. (shadowing)
- User has full confidence that the theme ships fully audited for accessibility, from first stable release.

## Bugs/Todos

- [x] (#14377) update design tokens to use theme-ui
  - [x] use theme-ui for dark/light mode
  - [x] compatible with typography.js
- [x] ~Prism seems to break [MDX inlineCode](https://github.com/gatsbyjs/gatsby/pull/14377/files#r289071919)~
- [x] (#14520) Audit default plugins (ex., the ported site included the google analytics plugin, which doesn't need to ship with a blog theme)
- [x] (#14519) Update default colors to be "Gatsby-esque"
- [x] (#14523) Audit for any hard-coded text so we’re always pulling from config (ex. footer component, SEO component. Hard-coded while we decide about the fragment pattern)
- [x] (#14523) Audit default text to be genericized more
- [x] Set up generic abstraction using schema customization (@freiksenet, @ChristopherBiscardi)
- [x] Test for accessibility (meeting planned with @marcysutton)
- [x] (#14596) Refactor the dark/light toggle (@jxnblk)

## Rollout todos
- [ ] ⛔️: Run through empty install to make sure we don’t get warnings/errors
- [ ] ⛔️: Add a note to the README about stopping/starting when adding the first post to hide default posts

## Discussion points 

- ~How to style imported MDX fragments (e.g. bio fragment)~
- ~Get a CSS (modules) theme put together to make themes more approachable to people who aren’t comfortable with CSS-in-JS?.~
- Further discussion on use of [`Global`]?(https://github.com/gatsbyjs/gatsby/pull/14377/files#r288577341)
- Use shadowing for theme changing / typography changes.


ChristopherBiscardi:
❤️ ❤️ This is fantastic!

I'd like to add a basic schema customization implementation here too so that we're querying `allBlogPost` instead of `allMdx`. It can be much less complicated than [the one I use in my blog](https://github.com/ChristopherBiscardi/christopherbiscardi.github.com/blob/24227770908405592bb916938944579ba50b43c5/www/gatsby-node.js#L22-L36) and be a concrete type for now instead of an interface.

> How to style imported MDX fragments (e.g. bio fragment)

Maybe we make this a React components anyway? The other option I see being possible is to use an MDXProvider in the component that renders the MDX. I like the .mdx fragment shadowing pattern for replaceable content chunks and this could translate well into sourcing the bio from headless CMSs as we work on the data abstractions.

> Get a CSS (modules) theme put together to make themes more approachable to people who aren’t comfortable with CSS-in-JS?

Approachable to who specifically and for what use cases? Theme authors? I'm hesitant to fragment `gatsby-theme-blog` (for example) into a "css-in-js implementation" and a "CSS Modules implementation". This would bring a higher maintenance burden (same changes in two places, rectifying differences in implementations like custom props vs themeproviders) and confusion for users as to which to use. People will look to the official themes for "best practices" and when thinking about how shadowing and composition interact with the most common styling practices the best practice is css-in-js right now with the css prop.

The way I could see this being supported in the future is through theme-ui "backends", where you could swap out [emotion](https://emotion.sh/) (which currently powers theme-ui) for [linaria](https://github.com/callstack/linaria) if you wanted static extraction, etc. This doesn't seem like a high priority though.

Overall I'd say alternative CSS implementations are better off as blog posts than as officially maintained themes.

themes/gatsby-theme-blog/src/components/layout.js
```diff
@@ -1,72 +1,130 @@
...
+import { rhythm, scale } from "../utils/typography";
+import sun from "../../content/assets/sun.png";
+import moon from "../../content/assets/moon.png";
 
-class Template extends React.Component {
-  render() {
-    const { location, children } = this.props
-    const rootPath = `${__PATH_PREFIX__}/`
-    let header
+const renderHeader = props => {
```
wardpeet: Bikeshed: should we create 2 components instead?
Feel free to ignore but I think this is more readable

```js
const HomepageHeader = (props) => (
  <h1
    style={{
      ...scale(0.75),
      marginBottom: 0,
      marginTop: 0
    }}
  >
    <Link
      style={{
        boxShadow: "none",
        textDecoration: "none",
        color: "var(--textTitle)"
      }}
      to={"/"}
    >
      {props.title}
    </Link>
  </h1>
);
```

```js
const Header = (props) => (
  <h3
    style={{
      fontFamily: "Montserrat, sans-serif",
      marginTop: 0,
      marginBottom: 0,
      height: 42, // because
      lineHeight: "2.625rem"
    }}
  >
    <Link
      style={{
        boxShadow: "none",
        textDecoration: "none",
        color: "rgb(102, 185, 191)"
      }}
      to={"/"}
    >
      {props.title}
    </Link>
  </h3>
);
```

```js
return location.pathname === rootPath ? <HomePageHeader {...props} /> : <Header {...props} />
```

jxnblk:
Another question I thought about while working on this is the routing/pages generated by the blog theme. With the notes theme, it defaults to `/notes` for the pages it creates. Do we want to follow a similar pattern here where the blog theme creates pages in `/blog` unless the user configures it to use the root path?

amberleyromo:
@jxnblk I think that makes a lot of sense (to have it create pages in `/blog` by default instead). I think it should be straightforward to configure to use the root path or a different path, though.

jxnblk:
Yeah, IIRC the notes theme lets you pass a `notesPath` option, where if you wanted it at the root of your site, you can set `notesPath: '/'`. That said, I feel like all our themes should use the same naming convention for this (i.e. `path` or `rootPath` instead of `notesPath`)

amberleyromo:
@jxnblk Actually, do you think both themes should default to `/` unless otherwise specified? IMO it makes sense for the default usage to be single theme, and then extensible based on the way you're composing other themes.

jxnblk:
@amberleyromo I like the idea of making themes as zero-config as possible, but I think we've been approaching themes with the idea that multiple themes can be composed together by default. I.e. a user could add multiple themes to their site without configuring them, and they work together. You could argue that the blog theme should be the main home page and the notes is a secondary page, but I'd guess a lot of people would want to change that behavior as well

amberleyromo:
@jxnblk I kind of think the opposite? Like a theme should default to just work (one theme one site), and then people can opt in consciously as they structure.

Like why `/notes`? Or why `/blog`? Some people will want `/journal`, or `/garden`, or (etc etc).

This way if someone starts a new site based on the notes theme, for notes, it just works. And then later decides to add a blog, they can _then_ think about where they want that all to live.

themes/gatsby-theme-blog/src/components/layout.js
amberleyromo: (i have seen this, @wardpeet! just leaving it alone while we're figuring some patterns out 😸)

jlengstorf:
I’d recommend we set this as a common setting (my vote would be `rootPath`) and default to `/`. I don't think composition is the default use case, and if it's a standard theme setting it's not too painful to adjust to `rootPath: '/blog'` or whatever. The happy path should be install a single theme and it Just Work™ as the home page; otherwise people will be confused if they install a theme and the home page 404s.

KyleAMathews:
Agree the rootPath should default to /. This is how I've imagined it. This would allow then adding multiple instances of a theme for multiple blogs (common in larger companies)

ChristopherBiscardi:
If we use a standardized rootPath we'll have to document the difference between that and pathPrefix

PR #14280, opened 2019-05-23
feat(gatsby): use graphiql-explorer

### LINKED ISSUE #11602: Consider integrating graphiql-explorer
[It looks](https://github.com/OneGraph/graphiql-explorer) very useful for exploring Gatsby internals.

pieh:
Replace current bare graphiql with graphiql-explorer + graphiql

Kapture 2019-05-23 at 21 50 03

Additional benefit that is not really related to graphiql-explorer, but implemented here:
new graphiql integration doesn't use external resources (currently used `express-graphql` uses https://cdn.jsdelivr.net to get react, react-dom, graphiql etc to render app correctly. This will use local bundle, so you can use it no problem on planes ;) 

TO-DO:

- [x] Write proper README
- [x] Discuss new package name (?)
- [x] Consider crafting nice default query - for the most part empty query filled with comments that will target Gatsby users (and maybe graphiql-explorer)
- [x] Add some e2e tests to to increase confidence in this (and any future changes to this)

Closes #11602


packages/gatsby-graphiql-explorer/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "gatsby-graphiql-explorer",
```
pieh: this probably need better package name - this is kind of express middleware, but not exactly, because it's not used as `express.use(someMiddleware())`, instead it is used as `graphiqlExplorer(expressApp, params)` as I need to have 2 route handlers (one for html and one for bundle)

packages/gatsby-graphiql-explorer/src/app/app.js
```diff
@@ -0,0 +1,159 @@
...
+// on visiting graphiql - in order it will be
+//  - query from query string (if set)
+//  - query stored in localStorage (which graphiql set when closing window)
+//  - default empty query
+const DEFAULT_QUERY =
+  parameters.query ||
+  (window.localStorage && window.localStorage.getItem(`graphiql:query`)) ||
+  `{
+
+}`
```
pieh: here we could add comments that will guide first time users

packages/gatsby-graphiql-explorer/src/app/app.js
```diff
@@ -0,0 +1,159 @@
...
+
+class App extends React.Component {
+  state = { schema: null, query: DEFAULT_QUERY, explorerIsOpen: true }
```
pieh: explorer is open by default - probably would be better to track state and save it to localStorage

DSchau:
@pieh I wrote a README for you, but can't push to the branch! Could you give collaborators (aka me) access?

packages/gatsby-graphiql-explorer/package.json
DSchau: It's easy to bike shed here--I don't really have a problem with the name. We could also scope it, e.g. `@gatsbyjs/graphiql` which is really what we've done here. We tweaked `graphiql` with some extra functionality.

packages/gatsby-graphiql-explorer/README.md
```diff
@@ -0,0 +1,16 @@
+# gatsby-graphiql-explorer
```
DSchau: Simple and concise 🤷‍♂ I think this will suffice!

packages/gatsby-graphiql-explorer/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "version": "0.0.1",
+  "description": "Stub description for gatsby-graphiql-explorer",
```
DSchau:
```suggestion
  "description": "GraphiQL IDE with custom features for Gatsby users",
```

packages/gatsby-graphiql-explorer/src/app/index.ejs
```diff
@@ -0,0 +1,17 @@
...
+    <title>GraphiQL</title>
+    <meta name="robots" content="noindex" />
+    <meta name="referrer" content="origin" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+  </head>
+  <body>
+    <div id="root" class="graphiql-container">Loading...</div>
+    <script
+      type="text/javascript"
+      src="/___graphql/<%= htmlWebpackPlugin.files.chunks.main.entry %>?<%= htmlWebpackPlugin.files.chunks.main.hash %>"
```
DSchau: Would this cause issues with #14234 (e.g. if navigating to /_graphql)? To avoid, we could just redirect to ___graphql, instead of exposing other endpoints--which is _probably_ a better solution anyways.

packages/gatsby-graphiql-explorer/src/app/webpack.config.js
```diff
@@ -0,0 +1,72 @@
...
+    ],
+  },
+  plugins: [
+    new HtmlWebpackPlugin({
+      template: path.resolve(__dirname, `index.ejs`),
+      filename: `index.html`,
+      inject: false,
+    }),
+    new webpack.DefinePlugin({
+      "process.env.NODE_ENV": JSON.stringify(`${mode}`),
```
DSchau: Guessing this is copypasta (so if so, ignore), but

```suggestion
      "process.env.NODE_ENV": JSON.stringify(mode),
```

packages/gatsby-graphiql-explorer/src/app/app.js
DSchau: I like the idea of exposing it by default. It's a useful feature, let's make it obvious! 💪

pieh: I should be more clear - I mean it will always start with explorer opened, so even if user close this and never use this, next time he will visit `/___graphiql` explorer will be opened again. This might be frustrating for users pretty familiar with gatsby schema that won't benefit that much from explorer

DSchau: No - I got that! I was commenting on the former part of the statement. I do think it could make sense to save a key to localStorage once closed (and opened).

packages/gatsby-graphiql-explorer/src/app/app.js
pieh: I should just copy & paste default graphiql "query" ( https://github.com/graphql/graphiql/blob/25006623dc56c78259cd1ab1401eea6b96343b42/src/components/GraphiQL.js#L1027-L1058 )

DSchau: This works with me. We could customize the query to be Gatsby specific too, e.g. a query that will typically always work, a la

```graphql
{
  site {
    siteMetadata {
      title
    }
  }
}
```

(also open to anything else here--just enough to convey a simple query that will rarely, ideally never, break)

packages/gatsby-graphiql-explorer/src/app/webpack.config.js
pieh: 🤦‍♂

packages/gatsby-graphiql-explorer/src/app/index.ejs
pieh: as long as `/___graphql` is handled by gatsby develop - this is fine regardless of which of the variants we use. Initially wanted to make this relative, but it would require trailing slash after `/___graphql/` to work

pieh: I can also adjust e2e test, to actually do something in IDE and verify that it's working (and not only check title as it works right now)

packages/gatsby-graphiql-explorer/src/app/app.js
pieh: This sounds good. 

I'll check how hard is to see if `site.siteMetadata.title` is actually in schema (we do instrospection query, so this information should be available) and will just fallback to `allSitePage.nodes.path` which will be available always (but not as useful as siteMetadata query)

pieh: Done

packages/gatsby-graphiql-explorer/src/app/app.js
pieh: done

pieh:
Will try to figure better tests e2e tests tomorrow, not quite sure why failing test is flaking in CircleCi

packages/gatsby-graphiql-explorer/src/app/index.ejs
pieh: e2e tests added

e2e-tests/development-runtime/cypress/integration/functionality/graphql-endpoint.js
```diff
@@ -1,10 +1,30 @@
 const endpoints = [`/___graphql`, `/_graphql`, `/___graphiql`]
 
+const testQueryString = `?query=%7B%0A%20%20site%20%7B%0A%20%20%20%20siteMetadata%20%7B%0A%20%20%20%20%20%20title%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D`
```
DSchau: Readability nit:

```suggestion
const testQueryString = `?query=${encodeURIComponent(`{
  site {
    siteMetadata {
      title
    }
  }
}`)}`
```

packages/gatsby-graphiql-explorer/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+A package to extend the default [GraphiQL][graphiql] IDE with useful features for Gatsby users.
+
+\_Note: accessible at http://localhost:8000/___graphql after running `gatsby develop`\_
```
DSchau:
```suggestion
*Note: accessible at http://localhost:8000/___graphql after running `gatsby develop`*
```

e2e-tests/development-runtime/cypress/integration/functionality/graphql-endpoint.js
pieh: Yeah, that's defenitely better ;)

Kevin-Hamilton:
@pieh - question, how does this impact https://www.gatsbyjs.org/docs/using-graphql-playground/ ?  Is the GraphQL Playground also updated with this feature?  If not, should that be mentioned in the Docs or Blog post?

PR #14283, opened 2019-05-23
refactor(gatsby): restrict actions available in Node APIs

stefanprobst:
Currently we pass all actions to every Node API. This PR:
* excludes internal actions from this (those not in the docs because marked `@private` in jsdoc). Specifically, this excludes `createPageDependency`, `deleteComponentsDependencies`, `replaceComponentQuery`, `replaceStaticQuery`, `queryExtracted`, `queryExtractionGraphQLError`, `queryExtractedBabelSuccess`, `queryExtractionBabelError`, `setProgramStatus`, `pageQueryRun`
* allows making certain actions only available to specific APIs. These include `createTypes` and `addThirdPartySchema`, which are currently made available to all APIs but need to be called before schema generation. This enforces the documented convention of calling schema-related actions in `sourceNodes`.
* not strictly related: removed [`maybeAddPathPrefix`](https://www.gatsbyjs.org/docs/actions/#maybeAddPathPrefix) and [`schemaFieldName`](https://www.gatsbyjs.org/docs/actions/#schemaFieldName) from the actions docs (i guess they aren't there on purpose?)


stefanprobst:
Hmm not sure why those integrations tests are failing. Is it possible they somehow keep a cached `redux/actions.js` file around which prevents accessing the new `redux/actions/index.js`?

pieh:
> Hmm not sure why those integrations tests are failing. Is it possible they somehow keep a cached `redux/actions.js` file around which prevents accessing the new `redux/actions/index.js`?

If this is not happening locally, it's quite possible some bug in gatsby-dev (i.e. we copy built files first and then install from npm which overwrites changes) - I will take a look at this

pieh:
> * not strictly related: removed [`maybeAddPathPrefix`](https://www.gatsbyjs.org/docs/actions/#maybeAddPathPrefix) and [`schemaFieldName`](https://www.gatsbyjs.org/docs/actions/#schemaFieldName) from the actions docs (i guess they aren't there on purpose?)

Also `actions` and `boundActionsCreators` are there (because those are exports from `actions.js` right now and we don't blacklist them :S) - will your changes remove them as well?

pieh:
> * allows making certain actions only available to specific APIs. These include `createTypes` and `addThirdPartySchema`, which are currently made available to all APIs but need to be called before schema generation. This enforces the documented convention of calling schema-related actions in `sourceNodes`.

Isn't this too late to do this? While using this action for creating schema has no effect, removing action for later APIs will crash builds if someone was using it. If possible I would vote for soft-removal - for example display warning and don't actually do anything, but don't crash if action is not applicable to the API (with full removal in v3)

stefanprobst:
>  will your changes remove them as well

yes those are removed as well - but makes sense, since they aren't actions themselves no?

> If possible I would vote for soft-removal

I agree we should have a deprecation warning for `onPreInit` and `onPreBootstrap`, and for the rest of the APIs show an error message instead of silently doing nothing. Question is where a good place would be to do this (other than listing all remaining APIs in `actions/restricted.js`)?
Also, for context: the original motivation was to ensure that a new `createFieldExtension` action is being called before schema generation. So the next step of my plan is to actually deprecate schema related actions in `sourceNodes` as well, and introduce a `createSchemaCustomization` Node API as the only place where schema related actions available.

pieh:
> yes those are removed as well - but makes sense, since they aren't actions themselves no?

Oh, absolutely! (I was just hoping they would be removed as nice side-effect! :P)

> Question is where a good place would be to do this (other than listing all remaining APIs in `actions/restricted.js`)?

I didn't think about implementation of this (just from high level / user / plugin perspective)

> So the next step of my plan is to actually deprecate schema related actions in `sourceNodes` as well, and introduce a `createSchemaCustomization` Node API as the only place where schema related actions available.

Just from my experience working with adding types to contentful - in that case `sourceNodes` is actually best place for defining types, as we get both data and type shapes from single API call. Restriciting it just to this single API hook would add some hoops (either doing same API call again or storing data in some global context).

If main goal of restriction is to make sure action is not called after schema was already generated I would really vote to still enable it for API hooks that are called before schema generation (and not just restrict to single API)

stefanprobst:
> in that case `sourceNodes` is actually best place for defining types, as we get both data and type shapes from single API call.

Thanks, that's valuable info!

KyleAMathews:
💯 for the change and 👍 to @pieh's suggestion to deprecatete then remove in v3.

stefanprobst:
> it's quite possible some bug in gatsby-dev

It does look like it. For testing I added `"predevelop": "rm node_modules/gatsby/dist/redux/actions.js*"` to `development_runtime` scripts and now it passes.

packages/gatsby/src/redux/actions/restricted.js
```diff
@@ -0,0 +1,276 @@
...
+  actionNames.reduce((acc, action) => {
+    acc[action] = () => {
+      report.error(
+        `\`${action}\` is not available in the \`${api}\` API. ` +
+          `Please use: ${alternativeAPI.map(a => `\`${a}\``).join(`, `)}.`
+      )
+      // FIXME: Dummy error action - required because action creators
+      // must return an action. Better would be to have a logging middleware
+      // so we could just `return log({ message, type: 'error' })`
+      return { type: `DISPATCH_ERROR` }
```
pieh: Maybe we could return empty array or adjust our redux middleware to skip `null` or `undefined` "actions"?

www/src/pages/docs/actions.js
```diff
@@ -80,7 +96,23 @@ export default ActionCreatorsDocs
 
 export const pageQuery = graphql`
   query {
-    file(relativePath: { eq: "gatsby/src/redux/actions.js" }) {
+    public: file(relativePath: { eq: "gatsby/src/redux/actions/public.js" }) {
```
pieh: Is split here beneficial? We could maybe do
```suggestion
    public: file(relativePath: { in: ["gatsby/src/redux/actions/public.js", "gatsby/src/redux/actions/restricted.js"] }) {
```
if we just display them all to keep all actions sorted. and adjust `githubPath` separately. We already have second way of providing github path (used in https://www.gatsbyjs.org/docs/node-apis/ / https://github.com/gatsbyjs/gatsby/blob/master/www/src/pages/docs/node-apis.js where almost all APIs are called in different files)

packages/gatsby/src/redux/actions/restricted.js
```diff
@@ -0,0 +1,276 @@
...
+
+import type { Plugin } from "./types"
+
+const actions = {}
+
+/**
+ * Add a third-party schema to be merged into main schema. Schema has to be a
+ * graphql-js GraphQLSchema object.
+ *
+ * Only available in [`sourceNodes`](/docs/node-apis/#sourceNodes) API.
```
pieh: This is ok, but for future maybe we could use some custom jsdoc tag to mark in what APIs action is available, so we can display it nicely in our action reference docs (so it isn't lost in description's wall of text)

packages/gatsby/src/redux/actions/restricted.js
stefanprobst: yes i think easiest is probably to use middleware we already have and just return a thunk that does not dispatch.

www/src/pages/docs/actions.js
stefanprobst: yes that makes way more sense!

packages/gatsby/src/redux/actions/restricted.js
stefanprobst: will `documentationjs` parse custom tags?

pieh: It does parse in show just full bag of all tags (even custom ones), we probably would need to create extra field in `DocumentationJS` type with something like `createResolvers` to access it nicely for .org (as this would be custom thing not applicable for all users of the plugin)

pieh: but this is not a blocker for this PR - just sidenote

packages/gatsby/src/redux/actions/restricted.js
```diff
@@ -0,0 +1,276 @@
...
+const deprecatedInAPI = {
+  ...apiWithDeprecatedSchemaActions,
+}
+
+const allowedInAPI = {
+  sourceNodes: {
+    createTypes: actions.createTypes,
+    addThirdPartySchema: actions.addThirdPartySchema,
+  },
+}
```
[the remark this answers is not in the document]
stefanprobst: Hope it's a bit more readable now

packages/gatsby/src/redux/actions/restricted.js
stefanprobst: Two questions:
* `gatsby-transformer-documentationjs` needs to preserve custom tags on the created nodes. better to preserve all, or just the one we care about (`@availableIn`)? I've opted for the latter now.
* we would not need to use `createResolvers` if we rely on inference - I've however now added createResolvers call and made the `DocumentationJs` type `@dontInfer`

stefanprobst:
Hmm looks like #14347 did not 100% solve this. Any advice how to get this through CI?

pieh:
> Hmm looks like #14347 did not 100% solve this. Any advice how to get this through CI?

After some publishing problems, this is resolved - I restarted CircleCi to use fresh gatsby-dev and it's passing now

packages/gatsby-transformer-documentationjs/src/gatsby-node.js
```diff
@@ -426,6 +426,21 @@ exports.onCreateNode = async ({ node, actions, ...helpers }) => {
...
+        )
+        if (availableIn) {
+          picked.availableIn = availableIn.description
+            .split(`\n`)[0]
+            .replace(/[[\]]/g, ``)
+            .split(`,`)
+            .map(api => api.trim())
+        }
+      }
+
```
pieh: could we just add `tags` to `picked` object, and add `availableIn` resolver (that will find custom `availableIn` tag) in just .org? We wouldn't need to add `tags` to field defs (tho it wouldn't hurt I guess)

stefanprobst: done.
did not add `tags` to the typedefs though - i think we should do a different PR that makes `DocumentationJs` `@dontInfer` and explicitly adds all fields. These are currently missing from the typedefs:

```graphql
commentNumber: Int
deprecated: DocumentationJSComponentDescription
description: DocumentationJSComponentDescription
level: Int
lineNumber: Int
optional: Boolean
tags: [DocumentationJsTags]
title: String
```

packages/gatsby/src/redux/actions/restricted.js
```diff
@@ -0,0 +1,269 @@
...
+      set(availableActionsByAPI, api, actionName, action)
+    )
+
+    const deprecatedIn = restrictions[actionName][DEPRECATED_IN]
+    deprecatedIn.forEach(api =>
+      set(
+        availableActionsByAPI,
+        api,
+        actionName,
+        withDeprecationWarning(action, api, allowedIn)
```
pieh:
```suggestion
        withDeprecationWarning(actionName, api, allowedIn)
```

`action` is function, which cause weird output (for warning) / crashes (for error):
```
warn Calling `(types, plugin, traceId) => {
  return {
    type: `CREATE_TYPES`,
    plugin,
    traceId,
    payload: types
  };
}` in the `onPreInit` API is deprecated. Please use: `sourceNodes`.
error gatsby-node.js returned an error


  TypeError: actions[action] is not a function
  
  - restricted.js:183 args
    [pr14283]/[gatsby]/dist/redux/actions/restricted.js:183:25
```

packages/gatsby/src/redux/actions/restricted.js
```diff
@@ -0,0 +1,269 @@
...
+
+    const forbiddenIn = nodeAPIs.filter(
+      api => ![...allowedIn, ...deprecatedIn].includes(api)
+    )
+    forbiddenIn.forEach(api =>
+      set(
+        availableActionsByAPI,
+        api,
+        actionName,
+        withErrorMessage(action, api, allowedIn)
```
pieh:
```suggestion
        withErrorMessage(actionName, api, allowedIn)
```
part 2 of above comment

packages/gatsby/src/redux/actions/restricted.js
stefanprobst: oops that's embarassing. sorry for that. let me actually add a test for these.

PR #14284, opened 2019-05-23
feat(gatsby-image): export TS interfaces of gatsby-image object types

kripod:
## Description

When passing optimized images through type-safe components, the need for using gatsby-image object types may emerge.

E.g.:

```tsx
import Img, { FixedObject } from 'gatsby-image';

type UserProps = {
  avatar: FixedObject;
  // ...
};

export default function User({ avatar }: UserProps) {
  return (
    <div>
      <Img fixed={avatar} />
      {/* ... */}
    </div>
  );
}
```

### DIFF
--- a/packages/gatsby-image/index.d.ts
+++ b/packages/gatsby-image/index.d.ts
@@ -1,6 +1,6 @@
 import * as React from "react"
 
-interface FixedObject {
+export interface FixedObject {
   width: number
   height: number
   src: string
@@ -11,7 +11,7 @@ interface FixedObject {
   srcSetWebp?: string
 }
 
-interface FluidObject {
+export interface FluidObject {
   aspectRatio: number
   src: string
   srcSet: string


joealden:
@kripod could you please explain to me why these need to be exported? I have been using these interfaces for a long time now without problems (when they weren't marked as exported). I can't find anything in the TypeScript docs on declaration files about this, so I wondered if you know why this is allowed?

To get a better understanding of what I mean, here is a screenshot of the named imports available to me with `gatsby-image` 2.1.1 (the version that contains this PR's commits):

screenshot

As you can see, I can import and use the `GatsbyImageProps` interface in my own code, and as the file currently exists, this interface isn't marked as exported. This is what it was like for me previously in `gatsby-image` in relation to the `FixedObject` and `FluidObject` interfaces.

The only reason I can think this would have ever worked is because maybe TypeScript treats `index.d.ts` files specially? Maybe they export everything in them by default or something? I am just guessing though, as like I said, I can't find any documentation that suggests this is the case.

PR #14286, opened 2019-05-24
Page data loading resilience

Moocar:

## Description

The [resource-loading-resilience](https://github.com/gatsbyjs/gatsby/blob/master/e2e-tests/production-runtime/cypress/integration/resource-loading-resilience.js) tests in [e2e-tests/production-runtime](https://github.com/gatsbyjs/gatsby/tree/master/e2e-tests/production-runtime) test how the gatsby browser application handles blocked resources. E.g ad blockers not allowing json files, or network connectivity problems.

These tests were all broken because they relied on the pages-manifest, which has now been removed. This PR does 2 main things.

1. It makes the new page-data.json code more resilient to errors. In the case that a component or page-data.json file is blocked, the app will now test that the page HTML file exists on the server, and if so, will force a reload of the browser. The old implementation didn't have to perform this test since it knew whether a page existed or not by looking up the pages-manifest
1. Updates the `production-runtime` tests to block `page-data` instead of `query-results`. It also uses the [Cypress task](https://docs.cypress.io/api/commands/task.html) API to simulate blocking behavior instead of having to call npm scripts each time, which radically speeds up test speed.

## Related Issues

- https://github.com/gatsbyjs/gatsby/pull/13409#issuecomment-487730702


Moocar:
Barring any more issues, this is the final required work before we PR `per-page-manifest` into `master`!! @pieh If you have a moment, I think you'd be in the best position to review this code. Especially the tests. Would love a gander from other folks too (@KyleAMathews @wardpeet etc :) )

Also, can anyone cast any light on the `.github/main.workflow` error?

pieh:
I think it's safe to ignore main.workflow here - seems like flake ( https://github.com/gatsbyjs/gatsby/pull/14286/checks?check_run_id=135248022 ). Will review today

e2e-tests/production-runtime/cypress/plugins/block-resources.js
```diff
@@ -0,0 +1,120 @@
...
+}
+
+const blockAsset = assetFileName => {
+  moveAsset(getAssetPath(assetFileName), getHiddenAssetPath(assetFileName))
+}
+
+const blockAssetsForChunk = ({ chunk, filter }) => {
+  const assetManifest = getAssetManifest()
+  assetManifest[chunk].forEach(blockAsset)
+  console.log(`Blocked assets for chunk "${chunk}"`)
```
wardpeet: I'm unsure what the difference is but I would go for cypress.log 🤷‍♂  (same for the others below)
https://docs.cypress.io/api/commands/log.html#Syntax
```suggestion
  cy.log(`Blocked assets for chunk "${chunk}"`)
```

feel free to ignore

Moocar: Nice! Didn't know about that one

packages/gatsby/cache-dir/loader.js
```diff
@@ -359,6 +363,20 @@ const queue = {
       return null
     }
   },
+
+  fetchPageHtml: rawPath => {
```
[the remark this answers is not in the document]
Moocar: `doesPageExist` is much better.

Agree on `HEAD` too. I was hoping that by doing a full GET request, the browser would be able to use that cached version when it reloaded. But html isn't cached at all so it will have to do two requests anyway.

e2e-tests/production-runtime/cypress/plugins/block-resources.js
Moocar: Ahh, except that `plugins/block-resources.js` runs in node, not the browser. So no, we can't use this.

wardpeet: Thanks for checking!

packages/gatsby/cache-dir/loader.js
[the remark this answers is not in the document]
wardpeet: Why wouldn't it be cached? It all depends on the settings of the server, right? If the server has proper cache-control headers for HTML it should cache it? (or is it browser behaviour?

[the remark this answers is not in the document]
Moocar: My understanding is that html files are normally uncached whereas other assets are. Since html files aren't content hashed. Either way, I think `HEAD` makes more sense since the main purpose is to find out whether the html file exists or not, so we know whether to show a 404.

packages/gatsby/cache-dir/loader.js
```diff
@@ -397,6 +415,7 @@ export const publicLoader = {
   loadPage: queue.loadPage,
   loadPageSync: queue.loadPageSync,
   loadPageOr404Sync: queue.loadPageOr404Sync,
```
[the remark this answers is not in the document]
Moocar: Nope, good find. Fixed.

packages/gatsby/cache-dir/ensure-resources.js
```diff
@@ -5,19 +5,55 @@ import shallowCompare from "shallow-compare"
...
+// ## Page data or component blocked:
+//
+// This can happen for e.g if an ad blocker blocks json or js, or
+// internet connection cutoff halfway through request. In this case,
+// when navigation occurs, navigate.js calls loader.loadPage, which
+// returns null. `getDerivedStateFromProps` updates props with null
+// pageResources. `shouldComponentUpdate` sees no resources, so calls
+// `retryResources`, which performs a request for the page html. If
+// the html page exists, then we have a fallback, so we reload the
+// entire page, resulting in the pure HTML being rendered.
```
KyleAMathews: This is super cool — extra security for Gatsby sites

Moocar: Just to be clear, I didn't introduce this functionality. I merely refactored and documented it 👼

PR #14289, opened 2019-05-24
feat(gatsby): enable modern builds for gatsby

wardpeet:
## Description
Gatsby generates 2 bundles. One .js & one .mjs, the .js will be used for legacy browsers (<script nomodule>) & mjs will be used for es supported modules.

Changes I made:

- [x] remove process.env from babel-preset-env, use options instead.
- [x] add modern option to babel-preset-env
- [x] enable js & mjs bundling by webpack
- [x] add module, nomodule script tags to pages (make sure we fix the ios/safari bug)
- [x] add option do disable modern builds
- [x] add e2e tests

I wonder if we should detect if a given browserlist is already > modern zo we don't have to add 2 script tags. Perhaps just disable when a custom babelrc and browserlist are given?

@DSchau created a killer todo, so for more info:
https://github.com/gatsbyjs/gatsby/issues/2114#issuecomment-467495330

## Related Issues
#2114

## Note
We lose preload for legacy browsers on first visit when a modern build is generated because there is no way for us to conditionally preload one set (legacy or modern) before knowing the browser and by then it is perhaps too late to add preload tags


KyleAMathews:
Great work! Is the idea we'll release this behind a command-line flag / experimental config value?

wardpeet:
I was thinking of writing an issue about creating a new options inside `gatsby-config.js` called `_experiments` where we could have these kinds of feature flags (Release Toggles).

packages/gatsby/cache-dir/static-entry.js
```diff
@@ -275,6 +298,8 @@ export default (pagePath, callback) => {
   scripts
     .slice(0)
     .reverse()
+    // remove legacy scripts from our preload as we preload our modern files instead
```
wardpeet: I'm unsure if there is a workaround for this that we can have our cake and eat it too. A way to do this is document.write but I rather don't want to use it.

DSchau:
@wardpeet please do! I've been thinking that would be really valuable--basically a way that we can get more fine-grained control (and likely analytics data, as well) of these experiments and make them a little more user-friendly than setting an environment variable.

packages/gatsby/cache-dir/static-entry.js
```diff
@@ -368,14 +393,44 @@ export default (pagePath, callback) => {
...
+    .map(s => {
+      const scriptPath = `${__PATH_PREFIX__}/${JSON.stringify(s.name).slice(
+        1,
+        -1
+      )}`
+      return <script key={scriptPath} src={scriptPath} type="module" async />
+    })
+
+  // Patches browsers who have a flawed noModule - module system
+  // Sadly we lose preload for legacy browsers
```
wardpeet:
```suggestion
  // Sadly we lose the benefit of the preload scanner for legacy browsers
  // because the the sources are hidden in javascript.
```

packages/gatsby/cache-dir/static-entry.js
```diff
@@ -368,14 +393,44 @@ export default (pagePath, callback) => {
...
+  // @see https://caniuse.com/#feat=es6-module
+  // 1. Safari 10.1 supports modules, but does not support the `nomodule` attribute - it will load <script nomodule> anyway.
+  // 2. Edge does not executes noModule but still fetches it
+  const noModuleBugFixScripts = `(function(b){function c(e){var d=b.createElement("script");d.src=e;b.body.appendChild(d)}"noModule"in b.createElement("script")||/Version\\/10\\.1(\\.\\d+)* Safari|Version\\/10\\.\\d(\\.\\d+)*.*Safari|Edge\\/1[6-8]\\.\\d/i.test(navigator.userAgent)||(%scripts%)})(document)`
```
wardpeet: I had no clue on how to write this in a nicer way. Basically it detects if noModule exists or it does not exists and is safari 10.1.x or safari ios 10.x as they do have know `type="module"` but not nomodule

```suggestion
  const noModuleBugFixScripts = `(function(b){function c(e){var d=b.createElement("script");d.src=e;b.body.appendChild(d)}"noModule"in b.createElement("script")||/Version\\/10\\.1(\\.\\d+)* Safari|Version\\/10\\.\\d(\\.\\d+)*.*Safari/i.test(navigator.userAgent)||(%scripts%)})(document)`
```

wardpeet:
@DSchau I think it's best to keep this under an experiment for now, I have no idea if this is going to be flawless 😄 for all websites.

sidharthachatterjee:
I think you should document 3 things on the blog post/documentation page:
- that we "handle" the Safari bug with the dynamic script tag addition
- the Chrome Core JS catastrophe
- Side note for folks to beware if they have custom plugin code that reads `webpack.stats.json` while mentioning that we handle the official ones (guess-js, postcss, netlify cms, netlify headers, offline plugin)

Once per page manifest lands, we have to update this (cc @anthony)

Random thought: In the future, a build that is set using browserslist might be "more modern" than modern so we might need to be careful about naming here 😉

wardpeet:
Some notes from our pairing session from @sidharthachatterjee 

> Create some documentation/blog post after this PR:
I think we should document 3 things on the blog post/documentation page:
- that we "handle" the Safari bug with the dynamic script tag addition
- the core-js polyfill issues with esmodules=true
- Side note for folks to beware if they have custom plugin code that reads `webpack.stats.json` or runs own webpack while mentioning that we handle the official ones (guess-js, postcss, netlify cms, netlify headers, offline plugin)

> Once per page manifest lands, we have to update this (cc @anthony)
There shouldn't be much differences but there are a few.

> Random thought: In the future, a build that is set using browserslist might be "more modern" than modern so we might need to be careful about naming here :wink: (edited) 

We'll need to see how the ecosystem moves.

moonmeister:
This  may be helpful:

https://jasonformat.com/modern-script-loading/

wardpeet:
Thanks @moonmeister! I will indeed thinker a bit about that blog post

packages/gatsby/cache-dir/static-entry.js
sidharthachatterjee: Added the note!

e2e-tests/modern-builds/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "production-runtime",
+  "description": "Gatsby default starter",
```
sidharthachatterjee:
```suggestion
  "description": "A Gatsby project to test out modern builds",
```

e2e-tests/modern-builds/package.json
```diff
@@ -0,0 +1,34 @@
...
+  "devDependencies": {
+    "cross-env": "^5.2.0",
+    "cypress": "^3.1.3",
+    "gatsby-cypress": "^0.1.7",
+    "prettier": "^1.15.2",
+    "start-server-and-test": "^1.7.11"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/gatsbyjs/gatsby-starter-default"
```
sidharthachatterjee:
```suggestion
    "url": "https://github.com/gatsbyjs/gatsby"
```

packages/gatsby/cache-dir/static-entry.js
```diff
@@ -3,19 +3,38 @@ const fs = require(`fs`)
...
 )
+const statsModern = JSON.parse(
+  fs.readFileSync(`${process.cwd()}/public/webpack.stats.modern.json`, `utf-8`)
+)
 
-const chunkMapping = JSON.parse(
+const chunkMappingLegacy = JSON.parse(
   fs.readFileSync(`${process.cwd()}/public/chunk-map.json`, `utf-8`)
 )
+const chunkMappingModern = JSON.parse(
```
sidharthachatterjee: This needs to be conditional to `ENABLE_MODERN_BUILDS`

packages/gatsby/cache-dir/static-entry.js
```diff
@@ -3,19 +3,38 @@ const fs = require(`fs`)
...
+  fs.readFileSync(`${process.cwd()}/public/chunk-map.modern.json`, `utf-8`)
```
sidharthachatterjee: This as well

ChristopherBiscardi:
Looks like this takes advantage of webpack multi-compiler yeah? Might affect gatsby-plugin-mdx's `html` field resolution. Would love a ping whenever there's a testable release.

wardpeet:
I'm going to restructure this PR in small chunks and probably remove the multi compiler instance as it's probably going to break things.

I'll close this PR for now and re-open smaller chunks. I'll keep the branch alive.

PR #14291, opened 2019-05-24
Add Kentico Cloud example 

Simply007:
## Description

AddKentico Cloud among the example


Simply007:
Could you tell me the tracking ID I should use for Google Analytics?

```
{
      resolve: 'gatsby-plugin-google-analytics',
      options: {
        trackingId: <?>
      }
}
```

KyleAMathews:
Hey!

Thanks for this PR. We don't have this written down anywhere just yet but our policy is that example sites here should only be around plugins that we also maintain as obviously it's harder to keep things up to date otherwise.

So this looks great but would be best on a repo in your org and then you can link to it from your source plugin, etc.

PR #14294, opened 2019-05-24
[WIP] Add missing dependencies

batfink:
## Description

The default starter is missing some dependencies

- https://github.com/gatsbyjs/gatsby/blob/master/starters/default/src/components/header.js#L1

### DIFF
--- a/starters/default/package.json
+++ b/starters/default/package.json
@@ -7,6 +7,7 @@
   "dependencies": {
     "gatsby": "^2.7.1",
     "gatsby-image": "^2.1.0",
+    "gatsby-link": "^2.1.1",
     "gatsby-plugin-manifest": "^2.1.1",
     "gatsby-plugin-offline": "^2.1.1",
     "gatsby-plugin-react-helmet": "^3.0.12",


pieh:
`gatsby-link` is dependency of `gatsby` and is being re-exported by `gatsby`, so you should never install `gatsby-link`.

This was one of changes in gatsby v2 ( https://www.gatsbyjs.org/docs/migrating-from-v1-to-v2/#import-link-from-gatsby )

batfink:
Ok, then the problem lies elsewhere, I’ll dig some more. Trying to get gatsby to work with pnpm.

jeffsee55:
@pieh I think this may be a transitive dependency bug. `gatsby-link` is not listed in Gatsby's [dependencies](https://github.com/gatsbyjs/gatsby/blob/master/package.json) so a strict package manager like pnpm won't install it. 

@batfink for what it's worth I was able to get gatsby working with pnpm with some hooks, your mileage may vary...

```js
// ./pnpmfile.js
module.exports = {
  hooks: {
    readPackage,
  },
}

function readPackage(pkg, context) {
  if (pkg.name === 'gatsby') {
    pkg.dependencies = {
      ...pkg.dependencies,
      'babel-plugin-styled-components': '^1.10.0',
      '@babel/runtime': '^7.4.5',
      'better-opn': '^0.1.4',
      minimatch: '3.0.4',
      'prop-types': '15.7.2',
      pify: '4.0.1',
      glob: '^7.1.4',
      '@mikaelkristiansson/domready': '1.0.9',
      '@reach/router': '1.2.1',
      'gatsby-react-router-scroll': '2.0.7',
      mitt: '1.1.3',
      'react-error-overlay': '5.1.6',
      'regenerator-runtime': '0.13.2',
      'shallow-compare': '1.2.2',
    }
    context.log('Adding missing gatsby depenedencies')
  }

  return pkg
}
```

And then a bit of a hack on the module resolution with webpack thanks to https://github.com/pnpm/pnpm/issues/991#issuecomment-446330885:

```js
let gatsbyNodeModules = require('fs').realpathSync('node_modules/gatsby')
gatsbyNodeModules = require('path').resolve(gatsbyNodeModules, '..')

exports.onCreateWebpackConfig = ({ stage, actions }) => {

  actions.setWebpackConfig({
    resolve: {
      modules: [gatsbyNodeModules, 'node_modules'],
    },
  })
  
}
```

PR #14295, opened 2019-05-24
[ gatsby-source-drupal] Losing relationships data if no ID provided.

devSnig:
In some cases, we are losing relationships data if there no id. Here some example:

I'm using Drupal 8 JSON:API and I need to get images in different sizes from Node. I'm using [Consumer Image Styles](https://www.drupal.org/project/consumer_image_styles) to achieve this, as it has done on Contenta. My changes on a field level transform my JSON to this:

`field_image: {
  data: {...}
  links: {...}
}`

so we are losing this data because we don't have `type` and `id` params.

### DIFF
--- a/packages/gatsby-source-drupal/src/gatsby-node.js
+++ b/packages/gatsby-source-drupal/src/gatsby-node.js
@@ -182,6 +182,10 @@ exports.sourceNodes = async (
           } else if (ids[v.data.id]) {
             node.relationships[`${k}___NODE`] = createNodeId(v.data.id)
           }
+          if (typeof node.relationships[`${k}___NODE`] === 'undefined') {
+            // Do not lose relationships data.
+            node.relationships[k] = v.data
+          }
         })
       }
 


pieh:
cc @gatsbyjs/drupal-working-group

wardpeet:
Hey @devSnig!

In the interest of moving this along most effectively, we strongly encourage our contributors to provide an example project which we can then use to validate the changes in this PR. This will allow us to quickly and easily ensure that this does indeed address the purported feature or fix in this PR!

An example project:

can be run locally or credentials have been provided
Feel free to message on Twitter, Discord, etc.
can be used to validate this functionality
In other words, the necessary functionality has been enabled/disabled in the project for us to easily validate these > changes>

💜

sidharthachatterjee:
@devSnig As far as I understand, this shouldn't be necessary anymore. I rebased your branch against `master` to fix conflicts (couldn't push to your fork though) and based on my observations

- Reference handling code is moved over to https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-source-drupal/src/utils.js
- In line https://github.com/gatsbyjs/gatsby/blob/15c22d3657bfd77785ba1fd69d27c36e39acfba5/packages/gatsby-source-drupal/src/utils.js#L46 `v.data.meta` is always set to `node[k]` so relationships should always be preserved now 

Please correct me in case I'm missing something or if I'm wrong! 

As a result of the above, I'm closing this PR but please feel free to reopen this if I'm missing something or if this is still necessary (as a fix for the issue you were seeing) 

Thank you so much once again for taking the time to contribute back! We look forward to more contributions from you in the future 💜

PR #14297, opened 2019-05-24
Add headless CMS integration into plugins slider.

makma:
## Description

In the plugins slider, there are various sample plugins and integrations, however, I missed there some headless CMS integration.


calcsam:
Sorry, we like this the way it is :)

PR #14302, opened 2019-05-24
fix(gatsby-plugin-sharp): Image rotation

### LINKED ISSUE #5086: resize(rotate: xx) in gatsby-transformer-sharp is not work.
## Description
GraphQL function`resize(width: 125, height: 125, rotate: 180)` is not work.

### Steps to reproduce
Visit demo page https://image-processing.gatsbyjs.org/
Compare images in **resize** part with the original [images](https://github.com/gatsbyjs/gatsby/tree/master/examples/image-processing/src/images)

I also checked that problem on my local test code.

### Expected result
In demo page, images in **resize** part should be rotated.

### Actual result
In demo page, images in **resize** part is not rotated.

### Environment
Please check here https://github.com/gatsbyjs/gatsby/blob/master/examples/image-processing/package.json

cwgw:
## Description

Rotating images with gatsby-plugin-sharp via the `rotate` argument is currently broken.

The plugin uses sharp's rotate method without args to lock-in orientation from EXIF. Unfortunately, this makes additional calls to rotate ineffectual.

This PR adds a check for the rotate arg and skips the early rotate call if the arg is provided and not equal to zero.

## Related Issues

Fixes #5086


packages/gatsby-plugin-sharp/src/index.js
```diff
@@ -144,7 +144,11 @@ async function generateBase64({ file, args, reporter }) {
   })
   let pipeline
   try {
-    pipeline = sharp(file.absolutePath).rotate()
+    pipeline = sharp(file.absolutePath)
+
+    if (!options.rotate || options.rotate === 0) {
```
wardpeet: 0 is a falsy value so it should be covered by `!options.rotate`
```suggestion
    if (!options.rotate) {
```

packages/gatsby-plugin-sharp/src/process-file.js
```diff
@@ -91,6 +89,10 @@ exports.processFile = (file, transforms, options = {}) => {
 
     let clonedPipeline = transforms.length > 1 ? pipeline.clone() : pipeline
 
+    if (!args.rotate || args.rotate === 0) {
```
wardpeet: same as above
```suggestion
    if (!args.rotate) {
```

packages/gatsby-plugin-sharp/src/trace-svg.js
```diff
@@ -16,7 +16,11 @@ exports.notMemoizedPrepareTraceSVGInputFile = async ({
 }) => {
   let pipeline
   try {
-    pipeline = sharp(file.absolutePath).rotate()
+    pipeline = sharp(file.absolutePath)
+
+    if (!options.rotate || options.rotate === 0) {
```
wardpeet: same as above
```suggestion
    if (!options.rotate) {
```

cwgw:
Fair enough. Thank you.

I'm not sure why that gatsby-image test is failing. Is this fix breaking something else or is that unrelated?

PR #14307, opened 2019-05-24
chore: update Prettier, ESLint, SVGO and plugins

lipis:
## Description

Updated to the latest versions of Prettier, ESLint, SVGO, Plugins, Husky, Lint-staged


KyleAMathews:
What breaking changes does husky have?

lipis:
@KyleAMathews https://github.com/typicode/husky/blob/master/CHANGELOG.md#200
- Breaking change drop Node 6 support

lipis:
I reverted husky since we support node 6.

wardpeet:
we only support node 6 for dependencies. We don't need to support it for dev stuff :) So feel free to upgrade to the lasest version as long as gatsby itself is node 6 compliant we are good!

lipis:
Updated.. but not sure why it's complaining.. synced with master as well.

Issue #14310, opened 2019-05-24
[gatsby-plugin-facebook-analytics] Add version, cookie and xfbml config options


keithcrooks:
## Summary

The Facebook Analytics plugin has the SDK version hard coded as `v2.12`. This should be a configurable option so that the user can choose their own preferred version without the plugin having to be updated.

### Basic example

Previous plugin config:
```javascript
plugins: [
  {
    resolve: `gatsby-plugin-facebook-analytics`,
    options: {
      appId: `YOUR_APP_ID`,
      includeInDevelopment: false,
      debug: false,
      language: `en_US`,
    },
  },
]
```

Proposed config:

```javascript
plugins: [
  {
    resolve: `gatsby-plugin-facebook-analytics`,
    options: {
      appId: `YOUR_APP_ID`,
      version: `v2.12`,
      xfbml: true,
      cookie: false,
      includeInDevelopment: false,
      debug: false,
      language: `en_US`,
    },
  },
]
```

### Motivation

The plugin is currently hard-coded to v2.12. The latest version of the Facebook JavaScript SDK is v3.3. If this change it made, it will future-proof the plugin as new versions of the SDK can be specified by providing the required version in the plugin options.

v2.12 should be the default value so as not to break backwards compatibility.

The xfbml value is also hard-coded to true. Unless the user is also using the SDK for Social Plugins, this value can be set to false to speed up page loads.

It would be a nice-to-have to also include an option for the cookie init parameter.

[Facebook JavaScript SDK initialization reference](https://developers.facebook.com/docs/javascript/reference/FB.init/v3.3)

DSchau:
Agreed! This would be a great first contribution for somebody.

All plugin APIs have their second argument defined as pluginOptions (options passed to plugins!) so we could add this as an option and default it to 3.3.

I would specifically propose:

1. Let's take this opportunity to update to 3.3 (if it requires a major bump--that's OK!)
    - To the implementor, we're only using [`FB.AppEvents.logPageView`](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-facebook-analytics/src/gatsby-browser.js#L4) which _probably_ isn't a breaking change between versions
[...]

PR #14311, opened 2019-05-24
feat(gatsby-plugin-facebook-analytics): Add version, cookie and xfbml config options

keithcrooks:
## Description

* Adds version, cookie and xfbml config options. Default values ensures backwards compatibility is maintained.
* Fixes tests as snapshots were incorrect which lead to false positive results.
* Updates the README documentation to include further information about using the plugin.

## Related Issues

Implements #14310

### DIFF
--- a/packages/gatsby-plugin-facebook-analytics/src/gatsby-ssr.js
+++ b/packages/gatsby-plugin-facebook-analytics/src/gatsby-ssr.js
@@ -9,32 +9,32 @@ export const onRenderBody = ({ setPostBodyComponents }, pluginOptions) => {
     language = `en_US`,
   } = pluginOptions
 
+  const sdkFile = debug === true ? `sdk/debug.js` : `sdk.js`
+
   if (process.env.NODE_ENV === `production` || includeInDevelopment || debug) {
     setPostBodyComponents([
       <script
         key="plugin-facebook-analitycs"
         dangerouslySetInnerHTML={{
-          __html: stripIndent`
-          window.fbAsyncInit = function() {
-            FB.init({
-              appId      : ${appId},
-              xfbml      : true,
-              version    : 'v2.12'
-            });
+          __html: stripIndent(`
+            window.fbAsyncInit = function() {
+              FB.init({
+                appId      : ${appId},
+                xfbml      : true,
+                version    : 'v2.12'
+              });
 
-            FB.AppEvents.logPageView();
+              FB.AppEvents.logPageView();
+            };
 
-          };
-
-          (function(d, s, id){
-             var js, fjs = d.getElementsByTagName(s)[0];
-             if (d.getElementById(id)) {return;}
-             js = d.createElement(s); js.id = id;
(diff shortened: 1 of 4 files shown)


packages/gatsby-plugin-facebook-analytics/src/gatsby-ssr.js
```diff
@@ -4,37 +4,41 @@ import { stripIndent } from "common-tags"
 export const onRenderBody = ({ setPostBodyComponents }, pluginOptions) => {
   const {
     appId,
-    includeInDevelopment = false,
+    cookie = false,
     debug = false,
+    includeInDevelopment = false,
     language = `en_US`,
+    version = `v2.12`,
```
DSchau: Are there any breaking change from 2 - 3? Otherwise, is there a reason we wouldn't just use the latest and greatest?

It's also just fine to make a breaking change in the plugin itself to accommodate using 3.3.

DSchau: It is kinda nice to keep it _for sure_ not a breaking change though--so I appreciate the effort here to enable that!

keithcrooks: I'm new to Facebook Analytics myself. However, looking at the Facebook JavaScript SDK reference documents and the plugin code, I think you are correct and that it should be okay to bump the default to the latest version 👍

keithcrooks: I've implemented the change to set the default version at v3.3 👍

DSchau: Do you have a Facebook Analytics account you can test the changes with? Doesn't seem like we'd have broken anything here--but worth validating!

Otherwise one of us can validate manually early next week!

DSchau: I think we're probably good -> https://developers.facebook.com/docs/graph-api/changelog/breaking-changes/

keithcrooks: @DSchau Yes, I followed the [How to Make a Reproducible Test Case](https://www.gatsbyjs.org/contributing/how-to-make-a-reproducible-test-case/) to setup a test site and then used the gatsby-dev-cli to pull in the changes I had made and set the configuration to point to a test Facebook App and have seen some data come through.
Screen Shot 2019-05-24 at 23 04 08

DSchau: Nicely done!! Let's get this merged, then. Thanks so much!

keithcrooks: It was a pleasure to contribute and thank you for the feedback and reviewing so quickly 😄

DSchau:
Published as `gatsby-plugin-facebook-analytics`@2.1.0 (minor version bump!).

PR #14314, opened 2019-05-25
feat(gatsby-remark-images): enable relative images in jsx content

ChristopherBiscardi:
Enables image processing for MDX with JSX tags for content such as the example
below.

```html
<figure>
  <img alt="The modularization of the CMS" height="400" src="./modular-cms-architecture.png" />
  <figcaption>
   The modularization of the CMS
  </figcaption>
</figure>
```

Issue can be reproduced without this change by looking at `/` (index.mdx) in this repro repo: https://github.com/ChristopherBiscardi/gatsby-mdx-remark-images-jsx-tags-repro. Note that the markdown syntax works and the JSX does not before this change, and after this change the JSX syntax works as well.


pieh:
Published `gatsby-remark-images@3.0.13`

PR #14259, opened 2019-05-23
refactor(www): Color tokens

fk:
This refactors `www` to use a … decent color palette with 10 shades of purple, teal, blue, red, orange, yellow, magenta, green, grey (formerly `gray`), and (for now) simple black and white. 🌮 

It also introduces a very basic set (and def. to be discussed, improved, and expanded) of role-based color tokens, which should make theme creation (e.g. inverted/dark mode) a lot easier. 🍕 

---

- Remove most of the named color shortcuts until we figure out a good naming convention (or not ;-)) — „most“ to keep noise as low as possible:
  - `gatsbyDark` -> `purple[800]`
  - `gatsbyDarker` — not needed anymore
  - `lavender` -> `purple[400]`
  - `wisteria` -> `lavender` -> `purple[200]`
  - `accentLight` -> `orange[200]`
  - `accentDark` -> `orange[900]`
  - `skyLight` -> `teal[100]`
  - `skyDark` -> `blue[700]`
  - `mint` -> `teal[400]`
  - `success` -> `green[500]`
  - `lemon` -> `yellow[400]`
- Remove outdated comments in `src/utils/tokens/colors`.
- Consolidate `className`-basedCSS styles for SVG icons from `src/components/navigation-mobile` and `src/components/layout/page-heading`.
- Merge former `colors.gray` with new `palette.grey`, and move most of the former `colors.gray` to **role-based tokens** (defining their roles in the UI), laying the foundation for color themes:
  - `colors.gray.superLight` and `colors.gray.border` -> `colors.gray.border.subtle`
  - `colors.gray.lightCopy` and `colors.gray.bright` -> `colors.gray.calm` -> `colors.text.secondary`
  - `colors.gray.dark` -> `colors.text.header` — plus remove it from headlines where it’s redundant
  - `colors.gray.whisper` -> `colors.ui.background`
  - `colors.gray.light`, `colors.ui.whisper` -> various (`colors.lilac`, `colors.ui.border.form`)
  - `colors.ui.bright` -> `colors.purple[200]`

Along the way:
- Remove Webkit scrollbar styles for `/ecosystem/`.
- Remove a border-radius in `src/components/rotator` that shouldn’t have been there in the first place.
- Fix quite a few border colors (`ui.light` -> `ui.border.subtle` mostly).
- Fix (as in „comply to WCAG 2.0 AA“) color contrast of
  - `/plugins/` NPM download count (also simplify markup)
  - `/showcase/` and `/starters/` filter count
[...]

PR #14328, opened 2019-05-26
fix(www): Sidebar z-index; add „zIndices“ tokens

fk:
- fix sidebar navigation on mobile being covered by the mobile navigation, a regression I introduced in  #14259 while trying to fix the feedback widget on smaller screens covering the mobile navigation:
image (note that the last link in the sidebar, "Gatsby REPL", is covered by the mobile navigation)
- add `tokens/z-indices` (as per https://system-ui.com/theme)

### DIFF
--- a/www/src/components/banner.js
+++ b/www/src/components/banner.js
@@ -2,7 +2,7 @@ import React from "react"
 import styled from "@emotion/styled"
 import { OutboundLink } from "gatsby-plugin-google-analytics"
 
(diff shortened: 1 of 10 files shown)


fk:
Moar https://system-ui.com/theme FTW @jxnblk ;)

fk:
https://www.youtube.com/watch?v=OJd22xoSTNM ✌️

Issue #13945, opened 2019-05-09
[docs] [workflows] Installing and using Contentful


jasonellington:
Part of the **Top 25 Learning Workflows initiative**. See #13708 for the meta issue that this issue falls under.

# User story

As a new Gatsby user, I want to create a blog/site that pulls content and images from Contentful and allows me to query it.

# Evaluation

| Search | Discover | Complete | Linked | Tone | Style | Overall |
| --- | --- | --- | --- | --- | --- | --- |
| 😄 | 😐 | 😐 | 😞 | 😐 | 😄  | 😐 |

# Steps taken to implement

1. Searchability
   1. Searched `gatsby contentful` on Google. clicked [second link](https://www.gatsbyjs.org/packages/gatsby-source-contentful/).
   1. First link was to Contentful site
1. Discoverability
   1. Searchbar:
      1. Searched `gatsby contentful` in search bar; only blog articles show up.
      1. Searched`contentful`; [Sourcing from Contentful](https://www.gatsbyjs.org/docs/sourcing-from-contentful/) pops up.
      **[rec]** searching for `gatsby contentful` should bring up [Sourcing from Contentful](https://www.gatsbyjs.org/docs/sourcing-from-contentful/) as well
      1. Some blogs do not mention [Sourcing from Contentful](https://www.gatsbyjs.org/docs/sourcing-from-contentful/)
   1. Clickpath on .org:
      1. Clicked _docs_
      1. Clicked _Guides > Sourcing Content and Data > Sourcing from (Headless) CMSs > Sourcing from Contentful_
      1. Discoverable
1. Completeness
   1. Followed steps in [Sourcing from Contentful](https://www.gatsbyjs.org/docs/sourcing-from-contentful/)
   1. Clicked on Link [Quick Start guide](https://www.gatsbyjs.org/docs/) which actually links to `Gatsby.js Documentation`. 
  **[rec]** actually link to [Quick Start](https://www.gatsbyjs.org/docs/quick-start), the link above goes to docs.
   1. Followed steps in [Quick Start](https://www.gatsbyjs.org/docs/quick-start).
   1. Jumped back to [Sourcing from Contentful](https://www.gatsbyjs.org/docs/sourcing-from-contentful/)
      1. Followed `prerequisites`
[...]

PR #14339, opened 2019-05-27
updated quick link in docs from #13945

abhinavprasad47:
fixed one of the broken links i don't know if it's correct but it works


LekoArts:
This was already fixed in https://github.com/gatsbyjs/gatsby/pull/13962

PR #14347, opened 2019-05-27
fix(gatsby-dev-cli): refactor, wait for package installation before copying files, handle packages that are not direct deps of local project

pieh:
There's lot of things here:
 - some bug fixes - like waiting for package installation to finish before copying files directly
 - refactoring - move some utilities to separate files (easier to read code and also to test/mock)
 - finally, proper handling of packages that are not direct dependencies of current project (i.e. `gatsby-dev --packages gatsby-cli` will correctly watch `gatsby-cli`, but install `gatsby`)
 - also, clean up .js/.js.map files after installing packages from node_modules for packages that will copy files over to get rid of any stale files (this solves issue @stefanprobst hit in https://github.com/gatsbyjs/gatsby/pull/14283#issuecomment-495398172 )

Hopefully added test coverage add some clarity and confidence

/cc @wardpeet


packages/gatsby-dev-cli/src/__tests__/watch.js
```diff
@@ -142,11 +172,487 @@ describe(`watching`, () => {
...
+    installPackages.mockClear()
+    checkDepsChanges.mockClear()
+    promisifiedSpawn.mockClear()
+  })
+
+  afterAll(() => {
+    global.process = realProcess
+  })
+
+  describe(`publishing and installing packages`, () => {
```
wardpeet: should we also check if gatsby copy is working?

pieh: Like checking if `fs.copy` was called from monorepo to local project? If so I think we can do that in `installs from npm before copying files` test (last test from this file) - it also sounds like good idea

wardpeet: yes check if copy has been for the correct packages.

so for example we should test
```
checkDepsChanges.mockImplementationOnce(mockDepsChanges([`gatsby`]))
watch(process.cwd(), [`gatsby`, `gatsby-plugin-sharp`], {
  scanOnce: true,
  quiet: true,
  monoRepoPackages,
  localPackages: [`gatsby`, `gatsby-plugin-sharp`],
})
```

now it should install gastby and copy gatsby & gatsby-plugin-sharp

pieh: gotcha! makes total sense, will add

pieh: added https://github.com/gatsbyjs/gatsby/pull/14347/commits/36542cfa0b20b719e58f99ca793f0545985b2ced - let me know if this is what you had in mind

m-allanson:
@stefanprobst this wasn't published earlier, now available in `gatsby-dev-cli@2.4.16`

Issue #14205, opened 2019-05-21
Ignore merge commits in the PRs bot


m-allanson:
## Summary

The PR bot at https://github.com/gatsbyjs/gatsby/tree/master/.github/actions/high-priority-prs is great for highlighting PRs that need attention. However sometimes there are false positives.

In the `PRs with new commits awaiting review` list, any pr that has merge commits will be incorrectly highlighted. Merge commits should be ignored by the bot. This could be done by ignoring commits with messages like:

- ignore `merge branch 'master'`
- ignore `merge remote-tracking branch`

### Motivation

False positives are annoying and make the bot less useful, let's get rid of them.

sharat9211:
Hi, 
If nobody else is working, I am interested in taking up this issue.

pieh:
Hey @sharat9211! It seems like @mkan0141 already opened PR that tackles this issue

pieh:
#14352 was merged

PR #14352, opened 2019-05-27
feat(high-priority-prs): ignore master merge commits

mkan0141:
## Description

Ignoring commits with messages like:

- ignore ```merge branch 'master'```
- ignore ```merge remote-tracking branch```
## Related Issues

#14205

### DIFF
--- a/.github/actions/high-priority-prs/process.js
+++ b/.github/actions/high-priority-prs/process.js
@@ -83,6 +83,11 @@ const maintainers = {
   },
 }
 
+const ignoreMessages = [
+  "Merge branch 'master'",
+  "Merge remote-tracking branch"
+]
+
 module.exports = data => {
   const prs = data.repository.pullRequests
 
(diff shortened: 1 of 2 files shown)


.github/actions/high-priority-prs/process.js
```diff
@@ -151,11 +156,14 @@ module.exports = data => {
       _.maxBy(pr.comments.nodes, n => n.createdAt),
       `createdAt`
     )
+
+    const lastCommitMessage = pr.commits.nodes[0].commit.message
     const lastCommit = pr.commits.nodes[0].commit.authoredDate
```
pieh: Here we check only last commit and this can cause some false negatives in situation like this:

```
 - last comment
 - meaningful commit
 - merge commit
```
I think we need to iterate commit list until we find non-merge commit and use date of that commit

mkan0141: That's true. 
I will fix it soon.
(sorry... My english is not very good, but I'll try my best! )

pieh: Your english is great! And thanks for addressing my comment!

PR #14355, opened 2019-05-27
chore(showcase): Update sites.yml

ooloth:
Add two new sites and credit designer on my previously listed sites where applicable.


ooloth:
@KyleAMathews Haha. I appear to!

Thanks so much for the compliment! And, obviously, for Gatsby. 🙌

PR #14356, opened 2019-05-27
fix(gatsby-source-mongodb): fix null values

elmarburke:
## Description

when `preserveObjectIds: true` is enabled this function would throw on line 12 due to a `Object.key` if a value is `null`.

## Related Issues

no issue yet, why open an issue if you could fix it directly 😉


pieh:
`gatsby-source-mongodb@2.0.19` was published

PR #14358, opened 2019-05-27
feat(www): Add brand guidelines, design tokens docs (humble beginnings)

fk:
A humble first shot at documenting our brand guidelines, providing a place for people to find our logo, and document the design tokens used to create our OSS website UI.

Prior work shipped to mutability.netlify.com used https://styled-system.com/ under the hood. I really enjoyed working with it, but using it for .org `www` of course should be thoroughly discussed. I’m including it here to facilitate a discussion, …and to ship things! ;-)

- consolidate prior work with existing .org `www` design tokens
  - add a breakpoint `xxs`, required to make styled-system’s https://styled-system.com/responsive-styles work
  - add rudimentary `colors.blackFade` and `colors.whiteFade`
  - extend `tokens/font-sizes`, now generated via a formula that outputs (device) pixel values; generated values are manually adjusted, then converted to `rem` in `utils/presets`
- update link to mutability.netlify.com on `/contributing/organize-a-gatsby-event` to point to `/guidelines/logo/`
- add an item „Logo & Assets“ to the global footer, linking to `/guidelines/logo/`
- add `presets.fontWeights`
- add `presets.borders`

The original PoC implementation also makes use of `react-spring` and `react-visibility-sensor` to add some „text highlighting“ animations, and uses `react-copy-to-clipboard` and the `color` package to facilitate UI behavior and various calculations for the „Color“ part of the guidelines.

I’m not opposed to remove the animations — they don’t add much — but might need some help in dropping the `color` lib if we want to get that dep out of the bundle. Noob alert! ;-)

Oh, `react-copy-to-clipboard` probably should go, too. Unfortunately we still didn’t settle on anything reusable.

—

Wether we go for `styled-system` or not, we should think about using `emotion`’s context/`emotion-theming` moving forward. No need to explicitly import stuff esp. now that refactoring things have settled down.
I recall @jxnblk saying there are treeshaking/bundle size benefits to not doing so, though.

—

Todo:

- fix/expand „Color“ data, content and UI
  - fix missing color swatches for black and white on the „Logo“ page
- add `letterSpacing`, `lineHeights`, `fontWeights` token documentation to „Typography“, tidy examples here and there
- add partner logo lockup guide to „Logo“
- give the copy a very close second look
- lots more in and out of scope for this PR :-P — will follow up on that


fk:
Ugh, where is that button to change that back into a draft PR :-/ … closing! 😅

PR #14361, opened 2019-05-27
fix(gatsby-source-contentful): Don't set default widths

KyleAMathews:
This isn't the expected behavior and doesn't track what
gatsby-transformer-sharp does.

Most people do set widths so having the default doesn't affect them
but if you want to constraint the height and not the width, this is
currently impossible as e.g. you set `height: 200` and `width: 400` will
still get set.


pieh:
`gatsby-source-contentful@2.0.65` was published

Issue #13865, opened 2019-05-05
Allow unicode file path for gatsby-config


Ta2Rim:
## Description
![image](https://user-images.githubusercontent.com/41318449/57191994-95acfc00-6f66-11e9-9a11-ef6c9c6fb568.png)

when I run `gatsby devlop` causes error like above.

I've named my PC named '신태림' is my Korean name)
and I'd like to use gatsby on desktop directory.
but Windows make desktop directory located in the directory named PC name. (in this case, \신태림 exists on path)

I think it has to be solved for non-english speaking country

### Steps to reproduce

locate gatsby directory in some directory named with Unicode 
copy this --> 신태림

https://github.com/Ta2Rim/taesdevblog

### Expected result

Compiled successfully
( When I located my gatsby project in C drive, it works )

### Actual result
![image](https://user-images.githubusercontent.com/41318449/57191994-95acfc00-6f66-11e9-9a11-ef6c9c6fb568.png)

![image](https://user-images.githubusercontent.com/41318449/57192144-27693900-6f68-11e9-8079-c19b09056f18.png)

error occurs

### Environment

Run `gatsby info --clipboard` in your project directory and paste the output here.

  System:
    OS: Windows 10
    CPU: (4) x64 Intel(R) Core(TM) i5-7200U CPU @ 2.50GHz
  Binaries:
    Yarn: 1.13.0 - C:\Users\신태림\AppData\Roaming\npm\yarn.CMD
    npm: 6.4.1 - C:\Program Files\nodejs\npm.CMD
  Languages:
    Python: 2.7.14
  Browsers:
    Edge: 42.17134.1.0
  npmPackages:
    gatsby: ^2.3.34 => 2.4.2
    gatsby-image: ^2.0.41 => 2.0.41
    gatsby-plugin-manifest: ^2.0.29 => 2.1.1
    gatsby-plugin-offline: ^2.0.25 => 2.1.0
    gatsby-plugin-react-helmet: ^3.0.12 => 3.0.12
    gatsby-plugin-sharp: ^2.0.35 => 2.0.36
    gatsby-source-filesystem: ^2.0.33 => 2.0.33
    gatsby-transformer-remark: ^2.3.12 => 2.3.12
    gatsby-transformer-sharp: ^2.1.18 => 2.1.19

LekoArts:
Seems related to: https://github.com/gatsbyjs/gatsby/pull/13882

Ta2Rim:
related to #14126

PR #14372, opened 2019-05-28
remove slash

woodgear:
## Description
remove slash and implement it under utils
## Related Issues
#13865

### DIFF
--- a/packages/gatsby-source-filesystem/src/__tests__/utils.js
+++ b/packages/gatsby-source-filesystem/src/__tests__/utils.js
@@ -34,4 +34,8 @@ describe(`slash path`, () => {
       expect(slash(path)).toBe(expectRes)
     })
   })
+  it(`does not modify extended length paths`, () => {
+    const extended = `\\\\?\\some\\path`
+    expect(slash(extended)).toBe(extended)
+  })
 })
--- a/packages/gatsby-source-filesystem/src/__tests__/utils.js
+++ b/packages/gatsby-source-filesystem/src/__tests__/utils.js
@@ -1,4 +1,4 @@
-const { getRemoteFileExtension, getRemoteFileName } = require(`../utils`)
+const { getRemoteFileExtension, getRemoteFileName, slash } = require(`../utils`)
 
 describe(`create remote file node`, () => {
   it(`can correctly retrieve file name and extensions`, () => {
@@ -20,5 +20,15 @@ describe(`create remote file node`, () => {
       expect(getRemoteFileName(url)).toBe(name)
       expect(getRemoteFileExtension(url)).toBe(ext)
     })
+  })
+})
+
+describe(`slash path`, () => {
+  it(`can correctly slash path`, () => {
+    ;[[`foo\\bar`, `foo/bar`], [`foo\\中文`, `foo/中文`]].forEach(
+      ([path, expectRes]) => {
+        expect(slash(path)).toBe(expectRes)
+      }
+    )
   })
 })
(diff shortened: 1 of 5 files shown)


packages/gatsby-source-filesystem/src/__tests__/utils.js
```diff
@@ -22,3 +22,13 @@ describe(`create remote file node`, () => {
...
+
+describe(`slash path`, () => {
+  it(`can correctly slash path`, () => {
+    ;[[`foo\\bar`, `foo/bar`], [`foo\\中文`, `foo/中文`]].forEach(
+      ([path, expectRes]) => {
+        expect(slash(path)).toBe(expectRes)
+      }
+    )
+  })
+})
```
superhawk610: please add a second test asserting that extended length paths are not modified

packages/gatsby-source-filesystem/src/__tests__/utils.js
```diff
@@ -22,3 +22,13 @@ describe(`create remote file node`, () => {
     })
   })
 })
+
+describe(`slash path`, () => {
+  it(`can correctly slash path`, () => {
+    ;[[`foo\\bar`, `foo/bar`], [`foo\\中文`, `foo/中文`]].forEach(
```
superhawk610: remove unnecessary semicolon

```suggestion
    [[`foo\\bar`, `foo/bar`], [`foo\\中文`, `foo/中文`]].forEach(
```

packages/gatsby-source-filesystem/src/__tests__/utils.js
woodgear: thanks for review,I have already updated.

packages/gatsby-source-filesystem/src/utils.js
```diff
@@ -39,3 +39,22 @@ export function getRemoteFileExtension(url) {
...
+ * slash
+ * --
+ * Convert Windows backslash paths to slash paths: foo\\bar ➔ foo/bar
+ *
+ *
+ * @param  {String}          path
+ * @return {String}          slashed path
+ */
+export function slash(path) {
+  const isExtendedLengthPath = /^\\\\\?\\/.test(path)
```
superhawk610: Windows supports a special syntax for pathnames longer than 255 characters (you can read more [here](https://docs.microsoft.com/en-us/windows/desktop/fileio/naming-a-file#maximum-path-length-limitation)).

`slash` should not modify pathnames in this special syntax because they are only valid on Windows machines and thus must use forward slashes. You should add a test asserting that this is the case, like

```
it(`does not modify extended length paths`, () => {
    const extended = `\\\\?\\some\\path`
    expect(slash(extended)).toBe(extended)
})
```

woodgear: oh,that what you mean...

woodgear: I have already updated.

superhawk610:
@LekoArts this should be good to go, any idea why the CI is failing?

freiksenet:
Thanks a lot!

yiluomyt:
Here, I also found a similar problem using `gatsby-remark-images`, which is also caused by the dependency  `slash`. It will skip the images which contains unicode in file name.

https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-remark-images/src/index.js#L112-L124

packages/gatsby-source-filesystem/src/__tests__/utils.js
Athang663: Athang663

packages/gatsby-source-filesystem/src/create-file-node.js
Athang663, line 1: Athang663

Athang663: packages/gatsby-source-filesystem/src/create-file-node.js

Athang663 (commented):
packages/gatsby-source-filesystem/src/create-file-node.js

PR #14377, opened 2019-05-28
feat(themes): Use theme-ui for styles in blog theme

jxnblk:
## Description

This attempts to replace the use of global CSS with styles provided by theme-ui and adds a theme object for design tokens. Some of the styles have slightly changed as I attempted to normalize values in certain places, but I've tried to keep it as true to the original as possible. Please let me know if anything looks off design-wise. In the example site, since both `gatsby-theme-blog` and `gatsby-theme-notes` are installed, you can see how the theme object is currently merged and used in pages from both themes.

- Replaces custom ThemeProvider with the theme-ui provider
- Replaces the dark mode implementation with the one from theme-ui
- Replaces Typography.js plugin with `theme-ui-typography`, which converts Typography.js themes to the theme-ui object shape. This also replaces calls to the `rhythm` utility with values defined in `theme.space`
- Replaces inline style attributes with Emotion's `css` prop

## Related Issues

Related to #14276


themes/gatsby-theme-blog/src/components/layout.js
```diff
@@ -1,130 +1,124 @@
...
+    <Styled.root>
+      <Layout>
+        <Global
+          styles={theme =>
+            css({
+              body: {
+                color: `text`,
+                bg: `background`,
+              },
+            })(theme)
```
jlengstorf: Where is `theme` coming from? Is this a `theme-ui` special thing?

themes/gatsby-theme-blog/src/pages/404.js
```diff
@@ -12,8 +13,10 @@ class NotFoundPage extends React.Component {
     return (
       <Layout location={this.props.location} title={siteTitle}>
         <SEO title="404: Not Found" />
-        <h1>Not Found</h1>
-        <p>You just hit a route that doesn&#39;t exist... the sadness.</p>
+        <Styled.h1>Not Found</Styled.h1>
+        <Styled.p>
+          You just hit a route that doesn&#39;t exist... the sadness.
```
jlengstorf: This is something I keep forgetting to do, which is to update the 404 copy to be more helpful. Happy to skip this for another PR, but figured if you don’t mind we can just tack it on here.
```suggestion
          The page you’re trying to visit has moved or doesn’t exist. Please check the URL or use the site navigation to find the content you’re looking for.
```

themes/gatsby-theme-blog/src/theme/prism.js
```diff
@@ -0,0 +1,48 @@
+// todo: swap for theme colors
+export default {
+  ".attr-name": {
+    fontStyle: `italic`,
+  },
+  ".comment": {
+    color: `prism.comment`,
+  },
+  [[`.attr-name`, `.string`, `.url`]]: {
```
jlengstorf: Do you know why this requires a nested array?

themes/gatsby-theme-blog/src/theme/prism.js
```diff
@@ -0,0 +1,48 @@
+// todo: swap for theme colors
```
jlengstorf: Is this still a TODO?

themes/gatsby-theme-blog/src/theme/styles.js
```diff
@@ -0,0 +1,73 @@
...
+  root: {
+    fontFamily: `body`,
+  },
+  Container: {
+    maxWidth: 672,
+    px: 3,
+  },
+  pre: {
+    // references styles from theme.prism (src/theme/prism.js)
+    variant: `prism`,
```
jlengstorf: Is this kind of like using a parent selector for all the Prism styles? Something like `pre .tag`?

themes/gatsby-theme-blog/src/theme/typography.js
```diff
@@ -0,0 +1,4 @@
+import { toTheme } from "theme-ui-typography"
+import wordpress2016 from "typography-theme-wordpress-2016"
```
jlengstorf: It seems a little confusing to import the fonts in a different place from the typography theme. Would it be possible/make sense to include all of the typography stuff here instead?

themes/gatsby-theme-blog/gatsby-ssr.js
```diff
@@ -0,0 +1,6 @@
+import React from "react"
+import ThemeProvider from "./src/theme/provider"
```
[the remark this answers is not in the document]
jxnblk: It *looks* like they should be imported here as well

themes/gatsby-theme-blog/src/components/footer.js
```diff
@@ -1,32 +1,31 @@
...
-          marginTop: rhythm(2.5),
-          paddingTop: rhythm(1),
-        }}
+        css={css({
+          mt: 4,
+          pt: 3,
+        })}
       >
-        <a
+        <Styled.a
```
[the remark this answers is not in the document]
jxnblk: Really rough docs for `Styled` are here: https://theme-ui.now.sh/#styled-components – essentially it's a way to use styles from `theme.styles` (which are used in MDX) anywhere outside of an MDX file

themes/gatsby-theme-blog/src/components/layout.js
jxnblk: This is part of Emotion's API. The `css` prop and the Global components' `styles` prop accept a function as a value, passing Emotion's `theme` context as an argument. The Theme UI `css` function uses the `theme` object so that the color defined here is `theme.colors.text`

themes/gatsby-theme-blog/src/theme/prism.js
jxnblk: This is sorta weird syntax and isn't entirely necessary; the outer brackets are for the object key and the inner brackets are an array. Emotion will convert this to something like `.hashed-classname .attr-name, .hashed-classname .string` – it might be more readable to use a single string key here instead, which I think Emotion handles the same way

themes/gatsby-theme-blog/src/theme/prism.js
jxnblk: It could be considered done for now. The prism colors are defined in `theme.colors.prism`, but I was originally thinking that maybe these colors could come from the base color palette instead of being defined separately

themes/gatsby-theme-blog/src/theme/styles.js
jxnblk: Yes, `theme.prism` object is spread into `theme.styles.pre`, which will add child selectors to the `<pre>` tag in MDX.

The `variant` key in `@styled-system/css` is more intended as a way to define styles for something like different button styles, but here it's working more like the `extend` feature from Sass, which I think is unnecessary. Instead this object should probably be imported into the module and spread into this object.

themes/gatsby-theme-blog/src/theme/typography.js
jxnblk: That's a good point. Since the ThemeProvider is imported into both `gatsby-ssr.js` and `gatsby-browser.js`, I think that'd work

themes/gatsby-theme-blog/gatsby-ssr.js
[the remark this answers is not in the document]
KyleAMathews: yeah they should. The typefaces export the font-files css so if we don't import them here, then the font css won't get added until after JS loads which would delay font loading a lot.

themes/gatsby-theme-blog/src/theme/styles.js
```diff
@@ -0,0 +1,73 @@
+export default {
+  root: {
+    fontFamily: `body`,
+  },
+  Container: {
```
ChristopherBiscardi: Some questions I can see people asking:

* What is `root`?
* Does this create a new `Container` component or apply to any `Container` component?
* This is shadowable on the theme-level, so how does it combine when multiple themes use different styles.js'

I think this styles API is going to confuse people since it basically adds API surface area that does the same thing as component shadowing while reducing the ability to discover provenance of the styles and thus deviates from typical React approaches to adding more props, etc.

Also interesting will be how we merge styles like this across themes.

```
div: {
    "&.gatsby-highlight": {
    }
}
```

themes/gatsby-theme-blog/src/components/layout.js
```diff
@@ -1,130 +1,124 @@
...
-        <div
-          style={{
-            color: `var(--textNormal)`,
-            background: `var(--bg)`,
-            transition: `color 0.2s ease-out, background 0.2s ease-out`,
-            minHeight: `100vh`,
-          }}
+    <Styled.root>
+      <Layout>
+        <Global
```
ChristopherBiscardi: `Global` will still bleed into shadowed components. For example, if `Layout` has a `Heading`, user shadows `Heading` and applies their own styles and they now have to account for the `Global` styles exerting influence on their `Heading`.

themes/gatsby-theme-blog/src/theme/styles.js
ChristopherBiscardi: yeah, I think especially since these are object styles, a direct import makes more sense.

themes/gatsby-theme-blog/src/theme/styles.js
jxnblk: - The `styles.root` styles are for setting top-level styles and the component is only rendered with `<Styled.root>` to give you control over where these styles apply. For Typography.js styles, this is used in place of styling the `html` and `body` elements. For this PR, it's added here: https://github.com/gatsbyjs/gatsby/pull/14377/files#diff-c228aeead46e1216bd9de630c447985bR67
- `styles.Container` will apply to theme-ui's `Container` component if it's used. For this theme, it has a narrower max-width than the default styles. I'm not 100% sold on this approach, and an alternative I've thought about is defining values like that in e.g. `theme.sizes.container` instead of coupling them to a specific component. There will probably also be situations where things like header height, sidebar width, etc. come into play in other themes
- The `theme` object from multiple themes (currently) will be [deeply merged](https://github.com/system-ui/theme-ui/blob/master/theme-ui/src/core.js#L23-L27), with the inner most ThemeProvider taking precedence. The `styles` object is a shortcut for defining styled components that are consumed in MDX. Part of the idea with system-ui/theme-ui#30 is to handle what's being done here with the `wrapRootElement` API in a Gatsby plugin so that we can change how that works without causing breaking changes in themes like this one and `gatsby-theme-notes`. If you've installed both the blog and notes theme, I think the idea would be that your site ends up with one merged `theme` object that both themes consume, making the `Container` component the same `max-width` for the entire site. I'm open to exploring alternatives to this (maybe theme-ui components are provided as a Gatsby theme that can be shadowed?), and can open an issue on the `theme-ui` repo to discuss different approaches.

As for the `div > .gatsby-highlight` style, I suspect we might want to look into updating the output HTML from `gatsby-remark-prismjs` to better fit in with how MDX works. Curious if you have any thoughts there, @ChristopherBiscardi.

themes/gatsby-theme-blog/src/components/layout.js
jxnblk: Yeah, I'm definitely open to better alternatives for how to handle this. Here, `Global` is being used for the dark and light color modes only. Without setting the body's background color, there will be a visible white area in dark mode when scrolling past the viewport.

There's also an issue where the background color will flash from white to dark when loading a page that has dark mode on. The `useDarkMode` hook [uses a script](https://github.com/donavon/use-dark-mode/blob/develop/noflash.js.txt) to apply classnames to the `body` to avoid this, and I'm guessing we'll want something similar for `theme-ui`, but without using global CSS. We could define "scoped" classnames in a `Global` component for this, but not sure if that will really avoid the issue of styles bleeding into shadowed components. Maybe `theme-ui` can include something like a `Colors` component to handle this in a way that doesn't expose Emotion's `Global` component in Gatsby themes.

themes/gatsby-theme-blog/src/utils/global.css
```diff
@@ -1,3 +1,4 @@
+/* todo: are the language specific styles needed here? */
```
jxnblk: @amberleyromo do you have any thoughts on the RTL styles in this file? I assume they're here for the translations in the Overreacted blog, but not sure if we do anything like this in our other starters

amberleyromo: > I assume they're here for the translations in the Overreacted blog

Indeed, indeed.

When I started this I intended to include the internationalization element. I think we can remove it for now. Would love to add this feature in later though.

themes/gatsby-theme-blog/gatsby-ssr.js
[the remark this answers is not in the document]
jxnblk: Added the typography imports in `src/theme/typography.js` as suggested in the other comment

themes/gatsby-theme-blog/src/theme/styles.js
```diff
@@ -0,0 +1,73 @@
...
+    ...prism,
+  },
+  code: {
+    fontFamily: `monospace`,
+    // from typography overrideThemeStyles
+    // "h1 code, h2 code, h3 code, h4 code, h5 code, h6 code"
+    fontSize: `inherit`,
+  },
+  // todo: prism seems to break MDX inlineCode
+  inlineCode: {
```
jxnblk: I haven't yet looked into why this isn't working, but can either do that here or follow up with another PR

amberleyromo: Noted this on the main PR as a todo

themes/gatsby-theme-blog/src/theme/styles.js
amberleyromo: @jxnblk good to merge this and continue pondering this outside of this PR?

jxnblk: Yeah definitely, mostly wanted to provide some context here, but not blocking

themes/gatsby-theme-blog/src/theme/styles.js
jxnblk: I *think* adding the `gatsby-remark-prismjs` `noInlineHighlight` option fixes this, but can confirm after merging

PR #14379, opened 2019-05-28
feat(gatsby-remark-images): add option to use alt text as caption fallback

chasemccoy:
`gatsby-remark-images` has a wonderful option to show captions under images using the `title` attribute, or the `alt` attribute if no `title` is provided. This causes us some issues, because we have a lot of images that we _don't_ want captions for, but we _do_ want to provide alt text (for accessibility reasons).

~This PR adds a new configuration option to the `gatsby-remark-images` plugin:~

~`fallbackToAltTextForCaptions` - (defaults to `true`) If `showCaptions` is set to `true` and the title attribute is empty, this specifies whether the alt attribute should be used as a caption. If true it will be used, and if false the caption will be omitted if there is no title attribute.~

This is my first time contributing to Gatsby, so my apologies if I get any of this wrong. I am not sure what we want to call the option (I went with `fallbackToAltTextForCaptions` because it was clear, but happy to pick a less verbose name). I also wrote a test to cover this case and updated the README of the plugin to document the new option.


chasemccoy:
@DSchau I think you helped review the PR to accept the alt attribute as a fallback caption (#13489), so tagging you here for visibility.

m-allanson:
Thanks @chasemccoy! Thinking aloud, here's an alternative approach (maybe this is a terrible idea though?) - instead of creating a new option, we could override the existing `showCaptions` option.

- Passing a boolean `true` or `false` would retain the existing behaviour
- Passing an array with multiple values like `["alt", "title"]` or `["title", "alt"]` would specify the order that the fields should be checked. The default source value would be the first string in the array, with any further items acting as fallback source values
- Passing a single value array would specify the only source value, e.g. [`title`]

This is similar to how the `withWebP` and `tracedSVG` options work, and hopefully covers any ordering someone would like to use.

> If we _really_ wanted to, we could allow passing a string `title` or `alt` as a shortcut for a single value array.

What are your thoughts @DSchau and @chasemccoy?

chasemccoy:
@m-allanson Hmm, I like that idea quite a lot! Question: would passing `true` be equivalent to `["title", "alt"]`? I believe that is the default behavior now (use title first, alt second if the `showCaption` option is true)

chasemccoy:
@m-allanson Okay, I just pushed up my first stab at this behavior. I wrote some new tests as well to make sure everything works as expected, and I updates the README to reflect this new behavior. My implementation assumed that `true` is equivalent to `["title", "alt"]`, but let me know of that's not the case.

packages/gatsby-remark-images/src/index.js
```diff
@@ -81,12 +81,28 @@ module.exports = (
...
+  const getImageCaption = (node, alt, defaultAlt) => {
+    const captionOptions = Array.isArray(options.showCaptions)
+      ? options.showCaptions
+      : options.showCaptions === true
+      ? [`title`, `alt`]
+      : false
+
+    if (captionOptions) {
+      for (let option of captionOptions) {
+        /*eslint-disable */
```
chasemccoy: @m-allanson eslint didn't like me using fall-throughs in my switch statement here, so I disabled it. Let me know if there is a better way to handle this!

superhawk610: Generally you want to avoid blanket disabling eslint so it will still catch other linting errors aside from the ones you are OK with making; explicit is always better :)

```
for (const option of captionOptions) {
  // eslint-disable-next-line no-fallthrough, default-case
  switch (option) {
    // ...
  }
}
```

chasemccoy: Thank you! I just didn't know the proper syntax for this. I will update the PR :)

chasemccoy: @superhawk610 updated! thanks for your help!

freiksenet:
There is a linter failure in the switch block, could you fix it please? Otherwise looking good. CC @m-allanson

chasemccoy:
@freiksenet Okay, it should be fixed now! Thanks!

chasemccoy:
@m-allanson @freiksenet Could y'all give a 👍 on this if it's ready? Would love to get this published so that my team can use it 🙂

m-allanson:
Thanks @chasemccoy! 🎉

chasemccoy:
Thank you all so much!

wardpeet:
Published in `gatsby-remark-images@3.0.15`

PR #14383, opened 2019-05-29
fix(gatsby-cli): build successfully without optional dependencies

### LINKED ISSUE #14370: installing gatsby with --no-optional breaks gatsby-cli
## Description
When people install gatsby with `--no-optional` gatsby breaks because ink is not being installed. We made ink optional because of node 6 #14233 

We should fix our check here
https://github.com/gatsbyjs/gatsby/blob/6070279f68aab0c82f2d244f54088f8ad4a74503/packages/gatsby-cli/src/reporter/reporters/index.js#L4-L8

```js
let inkExists;
try {
  require.resolve('ink')
  inkExists = true
} catch(err) {}

if (inkExists && !isCI) {
```

### Steps to reproduce
open a gatsby project and run
`npm install --no-optional`
`npm run build build`

### Expected result
Gatsby should build the website.

### Actual result
```
> gatsby-starter-default@0.1.0 build C:\Users\WardPeeters\projectswin\tmp\gatsby\my-default-starter
> gatsby build

internal/modules/cjs/loader.js:584
    throw err;
    ^

Error: Cannot find module 'ink'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:582:15)
    at Function.Module._load (internal/modules/cjs/loader.js:508:25)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:22:18)
    at Object.<anonymous> (C:\Users\WardPeeters\projectswin\tmp\gatsby\my-default-starter\node_modules\gatsby\node_modules\gatsby-cli\lib\reporter\reporters\ink\index.js:10:12)
    at Module._compile (internal/modules/cjs/loader.js:701:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:712:10)
    at Module.load (internal/modules/cjs/loader.js:600:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:539:12)
    at Function.Module._load (internal/modules/cjs/loader.js:531:3)
```

Copied from https://github.com/gatsbyjs/gatsby/pull/14233#issuecomment-495895945

ryandrew14:
## Description
This PR solves the issue raised in #14370, where trying to build the site after installing without optional dependencies will fail due to `ink` not being installed. The change uses the straightforward solution proposed by @wardpeet of ensuring the module exists before using it, and defaulting to yurna in the case that it does not exist.

## Related Issues
Fixes #14370

### DIFF
--- a/packages/gatsby-cli/src/reporter/reporters/index.js
+++ b/packages/gatsby-cli/src/reporter/reporters/index.js
@@ -1,3 +1,4 @@
+const semver = require(`semver`)
 const { isCI } = require(`ci-info`)
 
 let inkExists = false
@@ -6,7 +7,7 @@ try {
   // eslint-disable-next-line no-empty
 } catch (err) {}
 
-if (inkExists && !isCI) {
+if (inkExists && semver.satisfies(process.version, `>=8`) && !isCI) {
   module.exports = require(`./ink`).default
 } else {
   module.exports = require(`./yurnalist`)
--- a/packages/gatsby-cli/src/reporter/reporters/index.js
+++ b/packages/gatsby-cli/src/reporter/reporters/index.js
@@ -1,7 +1,12 @@
-const semver = require(`semver`)
 const { isCI } = require(`ci-info`)
 
-if (semver.satisfies(process.version, `>=8`) && !isCI) {
+let inkExists = false
+try {
+  inkExists = require.resolve(`ink`)
+  // eslint-disable-next-line no-empty
+} catch (err) {}
+
+if (inkExists && !isCI) {
   module.exports = require(`./ink`).default
 } else {
   module.exports = require(`./yurnalist`)


packages/gatsby-cli/src/reporter/reporters/index.js
```diff
@@ -1,7 +1,12 @@
-const semver = require(`semver`)
 const { isCI } = require(`ci-info`)
 
-if (semver.satisfies(process.version, `>=8`) && !isCI) {
+let inkExists = false
+try {
+  inkExists = require.resolve(`ink`)
```
ryandrew14: Just wondering -- why do I see backticks used everywhere for normal strings inside requires? I followed this convention here, but I am not sure if I understand it.

packages/gatsby-cli/src/reporter/reporters/index.js
```diff
@@ -1,7 +1,12 @@
-const semver = require(`semver`)
 const { isCI } = require(`ci-info`)
 
-if (semver.satisfies(process.version, `>=8`) && !isCI) {
```
pieh: Can we still keep node version check - if for some reason `ink` is actually installed in node 6 env, we should not try to use it as it will crash

packages/gatsby-cli/src/reporter/reporters/index.js
pieh: This is just convention we use here. It's much nicer to work with template literals - in case we need to use interpolation we don't have to convert from "normal strings" to template literal

packages/gatsby-cli/src/reporter/reporters/index.js
ryandrew14: Got it, thanks! Updating now.

pieh:
`gatsby-cli@2.6.3` and `gatsby@2.7.5` were published

PR #14384, opened 2019-05-29
fix(gatsby-plugin-google-tagmanager): Properly communicate site title with GTM services

### LINKED ISSUE #13968: Google Tag Manager plug-in captures incorrect page title
## Description

Google Analytics captures incorrect page title when using Google Tag Manager via gatsby-plugin-google-tagmanager. See solution below.

### Steps to reproduce

1. Install plug in
2. Configure GTM and GA within GTM
3. Per plug-in ([README](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-google-tagmanager/README.md)) wire GA to the history change event in GTM.

### Expected result

Page navigation (internal and external) would correctly get recorded as page views with the correct titles.

### Actual result

Incorrect titles are captured. This is the same issue as discussed [here](https://github.com/gatsbyjs/gatsby/issues/9139) where the title is a page view behind because Helmet hasn't had a chance to update the page title yet.

### Environment

Not relevant to the issue but for completeness

```
  System:
    OS: macOS 10.14.4
    CPU: (8) x64 Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 11.12.0 - /usr/local/bin/node
  Languages:
    Python: 3.7.2 - /usr/local/opt/python/libexec/bin/python
  Browsers:
    Chrome: 74.0.3729.131
    Safari: 12.1
  npmPackages:
    gatsby: ^2.3.16 => 2.3.16 
    gatsby-cli: ^2.5.12 => 2.5.12 
    gatsby-image: ^2.0.37 => 2.0.37 
    gatsby-plugin-google-tagmanager: ^2.0.13 => 2.0.13 
    gatsby-plugin-manifest: ^2.0.29 => 2.0.29 
    gatsby-plugin-robots-txt: ^1.4.0 => 1.4.0 
    gatsby-plugin-root-import: ^2.0.5 => 2.0.5 
    gatsby-plugin-sass: ^2.0.11 => 2.0.11 
    gatsby-plugin-sharp: ^2.0.32 => 2.0.32 
    gatsby-plugin-sitemap: ^2.0.12 => 2.0.12 
    gatsby-source-filesystem: ^2.0.29 => 2.0.29 
    gatsby-transformer-sharp: ^2.1.17 => 2.1.17 
  npmGlobalPackages:
    gatsby-cli: 2.5.9
```

### Solution

In gatsby-browser.js fire a custom GTM event in onRouteUpdate - 
`window.dataLayer.push({'event': 'page-change'})`. Note the "page-change" is arbitrary.

In GTM create a trigger for that custom event and wire GA (and other relevant plug-ins) to it. Don't wire those plug-ins to the built in page view event nor a custom history change event.

My gatsby-browser.js code is based on the [recent PR](https://github.com/gatsbyjs/gatsby/pull/12033) to the gtag plug-in.

### Suggested Changes

At a minimum the README should be updated. It may also make sense to fire the custom event from the plug-in and tell people to wire it in the GTM UI.

ryandrew14:
## Description
This PR registers a Google Tag Manager event that will fire whenever the route in a Gatsby application changes. This fixes a previously described issue (#13968) which led to weird mismatched site titles and routes in Google Analytics. It also updates the documentation to give an example of how to use this event in your Google Analytics to detect Gatsby route changes.

I'm somewhat new to production-level JS development and I know this may not be entirely optimal (the 2000ms delay especially, it was just a number that seemed reasonable since some delay was necessary), so let me know what I can do to make this better!

## Related Issues
Fixes #13968

### DIFF
--- a/packages/gatsby-plugin-google-tagmanager/src/gatsby-browser.js
+++ b/packages/gatsby-plugin-google-tagmanager/src/gatsby-browser.js
@@ -2,5 +2,5 @@ exports.onRouteUpdate = ({ location }) => {
   // wrap inside a timeout to ensure the title has properly been changed
   setTimeout(() => {
     window.dataLayer.push({ event: `gatsby-route-change` })
-  }, 2000)
+  }, 50)
 }
--- a/packages/gatsby-plugin-google-tagmanager/src/gatsby-browser.js
+++ b/packages/gatsby-plugin-google-tagmanager/src/gatsby-browser.js
@@ -1,0 +1,6 @@
+exports.onRouteUpdate = ({ location }) => {
+  // wrap inside a timeout to ensure the title has properly been changed
+  setTimeout(() => {
+    window.dataLayer.push({ event: `gatsby-route-change` })
+  }, 2000)
+}
--- a/packages/gatsby-plugin-google-tagmanager/README.md
+++ b/packages/gatsby-plugin-google-tagmanager/README.md
@@ -29,6 +29,18 @@ plugins: [
 ]
 ```
 
+#### Tracking routes
+
+This plugin will fire a new event called `gatsby-route-change` whenever a route is changed in your Gatsby application. To record this in Google Tag Manager, we will need to add a trigger to the desired tag to listen for the event:
+
+1. Visit the [Google Tag Manager console](https://tagmanager.google.com/) and click on the workspace for your site.
+2. Navigate to the desired tag using the 'Tags' tab of the menu on the right hand side.
+3. Under "Triggering", click the pencil icon, then the "+" button to add a new trigger.
+4. In the "Choose a trigger" window, click on the "+" button again.
+5. Choose the trigger type by clicking the pencil button and clicking "Custom event". For event name, enter `gatsby-route-change`.
+
+This tag will now catch every route change in Gatsby, and you can add Google tag services as you wish to it.
+
 #### Note
 
-Out of the box this plugin will simply load Google Tag Manager on the initial page/app load. It's up to you to fire tags based on changes in your app. To automatically track page changes, in GA for instance, you can configure your tags in GTM to trigger on [History Change](https://support.google.com/tagmanager/answer/7679322?hl=en&ref_topic=7679108).
+Out of the box this plugin will simply load Google Tag Manager on the initial page/app load. It's up to you to fire tags based on changes in your app. See the above "Tracking routes" section for an example.


ryandrew14:
This failure is odd -- installing packages and running tests works great locally. Any ideas on what's happening?

pieh:
> This failure is odd -- installing packages and running tests works great locally. Any ideas on what's happening?

This is probably network error / flake and we can ignore it for now (will restart tests before merging)

packages/gatsby-plugin-google-tagmanager/src/gatsby-browser.js
```diff
@@ -0,0 +1,6 @@
+exports.onRouteUpdate = ({ location }) => {
+  // wrap inside a timeout to ensure the title has properly been changed
+  setTimeout(() => {
+    window.dataLayer.push({ event: `gatsby-route-change` })
+  }, 2000)
```
pieh: That's quite big timeout IMO, requesting animiation frame or `0`s timeout should work here too.

Also we should keep track of timeout and cancel it if another `onRouteUpdate` call was triggered to avoid pushing `gatsby-route-change` event twice for same path

ryandrew14: In what case is another onRouteUpdate triggered? I would think that if someone visits one page, then quickly visits another, we should log both page views.

pieh: Right, but this object you are pushing doesn't include any page metadata (just event name), so I assumed it's GTM services that extracts information on events. So if something like this would happen:

 - navigate to `/page-A`, (start setTimeout)
 - quickly (< 2s) navigate to `/page-B` (start another setTimeout)
 - first timeout callback is exucuted (push event when on `/page-B`)
 - second timeout callback is exucuted (push event when on `/page-B`)

In any case - I think 2000 is just too long here. If you are using `react-helmet` to handle `<title>` etc, it should be enough to use `setTimeout(<some-code>, 0)`

ryandrew14:
@pieh 0ms timeout was still mismatching title/route occasionally, so I upped it to 50ms to be safe.

packages/gatsby-plugin-google-tagmanager/src/gatsby-browser.js
ryandrew14: Got it, thanks for the insight. I drastically lowered the delay -- do you still think it's necessary to keep a reference to the timeout to cancel it?

pieh:
`gatsby-plugin-google-tagmanager@2.0.14` was published

PR #14385, opened 2019-05-29
revert gatsby-plugin-offline and update getResourceURLsForPathname

Moocar:

## Description

Addresses https://github.com/gatsbyjs/gatsby/pull/13004#issuecomment-496310446. For the `per-page-manifest` work, I replaced `onPostPrefetchPathname` with `onPostPrefetch`, which is passed the prefetched resource URLs vs forcing the offline plugin to request them via `getResourceURLsForPathname`. Unfortunately, this broke backwards compatibility, since I removed `onPostPrefetchPathname`, which is a gatsby browser API. At the time, I thought this was necessary, since there's no prefetching of query results by data path anymore. But having reviewed the code again, I realized we can simply pass the prefetch URL for the page-data.json instead.

So this PR reverts all changes to the offline plugin, and updates `getResourceURLsForPathname` to return the page-data.json URL instead of the query result URL.

## Related Issues

### DIFF
--- a/packages/gatsby-plugin-offline/src/gatsby-browser.js
+++ b/packages/gatsby-plugin-offline/src/gatsby-browser.js
@@ -1,9 +1,12 @@
 exports.registerServiceWorker = () => true
 
-const prefetchedResources = []
+const prefetchedPathnames = []
 const whitelistedPathnames = []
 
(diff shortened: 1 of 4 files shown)


Moocar:
ping @wardpeet @DSchau @KyleAMathews

PR #14399, opened 2019-05-29
Submit Submittable to showcase

g-crow:


docs/sites.yml
```diff
@@ -6114,16 +6114,14 @@
   built_by: Luke Secomb
   built_by_url: https://lukesecomb.digital
   featured: false
-- title: Global City Data
-  main_url: https://globalcitydata.com
-  url: https://globalcitydata.com
-  source_url: https://github.com/globalcitydata/globalcitydata
```
pieh: I think something went wrong when resolving merge conflicts? We don't want to remove "Global City Data" entry ;)

g-crow: Oops! I will try that again

PR #13293, opened 2019-04-11
chore(gatsby-source-mongodb): Use new createContentDigest helper

romulomachado:
This PR makes use of the new createContentDigest helper in gatsby-source-mongodb.

## Related Issues

Related to #8805

PR #14402, opened 2019-05-29
fix(gatsby-source-mongodb): fix mapping function signature

elmarburke:
## Description

In the mapping function (`map: {...}` in options), there was a mismatch of the function parameter.

Used `createNode` in place of `createContentDigest`.

## Related Issues

#13293


freiksenet:
Thanks!

PR #14409, opened 2019-05-29
chore(docs): updated selections in link

muescha:
## Description

updated selections in link

### DIFF
--- a/docs/docs/creating-a-source-plugin.md
+++ b/docs/docs/creating-a-source-plugin.md
@@ -142,7 +142,7 @@ In order to create a parent/child relationship, when calling `createNode` for th
 
 _Examples_
 
-[Here's the above example](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-transformer-remark/src/on-node-create.js#L34-L64) from the `gatsby-transformer-remark` source plugin.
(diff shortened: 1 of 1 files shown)


docs/docs/creating-a-source-plugin.md
```diff
@@ -142,7 +142,7 @@ In order to create a parent/child relationship, when calling `createNode` for th
 
 _Examples_
 
-[Here's the above example](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-transformer-remark/src/on-node-create.js#L34-L64) from the `gatsby-transformer-remark` source plugin.
+[Here's the above example](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-transformer-remark/src/on-node-create.js#L39-L67) from the `gatsby-transformer-remark` source plugin.
```
pieh: Maybe we should pin link to commit SHA so those lines don't move around?
```suggestion
[Here's the above example](https://github.com/gatsbyjs/gatsby/blob/72077527b4acd3f2109ed5a2fcb780cddefee35a/packages/gatsby-transformer-remark/src/on-node-create.js#L39-L67) from the `gatsby-transformer-remark` source plugin.
```

muescha: Yes. Thats better. 

PS: I checked the other links with hash and they are not changed since then

PR #14423, opened 2019-05-30
feat(gatsby): Allow sorting on resolved fields

stefanprobst:
* resolve `sort` fields as well as `filter` fields
* move some utility functions
* don't use `lodash.get` for getting values for sorting, to support array values


DSchau:
@stefanprobst could you provide a few things, specifically:

1. Slightly more in-depth description so it's clearer what this enables
1. An example of using this feature
1. Perhaps a unit test and/or an e2e test so we ensure we don't break this in the future?

stefanprobst:
@DSchau For context: Sorting on fields which don't exist on the nodes themselves, but are created by a field resolver has never worked in Gatsby. For example when defining a `frontmatter.author___NODE` field which links to an `AuthorJson` node, it is not possible to sort on `frontmatter.author.name`.
Note that *filtering* on this field *does* work. This is because for every `filter` field we manually call the field resolvers so all fields are available for querying. We currently don't do the same for `sort` fields. This PR fixes that.

The change from `lodash.get` to a homegrown utility has to do with the fact that we can make certain assumptions in a graphql schema context which a more generic utility cannot make, namely that the items in an array will always be of the same type.

There are [six unit tests](https://github.com/gatsbyjs/gatsby/pull/14423/files#diff-60f42b2e8402c867a19324d7dc2edb5d) included - anything specific that's missing?

Related:  #11368

DSchau:
@stefanprobst 🤦‍♂ I missed the tests somehow--thank you!

freiksenet:
@stefanprobst looking good, I'll merge after I do sanity check against some repos.

freiksenet:
Looking good, I'll publish and then merge it to publish as a minor.

PR #14443, opened 2019-05-30
fix(gatsby): Use anonymous requests when fetching resources

### LINKED ISSUE #14293: Make requests not send credentials
## Summary

Make gatsby not send credentials when making requests - currently in the following 2 places:

https://github.com/gatsbyjs/gatsby/blob/eaa08cfd7d9bf3e34209eec58a488585f25726e2/packages/gatsby/cache-dir/loader.js#L87

https://github.com/gatsbyjs/gatsby/blob/eaa08cfd7d9bf3e34209eec58a488585f25726e2/packages/gatsby/cache-dir/static-entry.js#L300

### Motivation

We have a slightly weird setup where our assets are on a CDN (enabled by the new `assetPrefix` option), which is a different domain from where the pages are served. This means we have to serve the assets with CORS headers. Moreover, we need them accessible from multiple origins and the headers must be static so we have to set `Access-Control-Allow-Origin` to `*`. Having this header set this way means that browsers error when doing XHRs with credentials.

My initial thought was that credentials can be removed for everyone without causing issues, but some people may have their static sites behind an authentication proxy which means credentials will need to be sent in some cases.

cameron-martin:
## Description

The [tests that I ran](https://github.com/gatsbyjs/gatsby/issues/14293#issuecomment-495654905) suggest that this is no longer needed, and it fixes the [issues I was experiencing](https://github.com/gatsbyjs/gatsby/issues/14293#issue-448125073).

## Related Issues

Fixes #14293.


pieh:
Sorry for long time without reply.

This looks correct, but there are few more changes that I think are needed for it to be complete:
- https://github.com/gatsbyjs/gatsby/blob/8dd7693c39e97af9c777bb72010ac4afd6e75711/packages/gatsby/cache-dir/prefetch.js#L39-L43
this is our fallback prefetching implementation (for browsers that don't support `<link rel="prefetch">`), and we most likely need to drop `withCredentials` here as well.
- https://github.com/gatsbyjs/gatsby/blob/master/docs/docs/asset-prefix.md#usage-with-gatsby-plugin-offline This probably need updates (or just removal) - will have to test it out if there is still extra setup needed

cameron-martin:
I've removed the other `withCredentials` and updated the docs.

Regarding the docs, `Access-Control-Allow-Origin` is still needed, but it can be set to `*` if credentials are not being sent.

pieh:
There is some trickiness that we would need to solve with those changes, when using asset prefix and relative `startUrl` in manifest file:
Screenshot 2019-06-14 at 19 40 44
This can be fixed by not using asset path when adding `<link rel="manifest">` or by using full path in manifest `startUrl` (instead of just `/` which we default to right now)

cameron-martin:
@pieh I'm not sure that I follow how this change introduces that issue.

pieh:
> @pieh I'm not sure that I follow how this change introduces that issue.

Sorry, I should clarify - this is about those changes in asset prefix docs (not the code changes). In the end we might need to revert those, I will need to do more testing on what's working and what doesn't. We also maybe will make change to manifest plugin itself there.

cameron-martin:
It seems that a relative `start_url` wouldn't perform as intended when the manifest is on a different domain, since [the manifest URL is used as the base URL](https://developer.mozilla.org/en-US/docs/Web/Manifest/start_url#Relative_URL) when resolving the relative URL.

I still don't understand how setting CORS headers differently would make the situation worse?

pieh:
> I still don't understand how setting CORS headers differently would make the situation worse?

This is pretty confusing to me too. It's not clear to me why browser is not complaining in this case - it might be weird thing that Chrome does? I didn't test it in other browsers to be honest. 

I think most sense is not to use asset prefix for manifest (so it always loads from the same origin as site itself). The problem here is that users would need to know to update both gatsby and manifest plugin so they are still compatible and we don't have anything to provide some warnings in case only gatsby package is updated :(

cameron-martin:
> This is pretty confusing to me too. It's not clear to me why browser is not complaining in this case - it might be weird thing that Chrome does?

Does the manifest actually work correctly (i.e. open the correct page on startup) before the CORS headers were changed?

> I think most sense is not to use asset prefix for manifest (so it always loads from the same origin as site itself).

Yeah that seems sensible. If I remember correctly, this already happens for the service worker.

pieh:
> Yeah that seems sensible. If I remember correctly, this already happens for the service worker.

Yeah, it does. Service worker registration is handled by gatsby core, while adding manifest link is done by plugin (so core can be independently updated and there can be situations where old manifest plugin is used together we new gatsby core producing broken experience/results :( )

cameron-martin:
Hmm, I'm can't think of any sensible way to avoid that.

xavivars:
Not sure I understand where this PR's status currently is... Anything that can be done to help?

pieh:
> Not sure I understand where this PR's status currently is... Anything that can be done to help?

Yeah, sorry about delay on that - we are busy with cleaning up bugs that creeped in recently with couple of major changes/additions (like themes and per-page-manifest).

This does need changes to `gatsby-plugin-manifest`, as well as some way to handle potentially mismatched versions of gatsby and gatsby-plugin-manifest

xavivars:
@cameron-martin it seems @pieh was waiting for your response in order for this PR to progress.

cameron-martin:
Oh sorry, is this a response on how to deal with mismatched versions? I don't have any good ideas on this unfortunately, although here is a few not-so-good ones to use as a starting point/possible inspiration:

* Emit a warning/error from Gatsby core at some point if it detects incompatible versions of plugins. However this necessarily makes Gatsby core in a sense dependent upon specific instances of plugins (if just by name), which is the wrong dependency direction.
* Increase the version of Gatsby such that it no longer satisfies the peer dependency version constraint of the plugin. However I imagine this would require a major version increment (haven't checked the constraint because I'm on my phone) and it seems that major version changes are rare and should have considerable changes in them so I'm not sure this is appropriate.
* Rely on people noticing the warning emitted by the browser and updating the plugin to fix it.

id0Sch:
is there anything new about this PR?
we really want to deploy our brand new gatsby site, and this is the only blocker :(
I would be happy to assist in any way if needed.

sidharthachatterjee:
@pieh Can you elaborate on what we need to get done in this to unblock it?

sidharthachatterjee:
So I'm just discussing this with @pieh right now and it looks like we'll have to do the following to get this in 

- We need to add a compatibility later to ensure that if the user only updates one of `gatsby` (and not `gatsby-plugin-manifest`) and if user happens to use `assetPrefix` then the domain will not be inferred correctly in case of relative `start_url` in manifest

sidharthachatterjee:
@pieh and I just fixed the edge case!

xavivars:
@sidharthachatterjee why does manifest needs to be be loaded from the HTML server, and can't be downloaded from the asset one? As far as I understood, the problem was with start_url not being a fully qualified url (meaning, starting with /), but if it's a full URL is there any problem with that?

PR #14444, opened 2019-05-30
chore: Update axios to 0.19.0 preventing potential DoS

rayriffy:
## Description

**CVE-2019-10742** is now official for axios which allows attackers to cause DoS attack by exceeding `maxContentLength`. This is patched in https://github.com/axios/axios/pull/1485 and released in 0.19.0

### DIFF
--- a/packages/gatsby-transformer-sqip/package.json
+++ b/packages/gatsby-transformer-sqip/package.json
@@ -8,7 +8,6 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.0.0",
-    "axios": "^0.19.0",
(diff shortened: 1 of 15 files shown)


packages/gatsby-transformer-sqip/package.json
```diff
@@ -8,7 +8,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.0.0",
-    "axios": "^0.18.0",
+    "axios": "^0.19.0",
```
m-allanson: It looks like this dep isn't used, I think it can be removed.

```suggestion
```

PR #14449, opened 2019-05-31
fix(gatsby-source-contentful): add query to the subsequent api call

sbezludny:
## Description

PR adds a missing argument `query` to the `pagedGet` invocation. See definition on [L134](https://github.com/gatsbyjs/gatsby/pull/14449/files#diff-6e25753afbe27b5e9223708cc884666fL134)

### DIFF
--- a/packages/gatsby-source-contentful/src/fetch.js
+++ b/packages/gatsby-source-contentful/src/fetch.js
@@ -153,6 +153,7 @@ function pagedGet(
         method,
         query,
         skip + pageLimit,
+        pageLimit,
         aggregatedResponse
       )
     }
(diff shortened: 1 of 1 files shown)


packages/gatsby-source-contentful/src/fetch.js
```diff
@@ -148,7 +148,13 @@ function pagedGet(
       aggregatedResponse.items = aggregatedResponse.items.concat(response.items)
     }
     if (skip + pageLimit <= response.total) {
-      return pagedGet(client, method, skip + pageLimit, aggregatedResponse)
+      return pagedGet(
+        client,
+        method,
+        query,
+        skip + pageLimit,
```
pieh: Function signature is:
```
function pagedGet(
  client,
  method,
  query = {},
  skip = 0,
  pageLimit = 1000,
  aggregatedResponse = null
)
```

and seems like we are also missing `pageLimit`?
```suggestion
        skip + pageLimit,
        pagelimit,
```

sbezludny:
fixed.

On a side note, this function (`pagedGet`) is used solely to fetch content types, and the bug this pr aims to address will be reproducible only if the number of content types is greater than 1000.
But if I remember correctly, a technical limit on content types per space is exactly 1000, therefore it should be safe to get rid of this logic in future.

PR #14456, opened 2019-05-31
chore(starters): add gatsby-starter-developer-diary

willjw3:
## Description

Humbly requesting that my starter be added to the Gatsby Starter Library


willjw3:
Thank you!

PR #14464, opened 2019-05-31
Fix the __experimentalThemes namespace

goodjack:
## Description

If I declare the theme like the following
```
__experimentalThemes: [`gatsby-theme-docz`]
```

It will show up the Gatsby Default Starter pages instead of the Docz pages

### DIFF
--- a/docs/tutorial/writing-documentation-with-docz.md
+++ b/docs/tutorial/writing-documentation-with-docz.md
@@ -35,7 +35,11 @@ Define `docz-theme-default` as a theme inside the `__experimentalThemes` of `gat
 ```js:title=gatsby-config.js
 module.exports = {
   //highlight-next-line
-  __experimentalThemes: [`gatsby-theme-docz`],
+  __experimentalThemes: [
+    {
+      resolve: "gatsby-theme-docz",
+    },
+  ],
   plugins: [`// your plugins go here`],
 }
 ```


muescha:
The only change i have found: now you use double quote `“` instead of backtics ```

muescha:
One more sidenote: in more than 50% of the docs we use double quotes but in code always backticks ([Search for `__experimentalThemes`](https://github.com/gatsbyjs/gatsby/search?utf8=%E2%9C%93&q=__experimentalThemes&type=))

goodjack:
> The only change i have found: now you use double quote `“` instead of backtics `

Oops, the code was copied from Docz's readme file
https://github.com/pedronauck/docz/tree/master/core/gatsby-theme-docz

and formatted by [Prettier](https://github.com/prettier/prettier).

goodjack:
> Could you provide more information as to what this is doing and what it's fixing?
> 
> In theory (and hopefully in practice) these should be doing effectively the same thing, so I don't as of yet understand the need!

I followed the tutorial to create a new site.

When I defined the `__experimentalThemes` given in this tutorial
```
__experimentalThemes: [`gatsby-theme-docz`],
```

and created the `docs/index.mdx`, I saw this:
image


If I modify `__experimentalThemes`:
```
__experimentalThemes: [
  {
    resolve: 'gatsby-theme-docz',
  },
],
```

It will show the correct result
image

I'm not sure why.

DSchau:
@goodjack this seems like a bug with Gatsby Themes, rather than an issue we'll want to "fix" with documentation.

I'll open an issue so we can fix the underlying issue. Thanks for opening this, but going to close as we'd prefer to fix the issue rather than workaround it!

goodjack:
I'll leave it to you. Thanks.

Issue #14300, opened 2019-05-24
[gatsby-source-shopify] Timeouts during load; Limit of 250 is too large?


cj2:
## Description

Gatsby Shopify source plugin throws unhandled Timeout exception while building cache. When modifying the GraphQL query to use `first: 150` rather than `first: 250`, the cache is built and Gatsby starts.

### Steps to reproduce

1. Create a new Gatsby project
2. Add the Shopify source with API key (our store has 4300 products inclusive of unique variants)
3. Run `yarn start`

### Expected result

Gatsby starts normally, with a fully built Shopify cache.

### Actual result

Gatsby fails to start

```
npm WARN lifecycle The node binary used for scripts is /var/folders/zr/18gw2jpj4xn9lfnl8rktth7m0000gp/T/yarn--1558714436525-0.14726619247761996/node but npm is using /Users/cj/.nvm/versions/node/v12.3.1/bin/node itself. Use the `--scripts-prepend-node-path` option to include the path for the node binary npm was executed with.

> gatsby-starter-default@0.1.0 develop /Users/cj/Companies/aspenandpine/aspenandpine-gatsby
> gatsby develop

success open and validate gatsby-configs — 0.006 s
success load plugins — 1.127 s
success onPreInit — 0.010 s
success initialize cache — 0.061 s
success copy gatsby files — 0.090 s
success onPreBootstrap — 0.018 s

gatsby-source-shopify/aspen-pine starting to fetch data from Shopify

gatsby-source-shopify/aspen-pine fetched and processed blogs: 1433.211ms

gatsby-source-shopify/aspen-pine fetched and processed productTypes: 1461.054ms

gatsby-source-shopify/aspen-pine fetched and processed policies: 1477.566ms
error (node:47418) [DEP0066] DeprecationWarning: OutgoingMessage.prototype._headers is deprecated

gatsby-source-shopify/aspen-pine fetched and processed collections: 3005.577ms


gatsby-source-shopify/aspen-pine fetched and processed articles: 3365.866ms
error 
error an error occured while sourcing data
error - 
  message: Timeout
error query: 
  """
    
      query GetProducts($first: Int!, $after: String) {
        shop {
          products(first: $first, after: $after) {
            pageInfo {
              hasNextPage
            }
            edges {
              cursor
              node {
                availableForSale
                createdAt
                description
                descriptionHtml
                handle
                id
                images(first: 250) {
                  edges {
                    node {
                      id
                      altText
                      originalSrc
[...]

PR #14470, opened 2019-05-31
feat(gatsby-source-shopify): add configurable 'paginationSize' parameter

cj2:
## Description

Cache building queries at startup result in timeouts. This PR adds a configurable "first" parameter to allow developers to reduce the "first" parameter on the GraphQL queries to Shopify.

## Related Issues

[Addresses #14300](https://github.com/gatsbyjs/gatsby/issues/14300)

### DIFF
--- a/packages/gatsby-source-shopify/README.md
+++ b/packages/gatsby-source-shopify/README.md
@@ -52,6 +52,11 @@ plugins: [
       // much time was required to fetch and process the data.
       // Defaults to true.
       verbose: true,
+
+      // Number of records to fetch on each request when building the cache
+      // at startup. If your application encounters timeout errors during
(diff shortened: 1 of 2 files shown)


KyleAMathews:
This is great!

Perhaps we could name this option something like `paginationSize`? Also, if 250 is reliably causing you trouble, does it slow things down much to change it to 150?

cj2:
> This is great!
> 
> Perhaps we could name this option something like `paginationSize`? Also, if 250 is reliably causing you trouble, does it slow things down much to change it to 150?

pagniationSize makes sense, change committed. I ended up settling on a page size of 100 to achieve consistent reliability. For us, it builds the cache in less than a minute. Pre-building all of the required pages is a different story :)

PR #14472, opened 2019-05-31
feat(gatsby): update graphiql-explorer

pieh:
This brings all new niceties - see for what changed https://www.onegraph.com/blog/2019/05/30/GraphiQL_Explorer_2_0_A_Power_User_Release.html 

We just needed small tweak to hide "add new mutation"/"add new subscription" buttons 
that would be pretty confusing to Gatsby users and @sgrove just pushed changes ( https://github.com/OneGraph/graphiql-explorer/commit/56df94da93e7ee280a131953cda8cfd3851f92fd ) and published it, so we are ready for update :)

### DIFF
--- a/packages/gatsby-graphiql-explorer/package.json
+++ b/packages/gatsby-graphiql-explorer/package.json
@@ -27,12 +27,7 @@
   },
   "license": "MIT",
   "dependencies": {
-    "@babel/runtime": "^7.0.0",
-    "graphiql": "^0.13.0",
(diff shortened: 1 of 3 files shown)


pieh:
There are some things I need to add this - so please don't merge yet

pieh:
Ok, this is ready for review. I also moved most deps to devDeps as we are bundling those, so users doesn't need to install those directly and be greeted with lot of peerDeps warning:

Screen Shot 2019-05-31 at 1 35 39 PM

PR #14483, opened 2019-06-01
fix(gatsby): Set CORS to * in serve 

stutrek:
I'm a long time user, first time contributor. Thanks for the great framework, the development experience for using and editing Gatsby is incredible.

## Description

Access control headers were added to the server about a month ago, using the host provided from the command line. I am running the server with `gatsby serve --host 0.0.0.0 --port 80` because it's being served on multiple domains, with routing and https handled before Gatsby. Currently these headers are served:

image

Unless I'm misunderstanding something, access control headers aren't useful in this situation because they'll always be the domain the browser is already on (with the exception of my 0.0.0.0 scenario). These headers were added [here](https://github.com/gatsbyjs/gatsby/pull/12128/files#r260549304) as part of #12128 so there may be a situation with `assetPrefix` on another domain that I'm unaware of. @DSchau can you confirm or deny? Should this be a CLI flag when running the server insetad?

I might not be running the tests properly, no tests failed after this change.

## Related Issues

Introduced by #12128


stutrek:
I don't have the slightest idea what to do about this Danger failure.

pieh:
This might be unnecesary or changed after https://github.com/gatsbyjs/gatsby/pull/14443 - will do some investigation.

> I don't have the slightest idea what to do about this Danger failure.

This is flake - somethings Peril/danger (tools we use for some workflows automation) errors out randomly with confusing message. You can ignore it

pieh:
Ok. let's not remove cors entirely - but with https://github.com/gatsbyjs/gatsby/pull/14443 we will be able to use wildcard for allowed origins (and remove allow-credentials).

pieh:
This is blocked until #14443 is merged.

PR #14492, opened 2019-06-01
Track JS errors on gatsbyjs.org with Sentry

KyleAMathews:
/cc @Moocar


docs/starters.yml
```diff
@@ -2718,8 +2718,6 @@
     - Styling with SCSS
     - Offline support
     - Web App Manifest
-- url: https://gatsby-starter-material-business-markdown.netlify.com/
```
KyleAMathews: this was breaking the build

www/gatsby-config.js
```diff
@@ -214,6 +215,7 @@ module.exports = {
               {
                 allMdx(
                   sort: { order: DESC, fields: [frontmatter___date] }
+                  limit: 10,
```
pieh: Is this intentional?

www/gatsby-config.js
```diff
@@ -282,6 +284,16 @@ module.exports = {
...
       },
     },
+    {
+      resolve: `gatsby-plugin-sentry`,
+      options: {
+        dsn: `https://2904ad31b1744c688ae19b627f51a5de@sentry.io/1471074`,
+        release: git.long(),
+        environment: process.env.NODE_ENV,
+        enabled: (() =>
+          [`production`, `stage`].indexOf(process.env.NODE_ENV) !== -1)(),
```
pieh: is IIFE needed here?

www/gatsby-config.js
KyleAMathews: Yeah, rss feeds typically only show the ten newest posts and rendering the MDX here takes a long time as it's rendering every blog post.

www/gatsby-config.js
KyleAMathews: I didn't test it but the idea I believe is to only send a true or false to the browser not a function (which wouldnt work anyways).

pieh: wouldn't
```js
enabled: [`production`, `stage`].indexOf(process.env.NODE_ENV) !== -1)
```
do the same? ;)

KyleAMathews: Yeah, I just copied the README sample code

DSchau: We should probably change the README 😅

PR #14510, opened 2019-06-03
fix(gatsby-dev-cli): add missing awaits and other various fixes

pieh:
there were couple issues in https://github.com/gatsbyjs/gatsby/pull/14347 did wasn't spotted and that caused some PRs to fail e2e/integration tests because of it

this adds some missing awaits, retry mechanism for copying files (seems like Promise.all copying hundreds of files cause weird ENOENT errors randomly)

EDIT: failing tests on this PR are example of those errors


wardpeet:
I'm hitting this issue on an old starter:
image

All ENOENT errors are gone so 👍

pieh:
There are some false warnings (X doesn't seem to be installed. restart gatsby-dev to publish it). Those don't really affect things, just are very confusing and just shouldn't be displayed in the middle of publishing dev version of package to verdaccio (will create separate PR for this)

PR #14511, opened 2019-06-03
Add yarn install command to readme

wiput1999:
## Description

Depend on yarn.lock in the repository I think we should have install command for yarn user in getting start guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,10 @@ You can get a new Gatsby site up and running on your local dev environment in 5 
 1. **Install the Gatsby CLI.**
 
    ```shell
+   # NPM users
    npm install -g gatsby-cli
+   # Yarn users
+   yarn global add gatsby-cli
 
    ```
 


pieh:
Hi @wiput1999! Thanks so much for opening a pull request!

We really appreciate you sending this over, but the change you’ve proposed is not going to be accepted because not adding `yarn` alternatives to our docs is decision we made to not clutter them with rationale that if users opted into using `yarn` as their package manager they would be able to translate `npm` commands. We would need to change this in 1000 different spots (tutorials, getting started, plugins documentation etc) and it's just not feasible to do so and having yarn instructions in some places but not other would be just confusing

We absolutely want to have you as a contributor, so please take a look at [our open issues][open-issues] for ideas, and please reach out to us [on Twitter at @gatsbyjs](https://twitter.com/gatsbyjs) with questions.

We offer pair programming sessions if you’d like to work with one of our maintainers to make another contribution. https://gatsby.dev/pairing

PR #14519, opened 2019-06-03
feat(themes): Add gatsbyesque colors

amberleyromo:
## Description

Changes blog theme to use a Gatsbyesque default color palette based on @fk's guidelines

## Related Issues

Related to #14276

### DIFF
--- a/themes/gatsby-theme-blog/src/theme/colors.js
+++ b/themes/gatsby-theme-blog/src/theme/colors.js
@@ -1,9 +1,8 @@
-const gatsby = `#663399`
-const gatsbyDarker = `#362066`
-const gatsbyBright = `#D9BAE8`
-const gray = `#48434F`
-const darkGray = `#232129`
-const almostBlack = `#1B1F23`
+const purple60 = `#663399`
+const purple30 = `#D9BAE8`
+const grey70 = `#48434F`
+const grey90 = `#232129`
+const black80 = `#1B1F23`
 const white = `#fff`
 const lightWhite = `rgba(255, 255, 255, 0.86)`
 const opaqueLightYellow = `rgba(255, 229, 100, 0.2)`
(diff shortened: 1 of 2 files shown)


amberleyromo:
I know something needs to change on the dark theme but @fk halp.

Screen Shot 2019-06-03 at 1 34 06 PM

Screen Shot 2019-06-03 at 1 34 15 PM

jxnblk:
This looks awesome!

themes/gatsby-theme-blog/src/theme/colors.js
```diff
@@ -1,19 +1,19 @@
-const teal = `#66b9bf`
-const darkTeal = `#078092`
-const darkBlue = `#282c35`
-const darkGray = `#222`
-const almostBlack = `#1a1a1a`
+const gatsby = `#663399`
+const gatsbyDarker = `#362066`
+const gatsbyBright = `#D9BAE8`
```
jxnblk: NB: I know @fk has some sort of number-based naming convention for .org. Even if we're only using a subset of the brand colors, it might make sense to use the same naming convention

amberleyromo: yeah i was looking at the draft brand guidelines for this actually! some of them are "named" named and some of them are numbered. down for whatever @fk thinks makes the most sense.

fk: TBH I don't have strong opinions on naming other than I think that most of these aliases

https://github.com/gatsbyjs/gatsby/blob/3c47ad1677952d14d9ac23463de3c9ab1c3a18a4/www/src/utils/tokens/colors.js#L119-L125

should go in favor of "role-based tokens" https://github.com/gatsbyjs/gatsby/blob/3c47ad1677952d14d9ac23463de3c9ab1c3a18a4/www/src/utils/tokens/colors.js#L143-L185

But of course having an alias for "Gatsby Purple" makes sense.

---

Sorry, bikeshedding. In regards to the names here I first wanted to say "I'm fine with the names as they are", but thinking about it again, @jxnblk is probably right in keeping things consistent — makes sense as a reference for ourselves. Then again we could solve the latter issue in so many ways. I feel like I'm not adding much here today anymore — gotta sleep, head hurts!

---

Oh @jxnblk — you know I'll take your veto re: `purple.60` any day, right? 😉
`blackFade[80]` already creeps me out: https://github.com/gatsbyjs/gatsby/blob/3c47ad1677952d14d9ac23463de3c9ab1c3a18a4/www/src/utils/tokens/colors.js#L127-L134

amberleyromo:
I do like that a lot better. ++ the purple60 changes to the highlight.

## Light

Screen Shot 2019-06-03 at 5 41 22 PM

## Dark

Screen Shot 2019-06-03 at 5 41 31 PM

PR #14520, opened 2019-06-03
feat(themes): Clean up blog theme config

jxnblk:
## Description

Removes some plugins from the blog theme that can be handled by the end-user's site:
- Google Analytics
- Netlify 
- Remark (this theme uses MDX for markdown)
- Manifest plugin
- Offline plugin

To do: 
- ~Fix heading styles – when moving the `gatsby-remark-autolink-headers` plugin to MDX,  the styles injected into the head do not appear to be working~

## Related Issues

Related to #14276


themes/gatsby-theme-blog/gatsby-config.js
```diff
@@ -95,12 +75,7 @@ module.exports = {
...
     },
     {
-      resolve: `gatsby-plugin-google-analytics`,
-      options: {
-        // trackingId: `ADD YOUR TRACKING ID HERE`,
-      },
-    },
-    {
+      // should this be included in the theme?
       resolve: `gatsby-plugin-manifest`,
```
jxnblk: Does having the manifest plugin make sense for a theme? I could see how it could make sense, but seems like the type of thing that might be better handled at the site level and in the starter that uses this theme

amberleyromo: i vote site level.

jxnblk:
It seems like `gatsby-remark-autolink-headers` [doesn't work super well with MDX](https://github.com/ChristopherBiscardi/gatsby-mdx/issues/204) at the moment. I think the combination of `remark-slug` and `remark-autolink-headings` with some custom styles in the theme might be a better option for the time being

themes/gatsby-theme-blog/gatsby-config.js
```diff
@@ -92,30 +70,26 @@ module.exports = {
...
         ],
+        remarkPlugins: [
+          require(`remark-slug`),
+          [
+            require(`remark-autolink-headings`),
+            {
+              content: {
+                type: `element`,
+                // custom MDX element - is this too magical??
+                tagName: `link-icon`,
```
jxnblk: By default `remark-autolink-headings` renders a `<span>` with a className. Here this is making a custom MDX element type, which is rendered by a custom MDX component that renders the Octicon link icon SVG. While this works, this seems a little too magical to me.

One alternative would be to use `remark-slug` and custom `h1` –`h6` MDX components that take `props.id` to render the link and icon – this would also open up other rendering options when shadowing the headings

amberleyromo: This strikes me as a bit too magical, partially because (and this comment may be borne of ignorance) but if the way the mdx plugin is handling it is different from the way it's handled in remark, isn't that something that should be fixed in the mdx plugin?

themes/gatsby-theme-blog/gatsby-config.js
KyleAMathews: I vote theme. Every site needs it and it's easy to drive from the theme by asking for a logo image and using e.g. the primary color. It's rarely something people touch after installing.

jxnblk: I can revert this change. One thing I'm wondering about is how multiple plugins like this would be merged, assuming the notes theme also has the manifest plugin

themes/gatsby-theme-blog/gatsby-config.js
jxnblk: MDX isn't doing anything non-standard here with the plugin, but I am with the `content` option here – and I definitely don't like this anymore

I think I'm leaning towards using custom MDX components for this sort of thing – when the markup generation happens in a Remark plugin, it's hidden away, whereas a React component would be easy to read and also shadow

themes/gatsby-theme-blog/gatsby-config.js
KyleAMathews: Hmm yeah, we'd need the same prop merging capabilities.

KyleAMathews: So yeah, I guess keeping this stuff at the site level does start to make sense 👍

KyleAMathews: Changed my mind haha 😂 I'm with y'all now!

themes/gatsby-theme-blog/src/components/headings.js
```diff
@@ -0,0 +1,54 @@
...
+    aria-hidden="true"
+  >
+    <path
+      fillRule="evenodd"
+      d="M4 9h1v1H4c-1.5 0-3-1.69-3-3.5S2.55 3 4 3h4c1.45 0 3 1.69 3 3.5 0 1.41-.91 2.72-2 3.25V8.59c.58-.45 1-1.27 1-2.09C10 5.22 8.98 4 8 4H4c-.98 0-2 1.22-2 2.5S3 9 4 9zm9-3h-1v1h1c1 0 2 1.22 2 2.5S13.98 12 13 12H9c-.98 0-2-1.22-2-2.5 0-.83.42-1.64 1-2.09V6.25c-1.09.53-2 1.84-2 3.25C6 11.31 7.55 13 9 13h4c1.45 0 3-1.69 3-3.5S14.5 6 13 6z"
+    />
+  </svg>
+)
+
+const heading = Tag => props => (
```
jxnblk: Instead of using the previous approach with `remark-autolink-headings`, this uses a custom MDX component to render a linked heading with SVG link icon on hover. The end result is essentially the same, but this would allow users to shadow the `components/headings.js` module to customize the rendered output

jxnblk:
Yeah, I think this is good for now. Are the failing tests anything I should be concerned about?

amberleyromo:
@jxnblk nope! checked with @DSchau, he's aware / knows whats up. all good.

PR #14523, opened 2019-06-03
feat(themes): Clean up blog theme default content

amberleyromo:
## Description

This cleans up some of the hard-coded defaults in the blog theme:
- Adds components to shadow for bio and footer content (rather than the mdx fragment pattern)
- Some random consistency refactors

## Related Issues

Related to #14276


themes/gatsby-theme-blog/src/components/bio-content.js
```diff
@@ -0,0 +1,14 @@
...
+ * Shadow me to add your own bio content
+ */
+
+export default () => (
+  <Fragment>
+    Words by <Styled.a href="http://example.com/">Jane Doe</Styled.a>.
+    <br />
+    Change me. This is all quite default.
+  </Fragment>
+)
```
jxnblk: I could go either way on this, but wanted to bring this question up for discussion. Do we want to follow this pattern of separating content from presentation in components like Bio and Footer (which means it's simpler to update the content while keeping styles in tact), or do we want to follow a pattern of having a single Bio component that accepts props and can be shadowed? Whichever we go with, we should probably make sure the notes theme follows a similar convention

amberleyromo: Open to thoughts. I did it this way because I think for the most minimal case (changing content) it's simplest to separate the two concerns.

amberleyromo: opinions @jlengstorf 	@ChristopherBiscardi 	?

jxnblk: Personally, I like this approach, and I don't think this should block the PR, but curious if anyone had thoughts on this

ChristopherBiscardi: no strong feelings either way. "bio-content" is a little weird, but this is really in addition to the other behavior that was mentioned so people can still shadow bio if they want.

Issue #14535, opened 2019-06-04
Why is there not a @types/gatsby package?


alexluong:
Hi,

I'm trying out TypeScript with Gatsby for the first time and quite surprised that there is no `@types/gatsby` package. I think it would be helpful, especially if someone want to implement the Node/Browser/SSR in TypeScript.

Some example code:

```ts
export function sourceNodes({ actions, createNodeId }: SourceNodesParam) {}
```

DSchau:
Types are available, and are distributed alongside the library.

https://unpkg.com/gatsby@2.8.2/index.d.ts

Hope this helps -- thanks!

alexluong:
Thanks for the response. This is exactly what I'm looking for.

alexluong:
@DSchau Sorry for bothering, but using types from Gatsby create a false error for me:

```ts
    createNode({
      ...obj,
      id: createNodeId(`Obj-${post.id}`),
      parent: "",
      children: [],
      internal: {
        type: "Obj",
        content: nodeContent,
        contentDigest: crypto
          .createHash("md5")
          .update(nodeContent)
          .digest("hex"),
      },
    })
```

It complains:

```
Type '{ type: string; content: string; contentDigest: string; }' is missing the following properties from type '{ type: string; mediaType: string; content: string; contentDigest: string; owner: string; description?: string | undefined; }': mediaType, owner
```

What I think is happening is here:

```
export interface Actions {
  ...

  createNode(node: Node, plugin?: ActionPlugin, options?: ActionOptions): void
}
```

It expect a Node as an argument where `mediaType` is not optional and `owner` field exists within `internal`.

Maybe there should be a `NodeInput` where `mediaType` is optional and there is no `owner` field?

DSchau:
@alexluong we'd love a PR making those types a little more flexible!

alexluong:
@DSchau Created a PR with a follow-up question. We can discuss over there! Thanks for the support.

muescha:
FYI: 
- https://www.gatsbyjs.org/plugins/?=typescript

PR #14537, opened 2019-06-04
fix(gatsby): Add NodeInput for createNode function types

alexluong:
## Description

Add `NodeInput` interface to be used by `createNode` function. With this, when writing Node API with TypeScript, users don't have to pass `internal.mediaType` and `internal.owner` to their object when creating new nodes.

Since we're already here, I'm wondering if we can also make `parent` and `children` field optional. I tried creating new nodes without those fields, and it didn't give me any errors.

## Related Issues

#14535


pieh:
> Since we're already here, I'm wondering if we can also make `parent` and `children` field optional. I tried creating new nodes without those fields, and it didn't give me any errors.

Yeah - this does make sense - minimal required things are `id`, `internal.type` and `internal.contentDigest`:

```
  actions.createNode({
    id: 'test',
    internal: {
      type: 'TestNode',
      contentDigest: 'foo'
    }
  })
}
```
Is valid.

I wonder if we can make `Node` to extend `NodeInput` to not redefine some of those common fields?

alexluong:
I'd change `NodeInput` to have only the minimal required fields and extend `Node` from it. Gotta wait until I get home to run prettier on it though. Thanks for the review.

alexluong:
So there is a failing test, but I'm not entirely sure what caused such regression. Any idea how I can fix it? Thanks.

@DSchau @pieh

wardpeet:
Sweet, thank you for fixing this one! I just restarted circle ci, seems like a flaky test.

I'm not 100% sure about this change as NodeInput seems a bit to loose now. What do you think of the following? I'm also pinging @orta as he is an expert on typescript ^^.

what this adds all possible props on nodeInput optional and makes the ones that are required on a Node mandatory. This helps us with autocompletion.

```ts
interface NodeInput {
  path?: string
  id: string
  parent?: string
  children?: Node[]
  fields?: Record<string, string>
  internal: {
    type: string
    mediaType?: string
    content?: string
    contentDigest: string
    owner?: string
    description?: string
  }
  resolve?: string
  name?: string
  version?: string
  pluginOptions?: PluginOptions
  nodeAPIs?: any[]
  browserAPIs?: any[]
  ssrAPIs?: any[]
  pluginFilepath?: string
  packageJson?: PackageJson
  siteMetadata?: Record<string, any>
  port?: string
  host?: string
  pathPrefix?: string
  polyfill?: boolean
  buildTime?: string
  jsonName?: string
  internalComponentName?: string
  matchPath?: unknown
  component?: string
  componentChunkName?: string
  context?: Record<string, any>
  pluginCreatorId?: string
  componentPath?: string
  [key: string]: unknown
}
export interface Node extends NodeInput {
  parent: string
  children: Node[]
  internal: NodeInput["internal"] & {
    mediaType: string
    content: string
    owner: string
  }
  [key: string]: unknown
}
```

pieh:
@wardpeet There are some fields that you can't set directly when creating nodes - like: `fields` or `internal.owner`. So what I'm thinking is something similar to what there was there initially (before second commit):

```ts
export interface NodeInput {
  id: string
  parent?: string
  children?: string[]
  internal: {
    type: string
    mediaType?: string
    content?: string
    contentDigest: string
    description?: string
  }
  [key: string]: unknown
}
```
and in Node just change `parent` and `children` to required and add `internal.owner` to required string.

I would seriously drop those properties from definition:
```
  resolve?: string
  name?: string
  version?: string
  pluginOptions?: PluginOptions
  nodeAPIs?: any[]
  browserAPIs?: any[]
  ssrAPIs?: any[]
  pluginFilepath?: string
  packageJson?: PackageJson
  siteMetadata?: Record<string, any>
  port?: string
  host?: string
  pathPrefix?: string
  polyfill?: boolean
  buildTime?: string
  jsonName?: string
  internalComponentName?: string
  matchPath?: unknown
  component?: string
  componentChunkName?: string
  context?: Record<string, any>
  pluginCreatorId?: string
  componentPath?: string
```
Those are very freeform and seems to be grabbed from some nodes that Gatsby internally produces, but it seems to be rare for user to actually ever use those?

alexluong:
Would this looks like a reasonable fix?

```ts

export interface NodeInput {
  id: string
  parent?: string
  children?: string[]
  internal: {
    type: string
    mediaType?: string
    content?: string
    contentDigest: string
    description?: string
  }
  [key: string]: unknown
}

export interface Node extends NodeInput {
  parent: string
  children: string[]
  internal: NodeInput["internal"] & {
    mediaType: string
    content: string
    owner: string
  }
  [key: string]: unknown
}

```

pieh:
@alexluong looks mostly great - just in `Node` - `mediaType` is still optional so I think we could drop that from Node and reuse what's in NodeInput. Also `content` is optional as well

alexluong:
Thanks @pieh for the review. I updated the PR. Let me know if there's anything else.

ryanditjia:
Hi, this broke one of my sites that uses `onCreatePage` API to delete pages, only to then recreate them for every locale.

I use `gatsby-node.ts` (that’s imported by root `gatsby-node.js`) to author my Gatsby Node APIs.

`gatsby-node.ts`
```ts
export const gatsbyNode: GatsbyNode = {
  onCreatePage: ({ page, actions }) => {
    const { createPage, deletePage, createRedirect } = actions

    // No need to recreate 404 page for every locale
    if (page.path.includes('404')) {  // TS error: Property 'includes' does not exist on type 'unknown'
      return
    }

    deletePage({
      component: page.component, // TS error: Type 'unknown' is not assignable to type 'string'
      path: page.path, // TS error: Type 'unknown' is not assignable to type 'string'
    })
 
    // recreate pages for every locale
    // ... rest of code
  }
}
```

Could we revert the removal of those properties? Especially `path` and `component`.

wardpeet:
I think we need to update https://github.com/gatsbyjs/gatsby/blob/bda7e6c7b87ba9db2f0b98b76eb545e9dfce88df/packages/gatsby/index.d.ts#L632 to have a `PageNode` which contains component & path.

ryanditjia:
I can submit a PR.

Will this be sufficient?

```diff
export interface Node extends NodeInput {
  parent: string
  children: string[]
  internal: NodeInput["internal"] & {
    owner: string
  }
+ path?: string
+ component?: string
  [key: string]: unknown
}
```

Or should we restore the rest of the properties too?

ryanditjia:
Logging one of the `page`s gets me the following object:

```json
{
  "page": {
    "internalComponentName": "ComponentIndex",
    "path": "/",
    "component": "/Users/ryanditjia/Workspace/neoreflexology.com/src/pages/index.tsx",
    "componentChunkName": "component---src-pages-index-tsx",
    "isCreatedByStatefulCreatePages": true,
    "context": {},
    "updatedAt": 1563525623803,
    "pluginCreator___NODE": "1013ede9-c58d-5483-b5f4-06465f558a78",
    "pluginCreatorId": "1013ede9-c58d-5483-b5f4-06465f558a78",
    "componentPath": "/Users/ryanditjia/Workspace/neoreflexology.com/src/pages/index.tsx"
  }
}
```

`context` property could be useful, I’ve seen it being used in [this plugin](https://github.com/gakimball/gatsby-plugin-global-context/blob/master/gatsby-node.js).

I think we should restore all of the properties, for thoroughness’ sake.

PR #14549, opened 2019-06-05
chore(showcase): Add the edit suite

Firthir:


KyleAMathews:
BTW, curious why there's the delay between switching pages each time? Looks like something you added deliberately?

Firthir:
> BTW, curious why there's the delay between switching pages each time? Looks like something you added deliberately?

Yeah we were having issues with full-page js which I won't be using again... it was the only fix that works across all browsers

PR #14560, opened 2019-06-05
showcase metatag images size fix [#14469]

### LINKED ISSUE #14469: [www/showcase] "Showcase" entry uses square image as meta `twitter:image`
E.g. for https://www.gatsbyjs.org/showcase/airbnb.io/ we are setting

as `twitter:image`.
Let's share the uncropped image instead:

abhijithvijayan:
## Description
Replaced the meta tag with the publicURL for the showcase images as the current images
 (eg: https://www.gatsbyjs.org/static/cbaf442ed29638ce3775856ede4ac483/d808c/8990f01f5a6d31fce109c16ab078b499.jpg) are cropped improperly.

## Related Issues
Fixes #14469

### DIFF
--- a/www/src/components/showcase-details.js
+++ b/www/src/components/showcase-details.js
@@ -123,6 +123,7 @@ const ShowcaseDetails = ({ parent, data, isModal, categories }) => (
               }
               childScreenshot {
                 screenshotFile {
+                  publicURL
                   childImageSharp {
                     resize(width: 200, height: 200) {
                       src
@@ -255,15 +256,13 @@ const ShowcaseDetails = ({ parent, data, isModal, categories }) => (
                 <meta
                   property="og:image"
                   content={`https://www.gatsbyjs.org${
-                    data.sitesYaml.childScreenshot.screenshotFile
-                      .childImageSharp.resize.src
+                    data.sitesYaml.childScreenshot.screenshotFile.publicURL
                   }`}
                 />
                 <meta
                   name="twitter:image"
                   content={`https://www.gatsbyjs.org${
-                    data.sitesYaml.childScreenshot.screenshotFile
-                      .childImageSharp.resize.src
+                    data.sitesYaml.childScreenshot.screenshotFile.publicURL
                   }`}
                 />
               </Helmet>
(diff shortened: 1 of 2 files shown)


abhijithvijayan:
The linting failed in windows.

I will try with Ubuntu

freiksenet:
CC @fk

www/src/components/showcase-details.js
```diff
@@ -255,15 +256,13 @@ const ShowcaseDetails = ({ parent, data, isModal, categories }) => (
                 <meta
                   property="og:image"
                   content={`https://www.gatsbyjs.org${
-                    data.sitesYaml.childScreenshot.screenshotFile
-                      .childImageSharp.resize.src
+                    data.sitesYaml.childScreenshot.screenshotFile.publicURL
```
fk: Hey @abhijithvijayan! Thanks for the PR! 🤗

I just gave this a spin, and `publicURL` here will always be `undefined`, unfortunately. :(

This is because `data.sitesYaml` is coming in from https://github.com/gatsbyjs/gatsby/blob/master/www/src/templates/template-showcase-details.js#L96-L130. You added `publicURL` to the `StaticQuery` that takes care of pulling in the screenshot thumbnails for the previous and next navigation shown (only) in the "showcase site" modal.

We were misleading you in https://github.com/gatsbyjs/gatsby/issues/14469#issuecomment-499029287—apologies for that!

Since I'm at it, I'll take the liberty and push a fix so we can get this merged!

If I was unclear in explaining why this doesn't work as you expected it, maybe my follow-up commits will help; if you have any questions, don't hesitate to ask! And please 🙏 don't let this discourage you from contributing! ❤️

fk:
This

- adds `og:title` meta information so that Twitter actually will generate a card when sharing a link to a site in the showcase (and expands the page `<title>` to match what we do for the "Starters" detail pages, e.g. https://www.gatsbyjs.org/starters/gatsbyjs/gatsby-starter-blog/)
- adjusts the image query to produce an image with a 1.91:1 aspect ratio instead of a square image
- now that we are serving an appropiate image, tells Twitter to generate a [`summary_large_image`](https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/summary-card-with-large-image.html) card

Before:
Bildschirmfoto 2019-06-10 um 23 44 02

👇 
After:
Bildschirmfoto 2019-06-10 um 23 26 42

www/src/components/showcase-details.js
abhijithvijayan: Thank You

PR #14562, opened 2019-06-05
fix/feat(gatsby-source-drupal): Copy relationship meta (if any) to a node's field

### LINKED ISSUE #10339: gatsby-source-drupal image alt tags

hanoii:
## Description

Drupal data structure allows for properties that are part of a relationship, and not on the relationship itself. A common example for this is an image field, which relates to a file entity, but each relationship has an "alt" field. (See #10339).

I believe as far as content goes, this meta information belongs not to the relationship, but to the node that uses the relationship.

So this means that if there's meta on the relationship, it will be copied to a field name of the same name, for an image field as an example, this now works:

```graphql
{
  allTaxonomyTermSite {
    edges {
      node {
        relationships {
          field_logo {
            localFile {
              publicURL
            }
          }
        }
        field_logo {
          alt
          title
          width
          height
        }
      }
    }
  }
}
```

Ideally, I like this to be part of the relationship data, but I think this it not possible currently with the inference or by changing too much on gatsby.

I.e., if you have 

A ----> Relates to ----> B

This is properly handled if A has data and B has data, but what happens when the "relationship" has data, which is what's happening in this place.

I might have expected that instead of doing `node.relationships[`${ref.type}___NODE`] = UUID`, I could have done something like:

```jsx
node.relationships[`${ref.type}___NODE`] = {
  id: UUID,
  foo: bar
}
```

And support both, I bet this is a lot of changing.

Maybe with this added and schema customization this can be improved, or maybe adding schema customization to the source plugin as well, but I haven't gotten there yet.

## Related Issues

Fixes #10339 
Another approach to #14187 
https://www.drupal.org/project/jsonapi/issues/2860886#comment-12040685

### DIFF
--- a/packages/gatsby-source-drupal/src/gatsby-node.js
+++ b/packages/gatsby-source-drupal/src/gatsby-node.js
@@ -179,8 +179,24 @@ exports.sourceNodes = async (
             node.relationships[`${k}___NODE`] = _.compact(
               v.data.map(data => (ids[data.id] ? createNodeId(data.id) : null))
             )
+            var meta = _.compact(
+              v.data.map(data => !_.isEmpty(data.meta) ? data.meta : null)
+            )
+            // If there's meta on the field and it's not an existing/internal one
+            // create a new node's field with that meta. It can't exist on both
+            // @see https://jsonapi.org/format/#document-resource-object-fields
+            if (!_.isEmpty(meta) && !(k in node)) {
+              node[k] = meta
+            }
+
           } else if (ids[v.data.id]) {
             node.relationships[`${k}___NODE`] = createNodeId(v.data.id)
+            // If there's meta on the field and it's not an existing/internal one
+            // create a new node's field with that meta. It can't exist on both
+            // @see https://jsonapi.org/format/#document-resource-object-fields
+            if (!_.isEmpty(v.data.meta) && !(k in node)) {
+              node[k] = v.data.meta
+            }
           }
         })
       }
--- a/packages/gatsby-source-drupal/src/gatsby-node.js
+++ b/packages/gatsby-source-drupal/src/gatsby-node.js
@@ -180,7 +180,7 @@ exports.sourceNodes = async (
               v.data.map(data => (ids[data.id] ? createNodeId(data.id) : null))
             )
             var meta = _.compact(
-              v.data.map(data => !_.isEmpty(data.meta) ? data.meta : null)
+              v.data.map(data => (!_.isEmpty(data.meta) ? data.meta : null))
             )
             // If there's meta on the field and it's not an existing/internal one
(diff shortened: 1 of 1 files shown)


bakeruk:
Will this work with nested entities from Drupal modules such as Paragraphs?

In my opinion, I find it odd to make duplicate references as well as separating the field data apart

hanoii:
> Will this work with nested entities from Drupal modules such as Paragraphs?

It does, this affects just Node A and its relationship with Node B (gatsby node, aka any entity on drupal). It doesn't really matter how depth the relationship goes:

```graphql
  allNodeSponsors {
    edges {
      node {
        relationships {
          field_content {
            ... on paragraph__affiliate {
              relationships {
                field_affiliate {
                  field_media_image {
                    alt
                    title
                    width
                    height
                  }
                  relationships {
                    field_media_image {
                      localFile {
                        publicURL
                      }
                    }
                  }
                }
	     }
            }
          }
        }
      }
    }
  }
```

> In my opinion, I find it odd to make duplicate references as well as separating the field data apart. Is this to get around the issue of having different `alt` and `title` tags on a node that uses the same image as another?

Yes, but also to keep it data consistent. alt/title is one specific case of meta in relationships with a file entity. But other things, like entity reference revisions can have meta too. I didn't want to touch each one on a case by case basis but rather something more fundamental.

freiksenet:
Thank you!

pascal-kubrick:
Thanks a lot!

DhillonN:
> {
  allTaxonomyTermSite {
    edges {
      node {
        relationships {
          field_logo {
            localFile {
              publicURL
            }
          }
        }
        field_logo {
          alt
          title
          width
          height
        }
      }
    }
  }
}

I cant't figure out how to manage case where field_logo is multi value field.

alessandrocalia:
> `{ allTaxonomyTermSite { edges { node { relationships { field_logo { localFile { publicURL } } } field_logo { alt title width height } } } } }`
> 
> I cant't figure out how to manage case where field_logo is multi value field.

+1

hanoii:
@DhillonN, @alessandrocalia 

In the case of multi_value field you should get an array of elements on both places, and array on field_logo as well as an array on the relationship, like any other multi-value field.

Is that not happening? What is the output of the GraphQL query (and the query) you are running?

PR #14572, opened 2019-06-05
Refactor gatsby-plugin-google-analytics gatsby-browser.js

BPScott:
## Description

The gatsby-browser.js file in gatsby-plugin-google-analytics and
gatsby-plugin-google-gtag have borderline identical functionality howeve
that is slightly obscured as their implementation is slightly different,
notably gatsby-plugin-google-analytics does not use the early return
pattern which means the nesting of everything is different if you try
and diff the two files.

This commit makes gatsby-plugin-google-analytics use the early return
pattern to make it applying the same changes to this an
gatsby-plugin-google-gtag easier.

Run `diff -u packages/gatsby-plugin-google-analytics/src/gatsby-browser.js packages/gatsby-plugin-google-gtag/src/gatsby-browser.js` to see that the contents is now nearly identical


freiksenet:
Thank you!

PR #14573, opened 2019-06-05
fix(docs): add labels to first relevant example

marcysutton:
A tiny PR to remove an accessibility antipattern from the Adding Forms page.


marcysutton:
I'm going to merge this, as it's really not worth holding up!

PR #14576, opened 2019-06-06
feat(gatsby-source-filesystem): add createFileNodeFromBuffer

### LINKED ISSUE #14538: Saving Buffer as file to utilize image processing
## Summary
How to save an image `Buffer` as file in order to utilize gatsby image processing?

If currently such use case is not available, which part of the code (core/plugin) that I can contribute to make it possible?

## Relevant information

- This use case arises from `gatsby-source-mysql`, whereby images could be retrived from database as `Buffer` object.
- It seems like `gatsby-source-filesystem` exports a `createRemoteFileNode` helper for file in remote. However, I could not find something like `createBufferFileNode`.

superhawk610:
## Description

In the same spirit as `createRemoteFileNode`, this PR adds a new named export to `gatsby-source-filesystem` intended for use with other source/transform plugins that have data stored in a buffer and wish to pass that data to other transforms that expect input in the form of a file node.

The API for `createFileNodeFromBuffer` is very similar to that of `createRemoteFileNode`, accepting a `Buffer` and optionally a precomputed content hash to skip duplicate work. This helper will write the contents of the buffer out to a file inside `gatsby-source-filesystem`'s disk cache and return a file node pointing to that cache file.

## Example

Here is a brief example of how this may be used within another source plugin:

```
// gatsby-node.js
const { createFileNodeFromBuffer } = require(`gatsby-source-filesystem`)

const { getBuffers } = require(`./internal`)

exports.sourceNodes = async ({ actions, createNodeId, store, cache }) => {

  const buffers = await getBuffers()

  for (let i = 0; i < buffers.length; i++) {
    const buffer = buffers[i]
    const fileNode = await createFileNodeFromBuffer({
      buffer,
      store,
      cache,
      createNode,
      createNodeId,
    })
    // optionally do something with `fileNode` (it's just a normal file node)
  }
}
```

## TODO

- [x] implement `createFileNodeFromBuffer`
- [x] write tests

## Related Issues

Closes #14538

### DIFF
--- a/packages/gatsby-source-filesystem/src/utils.js
+++ b/packages/gatsby-source-filesystem/src/utils.js
@@ -1,6 +1,5 @@
 const path = require(`path`)
 const Url = require(`url`)
-const crypto = require(`crypto`)
 
 /**
  * getParsedPath
--- a/packages/gatsby-source-filesystem/src/utils.js
+++ b/packages/gatsby-source-filesystem/src/utils.js
@@ -72,19 +72,3 @@ export function slash(path) {
 export function createFilePath(directory, filename, ext) {
   return path.join(directory, `${filename}${ext}`)
 }
-
-/**
- * md5Buffer
- * --
- * Hashes the contents of a buffer using MD5
- *
- *
- * @param {Buffer}            buffer
- * @return {String}           hash
- */
-export function md5Buffer(buffer) {
-  return crypto
-    .createHash(`md5`)
-    .update(buffer)
-    .digest(`hex`)
-}
--- a/packages/gatsby-source-filesystem/src/utils.js
+++ b/packages/gatsby-source-filesystem/src/utils.js
@@ -1,5 +1,6 @@
 const path = require(`path`)
 const Url = require(`url`)
+const crypto = require(`crypto`)
 
 /**
  * getParsedPath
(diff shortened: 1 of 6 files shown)


malcolm-kee:
Superb! The example fits my use case. I suppose the docs update would be tracked in separate PR? Would be glad to help on that after this is merged in and I could test on it.

superhawk610:
Yeah, we should add docs separately. Once this is merged I'll open a separate PR, or if you'd like to work on them you'll probably be better able to write more realistic examples since you've already been working on `gatsby-source-mysql`.

malcolm-kee:
@superhawk610 are u checking why the CI fail?
Anything I can help to move this forward?

superhawk610:
@malcolm-kee not sure why CI was failing but looks like it's good now! Since this makes changes to an existing package I have to wait for a review from a core team member, so I don't think there's much else we can do now but wait 👀 

Did you want to work on docs for `createFileNodeFromBuffer` or should I take them? If you want to give them a first pass and then I could review/edit them, that'd be cool.

malcolm-kee:
@superhawk610 can you take on the docs part?
I feel lost after I tried to re-implement what you did here into `gatsby-source-mysql` just to see how it works, and I no longer sure what's the best approach on using it 😭 . 
I would like to see what's your thought on `createFileNodeFromBuffer`.

packages/gatsby-source-filesystem/src/utils.js
```diff
@@ -58,3 +59,32 @@ export function slash(path) {
...
+/**
+ * md5Buffer
+ * --
+ * Hashes the contents of a buffer using MD5
+ *
+ *
+ * @param {Buffer}            buffer
+ * @return {String}           hash
+ */
+export function md5Buffer(buffer) {
```
wardpeet: Would you mind switching to our content-digest-helper located in gatsby package?
`const { createContentDigest } = require(`gatsby/utils`)`. It's basically the same thing what you wrote but to keep it consistent we would love that you use that one.

Thanks for understanding

wardpeet:
Publshed in `gatsby-source-filesystem:2.2.1`

PR #14596, opened 2019-06-06
feat(themes): add custom toggle switch component for blog theme

jxnblk:
## Description

This is to update the toggle switch component for turning dark mode on and off in themes. Ideally, this component is moved into theme-ui so that it can be shared across multiple themes, but I've included the component here for easier review.

~~The current implementation uses [react-toggle](https://github.com/aaronshaf/react-toggle) and a custom CSS file. The react-toggle component doesn't seem to have been updated recently and uses `UNSAFE_componentWillMount`, and I think we should use a more modern (and potentially smaller) approach that works well with the theme-ui context.~~

~~I've also looked at the [react-switch](https://github.com/markusenglund/react-switch) component, which seems to work and uses an `<input>` element for its implementation. Here, I've used a `<button>` component with `role="switch"`, `aria-checked`, and `aria-label`. This custom implementation does *not* currently handle touch events, but can add that if it makes sense to move forward with this approach.~~

I'm unsure what accessibility best practices are for a widget like this are, so my questions are:

- Does the `react-switch` library work well enough for our themes? There is a [live demo here](https://react-switch.netlify.com/) for viewing the rendered HTML.
- ~~Are there other components/libraries that I should look into?~~
- ~~If not (and if we decide to roll our own component), is the use of the `<button>` element here appropriate or should I change anything?~~

**Update** This now uses the `react-switch` component with some basic styling applied

## Related Issues

Related #14276


marcysutton:
I checked out the component in Safari and Chrome with Voiceover, and it seems to work fine; also in NVDA with Firefox. I can't quite tell if the demo you linked to is the right one though, as it uses a checkbox input and not a button element. Which markup should I be evaluating?

jxnblk:
@marcysutton That sounds good! There are two options I'm considering: `react-switch` or a custom implementation (with a `<button>` element). If the markup had a checkbox, that sounds like the official `react-switch` demo linked above. I'm unsure of the best way to deploy PRs for themes here, but the same custom implementation has been [deployed here](https://deploy-preview-45--theme-ui.netlify.com/switch/)

amberleyromo:
@jxnblk she was testing the `react-switch` demo.

If it's flexible enough, I think it'd be awesome to use `react-switch` instead of rolling our own. It looks like it was built with accessibility in mind (@marcysutton 	can 👍 / 👎 that when she has the chance to take another look).

marcysutton:
I agree that if we can reuse something and contribute fixes back to OSS, that's less for us to maintain! The react-switch demo worked in all major browser/AT combos except for JAWS and IE11 the first time I tried it, but then it seemed to fix itself just now. The custom version worked with all the major combos on the first try. I think either one would work, however. 

I should also note that I'm only testing with the latest AT's (assistive tech), but if we hear of problems we can always fix things later so I wouldn't let that hold everything up.

Here's some more info on accessible switches from Scott O'Hara, a11y wizard: https://scottaohara.github.io/aria-switch-control/

jxnblk:
Thanks for looking into this! I've *switched* the implementation in the PR here to use the `react-switch` library: system-ui/theme-ui#45, and it's deployed here: https://deploy-preview-45--theme-ui.netlify.com/switch/ – the styling needs a little work, and we can continue to iterate on how it integrates with the styling in Gatsby themes, but I think it works well enough for now. 

I based the custom implementation off of Scott's example, but I'll update this PR and remove the custom implementation in favor of using `react-switch`

amberleyromo:
@marcysutton this looks good to me -- do you have any blocking concerns?

marcysutton:
what's the latest link to review? this one? https://deploy-preview-45--theme-ui.netlify.com/switch/

jxnblk:
@marcysutton it's not the same *exact* code as what's used in this PR, but that link should be equivalent to what's rendered here in the blog theme and uses the same `react-switch` component with the same default props

PR #14599, opened 2019-06-06
Update instructions for installing XCode CLI

arafalov:
## Description
Apparently the xcode-select may no longer work. So, added alternative instructions that I tested myself.


freiksenet:
Thanks a lot!

PR #14600, opened 2019-06-06
fix(www): provide timezone to date to fix dates that dont match up

gillkyle:
## Description

Noticed dates on the blog preview cards differ from the dates displayed in the post:

image
image

This addresses that:

image

image


freiksenet:
Cheers!

PR #14604, opened 2019-06-07
fix(gatsby-source-shopify): fix pages query path

stevenwu:
## Description

Shopify page nodes are empty because the fetch results aren't using the `shop` key. I'm unsure how you want to handle this, so I duplicated `createNodes` and removed `shop` from the query path.

`PageNode` also needs to be updated to prevent `TypeError: nodeFactory(...) is not a function`.

### DIFF
--- a/packages/gatsby-source-shopify/src/gatsby-node.js
+++ b/packages/gatsby-source-shopify/src/gatsby-node.js
@@ -144,7 +144,7 @@ const createPageNodes = async (
   endpoint,
   query,
   nodeFactory,
-  { client, createNode, formatMsg, verbose, imageArgs, paginationSize },
+  { client, createNode, formatMsg, verbose, paginationSize },
   f = async () => {}
 ) => {
   // Message printed when fetching is complete.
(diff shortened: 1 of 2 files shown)


packages/gatsby-source-shopify/src/nodes.js
```diff
@@ -146,4 +146,4 @@ export const ProductVariantNode = imageArgs =>
 
 export const ShopPolicyNode = createNodeFactory(SHOP_POLICY)
 
-export const PagesNode = createNodeFactory(PAGE)
+export const PageNode = _imageArgs => createNodeFactory(PAGE)
```
pieh: is `_imageArgs` needed here if this has no effect?

packages/gatsby-source-shopify/src/gatsby-node.js
```diff
@@ -139,3 +139,25 @@ const createShopPolicies = async ({
...
+  f = async () => {}
+) => {
+  // Message printed when fetching is complete.
+  const msg = formatMsg(`fetched and processed ${endpoint}`)
+
+  if (verbose) console.time(msg)
+  await forEach(
+    await queryAll(client, [endpoint], query, paginationSize),
+    async entity => {
+      const node = await nodeFactory(imageArgs)(entity)
```
pieh: given that imageArgs has no effect, we could drop this?

PR #14608, opened 2019-06-07
feat(gatsby-plugin-preload-fonts): preload used fonts

### LINKED ISSUE #14281: Preload fonts used on pages
Fonts don't start downloading until the browser knows they're needed. Which means that the browser has to download/parse the HTML & CSS and start rendering a section of the page before it knows to start downloading the font.

This can cause significant delays in loading fonts.

A common suggestion is to add a `<link preload>` to the page for each font that's used on the page so the browser knows to start downloading the font files immediately.

But that raises the question of how would you know which files are used. Generally right now people manually note which fonts are used on specific pages and manually add preloads for each one.

This sort of manual bookkeeping is against everything Gatsby stands for! So how do we automate this? We can't (cheaply) know at build time which fonts are used on every page.

One idea I just prototyped is that we could use Puppeter to visit each page of the site and write out a file listing for each page, what font files are downloaded. Then at build-time, a plugin could add the `<link preload>` for these files.

The plugin would include a script to build/serve/visit all the pages and write out a file with the font file data. This file would need checked in and the script would need periodically re-run (or automate it e.g. nightly in a CI).

Someone want to write this :-D? This would be a great core plugin to add.

The puppeter script would work roughly like this:

```js
const puppeteer = require("puppeteer")
const path = require(`path`)

puppeteer.launch().then(async browser => {
  const page = await browser.newPage()
  page.on(`requestfinished`, e => {
    if (e._method === `GET` && path.parse(e._url).ext === `.woff2`) {
      console.log(e._url)
    }
  })
  await page.goto("https://www.gatsbyjs.org")
  await browser.close()
})
```

/cc @brimtown

superhawk610:
## Description

> Fonts don't start downloading until the browser knows they're needed. Which means that the browser has to download/parse the HTML & CSS and start rendering a section of the page before it knows to start downloading the font.
>
> This can cause significant delays in loading fonts.
>
> A common suggestion is to add a <link preload> to the page for each font that's used on the page so the browser knows to start downloading the font files immediately.

This plugin enables a two-step process for embedding `<link preload>` tags into each statically rendered page in your build output:

1. Run the included puppeteer script to scrape every route in your application and generate a mapping from routes to fonts.
2. Add the plugin to your `gatsby-config.js` and these tags will be automagically embedded into each page at build time.

## TODO

- [x] list all application routes
- [x] determine what fonts are required on each route
- [x] cache resource mapping between builds to prevent unnecessarily scrapes
- [x] inject tags at build time
- [x] configuration
- [x] docs
- [x] tests

## Related Issues

Closes #14281.


superhawk610:
Should the injected `<link>` tags have `cross-origin="anonymous"`?

superhawk610:
I'm currently working on unit tests but I figured I could go ahead and get some feedback on the plugin layout/implementation - in particular, I'm curious how my approach to the following tasks fit the Gatsby ecosystem:

- cache directory
- logging
- route fetching

thomaswangio:
Subscribed to this! Thanks for taking this on @superhawk610

packages/gatsby-plugin-preload-fonts/src/prepare/index.js
```diff
@@ -0,0 +1,114 @@
...
+
+  const browser = await puppeteer.launch()
+  const page = await browser.newPage()
+  let last = ``
+  page.on(`requestfinished`, e => {
+    if (!e._url.match(/(socket\.io|commons\.js)/)) {
+      logger.info(`load`, e._method, e._url)
+    }
+
+    if (
```
wardpeet: Checkign resourcetype is probably more robust than checking urls as we rely on chrome's internal api to see if something is a font.
```suggestion
if (request.resourceType() === 'font') {
```
https://pptr.dev/#?product=Puppeteer&version=v1.10.0&show=api-requestresourcetype

superhawk610:
@wardpeet I went with a separate script to follow the idea proposed in the original linked issue; I'm cool with either integrating with the dev server directly or having that as an option, my only concern is that this is a costly operation (minutes on smaller projects and potentially hours on larger projects). I added some basic route caching to prevent crawling if the list of routes hasn't changed, perhaps I should make that more granular and only crawl new routes by default?

cc @KyleAMathews: what are your thoughts on having the font-scraping script run onCreateDevServer vs manually via a package script?

packages/gatsby-plugin-preload-fonts/src/prepare/index.js
superhawk610: Nice. I didn't realize the other request properties had nice getter methods (`req.url()`, `req.method()`) so I think I'll switch to using those as well.

packages/gatsby-plugin-preload-fonts/README.md
```diff
@@ -0,0 +1,106 @@
...
+If an asset isn't showing up, this is likely due to its route not being scraped
+properly. To insure that the route in question is being scraped, you have a couple
+options. First, routes that don't show up under the `allSitePage` query on your
+site's GraphQL server will not be scraped.
+
+If you're still having trouble, you can run `gatsby-preload-fonts` with a lower
+log level to view what paths it's visiting in real time.
+
+```
+export LOG_LEVEL=info; npm run preload-fonts
```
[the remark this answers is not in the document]
superhawk610: `debug` seems to be more aimed at showing/hiding logs based on what package they originate from, while I need a level-based logger that hides messages below the configured level. Do you have any other package suggestions that are better suited for log levels?

I'll update the example with a Windows example as well 👍

KyleAMathews:
Yeah — my guess that this would be a really costly operation so best done in a periodic cron job — it looks like your testing has confirmed that so let's keep it a separate script.

packages/gatsby-plugin-preload-fonts/README.md
```diff
@@ -0,0 +1,110 @@
...
+waiting for them to resolve indirectly. In order to do this, `<link rel="preload">`
+tags will need to be added for each font asset required on each route at build time.
+
+To accomplish this, this plugin runs in two steps:
+
+1. Scrape each application route to determine what font assets are ultimately required.
+2. Inject `<link rel="preload">` tags for each asset at build time.
+
+The included `gatsby-preload-fonts` script spins up a [puppeteer](https://github.com/GoogleChrome/puppeteer) instance and visits each of your application's
+routes to determine the required font assets. Once it has done so, it caches
```
KyleAMathews: How is the caching done? My assumption is the script would write out a file which would get checked in via Git.

KyleAMathews: This would let someone run this in e.g. a CI server (which we should include instructions for)

superhawk610: The generated cache file `font-preload-cache.json` will be stored in the project root. In order to run this during CI, the user should be able to simply add `npm run preload-fonts`. Is that worth adding to the readme?

superhawk610:
Here's a couple screen recordings of the preload script:

## Initial Run

first-run

## Subsequent Run(s)

subsequent-run

packages/gatsby-plugin-preload-fonts/README.md
KyleAMathews: yeah! That'd be great. A section recommending people set this up in CI would be helpful. Thanks!

superhawk610:
Hey, sorry about the delay, I put this on the backburner until #15475 was addressed but it never was. I'll try and get a commit addressing @wardpeet's most recent feedback this week, hopefully the versioning thing will have sorted itself out.

wardpeet:
Oh yeah, it got fixed. I missed your issue. It would be awesome to get this one shipped 🤗.

packages/gatsby-plugin-preload-fonts/src/prepare/index.js
```diff
@@ -0,0 +1,110 @@
...
+
+    logger.info(`load`, req.method(), req.url())
+
+    if (req.method() === `GET` && req.resourceType() === `font`) {
+      logger.debug(`match`, req.url())
+
+      const { pathname } = new URL(page.url())
+
+      if (!Object.prototype.hasOwnProperty.call(cache.assets, pathname)) {
+        cache.assets[pathname] = {}
```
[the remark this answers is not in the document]
superhawk610: I like the idea of maps/sets, but the V8 serialization API is still marked experimental at time of writing, and I think `Object.create(null)` will still have non-trivial performance gains.

packages/gatsby-plugin-preload-fonts/src/prepare/fetch-routes.js
```diff
@@ -0,0 +1,57 @@
...
+
+      - ${bold(`HOST`)}          ${dim(`(default: localhost)`)}
+      - ${bold(`PORT`)}          ${dim(`(default: 8000)`)}
+      - ${bold(`LOG_LEVEL`)}     ${dim(`(default: error)`)}
+      - ${bold(`GRAPHQL_PATH`)}  ${dim(`(default: /___graphql)`)}
+
+`)
+  }
+
+  const hash = crypto.createHash(`md5`)
```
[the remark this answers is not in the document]
superhawk610: This loop updates the hash iteratively with each route's path, waiting to digest until the end, which I think is more idiomatic than something like appending to a string and then passing that to `createContentDigest`.

packages/gatsby-plugin-preload-fonts/src/prepare/cache.js
```diff
@@ -0,0 +1,50 @@
+const fs = require(`fs-extra`)
+const path = require(`path`)
+
+module.exports.getPath = getPath
+module.exports.load = load
+module.exports.save = save
+
+function getPath() {
```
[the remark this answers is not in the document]
superhawk610: Guess I was in a FP mood :upside_down_face:

packages/gatsby-plugin-preload-fonts/src/__tests__/index.test.js
```diff
@@ -0,0 +1,94 @@
...
+      [`/baz`]: [
+        createMockRequest(`/another/font.woff2`),
+        createMockRequest(`/some/external/font.ttf`),
+        createMockRequest(`/another/font.ttf`),
+      ],
+    })
+
+    await main()
+
+    expect(save.mock.calls[0][0]).toMatchSnapshot({
```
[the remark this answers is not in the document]
superhawk610: `save` is a mocked function, I think a full snapshot would be very noisy, and snapshotting specifically the first argument of the first call simplifies `expect.any(Number)` for `timestamp`.

superhawk610:
@wardpeet Should be g2g :fire:

wardpeet:
Awesome! :fire: I'll fix tests and give this plugin one more try and i'll merge afterwards! Thanks for your hard work. This is going to be awesome!

wardpeet:
@superhawk610 we have published your plugin `gatsby-plugin-preload-fonts@1.0.0`

universse:
Is this recommended for self-hosted fonts?

sidharthachatterjee:
@universse Yeah, should be beneficial whether self hosted or not

universse:
Cause I was trying to run the script ```"preload-fonts": "gatsby-preload-fonts"``` and got error
```
err could not establish a connection with the dev server
     attempted connection to http://localhost:8000/___graphql
```
So I started dev server, and run the scripts, which gives me this output
```javascript
// font-preload-cache.json
{
  ...
  "assets": {
    "/": {
      "http://localhost:8000/static/Spectral-ExtraBold-76dd69dbe4b4ddc4f3923bf08e883d43.woff2": true,
      ....
    },
  }
}
```
This works if fonts are loaded from external sources since you get the full external url. That's not the case for self-hosted ones with relative urls. So a way to configure the origin is needed. Not sure if i'm missing anything though.

wardpeet:
@universse good catch, you could add a PR to actually check the url if it matches the dev server we strip out the domain.

universse:
I made a PR https://github.com/gatsbyjs/gatsby/pull/17235

chocobuckle:
Maybe I'm missing something simple, but just curious as to how one is meant to use this with continuous deployment? If I run gatsby develop and then gatsby build locally then all works as expected, but when I push a commit and Netlify starts building my site then it obviously errors with a "err could not establish a connection with the dev server" message. Please explain, as if to an idiot! :)

superhawk610:
You'll need to tweak your CD config to run the dev server in the background, then run `gatsby-preload-fonts`, then kill the dev server once that's complete. I'm not totally sure of the best way to accomplish this on Netlify, perhaps @masives can shed some light?

nandorojo:
@superhawk610 Similar question to @chocobuckle here. Any idea how to get this to work on Gatsby Cloud when I deploy?

superhawk610:
@nandorojo I'm not super familiar with Gatsby Cloud configuration, but from [the docs](https://www.gatsbyjs.com/cloud/docs/cloud-configuration#can-i-use-a-custom-build-script) it seems like the best way to accomplish this would be to add a `postinstall` script that does what I suggested above, something like

```
// package.json
"scripts": {
  "postinstall": "scripts/update-font-preloads.sh"
}

// scripts/update-font-preloads.sh
#!/bin/bash
export CI=1 # Gatsby Cloud may set this for you, in which case
            # it's unnecessary; just make sure it's set or the
            # crawler will hang indefiniely waiting for input

echo "spinning up dev server..."
./node_modules/.bin/gatsby develop &
SERVER_PID=$!

echo "crawling routes to check for font preloads..."
./node_modules/.bin/gatsby-preload-fonts &
CRAWLER_PID=$!

wait $CRAWLER_PID
echo "done!"

kill -9 $SERVER_PID
```

(I haven't tested this but that's the basic idea)

botteu:
I'm in the same boat as @chocobuckle . We're hosting on Netlify and the build script is run when we push to branch X, but since the font mapping needs `gatsby develop` to run, it doesn't work.

masives:
@botteu I think you'd need to write a node script that would 
1. Spawn a process with gatsby develop
2. Wait for gatsby dev server to be available
3. Run the font preloading script
4. Kill the server

botteu:
@masives Yep. But I was hoping to avoid that 😉 But thanks anyways 👍

PR #14610, opened 2019-06-07
feat(gatsby): Allow adding custom field extensions

stefanprobst:
Continued from #13738

This PR allows adding custom field extensions with the `createFieldExtension` action.

It also introduced a new Node API `createSchemaCustomization` where all schema-related actions (`createTypes`, `createFieldExtension`, `addThirdPartySchema`) are made available, in addition to `sourceNodes`. This was [briefly discussed here](https://github.com/gatsbyjs/gatsby/pull/13738#discussion_r279738433).

Changes compared to #13738:
* extension registry moved into redux
* added type checking for extension args
* handles `defaultValue` for both directives and extensions
* added tests
* `process` renamed to `extend`

Open question:
* which field extension names should we reserve for internal use? should we prefix? (see @wardpeet [comment](https://github.com/gatsbyjs/gatsby/pull/13738#discussion_r280699333))


packages/gatsby/src/redux/actions/restricted.js
```diff
@@ -190,6 +190,77 @@ actions.createTypes = (
...
+ *         type: 'String',
+ *         args: {
+ *           sunshine: {
+ *             type: 'Int',
+ *             defaultValue: 0,
+ *           },
+ *         },
+ *         resolve(source, args, context, info) {
+ *           const motivation = (options.caffeine || 0) - args.sunshine
+ *           if (motivation > 5) return 'Work! Work! Work!'
```
freiksenet: :rofl:

packages/gatsby/src/redux/actions/restricted.js
```diff
@@ -258,12 +329,15 @@ const mapAvailableActionsToAPIs = restrictions => {
 }
 
 const availableActionsByAPI = mapAvailableActionsToAPIs({
+  createFieldExtension: {
+    [ALLOWED_IN]: [`sourceNodes`, `createSchemaCustomization`],
+  },
   createTypes: {
-    [ALLOWED_IN]: [`sourceNodes`],
+    [ALLOWED_IN]: [`sourceNodes`, `createSchemaCustomization`],
```
freiksenet: Should we make all those deprecated in `sourceNodes`?

freiksenet:
This is great!

Not 100% sure about namespacing. We don't really namespace eg types, so people can run into issue there. Ping @KyleAMathews @pieh for opinions on that.

Let's also figure out more use cases for this that we can add to a) guide b) an example (eg in using-schema-customization) c) a blog post announcing this.

packages/gatsby/src/redux/actions/restricted.js
stefanprobst: I kept `sourceNodes` as well because @pieh [comment here](https://github.com/gatsbyjs/gatsby/pull/14283#issuecomment-495664067)

pieh:
> Not 100% sure about namespacing. We don't really namespace eg types, so people can run into issue there. Ping @KyleAMathews @pieh for opinions on that.

It's bit hard to tell if we need to handle namespacing here. Type names are free-for-all and we didn't had any issues with that yet, so I would say to not worry about this now (other than making sure there are warning/errors if plugin tries to overwrite extension "owned" by some other plugin). And if this proves to be a problem we can act on it.

PR #14612, opened 2019-06-07
feat: allow inclusion/exclusion of fields in rich text-referenced nodes

### LINKED ISSUE #11364: gatsby-source-contentful: Maximum call stack size exceeded
Not sure if this is a bug or not. And unfortunately I don't really have much time this week to investigate. But I thought I might as well put this up here.

I'm using the `gatsby-source-contentful` plugin. In my contentful content model I'm having two content types, both with the new Richtext editor. Problems occurs when I have entry links to each other.

To be more clear. Say I have two pages, home and about. If I link from home -> about and from about -> home inside the Richtext editor I get this error.

Like I said before. Not sure if this is a bug.

Here's a callstack https://pastebin.com/fD676e6j

### Environment (if relevant)
```
  System:
    OS: macOS 10.14.3
    CPU: (4) x64 Intel(R) Core(TM) i5-8210Y CPU @ 1.60GHz
    Shell: 3.0.0 - /usr/local/bin/fish
  Binaries:
    Node: 10.15.0 - /var/folders/vh/cx86q1vs2xz5rxlf8b8cny2r0000gn/T/yarn--1548706327929-0.8647198662198463/node
    Yarn: 1.13.0 - /var/folders/vh/cx86q1vs2xz5rxlf8b8cny2r0000gn/T/yarn--1548706327929-0.8647198662198463/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v10.15.0/bin/npm
  Languages:
  Browsers:
    Chrome: 71.0.3578.98
    Safari: 12.0.3
  npmPackages:
    gatsby: 2.0.98 => 2.0.98 
    gatsby-plugin-layout: 1.0.11 => 1.0.11 
    gatsby-plugin-offline: 2.0.21 => 2.0.21 
    gatsby-plugin-react-helmet: 3.0.5 => 3.0.5 
    gatsby-plugin-styled-components: 3.0.4 => 3.0.4 
    gatsby-source-contentful: 2.0.26 => 2.0.26 
```

`gatsby-config.js`:
```
plugins: [
        {
            resolve: 'gatsby-plugin-styled-components',
            options: {
                displayName: false
            }
        },
        'gatsby-plugin-react-helmet',
        {
            resolve: 'gatsby-plugin-layout',
            options: {
                component: require.resolve('./src/components/Layout')
            }
        },
        {
            resolve: 'gatsby-source-contentful',
            options: {
                spaceId: '<spaceID>',
                accessToken: process.env.CONTENTFUL_ACCESS_TOKEN
            }
        },
        {
            resolve: '@contentful/gatsby-transformer-contentful-richtext',
            options: contentfulRenderOptions
        },
        'gatsby-plugin-offline'
    ]
```

jessepinho:
## Description
gatsby-source-contentful had a bug in which, if two entries have rich text fields, and each rich text field references the other entry, it got stuck in an infinite loop during the Gatsby bootup process.

This PR uses @simoneb's [approach](https://github.com/gatsbyjs/gatsby/issues/11364#issuecomment-462038897) and adds a `richText` plugin option to gatsby-source-contentful, which at the moment has two possible keys, `includeEntryFields` and `excludeEntryFields`. When using these options, you can control which fields to include/exclude on referenced entries in rich text, so that, e.g., you can avoid the infinite loop by excluding a referenced entry's rich text field.

Note that, as this change will not affected referenced entries unless the new plugin options are used, this is _not_ a breaking change.

## Related Issues
Fixes #11364


packages/gatsby-source-contentful/package.json
```diff
@@ -18,7 +18,7 @@
     "gatsby-plugin-sharp": "^2.1.3",
     "gatsby-source-filesystem": "^2.0.38",
     "is-online": "^7.0.0",
-    "joi": "^14.0.0",
+    "joi": "^14.3.1",
```
jessepinho: It was necessary to upgrade `joi` for the sake of being able to use Joi's `oxor`

jessepinho:
tagging @Khaledgarbaya for review!

jessepinho:
Hm, looks like flakey builds may be the reason for the CI failure.

pieh:
I'm not sure if this is best fix here. According to https://pastebin.com/fD676e6j from original issue - the infinite loop happens in gatsby core and not plugin.

We could add some mechanism to gatsby core to not crash by limiting maximum depth of recursion (we might need to do that in multiple places)

jessepinho:
@pieh ah, good point — but wouldn't the correct fix in core be that, instead of just limiting the max stack size, we reference existing nodes rather than creating new ones and making an infinite loop?

Meanwhile though, this bug is blocking for use of gatsby-source-contentful, so I'd love to get this merged if only temporarily. What do you think?

jessepinho:
@pieh hm — one more thought: maybe I could just change the documentation to clarify that the field filters can be used to make Gatsby bootstrapping faster (like it does with the `localeFilter` option). Thoughts? That way, this option could stay, even if the root Gatsby bug is fixed.

jessepinho:
(Also curious to hear @Khaledgarbaya 's thoughts on my last comment)

pieh:
> @pieh ah, good point — but wouldn't the correct fix in core be that, instead of just limiting the max stack size, we reference existing nodes rather than creating new ones and making an infinite loop?

The stack trace I linked crashes in part of code that recursively traverse single node object - (we don't ever check other nodes or create new ones there) so it seems like just Contentful SDK is creating circular structure in those cases and when gatsby tries to go traverse it, it get stuck in infinite loop. Maybe instead of checking depth we could check if we already traversed some object (or do both just to be safe)

jessepinho:
Yup, that does seem like the correct solution. Unfortunately, I'm not familiar enough with the Gatsby core codebase to know where to even begin to look for how to solve that. Do you know who would be a good person to talk to about it? And meanwhile, what do you think of my comment above — documenting the rich-text field filter as a performance booster?

jessepinho:
@pieh what do you think of my thoughts in [this comment](https://github.com/gatsbyjs/gatsby/pull/14612#issuecomment-500760873) above? That way we can merge this :)

kompounder:
Hi - is there a solution to this issue? It seems to severely limit Contentful's functionality with Gatsby. Any update would be much appreciated!

jessepinho:
@kompounder in my opinion, the best long-term solution is to follow the official Contentful GraphQL spec for rich text: https://www.contentful.com/developers/docs/references/graphql/#/reference/schema-generation/rich-text

This way, you could select the specific fields you want from embedded entries/assets _via a GraphQL query_, rather than in `gatsby-config.js`. 

Making the rich-text API work this way would take a bit of work and be a relatively hefty PR, but I think it's worth doing. @Khaledgarbaya maybe we could pair on this? 🤔

kompounder:
Thank you @jessepinho. I had considered using Contentful's Graphql API as well as you suggested, however it's available for large spaces only ($879/mo and above) at the moment, which makes sense in production in the long run but difficult to develop for.

jessepinho:
@kompounder right — I was just arguing for making the plugin mimic the way the real Contentful GraphQL API works.

packages/gatsby-source-contentful/README.md
```diff
@@ -152,6 +152,14 @@ For example, to filter locales on only germany `localeFilter: locale => locale.c
 
 List of locales and their codes can be found in Contentful app -> Settings -> Locales
 
+**richText.includeEntryFields** and **richText.excludeEntryFields** [array][optional] [default: `undefined`]
+
+If you have entries that reference each other via their rich-text fields, Gatsby may get stuck in an infinite loop during build and throw a `Maximum call stack size exceeded` error. To prevent this, you can filter which fields should be included in entries referenced by the field. Set a list of field names to include or exclude using this option, and Gatsby will filter the fields of referenced entries accordingly.
```
pvdz: I landed a fix that should address these circular references. I think this description can be updated to reflect that.

With that in mind, is this feature still necessary? Useful?

jessepinho: Thanks @pvdz! We have a ticket on our end to upgrade Gatsby to see if it fixes these issues. If it does, we'll close this PR. This should all happen in the next few weeks. Thanks again! Excited to try the fix :)

pvdz:
@jessepinho friendly ping for this PR. It's getting a bit stale.

jessepinho:
Hey @pvdz — we did upgrade Gatsby and it did not solve the problem for us. That said, I'm no longer working on the Gatsby site that I made this PR for, so perhaps we should close it. (Unless @yomete wants to pick it up.)

pvdz:
@yomete can you let us know, please? Otherwise I'll go ahead and close the PR.

yomete:
Hey @pvdz - Yes, we'll pick this back up from where Jesse left off and let you know how it goes :)

pvdz:
@yomete That's super! :purple_heart: In that case I'm going to put this branch in Draft mode. Please feel free to update (or whatever) it and put it back in review mode, or a new PR, once you're ready. Thank you!

bsgreenb:
Wasn't this already fixed by https://github.com/gatsbyjs/gatsby/pull/20674 ?

Do other people still have recursion issues with new Gatsby versions?

LekoArts:
Feel free to reopen this @yomete when you'll work on this, in the meantime I'll close this as it's stale.

PR #14616, opened 2019-06-07
feat(gatsby-source-contentful): Resolve locales in fields of entries referenced by rich-text fields

jessepinho:
## Description

When rich-text fields reference an entry or asset, the fields of the entry or asset don't have their locales resolved like the rest of the GraphQL response.

For example, consider the following GraphQL query:

```graphql
{
  contentfulArticle(slug: { eq: "my-article" }, node_locale: { eq: "en-US" }) {
    title
    bodyRichText {
      json
    }
  }
}
```

The response will look like this: 
```javascript
{
  contentfulArticle: {
    title: 'My article',
    bodyRichText: {
      json: {
        content: [{ nodeType: 'embedded-entry-block', data: { target: { fields: { title: { 'en-US': 'My other article', 'de-DE': 'Meine andere Artikel' } } } } }],
      } 
    }
  }
```

Note that, while the title of the outer article has a resolved locale — that is, the `title` property is simply the title string in the requested locale — the title of the inner article does _not_ have a resolved locale. Instead, the `title` property is an object containing `en-US` and `de-DE` properties. This makes for an inconsistent experience when working with the GraphQL response.

This PR fixes this issue by walking through the rich-text JSON tree and resolving locales of referenced entries. It also works recursively: if a referenced entry has its own reference to a third entry, it resolves the locales for the third (and fourth, and...) entry as well.  It does the same for assets (although not recursively, since assets don't have reference fields).

Note that this PR includes a major version bump, as it includes a breaking change for the content of the rich-text JSON.

cc @Khaledgarbaya


packages/gatsby-source-contentful/src/__tests__/normalize.js
```diff
@@ -55,7 +55,7 @@ describe(`Process contentful data`, () => {
     const createNodeId = jest.fn()
     createNodeId.mockReturnValue(`uuid-from-gatsby`)
     contentTypeItems.forEach((contentTypeItem, i) => {
-      normalize.createContentTypeNodes({
+      normalize.createNodesForContentType({
```
jessepinho: I felt that this name better explained what it's doing — i.e., it's not _just_ creating the content type node itself (e.g., `Article`), but also the entries in the content type.

packages/gatsby-source-contentful/package.json
```diff
@@ -1,13 +1,14 @@
 {
   "name": "gatsby-source-contentful",
   "description": "Gatsby source plugin for building websites using the Contentful CMS as a data source",
-  "version": "2.0.67",
+  "version": "3.0.0",
```
pieh: We are bumping version as part of publishing process, so please revert version bump here. I marked PR with breaking change label to keep track of it

jessepinho: @pieh fixed — thanks!

pieh:
Is this change compatible with https://github.com/contentful/rich-text/tree/master/packages/rich-text-react-renderer ?

jessepinho:
@pieh I haven't specifically tested these two together, but I can't see why not. The rich-text renderer simply walks to the node tree and passes each node to a renderer, and I haven't changed the _node_ tree at all. I've only changed the value of entry references that exist inside the node tree, which the rich-text renderer will take as-is and pass to the renderers that one provides in its config.

I feel like this is hard to explain by text 🙈 but does that make sense?

jessepinho:
(Thanks for taking a look at this btw! Excited to get it merged!)

packages/gatsby-source-contentful/src/__tests__/normalize.js
pieh: Yeah, thanks for that - I was confused quite few times by the name of the function in the past ;)

pieh:
> @pieh I haven't specifically tested these two together, but I can't see why not. The rich-text renderer simply walks to the node tree and passes each node to a renderer, and I haven't changed the _node_ tree at all. I've only changed the value of entry references that exist inside the node tree, which the rich-text renderer will take as-is and pass to the renderers that one provides in its config.

I will test it out. I honestly didn't use package I linked in the previous comment, so I don't know details, but given that it's linked in our docs/plugin readme - it would be good to make sure it does.

> I feel like this is hard to explain by text 🙈 but does that make sense?

I know the feel 🤗

jessepinho:
Thank you! 🙏

pieh:
I did test run and it still works with https://github.com/contentful/rich-text/tree/master/packages/rich-text-react-renderer with the change being that fields are localized from get-go (which is what we want).

I think it would be good to batch this PR and https://github.com/gatsbyjs/gatsby/pull/12816 together for major version bump. I can make prerelease with this change so you could install `gatsby-source-contentful@resolve-rich-text-locales` (or something like that) before we get the other PR in shape to be merged

jessepinho:
@pieh hm, is it not possible to simply do two major version bumps? My concern is just that, as this is a big repo, the chance of merge conflicts/etc. increases over time, and people won't be able to use this change in the meantime. Thoughts?

(If you absolutely want to wait on the major version bump, then the prerelease would help me a lot, thanks! However, does that mean I wouldn't get any new PRs that are merged into this repo until I switch back to `@latest`?)

jessepinho:
Wait, just one quick thing: if we could just merge and prerelease this now, that'd be great and would unblock me for some issues I'm facing — regardless of whether we do a version bump now or later :)

pieh:
> @pieh hm, is it not possible to simply do two major version bumps?

We could do that, but we don't want to make two major version bumps within week (or so) of each other for single package. There is no technical problem with that, just it's more respectful to our users to batch those breaking changes so they don't have to go through potential migrations multiple times if they upgrade often.

> Wait, just one quick thing: if we could just merge and prerelease this now, that'd be great and would unblock me for some issues I'm facing — regardless of whether we do a version bump now or later :)

This won't really work with how publishing with how lerna works (setup for our monorepo that helps with publishing to npm and package versioning). Anything merged into master will be published as `latest`.

jessepinho:
> > @pieh hm, is it not possible to simply do two major version bumps?
> 
> We could do that, but we don't want to make two major version bumps within week (or so) of each other for single package. There is no technical problem with that, just it's more respectful to our users to batch those breaking changes so they don't have to go through potential migrations multiple times if they upgrade often.

OK, makes sense.

> > Wait, just one quick thing: if we could just merge and prerelease this now, that'd be great and would unblock me for some issues I'm facing — regardless of whether we do a version bump now or later :)
> 
> This won't really work with how publishing with how lerna works (setup for our monorepo that helps with publishing to npm and package versioning). Anything merged into master will be published as `latest`.

OK. I've made a fork of this that includes my fixes, so that I can buy some time and not rush this process. So I'll just upgrade once the major version bump is done. Thanks!

jessepinho:
Note to @pieh: I've just pushed up another commit to handle array reference fields

jessepinho:
@pieh I've just pushed another commit as well to fix an infinite loop issue

jessepinho:
Hey @pieh, what's the status on this — waiting for the other PR to also merge?

jessepinho:
Hey @pieh, just checking in — will this definitely not be merged until #12816 is?

pieh:
#12816 is pretty much ready - it does need to be reviewed, so hopefully not much longer

jessepinho:
@pvdz I was out on holidays for most of the last month — sorry for the delay! I've created a ticket on our end to address your comments. I really like the idea of wrapping it in a config flag so that it's backwards-compatible! Will let you know when I've made more progress on this — thanks!

jessepinho:
Due to merge/rebase/conflict issues, I've superseded this PR with #21619. It hides this functionality behind a new config option, `richText.resolveFieldLocales`, and takes care of some of the optimization stuff @pvdz requested. Let's continue the convo over there.

PR #14622, opened 2019-06-07
fix(www): links in post-preview-component

NickyMeuleman:
Clicking anywhere within the center section on a contributor page goes to the same blogpost.

It was always the oldest blogpost, the last one in the HTML order.
What you are really clicking on is the "read more" label that makes the entire section a clickable target.

example: go to https://www.gatsbyjs.org/contributors/shannon-soper/ and click anywhere (but the name-tag) in the center section.

Added `pointer-events: none` to the CSS to prevent this.
Is this a servicable solution, what would be ideal?
/cc @marcysutton 	

edit: woops, went CSS, it's a JS object, camelcasing, I'm rusty
aaaaaaaaaaaand forgot to use backticks, that's what I get for using the github's web-interface.

### DIFF
--- a/www/src/components/blog-post-preview-item.js
+++ b/www/src/components/blog-post-preview-item.js
@@ -75,7 +75,7 @@ const BlogPostPreviewItem = ({ post, className }) => (
         textIndent: `-100%`,
         whiteSpace: `nowrap`,
         zIndex: 0,
-        pointer-events: 'none',
+        pointerEvents: 'none',
         "&&": { border: 0 },
       }}
     >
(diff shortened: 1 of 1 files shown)


DSchau:
Hey @NickyMeuleman, I think we actually want to solve this a little bit differently, e.g. by fixing the underlying positioning issue. (position: absolute without a parent position: relative on the blog post previews)

I've done that in #14540 (just merged!) which will also solve the problem that this PR was solving.

Thanks for your work here (and for the reminder to fix it/merge the PR!)! I'm going to close this in favor of the solution I implemented above--but please let me know if you'd like any more explanation for the _why_ of why the above solution is preferable!

DSchau:
To see _how_ (and validate!) that the above solution fixes it, go ahead and go to a contributor page and type this snippet into the Developer Console:

```
Array.from(document.querySelectorAll('article')).forEach(el => el.style.position = 'relative')
```

PR #14624, opened 2019-06-07
fix(gatsby): respect the GATSBY_CPU_COUNT env var, if set, by default for sharp

lovell:
## Description

Existing calls to `sharp.concurrency(cpuCoreCount())` are currently ignoring `GATSBY_CPU_COUNT` and will always use the physical CPU count, which can result in resource starvation in virtualised/container-based environments at build time.

The proposed change in this PR switches the default behaviour to observe and respect `GATSBY_CPU_COUNT`.

### DIFF
--- a/packages/gatsby/src/utils/cpu-core-count.js
+++ b/packages/gatsby/src/utils/cpu-core-count.js
@@ -4,7 +4,7 @@
  * @returns {number} Count of the requested type of CPU cores. Defaults to number of physical cores or 1
  */
 
-const cpuCoreCount = (useEnvVar = false) => {
+const cpuCoreCount = (useEnvVar = true) => {
   try {
     let coreCount = require(`./physical-cpu-count`) || 1
 
--- a/packages/gatsby/src/utils/__tests__/cpu-core-count.js
+++ b/packages/gatsby/src/utils/__tests__/cpu-core-count.js
@@ -1,0 +1,22 @@
+jest.mock(`../physical-cpu-count`, () => 1)
+const getCPUCoreCount = require(`../cpu-core-count`)
+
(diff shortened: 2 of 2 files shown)


DSchau:
Published as gatsby@2.8.6!

packages/gatsby/src/utils/cpu-core-count.js
```diff
@@ -4,7 +4,7 @@
  * @returns {number} Count of the requested type of CPU cores. Defaults to number of physical cores or 1
  */
 
-const cpuCoreCount = (useEnvVar = false) => {
+const cpuCoreCount = (useEnvVar = true) => {
```
stephenh: @DSchau FWIW is `useEnvVar` ever passed in as `false`? I only quickly looked at the `git blame` for when this code was introduced, but AFAICT it was only sharp that was calling this with `cpuCoreCount()`, and all the other callers were `cpuCoreCount(true)`.

Now that `cpuCoreCount()` is effectively `cpuCoreCount(true)`, I'm thinking its simpler to just remove the `useEnvVar` parameter all together.

DSchau: It's not - but this reserves the right to do so if needed.

stephenh: I personally find such "this _is_ configurable ... but oh is never actually used" patterns in code confusing b/c I get stuck wondering "...so when is this used?" only to realize ~X minutes of wasted time later that it's not, so would personally excise the extra/unused configuration, but your call. :-)

DSchau: I don't disagree :) PRs welcome!

PR #14627, opened 2019-06-07
style(blog): remove unnecessary closing tags

JohnTitor:
## Description

`<source>` is not needed to close so I removed `</source>`.

cf. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source


marcysutton:
I think if you're going to remove closing tags, they need to be self-closing: `<source />`

JohnTitor:
@marcysutton Thank you for your advice, I pushed following commit (should I squash commits?)

marcysutton:
> should I squash commits?

@JohnTitor no need, we'll squash on merge!

freiksenet:
Thanks!

PR #14630, opened 2019-06-07
feat(gatsby-source-drupal): added preview feature

grantglidewell:
## Description

Drupal source plugin Preview feature!

related Drupal module: https://www.drupal.org/project/gatsby


grantglidewell:
Merging the upstream changes to the source plugin has created issues issues, trying to work those out now.

grantglidewell:
I tested the upstream changes in the separate npm package `gatsby-source-drupal-preview` and it appears to be working. The caveat on existing projects is that you will need to kill node-modules and cache before a rebuild will work.

packages/gatsby-source-drupal/src/gatsby-node.js
```diff
@@ -29,7 +32,7 @@ exports.sourceNodes = async (
   // Default concurrentFileRequests to `20`
   concurrentFileRequests = concurrentFileRequests || 20
 
-  // Touch existing Drupal nodes so Gatsby doesn't garbage collect them.
+  // Touch existing Drupal nodes so Gatsby doesn`t garbage collect them.
```
KyleAMathews: this update seems like a mistake?

packages/gatsby-source-drupal/src/gatsby-node.js
```diff
@@ -264,4 +267,105 @@ exports.sourceNodes = async (
   for (const node of nodes) {
     createNode(node)
   }
+
+  if (process.env.NODE_ENV === `development` && preview) {
```
KyleAMathews: Curious if there's any way to unify logic between here & the regular fetches?

packages/gatsby-source-drupal/src/gatsby-node.js
```diff
@@ -264,4 +267,105 @@ exports.sourceNodes = async (
...
+    const server = micro(async (req, res) => {
```
pieh: Could you elaborate why do we need to create separate server?

grantglidewell: There is no way to access the express instance in this method, if it were exposed I would use it

packages/gatsby-source-drupal/src/gatsby-node.js
grantglidewell: could you elaborate?

packages/gatsby-source-drupal/src/gatsby-node.js
grantglidewell: yes, the result of moving from a different prettier and lint configuration (a find and replace) ill revert it

packages/gatsby-source-drupal/src/gatsby-node.js
pieh: Gotcha. I'll think about way to not have to do this. 

This introduces some potential issues - 8080 port is hardcoded, so there would be crashes when you have 2 drupal sources (or just trying to run 2 separate development servers with it). And instead of trying to address that it would be great to remove the need for spawning extra server

packages/gatsby-source-drupal/src/gatsby-node.js
```diff
@@ -15,6 +17,7 @@ exports.sourceNodes = async (
     headers,
     params,
     concurrentFileRequests,
+    preview,
```
pieh: I'm just wondering if we do need to make this feature opt-in? If we figure out a way to not have to spawn extra server - we could just always attach route handler

grantglidewell: I think if we're able to use the existing express instance there's no need to opt in.

packages/gatsby-source-drupal/src/gatsby-node.js
pieh: Checking more on this - seems like thing that are referenced are `backRefs`/`addBackRef` (that doesn't allow easily moving it as-is). I do think we can address this. I'll try to do some work on this before weekend.

packages/gatsby-source-drupal/src/gatsby-node.js
```diff
@@ -264,4 +267,105 @@ exports.sourceNodes = async (
...
+        } catch (e) {
+          // Ignore
+        }
+        if (fileNode) {
+          node.localFile___NODE = fileNode.id
+        }
+      }
+
+      node.internal.contentDigest = createContentDigest(node)
+      createNode(node)
```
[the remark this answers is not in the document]
grantglidewell: true, because of the way we have the Drupal plugin working, the dependent nodes are updated and then the final parent node is updated (the nodes to update are sent in a specific order). I haven't run into a problem with these getting out of sync, it works with paragraphs style data which seems to use this back dependency. There are definitely possible issues with this but I haven't found any that don't work correctly yet.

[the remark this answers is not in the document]
grantglidewell: I'm not sure if that's totally clear, but all affected nodes from a change are posted to the endpoint and updated

packages/gatsby-source-drupal/src/gatsby-node.js
grantglidewell: I totally agree, I know there would be potential ussies with exposing it to all source plugins, I believe @DSchau mentioned something about dragons 🐲

packages/gatsby-source-drupal/src/gatsby-node.js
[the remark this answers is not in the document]
pieh: Oh, so this is handled in Drupal module. Sorry for noise, I should have try debugging that.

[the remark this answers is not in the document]
grantglidewell: I still think there are potential issues, but for now it works because we have that ordering implemented there.

pieh:
Did another review pass and few notes:
 - there is no handling of content deletion currently. Not a blocker for initial merge, but need to know if there will be breaking changes to webhook payload to support that case
 - there is some trickiness with backreferences - right now those get duplicated and changes are not reflected - I'm working right now on fixing this (and some refactoring to share more code from initial sync handling and webhook handling)

packages/gatsby-source-drupal/src/gatsby-node.js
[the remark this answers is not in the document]
pieh: I was looking a little at drupal module code and it doesn't seem like there is any batching happening there - https://git.drupalcode.org/project/gatsby/blob/8.x-1.x/gatsby.module - am I missing something?

I added handling for back reference updates, so this shouldn't be an issue - I just wanted to fact check on this.

pieh:
All right, this should be good now:
 - removed `micro` (just using gatsby dev server now)
 - added handling of back references:
   - if updated node A no longer references other node (B) - that backreference get's removed
   - if updated node A has new reference - that backreference is added
 - added some basic tests for node updating and inserting

packages/gatsby-source-drupal/src/gatsby-node.js
```diff
@@ -114,170 +118,256 @@ exports.sourceNodes = async (
...
+    actions,
+    store,
+    cache,
+    createContentDigest,
+    reporter,
+  },
+  pluginOptions
+) => {
+  app.use(
+    `/___updatePreview/`,
```
pieh: I would propose to use `/__update/gatsby-source-drupal` or something like this here, so the update endpoint is namespaced. (This would need changes in Gatsby drupal module)

smthomas: If you provide me with the namespaces I can update the Drupal module to reflect this once it's merged. I'm assuming you would want to use `___update`, `___create` and `___delete`?

packages/gatsby-source-drupal/src/gatsby-node.js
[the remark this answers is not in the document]
smthomas: There is no batching currently happening. It currently sends all updates/inserts individually. This was the easiest way to get the proof of concept off the ground. If we can determine how we would like the data to be batched, we can add it to the roadmap for the Drupal module.

packages/gatsby-source-drupal/src/gatsby-node.js
pieh: I don't have opinion on separate endpoints for update/create/delete operations - it can stay as single endpoint and operation could be send with payload (for our use case update and create will have same handler anyway - only delete will have different handling). Only thing that I'm interested in is using namespace for `gatsby-source-drupal` (as we will likely start adding similar endpoints for other plugins)

pieh: For the sake of time. I think it's fine to merge this - this PR was in limbo for too long. I'll just add explicit note to README that there likely will be breaking changes coming that will require updating both gatsby-source-drupal and gatsby drupal module in future

pieh:
Published `gatsby-source-drupal@3.2.3`

PR #14639, opened 2019-06-08
[gatsby-remark-images] Add option for markdown captions

### LINKED ISSUE #14474: [gatsby-remark-images] When using `showCaption`, parse the caption as markdown
## Summary

When using `showCaptions: true` as an option for the `gatsby-remark-images`, it inserts the title of the image as raw html.
How about considering the caption as markdown and parsing it before inserting it?

### Motivation

https://github.com/mathieudutour/medium-to-own-blog/issues/5

Happy to work on it if that's something that you find interesting

johnletey:
## Description

I've added a `markdownCaptions` parameter that when set to true, the plugin will parse the markdown into html using a little helper function.

I did however add quite a few dependencies to do this, which in turn might make the package bloated?

## Related Issues

Fixes #14474

### DIFF
--- a/packages/gatsby-remark-images/package.json
+++ b/packages/gatsby-remark-images/package.json
@@ -14,7 +14,13 @@
     "mdast-util-definitions": "^1.2.0",
     "potrace": "^2.1.1",
     "query-string": "^6.1.0",
+    "rehype-document": "^3.0.1",
+    "rehype-format": "^2.3.2",
+    "rehype-stringify": "^6.0.0",
+    "remark-parse": "^6.0.3",
+    "remark-rehype": "^4.0.0",
     "slash": "^1.0.0",
+    "unified": "^7.1.0",
     "unist-util-select": "^1.5.0",
     "unist-util-visit-parents": "^2.0.1"
   },
--- a/packages/gatsby-remark-images/src/index.js
+++ b/packages/gatsby-remark-images/src/index.js
@@ -14,6 +14,23 @@ const Promise = require(`bluebird`)
 const cheerio = require(`cheerio`)
 const slash = require(`slash`)
 
+const unified = require(`unified`)
+const markdown = require(`remark-parse`)
+const remark2rehype = require(`remark-rehype`)
+const doc = require(`rehype-document`)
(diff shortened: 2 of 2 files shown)


mathieudutour:
Yeah that’s one way to do it but I’m not sure it’s very clean + it bypasses all the other remark plugin. 

Given it’s a remark plugin, I was hoping it’d be possible to reuse the existing remark pipeline but it turns out to be a lot more complicated than expected. Remark plugins don’t get access to remark unfortunately.  I’m not sure if there is an easy solution for it.

johnletey:
@mathieudutour This is the first thing that came to mind ... I’m open to other solutions!

DSchau:
@mathieudutour re: your comment:

> Remark plugins don’t get access to remark unfortunately. I’m not sure if there is an easy solution for it.

Why not solve this problem which will make _this_ problem simpler? We can definitely pass the remark instance, and I don't think that's necessarily a bad idea, actually.

As it stands currently -- I don't think we'll merge this PR. This does add a bit of complexity and bloat, and I think we'd prefer to not as this in as it's currently authored. Why?

1. I'd honestly prefer documenting how to use figcaption (and/or generating this structure in your medium -> md utility; note figcaption with img(src) will still be picked up by gatsby-remark-images), or
2. Pass the remark instance to child plugins
    - Note: you'll want to guard against it _not_ existing (e.g. if using a version of gatsby-transformer-remark that doesn't inject remark)

I'm OK with 1 or 2, but I think I'd push back against the current solution. Going to close this out until we can solve this more cleanly.

Hope that's OK, and thanks for the PR! If we can revisit 2) I think this could be worth merging if we can simplify some of the dependencies and complexity here!

mathieudutour:
2. sounds good but I’m not sure what you mean with 1.. Is there a markdown syntax for using figcaption directly?

johnletey:
Totally fine to close @DSchau ... it was just a quick fix that I through together. As far as moving forward, I don't totally understand what you want to do ... so maybe @mathieudutour can take over. Of course I'd be happy to make another PR but only once I know exactly what needs to be done!

PR #14649, opened 2019-06-08
feat(graphql): add support for $not $elemMatch filter

### LINKED ISSUE #14565: Filter out objects with array containing value
## Summary
When writing a graphql query, I want to be able to search for Articles, and remove any results where tags contains a tag with a value. For example, excluding all Articles, that contains a tag of value 'Featured' in the tags array.

I have attempted to use the elemMatch, but this does not appear to solve my issue.

## Relevant information
Example Article:
`
{
  title: 'Example',
  tags: [
    { tag: 'Featured' },
    { tag: 'Blog' },
    { tag: 'Custom' },
  ]
}
`

### Environment (if relevant)
  System:
    OS: macOS 10.14.4
    CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Yarn: 1.5.1 - /usr/local/bin/yarn
    npm: 6.9.0 - ~/.nvm/versions/node/v10.14.2/bin/npm
  Languages:
  Browsers:
    Chrome: 74.0.3729.169
    Firefox: 67.0
    Safari: 12.1
  npmPackages:
    gatsby: ^2.5.0 => 2.5.0 
    gatsby-image: ^2.0.41 => 2.0.41 
    gatsby-plugin-extract-schema: 0.0.5 => 0.0.5 
    gatsby-plugin-material-ui: ^2.0.1 => 2.0.1 
    gatsby-plugin-offline: ^2.1.0 => 2.1.0 
    gatsby-plugin-sharp: ^2.0.37 => 2.0.37 
    gatsby-remark-component: ^1.1.3 => 1.1.3 
    gatsby-source-filesystem: ^2.0.35 => 2.0.35 
    gatsby-source-strapi: 0.0.8 => 0.0.8 
    gatsby-transformer-remark: ^2.3.12 => 2.3.12 
    gatsby-transformer-sharp: ^2.1.19 => 2.1.19 
  npmGlobalPackages:
    gatsby-cli: 2.6.2

`gatsby-config.js`: N/A 
`package.json`: N/A 
`gatsby-node.js`: N/A 
`gatsby-browser.js`: N/A 
`gatsby-ssr.js`: N/A

superhawk610:
## Description

It's not currently possible to filter a set of inputs for those that don't contain an object with a given key. For example, given the following input data:

```
const data = [
  { id: 1, tags: [] },
  { id: 2, tags: [{ name: `Featured` }] },
  { id: 3, tags: [{ name: `Not Featured` }] },
]

// filter should return { id: 1 } and { id: 3 }
```

it is not possible to filter for all records where `tags` does not contain any tag where `tag.name === 'Featured'`. This _is_ supported by sift using `$not $elemMatch` as follows

```
{ tags: { $not: { $elemMatch: { tag: { $in: ['Featured'] } } } } }
```

This PR adds support for this filter pattern by adding a new key to the GQL type `*ListInput`, `not`, of type `*ListInputWithoutNot` (to prevent infinite recursion).

## Clarification

It's not immediately obvious why `$not $elemMatch` is different than `$nin`, which is current supported. To explain why this is required, consider the following dataset:

```
const data = [
  { id: 1, tags: [] },
  { id: 2, tags: [{ name: `Featured` }] },
  { id: 3, tags: [{ name: `Not Featured` }] },
  { id: 4, tags: [{ name: `Featured`, name: `Something Else` }] },
]

const filter = { tags: { $elemMatch: { tag: { $nin: ['Featured'] } } } }
```

You might expect this to perform the same as the previous set (returning `{ id: 1 }` and `{ id: 3 }`) but when filtering with `$nin` it will _also_ return `{ id: 4 }` because record 4 contains at least one tag where `tag.name !== 'Featured'`.

## Related Issues

Closes #14565.


superhawk610:
I'm having some difficulty getting the implementation for this, I'd appreciate some help from anyone more knowledgeable about Gatsby's GQL internals :smile:

packages/gatsby/src/redux/run-sift.js
```diff
@@ -15,7 +15,14 @@ const prepareQueryArgs = (filterFields = {}) =>
   Object.keys(filterFields).reduce((acc, key) => {
     const value = filterFields[key]
     if (_.isPlainObject(value)) {
-      acc[key === `elemMatch` ? `$elemMatch` : key] = prepareQueryArgs(value)
+      switch (key) {
+        case `not`:
+        case `elemMatch`:
+          acc[`$${key}`] = prepareQueryArgs(value)
```
stefanprobst: We need to do something similar [here](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/db/loki/nodes-query.js#L50) as well. (For context: Gatsby supports a second, not-very-advertised backend that uses `LokiJS`)

stefanprobst:
Thanks for working on this @superhawk610 !

One question: are there scenarios where `$not` could be useful other than in combination with `$elemMatch`?

Wrt to the implementation, [`dropQueryOperators`](https://github.com/gatsbyjs/gatsby/blob/09d0d10ebe1b7a0d85186266f15a855571269cee/packages/gatsby/src/db/common/query.js#L40) would also have to be adjusted to drop `$not` as well.

superhawk610:
@stefanprobst off the top of my head I can't think of any other situations where `$not` is required, generally `$neq` and `$nin` should be sufficient.

I'll try to get the implementation fixed up this evening, but if you can think of other instances where it may be necessary definitely let me know.

Naismith:
@superhawk610 @stefanprobst While there may not be any other $not implementations that are required, is there any other operators covered by sift that are not supported?

stefanprobst:
@superhawk610 How's it going with this? If you have any questions don't hesitate!

@Naismith While there are other operators supported by Sift (and Loki), I'd prefer we only add operators based on actual usecases, since most scenarios can also be handled by adding a custom field resolver.

superhawk610:
I've been sidetracked with some other work, hopefully should have some time in the next week to work on this!

sidharthachatterjee:
I fixed `dropQueryOperators` to support `not` and also added a test for this. The test passes for `sift` but fails for `loki` currently. 

Yet to add support for `loki`

sidharthachatterjee:
Loki doesn't seem to support `$not`

https://github.com/techfort/LokiJS/wiki/Query-Examples

Cc @freiksenet @stefanprobst

stefanprobst:
yes, i can pick this up -- but it's not a priority atm -- need to figure out how loki expects the $not operator

freiksenet:
Hello!

Sorry for dropping the ball on that.

I would prefer to be very conservative with adding additional query operators - we already have quite a lot of trouble implementing compatibility over two possible backends. In addition, I don't think we should add a new operator just for one use case.

I would however support adding `elemNotMatch` operator if that is possible to implement for both redux and loki backends.

Based on that I will be closing this PR. Thank you for working on this and I'm sorry that I have to do that.

PR #14657, opened 2019-06-09
Give gatsby-plugin-mdx

ChristopherBiscardi:
## Description

As discussed previously with @KyleAMathews, since gatsby-mdx is becoming an increasingly important part of the ecosystem, the code should live in the Gatsby repo. This PR is the moving of gatsby-plugin-mdx into the Gatsby repo, including the history of the plugin (for that reason, I'd suggest keeping the commits around instead of squashing when merging).

The development of gatsby-mdx has historically happened in a `packages/` directory, so the merge here is clean and affects no files currently in the gatsby repo.

The commits in this repo reflect the currently published version of gatsby-plugin-mdx entirely. There are some changes that I want to make to the README to reflect the new location, etc but I feel those changes should happen *after* this is merged so that we aren't stacking new changes on top of this PR without publishing. That means this PR is *just* the porting of the currently existing release.

I will still be around to maintain the code.

# TODO

### Before merging:

- [x] give publish rights to gatsby (@gatsbyjs/core who do I give the npm publish rights to?)

### After merging:

- [ ] Port issues from https://github.com/ChristopherBiscardi/gatsby-mdx/issues to the gatsby repo.
  - Many of these can be closed. There are 27 issues right now.
- [ ] Update documentation on gatsbyjs.org
  - Involves updating references to `gatsby-plugin-mdx` like on https://www.gatsbyjs.org/docs/mdx/getting-started 
- [ ] Add myself as a CODEOWNER to the gatsby-plugin-mdx package? (not sure what the convention is here with respect to labelling/codeowners/etc)
- [ ] @ChristopherBiscardi will write a blog post about the 1.0 and merging into the Gatsby repo. The 1.0 is dark-launched right now and used on .org, the blog post is for visibility.


ChristopherBiscardi:
`@mdx-js/mdx` doesn't work on node 6 due to syntax. What's the best way to skip node 6 for this test?

```
 FAIL  packages/gatsby-plugin-mdx/loaders/mdx-loader.test.js
  ● Test suite failed to run

    Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    /home/circleci/project/node_modules/@mdx-js/mdx/index.js:116
    async function compile(mdx, options = {}) {
          ^^^^^^^^

    SyntaxError: Unexpected token function

       9 | const htmlAttrToJSXAttr = require("./babel-plugin-html-attr-to-jsx-attr");
      10 | const BabelPluginPluckImports = require("./babel-plugin-pluck-imports");
    > 11 | 
         | ^
      12 | /*
      13 |  * function mutateNode({
      14 |  *   pluginOptions,

      at ScriptTransformer._transformAndBuildScript (node_modules/jest-runtime/build/ScriptTransformer.js:435:17)
      at Object.<anonymous> (packages/gatsby-plugin-mdx/utils/gen-mdx.js:11:13)
      at Object.<anonymous> (packages/gatsby-plugin-mdx/loaders/mdx-loader.js:35:16)
```

wardpeet:
@ChristopherBiscardi define best 😛 

Would something like this work?
```js
const testIfCondition = (condition) => condition ? it.skip : it;

testIfCondition(!/^v6/.test(process.version))('your test', ...)
```

lannonbr:
We should be able to remove the conditional for Node 6 for the tests now as we have dropped Node 6 support as of Gatsby 2.10.0

ChristopherBiscardi:
I've removed the node6 modifications and the tests, etc are passing :D

packages/gatsby-plugin-mdx/README.md
```diff
@@ -0,0 +1,466 @@
+![Logo](../../img/gatsby-mdx.png)
```
muescha: this image is missing in this commit ...

see: https://github.com/ChristopherBiscardi/gatsby-mdx/tree/master/img

PR #14666, opened 2019-06-09
[www] Homepage Call To Actions

andrewfairlie:
## Description

This is a suggestion to improve (imo) the main call to action on the homepage which is currently `Get Started` into the more active `Start Using Gatsby`.

It's a small change but one that I suspect will make a difference. `Get Started` feels like it's the beginning of a long journey that  I might not have time for right now. `Start Using Gatsby` sounds more immediate and less intimidating - which it is!

### DIFF
--- a/www/src/components/masthead.js
+++ b/www/src/components/masthead.js
@@ -69,7 +69,7 @@ const MastheadContent = () => (
       tracking="MasterHead -> Get Started"
       icon={<ArrowForwardIcon />}
     >
-      Get Started
+      Start Using Gatsby
     </Button>
   </div>
 )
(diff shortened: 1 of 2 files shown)


freiksenet:
CC @fk

gillkyle:
This is likely something @lindaleebumblebee has thought about so I'm also going to include her in this thread. They've been thinking a lot about this kind of marketing copy and probably have an opinion to weigh in with.

gillkyle:
I spoke to @lindaleebumblebee about this and she said she thought it best to leave it as 'get started' because it still gets the point across, is short, and it's consistent with the CTA on gatsbyjs.com as well. The folks working on marketing have been putting some good thought into this stuff so I trust them.

Thanks for the suggestion though @andrewfairlie!

PR #14679, opened 2019-06-10
fix(gatsby): Make createPageDependency public again

freiksenet:

### DIFF
--- a/packages/gatsby/src/redux/actions/public.js
+++ b/packages/gatsby/src/redux/actions/public.js
@@ -1175,6 +1175,9 @@ actions.createPageDependency = (
   }: { path: string, nodeId: string, connection: string },
   plugin: string = ``
 ) => {
+  console.warn(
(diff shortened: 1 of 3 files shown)


packages/gatsby/src/redux/actions/public.js
```diff
@@ -1158,4 +1158,35 @@ actions.createRedirect = ({
...
+actions.createPageDependency = (
+  {
+    path,
+    nodeId,
+    connection,
+  }: { path: string, nodeId: string, connection: string },
+  plugin: string = ``
+) => {
+  console.warn(
+    `Calling "createPageDependency" directly from actions in deprecated. Use "createPageDependency" from "gatsby/dist/redux/actions/add-page-dependency".`
```
pieh: I don't quite understand the warning message - people should import it from gatsby internals?

freiksenet: It is described in docs https://www.gatsbyjs.org/docs/page-node-dependencies/#add-page-dependencyjs

PR #14688, opened 2019-06-10
Adds missing semicolon

bravo-kernel:
The tutorial fails for me without this semicolon. Gatsby 2.6.5

### DIFF
--- a/docs/tutorial/part-four/index.md
+++ b/docs/tutorial/part-four/index.md
@@ -315,7 +315,7 @@ export default ({ children }) => {
         }
       }
     `
-  )
+  );
   // highlight-end
   // highlight-start
   return (


bravo-kernel:
Hmm, it looks like I might have been a bit too fast here, seem the semicolon is not accepted

m-allanson:
Thanks @bravo-kernel! Semicolons are mostly optional in javascript (except in some special cases). Gatsby uses [prettier](https://prettier.io) to remove semicolons wherever possible. 

If you're seeing a problem that's fixed by adding a semicolon it might indicate a problem in a different line of code. I'm going to close this PR, but please open an issue if the tutorial isn't working as expected. Thanks again!

bravo-kernel:
Apologies, I was probably too quick here.

PR #14691, opened 2019-06-10
fix(gatsby-remark-graphviz): fix graphviz on node < 10

pieh:
First commit just adds some basic tests - when tests are hopefully red, will push another commit that actually fixes it

### DIFF
--- a/packages/gatsby-remark-graphviz/src/__tests__/index.js
+++ b/packages/gatsby-remark-graphviz/src/__tests__/index.js
@@ -1,0 +1,73 @@
+const Remark = require(`remark`)
+const toHAST = require(`mdast-util-to-hast`)
+const hastToHTML = require(`hast-util-to-html`)
+const cheerio = require(`cheerio`)
+const plugin = require(`../`)
+
+const remark = new Remark().data(`settings`, {
+  commonmark: true,
+  footnotes: true,
+  pedantic: true,
+})
+
+const run = async content => {
+  const markdownAST = remark.parse(content)
+  await plugin({ markdownAST })
+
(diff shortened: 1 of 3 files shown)


packages/gatsby-remark-graphviz/src/index.js
```diff
@@ -11,13 +11,19 @@ module.exports = async ({ markdownAST }, pluginOptions = {}) => {
...
   visit(markdownAST, `code`, node => {
-    const chunks = (node.lang || ``).match(
-      /^(?<lang>\S+)(\s+(?<attrString>.+))?/
-    )
+    const chunks = (node.lang || ``).match(/^(\S+)(\s+(.+))?/)
+
+    if (!chunks || !chunks.length) {
+      return node
+    }
+
```
sidharthachatterjee:
```suggestion
hardcoded to indices 1 and 3 because we can't use named capture groups - those work in node 10+ and we don't transpile it right now
```

packages/gatsby-remark-graphviz/src/__tests__/index.js
```diff
@@ -0,0 +1,73 @@
...
+    expect($(`svg`).length).toBe(1)
+    expect($(`pre`).length).toBe(0)
+    expect($(`code`).length).toBe(0)
+  })
+
+  it(`unknown graph lang`, async () => {
+    const test = await run(`
+\`\`\`pieh-format
+  digraph graphname {
+    a :heart: b;
```
sidharthachatterjee: ❤️

sidharthachatterjee:
Published in `gatsby-remark-graphviz@1.0.11`

PR #14698, opened 2019-06-11
#14651 Fix `gatsby new` bug on initing a new project in a path with spaces

### LINKED ISSUE #14651: gatsby new fails if initialized with a path that has a space in it
## Description

When I make a new empty directory that contains a space in the folder new and type `gatsby new` a git error is thrown:

```
Error: Command failed: git clone https://github.com/gatsbyjs/gatsby-starter-default.git   /Users/trevorlitsey/Documents/code/3 notes/netlify --single-branch
```

### Steps to reproduce

```(bash)
mkdir 'some where' && cd 'some where'
gastby new
```

### Expected result

Gatsby should init a new project

### Actual result

A git error is throw because the path is split into two arguments

### Potential fix

The path should be wrapped in 'quotes'

### Environment

```
  System:
    OS: macOS 10.14.2
    CPU: x64 Intel(R) Core(TM) i5-4278U CPU @ 2.60GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Yarn: 1.13.0 - ~/.yarn/bin/yarn
  Browsers:
    Chrome: 74.0.3729.169
    Firefox: 67.0
    Safari: 12.0.2
  npmGlobalPackages:
    gatsby-cli: 2.4.8
    gatsby: 2.8.6
```

trevorlitsey:
## Description

When running `gatsby new` in an empty directory, call `git clone` for with args as array to preserve potential spaces in the user's path.

It would likely be a good idea to remove the `cmd.split(/\s+/)` logic altogether, but I don't see it causing any other issues at the moment. In the interesting of maintaining a small footprint here is a first pass at a solution.

## Related Issues

Fixes #14651

### DIFF
--- a/packages/gatsby-cli/src/init-starter.js
+++ b/packages/gatsby-cli/src/init-starter.js
@@ -17,8 +17,10 @@ const {
 const isTTY = require(`./util/is-tty`)
 const spawn = (cmd: string, options: any) => {
   const [file, ...args] = cmd.split(/\s+/)
-  return execa(file, args, { stdio: `inherit`, ...options })
+  return spawnWithArgs(file, args, options)
 }
+const spawnWithArgs = (file: string, args: string[], options: any) =>
+  execa(file, args, { stdio: `inherit`, ...options })
 
 // Checks the existence of yarn package and user preference if it exists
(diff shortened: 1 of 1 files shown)


freiksenet:
Thank you!

PR #14704, opened 2019-06-11
feat(gatsby-link): Add deprecation warning if Link is imported directly from gatsby-link

sidharthachatterjee:


packages/gatsby-link/src/index.js
```diff
@@ -4,6 +4,12 @@ import { Link } from "@reach/router"
 
 import { parsePath } from "./parse-path"
 
+if (!module.parent.id.includes(`gatsby-browser-entry`)) {
```
sidharthachatterjee: module.parent doesn't work with webpack and even if it did, we're expecting the `caller` where `parent` is set to the first module that imported the current module (and therefore resulted it in being added to the resolve cache)

sidharthachatterjee:
Only other potential approach (as recommended by @wardpeet is to move current exports to a separate file like `internal.js` and use an index.js file that re-exports them with a deprecation message) therefore showing the message if imported directly (gatsby will import from `internal.js`)

Doesn't seem like the deprecation message is worth the extra hassle and maintenance cost so we're closing this out 😞

wardpeet:
We might want to look into an eslint rule instead.

PR #14705, opened 2019-06-11
feat(gatsby): Re-export Img from gatsby

sidharthachatterjee:
This re-exports `Img` (from `gatsby-image`) from `gatsby` for convenience 

Tried adding a deprecation warning in https://github.com/gatsbyjs/gatsby/pull/14704 but doesn't seem like it is worth the effort


packages/gatsby/cache-dir/gatsby-browser-entry.js
```diff
@@ -82,6 +86,8 @@ function graphql() {
 }
 
 export {
+  Img,
+  ImgWithIEPolyfill,
```
KyleAMathews: Does this tree shake?

sidharthachatterjee: Good question, let me check!

marcysutton:
We could definitely do a search and replace to update references in the docs. A blog post would be a good idea, too! Love the codemod idea. 👍

marcysutton:
Also: does this mean `gatsby-image` no longer has to be installed separately?

packages/gatsby/cache-dir/gatsby-browser-entry.js
wardpeet: I don't think it does, maybe webpack is smart enough at the current version but it wasn't in the past, you'll have to do a direct export.

```js
export { default as Img } from "gatsby-image"
export { default as ImgWithIEPolyfill } from "gatsby-image/withIEPolyfill"
```

marcysutton:
Hey @sidharthachatterjee I wanted to check up on this PR. Can you chime in when you get a chance? It's on my mind as we work through the image workflow in the docs, as this would change things a bit.

m-allanson:
I like this change but I am a bit worried about the docs implications. Excluding the official Gatsby docs there's a _ton_ of documentation out there that won't update to use this way of importing `gatsby-image`.

I feel like having two ways of doing the same thing is likely to be more confusing than one slightly-less-good way of doing that thing.

I can see cases where people might have `gatsby-image` already installed, then start referencing official docs with `import { Img } from "gatsby-image"`, and end up with strange problems because they're using different versions of `gatsby-image` in different parts of the same project.

Not needing to install `gatsby-image` is a good improvement! But you'll probably still need to install `gatsby-transformer-sharp` and `gatsby-plugin-sharp` so this doesn't reduce friction that much.

I feel like this change might be better as a Gatsby v3 change? Then we could deprecate the old way of using `gatsby-image` (manually installing it), and recommend everyone switch to importing `gatsby-image` from `gatsby`. 

@marcysutton @sidharthachatterjee what do you think? Am I overthinking this and worrying too much? :D

packages/gatsby/cache-dir/gatsby-browser-entry.js
sidharthachatterjee: Done

wardpeet:
I'm on the same page as @m-allanson. I don't see much improvements on moving it to the gatsby package 🙊

sidharthachatterjee:
Sorry for missing your comment @marcysutton 🤦‍♂ 

Closing this after a discussion with @m-allanson and @KyleAMathews because 
- This doesn't really solve the original problem (simplifying the usage of `gatsby-image`) since users will still need to add `gatsby-plugin-sharp` and `gatsby-transformer-sharp`
- Breaking changes to `gatsby-image` will force a major bump for `gatsby` too

PR #14706, opened 2019-06-11
chore(docs): Fix use of navigate

dnakhla:
The navigate fn will not work on build since it needs access to the window which is only available in componentDidMount.


freiksenet:
Thank you!

PR #14710, opened 2019-06-11
feat(themes): header/page-layout theme concept

jxnblk:
## Description

This is a demonstration of an idea I mentioned to @ChristopherBiscardi the other day. When composing multiple themes together in a single site, there are a few different ways to share page layout components (e.g. headers/footers/etc) across multiple paths. One option is to leave page layout concerns to the end user building their site. While this approach is very intentional and flexible, it does require more work than simply installing a theme.

Another approach is to build page layout into individual themes and allow the user to import and reuse layout components from one theme in another. This also requires some amount of work to set up.

With the approach demonstrated in this PR, the page header component is provided by a separate `gatsby-theme-header` theme than can be imported, shadowed, and customized by themes or sites. Here, `gatsby-theme-header` is a dependency of both the blog and notes themes. Each one imports the `Header` component for use in its own layout. The notes theme imports the component without shadowing it. The blog theme imports the component *and* shadows the header component to include site metadata as the title and the dark mode toggle switch. The notes theme also uses this shadowed header component to render in its own layout.

The example `gatsby-theme-starter` site also shadows the header component to add a link to the `/notes` path (this hasn't been styled to match the themes, but is included as a demo). 

I've kept this scoped to a single component for demonstration purposes, but the idea could be handled with something like a `gatsby-theme-page-layout` theme that included multiple components.

### Questions

- What are your general thoughts on utilizing a pattern like this?
- Are there any obvious problems with this approach that I'm missing?
- Is this too complex (or too confusing) for the official themes?

## Related Issues

Related #14276

### DIFF
--- a/themes/gatsby-starter-theme/gatsby-config.js
+++ b/themes/gatsby-starter-theme/gatsby-config.js
@@ -1,18 +1,14 @@
 module.exports = {
   __experimentalThemes: [
-    /*
     {
       resolve: `gatsby-theme-blog`,
       options: {},
     },
-    */
     {
       resolve: `gatsby-theme-notes`,
       options: {
-        // mdx: false,
-        notesPath: `/`,
-        // homeText: `HOME`,
-        // breadcrumbSeparator: `⚡️`,
+        mdx: false,
+        notesPath: `/notes`,
       },
     },
   ],
--- a/themes/gatsby-starter-theme/gatsby-config.js
+++ b/themes/gatsby-starter-theme/gatsby-config.js
@@ -1,16 +1,18 @@
 module.exports = {
   __experimentalThemes: [
+    /*
     {
       resolve: `gatsby-theme-blog`,
       options: {},
     },
+    */
     {
       resolve: `gatsby-theme-notes`,
       options: {
-        mdx: false,
-        notesPath: `/txt`,
-        homeText: `HOME`,
-        breadcrumbSeparator: `⚡️`,
+        // mdx: false,
+        notesPath: `/`,
+        // homeText: `HOME`,
+        // breadcrumbSeparator: `⚡️`,
       },
     },
   ],
--- a/themes/gatsby-starter-theme/src/gatsby-theme-header/index.js
+++ b/themes/gatsby-starter-theme/src/gatsby-theme-header/index.js
@@ -1,0 +1,17 @@
+import React from 'react'
+import { Link } from 'gatsby'
+
+export default props =>
+  <header>
+    <h2>
+      Custom Header
+    </h2>
+    <ul>
+      <li>
+        <Link to='/'>Blog</Link>
+      </li>
+      <li>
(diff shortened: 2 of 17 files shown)


jlengstorf:
Overall this seems reasonable. I worry a little bit about having to install a half-dozen themes to get shared components in place, but I guess that's up to us to educate against.

It's out of scope for this PR, but this got me thinking about providing common shadowing for horizontally composed themes.

Like, if I want to do something like:

```
          / theme-a \
base --> |           | --> common-header-override --> site
          \ theme-b /
```

That seems like a reasonable pattern to expect, but right now you'd have to shadow twice. What might an API look like to allow for "common" overrides? 🤔

jxnblk:
@jlengstorf I might not be totally understanding the setup you're describing here...

> That seems like a reasonable pattern to expect, but right now you'd have to shadow twice

Why would it require shadowing twice? If `theme-a` and `theme-b` are both themes that are configured in `site`, and they both use `gatsby-theme-header`, then `site` can shadow both at `src/gatsby-theme-header/index.js`, right?

> I worry a little bit about having to install a half-dozen themes to get shared components in place

Yeah, I think this exact pattern could escalate out of control – I think having a more general purpose `gatsby-theme-layout` one, or even adding this to `gatsby-theme-ui` might make more sense.

ChristopherBiscardi:
> Yeah, I think this exact pattern could escalate out of control

IMO it's pretty hard to say whether this will end up out of control or not. It seems mostly like a pattern that is useful for theme groups (themes that are meant to be used together).

The experimentation is interesting and when themes become "just plugins" it's possible that this pattern becomes more stable as it's used more and more. 

> or even adding this to gatsby-theme-ui might make more sense.

I've been viewing theme-ui and gatsby-theme-ui, etc as more generally usable tools than just the official themes, so adding the official theme layouts there seems weird.

jxnblk:
@ChristopherBiscardi thanks! I think I generally agree with you, and with the idea that themes like this would essentially be plugins with components that can be shadowed, I suspect someone will do this somewhere eventually.

> I've been viewing theme-ui and gatsby-theme-ui, etc as more generally usable tools than just the official themes, so adding the official theme layouts there seems weird.

I definitely want Theme UI to be a more widely-applicable tool, and it might not make sense for the layout theme stuff to go in there – it is a fairly simple implementation and was more thinking about minimizing the number of hard dependencies

jlengstorf:
> If theme-a and theme-b are both themes that are configured in site, and they both use gatsby-theme-header, then site can shadow both at src/gatsby-theme-header/index.js, right?

Yeah, that's true. The challenge then becomes exposing _where_ to shadow, because to the person using this theme, they don't _have_ a `gatsby-theme-header` — they only have the A and B themes.

I think @ChristopherBiscardi is right, though; we can't know what will happen until people start using this stuff. Feel free to ignore me. 😄

jxnblk:
Closing in favor of #15215

PR #14711, opened 2019-06-11
chore(telemetry): add ability to inject tags from env and track websocket updates

jamo:


packages/gatsby/src/utils/websocket-manager.js
```diff
@@ -190,13 +227,19 @@ class WebsocketManager {
...
     }
   }
 
   emitPageData(data: QueryResult) {
     this.pageResults.set(data.id, data)
     if (this.isInitialised) {
       this.websocket.send({ type: `pageQueryResult`, payload: data })
+      telemetry.trackCli(`WEBSOCKET_EMIT_PAGE_DATA_UPDATE`, {
+        siteMeasurements: { clientsCount: this.connectedClients },
+      })
```
pieh: This might lead to unexpected results. If data update would make multiple queries/pages "dirty" (i.e. updating blog post cause updating blog posts listings, blog author personal page and updating actual blog post page) - this will be called N times for single data update

packages/gatsby-telemetry/src/telemetry.js
```diff
@@ -82,14 +101,18 @@ module.exports = class AnalyticsTracker {
...
 
     if (tags.error) {
       // `error` ought to have been `errors` but is `error` in the database
       tags.error = sanitizeErrors(tags.error)
     }
 
-    this.buildAndStoreEvent(eventType, Object.assign(tags, decoration))
+    this.buildAndStoreEvent(
+      eventType,
+      lodash.merge({}, tags, decoration, permanetDecoration)
```
KyleAMathews: s/permanet/permanent/

packages/gatsby/src/utils/websocket-manager.js
KyleAMathews: yeah, I'd aggregate what all the pages are that are updated in a go and send that all together. That + what page they're looking at seems like the right "event" not updating each page.

jamo:
Added debouncing, cleaned up the code a bit. And also adding a patch improving the `activePath` detection for the websocket manager.

jamo:
@KyleAMathews @pieh  I just updated the PR. 
Please take another look.

Also, let me know before releasing this as there is one change pending in the upstream api that I want to time with this change :)

packages/gatsby-telemetry/src/telemetry.js
```diff
@@ -71,9 +85,21 @@ module.exports = class AnalyticsTracker {
     }
 
     const decoration = this.metadataCache[type]
-    delete this.metadataCache[type]
     const eventType = `${baseEventType}_${type}`
-    this.buildAndStoreEvent(eventType, Object.assign(tags, decoration))
+
+    if (debounce.debounce) {
```
sidharthachatterjee: Can't this just be `debounce` as opposed to `debounce.debounce`?

sidharthachatterjee:
Blocked till upstream API is updated

packages/gatsby/src/utils/websocket-manager.js
```diff
@@ -172,9 +215,11 @@ class WebsocketManager {
 
       s.on(`disconnect`, s => {
         leaveRoom(activePath)
+        this.connectedClients -= 1
       })
 
       s.on(`unregisterPath`, path => {
+        console.log(`unregisterPath`, path)
```
sidharthachatterjee:
```suggestion

```

packages/gatsby-telemetry/src/telemetry.js
jamo: Updated, renamed the params object to opts for future expansibility.

packages/gatsby-telemetry/src/telemetry.js
```diff
@@ -60,7 +74,7 @@ module.exports = class AnalyticsTracker {
     }
     return undefined
   }
-  captureEvent(type = ``, tags = {}) {
+  captureEvent(type = ``, tags = {}, opts = { debounce: false }) {
```
sidharthachatterjee: Want to set debounce: true by default?

jamo: I'm slightly worried about the possible side-effects so hence I'd leave it as it was (i.e. no debouncing by default)

jamo:
@sidharthachatterjee this should be good to go, the upstream changes have been applied.

sidharthachatterjee:
Published in `gatsby-telemetry@1.0.12`

PR #14718, opened 2019-06-11
fix[gatsby-plugin-offline]: Ommit per-page-manifest preload link

zakdoek:
## Description

Removes the preload link, when the offline plugin is used, also for per-page-manifest data which are located differently under `/public/page-data/`.


KyleAMathews:
Published in 2.1.3

PR #14720, opened 2019-06-11
fix(gatsby-plugin-offline): use networkFirst caching for page-data.json files

Moocar:
## Description

`page-data.json` files are not content hashed. Their value can change between builds and the filename will remain the same. Therefore, in the offline plugin, we should use [networkFirst](https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook/#network-falling-back-to-cache) strategy.

## Related Issues

- https://github.com/gatsbyjs/gatsby/pull/13004

### DIFF
--- a/packages/gatsby-plugin-offline/src/gatsby-node.js
+++ b/packages/gatsby-plugin-offline/src/gatsby-node.js
@@ -92,7 +92,7 @@ exports.onPostBuild = (args, pluginOptions) => {
       },
       {
         // page-data.json files are not content hashed
-        urlPattern: /^https:?:*page-data\.json/,
+        urlPattern: /^https?:.*\page-data\/.*\/page-data\.json/,
         handler: `networkFirst`,
       },
       {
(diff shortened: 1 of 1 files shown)


packages/gatsby-plugin-offline/src/gatsby-node.js
```diff
@@ -90,6 +90,11 @@ exports.onPostBuild = (args, pluginOptions) => {
         urlPattern: /(\.js$|\.css$|static\/)/,
         handler: `cacheFirst`,
       },
+      {
+        // page-data.json files are not content hashed
+        urlPattern: /^https:?:*page-data\.json/,
```
KyleAMathews: This didn't seem to work when I tested it — something like this did https://regex101.com/r/WgnMvi/1

Moocar: Ahh interesting, I thought I saw it working locally, but I'll try again. With these changes.

Moocar: Ok, fixed and actually tested changing `page-data.json` to make sure the new results are shown on the first network request

Moocar:
Thanks for testing! I don't mind about the `$` suffix. The regex is already quite specific. Would it help with performance at all?

KyleAMathews:
Yeah, seems fine without it. Merging!

Issue #14468, opened 2019-05-31
[www/showcase] Twitter cards not working for individual "Showcase" entries


fk:
Happened to all three URLs I tested:

![image](https://user-images.githubusercontent.com/21834/58722475-cc482c00-83d7-11e9-829e-52e7c2b99688.png)

rlynn523:
I would like to pick this up if possible

fk:
Hey @rlynn523, that'd be great! 🙏

rlynn523:
@fk is there a dev or test env I can deploy to? I can only really validate the twitter cards when the code is deployed. I tried to use ngrok, but it didn't seem to work with the Twitter Card Validator. I could also just push my forked repo to gh-pages or something...what do you recommend?

fk:
> I can only really validate the twitter cards when the code is deployed. I tried to use ngrok, but it didn't seem to work with the Twitter Card Validator.

💯 Yeah … hmm, ngrok _should_ work — I used it before to test with the Twitter Card Validator. Did you try it with `develop` or `build`? I think I was successful with the latter. Want to give it another chance?

Alternatively, you could build `www` with `GATSBY_SCREENSHOT_PLACEHOLDER=true` in `.env.production` (see https://github.com/gatsbyjs/gatsby/tree/master/www#running-slow-build-screenshots-placeholder — `GATSBY_SCREENSHOT_PLACEHOLDER=true` also works for prod), and upload to Netlify. But that means uploading 1,5 GB if I remember correctly. :(

rlynn523:
I only tried it with `develop` but will see if using `build` works. Thanks!

fk:
Sure! You're welcome! Crossing fingers! 🤞
Thanks for looking into this!

rlynn523:
<img width="458" alt="twitter-card" src="https://user-images.githubusercontent.com/18073651/59139043-95858f00-895e-11e9-955c-856f502a9ff5.png">

It worked! I ended up creating a `SocialMetaTags` component that will create `og` and `twitter` meta tags. I did have one question, if a site in the showcase does not have a description, is there any copy you would like to use as a fallback? Something with the title maybe? Let me know if this is what you were expecting! I did see that there was a ticket in to fix which image is used in the cards...I can update that here or it can be resolved later and passed as props to the new component.

rlynn523:
[...]

PR #14725, opened 2019-06-12
Ryan L. | updated SiteMetaData and Layout components to allow for customizing Twitter cards across the site

rlynn523:
## Description
While I was working on #14468, I checked other parts of the Gatsby site and noticed that Twitter cards were broken elsewhere. I thought that by updating the `SiteMetaData` component that is pulled in by `Layout`, props to customize Twitter cards could be passed to `Layout` and subsequently `SiteMetaData`. There are defaults in place so that Twitter cards should always be available. 
Before:
Screen Shot 2019-06-11 at 9 01 49 PM

After: 
Screen Shot 2019-06-11 at 9 08 01 PM
(an image size prop can be added to change the size of the image shown for Twitter cards

Showcase sites: 
Screen Shot 2019-06-11 at 7 37 42 PM

My hope is that this will eliminate the need to add the necessary meta tags with React Helmet, and instead pass props to `Layout`.

### DIFF
--- a/www/src/components/layout.js
+++ b/www/src/components/layout.js
@@ -154,7 +154,12 @@ class DefaultLayout extends React.Component {
 
     return (
       <>
-        <SiteMetadata pathname={this.props.location.pathname} />
+        <SiteMetadata
+          pathname={this.props.location.pathname}
+          metaDescription={this.props.metaDescription}
+          metaImage={this.props.metaImage}
+          metaTitle={this.props.metaTitle}
+        />
         <SkipNavLink css={skipLink}>Skip to main content</SkipNavLink>
         <Banner />
         <Navigation pathname={this.props.location.pathname} />
--- a/www/src/components/showcase-details.js
+++ b/www/src/components/showcase-details.js
@@ -144,6 +144,12 @@ const ShowcaseDetails = ({ parent, data, isModal, categories }) => (
 
       return (
         <Layout
+          metaDescription={data.sitesYaml.description}
+          metaImage={`https://www.gatsbyjs.org${
+            data.sitesYaml.childScreenshot.screenshotFile.childImageSharp.resize
+              .src
+          }`}
+          metaTitle={`${data.sitesYaml.title}: Showcase | GatsbyJS`}
           location={parent.props.location}
           isModal={isModal}
           modalBackgroundPath={parent.getExitLocation()}
@@ -252,25 +258,6 @@ const ShowcaseDetails = ({ parent, data, isModal, categories }) => (
             >
               <Helmet>
                 <title>{data.sitesYaml.title}: Showcase | GatsbyJS</title>
-                <meta
-                  property="og:image"
-                  content={`https://www.gatsbyjs.org${
-                    data.sitesYaml.childScreenshot.screenshotFile
-                      .childImageSharp.resize.src
-                  }`}
-                />
-                <meta
-                  name="twitter:image"
-                  content={`https://www.gatsbyjs.org${
-                    data.sitesYaml.childScreenshot.screenshotFile
-                      .childImageSharp.resize.src
-                  }`}
-                />
-                <meta name="twitter:card" content="summary_large_image" />
-                <meta
-                  name="og:title"
-                  value={`${data.sitesYaml.title}: Showcase | GatsbyJS`}
-                />
                 <meta
                   property="og:image:width"
                   content={
--- a/www/src/components/site-metadata.js
+++ b/www/src/components/site-metadata.js
@@ -4,7 +4,7 @@ import { graphql, useStaticQuery } from "gatsby"
 
 import gatsbyIcon from "../assets/gatsby-icon.png"
 
-const SiteMetadata = ({ pathname }) => {
+const SiteMetadata = props => {
   const {
     site: {
       siteMetadata: { siteUrl, title, twitter },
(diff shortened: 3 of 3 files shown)


www/src/components/site-metadata.js
```diff
@@ -21,28 +21,43 @@ const SiteMetadata = ({ pathname }) => {
...
+        <meta property="og:url" content={siteUrl} />
+        <meta property="og:type" content="website" />
+        <meta property="og:locale" content="en" />
+        <meta property="og:site_name" content={metaTitle} />
+        <meta property="og:image:width" content="512" />
+        <meta property="og:image:height" content="512" />
+        <meta name="twitter:card" content="summary_large_image" />
+        <meta name="twitter:site" content={twitter} />
+      </Helmet>
+      {generateMetaTags(`description`, metaDescription)}
```
DSchau: Maybe just iterate over these as additional `meta` tags within the above Helmet? (e.g. line 54 currently)

www/src/components/site-metadata.js
```diff
@@ -21,28 +21,43 @@ const SiteMetadata = ({ pathname }) => {
...
-
-      <meta property="og:url" content={siteUrl} />
-      <meta property="og:type" content="website" />
-      <meta property="og:locale" content="en" />
-      <meta property="og:site_name" content={title} />
-      <meta property="og:image" content={`${siteUrl}${gatsbyIcon}`} />
-      <meta property="og:image:width" content="512" />
-      <meta property="og:image:height" content="512" />
+  const {
+    metaImage = `${siteUrl}${gatsbyIcon}`,
```
DSchau: This should maybe (?) prepend siteUrl so that we can just pass the absolute path as metaImage and always prefix with the siteUrl.

www/src/components/site-metadata.js
```diff
@@ -21,28 +21,43 @@ const SiteMetadata = ({ pathname }) => {
...
+    metaTitle = title,
+    metaDescription = `${siteUrl}`,
```
DSchau: Probably something better we can use here. Seems weird for the description to be the siteUrl.

www/src/components/showcase-details.js
```diff
@@ -252,25 +258,6 @@ const ShowcaseDetails = ({ parent, data, isModal, categories }) => (
             >
               <Helmet>
                 <title>{data.sitesYaml.title}: Showcase | GatsbyJS</title>
-                <meta
-                  property="og:image"
-                  content={`https://www.gatsbyjs.org${
```
DSchau: In my comment below, you can see how this would be useful, e.g. no need to prefix with this hard coded string.

www/src/components/layout.js
```diff
@@ -154,7 +154,12 @@ class DefaultLayout extends React.Component {
 
     return (
       <>
-        <SiteMetadata pathname={this.props.location.pathname} />
+        <SiteMetadata
+          pathname={this.props.location.pathname}
+          metaDescription={this.props.metaDescription}
```
DSchau: Few comments:

I'd either use a `meta` array as a prop _or_ remove the prefixes (e.g. because it's SiteMetadata it should be obvious that these will eventually become metatags).

In other words:

```jsx
<SiteMetadata
  pathname={this.props.location.pathname}
  description={this.props.description}
  image={this.props.image}
/>
```

www/src/components/site-metadata.js
rlynn523: I think I had something using the title instead of the siteUrl, but it was suggested by @fk to maybe use the siteUrl. I can change it to whatever!

www/src/components/layout.js
rlynn523: Agreed. I think I originally had it without prefixes.

www/src/components/site-metadata.js
rlynn523: Gotcha, yea that would make sense to just always prefix `siteUrl` so that you don't have to hardcode `https://www.gatsbyjs.org` for every image

www/src/components/site-metadata.js
rlynn523: Are you referring to lines 53 and 54 as the additional `meta` tags? I'm not sure I follow what you are suggesting, could you please explain?

DSchau: Something like this (inside of the existing Helmet tags)

```jsx
{
  []
    .concat(generateMetaTags(`description`, metaDescription))
    .concat(generateMetaTags(`whatever`, metaDescription))
    .concat(generateMetaTags(`other`, metaDescription))
    .map(({ property, description }, index) => <meta property={property} description={description} key={`${index-${description`} />)
}
```

You'll need to tweak what generateMetaTags returns (or just add a key property and return the element when mapping).

rlynn523: Gotcha, I see what you are saying. I didn't realize that most of the `Twitter` specific meta tags are unnecessary as long as there are the `og` counter parts.

I liked the idea you had of just passing a `meta` prop...so I implemented that to `Layout` and subsequently `SiteMetaData`. It will now accept an array of objects: 
```
<Layout
  meta={[
    { property: 'description', content: 'test' },
    { property: 'image', content: 'imagePath' },
    { property: 'title', content: 'This is the title' },
   ]}
/>
```
Which now fits nicely inside the one `Helmet` wrapper in `SiteMetaData`

```
<Helmet>
  <html lang="en" />
  <link rel="canonical" href={`${siteUrl}${pathname}`} />
  <meta name="docsearch:version" content="2.0" />
  <meta
    name="viewport"
    content="width=device-width,initial-scale=1,shrink-to-fit=no,viewport-fit=cover"
  />
  <meta property="og:url" content={siteUrl} />
  <meta property="og:type" content="website" />
  <meta property="og:locale" content="en" />
  <meta property="og:site_name" content={title} />
  <meta property="og:image:width" content="512" />
  <meta property="og:image:height" content="512" />
  <meta name="twitter:card" content="summary_large_image" />
  <meta name="twitter:site" content={twitter} />
{
  meta.map(({ property, content }, index) => {
    return (
      <meta property={`og:${property}`} content={content} key={`${index}-${property}`}/>
    )
  })
}
</Helmet>
```

Thoughts?

DSchau: @rlynn523 I don't think the iteration is helpful now! I'd just manually create those as JSX, e.g.

```jsx
<Helmet>
  <html lang="en" />
  <link rel="canonical" href={`${siteUrl}${pathname}`} />
  <meta name="docsearch:version" content="2.0" />
  <meta
    name="viewport"
    content="width=device-width,initial-scale=1,shrink-to-fit=no,viewport-fit=cover"
  />
  <meta property="og:url" content={siteUrl} />
  <meta property="og:type" content="website" />
  <meta property="og:locale" content="en" />
  <meta property="og:site_name" content={title} />
  <meta property="og:image:width" content="512" />
  <meta property="og:image:height" content="512" />
  <meta name="twitter:card" content="summary_large_image" />
  <meta name="twitter:site" content={twitter} />
  <meta property=og:description" content={props.description || title} />
  {/* and the rest of them */}
```

rlynn523: I thought you were suggesting to use an iteration?

rlynn523: > I thought you were suggesting to use an iteration?
> 
> I can just put in JSX tags. I did notice that when navigating from the root of the Gatsby site to, for example, a showcase site, the meta props don't get updated, only when you hit the page directly. I haven't looked into to the rendering of the `Layout` component. We could remove those tags from `SiteMetaData`, make a `SocialMetaTags` component that can just be imported to each page that wants to add Twitter cards.

Hey @DSchau , just wanted to check in and see if there was any other feedback/updates that should be made to this in order to have it merged. Thanks!

PR #14726, opened 2019-06-12
feat(gatsby-remark-images): gatsby-remark-images supports `mediaFolder` option

RaoHai:
## Description

The post images are not always store at the same path as post, as in https://github.com/alxshelepenok/gatsby-starter-lumen, posts located in `content/posts` but  images located in `static/`.  The `gatsby-remark-images` plugin cannot locale image file then fail.

the `mediaFolder` option allow user to set the path of images, in `gatsby-starter-lumen`, set it to `${__dirname}/static/media` will solve the problem.


KyleAMathews:
This is a pretty big change to how this plugin (and all remark plugins) works. They all assume that assets are co-located with the markdown file.

Perhaps it'd be easier to change gatsby-starter-lumen than this plugin?

PR #14728, opened 2019-06-12
fix(gatsby): Fix group and distinct when used with fragments

freiksenet:

### DIFF
--- a/packages/gatsby/src/db/common/query.js
+++ b/packages/gatsby/src/db/common/query.js
@@ -12,7 +12,7 @@ const getQueryFields = ({ filter, sort, group, distinct }) => {
   }
 
   if (distinct && !Array.isArray(distinct)) {
-    distinct = [group]
+    distinct = [distinct]
   } else if (distinct == null) {
     distinct = []
   }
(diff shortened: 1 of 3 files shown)


packages/gatsby/src/db/common/query.js
```diff
@@ -4,11 +4,24 @@ const { getNamedType } = require(`graphql`)
...
   const sortFields = (sort && sort.fields) || []
+
+  if (group && !Array.isArray(group)) {
+    group = [group]
+  } else if (group == null) {
+    group = []
+  }
+
+  if (distinct && !Array.isArray(distinct)) {
+    distinct = [group]
```
stefanprobst:
```suggestion
    distinct = [distinct]
```

freiksenet: Good catch.

packages/gatsby/src/schema/resolvers.js
```diff
@@ -26,14 +26,14 @@ const findOne = typeName => ({ args, context, info }) =>
 
 const findManyPaginated = typeName => async rp => {
   // Peek into selection set and pass on the `field` arg of `group` and
-  // `distinct` which might need to be resolved. The easiest way to check if
+  // `distinct` which might need to be resolved. The easiest wayF to check if
```
wardpeet: typo
```suggestion
  // `distinct` which might need to be resolved. The easiest way to check if
```

Issue #14729, opened 2019-06-12
Using Gatsby in Storybook broken in version 2.9.0


joealden:
## Description

In a project I am working on, I am using Gatsby's `Link` component within Storybook stories. To get this working, I followed the instructions found [here in the Gatsby docs](https://www.gatsbyjs.org/docs/visual-testing-with-storybook/). This has been working fine for over 2 months now.

However, when updating Gatsby from `2.8.8` to `2.9.0`, Storybook fails to build because of a Gatsby 
`cache-dir` error. This is what is logged to the console:

```
ERROR in ./node_modules/gatsby/cache-dir/loader.js
Module not found: Error: Can't resolve './match-paths.json' in '/x/x/node_modules/gatsby/cache-dir'
 @ ./node_modules/gatsby/cache-dir/loader.js 15:0-44 97:32-42
 @ ./node_modules/gatsby/cache-dir/public-page-renderer-dev.js
 @ ./node_modules/gatsby/cache-dir/public-page-renderer.js
 @ ./node_modules/gatsby/cache-dir/gatsby-browser-entry.js
 @ ./src/utils/UniversalLink.tsx
 @ ./stories/misc/UniversalLink.stories.tsx
 @ ./stories sync \.stories\.tsx$
 @ ./.storybook/config.js
 @ multi ./node_modules/@storybook/core/dist/server/common/polyfills.js ./node_modules/@storybook/core/dist/server/preview/globals.js ./.storybook/config.js (webpack)-hot-middleware/client.js?reload=true
```

The important part of this error is the following:

```
ERROR in ./node_modules/gatsby/cache-dir/loader.js
Module not found: Error: Can't resolve './match-paths.json' in '/x/x/node_modules/gatsby/cache-dir'
```

From this, I could tell that the problem was originating from a change made in 2.9.0 to `cache-dir/loader.js`, so I went and looked through the update diff. I found the following lines added: 

https://github.com/gatsbyjs/gatsby/compare/gatsby@2.8.8...gatsby@2.9.0#diff-06ce5193de7c2c01dd947a5001c147c3R6

On L7 of the file, `./match-paths.json` is being imported. On L6, there is a helpful comment noting that this file is "generated during bootstrap" (which I'm guessing means that it is generated only when `gatsby dev` or `gatsby build` is ran).

[...]

PR #14732, opened 2019-06-12
fix(gatsby): Move importing of match-paths.json outside of loader.js

Moocar:
## Description

As raised in https://github.com/gatsbyjs/gatsby/issues/14729, Gatsby-link can be used independently of Gatsby itself. And it depends on `loader.js` to enqueue pages for prefetching. Gatsby prescribes [overriding those loader functions to be noops](https://www.gatsbyjs.org/docs/visual-testing-with-storybook/), but it still has to compile `loader.js`.

The problem is that the [per-page-manifest](https://github.com/gatsbyjs/gatsby/pull/13004) changes introduced a `match-paths.json` into the loader itself. Which won't compile if a gatsby build hasn't occured. To fix, I moved it out into production-app.js and app.js, and then dynamically set it. 

## Related Issues

- Fixes https://github.com/gatsbyjs/gatsby/issues/14729

### DIFF
--- a/packages/gatsby/cache-dir/app.js
+++ b/packages/gatsby/cache-dir/app.js
@@ -7,6 +7,7 @@ import emitter from "./emitter"
 import { apiRunner, apiRunnerAsync } from "./api-runner-browser"
 import loader, { setApiRunnerForLoader } from "./loader"
 import syncRequires from "./sync-requires"
+import matchPaths from "./match-paths.json"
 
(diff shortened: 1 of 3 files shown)


beaucharman:
Awesome, thank you! we are getting the same issue

wardpeet:
Published in `gatsby@2.9.3` :tada:

PR #14739, opened 2019-06-12
Fix typography plugin not working

joelzanden:
Add gatsby-plugin-typography and some curly brackets before the resolve part


wardpeet:

joelzanden:
Hi. I actually think it fixes a bug. Copy-and-pasting from the tutorial doesn't work because it seems the name of the plugin is missing from line 3 in the gatsby-config.js file.

wardpeet:
The `gatsby-plugin-typography` is inside the resolve option.
```js
{
  resolve: `gatsby-plugin-typography`,
  options: {},
},
```

 If the tutorial isn't working, it has to be something. Could you create an issue instead?

Issue #14752, opened 2019-06-12
Gatsby-plugin-sharp generates source sets bigger than maxWidth parameter.


bigfanjs:
## Summary
`Gatsby-plugin-sharp` seems like it generates source sets(`srcset`) bigger than what specified in `maxWidth` parameter.

that is mentioned in `gatsby-transformer-sharp` docs:
> If the max width of the container for the rendered markdown file is 800px, the fluid sizes would then be: 200, 400, 800, 1200, 1600, 2400 – enough to provide close to the optimal image size for every device size / screen resolution.

Ref: https://image-processing.gatsbyjs.org/

## Relevant information
my query:
```
{
  contentfulArticle {
    id
    title
    image {
      fluid(maxWidth: 2080) {
        srcSet
      }
    }
  }
}
```

which generates a `srcset` like:
```
//images.ctfassets.net/image-url.jpg?w=520&h=347&q=50 520w,
//images.ctfassets.net/image-url.jpg?w=1040&h=693&q=50 1040w,
//images.ctfassets.net/image-url.jpg?w=2080&h=1387&q=50 2080w,
//images.ctfassets.net/image-url.jpg?w=3120&h=2080&q=50 3120w,
//images.ctfassets.net/image-url.jpg?w=4160&h=2773&q=50 4160w,     <--- this wont work
//images.ctfassets.net/image-url.jpg?w=5184&h=3456&q=50 5184w     <--- this wont work
```

Those images are coming from Contentful, and Contentful does not allow fetching images above `4000` in size. And the issue is that in large screens, the images would fail to load. because the browser would select the last source (`//images.ctfassets.net/image-url.jpg?w=5184&h=3456&q=50 5184w`) in large screens which would fail to load because the width specified as `5184`.

On Contentful docs:
>Specify width & height
You can resize the image to the desired width and height. The maximum allowed value is 4000 pixels. The default is the original image width and height.

Is there a work around to this issue, like preventing the plugin from generating image sources that go above a certain width. or deciding the image size for each breakpoint?

KyleAMathews:
This would be something to fix in gatsby-source-contentful as since they have their own image api, gatsby-plugin-sharp doesn't process images from Contentful. Would you like to add a PR restricting width to 4000px here https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-source-contentful/src/extend-node-type.js ?

bigfanjs:
[...]

PR #14760, opened 2019-06-13
fix(gatsby-source-contentful): Restricts images to 4000 max image width/height

bigfanjs:
## Description
Since Contentful does not allow fetching images above `4000` in size, this PR restricts images to not go above that maximum size to prevent images from not even loading in large screens.

## Related Issues
Related to #14752

### DIFF
--- a/packages/gatsby-source-contentful/src/extend-node-type.js
+++ b/packages/gatsby-source-contentful/src/extend-node-type.js
@@ -20,6 +20,7 @@ const {
   ImageCropFocusType,
 } = require(`./schemes`)
 
+// @see https://www.contentful.com/developers/docs/references/images-api/#/reference/resizing-&-cropping/specify-width-&-height
 const CONTENTFUL_IMAGE_MAX_SIZE = 4000
 
 const isImage = image =>
--- a/packages/gatsby-source-contentful/src/extend-node-type.js
+++ b/packages/gatsby-source-contentful/src/extend-node-type.js
@@ -20,7 +20,7 @@ const {
   ImageCropFocusType,
 } = require(`./schemes`)
 
-const contentful_max_size = 4000
+const CONTENTFUL_IMAGE_MAX_SIZE = 4000
 
 const isImage = image =>
   _.includes(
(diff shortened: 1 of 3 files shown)


packages/gatsby-source-contentful/src/extend-node-type.js
```diff
@@ -114,7 +116,9 @@ const resolveFixed = (image, options) => {
   fixedSizes = fixedSizes.map(Math.round)
 
   // Filter out sizes larger than the image's width.
-  const filteredSizes = fixedSizes.filter(size => size <= width)
+  const filteredSizes = fixedSizes.filter(
+    size => size <= contentful_max_size || size <= width
```
wardpeet: I've been testing this condition and I think you want `&&` as we want to filter out images that are bigger than 4000 or bigger than the image width. By changing this the test will fail but that's a good thing :wink: 

```suggestion
    size => size <= contentful_max_size && size <= width
```

packages/gatsby-source-contentful/src/extend-node-type.js
```diff
@@ -215,7 +219,9 @@ const resolveFluid = (image, options) => {
   fluidSizes = fluidSizes.map(Math.round)
 
   // Filter out sizes larger than the image's maxWidth.
-  const filteredSizes = fluidSizes.filter(size => size <= width)
+  const filteredSizes = fluidSizes.filter(
+    size => size <= contentful_max_size || size <= width
```
wardpeet: This line has the same issue as above but there is a catch 😂

A few lines below we have the following line of code
https://github.com/gatsbyjs/gatsby/blob/ca4f15af9a68302a78436af60801e973b9907416/packages/gatsby-source-contentful/src/extend-node-type.js#L228

This one fights with the contentful_max_size constant. If the image is bigger than contentful_max_size we still add it.

So I suggest changing that line to
`if (!filteredSizes.includes(parseInt(width)) && parseInt(width) < contentful_max_size) filteredSizes.push(width)`

packages/gatsby-source-contentful/src/extend-node-type.js
```diff
@@ -20,6 +20,8 @@ const {
   ImageCropFocusType,
 } = require(`./schemes`)
 
+const contentful_max_size = 4000
```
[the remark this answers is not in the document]
bigfanjs: Yeah sure I like this too.

packages/gatsby-source-contentful/src/extend-node-type.js
bigfanjs: Oops! What a mistake!

packages/gatsby-source-contentful/src/extend-node-type.js
bigfanjs: aah nice!!

sidharthachatterjee:
Published in `gatsby-source-contentful@2.0.72`

Issue #14564, opened 2019-06-05
[docs] new reference guide on contact forms in Gatsby


marcysutton:
## Summary

A [common question](https://github.com/gatsbyjs/gatsby/issues/12022) in Gatsby is how to build a functioning contact form (it is also a [key workflow](https://github.com/gatsbyjs/gatsby/issues/13708)). Without a server runtime environment, it isn't exactly intuitive how to hook up a form action to submit data. 

The first step to improving that workflow would be to create a [reference guide](https://www.gatsbyjs.org/docs/guides/) in the docs; reference guides are technical documentation outlining the various ways to build things in Gatsby (as opposed to tutorials, which are step-by-step instructions). Check out the [docs templates](https://www.gatsbyjs.org/contributing/docs-templates/) and [Gatsby Style Guide[(https://www.gatsbyjs.org/contributing/gatsby-style-guide/) for more information.

Here are some of the potential points to include in such a guide:

- Link to the [Adding Forms](https://www.gatsbyjs.org/docs/adding-forms/) doc
- Briefly mention [creating accessible forms](https://webaim.org/techniques/forms/) because it's a common barrier on the web, and we should take every opportunity to educate
- Use cloud/serverless functions for form action, like [Netlify](https://www.netlify.com/docs/form-handling/), [AWS Lambda](https://aws.amazon.com/blogs/architecture/create-dynamic-contact-forms-for-s3-static-websites-using-aws-lambda-amazon-api-gateway-and-amazon-ses/) or Azure
    - This could also mention Netlify's honeypot feature for spam prevention
- Stand up your own API server, i.e. with Express or similar
- Try services like FormSpree (be sure to test to make sure it works for submissions!)

This could be added to the docs sidebar under "Adding Website Functionality", under Adding Forms. The title of this one should differentiate that it is about hooking up the contact form, more than building out the markup; something like "Building a Contact Form" perhaps?

This is up for grabs if anyone wants to take it on!

dyyyl:
I can try taking this one up!

marcysutton:
@dyyyl awesome. Let us know if you have any questions!

dyyyl:
Hey @marcysutton! I'm getting close to PR time, but I'm struggling to add the new article to the sidenav. Can you please point me in the direction on how that's managed in the repo? Thanks!

dyyyl:
[...]

PR #14768, opened 2019-06-13
docs(www): add Build a Contact Form reference guide (#14564)

dyyyl:
## Description

Adds a reference guide for building contact forms in Gatsby!

Has sections regarding:
	- Adding a Contact Form
	- Creating an Accessible Form
	- Sending Form Data
	  - with Netlify
	  - with Formspree
	  - Running Your Own Server
	- Other Resources (a cool Scott Tolinski tutorial)

## Related Issues

Addresses #14564 🚀


docs/docs/building-a-contact-form.md
```diff
@@ -0,0 +1,140 @@
...
+  <button type="submit">Send</button>
+  <input type="reset" value="Clear" />
+</form>
+```
+
+### Netlify
+
+If you're hosting your site with Netlify, you gain access to their excellent form handling feature. Setting this up only invloves adding a few form attributes:
+
+```jsx:title=src/pages/contact.js
```
gillkyle: This could be a good place to use a `diff` file type to show what lines have changed from the example above and new code here.

I also had trouble trying to make this work, I put the default starter on Netlify with this code and I don't actually get submissions to go through: https://amazing-roentgen-d0c1a6.netlify.com/netlify/ I also looked at Netlify's example that worked okay, but it uses plenty of JavaScript even though their docs say you just need to add a few html attributes like this.

Do you have an example where you got this working by chance?

www/src/data/sidebars/doc-links.yaml
```diff
@@ -277,6 +277,8 @@
               link: /docs/building-a-site-with-authentication/
             - title: Adding Forms
               link: /docs/adding-forms/
+            - title: Building a Contact Form
```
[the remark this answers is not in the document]
marcysutton: I think this is fine for now actually (and simpler), since there aren't any plans for other form guides at the moment. We could always make it a sub-item later on, since it doesn't impact the URL.

docs/docs/building-a-contact-form.md
```diff
@@ -0,0 +1,140 @@
...
+    <textarea name="message" id="message" rows="5" />
+  </label>
+  <button type="submit">Send</button>
+  <input type="reset" value="Clear" />
+</form>
+```
+
+### Netlify
+
+If you're hosting your site with Netlify, you gain access to their excellent form handling feature. Setting this up only invloves adding a few form attributes:
```
marcysutton:
```suggestion
If you're hosting your site with Netlify, you gain access to their excellent [form handling feature](https://www.netlify.com/docs/form-handling/). Setting this up only involves adding a few form attributes:
```

docs/docs/building-a-contact-form.md
marcysutton: I have it working on my site: https://github.com/marcysutton/gatsby-site/blob/master/src/pages/contact.js

docs/docs/building-a-contact-form.md
```diff
@@ -0,0 +1,140 @@
...
+
+```jsx:title=src/pages/contact.js
+<form method="post" action="my-project-abcd123.now.sh/contact">
+  ...
+</form>
+```
+
+Now, all subsequent form submissions will be sent to your email address!
+
+For an in-depth guide on running your own mail server, you can refer to [this awesome guide by DataFire](https://medium.com/datafire-io/simple-backends-four-ways-to-implement-a-contact-us-form-on-a-static-website-10fc430984a4).
```
marcysutton: Are there any security/spam implications we should mention here?

dyyyl:
Hey @marcysutton @gillkyle, thank you for all of your input & patience! I've made a commit addressing the proposed changes, can't wait to see what you think ✨

docs/docs/building-a-contact-form.md
```diff
@@ -0,0 +1,148 @@
...
+  </label>
+  <label>
+    Message
+    <textarea name="message" id="message" rows="5" />
+  </label>
+  <button type="submit">Send</button>
+  <input type="reset" value="Clear" />
+</form>
+```
+
```
marcysutton: I feel like these vendor options need another heading to transition from the conceptual example to the practical how-tos. How about something like:

```## Form submission options in Gatsby```

dyyyl: Looks good, reflected in the latest commit!

docs/docs/building-a-contact-form.md
marcysutton: I found an article that goes into some of the detail of complexity and spam implications, though it isn't messaged from the perspective of a static site: https://www.digitalocean.com/community/tutorials/why-you-may-not-want-to-run-your-own-mail-server

PR #14770, opened 2019-06-13
docs(browser-support): Update Browser Support Docs

ARXChrono:
## Description

Add another browser support solution available (via gatsby plugin) be known in the docs. This is a solution I've personally used that works which is fairly easy to customize.

## Related Issues
NA


DSchau:
Thanks for this!

Unfortunately, I don't think this is a particularly good place for this change. While polyfill.io is _great_ I'd rather we keep this document focused on internal Gatsby browser support rather than relying on an external polyfilling service (which could go down, is tangentially related to core Gatsby stuff, etc.).

This is to keep this document as accessible, clear, and direct for as many people as possible.

Going to close this--but thank you for opening, and we hope to see more from you soon!

Thanks!

Issue #14418, opened 2019-05-30
Building JS and CSS bundles for 5 pages takes 88.374s


mikheevm:
## Description
After updating `gatsby-*`dependencies to the latest versions the cold build started to take really long compared to what it was before (like 4x or even more). After the cold one passes, succeeding ones will take 20-25 secs.
Approximately it became so after this commit: https://github.com/gnosis/dappcon-website/commit/56be14cc789c16355d81b2cc4be0d181adabc789

## BUILD LOG
```
$ npm run clean && gatsby build
npm WARN lifecycle The node binary used for scripts is /tmp/yarn--1559199574589-0.9665285201595042/node but npm is using /usr/bin/node itself. Use the `--scripts-prepend-node-path` option to include the path for the node binary npm was executed with.
> rimraf .cache public
success open and validate gatsby-configs - 0.013 s
success load plugins - 0.483 s
success onPreInit - 0.012 s
success delete html and css files from previous builds - 0.006 s
success initialize cache - 0.009 s
success copy gatsby files - 0.029 s
success onPreBootstrap - 0.007 s
success source and transform nodes - 0.163 s
success building schema - 0.244 s
success createPages - 0.074 s
success createPagesStatefully - 0.036 s
success onPreExtractQueries - 0.005 s
success update schema - 0.037 s
success extract queries from components - 0.179 s
success run static queries - 0.135 s — 2/2 15.57 queries/second
success run page queries - 0.552 s — 7/7 12.86 queries/second
success write out page data - 0.004 s
success write out redirect data - 0.003 s
success onPostBootstrap - 0.008 s
info bootstrap finished - 8.466 s
success Building production JavaScript and CSS bundles - 88.374 s
success Building static HTML for pages - 2.633 s — 7/7 11.08 pages/second
info Done building in 99.507033731 sec
Done in 100.05s.
```

### Steps to reproduce

- Clone this repo: https://github.com/gnosis/dappcon-website
- Install the dependencies with `yarn`
- Run `yarn build` 

### Expected result

The project has only 5 static pages, so I expect it to build production bundle in less than 100 seconds.

### Actual result

The build takes 100 seconds approximately. I could live with it, but on the CI environment it makes the build fail with error:
`FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory`

### Additional info
Sometimes I also get this error when trying to build:
```
error GraphQL Error Field "image" must not have a selection since type "String" has no subfields.

[...]

PR #14783, opened 2019-06-14
fix(gatsby-plugin-netlify-cms): Fix minimizer for production builds with gatsby-plugins-netlify-cms

emanueleperuffo:
## Description

When building for production with gatsby-plugin-netlify-cms, node hangs at 'Building production JavaScript and CSS bundles', eventually throwing an out of memory error.

I found out that the problem is with Webpack 'minimize'. 

Using the minimizer setting from the original Webpack configuration of Gatsby fixes the issue.

## Related Issues

Possibly related to and fixes https://github.com/gatsbyjs/gatsby/issues/14418


jcppman:
this PR breaks my `yarn start` because webpack config does not have the property `optimization` at all.

It will cause `TypeError: Canno read property 'minimizer' of undefined`.

PR #14792, opened 2019-06-14
chore(gatsby): upgrade webpack

### LINKED ISSUE #15775: upgrade webpack version
## Description

Gatsby fix webpack version at `~4.28.4` in [this commit](https://github.com/gatsbyjs/gatsby/commit/f460b7f59f3e2a71c9ca65276a3e2f0d2c9a7fef), because [there is an regression in webpack 4.29](https://github.com/webpack/webpack/issues/8656).
But now that regression is fixed. It should be ok to use `4.35.x` or later.

Some newer plugin expect [webpack's finishModules hook](https://github.com/webpack/webpack/blob/37d4480382d362a6f44325f13b2f728ef6f1bfb1/lib/Compilation.js#L242) to be `AsyncSeriesHook`. But in webpack~4.28.4, [it is a SyncHook](https://github.com/webpack/webpack/blob/fdb6b139b94273e5b18870262c193a93f65132e8/lib/Compilation.js#L242), which will complain `tapPromise is not supported on a SyncHook` and cause a build failure.

### Expected result

We should use newer webpack version.

### Actual result

Webpack version is fixed at `4.28.x`.

### Environment

all

wardpeet:
## Description
Upgrades webpack to the latest again so we can remove the mjs hack. It got fixed by webpack upgrading to acorn@^6.

I've linked an issue below that seems to have issues with webpack 4.29.0 on npm. I've tried reproduce it with their repo's and npm versions but sadly no way for me to reproduce.

To let people test I created a gatsby version on npm. `npm install gatsby@webpack-acorn`

## Related Issues
https://github.com/gatsbyjs/gatsby/issues/11198
fixes #15775

### DIFF
--- a/packages/gatsby/package.json
+++ b/packages/gatsby/package.json
@@ -127,7 +127,7 @@
     "util.promisify": "^1.0.0",
     "uuid": "^3.1.0",
     "v8-compile-cache": "^1.1.0",
-    "webpack": "~4.28.4",
+    "webpack": "^4.29.0",
     "webpack-dev-middleware": "^3.0.1",
     "webpack-dev-server": "^3.1.14",
     "webpack-hot-middleware": "^2.21.0",
--- a/packages/gatsby/package.json
+++ b/packages/gatsby/package.json
@@ -127,7 +127,7 @@
     "util.promisify": "^1.0.0",
     "uuid": "^3.1.0",
     "v8-compile-cache": "^1.1.0",
-    "webpack": "^4.29.0",
+    "webpack": "^4.35.3",
     "webpack-dev-middleware": "^3.0.1",
     "webpack-dev-server": "^3.1.14",
     "webpack-hot-middleware": "^2.21.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -3426,11 +3426,6 @@ acorn-dynamic-import@^3.0.0:
   dependencies:
     acorn "^5.0.0"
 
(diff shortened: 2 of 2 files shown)


freiksenet:
Did you try it with developrunner?

DSchau:
Marking as blocked until we fix [this comment](https://github.com/gatsbyjs/gatsby/issues/11198#issuecomment-502443775)

Also @wardpeet you'll want to remove the `.mjs` tests for webpack.config :)

wardpeet:
This is currently blocked by `@pieh/friendly-errors-webpack-plugin` on npm. 

```
There was a problem loading the local build command. Gatsby may not be installed. Perhaps you need to run "npm install"? Cannot find module 'webpack/lib/RequestShortener'



  Error: Cannot find module 'webpack/lib/RequestShortener'
```

This is fixed in `friendly-errors-webpack-plugin@v2.0.0-beta.2` still in beta.

pieh:
> This is currently blocked by `@pieh/friendly-errors-webpack-plugin` on npm.

I have to check but maybe we don't need to use my fork anymore

wardpeet:
@pieh we do, unless we are ok with going for a beta package?

pieh:
> @pieh we do, unless we are ok with going for a beta package? Your pr got merged but is only published in the beta package

Is the actual error:
```
  Error: Cannot find module 'webpack/lib/RequestShortener'
```
fixed there as well?

PR #14803, opened 2019-06-15
feat(docs): image and recipe docs updates

marcysutton:
## Description

Two small changes we need for workflows:

- a note about not necessarily needing GraphQL or Gatsby Image on the new API doc
- instructions for adding new recipe categories


marcysutton:
Thanks, Mike!

PR #14812, opened 2019-06-16
chore(showcase): Remove McDonald's Design System site

lannonbr:
## Description

It has changed to using Next.js so I removed it from the showcase


lannonbr:
pinging this as it should be ready to merge but the CI isn't running so it may be good to just merge it

PR #14818, opened 2019-06-16
chore: use faster `existsSync`

ZYSzys:
## Description

`fs-exists-cached` is faster than native `fs.existsSync`, so use `fs-exists-cached` instead.

```js
const Benchmark = require('benchmark');

const { existsSync } = require('fs');
const { sync } = require('fs-exists-cached');

const suite = new Benchmark.Suite;

// add tests
suite.add('Native fs', function() {
  existsSync(__filename);
  existsSync('./package.json');
})
.add('fs-exists-cached', function() {
  sync(__filename);
  sync('./package.json');
})
// add listeners
.on('cycle', function(event) {
  console.log(String(event.target));
})
.on('complete', function() {
  console.log('Fastest is ' + this.filter('fastest').map('name'));
})
// run async
.run({ 'async': true });

// logs:
// => Native fs x 134,342 ops/sec ±3.16% (79 runs sampled)
// => fs-exists-cached x 17,560,674 ops/sec ±5.86% (77 runs sampled)
// -> Fastest is fs-exists-cached
```

## Related Issues
None.

### DIFF
--- a/packages/gatsby-plugin-sharp/package.json
+++ b/packages/gatsby-plugin-sharp/package.json
@@ -10,6 +10,7 @@
     "@babel/runtime": "^7.0.0",
     "async": "^2.1.2",
     "bluebird": "^3.5.0",
+    "fs-exists-cached": "^1.0.0",
     "fs-extra": "^7.0.0",
     "imagemin": "^6.0.0",
     "imagemin-mozjpeg": "^8.0.0",
--- a/packages/gatsby-plugin-sharp/src/safe-sharp.js
+++ b/packages/gatsby-plugin-sharp/src/safe-sharp.js
@@ -11,7 +11,7 @@
 
 const childProcess = require(`child_process`)
 const path = require(`path`)
-const fs = require(`fs`)
+const existsSync = require(`fs-exists-cached`).sync
 const semver = require(`semver`)
 
 const originalConsoleError = console.error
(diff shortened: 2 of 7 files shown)


freiksenet:
Thank you for your PR @ZYSzys!

Most of this functions don't seem to be called more than couple of times during a gatsby build lifecycle, I'm not sure it's worth using another library if the savings aren't significant. Did you do any  tests that shows that this meaningfully improves Gatsby build performance?

ZYSzys:
> Did you do any tests that shows that this meaningfully improves Gatsby build performance?

Not yet. But one thing I'm sure is that our `gatsby-cli` has using this library:

```sh
 ~/Projects/gatsby/packages/gatsby-cli rg existsSync                                                                                                      [15:31:34]
src/init-starter.js
10:const existsSync = require(`fs-exists-cached`).sync
69:  if (existsSync(sysPath.join(rootPath, `.gitignore`))) {
120:  if (!existsSync(starterPath)) {
206:  if (existsSync(sysPath.join(rootPath, `package.json`))) {

src/create-cli.js
7:const existsSync = require(`fs-exists-cached`).sync
33:  const useYarn = existsSync(path.join(directory, `yarn.lock`))
```
So at least there should be some minor improvements right ?

sidharthachatterjee:
Thank you for opening this @ZYSzys 

While this was great work on your part, we're wary of merging this in because:

1. As @freiksenet mentioned, this looks like a micro optimisation and doesn't look seem to be a significant real world improvement 
2. Merging this in would result in us depending on one more dependency (`fs-exists-cached`) being maintained in the future 

If your benchmarks show a significant real world improvement, feel free to reopen! Thank you again 💜 💪

PR #14824, opened 2019-06-16
chore(starters): add gatsby-typescript-starter-default

RobertoMSousa:
## Description
There are no starters using eslint and prettier with typescript and since officially since the tslint will be deprecated it would be good for devs to start using eslint instead.


RobertoMSousa:
> Thank you for this contribution. Please add new entries to the end of file, not to the beginning. This simplifies diffs.

Thanks for your review. Just did the required changes.

Issue #12382, opened 2019-03-07
[gatsby-source-shopify] Fallback to cache with warning if API not reachable when developing


andru:
## Summary

Currently, source-shopify will cause `gatsby develop` to fail if the shopify API is not reachable, with the message:

`FetchError: request to https://[api-url]/api/graphql failed, reason: getaddrinfo ENOTFOUND`

### Basic example

I'm not totally familiar with the structure of source plugins, but it would be great if data fetched from the API is cached and used as a fallback during `gatsby develop` bootstrap if the shopify API is unreachable.

Even better would be if this could be passed as an option to `gatsby develop` or added to the config to actively avoid hitting up the live API. 

At least personally I find that most of the time during development, having the most up-to-date data from Shopify is not important, and the time it takes to hit the APIs would shave a good 10s from the gatsby dev bootstrap.

Having read some similar issues related to `source-wordpress`, this might be an issue of `gatsby develop` overly aggressively clearing cache and not something that source plugins have any control over, but - a bit ignorant as I am of the internals - I'm hopeful that's not the case.

### Motivation

Working on trains, planes, etc, or anywhere where network access may be unavailable or flaky. At present `gatsby-source-shopify` makes it impossible, unless there's an undocumented API or something I'm missing

DSchau:
Agreed!

We could follow a similar approach to how we've implemented it in [gatsby-source-contentful](https://www.gatsbyjs.org/packages/gatsby-source-contentful/#offline) which would be a nice feature addition to this plugin.

Happy to take PRs here, or offer any guidance!

andru:
I'd like to keep this open. It's a feature I'd be interested to contribute a PR to when I've got some time for it.

PR #14825, opened 2019-06-16
[gatsby-source-shopify] Add support for dev builds offline

andru:
## Description

This pull request is a first draft to introduce a check for network connectivity before attempting to make any requests to the Shopify API. Where the network is offline, it will attempt to use cached data from a previous build. 

Since it's rarely necessary to have the most recent data from Shopify every time `gatsby develop` runs, the environment variable `GATSBY_SHOPIFY_OFFLINE` can be set to `true` to force the use of cached data.

I mostly just copied the approach that `gatsby-source-contentful` takes in this regard.

In both cases, if there's no cached data available this will either fail silently and not generate any nodes, or error somewhat cryptically about missing nodes - e.g. if there are other nodes linking to the nodes created by this plugin. 

This does output a warning to the user ahead of time that when it's using the cache, but I welcome any suggestions on how this could detect an invalidated cache ahead of time to ignore the `GATSBY_SHOPIFY_OFFLINE` env where there's a network connection present, or provide a clearer error if there's no cache and no network connection.

## Related Issues

Addresses #12382

### DIFF
--- a/packages/gatsby-source-shopify/src/gatsby-node.js
+++ b/packages/gatsby-source-shopify/src/gatsby-node.js
@@ -1,6 +1,8 @@
 import { pipe } from "lodash/fp"
 import chalk from "chalk"
 import { forEach } from "p-iteration"
+import isOnline from "is-online"
+
 import { createClient, printGraphQLError, queryAll, queryOnce } from "./lib"
 import {
   ArticleNode,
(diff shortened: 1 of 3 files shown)


andru:
@DSchau just pinging you incase you don't automatically get notified about this.

packages/gatsby-source-shopify/src/gatsby-node.js
```diff
@@ -25,20 +27,54 @@ import {
...
+          touchNode({ nodeId: node.id })
+          // check for image nodes and touch them to persist in cache
+          if (node.image) {
+            touchNode({ nodeId: node.image.localFile___NODE })
+          }
+          if (node.images && node.images.edges) {
+            node.images.edges.map(edge => {
+              touchNode({ nodeId: edge.node.id })
+            })
+          }
```
pieh: This is fragile. Ideally we don't target `image` and `images`/`images.edges` here and we have more abstract way of traversing node to touch linked nodes. This might break later on when we introduce changes to other parts of shopify plugin code and this can go unnoticed

packages/gatsby-source-shopify/src/gatsby-node.js
```diff
@@ -25,20 +27,54 @@ import {
...
-    console.log(formatMsg(`starting to fetch data from Shopify`))
+    const online = await isOnline()
+    // if the user sets the GATSBY_SHOPIFY_OFFLINE environment variable
+    // force a fall-back to cached nodes, but never in production
+    const preferCache =
+      process.env.GATSBY_SHOPIFY_OFFLINE === `true` &&
+      process.env.NODE_ENV !== `production`
+
+    if (!online || preferCache) {
+      console.log(chalk`{bold.red Using offline cache for Shopify} ⚠️`)
```
pieh: It would be great if we would also show timestamp of last fetch.

wardpeet:
@andru any chance you can get back to this one? 😄

andru:
Thanks for the nudge! I’ve been super busy but I’ll try to get back to this
later in the month when my schedule frees up a bit.

packages/gatsby-source-shopify/src/gatsby-node.js
andru: Can you expand a little on what you mean by "a more abstract way of traversing node to touch linked nodes"? What other linked nodes are there I should be aware of? Any tips on how I can implement this in a way which ensures all linked nodes are touched?

packages/gatsby-source-shopify/src/gatsby-node.js
andru: Agreed, this would be really handy. I'm not sure how I'd go about this though. Is there somewhere that you know of where the timestamp of the last fetch is stored, or some easy way for me to add it?

andru:
@pieh just giving you a nudge, hoping you could give me some guidance on the above comments so I can proceed with changes in-line with your suggestions.

packages/gatsby-source-shopify/src/gatsby-node.js
pieh: We don't really store that information as gatsby doesn't have built in offline capabilities right now. But you can store this yourself using something like
```
cache.set(`last-fetch`, new Date().toJSON())
```

and retrieve it with
```
cache.get(`last-fetch`)
```

packages/gatsby-source-shopify/src/gatsby-node.js
pieh: > What other linked nodes are there I should be aware of? 

Yeah, the point is that we don't know and we should not make any assumptions that we know. It's very easy to desync other parts of code with feature like this that might not be used very often (and easy to break them).

> Any tips on how I can implement this in a way which ensures all linked nodes are touched?

I think if we do deep traverse of node, we can look for fields that end in `___NODE` (that's how we link nodes) and touch nodes that are linked:

```js
// pseudo-code

import { eachDeep } from 'deepdash-es/standalone';

eachDeep(node, (value, key) => {
  if (key.endsWith(`___NODE`)) {
    // can be either single node or array
	if (Array.isArray(value)) {
      value.forEach(linkedNodeId => touchNode({ nodeId: linkedNodeId }))
    } else {
      touchNode({ nodeId: value })
    }
  }
})
```

pieh:
> @pieh just giving you a nudge, hoping you could give me some guidance on the above comments so I can proceed with changes in-line with your suggestions.

Sorry for delay, just added some comments

LekoArts:
Hi @andru 	, a friendly nudge from us 😊 	Did you find time to work on this? Thanks a lot!

PR #14832, opened 2019-06-17
chore(docs): update Edit Tags-Category gatsby-node.js sample code

polishedwp:
## Description

Four months ago, as a new Gatsby user with no React/JS experience, I struggled to work through this tutorial [Adding Tags and Categories to Blog Posts](https://www.gatsbyjs.org/docs/adding-tags-and-categories-to-blog-posts/). I have since, successfully completed, and I think the struggle for a new user is the language associated with `gatsby-node.js`. 

 The paragraph before the code block reads `... make sure your createPages looks something like this`, but the reference to the markdown `slug` creation does not reflect this tutorial. The present code block reads `node.frontmatter.path`, and since the tutorial actually uses `node.fields.slug`, this PR makes this change for clarity.


marcysutton:
Thanks, @polishedwp! :taco:

Issue #14772, opened 2019-06-13
website & blog needs css background: #000


tomByrer:
After changing my Firefox's default background to dark, some of Gatsby's website & all the blog articles are unreadable.  Seems the font color everywhere is fixed, but assumed incorrectly everyone's browser background would be white

![gatsbyjs_org_firefox_background_color](https://user-images.githubusercontent.com/1308419/59440369-79e61280-8dbc-11e9-9253-e40082f90480.JPG)

I have not tested in other browsers, but I assume it would be same issue.

### Expected Action

Ensure CSS `background: #000` is added, instead of assuming that is the default, where needed.  Which is not everywhere, since much of the site has a light-purple background..

#### Advanced Action

Not sure if you use a Web Accessibility checker, but would be nice to add a rule there or a visual checker to ensure any additional components / templates have their CSS `background` set.

### Background of Issue

I have a head concussion, so extended bright lights are bothersome.  I don't expect websites to all have a 'dark theme' yet, so I was hoping to circumvent this by changing the default color & background colors & use openstyles / stylus to hack frequently visited websites.

thanks for your work!

mokimo:
Looks like a good suggestion, are you up for making a PR?
It should be somewhere around here: https://github.com/gatsbyjs/gatsby/blob/master/themes/

tomByrer:
Sure, I could poke around abit.

fk:
Hey @tomByrer! Thanks for brining this up, and sorry to hear that this is causing problems for you!

I just added #14838, which sets the default `html`(page) background for gatsbyjs.org to `#fff`/`white`. I _assume_ this is what you were after—I was confused that you mentioned `#000`, which is the (shorthand) HEX code for black, not white …

With #14838 merged and deployed, gatsbyjs.org will display "as designed"—with a white page background—no matter what your browser's default background is set to.

To address your issue in regards to gatsbyjs.org, maybe you want to give a browser extension like https://addons.mozilla.org/de/firefox/addon/dark-background-light-text/ a try? It seems to produce acceptable results:

![image](https://user-images.githubusercontent.com/21834/59619681-6247ba80-912b-11e9-8e53-4a20f20db9d6.png)

Let us know if this was helpful please! 🙏

fk:
@mokimo The source for gatsbyjs.org is located at https://github.com/gatsbyjs/gatsby/tree/master/www 😉—the `themes` directory is for Gatsby Themes, which you can read more about here:

[...]

PR #14838, opened 2019-06-17
fix(www): Set `background-color: white` for `html`

fk:
This sets `#fff` as the default background for `html`, and thus prevents the user-defined default background color (see e.g. [here](https://support.mozilla.org/en-US/kb/change-fonts-and-colors-websites-use) on how to set this in Firefox) from interfering with the design of gatsbyjs.org.

Context: #14772

---

Before: 

image

After: 

image


sidharthachatterjee:
that neon green tho

fk:
😎

PR #14840, opened 2019-06-17
chore(starters): add gatsby-starter-quiet

zhouyuexie:
## Starter

Add new starter `gatsby-starter-quiet`

### DIFF
--- a/docs/starters.yml
+++ b/docs/starters.yml
@@ -3033,3 +3033,20 @@
     - Includes plugins for offline support out of the box
     - TypeScript
     - Prettier & eslint to format & check the code
+- url: https://quietboy.net
+  repo: https://github.com/zhouyuexie/gatsby-starter-quiet
+  description: Gatsby out of the box blog, use TypeScript and highly customized style.
+  tags:
+    - TypeScript
+    - SwitchTheme
+    - FiraCode
+    - FontaweSome
+    - OutofBox
+  features:
+    - TypeScript
+    - TsLint & Prettier
+    - Tag list
+    - Custom page layout
+    - Switch the dark mode according to the system theme
+    - Scss
+    - Pagination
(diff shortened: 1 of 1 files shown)


docs/starters.yml
```diff
@@ -3033,3 +3033,20 @@
     - Includes plugins for offline support out of the box
     - TypeScript
     - Prettier & eslint to format & check the code
+- url: https://quietboy.net
+  repo: https://github.com/zhouyuexie/gatsby-starter-quiet
+  description: Gatsby out of the box blog, use TypeScript and highly customized style.
+  tags:
```
wardpeet: we don't support all these tags. What do you think of the following?

```json
tags:
  - Typescript
  - Styling:SCSS
  - Linting
```

for a full list you can checkout https://www.gatsbyjs.org/starters/

docs/starters.yml
```diff
@@ -3033,3 +3033,20 @@
     - Includes plugins for offline support out of the box
     - TypeScript
     - Prettier & eslint to format & check the code
+- url: https://quietboy.net
```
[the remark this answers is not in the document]
zhouyuexie: Very strange, I can open it here. . .
Do you have a response to ping this domain name?

docs/starters.yml
zhouyuexie: I am going to modify it.

zhouyuexie:
Since one of the goals of this template is out of the box, most styles are
covered by adding styled-component and sass in consideration of the scenes
that the user may use.

开 2019年6月18日 在 15:03:29, Ward Peeters (notifications@github.com) 写到:

I also have a question about styled-components. What's the reason for
adding styled-components instead of using SASS? You're using 99% sass and
than for one component switched to styled-components. In my eyes it doesn't
seem a good choice.

PR #14842, opened 2019-06-17
chore(gatsby): officially stop supporting Node 6

### LINKED ISSUE #14231: Update Gatsby to support Node 8 and above
## Summary

Node 6 has reached [its official end-of-life date](https://github.com/nodejs/Release) as of 30th April 2019. 

Today Gatsby _mostly_ supports Node 6. However as some of Gatsby's dependencies have upgraded, they've dropped Node 6 support. This means that e.g. `gatsby new` won't work when running Node 6, but an existing site will continue to build fine (for now).

Increasingly we are having to make decisions about dependencies that have (rightly) stopped supporting Node 6.

We should create a policy on how Gatsby intends to handle supporting Node versions, then implement that policy for Gatsby's Node 6 support.

Related reading: 
- [Gatsby Node 8 RFC](https://github.com/gatsbyjs/rfcs/blob/master/text/0006-node-8-minimum-version.md)
- [Node's releases doc](https://github.com/nodejs/Release)

## Proposals

Gatsby should support any Node version who's status is _Current Release_, _Active LTS_, or _Maintenance LTS_. I think that's not too controversial.

What do we do once a supported version reaches _End of Life_ status?

1. ~~Stop actively supporting that version of Node?~~
  - ~~This means it may work, but the Gatsby team will not run tests or fix issues related to that version of Node. In addition, Gatsby dependencies can be upgraded to versions that may drop support for that version of Node.~~
  - ~~Essentially - you can try and run Gatsby with this version of Node, but you're on your own.~~

2. **Explicitly drop support for that version of Node**
  - Release minor versions of Gatsby and official plugins that drop support for that version of Node. This would involve work like updating `babel-preset-gatsby` to target only newer Node versions, resulting in less code being shipped. 
  - Gatsby definitely won't work if you try and use this version of Node.

3. ~~Release a major version of Gatsby and related plugins that drops support for that version?~~
  - ~~Releasing major versions is a huge amount of churn for a change that most people won't notice. (It's worth noting here that Gatsby doesn't officially follow semver's versioning conventions)~~

4. ~~Something else?~~

### Motivation

Time passes, Node versions increment and fade away, Gatsby must adapt.

Edit: updated to show we've gone with option 2. See comments for to-dos.

DSchau:
## Description

This PR makes the breaking change (potentially! Only users on Node 6) that allows us to move off of support for Node 6. Why are we doing this? See: #14231 for more detail.

One change that merits further explanation:

It's required that babel-preset-gatsby-package takes an option (`nodeVersion` but open to calling it whatever) so that we transpile gatsby-cli specifically to Node 6 support. Otherwise we get e.g. Object rest spread errors like below

```
/Users/dschau/Code/Work/gatsby/gatsby/packages/gatsby-cli/lib/create-cli.js:104
      let args = { ...argv,
                  ^^^

SyntaxError: Unexpected token ...
```

Note: upon merging this PR, we should minor bump all packages (which will be every package that depends upon babel-preset-gatsby-package which is _every_ package).

BREAKING CHANGE: Gatsby no longer supports Node 6. Please upgrade to Node 8 or greater.

## Related Issues

Fixes #14231


packages/gatsby-cli/src/index.js
```diff
@@ -3,22 +3,24 @@
...
-const version = process.version
-const verDigit = Number(version.match(/\d+/)[0])
-
 const pkg = require(`../package.json`)
 const updateNotifier = require(`update-notifier`)
 // Check if update is available
 updateNotifier({ pkg }).notify()
 
-if (verDigit < 6) {
+const MINIMUM_NODE_VERSION = `8.0.0`
```
DSchau: Would be to nice to couple this with the version of Gatsby (core) or something. Not sure of the best approach for that--shouldn't block this PR, but something to think about.

DSchau: Maybe engines in gatsby/package.json?

DSchau: (I went with the engines approach for now!)

DanielRuf:
Do we follow SemVer? If so breaking changes should be just in major versions.

DSchau:
@DanielRuf there's a difference between _known_ breaking changes and potential breaking changes. Based on telemetry data, less than 1% of users are on Node 6, and the users who are on Node 6 are using a very old version of Gatsby and the Gatsby CLI. As such, the the intersection between users using Node 6 and users using the latest version(s) of Gatsby is incredibly small, perhaps an entirely null set.

As such--we feel comfortable "breaking" semantic versioning practice here and releasing it as a minor version, because the likelihood of this breaking real apps is very, very slim, if not entirely zero.

Does that make sense and does that seem fair? See #14231 for more rationale on this.

DSchau:
TODO:

- [ ] Remove engines
- [ ] Use 8.9.0 as our baseline (LTS)

wardpeet:
> Do we follow SemVer? If so breaking changes should be just in major versions.

We follow semver, but lots of packages don't. We wanted to keep node 6 alive in version 2. Sadly it's not feasible to support node 6 anymore, dependencies of dependencies are breaking semver.

Because of this, we don't want to bump a major as we don't think it's a substantive change to bump. We feel bad about it, but don't see another way. Yarn resolutions would have worked. There is no equivalent for this with npm. We have a lot of npm users so that would still break things.

@DSchau I feel like we should write a small blog post about this why we dropped node 6 and didn't bump a major.

DSchau:
@wardpeet it's a good idea to do a blog post, although this is really more of a "blurb" than a full-on post.

sidharthachatterjee:
@DSchau Small note: I know you removed `ci/circleci: unit_tests_node6` but it is set as a required check in this repo's settings. Not sure if that might need to be changed.

DSchau:
> but it is set as a required check in this repo's settings. Not sure if that might need to be changed.

Yeah - but I shouldn't really do that _until_ this is merged :)

DSchau:
Added engines to each package outside of `gatsby-cli`. `>=8.0.0` is the version of Node.js we're committing to supporting.

The benefit of this is that Yarn will fail to install, and npm will warn. This is a good signal to the end user that things _will_ break and they should upgrade, as is the existing content of this PR which will error from gatsby-cli on <8.0.0.

This is ready for another review. Once approved, I'll plan to publish tomorrow morning with minor release(s) of every package.

freiksenet:
SHIP IT

Issue #14827, opened 2019-06-16
childImageSharp is null when using allMarkdownRemark filter query


crutchcorn:
## Description

I have a repo that has a JSON file with author data for a blog I am making inside a folder `/src/data`:

```json
[
  {
    "id": "crutchcorn",
    "name": "Corbin Crutchley",
    "description": "Is a person who exists",
    "profileImg": "./crutchcorn.png"
  }
]
```

Inside that same folder exists the file `crutchcorn.png`

I am mapping this author data to the frontmatter by adding this into the `gatsby-config.js`:

```
mapping: {
  "MarkdownRemark.frontmatter.author": `AuthorsJson`,
}
```

I noticed that when I was using  `filter` query on `allMarkdownRemark`, I am not able to query the author profile picture `childImageSharp`, but I am when using `sort`. EG, when running:

```graphql
{
  allMarkdownRemark(sort: { fields: [frontmatter___date], order: DESC }) {
    edges {
      node {
        frontmatter {
          author {
            name
            profileImg {
              childImageSharp {
                fixed(width: 60, height: 60) {
                  base64
                }
              }
            }
          }
        }
      }
    }
  }
}
```

I get the base64 value I expect. However if I change the query to read:

```graphql
  allMarkdownRemark(filter: {frontmatter:{author: {id: {eq: "crutchcorn"}}}}) {
```

I am only returned `null`s for the `profileImg`. This is true both in the UI and in the GraphiQL

I have made a repo to reproduce the issue in question (though have not included a UI to reproduce, so GraphiQL will be able to verify the problem):

https://github.com/crutchcorn/reproduce-gatsby-childsharp-bug

### Environment

Run `gatsby info --clipboard` in your project directory and paste the output here.

<details>
<summary>Environment Info</summary>

```
 OS: macOS High Sierra 10.13.6
    CPU: (4) x64 Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 8.16.0 - ~/.nvm/versions/node/v8.16.0/bin/node
    Yarn: 1.16.0 - /usr/local/bin/yarn
    npm: 6.9.0 - ~/.nvm/versions/node/v8.16.0/bin/npm
  Languages:
    Python: 2.7.16 - /usr/local/bin/python
  Browsers:
    Chrome: 74.0.3729.169
    Safari: 12.1.1
  npmPackages:
[...]

PR #14843, opened 2019-06-17
fix(gatsby): Prefer nodes over inline objects in findRootNodeAncestor

stefanprobst:
* When calling `findRootNodeAncestor`, we should always prefer nodes looked up by `node.parent` over nodes looked up via node-tracking. This seems to fix the issue in #14827 -- although I'm not 100% sure why we have a node-tracking entry for author in the WeakMap.

* Also fixes the 2 tests in node-model which were broken (check with `describe.only` on the `findRootNodeAncestor` block)

### DIFF
--- a/packages/gatsby/src/db/node-tracking.js
+++ b/packages/gatsby/src/db/node-tracking.js
@@ -1,4 +1,5 @@
 const _ = require(`lodash`)
+const reporter = require(`gatsby-cli/lib/reporter`)
 
 /**
  * Map containing links between inline objects or arrays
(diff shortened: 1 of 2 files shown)


packages/gatsby/src/db/node-tracking.js
```diff
@@ -111,32 +112,25 @@ exports.trackInlineObjectsInRootNode = trackInlineObjectsInRootNode
...
+  while (iterations++ < 100) {
+    if (predicate && predicate(node)) return node
+
+    const parent = node.parent && getNode(node.parent)
+    const id = getRootNodeId(node)
+    const trackedParent = id && getNode(id)
+
+    if (!parent && !trackedParent) return node
+
+    node = parent || trackedParent
```
wardpeet: if I understand correclyt this is fixing the bug? Rest is just cleanup of the crazy while? 😛

stefanprobst: yes exactly

Issue #14796, opened 2019-06-14
[docs][workflows] Finding a Source Plugin


gillkyle:
Part of the **Top 25 Learning Workflows initiative**. See #13708 for the meta issue that this issue falls under.

# User story

As a new Gatsby user, I want to find a source plugin for my CMS (or other data source) of choice.

(going to consider Contentful, WordPress, Sanity, and Postgres)

# Evaluation

[Change emoji below based on your evaluation.]

| Search | Discover | Complete | Linked | Tone | Style | Overall |
| --- | --- | --- | --- | --- | --- | --- |
| 😄 | 😐 | 😄  | 😄 | 😐 | 😄 | 😄  |

# Steps taken to implement

1. Searchability
	1. searched `gatsby with contentful`; clicked second [result](https://www.gatsbyjs.org/packages/gatsby-source-contentful/), third [result](https://www.gatsbyjs.org/docs/sourcing-from-contentful/) is the doc on sourcing 👍 
	1. searched `gatsby with wordpress`; clicked first [result](https://www.gatsbyjs.org/docs/sourcing-from-wordpress/)
	1. searched `gatsby with sanity`; clicked first [result](https://www.gatsbyjs.org/packages/gatsby-source-sanity/), second [result](https://www.gatsbyjs.org/docs/sourcing-from-sanity/) is also the sourcing doc
	1. searched `gatsby with postgres`; clicked first [result](https://www.gatsbyjs.org/packages/gatsby-source-pg/)
2. Discoverability
	1. searched `wordpress` [Sourcing from WordPress](https://www.gatsbyjs.org/docs/sourcing-from-wordpress/) is the first documentation result (same for contentful, sanity, postgres)
	1. searched `database` and got right to Sourcing from Databases page, which could link to even more plugins than it shows [**rec**] update the lists of plugins for sourcing information on pages like Sourcing From Databases to include as many plugins as we currently have (maybe this info could be pulled in dynamically sort of like the plugin library is?..)
	1. clickpath to sourcing content on .org (**4 clicks**)
		1. clicked on *Docs* in Navbar
		1. clicked on *Reference Guides* in Sidebar
		1. clicked on *Sourcing Content and Data* in Sidebar [**rec**] from #13131 include an explanation on this overview page
		1. click on relevant page linked from overview (for Headless CMS this takes another 2 clicks) 
3. Completeness
[...]

PR #14846, opened 2019-06-17
docs(www): 25 Workflows - Finding a Source Plugin

gillkyle:
_Part of the 25 workflows initiative: #13708_

## Description

- lifted up information on source plugins to the overview page on sourcing data
- updated CMS list and converted it into a table instead of a weird list with links taking you all over the place (see screenshot)
- added more database plugins
- added note for plugin contributions

This workflow already seemed pretty well documented so the changes aren't too drastic.

## Related Issues

Made recommendations outlined in #14796

## Screenshot

<details>
<summary>Screenshots</summary>

Before:
image
After:
localhost_8000_docs_headless-cms_

</details>


www/src/data/sidebars/doc-links.yaml
```diff
@@ -109,16 +109,16 @@
...
-                  link: /docs/sourcing-from-wordpress/
-                - title: Sourcing from Drupal
-                  link: /docs/sourcing-from-drupal/
                 - title: Sourcing from Contentful
                   link: /docs/sourcing-from-contentful/
-                - title: Sourcing from Prismic
-                  link: /docs/sourcing-from-prismic/
                 - title: Sourcing from Netlify CMS
                   link: /docs/sourcing-from-netlify-cms/
+                - title: Sourcing from WordPress
```
marcysutton: What does the change in order represent? It doesn't look alphabetical...

gillkyle: Like the [comments](https://github.com/gatsbyjs/gatsby/blob/5bb6adec3b63d533aa570c48f485b4cd614c4971/www/src/data/sidebars/doc-links.yaml#L109-L111) above it say there should only be the top 5 CMSs and I reordered them based off of their rank in the plugin library (ie how many downloads they get, which maybe isn't the _best_ metric)

Drupal was an exception, it's the 5th one even though it's used a little less it seems to be a focus that it's supported.

gillkyle: Maybe it would be worth adding more than 5 though?

docs/docs/content-and-data.md
```diff
@@ -3,4 +3,12 @@ title: Sourcing Content and Data
 overview: true
 ---
 
+A core feature of Gatsby is its ability to **load data from anywhere**. This is what makes Gatsby more powerful than many other static site generators that are limited to only loading content from Markdown files.
```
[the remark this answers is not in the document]
gillkyle: What if it were reworded as: 

> This is what makes Gatsby more powerful than static site generators that...

Then it's not really implied that Gatsby is a static site generator, but can set it apart from then.

If not that, maybe:

> This is what makes Gatsby powerful as a modern site framework...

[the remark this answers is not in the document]
marcysutton: I think the first one would work fine!

www/src/data/sidebars/doc-links.yaml
marcysutton: This makes sense, I was just curious! I think we should keep it at 5. Drupal is a strategic choice, so that's why it's there.

PR #14863, opened 2019-06-18
fix(gatsby-cli): move `react` to peerDependencies

### LINKED ISSUE #14862: gatsby-cli has react in package.json dependencies
## Description

`gatsby-cli`'s `package.json` has `react` in `dependencies`. It makes all kind of unwanted behavior in monorepo setups. For example:

If you have `react` installed in the parent directory (`./`) and gatsby site installed by `gatsby new` command on a child directory (`./site`) directory, it creates a nasty bug. If you use hooks in your application, this error pops up:

```
Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
```

And it is specifically caused by the point number 1 described in the error message. If you remove the parent directory's `node_modules` (i.e. `./node_modules`) everything works fine.

### Steps to reproduce

1. `mkdir parent && cd parent && npm init`
2. `npm install --save-dev react react-dom`
3. `npx gatsby new site`
5. This will install `react` in `./site/node_modules` (which it should not).

### Expected result

`react` should not be installed automatically..

### Actual result

`react` is installed because `gatsby-cli` has `dependencies` with `react`.

MunifTanjim:
`react` in `dependencies` caused problem if parent directory's
node_modules had `react` installed.

fixes #14862

### DIFF
--- a/packages/gatsby-cli/package.json
+++ b/packages/gatsby-cli/package.json
@@ -34,7 +34,6 @@
     "opentracing": "^0.14.3",
     "pretty-error": "^2.1.1",
     "prompts": "^2.1.0",
-    "react": "^16.8.4",
     "resolve-cwd": "^2.0.0",
     "semver": "^6.0.0",
     "source-map": "0.5.7",
@@ -54,6 +53,9 @@
   "optionalDependencies": {
     "ink": "^2.0.5",
     "ink-spinner": "^3.0.1"
+  },
+  "peerDependencies": {
+    "react": "^16.8.4"
   },
   "files": [
     "lib"


sidharthachatterjee:
Let's remove it entirely? `gatsby-cli` doesn't actually need `react`

packages/gatsby-cli/package.json
```diff
@@ -55,6 +54,9 @@
     "ink": "^2.0.5",
     "ink-spinner": "^3.0.1"
   },
+  "peerDependencies": {
```
sidharthachatterjee: Let's remove it entirely? `gatsby-cli` doesn't actually need `react`

MunifTanjim: I see `react` imports in this directory: https://github.com/gatsbyjs/gatsby/tree/82848c9614933f8dcb97822e4723121fe340365e/packages/gatsby-cli/src/reporter/reporters/ink

That's why I didn't remove it entirely.

sidharthachatterjee: Oh never mind I missed that 🤦‍♂ Great catch!

sidharthachatterjee: Moving it to peerDependencies will be problematic then because when installing `gatsby-cli` globally, peer deps won't be installed and this will break, right?

MunifTanjim: :sweat: Yeah, you are right.

I don't see any workaround currently. Should I close this PR?

sidharthachatterjee: Yeah, let's close for now

PR #14874, opened 2019-06-18
Add status bar to bottom of screen during develop

KyleAMathews:
This is a simple start but we'll be adding more things to the
status bar over the next few days.

Screenshot 2019-06-18 17 28 24

### DIFF
--- a/packages/gatsby-cli/src/reporter/index.js
+++ b/packages/gatsby-cli/src/reporter/index.js
@@ -179,6 +179,7 @@ const reporter: Reporter = {
       span,
     }
   },
+  // Make private as we'll probably remove this in a future refactor.
   _setStage(stage) {
     if (reporterInstance.setStage) {
       reporterInstance.setStage(stage)
--- a/packages/gatsby-cli/src/reporter/index.js
+++ b/packages/gatsby-cli/src/reporter/index.js
@@ -179,6 +179,11 @@ const reporter: Reporter = {
       span,
     }
   },
+  _setStage(stage) {
+    if (reporterInstance.setStage) {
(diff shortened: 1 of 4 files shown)


packages/gatsby-cli/src/reporter/index.js
```diff
@@ -179,6 +179,11 @@ const reporter: Reporter = {
       span,
     }
   },
+  _setStage(stage) {
```
muescha: why is this method with underscore?

DSchau: It's a convention to denote that this method is private and should not be used outside of Gatsby Core.

muescha: thx

KyleAMathews: yeah, the implementation will change in the near future so I don't want anyone to start using it in the meantime.

packages/gatsby/src/commands/develop.js
```diff
@@ -428,7 +428,14 @@ module.exports = async (program: any) => {
...
   function printInstructions(appName, urls, useYarn) {
-    console.log()
+    report._setStage({
+      stage: `BootstrapFinished`,
+      context: {
+        url: urls.localUrlForBrowser,
+        appName,
+      },
+    })
+    console.log(`\x1B[2J\x1B[0f`)
```
wardpeet: what's this? :p maybe add a comment

KyleAMathews: clears the screen :-) yeah, good idea — added comment.

PR #14876, opened 2019-06-18
fix(gatsby-transformer-react-docgen): always create description nodes

jquense:
## Description

Gatsby started warning about `description` being resolved by two fields. In practice what was happening is that sometimes it'd just be the inferred string and other times the node here. This changes ensures that description is always a `ComponentDescription` node


jquense:
oops sorry, missed the template doing the github editor. I don't quite know how i'd unit test this, the "bug" is more in the interaction between gatsby's schema generation than anything the plugin does does.

jquense:
arg gonna have to actually pull this down sorry

jquense:
k this should be good, added some kinda of test as well

wardpeet:
Fixed in `gatsby-transformer-react-docgen@4.0.3`

PR #14902, opened 2019-06-19
Update prismic-source-graphql-tutorial.md

jordyvanraaij:
Fixed a couple of small typos / errors

## Description

- Added alt attribute from Prismic to Home Blog image code examples
- Fixed order of wrong export declaration in Home Blog query
- Fixed not existing key on blog post rendering by getting is from _meta


docs/tutorial/prismic-source-graphql-tutorial.md
```diff
@@ -124,7 +124,7 @@ export default ({ data }) => {
   return (
     <Layout>
       <div>
-        <img src={doc.node.image.url} alt="avatar image" />
+        <img src={doc.node.image.url} alt={doc.node.image.alt} />
```
marcysutton: Should we add a note about how to add alt text? I found this resource, although it only talks about SEO and not accessibility: https://user-guides.prismic.io/articles/768849-add-metadata-to-an-asset
Another good one would be the alt decision tree: https://www.w3.org/WAI/tutorials/images/decision-tree/

jordyvanraaij: Although I think accessibility is very important, I don't see how adding notes about accessibility is fitting in a readme about implementing the Prismic GraphQL api?

marcysutton: The motivation for adding it was to inform users where the data would come from. Does it provide an empty string if they haven't filled that data in?

marcysutton: bump! any input on alt being empty?

wardpeet: When it's not set it's probably going to be null. We could add a small note under the snippet to https://user-guides.prismic.io/articles/768849-add-metadata-to-an-asset. Adding the https://www.w3.org/WAI/tutorials/images/decision-tree/ might be a bit too much.

jordyvanraaij: @wardpeet I agree with your proposal, adding the w3 spec doesn't make sense imo. I will go ahead and add the metadata info from Prismic to the tutorial.

PR #14920, opened 2019-06-19
change layout component from class to function component

Strangehill:
in effort to keep the blog starter in sync with best practices, and with [gatsby v2 migration docs](https://www.gatsbyjs.org/docs/migrating-from-v1-to-v2/#remove-or-refactor-layout-components), I'm proposing the layout component be changed from a class into a function component. 
For showcasing class components, there is the index page at this time, though since this too would normally be better off as a function component, I'd ask if there couldn't be a better use case for a class component that could be turned into a new feature of the starter.

### DIFF
--- a/starters/blog/src/components/layout.js
+++ b/starters/blog/src/components/layout.js
@@ -3,73 +3,71 @@ import { Link } from "gatsby"
 
 import { rhythm, scale } from "../utils/typography"
 
-class Layout extends React.Component {
-  render() {
-    const { location, title, children } = this.props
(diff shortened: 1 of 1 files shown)


LekoArts:
Hi, thanks for the PR. But as mentioned in the other PR we really don’t see a value in changing these things just for the sake of changing something. Class Components are still perfectly fine to use. Thanks for your effort, we‘d be happy to get PRs that e.g. fix good first issues :)

Strangehill:
okie dokie

PR #14923, opened 2019-06-19
chore(telemetry): Track createNode utilization

jamo:
To better understand how the nodes are created and updated

### DIFF
--- a/packages/gatsby/src/db/loki/nodes.js
+++ b/packages/gatsby/src/db/loki/nodes.js
@@ -1,6 +1,7 @@
 const _ = require(`lodash`)
 const invariant = require(`invariant`)
 const { getDb, colls } = require(`./index`)
+const { trackCli } = require(`gatsby-telemetry`)
 
 /////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


packages/gatsby/src/db/loki/nodes.js
```diff
@@ -204,6 +205,8 @@ function createNode(node, oldNode) {
   invariant(node.internal.type, `node has no "internal.type" field`)
   invariant(node.id, `node has no "id" field`)
 
+  trackCli(`CREATE_LOKI_NODE`, {}, { debounce: true })
```
wardpeet: What's the reason for specifically track loki nodes? All node creation should be captured by https://github.com/gatsbyjs/gatsby/pull/14923/files#diff-c739d0ad63537276e8fc320522a8b985R594

jamo: I'm not 100% sure if that's necessary as I only have partial understanding of this part. The main idea was to ensure we don't miss out anyting if/when we move between  redux and loki.

PR #14924, opened 2019-06-19
feat(gatsby-cli): improve version output

iansu:
## Description

This change improves the output of `gatsby --version`. If run in a Gatsby project the output will look something like this:

```txt
Gatsby CLI version: 2.6.13
Gatsby version: 2.9.9
Note: this is the Gatsby version for the site at: /Users/iansu/Projects/oss/gatsby-test-blog
```

If run outside of a Gatsby project the output will look like this:

```txt
Gatsby CLI version: 2.6.13
```

## Related Issues

There's no issue for this but there is a maintainers team discussion: https://github.com/orgs/gatsbyjs/teams/maintainers/discussions/65?from_comment=7#discussion-65-comment-7

### DIFF
--- a/packages/gatsby-cli/src/create-cli.js
+++ b/packages/gatsby-cli/src/create-cli.js
@@ -314,8 +314,8 @@ function getLocalGatsbyVersion() {
 
 function getVersionInfo() {
   const { version } = require(`../package.json`)
-  const localVersion = isLocalGatsbySite()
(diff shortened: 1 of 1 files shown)


wardpeet:
I've updated the PR to get the real Gatsby version instead of the version defined in package.json as ^ could be a wrong one.

PR #14929, opened 2019-06-19
chore(sites): fix inactive/invalid Gatsby sites in site showcase

DSchau:
## Description

See title ☝️


DSchau:
@josefaidt wanted to give you a heads up that your site is down. We've removed it from the Site Showcase until you can get that fixed up. Thanks!

PR #14932, opened 2019-06-19
feat(gatsby): Support absolute certificate paths for development

cranberyxl:
## Description
All certificate paths are currently relative to the project directory. This change makes it support `/` and `~/`

```gatsby develop --https --key-file ~/.certs/star.dev.key --cert-file ~/.certs/star.dev.crt``` 

Before: `/Users/cranberyxl/gatsby-project/Users/cranberyxl/.certs/star.dev.key`

After: `/Users/cranberyxl/.certs/star.dev.key`

### DIFF
--- a/packages/gatsby/src/utils/get-ssl-cert.js
+++ b/packages/gatsby/src/utils/get-ssl-cert.js
@@ -3,6 +3,14 @@ const report = require(`gatsby-cli/lib/reporter`)
 const fs = require(`fs`)
 const path = require(`path`)
 
+const resolvePath = (directory, filePath) => {
+  // Support absolute paths
+  if (filePath[0] === `/`) {
+    return filePath
+  }
+  return path.join(directory, filePath)
+}
+
 module.exports = async ({ name, certFile, keyFile, directory }) => {
   // check that cert file and key file are both true or both false, if they are both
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/gatsby/src/utils/get-ssl-cert.js [hunk 1]
```diff
@@ -3,6 +3,14 @@ const report = require(`gatsby-cli/lib/reporter`)
 const fs = require(`fs`)
 const path = require(`path`)
 
+const resolvePath = (directory, filePath) => {
+  // Support absolute paths
+  if (filePath[0] === `/`) {
```
DSchau: How about [path.isAbsolute](https://nodejs.org/api/path.html#path_path_isabsolute_path)?

```suggestion
  if (path.isAbsolute(filePath[0])) {
```

Also why just one part of the path? I'm a little confused by that.

cranberyxl: `filePath` is a string, so I was looking at the first character. I'll adjust it to use the path method

wardpeet:
Published in `gatsby@2.11.0`

aaronjensen:
This is great, though afaict, self signed certs don't work anymore: https://github.com/gatsbyjs/gatsby/issues/14990

PR #14934, opened 2019-06-19
Remove mention to onRouteUpdate being removed from v2

sebastienfi:
## Description

[I see in the code](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-canonical-urls/src/gatsby-browser.js#L1) that `onRouteUpdate` and `onPreRouteUpdate` still exists while in the v1-to-v2 migration docs it is mentioned that these have been removed.

## Related Issues

none.


wardpeet:

The docs are correct. We used to add an `action` prop to the api call saying `push` or `replace`, we don't do it anymore.

PR #14939, opened 2019-06-20
feat(docs): updates to the glossary

marcysutton:
## Description

In reviewing https://github.com/gatsbyjs/gatsby/pull/14846 I wanted to add a link to "backend" in the glossary, but I noticed both that term and "frontend" weren't written to be consistent with the [Gatsby Style Guide](https://www.gatsbyjs.org/contributing/gatsby-style-guide/#capitalize-proper-nouns). I added a few terms that I've noticed were missing and updated a few terms, plus added a navigation block to the top (suggested improvements below). Finally, I noticed the Client Data Fetching doc wasn't actually in the sidebar despite a whole PR for that so I added it.

Something I could use help on is writing the navigation block more dynamically so we can give it better semantics. Right now the Markdown looks like this:

```
[A](#a) | [B](#b) | [C](#c) | [D](#d) | [E](#e) | [F](#f) | [G](#g) | [H](#h) | [I](#i) | [J](#j) | [K](#k) | [L](#l) | [M](#m) | [N](#n) | [O](#o) | [P](#p) | [Q](#q) | [R](#r) | [S](#s) | [T](#t) | [U](#u) | [V](#v) | [W](#w) | [X](#x) | [Y](#y) | [Z](#z)
```
The two things I think need to change for that are:

- wrap the list in `<nav><ul></ul></nav>`
- wrap each item in an `<li>` use a border for the separators instead of pipe characters

Seems like a job for MDX. @gillkyle could you help me with that?


gillkyle:
I can look into this later today, I think starting to use components for tables and navigation like this will be really helpful 👍

gillkyle:
Here's a comparison of the old vs new (the old one was removed in favor of the MDX component that's shown on the bottom):

mdx-rel-headers

www/src/components/horizontal-nav-list.js
```diff
@@ -0,0 +1,39 @@
...
+  &:hover {
+    background-color: ${colors.ui.hover};
+  }
+`
+
+const HorizontalNavList = ({ items = [], slug }) => (
+  <nav>
+    <LinkList>
+      {items.map(item => (
+        <Link key={item} to={`${slug.slice(0, -1)}#${item.toLowerCase()}`}>
```
marcysutton: The Link needs to go inside of the Li to properly structure a UL. So it would look like this:
```
<LinkList>
    {items.map(item => (
        <Li key={item}>
            <Link to={`${slug.slice(0, -1)}#${item.toLowerCase()}`}>{item}</Link>
        </Li>
    ))}
</LinkList>
```

docs/docs/glossary.md
```diff
@@ -1,8 +1,13 @@
...
----
+import HorizontalNavList from "../../www/src/components/horizontal-nav-list.js"
 
-When you're new to Gatsby there can be a lot of words that seem alien. This glossary aims to give you a 10,000ft overview of common terms and what they mean to the layperson.
+# Glossary
+
+When you're new to Gatsby there can be a lot of words to learn. This glossary aims to give you a 10,000-foot overview of common terms and what they mean for Gatsby sites.
+
+<HorizontalNavList
+items={"ABCDEFGHIJKLMNOPQRSTUVWXYZ".split("")}
```
marcysutton: So awesome! and you could pass an array for more flexibility later?

www/src/components/horizontal-nav-list.js
gillkyle: TIL! I guess I just always assumed a `<li>` could be any number of levels of descendancy deep in a list. Thanks for pointing this out rather than just correcting it yourself 🙂 

Fixing now...

marcysutton: You can put lists inside of list items, but they have to be properly structured with parents > children. You can read more here, note the "permitted content" section: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul

docs/docs/glossary.md
gillkyle: Yeah that was the idea! And it's a lot shorter than a big array or object of links and titles.

marcysutton:
I added two new terms: CSS and Progressive Enhancement. Also added a link to #frontend about How Browsers Work. It's such a great resource!

docs/docs/glossary.md
```diff
@@ -70,11 +83,11 @@ Content and data's origin point, typically integrated into Gatsby with [source p
 
 ### Database
 
-A database is a structured collection of data or content. Often a [CMS](#cms) will save to a database using [back-end technologies](#back-end). They're often accessed in Gatsby via a [source plugin](#source-plugin)
+A database is a structured collection of data or content. Often a [CMS](#cms) will save to a database using [back-end technologies](#backend). They're often accessed in Gatsby via a [source plugin](#source-plugin)
```
gillkyle:
```suggestion
A database is a structured collection of data or content. Often a [CMS](#cms) will save to a database using [backend technologies](#backend). They're often accessed in Gatsby via a [source plugin](#source-plugin)
```

Should the hyphen be omitted here too? It looks like it was changed in a couple other places.

docs/docs/glossary.md
```diff
@@ -70,11 +83,11 @@ Content and data's origin point, typically integrated into Gatsby with [source p
...
 
 ### Decoupled
 
-Decoupling describes the separation of different concerns. With [Gatsby](#gatsby) this most commonly means decoupling the [Front-End](#front-end) from the [Back-End](#back-end), like with [Decoupled Drupal](https://dri.es/how-to-decouple-drupal-in-2019) or [Headless Wordpress](https://www.smashingmagazine.com/2018/10/headless-wordpress-decoupled/).
+Decoupling describes the separation of different concerns. With [Gatsby](#gatsby) this most commonly means decoupling the [Front-End](#frontend) from the [Back-End](#backend), like with [Decoupled Drupal](https://dri.es/how-to-decouple-drupal-in-2019) or [Headless Wordpress](https://www.smashingmagazine.com/2018/10/headless-wordpress-decoupled/).
```
gillkyle:
```suggestion
Decoupling describes the separation of different concerns. With [Gatsby](#gatsby) this most commonly means decoupling the [frontend](#frontend) from the [backend](#backend), like with [Decoupled Drupal](https://dri.es/how-to-decouple-drupal-in-2019) or [Headless Wordpress](https://www.smashingmagazine.com/2018/10/headless-wordpress-decoupled/).
```

Same kind of thing? (maybe I'm splitting hairs with this)

docs/docs/glossary.md
```diff
@@ -84,6 +97,10 @@ The process of [building](#build) your website or app and uploading onto a [host
 
 The [environment](#environment) when you're developing your code. It's accessed through the [CLI](#cli) using `gatsby develop`, and provides extra error reporting and things to help you debug before building for [production](#production-environment).
 
+### DOM
+
+The Document Object Model, commonly referred to as "the DOM", is a standard browser API that connects web pages to scripts or programming languages by representing the structure of an HTML document in memory. Developers commonly interact with the DOM through [HTML](#html) markup (written in [JSX](#jsx) in Gatsby), as well as both [React](https://reactjs.org/docs/react-dom.html) and [vanilla JavaScript](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Introduction#DOM_and_JavaScript) code. Another important aspect of utilizing the DOM to its full potential is writing [accessible](#accessibility) HTML markup to expose a page's structure to assistive technology.
```
gillkyle: Reading this definition alone makes me want to link to this page for anyone trying to find info to start learning about web dev 🥇 these are awesome!

docs/docs/glossary.md
```diff
@@ -128,15 +145,15 @@ A markup language that every web browser is able to understand. It stands for Hy
 
 ### Headless CMS
 
-A [CMS](#cms) that only handles the [back-end](#back-end) content management instead of handling both the back-end and [front-end](#front-end). This type of setup is also referred to as [Decoupled](#decoupled).
+A [CMS](#cms) that only handles the [backend](#backend) content management instead of handling both the back-end and [frontend](#frontend). This type of setup is also referred to as [Decoupled](#decoupled).
```
gillkyle:
```suggestion
A [CMS](#cms) that only handles the [backend](#backend) content management instead of handling both the backend and [frontend](#frontend). This type of setup is also referred to as [Decoupled](#decoupled).
```

docs/docs/glossary.md
marcysutton: Yes, you got it! I thought I found them all with find and replace, but obviously not 😅

docs/docs/glossary.md
marcysutton: Yep, right on!

marcysutton:
Going to merge this since it's blocking another PR!

PR #14955, opened 2019-06-20
fix(gatsby): update tsc definition of cache in NodePluginArgs

d4rekanguok:
## Description

This PR fixes the type of `cache` in `NodePluginArgs`. It should be the instance of cache, instead of Cache.


d4rekanguok:
Urg I was hoping to be able to get away with editing the definition file directly via github, since it's adding less than 10 chars. Will submit one with proper format -- sorry!

d4rekanguok:
pinging @sidharthachatterjee

wardpeet:
Published in `gatsby@2.11.0`

Issue #14814, opened 2019-06-16
Add babel-plugin-transform-react-remove-prop-types


neo:
## Summary

I am not 100% sure, but it seems like `babel-plugin-transform-react-remove-prop-types` isn't being used when gatsby is building for production.

In my opinion, this is a very useful and practical thing to add.

### Basic example

N/A

### Motivation

To achieve smaller production build bundle.

sidharthachatterjee:
Sounds useful but I'm not _entirely_ sure if we should add it in our default config (which is super lightweight by design). 

Let's keep this open and hear what everyone else has to say. 

In the mean time, you could add this in a [custom .babelrc file](https://www.gatsbyjs.org/docs/babel/)

sidharthachatterjee:
Just spoke with @DSchau about this and we both agree that this is useful indeed! 

We would love to accept a PR with this. 💪

luisFilipePT:
Hi,
Can I try to tackle this one?

Cheers

sidharthachatterjee:
@luisFilipePT Sure! Thank you for picking it up 🙌

wardpeet:
@luisFilipePT You should be able to add it to [babel-preset-gatsby/src/index.js](https://github.com/gatsbyjs/gatsby/blob/master/packages/babel-preset-gatsby/src/index.js)

luisFilipePT:
@wardpeet thanks for the tip 👍 

I started to check it out last night, just not sure yet you to test it, this is what I have tried (so far)

1 - Added the npm package in `packages/babel-preset-react`
2 - Added the config to `.babelrc` in the same folder (so maybe this will go in `index.js` 🤔 )
3 - Built everything
4 - Build the default project in `starters`
5 - Go to `public` and search for `propTypes` in that folder
6 - Compare with the original build (before the change)

Am I on the right path? 😄

PS - The goal here, for now, is just to remove propTypes on Gatsby prod but not on the plugins, right?

Thanks guys

luisFilipePT:
kind of struggling here, this are my current changes:

I'm missing something 😕 

![image](https://user-images.githubusercontent.com/9373787/59847400-47a75880-935a-11e9-854e-d7f87e4a06c6.png)

0tho:
@luisFilipePT how is this going? I might be able to help you out

luisFilipePT:
> @luisFilipePT how is this going? I might be able to help you out

[...]

PR #14987, opened 2019-06-20
feat(babel-preset-gatsby): add babel-plugin-transform-react-remove-prop-types for production builds

luisFilipePT:
### PR Checklist

- [x] Tests for the changes have been added.

### Purpose of this PR
What kind of change does this Pull Request introduce?

- [x] feat:     A new feature

### Changes

Remove PropTypes from production build by adding babel-plugin-transform-react-remove-prop-types.

This change removes PropTypes and their imports from the production build.
This change DOES NOT remove proptypes from dependencies from external packages (in `node_modules`)

### Is there anything you would like reviewers to focus on?

No
### Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

### Related issue

Closes [#14814](https://github.com/gatsbyjs/gatsby/issues/14814).

### Reviewers

Gatsby team?
### Additional Reviewers

Please review: @sidharthachatterjee @neo @wardpeet


packages/babel-preset-gatsby/src/index.js
```diff
@@ -40,6 +43,18 @@ module.exports = function preset(_, options = {}) {
     }
   }
 
+  const plugins = []
+
+  if (PRODUCTION) {
```
pieh: I think we could target just `` stage === `build-javascript` `` here

luisFilipePT: Thanks, I will test, probably will keep both conditions.

wardpeet: `stage === `build-javascript` is the only condition you need as it's only run in production mode ^^

luisFilipePT: @wardpeet & @pieh refactored in [this](https://github.com/gatsbyjs/gatsby/pull/14987/commits/83c5645932bf6403f8fc4a82d792f319672c1a16) commit.
But tomorrow will try to do one more change in the tests - remove the `beforeEach` don't think it makes sense since its only one test. Maybe then the `Gatsby Build Service` gets unstuck 😄

pieh: Don't worry about `Gatsby Build Service` (that's not required :) )

packages/babel-preset-gatsby/src/index.js
```diff
@@ -45,6 +45,18 @@ module.exports = function preset(_, options = {}) {
     }
   }
 
+  const plugins = []
```
sidharthachatterjee: We shouldn't need this now that you've added it to the array in the bottom right?

pieh: Oh, that's my bad - I didn't remove that part before merging - I'll fix this up

neo:
should we expose some config like [`additionalLibraries`](https://www.npmjs.com/package/babel-plugin-transform-react-remove-prop-types#additionallibraries) or those people with needs of configurations should just create their own `.babelrc`?

sidharthachatterjee:
> should we expose some config like additionalLibraries or those people with needs of configurations should just create their own .babelrc?

@neo We could definitely add `additionalLibraries` as an option in a follow up PR

pieh:
@neo You can modify babel with .babelrc - see https://www.gatsbyjs.org/docs/babel/#how-to-use-a-custom-babelrc-file

neo:
yea, that's what I figured – I was just curious about where do we draw the line.
I think it makes sense to not exposing any options, otherwise we would need to choose what to expose...
so all good with me! 😆 glad it's merged.

wardpeet:
Proptypes is used in a lot of dependencies as it's the only runtime check supported by react. We want to be as fast and as small as possible. Having good defaults is key so removing prop-types by default in production makes a lot of sense.

PR #14996, opened 2019-06-20
fix(gatsby-transformer-sqip): make it work again

axe312ger:
### Issue 1

When sharp & contentful are enabled, gatsby crashes with:

```
Error: Schema must contain uniquely named types but contains multiple types named "Sqip".
```

This fixes the issue by using two different names for both SQIP objects

## Issue 2

Contentful does not cache sqips in richt folder (`node_modules/.cache/...`)

## Issue 3

In one of the merges in the last PR, one of us added code from the old queueing system.

This lead to broken and/or duplicate SQIP generation 😓 

It for sure was because the code was pretty weird, with this and the last PR it should be more maintainable now :)


freiksenet:
Thank you!

PR #15008, opened 2019-06-21
fix(showcase): Remove developerecosystem.com

sidharthachatterjee:
It's no longer a Gatsby site

### DIFF
--- a/docs/sites.yml
+++ b/docs/sites.yml
@@ -906,13 +906,6 @@
     - Blog
     - Web Development
     - Music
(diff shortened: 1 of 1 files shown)


durkinio:
I just noticed it disappeared after I added it again last month :(  

It's definitely a Gatsby site, and it will be forever and ever :) https://www.developerecosystem.com/

I'm even wearing Gatsby socks as I type this...

muescha:
but it still have the gatsby div tag?
`html > body > div#___gatsby`

PR #15009, opened 2019-06-21
fix(starters): restore vertical spacing in blog post layout

Strangehill:
fixes a sneaky side effect of removing negative margins in the layout for the starter-blog in #10758

### DIFF
--- a/starters/blog/src/templates/blog-post.js
+++ b/starters/blog/src/templates/blog-post.js
@@ -18,13 +18,19 @@ class BlogPostTemplate extends React.Component {
           title={post.frontmatter.title}
           description={post.frontmatter.description || post.excerpt}
         />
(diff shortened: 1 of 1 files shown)


wardpeet:
What issue are you seeing? (are you talking about the spacing between site titel & blog title?

before your fix:
image


after your fix:
image

Strangehill:
that's correct. That spacing was too large. I merely restored its original size. The enlargement was a side-effect that got through unnoticed on #10758 when a negative margin was removed from the layout. The reasoning for removing that negative margin was it messed up the headers on additional pages (such as a src/pages/blog.js). The layout was being used not just in the index page, but also in the blog-post template. When the negative margin in the layout was removed it inadvertently expanded that space between headers.

Issue #14993, opened 2019-06-20
Client-Side Routing Broken in Latest Release


alexlafroscia:
## Description

Client-side routing does not seem to be supported any longer when serving the app in "production" (either through `gatsby serve` or some other static file serving). An unrecognized route redirects back to the route where the client-side router has been mounted.

### Steps to reproduce

You can start with this repo here:

https://github.com/alexlafroscia/___gatsby-dynamic-routing-bug

Basically you

- Generate a new Gatsby app
- Configure Gatsby to treat all routes (`/*`) as being client-side
- View the production version of the app

### Expected result

I would expect that, if you hit something like the `/foobarbax`, it would render the client-side 404 page. This is actually working in local dev.

**Index Page**

<img width="936" alt="Screen Shot 2019-06-20 at 3 18 42 PM" src="https://user-images.githubusercontent.com/1645881/59885344-d5417300-936f-11e9-8d7f-2d23b9b8e463.png">

**Client-Side 404 Page**

<img width="936" alt="Screen Shot 2019-06-20 at 3 18 46 PM" src="https://user-images.githubusercontent.com/1645881/59885376-e5f1e900-936f-11e9-8dba-9ebeae3574f8.png">

### Actual result

In the fully static mode it redirects back to the root of the app instead of rendering my 404 component.

This was recorded when running `gatsby serve` in that repo

![Kapture 2019-06-20 at 15 23 23](https://user-images.githubusercontent.com/1645881/59885404-fbffa980-936f-11e9-94eb-ceb3b96d9bfa.gif)

Even after changing the URL, it redirects back to the root of the app instead.

### Environment

```txt
  System:
    OS: macOS 10.14.5
    CPU: (12) x64 Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
    Shell: 5.6.2 - /usr/local/bin/zsh
  Binaries:
    Node: 10.13.0 - /var/folders/nv/f8tyc7b960nf4kptb5890t400000gn/T/yarn--1561069776540-0.10964542529805676/node
    Yarn: 1.13.0 - /var/folders/nv/f8tyc7b960nf4kptb5890t400000gn/T/yarn--1561069776540-0.10964542529805676/yarn
    npm: 6.4.1 - ~/.asdf/installs/nodejs/10.13.0/bin/npm
  Languages:
    Python: 2.7.10 - /usr/bin/python
  Browsers:
    Chrome: 75.0.3770.100
    Firefox: 64.0.2
[...]

PR #15010, opened 2019-06-21
fix(gatsby): Fix client side routing match paths regression

Moocar:
Fix for https://github.com/gatsbyjs/gatsby/issues/14993. The issue was that I accidentally removed a check whether the URL matches the page's matchPath before explicitly navigating. I fixed and hopefully made the logic a bit easier to follow.


packages/gatsby/cache-dir/loader.js
```diff
@@ -66,7 +66,7 @@ const cleanAndFindPath = rawPathname => {
     return cleanAndFindPathCache[trimmedPathname]
   }
 
-  let foundPath = findMatchPath(matchPaths, trimmedPathname)
+  let foundPath = findMatchPath(trimmedPathname)
```
sidharthachatterjee: Leaving this for anyone reviewing after me:

This function is being changed to not accept `matchPaths` because, `loader.addMatchPaths(matchPaths)` is typically called before it which inits the matchPaths inside loader

packages/gatsby/cache-dir/loader.js
```diff
@@ -26,7 +26,7 @@ if (process.env.NODE_ENV !== `production`) {
 // Cache for `cleanAndFindPath()`. In case `match-paths.json` is large
 const cleanAndFindPathCache = {}
 
-const findMatchPath = (matchPaths, trimmedPathname) => {
+const findMatchPath = trimmedPathname => {
```
sidharthachatterjee: This doesn't seem to take into account `__BASE_PATH__` (which was checked earlier in https://github.com/gatsbyjs/gatsby/blob/daa2af3255131d94de065127b215068e82a4a412/packages/gatsby/cache-dir/production-app.js#L72)

Not sure if that's an issue

Moocar: Great catch, It is an issue. Fixed by stripping the BASE_PATH (like we do in loader.cleanAndFindPath).

sidharthachatterjee: Lovely, thanks!

wardpeet:
Tested the PR with examples given on the issues and using examples/client-only-paths. It seems to work. Also tested it with prefix-paths 👌 . Some tests would be great.

sidharthachatterjee:
Added some end to end tests to this!

e2e-tests/serve/gatsby-config.js
```diff
@@ -0,0 +1,65 @@
...
+        name: `gatsby-starter-default`,
+        short_name: `starter`,
+        start_url: `/`,
+        background_color: `#663399`,
+        theme_color: `#663399`,
+        display: `minimal-ui`,
+        icon: `src/images/gatsby-icon.png`, // This path is relative to the root of the site.
+      },
+    },
+    `gatsby-plugin-sitemap`,
```
wardpeet: Same here

e2e-tests/serve/gatsby-config.js
```diff
@@ -0,0 +1,65 @@
...
+  assetPrefix: `http://localhost:9001`,
+  pathPrefix: `/blog`,
+  siteMetadata: {
+    siteUrl: `http://localhost:9000`,
+    title: `Gatsby Default Starter`,
+  },
+  plugins: [
+    `gatsby-plugin-react-helmet`,
+    {
+      resolve: `gatsby-plugin-manifest`,
```
wardpeet: Any reason for this plugin?

e2e-tests/serve/gatsby-config.js
```diff
@@ -0,0 +1,65 @@
...
+
+module.exports = {
+  assetPrefix: `http://localhost:9001`,
+  pathPrefix: `/blog`,
+  siteMetadata: {
+    siteUrl: `http://localhost:9000`,
+    title: `Gatsby Default Starter`,
+  },
+  plugins: [
+    `gatsby-plugin-react-helmet`,
```
wardpeet: It doesn't do much harm but should we disable this one?

e2e-tests/serve/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "gatsby-starter-default",
+  "description": "Gatsby default starter",
+  "version": "1.0.0",
+  "author": "Kyle Mathews <mathews.kyle@gmail.com>",
+  "dependencies": {
+    "cypress": "^3.3.1",
+    "gatsby": "^2.8.0",
+    "gatsby-cypress": "^0.2.0",
+    "gatsby-plugin-manifest": "^2.2.0",
```
wardpeet: I would suggest dropping everything we don't need/aren't testing.

packages/gatsby/cache-dir/loader.js
wardpeet: Not typically, always 😄

rchrdnsh:
ETA until this fix is merged? Looks like it only needs one more test to pass...

...currently dead in the water on a site until this is fixed, sooooo...

...what's the web dev equivalent of having to pee really badly but stuck in a long line for the bathroom? 

🥺

wardpeet:
@rchrdnsh you can always lock gatsby on version < 2.9.0

rchrdnsh:
no version seems to work for me now...I've locked to 2.8.3, 2.8.6, 2.9.0, and I think a couple others, but none work for me, for some reason...🤷‍♂️

I might be doing it wrong, however. I go into package.json and change the gatsby version number, then delete and re-install my npm modules, and as well as delete the cache. Is this ok? Is there a better way to do it?

wiziple:
@rchrdnsh better do it using command because your gatsby version is already locked by `package-lock.json` or `yarn.lock`.

The other workaround is just removing existing lock file and then installing again.

rchrdnsh:
hmmmm...never messed with the lock file before...can I just delete that file then bpm install?

wardpeet:
@rchrdnsh yes you can. You could install `gatsby@resource-loading` to get a version with this fix

sidharthachatterjee:
This is good to go now. I've removed the new e2e test suite and instead have added tests for these in the existing suites, namely `development-runtime` and `production-runtime`

sidharthachatterjee:
Published in `gatsby@2.10.4`

PR #15013, opened 2019-06-21
fix(gatsby-plugin-preact): Make preact/compat server render HTML

knowler:
## Description

This just aliases `react-dom/server` to `preact/compat` so that SSR works as currently it doesn’t.


wardpeet:
Thanks for putting this PR up. Does this solve an issue or just making it more consistent?

knowler:
It makes it more consistent with how Gatsby and React work. Without it the root element isn't prerendered when you build. I couldn't find an open issue for it.

wardpeet:
I don't see any changes in output but I love consistency so 👍

Issue #15011, opened 2019-06-21
gatsby-plugin-create-client-paths: Sets matchPath incorrectly for client side path entry 


sidharthachatterjee:
When setting prefixes in the config for `gatsby-plugin-create-client-paths`, it seems that a client side path other than `/*` still gets its matchPath set to `/*` incorrectly. 

This means that if a project has two different client side path entry points, say `/*` and `/app/*`, even for the second case, pages will reach the entry point for `/*`

Reproduction is at https://github.com/sidharthachatterjee/gatsby-15011-reproduction

### Expected 

http://localhost:8000/amhjajk should render Not Found in Index

### Actual

http://localhost:8000/amhjajk renders Not Found in App

PR #15014, opened 2019-06-21
fix(gatsby-plugin-create-client-paths): set matchPath correctly

sidharthachatterjee:
Array.some short circuits early whereas we _need_ to run through all to get the _best_ match 

Fixes https://github.com/gatsbyjs/gatsby/issues/15011

### DIFF
--- a/packages/gatsby-plugin-create-client-paths/src/gatsby-node.js
+++ b/packages/gatsby-plugin-create-client-paths/src/gatsby-node.js
@@ -18,7 +18,7 @@ exports.onCreatePage = ({ page, store, actions }, { prefixes }) => {
       return resolve()
     }
 
-    prefixes.some(prefix => {
+    prefixes.forEach(prefix => {
       if (!re[prefix]) {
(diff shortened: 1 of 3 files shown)


packages/gatsby-plugin-create-client-paths/src/gatsby-node.js
```diff
@@ -18,7 +18,7 @@ exports.onCreatePage = ({ page, store, actions }, { prefixes }) => {
       return resolve()
     }
 
-    prefixes.some(prefix => {
+    prefixes.forEach(prefix => {
```
wardpeet: github commmenting is dumb 💤 

can you also remove the return true & return false from this function as forEach doesn't care.

I'm also thinking if we should drop the promise. What use has it?

sidharthachatterjee: Promise shouldn't be necessary since `createPage` is an action creator and therefore not async

Issue #9374, opened 2018-10-24
Tutorial: add `gatsby-image`


shannonbux:
## Summary

We want to create a new tutorial for `gatsby-image` for intermediate users. This learning resource should show how to create responsive images step-by-step. It would fit best as the first item on the [Additional Tutorials](https://www.gatsbyjs.org/tutorial/advanced-tutorials/) page. 

Here's some information on creating tutorials: https://www.gatsbyjs.org/contributing/docs-templates/#tutorials

### Motivation

Many Gatsby users don't know about `gatsby-image` and it's one of the most noticeable benefits to Gatsby site visitors. 

@jlengstorf continuing the discussion from https://github.com/gatsbyjs/gatsby/issues/9158#issuecomment-432853609

siddhant1:
On it !

jlengstorf:
All yours, @siddhant1!

@shannonbux Where do you think this should go? My thought is that it might be its own new section after Part 4 that could be called "Using Images in Gatsby".

Things to cover:

- Gatsby _can_ use regular images, just like any other web framework (show a small example)
- Gatsby has tools to automatically optimize images that have been packaged up in `gatsby-image`
- Show how to install `gatsby-image` and it's peer dependencies
- Configure the peer deps to load images from a folder
- Write a query to load the image data for `gatsby-image`
- Discuss the different options (e.g. traced SVG, WebP)
- Link to using-gatsby-image.gatsbyjs.org

Anything I'm missing here?

treb0r:
It would be great to see some documentation on how to leverage an external service such as [imgix](https://www.imgix.com/) with gatsby-image, possibly using [react-imgix](https://github.com/imgix/react-imgix).

justinformentin:
siddhant1 are you still working on this? Just wondering, because if not, I have it ready to go.

justinformentin:
@siddhant1 are you still working on this? Just wondering, because if not, I have it ready to go.

lvrbrtsn:
Can I take this?

siddhant1:
@Ivrbrtsn sure !

shannonbux:
@lvrbrtsn thanks! See @jlengstorf's comment above. https://github.com/gatsbyjs/gatsby/issues/9374#issuecomment-433088391

[...]

PR #15027, opened 2019-06-21
Create gatsby-image-tutorial

lstar19:
## Description

This is the first draft of a potential tutorial for using gatsby-image in various ways.

## Related Issues
Addresses #9374


docs/tutorial/gatsby-image-tutorial
```diff
@@ -0,0 +1,275 @@
...
+
+Now you can use it. Note that the key for pointing to the image corresponds to the way in which the image was processed. In this example that is `fixed`.
+
+```jsx
+<Img
+  className="headshot"
+  fixed={data.file.childImageSharp.fixed}
+  alt="headshot"
+/>
+```
```
marcysutton: I feel like a full component or page example showing the query and the component usage here would help tell a bit more of the story as well as set up the `useStaticQuery` example later on. Perhaps it could say something like "here's the query and usage all put together"

docs/tutorial/gatsby-image-tutorial
```diff
@@ -0,0 +1,275 @@
...
+        }
+      }
+    }
+  `)
+
+  return <Img fixed={data.file.childImageSharp.fixed} />
+}
+```
+
+Instead of a query constant and data that references the result, you can use a static query hook directly in the JXS code and then reference it. Note that the query language didn’t change and neither did the `Img` tag syntax, the only change was the location of the query and the usage of the `useStaticQuery` function to wrap it.
```
marcysutton:
```suggestion
Instead of a query constant and data that references the result like in the first section above, you can put the `useStaticQuery` hook directly in the JSX code and then reference it in the `Img` component. Note that the query language didn’t change and neither did the `Img` tag syntax; the only change was the location of the query and the usage of the `useStaticQuery` function to wrap it.
```

docs/tutorial/gatsby-image-tutorial
```diff
@@ -0,0 +1,275 @@
...
+
+### Multiple Queries and Aliasing
```
marcysutton:
```suggestion
## Multiple queries and aliasing
```

docs/tutorial/gatsby-image-tutorial
```diff
@@ -0,0 +1,275 @@
...
+
+The last use case you may come across is how to handle a situation where you have multiple queries in the same file/page.
+
+This example is attempting to query for all the data in `speaking.yaml` and the direct file query in our first example. In order to do this you want to use aliasing.
```
marcysutton:
```suggestion
This example is attempting to query for all the data in `speaking.yaml` and the direct file query in our first example. In order to do this you want to use aliasing in GraphQL.
```
Adding a qualifier here in case people want to go find an API for aliasing and they don't know where to look!

docs/tutorial/gatsby-image-tutorial
```diff
@@ -0,0 +1,275 @@
...
+
+The first thing to know is that an alias is assigning a name to a query. Below is a simple example.
+
+```jsx
```
marcysutton:
```suggestion
```graphql
```

docs/tutorial/gatsby-image-tutorial
```diff
@@ -0,0 +1,275 @@
...
+
+Instead of a query constant and data that references the result, you can use a static query hook directly in the JXS code and then reference it. Note that the query language didn’t change and neither did the `Img` tag syntax, the only change was the location of the query and the usage of the `useStaticQuery` function to wrap it.
+
+### Multiple Queries and Aliasing
+
+The last use case you may come across is how to handle a situation where you have multiple queries in the same file/page.
+
+This example is attempting to query for all the data in `speaking.yaml` and the direct file query in our first example. In order to do this you want to use aliasing.
+
+The first thing to know is that an alias is assigning a name to a query. Below is a simple example.
```
marcysutton:
```suggestion
The first thing to know is that an alias is assigning a name to a query. The second thing to know is that aliases are optional, but they can make your life easier! Below is an example.
```

docs/tutorial/gatsby-image-tutorial
```diff
@@ -0,0 +1,275 @@
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+```
+
+When you do that, you’ve changed the reference to that object in your JSX. While it was previously referenced as
```
marcysutton:
```suggestion
When you do that, you’ve changed the reference to the query object available in your JSX code. While it was previously referenced as this:
```

docs/tutorial/gatsby-image-tutorial
```diff
@@ -0,0 +1,275 @@
...
+
+```jsx
+{data.allSpeakingYaml.edges.map(({ node }) => ())
+```
+
+giving it an alias does not add a level of complexity to the response object, it just replaces it. So you end up with the same structure referenced as
```
marcysutton:
```suggestion
Giving it an alias does not add a level of complexity to the response object, it just replaces it. So you end up with the same structure, referenced like this (note the alias `talks` in place of the longer `allSpeakingYaml`):
```

docs/tutorial/gatsby-image-tutorial
```diff
@@ -0,0 +1,275 @@
...
+
+```jsx
+{data.talks.edges.map(({ node }) => ())
+```
+
+The top-level object name of data is implicit. This is important because when you conduct multiple queries as part of a single component `Gatsby` still passes the entire result to the component.
```
marcysutton:
```suggestion
The top-level object name of `data` is implicit. This is important because when you conduct multiple queries as part of a single component, Gatsby still passes the entire result to the component. 

Here's an example of data flowing into a component:
```

docs/tutorial/gatsby-image-tutorial
```diff
@@ -0,0 +1,275 @@
...
+
+```jsx
+const SpeakingPage = ({ data}) => {})
+```
+
+everything else gets referenced from that top-level return
```
marcysutton:
```suggestion
Everything else gets referenced from that top-level return
```

docs/tutorial/gatsby-image-tutorial
```diff
@@ -0,0 +1,275 @@
...
+name.
+
+With that understanding, you can combine two queries and use aliasing to distinguish between them.
```
marcysutton:
```suggestion
With that understanding, you can combine two queries referencing images and use aliasing to distinguish between them.
```

docs/tutorial/gatsby-image-tutorial
```diff
@@ -0,0 +1,275 @@
...
+
+```jsx
```
marcysutton:
```suggestion
```graphql
```

docs/tutorial/gatsby-image-tutorial
```diff
@@ -0,0 +1,275 @@
...
+{data.allSpeakingYaml.edges.map(({ node }) => ())
+```
+
+And then like this to access the image using the alias name `banner`.
+
+```jsx
+<Img fluid={data.banner.childImageSharp.fluid} />
+```
+
+### Aspect Ratio
```
marcysutton:
```suggestion
## Aspect ratio
```

docs/tutorial/gatsby-image-tutorial
```diff
@@ -0,0 +1,275 @@
...
+
+These examples should handle a fair number of your use cases. A couple bonus things. First, `gatsby-image` has a nice feature that gives you the ability to set an aspect ratio. This can be used for fixed or fluid processed images, it doesn't matter.
```
marcysutton:
```suggestion
`gatsby-image` has a feature that gives you the ability to set an aspect ratio to constrain image proportions. This can be used for fixed or fluid processed images; it doesn't matter.
```

docs/tutorial/gatsby-image-tutorial
```diff
@@ -0,0 +1,275 @@
...
+```jsx
+{data.allSpeakingYaml.edges.map(({ node }) => ())
+```
+
+And then like this to access the image using the alias name `banner`.
+
+```jsx
+<Img fluid={data.banner.childImageSharp.fluid} />
+```
+
```
marcysutton:
```suggestion

These examples should handle a fair number of use cases. A couple bonus things:
```
Moving this above the aspect ratio heading, as it might flow a little better to set up the next two examples.

docs/tutorial/gatsby-image-tutorial
```diff
@@ -0,0 +1,275 @@
...
+### Aspect Ratio
+
+These examples should handle a fair number of your use cases. A couple bonus things. First, `gatsby-image` has a nice feature that gives you the ability to set an aspect ratio. This can be used for fixed or fluid processed images, it doesn't matter.
+
+```jsx
+<Img sizes={{ ...data.banner.childImageSharp.fluid, aspectRatio: 21 / 9 }} />
+```
+
```
marcysutton: I think this would benefit from a little explanation of what's happening.

```suggestion
This example uses the `sizes` option on the `Img` component to specify the `aspectRatio` option along with the fluid image data. This processing is made possible by `gatsby-plugin-sharp`.
```

docs/tutorial/gatsby-image-tutorial
```diff
@@ -0,0 +1,275 @@
...
+### Bonus Error
```
marcysutton:
```suggestion
## Bonus error
```

docs/tutorial/gatsby-image-tutorial
```diff
@@ -0,0 +1,275 @@
...
+
+Now for errors to watch out for. If you change your image processing from `fixed` to `fluid` you may see this error.
+
+![In image cache error message.](https://tenmilesquare.com/wp-content/uploads/2019/02/Screen-Shot-2019-02-18-at-11.46.03-AM-1024x212.png)
```
marcysutton: We should import the image here in case that URL changes! Can you save the file locally and add it to this PR? You'd want to create a directory called `gatsby-image-tutorial` and then you can save the index.md and image in the same place.

docs/tutorial/gatsby-image-tutorial
```diff
@@ -0,0 +1,275 @@
...
+
+Despite its appearance, solving this doesn't actually require flushing any kind of cache. In reality, it has to do with incompatible references. You likely triggered it because you changed the query to process the image as `fluid` but the JSX key was still set to `fixed`, or visa versa.
+
+### The End
```
marcysutton:
```suggestion
## The end
```

docs/tutorial/gatsby-image-tutorial
```diff
@@ -0,0 +1,275 @@
+---
+title: "Using Gatsby-Image With Your Site Tutorial"
+---
```
marcysutton: To follow the template in [docs-templates](https://www.gatsbyjs.org/contributing/docs-templates/#tutorials) we should do a little more setup:

```
## What’s contained in this tutorial?

By the end of this tutorial, you’ll have done the following:

- learned how to use `gatsby-image` for responsive images
- queried for a single image with GraphQL
- sourced multiple images through YAML files
- learned how to troubleshoot common errors
```

docs/tutorial/gatsby-image-tutorial
```diff
@@ -0,0 +1,275 @@
...
+
+Now for errors to watch out for. If you change your image processing from `fixed` to `fluid` you may see this error.
+
+![In image cache error message.](https://tenmilesquare.com/wp-content/uploads/2019/02/Screen-Shot-2019-02-18-at-11.46.03-AM-1024x212.png)
+
+Despite its appearance, solving this doesn't actually require flushing any kind of cache. In reality, it has to do with incompatible references. You likely triggered it because you changed the query to process the image as `fluid` but the JSX key was still set to `fixed`, or visa versa.
+
+### The End
+
+So that's it. This post included a number of different possible use cases, so don't feel as if you need to explore them all. Pick the examples and tips that apply to your implementation.
```
marcysutton: Adding some resources to complete the loop!
```suggestion

## Other resources
- [Gatsby Image API docs](/docs/gatsby-image/)
- [Using Gatsby Image](/docs/using-gatsby-image/)
- [Other image and media techniques in Gatsby](/docs/images-and-files/)
```

lstar19:
Thanks Marcy! I'll wait for all the feedback to come in and then incorporate it together. Thanks for taking the time.

marcysutton:
Here's some feedback from @shannonbux!

> I think this has a good tone and fairly good pacing (level of detail seems like just enough to get the tasks done).
> The one thing that doesn't quite match the tutorial template is that this isn't really a series of steps. I think in order to be that kind of tutorial, it would need to be a series of steps that you complete. Right now it's more like a conversational guide. 
> One thing that could be cool is some sort of starter or preset collection of images so the tutorial can show screenshots of what the finished result looks like, and tips on how to test the performance of the images (I always like to slow my browser down to watch them load slower ;)

I think the most actionable thing here is to update the tutorial to match the format outlined in the [tutorial template](https://www.gatsbyjs.org/contributing/docs-templates/#tutorial-template), along with my comments. I love Shannon's suggestion of having a starter project for people to play with–that could be an improvement done later on, as a separate PR.

Thanks again for working on this tutorial, we really appreciate it and you!

lstar19:
Added all your recommended changes Marcy and tried to do some of the "step" stuff up top. What I'm realizing is there are so many paths to take depending on use case that breaking it down into steps is a bit more challenging past initial set up. Should we break this until multiple tutorials? Or are we ok with this branching?

docs/tutorial/gatsby-image-tutorial/gatsby-image-tutorial.md
```diff
@@ -0,0 +1,361 @@
...
+
+Now you're set up to use `gatsby-image`.
+
+### Step 3
+
+Determine where your image files are located. In this example they're in `src/data`.
+
+If you haven't already, make sure that your project is set up to see content inside that directory. That means doing two things:
+
+1.  Install `gatsby-source-filesystem`. Note: If you created your project using `gatsby new <name>`, this first step is already done for you.
```
marcysutton:
```suggestion
1.  Install `gatsby-source-filesystem`. Note: If you created your project using `gatsby new <name>`, this first step should already be done for you via the default starter.
```

docs/tutorial/gatsby-image-tutorial/gatsby-image-tutorial.md
```diff
@@ -0,0 +1,361 @@
...
+
+### Relative image paths and `gatsby-config.js`
+
+You might expect the relative path to be relative to the file the code sits in, in this case that's index.js. However, that doesn't work. The relative path is actually based on the line of code you put in the `gatsby-source-filesystem` config, which points to `src/data`.
+
+### Image fragments
+
+Another thing to note about this query is how it uses the fragment `GatsbyImageSharpFixed` to return a fixed width and height image. You could also use the fragment `GatsbyImageSharpFluid` which produces scalable images that fill their container instead of fitting specific dimensions. In `gatsby-image`, _fluid_ images are meant for images that don’t have a finite size depending on the screen, where as other images are _fixed_.
+
+The query will return a processed image at the highest depth. That is, the deepest part of the return object. The returned result will be automatically passed into the component and attached to the `data` prop. You can then display the image using JSX to automatically output responsive, highly performant HTML.
```
marcysutton:
```suggestion
The query will return a data object including the processed image in a format usable by the `gatsby-image` component. The returned result will be automatically passed into the component and attached to the `data` prop. You can then display the image using JSX to automatically output responsive, highly performant HTML.
```

I think this part will confuse folks–so I'd recommend rewording it slightly. An alternative suggestion would be to show more of the data object returned so the depth part is less abstract.

docs/tutorial/gatsby-image-tutorial/gatsby-image-tutorial.md
```diff
@@ -0,0 +1,361 @@
...
+import Img from "gatsby-image"
+import Layout from "../components/layout"
+
+const HomePage = ({ data }) => {
+  return (
+    <Layout>
+      <Img
+        className="headshot"
+        fixed={data.file.childImageSharp.fixed}
+        alt="headshot"
```
marcysutton:
```suggestion
        alt=""
```
"headshot" wouldn't be super useful for users of assistive technology, so I'd recommend leaving it blank or pulling in the author/site owner's name.

docs/tutorial/gatsby-image-tutorial/gatsby-image-tutorial.md
```diff
@@ -0,0 +1,361 @@
...
+      }
+    }
+  }
+`
+export default HomePage
+```
+
+## Querying for multiple images from YAML data
+
+Another way to source images is through YAML (or Markdown). This example uses the `gatsby-transformer-yaml` plugin to query the YAML files. More information about that plugin can be found in the [Gatsby plugin library](https://www.gatsbyjs.org/packages/gatsby-transformer-yaml/?=gatsby-transformer-yaml).
```
marcysutton:
```suggestion
Another way to source images is through YAML (or Markdown). This example uses the `gatsby-transformer-yaml` plugin to query the YAML files. More information about that plugin can be found in the [Gatsby plugin library](/packages/gatsby-transformer-yaml/?=gatsby-transformer-yaml).
```

docs/tutorial/gatsby-image-tutorial/gatsby-image-tutorial.md
```diff
@@ -0,0 +1,361 @@
...
+
+Here's an example of a query from a list of conferences in a YAML file:
```
marcysutton:
```suggestion
Here's an example of a query from a list of conferences in a YAML file with an image for each one:
```
In this section, I think it would help to get to the image part sooner so the tutorial doesn't get side-tracked with YAML query information. It would help mitigate the "forking" question you had earlier, @lstar19.

docs/tutorial/gatsby-image-tutorial/gatsby-image-tutorial.md
```diff
@@ -0,0 +1,361 @@
...
+
+```graphql
+{
+  allSpeakingYaml {
+    edges {
+      node {
+        conference
+        year
+      }
```
marcysutton:
```suggestion
        image {
          childImageSharp {
            fluid {
              ...GatsbyImageSharpFluid
            }
          }
        }
      }
```
Because the tutorial has already talked about queries, I think it would help to surface the image part of a YAML query sooner

docs/tutorial/gatsby-image-tutorial/gatsby-image-tutorial.md
```diff
@@ -0,0 +1,361 @@
...
+    }
+  }
+}
+```
+
+In this case the query starts with `allSpeakingYaml` to direct `graphql` to look for this data in the `speaking.yaml` file in your `src/data` folder referenced in `gatsby-config.js`. If you want to query a file named `blog.yaml`, for example, you'd start the query with `allBlogYaml`.
+
+### Troubleshooting queries
+
+A common error you might run into is `Unknown field allSpeakingYaml on type Query`, where `allSpeakingYaml` is replaced with something specific to your code. What the error boils down to is that `graphql` is unable to find a matching result. This can mean it couldn't find a file that matched the first line, meaning that no `speaking.yaml` file exists.
```
marcysutton: If we can cover the query gotchas in a few bullet points and keep the info limited to images, it will make this more succinct. See comments above about moving the image query stuff up to cover it sooner!

docs/tutorial/gatsby-image-tutorial/gatsby-image-tutorial.md
```diff
@@ -0,0 +1,361 @@
...
+      }
+    }
+  }
+}
+```
+
+Notice that this example uses aliasing for one query and not the other. This is allowed; there is no requirement that all your queries use aliasing. In this case, the JSX would look like this to access the `speaking.yaml` content.
+
+```jsx
+{data.allSpeakingYaml.edges.map(({ node }) => ())
```
marcysutton: It would help to surface the image part of this component to keep it relevant to `gatsby-image`.

lstar19:
Made some edits, looking at it with a narrow view to gatsby-image I'm wondering if I should remove the entire section titled `Multiple queries and aliasing`. It's helpful in understanding that you can query for multiple images from different places, but it is more about the querying than making use of gatsby-image itself.

marcysutton:
@lstar19 I think it's useful to include the aliases info. `gatsby-image` is one of those things where the line of concerns is fuzzy because it requires Gatsby's data layer to operate. So I think it's fine to leave in!

lstar19:
In that case, I think it might be ready?

lstar19:
Thanks Marcy!

PR #15034, opened 2019-06-21
feat: add gitpod config

svenefftinge:
Hey Gatsby folks 👋 

this PR adds support for onboarding contributors through Gitpod, a free online dev environment. The PR will allow anyone to get up and running with a browser-based dev environment in a few seconds, lowering the barrier for any code contributions.

This is what a dev environment looks like:
Screenshot 2019-06-21 at 19 35 27

The environment starts with a pre-built state of the given branch. On startup three processes are started in three different terminals watching changes in core and the example I picked (gatsbygram). All is based on a configuration file (`.gitpod.yml`) so hopefully pretty transparent. In case it is not, we're always happy to help.

Give it a try: https://gitpod.io/#https://github.com/svenefftinge/gatsby

Shaving off [the local setup](https://www.gatsbyjs.org/contributing/setting-up-your-local-dev-environment/) and a five minutes of build, I hope this will streamline your onboarding process a lot.

I also added a proposal for documentation. So if you think this is useful, please advice how and where you'd prefer to see information about this and of course feedback on the configuration would be great, too.

P.S.: Awesome framework! We really enjoyed building [our website](https://github.com/gitpod-io/website) with Gatsby.

### DIFF
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -1,0 +1,3 @@
+FROM gitpod/workspace-full
+
+RUN npm -g install gatsby-dev-cli
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,35 @@
+image:
+  file: .gitpod.Dockerfile
+tasks:
+  - init: >
+      yarn run bootstrap &&
+      cd ./examples/gatsbygram &&
+      yarn install &&
+      gatsby-dev --set-path-to-repo ../.. &&
+      gatsby-dev --scan-once &&
+      npx gatsby build &&
+      cd ../.. &&
+      touch /tmp/done.txt
+    command: >
+      yarn run watch --scope={gatsby,gatsby-image,gatsby-link}
+  - openMode: split-right
+    before: >
+      cd ./examples/gatsbygram
+    command: >
+      gp await-port 8000 &&
+      gatsby-dev --set-path-to-repo ../.. &&
+      gatsby-dev
+  - before: >
+      cd ./examples/gatsbygram
+    init: |
+      until [ -f /tmp/done.txt ]
+      do
+          sleep 2
+      done
+    command: >
+      npx gatsby develop
+ports:
+  - port: 8000
+    onOpen: open-preview
+  - port: 31997 # yarn mutex
+    onOpen: ignore


svenefftinge:
The failed `unit_test_windows` build says:
```
$ babel-node scripts/check-versions.js
gatsby-page-utils: 
  Depends on "babel-preset-gatsby-package@^0.1.4" 
  instead of "babel-preset-gatsby-package@0.2.0". 
```

I don't think that has anything to do with my change. Is the build a bit flaky?

pieh:
This is awesome! On my test I think it took ~2 minutes to get env ready which is pretty sweet. 

The config file seems straight forward (at least to modify test site to something else than gatsbygram). I do have a question - is there a way to provide multiple presets (using different example sites)? Flow I could see working - if you want to work on `gatsby-transformer-remark` - you probably would want to use `using-remark` example, but if you want to work on `gatsby-source-wordpress`, you would want to use `using-wordpress` etc.

I really need to take Code Review feature for a spin - this sounds awesome

> I don't think that has anything to do with my change. Is the build a bit flaky?

This is pretty weird. Definitely not related to the PR content. Don't worry about this. Only check that we need to make sure is passing is linting, and this is green, so all good!

rayriffy:
Is the file name `docs/contributing/using-an-onlie-dev-environment.md` mistyped?

svenefftinge:
> Flow I could see working - if you want to work on gatsby-transformer-remark - you probably would want to use using-remark example, but if you want to work on gatsby-source-wordpress, you would want to use using-wordpress etc.

That's a great idea. We can pass an environment variable in to overwrite the default example (gatsbygram). We could put buttons into the individual example Readmes.

svenefftinge:
> On my test I think it took ~2 minutes

Still pretty long, it usually was more like 40 seconds for me. But it might be because Gitpod runs in three regions (US, Europe, Asia) and I created the prebuilt in Europe, which is a big tar of the full workspaces including all node dependencies. If you start it from US it has to be transferred there first.

Prebuilds that are automatically triggered from the GitHub webhook are located in US (because GitHub is triggering from US). So then it might be the other way around. That said there's a change coming up that will improve the startup time further.

pieh:
> Still pretty long, it usually was more like 40 seconds for me.

Oh, I wasn't really measuring - it might have been shorter than 2 minutes and closer to 40s - my point was it was pretty quick - much faster than setting up local env usually :)

To move this forward with this addition, I think we could split this PR into 2:
 - one that adds config + Dockerfile (and I think there is absolutely no harm in merging that part as soon as possible) - this will enable us to test all the things (including Code Review flow)
 - second one that adds docs (this will need some review to go over language etc)

svenefftinge:
@pieh I have updated this PR to only include the config. Will follow up with the documentation once it's merged.

PR #15035, opened 2019-06-21
chore: change workflows around ~ use less resources

wardpeet:
## Description
Move some requires around to hopefully use less resources on failing builds.

Idea is to not run e2e tests if gatsby fails.

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,7 +66,8 @@ aliases:
           - /docs.+/
           - /blog.+/
     requires:
-      - bootstrap
+      # only run whne node 10 passes, don't waste resources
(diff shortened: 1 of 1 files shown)


pieh:
How about this (if possible) - as middle point between current and proposal:
run `lint` and one of unit tests (probably node 8) in parallel and run rest of them if both lint and unit tests pass?

Sometimes there are PRs that linting fail, but it's useful to know if unit tests fail too, so we don't end up fixing linting, wait for CIs to see there are unit tests that need to be handled as well (but this would still shortcicruit most expensive CI checks - e2e)

wardpeet:
Updated config

PR #15053, opened 2019-06-23
chore(showcase): Updated gvdrenovations URL, added new site

dkrasniy:
## Description

Updated GVD Renovations URL and added Cali's Finest Landscaping site to the site showcase


wardpeet:
Thanks for sharing your websites. It seems like both `https://www.gvdrenovations.com/` & `https://www.gvdrenovationsinc.com/` are working and resolving to the same site. Why do you want to do this change?

dkrasniy:
The old domain has been mistakenly marked as a malicious site and is blacklisted on some networks by McAfee, meaning it loads up a blank page for many people without redirecting. Hence, we moved the site to a completely new domain.

wardpeet:
Thanks for sharing!

PR #15054, opened 2019-06-23
fix(www): Center footer links

nelsonreitz:
This tweak fixes the footer links horizontal centering by removing the spaces after the last list item. 

Tested on mobile & desktop Chrome, Firefox & Safari.

### DIFF
--- a/www/src/components/shared/footer-links.js
+++ b/www/src/components/shared/footer-links.js
@@ -21,12 +21,12 @@ const FooterList = styled.ul`
 
     &:not(:last-of-type) {
       margin-right: 1em;
-    }
 
-    &:not(:last-of-type):after {
-      color: ${colors.grey[30]};
-      content: "•";
-      padding-left: 1em;
+      &:after {
+        color: ${colors.grey[30]};
+        content: "•";
+        padding-left: 1em;
+      }
     }
 
     a {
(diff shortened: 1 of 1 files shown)


www/src/components/shared/footer-links.js
```diff
@@ -18,18 +18,17 @@ const FooterList = styled.ul`
...
   li {
     display: inline-block;
-    margin-right: 1em;
 
-    &:after {
+    &:not(:last-of-type) {
+      margin-right: 1em;
+    }
+
+    &:not(:last-of-type):after {
```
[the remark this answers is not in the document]
nelsonreitz: Good point thanks!

www/src/components/shared/footer-links.js
```diff
@@ -18,16 +18,15 @@ const FooterList = styled.ul`
...
   li {
     display: inline-block;
-    margin-right: 1em;
 
-    &:after {
-      color: ${colors.grey[30]};
-      content: "•";
-      padding-left: 1em;
-    }
+    &:not(:last-of-type) {
```
wardpeet: `:not` is a selector you should try to avoid. It's a pretty heavy selector as the css parser has to get all elements and start excluding elements that match the not pattern.

nelsonreitz: Thanks a lot Ward, I didn't know about this behavior.

wardpeet: It's how the css parser works in the browser. It's not super bad as the css parser is super fast but if you don't need it, don't use it 😄.

Thanks for updating so fast!

PR #15063, opened 2019-06-23
chore: fix domain issue for GreenOrbit

Nina0708:
## Description
Fix domain issue.

### DIFF
--- a/docs/sites.yml
+++ b/docs/sites.yml
@@ -5893,8 +5893,8 @@
   built_by_url: https://herper.io
   featured: false
 - title: GreenOrbit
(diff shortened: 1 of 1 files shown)


wardpeet:
Hey, it seems like a PR just like this already got rejected a few days ago. What's the reason for changing `main_url: https://www.greenorbit.com/` => `main_url: https://greenorbit.com/` both urls work fine.

Nina0708:
Hi @wardpeet , it's a minor difference, but our URLs canonicalise to https://greenorbit.com, so we thought it might be more correct to use that URL in the Showcase. :)

Issue #3425, opened 2018-01-06
gatsby develop ok, but gatsby build failed


maidh91:
## Description

gatsby develop ok, but gatsby build failed

### Environment

Gatsby version: 1.1.28
Node.js version: 8.9.3
Operating System: Ubuntu 16.04

### File contents (if changed):

This is the source:
https://github.com/aquabubu/home-gatsby

### What happened:

'gatsby develop' ok, but 'gatsby build' failed.

```
success delete html files from previous builds — 0.174 s
success open and validate gatsby-config.js — 0.004 s
success copy gatsby files — 0.031 s
success onPreBootstrap — 0.005 s
success source and transform nodes — 0.587 s
success building schema — 0.296 s
success createLayouts — 0.009 s
success createPages — 0.524 s
success createPagesStatefully — 0.026 s
success onPreExtractQueries — 0.001 s
success update schema — 0.186 s
success extract queries from components — 0.109 s
success run graphql queries — 0.186 s
success write out page data — 0.029 s
success write out redirect data — 0.000 s
success onPostBootstrap — 0.000 s

info bootstrap finished - 3.592 s

Warning: Accessing PropTypes via the main React package is deprecated, and will be removed in  React v16.0. Use the latest available v15.* prop-types package from npm instead. For info on usage, compatibility, migration and more, see https://fb.me/prop-types-docs
Warning: Accessing createClass via the main React package is deprecated, and will be removed in React v16.0. Use a plain JavaScript class instead. If you're not yet ready to migrate, create-react-class v15.* is available on npm as a temporary, drop-in replacement. For more info see https://fb.me/react-create-class
 DONE  Compiled successfully in 11757ms                                                                                                                                                                                                                               23:05:07


You can now view aquabubu in the browser.

  http://localhost:8000/

View GraphiQL, an in-browser IDE, to explore your site's data and schema

  http://localhost:8000/___graphql

Note that the development build is not optimized.
To create a production build, use gatsby build

 WAIT  Compiling...                                                                                                                                                                                                                                                   23:05:07

 DONE  Compiled successfully in 445ms
```

However, gatsby build failed.
```
success delete html files from previous builds — 0.119 s
success open and validate gatsby-config.js — 0.005 s
success copy gatsby files — 0.020 s
success onPreBootstrap — 0.005 s
success source and transform nodes — 0.553 s
success building schema — 0.283 s
success createLayouts — 0.011 s
success createPages — 0.516 s
[...]

PR #15064, opened 2019-06-24
Change suggested directory for page templates

miguelpduarte:
I was having the errors mentioned [here](https://github.com/gatsbyjs/gatsby/issues/3425) and [here](https://stackoverflow.com/questions/54402112/gatsby-develop-command-runs-but-gatsby-build-gives-error), which were fixed by moving the page file from the `pages` directory to the `templates` directory. `pages` appears to work when using `gatsby develop` but not when using `gatsby build`.

## Description

I was having the errors mentioned [here](https://github.com/gatsbyjs/gatsby/issues/3425) and [here](https://stackoverflow.com/questions/54402112/gatsby-develop-command-runs-but-gatsby-build-gives-error) while using [`gatsby-plugin-create-client-paths`](https://www.gatsbyjs.org/packages/gatsby-plugin-create-client-paths/), which were fixed by moving the page file from the `pages` directory to the `templates` directory. `pages` appears to work when using `gatsby develop` but not when using `gatsby build`

## Related Issues

Might be related to #3425 (now closed)

### DIFF
--- a/packages/gatsby-plugin-create-client-paths/README.md
+++ b/packages/gatsby-plugin-create-client-paths/README.md
@@ -20,4 +20,4 @@ Then configure via `gatsby-config.js`:
 ```
 
 In this example, all paths prefixed by `/app/` will render the route described
-in `src/pages/app.js`.
+in `src/templates/app.js`.


wardpeet:
Could you post a reproduction of the issue you're having? src/pages should work fine.

wardpeet:
Friendly ping :)

miguelpduarte:
Sorry, have been out of the "office" and thus unable to post any updates. Will get back at the begining of the next week and will post some logs I have from a netlify deploy that better describe this error.

On 2 Aug 2019 09:37, Ward Peeters <notifications@github.com> wrote:

Friendly ping :)

wardpeet:
enjoy your vacation! 🌴

miguelpduarte:
Thanks!

Now for the long-deserved update to this (which should probably be an issue and not a PR now that I notise it, but past me was a bit of a GitHub Rambo it seems :sweat_smile:):

[Here is a gist](https://gist.github.com/miguelpduarte/7a4ea6d6dc2fc6e87021f12102973fb1) of the deploy log of a commit that had weird errors relative to this package (`gatsby-plugin-create-client-paths`). The working tree was at [this commit](https://github.com/miguelpduarte/move-ment/tree/96a49dba1f39b3690f7ca3d67ad831463c1f4337) (So that the source code can be fully explored).

(The slight delay was also some finishing touches made today to be able to make the project open source in order to be able to show the code :sweat_smile: Sorry for the delay!).

So, to better explain the situation: I was using the aformentioned plugin and ran into some problems when deploying using netlify (several ones in a row, but all looked similar and just like the one linked above). I narrowed it down to this package because there was some complaining about the route it added (`/stop/<id>`) and it resulted in basically calling some functions when building, despite them never being called in that way, which resulted in errors.

I believe that the hotfix used at the time was to add some guard clauses to prevent these errors and allow the build to progress normally, but have since moved from using this package due to not understanding some configurations, so might not be able to reply to some questions immediately (without testing some things again). I switched to using the other solution detailed [here](https://www.gatsbyjs.org/tutorial/authentication-tutorial/#creating-client-only-routes) (but also ran into some problems so might switch back for ease of use).

sidharthachatterjee:
@miguelpduarte Thank you for opening this and the detailed comment! 

`src/pages/app.js` mentioned in the readme is indeed correct and refers to how 
- /app will be created as a result of the `src/pages/app.js` file 
- client side route will be set for `/app`

I understand the issue you're seeing and would urge you to open a separate issue (with a reproduction, expected behaviour and visible behaviour) for it as this is most definitely unrelated to the change in this PR. 

Closing this PR now but please know that we appreciate you taking the time to investigate and open this and would love to have you as a contributor in the future! Thank you again

Issue #15028, opened 2019-06-21
[gatsby-source-drupal] menu link content isn't displaying parent as expect


xaviemirmon:
## Description

I am seeing a value for the  menu link content in the GraphiQL browser as null when I would expect it to be an ID

### Steps to reproduce

Browse your menu in GraphiQL when you have a menu on Drupal with nested menu items

<img width="1440" alt="Screenshot 2019-06-21 at 17 10 28" src="https://user-images.githubusercontent.com/5609840/59936502-11480700-9448-11e9-8eec-69951d4e9ad9.png">

### Expected result

According to the JSON:API there should be a string with the ID for nested menu links on the parent value.

<img width="581" alt="Screenshot 2019-06-21 at 17 12 08" src="https://user-images.githubusercontent.com/5609840/59936720-9d5a2e80-9448-11e9-8920-edee6d86d580.png">

Should be the string as follows in this instance:
menu_link_content:8d61eedd-a561-4001-8df3-a439bd2acd65

### Actual result

See image above it's returned null

### Environment

  System:
    OS: macOS 10.14.5
    CPU: (8) x64 Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 8.11.1 - /usr/local/bin/node
    Yarn: 1.7.0 - /usr/local/bin/yarn
    npm: 6.9.0 - ~/.npm-packages/bin/npm
  Languages:
    Python: 2.7.15 - /usr/local/bin/python
  Browsers:
    Chrome: 75.0.3770.100
    Firefox: 66.0.5
    Safari: 12.1.1
  npmPackages:
    gatsby: ^2.8.5 => 2.8.5 
    gatsby-cli: ^2.6.5 => 2.6.5 
    gatsby-image: ^2.1.2 => 2.1.2 
    gatsby-plugin-manifest: ^2.1.1 => 2.1.1 
    gatsby-plugin-netlify-cache: ^1.2.0 => 1.2.0 
    gatsby-plugin-offline: ^2.1.1 => 2.1.1 
    gatsby-plugin-react-helmet: ^3.0.12 => 3.0.12 
    gatsby-plugin-sharp: ^2.1.3 => 2.1.3 
    gatsby-source-drupal: ^3.1.10 => 3.1.10 
    gatsby-source-filesystem: ^2.0.38 => 2.0.38 
[...]

PR #15075, opened 2019-06-24
fix(gatsby-source-drupal): preserve drupal parent as drupal_parent_menu_item

xaviemirmon:
https://github.com/gatsbyjs/gatsby/issues/15028

## Description

Copies node.parent to node.drupal_parent so that it doesn't get overwritten later on.

## Related Issues

Fixes [#15028](https://github.com/gatsbyjs/gatsby/issues/15028)


pieh:
Let's add `drupal_parent` in https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-source-drupal/src/normalize.js (next to `drupal_id`) instead of change in gatsby-node file

xaviemirmon:
Cool gimme two mins and I'll change that around

pieh:
I also have question:
is `parent` coming from drupal core? or is this some module specific field? Looking at JSON:API result example in issue - it looks like value for it is something like `<type>:<uuid>` - if we could resolve this to node, user could use something like this in queries:

```
menuLinkContent {
  title
  drupal_parent {
    title
  }
}
```
instead of just getting `id`

xaviemirmon:
@pieh it's the id of the parent menu item e.g. if I have as follows:

Link
|-> Link nested

Screenshot 2019-06-24 at 13 56 04

The nested link then references the parent.  Unfortunately, Drupal doesn't provide children (just the parent in the child) otherwise we could do something as you've suggested.

xaviemirmon:
@pieh I've moved this to the normalize.js file.  Let me know if there are any issues.

xaviemirmon:
Yay! the tests are all green @pieh :smile:

pieh:
Before merging this I want to try out doing what I mentioned in https://github.com/gatsbyjs/gatsby/pull/15075#issuecomment-504988932 to not change `drupal_parent` field in future

xaviemirmon:
Cool @pieh! What I was trying to say in my [comment](https://github.com/gatsbyjs/gatsby/pull/15075#issuecomment-504999362) is that I don't think it's necessary as the relationship is the wrong way around.  I can't see a use case for it as the parent is already available further up the hierarchy. What would be ideal but requires quite a bit of work is the opposite relationship. 

```
menuLinkContent {
  title
  drupal_children {
    title
    link
  }
}
```

Do you know any examples elsewhere in other Gatsby plugins that transform the relationship to nest the child under a parent based on the child? 

Does that make sense what I am saying?

Thanks!
Xavier

pieh:
> Do you know any examples elsewhere in other Gatsby plugins that transform the relationship to nest the child under a parent based on the child? I might have a stab at it in a bit when I have more time

We do something like this already in drupal source plugin for "back references" - https://github.com/gatsbyjs/gatsby/blob/9fb78f8634d42e0f5547e320aafed9ae5b489223/packages/gatsby-source-drupal/src/gatsby-node.js#L146-L161


the idea is that we iterate over every node, grab `parent` field, try to find entry that represents the parent, and modify that one to add `drupal_children` there:

```diff
+ const childrenMap = {}
+ const addChild = (parentID, childID) => {
+   if (!childrenMap[parentID]) {
+     childrenMap[parentID] = []
+   }
+   childrenMap[parentID].push(createNodeId(childID))
+ }

  _.each(allData, contentType => {
    if (!contentType) return
    _.each(contentType.data, datum => {
+     if (datum.parent) {
+       const parentID = figureOutId(datum.parent)
+       if (parentID) {
+         addChild(parentID, datum.id)
+       }
+     }
      if (datum.relationships) {
        _.each(datum.relationships, (v, k) => {
          if (!v.data) return

          if (_.isArray(v.data)) {
            v.data.forEach(data => addBackRef(data.id, datum))
          } else {
            addBackRef(v.data.id, datum)
          }
        })
      }
    })
  })
```

And later on in `// process nodes` block we would add:
```js
node.drupal_children___NODE = childrenMap[datum.id]
```

The question is if `figureOutId` is feasible or not given that `parent` field can have weird things like `system.admin_content` you showed in example

packages/gatsby-source-drupal/src/normalize.js
```diff
@@ -6,6 +6,7 @@ const nodeFromData = (datum, createNodeId) => {
     id: createNodeId(datum.id),
     drupal_id: datum.id,
     parent: null,
+    drupal_parent: attributes.parent,
```
sidharthachatterjee: Based on the discussion, 

- `drupal_parent` can't be a Node because `attributes.parent` doesn't always have an `id`
- `drupal_children` is out of scope of this PR

I would recommend that we name this `drupal_parent_id` for now and call it a day for this particular issue. 

What do you think @xaviemirmon @pieh?

sidharthachatterjee:
```suggestion
    drupal_parent_id: attributes.parent,
```

xaviemirmon: Hey @sidharthachatterjee I think it needs to remain as drupal_parent because depending on  how Drupal has been setup it might not actually be an ID

sidharthachatterjee: Can you list possible values types for this key?

xaviemirmon: It's user-configurable so really depend on how the user configures it.

sidharthachatterjee:
In that case, I would be wary of adding this in here since a user configuration could result in differently inferred GraphQL types for this and break the user's site 

Also I want avoid calling it `drupal_parent` because as @pieh mentioned that should be the actual parent Node but since we can't always resolve to it, I don't want to use up that namespace only to have to come back to this when we _do_ figure out how to set the parent here

xaviemirmon:
@sidharthachatterjee 

This functionality is crucial for menus in Drupal and technically all data provided is changeable so I don't think it's worth parking.  What if I were to label it `drupal_parent_menu_item`?  I think would resolve the naming issue we are currently facing because of Gatsby protected use of `parent`.  

Then I proceed with a more standardised Drupal module as per [my comment on 15028](https://github.com/gatsbyjs/gatsby/issues/15028#issuecomment-509691737) so that you could better query the data.

What do you reckon?

sidharthachatterjee:
@xaviemirmon I think that’s a good approach and `drupal_parent_menu_item` looks good to me. Let’s make that change and get this in and follow up after that.

sidharthachatterjee:
@xaviemirmon Can you please merge master in and fix conflicts?

sidharthachatterjee:
Published in `gatsby-source-drupal@3.2.14`

Issue #15073, opened 2019-06-24
gatsby-cli missing the `better-opn` dependency


sandyklark:
## Description

Error when running `gatsby new`.  I believe the dependency is missing but is required [here](https://github.com/gatsbyjs/gatsby/blob/99fb7b44810d658805211073ddfc18e508b57c87/packages/gatsby-cli/src/init-starter.js#L13).

### Steps to reproduce

run `gatsby new`

### Expected result

a new project should be created

### Actual result

Get this error in the cli: `Error: Cannot find module 'better-opn'`

### Environment

Run `gatsby info --clipboard` in your project directory and paste the output here.

mgan59:
Same, was just getting ready to open an issue.  

I run `nvm`, and had the issue with both `10.16.0` and `11.13.0`

This from a fresh installation of `gatsby-cli` this morning.

----

```shell
$ gatsby new my-default-starter https://github.com/gatsbyjs/gatsby-starter-default

/Users/morgan/.nvm/versions/node/v11.13.0/lib/node_modules/gatsby-cli/node_modules/yoga-layout-prebuilt/yoga-layout/build/Release/nbind.js:53
        throw ex;
        ^

Error: Cannot find module 'better-opn'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:655:15)
    at Function.Module._load (internal/modules/cjs/loader.js:580:25)
    at Module.require (internal/modules/cjs/loader.js:711:19)
    at require (internal/modules/cjs/helpers.js:14:16)
    at Object.<anonymous> (/Users/morgan/.nvm/versions/node/v11.13.0/lib/node_modules/gatsby-cli/lib/init-starter.js:32:13)
    at Module._compile (internal/modules/cjs/loader.js:805:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:816:10)
    at Module.load (internal/modules/cjs/loader.js:672:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:612:12)
    at Function.Module._load (internal/modules/cjs/loader.js:604:3)
```

DSchau:
Ah - whoops! #14097 makes a call to better-opn but doesn't depend upon it.

Would either of you like to PR this change? It'll be adding the dependency to gatsby-cli!

YungTosti:
Just had the exact same problem, please provide a fix :)

sandyklark:
I am cloning now and will make a PR as soon as I can

[...]

PR #15077, opened 2019-06-24
Added better-opn as a dependency to package.json

YungTosti:
## Description
Just added better-opn to the package.json

## Related Issues
Fixed issue #15073


pieh:
The package should be added to `gatsby-cli` ( https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-cli/package.json ) not to monorepo package.json. Also thanks for jumping to help on this!

Issue #14918, opened 2019-06-19
[gatsby preview contentful] rich content text field in contentful not fetching new content in preview


FLoppix:
## Description

I have set up the gatbsy preview with contentful. However if you have a rich text field in one of your content types the preview instance is not fetching the new content. The data change is recognized but I dont receive the full payload to render the new data. Instead of getting all the fields back, I am only receiving the ID of the rich content field that changed. Any chances that there is a fix for this?  

```js
  nodeType: "embedded-entry-block",
  content: [],
  data: {
    target: {
      sys: { id: "c17UPYbEdX8LLIh9fFxkEqQ", type: "Link", linkType: "Entry" }
    }
  }
```

### Steps to reproduce

- Set up a gatsby project with contentful and the gatsby preview. 
- Add a rich content text field to one of content types.
- Modify the content and wait for the changes
- Gatsby Preview will recognize the data changes and refetch
- only the above payload is fetched and the new content is not available

### Expected result

The whole content of the rich content text field should be refetched for rendering new content. Instead I expect something some like this: 

```js
nodeType: "embedded-entry-block",
  content: [],
  data: {
    target: {
      sys: {
        space: {
          // space values
        },
        id: "c17UPYbEdX8LLIh9fFxkEqQ",
        type: "Entry",
        createdAt: "2019-06-11T15:24:15.054Z",
        updatedAt: "2019-06-18T17:05:51.976Z",
        environment: {
          sys: { id: "master", type: "Link", linkType: "Environment" }
        },
        revision: 11,
        contentType: {
          sys: { type: "Link", linkType: "ContentType", id: "headline" }
        }
      },
      fields: {
        // HERE I expect all of my relevant new fields with the content
      }
    }
  }
```

### Actual result

Only the above payload is fetched.

### Environment

  System:
    OS: macOS 10.14.1
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 8.11.3 - /var/folders/kk/lc2p7gkd2h7_hsrl11g2wylc0000gn/T/yarn--1560960233949-0.6171038925942727/node
    Yarn: 1.12.3 - /var/folders/kk/lc2p7gkd2h7_hsrl11g2wylc0000gn/T/yarn--1560960233949-0.6171038925942727/yarn
    npm: 5.6.0 - ~/.nvm/versions/node/v8.11.3/bin/npm
  Languages:
    Python: 2.7.15 - /usr/local/bin/python
  Browsers:
[...]

PR #15084, opened 2019-06-24
fix(gatsby-source-contentful): fix missing data with rich text fields

Khaledgarbaya:
## Description

This PRs is an attempt to fix missing data on content changes at contentful.
the gatsby-source-plugin is using the Contentful SDK which is calling the sync API behind the scene for the first sync the SDK will have all the data available to it so it can perform the link resolution however in the subsequent sync the API will return only the changed entries which is not enough data for the SDK to do the link resolution.  

## Solution
By saving the previous sync data and using it later to do link resolution we make sure that all links are resolved

## Related Issues

* https://github.com/contentful/rich-text/issues/94

## TODO
- [x] ~Add tests~ existing tests should do

### DIFF
--- a/packages/gatsby-source-contentful/src/fetch.js
+++ b/packages/gatsby-source-contentful/src/fetch.js
@@ -44,7 +44,7 @@ module.exports = async ({ syncToken, reporter, pluginConfig }) => {
         )}' were found but were filtered down to none.`
       )
     }
-    reporter.info(`default locale is: ${defaultLocale}`)
+    reporter.info(`Default locale is: ${defaultLocale}`)
   } catch (e) {
     let details
     let errors
--- a/packages/gatsby-source-contentful/src/fetch.js
+++ b/packages/gatsby-source-contentful/src/fetch.js
@@ -104,9 +104,9 @@ ${formatPluginOptionsForCLI(pluginConfig.getOriginalPluginOptions(), errors)}`)
   try {
     contentTypes = await pagedGet(client, `getContentTypes`, pageLimit)
   } catch (e) {
-    reporter.panic(`error fetching content types`, e)
+    reporter.panic(`Error fetching content types`, e)
   }
-  reporter.info(`contentTypes fetched ${contentTypes.items.length}`)
+  reporter.info(`Content types fetched ${contentTypes.items.length}`)
 
   let contentTypeItems = contentTypes.items
 
@@ -124,6 +124,8 @@ ${formatPluginOptionsForCLI(pluginConfig.getOriginalPluginOptions(), errors)}`)
     locales,
     space,
   }
+
+  console.timeEnd(`Fetch Contentful data`)
 
   return result
 }
--- a/packages/gatsby-source-contentful/src/fetch.js
+++ b/packages/gatsby-source-contentful/src/fetch.js
@@ -4,7 +4,7 @@ const chalk = require(`chalk`)
 const normalize = require(`./normalize`)
 const { formatPluginOptionsForCLI } = require(`./plugin-options`)
 
-module.exports = async ({ reporter, pluginConfig }) => {
+module.exports = async ({ syncToken, reporter, pluginConfig }) => {
   // Fetch articles.
   console.time(`Fetch Contentful data`)
 
--- a/packages/gatsby-source-contentful/src/fetch.js
+++ b/packages/gatsby-source-contentful/src/fetch.js
@@ -85,10 +85,14 @@ ${formatPluginOptionsForCLI(pluginConfig.getOriginalPluginOptions(), errors)}`)
   }
 
   let currentSyncData
+  const basicSyncConfig = {
+    limit: pageLimit,
+    resolveLinks: false,
+  }
   try {
     let query = syncToken
-      ? { nextSyncToken: syncToken }
-      : { initial: true, limit: pageLimit }
+      ? { nextSyncToken: syncToken, ...basicSyncConfig }
+      : { initial: true, ...basicSyncConfig }
     currentSyncData = await client.sync(query)
   } catch (e) {
     reporter.panic(`Fetching contentful data failed`, e)
--- a/packages/gatsby-source-contentful/src/fetch.js
+++ b/packages/gatsby-source-contentful/src/fetch.js
@@ -4,7 +4,7 @@ const chalk = require(`chalk`)
 const normalize = require(`./normalize`)
 const { formatPluginOptionsForCLI } = require(`./plugin-options`)
 
-module.exports = async ({ syncToken, reporter, pluginConfig }) => {
+module.exports = async ({ reporter, pluginConfig }) => {
   // Fetch articles.
   console.time(`Fetch Contentful data`)
 
--- a/packages/gatsby-source-contentful/src/gatsby-node.js
+++ b/packages/gatsby-source-contentful/src/gatsby-node.js
@@ -84,11 +84,6 @@ exports.sourceNodes = async (
 
   const pluginConfig = createPluginConfig(pluginOptions)
 
-  const createSyncToken = () =>
-    `${pluginConfig.get(`spaceId`)}-${pluginConfig.get(
-      `environment`
-    )}-${pluginConfig.get(`host`)}`
-
   // add preview/sync capabilitly to token
   let syncToken = await cache.get(CACHE_SYNC_TOKEN)
   let previousSyncData = {
@@ -114,12 +109,20 @@ exports.sourceNodes = async (
   })
 
   // compare data between currentSyncData and previousSyncData and update it
-  previousSyncData.entries.filter(entry =>
-    _.findIndex(currentSyncData.deletedEntries, o => o.sys.id === entry.sys.id)
+  previousSyncData.entries = previousSyncData.entries.filter(
+    entry =>
+      _.findIndex(
+        currentSyncData.deletedEntries,
+        o => o.sys.id === entry.sys.id
+      ) !== -1
   )
 
-  previousSyncData.assets.filter(asset =>
-    _.findIndex(currentSyncData.deletedAssets, o => o.sys.id === asset.sys.id)
+  previousSyncData.assets = previousSyncData.assets.filter(
+    asset =>
+      _.findIndex(
+        currentSyncData.deletedAssets,
+        o => o.sys.id === asset.sys.id
+      ) !== -1
   )
 
   // order is important here, fresh data first
--- a/packages/gatsby-source-contentful/src/gatsby-node.js
+++ b/packages/gatsby-source-contentful/src/gatsby-node.js
@@ -84,7 +84,16 @@ exports.sourceNodes = async (
 
   const pluginConfig = createPluginConfig(pluginOptions)
 
-  // add preview/sync capabilitly to token
+  /*
+   * Subsequent calls of Contentfuls sync API return only changed data.
+   *
+   * In some cases, especially when using rich-text fields, there can be data
+   * missing from referenced entries. This breaks the reference matching.
+   *
+   * To workround this, we cache the initial sync data and merge it
+   * with all data from subsequent syncs. Afterwards the references get
+   * resolved via the Contentful JS SDK.
+   */
   let syncToken = await cache.get(CACHE_SYNC_TOKEN)
   let previousSyncData = {
     assets: [],
@@ -140,15 +149,7 @@ exports.sourceNodes = async (
   // Store a raw and unresolved copy of the data for caching
   const currentSyncDataRaw = { ...currentSyncData }
 
-  // TODO run link resolution on the up to date data
-
-  /*
-   * on sync, we only get the synced data because there isn't enough data
-   * we want to trick the SDK to merge the data
-   * on initial sync, we save all the data
-   * on subsequent syncs, we get the updated data; merge with the old data
-   * if there are duplicate entries we take newest
-   */
+  // Use the JS-SDK to resolve the entries and assets
   const res = client.parseEntries({
     items: currentSyncData.entries,
     includes: {
--- a/packages/gatsby-source-contentful/src/gatsby-node.js
+++ b/packages/gatsby-source-contentful/src/gatsby-node.js
@@ -210,7 +210,6 @@ exports.sourceNodes = async (
   reporter.info(`Deleted entries ${currentSyncData.deletedEntries.length}`)
   reporter.info(`Updated assets ${currentSyncData.assets.length}`)
   reporter.info(`Deleted assets ${currentSyncData.deletedAssets.length}`)
-  console.timeEnd(`Fetch Contentful data`)
 
   // Update syncToken
   const nextSyncToken = currentSyncData.nextSyncToken
@@ -219,6 +218,10 @@ exports.sourceNodes = async (
     cache.set(CACHE_SYNC_KEY, currentSyncDataRaw),
     cache.set(CACHE_SYNC_TOKEN, nextSyncToken),
   ])
+
+  console.timeEnd(`Fetch Contentful data`)
+  reporter.info(`Building Contentful reference map`)
+
   // Create map of resolvable ids so we can check links against them while creating
   // links.
   const resolvable = normalize.buildResolvableSet({
@@ -240,6 +243,8 @@ exports.sourceNodes = async (
     space,
     useNameForId: pluginConfig.get(`useNameForId`),
   })
+
+  reporter.info(`Resolving Contentful references`)
 
   const newOrUpdatedEntries = []
   entryList.forEach(entries => {
@@ -270,6 +275,8 @@ exports.sourceNodes = async (
         )
       }
     })
+
+  reporter.info(`Creating Contentful nodes`)
 
   for (let i = 0; i < contentTypeItems.length; i++) {
     const contentTypeItem = contentTypeItems[i]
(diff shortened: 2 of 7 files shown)


packages/gatsby-source-contentful/src/gatsby-node.js
```diff
@@ -63,23 +63,26 @@ exports.sourceNodes = async (
...
     ]
   ) {
-    syncToken = store.getState().status.plugins[`gatsby-source-contentful`][
-      createSyncToken()
-    ]
+    const pluginState = store.getState().status.plugins[
+      `gatsby-source-contentful`
+    ][createStateKey()]
+    syncToken = pluginState.nextSyncToken
+    previousSyncData = pluginState.previousSyncData
```
Khaledgarbaya: I think this is wrong we might need to merge both data

DSchau:
Awesome @Khaledgarbaya! I'll pull this down and see if I'm able to fix any issues.

Thanks for prioritizing this--it's been a sore spot for some users so super excited to get this fix in as soon as possible!

FLoppix:
Thanks for the fast help and response ! @DSchau and @Khaledgarbaya. Super pumped when this is fixed and let me know when I can test it or you need help / reproduction cases!

Khaledgarbaya:
Sorry closed the PR by mistake.
@DSchau and @FLoppix I think can test with this change and let me know.
I might have some more time tomorrow to write some tests

pieh:
I've published `gatsby-source-contentful@rich-text-resolve` to make it easier to test

FLoppix:
I grabbed the package from @pieh and it still giving me not all the updated values. 
I described it here in more detail https://github.com/gatsbyjs/gatsby/issues/14918

Khaledgarbaya:
Hey @FLoppix Thanks for trying this out. @DSchau any chance I can get the space id and api key for the reproduction space?

Khaledgarbaya:
Hey @FLoppix and @DSchau I think my last change did fix the issue would you mind give it another try.

DSchau:
@Khaledgarbaya sorry - missed this! I’ll DM you on Twitter in 5.

FLoppix:
@Khaledgarbaya Your fix is working for me :)
Thank you so much ! This will help me a lot

odorisioe:
Hi @Khaledgarbaya, **2.1.2-rich-text-resolve.18** solves the hot reloading issue for me (thanks for that!), so I will wait for the merge.

However, I would like to check about an issue with entries included on rich text that could be introduced with this fix. I am not sure if it exists before cause I was not able to see rich text data because the error.

If I update an entry included on a rich text, the hot reload update the entry but does not update the json where it is included.

Example:
I created a content type Page with a rich text called "body" where I can embed VideoBlock entries.
If I update an VideoBlock entry, the reload works and updates the new values on Gatsby but the json still have old values. Only when I update a field on Page (like "title") the json is generated again.

Khaledgarbaya:
Hi @odorisioe, 
Thanks for the test would you mind making a small video about the issue so I can look at it.

odorisioe:
Hi @Khaledgarbaya  I will try to create a video soon but these are the steps to reproduce the error

### Steps
1. (On Contentful) Create two content types
**Page**
. title (field type: string)
. content (field type: rich text - allow to embed VideoBlock entries)
**VideoBlock**
. url (type: string)
. caption (type: string)

2. (On Contentful) Create a Page entry and add a title

3. (On Contentful) Embed a VideoBlock on the field content of the Page created on step 2.

4. (On Gatsby) Run "gatsby develop" with ENABLE_GATSBY_REFRESH_ENDPOINT=true

5. (On Gatsby) Go to /__graphql and query by id the Page and the VideoBlock. The values are sync with Contentful. 
```graphql
# Get created page by id - include the rich text as json
query PageByIdQuery($id: String!) {
  contentfulPage(id: { eq: $id }) {
    id
    content {
      json
    }
  }
}

# Get video block by id (the one embedded on the rich text)
query VideoBlockByIdQuery($id: String!) {
  contentfulVideoBlock(id: { eq: $id }) {
    id
    url
    caption
  }
}
```

6. (On Contentful) Update values on the VideoBlock - url and/or caption fields

7. (On Gatsbty) Do a POST request to /__refresh and the VideoBlock entry will be updated locally.
```
Refreshing source data
Starting to fetch data from Contentful
Fetching default locale
default locale is : en-US
contentTypes fetched 12
Updated entries  1
Deleted entries  0
Updated assets  0
Deleted assets  0
Fetch Contentful data: 3245.120ms
```

8. (On Gatsby) Repeat the same queries as the step 5 and compare the values of the VideoBlock.

### Result
VideoBlockByIdQuery includes the updated values added on step 6.
But the rich-text json on PageByIdQuery includes the VideoBlock with the previous values

Do POST request to /__refresh do not regenerate the rich-text json. 
**Unless** I make a change on the Page entry like edit the title. Then on next hot reloading, the json will be generated with the right values

Khaledgarbaya:
Hey @pieh @DSchau is it possible to release a new version of this to npm so people can play around with it more.

DSchau:
@Khaledgarbaya for sure. I'll do that today!

DSchau:
Successfully published:
 - gatsby-source-contentful@2.1.5-link-resolution-fix.19+787dd4b9f

packages/gatsby-source-contentful/src/fetch.js
```diff
@@ -67,7 +67,7 @@ ${formatPluginOptionsForCLI(pluginConfig.getOriginalPluginOptions(), errors)}`)
 
   let currentSyncData
   try {
-    let query = syncToken ? { nextSyncToken: syncToken } : { initial: true }
+    let query = { initial: true } // always initial to be able to resolve links
```
disintegrator: Hey @Khaledgarbaya, what is the impact of this change on spaces with _a lot_ of content?

Khaledgarbaya: I am trying to find another solution for this but for now, this will pull all the data.

sidharthachatterjee:
@odorisioe Can you confirm if the latest version of this fixes your issue? if yes, I'll go ahead and get this published this today 

Cc @Khaledgarbaya

packages/gatsby-source-contentful/src/fetch.js
sidharthachatterjee: @Khaledgarbaya This does seem problematic for very large spaces. Is this fix worth this impact or shall we wait for a better solution? 

@pieh Thoughts?

Khaledgarbaya: @sidharthachatterjee I think so too. I will try probably this week another approach and see.

bernhardw:
Not sure if it helps, but I have the same problem of missing content of rich text fields on hot-reloads during development and in production. Manually reloading the page solves the issue. It happens both with the current version `2.1.19` and with `2.1.5-rich-text-resolve.19`.

Khaledgarbaya:
Hey Folks,

Let me give some update on this.

I jumped on a call with @DSchau last Friday to discuss the issue.

Here's is what I believe is happening: 

The source plugin uses the Contentful js SDK under the hood which does the link resolution but the SDK can't resolve the links if the linked entry does not exist in the payload which is the case when using the Contentful sync API more specifically on non-initial sync.

I thought of 2 options to fix this :

1. Always use the initial sync but this will become a problem when the website is so big and also when using the Gatsby preview.

2. Save the old data and merge it with the new data and do a link resolution. We need to compare new data with old data and override the entries from the old payload with the same ID. After that, we can use the method `[parseEntries](https://contentful.github.io/contentful.js/contentful/7.9.1/ContentfulClientAPI.html#.parseEntries)` exposed by the SDK to resolve all the links and refresh the Nodes.

By now we think we are all in favor of option 2 which I think is the ideal approach but it might take some time. I will try my best to tackle it this week but also meanwhile I'll appreciate any help that you can provide in any form (code, test etc..)

I appreciate all your collaboration and feedback and hopefully, we'll manage to get this fixed soon

Cheers,
Khaled

MattBrooksManifesto:
Hi @Khaledgarbaya - thanks for the update above - do you have an ETA for this to be merged?

DSchau:
🚨 Extra extra, read all about it 🚨

@Khaledgarbaya and I paired on this this morning, and as of a660d32, rich text embeds seem to be persisting across reloads as we're sort of "hydrating" the data from the initial sync with updates while still keeping the embed content around.

This has been published as `gatsby-source-contentful`@`rich-text-resolve` and can be installed like so:

```shell
yarn add gatsby-source-contentful@rich-text-resolve
# or
npm install gatsby-source-contentful@rich-text-resolve
```

## Work Remaining

We're heavily leveraging the cache to serve as the base for our content, so we'll need to ensure that this is invalidated at the right moment (e.g. we need to fetch entirely new data if we swap access token, space id, etc.) so that data is always sparkly fresh.

1. [ ] Validate non-happy path (e.g. cache invalidation, space id changes, etc.)
1. [ ] Add a lot more tests to ensure functionality works as intended
1. [ ] Clean up some of the janky areas of the code

## How can _you_ help?

Test the version above on _your_ projects. Let us know if any issues arise, and specifically whether this solves the rich text resolve issue.

MattBrooksManifesto:
Hi @DSchau @Khaledgarbaya 

We've tested the updated version on our local instance this morning, and unfortunately, we have to report that it's not been successful in resolving the issue.

- We are now seeing the preview screen go white when we make any text changes (eg body copy). A reload of the page then updates with any text changes
- Embedded content items are not pulling through when added to a rich text block (even on reload after white screen issue)
- A restart of the instance DOES display embedded content but making any changes such as removing 1 embedded item results in all other embedded items disappear

Happy to provide access to our instance it that's useful

disintegrator:
@Khaledgarbaya is it possible the sync api can be fixed to correctly serve rich text? Perhaps the api response can contain the full rich text AST regardless of what parts of it changed. Right now sync API returns bad data for rich text fields and forcing a full sync on every build kills build performance on large projects.

packages/gatsby-source-contentful/src/gatsby-node.js
```diff
@@ -94,6 +96,44 @@ exports.sourceNodes = async (
...
+
+  /*
+   * on sync, we only get the synced data because there isn't enough data
+   * we want to trick the SDK to merge the data
+   * on initial sync, we save all the data
+   * on subsequent syncs, we get the updated data; merge with the old data
+   * if there are duplicate entries we take newest
+   */
+  const res = client.parseEntries({
+    items: currentSyncData.entries,
```
gambry: According to the documentation, `parseEntries()` is expecting raw JSON. I believe what we are passing is not a plain object?

packages/gatsby-source-contentful/src/gatsby-node.js
```diff
@@ -94,6 +96,44 @@ exports.sourceNodes = async (
...
+    includes: {
+      assets: currentSyncData.assets,
```
gambry: The [doc is not very clear](https://contentful.github.io/contentful.js/contentful/7.13.0/ContentfulClientAPI.html#.parseEntries) about what this `includes` is meant to do. If I should guess its meaning from [EntityCollection](https://contentful.github.io/contentful.js/contentful/7.13.0/Entities.html#.EntryCollection) includes the raw form of the links already present in the items.

Is what we are passing plain JSON?

packages/gatsby-source-contentful/src/gatsby-node.js
```diff
@@ -139,11 +179,10 @@ exports.sourceNodes = async (
   // Update syncToken
   const nextSyncToken = currentSyncData.nextSyncToken
 
-  // Store our sync state for the next sync.
-  const newState = {}
-  newState[createSyncToken()] = nextSyncToken
-  setPluginStatus(newState)
-
+  await Promise.all([
+    cache.set(CACHE_SYNC_KEY, currentSyncData),
```
gambry: This doesn't seem to work. The cache is not set, and IMO probably because currentSyncData is not string-safe, or a plain object?

Thebohemiangeek:
how is it goings does this solves the issue what ar the drawbacks of using rich-text-resolve ?

packages/gatsby-source-contentful/src/gatsby-node.js
gambry: Also I forgot to mention this bit generates a **Maximum call stack size exceeded** error on `node_modules/contentful/dist/contentful.node.js walkMutate()`. Probably because what's passed is not a JSON? Not sure.

With this on the latest version of `rich-text-resolve`  (`2.1.36-rich-text-resolve.14`) doesn't work at all.

gambry:
> how is it goings does this solves the issue what ar the drawbacks of using rich-text-resolve ?

AFAIK the current issue is triggered only if you Preview/Hot Reload/Refresh your content instance. If you build once, either locally or Prod environments, the issue shouldn't happen.

Also in the case you Preview/Hot Reload/Refresh version `gatsby-source-contentful@2.1.5-rich-text-resolve.19` works, although is not performant due fetching initial data all the time.

packages/gatsby-source-contentful/src/gatsby-node.js
gambry: Yep, error message on setting the cache is `Converting circular structure to JSON`

gambry:
I got this working with really few changes from the current WIP. I'm happy to create a MR if required, but briefly:

- on `fetch.js`: add `resolveLinks: false` as client option, so links are not resolved, fetched data doesn't have circular references and so object is cache-able (see comment https://github.com/gatsbyjs/gatsby/pull/15084#discussion_r365310633)
- on gatsby-node.js: before processing `currentSyncData` with `client.parseEntries()` clone it over like `const currentSyncDataUnparsed = {...currentSyncData};` and cache `currentSyncDataUnparsed` rather than `currentSyncData` as the latter may now have circular references and so it's not cacheable.

As a results I get pretty much everything working as expected, hot-reloads are fast and no issues with richtext or any other links.
Worth mentioning because entries links are not resolved anymore you may need to review and update your schema as auto-inferring may have not defined all the fields. On my local test only 2 out of about 50-60 fields are causing issues, which is not that bad.

**What's still missing**:

1. A good tidy up of the code would be good. Also I'm not completely sure why current WIP is swapping the state() with cache()? What's the point of persisting fetched data across builds? Don't hot-reload and Gatsby Preview-like system work on a running instance? Using back state() will help with reducing the work for finding a invalidation strategy.
2. Give reliable report numbers, as currently even if a single entry is updated the console still reports "Updated entries" with the TOTAL number of entries. Same for assets.

benrobertsonio:
@gambry do you have your changes on a branch / PR somewhere it could be reviewed?

gambry:
Not yet @benrobertsonio , I've been debugging and working through the package. But I'll apply the changes to a fork and create a PR.

gambry:
@benrobertsonio I created https://github.com/gatsbyjs/gatsby/pull/21165 .

I'm not primarily a gatsby developer, so please take that PR as an improvement to this one. I can't answer most of the questions about changes proposed in this PR. I can only answer the 2 additions RE not-resolving links on fetching and making the fetched object cacheable.

I've personally have - for example - some reservation on swapping `state` with `cache`. As hot-reloads happens in a running-instance context AFAIK the former may have better performances and the latter also introduces questions on how/when to invalidate the cache. But as said above I'm not the author of that bit so I can't fully answer what the best choice would be.

pvdz:
(I've converted this PR into a Draft PR.)

packages/gatsby-source-contentful/src/gatsby-node.js
axe312ger: By raw json the documentation means raw api output. That is what we are actually creating here manually.

This should be fine :)

packages/gatsby-source-contentful/src/gatsby-node.js
axe312ger: I guess we have to store the data in a raw state as these won't contain any circular references.

axe312ger:
FYI: Working on this right now

Khaledgarbaya:
@axe312ger @pvdz should we change this to a normal PR since Benedikt is working on it?

axe312ger:
Yes please change :)

pvdz:
You should be able to change this yourself. I had put it to draft because that's what it was and there was no work at the time. Draft PR's kinda fly under our radars. So it's totally up to you to change it back to review mode when ready for it :)

packages/gatsby-source-contentful/src/gatsby-node.js
axe312ger: Includes are all entries that are referenced from entries but itself are not part of the result list.


We will use the SDKs internal API here to resolve the custom request we creating out of the latest sync delta and the cached sync API responses.

packages/gatsby-source-contentful/src/gatsby-node.js
```diff
@@ -47,8 +49,11 @@ exports.sourceNodes = async (
   },
   pluginOptions
 ) => {
-  const { createNode, deleteNode, touchNode, setPluginStatus } = actions
-
+  const { createNode, deleteNode, touchNode } = actions
+  const client = createClient({
+    space: `none`,
```
Khaledgarbaya, line 54: @axe312ger is this intentional ? should we pull the token and the space id from the plugin option

axe312ger: @Khaledgarbaya I just used what was in the PR, probably your own code :P


I think we need to pass them otherwise `createClient` would fail to to missing config options. But as we do not connect to the space and use the SDK just for resolving the "fake" response, we can pass any space id and token

Khaledgarbaya: 🤦 make total sense

axe312ger:
Thanks @phoebeschmidt  & @Khaledgarbaya 

So lets continue for #24905

PR #15090, opened 2019-06-24
feat(themes): Standardize architecture across themes

amberleyromo:
## Description

Changes to make the blog and notes themes more consistent with each other. Main thing left in this area remains the single concrete type / schema customization from @ChristopherBiscardi.

- Separates page templates (data wrappers) and page components (consumers) as in the notes theme.
- Removes demo content from blog theme
- Various changes toward standardizing the approach in gatsby-node.
- Makes concept of `basePath` consistent across themes
- Adds a third starter so that we now have a starter for each theme, plus a starter for themes composed together:
  - `gatsby-starter-blog-theme` - for the blog theme (alone)
  - `gatsby-starter-notes-theme` - for the notes theme (alone)
  - `gatsby-starter-theme` for both, composed together

## Related Issues

Related #14276


themes/gatsby-theme-blog/gatsby-node.js
```diff
@@ -1,116 +1,158 @@
...
+// These templates are simply data-fetching wrappers that import components
+const PostTemplate = require.resolve(`./src/templates/post`)
+const PostsTemplate = require.resolve(`./src/templates/posts`)
+
+// Ensure that content directories exist at site-level
+exports.onPreBootstrap = ({ store }, themeOptions) => {
+  const { program } = store.getState()
+
+  basePath = themeOptions.basePath || `/`
+  contentDir = themeOptions.contentDir || `posts`
```
jlengstorf: Didn't we choose `contentPath` for this? That's what I'm teaching in the egghead course. 😕

themes/gatsby-theme-blog/src/components/post-footer-content.js
```diff
@@ -24,15 +24,15 @@ export default ({ previous, next }) => (
       >
```
jlengstorf: I thought we were moving this out to the post-footer? Or maybe I'm misremembering.

themes/gatsby-theme-blog/gatsby-node.js
amberleyromo: You are right, I double-checked - That's my mistake. Will change.

themes/gatsby-theme-blog/src/components/post-footer-content.js
amberleyromo: I don't remember that, but I don't have notes on it in my doc... We were going for the idea of consolidating content components for people to shadow, and then walking up in complexity from there.

themes/gatsby-starter-theme/gatsby-config.js
```diff
@@ -3,7 +3,8 @@ module.exports = {
     {
       resolve: `gatsby-theme-notes`,
       options: {
-        mdx: false,
+        mdx: true,
```
jxnblk: @ChristopherBiscardi are we still going to need to disable MDX like this for the time being? Wasn't sure what's needed to dedupe the MDX plugins or where we left off with that

johno: Yeah, for now we'll have to turn it off with this API since we haven't implemented a way for plugins to merge configs yet (AFAIK).

themes/gatsby-theme-blog/gatsby-config.js
```diff
@@ -1,88 +1,65 @@
...
+          path: options.assetPath || `content/assets`,
+          name: options.assetPath || `content/assets`,
+        },
+      },
+      `gatsby-transformer-sharp`,
+      `gatsby-plugin-sharp`,
+      `gatsby-plugin-react-helmet`,
+      `gatsby-plugin-twitter`,
+      `gatsby-plugin-emotion`,
+    ],
```
johno: I _think_ we're going to want to add a filter here in case `mdx` is set to false.

```suggestion
    ].filter(Boolean),
```

amberleyromo:
- Thanks for the catch on the filtering, committed that @johno 

On the content question @jlengstorf (https://github.com/gatsbyjs/gatsby/pull/15090#discussion_r298376156) @jxnblk (https://github.com/gatsbyjs/gatsby/pull/15090#pullrequestreview-255809472):

I don't believe we actually decided in that meeting, but I think pulling the content out of these shadowed components and into the config would be a poor authoring experience, as is. I perhaps didn't speak up enough when we all chatted. I think it would be cool to have all that content set up just through the config, but the way it is now is unideal enough that I don't like it.

Also, what we can do is go ahead and add the shadowed content components in the starters, so that what's necessary to shadow is clear from the get-go.

I'd like to merge this as-is for now.

PR #15091, opened 2019-06-24
chore(www): Tidy `assets` folder

fk:
- consolidate Futura PT webfont imports
- ➡️ move
  - font files from `src/fonts` to `src/assets/fonts` — this is where they (can) belong, and we should — not need to touch them again for a long time, so lets get them out of the way/editor file tree
  - SVGs for mobile navigation to `src/assets/icons` (that might change … WIP)
  - vendor logos used on `/features/` to `src/assets/vendor-logos` (that might still change, too)
- 🗑 remove
  - unused Squarespace logos `squarespace.jpg`, `squarespace.png`, `squarespace-small.png`
  - unused `bg-featured-sites.svg`, `showcase-magic.svg`
  - `src/used-by` and related logo SVGs — we have been using this for the homepage, but it is not very likely for this to come back (at least in this form, and with this content); lets not keep dead code around
  - `src/logo`, `src/monogram` SVGs to `src/assets`
- ↪️ replace `info-icon.svg` used on `/features/` with a Material Design icon, via `react-icons` (cc @gillkyle)


fk:
Oh … forgot about this. I'm not sure if it's ready to merge, actually—gotta take another look. Will do so today. I thought I had this one marked as draft. :-(

Thank you Sid!

fk:
This is not really finished, but good enough to go. Will push further in another PR. ✌️

PR #15108, opened 2019-06-25
docs(gatsby-plugin-preact): Change install script

rayriffy:
## Description

Current version of `preact` and `preact@latest` is at **8.x.x** which is **incompatible** with Gatsby v2

In this PR, I change install script to install `preact@beta` which is version **10.x.x-betax** instead to prevent anyone from confusion.


rayriffy:
Before merging I need to ask something

`preact@beta` is currently in **beta.1** but `preact@next` is now in **beta.3** which is less buggy than beta.

Which version would you prefer in this PR?

packages/gatsby-plugin-preact/README.md
```diff
@@ -10,7 +10,7 @@ React.
 
 ## Install
 
-`npm install --save gatsby-plugin-preact preact@beta preact-compat`
```
rayriffy: Since PreactX have it's own `preact/compat`. It unnecessary to add `preact-compat` into package.

PR #15116, opened 2019-06-25
chore(starters): add bunch of starters

smakosh:
## Description
Added my personal starters:
- Airtable starter
- Apollo starter
- Rest API starter with Auth
- Portfolio one page starter


endormi:
Well 😅

freiksenet:
@smakosh We can merge it once you fix the tags, see Danger result.

smakosh:
Yup, we need to add the new tags I added on this PR

smakosh:
All looks good now @endormi @freiksenet

docs/categories.yml
```diff
@@ -54,6 +54,13 @@ starter:
   - Styling:None
   - Testing
   - Transitions
+  - Apollo
```
freiksenet: Please don't add new tags.

smakosh: it was mentioned on Gatsby's docs: https://www.gatsbyjs.org/contributing/submit-to-starter-library/#adding-new-tag

freiksenet: While we support adding tags, we strongly discourage that. We have had plenty of tags that only have 1 instance and that made the UI confusing. We've just made a cleanup to remove them. Therefore the correct solution for your starters would be not adding new tags, but removing tags that don't exist from your entries.

Thank you!

freiksenet:
Thank you very much @smakosh!

Issue #14693, opened 2019-06-10
[www]: Fix broken internal links


gillkyle:
I did a sweep of the site to try and amend issues with broken links, running a [link checker](https://www.npmjs.com/package/broken-link-checker) gave me a report of about 300 broken links (some are redundant and show up on multiple pages, see the [gist](https://gist.github.com/gillkyle/ef24d6c991756c326951ae69089f6503) for full output).

The links to fix are the following:

## Broken URLs 

- [x] (#14960) [https://www.gatsbyjs.org/packages/gatsby/src/utils/create-content-digest.js](https://www.gatsbyjs.org/packages/gatsby/src/utils/create-content-digest.js (HTTP_404)) (HTTP_404)
- [x] [https://www.gatsbyjs.org/docs/gatsby-in-the-enterprise/](https://www.gatsbyjs.org/docs/gatsby-in-the-enterprise/ (HTTP_404)) (HTTP_404) - (this appears to be broken in the sidebar of all docs pages because of a `createRedirect` in `/src/www/gatsby-node.js` that is unnecessary, removing it seems like it would be enough to fix all of these issues)
- [x] [https://www.gatsbyjs.org/packages/gatsby/](https://www.gatsbyjs.org/packages/gatsby/ (HTTP_404)) (HTTP_404)
- [x] [https://www.gatsbyjs.org/plugins/gatsby-transformer-remark](https://www.gatsbyjs.org/plugins/gatsby-transformer-remark (HTTP_404)) (HTTP_404)
- [ ] [https://www.gatsbyjs.org/offline](https://www.gatsbyjs.org/offline (HTTP_404)) (HTTP_404)
- [x] [https://www.gatsbyjs.org/docs/gatsby-vendor-partnership/gatsbyjs.org](https://www.gatsbyjs.org/docs/gatsby-vendor-partnership/gatsbyjs.org (HTTP_404)) (HTTP_404)
- [ ] [https://www.gatsbyjs.org/docs/mdx/programmatically-creating-pages/api-reference/options/extensions](https://www.gatsbyjs.org/docs/mdx/programmatically-creating-pages/api-reference/options/extensions (HTTP_404)) (HTTP_404)
- [ ] [https://www.gatsbyjs.org/contributing/events/meenhacks.github.io](https://www.gatsbyjs.org/contributing/events/meenhacks.github.io (HTTP_404)) (HTTP_404) - (this looks it's being interpreted as a relative url, not an absolute url)

[...]

PR #15139, opened 2019-06-26
Fixes broken link from Gatsbygram

kindavishal:
## Description

Fixes a broken link to ```packages-gatsby``` and closes the third task inside #14693
## Related Issues

Closes third task in #14693

### DIFF
--- a/docs/blog/gatsbygram-case-study/index.md
+++ b/docs/blog/gatsbygram-case-study/index.md
@@ -374,7 +374,7 @@ generates all the configuration for you.
 
 Normally page resources are pre-cached with a service worker. But as several
 browsers (Safari/Microsoft Edge) still don't support Service Workers, the
-[Gatsby `<Link>` component](/packages/gatsby/) pre-caches resources for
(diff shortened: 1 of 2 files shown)


kindavishal:
Reverted the last commit to open a new PR which fixed another issue. I was working on the same branch, so forgot to switch!
The new PR is here https://github.com/gatsbyjs/gatsby/pull/15142

kindavishal:
Thank you!

m-allanson:
Thank you @kindavishal! Can I ask you to update these links in a new PR? They should be relative instead of absolute URLs. See my comments on https://github.com/gatsbyjs/gatsby/pull/15142 for an example.

The reason for relative urls is that when using absolute urls, clicking the link will trigger a full page refresh. Using relative urls means that Gatsby's Link component can prefetch the page content and update the window without having to refresh the page.

kindavishal:
> Thank you @kindavishal! Can I ask you to update these links in a new PR? They should be relative instead of absolute URLs. See my comments on #15142 for an example.
> 
> The reason for relative urls is that when using absolute urls, clicking the link will trigger a full page refresh. Using relative urls means that Gatsby's Link component can prefetch the page content and update the window without having to refresh the page.

Got it! Will do that :D

Issue #13475, opened 2019-04-19
[docs] Breadcrumbs for all docs pages


lannonbr:
## Summary

I am thinking it would be beneficial to have a component at the top of each docs page to show the hierarchy of where the current page is in the docs.

### Basic example

For instance, the [Working With Images in Gatsby](https://www.gatsbyjs.org/docs/working-with-images/) page could have a breadcrumbs of

```
Docs / Guides / Images, Files, and Video / Working With Images in Gatsby
```

And then you can click any point in the chain to walk up the hierarchy.

---

## Acceptance criteria

Acceptance criteria by @marcysutton (added by @prestonso)

- As a user with a larger viewport, I should encounter a link “home” to get a clear path to the homepage.
- As a user, I should encounter links in the breadcrumb list for all pages in the docs hierarchy except the current one, which is in plain text.
- As a user, I should see breadcrumbs displaying my current location in the site’s hierarchical structure, not my browser session history.
- As a user with a smaller viewport, I should see only the last item in the breadcrumb trail (the link that is the parent of the current page in view)
- As a sighted or low-vision user, I should see links and text with adequate color contrast.
- As a screen reader user, I should hear the widget announced as a nav landmark with a unique label to identify it from other landmarks.
    - This can be done with `<nav aria-label=”breadcrumb”>`
- As a screen reader user, inside the widget I should find useful links and current item text announced without repetitive separator characters.
    - Separators like `>` can be marked as presentational with `<span role=”presentation”>` or similar.

marcysutton:
Great intuition here! We've talked about it this week, actually. I think it could fit nicely into the #6896. I saw an example from @calcsam from Django where they have a nice "you are here" section in the sidebar. https://docs.djangoproject.com/en/2.2/howto/outputting-csv/

lannonbr:
Yeah I think these two issues could go hand in hand where mine is for the hierarchy of the global docs while Shannon's will dig into the local hierarchy of an individual page

andrewfairlie:
This would be great, particularly on mobile where the sidebar is collapsed it's currently sometimes quite hard to get a feel for where you are in the site structure.

shannonbux:
Yeah, @fk and I talked about breadcrumb trails maybe a year ago, too and I'd love to pair w/ @lannonbr and @marcysutton on this! 

[...]

PR #15165, opened 2019-06-26
feat(www): added breadcrumb to docs

gillkyle:
Co-authored-by: @mikeattara

## Description

Me and @mikeattara paired and worked this out together while pair programming 🎉 super fun!

breadcrumb

## Related Issues

Addresses #13475


KyleAMathews:
Nice!

Could we make them a bit less visually prominent? E.g. make them smaller and remove the underline?

KyleAMathews:
Also they could use a bit of margin between them and the header. 1/2 rhythm should work.

gillkyle:
Just made updates from the suggestions above, this is how it renders now:

image

KyleAMathews:
Nice!

marcysutton:
I agree the smaller font size works well! I will offer a counter-point for the link underlines that now we're only differentiating them with color, so my vote would be to keep them.

gillkyle:
Newest commit, and underlines are back:

image

mikeattara:
Good work @gillkyle

www/src/components/docs-breadcrumb.js
```diff
@@ -0,0 +1,90 @@
...
+        <Separator />
+        {activeItemParents.reverse().map(item => (
+          <React.Fragment key={item.title}>
+            <span>
+              <Link to={item.link}>{item.title}</Link>
+            </span>
+            <Separator />
+          </React.Fragment>
+        ))}
+        {activeItem.title}
```
marcysutton:
```suggestion
        <span aria-current="page">{activeItem.title}</span>
```
I think we should mark the current page somehow. `aria-current` is the recommended way to do that for assistive tech, and after consulting the [ARIA spec](https://www.w3.org/TR/wai-aria-1.1/#aria-current) it looks like it's allowed on any element.

gillkyle: I love getting this kind of feedback! Thanks for catching!

www/src/components/docs-breadcrumb.js
```diff
@@ -0,0 +1,90 @@
...
+const BreadcrumbNav = ({ children, mobile = false }) => (
+  <nav
+    aria-label="breadcrumb"
+    css={{
+      fontSize: fontSizes[1],
+      display: `${mobile ? `inherit` : `none`}`,
+      [mediaQueries.md]: {
+        display: `${mobile ? `none` : `inherit`}`,
+      },
+      marginBottom: rhythm(1 / 2),
```
fk: There's nothing wrong with still using `rhythm`, but I think we should stick to the design tokens.

www/src/utils/sidebar/get-active-item-parents.js
```diff
@@ -15,7 +15,7 @@ const isParentActive = (sections, parentTitle) => {
 const getActiveItemParents = (itemList, activeItemLink, activeItemParents) => {
   if (activeItemLink.parentTitle) {
     const bar = isParentActive(itemList, activeItemLink.parentTitle)
-    activeItemParents.push(bar.title)
+    activeItemParents.push(bar)
```
fk: Lol at `bar` (that was me) 🤦‍♂ … naming things is hard, but this one was lazy-stupid.

www/src/components/docs-breadcrumb.js
```diff
@@ -0,0 +1,90 @@
+import React from "react"
+import { Link } from "gatsby"
+import getActiveItem from "../utils/sidebar/get-active-item"
+import getActiveItemParents from "../utils/sidebar/get-active-item-parents"
+import { mediaQueries, space, fontSizes } from "../utils/presets"
+import { rhythm } from "../utils/typography"
+
+const Separator = ({ character = `>` }) => (
```
fk: I have to say I dislike using `>` (and `<`); they feel clunky, IMO they should be less wide. Maybe we can try a chevron (`react-icons/lib/md/chevron-right`)?

mikeattara: I think that would be cool. What's your thought? @gillkyle

marcysutton: That sounds good to me! I actually considered leaving feedback about the separator character, but didn't feel as strongly about it. Using a chevron would be slightly more sophisticated though.

gillkyle: Sounds like a great idea! Been swamped today so I could get to it later or if @mikeattara wants to take a stab at it that'd be great

gillkyle: Looks a lot better that way: 
image

www/src/components/docs-breadcrumb.js
```diff
@@ -0,0 +1,91 @@
...
+  // return a shorter version of the breadcrumb on the intro page
+  // because the docs intro page isn't generated from markdown
+  if (activeItem.title === `Introduction`) {
+    return (
+      <>
+        {/* only the breadcrumb nav of the proper viewport is displayed */}
+        <BreadcrumbNav>
+          <Link to="/">Home</Link>
+          <Separator />
+          Docs
```
marcysutton: This renders on the Tutorials and Contributing sections, but still says `docs`. We should fix that.

gillkyle: Just fixed this, now it will take the top level of the itemsList passed in and use it for the link and title displayed there which should reflect the highest item in sidebar

www/src/utils/sidebar/get-active-item-parents.js
```diff
@@ -15,7 +15,7 @@ const isParentActive = (sections, parentTitle) => {
 const getActiveItemParents = (itemList, activeItemLink, activeItemParents) => {
   if (activeItemLink.parentTitle) {
     const bar = isParentActive(itemList, activeItemLink.parentTitle)
-    activeItemParents.push(bar.title)
```
marcysutton: This gets tricky to update the Tutorials section, which is a bit of a unique case. Currently the top-level item says "Tutorial (step-by-step)" but it's a parent to both Additional Tutorials and the upcoming Themes Tutorials, so this title doesn't make sense to be a breadcrumb parent for those. But in the context of the main tutorial where this heading shows in the sidebar, it does make sense. 

I think the best solution would be to add a `breadcrumbTitle` or something to optionally alter the breadcrumb display since even though we could split on ` (` and remove the latter part, it still isn't quite right and mucking with pluralization of a single word seems like a bad way to go. I wanted to add my comment here since I tried updating it myself to move this forward, but had to switch gears to something else for the themes launch. So we can get to this when we get to it!

gillkyle: Alright I can get to this later then. My first solution actually just pulled the top level route: `/tutorial/part-one` => `tutorial` or `/docs/guides/plugins` => `docs` but maybe that wouldn't be the best solution either. Do you mean add a breadcrumbTitle to the sidebar links yaml files?

As far as comments go in those util functions, I can add some context there too.

marcysutton: Ah yes, that's probably thanks to our evolving (and confusing) IA! 😅 

Yeah, I meant a value in the tutorial-links yaml file. I should have clarified! This isn't a huge rush, I just wanted to comment while it was fresh in my mind. Thanks for your work on it!

gillkyle: I just pushed an update for this. It was pretty simple and works for those top level edge cases where the hierarchy might be a little weird.

marcysutton:
Circle e2e tests never seemed to run....but I think this is ready to merge.

lannonbr:
@marcysutton Yeah there's a config that if the PR's branch starts with docs, it will not run some of the things in CI

marcysutton:
Merged! Woo! Thanks all for your work on this.

mikeattara:
Happy to see this go live...

www/src/components/docs-breadcrumb.js
```diff
@@ -0,0 +1,101 @@
...
+    return (
+      <>
+        {/* only the breadcrumb nav of the proper viewport is displayed */}
+        <BreadcrumbNav>
+          <Link to="/">Home</Link>
+          <Separator />
+          {topLevelTitle}
+        </BreadcrumbNav>
+        <BreadcrumbNav mobile>
+          <Separator character="<" />
```
muescha: should it not be the ChevronLeft here also?

```suggestion
          <Separator character={<ChevronLeft />} />
```

www/src/data/sidebars/doc-links.yaml
```diff
@@ -28,54 +29,72 @@
...
             - title: Customizing Babel.js Config
               link: /docs/babel/
+              breadcrumbTitle: Babel.js Config
             - title: Using Babel Plugin Macros
               link: /docs/babel-plugin-macros/
-            - title: Customizing Webpack Config
+              breadcrumbTitle: Babel Plugin Macros
+            - title: Adding a Custom Webpack Config
               link: /docs/add-custom-webpack-config/
+              breadcrumbTitle: webpack Config
```
muescha: uppercase?
```suggestion
              breadcrumbTitle: Webpack Config
```

www/src/data/sidebars/doc-links.yaml
```diff
@@ -92,37 +111,50 @@
...
                 # Instead, please add links on the /docs/headless-cms/ page
                 - title: Sourcing from Contentful
                   link: /docs/sourcing-from-contentful/
+                  breadcrumbTitle: Contentful
                 - title: Sourcing from Netlify CMS
                   link: /docs/sourcing-from-netlify-cms/
+                  breadcrumbTitle: Netlify CMS
                 - title: Sourcing from WordPress
                   link: /docs/sourcing-from-wordpress/
+                  breadcrumbTitle: Wordpress
```
muescha: uppercase?
```suggestion
                  breadcrumbTitle: WordPress
```

www/src/data/sidebars/doc-links.yaml
muescha: fix in #15643

www/src/data/sidebars/doc-links.yaml
muescha: fix in #15643

www/src/components/docs-breadcrumb.js
gillkyle: Yeah that's a good catch! Wanna add that into your PR in #15643?

muescha: ok - included in PR #15643

PR #15173, opened 2019-06-27
Add config variable to avoid canonical redirect

xavivars:
## Description

This PR introduces a config variable that allows you to disable redirect when a page loads, in case the paths are different when generated and when displayed.

This used together with https://github.com/wardpeet/gatsby-plugin-static-site allows to use gatsby as a static page generator instead of a static site generator

## Related Issues

This PR is basically the same one as https://github.com/gatsbyjs/gatsby/pull/11914 and fixes https://github.com/gatsbyjs/gatsby/issues/4337

### DIFF
--- a/examples/using-__DISABLE_LOAD_TIME_CANONICAL_REDIRECT_CHECK__/.eslintrc.json
+++ b/examples/using-__DISABLE_LOAD_TIME_CANONICAL_REDIRECT_CHECK__/.eslintrc.json
@@ -1,0 +1,8 @@
+{
+  "env": {
+    "browser": true
+  },
+  "globals": {
+    "graphql": false
+  }
+} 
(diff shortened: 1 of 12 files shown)


KyleAMathews:
Generally speaking we don't add config options. Is there another way we can solve this?

xavivars:
I'd love to have another way to do this, but I can't think of another way to break this `if`

https://github.com/gatsbyjs/gatsby/blob/56950e6bd7da08a1a353d3d3ca1b711a786e1b43/packages/gatsby/cache-dir/production-app.js#L77

Up until 2.9, there was a workaround, that involved changing what is currently `window.pagePath` to browser's location hooking into `clientEntry` but that stopped working due to the new per-page `page-data.json` relying on that `window.pagePath`.

I don't now if creating a new hook at that point in the code is preferable to a config variable...

xavivars:
Maybe being able to override here the value of `browserLoc` to not to use the real browser `window.location` could also work, but I have no idea on how that could be done... 

https://github.com/gatsbyjs/gatsby/blob/0fa5955bde8b2e7c4dce6f0e88fc35a7fb432726/packages/gatsby/cache-dir/production-app.js#L65

xavivars:
This PR takes a much simpler approach, and enables the same thing:

https://github.com/gatsbyjs/gatsby/pull/15180

wardpeet:
Hey @xavivar, thank you for opening this pull request!

Unfortunately, the proposed change is not going to be merged. This is a fairly unusual usecase, and while we'd love to support every usecase for everyone we don't have the capacity to do that right now. It may seem like a small change but any time we expose apis publicly like this it means we're committed to supporting them, which can lead to problems when we change them later on.

PR #15178, opened 2019-06-27
Fixes No instruction to import header to index.js

### LINKED ISSUE #15098: No instruction to import header to index.js
## Description

In [Part One](https://www.gatsbyjs.org/tutorial/part-one/#-using-the-link--component) (Section "Using the <Link /> component") of the Gatsby tutorial, there is no instruction to add `import Header from "../components/header` to `index.js`. This line just appears in the code block of `index.js`. 

In a previous step, the user is instructed to add this line to `about.js`. The user should be instructed to also add this line to `index.js` to ensure their code matches that of the tutorial.

### Steps to reproduce

Follow Part one of the Gatsby tutorial

### Expected result

User should be instructed to add `import Header from "../components/header` to `index.js`.

### Actual result

The line just appears in the code block of `index.js` without any instruction given to the user.

### Environment

[Part One](https://www.gatsbyjs.org/tutorial/part-one/#-using-the-link--component) (Section "Using the <Link /> component") of the Gatsby tutorial

mehtavishwa30:
## Description

Added instructions to import Header component in the index.js file.

## Related Issues

Fixes #15098

### DIFF
--- a/docs/tutorial/part-one/index.md
+++ b/docs/tutorial/part-one/index.md
@@ -284,7 +284,10 @@ You'll often want to link between pages — Let's look at routing in a Gatsby si
 
 ### ✋ Using the `<Link />` component
 
-1.  Open the index page component (`src/pages/index.js`), import the `<Link />` component from Gatsby, add a `<Link />` component above the header, and give it a `to` property with the value of `"/contact/"` for the pathname:
+1.  Open the index page component (`src/pages/index.js`) 
+* Import Header by adding the line `import Header from "../components/header"` to the `index.js page`
+* Import the `<Link />` component from Gatsby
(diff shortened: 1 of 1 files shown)


alexdkaminski:
Thanks for this Vishwa.

marcysutton:
Hi there, thanks for the contribution! This detail is covered earlier in the tutorial, so I don't think it makes sense to add here because people following along will have already done it. If you go back to the section on [Using sub-components](https://www.gatsbyjs.org/tutorial/part-one/#-using-sub-components), that's where it's mentioned. The relevant parts of this section on Linking between Pages are highlighted in a "diff" format with a different color. 

But we absolutely still want to have you contributing to Gatsby, so we'd welcome future issues and PRs. let us know if you need any guidance, and thanks again!

PR #15184, opened 2019-06-27
[WIP] ci: trial out windows support

DSchau:
## Description

Test Windows integration with CircleCI

Goal: to run unit tests in a version of Node (probably 10, maybe 12 if I’m feeling cute)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -215,7 +215,10 @@ jobs:
     working_directory: c:\\cache
     steps:
       - checkout
-      - run: node --version # this probably won't work YOLO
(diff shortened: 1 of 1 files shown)


DSchau:
OK - well this “works” but installing Node.js is kind of a nightmare.

What worked:

- Able to access Windows environment
- Able to install chocolately
- Able to install nodejs/yarn

What didn’t work:

- Updating path _permanently_ to account for the global installs, which live in `c:program files/nodejs` and `c:program files (x86)/yarn/bin`

Close, but no cigar!

wardpeet:
Why bother? Azure will always have the best windows support.

PR #15185, opened 2019-06-27
Removed the difference between static and normal query from page.

Aswinpookkatt:
Will add this to seperate page on difference between Static and Normal query .

### DIFF
--- a/docs/docs/static-query.md
+++ b/docs/docs/static-query.md
@@ -89,10 +89,3 @@ Header.propTypes = {
 }
 ```
 
-## How StaticQuery differs from page query
-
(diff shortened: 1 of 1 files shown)


marcysutton:
I don't think we should remove this–more explanation is better than none, and this page lives in `/docs/guides`. The other one being edited is under "Gatsby Internals", and really this information should be in both places. I think it's fine to be redundant.

marcysutton:
I'm going to close this one as we should keep the information in docs/guides – repetition will only help for something that confuses a lot of people. Thanks for the contributions though, we're stoked to have you working with Gatsby!

PR #15186, opened 2019-06-27
 added How StaticQuery differs from page query

Aswinpookkatt:
Added the difference between Static query and Page query


marcysutton:
Thanks for the contribution!

Issue #15190, opened 2019-06-27
Gatsby@2.11.0 Build failure (JavaScript heap out of memory)


moonmeister:
## Description

Gatsby 2.11 breaks site build. I'm guessing it's the new functionality around building node_modules as well as the site (#14111). 

i was updating dependencies on [master](https://github.com/moonmeister/moonmeister.net/tree/master) when build started breaking. I've updated to 2.10.5 on an [update branch](https://github.com/moonmeister/moonmeister.net/tree/updatedeps) with no troubles but as soon as I move Gatsby to 2.11 the build breaks. 

### Steps to reproduce

1. Clone master or update branches from https://github.com/moonmeister/moonmeister.net
2. update gatsby dep to gatsby@2.11.0 
3. run build => see breakage.

### Expected result

build should be successful.

### Actual result
```
[BABEL] Note: The code generator has deoptimised the styling of /home/ajmoon/code/moonmeister/moonmeister.net/node_modules/netlify-cms-app/dist/netlify-cms-app.js as it
⠦ Building production JavaScript and CSS bundles

<--- Last few GCs --->

[19712:0x2dbca30]    45370 ms: Scavenge 1338.2 (1423.9) -> 1337.3 (1423.9) MB, 2.9 / 0.0 ms  (average mu = 0.163, current mu = 0.102) allocation failure 
[19712:0x2dbca30]    45374 ms: Scavenge 1338.4 (1423.9) -> 1337.5 (1423.9) MB, 2.8 / 0.0 ms  (average mu = 0.163, current mu = 0.102) allocation failure 
[19712:0x2dbca30]    45378 ms: Scavenge 1338.6 (1423.9) -> 1337.7 (1424.4) MB, 2.6 / 0.0 ms  (average mu = 0.163, current mu = 0.102) allocation failure 


<--- JS stacktrace --->

==== JS stack trace =========================================

    0: ExitFrame [pc: 0x34eb54adbe1d]
    1: StubFrame [pc: 0x34eb54a875c2]
Security context: 0x2caa7a21e6e9 <JSObject>
    2: /* anonymous */(aka /* anonymous */) [0x23804dd52ac9] [/home/ajmoon/code/moonmeister/moonmeister.net/node_modules/@babel/core/lib/transformation/file/merge-map.js:~155] [pc=0x34eb5534b963](this=0x01389b5022b1 <null>,m=0x1c2e6adbae29 <Object map = 0x11c6eb590b11>)
[...]

PR #15191, opened 2019-06-27
fix(gatsby-plugin-netlify-cms): exclude node_modules from cms

erquhart:
## Description

#14111 caused a regression for `gatsby-plugin-netlify-cms`, which reuses a modified version of Gatsby's final webpack config and requires that `node_modules` be excluded from all builds. This PR adds the exclusion directly to the plugin.

Related to #15190

### DIFF
--- a/packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
+++ b/packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
@@ -1,5 +1,5 @@
 import path from "path"
-import { get, mapValues, isPlainObject, trim } from "lodash"
+import { get, mapValues, isPlainObject, trim, castArray } from "lodash"
 import webpack from "webpack"
 import HtmlWebpackPlugin from "html-webpack-plugin"
 import HtmlWebpackExcludeAssetsPlugin from "html-webpack-exclude-assets-plugin"
@@ -7,23 +7,17 @@ import MiniCssExtractPlugin from "mini-css-extract-plugin"
 // TODO: swap back when https://github.com/geowarin/friendly-errors-webpack-plugin/pull/86 lands
 import FriendlyErrorsPlugin from "@pieh/friendly-errors-webpack-plugin"
 
-/**
- * Deep mapping function for plain objects and arrays. Allows any value,
- * including an object or array, to be transformed.
- */
+// Deep mapping function for plain objects and arrays. Allows any value,
+// including an object or array, to be transformed.
 function deepMap(obj, fn) {
-  /**
-   * If the transform function transforms the value, regardless of type,
-   * return the transformed value.
-   */
+  // If the transform function transforms the value, regardless of type, return
+  // the transformed value.
   const mapped = fn(obj)
   if (mapped !== obj) {
     return mapped
   }
 
-  /**
-   * Recursively deep map arrays and plain objects, otherwise return the value.
-   */
+  // Recursively deep map arrays and plain objects, otherwise return the value.
   if (Array.isArray(obj)) {
     return obj.map(value => deepMap(value, fn))
   }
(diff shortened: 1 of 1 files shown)


moonmeister:
@erquhart Can you give some context, maybe the original issue or a explanation of the regression? I just ran into some issues while updating my site but was able to get it to build and deploy. It appears to be working. see #15190 for my troubles.

P.S. I'm mostly just curious and want to understand...not saying you're incorrect.

erquhart:
@wardpeet thanks for the edits! I think I might see a couple of issues, let me give a quick look before you merge, I'll be at my machine soon.

packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
```diff
@@ -33,6 +27,38 @@ function deepMap(obj, fn) {
   return obj
 }
 
+function replaceRule(value) {
+  // If `value` does not have a `test` property, it isn't a rule object.
+  if (!value || !value.test) {
```
erquhart: `value` could be a string - we're recursing the entire received object, in which case `!value.test` would throw.

packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
```diff
@@ -33,6 +27,38 @@ function deepMap(obj, fn) {
...
+    return value
+  }
+
+  // when javascript we exclude node_modules
+  if (value.type === `javascript/auto` && value.exclude) {
+    return {
+      ...value,
+      exclude: new RegExp(
+        [value.exclude.source, `node_modules|bower_components`].join(`|`)
```
erquhart: We don't know for sure that the `exclude` is and will continue to be a RegExp, it could easily be an array. `castArray` is the safe route, its guaranteed not to fail, we always know what we're getting, and the output is equally compatible with Webpack.

erquhart:
@moonmeister ah, your issue is exactly what this addresses, I'll comment there.

erquhart:
@wardpeet sorry to bug! Any chance of getting this in before the weekend? Bit of a showstopper for affected users.

Jamamuuga:
Will any additional changes be needed for a developer who uses the plugin & uses Yarn's PnP feature?

erquhart:
Update: comments on #15190 show that this isn't limited to the Netlify CMS plug-in. The appropriate fix to that issue may supercede this PR.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
wardpeet: if test is a string it will give a falsy value (undefined) when `value.test` is used. So no harm here, thanks for bringing it up to my attention.

packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
wardpeet: I'm checking that it's a regex now. If people have a custom js rule they should make sure to exclude node_modules themselves.

packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
erquhart: No, `value` could be a string. This will throw for some users.

erquhart:
@wardpeet you misread my comment: `value` may be a string, in which case `value.test` will throw. I also commented that a proper fix to #15190 would likely supercede this PR, which appears to have happened in #15270. I recommend reverting this PR.

packages/gatsby-plugin-netlify-cms/src/gatsby-node.js
DSchau: @erquhart do you have an example?

```js
var a = `1234`

if (!a || !a.test) {
  // we will be in this block
}
```

erquhart: Hmm not sure why I was thinking that would throw as if `value` is undefined. I know I ran into issues deep mapping the webpack config and needed safe getters, but you're right, this wouldn't be an issue at this line.

erquhart:
Took another look based on @DSchau's comment on the review, and you're right, `value` wouldn't throw. I do still think this should be reverted since the bug it addressed has itself been reverted.

rdela:
@wardpeet @erquhart In case it helps figure any of this out, this repo that uses Netlify CMS is still broken under 2.13.6, but everything is fine under 2.10.5. See [master branch](https://github.com/rdela/voidcluster) vs https://github.com/rdela/voidcluster/pull/54; build failure log at https://app.netlify.com/sites/voidcluster/deploys/5d20f59e9f191b000a971e08

Issue #5995, opened 2018-06-18
Feature Request: Manual component code splitting


alexlande:
## Summary

I'm working on an app that would benefit from component-based code splitting that works with static rendering. I tweeted about it here: https://twitter.com/gatsbyjs/status/1007760243596513280. Basically, I'd like to be able to code split components using `import()` while rendering them in statically built pages.

The result would be that the HTML for a page contains the lazy loaded component's markup and the app isn't hydrated on the client until components used on the page are loaded. After the initial page load, other pages would lazy load these components as you navigate through the app.

Code splitting already works, so preloading before rendering and statically rendering code split components would be the new behavior here.

### Tools

[`react-static`](https://github.com/nozzle/react-static) handles this use case with [`react-universal-component`](https://github.com/faceyspacey/react-universal-component). Usage example: https://github.com/nozzle/react-static/blob/master/examples/dynamic-imports/src/containers/About.js

I typically use [`react-loadable`](https://github.com/jamiebuilds/react-loadable) for component code splitting. That project has pretty comprehensive instructions on server rendering, but there are a lot of moving parts-- a babel plugin, a webpack plugin, a component to capture a list of rendered loadable components, and preload methods to be used on the client and the server to ensure that the initial render + hydration works.

### Motivation

The project I'm working on is a large site where page customization is driven by components rather than templates. There's only a single template, but authors using the connected CMS can mix and match a large number (100+) of content components that fit into slots in the template. This means that Gatsby's existing template code splitting doesn't help us much, and if we were to load all of the components in the template the resulting JS bundle would be huge.

KyleAMathews:
This is definitely something we want to support. Question is how :-) This space seems like a moving target so not entirely sure what we want to do. Also question is can we support react-loadable/react-universal-component in plugins as that'd be ideal as we'd decouple core from supporting these different options.

alexlande:
Yep, agreed that a plugin would be best. I ran into two issues looking at this that make me unsure if it needs deeper integration than a plugin can do:

1. `react-loadable` has an async [`preloadAll`](https://github.com/jamiebuilds/react-loadable#loadablepreloadall) method used to preload components in Node, that needs to be called before rendering. I think this would require async `replaceRenderer` support along the lines of https://github.com/gatsbyjs/gatsby/pull/2423 if you were going to do it through a plugin.
[...]

PR #15196, opened 2019-06-28
Moved api-node-docs initialization so that it can be overridden.

jrestall:
This is a small change to move the api list initialization out of global scope so that my gatsby theme has a chance to override the allowed Gatsby APIs before it is initialized.

<details>
  <summary>Why would you want to do that?!</summary>

1. I'm working on integrating OrchardCore CMS with GatsbyJS by creating a number of themes and a plugin.
2. The themes are able to compose an entire site based on the structured data in the CMS but need to have a say in how the graphql page query is built, how the page query response is processed and contribute context to the created page nodes.
    * gatsby-theme-orchardcore-images needs to process the page query and look for any widgets containing images, then replace the url with a gatsby-image supported fluid object.
    * gatsby-theme-orchardcore-layers needs to contribute widget names from any of the dynamic CMS defined layers using a custom 'onCreateTemplate' API, so that the dynamically created template file will include all the correct imports and code splitting will include only the subset of widgets used on that one page. https://github.com/gatsbyjs/gatsby/issues/5995#issuecomment-502891938  
    * gatsby-theme-orchardcore-flows needs to search for widgets and add them as fragments to the page query using the 'sourcePageQuery' API.
    * Etc...  

3. I want to reuse the flexibility of the gatsby nodeRunner to support these custom APIs and allow my themes the same level of flexibility for building OrchardCore pages as the standard node API gives us when building sites.
4. Because gatsby's plugin-loader and nodeRunner by default prohibits plugins exporting and calling unknown Gatsby-APIs I need to be able to override the allowed API list in api-node-docs when loading my main theme. This is done as below but requires the PR change to work:
```
// gatsby-config.js

const apiList = require(`gatsby/dist/utils/api-node-docs`)

module.exports = _ => {

  // Tell gatsby to allow these non-standard API calls
  apiList.sourcePageQuery = true
  apiList.onCreatingTemplate = true
  apiList.onCreatingPage = true

  return ({...})
}
```
</details>

If interested in OrchardCore integration or how I'm solving code splitting, very early code is here: https://github.com/jrestall/gatsby-orchardcore

### DIFF
--- a/packages/gatsby/src/bootstrap/load-plugins/index.js
+++ b/packages/gatsby/src/bootstrap/load-plugins/index.js
@@ -11,12 +11,6 @@ const {
   handleMultipleReplaceRenderers,
 } = require(`./validate`)
 
-const apis = {
-  node: _.keys(nodeAPIs),
-  browser: _.keys(browserAPIs),
-  ssr: _.keys(ssrAPIs),
-}
-
 // Create a "flattened" array of plugins with all subplugins
 // brought to the top-level. This simplifies running gatsby-* files
 // for subplugins.
@@ -38,6 +32,12 @@ const flattenPlugins = plugins => {
 }
 
 module.exports = async (config = {}, rootDir = null) => {
+  const apis = {
+    node: _.keys(nodeAPIs),
+    browser: _.keys(browserAPIs),
+    ssr: _.keys(ssrAPIs),
+  }
+
   // Collate internal plugins, site config plugins, site default plugins
   const plugins = loadPlugins(config, rootDir)
 


freiksenet:
Hello @jrestall!

I'm afraid we won't be merging this PR as is. We don't think this would be a good thing to allow additional custom hooks added this way. In addition, calling `apiRunnerNode` directly from other api runners isn't a supported thing.

It seems that you are calling those APIs directly from your `createPages`, so technically you don't need an API runner to do that. You can import relevant processing functions directly from the plugins and call them inside your `createPages`.

Regardless, we would be willing to add new node APIs if we see that there are some things that are impossible with our current APIs and that the new API will enable a lot of use cases currently not enabled by Gatsby.

Issue #13134, opened 2019-04-04
Documentation missing installation step


idahogurl:
The documentation here https://www.gatsbyjs.org/docs/gatsby-config/#mapping-node-types fails to mention you need to install the Yaml plugin.

pieh:
Hey @idahogurl! Thanks for opening issue. 

I'm not sure what's the best approach here, because that part of documentation is using Markdown, Yaml and Json as examples, but mapping is not limited to those and those are just used to illustrate mapping feature.

To me, adding installation instruction for those plugins would divert attention from the essence of content there.

idahogurl:
Perhaps a short comment that they require the plugin to be installed.

wgao19:
Hi @pieh @idahogurl  I encountered this issue too when I tried to add mapping to my site. I also learned that the file needs to be in the directory picked up by the `gatsby-source-filesystem` as well. Since I'm pretty new to Gatsby this was a #TIL for me. 

What do you think of the following:

```diff
You can map between the author field in frontmatter to the id in the author.yaml objects by adding to your gatsby-config.js:

module.exports = {
  plugins: [...],
  mapping: {
    "MarkdownRemark.frontmatter.author": `AuthorYaml`,
  },
}

+ You may need to install the appropriate file transformer and source plugins for Gatsby to pick up the mapping files. This applies to other file types later mentioned in this segment as well.

Gatsby then uses this mapping when creating the GraphQL schema to enable you to query data from both sources:
```

polishedwp:
Hi @pieh this has happened again. See #13660. 
I helped the user with a PR to their repo and worked them through the steps for the mapping to work, but I think we at least add a note to remind the read to add the plugin with a link to the [gatsby-transformer-yaml](https://www.gatsbyjs.org/packages/gatsby-transformer-yaml/?=yaml). I think it would help new users.

k-conway:
I'm happy to open a PR and add the comment from @wgao19 if folks have decided that will help.

wgao19:
Oh just seeing the added label. I am also happy PR. It’ll be my first contribution to this repo 😁 Although, I don’t have too much more insight beyond that line so if you got more to add on I think it’ll be nicer for you to do it @k-conway

k-conway:
I was pretty much going to update with your comment so feel free to take it since you already did the hard part :-)

wgao19:
Thank you @k-conway! Just opened #15197, let's hear what people will say about it :)

[...]

PR #15197, opened 2019-06-28
docs: explain missing file loading step for schema mapping

wgao19:
## Description

Explain the file transformer dependency and plugin setup for schema mapping.

## Related Issues

Addresses #13134


docs/docs/gatsby-config.md
```diff
@@ -115,6 +115,8 @@ module.exports = {
 }
 ```
 
+You may need to install the appropriate file transformer and source plugins for Gatsby to pick up the mapping files. This applies to other file types later mentioned in this segment as well.
```
marcysutton:
```suggestion
You may need to install the appropriate file transformer and source plugins for Gatsby to pick up the mapping files. This applies to other file types later mentioned in this segment as well. For example, to handle YAML files, you would install [`gatsby-transformer-yaml`](/packages/gatsby-transformer-yaml/?=yaml). 

> _Note: Additional source plugins can be found in the [plugin library](/plugins?=gatsby-source)._
```

What do you think about adding some links to common source plugins?

wgao19: Hey @marcysutton, I added a link to the YAML transformer plugin and rephrased a little bit. How do you think it reads now?

docs/docs/gatsby-config.md
```diff
@@ -115,6 +115,8 @@ module.exports = {
 }
 ```
 
+You may need to install the appropriate file transformer (in this case [YAML](https://www.gatsbyjs.org/packages/gatsby-transformer-yaml/)) and set up [gatsby-source-filesystem](https://www.gatsbyjs.org/packages/gatsby-source-filesystem/) properly for Gatsby to pick up the mapping files. This applies to other file types later mentioned in this segment as well.
```
marcysutton:
```suggestion
You may need to install the appropriate file transformer (in this case [YAML](/packages/gatsby-transformer-yaml/)) and set up [gatsby-source-filesystem](/packages/gatsby-source-filesystem/) properly for Gatsby to pick up the mapping files. This applies to other file types later mentioned in this segment as well.
```
We should keep these URLs relative to the docs site. But that's it! Looking great!

wgao19: @marcysutton updated :)))))

PR #15198, opened 2019-06-28
added few info on non-page components

### LINKED ISSUE #15195: [docs][guides] update Building With Components page with non-page components
## Summary

In reviewing #15079, I wanted to link to the [Building with Components](https://www.gatsbyjs.org/docs/building-with-components/) Conceptual Guide but noticed it doesn't cover everything it should. There are sections on [Page Components](https://www.gatsbyjs.org/docs/building-with-components/#page-components) and [Page Template Components](https://www.gatsbyjs.org/docs/building-with-components/#page-template-components), and then it goes into [`html.jsx`](https://www.gatsbyjs.org/docs/building-with-components/#html-component), and doesn't mention anything else. 

It would really help users of StaticQuery and PageQuery to know what "non-page components" means in detail, and this page would be a great place to expand on that. Ideally, we could craft a new bit of text that rephrases the definition in a new way rather than copying so if the other explanations don't make sense to someone, they'll get another chance to understand. This is a Conceptual Guide, so deeper technical explanations are welcome.

The `html.jsx` section should probably also explicitly mention that overriding `html.js` is optional. It mentions it in the third paragraph with "Typically you should omit this from your site", but it should be more clearly stated in the first sentence or two that you don't need to create that file unless you've got a specific reason.

The page to update: https://www.gatsbyjs.org/docs/building-with-components/

With these related pages and an updated Conceptual Guide, we'll have more explanations throughout the docs for extra clarity on component types:
- [More on Static Query for querying data in components](https://www.gatsbyjs.org/docs/static-query/)
- [Gatsby Internals: Static vs. Normal Queries](https://www.gatsbyjs.org/docs/static-vs-normal-queries/) (updated in #15186)

Aswinpookkatt:
## Description
Querying data using static query docs and example added

## Related Issues
 fixes #15195


docs/docs/building-with-components.md
```diff
@@ -167,3 +167,35 @@ These are examples of the different ways React components are used in Gatsby
 sites. To see full working examples, check out the
 [examples directory](https://github.com/gatsbyjs/gatsby/tree/master/examples) in
 the Gatsby repo.
+
+### Querying data in components using StaticQuery
+Gatsby v2 introduces StaticQuery, a new API that allows components to retrieve data via GraphQL query.
+<br>
```
marcysutton:
```suggestion

```
These break tags aren't necessary in Markdown.

docs/docs/building-with-components.md
```diff
@@ -167,3 +167,35 @@ These are examples of the different ways React components are used in Gatsby
 sites. To see full working examples, check out the
 [examples directory](https://github.com/gatsbyjs/gatsby/tree/master/examples) in
 the Gatsby repo.
+
+### Querying data in components using StaticQuery
```
marcysutton: I think this should be more about what "non-page components" are than how to use StaticQuery. The reason to write about it is to support StaticQuery docs, but this one should be focused more on the type of component that StaticQuery can be used in.

docs/docs/building-with-components.md
```diff
@@ -167,3 +167,8 @@ These are examples of the different ways React components are used in Gatsby
 sites. To see full working examples, check out the
 [examples directory](https://github.com/gatsbyjs/gatsby/tree/master/examples) in
 the Gatsby repo.
+
+### Non-page components
+A Non-page component is embedded inside some other component forming a hierarchy that can get quite deep
```
marcysutton:
```suggestion
A Non-page component is one that's embedded inside some other component, forming a component hierarchy. An example would be a Header component that's included in multiple page components.
```

docs/docs/building-with-components.md
```diff
@@ -167,3 +167,8 @@ These are examples of the different ways React components are used in Gatsby
...
+Gatsby uses GraphQL to enable components to declare the data they need.Using StaticQuery, you can colocate a component with its data.
```
marcysutton:
```suggestion
Gatsby uses GraphQL to enable components to declare the data they need. Using the [StaticQuery](/docs/static-query/) component or [useStaticQuery hook](/docs/use-static-query/), you can colocate a non-page component with its data.
```

PR #15215, opened 2019-06-28
feat(themes): Notes theme header

jxnblk:
# Description

This is an alternative to #14710 that adds the blog theme's header to the notes theme's pages through shadowing. I'd like to wait until #15090 is merged in before this one, but opening for earlier review.

- The blog's header is pulled out into its own module for shadowing
- The `gatsby-starter-theme` example shadows the `header` component from `gatsby-theme-blog`, adding a navigation link to the `/notes` route and passing a `title` prop in
- The example also shadows `layout` from `gatsby-theme-notes`
- This shadowed layout imports the blog's `header` component to include it in the notes `layout`
- I've also removed the theme-ui layout components here because I think they were unnecessary and potentially more confusing than helpful
- Updated the `theme-ui` dependencies

## Related Issues

Related to #14276


themes/gatsby-theme-blog/src/components/bio.js
```diff
@@ -21,7 +21,7 @@ const Bio = () => {
   } = data
 
   return (
-    <Flex mb={4}>
+    <Flex css={css({ mb: 4 })}>
```
jxnblk: The latest thinking here is to move away from using styled-system like style props in favor of the `css` (and `sx` prop when using the custom JSX pragma) – this helps reduce the overall bundle size of theme-ui

themes/gatsby-theme-blog/src/gatsby-theme-ui/typography.js
```diff
@@ -1,6 +1,6 @@
 import "typeface-montserrat"
 import "typeface-merriweather"
-import { toTheme } from "theme-ui-typography"
+import { toTheme } from "@theme-ui/typography"
```
jxnblk: This package was renamed to use the scope like other theme-ui packages

themes/gatsby-theme-notes/src/components/layout.js
```diff
@@ -12,8 +12,6 @@ export default props => (
         },
         body: {
           margin: 0,
-          color: `text`,
-          bg: `background`,
```
jxnblk: This was already handled by `gatsby-theme-ui` so it should be needed here

themes/gatsby-theme-notes/src/components/note.js
```diff
@@ -7,8 +7,9 @@ const WikiPage = ({
   data: {
     note: { body },
   },
+  ...props
 }) => (
-  <Layout>
+  <Layout {...props}>
```
jxnblk: The blog header uses `props.location`, so there's a little bit of props drilling added

themes/gatsby-theme-notes/src/components/breadcrumbs.js
```diff
@@ -7,7 +7,7 @@ import BreadcrumbDivider from "./breadcrumb-divider"
 import BreadcrumbHome from "./breadcrumb-home"
 
 export default ({ links }) => {
-  const { notesPath, homeText, breadcrumbSeparator } = useOptions()
+  const { basePath = `/`, homeText, breadcrumbSeparator } = useOptions()
```
amberleyromo: good catch

themes/gatsby-starter-theme/src/gatsby-theme-blog/components/header.js
```diff
@@ -0,0 +1,26 @@
+import React from "react"
+import { Link } from "gatsby"
+import { css, Styled } from "theme-ui"
+import Header from "gatsby-theme-blog/src/components/header"
+
+export default props => (
+  <Header {...props} title="Custom Title">
```
jxnblk: The `title` prop here changes what the starter shows – should I change it back to say `Shadowed Site Title`, @amberleyromo?

jxnblk: Okay, dug into this again, and I have a question: the notes layout does not render the `title` from `siteMetadata` because that query is currently in the blog theme's template. How do we want to handle component-level queries like this?

amberleyromo: can the notes theme make its own query? it's querying the same config, so it would still have `title` in `siteMetadata` on the site level.

ChristopherBiscardi: Why not put the `StaticQuery` in the header itself?

jxnblk: If the query is in the header component, then it'd be on the user's end to also include that in the shadowed component – that seems a little less-than-ideal. If we make all official themes follow the same siteMetadata setup, that seems like it'd be a case of it works here, but not with other themes

amberleyromo: I think that makes sense though. These are built to work together -- if you're composing another theme in that wasn't built for that, it makes sense that you might need to do more manual work to tune it. But these were designed to work together.

ChristopherBiscardi: > If the query is in the header component, then it'd be on the user's end to also include that in the shadowed component – that seems a little less-than-ideal.

That seems pretty ideal to me. If I don't want the data in my header then the entire query goes away, and this also means it's not passed through other arbitrary components that don't care about the data like `Layout`.

If you're shadowing a component you can copy the whole file over, which gives you the data and the component in a single location rather than having to learn how page context works.

PR #15227, opened 2019-06-28
Show structured errors for graphql errors from the graphql function used in gatsby-node.js

### LINKED ISSUE #15212: Catch graphql errors from the graphql function and pretty-print errors
## Description

If in your `gatsby-node.js` you send a bad graphql query to the graphql function (this is commonly done to query data to create pages), you won't get any error. Instead you'll generally get an error that you can't access expected data from the graphql result e.g.

Screenshot 2019-06-28 07 28 22

Ideally we'd catch errors return from the graphql function and create structured errors for common problems.

### Steps to reproduce

Create a site add add the following code to your `gatsby-node.js`:

```js
  return graphql(
    `
      query HomePageQuery {
        site {
          # Typo here
          siteMetadat {
            description
          }
        }
      }
    `
  ).then(result => {
    const doesthisexist = result.data.siteMetadat
  })
}

```

If you run `gatsby build`, it'll give an error like the following:

Screenshot 2019-06-28 07 35 27

It should instead show that there was an error with your graphql query including what file/line number it happened in and the GraphQL error message.

GraphQL

0

KyleAMathews:
Fixes #15212

Error before:
Screenshot 2019-06-28 15 20 05

Error after:
Screenshot 2019-06-28 15 20 36

### DIFF
--- a/packages/gatsby/src/bootstrap/index.js
+++ b/packages/gatsby/src/bootstrap/index.js
@@ -396,22 +396,29 @@ module.exports = async (args: BootstrapArgs) => {
     ).then(result => {
       if (result.errors) {
         report.panicOnBuild(
-          result.errors.map(e => {
-            // Find the file where graphql was called.
-            const file = stackTrace
-              .parse(e)
-              .find(file => /createPages/.test(file.functionName))
-            if (file) {
-              return errorParser({
-                message: e.message,
-                location: {
-                  start: { line: file.lineNumber, column: file.columnNumber },
-                },
-                filePath: file.fileName,
-              })
-            }
-            return null
-          })
+          result.errors
+            .map(e => {
+              // Find the file where graphql was called.
+              const file = stackTrace
+                .parse(e)
+                .find(file => /createPages/.test(file.functionName))
+              if (file) {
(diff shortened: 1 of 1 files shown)


packages/gatsby/src/bootstrap/index.js
```diff
@@ -385,13 +387,37 @@ module.exports = async (args: BootstrapArgs) => {
 
   const graphqlRunner = (query, context = {}) => {
     const schema = store.getState().schema
+    console.log(`super`)
```
m-allanson:
```suggestion
```

KyleAMathews:
Thinking about this a bit more, I think it'd make sense to use a different error code for graphql errors from the graphql function. It'll be helpful from a product standpoint to differentiate the two and also it might be useful to have different error messages.

KyleAMathews:
Also it should be possible to print out the query. Would need to go looking for the graphql function for printing the query with the error highlighted.

KyleAMathews:
Decided to keep the same id & just note in the context that the error came from a graphql function call.

We should print out the query — let's work on that for this and other graphql errors in a future PR.

PR #15243, opened 2019-06-29
gatsby-theme-blog: Mdx -> BlogPost

ChristopherBiscardi:
## Description

Replace the queries for `Mdx` with queries for `BlogPost`. This provides a better potential for future compatibility with generic data interface work.

### DIFF
--- a/themes/gatsby-theme-blog/gatsby-node.js
+++ b/themes/gatsby-theme-blog/gatsby-node.js
@@ -1,6 +1,7 @@
 const fs = require(`fs`)
 const path = require(`path`)
 const mkdirp = require(`mkdirp`)
+const crypto = require(`crypto`)
 const Debug = require(`debug`)
 
 // @TODO document theme options:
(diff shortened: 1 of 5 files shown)


themes/gatsby-theme-blog/src/templates/post.js
```diff
@@ -9,12 +9,10 @@ export default ({ pathContext: { previous, next }, location, data }) => (
 
 export const pageQuery = graphql`
   query($id: String!) {
-    post: mdx(id: { eq: $id }) {
+    post: blogPost(id: { eq: $id }) {
```
jxnblk: This is cool

themes/gatsby-theme-blog/gatsby-node.js
```diff
@@ -39,6 +40,55 @@ exports.onPreBootstrap = ({ store }, themeOptions) => {
...
+    schema.buildObjectType({
+      name: `BlogPost`,
+      fields: {
+        id: { type: `ID!` },
+        title: {
+          type: `String!`,
+        },
+        date: { type: `Date`, extensions: { dateformat: {} } },
+        tags: { type: `[String]!` },
+        keywords: { type: `[String]!` },
```
jxnblk: Maybe a dumb question but what's the difference between *tags* and *keywords*?

ChristopherBiscardi: keywords seems to be for SEO purposes while I'd expect tags to be user-facing. I'm open to removing this

ChristopherBiscardi:
missing `slug: { type: 'String' }` in the `sourceNodes` type for `BlogPost`

themes/gatsby-theme-blog/gatsby-node.js
amberleyromo: ^yep. keywords are tag-oriented, for tags we could do tag filtering later.

PR #15251, opened 2019-06-30
feat(docs): Table of Contents component

### LINKED ISSUE #6896: Floating Table of Contents for longer docs (ToC)

lannonbr:
## Description

Added a Table of Contents (TOC) component to pages in the docs that separate parts of the page using headings. This is all based upon discussions in #6896 and pair programming sessions I had with @marcysutton and @shannonbux.

## TODO

These things need to be resolved before getting merged:

- [ ] Make sure this is accessible
- [x] In the template for the docs, I am adding some styling to some various components to expand the container so on most pages on a large display, the TOC won't overlap as it scrolls down the page. I think this can likely be handled better.
- [x] Clean up some of the design with advice from @fk. I tried using some of the design tokens now set up in `src/utils/presets.js`

As well, as I have been going through some pages to test this out, some things have appeared to me. Some of these can likely be tackled as separate PRs.

- [x] some pages have a TOC already that is static. We could remove the static TOC.
- [x] We may want to disable the TOC on some pages. For instance, on the [Glossary](https://www.gatsbyjs.org/docs/glossary/) page we may want to use the alphabet component as a TOC instead of this TOC. We could maybe have a custom field in the frontmatter to disable this TOC.

## Screenshots

Here's a few screenshots of the current state:

On large screens, there now will be a floating component on the right that displays the Table of Contents. This uses `position: sticky` so it will stick to the viewport as the page is scrolled down.

GraphQL Reference page with TOC on right for large display

On mobile and small screens, it will be in the flow of the regular document:

Introducing GraphQL page with TOC inline on mid-sized display
Introducing GraphQL page with TOC inline on phone-sized display

## Related Issues

Fixes #6896


www/src/components/docs-table-of-contents.js
```diff
@@ -0,0 +1,61 @@
...
+          padding: space[6],
+          boxShadow: shadows.overlay,
+          float: `right`,
+          marginLeft: space[10],
+          position: `sticky`,
+          top: 130,
+        },
+      }}
+    >
+      <h1
```
gillkyle: I think these should be `h2` tags since we already have an `h1` on the page for the title. Then it will also match the subheaders that it falls inline with on the smaller viewport size.

www/src/components/docs-table-of-contents.js
```diff
@@ -0,0 +1,61 @@
...
+function TableOfContents({ page, location }) {
+  return page.tableOfContents.items ? (
+    <div
+      css={{
+        background: colors.white,
+        zIndex: zIndices.feedbackWidget,
+
+        [mediaQueries.xxl]: {
+          padding: space[6],
+          boxShadow: shadows.overlay,
```
gillkyle: IMO the shadow isn't totally necessary, I think it looks good sitting flat and it feels a little more visually consistent (with the rest of the page and with the mobile view as well) to me without it:

image

```suggestion
```

www/src/components/docs-table-of-contents.js
```diff
@@ -0,0 +1,61 @@
...
+    ))
+  )
+}
+
+function TableOfContents({ page, location }) {
+  return page.tableOfContents.items ? (
+    <div
+      css={{
+        background: colors.white,
+        zIndex: zIndices.feedbackWidget,
```
gillkyle: I think this might not be needed, if the shadow is taken off and it sits flat on the page, it'd simplify against situations like this:

image

where the feedback widget and it overlap.

```suggestion
```

www/src/components/docs-table-of-contents.js
fk: 👍

www/src/components/docs-table-of-contents.js
fk: :ok

fk:
@lannonbr I want to second everything @gillkyle said — this is awesome!

www/src/components/docs-table-of-contents.js
lannonbr: Done 😄

www/src/components/docs-table-of-contents.js
```diff
@@ -0,0 +1,52 @@
...
+      <h2
+        css={{
+          textTransform: `uppercase`,
+          fontSize: fontSizes[2],
+          color: colors.grey[60],
+        }}
+      >
+        Table of Contents
+      </h2>
+      <nav>
```
marcysutton: Can the h2 go in the nav here to act as a label for it?

shannonbux:
Hello! Fun to see this! I don't know if someone has mentioned this yet. I
also think the design could use colors that are a little softer (e.g. the
text could be medium grey and the bullet points perhaps medium grey too).
Because we want people to be able to reference the ToC, yet not have it
compete visually with the main document. Perhaps removing the drop shadow
would also make the area less emphasized. Exciting to see this progressing!

marcysutton:
>  the design could use colors that are a little softer 

As long as the colors meet [contrast requirements for regular text](https://dequeuniversity.com/tips/color-contrast), sounds like a good design decision to me!

fk:
☝️👍 

Contrast scores also are available in the modal for each color group on https://www.gatsbyjs.org/guidelines/color/ btw.

www/src/components/docs-table-of-contents.js
```diff
@@ -0,0 +1,52 @@
...
+          marginLeft: space[10],
+          position: `sticky`,
+          top: 130,
+        },
+      }}
+    >
+      <h2
+        css={{
+          textTransform: `uppercase`,
+          fontSize: fontSizes[2],
```
fk: I'd go for `fontSizes[1]` here, and throw in a `letterSpacing: letterSpacings.tracked`—ref. the excellent https://practicaltypography.com/letterspacing.html for why we want extra letterspacing for all caps text.

www/src/templates/template-docs-markdown.js
```diff
@@ -49,11 +51,26 @@ function DocsTemplate({ data, location }) {
         enableScrollSync={urlSegment === `docs` ? false : true}
       >
         <DocSearchContent>
-          <Container>
+          <Container
```
fk: I just gave this a spin locally (insomnia), and noticed that the main content currently sticks to the left of the…"main sidebar nav" (the one to the left, containing the navigation for "Docs/Tutorial/Features")…on large screens where the TOC is positioned to the right of the main content:

image

IMO we want the whole content, including our new TOC sidebar, to stay centered. Not super trivial, but also not too hard to do—so I got tempted, and gave that a quick shot. Pushed my stuff here:
https://github.com/gatsbyjs/gatsby/commit/8616bc169350fe0af1895b545abaa48d143d8039

fk: Here's how what is in that PoC branch looks like:

toc-poc-low

It's not perfect, but maybe it helps a little.

fk: - To make it a bit easier to get there (and understand/maintain this stuff in the future), this drops pulling `.gatsby-highlight` & Co. to the left/right for `mediaQueries.lg`—no need to worry about this detail IMO. It was fun while it lasted, time to move on. (happened in `typography.js`)
- Also I thought it's time to drop defining `<Container>`s `maxWidth` as `rhythm`, and added those values to `tokens/sizes` as `rem`s.
- That commit might also include what I mentioned in https://github.com/gatsbyjs/gatsby/pull/15251/files#r302286146 `¯\_(ツ)_/¯` :wink:

Infamous last words: We eventually have to take care of the TOC sidebar _height_, too.

lannonbr: Yeah thank you @fk for this. The container being pushed around felt very hacky when I built it but this is looking to be much better! I will look into this deeper later

fk: Thank _you_, Ben—your work made it really easy for me to jump in to do this! 🙌 

Sorry for the sloppy commit there—it looks a bit much, but IMO contains a lot of valuable changes. Please don't hesitate to reach out if something is unclear, or just pick what you think is valuable. One thing I want to trying clarifying is what I meant by

> started off with a different approach, so it's not fully fluid yet, container size is "hardcoded" a little more than it needs still, so if the ToC content doesn't need the full width of what the layout container currently reserves for it, things look a little weird (uncentered)

—I think we might not need to define a `maxWidth` for the outer container at all:

https://github.com/gatsbyjs/gatsby/commit/8616bc169350fe0af1895b545abaa48d143d8039#diff-731b5641a3566389a9aa7e6d232fe6adR26

fk: ^ Ah no, ignore that last bit about `maxWidth`—I think that PoC is as good as it gets with the `<h1>` and the requirements defined in https://github.com/gatsbyjs/gatsby/issues/6896#issuecomment-499972360 (ToC following the main headline on smaller screens), at least if we don't want to throw in some JS yada yada, or go for left-aligned content once the ToC on the right kicks in (which isn't too bad, thinking about that again), or revisit the requirements.

Better to worry about content becoming inaccessible if the ToC height exceeds the viewport height IMO. Wondering if we want to avoid two scrollbars relatively close to each other here? An approach as https://react-sticky-box.codecks.io/ looks interesting irt to that.

fk: Pushed a quick rudimentary fix for the latter to https://github.com/gatsbyjs/gatsby/tree/topics/15251-layout (https://github.com/gatsbyjs/gatsby/commit/5951933a556456e0beafbd77776ebac4fd5a27d0).

lannonbr:
Just a notice I will likely continue work on this PR on Thursday. Been busy with other things this week but the end of the week is much lighter

www/src/components/docs-table-of-contents.js
lannonbr: Done

www/src/components/docs-table-of-contents.js
lannonbr: Done

lannonbr:
Alright, so I continued some work with this. I merged in the changes @fk made in a side branch so the container stuff is a bit more succinct and the design fits better.

With what @marcysutton commented for the disabling of the table of contents, I added a `disableTableOfContents` field in the frontmatter that when set to be true will not render this TOC and just use the one that may exist already on the page. I added this to the glossary page as an example but we can do a scan of other pages at a later point of if we want to disable it on other pages.

There seems to be some issues with the www unit tests. I will try debugging those later this weekend.

www/src/templates/template-docs-markdown.js
lannonbr: Done. This has been merged into my branch.

marcysutton:
I just saw that same error locally that's coming up on Circle `unit_tests_www` actually: `Error: Uncaught [TypeError: Cannot read property 'withSidebar' of undefined]`

fk:
@lannonbr @marcysutton Tests are failing because when merging `master` in https://github.com/gatsbyjs/gatsby/commit/fe93df40ba503e7916a28e421d0b63c341d6180a, git got a little too smart and moved `ww/src/utils/tokens/sizes.js` to `packages/gatsby-design-tokens/src/sizes.js`, and the lines causing problems (https://github.com/gatsbyjs/gatsby/blob/fe93df40ba503e7916a28e421d0b63c341d6180a/packages/gatsby-design-tokens/src/sizes.js#L11-L14) with it. That's what we (I 😉) get for moving the design tokens into their own package, and including stuff in there which shouldn't have been there in the first place, namely the `sizes` tokens that we added `mainContentWidth` to.

I'll push a fix. ✌️

fk:
- tests fixed in 792dccb
- changes to `packages/gatsby-design-tokens` reverted in d623592
- cherry-picked https://github.com/gatsbyjs/gatsby/commit/5951933a556456e0beafbd77776ebac4fd5a27d0 in aaf8aae

lannonbr:
Thanks so much for the fixes @fk!

fk:
@lannonbr You're welcome! 🤗

lannonbr:
This should be ready for another review

gillkyle:
Just merged master to resolve problems with the prev and next buttons that got merged recently. Also removed the Table of Contents that were on pages that I could find doing a quick couple of regex searches across the repo and disabled the ToC on the tutorial where the ToC is a little redundant since that the majority of that info is already displayed in the sidebar.

gillkyle:
Thanks for all the awesome work on this everyone! @lannonbr, @shannonbux, @marcysutton, @fk 🎉

fk:
🎉 …and thank you Kyle!

muescha:
I miss some documentation of the TOC component and the usage `disableTableOfContents`

PR #15268, opened 2019-07-01
fix(gatsby): various Typescript definitions

d4rekanguok:
## Description

Hi folks, sorry about these droplet of tiny fixes, I wasn't actively looking for issues & only fix them as I go. If you keep this one open for a few days, I'll push here whenever I find more issues.

- [gatsby] Fix `touchNode` def so `plugin` is the 2nd argument instead of being nested in the first one ([reference](https://github.com/gatsbyjs/gatsby/blob/7c6a623f6d468b81c8771bbaddc003a6430bc3fd/packages/gatsby/src/redux/actions/public.js#L749-L770))

- [gatsby] Fix `createRedirect` def: make `fromPath` & `toPath` the only 2 required fields ([reference](https://github.com/gatsbyjs/gatsby/blob/7c6a623f6d468b81c8771bbaddc003a6430bc3fd/packages/gatsby/src/redux/actions/public.js#L1141-L1163))

- [gatsby-source-filesystem] Fix Cache definition in createRemoteFileNode args (related to [this one](https://github.com/gatsbyjs/gatsby/pull/14955))

- [gatsby-source-filesystem] add definition for createFileNodeFromBuffer ([reference](https://github.com/gatsbyjs/gatsby/blob/f55250e67b89a2cbfda7700719f1019702249604/packages/gatsby-source-filesystem/src/create-file-node-from-buffer.js#L135))

### DIFF
--- a/packages/gatsby-source-filesystem/index.d.ts
+++ b/packages/gatsby-source-filesystem/index.d.ts
@@ -22,7 +22,7 @@ export interface CreateFilePathArgs {
 export interface CreateRemoteFileNodeArgs {
   url: string
   store: Store
-  cache: Cache
+  cache: Cache["cache"]
   createNode: Function
   createNodeId: Function
   parentNodeId?: string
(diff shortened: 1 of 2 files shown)


freiksenet:
Thanks a lot!

sidharthachatterjee:
Published in `gatsby-source-filesystem@2.1.2` and `gatsby@2.12.1`

PR #15270, opened 2019-07-01
fix(*): enable separate config for node_modules transpilation

wardpeet:
## Description
Implements point 1 & 2 from #15213. Mostly stolen from create-react-app.

We revert the changes done by https://github.com/gatsbyjs/gatsby/pull/14111 in the js loader and create a new loader called `dependencies` to do specific work on node_modules.

Inside the `dependencies` loader, we set every optimisation to false. We also disable sourcemap generation as this might fail on big files (ex #15190).

It's not really much faster.

| site | before node transpilation | after node transpilation | This PR |
|-----------------|---------------------------|--------------------------|---------|
| gatsbyjs.org | 88.907s | 110s | 91.936s |
| moonmeister.net | 55.749s | 82s | 57.216s |
| default-starter | 9.444s | 13s | 13s |

More PRs will follow to speedup webpack build time. Sadly most of these PRs will only work on next build.

## Related Issues
Fixes #15190
Fixes #15183
Fixes #15207

PR #15284, opened 2019-07-01
feat(gatsby): enable babel in deps

sidharthachatterjee:
This changes our setup to run babel on dependencies in `node_modules` that use `gatsby`, during both:
- query compilation
- code compilation

This follows up the work in https://github.com/gatsbyjs/gatsby/pull/15270 and https://github.com/gatsbyjs/gatsby/pull/15179

A follow up to this will include a better (and faster) heuristic when fetching the dependency tree as recommended by @KyleAMathews


packages/gatsby/src/utils/gatsby-dependents.js
```diff
@@ -1,13 +1,14 @@
 import { store } from "../redux"
+import { memoize } from "lodash"
 import rpt from "read-package-tree"
 
 // Returns [Object] with name and path
-module.exports = async () => {
+module.exports = memoize(async () => {
```
wardpeet: I wonder if we need to memoize and not just store the promise... Not sure if there is any benfit over my approach as we use lodash already in gatsby.

```js
let memoizedPromise;
module.exports = () => {
  if (memoizedPromise) {
    return memoizedPromise;
  }

  memoizedPromise = new Promise(async resolve => {
    const { program } = store.getState()
    const allNodeModules = await rpt(program.directory)
    resolve(allNodeModules.children.filter(
      node =>
        (node.package.dependencies && node.package.dependencies[`gatsby`]) ||
        (node.package.peerDependencies && node.package.peerDependencies[`gatsby`])
    ))
  });

  return memoizedPromise;
})

File: packages/gatsby/src/utils/webpack-utils.js [hunk 2]
```diff
@@ -290,13 +290,25 @@ module.exports = async ({
...
   {
-    let js = (options = {}) => {
+    let js = (
+      { modulesThatUseGatsby, ...options } = { modulesThatUseGatsby: [] }
+    ) => {
       return {
         test: /\.(js|mjs|jsx)$/,
-        exclude: vendorRegex,
+        include: modulePath => {
+          if (vendorRegex.test(modulePath)) {
```
wardpeet: I would write this with an early return, doesn't really matter but that's my code style, no big deal though 😛 
```js
// when it's not coming from node_modules we treat it as a source file.
if (!vendorRegex.test(modulePath)) {
  return true;
}

// If the module uses Gatsby as a dependency we want to treat it as src so we can extract queries, ...
return modulesThatUseGatsby.some(module =>
  modulePath.includes(module.path)
)
```

packages/gatsby/src/utils/webpack-utils.js
```diff
@@ -326,7 +342,28 @@ module.exports = async ({
 
       return {
         test: /\.(js|mjs)$/,
-        exclude: /@babel(?:\/|\\{1,2})runtime|core-js/,
+        exclude: modulePath => {
+          if (vendorRegex.test(modulePath)) {
+            // If dep uses Gatsby, exclude
+            if (
```
wardpeet: In this one I would drop the else if and just use a regular if
```
if (
  modulesThatUseGatsby.some(module =>
    modulePath.includes(module.path)
  )
) {
  return true
}
// If dep is babel-runtime or core-js, exclude
if (/@babel(?:\/|\\{1,2})runtime|core-js/.test(modulePath)) {
  return true
}
```

packages/gatsby/src/utils/webpack.config.js
```diff
@@ -249,7 +251,9 @@ module.exports = async (program, directory, suppliedStage) => {
     // Common config for every env.
     // prettier-ignore
     let configRules = [
-      rules.js(),
+      rules.js({
+        modulesThatUseGatsby: modulesThatUseGatsby
```
wardpeet:
```suggestion
        modulesThatUseGatsby,
```

packages/gatsby/src/utils/webpack.config.js
```diff
@@ -260,7 +264,11 @@ module.exports = async (program, directory, suppliedStage) => {
     // Speedup 🏎️💨 the build! We only include transpilation of node_modules on javascript production builds
     // TODO create gatsby plugin to enable this behaviour on develop (only when people are requesting this feature)
     if (stage === `build-javascript`) {
-      configRules.push(rules.dependencies())
+      configRules.push(
+        rules.dependencies({
+          modulesThatUseGatsby: modulesThatUseGatsby,
```
wardpeet:
```suggestion
          modulesThatUseGatsby,
```

sidharthachatterjee:
Blocked by https://github.com/gatsbyjs/gatsby/pull/15308

packages/gatsby/src/utils/gatsby-dependents.js
sidharthachatterjee: Think this is fine since that's what lodash is doing internally in this case anyway

packages/gatsby/src/utils/webpack-utils.js
sidharthachatterjee: Yes! Doing this

packages/gatsby/src/utils/webpack-utils.js
sidharthachatterjee: This is much better! Thank you

sidharthachatterjee:
Published in `gatsby@2.12.1`

ksjogo:
Is there some way to exclude specific folders from this?
I have a monorepo with a basepackage depending on gatsby and providing some config and template components. This package's src/ is pretranspiled via tsc into a lib/ folder.
The concrete gatsby site requires this shared package and should only read lib/
But it is actually reading src/ as well due to it being a package depending on gatsby.
npmignore doesn't help/work here, as the dhared  package is symlinked from the site and built in place.

PR #15285, opened 2019-07-01
feat(gatsby-transformer-sharp): Add default types for ImageSharp

sidharthachatterjee:
Now that themes is _almost_ ready to go, we've surfaced upon an issue in https://github.com/ChristopherBiscardi/gatsby-themes-stable-test/issues/1 where a theme without any content will break with 

```
There was an error in your GraphQL query:

Unknown type "ImageSharpFixed".

File: node_modules/gatsby-transformer-sharp/src/fragments.js:5:37
```

This occurs because the node the above definition is trying to create a fragment on (`ImageSharp`) doesn't exist because no image file was present in the project. 

To fix this, we need to add default types for `ImageSharp`, `File` and more in the respective plugins that _own_ them. 

CC @freiksenet @pieh @KyleAMathews

### DIFF
--- a/packages/gatsby-source-filesystem/src/gatsby-node.js
+++ b/packages/gatsby-source-filesystem/src/gatsby-node.js
@@ -47,7 +47,7 @@ exports.sourceNodes = (
 
   if (createTypes) {
     createTypes(`
-      type File implements Node {
+      type File implements Node @infer {
         id: ID!
       }
     `)
--- a/packages/gatsby-source-filesystem/src/gatsby-node.js
+++ b/packages/gatsby-source-filesystem/src/gatsby-node.js
@@ -43,7 +43,15 @@ exports.sourceNodes = (
   { actions, getNode, createNodeId, hasNodeChanged, reporter, emitter },
   pluginOptions
 ) => {
-  const { createNode, deleteNode } = actions
+  const { createNode, deleteNode, createTypes } = actions
+
+  if (createTypes) {
+    createTypes(`
+      type File implements Node {
+        id: ID!
+      }
+    `)
+  }
 
   // Validate that the path exists.
   if (!fs.existsSync(pluginOptions.path)) {
--- a/packages/gatsby-source-filesystem/src/gatsby-node.js
+++ b/packages/gatsby-source-filesystem/src/gatsby-node.js
@@ -43,15 +43,7 @@ exports.sourceNodes = (
   { actions, getNode, createNodeId, hasNodeChanged, reporter, emitter },
   pluginOptions
 ) => {
-  const { createNode, deleteNode, createTypes } = actions
-
-  if (createTypes) {
-    createTypes(`
-      type File implements Node @infer {
-        id: ID!
-      }
-    `)
-  }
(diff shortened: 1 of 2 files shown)


packages/gatsby-source-filesystem/src/gatsby-node.js
```diff
@@ -43,7 +43,15 @@ exports.sourceNodes = (
   { actions, getNode, createNodeId, hasNodeChanged, reporter, emitter },
   pluginOptions
 ) => {
-  const { createNode, deleteNode } = actions
+  const { createNode, deleteNode, createTypes } = actions
+
+  if (createTypes) {
+    createTypes(`
+      type File implements Node {
```
freiksenet:
```suggestion
      type File implements Node @infer {
```

packages/gatsby-transformer-sharp/src/gatsby-node.js
```diff
@@ -18,3 +18,15 @@ exports.onPreExtractQueries = async ({ store, getNodesByType }) => {
     `${program.directory}/.cache/fragments/image-sharp-fragments.js`
   )
 }
+
+exports.sourceNodes = ({ actions }) => {
+  const { createTypes } = actions
+
+  if (createTypes) {
+    createTypes(`
+      type ImageSharp implements Node {
```
freiksenet:
```suggestion
      type ImageSharp implements Node @infer {
```

packages/gatsby-source-filesystem/src/gatsby-node.js
```diff
@@ -43,7 +43,15 @@ exports.sourceNodes = (
   { actions, getNode, createNodeId, hasNodeChanged, reporter, emitter },
   pluginOptions
 ) => {
-  const { createNode, deleteNode } = actions
```
freiksenet: Use `createSchemaCustomization` instead of sourceNodes.

packages/gatsby-transformer-sharp/src/gatsby-node.js
```diff
@@ -18,3 +18,15 @@ exports.onPreExtractQueries = async ({ store, getNodesByType }) => {
     `${program.directory}/.cache/fragments/image-sharp-fragments.js`
   )
 }
+
+exports.sourceNodes = ({ actions }) => {
```
freiksenet: Ditto about `createSchemaCustomization`.

packages/gatsby-source-filesystem/src/gatsby-node.js
pieh: If we would use `createSchemaCustomization` here, users would be required to run at least `gatsby@2.12.0` and it will inevitably cause issues for people updating just plugin and not core gatsby package.

IMO, if this is just more correct API to use, but doesn't ultimately change how things work, it's just safer to keep it in `sourceNodes` (at least until we need to do major release for that plugin where we can bump minimal supported Gatsby version without interupting users)

sidharthachatterjee:
Published in `gatsby-transformer-sharp@2.2.1`

PR #15293, opened 2019-07-01
Add running yarn to tutorial

shansmith01:
When running yarn workspace example develop I was getting the following error

There was a problem loading the local develop command. Gatsby may not be installed in your site's "node_modules" directory. Perhaps you need to run "npm install"? You might need to delete your "package-lock.json" as well.

Packages need to be installed first

### DIFF
--- a/docs/blog/2019-05-22-setting-up-yarn-workspaces-for-theme-development/index.md
+++ b/docs/blog/2019-05-22-setting-up-yarn-workspaces-for-theme-development/index.md
@@ -140,7 +140,7 @@ Add Gatsby develop and build scripts to the example site's `package.json`.
 Install all your packages by running yarn in the root directory.
 
 ```shell
-yarn 
+yarn
 ```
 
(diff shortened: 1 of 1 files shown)


gillkyle:
Maybe I'm mistaken but shouldn't this step: https://github.com/gatsbyjs/gatsby/blob/master/docs/blog/2019-05-22-setting-up-yarn-workspaces-for-theme-development/index.md#L53 take care of Gatsby installation? 

Sorry you ran into trouble!

shansmith01:
No did not work for some reason. Unforutanlty I am not experienced with yarn. I normally use NPM, so I'm not sure if I'm messed up, but doing what I suggest worked

gillkyle:
I just went through the steps again from the beginning to verify and they seem to work okay. Thanks for offering the suggestion but unfortunately I'm going to close this since adding this step could confuse people as to why if it's necessary if things work okay without it.

If you have any more trouble and see that something is confusing feel free to open another issue!

PR #15310, opened 2019-07-02
fix(gatsby): adds the missing micromatch dependency

arcanis:
## Description

Package required here:

https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/redux/run-sift.js#L6

## Related Issues

n/a

### DIFF
--- a/packages/gatsby/package.json
+++ b/packages/gatsby/package.json
@@ -88,6 +88,7 @@
     "lodash": "^4.17.10",
     "md5": "^2.2.1",
     "md5-file": "^3.1.1",
+    "micromatch": "^3.1.10",
     "mime": "^2.2.0",
     "mini-css-extract-plugin": "^0.4.0",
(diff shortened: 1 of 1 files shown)


packages/gatsby/package.json
```diff
@@ -88,6 +88,7 @@
     "lodash": "^4.17.10",
     "md5": "^2.2.1",
     "md5-file": "^3.1.1",
+    "micromatch": "^3.1.10",
```
sidharthachatterjee: Need to update the top level `yarn.lock` as well

arcanis: Should be fine, workspaces aren't currently part of the lockfile (and `micromatch@^3.1.10` already was part of it).

sidharthachatterjee:
Published in `gatsby@2.12.1`

PR #15313, opened 2019-07-02
fix(gatsby-image): Update TypeScript types

brimtown:
## Description

In the course of merging #13395, the implementation was switched from using separate
`fluidImages` and `fixedImages` props to overloading `fluid` and `fixed` to take arrays.
The TypeScript types were not updated to match this. This fixes the types to reflect that: https://www.gatsbyjs.org/packages/gatsby-image/#gatsby-image-props


sidharthachatterjee:
Published in `gatsby-image@2.2.4`

PR #15321, opened 2019-07-02
fix(gatsby): Make root plural fields non nullable 

spawnia:
## Description

Change the setup of both the internal schema and the `gatsby-source-graphql` plugin to add root query fields as non-nullable.

This helps to reduce tedious null-checks when querying via `useStaticQuery` or other mechanisms. This became especially apparent when using TypeScript code generation:

image

Actually, the `file` query will never return null - and if it does, there is something wrong anyways, so throwing would be the right thing to do. I am forced to ignore the error or add type coercion like `data!.placeholderImage`.

With this change, TypeScript users will have a much better time:

image


spawnia:
Is this something you are generally interested in pursuing further? There are a bunch more places in the GraphQL schema where adding NonNull types would make type safety easier to achieve.

packages/gatsby/src/schema/schema.js
```diff
@@ -741,8 +741,10 @@ const addTypeToRootQuery = ({ schemaComposer, typeComposer }) => {
   const queryName = _.camelCase(typeName)
   const queryNamePlural = _.camelCase(`all ${typeName}`)
   schemaComposer.Query.addFields({
-    [queryName]: typeComposer.getResolver(`findOne`),
-    [queryNamePlural]: typeComposer.getResolver(`findManyPaginated`),
+    [queryName]: new GraphQLNonNull(typeComposer.getResolver(`findOne`)),
```
pieh: This technically isn't guaranteed to always not be null

pieh:
I think this makes a lot of sense to do, just need to be careful where non-null decorator is applied

packages/gatsby/src/schema/schema.js
spawnia: @pieh What would a scenario were that field returns `null` look like?

Can a page still be successfully built or does it indicate an error that fails the build?

spawnia:
I am having a hard time figuring out why the tests are failing.

It seems like the queries to the schema are set up in a way such that the data is not actually returned from the registered resolver, which is why the tests are failing when trying to query such a field:

```
  ● Query schema › on children fields › handles Node interface children field

    expect(received).toBeUndefined()

    Received: [[GraphQLError: Cannot return null for non-nullable field Query.allFile.]]

      182 |         },
      183 |       }
    > 184 |       expect(results.errors).toBeUndefined()
          |                              ^
      185 |       expect(results.data).toEqual(expected)
      186 |     })
      187 | 

      at Object.toBeUndefined (packages/gatsby/src/schema/__tests__/queries.js:184:30)
```

How can we fix this?

packages/gatsby/src/schema/schema.js
pieh: It might not be reasonable, but you can do
```
someType(id: { eq: "id-of-node-that-doesnt-exist" }) {
  id
}
```

Which will return `null` because node can't be found.

Other scenarios are - with schema customizations you can define types even if there are no nodes for that type - so those will always be empty (even without any filters)

spawnia: I do get the first scenario. Marking the field as NonNull could still be better, as it would uncover an error right away. My assumption is that you always require the data you want to find.

If there are fields where you try to find something dynamically **and** don't know about it's existence **and** handle the case of not finding something, a nullable field could make sense.

In the case of the `allSomeType` fields - i think those should always return at least an empty list, so `[SomeType]!` instead of `[SomeType]`.

packages/gatsby/src/schema/schema.js
```diff
@@ -925,7 +925,7 @@ const addTypeToRootQuery = ({ schemaComposer, typeComposer }) => {
       },
       resolve: findManyPaginated(typeName),
     },
-  })
+  }).makeFieldNonNull([queryName, queryNamePlural])
```
sidharthachatterjee: As @pieh noted `queryName` is _not_ non-nullable since there are indeed legitimate instances where a query to `someType(id: { eq: "id-of-node-that-doesnt-exist" }) { id }` could return `null`

Take for instance a series of queries one after another in `gatsby-node` whether `someType` is fetched for an `id` that may come from somewhere else. 

`queryNamePlural` can on the other hand safely be a `[Type]!`

What are your thoughts to switching over to `[Type]!` for just the plural form in this PR?

spawnia: I am now convinced that is correct, i changed the PR.

PR #15323, opened 2019-07-02
chore(scripts):  mirror theme starters to independent repos and add theme packages to lerna

sidharthachatterjee:
This is temporary until theme packages are moved over to `/packages` and theme starters are moved over to `/starters`

CC @amberleyromo @ChristopherBiscardi

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -207,6 +207,26 @@ jobs:
       - run: git config --global user.email "admin@gatsbyjs.com"
       - run: sh ./scripts/clone-and-validate.sh starters true
 
+  theme_starters_validate:
+    executor: node
+    steps:
+      - checkout
+      - run: ./scripts/assert-changed-files.sh "themes/*|.circleci/*"
+      - run: sh ./scripts/clone-and-validate-theme-starters.sh
+
+  theme_starters_publish:
+    executor: node
+    steps:
+      - checkout
+      - <<: *restore_cache
+      - <<: *install_node_modules
+      - <<: *persist_cache
+      - run: yarn markdown
+      - run: sudo apt-get update && sudo apt-get install jq # jq is helpful for parsing json
+      - run: git config --global user.name "GatsbyJS Bot"
+      - run: git config --global user.email "admin@gatsbyjs.com"
+      - run: sh ./scripts/clone-and-validate-theme-starters.sh true
+
 workflows:
   version: 2
   build-test:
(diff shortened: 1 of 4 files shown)


lerna.json
```diff
@@ -11,7 +11,11 @@
     }
   },
   "loglevel": "success",
-  "packages": ["packages/*"],
+  "packages": [
+    "packages/*",
+    "themes/gatsby-theme-blog",
```
DSchau: Is there some other heuristic we could use?

If we don’t want them published — what about `private: true` or something?

(also if this has bee discussed and I’ve missed it — ignore!)

scripts/clone-and-validate-theme-starters.sh
```diff
@@ -0,0 +1,46 @@
...
+CLONE=$1
+IS_CI="${CI:-false}"
+BASE=$(pwd)
+COMMIT_MESSAGE=$(git log -1 --pretty=%B)
+
+if [ "$IS_CI" = true ]; then
+  sudo apt-get update && sudo apt-get install jq
+fi
+
+for folder in "themes/gatsby-starter-blog-theme" "themes/gatsby-starter-notes-theme" "themes/gatsby-starter-theme"; do
```
DSchau: This seems a little less reusable than it was — and I’m not sure why it’s a separate script? Could you share some info?

It seems like the central problem that needs to be solved is iterating only a subset of themes, right?

.circleci/config.yml
```diff
@@ -207,6 +207,26 @@ jobs:
...
       - run: sh ./scripts/clone-and-validate.sh starters true
 
+  theme_starters_validate:
+    executor: node
+    steps:
+      - checkout
+      - run: ./scripts/assert-changed-files.sh "themes/*|.circleci/*"
+      - run: sh ./scripts/clone-and-validate-theme-starters.sh
+
+  theme_starters_publish:
```
DSchau: Because this is effectively the same as starters_publish, I’d recommend making it a template and exposing variables. (e.g. the way you’ve gone now you _could_ do a separate shell script path, but I’d probably just pass a different glob pattern)

lerna.json
sidharthachatterjee: The `themes` folder contains two themes packages as well. This is temporary till they're moved out to `/starters` so I didn't bother.

scripts/clone-and-validate-theme-starters.sh
sidharthachatterjee: Yeah, only want to go over those three specific folders in the `themes` directory (since there are also two more theme packages that this shouldn't run on) 

Again, didn't bother because will be moving these over to the `packages` and `starters` directories after we're stable and throw this away.

PR #15331, opened 2019-07-02
fix(gatsby): Default to empty arrays in load-themes

sidharthachatterjee:


sidharthachatterjee:
Published in `gatsby@2.13.1`

PR #15333, opened 2019-07-02
feat(themes): Update theme-ui plugin

jxnblk:
## Description

This is a small update to the themes for the latest stable API and the renamed theme-ui plugin.

- Updates themes to `gatsby@^2.13.0`
- Updates to use the (renamed) `gatsby-plugin-theme-ui` package
- Use the `plugins` array in `gatsby-config.js` for themes

**Questions:** 

- [x] It sounded like the `package-lock.json` files in the starters will automatically get updated, but is there anything additional I need to do here?
- [ ] Should I update the READMEs as well?


ChristopherBiscardi:
We'll want to set the lower bound for the version to 2.13.1, but otherwise looks good to me

PR #15348, opened 2019-07-03
feat(gatsby-transformer-remark): add meta support to code block

nihgwu:
## Description

ref https://github.com/syntax-tree/mdast-util-to-hast/pull/33

Another option is just use the code handler to parse the meta data if we don't want to upgrade the dependencies, as the meta data is a part of `lang` in the old version of `remark-parse`

as mentioned here https://github.com/remarkjs/remark/issues/342, info string support for fenced code block are common in both gfm and CommonMark, it would be nice we can use it, like the `gatsby-remark-prismjs` plugin,  https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-remark-prismjs/README.md#line-numbering, we can see that we can't have space between the language and the meta, so the language is actually invalid and there is no syntax highlight for this code block, with this change ,we can have both code highlight and the meta data for further usage


freiksenet:
This change is good, however I'm wondering if upgrading remark would be a breaking change for some users.

freiksenet:
I guess I'll make it a minor release.

freiksenet:
Published as `gatsby-transformer-remark@2.6.0`. Thank you for this!

PR #15349, opened 2019-07-03
chore(starters): add lewis-gatsby-starter-i18n

lewislbr:


pieh:
Hey, demo url - https://lewis-gatsby-starter-i18n.firebaseapp.com/ serves "Site Not Found" for me - is that correct url?

lewislbr:
True, it seems the previous deploy failed. 

Fixed it, thanks!

PR #15358, opened 2019-07-03
fix(gatsby): Add tests for and partially fix circular reference issue.

### LINKED ISSUE #11364: gatsby-source-contentful: Maximum call stack size exceeded

cameron-martin:
## Description

This particular issue occurs when a rich text hyperlink links to a page that is involved in a circular reference. This causes a circular data structure to be created, which causes problems in some algorithms that traverse this.

I believe this is the same issue as issue #11364.

I've added a new e2e test called "contentful" that contains a test that is a minimal reproduction of this issue. This uses a contentful space where I don't mind people having the API keys, but possibly for making tests deterministic this should be mocked. For convenience I have also [attached an export of the space](https://github.com/gatsbyjs/gatsby/files/3355189/contentful-export-mn2gkf1rxy7p-master-2019-07-03T14-34-73.zip).

This also contains a partial fix to the issue - it fixes the node tracking algorithm to deal with circular data structures. However, it now fails when doing schema inference.

## Related Issues

Fixes #11364 
Supersedes #14612


cameron-martin:
cc @wardpeet

cameron-martin:
@wardpeet Did you end up doing that second pair programming session regarding this issue that you mentioned during our session?

LekoArts:
Hey @cameron-martin , could you please move these tests to unit tests as suggested by Sid? That would be super awesome, thanks!

cameron-martin:
I don't see how unit tests would be sufficient to test this. The fix touches multiple different units (basically anywhere that breaks when a node is cyclic), and if we were to just rely on unit tests there would be nothing to verify that we haven't introduced any more such units in the future.

LekoArts:
> However, I can roll this test case into one of the other e2e suites due to speed.

Yeah, that would be awesome, too 👍 And thanks for understanding (as our e2e tests already take up quite some time).

cameron-martin:
So at the time I was working for an organisation that this problem was affecting but no longer do so I have little interest in fixing it since I'm no longer using Gatsby. Sorry

pieh:
> So at the time I was working for an organisation that this problem was affecting but no longer do so I have little interest in fixing it since I'm no longer using Gatsby. Sorry

This is ok! Don't be sorry! :) We just needed to know what's the status of this. Let's close this PR, as e2e tests would need to be reworked and node tracking have also changed since this PR started.

Issue #15148, opened 2019-06-26
[gatsby-source-wordpress] send custom cookies/headers with api request


karlsander:
## Summary

I would like a feature in gatsby-source-wordpress to send custom cookies. The reason is we want to send a "dev mode" cookie to never get cached API responses. I would be happy to create a PR for the feature but wanted to ask beforehand if you believe it makes sense for the plugin. Since cookies are (can be?) send as headers, it could also be full custom headers feature.

### Basic example

I would imagine it being configured like this

```js
    {
      resolve: 'gatsby-source-wordpress',
      options: {
        baseUrl: 'gatsbyjsexamplewordpress.wordpress.com',
        auth: {
          headers: {
            cookies: 'dev_mode_enabled=1;',
          },
        },
      },
    }
```

or

```js
   {
      resolve: 'gatsby-source-wordpress',
      options: {
        cookies: {
          dev_mode_enabled: '1',
        },
      },
    }
```

PR #15361, opened 2019-07-03
feat(gatsby-source-wordpress): add option to send cookies

karlsander:
## Description

This PR adds a "cookies" configuration option to gatsby-source-wordpress that takes an object of key value pairs to send as cookies with requests to the wordpress api.

## Related Issues

I previously created issue #15148 for this


freiksenet:
Thanks!

PR #15362, opened 2019-07-03
fix(lerna): add themes packages to workspaces

sidharthachatterjee:
`packages` in `lerna.json` is ignored when `useWorkspaces` is set to true


sidharthachatterjee:
Closed in favour of https://github.com/gatsbyjs/gatsby/pull/15572

PR #15368, opened 2019-07-03
Change default Referrer-Policy to no-referrer-when-downgrade

dja:
## Description

Changing this will allow support for passing referrer data on to integrated advertising scripts and subsequent subdomain destinations.

## Related Issues

Related to #13452


wardpeet:
Thanks for opening this PR. Sadly we'll be closing this one. Changing this header is considered a breaking change and it's not a change we would like to make. I do think it should be configurable.

I was hoping this would fix your issue:
```
    {
      resolve: `gatsby-plugin-netlify`,
      options: {
        headers: {
          '/*': [
            'Referrer-Policy: no-referrer-when-downgrade',
          ]
        }
      }
    },
```

Somehow we don't merge smartly. If you're up for it, we would love getting a PR that merges these headers properly.

PR #15369, opened 2019-07-03
docs(www): Revamp How Gatsby Works with GitHub Pages

### LINKED ISSUE #15229: [docs][guides] improve publishing to Github Pages doc
## Summary

In feedback from gatsbyjs.org, the [Deploying to GitHub Pages](https://www.gatsbyjs.org/docs/how-gatsby-works-with-github-pages/) guide could use an update. Specifically:

"What about if I wanted to use the doc/ folder option instead of a dedicated gh-pages branch?"

That process is described here: https://help.github.com/en/articles/configuring-a-publishing-source-for-github-pages#publishing-your-github-pages-site-from-a-docs-folder-on-your-master-branch

It would be good to add a section to the GitHub Pages deployment guide to provide that information inline, so people don't have to go hunting for it. If anyone else has encountered gotchas with GitHub pages, we could also include them as part of this effort.

The page needing an update: https://www.gatsbyjs.org/docs/how-gatsby-works-with-github-pages/

Aswinpookkatt:
## Description
**Added** 

- Enabling GitHub Pages to publish your site from master or gh-pages

- Publishing your GitHub Pages site from a /docs folder on your master branch

## Related Issues
 

- Fixes #15229


gillkyle:
Thanks @Aswinpookkatt! It looks like some of these changes would be helpful interspersed throughout the page as well. Instead of just appending them to the end adding info where it makes sense contextually could help this make more sense. 

For example, it looks like there's this existing note about how the `gh-pages` package moves files to a branch: https://github.com/gatsbyjs/gatsby/blob/master/docs/docs/how-gatsby-works-with-github-pages.md#L29

Your changes add some more context to that in each section where it says you need a branch with a certain name. I think this would benefit from updating the existing docs with the additional info you've added so it reads more like 1 section with context provided where it's needed rather than 3 or 4 different guides that each have similar steps?

marcysutton:
Hi there @Aswinpookkatt, do you have any time to work on the PR feedback above?

Aswinpookkatt:
Okay @marcysutton

gillkyle:
Hey @Aswinpookkatt have you been able to make any updates? If not, I'd be happy to jump in and try and help wrap this up!

Aswinpookkatt:
@gillkyle my checks are failing.
can you help me,why does this happen?

gillkyle:
@Aswinpookkatt Those e2e tests are failing right now for other PRs as well, since this docs page shouldn't affect those tests, you don't need to worry about them. 

I think more importantly we could combine the sections you wrote with the existing docs rather than just adding it to the bottom of the old doc since there is useful information in both that would work well together. Would you like to do that or should I take a crack at it?

gillkyle:
I went ahead and combined the updates into what was existing. Thanks for adding your help on this @Aswinpookkatt!

PR #15370, opened 2019-07-03
fix(gatsby-theme-blog): drop path.join to support windows

wardpeet:
## Description
creating urls with path.join isn working on osx & linux but sadly not on windows. This PR fixes that. I join by `/` and replace `//` or more into one `/`

before:
image

after:
image

### DIFF
--- a/themes/gatsby-theme-blog/gatsby-node.js
+++ b/themes/gatsby-theme-blog/gatsby-node.js
@@ -3,7 +3,7 @@ const path = require(`path`)
 const mkdirp = require(`mkdirp`)
 const crypto = require(`crypto`)
 const Debug = require(`debug`)
-const { joinPath } = require(`gatsby-core-utils`)
+const { urlResolve } = require(`gatsby-core-utils`)
 
 const debug = Debug(`gatsby-theme-blog`)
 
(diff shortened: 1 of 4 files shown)


sidharthachatterjee:
Can't we use
```

const url = require('url');
url.resolve(basePath, dir, node.parent.name)

```

wardpeet:
url.resolve has another use case. Check out https://nodejs.org/api/url.html#url_url_resolve_from_to it only works for 2 values

Currenly blocked on getting https://github.com/gatsbyjs/gatsby/pull/15451 shipped

talves:
@wardpeet I tested the `joinPath` utility you propose on the slug in `onCreateNode` function (`toPostPath`) in my example and it will not solve the issue of an unexpected slug path. Instead of getting a slug with back slashes, which is happening now, that should be forward slashes, it returns the path with double back slashes now.

The issue is a node `path.join` returns a slug path with back slashes on windows that should be forward slashes in the slug for the browser. There needs to probably be a utility that converts paths to a valid relative slug path instead like below.

```javascript
  const toSlugPath = node => {
    const { dir } = path.parse(node.relativePath)
    return join.path(basePath, dir, node.name).replace(/\\/g, '/')
  }
```

Of course, the `toNotesPath` function is a different issue that probably needs the proposed utility, but I did not look at that one.

wardpeet:
It seems like joinPath isn't doing what I was told it did. It converts `\` into `\\` to safely store it to disk.

I have a new PR up https://github.com/gatsbyjs/gatsby/pull/15510 that adds urlResolve

PR #15371, opened 2019-07-03
fix(gatsby-starter-blog-theme): fix package-lock

pieh:
I had partial changes to package-lock file that I didn't commit in https://github.com/gatsbyjs/gatsby/pull/15366 - this should fix this


wardpeet:
🤞 🌮

PR #15379, opened 2019-07-03
fix(gatsby): use joi validation result instead of payload

wardpeet:
## Description
`Joi.validate` returns a normalized config. Before this PR we weren't using it and we redid everything that joi can do for us. Now evertyhing is in one place.

Tests weren't working either so I've fixed that as well.


DSchau:
Successfully published:
 - gatsby@2.13.38

PR #15381, opened 2019-07-03
chore(starters): add lewis-gatsby-starter-basic

lewislbr:


lewislbr:
I don't know what went wrong.  😣

Screen Shot 2019-07-04 at 10 37 57

According to this there is a code style issue in the starters.yml due to a failed prettier run.

I checked the file and seems to be alright.

PR #15384, opened 2019-07-03
feat(gatsby-plugin-sharp): enable gatsby cloud processing

wardpeet:
## Description
This enables running gatsby-sharp out of process on gatsby cloud. By setting `GATSBY_CLOUD_IMAGE_SERVICE_URL`, the job will be offloaded from the current process.

cc @abhiaiyer91 @KyleAMathews @pieh


packages/gatsby-plugin-sharp/src/process-file.js
```diff
@@ -75,13 +76,42 @@ const argsWhitelist = [
 exports.processFile = (file, transforms, options = {}) => {
   let pipeline
   try {
+    // adds gatsby cloud image service to gatsby-sharp
+    // this is an experimental api so it can be removed without any warnings
+    if (process.env.GATSBY_CLOUD_IMAGE_SERVICE_URL) {
```
freiksenet: Arguably we should fall back to local processing if this fails.

wardpeet: We should fail hard. If this fails something isn't right.

freiksenet: Okay!

packages/gatsby-plugin-sharp/package.json
```diff
@@ -11,6 +11,7 @@
     "async": "^2.1.2",
     "bluebird": "^3.5.0",
     "fs-extra": "^7.0.0",
+    "got": "^9.6.0",
```
pieh: Only potential issue here is got needs node version >= 8.6 ( https://github.com/sindresorhus/got/blob/v9.6.0/package.json#L8-L10 ) while we technically should work on >= 8.0

thoughts on this?

wardpeet: Wow good catch, I can revert to something like axios or request

pieh: Or use `got@7` or `got@8` which works up to node 4 ;)

freiksenet:
:ship: :it:

packages/gatsby-plugin-sharp/src/process-file.js
```diff
@@ -75,13 +76,42 @@ const argsWhitelist = [
...
+    }
+
     pipeline = sharp(file)
 
     // Keep Metadata
     if (!options.stripMetadata) {
       pipeline = pipeline.withMetadata()
     }
   } catch (err) {
+    console.log(err)
```
wardpeet: need to remove this one

packages/gatsby-plugin-sharp/src/process-file.js
```diff
@@ -75,6 +76,31 @@ const argsWhitelist = [
...
+    if (process.env.GATSBY_CLOUD_IMAGE_SERVICE_URL) {
+      let cloudPromise
+
+      return transforms.map(transform => {
+        if (!cloudPromise) {
+          cloudPromise = got
+            .post(process.env.GATSBY_CLOUD_IMAGE_SERVICE_URL, {
+              body: {
+                file,
+                transforms,
```
muescha: you are sure to send all `transforms` instead of `transform`?

wardpeet: yeah we do 😄 Thanks for asking.

muescha: 🤦🏻‍♂️ Ok. Sure. I see now It is only called once for a transactions

PR #15390, opened 2019-07-04
Add placeholder page to fix 404 in themes docs

falconmick:
## Description

Currently there is a 404 on this page https://www.gatsbyjs.org/docs/themes/what-are-gatsby-themes/ (Getting Started Link) This adds a placeholder for SEO sake + making sure new developers understand that they didn't do anything wrong and that this page doesnt exist yet.

We should also make this page, still too green on themes to help as of yet

## Related Issues

### DIFF
--- a/docs/docs/themes/getting-started,md
+++ b/docs/docs/themes/getting-started,md
@@ -1,5 +1,0 @@
----
-title: Getting Started
----
-
-_Getting started docs are coming soon!_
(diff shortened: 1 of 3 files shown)


wardpeet:
Thank you for creating this PR. We use createRedirect to redirect from `/docs/themes/getting-started` to `/docs/themes/using-a-gatsby-theme`. [createRedirect](https://www.gatsbyjs.org/docs/actions/#createRedirect) by default doesn't redirect on client navigation, only when you visit the link directly.

I've updated the docs page to match the new url. Thank you for surfacing this bug!

Issue #15375, opened 2019-07-03
Blog post not loading on gatsbyjs.org


missmatsuko:
## Description
Not sure if this is the right place for this issue, but anyways~

I was trying to find out more about Gatsby Days. This post showed up in the blog search results, but doesn't load:
https://www.gatsbyjs.org/blog/2018-11-27-gatsby-days/

There's a lot of errors in the console for this page.

### Steps to reproduce
Go to https://www.gatsbyjs.org/blog/2018-11-27-gatsby-days/

### Expected result
The page should load.

### Actual result
The page doesn't load, and there's many errors in the console.

### Environment
n/a

freiksenet:
Temporarily fixed by removing the offending image. Likely a bug in how mdx treats images.

missmatsuko:
@freiksenet Do you know if the change live already? I saw the quick fix PR was merged, but the page still doesn't load for me. Maybe the site needs to be rebuilt, or I'm just seeing a cached version?

PR #15394, opened 2019-07-04
Remove image that breaks blog post

freiksenet:
Temporary fir for #15375


wardpeet:
We had a related bug https://github.com/gatsbyjs/gatsby/pull/15029

Issue #14870, opened 2019-06-18
fix: Allow local plugins to bypass node type owner in deleteNode


sidharthachatterjee:
In https://github.com/gatsbyjs/gatsby/pull/13492, we introduced a check where we only allow deletions in `deleteNode` if the plugin calling `deleteNode` is the owner of the node itself. 

However, as @jessepinho mentions in https://github.com/gatsbyjs/gatsby/pull/13492#issuecomment-503101035, local plugins (your project's `gatsby-node.js`) should be able to delete anything. 

We need to change the conditional to allow this. 

Related 
https://github.com/gatsbyjs/gatsby/issues/11930 and https://github.com/gatsbyjs/gatsby/pull/13492

PR #15399, opened 2019-07-04
feat: add a nodeFilter option

jessepinho:
## Description

This PR introduces a `nodeFilter` config option to the `gatsby-source-contentful` plugin. `nodeFilter` is a function that, when defined, will be called before calling `createNode` on each content type/entry/asset in Contentful, allowing other plugins (and `gatsby-node.js`) to control whether nodes are created based on particular business logic.

### DIFF
--- a/packages/gatsby-source-contentful/README.md
+++ b/packages/gatsby-source-contentful/README.md
@@ -152,6 +152,10 @@ For example, to filter locales on only germany `localeFilter: locale => locale.c
 
 List of locales and their codes can be found in Contentful app -> Settings -> Locales
 
+**`nodeFilter`** [function][optional] [default: `(node) => true`]
+
+Before `gatsby-source-contentful` calls `createNode` for each content type, entry, and asset, it first checks with `nodeFilter` to see whether it should. It passes the node to `nodeFilter`, and only calls `createNode` if `nodeFilter` returns true. This is an opportunity to filter out nodes before they're created based on custom business logic.
+
 **`forceFullSync`** [boolean][optional] [default: `false`]
 
 Prevents the use of sync tokens when accessing the Contentful API.
--- a/packages/gatsby-source-contentful/src/__tests__/normalize.js
+++ b/packages/gatsby-source-contentful/src/__tests__/normalize.js
@@ -53,6 +53,8 @@ describe(`Process contentful data`, () => {
   it(`creates nodes for each entry`, () => {
     const createNode = jest.fn()
     const createNodeId = jest.fn()
+    const nodeFilter = () => true
+
     createNodeId.mockReturnValue(`uuid-from-gatsby`)
     contentTypeItems.forEach((contentTypeItem, i) => {
       normalize.createContentTypeNodes({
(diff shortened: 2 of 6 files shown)


jessepinho:
cc @pieh @Khaledgarbaya

jessepinho:
Note: this PR basically supersedes #14870

jessepinho:
Is there anything I can do to help this get merged?

pieh:
What happens when you filter out nodes that are linked to from another nodes? (this is similar concern as with deleteNode approach)

jessepinho:
@pieh Hm — the approach here is that it prevents invalid nodes from ever being created. If nodes aren't even created yet, how could other nodes link to them? It's possible that I don't quite understand how node relationships are built in Gatsby...

pieh:
Nodes are linking to other nodes using following convention:
```
// for link to single node
fieldName___NODE: "id-of-other-node"
// for link to multiple nodes
fieldName___NODE: ["id-of-other-node","id-of-yet-other-node"]
```

so in example:
```
[
  {
    id: "id-of-some-article-node",
    title: "title of article",
    categories: ["id-of-category1-node"]
  },
  {
    id: "id-of-category1-node",
	title: "title of category"
  }
]
```

if you would filter out category nodes using `nodeFilter`. Then link to that node in article would still exist, but couldn't be resolved as category node it's linking to was filtered out and Gatsby would crash

jessepinho:
I see — I thought the references to other nodes required the ID that's created by Gatsby (via `createContentId` or similar).

And true, this does seem a bit more complicated than I originally thought. As you mentioned, we could run `nodeFilter` when even a _link_ is being created, so that we don't even create links for invalid nodes. The only trouble with this is that we're dealing with incomplete nodes at that point. For example, `nodeFilter` may, as part of its logic, check that the links from one node to another node are correct. But if they simply haven't been created yet, it would fail, even though it should pass after the links are created.

Here's another idea: `gatsby-source-contentful` already keeps track of links between nodes via the `foreignReferenceMap` variable. What if I kept the existing logic I have, but if `nodeFilter` ever returns `false`, I would look through the reference map and remove all links to the filtered-out node? 

(Counterpoint: that would mean modifying the nodes that have links to the filtered-out node, which in turn could make _them_ invalid per `nodeFilter`'s logic. hmm...)

sidharthachatterjee:
@jessepinho Just went through this thread and had a chat with @pieh and we both agree that there isn't really a way forward with this PR

To summarise:

- This is much harder to do than simply filtering nodes before creation 
- The change in the PR itself is not incrementally valuable without handling links so we can't merge and then follow up 
- Contentful's delta sync makes it even more complicated

> Here's another idea: gatsby-source-contentful already keeps track of links between nodes via the foreignReferenceMap variable. What if I kept the existing logic I have, but if nodeFilter ever returns false, I would look through the reference map and remove all links to the filtered-out node?

This would be tricky as well because it might lead to behaviour where filtering a certain Node removes another that might be useful. What we probably need is a _managed_ `deleteNode` (which deletes the Node _and_ references to it)

Let's close this for now based on the points above. 

Nonetheless, thank you so much for picking this up, working on it and for taking the time to have this discussion! We really appreciate it and can't wait to have you on as a contributor (looking at your other PRs) 💜

jessepinho:
@sidharthachatterjee Thanks for following up! This explanation makes sense — it's certainly a difficult problem to solve.

Regarding your point about a "managed `deleteNode`" — wouldn't that have some of the same problems? Imagine I have a content type A that has a _required_ field referencing content type B. If a B node gets filtered out or deleted, that would leave an invalid A node, since it now has an empty field that's supposed to be required.

Trying to think of the best way to solve this... if you have any brilliant ideas, please share!

Meanwhile, I suppose the only other solution is to simply filter in every query, which means duplicating a lot of filters.

Issue #11691, opened 2019-02-11
pageContext is not updated when context change


RageYL:
Hi,

The issue is not as straightforward as i thought it would be, so here is a repo to reproduce it: https://github.com/RageYL/gatsby-pageContext

And this is different from https://github.com/gatsbyjs/gatsby/issues/6097

I use context from createPage to pass data to a template (I started using gatsby 3 days ago, so not 100% sure if this is the good way), but when the data is updated the pageContext value is not updated.
The pageContext is not updated if:
- there is a query using MarkdownRemark (i tried without query or with sitePage and no problem)

If you confirm this is indeed an issue, i'm ready to dig in and submit a pull request.

jonniebigodes:
@RageYL i've picked up on your issue, i've cloned your repo into my machine installed it and did some tests, namelly:
- Created a json object with a similar structure and provided the data from the json file.
- Updated the template to implemement the livecycle method `componentDidUpdate` and `getDerivedStateFromProps` to see if it could solve the issue.
- Passed not the object but the property to `pageContext`, to see if produced an update.

As you can see bellow, the value of `datafromJson` on the right, the devtools is `x` and on the left is `b`
![noupdate](https://user-images.githubusercontent.com/22988955/52579216-a34a5e00-2e1d-11e9-87d6-ab649c2a1ca0.png)

 My reasoning for this behaviour  is that as is, the api hook, either with your object or mine, is on a "fire and forget" mode, it fetches the data once, passes it along and doesn't care about subsequent changes. What you could probably do is on `gatsby.node.js` filem is the following, you add a watcher to the file to see if it has changed, and if so, delete the page and recreate it. I have not tested this And it's not tested.

Also three days into Gatsby and already into one api that people tend to struggle more. Congratulations are in order.
Feel free to provide feedback ,so that we can close this issue

RageYL:
What i find very strange is that everything works fine if there is no query or if i change the query's content (maybe something related to MarkdownRemark resolvers and the cache).
I updated my gatsby-node.js to use createPageStatefully (better choice indeed as my data is not managed by gatsby): https://github.com/RageYL/gatsby-pageContext/blob/createPagesStatefully/gatsby-node.js

Now if i update the data and restart gatsby the change is reflected on the page, but when i update without restarting gatsby, the page still display the same value. So it's already better, but it seems there is still something wrong with the cache.

pieh:
[...]

PR #15404, opened 2019-07-04
fix(gatsby): reevaluate page query when page context is modified

RomanHotsiy:
## Description

When page context is updated by createPage the path is not marked as dirty and page query is not reevaluated resulting in the page not updating in develop mode.

It was already fixed by my in https://github.com/gatsbyjs/gatsby/pull/11048 but it appeared again in v2.2.11 because of refactoring by @KyleAMathews in https://github.com/gatsbyjs/gatsby/commit/91086d43e9b69ba357f7f7ad6bbf2eb0b8b5fbff

I added a basic test, but I'm not sure how to write a test to prevent this kind of regression in the future. If anyone has any ideas I would be happy to add a test.

## Related Issues

Related to #6097, May be related to #11691

### DIFF
--- a/packages/gatsby/src/redux/machines/__tests__/page-component.js
+++ b/packages/gatsby/src/redux/machines/__tests__/page-component.js
@@ -68,4 +68,12 @@ describe(`bootstrap`, () => {
     await sleep()
     expect(runQueuedQueries).toBeCalledWith(path)
   })
+
+  it(`will queue query when page context is changed`, async () => {
+    const service = getService({ isInBootstrap: false })
+    service.send({ type: `PAGE_CONTEXT_MODIFIED`, path: `/a/test.md` })
+    // there is setTimeout in action handler for `CONTEXT_CHANGES`
+    await sleep()
+    expect(enqueueExtractedQueryId).toBeCalledWith(`/a/test.md`)
+  })
 })
--- a/packages/gatsby/src/redux/machines/page-component.js
+++ b/packages/gatsby/src/redux/machines/page-component.js
@@ -22,6 +22,9 @@ module.exports = Machine(
       NEW_PAGE_CREATED: {
         actions: `setPage`,
       },
+      PAGE_CONTEXT_MODIFIED: {
+        actions: `rerunPageQuery`,
+      },
       QUERY_EXTRACTION_GRAPHQL_ERROR: `queryExtractionGraphQLError`,
       QUERY_EXTRACTION_BABEL_ERROR: `queryExtractionBabelError`,
     },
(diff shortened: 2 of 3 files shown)


RomanHotsiy:
Ping @pieh (you reviewed the previous PR)

pieh:
Right. There is problem with this that I'm not sure if I noted in previous implementation.

This will work only when context is modified while `gatsby develop` is running. It won't catch cases of context changes between builds as we don't persist pages created by previous builds right now

RomanHotsiy:
I think that was fixed here: https://github.com/gatsbyjs/gatsby/blob/38eda62659dfc2d97ab3045e37362df2762e073c/packages/gatsby/src/redux/actions/public.js#L344-L345

pieh:
This is what I actually meant. `oldPage` will be there only when page is updated while `gatsby develop` is running. It won't catch context changes if you change your content that would cause context to change when `gatsby develop` is not running

RomanHotsiy:
@pieh thanks for clarification!

> It won't catch cases of context changes between builds as we don't persist pages created by previous builds right now

Oh, I see. I just verified that this issue happens between builds:
- I run `gatsby build`
- I change somehow page context
- I run `gatsby build` and page is not updated since last time

That's a more severe issue... I don't know how to fix it as part of this PR though. Persisting page data is a separate bigger PR.

I believe we can't assume context to be changed every time when `oldPage` is not there as it may hurt performance. Am I right?

I believe this PR can be merged and a separate issue should be open for "between-builds" context changes. I will be able to work on this fix.

RomanHotsiy:
Sorry to bother you @pieh... but this is a severe issue: build may produce wrong results without the user noticing it.

Could you at least let me know what is your plan on this? Won't you merge this PR without a complete fix? Should I start working on a complete fix?

Thanks in advance.

pieh:
> That's a more severe issue... I don't know how to fix it as part of this PR though. Persisting page data is a separate bigger PR.
> 
> I believe we can't assume context to be changed every time when `oldPage` is not there as it may hurt performance. Am I right?
> 
> I believe this PR can be merged and a separate issue should be open for "between-builds" context changes. I will be able to work on this fix.

Yeah, this is bigger issue. The fix might be tiny (i.e. just persist `pages` redux namespace) - but there is a lot of things to consider here, as right now Gatsby operates under assumption that when it bootstraps pages are empty.

> Could you at least let me know what is your plan on this? Won't you merge this PR without a complete fix? Should I start working on a complete fix?

I think it makes sense to merge this in general, but I didn't have yet time to thoroughly review and see if it can potentially cause some regressions.

RomanHotsiy:
After some using in production, I noticed a severe performance issue with my changes: instead of rerunning only specific page query it was rerunning all page queries of a specific component.
I now fixed that and force-pushed changes.

@pieh any chance to get it merged soon?

sidharthachatterjee:
Published in `gatsby@2.13.54`

PR #15405, opened 2019-07-04
feat(gatsby): add joinPath to utils

wardpeet:
## Description
Add joinPath to our utils api.

## Related Issues
https://github.com/gatsbyjs/gatsby/pull/15370

PR #15406, opened 2019-07-04
fix(gatsby-plugin-manifest): use content-digest helper from utils file

wardpeet:
## Description
While working on https://github.com/gatsbyjs/gatsby/pull/15405 I've noticed manifest plugin isn't importing from gatsby/utils but from the file directly.

## Related Issues
https://github.com/gatsbyjs/gatsby/pull/15405


wardpeet:
i'll be fixing this with @gatsbyjs/utils so closing

PR #15407, opened 2019-07-04
Use v4 instead of v1 uuids

ctavan:
## Description

I'm a co-author of the [uuid npm module](https://github.com/kelektiv/node-uuid) which is being used in some places within the code base of Gatsby.

I'm currently trying to understand real-world use cases of [time-based UUIDs ("v1 UUIDs")](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_1_(date-time_and_MAC_address)).

I found two occurrences where v1 UUIDs were being used instead of v4 UUIDs and I was wondering if any of these cases really require the semantically much more complex v1 UUIDs or whether we could live equally well with purely random v4 UUIDs?

At least the test suite still passes after my changes, so apparently this doesn't seem to introduce any known regressions.

I'd be really curious to understand the motivation for choosing v1 over v4 UUIDs in the first place, so any feedback on this would be highly appreciated!

## Related Issues

none

### DIFF
--- a/packages/gatsby-telemetry/src/telemetry.js
+++ b/packages/gatsby-telemetry/src/telemetry.js
@@ -1,5 +1,5 @@
 const { createHash } = require(`crypto`)
-const uuid = require(`uuid/v1`)
+const uuid = require(`uuid/v4`)
 const EventStorage = require(`./event-storage`)
 const { sanitizeErrors, cleanPaths } = require(`./error-helpers`)
 const ci = require(`ci-info`)
--- a/packages/gatsby-telemetry/src/telemetry.js
+++ b/packages/gatsby-telemetry/src/telemetry.js
@@ -1,5 +1,5 @@
 const { createHash } = require(`crypto`)
-const uuid = require(`uuid/v4`)
+const uuidv4 = require(`uuid/v4`)
 const EventStorage = require(`./event-storage`)
 const { sanitizeErrors, cleanPaths } = require(`./error-helpers`)
 const ci = require(`ci-info`)
@@ -18,7 +18,7 @@ module.exports = class AnalyticsTracker {
   osInfo // lazy
   trackingEnabled // lazy
   componentVersion
-  sessionId = uuid()
+  sessionId = uuidv4()
 
   constructor() {
     try {
@@ -177,7 +177,7 @@ module.exports = class AnalyticsTracker {
     }
     let machineId = this.store.getConfig(`telemetry.machineId`)
     if (!machineId) {
-      machineId = uuid()
+      machineId = uuidv4()
       this.store.updateConfig(`telemetry.machineId`, machineId)
     }
     this.machineId = machineId


packages/gatsby-telemetry/src/telemetry.js
```diff
@@ -18,7 +18,7 @@ module.exports = class AnalyticsTracker {
   osInfo // lazy
   trackingEnabled // lazy
   componentVersion
-  sessionId = uuid()
+  sessionId = uuidv4()
```
jamo: For sessionId we are actually relying on this behavior of the uuid-v1 

> Note: The id is generated guaranteed to stay constant for the lifetime of the current JS runtime. (Future versions of this module may use persistent storage mechanisms to extend this guarantee.)

It's mostly because the gatsby-telemetry is required from multiple different modules and thanks we might have multiple instances of the same module required. (or that's around the lines what we figured out a while ago)

packages/gatsby-telemetry/src/telemetry.js
```diff
@@ -177,7 +177,7 @@ module.exports = class AnalyticsTracker {
     }
     let machineId = this.store.getConfig(`telemetry.machineId`)
     if (!machineId) {
-      machineId = uuid()
+      machineId = uuidv4()
```
jamo: 👍

packages/gatsby-telemetry/src/telemetry.js
jamo: We could also store the sessionId in `process.gatsbySessionId` and use uuid_v4 for it, that way we'd get similar semantics as with the uuid_v1

jamo: so we could do it like this:
```
getSessionId() {
     return ( process.gatsbyTelemetrySessionId || (process.gatsbyTelemetrySessionId = uuid_v4())
}
```

ctavan: > For sessionId we are actually relying on this behavior of the uuid-v1
> 
> > Note: The id is generated guaranteed to stay constant for the lifetime of the current JS runtime. (Future versions of this module may use persistent storage mechanisms to extend this guarantee.)
> 
> It's mostly because the gatsby-telemetry is required from multiple different modules and thanks we might have multiple instances of the same module required. (or that's around the lines what we figured out a while ago)

I think there might be some confusion around the note that you are quoting here: The `uuid.v1()` method _does not_ produce only one stable UUID for the entire lifetime of the current JS runtime! If you compare with the [latest Readme of the `uuid` npm module](https://github.com/kelektiv/node-uuid#version-1) it states:

> Note: The default node id (the last 12 digits in the UUID) is generated once, randomly, on process startup, and then remains unchanged for the duration of the process.

As this note says _only the last 12 digits_ in the UUID remain constant. Since the first digits are derived from the current timestamp, every invocation of the `uuid.v1()` method will generate a different UUID where only the last 12 digits are guaranteed to remain the same. You can try it out in a node repl:

```
> const uuidv1 = require('uuid/v1');
undefined
> uuidv1()
'6a05b400-9ea2-11e9-a3e9-23116db2ce06'
> uuidv1()
'6a960d20-9ea2-11e9-a3e9-23116db2ce06'
> uuidv1()
'6af0fe60-9ea2-11e9-a3e9-23116db2ce06'
> uuidv1()
'6b40a500-9ea2-11e9-a3e9-23116db2ce06'
> uuidv1()
'6b7b3d00-9ea2-11e9-a3e9-23116db2ce06'
> uuidv1()
'6ba27410-9ea2-11e9-a3e9-23116db2ce06'
> uuidv1()
'6bc676d0-9ea2-11e9-a3e9-23116db2ce06'
> uuidv1()
'6be410f0-9ea2-11e9-a3e9-23116db2ce06'
```

I think that in the current code, if you import the `AnalyticsTracker` class in multiple places and create multiple instances if it, they should all have different `sessionId` properties!

A workaround would be to generate only one UUID in the module during invocation, something along the lines of:

```js
const processSessionId = uuidv4();

module.exports = class AnalyticsTracker {
  // … other initializers
  sessionId = processSessionId

  // … rest of class
}
```

If I understand the intended behavior right then I'm pretty confident that the `uuid.v1()` implementation does not behave like you expect and that we might have found an actual bug in the `gatsby-telemetry` package that would be worth investigating, WDYT?

jamo: uuh, nice catch.  sounds like we got confused from reading the description from https://www.npmjs.com/package/uuid#version-1

Though based on experiments and looking at the data this id does seem to behave as expected.

But we'll need to change how this is used.

jamo: the `const processSessionId` approach won't work as the module is required from both globally installed `gatsby-cli` and in local node_modules from `gatsby`
Hence we'll need to put it into the `process` global.

freiksenet:
Closing in favor of https://github.com/gatsbyjs/gatsby/pull/15431 (which is this PR with a patch applied). Thanks a lot for your help @ctavan!

jamo:
Thanks, @ctavan for clarifying the v1 behavior!
Super happy we were able to fix it before it caused any real issues. Guess it worked fine just because both uuidV1 IDs were generated almost at the same time.

packages/gatsby-telemetry/src/telemetry.js
ctavan: I see the readme on npm is outdated and I'm deeply sorry for the confusion that it caused. Will try to get an updated version released.

Thanks for getting back so quickly and for merging my changes!

PR #15408, opened 2019-07-04
fix(gatsby): restore client-only-paths behaviour

### LINKED ISSUE #15401: Gatsby@2.13.2 - Client-only routes broken
## Description

Client-only routes broken in Gatsby@2.13.2

### Steps to reproduce

Use starter project https://github.com/dwyfrequency/gatsby-ghub that is using the 'gatsby-plugin-create-client-paths' plugin and the latest version of Gatsby.

To check that version 2.13.2 is the problem, fix Gatsby version to 2.13.1 and check that it is now working as expected.

### Expected result

Client-only routes should work

### Actual result

Client-only routes not working

### Environment

System:
    OS: macOS 10.14.4
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 11.12.0 - ~/.nvm/versions/node/v11.12.0/bin/node
    Yarn: 1.15.2 - ~/.nvm/versions/node/v11.12.0/bin/yarn
    npm: 6.7.0 - ~/.nvm/versions/node/v11.12.0/bin/npm
  Languages:
  Browsers:
    Chrome: 75.0.3770.100
    Firefox: 67.0
    Safari: 12.1
  npmPackages:
    gatsby: ^2.10.0 => 2.13.3
    gatsby-image: ^2.0.20 => 2.2.4
    gatsby-plugin-create-client-paths: ^2.0.2 => 2.1.1
    gatsby-plugin-manifest: ^2.0.9 => 2.2.1
    gatsby-plugin-offline: ^2.0.16 => 2.2.0
    gatsby-plugin-playground: ^1.0.6 => 1.0.6
    gatsby-plugin-react-helmet: ^3.0.2 => 3.1.0
    gatsby-plugin-sharp: ^2.0.14 => 2.2.2
    gatsby-source-filesystem: ^2.0.8 => 2.1.2
    gatsby-source-graphql: ^2.1.0 => 2.1.0
    gatsby-transformer-sharp: ^2.1.8 => 2.2.1
  npmGlobalPackages:
    gatsby: 2.7.3

pieh:
Fixes regression introduced in https://github.com/gatsbyjs/gatsby/pull/13197

Fixes #15401

### DIFF
--- a/e2e-tests/development-runtime/src/pages/client-only-paths.js
+++ b/e2e-tests/development-runtime/src/pages/client-only-paths.js
@@ -13,15 +13,11 @@ const routes = [`/`, `/profile`, `/dashboard`]
 
 const basePath = `/client-only-paths`
 
-const ClientOnlyPathPage = props => (
+const ClientOnlyPathPage = () => (
   <Layout>
-    <Router
-      location={props.location}
-      basepath={basePath}
-      id="client-only-paths-sub-router"
-    >
-      <Page path="/" page="index" />
-      <Page path="/:page" />
+    <Router id="client-only-paths-sub-router">
+      <Page path="/client-only-paths/" page="index" />
+      <Page path="/client-only-paths/:page" />
     </Router>
     <ul>
       {routes.map(route => (
--- a/e2e-tests/development-runtime/src/pages/paths.js
+++ b/e2e-tests/development-runtime/src/pages/paths.js
@@ -7,7 +7,7 @@ const NotFound = () => <h1>Not Found in App</h1>
 
 const AppPage = () => (
   <Router>
-    <App path="/app" />
+    <App path="/paths/app" />
     <NotFound default />
   </Router>
 )
--- a/e2e-tests/production-runtime/src/pages/client-only-paths.js
+++ b/e2e-tests/production-runtime/src/pages/client-only-paths.js
@@ -13,15 +13,11 @@ const routes = [`/`, `/profile`, `/dashboard`]
 
 const basePath = `/client-only-paths`
 
-const ClientOnlyPathPage = props => (
+const ClientOnlyPathPage = () => (
   <Layout>
-    <Router
-      location={props.location}
-      basepath={basePath}
-      id="client-only-paths-sub-router"
-    >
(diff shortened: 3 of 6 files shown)


pieh:
After more thought - this doesn't fully restore previous behaviour:

In case there are no nested routers in user code (most common scenario) - it works as it worked previously 

In case there are nested router in user code there will be behaviour changes with this.

Consider scenario - user has `/app/*` matchPath, Inside it have `/profile` route, and inside it `/settings` route (so full path is `/app/profile/settings`. Before 2.13.2 this would be setup with:

```jsx
<Router>
  <Profile path="/app/profile/*">
    <Router>
      { /*
        Notice how this nested route doesn't need 
        /app or /app/profile in front of settings.
        In fact it wouldn't work if you tried to use that!
      */ }
      <ProfileSettings path="/settings" />
    </Router>
  </Profile>
<Router>
```

In 2.13.2 this would need to be 

```jsx
<Router>
  { /*
    In here you don't need /app in front of top-level router in user code
    this is relative to `/app/*` matchPath
  */ }
  <Profile path="/profile/*">
    <Router>
      <ProfileSettings path="/settings" />
    </Router>
  </Profile>
<Router>
```

With changes in this PR (setting basepath to path prefix by default) - this would be:

```jsx
<Router>
  <Profile path="/app/profile/*">
    <Router>
      { /*
        In your nested router you would also need to use full path
        */
      }
      <ProfileSettings path="/app/profile/settings" />
    </Router>
  </Profile>
<Router>
```

Imo behaviour before 2.13.2 is completely weird, but we can't make breaking changes like that even when it makes sense, so as alternative to this PR here are WIP changes that restore previous behaviour without reverting reach focus fixes: https://github.com/gatsbyjs/gatsby/commit/2a8dd19717653ac8fc64c68b4c0c09c288e9c405 - The thing that fixes the problem is setting `BaseContext.Provider` below our usage of Router (issue is that `@reach/router` doesn't export that ... so I did very hacky monkey-patching webpack loader that add this export for us)

pieh:
This PR is replaced by https://github.com/gatsbyjs/gatsby/pull/15457

Issue #15287, opened 2019-07-01
chore(gatsby-source-filesystem): Deprecate multiple different versions of birth time on File 


sidharthachatterjee:
@pieh and I noticed that we have three different versions of birth time in File 

- birthTime
- birthtime
- birthtimeMs

Probably want to clean this up at some point 😂

spences10:
What's the work involved in this @sidharthachatterjee?

KyleAMathews:
No one noticed before @pieh went and added that explorer sidebar!!! 😅

sidharthachatterjee:
@spences10 We'll probably want to 

- mark two of them (birthtime and birthtimeMs likely) as deprecated in GraphQL and show a warning when they are used
- update usage across this repo to `birthTime`

Simonpedro:
Hey there! I took a look at this and I decided to follow what @sidharthachatterjee mentioned.
I searched for usages of `birthtime` and `birthtimeMs` in the codebase but I only found them in types declarations files, so I've updated them too.

henricazottes:
We still have those on version `2.1.0`

PR #15409, opened 2019-07-04
chore(gatsby-source-filesystem): Deprecate fields

Simonpedro:
## Description

I've just added a schema customization option to deprecate redundant birth time field versions.

## Related Issues

Addresses https://github.com/gatsbyjs/gatsby/issues/15287

### DIFF
--- a/packages/gatsby-source-filesystem/src/gatsby-node.js
+++ b/packages/gatsby-source-filesystem/src/gatsby-node.js
@@ -183,3 +183,14 @@ See docs here - https://www.gatsbyjs.org/packages/gatsby-source-filesystem/
 }
 
 exports.setFieldsOnGraphQLNodeType = require(`./extend-file-node`)
+
+exports.createSchemaCustomization = ({ actions }) => {
+  const { createTypes } = actions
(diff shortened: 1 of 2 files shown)


packages/gatsby-source-filesystem/src/gatsby-node.js
```diff
@@ -183,3 +183,14 @@ See docs here - https://www.gatsbyjs.org/packages/gatsby-source-filesystem/
 }
 
 exports.setFieldsOnGraphQLNodeType = require(`./extend-file-node`)
+
+exports.createSchemaCustomization = ({ actions }) => {
+  const { createTypes } = actions
+  const typeDefs = `
+  type File implements Node {
```
freiksenet:
```suggestion
  type File implements Node @infer {
```

freiksenet: Explicit `@infer` is recommended so that we use inference behavior that will be standard in Gatsby v3.

freiksenet:
Thank you @Simonpedro!

Issue #15151, opened 2019-06-26
Built site fails when deployed to Amazon S3 in 2.9.0 and later


gswirrl:
###  Description

Site builds and serves successfully with `gatsby build && gatsby serve`, but when I copy the built files to an AWS S3 web bucket, the site generates 'resource missing' errors.

(the same AWS bucket has served gatsby sites successfully in the past)

UPDATE: I believe rolling back in same project to

```
  "dependencies": {
    "gatsby": "2.8.8",
    ...
```

fixes the issue. 2.9.0 and later all have the problem)

### Steps to reproduce

(1)
```
gatsby new demo
cd demo
gatsby build
```

(2)
upload files to AWS S3 website-bucket

### Expected result

Website should appear as it does from `gatsby serve`

### Actual result

The js does not run. Console shows errors:

<img width="1277" alt="Screenshot 2019-06-26 at 13 44 07" src="https://user-images.githubusercontent.com/17145418/60180755-a5371b80-9818-11e9-82ad-7c5c382644f6.png">

According to the 'Network' tab of my browser, all files loaded successfully.

### Environment

  System:
    OS: macOS 10.14.5
    CPU: (4) x64 Intel(R) Core(TM) m3-7Y32 CPU @ 1.10GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 12.4.0 - ~/.nvm/versions/node/v12.4.0/bin/node
    Yarn: 1.16.0 - /usr/local/bin/yarn
    npm: 6.9.0 - ~/.nvm/versions/node/v12.4.0/bin/npm
  Languages:
    Python: 2.7.10 - /usr/bin/python
  Browsers:
    Chrome: 75.0.3770.100
  npmPackages:
    gatsby: ^2.10.4 => 2.10.4 
    gatsby-image: ^2.2.3 => 2.2.3 
    gatsby-plugin-manifest: ^2.2.0 => 2.2.0 
    gatsby-plugin-offline: ^2.2.0 => 2.2.0 
    gatsby-plugin-react-helmet: ^3.1.0 => 3.1.0 
    gatsby-plugin-sharp: ^2.2.1 => 2.2.1 
    gatsby-source-filesystem: ^2.1.0 => 2.1.0 
    gatsby-transformer-sharp: ^2.2.0 => 2.2.0 
  npmGlobalPackages:
    gatsby-cli: 2.7.2

gswirrl:
Still fails in 2.11.0

[...]

PR #15410, opened 2019-07-04
fix(gatsby): parse pageData response when contentType is not set

wardpeet:
## Description
Not all hosting providers set a content-type of `application/json`. We only accept page-data that have that content-type. This PR will also parse responeText and look if it has a proper pageData object. If it does, it's considered json.

## Related Issues
https://github.com/gatsbyjs/gatsby/issues/15151

### DIFF
--- a/packages/gatsby/cache-dir/loader.js
+++ b/packages/gatsby/cache-dir/loader.js
@@ -34,16 +34,23 @@ const loadPageDataJson = loadObj => {
   const url = createPageDataUrl(pagePath)
   return doFetch(url).then(req => {
     const { status, responseText } = req
+    let isJson = false
 
-    const contentType = req.getResponseHeader(`content-type`)
-    const isJson = contentType && contentType.startsWith(`application/json`)
+    // Handle 200
+    if (status === 200) {
+      try {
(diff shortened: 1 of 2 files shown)


tahasabih:
Would be great if this can be merged and released! :) 
Had to downgrade our gatsby version on prod because it stopped working on s3 :( This seems to be a good fix!

packages/gatsby/cache-dir/loader.js
```diff
@@ -34,16 +34,23 @@ const loadPageDataJson = loadObj => {
   const url = createPageDataUrl(pagePath)
   return doFetch(url).then(req => {
     const { status, responseText } = req
+    let isJson = false
```
pieh: NIT: this is never set to anything else and is used later in condition in https://github.com/gatsbyjs/gatsby/blob/482e4829f670c6bb910e4610b4db4e23b14dde1d/packages/gatsby/cache-dir/loader.js#L57 (there are no other usage of `isJson`) - we could remove this completely, right?

wardpeet: that's true! thanks!

pieh:
Published `gatsby@2.13.6`

Issue #14835, opened 2019-06-17
gatsby-plugin-catch-links - allow opt-outs


gethinw:
## Summary

It would be great to have some options to configure this plugin, to ignore certain paths. In my case, anything under /examples/ is served as separate static files outside of Gatsby, so I would like to only convert links that don't start /examples/

### Basic example

```
{
      resolve: 'gatsby-plugin-catch-links',
      options: {
        excludePattern: /^\/examples\//
      }
}
```
### Motivation

To support sites with different types of routing setups than a very basic one. Although https://github.com/gatsbyjs/gatsby/issues/11238 was resolved in a different way, this could also have solved that.

sidharthachatterjee:
This sounds pretty useful! We’d love to accept a PR adding support for this 🙌🏼

fboechats:
Can I work on this feature??

sidharthachatterjee:
@fboechats Sure! Thank you for picking it up 🙌

i-bsd:
What about using multiple exclude patterns?

This seems to only allow one single exclude regex.

AnansiOmega:
Hey, first time working on open source is this feature still being worked on? If not I'd love take this on, with hopefully some guidance?

himanshu211raj:
I wanted to work on this issue but as i am a beginner so i need guidance regarding how i start contributing to open source?

PR #15416, opened 2019-07-04
feat(gatsby-plugin-catch-links): Add ability to exclude links

fboechats:
I come up with this idea to do this exclude path option to the catch links plugins. It's missing some things, but just want check what you guys think about it and some helping with the unit tests

refs #14835


packages/gatsby-plugin-catch-links/src/gatsby-browser.js
```diff
@@ -2,8 +2,8 @@ import { navigate } from "gatsby"
 
 import catchLinks from "./catch-links"
 
-exports.onClientEntry = () => {
-  catchLinks(window, href => {
```
sidharthachatterjee: @pieh Would you consider a signature change to this function a breaking change? Not really used by anyone besides this file but just wondering 🤔

sidharthachatterjee:
Tests should be good now

packages/gatsby-plugin-catch-links/src/gatsby-browser.js
sidharthachatterjee: Probably not an issue since this file isn't meant to be a consumed by others and is not a public API

pieh: No - this is internals change - `catchLinks` function is not documented as public API

fboechats:
Hey guys, no problem about the delay and sorry for not having some free time to do those minor changes, really appreciate to be part of the Gatsby environment. I will try to be more active in the community.

PR #15442, opened 2019-07-05
fix(gatsby): Exclude node_modules when depth > 1

sidharthachatterjee:


packages/gatsby/src/query/query-compiler.js
```diff
@@ -101,7 +101,8 @@ class Runner {
   }
 
   async parseEverything() {
-    const filesRegex = path.join(`/**`, `*.+(t|j)s?(x)`)
+    const filesRegex = `*.+(t|j)s?(x)`
+    const pathRegex = `/{${filesRegex},!(node_modules)/**/${filesRegex}}`
```
pieh: How about some context:
```suggestion
    // Pattern that will be appended to searched directories.
    // It will match any .js, .jsx, .ts, and .tsx files, that are not
    // inside <searched_directory>/node_modules.
    const pathRegex = `/{${filesRegex},!(node_modules)/**/${filesRegex}}`
```

sidharthachatterjee:
Published in `gatsby@2.13.7`

PR #15457, opened 2019-07-05
fix(gatsby): restore client-only-paths behaviour

### LINKED ISSUE #15401: Gatsby@2.13.2 - Client-only routes broken

pieh:
This PR is different attempt than https://github.com/gatsbyjs/gatsby/pull/15408 to fix regression introduced in https://github.com/gatsbyjs/gatsby/pull/13197

Here I expanded e2e tests to cover nested routers in user code. My previous PR would fail those. I also run them against 2.13.1 (last release before publish with breaking changes).

This is definitely messy and hacky (and I hate it), but I just don't see another way of preserving behaviour of Gatsby v2 and still having focus managment fixes in.

For context on how client-only-paths are broken right now please check https://github.com/gatsbyjs/gatsby/pull/15408#issuecomment-508600347 that goes over sample scenario and how this was handled before 2.13.2 (this PR restores that) and in 2.13.2

Fixes #15401


packages/gatsby/src/utils/webpack.config.js
```diff
@@ -259,6 +259,18 @@ module.exports = async (program, directory, suppliedStage) => {
       rules.images(),
       rules.media(),
       rules.miscAssets(),
+
+      // This is a hack that exports one of @reach/router internals (BaseContext)
+      // to export list. We need it to reset basepath and baseuri context after
+      // Gatsby main router changes it, to keep v2 behaviour.
+      // We will need to most likely remove this for v3.
+      {
+        test: /reach\/router\/es\/index/,
```
wardpeet: This supports windows & linux. If we want to test one file we can do it with require.resolve.
```suggestion
        test: require.resolve(`@reach/router/es/index`),
```

pieh: Oh yeah, that makes much more sense. Thanks!

xavivars: Quick question: why is this requiring `@reach/router/es/index` instead of `@reach/router/index`?

jakubzloczewski:
@pieh, I think Link should be imported from '@reach/router'. Please correct me if I'm wrong or missing something.

https://github.com/gatsbyjs/gatsby/blob/561d33e2e491d3971cb2a404eec9705a5a493602/e2e-tests/development-runtime/src/pages/client-only-paths.js#L3

PR #15463, opened 2019-07-06
fix(PnP): updates the test to account for the property name change

arcanis:
## Description

We've slightly changed the thrown errors for compatibility reasons. Since lots of thing specifically check for `MODULE_NOT_FOUND`, we're moving the semantic error codes into a different property name (`pnpCode`) as to not override `code`.

I'm not sure why you use this regex instead of checking for `MODULE_NOT_FOUND` btw. Might make sense to add a comment about it if you remember why 🤔

Ref https://github.com/yarnpkg/berry/pull/279

### DIFF
--- a/packages/gatsby/src/utils/test-require-error.js
+++ b/packages/gatsby/src/utils/test-require-error.js
@@ -3,7 +3,10 @@
 export default (moduleName, err) => {
   // PnP will return the following code when a require is allowed per the
   // dependency tree rules but the requested file doesn't exist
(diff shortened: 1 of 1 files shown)


pieh:
> I'm not sure why you use this regex instead of checking for `MODULE_NOT_FOUND` btw. Might make sense to add a comment about it if you remember why 🤔

This is a bit legacy code so I don't have full context on this, but I think we want to check if exact requested module is not found (and not any of module that given modules require) - that's why we do have `moduleName` as part of regex. It might be done using more structured error properties (not sure).

This also possibly means that PnP check isn't doing exactly what we want to do here, as we don't check `moduleName` in this code path, but it's better than not handling this at all I guess

pieh:
Published `gatsby@2.13.5`

Issue #12755, opened 2019-03-22
Wrong issue links in plugin changelogs


millette:
## Description

Links to issues appear wrong in many package changelogs. Gatsby's own is ok:
<https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/CHANGELOG.md>

But not so for the plugins, such as:
<https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-image/CHANGELOG.md>

Instead of markdown like:
```markdown
[#12085](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-feed/issues/12085)
```

The manual link could be removed in favor of a simple:
```markdown
#12085
```

### Expected result

Should link issues to right (top-level) URLs.

For example: <https://github.com/gatsbyjs/gatsby/issues/10278>

### Actual result

Links to URLs like <https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-image/issues/10278>

The string `tree/master/packages/gatsby-image/` shouldn't be there.

pieh:
It does seem like default preset for generating our changelogs doesn't handle this correctly. My best guess is that it works for `gatsby` package because in package.json there we point to root of repo:
https://github.com/gatsbyjs/gatsby/blob/8b4d5deb1b766806ec959c061b555ff4cfa08fa4/packages/gatsby/package.json#L171-L174

And in `gatsby-image` we point to directory:
https://github.com/gatsbyjs/gatsby/blob/8b4d5deb1b766806ec959c061b555ff4cfa08fa4/packages/gatsby-image/package.json#L29

It's probably time to fork changelog preset we use and do some customization

millette:
I searched a few times but have no idea where could help with a fix. Any pointers would be appreciated.

wardpeet:
thanks for checking this out @millette 

@pieh will we put the changelog fork in gatsby monorepo or make a separate one?

DSchau:
Bit of a doozy here. As @pieh notes, this works for the gatsby package because the repository URL is the base from which we construct our issues. For all others, the base is _wrong_ because that's not actually where issues are filed in a monorepo.

There are (minimally) two solutions here:

1. PR conventional-changelog to tweak the Handlebars template to account for this
    - Kind of a rabbit hole, but conventional-changelog-core is probably the best place to tweak this
[...]

PR #15477, opened 2019-07-07
chore: Bugifx update all package repository fields

Graham42:
## Description

Looking at some gatsby packages on npm they are missing any link to the github repo source so it was difficult to find where the source code was.

This PR fixes all the `repository` fields for packages and add a script to be able to do this easily again in the future. Or if we want to make a sweeping change can update the script instead of manually updating each package.

Something I'm not sure about: I had to update the lerna config to include the `themes` directory. I think this should be ok but not sure what else this might affect.

## Related Issues

Related: https://github.com/gatsbyjs/gatsby/issues/12755

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -116,6 +116,7 @@ jobs:
       - <<: *persist_cache
       - run: yarn lint:code
       - run: yarn lint:other
+      - run: yarn check-repo-fields
 
   unit_tests_node8:
     executor:
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
   "scripts": {
     "bootstrap": "npm-run-all -s check-versions lerna-prepare",
     "check-versions": "babel-node scripts/check-versions.js",
+    "check-repo-fields": "babel-node scripts/check-repo-fields.js",
     "format": "npm run format:code && npm run format:other && npm run format:svg",
     "format:other": "npm run prettier -- --write",
     "format:code": "npm run lint:code -- --fix",
--- a/packages/babel-plugin-remove-graphql-queries/package.json
+++ b/packages/babel-plugin-remove-graphql-queries/package.json
@@ -2,6 +2,12 @@
   "name": "babel-plugin-remove-graphql-queries",
   "version": "2.7.0",
   "author": "Jason Quense <monastic.panic@gmail.com>",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/gatsbyjs/gatsby.git",
+    "directory": "packages/babel-plugin-remove-graphql-queries"
+  },
+  "homepage": "https://github.com/gatsbyjs/gatsby/tree/master/packages/babel-plugin-remove-graphql-queries#readme",
   "devDependencies": {
     "@babel/cli": "^7.0.0",
     "@babel/core": "^7.0.0",
--- a/packages/babel-preset-gatsby-package/package.json
+++ b/packages/babel-preset-gatsby-package/package.json
@@ -2,6 +2,12 @@
   "name": "babel-preset-gatsby-package",
   "version": "0.2.0",
   "author": "Philipp Spiess <hello@philippspiess.com>",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/gatsbyjs/gatsby.git",
+    "directory": "packages/babel-preset-gatsby-package"
+  },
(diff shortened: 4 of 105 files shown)


Graham42:
bootstrap is failing because the check-versions script is failing
```
$ babel-node scripts/check-versions.js
gatsby-starter-blog-theme:
  Depends on "gatsby-theme-blog@^1.0.0"
  instead of "gatsby-theme-blog@0.0.11".

gatsby-starter-notes-theme:
  Depends on "gatsby-theme-notes@^1.0.0"
  instead of "gatsby-theme-notes@0.0.3".

gatsby-starter-theme:
  Depends on "gatsby-theme-blog@^1.0.0"
  instead of "gatsby-theme-blog@0.0.11".

  Depends on "gatsby-theme-notes@^1.0.0"
  instead of "gatsby-theme-notes@0.0.3".
```

It appears there are `1.0.0` versions of some packages published but that doesn't match the version in the source?

pieh:
> bootstrap is failing because the check-versions script is failing

Those are caused by changes you did to `package.json` -> `workspaces` and fact that we didn't update deps in `themes/*` - this is change that we will likely make, but it affect how themes and starters are handled, so we are not confident in making this change yet

Graham42:
Ok, I've dropped the commit that updated the package json `worskpaces` field.

wardpeet:
Thank you!

I'm wondering why the following syntax is not correct?
```
"repository": {
  "type": "git",
  "url": "https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-codemods"
}
```

Official npm docs say it's okay https://docs.npmjs.com/files/package.json#repository

Graham42:
Yes that would also be fine, I chose the style that most commonly already exists in this repo. I don't think it makes any significant difference.

Graham42:
Made a tweak to the script to not insert repository as the last key in package.json which makes merge conflicts more likely, and rebased on the master branch.

wardpeet:
npm has support for monorepos now so I would suggest moving to this syntax
```
"repository": {
  "type" : "git",
  "url" : "https://github.com/gatsbyjs/gatsby.git",
  "directory": "packages/gatsby"
}
```

Sorry for the back and forth

Graham42:
I tried that syntax and it doesn't work for the npm page link to GitHub. Until it does, I would recommend the current format. See my comment here: https://github.com/gatsbyjs/gatsby/pull/15477/commits/ea1d7d1be084d60ba4f15a276404d509295f289d#diff-5f64c3c255be83cf700e83e922506fa3R22

wardpeet:
Thank you! I didn't go through the code yet 🙊. PR looks great!

moonmeister:
So I've noticed that most of the change logs are broken. I think this is due to Lerna prefixing urls with the "repository" url. This means "tree/master/packages/..." get's added to every link to issues, prs, commits, etc...and breaks the url. I'm guessing the new format (https://github.com/gatsbyjs/gatsby/pull/15477#issuecomment-510106232) mentioned by @wardpeet would fix this, but I get it may have other drawbacks.

moonmeister:
Okay, did some more reading, I now have some thoughts:

@Graham42 How does the monorepo syntax not work for the nom page link? (my assumption is you mean it doesn't link you to the sub-directory) The more I think about this the repository link should be the root...and not include the folder path. NPM makes it clear that the 'repository" field is not for human consumption but for computers. 

> The URL should be a publicly available (perhaps read-only) url that can be handed directly to a VCS program without any modification. It should not be a url to an html project page that you put in your browser. It’s for computers.

The "homepage" field already exists in the `package.json` to get the user into the sub-directory. I think the repository url should be the new syntax. 

```
"repository": {
  "type" : "git",
  "url" : "https://github.com/gatsbyjs/gatsby.git",
  "directory": "packages/gatsby"
}
```

Let me know if you have any questions.

Graham42:
Correct, I was meaning the sub-directory. 

I missed the `homepage` field, that looks like a better fit, thanks! I'll update the script.

Graham42:
Updated script to use the monorepo format, and to also update the `homepage` field.

After having to update this PR multiple times I've realize that the repo would easily drift after this fix so I've refactored the script a bit to work as a check that runs as part of the circle ci lint stage.

Not sure if this might also fix https://github.com/gatsbyjs/gatsby/issues/12755 now?

moonmeister:
Yes, it would close #12755

PR #15492, opened 2019-07-07
chore(showcase): Added Cosmotory to showcase

Hyraze:


Hyraze:
> Remove www from url

Did it.Thanks cris

PR #15498, opened 2019-07-07
fix(www): Update filtering logic and derrive Open Source category (#15496 #15523)

### LINKED ISSUE #15496: [www] Filtering does not work as expected on showcase tab
## Description
When browsing for showcase projects, the filtering mechanism should work as an "AND" mechanism.  Currently, if you select multiple categories, the page with show you the UNION of the categories.  It makes more sense for the filtering mechanism to show the INTERSECTION of the categories when selecting multiple.

### Steps to reproduce
1. Navigate to https://www.gatsbyjs.org/showcase/
2. Select "API" from the categories
3. Select "Accessibility" from the categories

### Expected result
The number of results shown should be 13 or less (the number of sites with the category of "API").  Additionally, once the "API" category is selected, the available categories should also reflect the INTERSECTION with the selected category(s).
What should happen?

### Actual result
The number of results increases from 13 to 19 because that is the number of sites which have "API" OR "Accessibility"

jd-carroll:
## Description
Move the filtering logic on the showcase page to be the intersection of categories instead of the union of categories

## Related Issues
Fixes #15496 #15523


jd-carroll:
Note: Both failures are due to the linting issues. I am happy to update the file to address the linting issues, but those changes would go beyond the scope of #15496

jd-carroll:
@lannonbr Is the title of this PR correct or should it be updated?  Specifically, should it be `fix(www)` or `fix(showcase)`?

lannonbr:
In this usecase I would say fix(www) is fine given it is updating the website code. usually when we use showcase in the title it is chore(showcase) for adding or removing entries from the showcase

jd-carroll:
I'm running into an issue getting these features tested, any time I run `npm start` from the `www` directory it fails to snapshot ~250 sites.  This ultimately leads to a failure of `npm start`.  Any ideas on how to get his working? (Is `npm start` the right way to launch a local dev server?)

Example:
```
Failed to screenshot http://capricious-spring.surge.sh/. Retrying...
Failed to screenshot https://gatsby-starter-bloomer.netlify.com/. Retrying...
Failed to screenshot https://gatsby-starter-business.netlify.com/. Retrying...
...
```

jd-carroll:
Any thoughts or opinions on this PR, happy to have a discussion

fk:
Hey @jd-carroll! 👋 
I was the one responsible for making those filters work the way they do right now. TBH I can not reproduce the reasoning anymore, and I follow your arguments for changing the behavior.

Opinions @lannonbr @gillkyle @freiksenet @sidharthachatterjee?

fk:
cc @shannonbux

gillkyle:
Yeah I like the idea of it being an intersection. That seems like how I'd expect the filters to work in most cases.

www/src/views/showcase/filtered-showcase.js
```diff
@@ -72,31 +80,37 @@ class FilteredShowcase extends Component {
...
+    }, {})
 
     // get sorted set of categories to generate list with
-    const categoryKeys = Object.keys(aggregatedCategories).sort((a, b) => {
-      if (a < b) return -1
-      if (a > b) return 1
-      return 0
-    })
+    const categoryKeys = Object.keys(aggregatedCategories).sort((str1, str2) =>
+      str1.toLowerCase().localeCompare(str2.toLowerCase())
```
gillkyle: Didn't know about the `localeCompare` method, nice!

www/src/views/showcase/filtered-showcase.js
```diff
@@ -72,31 +80,37 @@ class FilteredShowcase extends Component {
...
+    const aggregatedCategories = items.reduce((categories, edge) => {
+      if (!edge.node.categories) {
+        edge.node.categories = []
+      }
+      const idx = edge.node.categories.indexOf(OPEN_SOURCE_CATEGORY)
+      if (idx !== -1) {
+        edge.node.categories.splice(idx, 1)
+      }
+      if (edge.node.source_url) {
+        edge.node.categories.push(OPEN_SOURCE_CATEGORY)
```
gillkyle: This is also a nice touch, adding any sites with a public repository to the Open Source category, seems like a good idea to me!

shannonbux: oh yeah, great idea!

shannonbux:
@fk I agree with you, this argument for changing the behavior makes sense and is something we intended to change so thank you @jd-carroll for doing it! I just tested it on the live site and it's looking like it's working!

fk:
❤️ @jd-carroll and everyone! 🙏

Issue #15272, opened 2019-07-01
Measure key performance numbers in CI (and via telemetry when makes sense)


KyleAMathews:
Gatsby stakes its reputation on being fast.

We should measure closely key performance numbers so we don't introduce accidental regressions (and so we can know when various improvements are working).

We should measure these in CI and fail builds that slow things down. We should also track bootstrap numbers in telemetry so we can see how things look in the wild.

- develop & build bootstrap up to the first API invocation to measure our startup time
- individual activity timing during build for hello-world starter
- how big is the JS bundles for the front page of the hello world starter

jamo:
Uuh, we've captured the two first points a while back. I should look more into the last one.
Thanks @gatsbybot 😉

PR #15503, opened 2019-07-08
fix(gatsby): speed up and lower memory of createPages

me4502:
## Description

I was noticing very slow createPages times with 75k pages, so I started profiling and found that it was allocating 18GB(!!!) of RAM, and spending a significant amount of time garbage collecting.

I tracked it down to the line modified in this PR, and replaced it with a push rather than a concat. The places it is used don't appear to care if the initial array is modified or not, however whether this is guaranteed I'm unsure.

Tests appear to still work for me, as does the site with 75k pages.

This took RAM from 18GB to 1.5GB, and createPages times from 240-300 to 100-150.

This is one of two sections that use ridiculous amounts of RAM at large page counts. The other is the creation of bundles and HTML pages, however I haven't had a chance to look into this yet. They easily use ~25GB of RAM in 75k pages.

### DIFF
--- a/packages/gatsby/src/redux/machines/page-component.js
+++ b/packages/gatsby/src/redux/machines/page-component.js
@@ -105,8 +105,8 @@ module.exports = Machine(
                 queryUtil.runQueuedQueries(event.path)
               }
             }, 0)
-
-            return ctx.pages.concat(event.path)
+            ctx.pages.push(event.path)
+            return ctx.pages
           } else {
             return ctx.pages
           }


pieh:
Common approach is to use immutable structures (like we have right now) for cases like this, but in fact we don't actually benefit from it at all as we don't really do things like time-travel and this is not tied to UI (re-render on props changes) so it should be safe to apply this change.

The only reason AFAIR is that we use list of pages to determine if new page was created or not to potentially trigger query running (in https://github.com/gatsbyjs/gatsby/blob/811ee778d356c8ce4103101bc4a942d0910f6ef5/packages/gatsby/src/redux/reducers/components.js#L41-L43 )

pieh:
Published `gatsby@2.13.8`

KyleAMathews:
Thanks for the fix! This is a regression that happened sometime in the last few months as I can't recreate this on older versions of https://github.com/gatsbyjs/gatsby/tree/master/benchmarks/create-pages (just tested for five minutes so didn't track down where the regression happened).

More reason to get automatic perf tests in CI https://github.com/gatsbyjs/gatsby/issues/15272

BTW, to make perf issues easy to reproduce, we have benchmark sites https://github.com/gatsbyjs/gatsby/tree/master/benchmarks/create-pages

I noticed also that you mentioned in Discord that Ink is slowing things down a lot. I validated that with the createPages benchmark on the latest release. We should throttle updates to INK to maybe 30 FPS /cc @pieh @wardpeet

KyleAMathews:
Oh on my validation — the old version of Gatsby took ~12 seconds to create pages for 75k pages. Before this PR it OOMed. With this PR, it took 37 seconds so 3x still the old time.

nakleiderer:
I just wanted to drop a quick thank you to everyone involved with this fix! My website stopped building in CI after I updated Gatsby and I wasn't able to reproduce on my local machine. This fix was very timely for me :) keep up the great work!

PR #15510, opened 2019-07-08
 feat(gatsby-core-utils): add url resolve

wardpeet:
## Description
2 things happen here:
1) I renamed the gatsby-utils folder to match the gatsby-core-utils package name
2) added urlResolve util as joinPath is used to store windows path as a string and not to join paths with '/'.

The reason for not using `path.posix.join` is that it doesn't covert `\` into `/`

## Related Issues
https://github.com/gatsbyjs/gatsby/pull/15370


sidharthachatterjee:
@wardpeet Linting issue 

[gatsby-core-utils] repository and/or homepage field in package.json are out of date.Run "yarn run check-repo-fields -- --fix" to update.

PR #15533, opened 2019-07-08
fix(gatsby): use Set instead of Array in page-component state machine

pieh:
Using https://github.com/gatsbyjs/gatsby/tree/master/benchmarks/create-pages benchmark with 75000 pages

before:
Screenshot 2019-07-08 at 20 59 29

after:
Screenshot 2019-07-08 at 21 11 25

Seems like pretty quick low hanging perf win.

Ref https://github.com/gatsbyjs/gatsby/pull/15503#issuecomment-509300935


KyleAMathews:
@pieh does this match perf numbers for earlier versions of Gatsby v2?

KyleAMathews:
Hmmm trying this locally with gatsby-dev-cli and it's freezing now on the create-pages benchmark with 75k pages.

me4502:
Oddly enough, on the site I work on, I can't see too significant a difference with this. It's within margin of error, at least.

From looking at the change, it should impact any page insertion that uses a template that has been used before, which is 99% of page creations on this site.

It should in theory make a massive difference, I'm unsure why I'm not seeing much of a change.

pieh:
> Hmmm trying this locally with gatsby-dev-cli and it's freezing now on the create-pages benchmark with 75k pages.

Ok, this didn't happen here - but it is worrying - will go over this code again. One thing that might happen which could cause problems is if gatsby didn't invalidate .cache directory - it would load old array like persisted data - but then it should crash and not freeze I think

me4502:
I've been running this for about 2 weeks now (actually improves things a lot after I disabled Loki), and have yet to run into the freezing issue. I do regularly clean my Gatsby cache though, could it potentially be a caching issue?

One "gotcha" that I've ran into before regarding sets, is that they aren't serialised correctly (Serialises as "{}"). I'm not too familiar with redux internals, but does it ever serialise to string? It could be related to that.

DanielRuf:
Actually it depends on the v8 version and a few other things.
You should also mention that `Set` keeps only unique names so any additional filtering to get unique values is not needed with this anymore.

pieh:
> One "gotcha" that I've ran into before regarding sets, is that they aren't serialised correctly (Serialises as "{}"). I'm not too familiar with redux internals, but does it ever serialise to string? It could be related to that.

We use `v8.serialize`, which does correctly serialize/deserialize Maps and Sets for persisting state. We were using JSON.stringify before as fallback for node 6 (where v8.serialize was not available), but we did drop node 6 support recently, so now v8.serialize should be used always.

wardpeet:
@me4502 what node version are you running?

me4502:
I'm on v11.15.0

wardpeet:
v11 might be a bit buggy as it's not an LTS release. Could you see if node 12 or node 10 also causes the `{}` bug?

me4502:
`JSON.stringify(new Set([1,2,3]))` returns '{}' on Node 10.16.0, 11.15.0, and 12.7.0

However, as JSON.stringify serialisation isn't used anywhere this PR touches it shouldn't matter.

KyleAMathews:
Tested this again and it didn't freeze this time 😬 

Anyways, validated that with 50k pages on the create-pages benchmark, this PR is 10 seconds faster for creating pages — 21s on master and 11s on this PR. So seems good to go 👍

wardpeet:
Successfully published:
 - gatsby@2.13.46

PR #15534, opened 2019-07-08
Remove /app for path route

ipatate:
Related to #15532


pieh:
@ipatate we will be fixing issue that introduced the need for this change (it's considered a bug because it breaks v2 behaviour)

ipatate:
Ok, i close the PR.

ipatate:
@pieh i tried the version 2.13.10, it's ok now. That work fine.

pieh:
Awesome! Thanks for update on that @ipatate!

Issue #14755, opened 2019-06-13
[docs][workflows]: add more info & controls for Working with GIFs


marcysutton:
## Summary

Continuing on from https://github.com/gatsbyjs/gatsby/issues/14529, the new page on [Working with GIFs](https://www.gatsbyjs.org/docs/working-with-gifs/) has two improvements to be made:

1. Add controls to pause the animating GIF on page load to not pose accessibility problems for users with motion sensitivity, perhaps with [react-gif-player](https://www.npmjs.com/package/react-gif-player);
2. Explain the phrase "beware that flashing and autoplaying GIFs can cause accessibility problems" with a bit more detail. We could certainly mention it violates WCAG 2.2.2 for "pause, stop, hide". Some resources for reference material: 
    - https://source.opennews.org/articles/motion-sick/
    - https://www.w3.org/TR/UNDERSTANDING-WCAG20/time-limits-pause.html

We need to do this as part of our own commitment to accessibility in Gatsby, but we also want to take every opportunity to teach best practices–and this is a best practice for working with animation and motion on the web.

Issue #14095, opened 2019-05-17
[docs] Define custom MDX components needed for the docs


jlengstorf:
As soon as gatsbyjs/gatsby#14071 is merged, we’ll be able to add custom MDX shortcodes to improve the docs authoring experience and the readability of the docs.

@marcysutton is collecting requirements for what we need to create — this issue tracks that work and will eventually evolve into an umbrella issue tracking progress on the shortcodes.

marcysutton:
The components were spec'd out in a separate doc: https://docs.google.com/document/d/1XYzToB-cNJ6Ot0ujdg5WaO5jmKH7moJ8305VwwwA2UY/edit#heading=h.nt3t0p9y6kx5

We need this project to be prioritized to move onto the next stage. Some of the components will be addressed with issues like #13475 and #6896, which I'll follow-up on, but as for the spec'ing part, I believe it is done and the issue can be closed.

PR #15539, opened 2019-07-08
fix: remove motion from GIFs page

marcysutton:
I looked into adding `react-gif-player` to gatsbyjs.org, but it would need to be added as a client-only script and I didn't have time to work through that...so I changed the animated GIF to show only one frame instead.

Closes https://github.com/gatsbyjs/gatsby/issues/14755


marcysutton:
Thanks, @amberleyromo!

As for future work around a GIF playback component, we'll tackle that as part of https://github.com/gatsbyjs/gatsby/issues/14095 later in Q3. So it isn't being dropped entirely.

PR #15557, opened 2019-07-09
Update HREF for "tag function"

devolasvegas:
Updated the link to take user directly to the section on the page describing tagged templates, which is more pertinent in this particular context.


devolasvegas:
✌😃

Issue #15279, opened 2019-07-01
[feat]: pass data from webhook to sourceNodes function


DSchau:
## Summary

Gatsby Preview invokes [sourceNodes](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/commands/develop.js#L186) kinda blindly from the webhook.

To make authoring _some_ source plugins easier, we could directly pass the webhook data from the CMS -> Gatsby Preview -> Gatsby, which oftentimes contains the incremental data that is needed to refresh Gatsby Preview content at the plugin level.

### Basic example

We should pass `webhookData` that is the `req.body` (as JSON) from the POST invocation. This will contain the entire payload from the webhook request from the CMS.

### Motivation

This unblocks even more source plugins from being able to refresh content in a Gatsby Preview environment, presuming they pass the differential data to the webhook (many of which do, notably CosmicJS).

PR #15564, opened 2019-07-09
feat(gatsby): pass webhook data to sourceNodes functions

DSchau:
## Description

Introduces functionality to pass a property `webhookBody` to source plugins that are Gatsby Cloud aware.

This can be used to fetch remote content with the “diff” that could be passed to the webhook from the vendor/CMS.

I’m happy to bikeshed on the name of the parameter (e.g. webhookData, webhookBody, etc.) so let me know what people prefer here.

cc @cosmicjs

## Related Issues

Implements #15279


DSchau:
Will iron out the failing test!

DSchau:
One sec — will fix the weird merges.

DSchau:
@pieh presuming the e2e tests pass — this should be ready for another look and merge!

For context — this is a change worth doing, but longer term, we may want to offer up a separate endpoint per _source_ plugin (e.g. `/__refresh/gatsby-source-cosmicjs`) which will only invoke create/edit/delete nodes for that particular source plugin.

packages/gatsby/src/commands/develop.js
```diff
@@ -131,6 +131,7 @@ async function startServer(program) {
     })
   )
 
+  app.use(express.json())
```
pieh: Could we perhaps apply this middleware only to refresh endpoint?

pieh:
If this test passes locally, but not in CircleCi I suspect there is timing problem (and I think I saw updating tests being flaky sometimes). How do you feel about adding more timeout to assertions (I think default cypress one is 3s)?

DSchau:
OK — besides the failing tests (which probably need master merged?) this is ready to go!

DSchau:
@gatsbyjs/core this is ready for another look!

packages/gatsby/src/utils/source-nodes.js
```diff
@@ -31,11 +31,12 @@ function discoverPluginsWithoutNodes(storeState) {
   return _.difference(nodeCreationPlugins, nodeOwners)
 }
 
-module.exports = async ({ parentSpan } = {}) => {
+module.exports = async ({ webhookBody = {}, parentSpan } = {}) => {
```
pieh: I have question about setting empty object as default - I wonder if it would be better to not do that - as user have to do more checks than just looking if `webhookBody` exists. Also body could technically not be plain object, but an array (or whatever) - so assumption that it will be object might be more problematic in those cases as well?

DSchau: For sure. I was thinking of the case where e.g. `if (webhookBody.updates)` or something, and didn’t want to _have_ to guard against null. In other words, I think it’s likely this would be a nested structure passed from the CMS, so they’d check for a nested property instead of the existence of the object itself.

I’m happy to do either though.

pieh: I would not set default, but this is not the hill I will die on ;) I leave that decision to You, just wanted to showcase edge cases.

DSchau:
@pieh you’re the best :) I’m gonna merge this as-is, but as no one is using this yet, we can follow-up as needed.

PR #15572, opened 2019-07-09
chore(scripts): clean up CI scripts for starters

sidharthachatterjee:
Cleans up a lot of https://github.com/gatsbyjs/gatsby/pull/15323


package.json
```diff
@@ -105,6 +105,7 @@
     "watch": "lerna run watch --no-sort --stream --concurrency 999"
   },
   "workspaces": [
-    "packages/*"
+    "packages/*",
+    "themes/gatsby-theme-*"
```
sidharthachatterjee: Is adding a folder to multiple workspaces defined behaviour? 

In this case `themes/gatsby-theme-*` is in both the root level workspace here and the ones in `themes/`

@arcanis Thoughts?

arcanis: We discourage nested workspaces (so to make `themes/` contain its own workspaces) because it isn't officially supported in the v1. It cause problems where running `yarn install` inside `themes/gatsby-theme-foo` will run an install in `themes/` but *not* at the root of the repository as you'd expect. Similarly, running `yarn add` in `themes/gatsby-theme-foo` will not update the root lockfile.

This behavior will change in the v2, and nested workspaces will be properly supported (the lockfile will only exist at the root, no longer in `themes/`).

sidharthachatterjee: Ah! Best wait for v2 then. Thank you so much @arcanis 🙌

renovate.json
```diff
@@ -9,5 +9,5 @@
   "rebaseStalePrs": false,
   "rangeStrategy": "bump",
   "bumpVersion": null,
-  "semanticCommitScope": "starters"
+  "semanticCommitScope": null
```
sidharthachatterjee: @pieh This is because PRs updating dependencies on theme packages at the moment still say `chore(starters): blah` 🤦‍♂ 

With this all PRs from Renovate will say `chore: blah`

PR #15574, opened 2019-07-09
feat(gatsby-plugin-google-gtag): Add link preconnect for Google Analytics

sheepsteak:
## Description

`gatsby-plugin-google-analytics` currently has a `link` tag added with a `preload` for the Google Analytics domain as it's recommended by Lighthouse (see [here](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-google-analytics/src/gatsby-ssr.js#L26-L32)).

This PR adds the same feature to `gatsby-plugin-google-gtag`.


packages/gatsby-plugin-google-gtag/src/__tests__/gatsby-ssr.js
```diff
@@ -15,8 +16,31 @@ it(`does not crash when no pluginConfig is provided`, () => {
 const DO_NOT_TRACK_STRING = `!(navigator.doNotTrack == "1" || window.doNotTrack == "1")`
 
 describe(`respectDNT`, () => {
+  it(`adds a preconnect link for Google Analytics`, () => {
```
wardpeet: could you move this test out of respectDNT describe? We're missing an root describe here but for now moving the `it` statement to the root is fine.

sheepsteak: Done!

PR #15592, opened 2019-07-10
Add support for preloaded global critical styles

jcayzac:
## Description
This allows injecting critical style without inlining (antipattern in HTTP/2).

```js
// gatsby-browser.js
import(/* webpackPreload: true */ `./src/styles/index.sass`)
```

The above generates a `<link as="style" rel="preload" href="/styles.9b011d582aac2f4be01f.css"/>` tag in
each page's `<head>`.

## Related Issues

- https://github.com/gatsbyjs/gatsby/commit/981bc8ca0aa726f9459cef99168c18cc2173740b
- #9828 gatsby-plugin-netlify v2 excludes some link headers
- #12521 PR adding links to `_headers`, for Netlify. Only preloads bundles.
- `from: jc#3330 static-entry.js` search on Discord.

## Caveats

`gatsby-plugin-netlify` doesn't parse `.namedChunkGroups.app.childAssets.preload` in the webpack stats, and simply ignore all assets marked for preloading, so this commit still doesn't enable the expected HTTP/2 behavior on Netlify.

### DIFF
--- a/examples/using-preloaded-global-styles/.eslintrc.json
+++ b/examples/using-preloaded-global-styles/.eslintrc.json
@@ -1,0 +1,8 @@
+{
+  "env": {
+    "browser": true
+  },
+  "globals": {
+    "graphql": false
+  }
+}
--- a/examples/using-preloaded-global-styles/README.md
+++ b/examples/using-preloaded-global-styles/README.md
@@ -1,0 +1,7 @@
+# Using Preloading Global Critical Styles
+
+Demonstrates how to use `webpackPreloaded` to apply a critical style globally.
+
+## References
+
+- [Prefetching/Preloading Modules](https://webpack.js.org/guides/code-splitting/#prefetching-preloading-modules).
--- a/examples/using-preloaded-global-styles/gatsby-browser.js
+++ b/examples/using-preloaded-global-styles/gatsby-browser.js
@@ -1,0 +1,1 @@
+import(/* webpackPreload: true */ `./src/styles/index.sass`)
--- a/examples/using-preloaded-global-styles/gatsby-config.js
+++ b/examples/using-preloaded-global-styles/gatsby-config.js
@@ -1,0 +1,6 @@
+module.exports = {
+  siteMetadata: {
+    title: `Gatsby with global style`,
+  },
+  plugins: [`gatsby-plugin-sass`],
+}
--- a/examples/using-preloaded-global-styles/package.json
+++ b/examples/using-preloaded-global-styles/package.json
@@ -1,0 +1,24 @@
+{
+  "name": "gatsby-example-using-preloaded-global-critical-styles",
+  "private": true,
+  "description": "Gatsby example site using preloaded global critical style",
+  "version": "2.0.0",
+  "author": "Julien Cayzac <julien.cayzac@gmail.com>",
+  "dependencies": {
+    "gatsby": "next",
+    "gatsby-plugin-sass": "^2.1.0",
+    "node-sass": "^4.12.0",
+    "react": "^16.8.6",
+    "react-dom": "^16.8.6"
+  },
+  "keywords": [
+    "gatsby"
+  ],
+  "license": "MIT",
+  "main": "n/a",
+  "scripts": {
+    "develop": "gatsby develop",
+    "build": "gatsby build",
+    "start": "npm run develop"
+  }
+}
--- a/examples/using-preloaded-global-styles/src/pages/index.jsx
+++ b/examples/using-preloaded-global-styles/src/pages/index.jsx
@@ -1,0 +1,5 @@
+import React from "react"
+
+const Index = () => <div>Hello</div>
+
+export default Index
--- a/examples/using-preloaded-global-styles/src/styles/index.sass
+++ b/examples/using-preloaded-global-styles/src/styles/index.sass
@@ -1,0 +1,4 @@
+html
+	background-color: rebeccapurple
+	color: white
+
--- a/packages/gatsby/cache-dir/static-entry.js
+++ b/packages/gatsby/cache-dir/static-entry.js
@@ -314,10 +314,10 @@ export default (pagePath, callback) => {
     .slice(0)
     .reverse()
     .forEach(style => {
-      // Add <link>s for styles that should be prefetched
+      // Add <link>s for styles that should be prefetched or preloaded
       // otherwise, inline as a <style> tag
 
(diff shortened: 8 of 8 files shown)


wardpeet:
Hey! Thank you for your PR. The code looks great 👍.

Do you have some blog posts or extra information about "antipattern in HTTP/2"? Could you have a look at e2e tests and also update them with our changes so we make sure we test your changes.

jcayzac:
@wardpeet a good start would be https://developers.google.com/web/fundamentals/performance/http2/#server_push

Inlining CSS basically prevents the client from declining it, should it already have it in its cache (which it doesn’t because it’s not in a separate file). You’re just pushing the same resource over and over like you would have done in HTTP/1.1 days.

I’ll have a look at the tests.

pieh:
> Inlining CSS basically prevents the client from declining it, should it already have it in its cache (which it doesn’t because it’s not in a separate file). You’re just pushing the same resource over and over like you would have done in HTTP/1.1 days.

True, but worth noting that html file is pushed just for first page as afterwards navigation happens on the client. Granted - if user open new tab / refresh they will get same inlined css again.

We also can't make assumptions if user will configure http/2 push, and if they don't - inlining usually is better idea.

From looking at failed tests it seems like styles are not being loaded at all for some reason (e2e_tests_production_runtime failing) and seems that this change affects styles that are not imported with `/* webpackPreload: true */` hint (e2e_tests_path-prefix failing)

jcayzac:
> True, but worth noting that html file is pushed just for first page as afterwards navigation happens on the client.

This is only true for clients with Javascript enabled. Other clients get a new full page with the same inlined CSS.

> We also can't make assumptions if user will configure http/2 push, and if they don't - inlining usually is better idea.

If you’re talking about end users, I’m not sure I understand. HTTP/2 is a 5-year old standard, the current version of the HTTP protocol and HTTP/3 is to be released later this year. HTTP/2 server push is enabled by default in all modern browsers. Users advanced enough to find the flag to disable it probably know about the performance degradation it will lead to everywhere.

If talking about Gatsby users, I think explicitly having to write the `webpackPreload` bit is a pretty good opt-in / no assumption made mechanism. It also follows the same behavior already existing for `webpackPrefetch` (which also disables inlining, but sets `rel` to `prefetch` which is useless for critical CSS).

pieh:
> This is only true for clients with Javascript enabled. Other clients get a new full page with the same inlined CSS.

True.

> If you’re talking about end users, I’m not sure I understand. HTTP/2 is a 5-year old standard, the current version of the HTTP protocol and HTTP/3 is to be released later this year. HTTP/2 server push is enabled by default in all modern browsers. Users advanced enough to find the flag to disable it probably know about the performance degradation it will lead to everywhere.

Don't you still need to configure your server to use push/let it know what to push for which pages? I found `http2_push_preload` is config option in nginx, but not sure how common is this used (it doesn't seem to be default) - do you have more insight into this matter?

> If talking about Gatsby users, I think explicitly having to write the `webpackPreload` bit is a pretty good opt-in / no assumption made mechanism. It also follows the same behavior already existing for `webpackPrefetch` (which also disables inlining, but sets `rel` to `prefetch` which is useless for critical CSS).

Agreed - but just looking at tests, it doesn't seem to work as expected (at least in current implementation). If we can make it behave like that when user opts in - I'm on board for this feature.

jcayzac:
> Don't you still need to configure your server to use push/let it know what to push for which pages? I found http2_push_preload is config option in nginx, but not sure how common is this used (it doesn't seem to be default) - do you have more insight into this matter?

This is certainly true. Now, this is not required if you're deploying to a provider with builtin server push support, and they're becoming more common. Netlify for instance.

> just looking at tests, it doesn't seem to work as expected (at least in current implementation). If we can make it behave like that when user opts in - I'm on board for this feature.

Yes, I still need to have a look at the test. Hopefully over the weekend.

jcayzac:
All the tests fail, on my system (with or without this commit).
Do you have a Docker image?

jcayzac:
It seems there's a dependency on `yarn global` somewhere. It's trying to create the global yarn prefix (`/nope`), which is forbidden on my system (global modules use npm -g, yarn is only for projects).

```sh
jcayzac ~/github/gatsby/e2e-tests/path-prefix [master * u= origin/master] gatsby-dev --set-path-to-repo ../../
jcayzac ~/github/gatsby/e2e-tests/path-prefix [master * u= origin/master] gatsby-dev --scan-once --quiet
Copied 142 files
jcayzac ~/github/gatsby/e2e-tests/path-prefix [master * u= origin/master] chmod +x ./node_modules/.bin/gatsby
jcayzac ~/github/gatsby/e2e-tests/path-prefix [master * u= origin/master] yarn test
yarn run v1.16.0
$ CYPRESS_SUPPORT=y npm run build && npm run start-server-and-test
Error: EACCES: permission denied, mkdir '/nope'
TypeError: Cannot read property 'loaded' of undefined
    at exit (/Users/jcayzac/.nvm/versions/node/v12.6.0/lib/node_modules/npm/lib/utils/error-handler.js:98:27)
    at errorHandler (/Users/jcayzac/.nvm/versions/node/v12.6.0/lib/node_modules/npm/lib/utils/error-handler.js:216:3)
    at /Users/jcayzac/.nvm/versions/node/v12.6.0/lib/node_modules/npm/bin/npm-cli.js:77:20
    at cb (/Users/jcayzac/.nvm/versions/node/v12.6.0/lib/node_modules/npm/lib/npm.js:225:22)
    at /Users/jcayzac/.nvm/versions/node/v12.6.0/lib/node_modules/npm/lib/npm.js:263:24
    at /Users/jcayzac/.nvm/versions/node/v12.6.0/lib/node_modules/npm/lib/config/core.js:83:7
    at Array.forEach (<anonymous>)
    at /Users/jcayzac/.nvm/versions/node/v12.6.0/lib/node_modules/npm/lib/config/core.js:82:13
    at f (/Users/jcayzac/.nvm/versions/node/v12.6.0/lib/node_modules/npm/node_modules/once/once.js:25:25)
    at afterExtras (/Users/jcayzac/.nvm/versions/node/v12.6.0/lib/node_modules/npm/lib/config/core.js:173:20)
/Users/jcayzac/.nvm/versions/node/v12.6.0/lib/node_modules/npm/lib/utils/error-handler.js:98
  var doExit = npm.config.loaded ? npm.config.get('_exit') : true
                          ^

TypeError: Cannot read property 'loaded' of undefined
error Command failed with exit code 7.
```

Issue #11736, opened 2019-02-13
gatsby-transformer-xml uses unmaintained dependencies to parse xml


laurenskling:
## Summary

`gatsby-transformer-xml` uses https://github.com/segmentio/xml-parser to parse XML, but this package is not being maintained. 
For me, it's lacking CDATA support, but without a maintainer there is no fix for me to get this working in Gatsby :(
What can we do about it?

### File contents (if changed)

`gatsby-config.js`: N/A 
`package.json`: N/A 
`gatsby-node.js`: N/A 
`gatsby-browser.js`: N/A 
`gatsby-ssr.js`: N/A

jonniebigodes:
@laurenskling thank you for opening the issue.
As this is being addressed i'm currently working on a alternative so that you can parse xml documents with Gatsby. If you don't mind waiting a bit longer, i'll post a alternative for you shortly, i have almost complete just needs a bit of tweaking, sounds good?

pieh:
@jonniebigodes does your alternative have similar api to `gatsby-transformer-xml`? Maybe we should just bump major version and use that for it?

pieh:
But back to the question - we are happy to review / accept PR that will swap unmaintained dependency in the plugin - ideally if it would mean breaking changes for users, but it wouldn't be blocker if there would be (as mentioned in previous comment - we can bump major version if there would be breaking changes)

laurenskling:
@jonniebigodes , sure I can wait :) currently I'm running a fork of `xml-parser` with a dirty CDATA `.replace`, but would love to get a stable fix.

I'm no hero in XML, but would love to help out testing a new tool.

jonniebigodes:
@laurenskling i've created a simple proof of concept like i mentioned. But before we head down that road, a side note.
Before some actual implementation i did a little research on alternative packages that could be more of help.
After applying some metrics to it, namelly:
 - Package update ratio
 - Number of uses/downloads in both github and npmjs.com
- Options available
- Ease of use
- Adaptability to the current plugin.
- Package dependencies

After taking some "out for a spin"( pardon the bad pun), i narrowed it down to these 3
- [xml-js](https://github.com/nashwaan/xml-js)
- [fast-xml-parser](https://github.com/NaturalIntelligence/fast-xml-parser)
- [xml2js](https://github.com/Leonidas-from-XIV/node-xml2js)

Regarding update they all seem pretty active in the past year.
[...]

PR #15598, opened 2019-07-10
feat(gatsby-transformer-xml): replace xml-parser with xml-js

MadsMadsDk:
## Description

xml-parser does not seem to be maintained any longer, and didn't support cdata. xml-js supports cdata. This implementation is largely created by @jonniebigodes ([based upon this implementation](https://github.com/jonniebigodes/test_gatsby_xml_parser_alternative/blob/master/plugins/gatsby-transformer-xml-alternative/gatsby-node.js])) and tweaked a little by me.

## Related Issues

#11736


jonniebigodes:
@MadsMadsDk go for it...i didn't fully tested the code i had in that repository, simply put i just didn't have the time to do it. But if it's working i'm hoping to see this merged and the plugin updated. Thank you for working on this. 👍

pieh:
In case that breaking changes are desired, it would be great to provide small migration guide in plugin's README to walk people through, by adding section called "Migrating from version 2 of `gatsby-transformer-xml`" or something like this

jonniebigodes:
@pieh if you don't mind me "barging in" when i created that repo with the code behind this PR, it was my first go at gatsby plugins, And i wasn't too familiar with the options concept. But implementing them via a configuration shouldn't be too hard. 

Also the 3rd level was hardcoded because i wanted to generate a fully functional node schema/tree first and then optimize it so that it could be iterated recursively and with that allow a n depth search.

MadsMadsDk:
@pieh thanks for the feedback. I’ll look into improving those things tomorrow, they make perfectly good sense. 

@jonniebigodes thanks for laying the groundwork for this :)

pieh:
> @pieh if you don't mind me "barging in" when i created that repo with the code behind this PR, it was my first go at gatsby plugins, And i wasn't too familiar with the options concept. But implementing them via a configuration shouldn't be too hard.

I don't mind at all! :) It was really more of a question than suggestion about plugin options - it might not even make sense to expose configuration at all honestly, as it seems that those transforms that are happening after initial conversion rely on using `compact` and some of the `<something>Key` options (did I gathered after looking more into `xml-js` docs. I think also exposing just some of the config options might make sense - i.e. maybe exposing option to set `trim` (to preserve white space as they are) and `ignoreX` (to have cleaner schema by omitting bunch of fields) config might be useful, but others might not.

> Also the 3rd level was hardcoded originally because i wanted to generate a fully functional node schema/tree first and then optimize it so that it could be iterated recursively and with that allow a n depth search.

Understandable and also very valid thing to do to not overcomplicate things if not needed - specifically for your own personal usage.

packages/gatsby-transformer-xml/src/gatsby-node.js
```diff
@@ -8,37 +8,98 @@ async function onCreateNode({
...
+    trim: true,
+    ignoreComment: true,
+    declarationKey: `declaration`,
+    instructionKey: `instruction`,
+    attributesKey: `attributes`,
+    textKey: `text`,
+    cdataKey: `cdata`,
+    doctypeKey: `doctype`,
+    parentKey: `parent`,
+  })
```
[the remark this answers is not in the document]
MadsMadsDk: Alright, I've added the possibility to pass meaningful options to the plugin. Backwards compatibility is available behind a flag as well.

packages/gatsby-transformer-xml/src/gatsby-node.js
```diff
@@ -8,37 +8,98 @@ async function onCreateNode({
...
+        Object.entries(childvalue).forEach(([gckey, gvvalue]) => {
+          if (Array.isArray(gvvalue)) {
+            transformedObject.root.children = [
+              ...transformedObject.root.children,
+              { name: gckey, children: gvvalue },
+            ]
+          }
+        })
+      }
+    })
```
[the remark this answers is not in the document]
MadsMadsDk: Should be fixed now, so arbitrary depth is supported (for both v2.x version of the plugin and the new graphql syntax).

MadsMadsDk:
> In case that breaking changes are desired, it would be great to provide small migration guide in plugin's README to walk people through, by adding section called "Migrating from version 2 of `gatsby-transformer-xml`" or something like this

Should be added now.

packages/gatsby-transformer-xml/src/gatsby-node.js
```diff
@@ -1,44 +1,218 @@
...
+    ignoreCdata,
+    ignoreDoctype,
+    ignoreText,
+  } = pluginOptions
+
+  // merge defaults with plugin options
+  const options = _.pickBy(
+    Object.assign({}, defaultOptions, {
+      nativeType: typeof nativeType !== `undefined` ? nativeType : false,
+      trim,
```
muescha: why you for some settings are set default values and other not?

packages/gatsby-transformer-xml/README.md
```diff
@@ -13,6 +13,43 @@ Parses XML files. It supports also attributes
...
+```javascript
+// In your gatsby-config.js
+plugins: [
+  {
+    resolve: `gatsby-transformer-xml`,
+    options: {
+      // Whether to trim whitespace characters that may exist before and after the text.
+      trim: false,
+      // Whether to attempt converting text of numerals or of boolean values to native type. For example, "123" will be 123 and "true" will be true
+      nativeType: false,
```
muescha: you should mention which are mandatory and which are not ...

packages/gatsby-transformer-xml/package.json
```diff
@@ -10,7 +10,7 @@
     "@babel/runtime": "^7.0.0",
     "bluebird": "^3.5.0",
     "lodash": "^4.17.10",
-    "xml-parser": "^1.2.1"
+    "xml-js": "^1.6.11"
   },
```
muescha: you forget to change the plugin version number in [line 4](https://github.com/gatsbyjs/gatsby/pull/15598/files#diff-d9d0fb671b49b5f38c6211621468427fR4) from `2.1.0` to `3.0.0`

MadsMadsDk: Oh sure - I didn’t know if there was mechanics in place, somewhere in the pipeline, that would bump the package in the event of an accepted pull resurse containing breaking changes. Good to know, thanks :)

packages/gatsby-transformer-xml/README.md
MadsMadsDk: I can make it more explicit, sure :)
All options are optional - I’ll adjust the docs.

packages/gatsby-transformer-xml/package.json
muescha: i also don't know it exactly, but you mentioned `3.0` it in the `README`

packages/gatsby-transformer-xml/README.md
muescha: thats good - but if all are optional then in the merge part you should check all values against `undefined`

packages/gatsby-transformer-xml/src/gatsby-node.js
MadsMadsDk: Good question! It was mostly an assesment from my part, that the document declaration doesn’t provide too much useful data. There might be cases though. Same for vommenes. The nativeType is default set to false, to interpret output as strings, but I’m not sure if it’s better to set that to true.
The rest are set to false per default in the xml-js library, so no need for default values as undefined is falsy.

MadsMadsDk:
It just hit me, that since this pull request also supports the old syntax, would it make more sense to hide the new syntax behind a flag instead? So it would be a minor version bump instead of a major.

packages/gatsby-transformer-xml/package.json
LekoArts: We always change the version number when running lerna (for publishing). So please don’t change the version numbers in this or future PRs - thanks!

MadsMadsDk: Thanks for clarifying :)

MadsMadsDk:
@LekoArts should I close this pull request and open a new one, since I've realised we can update this plugin without introducing breaking changes, but rather introduce new features such as options, Cdata-support (which the original issue was adressing) and using XML element names as keys?

muescha:
@pieh since the latest commit the label `braking change` can be removed?

pieh:
> @pieh since the latest commit the label `braking change` can be removed?

Yeah, if it is no longer breaking change then sure. We use that label to remember when we need to bump major version when publishing

packages/gatsby-transformer-xml/src/__tests__/__snapshots__/gatsby-node.js.snap
```diff
@@ -4,12 +4,8 @@ exports[`Process XML nodes correctly correctly creates nodes from XML 1`] = `
 Array [
   Array [
     Object {
-      "attributes": Object {
-        "id": "bk101",
-      },
       "children": Array [],
-      "content": "",
-      "id": "bk101",
+      "id": "uuid-from-gatsby",
```
pieh: Seems like this is something that changed. Top level `id` attribute (if it exist) should be used as node's `id`

packages/gatsby-transformer-xml/src/__tests__/__snapshots__/gatsby-node.js.snap
```diff
@@ -4,12 +4,8 @@ exports[`Process XML nodes correctly correctly creates nodes from XML 1`] = `
 Array [
   Array [
     Object {
-      "attributes": Object {
-        "id": "bk101",
-      },
```
pieh: `attributes` seems to no longer show up like this - instead they go into `xmlChildren`:

```
      "xmlChildren": Array [
        Object {
          "content": "bk101",
          "name": "attributes",
        },
```

sidharthachatterjee:
@MadsMadsDk Is this good to go? Can you please clarify if this is a breaking change anymore or not. Thanks!

MadsMadsDk:
@sidharthachatterjee As @pieh pointed out, there's still an issue with attributes showing up in xmlChildren, so that's a breaking change. I haven't had the time to look into it, but I hope I can free up some time soon.

pieh:
I'll mark PR as "work in progress" in meantime

laurenskling:
@MadsMadsDk , any news on this mate? Would love to help out. The old xml-parser needs to go

MadsMadsDk:
@laurenskling I'm taking another swing at this today. Hopefully get it resolved.

packages/gatsby-transformer-xml/src/__tests__/__snapshots__/gatsby-node.js.snap
MadsMadsDk: This should be resolved with the latest commits.

packages/gatsby-transformer-xml/src/__tests__/__snapshots__/gatsby-node.js.snap
MadsMadsDk: Should be fixed as well.

pieh:
There are still differences in results between current implementation and one in this PR, but it might be just time to accept that this just need to be breaking change and just bump major version for `gatsby-transformer-xml`.

So let's just drop trying to make it non-breaking change and mark it as breaking change (needing major version bump). I'll add review of this PR through that lense for my tasks for tomorrow

packages/gatsby-transformer-xml/src/gatsby-node.js
```diff
@@ -1,44 +1,198 @@
...
+    trim: typeof trim !== `undefined` ? trim : false,
+    ignoreDeclaration:
+      typeof ignoreDeclaration !== `undefined` ? ignoreDeclaration : true,
+    ignoreInstruction:
+      typeof ignoreInstruction !== `undefined` ? ignoreInstruction : false,
+    ignoreAttributes:
+      typeof ignoreAttributes !== `undefined` ? ignoreAttributes : false,
+    ignoreComment: typeof ignoreComment !== `undefined` ? ignoreComment : true,
+    ignoreCdata: typeof ignoreCdata !== `undefined` ? ignoreCdata : false,
+    ignoreDoctype: typeof ignoreDoctype !== `undefined` ? ignoreDoctype : false,
```
muescha, line 49: It can refactored into a `defaultPluginOptions` ?

And then:
```javascript
Object.assign({}, defaultOptions, defaultPluginOptions, pluginOptions)
```

pieh: Right, I don't quite follow why we have 2 sets of defaults and they are implemented in bit different way.

I guess this might be related to this note in README:

```
Only a subset of the [xml-js options](https://www.npmjs.com/package/xml-js#options-for-converting-xml-%E2%86%92-js-object--json) are available. Allowing for non-compact mode would currently break the implementation.
```

but this could also be done with some white-listing (for example using `lodash`'s `_.pick`) if that's the case

pieh:
If we end up dropping the idea of making this non-breaking change (still need to look over the snapshot changes to see if that's the case or not), it might also be beneficial to change some defaults to be more sensible (particularly `useElementNamesAsKey` seems like more sensible default than using `xmlChildren` array) - still TBD just writing my thoughts as they go in.

pieh:
Ok, seems like I misunderstood `useElementNamesAsKey` - it doesn't seem to try to match the description in the readme. In the current state it doesn't seem like the PR tries to be backward compatible (there is no sign of `xmlChildren` in the code). I would be fine with making breaking change like that (and bumping major), but the documentation just doesn't reflect the changes (it still mentions `xmlChildren`).

I will put this in draft mode (but not sure if @MadsMadsDk is still interested in pursuing this) - I tried to get make small changes to make it work but it's clear that it still need lot of work.

One of additional problems I discovered is when top level wrapping xml element have attribute (for example adding `something="foo"` in https://github.com/gatsbyjs/gatsby/blob/b9a28f779eb355692c0a455eb00893adfc85f108/packages/gatsby-transformer-xml/src/__tests__/gatsby-node.js#L20 ) also breaks a lot of things

LekoArts:
@MadsMadsDk Hi! Would you still be interested in moving forward with this PR?

MadsMadsDk:
> @MadsMadsDk Hi! Would you still be interested in moving forward with this PR?

I indeed very much would, but I'm afraid I don't have the time currently. If anyone could take over, that would be pretty great. Not sure when I'll have time to look into it again - also not entirely sure what's still missing.

LekoArts:
Okay! Then we'll close this for now. If anyone wants to pick this up, please comment here.

Issue #15504, opened 2019-07-08
Google Analytics plugin should have configurable delay before sending Page View


Simon-Tang:
## Summary

To build on the fix for https://github.com/gatsbyjs/gatsby/issues/9139 in https://github.com/gatsbyjs/gatsby/pull/10917, we need to support an optional configurable delay before invoking `sendPageView` for sites that have custom page transitions. (An example plugin: [gatsby-plugin-transition-link](https://transitionlink.tylerbarnes.ca/docs/transitionlink/)).

I can probably make this change in the code, but I wanted to see if anyone else had thoughts on this (e.g. the naming of the option).

### Basic example

In `gatsby-config.js`:

```js
{
  resolve: 'gatsby-plugin-google-analytics`,
  options: {
    pageViewDelay: 250,
  }
},
```

### Motivation

With page transitions, such as when using the plugin linked above, the next route can be mounted after an arbitrary delay, so the current `requestAnimationFrame`/`setTimeout` in `onRouteUpdate` will need to be replaced with a user-configured delay in this case.

Simon-Tang:
Fixed in #15610

PR #15610, opened 2019-07-10
feat(gatsby-plugin-google-analytics): Add pageTransitionDelay option

Simon-Tang:
## Description

Adds a `pageTransitionDelay` option to `gatsby-plugin-google-analytics` to delay the "Page View" event when necessary for page transitions.

## Related Issues

https://github.com/gatsbyjs/gatsby/issues/15504 - contains more context.

### DIFF
--- a/packages/gatsby-plugin-google-analytics/src/__tests__/gatsby-browser.js
+++ b/packages/gatsby-plugin-google-analytics/src/__tests__/gatsby-browser.js
@@ -81,7 +81,7 @@ describe(`gatsby-plugin-google-analytics`, () => {
 
           jest.runAllTimers()
 
-          expect(setTimeout).toHaveBeenCalledTimes(1)
+          expect(setTimeout).toHaveBeenCalledWith(expect.any(Function), 1000)
           expect(window.ga).toHaveBeenCalledTimes(2)
         })
       })
--- a/packages/gatsby-plugin-google-analytics/src/__tests__/gatsby-browser.js
+++ b/packages/gatsby-plugin-google-analytics/src/__tests__/gatsby-browser.js
@@ -29,9 +29,6 @@ describe(`gatsby-plugin-google-analytics`, () => {
         beforeEach(() => {
           jest.useFakeTimers()
           window.ga = jest.fn()
-          window.requestAnimationFrame = jest.fn(cb => {
-            cb()
-          })
         })
 
         afterEach(() => {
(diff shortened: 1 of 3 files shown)


packages/gatsby-plugin-google-analytics/src/gatsby-browser.js
```diff
@@ -20,13 +20,14 @@ export const onRouteUpdate = ({ location }) => {
     window.ga(`send`, `pageview`)
   }
 
-  if (`requestAnimationFrame` in window) {
+  const delay = pluginOptions.pageTransitionDelay
```
wardpeet: Let's rewrite to something like below. Let's move to setTimeout for all the things.

```suggestion
  const delay = Math.max(32, pluginOptions.pageTransitionDelay)
  setTimeout(sendPageView, delay)
```

packages/gatsby-plugin-google-analytics/src/__tests__/gatsby-browser.js
```diff
@@ -29,12 +27,17 @@ describe(`gatsby-plugin-google-analytics`, () => {
         })
 
         beforeEach(() => {
+          jest.useFakeTimers()
           window.ga = jest.fn()
           window.requestAnimationFrame = jest.fn(cb => {
```
wardpeet: with removing requestAnmationFrame from the code below you can remove this mock.

packages/gatsby-plugin-google-analytics/src/__tests__/gatsby-browser.js
```diff
@@ -72,6 +75,15 @@ describe(`gatsby-plugin-google-analytics`, () => {
           expect(setTimeout).toHaveBeenCalledTimes(1)
           expect(window.ga).toHaveBeenCalledTimes(2)
         })
+
+        it(`uses setTimeout when pageTransitionDelay is set`, () => {
+          onRouteUpdate({}, { pageTransitionDelay: 1000 })
+
+          jest.runAllTimers()
+
+          expect(setTimeout).toHaveBeenCalledTimes(1)
```
wardpeet: We should check if the correct delay is used
```suggestion
          expect(setTimeout).toHaveBeenCalledWith(expect.any(Function), 1000)
```

Simon-Tang:
Good catches @wardpeet, thanks!

wardpeet:
fixed in `gatsby-google-analaytics@2.1.3`

Issue #13030, opened 2019-04-02
Gatsby does not recognize Contentful 404 page


janosh:
## Description

According to [the docs](https://www.gatsbyjs.org/docs/add-404-page), creating a page whose path matches the regex `/404*` should be recognized as the 404 page. However, the development 404 page says no custom 404 page was detected even though a page with such a path exists.

<img width="1422" alt="custom-404-not-detected" src="https://user-images.githubusercontent.com/30958850/55406441-6d437380-555c-11e9-81df-8ca2f3bc8298.png">

### Steps to reproduce

Create a page in Contentful with slug `/404`.

### Expected result

Show button to preview custom 404 page.

floriangaechter:
Do you still have the `src/pages/404.js` file?

floriangaechter:
Sorry, strike that – I just saw in your error message that you deleted it…

I had to delete the cache folder for Gatsby to recognize the 404 page coming from Contentful when I just tried it out.

janosh:
@floriangaechter I don't have a `src/pages/404.js` file? I deleted that file since it essentially just reproduces `src/template/page.js`. With `src/pages/404.js` everything works as expected, without it, Gatsby doesn't seem to pick up on the Contentful page even though it matches the path prescribed by the docs.

floriangaechter:
Did you try deleting the `.cache` folder? I had issues with the 404 page coming from Contentful, but once I deleted the `.cache` folder it worked…

janosh:
@floriangaechter I did attempt `gatsby clean && gatsby develop` multiple times without success. Do you have a link to the repo with the Contentful 404 page? Maybe I can find the problem by comparing with your code.

floriangaechter:
Ok, I just tried to recreate your issue and noticed that Gatsby actually creates a 404 page from the Contentful source. If you have a look in your public directory you'll see that Gatsby created a `404/index.html` file but the actual 404 page is expected to be at /404.html (which doesn't exist if you deleted the `src/pages/404.js` file).

I guess you'd have to figure out a way to make Gatsby create the `/404.html` file instead of the `/404/index.html` file. Or you stick with the static 404 page and pull in the content via page query.

janosh:
[...]

Issue #15429, opened 2019-07-05
404 Develop Page: Blank in theme development


LekoArts:
## Description

Following Jason's egghead course until step 5. The 404 develop page should show up showing the available pages, instead the pages stays blank and displays nothing.

### Steps to reproduce

Setup a monorepo with a theme and example site, add some pages to the theme (basically following Jason's course).

Run `yarn workspace my-theme develop` and try to visit `localhost:8000/asdf`.

### Expected result

The dev browser should display the 404 dev page.

### Actual result

Blank page and error in the terminal:

```
Page not found /404.html
Error loading a result for the page query in "/asdf". Query was not run and no cached result was found.
Page not found /asdf
```

### Environment

  System:
    OS: macOS 10.14.5
    CPU: (8) x64 Intel(R) Core(TM) i5-8279U CPU @ 2.40GHz
    Shell: 5.7.1 - /opt/local/bin/zsh
  Binaries:
    Node: 10.16.0 - ~/.nvm/versions/node/v10.16.0/bin/node
    Yarn: 1.16.0 - /opt/local/bin/yarn
    npm: 6.9.0 - ~/.nvm/versions/node/v10.16.0/bin/npm
  Languages:
    Python: 2.7.10 - /usr/bin/python
  Browsers:
    Chrome: 75.0.3770.100
    Safari: 12.1.1
  npmGlobalPackages:
    gatsby-cli: 2.7.7

ryanwiemer:
@LekoArts I noticed this as well when taking the course. 

This brings up a related question. How should static pages be created when developing a theme? 

Not only is the 404 page not configured automatically it doesn't appears to be created at all. 404 pages or any other static page normally I add them in `src/pages/` and Gatsby creates the page for me without requiring me to specifically include it in `gatsby-node.js`. With theme development however it seems that functionality doesn't work. It's not really clear to me if that is intentional or just a bug.

ChristopherBiscardi:
> How should static pages be created when developing a theme?

Static pages can be created using `gatsby-plugin-page-creator` (the same plugin gatsby itself uses), or by making `createPage` calls. Using `gatsby-plugin-page-creator` allows the user's site to shadow a relevant page by creating their own page at `src/pages/my-static-page.js`.

The 404 page is a special case that isn't handled in the same way. I'm not sure why this is currently happening and have to look into it.

Polcius:
Same issue here. 

[...]

PR #15615, opened 2019-07-10
 fix(gatsby): handle `/404/`, `/404`, `404/` or `404` page paths

pieh:
Relax current logic of finding 404 page.

Currently page must be exactly `/404/` - which is what `gatsby-plugin-page-creator` will use when user have `src/pages/404.html`.

With this change `/404/`, `/404`, `404/` or `404` all will be considered 404 page. This is mostly quality of life improvement, so users creating 404 pages don't have to scratch their head and look in documentation (or worse in source code) why it didn't work if path wasn't exactly what Gatsby expects

Related: https://github.com/gatsbyjs/gatsby/issues/15429

Fixes: https://github.com/gatsbyjs/gatsby/issues/13030

### DIFF
--- a/docs/docs/add-404-page.md
+++ b/docs/docs/add-404-page.md
@@ -3,7 +3,7 @@ title: "Adding a 404 Page"
 ---
 
 To create a 404 page create a page whose path matches the regex
-`/404*`. Most often you'll want to create a React component page at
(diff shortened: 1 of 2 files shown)


sidharthachatterjee:
Published in `gatsby@2.13.15`

janosh:
Nice work! Fix confirmed.

PR #15619, opened 2019-07-11
fix(gatsby): Make createPage action errors structured

KyleAMathews:
And improve them a bit.

### DIFF
--- a/packages/gatsby-cli/src/structured-errors/error-map.js
+++ b/packages/gatsby-cli/src/structured-errors/error-map.js
@@ -104,7 +104,7 @@ const errorMap = {
     text: context =>
       `${
         context.pluginName
-      } must set the page path when creating a page\n\nThe page object passed to createPage:\n${JSON.stringify(
+      } must set the page path when creating a page.\n\nThe page object passed to createPage:\n${JSON.stringify(
         context.pageObject,
         null,
         4
--- a/packages/gatsby-cli/src/structured-errors/error-map.js
+++ b/packages/gatsby-cli/src/structured-errors/error-map.js
@@ -89,6 +89,77 @@ const errorMap = {
     type: `PLUGIN`,
     level: `ERROR`,
   },
+  "11322": {
+    text: context =>
+      `${
+        context.pluginName
+      } created a page and didn't pass the path to the component.\n\nThe page object passed to createPage:\n${JSON.stringify(
+        context.pageObject,
+        null,
+        4
+      )}\n\nSee the documentation for the "createPage" action — https://www.gatsbyjs.org/docs/actions/#createPage`,
+    level: `ERROR`,
+  },
+  "11323": {
+    text: context =>
+      `${
+        context.pluginName
+      } must set the page path when creating a page\n\nThe page object passed to createPage:\n${JSON.stringify(
+        context.pageObject,
+        null,
+        4
+      )}\n\nSee the documentation for the "createPage" action — https://www.gatsbyjs.org/docs/actions/#createPage`,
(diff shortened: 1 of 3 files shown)


KyleAMathews:
Before:
Screenshot 2019-07-10 17 27 53

After:
Screenshot 2019-07-10 19 41 06

packages/gatsby-cli/src/structured-errors/error-map.js
```diff
@@ -89,6 +89,77 @@ const errorMap = {
...
   },
+  "11322": {
+    text: context =>
+      `${
+        context.pluginName
+      } created a page and didn't pass the path to the component.\n\nThe page object passed to createPage:\n${JSON.stringify(
+        context.pageObject,
+        null,
+        4
+      )}\n\nSee the documentation for the "createPage" action — https://www.gatsbyjs.org/docs/actions/#createPage`,
```
wardpeet: Should we move to docsUrl instead? Or is docsUrl not really for this usecase
```suggestion
      )}`,
    docsUrl: `https://www.gatsbyjs.org/docs/actions/#createPage`
```

packages/gatsby/src/redux/actions/public.js
```diff
@@ -325,12 +332,21 @@ ${reservedFields.map(f => `  * "${f}"`).join(`\n`)}
...
       }
 
       if (!includesDefaultExport) {
+        report.panicOnBuild({
+          id: `11328`,
+          context: {
+            fileName,
+          },
+        })
         report.panicOnBuild(
```
wardpeet: should we remove this line?

packages/gatsby-cli/src/structured-errors/error-map.js
KyleAMathews: This is why we need MDX actually — `docsUrl` is nice but it hardcodes the layout of stuff where I want a link but I want to preface it with a cue for when you'd need to go visit the page.

So I'll leave it as is for now.

packages/gatsby/src/redux/actions/public.js
KyleAMathews: oh oops yeah

packages/gatsby/src/redux/actions/public.js
```diff
@@ -172,97 +174,110 @@ ${reservedFields.map(f => `  * "${f}"`).join(`\n`)}
...
-        // version.  People in v1 often thought that they needed to also pass
-        // the path to context for it to be available in GraphQL
-      } else if (invalidFields.some(f => page.context[f] !== page[f])) {
-        report.panic(error)
       } else {
-        if (!hasWarnedForPageComponentInvalidContext.has(page.component)) {
-          report.warn(error)
-          hasWarnedForPageComponentInvalidContext.add(page.component)
-        }
+        report.panic({
```
sidharthachatterjee: Is deduping not necessary anymore?

(Like we have in `hasWarnedForPageComponentInvalidContext` in the before)

KyleAMathews: I just changed it to error every time instead of warning.

I added this code during pre-v2 as there'd been a bad pattern that had developed with people creating pages during v1 and I wanted to warn people against it. So I made it a warning during development and error during builds. It's been long enough that I thought just simplifying things to always error would work.

sidharthachatterjee: We'll probably want to not change this behaviour (warn versus error) as a hard fail would be a breaking change IMO

As a follow up to that, if we're warning, we'll likely want to keep the dedupe logic

PR #15636, opened 2019-07-11
fix(svg): don't try to use convertShapeToPath svgo optim

### LINKED ISSUE #15629: Svg Loading when using `gatsby develop` but not `gatsby serve`
## Description

I have an svg that I'm referencing directly in my css. I can see it render in the `gatsby develop` view of my site, but not when I build and serve.

### Steps to reproduce

This is the repository and the specific line in question: https://github.com/lstar19/website-gatsby/blob/87f183ff4a3d5d785cb2dd47a3983aa22d4a5642/src/assets/css/test.css#L327

Run `gatsby develop` and on the http://localhost:8000/blog/ page, each blog post block will have subtle lines.

Run `gatsby build`, `gatsby serve` and navigate to http://localhost:9000/blog/. The lines will no longer appear.

### Expected result

These pages should be the same.

### Actual result

The svg only appears in develop mode.

### Environment

  System:
    OS: macOS 10.14.4
    CPU: (8) x64 Intel(R) Core(TM) i7-4980HQ CPU @ 2.80GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 11.0.0 - ~/.nvm/versions/node/v11.0.0/bin/node
    Yarn: 1.17.0 - /usr/local/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v11.0.0/bin/npm
  Languages:
    Python: 2.7.12 - /usr/local/bin/python
  Browsers:
    Chrome: 75.0.3770.100
    Firefox: 49.0.1
    Safari: 12.1
  npmPackages:
    gatsby: ^2.13.12 => 2.13.12
    gatsby-background-image: ^0.2.74 => 0.2.74
    gatsby-image: ^2.2.4 => 2.2.4
    gatsby-plugin-google-gtag: ^1.1.1 => 1.1.1
    gatsby-plugin-gtag: ^1.0.11 => 1.0.11
    gatsby-plugin-manifest: ^2.2.1 => 2.2.1
    gatsby-plugin-offline: ^2.2.1 => 2.2.1
    gatsby-plugin-react-svg: ^2.1.1 => 2.1.1
    gatsby-plugin-sass: ^2.1.0 => 2.1.0
    gatsby-plugin-sharp: ^2.2.3 => 2.2.3
    gatsby-source-filesystem: ^2.1.3 => 2.1.3
    gatsby-transformer-remark: ^2.6.1 => 2.6.1
    gatsby-transformer-sharp: ^2.2.1 => 2.2.1
    gatsby-transformer-yaml: ^2.2.0 => 2.2.0
  npmGlobalPackages:
    gatsby-cli: 2.4.17

pieh:
This is pretty messy - it fixes exact problem described in #15629, but I wonder if we should just disable svgo all together, as it proved to cause lot of issues when trying to optimize svg and user doesn't have any control over it

Fixes #15629


KyleAMathews:
Has it caused a lot of bugs?

pieh:
> Has it caused a lot of bugs?

Maybe not lot of bugs, but definitely some - and those are usually pretty frustrating ones, as user can't really turn this off or themselves (or overwrite configuration).

From my personal experience - another optim in `svgo` that likes to cause troubles is `removeViewBox`.

It probably make sense to evaluate each optim case by case, as there is definitely benefit to using svgo in general. It might be that my frustration in figuring out what caused this exactly took better of me (and reminded me of frustrating problems I saw in the past)

KyleAMathews:
>It probably make sense to evaluate each optim case by case, as there is definitely benefit to using svgo in general. It might be that my frustration in figuring out what caused this exactly took better of me (and reminded me of frustrating problems I saw in the past)

Can we perhaps whitelist optimizations instead of blacklisting them? Use a minimal set?

pieh:
> Can we perhaps whitelist optimizations instead of blacklisting them? Use a minimal set?

It doesn't seem like it. Looking at `svgo` docs they list available optims with notes if something is disabled by default - https://github.com/svg/svgo#what-it-can-do - and their default status seems to come from `active` exports for each optim module (for example https://github.com/svg/svgo/blob/95f7603983d50aa579b6ebb21fa747d0436e5581/plugins/convertShapeToPath.js#L5 ) and I didn't find any other defaults settings up the dependency chain SVGO < postcss-svgo < cssnano-preset-default < cssnano < optimize-css-assets-webpack-plugin.

So maybe approach in this PR of selectively addressing problematic optimizations is not the worst idea.

m-allanson:
It seems we could whitelist by using the `full: true` config option, and supplying the full list of svgo plugins that should be run. I'll try updating this PR.

m-allanson:
I've updated this to use a whitelist approach. SVGO's default config is whitelisted with the exception of `convertShapeToPath` and `removeViewBox`.

In future we could look at disabling all destructive optimisations by default? See [discussion here](https://github.com/svg/svgo/issues/646#issuecomment-273513053) and [an example in the ImageOptim repo](https://github.com/ImageOptim/SVGO-bin/commit/d5355262a8d22116821f7fad5bce603dc2d28edc).

wardpeet:
Fixed in `gatsby@2.13.16`

spikyjt:
I know this PR is rather old now, but it causes a bug in current gatsby, when using svgs as uris in scss (at least).

I'm not sure where the whiteliist came from, but the format is not correct. The plugins should either be listed as an array of strings, or as an array of objects, where each object has `name` and `active` properties as a minimum. See the docs on plugins at https://github.com/svg/svgo

The correct way to do this now would be either to pass a whitelist like this:
```
...
plugins: [
  {
    name: 'cleanupListOfValues',
    active: true,
  },
  ...
],
```

or (possibly better depending on the intention) add the extras and disable the ones we don't want using `extendDefaultPlugins`:
```
import { extendDefaultPlugins } from "svgo"

...
plugins: extendDefaultPlugins([
  {
    name: 'convertShapeToPath',
    active: false,
  },
  ...
  {
    name: 'cleanupListOfValues',
    active: true,
  },
  ...
]),
```

Additionally, there are some plugins, such as `addAttributesToSVGElement` that require additional parameters to configure, and make no sense being simply enabled.

What was the intention for the added plugins in this PR? Was it simply to use the defaults but disable the ones seen to be risky? Or was it to specifically set a known working list of plugins? (I don't see how the second option can be true, looking at the list).

Based on the answer to the above, I can create a new PR to address this. as I've done some tweaking on a test setup to identified the cause of the errors.

nakedible-p:
I have the same problem! Please create a new pr, I'd really like to see this fixed.

nakedible-p:
Nevermind, fixed in https://github.com/gatsbyjs/gatsby/pull/31620 - it just was not linked here.

PR #15638, opened 2019-07-11
fix(gatsby-plugin-mdx): add missing change-case dependency

DSchau:
## Description

There’s a babel utility in gatsby-plugin-mdx that depends on `change-case` implicitly, but does not add the dependency and so is thus failing builds on www with:

> Cannot find module 'change-case'


sidharthachatterjee:
Published in `gatsby-plugin-mdx@1.0.11`

PR #15642, opened 2019-07-11
chore(www): update banner text to point users to Gatsby Preview

gillkyle:
## Description

Pretty self explanatory really.

image

### DIFF
--- a/www/src/components/__tests__/banner.js
+++ b/www/src/components/__tests__/banner.js
@@ -3,7 +3,7 @@ import { render } from "react-testing-library"
 
 import Banner from "../banner"
 
-const matchers = [`Watch now`, `Register now`]
+const matchers = [`Watch now`, `Register now`, `Gatsby Preview`]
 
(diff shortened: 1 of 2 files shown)


www/src/components/__tests__/banner.js
```diff
@@ -3,7 +3,7 @@ import { render } from "react-testing-library"
 
 import Banner from "../banner"
 
-const matchers = [`Watch now`, `Register now`]
+const matchers = [`Watch now`, `Register now`, `Gatsby Preview`]
```
muescha: the old matchers are unused?

gillkyle: Yeah that's because they've bounced back between before/after webinars. When the text of the banner changes we won't have to update the test again. It's not exactly the best or most useful test anyway.

muescha: maybe better solution are rotating banners?

PR #15643, opened 2019-07-11
Fix(www): uppercase breadcrumbTitles and change separator to ChevronLeft

muescha:
## Description

Fixes breadcrumb titles and change separator to ChevronLeft

## Related Issues

introduced in #15165

### DIFF
--- a/www/src/data/sidebars/doc-links.yaml
+++ b/www/src/data/sidebars/doc-links.yaml
@@ -77,7 +77,7 @@
               breadcrumbTitle: Babel Plugin Macros
             - title: Adding a Custom Webpack Config
               link: /docs/add-custom-webpack-config/
-              breadcrumbTitle: webpack Config
+              breadcrumbTitle: Webpack Config
             - title: Customizing html.js
               link: /docs/custom-html/
               breadcrumbTitle: html.js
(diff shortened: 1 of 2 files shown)


www/src/data/sidebars/doc-links.yaml
```diff
@@ -77,7 +77,7 @@
               breadcrumbTitle: Babel Plugin Macros
             - title: Adding a Custom Webpack Config
               link: /docs/add-custom-webpack-config/
-              breadcrumbTitle: webpack Config
+              breadcrumbTitle: Webpack Config
```
gillkyle: I'm always confused by this because webpack's branding guidelines say to always write it in lowercase, but the docs seem to break that rule all over the place already 🤔 

https://webpack.js.org/branding/

muescha: mhh - but there is no rule for title - should it then be: `Adding a Custom webpack Config`

gillkyle: I think it's fine how it is, this would be really easy to update in one fell swoop across the whole repo if we want to. We capitalize it everywhere else so I think it's good with your change 🙂

PR #15644, opened 2019-07-11
fix(gatsby-plugin-mdx): Fix Safari 9 `const` syntax error

jonathanphz:
## Description

Gatsby sites using mdx, break in Safari 9 (possibly other older browsers) because it's using es6 `const`.

`SyntaxError: Unexpected keyword 'const'. Const declarations are not supported in strict mode.`
Dashboard


wardpeet:
fixed in `gatsby-plugin-mdx@1.0.12`

PR #15637, opened 2019-07-11
chore(themes): Add some basic end to end tests for themes

sidharthachatterjee:
Ultra simple tests that assert 
- page shadowing
- component shadowing

And also workspace functionality implicitly 😉 

I'll add some more for the development runtime as well

PR #15649, opened 2019-07-12
feat(gatsby-dev-cli): add initial support for handling yarn workspaces

pieh:
This will be needed to properly support e2e tests for themes ( https://github.com/gatsbyjs/gatsby/pull/15637 )


packages/gatsby-dev-cli/src/local-npm-registry/install-packages.js
```diff
@@ -1,22 +1,82 @@
...
+    // in workspaces which should preserve node_modules structure
+    // (packages being mostly hoisted to top-level node_modules)
+
+    const { stdout } = await promisifiedSpawn([
+      `yarn`,
+      [`workspaces`, `info`, `--json`],
+      { stdio: `pipe` },
+    ])
+
+    const workspacesLayout = JSON.parse(JSON.parse(stdout).data)
```
sidharthachatterjee: 😄

packages/gatsby-dev-cli/src/local-npm-registry/install-packages.js
```diff
@@ -1,22 +1,82 @@
...
+    Object.keys(workspacesLayout).forEach(workspaceName => {
+      const { location } = workspacesLayout[workspaceName]
+      const pkgJsonPath = path.join(yarnWorkspaceRoot, location, `package.json`)
+      if (!fs.existsSync(pkgJsonPath)) {
+        return
+      }
+      const pkg = JSON.parse(fs.readFileSync(pkgJsonPath, `utf8`))
+
+      let changed = false
+      changed |= handleDeps(pkg.dependencies)
```
sidharthachatterjee: Nice Bitwise OR!

But I'd change this to 
```
changed = changed || handleDeps(pkg.dependencies) || handleDeps(pkg.devDependencies) || handleDeps(pkg.peerDependencies)
```
for readability

packages/gatsby-dev-cli/src/watch.js
```diff
@@ -103,19 +113,30 @@ function watch(
...
-    packages.forEach(pkg => {
-      if (localPackages.includes(pkg)) {
-        packagesToInstall.push(pkg)
-      }
-      if (depTree[pkg]) {
-        getPackagesToInstall([...depTree[pkg]])
-      }
-    })
+    return () => {
+      ignoredPackageJSON.delete(ignoredPackageJSON)
```
sidharthachatterjee: Shouldn't this be 
```suggestion
      ignoredPackageJSON.delete(packageName)
```

packages/gatsby-dev-cli/src/local-npm-registry/install-packages.js
pieh: This will change behaviour - if `handleDeps(pkg.dependencies)` then next one won't ever be called and I do need to change it everywhere

packages/gatsby-dev-cli/src/watch.js
pieh: Oh yeah! It was like that before and wasn't caught by anyone ops

packages/gatsby-dev-cli/src/local-npm-registry/install-packages.js
pieh: So to avoid `|=` thsi would need to be:
```
if (handleDeps(pkg.dependencies)) {
  changed = true
}
if (handleDeps(pkg.devDependencies)) {
  changed = true
}
if (handleDeps(pkg.peerDependencies)) {
  changed = true
}
```

sidharthachatterjee: Ah, I'd typically want to avoid functions with side effects in conditionals due to this short circuiting 

But not a blocker for now!

PR #15656, opened 2019-07-12
chore(showcase): Add marcelhauri.ch to showcase

mhauri:
## Description

Adding marcelhauri.ch to showcase


mhauri:
Thank you @wardpeet 😊

PR #15657, opened 2019-07-12
fix(babel-preset-gatsby): Add babel-plugin-dynamic-import-node for tests

sidharthachatterjee:
Node 8 (which we still support) doesn't understand dynamic imports. While it's okay to keep these untranspiled (like in `gatsby-image/withIEPolyfill`) because webpack will handle them anyway, this breaks tests run on Node 8 (https://circleci.com/gh/gatsbyjs/gatsby/169116?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). 

To fix this, we need to conditionally transpile these to deferred requires for Node in test environments. This PR does that for this monorepo's jest config and for `babel-preset-gatsby` which we recommend in https://www.gatsbyjs.org/docs/unit-testing/

### DIFF
--- a/jest-transformer.js
+++ b/jest-transformer.js
@@ -1,5 +1,2 @@
 const babelPreset = require(`babel-preset-gatsby-package`)()
-module.exports = require(`babel-jest`).createTransformer({
-  ...babelPreset,
-  plugins: [...babelPreset.plugins, `babel-plugin-dynamic-import-node`],
-})
+module.exports = require(`babel-jest`).createTransformer(babelPreset)
--- a/jest-transformer.js
+++ b/jest-transformer.js
@@ -1,2 +1,5 @@
 const babelPreset = require(`babel-preset-gatsby-package`)()
-module.exports = require(`babel-jest`).createTransformer(babelPreset)
+module.exports = require(`babel-jest`).createTransformer({
+  ...babelPreset,
+  plugins: [...babelPreset.plugins, `babel-plugin-dynamic-import-node`],
+})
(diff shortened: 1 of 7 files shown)


jest-transformer.js
```diff
@@ -1,2 +1,5 @@
 const babelPreset = require(`babel-preset-gatsby-package`)()
-module.exports = require(`babel-jest`).createTransformer(babelPreset)
+module.exports = require(`babel-jest`).createTransformer({
```
wardpeet: I would add a test environment to babel-reset-gatsby-package to include it so other plugins have access to it too.

https://babeljs.io/docs/en/options#env

packages/babel-preset-gatsby/src/index.js
```diff
@@ -2,9 +2,11 @@ const path = require(`path`)
 
 const resolve = m => require.resolve(m)
 
+const TEST = process.env.NODE_ENV === `test`
```
wardpeet: NIT
```suggestion
const IS_TEST = process.env.NODE_ENV === `test`
```

jest-transformer.js
sidharthachatterjee: Added it to `babel-preset-gatsby` so that other folks have access to it for their unit tests. Wanted to avoid adding it to `babel-preset-gatsby-package` for now after a conversation with @pieh because it doesn't seem necessary since this `jest-transformer.js` gets run for all tests at the moment.

wardpeet: We could ship it like this and re-iterate

packages/babel-preset-gatsby/src/index.js
sidharthachatterjee: Fixed

PR #15659, opened 2019-07-12
chore(telemetry): collect bootstrap times

jamo:
## Description

Start tracking bootstrap times 

This change doesn't require any upstream changes

### DIFF
--- a/packages/gatsby-cli/src/reporter/index.js
+++ b/packages/gatsby-cli/src/reporter/index.js
@@ -8,6 +8,8 @@ const { getErrorFormatter } = require(`./errors`)
 const reporterInstance = require(`./reporters`)
 const constructError = require(`../structured-errors/construct-error`)
 const errorFormatter = getErrorFormatter()
+const { trackCli } = require(`gatsby-telemetry`)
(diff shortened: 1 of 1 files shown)


packages/gatsby-cli/src/reporter/index.js
```diff
@@ -139,6 +142,16 @@ const reporter: Reporter = {
...
       },
       end() {
+        const diff = process.hrtime(startTime)
+
+        const NS_PER_SEC = 1e9
+        const MS_PER_NSEC = 1e6
+
+        trackCli(`ACTIVITY_DURATION`, {
+          name: name,
+          duration: Math.round((diff[0] * NS_PER_SEC + diff[1]) / MS_PER_NSEC),
```
pieh: Could we use https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-cli/src/util/calc-elapsed-time.js (or https://www.npmjs.com/package/convert-hrtime if precission returned by our util seems insufficient) instead of magic using magic numbers?

jamo: yess, thanks!

Updated the pr to use the convert-hrtime module.

PR #15665, opened 2019-07-12
fix(actions): fix PR bot action via await on each report call

gillkyle:
## Description

The GitHub action for the Learning PR bot failed yesterday with this stacktrace:

```
(node:1) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'id' of null
    at pr.reviewRequests.nodes.map.node (/src/filter.js:13:45)
    at Array.map (<anonymous>)
    at filteredQueues.(anonymous function).queues.(anonymous function).filter.pr (/src/filter.js:13:10)
    at Array.filter (<anonymous>)
    at Object.keys.forEach.key (/src/filter.js:11:39)
    at Array.forEach (<anonymous>)
    at filter (/src/filter.js:10:23)
    at teams.forEach.team (/src/index.js:28:16)
    at Array.forEach (<anonymous>)
    at start (/src/index.js:22:11)
```

I haven't see this issue triggering the action manually so I presume that there was just some race condition when GitHub fires the action and that adding an `await` to the `report()` function will let the message for the Core team fire and then wait for the Learning time message. 

I also added an if to protect for finding the `id` on null properties that the error showed.

Things run fine locally (but they also did before) so I'm kind of crossing my fingers this will just resolve things 🤷‍♂.

(I also added Amberley to the list of maintainers)


.github/actions/high-priority-prs/src/index.js
```diff
@@ -19,13 +19,13 @@ const start = async function() {
   const data = await fetch()
   if (data) {
     const queues = processData(data)
-    teams.forEach(team => {
+    teams.forEach(async team => {
```
wardpeet: does it make sense to go for
```
      await Promise.all(teams.map(async team => {
        if (team.name === `core`) {
          // send report of all PRs (without filtering) to Core
          await report({ queues, channelId: team.channelId })
        } else {
          // send filtered reports to respective teams
          await report(filter(queues, team))
        }
      }))
```

I don't know this code so this might not make any sense

gillkyle: I guess this would allow them to start running at the same time right? So rather than having to wait for the first one to report and then get to the next iteration the next `report` call could start. Seems worthwhile.

sidharthachatterjee:
Looking great! Thank you so much @gillkyle 🥇 

> Things run fine locally (but they also did before) so I'm kind of crossing my fingers this will just resolve things 🤷‍♂.

Was maybe failing because the Docker container in GitHub actions doesn't wait on unresolved promises (for an empty event loop) before exiting 🤷‍♂

.github/actions/high-priority-prs/src/index.js
```diff
@@ -19,15 +19,15 @@ const start = async function() {
   const data = await fetch()
   if (data) {
     const queues = processData(data)
-    teams.forEach(team => {
+    await Promise.all(teams.map(async team => {
```
sidharthachatterjee: Nit: The callback doesn't need to be an async function anymore

gillkyle:
> Was maybe failing because the Docker container in GitHub actions doesn't wait on unresolved promises (for an empty event loop) before exiting 🤷‍♂

Interesting, good to know!

.github/actions/high-priority-prs/src/index.js
sidharthachatterjee:
```suggestion
    await Promise.all(teams.map(team => {
```

PR #15689, opened 2019-07-13
feat(gatsby): Add enum and scalar type builders

stefanprobst:
This adds Gatsby Type Builders for enum and scalar types.
Especially enum type builder seems good to have because enums in SDL cannot have custom internal values.


sidharthachatterjee:
Published in `gatsby@2.13.21`

rhengles:
The `buildScalarType` appears to only create a type, is that right? How can that type be registered in gatsby schema for later use in my `createTypes(...)` ?

Issue #15347, opened 2019-07-03
[gatsby-source-wordpress] Differences in WPCOM API results in pagination errors


himynameistimli:
## Description

WPCOM's REST API differs slightly from the typical WP REST API, so using the header to determine number of items and number of pages does not work properly.

For instance, if you have over 100 tags, the endpoint will not pull more than 100, and then posts that link to those tags will result in:

```
TypeError: Cannot read property 'id' of undefined
```

If you're in a bind, you can remove some tags, and the build will proceed normally.

I'm talking to a WP support admin to determine what the proper path forward is. I'll update this post as I learn more.

When I get a chance, I'll make a PR to fix this issue, but I wanted to put up this issue in case others are using Wordpress.com instead of the self-hosted version.

freiksenet:
@himynameistimli Is it possible to page the endpoint to request rest of the tags? In that case it would be possible to update wordpress plugin to do that.

CC @jasonbahl

freiksenet:
Thank you for your patch!

PR #15695, opened 2019-07-14
docs(gatsby-source-wordpress): add note about limitation of pagination in WordPress.com API

himynameistimli:
## Description

This is just a documentation update for gatsby-source-wordpress. I've added some additional information for WordPress.com users, as well as added some troubleshooting documentation for a common issue.

This does not fix issue https://github.com/gatsbyjs/gatsby/issues/15347 I raised here, but rather provides some WordPress.com users with an alternative.

## Changes

- Add more information for WordPress.com users based on which plan they have
- Add note about limitation for some WordPress.com users in troubleshooting section (I guess it would be removed once the fix has been introduced)

### DIFF
--- a/packages/gatsby-source-wordpress/README.md
+++ b/packages/gatsby-source-wordpress/README.md
@@ -893,7 +893,7 @@ If you have custom post types or metadata that are not showing up within the sch
 
 - **Custom Meta**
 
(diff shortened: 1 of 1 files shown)


pieh:
I think it's ok to merge this in if this is known, not handled issue.

Do you know if wp.com provide alternative metadata for `X-WP-Total`/`X-WP-TotalPages`? If not, maybe it might make sense to try to get next pages for wp.com if we don't know if there are anymore or not? For example in situation when result length equals to 100 (or whatever posts_per_page is set) and those headers are not there we should try getting next page (and next one after that etc) - that's for follow up PRs not this one

Issue #14017, opened 2019-05-13
Define scope of gatsbyjs.org improvements


jlengstorf:
Thanks to insights from the community and hard work by @shannonbux, @amberleyromo, @marcysutton, and others, we have a backlog of issues to address on gatsbyjs.org.

This scope is intended to define the desired outcomes and how we will measure them. From this scope, individual issues will be created so this work can be done incrementally with confidence that we're all rowing in the same direction.

## What are the outcomes we want to achieve with these improvements?

This work will focus on the overall design + four critical sections of the gatsbyjs.org site.

**Anything not called out in this scope will not be addressed in this project, so please review this scope carefully to make sure we didn't miss anything!**

### Overall desired outcomes

The following aren't page-specific, but should be true for the overall site experience.

-  [code quality] Developers working on the site should have clear direction on how to use Gatsby's design tokens and common components
    - gatsbyjs.org is regularly referenced as a canonical example of implementing Gatsby. We need to ensure the code is using best practices and give developers solid examples of Gatsby patterns.
-  [discovery] Community members should be able to quickly find information about the most critical workflows
    - how to get started with Gatsby (i.e. quick start, tutorial)
    - how to find answers for specific questions about Gatsby (e.g. docs)
    - how to get involved in Gatsby (i.e. contributing section)
    - how to find themes, starters, plugins, and other community-powered ecosystem projects
-  [community] Visitors to the site should immediately feel the fun, welcoming personality of Gatsby and its community
    - Visitors should get a sense of Gatsby’s [core values](https://www.gatsbyjs.org/blog/2018-09-07-gatsby-values/), including our commitment to [accessibility](https://www.gatsbyjs.org/accessibility-statement) and our community [code of conduct](https://www.gatsbyjs.org/contributing/code-of-conduct/)
    - Visitors should see what other community members are creating (e.g. the site showcase, starters, plugins, tweets)
-  [search] Users should be able to find the information they're after with robust site search
    - Site searches should include the contributing section, plugins, starters, and other site content (see #12503)
-  [cloud] Visitors should know that Gatsby Cloud exists, understand the high-level value it provides, and know where to look for more information

### Home page

1. First-time visitors should leave with a clear understanding of Gatsby’s purpose, value, and credibility
    - what _is_ Gatsby?
    - what can Gatsby do?
    - who is using Gatsby?
2. First-time visitors should have a clear call to action to start with Gatsby

#### How do we measure success?

[...]

PR #15697, opened 2019-07-14
feat(www) : Add starters library to top navigation menu

ImedAdel:
## Description

The current navigation menu lacks a link to the starters library, which leads to using the browser's history or searching in google to find the page. This commit would add a link to starters library to make it easier to navigate to the page.

### DIFF
--- a/www/src/components/navigation.js
+++ b/www/src/components/navigation.js
@@ -126,6 +126,7 @@ const Navigation = ({ pathname }) => {
             <NavItem linkTo="/docs/">Docs</NavItem>
             <NavItem linkTo="/tutorial/">Tutorials</NavItem>
             <NavItem linkTo="/plugins/">Plugins</NavItem>
+            <NavItem linkTo="/starters/">Starters</NavItem>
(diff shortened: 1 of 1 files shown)


wardpeet:
Tagging @fk @gatsbyjs/ecosystem  @gatsbyjs/docs

m-allanson:
Thanks @ImedAdel - I agree the starters could be more visible, however I'm going to close this PR for now. That's because making this change is more complex than it initially seems, and there's a future project that should improve this part of the site.

This change would need updates to the CSS to ensure the updated navigation fits at all browser widths - currently this code is different for the homepage and the rest of the site. In addition, there's no space at smaller breakpoints to include this in the mobile nav, so it won't be visible to all visitors.

There's a project to make several improvements to gatsbyjs.org, and making sure key parts of the site are discoverable is one of the aims there. See https://github.com/gatsbyjs/gatsby/issues/14017 (although I think it's on hold for the moment).

For the above reasons I will close this. Thank you again for the PR and I hope to see more contributions from you in the future!

PR #15701, opened 2019-07-14
feat(Dockerfile): Set system locale to UTF-8

### LINKED ISSUE #15700: [www - gatsby-transformer-screenshot] Puppeteer cannot capture Chinese or UTF-8 properly
## Description

Puppeteer itself cannot recognize UTF-8 character (Chinese, Thai etc.) and fallback to square character instead

Screenshot

Screenshot

I think there's a way to resolve this by set system locale to UTF-8 in `Dockerfile` before Gatsby site was built, I will launch PR there.

### Steps to reproduce

Capture some Chinese sites screenshot by using `gatsby-transformer-screenshot`

### Expected result

I will use https://arnondora.in.th as an example

Screenshot

### Actual result

Screenshot

### Environment

I think there's no need here

rayriffy:
## Description

Set Docker image system locale to UTF-8 in order to rendering foreign character properly. *Need confirmation that this PR works and it don't break any screenshots in the site.*

## Related Issues

closes #15700


wardpeet:

We don't use the dockerfile to generate screenshots but run on an aws lambda. See https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-transformer-screenshot/lambda/index.js for the code. We're open to a complete rewrite of the screenshot service that's more reliable and works with uf8 characters.

Issue #14977, opened 2019-06-20
[docs][recipes] Querying data in pages


marcysutton:
## Summary

To improve key learning workflows in the Recipes section of the docs, there should be a short, actionable recipe on querying for data in pages with GraphQL.

This recipe should show how to write a pageQuery and provide the returned data to a page _inline_ as concisely as possible. It should briefly describe when someone would need a page query vs other kinds of queries (even if obvious) and include code snippets.

Recipes now follow this format:

- Requirements or prerequisites 
- Step-by-step instructions
- Relevant links to any relevant information
    - More on Page Query: https://www.gatsbyjs.org/docs/page-query/
    - Link to a live example

This recipe should go into the "Querying data" section of [`docs/docs/recipes.md`](https://github.com/gatsbyjs/gatsby/blob/master/docs/docs/recipes.md), replacing the existing text. There is a recipe template in the contributing docs: https://www.gatsbyjs.org/contributing/docs-templates/#recipes

For reference, the new recipe collection is also listed in a spreadsheet: https://docs.google.com/spreadsheets/d/1yNUQi_-3TsuDtE98hCIJQvLZ6X1YHlDmC-EJwOGxiHk/edit#gid=1252007048

marcysutton:
Closed with #15704

PR #15704, opened 2019-07-14
[docs][recipes] Querying data in pages

juhosa:
## Description

A recipe for querying data in pages.

There are four steps in the directions and they are all represented in one code block. I thought for a while to separate the steps in to different blocks but as they are fairly small and should be easy to understand I think it's okay to have them in a single one.

Happy to receive any feedback!

## Related Issues

This PR addresses the issue #14977


docs/docs/recipes.md
```diff
@@ -131,6 +131,54 @@ Showtime.
...
+
+You can use the `graphql`-tag to query data in your pages.
+
+#### Directions
+
+1. Import `graphql` from `gatsby`.
+
+2. Export a constant named `query` and set its value to be a `graphql` [tagged template](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) with the query between two backticks.
+
+3. Destruct `data` from props.
```
marcysutton:
```suggestion
3. Pass in `data` as a prop to the component.
```
I think the terminology could be simplified here so that people aren't thrown off by destructuring. Either that or we'll have to explain it.

docs/docs/recipes.md
```diff
@@ -131,6 +131,54 @@ Showtime.
...
+
+4. The `data` variable now holds the queried data with the expected shape.
```
marcysutton: If you end up going with the simplified terminology suggestion above, this could flow a little better:
```suggestion
4. The `data` variable which holds the queried data with the expected shape can be referenced in JSX to output HTML.
```

marcysutton:
This is excellent! It's getting really close. I made two suggestions for the steps to make sure they're clear to users of all skill levels. Let me know what you think!

juhosa:
I merged your suggestions @marcysutton as they made sense 👍🏽

PR #15705, opened 2019-07-14
[RFC] Add @memoize field extension

stefanprobst:
**DISCLAIMER**: I have not really thought this through -- would be cool to get feedback though if this makes sense.

This suggests a `@memoize` field extension. By default it will use `args` and the field value as cache key. This can be adjusted with the `fields` extension option.

Usage example:
```js
type AuthorYaml implements Node {
  name: String
  slug: String
  posts: [MarkdownRemark] @link(by: "frontmatter.authors.slug", from: "slug") @memoize(fields: ["slug"])
}
  frontmatter: Frontmatter
}
type Frontmatter {
  title: String
  authors: [AuthorYaml] @link(by: "slug") @memoize
}
```

### DIFF
--- a/packages/gatsby/src/schema/extensions/index.js
+++ b/packages/gatsby/src/schema/extensions/index.js
@@ -115,6 +115,68 @@ const builtInFieldExtensions = {
       }
     },
   },
+
+  memoize: {
+    name: `memoize`,
+    args: {
+      fields: [`String!`],
+    },
+    extend(options, fieldConfig) {
+      // TODO: Use `cache-manager` instance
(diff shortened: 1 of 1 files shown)


freiksenet:
This got me thinking that maybe we should cache results of such resolvers automatically. Practically a single build should have consistent node content, so why not cache it all, esp if it's used multiple times?

stefanprobst:
Yes if it works well we should definitely bake this into core. My thinking was that having this as an extension first means it's opt-in and we can collect some data how much difference it makes.

PR #15708, opened 2019-07-15
Recommend npx for one-time CLI use

cglong:
## Description

Update documentation for `gatsby-cli` to recommend using [npx](https://npm.im/npx) for one-time usage of the CLI, particularly when running the `new` command.

## Related Issues

None


cglong:
After opening this, I came across #12195 which encouraged the opposite approach of this PR. I've now opened #15712 to address my confusion 🙂

PR #15710, opened 2019-07-15
chore(showcase): add MAZECX

wladiston:

### DIFF
--- a/docs/sites.yml
+++ b/docs/sites.yml
@@ -6554,3 +6554,15 @@
   built_by: Stephan Weinhold
   built_by_url: https://stephanweinhold.com/
   featured: false
+- title: MAZECX
+  url: https://maze.cx
+  main_url: https://maze.cx
+  description: >
+    MAZECX is a digital agency designed to solve challenges by unleashing human potential to process more and execute faster using technology.
+  categories:
+    - Agency
+    - Design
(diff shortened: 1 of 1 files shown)


m-allanson:
Thanks @wladiston for submitting this site to the showcase! This is slightly unusual for us, but I'm going to request a change to the site before we can accept it. Would you be able to add a content warning that advises people about the flashing content?

The Gatsby team treat accessibility as a first class concern, and while we don't generally comment on site showcase designs I thought it's worth mentioning in this case. 

For reference check out [the seizures section of the Web Content Accessibility Guidelines (WCAG)](https://www.w3.org/TR/WCAG20/#seizure) and [this blog post announcing Gatsby's Accessibility Statement](https://www.gatsbyjs.org/blog/2019-04-18-gatsby-commitment-to-accessibility/).

As an example, displaying a banner or message _before_ any flashing content gives people the opportunity to leave if they want to. The wording could be. something like "This site may potentially trigger seizures for people with photosensitive epilepsy. Viewer discretion is advised." ([ref Wikipedia](https://en.wikipedia.org/wiki/Template:Seizure_warning)).

I'm going to close this PR for now, but would love to see a new PR from you if you're open to updating the site :purple_heart:

PR #15712, opened 2019-07-15
docs(gatsby-cli): Encourage installing CLI globally

### LINKED ISSUE #12195: Revert instructing people to use npx and npm scripts

cglong:
## Description

Removes warnings against installing `gatsby-cli` globally from docs, following discussion that this should instead be encouraged.

## Related Issues

Reverts part of #11556


wardpeet:
Let us keep mentioning adding scripts to package.json as this are shortcuts for people to use. People don't have to install gatsby-cli if they want to run a project.

We encourage them to install `gatsby-cli` as it will make it more comfortable to interact with gatsby in the future.

gillkyle:
Hey @cglong, thanks for updating this, like @wardpeet mentioned could we keep the comment about scripts in the `package.json` but just reword it so that the recommended method is the global installation? Maybe the note could go below to help convey that it's less important.

Would you have a chance to update that?

PR #15713, opened 2019-07-15
added footer

nguyenami:


wardpeet:

We want to keep the hello-starter as small as possible so adding a footer would only make it more complex.

Issue #15305, opened 2019-07-02
feat(gatsby/schema): Allow defining child relationships through metadata


freiksenet:
## Summary

Schema customization allows defining all kind of types and fields, except for the fields created through child/parent relationships. Defining child convenience fields isn't a good thing to do on parent level, therefore it needs to be done from a child point of view and we currently don't have any backreference mechanism in place. This issue defines API that transformers can use to specify child/parent relationships on type level, rather than node level.

### Basic example

```
// Generic parent children relationship
type SomeParent implements Node @infer {
  id
}

type SomeChild implements Node @infer @childOf(types: ['SomeParent']) {
  id
}

// Mime type based relationships

// File has all mimeTypes
type File implements Node @infer @mimeTypes {
  id
}

type SomeMarkdownType implements Node @infer @mimeTypes(types: ['text/markdown']) {
  id
}

type MarkdownRemark implements Node @infer @childOf(mimeTypes: ['text/markdown']) {
  id
}

type ImageSharp implements Node @infer @childOf(mimeTypes: ['image/jpeg']) {
  id
}
```

### Detailed explanation

Add `mimeTypes` extensions to types, that define possible mime types for that type.

Add `childOf` extension, that allows specifying that a type can be a child of either some type or a type with a particular mimeType. This allows for extensible parent/child relationship that aren't bound to `File`. We special case `File` to have all mimeTypes.

When we create child convenient fields, we add them not only based on nodes, but also based on the above system. @infer/@dontInfer start affecting this, with @dontInfer only adding explicit children.

pauleveritt:
Should this work with deletion of the parent?

I have a parent-child transformer working using:

```javascript
        createParentChildLink({
            parent: node,
            child: resourceNode
        });
```

When I comment out that line, and add `@childOf(types: ["Mdx"])` on my type, my queries fail to grab the parent (and don't delete the child.)

PR #15715, opened 2019-07-15
feat(gatsby): Allow explicitly defining parent-child relations with type extensions

stefanprobst:
This allows defining parent-child relations between types with the `@childOf` type extension, either by specifying parent type names, or by defining parent mime-types.
On types with the `@dontInfer` extension, only explicitly defined children fields will be added.

More details in #15305.

Fixes: #15305
Related: #15545


freiksenet:
@pieh made a good point that this breaks behavior of `@dontInfer`. We might need an argument for `@dontInfer` again and maybe a deprecation in v3 that will default to non-children inferred behaviour, with current behaviour still inferring children.

stefanprobst:
We can just leave these lines commented out, and wait until v3:
https://github.com/gatsbyjs/gatsby/blob/d9ccb9c639d97bc5cd7f3017c5faf3228d295e6e/packages/gatsby/src/schema/schema.js#L772-L777
and maybe add a deprecation warning?

I'm not so sure about adding a directive argument to `@dontInfer`, might make controlling inference behavior too complicated since we already have deprecated `noDefaultResolvers` arg? Would be easy to add though.

stefanprobst:
Now handles interfaces with `@nodeInterface` extension as well - tests should hopefully cover all scenarios.

freiksenet:
> and maybe add a deprecation warning?
> I'm not so sure about adding a directive argument to @dontInfer, might make controlling inference behavior too complicated since we already have deprecated noDefaultResolvers arg? Would be easy to add though.

You are right. Let's add deprecation warning when implicit children are added through `dontInfer` (when they weren't added explicitly).

Issue #15699, opened 2019-07-14
Find linked file node issue


zce:
## Description

<img width="1440" alt="屏幕快照 2019-07-14 21 53 19" src="https://user-images.githubusercontent.com/6166576/61184664-09643700-a683-11e9-88ef-28cff461749a.png">

### Steps to reproduce

create an image field by:

```js
createNodeField({ node, name: 'cover', value: node.frontmatter.cover })
```

In fact, `frontmatter.cover` and `fields.cover` are just aliases.

### Environment

  System:
    OS: macOS 10.14.5
    CPU: (4) x64 Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.16.0 - ~/.nvm/versions/node/v10.16.0/bin/node
    Yarn: 1.16.0 - ~/.config/yarn/bin/yarn
    npm: 6.9.0 - ~/.nvm/versions/node/v10.16.0/bin/npm
  Languages:
    Python: 2.7.16 - /usr/local/bin/python
  Browsers:
    Chrome: 77.0.3833.0
    Safari: 12.1.1
  npmPackages:
    gatsby: latest => 2.13.20 
    gatsby-image: latest => 2.2.6 
    gatsby-plugin-feed: latest => 2.3.4 
    gatsby-plugin-nprogress: latest => 2.1.2 
    gatsby-plugin-offline: latest => 2.2.4 
    gatsby-plugin-react-helmet: latest => 3.1.2 
    gatsby-plugin-sharp: latest => 2.2.7 
    gatsby-plugin-sitemap: latest => 2.2.3 
    gatsby-plugin-typography: latest => 2.3.2 
    gatsby-remark-autolink-headers: latest => 2.1.3 
    gatsby-remark-copy-linked-files: latest => 2.1.3 
    gatsby-remark-images: latest => 3.1.6 
    gatsby-remark-prismjs: latest => 3.3.3 
    gatsby-remark-responsive-iframe: latest => 2.2.4 
    gatsby-remark-smartypants: latest => 2.1.2 
    gatsby-source-filesystem: latest => 2.1.5 
    gatsby-transformer-remark: latest => 2.6.5 
    gatsby-transformer-sharp: latest => 2.2.3 
    gatsby-transformer-yaml: latest => 2.2.3

zce:
[...]

PR #15720, opened 2019-07-15
fix(gatsby): Track nodes when mutated

stefanprobst:
After a field is added to a node with `createNodeField`, we should re-track inline objects.

Fixes:  #15699


stefanprobst:
Failing tests look unrelated, failure with NPM registry?

pieh:
> Failing tests look unrelated, failure with NPM registry?
Yeah, seems so - I restarted workflow - those tests don't fail anymore, so it probably was temporary registry/CI problem

pieh:
I restarted just CircleCi tests as I have no idea how to do that for azure and seems to work

zce:
When do you plan to merge?

pieh:
Oh, sorry @zce. It was suppose to automerge, but seems like some recently required tests didn't even run. Let me update PR be syncing master and if nothing breaks, it will automerge.

Issue #15596, opened 2019-07-10
[www] Move design tokens to NPM package


fk:
… so we can easily reuse them for other sites/projects.

I understand we ideally want to keep this package "dependency-free", prerequisite for that is to refactor https://github.com/gatsbyjs/gatsby/blob/master/www/src/utils/tokens/space.js, which is currently put through typography.js' `rhythm` in https://github.com/gatsbyjs/gatsby/blob/97522c172d9441dfe644a0e2b3fe84d2e2aea1d1/www/src/utils/presets.js#L26, to px/rem values.

We might want to take a closer look at adjusting the setup to be in line with the ideas of `theme-ui` as much as possible—not sure how off we currently are. _Maybe, just maybe_ it's worth thinking through about the tradeoff of putting .org on `theme-ui` now instead of later on…tbh I'm not really sure what I'm talking about here. Related: #8789

Other thoughts:

- should https://github.com/gatsbyjs/gatsby/blob/master/www/src/utils/tokens/sizes.js and https://github.com/gatsbyjs/gatsby/blob/master/www/src/utils/tokens/z-indices.js be part of that package? At least a bunch of values here are quite .org `www`-specific.
- maybe the same is true for https://github.com/gatsbyjs/gatsby/blob/master/www/src/utils/tokens/breakpoints.js?

/cc @jxnblk @greglobinski @KyleAMathews

fk:
PR that detaches `space` tokens from `typography.js` here: https://github.com/gatsbyjs/gatsby/pull/15728

fk:
Done, thanks @DSchau and @wardpeet, and @greglobinski, and @jxnblk, and @pieh, and @sidharthachatterjee <3

PR #15728, opened 2019-07-15
chore(www): Convert `utils/tokens/space` to pixels

fk:
…and detach them from `typography.js` and its `rhythm`.
Convert them to `rem` in `utils/presets.js`.

Context: Prerequisite for #15596


KyleAMathews:
BTW — what's the status of the discussion to move to theme-ui?

PR #15731, opened 2019-07-15
chore(starters): add gatsby-starter-nagui

kimnagui:
## Description

Added gatsby-starter-nagui starter.


pieh:
Hey there is similar PR ( https://github.com/gatsbyjs/gatsby/pull/15733 ) - we should close one of those I think

kimnagui:
> Hey there is similar PR ( #15733 ) - we should close one of those I think

Sorry, I forgot close.

Issue #15686, opened 2019-07-13
Official themes should include "Powered by Gatsby" in footer


KyleAMathews:
Or "Gatsby Inside" logo if we want to go with that.

polishedwp:
I will be glad to work on this one.

sidharthachatterjee:
CC @fk in case we need a logo for this

polishedwp:
What about the "starters?"  Right now their footer's state "[Built with Gatsby](https://github.com/gatsbyjs/gatsby/blob/ef4eea27d6570573ba2cde10d37899ca67b55b72/starters/blog/src/components/layout.js#L66}."
Do we want the themes to match them for consistency?

nguyenami:
@polishedwp do you mind sharing one? I would like to do one for the gatsby-starter-hello-world.

polishedwp:
So, I installed each of the starters and the only starter that displays the 404 error in the terminal is the hello-world starter. 

The 404 message appears in each of the default theme starters including the new workspace theme starter.

PR #15738, opened 2019-07-15
feat: Added Powered by Gatsby note to theme footers.

eclectic-coding:
Fixes: #15686 

## Description

Added "Powered by Gatsby" to the footer of `gatsby-theme-blog,` and created a footer for `gatsby-theme-notes.`
If the preference is to include a new image @fk I will be glad to do that. 

## Related Issues

Refer to my previous note, it might be a good idea to change the footers in the default Gatsby starters for consistency across the Gatsby Brand.


fk:
Thanks @eclectic-coding, and sorry I forgot about this. I think this is a great solution, no need for an image.

Nitpick: We should use "Twitter" and "GitHub". Will get a PR up for that.

fk:
^ Done: #16017

Issue #15101, opened 2019-06-25
Gatsby's matchPath still does not work as expected after v2.9.0


wiziple:
## Description

👋 I'm filing an ongoing issue on behalf of `gatsby-plugin-intl` users since some people are keep asking me about the reason.
This issue happens since gatsby@2.9.0 and still happens on the latest version of gatsby@2.10.4.

The `matchPath` does not work as expected or at least it used to be. There was a related fix to solve some issues on the gatsby@2.10.4, but it still has some issues described below.

### Steps to reproduce

In order to describe as simply as possible, I'm going to use an official example and document.
https://www.gatsbyjs.org/docs/creating-prefixed-404-pages-for-different-languages/

I cloned `gatsby-starter-default` and followed the simple steps on the instruction above, then created `pages/en/hello.js` and `pages/de/hello.js` files.

[Repo] https://github.com/wiziple/gatsby-example-matchpath
[Netlify Build] https://sleepy-austin-39aafc.netlify.com

### Expected result

https://sleepy-austin-39aafc.netlify.com/en/hello/ should render `pages/en/hello.js`
https://sleepy-austin-39aafc.netlify.com/de/hello/ should render `pages/de/hello.js`

### Actual result

https://sleepy-austin-39aafc.netlify.com/en/hello/ renders `pages/en/404.js`
https://sleepy-austin-39aafc.netlify.com/de/hello/ renders `pages/de/404.js`

### Environment

  System:
    OS: macOS 10.14.3
    CPU: (4) x64 Intel(R) Core(TM) i5-4258U CPU @ 2.40GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.16.0 - /usr/local/bin/node
    Yarn: 1.15.2 - /usr/local/bin/yarn
    npm: 6.9.0 - /usr/local/bin/npm
  Languages:
    Python: 2.7.10 - /usr/bin/python
  Browsers:
    Chrome: 75.0.3770.100
    Safari: 12.0.3
  npmPackages:
    gatsby: ^2.10.4 => 2.10.4 
    gatsby-image: ^2.2.1 => 2.2.3 
    gatsby-plugin-manifest: ^2.2.0 => 2.2.0 
    gatsby-plugin-offline: ^2.2.0 => 2.2.0 
    gatsby-plugin-react-helmet: ^3.1.0 => 3.1.0 
    gatsby-plugin-sharp: ^2.2.1 => 2.2.1 
[...]

PR #15762, opened 2019-07-15
fix(gatsby): fix matchpath prioritisation

wardpeet:
## Description
I rewrote the matchpath functionality in 2 separate functions. We first try to get page-data from the current route. If it returns a 404 we try to fetch a matchPath route.

Todo:
- [x] Have another look at the code 👀 
- [x] Make sure 404 still works
- [x] write tests

I've tested the match path logic in a unit test but a e2e test would be great as well. I'm leaving that in the hand of @sidharthachatterjee :smile:

I'll probably do another PR after this to cleanup the loader.js file a bit to make it more readable.

## Related Issues
#15101


packages/gatsby/cache-dir/loader.js
```diff
@@ -242,22 +258,37 @@ export class BaseLoader {
...
     if (!this.shouldPrefetch(pagePath)) {
       return false
     }
+
+    let hasPrefetched = true
+
     // Tell plugins with custom prefetching logic that they should start
     // prefetching this path.
-    if (!this.prefetchTriggered.has(pagePath)) {
+    if (!this.prefetchTriggered.has(realPath)) {
```
wardpeet:
```suggestion
    if (!this.prefetchTriggered.has(pagePath)) {
```

packages/gatsby/cache-dir/loader.js
```diff
@@ -242,22 +258,37 @@ export class BaseLoader {
...
       this.apiRunner(`onPrefetchPathname`, { pathname: pagePath })
-      this.prefetchTriggered.add(pagePath)
+      hasPrefetched = false
+      this.prefetchTriggered.add(realPath)
```
wardpeet:
```suggestion
      this.prefetchTriggered.add(pagePath)
```

packages/gatsby/cache-dir/loader.js
```diff
@@ -242,22 +258,37 @@ export class BaseLoader {
   }
 
   prefetch(pagePath) {
```
wardpeet: need to look at prefetch

PR #15763, opened 2019-07-15
chore(gatsby-theme-blog): Use deepmerge instead lodash.merge

### LINKED ISSUE #15692: [gatsby-theme-blog] use deepmerge instead of lodash.merge
It’s a lot smaller — 3.7kb vs. 0.664 kb (minified and gzipped)

https://github.com/gatsbyjs/gatsby/blob/4a10c54b2ba7866da541257eadbed6c7e206f2f5/themes/gatsby-starter-blog-theme/src/gatsby-theme-blog/gatsby-plugin-theme-ui/colors.js#L1

dantehemerson:
## Description
Use deepmerge instead lodash.merge.

## Related Issues

Fixes #15692


dantehemerson:
Change the `gatsby-theme-blog` packages as suggested by @KyleAMathews .
I'd like you to give me a review about what I did.

However as for the package version. I'm confused since `gatsby-starter-blog-theme` was using version `v1.0.0` but the `gatsby-theme-blog` package was still in version `v0.0.11`.

Looking at npm I saw that there is also another version `v.0.0.12`. I do not know how they are handling this. If you could give me some explanation I would really appreciate it.
https://www.npmjs.com/package/gatsby-theme-blog
Screen Shot 2019-07-15 at 4 05 29 PM

themes/gatsby-theme-blog/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "gatsby-theme-blog",
-  "version": "0.0.11",
+  "version": "0.0.13",
```
sidharthachatterjee: Versioning is handled by `lerna` so this is unnecessary and can be removed

themes/gatsby-starter-blog-theme/package.json
```diff
@@ -9,7 +9,7 @@
   },
   "dependencies": {
     "gatsby": "^2.13.23",
-    "gatsby-theme-blog": "^1.0.0",
+    "gatsby-theme-blog": "^0.0.13",
```
sidharthachatterjee: We update `package.json` and lock files automatically via Renovate so this is unnecessary as well

dantehemerson:
Thanks for the explanation, @sidharthachatterjee

PR #15764, opened 2019-07-15
feat(docs): Add docs for Theme UI

### LINKED ISSUE #15746: [docs][themes] add a reference guide on using theme-ui
## Description

Add a reference guide on "using theme-ui" to `/docs/themes`.

## Related

#15382

jxnblk:
## Description

This adds some high-level docs for using Theme UI with Gatsby. It's loosely based on https://theme-ui.com/getting-started, but geared more towards using the library specifically with Gatsby.

**Questions**

- Is `Docs/Reference Guides/Styling/CSS Libraries and Frameworks` the correct place to add this in the navigation?
- Is this the right level of detail for this doc?

## Related Issues

Fixes #15746


amberleyromo:
> Is Docs/Reference Guides/Styling/CSS Libraries and Frameworks the correct place to add this in the navigation?

This is an interesting question. We had been thinking about it in the themes perspective, but this is probably right. And we can link to the guide from the themes section. @marcysutton 	?

docs/docs/theme-ui.md
```diff
@@ -0,0 +1,260 @@
...
+  // highlight-end
+}
+```
+
+Feel free to include as many additional values as you'd like in your theme.
+Read more about theming in the [Theme UI documentation](https://theme-ui.com/theming).
+
+## Adding styles to elements
+
+Theme UI uses a custom [JSX pragma][] comment to configure the create element function used in JSX syntax.
```
amberleyromo: Can you include some verbiage about what a custom pragma is, and why you would need it?

www/src/data/sidebars/doc-links.yaml
```diff
@@ -275,6 +275,8 @@
                   link: /docs/tailwind-css/
                 - title: Using Bulma
                   link: /docs/bulma/
+                - title: Theme UI
```
amberleyromo:
```suggestion
                - title: Using Theme UI
```

docs/docs/theme-ui.md
```diff
@@ -0,0 +1,260 @@
...
+    text: "#222",
+    primary: "tomato",
+  },
+})
+```
+
+## Styling MDX content
+
+Theme UI includes a way to style elements in MDX documents without the need to add global CSS to your site.
+This is useful for authoring Gatsby themes that can be installed alongside othere themes.
```
gillkyle:
```suggestion
This is useful for authoring Gatsby themes that can be installed alongside other themes.
```

www/src/data/sidebars/doc-links.yaml
```diff
@@ -275,6 +275,8 @@
                   link: /docs/tailwind-css/
                 - title: Using Bulma
                   link: /docs/bulma/
+                - title: Using Theme UI
```
marcysutton: > We had been thinking about it in the themes perspective, but this is probably right. And we can link to the guide from the themes section

One basic question: does Theme UI to scenarios where themes aren't involved? That might determine where it should live. But my intuition is that it's fine–it would probably be more intuitive to find it here from a styling perspective, with links from the themes docs.

jxnblk: I think it does apply to other scenarios – the only part that's explicitly tied to themes is the plugin. If we'd rather put together Gatsby theme-specific docs for using Theme UI, I could try to write a doc that's more geared towards that use case.

docs/docs/theme-ui.md
jxnblk: I tried to write a little more about how the JSX pragma comments work, but let me know if you have any ideas on how to make this better

docs/docs/theme-ui.md
```diff
@@ -0,0 +1,264 @@
+---
+title: Theme UI
+---
+
+[Theme UI][] is a library for styling React apps with user-configurable design constraints.
+It allows you to style any component in your application with typographic, color, and layout values defined in a shared theme object.
+Theme UI is currently used in our official Gatsby themes,
```
marcysutton:
```suggestion
Theme UI is currently used in Gatsby's official themes,
```
Updating per the Gatsby Style Guide, which recommends focusing on the reader

docs/docs/theme-ui.md
```diff
@@ -0,0 +1,264 @@
...
+  // highlight-end
+}
+```
+
+Feel free to include as many additional values as you'd like in your theme.
+Read more about theming in the [Theme UI documentation](https://theme-ui.com/theming).
+
+## Adding styles to elements
+
+Theme UI uses a custom [JSX pragma][] comment to add support for Emotion's `css` prop,
```
marcysutton: Since this term will be new to many people, is there a short way to summarize what it is? e.g. "a custom JSX pragma comment, a fancy type of function" or whatever the thing is? (i couldn't immediately tell from reading the Babel doc)

docs/docs/theme-ui.md
```diff
@@ -0,0 +1,264 @@
...
+which can be used to style elements.
+By default JSX compiles to the `React.createElement` function.
```
marcysutton: I feel like these two sentences about JSX would be great as an advanced callout. @amberleyromo your idea about little graphic icons in the docs would fit well here. But maybe there's something we can add in the meantime, like a note? My concern is that people will come to this section and think "this is too advanced for me" and give up early.

docs/docs/theme-ui.md
```diff
@@ -0,0 +1,264 @@
...
+      color: theme.colors.background,
+      backgroundColor: theme.colors.primary,
+    })}
+  >
+    {props.children}
+  </header>
+)
+```
+
+Theme UI includes the [`sx` prop][], which can be used as a replacement for the `css` prop.
```
marcysutton: This section could be another advanced note like I mentioned above.

docs/docs/theme-ui.md
```diff
@@ -0,0 +1,264 @@
...
+The `sx` prop allows values from the theming context to be referenced with a terser syntax.
+
+[`sx` prop]: https://theme-ui.com/sx-prop
+
+The example from above can be rewritten like this:
```
marcysutton: I'm hesitant to introduce the shorthand into these docs, because it's _so much_ to learn at once and we want this to  be as easy as possible. Maybe making it clear that this is completely optional would help, with an "Advanced usage" heading or something. The way it's worded kinda makes it sound like everyone would want to rewrite it that way, when it could introduce more cognitive overhead for people following along (even with the "can" part).

docs/docs/theme-ui.md
```diff
@@ -0,0 +1,264 @@
...
+      // can reference other values in the theme
+      fontFamily: "heading",
+      fontWeight: "heading",
+      lineHeight: "heading",
+      marginTop: 0,
+      marginBottom: 3,
+    },
+    a: {
+      color: "primary",
+      ":hover": {
```
marcysutton:
```suggestion
      ":hover, :focus": {
```
always good to remind! :)

www/src/data/sidebars/doc-links.yaml
marcysutton: I think both could be useful. We could start with this and some links, and make further improvements if necessary?

docs/docs/theme-ui.md
jxnblk: Yeah, this part of JSX is not the easiest part to explain. I added more in this section based on @amberleyromo's suggestion, but I'm not sure how detailed to go here. I think a special callout would make sense – do we have a way to do that today?

docs/docs/theme-ui.md
jxnblk: In the [Theme UI docs](https://theme-ui.com/sx-prop) it introduces the `sx` prop first, and doesn't mention how to use the `css` prop until the end. The `sx` prop is the preferred way to add styles (i.e. without the `sx` prop, Theme UI isn't doing much beyond what you can do with Emotion), and `css` is sort of an escape hatch. I moved the `css` prop section before in an attempt to explain how it works for people who are familiar with Emotion, but maybe that could be left out of this doc entirely. Curious to hear what others think

docs/docs/theme-ui.md
marcysutton: Here's a suggestion based on [an explainer I read](https://jasonformat.com/wtf-is-jsx/): 

```suggestion
Theme UI uses a custom [JSX pragma][], a per-file or per-function comment to tell [Babel](/docs/glossary#babel) the name of Emotion's `css` prop.
```

docs/docs/theme-ui.md
```diff
@@ -0,0 +1,264 @@
...
+Theme UI uses a custom [JSX pragma][] comment to add support for Emotion's `css` prop,
+which can be used to style elements.
+By default JSX compiles to the `React.createElement` function.
+When you add the `/** @jsx jsx */` comment to the top of a file, JSX will compile to the function name in the comment instead.
+For example, `<button />` would compile to `jsx('button')`.
+Theme UI's `jsx` function still renders React elements but lets you use the `css` prop for styling.
+
+[jsx pragma]: https://babeljs.io/docs/en/babel-plugin-transform-react-jsx#pragma
+
+The following example uses the `css` prop to add styles to an element.
```
marcysutton: Listing these two code examples first in the Adding Styles to Elements section and then explaining the advanced pragma stuff could strike a nice balance between providing information and not overwhelming people who just want to read code and try stuff out. Then, we could add an `### Advanced tips` heading and the rest of the explanation. How does that sound to you?

docs/docs/theme-ui.md
marcysutton: Perhaps showing code might be the most effective. If it's a necessary concept, can you try minimal code blocks to show the difference between the `React.createElement` and pragma output? Similar to line 128 with `jsx('button')`, but called out a bit more. If line 129 about Theme UI's `jsx` function could be shown with code, that would help too.

docs/docs/theme-ui.md
marcysutton: With https://github.com/gatsbyjs/gatsby/pull/15764/files#r305564999, this should be fine as-is!

docs/docs/theme-ui.md
jxnblk: I'm taking another look at this now, and I feel like going into how JSX works is maybe a little too advanced for the concepts this doc is trying to introduce. I think with the first line reworked a little, we could avoid explaining JSX and perhaps we could link to https://jasonformat.com/wtf-is-jsx/ instead of going into too much detail here

docs/docs/theme-ui.md
amberleyromo: I think if `sx` is the preferred pattern we want to promote, then it should be prioritized in this doc.

docs/docs/theme-ui.md
amberleyromo: So I love the article you've linked here, but I think it _really_ doesn't solve the problem of explaining this on a more approachable level to beginner folks. We would need to explicitly mark that as an advanced read.

www/src/data/sidebars/doc-links.yaml
amberleyromo: I think separate Gatsby theme-specific docs for using Theme UI would be helpful -- particularly for how using `gatsby-plugin-theme-ui` affects usage of multiple themes composed in a single project.

jxnblk:
I've tried to completely rework and strip back the `sx` prop usage section based on chatting with @amberleyromo, so this will need some fresh eyes - happy to work in any links or *advanced* sections that would make sense, but I think I was originally getting too deep into the implementation here.

amberleyromo:
@marcysutton a few more notes on the update: @jxnblk 	and I discussed trying to find a balance between enough to make sense, but not so much as to overwhelm. I started pulling together some introductory explanatory language for jsx pragma, and there's a lot we can do on the beginner/intermediate/advanced levels there. 

We talked about that potentially being a blog post of its own. I'm not certain it makes sense as a standalone doc.

PR #15766, opened 2019-07-15
chore(www): Add a nightly job to build the .org site

m-allanson:
## Description

This is the first step towards adding a daily test for building the .org site. It shouldn't interfere with any of the existing tests. Env vars are set in the CircleCI web UI. This won't get reported anywhere (except within Circle CI), and doesn't deploy anything yet.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -222,6 +222,17 @@ jobs:
       - run: git config --global user.email "admin@gatsbyjs.com"
       - run: sh ./scripts/publish-starters.sh "starters/* themes/gatsby-starter-*"
 
+  build_www:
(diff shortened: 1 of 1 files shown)


wardpeet:
Nice! should we use gatsby-dev-cli and copy things over?

m-allanson:
> Nice! should we use gatsby-dev-cli and copy things over?

Yeah I'm going to experiment a bit with this - my first plan is to just run this every 24hrs. Then maybe look into running it for every PR later.

m-allanson:
Failing tests fixed in https://github.com/gatsbyjs/gatsby/pull/15776, merging this.

PR #15768, opened 2019-07-15
chore(.github): PR bot logs

gillkyle:
## Description

chore: add logs for debugging production runs of the pr bot

The PR bot is returning empty messages for the learning team when run from the actions container. I suspect it's an issue with auth tokens though the `GITHUB_TOKEN` that is baked into GitHub actions _should_ have the necessary permissions. I added some logs as a rudimentary check to see if review requests are coming through and also see that things are running in the correct order and there isn't an async problem somewhere.


gillkyle:
YOLO merged this so I can debug when it runs on the schedule tonight

PR #15778, opened 2019-07-16
Add packages/gatsby-design-tokens

fk:
This is a draft to get feedback on how to add a new package `gatsby-design-tokens`.
It solves #15596 in the simplest way possible, exporting just the bare-bones design tokens, leaving e.g. conversion of the token pixel values to `rem` like we e.g. require for the current `www` setup.

It currently contains changes to `www` so I could test if the new package is working…
@sidharthachatterjee Am I right assuming that I really want to provide two separate PRs—one just adding `gatsby-design-tokens`, which then would be published, and another one switching `www` to consume the package after it has been published to NPM?


fk:
Nah. 😅

PR #15779, opened 2019-07-16
feat(gatsby-design-tokens): add gatsby-design-tokens utility

fk:
This adds a new package `gatsby-design-tokens`, containing the…design tokens 😜 currently in use for `www` (`www/src/utils/tokens`).

It exports just the bare-bones design tokens, leaving e.g. conversion of pixel values to `rem`—like we require for the current `www` setup—up to the user.

It also includes tokens that are quite specific to the `www` setup—`sizes` and `z-indicies`.


packages/gatsby-design-tokens/src/transition.js
```diff
@@ -0,0 +1,10 @@
+export default {
+  curve: {
+    default: `cubic-bezier(0.4, 0, 0.2, 1)`,
+  },
+  speed: {
+    default: `250ms`,
+    fast: `100ms`,
+    slow: `350ms`,
```
greglobinski: could we add another one 
`slower: 500ms`
we use it a lot on Cloud, I mean 500ms, the name is less important, if you have better name, :+1:

packages/gatsby-design-tokens/src/z-indices.js
```diff
@@ -0,0 +1,9 @@
+export default {
+  feedbackWidget: 2,
```
greglobinski: maybe 
```
widget: 2
```
it could be used for other widgets in the future too

packages/gatsby-design-tokens/src/z-indices.js
```diff
@@ -0,0 +1,9 @@
...
+  navigation: 5,
+  banner: 10,
+  modal: 10,
+  sidebar: 10,
+  sidebarToggleButton: 20,
```
greglobinski: In this case `sidebarToggleButton` also would like to see more general name, but have no idea for it, so will have to live if it will stay, as it is :)

packages/gatsby-design-tokens/src/index.js
```diff
@@ -0,0 +1,33 @@
+import borders from "./borders"
```
wardpeet: please use the export syntax so we can do treeshaking.
```suggestion
export { default as borders } from "./borders"
```

fk: Done (I think 😬): 47ab9d8

packages/gatsby-design-tokens/src/transition.js
fk: Done in 074716c

packages/gatsby-design-tokens/src/z-indices.js
fk: Renamed in 273bb2e

I'm not super sure that all widgets will work with a z-index of 2 though.
Want to get together and work through some solid definitions for "depth/elevation" (`z-index`)?

packages/gatsby-design-tokens/src/z-indices.js
fk: Renamed it to `floatingActionButton` (which it indeed is ;-)) in f599bd6.

packages/gatsby-design-tokens/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "gatsby-design-tokens",
+  "version": "1.0.0-alpha.1",
```
DSchau: @fk can I release this as 1.0.0 or do you want a pre-release first?

packages/gatsby-design-tokens/src/z-indices.js
greglobinski: the shorter version is just less specific, but sure it could happen that it will not work in all cases. But you are right, maybe we should talk about something truly general like `level0`, 'level1`

packages/gatsby-design-tokens/package.json
fk: I guess it doesn't _really_ matter, but I expect some changes/cleanup to be happening still to call this a true 1.0.0. Any worries about/drawbacks in consuming this as `1.0.0-alpha.1`?

packages/gatsby-design-tokens/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "main": "index.js",
+  "author": "Florian Kissling",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/gatsbyjs/gatsby.git",
+    "directory": "packages/gatsby-design-tokens"
+  },
+  "scripts": {
+    "build": "babel src --out-dir . --ignore **/__tests__",
```
wardpeet: should we build to a dist folder? This isn't a gatsby-plugin package so we can add a `files: ["dist/"]` instead of doing the npmignore dance.

packages/gatsby-design-tokens/package.json
DSchau: @fk not at all — this hasn’t been published whatsoever right? I can just release it as a pre-release!

packages/gatsby-design-tokens/package.json
DSchau: @wardpeet I’ll do this in a follow-up. Really it doesn’t _matter_ 🤷‍♂

wardpeet: yeah, it's easier to only have to manage gitignore instead of gitignore and npmignore when not setup properly. That's all but like you said it doesn't really matter

DSchau:
Published as `gatsby-design-tokens`@`0.0.1`

fk:
<3 <3 <3 PR for `www` incoming!

packages/gatsby-design-tokens/package.json
fk: Sounds good to my n00b ears ;)

packages/gatsby-design-tokens/src/z-indices.js
fk: 👍

fk: Just realized I messed up the renaming of the two `zIndices` things—did them in `utils/tokens/z-indices` … 🤦‍♂ https://github.com/gatsbyjs/gatsby/commit/dded37dd028ebb311dbe6930b7b2633148fcf612#diff-d3cc05df2a27981e190be3123cae8e92 … ah well. `0.0.2` incoming.

PR #8379, opened 2018-09-20
feat(gatsby): selective cache invalidation

DSchau:
Fixes #8324

This PR introduces a cache removal mechanism where we don't remove the cached results from `createRemoteFileNode`

## To-do

1. [x] Add an integration test validating the functionality
1. [x] Add more unit tests validating the functionality
1. [x] Test with real-world projects (e.g. gatsby-image, reactjs.org, gatsbyjs.org, etc.)

Issue #8324, opened 2018-09-19
Don't clear downloaded files from createRemoteFileNode calls when we clear cache on package/code changes


KyleAMathews:
Right now we clear the entire `.cache` folder when gatsby-node.js changes or a user upgrades their packages. Which cause a lot of slowdowns on the next run of `gatsby develop` as then any plugins that pull remote files have to pull the same files again.

We should start moving to more selective cache clearing and _not_ clear data that we know is immutable.

PR #5862, opened 2018-06-12
[v2] Move onPreBootstrap to run before the cache is loaded

axe312ger:
I worked on https://github.com/axe312ger/gatsby-plugin-netlify-cache to provide a caching plugin for Netlify. It currently does cache `public` and `.cache` and restores that cache `onPreBootstrap`.

But this does not really work that way since Gatsby does clean up the cache and init the database BEFORE `preBootstrap`.

This PR adds a new hook which gets triggered basically before Gatsby does anything. That way my plugin could restore everything from the custom cache directory and Gatsby can still do its cleanup magic.

What do you think?

```
10:52:54 AM: > gatsby build
10:52:56 AM: success delete html and css files from previous builds — 0.047 s
10:52:56 AM: success open and validate gatsby-config — 0.006 s
10:52:56 AM: success copy gatsby files — 0.015 s
10:52:57 AM: Found 57 cached files for public directory with 1 files.
10:52:57 AM: Found 31 cached files for .cache directory with 26 files.
10:52:57 AM: Netlify cache restored
10:52:57 AM: success onPreBootstrap — 0.850 s
```

See https://github.com/axe312ger/gatsby-plugin-netlify-cache/issues/3

PR #15783, opened 2019-07-16
add gatsby-transformer-video (beta available)

axe312ger:
Kapture 2020-02-14 at 23 38 27

This integrates [fluent-ffmpeg](https://github.com/fluent-ffmpeg/node-fluent-ffmpeg) with Gatsby 🎥 

* A beta is already released: `npm i gatsby-transformer-video`
* Docs: https://github.com/gatsbyjs/gatsby/tree/gatsby-transformer-video/packages/gatsby-transformer-video
* Example: https://github.com/gatsbyjs/gatsby/tree/gatsby-transformer-video/examples/using-gatsby-transformer-video

## Features

- [x] Source: Works with nodes from `gatsby-source-filesystem` and `gatsby-source-contentful`
- [x] Defaults optimized for small files with decent quality and quick and seamless streaming
- [x] Supported codecs: h264, h265, VP9, WebP & gif
- [x] Several parameters to tweak the output: maxWidth/maxHeight, overlay/watermark, saturation, duration, fps, ...
- [x] Create video conversion profiles. Create a converter function using `fluent-ffmpeg` to unlock all FFMPEG features.
- [x] Downloads binaries from official sources when FFMPEG is not (properly) installed
- [x] Take screenshots at any position of the video
- [x] Multiple cache layers to prevent accidental reconversion (Discussion see below)
- [x] Example project to demonstrate the usage

### Todos:

- [ ] Find somebody with extended FFMPEG knowledge and verify our default profile settings: https://github.com/gatsbyjs/gatsby/tree/gatsby-transformer-video/packages/gatsby-transformer-video/src/profiles
- [x] Videos converted via the h265 profile work fine on iOS & in VLC, but Safari & Quicktime struggle to play the files
- [ ] Overlays have fixed size, FFMPEG can resize them to better fit into the video. We should support this
- [ ] We currently cache into `.node_moduels/.cache` and queue our FFMPEG executions on our own. This is just a workaround as Gatsby has no system to queue such time consuming transformations. We as community should implement an API that allows all plugins play well next to `gatsby-transformer-sharp`,  this video transformer & others.
- [ ] Write tests
- [ ] ... There is more, look in the discussion below...


me4502:
Generally GIFs should never be used on the web - https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/replace-animated-gifs-with-video/

DSchau:
@me4502 while I'd agree, exposing `gif` nicely creates an experience where video elements are _preferred_ and we can fall back to `gif` as a _last resort. (like in a picture/source stack as the example shows).

So... noted, but I think it's reasonable to provide it so we could create e.g. `gatsby-video` component or something.

axe312ger:
Currently thinking of providing one field per codec. Opens up flexibility and would remove that preview field.

That way users could generate a gif which way less fps as the other previews. Or just simply provide a jpg as preview.

yarn.lock
```diff
@@ -276,6 +276,18 @@
     "@babel/helper-plugin-utils" "^7.0.0"
     "@babel/helper-replace-supers" "^7.2.3"
 
+"@babel/helper-create-class-features-plugin@^7.5.0":
```
KyleAMathews: we don't check in yarn.lock files for packages

packages/gatsby-transformer-video/src/ffmpeg.js
```diff
@@ -0,0 +1,431 @@
...
+
+    if (ffmpegPath) {
+      ffmpeg.setFfmpegPath(ffmpegPath)
+    }
+    if (ffprobePath) {
+      ffmpeg.setFfprobePath(ffprobePath)
+    }
+  }
+
+  cacheContentfulVideo = async ({
```
KyleAMathews: shouldn't this be using createRemoteFileNode?

packages/gatsby-transformer-video/README.md
```diff
@@ -0,0 +1,86 @@
...
+- Create animated previews as mp4, animated webp and animated gif.
+- Add watermarks to you videos
+- More?
+
+## Installation
+
+```sh
+npm i gatsby-transformer-video
+```
+
```
robertcoopercode: The usage of `fluent-ffmpeg` requires that `ffmpeg` be installed on your machine. Should there be a note regarding this installing?

axe312ger: Yes. We might even notify the user on CLI when installing if ffmpeg is missing?

axe312ger:
Did not try it out yet but the following should speed up the playback start delay:

> Add `-movflags +faststart`. This will relocate the moov atom from the end of the file to the beginning. This will allow the video to begin playback while it is still being downloaded. Otherwise the whole video must be completely downloaded before it can begin playing.

Source: https://stackoverflow.com/a/25573078

axe312ger:
FYI: planning to work on this on christmas holidays :)

me4502:
I personally have been pre-encoding all videos with the recommendations from here, https://gist.github.com/mikoim/27e4e0dc64e384adbcb91ff10a2d3678

These seem to perform super well, but the benefits of this plugin to allow better sizing / multiple formats will be amazing! :)

packages/gatsby-transformer-video/src/gatsby-node.js
```diff
@@ -0,0 +1,209 @@
...
+  GraphQLFloat,
+} from "gatsby/graphql"
+import { ensureDir } from "fs-extra"
+
+import FFMPEG from "./ffmpeg"
+
+exports.setFieldsOnGraphQLNodeType = (
+  { type, store },
+  { ffmpegPath, ffprobePath }
+) => {
```
muescha: on the page [Building the GqlType](https://www.gatsbyjs.org/docs/schema-gql-type/) is an outdated notification:

> This documentation isn’t up to date with the latest [schema customization changes](https://www.gatsbyjs.org/docs/schema-customization).
>
> Outdated areas are:
> ...
> `setFieldsOnGraphQLNodeType` has been deprecated due to the new `createTypes` action
> ...

axe312ger: Thanks will take this in to account

axe312ger:
So my idea is to refactor this till end of the year, including all the given feedback and especially:

> Currently thinking of providing one field per codec. Opens up flexibility and would remove that preview field.

So instead of a `video`  && `videoPreview` field with the codecs as subfields, I'd create a field per codec/file-type, supporting codec/file-format related attributes + easy support for fragments.

To use the current version of the plugin you would do something similar like:

```graphql
fragment BlogPostData on BlogPost {
  ...
  mediaFieldWithVideoData {
    # convert source video into optimized for web version with default settings
    video {
      h264
      h265
    }
    # Create a preview version of the video with reduced dimension, speed up and very limited frames per second
    videopreview(width: 600, fps: 4, duration: 3) {
      mp4
      webp
      gif
    }
  }
}
```

While the refactored way I propose would allow you way more fine granular configuration. Which is necessary as video conversion is complex and the settings depend on a lot of content and project specific factors.

```graphql
fragment BlogPostData on BlogPost {
  ...
  mediaFieldWithVideoData {
    videoH264 (maxBuffer: '2M', ...) {
      path
      size
    }
    videoH265 (maxBuffer: '1M', fancyH265Setting: 'bar', ...) {
      path
      size
    }
    previewMp4: videoH264 (desaturate: true, width: 600, fps: 8, duration: 3) {
      path
    }
    # Note that we can give the gif way smaller dimensions to reduce preview file size for clients without mp4 support
    previewGif: videoGif (desaturate: true, width: 300, fps: 4, duration: 3) {
      path
    }
    ...
  }
}
```

muescha:
> muescha unassigned KyleAMathews 2 days ago

sorry Kyle - this must done by mistake - i don't remember doing something like this :( 
i reassigning you

packages/gatsby-transformer-video/src/ffmpeg.js
```diff
@@ -0,0 +1,431 @@
...
+    if (!path) {
+      throw new Error(
+        `Unable to extract asset file path for ${type} (${video.id})`
+      )
+    }
+
+    const optionsHash = crypto
+      .createHash(`md5`)
+      .update(JSON.stringify(fieldArgs))
+      .digest(`hex`)
```
muescha: new code style suggestion from:

#8805 Issue: feat: update gatsby plugins to use new createContentDigest helper

```suggestion
    const optionsHash = createContentDigest(fieldArgs)
```
and add to imports:
```javascript
import { createContentDigest } from "gatsby-core-utils"
```

axe312ger: cool! will do! thx

packages/gatsby-transformer-video/README.md
axe312ger: If binaries are missing, the CLI outputs: `Cannot find ffprobe`

Writing on an updated README at this very second as well

packages/gatsby-transformer-video/src/gatsby-node.js
axe312ger: done :)

packages/gatsby-transformer-video/README.md
axe312ger: Mentioned it in the readme, the CLI gives an reasonable output by default as well

packages/gatsby-transformer-video/src/ffmpeg.js
axe312ger: done :)

yarn.lock
axe312ger: done :)

packages/gatsby-transformer-video/src/ffmpeg.js
axe312ger: done :)

axe312ger:
So refactoring is done to have one field per codec. Additionally it uses the new schema customisation API and removes all code-smells from the last reviews.

This plugin now adds the following subfields to all file nodes:

* `videoH264`
* `videoH265`
* `videoGif`
* `videoWebP`

Next steps:

* Apply recommendations from above (tweak FFMPEG)
* Add VP9 codec support
* Add `videoPreset` which executes a configurable function to execute `fluent-ffmpeg` on your own
* Posterimage feature

axe312ger:
I implemented VP9 and some performance & quality improvements for h264/h265.

Thanks to https://developers.google.com/media/vp9/settings/vod/#bitrate we can even automatically determine a fitting bitrate based on our source data and given field args (like maxWidth)


The generated files now have a decent (not super high) quality, small file size and the file sizes differ as expected based on coded.

* source mov: 15,7 MB (720p)
* h265: 1.5mb
* VP9: 3.5mb
* h264: 4.9mb

I won't claim the settings are perfect but its a clear improvement to the first version :)

packages/gatsby-transformer-video/README.md
```diff
@@ -0,0 +1,123 @@
...
+### Linux
+
+Linux users might be able to: `sudo apt-get install ffmpeg`
+
+Make sure you have `ffprobe` installed as well.
+
+### OSX
+
+I could not figure out how to use `brew` to install ffprobe as well
```
muescha: should the `brew install ffmpeg` also install the ffprobe?

i think a link to the official wiki would be the best: https://trac.ffmpeg.org/wiki/CompilationGuide/macOS#ffmpegthroughHomebrew

packages/gatsby-transformer-video/src/gatsby-node.js
muescha: i think i can close this comment as resolved ...

packages/gatsby-transformer-video/src/ffmpeg.js
```diff
@@ -0,0 +1,443 @@
...
+import ffmpeg from "fluent-ffmpeg"
+import imagemin from "imagemin"
+import imageminGiflossy from "imagemin-giflossy"
+import PQueue from "p-queue"
+
+export default class FFMPEG {
+  constructor({ rootDir, ffmpegPath, ffprobePath }) {
+    this.queue = new PQueue({ concurrency: 1 })
+    this.cacheDir = resolve(
+      `${rootDir}/node_modules/.cache/gatsby-transformer-video/`
```
muescha: a) should the cache of user data better somewhere in `www/.cache/gatsby-transformer-video/`?

muescha: b) i can not see that `cacheDir` is used anymore?

packages/gatsby-transformer-video/src/ffmpeg.js
```diff
@@ -0,0 +1,443 @@
...
+        `-movflags +faststart`,
+        `-profile:v high`,
+        `-bf 2	`,
+        `-g ${Math.floor((fieldArgs.fps || currentFps) / 2)}`,
+        `-coder 1`,
+        `-pix_fmt yuv420p`,
+      ].filter(Boolean)
+
+      console.log(
+        `ffmpeg command:\n\nffmpeg -i ${path} -vf "${filters}" -c:v libx264 ${outputOptions.join(
```
muescha: - for all `ffmpeg command` logs: i would avoid "guessing" the resulting ffmpeg command, this info can not be always in sync with the generated command (if you forget to change it or if `fluent-ffmpeg` changed internally) and gives you wrong debug informations (for example it not shows the changes from `enhanceFfmpegForFilters`)
- i think it is better to use the "real" debug info from fluent-ffmpeg with `.on('start', function(commandLine) { console.log(commandLine);})`
- see my comment at `executeFfmpeg`

packages/gatsby-transformer-video/src/ffmpeg.js
```diff
@@ -0,0 +1,443 @@
...
+      })
+    })
+
+  executeFfmpeg = async (ffmpegSession, publicPath) => {
+    let lastLoggedPercent = -10
+    return new Promise((resolve, reject) => {
+      ffmpegSession
+        .on(`start`, function() {
+          console.log(`[FFMPEG] Spawned`)
+        })
```
muescha:
```js
  .on('start', function(commandLine) {
    console.log('[FFMPEG] Spawned with command: \n\n' + commandLine + '\n\n');
  });
```

(see https://github.com/fluent-ffmpeg/node-fluent-ffmpeg#start-ffmpeg-process-started)

axe312ger:
Crunching more numbers:

I got 7 source videos, mostly 720p from iPhone cameras or final cut.

It takes about ~16 minutes to convert all videos on my i9.

Gatsby cloud took 1h 5min to convert all of it.

Restarting the build with the same cache, 30 seconds build time 💪 😍

packages/gatsby-transformer-video/README.md
axe312ger: Wonderful, good catch. Tapping the formula makes ffprobe available :)

packages/gatsby-transformer-video/src/ffmpeg.js
axe312ger: Thanks, I took this into account in the last refactor :)

packages/gatsby-transformer-video/src/ffmpeg.js
axe312ger: Doing so now :)

packages/gatsby-transformer-video/src/ffmpeg.js
axe312ger: I readded this caching layer and moved it into `.cache`

Curious what gatsby cloud now will do, before this change it had to reconvert all videos on new branch builds

axe312ger:
Next refactoring is done, took everything @muescha mentioned into account.

The code should be more flexible and readable as well as I moved the codec specific code into a subdirectory, one file per codec/profile

https://github.com/gatsbyjs/gatsby/tree/gatsby-transformer-video/packages/gatsby-transformer-video/src/profiles

Implementing the custom profile function should  now be pretty low efford :)

I will do a complete local run on my machine and publish `v0.1.1` afterwards.

packages/gatsby-transformer-video/README.md
muescha: done

packages/gatsby-transformer-video/src/ffmpeg.js
muescha: done

packages/gatsby-transformer-video/src/ffmpeg.js
muescha: done

axe312ger:
Screenshot feature implemented and released as well.

Based on https://github.com/fluent-ffmpeg/node-fluent-ffmpeg#screenshotsoptions-dirname-generate-thumbnails

Example project code can be found here:
https://github.com/gatsbyjs/gatsby/tree/gatsby-transformer-video/examples/using-gatsby-transformer-video

The example will render:

Kapture 2020-02-14 at 23 38 27

packages/gatsby-transformer-video/src/gatsby-node.js
```diff
@@ -0,0 +1,273 @@
...
+  const { createNode } = actions
+
+  const program = store.getState().program
+  const rootDir = program.directory
+  const cacheDir = resolve(
+    rootDir,
+    `node_modules`,
+    `.cache`,
+    `gatsby-transformer-video`
```
muescha: > cache in node_modules as it is more persistent

but then `gatsby clean` don't clean up the user data cache - i guess that behaviour is surprising for the user

axe312ger: Usually, yes! We had this discussion before with `gatsby-transformer-sqip`.

Let me explain my decision:

Transforming videos is slow, when I keep the cache in gatsbys `.cache`, an update to any dependency, any change to gatsby-config.js or running a production build after using develop mode will trigger new videos to be build.

This means in my test case:
* multiple times of waiting 15 minutes when developing the project, even when not touching video related stuff
* rerendering of the exact same videos when doing a test production build on my machine
* any case cleaning gatsby cache will trigger 1h of waiting on the CI as well. Means every new pull request, most updates to pull requests etc. trigger full video conversion

Keeping the number of reconversions of videos as low as possible should be our goal to achieve neat DX/UX and reducing the load on CI providers.

On the other hand, the `node_modules/.cache` will persist way better. CI environments like netlify & gatsby cloud (just tested it with latest version) keep this directory till you install new dependencies. Also, pull requests will receive the masters `node_modules/.cache` directory.


The downsides of the current implementation I see so far:

* When query field arguments get changed, the videos option hash will change and a new file is generated. But old files won't be deleted. (We could create a list of "active" video files and remove old files from the cache in a post build gatsby event)
* Installing dependencies is a case where it is very unlikely that the converted videos will result in another file as before. At least his is a more rare case.


I am very open for suggestions and gatsby-wide solutions. The issue is very much the same for https://github.com/gatsbyjs/gatsby/tree/gatsby-transformer-video :)

axe312ger: To work around this issue, I came up with this. Still works great for one of my projects:

https://github.com/axe312ger/gatsby-plugin-sftp-cache

Basically uses a ftp server as cache. The great benefit here is, that I can do heavy processing of  files on my local development machine, sync the results via `gatsby build` to my FTP and my CI just has to download if they are missing files.

muescha: i understand why you do this - i see it also with the thumbnails creation, for development they cleaned away ....

i think it should be solved gatsby-wide

is there an already existing issue for this?

muescha: if you place it into `node_modules` then there should be a note in the readme

axe312ger: There is no open issue for this. I know I discussed it in person with @KyleAMathews long time ago when working on the first version of SQIP

Also @pieh was involved back then.

Here are related comments I found:

https://github.com/gatsbyjs/gatsby/pull/5862 (this made the netlify cache plugin possible)
https://github.com/gatsbyjs/gatsby/issues/5224#issuecomment-387327378 (Details of the netlify workaround)


Ahh here, This mentions permanent cache and says its a v3 topic and on Kyles list: https://github.com/gatsbyjs/gatsby/pull/4205#issuecomment-368640240

Gatsby Cloud was also mentioned by Kyle at some point that it might help out at some point.

packages/gatsby-transformer-video/src/ffmpeg.js
```diff
@@ -0,0 +1,492 @@
...
+            const percent = Math.floor(progress.percent)
+            const elapsedTime = Math.ceil(
+              (performance.now() - startTime) / 1000
+            )
+            const estTotalTime = (100 / percent) * elapsedTime
+            const estTimeLeft = Math.ceil(estTotalTime - elapsedTime)
+            const loggedTimeLeft =
+              estTimeLeft !== Infinity && ` (~${estTimeLeft}s)`
+
+            console.log(`${loggingPrefix} ${percent}%${loggedTimeLeft}`)
```
muescha: not in the scope of this PR but as a follow up PR: other plugins are using ProgressBar for the logging ... [Code Search ProgressBar](https://github.com/gatsbyjs/gatsby/search?q=progressbar&unscoped_q=progressbar)

muescha: time calculation can be done then by the ProgressBar:

https://github.com/gatsbyjs/gatsby/blob/db36d6995659c190b4aefda2d60828d48a430611/packages/gatsby-plugin-sharp/src/utils.js#L12-L19

see Tokens: https://www.npmjs.com/package/progress#tokens

axe312ger: +1 :)

axe312ger:
This is how the example `using-gatsby-transformer-video` looks like:

Kapture 2020-02-14 at 23 38 27

axe312ger:
Current status:

* Refactorings are done, IMHO the code is now pretty maintainable and easy to extend.
* It is now running fine on two production projects, another one is coming
* Seems like h265 finally running on iOS (https://brandur.org/fragments/ffmpeg-h265#section-4 😀)


Looking forward for more reviews :)

muescha:
file view of this PR is broken :(

axe312ger:
> file view of this PR is broken :(

Yes, already contacted github support 😓

muescha:
just a side note: i will just point to #20949 `☂️ [Umbrella issue] Yarn 2 issues` where is mentioned that `node_modules` would be read only with Yarn 2
so the current `CACHE_FOLDER_VIDEOS` would not work

muescha:
maybe you can hop on this "selective cache invalidation", as i understand this is similar functionality (but i am not sure about it)

issues/PR:
- #22090 `chore(gatsby-source-filesystem): allow createRemoteFileNode and createFileNodeFromBuffer to receive getCache`
- #8379 `feat(gatsby): selective cache invalidation`
- #8324 `Don't clear downloaded files from createRemoteFileNode calls when we clear cache on package/code changes`

pieh:
About "selective cache invalidation" - this will not land fully in v2 (if it lands - it will be behind feature flag, because there will be a lot more restrictions needed in place to properly invalidate parts of cache - mainly fact that currently plugins can put stuff to other plugin caches making it impossible to reliably invalidate part of cache).

As for what selective cache invalidation could do for `gatsby-transformer-video` - once it lands and when enabled (possibility of feature flag in v2) - it means we will not delete cache for a plugin (`cache`, `cache.directory` directory, nodes) if the plugin didn't change since last run. So it still won't be trully permament cache as updating the plugin would blow away the cache.

There is separate (somewhat connected) need for more permanent caches. File downloading is good example here as `gatsby-source-filesystem` receive fair share of updates, but most of them don't impact file downloading. So `gatsby-source-filesystem` (or any other plugin) potentially could have ability to create separate cache that is versioned "manually" by plugin (and not automatically by plugin version), that would allow to keep artefacts around in more controlled manner.

axe312ger:
Files view is working again for this PR.

Happy reviewing :)

axe312ger:
Currently I found the following issues:

* If downloading of a contentful asset fails (eG not published), the error message is build the wrong way:

```
Cannot create property 'message' on string 'failed to process https://videos.ctfassets.net/.../.../.../canggu.mp4
Error: ENOENT: no such file or directory, open '/Users/bene/.../.cache/caches/gatsby-transformer-video/tmp-c427f69f25b6730b18af9cf0ca78f510.mp4''
```

* current download of videos is pretty slow as it downloads only one video at once
* we should really get `createRemoteFileNode ` to properly retry on its own
* even when we have the files on disk, it looks like `createRemoteFileNode ` still downloads them even when it should just create the node from the already downloaded file
* in case the contentful download request fails, there is currently no way to get the request id to contact contentful if the error was on their side or ours (failed requests happend very often lately on Gatsby cloud for unknown reasons 🤷‍♂ )

pvdz:
(I've converted this PR into a Draft PR)

axe312ger:
More things popped up that might be needed:

* Add AV1 support as it (seems to be) great, is open source and supported by FF, Chrome and Edge: 
  * https://trac.ffmpeg.org/wiki/Encode/AV1
  * https://hacks.mozilla.org/2019/05/firefox-brings-you-smooth-video-playback-with-the-worlds-fastest-av1-decoder/
  * https://caniuse.com/#feat=av1
* Firefox seems to have issues detecting if it can play the h264 or not. We need to pass the proper codec parameter to the <source type /> attribute. So `<source src="...-h264.mp4" type="video/mp4">` will become `<source src="...-h264.mp4" type="video/mp4; codecs=avc1.4d4032">`
  * Unfortunately ffprobe does not give us the value directly
  * There are tools like https://github.com/gpac/mp4box.js/ + https://www.npmjs.com/package/mpeg4-mimetype
  * There are a **lot** of coded specific rules to consider https://developer.mozilla.org/en-US/docs/Web/Media/Formats/codecs_parameter
* Taking screenshots for poster images and so on should be its own field. They are unrelated to the codec but might get special dimensions / color filters. Additionally, the data structure becomes more easy to handle when rendering the data (Screenshots not hidden within a codec field)

axe312ger:
We found a bug:

* When you reupload a file in contentful, and the filename does not change, the current code won't redownload the source file and the video won't be transformed again

muescha:
maybe that is a general cache problem with contentful? is there a content hash / last updated field available for a file from contentful?

axe312ger:
@muescha: no the bug is in my custom video download code I wrote as replacement for getRemoteFile as it was/is buggy/not picking up files that got already downloaded earlier.

See here: https://github.com/gatsbyjs/gatsby/pull/15783/files#diff-5aca1886e81322607f4efde72d55ba20

axe312ger:
I cleaned up the code and applied several fixes and improvements we stumbled upon while using v0.1 in actual projects.

#### 0.2.0 (2020-11-02)

##### Build System / Dependencies

- upgrade dependencies ([2457d172](https://github.com/gatsbyjs/gatsby/commit/2457d1722cf26404ac3db33068d202ff1f58df0d))

##### Chores

- remove debug console.log ([7ee4556d](https://github.com/gatsbyjs/gatsby/commit/7ee4556dda9c588896b2b6eb82c156948df95124))

##### Documentation Changes

- mark as beta and link usage example ([8438fccd](https://github.com/gatsbyjs/gatsby/commit/8438fccd67d95cc22550300a91614c380e4c715f))

##### New Features

- add additional information when video could not be analyzed cus of a broken or unsupported file ([a70ea950](https://github.com/gatsbyjs/gatsby/commit/a70ea950f93e5c26a656c6e95f1ef4a381b2d6d0))
- add width, height and aspect ratio to video nodes ([e1370884](https://github.com/gatsbyjs/gatsby/commit/e1370884e875335751eedb697e3eeb5f949ea301))
- align file naming for files and contentful assets ([240eefe9](https://github.com/gatsbyjs/gatsby/commit/240eefe9213ce818e55c26ef88f1324239527482))
- clean up logging and reporting ([91d50ec7](https://github.com/gatsbyjs/gatsby/commit/91d50ec7df9698b80e789e6be167e635a8265167))
- slightly increase default quality for h264 and h265 ([4b05c482](https://github.com/gatsbyjs/gatsby/commit/4b05c4821506defebc9ebd826f5c59324bbdeb63))
- move default directory into static folder ([767e75f9](https://github.com/gatsbyjs/gatsby/commit/767e75f9fd27bbfb348a543117f374a862d676c6))
- use new reporter API ([dd80e1aa](https://github.com/gatsbyjs/gatsby/commit/dd80e1aa50419edf9ff48f8b54e06fbdc3edf38f))
- move screenshots to extra videoScreenshots resolver ([242e99ea](https://github.com/gatsbyjs/gatsby/commit/242e99ea6956b44e782ac24ed9873b3a4542b582))
- switch to medium preset for h264 and h265 for higher quality, smaller size and almost same conversion time ([370c0d07](https://github.com/gatsbyjs/gatsby/commit/370c0d07e2ec546260a6058dfba7a2de4d7bbbe9))

##### Bug Fixes

- redownload files if actual file on Contentful changed ([53909a2e](https://github.com/gatsbyjs/gatsby/commit/53909a2e57fe6055583a15e7e5e8e1aaca8788f6))
- ensure screenshots are queued ([c61df862](https://github.com/gatsbyjs/gatsby/commit/c61df8623f16e94e7f28c9f5fc0a43b80f0da16c))

##### Refactors

- adjust example to match new screenshots resolver ([73f67b33](https://github.com/gatsbyjs/gatsby/commit/73f67b33b5fbb984fd605f41649a3b4a3da6640b))
- update babelrc to use babel-preset-gatsby-package ([c7371733](https://github.com/gatsbyjs/gatsby/commit/c73717330f8f214166ddfcec4b319805683b866f))

##### Code Style Changes

- reapply eslint ([1c21a16f](https://github.com/gatsbyjs/gatsby/commit/1c21a16f5de8e092a795a2bdb1f4bba0e0b00cca))

axe312ger:
Planning on a v0.3 release that supports Gatsby v3, is less spammy in the logs and actually caches screenshots/poster images

ascorbic:
@axe312ger Would it make sense to move this into its own repo? It's not going to be merged any time soon, and publishing releases from a branch isn't great practice.

axe312ger:
@ascorbic yeah maybe we should do that.

But I was hoping to not end up alone as maintainer :innocent:

wardpeet:
cc @KyleAMathews what do you suggest here? I rather not add it to the monorepo as we don't have any knowledge or time to dig into this. We also don't add ffmpeg to cloud so this wouldn't work there.

axe312ger:
> We also don't add ffmpeg to cloud so this wouldn't work there.

it does work, just needs to download a 90mb zip with two binaries 🙈

KyleAMathews:
I love the idea of this but... if image processing trips up a lot of people using Gatsby, well, video processing is like the sun to image processing's little match. It's immensely expensive & I don't think people would have a good experience using a video transform plugin. I think most people would be much better off using a dedicated service for video hosting. There's some sites for sure where this transformer would be great but I don't think the number is enough that we'd want to make it an official project.

ascorbic:
Yes, and if sharp causes people issues, ffmpeg is another level. This is a great plugin, but making it official would take a lot of work, including fixing caching.

axe312ger:
> This is a great plugin, but making it official would take a lot of work, including fixing caching.

Fixing/improving the cache would be very helpful, for this plugin, but also for others 🙃

axe312ger:
I'll move this out of this repo and make a community plugin out of it as soon I find time for it.

When doing this, I'll also rewrite to Gatsby v3 and maybe to TS.

ascorbic:
There's now an experimental flag to enable retaining downloaded assets when auto-cleaning. I wonder if that will be enough to let this use the regular asset cache?

axe312ger:
Sounds good. You will hear from me for sure when I move this code 😅

axe312ger:
Hello everyone,

I moved the code of this PR to its own repository: https://github.com/hashbite/gatsby-transformer-video
All open ToDos and known bugs are moved into issues: https://github.com/hashbite/gatsby-transformer-video/issues

Also `v0.3.0` got released. It just bumps the dependencies for Gatsby v3. Works fine in my first tests.
More features and bug fixes will follow as soon I find time, especially the screenshot bug is annoying.

Hope to get some collaboration going in the new repo, see ya over there 👋

PR #15788, opened 2019-07-16
chore(starters): add gatsby-starter-newage

chiragswadia:
## Description

added gatsby-starter-newage design by startbootstrap


wardpeet:
That's slick! Thanks for creating this nice starter. Let's what for CI to turn green.

PR #15791, opened 2019-07-16
chore(www): Switch to gatsby-design-tokens package

fk:
Following up on #15596/#15779, this 

* adds the new `gatsby-design-tokens` package (and pins it at `0.0.1`—smelling a `0.0.2` already)
* removes `www/src/utils/tokens`
* rewires things for `gatsby-design-tokens`
* orders imports in `www/src/utils/presets.js` alphabetically (b/c OCD)


www/package.json
```diff
@@ -22,6 +22,7 @@
     "emotion-theming": "^10.0.10",
     "fuse.js": "^3.2.0",
     "gatsby": "^2.11.0",
+    "gatsby-design-tokens": "0.0.1",
```
DSchau: Why pin?

sidharthachatterjee:
```suggestion
    "gatsby-design-tokens": "^0.0.1",
```

sidharthachatterjee: Spoke to @fk about this. He prefers pinning here to ensure publishing design changes to `www` are deliberate.

fk:
Thanks Dustin and Sid!

PR #15793, opened 2019-07-16
chore(gatsby-design-tokens): add homepage and stub README

pieh:
Recently added script that makes sure all packages have proper metadata ( https://github.com/gatsbyjs/gatsby/pull/15477 ) is causing failed lint tests after merging https://github.com/gatsbyjs/gatsby/pull/15779. This is quick chore to make sure it passes.

Would be great to have 1 paragraph explanation of what this package is for, but I have no context so just added TODO to readme.

### DIFF
--- a/packages/gatsby-design-tokens/README.md
+++ b/packages/gatsby-design-tokens/README.md
@@ -1,3 +1,3 @@
 # gatsby-design-tokens
 
-_TODO_: document purpose and usage once we hit 1.0.0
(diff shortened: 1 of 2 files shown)


fk:
Thanks Michal, sorry for being late on this. Thanks @sidharthachatterjee for jumping in! 🤗 
I'll get that README updated ASAP (and some v0.0.2 fixed in 😅).

PR #15814, opened 2019-07-17
update theme core API docs

ChristopherBiscardi:


docs/docs/themes/api-reference.md
```diff
@@ -42,11 +40,12 @@ module.exports = {
 }
 ```
 
-In your theme's `gatsby-config.js` you can return a function: the argument it receives are the options:
+You can access options that are passed to your theme in your theme's `gatsby-config`. You can use this to make filesystem sourcing configurable, accept different nav menu items, or change branding colors from the default. To take advantage of the options that are passed in when configuring your theme in a user's site, return a function in your theme's `gatsby-config.js`. The argument the function receives are the options the user passed in.
```
jlengstorf:
```suggestion
You can access options that are passed to your theme in your theme's `gatsby-config`. You can use options to make filesystem sourcing configurable, accept different nav menu items, change branding colors from the default, and anything else you want to make configurable.

To take advantage of the options that are passed in when configuring your theme in a user's site, return a function in your theme's `gatsby-config.js`. The argument the function receives is the options the user passed in.
```

docs/docs/themes/api-reference.md
```diff
@@ -56,39 +55,86 @@ module.exports = themeOptions => {
 }
 ```
 
-Then, in your theme's `gatsby-node.js` you can access them as the second argument to `createPages`:
+While using the usual object export (`module.exports = {}`) in your theme means that you can run the theme standalone as it's own site, When using a function in your theme to accept options you will need to run the theme as part of an example site. See how the [theme authoring starter](https://github.com/gatsbyjs/gatsby/tree/master/themes/gatsby-starter-theme-workspace) handles this using Yarn Workspaces.
```
jlengstorf:
```suggestion
While using the usual object export (`module.exports = {}`) in your theme means that you can run the theme standalone as its own site, when using a function in your theme to accept options you will need to run the theme as part of an example site. See how the [theme authoring starter](https://github.com/gatsbyjs/gatsby/tree/master/themes/gatsby-starter-theme-workspace) handles this using Yarn Workspaces.
```

docs/docs/themes/api-reference.md
```diff
@@ -56,39 +55,86 @@ module.exports = themeOptions => {
...
+
+### Accessing Options elsewhere
+
+Note that because themes are also plugins you can also access the options in any of the lifecycle methods taht you're used to. For example, in your theme's `gatsby-node.js` you can access the options as the second argument to `createPages`:
```
jlengstorf:
```suggestion
Note that because themes are plugins you can also access the options in any of the lifecycle methods that you're used to. For example, in your theme's `gatsby-node.js` you can access the options as the second argument to `createPages`:
```

docs/docs/themes/api-reference.md
```diff
@@ -56,39 +55,86 @@ module.exports = themeOptions => {
...
 ```js:title=gatsby-node.js
 exports.createPages = async ({ graphql, actions }, themeOptions) => {
   console.log(themeOptions)
 }
 ```
 
-## Component Shadowing
+## Shadowing
+
+Since themes are usually deployed as NPM packages that other people use in their sites, we need a way to modify certain files, such as React components, without making changes to the source code of the theme. This is called Shadowing. Shadowing is a filesystem-based API that allows us to replace one file with another at build time. For example, if we had a theme with a `Header` component we could replace that `Header` with our own by creating a single file using Shadowing.
```
jlengstorf:
```suggestion
Since themes are usually deployed as npm packages that other people use in their sites, we need a way to modify certain files, such as React components, without making changes to the source code of the theme. This is called _Shadowing_.

Shadowing is a filesystem-based API that allows us to replace one file with another at build time. For example, if we had a theme with a `Header` component we could replace that `Header` with our own by creating a new file in our site and placing it in the correct location for Shadowing to find it.
```

docs/docs/themes/api-reference.md
```diff
@@ -56,39 +55,86 @@ module.exports = themeOptions => {
...
+
+### Overriding
+
+Taking a closer look at our `Header` example, let's say we have a theme called `gatsby-theme-amazing`. That theme uses a `Header` component to render navigation and other miscellaneous items. The path to the component from the root of the npm package is `gatsby-theme-amazing/src/components/header.js`.
+
+We'd like the `Header` component to do something different (maybe change colors, maybe add additional navigation items, really anything you can think of). To do that, we create a file in our site at `src/gatsby-theme-amazing/components/header.js`. We can now export any React component we want from this file and Gatsby will use it instead of the theme's component.
+
+note: you can shadow components from other themes too using the same method. Use the destination path in your new theme instead of in your site. Read more about advanced applications of this in [latent shadowing](https://johno.com/latent-component-shadowing).
```
jlengstorf: I had to read this a couple times to understand what it meant, so I shortened it to the bare essentials and figure people will click the link to get more context.
```suggestion
> 💡 Note: you can shadow components from other themes using the same method. Read more about advanced applications in [latent shadowing](https://johno.com/latent-component-shadowing).
```

docs/docs/themes/api-reference.md
```diff
@@ -56,39 +55,86 @@ module.exports = themeOptions => {
...
+
+// these props are the same as the original component would get
+export default props => <Header {...props} myProp="true" />
 ```
 
-Gatsby Themes also allow you to customize any file in a theme's `src` directory by following a file naming convention.
-If you're using `gatsby-theme-tomato` which uses a `ProfileCard` component located at `src/components/profile-card.js` you can override the component by creating `src/gatsby-theme-tomato/components/profile-card.js`. If you want to see what props are passed you can do so by putting the props into a `pre` tag:
+Taking this approach means that when we upgrade our theme later we can also take advantage of all the updates to the `Header` component because we haven't fully replaced it, just modified it.
+
+### What path do use to shadow a file?
```
jlengstorf:
```suggestion
### What path should be used to shadow a file?
```

docs/docs/themes/api-reference.md
```diff
@@ -56,39 +55,86 @@ module.exports = themeOptions => {
...
 
-```js:title=src/gatsby-theme-tomato/components/profile-card.js
-import React from "react"
+Until we build tooling to support automatically handling shadowing you will have to figure out how to convert paths in a theme into paths in your site on your own. Luckily the way to do that is few steps. Take the `src` directory from the theme, and move it to the front of the path, then write a file at that location in your site. Looking back on our `Header` example, this is the path to the component in our theme:
```
jlengstorf:
```suggestion
Until we build tooling to support automatically handling shadowing, you will have to manually locate paths in a theme and create the correct shadowing paths in your site.

Luckily, the way to do that is only a few steps. Take the `src` directory from the theme, and move it to the front of the path, then write a file at that location in your site. Looking back on our `Header` example, this is the path to the component in our theme:
```

docs/docs/themes/api-reference.md
```diff
@@ -56,39 +55,86 @@ module.exports = themeOptions => {
...
+
+and here is the path where we would shadow it in our site:
+
+```
+<our-site>/src/gatsby-theme-amazing/components/header.js
+```
+
+Shadowing only works on imported files in the `src` directory. This is because shadowing is built on top of Webpack, so the module graph needs to include the shadowable file.
+
+Since we can use many themes in a given site, there are many potential places to shadow a given file (one for each theme and one for the user's site). In the event that there exist many files attempting to shadow `gatsby-theme-amazing/src/components/header.js`, the last one included in the plugins array will win although the user's site will beat any plugin.
```
jlengstorf:
```suggestion
Since we can use multiple themes in a given site, there are many potential places to shadow a given file (one for each theme and one for the user's site). In the event that more than one themes are attempting to shadow `gatsby-theme-amazing/src/components/header.js`, the last theme included in the plugins array will win. The site itself takes the highest priority in shadowing.
```

docs/docs/themes/api-reference.md
```diff
@@ -56,39 +55,86 @@ module.exports = themeOptions => {
...
+```js:title=my-site/gatsby-config.js
 module.exports = {
-  plugins: ["gatsby-theme-blog"],
+  plugins: [`gatsby-theme-blog`, `gatsby-theme-notes`],
 }
 ```
+
+Themes at their core are an algorithm that merges multiple `gatsby-config.js` files together into a single config your site can use to build with. To do that we need to define how to combine two `gatsby-config.js`s together. Before we can do that, we need to flatten the parent/child relationships into a single array. This results in the final ordering when considering which shadowing file to use if multiple are available.
+
+Our first example results in a final ordering of `['gatsby-theme-parent', 'gatsby-theme-child']` (parents always come before their children so that children can override functionality), while our second example results in `['gatsby-theme-blog', 'gatsby-theme-core']`.
```
jlengstorf:
```suggestion
Our first example results in a final ordering of `['gatsby-theme-parent', 'gatsby-theme-child']` (parents always come before their children so that children can override functionality), while our second example results in `['gatsby-theme-blog', 'gatsby-theme-notes']`.
```

docs/docs/themes/api-reference.md
```diff
@@ -56,39 +55,86 @@ module.exports = themeOptions => {
...
+
+Once we have the final ordering of themes we merge them together using a reduce function. [This reduce function](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/utils/merge-gatsby-config.js) specifies the way each key in `gatsby-config.js` will merge together. Unless otherwise specified below, the last value wins.
+
+- `siteMetadata` and `mapping` both merge deeply using lodash's `merge` function.
```
jlengstorf:
```suggestion
- `siteMetadata` and `mapping` both merge deeply using lodash's `merge` function. This means a theme can set default values in `siteMetadata` and the site can override them using the standard `siteMetadata` object in `gatsby-config.js`.
```

docs/docs/themes/api-reference.md
```diff
@@ -56,39 +55,86 @@ module.exports = themeOptions => {
...
+- `plugins` are normalized to remove duplicates, then concatenated together.
```
jlengstorf: Is it worth explaining how duplicates are removed? E.g. last declared plugin wins?

docs/docs/themes/api-reference.md
ChristopherBiscardi: hmm yeah, there are two thing I wanted to convey here. One is that shadowing works in your site and in themes, and 2 that this can be used to add support for themes that haven't been installed yet.

docs/docs/themes/api-reference.md
ChristopherBiscardi: they're duplicates by definition so it shouldn't matter how they're removed (same `resolve`, same `options`). We can revisit when "plugin merging" gets in, but that will likely go under it's own API heading anyway since it will have to be `onPluginSetOptions` or something.

docs/docs/themes/api-reference.md
ChristopherBiscardi: whoa, this `core` reference totally came out of my hands without realizing I was writing it.

PR #15834, opened 2019-07-17
feat(www): add copy button to code snippets

### LINKED ISSUE #5030: Docs enhancement - add a copy button to code examples
## Summary
Code examples in tutorials, READme files, and guides could use a copy button so people don't miss a single thing when they copy code.

## Motivation 
When we did a Gatsby workshop a while ago, many workshop attendees' computer screens didn't show backtics very clearly and they missed copying them a lot.

Another related issue brought up by @twhiteaker is that when copying code from Firefox to VS code (and Notepad ++), the code ended up having two carriage returns added to the top and bottom. This actually mattered for a markdown file, it appears.

## Acceptance criteria (by @marcysutton)

- As a keyboard user, I should be able to reach and operate a copy button for each code example and see a visible focus style.
- As a sighted user, I should see a graphic icon with adequate color contrast for each code example that is intuitive and clear that it will copy code to the clipboard. 
    - If the icon button is unclear through user testing, adding the text “Copy” or “Copy to clipboard” may help
- As an assistive technology user, when navigating through the copy buttons I should hear unique button names like `${filename}: copy code to clipboard` to know which file I’m copying to the clipboard.
- As a mouse user, clicking on the button should copy the adjacent code to the user’s clipboard in all major browsers.
- As a keyboard or screen reader user, focusing on the button and hitting the ENTER or SPACE keys should copy the adjacent code to the clipboard in all major browsers.
- As a sighted user, upon activating the button, I should see a visible “copied to clipboard!” toast notification.
- As a keyboard user, upon activating the copy button and toast notification, I should be able to hit the TAB key to reach a visible “Close” or “X” button with a visible focus style. When the toast notification is not visible, the button should be rendered inoperable.
- As a user with limited mobility or a cognitive impairment, I should have time to read the toast notification before it is dismissed by only closing it when activating the close button.
- As a sighted user, I should be able to dismiss the toast notification by clicking the close button or anywhere else on the screen. 
- As a screen reader user, I should hear an announcement of `${filename} copied to clipboard` after activating the button. 
    - This can be accomplished with a persistent, visually hidden DOM element with ARIA `role=”status”.
    - Add and announce a keyboard shortcut, like “ENTER” to close? (We can play with it)

Potential icon: 

copy to clipboard icon

Alternatively, there could be a button that just visually says "copy" without an icon.

DSchau:
## Description

At a workshop yesterday, @KyleAMathews and I noticed that people tend to highlight the wrong lines, or lose a few lines when highlighting leading to strange errors and confusion. We can avoid this by introducing a copy functionality to our code snippets.

This PR does just that while also maintaining _most_ of our other features, notably:

1. Syntax highlighting
1. Titles from the language (e.g. `jsx:title=gatsby-config.js`)
    - Note: we should make this component re-usable, and then we can pass a `title` as a JSX prop or something
1. Language badges

A `gif` is worth _lots_ of words, so it looks something like this:

copy-button

Also @fk would love your design eye — I think it looks OK but could certainly be improved :) Feel free to tweak anything!

## Related Issues

Fixes #5030


www/src/components/pre.js
```diff
@@ -0,0 +1,51 @@
...
+      return merged
+    }, {})
+  )
+}
+
+export default ({ children }) => {
+  const [language, { title = `` }] = getParams(children.props.className)
+  const content = children.props.children
+  return (
+    <Highlight
```
DSchau: Note: we’ll probably want to pass our custom Prism theme as a theme here — will do just that!

sidharthachatterjee:
Just stopping by to say that this is incredible @DSchau 💜

DSchau:
@marcysutton @gillkyle could use some assistance! Let me address what I’ve tried to improve re: accessibility, and then we can see if there’s room for improvement!

> As a keyboard user, I should be able to reach and operate a copy button for each code example and see a visible focus style.

✅ 

> As a sighted user, I should see a graphic icon with adequate color contrast for each code example that is intuitive and clear that it will copy code to the clipboard.

I used the text `Copy` and `Copied` instead of an icon and toast widget? That OK?

> As an assistive technology user, when navigating through the copy buttons I should hear unique button names like ${filename}: copy code to clipboard to know which file I’m copying to the clipboard.

Maybe done? I used the text you describe as the `name` attribute on the button and swap between them when the button is in copied state or not.

> As a mouse user, clicking on the button should copy the adjacent code to the user’s clipboard in all major browsers.

✅ 

> As a keyboard or screen reader user, focusing on the button and hitting the ENTER or SPACE keys should copy the adjacent code to the clipboard in all major browsers.

✅ 

> As a sighted user, upon activating the button, I should see a visible “copied to clipboard!” toast notification.

I didn’t do this because we don’t really have a pattern for this — and it’s a bit out of scope for this PR. I _can_ do this, but I’m thinking swapping the text could be enough? More I can do here?

> As a user with limited mobility or a cognitive impairment, I should have time to read the toast notification before it is dismissed by only closing it when activating the close button.

5 seconds? More or less?

> As a screen reader user, I should hear an announcement of ${filename} copied to clipboard after activating the button.

✅ I _think_. I stole your ScreenReaderText component!

fk:
<3 Will have a mock for this by tomorrow.

gillkyle:
I think swapping the text as a notification might actually simplify some of the accessibility concerns. I recall @marcysutton talking about how toast notifications can be difficult with focus though she'll certainly have a more definite opinion.

DSchau:
Woo-hoo! I wrote some relatively janky-ish code to add support for “directives” (i.e. highlight-start, highlight-end, hide-line, etc.), and the accessibility seems to be much approved — but still could use a once-over from someone a little more well-versed than myself.

I think this is ready for a set of 👀 Still need that slick design tho @fk but no rush!

Longer term, we should probably open-source this as a separate package; I can imagine it being pretty useful in any MDX setup!

marcysutton:
(I thought I commented on this yesterday, but didn't hit send!)

Yay, that all sounds good to me! I'm so glad I wrote those success criteria. After user testing I actually agree that putting the "copied" text right on the button makes a lot of sense. I wouldn't worry about changing the design for that, but we might need to do a little more to make sure the "${filename} copied" message is announced in a screen reader. Thanks so much for working on this!

KyleAMathews:
Are we using the same code highlighting component as in gatsby-theme-blog?

DSchau:
@KyleAMathews nope. That one is `theme-ui/prism`, which is [here](https://github.com/system-ui/theme-ui/blob/master/packages/prism/src/index.js)

It is fairly similar, but we’re not using theme-ui (yet…) on .org, so I think it makes some sense to have a separate component here.

fk:
AFAIK the UI needs to support four different scenarios:

- plain code blocks — no "language" label, no title
- code blocks with just a title
- code blocks with just a "language" label
- code blocks with "language" label and title

I'm assuming we want to have a "Copy" button for every block of code.

---

1. Just adding a "header" (like the one we show when there's a title) and always show the "Copy" button there would be the easiest, but also the most obstrusive solution, esp. when there's no title:

Code Block_ Language badge only

2. This layout works well when we have title and language badge, but—and this is _very_ nitpicky—I don't really like that the "Copy" button looks like it's associated with the filename:

Code Block_ Title, language badge

This might be the most common scenario in our docs, but let's take a look at some more, e.g. code block showing just one line of code.

3. So here's another variant: Only show the "Copy" button when a user hovers the code block—but that would lock out touch(-only)-device users:

Code Block_ Language badge only, switch to _Copy_ button on hover

4. We also could show the "Copy" button persistent though, like e.g. the [Bootstrap docs](https://getbootstrap.com/docs/4.3/getting-started/introduction/) do, and move the language badge to the left:

Code Block_ Language badge, persistent _Copy_ button a la getbootstrap com

This is not super-duper touch friendly though (irt the button size), but would probably be an acceptable tradeoff (assuming that the majority of our users are on a desktop-like device).

5. So, finally, here's what I ended up with—persistent "Copy" button with slightly increased hozizontal padding; for code blocks with a title, we move the language badge around to be less distracting:

Code Block_ Final

If we'd want the "Copy" button to look more like a button, we could decrease the overall padding:

Code Block_ Final, variant where the button shape is clearer

fk:
Here's a variant of the last approach, with (somewhat) consistent position for the language badge:

Code Block_ Final, variant where the language badge position is consistent

IMO that's a little too busy. We can make that work a little better if we ditch the "per-language" background color:

Code Block_ Final, variant where the language badge position is consistent, and we don't do colo

DSchau:
@fk implemented! 🙌 

button

fk:
Hmm, for some reason our monospace font stack doesn't get applied anymore: https://github.com/gatsbyjs/gatsby/blob/156b7f3dc065dbb012f148e8fe83238beb315cac/www/src/utils/typography.js#L79

DSchau:
@fk I think that it’s a string now? Let me fix that.

DSchau:
This should be ready to go! 🚀

marcysutton:
Wow, nice work @fk and @DSchau! I didn't get a chance to mention this earlier, but I'm wondering if users will look at "Copy" in text and assume it's not interactive, while the language badges with a background color look more like a button. Does anyone else see that?

DSchau:
@marcysutton for sure! @fk and I were definitely inspired by [Bootstrap](https://getbootstrap.com/docs/4.3/getting-started/introduction/?origin_team=T6VN36NMP) here, but if we need it to be more button-y (that’s a word, trust me), we can do that too!

DSchau:
Going to merge! Thanks for the thoughtful review, comments, and design details everyone 🚀

marcysutton:
Thanks for merging Dustin! We can follow up with tweaks and stuff...once it's deployed on the site I'll take a look and see if I have any recommendations to chat through with @fk. This is going to be so helpful for users!

iamskok:
@DSchau You are killing it! Any plans to publish it as a separate package? I will update [gatsby-remark-code-buttons](https://github.com/iamskok/gatsby-remark-code-buttons/) README.md to link to this package.

Blackglade:
@DSchau How is this implemented? I can't seem to find documentation anywhere on how enable a copy button? I see the functionality on this page: https://www.gatsbyjs.org/docs/use-static-query/ but haven't been able to find what to do to implement it

DSchau:
@iamskok definitely on the roadmap! What we’d probably like to do is enhance the theme-ui component to include these line highlighting changes, as well!

@Blackglade sorry — it’s in this PR? You should be able to check out the [files changed](https://github.com/gatsbyjs/gatsby/pull/15834/files) to discover more?

As far as using it — click the button and it’ll copy to your clipboard!

Blackglade:
@DSchau Oops my bad, I was under the impression that this was implemented as a default for all code blocks in Gatsby, but it seems you only did it specific to the GatsbyJS website. Would love to see a generalized plugin as @iamskok mentioned available for everyone!

www/src/components/code-block/index.js
```diff
@@ -0,0 +1,96 @@
...
+/*
+ * MDX passes the code block as JSX
+ * we un-wind it a bit to get the string content
+ * but keep it extensible so it can be used with just children (string) and className
+ */
+export default ({
+  children,
+  className = children.props ? children.props.className : ``,
+}) => {
+  const [language, { title = `` }] = getParams(className)
```
CanRau: @DSchau I love it 😍 and plan to use some of that to add that nice copy feature to coding4.gaiama.org too 👍 🚀 

By the way `mdx` passes additional space separated props as well so you could get the `title` from `props.title` if you pass the title like so
```md
```js title=gatsby-config.js
```
instead of
```md
```js:title=gatsby-config.js
```
then it won't end up in className and you wouldn't have to extract

Yet it's working as is and would be quite a pain to migrate all docs I guess 😅

muescha: FYI: i found your described behaviour here in the mdx docs:

https://mdxjs.com/guides/live-code#code-block-meta-string

CanRau: Uh yeah thanks for the link that's where I know this from 😁🙏

PR #15837, opened 2019-07-17
chore: record e2e in cypress dashboard

pieh:


.circleci/config.yml
```diff
@@ -180,17 +189,23 @@ jobs:
     steps:
       - e2e-test:
           test_path: e2e-tests/production-runtime
-          test_command: yarn test && yarn test:offline
+          test_command: CYPRESS_PROJECT_ID=is8aoq CYPRESS_RECORD_KEY=cb4708d2-1578-4665-9a07-c59f8db28d91 yarn test && CYPRESS_PROJECT_ID=htpvkv CYPRESS_RECORD_KEY=0d734841-c613-41d2-86e5-df0b5968f93f yarn test:offline
```
wardpeet: why is this different from other commands? ^^

pieh: We run 2 separate cypress tests - normal production and production with offline plugin - using env vars here is so results go to different buckets in cypress dashboard. This is not strictly needed, but I think it make things a little bit more organized.

pieh: Alternatively I could use [Grouping test runs](https://docs.cypress.io/guides/guides/parallelization.html#Grouping-test-runs) so normal and offline tests end in same bucket, but have few groups - will try it now

pieh: Ok, grouping seems nicer ( https://dashboard.cypress.io/#/projects/is8aoq/runs/6/specs ). So I'll keep grouping instead of separate projects

pieh: After some tests - grouping mess with re-running tests - so I reverted back to what was before

wardpeet: oh yeah the values differ 🤦‍♂

pieh:
I merged it - won't make it any better for now and no point in delaying it. If it cause problems we can revert

PR #15844, opened 2019-07-17
feat(gatsby-plugin-sharp) Concurrency adjustment for sharp scheduler job queue.

smurrayatwork:
## Description
Currently, gatsby-plugin-sharp only processes one job at a time within the scheduler, causing the generation of a large amount of image thumbnails for a site to take a long time. Increasing this (or allowing adjustments) by using cpu-core-count as provided by the gatsby package significantly decreases the amount of time to process and create these thumbnail images.

Please let me know if I need to add any extra tests/benchmarks to this PR, thanks!
## Related Issues
#15845

### DIFF
--- a/packages/gatsby-plugin-sharp/src/scheduler.js
+++ b/packages/gatsby-plugin-sharp/src/scheduler.js
@@ -3,12 +3,13 @@ const { existsSync } = require(`fs`)
 const queue = require(`async/queue`)
 const { processFile } = require(`./process-file`)
 const { createProgress } = require(`./utils`)
+const cpuCoreCount = require(`gatsby/dist/utils/worker/cpu-core-count`)
 
 const toProcess = {}
 let totalJobs = 0
 const q = queue((task, callback) => {
   task(callback)
-}, 1)
+}, cpuCoreCount())
 
 let bar
 // when the queue is empty we stop the progressbar


pieh:
Concurrent jobs within same process in Node.js will always use single CPU core. You would need to use something like `jest-worker` to create multiple workers (each running in its own process meaning that you can use multiple cores).

Another thing is that `sharp` already make use of multiple cores - it has bindings to native code that isn't limited but single process Node.js execution.

smurrayatwork:
@pieh ,
Thanks for the reply! I agree with what you said, but I'm just adjusting to what functionality the async library provides out of the box, which should be splitting the queued up jobs among multiple workers.

pieh:
Consider having 4 cores - setting concurrent in this job to 4 will cause running 4 sharp for times and each sharp will use 4 threads, meaning it will use 16 threads on 4 cores, which can lead into starving threads and decrease in performance.

Note that this doesn't 100% mean that perf will be decreased. When using 1 concurrent job like we do right now means that that we might not utilize available cores/threads fully as queue scheduler becomes bottleneck (particularly when 1 job finished/next job starts).

All I wanted to say that this is not that straight forward, and also using cpu count for queue concurrency in Node.js won't work as expected, because queue itself still operates in single thread. Check concurrency vs parallelism

smurrayatwork:
@pieh,

I understand. :) My issue was that I didn't realize that async wasn't actually using worker threads but a similar name instead! Either way, I agree with what you're saying. What about the option to potentially increase concurrent with a new, separate, variable, and adding notes in the docs about setting it? That way it's on the user to decide if it's worth trying.

pieh:
One thing that is good approximation if something utilize worker threads or not is API design (of queues in this example) without checking source code. If queue accept function as worker (and not require you to point to a file that implements worker) - it's most likely not using worker threads. Reason for this is that in Node.js you need to serialize parameters when creating worker thread - and functions are not serializable. There are hacks for this - for example you can do something like this:

```
(() => { console.log('hi')}).toString()
```

which will produce string like:
```
() => {
        console.log('hi');
      }
```

which technically you can `eval` and turn back into function, but if inside this function you try to reference anything outside of function scope - it will just fail.

sidharthachatterjee:
Thank you for your work on this @smurrayatwork 

As @pieh mentioned, this wouldn't likely result in an improvement and might actually _degrade_ performance due to thrashing. Since `sharp` already uses multiple cores, I think it's best to set concurrency to 1 here (as it already is). 

Closing this PR at the moment. Thank you so much nonetheless for thinking about this and trying this out! Looking forward to other contributions from you in the future 💜

PR #15848, opened 2019-07-17
chore(docs): Improve readability in environment-variables.md

k-sav:
## Description

An attempt to improve the readability of this intro paragraph, mindful that many readers won't necessarily use english as their native language (Hemingway score has been improved from 12 to 8).

(I must have read it 5 times before being  able to understand what it was getting at.)

If possible, I would also like to replace the phrase _special places_, but I'm not exactly sure what it is referring to. Does anyone have a suggestion?

> There are env variables that are defined in special places intended to be used in different deployment environments.

## Related Issues

None


k-sav:
I just noticed that this is failing a prettier test. Do I need to clone the repo in order to do that? 

[Options for contributing to the Gatsby docs
](https://www.gatsbyjs.org/contributing/docs-contributions/#options-for-contributing-to-the-gatsby-docs) suggests simply editing directly in the GitHub UI, which is what I have done.

PR #15852, opened 2019-07-17
test(gatsby): assert shadow support of org scoped packages

adrw:
## Description

Added test to ensure support of [NPM org scoped packages](https://docs.npmjs.com/about-org-scopes-and-packages) with Gatsby Themes component shadowing logic.

### DIFF
--- a/packages/gatsby/src/internal-plugins/webpack-theme-component-shadowing/__tests__/index.js
+++ b/packages/gatsby/src/internal-plugins/webpack-theme-component-shadowing/__tests__/index.js
@@ -4,12 +4,14 @@ import ShadowingPlugin from "../"
 describe(`Component Shadowing`, () => {
   it(`gets matching themes`, () => {
     const plugin = new ShadowingPlugin({
-      themes: [`a-theme`, `theme-b`, `gatsby-theme-c`].map(name => {
-        return {
-          themeName: name,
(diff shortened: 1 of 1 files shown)


sidharthachatterjee:
So it seems that this was missing something and broke the tests (which I fixed in https://github.com/gatsbyjs/gatsby/commit/c7e2361ffbe55ae69d95e977715e4bc847058dd0) but I'm not sure why it wasn't caught in the checks for this PR

Any ideas @pieh @DSchau ?

DSchau:
@sidharthachatterjee same reason as others! process.exit was invoked before other (failing) tests

```
prefetch
      ✓ shouldn't prefetch when shouldPrefetch is false
      ✓ should prefetch when not yet triggered (2ms)
      ✓ should only run apis once (2ms)

  ●  process.exit called with ""

      168 |   const watchEvents = [`change`, `add`]
      169 | 
    > 170 |   chokidar
          |           ^
      171 |     .watch(watchers, {
      172 |       ignored: [filePath => _.some(ignored, reg => reg.test(filePath))],
      173 |     })
```

sidharthachatterjee:
@DSchau 😢 

@gatsbyjs/core Please remember to merge master into all current PRs before you _trust_ green checks!

PR #15856, opened 2019-07-17
feat(gatsby): enable local themes

ChristopherBiscardi:


pieh:
Hey @ChristopherBiscardi! What's the status on this? Is this generally good to go?

If so, I can take over and add basic test for e2e themes to cover local plugin and get this in?

ChristopherBiscardi:
@pieh It's generally good to go I tested it and it worked last time I ran it. I feel it needs a bit more testing so that the error messages are more useful. (like, "gatsby checked X and Y locations but didn't find plugin Z" or something)

Happy for you to take this over with e2e tests! Thanks for the offer :)

pieh:
> Happy for you to take this over with e2e tests! Thanks for the offer :)

That's pushed (finally, sorry for delay on this)

> @pieh It's generally good to go I tested it and it worked last time I ran it. I feel it needs a bit more testing so that the error messages are more useful. (like, "gatsby checked X and Y locations but didn't find plugin Z" or something)

This is still on TO-DO (original debugging logs are still there)

pieh:
Additional notes that I just started wondering - I think current code will support local themes only in main site's `plugins` dir - is this something that should be allowed in nested themes? If so, I can do work on that

pieh:
> Additional notes that I just started wondering - I think current code will support local themes only in main site's `plugins` dir - is this something that should be allowed in nested themes? If so, I can do work on that

After some chat I decided against it. This could produce weird conflicts with shadowing when multiple themes have local theme with the same name, but are actually different things.

Consider scenario we have theme `gatsby-theme-a` and `gatsby-theme-b` and both of those have local theme `gatsby-theme-local` that have different component in `src/components/foo.js` - how would you shadow it and target only one of those themes? It seems non-deterministic. One way would be to not allow loading themes with same names multiple times, but then theme consumers would be left with issue they can't resolve. So for now (until problem is figured out), we will load local themes only from main site.

sidharthachatterjee:
> So for now (until problem is figured out), we will load local themes only from main site.

@pieh Sounds good! Does that mean this is good to go in for now then?

pieh:
Not yet, I need to display actual proper error as mentioned in https://github.com/gatsbyjs/gatsby/pull/15856#issuecomment-522186393

packages/gatsby/src/bootstrap/load-themes/index.js
```diff
@@ -5,25 +5,46 @@ const _ = require(`lodash`)
...
   let themeDir
   try {
+    // theme is an node-resolvable module
     themeDir = path.dirname(require.resolve(themeName))
   } catch (e) {
-    // this can be local plugin, and require.resolve will throw
-    // in this case - let's return partial entry
-    return { themeName, themeSpec }
+    // is a local plugin OR it doesn't exist
+    const pathToLocalTheme = path.join(path.resolve(`.`), `plugins`, themeName)
```
pieh: we always resolve to main site local plugin which results in error like this (if plugin declared by some theme can't be found):
```
Couldn't find "wat" plugin declared in "/Users/misiek/dev/gatsby/e2e-tests/themes/gatsby-theme-about/gatsby-config.js".

Tried looking for a local plugin in /Users/misiek/dev/gatsby/e2e-tests/themes/production-runtime/plugins/wat
```
So plugin is declared in theme but we look for it in main site local plugins.

This is very confusing - I think we should not even look for local plugin in those cases, and adjust error message (on it right now)

packages/gatsby/src/bootstrap/load-themes/index.js
```diff
@@ -5,25 +5,66 @@ const _ = require(`lodash`)
...
-    // in this case - let's return partial entry
-    return { themeName, themeSpec }
+    let panic = false
+    let pathToLocalTheme
+
+    // only try to look for local theme in main site
+    // local themes nested in other themes is potential source of problems:
+    // because those are not hosted by npm, there is potential for multiple
+    // local themes with same name that do different things and name being
+    // main identifier that gatsby uses right now, it's safer not to support it for now.
```
sidharthachatterjee: Ah, this should go in docs somewhere

packages/gatsby/src/bootstrap/load-themes/index.js
```diff
@@ -5,25 +5,66 @@ const _ = require(`lodash`)
...
+      try {
+        themeDir = path.dirname(require.resolve(pathToLocalTheme))
+      } catch (localErr) {
+        panic = true
+      }
+    } else {
+      panic = true
+    }
+
+    if (panic) {
```
sidharthachatterjee: Nit: Moving this to finally might be a little more readable

packages/gatsby-cli/src/structured-errors/error-map.js
```diff
@@ -160,6 +160,21 @@ const errorMap = {
     type: `CONFIG`,
     level: `ERROR`,
   },
+  "10226": {
+    text: context =>
+      [
+        `Couldn't find the "${context.themeName}" plugin that is declared in "${context.configFilePath}".`,
```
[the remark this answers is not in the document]
sidharthachatterjee: Changed my mind here 😛

packages/gatsby/src/bootstrap/load-themes/index.js
pieh: did some cleanup here (hopefully)

PR #15869, opened 2019-07-18
chore(showcase): Removing Devol's Dance from showcase

lannonbr:
## Description

@coreyward just for notice. Devol's Dance is now using YUI for their splash page so we are going to remove it from the showcase. We are always welcome to adding sites back if they switch back to Gatsby.


coreyward:
@lannonbr Bummer! Thanks for the heads up.

Issue #14662, opened 2019-06-09
Gatsby Image Offset For Lazy Loading


lnikell:
**Problem**
Using Gatsby Image the images only start loading when you scroll to the object, so it always causes the issue that for few ms the image is not visible and then it appears from blur/or with fade effect. 

**Expected behaviour**
I want to make sure that there is some offset definition to make sure that if I'm currently on the first section of the page, the second section already will load images, then when I'm on 2nd section, the 3rd section images will start loading.

Currently, I could either define loading images immediately(set critical attribute) without waiting for scroll or default lazy loading which triggers actually only when the image appears in the viewport, without any offset.

freiksenet:
There is already [root margin specified](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-image/src/index.js#L97) for Gatsby Image intersection observer, so it should happen before hand. It's possible to make a PR to add a configuration for it in gatsby-image API. However, 200px is not a small value, so I wonder if you could try it first by increasing this value by editing it in `node_modules` directly.

lnikell:
Thank you @freiksenet! I will definitely try it and probably will get back with PR 😀

lnikell:
So the conclusion is that rootMargin works correctly but unfortunately 200px is very small. 
Proposal: It should be the customizable property

Content on modern websites usually section based, each section usually are about full height of the screen, users also scroll by sections rather than just by 200px. In that case, if you have  images in the section 2 closer to the bottom of the section, the user always will experience loading/fade in effect when he scrolls through website

![image](https://user-images.githubusercontent.com/2697570/60336398-8bc0db80-99a0-11e9-9600-2417f4dcd898.png)

0tho:
Hello, I was looking something to contribute to gatsby and found this issue, so I made the [following](https://github.com/gatsbyjs/gatsby/compare/master...0tho:14662-IntersectionObserverOptions?expand=1) changes enabling to reconfigure IO as wanted. I still couldn't test it, so I haven't opened a PR yet.

Any comments are appreciated

freiksenet:
@0tho This looks good, could you make a PR so that we can make a review?

proAlexandr:
Hi, I have created a PR for this feature (https://github.com/gatsbyjs/gatsby/pull/16004). Could someone, please, review it?

PR #15871, opened 2019-07-18
(WIP) Allow to configure Intersection Observer

0tho:
## Description

Enhance gatsby-image IO interface.

Based on MDN documentation of Intersection Observer(IO) add a prop to
Image that allows to reconfigure entirely how the image instace of IO
behaves.

Following the documentation there are 3 possible properties
- root
- rootMargin
- threshold

So the new prop, ioOptions, allows an object containing only this
options to be passed down.

https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API

## Related Issues

#14662 

This PR still needs some testing

### DIFF
--- a/packages/gatsby-image/src/index.js
+++ b/packages/gatsby-image/src/index.js
@@ -88,7 +88,7 @@ const hasIOSupport = isBrowser && window.IntersectionObserver
 let io
 const listeners = new WeakMap()
 
-function getIO() {
+function getIO({ root = null, rootMargin = "200px", threshold = 0 } = {}) {
   if (
     typeof io === `undefined` &&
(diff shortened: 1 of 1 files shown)


wardpeet:
Thank you for opening this Pull Request.

Sadly we're not going to merge this one. Lazy loading is coming to the web and passing down `ioOptions` won't have any effect anymore for browsers that support this technique. The browser will figure out the best behaviour on when to load images.

More information can be found https://addyosmani.com/blog/lazy-loading/ or https://chromestatus.com/feature/5645767347798016

@0tho would love to see people experimenting with different values other than 200px — if we decide a different value works better, we can definitely change it to something else.

PR #15876, opened 2019-07-18
fix(gatsby): don't break if an ancestor directory is inaccessible

### LINKED ISSUE #14962: Gatsby build fails on shared hosting where listing of folder content (e.g /home) is not allowed
## Description

I've tried running `gatsby build` on my shared hosting, but the build failed. More specifically, it seems to fail due to the fact that one of the parent directories (in this case `/home`) does not permit listing of files. The `true-case-path` library, used by gatsby, throws an error because it is unable to `scandir` a component path's parent folder (in this case the component path is `/home/myusername/gatsby-site/src/pages/404.js`). I think the build should proceed in this case, and handle this error silently.

### Steps to reproduce

```
cd /home/myusername
npm install -g gatsby
gatsby new gatsby-site
cd gatsby-site
npm run build
```

### Expected result

Build should complete successfully. If parent directories do not allow files to be listed, the build should still proceed.

### Actual result

```
[myusername@server gatsby-site]$ npm run build

> gatsby-starter-default@0.1.0 build /home/myusername/gatsby-site
> gatsby build

success open and validate gatsby-configs - 0.011 s
success load plugins - 0.491 s
success onPreInit - 0.010 s
success delete html and css files from previous builds - 0.010 s
success initialize cache - 0.021 s
success copy gatsby files - 0.043 s
success onPreBootstrap - 0.012 s
success source and transform nodes - 0.073 s
success building schema - 0.312 s
success createPages - 0.002 s
/home/myusername/gatsby-site/src/pages/404.js
error UNHANDLED REJECTION


  Error: EACCES: permission denied, scandir '/home'

  - sync.js:288 GlobSync._readdir
    [gatsby-site]/[glob]/sync.js:288:41

  - sync.js:137 GlobSync._processReaddir
    [gatsby-site]/[glob]/sync.js:137:22

  - sync.js:132 GlobSync._process
    [gatsby-site]/[glob]/sync.js:132:10

  - sync.js:207 GlobSync._processReaddir
    [gatsby-site]/[glob]/sync.js:207:10

  - sync.js:132 GlobSync._process
    [gatsby-site]/[glob]/sync.js:132:10

  - sync.js:48 new GlobSync
    [gatsby-site]/[glob]/sync.js:48:10

  - sync.js:26 Function.globSync [as sync]
    [gatsby-site]/[glob]/sync.js:26:10

  - index.js:29 trueCasePathSync
    [gatsby-site]/[true-case-path]/index.js:29:15

  - public.js:198 actions.createPage
    [gatsby-site]/[gatsby]/dist/redux/actions/public.js:198:41

  - redux.js:468
    [gatsby-site]/[redux]/lib/redux.js:468:35

  - api-runner-node.js:68 doubleBoundActionCreators.(anonymous function).args
    [gatsby-site]/[gatsby]/dist/utils/api-runner-node.js:68:20

  - gatsby-node.js:141 _createPage
    [gatsby-site]/[gatsby-plugin-page-creator]/gatsby-node.js:141:3

  - gatsby-node.js:76
    [gatsby-site]/[gatsby-plugin-page-creator]/gatsby-node.js:76:22

  - Array.forEach

  - gatsby-node.js:75 Object._callee$
    [gatsby-site]/[gatsby-plugin-page-creator]/gatsby-node.js:75:19
```

### Environment

```
  System:
    OS: Linux 3.10 CentOS Linux 7 (Core)
    CPU: (48) x64 Intel(R) Xeon(R) CPU E5-2650 v4 @ 2.20GHz
    Shell: 4.2.46 - /bin/bash
  Binaries:
    Node: 10.15.1 - ~/webapps/nodejs/bin/node
    npm: 6.4.1 - ~/webapps/nodejs/bin/npm
  Languages:
    Python: 2.7.5 - /usr/local/bin/python
  npmPackages:
    gatsby: ^2.9.11 => 2.9.11
    gatsby-image: ^2.1.4 => 2.1.4
    gatsby-plugin-offline: ^2.1.3 => 2.1.3
    gatsby-plugin-sharp: ^2.1.9 => 2.1.9
    gatsby-source-filesystem: ^2.0.43 => 2.0.43
    gatsby-transformer-sharp: ^2.1.21 => 2.1.21
  npmGlobalPackages:
    gatsby: 2.10.0
```

waltercruz:
## Description

Fix builds on hosts where the user doesn't have access to the entire folder structure until /

## Related Issues

Fixes #14962

## Needs help

I still need help on ideas on what to do on the catch part. Any suggestions?


sidharthachatterjee:
I would prefer if this was handled in true-case-path to be honest https://github.com/barsh/true-case-path/issues/2

caseyWebb:
Developer of true-case-path (moved to Profiscience/true-case-path) here.

Ideally this would be handled in true-case-path, but the only solution I can think of is providing the ability to pass an argument for the "known good base path" which will be omitted from case-correction. I looked over this PR, but I'm not familiar enough with the problem to know if this is a solution that can be applied here. Is this a situation where Gatsby would be able to provide this (by means of something like __dirname or process.cwd() to get a known good relative path to work from)?

FWIW, the issue is that the way true-case-path works is it lists all of the directories for each segment of the file path and then uses regex to compare.

waltercruz:
Hi @caseyWebb ! I'm really a gatsby core developer (was just trying to fix my case), but I'm sure that it's possible do use __dirname. I like the idea about a "known good base directory" that you talked about answering my email.

packages/gatsby/src/redux/actions/public.js
```diff
@@ -258,32 +258,36 @@ ${reservedFields.map(f => `  * "${f}"`).join(`\n`)}
...
-          )
-          hasWarnedForPageComponentInvalidCasing.add(page.component)
+            )
+            hasWarnedForPageComponentInvalidCasing.add(page.component)
+          }
         }
-
         page.component = trueComponentPath
+      } catch (e) {
+        console.log(e)
```
sidharthachatterjee: Can we print a richer error here with some more context?

sidharthachatterjee: Also currently it looks like this will be printed for every instance of this issue. Maybe we want to set a flag and only print this once?

caseyWebb:
true-case-path v2.2.0 adds support for an optional `basePath` argument. When provided, the first argument is expected to be the file path to be checked relative from the known good base path.

waltercruz:
Hi @sidharthachatterjee! I updated to the latest true-case-path, that uses the optional basePath. It works ok now - tested on my local and on my dev server, where the issue happened. But there are two failing tests on circleci and two tests on local fail too :( 

the local tests that fail are these on packages/gatsby/src/schema/extensions/__tests__/field-extensions.js and I don't see how they could be related to the commit here.

any tips?

pieh:
> Hi @sidharthachatterjee! I updated to the latest true-case-path, that uses the optional basePath. It works ok now - tested on my local and on my dev server, where the issue happened. But there are two failing tests on circleci and two tests on local fail too :(

If you look at the path that true-case-path fails with - it's `e2e-tests/themes/gatsby-theme-about/src/pages/about.js,`

And actual CWD is `e2e-tests/themes/production-runtime` (or `development-runtime`), so path to page components is not inside current CWD (which I guess is why `true-case-path` throws error). It seems to me that `cwd` is not best `basePath` here. I'm not sure if it's possible to determine what should be actual `basePath` - maybe gatsby (or true-case-path) could walk directories up to the point it can't list files anymore and use last path that it could list files as `basePath`?

Also this tests fail correctly - this change has potential to break themes in yarn workspaces - so I'm glad this was caught.

> the local tests that fail are these on packages/gatsby/src/schema/extensions/**tests**/field-extensions.js and I don't see how they could be related to the commit here.

Ok, so your branch was created when we had some issues with running unit tests in CircleCI - if you look at tests that passed (for example - https://circleci.com/gh/gatsbyjs/gatsby/180224?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) - it didn't actually run all tests - it crashed, but CircleCI still didn't fail.

You should be able to merge master in
```
git pull <remote-name-for-gatsbyjs/gatsby> master
```

And push it to your fork - and then we might see same failure here and investigate what is happening

waltercruz:
Hi @sidharthachatterjee and @pieh, what do you guys think about this version:

using the original trueCasePath, that should work on most systems, and in case of exception, using trueCasePathSync with the optional basePath, that takes care of the case where user doesn't have access to the / directory (shared host, etc).

pieh:
Yeah - it's nicer. But it will still cause crashes in some cases (yarn workspaces + not being able to list dirs/files in `/`.

What about this - instead of using `process.cwd()` / `state.program.directory` as basePath, we figure out common directory that `state.program.directory` and `page.component`. As example:

```
const projectDirectory = state.program.directory
// for exercise, let's assume projectDirectory is "root/project/e2e-tests/themes/development-runtime"
// Let's assume page.component is "root/project/e2e-tests/themes/gatsby-theme-about/src/pages/about.js"
// we do need to get common directory path - which we should be able to list files from - otherwise there will be lot other problems anyway.

const commonDir = getCommonDir(state.program.directory, page.component)
// getting common directory from paths:
// - root/project/e2e-tests/themes/development-runtime
// - root/project/e2e-tests/themes/gatsby-theme-about/src/pages/about.js
// = root/project/e2e-tests/themes
```

waltercruz:
> Yeah - it's nicer. But it will still cause crashes in some cases (yarn workspaces + not being able to list dirs/files in /.

any tips to create such environment on local @pieh?

waltercruz:
What do you guys think about now?

packages/gatsby/src/redux/actions/public.js
```diff
@@ -260,7 +270,22 @@ ${reservedFields.map(f => `  * "${f}"`).join(`\n`)}
...
+      try {
+        // most systems
+        trueComponentPath = slash(trueCasePathSync(page.component))
+      } catch {
+        // systems where user doesn't have access to /
+        const commonDir = getCommonDir(
+          store.getState().program.directory,
+          page.component
+        )
+        const relativePath = page.component
```
pieh: There is helper in Node for this:
```
const relativePath = path.relative(commonDir, page.component)
```

packages/gatsby/src/redux/actions/public.js
```diff
@@ -35,6 +35,16 @@ const findChildrenRecursively = (children = []) => {
...
 
+const getCommonDir = (path1, path2) => {
+  var iChar
+  for (iChar = 0; iChar < path1.length; iChar += 1) {
+    if (path1[iChar] !== path2[iChar]) {
+      return path1.substring(0, iChar)
+    }
+  }
+  return path1
+}
```
pieh: Comparing paths character by character is potentially flaky:
 - there might be weirdness on Windows if one path contains slashes and other path contains backslashes (tho to be fair, we shouldn't hit this code path on Windows ever (`/` being inaccessible).
 - There might be casing difference that would cause characters to differ, but still should be considered same path.

I'll try to work it out, so we can merge this today

sidharthachatterjee:
Thanks so much @waltercruz and for finishing this up @pieh 👍

sidharthachatterjee:
Published in `gatsby@2.13.66`

Issue #14983, opened 2019-06-20
[docs][recipes] Querying data: aliases


marcysutton:
## Summary

To improve key learning workflows in the Recipes section of the docs, there should be a short, actionable recipe on using aliases with GraphQL. See also: #14977, #14979, #14980, #14981, and #14982 for previous query recipes.

Aliases make it easier to include multiple queries of the same type, and to identify which query is which. This recipe should show how to add aliases to a GraphQL query _inline_ as concisely as possible with a code snippet and most likely, a GraphiQL embed like on the [reference doc](https://www.gatsbyjs.org/docs/graphql-reference/#aliasing).

Recipes now follow this format:

- Requirements or prerequisites 
- Step-by-step instructions
- Relevant links to any relevant information
    - Gatsby GraphQL reference for aliasing: https://www.gatsbyjs.org/docs/graphql-reference/#aliasing
    - Link to a live example

This recipe should go into the "Querying data" section of [`docs/docs/recipes.md`](https://github.com/gatsbyjs/gatsby/blob/master/docs/docs/recipes.md), replacing the existing text. There is a recipe template in the contributing docs: https://www.gatsbyjs.org/contributing/docs-templates/#recipes

For reference, the new recipe collection is also listed in a spreadsheet: https://docs.google.com/spreadsheets/d/1yNUQi_-3TsuDtE98hCIJQvLZ6X1YHlDmC-EJwOGxiHk/edit#gid=1252007048

marcysutton:
Closed with #15882

Issue #14981, opened 2019-06-20
[docs][recipes] Querying data: sorting


marcysutton:
## Summary

To improve key learning workflows in the Recipes section of the docs, there should be a short, actionable recipe on sorting data with GraphQL. See also: #14977, #14979, and #14980 for previous query recipes.

Common use cases include sorting content by date or alphabetical titles, either ascending or descending order. This recipe should show how to sort data with GraphQL _inline_ as concisely as possible with code snippets and most likely, a GraphiQL embed like on the [reference doc](https://www.gatsbyjs.org/docs/graphql-reference/#sort).

Recipes now follow this format:

- Requirements or prerequisites 
- Step-by-step instructions
- Relevant links to any relevant information
    - Gatsby GraphQL reference for sorting: https://www.gatsbyjs.org/docs/graphql-reference/#sort
    - Link to a live example

This recipe should go into the "Querying data" section of [`docs/docs/recipes.md`](https://github.com/gatsbyjs/gatsby/blob/master/docs/docs/recipes.md), replacing the existing text. There is a recipe template in the contributing docs: https://www.gatsbyjs.org/contributing/docs-templates/#recipes

For reference, the new recipe collection is also listed in a spreadsheet: https://docs.google.com/spreadsheets/d/1yNUQi_-3TsuDtE98hCIJQvLZ6X1YHlDmC-EJwOGxiHk/edit#gid=1252007048

marcysutton:
Closed with #15882

Issue #14980, opened 2019-06-20
[docs][recipes] Querying data: filtering


marcysutton:
## Summary

To improve key learning workflows in the Recipes section of the docs, there should be a short, actionable recipe on filtering data with GraphQL. See also: #14977 and #14979 for previous query recipes.

Common use cases for filtering include filtering content by type (published or draft, post or page, categories, with regex, etc.). This recipe should show how to filter data with GraphQL _inline_ as concisely as possible with code snippets and most likely, a GraphiQL embed like on the [reference doc](https://www.gatsbyjs.org/docs/graphql-reference/#filter). If this needs to be broken up into multiple task-based recipes, that would be fine too.

Recipes now follow this format:

- Requirements or prerequisites 
- Step-by-step instructions
- Relevant links to any relevant information
    - Gatsby GraphQL reference for filtering: https://www.gatsbyjs.org/docs/graphql-reference/#filter
    - Link to a live example

This recipe should go into the "Querying data" section of [`docs/docs/recipes.md`](https://github.com/gatsbyjs/gatsby/blob/master/docs/docs/recipes.md), replacing the existing text. There is a recipe template in the contributing docs: https://www.gatsbyjs.org/contributing/docs-templates/#recipes

For reference, the new recipe collection is also listed in a spreadsheet: https://docs.google.com/spreadsheets/d/1yNUQi_-3TsuDtE98hCIJQvLZ6X1YHlDmC-EJwOGxiHk/edit#gid=1252007048

marcysutton:
Closed with #15882

PR #15882, opened 2019-07-18
[docs][recipes] GraphQL arguments

gillkyle:
## Description

Adds recipes for:
- limiting (#14982) 
- sorting (#14981)
- filtering (#14980)
- aliasing (#14983)

Getting a really brief length for this is a little tricky and managing the overlap between them is something worth discussing. Since the prerequisites are fairly similar maybe there'd be a better way to format these. Open to any suggestions as always!

<details>
<summary>Details</summary>

image

</details>


docs/docs/recipes.md
```diff
@@ -218,6 +218,220 @@ export default NonPageComponent
...
+      node {
+        id
+        path
+      }
+    }
+  }
+}
+```
+
+4. Add a `sort` argument to the `allSitePage` field and give it an object with the a `fields` and `order` attribute. The value for `fields` can be a field or an array of fields to sort by (this example uses the `path` field), the `order` can be either `ASC` or `DESC` for ascending and descending respectively.
```
stefanprobst: Should we mention that `order` can also be an array -- i.e. when sorting by an array of fields like `fields: [first, second]`, it is possible to sort by `first` in ascending order, and `second` in descending order with `order: [ASC, DESC]`

gillkyle: I do like that idea, I battled with including a lot of more details (and even including as much detail as I did) with the format of the recipes though since they're meant to be short. Is it better to just limit to one really simple use case or spell out some more complicated ones too. I aired more on the side of the latter for this one.

marcysutton: Yes - i'd say let's keep this simple to start and we can add more detail in the GraphQL docs guides.

marcysutton:
I made some content updates to get this updated with master. I also made some updates to the prerequisites - let me know what you think. Once we have sourcing recipes it will be easier to set up things (like sourcing `allMarkdownRemark` nodes), and we can show rather than tell.

gillkyle:
I love the updates you made @marcysutton, I felt like my descriptions of the recipes were a little redundant and adding in words like "collection" of nodes make them stand alone better. 👍

PR #15890, opened 2019-07-19
fix(www): potentially fix intermittent screenshot failures

DSchau:
## Description

I haven’t been able to reproduce locally, but this _may_ help guard against screenshot failures.

I also did two other things (for longer term fixes):

1. Added schema customization for known yaml types; this will let us “trim down” sites and showcases to smaller subsets without breaking the builds
1. Moved the featured 5 showcases/starters to the top, this will let us trim the yaml as part of a simpler develop mode (e.g. a script to only use the first 5 entries and then pop them back on before committing)


docs/starters.yml
```diff
@@ -1,3 +1,69 @@
+- url: https://gatsby-starter-blog-demo.netlify.com/
```
wardpeet: What are these changes?

sidharthachatterjee: Looks like he has moved 5 ones that always work to the top

DSchau: Ha -- I'm so sorry I missed these!

I did this for a reason :) I'd like to eventually make it easier to get up and running with .org, and popping off the top 5 seems like an easy approach 🎉

PR #15907, opened 2019-07-19
fix(gatsby-cli): fix cut off text

### LINKED ISSUE #14658: CLI text seems to be cut off with Ink
I’m not sure if the switch to Ink is the culprit, but I’ve started seeing messages in the CLI get cut off since about the time Ink rolled out.

For example, the rest of this `info` notice is missing from the output:

Screen Shot 2019-06-08 at 5 52 43 PM

I’ve seen this a few times — I can’t remember exactly where, but I _think_ it was error messages.

wardpeet:
## Description
Fixes ink not showing all text.

Old:
image

New:
image

## Related Issues
Fixes #14658

### DIFF
--- a/packages/gatsby-cli/package.json
+++ b/packages/gatsby-cli/package.json
@@ -55,7 +55,7 @@
     "cross-env": "^5.1.4"
   },
   "optionalDependencies": {
-    "ink": "^2.0.5",
+    "ink": "^2.3.0",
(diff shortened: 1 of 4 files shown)


sidharthachatterjee:
Published in `gatsby-cli@2.7.17` and `gatsby@2.13.31`

Issue #15918, opened 2019-07-19
touchNode does not populate typeOwners resulting errors when deleting cached nodes


datakurre:
## Description

We have a source plugin "that can return only nodes that have updated" and can also delete removed nodes while on `gatsby develop`. When I

1. Run `gatsby build` to fetch all nodes once

2. immediately start `gatsby develop` resulting mostly touchNode-calls, and

3. delete a page from the source system

I get an error that exists gatsby develop:

```
  Error: The plugin "..." deleted a node of a type owned by another plugin.
  The node type "..." is owned by "undefined".
```

The reason for this seems to be that `typeOwners` are only populated when source plugin is calling `createNode` (not from `touchNode`). If no `createNode` has been called for the type of the node that is being deleted, it does not have owner yet (undefined) and deletion raises an error.

I would expect that calling `touchNode` would also populate `typeOwners`.

### Environment

```

  System:
    OS: Linux 5.1 NixOS 19.03.173068.9ec7625cee5 (Koi) 19.03.173068.9ec7625cee5 (Koi)
    CPU: (4) x64 Intel(R) Core(TM) i5-2540M CPU @ 2.60GHz
    Shell: 4.4.23 - /nix/store/93h01q6yg13xdrabvqbddzbk11w6a928-bash-interactive-4.4-p23/bin/bash
  Binaries:
    Node: 10.15.3 - /run/user/1000/yarn--1563572643943-0.09144924431626666/node
    Yarn: 1.13.0 - /run/user/1000/yarn--1563572643943-0.09144924431626666/yarn
    npm: 6.4.1 - /nix/store/ll4pkaa7ajycqippwp47yypqy5b3x58h-nodejs-10.15.3/bin/npm
  Languages:
    Python: 3.7.3 - /nix/store/gyffk0z9rwcdl1l1654z3b0sry98ddrk-python3-3.7.3/bin/python
  Browsers:
    Firefox: 67.0.4
  npmPackages:
    gatsby: 2.13.8 => 2.13.8 
    gatsby-image: ^2.2.6 => 2.2.6 
    gatsby-plugin-react-helmet: ^3.1.2 => 3.1.2 
    gatsby-plugin-sass: ^2.1.3 => 2.1.3 
    gatsby-plugin-sharp: ^2.2.8 => 2.2.8 
    gatsby-source-filesystem: ^2.1.5 => 2.1.5 
    gatsby-transformer-sharp: ^2.2.4 => 2.2.4 
```

[...]

PR #15919, opened 2019-07-19
fix(gatsby): Add touchNode to populate typeOwners (#15918)

datakurre:
## Description

Gatsby keeps track of different node types by different source plugins and allows source plugins to only manage nodes of types they own. Unfortunately typeOwners mapping is only populated when nodes are being created (by createNode API). If nodes of certain type have only been touched so far (with touchNode), no source plugin owns that node type. This prevents the real owner of that node type from deleting cached nodes until it has created at least a single node of that type.

This seems to be the minimal change to make touchNode populate typeOwners mapping, but I don't know if that's the correct one.

## Related Issues

https://github.com/gatsbyjs/gatsby/issues/15918

### DIFF
--- a/packages/gatsby/src/redux/actions/public.js
+++ b/packages/gatsby/src/redux/actions/public.js
@@ -792,6 +792,11 @@ actions.touchNode = (options: any, plugin?: Plugin) => {
     nodeId = options
   }
 
+  const node = getNode(nodeId)
+  if (node && !typeOwners[node.internal.type]) {
+    typeOwners[node.internal.type] = node.internal.owner
+  }
+
   return {
     type: `TOUCH_NODE`,
     plugin,


pieh:
I see 2 potential alternatives to achieve same thing:
1. Persist typeOwners together with nodes in redux state
2. Re-concstuct typeOwners when loading cached redux state.

But this is also correct, so let's merge this.

pieh:
Published `gatsby@2.13.34`

datakurre:
@pieh FYI. Similar issue may still happen when a removal of node is detected in incremental build and no node of the same type has been created or touched before.

pieh:
Oh, very interesting - so maybe we do need to revisit solution and actually persist typeOwners

juanpprieto:
> Oh, very interesting - so maybe we do need to revisit solution and actually persist typeOwners

Any updates/plans on persisting typeOwner?

Issue #15676, opened 2019-07-13
[gatsby-remark-images]: backgroundColor option not working for transparent pngs


danielj41:
## Description

I have the following `gatsby-config.js`, setting `gatsby-remark-images`'s backgroundColor option to `#222`. 

```js
    {
      resolve: `gatsby-transformer-remark`,
      options: {
        plugins: [
          {
            resolve: `gatsby-remark-images`,
            options: {
              maxWidth: 590,
              backgroundColor: "#222",
            },
          },
        ]
      }
    }
```

I expect that to make the transparent portion of my pngs show up as `#222`, but they remain the default `white`.

### Steps to reproduce

- Check out demo project https://github.com/danielj41/blog-gatsby/blob/demo-for-issue/gatsby-config.js#L36-L37 .
- Check out branch `demo-for-issue`.
- Run `yarn build`, serve `public`. (`gatsby develop` was giving me inconsistent results)
- Load `/`. Click the "dark mode" checkbox.
- Load `/js-strict-mode`.

### Expected result

The transparent portion of the png should match the page background (`#222`)

<img width="667" alt="Screen Shot 2019-07-12 at 10 44 22 PM" src="https://user-images.githubusercontent.com/5705347/61167673-09373080-a4f7-11e9-822d-0dc95bd46369.png">

### Actual result

The transparent portion of the png shows up as `white`, the default background color setting.

<img width="696" alt="Screen Shot 2019-07-12 at 10 44 53 PM" src="https://user-images.githubusercontent.com/5705347/61167677-16541f80-a4f7-11e9-9b8e-62d4a3211884.png">

### Suggested fix

I was able to get it working as expected by changing https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-remark-images/src/gatsby-browser.js#L8 from

```js
exports.onRouteUpdate = ({ pluginOptions }) => {
```

to

```js
exports.onRouteUpdate = (apiCallbackContext, pluginOptions) => {
```

That said, I don't know Gatsby well enough to say if that's the correct fix.

### Environment

```

  System:
    OS: macOS 10.14.5
    CPU: (8) x64 Intel(R) Core(TM) i7-8569U CPU @ 2.80GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.16.0 - ~/.nvm/versions/node/v10.16.0/bin/node
    Yarn: 1.17.0 - /usr/local/bin/yarn
[...]

PR #15940, opened 2019-07-20
gastby-remark-images: fix onRouteUpdate signature

danielj41:
## Description

According [to the `onRouteUpdate` docs](https://www.gatsbyjs.org/docs/browser-apis/#onRouteUpdate), the correct signature is `(apiCallbackContext, pluginOptions)`. Change

```js
exports.onRouteUpdate = ({ pluginOptions }) => {
```

to

```js
exports.onRouteUpdate = (apiCallbackContext, pluginOptions) => {
```

With the old signature, `({ pluginOptions })`, `pluginOptions` will always be undefined, causing the default options to be used. This causes the `box-shadow` later in the file to always be `white`, the default `options.backgroundColor`.

## Related Issues

Fixes https://github.com/gatsbyjs/gatsby/issues/15676. See that issue for bug repro instructions.

cc @birjolaxew

### DIFF
--- a/packages/gatsby-remark-images/src/__tests__/gastby-browser.js
+++ b/packages/gatsby-remark-images/src/__tests__/gastby-browser.js
@@ -1,0 +1,37 @@
+const {
+  imageClass,
+  imageWrapperClass,
+  imageBackgroundClass,
+} = require(`../constants`)
+const { onRouteUpdate } = require(`../gatsby-browser`)
+
+const createImageElement = () => {
+  global.document.body.innerHTML = `
+    <span class="${imageWrapperClass}">
+      <span class="${imageBackgroundClass}"></span>
+      <img class="${imageClass}">
+    </span>
+  `
+
+  return global.document.querySelector(`.${imageClass}`)
+}
+
+test(`it sets correct box-shadow on img element`, () => {
+  const pluginOptions = {}
+  const imageElement = createImageElement()
+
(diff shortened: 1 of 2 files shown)


packages/gatsby-remark-images/src/__tests__/gastby-browser.js
```diff
@@ -0,0 +1,37 @@
...
+
+test(`it sets correct box-shadow on img element`, () => {
+  const pluginOptions = {}
+  const imageElement = createImageElement()
+
+  onRouteUpdate({}, pluginOptions)
+  imageElement.dispatchEvent(new Event(`load`))
+
+  expect(imageElement.style[`box-shadow`]).toBe(`inset 0px 0px 0px 400px white`)
+})
```
danielj41: @birjolaxew [mentioned that it'd be nice to not rely on a 400px `box-shadow`](https://github.com/gatsbyjs/gatsby/issues/15676#issuecomment-511743350), since it won't work for transparent images more than 800px wide. Any thoughts on handling that?

- Could set the `box-shadow` to `options.maxWidth / 2`, but I'm not sure if that'd work for images that are taller than they are wide.
- Could use `background-color` instead of `box-shadow`, which seems to work for me, but there may be implications I'm unaware of.

KyleAMathews: Let's continue this discussion in a new issue!

PR #15956, opened 2019-07-21
chore(starters): add gatsby_minimal_portfolio

Mr404Found:

### DIFF
--- a/docs/starters.yml
+++ b/docs/starters.yml
@@ -3406,3 +3406,15 @@
     - Ready made Components
     - Responsive Design
     - Deployed with netlify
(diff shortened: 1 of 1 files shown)


lannonbr:
Hi @Mr404Found, So given this looks like you are using @LekoArts's Cara starter, I am going to actually recommend to moving this instead of a starter to instead be an entry in the Site Showcase highlighting your work.

You can learn how to submit an entry to the showcase here: https://www.gatsbyjs.org/contributing/site-showcase-submissions/

With that, I am going to close this PR. Thanks for using Gatsby! 🎉

PR #15963, opened 2019-07-21
feat(relay): use relay-compiler v5, fix #11536

### LINKED ISSUE #11536: gatsby: Don't use fork of relay-compiler
## Description

When having literal enums that have non-string values (eg ImageCropFocus), Relay fails to print the query. 

### Expected result

Should print normally.

### Actual result

Fails with exception.

Reported upstream in https://github.com/facebook/relay/issues/2633.

We should consider downgrading back to relay-compiler 1.

sibelius:
## Description

This uses relay-compiler v5 instead of gatsby fork

## Related Issues

#11536 

I've tried to run all tests locally, but there are a lot of them breaking (yarn jest)

### DIFF
--- a/packages/gatsby/package.json
+++ b/packages/gatsby/package.json
@@ -16,7 +16,7 @@
     "@babel/polyfill": "^7.0.0",
     "@babel/runtime": "^7.0.0",
     "@babel/traverse": "^7.0.0",
-    "@gatsbyjs/relay-compiler": "2.0.0-printer-fix.2",
+    "relay-compiler": "5.0.0",
     "@hapi/joi": "^15.0.0",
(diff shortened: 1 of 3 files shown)


freiksenet:
Sadly Relay have regressed on the bugs that we fixed in their code, Relay should be fixed before we can use it.

sibelius:
is there an issue there about it?

sidharthachatterjee:
@sibelius Not sure about an issue there at the moment but I think @freiksenet is referring to the fixes he made in https://github.com/facebook/relay/pull/2638. 

@freiksenet Want to add a test asserting the behaviour that we depend on to better track this? 

Closing this in the mean time but thank you for opening this @sibelius and apologies for the confusion here!

PR #15964, opened 2019-07-22
chore: speed up and lower memory usage of jobs

me4502:
## Description

In the case of Gatsby, immutable operations inside the reducers can be rather slow due to the potential size of the state data. Normally mutable operations here are frowned upon, but as Gatsby is using Redux in this way, it's perfectly fine to keep them mutable for performance reasons.

This change appears to have trimmed 60 seconds off a build time of 400 seconds. It is however a very image-heavy build.

### DIFF
--- a/packages/gatsby/src/redux/reducers/jobs.js
+++ b/packages/gatsby/src/redux/reducers/jobs.js
@@ -18,20 +18,14 @@ module.exports = (state = { active: [], done: [] }, action) => {
         })
 
         state.active[index] = mergedJob
-        return {
-          done: state.done,
-          active: state.active,
-        }
+        return state
       } else {
         state.active.push({
           ...action.payload,
           createdAt: Date.now(),
           plugin: action.plugin,
         })
-        return {
-          done: state.done,
-          active: state.active,
-        }
+        return state
       }
     }
     case `END_JOB`: {
@@ -53,10 +47,7 @@ module.exports = (state = { active: [], done: [] }, action) => {
         runTime: moment(completedAt).diff(moment(job.createdAt)),
       })
 
-      return {
-        done: state.done,
-        active: state.active,
-      }
+      return state
     }
   }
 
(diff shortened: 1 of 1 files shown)


packages/gatsby/src/redux/reducers/jobs.js
```diff
@@ -10,31 +10,27 @@ module.exports = (state = { active: [], done: [] }, action) => {
         throw new Error(`An ID must be provided when creating or setting job`)
       }
       const index = _.findIndex(state.active, j => j.id === action.payload.id)
-      const mergedJob = _.merge(state.active[index], {
```
me4502: Was there a reason merge was happening outside of the branch that was using it? _.merge appears to be immutable from the docs so this appears to be an entirely useless call on initial creation.

packages/gatsby/src/redux/reducers/jobs.js
```diff
@@ -10,31 +10,27 @@ module.exports = (state = { active: [], done: [] }, action) => {
...
-      })
       if (index !== -1) {
+        const mergedJob = _.merge(state.active[index], {
+          ...action.payload,
+          createdAt: Date.now(),
+          plugin: action.plugin,
+        })
+
+        state.active[index] = mergedJob
         return {
```
pieh: should we return just `state` if we are mutating it anyway?

packages/gatsby/src/redux/reducers/jobs.js
```diff
@@ -10,31 +10,27 @@ module.exports = (state = { active: [], done: [] }, action) => {
...
-          ],
+          active: state.active,
         }
       } else {
+        state.active.push({
+          ...action.payload,
+          createdAt: Date.now(),
+          plugin: action.plugin,
+        })
         return {
```
pieh: same here - it seems like we can just return `state`

packages/gatsby/src/redux/reducers/jobs.js
pieh: Probably no reason. Moving it inside conditional makes more sense

packages/gatsby/src/redux/reducers/jobs.js
me4502: Oh good point, will do

PR #15983, opened 2019-07-22
fix(www): copy button now works on iOS

DSchau:
## Description

This fixes an issue with the copy button that `contenteditable` and `range` must be used to work on iOS Safari. _Fun_.

### DIFF
--- a/www/src/components/code-block/__tests__/index.js
+++ b/www/src/components/code-block/__tests__/index.js
@@ -3,7 +3,14 @@ import CodeBlock from ".."
 import { fireEvent, render } from "react-testing-library"
 
 beforeEach(() => {
+  document.createRange = jest.fn()
   document.execCommand = jest.fn()
+  window.getSelection = jest.fn().mockImplementation(() => {
+    return {
+      addRange: jest.fn(),
+      removeAllRanges: jest.fn(),
+    }
+  })
 })
 
 describe(`basic functionality`, () => {
@@ -16,7 +23,7 @@ describe(`basic functionality`, () => {
       expect(queryByText(`copy`)).toBeDefined()
     })
 
-    it(`copies text to clipboard`, () => {
+    it.only(`copies text to clipboard`, () => {
       const text = `alert('hello world')`
       const { queryByText } = render(
         <CodeBlock language="jsx">{text}</CodeBlock>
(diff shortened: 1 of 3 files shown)


www/src/components/code-block/index.js
```diff
@@ -93,3 +97,15 @@ export default ({
     </Highlight>
   )
 }
+
+CodeBlock.propTypes = {
+  children: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired,
+  className: PropTypes.string,
+  disableCopy: PropTypes.bool,
```
DSchau:
```suggestion
  copy: PropTypes.bool,
```

www/src/components/copy.js
```diff
@@ -13,15 +13,21 @@ import {
 } from "../utils/presets"
 
 const copyToClipboard = content => {
```
wardpeet: NOT a blocking comment!

no big deal for now but have you seen the async clipboard api?
https://googlechrome.github.io/samples/async-clipboard/
https://alligator.io/js/async-clipboard-api/

It has pretty good support and has less bugs than the old api:
https://www.chromestatus.com/feature/5861289330999296

We can feature detect this and if browsers don't support it we don't show it. I don't think we should support old clipboard api.

www/src/components/code-block/__tests__/index.js
```diff
@@ -16,7 +23,7 @@ describe(`basic functionality`, () => {
       expect(queryByText(`copy`)).toBeDefined()
     })
 
-    it(`copies text to clipboard`, () => {
+    it.only(`copies text to clipboard`, () => {
```
wardpeet: Should we revert only?
```suggestion
    it(`copies text to clipboard`, () => {
```

DSchau: Yasss -- awesome catch, thank you!

www/src/components/copy.js
DSchau: Nice. I was thinking of extracting this into a utility, and this seems like the kind of "branching logic" that can be tested and extracted. I'll do this in a bit -- love this idea!

DSchau:
@wardpeet merging this, then gonna add the async clipboard stuff next!

www/src/components/copy.js
DSchau: Done in #16012

PR #15984, opened 2019-07-22
feat(docs): Include explicit mention of GraphQL arguments coming from context

taylorham:
## Description

I got confused as to why certain fields weren't being sent into my GraphQL queries and spent an embarrassing amount of time trying to debug it in all the wrong ways.

This PR adds a sentence the "Programmatically Create Pages From Data" section of the docs explicitly clarifying that all values passed to `context` are made available to GraphQL queries as arguments.

:bulb: It might be worth also explaining the related inverse, where any argument you expect to use in GraphQL queries must be specified on the `context`. **Opinions appreciated!**

## Related Issues

My own lack of understanding 😁

I also want to thank whoever sees this for doing incredibly impressive work on Gatsby thus far. I've had a blast working on a few `theme` projects for my company and I hope I can contribute more in the coming weeks!

Also big thanks to @lannonbr for pointing out the fix in the [`#need-help` Discord channel](https://discordapp.com/channels/484383807575687178/537691356487876624)!

### DIFF
--- a/docs/docs/programmatically-create-pages-from-data.md
+++ b/docs/docs/programmatically-create-pages-from-data.md
@@ -51,11 +51,21 @@ exports.createPages = async function({ actions, graphql }) {
 
 For each page we want to create we must specify the `path` for visiting that
 page, the `component` template used to render that page, and any `context`
-we need in the component for rendering. The `context` parameter is
-_optional_ though often times it will include a unique identifier that can
-be used to query for associated data that will be rendered to the page. All
-`context` values are made available to a template's GraphQL queries as
-arguments prefaced with `$` (ex. `slug` -> `$slug`).
+we need in the component for rendering.
+
+The `context` parameter is _optional_, though often times it will include a
+unique identifier that can be used to query for associated data that will be
+rendered to the page. All `context` values are made available to a template's
+GraphQL queries as arguments prefaced with `$`, so from our example above the
+`slug` property will become the `$slug` argument in our page query:
+
+```js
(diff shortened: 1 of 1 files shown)


docs/docs/programmatically-create-pages-from-data.md
```diff
@@ -53,7 +53,9 @@ For each page we want to create we must specify the `path` for visiting that
 page, the `component` template used to render that page, and any `context`
```
wardpeet: What do you think of the following sentence?

For each page we want to create, we must specify the path for visiting that page, the component template used to render that page, and any context we need to render the component. The context parameter is optional though often times it will include a unique identifier that can be used to query for associated data that will be used inside your page. All context values are made available in the page query as arguments prefaced with $ (ex. slug -> $slug).
```js
export const query = graphql`
  query($slug: String!) {	  query($slug: String!) {
  }
`
```

I'm not sure if we need to show an example as it's a few lines below anyway.

cc @gillkyle

gillkyle: See my comments [here](https://github.com/gatsbyjs/gatsby/pull/15984#pullrequestreview-265458577), I think this would be great 🙂

taylorham:
Thank you all for the suggestions! I'm very happy to help and should have time before the weekend to include the additions and finalize the PR.

taylorham:
I changed a few things around for clarity and emphasis, let me know if I went overboard 🚢

PR #15988, opened 2019-07-22
feat(gatsby-theme-blog): use dirname as fallback as slug

pomber:
Some people (me included) use a folder for each blog post with an `index.mdx`.

Currently, if I have `content/posts/my-first-post/index.mdx` it creates the page at `/my-first-post/index`. I changed so it creates the page at `/my-first-post`.

For files other than `index` it uses the full path (same as before): `content/posts/my-first-post/foo.mdx` -> `/my-first-post/foo`

### DIFF
--- a/themes/gatsby-theme-blog/gatsby-node.js
+++ b/themes/gatsby-theme-blog/gatsby-node.js
@@ -169,7 +169,10 @@ exports.onCreateNode = ({ node, actions, getNode, createNodeId }) => {
 
   const toPostPath = node => {
     const { dir } = path.parse(node.relativePath)
-    return urlResolve(basePath, dir, node.name)
+    const isIndex = node.name === `index`
+    return isIndex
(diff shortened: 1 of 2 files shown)


KyleAMathews:
There's a helper function for this in gatsby-source-filesystem that handles this (and more) — could you try switching to it? https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-source-filesystem/src/create-file-path.js

e.g. from the tutorial:

```js
const { createFilePath } = require(`gatsby-source-filesystem`)

exports.onCreateNode = ({ node, getNode }) => {
  if (node.internal.type === `MarkdownRemark`) {
    console.log(createFilePath({ node, getNode, basePath: `pages` }))
  }
}
```

https://www.gatsbyjs.org/tutorial/part-seven/#creating-slugs-for-pages

PR #15990, opened 2019-07-22
fix(PrismJS): Documentation highlighting

LorrisSaintGenez:
## Description

Browsing through your doc, I've noticed some highlighting was broken.
Using `display: inline` on the `.token` class instead of `inline-block` fixed it.

**Before:**
Screenshot 2019-07-22 at 14 03 38

**After:**
Screenshot 2019-07-22 at 14 03 48

(Page of the screenshot : https://www.gatsbyjs.org/docs/node-apis/?origin_team=T026AN50K#createPages)

### DIFF
--- a/www/src/utils/typography.js
+++ b/www/src/utils/typography.js
@@ -315,7 +315,7 @@ const _options = {
       // PrismJS syntax highlighting token styles
       // https://www.gatsbyjs.org/packages/gatsby-remark-prismjs/
       ".token": {
-        display: `inline-block`,
+        display: `inline`,
       },
       ".token.comment, .token.block-comment, .token.prolog, .token.doctype, .token.cdata": {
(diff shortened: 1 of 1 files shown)


DSchau:
Oh interesting! Thanks for this.

This is a good fix -- but it also will "break" some other snippets. Specifically we need inline-block on empty lines, e.g. with this change you'll see the following as an example:

Screen Shot 2019-07-22 at 2 45 22 PM

So we'll need a different/better mechanism that also keeps the above working. I'll keep thinking on it!

DSchau:
Let's get this merged -- it seems to be generally better, with only a few cases (blank lines!) where it's merged.

I'll open an issue to track on the blank line issue!

LorrisSaintGenez:
Thank you @DSchau 🎉

PR #16004, opened 2019-07-23
feat(gatsby-image): Add lazyOffset prop

proAlexandr:
## Description

`lazyOffset` prop was added to `gatsby-image` component which passes to IntersectionObserver as a `rootMargin` property. The default value remains `200px`. It makes possible to control how early or lately an image will be loaded.

## Related Issues

Implementation of #14662

### DIFF
--- a/docs/docs/gatsby-image.md
+++ b/docs/docs/gatsby-image.md
@@ -370,6 +370,8 @@ After you've made a query, you can pass additional options to the gatsby-image c
 | `objectPosition`       | `string`            | Passed to the `object-fit-images` polyfill when importing from `gatsby-image/withIEPolyfill`. Defaults to `50% 50%`.          |
 | `loading`              | `string`            | Set the browser's native lazy loading attribute. One of `lazy`, `eager` or `auto`. Defaults to `lazy`.                        |
 | `critical`             | `bool`              | Opt-out of lazy-loading behavior. Defaults to `false`. Deprecated, use `loading` instead.                                     |
+| `draggable`            | `string`            | Passed to the rendered HTML `img` element                                                                                     |
+| `lazyOffset`           | `string`            | An offset for lazy-loading. The same as `IntersectionObserver.rootMargin`, Defaults to `200px`.                               |
 
 Here are some usage examples:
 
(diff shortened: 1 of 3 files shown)


wardpeet:
Thank you for opening this Pull Request.

Sadly we're not going to merge this one. Lazy loading is coming to the web and passing down `lazyOffset` won't have any effect anymore for browsers that support this technique. The browser will figure out the best behaviour on when to load images.

More information can be found https://addyosmani.com/blog/lazy-loading/ or https://chromestatus.com/feature/5645767347798016

@proAlexandr would love to see people experimenting with different values other than 200px — if we decide a different value works better, we can definitely change it to something else.

lnikell:
Hi @wardpeet, appreciate for your feedback. Am I right that you would like in future completely move the control of lazy loading to the browser? Currently we got 3 requests in a row from our client who just don't like that images load only when you scroll to it, and I personally don't like it as well since 200px is very small and the most website are section based that described https://github.com/gatsbyjs/gatsby/issues/14662 
In that case it seems for me that there is a good place for another plugin that will give more control over lazy loading.

wardpeet:
> Am I right that you would like in future completely move the control of lazy loading to the browser
Yes you're right.

We're happy to accept any other value than 200px if there is a better value. I've used 50% of the viewport before. 

Feel free to fork gatsby-image, modify it to your liking and publish your own version of it. https://www.gatsbyjs.org/contributing/submit-to-plugin-library/

polarathene:
Technically, if you combine the IntersectionObserver trigger to make note of if native lazyloading is in use, you could then also check if the image is loaded, if not, switch from lazy to eager to ensure your image begins loading within that trigger range always.

So even with native browser support on the way(which is still only limited to chrome or blink browsers atm?), this would be a way for the required use-case to work. If the core team does not want to merge that for maintenance reasons, fair enough. I'd suggest waiting until a future point when using hooks is permitted, then gatsby-image can be better refactored for composing custom components without having to maintain your own fork.

fabio-arsenal:
Damnit. I need this.

0xsven:
Damnit. I need this.

PR #16005, opened 2019-07-23
Add decoding prop to gatsby-image img element

xxczaki:
## Description

This PR adds the decoding prop to img element inside gatsby-image package.

* [Information about the decoding attribute on MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/decoding)

## Notice

Let me know if I should add a default decoding prop value, for example `async`.


wardpeet:
This looks great! I'm not sure if we should enable async on every image. Not sure what the performance implications are. Do you know if there is any lecture on this?

xxczaki:
@wardpeet Hi, here are some useful resources about the `decoding` attribute:

* [img decoding attribute explainer](https://docs.google.com/document/d/1VvrEGAjcnto5egR_rnX7PR5ymqWayMNRPYSbCf1bC4I/edit#heading=h.83s717wno266)
* [Spec PR](https://github.com/whatwg/html/pull/3221)
* [Addy Osmani's tweet & discussion](https://twitter.com/addyosmani/status/938078402430382080)

Hope this helps ;)

wardpeet:
You're the best!

KyleAMathews:
Hmm interesting — I've never seen this attribute before — is there any reason to not set the attribute to `async` always? That way we could avoid adding another prop.

xxczaki:
@KyleAMathews I think we should use the `decoding="async"` value as default, because it improves loading performance & we are using `loading="lazy"` as default too. 

However, using `decoding="async" sometimes might lead to a visual flash and I guess not everyone will like that. So IMO it would be better if there was an option for developers using Gatsby to change the default `decoding` attribute when needed.

Still, I'm not a web performance expert and maybe we should ask more experienced people (such as Addy Osmani) for their opinion :smile:

xxczaki:
cc @addyosmani

addyosmani:
My recommendation here would be for someone to put together a test page or site that uses loading=lazy and decoding=async with Gatsby's particular implementation. We can then test that on a slow network and device and see what the typical user experience might be.

Does that sound reasonable? I can give you advice but I think in this case as we're trying to make a default for a wide range of use cases measurement might be our best option :)

wardpeet:
Hey @xxczaki,

I've created an issue to create a demo gatsby site with the decoding prop. https://github.com/gatsbyjs/gatsby/issues/17941

As stated above the change you’ve proposed is not going to be accepted until the issue above has been fixed.

PR #16012, opened 2019-07-23
refactor(www): use new clipboard.writeText API if it exists

DSchau:
## Description

Shout-out @wardpeet for letting me know about the feature _and_ for not blocking the other PR to introduce it. The real MVP 🙌

## Related Issues

Follow-up to #15983

### DIFF
--- a/www/src/utils/copy-to-clipboard.js
+++ b/www/src/utils/copy-to-clipboard.js
@@ -1,22 +1,30 @@
-// https://hackernoon.com/copying-text-to-clipboard-with-javascript-df4d4988697f
 const copyToClipboard = str => {
-  const el = document.createElement(`textarea`) // Create a <textarea> element
-  el.value = str // Set its value to the string that you want copied
-  el.setAttribute(`readonly`, ``) // Make it readonly to be tamper-proof
-  el.style.position = `absolute`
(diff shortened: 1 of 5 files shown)


www/src/utils/copy-to-clipboard.js
```diff
@@ -1,22 +1,30 @@
...
-  if (selected) {
-    // If a selection existed before copying
-    document.getSelection().removeAllRanges() // Unselect everything on the HTML document
-    document.getSelection().addRange(selected) // Restore the original selection
+  const clipboard = window.navigator.clipboard
+  /*
+   * fallback to older browsers (including Safari)
+   * if clipboard API not supported
+   */
+  if (!clipboard || typeof clipboard.writeText !== `function`) {
```
wardpeet: oh crap clipboard isn't supported on Safari, chrome statuswindow.navigator.clipboard is lying to me :scream: 
https://www.chromestatus.com/feature/5861289330999296

DSchau: This feature detection should still work just fine tho right?

wardpeet: yeah! 👌

PR #16017, opened 2019-07-23
chore(gatsby-theme-blog): Capitalization in social link names

fk:
Following up on #15738, changing image to say `Twitter` and `GitHub`.

I don't know how to test changes to themes yet, but I believe this shouldn't break anything?

### DIFF
--- a/themes/gatsby-theme-blog/README.md
+++ b/themes/gatsby-theme-blog/README.md
@@ -70,11 +70,11 @@ module.exports = {
     // Used for social links in the root footer
     social: [
       {
-        name: `twitter`,
+        name: `Twitter`,
         url: `https://twitter.com/gatsbyjs`,
       },
       {
-        name: `github`,
+        name: `GitHub`,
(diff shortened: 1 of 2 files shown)


wardpeet:
Technically this is a breaking change but we are still under the 0 major so we could ship this as a 0.1.0 or just as a patch. I don't think it matters much so let's 🚢 it

KyleAMathews:
It's not a breaking change unless we change an interface e.g. what data is passed to a component, etc. So a file someone has shadowed expects certain data or normal stuff like options passed into a theme.

We can make cosmetic changes like this all we want. That's the benefit of having themes is that people _can_ improve themes and everyone gets the improvements.

fk:
@eclectic-coding 👋 Hey Chuck! Nice to meet you!
`text-decoration: capitalize` wouldn't have helped with "Git*H*ub", so unfortunately this

> The capitalization depends on the user setting the capitalization correctly in the SiteMetadata properly.

is the only way.

PR #16025, opened 2019-07-23
chore(gatsby-plugin-sharp): pass md5 hash of input file to Cloud

DSchau:
## Description

This simple PR adds a content hash mechanism to send to Gatsby Cloud so that subsequent Cloud image requests can be cached based on the content and transforms arguments, rather than the file argument (which could change to a cache directory on subsequent builds).

Why?

Cloud can't compute a unique hash based on input file file alone (which changes between runs), we need a content hash and transforms to compute a cache key, which we can then cache as long as those are unique/non-changing.

## Questions

Both of the below were fixed by threading the digest down to the scheduler 🎉 

1. ~Could be fairly expensive to compute md5-file for each file; we have this in the file node; is there a better way we can use to avoid computing it each time?~
1. ~We should compute md5-file based on _original input file_ instead of the cached file (which could theoretically change?)~
	- In other words: first invocation `src/images/gatsby-icon.png`
	- Second invocation: `.cache/caches/gatsby-source-filesystem/gatsby-icon-1234.png`

cc @rase- @wardpeet (since you worked on this previously!)

### DIFF
--- a/packages/gatsby-plugin-sharp/package.json
+++ b/packages/gatsby-plugin-sharp/package.json
@@ -18,7 +18,6 @@
     "imagemin-pngquant": "^6.0.0",
     "imagemin-webp": "^5.0.0",
     "lodash": "^4.17.14",
-    "md5-file": "^4.0.0",
     "mini-svg-data-uri": "^1.0.0",
     "potrace": "^2.1.1",
     "probe-image-size": "^4.0.0",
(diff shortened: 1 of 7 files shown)


DSchau:
Meant to mark this a draft -- but that's OK! Adding WIP tags to make my intent clearer here :)

KyleAMathews:
Why can't we use the contentDigest from the `File` node?

DSchau:
@KyleAMathews I didn't go far enough up the chain 🙃 Threading it like that should work. One sec.

DSchau:
Let's keep this WIP/non-merged until we test it in Cloud, please!

DSchau:
@wardpeet I have! Waiting for Cloud team's approval that this works, then we can publish!

sidharthachatterjee:
This should be okay to ship for now! Thank you so much again @DSchau 🥇

Issue #15281, opened 2019-07-01
[feat]: expose API to validate pluginOptions


DSchau:
## Summary

Passing options to plugins is a key construct in Gatsby, and the plugin ecosystem is key to the power and reusability of any Gatsby application.

However — currently we’re encouraging plugins to implement their own plugin validation API, or worse yet, to not validate options whatsoever. This leads to failing much later than needed, and we should fail early with a nice, helpful error message when an invalid plugin option shape is detected.

### Basic example

Let’s consider Contentful, as an example, specifically gatsby-source-contentful. It accepts `accessToken`, `spaceId`, `host`, etc. to connect to the Contentful instance. These environment variables are oftentimes environment variables (secrets!), which makes this even harder to debug. As a user using this functionality, it may look something like:

```js
require('dotenv').config({
  path: `.env.${process.env.NODE_ENV}`
})

module.exports = {
  plugins: [
    {
      resolve: `gatsby-source-contentful`,
      options: {
        spaceId: process.env.CONTENTFUL_SPACE_ID,
        accessToken: process.env.CONTENTFUL_ACCESS_TOKEN,
        host: process.env.CONTENTFUL_HOST
      }
    }
  ]
}
```

However — if any of those environment variables are undefined, the source plugin will not fail early with a helpful error message, but will rather fail with something like:

> [Gatsby] error Accessing your Contentful space failed.

### Code and Approaches

I see (minimally) two ways we could do this.

1. Overload a lifecycle method (e.g. onPreInit) and pass in a helper/helpers to validate plugin optionsn
1. Create a new lifecycle method (e.g. `validatePluginOptions`)
    - Note: this should be async (Joi.validate is async) and in case e.g. you could do deep validation, e.g. check the accessToken and spaceId are not only provided but valid

I think 2. is preferable, because it will just be a “no-op” in older versions of Gatsby, whereas 1. would potentially break if the helper wasn’t passed as expected.

### Code example

Again, I think there are two approaches here:

1. We use Joi (or equivalent?) and pass it as an option to `validatePluginOptions`
1. We use GraphQL to validate the shape of a plugin’s options, e.g. with an SDL-like approach like the below

```js
exports.validatePluginOptions = ({ validateOptions }) => {
  return validateOptions(`
    # note: not the final API, nor even the best API
    # strawman to begin the discussion
    type SourceContentfulPluginOptions implements PluginOptions {
      accessToken: String!
[...]

PR #16027, opened 2019-07-23
feat: start work on validating plugin options

DSchau:
## Description

_Extremely_ WIP work on implementing #15281 

## Issues to Solve

1. Implementing this API in a plugin is a breaking change (because Gatsby will fail with an unknown API error)
	- Note: we should probably solve this problem, e.g. to implement new APIs isn't _always_ a breaking change, e.g. a noop in older versions of Gatsby would've been OK here
1. I'd _probably_ like to use this API to coerce values, as well (e.g. if expecting a number, Joi will coerce to a string a la [this example](https://runkit.com/dschau/5d378554b0f6a20023ca6f32)) and set the result as `pluginOptions` for the plugin from that point
1. This should only be available to plugins (e.g. disallow in root-level gatsby-node.js?)
1. Need to set defaults on passed-in Joi instance
	- `allowUnknown: true` to avoid empty `plugins` option causing issues
    - `abortEarly: false` to show _all_ errors

and probably more I'm forgetting. It's a start! 🚀 

Screen Shot 2019-07-23 at 3 16 31 PM

## Still to do

- [ ] Solve all those issues
- [ ] Ensure _not_ using the passed in `validator` doesn't cause any issues with the error message
- [ ] Add tests
- [ ] Dogfood with plugins in gatsby/* monorepo (e.g. gatsby-source-contentful)
- [ ] Make structured error message a little nicer and actionable


pieh:
> 3\. This should only be available to plugins (e.g. disallow in root-level gatsby-node.js?)

Is this distinction that need to be made?

Only usecase I see where this could be problematic (depending how it's handled) is if there is theme that also doubles as starter (where you would want validation for theme options, but maybe not when using it as a starter (tho then, using default values might actually be good idea)

DSchau:
> Is this distinction that need to be made?

It feels like a "useless" API in the root of a project, but yeah -- perhaps not a fix that needs to be made. In other words, if a user wants to use it (even though it's not valuable) -- so be it!

DSchau:
Because this is quite breaking for any plugins that implement, let's either:

1) Implement in plugins outside the scope of this PR, or
2) Wait until #16027 lands

I think I'm leaning towards 1 in the interest of shipping the base functionality, however, I do also think it's quite important to _use_ the API in various plugins so that we can ensure it accommodates all use cases.

DSchau:
This is no longer blocked since #16027 landed. I'll revisit work on this today and this week.

There will be some churn in _any_ new APIs until people upgrade to the latest Gatsby, but I think that's _OK_.

pieh:
> This is no longer blocked since #16027 landed.

I think that's wrong PR/Issue number (it's links to this PR) ;)

DSchau:
lol -- I double down'd on my assumption.

*** #16105

DSchau:
OK -- got a lot done on this.

- [x] Values are coerced by Joi/validation
    - Note: the idea is that validatePluginOptions returns either a Joi validation instance (e.g. Joi.object().keys()) or the user can directly return plugin options which get mutated.
- [ ] Joi doesn't warn on e.g. plugins or other misc. fields; _all_ errors are shown instead of just first
- [x] JSON types for SitePlugin fields
    - Note: didn't do separate plugin options (e.g. SitePluginGatsbySourceContentful or anything) but I think the schema customization API solves that purpose rather nicely
- [x] Dogfooded on a plugin (`gatsby-source-contentful`)

What's not done?

- [ ] Themes don't get their options coerced because their gatsby-config.js is a special option and we run this API after themes are loaded
    - This will be a little challenging to solve but should be doable
    - This is a "big deal" because we can set defaults in this API and coerce values, which we want to make available to gatsby-config.js and other plugins the theme may invoke
- [ ] Tests
- [ ] Dogfood more packages

packages/gatsby/src/utils/api-runner-node.js
```diff
@@ -220,14 +221,36 @@ const runAPI = (plugin, api, args) => {
...
-        apiFinished = true
-        return res
-      })
+      chain = Promise.resolve(result)
+        .then(res => {
+          /*
+           * validatePluginOptions is special
+           * it validates plugin options and (optionally) mutates them
+           */
+          if (api === `validatePluginOptions`) {
```
DSchau: This feels _extra janky_. Any suggestions for improvements?

packages/gatsby/src/utils/api-runner-node.js
```diff
@@ -220,14 +221,36 @@ const runAPI = (plugin, api, args) => {
...
+            if (res) {
+              let resChain = Promise.resolve(res)
+              if (res.validate) {
+                resChain = res.validate(plugin.pluginOptions, {
+                  abortEarly: false,
+                  allowUnknown: true,
+                })
+              }
+              return resChain.then(options => {
```
DSchau: Here's where options are actually coerced. A "gotcha" here is that whether using Joi or not, the returned structure is what will be passed to other APIs that consume plugin options.

moonmeister:
@DSchau because of the complex plugin options for `gatsby-plugin-manifest` it could be a good dog fooder. I should have some time in the next week to test this, anything i should know?

DSchau:
@moonmeister I was just gonna pick this up again today!

Yeah -- that'd be a great one. It should be ready to test for plugins. Would you want to open a PR to this branch when you test things out?

moonmeister:
@dschau that works for me.

packages/gatsby/src/bootstrap/load-themes/index.js
```diff
@@ -5,6 +5,34 @@ const _ = require(`lodash`)
...
 const preferDefault = require(`../prefer-default`)
 const getConfigFile = require(`../get-config-file`)
+const {
+  validatePluginOptions,
+  formatOptionsError,
+} = require(`../../utils/validate-plugin-options`)
+const reporter = require(`gatsby-cli/lib/reporter`)
+const Joi = require(`@hapi/joi`)
+
+const getPluginOptions = theme => {
```
DSchau: I kinda hate this -- the idea is that we need to replicate the coercion that the validatePluginOptions API will normally do.

However -- themes are a special kind of plugin (that can create other plugins) so we have to do some hackery here to ensure that options are coerced (and thrown) and passed to the theme as expected.

Suggestions welcome!

packages/gatsby/src/bootstrap/load-themes/index.js
```diff
@@ -5,6 +5,34 @@ const _ = require(`lodash`)
...
+  const options = theme.options || {}
+  try {
+    const { validatePluginOptions: validateAPI } = require(path.join(
+      theme.themeDir,
+      `gatsby-node.js`
+    ))
+    // note: this is fairly naive; it doesn't pass the additional API options
+    // that other APIs have available (and e.g. that a plugin implementing this API) would have
```
DSchau: This is sort of the big reason I kinda hate it. I don't think there are many other valid/useful API options -- but I don't like the "gotcha" that this kind of API is special for themes and is treated differently than e.g. a plugin.

packages/gatsby/src/utils/api-node-docs.js
```diff
@@ -411,3 +411,21 @@ exports.onPreExtractQueries = true
  * }
  */
 exports.onCreateDevServer = true
+
+/**
+ * Run before sourcing plugins from gatsby-config.js. Useful for
+ * failing quickly with a helpful error message for in invalid options shape.
+ * Note: we provide an instance of "Joi" for validation, but can also do any kind of async validation necessary here.
+ * @gatsbyVersion 2.14.0
```
DSchau: TODO: change this when it's ready to be merged.

packages/gatsby/src/utils/validate-plugin-options.js
```diff
@@ -0,0 +1,23 @@
+exports.validatePluginOptions = (res, options = {}) => {
+  let chain = Promise.resolve(res)
+  if (res && res.validate) {
+    chain = res.validate(options, {
```
DSchau: There's a trade off here -- I think this is a best practice (e.g. we'll want to capture _all_ errors and allow unknown) but some plugins would possibly want more control over this.

moonmeister:
@DSchau Hey, so I just added a huge commit for validating the entire WebAppManifest spec. I ran into one thing that we should fix. Having to add `plugins: Joi.array()` to every config's preferences is redundant and tedious (imo). I figured out how to merge Joi objects since our plugin options are intertwined with the manifest definition itself. Using the code below we add this as default:

```js
//not sure if this is how you call validate, but hopefully you get the idea

 let newObject = Joi.object()
    .keys({
      plugins: Joi.array()
    })
    .concat(joiObjectFromAPI)
);

newObject.validate( pluginOptions)
```

because we're creating an object and appending the provided object from the API, it means that `plugins` settings can be extended and overridden from a plugin.

so if I set my plugins object to 
```js 
plugins: Joi.array().require().items(Joi.string())
``` 
then the plugins option would be required. See my PR for multiple example of this.

DSchau:
Hey @moonmeister!

Awesome! Let me add a few comments/questions:

> Having to add plugins: Joi.array() to every config's preferences is redundant and tedious (imo)

Why is this necessary? Removing "strict" (e.g. allowing unknown keys) should prevent this from being necessary?

Agreed the concat approach would work, but ideally we don't need that at all.

moonmeister:
@DSchau Hmm, I suppose I'd lean towards the strict mode by default. It eliminates the typo case. I meant to se property `date` and accidentally typed `data` (might have spent too much time debugging this very thing at work this week). Without strict mode I don't get any error that there is an unknown key and I have to notice it myself. If we could make this throw a warning that would be fine with me, but I think it's worth throwing something!

DSchau:
@moonmeister I think you should check out the code I linked to!

Strict (as I used it) basically means it doesn't fail on extra keys passed that aren't part of the validation schema (in this case, plugins!).

moonmeister:
@DSchau Opps, you must have added code later, cause I didn't see it originally. Either way What you describe is what I'm suggesting, I want it to throw when I pass 'data' cause it's not part of the validation schema. I only ran into this issue when i added `.unknown(false)` to my object.

DSchau:
You could directly use Joi to workaround the baked-in defaults. I _think_ what you're describing is a case that I'm not too concerned about.

If we think about your exact example, if you typo `data` as `date`, the validation _will_ fail because you haven't passed `data`. Sure -- it won't fail if you merely only pass `data` (because it will be an unknown key) but I'm not sure I'm convinced of the negative impact, yet!

Happy to be proven wrong!

moonmeister:
@DSchau Okay I get what your saying, there is still the case if `data` is optional then it won't error. I guess I don't see why making this strict and including `plugins` by default is such a big deal. Why is it "ideal" that we don't do this?

DSchau:
@moonmeister what if a plugin wants to enforce a particular shape for plugins? We then have a non-overridable key, right?

I guess I'd toss it back to ya -- why is `allowUnknown` such a big deal? I'm not yet convinced of the negative impact!

Also to be clear, if this is a concern, you can use your own instance of Joi (or really any validation). Ours is passed as a convenience, it's not strictly necessary.

DSchau:
@moonmeister in thinking more about my above statement -- I think `plugins` is a fairly typical known shape. It's either a string, or an object with keys `resolve` and `options`. I don't think this pattern falls down anywhere, so maybe it's OK to enforce it by default?

moonmeister:
> what if a plugin wants to enforce a particular shape for plugins? We then have a non-overidable key, right?

`plugins` would always be an optional array because gatsby always injects it (in the validation obj and in the plugin options). If a plugin wants to make it a required string they could. whatever is concatenated onto a Joi object merges with existing object properties but takes president. Meaning if `required()` is set on the `plugins` property in the `object` returned from the plugin then plugins becomes required.  UPDATE: but as you've now noted if I were to override this to a random object that'd be breaking the Gatsby API. I think this should be in there to make sure it's being validated as Gatsby expects it...if folks want to set it as required for their plugin that's okay. 

> I guess I'd toss it back to ya -- why is `allowUnknown` such a big deal? I'm not yet convinced of the negative impact!

It's not about the negative, it's about the positive experience it provides. I think I've articulated a very valid and probably common case where a dev (especially new ones) miss type or are trying to add properties not understanding where they should go. Explicitly setting `allowUnknown` to `false` would improve developer experience with no drawbacks that i can see.  Right now I've forced my object to not allow but if another plugin really needed, they could force to `allowUnknown`.  So why not?

DSchau:
@moonmeister you've persuaded me! I'll keep thinking if there are any plugins that may not match this expected structure (I suppose we could just make it more generic, e.g. validate that plugins is an array!), but for now, I agree that this is an improvement.

It does mean that developers will have to be _strict_ about valid options, but perhaps that's a valuable outcome?

See f0ef733 for the implementation and 870def4 for the tests

moonmeister:
I think it's valuable.

moonmeister:
@Dschau Where we at on this? Anything else I can do to help?

LekoArts:
This was successfully implemented in a bunch of others PRs :)
https://github.com/gatsbyjs/gatsby/pull/27337

PR #16034, opened 2019-07-24
Changing feature to future

scotttharvey:
## Description

Correcting sentence. The word feature should be future when refering to development


KyleAMathews:
Thanks for the PR but I'm pretty sure "feature development" was the intended phrase.

PR #16037, opened 2019-07-24
Refactor: refactor adding tags

marcobiedermann:
## Description

Refactor adding tags to blog post documentation

***

## Steps of implementation

* Loop over posts and get tags using `map()`
* Flatten nested arrays using `flat()`
* Combine both statements using `flatMap()`
* Remove duplicates by adding them to a `Set`

### DIFF
--- a/docs/docs/adding-tags-and-categories-to-blog-posts.md
+++ b/docs/docs/adding-tags-and-categories-to-blog-posts.md
@@ -208,16 +208,8 @@ exports.createPages = ({ actions, graphql }) => {
       })
     })
 
-    // Tag pages:
-    let tags = []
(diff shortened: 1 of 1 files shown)


pieh:
Both `.flat()` and `.flatmap()` was added to node in version 11. We do support node 8, so we can't use them (yet)

marcobiedermann:
@pieh

Okay, fair enough. Should I add this as a alternative in the comments or would you get rid of it entirely until you upgraded to Node V11?

sidharthachatterjee:
While this is helpful (and I personally like it too) I think we should hold on to this till our minimum supported Node version is 11. 

Closing this in the mean time but thank you so much for this @marcobiedermann 💜

PR #16043, opened 2019-07-24
fix(gatsby-plugin-mdx): Add missing HTML attribute translations

PedroLamas:
## Description

This introduces some missing conversions from regular HTML attributes to JSX ones.

I've been getting some in-browser console warnings from converted MDX files where I embedded a twit from Twitter, and it contains a "charset" attribute that should be "charSet" in JSX.

I went through the Typescript definitions file for ReactJs and got most of the cases that could cause this issue and added to the translations object.


PedroLamas:
FYI, the unit tests have passed (check the details link) but GitHub status has not updated accordingly...

johno:
Yeah I _think_ for now we can mirror the larger object over and reference where it originated as @sidharthachatterjee mentioned.

We have plans to pull this plugin out of `gatsby-plugin-mdx` and into MDX core so it'll eventually be something that Gatsby core doesn't have to maintain (including the dictionary ofc), but it's sort of low on the list right now.

PedroLamas:
So are you ok for me to just pull the whole object here with a link to where it came from?

johno:
That works for me @PedroLamas!

PedroLamas:
@johno done! Full object from React is now here!

pieh:
Published `gatsby-plugin-mdx@1.0.17`

Issue #16055, opened 2019-07-24
[feat] Improve experience of introducing new APIs to Gatsby Core


DSchau:
## Summary

When we have a need to introduce a new API, e.g. see #16027, we also unfortunately tend to break support/introduce some churn for older versions of Gatsby in the process.

For example, given the above API `validatePluginOptions` -- if we added that API to a plugin, and that plugin was updated but the gatsby version was kept constant (very common!) the following error would appear:

> The plugin "gatsby-source-contentful@2.1.5" is exporting a variable named "validatePluginOptions" which isn't an API.

See #15680 for a real world example of the issue.

### Basic example

This is a _straw man_ (aka a proposal that potentially has flaws, and that I'd love to iterate to something better) proposal.

1. We create (at build/publish time) a list of known APIs in api-node-docs.js as JSON (e.g. gatsby/node-apis.json, gatsby/browser-apis.json, etc. or perhaps one big api.json file with keys for node, browser, etc.)
    - We'll probably want to use Babel or something to extract all the exports
    - Ideally, would also include the version introduced and then some metadata as to whether it should be "breaking" to older versions or not (e.g. some are crucial for functionality, others can/should be a noop as they're an enhancement, not crucial)
1. We build into bootstrap a request to these files (with Unpkg or equivalent), to know about the _latest_ APIs
1. If a breaking API, break with a message like the [below](#breaking-message)
1. If not breaking, _warn_ with a similar message

The structure could look something like this:

```json
{
  "createPages": {
    "version": "0.0.0",
    "breaking": true
  },
  "validatePluginOptions": {
    "version": "2.2.0",
    "breaking": false
  }
}
```

### Breaking Message

> The plugin {PLUGIN_NAME}@{PLUGIN_VERSION} is trying to use the API: `{API_NAME}` which is not available in gatsby@{GATSBY_VERSION}
>
> To fix, upgrade to gatsby@{GATSBY_API_VERSION} with:
>
> npm install gatsby@^{GATSBY_API_VERSION}

_Note: if the network call fails, we need to fail quickly, and then tweak the message to be less useful and probably presumed breaking._

### Motivation

We need to introduce new APIs. We can't break people's builds when they (very likely!) upgrade plugin versions, while keeping Gatsby versions constant.

We should _encourage_ to upgrade, but not break builds whenever possible. We want to continue to iterate and build in useful functionality, but we also don't want to cultivate an environment of churn and frustration.

PR #16061, opened 2019-07-24
chore(gatsby): output apis.json for current valid API list

DSchau:
## Description

This PR begins implementing one chunk of the work required to improve the experience outlined in #16055.

Specifically, this PR outputs an `apis.json` file that we can then distribute with NPM (and query with e.g. unpkg).

This uses a doclet `@gatsbyVersion` to denote when an API is available. This will be used to deliver richer error messages (or warnings) like so:

```
The plugin gatsby-source-filesystem is using the Gatsby API: `createSchemaCustomization`.

This API was introduced in gatsby@2.12.0. Upgrade your version of Gatsby (currently 2.1.0) for the best possible experience:

npm install gatsby@^2.12.0
```

A harder problem to solve is if we tweak an existing API (e.g. to pass a new utility or something) but I'd contend that we can either: 1) not solve that problem, or 2) Solve it later with additional metadata. As an example of this problem, `createTypes` was a new utility passed to `sourceNodes` in gatsby@2.2.0.

The idea is we'll progressively improve the error message when/if we add more metadata to `apis.json` in the future. For context, the below is a sample `apis.json` that will be produced with this PR.

```json
{
  "ssr": {
    "replaceRenderer": {},
    "onRenderBody": {},
    "onPreRenderHTML": {},
    "wrapPageElement": {},
    "wrapRootElement": {}
  },
  "browser": {
    "onClientEntry": {},
    "onInitialClientRender": {},
    "onPreRouteUpdate": {},
    "onRouteUpdateDelayed": {},
    "onRouteUpdate": {},
    "shouldUpdateScroll": {},
    "registerServiceWorker": {},
    "replaceComponentRenderer": {
      "deprecated": true
    },
    "wrapPageElement": {},
    "wrapRootElement": {},
    "onPrefetchPathname": {},
    "onPostPrefetchPathname": {},
    "disableCorePrefetching": {},
    "replaceHydrateFunction": {},
    "onServiceWorkerInstalled": {},
    "onServiceWorkerUpdateFound": {},
    "onServiceWorkerUpdateReady": {},
    "onServiceWorkerActive": {},
    "onServiceWorkerRedundant": {}
  },
  "node": {
    "resolvableExtensions": {},
    "createPages": {},
    "createPagesStatefully": {},
    "sourceNodes": {},
    "onCreateNode": {},
    "onCreatePage": {},
    "setFieldsOnGraphQLNodeType": {},
    "createSchemaCustomization": {
      "version": "2.12.0"
    },
    "createResolvers": {
      "version": "2.2.0"
    },
    "preprocessSource": {},
    "generateSideEffects": {},
    "onCreateBabelConfig": {},
    "onCreateWebpackConfig": {},
    "onPreInit": {},
    "onPreBootstrap": {},
    "onPostBootstrap": {},
    "onPreBuild": {},
    "onPostBuild": {},
    "onPreExtractQueries": {},
    "onCreateDevServer": {}
  }
}
```

Any thoughts @gatsbyjs/core? Any metadata that is currently missing that we could use to improve this experience?

## Related Issues

Begins implementing #16055.


DSchau:
I have no idea why unit tests failed -- seems unrelated 😢

pieh:
Adding `documentation` to gatsby's devDependencies seems to cause different node_modules structure (causing snapshot to differ).

I think this happens because we do already have `documentation` package in `gatsby-transformer-documentation` - but we use `documentation@10`, while this new package is `documentation@12` so both can't be hoisted

DSchau:
@pieh good catch -- it's fixed now.

Agreed, it's kinda why I've been leaning towards quite a bit less snapshots. They're kind of a fragile "easy" way to test things, when it seems better to perform tests on specific properties of the object's shape.

DSchau:
Note: after changing documentation to 10.x, `apis.json` still has the same content (e.g. version for those few APIs, deprecated field for that one API, etc.).

packages/gatsby/scripts/output-api-file.js
```diff
@@ -0,0 +1,48 @@
...
+            contents,
+            api
+          ]
+        })
+    })
+  )
+
+  const output = apis.reduce((merged, [output, api]) => {
+    merged[api] = output.reduce((mergedOutput, doc) => {
+      if (doc.namespace.startsWith('.')) {
```
pieh: I can't really figure out what this does - worth adding small note why this is here?

DSchau: Yeah -- this should block, good catch!

It's because of this line in gatsby-ssr.js, it's captured as an item in the output when it's not really an API.

```
/**
 * Object containing options defined in `gatsby-config.js`
 * @typedef {object} pluginOptions
 */
```

That creates:

```json
{
  "namespace": "pluginOptions"
}
```

whereas every other one is an export so namespace is `.replaceRenderer` for instance.

DSchau:
Woo-hoo 🚀 

https://unpkg.com/gatsby@2.13.41/apis.json

packages/gatsby/scripts/output-api-file.js
pieh: Right, I now do remember I similar thing to our docs (just implemented differently):

https://github.com/gatsbyjs/gatsby/blob/aea3e39d3e5b19e0c905857f8068bc4d351c2b87/www/src/pages/docs/browser-apis.js#L15-L20

(and I didn't add note on what it does and why it's there :S)

PR #13128, opened 2019-04-04
allow accesing server over the network by default

pgarciacamou:
## Description

Add `-H 0.0.0.0` as default to the templates when running the `develop` script.

## Related Issues

https://github.com/gatsbyjs/gatsby/issues/5801#issuecomment-395735520

PR #171, opened 2016-03-05
Change default develop host to "localhost" on windows

jquense:
Windows won't resolve 0.0.0.0 as localhost/127.0.0.1 implicitly, so hot loading and links break. This just changes the default to a more (I think) helpful default for win32 users, rather then getting surprised by a default that doesn't work.

(fair warning I wrote this in the browser editor, so sorry if something is off)

PR #16070, opened 2019-07-24
chore(docs): added quick tip for using host option with gatsby develop

gillkyle:
## Description

Added a note based on a tip from a tweet that people seem to be enjoying: https://twitter.com/nelsreitz/status/1154131687950954497

Note: @pieh mentioned that 0.0.0.0 might not work on all platforms

### DIFF
--- a/docs/docs/gatsby-cli.md
+++ b/docs/docs/gatsby-cli.md
@@ -74,7 +74,7 @@ You can now view gatsbyjs.org in the browser.
   On Your Network:  http://192.168.0.212:8000/ // highlight-line
 ```
 
-**Note**: you can't visit 0.0.0.0:8000 on Windows (but things will work with using either localhost:8000 or the "On Your Network" URL on Windows)
+**Note**: you can't visit 0.0.0.0:8000 on Windows (but things will work using either localhost:8000 or the "On Your Network" URL on Windows)
 
 ### `build`
 
--- a/docs/docs/gatsby-cli.md
+++ b/docs/docs/gatsby-cli.md
@@ -74,6 +74,8 @@ You can now view gatsbyjs.org in the browser.
   On Your Network:  http://192.168.0.212:8000/ // highlight-line
 ```
 
+**Note**: you can't visit 0.0.0.0:8000 on Windows (but things will work with using either localhost:8000 or the "On Your Network" URL on Windows)
+
 ### `build`
 
 At the root of a Gatsby site, compile your application and make it ready for deployment:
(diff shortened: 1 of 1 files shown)


marcysutton:
@pieh do you have any specific platforms in mind that we could test? This worked for me on the Mac host side and I could view the IP address it output in the logs from the Windows VM side.

KyleAMathews:
My memory was it doesn't work on windows. But perhaps it was Linux. You can search old issues. This used to be the default host for a bit

gillkyle:
See also: 

#171
#13128

Seems like maybe to specific of a tip to include based on discussion in those PRs.

pieh:
I think we should do some testing and see if it's feasible to set `0.0.0.0` by default on platforms that are known to be working.

But in meantime I don't see harm with getting this one in?

marcysutton:
I think these instructions would help illustrate usage–where otherwise the list of options aren't necessarily motivated by use cases. We could add a note about platform inconsistencies to give people a heads up?

pieh:
I just did some tests with few platforms: Mac OS / Windows / Linux (Ubuntu). Windows and Linux were in virtual machines.

I was running `gatsby develop -H 0.0.0.0` on each platform and trying to access development server on all platforms. Using Chrome on Mac OS and Windows, using Firefox on Linux.

I was checking:
 - Initial load
 - Hot webpack update (code change)
 - Hot data update (content change)

In my setup things worked in all possible combinations when using actual IP addresses (or `localhost:8000`). Only thing that didn't work is trying to visit `0.0.0.0:8000` on Windows (works on Linux + Mac OS) - so to me it seems like we could try default to using 0.0.0.0 and instead just print out `localhost` for Windows as local URL (and maybe other platforms)

docs/docs/gatsby-cli.md
```diff
@@ -57,6 +57,23 @@ Once you've installed a Gatsby site, go to the root directory of your project an
...
+
+Then the terminal will log information as usual, but will additionally include a URL that you can navigate to from a client on the same network to see how the site renders.
+
+```
+You can now view gatsbyjs.org in the browser.
+⠀
+  Local:            http://0.0.0.0:8000/
+  On Your Network:  http://192.168.0.212:8000/ // highlight-line
+```
+
```
pieh: After my testing it seems like this could be shipped after adding note that you can't visit `0.0.0.0:8000` on Windows (but will work with either `localhost:8000` or "On Your Network" URL.

And then separately work on 0.0.0.0 + Windows displaying localhost instead and making 0.0.0.0 default?

marcysutton: Sounds good to me!

PR #16071, opened 2019-07-25
docs: Update Contentstack guide url

johnpkel:
## Description

Updated to the relative /docs/sourcing-from-contentstack url


gillkyle:
It looks like this change is already reflected on the most up to date version of `master` https://github.com/gatsbyjs/gatsby/blob/master/docs/docs/headless-cms.md#L32. I think I changed that when I formatted some things on the last PR, thanks @johnpkel!

PR #16079, opened 2019-07-25
little mini tiny typo

andihaki:
removing extra tick '


ajschmidt8:
pretty sure this is a valid apostrophe. it's used to show possession and indicates that the _dynamically-created class names_ belongs to Styled Components. gonna close this.

src: https://www.grammarbook.com/punctuation/apostro.asp

marcysutton:
That is true! The apostrophe was correct in this case, so it should stay in. But we're very happy to have you as a contributor, @andihaki, and I look forward to your future contributions!

PR #16082, opened 2019-07-25
fix(gatsby-cli): don't try to use local binaries

arcanis:
## Description

Execa pre-2.0 had a weird default: `preferLocal` was causing execa to modify the PATH in order to prepend the directory containing the Node binary. This is annoying because if you happen to run a custom `yarn` binary while having a Yarn build installed next to the Node binary (which is fairly common), then execa will spawn the global one rather than the one that should have been selected according to the regular PATH traversal.

This behavior got [disabled by default in execa 2.0+](https://codeburst.io/execa-v2-20ffafeedfdf#18d9), and I think it would make sense to do it here as well (you seem to only be using execa on git, yarn, and npm).


pieh:
Published `gatsby-cli@2.7.20`

PR #16084, opened 2019-07-25
fix(babel-preset-gatsby): Adds a missing `resolve` call

arcanis:
## Description

Seems to be a typo. The resolve being missing caused `preset-env` to be resolved starting from the root of the project rather than `babel-preset-gatsby` as it should have been 🙂


pieh:
Published `babel-preset-gatsby@0.2.8` / `gatsby@2.13.40`

PR #16085, opened 2019-07-25
fix(gatsby): group and distinct optional fields

afenton90:
## Description

`gatsby@2.13.22` seems to have introduced an error when running queries for some `nodes.children`. The specific stack trace was:

```
TypeError: Cannot read property 'args' of undefined",
        "    at getProjectedField (/Users/alexanderfenton/test-site/node_modules/gatsby/dist/schema/resolvers.js:237:92)",
        "    at /Users/alexanderfenton/test-site/node_modules/gatsby/dist/schema/resolvers.js:42:17",
        "    at resolveFieldValueOrError (/Users/alexanderfenton/test-site/node_modules/graphql/execution/execute.js:486:18)",
        "    at resolveField (/Users/alexanderfenton/test-site/node_modules/graphql/execution/execute.js:453:16)",
        "    at executeFields (/Users/alexanderfenton/test-site/node_modules/graphql/execution/execute.js:294:18)",
        "    at executeOperation (/Users/alexanderfenton/test-site/node_modules/graphql/execution/execute.js:238:122)",
        "    at executeImpl (/Users/alexanderfenton/test-site/node_modules/graphql/execution/execute.js:85:14)",
        "    at execute (/Users/alexanderfenton/test-site/node_modules/graphql/execution/execute.js:62:256)",
        "    at /Users/alexanderfenton/test-site/node_modules/express-graphql/dist/index.js:180:37",
        "    at processTicksAndRejections (internal/process/task_queues.js:85:5)
```

This was being caused because `group` and `distinct` where expected on all `*Connection` types. However, these fields are not always generated.

## Related Issues

N/A

### DIFF
--- a/packages/gatsby/src/schema/resolvers.js
+++ b/packages/gatsby/src/schema/resolvers.js
@@ -29,7 +29,7 @@ const findManyPaginated = typeName => async (source, args, context, info) => {
   // `distinct` which might need to be resolved.
   const group = getProjectedField(info, `group`)
   const distinct = getProjectedField(info, `distinct`)
-  const extendedArgs = { ...args, group: group || [], distinct: distinct || [] }
+  const extendedArgs = { ...args, group, distinct }
 
   const result = await findMany(typeName)(source, extendedArgs, context, info)
   return paginate(result, { skip: args.skip, limit: args.limit })
@@ -234,10 +234,11 @@ const getProjectedField = (info, fieldName) => {
     info
   )
 
+  const fields = getNullableType(info.returnType).getFields()
+  if (!fields[fieldName]) return []
+
   const fieldEnum = getNullableType(
-    getNullableType(info.returnType)
-      .getFields()
-      [fieldName].args.find(arg => arg.name === `field`).type
+    fields[fieldName].args.find(arg => arg.name === `field`).type
   )
 
   return fieldNodes.reduce((acc, fieldNode) => {


stefanprobst:
Could you clarify the scenario that is currently breaking and this PR fixes? -- would be good to have a test for that so we don't break in the future (good place for a test would be [here](https://github.com/gatsbyjs/gatsby/blob/510dad67a22bac1569b8fa6ed3e2b3f3fb17507d/packages/gatsby/src/schema/__tests__/queries.js#L658)) Thanks!

afenton90:
Hey @stefanprobst here is a codesandbox to demonstrate the issue https://codesandbox.io/s/group-and-distinct-example-pl9wc

If you run the following query in graphiql you will get the error mentioned above:
```
{
  allTagGroup {
    totalCount
  }
}
```

I'll work on adding a test in.

afenton90:
I agree this could really do with a test, however I'm not sure where to add a test?
It could do with something like:
```
beforeAll(() => {
 addCustomAllQuery()
})

it('does not error when querying all* query without group and distinct fields`, () => {
....
})
```

stefanprobst:
@afenton90 thanks for the codesandbox!

so it looks like this is actually a different issue: in your example, you create a `Tag` type and a `TagGroup` type.

Gatsby will for every top-level type create some extra types TypeFilterInput, TypeSortInput, TypeEdge, TypeConnection, TypeGroupConnection.

So Gatsby will create
* for `Tag`: `TagConnection`, `TagGroupConnection` etc
* for `TagGroup`: `TagGroupConnection`, `TagGroupGroupConnnection`
=> we have two types called `TagGroupConnection` and this messes things up.

You can try your example by renaming `TagGroup` to e.g. `TagsGroup` and the query will work.

afenton90:
@stefanprobst oh wow yeah that was the issue! Thanks for that.
If there are no docs for the types that are generated I may add some in. I appreciate having a warning for clashing types it in the console would be undesirable as it would be really difficult to see which type overrides were intentional and which were not.

PR #16091, opened 2019-07-25
feat(gatsby): Materialization

freiksenet:
Add a node backend independent materialization layer that triggers when nodes are searched on fields only available through resolvers.

In addition, this fixes various incompatibilities that loki had due to never being tested through materialization.

In addition, node tracking is now happening on node-model level too.

In addition, this adds extensions to decide which fields are sortable and searchable and which needs resolve, so we can materialize.

Can be tested via `gatsby@materialization`


sidharthachatterjee:
@Ishaan28malik Please do not approve draft PRs or other PRs without giving them a review

packages/gatsby/src/schema/node-model.js
```diff
@@ -168,40 +202,137 @@ class LocalNodeModel {
...
+    } else if (result) {
+      result.forEach(node => this.trackInlineObjectsInRootNode(node))
     }
 
     return this.trackPageDependencies(result, pageDependencies)
   }
 
+  prepareNodes(type, queryFields, fieldsToResolve, nodeTypeNames) {
+    const typeName = type.name
+    if (!this._prepareNodesQueues[typeName]) {
```
freiksenet: Instead of running resolving immediately, we delay it.

This is because GraphQL is async, several resolvers that all need to prepare fields can query asynchronously. By queueing them, we make sure that all the promises inside this particular tick are queued, thus we shouldn't have conflicts.

packages/gatsby/src/schema/node-model.js
```diff
@@ -168,40 +202,137 @@ class LocalNodeModel {
...
+    }
+
+    this._prepareNodesQueues[typeName].push({
+      queryFields,
+      fieldsToResolve,
+    })
+
+    if (!this._prepareNodesPromises[typeName]) {
+      this._prepareNodesPromises[typeName] = new Promise(resolve => {
+        process.nextTick(async () => {
```
freiksenet: Enqueue actual resolving.

packages/gatsby/src/schema/node-model.js
```diff
@@ -168,40 +202,137 @@ class LocalNodeModel {
...
+    return this._prepareNodesPromises[typeName]
+  }
+
+  async _doResolvePrepareNodesQueue(type, nodeTypeNames) {
+    const typeName = type.name
+    const queue = this._prepareNodesQueues[typeName]
+    this._prepareNodesQueues[typeName] = []
+    this._prepareNodesPromises[typeName] = null
+
+    const { queryFields, fieldsToResolve } = queue.reduce(
```
freiksenet: Merge fields that everything in a queue wants to resolve.

packages/gatsby/src/schema/node-model.js
```diff
@@ -168,40 +202,137 @@ class LocalNodeModel {
...
+          fieldsToResolve: _.merge(fieldsToResolve, nextFieldsToResolve),
+        }
+      },
+      {
+        queryFields: {},
+        fieldsToResolve: {},
+      }
+    )
+
+    const actualFieldsToResolve = deepObjectDifference(
```
freiksenet: Check what we already have resolved and cached.

packages/gatsby/src/schema/node-model.js
```diff
@@ -212,15 +343,48 @@ class LocalNodeModel {
...
-   * @param {(Object|Array)} obj An object belonging to a Node, or a Node object
-   * @param {Function} [predicate] Optional condition to match
-   * @returns {(Node|null)}
+   * Adds link between inline objects/arrays contained in Node object
+   * and that Node object.
+   * @param {Node} node Root Node
    */
-  findRootNodeAncestor(obj, predicate) {
-    return this.nodeStore.findRootNodeAncestor(obj, predicate)
+  trackInlineObjectsInRootNode(node) {
```
freiksenet: Track inline object / findRootNode now live in node model.

packages/gatsby/src/schema/node-model.js
```diff
@@ -253,6 +413,60 @@ class LocalNodeModel {
   }
 }
 
+class ContextualNodeModel {
```
freiksenet: Node model that has a context, in tihs case it's basically just path. It forwards everything but `trackPageDependencies` to node model and adds default path otherwise.

packages/gatsby/src/schema/schema.js
```diff
@@ -153,6 +153,11 @@ const processTypeComposer = async ({
       fieldExtensions,
       parentSpan,
     })
+    await determineSearchableFields({
```
freiksenet: We mark fields that are searchable (so should be included in filter), sortable (so should be included in sort) and resolvable (so should be prepared) via extensions metadata.

packages/gatsby/src/schema/schema.js
```diff
@@ -689,6 +703,40 @@ const addCustomResolveFunctions = async ({ schemaComposer, parentSpan }) => {
   })
 }
 
+const determineSearchableFields = ({ schemaComposer, typeComposer }) => {
```
freiksenet: Currently we make everything searchable and resolvable. However in future we might change that behaviour, so that weird things like `fixed` in sharp aren't searchable.

packages/gatsby/src/schema/infer/type-conflict-reporter.js
```diff
@@ -22,9 +21,10 @@ const isNodeWithDescription = node =>
 
 const findNodeDescription = obj => {
   if (obj) {
-    const node = findRootNodeAncestor(obj, isNodeWithDescription)
-    if (isNodeWithDescription(node)) {
-      return node.internal.description
+    // TODO: Maybe get this back
```
freiksenet: Not sure if this ever happened, not sure how to make it work now. I think it's not a big deal, but I can try to restore that behaviour otherwise.

packages/gatsby/src/db/loki/nodes-query.js
```diff
@@ -48,12 +50,16 @@ emitter.on(`DELETE_CACHE`, () => {
...
 function toMongoArgs(gqlFilter, lastFieldType) {
+  lastFieldType = getNullableType(lastFieldType)
   const mongoArgs = {}
   _.each(gqlFilter, (v, k) => {
     if (_.isPlainObject(v)) {
       if (k === `elemMatch`) {
-        const gqlFieldType = lastFieldType.ofType
-        mongoArgs[`$elemMatch`] = toMongoArgs(v, gqlFieldType)
+        mongoArgs[`$elemMatch`] = toMongoArgs(v, getNamedType(lastFieldType))
+      } else if (lastFieldType instanceof GraphQLList) {
```
sidharthachatterjee: Unsure what's going on here

packages/gatsby/src/db/loki/nodes-query.js
```diff
@@ -244,28 +293,71 @@ function ensureFieldIndexes(coll, lokiArgs) {
  * @returns {promise} A promise that will eventually be resolved with
  * a collection of matching objects (even if `firstOnly` is true)
  */
-async function runQuery({ gqlType, queryArgs, firstOnly }) {
+async function runQuery(args) {
+  if (args.nodeTypeNames.length > 1) {
```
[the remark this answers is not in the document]
sidharthachatterjee: Does Loki not support querying more than 1 table?

packages/gatsby/src/db/nodes-query.js
```diff
@@ -1,26 +1,5 @@
...
-
-  const { queryArgs, gqlType } = args
-  const { filter, sort, group, distinct } = queryArgs
-  const fields = getQueryFields({ filter, sort, group, distinct })
-
-  // NOTE: `hasFieldResolvers` is also true for Date fields
-  if (
-    backend === `loki` &&
-    !args.nodes &&
-    !hasFieldResolvers(gqlType, fields)
```
sidharthachatterjee: Does this mean that `hasFieldResolvers` is now supported for Loki backend as well?

freiksenet: Yes, it's all moved on one level up, nodes-query already expects materialization to have happened.

packages/gatsby/src/query/index.js
```diff
@@ -249,17 +250,36 @@ const runQueuedQueries = () => {
  *
  * For what constitutes a dirty query, see `calcQueries`
  */
-const startListening = queue => {
+const startListeningToDevelopQueue = queue => {
```
m-allanson: Can this `queue` arg be removed now?

packages/gatsby/src/query/index.js
```diff
@@ -249,17 +250,36 @@ const runQueuedQueries = () => {
...
   )
 
   emitter.on(`API_RUNNING_QUEUE_EMPTY`, runQueuedQueries)
+  ;[
+    `DELETE_CACHE`,
+    `CREATE_NODE`,
+    `DELETE_NODE`,
+    `DELETE_NODES`,
+    `SET_SCHEMA_COMPOSER`,
+    `SET_SCHEMA`,
```
m-allanson: Do we need to also handle `ADD_FIELD_TO_NODE` and `ADD_CHILD_NODE_TO_PARENT_NODE`?

Or can we remove these two in a later version of Gatsby?

packages/gatsby/src/redux/actions/public.js
```diff
@@ -616,6 +618,9 @@ const createNode = (
     node.internal = {}
   }
 
+  NODE_COUNTER++
```
m-allanson:
```suggestion
// allows 'insertion order' sorting when needed (date is not precise enough)
  NODE_COUNTER++
```

packages/gatsby/src/redux/actions/public.js
```diff
@@ -616,6 +618,9 @@ const createNode = (
...
+  node.internal.$counter = NODE_COUNTER
```
m-allanson: Should this be added to the Joi schema validation? Why isn't it causing errors?

m-allanson: Also do we need the $ prefix?

packages/gatsby/src/redux/nodes.js
```diff
@@ -82,3 +82,47 @@ exports.getNodeAndSavePathDependency = (id, path) => {
   createPageDependency({ path, nodeId: id })
   return node
 }
+
+exports.updateNodesByType = async (typeName, updater) => {
```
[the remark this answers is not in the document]
m-allanson: Let's call this `saveResolvedNodes` or something similar, to be more specific about purpose.

packages/gatsby/src/redux/run-sift.js
```diff
@@ -68,29 +65,96 @@ function handleFirst(siftArgs, nodes) {
...
+      })
+      .map(field => v => getValueAt(v, field))
     const sortOrder = sort.order.map(order => order.toLowerCase())
 
     result = _.orderBy(result, sortFields, sortOrder)
   }
   return result
 }
 
+const toDottedFields = (filter, acc = {}, path = []) => {
```
wardpeet: Can you add jsdoc with @example so we know what goes in and what goes out?

packages/gatsby/src/redux/run-sift.js
```diff
@@ -68,29 +65,96 @@ function handleFirst(siftArgs, nodes) {
...
+    } else if (_.isPlainObject(nextValue)) {
+      toDottedFields(value, acc, path.concat(key))
+    } else {
+      acc[path.concat(key).join(`.`)] = value
+    }
+  })
+  return acc
+}
+
+const objectToDottedField = (obj, path = []) => {
```
wardpeet: Can you add jsdoc with @example so we know what goes in and what goes out?

packages/gatsby/src/schema/node-model.js
```diff
@@ -51,14 +56,31 @@ export interface NodeModel {
...
 }
 
 class LocalNodeModel {
-  constructor({ schema, nodeStore, createPageDependency, path }) {
+  constructor({
+    schema,
+    schemaComposer,
+    nodeStore,
+    createPageDependency,
+    preparedNodesCache,
```
davidbailey00: no longer used; can remove

packages/gatsby/src/schema/node-model.js
davidbailey00: use `nextTick` to prevent race conditions by waiting for other promises

packages/gatsby/src/schema/node-model.js
```diff
@@ -253,6 +413,60 @@ class LocalNodeModel {
...
+    return this.nodeModel.trackInlineObjectsInRootNode(...args)
+  }
+
+  findRootNodeAncestor(...args) {
+    return this.nodeModel.findRootNodeAncestor(...args)
+  }
+
+  trackPageDependencies(result, pageDependencies) {
+    const fullDependencies = {
+      path: this.path,
```
davidbailey00: should be context.path

packages/gatsby/src/schema/node-model.js
wardpeet: Any reason why we don't use inheritance?

packages/gatsby/src/db/loki/nodes-query.js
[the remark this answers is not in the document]
freiksenet: No.

packages/gatsby/src/redux/actions/public.js
freiksenet: It's not in loki because it's added after validation. I've moved it to be just "counter".

packages/gatsby/src/schema/types/filter.js
```diff
@@ -41,13 +49,25 @@ const convert = ({
   fieldNames.forEach(fieldName => {
     const fieldConfig = inputTypeComposer.getFieldConfig(fieldName)
     const type = getNamedType(fieldConfig.type)
+    const searchable = typeComposer.getFieldExtension(fieldName, `searchable`)
+
+    if (searchable === SEARCHABLE_ENUM.NOT_SEARCHABLE) {
+      return
+    } else if (searchable === SEARCHABLE_ENUM.DEPRECATED_SEARCHABLE) {
+      deprecationReason = `Filtering on fields that need arguments to resolve is deprecated.`
```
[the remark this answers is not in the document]
freiksenet: Eg filtering on `fixed` in ImageSharp.

packages/gatsby/src/schema/node-model.js
freiksenet: I just thought this is cleaner.

packages/gatsby/src/db/loki/nodes-query.js
```diff
@@ -152,6 +158,24 @@ const toDottedFields = (filter, acc = {}, path = []) => {
   return acc
 }
 
+// Like above, but doesn't handle $elemMatch
+const objectToDottedField = (obj, path = []) => {
```
stefanprobst: Looks like this is the same in `redux/run-sift` and `loki/nodes-query`? can we move it do `src/db/common/`?

packages/gatsby/src/db/loki/nodes-query.js
```diff
@@ -180,9 +204,34 @@ const fixNeTrue = filter =>
     return acc
   }, {})
 
+const liftResolvedFields = (args, resolvedFields) => {
```
stefanprobst: Same as above: Looks like this is the same in `redux/run-sift` and `loki/nodes-query`? can we move it do `src/db/common/`?

packages/gatsby/src/db/loki/nodes-query.js
```diff
@@ -244,28 +294,71 @@ function ensureFieldIndexes(coll, lokiArgs) {
...
  * a collection of matching objects (even if `firstOnly` is true)
  */
-async function runQuery({ gqlType, queryArgs, firstOnly }) {
+async function runQuery(args) {
+  if (args.nodeTypeNames.length > 1) {
+    const nodes = args.nodeTypeNames.reduce(
+      (acc, typeName) => acc.concat(getNodesByType(typeName)),
+      []
+    )
+    return runSiftOnNodes(nodes, args, getNode)
```
stefanprobst: Do we have to fall back on sift here? can't we query each node type collection separately and then concatenate and sort the results? this way we could potentially benefit from loki indices when filtering, and had a better separation between the backends.

packages/gatsby/src/db/loki/nodes-query.js
```diff
@@ -244,28 +294,71 @@ function ensureFieldIndexes(coll, lokiArgs) {
...
+  const {
+    gqlType,
+    queryArgs,
+    firstOnly,
+    resolvedFields = {},
+    nodeTypeNames,
+  } = args
   // Clone args as for some reason graphql-js removes the constructor
   // from nested objects which breaks a check in sift.js.
   const gqlArgs = JSON.parse(JSON.stringify(queryArgs))
```
stefanprobst: i don't think this is still needed?

packages/gatsby/src/db/nodes-query.js
stefanprobst: i think `hasFieldResolvers` is not used anymore and can be removed from `db/common/query`?

packages/gatsby/src/db/nodes-query.js
```diff
@@ -52,10 +33,14 @@ function chooseQueryEngine(args) {
  * a collection of matching objects (even if `firstOnly` is true, in
  * which case it will be a collection of length 1 or zero)
  */
-function run(args) {
-  const queryFunction = chooseQueryEngine(args)
+async function run(args) {
+  const { backend } = require(`./nodes`)
 
-  return queryFunction(args)
+  if (backend === `redux`) {
```
stefanprobst: in the future would be good if we didn't have to check backend on every query

packages/gatsby/src/redux/nodes.js
```diff
@@ -82,3 +82,45 @@ exports.getNodeAndSavePathDependency = (id, path) => {
...
+
+exports.saveResolvedNodes = async (typeName, resolver, nodeTypeNames) => {
+  const nodes = store.getState().nodesByType.get(typeName)
+  const resolvedNodes = new Map()
+  if (nodes) {
+    for (const node of nodes.values()) {
+      const resolved = await resolver(node)
+      resolvedNodes.set(node.id, resolved)
+    }
+    store.dispatch({
```
stefanprobst: could you add comment why we do this differently here than we do with loki? if i understand correctly, with loki we update the collection with the resolved nodes, but here we use a separate `resolvedNodes` Map? Also, how are `nodeTypeNames` handled here?

packages/gatsby/src/db/loki/nodes.js
```diff
@@ -290,6 +293,63 @@ function deleteNodes(nodes) {
   }
 }
 
+const saveResolvedNodes = async (typeName, resolver, nodeTypeNames) => {
```
stefanprobst: do we use `typeName` here?

packages/gatsby/src/redux/run-sift.js
```diff
@@ -41,11 +39,11 @@ const getFilters = filters =>
 // Run Sift
 /////////////////////////////////////////////////////////////////////
 
-function isEqId(firstOnly, fieldsToSift, siftArgs) {
+function isEqId(firstOnly, siftArgs) {
   return (
     firstOnly &&
-    Object.keys(fieldsToSift).length === 1 &&
-    Object.keys(fieldsToSift)[0] === `id` &&
+    siftArgs.length > 0 &&
```
stefanprobst: should this be `siftArgs.length === 1`? if we have more than one filter, this special case does not apply?

packages/gatsby/src/redux/run-sift.js
```diff
@@ -68,29 +65,128 @@ function handleFirst(siftArgs, nodes) {
...
+//   "id": {
+//     $regex: // as above
+//   }
+// }
+const toDottedFields = (filter, acc = {}, path = []) => {
+  Object.keys(filter).forEach(key => {
+    const value = filter[key]
+    const nextValue = _.isPlainObject(value) && value[Object.keys(value)[0]]
+    if (key === `$elemMatch`) {
+      acc[path.join(`.`)] = { [`$elemMatch`]: value }
```
stefanprobst: this line is different than in the `toDottedFields` loki version, is that on purpose? if not, we could also move this to `db/common`

packages/gatsby/src/query/graphql-runner.js
```diff
@@ -0,0 +1,40 @@
...
+      query,
+      context,
+      withResolverContext({
+        schema,
+        schemaComposer: schemaCustomization.composer,
+        context,
+        customContext: schemaCustomization.context,
+        nodeModel: this.nodeModel,
+      }),
+      context
```
stefanprobst: this is not important but do we know what depends on injecting `context` for root value and variables?

packages/gatsby/src/schema/__tests__/node-model.js
```diff
@@ -36,335 +35,616 @@ describe(`NodeModel`, () => {
...
+        { path: `/` }
+      )
+      expect(resolveBetterTitleMock.mock.calls.length).toBe(1)
+      expect(resolveOtherTitleMock.mock.calls.length).toBe(1)
+      await nodeModel.runQuery(
+        {
+          query: {
+            filter: { betterTitle: { eq: `foo` }, otherTitle: { eq: `Bar` } },
+          },
+          firstOnly: false,
```
stefanprobst: can we add a test with `firstOnly: true` as well?

packages/gatsby/src/schema/node-model.js
```diff
@@ -85,6 +107,10 @@ class LocalNodeModel {
       result = nodeTypeNames.includes(node.internal.type) ? node : null
     }
 
+    if (result) {
+      this.trackInlineObjectsInRootNode(node)
```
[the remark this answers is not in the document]
stefanprobst: do we now track nodes on every query? i *think* that previously we only updated the rootnodemap when we resolved fields?

packages/gatsby/src/schema/node-model.js
```diff
@@ -278,6 +485,251 @@ const toNodeTypeNames = (schema, gqlTypeName) => {
...
+
+const merge = (...objects) => {
+  const [first, ...rest] = objects.filter(Boolean)
+  return rest.reduce((acc, obj) => mergeObjects(acc, obj), { ...first })
+}
+
+async function resolveRecursive(
+  nodeModel,
+  schemaComposer,
+  schema,
```
stefanprobst: not for this PR: but since we get the schema here porting  #16679 should be fairly unproblematic

packages/gatsby/src/schema/node-model.js
```diff
@@ -278,6 +485,251 @@ const toNodeTypeNames = (schema, gqlTypeName) => {
...
+  node,
+  gqlField,
+  fieldName
+) {
+  const withResolverContext = require(`./context`)
+  return gqlField.resolve(
+    node,
+    {},
```
stefanprobst: we should pass the args defaultValues here, something like
```suggestion
    gqlField.args.reduce((acc, arg) => {
      acc[arg.name] = arg.defaultValue
      return acc
    }, {}),
```

packages/gatsby/src/schema/schema.js
```diff
@@ -717,6 +731,40 @@ const addCustomResolveFunctions = async ({ schemaComposer, parentSpan }) => {
...
+  typeComposer.getFieldNames().forEach(fieldName => {
+    const field = typeComposer.getField(fieldName)
+    const extensions = typeComposer.getFieldExtensions(fieldName)
+    if (field.resolve) {
+      if (extensions.dateformat) {
+        typeComposer.extendFieldExtensions(fieldName, {
+          searchable: SEARCHABLE_ENUM.SEARCHABLE,
+          sortable: SORTABLE_ENUM.SORTABLE,
+          needsResolve: extensions.proxyFrom ? true : false,
```
stefanprobst: hmm this is a very specific scenario

stefanprobst:
```suggestion
          needsResolve: extensions.proxy ? true : false,
```

packages/gatsby/src/schema/schema.js
```diff
@@ -717,6 +731,40 @@ const addCustomResolveFunctions = async ({ schemaComposer, parentSpan }) => {
   })
 }
 
+const determineSearchableFields = ({ schemaComposer, typeComposer }) => {
+  typeComposer.getFieldNames().forEach(fieldName => {
+    const field = typeComposer.getField(fieldName)
+    const extensions = typeComposer.getFieldExtensions(fieldName)
+    if (field.resolve) {
+      if (extensions.dateformat) {
+        typeComposer.extendFieldExtensions(fieldName, {
```
stefanprobst: not necessarily in this PR, but we should have some tests for the new extensions

freiksenet:
> Does this also solve the issue that source was different in a field resolver depending on whether the field had already been resolved as a filter?

Yes, source is always original node.

> For the future, can we somehow use the cache also when resolving fields in the selection set?

We should make a wrapper for `prepareNodes` in nodeModel that accepts projection. It can use the cache in that case, because it will still go through prepare nodes. I think actually nodes will have the `__gatsby_resolved` field already if they come through the query. We can modify other accessors to do that too.

packages/gatsby/src/db/loki/nodes-query.js
freiksenet: That is a good idea, but let's add it in future PR.

packages/gatsby/src/schema/node-model.js
freiksenet: :+1:

packages/gatsby/src/schema/schema.js
freiksenet: It is. I'm open to better logic here. The idea is that our dateformat should always be searchable, even with args. Because our resolvers are ultimately functions, I'm not sure what would be better logic here.

stefanprobst:
> Yes, source is always original node.

Nice! This allows us to clean up quite a bit in our internal resolvers

>  wrapper for `prepareNodes` in nodeModel that accepts projection

:+1:

packages/gatsby/src/redux/nodes.js
freiksenet: I didn't want to mutate nodes directly in Redux store, because I was worried there would be some kind of sideeffects of that.

packages/gatsby/src/db/loki/nodes.js
freiksenet: No, because loki has separate tables, we store it via implementation type names.

packages/gatsby/src/redux/run-sift.js
freiksenet: Yes, thanks!

packages/gatsby/src/query/graphql-runner.js
freiksenet: I don't think anything depends on root value. I'd fix it in a separate PR, move it so that we pass context and variables separately.

freiksenet:
> do we now track nodes on every query? i think that previously we only updated the rootnodemap when we resolved fields?

Somehow this is not answerable inline. That's a great point, we should add a cache that skips that.

freiksenet:
Published as `gatsby@2.14.0-rc.1` / `gatsby@materialization`.

packages/gatsby/src/db/loki/nodes.js
stefanprobst: can we then remove `typeName` from the signature -- in the redux side of `saveResolvedNodes` we'd also want to use `nodeTypeNames` not `typeName` right?

packages/gatsby/src/schema/schema.js
stefanprobst: what happens when `dateformat` is combined with an extension other than `proxy` that also needs resolving, e.g. `projection` or similar? maybe these enums should apply per-resolver/extension, not per-field, since a field can have multiple wrapped resolvers?

packages/gatsby/src/db/loki/nodes.js
freiksenet: That is actually true. Good point.

freiksenet:
Published as `gatsby@2.14.0-rc.2` / `gatsby@materialization`

packages/gatsby/src/schema/node-model.js
```diff
@@ -341,13 +342,10 @@ class LocalNodeModel {
...
    */
   trackInlineObjectsInRootNode(node) {
-    return addRootNodeToInlineObject(
-      this._rootNodeMap,
-      node,
-      node.id,
-      true,
-      true
-    )
+    if (!this._trackedRootNodes.has(node.id)) {
```
stefanprobst: does this work with hot-reloading, i.e. when the node.id stays the same but some nested field changes and needs to be re-tracked?

packages/gatsby/src/schema/__tests__/__snapshots__/kitchen-sink.js.snap
```diff
@@ -10,9 +10,9 @@ Object {
         "likes": 33,
       },
       Object {
-        "code": "BdIpEzalAJx",
-        "id": "1677771511187112561",
-        "likes": 29,
+        "code": "Bam3G5XFOTK",
+        "id": "1632234281433883850",
+        "likes": 30,
```
wardpeet: how come this is 30?

packages/gatsby/src/schema/node-model.js
wardpeet: from my perspective it isn't but I don't have a strong opinion about it.

stefanprobst:
`getQueryFields`, `mergeObjects`, `merge`, `pathToObject`, `dropQueryOperators` are now in `node-model` and can be removed from `db/common/query` i think

packages/gatsby/src/schema/node-model.js
freiksenet: Changing nodes resets node model.

freiksenet:
Published as `gatsby@2.14.0-rc.3` / `gatsby@materialization`

freiksenet:
> getQueryFields, mergeObjects, merge, pathToObject, dropQueryOperators are now in node-model and can be removed from db/common/query i think

Cheers, fixed.

packages/gatsby/src/schema/node-model.js
```diff
@@ -51,14 +56,26 @@ export interface NodeModel {
...
+
+    this._rootNodeMap = new WeakMap()
+    this._trackedRootNodes = new Set()
+    this._prepareNodesQueues = {}
+    this._prepareNodesPromises = {}
+    this._preparedNodesCache = new Map()
+  }
+
+  withContext(context) {
+    return new ContextualNodeModel(this, context)
```
wardpeet: I do wonder why we not augment this model instead?

packages/gatsby/src/schema/node-model.js
```diff
@@ -278,6 +483,254 @@ const toNodeTypeNames = (schema, gqlTypeName) => {
...
+  Object.keys(fields).forEach(fieldName => {
+    const field = fields[fieldName]
+    const gqlField = gqlFields[fieldName]
+    const gqlFieldType = getNamedType(gqlField.type)
+    const typeComposer = schemaComposer.getAnyTC(type.name)
+    const needsResolve = typeComposer.getFieldExtension(
+      fieldName,
+      `needsResolve`
+    )
+    if (_.isObject(field) && gqlField) {
```
wardpeet:
```js
	
    if (_.isObject(field) && gqlField) {
      const innerResolved = determineResolvableFields(
        schemaComposer,
        schema,
        gqlFieldType,
        field
      )
      if (!_.isEmpty(innerResolved)) {
        fieldsToResolve[fieldName] = innerResolved
      }
    }

    if (!fieldsToResolve[fieldName] && needsResolve) {
      fieldsToResolve[fieldName] = true
    }

```

packages/gatsby/src/schema/node-model.js
```diff
@@ -278,6 +483,254 @@ const toNodeTypeNames = (schema, gqlTypeName) => {
...
+  rootNodeMap,
+  data,
+  nodeId,
+  isNode = false
+) => {
+  const isPlainObject = _.isPlainObject(data)
+
+  if (isPlainObject || _.isArray(data)) {
+    _.each(data, (o, key) => {
+      if (isNode && key === `internal`) {
```
wardpeet: we can drop the return 
```suggestion
      if (!isNode || key !== `internal`) {
        addRootNodeToInlineObject(rootNodeMap, o, nodeId)
      }
```

packages/gatsby/src/schema/node-model.js
```diff
@@ -278,6 +483,254 @@ const toNodeTypeNames = (schema, gqlTypeName) => {
...
+    }
+  })
+  return fieldsToResolve
+}
+
+/**
+ * Add link between passed data and Node. This function shouldn't be used
+ * directly. Use higher level `trackInlineObjectsInRootNode`
+ * @see trackInlineObjectsInRootNode
+ * @param {(Object|Array)} data Inline object or array
```
wardpeet: jsdoc isn't correct. remove params or fix

packages/gatsby/src/schema/node-model.js
```diff
@@ -278,6 +483,254 @@ const toNodeTypeNames = (schema, gqlTypeName) => {
...
+    if (_.isPlainObject(value) && _.isPlainObject(v)) {
+      acc[key] =
+        k === `elemMatch` ? dropQueryOperators(v) : dropQueryOperators(value)
+    } else {
+      acc[key] = true
+    }
+    return acc
+  }, {})
+
+const mergeObjects = (obj1, obj2) =>
```
wardpeet: i wonder if we can leverage lodash helpers instead.

packages/gatsby/src/schema/__tests__/__snapshots__/kitchen-sink.js.snap
freiksenet: Sorting was broken, I fixed it.

packages/gatsby/src/schema/node-model.js
freiksenet: There can be multiple contextual node models rooting from one node model. They share state, but have different paths.

packages/gatsby/src/schema/node-model.js
freiksenet: Cheers!

packages/gatsby/src/schema/node-model.js
freiksenet: Seems we can.

freiksenet:
Published as `gatsby@2.14.0-rc.4` / `gatsby@materialization`

PR #16094, opened 2019-07-25
feat(gatsby-transformer-react-docgen): upgrade to react-docgen 5

jquense:
BREAKING CHANGE: maybe, the interal api's chagned so folks using custom resolvers will probably be broken, but generally the default user should be unaffected

### DIFF
--- a/packages/gatsby-transformer-react-docgen/package.json
+++ b/packages/gatsby-transformer-react-docgen/package.json
@@ -11,7 +11,7 @@
     "@babel/runtime": "^7.0.0",
     "@babel/types": "^7.0.0",
     "common-tags": "^1.4.0",
-    "react-docgen": "^4.1.0",
+    "react-docgen": "^5.0.0-beta.1",
     "react-docgen-displayname-handler": " ^2.1.0"
   },
   "devDependencies": {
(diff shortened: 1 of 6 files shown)


jquense:
yup to ya'll whether this counts as a breaking change. Strictly speaking some users will be broken tho

jquense:
ping

sidharthachatterjee:
@jquense Thank you for doing this! The new release looks great (TypeScript support 😍)

I would definitely want to release this as a major version bump (owning to the breaking changes in the resolver signature documented in https://github.com/reactjs/react-docgen/releases/tag/v5.0.0-beta.1) 

Shall we wait till `react-docgen` v5 is out of beta?

jquense:
that is likely to be months at the pace the project tends to move, integers are cheap, so why wait, can always bump again if needed. Up to ya'll but i've published my own version for now so we can get typescript support

sidharthachatterjee:
@jquense Fair enough! I’ll get this published as a major today

sidharthachatterjee:
Published in `gatsby-transformer-react-docgen@5.0.0`

jquense:
🎉 Thanks!

PR #16100, opened 2019-07-25
fix(docs): miscellaneous docs reorg

amberleyromo:
##  Description

- Link accessibility guide in the "building themes" guide
- Move theme API doc to api-reference section, create redirect
- Add theme tutorials section to tutorial index
- Rearrange docs landing page to highlight two main pathways.


www/gatsby-node.js
```diff
@@ -46,6 +46,12 @@ const slugToAnchor = slug =>
 exports.createPages = ({ graphql, actions, reporter }) => {
   const { createPage, createRedirect } = actions
 
+  createRedirect({
+    fromPath: `/docs/themes/api-reference`, // Merged Component CSS and CSS Modules
+    toPath: `/docs/themes-api/`,
```
johno: It looks like the path down below is `/theme-api/` rather than `/themes-api/`, not sure if that's intentional?

amberleyromo: oh, nope oversight. gracias.

marcysutton: That's the only thing that needs to be updated, IMO!

amberleyromo: fixed!

marcysutton:
One thing we should add is a link to the Theme API docs in their new location. I think the Building Themes page would be a good place for that, what do you think? https://www.gatsbyjs.org/docs/themes/building-themes

marcysutton:
Nice work on this, a great collection of changes!

PR #16109, opened 2019-07-26
feat(gatsby-transformer-sharp): define ImageSharp as child of File type

### LINKED ISSUE #16099: Schema Customization - Cannot Query Field "childImageSharpe" on type "File"
## Description

While developing a Gatsby Theme I customized the GraphQL schema  ([gatsby-node.js](https://github.com/ryanwiemer/gatsby-theme-amsterdam/blob/master/theme/gatsby-node.js#L20-L38)). This works as expected with all fields except the `cover` front matter field used for a cover image. If I omit the cover image on all posts the site still builds but if there are no posts at all I receive the following error message: 

`GraphQLError: Cannot query field "childImageSharp" on type "File"` 

### Steps to reproduce

1. `git clone https://github.com/ryanwiemer/gatsby-theme-amsterdam`
2. `yarn install`
3. Remove the `content` folder in `site/content/`
4. `yarn workspace site develop` - To start development using the site folder
5. You should now see the error message: `GraphQLError: Cannot query field "childImageSharp" on type "File"` 

### Expected result

Since I defined the schema through the Schema Customization I expect the site to build even when there is no content.

### Actual result

The site fails to build. 

### Environment

  System:
    OS: macOS High Sierra 10.13.6
    CPU: (8) x64 Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 12.6.0 - /var/folders/v8/dxrbb0vj7z9dtxj7mq5tc101d1qq1d/T/yarn--1564090827764-0.3745122210652396/node
    Yarn: 1.17.3 - /var/folders/v8/dxrbb0vj7z9dtxj7mq5tc101d1qq1d/T/yarn--1564090827764-0.3745122210652396/yarn
    npm: 6.10.2 - /usr/local/bin/npm
  Languages:
  Browsers:
    Chrome: 75.0.3770.142
    Firefox: 67.0.4
    Safari: 12.1.2

stefanprobst:
Querying on `File.childImageSharp` in a site/theme where no image files are present currently fails because children fields are not being created. This PR uses the new `@childOf` extension to explicitly define `ImageSharp` nodes as children of `File`nodes.

Fixes  #16099

### DIFF
--- a/packages/gatsby-transformer-sharp/src/gatsby-node.js
+++ b/packages/gatsby-transformer-sharp/src/gatsby-node.js
@@ -24,7 +24,7 @@ exports.sourceNodes = ({ actions }) => {
 
   if (createTypes) {
     createTypes(`
-      type ImageSharp implements Node @infer {
+      type ImageSharp implements Node @infer @childOf(types: ["File"]) {
         id: ID!
       }
     `)


packages/gatsby-transformer-sharp/src/gatsby-node.js
```diff
@@ -24,7 +24,7 @@ exports.sourceNodes = ({ actions }) => {
 
   if (createTypes) {
     createTypes(`
-      type ImageSharp implements Node @infer {
+      type ImageSharp implements Node @infer @childOf(types: ["File"]) {
```
pieh: What will happen when version of gatsby doesn't handle those newer directives? Will this be silent no-op or will it complain about unknown directives?

stefanprobst: Only directives handled here are turned into extensions, so this is a noop in older versions:
https://github.com/gatsbyjs/gatsby/blob/42317c16ea1f86b1b3fe04b91cc25e16b79f757b/packages/gatsby/src/schema/schema.js#L322-L355

PR #16066, opened 2019-07-24
fix(gatsby-cli): don't crash when creating project from local starter

marcos-abreu:
## Description

Fix issue when attempting to create a new project using a local starter.

## Related Issues

Fixes #16065

Issue #16065, opened 2019-07-24
Issue when attempting to create a new project from a local starter


marcos-abreu:
## Description

Gatsby CLI fails when creating a project from a local starter.

I've encountered this issue when trying to create a starter that cannot be hosted on github.com.

Since `gatsby` don't support non-public repository hosts (yet, I hope to create a PR that will help on this later), creating a project from a local starter is currently the only way, but I when I tested I found this issue.

### Steps to reproduce

1. clone the default starter at https://github.com/gatsbyjs/gatsby-starter-default.git

2. attempt to create a project based on the local starter: `gatsby new my-project /path/to/local/gatsby-starter-default`

### Expected result

A new project is created using the local starter as a template.

### Actual result

No project was created and the following error is thrown:

```
  TypeError: Cannot read property 'shortcut' of undefined

  - init-starter.js:357
    [lib]/[gatsby]/[gatsby-cli]/lib/init-starter.js:357:31
```

### Environment

```
  System:
    OS: Linux 4.4 Ubuntu 16.04.3 LTS (Xenial Xerus)
    CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
    Shell: 4.3.48 - /bin/bash
  Binaries:
    Node: 8.11.1 - ~/.nvm/versions/node/v8.11.1/bin/node
    Yarn: 1.6.0 - ~/.nvm/versions/node/v8.11.1/bin/yarn
    npm: 5.6.0 - ~/.nvm/versions/node/v8.11.1/bin/npm
```

marcos-abreu:
I've looked into the issue and I will be submitting a PR soon.

PR #16121, opened 2019-07-26
docs(starters): Add documentation for using local starters

marcos-abreu:
## Description

Document creating a project using local starters

## Related Issues

Related to #16065 
Related to #16066


marcos-abreu:
@gatsbyjs/learning - I know you probably have many priorities, just was wondering if there are any updates on this PR?

docs/docs/starters.md
```diff
@@ -44,6 +44,20 @@ If you don't specify a custom starter, your site will be created from the [defau
...
 
+### Using a local starter
+
+Another option is to supply a path (relative or absolute) to a local folder containing a starter:
+
+```shell
+gatsby new [SITE_DIRECTORY] [STARTER/LOCAL/PATH]
+```
+
+Here is an example assuming a local starter exists on the path `./my-local-starter`:
```
marcysutton: I think showing it installing from a different folder would motivate why someone might want to do this. Because this seems kinda unnecessary to me without more explanation

sidharthachatterjee: Good point, fixed

PR #16124, opened 2019-07-26
docs: fix wrong link to create-content-digest.js

georgiee:
## Description
Fixed a link into the gatsby source of file `create-content-digest.js` as I was reading this article just now.

## Discussion/Idea
Maybe it would be better to use permalinks in the docs. Those would include the commit hash (key y on github)? Everyone clicking would always get a working link regardless of the file being moved or deleted. That way I would get at least the idea of the file when it was linked. If I'm truly interested in finding the actual implementation, I think I would be able to find it by using git blame. What do you think ? When changing the file I can still update the permalink to always point to the current version but missing it is not yielding a 404 instantly like now.

The actual permalink (at the time of writing this) would be: https://github.com/gatsbyjs/gatsby/blob/42317c16ea1f86b1b3fe04b91cc25e16b79f757b/packages/gatsby-core-utils/src/create-content-digest.js

I'm not sure how this would look like in the bigger image, if internal source links are pointing to different states of the implementation. That's why I kept the master based link in this PR for now.

### DIFF
--- a/docs/docs/node-interface.md
+++ b/docs/docs/node-interface.md
@@ -31,7 +31,7 @@ Reserved for plugins who wish to extend other nodes.
 
 Digest "Hash" (for example `md5sum`) of the content of this node.
 
-The digest should be unique to the content of this node since it's used for caching. If the content changes, this digest should also change. There's a helper function called [createContentDigest](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/utils/create-content-digest.js) to create an `md5` digest.
+The digest should be unique to the content of this node since it's used for caching. If the content changes, this digest should also change. There's a helper function called [createContentDigest](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-core-utils/src/create-content-digest.js) to create an `md5` digest.
 
 ### `mediaType`
 


DSchau:
This is not an incredibly easy problem to solve.

I like your thinking here -- and we certainly don't want broken links. We want to find a solution that:

1. Is as correct as possible
1. Does not have a ton of churn (e.g. it breaks every other week, and we are _reactive_ in fixing)
1. Provides helpful info for the reader

Towards achieving each of these as best as we can -- I think the best path is to include the commit hash of current master (e.g. in this example would be https://github.com/gatsbyjs/gatsby/blob/42317c16ea1f86b1b3fe04b91cc25e16b79f757b/packages/gatsby-core-utils/src/create-content-digest.js as you noted).

Why?

It strikes the balance between 2 and 3, while not being _too terribly inaccurate _ as far as 1 goes. This doesn't mean this is the correct thing to do in every case -- but in this case when it's provided as a helpful resource for the reader to learn more, it's OK to link to a commit hash which may change slightly from the hash and master.

georgiee:
Thank you for your response. Looks like we are on the same page 👌 Such a doc link would always show the gist of a mentioned file — which is so much better than any expired link. 

I still have a thought on this: Once the link is updated with a permalink, no one will come back to the file link in the future as it will never show any obvious signs of being outdated anymore. Is this only a constructed problem because future dramatic changes in the codebase would require the docs to be updated or even rewritten anyway? 

Thanks for taking your time checking on this.

wardpeet:
@DSchau we might want to go for something like `gatsby.dev/utils/create-content-digest` and manage the redirection ourselves?

I suggest we should move forward with a permalink https://github.com/gatsbyjs/gatsby/blob/82a2a07359f03c683d64903149bd5acfa57cce71/packages/gatsby-core-utils/src/create-content-digest.js. I don't think it will change anytime soon.

PR #16132, opened 2019-07-26
feat(www): add aria attributes to pagination

NickyMeuleman:
Added aria attributes to pagination.

Adding aria attributes can do more harm than good if not done correctly, does this move the needle in the right direction?
Used these resources:
https://a11y-style-guide.com/style-guide/section-navigation.html
https://tink.uk/using-the-aria-current-attribute/
https://www.a11ymatters.com/pattern/pagination/

### DIFF
--- a/www/src/components/pagination/index.js
+++ b/www/src/components/pagination/index.js
@@ -94,7 +94,7 @@ class Pagination extends React.Component {
                 value={`${i === 0 ? `` : i + 1}`}
                 key={`pagination-number${i + 1}`}
                 aria-label={`Goto Page ${i + 1}`}
-                aria-current={currentPage === i + 1 && `page`}
+                aria-current={currentPage === i + 1}
               >
                 {i + 1}
               </option>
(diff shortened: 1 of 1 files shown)


www/src/components/pagination/index.js
```diff
@@ -92,6 +93,8 @@ class Pagination extends React.Component {
               <option
                 value={`${i === 0 ? `` : i + 1}`}
                 key={`pagination-number${i + 1}`}
+                aria-label={`Goto Page ${i + 1}`}
+                aria-current={currentPage === i + 1 && `page`}
```
wardpeet: Why do you need the `` && `page` ``?
```suggestion
                aria-current={currentPage === i + 1}
```

NickyMeuleman: Wondered that myself, since I didn't notice a difference when using the NVDA screenreader. Took the recommendation from https://tink.uk/using-the-aria-current-attribute/
Search for 'page' there. It's under the header: The aria-current attribute.

PS: Yay, nen anderen Belg, hoera, hoera!

sidharthachatterjee: We don't need the `&& `page``

@NickyMeuleman Can you apply the suggestion? I'll go ahead and merge 🙂

NickyMeuleman: done :)

Issue #15816, opened 2019-07-17
[docs] Improve documentation around "mapping"


LekoArts:
## Summary

While "naively" searching for a solution to this [Spectrum thread](https://spectrum.chat/gatsby-js/general/connecting-gatsby-image-results-to-items-in-a-list~41586191-994e-4d83-9ab8-c768a8762c0e) via Google and the docs search itself, I didn't find the bits I needed to find. I'm familiar with the mapping via [gatsby-config](https://www.gatsbyjs.org/docs/gatsby-config/#mapping-node-types) but couldn't find the new [schema customization](https://www.gatsbyjs.org/docs/schema-customization/#foreign-key-fields) way of doing things by myself.

### Possible reasons

- The documentation on gatsby-config doesn't mention that there's another way of doing this now
- The schema customization doc doesn't mention the word "mapping" at all (which is "necessary" for Google/Local search IMO)

## Proposal

- Add a paragraph to the gatsby-config doc to mention that [this](https://www.gatsbyjs.org/docs/schema-customization/#foreign-key-fields) is now the new way of doing this
- Add a paragraph to the schema customization doc saying something like "previously you used to use the mapping feature in gatsby-config to do ...."

---

**On a related note:** Could we add recipes for things like that or schema customization stuff in general?

Issue #16118, opened 2019-07-26
Mapping multiple nodes to a single source node fails


nathobson:
## Description

I can successfully map a single node to another node [as described in the docs](https://www.gatsbyjs.org/docs/gatsby-config/#mapping-node-types). However, I can't seem to find a way to map multiple nodes to a single other node.

### Steps to reproduce

Add something like this to `gatsby-config`

```
mapping: {
    'WordPressAcf_call_to_action.buttons': ['wordpress__wp_projects.wordpress_id', 'wordpress__PAGE.wordpress_id']
  }
```

The project I'm working on is private so don't have a repo. Can try to abstract something out if that would be needed.

### Expected result

I would expect to be able to pass an array of nodes, which would each then be mapped to the single other node. If multiple keys with the same value are passed, it will not fail but only the last value will be mapped (which makes sense).

### Actual result

This fails on `gatsby develop` with the following error message:

`UNHANDLED REJECTION typeMapping[selector].split is not a function or its return value is not iterable`

Using a single key:value pair works as expected:

```
mapping: {
    'WordPressAcf_call_to_action.buttons': `wordpress__wp_projects.wordpress_id`
  }
```

### Environment

```
  System:
    OS: macOS 10.14.3
    CPU: (12) x64 Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 10.15.3 - /usr/local/bin/node
    Yarn: 1.15.2 - /usr/local/bin/yarn
    npm: 6.4.1 - /usr/local/bin/npm
  Languages:
    Python: 2.7.10 - /usr/bin/python
  Browsers:
    Chrome: 75.0.3770.142
    Firefox: 67.0.2
    Safari: 12.0.3
  npmPackages:
    gatsby: ^2.8.6 => 2.13.28 
    gatsby-image: ^2.0.31 => 2.2.6 
    gatsby-plugin-manifest: ^2.0.22 => 2.2.3 
    gatsby-plugin-offline: ^2.0.24 => 2.2.4 
    gatsby-plugin-react-helmet: ^3.0.8 => 3.1.2 
    gatsby-plugin-react-svg: ^2.1.1 => 2.1.1 
    gatsby-plugin-sharp: ^2.0.25 => 2.2.8 
    gatsby-plugin-styled-components: ^3.1.2 => 3.1.2 
[...]

Issue #15920, opened 2019-07-19
I need help creating a relationship between two Gatsby sources


mrbjjackson:
Hi everyone,

So I have two sources loaded into my GraphQL...

Firstly, a list of _Artist_ nodes, coming from a _Contentful_ content model.

This content type contains the _Artist name_, some information about them, and an ID.

Secondly a list of _Track_ nodes automatically imported from a _Spotify playlist_. Each track has a _Track Name_, an _Artist name_ and an ID.

I would like to create a relationship between the _Spotify Tracks_ to the _Contentful Artists_ based on the _Artist name_, when it matches between the two so I can have write some information about an artist on Contentful and then on my website, easily display this info along with a list of their tracks below.

From my reading it looks like I have to use `onCreateNode`. And if the node type is Contentful Artist, use `createNodeField` and do something like `LinkedTracks___NODE = <Spotify Track ID>`

The trouble I'm having is, from within `onCreateNode`, I don't know how to access the information in the Spotify Tracks nodes to work out which tracks have the same Artist as the current Contentful Artist.

I hope that makes sense.

Any help would be much appreciated. Thanks, Ben

cluk3:
https://www.gatsbyjs.org/docs/gatsby-config/#mapping-node-types
If I understood well that's exactly what you need, right?

stefanprobst:
You could also use the schema customization APIs to [create a foreign-key relation](https://www.gatsbyjs.org/docs/schema-customization/#foreign-key-fields).

mrbjjackson:
Thanks very much for the speedy replies. I'll try them both out.
On first look over it looks like the Mapping solution may not work because the Artist field is not an ID in either the Contentful or Spotify sources. It's just a general field. I'll investigate further.
The foreign-key relation looks very promising.

mrbjjackson:
So I did manage to get this working using the info linked by @stefanprobst .
The code is pretty specific to my needs so may not make sense out of context but I thought I would post it in case it helps anyone else.
```
exports.createSchemaCustomization = ({ actions, schema }) => {
  const { createTypes } = actions
  const typeDefs = [
    `
    #the following three help put the linked artist node into the track node
    type SpotifyTrack implements Node {
      track: Track
    }
    
    type Track{
        artists: [Artist]
    }

    type Artist implements Node {
      artist: [ContentfulArtist] @link(by: "artist", from: "name")
    }

    `
  ]
  createTypes(typeDefs)
}
```

[...]

PR #16152, opened 2019-07-27
docs: Link mapping to @link

stefanprobst:
Link to creating foreign-key relations with the `@link` directive from mappings, and vice versa.

Fixes:  #15816
Related: #16118, #15920

### DIFF
--- a/docs/docs/schema-customization.md
+++ b/docs/docs/schema-customization.md
@@ -319,6 +319,10 @@ Gatsby's automatic type inference has one trick up its sleeve: for every field
 that ends in `___NODE` it will interpret the field value as an `id` and create a
 foreign-key relation.
 
+> Note: Before the introduction of the Schema Customization APIs in Gatsby v2.2,
+> there were two mechanisms to create links between node types: with the `___NODE`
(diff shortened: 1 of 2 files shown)


docs/docs/schema-customization.md
```diff
@@ -319,6 +319,10 @@ Gatsby's automatic type inference has one trick up its sleeve: for every field
 that ends in `___NODE` it will interpret the field value as an `id` and create a
 foreign-key relation.
 
+> Note: Before the introduction of the Schema Customization APIs in Gatsby v2.2,
```
m-allanson: I'm unsure what the takeaway is here - as a reader should this inform me that using mappings or `___NODE` are equivalent, and that in both cases I can now use schema customisation instead?

stefanprobst: yes -- previously a plugin author would use the `___NODE` naming convention to create links between nodes when *creating the nodes* (i.e. used by source/transformer plugins), and a user would use the `mappings` config key. both users and plugin authors can now use the `@link` extension.

m-allanson: That's great - I added one more line at the end that matches your comment here

Issue #15761, opened 2019-07-15
[gatsby-source-drupal] fetching data fails when there are 0 images


pieh:
## Description

When there are 0 images in drupal site, `gatsby-source-drupal` crashes with:
![D_ilNTNX4Ag4c-0](https://user-images.githubusercontent.com/419821/61247048-99bc7d80-a750-11e9-931a-4bbc83ac3d23.png)

This happens in this part of code - https://github.com/gatsbyjs/gatsby/blob/98511d678f58fc63387ebfa8aa898c7a9761c1e4/packages/gatsby-source-drupal/src/gatsby-node.js#L139-L154

plugin should determine array of File nodes first:
```
const fileNodes = [...nodes.values()].filter(isFileNode)
```

and only use `asyncPool` (and reporter for logging) when there are any images.

harshil1712:
@pieh I would like to take this, but I would need your help :sweat_smile:

jdiglesias:
Thanks Kayla, your fix works for me. In order to test it, you'll have to include a mock fixture that returns no file nodes (i.e. [this file](https://github.com/gatsbyjs/gatsby/blob/98511d678f58fc63387ebfa8aa898c7a9761c1e4/packages/gatsby-source-drupal/src/__tests__/fixtures/file.json) but with an empty array as the data field) and run a test using that fixture.

PR #16157, opened 2019-07-27
fix(gatsby-source-drupal): don't show "file download" activity if there are no files to download

skippednote:
If the nodes don't have any files attached like images Gatsby would throw the following error
```
"gatsby-source-drupal" threw an error while running the sourceNodes lifecycle:
Parameter `array` must have at least one item
```

To prevent false warnings we are checking if there are any images to download to begin with.


skippednote:
Fixes #15761

pieh:
Hey @skippednote! 

Similar pull request was just merged ( https://github.com/gatsbyjs/gatsby/pull/16160 ), but I think we could still merge this in with small change/improvement. Right now `downloadingFilesActivity.start()` and `downloadingFilesActivity.end()` are outside of the `if (fileNodes.length)` conditional block - if you could move those inside, we could merge this PR as well!

packages/gatsby-source-drupal/src/gatsby-node.js
```diff
@@ -140,16 +140,15 @@ exports.sourceNodes = async (
   downloadingFilesActivity.start()
```
pieh: Please move this inside `if (fileNodes.length)` conditional code path

packages/gatsby-source-drupal/src/gatsby-node.js
```diff
@@ -140,16 +140,15 @@ exports.sourceNodes = async (
...
       await downloadFile(
         { node, store, cache, createNode, createNodeId },
         pluginOptions
       )
-    }
-  )
+    })
+  }
 
   downloadingFilesActivity.end()
```
pieh: Please move this inside `if (fileNodes.length)` conditional code path

skippednote: @pieh Thank you for catching that. I've made those changes.

PR #16159, opened 2019-07-27
chore(docs): Fix snippet filenames & cleanup in programmatically creating pages.md

jameygleason:
## Description

Updated docs for gatsby-plugin-mdx docs for programmatically creating pages. Updated my site this morning. This is cool!

Tidying up title attributes on the code blocks. Remove the nested "body" in the query. It's no longer within "code". Wrote createPages with async/await.

PR #16163, opened 2019-07-28
updating graphql query in documentation

andihaki:
updating graphql query at this page : [https://www.gatsbyjs.org/docs/mdx/programmatically-creating-pages/#make-a-template-for-your-posts](https://www.gatsbyjs.org/docs/mdx/programmatically-creating-pages/#make-a-template-for-your-posts)
i'm using :

> "gatsby": "^2.13.40",
> "gatsby-plugin-mdx": "^1.0.17",
> "gatsby-source-filesystem": "^2.1.6",

there's no `code.body`, but `rawBody` instead :
```
mdx(id: { eq: $id }) {
      id
      frontmatter {
        title
      }
      # code {
      # body
      # }
      rawBody
    }
```
screenshot :
image

ps: please check, there's extra apostrophe in styled-component page.
i think wrong:

> Here's an example where the class name `container` is added to the DOM along with the Styled Components' dynamically-created class names:

i think right:

> Here's an example where the class name `container` is added to the DOM along with the Styled Components dynamically-created class names:


lannonbr:
I'm going to close this PR as there is already a PR opened (#16159) that resolves this issue in the mdx docs.

As well, I think the apostrophe is fine here given it is used as possessive (the dynamic created classnames are of Styled Components)

PR #16166, opened 2019-07-28
Gatsby theme blog core

ChristopherBiscardi:
cc/ @NickyMeuleman 

# Create gatsby-theme-blog-core

This PR creates a parent theme called gatsby-theme-blog-core which is then used as the parent of gatsby-theme-blog.

## gatsby-theme-blog-core

The core theme is a theme meant to encourage re-use of the `BlogPost` core type to build an ecosystem of compatible blog themes on the same base. The `BlogPost` interface itself enables people to create backing implementations from any node type in the Gatsby ecosystem, leading us one step further down the path to supporting any CMS. The core theme implements this approach for `Mdx` nodes sourced from the included filesystem source. More info on the ability to source from multiple node types [here](https://www.christopherbiscardi.com/post/sourcing-data-from-multiple-locations-in-gatsby).

The theme:
- contains all data structures, sources, and processing
- Contains page creation (`createPage` calls) with templates that render `JSON.stringify(props.data, null, 2)`

## gatsby-theme-blog

`gatsby-theme-blog` becomes effectively a collection of components used to render a Gatsby site. This allows anyone to do the same by shadowing two components from the core theme (`src/gatsby-theme-blog-core/components/posts.js` and `src/gatsby-theme-blog-core/components/post.js`).

- removed `gatsby-node.js` as it no longer does anything
- shadowed gatsby-theme-blog-core components to implement theme-ui, etc based layouts


NickyMeuleman:
Thanks for tagging me to this PR!

I read your comment on **pagination** in the other thread.
How do you feel about it being an option to include (default is no pagination)?

I did this in my themejam submission, the relevant section in `gatsby-node`:
https://github.com/NickyMeuleman/gatsby-theme-nicky-blog/blob/d37c8f35c2dbde1365edcce001e926fc3005ccdb/theme/gatsby-node.js#L188-L224

---

Also on that site, I'm transforming the **tags** array in the frontmatter to an array of objects in graphQL that have this form `{ name: String, slug: String }`. That worked surprisingly well, ability to query both fields.
Unfortunately, for the page that groups all tags. I couldn't figure out how to query a group of them AND access both fields in graphQL: 
```
query MyQuery {
  allBlogPost {
    group(field: tags) {
      fieldValue {
        name
        slug
      }
    }
  }
}
```
errors out, as the fieldValue alone returns a string: `"object Object"`

themes/gatsby-theme-blog-core/gatsby-node.js
```diff
@@ -209,3 +130,59 @@ exports.onCreateNode = ({ node, actions, getNode, createNodeId }) => {
     createParentChildLink({ parent: fileNode, child: node })
   }
 }
+
+// These templates are simply data-fetching wrappers that import components
+const PostTemplate = require.resolve(`./src/templates/post`)
+const PostsTemplate = require.resolve(`./src/templates/posts`)
+
```
KyleAMathews: Perhaps we should we rename these to something more descriptive? E.g. post-query.js I find it odd currently when searching through files as both post.js come up and I have to check to distinguish them.

ChristopherBiscardi: yeah, I thought about this for sure when I was translating it to the -core theme. `posts` and `post` can change in -core, but it would be a breaking change for shadowing in `gatsby-theme-blog`.

ChristopherBiscardi:
@NickyMeuleman 

> Thanks for tagging me to this PR!

🎉 

> I read your comment on **pagination** in the other thread.
> How do you feel about it being an option to include (default is no pagination)?

I don't like including it by default. I think we should include it as an extension to the data model in some way. I don't know exactly what that looks like yet, but it means we won't have to support n* features in -core if we can pull off "data extension" like this.

(personally I don't think pagination is useful, especially for the amount of content the "average" blogger puts out and the purpose of having a page listing. Why make it harder for people to CMD+F through your list of posts? ...but this is a different discussion than the data one above).

> Unfortunately, for the page that groups all tags. I couldn't figure out how to query a group of them AND access both fields in graphQL:

I generally avoid nested fields unless it's an additional type. I think in -core we'll solve this problem using a `Tag` extension that can be used across other interfaces (`BlogPost`, `Note`, etc). Then you'd be able to do this if you wanted a list of all tags and such:

```js
query MyQuery {
  allTag {
    name
    slug
  }
}
```

ChristopherBiscardi:
Implemented `BlogPost` as an Interface in the last commit. Nothing had to change in gatsby-theme-blog

KyleAMathews:
👍 to post-query.js

themes/gatsby-theme-blog-core/gatsby-node.js
```diff
@@ -53,11 +42,24 @@ const mdxResolverPassthrough = fieldName => async (
...
+  createTypes(`interface BlogPost @nodeInterface {
+      id: ID!
+      title: String!
+      body: String!
+      slug: String!
+      date: Date! @dateformat(formatString: "DD-MM-YYYY")
+      tags: [String]!
+      keywords: [String]!
+      excerpt: String!
+      body: String!
```
stefanprobst: Duplicated `body` field
```suggestion
```

themes/gatsby-theme-blog-core/gatsby-node.js
```diff
@@ -53,11 +42,24 @@ const mdxResolverPassthrough = fieldName => async (
...
-exports.sourceNodes = ({ actions, schema }) => {
+
+exports.createSchemaCustomization = ({ actions, schema }) => {
   const { createTypes } = actions
+  createTypes(`interface BlogPost @nodeInterface {
+      id: ID!
+      title: String!
+      body: String!
+      slug: String!
+      date: Date! @dateformat(formatString: "DD-MM-YYYY")
```
stefanprobst:
```suggestion
      date: Date! @dateformat
```
Or, if we want a default format string, we should have it on `MdxBlogPost` as well

themes/gatsby-theme-blog-core/package.json
```diff
@@ -0,0 +1,31 @@
...
+  },
+  "dependencies": {
+    "gatsby-plugin-mdx": "^1.0.16",
+    "gatsby-plugin-sharp": "^2.2.9",
+    "gatsby-remark-copy-linked-files": "^2.1.3",
+    "gatsby-remark-images": "^3.1.7",
+    "gatsby-remark-numbered-footnotes": "^1.0.0",
+    "gatsby-remark-smartypants": "^2.1.2",
+    "gatsby-source-filesystem": "^2.1.6",
+    "gatsby-transformer-sharp": "^2.2.4",
```
jxnblk: I see the benefit in having the sharp image stuff baked into the core theme, but also wonder if we should strive to make it even less opinionated by not having sharp as a dependency – and maybe fewer of the Remark plugins. In the future, I suspect we'll be able to handle some of this with components, which seems easier to change from the end-user's perspective, but we don't have those yet, so I'm fine keeping this in. Just wanted to pose the question

themes/gatsby-starter-blog-theme-core/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+## 🚀 Quick start
+
+1.  **Create a Gatsby site.**
+
+    Use the Gatsby CLI to create a new site, specifying the blog theme starter.
+
+    ```sh
+    # create a new Gatsby site using the blog theme starter
+    gatsby new my-themed-blog https://github.com/gatsbyjs/gatsby-starter-blog-theme
```
jxnblk: Should this be referencing `gatsby-starter-blog-theme-core` in this README?

```suggestion
    gatsby new my-themed-blog https://github.com/gatsbyjs/gatsby-starter-blog-theme-core
```

themes/gatsby-starter-blog-theme-core/src/gatsby-theme-blog/gatsby-plugin-theme-ui/colors.js
```diff
@@ -0,0 +1,25 @@
+import merge from "deepmerge"
```
jxnblk: I might be misunderstanding the purpose of `gatsby-starter-blog-theme-core`, but is this intended for latent shadowing? I'd imaging that the starter for blog-core would be fairly minimal and decoupled from `gatsby-theme-blog`

themes/gatsby-theme-blog/gatsby-config.js
```diff
@@ -2,62 +2,15 @@ module.exports = options => {
   const { mdx = true } = options
 
   return {
-    siteMetadata: {
```
jxnblk: I love how much this simplifies this theme

themes/gatsby-starter-blog-theme-core/src/gatsby-theme-blog/gatsby-plugin-theme-ui/colors.js
ChristopherBiscardi: I think I copied this starter and left the directory here. will delete.

themes/gatsby-theme-blog-core/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "remark-slug": "^5.1.2"
+  },
+  "devDependencies": {
+    "@mdx-js/mdx": "^1.1.0",
```
johno: Just took this for a spin and it errored out here due to the lack of `@mdx-js/mdx`. I _think_ we can include this dependency because it won't be loaded on the client or cause duplicate contexts to stomp eachother like `@mdx-js/react` would.

(after manually adding this dep it's all working out great tho!)

ChristopherBiscardi: yeah, I guess you could always upgrade by forcing resolutions if you needed a bugfix. Same for upgrading child themes and such.

themes/gatsby-theme-blog-core/package.json
ChristopherBiscardi: Since sharp is the 3rd, 4th, and 5th most downloaded plugins I'm going to keep sharp in for now. I think that moving it somewhere else is something we should come back to in the future.

image

ChristopherBiscardi:
Rebased and made all suggested changes. This should be good to go.

themes/gatsby-theme-blog-core/gatsby-node.js
```diff
@@ -190,22 +122,80 @@ exports.onCreateNode = ({ node, actions, getNode, createNodeId }) => {
...
+
+  const result = await graphql(`
+    {
+      allBlogPost(sort: { fields: [date, title], order: DESC }, limit: 1000) {
+        edges {
+          node {
+            id
+            excerpt
+            slug
+            title
```
KyleAMathews: title, date, excerpt aren't used anymore right?

themes/gatsby-starter-blog-theme-core/README.md
```diff
@@ -0,0 +1,89 @@
...
+    Your site is now running at `http://localhost:8000`!
+
+    To get started, check out the guide to [using the Gatsby blog theme starter](http://gatsbyjs.org/docs/themes/using-a-gatsby-theme), or the longer, [more detailed tutorial](http://gatsbyjs.org/tutorial/using-a-theme).
+
+## 🧐 What's inside?
+
+Here are the top-level files and directories you'll see in a site created using the blog theme starter:
+
+```
+gatsby-starter-blog-theme
```
johno:
```suggestion
gatsby-starter-blog-theme-core
```

PR #16170, opened 2019-07-28
feat(gatsby-remark-prismjs): Add support for Prism command-line plugin

elken:
## Description

Adds prompts (customised with css) to shell output using command-line plug-in. Output can be changed via outputLines directive. Disabled by default, enabled by specifying user & host in settings.


wardpeet:
Any chance you can show us a screenshot of what this does? :smile: I'm not super familiar with prismjs.

elken:
Of course! I can also attach to the README if you think it'd be relevant.

Output

packages/gatsby-remark-prismjs/README.md
```diff
@@ -68,6 +68,12 @@ plugins: [
                 },
               },
             ],
+            // Customize the prompt used in shell output
+            // Leave empty to hide prompts
+            prompt: {
```
wardpeet: I wonder if we should move this to the snippet itself instead. We do the same with line numbers and other plugins.
```
```bash prompt=root@localhost
```

```
```bash prompt={user:"root",host:"localhost"}
```

What do you think?

elken: I can see the idea behind it but I made this decision because

a) It reduces clutter in the snippet specification (esp. with other plugins)
b) As far as I can see it, it's something you either want on or off

But I'm happy to amend if you disagree with these points

elken: I'm working on adding the best of both worlds. So new logic will be:

- `prompt.global` setting defaulting to false (which will add prompts to every line)
- Including `{outputLines: <range>}` which will add prompts to any line _not_ in the range

elken:
As reflected in the README, default behaviour is off, can be changed with `prompt.global` in settings or by specifying `outputLines`. I've also made `outputLines` and `numberLines` work together in options (should still be backwards compatible but I guess the tests will show that).

packages/gatsby-remark-prismjs/package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "gatsby-remark-prismjs",
+  "name": "@elken/gatsby-remark-prismjs",
```
sidharthachatterjee:
```suggestion
  "name": "gatsby-remark-prismjs",
```

packages/gatsby-remark-prismjs/package.json
```diff
@@ -1,7 +1,7 @@
...
   "description": "Adds syntax highlighting to code blocks at build time using PrismJS",
-  "version": "3.3.3",
+  "version": "0.3.3",
```
sidharthachatterjee:
```suggestion
  "version": "3.3.3",
```

We don't need to change versions in a PR since versioning is handled by lerna

packages/gatsby-remark-prismjs/package.json
elken: Apologies, leftover from testing.

elken:
Few testing tweaks slipped in, apologies. Both resolved.

alexfornuto:
Thanks for this addition! While I'm not sure how to implement, is there any to expand this so that we could use it for other shells besides bash? For example, I'd love to have:

````
```redis{promptUser: redis}
redis> KEYS *a*
$17
englash bigkahuna
```
````
with `redis>` as the prompt. 

I'm not asking for this specific example, just the ability to make additional prompts that work for other highlighters besides bash.

elken:
The plugin only works for command line, hence the name. Feel free to write your own Prism plugin.

Also create an issue rather than comment on closed PR's.

PR #16172, opened 2019-07-28
feat(gatsby-transformer-remark): subscribe to mimeTypes in type definitions

pieh:
Built on top of features added in https://github.com/gatsbyjs/gatsby/pull/15715

This should be no-op on older versions of Gatsby that don't support `@childOf` directive and not cause any unknown API situation.

/cc @stefanprobst Let me know if this is correct - it does seem to work and create `childMarkdownRemark` fields (when parent node types have `@mimeTypes` directive/extension)


wardpeet:
Published in `gatsby-transformer-remark@2.6.10`

PR #16178, opened 2019-07-29
chore(themes): move themes into starters and packages

sidharthachatterjee:
## Why
Theme packages should now be publishable (is that even a word) by lerna

## What 
This pull request 

- moves theme starters into the starters directory
- moves theme packages into the package directory
- updates fields in package.json accordingly
- cleans up renovate config
- cleans up lerna config
- cleans up CI config
- excludes theme starter README from markdown magic


sidharthachatterjee:
@wardpeet Shouldn’t need any testing per se but let’s wait on a go ahead from the themes team? 

Cc @johno @ChristopherBiscardi

sidharthachatterjee:
So turns out adding starters to the workspace breaks in all sorts of ridiculous ways 😂 

https://app.circleci.com/jobs/github/gatsbyjs/gatsby/182460 fails with `sh: 1: gatsby: not found` because dependencies are hoisted to the root as a result of adding this to the workspace and scripts/validate-starters.sh deletes all of `node_modules` and then runs `yarn` which for some reason results in no `gatsby` in `.bin`

https://app.circleci.com/jobs/github/gatsbyjs/gatsby/182462 fails because of some snapshot mismatches where it expects `"<PROJECT_ROOT>/packages/babel-preset-gatsby/node_modules/@babel/runtime"` but gets `"<PROJECT_ROOT>/node_modules/@babel/runtime"` which I guess makes sense because of hoisting 🤔 

Another snapshot test fails (it transforms HTML img tags). This one is strange.

sidharthachatterjee:
Closing this in favour of a new one because holy hell look at those conflicts

ChristopherBiscardi:
leaving a post-close comment that this approach is good with me and we'll work out whatever we need to with `gatsby-dev` later.

PR #16190, opened 2019-07-29
chore(e2e-tests): Only record cypress runs on CI

sidharthachatterjee:
Currently end to end tests don't work locally when run via `yarn test` because cypress complains of being called with `--record` but without a secret and project ID 

Since cypress dashboard secrets and project IDs are in Circle CI config and we shouldn't record local cypress runs IMO anyway, this PR checks you're on CI and only then passes in `--record` when running cypress


sidharthachatterjee:
Yeah. We should definitely clean up this once cypress supports passing in the record options via the config json or an env variable @pieh

PR #16205, opened 2019-07-30
fix(gatsby): enable babel.config.js once again

### LINKED ISSUE #16134: babel.config.js not picked up since 2.13.30 (regression)
## Description

Since updating Gatsby to `2.13.30` (or newer), I can no longer use my own custom `babel.config.js`. `babel.config.js` is the new, better way to configure Babel. See Babel [config files docs](https://babeljs.io/docs/en/config-files).

### Steps to reproduce

I've created a minimal repro here: https://github.com/ahfarmer/gatsby-babel-config-issue
To create the repro I just ran `gatsby new`, then added `babel.config.js` and deleted some unnecessary files.

Here's how you can reproduce easily:
```
git clone https://github.com/ahfarmer/gatsby-babel-config-issue
cd gatsby-babel-config-issue
yarn install
gatsby develop
```

Note that the console.log line that I have placed in `babel.config.js` does NOT print out. There are other ways you could show that `babel.config.js` is not being considered, but I think this is the easiest one.

### Expected result

We should see this line printed to the console: `"👍 babel.config.js is being used"`

### Actual result

`babel.config.js` is not used for some reason. The code within does not run and the log statement doesn't print out.

### Workaround

```
yarn add gatsby@2.13.27
yarn develop
```

With this older version of Gatsby, you will see that `babel.config.js` is being used.

### Related Issues

Basically the same issue reported here: https://github.com/birkir/gatsby-source-graphql-universal/issues/3

I opened this issue to show that it is not an issue with `gatsby-source-graphql-universal`, which I am not using.

They are reporting that this change caused the issue: https://github.com/gatsbyjs/gatsby/pull/15875

### Environment

```
  System:
    OS: macOS 10.14.5
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 10.15.3 - /usr/local/bin/node
  Languages:
    Python: 3.6.8 - /anaconda3/bin/python
  Browsers:
    Chrome: 75.0.3770.142
    Firefox: 68.0.1
    Safari: 12.1.1
  npmPackages:
    gatsby: ^2.13.39 => 2.13.39 
    gatsby-image: ^2.2.6 => 2.2.6 
  npmGlobalPackages:
    gatsby-cli: 2.7.17
```

wardpeet:
## Description
Bring back support for babel.config.js which we should disable in V3 as it's a breaking change for v2. It's not encouraged to use it the way people are using it for gatsby.

## Related Issues
Fixes #16134

### DIFF
--- a/packages/gatsby/src/utils/__tests__/__snapshots__/webpack-utils.js.snap
+++ b/packages/gatsby/src/utils/__tests__/__snapshots__/webpack-utils.js.snap
@@ -33,7 +33,7 @@ Object {
       "loader": "<PROJECT_ROOT>/packages/gatsby/src/utils/babel-loader.js",
       "options": Object {
         "compact": true,
-        "configFile": false,
+        "configFile": true,
         "stage": "develop",
       },
     },
--- a/packages/gatsby/src/utils/webpack-utils.js
+++ b/packages/gatsby/src/utils/webpack-utils.js
@@ -326,7 +326,6 @@ module.exports = async ({
         use: [
           loaders.js({
             ...options,
-            // TODO(v3): set configFile to false
             configFile: true,
             compact: true,
           }),
--- a/packages/gatsby/src/utils/webpack-utils.js
+++ b/packages/gatsby/src/utils/webpack-utils.js
@@ -326,7 +326,8 @@ module.exports = async ({
         use: [
           loaders.js({
             ...options,
-            configFile: false,
+            // TODO(v3): set configFile to false
+            configFile: true,
             compact: true,
           }),
         ],


wardpeet:
Oh gatsby is always used as a "root" project and I don't see an immediate use case to programmatically add/remove config to babel for a gatsby project. Configs are mostly static and might change depending on build (production) or develop (development).

If you want to share configs between projects you can use a babel.config.js in your root. For themes, this might be a good solution 🤔.

https://babeljs.io/docs/en/configuration#what-s-your-use-case

Let me revert the comment as it might not be the best idea to disable it. Thanks @m-allanson for letting me re-iterating

ahfarmer:
I've got Gatsby as one of the projects in my repo. I share code between two Gatsby projects and a React Native project. 

I use babel.config.js to ensure that code I import has my project Babel rules applied - despite not descending from my Gatsby project root. 

Please reconsider. Frustrating to see Babel come out with the perfect solution, and then to hear that Gatsby will just turn it off. 

@wardpeet I've seen you arguing that we shouldn't use this feature but haven't seen any reasons why. Why not simply leave this feature available I'm all Gatsby versions? What benefit do you gain from disabling it?

m-allanson:
Hey @ahfarmer :wave: we reached the same conclusion as you in this PR - the ability to use `babel.config.js` has been added back in and is available in `gatsby@2.13.44`

wardpeet:
@ahfarmer https://github.com/gatsbyjs/gatsby/pull/16205#issuecomment-516320418

I've updated my view on babel.config.js

PR #16208, opened 2019-07-30
fix(gatsby): run onPrefetchPathname when prefetch is disabled

wardpeet:
## Description
When changing the page-manifest logic of gatsby we also broke the `onPrefetchPathname` api. It was not being called when prefetching was being disabled by plugins like guess-js.

Before page manifest:
https://github.com/gatsbyjs/gatsby/blob/8291044020254b04eb47b021101e79743794243a/packages/gatsby/cache-dir/loader.js#L237-L247

## Related Issues
No bugs yet but guess-js doesn't work anymore, it doesn't crash so that's good 👍


wardpeet:
Ok seems like we don't call `onPrefetchPathname` when we're on a slow connection. I've broke down the shouldPrefetch function in a connection check & prefetchDisabled boolean.

This should fix e2e tests

Issue #15724, opened 2019-07-15
Error loading a result for the page query in "/404.html". Query was not run and Page not found /404.html


sidharthachatterjee:
## Description

When using `gatsby develop` with a theme, the following error is printed in the terminal after a page is loaded in the browser:

```
Error loading a result for the page query in "/404.html". Query was not run and
Page not found /404.html
```

### Steps to reproduce

- `gatsby new my-blog gatsbyjs/gatsby-starter-blog-theme`
- `cd my-blog`
- `gatsby develop`
- browse to `http://localhost:8000`
- check the terminal to see this message:

```
Error loading a result for the page query in "/404.html". Query was not run and no cached result was
Page not found /404.html
```

- This message will be printed to the terminal every time the page is refreshed.

There's also a repro available at https://github.com/polishedwp/gatsby-starter-theme-eclectic courtesy of @polishedwp 

### Expected result

The _Error loading a result for the page query..._ message should not appear in the terminal.

### Actual result

The message appears.

### Environment

```

  System:
    OS: macOS 10.14.5
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.16.0 - ~/.nvm/versions/node/v10.16.0/bin/node
    Yarn: 1.16.0 - /usr/local/bin/yarn
    npm: 6.9.0 - ~/.nvm/versions/node/v10.16.0/bin/npm
  Languages:
    Python: 3.5.3 - /Users/mike/.pyenv/shims/python
  Browsers:
    Chrome: 75.0.3770.100
    Firefox: 67.0.4
    Safari: 12.1.1
  npmPackages:
    gatsby: ^2.13.21 => 2.13.21 
    gatsby-theme-blog: ^1.0.0 => 1.0.0 
  npmGlobalPackages:
    gatsby-cli: 2.7.7
    gatsby-dev-cli: 2.4.19
    gatsby: 2.8.8

```

m-allanson:
Note that running `gatsby build && gatsby serve` will show the same error as `error Error: ENOENT: no such file or directory, stat '/Users/mike/d/gatsby/starters/hello-world/public/404.html'`

polishedwp:
So, I installed each of the starters and the only starter that displays the 404 error in the terminal is the hello-world starter.

The 404 message appears in each of the default theme starters including the new workspace theme starter.

[...]

PR #16220, opened 2019-07-30
fix(gatsby): Check if page exists before trying to get page data

sidharthachatterjee:
Fixes https://github.com/gatsbyjs/gatsby/issues/15724


sidharthachatterjee:
Interesting that this broke a couple of end to end tests. Will fix!

PR #16227, opened 2019-07-30
fix(www): be fault tolerant for screenshots for starters

DSchau:
## Description

Wasn't falling back like we were expecting 🙄 JavaScript is hard.

### DIFF
--- a/www/src/templates/template-starter-page.js
+++ b/www/src/templates/template-starter-page.js
@@ -12,7 +12,7 @@ import StarterSource from "../views/starter/source"
 import StarterDetails from "../views/starter/details"
 import FooterLinks from "../components/shared/footer-links"
 
-const getFallbackScreenshot = (data, fallback) => {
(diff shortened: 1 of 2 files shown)


www/src/templates/template-starter-page.js
```diff
@@ -28,10 +29,14 @@ class StarterTemplate extends React.Component {
       url: demoUrl,
       repo: repoUrl,
       fields: { starterShowcase },
-      childScreenshot = getFallbackScreenshot(fallback),
```
DSchau: childScreenshot wasn't undefined, it had a structure like:

```json
{
  "screenshotFile": null
}
```

so this fallback was never invoked. Reminder to myself to not get too clever next time :)

DSchau:
@amberleyromo it's basically a noop in most scenarios -- but I changed the name. Agreed it could be a little deceptive!

PR #16232, opened 2019-07-30
fix(gatsby): Add error message for field extension validation

stefanprobst:
Improve error message in field extension arg validation.

### DIFF
--- a/packages/gatsby/src/schema/schema.js
+++ b/packages/gatsby/src/schema/schema.js
@@ -386,6 +386,15 @@ const addExtensions = ({
         .filter(name => !internalExtensionNames.includes(name))
         .forEach(name => {
           const args = fieldExtensions[name]
+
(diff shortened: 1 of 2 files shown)


packages/gatsby/src/schema/schema.js
```diff
@@ -386,6 +386,15 @@ const addExtensions = ({
         .filter(name => !internalExtensionNames.includes(name))
         .forEach(name => {
           const args = fieldExtensions[name]
+
+          if (!args || typeof args !== `object`) {
+            report.error(
+              `Field extension arguments must be provided as an object. ` +
```
KyleAMathews: Could you switch this to use the new structured errors?

There's been a number of PRs recently switching code over that you can draw from. New errors should all use it as it provides better tracking plus makes it a lot easier to refine the help we give people.

Thanks!

PR #16238, opened 2019-07-31
fix(starters): Format all the supported files

lipis:
works also in Windows :)


lipis:
Why not JSON as well.. perhaps not the lock file but package is fine.

lannonbr:
> 
> 
> Why not JSON as well.. perhaps not the lock file but package is fine.

When you do either `yarn install` or `npm install` to grab a new package, they have their own format which is different than Prettier, so it would be good to add `package.json` to that ignore list.

lipis:
I think prettier is now using the same..

lipis:
There is no harm since there are no git hooks or anything.. it's up to the end users..

sidharthachatterjee:
> There is no harm since there are no git hooks or anything.. it's up to the end users..

While it _is_ true that there aren't any git hooks, I'd rather leave this to the user if they'd like to add it themselves. Including formatting `json` seems a little too opinionated to include by default and the JSON files included by default (like `package.json`) aren't user land code anyway. 

I think sticking to `js`, `jsx` and `md` by default is a good balance between lacking and _too_ opinionated. What do you think @lipis?

lipis:
My point was that in many Gatsby projects, people are using JSON as data in the source.. so there is that, and the way it is right now, Prettier is using the same formatter as yarn and npm for package.json..

I can put it in the ignore list.. but removing JSON files I would pass if needed after all

sidharthachatterjee:
Fair enough! Let's add `package.json` to `prettierignore` and get this shipped 🙂

lipis:
Done!

wardpeet:
Do we want to add package-lock.json and npm-shrinkwrap.json to the ignore list as well?

lipis:
We recommend of using Yarn... if they decide to go with NPM, they will figure out which files they want to ignore or not.

wardpeet:
not entirely true, we advocate for yarn but we should have support for npm by default so I'll add the lock file to your pr

PR #16249, opened 2019-07-31
Generate 3x resolution images

hiddentao:
From https://kapeli.com/cheat_sheets/iOS_Design.docset/Contents/Resources/Documents/index, we need `3x` resolution assets too.

### DIFF
--- a/packages/gatsby-plugin-sharp/README.md
+++ b/packages/gatsby-plugin-sharp/README.md
@@ -60,7 +60,7 @@ plugins: [
 
 ### fixed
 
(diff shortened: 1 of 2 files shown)


LekoArts:
The 3x was removed on purpose :)
https://github.com/gatsbyjs/gatsby/pull/14201

hiddentao:
Oh wow, I had no idea!

DSchau:
As @LekoArts intimated -- this was done intentionally to exclude this dimension.

Thank you so much for this PR -- but let's go ahead and close this out. If some crazy new display tech comes out in the future that necessitates the need for this, we can revisit.

Thanks!

PR #16258, opened 2019-07-31
fix(gatsby-telemetry): Don't display analytics notification if opted out

### LINKED ISSUE #16254: Don't display the opt-out text after opting out of telemetry
## Description

The telemetry 'opt-out' instructions can display after opting out of telemetry. The text should not display after opting out.

### Steps to reproduce

See https://twitter.com/code_punkt/status/1156597414486298624

### Expected result

'opt-out' text should not display

### Actual result

'opt-out' text is displayed

### Environment

Run `gatsby info --clipboard` in your project directory and paste the output here.

alexluong:
Fixes #16254

### DIFF
--- a/packages/gatsby-telemetry/src/postinstall.js
+++ b/packages/gatsby-telemetry/src/postinstall.js
@@ -1,4 +1,5 @@
 const showAnalyticsNotification = require(`./showAnalyticsNotification`)
+const isTruthy = require("./is-truthy")
 
 let enabled = false
 try {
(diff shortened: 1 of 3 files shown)


packages/gatsby-telemetry/src/postinstall.js
```diff
@@ -1,12 +1,14 @@
...
+const isTruthy = require(`./is-truthy`)
 
 let enabled = false
 try {
   const ci = require(`ci-info`)
   const Configstore = require(`configstore`)
   const config = new Configstore(`gatsby`, {}, { globalConfigPath: true })
   enabled = config.get(`telemetry.enabled`)
-  if (enabled === undefined && !ci.isCI) {
+  const disabled = isTruthy(process.env.GATSBY_TELEMETRY_DISABLED)
```
jamo: It would be cool to have this check live in only a single place. 
One option would be to expose the disabled field from the event store instead.

sidharthachatterjee: I went ahead and did this

alexluong:
Thank you @sidharthachatterjee and @jamo for the review. It was fun.

PR #16266, opened 2019-07-31
fix(docs): Update theme building tutorial re: function config

amberleyromo:
## Description

Clarify that function configs are only supported in themes, and not in Gatsby sites. (See [tweet thread](https://twitter.com/_brotzky/status/1156662134530379776).


ChristopherBiscardi:
Content looks good to me!

PR #16268, opened 2019-07-31
docs(tutorial): Update writing-documentation-with-docz

jstacoder:
## Description 

fixed reference to wrong theme in gatsby config in docs


lannonbr:
This may want to be tweaked further given this tutorial still uses the `__expirimentalThemes` key in the gatsby-config which was deprecated as of 2.13.0.

@gatsbyjs/learning should we look into merging this in as a temporary step and then look into a PR to clean the rest of the tutorial up to be up to date with the stable api themes?

lannonbr:
And just opened an issue to triage that we need to update the doc at large.

PR #16269, opened 2019-07-31
feat(blog): JSX pragma

amberleyromo:
## Description

A short post explaining custom JSX pragma, in the context of Theme UI.


docs/blog/2019-07-31-what-is-jsx-pragma/index.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Set a pragma comment
+
+Using a pragma comment will limit the change to the modules the comment is added to. Therefore, you can use `React.createElement` by default, and _opt in_ to the custom function only where needed. This is [the approach taken in Theme UI](/docs/theme-ui/#adding-styles-to-elements).
+
+## Why use a custom JSX pragma?
+
+Using a custom JSX pragma is useful when you want to customize the transform process of JSX => JavaScript.
+
+- Emotion uses a [custom JSX pragma to use the Emotion `css` prop](https://emotion.sh/docs/css-prop#jsx-pragma).
```
DSchau: _Could_ (and I leave this up to you) be worth mentioning that we automatically enable this so you don't _have_ to use a custom pragma with `gatsby-plugin-emotion`.

More broadly -- I wasn't really certain why we wouldn't do the same with gatsby-plugin-theme-ui. Presumably there could be a conflict with e.g. emotion's pragma and css prop and the sx prop? I actually don't know 🤷‍♂

docs/blog/2019-07-31-what-is-jsx-pragma/index.md
```diff
@@ -0,0 +1,88 @@
...
+## Using a custom JSX pragma
+
+There are two ways to specify a custom function (and therefore replace `React.createElement`):
+
+1. Add an option to the Babel plugin
+2. Set a _pragma comment_ at the beginning of a module
+
+### Add an option to the Babel plugin
+
+Changing the function in tha Babel plugin will transform _all_ JSX in an application to use the specified function.
```
jxnblk:
```suggestion
Changing the function in the Babel plugin will transform _all_ JSX in an application to use the specified function.
```

amberleyromo: bahaha thank you.

docs/blog/2019-07-31-what-is-jsx-pragma/index.md
amberleyromo: psst @jxnblk?

jxnblk: Originally, Theme UI was using the same `css` prop and they *would* conflict, but since we renamed it to a different prop, the two *should* work well together.

My stance right now (which could be wrong) is that being a little cautious about where the Theme UI `jsx` function is used (i.e. only in certain themes) could be a little safer than switching over an entire Gatsby site. Emotion should work fine, but if there are other cases where a different JSX function is used in a Gatsby site, I wouldn't want a theme to blow that up. I think this is definitely something we should consider since having the `sx` prop available to any site that has `gatsby-plugin-theme-ui` is certainly a friendlier API.

ChristopherBiscardi: I think last time I dove into this, the only concern would be that the babel transform-react-jsx might have a bug where locally defined pragmas don't override a globally supplied one. there really aren't many pragmas in the wild right now though, and we own or use basically all of them so I think we'd be safe to do it and fix the bug in the babel transform later if we wanted to

docs/blog/2019-07-31-what-is-jsx-pragma/index.md
```diff
@@ -0,0 +1,88 @@
...
+
+A pragma is a compiler directive. It tells the compiler how it should handle the contents of a file.
+
+An example of this in JavaScript is `'use strict'` mode. `'use strict'` is a directive that enables JavaScript's [Strict Mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode), which is a way to opt in to a more restricted variant of JavaScript. It denotes that the code should be processed in a specific way.
+
+## What is JSX pragma?
+
+JSX syntax on its own isn’t readable by the browser. In order to ship something readable to the browser, JSX needs to be converted to plain JavaScript.
+
+Most React-based frameworks (like Gatsby), come with tooling already set up to support this conversion (usually Babel). How does that tooling know how to transform JSX? By default, the Babel plugin will convert JSX into JavaScript that calls the `React.createElement` function.
```
ChristopherBiscardi: Added a link to the babel plugin the vast majority or react projects are using:

```suggestion
Most React-based frameworks (like Gatsby), come with tooling already set up to support this conversion (usually Babel). How does that tooling know how to transform JSX? By default, [the Babel plugin](https://babeljs.io/docs/en/babel-plugin-transform-react-jsx) will convert JSX into JavaScript that calls the `React.createElement` function.
```

docs/blog/2019-07-31-what-is-jsx-pragma/index.md
amberleyromo: Unless anyone objects, I think I'm going to leave this as-is. The point of it is to note uses of custom pragma elsewhere in the ecosystem, so I don't feel a need to go into detail about implementation it the Gatsby plugin specifically here.

docs/blog/2019-08-02-what-is-jsx-pragma/index.md
```diff
@@ -0,0 +1,88 @@
...
+title: What is JSX pragma?
+date: 2019-08-02
+author: Amberley Romo
+excerpt: "Theme UI uses custom JSX pragma to make JSX theme-aware. But what is JSX pragma?"
+tags:
+  - themes
+  - theme-ui
+---
+
+If you’re looking into using [Theme UI](https://theme-ui.com/) , you’ll come across some (potentially) unfamiliar looking syntax:
```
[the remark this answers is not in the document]
marcysutton: ha, and i merged before addressing this. Feel free to disregard 😅

PR #16272, opened 2019-07-31
fix(gatsby): Panic when root config is a function

### LINKED ISSUE #16265: Warn if function is used in root site config
## Description

If you write a gatsby-config.js in the root of your site with the function style, it will fail in weird ways because plugins aren't added when you think they are.

(I'm on my phone and accidentally deleted the rest of the template I'm sorry 😭)

# Solution 

In bootstrap/index.js we should add a typeof config === function check on the root config before it is passed in to loadThemes and reporter.panic with a nice error message if it is a function

cephalization:
Not 100% sure if or where tests for this should be written, I could not find any other tests for bootstrap's index. I did use gatsby-dev to test that this change works however and it appears to do so.

Additionally, not sure if we want this panic to occur on configs that are still using experimental configs _AND_ non experimental configs so it does it before loading either config.

## Description

When bootstrap loads the root config, throw a panic if the config is of type `Function`.

## Related Issues

Fixes #16265

### DIFF
--- a/packages/gatsby-cli/src/structured-errors/error-map.js
+++ b/packages/gatsby-cli/src/structured-errors/error-map.js
@@ -83,11 +83,13 @@ const errorMap = {
   },
   "10126": {
     text: context =>
-      `A ${
+      `${context.path}/${context.configName} cannot export a function.` +
+      `\n\nA ${
         context.configName
-      } that exports a function can only be used as a theme and not run directly.` +
+      } exported as a Function can only be used as a theme and not run directly.` +
       `\nIf you are trying to run a theme directly, use the theme in an example site or starter instead and run that site to test.` +
-      `\nIf you are in the root gatsby-config.js for your site, change the export to be an object and not a function as functions are not supported in the root gatsby-config.`,
+      `\nIf you are in the root gatsby-config.js for your site, change the export to be an object and not a function as functions` +
+      `\nare not supported in the root gatsby-config.`,
     type: `CONFIG`,
     level: `ERROR`,
   },
--- a/packages/gatsby-cli/src/structured-errors/error-map.js
+++ b/packages/gatsby-cli/src/structured-errors/error-map.js
@@ -81,6 +81,17 @@ const errorMap = {
     type: `CONFIG`,
     level: `ERROR`,
   },
+  "10126": {
+    text: context =>
+      `A ${
+        context.configName
+      } that exports a function can only be used as a theme and not run directly.` +
+      `\nIf you are trying to run a theme directly, use the theme in an example site or starter instead and run that site to test.` +
+      `\nIf you are in the root gatsby-config.js for your site, change the export to be an object and not a function as functions are not supported in the root gatsby-config.`,
+    type: `CONFIG`,
+    level: `ERROR`,
+  },
+  // Plugin errors
   "11321": {
     text: context =>
       `"${context.pluginName}" threw an error while running the ${
--- a/packages/gatsby/src/bootstrap/index.js
+++ b/packages/gatsby/src/bootstrap/index.js
@@ -75,12 +75,14 @@ module.exports = async (args: BootstrapArgs) => {
     getConfigFile(program.directory, `gatsby-config`)
   )
 
+  // The root config cannot be exported as a function, only theme configs
   if (typeof config === `function`) {
-    report.panic(
-      `A gatsby-config that exports a function can only be used as a theme and not run directly.
-      If you are trying to run a theme directly, use the theme in an example site or starter instead and run that site to test.
-      If you are in the root gatsby-config.js for your site, change the export to be an object and not a function as functions are not supported in the root gatsby-config.`
-    )
+    report.panic({
+      id: `10126`,
+      context: {
+        configName: `gatsby-config`,
+      },
+    })
   }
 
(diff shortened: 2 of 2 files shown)


packages/gatsby/src/bootstrap/index.js
```diff
@@ -75,6 +75,12 @@ module.exports = async (args: BootstrapArgs) => {
     getConfigFile(program.directory, `gatsby-config`)
   )
 
+  if (typeof config === `function`) {
+    report.panic(
+      `Custom gatsby-config as a Function is not supported in site root.`
+    )
+  }
+
   // theme gatsby configs can be functions or objects
```
cephalization: Should we update this comment as well?

packages/gatsby/src/bootstrap/index.js
```diff
@@ -75,6 +75,12 @@ module.exports = async (args: BootstrapArgs) => {
     getConfigFile(program.directory, `gatsby-config`)
   )
 
+  if (typeof config === `function`) {
+    report.panic(
+      `Custom gatsby-config as a Function is not supported in site root.`
```
ChristopherBiscardi:
```suggestion
      `A gatsby-config that exports a function can only be used as a theme and not run directly.
If you are trying to run a theme directly, use the theme in an example site or starter instead and run that site to test.
If you are in the root gatsby-config.js for your site, change the export to be an object and not a function as functions are not supported in the root gatsby-config.`
```

packages/gatsby/src/bootstrap/index.js
ChristopherBiscardi: This seems fine. What would you want to update it to?

cephalization: We now panic and crash if the `gatsby-config` is a function, this seems to be the opposite of what the comment says. Unless the root config is not considered a theme config in which case I misunderstood

packages/gatsby/src/bootstrap/index.js
cephalization: This is a much more helpful message haha, thanks!

KyleAMathews:
Could you convert the error to be a structured error? E.g. https://github.com/gatsbyjs/gatsby/pull/15619

We're switching all errors over and we shouldn't add new non-structured errors.

To convert it, you need to add an error entry to https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-cli/src/structured-errors/error-map.js and then reference that in the error object you pass into `report.panic`.

Thanks! Can answer any questions you have. We need to create a doc page for this as well.

cephalization:
> Could you convert the error to be a structured error? E.g. #15619
> 
> We're switching all errors over and we shouldn't add new non-structured errors.
> 
> To convert it, you need to add an error entry to https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-cli/src/structured-errors/error-map.js and then reference that in the error object you pass into `report.panic`.
> 
> Thanks! Can answer any questions you have. We need to create a doc page for this as well.

I can give it a shot 👍

KyleAMathews:
Thanks!

--
Kyle Mathews

Blog: http://bricolage.io

cephalization:
Does the id of the structured error just iterate off of the last error id?

KyleAMathews:
The errors are pretty random actually :-D

Part of the goal of structured errors is that people can google the numbers and find something specific to the error. So it doesn't really matter what you pick as long as it isn't the same as another error and it isn't a "common" number e.g. 123456.

Incremented IDs tend to come up when someone is doing a PR converted unstructured errors and converts several together. There's no meaning otherwise.

cephalization:
2019-07-31 20_12_59-gatsby-config js - gatsby-test - Visual Studio Code

packages/gatsby-cli/src/structured-errors/error-map.js
```diff
@@ -81,6 +81,15 @@ const errorMap = {
     type: `CONFIG`,
     level: `ERROR`,
   },
+  "10126": {
+    text: context =>
+      `A ${
+        context.configName
+      } that exports a function can only be used as a theme and not run directly.\nIf you are trying to run a theme directly, use the theme in an example site or starter instead and run that site to test.\nIf you are in the root gatsby-config.js for your site, change the export to be an object and not a function as functions are not supported in the root gatsby-config.`,
```
cephalization: This is a huge giant line with newline characters because if I just move them to the next line manually I get indentation. I could not use `report.stripIndent` while using structured errors as far as I could tell

cephalization:
Not sure why circle is choking, will look into it tomorrow

packages/gatsby/src/bootstrap/index.js
```diff
@@ -75,6 +75,16 @@ module.exports = async (args: BootstrapArgs) => {
     getConfigFile(program.directory, `gatsby-config`)
   )
 
+  // The root config cannot be exported as a function, only theme configs
+  if (typeof config === `function`) {
+    report.panic({
+      id: `10126`,
+      context: {
+        configName: `gatsby-config`,
```
KyleAMathews: This should pass in the path to the config file as well so it's clear which one to use.

cephalization:
Screen Shot 2019-08-01 at 6 42 48 AM

Issue #16097, opened 2019-07-25
Requests for page-data.json return 404 in client side routes


kirbydcool:
## Description

Hey team! I've noticed an issue with client side routes recently.

When navigating to client side routes, Gatsby requests the associated `page-data.json` which returns a 404. For example, navigating to `/app/profile` loads `/page-data/app/profile/page-data.json`, which returns a 404 if the `/app` page has a matchPath of `/app/*`. This is reproducible in the [simple-auth example](https://github.com/gatsbyjs/gatsby/tree/master/examples/simple-auth) which has a `gatsby-node.js` like this:

```js
exports.onCreatePage = async ({ page, actions }) => {
  const { createPage } = actions

  // page.matchPath is a special key that's used for matching pages
  // only on the client.
  if (page.path.match(/^\/app/)) {
    page.matchPath = `/app/*`

    // Update the page.
    createPage(page)
  }
}
```

Note this doesn't seem to affect the rendering of the page. It looks like it's still getting the correct page data for the matchPath page even though Gatsby is making a request for different page data.

### Steps to reproduce

You can reproduce this in the the [simple-auth example](https://github.com/gatsbyjs/gatsby/tree/master/examples/simple-auth).

1. Clone the simple-auth example
2. Install the latest gatsby
3. `gatsby build` and `gatsby serve`
4. Navigate to `localhost:9000/app/login`

You get some 404s for page-data
![image](https://user-images.githubusercontent.com/696150/61902918-1cbf9f80-aed8-11e9-994a-beeedf19ab93.png)

### Expected result

Gatsby shouldn't request `page-data.json` for client side routes.

### Actual result

Gatsby requests `page-data.json` for client side routes and receives a 404.

### Environment

  System:
    OS: macOS 10.14.5
    CPU: (8) x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
    Shell: 3.0.2 - /usr/local/bin/fish
  Binaries:
    Node: 12.3.1 - ~/.asdf/shims/node
    Yarn: 1.16.0 - ~/.asdf/shims/yarn
    npm: 6.9.0 - ~/.asdf/shims/npm
  Languages:
    Python: 2.7.15 - /Users/thanx/.asdf/shims/python
  Browsers:
    Chrome: 75.0.3770.142
    Safari: 12.1.1
  npmPackages:
    gatsby: ^2.13.41 => 2.13.41
[...]

PR #16275, opened 2019-08-01
fix: Check for match-path when loading page-data json

scottyeck:
## Description

Per #16097 , requests made to `page-data.json` don't currently reflect client-only routes. This PR addresses that issue by checking for a corresponding `matchPath` before determining the path to the desired `page-data.json` on a given page.

### Current State

As described in [this comment](https://github.com/gatsbyjs/gatsby/tree/master/examples/client-only-paths), the underlying issue can be seen in the `client-only-paths` example. Because there's a link to several client-only paths on the page - (eg. `/page/1`), the `BaseLoader` tries to pre-fetch the corresponding `page-data.json` assets, which it presumes to be at `/page-data/page/1/page-data.json`. Because that's a client-only path, the corresponding _page_ doesn't even exist and as such, neither does the `page-data.json` at that path, resulting in the 404.

localhost_8080_page_4

### The Fix

With the fix, however, for each individual route (eg. `/page/1`), a `matchPath` of `/` is identified which corresponds to an entry in the `pagesDb` on the `BaseLoader` instance, thus preventing the additional requests from being issued, since `/page-data/index/page-data.json` has already been requested and cached.

localhost_8080_page_4

## Related Issues

#16097

## Additional questions

I'd love to add tests but I'm not sure where to begin in that regard - it's been a long while since I've contributed and the codebase has evolved tremendously to say the least. Any additional direction would be super helpful.

### DIFF
--- a/packages/gatsby/cache-dir/loader.js
+++ b/packages/gatsby/cache-dir/loader.js
@@ -144,7 +144,7 @@ export class BaseLoader {
   }
 
   loadPageDataJson(rawPath) {
-    const pagePath = cleanPath(rawPath)
+    const pagePath = cleanPath(findMatchPath(rawPath) || rawPath)
     if (this.pageDataDb.has(pagePath)) {
       return Promise.resolve(this.pageDataDb.get(pagePath))
     }


scottyeck:
Welp. TIL that static siblings living alongside client-only paths are a thing. I'd love to treat this as some kind of fallback but I'm not entirely sure how I'd go about doing that. Gonna noodle on it. Feel free to close if y'all think it's entirely impossible and don't wish to see me blabbering on.

In other news, the test-bed y'all have put together is absolutely incredible. Well done!

scottyeck:
Follow-up thought. Detect a `404` from the initial fetch and _then_, if the path contains a `matchPath`, attempt to fetch `page-data` for the corresponding `match-path` root. It doesn't help from polluting the console with unnecessary `404`s, but it sounds like it'll solve the problem of missing `page-data`.

wardpeet:
Hey :wave:,

I do get your concerns but sadly we can't fix this. If we don't do the page-data check for the current page we'll bring back #15101. 

Consider the following routes
```
[
  {
    "path": "/app/dashboard",
    "matchPath": "/app/*"
  },
  {
    "path": "/app/login",
  },
]
```

With that configuration, we won't be able to load /app/login as it's part of the matchPath logic. We will simply ignore the page and fallback to whatever is inside /app/*. The change to fix this behaviour was introduced in 
https://github.com/gatsbyjs/gatsby/pull/15762

A solution we could add is to save all paths that are inside a matchPath also into the matchPath.json so we know about all routes. This might bring back the page-manifest performance problem for site that heavily uses this behaviour.

Matchpath will look like
```
[
  {
    "path": "/app/dashboard",
    "matchPath": "/app/*"
  },
  {
    "path": "/app/login",
    "matchPath": "/app/login"
  },
]
```

scottyeck:
> Follow-up thought. Detect a 404 from the initial fetch and then, if the path contains a matchPath, attempt to fetch page-data for the corresponding matchPath root. It doesn't help from polluting the console with unnecessary 404s, but it sounds like it'll solve the problem of missing page-data.

^^ @wardpeet Any thoughts on this potential approach?

wardpeet:
> Follow-up thought. Detect a 404 from the initial fetch and then, if the path contains a matchPath, attempt to fetch page-data for the corresponding matchPath root. It doesn't help from polluting the console with unnecessary 404s, but it sounds like it'll solve the problem of missing page-data.

Unsure what this means to be honest. Isn't that what we're doing? We fetch the current path and if it's a 404 we fallback to the matchPath?

wardpeet:
@scottyeck Unfortunately, we're not going to merge this at this time. We'll go forward with a PR that does https://github.com/gatsbyjs/gatsby/pull/16275#issuecomment-517382570 so we'll fix the 404 errors which will make this PR obsolete.

As such--we're going to close this out. We want you to know that we certainly want to keep you as a contributor to Gatsby, so we'd encourage you to check out our open issues.

scottyeck:
@wardpeet No problem. I recognize that this solution didn't quite cover the current specifications for client-routes.

I've been following along on the other thread and am pretty thrilled that y'all intend to patch after all!

PR #16282, opened 2019-08-01
fix(telemetry): don't show telemetry notification when telemetry collection is disabled via env flag

jamo:
The env check was also implemented deeper just before sending events, and
hence the flag did work even though we showed the notification. 

fixes #16258


jamo:
Uuh, looks like there was an earlier PR for this: https://github.com/gatsbyjs/gatsby/pull/16258

PR #15979, opened 2019-07-22
fix(blog): blog tag audit and overhaul

gillkyle:
## Description

Me and @calcsam paired for a day on auditing the tags we have on the blog to cut down on topics that are too specific or redundant. 

The audit of tags happened on this sheet: https://docs.google.com/spreadsheets/d/1HwDkk55h85dLWJTPHk_3GvGfJmqqL_iunlOYPjdGbcs/edit#gid=0

We pulled from key marketing messages that have been outlined in narrowing the tags, the goal was to remove 1/3 of the existing tags but we ended up removing or renaming nearly 50% (**from 165 tags to 81**)

## Screenshots

A popular tags section was added to highlight some commonly used tags:

<details>
<summary>Tags page (BEFORE)</summary>

image

</details>

<details>
<summary>Tags page (AFTER)</summary>

image

</details>

And the tag details pages were updated to include a link back to docs for tags that have docs associated with them:

<details>
<summary>Tag details</summary>

image

</details>

PR #16309, opened 2019-08-01
chore(blog): Rename md files to mdx

sidharthachatterjee:
We recently broke a build for `www` because of some incorrect markup in docs/blog/2018-09-27-reach-router/index.md (https://github.com/gatsbyjs/gatsby/pull/15979)

This was not caught by prettier in the `md` mode but does indeed get caught in `mdx` mode! 

So we're renaming all our blog entries to the `mdx` extension in this PR. That should prevent issues like this in the future since our linting check will catch them.


m-allanson:
@sidharthachatterjee any idea why the lints are failing here? I'm guessing it's fairly quick to recreate this PR so I'm going to close this for now. To prevent blocking other docs PRs.

sidharthachatterjee:
@m-allanson Looks like strange behaviour on prettier’s part. The same file fails linting checks after running prettier format on it. Will need to debug.

PR #16312, opened 2019-08-01
feat(gatsby-source-shopify): add metafields to product nodes

kellyvaughn:
## Description

Shopify introduced access to metafields via the Storefront API back in May. This adds the metafield endpoint to products via gatsby-source-shopify.

## Related Issues

N/A

### DIFF
--- a/packages/gatsby-source-shopify/src/nodes.js
+++ b/packages/gatsby-source-shopify/src/nodes.js
@@ -14,6 +14,7 @@ const COMMENT = `Comment`
 const PRODUCT = `Product`
 const PRODUCT_OPTION = `ProductOption`
 const PRODUCT_VARIANT = `ProductVariant`
+const PRODUCT_METAFIELD = `ProductMetafield`
 const SHOP_POLICY = `ShopPolicy`
 const PRODUCT_TYPE = `ProductType`
 const PAGE = `Page`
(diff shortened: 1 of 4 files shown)


packages/gatsby-source-shopify/src/nodes.js
```diff
@@ -129,6 +138,9 @@ export const ProductNode = imageArgs =>
     return node
   })
 
+export const ProductMetafieldNode = _imageArgs =>
```
DSchau: Not really new here -- but I wonder why this pattern is used through all of these node types?

Oh well!

kellyvaughn: That's a really good question - I didn't want to reinvent anything, so I just followed the same pattern as before. I don't totally understand it to be entirely honest.

DSchau: 🤷‍♂ Your guess is as good as mine. Not hurting anything -- we'll clean it up later!

DSchau:
Published as [`gatsby-source-shopify`@`2.2.0`](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-source-shopify/CHANGELOG.md#220-2019-08-01).

Enjoy the swag 🎉

PR #16319, opened 2019-08-02
ci: add an (optional) approval step to build out www

DSchau:
## Description

We can re-run a job that (possibly) failed for an automatic deploy, but that has the problem of being a snapshot at that commit hash. If we'd like to grab _new changes_ this should do the trick.

This should be used sparingly, only when something needs to go live at a certain time when the automated schedule won't suffice.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -318,6 +318,17 @@ workflows:
             branches:
               only:
                 - master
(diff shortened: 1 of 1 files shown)


DSchau:
This _may not work_ because an approval job will mark the entire suite of checks as pending. Let's see.

Issue #16317, opened 2019-08-01
[gatsby-plugin-shopify] Remove undocumented ProductType query


jlengstorf:
This would be a breaking change, but Shopify's GraphQL API no longer documents the `ProductType` type and the query for product types has started timing out. The product type data is still available via the `product` query, so we don't lose any data by removing this.

Recommendation:

1. Remove the `PRODUCT_TYPE` query in `gatsby-source-shopify`
	- Remove the query itself ([here](https://github.com/gatsbyjs/gatsby/blob/944363288d51d2cd0b845dc46f028122fba9d9a9/packages/gatsby-source-shopify/src/queries.js#L224-L238)) and any code that uses it
2. Remove the `ProductType` node ([here](https://github.com/gatsbyjs/gatsby/blob/944363288d51d2cd0b845dc46f028122fba9d9a9/packages/gatsby-source-shopify/src/nodes.js#L19)) an all the code that relies on it

**NOTE:** This will be a breaking change for the plugin!

As a bonus: some of the queries are deprecated in this plugin. This is not an issue, strictly speaking, but refactoring to use the up-to-date queries will help us stay inline with Shopify's API and docs.

This could be a great way for someone new to Gatsby to get a first PR in!

nivekselig:
I'll take a look at it. I just started using Gatsby this week.

mimibar:
Is this assigned? I think I have identified the changes needed.

pqt:
It's not, would you like to take this one on @mimibar?

cj2:
Thanks, @jlengstorf! This helped me get past the build issue.

Aswinpookkatt:
Is issue still open?

pqt:
Yeah, just pending the merge of #16320 @Aswinpookkatt

sidharthachatterjee:
Published in `gatsby-source-shopify@3.0.0` thanks to @mimibar 🥇

PR #16320, opened 2019-08-02
[gatsby-plugin-shopify] Remove undocumented ProductType query #16317

mimibar:


mimibar:
@jlengstorf  Would you mind taking a look and see if anything is missing?

Thanks a lot!

kang-consulting:
I was still having the same issue after testing this and found some more references to product type by searching for "ProductTypes". After removing those, the error went away.

m-allanson:
Thanks @mimibar this looks good! 🎉

This will be a breaking change - @sidharthachatterjee had some ideas on how to make the upgrade process a bit smoother by using Gatsby's schema customisation APIs. 

This plugin didn't have any tests yet, so I've added a fairly barebones snapshot test for now.

sidharthachatterjee:
Thanks for the tag @m-allanson 🥇 

@mimibar This looks good but I'd like to avoid shipping a major version/breaking change as much as possible and in that regard, I was wondering if we could avoid the same by

- Using `createTypes` to add `ProductType` 
- Mark it as deprecated in the schema 
- Add a custom resolver to always return `null` for it

This should give users a gentler migration experience but I'm not sure if it's worth the added maintenance overhead. 

@pieh Do you think this is worth the effort/maintenance overhead as opposed to shipping a major version?

sidharthachatterjee:
Published in `gatsby-source-shopify@3.0.0`

PR #16328, opened 2019-08-02
fix(gatsby): Prefer a user's shouldUpdateScroll over a plugin's

### LINKED ISSUE #12038: [gatsby-remark-autolink-headers] Cannot override scroll behavior via shouldUpdateScroll when gatsby-remark-autolink-headers plugin is used.
## Description

I have a page on which I want to disable scroll update. In order to achieve that I defined `shouldUpdateScroll` function in `gatsby-browser.js` file in the root of the project  which returns false when location matches my page's url. But when `gatsby-remark-autolink-headers` is used result of my `shouldUpdateScroll` is ignored.

As far as I understand, when Gatsby decides whether it should update scroll or not, it runs `shouldUpdateScroll` functions defined by plugins first and then it runs the function defined in the project. But only the first result is used (in my case the result returned from `gatsby-remark-autolink-headers` ): https://github.com/gatsbyjs/gatsby/blob/3d77520272e2ea885bd09a7f9f8e0004d378387b/packages/gatsby/cache-dir/navigation.js#L90-L101

`gatsby-remark-autolink-headers` defines its own `shouldUpdateScroll` function which returns `true` if url does not contain a hash:
https://github.com/gatsbyjs/gatsby/blob/390803c8f679e53d398d40b9535c1a8784380a48/packages/gatsby-remark-autolink-headers/src/gatsby-browser.js#L27-L30

Probably `shouldUpdateScroll` function defined in `gatsby-remark-autolink-headers` should return `undefined` instead of `true` if URL does not contain a hash. In this case its result will be ignored and it will be possible to disable scroll update for certain pages.
https://github.com/gatsbyjs/gatsby/blob/3d77520272e2ea885bd09a7f9f8e0004d378387b/packages/gatsby/cache-dir/api-runner-browser.js#L36-L37

### Steps to reproduce

Here's a link to the demo project: https://github.com/meyginix/gatsby-issue-reproduction.

### Expected result

Developer should be able to change scroll update behavior by defining `shouldUpdateScroll` function when `gatsby-remark-autolink-headers` plugin is included into a project.

Probably `shouldUpdateScroll` function defined in `gatsby-remark-autolink-headers` should return `undefined` instead of true if URL does not contain a hash. In this case its result will be ignored and it will be possible to disable scroll update for certain pages.

### Actual result

`gatsby-remark-autolink-headers` plugin does not allow to change scroll update behavior via `shouldUpdateScroll` function.

### Environment

```
System:
    OS: Linux 4.15 Linux Mint 18.3 (Sylvia)
    CPU: (8) x64 Intel(R) Core(TM) i5-8250U CPU @ 1.60GHz
    Shell: 4.3.48 - /bin/bash
  Binaries:
    Node: 8.11.4 - /usr/bin/node
    Yarn: 1.13.0 - /usr/bin/yarn
    npm: 6.4.1 - /usr/bin/npm
  Languages:
    Python: 2.7.12 - /usr/bin/python
  Browsers:
    Chrome: 72.0.3626.119
    Firefox: 65.0.1
  npmPackages:
    gatsby: ^2.1.17 => 2.1.17 
    gatsby-plugin-manifest: ^2.0.19 => 2.0.19 
    gatsby-remark-autolink-headers: ^2.0.14 => 2.0.14 
    gatsby-source-filesystem: ^2.0.22 => 2.0.22 
    gatsby-transformer-remark: ^2.2.6 => 2.2.6 
  npmGlobalPackages:
    gatsby-cli: 2.4.4
```

stefanprobst:
When a plugin's and a user's `gatsby-browser.js` both define a `shouldUpdateScroll`, we currently pick the result from the plugin. We should allow to overwrite the plugin-defined behavior by the user config.

Fixes #12038

### DIFF
--- a/packages/gatsby/cache-dir/navigation.js
+++ b/packages/gatsby/cache-dir/navigation.js
@@ -129,6 +129,8 @@ function shouldUpdateScroll(prevRouterProps, { location }) {
     getSavedScrollPosition: args => this._stateStorage.read(args),
   })
   if (results.length > 0) {
+    // Use the latest registered shouldUpdateScroll result, this allows users to override plugin's configuration
+    // @see https://github.com/gatsbyjs/gatsby/issues/12038
     return results[results.length - 1]
   }
 
--- a/packages/gatsby/cache-dir/navigation.js
+++ b/packages/gatsby/cache-dir/navigation.js
@@ -129,7 +129,7 @@ function shouldUpdateScroll(prevRouterProps, { location }) {
     getSavedScrollPosition: args => this._stateStorage.read(args),
   })
   if (results.length > 0) {
-    return results[0]
+    return results[results.length - 1]
   }
 
   if (prevRouterProps) {


packages/gatsby/cache-dir/navigation.js
```diff
@@ -129,7 +129,7 @@ function shouldUpdateScroll(prevRouterProps, { location }) {
     getSavedScrollPosition: args => this._stateStorage.read(args),
   })
   if (results.length > 0) {
-    return results[0]
+    return results[results.length - 1]
```
wardpeet:
```suggestion
    // Use the latest registered shouldUpdateScroll result, this allows users to override plugin's configuration
    // @see https://github.com/gatsbyjs/gatsby/issues/12038
    return results[results.length - 1]
```

stefanprobst: :+1: thanks!

PR #16329, opened 2019-08-02
feat(gatsby-plugin-sitemap): add siteUrl validation

### LINKED ISSUE #15754: gatsby-plugin-sitemap does not throw error if siteUrl is not filled
Using the sitemap gatsby plugin I mentioned that there is no validation of siteUrl and if you don't fill it you will get an error on the building HTML step, but without any clear message that could reference you to this issue.

https://www.gatsbyjs.org/docs/creating-a-sitemap/

In the documentation also that section with gatsby-config with siteMetadata more looks like typical description of the plugin when siteMetadata inserted just to get user an understanding how the default gatsby-config looks like.

Proposal:
- Add in the documentation information about "siteUrl" key
- Add validation for that key and throw proper error message when it is empty

doc-han:
## Description
This PR is a fix for #15754 

1. Validating siteUrl and throwing an error if it's an empty string
2. Adding a note to docs to prompt that siteUrl can't be left blank or undefined

## Related Issues
Fixes #15754

### DIFF
--- a/packages/gatsby-plugin-sitemap/src/internals.js
+++ b/packages/gatsby-plugin-sitemap/src/internals.js
@@ -29,18 +29,13 @@ export const runQuery = (handler, query, excludes, pathPrefix) =>
     })
 
     // siteUrl Validation
-    if (!r.data.site.siteMetadata.siteUrl) {
-      throw new Error(
-        `SiteMetaData 'siteUrl' property is required. Check out the documentation to see a working example: https://www.gatsbyjs.org/packages/gatsby-plugin-sitemap/#how-to-use`
-      )
-    }
-
     if (
+      !r.data.site.siteMetadata.siteUrl ||
       r.data.site.siteMetadata.siteUrl == null ||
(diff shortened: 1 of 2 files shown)


doc-han:
Anyone to check this out?

packages/gatsby-plugin-sitemap/src/internals.js
```diff
@@ -28,12 +28,22 @@ export const runQuery = (handler, query, excludes, pathPrefix) =>
...
     })
 
+    // siteUrl Validation
     if (!r.data.site.siteMetadata.siteUrl) {
       throw new Error(
         `SiteMetaData 'siteUrl' property is required. Check out the documentation to see a working example: https://www.gatsbyjs.org/packages/gatsby-plugin-sitemap/#how-to-use`
       )
     }
 
+    if (
```
wardpeet: I think we should augment the above if instead of creating a whole new one.

doc-han: okay

PR #16331, opened 2019-08-02
fix(gatsby-source-mongodb): work with _id types that are not ObjectIds

senorpedro:
## Description

Using mongodb, it is possible to use arbritary values (e.g. strings) as _id.
In that case the current implementation will throw
"TypeError: _id.toHexString is not a function"
This fix will prevent this, at least for string _ids.


senorpedro:
> I hate that Mongo supports arbitrary data types for `_id` but this seems like a fair change to handle edge cases like those.
> 
> Thank you for patching this @senorpedro 👍

You're welcome! Thanks for creating such awesome software! :-)

PR #16335, opened 2019-08-02
feat(gatsby): Allow all dateformat options as directive args

stefanprobst:
We currently only have `locale` and `formatString` as config arguments for the `@dateformat` extension. This PR adds `fromNow` and `difference`.


wardpeet:
Published in gatsby@2.13.51

PR #16344, opened 2019-08-02
fix(gatsby-plugin-mdx): fix filter by heading depth

trevorblades:
This branch fixes an issue where heading fields on an MDX node can't be filtered by depth. Here's the query that I was using:

```graphql
{
  allMdx {
    nodes {
      headings(depth: h2) {
        depth
      }
    }
  }
}
```

In this case, all headings would be returned, regardless of their depth. This was happening because in the `headings` resolver, the `depth` arg was defined as a `HeadingMdx` [here](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-mdx/gatsby/source-nodes.js#L162), but filtering would only happen [if `depth` is a number](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-mdx/gatsby/source-nodes.js#L175).

I changed the logic to only perform filtering when `depth` is one of the values that `HeadingMdx` allows, and adjusted the filtering to make this strategy work properly.

### DIFF
--- a/packages/gatsby-plugin-mdx/gatsby/source-nodes.js
+++ b/packages/gatsby-plugin-mdx/gatsby/source-nodes.js
@@ -59,7 +59,7 @@ module.exports = (
   const { createTypes } = actions;
 
   const options = defaultOptions(pluginOptions);
-
+  const headingsMdx = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'];
   createTypes(`
     type MdxFrontmatter {
       title: String!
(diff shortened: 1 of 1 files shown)


ChristopherBiscardi:
😅 thanks for catching this :)

trevorblades:
@chrisbiscardi the www build failed, but do you think we can merge this in and publish a new version anyway?

ChristopherBiscardi:
@trevorblades unfortunately merging is tied to WWW passing now. @gatsbyjs/core any idea what the github token issue is here?

DSchau:
Published in gatsby-plugin-mdx@1.0.23

Thanks @trevorblades!

PR #16348, opened 2019-08-03
docs(mdx): Update getting-started.md

davidkpiano:
Remove outdated steps, which use an outdated setup and dependencies and currently does not work as-is.

### DIFF
--- a/docs/docs/mdx/getting-started.md
+++ b/docs/docs/mdx/getting-started.md
@@ -2,36 +2,9 @@
 title: Getting Started with MDX
 ---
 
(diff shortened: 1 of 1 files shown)


ChristopherBiscardi:
Hey @davidkpiano! Thanks for bringing this up. I've created an issue to track this (https://github.com/gatsbyjs/gatsby/issues/16350) and have updated all the dependencies in the original repo in the meantime. Instead of removing the quickstart, I'm going to close this PR pending resolution of the issue I just created (moving the starter to be under Gatsby ownership).

PR #16354, opened 2019-08-03
fix(gatsby): Don't delete 404 page-data so we don't crash build

### LINKED ISSUE #15273: Querying GraphQL in 404 page not working
## Description

I'm getting error when running `gatsby build` for a second time (with cache present).
It happens only when I query image in 404 page.

### Steps to reproduce

1. Fork this sandbox https://codesandbox.io/s/gatsby-starter-default-edlku where I replaced `404.js` page content with content from `components/image.js`.
2. Run `yarn build`

### Expected result

Everything works as expected.

### Actual result

```
Error: ENOENT: no such file or directory, open '/sandbox/public/page-data/404.html/page-data.json'
```

### Environment

Codesandbox

stefanprobst:
We currently crash on `build` with a warm cache when there is a query in the 404 page, because we delete `public/page-data/404.html` during bootstrap but expect it is still there later.

Fixes #15273

### DIFF
--- a/packages/gatsby/src/bootstrap/index.js
+++ b/packages/gatsby/src/bootstrap/index.js
@@ -144,6 +144,7 @@ module.exports = async (args: BootstrapArgs) => {
     await del([
       `public/*.{html,css}`,
       `public/**/*.{html,css}`,
+      `!public/page-data/404.html`,
       `!public/static`,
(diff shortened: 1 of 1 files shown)


wardpeet:
Fixed in gatsby@2.13.51

PR #16359, opened 2019-08-03
feat(gatsby): Allow plugins to make functionality available on field resolver context

stefanprobst:
Allow plugins to put methods on `context`, which would allow using plugin functionality to compose new custom types (e.g. more than one markdown type with different frontmatter fields).


stefanprobst:
Yes sure!

PR #16369, opened 2019-08-04
fix(gatsby): createSchemaCustomization typings

RIP21:
fix: update GatsbyNode typings adding `createSchemaCustomization` typings

### DIFF
--- a/packages/gatsby/index.d.ts
+++ b/packages/gatsby/index.d.ts
@@ -374,7 +374,7 @@ export interface GatsbyNode {
     options: PluginOptions,
     callback: PluginCallback
   ): void
-    
+
   /**
    * Customize Gatsby’s GraphQL schema by creating type definitions, field extensions or adding third-party schemas.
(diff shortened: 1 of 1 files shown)


RIP21:
I didn't run prettier or whatever. I did this change right from the GitHub UI, sorry :) Will be grateful for some to apply proper code formatting for me :P

wardpeet:
Hey @RIP21

I cleaned up the PR a little as the typings weren't 100% correct. I do get why you've used pick for the actions but they aren't 100% on the same line as the args the `createSchemaCustomization` actually gets.

RIP21:
@wardpeet question. It's a second time I see in the typings people using backticks to define some string. Isn't it not valid syntax? Cause Webstorm for example, complains :)

Anyway, thanks a lot, my main idea was to show the absence of such thing in typings :) Will be glad that to be merged and added next patch or whatever version :)

wardpeet:
yeah sorry bad copy :) Good catch 😉

PR #16372, opened 2019-08-05
feat(gatsby-cli): Added support and docs for NO_COLOR env variable

### LINKED ISSUE #16324: Support NO_COLOR environment variable
## Summary

The console output of e.g. `gatsby build` is not very readable in no-color environments or CI/CD tools, where this is not supported.

### Basic example

Gatsby should support the `NO_COLOR` environment variabel (https://no-color.org/) and disable colored output accordingly.

### Motivation

- Better readability of logs in CI environments
- Better accessibility for users who prefer non-colored output

evolkmann:
## Description

Adds support for the `NO_COLOR` environment variable as described on https://no-color.org/.
As described in the spec, the simple presence of the env variable is checked, not its value (e.g. `NO_COLOR=false` also produces un-colored output).

Usage Example:

`export NO_COLOR=somevalue && gatsby build`

or set the env variable globally.

## Related Issues

Closes #16324.

### DIFF
--- a/packages/gatsby-cli/src/create-cli.js
+++ b/packages/gatsby-cli/src/create-cli.js
@@ -83,8 +83,7 @@ function buildLocalCommands(cli, isLocalSite) {
     return argv => {
       report.setVerbose(!!argv.verbose)
 
-      const noColor = argv.noColor || process.env.NO_COLOR
-      report.setNoColor(noColor)
+      report.setNoColor(argv.noColor || process.env.NO_COLOR)
 
       process.env.gatsby_log_level = argv.verbose ? `verbose` : `normal`
       report.verbose(`set gatsby_log_level: "${process.env.gatsby_log_level}"`)
--- a/packages/gatsby-cli/src/create-cli.js
+++ b/packages/gatsby-cli/src/create-cli.js
@@ -82,7 +82,7 @@ function buildLocalCommands(cli, isLocalSite) {
   function getCommandHandler(command, handler) {
     return argv => {
       report.setVerbose(!!argv.verbose)
-      if (argv.noColor) {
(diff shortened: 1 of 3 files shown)


evolkmann:
As soon as this is merged, I will also update the list of software at https://github.com/jcs/no_color.

packages/gatsby-cli/src/create-cli.js
```diff
@@ -82,7 +82,7 @@ function buildLocalCommands(cli, isLocalSite) {
   function getCommandHandler(command, handler) {
     return argv => {
       report.setVerbose(!!argv.verbose)
-      if (argv.noColor) {
+      if (argv.noColor || process.env.NO_COLOR) {
```
wardpeet: On line 92 there is another `argv.noColor` usage. So I would put the noColor in a variable so we can re-use it in the function.
```suggestion
      const noColor = argv.noColor || process.env.NO_COLOR
      if (noColor ) {
```

evolkmann: Thanks for your feedback, I refactored the section in https://github.com/gatsbyjs/gatsby/pull/16372/commits/ad4c029f6a2f131f97cdee319d792fd3c12713d1.

Another suggestion from my side is to put [lines 87-90](https://github.com/gatsbyjs/gatsby/pull/16372/commits/ad4c029f6a2f131f97cdee319d792fd3c12713d1#diff-dfccf547543bb987b0ff0a7281348933R87-R90) into the `reporter.setNoColor()` function, since it is just another internal color setting. What do you think?

wardpeet: oh yeah that sounds great! can you do that refactor in this pr? 😄

PR #16373, opened 2019-08-05
docs: Update import for Emotion 10

johncookv:

### DIFF
--- a/docs/docs/testing-css-in-js.md
+++ b/docs/docs/testing-css-in-js.md
@@ -35,10 +35,7 @@ In order to tell Jest to use the serializer you'll need to create the file `setu
 
 ```js:title=setup-test-env.js
 import { createSerializer } from "jest-emotion"
-// for Emotion 10
 import * as emotion from "@emotion/core"
-// for Emotion 9 and below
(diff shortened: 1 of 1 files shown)


gillkyle:
Hey @johncookv, thanks for the PR!

I think in a case like this we are probably fine just including the most recent version of import syntax. I can see these comments getting out of date in future emotion releases so how about we just update it to show the new import from `@emotion/core`?

docs/docs/testing-css-in-js.md
```diff
@@ -35,7 +35,10 @@ In order to tell Jest to use the serializer you'll need to create the file `setu
 
 ```js:title=setup-test-env.js
 import { createSerializer } from "jest-emotion"
-import * as emotion from "emotion"
+// for Emotion 10
+import * as emotion from "@emotion/core"
```
gillkyle: I think we'd be good just keeping this line as a change, the comments are accurate but I think that could be found on emotion's docs so we don't have to be responsible for it 🙂

johncookv: Noted. Thanks! I went ahead and updated per your suggestion.

PR #16389, opened 2019-08-05
feat(build): builds use a chunk group hash

### LINKED ISSUE #15872: Can you opt-out of webpackCompilationHash (to improve Netlify build times)
## Summary
Within the `gatsby-script-loader` script at the bottom of any page there is a `window.webpackCompilationHash` set. This hash is updated during every change to anything in the `src` directory.

This means deploying onto services such as Netlify take longer than expected. Netlify has advanced features for builds so only assets which content has updated since the last build are transferred to the underlying CDN. However, because the `webpackCompilationHash` is updated if anything changes in `src` this means all files will be transferred for only a small change.

For example:

1. `gatsby build` - To give an initial built site.
2. Change the `404` template.
3. `gatsby build` - Builds site but all outputted pages are changed because a single file has been updated.

For larger sites generating 1000s of pages this is a real problem. An update to a fairly static page will result in the dynamic pages also being updated.

Is it possible that we could add an option for the hash to not be generated an inlined in HTML output?

How integral is it to the inner workings of Gatsby?

## Relevant information

Mainly affects large sites being deployed on services such as Netlify.

afenton90:
## Description

`webpackCompilationHash` is currently used in `page-data.json` & HTML files to decipher whether the Gatsby app running in the browser is at the same version as the code on the server/CDN.

At the moment this hash changes if anything within `./src` has changed between builds. Meaning a 404 file could be changed which would result in all `page-data.json` & HTML files being updated even if they have not changed between builds.

This causes some issues when using some cloud CDN providers as described in this issue #15872 .

This PR now constructs the `webpackCompilationHash` from hashes for each named chunk in webpack. Meaning that if a page or any associated files are changed between builds, only those files that have been updated will receive a new `webpackCompilationHash`, allowing for diffs to take place between what is hosted in a CDN and what is about to be pushed to a CDN.
 
## Related Issues

closes #15872


sidharthachatterjee:
Needs more tests. Alex and I will be pairing on these on Tuesday!

sidharthachatterjee:
Alex and I paired on this some more today and managed to verify the chunk hash behaviour. However, we found that there are edge cases that this doesn't currently handle:

- First time build sometimes leaves hash set to `undefined` 
- Webpack runtime and app chunk are parents of every page chunk and invalidation can be problematic if the hash is not a function of them as well

afenton90:
Closing in favour of #16686

PR #16395, opened 2019-08-05
Update sites.yml

97thjingba:

### DIFF
--- a/docs/sites.yml
+++ b/docs/sites.yml
@@ -6850,11 +6850,11 @@
   built_by: Fourpost
   built_by_url: https://github.com/fourpost
   featured: false
-  - title: PersonalBlog Example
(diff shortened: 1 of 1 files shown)


docs/sites.yml
```diff
@@ -6850,3 +6850,14 @@
   built_by: Fourpost
   built_by_url: https://github.com/fourpost
   featured: false
+  - title: PersonalBlog Example
```
wardpeet: Can you give it a proper title? PersonalBlog example sounds more like a starter than a website. Are you sure you want to add it to the showcase?
```suggestion
  - title: PersonalBlog Example
```

97thjingba: Thank you for your correction, I have changed

PR #16342, opened 2019-08-02
chore(docs): Deploying to ZEIT Now has become even easier

leo:
## Description 

Now that deploying a Gatsby site to [ZEIT Now](http://zeit.co) is as easy as...

```
gatsby new gatsby-site
now gatsby-site
```

...I've adjusted the documentation to reflect that.

## Related Issues

No related issues.

PR #16398, opened 2019-08-05
docs: Replace Now with ZEIT Now

leo:
## Description

This is a follow-up PR to https://github.com/gatsbyjs/gatsby/pull/16342 and replaces "Now" with "ZEIT Now", as that's the new brand name.

## Related PRs

https://github.com/gatsbyjs/gatsby/pull/16342

### DIFF
--- a/docs/docs/deploying-to-zeit-now.md
+++ b/docs/docs/deploying-to-zeit-now.md
@@ -1,5 +1,5 @@
 ---
-title: Deploying to Now
+title: Deploying to ZEIT Now
 ---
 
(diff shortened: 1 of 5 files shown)


docs/docs/deploying-to-zeit-now.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Deploying to Now
+title: Deploying to ZEIT Now
```
sidharthachatterjee: We need to add a redirect if we're changing the URL

leo: Done!

PR #16414, opened 2019-08-06
chore(ci): revert "ci: add an (optional) approval step to build out www"

DSchau:
Reverts gatsbyjs/gatsby#16319

It can be triggered by users and isn't being used by _us_ currently. We can add this back later if needed.


DSchau:
Let's just do @wardpeet's suggestion of making it master only!

PR #16437, opened 2019-08-07
fix(docs): adds back some missing guides to sidebar

gillkyle:
## Description

I merged a guide #16391 without remembering we needed to add it back into the sidebar. 

I figured this would also be a good time to add back some guides that had been missing from the sidebar that could be useful into sections I thought they made sense. Happy to hear opinions about where they'd belong better but this makes them much more discoverable as guides.

### DIFF
--- a/www/src/data/sidebars/doc-links.yaml
+++ b/www/src/data/sidebars/doc-links.yaml
@@ -401,7 +401,7 @@
                   link: /docs/rendering-sidebar-navigation-dynamically/
                 - title: Client-only Routes & User Authentication
                   link: /docs/client-only-routes-and-user-authentication/
-                  breadcrumbTitle: Client routes & User auth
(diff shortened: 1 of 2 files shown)


www/src/data/sidebars/doc-links.yaml
```diff
@@ -395,10 +399,13 @@
                   link: /docs/centralizing-your-sites-navigation/
                 - title: Rendering Sidebar Navigation Dynamically*
                   link: /docs/rendering-sidebar-navigation-dynamically/
+                - title: Client-only Routes & User Authentication
+                  link: /docs/client-only-routes-and-user-authentication/
+                  breadcrumbTitle: Client routes & User auth
```
muescha:
```suggestion
                  breadcrumbTitle: Client Routes & User Auth
```

PR #16441, opened 2019-08-07
chore(gatsby): Improve typings for reporter

Everspace:
## Description

`panic`, `painicOnBuild`, and `error` should now match what is described on https://www.gatsbyjs.org/docs/node-api-helpers/#reporter


muescha:
maybe the `logMessageTyype` also not match exactly to the documentation?

Issue #15716, opened 2019-07-15
Lazy load for gatsby-remark-images


humphreybc:
## Summary

Out of the box, Gatsby does a good job of lazy loading images below the fold through `gatsby-image`. However, `gatsby-remark-images` doesn't appear to do this, which results in “Defer offscreen images” complaints from Lighthouse when testing a Markdown-generated page.

It would be nice if `gatsby-remark-images` added lazy loading so that images below the fold are not loaded until the user scrolls down the page. Bonus points if [support for the new Chrome native lazy loading API](https://twitter.com/chatsidhartha/status/1129118956193640448) was extended to `gatsby-remark-images` too.

## Previous work

There's been an attempt to add lazy-loading to `gatsby-remark-images` in [this plugin](https://www.npmjs.com/package/gatsby-remark-lazy-load). However it doesn't seem to work very well, and requires an external dependency in `lazysizes`. It would be better if this functionality was built into `gatsby-remark-images` like it is with `gatsby-image`.

Another relevant issue: https://github.com/gatsbyjs/gatsby/issues/2288

## Motivation

Improve the performance for Markdown-generated pages.

sidharthachatterjee:
Thank you for opening this @humphreybc 

We'd love to accept a pull request porting the lazy load implementation over to `gatsby-remark-images`

Marking this as `help wanted` for the moment!

humphreybc:
@sidharthachatterjee If you could drop a link to the relevant code in `gatsby-image`, I could take a look and see how difficult it would be to port to `gatsby-remark-images`. The Chrome stuff is just adding an attribute to the HTML tag, so that should be pretty easy.

0tho:
@humphreybc have a look at gatsby-image, it uses IntersectionObserver to 'defer image loading'.

[gatsby-image](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-image/src/index.js#L97)
[mdn IO](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API)

karlhorky:
not stale

sidharthachatterjee:
gatsby-remark-images@3.1.10 supports native lazy loading thanks to @gabefromutah 🥇

PR #16448, opened 2019-08-07
feat(gatsby-remark-images): Add support for native lazy loading

gabefromutah:
## Description
`gatsby-image` has already implemented this feature a while back. Now that Chrome 76 is available, let us also update `gatsby-remark-images` to support lazy-loading. This is one of my first PRs and I'm not 100% certain I've covered all the areas to make this happen. 

Fixes https://github.com/gatsbyjs/gatsby/issues/15716


packages/gatsby-remark-images/README.md
```diff
@@ -44,17 +44,18 @@ plugins: [
...
+| `loading`              | `string` | Set the browser's native lazy loading attribute. One of `lazy`, `eager` or `auto`. Defaults to `lazy`.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             |
```
gabefromutah: Not sure what is going on with the diff, I'll try to update the PR so that only this line shows as changed.

KyleAMathews:
Nice!

Perhaps we could default to `lazy`?

packages/gatsby-remark-images/README.md
```diff
@@ -55,6 +55,7 @@ plugins: [
...
+| `loading`              | `lazy`  | Set the browser's native lazy loading attribute. One of `lazy`, `eager` or `auto`.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 |
```
muescha: Should we note which is the default setting here?

gabefromutah: The second column in the table shows the default value. I was sort of confused by this since the second column in the `gastby-image` readme shows data type.

muescha: You are right. I muss this column

sidharthachatterjee:
Published in `gatsby-remark-images@3.1.10`

PR #16454, opened 2019-08-08
fix(tests): Remove unreliable timeout logic from development-runtime

### LINKED ISSUE #16432: Waiting for route change in Cypress is unreliable and hacky
See-also: #16473

## Description

Using the custom `waitForRouteChange` Cypress command is sometimes unreliable - especially when it comes to missing resources or 404 pages - which I believe is why `waitForAPIorTimeout` was created.

However, using this leaves us prone to other errors occurring which can no longer be spotted e.g. if the `onRouteUpdate` is not called correctly, and results in it being very hard to understand what is actually happening when debugging.

Notably in #13502, I've tried to remove the timeout logic in order to make Cypress tests more reliable across platforms, although this has resulted in me spending much more time than I initially spent on the main feature of that PR (improvements to the offline plugin) in order to fix the issues that this has exposed. As a result I think significantly larger changes are needed in a separate PR in order to have both reliable and clean code, rather than one or the other.

### Steps to reproduce

1. Replace `waitForAPIorTimeout('onRouteUpdate')` with `waitForRouteChange()` in various tests
2. Run said tests

### Expected result

The tests still pass

### Actual result

Various tests fail due to a variety of reasons - in some cases we can expect `waitForRouteChange()` to fail e.g. when we're intentionally deleting files, in which cases we should not call it in the first place. In others, Cypress sometimes reaches a timeout with no clear reason, sometimes inconsistently with each run.

### Environment

```
  System:
    OS: Linux 5.0 Pop!_OS 19.04
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
    Shell: 5.5.1 - /bin/zsh
  Binaries:
    Node: 12.8.0 - ~/.nvm/versions/node/v12.8.0/bin/node
    Yarn: 1.17.3 - ~/.nvm/versions/node/v12.8.0/bin/yarn
    npm: 6.10.3 - ~/.nvm/versions/node/v12.8.0/bin/npm
  Languages:
    Python: 2.7.16 - /usr/bin/python
  Browsers:
    Chrome: 76.0.3809.100
    Firefox: 68.0.1
```

davidbailey00:
## Description

Refactors Cypress tests to avoid the need of `waitForAPIorTimeout` in order to make testing more reliable.

## Related Issues

Fixes #16432


packages/gatsby-cypress/src/commands.js
```diff
@@ -6,13 +6,18 @@ Cypress.Commands.add(`getTestElement`, (selector, options = {}) =>
...
 
-const TIMEOUT = 9999
+const TIMEOUT = 5000
 
 Cypress.Commands.add(
   `waitForAPI`,
   { prevSubject: `optional` },
   (subject, api, { timeout = TIMEOUT } = {}) => {
-    cy.window().then({ timeout: timeout }, win => {
+    Cypress.log({
```
sidharthachatterjee: This is a great usability improvement!

PR #16486, opened 2019-08-09
feat(www): Add active style for TableOfContents

ImedAdel:
## Description

This PR adds an active style to the selected ToC item.

### DIFF
--- a/www/src/components/docs-table-of-contents.js
+++ b/www/src/components/docs-table-of-contents.js
@@ -35,6 +35,10 @@ function createItems(items, location) {
               },
             },
           }}
+          activeStyle={{
+            color: colors.link.color,
+            borderBottom: `1px solid ${colors.link.hoverBorder}`,
+          }}
           to={location.pathname + item.url}
         >
           {item.title}
--- a/www/src/components/docs-table-of-contents.js
+++ b/www/src/components/docs-table-of-contents.js
@@ -35,10 +35,16 @@ function createItems(items, location) {
               },
             },
           }}
-          activeStyle={{
-            color: colors.link.color,
-            borderBottom: `1px solid ${colors.link.hoverBorder}`,
-          }}
+          getProps={({ href, location }) =>
+            location.href.includes(href)
+              ? {
+                  style: {
+                    color: colors.link.color,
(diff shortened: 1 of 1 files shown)


ImedAdel:
@sidharthachatterjee I'm sorry, I didn't develop it locally, I just added the `activeStyle` part on GitHub. But this is how it _should_ look like.

Before:

2019-08-09 12_02_57-Gatsby Link API _ GatsbyJS

After:

2019-08-09 12_03_29-Gatsby Link API _ GatsbyJS

www/src/components/docs-table-of-contents.js
```diff
@@ -35,6 +35,10 @@ function createItems(items, location) {
               },
             },
           }}
+          activeStyle={{
```
sidharthachatterjee: Looks like this never gets applied because Link from reach doesn’t take hash into account when evaluating whether the current link is active 🤦🏻‍♂️

We could use getProps to check ourselves and set these styles

ImedAdel:
Oh, this is really unfortunate. I think we can close the PR for now?

sidharthachatterjee:
@ImedAdel Nah, I went ahead and fixed it! Let's get this in 🙌

ImedAdel:
That's amazing 😄 I didn't know about `getProps` though, it doesn't seem to be documented on https://www.gatsbyjs.org/docs/gatsby-link/

sidharthachatterjee:
@ImedAdel Want to PR that back to the docs mentioning this use case?

ImedAdel:
@sidharthachatterjee Done in #16511 --- I always see this `feat|doc|fix|...: something` naming convention, but I feel stupid for not knowing what to put inside the paranthesis. Do you have a list for that?

sidharthachatterjee:
@ImedAdel To be honest, we don’t have a list and mostly just set it based on discretion. My thought process is usually

- if it affects a particular package, put the package name
- if it affects the website, put www 
- if it affects other infrastructure (like gh-actions), put gh-actions

fk:
Sweeeeeet! Thank you @ImedAdel and @sidharthachatterjee!! 🙏 🤗

PR #16494, opened 2019-08-09
feat(gatsby-plugin-sass): Add ability to override sassRuleTest and sassRuleModulesTest

tkbky:
##  Description

This PR adds 2 new options to gatsby-plugin-sass
- `sassRuleTest` to configure regex for sass files.
- `sassRuleModulesTest` to configure regex for css modules files.

## Related Issues

Fixes [#15879](https://github.com/gatsbyjs/gatsby/issues/15879)

Issue #15879, opened 2019-07-18
[gatsby-plugin-sass] Add a configuration for sass plugin file regexes


abumalick:
## Summary
We want to use css modules with sass, but we don't want to add `.modules.` to the filename for every scss file in our app. Instead, we want to add `.global.` to global files, and normal scss files are modules by default.

https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-sass/src/gatsby-node.js#L37
https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-sass/src/gatsby-node.js#L20

### Basic example

add `sassRuleTest` and `sassRuleModulesTest` to the plugin options, that would accept a regex. If no regex is provided we use the current regexes.

### Motivation

- Sass modules files are usually a lot more than global files.
- We copied the plugin code and modified these two tests in our app since we switched to gatsby and want to switch back to official plugin.

Thank you for the excellent framework 💪

tkbky:
@abumalick 
I've made [a PR](https://github.com/gatsbyjs/gatsby/pull/16494) for this, feel free to take a look.

sidharthachatterjee:
Published in `gatsby-plugin-sass@2.1.6` thanks to @tkbky

PR #16514, opened 2019-08-10
docs(gatsby-plugin-sass): Improve example configuration 

abumalick:
## Description
- `sassRuleModulesTest` selects files that will be loaded via css modules. Having `global` in the filename is misleading as global is usually used for global style sheets.
- The configuration in the example have the same test for both rules. Which one will webpack use for loading `*.global.scss` files ?

## Related Issues
https://github.com/gatsbyjs/gatsby/issues/15879
https://github.com/gatsbyjs/gatsby/pull/16494


abumalick:
Thank you very much

PR #10861, opened 2019-01-06
feat(gatsby-plugin-postcss): added css-loader options

magicspon:
## Description

This commit allows users to define some css-loader options, such as `localIdentName` 

```javascript
{
	resolve: `gatsby-plugin-postcss`,
	options: {
		postCssPlugins: plugins,
		cssOptions: {
			localIdentName: 'mod-[hash:base64:8]'
		}
	}
}
``` 

`cssOptions` is the new bit.

## Related Issues

Some of my class names were getting stupidly long!

PR #16521, opened 2019-08-10
fix(gatsby-plugin-postcss): Always ensure postcss.config.js is loaded

mdreizin:
## Description

If you specify `cssLoaderOptions` then `postcss-loader` wound not picked up `postcss.config.js` config.

The `postcssOptions` options would contain `cssLoaderOptions` property (`{cssLoaderOptions: {...}}`) and `postcss-loader` gets it as is and does not use some built-in defaults.

This PR aims to fix that and also add some code-style adjustments to align with existing codebase.

## Related Issues

https://github.com/gatsbyjs/gatsby/pull/10861

### DIFF
--- a/packages/gatsby-plugin-postcss/src/__tests__/__snapshots__/gatsby-node.test.js.snap
+++ b/packages/gatsby-plugin-postcss/src/__tests__/__snapshots__/gatsby-node.test.js.snap
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`gatsby-plugin-postcss Stage: build-html / No options 1`] = `
+exports[`gatsby-plugin-postcss Stage: build-html / Css options 1`] = `
 [MockFunction] {
   "calls": Array [
     Array [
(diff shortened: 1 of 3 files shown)


mdreizin:
> Thank you for fixing this! Could you please add a test that asserts the correct behaviour here so that we don't see regressions in the future?

Yes, I could. I will do my best to add missing tests this week.

pieh:
Hey @mdreizin! Do you think you will have time to take a look at those tests? If not - that's not a problem - I can add test case for `cssLoaderOptions`. If you want to do this, but have any questions - don't hesitate to ask!

mdreizin:
@pieh No don’t. I will be able to write them only at the weekend. It would be nice if you could add missing tests.

sidharthachatterjee:
Thanks for the tests @pieh 💜

pieh:
Diff for snapshot file is messy, mostly because we didn't reset mocked calls before.

Here's somewhat easier to see snapshot diff between master and this PR - https://gist.github.com/pieh/6ee5420e34bd61ca36d4392cbd64171c

PR #16530, opened 2019-08-11
feat(gatsby): Display network url for GraphiQL when using -H

NickyMeuleman:
Been using `gatsby develop -H 0.0.0.0` on Windows a lot lately.

This adds a convenience link to take you directly to the gaphql IDE in the output when running that command.

### DIFF
--- a/packages/gatsby/src/commands/develop.js
+++ b/packages/gatsby/src/commands/develop.js
@@ -462,7 +462,21 @@ module.exports = async (program: any) => {
       }, an in-browser IDE, to explore your site's data and schema`
     )
     console.log()
-    console.log(`  ${urls.localUrlForTerminal}___graphql`)
+
+    if (urls.lanUrlForTerminal) {
+      console.log(
+        `  ${chalk.bold(`Local:`)}            ${
+          urls.localUrlForTerminal
+        }___graphql`
+      )
+      console.log(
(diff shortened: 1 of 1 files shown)


sidharthachatterjee:
@NickyMeuleman It's useful to display both the Local and Network urls for the site but is it really useful to do the same for GraphiQL? The Local url should continue to work on Windows from what I understand (on the same machine that is)

And the use case of running your build on one machine and viewing GraphiQL on another seems a little niche? 

I'd rather keep this out in effort to leave unnecessary noise out of the CLI output

NickyMeuleman:
Sadly, that local url doesn't work if `-H` is set to `0.0.0.0`.
Probably a frequent usecase.
From https://www.gatsbyjs.org/docs/gatsby-cli/
> Note: you can’t visit 0.0.0.0:8000 on Windows (but things will work using either localhost:8000 or the “On Your Network” URL on Windows)

When using the Windows subsystem for linux, localhost doesn't work either.

So a current usecase for Windows users is:
- `gatsby develop -H 0.0.0.0`
- Open the URL on the network for the regular site
- Add `/___graphql` to it and visit that.

Since the change makes the output **consistent** with the one right above it (and  only adds to it if `-H` is set), I don't consider it to be noise.

PR #16538, opened 2019-08-12
fix(gh-actions): Updating workflow file to new syntax

### LINKED ISSUE #16513: [gh-actions] Switch to the new workflow syntax
## Summary

As of this past week, GitHub announced a new version of [GitHub Actions with CI/CD](https://github.blog/2019-08-08-github-actions-now-supports-ci-cd/). The main thing is they are switching from HCL over to YAML for the workflow files. They have a are ending support for the HCL syntax at end of September, so we should eventually switch over to the new syntax.

You can either look at the docs for the [new workflow syntax ](https://help.github.com/en/articles/workflow-syntax-for-github-actions) and migrate manually or use their [migrate cli](https://github.com/actions/migrate) to convert it automatically.

Alongside this, we can as well drop the Docker requirement for our actions by using the new [JS Actions](https://help.github.com/en/articles/about-actions#javascript-actions) defined on this page, but a note is that to run this if using node modules is we need to bundle it into a single file and have that be the entrypoint for the action. This could be done with Webpack, Parcel, Rollup, or any other bundler.

lannonbr:
## Description

It seems someone accidentally pushed the button to migrate the repo over to the new version of GitHub Actions. That said, it disabled workflows that ran with the `.github/main.workflow` file. 

Due to this, I updated to the new syntax so we could get the actions back up and running. This PR used the [actions/migrate](https://github.com/actions/migrate) tool that the Actions team made to migrate from the `.workflow` file over to the new YAML based syntax.

## Related Issues

Fixes #16513


lannonbr:
Thanks Sid for the quick merge. I'm likely to bring up this and more info on the new beta of Actions at the maintainers meeting on Wednesday, but as it stands right now, everything should hopefully just work as it stands.

PR #16543, opened 2019-08-12
docs(visual-testing-with-storybook): Update static file directory

PsychedMagnet:
## Description

Change the static file directory in the storybook docs to "public". I don't know if this is a mistake, or the doc is outdated, but the static files from Gatsby, AFAIK go in a directory called public, in which there is a directory called static. This can lead to confusion (definitely confused me, maybe because I am new to Gatsby)

## Related Issues

None

### DIFF
--- a/docs/docs/visual-testing-with-storybook.md
+++ b/docs/docs/visual-testing-with-storybook.md
@@ -164,8 +164,8 @@ However, if you use `StaticQuery` or `useStaticQuery` in your project Storybook 
 ```json:title=package.json
 {
   "scripts": {
-    "storybook": "NODE_ENV=production start-storybook -s static",
-    "build-storybook": "NODE_ENV=production build-storybook -s static"
+    "storybook": "NODE_ENV=production start-storybook -s public",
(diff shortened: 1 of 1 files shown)


sidharthachatterjee:
@PsychedMagnet Just had a chat with @LekoArts and he mentioned that this might have to be `-s public,static` 

Would you like to test that out and PR it in?

Also, congratulations on your first contribution and enjoy the swag!

PsychedMagnet:
> @PsychedMagnet Just had a chat with @LekoArts and he mentioned that this might have to be `-s public,static`
> 
> Would you like to test that out and PR it in?
> 
> Also, congratulations on your first contribution and enjoy the swag!

Thank you! I will definitely enjoy the swag! 😄 

I tested it with both public and static, and it does work, but I am not sure both are needed. My understanding is that Storybook adds the folders listed to serve statically, but since folder "static" does not exist (at least not outside "public"), it just ignores it. I mean, it doesn't have a negative effect to have "static" listed too, but to my understanding, it is unnecessary.

PR #16546, opened 2019-08-12
fix(gatsby): improve hot-reloading for hooks

theKashey:
> Reincarnation of #16146

## Description
This is a follow up for #13713(RHL+Hooks), and https://github.com/gaearon/react-hot-loader/issues/1309 (RHL + Gatsby)

The problem: application breaking if `node_modules` exists in the `src` dir. There is no question "why", the question is "why".

### Fix

In short - there is no need to apply `react-hot-loader/webpack` for all `node_modules`, only for `react-dom` to land the patch, and that's all the change. 
+ Plus bump `react-hot-loader` to fix the problems we found in the previous PR with cyclic dependencies, and make it React 16.9 compatible.

### Outcome:
The problem is fixed and everything is a bit faster.

## Related Issues

- #13713 introduced `react-hot-loader/webpack`
- https://github.com/gaearon/react-hot-loader/issues/1309 discovered the problem.


pieh:
Published `gatsby@2.13.60`

sidharthachatterjee:
Thank you so much for your incredible here @theKashey 💜

theKashey:
You also helped to surface a very nasty issue. So all kudos go for you.

Issue #15493, opened 2019-07-07
[gatsby-theme-notes] showing improper breadcrumbs when basePath is set to default value.


jakewies:
## Description

I’ve been playing around with `gatsby-theme-notes` and noticed a small issue related to breadcrumbs inside of subdirectories of the `contentPath` (i.e. `/content/notes`).

### Steps to reproduce

1. Bootstrap a quick example with `gatsby-theme-notes`:

```
gatsby new my-themed-notes https://github.com/gatsbyjs/gatsby-starter-notes-theme

cd my-themed-notes

yarn develop
```

No custom configuration is needed, as the issue only appears under conditions where the user of `gatsby-theme-notes` has not provided a [custom `basePath` option](https://github.com/gatsbyjs/gatsby/tree/master/themes/gatsby-theme-notes#options).

2. In your browser, navigate to `localhost:8000/example-dir`.

### Expected result

The breadcrumbs at the top of the page should look like this:

```
~ / example-dir
```

### Actual result

The breadcrumbs at the top of the page look like this:

```
~ / / example-dir
```

### Environment

```
❯ gatsby info --clipboard

  System:
    OS: macOS 10.14.5
    CPU: (8) x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 11.14.0 - /usr/local/bin/node
    Yarn: 1.15.2 - /usr/local/bin/yarn
    npm: 6.9.2 - ~/.npm-packages/bin/npm
  Languages:
    Python: 2.7.16 - /usr/local/bin/python
  Browsers:
    Chrome: 77.0.3833.0
    Firefox: 65.0.1
    Safari: 12.1.1
  npmPackages:
    gatsby: ^2.13.6 => 2.13.6
    gatsby-theme-notes: ^1.0.0 => 1.0.0
  npmGlobalPackages:
    gatsby-cli: 2.7.3
```

### Debugging

I’ve spent a little bit of time debugging the issue, and I’ve traced it back to 2 separate parts of the same file:

1. [Rendering the basePath text in the breadcrumb](https://github.com/gatsbyjs/gatsby/blob/master/themes/gatsby-theme-notes/src/components/breadcrumbs.js#L26):

```js
basePath.replace(/^\//, ``)
```

[...]

PR #16556, opened 2019-08-12
fix(gatsby-theme-notes): Render correct breadcrumb path based on basePath

jakewies:
## Description

In scenarios where the `basePath` option is not provided, it defaults to `"/"`. When this happens, the breadcrumbs render as such:

```
~ / / example-dir
```

This is due to a regex replacement that causes the base breadcrumb text to be an empty string:

```js
basePath.replace(/^\//, ``)
```

The initial breadcrumb divider should only render if there is a custom basePath (or more importantly, if the basePath has a text length). This results in proper breadcrumb text:

```
~ / example-dir
```

## Related Issues

Closes https://github.com/gatsbyjs/gatsby/issues/15493


jakewies:
Thanks @sidharthachatterjee !

PR #16561, opened 2019-08-12
feat(gatsby-plugin-schema-snapshot): First version

stefanprobst:
This is a thin wrapper plugin over the schema snapshotting functionality added in  #16291


packages/gatsby-plugin-schema-snapshot/gatsby-node.js
```diff
@@ -0,0 +1,34 @@
+const fs = require(`fs`)
+const path = require(`path`)
+
+exports.createSchemaCustomization = ({ actions, reporter }, options = {}) => {
+  const { createTypes, printTypeDefinitions } = actions
+
+  if (!printTypeDefinitions) {
+    reporter.error(
+      `\`gatsby-plugin-schema-snapshot\` needs Gatsby v2.13.55 or above.`
+    )
```
pieh:
```suggestion
    )
    return
```
seems like early bail make sense here given we log problem and we know rest won't work?

stefanprobst: oops yes sure

sidharthachatterjee:
Published in `gatsby-plugin-schema-snapshot@1.0.0`

PR #16571, opened 2019-08-13
chore(gatsby-link): Minor refactor

jamesgeorge007:
## Description
Minor refactor to have a more concise version.

## Related Issues
N/A

### DIFF
--- a/packages/gatsby-link/src/index.js
+++ b/packages/gatsby-link/src/index.js
@@ -178,6 +178,11 @@ GatsbyLink.propTypes = {
   replace: PropTypes.bool,
 }
 
+const showWarningMessage = action =>
+  console.warn(
+    `The "${action}" method is now deprecated and will be removed in Gatsby v3. Please use "navigate" instead.`
+  )
+
 export default React.forwardRef((props, ref) => (
   <GatsbyLink innerRef={ref} {...props} />
 ))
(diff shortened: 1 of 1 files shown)


packages/gatsby-link/src/index.js
```diff
@@ -178,6 +178,11 @@ GatsbyLink.propTypes = {
   replace: PropTypes.bool,
 }
 
+const showWarningMessage = action =>
```
sidharthachatterjee: Want to call the function showDeprecationWarning and accept functionName, altFunctionName and version?

jamesgeorge007:
cc @sidharthachatterjee

packages/gatsby-link/src/index.js
sidharthachatterjee: Would've preferred to take `altFunctionName` and `version` too as parameters too (so we could reuse this function) but this is valuable as well so thank you!

jamesgeorge007:
@sidharthachatterjee should I shoot up yet another PR with the changes as you suggested above :question:

sidharthachatterjee:
@jamesgeorge007 Sure! That would be awesome 🥇

jamesgeorge007:
@sidharthachatterjee would also love to know your thoughts on [Mevn-CLI](https://github.com/madlabsinc/mevn-cli). Consider, leaving a :star:

PR #16573, opened 2019-08-13
fix(gatsby): adjust regex to handle MODULE_NOT_FOUND errors thrown by Bazel

bweston92:
## Description

I have been trying to build my Gatsby site with Bazel. However `rules_nodejs` patches the loader script. They return an error that is not caught by Gatsby correctly. Upon further investigation Bazel is throwing an error that is return with a code `MODULE_NOT_FOUND` which is actually the same as Node's own loader script.

--

The MODULE_NOT_FOUND error is provided by a loader script when the module is not found.

References to loaders:
- **Bazel rules_nodejs** - https://github.com/bazelbuild/rules_nodejs/blob/968b3a35cf6e4e3bf20fae70f7461514e28f24e6/internal/node/node_loader.js#L473
- **Nodes internal** - https://github.com/nodejs/node/blob/a49b20d3245dd2a4d890e28582f3c013c07c3136/lib/internal/modules/cjs/loader.js#L268

### DIFF
--- a/packages/gatsby/src/utils/__tests__/test-require-error.js
+++ b/packages/gatsby/src/utils/__tests__/test-require-error.js
@@ -37,4 +37,36 @@ describe(`test-require-error`, () => {
       )
     }
   })
+
+  describe(`handles error message thrown by Bazel`, () => {
+    it(`detects require errors`, () => {
+      const bazelModuleNotFoundError = new Error(`Error: //:build_bin cannot find module './fixtures/module-does-not-exist' required by '/private/var/tmp/_bazel_misiek/eba1803983a26276494495d851e478a5/execroot/com_github_bweston92_debug_gatsby_bazel_rules_nodejs/bazel-out/darwin-fastbuild/bin/build.runfiles/com_github_bweston92_debug_gatsby_bazel_rules_nodejs/node_modules/gatsby/dist/bootstrap/get-config-file.js'
+      looked in:
+        built-in, relative, absolute, nested node_modules - Error: Cannot find module './fixtures/module-does-not-exist'
+        runfiles - Error: Cannot find module '/private/var/tmp/_bazel_misiek/eba1803983a26276494495d851e478a5/execroot/com_github_bweston92_debug_gatsby_bazel_rules_nodejs/bazel-out/darwin-fastbuild/bin/build.runfiles/fixtures/module-does-not-exist'
+        node_modules attribute (com_github_bweston92_debug_gatsby_bazel_rules_nodejs/node_modules) - Error: Cannot find module '/private/var/tmp/_bazel_misiek/eba1803983a26276494495d851e478a5/execroot/com_github_bweston92_debug_gatsby_bazel_rules_nodejs/bazel-out/darwin-fastbuild/bin/build.runfiles/com_github_bweston92_debug_gatsby_bazel_rules_nodejs/node_modules/fixtures/module-does-not-exist'`)
+
+      expect(
+        testRequireError(
+          `./fixtures/module-does-not-exist`,
+          bazelModuleNotFoundError
+        )
+      ).toEqual(true)
+    })
+
+    it(`detects require errors`, () => {
+      const bazelModuleNotFoundError = new Error(`Error: //:build_bin cannot find module 'cheese' required by '/private/var/tmp/_bazel_misiek/eba1803983a26276494495d851e478a5/execroot/com_github_bweston92_debug_gatsby_bazel_rules_nodejs/bazel-out/darwin-fastbuild/bin/build.runfiles/com_github_bweston92_debug_gatsby_bazel_rules_nodejs/node_modules/gatsby/dist/bootstrap/fixtures/bad-module-require.js'
+      looked in:
+        built-in, relative, absolute, nested node_modules - Error: Cannot find module 'cheese'
+        runfiles - Error: Cannot find module '/private/var/tmp/_bazel_misiek/eba1803983a26276494495d851e478a5/execroot/com_github_bweston92_debug_gatsby_bazel_rules_nodejs/bazel-out/darwin-fastbuild/bin/build.runfiles/cheese'
+        node_modules attribute (com_github_bweston92_debug_gatsby_bazel_rules_nodejs/node_modules) - Error: Cannot find module '/private/var/tmp/_bazel_misiek/eba1803983a26276494495d851e478a5/execroot/com_github_bweston92_debug_gatsby_bazel_rules_nodejs/bazel-out/darwin-fastbuild/bin/build.runfiles/com_github_bweston92_debug_gatsby_bazel_rules_nodejs/node_modules/cheese'`)
+
+      expect(
+        testRequireError(
+          `./fixtures/bad-module-require`,
+          bazelModuleNotFoundError
+        )
+      ).toEqual(false)
+    })
+  })
 })
--- a/packages/gatsby/src/utils/test-require-error.js
+++ b/packages/gatsby/src/utils/test-require-error.js
@@ -5,7 +5,8 @@ export default (moduleName, err) => {
   // dependency tree rules but the requested file doesn't exist
   if (
     err.code === `QUALIFIED_PATH_RESOLUTION_FAILED` ||
-    err.pnpCode === `QUALIFIED_PATH_RESOLUTION_FAILED`
+    err.pnpCode === `QUALIFIED_PATH_RESOLUTION_FAILED` ||
+    err.code === `MODULE_NOT_FOUND`
   ) {
     return true
   }
--- a/packages/gatsby/src/utils/test-require-error.js
+++ b/packages/gatsby/src/utils/test-require-error.js
@@ -5,14 +5,13 @@ export default (moduleName, err) => {
   // dependency tree rules but the requested file doesn't exist
   if (
     err.code === `QUALIFIED_PATH_RESOLUTION_FAILED` ||
-    err.pnpCode === `QUALIFIED_PATH_RESOLUTION_FAILED` ||
-    err.code === `MODULE_NOT_FOUND`
+    err.pnpCode === `QUALIFIED_PATH_RESOLUTION_FAILED`
   ) {
     return true
   }
 
   const regex = new RegExp(
-    `Error: Cannot find module\\s.${moduleName.replace(
+    `Error:\\s(\\S+\\s)?[Cc]annot find module\\s.${moduleName.replace(
       /[-/\\^$*+?.()|[\]{}]/g,
       `\\$&`
     )}`


pieh:
Hey @bweston92!

We do need to make sure that not found module is the one that is being directly tested (that's why in regex below `moduleName` is explicitly checked). In failing test we try to import module, which in turn tries to import another one that doesn't exist. 

Existing condition that are you changing is for Yarn PnP and it has same issue I described (but it's there to enable PnP at all).

I pretty much hate that we do regex test on error message - but maybe we could modify existing regex to cover Bazel error message?

bweston92:
@pieh sure, the Bazel message is pretty verbose. What parts of the message do we need to guarantee? Example error message:

```
  Error: //:build_bin cannot find module '/private/var/tmp/_bazel_bweston/8d1f338118b584267486f5711fe24c3f/execroot/com_github_legalweb_commercialfinancegroup_co_uk/bazel-out/darwin-fastbuild/bin/build.runfiles/com_github_legalweb_commercialfinancegroup_co_uk/node_mo  dules/gatsby-plugin-manifest/gatsby-config' required by '/private/var/tmp/_bazel_bweston/8d1f338118b584267486f5711fe24c3f/execroot/com_github_legalweb_commercialfinancegroup_co_uk/bazel-out/darwin-fastbuild/bin/build.runfiles/com_github_legalweb_commercialfinancegr  oup_co_uk/node_modules/gatsby/dist/bootstrap/get-config-file.js'
    looked in:
      built-in, relative, absolute, nested node_modules - Error: Cannot find module '/private/var/tmp/_bazel_bweston/8d1f338118b584267486f5711fe24c3f/execroot/com_github_legalweb_commercialfinancegroup_co_uk/bazel-out/darwin-fastbuild/bin/build.runfiles/com_github_le  galweb_commercialfinancegroup_co_uk/node_modules/gatsby-plugin-manifest/gatsby-config'
      runfiles - Error: Cannot find module '/private/var/tmp/_bazel_bweston/8d1f338118b584267486f5711fe24c3f/execroot/com_github_legalweb_commercialfinancegroup_co_uk/bazel-out/darwin-fastbuild/bin/build.runfiles/private/var/tmp/_bazel_bweston/8d1f338118b584267486f57  11fe24c3f/execroot/com_github_legalweb_commercialfinancegroup_co_uk/bazel-out/darwin-fastbuild/bin/build.runfiles/com_github_legalweb_commercialfinancegroup_co_uk/node_modules/gatsby-plugin-manifest/gatsby-config'
      node_modules attribute (com_github_legalweb_commercialfinancegroup_co_uk/node_modules) - Error: Cannot find module '/private/var/tmp/_bazel_bweston/8d1f338118b584267486f5711fe24c3f/execroot/com_github_legalweb_commercialfinancegroup_co_uk/bazel-out/darwin-fastb  uild/bin/build.runfiles/com_github_legalweb_commercialfinancegroup_co_uk/node_modules/private/var/tmp/_bazel_bweston/8d1f338118b584267486f5711fe24c3f/execroot/com_github_legalweb_commercialfinancegroup_co_uk/bazel-out/darwin-fastbuild/bin/build.runfiles/com_github_  legalweb_commercialfinancegroup_co_uk/node_modules/gatsby-plugin-manifest/gatsby-config'
```

pieh:
I think we would need just `cannot find module '${request}'` part of error message (from https://github.com/bazelbuild/rules_nodejs/blob/968b3a35cf6e4e3bf20fae70f7461514e28f24e6/internal/node/node_loader.js#L470-L472 )

where request _should_ match `moduleName` argument of our function

bweston92:
It actually looks like we can test for `node_modules/gatsby/dist/bootstrap/get-config-file.js` because that is the parent it is on about which is obviously us calling it and would make this super simple.

pieh:
> It actually looks like we can test for `node_modules/gatsby/dist/bootstrap/get-config-file.js` because that is the parent it is on about which is obviously us calling it and would make this super simple.
> 
> Is that OK with you?

Right, but we don't exactly care about parent/caller - we also use `test-require-error` in at least one different place (which will have different parent)

We want to check what module actually doesn't exist.

Consider scenario:

```js
// file a.js
require(`./b.js`)
```

```js
// file b.js
require(`./doesnt-exist.js`)
```

In file `a.js` we will get MODULE_NOT_FOUND - but `a.js` exist - it's one of `a.js` imports that doesn't exist.

in `test-require-error` file, we want to check if not existing module is the one that was directly imported.

bweston92:
@pieh if the error message is from Bazel the message contains a statement regarding the parent being `/node_modules/gatsby/dist/bootstrap/get-config-file.js` there is never going to be a scenario where the error isn't going to be about the direct module you're importing.

We could try and reconstruct the exact module path to match on but seems complicated for example: `cannot find module '/private/var/tmp/_bazel_bweston/8d1f338118b584267486f5711fe24c3f/execroot/com_github_legalweb_commercialfinancegroup_co_uk/bazel-out/darwin-fastbuild/bin/build.runfiles/com_github_legalweb_commercialfinancegroup_co_uk/node_mo  dules/gatsby-plugin-manifest/gatsby-config'`

pieh:
> @pieh if the error message is from Bazel the message contains a statement regarding the parent being `/node_modules/gatsby/dist/bootstrap/get-config-file.js` there is never going to be a scenario where the error isn't going to be about the direct module you're importing.

Is bazel behaving differently than vanilla Node.js? I'm pretty confused here.

Are you able to share bazel setup (at least parts related to running gatsby)? I'd love to do some local testing on this

pieh:
> We could try and reconstruct the exact module path to match on but seems complicated for example: `cannot find module '/private/var/tmp/_bazel_bweston/8d1f338118b584267486f5711fe24c3f/execroot/com_github_legalweb_commercialfinancegroup_co_uk/bazel-out/darwin-fastbuild/bin/build.runfiles/com_github_legalweb_commercialfinancegroup_co_uk/node_mo dules/gatsby-plugin-manifest/gatsby-config'`

I think in case of `gatsby-config` (and probably others) - it's actually how we try to import them (by doing `require(<absolutePathToPotentialGatsbyConfig>)` - so `moduleName` should actually be full absolute path

bweston92:
@pieh oh sorry for my misunderstanding I though rules_nodejs was adding the full path making it harder to regex match.

I do have a reproducable git repo here: https://github.com/bweston92/debug-gatsby-bazel-rules-nodejs

pieh:
All righty, just checked repo out

And here's `moduleName` passed to `test-require-error`:

```
/private/var/tmp/_bazel_misiek/eba1803983a26276494495d851e478a5/execroot/com_github_bweston92_debug_gatsby_bazel_rules_nodejs/bazel-out/darwin-fastbuild/bin/build.runfiles/com_github_bweston92_debug_gatsby_bazel_rules_nodejs/node_modules/gatsby-plugin-react-helmet/gatsby-config
```

And error message is

```
Error: //:build_bin cannot find module '/private/var/tmp/_bazel_misiek/eba1803983a26276494495d851e478a5/execroot/com_gi  thub_bweston92_debug_gatsby_bazel_rules_nodejs/bazel-out/darwin-fastbuild/bin/build.runfiles/com_github_bweston92_debug  _gatsby_bazel_rules_nodejs/node_modules/gatsby-plugin-react-helmet/gatsby-config' required by '/private/var/tmp/_bazel_  misiek/eba1803983a26276494495d851e478a5/execroot/com_github_bweston92_debug_gatsby_bazel_rules_nodejs/bazel-out/darwin-  fastbuild/bin/build.runfiles/com_github_bweston92_debug_gatsby_bazel_rules_nodejs/node_modules/gatsby/dist/bootstrap/ge  t-config-file.js'
    looked in:
      built-in, relative, absolute, nested node_modules - Error: Cannot find module '/private/var/tmp/_bazel_misiek/eba18  03983a26276494495d851e478a5/execroot/com_github_bweston92_debug_gatsby_bazel_rules_nodejs/bazel-out/darwin-fastbuild/bi  n/build.runfiles/com_github_bweston92_debug_gatsby_bazel_rules_nodejs/node_modules/gatsby-plugin-react-helmet/gatsby-co  nfig'
      runfiles - Error: Cannot find module '/private/var/tmp/_bazel_misiek/eba1803983a26276494495d851e478a5/execroot/com_  github_bweston92_debug_gatsby_bazel_rules_nodejs/bazel-out/darwin-fastbuild/bin/build.runfiles/private/var/tmp/_bazel_m  isiek/eba1803983a26276494495d851e478a5/execroot/com_github_bweston92_debug_gatsby_bazel_rules_nodejs/bazel-out/darwin-f  astbuild/bin/build.runfiles/com_github_bweston92_debug_gatsby_bazel_rules_nodejs/node_modules/gatsby-plugin-react-helme  t/gatsby-config'
      node_modules attribute (com_github_bweston92_debug_gatsby_bazel_rules_nodejs/node_modules) - Error: Cannot find mod  ule '/private/var/tmp/_bazel_misiek/eba1803983a26276494495d851e478a5/execroot/com_github_bweston92_debug_gatsby_bazel_r  ules_nodejs/bazel-out/darwin-fastbuild/bin/build.runfiles/com_github_bweston92_debug_gatsby_bazel_rules_nodejs/node_mod  ules/private/var/tmp/_bazel_misiek/eba1803983a26276494495d851e478a5/execroot/com_github_bweston92_debug_gatsby_bazel_ru  les_nodejs/bazel-out/darwin-fastbuild/bin/build.runfiles/com_github_bweston92_debug_gatsby_bazel_rules_nodejs/node_modu  les/gatsby-plugin-react-helmet/gatsby-config'
```

So we should be able to construct regex just fine as `moduleName` matches that path in error message

alexeagle:
I would rather fix how bazel invokes npm so that the error message matches. We should strive to have bazel be a drop-in replacement for any other build tool like gulp.
Put another way, if gatsby needs a change, then lots of other tools probably do too.

Do we just need to change the first line in that error message?

pieh:
> I would rather fix how bazel invokes npm so that the error message matches. We should strive to have bazel be a drop-in replacement for any other build tool like gulp.
> Put another way, if gatsby needs a change, then lots of other tools probably do too.

To be fair here, we really shouldn't try to parse error message and use any regexes on it and instead rely on more structured metadata in errors. But there are no structured metadata on errors (at least in Node 8 and 10 - i see there's more added in Node 11 or 12), so we fallback to this crude regex.

I do think that making that error in Bazel to resemble Node.js error does make sense, but I have little to none context on Bazel - so this is up to you if you want to take action to change error message. (Also in same way we rely on Node.js error message here I wonder if someone else wouldn't rely on Bazel's error - sometimes hard to decide if something would be breaking change or not ;) )

> Do we just need to change the first line in that error message?

For our usage yeah, we parse whole message, but really our test only cares about first line

I'm also fine with adding regex here to match Bazel's MODULE_NOT_FOUND error message (possibly just temporarily, if Bazel was to change emitted error)

bweston92:
Bazel has its reasons about everything and I’m no authority to change it.

The MODULE_NOT_FOUND follows Nodes internal loader script too so
technically they’re compliant with Node.

It is the fact Gatsby cares to treat plugins dependency resolution failures
the same as a plugin which I find odd but I’m sure there is some reason.

I will propose it to the rules_nodejs repository but likely they’ll not
want to change the message.

pieh:
> It is the fact Gatsby cares to treat plugins dependency resolution failures the same as a plugin which I find odd but I’m sure there is some reason.

Let me step back from details of `test-require-error` and explain what it's actually used for:

Plugins, themes or even main site itself can have some optional files - `gatsby-config.js`, `gatsby-node.js`, etc. We don't want to fail completely when trying to import those files if they don't exist. So we wrap those optional imports in try/catch and if error is thrown we examine error to see if it was thrown because module we try to import doesn't exist or if there is something else causing error during import (in this case we want to fail).

Now after some of discussion here I started to think if we can redesign how we do those checks. First thought was to separate module existence (i.e. use `require.resolve` + `fs.exists`) from checking if there are other errors that can happen as result of module import. But this flow probably won't work for Yarn 2 PnP implementation, where modules aren't guaranteed to be actual files (Yarn 2 can store packages as tarballs, so `fs.exists` on single module from some package wouldn't probably work and we would need yarn 2 specific checks for module existence which make refactor questionable as we trade one tool specific code for another.

> I will propose it to the rules_nodejs repository but likely they’ll not want to change the message.

Alex is one of maintainers of `rules_nodejs` - https://github.com/bazelbuild/rules_nodejs/blob/master/CODEOWNERS, so I think he has pretty good authority when it comes to proposing changes there ;)

bweston92:
Sounds good to me, anything I can do to help?

I’m not really aware of implementation specifics with Node/Yarn/npm.

pieh:
I think that for now we should still add regex that would pick up on error from current version of Bazel. Even if error message will change in future version of Bazel, this way we can support previous versions before that change.

pieh:
I'll pick it up, add or adjust regex and add some test to not regress in the future.

bweston92:
Thanks I’m going to submit a PR shortly to rules_nodejs if they’ve not
already done it as they was happy to move to Node’s error message.

pieh:
I have regex working (will push when I add test), but there seems to be more work needed

In your reproduction I had to make some changes to BUILD:
```diff
diff --git a/BUILD b/BUILD
index 8c0a9736..50be333f 100644
--- a/BUILD
+++ b/BUILD
@@ -3,8 +3,10 @@ load("@build_bazel_rules_nodejs//:defs.bzl", "nodejs_binary")
 filegroup(
     name = "node_modules",
     srcs = glob([
+        "node_modules/sharp/**/*",
+        "node_modules/pngquant-bin/**/*",
         "node_modules/**/*.js",
-        "node_modules/**/*.json",
+        "node_modules/**/*.json"
     ]),
     visibility = ["//visibility:public"],
 )
```
as there were some things missing in "Bazel'ed node_modules", but ultimately It still fails at building webpack bundles stage:

```
 ERROR #98123  WEBPACK

Generating JavaScript bundles failed

/Users/misiek/test/i16573/node_modules/gatsby/cache-dir/production-app.js: Unexpected token (49:4)

  47 |   // Remove this in v3
  48 |   const RouteHandler = props => (
> 49 |     <BaseContext.Provider
     |     ^
  50 |       value={{
  51 |         baseuri: `/`,
  52 |         basepath: `/`,

File: ../../../../../../../../../../../../Users/misiek/test/i16573/node_modules/gatsby/cache-dir/production-app.js:49:4
```

this seems to suggest that our webpack rules when transpiling aren't working with this setup. I think there are at least 2 things to research here:
 - maybe `BUILD` config in your project needs some changes (I have absolutely no context on Bazel, so this might be dead alley)
 - we might need adjustments to webpack rules to handle setups like this - I'll do some more checks after I finish writing test for `test-require-error` changes

packages/gatsby/src/utils/__tests__/test-require-error.js
```diff
@@ -37,4 +37,36 @@ describe(`test-require-error`, () => {
...
+      looked in:
+        built-in, relative, absolute, nested node_modules - Error: Cannot find module './fixtures/module-does-not-exist'
+        runfiles - Error: Cannot find module '/private/var/tmp/_bazel_misiek/eba1803983a26276494495d851e478a5/execroot/com_github_bweston92_debug_gatsby_bazel_rules_nodejs/bazel-out/darwin-fastbuild/bin/build.runfiles/fixtures/module-does-not-exist'
+        node_modules attribute (com_github_bweston92_debug_gatsby_bazel_rules_nodejs/node_modules) - Error: Cannot find module '/private/var/tmp/_bazel_misiek/eba1803983a26276494495d851e478a5/execroot/com_github_bweston92_debug_gatsby_bazel_rules_nodejs/bazel-out/darwin-fastbuild/bin/build.runfiles/com_github_bweston92_debug_gatsby_bazel_rules_nodejs/node_modules/fixtures/module-does-not-exist'`)
```
pieh: I feel silly about inlining this, but spinning env with Bazel would be much more costly than hard coding those errors

pieh:
In any case - I'll merge this PR as-is now, we can work on adding better support in follow up PRs

packages/gatsby/src/utils/__tests__/test-require-error.js
bweston92: I’m just happy it will allow others to use Gatsby with Bazel :) thanks @pieh !

PR #16574, opened 2019-08-13
feat(gatsby-remark-images): Add markdownCaptions option

### LINKED ISSUE #14474: [gatsby-remark-images] When using `showCaption`, parse the caption as markdown

mathieudutour:
## Description

This PR adds an option to parse image captions as markdown. To do so, the gatsby remark plugin needs access to the "compiler", eg. parsing a markdown string and generating an html string from it.
So the PR also contains a change to `gatsby-remark-transformer` to pass that down to plugins.

## Related Issues

Fixes #14474

## Caveat

`gatsby-plugin-mdx` doesn't pass down the compiler yet, so the option is going to be ignored when used with the mdx plugin. This will be address in a follow-up PR because there isn't any great way to it yet.


pieh:
I do wonder if there are any instances where user wouldn't want to parse captions as markdown (i.e. if we can get away without adding extra config option, and just parse it every time)?

Question of backward compat here - if raw text (without any markdown formatting) using parsing and generation produces same thing as before?

mathieudutour:
Indeed, for backward compatibility we need a new option.

packages/gatsby-remark-images/package.json
```diff
@@ -23,7 +23,9 @@
     "@babel/cli": "^7.0.0",
     "@babel/core": "^7.0.0",
     "babel-preset-gatsby-package": "^0.2.2",
-    "cross-env": "^5.1.4"
+    "cross-env": "^5.1.4",
+    "hast-util-to-html": "^6.0.2",
+    "mdast-util-to-hast": "^6.0.1"
```
pieh: Just a note - those 2 should go to `dependencies` not `devDependencies`

mathieudutour: no it's only used by the tests

pieh: Ah, my bad - right - you need those to "mock" gatsby-transformer-remark passing `compiler` thingy.

dcastil:
Thanks for this awesome feature @mathieudutour!

> `gatsby-plugin-mdx` doesn't pass down the compiler yet, so the option is going to be ignored when used with the mdx plugin. This will be address in a follow-up PR because there isn't any great way to it yet.

Is there a PR for mdx support already? Would love to subscribe to it!

CanRau:
Hey @mathieudutour that's a super useful feature 👏 
Any update on `gatsby-plugin-mdx` support?

Issue #16581, opened 2019-08-13
gatsby-image opacity not configurable


tkvw:
Currently revealed opacity is configured as 1 and unrevealed as 0 and there's no way to override this. I would like to use `0.8` and a backgroundColor of `#000` to darken the image a bit. 

Something like this, would be ideal: 
```
<Img revealedStyle={{opacity: 0.8}} /> // Having revealedStyle default to {opacity:1}
```
I can PR if requested.

pieh:
Hey @tkbky! We won't be adding prop like that to `gatsby-image` as the usecase you presented is quite niche and in our opinion doesn't warrant adding more surface area to `gatsby-image`.

For cases like that I would recommend vendoring `gatsby-image` (copy code from https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-image/src/index.js to your project and adjust it locally and use that instead of `gatsby-image`).

tkbky:
@pieh You have mentioned the wrong person. 😬 

@tkvw FYI

gabrilator:
+1 to @tkvw's request, I don't think darkening an image should be considered a niche case!!

joelbowen:
darkening an image is a nice feature for dark-mode ([CSS Tricks article on the subject](https://css-tricks.com/dark-modes-with-css/#its-not-just-setting-dark-backgrounds-with-light-text)), or creating hover effects for interactions.

This would be a great feature to have.

PR #16596, opened 2019-08-13
Added support to override style on revealed image

nasirhjafri:
## Description

Currently revealed opacity is configured as 1 and unrevealed as 0 and there's no way to override this. I would like to use 0.8 and a backgroundColor of #000 to darken the image a bit.

Added support to override style when the image is revealed. 

```
<Img revealedStyle={{opacity: 0.8}} /> // Having revealedStyle default to {opacity:1}
```

## Related Issues

Addresses #16581

### DIFF
--- a/packages/gatsby-image/src/index.js
+++ b/packages/gatsby-image/src/index.js
@@ -381,6 +381,7 @@ class Image extends React.Component {
       itemProp,
       loading,
       draggable,
+      revealedStyle = {},
     } = convertProps(this.props)
 
     const shouldReveal = this.state.fadeIn === false || this.state.imgLoaded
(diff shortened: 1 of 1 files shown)


pieh:
> This seems like a fairly _niche_ use case. I'd want to avoid adding props to `Img` unless they're a fair use case.
> 
> Thoughts @gatsbyjs/core?

Yeah, I very much agree. 

@nasirhjafri We really appreciate you sending this over, but the change you’ve proposed is not going to be accepted because it adds extra public API (prop) and extra maintenance burden for not very common usecase.

I'll also comment on issue that this PR is addressing (https://github.com/gatsbyjs/gatsby/issues/16581) on how this can be implemented in userland, without adding more surface area to `gatsby-image`.

PR #16597, opened 2019-08-13
docs(gatsby-plugin-offline): Clarify how to pass in options

jkjustjoshing:
I wasn't initially clear on where to put the `options` object - this should clarify that.

### DIFF
--- a/packages/gatsby-plugin-offline/README.md
+++ b/packages/gatsby-plugin-offline/README.md
@@ -21,10 +21,10 @@ plugins: [`gatsby-plugin-offline`]
 
 ## Overriding options
 
-When adding this plugin to your `gatsby-config.js`, you can pass in options to
+When adding this plugin to your `gatsby-config.js`, you can pass in options (via the `options` key) to
 override the default [Workbox](https://developers.google.com/web/tools/workbox/modules/workbox-build) config.
 
-The default config is as follows. Warning: you can break the offline support by
+The default config is as follows. Warning: You can break the offline support by
 changing these options, so tread carefully.
 
 ```javascript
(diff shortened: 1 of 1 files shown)


packages/gatsby-plugin-offline/README.md
```diff
@@ -62,6 +62,14 @@ const options = {
...
   clientsClaim: true,
 }
+
+// In your gatsby-config.js
+plugins: [
+  {
+    resolve: `gatsby-plugin-offline`,
+    options: options
+  }
+]
```
pieh: How about this whole codefence be something like this:

```javascript
// In your gatsby-config.js
module.exports = {
  plugins: [
    {
      resolve: `gatsby-plugin-offline`,
      options: {
        importWorkboxFrom: `local`,
        globDirectory: rootDir,
        globPatterns,
        modifyUrlPrefix: {
          // If `pathPrefix` is configured by user, we should replace
          // the default prefix with `pathPrefix`.
          "/": `${pathPrefix}/`,
        },
        cacheId: `gatsby-plugin-offline`,
        // Don't cache-bust JS or CSS files, and anything in the static directory,
        // since these files have unique URLs and their contents will never change
        dontCacheBustUrlsMatching: /(\.js$|\.css$|static\/)/,
        runtimeCaching: [
          {
            // Use cacheFirst since these don't need to be revalidated (same RegExp
            // and same reason as above)
            urlPattern: /(\.js$|\.css$|static\/)/,
            handler: `cacheFirst`,
          },
          {
            // Add runtime caching of various other page resources
            urlPattern: /^https?:.*\.(png|jpg|jpeg|webp|svg|gif|tiff|js|woff|woff2|json|css)$/,
            handler: `staleWhileRevalidate`,
          },
          {
            // Google Fonts CSS (doesn't end in .css so we need to specify it)
            urlPattern: /^https?:\/\/fonts\.googleapis\.com\/css/,
            handler: `staleWhileRevalidate`,
          },
        ],
        skipWaiting: true,
        clientsClaim: true,
      }
    }
  ]
}
```
?

PR #16599, opened 2019-08-13
docs(recipes): Creating source nodes

juhosa:
## Description

A recipe for creating source nodes yourself.

I wrote it in the style that would have made sense to me when I was looking for this information in the recipes a few days ago. Meaning that the reader is not totally new to Gatsby, JavaScript and, GraphQL and knows where to find `gatsby-node.js` for example.

With that being said, all suggestions are more than welcome!

## Related Issues

No related issue.

### DIFF
--- a/docs/docs/recipes.md
+++ b/docs/docs/recipes.md
@@ -656,11 +656,11 @@ exports.sourceNodes = ({ actions, createNodeId, createContentDigest }) => {
     actions.createNode(node)
   })
 }
-
 ```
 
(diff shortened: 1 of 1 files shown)


juhosa:
The ci/circleci: unit_tests_www failed because of
```
An unexpected error occurred: "https://registry.yarnpkg.com/core-js/-/core-js-2.6.9.tgz: Request failed \"503 Service Unavailable\"".
```
😯

PR #16602, opened 2019-08-13
fix(gatsby-transformer-sqip): Return null early for unsupported file types

axe312ger:
## Description

If your list of nodes contains non-pixel images like SVGs, sqip will fail and break your side. Can be workaround via multiple queries and filtering, but, lets just fix the actual issue.

Return `null` when a file type is not supported.

Ps.: This issue has the potential to break a site only through content editing :scream:

### DIFF
--- a/packages/gatsby-transformer-sqip/src/__tests__/generate-sqip.js
+++ b/packages/gatsby-transformer-sqip/src/__tests__/generate-sqip.js
@@ -32,7 +32,7 @@ describe(`gatsby-transformer-sqip`, async () => {
   const absolutePath = resolve(
     __dirname,
     `images`,
-    `this-file-does-not-neet-to-exist-for-the-test.jpg`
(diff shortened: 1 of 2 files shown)


sidharthachatterjee:
```
@axe312ger Looks like a test needs to be updated 

Expected mock function to have been called one time, but it was called zero times.

      113 |       expect(result).toBe(null)
      114 | 
    > 115 |       expect(sqip).toHaveBeenCalledTimes(1)
          |                    ^
      116 |       expect(exists).toHaveBeenCalledTimes(1)
      117 |       expect(writeFile).toHaveBeenCalledTimes(0)
      118 |       expect(readFile).toHaveBeenCalledTimes(0)

      at Object.toHaveBeenCalledTimes (packages/gatsby-transformer-sqip/src/__tests__/generate-sqip.js:115:20)
```

axe312ger:
Ooops, not committed :sweat_smile:

PR #16611, opened 2019-08-14
gatsby-theme-blog path generation fixes

ChristopherBiscardi:
## Description

https://github.com/gatsbyjs/gatsby/pull/15988 caused a regression in the url generation for `gatsby-theme-blog` that resulted in the `basePath` theme option never being applied to the url. This PR adds that back in and also makes sure that you can override the slug from the frontmatter. If the slug is relative, we prepend `basePath`, if the slug is absolute, we use that URL directly. If there is no slug, we fall back to `createFilePath` which included the `index.js` behavior originally added in #15988

This fix happens in `gatsby-theme-blog-core` now because that's where the logic is.

### DIFF
--- a/themes/gatsby-theme-blog-core/gatsby-node.js
+++ b/themes/gatsby-theme-blog-core/gatsby-node.js
@@ -113,7 +113,7 @@ exports.onCreateNode = async (
     let slug
     if (node.frontmatter.slug) {
       if (path.isAbsolute(node.frontmatter.slug)) {
-        // absolute paths take precedance
+        // absolute paths take precedence
         slug = node.frontmatter.slug
       } else {
         // otherwise a relative slug gets turned into a sub path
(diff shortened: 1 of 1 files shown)


themes/gatsby-theme-blog-core/gatsby-node.js
```diff
@@ -109,12 +110,26 @@ exports.onCreateNode = async (
...
-    const slug = createFilePath({
-      node: fileNode,
-      getNode,
-      basePath: contentPath,
-    })
-
+    let slug
+    if (node.frontmatter.slug) {
+      if (path.isAbsolute(node.frontmatter.slug)) {
+        // absolute paths take precedance
```
ChristopherBiscardi:
```suggestion
        // absolute paths take precedence
```

themes/gatsby-theme-blog-core/gatsby-node.js
```diff
@@ -109,12 +110,26 @@ exports.onCreateNode = async (
...
+      // otherwise use the filepath function from gatsby-source-filesystem
+      const filePath = createFilePath({
+        node: fileNode,
+        getNode,
+        basePath: contentPath,
+      })
+
+      slug = urlResolve(basePath, filePath)
+    }
+    console.log("slug", slug)
```
ChristopherBiscardi:
```suggestion
```

PR #16623, opened 2019-08-14
tests(gatsby-remark-images): Add tests for backgroundColor

kimbaudi:
## Description

Improve [gatsby-remark-images](https://www.gatsbyjs.org/packages/gatsby-remark-images/) documentation. Specifically, improve what `backgroundColor` option does and how to disable it.

Also, add more unit tests for this plugin

I was going to create a new PR to be able to disable `box-shadow` css property on `` element, but realize that setting `backgroundColor: 'none'` does exactly what I need. Others have suggested `backgroundColor: 'transparent'` to work-around their issue. So I thought it would be a great idea to mention this in the doc.


packages/gatsby-remark-images/README.md
```diff
@@ -52,7 +52,7 @@ plugins: [
...
+| `backgroundColor`      | `white` | Set the box-shadow color of the image to match the background image of your design.<br /><br />**Note:** To completely disable box-shadow, use `backgroundColor: none`. You can also use `backgroundColor: transparent` for a transparent box-shadow (i.e., `box-shadow: inset 0px 0px 0px 400px transparent`).                                                                                                                                                                                                                                                                                                                    |
```
pieh: There are 2 separate things here:
 - resulting effect (which is background after all)
 - implementation (which uses box-shadow)

Just specifying one of those will always confuse someone, so maybe adjusting this to somehow use `background` to describe effect and `box-shadow` to describe implementation will make it so users looking for `box-shadow` will still find this and people just looking at prop list will not be confused?

kimbaudi: hi, my main reason for mentioning `backgroundColor: 'transparent'` and `backgroundColor: 'none'` was because I wanted to remove `box-shadow` entirely from ``. I was a bit confused with the `backgroundColor` option because I thought I can only change the box-shadow color, but not remove it. After I changed from `backgroundColor: 'transparent'` to `backgroundColor: 'none'`, I noticed that this would remove the `box-shadow` from the `` element. I thought mentioning this in the doc might help someone else, but I don't want this to confuse anyone. Maybe we should just close this PR. I can work on something else instead.

kimbaudi: Actually, maybe we can remove the changes to README, but still keep the unit tests

m-allanson: @kimbaudi I think this is useful info and we should include it 👍 However, I agree with @pieh that people don't really need to know implementation details. I'd propose something like this:

```suggestion
| `backgroundColor`      | `white` | Set the background color of the image to match your design.<br /><br />**Note:** To completely remove the background color, set this option to `none`.                                                                                                                                                                                                                                                                                                                    |
```

This keeps your update on how to disable the background color, but removes the info about `backgroundColor: transparent` as that's effectively the same as `backgroundColor: none`. It also removes any mention of box-shadow, as that knowledge isn't required to understand this option. 

What do you think?

kimbaudi: Hi I made an edit based on you and pieh's suggestion. Thank you both! I will push the changes.

kimbaudi:
@pieh and @m-allanson - I changed the description for the `backgroundColor` option based on your suggestions (removed implementation details. focus on the resulting effect of `transparent` and `none`)

But I want to add one more unit test to make sure that setting `backgroundColor: 'none'` does not add `box-shadow` styling to `` tag since we don't have that test right now. Let me do that right now before you guys decide to accept this PR.

kimbaudi:
I updated the unit tests to use `window.getComputedStyle().getPropertyValue` instead of `style` property to see if makes any difference, but it doesn't 😞 . So even though I added another test for `backgroundColor: none`, it is pointless. To really test whether `backgroundColor: 'none'` remove `box-shadow` requires testing on a browser. Maybe there is a way to do it in Jest by connecting to a browser, but I don't know anything about that.

Still, I think the unit tests are still useful so I hope you guys will still accept this PR 🤞

kimbaudi:
thank you for reviewing the changes @sidharthachatterjee! yes, I should probably take a look at cypress for end-to-end testing.

PR #16630, opened 2019-08-14
docs(gatsby-plugin-netlify): Remove H/2 server push section

sw-yx:
## Description

hi! i work at netlify. add note that H/2 server push does not work on Netlify right now

## Related Issues

n/a


KyleAMathews:
Maybe just remove the section altogether?

sw-yx:
idk, what if we manage to fix it someday? your call you wrote it :)

sidharthachatterjee:
Let's remove the section for now and add it back once it works consistently!

sw-yx:
done

PR #16642, opened 2019-08-15
chore(docs): Add note about react-loadable

slk333:
using 'react loadable' module fixed the error that blocks SSR. (no window object problem)
The webpack hack provided in the article on the other hand didn't work.

## Description

The solution provided in the article (changing the gatsby-node.js) didn't work. Using 'react loadable' is a very decent alternative solution that should be mentioned.

### DIFF
--- a/docs/docs/debugging-html-builds.md
+++ b/docs/docs/debugging-html-builds.md
@@ -72,4 +72,4 @@ exports.onCreateWebpackConfig = ({ stage, loaders, actions }) => {
 }
 ```
 
-Another solution is to use a package like [react-loadable] (https://github.com/jamiebuilds/react-loadable). The module that tries to use `window` will be dynamically loaded only on the client side (and not during SSR).
+Another solution is to use a package like [react-loadable](https://github.com/jamiebuilds/react-loadable). The module that tries to use `window` will be dynamically loaded only on the client side (and not during SSR).
--- a/docs/docs/debugging-html-builds.md
+++ b/docs/docs/debugging-html-builds.md
@@ -72,4 +72,4 @@ exports.onCreateWebpackConfig = ({ stage, loaders, actions }) => {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


docs/docs/debugging-html-builds.md
```diff
@@ -71,3 +71,5 @@ exports.onCreateWebpackConfig = ({ stage, loaders, actions }) => {
   }
 }
 ```
+
+Another solution is to use a package named [reactloadble] (https://github.com/jamiebuilds/react-loadable). The module that wants to use the window object will be imported in a way that doesn't trigger a build error.
```
sidharthachatterjee:
```suggestion
Another solution is to use a package like [react-loadable] (https://github.com/jamiebuilds/react-loadable). The module that tries to use `window` will be dynamically loaded only on the client side (and not during SSR). 
```

m-allanson:
@slk333 are you able to give more info on what didn't work with the webpack config in the article? Maybe we can update that to have clearer instructions.

slk333:
@m-allanson 
```
exports.onCreateWebpackConfig = ({ stage, loaders, actions }) => {
  if (stage === "build-html") {
    actions.setWebpackConfig({
      module: {
        rules: [
          {
            test: /bad-module/,
            use: loaders.null(),
          },
        ],
      },
    })
  }
}
```
the package I wanted to use was called "react-paypal-button-v2"
so I wrote the line like this
```
 {
            test: /react-paypal-button-v2/,
            use: loaders.null(),
          },
```
as it didn't work, I tried to use simple quotes
```
 {
            test: '/react-paypal-button-v2/',
            use: loaders.null(),
          },
```
and even
```
 {
            test: 'react-paypal-button-v2',
            use: loaders.null(),
          },
```
but with no success.
I'm not sure if it failed because I didn't input the name correctly, or because the command altogether didn't work. I got some errors regarding regular expresions. I tried for like 10 minutes and gave up. I wish it's just because I didn't input the name correctly!

slk333:
just as a side note, the 'react-paypal-button-v2' module got updated to work with SSR, so we cannot trigger the build error with that module anymore. The previous version of it can still be used to test the build error.

PR #16657, opened 2019-08-15
fix(gatsby): Set default for modulesThatUseGatsby

ValeraS:


sidharthachatterjee:
Thank you for opening this @ValeraS 

It doesn't look like this _corrects_ or _changes_ any behaviour and is a matter of style from what I can tell. The default value is set in both cases. 

We prefer to avoid churn due to purely stylistic changes so I'm closing this for now. 

Thank you for your contribution nonetheless! We appreciate it 💜

ValeraS:
@sidharthachatterjee It's not purely stylistic changes. It fixes the next bug if you want to use this function to get custom rules like this:
```js
rules.js({
	exclude: ...
})
``` 

you will get an error during a webpack config creation, because of modulesThatUseGatsby will `undefine`.

sidharthachatterjee:
@ValeraS Oh, you're right! My mistake. Thank you for pointing that out. 

Merging this in

sidharthachatterjee:
Published in `gatsby@2.13.65`

Issue #10048, opened 2018-11-20
Better default for gatsby-remark-copy-linked-files


lekevicius:
## Summary

Default for file URLs should be `hash/filename.ext`, not `filename-hash.ext`

### Basic example

Currently the default is `file.pdf` -> `file-ea0f4f14f1b991e92c681da93172497d.pdf`. 

It is set in [this function](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-remark-copy-linked-files/src/index.js#L21):

```js
const newFileName = linkNode =>
  `${linkNode.name}-${linkNode.internal.contentDigest}.${linkNode.extension}`
```

I suggest to make the default URL be:

```js
const newFileName = linkNode =>
  `${linkNode.internal.contentDigest}/${linkNode.name}.${linkNode.extension}`
```

This would follow the general structure of gatsby-image paths, and make downloaded file name much nicer.

If default is not changed then at least this should be a customizable option, perhaps a function that could be passed.

### Motivation

Currently, for every user downloading the file, filename is very long. In many cases the "hash" part is much longer than the actual file name. Default setting should be nicer.

pieh:
I think that's good idea - we would possibly need to ensure that extra directory exists.

@lekevicius would you be interested in creating PR with this change?

KyleAMathews:
Yeah this would be a good change. We made the same change for image paths.

lekevicius:
The Kyle has spoken.

I'll try to make a PR this week.

kimbaudi:
Hi, I would like to implement this feature. I will submit a PR shortly 🤞

kimbaudi:
Hi, I completed implementing this feature (https://github.com/kimbaudi/gatsby/commit/f7fbd10d2268a4848df7993910d52680516ead5b):
1. change the default destination format from `filename-hash.ext` to `hash/filename.ext`
2. allow customization of the destination format:
    ```js
    destinationDir: f => `${f.hash}/${f.name}`  // f7fbd/myfile.txt
    destinationDir: f => `${f.name}-${f.hash}`  // myfile-f7fbd.txt
    destinationDir: f => `foo${f.hash}---${f.name}bar`  // foof7fbd---myfilebar.txt
    destinationDir: f => `foo${f.hash}/${f.name}bar`  // foof7fbd/myfilebar.txt
    ```

I perform same validation to make sure that the path is a child of `public` folder:

```js
destinationDir: `../`  // error
[...]

PR #16669, opened 2019-08-15
(feat)gatsby-remark-copy-linked-files: create hashLength option to specify the digest hash length

kimbaudi:
## Description

This PR adds a new option called `hashLength` that would allow users to specify the length of [contentDigest](https://www.gatsbyjs.org/docs/node-interface/#contentdigest)

```
{
  resolve: `gatsby-remark-copy-linked-files`,
  options: {
      hashLength: 9
  },
},
```

### DIFF
--- a/packages/gatsby-remark-copy-linked-files/src/__tests__/index.js
+++ b/packages/gatsby-remark-copy-linked-files/src/__tests__/index.js
@@ -418,6 +418,36 @@ describe(`gatsby-remark-copy-linked-files`, () => {
     })
   })
 
+  describe(`options.hashLength`, () => {
+    const imagePath = `images/sample-image.gif`
+
+    it(`sets the contentDigest hash to the length supplied by hashLength`, async () => {
+      const markdownAST = remark.parse(`![some absolute image](${imagePath})`)
+      const validHashLength = 3
+      const expectedNewPath = path.posix.join(
+        process.cwd(),
+        `public`,
+        `/undefined/undefined.gif`
+      )
+      expect.assertions(3)
+      await plugin(
+        {
+          files: getFiles(imagePath),
+          markdownAST,
+          markdownNode,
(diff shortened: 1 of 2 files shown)


kimbaudi:
I am new to Jest and unit testing, but feel that the unit tests for this plugin requires actual `fs-extra` functionality (i.e. `existsSync`, `copy`, and `ensureDir`) instead of mocked ones. [Currently, the unit tests expects `undefined` values](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-remark-copy-linked-files/src/__tests__/index.js#L416):

```
expect(imageURL(markdownAST)).toEqual(`/undefined/undefined.gif`)
```

In order to test this new `hashLength` option, I would have to be able to test it using an actual file. Does anybody know if this is possible with Jest? I'll start looking into it, but I am unsure. Maybe this type of test requires end-to-end testing with Cypress instead of Jest? If anyone can point me in the right direction, that would be really helpful.

sidharthachatterjee:
I would be wary of adding this because it if I understand correctly, it strips off characters off the `contentDigest` from a Node. 

This _could_ lead to collisions. Also, if a user needs to set a specific length for a Node's `contentDigest`, they are free to set that themselves. 

Based on the above, I'd prefer to close this and _not_ add this option. What do you think @kimbaudi?

sidharthachatterjee:
@pieh Thoughts?

pieh:
Why do you need that option? I can't really think of situation when I would want/need to mess with `contentDigest`?

kimbaudi:
personally, its fine to close this. I really don't need it. I think I misread something about long filename due to hashing from https://github.com/gatsbyjs/gatsby/issues/10048. I will close this.

sidharthachatterjee:
Thanks @kimbaudi 👍

PR #16679, opened 2019-08-16
fix(gatsby): Use concrete type resolvers when filtering on node interface fields

stefanprobst:
When filtering on `@nodeInterface` fields, we currently use the resolver on the interface type. This PR changes this to use the resolver on the type returned from the interface's `resolveType`, which is in line with what graphql does when resolving the fields in the selection set.

Related: https://github.com/gatsbyjs/gatsby/issues/12272#issuecomment-520039307
Fixes:  #16890

Note: I tried to change as little as possible in `prepare-nodes` which will be replaced by @freiksenet's work in the `materialization` branch anyway.


freiksenet:
This will cause quite a conflict with what I'm doing with materialization. I can try porting that so it's based on that.

stefanprobst:
Superseded by #17284

PR #16681, opened 2019-08-16
chore: Move back to upstream relay-compiler

stefanprobst:
Move back to upstream `relay-compiler` -- seems to work fine.


sidharthachatterjee:
So this was attempted previously in https://github.com/gatsbyjs/gatsby/pull/15963 but we closed it because Relay regressed on the fix that @freiksenet made! 

Do you know about this regression (I don't)? How do you feel about adding a test that catches that regression if yes?

stefanprobst:
tbh i don't know the exact reason for the fork. i tested this with .org, and checked the sorting of the blog posts (so the sort enum works), but i might not have the full context here

stefanprobst:
if we are not confident to switch back, we should release a new version of our fork with https://github.com/freiksenet/relay/pull/1

sidharthachatterjee:
https://github.com/freiksenet/relay/pull/1 is now published as `2.0.0-printer-fix.3` and an upgrade to it is in https://github.com/gatsbyjs/gatsby/pull/16986

Closing this

stefanprobst:
Can we keep this open until we figure out why we do/don't need to stay on our fork of `relay-compiler`?

sidharthachatterjee:
@stefanprobst Let's track that in https://github.com/gatsbyjs/gatsby/issues/11536

I think we should add a (failing) test if possible but I'd recommend against keeping this open because it's bound to out of date/stale with time

stefanprobst:
@sidharthachatterjee i just think we should figure this out soon. @freiksenet mentioned in #11536 that last time we tried it broke out builds, but i could not reproduce this with v5 anymore.
if we can't remember the reason for the fork it's maybe best to actually move back and wait for a bug report?

freiksenet:
Have you tried www with latest relay? It was fixed with my patch at 2 then broken again in 3.

m-allanson:
@stefanprobst said this:

>tbh i don't know the exact reason for the fork. i tested this with .org, and checked the sorting of the blog posts (so the sort enum works), but i might not have the full context here

So it seems like it's fine on .org? Is there anything else we should check? I agree we should switch back if there's nothing preventing the change.

PR #16683, opened 2019-08-16
docs(debugging-html-builds): Mention the useEffect hook, reference React docs, use ternary operator

ImedAdel:
## Description

Extend the documentation with the `useEffect` hook in React. 

Also, link to React's documentation in case the reader does not know about hooks or lifecycles yet. 

Finally, it introduces a ternary operator to check if a window is defined or not, in case that module has to be defined (I use this daily at work).


ImedAdel:
Would it be okay if I went over all the examples and replaced lifecycles and classes with hooks and functions? :)

Issue #15873, opened 2019-07-18
[gatsby-plugin-page-creator] (Suggestion) Use explicitly plugin page creator for new gatsby sites


matthieuauger:
## Summary

Hello 🙋‍♂️I'm currently working on the migration of a gatsby site to a gatsby theme. One of the issues I faced is [explained here](https://github.com/gatsbyjs/gatsby/blob/69445cec9b93897a5819f344965da57bfddd5f22/docs/docs/themes/converting-a-starter.md#sourcing-pages
). By default a gatsby site will load pages in src/pages directory. A theme doesn't and need to have the gatsby-page-creator plugin installed and enabled.

```
{
    resolve: `gatsby-plugin-page-creator`,
    options: {
        path: `${__dirname}/src/pages`,
    },
},
```

In order to have consistency and ease the migration of a site to a theme, could we explicitly add the plugin page creator in new sites created in Gatsby as well ?

matthieuauger:
@ChristopherBiscardi As this is related to themes, you may have an opinion on this, your feedback would be really helpful 💯

matthieuauger:
Hello @gatsbyjs/themes @gatsbyjs/core ✋Does it make sense to you? I'm working to ease the migration of gatsby sites to gatsby themes and this change could save a bit of time to gatsby theme creators.

If I'm correct, Gatsby already uses gatsby-plugin-page-creator under the hood and [seems loaded here](https://github.com/gatsbyjs/gatsby/blob/34e4add901738e2b15fd1b5875ccb76871ce57a3/packages/gatsby/src/bootstrap/load-plugins/load.js#L196)

I think we may explicitly expose the configuration in gatsby-config.js to ease theme creation

ChristopherBiscardi:
https://github.com/gatsbyjs/gatsby/issues/15873

sidharthachatterjee:
Published in `gatsby@2.15.6`

PR #16698, opened 2019-08-17
feat(gatsby): enable page-creator on all themes

ChristopherBiscardi:
## Description

this enables the page-creator plugin on all themes (and by extension, all plugins), allowing themes to ship pages in `src/pages`. The order of page-creators matches the order of theme application. No one should be using `src/pages` for anything today unless it's an additional page-creator instance, so this should be non-breaking as a change.

I thought we did this already but apparently it slipped through the cracks.

## Related Issues

Fixes https://github.com/gatsbyjs/gatsby/issues/15873


ChristopherBiscardi:
The failing test here is because of (as far as I can tell) a malformatted test plugin spec that doesn't include `resolve`

```
{ id: 'c731052f-1e62-5905-9fb3-edc9057be8f1',
  name: 'TEST',
  pluginOptions: { plugins: [] } }
```

ChristopherBiscardi:
Turns out that in the source code for the relevant function there's special handling for this test input specifically: https://github.com/gatsbyjs/gatsby/blob/71f133b9b8d90d5d029b268f7a548c40e201805f/packages/gatsby/src/bootstrap/load-plugins/load.js#L143-L153 that doesn't return a full spec, which is then hidden by using the `replaceFieldsThatCanVary` function, which inserts a `resolve` field after the fact.

```js
      // Add some default values for tests as we don't actually
      // want to try to load anything during tests.
      if (plugin.resolve === `___TEST___`) {
        const name = `TEST`

        return {
          id: createPluginId(name, plugin),
          name,
          pluginOptions: {
            plugins: [],
          },
        }
      }
```

ChristopherBiscardi:
It is unclear to me what needs to be done to get the tests passing on windows and node v8. The id generation for plugins seems somehow unstable across node versions and platforms.

wardpeet:
How would you disable pages in themes? Like I want to use a theme but it's also serving pages. Should I run `deletePage` for the pages I don't want in gatsby-node?

Maybe this is an edge case but we probably want to add documentation on how to remove them as well? I might be overthinking this :joy:.

wardpeet:
to make tests work, you should try
```
 const replaceFieldsThatCanVary = plugins =>
    plugins.map(plugin => {
      if (plugin.pluginOptions && plugin.pluginOptions.path) {
        plugin.pluginOptions = {
          ...plugin.pluginOptions,
          path: plugin.pluginOptions.path.replace(
            slash(process.cwd()),
            `<PROJECT_ROOT>`
          ),
        }
      }

      return {
        ...plugin,
        resolve: ``,
        version: `1.0.0`,
      }
    })
```

We probably want to rewrite the `jest-serializer-path` to also convert posix paths.

ChristopherBiscardi:
> How would you disable pages in themes? 

yeah, `deletePage` would be it. Or you can shadow them.

sidharthachatterjee:
Published in `gatsby@2.15.6`

ChristopherBiscardi:
Thanks for getting this over the finish line @sidharthachatterjee!

universse:
Anyone keen to look into this issue #17373. Seems like this PR caused it.

PR #16701, opened 2019-08-17
Add new image processing benchmark site

KyleAMathews:
Benchmark for downloading and processing images.

There's a local plugin with a list of ~1300 images all 1280px wide. When you run, the site will generate a page for each image included (defaults to 100 which you can vary by setting the `NUM_IMAGES` env variable) and set to be processed to a 400 x 400 fixed image.


benchmarks/markdown/gatsby-node.js
```diff
@@ -1,32 +1,32 @@
 const NUM_PAGES = parseInt(process.env.NUM_PAGES || 5000, 10)
-const template = require(`./page-template`)
+const template = require('./page-template')
```
wardpeet: looks like prettier isn't handling this correctly. It's different from the other packages inside the monorepo.

benchmarks/image-processing/src/pages/index.js
```diff
@@ -0,0 +1,31 @@
...
+        <Link to={`/${n.id}/`}>{n.id}</Link>
+      </div>
+    ))}
+  </Layout>
+)
+
+export default IndexPage
+
+export const query = graphql`
+  query MyQuery {
```
wardpeet: Nit:
No need to name the query
```suggestion
  query {
```

benchmarks/image-processing/plugins/gatsby-source-remote-images/fetch-image-urls.js
```diff
@@ -0,0 +1,41 @@
+const { Sema } = require(`async-sema`)
+const fetch = require(`node-fetch`)
+const _ = require(`lodash`)
+const fs = require(`fs`)
+const path = require(`path`)
+
+const baseUrl = `http://www.splashbase.co/api/v1/images/`
+
+const s = new Sema(
+  10, // Allow 4 concurrent async calls
```
wardpeet: Comment is incorrect
```suggestion
  10, // Allow 10 concurrent async calls
```

benchmarks/image-processing/plugins/gatsby-source-remote-images/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "gatsby-source-unsplash",
```
wardpeet:
```suggestion
  "name": "gatsby-source-remote-images",
```

benchmarks/image-processing/plugins/gatsby-source-remote-images/gatsby-node.js
```diff
@@ -0,0 +1,41 @@
+const fs = require(`fs`)
+const { createRemoteFileNode } = require(`gatsby-source-filesystem`)
+
+const NUM_IMAGES = parseInt(process.env.NUM_IMAGES || 100, 10)
```
wardpeet:
```suggestion
const NUM_IMAGES = Number(process.env.NUM_IMAGES || 100)
```

benchmarks/image-processing/src/pages/index.js
KyleAMathews: I blame @pieh since I copy/pasted from the explorer :-P

benchmarks/image-processing/plugins/gatsby-source-remote-images/gatsby-node.js
KyleAMathews: We want an int not a Number (which could be a float)

PR #16705, opened 2019-08-17
docs(tutorial): add missing query keyword in gatsby-node.js snippet

dantereve:
## Description

In tutorial part 7, add missing query keywords in the GraphQL query of the gatsby-node.js snippet. 

## Related Issues

none.


pieh:
`query` keyword is optional, so strictly speaking it's not needed. As example here's working piece of gatsby-node code that's not using `query` keyword: https://github.com/gatsbyjs/gatsby/blob/71f133b9b8d90d5d029b268f7a548c40e201805f/starters/blog/gatsby-node.js#L8-L28

pieh:
Said above, in earlier parts of tutorial we do use `query` keyword (for example in https://www.gatsbyjs.org/tutorial/part-six/#create-a-list-of-your-sites-markdown-files-in-srcpagesindexjs ), so this change is very reasonable - let's get this in!

Issue #15947, opened 2019-07-21
How to refresh the localhost https certificate?


DevinKott:
## Summary

I tried following the gatsbyjs local https guide [here](https://www.gatsbyjs.org/docs/local-https/), but I got an error while setting it up. Now, when I try running `npm run develop -- --https`, it does not start because of the error
`UNHANDLED REJECTION request to https://localhost:8000/___graphql failed, reason: unable to verify the first certificate`
## Relevant information

When trying to set up local https, I put in the wrong sudo password a couple of times, and exited the process. Now it's automatically trying to generate certificates on the fly that are wrong and the building/running fails every time.

### File contents (if changed)

`package.json`: N/A 
`"start": "npm run develop -- --https",`

The solution [here](https://github.com/gatsbyjs/gatsby/issues/15441) worked, but firefox/chrome said it was a security risk to visit the localhost website. How can i reset it and try setting up https again?

buz-zard:
I'm having the same issue. I'm trying to upgrade from `"gatsby": "^2.8.2",` to `"gatsby": "^2.13.50",` and the build fails:

```
UNHANDLED REJECTION request to https://localhost:8000/___graphql failed, reason: unable to verify the first certificate



  FetchError: request to https://localhost:8000/___graphql failed, reason: unable to verify the first certificate

  - index.js:1455 ClientRequest.<anonymous>
    [locker-web]/[node-fetch]/lib/index.js:1455:11

  - destroy.js:82 emitErrorNT
    internal/streams/destroy.js:82:8

  - destroy.js:50 emitErrorAndCloseNT
    internal/streams/destroy.js:50:3

  - next_tick.js:63 process._tickCallback
    internal/process/next_tick.js:63:19
```

buz-zard:
For now my workaround is `NODE_TLS_REJECT_UNAUTHORIZED=0 gatsby develop --https`

geoffdavis92:
Would also like to see this looked at. Ran the `--https` command without sudo, now unable to setup a (truly secure) HTTPS connection through Gatsby.

One thing that would be helpful– and I may be able to help with this– is to present a warning on the [`local-https` docs page](https://www.gatsbyjs.org/docs/local-https/) to indicate users should maybe first try this using `sudo` before they run it? At least until this bug is resolved.

Bulletninja:
I simply ran `gatsby develop --https` for the first time in a new project and get a similar error.
```
[...]

Issue #15441, opened 2019-07-05
FetchError: request to https://localhost:8000/___graphql failed, reason: unable to verify the first certificate


sixclones:
Hello,

The first time I run Gatsby with SSL activated, an error occured. So now the error is fixed, when I 
run `gatsby develop "-S"`, I got this :
`FetchError: request to https://localhost:8000/___graphql failed, reason: unable to verify the first certificate`
My tought is I haven't setting up my SSL certificate and I don't know how to "reset" it to setting up in a good way.
I tried to clean gatsby's cache and `rm -rf node_modules`, change nothing.

## Environment
```
  System:
    OS: Linux 5.0 Ubuntu 19.04 (Disco Dingo)
    CPU: (4) x64 Intel(R) Core(TM) i3-6100 CPU @ 3.70GHz
    Shell: 5.5.1 - /usr/bin/zsh
  Binaries:
    Node: 10.15.3 - ~/.nvm/versions/node/v10.15.3/bin/node
    Yarn: 1.1.0 - /usr/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v10.15.3/bin/npm
  Languages:
    Python: 2.7.16 - /usr/bin/python
  Browsers:
    Chrome: 74.0.3729.131
    Firefox: 67.0.4
  npmPackages:
    gatsby: ^2.13.3 => 2.13.3 
    gatsby-plugin-postcss: ^2.1.0 => 2.1.0 
    gatsby-plugin-react-helmet: ^3.1.0 => 3.1.0 
    gatsby-plugin-sitemap: ^2.2.1 => 2.2.1 
```

fraserisland:
This is a known bug, you can use "NODE_TLS_REJECT_UNAUTHORIZED=0 gatsby develop --https" as a workaround though until it gets fixed though

sixclones:
Thx for the answer.
FYI the related bug your are quoting is actually this bug. ^^
Just by curiousity, what's the bug ?

bcanseco:
Seems like this is it https://github.com/gatsbyjs/gatsby/issues/14990

rwitchell:
anyone coming here: if you are running SSL certificates for local development, using mkcert: you might need to run

```export NODE_EXTRA_CA_CERTS="$(mkcert -CAROOT)/rootCA.pem"```

source: https://github.com/FiloSottile/mkcert#using-the-root-with-nodejs

Issue #14990, opened 2019-06-20
Regression: self-signed certificates no longer work


aaronjensen:
## Description

I'm having a hard time sorting out what package version where causes the issue, but when upgrading from gatsby 2.9.8 to 2.9.9 I now get this error when starting gatsby with a self-signed certificate:

```
error UNHANDLED REJECTION


  FetchError: request to https://localhost:8002/___graphql failed, reason: self signed certificate

  - index.js:1455 ClientRequest.<anonymous>
    [hello-epics]/[node-fetch]/lib/index.js:1455:11

  - destroy.js:82 emitErrorNT
    internal/streams/destroy.js:82:8

  - destroy.js:50 emitErrorAndCloseNT
    internal/streams/destroy.js:50:3

  - next_tick.js:63 process._tickCallback
    internal/process/next_tick.js:63:19
```

My guess is it has to do with the upgrade of `node-fetch` in `gatsby-cli`

### Steps to reproduce

Create and trust ssl certificates, then run gatsby using them. I've also tried with the auto generated certificates and I get a different error.

```
gatsby develop --port 8002 --https --key-file ../config/localhost/key.pem --cert-file ../config/localhost/cert.pem
```
### Expected result

It should start gatsby

### Actual result

```
error UNHANDLED REJECTION


  FetchError: request to https://localhost:8002/___graphql failed, reason: self signed certificate

  - index.js:1455 ClientRequest.<anonymous>
    [hello-epics]/[node-fetch]/lib/index.js:1455:11

  - destroy.js:82 emitErrorNT
    internal/streams/destroy.js:82:8

  - destroy.js:50 emitErrorAndCloseNT
    internal/streams/destroy.js:50:3

  - next_tick.js:63 process._tickCallback
    internal/process/next_tick.js:63:19
```
### Environment

Run `gatsby info --clipboard` in your project directory and paste the output here.

```
  System:
    OS: macOS 10.14.5
    CPU: (8) x64 Intel(R) Core(TM) i7-7920HQ CPU @ 3.10GHz
    Shell: 5.6.2 - /usr/local/bin/zsh
  Binaries:
    Node: 10.15.3 - /var/folders/p4/0ptjct956432wm2q1dd048g40000gn/T/yarn--1561063929532-0.7049991999729923/node
    Yarn: 1.16.0 - /var/folders/p4/0ptjct956432wm2q1dd048g40000gn/T/yarn--1561063929532-0.7049991999729923/yarn
[...]

PR #16726, opened 2019-08-19
fix(gatsby): Switching from devcert-san to devcert to fix HTTPS issues

### LINKED ISSUE #16212: --https fails if the path of the project contains spaces
## Description

command `gatsby run develop --https` fails in the generate certificate step if the path of the project contains spaces

### Steps to reproduce

I'm on windows, so I don't know if the same steps will reproduce the issue on other OS's
1. Create a user whose name contains spaces
1. Create a project anywhere
1. Install openssl and add it to the PATH env variable
1. run the command `gatsby develop --https`

### Expected result

Gatsby generates certificate and starts an https server

### Actual result

Error
`Error: Command failed: openssl genrsa -out C:\Users\User Name With Spaces\AppData\Local\devcert\config\devcert-ca-root.key 2048`
and process exits

### Environment

  System:
    OS: Windows 10
    CPU: (4) x64 Intel(R) Core(TM) i7-6500U CPU @ 2.50GHz
  Binaries:
    npm: 6.4.1 - C:\Program Files\nodejs\npm.CMD
  Languages:
    Python: 2.7.15
  Browsers:
    Edge: 42.17134.1.0
  npmPackages:
    gatsby: ^2.13.3 => 2.13.3 
    gatsby-image: ^2.2.4 => 2.2.4 
    gatsby-plugin-intl: ^0.2.6 => 0.2.6 
    gatsby-plugin-manifest: ^2.2.1 => 2.2.1 
    gatsby-plugin-offline: ^2.2.0 => 2.2.0 
    gatsby-plugin-sharp: ^2.2.2 => 2.2.2 
    gatsby-source-filesystem: ^2.1.2 => 2.1.2 
    gatsby-transformer-sharp: ^2.2.1 => 2.2.1

rmorabia:
- Upgrade `devcert-san` to `devcert`
- Update calls from package to match the new changes

## Description

I changed `devcert-san` to `devcert` as described by @Js-Brecht in this comment: https://github.com/gatsbyjs/gatsby/issues/16212#issuecomment-520059430

This should work, however, I'm running on Linux (unix line endings?) and can't confirm it. Can someone on Windows run the repo before the PR is merged and confirm it's fixed?

## Related Issues

Fixes #16212


Js-Brecht:
Did you have a moment to investigate the key/cert paths shown here?

> https://github.com/gatsbyjs/gatsby/pull/16726/files#diff-b7d3f367223033fe45bc0ed8f2915244R28-R29

Since they're not returned by `certificateFor()`, it could be misleading having the paths returned from one branch when the user specifies the cert, but not the other when returned by `certificateFor`.  Somebody down the line might want to try using those paths, only to encounter some unexpected behavior when they come back undefined.

I guess if they're needed, somebody could issue a PR over @ `devcerts` to include them in the return value for that function. :man_shrugging:

Js-Brecht:
AFAICT, this fixes the issue (with paths) on Windows. 👍 

However, it causes other issues.  It installs the certificate in the Certificate Authority store, but when it tries to open Firefox to run the wizard for trusting the cert, the page it opens results in an endless refresh loop.  And that CA cert is basically useless.  No trust... "SEC_ERROR_BAD_SIGNATURE".  Site still shows security errors.

I've opened issue davewasmer/devcert#35

This is definitely a regression from the previous version.  On a path with no spaces, the previous version works correctly.

wardpeet:
I'm getting `UNHANDLED REJECTION request to https://localhost:8000/___graphql failed, reason: unable to verify the first certificate` I do have the certificate installed 
image

Js-Brecht:
@wardpeet I noticed that same thing, but it's an existing issue (#15947, #15441, #14990).  I had the same problem using the old version.  The work around [here](https://github.com/gatsbyjs/gatsby/issues/15947#issuecomment-518124400) sems to work.

rmorabia:
Yep, had the same issues with the cert stuff. I had assumed that it was the key cert path issues that was causing this so I didn't change this. However, I had the same issue even without my changes, so I just decided to submit the PR anyway.

Right now, the rest of this is above my level of understanding, so I guess let me know if there's anything I can do.

wardpeet:
i'll see what I can do :)

Js-Brecht:
Is it just me, or is there something weird going on with `gatsby-graphiql-explorer` in this repo?  Every time I run `gatsby-dev`, I have to manually install `gatsby-graphiql-explorer` in the target site in order to run `gatsby develop`.  Gets rather frustrating, especially because `gatsby-dev` kept crashing when running `yarn build` in `rmorabia/gatsby`.  So, it turned into `yarn build` -> `gatsby-dev` -> `yarn add gatsby-graphiql-explorer` -> `cross-env NODE_TLS_REJECT_UNAUTHORIZED=0 gatsby develop --https`.  I don't think it would affect the master; it just slows down testing.

Really, it could just be me.  I didn't try wiping it all out and starting over.  Could be the yarn cache, too...

rmorabia:
Nope, had the same issue! I thought that was a general `gatsby-dev` issue, happened to me no matter what I tried. I just ran `yarn --force` to fix it. Asked about it in the Discord channel as well.

I have no clue how my 5 lines of code change could mess up that much, lol.

Js-Brecht:
@rmorabia have you tried updating your repo?  It looks like you're on `2.13.52`, and the master is `2.13.71`.  Could pull the changes from the master into a new branch, and see if it fixes the problem 🤷‍♂.

rmorabia:
Let me try messing around with that and see if that fixes the issue there. Not a huge deal either way I think, the certs issue is still way above my head, but i can try and get the `gatsby-dev` stuff working a bit better.

Js-Brecht:
I found a fix for the `unable to verify the first certificate` error.  Node apparently doesn't look in the OS trust store.  You have to add the CA cert to the list of global CAs:

```js
    let https = require(`https`)
    let cas = https.globalAgent.options.ca || [];
    cas.push(`
-----BEGIN CERTIFICATE-----
MIIDDTCCAfWgAwIBAgIJANJT2lcJ+EIWMA0GCSqGSIb3DQEBCwUAMBIxEDAOBgNV
BAMMB2RldmNlcnQwHhcNMTkwODIxMjE1NDM0WhcNMTkwOTIwMjE1NDM0WjASMRAw
DgYDVQQDDAdkZXZjZXJ0MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEA
xU9F35WajjIakpCVBntez6GZFc/NpqjcIkkZwq/1nBU9U43nD+ND43fCzN9Gd9Z4
c2U/z+S4hgB56L8Xcu9mfXM3Ctz+mAP5sEW9bcBjZszTOTiq+ixQWzMIeJTqVWbN
uXTOT3wVdQWLpapGCndHkbEIRybvOg+28BkH0n2GLrxNb/vjICFOYlNHGMRg0m15
vuuZgMmQVCrny/aMCZ5J9TL2btIeoK7it64beQViqRaDJnyn0CVocxZZdWtp/PSv
+S/gj5YKS6bs2KN5cSJoYCK1Tl2Rvj6YmEyL6GWwKDfCBd7N6O/TqiLERJHpZKEy
EDvzsdmKuTXPcIkGhxTGJQIDAQABo2YwZDAdBgNVHQ4EFgQUfn+zP7IOuQtUcpI1
2O7zlEae3JMwHwYDVR0jBBgwFoAUfn+zP7IOuQtUcpI12O7zlEae3JMwEgYDVR0T
AQH/BAgwBgEB/wIBADAOBgNVHQ8BAf8EBAMCAYYwDQYJKoZIhvcNAQELBQADggEB
ABRrQvbMyo+SZz/e67yKGDpesEJTppq1lb6ikeS5MAeUIF54hpl8ipLoeW+EHfoo
aVLsOpokx6ilCMkfn6ggeN077g4Sby4ZQK40LwUoyAib8UrWIR7HpUxZKhazv7Pp
04yCwkLKJzGZzKLfYvhDCC5IAlS1oLDlv96+FpFVGk7j6KirqHPzvmUACIO2PkjR
2CAbleFDylUwKtLCZFulLPEIovLBN5tyTJ6NU4ZqBEWguQ83TQYvhzjAzwRjfK4W
7Km3Y3pYyRDgx6mc/gxKOzV6FYsholkM05TMGrVOFFdJNUDyP4MBEqYmi4uK87ME
TNkif2W2D8d1iFsTNKvMgSQ=
-----END CERTIFICATE-----`)
    https.globalAgent.options.ca = cas
    server = https.createServer(program.ssl, app)
```

I had to go through some steps to get to that point.  `devcert` does not give you access to the CA cert.  In fact, when it writes it out to the disk (on windows), it writes it encrypted using the provided password as a key, and when it decrypts it, it asks for the password.  On linux, when it writes it to disk, it makes it readonly for root, and when it reads it, it runs `sudo`.

This is unfortunate, because that means it doesn't give you the needed access to read the certificate, which doesn't make sense to me.  Reading the certificate is fine; all machines need to do that in order to trust the CA; and since this is being installed in the **user's** trust store, it is, by extension, readable by the user, too.  It's just the private key that needs to be protected, and `devcert` does that just fine.

So, on Windows, you can export the certificate, and use openssl to convert it to `pem` format, or you can use `devcert`:
```bash
node -e "require(`devcert/dist/platforms`).default.readProtectedFile(`C:\\path\\to\\localAppData\\devcert\\certificate-authority\\certificate.cert`).then(console.log);"
```
On linux, you can do:
`sudo cat /usr/local/share/ca-certificates/devcert.crt`

The important takeaway here is that the trusted certificate authority needs to be included in the https server's globalAgent CAs.  That can be a path name, or it can be a `readFile` buffer, but the file does need to be readable, no matter how you do it.

It would be impossible to make this production ready from `gatsby` side.  You can't hardcode certificates, for one.  Any time you wanted to read the certificate authority through `devcert`, you'd be entering a password; so, every single time you started the dev server.  And in order to read from the OS trust store, it would require far more work than is necessary, and on Linux, it would still require a password every. single. time.

I think it would probably be better to submit a PR over @ `devcert` that does away with encrypting the CA cert, and returns the CA certificate data buffer along with the domain cert/key.  Then it can be picked up from `gatsby`, and added to the globalAgent's CAs.

One caveat to **that** approach:
The `ca` property would have to be removed from the `devcert` return value before it is fed to `https.createServer()`.  Putting a `ca` value there is for authenticating peers, and that's not what we're trying to do.  Probably the best way would be to do `program.ssl.ca = undefined` before it goes to `https.createServer()`, since `delete` has some interesting effects, which impact performance.

---

This exposes more issues.  `gatsby` is trying to start the server at `localhost` domain, but `devcert` only includes the domain that is provided to it as the subject, and `*.<domain>` as subject alternative names.  That means starting the server still fails because `localhost` is not in the SANs.

`devcert` **does** add the domain to the local `hosts` file, so that it can be resolved, however.  So if `gatsby` started the server on the same host as it gives to `devcert`, then it should work, theoretically.

Alternatively, we could submit a PR to `devcert` to include `localhost` in the subject alternative names [here](https://github.com/davewasmer/devcert/blob/master/openssl-configurations/domain-certificates.conf#L37-L39).

Js-Brecht:
> This exposes more issues. gatsby is trying to start the server at localhost domain, but devcert only includes the domain that is provided to it as the subject, and *.<domain> as subject alternative names. That means starting the server still fails because localhost is not in the SANs.
> 
> devcert does add the domain to the local hosts file, so that it can be resolved, however. So if gatsby started the server on the same host as it gives to devcert, then it should work, theoretically.
> 
> Alternatively, we could submit a PR to devcert to include localhost in the subject alternative names here.

Another alternative would be this: [develop.js](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/commands/develop.js#L338-L345)
```diff
  if (program.https) {
    program.ssl = await getSslCert({
-     name: program.sitePackageJson.name.replace(`@`, ``).replace(`/`, `--`),
+     name: `localhost`,
      certFile: program[`cert-file`],
      keyFile: program[`key-file`],
      directory: program.directory,
    })
  }
```

This is a development server.  It's not going into production, so, IMO, it doesn't really need to use the project name as the host.  Besides, it tells the user to go to `https://localhost:<whatever>` anyway.  Also, the certificate doesn't really need to be different for every single project, does it?

Js-Brecht:
_Thought I'd post this here, since it would affect this PR, and can't be a PR of its own...._

You guys care to give [this branch](https://github.com/Js-Brecht/gatsby-1/tree/devcert-upgrade) in my dev repo a spin?

I'm working on a PR to fix some of the issues in `devcert`, and at the same time, I figured I'd fix the `unable to verify the first certificate` error.  I have tested it on Windows, and it works for me.  Still get a `SEC_ERROR_BAD_SIGNATURE` error in Firefox.  Will need to spend some time tracking that down before work on it is finished.

I've included the package in the gatsby monorepo (`gatsby-devcert`) for testing purposes.  It's not going to remain.

I don't have access to my Linux machine from where I'm at right now, so cannot test with it.  And I'm going on vacation tomorrow, so I doubt I'll get around to that till I get back.

What I've done so far with `devcert` aka `gatsby-devcert` :
* Use `prompts` for password prompt in Windows (the other package that was being used (`password-prompt`, is buggy, and not worth fixing).
* Remove the process that uses Firefox to try to trust CA in Windows.  It doesn't work... and, honestly, shouldn't be necessary anyway.
* Eliminate encrypting certificate authority Cert on Windows.  Need to be able to read it, and only the key needs to be encrypted.  **_Users_** can read CA certs as much as they want, anyway, directly from the root store... in fact, they HAVE to, in order for their browsers to trust the certs, so encrypting that cert is worthless; just a waste of time & resources, and causes unnecessary complication.
* Reduce file perm restrictions on CA certificates in Linux & Darwin.  Again, same thing.  Just the key needs to be protected, because that's what's used for signing certs.
    * File perms will now be 666 for the cert, and 600 for the key (key process doesn't change).
* Return the CA certificate in a tuple as a `readFile` buffer along with the other `key` and `cert` values that are passed to `https.createServer`.
    * Return value is now [ca, { key, cert }]
* Fixed various other latent bugs having to do with paths names, and the havoc spaces would cause
* **TODO**
    * `crypto.createCipher` is deprecated.  Should probably use a different encryption module
    * Track down the cause of the `SEC_ERROR_BAD_SIGNATURE` Firefox error.
        * Most likely an `openssl` misconfiguration somewhere

The changes I've made so far would require additional updates to `gatsby`
What I've done so far to accommodate the changes in `devcert`:
* `get-ssl-cert.js` and `develop.js`
    * Receive tuple ssl values from `devcert`
* `develop.js`
    * Changed `getSslCerts()` hostname argument to `localhost`
    * Setup global CA values with received  CA certificate before `createServer()`

There is no way I would want to submit this as a PR to `gatsby`, because of the extra `gatsby-devcert` package, and because of the changes that conflict with this PR; and I'm not yet ready to submit a PR to `devcert`, because I want to make sure it's working 100% with `gatsby` before I split it up again.  If anybody wants to contribute, I'd can add you as a collaborator

Js-Brecht:
@wardpeet So, it looks like `devcert` is a pretty stale project.  There's PRs fixing bugs and adding features that have just been sitting there for months... quite a few more than a year old.  I don't even see any action in the issues.  It's like a ghost town over there. It hurts my brain to think about using a project that has these issues, plus isn't even being maintained.

I'd like to get your thoughts on building a new Gatsby module/plugin that would handle SSL certs.  I've looked at a couple projects that we might be able to use:
* [pem](https://www.npmjs.com/package/pem)
* [trust-cert](https://www.npmjs.com/package/trust-cert)
* [selfsigned-ca](https://github.com/MikeKovarik/selfsigned-ca)

I haven't actually tried to put anything together with them, but they seem promising... and they seem to be a bit more actively maintained.  There's quite a few projects out there dealing with SSL Certificates, so I just kind of skimmed the top.  That last one _might_ do everything that needs to be done; it would just be a bit more involved than what is already being used, but also wouldn't be so much more complex that it would need it's own plugin (could do it in one module).

I suppose an alternative could be to fork `devcert`, fix the issues, and use the fork instead.... but that feels so much like plagiarism to me 😬

Js-Brecht:
@wardpeet I'm pretty sure this can be merged, yeah?  I could run another test, just to make sure it doesn't cause any issues... the other problems that were noticed before are kind of a separate issue.

Merging this would at least close #16212.  The other issues with certificates should be fixed by https://github.com/davewasmer/devcert/pull/41 if/when it gets merged (and #18703, by extension).

One caveat:
Issues will begin to crop up for users of MacOS Catalina: https://github.com/davewasmer/devcert/issues/39
Fix is here https://github.com/davewasmer/devcert/pull/45

rmorabia:
Sorry I ghosted on this -- I couldn't find the issue, I assume it's the package.json/yarn.lock stuff. Up to y'all what you want to do on this.

Js-Brecht:
Oh... I forgot... This upgrade would break the `NODE_EXTRA_CA_CERTS` workaround for #14990.  The new version locks down the certificate on the disk, so it can't be read 😤

Just trying to some of these things off my list, and since this PR conflicts with #18703, it stays.

Js-Brecht:
@rmorabia This should fix the issue with the tests 👇 
<details>
<summary>Diff output... see attached if you want to do a patch</summary>

```diff
diff --git a/packages/gatsby/src/utils/__tests__/get-ssl-cert.js b/packages/gatsby/src/utils/__tests__/get-ssl-cert.js
index 7191c7496..8ef7651b6 100644
--- a/packages/gatsby/src/utils/__tests__/get-ssl-cert.js
+++ b/packages/gatsby/src/utils/__tests__/get-ssl-cert.js
@@ -11,13 +11,13 @@ jest.mock(`gatsby-cli/lib/reporter`, () => {
     info: jest.fn(),
   }
 })
-jest.mock(`devcert-san`, () => {
+jest.mock(`devcert`, () => {
   return {
-    default: jest.fn(),
+    certificateFor: jest.fn(),
   }
 })

-const devcertSan = require(`devcert-san`).default
+const { certificateFor } = require(`devcert`)
 const reporter = require(`gatsby-cli/lib/reporter`)
 const getSslCert = require(`../get-ssl-cert`)

@@ -25,7 +25,7 @@ describe(`gets ssl certs`, () => {
   beforeEach(() => {
     reporter.panic.mockClear()
     reporter.info.mockClear()
-    devcertSan.mockClear()
+    certificateFor.mockClear()
   })
   describe(`Custom SSL certificate`, () => {
     it.each([[{ certFile: `foo` }], [{ keyFile: `bar` }]])(
@@ -60,11 +60,11 @@ describe(`gets ssl certs`, () => {
   describe(`automatic SSL certificate`, () => {
     it(`sets up dev cert`, () => {
       getSslCert({ name: `mock-cert` })
-      expect(devcertSan).toBeCalledWith(`mock-cert`, { installCertutil: true })
+      expect(certificateFor).toBeCalledWith(`mock-cert`, { installCertutil: true })
       expect(reporter.info.mock.calls).toMatchSnapshot()
     })
     it(`panics if certificate can't be created`, () => {
-      devcertSan.mockImplementation(() => {
+      certificateFor.mockImplementation(() => {
         throw new Error(`mock error message`)
       })
       getSslCert({ name: `mock-cert` })
```
[get-ssl-cert.diff.txt](https://github.com/gatsbyjs/gatsby/files/3872170/get-ssl-cert.diff.txt)

</details>

The other issue (https://github.com/gatsbyjs/gatsby/pull/16726#issuecomment-523103877) I think was because of only doing an install, not a `yarn bootstrap`.  I don't have that problem if I do a full bootstrap.  Should have been obvious 😆

rmorabia:
I'll fix the tests tonight/tomorrow night.

rmorabia:
Ahh, still getting failing tests because `devcert` isn't installed. I thought I ...would've done that for this PR to exist in the first place. I've lost track of this since I wrote it so long ago. Not sure what I'm supposed to do here.

packages/gatsby/package.json
```diff
@@ -47,23 +47,24 @@
...
-    "eslint-plugin-jsx-a11y": "^6.0.3",
-    "eslint-plugin-react": "^7.8.2",
-    "event-source-polyfill": "^1.0.5",
-    "express": "^4.16.3",
-    "express-graphql": "^0.7.1",
+    "css-loader": "^1.0.1",
+    "debug": "^3.2.6",
+    "del": "^5.1.0",
+    "detect-port": "^1.3.0",
+    "devcert-san": "^0.3.3",
```
Js-Brecht: Here is the problem.  When you merged the changes from the master branch into your fork, it overwrote your changes to `package.json`.  It has reverted back to the original `devcert-san`

packages/gatsby/src/utils/__tests__/get-ssl-cert.js
```diff
@@ -11,21 +11,21 @@ jest.mock(`gatsby-cli/lib/reporter`, () => {
...
-jest.mock(`devcert-san`, () => {
+jest.mock(`devcert`, () => {
   return {
-    default: jest.fn(),
+    certificateFor: jest.fn(),
   }
 })
 
-const devcertSan = require(`devcert-san`).default
+const certificateFor = require(`devcert`)
```
Js-Brecht: This should either be destructured, or aliased.  Right now it's importing as a namespace.  This will likely make the test fail further down.

`devcert` is written in typescript, and using `modules.export = function()` is generally frowned upon; it uses named exports for all of its functions.  So this namespace won't be a function you can call; instead it'll be an object containing the functions.  You'll notice that the `jest.mock('devcert')` module above is set up to mirror `devcert`'s structure.  It returns an object with a function, not a function.

You could do one of these:
```ts
// destructure the import
const { certificateFor } = require(`devcert`);
// or, in keeping with the source get-ssl-cert.js
const certificateFor = require(`devcert`).certificateFor;
// If you wanted the function name to fit its purpose a bit more clearly, you could alias it
const getDevCert = require(`devcert`).certificateFor;
```

Js-Brecht:
There's other things that might need to happen for this update to go smoothly.

For one, the common name of the certificate generated will match the project name; meanwhile, `gatsby develop --https` will report that the server is running on `https://localhost:<port>`.  This means that if the user opens the url that `gatsby develop` reports, then an error will be shown indicating that the certificate is invalid for that domain.

@rmorabia This is something that I address in PR #18703, but if you wanted to fix it here, I will make sure mine aligns with yours.

https://github.com/gatsbyjs/gatsby/pull/18703/files#diff-346c3005d97c1ca0b5efb170af1b43f6R358-R367

Another problem: the password prompt that's generated for Windows users doesn't work very well.  Not a very big deal, and a fix for it is already part of #18703.  It will probably hold until then.

---

Progress is being made on `devcert`, so it shouldn't be much longer before all of the certificate issues can be fixed.

rmorabia:
I'm going to hold off on making fixes outside of just making this fix work. I'll continue working on it tonight since things still seem to be failing.

Js-Brecht:
AFAICT, the failing test doesn't have anything to do with this PR.  Just a security audit; the `mem` package has a security vulnerability.

LekoArts:
Hi @rmorabia 👋 Do you have time to rebase your PR onto master and fix the merge conflict?

rmorabia:
> Hi @rmorabia 👋 Do you have time to rebase your PR onto master and fix the merge conflict?

Done!

wardpeet:
Fixed in `gatsby@2.19.24`

Issue #16287, opened 2019-08-01
[Firefox] Unhandled promise rejection TypeError: "this.loadPageDataJson(...).then(...).finally is not a function


arnauguadall:
## Description

When I run in localhost works ok with all the browsers including Firefox but when it goes to production the Firefox v68.0.1 looks misaligned caused (I guess) for this `.finally` function.

### Steps I tried

I tried to use `"gatsby-plugin-compile-es6-packages": "^2.1.0"` to compile es6 features like they are using in #2177 but without luck.

Also I tried:
```
exports.modifyWebpackConfig = ({ config, stage }) => {
  switch (stage) {
      case "build-javascript":
          const app = config._config.entry.app;
          config._config.entry.app = [require.resolve("./polyfills"), app];

          break;
      default:
          break;
  }

  return config;
};
```

### Expected result

What should happen?

Works ok with Chrome (v75.0.3770.142).

![Chrome version](https://i.imgur.com/zwgqFMQ.png)

### Actual result

Returns a console error which for what I found it's something that `.finally()` function is quite new and maybe is not fully supported yet.

```
Unhandled promise rejection TypeError: 
"this.loadPageDataJson(...).then(...).finally is not a function"
```

How it looks like in Firefox (v68.0.1)
![Firefox version](https://i.imgur.com/anYl0YJ.png)

Also I'm using  `"antd": "^3.12.3"` package to handle the UI and in this CSS class if you change `flex-direction: column` to `row` it looks ok but the error still persists.

```
.ant-layout {
    display: flex;
    flex: auto;
    flex-direction: column;
    min-height: 0;
    background: #f0f2f5;
```

### Environment

```
System:
    OS: macOS 10.14.5
    CPU: (8) x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 10.16.0 - /usr/local/bin/node
    Yarn: 1.16.0 - /usr/local/bin/yarn
    npm: 6.9.0 - /usr/local/bin/npm
  Languages:
    Python: 2.7.10 - /usr/bin/python
  Browsers:
    Chrome: 76.0.3809.87
    Firefox: 68.0.1
    Safari: 12.1.1
  npmPackages:
    gatsby: ^2.0.76 => 2.13.45
    gatsby-image: ^2.0.20 => 2.2.7
    gatsby-mdx: 0.6.3 => 0.6.3
[...]

PR #16734, opened 2019-08-19
fix(gatsby): remove finally to enable ie11 + firefox support

peter-mouland:
## Description

since Gatsby 12.2.1 firefox + ie11 have been broken due to the use of `.finally`.

I've opted not to use `async/await` as that would  change the behaviour of the function, instead i've dupe'd the `finally` code in a `then` and `catch`.  This mean there is no change in code behaviour.

## Related Issues

fixed https://github.com/gatsbyjs/gatsby/issues/16287


peter-mouland:
thanks for the feedback - i've updated 👍

PR #16746, opened 2019-08-19
fix(gatsby-image): Fix eager loading

### LINKED ISSUE #14972: gatsby-image loading="eager" does not work, only critical={true} works
## Description

I'm trying to enable eager loading on images using gatsby-image.

### Steps to reproduce

1. npm i --save gatsby-image@latest
2. Import image from gatsby graphql w/ webp 
3. Place below the fold and turn off chrome cache, and scroll down to image:
```
<Image loading="eager" />
<Image criticial />
```
### Expected result

Critical and loading="eager" behaves the same way

### Actual result

Critical loads the image on DOM entry, loading="eager" behaves like loading="lazy"

### Other info

It looks like the gatsby-image source still references the props.critical instead of checking props.loading.

https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-image/src/index.js#L216

Oddly there is a warning that gets thrown here to switch to the new loading prop vs eager:

https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-image/src/index.js#L36

sagarnasit:
## Description

Eager load image which has `loading="eager"` props defined. Previously eager loading only worked with `critical` prop even if `loading="eager"` is defined.

## Related Issues

Fixes #14972


sagarnasit:
Anyone available to validate this PR. Please let me know if any changes or addition needed. Thank you :)

sidharthachatterjee:
Published in `gatsby-image@2.2.9`

sagarnasit:
Thank you @sidharthachatterjee. I would love to contribute more to gatsbyJS, Let me know if i can help with anything.

sidharthachatterjee:
Thanks @sagarnasit 💜 Take a look at issues labelled _help wanted_ or _good first issue_

SilencerWeb:
Quick questions, does `loading="eager"` works the same way as `critical` but it is supported natively by the browser?

And now `critical` doesn't work

sidharthachatterjee:
@SilencerWeb Yes, loading="eager" uses the native lazy loading recently launched in Chrome and falls back to Gatsby's Intersection Observer based implementation if the current browser doesn't support it. 

You can still use `critical` for this but it is deprecated and will be removed in the next major version. 

In case `critical` isn't working for you, please open an issue with a reproduction. Thanks!

glennreyes:
I think this fix causes the bug not transitioning from the blurred to the actually image, which I reported [here](https://github.com/gatsbyjs/gatsby/issues/17007). Passing `loading="eager"` is working for me as it should in 2.2.8. Is there any specific edge case for this bug?

PR #16747, opened 2019-08-19
chore(telemetry): make interval configurable for easier local testing

jamo:


packages/gatsby-telemetry/src/index.js
```diff
@@ -7,7 +7,10 @@ process.on(`exit`, flush)
 
 // For longrunning commands we want to occasinally flush the data
 // The data is also sent on exit.
-const interval = 10 * 60 * 1000 // 10 min
+
+const interval = process.env.TELEMETRY_BUFFER_INTERVAL
+  ? Number(process.env.TELEMETRY_BUFFER_INTERVAL)
```
pieh:
```suggestion
  ? Math.max(Number(process.env.TELEMETRY_BUFFER_INTERVAL), 1000)
```

just sanity clamp to not spawn flush process too often if wrongly configured - 1 second is my suggestion, feel free to adjust

jamo: Great idea for a safeguard, thanks!

packages/gatsby-telemetry/src/index.js
```diff
@@ -7,7 +7,10 @@ process.on(`exit`, flush)
 
 // For longrunning commands we want to occasinally flush the data
 // The data is also sent on exit.
-const interval = 10 * 60 * 1000 // 10 min
+
+const interval = process.env.TELEMETRY_BUFFER_INTERVAL
```
m-allanson: Could you add some checks to ensure that interval is always a number? If it ends up as `NaN` or any falsy value then the timeouts will fire immediately.

m-allanson: Oh just saw @pieh's comment :) I think this could still cause problems depending on the input. e.g: `Math.max(Number('oops this is a string'), 1000)` will be `NaN`.

jamo: Indeed

jamo: Updated, is the Number.isFinite too much of a hack?

m-allanson: Nah that should be good, thanks!

DSchau: Available in Node 0.10 :) Not _that_ hacky ™️

PR #16766, opened 2019-08-20
fix(gatsby): keep all page-data files in public

### LINKED ISSUE #16645: Building static HTML failed for path "/404.html"
## Description

First of all, I saw a number of similar issues but those topics are now closed and none of the proposed solutions seem to work so far.

The problem: I'm localizing my blog and the 404 page is the only page having issues building.
The build fails with the following message:

> Building static HTML failed for path "/404.html"

>   Error: ENOENT: no such file or directory, open 'D:\...\public\page-data\404.html\page
>   -data.json'

**This happens only from the second build forth**, meaning that the first build after _gatsby clean_ goes fine. Moreover, _gatsby develop_ only has problems displaying 404 pages when I don't run _gatsby clean_ before.

I don't know if it can result in serious problems going forward.

### Steps to reproduce

Clone the repository here: [gatsby-bug-demo](https://github.com/pleopardi/gatsby-bug-demo)

Run `yarn` and then `yarn prod:test` to reproduce the error.

If you checkout to the fourth last commit (hash: _d0e239f7f44a3273fa4a5d088e2abeaa8e56f4ac_), right before starting to localize the 404 page, the issue disappears.

### Expected result

I expect the build to end without problems.

### Actual result

The build fails with the error above.

### Environment
Happens on both MacOs and Windows. on Windows: 

System:
  OS: Windows 10
  CPU: (8) x64 Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz
Binaries:
  Node: 10.16.0 - D:\Programmi\nodejs\node.EXE
  Yarn: 1.16.0 - D:\Programmi\Yarn\bin\yarn.CMD
  npm: 6.9.0 - C:\Users\leopa\AppData\Roaming\npm\npm.CMD
Languages:
  Python: 2.7.16
Browsers:
  Edge: 44.18362.1.0
npmPackages:
  gatsby: ^2.13.42 => 2.13.42
  gatsby-image: ^2.2.7 => 2.2.7
  gatsby-plugin-emotion: ^4.1.2 => 4.1.2
  gatsby-plugin-google-analytics: ^2.1.4 => 2.1.4
  gatsby-plugin-mdx: ^1.0.0 => 1.0.18
  gatsby-plugin-react-svg: ^2.1.1 => 2.1.1
  gatsby-plugin-webfonts: ^1.1.0 => 1.1.0
  gatsby-remark-images: ^3.1.7 => 3.1.7
  gatsby-transformer-json: ^2.2.2 => 2.2.2
  gatsby-transformer-remark: ^2.6.10 => 2.6.10
  gatsby-transformer-sharp: ^2.2.5 => 2.2.5

wardpeet:
## Description
We remove all {.html,.css} files from public which also includes page-data files that end with .html. For example page-data/page-3.html/page-data.json also gets removed. This pr fixes it.

## Related Issues
Related to #16354
Fixes #16645

### DIFF
--- a/packages/gatsby/src/bootstrap/index.js
+++ b/packages/gatsby/src/bootstrap/index.js
@@ -145,9 +145,8 @@ module.exports = async (args: BootstrapArgs) => {
     )
     activity.start()
     await del([
-      `public/*.{html,css}`,
       `public/**/*.{html,css}`,
-      `!public/page-data/404.html`,
+      `!public/page-data/**/*`,
       `!public/static`,
       `!public/static/**/*.{html,css}`,
     ])


sidharthachatterjee:
Published in `gatsby@2.13.71`

drake-smith:
I see that this PR fixes crashing the build, and indeed the build does not break on this and I get, say, a file in public that's `page-data/page-3.html/page-data.json`, but when it's deployed in production (in my case, tested via gatsby serve, on Netlify and on AWS S3/CloudFront), I get 404 for page-data.json files in my Network tab because it is looking for `page-data/page-3/page-data.json`, not `page-data/page-3.html/page-data.json`. For some reason the request to that URL omits the `.html` part. 

Referencing my open bug report below that includes a manual workaround (aka I cannot use my CI/CD setup to automatically deploy changes to production and have to go into each build's public folder to change the <link href= /> URL for these files.
#21252

Issue #16434, opened 2019-08-07
gatsby develop -H 0.0.0.0 --open does not open correct page on Windows


crhistianramirez:
## Description
`--open` does not play nicely with `-H`. Sends me to and prints that I should be able to access local dev server on `http://0.0.0.0:8000/` however I get a "site can't be reached message"

### Steps to reproduce
1. run `gatsby develop -H 0.0.0.0 --open`

### Expected result
Should open my app

What should happen?
Should open my app

### Actual result
Renders a non functional page

### Environment

System:
    OS: Windows 10
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
  Binaries:
    Node: 10.16.0 - C:\Users\CRHIST~1\AppData\Local\Temp\yarn--1565192299909-0.8662975271117463\node.CMD
    Yarn: 1.13.0 - C:\Users\CRHIST~1\AppData\Local\Temp\yarn--1565192299909-0.8662975271117463\yarn.CMD
    npm: 6.9.0 - C:\Program Files\nodejs\npm.CMD
  Languages:
    Python: 3.7.1 - C:\Users\CrhistianR\AppData\Local\Programs\Python\Python37-32\python.EXE
  Browsers:
    Edge: 42.17134.1.0
  npmPackages:
    gatsby: ^2.13.38 => 2.13.39
    gatsby-image: ^2.2.0 => 2.2.6
    gatsby-plugin-catch-links: ^2.1.0 => 2.1.2
    gatsby-plugin-manifest: ^2.2.0 => 2.2.4
    gatsby-plugin-material-ui: ^2.0.1 => 2.1.4
    gatsby-plugin-mdx: ^1.0.7 => 1.0.16
    gatsby-plugin-offline: ^2.2.0 => 2.2.4
    gatsby-plugin-react-helmet: ^3.1.0 => 3.1.2
    gatsby-plugin-sharp: ^2.2.1 => 2.2.9
    gatsby-plugin-typescript: ^2.1.0 => 2.1.2
    gatsby-plugin-web-font-loader: ^1.0.4 => 1.0.4
    gatsby-remark-autolink-headers: ^2.1.0 => 2.1.3
    gatsby-remark-images: ^3.1.2 => 3.1.7
    gatsby-remark-prismjs: ^3.3.0 => 3.3.3
    gatsby-source-filesystem: ^2.1.6 => 2.1.6
[...]

PR #16769, opened 2019-08-20
fix(gatsby-cli): Open Lan URL if Windows

sidharthachatterjee:
Because Windows can't handle http://0.0.0.0 😞 

Fixes https://github.com/gatsbyjs/gatsby/issues/16434


packages/gatsby/src/commands/develop.js
```diff
@@ -545,15 +545,15 @@ module.exports = async (program: any) => {
...
-    // console.log(chalk.green(`Compiled successfully!`))
-    // }
-    // if (isSuccessful && (isInteractive || isFirstCompile)) {
     if (isSuccessful && isFirstCompile) {
       printInstructions(program.sitePackageJson.name, urls, program.useYarn)
       printDeprecationWarnings()
       if (program.open) {
-        Promise.resolve(openurl(urls.localUrlForBrowser)).catch(err =>
+        const urlToOpen =
+          process.platform === `win32`
```
wardpeet: I would fix this in prepareUrls where we convert 0.0.0.0 to localhost. We should also do the isWSL check as we'll have the same issue.

CRA does this too
https://github.com/facebook/create-react-app/blob/eaf0b176e2d2af4f9a1f092caf5fd7356cd73844/packages/react-dev-utils/WebpackDevServerUtils.js#L44

wardpeet:
image

PR #16801, opened 2019-08-20
feat(blog): Add blog post for implementing microfrontends with Ara framework

marconi1992:
## Description

 
Article about how to implement microfrontends in GatsbyJS using Ara Framework.


sidharthachatterjee:
Thanks so much for taking the time to write this @marconi1992 

Unfortunately, we won't be accepting this submission. We believe this best belongs in your personal blog because it seems like the article is not relevant for the entire Gatsby blog audience.

This, however, doesn't mean that it isn't a valuable blog post. We would love to see it published on your blog (if it isn't already)

Thank you once again for taking the time to write this and for using Gatsby 💜

PR #16814, opened 2019-08-20
fix(gatsby-plugin-feed): respect custom options and require title in config

### LINKED ISSUE #16177: Gatsby feed plugin ignores custom options
## Description

The `gatsby-plugin-feed` completely ignores all custom feed / channel properties.

```js
{
  resolve: 'gatsby-plugin-feed',
  options: {
    feeds: [
      {
        output: "/rss.xml",
        title: "I am ignored :(", // ignored
        language: "en", // ignored
      },
    ],
  }
}
```

## Suspect

I believe this is a bug in the default `setup()` method:
https://github.com/gatsbyjs/gatsby/blob/00fd5faa31fefce22df7a61aa8482b44a147aa01/packages/gatsby-plugin-feed/src/internals.js#L32
`...rest` should be outside the `query` object.

## Workaround

A workaround is simply to override the default `setup` method with a fixed version:
```js
{
  resolve: 'gatsby-plugin-feed',
  options: {
    setup:
      ({
        query: {
          site: { siteMetadata },
        },
        ...rest // outside of the query
      }) => {
        return {
          ...siteMetadata,
          ...rest,
        }
      }
    ,
    feeds: [
      {
        output: "/rss.xml",
        title: "They see me :)", // not ignored
        language: "en", // not ignored
      },
    ],
  }
}
```

### Environment

```txt
$ gatsby info --clipboard
  System:
    OS: Linux 4.4 Ubuntu 16.04.6 LTS (Xenial Xerus)
    CPU: (4) x64 Intel(R) Core(TM) i7-5600U CPU @ 2.60GHz
    Shell: 4.3.48 - /bin/bash
  Binaries:
    Node: 10.16.0 - ~/n/bin/node
    npm: 6.9.0 - ~/n/bin/npm
  Languages:
    Python: 2.7.12 - /usr/bin/python
  npmPackages:
    gatsby: ^2.0.22 => 2.8.6
    gatsby-image: ^2.2.1 => 2.2.1
    gatsby-plugin-catch-links: ^2.0.4 => 2.0.15
    gatsby-plugin-feed: ^2.0.8 => 2.2.3
    gatsby-plugin-google-analytics: ^2.0.21 => 2.0.21
    gatsby-plugin-google-fonts: latest => 1.0.1
    gatsby-plugin-layout: ^1.1.2 => 1.1.2
    gatsby-plugin-netlify: ^2.0.17 => 2.0.17
    gatsby-plugin-offline: ^2.1.1 => 2.1.1
    gatsby-plugin-react-helmet: ^3.0.0 => 3.0.12
    gatsby-plugin-robots-txt: ^1.4.0 => 1.4.0
    gatsby-plugin-sass: ^2.0.1 => 2.0.11
    gatsby-plugin-sharp: ^2.0.6 => 2.1.3
    gatsby-plugin-sitemap: ^2.1.0 => 2.1.0
    gatsby-remark-copy-linked-files: ^2.0.5 => 2.0.13
    gatsby-remark-images: ^2.0.4 => 2.0.6
    gatsby-remark-prismjs: ^3.2.11 => 3.2.11
    gatsby-remark-reading-time: ^1.0.1 => 1.0.1
    gatsby-remark-responsive-iframe: ^2.0.5 => 2.1.1
    gatsby-remark-smartypants: ^2.0.5 => 2.0.9
    gatsby-source-filesystem: ^2.0.3 => 2.0.38
    gatsby-transformer-remark: ^2.3.12 => 2.3.12
    gatsby-transformer-sharp: ^2.1.3 => 2.1.21
  npmGlobalPackages:
    gatsby-cli: 2.6.13
```

**Note** that this is a WSL installation on Windows 10.0.17134.799.

eyalroth:
Fixes #16177.

Note that the plugin will now require a title to be configured for each feed, as [per the documentation](https://www.gatsbyjs.org/packages/gatsby-plugin-feed/).


packages/gatsby-plugin-feed/src/internals.js
```diff
@@ -29,11 +29,13 @@ export const defaultOptions = {
   setup: ({
     query: {
       site: { siteMetadata },
-      ...rest
     },
+    generator,
```
sidharthachatterjee: This doesn't seem necessary! 

```suggestion

```

packages/gatsby-plugin-feed/src/internals.js
```diff
@@ -29,11 +29,13 @@ export const defaultOptions = {
...
+    ...rest
   }) => {
     return {
       ...siteMetadata,
+      generator,
```
sidharthachatterjee:
```suggestion

```

packages/gatsby-plugin-feed/src/internals.js
```diff
@@ -29,11 +29,13 @@ export const defaultOptions = {
   setup: ({
     query: {
       site: { siteMetadata },
-      ...rest
     },
+    generator,
+    ...rest
```
sidharthachatterjee: So it looks like this does indeed fix the issue! Took a while for us to catch it so we need it now!

packages/gatsby-plugin-feed/src/internals.js
eyalroth: This is necessary; otherwise, the default `generator` value "GatsbyJs" will be ignored.

davidbailey00: Functionally it will behave the same now that you've fixed the issue with `...rest` being in the wrong place - I've checked that all your tests still pass just to make sure. Maybe it was only necessary before you fixed the primary issue?

eyalroth: @davidbailey00 Interestingly enough, that is true. It's funny because that only showed up on the tests _after_ I had already fixed the primary issue. Oh well, who knows why, I removed it.

sidharthachatterjee:
Published in `gatsby-plugin-feed@2.3.8`

PR #16820, opened 2019-08-20
Add excerpt_separator option to gatsby-plugin-mdx

riywo:
## Description

This commit adds `excerpt_separator` option to `gatsby-plugin-mdx` just like `gatsby-transformer-remark`.

## Related Issues

N/A


ChristopherBiscardi:
Thanks for taking a crack at this! Unfortunately this is an approach that introduces complexity that doesn't belong in gatsby-plugin-mdx. To be more specific: gray-matter, which we already use in gatsby-plugin-mdx, includes support for this sort of separator and we should lean on the library for this processing instead of writing custom markdown node types, etc processing. See https://github.com/ChristopherBiscardi/gatsby-mdx/issues/386 for more information on that.

riywo:
Ok, I’ll look into `gray-matter` and submit a new PR.

PR #16844, opened 2019-08-21
tests: Move Windows unit tests to CircleCI

sidharthachatterjee:
Because Azure pipelines are too slow

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,5 +107,8 @@
   },
   "workspaces": [
     "packages/*"
-  ]
+  ],
+  "resolutions": {
+    "write-file-atomic": "2.4.1"
+  }
 }
--- a/package.json
+++ b/package.json
@@ -110,5 +110,10 @@
   ],
   "resolutions": {
     "write-file-atomic": "2.4.1"
+  },
+  "comments": {
+    "resolutions": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -107,5 +107,8 @@
   },
   "workspaces": [
     "packages/*"
-  ]
+  ],
+  "resolutions": {
+    "write-file-atomic": "2.4.1"
```
wardpeet: Any chance you can elaborate on this change? :)

sidharthachatterjee: Oh right! Sorry! Jest has a known issue that causes a race condition on Windows and breaks some tests. This is the recommended fix in the issue that tracks this bug (https://github.com/facebook/jest/issues/4444)

package.json
```diff
@@ -107,5 +107,8 @@
...
+  }
```
wardpeet: We could add something like this to keep track on the error. Maybe creating an issue in gatsby and tagging it with upstream is a good place too 😛 
```suggestion
  },
  "comments": {
    "resolutions": {
      "write-file-atomic": "Jest bug on windows: https://github.com/facebook/jest/issues/4444"
    }
  }
```

sidharthachatterjee: Documented in https://github.com/gatsbyjs/gatsby/issues/16885

PR #16887, opened 2019-08-21
feat: Use gatsby clean for the build command

endormi:
## Description

Using gatsby clean for the build command on the examples to remove cache rather than doing it manually if needed.

### DIFF
--- a/examples/client-only-paths/package.json
+++ b/examples/client-only-paths/package.json
@@ -24,7 +24,7 @@
   "main": "n/a",
   "scripts": {
     "develop": "gatsby develop",
-    "build": "gatsby build",
+    "build": "gatsby build && gatsby clean",
     "start": "npm run develop"
   }
 }
--- a/examples/feed/package.json
+++ b/examples/feed/package.json
@@ -19,7 +19,7 @@
   "main": "index.js",
   "scripts": {
     "develop": "gatsby develop",
-    "build": "gatsby build",
+    "build": "gatsby build && gatsby clean",
     "start": "npm run develop"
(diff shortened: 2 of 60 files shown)


sidharthachatterjee:
This shouldn't be necessary. We only want to delete `.cache` and `public` (via `gatsby clean`) when needed and not after every build.

endormi:
Okay, thanks for clarifying

2x2xplz:
There are some builds which seem to fail more than half the time without clearing the cache. There are also environments (ahem, Netlify) where developers don't have an entire CLI, just a single command we can execute to build our projects. While I agree that clearing the cache shouldn't be the default, an option like `gatsby build --clean` should be possible which would internally, run `gatsby clean` then `gatsby build`

endormi:
Thoughts @sidharthachatterjee ?

m-allanson:
Hey @2x2xplz!

If there's an issue with builds failing intermittently, let's try and find out the root cause and fix it. If you're able to create a [minimal reproduction](https://www.gatsbyjs.org/contributing/how-to-make-a-reproducible-test-case/) that regularly fails to build, please open an issue and link to it.

In the mean time, for environments where you can just run a single command there are workarounds. You could add a command to your `package.json` file that runs a couple of commands together. An example might look like:

```js
// package.json
// ...
"scripts": {
    "clean-build": "gatsby clean && gatsby build"
},
// ..etc
```

There are packages like [npm-run-all](https://github.com/mysticatea/npm-run-all) that provide more control.

I hope that helps!

PR #16913, opened 2019-08-21
feat(gatsby-transformer-react-docgen): vendor in react-docgen-displayname-handler

jquense:
## Description

Vendors this particular handler so it's easier to keep up to date. The original project is fairly dead, and this seems easier/better than trying to revive it given how little code there is. Should remove some peer dep warnings


packages/gatsby-transformer-react-docgen/src/__tests__/displayname-handler.js
```diff
@@ -0,0 +1,150 @@
+import test from "ava"
```
wardpeet: I believe this fixes jest. It would be nice if you could transform it into a more gatsby like test with `describe` and `it`
```suggestion
```

jquense:
oops, updated

Issue #15544, opened 2019-07-09
Allow themes to modify existing GraphQL types


wangyi7099:
I have a graphQL query like this:
```js
export const pageQuery = graphql`
  query TemplateDocsMarkdown($slug: String!) {
    allMdx(
      sort: { fields: fields___slug, order: DESC }
      filter: { fields: { slug: { glob: "/docs/*" } } }
    ) {
      edges {
        node { 
          frontmatter {
            title
            type
            order
          }
        }
      }
    }
  }
`;
```

But if my `md` file doesn't have a frontmatter `type` . then gatsby will show the error: 
`Unknown field 'type' on type 'MdxFrontmatter'.`

Can I get around this? Because I want  to make this field **optional**  in the frontmatter of each file.

freiksenet:
Hi!

Take a look at [schema customization docs](https://www.gatsbyjs.org/docs/schema-customization/#explicitly-defining-data-types).

For your use case, you would need something like the following:

```
createTypes(`
  type MdxFrontmatter @infer {
    type: String
  }

  type Mdx implements Node @infer {
    frontmatter: MdxFrontmatter
  }
`)
```

wangyi7099:
Hi I followed the docs and use the `createSchemaCustomization`  api in `gatsby-node.js`  which doesn't work. And gatsby show the errors:
`The following exports aren't APIs. Perhaps you made a typo or your plugin is outdated?`

This is my code, and I use it in a custom theme plugin:  
```js
exports.createSchemaCustomization = ({ actions }) => {
  const { createTypes } = actions;
  const typeDefs = `
  type MdxFrontmatter @infer {
    type: String
  }

  type Mdx implements Node @infer {
    frontmatter: MdxFrontmatter
  }
  `;
  createTypes(typeDefs);
};
```

stefanprobst:
@wangyi7099 The `createSchemaCustomization` API is only available in Gatsby v2.12 and above. If you are using a different version, you can use `sourceNodes` instead of `createSchemaCustomization`.

wangyi7099:
Thanks!
@freiksenet @stefanprobst I have updated to the lastest version and works properly!

wangyi7099:
Hi. I have another problem just now..  
When I change the plugin from `gatsby-mdx` to `gatsby-plugin-mdx`..  The code below still don't work.
```js
exports.createSchemaCustomization = ({ actions }) => {
  const { createTypes } = actions;
  const typeDefs = `
[...]

PR #16620, opened 2019-08-14
docs(themes): Make naming convention required and not recommended

sant0shg:
As discussed during the gatsby core meeting, prefixing the package name of gatsby themes with "gatsby-theme" is required, (and not only a recommendation)

PR #16928, opened 2019-08-21
feat: Allow themes to modify GraphQL types

### LINKED ISSUE #16529: Clarify GraphQL type ownership and permissions for themes
When building the GraphQL schema, we keep track of which types are created by which plugins. This information is used to prevent plugins from modifying types they do not own.

The exception to this is `default-site-plugin`, i.e. a user's `gatsby-node.js`, which is always allowed to modify types. When a type has been modified by a user, it is then owned by the user.

We should clarify, how this should work with themes. Currently, to the schema builder a theme appears like any other plugin and is therefore not allowed to modify types owned by other plugins. For example, a theme's `gatsby-node.js` would not be allowed to modify `MdxFrontmatter`, which is owned by `gatbsy-plugin-mdx`.

Options:
* treat a theme like any other plugin, which means modifying types from a theme's `gatsby-node` is not allowed
* treat a theme's `gatsby-node` like a top-level `default-site-plugin`
* define a special set of permissions for a theme's `gatsby-node`

the last two options would both require a way to differentiate a theme from other plugins -- by `gatsby-theme-*` naming convention or something else.

personally, I opt for option 2+naming convention

/cc @freiksenet @ChristopherBiscardi 

Related  #15544

stefanprobst:
Currently, a GraphQL type can only be modified by the plugin which created that type, or by a user's `gatsby-node`. This excludes a theme's `gatsby-node`.

This PR allows a theme's `gatsby-node` to modify types. A top-level `gatsby-node` always takes precedence. If a type has already been modified by another theme, modification is not allowed.

Fixes #16529, #15544

### DIFF
--- a/packages/gatsby/src/schema/schema.js
+++ b/packages/gatsby/src/schema/schema.js
@@ -299,11 +299,15 @@ const mergeTypes = ({
   parentSpan,
 }) => {
   // Only allow user or plugin owning the type to extend already existing type.
+  // Themes are also allowed to modify plugin-owned types, but not types already
+  // modified by another plugin.
   const typeOwner = typeComposer.getExtension(`plugin`)
   if (
     !plugin ||
+    plugin.name === typeOwner ||
     plugin.name === `default-site-plugin` ||
-    plugin.name === typeOwner
+    (plugin.name.startsWith(`gatsby-theme-`) &&
+      !typeOwner.startsWith(`gatsby-theme-`))
   ) {
     typeComposer.merge(type)
     if (isNamedTypeComposer(type)) {
--- a/packages/gatsby/src/schema/__tests__/build-schema.js
+++ b/packages/gatsby/src/schema/__tests__/build-schema.js
@@ -804,6 +804,165 @@ describe(`Build schema`, () => {
       )
     })
 
+    it(`does not merge plugin-defined type with type defined by theme`, async () => {
+      createTypes(
+        `type PluginDefined implements Node { foo: Int, baz: PluginDefinedNested }
+         type PluginDefinedNested { foo: Int }`,
+        {
+          name: `gatsby-theme-random`,
+        }
+      )
+      createTypes(
+        `type PluginDefined implements Node { bar: Int, qux: PluginDefinedNested }
+         type PluginDefinedNested { bar: Int }`,
+        {
+          name: `some-other-gatsby-plugin`,
+        }
+      )
+      const schema = await buildSchema()
+      const nestedFields = schema.getType(`PluginDefinedNested`).getFields()
+      const fields = schema.getType(`PluginDefined`).getFields()
+      expect(Object.keys(nestedFields)).toEqual([`foo`])
+      expect(Object.keys(fields)).toEqual([
+        `foo`,
+        `baz`,
+        `id`,
+        `parent`,
+        `children`,
+        `internal`,
+      ])
+      expect(report.warn).toHaveBeenCalledWith(
+        `Plugin \`some-other-gatsby-plugin\` tried to define the GraphQL type ` +
+          `\`PluginDefinedNested\`, which has already been defined by the plugin ` +
+          `\`gatsby-theme-random\`.`
+      )
+      expect(report.warn).toHaveBeenCalledWith(
+        `Plugin \`some-other-gatsby-plugin\` tried to define the GraphQL type ` +
+          `\`PluginDefined\`, which has already been defined by the plugin ` +
+          `\`gatsby-theme-random\`.`
+      )
+    })
+
+    it(`merges theme-defined type with plugin-defined type`, async () => {
+      createTypes(
+        `type PluginDefined implements Node @infer { foo: Int, baz: PluginDefinedNested }
+         type PluginDefinedNested { foo: Int }`,
+        {
+          name: `some-gatsby-plugin`,
+        }
+      )
+      createTypes(
+        `type PluginDefined implements Node @dontInfer { bar: Int, qux: PluginDefinedNested }
+         type PluginDefinedNested { bar: Int }`,
+        {
+          name: `gatsby-theme-random`,
+        }
+      )
+      const schema = await buildSchema()
+      const PluginDefinedNested = schema.getType(`PluginDefinedNested`)
+      const nestedFields = PluginDefinedNested.getFields()
+      const PluginDefined = schema.getType(`PluginDefined`)
+      const fields = PluginDefined.getFields()
+      expect(Object.keys(nestedFields)).toEqual([`foo`, `bar`])
+      expect(Object.keys(fields)).toEqual([
+        `foo`,
+        `baz`,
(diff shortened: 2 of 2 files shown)


stefanprobst:
apparently node 11 (or v8 really) has changed the sort implementation, which is why this doesn't work anymore:
https://github.com/gatsbyjs/gatsby/blob/c2fc3eef2a6c2231af7e46100a1bbe0898976d06/packages/gatsby/src/schema/index.js#L42-L44

packages/gatsby/src/schema/schema.js
```diff
@@ -299,11 +299,15 @@ const mergeTypes = ({
...
+  // Themes are also allowed to modify plugin-owned types, but not types already
+  // modified by another plugin.
   const typeOwner = typeComposer.getExtension(`plugin`)
   if (
     !plugin ||
+    plugin.name === typeOwner ||
     plugin.name === `default-site-plugin` ||
-    plugin.name === typeOwner
+    (plugin.name.startsWith(`gatsby-theme-`) &&
+      !typeOwner.startsWith(`gatsby-theme-`))
```
wangyi7099: Just wonder that does the name of a theme have to starts with "gatsby-theme" ?  If so, I  think that's a little unreasonable. My theme plugin's is `antdsite` , not starts with "gatsby-theme". I think we should judge theme in another way.

stefanprobst: the `gatsby-theme-` naming is a requirement since #16620, so we can rely on that. it's not totally optimal since i don't think it is actually enforced anywhere, otoh the only thing you'll lose (currently) when naming your theme differently is the ability to modify types owned by plugins. i'm open to other proposals, please post them in #16529.

wangyi7099: Lol All right, since you haved discussed.

packages/gatsby/src/schema/index.js
```diff
@@ -39,9 +39,14 @@ module.exports.build = async ({ parentSpan }) => {
   const typeConflictReporter = new TypeConflictReporter()
 
   // Ensure that user-defined types are processed last
-  const sortedTypes = types.sort(
-    type => type.plugin && type.plugin.name === `default-site-plugin`
-  )
+  const sortedTypes = [
```
wardpeet: This should fix sorting on all node versions. From node 12 (11.x) stable sort is being used and booleans aren't the best way to sort.

```
types.sort( 
   type => type.plugin && type.plugin.name === `default-site-plugin` ? 1 : -1
 ) 
```

michaellopez:
@wardpeet 
> I do have a question about the themes spec. Don't we want themes to override/use other themes and maybe also augment the graphql types of those child themes?

We have this use case, currently we work around it via `createNode` + `onCreateNode` plus `mediaType`. But would definitely want that in core if possible.

wardpeet:
Sweet! Good workaround 👍. My comment about parent/child augmentation isn't blocking to get this PR merged so most people will already be happy

stefanprobst:
> Don't we want themes to override/use other themes and maybe also augment the graphql types of those child themes?

#16529 did not really get lots of responses so i think this needs more discussion wrt to themes modifying types owned by other themes. this PR is a conservative start (following @KyleAMathews comment on slack)

packages/gatsby/src/schema/index.js
stefanprobst: unfortuntately this does not seem to work as expected

wardpeet: it seems to work for me 🤷‍♂ Oh looks like there was an issue when sorting wasn't stable.

```
types.sort( 
   (a, type) => type.plugin && type.plugin.name === `default-site-plugin` ? -1 : 1
 ) 
```
on runkit you can switch versions.
https://runkit.com/wardpeet/5d68ca5da5888a00132fe3c7

packages/gatsby/src/schema/schema.js
```diff
@@ -299,11 +299,15 @@ const mergeTypes = ({
...
   // Only allow user or plugin owning the type to extend already existing type.
+  // Themes are also allowed to modify plugin-owned types, but not types already
+  // modified by another plugin.
   const typeOwner = typeComposer.getExtension(`plugin`)
   if (
     !plugin ||
+    plugin.name === typeOwner ||
     plugin.name === `default-site-plugin` ||
-    plugin.name === typeOwner
+    (plugin.name.startsWith(`gatsby-theme-`) &&
```
ChristopherBiscardi: This is a heavy handed suggestion but not a requirement, as detailed in https://github.com/gatsbyjs/gatsby/issues/16529#issuecomment-526687220

ChristopherBiscardi: Also this code would need to be modified to work for scoped packages like `@namespace/gatsby-theme-thing` as well

packages/gatsby/src/schema/schema.js
stefanprobst: sorry, looks like my assumption was wrong

packages/gatsby/src/schema/index.js
stefanprobst: Thanks for the link @wardpeet!

So, the following would work, but is actually quite a bit slower than filtering:
```js
const sortedTypes = types.sort((a, b) => {
  if (a.plugin && a.plugin.name === `default-site-plugin`) return 1
  if (b.plugin && b.plugin.name === `default-site-plugin`) return -1
  return 0
})
```

wardpeet:
Could you add some tests for non vs scoped packages?

stefanprobst:
Sure, but let's first decide if we can actually rely on the naming convention?

freiksenet:
@stefanprobst I feel that we should probably remove restriction on modifying types by other plugins. I'm not sure there is any other solution, we don't want a "theme-class" and "non-theme-class" plugins.

CC @ChristopherBiscardi @KyleAMathews

KyleAMathews:
Perhaps we could do what we do for nodes? E.g. reserve a "fields" key on every type that other plugins can add additional fields to.

free-easy:
@KyleAMathews What I'd like to be possible is to override some types, instead of extending them.
E.g. override and make explicit, instead of inferred, the type of a frontmatter from markdown.

stefanprobst:
@freiksenet 

> I feel that we should probably remove restriction on modifying types by other plugins. I'm not sure there is any other solution, we don't want a "theme-class" and "non-theme-class" plugins.

I'm not really clear why that is? To me it seems that the issue of "who is allowed to modify which types" is exactly a case where having some way to differentiate between themes and other-plugins would make sense (even if only by naming convention). Not sure I like dropping type ownership just to preserve the theme/plugin symmetry.

KyleAMathews:
What we have to prevent is scenarios where s site is working and someone installs a plugin (or theme) and the site stops working. This is incredibly confusing for everyone.

Allowing plugins or themes to override types leads to exactly that sort of problem.

PR #16952, opened 2019-08-22
chore: Remove flow type annotations

### LINKED ISSUE #16472: Remove flow annotations
We don't really use them and there are very few sprinkled through the code base.

crutchcorn:
## Description

This PR removes flow type annotations in favor of `.d.ts` definition files and inline JSDoc type annotated comments. While the original issue pertains to only removing flow annotations and adding TS definition files, I felt that inline JSDoc type comments were an important add to preserve some of the developer niceties within these files (it's worth mentioning that we don't define the type shape in JSDoc, as was prosed as a question in the original issue, but rather simply leaves the type information that was not it's own type definition).

Something I was somewhat unclear with in the issue was how to track new definition files. I currently have them in `d.ts` files that match the original filename, but could move them to a global definition file for Gatsby as well if requested (I have no strong opinions either way, although moving them to a global definition file will potentially need namespacing/renaming [due to] for some of the type names)

## Related Issues

Fixes #16472


crutchcorn:
There seems to be a failing test, but it appears to be unrelated:

```
Summary of all failing tests
  ● gatsby-plugin-sharp › tracedSVG › runs on demand

: Timeout - Async callback was not invoked within the 5000ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 5000ms timeout specified by jest.setTimeout. 

    350 |     })                                                                                                                       
    351 |
  > 352 |     it(`runs on demand`, async () => {
        |     ^
    353 |       const args = {
    354 |         maxWidth: 100,
    355 |         width: 100,       

at new Spec (node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
at Suite.it (packages/gatsby-plugin-sharp/src/__tests__/index.js:352:5)
at Suite.describe (packages/gatsby-plugin-sharp/src/__tests__/index.js:329:3)
at Object.describe (packages/gatsby-plugin-sharp/src/__tests__/index.js:40:1)
```

I am the most up-to-date with `master` as of a few hours ago

packages/gatsby-cli/src/init-starter.d.ts
```diff
@@ -0,0 +1,3 @@
+export interface InitOptions {
+  rootPath?: string
```
eknowles: this would make this file empty right? is it worth keeping `InitOptions` in there anyway as it's TS?

packages/gatsby-cli/src/reporter/errors.js
```diff
@@ -52,9 +50,10 @@ function getErrorFormatter() {
 /**
  * Convert a stringified webpack compilation error back into
  * an Error instance so it can be formatted properly
- * @param {string} errorStr
+ * @param {string} [errorStr]
```
eknowles:
```suggestion
 * @param {string} [errorStr='']
```
if you're adding jsdoc, I think you would declare this default value like this?

packages/gatsby-cli/src/reporter/types.d.ts
```diff
@@ -0,0 +1,29 @@
+export interface ActivityTracker {
+  start(): Function
```
eknowles: I think all these need terminators at the end `;` satisfies a common TSLint and prettier default config

packages/gatsby-page-utils/src/create-path.js
```diff
@@ -1,7 +1,6 @@
-// @flow
 const path = require(`path`)
 
-module.exports = (filePath: string): string => {
+module.exports = filePath => {
```
eknowles:
```suggestion
module.exports = (filePath) => {
```
is this enforced anywhere?

packages/gatsby/src/bootstrap/index.d.ts
```diff
@@ -0,0 +1,5 @@
+type BootstrapArgs = {
```
eknowles: I think this should be an `interface` instead of a `type` in this case?

packages/gatsby/src/db/nodes.d.ts
```diff
@@ -0,0 +1,11 @@
+interface NodeStore {
+  getNodes: () => Array<any>
+  getNode: (id: string) => any | undefined
+  getNodesByType: (type: string) => Array<any>
+  getTypes: () => Array<string>
```
eknowles:
```suggestion
  getTypes: () => string[];
```

packages/gatsby/src/db/nodes.d.ts
```diff
@@ -0,0 +1,11 @@
+interface NodeStore {
+  getNodes: () => Array<any>
```
eknowles:
```suggestion
  getNodes: () => any[];
```

packages/gatsby/src/query/graphql-errors.d.ts
```diff
@@ -0,0 +1 @@
+type RelayGraphQLError = Error & { validationErrors?: Object }
```
eknowles:
```suggestion
interface RelayGraphQLError extends Error {
  validationErrors?: Object;
}
```
should this be an interface here as it extends an object type?

packages/gatsby/src/query/graphql-errors.js
```diff
@@ -120,11 +134,12 @@ function getCodeFrameFromRelayError(
...
 }
 
-export function multipleRootQueriesError(
-  filePath: string,
-  def: any,
-  otherDef: any
-) {
+/**
+ * @param {string} filePath
+ * @param {any} def
```
eknowles:
```suggestion
 * @param {*} def
```
I think `any` isn't a valid JSDoc tag type? https://jsdoc.app/tags-type.html - I might be wrong here, otherwise ignore me

packages/gatsby/src/query/query-compiler.js
```diff
@@ -75,18 +61,34 @@ const resolveThemes = (themes = []) =>
...
+  /**
+   * @type {string}
+   */
+  fragmentsDir
+
+  /**
+   * @param {string} base
+   * @param {string[]} additional
+   * @param {GraphQLSchema} schema
+   * @param $3 {}
```
eknowles:
```suggestion
   * @param $3.parentSpan
```

eknowles:
looks good, personally I would lean to exposing most of the `types` as `interfaces` as plugin authors, consumers might want to use these and find it tricky

packages/gatsby-cli/src/reporter/errors.js
crutchcorn: Oh wow, I didn't know JSDoc supported default values!

packages/gatsby-cli/src/init-starter.d.ts
crutchcorn: And this is a broader question that I wasn't sure how to address. I simply took the definition files from flow annotations and placed them into `.d.ts` files with the same name. If we're wanting to only keep the definitions, I'm not sure this is the right way to do them (as, like you'd mentioned, the TS language extensions MIGHT (?) assume that these files are empty), but I'm not sure how else we'd want to handle these files. 

That said, I've not noticed any differences with IDE behavior (including Ctrl+Click) in either Webstorm or VS Code so I'm unsure if this is the behavior which occurs.

packages/gatsby-cli/src/reporter/types.d.ts
crutchcorn: They were removed after running the current `prettierlint --fix` and was throwing linting errors when they were present. I'm unsure how to follow (as I saw errors from webstorm mentioning this as well)

packages/gatsby-page-utils/src/create-path.js
crutchcorn: This doesn't appear to be enforced using linting rules anywhere. This PR at this stage passes linting rules. That said, I'll make this change all-the-same

packages/gatsby/src/bootstrap/index.d.ts
crutchcorn: I generally agree. I'll make these changes

packages/gatsby/src/db/nodes.d.ts
crutchcorn: I think it's a good idea to expand the TS linting rules. I'll see about setting this up

packages/gatsby/src/query/graphql-errors.js
crutchcorn: Ahh, you are indeed correct. Apologies

packages/gatsby-cli/src/init-starter.d.ts
eknowles: so flow types and typescript are not the same, they have slightly different syntax

crutchcorn: Well and there are various implementation differences. I've written a bit of both. A `.d.ts` file is just a definition file, though. It's not a typescript logic file. All the `.d.ts`s are doing in this PR is enabling the JSDoc comments to be able to pick up the formed type for IDE TS Language extensions (built into TS itself, provided by MS for other IDEs even outside of vs code)

That said, I could continue writing the definition files to include the actual shapes of the file's exports if that's what we want, but I feel they're out-of-scope for this issue itself

wardpeet:
Hey, sorry for my late answer and thank you for your patience. I was waiting to respond as I wanted to get https://github.com/gatsbyjs/gatsby/pull/16091 in first because of the conflicts.

Thank you for removing our flow code as we're not using it anyway. From the Gatsby side we haven't decided on our typescript strategy yet. We're not sure if we should move to typescript (ts files) or go for jsdoc. We don't want to introduce `tsc` yet to do our validation.

If possible at this stage, I would like to keep most types/interfaces in jsdoc where the flow types were. It makes it easier to review and decide on what we're going to do next. If you need a d.ts file, I would put them inside the types folder in packages/gatsby and not next to the current file.

Again, thank you very much for all your code changes, which are a lot. I'm happy to answer any questions or pair to get this over the finish line.

LekoArts:
Hey @crutchcorn 	👋 	
Do you have time to resolve this conflicts? If not @brolag would be interested in doing that.

crutchcorn:
@LekoArts I can take care of the conflicts. I had some other questions that I got sidetracked away from, but I'll have the conflicts handled on this one today

crutchcorn:
I am 100% certain that I broke something during my conflict handling. I am going to go through my diff from master to fix the problems

crutchcorn:
Merge conflicts have been handled. There's still some question what should be done with the .d.ts files vs inline JSDoc files. I've DM'd @wardpeet about this on Discord as I just think there needs to be a quick back-n-forth so I can understand a bit better :)

blainekasten:
@crutchcorn we're currently exploring how we're going to start using TS in the gatsby project. I would be happy to merge this PR if we could reduce the scope to only remove flow (js doc stuff is fine, though it might be removed later).

I'd like to not overload this PR, it's already large! 😆

crutchcorn:
What should be done with the types that are present in the `.d.ts`? Should it just be plainly removed?

blainekasten:
@crutchcorn you can leave them unchanged. We'll be following up and fixing a lot of the typescript stuff. So if the scope of this PR could just be removing flow that would be super helpful!

blainekasten:
@crutchcorn are you still interested in finishing up the work to remove flow?

crutchcorn:
While I'm still interested in contributing to Gatsby (i'm actually going to be diving into a bug in the next week or so), I'm not wanting to move forward with removing flow (sorry). Too many required rebases due to the scope and some confusion between different folks have prevented me from taking another pass at this (primarily the first. Each rebase takes a decent chunk of time for me)

blainekasten:
@crutchcorn Absolutely no problem. So sorry for sending you in loops. This triggered philosophical discussions within the team and required us to make larger decisions that you got caught in the middle of a bit. So happy to hear you are still excited to be involved as your code is great and your contributions are wonderful. 💜

crutchcorn:
No worries in the slightest! I figured that's what'd happened and I know how extremely difficult it can be to sync the public with internal dialog (no sarcasm here - I know communication and organization are some of the trickiest bits of managing something like this! 💜 ). Glad to be pinged and updated - looking forward to more :)

PR #16961, opened 2019-08-22
Add grayMatter option to gatsby-plugin-mdx

### LINKED ISSUE #16865: Richer frontmatter support in gatsby-plugin-mdx
## Summary

We use gray-matter for gatsby-plugin-mdx frontmatter support. We should passthrough the `excerpt_separator` option.

### Basic example

We would likely just enable passthrough in the options as such:

```js
module.exports = {
  plugins: [
    {
      resolve: `gatsby-plugin-mdx`,
      options: {
        grayMatter: {
          excerpt_separator: "<<<my-separator>>>"
        }
      }
    }
  ]
};
```

Would set this mdx content:

```md
---
title: Blog
---
My awesome blog.
<<<my-separator>>>
<h1>Hello world</h1>
```

to this result:

```js
{ 
  data: { title: 'Blog'}, 
  excerpt: 'My awesome blog.', 
  content: 'My awesome blog.\n<<<my-separator>>>\n<h1>Hello world</h1>' 
}
```

### Open Questions

* Does the excerpt stay in the content afterwards?
* gray-matter supports a lot of options, some of which are literal functions. Is there a better way to support those since gatsby-config is serialized?
* How does the excerpt interact with the gatsby-plugin-mdx excerpt resolver. Currently if there is an excerpt we'd prefer it over the resolver.
* What happens to react components in excerpts? currently we don't render them.
* We should probably test what already works (like frontmatter language inference) before attempting to add this functionality.

### Motivation

When porting from other ecosystems it can be easier if we support their frontmatter delimiters and excerpt formats. That was the use case in [the original gatsby-mdx issues](https://github.com/ChristopherBiscardi/gatsby-mdx/issues/386)

ported from https://github.com/ChristopherBiscardi/gatsby-mdx/issues/386

riywo:
## Description
This commit is adding a new option for `gray-matter` used inside `gatsby-plugin-mdx`. The main motivation of this PR is for adding `excerpt_separator` which is commonly used in blogs.

When you specify:
```
module.exports = {
  plugins: [
    {
      resolve: `gatsby-plugin-mdx`,
      options: {
        grayMatter: {
          excerpt_separator: "<<<my-separator>>>"
        }
      }
    }
  ]
};
```

`gray-matter` automatically pick content before the `excerpt_separator`. ~~However, it contains mdx resources, so we need to translate it. In the `excerpt` resolver, it runs `processMDX()` for the extracted excerpt string just like it does for without `excerpt_separator`, then concat them.~~ Any MDX syntax won't be transformed since `gray-matter` doesn't know anything about MDX.

## Related Issues
Fixes #16865 

Note: Previously I tried to implement `excerpt_separator` inside the resolver but it seemed too complex. #16820

### DIFF
--- a/packages/gatsby-plugin-mdx/gatsby/on-create-node.js
+++ b/packages/gatsby-plugin-mdx/gatsby/on-create-node.js
@@ -49,6 +49,7 @@ module.exports = async (
     id: createNodeId(`${node.id} >>> Mdx`),
     node,
     content,
+    options,
   })
 
   createNode(mdxNode)
--- a/packages/gatsby-plugin-mdx/utils/gen-mdx.js
+++ b/packages/gatsby-plugin-mdx/utils/gen-mdx.js
@@ -1,5 +1,5 @@
 const babel = require(`@babel/core`)
-const grayMatter = require(`gray-matter`)
+const grayMatterWrapper = require(`./gray-matter-wrapper`)
 const mdx = require(`@mdx-js/mdx`)
 const objRestSpread = require(`@babel/plugin-proposal-object-rest-spread`)
 
@@ -76,7 +76,10 @@ module.exports = async function genMDX(
 
   // pull classic style frontmatter off the raw MDX body
   debug(`processing classic frontmatter`)
-  const { data, content: frontMatterCodeResult } = grayMatter(node.rawBody)
+  const { data, content: frontMatterCodeResult } = grayMatterWrapper(
+    node.rawBody,
+    options
+  )
   const content = `${frontMatterCodeResult}
 
 export const _frontmatter = ${JSON.stringify(data)}`
--- a/packages/gatsby-plugin-mdx/utils/gray-matter-wrapper/index.js
+++ b/packages/gatsby-plugin-mdx/utils/gray-matter-wrapper/index.js
@@ -1,0 +1,8 @@
+const grayMatter = require(`gray-matter`)
+
+module.exports = (input, options) => {
+  const matter = grayMatter(input, options.grayMatter || {})
+  const { data: origData, excerpt, content } = matter
+  const data = excerpt === `` ? origData : { excerpt, ...origData }
+  return { data, content }
+}
--- a/packages/gatsby-plugin-mdx/utils/mdx.js
+++ b/packages/gatsby-plugin-mdx/utils/mdx.js
@@ -1,5 +1,5 @@
 const mdx = require(`@mdx-js/mdx`)
-const grayMatter = require(`gray-matter`)
+const grayMatterWrapper = require(`./gray-matter-wrapper`)
 /**
  * Converts MDX to JSX, including converting classic frontmatter to an
  * exported variable.
@@ -9,7 +9,7 @@ const grayMatter = require(`gray-matter`)
  * @return {String}         JSX source
  */
 module.exports = async function mdxToJsx(source, options) {
-  const { data, content } = grayMatter(source)
+  const { data, content } = grayMatterWrapper(source, options || {})
 
   let code = await mdx(content, options || {})
 
--- a/packages/gatsby-plugin-mdx/loaders/mdx-loader.js
+++ b/packages/gatsby-plugin-mdx/loaders/mdx-loader.js
@@ -119,6 +119,7 @@ module.exports = async function(content) {
       id: `fakeNodeIdMDXFileABugIfYouSeeThis`,
       node: fileNode,
       content,
+      options,
     })
   } catch (e) {
     return callback(e)
--- a/packages/gatsby-plugin-mdx/utils/create-mdx-node.js
+++ b/packages/gatsby-plugin-mdx/utils/create-mdx-node.js
@@ -3,10 +3,10 @@ const { createContentDigest } = require(`gatsby-core-utils`)
 const mdx = require(`../utils/mdx`)
 const extractExports = require(`../utils/extract-exports`)
 
-module.exports = async ({ id, node, content }) => {
+module.exports = async ({ id, node, content, options }) => {
   let code
   try {
-    code = await mdx(content)
+    code = await mdx(content, options)
   } catch (e) {
     // add the path of the file to simplify debugging error messages
     e.message += `${node.absolutePath}: ${e.message}`
--- a/packages/gatsby-plugin-mdx/utils/default-options.js
+++ b/packages/gatsby-plugin-mdx/utils/default-options.js
@@ -19,6 +19,7 @@ module.exports = ({ mdPlugins, hastPlugins, ...pluginOptions }) => {
       gatsbyRemarkPlugins: [],
       globalScope: `export default {}`,
       shouldBlockNodeFromTransformation: () => false,
+      grayMatter: {},
     },
     pluginOptions
   )
--- a/packages/gatsby-plugin-mdx/utils/gray-matter-wrapper/__snapshots__/index.test.js.snap
+++ b/packages/gatsby-plugin-mdx/utils/gray-matter-wrapper/__snapshots__/index.test.js.snap
@@ -1,0 +1,130 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`gray-matter excerpt snapshot with body 1`] = `
+Object {
+  "content": "
+This is an excerpt.
+This is also an excerpt.
+
+This could be an excerpt.
+---
+This is content.",
+  "data": Object {},
+}
+`;
+
+exports[`gray-matter excerpt snapshot with excerpt 1`] = `
+Object {
+  "content": "
+This is an excerpt.
+This is also an excerpt.
+
+This could be an excerpt.
+---
+This is content.",
+  "data": Object {
+    "excerpt": "
+This is an excerpt.
+This is also an excerpt.
+
+This could be an excerpt.
+",
+  },
+}
+`;
+
+exports[`gray-matter excerpt snapshot with excerpt_separator 1`] = `
+Object {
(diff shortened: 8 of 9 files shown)


packages/gatsby-plugin-mdx/gatsby/source-nodes.test.js
```diff
@@ -0,0 +1,178 @@
+const { graphql } = require(`gatsby/graphql`)
```
ChristopherBiscardi: it seems like this test requires mocking a lot of code, making it prone to breakage and extra maintenance burden in the future. Have any ideas to simplify this?

packages/gatsby-plugin-mdx/gatsby/source-nodes.js
```diff
@@ -131,19 +146,17 @@ module.exports = (
         },
         async resolve(mdxNode, { pruneLength }) {
           if (mdxNode.excerpt) {
-            return Promise.resolve(mdxNode.excerpt)
+            const mdxExcerptNode = await createMDXNode({
```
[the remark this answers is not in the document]
riywo: `gray-matter` can't interpret any `import` or `export` statement. So, if we use the string excerpt, it could contain unnecessary text.

I'm ok with it for my use case, though. I'll revert this.

packages/gatsby-plugin-mdx/gatsby/on-create-node.js
```diff
@@ -53,11 +53,14 @@ module.exports = async (
     id: createNodeId(`${node.id} >>> Mdx`),
     node,
     content,
+    options,
   })
 
   createNode(mdxNode)
   createParentChildLink({ parent: node, child: mdxNode })
 
+  if (!cache) return
```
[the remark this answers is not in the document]
riywo: Yeah, this was added to pass the unit test. I'll revert and try to modify the test itself.

packages/gatsby-plugin-mdx/utils/gen-mdx.js
```diff
@@ -76,10 +76,16 @@ module.exports = async function genMDX(
 
   // pull classic style frontmatter off the raw MDX body
   debug(`processing classic frontmatter`)
-  const { data, content: frontMatterCodeResult } = grayMatter(node.rawBody)
+  const grayMatterOptions =
+    typeof options === `undefined` ? {} : options.grayMatter || {}
```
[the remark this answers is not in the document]
riywo: I see. Will do that.

packages/gatsby-plugin-mdx/utils/gen-mdx.js
```diff
@@ -76,10 +76,16 @@ module.exports = async function genMDX(
...
+  const { data, excerpt, content: frontMatterCodeResult } = grayMatter(
+    node.rawBody,
+    grayMatterOptions
+  )
+  const frontmatter = excerpt === `` ? data : { ...data, excerpt }
```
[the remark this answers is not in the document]
riywo: Good point. I'll fix it.

packages/gatsby-plugin-mdx/gatsby/source-nodes.test.js
riywo: This test was mostly imported from `gatsby-transfomer-remark`. I'll try to find the way to test GraphQL query easier...

riywo:
> How do you see this PR answering the open questions in #16865?

- Does the excerpt stay in the content afterwards?
    - I believe so. I'll try to add such test.
- gray-matter supports a lot of options, some of which are literal functions. Is there a better way to support those since gatsby-config is serialized?
    - I have no idea so far.
- How does the excerpt interact with the gatsby-plugin-mdx excerpt resolver. Currently if there is an excerpt we'd prefer it over the resolver.
    - As I commented, it can work by just leaving an excerpt found by gray-matter but it could contain unnecessary `export` or `import` statements.
- What happens to react components in excerpts? currently we don't render them.
    - Same above. It returns as is. We could strip `export` and `import` using regex.
- We should probably test what already works (like frontmatter language inference) before attempting to add this functionality.
    - Added a test but it seems too complex. Maybe we need finer-grained tests, not E2E for GraphQL.

riywo:
@ChristopherBiscardi I've updated the PR.

- Introduced `utils/gray-matter` to wrap the normal `gray-matter` and make testing easier.
- Add snapshot tests for `utils/gray-matter`.
    - So far, they are only focusing on `excerpt` related options. We can add more tests for other options later.
- Note: Any MDX syntax won't be transformed e.g. bare `import`, `export` or `JSX` including shortcodes.

riywo:
BTW, I think rendering Markdown (not MDX) is at least required like `gatsby-transformer-remark`'s `PLAIN` format...

packages/gatsby-plugin-mdx/utils/gray-matter/index.js
```diff
@@ -0,0 +1,11 @@
+const grayMatter = require(`gray-matter`)
+
+const stripExcerpt = (excerpt, _options) => excerpt.trim().replace(/\n+/g, ` `)
```
muescha: question: is stripping new lines per default right here or should it return 1:1 the excerpt and its up to the user to change it as needed?

how is it done in [packages/gatsby-transformer-remark](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-transformer-remark) since there is also an excerpt option there ?

dabrady: The `gatsby-transformer-remark` plugin [apparently uses `gray-matter`](https://using-remark.gatsbyjs.org/excerpts/) under the hood to retrieve the excerpt.

But if I'm understanding [the actual implementation](https://github.com/gatsbyjs/gatsby/blob/69ecf6a534b2be20e3b1ffe0b694ea677b68bb66/packages/gatsby-transformer-remark/src/extend-node-type.js#L436-L439) correctly, the `gray-matter` excerpt is only used in one particular case: when the `format:MARKDOWN` option is specified in the GraphQL query _and_ the `excerpt_separator` option is configured; in all other cases, the excerpt parsing is hand-rolled.

A quick test on my own site shows the behavior of the different excerpt formatting modes when an `excerpt_separator` is specified and the excerpt is rendered in a `pre` tag (to preserve whitespace and prevent text-wrapping).

| `PLAIN` | `MARKDOWN` | `HTML` |
:--------:|:----------:|:-------:
Screen Shot 2020-01-05 at 21 21 06 | Screen Shot 2020-01-05 at 21 21 16 | Screen Shot 2020-01-05 at 21 21 24

The results are interesting. It seems `gatsby-transformer-remark` plugin is trying to preserve newlines (though in the case of `PLAIN` I think there is a slight bug: entirely blank lines cause the preceding and following line to be joined).

So if we are trying to mirror the behavior of the `Remark` plugin, I think we should preserve the output of `gray-matter` exactly, and let the user decide if they want to manipulate it further.

riywo: Updated.

riywo:
With this change, it behaves apparently differently between with and without `excerpt_separator`: https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-mdx/gatsby/source-nodes.js#L133-L146

- If `grayMatter.excerpt_separato` is specified: The excerpt is not processed by MDX
- If not: The excerpt is processed by MDX

packages/gatsby-plugin-mdx/utils/mdx.js
```diff
@@ -1,5 +1,5 @@
 const mdx = require(`@mdx-js/mdx`)
-const grayMatter = require(`gray-matter`)
+const grayMatter = require(`./gray-matter`)
```
muescha: I would name it different, otherwise other devs are expecting the original GrayMatter functionality here. 

one possible naming
```suggestion
const grayMatterWrapper = require(`./gray-matter-wrapper`)
```

riywo: Thank you for the suggestion. I've updated.

packages/gatsby-plugin-mdx/utils/gray-matter-wrapper/index.js
```diff
@@ -0,0 +1,8 @@
+const grayMatter = require(`gray-matter`)
+
+module.exports = (input, options) => {
+  const matter = grayMatter(input, options.grayMatter || {})
+  const { data: origData, excerpt, content } = matter
+  const data = excerpt === `` ? origData : { excerpt, ...origData }
```
pvdz, line 6: Can `excerpt` not be `null` or `undefined` here in case of absence? Just wondering whether a `!excerpt` wouldn't be better here, if it can also be the empty string.

mathieudutour:
on `gatsby-transformer-remark`, the `excerpt_separator` option is at the root: https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-transformer-remark#example-excerpts

Shall it be the same here to be consistent?

riywo:
> on `gatsby-transformer-remark`, the `excerpt_separator` option is at the root: https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-transformer-remark#example-excerpts
> 
> Shall it be the same here to be consistent?

That's my first attempt but rejected. https://github.com/gatsbyjs/gatsby/pull/16820

Anyway, even if this is merged, the graymatter provided excerpt is useless since it is a plain text and the customer needs to process it by MDX (I don't know how I can do that). So, in my use case, I gave up this approach and decided to specify excerpt in frontmatter explicitly (used unified to scrape the excerpt content in the existing MD files.)

Thus, I don't have any motivation to push this change. If it has any use case, please reopen this PR.

PR #16962, opened 2019-08-22
chore(gatsby-image): Clarify IntersectionObserver support in README.md

bjfresh:
The latest Safari releases for Desktop and iOS now include support for IntersectionObserver:
https://caniuse.com/#feat=intersectionobserver

## Description

Documentation edits gatsby-image

### DIFF
--- a/packages/gatsby-image/README.md
+++ b/packages/gatsby-image/README.md
@@ -435,4 +435,4 @@ While you could achieve a similar effect with plain CSS media queries, `gatsby-i
   to use a gif with `gatsby-image`, it won't work. For now, the best workaround is
   to [import the gif directly](/docs/adding-images-fonts-files).
 - Lazy loading behavior is dependent on `IntersectionObserver` which is not available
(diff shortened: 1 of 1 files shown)


lannonbr:
I'm curious if it is needed to comment for people using older versions of safari given the quick turnaround of people upgrading to new versions of MacOS & iOS. For instance here on [Apple's App Store support site](https://developer.apple.com/support/app-store/), about 88% of people on iOS are running at least iOS 12 now. So if there's less and less people on these older versions of iOS / MacOS, should we need to say this for older versions or just comment that if you are supporting IE that you need polyfills

bjfresh:
I weighed both options, figuring that a March 2019 release was recent enough to warrant the specificity, especially given that there's just 88% support for IntersectionObserver globally. Of course, that'll fade quickly. According to the stats, 3% are still using incompatible versions of Safari. Realistically it's probably enough to simply encourage people to check browser support, and highlight IE as an outlier.

wardpeet:
I agree with @lannonbr  if we can just drop Safari that would be great :)

bjfresh:
OK, done. Safari is out.

PR #16973, opened 2019-08-22
chore(ci): Fix lerna concurrency

sidharthachatterjee:
The new dependency updates seem to cause https://github.com/lerna/lerna/issues/1798 and `bootstrap` fails as a result on CircleCI. 

This is happening because CircleCI incorrectly reports the number of CPUs and lerna tries to run too many at once and runs out of memory. 

The fix here is to set concurrency to 2 for `lerna prepare` (only when running in CI)

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     }
   },
   "scripts": {
-    "bootstrap": "npm-run-all -s check-versions lerna-prepare",
+    "bootstrap": "npm-run-all -s check-versions \"lerna-prepare -- --{@}\" --",
     "check-versions": "babel-node scripts/check-versions.js",
     "check-repo-fields": "babel-node scripts/check-repo-fields.js",
     "format": "npm run format:code && npm run format:other && npm run format:svg",
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,7 +103,7 @@ jobs:
       - <<: *restore_cache
       - <<: *install_node_modules
       - <<: *persist_cache
-      - run: yarn bootstrap
+      - run: yarn bootstrap -- concurrency=2
       - persist_to_workspace:
           root: packages
           paths:


sidharthachatterjee:
@KyleAMathews Yeah, really frustrating that they misreport `os.cpus()`

wardpeet:
Please think of windows users :) This change won't allow me to bootstrap. 

`\"lerna-prepare -- --{@}\" --"` this voodoo is breaking windows.
`npm-run-all -s  "lerna-prepare -- --{@}" -- --concurrency=2`

sidharthachatterjee:
Oops! Sorry, totally didn't think of that @wardpeet 😱

package.json
```diff
@@ -72,7 +72,7 @@
     }
   },
   "scripts": {
-    "bootstrap": "npm-run-all -s check-versions lerna-prepare",
+    "bootstrap": "npm-run-all -s check-versions \"lerna-prepare -- --{@}\" --",
```
wardpeet: We use yarn on CI so we shouldn't need lerna-prepare with -- as yarn works without -- to pass arguments down.
```suggestion
    "bootstrap": "npm-run-all -s check-versions lerna-prepare",
```

sidharthachatterjee: Already tried this. It doesn't work because both `check-versions` and `lerna-prepare` are scripts in `package.json`

sidharthachatterjee:
@wardpeet Voodoo seemed to have worked fine on Windows ;)

https://circleci.com/gh/gatsbyjs/gatsby/203683?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

This runs `yarn bootstrap`

package.json
wardpeet: My bad, i had a dependency error, not a lerna-prepare issue. This works on windows too! :) great job! Sorry about being faulty on this one! <3

sidharthachatterjee:
This was caused by https://github.com/lerna/lerna/pull/1931 by the way which was in a new `lerna` version which we upgraded to recently in our mass upgrade!

PR #16979, opened 2019-08-23
fix(gatsby): update check for default exports

sidharthachatterjee:
```
    return _react["default"].createElement(_react["default"].Fragment, null);
  };

  return AppShell;
}(_react["default"].Component);

var _default = AppShell;
exports["default"] = _default;
```

The code above throws with 

```
error #11328 

A page component must export a React component for it to be valid. Please make sure this file exports a React component:

/root/project/e2e-tests/production-runtime/node_modules/gatsby-plugin-offline/app-shell.js
```

This happens because our check doesn't handle `exports["default"]`

(babel prior to https://github.com/gatsbyjs/gatsby/pull/16929 output `exports.default = _default;` but now outputs `exports["default"] = _default;`)

This PR adds that

### DIFF
--- a/packages/gatsby/src/redux/actions/public.js
+++ b/packages/gatsby/src/redux/actions/public.js
@@ -359,6 +359,7 @@ ${reservedFields.map(f => `  * "${f}"`).join(`\n`)}
       !fileContent.includes(`export default`) &&
       !fileContent.includes(`module.exports`) &&
       !fileContent.includes(`exports.default`) &&
+      !fileContent.includes(`exports["default"]`) &&
       // this check only applies to js and ts, not mdx
       /\.(jsx?|tsx?)/.test(path.extname(fileName))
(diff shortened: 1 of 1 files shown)


sidharthachatterjee:
@wardpeet Will be painful to write as an end to end because we'll need different versions of babel for the different outputs 😓 

But we can add fixtures for each compiled case and write some unit I think. Let me add that.

wardpeet:
I didn't really look at the whole file so it was more a go to responds :p So not worth at this point to actually add a test. I also think we have covered all possible options so 👍

sidharthachatterjee:
@wardpeet Yeah, I agree. Let's table this for now. Perhaps in the future, we can improve the heuristic that it uses to check whether there is a default export at all. For now, we're good! 

Thanks! 💜

PR #16992, opened 2019-08-23
chore: update eslint-utils to 1.4.2

nikoladev:
## Description

From [npm's security advisory](https://www.npmjs.com/advisories/1118):

> Versions of `eslint-utils` >=1.2.0 or <1.4.1 are vulnerable to Arbitrary Code Execution. The `getStaticValue` does not properly sanitize user input allowing attackers to supply malicious input that executes arbitrary code during the linting process. The `getStringIfConstant` and `getPropertyName` functions are not affected.

More info:

- https://github.com/mysticatea/eslint-utils/security/advisories/GHSA-3gx7-xhv7-5mx3
- https://eslint.org/blog/2019/08/eslint-v6.2.1-released

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -8637,8 +8637,11 @@ eslint-scope@^4.0.3:
     estraverse "^4.1.1"
 
 eslint-utils@^1.3.1:
-  version "1.3.1"
-  resolved "https://registry.yarnpkg.com/eslint-utils/-/eslint-utils-1.3.1.tgz#9a851ba89ee7c460346f97cf8939c7298827e512"
+  version "1.4.2"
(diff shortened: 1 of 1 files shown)


nikoladev:
Oops, I see that I've given the branch an incorrect name! This fix updated `eslint-utils` to `1.4.2`, to be clear.

wardpeet:
Hey, changing yarn.lock doesn't really fixes this issue. It will be fixed for yarn users that use this repository but it won't be used by people that install gatsby and create an application with it.

We don't have eslint-utils specified anywhere except our dependencies so we can't really solve this problem. We'll need to upgrade to eslint 6 which we cant as we suport node 8.0 and eslint 6 needs 8.10.0 or something

PR #16993, opened 2019-08-23
initial migration from cypress to jest-puppeteer

davidbailey00:
## Description

Experiment with migrating tests from Cypress to `jest-puppeteer`


davidbailey00:
different behaviour on CI vs locally with production build tests w/ offline plugin (remind you of something? :D) so I'll put this on hold for now to look at other issues

PR #16996, opened 2019-08-23
chore(gatsby-source-graphql): move graphql to peerdependencies

stefanprobst:
Make `gatsby-source-graphql` use gatsby's graphql package to avoid having two `graphql` packages in a project.

### DIFF
--- a/packages/gatsby-source-graphql/package.json
+++ b/packages/gatsby-source-graphql/package.json
@@ -10,7 +10,6 @@
     "@babel/runtime": "^7.5.5",
     "apollo-link": "1.2.12",
     "apollo-link-http": "^1.5.15",
-    "graphql": "^14.5.0",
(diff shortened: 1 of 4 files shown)


packages/gatsby-source-graphql/package.json
```diff
@@ -29,7 +28,8 @@
   ],
   "license": "MIT",
   "peerDependencies": {
-    "gatsby": "^2.0.15"
+    "gatsby": "^2.0.15",
+    "graphql": "^14.5.0"
```
wardpeet: why move graphql to peerdependencies if you're not using it? Is it to satisfy other dependencies?

stefanprobst: oh yes you're right, we can get rid of it -- i'll open a new one though because i accidentally force-pushed...

PR #17006, opened 2019-08-23
fix(gatsby-image): Fix fluid not respecting maxWidth/maxHeight

### LINKED ISSUE #15167: gatsby-image fluid not respecting maxHeight
## Description

When I use the `fluid` function and specify a `maxHeight` value, it doesn't stick to that size. However, when I use the `fixed` version of the following, everything renders just fine. I have tried `gatsby clean`, `gatsby build`, and `gatsby serve` as well as re-installing all my node_modules. I'm not sure if this issue is related to #14988 but I didn't want to comment on it if they are separate issues... but I apologize for duplicating if they are not.

`@t2ca#1075` has also been kind enough to help me on discord but we haven't had any luck in the troubleshooting we did. I've done a search through the issues as best I can but I didn't find much in the short time I spent.

### Steps to reproduce

```js
import React from "react";
import { graphql, StaticQuery } from "gatsby";
import Img from "gatsby-image";

export default function MyImage() {
  return (
    <StaticQuery
      query={graphql`
        query {
          logoImage: file(relativePath: { eq: "paddle.png" }) {
            childImageSharp {
              fluid(maxHeight: 35) {
                ...GatsbyImageSharpFluid
              }
            }
          }
        }
      `}
      render={data => <Img fluid={data.logoImage.childImageSharp.fluid} />}
    />
  );
}
```

### Expected result

I should see a fluid image with a maxHeight as specified.

### Actual result

An image that is too large is rendered. I see no errors in the console.

### Environment
```
  System:
    OS: macOS 10.14.5
    CPU: (4) x64 Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
    Shell: 5.7.1 - /usr/local/bin/zsh
  Binaries:
    Node: 12.4.0 - /usr/local/bin/node
    Yarn: 1.17.0 - /usr/local/bin/yarn
  Languages:
  Browsers:
    Chrome: 75.0.3770.100
    Safari: 12.1.1
  npmPackages:
    gatsby: ^2.10.5 => 2.10.5 
    gatsby-image: ^2.2.3 => 2.2.3 
    gatsby-plugin-create-client-paths: ^2.1.1 => 2.1.1 
    gatsby-plugin-manifest: ^2.2.0 => 2.2.0 
    gatsby-plugin-offline: ^2.2.0 => 2.2.0 
    gatsby-plugin-sharp: ^2.2.1 => 2.2.1 
    gatsby-source-filesystem: ^2.1.1 => 2.1.1 
    gatsby-transformer-sharp: ^2.2.0 => 2.2.0 
```

sbardian:
## Description
An attempt at fixing gatsby-image fluid not respecting maxHeight #15167 issue.  Instead of using the sizes string to pull out max-width/height, I decided to try to add to maxHeight/maxWidth fields to the node and use those in `gatsby-image` on the `gatsby-image-wrapper` div.  This might be way off base, and may presents other problems I'm not seeing at this point, but thought I would give it a try.  This change touches many different pieces of code, and I am in no way confident this is a great/easiest/best way to fix this.  :) My manual testing _seems_ to fix the issue though. 

## Related Issues
closes #15167


wardpeet:
I'm not 100% sure what to do here. We have presentationWidth & presentationHeight  which we can already use for this. Another thing is I'm unsure if baking this in gatsby-image is the right decision. 

Problems with this solution is that we need to update our fragments to make it universal to always include presentationWidth & presentationHeight  inside fluid fragments, which is probably not a big deal.

```js

const Image = () => {
  const data = useStaticQuery(graphql`
    query {
      placeholderImage: file(relativePath: { eq: "gatsby-astronaut.png" }) {
        childImageSharp {
          fluid(maxHeight: 150) {
            presentationWidth,
            presentationHeight,
            ...GatsbyImageSharpFluid
          }
        }
      }
    }
  `)

  return <Img fluid={data.placeholderImage.childImageSharp.fluid} style={{ width: data.placeholderImage.childImageSharp.fluid.presentationWidth, height: data.placeholderImage.childImageSharp.fluid.presentationHeight }} />
}
```

sbardian:
@wardpeet you are right, I didn't even notice those were getting the same values.  I can drop the 3 commits dealing with adding maxHeight/maxWidth (4ac71dd, a838a30, c34d5b4) and work from there using `presentationHeight` and `presentationWidth` instead.  Unless you think we should close this PR and come at this from a totally different angle?

sbardian:
@wardpeet, I have pushed the changes we talked about, for you to take a look.  I _think_ I updated the fragments correctly, but I had a heck of a time getting my test project to _see_ / _take_ those updates, so maybe there is somewhere else I'm not aware of that will need to be updated as well?

sbardian:
thanks @wardpeet 👍

pieh:
This changes default behaviour and users will get different results for same code (breaking change) - some people rely on current behaviour. I think it's also easier to to set `maxWidth`,`maxHeight` than unset it. Also the `maxWidth` and `maxHeight` argument don't apply to presentational dimensions - but rather to size of generated images. Changing that removes "fluid" characteristic of `fluid` image ;) (above certain threshold)

As a note - please check https://github.com/gatsbyjs/gatsby/pull/10460 for PR that was pretty similar to this one.

sbardian:
Thought I would give this one last go 😀 sorry @pieh 

This still might "break" people who are relying on the current behavior...

Thought here is adding maxWidth and maxHeight to the node, to use in `gatsby-sharp-plugin` and `gatsby-image`.  Also, adding a container around gatsby-image-wrapper to keep the `fluid` image behavior. This seems to work as one might expect maxHeight and maxWidth to work.  

I see now that you say it, `presentationHeight` and `presentationWidth` "max out" at the image dimensions, with this update they respect the maxHeight and maxWidth args passed in the query. 

Testing this seems to function pretty well, but I understand it could throw everyone using this for a loop.  We can just close it out and go with the documentation with the easy fix 👍 no big.

sbardian:
one thing I can't seem to get to work, is getting the update fragments in my project.  so when testing I have to include `maxHeight` and `maxWidth` in my query with `...GatsbyImageSharpFluid`.  Tried using `gatsby clean` / running build on `gatsby-transformer-sharp` then running `gatsby develop` /etc.  Can't seem to get the new fragments to populate in my dev projects cache.

sbardian:
I have created a gatsby test site for testing this PR.  https://github.com/sbardian/my-default-starter.  directions in README for anyone who needs them.

pieh:
Hey, don't worry about me ;)

I know this is confusing and problematic (hence attempt to address it with this PR), only thing I wanted to convey is that we have to be careful about breaking changes. 

I don't have proper answer on how to best solve this. Adding documentation is "safest way out" (in that we avoid making breaking changes, and that's why we opted into this way in that previous PR), but this will still bite people that don't expect our current behaviour and it's not unreasonable to think that meaningful portion of users might never find our documentation for this problem

sbardian:
Will wait and see what wardpeet or anyone else that might check it out thinks.  Give it a test drive and let me know.  Thanks!

sbardian:
@pieh @wardpeet I was looking through the documentation on the manual "fix" for this.  I don't think this will be a breaking change.  I think people using the documented fix would just end up overriding the new maxHeight and/or maxWidth values I'm setting on the image, when they pass their `normalizedProps.styles` to the `gatsby-image` component.   So, no breaking change for people using the documentation as far as I can see.

sbardian:
@pieh @wardpeet looking at the documented fix, it will not always actually work.  `presentationWidth` and `presentationHeight` will not always equal the `maxHeigth` and `maxWidth` args.  They max out at the image dimensions.  So using the gatsby-astronaut.png from the default starter as an example, if you had the following query: 

```
query MyQuery {
  allImageSharp(filter: {fluid: {src: {regex: "/astronaut/g"}}}) {
    nodes {
      fluid(maxWidth: 2000, maxHeight: 2000) {
        src
        maxWidth
        maxHeight
        presentationWidth
        presentationHeight
      }
    }
  }
}
```
you will get the following results (using my branch) or the same minus maxHeight/maxWidth using master branch: 
```
{
  "data": {
    "allImageSharp": {
      "nodes": [
        {
          "fluid": {
            "src": "/static/6d91c86c0fde632ba4cd01062fd9ccfa/eb6cb/gatsby-astronaut.png",
            "maxWidth": 2000,
            "maxHeight": 2000,
            "presentationWidth": 800,
            "presentationHeight": 800
          }
        }
      ]
    }
  }
}
```

So with the fix you would be setting the gatsby-image styles to maxHeight 800, not the 2000 you requested in your query.

sbardian:
@gillkyle have you had a chance to review? Thoughts? Anyone, else I should reach out to?

gillkyle:
Hey @sbardian, all your work on this looks awesome and your research into the problem is really helpful!

I probably don't have as much context as @pieh and @wardpeet (who can probably make the final say better), but I do like the idea of documenting what you've found tricky with this better in the [API doc page](https://www.gatsbyjs.org/docs/gatsby-image/). I actually wasn't aware of the `presentationHeight` and `presentationWidth` in the sharp plugin, if that's actually a use case that solves some problems it'd be worth documenting that somewhere too (doesn't look like it's in the `gatsby-plugin-sharp` README or the docs AFAIK). Gatsby image has always been a little hard to teach, so deciding what to do under the hood and how much to try and explain through docs is a tricky balance. All of these efforts should help a lot though 🙂

wardpeet:
Hey @sbardian sorry for taking so long but could you make an example site which has a lot of different image resolutions and see if this actually fixes things for all cases. It's going to be easier to get this through.

Thanks for your patience and sorry for taking so long.

sbardian:
@wardpeet sure thing, will start working on it.

sbardian:
@wardpeet I have deployed two sites to netlify for this test, so we can see them side by side.  I can add more tests or provide more specifics on the queries if needed? 

gatsby-image-current-master: https://gatsby-image-current-master.netlify.com/
gatsby-image-with-pr17006: https://gatsby-image-with-pr17006.netlify.com/

sbardian:
@wardpeet, a comment a day keeps the stale bot away. . . 🍎

sbardian:
@wardpeet figured I would ping you on this one, any time to check it out yet?

smashercosmo:
I think it's time for another friendly ping @pieh @wardpeet :) these are very much needed changes

vsimak:
bump!

gentksb:
I wish this PR to be merged!!

wardpeet:
In the current state with the extra wrapper `gatsby-image-wrapper-container` div makes this PR a breaking change as this might break sites who do child selector css on gatsby-image. I'll play with it and see what we need to do to get this in without a breaking change.

sbardian:
@wardpeet good point.  I'll dig about as well.  Let me know if you need anything else from me.

sbardian:
@wardpeet maybe I just have not looked at this code for a long time. . . but it seems to work fine just removing that `gatsby-image-wrapper-container` div  . . .?  Could have sworn that was needed though :)

sbardian:
@wardpeet  I removed that `gatsby-image-wrapper-container` div, as I can't find why it is needed anymore :)  I have also updated the test/demo site on [netlify](https://gatsby-image-with-pr17006.netlify.com/) and that demo sites repo can be found [here](https://github.com/sbardian/my-default-starter/blob/master/src/components/image.js).  I guess play around, let me know if you see where I might have broke what I was attempting prior with that `div` as I'm not seeing it . . .

sbardian:
@wardpeet one last ping. . .

packages/gatsby-image/src/index.js
```diff
@@ -546,7 +552,7 @@ class Image extends React.Component {
                 src={image.src}
                 crossOrigin={this.props.crossOrigin}
                 srcSet={image.srcSet}
-                style={imageStyle}
+                style={{ imageStyle }}
```
smashercosmo: I'm not sure this is correct

wardpeet:
Sorry for letting you wait again... I have it on my todo list. I'll move it up so we can ship this 👍

packages/gatsby-image/src/index.js
sbardian: @smashercosmo hmm agreed, not sure where this is coming from actually. . . I didn't update this intentionally.  Pushing fix.

sbardian:
@wardpeet no problem.  I think I fixed up the issue @smashercosmo caught.

AAverin:
Is this coming any time soon?
At the moment specifying maxHeight for fluid images doesn't size the image to expected maximal height instead seemingly ignoring the property.

wardpeet:
This looks great, I'm running some tests and hopefully, I can merge this today.

sbardian:
@wardpeet you might still be working on this, if so ignore me. Pulled down the update and it seems to break the fix.  Though I'm not able to run `yarn run bootstrap` and the project might not be getting loaded into my dev environment correctly?  FYI . . .

wardpeet:
Hey @sbardian, what do you mean it doesn't work anymore? It seems to work fine for me but maybe I'm testing it the wrong way :grimacing:

```js
export const query = graphql`
  query {
    allImageSharp {
      nodes {
        wh: fluid(maxWidth: 800) {
          ...GatsbyImageSharpFluid_withWebp
        }
        hw: fluid(maxHeight: 500) {
          ...GatsbyImageSharpFluid_withWebp
        }
      }
    }
  }
```

localhost_8000_page-2_

sbardian:
@wardpeet let me pull down the latest commits and see what I get today.  I was testing against the project I used before to test [here](https://github.com/sbardian/my-default-starter).  If you have time pull it down and see if it is still working for you as well maybe?  I'll follow up here in a few. Thanks!

packages/gatsby-image/src/index.js
```diff
@@ -480,6 +480,8 @@ class Image extends React.Component {
           style={{
             position: `relative`,
             overflow: `hidden`,
+            maxWidth: image.maxWidth ? `${image.maxWidth}px` : null,
```
sbardian, line 483: @wardpeet testing it against my test site, I think the issue is here.  No need to use `${image.maxWidth}px` if you log maxWidth/maxHeight it already has px appended.  When I remove that and just use `image.maxWidth/image.maxHeight` it works again. Thought not sure how it is working for you? :)

sbardian:
@wardpeet with your update to the fragments I was also not getting maxWidth/maxHeight coming acrossing my my project.  But I think that is only because I'm not sure how to get gatsby to update the fragments still, when I add them specifically I get the values.  

```
      cropWidth: file(relativePath: { eq: "gatsby-astronaut.png" }) {
        childImageSharp {
          fluid(maxWidth: 200, maxHeight: 600) {
            ...GatsbyImageSharpFluid
            maxHeight
            maxWidth
          }
        }
      }
```

packages/gatsby-image/src/index.js
sbardian: I also think that is why the PropType for `maxHeight/maxWidth` were both `string` instead of number under `fluidObject`.  As I'm not getting an error for invalid prop on those being number.

packages/gatsby-plugin-sharp/src/index.js
```diff
@@ -615,19 +615,9 @@ async function fluid({ file, args = {}, reporter, cache }) {
   }
 
   // when using maxHeight and/or maxWidth determine appropriate aspect ratios
-  let maxWidth
-  let maxHeight
   if (!options.aspectRatio) {
     options.aspectRatio = presentationWidth / presentationHeight
   }
-  if (options.maxHeight || options.maxWidth) {
```
sbardian: @wardpeet looking at `gatsby-plugin-sharp` I'm not seeing maxHeight/maxWidth being returned anymore for fluid image types, or the dimensions being calculated if only one (maxHeight or maxWidth) is used.  So maybe I'm just lost at this point on the new changes :)

wardpeet:
Hey sorry, i changed the PR a bit so we don't change behavior in gatsby-image v2. In v3 we should do stuff differently like we do now.

I've added a new fragment called `GatsbyImageSharpFluidLimitPresentationSize` that you can add to your sharp query to make this work. See it as an opt-in feature. After the merge, I'll chat with our docs team to make it more obvious on how to use it in our tutorials.

You'll have to do the following:
```
       childImageSharp {
          fluid(maxWidth: 500) {
            ...GatsbyImageSharpFluid
            ...GatsbyImageSharpFluidLimitPresentationSize
          }
        }
```

What do you think of this change?

sbardian:
@wardpeet I see.  Looks good.  Trying to test it but it thinks the fragment doesn't exist.  I'm not sure how to get gatsby to "refresh" the fragments.

packages/gatsby-transformer-sharp/src/fragments.js
```diff
@@ -119,7 +119,7 @@ export const GatsbyImageSharpFluid = graphql`
  * Presentation sizes to make sure a fluid container does not overflow
  * @type {Fragment}
  */
-export const GatsbyImageSharpFluidLimitPresentationSize = `
+export const GatsbyImageSharpFluidLimitPresentationSize = graphql`
   fragment GatsbyImageSharpFluidLimitPresentationSize on ImageSharpFluid {
     maxHeight: presentationHeight
```
sbardian, line 124: @wardpeet if you are setting the maxHeight/maxWidth to the presentationHeight/presentationWidth I'm not sure this will end up being correct, see this comment of mine on the issue of using presentationHeight/Width as maxHeight/Width. [comment](https://github.com/gatsbyjs/gatsby/pull/17006#issuecomment-532056611).  Basically the presentationHeight/Width max out at the images dimensions, so if your image is 800x800 and you set maxWidth to 1500, it will still only be maxWidth 800.

sbardian: and this [comment](https://github.com/gatsbyjs/gatsby/pull/17006#issuecomment-532757200) also.

sbardian: I was finally able to get the fragments to update and working using: 

```
       childImageSharp {
          fluid(maxWidth: 500) {
            ...GatsbyImageSharpFluid
            ...GatsbyImageSharpFluidLimitPresentationSize
          }
        }
```

and it does appear to be using presentationHeight/Width.  So any maxHeight/width will max out at the image dimensions.  This is why I had the logic in `gatsby-plugin-sharp` 

```
  let maxWidth
  let maxHeight
  if (options.maxHeight || options.maxWidth) {
    maxWidth = options.maxWidth
      ? `${options.maxWidth}px`
      : `${options.aspectRatio * options.maxHeight}px`
    maxHeight = options.maxHeight
      ? `${options.maxHeight}px`
      : `${options.aspectRatio * options.maxWidth}px`
  }
```
and specifically returned `maxHeight/maxWidth` it as part of the fluid return object.

wardpeet: Setting it to presentationWidth/height feels more correct as your the container shouldn't be bigger than image. If you need that, you'll need a bigger image, else you'll end up with quality drop.

If you want to center the image or override the behavior you can still add a wrapper or set the maxWidth/maxHeight yourself on gatsby-image.

sbardian: If this makes more sense I'm good with it.

sbardian:
Thanks @wardpeet !

sbardian:
@wardpeet sorry to be a downer. . . but it appears using `presentationHeight/presentationWidth` also broke cropping.  So with the following query: 
```
      cropWidth: file(relativePath: { eq: "gatsby-astronaut.png" }) {
        childImageSharp {
          fluid(maxWidth: 200, maxHeight: 600) {
            ...GatsbyImageSharpFluid
            ...GatsbyImageSharpFluidLimitPresentationSize
          }
        }
      }
```
Instead of getting a skinny tall image cropped to your specifics, you end up with a 200x200 image.  One way to fix this is to remove the documentation comment in the example [here](https://www.gatsbyjs.org/packages/gatsby-image/?=gat#query-1) so no one expects a cropping behavior I guess?   Thoughts?

pieh:
The PR was published in:
- `gatsby-image@2.4.3`
- `gatsby-plugin-sharp@2.6.2`
- `gatsby-transformer-sharp@2.5.2`

The cropping problem might need to be looked at :(

wardpeet:
I will :)

sbardian:
@wardpeet I think I have a fix for this

by replacing `gatsby-plugin-sharp` /src/index.js lines 477-484 [here](https://github.com/gatsbyjs/gatsby/blob/8e6d5c6fd91b06c368c78201d3683e1cc6d07cb1/packages/gatsby-plugin-sharp/src/index.js#L477) with the following: 

```
  let presentationWidth, presentationHeight

  if (!options.aspectRatio) {
    options.aspectRatio = width / height
  }

  if (fixedDimension === `maxWidth`) {
    presentationWidth = Math.min(options.maxWidth, width)
    presentationHeight = Math.round(
      options.maxHeight
        ? options.maxHeight
        : options.maxWidth * options.aspectRatio
    )
  } else {
    presentationHeight = Math.min(options.maxHeight, height)
    presentationWidth = Math.round(
      options.maxWidth
        ? options.maxWidth
        : options.maxHeight * options.aspectRatio
    )
  }
```
We get cropping back.  The only issue is the image will grow to match a maxWidth or maxHeight query value if given.  BUT it will not grow beyond the dimensions of the image itself.  So if the image was 200x500 and they had maxHeight: 2000, it will max out at the 500 image height.  

I also noticed some odd behavior in the plugin `healOptions` function.  It returns 800 width for images maxWidth if one is not supplied.  I would think this should default to the images width dimension?  Anyway, not sure if you wanted a new PR to help fix this issue or not.  Let me know if so.  I have also deployed this site [here](https://5eb45425488e8d02695c96aa--gatsby-image-with-pr17006.netlify.app/) This demo site has updated images and the query descriptions towards the bottom.  You can see the queries and demo code [here](https://github.com/sbardian/my-default-starter/blob/master/src/components/image.js)

wardpeet:
@sbardian thanks for your thorough comment! Seems like calculating presentationWidth & height is a mistake, we should get it from the image itself. I've created https://github.com/gatsbyjs/gatsby/pull/23905 to deal with that.

> Basically remove first if condition, no point setting maxWidth if the user does not pass one, we will use the image dimensions.

Sadly changing this default will be a breaking change, I do think in a next version we should make maxWidth or maxHeight mandatory.

Issue #16987, opened 2019-08-23
[gatsby-source-filesystem] Passing folder in graphql. windows/linux


trickydisco78:
When running gatsby graphql query that has folder access

```  
allFile(filter: { relativeDirectory: { eq: "player/profiles" } }) {
      edges {
        node {
          publicURL
          relativePath
        }
      }
    }
  }
```

On windows this won't generate the json page data as it needs to be ```players\\profiles``` but on mac/linux it needs to be ```players/profiles```

How do you i resolve this? how do i pass ```path.resolve('players/profiles')``` into that query without using createPage in gatsby-node. Is it possible to pass query variables inside of a react component?

stefanprobst:
It looks like this is a bug in `gatsby-source-filesystem`. When storing file paths in `File` nodes, we normally use the `slash` package to convert to forward slash format. We do this for the `absolutePath`, `relativePath` and `dir` fields, but apparently not for `relativeDirectory`.

Would need to wrap this line in `slash`:
https://github.com/gatsbyjs/gatsby/blob/a492167f23cfea28080e21d901a1858301237bad/packages/gatsby-source-filesystem/src/create-file-node.js#L21-L24

Not 100% sure though if this would be bug fix or breaking change?

sidharthachatterjee:
> Not 100% sure though if this would be bug fix or breaking change? 

This is a bug fix! Thank you so much @stefanprobst 

Fixed and published in `gatsby-source-filesystem@2.1.18`

PR #17048, opened 2019-08-24
fix(gatsby-source-filesystem): Use forward slashes on relativeDirectory

stefanprobst:
When creating `File` nodes, `gatsby-source-filesystem` ensures file paths in forward-slash format for the `dir`, `relativePath` and `absolutePath` fields, but not for `relativeDirectory`.
This PR fixes that, but I'm not sure if anybody on Windows might rely on the previous behavior?

Related: #16987


sidharthachatterjee:
Published in `gatsby-source-filesystem@2.1.18`

PR #17056, opened 2019-08-24
fix(gatsby-link) Clean up IntersectionObservers to fix a memory leak

### LINKED ISSUE #12198: Potential memory leak when navigating between pages
## Description

If this is expected behavior or I'm reading the Memory tab wrong, feel free to close this  😅 

What I'm seeing is when navigating between pages (even retracing steps back to an already-visited page), the JS heap size keeps growing.

### Steps to reproduce

This happens on both my website and the React site. 

1. Open https://reactjs.org/
2. Open DevTools > Memory > Take Heap Snapshot
3. Result: 13.7MB
4. Click 'Get Started' and click the 'Main Concepts' subheading on the right. Navigate through every page.
5. Take another snapshot > 22.6MB
6. Retrace all the steps backwards (already visited pages)
7. Take another snapshot > 28.9MB

### Expected result

JS heap doesn't keep growing when navigating around, esp. already visited pages?

### Actual result

Continues to grow

phacks:
## Description

This PR fixes a memory leak caused by `IntersectionObserver`s not being cleaned up in case the intersection is never reached.

Many thanks to @atomiks who did a thorough investigation on the matter and came up with this solution 👏

I tested it on https://reactjs.org with the following procedure:

- Head to https://reactjs.org/docs/getting-started.html
- Take a Heap Snapshot
- Switch back and forth to https://reactjs.org/docs/add-react-to-a-website.html multiple times
- Take another Heap snapshot
- The presence of the memory leak can be detected by filtering with `Intersection` in the memory heap snapshot and seeing `Detached…` entries.

Here’s a comparison of before and after the fix (tested locally):

| 🙁Before | 🙂After |
|---|---|
| Memory Leak before PR | No memory leak after PR |

## Related Issues

Fixes #12198.

### DIFF
--- a/packages/gatsby-link/src/index.js
+++ b/packages/gatsby-link/src/index.js
@@ -79,6 +79,9 @@ class GatsbyLink extends React.Component {
   }
 
   componentWillUnmount() {
+    if (!this.io) {
+      return
+    }
     const { instance, el } = this.io
 
     instance.unobserve(el)
--- a/packages/gatsby-link/src/index.js
+++ b/packages/gatsby-link/src/index.js
@@ -30,7 +30,7 @@ const NavLinkPropTypes = {
 }
 
 // Set up IntersectionObserver
-const handleIntersection = (el, cb) => {
+const createIntersectionObserver = (el, cb) => {
   const io = new window.IntersectionObserver(entries => {
     entries.forEach(entry => {
       if (el === entry.target) {
@@ -46,6 +46,7 @@ const handleIntersection = (el, cb) => {
   })
   // Add element to the observer
   io.observe(el)
+  return { instance: io, el }
 }
 
 class GatsbyLink extends React.Component {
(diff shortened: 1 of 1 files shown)


KyleAMathews:
Nice! Thanks!

Does this need done in gatsby-image as well?

phacks:
Don’t know yet, will have a look tomorrow!

phacks:
I did a bit of testing and could not reproduce the memory leak for `gatsby-image`. As it turns out, the leak has already been fixed by @alexandernanberg in #10278!

https://github.com/gatsbyjs/gatsby/blob/49ea6dc15e340fe0a8231509b14289f0aae1b4f3/packages/gatsby-image/src/index.js#L322-L326

packages/gatsby-link/src/index.js
```diff
@@ -77,6 +78,13 @@ class GatsbyLink extends React.Component {
     }
   }
 
+  componentWillUnmount() {
+    const { instance, el } = this.io
```
wardpeet: Let us make sure this.io is set.

```js
if (!this.io) { return }
```

phacks: Ohh yes, forgot about that. Fixed in 5b827ab!

PR #17060, opened 2019-08-25
docs: authenticated graphql server with auth0 and netlify functions

omonk:
## Description

This PR introduces a new example detailing how to create a secure Apollo GraphQL server using Auth0, Netlify functions and Gatsby

I open this PR for two main reasons:
- Share some work I've put together to help others quick start an authenticated GraphQL server
- Look for feedback and advice on whether how I'm performing auth is the correct way. Everything works but auth is a big topic of discussion and something I'm fairly new to, I'd like to get everything right so feedback would be greatly appreciated.

## Feedback requested

Is having different content in gatsby-ssr and gatsby-browser ok? I would keep them the same but as the `useAuth` hook would be fired on SSR and I need to pass an auth token to my client in the `<ApolloProvider>` the build would fail as it was trying to use the hook on the server. I tested out wrapping the AuthProvider in a conditional but due to the way hooks work it's advised not to do that... any thoughts on this would be great.

## Missing functionality
~~- Keeping user logged in between refreshes... this is something I had working and [added to the useAuth docs](https://github.com/Swizec/useAuth#persisting-login-after-refresh) to fix but it appears to have broken... I need to do some more investigation as to why its now broken.~~
Privacy badger prevent the cookies being set by Auth0, now it's working!


omonk:
I'm not sure why the linter is saying it can't find `eslint-config-xo-react/space`, it's been in this package for around 3 years... 

https://github.com/xojs/eslint-config-xo-react/blob/master/space.js

wardpeet:
Thanks @omonk  for your hard work on this example. It's super comprehensive. Sadly I don't think it's something we would like to have in our examples. This will be a burden to keep it up to date. I would love it if you could publish this as a starter or theme where you can own it and get the recognition you deserve for this example!

Thank you for using gatsby :muscle: and creating this excellent boilerplate! Many people will find this super helpful.

omonk:
@wardpeet - totally understand! 

Will try and find some time to make it into a boilerplate!

Issue #16651, opened 2019-08-15
importing JSON file into a page adds it in all pages


tareefdev:
## Description

Importing JSON file into a page directly (`import things from '../things'`) adds the imported file to all pages 

### Steps to reproduce

Here is a [repo](https://github.com/tareefdev/gatsby-json) to reproduce:

- Add a JSON file to the default starter and import it directly without using GraphQL
- Build the project and check pages size

### Expected result

- Just one page (where the file was imported) should increase in the size.

### Actual result

- All pages now increased in size from 1 MB to 5.7 MB

### Environment

  System:
    OS: Linux 4.15 Ubuntu 18.04.2 LTS (Bionic Beaver)
    CPU: (2) x64 Intel Xeon Processor (Skylake, IBRS)
    Shell: 4.4.20 - /bin/bash
  Binaries:
    Node: 8.16.0 - ~/.nvm/versions/node/v8.16.0/bin/node
    Yarn: 1.17.3 - /usr/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v8.16.0/bin/npm
  Languages:
    Python: 2.7.15+ - /usr/bin/python
  npmPackages:
    gatsby: ^2.13.64 => 2.13.64
    gatsby-image: ^2.2.8 => 2.2.8
    gatsby-plugin-feed: ^2.3.6 => 2.3.6
    gatsby-plugin-google-analytics: ^2.1.7 => 2.1.7
    gatsby-plugin-manifest: ^2.2.5 => 2.2.5
    gatsby-plugin-offline: ^2.2.6 => 2.2.6
    gatsby-plugin-react-helmet: ^3.1.3 => 3.1.3
    gatsby-plugin-sharp: ^2.2.11 => 2.2.11
    gatsby-plugin-typography: ^2.3.2 => 2.3.2
    gatsby-remark-copy-linked-files: ^2.1.6 => 2.1.6
    gatsby-remark-images: ^3.1.11 => 3.1.11
    gatsby-remark-prismjs: ^3.3.5 => 3.3.5
    gatsby-remark-responsive-iframe: ^2.2.4 => 2.2.4
    gatsby-remark-smartypants: ^2.1.2 => 2.1.2
    gatsby-source-filesystem: ^2.1.9 => 2.1.9
    gatsby-transformer-remark: ^2.6.14 => 2.6.14
    gatsby-transformer-sharp: ^2.2.6 => 2.2.6
  npmGlobalPackages:
[...]

PR #17061, opened 2019-08-25
docs: Importing JSON files correctly outside GraphQL

tareefdev:
## Description

`gatsby-transformer-json` is a great plugin to parse raw JSON. Although the plugin works perfectly with well-structured JSON files, it's faile to read and parse some type of files like Topojson.
The "Troubleshooting" section within plugin's README suggests rewriting the data, which not possible most of the time.

Trying to import those files directly inside a JS file results in adding them to the app bundle and that would increase the size of all pages, as I discovered in #16651 issue.  

## Related Issues
 Related to #16651


tareefdev:
Thanks for your review, I'm thinking of adding a similar note to the [Troubleshooting](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-transformer-json/README.md#troubleshooting) section of gatsby-transformer-json plugin, where the document suggests restructuring the data when it's not has a consistent schema, which is not possible in most cases.

wardpeet:
Hey, sorry for my late response. The advice you're giving isn't 100% correct. Instead of advertising a http request libs like axios. We can simply use import('../file.json') which creates a separate chunk and gives us the same ability as axios but without the extra lib. The useEffect/componentDidMount part is still necessary so :ok-hand: on that!

If your json doesn't have a consistent schema, we recommend writing your own source-plugin. A source-plugin gives you more benefits on the grep the data you need heuristic of gatsby.

On the other hand, it's more complex setting up than using an import() statement. I'm fine with you just rewriting the recommendation using dynamic import instead of Axios.

What do you think?

LekoArts:
Hi @tareefdev 	👋 	Do you have time to address Ward's comments? I agree with him that we shouldn't merge this as-is. Thanks!

tareefdev:
Hello all,

I used `axios` here because several documentation pages mentioned it, so I thought there is no problem. But if it's preferred to use dynamic import I could change that.

wardpeet:
@tareefdev we like to keep our advice dependency-free so dynamic imports is what we want for these. One of the benefits here is that it becomes part of the webpack bundle which gets hashed and only updated when it changes so you don't have to think about clearing cache, ...

docs/docs/static-folder.md
```diff
@@ -48,3 +48,4 @@ folder is useful as a workaround for a number of less common cases:
   [`pace.js`](http://github.hubspot.com/pace/docs/welcome/) outside of the
   bundled code.
 - Some libraries may be incompatible with Webpack and you have no other option but to include it as a `<script>` tag.
+- You need to import JSON file that doesn't have a consistent schema, like [TopoJSON files](https://en.wikipedia.org/wiki/GeoJSON#TopoJSON), which GraphQL can't handle correctly. Note that importing JSON files directly inside a page, a template, or a component using `import` syntax results in adding that file to the app bundle and increasing the size of all site's pages. Instead, it's better to place your JSON file inside the `static` folder and then fetch that file using something like [axios](https://github.com/axios/axios/) within `componentDidMount` lifecycle or `useEffect` hook.
```
wardpeet: WDYT of? Mainly your words but mixed in the dynamic import part.
```suggestion
- You need to import JSON file that doesn't have a consistent schema, like [TopoJSON files](https://en.wikipedia.org/wiki/GeoJSON#TopoJSON), which is difficult to handle with GraphQL. Note that importing JSON files directly inside a page, a template, or a component using `import` syntax results in adding that file to the app bundle and increasing the size of all site's pages. Instead, it's better to place your JSON file inside the `static` folder and use the dynamic import syntax (`import('/static/myjson.json')`) within the `componentDidMount` lifecycle or the `useEffect` hook
```

pieh: I think @wardpeet advice here makes a lot of sense. Given that it changes actual advice from using `static` directory to using webpack's dynamic import - I think it should also be moved to different documentation page - for example https://www.gatsbyjs.org/docs/importing-assets-into-files/#importing-assets-with-webpack

tareefdev: Hello again, I'm sorry to be late on this. 
@pieh I think it's important to add the same advice to gatsby-transformer-json`s README file too, as the plugin couldn't handle files which have a random schema. If it sounds OK I'm ready to send a small commit to add the same mention there.

pieh: Yeah, sounds reasonable to mention "gotcha"s in json transformer docs. Just one note - json transformer isn't limited to local files (it can work on content pulled from CMSes), so the advice for objects without consistent schema will not apply to all potential scenarios

muescha: in the import from webpack there is no mention of `json` maybe it should also added there?

muescha:
```
To reduce the number of requests to the server, importing images that are less than 10,000 bytes returns a data URI instead of a path. This applies to the following file extensions: svg, jpg, jpeg, png, gif, mp4, webm, wav, mp3, m4a, aac, and oga.
```

oh i think i misunderstood the second sentence

wardpeet:
@tareefdev I've updated the readme with the review comment. Let's merge this. Please create a new PR to update gatsby-transformer-json README.md

tareefdev:
Thanks for all of you, I'm really happy to contribute here.

Issue #14173, opened 2019-05-20
fetch data from wordpress, Failed to process remote content.


xuweiwei117526:
during our Gatsby build,fetch data from wordpress, we will see images which say Failed to process remote content.

Any assistance would be appreciated!

.source and transform nodes
.source and transform nodesFailed to process remote content http://site-url/sites/default/files/Name%20of%20file_2018_1.jpg
Failed to process remote content http://site-url/sites/default/files/Name%20of%20file_2018_1.jpg
Failed to process remote content http://site-url/sites/default/files/NameOfFile.gif

josh-meredith:
I have experienced this same issue, and it seems the only fix for now is to manually reduce the limit of concurrent downloads within the `gatsby-source-filesystem` plugin. The problem is most likely due to server/network overload.

Please see #13073 for the fix I'm referring to. Unfortunately it doesn't look like there has been much progress with the PR.

bakeruk:
@wardpeet This looks like the same issue as the recent `gatsby-source-drupal` that was merged into master (#13943).

All to do with unresolved promises overloading the callstack using with using `Promise.all` in the code

```
// Downloads media files and removes "sizes" data as useless in Gatsby context.
exports.downloadMediaFiles = async ({
  entities,
  store,
  cache,
  createNode,
  createNodeId,
  touchNode,
  getNode,
  _auth,
}) =>
  Promise.all(
    entities.map(async e => {
      let fileNodeID
      if (e.__type === `wordpress__wp_media`) {
        const mediaDataCacheKey = `wordpress-media-${e.wordpress_id}`
        const cacheMediaData = await cache.get(mediaDataCacheKey)

        // If we have cached media data and it wasn't modified, reuse
        // previously created file node to not try to redownload
        if (cacheMediaData && e.modified === cacheMediaData.modified) {
          const fileNode = getNode(cacheMediaData.fileNodeID)

          // check if node still exists in cache
          // it could be removed if image was made private
          if (fileNode) {
            fileNodeID = cacheMediaData.fileNodeID
            touchNode({
              nodeId: fileNodeID,
            })
          }
        }

        // If we don't have cached data, download the file
        if (!fileNodeID) {
          try {
            const fileNode = await createRemoteFileNode({
              url: e.source_url,
              store,
              cache,
              createNode,
              createNodeId,
              parentNodeId: e.id,
              auth: _auth,
            })
...
```

[...]

Issue #6654, opened 2018-07-22
[gatsby-source-wordpress] Large WordPress site causing extremely slow build time (stuck at 'source and transform nodes')


dustinhorton:
## Description

`gatsby develop` hangs on `source and transform nodes` after querying a large WordPress installation (~9000 posts, ~35 pages).

Is there any guides as to what's too big for Gatsby to handle in this regards?

### Environment

```
  System:
    OS: macOS High Sierra 10.13.6
    CPU: x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 8.10.0 - ~/n/bin/node
    Yarn: 1.5.1 - ~/n/bin/yarn
    npm: 5.6.0 - ~/n/bin/npm
  Browsers:
    Chrome: 67.0.3396.99
    Safari: 11.1.2
  npmPackages:
    gatsby: ^1.9.273 => 1.9.273
    gatsby-image: ^1.0.54 => 1.0.54
    gatsby-link: ^1.6.45 => 1.6.45
    gatsby-plugin-google-analytics: ^1.0.27 => 1.0.31
    gatsby-plugin-postcss-sass: ^1.0.22 => 1.0.22
    gatsby-plugin-react-helmet: ^2.0.10 => 2.0.11
    gatsby-plugin-react-next: ^1.0.11 => 1.0.11
    gatsby-plugin-resolve-src: 1.1.3 => 1.1.3
    gatsby-plugin-sharp: ^1.6.48 => 1.6.48
    gatsby-plugin-svgr: ^1.0.1 => 1.0.1
    gatsby-source-filesystem: ^1.5.39 => 1.5.39
    gatsby-source-wordpress: ^2.0.93 => 2.0.93
    gatsby-transformer-sharp: ^1.6.27 => 1.6.27
  npmGlobalPackages:
    gatsby-cli: 1.1.58
```

edit: Just want to reiterate—this is *not* something easily fixable by deleted `.cache/`, `.node_modules`, etc. If that resolves your problem, you weren't experiencing this issue.

pieh:
Can You prepare reproduction repo? Number of posts shouldn't be a problem (at least at this step) - v1 might get into memory problems but this would be in later build step and shouldn't get stuck

dustinhorton:
@pieh Whoops, posted that before I saw you'd just replied. I can try to get this site up remotely tomorrow.

[...]

PR #17063, opened 2019-08-25
feat(gatsby-source-filesystem): Added delay to download queue

robmarshall:
## Description

This change adds the ability to delay between remote file downloads. It follows the same logic as the GATSBY_CONCURRENT_DOWNLOAD solution. 

GATSBY_CONCURRENT_DOWNLOAD can reduce server stress, but running the create-remote-file-node process to pull a large number of files from a cheaper hosting service is too much. The server seems to be unable to deal with such a high demand of constant requests (on top of running another other queries). 

The outcome is having bandwidth/CPU limited and killing any current processes.

By introducing a delay gives the server a chance to breath. An example of usage for a low end server would be concurrent downloads set to 10, with delay set to 5 seconds.

Although not perfect (would be better to re-write the whole section) it is a quick fix for a large problem.

## Related Issues

I believe the following issues would be fixed by this PR:

https://github.com/gatsbyjs/gatsby/issues/6654
https://github.com/gatsbyjs/gatsby/issues/12848
https://github.com/gatsbyjs/gatsby/issues/14173
https://github.com/graysonhicks/gatsby-plugin-remote-images/issues/16

## Future

Following this logic, it would be great to split the queuing section into its own function, and pass more arguments for [better-queue](https://github.com/diamondio/better-queue) such as batchSize and batchDelay. This would give the user a little more flexibility in managing their own stack needs.

### DIFF
--- a/packages/gatsby-source-filesystem/src/create-remote-file-node.js
+++ b/packages/gatsby-source-filesystem/src/create-remote-file-node.js
@@ -76,6 +76,7 @@ const queue = new Queue(pushToQueue, {
   id: `url`,
   merge: (old, _, cb) => cb(old),
   concurrent: process.env.GATSBY_CONCURRENT_DOWNLOAD || 200,
+  afterProcessDelay: process.env.GATSBY_DELAY_DOWNLOAD || 0,
 })
 
 // when the queue is empty we stop the progressbar


KyleAMathews:
It'd be nice to make downloading smarter and not add more options. Most people aren't going to know how to use options like this so are going to have things fail and then give up.

robmarshall:
I completly agree. But the issues that have been shown above need more control over the download rates. 

It would make more sense for the queue initiation to be packaged up into a function, giving users the opportunity to add extra options where needed. It can then fall back to default when nothing is supplied.

I suppose one way it could work would be similar to a try catch. If a file cannot be downloaded (gets a 504 from being limited) it gets delayed slightly and retried. Reducing download rates/concurrent files each try up to 5 attempts.

My concern would be this approach pushes the server to breaking, and can have a knock on effect to other services. Maybe a combination of options/try/catch could work?

Would you be happy to merge this PR to solve the current issue, and I will look into a more permanent solution over the coming weeks?

wardpeet:
Hey @robmarshall, as Kyle stated this isn't a proper fix. Delaying requests will slowdown all usages of createRemoteFile.

This has been a known issue for a while, we have created an issue https://github.com/gatsbyjs/gatsby/issues/10049 on how to improve it. We really love to get this solved but only using a fix that will fix all source plugins.

robmarshall:
That it interesting! Nice idea. Didn't come across that whilst researching. Thanks for the link!

PR #17065, opened 2019-08-25
fix(gatsby-image): Make GatsbyImageProps's FixedObject and FluidObject props optional

patrick91:
When running codegen on GatbsyImage fields the returned types have null in it (which is the expected behaviour in GraphQL when a field is optional).

This should fix code generation issues (it does fix it for me at least), the only issue I see is
that maybe the optional fields I added should not be optional, in that case we should change the
GraphQL schema too, let me know if that's the case and I'll see if I can change that too :)

### DIFF
--- a/packages/gatsby-image/index.d.ts
+++ b/packages/gatsby-image/index.d.ts
@@ -1,34 +1,34 @@
 import * as React from "react"
 
 export interface FixedObject {
-  width: number
-  height: number
-  src: string
-  srcSet: string
-  base64?: string
-  tracedSVG?: string
-  srcWebp?: string
-  srcSetWebp?: string
-  media?: string
+  width?: number | null
+  height?: number | null
+  src?: string | null
+  srcSet?: string | null
+  base64?: string | null
+  tracedSVG?: string | null
+  srcWebp?: string | null
+  srcSetWebp?: string | null
+  media?: string | null
 }
 
 export interface FluidObject {
-  aspectRatio: number
-  src: string
-  srcSet: string
-  sizes: string
-  base64?: string
-  tracedSVG?: string
-  srcWebp?: string
-  srcSetWebp?: string
-  media?: string
+  aspectRatio?: number | null
+  src?: string | null
+  srcSet?: string | null
+  sizes?: string | null
+  base64?: string | null
+  tracedSVG?: string | null
+  srcWebp?: string | null
+  srcSetWebp?: string | null
+  media?: string | null
 }
 
 interface GatsbyImageProps {
   resolutions?: FixedObject
   sizes?: FluidObject
-  fixed?: FixedObject | FixedObject[]
-  fluid?: FluidObject | FluidObject[]
+  fixed?: FixedObject | FixedObject[] | null
+  fluid?: FluidObject | FluidObject[] | null
   fadeIn?: boolean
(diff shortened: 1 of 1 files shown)


wardpeet:
This seems to make sense as `undefined` & `null` are not treated by `?`, only `undefined` is. I'll need to have a look if the typings are correct but as far as I can tell this looks ACE! 💯

wardpeet:
Could you give me an example of how you generate types and which package you use to do so? The typescript file doesn't match with proptypes.

patrick91:
@wardpeet I'm using GraphQL codegen, so the types are coming directly from GraphQL, so they might not match the proptypes, not sure which are the wrong ones :)

Do you want me to setup a small demo to show the generated types? I have one on github but the setup is a bit complicated, so if needed I can create a small repro :) 

Happy to send PR to fix either the GraphQL types or the prop types :)

wardpeet:
Oh a small repro would work pretty great for me :)

patrick91:
@wardpeet will do that tomorrow then!

patrick91:
@wardpeet did this real quick: https://github.com/patrick91/gatsby-typing-example
hope it helps :)

wardpeet:
The problem I'm having is that the graphql schema we output isn't correct. Some values should have notnull in there as well which we don't so everything is Maybe...

You'll always need to cast it because of our faulty schema. I'm a bit afraid of editing our current schema with notNull to have better support for this.

patrick91:
> I'm a bit afraid of editing our current schema with notNull to have better support for this.

why?

wardpeet:
I'm unsure if nullable types are intended or not. This could break lots of sites if we make a wrong guess.

wardpeet:
I talked with @pieh as he has more context on these things. We can go forward and mark all non-nullable fields in https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-transformer-sharp/src/extend-node-type.js with non-nullable. So basically it's everything except webpsrc.

patrick91:
@wardpeet awesome, I'll try to send a PR in the next few days!

LekoArts:
Hi @patrick91 	👋 	Did you find time to update your PR? 😊

patrick91:
@LekoArts not yet, it is in my todo list, but I'm a bit busy!

pieh:
Hey @patrick91! Given this comment from @wardpeet 

> I talked with @pieh as he has more context on these things. We can go forward and mark all non-nullable fields in https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-transformer-sharp/src/extend-node-type.js with non-nullable. So basically it's everything except webpsrc.
> 
> Could you update those and hopefully we don't have to change gatsby-image/index.d.ts

One update to the comment above is that referenced file has moved recently to https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-transformer-sharp/src/customize-schema.js.

We will need to change types to be wrapped by `GraphQLNonNull` higher order type - for example:

```js
          width: {
            type: new GraphQLNonNull(GraphQLFloat)
          },
```

I'll close this PR as the change need to happen in `gatsby-transformer-sharp` and not in gatsby-image props types, so there is no value in keeping this PR open.

Thanks again for being part of the Gatsby community!

PR #17067, opened 2019-08-25
feat(gatsby-plugin-google-analytics): Add more checks of Do Not Track preference

crazko:
## Description

This PR adds more specific condition to check whether user wants to send Do Not Track preference.

- adds older browsers compatibility (IE9, IE10, Gecko-based pre 32 version), see https://caniuse.com/#feat=do-not-track
- updates condition for possible DNT value extensions, see https://www.w3.org/TR/tracking-dnt/#dnt-extensions

### DIFF
--- a/packages/gatsby-plugin-google-analytics/src/gatsby-ssr.js
+++ b/packages/gatsby-plugin-google-analytics/src/gatsby-ssr.js
@@ -78,7 +78,7 @@ export const onRenderBody = (
   if(${
     typeof pluginOptions.respectDNT !== `undefined` &&
     pluginOptions.respectDNT == true
-      ? `!(navigator.doNotTrack == "1" || window.doNotTrack == "1")`
+      ? `!(parseInt(navigator.doNotTrack) === 1 || parseInt(window.doNotTrack) === 1 || parseInt(navigator.msDoNotTrack) === 1 || navigator.doNotTrack === "yes")`
       : `true`
   }) {
     (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){


packages/gatsby-plugin-google-analytics/src/gatsby-ssr.js
```diff
@@ -78,7 +78,7 @@ export const onRenderBody = (
   if(${
     typeof pluginOptions.respectDNT !== `undefined` &&
     pluginOptions.respectDNT == true
-      ? `!(navigator.doNotTrack == "1" || window.doNotTrack == "1")`
+      ? `!(parseInt(navigator.doNotTrack) === 1 || parseInt(window.doNotTrack) === 1 || parseInt(navigator.msDoNotTrack) === 1 || navigator.doNotTrack === "yes")`
```
wardpeet: We can move to `==` instead of `===` which fixes your issue too with fewer bytes.

```suggestion
      ? `!(navigator.doNotTrack == '1' || window.doNotTrack) == '1' || navigator.msDoNotTrack == '1' || navigator.doNotTrack === "yes")`
```

crazko: Thanks for the review. The reason of `parseInt()` is the possible [extension of doNotTrack](https://www.w3.org/TR/tracking-dnt/#dnt-extensions) value, which could be e.g. `1xyz`, so this way the value is parsed correctly. What do you think?

wardpeet: Oh make sense! 👍 thanks for clarifying.

crazko:
Thanks!

PR #17072, opened 2019-08-25
fix(gatsby-transformer-json): Coerce id field to always be a String

sidharthachatterjee:
We expect any field called `id` on a Node to be a String and break if it is not. So it makes sense for transformer-json to coerce any values in id to a String so that a user isn't required to manipulate input JSON data when using `gatsby-transformer-json`


sidharthachatterjee:
https://circleci.com/gh/gatsbyjs/gatsby/206608?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

This particular test is painfully flaky 😞 CC @davidbailey00

PR #17078, opened 2019-08-26
fix(gatsby): fix an infinite loop in node child collection

me4502:
## Description

Currently there are cases where a cyclic node graph can cause findChildrenRecursively to hit the limit of the stack. Also, recursive functions are notoriously slow in JS.

I've replaced this with a non-recursive implementation of a pre-order depth first search tree traversal, and also made it keep track of traversed nodes. This will prevent duplicate nodes being found, and also prevent infinite loops with cyclic node trees.

~Performance wise this should theoretically be the same or possibly slightly better - I have not done a performance benchmark.~ Edit: Just made a change and benchmarked it, this is 3x faster than before.

Correct me if I'm wrong - but renaming this method is fine as it is not exported, and all local usages have been fixed.

### DIFF
--- a/packages/gatsby/src/redux/actions/public.js
+++ b/packages/gatsby/src/redux/actions/public.js
@@ -27,7 +27,7 @@ const findChildren = initialChildren => {
   const traversedNodes = new Set()
 
   while (queue.length > 0) {
-    const currentChild = getNode(queue.shift())
+    const currentChild = getNode(queue.pop())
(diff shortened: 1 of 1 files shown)


ehowey:
Any chance this relates to what I am experiencing in this issue?

https://github.com/gatsbyjs/gatsby/issues/17131

me4502:
Very unlikely as this will actually crash node due to exhausting the stack

Issue #17062, opened 2019-08-25
[gatsby-transformer-remark] Filtering on tableOfContents in createResolvers makes Gatsby crash


problem233:
## Description

Filtering on `tableOfContents` with `context.nodeModel.runQuery` in `createResolvers` makes Gatsby crash.

### Steps to reproduce

A minimal demo: https://github.com/problem233/gatsby-bug-demo

Related code: [link](https://github.com/problem233/gatsby-bug-demo/blob/0c2f0c3fd7c3ebda36c2eeadfd3cd3e32444186a/gatsby-node.js#L16)

Run `npm run dev`, open Graph*i*QL and run this query:

```graphql
query MyQuery {
  someType {
    toc
  }
}
```

### Expected result

Data:
```json
{
  "data": {
    "someType": {
      "toc": "<ul>\n<li>\n<p><a href=\"//#hello\">Hello</a></p>\n<ul>\n<li><a href=\"//#world\">World</a></li>\n</ul>\n</li>\n</ul>"
    }
  }
}
```

### Actual result

Gastby crashed.

### Environment

```
  System:
    OS: Linux 5.2 Arch Linux undefined
    CPU: (4) x64 Intel(R) Core(TM) i5-7300HQ CPU @ 2.50GHz
    Shell: 5.0.7 - /bin/bash
  Binaries:
    Node: 12.9.0 - /usr/bin/node
    npm: 6.11.2 - /usr/bin/npm
  Languages:
    Python: 3.7.4 - /usr/bin/python
  Browsers:
    Firefox: 68.0.2
  npmPackages:
    gatsby: ^2.13.79 => 2.13.82 
    gatsby-source-filesystem: ^2.1.16 => 2.1.18 
    gatsby-transformer-remark: ^2.6.19 => 2.6.19 
```

stefanprobst:
There are a couple of issues here: the first is that the `tableOfContents` field currently requires a `pathToSlug` argument to produce a toc (not totally sure why tbh), and you cannot meaningfully run a filter on such a field. We actually plan to exclude fields that need an arg to produce a value from the filter type in the future. What you could do is get the field resolver from `info.schema.getType("MarkdownRemark").getFields().tableOfContents.resolve` and call it directly which is not pretty but should work.

The crash itself is because `tableOfContents` actually produces an invalid AST when no `pathToSlug` arg is provided which should be fixed in #17080

problem233:
[...]

PR #17080, opened 2019-08-26
fix(gatsby-transformer-remark): Don't produce invalid AST in TOC

stefanprobst:
The table of contents generation in `gatsby-transformer-remark` can produce an invalid AST (when no `pathToSlug` field is provided, `node.children` will be a sparse `[null]` array.

Unrelated: what's the reason `pathToSlug` is required in the first place?

### DIFF
--- a/packages/gatsby-transformer-remark/src/__tests__/__snapshots__/extend-node.js.snap
+++ b/packages/gatsby-transformer-remark/src/__tests__/__snapshots__/extend-node.js.snap
@@ -593,7 +593,14 @@ Object {
   "frontmatter": Object {
     "title": "my little pony",
   },
-  "tableOfContents": null,
+  "tableOfContents": "<ul>
+<li>
+<p></p>
+<ul>
+<li></li>
+</ul>
+</li>
+</ul>",
 }
 `;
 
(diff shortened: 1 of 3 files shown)


stefanprobst:
Also updated test+snapshot -- the test was returning `null` only because the [this line](https://github.com/gatsbyjs/gatsby/blob/9e3581a03587f731632933c06166b92e3c4822c4/packages/gatsby-transformer-remark/src/extend-node-type.js#L304) actually throws.

wardpeet:
How do I test if this PR works? I've run the example on the bug report but it still fails with

stefanprobst:
>  the example on the bug report but it still fails

yes, see [my comment](https://github.com/gatsbyjs/gatsby/issues/17062#issuecomment-525291717) there. this PR only fixes one part, which is the invalid ast produced by the `tableOfContents` resolver.

I've fixed the transformer-remark test setup with https://github.com/gatsbyjs/gatsby/pull/17080/commits/a2b9eb2bf68502f1a3cdff2df14e585c2fc4a110 , so without this PR one of the `tableOfContents` tests would now correctly fail.

Issue #16661, opened 2019-08-15
How to improve webpack splitChunks heuristics?


slorber:
[@Gatsby](https://twitter.com/gatsbyjs/status/1162047124764155905) suggested me on Twitter to open an issue on this subject.

I think it would be nice to document how an user can enhance gatsby/webpack default code splitting heuristics to achieve better code splitting.

The Webpack splitChunks documentation:
https://webpack.js.org/plugins/split-chunks-plugin/

The current Gatsby config is:
https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/utils/webpack.config.js#L469

```js
      splitChunks: {
        name: false,
        cacheGroups: {
          // Only create one CSS file to avoid
          // problems with code-split CSS loading in different orders
          // causing inconsistent/non-determanistic styling
          // See https://github.com/gatsbyjs/gatsby/issues/11072
          styles: {
            name: `styles`,
            // This should cover all our types of CSS.
            test: /\.(css|scss|sass|less|styl)$/,
            chunks: `all`,
            enforce: true,
          },
        },
      },
```

As we can see, apart from disabling css code splitting, Gatsby does not do much and relies on Webpack defaults.

I think it would be nice to document how a user can enhance those heuristics so that webpack can give more optimized outputs. 

Also it's probably worth discussing what is an optimized webpack output in the first place, in the context of a Gatsby static site. We might favor multiple smaller bundles so that they can be better cached independently, or downloaded/parsed in parallel. 

Webpack can't really know by himself how our code is going to change over time, and recipes or a plugin to tell him new heuristics could be nice.

I've been successful with this kind of config:

```js
exports.onCreateWebpackConfig = ({ stage, actions, getConfig }) => {
  if (stage === `build-javascript`) {
    const config = getConfig();

    // Optimize the default code splitting config by putting heavy things
    // that changes independently on separate bundles
    config.optimization.splitChunks.cacheGroups = {
      ...config.optimization.splitChunks.cacheGroups,
      vendors: {
        test: /[\\/]node_modules[\\/]/,
        enforce: true,
        chunks: 'all',
        priority: 1,
      },
      translations: {
        test: /[\\/]translations[\\/]/,
        enforce: true,
        chunks: 'all',
        priority: 1,
      },
    };

    // Ensure Gatsby does not do any css code splitting
    config.optimization.splitChunks.cacheGroups.styles.priority = 10;

    actions.replaceWebpackConfig(config);
  }
};
```

[...]

PR #17093, opened 2019-08-26
feat(gatsby): add better splitchunks config

### LINKED ISSUE #15302: Code splitting common vendor files
# Summary

I have a rather large package that is being built into all of my page bundles. 

I was under the impression from the docs that this would be ripped out into a common bundle, which is then only imported once? Am I missing a required option for this?

## Relevant information

Here is my bundle analysis;

Screen Shot 2019-07-02 at 12 59 31

As you can see `Dexie.es.js` is present in most of the bundles, along with some smaller common packages.

### Environment (if relevant)

```
  System:
    OS: macOS High Sierra 10.13.2
    Shell: 5.3 - /bin/zsh
  Binaries:
    Yarn: 1.15.2 - /usr/local/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v10.14.2/bin/npm
  Languages:
  Browsers:
    Chrome: 75.0.3770.100
    Safari: 11.0.2
  npmPackages:
    gatsby: 2.6.0 => 2.6.0 
    gatsby-image: ^2.0.15 => 2.0.25 
    gatsby-plugin-create-client-paths: ^2.0.2 => 2.0.3 
    gatsby-plugin-google-gtag: ^1.0.8 => 1.0.8 
    gatsby-plugin-manifest: ^2.0.5 => 2.0.12 
    gatsby-plugin-offline: ^2.0.11 => 2.0.20 
    gatsby-plugin-react-css-modules: ^2.0.7 => 2.0.7 
    gatsby-plugin-react-helmet: ^3.0.0 => 3.0.4 
    gatsby-plugin-sass: ^2.0.10 => 2.0.10 
    gatsby-plugin-sharp: ^2.0.18 => 2.0.18 
    gatsby-plugin-sitemap: ^2.0.12 => 2.0.12 
    gatsby-plugin-transition-link: ^1.12.2 => 1.12.2 
    gatsby-plugin-webpack-bundle-analyzer: ^1.0.5 => 1.0.5 
    gatsby-source-filesystem: ^2.0.4 => 2.0.12 
    gatsby-transformer-sharp: ^2.1.12 => 2.1.12 
```

Can include these if relevant.

wardpeet:
## Description
Creates a basic splitChunks config that works better than our current one. It's not the final config we would like as that one needs more testing. This one works pretty well so far.

Credits go to Nextjs as I copied the code from their codebase.
https://github.com/zeit/next.js/blob/canary/packages/next/build/webpack-config.ts#L200-L216

We want to go to a granular way of splitting chunks like nextjs does https://github.com/zeit/next.js/blob/canary/packages/next/build/webpack-config.ts#L217-L275 but create one that works for Gatsby.

This PR saves around **400kb** on .org and lowers visual-complete and TTI (fist-cpu-idle) by 400ms which will lower even more when guess-js is fixed.

Comparison can be found here:
https://www.webpagetest.org/video/compare.php?tests=190826_2H_5a44521034c35d43ff27a3af95b7c53c%2C190826_D7_ed0d076acfadfd3575943dfcf6cf34b8&thumbSize=200&ival=100&end=visual

Test current:
https://www.webpagetest.org/result/190826_D7_ed0d076acfadfd3575943dfcf6cf34b8/

Test this pr:
https://www.webpagetest.org/result/190826_2H_5a44521034c35d43ff27a3af95b7c53c/

## Related Issues
Fixes #15302

### DIFF
--- a/packages/gatsby/src/utils/webpack.config.js
+++ b/packages/gatsby/src/utils/webpack.config.js
@@ -468,13 +468,30 @@ module.exports = async (
   }
 
   if (stage === `build-javascript`) {
+    const componentsCount = store.getState().components.size
+
     config.optimization = {
       runtimeChunk: {
         name: `webpack-runtime`,
       },
       splitChunks: {
         name: false,
+        chunks: `all`,
         cacheGroups: {
+          default: false,
+          vendors: false,
+          commons: {
+            name: `commons`,
+            chunks: `all`,
+            // if a chunk is used more than half the components count,
+            // we can assume it's pretty global
+            minChunks: componentsCount > 2 ? componentsCount * 0.5 : 2,
+          },
+          react: {
+            name: `commons`,
+            chunks: `all`,
+            test: /[\\/]node_modules[\\/](react|react-dom|scheduler)[\\/]/,
+          },
           // Only create one CSS file to avoid
           // problems with code-split CSS loading in different orders
           // causing inconsistent/non-determanistic styling


packages/gatsby/src/utils/webpack.config.js
```diff
@@ -468,13 +468,30 @@ module.exports = async (
...
+        chunks: `all`,
         cacheGroups: {
+          default: false,
+          vendors: false,
+          commons: {
+            name: `commons`,
+            chunks: `all`,
+            // if a chunk is used more than half the components count,
+            // we can assume it's pretty global
+            minChunks: componentsCount > 2 ? componentsCount * 0.5 : 2,
```
sidharthachatterjee: More than half components count is a nice metric 😅

sidharthachatterjee: Does `store.getState().components.size` include pages?

wardpeet: I need components as they are the ones that create chunks. We can have 1000 of pages that only use 1 component so we need to check components and not pages.

sidharthachatterjee: Hmm, fair enough

wardpeet:
Added analysers
current master: https://5d63face1c5e2fc7e4802b5d--gatsbyjs-dev.netlify.com/old-chunks.html
PR: https://5d63face1c5e2fc7e4802b5d--gatsbyjs-dev.netlify.com/splitchunks.html

packages/gatsby/src/utils/webpack.config.js
slorber: Crazy idea: GuessJS for webpack config :D

slorber: Note I've opened a related issue where you'll find links to what has been done recently in NextJS which has a new "granularChunks" config: https://github.com/gatsbyjs/gatsby/issues/16661

wardpeet: Yeah so the idea is to go on an even more granular level, like nextjs but I wanted something I could get in fairly easy without having to check x amount of sites.

PR #17107, opened 2019-08-27
fix(gatsby-plugin-feed): replace title feed requirement with a warning

### LINKED ISSUE #17100: [gatsby-plugin-feed] 🚨BREAKING CHANGE ⚠️ builds now fail Config Validation with "title" is required
My builds on https://github.com/rdela/rdela.com started failing all of a sudden with a gatsby-plugin-feed error, `Config Validation with "title" is required`:

```
 ERROR #11321  PLUGIN

"gatsby-plugin-feed" threw an error while running the onPreBootstrap lifecycle:

[Config Validation]: "title" is required

  Error: [Config Validation]: "title" is required
  
  - gatsby-node.js:84 _callee$
    [rdela.com]/[gatsby-plugin-feed]/gatsby-node.js:84:19
```

Here is the PR from Renovate that introduced this bug/error: https://github.com/rdela/rdela.com/pull/186

```
gatsby-plugin-feed | dependencies | patch | 2.3.7 -> 2.3.9
```

By examining the link attached to 2.3.7 -> 2.3.9 in the PR description:
https://renovatebot.com/package-diff?name=gatsby-plugin-feed&from=2.3.7&to=2.3.9

I was able to track down https://github.com/gatsbyjs/gatsby/pull/16814 where @eyalroth writes 

> Note that the plugin will now require a title to be configured for each feed, as [per the documentation](https://www.gatsbyjs.org/packages/gatsby-plugin-feed/).

@sidharthachatterjee approved and merged the PR.

That bit in the readme was only added 5 months ago by @DSchau:
https://github.com/gatsbyjs/gatsby/commit/626cab461640bd6598806f3ec5b98e196f83dc9a

Dustin wrote at the time in the commit and PR description:

> I propose that we make this a minor version bump for this plugin, once merged.

@secretfader and @pieh approved the changes and Michal merged the PR:
https://github.com/gatsbyjs/gatsby/pull/12085

I think it goes without saying that this is a breaking change, and would have been whether or not the docs had said this all along, but they didn't. I cannot see how there is any way this should have happened under a [PATCH version](https://semver.org/).

Further, my RSS feed had a title the whole time anyway.

Screen Shot 2019-08-26 at 160054

Here is my config at the time the breaking change was introduced:
https://github.com/rdela/rdela.com/blob/755f3c98ab727f647377941b84dcc35a5b020e29/gatsby-config.js#L110-L163

Please try to respect SemVer. I know how to track down errors like these, but I am sure there are people out there with broken builds scratching their heads who will not drop by to weigh in or raise their voice in the issues, even if they are able to fix their builds. Gatsby is difficult and complicated enough to use without surprise breakage like this.

I appreciate all of your hard work on Gatsby and that there were other bugs being fixed as these breaking changes were introduced. Yet none of those bugs effected me, but the breaking change in gatsby-plugin-feed 2.3.8 did. I am able to fix the build by adding the title field in my config, but I still do not understand why I should have to add redundant information that exists in `siteMetadata` in the same file to replicate the result I was already getting without it.

Please let me know if I have misunderstood something and if you spot any problems with my site, issues and PRs are very welcome, as always.

eyalroth:
Fixes #17100.

### DIFF
--- a/packages/gatsby-plugin-feed/src/__tests__/__snapshots__/gatsby-node.js.snap
+++ b/packages/gatsby-plugin-feed/src/__tests__/__snapshots__/gatsby-node.js.snap
@@ -9,5 +9,3 @@ exports[`Test plugin feed default settings work properly 1`] = `"<?xml version=\
 exports[`Test plugin feed options validation throws when invalid plugin options 1`] = `[Error: [Config Validation]: "output" is required]`;
 
 exports[`Test plugin feed options validation throws when invalid plugin options 2`] = `[Error: [Config Validation]: "query" is required]`;
-
-exports[`Test plugin feed options validation throws when invalid plugin options 3`] = `[Error: [Config Validation]: "title" is required]`;
--- a/packages/gatsby-plugin-feed/src/__tests__/gatsby-node.js
+++ b/packages/gatsby-plugin-feed/src/__tests__/gatsby-node.js
@@ -47,6 +47,25 @@ describe(`Test plugin feed`, async () => {
       )
     })
 
+    it(`warns when individual feed does not have title`, async () => {
+      const options = {
+        feeds: [
+          {
+            output: `rss.xml`,
+            query: `{}`,
+            serialize: () => {},
+          },
+        ],
+      }
+
+      const [reporter] = await setup(options)
+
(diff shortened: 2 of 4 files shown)


sidharthachatterjee:
Published in `gatsby-plugin-feed@2.3.11`

PR #17111, opened 2019-08-27
fix(gatsby): prevent unmount wrapPageElement while page-data loading in dev

### LINKED ISSUE #16489: wrapPageElement still remounts everything on certain pages
So I am using wrapPageElement to wrap pages with a layout that I don't want remounting when transitioning from page to page. What's weird is that when I start on a page and go to any page other than home for the first time everything remounts. Then if I go back to the first page that I already accessed it transitions as it should. Plus, if I start on any page other than the homepage then go to the homepage it transitions the way it should. All of my pages are being created during build with createPage. Any ideas??

ValeraS:
In rare cases, when you go to a page, its data may not be loaded yet, which causes unmounting of `wrapPageComponent`.

Closes #16489

### DIFF
--- a/packages/gatsby/cache-dir/dev-loader.js
+++ b/packages/gatsby/cache-dir/dev-loader.js
@@ -10,10 +10,11 @@ class DevLoader extends BaseLoader {
 
   loadPage(pagePath) {
     const realPath = cleanPath(pagePath)
-    return super.loadPage(realPath).then(result => {
-      require(`./socketIo`).getPageData(realPath)
-      return result
-    })
+    return super.loadPage(realPath).then(result =>
+      require(`./socketIo`)
+        .getPageData(realPath)
+        .then(() => result)
+    )
   }
 
   loadPageDataJson(rawPath) {
--- a/packages/gatsby/cache-dir/json-store.js
+++ b/packages/gatsby/cache-dir/json-store.js
@@ -72,12 +72,6 @@ class JSONStore extends React.Component {
     // - page data for path changed
     // - static query results changed
 
-    // Always return false if we're missing resources.
-    const data = nextState.pageQueryData[normalizePagePath(nextState.path)]
-    if (!data) {
-      return false
-    }
-
     return (
       this.props.location !== nextProps.location ||
       this.state.path !== nextState.path ||
--- a/packages/gatsby/cache-dir/json-store.js
+++ b/packages/gatsby/cache-dir/json-store.js
@@ -72,6 +72,12 @@ class JSONStore extends React.Component {
     // - page data for path changed
     // - static query results changed
 
+    // Always return false if we're missing resources.
+    const data = nextState.pageQueryData[normalizePagePath(nextState.path)]
+    if (!data) {
+      return false
+    }
+
     return (
       this.props.location !== nextProps.location ||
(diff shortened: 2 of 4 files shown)


packages/gatsby/cache-dir/json-store.js
```diff
@@ -72,6 +72,12 @@ class JSONStore extends React.Component {
     // - page data for path changed
     // - static query results changed
 
+    // Always return false if we're missing resources.
+    const data = nextState.pageQueryData[normalizePagePath(nextState.path)]
+    if (!data) {
+      return false
```
ValeraS: With this approach, e2e tests, which wait for `onRouteUpdate` and check focused element, randomly fail, because of the route changed but page-data did not load yet.

ValeraS: Or maybe instead of this, change render function like this:

```js
  render() {
    const data = this.state.pageQueryData[getPathFromProps(this.props)]
	let { pageResources, ...otherProps } = this.props
    // eslint-disable-next-line
    if (!data) {
	  pageResources = Object.assign({}, pageResources, { component: () => <div /> })
    }
    return (
      <StaticQueryContext.Provider value={this.state.staticQueryData}>
        <PageRenderer pageResources={pageResources} {...otherProps} {...data} />
      </StaticQueryContext.Provider>
    )
  }
```

pieh:
This is interesting, it seems to me we should prevent this earlier. We do have `<EnsureResources>` ( https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/cache-dir/ensure-resources.js ) which does ensure we don't switch component (or emit `onRouteUpdate` events) until page is actually ready - see https://github.com/gatsbyjs/gatsby/blob/9468280bb28144c5283b06144631bc1a3ec1b745/packages/gatsby/cache-dir/root.js#L61-L63

this works pretty well in production builds, but seems like it doesn't cover development

I think there is regression happening there - `loader` used there to make sure resources are loaded doesn't seem to handle data, looking now at 
https://github.com/gatsbyjs/gatsby/blob/9468280bb28144c5283b06144631bc1a3ec1b745/packages/gatsby/cache-dir/dev-loader.js#L11-L17

it seems like we are not waiting for getting data from websocket - it should chain then there - otherwise it creates orphaned promise (in `require(`./socketIo`).getPageData(realPath)` code, that will resolve sometime in future, but `loadPage` itself can resolve actually faster

packages/gatsby/cache-dir/dev-loader.js
```diff
@@ -10,10 +10,11 @@ class DevLoader extends BaseLoader {
 
   loadPage(pagePath) {
     const realPath = cleanPath(pagePath)
-    return super.loadPage(realPath).then(result => {
-      require(`./socketIo`).getPageData(realPath)
-      return result
-    })
+    return super.loadPage(realPath).then(result =>
+      require(`./socketIo`)
+        .getPageData(realPath)
```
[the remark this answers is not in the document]
pieh: That's good catch. Yeah websocket-manager should sent some message if page doesn't exist so the promise can be resolved (or maybe rejected - would need to to check what happens in production builds)

pieh:
Published in `gatsby@2.15.23`

PR #17118, opened 2019-08-27
feat(gatsby): Add support for urls in assets

Eddie-CooRo:
## Description

This PR adds support for having cdn links in assets (added to assets using `dynamic-cdn-webpack-plugin`) by stopping adding "/" and __PATH_PREFIX__ to the start of links and urls which are starting with `http://` or `https://`.

### DIFF
--- a/packages/gatsby/cache-dir/static-entry.js
+++ b/packages/gatsby/cache-dir/static-entry.js
@@ -256,15 +256,24 @@ export default (pagePath, callback) => {
         return null
       }
 
+      const getAssetType = name =>
+        name && (name.startsWith(`https://`) || name.startsWith(`http://`))
+          ? `url`
+          : `local`
+
       chunks = chunks.map(chunk => {
         if (chunk === `/`) {
           return null
         }
-        return { rel: `preload`, name: chunk }
+        return { rel: `preload`, name: chunk, type: getAssetType(chunk) }
       })
 
       namedChunkGroups[s].assets.forEach(asset =>
-        chunks.push({ rel: `preload`, name: asset })
+        chunks.push({
+          rel: `preload`,
+          name: asset,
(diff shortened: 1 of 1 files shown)


wardpeet:
I don't really like adding support for things like `dynamic-cdn-webpack-plugin`. It's mostly bad for performance as you loose on a lot of benefits that webpack is providing like tree-shaking, code splitting & bundle optimisations.

You can argue that dependencies are mostly in the browser cache and that might be true for react, react-dom but not for most node_modules. You can achieve the same thing by adding an externals option to webpack and using our [SSR api onRenderBody](https://www.gatsbyjs.org/docs/ssr-apis/#onRenderBody)

Eddie-CooRo:
About losing things like tree-shaking and code-splitting, I totally agree with you; But as you stated, React & ReactDOM won't benefit from these things at all. And with `dynamic-cdn-webpack-plugin` we can turn on cdn loading only for these modules. It would be nice if we could just add the plugin to the webpack config and go with it. And to me, it seems pretty logical to not prefixing the absolute paths while we are rendering the HTML.
But about the CDN use-case (Although I can't find any other use-case at the moment), It's OK for me. I can achieve the same behavior using the life-cycles gatsby currently provides.

Thank you for your time.

wardpeet:
More on this topic: Browsers are moving to double-key caching strategy for resources. Doube-key caching means that resources are cached with the origin as the cache key. If you have `react.js` on a global CDN which is cached on site A. The visitor loads your page, the user will still need to download `react.js`. So the benefit of downloading from a global CDN has nothing to do with performance anymore, especially in Gatsby land where everything is static.
This is already implemented in Safari, and it's coming to chrome soon (I don't know the version by heart). 

The change you’ve proposed is not going to be accepted because of the reasons stated above.

PR #17120, opened 2019-08-27
feat(gatsby-graphiql-explorer): Implement CodeExporter

### LINKED ISSUE #14476: Generate page code from graphiql
An idea from @sgrove / @pieh 

OneGraph generates code for you from GraphQL queries — we could do the same thing for Gatsby as a very common workflow is going to GraphiQL to sketch out the query for a page and then using that query for the new page. We could add a button that generates code like the following:

```jsx
import React from "react"
import { graphql } from "gatsby"

export default ({ data }) => <pre>{JSON.stringify(data, null, 4)}</pre>

export const query graphql`
INSERT GRAPHQL QUERY HERE
`
```

Initially we could just give them code to copy paste. There's been conversations about adding an API to Gatsby so we can write out to the file from tooling in the browser so eventually we might be able to just ask them what path they want the new page at and we'd directly create the file there and give them a link to open.

herecydev:
## Description

Add's `graphiql-code-exporter` into `gatsby-graphiql-explorer` allowing the user to choose between 3 graphql snippets which can be copied/pasted.

## Examples

Page query template
image

useStaticQuery Hook
image

StaticQuery Component
image

## Related Issues

Fixes #14476

### DIFF
--- a/packages/gatsby-graphiql-explorer/src/app/app.js
+++ b/packages/gatsby-graphiql-explorer/src/app/app.js
@@ -4,11 +4,14 @@ import ReactDOM from "react-dom"
 import GraphiQL from "graphiql"
 import GraphiQLExplorer from "graphiql-explorer"
 import { getIntrospectionQuery, buildClientSchema, parse } from "graphql"
+import CodeExporter from "graphiql-code-exporter"
+import snippets from "./snippets"
 
 import "whatwg-fetch"
 
 import "graphiql/graphiql.css"
 import "./app.css"
+import "graphiql-code-exporter/CodeExporter.css"
 
 const parameters = {}
 window.location.search
@@ -142,11 +145,22 @@ const storedExplorerPaneState =
     ? window.localStorage.getItem(`graphiql:graphiqlExplorerOpen`) !== `false`
     : true
 
+const storedCodeExporterPaneState =
+  typeof parameters.codeExporterIsOpen !== `undefined`
+    ? parameters.codeExporterIsOpen === `false`
+      ? false
+      : true
+    : window.localStorage
+    ? window.localStorage.getItem(`graphiql:graphiqlCodeExporterIsOpen`) !==
+      `false`
+    : true
+
 class App extends React.Component {
   state = {
     schema: null,
     query: DEFAULT_QUERY,
     explorerIsOpen: storedExplorerPaneState,
+    codeExporterIsOpen: storedCodeExporterPaneState,
   }
 
   componentDidMount() {
@@ -269,8 +283,29 @@ class App extends React.Component {
     this.setState({ explorerIsOpen: newExplorerIsOpen })
   }
 
+  _handleToggleExporter = () => {
+    const newCodeExporterIsOpen = !this.state.codeExporterIsOpen
+    if (window.localStorage) {
+      window.localStorage.setItem(
+        `graphiql:graphiqlCodeExporterOpen`,
+        newCodeExporterIsOpen
+      )
+    }
+    parameters.codeExporterIsOpen = newCodeExporterIsOpen
+    updateURL()
+    this.setState({ codeExporterIsOpen: newCodeExporterIsOpen })
+  }
+
   render() {
-    const { query, schema } = this.state
+    const { query, schema, codeExporterIsOpen } = this.state
+    const codeExporter = codeExporterIsOpen ? (
+      <CodeExporter
+        hideCodeExporter={this._handleToggleExporter}
+        snippets={snippets}
+        query={query}
+        codeMirrorTheme="default"
+      />
+    ) : null
 
     return (
       <React.Fragment>
@@ -309,8 +344,14 @@ class App extends React.Component {
               label="Explorer"
               title="Toggle Explorer"
             />
+            <GraphiQL.Button
+              onClick={this._handleToggleExporter}
+              label="Code Exporter"
+              title="Toggle Code Exporter"
+            />
           </GraphiQL.Toolbar>
         </GraphiQL>
+        {codeExporter}
       </React.Fragment>
     )
   }
--- a/packages/gatsby-graphiql-explorer/src/app/app.js
+++ b/packages/gatsby-graphiql-explorer/src/app/app.js
@@ -151,8 +151,8 @@ const storedCodeExporterPaneState =
       ? false
       : true
     : window.localStorage
-    ? window.localStorage.getItem(`graphiql:graphiqlCodeExporterOpen`) !==
-      `false`
+    ? window.localStorage.getItem(`graphiql:graphiqlCodeExporterOpen`) ===
+      `true`
     : false
 
 class App extends React.Component {
--- a/packages/gatsby-graphiql-explorer/src/app/app.js
+++ b/packages/gatsby-graphiql-explorer/src/app/app.js
@@ -153,7 +153,7 @@ const storedCodeExporterPaneState =
     : window.localStorage
     ? window.localStorage.getItem(`graphiql:graphiqlCodeExporterIsOpen`) !==
       `false`
-    : true
+    : false
 
 class App extends React.Component {
   state = {
--- a/packages/gatsby-graphiql-explorer/src/app/app.js
+++ b/packages/gatsby-graphiql-explorer/src/app/app.js
@@ -151,7 +151,7 @@ const storedCodeExporterPaneState =
       ? false
       : true
     : window.localStorage
-    ? window.localStorage.getItem(`graphiql:graphiqlCodeExporterIsOpen`) !==
+    ? window.localStorage.getItem(`graphiql:graphiqlCodeExporterOpen`) !==
       `false`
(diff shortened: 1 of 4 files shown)


pieh:
That's very cool!

What do you think about adding 4th snippet for creating pages inside `gatsby-node.js`?

pieh:
Another note I would like to make is about query names (which are `MyQuery` in example screenshots): Gatsby requires query names (if provided) to be unique. I can imagine workflows when you scaffold multiple pages using GraphiQL explorer -> code exporter that you could end up with multiple queries with same name.

Gatsby also allow users to skip query names (and those will be generated under the hood) - so maybe we could try to strip actual query names from code snippets? It wouldn't be prettiest solution, but doing some regular expression replacement could work here?

Something along the lines of:
```js
getQuery(arg).replace(/query\s.+{/gmi, `query {`)
```
which I did quick very quick check that it could work

KyleAMathews:
Amazing! 🙏

packages/gatsby-graphiql-explorer/src/app/app.js
```diff
@@ -142,11 +145,22 @@ const storedExplorerPaneState =
...
 
+const storedCodeExporterPaneState =
+  typeof parameters.codeExporterIsOpen !== `undefined`
+    ? parameters.codeExporterIsOpen === `false`
+      ? false
+      : true
+    : window.localStorage
+    ? window.localStorage.getItem(`graphiql:graphiqlCodeExporterIsOpen`) !==
+      `false`
+    : true
```
pieh: should we default to `false` here? having 2 panes open (explorer and exporter) on first visit might hide too much of core graphiql in viewport?

packages/gatsby-graphiql-explorer/src/app/snippets.js
```diff
@@ -0,0 +1,59 @@
+const getQuery = arg => {
+  const { operationDataList } = arg
+  const { query } = operationDataList[0]
```
pieh: I think it would be useful to do some sanity check here - if there are no queries there will "cannot read query of undefined" error happening

herecydev: I will check this again, I felt like it's handled but if not I'll put something in place

packages/gatsby-graphiql-explorer/src/app/snippets.js
```diff
@@ -0,0 +1,59 @@
...
+  name: `StaticQuery hook`,
+  language: `JavaScript`,
+  codeMirrorMode: `jsx`,
+  options: [],
+  generate: arg => `import React from "react"
+import { useStaticQuery, graphql } from "gatsby"
+
+export default () => {
+const data = useStaticQuery(graphql\`
+  ${getQuery(arg)}
```
herecydev: Minor formatting issue is that the first line gets spaces added, but following lines don't. Is it worth padding the lines with a couple of spaces to keep things nicely formatted?

packages/gatsby-graphiql-explorer/src/app/snippets.js
```diff
@@ -0,0 +1,59 @@
...
+export default ({ data }) => <pre>{JSON.stringify(data, null, 4)}</pre>
+
+export const query = graphql\`
+${getQuery(arg)}
+\`
+`,
+}
+
+const staticHook = {
+  name: `StaticQuery hook`,
```
herecydev: Are the query names appropriate?

packages/gatsby-graphiql-explorer/src/app/snippets.js
pieh: Oh, it might be handled by code-exporter itself maybe (not sure) - let me know what happens when you have completely empty query editor ;)

herecydev:
@pieh I've made some further changes if you can take a look. With regards to how errors in the query are handled. The library handles it before running the snippets:

image

packages/gatsby-graphiql-explorer/src/app/snippets.js
pieh: This is handled by code-exporter internally, so I'll mark this note as resolved

packages/gatsby-graphiql-explorer/src/app/snippets.js
pieh: I would use `useStaticQuery`, but I think having some note on what it is (and possibly link to documentation) would be great (I'm not sure if code-exporter has support to provide notes that are not part of code snippet?)

pieh:
I committed some minor fixes and expanded a bit `createPages` snippet to:
 - be able to handle potential query errors
 - show how to actually create pages from results of query - this is tough, because this is pretty abstract and that part isn't reflecting actual query, but rather is blueprint for how this need to be done. Ideally that part could also be autogenerated, but it would require to set some params in code-exporter (which part should be used for `path` etc and it's much more complicated). I think this should be good to go now and can be iterated on in future if anyone is interested. 

Screenshot 2019-09-02 at 12 57 00

Let me know what you think of my changes @herecydev

herecydev:
Changes are good, I specifically avoided adding code after the query simply because you had to assume a lot, for example that you were query would return `allData` and that would have `nodes` etc...

I dug into the CodeExporter a little more and it provides a huge tree of the query, so my latest commit is my attempt at producing a much deeper code snippet:

image

(and with comments)

image


A few things to note:

- Checks for the presence of `nodes` on the query and suggests alternatives if it can't be found
- Uses the node type to work out the variable name in `forEach`, i.e. allFoo will become foo
- Checks for a field `slug` and uses that in the path
- Checks for a field `id` and will add that to the context (as this is a good candidate)
- Path is generated with the node type in mind, i.e. allFoo will become "./src/templates/foo.js"
- Option to toggle comments

herecydev:
I suppose one other thing I would like to change is that although using `export default ( ) =>` is nice and easy. I think it's a slippery slope and would always recommend explicitly naming your component otherwise you end up with a bunch of `Unknown` components in your React Tree where it has no information to work with.

What do you think to adding an option checkbox to toggle that behavior?

pieh:
> Changes are good, I specifically avoided adding code after the query simply because you had to assume a lot, for example that you were query would return `allData` and that would have `nodes` etc...

Yeah, I do agree that all the snippet changes assumes much (tho, I tried to make it clear this is sample code), alternatively instead of adding code we could link to documentation for creating pages.

> I dug into the CodeExporter a little more and it provides a huge tree of the query, so my latest few commits are my attempt at producing a much deeper code snippet:

Nice! `huge tree of the query` is called Abstract Syntax Tree (AST) ;) There is handy tool for inspecting them - https://astexplorer.net/#/gist/0810c322d06bf3d9ea7c9fb14d82fb9f/f19bcf0b1eb515f3535dd21d2f61a6724317a1dc (there all kinds of ASTs - so be sure to use `GraphQL` there if you end up using that)

Honestly I didn't want to this initially, because it complicates handler a lot. And it delays merging of this PR. Maybe we should drop `createPages` snippet from this PR, so we could merge initial implementation for page queries and both static queries variant and continue iterating on `createPages` in next PR as follow up PR?

> I suppose one other thing I would like to change is that although using `export default ( ) =>` is nice and easy. I think it's a slippery slope and would always recommend explicitly naming your component otherwise you end up with a bunch of `Unknown` components in your React Tree where it has no information to work with.
> 
> What do you think to adding an option checkbox to toggle that behavior?

I think we should avoid options when possible and try to generate snippets that would use as much of good practices as possible - so to me it makes perfect sense to use named Component by default.

packages/gatsby-graphiql-explorer/src/app/snippets.js
pieh: Just checked, and it doesn't seem like we can provide notes with current version of `graphiql-code-exporter` so my thought is irrelevant. Let's keep your name - we can always change it later if we need to.

herecydev:
> Honestly I didn't want to this initially, because it complicates handler a lot

I fully expected this answer, there's a lot of value to be had but there's quite some complication as well.

Alright let me:

1. drop the `createPages` snippet for now, we can revisit that later
2. Add named components, there's really nothing to go on so I'll call it `ComponentName` unless you have a better suggestion

Can you confirm you're happy with the query names like `useStaticQuery hook`. Are they obvious how they differentiate and what they do?

pieh:
> 2\. Add named components, there's really nothing to go on so I'll call it `ComponentName` unless you have a better suggestion

I don't have better suggestion. 👍 for it.

> Can you confirm you're happy with the query names like `useStaticQuery hook`. Are they obvious how they differentiate and what they do?

I can't really come up with name that's objectively better, so 👍 for it (not a blocker)

pieh:
One note on `ComponentName` and `useStaticQuery` - we will likely hear from broader community of users when they start using them and potentially propose changes to those - will make sure to /cc you there ;)

pieh:
Published `gatsby@2.15.3` / `gatsby-graphiql-explorer@0.2.10`

Issue #17126, opened 2019-08-28
Adding large numbers of redirects causes significant createPages slowdown


me4502:
## Description

When creating a large number of redirects (around 17k), the createPages function goes from 8 seconds up to 180 seconds. This is significant, as there are 60k pages which are completed in 22x less time.

The slowdown appears to be caused by `if (!state.some(redirect => _.isEqual(redirect, action.payload)))` in the redirects reducer.

Ideally this would be replaced with a Map of fromPath to the object for constant time lookup, however as this is used by plugins that would be a breaking change.

Funnily enough, replacing that entire block with:

```
if (!new Set(state.map(redirect => redirect.fromPath)).has(action.payload.fromPath)) {
  // Add redirect only if it wasn't yet added to prevent duplicates
  state.push(action.payload);
  return state; 
}
```

is twice as fast (although this would be /awful/ on memory).

I was looking into maybe a way to proxy the value of the object (to retain compatibility) using ES6 proxies, however that feels fairly hacky.

### Steps to reproduce

1. Create a lot of redirects

### Expected result

Redirects should not introduce significantly more overhead than creating an actual page

### Actual result

Redirects are 68x slower than creating pages.

### Environment

  System:
    OS: macOS 10.15
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 11.15.0 - ~/.nvm/versions/node/v11.15.0/bin/node
    Yarn: 1.17.3 - ~/.nvm/versions/node/v11.15.0/bin/yarn
    npm: 6.7.0 - ~/.nvm/versions/node/v11.15.0/bin/npm
  Languages:
    Python: 2.7.16 - /usr/local/bin/python
  Browsers:
    Chrome: 76.0.3809.132
    Safari: 13.0

me4502:
Replacing the entire function with

```
const fromPaths = new Set();

module.exports = (state = [], action) => {
  switch (action.type) {
    case `CREATE_REDIRECT`:
      {
        if (!fromPaths.has(action.payload.fromPath)) {
          // Add redirect only if it wasn't yet added to prevent duplicates
          state.push(action.payload);
          fromPaths.add(action.payload.fromPath);
          return state;
        }

        return state;
      }

    default:
      return state;
  }
};
```

[...]

PR #17128, opened 2019-08-28
chore(gatsby): speed up redirect creation

me4502:
## Description

This PR prevents redirects massively slowing down the build. The tests needed to be changed so that they don't conflict on the fromPath, and that was a better change than exporting fromPaths.

## Related Issues

https://github.com/gatsbyjs/gatsby/issues/17126

### DIFF
--- a/packages/gatsby/src/bootstrap/redirects-writer.js
+++ b/packages/gatsby/src/bootstrap/redirects-writer.js
@@ -12,7 +12,9 @@ const writeRedirects = async () => {
   let { program, redirects } = store.getState()
 
   // Filter for redirects that are meant for the browser.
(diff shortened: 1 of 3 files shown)


me4502:
Ah, nice. Thanks - that was my original fix idea however I was unsure whether it was safe to modify the type of state due to it being used in plugins.

wardpeet:
Redux state is a "private" api and on your own risk to use it so no harm done.

KyleAMathews:
@wardpeet -- we haven't considered redux state private. We need to deprecate it and replace it with data access APIs.

Could you revert this to the previous way @me4502 had it?

wardpeet:
I'll do that. Thanks for telling me.

PR #17129, opened 2019-08-28
remove broken Twitter link to example

dschlabach:
Link to Kyle Gill's example on Twitter is broken


wardpeet:
Thank you, good catch! 🥇

Issue #16502, opened 2019-08-09
Adding HTML to MDX crashes hot reloader


whoisryosuke:
## Description

I'm trying to create documentation using Gatsby and MDX. I'm taking the HTML examples from the Semantic UI docs and inserting them directly into MDX files.

MDX crashes when adding a lot of HTML to a MDX file and produces the following error:

`TypeError: Cannot read property 'page' of undefined`

The error is not very descriptive of what's actually breaking it (the HTML), and it becomes impossible to debug the issue directly.

### Steps to reproduce

Using project with MDX. Basic setup grabs MDX files, creates slug based off frontmatter, and runs through template to create a page layout.

I create an MDX file with pure markdown, it works fine. I add a large amount of HTML snippets to it, HTML shows as a blockquote (it's snippets wrapped in a container div, the container gets rendered then MDX assumes the inner HTML is a blockquote -- a whole other issue 😅).

I then edit the MDX file (removing a few lines of HTML  - the container specifically, as well as un-indenting the HTML). Page doesn't refresh, when I hard refresh it shows the following error message above.

If I remove a lot of the HTML, I get the page. Sometimes requires a restart of the dev server though.

Here's the HTML in question, it's from the Semantic UI docs

```html
<h2 class="ui dividing header">Introduction</h2>
<div class="highlighted example" data-class="grid">
    <h4 class="ui header">Grids</h4>
    <p>A grid is a structure with a <a href="http://99designs.com/designer-blog/2013/03/21/history-of-the-grid-part-1/" target="_blank">long history</a> used to align negative space in designs.</p>
    <p>Using a grid makes content appear to flow more naturally on your page.</p>
    <div class="ui ignored toggle animation checkbox">
    <input type="checkbox" name="animation" />
    <label>Toggle Animation</label>
    </div>
    <div class="ui ignored hidden divider"></div>
    <div class="ui grid">
    <div class="four wide column"></div>
    <div class="four wide column"></div>
    <div class="four wide column"></div>
    <div class="four wide column"></div>
    </div>
</div>

<div class="highlighted example" data-use-content="true" data-class="column">
    <h4 class="ui header">Columns</h4>
    <p>Grids divide horizontal space into indivisible units called "columns". All columns in a grid must specify their width as proportion of the total available row width.</p>
    <p>All grid systems choose an arbitrary column count to allow per row. Semantic's default theme uses <b>16 columns</b>.</p>

[...]

Issue #16107, opened 2019-07-26
[gatsby-plugin-mdx] Don't crash `develop` server on broken Markdown


karlhorky:
## Summary

Broken Markdown will cause `gatsby-plugin-mdx` to crash the `gatsby develop` server and prevent starting again. This issue is a request to change that behavior so that:

1. MVP: the `develop` server doesn't crash (maybe the last working result is cached?)
2. Ideally, the error appears in the browser.

### Basic example

The following MDX content will crash the `develop` server with the error below.

```mdx
asdf</p>
```
<details>
<summary><b>Show Full Error Message</b></summary>

```
ERROR

UNHANDLED REJECTION unknown: Expected corresponding JSX closing tag for <MDXLayout> (7:15)

   5 |       {...props}
   6 |       components={components}>
>  7 | <p>{`asdf`}</p></p>
     |                ^
   8 |     </MDXLayout>
   9 |   )
  10 | };src/pages/c/index.mdx: unknown: Expected corresponding JSX closing tag for <MDXLayout> (7:15)

   5 |       {...props}
   6 |       components={components}>
>  7 | <p>{`asdf`}</p></p>
     |                ^
   8 |     </MDXLayout>
   9 |   )
  10 | };



  SyntaxError: unknown: Expected corresponding JSX closing tag for <MDXLayout> (7:15)
     5 |       {...props}
     6 |       components={components}>
  >  7 | <p>{`asdf`}</p></p>
       |                ^
     8 |     </MDXLayout>
     9 |   )
    10 | };src/pages/c/index.mdx: unknown: Expected corresponding JSX closing tag for   <MDXLayout> (7:15)
     5 |       {...props}
     6 |       components={components}>
  >  7 | <p>{`asdf`}</p></p>
       |                ^
     8 |     </MDXLayout>
     9 |   )
    10 | };

  - index.js:6325 Object.raise
    [@babel]/parser/lib/index.js:6325:17

  - index.js:3563 Object.jsxParseElementAt
    [@babel]/parser/lib/index.js:3563:16

  - index.js:3589 Object.jsxParseElement
    [@babel]/parser/lib/index.js:3589:17

  - index.js:3596 Object.parseExprAtom
    [@babel]/parser/lib/index.js:3596:19

  - index.js:8412 Object.parseExprSubscripts
    [@babel]/parser/lib/index.js:8412:23

  - index.js:8392 Object.parseMaybeUnary
[...]

PR #17135, opened 2019-08-28
fix(gatsby-plugin-mdx): prevent crash when Markdown syntax error

tureki:
## Description
This PR will catch the error from MDX compile. Avoid the develop server crash and show the error message in the browser.

## Related Issues
Fixes https://github.com/gatsbyjs/gatsby/issues/16107
Fixes https://github.com/gatsbyjs/gatsby/issues/16502

### Before
before

### After
after


wardpeet:
Thanks, this seems to work as expected! I'm not sure why we aren't wrapping the whole function in try-catch.

cc @gatsbyjs/themes

PR #17137, opened 2019-08-28
gatsby-plugin-preact: remove 'preact/debug'

sndrs:
## Description

Preact is not actually used by the plugin in development: https://github.com/gatsbyjs/gatsby/blob/193f4cd3728cc6af3e55640fb712747e1b369c7e/packages/gatsby-plugin-preact/src/gatsby-node.js#L2-L4 

This PR updates the docs to make that clear.

The current version of `preact/debug` is also not compatible with latest react devtools: https://github.com/preactjs/preact/issues/1885, throwing errors in the console. 

Since Preact is not being used anyway, this PR also removes import of `preact/debug`.

### DIFF
--- a/packages/gatsby-plugin-preact/README.md
+++ b/packages/gatsby-plugin-preact/README.md
@@ -18,3 +18,9 @@ React.
 // In your gatsby-config.js
 plugins: [`gatsby-plugin-preact`]
 ```
+
+## Usage in a development environment
+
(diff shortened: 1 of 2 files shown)


sndrs:
@wardpeet just out of interest, what's the benefit of requiring `preact/debug` in envs where preact isn't being bundled? does it still have an effect?

wardpeet:
The idea is to move develop to preactjs too when the time is right :joy:

PR #17140, opened 2019-08-28
fix(www): large output of guess-js

wardpeet:
## Description
Guess-js is making our webpack-runtime super big adds 2mb to it. We need to switch to guess AOT mode but for now this works.

Thanks for the quickfix @mgechev!


mgechev:
This should be accurate enough! :)

PR #17142, opened 2019-08-28
Chore(CLEANING): Cleaning all files and changing from class based to …

Bvlktech:
Cleaning up all files

## Description

Cleaning up all the files in the www/*.js folder and making all class based components into function based ex ->

class something extends component{ // ext... } 

will be changed to

const something = () => { // ext... }


wardpeet:
Hey, thank you for taking the time to do a "cleanup" our website's code. 
I don't feel like this is a significant enough change to go forward with. Here are a few notes why:
- Functional components won't give any benefits over class components.  Gzip will make bytes size a marginal factor to be valuable here. We probably want to take this one while we're rebranding gatsbyjs.org instead of converting everything right now.
- Ordering imports will be a lost effort. We don't enforce import ordering as it makes it a little bit harder to contribute. There is no benefit from doing so.
- Adding comments to blog posts (discus) -- this might look like a good idea but the time and effort it costs to moderate these is not worth the extra engagement to the community.

We love you taking the time and effort to fix this. As of now, i'm closing the PR for the reasons stated above. Thank you for understanding.

Bvlktech:
No problem @wardpeet i just did this as something fun to do although the discus thing was a suggestion that's why i just left it commented out in the code rather than actually adding it in ill delete this branch in a few and look at some actual bugs that i can work on / hopefully fix thanks for taking a look at it tho

PR #17200, opened 2019-08-29
docs:UserWrapper is not defined

scottwater:
In the quickstart recipe on styled-components (https://www.gatsbyjs.org/docs/recipes/#using-styled-components), the component UserWrapper is not defined. My suggested change is to just use a Fragment since UserWrapper did not have any true functionality (as far as I could tell).

The alternative would be something like the following, but I do not see that being any more helpful.

```jsx
const UserWrapper = ({ children }) => {
  return <div>{children}</div>
}
```

### DIFF
--- a/docs/docs/recipes.md
+++ b/docs/docs/recipes.md
@@ -314,10 +314,10 @@ const Username = styled.h2`
 `
 
 const User = props => (
-  <UserWrapper>
+  <>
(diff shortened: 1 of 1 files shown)


wardpeet:
This looks good, thank you!. I meant to track down where UserWrapper came from but it came from nowhere so <> & </> seems right.

PR #17213, opened 2019-08-30
Show highlight-* comments in code samples in docs

dkthehuman:
### Problem

In the [Line highlighting section](https://www.gatsbyjs.org/contributing/gatsby-style-guide/#code-formatting-line-highlighting) of the Gatsby Style Guide, the code samples don't actually display the comments (e.g. `// highlight-line`). As a result, reading that section of the docs is quite confusing because the comments being referenced don't actually appear in the code sample.

Based on the assumption that the comment is being processed by `gatsby-remark-prismjs`, I tried various ways to make the comment show up (escaping the slashes, putting the comment inside a <span> element, etc.), but the only approach that  worked was replacing the hyphen in the keywords like "highlight-line" with another dash character (in this case, the figure dash (U+2012)).

The downside of this approach is that if someone were to copy and paste the code sample, it wouldn't work for them since "highlight‒line" (with a figure dash) wouldn't trigger `gatsby-remark-prismjs`, but I think that this is a better solution than not having the comments show up at all in the code snippet.

Open to other suggestions if there's a better way around this!

### Before

Screen Shot 2019-08-29 at 9 41 00 PM

### After

Screen Shot 2019-08-29 at 9 41 18 PM

### DIFF
--- a/docs/contributing/gatsby-style-guide.md
+++ b/docs/contributing/gatsby-style-guide.md
@@ -273,7 +273,7 @@ You may also choose to include line highlighting in your code snippets, using th
 ```javascript:title=gatsby-config.js
 module.exports = {
 	siteMetadata: {
-		title: `GatsbyJS`, // highlight-line
+		title: `GatsbyJS`, // highlight‒line
 		siteUrl: `https://www.gatsbyjs.org`,
 	},
 }
@@ -296,7 +296,7 @@ module.exports = {
 module.exports = {
 	siteMetadata: {
 		title: `GatsbyJS`,
-		// highlight-next-line
+		// highlight‒next‒line
 		siteUrl: `https://www.gatsbyjs.org`,
 	},
 }
@@ -318,12 +318,12 @@ module.exports = {
 ````
 ```javascript:title=gatsby-config.js
 module.exports = {
-	// highlight-start
+	// highlight‒start
 	siteMetadata: {
 		title: `GatsbyJS`,
 		siteUrl: `https://www.gatsbyjs.org`,
 	},
-	// highlight-end
+	// highlight‒end
 }
 ```
 ````


dkthehuman:
Aw, I got really excited by your suggestion because that would've been a much better solution, but `gatsby-remark-prismjs` still eats the comment even when the language is set to `none` as follows:

`````
````none
```javascript:title=gatsby-config.js
module.exports = {
	siteMetadata: {
		title: `GatsbyJS`, // highlight-line
		siteUrl: `https://www.gatsbyjs.org`,
	},
}
```
````
`````

Still shows up as:

Screen Shot 2019-08-30 at 9 04 47 PM

**Comment by dkthehuman** [comment]
I thought that my results here indicated that there's a bug in `gatsby-remark-prismjs` so I set out to create a test site that would show that `// highlight-line` gets processed even when the language is not set. But I found that it works as intended (syntax-highlighting is disabled when the language is not set or set to `none`):

Netlify demo: https://goofy-hermann-1dc856.netlify.com/hello-world/
Test repo: https://github.com/dkthehuman/gatsby-prismjs-highlight-demo

Looks like something else is eating those `highlight-*` comments?

**Comment by pieh** [comment]
That might be related to .org using `gatsby-plugin-mdx` and not `gatsby-transformer-remark` - worth checking at least (to eliminate as much variables as possible)

**Comment by dkthehuman** [comment]
@pieh Gotcha, thanks for the pointer. When is `gatsby-transformer-remark` used instead of `gatsby-plugin-mdx` since it's included in www's [gatsby-config.js](https://github.com/gatsbyjs/gatsby/blob/master/www/gatsby-config.js#L158) file?

Also, I'm a bit confused by this setup since it looks like `gatsby-remark-prismjs` is only [used as a plugin]((https://github.com/gatsbyjs/gatsby/blob/master/www/gatsby-config.js#L178)) for `gatsby-transformer-remark` but not for `gatsby-plugin-mdx` but the docs are [generated using the `gatsby-plugin-mdx`](https://github.com/gatsbyjs/gatsby/blob/master/www/gatsby-node.js#L641). I [don't see a syntax highlighter](https://github.com/gatsbyjs/gatsby/blob/master/www/gatsby-config.js#L135) set as a plugin for `gatsby-plugin-mdx`. What am I missing?

**Comment by pieh** [comment]
I was wrong with this, seems like we use remark for all readme files in plugin library (including ones we maintain in the same repo) - so it doesn't seem mdx related:
Screenshot 2019-09-02 at 18 28 50

But at least one variable is excluded :S

**Comment by dkthehuman** [comment]
FWIW, I updated my [test repo](https://github.com/dkthehuman/gatsby-prismjs-highlight-demo/blob/master/gatsby-config.js#L13) to use `gatsby-plugin-mdx` instead of `gatsby-remark-markdown` and `gatsby-remark-prismjs` performs as expected. (It shows `// highlight-line` when no language is set instead of processing it.)

**Comment by pieh** [comment]
It does seem like maybe it's not prismjs plugin doing - maybe is something else (?), I'll try to pull down your changes and point out what exactly is happening

**Comment by pieh** [comment]
Ok, it does seems roughly related to mdx after all - but mostly to our custom CodeBlock component for .org site:

https://github.com/gatsbyjs/gatsby/blob/edbd0d3c6dd3566a33626ff879a6508c5d619943/www/src/components/code-block/index.js#L37-L42

In here `normalize` will strip "directives" (so comments like `// highlight-line` regardless of lang). We do need to fix that, but not quite sure yet what's proper way to do this.

Replacing that `normalize` call with:
```
const [content, highlights] = [
    children.props && children.props.children
      ? children.props.children
      : children,
    [],
  ]
```
Results in:
Screenshot 2019-09-02 at 19 21 50

It completely get rids of highlighting (and stripping highlight directives), but this is not proper fix (just showcase that this is where the problem is). We do need this `normalize` function to behave similarly as code in `prismjs` plugin

LekoArts:
Hey @DSchau 	, friendly reminder 😊

DSchau:
@dkrasniy I think a better solution is to just quick return (e.g. return just the content) if a language isn't specified for a code block.

The code is here: https://github.com/gatsbyjs/gatsby/blob/master/www/src/components/code-block/normalize.js#L64

We'll want to update this function to short-circuit when an empty "className" (which is something like lang-javascript or empty if no language), and we'll want to return a similar structure to what gets returned by default with the normalize function (e.g. an array of arrays).

Let me know if you'd like any more info here -- hopefully this helps!

PR #17245, opened 2019-08-30
feat(gatsby): Implement tracing for graphql

freiksenet:
This adds tracing to GraphQL resolvers.

A caveat - it measures only resolver time, not children time yet (though it displays them correctly).


freiksenet:
CC @stefanprobst

packages/gatsby/src/query/graphql-runner.js
```diff
@@ -18,23 +20,96 @@ class GraphQLRunner {
     })
   }
 
-  query(query, context) {
+  async query(query, context, { parentSpan } = {}) {
     const { schema, schemaCustomization } = this.store.getState()
 
-    return graphql(
+    const activity = new SpanTracker(`GraphQL Query`, {
```
wardpeet: I would rename to spanTracker as it collides with createActivity.

```suggestion
    const spanTracker = new SpanTracker(`GraphQL Query`, {
```

packages/gatsby/src/query/graphql-runner.js
```diff
@@ -18,23 +20,96 @@ class GraphQLRunner {
...
+
+  start() {}
+
+  done() {
+    this.parentSpan.finish()
+  }
+
+  createActivity(path, name, args) {
+    let prev = path.prev
+    if (prev && Number.isInteger(prev.key)) {
```
wardpeet: https://github.com/graphql/graphql-js/blob/a482d3b24d7a6d4a8027614ab6304e980548e16f/tstypes/jsutils/Path.d.ts

says it can also be a string, i'm unsure where this comes from.

packages/gatsby/src/query/graphql-runner.js
```diff
@@ -18,23 +20,96 @@ class GraphQLRunner {
...
+      start() {},
+
+      done() {
+        span.finish()
+      },
+    }
+  }
+
+  getSpan(gqlPath) {
+    const path = pathToArray(gqlPath)
```
wardpeet: It's hard to understand where you do the null check but basically pathToArray is guarding us because of the while loop. Maybe make it explicit.

```js
if (!gqlPath) {
  return this.parentSpan;
}
```

packages/gatsby/src/query/queue.js
```diff
@@ -99,7 +99,15 @@ const processBatch = async (queue, jobs, activity) => {
   if (numJobs === 0) {
     return Promise.resolve()
   }
-  const runningJobs = jobs.map(job => pushJob(queue, job, activity))
+
+  const runningJobs = jobs.map(job =>
+    pushJob(queue, job).then(v => {
```
muescha: this merge removed here the `activity` param which is needed in `pushJob`:

https://github.com/gatsbyjs/gatsby/blob/88e7c7f1a889e2c756e572878385e54165770c94/packages/gatsby/src/query/queue.js#L84-L90

m-allanson:
Is there a strong need for this at the moment? If not, let's close and revisit when it's needed.

freiksenet:
Closing in favour of #23589

PR #17253, opened 2019-08-30
feat(gatsby-image): Add support for Save-Data header

### LINKED ISSUE #17202: Respect Save-Data client hint in gatsby-image
Many people on constrained data plans turn on data saving — their smartphone browsers then send a "Save-Data" header to servers.

We can use this in the client as well to choose lower-data rendering options for these people.

For gatsby-image, we can render regular resolution images for these people instead of retina.

```js
if ("connection" in navigator) {
    if (navigator.connection.saveData === true) {
        // Implement data saving operations here.
    }
}
```

We can detect in the component if data saving is enabled and then when rendering, only render the `src` and not `srcset` thus avoiding making the user download retina (so much bigger) images.

The source code for gatsby-image is at https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-image/src/index.js

We can use a similar pattern for this as we do with detecting native lazy loading support https://github.com/gatsbyjs/gatsby/blob/693f3edc43f0dee0766f5e3a07de32a4cfbba22c/packages/gatsby-image/src/index.js#L81

For background reading on Save-Data:

- https://timkadlec.com/remembers/2019-08-29-save-data-usage/
- https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/save-data/

charlyx:
Hi 👋 

I've just started using `gatsby-image` so I thought I'd help by working on some issues.
Great job though!

## Description

This PR should add support for Save-Data header.

It actually removes srcSet attribute from img tag when Save-Data is enabled.
I thought I should remove source tag from the DOM when Save-Data is enabled so I did: I'm wondering if did the right thing though?

## Related Issues

Fixes #17202


packages/gatsby-image/src/index.js
```diff
@@ -82,6 +82,10 @@ const hasNativeLazyLoadSupport =
   typeof HTMLImageElement !== `undefined` &&
   `loading` in HTMLImageElement.prototype
 
+// Save-Data support: https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/save-data/#detecting_the_save-data_setting
+const hasSaveDataEnabled = () =>
+  `connection` in navigator && navigator.connection.saveData === true
```
pieh: `navigator` global object doesn't exist when generating html pages (we are running in Node.js and not in browser). So we will need to use something like `` typeof window === `undefined` `` to not crash during html generation

Question is: what do we want to display when JavaScript is disabled or in first mount (hydration doesn't patch differences in attributes - see notes in https://reactjs.org/docs/react-dom.html#hydrate). Should we use what we currently use (so include srcSet) or not?

packages/gatsby-image/src/index.js
```diff
@@ -415,7 +419,11 @@ class Image extends React.Component {
 
     if (fluid) {
       const imageVariants = fluid
-      const image = imageVariants[0]
+      const image = { ...imageVariants[0] }
+
+      if (hasSaveDataEnabled()) {
+        delete image.srcSet
+      }
```
pieh: Should we instead use `hasSaveDataEnabled` check in place where we generate responsive sizes elements? Creating copy, deleting field and then checking for field seems pretty hacky?

packages/gatsby-image/src/index.js
pieh: oh, we do have `isBrowser` thingie just below this - so we can incorporate that in our check

pieh:
I'm also thinking about scenarios where removing srcSet might actually be hurtful for data usage? It totally makes sense to cut out retina resolutions, but removing entire `srcSet` removes also those with smaller sizes than nominal image, right?

packages/gatsby-image/src/index.js
charlyx: Unfortunately, we cannot. Those information are the result of GraphQL queries using gatsby-plugin-sharp. They're then passed as props to the Image component.
It creates a copy in order to keep the original object as is.

pieh: Sure, but we don't need to delete `srcSet` field - we can just not use it if `hasSaveDataEnabled()` is true, right?

packages/gatsby-image/src/index.js
```diff
@@ -85,6 +85,12 @@ const hasNativeLazyLoadSupport =
 const isBrowser = typeof window !== `undefined`
 const hasIOSupport = isBrowser && window.IntersectionObserver
 
+// Save-Data support: https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/save-data/#detecting_the_save-data_setting
+const hasSaveDataEnabled = () =>
+  isBrowser &&
+  `connection` in navigator &&
+  navigator.connection.saveData === true
```
pieh: Does this need to be a function? Can we just set `hasSaveDataEnabled` variable once and not "recalculate" it?

charlyx: I need `hasSaveDataEnabled` to be a function in order to unit test this behaviour.
How would you test it if this value is computed once and for all?

pieh: Yeah, this would make unit testing more difficult and there is not enough upside for doing that - so let's keep it as is

charlyx:
@pieh I agree with you and I asked myself the exact same question.
I made this PR the same as requested in the issue (unless I missed something). Also, I did some test and it appears that the image provided to img's src attribute is not the biggest one but more like the "average".

I don't see any other solution than parsing the srcSet and check every specified size against the current viewport to select the best one.

Shall we introduce more complexity to this PR or make another one in order to improve it?

pieh:
> @pieh I agree with you and I asked myself the exact same question.
> I made this PR the same as requested in the issue (unless I missed something).

I think this warrants asking clarifying question in the issue. Details like this can be often missed when creating feature requests and it's always good to clarify it

> Also, I did some test and it appears that the image provided to img's src attribute is not the biggest one but more like the "average".

I called `src` "nominal" (I lack better word for it) - it should match arguments passed to queries for `fixed` (exactly width x height specified in args) and `fluid` (constrained by `maxWidth` arg). 

`gatsby-plugin-sharp` will create both larger (retina) and smaller versions for `fluid` variant:
https://github.com/gatsbyjs/gatsby/blob/d3960642a69873fe8188eca2042bfbd35e91e073/packages/gatsby-plugin-sharp/src/index.js#L319-L336

but just higher resolution versions for `fixed` variant:
https://github.com/gatsbyjs/gatsby/blob/d3960642a69873fe8188eca2042bfbd35e91e073/packages/gatsby-plugin-sharp/src/index.js#L456-L460

> I don't see any other solution than parsing the srcSet and check every specified size against the current viewport to select the best one.

Yeah, this is tough. We don't exactly provide enough information to easily filter retina version out. It's also important to note that `fixed` and `fluid` uses a bit different methods (`fluid` will use both `srcSet` and `sizes`, while `fixed` will use only `srcSet`). `fluid` follow this https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#Resolution_switching_Different_sizes, while `fixed` uses https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#Resolution_switching_Same_size_different_resolutions (note `x` there).

It get's bit more difficult when we realize that `gatsby-image` isn't used only for `gatsby-transformer-sharp` results - lot of plugins provide `gatsby-image` compatible data:
 - Contentful 
 - DatoCMS - https://github.com/datocms/gatsby-source-datocms#integration-with-gatsby-image
 - Sanity - https://github.com/sanity-io/gatsby-source-sanity#using-images

So we also can't make too much assumptions here :(

It really seems like we need to implement something that need to emulate browser :/ to make it work correctly

> Shall we introduce more complexity to this PR or make another one in order to improve it?

Given that goal of the feature is to lower data bandwidth for those on constrained data plans, making it potentially worse for them by removing lower image resolution variants (for mobile devices) probably warrant iterating on it before merging it

wardpeet:
We can be pretty aggressive about our choices as the api of gatsby-image is pretty clear. Fixed images have the 1x, 2x part & fluid have the viewport (w & h) indicators. We can write a function to parse a srcset and take in considerations of viewport & sizes.

I wrote a function that does this (expect the sizes) part. I didn't check any edge cases and I've definitely forgotten a few srcset variants. So use this as a guideline.
https://runkit.com/wardpeet/5d6e22275015e5001be24048

To see what sizes we need to match we can use MatchMedia which is a pretty good api.

My function only takes in account when there is one fluid & one fixed version. We also support art-direction which has multiple <source tags inside a picture.

More info can be found https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#Art_direction

muescha:
Instead of reverse parsing the `srcSet`: maybe the `gatsby-image` can give more structured meta data to the gtaphql?

KyleAMathews:
@wardpeet is the hold up on this PR getting your suggestions incorporated?

pvdz:
Hi @charlyx :) I know it's been a while, sorry about that.

I'm afraid we can't merge the PR as it stands because we really have to solve the approach to first-render before we can implement it. I've left some notes on #17202 regarding that.

What we can do is implement the client-side version of it, where we conditionally set the `src` of an image to the lowest known rez version. Gatsby should have a set of sources ordered by resolution. Worst case that's still goign to be the same image, best case it'll be lowrez.

My suggestion would be to either update this PR, or to create a new one, that implements this for the client side. If datasaver, use the lowest rez that Gatsby is aware of. Otherwise use the default. But don't change anything for the statically generated html.

@charlyx can you please let me know how you feel about this? Do you want to follow up with this? I think it'd be super nice to have this feature in Gatsby!

pvdz:
Considering @charlyx hasn't chimed in since September, is not responding to my last message, and nobody else offered to amend the PR, I'm going to close this.

We're not against supporting this feature but the PR will need some work in its current form. If you, or somebody, wants to continue working on this we'd be happy to help.

charlyx:
Sorry about the radio silence @pvdz :(
Unfortunately, I won't be able to work on it anytime soon, so you did well.

I might have a try later if it's still available but for the time being feel free to make it :)

PR #17261, opened 2019-08-30
feat(gatsby): Add support for webpack multicompiler mode

Eddie-CooRo:


wardpeet:
@Eddie-CooRo could you give us some information on why you need this change? And what the problem is you're having? We're likely not going to accept this PR as multi-webpack compilation will break a lot of things inside gatsby. I'm happy to understand your use case and maybe workaround it.

Eddie-CooRo:
Hey @wardpeet, thank you for your response.
I am aware of some problems that multi compiler mode causes. But I think they are fixable by, for example, using a single instance of webpack-assets-plugin in both configurations. If it's not the case, then feel free to close the PR.
And about the use-case. I was trying to achieve some sort of differential serving without changing the gatsby internals too much.

wardpeet:
> And about the use-case. I was trying to achieve some sort of differential serving without changing the gatsby internals too much.

What do you mean with differential serving? Why not go for a webpack plugin instead? You have access to childcompiler and all assets available in webpack.

Eddie-CooRo:
By differential serving I mean something like what you are doing in #2114, But with I try to do it mostly in the user side code than the gatsby's codebase itself. Today I digged into the gatsby codes a little, and I found out why this PR won't work. Probably I'll take the webpack plugin path.
Sorry for taking your time, I'll let you know when I reached somewhere with the plugin.

PR #17277, opened 2019-08-31
docs(gatsby-cli): update readme

rishabh3112:
Documentation Update 

## Description
1. Adds version badge
2. Links in commands to browse quikly

## Related Issues
N/A


packages/gatsby-cli/README.md
```diff
@@ -1,4 +1,5 @@
 # gatsby-cli
+[![npm version](https://badge.fury.io/js/gatsby-cli.svg)](https://badge.fury.io/js/gatsby-cli)
```
wardpeet: We're not super fond of adding badges to our readme files as there are a lot of dependencies that we could add badges for, but I don't think that helps people in deciding to use Gatsby or not.

```suggestion
```

rishabh3112: Sure, applied the suggestion!

PR #17299, opened 2019-09-01
feat(gatsby): pass PageComponent prop to wrapPageElement APIs

### LINKED ISSUE #15820: Pass `PageComponent` prop to `wrapPageElement` APIs
I'm using gatsby-plugin-layout for sharing states that persist across routes. Can pages be components instead of elements when passed to Layout as children?

So, instead of
```javascript
export default function Layout ({ children }) {
  return <State>{children}</State>
}
```

It can be
```javascript
export default function Layout ({ PageComponent }) {
  return <State><PageComponent /></State>
}
```

This is similar to how Next.js does it https://nextjs.org/docs#custom-app.
Use case is passing props from Layout to page components. Since page is direct children of layout, using Context can be a bit of an overkill for some cases.
I understand that React.cloneElement can be used. Yet, I feel that passing props to components is more idiomatic.

universse:
## Description
N.A

## Related Issues
Fixes #15820

### DIFF
--- a/packages/gatsby/cache-dir/page-renderer.js
+++ b/packages/gatsby/cache-dir/page-renderer.js
@@ -6,12 +6,12 @@ import { apiRunner } from "./api-runner-browser"
 // Renders page
 class PageRenderer extends React.Component {
   render() {
-    const Page = this.props.pageResources.component
+    const PageComponent = this.props.pageResources.component
 
     const props = {
       ...this.props,
       pathContext: this.props.pageContext,
-      Page,
+      PageComponent,
     }
 
     const [replacementElement] = apiRunner(`replaceComponentRenderer`, {
@@ -21,7 +21,7 @@ class PageRenderer extends React.Component {
 
     const pageElement =
       replacementElement ||
-      createElement(Page, {
+      createElement(PageComponent, {
         ...props,
         key: this.props.path || this.props.pageResources.page.path,
       })
--- a/packages/gatsby/cache-dir/page-renderer.js
+++ b/packages/gatsby/cache-dir/page-renderer.js
@@ -6,9 +6,12 @@ import { apiRunner } from "./api-runner-browser"
 // Renders page
 class PageRenderer extends React.Component {
   render() {
+    const Page = this.props.pageResources.component
+
     const props = {
       ...this.props,
       pathContext: this.props.pageContext,
+      Page,
     }
 
     const [replacementElement] = apiRunner(`replaceComponentRenderer`, {
@@ -18,7 +21,7 @@ class PageRenderer extends React.Component {
 
     const pageElement =
       replacementElement ||
-      createElement(this.props.pageResources.component, {
+      createElement(Page, {
         ...props,
         key: this.props.path || this.props.pageResources.page.path,
       })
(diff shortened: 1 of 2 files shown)


packages/gatsby/cache-dir/page-renderer.js
```diff
@@ -6,9 +6,12 @@ import { apiRunner } from "./api-runner-browser"
 // Renders page
 class PageRenderer extends React.Component {
   render() {
+    const Page = this.props.pageResources.component
```
wardpeet: Let's call it PageComponent to make it more verbose

packages/gatsby/cache-dir/static-entry.js
```diff
@@ -173,17 +173,17 @@ export default (pagePath, callback) => {
 
   class RouteHandler extends React.Component {
     render() {
+      const Page = syncRequires.components[componentChunkName]
```
wardpeet: Same here

universse:
> 
> 
> Thanks for doing the work! I added a small nit on the variable naming. Could you share a small example on how to test this behaviour properly?

Noted. As for the example, I intend to directly modify the files in node_modules of my own project. Is that sufficient?

wardpeet:
Basically I need to replace my wrapPageElement with 
```
export default function Layout ({ PageComponent }) {
  return <State><PageComponent /></State>
}
```

instead of children to test this?

universse:
It would be
```
export default function Layout ({ children, PageComponent, ...pageProps }) {
  return <State><PageComponent {...pageProps} extraProp='hi' /></State>
}
```
I guess the benefit of this is to be able to pass additional props without having to use Context.

pieh:
> And guess I should update the doc for [gatsby-plugin-layout](https://www.gatsbyjs.org/packages/gatsby-plugin-layout/) too.

You wouldn't be able to use this feature with `gatsby-plugin-layout` as we do use `element` there (and not `PageComponent` - https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-layout/src/wrap-page.js#L22), but we would need to update documentation for `wrapPageElement`:
SSR APIs docs - https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/cache-dir/api-ssr-docs.js#L142-L164
Browser APIs docs - https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/utils/api-browser-docs.js#L127-L149

pieh:
Ok, one important thing to note here:
using `PageComponent` instead of `element` is potentially dangerous as it could discard any changes made by other plugins.

The way it works now is that `element` is passed to subsequent plugins is being updated with each `wrapPageElement` call:

so if we have 2 plugins (let's call them A and B) that use this API, it works like that:

we call `wrapPageElement` in plugin A and it returns wrapped element (let's say it wraps page with some Provider component). In plugin B, `element` now will be page that is wrapped in provider, but PageComponent still be just original Page Component - so if in plugin B you use that, it will discard Provider used in plugin A - so this need to be documented (because I'm not sure if this can be fixed)

universse:
Thanks for the info. It's something I haven't considered. Will definitely need more time for this.

wardpeet:
@universse, I'll be closing this one. Feel free to re-open when you get to this one.

PR #17314, opened 2019-09-02
feat(telemetry): Refactor and test CI detection and add support for heroku and now

jamo:
This PR refactors and adds tests for CI detection. 

While `ci-info` detects many CIs we've already had additional logic in place for detecting a few more CIs. This cleans it up and adds more accurate detection for heroku and Now v1 and v2 platforms.

This is to be appended while we identify new CI.

### DIFF
--- a/packages/gatsby-telemetry/src/ci.js
+++ b/packages/gatsby-telemetry/src/ci.js
@@ -1,0 +1,70 @@
+const ci = require(`ci-info`)
+
+const CI_DEFINITIONS = [
+  envFromCIandCIName,
+  getEnvDetect({ key: `NOW_BUILDER_ANNOTATE`, name: `ZEIT Now` }),
+  getEnvDetect({ key: `NOW_REGION`, name: `ZEIT Now v1` }),
+  herokuDetect,
+  getEnvFromCIInfo,
+  envFromCIWithNoName,
+]
+
+function lookupCI() {
+  for (const fn of CI_DEFINITIONS) {
+    try {
+      const res = fn()
+      if (res) {
+        return res
+      }
+    } catch (e) {
+      // ignore
+    }
+  }
+  return null
+}
+const CIName = lookupCI()
+
+function isCI() {
+  return !!CIName
+}
+
+function getCIName() {
+  if (!isCI()) {
+    return null
+  }
+  return CIName
+}
+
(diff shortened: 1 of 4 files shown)


wardpeet:
@jamo could you explain a bit what the incentive is here?

jamo:
@wardpeet added the description

packages/gatsby-telemetry/src/ci.js
```diff
@@ -0,0 +1,60 @@
...
+  return function() {
+    if (process.env[key]) {
+      return name
+    }
+    return null
+  }
+}
+
+function herokuDetect() {
+  return /\.heroku\/node\/bin\/node/.test(process.env.NODE) && `Heroku`
```
pieh: I'm not sure of reliability of this - I checked one of my Dynos and `process.env.NODE` is not set there:

```shell
~ $ node -e "console.log(process.env)"
{ _: '/app/.heroku/node/bin/node',
  NODE_HOME: '/app/.heroku/node',
  NODE_ENV: 'production',
  WEB_MEMORY: '512',
  DYNO: 'run.7506',
  PWD: '/app',
  HOME: '/app',
  PORT: '33175',
  MEMORY_AVAILABLE: '512',
  TERM: 'xterm-256color',
  WEB_CONCURRENCY: '1',
  SHLVL: '2',
  PATH:
   '/app/.heroku/node/bin:/app/.heroku/yarn/bin:/usr/local/bin:/usr/bin:/bin:/app/bin:/app/node_modules/.bin',
  PS1:
   '\\[\\033[01;34m\\]\\w\\[\\033[00m\\] \\[\\033[01;32m\\]$ \\[\\033[00m\\]' }
~ $ node -e "console.log(process.env.NODE)"
undefined
```

pieh: I might completely missconfigure my dyno - but it also means people can run gatsby on Heroku in same way I did :P

jamo: It appears to be set in the build process. But indeed they don't set any `heroku=1` like environment variables. And in my couple samples the build process has NODE set like this: `NODE=/tmp/build_8a43526a8849e690a3b67906d404e434/.heroku/node/bin/node`

PR #17322, opened 2019-09-02
fixed typos

aneeshjose:
## Description
fixed typo

## Related Issues
n/a


pieh:
In our style guide we have preference for US english ( https://www.gatsbyjs.org/contributing/gatsby-style-guide/#prefer-us-english ), so we should keep `flavors` here

PR #17332, opened 2019-09-02
docs(blog): implementing draft feature in a gatsby blog

kingingcole:
## Description

I added a new article.


kingingcole:
Are these tests that failed from my end? I cannot seem to understand why they failed.

lannonbr:
@kingingcole Yes it is coming back that your blogpost is not being properly linted. Try installing [Prettier](https://prettier.io/) and running it on your index.md file.

sidharthachatterjee:
Thanks so much for taking the time to write this @kingingcole 

Unfortunately, we won't be accepting this submission. We believe this best belongs in your personal blog because it seems like the article is _not_ relevant for the entire Gatsby blog audience. 

This, however, doesn't mean that it isn't a valuable blog post. We would love to see it published on your blog (if it isn't already) 

Thank you once again for taking the time to write this and for using Gatsby 💜

PR #17334, opened 2019-09-03
fix(gatsby-plugin-catch-links): fix pluginOptions

### LINKED ISSUE #17324: [gatsby-plugin-catch-links] documentation for plugin options is not correct
## Description

The example that is given for Plugin Options for gatsby-plugin-catch-links shows 'excludePattern' the expected value is 'excludeRegex'

### Steps to reproduce

https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-catch-links

universse:
## Description

I edited source instead of doc since users are already following current doc.

## Related Issues

Fixes #17324

### DIFF
--- a/packages/gatsby-plugin-catch-links/src/catch-links.js
+++ b/packages/gatsby-plugin-catch-links/src/catch-links.js
@@ -146,6 +146,10 @@ export const routeThroughBrowserOrApp = (
 
   if (hashShouldBeFollowed(origin, destination)) return true
 
+  if (pluginOptions.excludeRegex && !pluginOptions.excludePattern) {
+    pluginOptions.excludePattern = pluginOptions.excludeRegex
+  }
+
   if (pluginOptions.excludePattern) {
     console.log(
       `pluginOptions.excludePattern.test(destination.pathname)`,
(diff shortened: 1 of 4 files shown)


packages/gatsby-plugin-catch-links/src/catch-links.js
```diff
@@ -146,12 +146,12 @@ export const routeThroughBrowserOrApp = (
 
   if (hashShouldBeFollowed(origin, destination)) return true
 
-  if (pluginOptions.excludeRegex) {
+  if (pluginOptions.excludePattern) {
```
wardpeet: Let's make sure we support both cases, just in case.

```suggestion
  if (pluginOptions.excludeRegex && !pluginOptions.excludePattern) {
    pluginOptions.excludePattern = pluginOptions.excludeRegex
  }

  if (pluginOptions.excludePattern) {
```

wardpeet:
Seems like we never used the regex as json.stringify removes all types of a regex so I've added a small change to this PR to make it work.

universse:
Thanks for reviewing and fixing!

PR #17349, opened 2019-09-03
Docs: Added content to SaaS services stub

kevee:
Added content to the [Sourcing from SaaS Services](https://www.gatsbyjs.org/docs/sourcing-from-saas-services/)


LekoArts:
Hi, thanks for creating the PR! @marcysutton 	will have a look at this eventually (and maybe rephrase what she had in mind for that page).

I personally imagine this page do not only contain links to some other pages (btw, you could link guides instead of the packages if existent) but also include a general overview of this approach, e.g. that you need source plugins for them (or create your own), webhooks are good use, the APIs can differ in how much they allow/expose etc.

kevee:
I agree that we should front-load a bit more information. I was following the format of the [Sourcing from Databases](https://www.gatsbyjs.org/docs/sourcing-from-databases/) page, but given how broad the topic of this page could be, I agree a simple list might not cut it.

docs/docs/sourcing-from-saas-services.md
```diff
@@ -2,7 +2,16 @@
 title: Sourcing from SaaS Services
 ---
 
-This is a stub. Help our community expand it.
+There are many Gatsby plugins that integrate with SaaS services. You can search for more [Gatsby plugins](https://www.gatsbyjs.org/plugins/) and [Gatsby starters](https://www.gatsbyjs.org/starters/) that integrate with hosted services. Some examples include:
```
marcysutton:
```suggestion
There are many Gatsby plugins that integrate with SaaS (Software as a Service). You can search for [Gatsby plugins](/plugins/) as well as [Gatsby starters](/starters/) that integrate with hosted services to source content and functionality. Some examples include:
```

docs/docs/sourcing-from-saas-services.md
```diff
@@ -2,7 +2,16 @@
...
+- [AWS DynamoDB](/packages/gatsby-source-dynamodb)
+- [Contentful](/packages/gatsby-source-contentful/)
+- [Firebase](/packages/gatsby-source-firebase)
+- [Google Docs](/packages/gatsby-source-google-docs)
+- [Prismic](/packages/gatsby-source-prismic-graphql)
+- [Shopify](/packages/gatsby-source-shopify)
+
+These plugins will pull data from a hosted service and make it available to your Gatsby site via GraphQL.
+
+You can also [write your own source plugin](/docs/creating-a-source-plugin/).
```
marcysutton:
```suggestion
You can also [write your own source plugin](/docs/creating-a-source-plugin/) to integrate with a service that isn't already represented.
```

kevee:
Thanks for the feedback, @marcysutton! I think this could either go forward as it is, or we can regroup and add more. The general topic of the stub was so broad that I think it should be a large document, or a way to disambiguate SaaS sources.

marcysutton:
I personally think we should ship this PR and if you're interested in making further improvements, those could come later!

kevee:
Lets ship it.

marcysutton:
Thanks again for the PR, we really appreciate your contributions! :taco:

PR #17366, opened 2019-09-03
fix(gatsby-graphiql-explorer): Add missing dep

sidharthachatterjee:
Well, that was fun


sidharthachatterjee:
schrodinger's webpack 😄 

It's both there and not there at the same time

Issue #13965, opened 2019-05-10
Gatsby clientside pages: inject route match params as props


slorber:
# Summary 
Currently if you do

`createPage({path: "/item/:itemId", matchPage: "/item/:itemId", component: itemCompPath})`

And when you go to `/item/123`

It's not straightforward at all how to extract the 123 value from the url elegantly.

I think Gatsby should make this usecase a lot better by providing those params as page props. 

# Details

There are some mentionned examples here and there which rely on importing reach router `<Router>` comp (which by the way does not play well at all with Typescript). This is probably the way to go if you have a lot of pages under the same match param like `/authenticated/*`, but it's not convenient at all when you need to declare a single path.

There is also the `Match` component that can be used (it plays better with TS)

Internally, Gatsby is using `const { match: reachMatch } = require(`@reach/router/lib/utils`)` to do that. 

I think Gatsby should extract reach router params automatically for us and inject them directly under the page component. These params are already provided by `match()` so Gatsby should be able to add these params to the props.

# Workaround

If you use the layout plugin, it is possible to extract the params using that plugin and inject them yourself in the page.

It took me some time to find the best/reusable solution for this problem, hope this will be helpful to others who need the same feature

```
import React from 'react';

// See https://github.com/reach/router/issues/74#issuecomment-397085896
// @ts-ignore
import { match } from '@reach/router/lib/utils';

export const extractPathParams = (props: any): any | null => {
  if (props.pageContext.matchPath) {
    const result = match(props.pageContext.matchPath, props.location.pathname);
    if (result && result.params) {
      return result.params;
    }
  }
  return null;
};

const RootLayout = (props: any) => {
  return React.cloneElement(props.children, extractPathParams(props));
};

export default RootLayout;
```

```
    {
      resolve: `gatsby-plugin-layout`,
      options: {
        component: require.resolve(`./src/layout/rootLayout.tsx`),
      },
    },
```

cc @DSchau

universse:
You can do this also

```javascript
const getParamFromPathname = pathname =>
  pathname
    .split('/')
    .pop()

export default Layout ({ location: { pathname } }) {
  const param = getParamFromPathname(pathname)

  // ...
}
```
One caveat is this works if pathname doesn't have trailing slash. To work with trailing slash e.g. /product/123/
```javascript
const getParamFromPathname = pathname =>
  pathname.slice(0, -1)
    .split('/')
    .pop()
```

rayriffy:
[...]

PR #17384, opened 2019-09-04
docs(gatsby): document client-only route param

universse:
## Description

I'm not sure if anything need to be done server side. If approved, need to update doc also.

## Related issues

#13965

### DIFF
--- a/packages/gatsby/cache-dir/page-renderer.js
+++ b/packages/gatsby/cache-dir/page-renderer.js
@@ -1,6 +1,5 @@
 import React, { createElement } from "react"
 import PropTypes from "prop-types"
-import { match } from "@reach/router/lib/utils"
 import { publicLoader } from "./loader"
 import { apiRunner } from "./api-runner-browser"
 
@@ -10,15 +9,6 @@ class PageRenderer extends React.Component {
     const props = {
       ...this.props,
       pathContext: this.props.pageContext,
-    }
-
-    if (this.props.pageContext.matchPath) {
-      const { params } = match(
-        this.props.pageContext.matchPath,
-        this.props.location.pathname
-      )
-      props.pageContext.params = params
-      props.pathContext.params = params
     }
 
     const [replacementElement] = apiRunner(`replaceComponentRenderer`, {
--- a/packages/gatsby/cache-dir/page-renderer.js
+++ b/packages/gatsby/cache-dir/page-renderer.js
@@ -1,5 +1,6 @@
 import React, { createElement } from "react"
 import PropTypes from "prop-types"
+import { match } from "@reach/router/lib/utils"
 import { publicLoader } from "./loader"
 import { apiRunner } from "./api-runner-browser"
 
@@ -9,6 +10,15 @@ class PageRenderer extends React.Component {
     const props = {
       ...this.props,
       pathContext: this.props.pageContext,
+    }
+
+    if (this.props.pageContext.matchPath) {
+      const { params } = match(
+        this.props.pageContext.matchPath,
+        this.props.location.pathname
+      )
+      props.pageContext.params = params
+      props.pathContext.params = params
     }
 
     const [replacementElement] = apiRunner(`replaceComponentRenderer`, {


packages/gatsby/cache-dir/page-renderer.js
```diff
@@ -11,6 +12,15 @@ class PageRenderer extends React.Component {
       pathContext: this.props.pageContext,
     }
 
+    if (this.props.pageContext.matchPath) {
+      const { params } = match(
+        this.props.pageContext.matchPath,
+        this.props.location.pathname
+      )
+      props.pageContext.params = params
+      props.pathContext.params = params
```
muescha: 2x? To path and page?

universse: I guess it's for compatibility purpose. Cuz right above there is `pathContext: this.props.pageContext,`

slorber:
+1

On my app I use this:



```jsx
export const extractPageMatchParams = (props: any): any | null => {
  if (props.pageContext.matchPath) {
    const result = match(props.pageContext.matchPath, props.location.pathname);
    if (result && result.params) {
      return result.params;
    }
  }
  return {};
};
```

I prefer to return {} if no match params, which makes user code simpler (no need to check for null/undefined of the params object...).

Note during SSR params will be always {} (or undefined)

It's more convenient to write

```
if ( pageContext.params.userId ) {}
```

compared to 

```
const userId = isServer() ? undefined : pageContext.params.userId;
if ( pageContext.params.userId ) {}
```

universse:
> I prefer to return {} if no match params, which makes user code simpler (no need to check for null/undefined of the params object...).

Just to clarify, what are the cases where there is no matched params. Cuz `match` will return empty string e.g. if you go to path `/user` , `match` will return `{ userId: '' }`.

slorber:
hmm, maybe my code is useless then, probably added some useless securities

packages/gatsby/cache-dir/page-renderer.js
pieh: Yeah, we changed it between v1 and v2 and kept `pathContext` (alias really) for backward compatibility.

That said, I feel like this shouldn't be in neither of them and maybe this should be top level prop? I need to check how @reach/router does this - because I feel like <Router> should automatically give us this props and we just might drop those on way from <Router> to user's page component (I don't think we should need to match again, as we already matched before that)

pieh: So reach/router, given `invoices/:invoiceId` matchPath will use following props:
```
{
  "path": "invoices/:invoiceId",
  "invoiceId": "123",
  "uri": "/invoices/123",
  "location": {
    "replace": "fn()",
    "href": "https://k9lzn45065.csb.app/invoices/123",
    "ancestorOrigins": "DOMStringList(…)",
    "origin": "https://k9lzn45065.csb.app",
    "protocol": "https:",
    "host": "k9lzn45065.csb.app",
    "hostname": "k9lzn45065.csb.app",
    "port": "",
    "pathname": "/invoices/123",
    "search": "",
    "hash": "",
    "assign": "assign()",
    "reload": "reload()",
    "toString": "toString()",
    "state": {
      "key": "1567638513581"
    },
    "key": "1567638513581"
  },
  "navigate": "navigate()"
}
```
so param is being inserted just as top level prop - it sounds like good idea to preserve this behaviour - what do you think?

universse: Thanks for reviewing. That looks good too. However, there is a rare edge case where the param key is the same as one of the props e.g. `/uris/:uri` or `/paths/:path`. But I guess that's not a big concern.
cc @slorber

universse:
> 
> 
> hmm, maybe my code is useless then, probably added some useless securities

Oops sorry this only applies when param is `*`. Your solution is correct.

packages/gatsby/cache-dir/page-renderer.js
universse: Anw looks like @reach/router alr does this. From Page component, we can simply access `props[paramKey]` to get our param.
One important caveat is @reach/router will return `null` when we access index path e.g. `/user` for matchPath `/user/:userId`.
Guess we need to update the doc then.

slorber: Yeah that would be fine if we just forwarded the gatsby match call result to the page instead of rematching.

universse: And what I meant is we cal alr do it currently. Just use props[paramKey] from page components.

LekoArts:
Hi, what's the status of this PR? :)
You haven't committed any changes.

universse:
I looked through Gatsby doc for client-only route and there's currently nothing that covers the use case mentioned in #13965.

I was thinking of adding a guide to [this section](https://www.gatsbyjs.org/docs/adding-website-functionality/), something along the line of `Handling highly dynamic data/user-generated contents`. E.g. fetching comments for blog posts. This showcases how to handle both **static** (blog posts) and **dynamic, user-generated** data (comments).

What do you guys think?

LekoArts:
Feel free to open up an issue to discuss this topic and best approaches before creating a PR. In the meantime I'll close this PR :)

PR #17389, opened 2019-09-04
fix: close html comment in recipes.md

wardpeet:
## Description
Typo fix :joy:

Closing html comment so .org builds again.

## Related Issues
¯\_(ツ)_/¯


wardpeet:
locally worked so it should work 💃

marcysutton:
Thanks for this! Maybe that's the problem I was running into 🤔

sidharthachatterjee:
@marcysutton Yup, was most likely this!

PR #17411, opened 2019-09-05
fix(gatsby): order matchpaths paths by specificity

### LINKED ISSUE #16098: Match path's aren't correctly sorted
## Description

Hi team! I've been encountering some inconsistencies with match path sorting. With two pages with match paths of `/*` and `/txt/*`, the `/*` page gets sorted before the `/txt/*` page. But if I include a `/app/*` page, that gets correctly sorted before `/*`.

This was addressed by @pieh  in https://github.com/gatsbyjs/gatsby/pull/9719, but the fix seems incomplete.

### Steps to reproduce

1. Starting from the default Gatsby starter, create a `src/pages/app.js` and `src/pages/txt.js`, and create have a gatsby-node.js like this:
```js
exports.onCreatePage = ({ page, actions }) => {

  if (page.path.match(/^\/app/)) {
    createPage({
      ...page,
      matchPath: "/app/*",
    })
  }

  if (page.path.match(/^\/txt/)) {
    createPage({
      ...page,
      matchPath: "/txt/*",
    })
  }

  if (page.path === "/") {
    createPage({
      ...page,
      matchPath: "/*",
    })
  }
}
```

2. Run `gatsby build`

3. Check `.cache/match-paths.json`. It should look like this:

```json
[
    {
        "path": "/app/",
        "matchPath": "/app/*"
    },
    {
        "path": "/",
        "matchPath": "/*"
    },
    {
        "path": "/txt/",
        "matchPath": "/txt/*"
    }
]
```

Note: Changing the `/txt/` page's matchPath to `/app/*` doesn't change the ordering. It looks like matchPath isn't taken into account or is superseded by some other sorting.
```json
[
    {
        "path": "/app/",
        "matchPath": "/app/*"
    },
    {
        "path": "/",
        "matchPath": "/*"
    },
    {
        "path": "/txt/",
        "matchPath": "/app/*"
    }
]
```

### Expected result

`/txt/*` should be sorted before `/*` because it is more specific (more path segments)

### Actual result

`/txt/*` is sorted after `/*`

### Environment

  System:
    OS: macOS 10.14.5
    CPU: (8) x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
    Shell: 3.0.2 - /usr/local/bin/fish
  Binaries:
    Node: 12.3.1 - ~/.asdf/shims/node
    Yarn: 1.16.0 - ~/.asdf/shims/yarn
    npm: 6.9.0 - ~/.asdf/shims/npm
  Languages:
    Python: 2.7.15 - /Users/thanx/.asdf/shims/python
  Browsers:
    Chrome: 75.0.3770.142
    Safari: 12.1.1
  npmPackages:
    gatsby: ^2.13.39 => 2.13.39 
    gatsby-image: ^2.2.6 => 2.2.6 
    gatsby-transformer-sharp: ^2.2.4 => 2.2.4

wardpeet:
## Description
I've updated sorting for matchPaths so our most specific matchPath will be hit first.

I've added a unit test to show how it works. I just sort by the number of `/` in the string. If the amount of `/` match we order by index key (last one wins), I don't think we actually need the index check but it shows that we're handling those things too.

## Related Issues
Fixes #16098

### DIFF
--- a/packages/gatsby/src/bootstrap/requires-writer.js
+++ b/packages/gatsby/src/bootstrap/requires-writer.js
@@ -18,11 +18,37 @@ const getComponents = pages =>
     .map(pickComponentFields)
     .uniqBy(c => c.componentChunkName)
     .value()
+/**
+ * Get all dynamic routes and sort them by most specific at the top
+ * code is based on @reach/router match utility (https://github.com/reach/router/blob/152aff2352bc62cefc932e1b536de9efde6b64a5/src/lib/utils.js#L224-L254)
+ */
+const getMatchPaths = pages => {
+  const filteredPaths = []
+  pages.forEach((page, index) => {
+    if (page.matchPath) {
+      filteredPaths.push({
+        ...page,
+        index,
+        score: page.matchPath.split(`/`).length,
+      })
+    }
+  })
 
-const pickMatchPathFields = page => _.pick(page, [`path`, `matchPath`])
+  return filteredPaths
+    .sort((a, b) => {
+      // The higher the score, the higher the specificity of our matchPath
+      const order = b.score - a.score
+      if (order !== 0) {
+        return order
+      }
 
-const getMatchPaths = pages =>
-  pages.filter(page => page.matchPath).map(pickMatchPathFields)
+      // if specificty is the same we use the array index
+      return b.index - a.index
+    })
+    .map(({ path, matchPath }) => {
+      return { path, matchPath }
+    })
+}
 
 const createHash = (matchPaths, components) =>
   crypto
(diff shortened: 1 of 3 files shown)


packages/gatsby/src/bootstrap/requires-writer.js
```diff
@@ -18,11 +18,37 @@ const getComponents = pages =>
...
-  pages.filter(page => page.matchPath).map(pickMatchPathFields)
+  return filteredPaths
+    .sort((a, b) => {
+      // The higher the score, the higher the specificity of our matchPath
+      const order = b.score - a.score
+      if (order !== 0) {
+        return order
+      }
+
+      // if specificty is the same we use the array index
```
sidharthachatterjee: I see a typo fix incoming 😅

PR #17420, opened 2019-09-05
fix(gatsby): Allow overriding options for default instance of gatsby-plugin-page-creator

### LINKED ISSUE #17379: feat(gatsby): override default options of gatsby-plugin-page-creator
## Summary
When a user specifies a `gatsby-plugin-page-creator` plugin with a path pointing to `src/pages` we replace the config of our default `gatsby-plugin-page-creator` in gatsby core. This allows people to add `ignore` options.

A step in the right direction:
1) open https://github.com/gatsbyjs/gatsby/blob/abf888147de0c87d5c8560eaf2de18a0da4b5fca/packages/gatsby/src/bootstrap/load-plugins/load.js
2) find a `gatsby-plugin-page-creator` plugin with the path pointing to src/pages and temporary save the config in a variable
3) Replace the options on line (https://github.com/gatsbyjs/gatsby/blob/abf888147de0c87d5c8560eaf2de18a0da4b5fca/packages/gatsby/src/bootstrap/load-plugins/load.js#L217) with your variable.
4) write tests to make sure this works correctly.

### Basic example
```js
// in gatsby-config.js
{
  plugins: [{
    resolve: require.resolve(`gatsby-plugin-page-creator`),
    options: {
      path: `${__dirname}/src/pages`,
      ignore: [`**/examples/**`],
    },
  }]
}
```
Should change the default `gatsby-plugin-page-creator` config in 
https://github.com/gatsbyjs/gatsby/blob/abf888147de0c87d5c8560eaf2de18a0da4b5fca/packages/gatsby/src/bootstrap/load-plugins/load.js#L214-L222

Outcome should be
```js
  plugins.push(
    processPlugin({
      resolve: require.resolve(`gatsby-plugin-page-creator`),
      options: {
        path: slash(path.join(program.directory, `src/pages`)),
        ignore: [`**/examples/**`],
      },
    })
  )
```

### Motivation
We get a lot of requests to override the default `gatsby-plugin-page-creator` plugin. At this point we don't offer a good solution for this problem, there is a workaround but it's not fixing all use cases.

akshayymahajan:
## Description

Override the default options for gatsby-plugin-page-creator with options specified in gatsby-config.js

## Related Issues

Fixes #17379

### DIFF
--- a/packages/gatsby/src/bootstrap/load-plugins/load.js
+++ b/packages/gatsby/src/bootstrap/load-plugins/load.js
@@ -211,13 +211,27 @@ module.exports = (config = {}, rootDir = null) => {
     )
   })
   const program = store.getState().program
+
+  // default options for gatsby-plugin-page-creator
+  let pageCreatorOptions = {
+    path: slash(path.join(program.directory, `src/pages`)),
+    pathCheck: false,
+  }
+
+  if (config.plugins) {
+    const pageCreatorPlugin = config.plugins.find(
+      plugin => plugin.resolve === `gatsby-plugin-page-creator`
+    )
+    if (pageCreatorPlugin) {
+      // override the options if there are any user specified options
+      pageCreatorOptions = pageCreatorPlugin.options
+    }
+  }
+
   plugins.push(
     processPlugin({
       resolve: require.resolve(`gatsby-plugin-page-creator`),
-      options: {
-        path: slash(path.join(program.directory, `src/pages`)),
-        pathCheck: false,
-      },
+      options: pageCreatorOptions,
     })
   )
 
--- a/packages/gatsby/src/bootstrap/load-plugins/load.js
+++ b/packages/gatsby/src/bootstrap/load-plugins/load.js
@@ -222,7 +222,7 @@ module.exports = (config = {}, rootDir = null) => {
     const pageCreatorPlugin = config.plugins.find(
       plugin =>
         plugin.resolve === `gatsby-plugin-page-creator` &&
-        plugin.options.path.match(/src\/pages$/)
+        plugin.options.path === slash(path.join(program.directory, `src/pages`))
     )
     if (pageCreatorPlugin) {
       // override the options if there are any user specified options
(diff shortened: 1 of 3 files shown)


akshayymahajan:
This is my first attempt at contributing to gatsbyjs, would really appreciate some feedback..

karlhorky:
Hm, seems like this `1) added by themes can be shadowed` test is failing:

```
CypressError: Timed out retrying: Expected to find element: '[data-testid="title"]', but never found it.
```

https://circleci.com/gh/gatsbyjs/gatsby/217604?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Not sure what that could be, maybe unrelated to this? Wonder if master is currently passing...

akshayymahajan:
Yes @karlhorky. But when I try to run the same e2e test locally on my system, it passes all the tests without any issue..

muescha:
Somehow at a first glimpse the code not do the expected...

A) it overwrites the plugin options with default. But it should be otherwise. The default with the specific 

B) if there found the plugin with the same path it should not pushed a second time. 



As I understand the documentation of the plugin:

a) it is by default activated (pushed to plugins)
b) if it defined manually in plugins it uses the new options (overwriting options from default one) and no additional push of the plugin at the end


@wardpeet could confirm my assumptions?

packages/gatsby/src/bootstrap/load-plugins/load.js
```diff
@@ -211,13 +211,29 @@ module.exports = (config = {}, rootDir = null) => {
...
+  let pageCreatorOptions = {
+    path: slash(path.join(program.directory, `src/pages`)),
+    pathCheck: false,
+  }
+
+  if (config.plugins) {
+    const pageCreatorPlugin = config.plugins.find(
+      plugin =>
+        plugin.resolve === `gatsby-plugin-page-creator` &&
+        plugin.options.path.match(/src\/pages$/)
```
muescha: It should not match only the last chars but the full path, because all path for this plugin are full path

muescha: can be also `${__dirname}/myguides/src/pages`

akshayymahajan: Updated the condition to override the defaults only if the path matches <gatsby-site-root>/src/pages

sidharthachatterjee:
Published in `gatsby@2.15.11`

karlhorky:
Ah this feature seems to be broken with Gatsby `4.9.0` and the TypeScript configs: https://github.com/gatsbyjs/gatsby/discussions/34613#discussioncomment-2275411

PR #17426, opened 2019-09-05
feat(gatsby-plugin-manifest): Update manifest on navigation

### LINKED ISSUE #17255: [gatsby-plugin-manifest] Localized manifest not selected on navigation
## Description

When a localized page is loaded using the `<Link />` component, the localized manifest is not selected in `<head>`.

### Steps to reproduce

Demo repo: https://github.com/dkthehuman/gatsby-manifest-l10n-bug-demo
Netlify: https://affectionate-franklin-dc4bfb.netlify.com

1. Navigate to the Netlify demo (or run `gatsby build && gatsby serve` on the test repo)
2. Open DevTools and confirm that the manifest in `<head>` is `manifest-en.webmanifest` as expected.
3. Click on the "Spanish" link to navigate to `/es/`.
4. See that the manifest is still `manifest-en.webmanifest` rather than `manifest-es.webmanifest`.

### Expected result

The manifest in `<head>` should be `manifest-es.webmanifest`.

### Actual result

The manifest in `<head>` is still `manifest-en.webmanifest`.

### Considerations

In production builds (`gatsby build && gatsby serve`), the correct manifest is loaded when the localized page is the initial page or if you refresh on it.

In development builds (`gatsby develop`), the correct manifest is never loaded, regardless of whether you refresh or not.

### `gatsby-plugin-manifest` configuration

```javascript
{
      resolve: "gatsby-plugin-manifest",
      options: {
        name: "Hello World",
        short_name: "Test",
        start_url: "/",
        display: "minimal-ui",
        icon: "src/images/gatsby-icon.png",
        lang: "en",
        localize: [
          {
            start_url: "/es/",
            lang: "es",
            name: "Hola Mundo",
          },
        ],
      },
    },
```

### Environment

```

  System:
    OS: macOS 10.14.6
    CPU: (4) x64 Intel(R) Core(TM) i7-7567U CPU @ 3.50GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 12.9.1 - /usr/local/bin/node
    npm: 6.10.3 - /usr/local/bin/npm
  Languages:
    Python: 3.7.4 - /usr/local/opt/python/libexec/bin/python
  Browsers:
    Chrome: 76.0.3809.132
    Firefox: 68.0
    Safari: 12.1.2
  npmPackages:
    gatsby: ^2.13.73 => 2.13.73 
    gatsby-plugin-manifest: ^2.2.6 => 2.2.6 
  npmGlobalPackages:
    gatsby-cli: 2.7.33
```

dkthehuman:
## Description

With the addition of the `localize` option in `gatsby-plugin-manifest` (#13471), multiple manifests can be generated and served depending on the path of the URL. While the correct manifest is served on initial page load, the link to the manifest is never updated even if the user navigates to a page that should include a different manifest.

This PR hooks into `onRouteChange` in `gatsby-browser.js` in order to update the manifest as necessary.

## Related Issues

Fixes #17255

### DIFF
--- a/packages/gatsby-plugin-manifest/src/gatsby-browser.js
+++ b/packages/gatsby-plugin-manifest/src/gatsby-browser.js
@@ -1,5 +1,6 @@
 /* global __MANIFEST_PLUGIN_HAS_LOCALISATION__ */
 import { withPrefix as fallbackWithPrefix, withAssetPrefix } from "gatsby"
+import getManifestForPathname from "./get-manifest-pathname"
 
 // when we don't have localisation in our manifest, we tree shake everything away
 if (__MANIFEST_PLUGIN_HAS_LOCALISATION__) {
@@ -13,22 +14,5 @@ if (__MANIFEST_PLUGIN_HAS_LOCALISATION__) {
     if (manifestEl) {
       manifestEl.setAttribute(`href`, withPrefix(manifestFilename))
     }
-  }
-
-  function getManifestForPathname(pathname, localizedApps) {
-    const defaultFilename = `manifest.webmanifest`
-    if (!Array.isArray(localizedApps)) {
-      return defaultFilename
-    }
-
-    const appOptions = localizedApps.find(app =>
-      pathname.startsWith(app.start_url)
-    )
-
-    if (!appOptions) {
-      return defaultFilename
-    }
-
-    return `manifest_${appOptions.lang}.webmanifest`
   }
 }
--- a/packages/gatsby-plugin-manifest/src/gatsby-browser.js
+++ b/packages/gatsby-plugin-manifest/src/gatsby-browser.js
@@ -1,0 +1,26 @@
+import { withPrefix as fallbackWithPrefix, withAssetPrefix } from "gatsby"
+
+const withPrefix = withAssetPrefix || fallbackWithPrefix
+
+exports.onRouteUpdate = function({ location }, pluginOptions) {
+  const { localize } = pluginOptions
+  const manifestFilename = getManifestForPathname(location.pathname, localize)
+
+  const manifestEl = document.head.querySelector(`link[rel="manifest"]`)
+  if (manifestEl) {
+    manifestEl.setAttribute(`href`, withPrefix(manifestFilename))
+  }
+}
+
+function getManifestForPathname(pathname, localizedApps) {
+  let suffix = ``
+  if (Array.isArray(localizedApps)) {
+    const appOptions = localizedApps.find(app =>
+      RegExp(`^${app.start_url}.*`, `i`).test(pathname)
+    )
+    if (appOptions) {
+      suffix = `_${appOptions.lang}`
+    }
+  }
+  return `manifest${suffix}.webmanifest`
+}
--- a/packages/gatsby-plugin-manifest/src/gatsby-browser.js
+++ b/packages/gatsby-plugin-manifest/src/gatsby-browser.js
@@ -1,29 +1,34 @@
+/* global __MANIFEST_PLUGIN_HAS_LOCALISATION__ */
 import { withPrefix as fallbackWithPrefix, withAssetPrefix } from "gatsby"
 
-const withPrefix = withAssetPrefix || fallbackWithPrefix
+// when we don't have localisation in our manifest, we tree shake everything away
+if (__MANIFEST_PLUGIN_HAS_LOCALISATION__) {
+  const withPrefix = withAssetPrefix || fallbackWithPrefix
 
-exports.onRouteUpdate = function({ location }, pluginOptions) {
-  const { localize } = pluginOptions
-  const manifestFilename = getManifestForPathname(location.pathname, localize)
+  exports.onRouteUpdate = function({ location }, pluginOptions) {
+    const { localize } = pluginOptions
+    const manifestFilename = getManifestForPathname(location.pathname, localize)
 
-  const manifestEl = document.head.querySelector(`link[rel="manifest"]`)
-  if (manifestEl) {
-    manifestEl.setAttribute(`href`, withPrefix(manifestFilename))
-  }
-}
-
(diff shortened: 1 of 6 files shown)


dkthehuman:
Hm, not sure why the e2e test is failing since I use the asset prefix when setting the manifest: https://github.com/gatsbyjs/gatsby/pull/17426/files#diff-9d857a76b97041cf4aefbf1950cd3c9fR11

packages/gatsby-plugin-manifest/src/gatsby-browser.js
```diff
@@ -0,0 +1,26 @@
...
+  if (manifestEl) {
+    manifestEl.setAttribute(`href`, withPrefix(manifestFilename))
+  }
+}
+
+function getManifestForPathname(pathname, localizedApps) {
+  let suffix = ``
+  if (Array.isArray(localizedApps)) {
+    const appOptions = localizedApps.find(app =>
+      RegExp(`^${app.start_url}.*`, `i`).test(pathname)
```
moonmeister: I haven't tested this regex but just crossed my mind that we need to be super careful with string matching. For instance, there's a valid country code 'an'. this could erroneously be detected if it detects the `m(an)ifest.webmanifest`. Like I said, not sure if it does but we should do some thorough testing.

moonmeister: i'm guessing this one could be an issue:

```
Type: language
Subtag: st
Description: Southern Sotho
Added: 2005-10-16
Suppress-Script: Latn
```

packages/gatsby-plugin-manifest/src/gatsby-browser.js
```diff
@@ -0,0 +1,26 @@
...
+
+  const manifestEl = document.head.querySelector(`link[rel="manifest"]`)
+  if (manifestEl) {
+    manifestEl.setAttribute(`href`, withPrefix(manifestFilename))
+  }
+}
+
+function getManifestForPathname(pathname, localizedApps) {
+  let suffix = ``
+  if (Array.isArray(localizedApps)) {
```
moonmeister: with this if statement and the one below I'd recommend you put the happy path outside the if. So, negate both these and return early. it helps make the code more readable to keep your happy path out of if statements when possible imo.

packages/gatsby-plugin-manifest/src/gatsby-browser.js
moonmeister: I am way to tired to know if these last two comments are genius or rubbish. I have a feeling their rubbish and it's not going to be a problem...please read and disregard or fix as necessary. I'll re read when I'm not wiped. :sob:

dkthehuman: Hahah, get some rest!

This regex is the same as the one already used in `gatsby-ssr.js`: https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-manifest/src/gatsby-ssr.js#L22

I might be misunderstanding what you meant, but this line doesn't have anything to do with country codes or the name of the manifest file. The line checks to see if the path of the URL (e.g. `/es/`) matches the `start_url` specified in the localized app's manifest.

packages/gatsby-plugin-manifest/src/gatsby-browser.js
dkthehuman: I wanted to avoid repeating the string `manifest.webmanifest` in the function, but I wrote it out using the "return early" style to compare:

```javascript
function getManifestForPathname(pathname, localizedApps) {
  const defaultFilename = `manifest.webmanifest`
  if (!Array.isArray(localizedApps)) {
    return defaultFilename
  }

  const appOptions = localizedApps.find(app =>
    RegExp(`^${app.start_url}.*`, `i`).test(pathname)
  )
  if (!appOptions) {
    return defaultFilename
  }
  
  const suffix = `_${appOptions.lang}`
  return `manifest${suffix}.webmanifest`
}
```

I slightly prefer the current implementation because it avoids repeating the filename, but let me know if you'd like me to switch because I don't feel strongly about it. 😎

dkthehuman:
Hoping I can also get some help regarding this ESLint warning I get when developing on the plugin as a local plugin:
```
Module Warning (from ./node_modules/eslint-loader/index.js):

/Users/dskang/Code/dkthehuman.com/plugins/gatsby-plugin-manifest-multiple/gatsby-browser.js
  1:1  warning  'use strict' is unnecessary inside of modules  strict
```

Babel adds "use strict" to the top of every file and ESLint (I'm using Gatsby's default config) complains about it. Have you encountered this before when working on plugins as a local plugin?

packages/gatsby-plugin-manifest/src/gatsby-browser.js
moonmeister: I know it's a little more verbose, but I think it's more readable. :man_shrugging:

dkthehuman: Done!

wardpeet:
I'm pretty confident that this is ready to be shipped.
- fixed default localized manifest & sw support (gatsby-plugin-offline wasn't caching the correct file)
- Only add gatsby-browser code when necessary

I think we can iterate on this if the startWith/regex doesn't suffice.

wardpeet:
@moonmeister, yeah offline plugin has issues with it. I think it's good to just keep the default one manifest.webmanifest and only suffix the ones that are in the localized string.

Thanks for pointing out gatsby-ssr, I forgot about it and looks like I didn't have any issue because of rehydration XD.

The latest offline plugin doesn't prefetch manifest anymore when it's not found so the issue looks like it doesn't exist anymore. I'll bring that feature back as it's a regression.

I updated tests and I think, it's ready to go.

moonmeister:
What about caching the non-default manifests?

PR #17436, opened 2019-09-06
docs: Move `__tests__` dir out of `src`

wataruoguchi:
## Description

I've got the realization that it's not intuitive to have `__tests__` dir in `src` dir while having the reply: https://github.com/gatsbyjs/gatsby/pull/17077#pullrequestreview-279460456

I sort of revert my previous change PR:17077 and move `__tests__` dir to project root dir instead, so people don't need extra work to handle the test dir.

## Related Issues

Not an issue but PR: https://github.com/gatsbyjs/gatsby/pull/17077


LekoArts:
> I've got the realization that it's not intuitive to have __tests__ dir in src dir

Sadly I have to disagree with this. The general (best) practice is to keep your tests beside your actual code and I think we should keep it that way as most of the other users of Jest do it like this.

wataruoguchi:
Thanks @LekoArts for reviewing & describing it. I understand the intention. Closing this PR.

PR #17442, opened 2019-09-06
feat(gatsby-plugin-netlify-cms): new option customizeWebpackConfig

emanueleperuffo:
## Description

I needed a custom Webpack plugin to be able to use some of my components in the preview pane. I ended up modifying the Webpack configuration for the CMS. This pull request allows anyone to pass an optional function to the plugin to allow them to customize the configuration of Webpack.

In addition, I suggest that we include in `plugins:` the same `definePlugin` present the configuration of `getConfig()`.


sidharthachatterjee:
Cc @erquhart

wardpeet:
@emanueleperuffo Why don't you augment the gatsby webpack config instead? We're a bit worried if we add custom webpack configs to Netlify plugin we will get more bugs as people tend not to know how do configure webpack. Also, it re-uses the gatsby webpack config which might break other things inside Netlify plugin when modifying it badly.

emanueleperuffo:
@wardpeet I already augment the webpack config of gatsby with my own definitions like the code below.

I absolutely need that `plugins.define(envForDefinePlugin)` because my components rely on it. Without adding that to the webpack config of netlify cms my components are not getting the definitions in `plugins.define(envForDefinePlugin)`.

Also to note, without any of my changes in this PR, none of the definitions in the gatsby files *.env.development* and *.env.production* are compiled for netlify cms. If someone rely on those definitions (which are part of gatsby) in their components they will get errors. https://www.gatsbyjs.org/docs/environment-variables/

My PR wanted to create a solution for other problems as well because someone may need some webpack plugin injected in the configuration, or some other loader, etc... I expect that people should know what to do when changing the configuration of webpack like you do. If this is your main concern, will it be acceptable to rename that function to something like *dangerouslyReplaceWebpackConfig*?

```javascript
const fs = require('fs');
const path = require('path');
const dotenv = require('dotenv');
const DuplicatePackageCheckerPlugin = require('duplicate-package-checker-webpack-plugin');

const env = process.env.APP_ENV || process.env.NODE_ENV || 'development';
const envBuffer = fs.readFileSync(path.resolve(__dirname, `.env.${env}`));

const envForDefinePlugin = {
  ...Object.entries(dotenv.parse(envBuffer)).reduce((res, [k, v]) => ({...res, [`process.env.${k}`]: JSON.stringify(v)}), {}),
};

exports.onCreateWebpackConfig = (args) => {
  const {getConfig, actions, plugins} = args;

  const config = getConfig();

  actions.replaceWebpackConfig({
    ...config,

    plugins: [
      plugins.define(envForDefinePlugin), // Overrides the default .env.production or .env.development to custom environments like .env.staging
      ...config.plugins,
      new DuplicatePackageCheckerPlugin(),
    ],
  });
};
```

wardpeet:
@emanueleperuffo sorry for my late response, I've tried overriding env and found the issue you were describing.

onCreateWebpackConfig inside gatsby-node doesn't augment the config on time for Netlify to use. It's the bad architecture of `gatsby-netlify-cms` that is the problem as it should use childCompiler instead. Nonetheless, I do feel like your option fixes it without doing a whole rewrite.

wardpeet:
Published in `gatsby-plugin-netlify-cms@4.1.18`

PR #17444, opened 2019-09-06
plugin(gatsby-plugin-emotion): Add note about supported SSR APIs

amitnovick:


packages/gatsby-plugin-emotion/README.md
```diff
@@ -9,6 +9,15 @@ rendering.
...
 guide](https://emotion.sh/docs/migrating-to-emotion-10#incremental-migration) for more information on how to upgrade.
 
+## What this plugin does
+
+This plugin lets you leverage Emotion's server-side rendering (SSR) on the following Emotion APIs:
+
+- `css` component prop from `@emotion/core` with the `jsx` babel transformation
+- `styled` function with `@emotion/styled`
+
+NOTE: Unlike the APIs mentioned above, the `css` function from `emotion` **will not leverage SSR using this plugin**
```
wardpeet: Can you put in an example? I'm not 100% sure what you mean with this line.

packages/gatsby-plugin-emotion/README.md
```diff
@@ -14,26 +14,47 @@ guide](https://emotion.sh/docs/migrating-to-emotion-10#incremental-migration) fo
...
+
+const Header = () => (
+  <h2
+    className={css`
+      color: hotpink;
+    `}
+  >
+    Welcome
+  </h2>
+) // ✖ SSR
```
amitnovick: @wardpeet :
> Can you put in an example? I'm not 100% sure what you mean with this line.

I added some example snippets
Is it easier to understand now?

wardpeet: Oh yeah definitely, Thank you for adding these.

marcysutton:
I don't see a GitHub issue related to this PR - can you expand on what behavior this is documenting? CSS is inherently a client-side technology, but Gatsby bundles styles at build time. Are you saying there is a build-time difference with Emotion `styled` vs `css`? Because that is not what I've found in my experience.

amitnovick:
> I don't see a GitHub issue related to this PR - can you expand on what behavior this is documenting? CSS is inherently a client-side technology, but Gatsby bundles styles at build time. Are you saying there is a build-time difference with Emotion `styled` vs `css`? Because that is not what I've found in my experience.

Hi, I am using the following packages in my Gatsby project:
- `gatsby@2.3.32`
- `gatsby-plugin-emotion@4.1.5
- `emotion@10.0.17`

and added the Gatsby plugin to my `gatsby-config.js` as per instructions:
```js
module.exports = {
  // ...
  plugins: [
    // ...
    {
      resolve: `gatsby-plugin-emotion`,
      options: {
        // Accepts all options defined by `babel-plugin-emotion` plugin.
      },
    },
  ],
}
```

And the behaviour is the following:
1.  
```jsx
import { css } from "@emotion/core"
const Header = () => (
  <h2
    css={css`
      color: hotpink;
    `}
  >
    Welcome
  </h2>
) // ✅ SSR
```

2.
 ```jsx
import styled from "@emotion/styled"
const Header = styled.h2`
  color: hotpink;
` // ✅ SSR
```

3.
 ```jsx
import { css } from "emotion"
const Header = () => (
  <h2
    className={css`
      color: hotpink;
    `}
  >
    Welcome
  </h2>
) // ✖ SSR
```

muescha:
`✖ SSR` this cross should be more clear and add some words that this is not working. on my browser the `✖` looks like a star and not like the big red cross

muescha:
is there a battle for the error between:
- `import { css } from "emotion"` vs `import { css } from "@emotion/core"`

or between:
- `className={css` vs `css={css` 


or both only in combination?

amitnovick:
@muescha
Thanks for providing appropriate suggestions :heart: . I went ahead and implemented them in [this last commit](https://github.com/gatsbyjs/gatsby/pull/17444/commits/d3ec3e2ecaeec45264574e43d167e700a4ff1617).

> is there a battle for the error between:
> 
>     * `import { css } from "emotion"` vs `import { css } from "@emotion/core"`
> 
> 
> or between:
> 
>     * `className={css` vs `css={css`
> 
> 
> or both only in combination?
> 

Not sure what you mean by "battle for the error".
My usage of `css` from `emotion` is strictly by passing the string value it generates to the `className` prop of a React component.

muescha:
since you changed 2 different things in the example (compared to the `css` above i am confused which change would cause the error...

amitnovick:
> since you changed 2 different things in the example (compared to the `css` above i am confused which change would cause the error...

Sorry but I can't understand what you mean here. The two example snippets I posted in this thread of the `css` function from `emotion` not working are the same: [1](https://github.com/gatsbyjs/gatsby/pull/17444#issuecomment-530064565) [2](https://github.com/gatsbyjs/gatsby/pull/17444#issuecomment-530135797)

muescha:
you changed between your example `1.`:

```javascript
import { css } from "@emotion/core"
                     ~~~~~~~~~~~~~

const Header = () => (
  <h2
    css={css`
    ~~~
      color: hotpink;
    `}
  >
    Welcome
  </h2>
)
```

and your example `3.`:

```javascript
import { css } from "emotion"
                     ~~~~~~~

const Header = () => (
 <h2
   className={css`
   ~~~~~~~~~
     color: hotpink;
   `}
 >
   Welcome
 </h2>
) 
```

exactly this 2 things:

a) first change the import:  `@emotion/core` to `emotion`
b) second change tag name:  `css=` to `className=`


the question is: when the server side rendering fails?

1) only with the `a)` change -> because of import of a different module?
2) only with the `b)` change -> because of the different tag name
3) only with the `a)` AND `b)` change -> because of different module AND different tag name


if it is the `2)` change then we should change the example to number `4.`:

```javascript
import { css } from "@emotion/code"

const Header = () => (
 <h2
   className={css`
     color: hotpink;
   `}
 >
   Welcome
 </h2>
) 
```

marcysutton:
Thanks for the comment, however I think this would be better solved in the plugin itself. The README mentions it should be used with `@emotion/core`, yet that last example is using `emotion`. Is there a reason why you'd want to deviate from the recommended install?

DSchau:
Hey @amitnovick!

Thanks for opening this PR.

I see what you're doing here, and I very much appreciate it!

If I'm understanding correctly, you've personally encountered an issue where you were using the `emotion` package rather than the recommended `@emotion/core` package. The former of these will not be server-side rendered using the latest version(s) of gatsby-plugin-emotion, which does make some sense because we don't document installing/using the bare `emotion` package.

That being said -- I think one of the most helpful things contributors/users can do is improve documentation when they've personally experienced a frustrating/confusing error or issue. However, in this case, I don't think this really improves comprehension. Our installation steps specifically list using the `@emotion/core` and `@emotion/styled` packages, rather than the `emotion` package as you've noted here.

As a user reading this documentation, I would personally wonder why `emotion` is even referenced, and may even grow more confused, which is sort of the opposite of what we'd hope would occur!

From this basis, I'd prefer we close this at this time and not merge this change, just to ensure that other users don't grow more confused with this change. If you can think of a slightly more user-friendly way to introduce a change that accomplishes a similar effect, we'd love to hear it!

Thanks again, and thanks for contributing to Gatsby 💜

muescha:
@DSchau i found the addition to the docs very useful. 
Also the "so not do this" part is very helpful to avoid pitfalls. 
Only I missed a better explanation what are the in the form:

You have done "this" - but the right syntax is "this" because of "that"

Maybe @marcysutton has a better idea how to write it in a good way that helps to avoid pitfalls

PR #17446, opened 2019-09-06
fix(gatsby-plugin-offline): use base path instead of path prefix

davidbailey00:


sidharthachatterjee:
Okay, that tests seems to be failing pretty consistently 😞

sidharthachatterjee:
Published in `gatsby-plugin-offline@3.0.6`

PR #17447, opened 2019-09-06
feat(www): Scroll sidebar to current page

### LINKED ISSUE #13602: [docs] Current page not always in view in sidebar when navigating around docs
just a side note: when i jump to a doc then in the doc hierarchy is bolded. 

but i miss here that the bolded selected doc is also moved to be visible area. i always need to scroll down to find the bolded text

_Originally posted by @muescha in https://github.com/gatsbyjs/gatsby/issues/13475#issuecomment-486307738_

kevee:
## Description

Scrolls the docs sidebar to the current page, helpful if the sidebar is expanded and the current page is buried.

Currently it does so with *every* page, even as you navigate around. I'm not sure whether this is desired behavior or not... it might be better to only scroll if the user opens the page initially.

## Related Issues

Fixes #13602


www/src/components/sidebar/item.js
```diff
@@ -14,6 +14,19 @@ const isItemActive = (activeItemParents, item) => {
 }
 
 class Item extends React.PureComponent {
+  constructor(props, context) {
+    super(props, context)
+    this.itemRef = React.createRef()
```
muescha: I would name it `activeItemRef` to make clear it is not on every item

wardpeet:
Can you give an example of a page where this behaviour occurs? Sorry, by reading the issue and/or PR i'm not sure what pages we're talking about.

marcysutton:
@wardpeet it happens on all of the docs pages with the sidebar, from my experience. We improved UX a bit with a breadcrumb, but the fact that the sidebar doesn't scroll to the relevant, open position when you deep link to a doc contributes to a general feeling of being lost or overwhelmed.

www/src/components/sidebar/item.js
```diff
@@ -14,6 +14,19 @@ const isItemActive = (activeItemParents, item) => {
...
+  }
+
+  componentDidMount() {
+    if (this.itemRef.current) {
+      this.itemRef.current.scrollIntoView({
```
marcysutton: In testing, this doesn't seem to work reliably - some items will scroll, while others won't. When I log the itemRef it's accurate, but the scrolling doesn't occur. If I add a short delay with `setTimeout`, it does work...perhaps there's a cleaner way to make it more consistent?

muescha: maybe: with `useEffect` and empty array `,[])` it runs also only one time ...

muescha: oh - it's a `class` - no hooks

kevee: Right, this is probably an issue with the DOM node sometimes not being available. Should we convert it to a function component with hooks?

www/src/components/sidebar/item.js
kevee: Agreed, done.

www/src/components/sidebar/item.js
```diff
@@ -13,76 +13,75 @@ const isItemActive = (activeItemParents, item) => {
...
-      this.activeItemRef.current.scrollIntoView({
+  useEffect(() => {
+    if (!didScroll && activeItemRef.current) {
+      activeItemRef.current.scrollIntoView({
         block: `center`,
       })
+      setDidScroll(true)
     }
-  }
+  })
```
muescha: If there ist an empty dependency array then useEffekt will run only once
```suggestion
  },[])
```

muescha: Then you not need `didScroll`

marcysutton:
I think I figured out what's going on: if a different subsection is expanded than the one you're on (saved from a previous visit) and you navigate to a page, it doesn't scroll you to the correct position. Steps to reproduce:

- Navigate to a subsection like `Querying Your Data with GraphQL: Why Gatsby Uses GraphQL`
- Navigate to another subsection like `MDX: Writing Pages`
- Navigate elsewhere, such as the Contributing docs
- Go back to `MDX: Writing Pages`: the sidebar isn't scrolled to the right spot

@DSchau @fk any input on how to handle that scenario? Should we collapse sections on page load that aren't relevant to the one you're in?

fk:
I gave this a quick shot on localhost, and can reproduce your issue like 20% of times @marcysutton — worked for me 80% of times. In general, things seem to work 50% of the time … 🤔 🤷‍♂ 

I remember we are "saving" the sidebar scroll position in https://github.com/gatsbyjs/gatsby/blob/master/www/src/components/sidebar/scrollbar-position-provider.js, that seems like it has the potential to mess up things? This PR's functionality seems like it could/should replace the aforementioned?

gillkyle:
I've been trying to test this locally myself and am not quite sure what behavior I should be expecting. It seems like the initial page load is scrolling to the correct place when I open a new tab and paste in a link, but if the link I'm pasting is lower in the sidebar -- like Commands (Gatsby CLI) for example -- it won't scroll because the item in the sidebar is too low on the screen and won't open to it.

scrol-behavior

I'm also seeing something interesting with the sidebar starting to slow down and making it so the links I click on in the sidebar aren't always taking me to the right doc (which I'm not sure is an issue of running the site in develop vs the production build, or if it's a result of the `useEffect` or what).

scroll-behavior

gillkyle:
I've been investigating this more and found the usage of a [callback ref]() to get the initial page load to work every time. The struggle has been getting subsequent page changes (change of `location.pathname`) to `scrollIntoView` again.

```
const itemRef = useCallback(
  async node => {
    if (item.link === activeItemLink.link && node !== null) {
      console.log(node)
      // await new Promise(f => setTimeout(f, 1000))
      node.scrollIntoView({ block: `center` })
    }
  },
  [location.pathname]
)
```

What seems to happen is that the ref for the `Item` isn't updated fast enough to (not sure what the proper terminology is..) have the `scrollIntoView` method being called. It's illustrated in these two recordings, the first shows the above code, and the second shows the above code with the commented line uncommented -- thus waiting 1 second each time there's a page change and calling `scrollIntoView`, which works each time.

no-await

await

I think we could just call this on page load to center the sidebar at the right article which would be an improvment. The next behavior of being able to move the sidebar each time you navigate once you're already on the site is the trickier part.

gillkyle:
Here's a recording showing that the initial page load gets the user to the right place with the sidebar open at least: 

load

muescha:
maybe it is possible to wait for some event so there is no need for the `setTimeout`

gillkyle:
I just opened #19556 to fix this with the solutions I researched. I'll close this now and we can continue discussion there.

PR #17449, opened 2019-09-06
chore(starters): add gatsby_starter_portfolio

mohanmonu777:


docs/starters.yml
```diff
@@ -3639,6 +3639,7 @@
     - Ready made Components
     - Responsive Design
     - Syntax Highlight in Code.
+- url: https://gastby-starter-organization.netlify.com/
```
sidharthachatterjee: This should be removed?

mohanmonu777: Nope

sidharthachatterjee: This is unrelated to your entry, correct?

PR #17456, opened 2019-09-06
chore(gatsby-cli): changed the new command gist

LOLdevelopr:
## Description
In the CLI, the new command gist is `gatsby new gatsby-site` which is not so beginner  friendly as many will be fond to create their project with `gatsby-site` name, instead added a template mark to show it as it is customizable

## Related Issues
NA

### DIFF
--- a/packages/gatsby-cli/README.md
+++ b/packages/gatsby-cli/README.md
@@ -32,7 +32,7 @@ You can also use the `package.json` script variant of these commands, typically 
 
 ### `new`
 
-`gatsby new <my-gatsby-site>`
+`gatsby new <my-gatsby-site> [<gatsby-starter>]`
 
 See the [Gatsby starters docs](https://www.gatsbyjs.org/docs/gatsby-starters/)
 for more.
--- a/packages/gatsby-cli/README.md
+++ b/packages/gatsby-cli/README.md
@@ -32,7 +32,7 @@ You can also use the `package.json` script variant of these commands, typically 
 
 ### `new`
 
-`gatsby new gatsby-site`
+`gatsby new <my-gatsby-site>`
 
 See the [Gatsby starters docs](https://www.gatsbyjs.org/docs/gatsby-starters/)
 for more.
--- a/packages/gatsby-cli/README.md
+++ b/packages/gatsby-cli/README.md
@@ -34,7 +34,7 @@ You can also use the `package.json` script variant of these commands, typically 
 
 `gatsby new <my-gatsby-site> [<gatsby-starter>]`
 
-See the [Gatsby starters docs](https://www.gatsbyjs.org/docs/gatsby-starters/)
+See the [Gatsby starters docs](https://www.gatsbyjs.org/docs/starters/)
 for more.
 
 ### `develop`


packages/gatsby-cli/README.md
```diff
@@ -32,7 +32,7 @@ You can also use the `package.json` script variant of these commands, typically
 
 ### `new`
 
-`gatsby new gatsby-site`
+`gatsby new <my-gatsby-site>`
```
muescha: add starter parameter

```suggestion
`gatsby new <my-gatsby-site> [<gatsby-starter>]`
```

packages/gatsby-cli/README.md
```diff
@@ -32,7 +32,7 @@ You can also use the `package.json` script variant of these commands, typically
...
 
 See the [Gatsby starters docs](https://www.gatsbyjs.org/docs/gatsby-starters/)
```
muescha: should point to the docs instead to the list of starters 

```suggestion
See the [Gatsby starters docs](https://www.gatsbyjs.org/docs/starters/)
```

sidharthachatterjee:
Thank you for taking the time to open this @LOLdevelopr 👍 

Unfortunately, we can't accept this because 

- <site-name> syntax can be confusing to folks who are new 
- <site-name> isn't very copy pastable either

Nonetheless, we could add a small note at the bottom there mentioning that _gatsby-site_ is arbitrary. 

Thank you again!

LOLdevelopr:
@sidharthachatterjee  is there any way to make it more beginner friendly. ? 
Cause now the current is bit confusing too and i tried to solve that only.
Maybe foot note can be a solution with examples about how to use it

LOLdevelopr:
Like this 
`gatsby new <site-name> <starter-url>`

**examples**

- `gatsby new my-app https://github.com/gatsbyjs/gatsby-starter-blog` 
- `gatsby new my-app https://github.com/netlify-templates/gatsby-starter-netlify-cms`

muescha:
I think a notation where I see what are the options is more beginner friendly then an command where I have to guess and look on other docs before it is clear

Ever command line program has a notion of options

muescha:
In the readme pay fee the option "gatsby-site" looks like a *must have* parameter because of the Gatsby word in it

muescha:
I would go to explain all commnsa with this command line syntax: http://docopt.org/

LOLdevelopr:
Yeah actually the current readme command is not so clear for a first time user. it doesnt represent that `gatsby-site` is an arbitrary or can be overridden. 
Other CLIs like
-  [`webpack-cli`](https://github.com/webpack/webpack-cli/blob/next/packages/init/README.md#cli-via-webpack-cli) *(check the scaffold part)* 
-  [`preact-cli`](https://github.com/preactjs/preact-cli#usage)  

these all use these kind of templating. It seems these can be more friendly

muescha:
Maybe open a new issue for general discussion about the command line syntax for Gatsby docs

LOLdevelopr:
yeah we can have a discussion there then !

marcysutton:
I agree, those instructions could be clarified. An issue is indeed the best place to discuss and figure out a good solution! Thanks so much for bringing it up 😀

LOLdevelopr:
@marcysutton here is the issue https://github.com/gatsbyjs/gatsby/issues/17473

Issue #16458, opened 2019-08-08
Resource Loading Resilience tests don't run 


sidharthachatterjee:
It seems like the tests at https://github.com/gatsbyjs/gatsby/tree/master/e2e-tests/production-runtime/cypress/integration/resource-loading-resilience.js aren't running correctly at the moment!

### Expected behaviour 
`runBlockedScenario` needs to run for every describe block under `runSuiteForPage` 

### Current behaviour 
After resources are blocked, no tests are run and the suite completes with everything green

Darking360:
I can take a look during this week 🚀  I'll let you know as soon as I have progress on it 🙏

PR #17458, opened 2019-09-06
fix(gatsby): Fixing Cypress blocked scenarios tests ⚙︎

Darking360:
# Cypress blocked scenarios tests ⚙︎

## Description

This is related to an issue opened by @sidharthachatterjee, the problem is related with the Cypress tests here https://github.com/gatsbyjs/gatsby/tree/master/e2e-tests/production-runtime/cypress/integration/resource-loading-resilience.js, the thing is that one different resource is being blocked for each test round, so for example blocking `Index` should run all the `Loads Index`, `Navigates to second page` and so on, the problem is that before it only restore the blocked resources but failed to run the tests all over to the other page cases like `Index`, `Page-2` and `404`.

Passing down the `tag` of the scenario brings a better reading for the test `describe` and it `statements` run every time, also having the `runTests` inside an `it` case didn't add the dynamic test cases correctly, so refactoring this into a `describe` statement running an async `before` function before each dynamic `it` that is going to be generated we guarantee that the resources are being restored and blocking the next case.

Now when you see Cypress tests, you'd see around 70 test cases being generated which is the correct behavior. Now the problem is that there are some test cases failing because of the blocking behavior, if we're blocking the `Index` for example, we would not get navigation so we cannot assert on navigation item for page 2 for example 🤔 I don't know if I should fix the test cases too, or only fix the Cypress problem in this PR, waiting on @sidharthachatterjee response for this 🚀 

## Related Issues

https://github.com/gatsbyjs/gatsby/issues/16458

### Attachments 

Here is the [link](https://dashboard.cypress.io/#/projects/is8aoq/runs/1030/specs) for the Cypress run on the tests for all scenarios.

Some attachments to show Cypress tests running the test cases:

image

image


wardpeet:
Oh man, this is awesome! :fire: Great job, we definitely should fix it in this PR. If you're up for it please do. If not I can take over but I would love that you do it and if you need any assistant just shout.

Thanks a bunch!

Darking360:
Hey @wardpeet sure no problem, as soon as I have it ready I'll let you know, or shout for help hehe thank you :rocket:

wardpeet:
Awesome! :fire: Thank you! :pray:

Darking360:
Hey @wardpeet how are you? Sorry for the delay hehe just finished the changes, I was fixing some duplicated code I had 😅 What I did is basically add the expected behavior in a little object of options to the function that runs the tests for each scenario, the challenge here is that each time we block a resource for `page, `query results`, `page assets` and so on, we get dynamic tests which are the `why` of this PR but we need to assert different scenarios results for blocking the Index, the second page and the 404, for example, we have different behavior there since we test it with `cy.visit()`. 

That being said, I think I found the pattern of those adding an object with some variables for link working or not, if it should stay (if the page was removed from page template asset for example) or not and for 404 safe behavior (due to the `cy.visit()`). All 72 dynamic generated tests are working now for both `online` and `offline` behaviors; also included conditionals to toggle some variables here depending on offline or not.

All the asserts work depending on the blocked or not resource and based on the variables from the options object, I pull a helper for asserting on the navigation from a click element called `assertOnNavigate` to reduce duplicated code in general. I think that'd be it, let me know your thoughts on this one @wardpeet and thank you 🙏

Darking360:
I'm going to close this PR and open it in another one, to make it count on Hacktober fest 🙈

PR #17466, opened 2019-09-07
chore(showcase): Remove oern.tv

lannonbr:
## Description

@cadejscroggins just a heads up. The Gatsby Site Showcase Validator found that this site returned the GitHub Pages 404 so it seems the site is no longer available. I'm removing it from the showcase as such.


muescha:
The GitHub repo ist also gone

PR #17484, opened 2019-09-08
Add Fathom Analytics to Gatsby Starter Blog

bcye:
## Description
This PR adds the functionality to add the [Fathom Analytics](https://usefathom.com) tracking script to your site created with the gatsby-starter-blog

### DIFF
--- a/starters/blog/gatsby-config.js
+++ b/starters/blog/gatsby-config.js
@@ -53,6 +53,15 @@ module.exports = {
         //trackingId: `ADD YOUR TRACKING ID HERE`,
       },
     },
+    {
+      resolve: 'gatsby-plugin-fathom',
+      options: {
+        // Fathom server URL. Defaults to `cdn.usefathom.com`
+        // trackingUrl: 'your-fathom-instance.com',
+
+        // siteId: 'ADD YOUR SITE ID HERE'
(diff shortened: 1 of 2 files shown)


sidharthachatterjee:
The official starters are meant to be unopinionated and as minimal as possible so while I think Fathom is awesome, I don't think adding it by default is a good idea. 

Nonetheless, thank you for taking the time to open this! 💜

bcye:
So why is Google Analytics getting an exception? Because of its size?

sidharthachatterjee:
@bcye That's actually a great question. I'd forgotten that that's included by default! I reckon it's there since it's popular enough to be a default but I wouldn't be opposed to removing it to be honest. 

Curious what the rest of @gatsbyjs/core thinks

bcye:
Another option might be to simply document in the README how to install either into the project (or maybe add it as a blog post or a page in the docs something along the lines of "How to add analytics to your Gatsby website").

bcye:
Would be interested in working on that if that's something wanted.

bcye:
So thumbs up means yes?

bcye:
There is already a doc page about this: https://gatsbyjs.org/docs/adding-analytics/

PR #17492, opened 2019-09-08
fix(gatsby): Handle Windows drive letter casing correctly

### LINKED ISSUE #16721: gatsby v2.13.65 -> v2.13.66 introduces a lot of warnings in Windows due to path changes
## Description

I am on Windows 10. After updating from Gatsby v2.13.65 -> 2.13.66 I get a lot of warnings resulting from the path letter being different, such as this

```
 warning  in D:/my/Websites/node_modules/object-assign/index.js
⠀
There are multiple modules with names that only differ in casing.
This can lead to unexpected behavior when compiling on a filesystem with other case-semantic.
Use equal casing. Compare these module identifiers:
* D:\my\Websites\node_modules\object-assign\index.js
    Used by 160 module(s), i. e.
    D:\my\Websites\node_modules\react\cjs\react.development.js
* d:\my\Websites\node_modules\object-assign\index.js
    Used by 4 module(s), i. e.
    d:\my\Websites\node_modules\react\cjs\react.development.js
```

I believe the error results from https://github.com/gatsbyjs/gatsby/commit/cac4246

Up until Gatsby v2.13.65 all path letters were in lower cases, which was good.

I also believe the Windows path letters in packages/gatsby/src/utils/__tests__/path.js should be lower case.

waltercruz:
Fix the weird way that windows handles drive letter upper/lower case.

## Description

When running on windows, change the first path of path component (drive letter) to uppercase to match true-case-path

## Related Issues

Fix #16721

See also [this issue](https://github.com/Profiscience/true-case-path/issues/3#issuecomment-528832215) on true-case-path


sidharthachatterjee:
Published in `gatsby@2.15.10`

muescha:
@sidharthachatterjee i think it is a common use case -> should this go into some gatsby core helpers?

sidharthachatterjee:
@muescha Yeah, I think this could live in utils

PR #17503, opened 2019-09-09
fix: upgrade chrome to fix e2e-test flake

wardpeet:
# Description

Upgrading chrome to a newer version.
I also added cross-env to have proper local support for windows users


sidharthachatterjee:
I wonder if this will fix it. I remember @davidbailey00 tried upgrading Chrome earlier which broke in all sorts of strange ways 😅

wardpeet:
no clue, i'm also trying to run 69 on docker locally to see if I can mimic behaviour

PR #17511, opened 2019-09-09
fix(gatsby): fix node internal validation

pieh:
we were allowing custom fields in `internal` to pass validation because of misconfiguration of joi schema - `unknown()` modifier accept boolean (called `allow`), and not object with `allow` field. When using object, parameter was truthy which essentially was doing exact opposite than what we wanted it to do.

I also added very basic tests for this


pieh:
This probably needs some more changes when handling validation errors - right now we short circuit node creation and output error to console, but this means we can display 1000s of errors - so some grouping need to happen in https://github.com/gatsbyjs/gatsby/blob/8562e93c9d413b2660a40aaad34fb10b33d98921/packages/gatsby/src/redux/actions/public.js#L671-L677

pieh:
note - failing test doesn't seem related to my changes (happens in master as well) - https://github.com/gatsbyjs/gatsby/pull/17515 is separate PR that should address it

pieh:
Preview of updated error:

- if call site can be figured out:
Screenshot 2019-09-09 at 19 50 49
- if it can't (missing code frame + location)
Screenshot 2019-09-09 at 19 49 21

packages/gatsby/src/joi-schemas/__tests__/joi.js
```diff
@@ -101,3 +101,57 @@ describe(`gatsby config`, () => {
     }
   })
 })
+
+describe.only(`node schema`, () => {
```
sidharthachatterjee:
```suggestion
describe(`node schema`, () => {
```

packages/gatsby/src/utils/stack-trace-utils.js
```diff
@@ -13,7 +17,9 @@ const getNonGatsbyCallSite = () =>
       callSite =>
         callSite &&
         callSite.getFileName() &&
-        !callSite.getFileName().includes(gatsbyLocation)
+        !callSite.getFileName().includes(gatsbyLocation) &&
```
sidharthachatterjee: Wonder if there are any others we want to add in here?

packages/gatsby/src/joi-schemas/__tests__/joi.js
pieh: 😅

packages/gatsby/src/utils/stack-trace-utils.js
pieh: I'm not exactly happy with those (it was kind of trial and error, seeing what will pop up there) and we should only handle packages that are there because of gatsby internals (and nothing else) - so I don't think we should add any more (right now)

pieh:
> Looking good! We should batch those error messages though to prevent an overwhelming bunch of logs at once like you described

I already check validation exact error, so it display only once for given validation error - so if you create 1000 nodes that have `internal.custom` field - it will be displayed just once.

However if you have 1000 nodes that have `internal.custom_${N}`, it will still be displayed 1000 times, because it's different validation error message

Still better than what we had before (when it worked) which just displayed it for every node

packages/gatsby/src/utils/stack-trace-utils.js
sidharthachatterjee: Okay, that's fair

sidharthachatterjee:
Published in `gatsby@2.15.13`

PR #17544, opened 2019-09-11
fix(gatsby): add / to prefix when stripping prefixes

### LINKED ISSUE #16457: The configured pathPrefix is stripped from the path of pages that start with the same prefix
## Description

Between version 2.8.8 and 2.9.0 of gatsby a change was made that caused the configured `pathPrefix` to be stripped from the path of pages that start with the same prefix.

The issue is still present in 2.13.52

### Steps to reproduce

I have set `pathPrefix` to `/p`, and I have a directory in the page root directory called `/privacynotice`.
gatsby will incorrectly strip `/p` from the page location, resulting in `rivacynotice` being used.
When hovering a `<Link to="/privacynotice">...</Link>` the browser will attempt to fetch the file `/p/page-data/rivacynotice/page-data.json`.

robin-macpherson:
## Description
*_Copied from the issue description_*:
Between version 2.8.8 and 2.9.0 of gatsby a change was made that caused the configured pathPrefix to be stripped from the path of pages that start with the same prefix.

I will try my best to investigate the potential solution presented by the issue reporter, by modifying the `trimPathname` function so that it only strips the prefix if it matches the specified pathPrefix followed by a `/`, and try to think of any problems with that solution.

## Related Issues
Fixes #16457: 

### Steps to reproduce the issue
* set `pathPrefix` to `/p`, and create a directory in the page root directory called `/privacynotice`.
* gatsby incorrectly strips `/p` from the page location, resulting in `rivacynotice` being used.
* When hovering a `<Link to="/privacynotice">...</Link>` the browser will attempt to fetch the file `/p/page-data/rivacynotice/page-data.json`

### Todos
- [x] Identify best place to fix this - current best lead is the `trimPathName()` function in `find-path.js`
- [x] Find and implement the fix
- [x] Seek guidance on any potential flaws with this approach
- [x] Update unit tests - `gatsby/packages/gatsby/cache-dir/__tests__/find-path.js`
- [x] e2e tests


robin-macpherson:
I had a look and the suggested fix seemed okay to me, so I thought it might be a good idea to implement it and work on adding some unit tests. In the meantime, I would love some guidance from the core team on any potential downsides or consequences that I'm not seeing here. Thank you in advance!

wardpeet:
Hey @robin-macpherson, i'm sorry to push a commit to your PR but it was a bit difficult to explain XD.

You're fix was a great workaround 👍 but it wasn't fixing the real problem. We remove the prefix twice in loader.js so that's something I've cleaned up.

It would be great if you could add a test to https://github.com/gatsbyjs/gatsby/tree/master/e2e-tests/path-prefix to make sure we don't have this regression again.

Does that sound good?

robin-macpherson:
@wardpeet ah that makes sense!! No problem at all about pushing a commit to my PR xD I think this would have been difficult for me to figure out and it's great to see what you did and get the short explanation 😄 I'll work on adding a test where you suggested and update you! 

Thanks again 💪🏼

wardpeet:
I've fixed the unit tests. Seems like there were some issues with my change. @robin-macpherson if you need any help with the e2e tests, feel free to reach out on discord.

robin-macpherson:
> I've fixed the unit tests. Seems like there were some issues with my change. @robin-macpherson if you need any help with the e2e tests, feel free to reach out on discord.

Thanks so much for all the help @wardpeet ! Sorry for the delay on my side, that last week was a bit insane. Could you point me in the right direction for updating e2e tests and I'll be more than happy to update them!

robin-macpherson:
Hi @wardpeet ! Just wanted to follow up here regarding the e2e tests. I'll send an e-mail too just in case that's better.

kmck:
Not sure about the e2e tests, but wanted to mention again that the bugfix to `packages/gatsby/cache-dir/loader.js` that prevents double-processing the path (af4b9280ad47c76a295b9a9e7729c91cec5c923e) seems like it needs to be re-added to this branch since it was lost in the merge conflict:

```diff
-  loadPageDataJson(rawPath) {
-    const pagePath = findPath(rawPath)
+  loadPageDataJson(pagePath) {
```

For reference, here's where it was removed: https://github.com/gatsbyjs/gatsby/commit/19a95fba14401b2adc62b1ab1b27db5c05653778#diff-06ce5193de7c2c01dd947a5001c147c3L147-R148

wardpeet:
Hey @Robin MacPherson sorry fo taking so long. I think I finally cracked your PR and made it work with e2e-tests. What do you think? Does this fix your issue?

robin-macpherson:
> Hey @robin MacPherson, sorry for taking so long. I think I finally cracked your PR and made it work with e2e-tests. What do you think? Does this fix your issue?

Hi @wardpeet, thanks for getting to this. Looks great and it is very helpful to see how you went about resolving the heart of the issue and adding all the necessary tests.

PR #17571, opened 2019-09-11
fix(gatsby-source-contentful): Fixing when assets are in draft state

tuliofaria:
When in draft, an asset could not have a file attached to it yet. Fixing this scenario.

## Description

When downloading assets from Contentful in preview, we can face problems with assets in draft that does not have a file attached to it yet. This leads to errors trying to process node.file.url.

### DIFF
--- a/packages/gatsby-source-contentful/src/download-contentful-assets.js
+++ b/packages/gatsby-source-contentful/src/download-contentful-assets.js
@@ -45,7 +45,7 @@ const downloadContentfulAssets = async gatsbyFunctions => {
       const { contentful_id: id, node_locale: locale } = node
       const remoteDataCacheKey = `contentful-asset-${id}-${locale}`
       const cacheRemoteData = await cache.get(remoteDataCacheKey)
-      if(!node.file){
+      if (!node.file) {
         return Promise.resolve() 
       }
       const url = `http://${node.file.url.slice(2)}`
--- a/packages/gatsby-source-contentful/src/download-contentful-assets.js
+++ b/packages/gatsby-source-contentful/src/download-contentful-assets.js
@@ -45,6 +45,9 @@ const downloadContentfulAssets = async gatsbyFunctions => {
       const { contentful_id: id, node_locale: locale } = node
       const remoteDataCacheKey = `contentful-asset-${id}-${locale}`
       const cacheRemoteData = await cache.get(remoteDataCacheKey)
+      if(!node.file){
+        return Promise.resolve() 
+      }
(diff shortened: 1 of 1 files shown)


wardpeet:
I'm not 100% familiar with contentful so could you give me a step by step process on how to mimic this behavior?

tuliofaria:
Hi @wardpeet , sure!
Contentful has two stages for content: preview (for drafts or waiting for revision content) and production (published content). Usually, it's common to develop and test with preview content - so we can mock data up. The problem is related to assets that can have those states as well (draft and published). Draft assets can happen to not have a file yet - and this raises the problem that this PR is fixing.

tuliofaria:
The step by step:
- configure gatsby-source-contentful to connect to preview api
- create an image asset without file

wardpeet:
Ok tested and your solution works. Thanks for the explanation. I wonder if we should still crash with a proper error message? I think you want an asset with an image nonetheless right?

tuliofaria:
Thanks @wardpeet . 
Maybe I better solution would be just warning about the issue and skip the draft asset. The problem of crashing is that we may stop a continuous deployment (CD) of preview content. My use scenario would be affected. Right now I have two CD pipelines:

- Contentful preview API >> triggers a preview website (all the editors and stakeholders review the content from there)
- Contentful production API >> trigger a production deployment

If we crash, pretty often I'm going to break my preview website.

packages/gatsby-source-contentful/src/download-contentful-assets.js
```diff
@@ -45,6 +45,9 @@ const downloadContentfulAssets = async gatsbyFunctions => {
       const { contentful_id: id, node_locale: locale } = node
       const remoteDataCacheKey = `contentful-asset-${id}-${locale}`
       const cacheRemoteData = await cache.get(remoteDataCacheKey)
+      if (!node.file) {
+        return Promise.resolve()
```
wardpeet: I'm not 100% sure if `node.name` exists but it should be the name of the asset. Something along the lines of this would be nice
```suggestion
        reporter.warn(`The asset ${node.name} contains no file.`)
        return Promise.resolve()
```

tuliofaria: I would add the id as well (it helps find the asset in contentful):
```
reporter.warn(`The asset ${node.name} with id: ${id} contains no file.`)
```

tuliofaria: @wardpeet checking here, I think it will be safer to just show the id - because even name could not be present in asset yet.

PR #17579, opened 2019-09-12
docs: deploying to Heroku, update to headers config

richhiggins:
## Description

URLs without a trailing slash were not applying the correct caching header.

This small config change means that `"Cache-Control": "public, max-age=0, must-revalidate"` headers will also get implicitly applied to page-data.json files, which is desirable, and the manifest.webmanifest file which I don't believe we'd want to set far-future (or even a near future) header on.

Other far-future header settings remain as they were.

### DIFF
--- a/docs/docs/deploying-to-heroku.md
+++ b/docs/docs/deploying-to-heroku.md
@@ -44,7 +44,7 @@ The following configuration will give you a good start point in line with Gatsby
 {
   "root": "public/",
   "headers": {
-    "/**/": {
+    "/**": {
       "Cache-Control": "public, max-age=0, must-revalidate"
     },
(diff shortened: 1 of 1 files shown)


wardpeet:
Oh I'll have to checkout the heroku recipe to see if the cache headers are correct. We don't want to cache page-data at all.

richhiggins:
@wardpeet this config does apply a no-cache header to the page-data.json files (as well as page requests without trailing slash). 

It's essentially setting a default no-cache header with the aggressive caching headers then specified more specifically for css, js, icons, etc... 

The above config is in use on this live site https://co-operate.coop.co.uk/ so you can see how the headers are being applied.

Issue #16981, opened 2019-08-23
[docs][workflows] using GraphQL in Gatsby


marcysutton:
Part of the **Top 25 Learning Workflows initiative**. See #13708 for the meta issue that this issue falls under.

**Note: this is a placeholder issue for work in progress.**

## User story

As a new Gatsby user, I want to understand how GraphQL works in Gatsby and get it working properly in a project.

The motivation for this issue is to make sure users aren't deterred from using Gatsby by what they find when searching the docs for GraphQL information. There's a ton of it on there already, including why Gatsby uses GraphQL and how to build sites without it. However, some of the docs don't have the most intuitive titles or locations for people to find information easily (like API docs). We have added recipes recently, which helps! But given the importance of GraphQL in Gatsby, we should take a deeper look at it similar to what we did with images.

## Evaluation

_in progress by @gillkyle_

| Search | Discover | Complete | Linked | Tone | Style | Overall |
| --- | --- | --- | --- | --- | --- | --- |
| 😄 | 😐 | 😄 | 😐 | 😐 | 😐 | 😄 |

# Steps taken to implement

1. Searchability:
    - Searched `graphql in gatsby`: first result [Querying with GraphQL](https://www.gatsbyjs.org/docs/querying-with-graphql/) (name of that article is a little confusing when this a conceptual guide, maybe should include a note distinguishing this from the reference guide on GraphQL, also renaming this to show that's conceptual would be helpful and would solve the naming collision between this and the reference guide page)
    - Search `gatsby graphql` same first and second result as the first search (Querying data with GraphQL and then GraphQL and Gatsby)

2. Discoverability:
    - Searched `graphql` in the .org search bar which shows a blog post for Kentico Cloud as the first result and "Gatsby without GraphQL" as the second 😱(which seems to be a result of this section being in the sourcing data section before GraphQL is introduced, maybe move Using Gatsby Without GraphQL in the sidebar to below Sourcing from (Headless) CMSs), the 3rd result is "GraphQL and Gatsby"
3. Completeness:
	- Good level of content addressing lots of use cases, the recipes give some quick, digestable examples and there are several reference guides, mostly on querying and creating pages
	- Followed steps in [Creating and Modifying Pages](https://www.gatsbyjs.org/docs/creating-and-modifying-pages/) as it appeared to be the most relevant by title to actually writing code
[...]

Issue #15616, opened 2019-07-10
[docs][themes] Add visual aids to explain themes concept better in docs


prestonso:
Per conversation with @amberleyromo last week!

It would be really strong to have some visuals to support our theme docs:

* Brainstorming how we can better visually explain the concept of themes, taking inspiration from the visuals associated with [Trevor’s case study](https://www.gatsbyjs.org/blog/2019-07-03-using-themes-for-distributed-docs/).
* Designing and implementing these visual aids in the docs.

This is a relatively low-priority issue, but it could also be picked up by the community. This is a minor Learning improvement but is important due to our push for greater Themes adoption.

PR #17589, opened 2019-09-12
docs: layer model interactive component

gillkyle:
## Description

In an effort to explain the GraphQL docs better I drew up what I thought would explain some of the "layers of Gatsby" as I thought of them. These are pretty simplified but generally express what happens in Gatsby in an order that shows data going all the way from a source to being consumed in a component.

I did some hacking to make the Content layer have the data be able to cycle through different data sources that all have the same simple content to express that data can come from anywhere and all ends up looking the same in the GraphQL layer. 

I petitioned @fk for some design critiquing and used his input to clean up the design to make it less busy.

layers-model

Added the component to a handful of guides in the Conceptual section where I think it could do the most help expressing what's happening at different places.

Happy to hear any suggestions on how to improve content of each section, design, or whatever it may be.

## Related Issues

Related to #15616 about adding more visual aids to docs (though this is about GraphQL, _not_ themes like that issue mentionos)

This came about while working on the GraphQL workflow in #16981, trying to explain these concepts better


LekoArts:
> I did some hacking to make the Content layer have the data be able to cycle through different data sources that all have the same simple content to express that data can come from anywhere and all ends up looking the same in the GraphQL layer.

It took me some time to get what "cycle" actually means in the context, so I'd propose to make the label more descriptive (e.g. "Cycle data source"). Longer text will impact your design, but I think this makes it more clear.

---

Love the general idea and execution, well done!

gillkyle:
Thanks for the review @LekoArts, I updated the button to say "cycle source" which I think would suffice 👍

KyleAMathews:
Nifty! Looks great

www/src/assets/icons/layer-icon.js
```diff
@@ -0,0 +1,324 @@
...
+        <path
+          d="M32.5543 8.65346C34.4648 8.65346 36.0136 7.10468 36.0136 5.19416C36.0136 3.28364 34.4648 1.73486 32.5543 1.73486C30.6438 1.73486 29.095 3.28364 29.095 5.19416C29.095 7.10468 30.6438 8.65346 32.5543 8.65346Z"
+          stroke={fillColor}
+          strokeWidth="1.52728"
+          strokeLinecap="round"
+          strokeLinejoin="round"
+        />
+      </svg>
+    ),
+    Data: (
```
DSchau: Little nit-picky, but we can clean this file up quite a bit if we break these SVG icons out into either their raw SVG equivalent (and import with raw-loader) or as a React component and then the file would look like:

```js
import React from 'react'

import Data from './data'

const lookup = {
  Data
}

const LayerIcon = ({ name }) => {
  const Icon = lookup[name]
  return Icon ? <Icon {...whateverProps /> : null
}
```

www/src/components/layer-model/layer-content-sections.js
```diff
@@ -0,0 +1,352 @@
...
+  mediaQueries,
+} from "../../utils/presets"
+
+// Components for building sections used in the model
+const LayerContentWrapper = ({ children }) => (
+  <div
+    css={{
+      padding: `${space[4]} 0`,
+      display: `grid`,
+      gridTemplateRows: `1fr 1fr`,
```
DSchau: little trick :)

```suggestion
      gridTemplateRows: `repeat(2, 1fr)`,
```

www/src/components/layer-model/layer-content-sections.js
```diff
@@ -0,0 +1,352 @@
...
+      gridTemplateAreas: `"example" "content"`,
+      gridGap: 0,
+      [mediaQueries.md]: {
+        gridTemplateRows: `1fr`,
+        gridTemplateAreas: `"example content"`,
+        gridTemplateColumns: `1fr 1fr`,
```
DSchau: same idea here

```suggestion
        gridTemplateColumns: `repeat(2, 1fr)`,
```

www/src/components/layer-model/layer-content-sections.js
```diff
@@ -0,0 +1,352 @@
...
+              right: space[3],
+              backgroundColor: `transparent`,
+              border: `none`,
+              color: colors.grey[60],
+              cursor: `pointer`,
+              padding: `${space[2]} ${space[2]}`,
+              transition: transitionProperty,
+              borderRadius: radii[2],
+              whiteSpace: `nowrap`,
+              ":focus": {
```
DSchau: We could probably clean this up a bit by grouping, right?

```js
css={{
  ":focus, :hover, :active": {
    boxShadow: shadows.floating,
    color: colors.white,
  }
}}
```

then no need to split up!

www/src/components/layer-model/layer-content-sections.js
```diff
@@ -0,0 +1,352 @@
...
+              },
+              ":hover": {
+                ...pseudoProperties,
+                backgroundColor: colors.purple[50],
+              },
+              ":active": {
+                ...pseudoProperties,
+                backgroundColor: colors.purple[60],
+              },
+              ":focus:before": {
```
DSchau: ::before right?

```suggestion
              ":focus::before": {
```

www/src/components/layer-model/layer-content-sections.js
```diff
@@ -0,0 +1,352 @@
...
+                content: `"cycle source "`,
+              },
+              ":hover::before": {
+                content: `"cycle source "`,
+              },
+            }}
+            onClick={() =>
+              setSourceIndex(
+                sourceIndex === sources.length - 1 ? 0 : sourceIndex + 1
```
DSchau: Writing this without testing, but neat little trick :)

```suggestion
                (sourceIndex + 1) % sources.length
```

it'll be the next index if not greater than the length of the array. Not necessary though, use your discretion if it's better!

www/src/assets/icons/layer-icon.js
DSchau: However -- since these seem to only be used by _this file_ it could be prematurely optimizing a bit :) Up to you!

docs/docs/querying-with-graphql.md
```diff
@@ -19,6 +21,10 @@ components](/docs/building-with-components/) to declare what data they and their
 sub-components need. Then, Gatsby makes that data available in
 the browser when needed by your components.
 
+Data from any number of sources is made queryable in one unified layer:
```
marcysutton: Because the interactive piece includes other aspects of Gatsby while focusing on the data layer by default, it would help to mention somehow why the other sections are present. How does data relate to the rest of it?

Here's one attempt, let me know what you think: "Data from any number of sources is made queryable in one unified layer, a key part of Gatsby the building process:"

www/src/assets/icons/layer-icon.js
```diff
@@ -0,0 +1,324 @@
...
+        />
+      </svg>
+    ),
+    App: (
+      <svg
+        width="37"
+        height="32"
+        viewBox="0 0 37 32"
+        fill="none"
+        xmlns="http://www.w3.org/2000/svg"
```
marcysutton:
```suggestion
        xmlns="http://www.w3.org/2000/svg"
        focusable="false"
        role="presentation"
```

www/src/components/layer-model/layer-model.js
```diff
@@ -0,0 +1,123 @@
...
+        ":focus": {
+          outline: 0,
+          boxShadow: `0 0 0 3px ${hex2rgba(colors[baseColor][30], 0.5)}`,
+        },
+        ":hover": {
+          backgroundColor: colors[baseColor][5],
+        },
+      }}
+    >
+      <div
```
marcysutton: Can these items inside the button be changed to `<span>` elements and styled with CSS to avoid putting DIVs inside?

www/src/components/layer-model/layer-model.js
```diff
@@ -0,0 +1,123 @@
...
+        cursor: `pointer`,
+        borderRadius: radii[3],
+        padding: space[2],
+        color: colors.grey[60],
+        fontWeight: selected ? `bold` : `normal`,
+        backgroundColor: colors.grey[5],
+        border: selected
+          ? `2px ${colors[baseColor][60]} solid`
+          : `2px transparent solid`,
+        ":focus": {
```
marcysutton: Can we use the default focus style to be more consistent with the rest of the site?

www/src/components/layer-model/layer-model.js
```diff
@@ -0,0 +1,123 @@
...
+  ViewLayerContent,
+  AppLayerContent,
+} from "./layer-content-sections"
+
+const Layer = ({ layer, onClick, selected, children }) => {
+  const { baseColor, title } = layer
+
+  return (
+    <button
+      onClick={onClick}
```
marcysutton: I love how these are focusable and they work with the keyboard. This component looks like a tab switcher to me, though, so it would be amazing to make it function with one tab stop for the controls. Here's the ARIA Authoring Practices Guide for tabpanels, which will show semantics and navigation conventions: https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel

And here's a Codepen demo I made that might help: https://codepen.io/marcysutton/pen/oKBzdr

docs/docs/gatsby-lifecycle-apis.md
```diff
@@ -15,6 +17,10 @@ Gatsby's design principles include:
 
 # High level Overview
 
+The following model gives a conceptual overview of how data is source and transformed in a Gatsby site:
```
marcysutton:
```suggestion
The following model gives a conceptual overview of how data is source and transformed in the process of building a Gatsby site:
```

docs/docs/building-with-components.md
```diff
@@ -15,6 +17,10 @@ open source components, tutorials, and tooling that can be used seamlessly for
 building sites with Gatsby. Gatsby is built to behave almost exactly like a
 normal React application.
 
+The following model shows how data from a source can be queried by GraphQL for use inside components:
```
marcysutton:
```suggestion
The following model shows how data from a source can be queried by GraphQL for use inside components in the process of building a Gatsby site:
```

docs/docs/gatsby-lifecycle-apis.md
```diff
@@ -15,6 +17,10 @@ Gatsby's design principles include:
...
+
+<LayerModel />
```
marcysutton: If there isn't an initialView specified, it would help to default to the first tab since the component otherwise looks like it relates to the text content below.

muescha: @marcysutton yes. I like the interactive tab switcher. It is a new component for the docs. 

But I see there is a problem that nobody knows that I can tap on the icons and see more conten. I don't know how it could be done better. Maybe a hint "Klick on the icons and switch the datasource" or something like this.

I see also other elements like this cycle through the layers ever xx seconds to show there is more content 

Also it would be nice to have all 5 options all visible for printing or something (like on some pages you can have a gallery and also.the option "view all on one page)

marcysutton: I would advocate against auto-rotating, personally – it would distract from the reading experience and cause accessibility problems. If the icons don't look interactive enough, perhaps there's a visual design tweak we could make. Having to say "click these icons" suggests the design could be improved -- perhaps making the buttons look more like buttons with edges/borders could help, even just having one selected by default could help.

I agree about showing all content for a the print stylesheet -- that's a great point, and if it could be handled with CSS now, it would make future work to print the docs go a little bit faster.

gillkyle: I think defaulting to the initial view will help with the issue that they may not look clickable, that's great feedback 👍

www/src/assets/icons/layer-icon.js
gillkyle: Yeah I think I'm going to go the path of least resistance and just keep them in the same file until it feels more necessary to break them up 😬

docs/docs/querying-with-graphql.md
gillkyle: Good point, I like how that's phrased. I think the paragraph above this one helps give some context as well since it explains that data is made available to components in the browser and hopefully the little examples can help that sink in.

www/src/components/layer-model/layer-model.js
gillkyle: I think I got this all sorted out! Maybe I should have just used reach UI's tabs but I think I've got everything the spec requires 🙂

www/src/components/layer-model/layer-content-sections.js
```diff
@@ -0,0 +1,345 @@
...
+                backgroundColor: colors.purple[60],
+              },
+              ":focus::before": {
+                content: `"cycle source "`,
+              },
+              ":hover::before": {
+                content: `"cycle source "`,
+              },
+            }}
+            onClick={() => setSourceIndex((sourceIndex + 1) % sources.length)}
```
marcysutton:
```suggestion
            onClick={() => setSourceIndex((sourceIndex + 1) % sources.length)}
            aria-label="Update code source"
```
The button only has text in it when you focus, which means it will be empty in a screen reader's elements list.

gillkyle:
Thanks @marcysutton! Awesome feedback! I'll fix that up now!

muescha:
is there any chance to make this component more generic?

For example:

Cards for the upper big icons row
```jsx
<cards default=“data” layout=“top-icon”>
  <card name=“data” icon=“data-icon” title=“Data Layer”>
    {children}
   </card>
   <card name=“Source” icon=“Source-icon” title=“Sourcing”>
    {children}
   </card>
</cards>
```

Tabs for the small code snippets
```jsx
<tabs default=“Js” layout=“tabs-icon-and-text”>
  <tab name=“Js” icon=“data-icon” title=“Data Layer”>
    {children}
   </tab>
   <tab name=“gql” icon=“gql-icon” title=“GraphQL”>
    {children}
   </tab>
</tabs>
```


Maybe we can combine the cards and tabs and only differnicate by a laout:

The hole is:
- a Big Icons on top layout. 
- Content below

The code snippes:
- small icons (for tech symbols depending on language and a title
- content below


Benefits:
- reusable. I dreaming all time of a language toggle tab layout to display yarn commands instead of the npm (selected language is stored)
- content stays only in doc document and is available for translation team

muescha:
As a fast Google search: there also react compontens in the wild, but I don't know how well the fit: all under the keyword: 

React content tabs

gillkyle:
@muescha I wasn't sure it was worth the effort of making it more generic at this point because I couldn't think of many more use cases. Having an npm/yarn toggle could be similar but might have different requirements and could probably happen in a PR for something more specific like that.

muescha:
Reusable is one side effect 

But i18n is on the way - and for this all text should be stay in /docs/

https://github.com/gatsbyjs/rfcs/pull/42

jonniebigodes:
@gillkyle awesome job. i've gone over it and it looks awesome. 

@muescha if by more generic you mean abstracting it to allow the content/text of the component to be translated to someone that visits the page. I think it's best to leave it as is for now until we a solid foundation in the translation effort. A mid future goal should be have a good portion of the documentation already translated in some languages. A more end goal would components like this one that was created and possibly others.

muescha:
Generic i mean that all the localized content is in doc. Presentation is in the component

Like my draft code example: `{children}` is any content from the mdx doc page

marcysutton:
@muescha perhaps that could be discussed later for a future enhancement? I'd prefer to keep the scope of this limited so we can move it forward.

alexlbr:
@gillkyle I think this visual representation is a big step to illustrate the process, it's helped me understand the puzzle much better. Although it's very helpful, I got a bit confused with the "build" concept. I see in the text that "build" refers to the process itself and it's also used in one of the tabs as a step in the process as well. It also confuses me a bit to use the GraphQL logo in the tab that displays a JSON and not on the previous one where the GraphQL query is.
image

What I understood from the examples is that during the build process the "content" is "queried" to get "data" for the "view" to be rendered in the "app", meaning:

image

Sorry for the late feedback :)

muescha:
mhh also i miss there a "import data" build step where it is said that data is fetched imported into the gatsby data warehouse in graphql. it says only "can be connected to Gatsby" but not exactly where the data is stored.

gillkyle:
@alexlbr this is great feedback! 

I think in order to resolve that confusion maybe the right move would be to move the information in the build step (those 2 paragraphs on the right in what is currently the build layer) into the initial content step and explain that at that point the content gets pulled from different places and a schema is made all during Gatsby's processing.  

I like the idea of moving that up, I wonder if anyone else from @gatsbyjs/learning  has any thoughts?

@muescha I think for clarity the build step could be reworded, though if changes are made to resolve what @alexlbr found confusing perhaps that would fix that anyway. I didn't include really granular details about the build on purpose because this is a pretty conceptual thing, though I'm happy to hear how you'd reword it!

muescha:
@gillkyle 👍 

yes you are right, too much granular would be not good. my thing was just an expectation and can be reworded be more short, like the suggested changes in your answer to @alexlbr

PR #17590, opened 2019-09-12
feat(gatsby-plugin-offline): "Magic" JS detection to make sites function correctly when JS is disabled retroactively

### LINKED ISSUE #17386: Page HTML isn't served when visiting a site with offline plugin, then disabling JS
## Description

HTML for a page isn't displayed when visiting a site with gatsby-plugin-offline initially with JS enabled, then disabling it.

### Steps to reproduce

1. Build a site with gatsby-plugin-offline
2. Visit it
3. Disable JS in the browser
4. Reload the page

### Expected result

JS-free version of the site loads

### Actual result

"This site works best with JavaScript enabled" and no actual content

### Environment

n/a

davidbailey00:
## Description

`gatsby-plugin-offline`'s service worker (which still runs if JavaScript has been disabled retroactively) can now detect via a conditional network request whether or not JavaScript has been disabled in the browser, and serve a page from the network if this is the case.

## Related Issues

Closes #17386


KyleAMathews:
Nice "magic" :-D

packages/gatsby/cache-dir/default-html.js
```diff
@@ -15,9 +15,6 @@ export default function HTML(props) {
       </head>
       <body {...props.bodyAttributes}>
         {props.preBodyComponents}
-        <noscript key="noscript" id="gatsby-noscript">
-          This app works best with JavaScript enabled.
-        </noscript>
```
pieh: should this change be part of this PR?

pieh: Also for reference - this is PR that added this initially https://github.com/gatsbyjs/gatsby/pull/10945 so worth checking reasoning and see if we don't break that (or at least it's conscious decision)

PR #17613, opened 2019-09-13
fix(gatsby-plugin-sharp): [gatsby-cloud] actually wait for cloud processing to finish

pieh:
Request we are currently making to cloud service to process image only queue image processing, and respond when it was put in the queue, but not when image were actually processed

We need to wait for message from cloud to let us know when image processing finished and when assets where moved over to `public` directory to properly await for jobs to be actually finished.

This change is prototype that need to be improved, but generally it works.

We need to iron out what the message (received via ipc) should look like - need to spec proper contract here - this should include handling success and failures. Error objects are poorly serializable (you can, but you receive object on the other side) - maybe just boolean switch would be enough (?)

In current implementation message looks like this:
```
    gatsbyProcess.send({
      type: `GATSBY_CLOUD_IMAGE_SERVICE`,
      id: req.body.id,
    })
```
but this need to be carefully thought through - I expect in the future gatsby core would have proper spec for receiving events via ipc


packages/gatsby-plugin-sharp/src/scheduler.js
```diff
@@ -6,9 +6,12 @@ const { createProgress } = require(`./utils`)
...
 const toProcess = {}
 let totalJobs = 0
-const q = queue((task, callback) => {
-  task(callback)
-}, 1)
+const q = queue(
+  (task, callback) => {
+    task(callback)
+  },
+  process.env.GATSBY_CLOUD_IMAGE_SERVICE_URL ? Number.MAX_SAFE_INTEGER : 1
```
pieh: cloud service shouldn't be limited to 1 job at a time, as it just queues jobs - we don't want this internal queue to be bottlenech

packages/gatsby/src/commands/build.js
```diff
@@ -33,10 +33,14 @@ const waitJobsFinished = () =>
       if (store.getState().jobs.active.length === 0) {
         resolve()
         emitter.off(`END_JOB`, onEndJob)
+        return true
       }
+      return false
+    }
+    if (!onEndJob()) {
+      console.log(`[gatsby] waiting for jobs to finish`)
+      emitter.on(`END_JOB`, onEndJob)
```
pieh: changes here are just for debugging purpose - they will need to be removed later

packages/gatsby-plugin-sharp/src/scheduler.js
```diff
@@ -152,6 +155,7 @@ function runJobs(
     )
 
     Promise.all(promises).then(() => {
+      console.log(`[gatsby-plugin-sharp] finishing job ${job.inputPath}`)
```
pieh: changes here are just for debugging purpose - they will need to be removed later

packages/gatsby-plugin-sharp/src/process-file.js
```diff
@@ -43,6 +44,18 @@ const argsWhitelist = [
   `background`,
 ]
 
+let cloudEventsEmitter
+if (process.env.GATSBY_CLOUD_IMAGE_SERVICE_URL) {
```
pieh: I went with assumption that cloud worker doesn't have this env var set - otherwise `processFile` would be circularly calling itself I think

packages/gatsby-plugin-sharp/src/process-file.js
```diff
@@ -75,25 +88,50 @@ exports.processFile = (file, contentDigest, transforms, options = {}) => {
...
               },
               json: true,
             })
+            .then(
+              () =>
+                new Promise((resolve, reject) => {
+                  console.log(`Sent request ${cloudJobID} / ${file}`)
+                  const eventToListenFor = `GATSBY_CLOUD_IMAGE_SERVICE_${cloudJobID}`
+                  const handler = msg => {
+                    cloudEventsEmitter.off(eventToListenFor, handler)
```
pieh: `mitt` doesn't have `once` so we need to manually stop listening

packages/gatsby-plugin-sharp/src/process-file.js
```diff
@@ -75,25 +88,50 @@ exports.processFile = (file, contentDigest, transforms, options = {}) => {
...
+                })
+            )
             .then(() => transform)
 
           return cloudPromise
         }
 
-        return Promise.resolve(transform)
+        // actually wait for cloud promise so reporting is correct
+        return cloudPromise.then(() => transform)
```
pieh: did this change, because if job has 6 images to produce - it would immediately say that 5 of them are done while actually waiting for all of them

pieh:
I don't think we should merge it yet. Shape of message is important - it's first incoming ipc message that gatsby would support, and currently it kind of lives in global space, so some kind of namespacing would be beneficial here - any suggestions what we could check for prior art in different project? I was thinking about checking a bit of chrome debugging protocol to take some clues - https://chromedevtools.github.io/devtools-protocol/

It's good enough for canary, so I did publish `gatsby-plugin-sharp@cloud-image-sharp`

packages/gatsby-plugin-sharp/src/scheduler.js
muescha: Is there not an verbose/debug mode for Gatsby to show all this debug messages ?

pieh: there is, but even then - this would be way to spammy

sidharthachatterjee:
@pieh @wardpeet Shall we close this as we'll defer to the upcoming Jobs API for this?

wardpeet:
Thanks @sidharthachatterjee for the follow up. Let's close this

PR #17627, opened 2019-09-14
chore(showcase): Add nicaraguabestguides.com to showcase

nephlin7:
## Description
Add nicaraguabestguides.com to showcase


nephlin7:
> 😍

thanks :D

docs/sites.yml
```diff
@@ -7273,3 +7273,17 @@
   built_by: Gerald Martinez
   built_by_url: https://twitter.com/GeraldM_92
   featured: false
+- title: Nicaragua Best Guides
+  main_url: https://www.nicaraguasbestguides.com
+  url: https://www.nicaraguasbestguides.com
+  description: >
+    Full-Service Tour Operator and Destination Management Company (DMC)
+  categories:
+    - CMS:Contentful
```
lannonbr: The first 3 categories are not valid. The last 2 can be kept (Agency & Travel). Here's the list of all [acceptable site categories](https://github.com/gatsbyjs/gatsby/blob/master/docs/categories.yml#L62)

nephlin7: > The first 3 categories are not valid. The last 2 can be kept (Agency & Travel). Here's the list of all [acceptable site categories](https://github.com/gatsbyjs/gatsby/blob/master/docs/categories.yml#L62)

Thanks @lannonbr, already fix that

PR #17629, opened 2019-09-14
Change command to gatsby develop

kosvrouvas:
the correct command here is probably `gatsby develop` and not `npm run develop`


muescha:
Good catch 👍

PR #17632, opened 2019-09-14
chore(docs): Add section on client only routes

### LINKED ISSUE #17622: [docs] add a brief section to Building Apps with Gatsby doc on client-only routes
We've done some [shuffling around of docs recently](https://github.com/gatsbyjs/gatsby/issues/16133) to make a standalone doc on [Creating Client-Only Routes](https://www.gatsbyjs.org/docs/client-only-routes-and-user-authentication/), and the trail of information has been lost somewhat as a result. Take this StackOverflow post: https://stackoverflow.com/questions/48069170/is-it-possible-to-run-gatsby-side-by-side-with-a-regular-react-single-page-appli

Kyle points to the docs at https://www.gatsbyjs.org/docs/creating-and-modifying-pages/#creating-client-only-routes, which references a section that no longer exists. There's a link in the intro section to client-only routes on https://www.gatsbyjs.org/docs/building-apps-with-gatsby/, which contains a paragraph about it and points to https://www.gatsbyjs.org/docs/building-apps-with-gatsby/#client-only-routes--user-authentication

This issue is about adding some information back to the original reference page of [Creating and Modifying Pages](https://www.gatsbyjs.org/docs/creating-and-modifying-pages/) to include a short section on client-only routes and link to the [new page](https://www.gatsbyjs.org/docs/client-only-routes-and-user-authentication/). I'm filing it here in case someone else beats me to it (help wanted)!

thecodingaviator:
## Description

@marcysutton I have tried to follow what you wrote in your issue but I'm not sure I've got it right.

## Related Issues

Closes #17622

### DIFF
--- a/docs/docs/creating-and-modifying-pages.md
+++ b/docs/docs/creating-and-modifying-pages.md
@@ -172,4 +172,4 @@ Page context is serialized before being passed to pages: This means it can't be 
 
 ## Creating Client-only routes
 
-In specific cases, you might want to create a site with client-only portions that are gated by authentication. For more on how to achieve this, refer to [Client-only routes & user authentication](https://www.gatsbyjs.org/docs/client-only-routes-and-user-authentication/).
+In specific cases, you might want to create a site with client-only portions that are gated by authentication. For more on how to achieve this, refer to [client-only routes & user authentication](https://www.gatsbyjs.org/docs/client-only-routes-and-user-authentication/).
--- a/docs/docs/creating-and-modifying-pages.md
+++ b/docs/docs/creating-and-modifying-pages.md
@@ -169,3 +169,7 @@ export default Page
 ```
 
 Page context is serialized before being passed to pages: This means it can't be used to pass functions into components.
+
+## Creating Client-only routes
+
+In specific cases, you might want to create a site with client-only portions that are gated by authentication. For more on how to achieve this, refer to [Client-only routes & user authentication](https://www.gatsbyjs.org/docs/client-only-routes-and-user-authentication/).


wardpeet:
Previously we had the following text: (https://github.com/gatsbyjs/gatsby/commit/e29ea7bd39e1039f2ccd5c11988c889312769218#diff-6c3490bd686d41d730f3331f5af24f00)

> ### Creating client-only routes
> 
> If you're creating a "hybrid" Gatsby app with both statically rendered pages as
> well as client-only routes (e.g. an app that combines marketing pages and your
> app that lives under `/app/*`), you want to add code to your `gatsby-node.js`
> like the following:
> 
> _Note: There's also a plugin that will set up the creation of client-paths declaratively:
> [gatsby-plugin-create-client-paths](/packages/gatsby-plugin-create-client-paths/)_.
> 
> ```javascript
> // Implement the Gatsby API “onCreatePage”. This is
> // called after every page is created.
> exports.onCreatePage = async ({ page, boundActionCreators }) => {
>   const { createPage } = boundActionCreators;
>   // page.matchPath is a special key that's used for matching pages
>   // only on the client.
>   if (page.path.match(/^\/app/)) {
>     page.matchPath = "/app/:path";
>     // Update the page.
>     createPage(page);
>   }
> };
> ```

Maybe use the term hybrid as well instead of "client-only portions". I'm not sure which is better. Let's wait for @gatsbyjs/learning to review.

Thank for doing this! 💪

thecodingaviator:
@wardpeet Do you want me to change the text right now or wait for the review for both?

wardpeet:
I rather wait for the review of @gatsbyjs/learning. I'm fine with what you have atm 👍

PR #17644, opened 2019-09-15
fix: update dependency css-loader to ^3.2.0

dreyks:
## Description

Update css-loader to 3.x
This is needed for correct work of some libraries such as `asotroturf`

### DIFF
--- a/packages/gatsby-graphiql-explorer/package.json
+++ b/packages/gatsby-graphiql-explorer/package.json
@@ -40,7 +40,7 @@
     "babel-loader": "^8.0.6",
     "babel-preset-gatsby-package": "^0.2.4",
     "cross-env": "^5.2.1",
-    "css-loader": "^1.0.1",
+    "css-loader": "^3.2.0",
     "graphiql": "^0.14.2",
     "graphiql-code-exporter": "^2.0.5",
     "graphiql-explorer": "^0.4.4",
--- a/packages/gatsby/package.json
+++ b/packages/gatsby/package.json
@@ -47,7 +47,7 @@
     "copyfiles": "^1.2.0",
     "core-js": "^2.6.9",
     "cors": "^2.8.5",
-    "css-loader": "^1.0.1",
+    "css-loader": "^3.2.0",
     "debug": "^3.2.6",
     "del": "^5.1.0",
     "detect-port": "^1.3.0",
--- a/packages/gatsby/src/utils/webpack-utils.js
+++ b/packages/gatsby/src/utils/webpack-utils.js
@@ -193,14 +193,15 @@ module.exports = async ({
 
     css: (options = {}) => {
       return {
-        loader: isSSR
-          ? require.resolve(`css-loader/locals`)
-          : require.resolve(`css-loader`),
+        loader: require.resolve(`css-loader`),
         options: {
+          onlyLocals: isSSR,
           sourceMap: !PRODUCTION,
-          camelCase: `dashesOnly`,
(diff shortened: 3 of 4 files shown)


wardpeet:
@dreyks Thanks for upgrading to css-loader. Can you give us the error message you're getting with astroturf?

Upgrading css-loader from version 1x to 3x is a big jump. There are some breaking changes https://github.com/webpack-contrib/css-loader/releases.

I don't feel super confident upgrade without making sure we're 100% backward compatible which I don't think we can be.

dreyks:
astroturf relies on some css-loader@3 magic to allow component interpolation. more info https://github.com/4Catalyzer/astroturf/pull/267
without the info from the loader, it's impossible to do that

This is the list of all breaking changes from v1 to v3. I've ticked the ones that I've handled. I'll try to figure out what to do with the rest but maybe someone from the core team has any tips for me?

V3
- [x] minimum required nodejs version is 8.9.0
- [ ] `@value` at rules now support in `selector`
- [ ] invert `{Function}` behavior for `url` and `import` options (need return `true` when you want handle `url`/`@import` and return `false` if not)
- [x] `exportLocalsStyle` option was remove in favor `localsConvention` option, also it is accept only `{String}` value (use `camelCase` value if you previously value was `true` and `asIs` if you previously value was `false`)
- [x] `exportOnlyLocals` option was remove in favor `onlyLocals` option
- [x] modules option now can be {Object} and allow to setup CSS Modules options:
  - `localIdentName` option was removed in favor `modules.localIdentName` option
  - `context` option was remove in favor `modules.context` option
  - `hashPrefix` option was removed in favor `modules.hashPrefix` option
  - `getLocalIdent` option was removed in favor `modules.getLocalIdent` option
  - `localIdentRegExp` option was removed in favor `modules.localIdentRegExp` option
V2
- [ ] resolving logic for `uls()` and `import` at-rules works the same everywhere, it does not matter whether css modules are enabled (with `global` and `local` module) or not. Examples - `url('image.png')` as `require('./image.png')`, `url('./image.png')` as `require('./image.png')`, `url('~module/image.png')` as `require('module/image.png')`.
- [ ] by default css modules are disabled (now `modules: false` disable all css modules features), you can return old behaviour change this on `modules: 'global'`
- [x] `css-loader/locals` was dropped in favor `exportOnlyLocals` option
import option only affect on import at-rules and doesn't affect on composes declarations
- [x] invalid `@import` at rules now emit warnings
- [ ] use `postcss@7`

dreyks:
Also, worth noting that in any case we might break other users' configs/plugins where they modify the css-loader's options.

I fully understand that this would be a breaking change for gatsby itself and we might not just merge it as is, but I'm glad to try guiding it towards something if this is the direction that we want to move for say next major release

pieh:
In case there are breaking changes that are just not possible to rectify with any config magic, I think way forward would be with creating plugin that replace builtin default css rules (which would still use css-loader@1) with rules that use css-loader@3

wardpeet:
I was thinking the same as @pieh. You could create a separate plugin so others can use it too for their purpose or bake it inside https://github.com/silvenon/gatsby-plugin-astroturf

dreyks:
hmm... that's an interesting thought. thanks! will try to figure something out

wardpeet:
@dreyks we're here to help if you need any. Checkout [Discord](https://gatsby.dev/discord) or register for a [pair programming session](https://www.gatsbyjs.org/contributing/pair-programming/)

The change you’ve proposed is not going to be accepted because it's not a change we can do in gatsby v2.

Issue #17370, opened 2019-09-04
Support for Emotion’s extractCritical SSR rendering


fionawhim:
## Summary

Emotion’s default SSR rendering inserts `<style>` tags in the HTML body above each element being styled by an inline `css` (or `sx`) prop. This means that, until the React app loads and the browser re-renders the page, there are a number of extra elements in the DOM that the developer didn’t expect. Even though they’re themselves invisible, they disrupt any `:first-child` or `* + *`–type selectors in the CSS.

If the site uses those selectors, it can lead to a flash of the page’s content moving around as everything loads.

This is a known issue with Emotion 10’s zero-configuration SSR implementation: [emotionjs/emotion#1178](https://github.com/emotion-js/emotion/issues/1178)

As mentioned in the bug, `gatsby-plugin-emotion` could eliminate this problem, either by default or as an option, by using Emotion’s other SSR API, `extractCritical`.

### Basic example

You can see these elements easily on the [Theme UI](https://theme-ui.com/) homepage by turning off JavaScript in devtools and inspecting the DOM:

<img width="543" alt="Screen Shot 2019-09-03 at 8 55 28 PM" src="https://user-images.githubusercontent.com/260837/64217892-3962d600-ce8d-11e9-8abf-d2e1b38243d9.png">

(Though that site does not obviously use the type of selectors that these rogue elements disrupt.)

### Motivation

Emotion 10’s default SSR has two primary advantages: no special configuration, and support for React’s streaming rendering mode. Because Gatsby lets us distribute built configurations, and because it pre-renders complete HTML pages, neither of the Emotion 10 advantages is particularly compelling.

By switching to use `extractCritical`, Gatsby would eliminate the surprise that comes from these elements that do not appear in `develop` mode and can cause page loading to be unnecessarily jumpy.

While this should go by unnoticed by most users of `gatsby-plugin-emotion`, there is the potential for conflicting with sites that are already using the SSR hooks. Therefore it might be worth making this SSR rendering strategy an option rather than the default.

fionawhim:
I’d be willing to try and implement this, if folks feel like it could be accepted.

greenie:
@fionawhim I've been experiencing this issue as well when using `gatsby-plugin-emotion` together with [Typography.js](https://kyleamathews.github.io/typography.js/) in my site's global reset stylesheet.

[...]

PR #17646, opened 2019-09-15
feat(gatsby-plugin-emotion): add option to use extractCritical API

greenie:
## Description

As mentioned in the related issue, Emotion's zero-configuration SSR implementation creates extra `<style>` elements above elements styled by Emotion. This can cause rendering differences when using selectors such as `:first-child` and `* + *` that would be affected by the presence of extra elements.

This branch adds the option to use Emotion's [extractCricital](https://emotion.sh/docs/ssr#extractcritical) API, which pulls out critical styles into the `<head>` of the page. This is an opt-in feature so as not to cause compatibility issues with plugin users that are already implementing the API themselves.

I wanted to add some tests around the Gatsby APIs this plugin would now implement, but I struggled with Jest not being able to find `@emotion/core`.

Given the following:
```
// packages/gatsby-plugin-emotion/src/__tests__/gatsby-ssr.js

import { onRenderBody } from "../gatsby-ssr"

describe(`gatsby-plugin-emotion`, () => {
  describe(`onRenderBody`, () => {
    describe(`using simple server side rendering`, () => {
      it(`does not call setHeadComponents`, () => {
        const setHeadComponents = jest.fn()

        onRenderBody({ setHeadComponents }, {})

        expect(setHeadComponents).not.toHaveBeenCalled()
      })
    })
  })
})
```

Running `yarn run jest packages/gatsby-plugin-emotion` would give me the error:

```
 FAIL  packages/gatsby-plugin-emotion/src/__tests__/gatsby-browser.js
  ● Test suite failed to run

    Cannot find module '@emotion/core' from 'gatsby-browser.js'

    However, Jest was able to find:
    	'../gatsby-browser.js'

    You might want to include a file extension in your import, or update your 'moduleFileExtensions', which is currently ['js', 'jsx', 'ts', 'tsx', 'json'].

    See https://jestjs.io/docs/en/configuration#modulefileextensions-array-string

      1 | import React from "react"
    > 2 | import { CacheProvider } from "@emotion/core"
        | ^
      3 | import { cache, hydrate } from "emotion"
      4 | 
      5 | export const onClientEntry = (_, pluginOptions) => {

      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:259:17)
      at Object.<anonymous> (packages/gatsby-plugin-emotion/src/gatsby-browser.js:2:1)
```

Any pointers on how to solve this issue so I can add the tests would be much appreciated.

Thanks to @fionawhim for their examples on how they added support for extractCritical.

## Related Issues

Addresses #17370


LekoArts:

We don't want our plugins to be *too* opinionated or offer options that could be seen as bad practice. They should default to whatever the core library does. Hence I'll close this PR as we don't want this feature to be available in the plugin itself. Feel free to publish a little plugin for this or a blog post :)

PR #17652, opened 2019-09-15
chore(showcase): Updating Quick Stop Nicaragua catgories

lannonbr:
## Description

Removing some categories from the Quick Stop Nicaragua entry as it was triggering Danger as some of the categories were not valid categories.

## Related Issues

N/A


m-allanson:
Oops - thanks @LekoArts and @lannonbr 👍

PR #17666, opened 2019-09-16
chore(docs): Lazy/suspense as workaround for client side only

giacomorebonato:
## Description

I added a dependency to my [Gatsby starter project](https://github.com/giacomorebonato/semantic-starter) that fails on server side rendering: it's [Firebase UI authentication](https://github.com/firebase/firebaseui-web-react).
`React.lazy` and `suspense` don't work already on server side rendering, but still they can be forced to be used on client side only by doing a check on the existence of the window object.

This seemed to solve the situation for me; do you think that this is a viable workaround to introduce to the docs?

### DIFF
--- a/docs/docs/using-client-side-only-packages.md
+++ b/docs/docs/using-client-side-only-packages.md
@@ -80,6 +80,29 @@ const LoadableBuyButton = Loadable(() => import("./ShopifyBuyButton"))
 export default LoadableBuyButton
 ```
 
+## Workaround 4: Use React.lazy and Suspense on client side only
+
+React.lazy and Suspense are still not ready for server side rendering but they can still be used by checking if we are doing server side rendering.
+
+```jsx
+const MyPage = () => {
+  const isSSR = typeof window === "undefined"
+  const ClientSideOnlyLazy = React.lazy(() =>
+    import("../components/ClientSideOnly")
+  )
+
+  return (
+    <>
+      {!isSSR && (
+        <React.Suspense fallback={<div />}>
+          <ClientSideOnlyLazy />
+        </React.Suspense>
+      )}
+    </>
+  )
+}
+```
+
 > **Note:** There are other potential workarounds than those listed here. If you've had success with another method, check out the [contributing docs](/contributing/docs-contributions/) and add yours!
 
 If all else fails, you may also want to check out the documentation on [Debugging HTML Builds](/docs/debugging-html-builds/).
--- a/docs/docs/using-client-side-only-packages.md
+++ b/docs/docs/using-client-side-only-packages.md
@@ -82,7 +82,9 @@ export default LoadableBuyButton
 
 ## Workaround 4: Use React.lazy and Suspense on client side only
 
-React.lazy and Suspense are still not ready for server side rendering but they can still be used by checking if we are doing server side rendering.
+React.lazy and Suspense are still not ready for server side rendering, but they can still be used by checking that the code is executed only on the client.
+While this solution is inferior to `loadable-components`, that works both on server side and client, it still provides an alternative for dealing with client side only packages, without an added dependency.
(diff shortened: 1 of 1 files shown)


docs/docs/using-client-side-only-packages.md
```diff
@@ -80,6 +80,29 @@ const LoadableBuyButton = Loadable(() => import("./ShopifyBuyButton"))
 export default LoadableBuyButton
 ```
 
+## Workaround 4: Use React.lazy and Suspense on client side only
+
+React.lazy and Suspense are still not ready for server side rendering but they can still be used by checking if we are doing server side rendering.
```
muescha:
```suggestion
`React.lazy` and `React.Suspense` are still not ready for server side rendering but they can still be used by checking if we are doing server side rendering.
```

docs/docs/using-client-side-only-packages.md
```diff
@@ -80,6 +80,29 @@ const LoadableBuyButton = Loadable(() => import("./ShopifyBuyButton"))
 export default LoadableBuyButton
 ```
 
+## Workaround 4: Use React.lazy and Suspense on client side only
```
muescha:
```suggestion
## Workaround 4: Use React.lazy and React.Suspense on client side only
```

docs/docs/using-client-side-only-packages.md
```diff
@@ -80,6 +80,29 @@ const LoadableBuyButton = Loadable(() => import("./ShopifyBuyButton"))
...
+
+React.lazy and Suspense are still not ready for server side rendering but they can still be used by checking if we are doing server side rendering.
+
+```jsx
+const MyPage = () => {
+  const isSSR = typeof window === "undefined"
+  const ClientSideOnlyLazy = isSSR
```
muescha: It is never used when `isSSR == true` 
so you don't need this here?

giacomorebonato: You're right. I didn't notice it's redundant.
I'll test it anyway and do the change.

giacomorebonato: Yes, of course it works :) I just wasn't really sure what `React.lazy` is doing to webpack.
Docs updated.

wardpeet:
I'm not sure we want to advocate React.lazy at this point. We advocate loadable as it's the same as React.lazy but it works on server & client. I feel if we advocate React.lazy we will get in the same boat as window is undefined. You're code guards it but it's easily missed.

Let's hear what others have to say 😄

muescha:
Maybe: then we add a warning note that `loadable` is the better solution because it works on server and client but for completeness we add the last/suspend.

giacomorebonato:
This page is called "Using Client-Side Only Packages" and I have found it because [Firebase Auth UI](https://github.com/firebase/firebaseui-web-react) doesn't render server side.
SSR is not always better, sometimes like in this case, you don't want it and with `react.lazy` you can achieve this without an added dependency.

I'll try to write down a better description to the functionality.

giacomorebonato:
I have updated the description.
Let me know it makes sens to you, otherwise I'll close the PR.
Thanks for the discussion :)

marcysutton:
I think this works well, especially if it eliminates an extra dependency; thanks for updating the PR based on feedback. What do you think about it now, @wardpeet ?

wardpeet:
I'm still not 100% convinced we should advocate it in Gatsby as it's still could lead to problems because of the `isSSR` guard. Documentation itself is good though and it looks like more people like to have it in our docs so 👍. If it becomes a an issue we can remove it again so I'll approve & merge

giacomorebonato:
Thanks, I understand the concerns and if you decide to step back.
React Suspense is a promising technique that hopefully will improve.

It's up to you to merge right?

PR #17673, opened 2019-09-16
chore(gatsby-plugin-typography): tweak readme for clarity

kosvrouvas:
Missing commas on the examples throw errors on the terminal. Suggest we take the example from the package itself (found here: https://www.gatsbyjs.org/packages/gatsby-plugin-typography/).

### DIFF
--- a/docs/docs/typography-js.md
+++ b/docs/docs/typography-js.md
@@ -21,11 +21,11 @@ module.exports = {
     {
       resolve: `gatsby-plugin-typography`,
       options: {
-        pathToConfigModule: `src/utils/typography`
-      }
-    }
+        pathToConfigModule: `src/utils/typography`,
+      },
+    },
     // highlight-end
-  ]
+  ],
 }
 ```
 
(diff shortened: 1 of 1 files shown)


wardpeet:
Could you show what error you get? The syntax at least is correct so unsure what error you get.

kosvrouvas:
Hi, I tried to replicate it try different ways to see why I got the error.

It seems that the problem is somewhere in between (and probably because it was early AM when it happened to me the first time).

The example is wrapped inside a 

```
plugins: [

]
```

Assuming you add it at the end of it, you can run `develop` without any problems. But, adding it in the middle somewhere, it needs a comma after the second bracket.

So this probably doesn't need to be fixed at all (we can only assume that the one reading the docs will understand this) but it's a chance to change the diff to JS and highlight it like @DSchau said!

docs/docs/typography-js.md
```diff
@@ -14,19 +14,18 @@ You can install the plugin and its peer dependencies into your project by runnin
...
-+ }
-],
+  plugins: [
+    // highlight-start
+    {
+      resolve: `gatsby-plugin-typography`,
+      options: {
+        pathToConfigModule: `src/utils/typography`
+      }
+    }
```
muescha:
```suggestion
    },
```

I think it don't hurts to have here a comma

PR #17679, opened 2019-09-16
Edited for end-user stylesheets

### LINKED ISSUE #15232: [docs][guides] add details about user stylesheets with Styled Components
## Summary

In feedback from gatsbyjs.org, the [Styled Components](https://www.gatsbyjs.org/docs/styled-components/) styling doc could use a little more information around user stylesheets. Specifically:

> in the last part where would a user write their CSS for the line "A site user could then write their own CSS styles matching HTML elements with a class name of .container, and it wouldn’t be affected if the CSS-in-JS output changed."

It could be that the term "user" is what's confusing here, since this section is referring to end-users creating their own stylesheets for accessibility and other purposes. I have three recommendations:

- Refer to users in this scenario as "end-users" to differentiate from "Gatsby users" (developers)
- Include a short CSS snippet that references the `.container` class, and give it a file title like `css:title=user-stylesheet.css`.
- Add a link to more information on writing and including user stylesheets. Here's a recent-ish-looking post we could link to by Chris Coyier: https://mediatemple.net/blog/tips/bend-websites-css-will-stylish-stylebot/

kevee:
## Description

More accurately described who constitutes "users" for user stylesheets, and included an example of user stylesheets.

## Related Issues

Fixes #15232


docs/docs/styled-components.md
```diff
@@ -114,7 +114,7 @@ export default () => (
 
 ### Enabling user stylesheets with a stable class name
 
-Adding a persistent CSS `className` to your styled components can make it easier for users to take advantage of [User Stylesheets](https://www.viget.com/articles/inline-styles-user-style-sheets-and-accessibility/) for accessibility.
```
kevee: I personally don't think that "user stylesheets" should be capitalized, but I'm not super committed either way.

PR #17681, opened 2019-09-16
feat(gatsby): Enable shadowing of page template queries

### LINKED ISSUE #16927: Shadowing templates doesn't allow you to override the page query
## Description

Shadowing `gatsby-theme-blog-core/src/templates/post-query.js` lets you shadow the component, but doesn't let you shadow the page query.

### Steps to reproduce

```
gatsby new shadow-templates https://github.com/gatsbyjs/gatsby-starter-blog-theme
```

shadow `gatsby-theme-blog-core/src/templates/post-query.js` and use whatever query you want (I used `site { siteMetadata { author }}`). You'll get a "this query is not used" warning pointing at the shadowed file.

### Expected result

the "this query is not used" warning should point at the node_module file, `gatsby-theme-blog-core/src/templates/post-query.js` since that query would be no longer used.

### Actual result

warns on the shadowed file

ChristopherBiscardi:
## Description

fixes #16927 

Enables shadowing for createPage calls with respect to template queries. This is a shadowing algorithm application *outside* of webpack and as such needed explicit support.

### DIFF
--- a/e2e-tests/themes/development-runtime/package.json
+++ b/e2e-tests/themes/development-runtime/package.json
@@ -4,7 +4,7 @@
   "version": "1.0.0",
   "author": "Sidhartha Chatterjee <sid@gatsbyjs.com>",
   "dependencies": {
-    "gatsby": "gatsby-dev",
+    "gatsby": "^2.13.14",
     "gatsby-theme-about": "*",
     "react": "^16.8.0",
     "react-dom": "^16.8.0"
@@ -27,7 +27,7 @@
   "devDependencies": {
     "cross-env": "^5.2.0",
     "cypress": "^3.1.3",
-    "gatsby-cypress": "gatsby-dev",
+    "gatsby-cypress": "^0.2.2",
     "is-ci": "^2.0.0",
     "prettier": "^1.15.2",
     "start-server-and-test": "^1.7.11"
--- a/e2e-tests/themes/development-runtime/package.json
+++ b/e2e-tests/themes/development-runtime/package.json
@@ -4,7 +4,7 @@
   "version": "1.0.0",
   "author": "Sidhartha Chatterjee <sid@gatsbyjs.com>",
   "dependencies": {
-    "gatsby": "^2.13.14",
+    "gatsby": "gatsby-dev",
     "gatsby-theme-about": "*",
     "react": "^16.8.0",
     "react-dom": "^16.8.0"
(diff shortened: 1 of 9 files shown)


packages/gatsby/src/redux/actions/public.js
```diff
@@ -226,6 +227,45 @@ ${reservedFields.map(f => `  * "${f}"`).join(`\n`)}
     }
   }
 
+  // TODO: start shadowing implementation
+  const shadowingPlugin = new GatsbyThemeComponentShadowingResolverPlugin({
+    // TODO: extensions needs to be sourced from store somehow
+    extensions: [`.js`],
```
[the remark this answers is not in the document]
ChristopherBiscardi: also move this logic to separate file

ChristopherBiscardi:
@gatsbyjs/core this is rebased and ready for review

pieh:
Thanks! Sorry for the delay - code LGTM

I do have worries about adding/removing shadowed page templates while `gatsby develop` is running, but this part is already problematic with webpack plugin so I don't want this to block the PR.

I will add test cases for e2e themes tests

e2e-tests/themes/development-runtime/package.json
```diff
@@ -4,7 +4,7 @@
   "version": "1.0.0",
   "author": "Sidhartha Chatterjee <sid@gatsbyjs.com>",
   "dependencies": {
-    "gatsby": "^2.13.14",
+    "gatsby": "gatsby-dev",
```
pieh: oooopsie

sidharthachatterjee: 😂

pieh:
Published `gatsby@2.18.1`

laurenskling:
Can confirm this is working for my project <3 thanks guys

jsg2021:
I hit an edge case. Create a site from `@lekoarts/gatsby-theme-minimal-blog` and attempt to shadow the `template/homepage-query.tsx` with a local `src/@lekoarts/gatsby-theme-minimal-blog-core/templates/homepage-query.js`. It won't shadow over the in-theme file. Rename it with the same extension...it works.

the documents for `createPage` tells you to use an absolute path to the component. So, the theme author didn't do anything wrong. (per docs) The component will always have an extension, so the algorithm resolving the shadowed component will not allow extension overriding. :(

Kalmuraee:
I have the same problem shadowing `data.query.js` in tsx template :(

PR #17682, opened 2019-09-17
fix(query): improve query performance

me4502:
# Description

One of the big caveats currently of large sites in Gatsby is having to use gatsby-node when performing large numbers of queries. This has downsides, such as losing the ability to reload pages when the data changes.

I encountered an issue recently which made it much more preferable to use page queries, so I've looked into /why/ this is slow.

This PR is only a start, but it's taken 18k page queries from 11 minutes to 4.5 minutes, so over a 50% reduction in build time, which I feel is a decent improvement.

### DIFF
--- a/packages/gatsby/src/query/index.js
+++ b/packages/gatsby/src/query/index.js
@@ -36,11 +36,15 @@ const popExtractedQueries = () => {
 }
 
 const findIdsWithoutDataDependencies = state => {
-  const allTrackedIds = new Set(
-    _.flatten([
-      ...state.componentDataDependencies.nodes.values(),
-      ...state.componentDataDependencies.connections.values(),
-    ])
+  const allTrackedIds = new Set()
+  const boundAddToTrackedIds = allTrackedIds.add.bind(allTrackedIds)
+  state.componentDataDependencies.nodes.forEach(dependenciesOnNode => {
+    dependenciesOnNode.forEach(boundAddToTrackedIds)
+  })
+  state.componentDataDependencies.connections.forEach(
+    dependenciesOnConnection => {
+      dependenciesOnConnection.forEach(boundAddToTrackedIds)
+    }
   )
 
   // Get list of paths not already tracked and run the queries for these
--- a/packages/gatsby/src/query/index.js
+++ b/packages/gatsby/src/query/index.js
@@ -72,7 +72,11 @@ const popNodeQueries = state => {
 
     // Find components that depend on this node so are now dirty.
     if (state.componentDataDependencies.nodes.has(node.id)) {
-      state.componentDataDependencies.nodes.get(node.id).forEach(dirtyIds)
+      state.componentDataDependencies.nodes.get(node.id).forEach(n => {
+        if (n) {
+          dirtyIds.add(n)
+        }
+      })
     }
 
     // Find connections that depend on this node so are now invalid.
--- a/packages/gatsby/src/query/index.js
+++ b/packages/gatsby/src/query/index.js
@@ -8,16 +8,14 @@ const { boundActionCreators } = require(`../redux/actions`)
 const queryQueue = require(`./queue`)
 const GraphQLRunner = require(`./graphql-runner`)
 
-let seenIdsWithoutDataDependencies = []
+const seenIdsWithoutDataDependencies = new Set()
 let queuedDirtyActions = []
 const extractedQueryIds = new Set()
 
 // Remove pages from seenIdsWithoutDataDependencies when they're deleted
 // so their query will be run again if they're created again.
 emitter.on(`DELETE_PAGE`, action => {
-  seenIdsWithoutDataDependencies = seenIdsWithoutDataDependencies.filter(
-    p => p !== action.payload.path
-  )
+  seenIdsWithoutDataDependencies.delete(action.payload.path)
 })
 
 emitter.on(`CREATE_NODE`, action => {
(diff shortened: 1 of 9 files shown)


KyleAMathews:
Wow! This is awesome.

Could you check how this improves the markdown benchmark scores? https://github.com/gatsbyjs/gatsby/blob/master/benchmarks/markdown/README.md

me4502:
I just ran the benchmark, it doesn't seem to be as dramatic in there (assuming I ran it correctly)

I've been testing with my normal site in develop though, and I feel this would make more of an impact there.

Here are the benchmark times for build,

```
Before: 424.935
After: 381.775
```

I ran it, then checked out the other branch then ran it again - is that correct?

KyleAMathews:
That's a good way to test things.

Query running used to at least have more caching than during development. That might have changed with all the changes @gatsbyjs/core has been doing.

me4502:
I've worked out why it's much faster in my site - due to extra data dependencies this code gets called a lot more often for the same page count.

me4502:
A vast majority of the time is now spent within Sift itself. I ran a test with Loki, and my local site that was 11 minutes before this change (using Sift), is now 80 seconds with Loki.

me4502:
So just as a final summary of this PR:

Using Loki, queries are almost instant from my testing. The second the query needs to offload to Sift it becomes slower. It's still magnitudes faster than it was before this PR, but Sift now appears to be the biggest slowdown.

muescha:
In node-model you also changed to `new Set() ... Has` from `includes`
Maybe revert this also

me4502:
Ah thanks, I’ll do that when I’m back at a computer (if a Gatsby team member wants to merge before then feel free to change those two instances in node-model)

packages/gatsby/src/query/index.js
```diff
@@ -38,54 +36,62 @@ const popExtractedQueries = () => {
...
-      dirtyIds = dirtyIds.concat(state.componentDataDependencies.nodes[node.id])
+  const uniqDirties = actions.reduce((dirtyIds, action) => {
+    const node = action.payload
+
+    if (!node || !node.id || !node.internal.type) return dirtyIds
+
+    // Find components that depend on this node so are now dirty.
+    if (state.componentDataDependencies.nodes.has(node.id)) {
+      state.componentDataDependencies.nodes.get(node.id).forEach(n => {
+        if (n) {
```
pieh: This is interesting - did you see some falsy entries in there that you needed to add this check?

pieh: If this is the case then there is some bug somewhere:
- it can be be not problematic (i.e. we try to add dependency from running queries in gatsby-node, which will never have query identifier and we should just not add dependencies in those cases) 
- or might be very problematic if page/static queries don't add dependency properly in some cases, and so we don't properly invalidate query results for them

pieh: It's not in the scope of this PR to fix this, but it's nice that it potentially encovered problem

me4502: Yeah I was getting an error that undefined was being added into the Set. It’s possible it was caused by another issue I introduced and later fixed while creating the PR. I’ll run it without this line in the morning.

me4502:
Ah that makes sense, thanks. I’ll make the query a bit more complex in the benchmark in the morning as well :)

KyleAMathews:
Really exciting to see the speedups on your site! We're looking to make Loki default soon so other sites can see the benefits. Thanks for all your investigation work on this.

packages/gatsby/src/query/index.js
me4502: Yeah seems it now works without it - I'll remove it.

me4502: Nvm seems the E2E tests fail with it removed :^)

me4502: Before this was fine as arrays accept undefined, and the _.compact was clearing them out. It seems this issue must've been encountered in the past, hence why _.compact was added in the first place.

packages/gatsby/src/query/index.js
```diff
@@ -72,11 +72,7 @@ const popNodeQueries = state => {
 
     // Find components that depend on this node so are now dirty.
     if (state.componentDataDependencies.nodes.has(node.id)) {
-      state.componentDataDependencies.nodes.get(node.id).forEach(n => {
-        if (n) {
-          dirtyIds.add(n)
-        }
-      })
+      state.componentDataDependencies.nodes.get(node.id).forEach(dirtyIds)
```
pieh: I think this would need to `.forEach(dirtyIds.add)` (there was missing actual `.add` method (binding might also be necessary)

me4502: Ah, missed that yeah... I've just pushed it with .add and explicitly binding the n

me4502: Yeah it works now 🙃

packages/gatsby/src/query/index.js
```diff
@@ -38,54 +36,56 @@ const popExtractedQueries = () => {
...
-      _.concat(
-        _.values(state.componentDataDependencies.nodes),
-        _.values(state.componentDataDependencies.connections)
-      )
-    )
+  const allTrackedIds = new Set(
+    _.flatten([
+      ...state.componentDataDependencies.nodes.values(),
+      ...state.componentDataDependencies.connections.values(),
+    ])
```
pieh: This seems to be problematic:
this PR on the left vs latest/master on the right
Screenshot 2019-09-20 at 18 59 49

Currently it doesn't properly flatten our data structure, which means that it deems all ids to be untracked and cause to run all queries

pieh: Alternative to change above is:
```js
  const allTrackedIds = new Set()
  const boundAddToTrackedIds = allTrackedIds.add.bind(allTrackedIds)
  state.componentDataDependencies.nodes.forEach(dependenciesOnNode => {
    dependenciesOnNode.forEach(boundAddToTrackedIds)
  })
  state.componentDataDependencies.connections.forEach(
    dependenciesOnConnection => {
      dependenciesOnConnection.forEach(boundAddToTrackedIds)
    }
  )
```
Which seems to the trick (at least for 1to1 conversion with current algorithm)

packages/gatsby/src/redux/actions/add-page-dependency.js
```diff
@@ -24,7 +23,7 @@ function createPageDependency({ path, nodeId, connection }) {
   }
   if (
     connection &&
-    _.includes(state.componentDataDependencies.connections, connection)
+    state.componentDataDependencies.connections.has(connection)
```
pieh: Ok, this is actually causing some problems - in this check we never do any `path` check (and this is broken in master as well, but in master the condition actually never is true, so we might do some extra work, but it doesn't actually break data tracking.

With this change this become broken - we should use same approach as in `nodeId` check above:
```suggestion
    state.componentDataDependencies.connections.has(connection) &&
    state.componentDataDependencies.connections.get(connection).has(path)
```

me4502: Done

packages/gatsby/src/query/index.js
me4502: Done

me4502:
Yeah I'm happy with you pushing to the branch :)

pieh:
Hmm, seems like I can't push - can you check if "Allow edits from maintainers" is checked on this PR ( https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests )

me4502:
Hmm, that option appears to already be checked. It may be a setting on the organisation that the fork was made in preventing pushes, I don't think I have the ability to change those settings. If you send a patch file I could apply it?

pieh:
I hope to finish tests ( https://github.com/gatsbyjs/gatsby/pull/17813 ) soon and merge that one in too (just verifying if my changes there to mock persistent loki storage don't break regular usage of loki there now)

packages/gatsby/src/redux/reducers/component-data-dependencies.js
```diff
@@ -1,46 +1,51 @@
...
+          existingPaths = state.nodes.get(action.payload.nodeId)
+        }
+        if (!existingPaths.has(action.payload.path || action.payload.id)) {
+          existingPaths.add(action.payload.path || action.payload.id)
         }
-        const newPaths = _.uniq(
-          existingPaths.concat(action.payload.path || action.payload.id)
-        )
-        state.nodes[action.payload.nodeId] = newPaths
+        state.nodes.set(action.payload.nodeId, existingPaths)
```
pieh: I wonder if we could something more here to avoid initialising empty sets, and setting already set fields in set - maybe something like:

```js
let existingPaths = state.nodes.get(action.payload.nodeId)
if (!existingPaths) {
  existingPaths = new Set()
  state.nodes.set(action.payload.nodeId, existingPaths)
}

// not sure if `has` check is even needed here
if (!existingPaths.has(action.payload.path || action.payload.id)) {
  existingPaths.add(action.payload.path || action.payload.id)
}
```

pieh: and similar for the `connection` case

pieh:
Tests pr was merged - could you rebase again?

packages/gatsby/src/redux/reducers/component-data-dependencies.js
pieh: Let's skip this change, it doesn't seem to have any effect and keeping it more 1to1 gives me more confidence

pieh:
I think this is in good spot right now, let's just wait for
> Tests pr was merged - could you rebase again?

and we can merge this in

me4502:
I rebased, and the checks have all passed :)

pieh:
Published `gatsby@2.15.24`

sw-yx:
this is awesome!! question from the audience.. what is Loki and Sift?

jonniebigodes:
@sw-yx [sift](https://www.npmjs.com/package/sift) and [lokijs](https://www.npmjs.com/package/lokijs)

PR #17683, opened 2019-09-17
fix(gatsby-plugin-sitemap): Removed redundant check for siteUrl

dhakerShivnarayan:


wardpeet:
Published in gatsby-plugin-sitemap@2.2.13

PR #17693, opened 2019-09-17
fix(gatsby-plugin-sharp): Set a unique uuid for each job

sidharthachatterjee:
This was a bug we discovered while working on structured logging where multiple jobs for the same image (for instance a fluid version and a fixed version of the same image) create _two_ jobs (as one would expect) but the jobs have identical `id` fields and there is no way to differentiate them. 

We fix this by generating a `uuid` for every new job.


wardpeet:
Published in gatsby-plugin-sharp@2.2.23

PR #17694, opened 2019-09-17
docs: convert console/shell/terminal into shell

wardpeet:
# Description
Follow up on #17674 to replace all shell like languages into shell.


sagarhani:
Thanks. LGTM. 👍 
(My bad. I fixed only for those pages I read on docs. I should have replaced all.)

muescha:
Is there any MD linter checks to ensure not to use the `console|bash|sh`?
(Maybe with https://github.com/remarkjs/remark-lint )

PR #17717, opened 2019-09-18
chore(showcase): Add Fifteen Seconds Istanbul

ibrahimtelman:


lannonbr:
Hey @ibrahimtelman so the source_url seems to point to a private repo, so I would say either make the repo public or remove the source_url from the entry.

ibrahimtelman:
Hi, I've removed the source_url. thanks for review

PR #17723, opened 2019-09-18
fix(babel-preset-gatsby-package): Add corejs to config

gonzarodriguezt:
## Description

This is a warning in development about a missing `corejs` config in babel, the issue below tracks down the source of the problem, and this fix quiets it down, thanks!

Edit: I specified the version 3 as it seems to be the recommended version on the docs [(here)](https://github.com/zloirock/core-js#babelpreset-env)

## Related Issues

Fixed #17706

PR #17727, opened 2019-09-18
fix(babel-preset-gatsby-package): apply "corejs" only to non-browser target

sidharthachatterjee:
Appears to have been introduced in https://github.com/gatsbyjs/gatsby/pull/17723

### DIFF
--- a/packages/babel-preset-gatsby-package/__tests__/__snapshots__/index.js.snap
+++ b/packages/babel-preset-gatsby-package/__tests__/__snapshots__/index.js.snap
@@ -5,7 +5,6 @@ Array [
   Array [
     "@babel/preset-env",
     Object {
-      "corejs": 2,
       "debug": false,
       "loose": true,
       "modules": "commonjs",
@@ -17,7 +16,7 @@ Array [
           "not android 4.4.3",
         ],
       },
-      "useBuiltIns": "entry",
+      "useBuiltIns": false,
     },
   ],
   Array [
@@ -35,7 +34,6 @@ Array [
   Array [
     "@babel/preset-env",
     Object {
-      "corejs": 2,
       "debug": true,
       "loose": true,
       "modules": "commonjs",
@@ -47,7 +45,7 @@ Array [
           "not android 4.4.3",
         ],
       },
-      "useBuiltIns": "entry",
+      "useBuiltIns": false,
     },
   ],
   Array [
@@ -137,7 +135,6 @@ Array [
   Array [
     "@babel/preset-env",
     Object {
-      "corejs": 2,
       "debug": false,
       "loose": true,
       "modules": "commonjs",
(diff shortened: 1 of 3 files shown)


gonzarodriguezt:
Sorry! I should committed the updates on `yarn.lock` after bumping `core-js` to ^2.6.9?

sidharthachatterjee:
@gonzarodriguezt No problem at all! Yeah, the `yarn.lock` needed to be updated but also found an unusual setting for `useBuiltIns` in browser based environments. I'm not hundred percent sure of this change so let's wait on @wardpeet who is our tooling expert

gonzarodriguezt:
Oh okay @sidharthachatterjee let's wait then 🙏

pieh:
This looks quite weird, I don't think we want to add polyfills to browser code we ship in our npm packages? We only want to add them for site bundles (which is handled by `babel-preset-gatsby` (and not `babel-preset-gatsby-package`)?

sidharthachatterjee:
> This looks quite weird, I don't think we want to add polyfills to browser code we ship in our npm packages? We only want to add them for site bundles (which is handled by babel-preset-gatsby (and not babel-preset-gatsby-package)?

@pieh That _does_ make sense! So then if that `browser` is true, we want to remove the `corejs` field and override `useBuiltIns` to false?

pieh:
What is this exactly fixing? I think we are missing some context around why this is needed

sidharthachatterjee:
Good question, the error in https://circleci.com/gh/gatsbyjs/gatsby/226290?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

pieh:
> @pieh That _does_ make sense! So then if that `browser` is true, we want to remove the `corejs` field and override `useBuiltIns` to false?

Ok, so I think this make sense to add `corejs: 2` to `nodeConfig` only and not to all common config, and keep `browserConfig` as is currently (so revert that part)

sidharthachatterjee:
> Ok, so I think this make sense to add corejs: 2 to nodeConfig only and not to all common config, and keep browserConfig as is currently (so revert that part)

Yup, that is the cleanest way to go. Let me update this.

pieh:
And incidentally, `gatsby-transformer-sqip`'s `.babelrc` shouldn't use `browser` variant ;) We should audit .babelrc files really for this stuff

sidharthachatterjee:
@pieh Done

gonzarodriguezt:
You're awesome guys! Hopefully my next contrib will be less messy 😅thank you!

pieh:
> You're awesome guys! Hopefully my next contrib will be less messy 😅thank you!

Don't worry about it! We didn't catch that in review either!

sidharthachatterjee:
> You're awesome guys! Hopefully my next contrib will be less messy 😅thank you!

You did great! And thank you so much 🤗

PR #17732, opened 2019-09-19
blog: update graphiql improvements blog post with code exporter section

pieh:
this follows up in the blog post about graphiql improvements in gatsby

preview:
localhost_8001_blog_2019-06-03-integrating-graphiql-explorer_

/cc @herecydev ;)


KyleAMathews:
Why not make this a new blog post?

muescha:
We should also add a section in [Introducing GraphiQL](https://www.gatsbyjs.org/docs/introducing-graphiql/) in one step?

pieh:
> Why not make this a new blog post?

This was suggestion from @hashimwarren when we chatted about adding some content that can referenced when tweeting about code exporter feature:

> in august over about 116 people found that post through search, as they looked for info on graphiql :chart_with_upwards_trend:. So, I think updating has value, over creating a new post, or not updating

> We should also add a section in [Introducing GraphiQL](https://www.gatsbyjs.org/docs/introducing-graphiql/) in one step?

I think documentation should be updated, but not sure if trying to do both in single PR is best way to do this - IMO trying to limit scope of PRs is best way to ensure they are handled faster

Issue #3817, opened 2018-02-02
v2 - Don't generate source maps by default


tomazy:
## Description

Gatsby by default generates very expressive source maps. They include the full, non-minified source code of the site, including all comments like TODO, etc. They also expose absolute paths to the project directory leaking things like user names and naming conventions. 

Source maps in general should not be available to anyone willing to open the dev tools in their browser. Access to them should be restricted and this requires some additional adjustments on the web server level. 

I suggest we disable the default source map generation in the next version of the project.

See: https://github.com/gatsbyjs/gatsby/issues/3324#issuecomment-361401512

monsieurnebo:
A nice way of handling sourcemaps could be a config parameter, such as:

```js
// gatsby-config.js

module.exports = {

  pathPrefix    : "/blog",
  sourceMaps : true,
  // ...

}
```

It would let users choose when and where use them and would be particulary useful on multi-environments development (develop, staging, etc...).

jquense:
I don't think not generating source maps the right approach. For every production app I've made we've wanted them either to load when devtools were opened or to upload to sentry for clearer error tracking. I think if you don't want them in your uploaded site that's more of publishing concern than a building on?

KyleAMathews:
Instead of disabling sourcemaps by default — how about instead we a) investigate using relative paths, removing comments and otherwise making them more tidy by default and b) add a plugin to easily disable producing source maps by modifying the webpack config which also conveniently would speed up builds some.

monsieurnebo:
@KyleAMathews I cannot see the advantage of a plugin instead of a basic gatsby-config parameter? 🤔

KyleAMathews:
@monsieurnebo mainly because we avoid config options at (almost) all costs :-)

Adding config is a much weightier decision vs. a plugin. Also plugins are much more flexible than config.

monsieurnebo:
Ok! Would this plugin's purpose be only to disable sourcemaps via webpack configuration? Isn't that a bit overkill?

KyleAMathews:
@monsieurnebo that's how a config option would work so it's the same thing

siakaramalegos:
So it looks like that might actually work. Either my finder wasn't refreshing the file list or the /public/folder was holding on to old sourcemap files.

github0013:
This solved removing source map in v2.
https://github.com/gatsbyjs/gatsby/issues/6289

```bash
yarn add gatsby-plugin-no-sourcemaps@next
```

```json
[...]

PR #17733, opened 2019-09-19
Added ternary operator to allow sourcemaps on gatsby-cli develop

BlackFenix2:
## Description

I added a Ternary operator to allow the source-map devtool in development environments while using the 'gatsby-plugin-no-sourcemaps' plugin.

## Why did i make this PR?

Sourcemaps are still totes useful for debugging when running gatsby develop

## Could this be better?

yeah, if i could pass in a plugin option to configure sourcemap generation 

Ex:
```js
{
  resolve: `gatsby-plugin-no-sourcemaps`,
  options: {
    allowDev:true,
  }
}
```

### why didn't you make this better 👀 

... i don't know how to take plugin options as an argument 🤷‍♂. i'm welcome to suggestions.

## Related Issues

kind of a half-way fix for https://github.com/gatsbyjs/gatsby/issues/3817

### DIFF
--- a/packages/gatsby-plugin-no-sourcemaps/gatsby-node.js
+++ b/packages/gatsby-plugin-no-sourcemaps/gatsby-node.js
@@ -1,7 +1,7 @@
 exports.onCreateWebpackConfig = ({ stage, actions }) => {
   if (stage === `build-javascript`) {
     actions.setWebpackConfig({
-      devtool: false,
-    })
+      devtool: process.env.NODE_ENV === 'development' ? 'source-map' : false,
+    });
   }
 }


wardpeet:
Hey,

Thanks for your PR. It's better to do this condition inside gatsby-config and don't load the plugin on development.

You can checkout https://github.com/gatsbyjs/gatsby/blob/f5942535321dd59ba960165c358e02ec2c864bee/www/gatsby-config.js#L40-L54 on how we do this in gatsbyjs.org.

```js
const plugins = [];

if (process.env.NODE_ENV !== 'development') {
   plugins.push(`gatsby-plugin-no-sourcemaps`)
}

module.exports = {
  // gatsby config
  plugins,
};
```

PR #17735, opened 2019-09-19
fix(gatsby): add timeout so fetching API info doesn't fail on very slow connections

KyleAMathews:
I was on an airplane trying to run `gatsby develop` and bootstrap hung here. I dug around and found adding a timeout fixed it.

### DIFF
--- a/packages/gatsby/src/bootstrap/load-plugins/index.js
+++ b/packages/gatsby/src/bootstrap/load-plugins/index.js
@@ -38,14 +38,11 @@ const flattenPlugins = plugins => {
 }
 
 module.exports = async (config = {}, rootDir = null, latestAPIs = {}) => {
-  const apis = {
-    currentAPIs: getAPI({
-      browser: browserAPIs,
-      node: nodeAPIs,
-      ssr: ssrAPIs,
-    }),
-    latestAPIs,
-  }
+  const currentAPIs = getAPI({
+    browser: browserAPIs,
+    node: nodeAPIs,
+    ssr: ssrAPIs,
+  })
   // Collate internal plugins, site config plugins, site default plugins
   const plugins = loadPlugins(config, rootDir)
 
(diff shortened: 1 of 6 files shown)


muescha:
I am right that the latestApis only used in:

https://github.com/gatsbyjs/gatsby/blob/194a97c3854e3b8e4fcdd5b1eab1d72450dea867/packages/gatsby/src/bootstrap/load-plugins/validate.js#L35-L41

And only in case of error:

https://github.com/gatsbyjs/gatsby/blob/194a97c3854e3b8e4fcdd5b1eab1d72450dea867/packages/gatsby/src/bootstrap/load-plugins/validate.js#L106-L115


Errors are not so much here. 

That's why I ask: 

1) why we need to fetch on every `gatsby develop` the `lastestApis` from an external source?
2) would it be enough only in case of error to fetch this data?
3) why this latestapis need to stay on an external server?

KyleAMathews:
>would it be enough only in case of error to fetch this data?

Good question. @dschau didn't you add this originally? Thoughts?

DSchau:
@KyleAMathews yep -> https://github.com/gatsbyjs/gatsby/pull/16105#discussion_r307553931

Wouldn't _reasonably slow_ connections short-circuit here and surpass the 1s limit? Perhaps that's OK.

@muescha I like your idea to only fetch these APIs on error. That's a nicer solution.

@KyleAMathews I'll PR some tweaks and get this merged!

DSchau:
@KyleAMathews I tried a new approach in 4f812d7.

This will only make the API request when bad exports are detected, which seems like a better approach. I added a few tests, and validated locally.

Seems fine and would solve the underlying issue you identified as the API request will no longer be made in bootstrap _unless_ there are bad exports.

muescha:
in case of timeout:
- it would be not good to have just an fetch timeout error
-  it would be good to print then an describing error message about the signature check and the user should byself lookup at the file later - so he can go on and debug and can see easy errors like small typos

muescha:
Maybe add also some meta data to the `apis.json` like:

`You are running now gatsby version 2.0 there is a newer version 2.1 available.`

KyleAMathews:
Looks great! Yeah maybe have a timeout of e.g. five seconds. You could still fetch the data on bootstrap but just don't make the call blocking.

packages/gatsby/src/bootstrap/load-plugins/index.js
```diff
@@ -38,14 +38,11 @@ const flattenPlugins = plugins => {
 }
 
 module.exports = async (config = {}, rootDir = null, latestAPIs = {}) => {
```
wardpeet: latestAPIs seems to be unused
```suggestion
module.exports = async (config = {}, rootDir = null) => {
```

DSchau:
@muescha 

> it would be not good to have just an fetch timeout error

It falls through to the catch block with `ECONABORTED`, which will grab the (hopefully cached) or current apis.json.

> Maybe add also some meta data to the apis.json So we can print also infos like:
>
> You are running now gatsby version 2.0 there is a newer version 2.1 available.

This is what it does. If you're using an API in a later version of Gatsby (but with an older version) you'll get a prompt as to what version it was introduced in.

I've added a timeout (5s seems _fair_), this should be good to go.

DSchau:
Going to merge and publish -- thanks Kyle for tagging me in!

DSchau:
Successfully published gatsby@2.15.20

PR #17742, opened 2019-09-19
chore(gatsby-design-tokens): update bundle config

wardpeet:
## Description
Introducing better bundling for gatsby-design-token. Currently, the package isn't 100% tree-shakeable. There are a few side-effects that keep this from happening. I've added microbundle to bundles things together to one file.

I'm outputting a cjs & es module so it can be used in the browser & node like environments.

Agadoo is a tool to check if you're 100% tree-shakeable. It's pretty good if you're exposing a library. It's great to make sure we don't regress.

As seen in this Pr, I had a few issues to make it tree-shakeable so I introduced preval macro to precompile a few things.

Sadly microbundle needs node `10.16.0` so i'm pretty sure CI will fail... it's a dev dependency so I'm not sure how to go forward with it.

The reason for this PR is a prerequisite to improve bundling of https://github.com/gatsby-inc/gatsby-interface

Size:
```
Build "gatsbyDesignTokens" to dist:
       3940 B: index.js
       1908 B: index.js.gz
       1581 B: index.js.br
       3920 B: index.esm.js
       1939 B: index.esm.js.gz
       1607 B: index.esm.js.br
```

This is what bundlephobia shows atm:
https://bundlephobia.com/result?p=gatsby-design-tokens@1.0.5
image


pieh:
Do we need microbundle 0.12? 0.11 works on node 8

wardpeet:
For this module no, but we will run into this issue eventually when we'll be doing modern compilation for gatsby-image and other gatsby packages.

I'm not sure it's a big deal for a devDependency? We could just kill azure as we have windows ci on circleci. I don't think we need to build assets on node 8.

pieh:
> We could just kill azure as we have windows ci on circleci.

Or bump node version on azure to not spend time on figuring out how to disable it ;)

Let's do either one of those and get this in?

packages/gatsby-design-tokens/package.json
```diff
@@ -2,27 +2,30 @@
...
     "type": "git",
     "url": "https://github.com/gatsbyjs/gatsby.git",
     "directory": "packages/gatsby-design-tokens"
   },
+  "files": [
+    "./dist/**/*"
+  ],
   "scripts": {
-    "build": "babel src --out-dir . --ignore **/__tests__",
+    "build": "microbundle build --sourcemap=false -f es,cjs && agadoo dist/index.esm.js",
```
pieh: why no source maps?

wardpeet: wasn't sure we needed it. I don't know if sourcemaps are being shipped with node_modules. Happy to enable it :joy:

pieh: I thought we do for our browser packages, but after checking it, apparently we don't. Let's keep it as-is and maybe we can change it separately

pieh:
This is good, for DX let's add a note if build / watch file that you need Node >=10.16 for now and maybe later try PR to microbundle so maybe it doesn't fail completely if `zlib_1.brotliCompress` is not a function as it's used only to generate stats and doesn't impact actual bundling process

PR #17752, opened 2019-09-19
fix(gatsby): support `export { X as default }` syntax when checking if page template files have default export

Schmavery:
## Description
`gatsby build` checks to see that all pages export a default react component. It does this by searching the file for strings like `module.exports`, `exports.default`,  `exports["default"]`  etc.

While this covers many of the [possible ways](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export) to export a default value, one is missed:
```js
export { name1 as default, … };
``` 

Unfortunately, this is the form used by [bucklescript](https://bucklescript.github.io/), a ReasonML-to-JS compiler that I'm trying to use with gatsby. 

Bucklescript generates code for default export that looks something like this:
```js
var $$default = Index;
export {
  Styles ,
  make ,
  $$default ,
  $$default as default,
}
```

Looks a little funny but the main point is that it uses the  ` as default` syntax. I tried to add detection for this case, it seems to be working locally on my machine. Not sure how precise you want this check to be, we could also do something simpler like `!fileContent.includes(' as default')`.

Thanks for taking a look and let me know if I misunderstood anything!

## Related Issues
Fixes https://github.com/Schmavery/reason-gatsby/issues/1

### DIFF
--- a/packages/gatsby/src/redux/actions/public.js
+++ b/packages/gatsby/src/redux/actions/public.js
@@ -379,7 +379,7 @@ ${reservedFields.map(f => `  * "${f}"`).join(`\n`)}
       !fileContent.includes(`module.exports`) &&
       !fileContent.includes(`exports.default`) &&
       !fileContent.includes(`exports["default"]`) &&
-      !fileContent.match(/export {.* as default.*}/s) &&
+      !fileContent.match(/export \{.* as default.*\}/s) &&
       // this check only applies to js and ts, not mdx
       /\.(jsx?|tsx?)/.test(path.extname(fileName))
     ) {
--- a/packages/gatsby/src/redux/actions/public.js
+++ b/packages/gatsby/src/redux/actions/public.js
@@ -379,6 +379,7 @@ ${reservedFields.map(f => `  * "${f}"`).join(`\n`)}
       !fileContent.includes(`module.exports`) &&
       !fileContent.includes(`exports.default`) &&
       !fileContent.includes(`exports["default"]`) &&
+      !fileContent.match(/export {.* as default.*}/s) &&
       // this check only applies to js and ts, not mdx
       /\.(jsx?|tsx?)/.test(path.extname(fileName))
     ) {


Schmavery:
Looks like babel is dying on my regex, I'll try to debug locally unless you think it's ok to just use `!fileContent.includes(' as default')`

packages/gatsby/src/redux/actions/public.js
```diff
@@ -379,6 +379,7 @@ ${reservedFields.map(f => `  * "${f}"`).join(`\n`)}
       !fileContent.includes(`module.exports`) &&
       !fileContent.includes(`exports.default`) &&
       !fileContent.includes(`exports["default"]`) &&
+      !fileContent.match(/export \{.* as default.*\}/s) &&
```
pieh:
```suggestion
      !fileContent.match(/export\s*\{.*\s+as\s+default.*\}/s) &&
```
Not sure if this suggestion is totally correct, but I think this should match a bit more loose when it comes to whitespaces

so examples like:
```
export{X as default}
export     {X      as      default}
```
are also valid

pieh: (that said, I realize our current code - specifically one for `export default` doesn't handle this nicely)

Schmavery: Yeah I figured this was more or less true of almost all the checks and was trying to keep the regex simple.
Do you have any thoughts on the even simpler `!fileContent.includes('as default')`?

pieh: I think we can start with your current code (and not apply my suggestion) and think about making all of those tests to be more resilient when we will get reports of problems

Schmavery: sgtm, thanks :)

pieh:
Will publish new release in few minutes

pieh:
`gatsby@2.15.19` is published!

PR #17747, opened 2019-09-19
chore(docs): Add documentation for tailwindcss support

nextlevelshit:
# Description

I was missing a documentation how to implement the tailwindcss framework. After searching the web I found lot of blog posts about how to implement tailwindcss with postcss plugin, but not with sass/scss support.

This plugin looks to be perfect for this and makes it so easy to implement tailwindcss, but there is somethings to care for. That's why I tought about adding a brief instruction how to implement tailwindcss.

Additionally I've found some mistakes in the highlighting and some missing information about which file is displayed in the code samples. Instead of splitting it into different PR's I've done it at once to save some time.

PR #17762, opened 2019-09-20
Added instructions for Sass/SCSS implementation

nextlevelshit:
# Description

I was missing a documentation how to implement the tailwindcss framework. After searching the web I found lot of blog posts about how to implement tailwindcss with postcss plugin, but not with sass/scss support.

This plugin looks to be perfect for this and makes it so easy to implement tailwindcss, but there is somethings to care for. That's why I tought about adding a brief instruction how to implement tailwindcss.

Additionally I've found some mistakes in the highlighting and some missing information about which file is displayed in the code samples. Instead of splitting it into different PR's I've done it at once to save some time.

### DIFF
--- a/docs/docs/tailwind-css.md
+++ b/docs/docs/tailwind-css.md
@@ -6,12 +6,13 @@ Tailwind is a utility-first CSS framework for rapidly building custom user inter
 
 ## Overview
 
-There are two ways you can use Tailwind with Gatsby:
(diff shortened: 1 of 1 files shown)


LekoArts:
Please wait for the confirmation I mentioned here: https://github.com/gatsbyjs/gatsby/pull/17747#issuecomment-533532805

Also please then commit the changes to your current PR instead of opening a new one. Thanks!

PR #17779, opened 2019-09-21
add Spanish translations

horacioh:
## Description

First Spanish translations to the main repo

## Related Issues

as @tesseralis and I discussed in the Discord channel, I'm adding the current spanish translations to the main repo


jonniebigodes:
@horacioh looked over it and it's looking good. But one thing that came to mind while going over this pr, are going to copy over the image files for each language or can we find a more pragmatic approach to it. Probably having only one copy of them and reference them for each language? Is it doable? Also i believe it would benefit new translators in their workflow and the repo by preventing the already considerable size to inflate even more. 

@tesseralis as you´re  spearheading the initiative your thoughts on this? and also the gatsby teams involved care to chime in?

tesseralis:
@jonniebigodes I also agree that it's important to figure out how to prevent images from being copied excessively.

I think this is a general question too: there are a lot of links that refer to different pages, some of which aren't translated yet. I wonder if we can change the routing logic to be something like:

* If the link (to an image or another page) exists in this language, use it
* Otherwise, reroute to the English version

That said, there is the possibility of copying over images if those images have English text on them that we would like to translate.

For now, @horacioh could you remove the image files until we figure out what to do?

tesseralis:
One thing I'm realizing is that, with React, we copied over the entire repo, so when you made a translation, you had the original version to compare against in the PR (for example: https://github.com/reactjs/es.reactjs.org/pull/10/files).

We don't get that here, so I'd imagine that it would be difficult to check that the translation is actually accurate. As someone who's had experience with this, what do you think @Darking360?

I'm not sure what the best solution is here. If we copy over the entire `docs/` directory for each language, the whole thing will just become bloated... Maybe we can just have the bot copy over one section at a time (like tutorials/).

horacioh:
you are right, copying the images does not sound like a good idea.

I can delete the images from the translation, do I need to change all the links or should I leave it as is?

Darking360:
Hey, just looking at this 👀  

For the images is already said, but yes, we should kind of centralize images in one place so they can be referenced from another place and don't have duplicated all over the languages, now I don't know if we're going to translate some images (I haven't take a look in-depth to check if there are images that would need it). And for links, I'm sure that there should be a way to do it 🤔 Don't Gatsby has like a manifest were all the pages that should be indexed are in? Something like `algolia` [here](https://github.com/algolia/docsearch-configs/blob/master/configs/gatsbyjs.json) maybe?

Now talking about the having the original to compare, I get it, and wow is tricky 😱 Maybe for the translator is not that bad since they can compare with a local version side to side in their code editor and that'd be it, but for the reviewers and maintainers is tricky since they won't have an easy way to compare with the original to check correct translations, correct line numbers against original, etc. I think maybe in 2 ways to address this maybe:

- In the PR, the first commit will always be English version, and then the next ones would be the translation, with this we could see the diffs at least, but I think it's still tricky 😓 
- We can merge the English versions first of the ones that are in the works right now, so we can have the original to compare, both the translator and the reviewers 🤔 This sound easier, but I don't know if we have a way to not index those before the translation is done, we could figure that out, as you mentioned with the bot approach @tesseralis 

The closest problem is the one related to the original versions 🤔 But I think the bot approach could work 🚀 Let me know and I can help, and thank you @horacioh for the first translation, with this the work starts to get more visible and we can iterate on top of that, and to @jonniebigodes for jumping in 💪

Darking360:
@horacioh Some work from the starter Russian translation here, more of what we said to remove the images, but I think is pretty close to what you've done here, this is the [PR](https://github.com/gatsbyjs/gatsby/pull/17840) :eyes: 

If we start like that we could start pulling the translations from your fork, and start doing the PRs here, for example, the tutorial part one that I'm finishing and so on :rocket:

horacioh:
I removed all the relevant images I saw on the translations I copied before. Hope it's everything OK and we can continue with more data to translate! ;)

horacioh:
`icon.png` image removed!

@tesseralis let me know if you need anything else! :)

tesseralis:
@horacioh @Darking360 After a lot of thought I decided that the best way to move forward is to use a separate repo for each of the translated docs, like we did with React. I made a copy of the gatsby monorepo containing only the `docs/` directory here: https://github.com/gatsbyjs/gatsby-es/tree/master

@horacioh could you re-open a pull request to that repo? You should be able to just clone it and replace the files with the ones you changed (no need to touch the images at all!) I believe that this will greatly improve the review experience, since we can have side-by-side with the original English :)

Sorry to everyone for the extra work -- I tried to figure out a way to do it in the monorepo but as it turns out, this was the easiest path all along.

horacioh:
perfect @tesseralis !!
I will re-open the PR on the other repo :)

PR #17804, opened 2019-09-22
Allow the use of alternative config files for the root config

Bouncey:
## Description
Allow the use of alternate root `gatsby-config` files.

This PR creates a new cli option `-f : --config-file` that allows a user to specify a different config file for `develop`, `build` and `serve` commands.


Bouncey:
I can't seem to get the new e2e tests picked up be circle-ci, most probably because I have done something incorrectly or not at all.

pieh:
What's the use case for this?

pieh:
You can use env var inside your `gatsby-config` file for alternative configuration without adding more things that gatsby core need to support:

```
// in gatsby-config
module.exports = {
  site: {
    siteMetadata: {
      title: process.env.GATSBY_VARIANT === 'alternate' ? 'Alternate' : 'Default'
    }
  }
}
```

and run with `GATSBY_VARIANT=alternate gatsby build`

or even to mimic your approach:
```
// in gatsby-config
if (process.env.GATSBY_VARIANT === `alternate`) {
  module.exports = require(`./alternative-config`)
} else {
  module.exports = {
    // your regular export
  }
}
```

Bouncey:
My use case is running spec e2e's on a large site that has a long build time. My alternate configs would only source node appropriate to the tests being run.

wardpeet:

The change you’ve proposed is not going to be accepted because we have a few ways to work around your issue. This feature will make Gatsby a little bit more complex which we want to avoid. 

For the use case, you described in your PR description. You can create a gatsby-config.serve.js, gatsby-config.develop.js & gatsby-config.build.js
```js
// in gatsby-config.js
module.exports = require(`./gatsby-config.${process.env.gatsby_executing_command}.js`)
```
Another way to mimic your behavior is the following code, you'll be able to run `yarn build --config-file ./gatsby-config.develop.js`
```js
// in gatsby-config
const configFileIndex = process.argv.findIndex(arg => arg === '--config-file')
let config;
if (configFileIndex) {
  config = require(process.argv[configFileIndex + 1]);
} else {
  config = {
     // default gatsby config
  }
}

module.exports = config;
```

muescha:
How about to make a small receipt with the suggested workarounds?

Issue #17385, opened 2019-09-04
[docs][guides]: Reference guide on sourcing images from external urls


lannonbr:
## Summary

Gatsby provides many features to make performant images when using gatsby-transformer-sharp and gatsby-image. That said, it may not be as clear how to do such if you want to have the same benefits from images like on S3 or from other online sources.

The reccomended way if there isn't a source plugin for the particular provider is to use gatsby-source-filesystem's [createRemoteFileNode](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-source-filesystem#createremotefilenode) API which allows you to generate new File nodes from urls. These then will be digested into gatsby-transformer-sharp.

I am thinking it would be good to have a reference guide in the "Images, Files and Video" section explaining how to do this and I am willing to write it up.

lannonbr:
Not stale, just waiting for final reviews on the PR.

PR #17806, opened 2019-09-22
Update working-with-images-in-markdown.md

nratter:
I was having lots of troubles getting this to recognize the path I used for my image in my markdown file as a File. It turns out it's because I wasn't loading my images directory in `gatsby-source-filesystem`. This should help prevent others from having issues with this.


muescha:
But this change is only with a different images folder. But if a user like to have it in same dir like the MD file then you not need this. Maybe it would better to have a new section with info if the file is not in the sourcing dirs

nratter:
Yeah I see your point, but I just think they'll know how to proceed if they want to put the image files in the same directory as the markdown file.

muescha:
i assume that users reading this doc are new to gatsby and try to find out how to include images. so many was should be shown. also your case `images in different folder`.

i would expect this different use cases as different sections:
- same folder
- different folder than in the source plugin before
- external images (when the MD files coming from a headless cms)

marcysutton:
I think a short, written note about sourcing images from the same folder would suffice. As for sourcing external images, there's already a PR in the works: https://github.com/gatsbyjs/gatsby/issues/17385

nratter:
@marcysutton I've updated the pull request.

docs/docs/working-with-images-in-markdown.md
```diff
@@ -32,16 +32,24 @@ module.exports = {
...
+      resolve: `gatsby-source-filesystem`,
+      options: {
+        path: `${__dirname}/src/images`,
+      },
+    },
   ],
 }
 ```
 
+Note: Resolving the images directory separately (as shown above) isn't necessary if the images are in the same directory as the markdown files.
```
marcysutton:
```suggestion
Note: If the images are in the same directory as the Markdown files, sourcing and resolving the images directory separately (as shown above) isn't necessary.
```

birnan:
But this change is only with a different images folder. But if a user like to have it in same dir like the MD file then you not need this. Maybe it,,

birnan:
i would expect this different use cases as different sections,,

marcysutton:
That's a fair suggestion @birnan. Are you able to add that, @nratter?

birnan:
This doc are new to gatsby and try to find out how to include images. so many was should be shown. also your..

marcysutton:
I can't quite follow your last comment, @birnan. Can you clarify?

muescha:
i think it is the same as i requested in https://github.com/gatsbyjs/gatsby/pull/17806#issuecomment-534629123 with many sections and also some more image sources?

nratter:
I disagree with the above comments. I think having multiple sections is unnecessarily verbose. The important piece of information is that both the images AND the markdown file have to be sourced. If someone has their images in the same directory as their MD then I think it's perfectly reasonable to assume they'll understand to remove the images directory from the sourcing. The added note should suffice IMO. If everyone disagrees feel free to modify as needed.

muescha:
i think for new users it's better to have different sections they can jump right to the section they need and read only this apply it.

advanced users are abled to understand what to remove and what not. and he know which section is easy to skip. 
 
for new users is it's a big huddle at first to understand whats going on and to find out what to tweak :(

LekoArts:
I think this document shouldn't unnecessarily complicate things and show the happy path and most common way of doing things -- which is in our opinion placing the images besides the markdown file. A note should link to a different doc page that explains sourcing images from somewhere else.

marcysutton:
I went ahead and finished this PR to resolve the issue. I'd love to get everyone's take on it to see if it satisfies the various use cases. Let me know if you have any feedback, otherwise we'll merge it and keep moving forward. Thank you!

PR #17810, opened 2019-09-22
docs: add accordion component with sample usages

jbachhardie:
## Description

This introduces Accordion, Details and Summary components to declutter docs that consist of lists of options or terminology that the reader isn't expected to read linearly from top to bottom. I've used them in a few pages that I thought particularly demonstrated the advantages of this pattern as kind of an example but I'm sure they could be more widely applicable.

gastby-docs-accordion

The Accordion component groups multiple detail blocks together so that only one of them can be open at any given time. It's entirely optional and without it the Details and Summary components fall back to working like the standard HTML elements they're based on.

The are [a few known bugs](https://www.scottohara.me/blog/2018/09/03/details-and-summary.html) in how screen readers treat `<summary>` but this is probably still the least-bad approach here. I considered using a `button` / `div` setup but then you get into the trouble of assigning IDs and wrapping the buttons in `h#` tags and it just results in much messier MDX.

IE11, Edge and Oracle Mini don't handle `details` / `summary` but they should fall back to displaying them open which is perfectly acceptable.

### DIFF
--- a/docs/docs/gatsby-internals-terminology.md
+++ b/docs/docs/gatsby-internals-terminology.md
@@ -6,7 +6,15 @@ Throughout the Gatsby code, you'll see the below object fields and variables men
 
 ## Page
 
+<Accordion>
+
+<Details>
+
+<Summary>
+
 ### Page Object
+
+</Summary>
 
 created by calls to [createPage](/docs/actions/#createPage) (see [Page Creation](/docs/page-creation)).
 
@@ -21,7 +29,15 @@ created by calls to [createPage](/docs/actions/#createPage) (see [Page Creation]
 
 The above fields are explained below
 
+</Details>
+
+<Details>
+
+<Summary>
+
 ### path
+
+</Summary>
 
 The publicly accessible path in the web URL to access the page in question. E.g
 
@@ -29,11 +45,27 @@ The publicly accessible path in the web URL to access the page in question. E.g
 
 It is created when the page object is created (see [Page Creation](/docs/page-creation/))
 
+</Details>
+
+<Details>
+
+<Summary>
+
 ### Redux `pages` namespace
+
+</Summary>
 
 Contains a map of Page [path](#path) -> [Page object](#page-object).
 
+</Details>
+
+<Details>
+
+<Summary>
+
 ### matchPath
+
+</Summary>
 
 Think of this instead as `client matchPath`. It is ignored when creating pages during the build. But on the frontend, when resolving the page from the path ([find-path.js]()), it is used (via [reach router](https://github.com/reach/router/blob/master/src/lib/utils.js)) to find the matching page. Note that the [pages are sorted](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/pages-writer.js#L38) so that those with matchPaths are at the end, so that explicit paths are matched first.
 
@@ -41,7 +73,15 @@ This is also used by [gatsby-plugin-create-client-paths](/packages/gatsby-plugin
 
 It is also used by [gatsby-plugin-netlify](/packages/gatsby-plugin-netlify/?=netlify) when creating `_redirects`.
 
+</Details>
+
+<Details>
+
+<Summary>
+
 ### jsonName
+
+</Summary>
 
 The logical name for the page's query json result. The name is constructed during [createPage](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/redux/actions.js#L229) using a kebabHash of page path. E.g. For the above page path, it is:
 
@@ -49,7 +89,15 @@ The logical name for the page's query json result. The name is constructed durin
 
 The actual json file is written to disk after [Query Execution](/docs/query-execution/#save-query-results-to-redux-and-disk/).
 
+</Details>
+
+<Details>
+
+<Summary>
+
 ### component
+
+</Summary>
 
 The path on disk to the JavaScript file containing the React component. E.g
 
@@ -57,7 +105,15 @@ The path on disk to the JavaScript file containing the React component. E.g
 
 Think of this as `componentPath` instead.
 
+</Details>
+
+<Details>
+
+<Summary>
+
 ### Redux `components` namespace
+
+</Summary>
 
 Mapping from `component` (path on disk) to its [Page object](#page-object). It is created every time a page is created (by listening to `CREATE_PAGE`).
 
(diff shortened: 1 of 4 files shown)


muescha:
I like your component - it's nice 

But:
-  I personally don't like if content is hidden and if I like to compare all options I need to click always on the accordion to see the full content. 
- hidden content is not searchable with browser search. 
- users coming from Google searching for specific keywords need also.click through all sections to find it

jbachhardie:
@muescha That's a great point that any use of something like this is a usability tradeoff between readers who skim and readers who search. It definitely should be used sparingly and only on non-key content. 

With that in consideration, perhaps collapsing the terminology terms is excessive but what do you think of cases like the internationalization page, where we're collapsing different options for doing the same thing? I thought this component would be useful precisely because I've found it frustrating to search for something, find it multiple times on the page and then have to scroll up each time to figure out what option it's associated with.

docs/docs/gatsby-internals-terminology.md
```diff
@@ -6,8 +6,16 @@ Throughout the Gatsby code, you'll see the below object fields and variables men
 
 ## Page
 
+<Accordion>
```
marcysutton: Obfuscating the terminology in collapsed sections on this page doesn't seem super helpful, as users can no longer scan the contents. I'd recommend not using the accordion here.

docs/docs/localization-i18n.md
```diff
@@ -12,8 +12,16 @@ This guide is a brief look at the options that exist for enhancing your Gatsby p
 
 There are a few React i18n packages out there. Several options include [react-intl](https://github.com/yahoo/react-intl), the community [Gatsby plugin](https://www.npmjs.com/package/gatsby-plugin-i18n) and [react-i18next](https://github.com/i18next/react-i18next/). There are several factors to consider when choosing a package: Do you already use a similar package in another project? How well does the package meet the needs of your users? Are you or your team already familiar with a certain package? Is the package well documented and maintained?
 
+<Accordion>
```
marcysutton: This usage works better than the internals terminology doc in my opinion, as the interface presents three package options as choices the user has to make. Ultimately, Gatsby should be more decisive about which direction we recommend for localization – but we won't have that solidified quite yet until the [Internalization project](https://github.com/gatsbyjs/rfcs/pull/42) is farther along.

docs/docs/localization-i18n.md
```diff
@@ -12,8 +12,16 @@ This guide is a brief look at the options that exist for enhancing your Gatsby p
 
 There are a few React i18n packages out there. Several options include [react-intl](https://github.com/yahoo/react-intl), the community [Gatsby plugin](https://www.npmjs.com/package/gatsby-plugin-i18n) and [react-i18next](https://github.com/i18next/react-i18next/). There are several factors to consider when choosing a package: Do you already use a similar package in another project? How well does the package meet the needs of your users? Are you or your team already familiar with a certain package? Is the package well documented and maintained?
```
marcysutton:
```suggestion

Here are three potential packages you could choose for internationalization in Gatsby:

```
Adding a lead-in sentence.

www/src/components/accordion.js
```diff
@@ -0,0 +1,150 @@
+import React, { useState, useMemo, useContext, useRef, useEffect } from "react"
+import { colors, transition, space, radii } from "../utils/presets"
+import ChevronSvg from "./sidebar/chevron-svg"
+
+const AccordionContext = React.createContext()
+
+export const Accordion = ({ children }) => {
+  const [childState, setChildState] = useState([])
+  const lastRegisteredId = useRef(0)
+  const accordionContextValue = useMemo(() => {
```
marcysutton: I haven't seen that hook before, super cool!

www/src/components/accordion.js
```diff
@@ -0,0 +1,150 @@
...
+export const Details = props => {
+  const [open, toggle] = useAccordionState()
+  const contextValue = useMemo(() => {
+    return {
+      toggle,
+    }
+  }, [toggle])
+  return (
+    <DetailsContext.Provider value={contextValue}>
+      <details open={open} {...props} />
```
marcysutton: The details and summary elements are great, but they aren't supported everywhere. For browsers that don't support it (such as Edge, until users upgrade to the Chromium-based version), this is how it renders:

details collapsed with summaries showing

Can we shim support somehow? Here's an article on it: https://www.smashingmagazine.com/2014/11/complete-polyfill-html5-details-element/

www/src/components/accordion.js
```diff
@@ -0,0 +1,150 @@
...
+          "&:before": {
+            background: colors.link.color,
+            transform: `scale(1)`,
+          },
+        },
+      }}
+      {...props}
+    >
+      {children}
+      <ChevronSvg
```
marcysutton: In Safari, the icon is dropping to the next line:

Three internationalization packages collapsed

www/src/components/accordion.js
```diff
@@ -0,0 +1,150 @@
...
+import ChevronSvg from "./sidebar/chevron-svg"
+
+const AccordionContext = React.createContext()
+
+export const Accordion = ({ children }) => {
+  const [childState, setChildState] = useState([])
+  const lastRegisteredId = useRef(0)
+  const accordionContextValue = useMemo(() => {
+    return {
+      registerChild: ref => {
```
marcysutton: Some code comments would be helpful to explain what these methods are doing.

www/src/components/accordion.js
jbachhardie: We could easily shim support, yeah. I was worried about accessibility. since any shim would have to have `role="button"` but that would hide the heading inside the `summary` element from the a11y tree because buttons can't have headings inside. 

What would you prefer we lose in legacy browsers, a bit of screen-reader usability or the ability to expand and collapse?

I'll fix the styling so the spacing is less awkward in either case. I might need some pointers on getting Gatsby building in IE11 compatibility mode the default develop server didn't load at all when I tried it.

www/src/components/accordion.js
jbachhardie: Well spotted! This is apparently a [bug in WebKit](https://bugs.webkit.org/show_bug.cgi?id=190065) I didn't know about. I've worked around it.

marcysutton:
Thank you so much for this PR and the time you spent on it! 

Because the details and summary elements are provided natively by many browsers, having a React component that effectively does the same thing doesn't seem like the best use of JavaScript code, which has to be shipped to end users. I can envision a different version of this that shims support for the native elements to browsers that need it, and if you're interested in working on that we would welcome your contribution. However, I'm going to close this PR as it doesn't quite meet the needs for the Gatsby docs. But I want to thank you again for your contributions. If you have any questions, let us know!

PR #17813, opened 2019-09-22
test(gatsby): add some tests for query running caching

pieh:
When reviewing https://github.com/gatsbyjs/gatsby/pull/17682 I found few issues there that are pretty hard to even discover without deep understanding of how query caching, query results invalidation and none of our automated tests caught that.

This is attempt to add tests for those.

I don't particularly like the way I implemented `setup` function in this test suite (re-implementing parts of bootstrap, build and develop commands), but at least it works in detecting issues I found in referenced PR:
```
 FAIL  packages/gatsby/src/query/__tests__/data-tracking.js (5.617s)
  query caching between builds
    Baseline
      ✓ first run - should run all queries (1413ms)
      ✓ rerunning without data changes and without restart shouldn't run any queries (3ms)
      ✕ rerunning without data changes after restart shouldn't run any queries (271ms)
      ✓ rerunning after cache clearing - should run all queries (235ms)
    Changed data node used for static query
      ✓ rerunning after cache clearing - should run all queries (213ms)
      ✓ reruns only static query (no restart) (7ms)
      ✕ reruns only static query (with restart) (240ms)
    Changed data node used for pages
      ✓ rerunning after cache clearing - should run all queries (242ms)
      ✓ reruns queries only for listing and detail page that uses that node (no restart) (7ms)
      ✕ reruns queries only for listing and detail page that uses that node (with restart) (261ms)
    Changed data not used by either page or static queries
      ✓ rerunning after cache clearing - should run all queries (336ms)
      ✓ changing node not used by anything doesn't trigger running any queries (no restart) (5ms)
      ✕ changing node not used by anything doesn't trigger running any queries (with restart) (288ms)
    Changing data used in multiple queries properly invalidates them
      ✓ should run all queries after clearing cache (237ms)
      ✕ should run only queries that use changed data (no restart) (8ms)
      ✕ should run only queries that use changed data (with restart) (244ms)
    Running "queries" for page without query
      ✓ rerunning after cache clearing - should run all queries (230ms)
      ✓ rerunning should not run any queries (no restart) (2ms)
      ○ skipped rerunning should not run any queries (with restart)
```

Alternative solution would be to run `gatsby/src/commands/develop` (and/or `build`) and:
 - mock out bunch of things that those tests should not care about
 - make it easier to disable some things programatically (things like building webpack, plugin loading, html building etc) - that would honestly be preferred as it would be much easier to work on integration style tests for other things (and possibly make integration tests for source/transformer plugin much easier)

---Edit:
Let me know what you think about this before I dive into making it work with loki

### DIFF
--- a/packages/gatsby/src/db/loki/index.js
+++ b/packages/gatsby/src/db/loki/index.js
@@ -61,10 +61,11 @@ function ensureNodeCollections(db) {
   })
 }
 
-function startFileDb(saveFile) {
+function startFileDb({ saveFile, adapter }) {
   return new Promise((resolve, reject) => {
     const dbOptions = {
       autoload: true,
+      adapter,
       autoloadCallback: err => {
         if (err) {
           reject(err)
@@ -94,14 +95,14 @@ async function startInMemory() {
  * the existing state has been loaded (if there was an existing
  * saveFile)
  */
-async function start({ saveFile } = {}) {
+async function start({ saveFile, adapter } = {}) {
   if (saveFile && !_.isString(saveFile)) {
     throw new Error(`saveFile must be a path`)
   }
   if (saveFile) {
     const saveDir = path.dirname(saveFile)
     await fs.ensureDir(saveDir)
-    await startFileDb(saveFile)
+    await startFileDb({ saveFile, adapter })
   } else {
     await startInMemory()
   }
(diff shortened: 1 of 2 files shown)


pieh:
Missing tests here are for invalidating query results when changing context (which is also known issue) after restarting gatsby (it works when it happens during `gatsby develop`), but the way it's currently implemented make it pretty difficult to test (queueing rerunning with setTimeout - https://github.com/gatsbyjs/gatsby/blob/990b352ab11be004a8a6132cff6e355910db1018/packages/gatsby/src/redux/machines/page-component.js#L84-L88 ), so I'll skip that scenario and just merge initial set of tests

PR #17856, opened 2019-09-25
fix(www): fix plugin searchbar query

### LINKED ISSUE #17854: Query parameters incorrectly handled by https://www.gatsbyjs.org/plugins/
## Description
The plugins page puts all query parameters (except for the first character) directly into the search box.

### Steps to reproduce
On the production page, visit the URL: `https://www.gatsbyjs.org/plugins/?hello=world&ga=true`

The plugins page puts all query parameters (except for the first character) directly into the search box, such that this link will open to a plugins page searching for `ello=world&ga=true`

### Expected result

The search box should ignore unrelated query parameters, particularly the the analytics linker id picked up from gatsbyjs.com. Ideally a specific parameter would be used (ie. `?search=gatsby-plugin-`)

### Actual result

The search box is filled with the raw query parameter text.

### Environment

N/A - This is a bug in the production www page.

universse:
## Related Issues
Fixes #17854

### DIFF
--- a/www/src/components/plugin-searchbar-body.js
+++ b/www/src/components/plugin-searchbar-body.js
@@ -467,17 +467,16 @@ class PluginSearchBar extends Component {
 
   urlToSearch = () => {
     if (this.props.location.search) {
-      // ignore this automatically added query parameter
-      const search = this.props.location.search
-        .replace(`no-cache=1`, ``)
-        .slice(2)
+      const params = new URLSearchParams(this.props.location.search)
+      // only take into account 'search' query
+      const search = params.get(`search`)
       return decodeURIComponent(search)
     }
     return ``
   }
 
   updateHistory(value) {
-    reachNavigate(`${this.props.location.pathname}?=${value.query}`, {
+    reachNavigate(`${this.props.location.pathname}?=search=${value.query}`, {
       replace: true,
     })
   }
--- a/www/src/components/plugin-searchbar-body.js
+++ b/www/src/components/plugin-searchbar-body.js
@@ -467,18 +467,20 @@ class PluginSearchBar extends Component {
 
   urlToSearch = () => {
     if (this.props.location.search) {
-      const params = new URLSearchParams(this.props.location.search)
-      // only take into account 'search' query
-      const search = params.get(`search`)
-      return decodeURIComponent(search)
+      const match = /(\?|&)=([^&]+)/.exec(this.props.location.search)
+      if (match) return decodeURIComponent(match[2])
+      return ``
     }
     return ``
   }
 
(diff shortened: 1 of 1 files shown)


www/src/components/plugin-searchbar-body.js
```diff
@@ -467,17 +467,16 @@ class PluginSearchBar extends Component {
 
   urlToSearch = () => {
     if (this.props.location.search) {
-      // ignore this automatically added query parameter
-      const search = this.props.location.search
-        .replace(`no-cache=1`, ``)
-        .slice(2)
+      const params = new URLSearchParams(this.props.location.search)
```
pieh: Not sure if it's relevant to us, but `URLSearchParams` is not supported by IE11 ( https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams#Browser_compatibility )

www/src/components/plugin-searchbar-body.js
```diff
@@ -467,17 +467,16 @@ class PluginSearchBar extends Component {
...
+      // only take into account 'search' query
+      const search = params.get(`search`)
```
pieh: Would be great if we don't change param name (or maybe just support previous one) - so something like
```
const search = params.get(`search`) || params.get(``)
```
(dependning if we end up using `URLSearchParams` or not, see comment above)

universse: Guess we can use `qs` instead of `URLSearchParams` since it's also being used in other places. Supporting the old behavior might be tricky though since `qs.parse('?=gatsby')` would result in `{}`. Anw I can give it a try.

pieh: Not ideal, but we could maybe check if string starts with `?=` and replace it with something that `qs` supports

sever1an: Would it work to use a simple regular expression to extract the text we want? Since the text is prefixed with `=` I doubt that it would cause problems for other query string users. The content of props.location.serach is still url encoded so it is safe to use `&` for tokenization. 

```javascript
const extractSearch = queryString => {
  let match = /(\?|&)=([^&]+)/.exec(queryString);
  if (match) return match[2];
  return null;
};
```

www/src/components/plugin-searchbar-body.js
muescha: there is a polyfill https://github.com/ungap/url-search-params

www/src/components/plugin-searchbar-body.js
universse: Yep. That should work. Thank you.

PR #17858, opened 2019-09-25
chore(docs): Update caching

rwieruch:
Stumbled across this when I updated my nginx configuration.

Isn't it just the static folder? In my production build, I can only find this director.

### DIFF
--- a/docs/docs/caching.md
+++ b/docs/docs/caching.md
@@ -14,13 +14,13 @@ The `cache-control` header should be `cache-control: public, max-age=0, must-rev
 
 ## Page data
 
-Similar to HTML files, the JSON files in the `page-data` directory should never be cached by the browser. In fact, it's possible for these files to be updated even without doing a redeploy of your site. Because of this, browsers should be instructed to check on every request if the data in your application has changed.
+Similar to HTML files, the JSON files in the `page-data/` directory should never be cached by the browser. In fact, it's possible for these files to be updated even without doing a redeploy of your site. Because of this, browsers should be instructed to check on every request if the data in your application has changed.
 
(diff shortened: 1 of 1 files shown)


LekoArts:
Well, when you run `gatsby build` it outputs to `public` which in turn will have a `static` folder. So I guess technically speaking `public/static` is correct. People might also confuse it with the `<rootDir>/static` folder whose contents are copied over on the build to `public`.

rwieruch:
Okay. Then it should be `public/page-data` for the previous section to keep it consistent?

sidharthachatterjee:
> Then it should be public/page-data for the previous section to keep it consistent?

Yup, that sounds good to me!

docs/docs/caching.md
```diff
@@ -14,13 +14,13 @@ The `cache-control` header should be `cache-control: public, max-age=0, must-rev
 
 ## Page data
 
-Similar to HTML files, the JSON files in the `page-data` directory should never be cached by the browser. In fact, it's possible for these files to be updated even without doing a redeploy of your site. Because of this, browsers should be instructed to check on every request if the data in your application has changed.
+Similar to HTML files, the JSON files in the `page-data/` directory should never be cached by the browser. In fact, it's possible for these files to be updated even without doing a redeploy of your site. Because of this, browsers should be instructed to check on every request if the data in your application has changed.
```
sidharthachatterjee:
```suggestion
Similar to HTML files, the JSON files in the `public/page-data/` directory should never be cached by the browser. In fact, it's possible for these files to be updated even without doing a redeploy of your site. Because of this, browsers should be instructed to check on every request if the data in your application has changed.
```

KyleAMathews:
We added public to page-data and removed it from static? Why aren't we consistent?

PR #17871, opened 2019-09-25
fix(gatsby-theme-blog-core): Normalize trailing slashes

JRGould:
## Description
`gatsby-theme-blog-core` was producing inconsistent urls based on wether a slug was supplied via frontmatter or inferred from the filepath. When a slug was provided via frontmatter, a trailing slash would need to be explicitly specified (eg: `slug: my-cool-post/`) but slugs inferred from the filepath always contained a trailing slash.

This PR normalizes urls by adding a `useTrailingSlash` option and ensuring that all blog post urls either do or do not have a trailing slash based on wether that option is set to `true` or `false` (default).

 - `gatsby-node.js` has been modified to normalize the trailing slash and utilize the new option
 - `default-options.js` has been modified to provide a default for `useTrailingSlash`
 - `readme.md` has been updated 

## Related Issues

I didn't see any open issues related to this and figured that it would be easy enough just to create a PR since this is a relatively small change.


pieh:
I think just normalizing it to always contain trailing slash (as this is what we for example for pages created from `src/pages` components) and not adding config options would be cleaner approach (avoid adding options and alternative code paths if possible). Also I think this option wouldn't affect pages not created from mdx and that be pretty confusing

We also have have plugins that either remove trailing slashes ( https://www.gatsbyjs.org/packages/gatsby-plugin-remove-trailing-slashes/ ) or force them ( https://www.gatsbyjs.org/packages/gatsby-plugin-force-trailing-slashes/ ) that target all pages which seems like better approach than adding this option here.

JRGould:
Thanks @pieh, that makes sense. Replaced the `feat` commit with a `fix` commit that focuses on the actual bug.

PR #17880, opened 2019-09-25
fix(gatsby-remark-images-contentful): avoid modifying options when passed by reference

jdcataldo:
## Description
When using `gatsby-remark-images-contentful`, images of different sizes would end up having the wrong sizes for breakpoints defined and were based on the first image that finished processing. This is because `options` is being passed in by reference and the `sizes` key was being set on it causing `options.sizes` to now be defined for all images. 

This change makes it so that nothing is ever set on `options` within the `buildResponsiveSizes` function and all option values are destructured from the passed in `options` object. This ensure that each image will have the proper sizes set based on the images metadata or the size option that is set.

I also fixed a typo for `cahedRawHTML` to `cachedRawHTML`.

### DIFF
--- a/packages/gatsby-remark-images-contentful/src/index.js
+++ b/packages/gatsby-remark-images-contentful/src/index.js
@@ -61,10 +61,10 @@ module.exports = async (
     const optionsHash = createContentDigest(options)
 
     const cacheKey = `remark-images-ctf-${fileName}-${optionsHash}`
-    let cahedRawHTML = await cache.get(cacheKey)
+    let cachedRawHTML = await cache.get(cacheKey)
 
-    if (cahedRawHTML) {
-      return cahedRawHTML
+    if (cachedRawHTML) {
+      return cachedRawHTML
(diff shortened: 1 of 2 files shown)


packages/gatsby-remark-images-contentful/src/index.js
```diff
@@ -61,10 +61,10 @@ module.exports = async (
     const optionsHash = createContentDigest(options)
 
     const cacheKey = `remark-images-ctf-${fileName}-${optionsHash}`
-    let cahedRawHTML = await cache.get(cacheKey)
+    let cachedRawHTML = await cache.get(cacheKey)
 
-    if (cahedRawHTML) {
-      return cahedRawHTML
+    if (cachedRawHTML) {
+      return cachedRawHTML
```
pieh: 🙈 nice catch ;)

pieh:
Published `gatsby-remark-images-contentful@2.1.15`

PR #17883, opened 2019-09-25
chore(showcase): add GatsbyFinds

bvlktech:
## Description
Hi i would love to add GatsbyFinds as a showcase to show my love and support for GatsbyJS

### DIFF
--- a/docs/sites.yml
+++ b/docs/sites.yml
@@ -7341,3 +7341,15 @@
   built_by: The Delta Studio
   built_by_url: https://www.thedelta.io
   featured: false
+- title: GatsbyFinds
+  main_url: https://gatsbyfinds.netlify.com
+  url: https://gatsbyfinds.netlify.com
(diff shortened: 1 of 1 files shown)


muescha:
I think there is a little problem with the logo usage?
see https://www.gatsbyjs.org/guidelines/logo/

bvlktech:
@muescha sorry didn't get notified of this but seeing as im not claiming ownership of it nor will i ever nor do i intend to because i don't as well as there are several several links back to several Gatsby owned projects libraries ext.. its not really an issue and in the whole project its based around well GatsbyJS what else am i going to use? furthermore this logo is the default logo that comes with the default Gatsby starter config

marcysutton:
Is this your project @bvlktech? Because I agree that it should follow Gatsby’s brand guidelines. They ask not to “integrate the Gatsby logo or monogram into your own logo”. If you do own the site, I would recommend changing it to follow the brand guide.

PR #17938, opened 2019-09-27
fix: update fsevents via chokidar to fix MacOS bug

### LINKED ISSUE #17131: Stuck at "source and transform nodes" / "createPagesStatefully" on MacOS
## Description

I am working on a theme, local build was working fine with no problems, and recently upgraded all the dependencies to Gatsby 2.14.0 and both `gatsby develop` and `gatsby build` hang at `source and transform nodes` in my local dev environment.

Interestingly it works and builds on Netlify.  This would point to it being something on my system. I have deleted the node modules folders in the workspaces and the root workspace folder and done a fresh yarn command.  I also deleted the yarn.lock and package.lock files...not sure if this would cause problems.

### Steps to reproduce

Theme repo is here: [Gatsby-Theme-Catalyst-Core](https://github.com/ehowey/gatsby-theme-catalyst-core)

Starter repo is here: [Gatsby-Starter-Catalyst-Core](https://github.com/ehowey/gatsby-starter-catalyst-core)

I have this setup in a yarn workspace for development however the same issue occurs if you do a fresh install of the starter using `gatsby new my-catalyst-starter-core https://github.com/ehowey/gatsby-starter-catalyst-core`

### Expected result

Build successfully

### Actual result

yarn workspace v1.17.3
yarn run v1.17.3
$ gatsby develop
success open and validate gatsby-configs - 0.122 s
success load plugins - 1.964 s
success onPreInit - 0.073 s
success initialize cache - 0.056 s
success copy gatsby files - 0.242 s
success onPreBootstrap - 0.087 s
⠙ source and transform nodes

### Environment

  System:
    OS: macOS High Sierra 10.13.6
    CPU: (2) x64 Intel(R) Core(TM)2 Duo CPU     P8600  @ 2.40GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 12.9.1 - /usr/local/bin/node
    npm: 6.11.2 - /usr/local/bin/npm
  Languages:
    Python: 2.7.16 - /usr/local/bin/python
  Browsers:
    Chrome: 76.0.3809.100
    Firefox: 67.0.2
    Safari: 12.1.2
  npmGlobalPackages:
    gatsby-cli: 2.7.40

Js-Brecht:
## Description

This is a rather quirky issue, where `gatsby develop` (and likely `gatsby build`) will hang at various stages while using MacOS.  Most common was `source and transform`, and usually when running plugin `gatsby-source-filesystem`.  It also hangs in `dev-404-page`, `internal-data-bridge`, and others.

As it turns out, there is an issue with `chokidar.watch()` on MacOS.  It uses `fsevents` by default, in an effort to increase speed and reduce cpu load.

Disabling `fsevents` has been tested by @ehowey, who was experiencing this issue regularly.  It has been shown to result in a successful build.  It will also increase the load on the cpu, so for larger sites, there will be a performance impact.  Use env var `CHOKIDAR_USEPOLLING=1` to disable `fsevents`.

**UPDATE 10/10/19**
The issue has been tracked down to a bug in `fsevents` itself.  Seems there was a race condition that was causing the thread to hang.  Thanks to @RomanHotsiy for his hard work! 🎉 🎉  As soon as the fix has trickled down, we can do updates to fix the problem.

[See comment here](https://github.com/gatsbyjs/gatsby/pull/17938#issuecomment-540661991) for instructions for testing

**UPDATE 10/13/19**
`fsevents@2.1.1` has been published, which resolves the issue.  Just requires updating lockfiles.

## Related Issues

Fixes #17131

### DIFF
--- a/packages/gatsby-dev-cli/src/watch.js
+++ b/packages/gatsby-dev-cli/src/watch.js
@@ -170,7 +170,7 @@ async function watch(
     .watch(watchers, {
       ignored: [filePath => _.some(ignored, reg => reg.test(filePath))],
       // Setting useFsEvents to false fixes https://github.com/gatsbyjs/gatsby/issues/17131
-      useFsEvents: process.env.GATSBY_USE_FSEVENTS || true,
+      useFsEvents: process.env.GATSBY_USE_FSEVENTS !== `0`,
     })
     .on(`all`, async (event, filePath) => {
       if (!watchEvents.includes(event)) {
--- a/packages/gatsby-dev-cli/src/watch.js
+++ b/packages/gatsby-dev-cli/src/watch.js
@@ -169,6 +169,8 @@ async function watch(
   chokidar
     .watch(watchers, {
       ignored: [filePath => _.some(ignored, reg => reg.test(filePath))],
+      // Setting useFsEvents to false fixes https://github.com/gatsbyjs/gatsby/issues/17131
+      useFsEvents: process.env.GATSBY_USE_FSEVENTS || true,
     })
     .on(`all`, async (event, filePath) => {
       if (!watchEvents.includes(event)) {
--- a/packages/gatsby-dev-cli/src/watch.js
+++ b/packages/gatsby-dev-cli/src/watch.js
@@ -169,6 +169,7 @@ async function watch(
   chokidar
     .watch(watchers, {
       ignored: [filePath => _.some(ignored, reg => reg.test(filePath))],
+      // Setting useFsEvents to false fixes https://github.com/gatsbyjs/gatsby/issues/17131
       useFsEvents: false,
     })
     .on(`all`, async (event, filePath) => {
--- a/packages/gatsby-dev-cli/src/watch.js
+++ b/packages/gatsby-dev-cli/src/watch.js
@@ -169,8 +169,6 @@ async function watch(
   chokidar
     .watch(watchers, {
       ignored: [filePath => _.some(ignored, reg => reg.test(filePath))],
-      // Setting useFsEvents to false fixes https://github.com/gatsbyjs/gatsby/issues/17131
-      useFsEvents: process.env.GATSBY_USE_FSEVENTS !== `0`,
     })
     .on(`all`, async (event, filePath) => {
       if (!watchEvents.includes(event)) {
--- a/packages/gatsby-dev-cli/src/watch.js
+++ b/packages/gatsby-dev-cli/src/watch.js
@@ -169,8 +169,6 @@ async function watch(
   chokidar
     .watch(watchers, {
       ignored: [filePath => _.some(ignored, reg => reg.test(filePath))],
-      // Setting useFsEvents to false fixes https://github.com/gatsbyjs/gatsby/issues/17131
-      useFsEvents: false,
     })
     .on(`all`, async (event, filePath) => {
       if (!watchEvents.includes(event)) {
--- a/packages/gatsby-dev-cli/src/watch.js
+++ b/packages/gatsby-dev-cli/src/watch.js
@@ -169,6 +169,7 @@ async function watch(
   chokidar
     .watch(watchers, {
       ignored: [filePath => _.some(ignored, reg => reg.test(filePath))],
+      useFsEvents: false,
     })
     .on(`all`, async (event, filePath) => {
       if (!watchEvents.includes(event)) {
--- a/packages/gatsby-page-utils/src/watch-directory.js
+++ b/packages/gatsby-page-utils/src/watch-directory.js
@@ -5,6 +5,7 @@ const slash = require(`slash`)
 module.exports = async (path, glob, onNewFile, onRemovedFile) =>
   new Promise((resolve, reject) => {
     chokidar
+      // Setting useFsEvents to false fixes https://github.com/gatsbyjs/gatsby/issues/17131
       .watch(glob, { cwd: path, useFsEvents: false })
       .on(`add`, path => {
         path = slash(path)
--- a/packages/gatsby-page-utils/src/watch-directory.js
+++ b/packages/gatsby-page-utils/src/watch-directory.js
@@ -8,7 +8,7 @@ module.exports = async (path, glob, onNewFile, onRemovedFile) =>
       .watch(glob, {
         cwd: path,
         // Setting useFsEvents to false fixes https://github.com/gatsbyjs/gatsby/issues/17131
-        useFsEvents: process.env.GATSBY_USE_FSEVENTS || true,
+        useFsEvents: process.env.GATSBY_USE_FSEVENTS !== `0`,
       })
       .on(`add`, path => {
         path = slash(path)
--- a/packages/gatsby-page-utils/src/watch-directory.js
+++ b/packages/gatsby-page-utils/src/watch-directory.js
@@ -5,8 +5,7 @@ const slash = require(`slash`)
 module.exports = async (path, glob, onNewFile, onRemovedFile) =>
   new Promise((resolve, reject) => {
     chokidar
-      // Setting useFsEvents to false fixes https://github.com/gatsbyjs/gatsby/issues/17131
-      .watch(glob, { cwd: path, useFsEvents: false })
+      .watch(glob, { cwd: path })
       .on(`add`, path => {
         path = slash(path)
         onNewFile(path)
--- a/packages/gatsby-page-utils/src/watch-directory.js
+++ b/packages/gatsby-page-utils/src/watch-directory.js
@@ -5,7 +5,7 @@ const slash = require(`slash`)
 module.exports = async (path, glob, onNewFile, onRemovedFile) =>
   new Promise((resolve, reject) => {
     chokidar
-      .watch(glob, { cwd: path })
+      .watch(glob, { cwd: path, useFsEvents: false })
       .on(`add`, path => {
         path = slash(path)
         onNewFile(path)
--- a/packages/gatsby-page-utils/src/watch-directory.js
+++ b/packages/gatsby-page-utils/src/watch-directory.js
@@ -5,7 +5,11 @@ const slash = require(`slash`)
 module.exports = async (path, glob, onNewFile, onRemovedFile) =>
   new Promise((resolve, reject) => {
     chokidar
-      .watch(glob, { cwd: path })
+      .watch(glob, {
+        cwd: path,
+        // Setting useFsEvents to false fixes https://github.com/gatsbyjs/gatsby/issues/17131
+        useFsEvents: process.env.GATSBY_USE_FSEVENTS || true,
+      })
       .on(`add`, path => {
         path = slash(path)
         onNewFile(path)
--- a/packages/gatsby-page-utils/src/watch-directory.js
+++ b/packages/gatsby-page-utils/src/watch-directory.js
@@ -5,11 +5,7 @@ const slash = require(`slash`)
 module.exports = async (path, glob, onNewFile, onRemovedFile) =>
   new Promise((resolve, reject) => {
     chokidar
-      .watch(glob, {
-        cwd: path,
-        // Setting useFsEvents to false fixes https://github.com/gatsbyjs/gatsby/issues/17131
-        useFsEvents: process.env.GATSBY_USE_FSEVENTS !== `0`,
-      })
+      .watch(glob, { cwd: path })
       .on(`add`, path => {
         path = slash(path)
         onNewFile(path)
--- a/packages/gatsby-source-filesystem/src/gatsby-node.js
+++ b/packages/gatsby-source-filesystem/src/gatsby-node.js
@@ -173,7 +173,7 @@ See docs here - https://www.gatsbyjs.org/packages/gatsby-source-filesystem/
       ...(pluginOptions.ignore || []),
     ],
     // Setting useFsEvents to false fixes https://github.com/gatsbyjs/gatsby/issues/17131
-    useFsEvents: process.env.GATSBY_USE_FSEVENTS || true,
+    useFsEvents: process.env.GATSBY_USE_FSEVENTS !== `0`,
   })
 
   watcher.on(`add`, path => {
--- a/packages/gatsby-source-filesystem/src/gatsby-node.js
+++ b/packages/gatsby-source-filesystem/src/gatsby-node.js
@@ -172,8 +172,6 @@ See docs here - https://www.gatsbyjs.org/packages/gatsby-source-filesystem/
       `../**/dist/**`,
       ...(pluginOptions.ignore || []),
     ],
-    // Setting useFsEvents to false fixes https://github.com/gatsbyjs/gatsby/issues/17131
-    useFsEvents: process.env.GATSBY_USE_FSEVENTS !== `0`,
   })
 
   watcher.on(`add`, path => {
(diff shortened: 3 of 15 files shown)


packages/gatsby/src/utils/get-static-dir.js
```diff
@@ -44,7 +44,7 @@ exports.copyStaticDirs = () => {
 exports.syncStaticDir = () => {
   const staticDir = nodePath.join(process.cwd(), `static`)
   chokidar
-    .watch(staticDir)
+    .watch(staticDir, { useFsEvents: false })
```
wardpeet:
```suggestion
    // Setting useFsEvents to false fixes https://github.com/gatsbyjs/gatsby/issues/17131
    .watch(staticDir, { useFsEvents: false })
```

packages/gatsby/src/query/query-watcher.js
```diff
@@ -232,10 +232,12 @@ const watch = async rootDir => {
...
 
   watcher = chokidar
-    .watch([
-      slash(path.join(rootDir, `/src/**/*.{js,jsx,ts,tsx}`)),
-      ...packagePaths,
-    ])
+    .watch(
+      [slash(path.join(rootDir, `/src/**/*.{js,jsx,ts,tsx}`)), ...packagePaths],
+      {
+        useFsEvents: false,
```
wardpeet:
```suggestion
        // Setting useFsEvents to false fixes https://github.com/gatsbyjs/gatsby/issues/17131
        useFsEvents: false,
```

packages/gatsby/src/internal-plugins/internal-data-bridge/gatsby-node.js
```diff
@@ -113,20 +113,22 @@ exports.sourceNodes = ({ createContentDigest, actions, store }) => {
...
-      // Delete require cache so we can reload the module.
-      delete require.cache[require.resolve(pathToGatsbyConfig)]
-      const config = require(pathToGatsbyConfig)
-      createGatsbyConfigNode(config)
-    } catch (e) {
-      // Restore the old cache since requiring the new gatsby-config.js failed.
-      if (oldCache !== undefined) {
-        require.cache[require.resolve(pathToGatsbyConfig)] = oldCache
+  chokidar
+    .watch(pathToGatsbyConfig, { useFsEvents: false })
```
wardpeet:
```suggestion
    // Setting useFsEvents to false fixes https://github.com/gatsbyjs/gatsby/issues/17131
    .watch(pathToGatsbyConfig, { useFsEvents: false })
```

packages/gatsby/src/internal-plugins/dev-404-page/gatsby-node.js
```diff
@@ -19,7 +19,7 @@ exports.createPagesStatefully = async ({ store, actions }, options, done) => {
       path: `/dev-404-page/`,
     })
     chokidar
-      .watch(source)
+      .watch(source, { useFsEvents: false })
```
wardpeet:
```suggestion
      // Setting useFsEvents to false fixes https://github.com/gatsbyjs/gatsby/issues/17131
      .watch(source, { useFsEvents: false })
```

packages/gatsby/src/commands/develop.js
```diff
@@ -298,7 +298,7 @@ async function startServer(program, { activity }) {
     slash(directoryPath(path))
   )
 
-  chokidar.watch(watchGlobs).on(`change`, async () => {
+  chokidar.watch(watchGlobs, { useFsEvents: false }).on(`change`, async () => {
```
wardpeet:
```suggestion
  // Setting useFsEvents to false fixes https://github.com/gatsbyjs/gatsby/issues/17131
  chokidar.watch(watchGlobs, { useFsEvents: false }).on(`change`, async () => {
```

packages/gatsby-source-filesystem/src/gatsby-node.js
```diff
@@ -160,6 +160,7 @@ See docs here - https://www.gatsbyjs.org/packages/gatsby-source-filesystem/
   })
 
   const watcher = chokidar.watch(pluginOptions.path, {
+    useFsEvents: false,
```
wardpeet:
```suggestion
    // Setting useFsEvents to false fixes https://github.com/gatsbyjs/gatsby/issues/17131
    useFsEvents: false,
```

packages/gatsby-page-utils/src/watch-directory.js
```diff
@@ -5,7 +5,7 @@ const slash = require(`slash`)
 module.exports = async (path, glob, onNewFile, onRemovedFile) =>
   new Promise((resolve, reject) => {
     chokidar
-      .watch(glob, { cwd: path })
+      .watch(glob, { cwd: path, useFsEvents: false })
```
wardpeet:
```suggestion
      // Setting useFsEvents to false fixes https://github.com/gatsbyjs/gatsby/issues/17131
      .watch(glob, { cwd: path, useFsEvents: false })
```

packages/gatsby-dev-cli/src/watch.js
```diff
@@ -169,6 +169,7 @@ async function watch(
   chokidar
     .watch(watchers, {
       ignored: [filePath => _.some(ignored, reg => reg.test(filePath))],
+      useFsEvents: false,
```
wardpeet:
```suggestion
      // Setting useFsEvents to false fixes https://github.com/gatsbyjs/gatsby/issues/17131
      useFsEvents: false,
```

KyleAMathews:
How does this PR change performance on a larger site?

Js-Brecht:
@KyleAMathews I'm very interested in the answer to that question as well.  **Theoretically**, it shouldn't hurt any.  Basically, what this does is fallback to polling `kqueue`, and because of the nature of `libuv`, polling takes a back seat if there are any active setTimeout(), nextTick(), Promises, handlers/setImmediate(); so, it shouldn't be blocking processor time for any running processes.

I think it would have to do with how `chokidar` implements the polling mechanism.  TBH, I don't know a lot about `chokidar`; I've never looked at their code.  I'd very much like to see this tested on a variety of different sites, small & large, especially those that extend the build process beyond the standard.

Js-Brecht:
@wardpeet Done.  Sorry, didn't even think about that.  I was just focused on getting it out, so I could go to sleep 😄

ehowey:
@Js-Brecht Wow - you worked fast on that!  I tested it on my end - it seems to completely fix this error for me.  I tested it with both `gatsby-starter-hello-world` and `gatsby-advanced-starter`.  In both cases I ran `gatsby-dev --scan-once` followed by the `gatsby-develop` command 10+ times in each repo.  I can report that it did not hang or stop in the build process ever.

Js-Brecht:
@ehowey Do you have any production sites, or sites you're developing that might be more complex that you could test it on?

Js-Brecht:
Need to stress test it

ehowey:
To be honest, not really.  My "production" sites are pretty bare bones when it comes to the size and complexity of the site.  Not in a way that provide any kind of system stress anyways.  Maybe I could pull down a larger open source site and do something with that?  Got any ideas that might fit the bill?

paulmillr:
Chokidar maintainer here. The performance would definitely be shitty after this change. Be careful.

fsevents uses one handler for 100,000 files. When it's disabled, chokidar would use 100,000 handlers. Which will use tons of RAM and CPU.

Js-Brecht:
@paulmillr good to know, thanks.  Perhaps you could weigh in on the cause of the issue, and if there's a potential fix?  TBH, I think it would be a better solution for the root of the issue to be addressed than to actually use a band-aid fix like this one.

Js-Brecht:
Nevermind, I see you have a fix published already.  Excellent!

paulmillr:
Yeah, try chokidar@master —  if it doesn't work, please provide more detatils. I've ran your loop and it didn't end, everything was fine. Perhaps it's related to "put disks to sleep" setting on MacOS or something like that.

Js-Brecht:
@ehowey care to run some more tests, but with another branch?  Instead of branch `macos-fsevents-fix`, try branch `chokidar-test`.  I've updated the chokidar dependencies to point to chokidar's github master branch.

muescha:
it is published in release [chokidar 3.2.0](https://github.com/paulmillr/chokidar/releases/tag/3.2.0)

Js-Brecht:
Updated chokidar to version 3.2.1.  Needs testing... I'd like to know that this fixes the issue, or if more investigation needs to be done.

ehowey:
I'm seeing it hang again with chokidar 3.2.1 - sorry @Js-Brecht, I don't know if it is worth too much more work on your part.  I really appreciate all of the help and effort you have taken for a random stranger on the hyperwebs! The resizing terminal window fix works reliably for me...and I dont think this bug is affecting too many people as the issues would be flooded.

Jeddf:
I've been seeing this same hanging at `createPagesStatefully` or `source and transform nodes` on my mac when running develop.

Got an improvement like @ehowey when doing runs just now, without this chokidar update 5/10 develop runs hang, with the update I only saw 2/10 hang 👍

Js-Brecht:
Because it looks like the fsevents workaround might be a bad idea, and since I have a general idea of what is causing the issue, I think what I'm going to do is start a new repository, and try to reproduce it as simply as I can.  That way we have a repro case that is easier to work with; something that @paulmillr can hoppefully use to debug.

This would be much easier if I had a Mac 😆.  I will probably be recruiting some of you that are experiencing this to do some testing.

RomanHotsiy:
> I will probably be recruiting some of you that are experiencing this to do some testing.

I will be happy to help here. I'm experiencing this issue sometimes too. So let me know if you need any help.

Js-Brecht:
Hey guys, sorry it's taking me so long.  I've been working overtime on a critical business process @ work.

I've updated the PR so that you can **_opt out_** of using fsevents, since this still seems to be an issue with the updated version of chokidar.  This way, it won't effect everyone, but if anybody is experiencing this issue and they aren't concerned about performance (small site?), they can set `GATSBY_USE_FSEVENTS=0`, and it will disable it.

I have a little time today, so I'm going to begin work on the reproduction repository, so we can start moving towards an actual resolution.

paulmillr:
There is no need for the env var. use chokidar_use_fsevents.

Jeddf:
Sounds like there is high confidence already but I did some runs on my mac to sanity check the disable fsevents workaround flag. No obvious problems with functionality either:

15/15 runs not hanging with `GATSBY_USE_FSEVENTS=0 gatsby develop`

Js-Brecht:
Even better, @paulmillr; glad you're here, because I completely missed that in the docs.  Definitely no reason for our own env var, if there is one that chokidar itself uses to do the same thing.  Will save in future maintenance, for sure; less to remember to do if another instance of `chokidar.watch()` needs to be used anywhere.

I'll update this PR later today, and just leave it as an upgrade.  It works out, because that's likely what it will be anyway, assuming I can narrow down the issue.

muescha:
maybe it would be nice have some docs about this and what this env var solves?

ZHANGTYN:
Show to result in a successful build. It will also increase the load on the cpu, so for larger sites, there will be a performance impact.

Js-Brecht:
@muescha I will be removing it.  I didn't realize there was an env var for chokidar already existing, so this would just be redundant.  Instead of `GATSBY_USE_FSEVENTS=0`, people will be able to use `CHOKIDAR_USEPOLLING=1`.

muescha:
i mean documenting this `CHOKIDAR_USEPOLLING` and to show what problem it should solves in combination with gatsby. if you have this problem with `fsevents` then maybe other users have the same problem.

maybe in some new troubleshooting doc?

wardpeet:
@Js-Brecht We could just set the polling to true at the start of gatsby-cli. If we make it opt-in, we probably want to put it in https://www.gatsbyjs.org/docs/debugging/ or something.

Js-Brecht:
Sure, I can write up a doc for it.

I'm hesitant now to make it the default setting, @wardpeet, considering how it could affect some sites.  It just feels to me like the most effective/efficient method should be the default, and while there are more than just a couple people experiencing this, it's not really enough to warrant bombing everybody.

v3.2.1 seemed to help, some; reduced the frequency, anyway.  I think I might be making some headway on a reproduction, too.

RomanHotsiy:
Looks like this is caused by `fsevents` issue. I already opened a PR to fix it: fsevents/fsevents#290

For me it is fixing the issue completely.

Would be great if someone else can try it also.

Here is how to test it:
- clone the forked fsevents repo: `git clone --single-branch -b fix/pthread_cond_wait-race-condition git@github.com:Redocly/fsevents.git`
- run `npm run build` to produce `fsevents.node`
- replace `node_modules/fsevents/fsevents.node` in your gatsby site dependencies with the resulting `fsevents.node` from the previous step
- run `gatsby develop` or `npm start` in your gatsby site

cc @Js-Brecht @paulmillr

ehowey:
@RomanHotsiy I got the following error trying access your repo:
```
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```
Do you need to update the permissions?

Jeddf:
Looks to have completely solved it for me too 🎉 Awesome!

It didn't work at first but then I realised I had a few nested node_modules/fsevents to copy my fsevents.node file to. This find command that google turned up was useful for listing them all: `find . -path '*/fsevents.node'`

ehowey:
I tested it on the hello-world starter and also on a more complex project I am working on using themes and workspaces.  It seems to fix it entirely!  You do have to make sure you track down all instances of `fsevents.node` as I had a few of them tucked into the node_modules folder in a couple of spots as commented on above.

@Js-Brecht and @RomanHotsiy it was awesome collaborating with you both to track this bug down and I learned some stuff along the way.  Kudos to your persistance in figuring this one out.  Way beyond what I ever would have gotten to on my own.

Js-Brecht:
Can you guys drop a list of the packages that are using `fsevents`?  I just want to look at their `fsevents` dependency version specifier.

FYI, you can use [yarn resolutions](https://yarnpkg.com/lang/en/docs/selective-version-resolutions/) if you want to override the version of `fsevents` being used down through your dependency tree.  Or, if you're using pnpm, check out [pnpm hooks](https://pnpm.js.org/en/pnpmfile) (pnpm is pretty cool).  Using either one of those, you can point the dependency for `fsevents` to @RomanHotsiy's fork

Jeddf:
chokidar, gatsby, gatsby-page-utils, gatsby-source-filesystem
```
yarn list v1.19.0
├─ fsevents@1.2.9
├─ gatsby-page-utils@0.0.25
│  └─ fsevents@2.1.0
├─ gatsby-source-filesystem@2.1.31
│  └─ fsevents@2.1.0
└─ gatsby@2.15.36
   └─ fsevents@2.1.0
```

thanks for the `yarn resolutions` tip

Js-Brecht:
Oh, I was thinking they were external dependencies 😆

paulmillr:
New fsevents is out, so just rebuild your lockfiles - chokidar will auto-use it

Js-Brecht:
Thanks @paulmillr.

@wardpeet not really a need for this PR anymore; issue is resolved by `fsevents@2.1.1`.  Unless you guys still want me to write a doc for the `CHOKIDAR_USEPOLLING` switch for debugging, we can just close it.

wardpeet:
@Js-Brecht if it's resolved we don't really need to write a doc about it. Thanks for your work and let's close this for now.

Thanks everyone for jumping on this thread! <3

PR #17945, opened 2019-09-27
chore(gatsby-design-tokens): add message when microbundle fails

wardpeet:
## Description
Improve dx for people using node < 10.16 and trying to run `yarn bootstrap` or building `gatsby-design-tokens`

image

### DIFF
--- a/packages/gatsby-design-tokens/package.json
+++ b/packages/gatsby-design-tokens/package.json
@@ -17,7 +17,7 @@
     "./dist/**/*"
   ],
   "scripts": {
-    "build": "(microbundle build --sourcemap=false -f es,cjs && agadoo dist/index.esm.js) || node -e \"console.log('\\nMicrobundle failed, make sure you\\'re using node version 10.16 or higher\\n')\"",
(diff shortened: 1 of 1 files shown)


packages/gatsby-design-tokens/package.json
```diff
@@ -17,7 +17,7 @@
     "./dist/**/*"
   ],
   "scripts": {
-    "build": "microbundle build --sourcemap=false -f es,cjs && agadoo dist/index.esm.js",
+    "build": "(microbundle build --sourcemap=false -f es,cjs && agadoo dist/index.esm.js) || node -e \"console.log('\\nMicrobundle failed, make sure you\\'re using node version 10.16 or higher\\n')\"",
```
pieh: actually, we should still `exit 1` there

Node <10.16 might not be only reason this build fail, and we might end up publishing broken release

sidharthachatterjee: 😂 Sorry for adding the label!

wardpeet: oh good catch

wardpeet: image

wardpeet:
It was super valuable, so no need to apologize!

Issue #15959, opened 2019-07-21
ESLint not working


ndstephens:
## Description

ESLint not running/working

### Steps to reproduce

Started new "hello world" Gatsby project, ran "npm run develop"

### Expected result

According to Gatsby docs "Gatsby ships with a built-in ESLint setup. For most users, our built-in ESlint setup is all you need."

### Actual result

Get ESLint warning from ESLint VSCode extension.  

"No ESLint configuration (e.g .eslintrc) found for file: /Users/.../src/pages/index.js"

### Environment

  System:
    OS: macOS 10.14.4
    CPU: (8) x64 Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 11.9.0 - ~/.nvm/versions/node/v11.9.0/bin/node
    npm: 6.10.1 - ~/.nvm/versions/node/v11.9.0/bin/npm
  Languages:
    Python: 2.7.13 - /Applications/MAMP/Library/bin/python
  Browsers:
    Chrome: 75.0.3770.142
    Safari: 12.1
  npmPackages:
    gatsby: ^2.13.31 => 2.13.31 
  npmGlobalPackages:
    gatsby-cli: 2.7.17

wardpeet:
Hey, thanks for reporting. We use webpack to do eslint checks. You can override this behaviour by creating your own eslint configuration file in your project.

The default eslint config we use can be found https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/utils/eslint-config.js

ndstephens:
Is that a similar approach to how create-react-app sets up its eslint config?  

I'm asking b/c the ESLint extension in VSCode completely recognizes the eslint config whenever I start a project using CRA.  However with the default Gatsby project it does not.  I'd be happy to use the default eslint checks, but it's not recognized.  I hope i'm making sense here.

wardpeet:
It seems like CRA adds the following snippet to package.json.
```
"eslintConfig": {
  "extends": "react-app"
}
```

This snippet is picked up by vscode eslint. We don't expose our eslint file. We can ship this with our official starters but it's pretty hard to do with gatsby-core.

stevengrimaldo:
This is still an issue btw, what the OP posted. Not sure how this responsive even remotely addresses the question either:

[...]

PR #17958, opened 2019-09-27
Correct Gatsby ESLint config link

dkthehuman:
I was looking into how to have VS Code detect my ESLint config and found [an issue](https://github.com/gatsbyjs/gatsby/issues/15959) which advises creating a custom ESLint config file.

The file referenced by @wardpeet in that comment is https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/utils/eslint-config.js whereas the link in the [Gatsby's ESLint doc](https://www.gatsbyjs.org/docs/eslint/) points to the file https://github.com/gatsbyjs/gatsby/blob/master/.eslintrc.json.

I believe the difference is that the former is for projects *using* Gatsby whereas the latter is for *working on* Gatsby itself. In the case of people looking into adding ESLint to their Gatsby project, the former seems to be more relevant.


sidharthachatterjee:
Merge on green didn't work because checks didn't complete (because the branch name has docs in it) 

Squashing and merging! Thank you @dkthehuman 👍

Issue #17954, opened 2019-09-27
add margin to code blocks in plugin document page


XHMM:
code block in plugin document page missed `margin` so caused collapse effect , it's bad for read：
![image](https://user-images.githubusercontent.com/24786522/65787346-496b8e00-e18b-11e9-8407-ff6ac23e16cb.png)

should add `margin`  to `gatsby-code-title` class and it will look good as below：
![image](https://user-images.githubusercontent.com/24786522/65787557-c991f380-e18b-11e9-9bf1-f2c77e59a0b3.png)

[associated page link](https://www.gatsbyjs.org/packages/gatsby-plugin-offline/)

sidharthachatterjee:
Fixed in https://github.com/gatsbyjs/gatsby/pull/17969

Thank you so much @ashu8912 🥇

PR #17969, opened 2019-09-28
added margin to code blocks solves #17954

ashu8912:
## Description
Code block in plugin document page missed margin so caused collapse effect , it's bad for read：
So Added a margin from top and bottom

## Related Issues
https://github.com/gatsbyjs/gatsby/issues/17954


fk:
This adds a margin below `.gatsby-code-title` that we do not want :(

image

gillkyle:
Yeah looks like this should be a one off adjustment on the page where we're seeing the overlap in the code blocks. I'll open up another issue.

PR #17977, opened 2019-09-28
Only trim pathPrefix when loading from location.pathname

kmck:
## Description

When a page had a path that coincidentally started with the same string as `pathPrefix`, its data request would fail. This happened because the path processing was overzealous about trying to remove the prefix from the string.

This PR changes `trimPathname` so that it no longer tries to remove the path prefix. Instead, it calls `stripPrefix` on the location just before it's passed to `loadPage` or `loadPageSync`. The idea is that we should only attempt to remove a path prefix when the path may actually be prefixed.

## Related Issues

Fixes https://github.com/gatsbyjs/gatsby/issues/16457

### Side effects?

This change means that `findPath()` will no longer attempt to strip the path prefix. I'm not sure whether anything was relying on this behavior (seems like it shouldn't in the first place?), but hopefully someone who knows a little more can flag it if this is going to be a problem.

### DIFF
--- a/packages/gatsby/cache-dir/__tests__/find-path.js
+++ b/packages/gatsby/cache-dir/__tests__/find-path.js
@@ -14,9 +14,9 @@ describe(`find-path`, () => {
       expect(cleanPath(`/index.html`)).toBe(`/`)
     })
 
-    it(`strip out a basePrefix`, () => {
+    it(`does NOT strip out a pathPrefix`, () => {
       global.__BASE_PATH__ = `/blog`
-      expect(cleanPath(`/blog/mypath`)).toBe(`/mypath`)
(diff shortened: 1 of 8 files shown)


robin-macpherson:
@kmck thanks for opening this PR up! I'm not sure if you saw the comments / PR reference in the issue but I already have a PR open that fixes this and got some helpful input from the core team. It's just waiting on me to finish up the e2e tests now.

https://github.com/gatsbyjs/gatsby/pull/17544

kmck:
Yes, I saw that! I thought your PR was still just checking for a slash after the prefix, but I see now that it's been changed to avoid double-processing the path (and it's more concise 😁)

Looks like I can close this. Thanks!

robin-macpherson:
Thank you @kmck ! 🙂

PR #18002, opened 2019-09-30
chore(docs): add steps in tutorial part four

### LINKED ISSUE #17979: Tutorial Part Four Issue
## Summary

I was following the Data in Gatsby ([tutorial part four](https://www.gatsbyjs.org/tutorial/part-four/)) and after performing the changes in the "[Use a page query](https://www.gatsbyjs.org/tutorial/part-four/#use-a-page-query)" section the About page gives the error `TypeError: data is undefined`

## Relevant information

After updating `src/pages/about.js` with the code in the "[Use a page query](https://www.gatsbyjs.org/tutorial/part-four/#use-a-page-query)" section, when the page reloads you get the error mentioned above along with some diagnostic info.

I've found that the development server needs to be stopped and restarted after the about.js file is updated with the code specified. Then you are able to proceed onto the next part of the tutorial. Note that the tutorial does say to restart the development server, but says to do so after you add the siteMetadata in the prior step before updating `src/pages/about.js`.

Continuing on to the next part of the tutorial "[Use a StaticQuery](https://www.gatsbyjs.org/tutorial/part-four/#use-a-staticquery)" I ran into a similar issue after updating `src/components/layout.js` and saving when the page reloads it yields the error:
```
Error: The result of this StaticQuery could not be fetched.

This is likely a bug in Gatsby and if refreshing the page does not fix it, 
please open an issue in https://github.com/gatsbyjs/gatsby/issues
```

Again, a restart of the development server fixes the issue.

I think the documentation just needs to be updated to state that restarts are needed after the changes are made in the steps indicated above.

I've been through a clean start of this tutorial a couple of times to verify, (with my last attempt being sure to copy and paste the code directly from the tutorial), but if anyone else can confirm that this isn't just me, then I will make an update to the docs and submit a PR.

rgroves:
## Description

A restart of the development server is required after the "Use a page query" section and after the "Use a StaticQuery" section changes. Without a server restart at those points, the user hits errors at those points in the tutorial.

## Related Issues

Fixes #17979

### DIFF
--- a/docs/tutorial/part-four/index.md
+++ b/docs/tutorial/part-four/index.md
@@ -235,8 +235,6 @@ module.exports = {
 }
 ```
 
-Restart the development server.
-
 ### Use a page query
 
 Now the site title is available to be queried; Add it to the `about.js` file using a [page query](/docs/page-query):
@@ -270,7 +268,9 @@ export const query = graphql`
 // highlight-end
 ```
 
-It worked! 🎉
+When the page reloads, you'll see an error message at this point. After making the changes above you'll need to restart the development server. The restart is required because page queries happen at build time (in this case, when you run `gatsby develop`). This is why you saw an error message, the data the page expects to use was not available because the page query you added was not there at build time previously.
+
+After restarting the development server, refresh your page and you'll see that it worked! 🎉
 
 ![Page title pulling from siteMetadata](site-metadata-title.png)
 
@@ -354,7 +354,9 @@ export default ({ children }) => {
 // highlight-end
 ```
 
-Another success! 🎉
+When the page reloads, you'll see an error message just like when you added the page query in the previous section. The development server needs to be restarted again. This is because the graphql query (specified in the useStaticQuery hook) happens at build time as well, just like a page query.
+
+After restarting the development server, refresh your page. Another success! 🎉
 
 ![Page title and layout title both pulling from siteMetadata](site-metadata-two-titles.png)
 


sidharthachatterjee:
I like this but I wonder if the addition is a little too verbose? 🤷‍♂ Curious to know what @marcysutton has to say!

rgroves:
If the tutorial wasn't aimed at beginners I may have used something terse like, "Restart the development server." Since it is though, I think telling them what they need to do, as well as why they need to do it is beneficial. More than happy to edit it down if required though.

dhhdev:
It would be a good place to actually say why someone needs to restart their development server. It might avoid misconceptions later on in the learning curve.

When I first dived into Gatsby, less than a week ago - I didn't fully understand how/why it actually needed to restart. I just got told that I should restart after every query. (Jason Lengstorf makes some good live streams and posts them to YouTube as well.)

I would say a tutorial's main job is to be **verbose**. Not everyone knows what React, GraphQL etc. is, when they tried out their first Gatsby tutorial. I surely didn't...

KyleAMathews:
What errors are you seeing? You shouldn't have to restart the server when changing queries so perhaps you've run into a bug which needs fixed.

rgroves:
@KyleAMathews it's documented in the issue I opened for it here: #17979  
If you walk through the tutorial step-by-step as written you'll run into the error described.

rgroves:
The docs state that a page query and the useStaticQuery hook happen at build time:

> During the Gatsby build process, GraphQL queries are pulled out of the original source for parsing.
— (source: https://www.gatsbyjs.org/docs/page-query/#the-short-answer)

> Gatsby v2.1.0 introduces useStaticQuery, a new Gatsby feature that provides the ability to use a React Hook to query with GraphQL at build time.
— (source: https://www.gatsbyjs.org/docs/use-static-query/)

So if the code specified in the tutorial is added after the development server is running I don't understand how it could work without error. This is a legit question, as I'm not familiar with Gatsby (or React or GrahpQL) and am coming at this as a beginner trying to improve the tutorials for other beginners who might otherwise just give up when running into an error during the tutorials.

I thought this would work toward the goal stated on the main tutorials page: 

> The full tutorial is intended to be as accessible as possible to people without much web development experience (yet!) — no need to be an expert. 
— (source: https://www.gatsbyjs.org/tutorial/)

muescha:
and for a tutorial and for a beginner this can be also unknown what means:

`restarting the development server`

just refreshing the page? running some commands to rebuilds?

it would be nice if it gives a hint:
```
restarting the development server (hit CTRL+C and `gatsby deploy`)
```

jonniebigodes:
The document has to factor in that the person that follows this might have or not knowledge of graphql or react, based on what @dhhdev mentioned.  Reading up on it i would simplify it, Before the "It worked! 🎉" i would add something like the following "Based on the nature of how Gatsby handles graphql queries, after making this change, restart the development server so that the newly created query, can be picked up by Gatsby and do the same in the next section". At this point, as this is a tutorial, the reader should be focused on going through the tutorial and see the results in action and grasp the features of gatsby, it does not need to know that queries execute at build time, there's a time for that and a place(documentation). But nonetheless it's really nice that you've made the changes so readily and also going to defer to the @gatsbyjs/learning and @marcysutton on how to proceed on this.

rgroves:
@KyleAMathews I noticed you are on the gatsby/learning team, and so hopefully can provide further direction on this. In response to your question above, do you need more info than was is reported in the original issue (#17979) by myself and @jonniebigodes (who added the helpful screen shots of the actual error messages)?

wardpeet:
Hi @rgroves. Thanks for writing these additions to our tutorial. I followed the tutorial again and I didn't have to restart my development server. I didn't get any error.

Could you maybe create a reproduction in like a gif or video that we could watch? Graphql queries are running on build while doing `gatsby build`, when doing `gatsby develop` we have hot module reloading (HMR) enabled so we listen for changes in page & static queries. Which means we update your site when queries change.

jonniebigodes:
@wardpeet i've helped triage this and the issue existed. As you can see based on my response in the original issue. I checked and i still have the reproduction used. If you want i can hoist the code to a repo and you can take a look it.

rgroves:
> Could you maybe create a reproduction in like a gif or video that we could watch? 

@wardpeet I'll create a vid showing the problem as soon as I am able to. This issue only presents itself if you follow along step-by-step in the order things are presented in the tutorial. If all of the files are added first and then `gatsby develop` is done then changes to the query data work as expected. 

> Graphql queries are running on build while doing `gatsby build`, when doing `gatsby develop` we have hot module reloading (HMR) enabled so we listen for changes in page & static queries. Which means we update your site when queries change.

I'm not familiar with how the HMR works, but I have a feeling there might be a difference between how things are working when queries are present during the initial `gatsby develop` step and changing them after (which works) versus already having done `gatsby develop` and adding a query to a page or component that did not have one prior to doing the initial `gatsby develop`.

Also, in case it matters, at the time I went through the tutorial I was using the following:
```
Gatsby CLI version: 2.7.50
Gatsby version: 2.15.28
```

rgroves:
I tried walking through tutorial part four again, and I can no longer reproduce the issue. 

There are various differences between the package.json and lock file for the starter repo used in this tutorial (gatsby-starter-hello-world) as they exist freshly cloned today versus the one cloned back on 9/28. If I go back and use the package.json and package-lock.json files as they existed on the 28th I'm able to reproduce.

In any case, I assume some other update has fixed the original reported issue so I'm closing this PR. This also closes issue #17979

PR #18007, opened 2019-09-30
chore(showcase): Add surge 3

dillonbrowne:
add surge3.com to showcased websites.


sidharthachatterjee:
There's a couple of invalid categories here (https://github.com/gatsbyjs/gatsby/runs/242530589)

@dillonbrowne Could you remove them please?

PR #18008, opened 2019-09-30
perf(schema): speed up and lower memory usage of type inference

me4502:
## Description

Type inference was completely hammering the garbage collector before with a large schema, and started to cause issues and node crashes when I tried lowering the memory. This takes it from 20s to 5s, and prevents the massive GC churn.

### DIFF
--- a/packages/gatsby/src/schema/infer/example-value.js
+++ b/packages/gatsby/src/schema/infer/example-value.js
@@ -100,7 +100,12 @@ const getExampleObject = ({
           value = value.find(v => v != null)
           arrays--
         }
-        return acc.concat(value)
+        if (_.isArray(value)) {
+          acc.push(...value)
+        } else {
+          acc.push(value)
+        }
+        return acc
       }, [])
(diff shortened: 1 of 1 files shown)


packages/gatsby/src/schema/infer/example-value.js
```diff
@@ -100,7 +100,12 @@ const getExampleObject = ({
           value = value.find(v => v != null)
           arrays--
         }
-        return acc.concat(value)
+        if (_.isArray(value)) {
+          acc.push(...value)
```
muescha: i need to test for Array? why not always add with `...`?

me4502: Because if you spread an object it behaves very differently to just adding it

pieh:
Published in `gatsby@2.16.3`

rroslaniec:
After that merge request our pipeline fails just like that...

```
info Done building in 135.270751079 sec
error Command failed with exit code 1.
ERROR: Job failed: exit code 1
```

PR #18017, opened 2019-10-01
docs(blog): lowercase netlify toml keys

NickyMeuleman:
Change to: https://www.gatsbyjs.org/blog/2018-12-17-turning-the-static-dynamic/

Lowercased the `netlify.toml` keys in the example to fit [the netlify reference](https://www.netlify.com/docs/netlify-toml-reference/)

Thank for the blogpost @sw-yx ! Really helped a lot setting up a project that uses Netlify functions (with TypeScript, hence `netlify-lambda` and not `netlify-dev`)


sw-yx:
yeah its actually case insensitive so doesnt matter but apparently according to the spec its lowercase? idk

PR #18035, opened 2019-10-02
fix(gatsby-image): add itemprop on placeholder images

### LINKED ISSUE #17419: feat(gatsby-image): Have 'itemProp' attribute also rendered in HTML
## Summary

With the current [`itemProp` implementation](https://github.com/gatsbyjs/gatsby/pull/11296) in gatsby-image, the `itemprop` HTML attribute is applied to the `` after the JS (ie, lazyloading) is run. I think it would be worthwhile to also apply the attribute to the `` that lives in `<noscript>` tag; that way, it's already in the source and ready for search engines to crawl.

### Motivation

It's unclear whether or not structured data—microdata format, specifically—that is added via JS is readable by search engines.

According to [Google's docs](https://developers.google.com/search/docs/guides/intro-structured-data#structured-data-format) and a [discussion in a Google forum](https://support.google.com/webmasters/forum/AAAA2Jdx3sUgsdps9ncUQA/?hl=en&gpf=%23!topic%2Fwebmasters%2Fgsdps9ncUQA), it's possible if the format is JSON-LD but there's no info on it working with microdata.

jamessimone:
* Fixes #17419 
* Fixes a logging issue generated by no key value set within a map statement for the `recipe-gatsby-images` when running Gatsby test suite


examples/recipes-gatsby-image/src/pages/examples/directory.js
```diff
@@ -42,8 +42,9 @@ export default () => {
 
   return (
     <Layout>
-      {images.map(image => (
+      {images.map((image, index) => (
         <Img
+          id={index}
```
wardpeet: could you explain why you've putten id here?

jamessimone:
@wardpeet good question - how embarrassing though, that was meant to add the `key` prop to get rid of a warning I saw while running the tests. Pushed up that change, thanks for double-checking!

jamessimone:
@wardpeet thanks!

PR #18051, opened 2019-10-02
fix: runtime network/resource loading resilience fixes + restore e2e tests for it

Darking360:
# Cypress blocked scenarios tests ⚙︎

## Description

This is related to an issue opened by @sidharthachatterjee, the problem is related with the Cypress tests here https://github.com/gatsbyjs/gatsby/tree/master/e2e-tests/production-runtime/cypress/integration/resource-loading-resilience.js, the thing is that one different resource is being blocked for each test round, so for example blocking `Index` should run all the `Loads Index`, `Navigates to second page` and so on, the problem is that before it only restore the blocked resources but failed to run the tests all over to the other page cases like `Index`, `Page-2` and `404`.

Passing down the `tag` of the scenario brings a better reading for the test `describe` and it `statements` run every time, also having the `runTests` inside an `it` case didn't add the dynamic test cases correctly, so refactoring this into a `describe` statement running an async `before` function before each dynamic `it` that is going to be generated we guarantee that the resources are being restored and blocking the next case.

Now when you see Cypress tests, you'd see around 70 test cases being generated which is the correct behavior. Now the problem is that there are some test cases failing because of the blocking behavior, if we're blocking the `Index` for example, we would not get navigation so we cannot assert on navigation item for page 2 for example 🤔 I don't know if I should fix the test cases too, or only fix the Cypress problem in this PR, waiting on @sidharthachatterjee response for this 🚀 

## Related Issues

https://github.com/gatsbyjs/gatsby/issues/16458

### Attachments 

Here is the [link](https://dashboard.cypress.io/#/projects/is8aoq/runs/1030/specs) for the Cypress run on the tests for all scenarios.

Some attachments to show Cypress tests running the test cases:

image

image

### Old PR

https://github.com/gatsbyjs/gatsby/pull/17458

Opened it again because of Hacktober fest 🚀 

### New tests fixed

I was fixing some duplicated code I had 😅 What I did is basically add the expected behavior in a little object of options to the function that runs the tests for each scenario, the challenge here is that each time we block a resource for page, query results, page assets and so on, we get dynamic tests which are the why of this PR but we need to assert different scenarios results for blocking the Index, the second page and the 404, for example, we have different behavior there since we test it with cy.visit().

That being said, I think I found the pattern of those adding an object with some variables for link working or not, if it should stay (if the page was removed from page template asset for example) or not and for 404 safe behavior (due to the cy.visit()). All 72 dynamic generated tests are working now for both online and offline behaviors; also included conditionals to toggle some variables here depending on offline or not.

All the asserts work depending on the blocked or not resource and based on the variables from the options object, I pull a helper for asserting on the navigation from a click element called assertOnNavigate to reduce duplicated code in general. I think that'd be it, let me know your thoughts on this one @wardpeet and thank you 🙏


pieh:
> Now when you see Cypress tests, you'd see around 70 test cases being generated which is the correct behavior. Now the problem is that there are some test cases failing because of the blocking behavior, if we're blocking the `Index` for example, we would not get navigation so we cannot assert on navigation item for page 2 for example 🤔 I don't know if I should fix the test cases too, or only fix the Cypress problem in this PR

Let's start by restoring those tests, fixing bugs in core might take longer but we need to fix them.

> That being said, I think I found the pattern of those adding an object with some variables for link working or not, if it should stay (if the page was removed from page template asset for example) or not and for 404 safe behavior (due to the cy.visit()). All 72 dynamic generated tests are working now for both online and offline behaviors; also included conditionals to toggle some variables here depending on offline or not.

So goal here is that even if page template can't be loaded (either for page you visit with `cy.visit` or by "clicking" gatsby-link element) we should end up seeing that page (we should fallback to pure html). So in scenario where page template chunk for `page-2` is blocked, and we try to navigate to it from `index` we should still end up on page-2 (even if js won't be working) - this is better solution than failing silently and staying in `index` page. This is what those tests were doing initially, before they broke :S

I will be going over those fixes tomorrow and adjusting where needed, because I just tested this locally (no sure why CircleCI refuses that run tests here) and tests are passing, but behaviour is not what we want. So after my changes we will probably see bunch of failing tests and this is something we will need to fix in core

packages/gatsby/cache-dir/loader.js
```diff
@@ -291,7 +291,7 @@ export class BaseLoader {
   }
 
   hovering(rawPath) {
-    this.loadPage(rawPath)
+    // this.loadPage(rawPath)
```
pieh: debugging only - will need to restore

packages/gatsby/cache-dir/loader.js
```diff
@@ -353,7 +353,12 @@ const createComponentUrls = componentChunkName =>
 export class ProdLoader extends BaseLoader {
   constructor(asyncRequires, matchPaths) {
     const loadComponent = chunkName =>
-      asyncRequires.components[chunkName]().then(preferDefault)
+      asyncRequires.components[chunkName]()
+        .then(preferDefault)
+        .catch(() => {
+          console.log(`failed to load component but not an unhandled exception`)
+          return null
```
pieh: not sure if best place to catch - but we do handle cases when this resolves to `null` in loader - so this was quickest way to "fix" that

packages/gatsby/cache-dir/loader.js
```diff
@@ -55,7 +55,7 @@ const loadPageDataJson = loadObj => {
       // If the request was for a 404 page and it doesn't exist, we're done
       if (pagePath === `/404.html`) {
         return Object.assign(loadObj, {
-          status: `failure`,
+          status: `error`,
```
pieh: this is tricky - if for some reason page-data for 404.html doesn't exist and we set failure then we hit this code path:
https://github.com/gatsbyjs/gatsby/blob/8fd5877ed74d89d17633ea380222d8b9d7a9fc09/packages/gatsby/cache-dir/loader.js#L186-L191

but this also just keep user on current page

I do like throwing error so error tracked could pick this up, but I feel like trying no navigate to page will be better end user experience than keeping them on current page - would still love some error for error trackers tho

packages/gatsby/cache-dir/loader.js
```diff
@@ -381,6 +386,47 @@ export class ProdLoader extends BaseLoader {
...
+            if (req.status === 200) {
+              // page actually exist (or we asked for 404 )
+              return {
+                status: `error`,
+              }
+            } else if (req.status === 404) {
+              // page truly doesn't exist
+              return data
+            } else {
+              console.log(`[prod-loader] something else???`, req.status)
```
pieh: in json loading code we only handle code 200, 404 ... and 500 - do we need to handle more in HEAD request check than 200 and 404?

packages/gatsby/cache-dir/loader.js
```diff
@@ -353,9 +372,20 @@ const createComponentUrls = componentChunkName =>
...
+          () =>
+            // failed to load component but not an unhandled exception
+            // because we handle case of component resolving to null
+            null
+        )
 
-    super(loadComponent, matchPaths)
+    // check if component is available in chunks dictionary
+    const pageDataValidator = pageData =>
+      !!asyncRequires.components[pageData.componentChunkName]
```
pieh: Controversial:
What this does it adds additional validation to page-data. We currently check if file is event a json file and also if it contains `path` field.

This adds check to see if currently loaded bundle have listed page template inside (we need some check for it, because otherwise we will crash with `asyncRequires.components[pageData.componentChunkName] is not a function` error while we try to `loadComponent`.

It's marked as controversial here, because of the flow that happens when page-data is considered invalid - we load 404 (in browser).

There can be case been made that we instead should try to navigate to this page using regular browser navigation: loaded webpack bundle might be already outdated (there was new deploy happening), tho it's not clear how to best achieve that. Navigating using regular browser navigation (not clientside one) would automatically reload bundles (if html site - either existing one or 404 one have updated script resources).

packages/gatsby/cache-dir/__tests__/loader.js
```diff
@@ -388,22 +391,18 @@ describe(`Production loader`, () => {
       expect(emitter.emit).toHaveBeenCalledTimes(0)
     })
 
-    it(`should throw an error when 404 cannot be fetched`, async () => {
+    it(`should return am error when 404 cannot be fetched`, async () => {
```
wardpeet:
```suggestion
    it(`should return an error when 404 cannot be fetched`, async () => {
```

PR #18079, opened 2019-10-03
feat(gatsby-source-wordpress): add normalizers option to modify normalizers

paweljedrzejczyk:
## Description

Adding `normalizers` option that allows to add/remove normalizers

## Example use
This allows for adding the normalizer for e.g. filtering out media library items from being processed/downloaded when they don't meet the criteria - e.g. are not attached to any page.

```js
const dropUnusedMediaNormalizer = {
  name: "dropUnusedMediaNormalizer",
  normalizer: function({ entities }) {
    return entities.filter(e => {
      if (e.__type === "wordpress__wp_media" && !e.post) {
        return false
      }
      return true
    })
  },
}
``` 

```javascript
module.exports = {
  plugins: [
    {
      resolve: "gatsby-source-wordpress",
      options: {
        // ...
        normalizers: normalizers => [dropUnusedMediaNormalizer, ...normalizers],
      },
    },
  ],
}
```


wardpeet:
Can you explain your use case for this? Is it just to speed up things?

paweljedrzejczyk:
@wardpeet We have a wordpress site with large media library due to many custom post types and at the same time the same wordpress instance is used to generate site with gatsby. We don't need any of the media files that are attached to custom post types in gatsby, just those attached to pages in ACF.

It greatly speeds up the build time, since for our case it don't need to download 2000 extra images every time.

default `normalizer` option that could be used to do it is triggered after all media is downloaded, this is the same `normalizer` function but trigerred just after fetching the entities before any gatsby processing.

wardpeet:
I don't mind this change as I think it's an easy way to filter out entities before they are being fetched. I would suggest naming the option: filterEntities or something similar, normalizer doesn't seem right to me.

let me ping our wordpress folks @jasonbahl @TylerBarnes

TylerBarnes:
This is something I've wanted in this plugin before, it's definitely a useful feature. Thanks @paweljedrzejczyk ! 
I think calling it a normalizer still makes sense since it runs before all the other normalizers and could be used either to filter or normalize entities.
The naming is a bit clunky but I can't think of anything better. Maybe `initialNormalizer`? Not sure.

@pieh I believe you've worked on this plugin the most, it would be good to hear your thoughts.

pieh:
This sets a bit of dangerous precedence of adding different config field that will run similar function as one that already exists (`normalizer`) just before all the builtin normalizer functions.

I don't oppose adding some way to do this, because there are definitely use cases for it as described in https://github.com/gatsbyjs/gatsby/pull/18079#issuecomment-540015983 .

But from API design perspective this is troublesome, because what if someone will have use case to add something similar in the middle of normalizers chain we have in the plugin ( https://github.com/gatsbyjs/gatsby/blob/c7a5367cb01d358a08524117db5c4d6025ee3818/packages/gatsby-source-wordpress/src/gatsby-node.js#L91-L188 ) to workaround something else?

I feel like we should have a way to declare order where we want to insert the normalizer - maybe similar to priority you can declare in wordpress action/filter setup (?). This is something I don't like very much honestly (using arbitrary priority numbers) in the user code, so maybe something that could work like this:

```js
// get internal normalizers identifier so you could order normalizer function execution relative to those
const { NormalizerSteps } = require(`gatsby-source-wordpress`)

module.exports = {
  plugins: [
    {
      resolve: `gatsby-source-wordpress`,
      options: {
        // [...] regular options
        // overload normalizer option (make sure current usage continue to work)
        normalizer: [
          {
            // declare that this should run before this step
            // for completeness, probably should support both before and after
            before: NormalizerSteps.mapEntitiesToMedia,
			cb: entities => {
              // do something with entities
              return entities
            }
          }
        ]
      }
  ]
}
```

This is just an idea (probably not very good one). This is open for discussion

TylerBarnes:
@pieh I like that idea. Are you thinking about priority in case sub-plugins also use the option and there are multiple normalizers on the same step?
For now it seems like it might be unnecessary but the rest holds up to me. 

It's just semantics but I like the idea of adding a new option `normalizers` instead of using the existing one:

```javascript
module.exports = {
    plugins: [
        {
            resolve: `gatsby-source-wordpress`,
            options: {
                // [...] regular options
                // overload normalizer option (make sure current usage continue to work)
                normalizers: [
                    {
                        // declare that this should run before this step
                        before: NormalizerSteps.mapEntitiesToMedia,
                        normalizer: entities => {
                            // do something with entities
                            return entities
                        }
                    },
                    {
                        // declare that this should run after this step
                        after: NormalizerSteps.mapEntitiesToMedia,
                        normalizer: entities => {
                            // do something with entities
                            return entities
                        }
                    }
                ]
            }
        }
    ]
}
```

pieh:
> Are you thinking about priority in case sub-plugins also use the option and there are multiple normalizers on the same step?

Yeah, that would be tricky with what I proposed. But this is generally difficult because just the nature of sub plugins should be that they shouldn't need to know or care about other subplugins. If they would need to handle races/ordering between other subplugins, it would break subplugin separation contract. We kind of see this problem with some of remark subplugins, where order of declaration in subplugin list matters in some cases, so this is defenitely something to think about, but at this point I don't have any good solutions.

> It's just semantics but I like the idea of adding a new option `normalizers` instead of using the existing one:

Yeah, agree here. Implementation wise, we would probably convert current `normalizer` plugin option to use `normalizers` construct internally, so all normalizers use same code path in the end.

> Edit: We could maybe even just have the normalizer steps return a number and sort all normalizers by that field before running them one by one.

This seems dangerous for user code. We would assign arbitrary numbers there, and users might end up using literal numbers instead of importing them. Which means that if we needed to shift normalizer "priority" (i.e. if we would have normalizer "A" (priority 10) and normalizer "B" (priority 20), add later 10 more normalizers in between we could run out of integer number in range we gave ourselves - should we use floating point priorities there?). I'm pretty surprised that this technique works in wordpress (from things I know about where they use this convention is priority for add_action/add_filter, admin menu items positions) - maybe I just never hit edge cases with that, when some plugins would clash for same priority causing problems in the end.

TylerBarnes:
In WordPress if you add two filters with the same priority, they run in the order they were added. If we were to compare this directly to WordPress though, each normalizer here would probably have it's own filter name which minimizes the problem of adding additional internal normalizers later. In that comparison it would be a mix of the two ideas:

```javascript
const { NormalizerSteps } = require(`gatsby-source-wordpress`)

module.exports = {
    plugins: [
        {
            resolve: `gatsby-source-wordpress`,
            options: {
                // [...] regular options
                normalizers: [
                    {
                        step: NormalizerSteps.A,
                        priority: 11,
                       // this would run after A
                        normalizer: entities => {
                            // do something with entities
                            return entities
                        }
                    },
                    {
                        step: NormalizerSteps.B,
                        priority: -1,
                       // this would run before B
                        normalizer: entities => {
                            // do something with entities
                            return entities
                        }
                    }
                ]
            }
        }
    ]
}
```

Then if we needed to add 3 more normalizers between normalizer A and B, everything would still work since the internal normalizer order would be based on name, not number and the order of external normalizers would be relative to the name, but offset by number.

I guess the main difference is that with after/before you have 2 priorities (where all external normalizers with the same before or after priority run in whichever order they were added) and if you use integer priorities, you can change the order of external normalizers that are using the same internal step (which would play nice with sub-plugins).

I'm also not sure if this is a good way to do it but it feels like it's not far off at least.

## Alternative idea
We could also import all internal normalizers to determine which order the external normalizer should come in. I think that falls over in the case of sub-plugins though.

```javascript
const { Normalizers } = require(`gatsby-source-wordpress`)

module.exports = {
    plugins: [
        {
            resolve: `gatsby-source-wordpress`,
            options: {
                // [...] regular options
                normalizers: [
                    {
                        normalizer: entities => {
                            return entities
                        }
                    },
                    Normalizers,
                    {
                        normalizer: entities => {
                            return entities
                        }
                    }
                ]
            }
        }
    ]
}
```

`Normalizer` could just be an object of normalizers, so you could change the order by doing something like this:

```javascript
const { Normalizers } = require(`gatsby-source-wordpress`)

const { mapEntitiesToMedia, ...normalizers } = Normalizers

module.exports = {
    plugins: [
        {
            resolve: `gatsby-source-wordpress`,
            options: {
                // [...] regular options
                normalizers: [
                    {
                        normalizer: entities => {
                            return entities
                        }
                    },
                    mapEntitiesToMedia,
                    {
                        normalizer: entities => {
                            return entities
                        }
                    },
                    normalizers
                ]
            }
        }
    ]
}
```

If no internal normalizers were added to the plugin option, we could just prepend them to the array.

muescha:
you mean `Normalizers` is already in the right order?

how about an array with an name so it can be manipulated by finding the index of an normalizer in the array and then insert there?

```javascript
Normalizers = {
        before: ...,
		after: ...,
		values: [
			{
				name:'mapTagsCategoriesToTaxonomies',
 				normalizer:normalize.mapTagsCategoriesToTaxonomies 
    		},
			{
				name:'searchReplaceContentUrls',
 				normalizer:normalize.searchReplaceContentUrls 
    		}
		]
	}
```

and use the helpers:
```javascript

const { Normalizers } = require(`gatsby-source-wordpress`)

Normalizers.before('searchReplaceContentUrls', {
						name: 'myFirstNormalizer'
                        normalizer: entities => {
                            return entities
                        }
                    })

module.exports = {
    plugins: [
        {
            resolve: `gatsby-source-wordpress`,
            options: {
                // [...] regular options
                normalizers: Normalizers.values
```

TylerBarnes:
@muescha , yeah they would already be in the right order. I like the idea of just passing an array of normalizers. In the spirit of keeping the API as simple as possible maybe it should take a function that's expected to return an array, so similar to what you have @muescha , but just using native JS Array methods to add your own normalizers. For ex:

```javascript
module.exports = {
    plugins: [
        {
            resolve: `gatsby-source-wordpress`,
            options: {
                // [...] regular options
                normalizers: normalizers => {
                    // modify normalizers array as needed, splice, push, etc.
                    return normalizers;
                }
            }
        }
    ]
}
```

The nice thing about that is it's probably the simplest way we could implement this and it will also play nicely with sub-plugins. Each normalizer in the array would be an object with `name` and `normalizer` properties like you mentioned @muescha for easy array manipulation. 

When we loop over them and run the normalizers on the `gatsby-source-wordpress` side of things, the loop can just check if each is a function or if it has a normalizer property that's a function, then either run the function or run the nested function in the normalizer property.

The downside is that someone can remove any normalizers, but that's also an upside and makes `gatsby-source-wordpress` much more flexible.

muescha:
sounds good as plain Array


then maybe an helper as second parameter:
```javascript
                normalizers: (normalizers, helper) => {
```

paweljedrzejczyk:
@muescha @TylerBarnes @pieh @wardpeet 

I rewritten the PR with the ideas discussed above (Added a normalizers option to add/remove normalizers that are stored in an Array.) Let me know what you think.

packages/gatsby-source-wordpress/README.md
```diff
@@ -140,6 +140,11 @@ module.exports = {
         // URLs for thumbnails, or any other media detail.
         // Defaults to false
         keepMediaSizes: false,
+        // add/remove wordpress data normalizers
+        // each normalizer is an object with name and normalizer property being a function that accepts
+        // object with entities array and different helpers and configuration settings
+        normalizers: normalizers =>
+          [dropUnusedMediaNormalizer].concat(normalizers),
```
muescha: `unshift` add at beginning of an array:

```suggestion
          normalizers.unshift(dropUnusedMediaNormalizer),
```

packages/gatsby-source-wordpress/src/gatsby-node.js
```diff
@@ -88,111 +91,190 @@ exports.sourceNodes = async (
...
+        cache,
+        createNode,
+        createNodeId,
+        touchNode,
+        getNode,
+        auth,
+        reporter,
+        keepMediaSizes,
+      }) =>
+        await normalize.downloadMediaFiles({
```
muescha: maybe can be shorter?

```suggestion
        async(helpers) => await normalize.downloadMediaFiles(helpers)
```

packages/gatsby-source-wordpress/README.md
muescha: yes. Somehow better for readability. But that's only nitpicks

muescha:
👍 Looks good to me. 

Hard to distinct the code between `normalizer` and `normalizers` with additional `s`


I would pass always all `props` or `helpers` to to subfunction  - and destruct there the props. But that's only my personal taste: "always equal"

muescha:
You are assign the values to an underscore value - but later You not use always the underscore values. Is this redundant?

paweljedrzejczyk:
> You are assign the values to an underscore value - but later You not use always the underscore values. Is this redundant?

I see that it mixed up in the whole gatsby-node.js file. I might be missing something but I don't see a reason for plugin accepting `foo` option and then assigning that to `let _foo = foo;`. So all of that code could be cleaned up with just passing the values as is instead of assigning those to underscore values first.

I wanted to change as little things as possible. Currently the `downloadMediaFiles` normalizer accepts `_auth`, but custom normalizer that we can currently add with `normalizer` option is triggered with `auth` not `_auth`.

paweljedrzejczyk:
> Hard to distinct the code between `normalizer` and `normalizers` with additional `s`

I am looking for a better name for it. Maybe `normalizersCustomizer`

paweljedrzejczyk:
@TylerBarnes @muescha @pieh @wardpeet This PR has been sitting for a while. Do you have an idea what is required to get this into completion before can be merged? Would love to give this a final touch.

packages/gatsby-source-wordpress/README.md
```diff
@@ -755,15 +751,17 @@ To learn more about image processing check
 ## Using a custom normalizer
 
 The plugin uses the concept of normalizers to transform the json data from WordPress into
-GraphQL nodes. You can extend the normalizers by passing a custom function to your `gatsby-config.js`.
+GraphQL nodes. You can extend the normalizers by modyfing the normalizers array in plugin options in `gatsby-config.js`.
```
TylerBarnes: Should be "modifying"

packages/gatsby-source-wordpress/src/gatsby-node.js
```diff
@@ -91,139 +90,190 @@ exports.sourceNodes = async (
...
+    )
+  }
+
+  // Normalize data & create nodes
+  await wordpressDataNormalizers
+    .map(n => n.normalizer)
+    .reduce(async (entities, normalizerFunction) => {
+      entities = await entities
+      return await normalizerFunction({ entities, ...normalizerHelpers })
+    }, Promise.resolve(entities))
```
TylerBarnes: Is `Promise.resolve(entities)` necessary here?

TylerBarnes:
@paweljedrzejczyk thanks for doing this, it looks good and works well from my local testing. @pieh curious if you have any more thoughts on this?
The PR adds this new option but also replaces `normalizer` with `normalizers` in the docs/example. I'm not sure what best practises are around new features that replace old ones and documentation. Should we keep both options and document both, or deprecate `normalizer` in favour of `normalizers`?

packages/gatsby-source-wordpress/README.md
```diff
@@ -140,10 +140,10 @@ module.exports = {
...
         // Defaults to false
         keepMediaSizes: false,
-        // use a custom normalizer which is applied after the built-in ones.
-        normalizer: function({ entities }) {
-          return entities
-        },
+        // add/remove wordpress data normalizers
+        // each normalizer is an object with name and normalizer property being a function that accepts
+        // object with entities array and different helpers and configuration settings
+        normalizers: normalizers => [dropUnusedMediaNormalizer, ...normalizers],
```
pieh: Can we try not to use `dropUnusedMediaNormalizer` here (as it's not explained what it is at this point).

Wondering if this could be something like this:
```suggestion
        normalizers: normalizers => [
          ...normalizers,
          {
            name: "nameOfTheFunction",
            normalizer: function({ entities }) {
             // manipulate entities here
             return entities
           },
          }
        ],
```
to vizualize the structure of normalizer here. The `dropUnusedMediaNormalizer` is used as example below, so this won't be missed

packages/gatsby-source-wordpress/src/gatsby-node.js
```diff
@@ -88,111 +91,190 @@ exports.sourceNodes = async (
...
-      auth,
-      verboseOutput,
-      perPage,
-      searchAndReplaceContentUrls,
-      concurrentRequests,
-      excludedRoutes,
-      keepMediaSizes,
+    wordpressDataNormalizers.push({
+      name: `customNormalizer`,
+      normalizer: helpers => _normalizer(helpers),
```
pieh:
```suggestion
      normalizer: _normalizer,
```

packages/gatsby-source-wordpress/src/gatsby-node.js
```diff
@@ -88,111 +91,190 @@ exports.sourceNodes = async (
...
-    createContentDigest,
+  wordpressDataNormalizers.push({
+    name: `createNodesFromEntities`,
+    normalizer: ({ entities, createNode, createContentDigest }) =>
+      normalize.createNodesFromEntities({
+        entities,
+        createNode,
+        createContentDigest,
+      }),
   })
```
pieh: Does it make sense to allow users to mess with placement of `createNodesFromEntities` normalizer? I think we will always want this to be executed last, so IMO, this should be pushed to normalizers array after we execute `_normalizers` function. What do you think?

packages/gatsby-source-wordpress/src/gatsby-node.js
```diff
@@ -88,111 +91,190 @@ exports.sourceNodes = async (
...
 
+  if (typeof _normalizers === `function`) {
+    wordpressDataNormalizers = _normalizers(
+      [...wordpressDataNormalizers],
+      normalizerHelpers
```
pieh: Should this second parameter be here? If so we should have it documented as well (docs changes just mention first one - `normalizers`

packages/gatsby-source-wordpress/README.md
```diff
@@ -140,10 +140,10 @@ module.exports = {
         // URLs for thumbnails, or any other media detail.
         // Defaults to false
         keepMediaSizes: false,
-        // use a custom normalizer which is applied after the built-in ones.
-        normalizer: function({ entities }) {
-          return entities
-        },
```
pieh: I don't think we should remove this from documentation completely. People might be using that and it will still conitnue to work. This is essentially simplified version of `normalizers` (with less power), so maybe we could just add mention that if you need more control over order of normalizers (internals / custom one) - to use `normalizers`

packages/gatsby-source-wordpress/src/gatsby-node.js
pieh: Can we switch from this map + reduce to something like this:
```js
  // Normalize data & create nodes
  for (const { normalizer } of wordpressDataNormalizers) {
    entities = await normalizer({
      entities,
      ...normalizerHelpers,
    })
  }
```

I think it's easier to read and we don't have this weird `entities = await entities` that is confusing

packages/gatsby-source-wordpress/README.md
```diff
@@ -797,6 +799,38 @@ module.exports = {
...
 and also your `wordpress-source-plugin` options from `gatsby-config.js`. To learn more about the passed object see the [source code](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-source-wordpress/src/gatsby-node.js).
 
+### Example with modyfing gatsby-source-wordpress normalizers
+
+```javascript
+const dropUnusedMediaNormalizer = {
+  name: "dropUnusedMediaNormalizer",
+  normalizer: function({ entities }) {
+    return entities.filter(e => {
+      if (e.__type === "wordpress__wp_media" && !e.post) {
```
muescha: I know it is example code - but it can be shorter?

```suggestion
      return !(e.__type === "wordpress__wp_media" && !e.post) {
```

paweljedrzejczyk:
@pieh @muescha @TylerBarnes fixed issues after code review, let me know if there is anything more to fix

packages/gatsby-source-wordpress/README.md
```diff
@@ -144,6 +144,19 @@ module.exports = {
         normalizer: function({ entities }) {
           return entities
         },
+        // add/remove wordpress data normalizers
+        // each normalizer is an object with name and normalizer property being a function that accepts
+        // object with entities array and different helpers and configuration settings
```
TylerBarnes: After reading this again after so long it occurs to me these comments could be a little clearer. What do you think about this?

```
        // The normalizers option allows you to manipulate the array of internal
        // normalizers that are applied to entities after they're fetched
        // from WordPress.
        // You can add your own normalizers to this array by adding an object
        // that contains name and normalizer properties.
        // Name is the name of your normalizer, and normalizer is a function that
        // should return the array of entities that are passed to it.
        // This is useful if you need more control over the order of normalizers,
        // instead of your normalizer being applied after the built in normalizers (as is the case with the normalizer option).
```

TylerBarnes:
Hey @paweljedrzejczyk , apologies for the delayed response! The holidays delayed things a bit. Thanks again for this awesome PR and for making all the requested changes. I requested/asked-about one minor additional change. I think after that's resolved this is ready to be merged!

packages/gatsby-source-wordpress/README.md
TylerBarnes: Awesome, thanks! Just merging master back into this PR to see if it fixes the  failing tests, then LGTM!

TylerBarnes:
@pieh the requested changes were made and I've tested & approved. Looks like GH wants you to approve as well before we can merge.

TylerBarnes:
Published in `gatsby-source-wordpress@3.1.57`. Thanks again @paweljedrzejczyk !

PR #18082, opened 2019-10-03
chore(gatsby-plugin-mdx): add typings

### LINKED ISSUE #15924: [gatsby-plugin-mdx] Provide TypeScript typings
## Summary

It'd be nice to have a TypeScript type at least for `MDXRenderer` which is used in code.

Screen Shot 2019-07-19 at 22 58 14

### Motivation

TypeScript is great. Supporting TypeScript is great!

thchia:
## Description
Added a type file for gatsby-plugin-mdx, typing the `MDXRenderer` component.

## Related Issues
Fixes #15924 

This relies on the presence of `@types/react` in the parent project, which should be there if the parent project is a TypeScript one.

### DIFF
--- a/packages/gatsby-plugin-mdx/index.d.ts
+++ b/packages/gatsby-plugin-mdx/index.d.ts
@@ -1,0 +1,12 @@
+import * as React from "react"
+
+export interface MDXRendererProps {
+  scope?: any
+  components?: {
+    [key: string]: React.ComponentType<any>
+  }
+  children: string
+  [propName: string]: any
+}
+
+export class MDXRenderer extends React.Component<MDXRendererProps> {}
--- a/packages/gatsby-plugin-mdx/package.json
+++ b/packages/gatsby-plugin-mdx/package.json
@@ -72,5 +72,6 @@
     "markdown",
     "remark",
     "rehype"
-  ]
+  ],
+  "types": "index.d.ts"
 }


packages/gatsby-plugin-mdx/index.d.ts
```diff
@@ -0,0 +1,43 @@
+declare module "gatsby-plugin-mdx" {
```
wardpeet: we probably want to use a typing file like
```
import * as React from "react"

export interface MDXRendererProps {
  scope?: any
  components?: {
    [key: string]: React.ComponentType<any>
  }
  children: string
}

export class MDXRenderer extends React.Component<MDXRendererProps> {}
```

thchia: @wardpeet done. Thanks 🙂

thchia:
The starter validation check failed with this error:

```
npm ERR! code ENOAUDIT
npm ERR! audit Your configured registry (https://registry.npmjs.org/) does not support audit requests, or the audit endpoint is temporarily unavailable.
```
I don't have permissions to restart the build.

packages/gatsby-plugin-mdx/index.d.ts
```diff
@@ -0,0 +1,39 @@
+import * as React from "react"
+
+// From https://mdxjs.com/advanced/typescript
+type ComponentType =
```
wardpeet: We don't need this with the suggestion below

```suggestion
```

wardpeet:
Successfully published  gatsby-plugin-mdx@1.0.51

PR #18104, opened 2019-10-04
feat(gatsby-transformer-csv): support customizing node type names

Mike-Dax:
## Description

gatsby-transformer-json and gatsby-transformer-csv aren't quite at feature parity. This PR brings -csv closer to the functionality that -json provides.

There are two major changes here, I'm happy to split them into separate PRs if desired. These changes are non-breaking, maintaining current behavior with defaults.

---

The first change is a developer customisable `typeName`.

By default the `typeName` option creates a type name based on the file.

It can be set to `from-filename` to explicitly match current behaviour, this is the default.

It can be set to `from-dir` to match the behaviour of `gatsby-transformer-json` where the typename is derived from the directory of the .csv files.

For example, a folder with contents:

```
ingredients
├── grain.csv
└── vegetables.csv
```

The graphql query would be named `allIngredientsCsv`, instead of separate queries for `allVegetablesCsv`, `allGrainCsv`.

Alternatively it can be set to a function which can take the node and parsed contents of the CSV and return a typename.

---

The second change is the `nodePerFile` option.

Instead of creating a node per line, it creates a node per csv file. It can either be `false`, maintaining current behaviour (this is the default), `true`, which creates a key called `items` which contains the contents of the csv object, or it can be a string which defines the key.

An example that combines both changes:

The config:

```javascript
{
  resolve: `gatsby-transformer-csv`,
  options: {
    typeName: () => `Foodstuffs`,
    nodePerFile: `ingredients`,
  },
},
```

A query:

```graphql
{
  allFoodstuffs {
    nodes {
      ingredients {
        ingredient
        amount
      }
      parent {
        ... on File {
          name
        }
      }
    }
  }
}
```

The result:

```json
{
  "data": {
    "allFoodstuffs": {
      "nodes": [
        {
          "parent": {
            "name": "vegetables"
          },
          "ingredients": [
            {
              "ingredient": "potato",
              "amount": 32
            },
            {
              "ingredient": "lettuce",
              "amount": 12
            }
          ]
        },
        {
          "parent": {
            "name": "grains"
          },
          "ingredients": [
            {
              "ingredient": "barley",
              "amount": 2
            },
            {
              "ingredient": "wheat",
              "amount": 42
            }
          ]
        }
      ]
    }
  }
}
```


wardpeet:
I haven't gone through the code yet, it sounds great to align them feature-wise. Could you add tests before we go foward with this one?

Mike-Dax:
@pieh Thanks for the ping. I've added tests. 

The tests and linting are passing locally for me - the errors I'm seeing seem unrelated to my changes.

I can't seem to force push my rebased (against the updated master) branch - what's the intended workflow for this kind of thing?

packages/gatsby-transformer-csv/src/gatsby-node.js
```diff
@@ -25,30 +26,67 @@ async function onCreateNode(
...
+      return _.upperFirst(_.camelCase(`${path.basename(node.dir)} Csv`))
+    } else if (
+      pluginOptions &&
+      _.isString(typeName) &&
+      typeName === `from-filename`
+    ) {
+      // be explicit
+      return _.upperFirst(_.camelCase(`${node.name} Csv`))
+    } else {
+      return _.upperFirst(_.camelCase(`${node.name} Csv`))
```
pieh: Can we do some plugin option validation and applying defaults ahead of doing any processing so we don't have extra code branches here that do similar thing?

For example on line 31 we could have:
```js
let { typeName, nodePerFile, ...options } = pluginOptions || {}
if (nodePerFile && !_.isString(nodePerFile)) {
  // if nodePerFile is truthy, but not a string - let's set default "items"
  nodePerFile = 'items' 
}

if (!typeName) {
  // if typeName is not set, use default:
  typeName = `from-filename`
}
```

Mike-Dax:
@pieh I've replaced the string options with exported functions.

I've also included a string option where the type is just the literal string provided.

packages/gatsby-transformer-csv/src/gatsby-node.js
```diff
@@ -25,33 +32,62 @@ async function onCreateNode(
...
+        transformObject(obj, i)
+      })
+    }
   }
 
   return
 }
 
+exports.typeNameFromDir = typeNameFromDir
+exports.typeNameFromFile = typeNameFromFile
```
pieh: I think we need to move those functions and exports to `index.js` file:
 - this is default entrypoint for packages (otherwise ``const { typeNameFromDir } = require(`gatsby-transformer-csv`)`` probably won't work
 - gatsby will warn (or error out) when gatsby-node export names doesn't match gatsby-node API extension points

Mike-Dax:
> Looks almost ready - there's one potential issue with `typeNameFromDir` and `typeNameFromFile` exported functions (see inline comment about it for details).
> 
> I'm not sure why there are access permissions changed on `gatsby-node` file (`100644 → 100755`) - this shouldn't be needed

Sure, and yeah that's very strange, I've explicitly set them back.

Mike-Dax:
Awesome, great to hear. Thank you for your help!

PR #18112, opened 2019-10-04
fix(gatsby): catch more browser only variable for SSR error

igogo5yo:
##  Description

This is the improvement of error handling for errors related to not defined References during SSR.

## Related Issues

I haven't created any issues, but the related error is easy reproduces if you will try to run `gatsby build` and your code will have some browser-related objects like `navigator`, `localStorage`, etc.


packages/gatsby/src/commands/build.js
```diff
@@ -150,16 +150,20 @@ module.exports = async function build(program: BuildArgs) {
     })
   } catch (err) {
     let id = `95313` // TODO: verify error IDs exist
-    if (err.message === `ReferenceError: window is not defined`) {
+    const context = {
+      errorPath: err.context && err.context.path,
+    }
+
+    const match = err.message.match(/ReferenceError: (\w+) is not defined/)
```
pieh: I'm bit hesitant on using `\w+` here (as it will catch any variables and not just browser-only ones - should we have list of potential variables (`(browser|navigator|localStorage)` etc) here instead maybe?

pieh:
We will also need to update jest snapshot - you can run `yarn jest packages/gatsby-cli/src/reporter/__tests__/index.js -u` in root of the repo, and commit snapshot change

packages/gatsby/src/commands/build.js
igogo5yo: @pieh good idea! I will fix it.

igogo5yo: @pieh I suggest to add only a most common list `window|document|localStorage|navigator|alert|location`, because browser global scope contains a lot more variables.

pieh: Sounds perfect! We can always add more browser-only global variables here in future

pieh:
All right, awesome! I'll just master in to fix CI check failure (unrelated to this PR), and will merge this in.

Issue #17921, opened 2019-09-26
blog: except shows unescaped `\$15' instead of `$15'


muescha:
## Description

Blog except shows unescaped `\$15' instead of `$15'

### Steps to reproduce

open https://www.gatsbyjs.org/blog/

### Expected result

normal text: `15$`

### Actual result

![Bildschirmfoto 2019-09-26 um 20 48 30](https://user-images.githubusercontent.com/184316/65716440-704c9600-e09f-11e9-8295-52bba48faa3a.png)

### Environment

n/a

sidharthachatterjee:
This was an upstream issue in MDX that caused prettier to incorrectly escape blog: `$15' to `\$15' 

This is fixed in upstream so I'll close this once `www` is updated visa Renovate

gillkyle:
I just updated the MDX version to verify this happened since it doesn't seem like renovate is actually making those changes, however I'm seeing the deployed version of the blog doesn't seem to have the backslashes anyway 🤔 hm.

PR #18115, opened 2019-10-04
www(fix): upgrade mdx to fix prepended backslashes

gillkyle:
## Description

Prepended backslashes have been an issue in the site for awhile, this got fixed in MDX here: https://github.com/mdx-js/mdx/pull/740

The issue keeps getting reported and it turns out renovate bot hasn't been bumping dependencies in `www` so I just updated this real quick myself to be sure to do away with this pesky problem once and for all 🎉 

## Related Issues

Related to #17921 and probably like 50 others I'm too lazy to go find.


fk:
🙏 YAAAS

PR #18123, opened 2019-10-04
feat(gatsby-plugin-offline): Allow configuring Workbox debug mode

### LINKED ISSUE #17782: Option to enable workbox debug output in gatsby-module-offline
## Summary

Expose option to enable workbox debug mode in `gatsby-plugin-offline`.

### Basic example

The basic objective would be to add `debug: true`  when running `workbox.setConfig()`

```javascript
workbox.setConfig({modulePathPrefix: "workbox-v4.3.1", debug:true});
```

The hacky way that I enable it now is to actually edit `sw.js` in the `public` directory before deployment. This task might be automated using `sed`, depending on the outcome of this feature request .

I am not sure where the ideal place to do this would be, gatsby-plugin-offline seems to only run in actual builds from `gatsby build`. Does `gatsby build` have some notion of a debug build? It might possibly be enabled with the `--no-uglify` flag, but that seems rather presumptuous. 

An option to enable it in gatsby-config would be useful to me... it feels a bit ephemeral to exist there.

### Motivation

This is the only place to enable debugging for workbox, which provides incredibly detailed and useful logging for users of `appendScript` or `workboxConfig`. It does not seem possible within an appended script because `setConfig` needs to be used before any modules are accessed, near the top of `sw.js`.

davidbailey00:
## Description

Add new `debug` option to configure Workbox's debugging logs

## Related Issues

Fixes #17782


wardpeet:
Oh my, I didn't test it but it seems like we do lots of hacky stuff with workbox so i'm okay with approving after this is working. I wonder if we should switch to https://developers.google.com/web/tools/workbox/modules/workbox-build#injectmanifest instead of generateSw

davidbailey00:
The reason we use `generateSW` rather than `injectManifest` is because the latter appends to a file, whereas we need to have the generated config at the top followed by our own options in order to have access to the Workbox global variables.

Also I wish there were a less hacky way of doing this, but Workbox doesn't allow configuring this option when using either `generateSW` or `injectManifest` :(

PR #18130, opened 2019-10-04
Add allowFullScreen attribute to iframe

### LINKED ISSUE #18127: [www] add allowFullScreen to EggheadEmbed component
## Description

From feedback received in the feedback widget of the docs page, it mentioned that videos can't be opened fullscreen which is definitely a behavior we'd like to support, especially because the bitmovin video player that the egghead embed component uses has a button for it.

The `allowFullScreen` attribute is likely enough to make this behavior possible.

One page where you can see an Egghead video with the embed is on the [Quick Start](https://www.gatsbyjs.org/docs/quick-start/) page.

## What needs to be done

- [ ] Add the `allowFullScreen` attribute to the iframe in this component: https://github.com/gatsbyjs/gatsby/blob/master/www/src/components/shared/egghead-embed.js#L6-L12
- [ ] Verify that the above fix works

sagarnasit:
## Description

Added `allowFullScreen` attribute in iframe.

## Related Issues
 closes #18127

## What needs to be done

- [x] Add the `allowFullScreen` attribute to the iframe in this component: https://github.com/gatsbyjs/gatsby/blob/master/www/src/components/shared/egghead-embed.js#L6-L12
- [x] Verify that the above fix works


sagarnasit:
@gillkyle I have added attributed and tested. It is working as expected and allows a user to see iframe video on full screen.

PR #18131, opened 2019-10-04
Add allowFullScreen attribute to the iframe.

### LINKED ISSUE #18127: [www] add allowFullScreen to EggheadEmbed component

ianbaxter:
## Description

Added allowFullScreen attribute to the iframe to enable fullscreen behavior

## Related Issues

Fixes #18127

### DIFF
--- a/www/src/components/shared/egghead-embed.js
+++ b/www/src/components/shared/egghead-embed.js
@@ -9,6 +9,7 @@ const EggheadEmbed = ({ lessonLink, lessonTitle }) => (
       height={348}
       src={`${lessonLink}/embed`}
       title={`Video: ${lessonTitle}`}
+      allowfullscreen="allowfullscreen"
     />
 
     <p>


gillkyle:
Hey @ianbaxter, since @sagarnasit opened a PR for this issue just before you did I merged in his PR so I'm afraid I'll just have to close this now. 

There will definitely be other issues and I'll try and write some up with similar detail so you can find more low hanging fruit to contribute with, we'd love to have you as a contributor to the repository 🙂 sorry this had several people jump on it so quickly.

PR #18133, opened 2019-10-04
Add run-static-queries cli command

ascorbic:
This is a bit of a speculative PR, because I'm not sure if this is a useful addition to the CLI, and if my approahc is good. I was experimenting with ways to get Storybooks working with static queries and found a solution, and thought it might help others.

Static queries make it hard to use components outside of a full Gatsby site, because the component doesn't have access to the data from the query. Mocking can solve this for unit tests, but there are still situations where it is useful to be able to use a component with a static query outside of a site. I ran into this when trying to run Storybooks. If I didn't use the Babel plugin to remove queries then it would fail when it tried to call `graphql`, but if I did include it then there would be no data unless I had built the site. This was fine for local development, but I like to build Storybooks in CI and it's not realistic to build the whole site.

This PR adds a new CLI command: `gatsby run-static-queries`. This runs the static queries in the site, and generates the JSON files in `public/static`. This is a lot faster than doing a full site build, and can be run from CI before building Storybooks.

I've not written any docs on this, but if people think this will be a useful addition to the CLI then I can write docs.

### DIFF
--- a/packages/gatsby-cli/src/create-cli.js
+++ b/packages/gatsby-cli/src/create-cli.js
@@ -259,6 +259,15 @@ function buildLocalCommands(cli, isLocalSite) {
   })
 
   cli.command({
+    command: `run-static-queries`,
+    desc: `Run static queries and generate JSON files`,
+    handler: getCommandHandler(`run-static-queries`, (args, cmd) => {
+      process.env.NODE_ENV = process.env.NODE_ENV || `development`
+      return cmd(args)
+    }),
+  })
+
+  cli.command({
     command: `repl`,
     desc: `Get a node repl with context of Gatsby environment, see (https://www.gatsbyjs.org/docs/gatsby-repl/)`,
     handler: getCommandHandler(`repl`, (args, cmd) => {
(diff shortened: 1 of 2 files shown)


pieh:
Hey @ascorbic! We won't be merging this PR, I wish you created issue first to discuss this (please do that next time). This seems like half measure - it will handle static queries, but not page queries - where do we draw the line here?

As far as I know storybook is tool "for developing UI components in isolation" (copied from their website) and (to me) it seems like trying to use production data there violates the purpose of it?

I would much prefer better solution for mocking out query results (for unit tests and "stories" in storybook-like tools), so it's not copy & pasting instead of this command.

That being said - if you still want to use flow like this - you can publish your command separately on your own (change imports to grab from `gatsby/bootstrap` etc), install it as dev dependency and set up scripts to run it before using storybook.

ascorbic:
No problem. As I said, it was a bit speculative and was mostly the result of poking around in the core to see what could be done so I'm not too worried about the time I spent on it! I decided against doing page queries because by that point you might as well build the whole thing. You're right that best of all would be a way to handle graphql queries in isolation.

PR #18139, opened 2019-10-04
chore(gastby-plugin-sharp): Use createContentDigest helper

jbutz:
## Description

Updates the plugin to use the helper instead of manually handling the crypto work.

## Related Issues

Related to #8805


jbutz:
@wardpeet, even better! I will try and get a commit pushed up to shift to using that ASAP

wardpeet:
@jbutz awesome work! Thanks for using the new code snippet. Let's merge when everything is green! :tada:

Issue #15019, opened 2019-06-21
Error messages - contribute to improvements!


shannonbux:
# The Great Error Message Voyage
## Summary
Great error messages are part of helping users recognize and recover from errors. This project summarizes how to make a list of errors and then improve the errors.

## 1. How to make a complete list of errors
A) When you find an error that is not yet on the list below, create an issue using [this template](https://docs.google.com/document/d/14e_z79hk6NmDYTMnveplM4uj-k_KzPQNNRm7U8Jlq4o/edit?usp=sharing). 
B) Add your issue to the "List of errors" below and link to it.

### List of errors
- [x] [insert link to issue describing the error here] [insert category here - GraphQL, Webpack, or "unknown"] [insert grade here]

## 2. How to improve errors
A) Choose an error from the checklist above
B) Look at the grade for the error
C) Open a PR (linked to the corresponding issue) that improves one error message by at least one grade level (mention this issue #15019 in your issue). Take an error from 1 to 2, for example.

## Example
Here is an example of an error message that was taken from level 1 to level 3.
<img width="719" alt="unknown-fragment-before" src="https://user-images.githubusercontent.com/32467162/59928542-44d26380-943f-11e9-834c-08aac5d8422b.png">

<img width="403" alt="unknown-fragment-after" src="https://user-images.githubusercontent.com/32467162/59928564-4ef46200-943f-11e9-831d-bf173d0a7e6a.png">

[insert issue describing a recently improved error]
[insert PR that resolved the issue above. PR should show a screenshot of what the new error message looks like in the CLI]

## Motivation
The faster that users can recognize, diagnose, and recover from an error, the faster users can build a super awesome, fast Gatsby project!

## Key for Grades

This is what the grades mean:
0: Silent error
1: An error is thrown, and the message is missing one or more of the following: id, category, name, description, prompt for fix.
2: Error message with id, category, name, and description
3: Error message with id, category, name, and description + prompt for fix (can be URL to docs page, code example, example site, suggestion, etc) 
4: Error message with id, category, name, and description + prompt for fix + adaptive prompting

nwthomas:
I want to be involved with this as I'm just coming off writing custom thrown errors for [React-Conflux](https://github.com/dustinmyers/react-conflux). 

As this is my first Gatsby issue, I'm willing to do whatever is needed. I'd love to take on some of the first errors posted here and get my feet wet.

chrismgonzalez:
[...]

PR #18142, opened 2019-10-04
chore: Convert local SSL certificate errors into structured logging

izontm:
## Description
This PR makes structured errors around local SSL certificate, including links to documentation.

Original:
image

New:
image

## Related Issues

Related to #15019


packages/gatsby/src/utils/get-ssl-cert.js
```diff
@@ -15,9 +15,12 @@ module.exports = async ({ name, certFile, keyFile, directory }) => {
   // check that cert file and key file are both true or both false, if they are both
   // false, it defaults to the automatic ssl
   if (certFile ? !keyFile : keyFile) {
-    report.panic(
-      `for custom ssl --https, --cert-file, and --key-file must be used together`
-    )
+    report.panic({
+      id: `11521`,
+      context: {
```
wardpeet: Is this bit necesary? maybe empty object or just remove it as we're not using it in the message.

Wdyt?

ANDIKKY:
This would allow for an instance of gatsby develop to be served from behind a proxy server while allowing the socket connection to go directly back to the instance instead of through the proxy...

packages/gatsby/src/utils/get-ssl-cert.js
izontm: I had tried without passing context, but apparently the message is needed for the gatsby-telemetry package.  I couldn't solve it any other way.  If you have any idea how to overcome this, I can try to solve it.  Thanks.

pieh: Let's use empty `context: {}` - this should work ok. But this is separate issue, and we should fix this in telemetry package to not crash on undefined context

PR #18146, opened 2019-10-04
chore(showcase): Add blog.kwst.site

SatoshiKawabata:
## Description

Add https://blog.kwst.site into showcase.


docs/sites.yml
```diff
@@ -7530,3 +7530,14 @@
   categories:
     - Education
     - Learning
+- title: blog.kwst.site
+  main_url: https://blog.kwst.site
+  url: https://blog.kwst.site
+  description: 大名ではたらくフロントエンジニアのブログ
```
wardpeet: Could you add an English translation to this description as well? We want to make sure the showcase is accessible for "all" users.

SatoshiKawabata: OK. I understand. I'll translate it. Thanks.

SatoshiKawabata:
I fixed it.

PR #18200, opened 2019-10-05
chore(memory): very minor memory churn improvements

me4502:
## Description

I've been dealing with a few overly-active GC events and did a bit of an investigation throughout Gatsby to just hit a few of the more minor locations that don't show up too much on a profiler, but over time add up. The main thing I looked for was `reduce` statements that created a new array each time, however, I believe a vast majority of the ones that can make an impact have already been fixed.

The change to the query queue feels like it's made a small impact, however it is hard to measure.  The query compiler most likely does not at this stage, but I feel will improve things once the temporary code immediately following it is removed.


wardpeet:
Published in gatsby@2.15.33

Issue #17902, opened 2019-09-26
using useStaticQuery hooks but can't find any documentation how test components with GraphQL 


HUFGhani:
i've looked at website docs "https://www.gatsbyjs.org/docs/testing-components-with-graphql/" but there is only have reference to StaticQuery but not useStaticQuery can the docs be update

simrankaurmann:
Would it be sufficient to edit the above page to include a useStaticQuery mock implementation example?

gillkyle:
That would be great @simrankaurmann! A section similar to the section titled "Testing StaticQuery", maybe called "Testing useStaticQuery" would be welcome for this! Feel free to reference what's already been done for this guide and add a new section, [the guides on using useStaticQuery](https://www.gatsbyjs.org/docs/use-static-query/#how-to-use-usestaticquery-in-components) might be helpful in finding some code to use or showing you how that testing would happen! If you'd like to take it on we can label it Hacktoberfest - Claimed 🙂

imthinhvu:
@gillkyle - since I didn't yet see this issue being officially claimed, I ended up making a PR for this here: https://github.com/gatsbyjs/gatsby/pull/18202

@simrankaurmann - Feel free to add onto my PR if you have any other ideas. Since you had originally requested this, I also don't mind closing my PR and finding something else to work on. I was mostly going through the open Hacktoberfest issues that weren't yet marked as claimed.

Either way, would be great to get both of your reviews on it.

PR #18202, opened 2019-10-05
docs: add messaging about useStaticQuery in existing GraphQL testing doc

imthinhvu:
## Description

Adds messaging for testing `useStaticQuery` based off existing GraphQL testing docs. Specifically mentions how to rewrite the `Header` component with `useStaticQuery` hook, and how keeping components pure also aid with converting previous tests to work with Hooks.

## Related Issues
Addresses https://github.com/gatsbyjs/gatsby/issues/17902


docs/docs/testing-components-with-graphql.md
```diff
@@ -249,10 +249,45 @@ export const Header = props => (
...
+
+export const PureHeader = ({ data }) => (
+  <header>
+    <h1>{data.site.siteMetadata.title}</h1>
+  </header>
+)
+
+export const Header = props => (
+  const data = useStaticQuery(graphql`
+    query HeaderQuery {
```
muescha: `HeaderQuery` is new in this change - we need this here? 
if yes: so then it should be mentioned in the text 
if no: then we should leave it off to avoid confusion that this is mandatory

imthinhvu: Thanks! I removed it.

PR #18218, opened 2019-10-06
fix(www): Add space between prev and next article links

lekterable:
## Description

In some cases when lines of text of a prev and next article take the full width of their containers there is no space between them so it's hard to tell that they are actually two different elements [Like here](https://www.gatsbyjs.org/blog/2018-08-09-swag-store/#reach-skip-nav). I've added some space to make sure that it won't happen anymore.

Before:
Zrzut ekranu 2019-10-6 o 13 39 58

After:
Zrzut ekranu 2019-10-6 o 13 39 50

### DIFF
--- a/www/src/components/prev-and-next.js
+++ b/www/src/components/prev-and-next.js
@@ -42,12 +42,13 @@ const PrevAndNext = ({ prev = null, next = null, ...props }) => {
       css={{
         [mediaQueries.sm]: {
           display: `flex`,
+          justifyContent: `space-between`,
           width: `100%`,
         },
       }}
       {...props}
     >
-      <div css={{ [mediaQueries.sm]: { width: `50%` } }}>
+      <div css={{ [mediaQueries.sm]: { width: `48%` } }}>
         {prev && (
           <Link to={prev.link} css={prevNextLinkStyles}>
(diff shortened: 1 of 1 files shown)


fk:
Oh man 🙄… thank you @lekterable! 🙏 🤗

lekterable:
hey @wardpeet , I've noticed that you've changed the title before merge to follow the convention `type(scope?): subject`, how about I configure commitlint to do it? do you think it's a good idea?

wardpeet:
Hi @lekterable,

commitlint will check every commit and will make it harder for newcomers to contribute. So we want to put the burden on us instead of the contributors. Sometimes we forget to do it but that's not a big deal.

Thanks for offering the help!

lekterable:
hmm, that's a fair point, how about a small note in `How to contribute` that this is the preferred convention? I think that would give you less work as the more experienced people usually read it to make sure that they do everything according to the guidelines, what do you think @wardpeet ?

wardpeet:
That sounds like a good idea, maybe it would best to put it in https://www.gatsbyjs.org/contributing/how-to-open-a-pull-request/

lekterable:
awesome, I'm on it!

Issue #1966, opened 2017-08-30
Warning: You cannot change <Router history> in browser.js/HMR stops working when changing layout, template or page


fk:
When picking up work on #1765 (www) today I merged the current master. After that I started noticing that in `gatsby develop` whenever I changed a layout, template or page, HMR stopped working (until I reload the page). Console says `Warning: You cannot change <Router history>`:

![image](https://user-images.githubusercontent.com/21834/29851708-f4022a38-8d35-11e7-90eb-00815a3375d7.png)

I can reproduce this in www of the current master, too (Gatsby v1.9.10). Also tried v1.9.14 and run into the same. Searched and found #1913 and #1936 … confused now! ;-) `¯\_(ツ)_/¯`

fk:
Seems I'm not the only one seeing this. @3axap4eHko not sure if this is related to your PR, but maybe you can take a look?

jamesslock:
I am also having the same issue.

stolinski:
+1

fk:
@KyleAMathews Thank you! 🎉

KyleAMathews:
Sorry for the slow fix here — gatsby 1.9.31 has the fix from https://github.com/gatsbyjs/gatsby/pull/2110

PR #18234, opened 2019-10-06
chore(gatsby-plugin-preact): update docs after Preact X release

vitordino:
## Description

Preact X has released! And as I we’ve already had this plugin dependency updated (on #18057) i imagine it would be nice to update our docs too 👐

### DIFF
--- a/packages/gatsby-plugin-preact/README.md
+++ b/packages/gatsby-plugin-preact/README.md
@@ -6,11 +6,11 @@ While Preact doesn't provide full support for the React ecosystem, it is an
 intriguing option for Gatsby sites as it saves ~30kb of JavaScript vs. using
 React.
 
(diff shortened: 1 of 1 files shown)


gustavpursche:
@vitordino I was using the last preact beta and had to realize that `gatsby-link` is not compatible with preact (the URL was updated, but the new content isn't loaded). In case this problem still exists, do you think it would make sense to mention it here?

vitordino:
hi @gustavpursche doesn’t your problem isn’t the same as [**preact**’s `#1966`](https://github.com/preactjs/preact/issues/1966)? if so, it’s probably fixed, and you just have to pin the stable `10.0.0` version

gustavpursche:
✨ 

> Yes but which version of Preact X, that bug was present in rc.2 and rc.3 but should be resolved in the final.

Ohlala. Thanks for the pointer, seems to be fixed!

wardpeet:
published in gatsby-plugin-preact@3.1.15

PR #18236, opened 2019-10-06
chore: Remove unused parameters

DanielRuf:
## Description

This PR removes unused parameters.

### DIFF
--- a/benchmarks/plugin-manifest/index.js
+++ b/benchmarks/plugin-manifest/index.js
@@ -46,7 +46,7 @@ async function runTest() {
 
   for (let i = 0; i < rounds; i++) {
     // process.stdout.write(`Round ${i + 1}: `)
-    await executeBootstrap(i)
+    await executeBootstrap()
   }
 
   let averageSum = [0, 0]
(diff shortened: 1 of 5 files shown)


DanielRuf:
> Did you use any tool to find those unused parameters? If so, maybe we could have it run as part of CI checks to keep the code clean?

Yes, I have used SonarCloud. You can easily add it to a GitHub project and it gives feedback on pushes and PRs.

See https://sonarcloud.io/dashboard?id=DanielRuf_gatsby

DanielRuf:
I would highly recommend adding / using SonarCloud as additional check =)

See also https://github.com/DanielRuf/sonarcloud-project/pull/1 and https://github.com/DanielRuf/talks/blob/master/08_codequality_sonarsource_phpugmrn/slides.pdf

pieh:
This looks very interesting. Do you know if we would add check right now, if it would continuously fail because of all the already existing problems (until we fix them)? Or would it just pick up problems introduced with changes in PR?

I don't want to add check that would fail on all new PRs, because this will be very confusing for people, that it fails on unrelated code paths

DanielRuf:
> This looks very interesting. Do you know if we would add check right now, if it would continuously fail because of all the already existing problems (until we fix them)? Or would it just pick up problems introduced with changes in PR?

We can do both. Do a first manual check using `sonar-scanner` and the PR check.
The GitHub integration / part only checks new changes normally.

In general we can test this on a fork.

Issue #17703, opened 2019-09-18
[docs] [workflows] Improve docs for "Working with fonts and typography"


nicholaspung:
Part of the **Top 25 Learning Workflows initiative**. See  [#13708](https://github.com/gatsbyjs/gatsby/issues/13708) for the meta issue that this issue falls under.

# User story
As a new Gatsby user, I want to work with fonts and typography in my Gatsby project.

# Evaluation
Search | Discover | Complete | Linked | Tone | Style | Overall
-- | -- | -- | -- | -- | -- | --
😄 | 😄 | 😐 | 😄 | 😄 | 😐 | 😐

# Steps taken to implement
1. Searchability
    - Gatsby fonts and typography
        1. result: https://www.gatsbyjs.org/docs/typography-js/
    - Gatsby fonts
        1. result: https://www.gatsbyjs.org/packages/gatsby-plugin-web-font-loader/
    - Gatsby typography
        1. result: https://www.gatsbyjs.org/docs/typography-js/ 
        2. result: https://www.gatsbyjs.org/packages/gatsby-plugin-typography/ 
        3. result: https://www.gatsbyjs.org/guidelines/typography/
    - Gatsby change font
        1. result: https://www.gatsbyjs.org/docs/typography-js/ 
        2. result: https://www.gatsbyjs.org/packages/gatsby-plugin-web-font-loader/
    - Gatsby change typography
        1. result: https://www.gatsbyjs.org/docs/typography-js/

2. Discoverability
    - Search bar:
        - fonts: 
            1. result: https://www.gatsbyjs.org/docs/recipes/#adding-a-local-font
                - only recipes for fonts shows up
        - typography
            1. result: https://www.gatsbyjs.org/docs/typography-js/
                - all results go to typography-js docs
        - fonts and typography
            1. results: https://www.gatsbyjs.org/docs/typography-js/
            2. blogs: 
                - 1 [article](https://www.gatsbyjs.org/blog/2019-08-23-creating-a-purpose-driven-media-platform/#building-a-typographic-system-from-scratch) not related to fonts and typography
                - 1 [article](https://www.gatsbyjs.org/docs/importing-assets-into-files/#reach-skip-nav) links to importing images, fonts, and files
                - no article on using CSS modules/CSS in JS

    - Clickpath on .org:
        - docs -> recipes -> styling with css -> adding a local font/ using google fonts -> scroll down to additional resources - typography.js 
[...]

PR #18239, opened 2019-10-07
docs: add links to Typography.js docs and tweaks for understanding

nicholaspung:
## Description

Simplified instructions to Grade 8 level, added instructions to remove previous font/css, highlighted code changed, changed a lot of wording

## Related Issues

Addresses one of the recommendations I made in #17703


nicholaspung:
Hey @gillkyle , can I have some help with this check? I looked through CircleCI to see what was failing, and for some reason it's failing on a test file on the string "gitlab" vs "github". I'm not sure how to solve it since I haven't touched the test file and not too sure about the gitlab vs github situation.

Unless the solution is to change the test file to be "github" instead of "gitlab". I'm really at a loss...

gillkyle:
Hey @nicholaspung, that test failing doesn't look related to anything you've changed so don't worry about it. I'll trigger the test suite again and see if they pass (which I'm guessing they will.. 🤞). If they keep having issues I'll look into it more.

docs/docs/typography-js.md
```diff
@@ -64,32 +60,38 @@ const typography = new Typography({
...
-To use the theme, edit the `typography.js` file that you created before and inform typography of the new configuration:
+To use the theme, edit the `typography.js` file that you created before and inform Typography of the new configuration:
 
 ```diff:title=src/utils/typography.js
 import Typography from "typography";
+// highlight-start
 + import funstonTheme from 'typography-theme-funston'
+// highlight-end
 const typography = new Typography(
+// highlight-start
```
gillkyle: It looks like this could go on the line before `funstonTheme` on line 86? Since the other configurations here like baseFontSize were included previously?

nicholaspung: I wanted to highlight all the code changes, additions and subtractions. Is it more intuitive if we only highlight the additions?

nicholaspung: Yup, that makes sense! Just made the changes. We could always change it later if it turns out people aren't understanding the highlights for code additions/subtractions.

Let me know if there's anything else of concern!

Issue #18135, opened 2019-10-04
gatsby-remark-embed-video update causes gatsby-plugin-mdx to throw sourceNodes lifecycle error


janosh:
## Description

Trying `gatsby develop` in [this repo](https://github.com/janosh/janosh.io) fails with 

```
ERROR #11321  PLUGIN

"gatsby-plugin-mdx" threw an error while running the sourceNodes lifecycle:

Cannot destructure property `beginMarker` of 'undefined' or 'null'.

  - index.js:43 Function.setParserPlugins
    [janosh.io]/[gatsby-remark-embed-video]/dist/index.js:43:27
  
  - source-nodes.js:92 Object.module.exports [as sourceNodes]
    [janosh.io]/[gatsby-plugin-mdx]/gatsby/source-nodes.js:92:47
```

### Environment

```
System:
  OS: macOS 10.14.6
  CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
  Shell: 5.3 - /bin/zsh
Binaries:
  Node: 12.11.1 - /usr/local/bin/node
  Yarn: 1.19.0 - /usr/local/bin/yarn
  npm: 6.11.3 - /usr/local/bin/npm
Languages:
  Python: 2.7.10 - /usr/bin/python
Browsers:
  Chrome: 76.0.3809.87
  Firefox: 68.0.2
  Safari: 13.0.1
npmPackages:
  gatsby: 2.15.29 => 2.15.29 
  gatsby-image: 2.2.24 => 2.2.24 
  gatsby-plugin-algolia: ^0.3.4 => 0.3.4 
  gatsby-plugin-catch-links: 2.1.12 => 2.1.12 
  gatsby-plugin-favicon: ^3.1.6 => 3.1.6 
  gatsby-plugin-google-analytics: 2.1.20 => 2.1.20 
  gatsby-plugin-lodash: ^3.1.10 => 3.1.10 
  gatsby-plugin-manifest: 2.2.20 => 2.2.20 
  gatsby-plugin-mdx: ^1.0.47 => 1.0.47 
  gatsby-plugin-netlify-cache: ^1.2.0 => 1.2.0 
  gatsby-plugin-offline: 3.0.11 => 3.0.11 
  gatsby-plugin-react-helmet: 3.1.10 => 3.1.10 
  gatsby-plugin-sharp: 2.2.28 => 2.2.28 
[...]

PR #18243, opened 2019-10-07
fix(gatsby-plugin-mdx): pass the correct options to setParserPlugins

d4rekanguok:
## Description

Hey folks!

In `gatsby-plugin-mdx`, when `setParserPlugins` is called, it was passed in the wrong options<sup> [source](https://github.com/gatsbyjs/gatsby/blob/561d33e2e491d3971cb2a404eec9705a5a493602/packages/gatsby-plugin-mdx/gatsby/source-nodes.js#L93)</sup>:

```
for (let plugin of options.gatsbyRemarkPlugins) {
  ...
    for (let parserPlugin of requiredPlugin.setParserPlugins(
        plugin.pluginOptions   // <--- undefined
      )) { ... }
```

Instead of `plugin.pluginOptions`, it should have been `plugin.options`.

## Reproduction of bug

https://github.com/d4rekanguok/gatsby-bug-repro-18243


pieh:
Seems like this is pretty much copy & paste from remark plugin - https://github.com/gatsbyjs/gatsby/blob/68dbb6d8eaeeb07613a810d3f96cd7ec3636e2cb/packages/gatsby-transformer-remark/src/extend-node-type.js#L128-L142

Which means if that's a bug we have it in both mdx and remark.

Do you have any reproduction so we could verify this change quickly?

d4rekanguok:
Hi @pieh! I actually discovered this while debugging a remark plugin that worked with gatsby-transformer-remark but not gatsby-plugin-mdx, so I think the issue only exists with mdx.

I can quickly make & share a reproduction repo in an hour or so

d4rekanguok:
@pieh I've just put this repro together: 

https://github.com/d4rekanguok/gatsby-bug-repro-18243

The step to reproduce is detailed in the readme.

I think it might have something to do with the fact that remark plugins for mdx are passed in via options as `gatsbyRemarkPlugins`, while with remark it's `plugins`. I'm guessing Gatsby might have a recursive system that store plugin options as `pluginOptions`, but only recognize plugins set via the `plugins` attribute?

pieh:
> think it might have something to do with the fact that remark plugins for mdx are passed in via options as `gatsbyRemarkPlugins`, while with remark it's `plugins`. I'm guessing Gatsby might have a recursive system that store plugin options as `pluginOptions`, but only recognize plugins set via the `plugins` attribute?

Oh yeah, that would explain the difference.

You are correct - `plugins` field in `options` does have special handling, and we do some transformation, so if `gatsbyRemarkPlugins` is used, this transformation won't be done.

d4rekanguok:
Oh in that case, instead of fixing this like I did in this PR, maybe it's better to also handle `gatsbyRemarkPlugins` specially to keep the plugin chain going?

This might be a really edge case however, since gatsby-plugin-mdx has been released for a long time & no one has yet filed a bug on this

pieh:
I think this is fine. Other subplugin API use `options` instead of `pluginOptions` in mdx, so I think this is fine for now.

mdx (using `plugin.options`):
https://github.com/gatsbyjs/gatsby/blob/b4e3c5eb0192832403beebad5a44db08efbfa0ca/packages/gatsby-plugin-mdx/utils/get-source-plugins-as-remark-plugins.js#L53-L73


remark (using `plugin.pluginOptions`):

https://github.com/gatsbyjs/gatsby/blob/b4e3c5eb0192832403beebad5a44db08efbfa0ca/packages/gatsby-transformer-remark/src/extend-node-type.js#L216-L235

packages/gatsby-plugin-mdx/gatsby/source-nodes.js
```diff
@@ -90,7 +90,7 @@ module.exports = (
     debug(`required`, plugin)
     if (_.isFunction(requiredPlugin.setParserPlugins)) {
       for (let parserPlugin of requiredPlugin.setParserPlugins(
-        plugin.pluginOptions
+        plugin.options
```
pieh: Let's ensure it's not undefined so plugins don't have to check that
```suggestion
        plugin.options || {} 
```
( Same as in https://github.com/gatsbyjs/gatsby/blob/b4e3c5eb0192832403beebad5a44db08efbfa0ca/packages/gatsby-plugin-mdx/utils/get-source-plugins-as-remark-plugins.js#L67 )

d4rekanguok: Good call, I've actually just run into this issue a few days ago https://github.com/gatsbyjs/gatsby/issues/18135#event-2688457725

d4rekanguok:
A few of my PRs is failing this unit_tests_node12:

> FAIL  peril/tests/validate-yaml/validate-yaml-starters.test.ts
>  ● a new PR › Doesn't allow non GitHub repos

I'm not submitting new starters though, so I'm not sure why it's failing... any ideas? @pieh

pieh:
> I'm not submitting new starters though, so I'm not sure why it's failing... any ideas? @pieh

Oh yeah, this has something with recent Node 12 release (I think) - we "fixed" this (or rather workarounded it) in https://github.com/gatsbyjs/gatsby/pull/18247, so I'll just merge master in this PR and that should fix that failing test

pieh:
Published `gatsby-plugin-mdx@1.0.49`

PR #18266, opened 2019-10-07
fix(gatsby-remark-responsive-iframe): content is wrapped in <undefined> block when used with gatsby-plugin-mdx

d4rekanguok:
## Description

When used with gatsby-plugin-mdx, the html generated by this plugin is wrapped in a `<undefined>` block:

```html
<undefined>
    <div class="gatsby-resp-iframe-wrapper" style="padding-bottom: 75%; position: relative; height: 0px; overflow: hidden; margin-bottom: 1.0725rem;">
      <iframe src="asdads" style="position: absolute; top: 0px; left: 0px; width: 100%; height: 100%;"></iframe>
    </div>
</undefined>
```

## Cause

The html generated by this plugin has a bit of extra space at the end <sup>[source](https://github.com/gatsbyjs/gatsby/blob/57e5d69c17abfed74fa05c069a7467107e1aadfc/packages/gatsby-remark-responsive-iframe/src/index.js#L75)</sup>:

```js
          const rawHTML = `
          <div ...>
            ${newIframe}
          </div>
          ` // <-- extra space with new line
```

When passed via `hast-util-raw` in mdxjs/mdx <sup>[source](https://github.com/mdx-js/mdx/blob/4ee999cbe3e82bb4a56e43df14151835f14da400/packages/mdx/index.js#L67)</sup>, this resulted in a tree that looks like this:

```js
{ type: 'root',
  children:
   [ { type: 'element',
       tagName: 'div',
       properties: [Object],
       children: [Array] },
     { type: 'text', value: '\n          ' } ], // <----- extra space is now turned into a child of type 'text'
  data: { quirksMode: false } }
```

Unfortunately, `hast-util-raw` can't unwrap (remove root) nodes with more than 1 child <sup>[source](https://github.com/syntax-tree/hast-util-raw/blob/aba08a50352998f770de7656c7ca5138a2883c64/index.js#L47)</sup>.

```js
  // Unpack if possible and when not given a `root`.
  if (tree.type !== 'root' && result.children.length === 1) { // <-- here, result.children.length is 2
    return result.children[0]
  }
```

Because of this, the `root` node was not removed, and thus cause `const { tagName } = raw(node)` in mdx-js/mdx to be `undefined`, which made it to the rendered html. 

## Related Issues

https://github.com/borgfriend/gatsby-remark-embed-video/issues/20

### DIFF
--- a/packages/gatsby-remark-responsive-iframe/src/index.js
+++ b/packages/gatsby-remark-responsive-iframe/src/index.js
@@ -71,8 +71,7 @@ module.exports = ({ markdownAST }, pluginOptions = {}) =>
           }"
           >
             ${newIframe}
-          </div>
-          `
+          </div>`
 
           node.type = `html`
           node.value = rawHTML


d4rekanguok:
Here is a reproduction site that demonstrate the bug: 

site: https://nifty-euclid-b36a51.netlify.com/hello-world/

repo: https://github.com/d4rekanguok/gatsby-bug-repro-18266 

Inspected markup of the iframe map:
```
<undefined>
  <div class="gatsby-resp-iframe-wrapper" style="padding-bottom: 66.6667%; position: relative; height: 0px; overflow: hidden; margin-bottom: 1.0725rem;">
    <iframe id="inlineFrameExample" title="Inline Frame Example" src="https://www.openstreetmap.org/export/embed.html?bbox=-0.004017949104309083%2C51.47612752641776%2C0.00030577182769775396%2C51.478569861898606&amp;layer=mapnik" style="position: absolute; top: 0px; left: 0px; width: 100%; height: 100%;"></iframe>
  </div>
</undefined>
```

d4rekanguok:
Oops I forgot to update the jest snapshot, sorry about that. Should I add a test against `util-hast-raw` as well, to ensure the resulted html ast is always unwrappable?

wardpeet:
Good find! Thanks for the reproduction.

How do you feel about stripping all newlines and whitespaces instead of manually fixing this issue.

We use a package called [common-tags](https://www.npmjs.com/package/common-tags) on a few places. We could do and we'll get a single line in our html.

```js
const { oneLine } = require("common-tags")

// more code

const rawHTML = oneLine`
html
`
```

d4rekanguok:
@wardpeet Sure, updated!

d4rekanguok:
Hhhm it looks like the existing test snapshot has a few issues that's not related to this PR -- I spot quite a few `padding-bottom: NaN%` in there. I've made sure those are not introduced by this PR <sup>[source](https://github.com/gatsbyjs/gatsby/blob/16e4e0e8cfe210f4e47ed6eb88d63b779cc08c42/packages/gatsby-remark-responsive-iframe/src/__tests__/__snapshots__/index.js.snap#L65)</sup>.

I'll take a look and possibly send another PR for those later

ANDIKKY:
Should I add a test against util-hast-raw as well, to ensure the resulted html ast is always unwrappable,,

Issue #18284, opened 2019-10-07
[docs] Gatsby Style Guide blitz for those new to open source


marcysutton:
## Summary

There are a ton of opportunities for small improvements to the [Gatsby Docs](https://www.gatsbyjs.org/docs/) that Hacktoberfest fans could take on. For folks new to open source, here are some ideas for contributing to the docs to make them adhere to the [Gatsby Style Guide](https://www.gatsbyjs.org/contributing/gatsby-style-guide/):

- Update Markdown docs titles to use [Title Case](https://www.gatsbyjs.org/contributing/gatsby-style-guide/#format-titles-and-headers)
- Update titles in the [docs sidebar](https://github.com/gatsbyjs/gatsby/blob/master/www/src/data/sidebars/doc-links.yaml) to use Title Case
- Update pronouns in a doc to use ["you" and "your"](https://www.gatsbyjs.org/contributing/gatsby-style-guide/#use-you-as-the-pronoun) instead of "we" and "our"
- Fix the [heading order](https://www.gatsbyjs.org/contributing/docs-contributions/#headings) of a doc to create an accessible content hierarchy (e.g. removing an extra h1, making headings go in order, etc.)

Other suggestions are welcome: chime in through the comments below! One approach could be to do a small set of changes to 5-10 docs in a PR, but one change at a time would be welcome too.

### Process recommendations

To make sure your PR is accepted, try to limit its scope to changes that form logical groups. For example, a PR with 5-10 title case changes or a PR with pronoun changes from "we" to "you" would be helpful (see the [Style Guide](https://www.gatsbyjs.org/contributing/gatsby-style-guide/#use-you-as-the-pronoun) for when the subject of "we" is referring to the code).

### Motivation

Grammar enthusiasm and Hacktoberfest for the win!

### Additional resources

- [Gatsby Hacktoberfest umbrella issue](https://github.com/gatsbyjs/gatsby/issues/18129)
- [Gatsby Style Guide](https://www.gatsbyjs.org/contributing/gatsby-style-guide/)
- [How to Open a PR](https://www.gatsbyjs.org/contributing/how-to-open-a-pull-request/)
- [Docs contributions](https://www.gatsbyjs.org/contributing/docs-contributions/)

If you have any questions, feel free to reach out to us here on GitHub or on Twitter or Discord. https://www.gatsbyjs.org/contributing/how-to-contribute/#not-sure-how-to-start-contributing

aditsachde:
[...]

PR #18297, opened 2019-10-08
Make minor edits to theme docs

jacob-shu:
## Description

Changed h2/h3s to sentence case per the Style Guide. Made minor clarifications.
#
## Related Issues
Addresses Issue: [#18284](https://github.com/gatsbyjs/gatsby/issues/18284#issue-503707016)

### DIFF
--- a/docs/docs/themes/converting-a-starter.md
+++ b/docs/docs/themes/converting-a-starter.md
@@ -36,7 +36,7 @@ const createPosts = (createPage, createRedirect, edges) => {
 }
 ```
 
-Since `path.resolve` is being used the starter will result `src/templates/post.js` rather than `node_modules/gatsby-theme-awesome-blog/src/templates/post.js`. In order to fix this you can use `require.resolve` which will look relative to the theme so the correct template is found.
+Since `path.resolve` is being used the starter will resolve `src/templates/post.js` rather than `node_modules/gatsby-theme-awesome-blog/src/templates/post.js`. In order to fix this you can use `require.resolve` which will look relative to the theme so the correct template is found.
 
 ```js
 const createPosts = (createPage, createRedirect, edges) => {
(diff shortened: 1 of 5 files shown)


docs/docs/themes/converting-a-starter.md
```diff
@@ -36,7 +36,7 @@ const createPosts = (createPage, createRedirect, edges) => {
 }
 ```
 
-Since `path.resolve` is being used we result in `src/templates/post.js` rather than `node_modules/gatsby-theme-awesome-blog/src/templates/post.js`. In order to fix this we can use `require.resolve` which will look relative to the theme so the correct template is found.
+Since `path.resolve` is being used the starter will result `src/templates/post.js` rather than `node_modules/gatsby-theme-awesome-blog/src/templates/post.js`. In order to fix this you can use `require.resolve` which will look relative to the theme so the correct template is found.
```
johno:
```suggestion
Since `path.resolve` is being used the starter will resolve `src/templates/post.js` rather than `node_modules/gatsby-theme-awesome-blog/src/templates/post.js`. In order to fix this you can use `require.resolve` which will look relative to the theme so the correct template is found.
```

docs/docs/themes/shadowing.md
```diff
@@ -163,9 +163,9 @@ This is a quick and efficient way to customize rendering without needing to worr
 
 ### Applying new props
 
-In some cases components offer prop APIs to change their behavior. With component extending you can import that component and then add your new prop to change it.
+In some cases components offer prop APIs to change their behavior. By extending this type of component you can import it and then add your new prop to change its behavior.
```
johno:
```suggestion
In some cases components offer prop APIs to change their behavior. To extend a component you can import it and then add a new prop.
```

Issue #18242, opened 2019-10-07
[docs] [workflows] Using Gatsby Themes


thomaswangio:
Part of the **Top 25 Learning Workflows initiative**. See #13708 for the meta issue that this issue falls under.

# User story

As a new Gatsby user, I want to use Themes in my Gatsby project.

# Evaluation

| Search | Discover | Complete | Linked | Tone | Style | Overall |
| --- | --- | --- | --- | --- | --- | --- |
| 😄 | 😄 | 😄 | 😄 | 😄 | 😄 | 😄 |

# Steps taken to implement

1. Searchability
	1. Searched `gatsby themes` on Google; clicked on the [first result - What Are Gatsby Themes?](https://www.gatsbyjs.org/docs/themes/what-are-gatsby-themes/)
	1. [Building a Theme](https://www.gatsbyjs.org/tutorial/building-a-theme/) tutorial has incorrect date (SEO)
		1. ✅**[rec]** Fix seo date for this tutorial

<img width="577" alt="Screen Shot 2019-10-23 at 5 31 45 PM" src="https://user-images.githubusercontent.com/41944255/67443787-27fb9780-f5bb-11e9-8804-b42b8d97ba9a.png">

2. Discoverability
	1. Searched `themes` on .org; clicked on the [first result - Themes](https://www.gatsbyjs.org/docs/themes/)
	1. Clickpath on .org (**3 clicks**):
		1. clicked on *Docs* in Navbar
		1. clicked on *Reference Guides* in Sidebar
		1. clicked on *Themes* in *Reference Guides* page, under *In this section:*
		- ✅ **[rec]**: Add Themes to "Ecosystem" on .org home page, 1 click for direct link to documentation
3. Completeness
	1. Read through [What Are Gatsby Themes?](https://www.gatsbyjs.org/docs/themes/what-are-gatsby-themes/). 
		- ✅ **[rec]** Add explanation of when to use themes vs. starters
		- ✅ **[rec]** Add all themes sections to "What's next?"
		- ✅ **[rec]** Put a one-liner of what Gatsby Themes are at the top of the doc on What are Gatsby Themes?
	1. Read through [Using a Gatsby Theme](https://www.gatsbyjs.org/docs/themes/using-a-gatsby-theme/)
		- ✅ **[rec]** rename doc as "Getting Started"
		- ✅ **[rec]** Make this page about how to install a theme directly. Answer where themes must be published to pull into your project
[...]

PR #18301, opened 2019-10-08
feat(www): Add Themes to "Ecosystem" on home page (option 1)

thomaswangio:
## Description

In this PR, to improve the discoverability of Themes, adding Themes to:
1. the homepage, in the "Ecosystem" section. 
1. the "Ecosystem" page.

#### Home Page
Screen Shot 2019-10-08 at 2 55 49 PM

#### Ecosystem Page

Screen Shot 2019-10-08 at 2 53 54 PM

#### Home Page
Screen Shot 2019-10-08 at 2 51 32 PM

#### Ecosystem Page
Screen Shot 2019-10-08 at 2 53 46 PM

## Related Issues/PRs

Part of Workflows Issue https://github.com/gatsbyjs/gatsby/issues/18242
https://github.com/gatsbyjs/gatsby/pull/18586 (option 2)


wardpeet:
On the homepage it looks a bit misplaced in my opinion. I'm unsure how this fits nicely in the current design. We could just mix it up with starters and add a themes marker 🤷‍♂.

Let's wait for @gatsbyjs/learning, @shannonbux and @fk

thomaswangio:
> On the homepage it looks a bit misplaced in my opinion. I'm unsure how this fits nicely in the current design. We could just mix it up with starters and add a themes marker 🤷‍♂.
> 
> Let's wait for @gatsbyjs/learning, @shannonbux and @fk

Yea, agreed, could definitely use a re-design. I noticed @fk just merged his re-design with dark mode so I'll post the new screenshots and see how it looks. 

I think the issue is that Themes are actually technically plugins (but not all plugins are themes), and Themes function similarly to a starter, effectively making Themes a combination of both plugins and starters. Would like to get @marcysutton's opinion on the best way to convey how Themes fit in with the rest of the Ecosystem, since people are having trouble understanding how to use Themes.

thomaswangio:
> > On the homepage it looks a bit misplaced in my opinion. I'm unsure how this fits nicely in the current design. We could just mix it up with starters and add a themes marker 🤷‍♂.
> > Let's wait for @gatsbyjs/learning, @shannonbux and @fk
> 
> Yea, agreed, could definitely use a re-design. I noticed @fk just merged his re-design with dark mode so I'll post the new screenshots and see how it looks.
> 
> I think the issue is that Themes are actually technically plugins (but not all plugins are themes), and Themes function similarly to starters, effectively making Themes a combination of both plugins and starters. Would like to get @marcysutton's opinion on the best way to convey how Themes fit in with the rest of the Ecosystem, since people are having trouble understanding how to use Themes.

## New screenshots

Screen Shot 2019-10-08 at 2 51 32 PM
Screen Shot 2019-10-08 at 2 53 46 PM

thomaswangio:
Updated description of Themes to: "Themes are special Plugins that add preconfigured functionality to Gatsby sites. You can update central themes across sites and use multiple themes in your project!"

marcysutton:
Here's an idea: what if we rename "Plugins" to be "Plugins & Themes" so that we don't have to change the layout? Since themes are essentially plugins with a `gatsby-config` file, that might make the most sense -- and it would make it less of an impact on the layout.

marcysutton:
We decided to close this to allow more of a discussion for homepage approaches, since that cuts across multiple teams at Gatsby. Thanks for your thinking on this though, @thomaswangio!

PR #18303, opened 2019-10-08
feat(gatsby-transformer-remark): Handle CJK characters for timeToRead

### LINKED ISSUE #17988: Word count for CJK character
## Summary

Currently, `gatsby-transformer-remark` use lodash.words to count words and generate reading time, which doesn't count any CJK characters at all.

I think just count the number of CJK characters and add to the result is fine enough since basically `hexo-reading-time` behave like this. I'd like to know if pr is welcomed

Austaras:
## Description

Currently gatsby-transformer-remark uses lodash.words to count words for timeToRead which only count latin characters. This PR make it also count CJK characters

## Related Issues

Fixes #17988


packages/gatsby-transformer-remark/src/extend-node-type.js
```diff
@@ -586,6 +586,11 @@ module.exports = (
             const avgWPM = 265
             const wordCount = _.words(pureText).length
             timeToRead = Math.round(wordCount / avgWPM)
+            const avgCJKPM = 400
+            const CJKCount = pureText.match(
```
wardpeet: can you elaborate a bit on what we're matching here? I'm sorry but I'm not familiar with CJK can't we combine 2 reading counters into 1?

Austaras: > can you elaborate a bit on what we're matching here? I'm sorry but I'm not familiar with CJK can't we combine 2 reading counters into 1?

Chinese and Japanese words aren't separated by space so the most easy solution is counting characters. What is counted here is Chinese characters(`\p{sc=Han}`) and Japanese characters(`\p{sc=Katakana}\p{sc=Hiragana}\p{sc=Han}`).

It's possible to combine into one pass but that requires not using lodash.words. I will look into that later

Austaras:
After reading wiki I found that Korean Hangul is separated by spaces so they are covered by original count. So it's actually count CJ for timeToRead

sidharthachatterjee:
@Austaras Perhaps some tests here might help?

Austaras:
test added

daacheen:
I am pretty looking forward to this PR, but it looks like this fix hasn't been released yet. Could anyone take a look at it please? There is a build error according to the build log.

devrsi0n:
Hi guys, I installed latest `gatsby-transformer-remark@2.6.30` still can't fix this issue.😢

I don't think code publish successfully as `node_modules` code show below.
Snipaste_2019-10-24_20-59-44

gatsby repo master code show below.
image


🙏 Please fix the problem and republish this package.

PR #18313, opened 2019-10-08
docs: document the commit naming convention

lekterable:
## Description

Just a small note of our preferred commit-naming convention so that it would leave less work for the people merging them. [Link](https://github.com/gatsbyjs/gatsby/pull/18218#issuecomment-539431399) to a small discussion we've had with @wardpeet regarding this.

### DIFF
--- a/docs/contributing/how-to-open-a-pull-request.md
+++ b/docs/contributing/how-to-open-a-pull-request.md
@@ -52,6 +52,10 @@ To test changes locally against the Gatsby [site and project files](https://gith
   git add .
   ```
   - Using a visual tool like [GitHub Desktop](https://desktop.github.com/) or [GitX](https://rowanj.github.io/gitx/) can help for choosing which files and lines to commit.
+- To make it easier for the new contributors we don't lint the commit names, but our preffered convention is [conventional commits](https://www.conventionalcommits.org)
+  ```shell
+  git commit -m '<type>(optional scope): <description>'
+  ```
 - Committing code will run the automated linter using [Prettier](https://prettier.io). To run the linter manually, run an npm script in the project's base directory:
   ```shell
   npm run format


pieh:
Hey! Just some clarification - we don't want to force commit naming convention on each individual commits in their branches. We only really apply the convention to Pull Request titles (because we squash and merge PRs, those end up as single commit in `master` branch, and PR title is being used as commit message automatically).

lekterable:
Hey @pieh , thanks for the clarification. Do you think that I should put this note somewhere else? My idea was that following the convention while creating individual commits was just a good habit which is still as noted just `preferred` and can be only beneficial. + for PR's with just one commit, which I believe most of the first-timers create as their first contribution, it's the commit's msg that will be used as a default title and people not always remember to edit it

pieh:
I think this document should have as little info as necessary to open pull request. I think majority of users that benefit from that document are people that are about to open their first Pull Request. Making sure that we don't introduce detours for optional things makes it more digestible.

That said, I'm not sure if we have better place to list optional things, so maybe putting them in some kind of appendix, might be way forward?

But I'm not expert on documentation, so let's ask @gatsbyjs/learning about this ;)

lekterable:
Other bullet which IMO seems like a good place to put such note would be:

Screenshot 2019-10-08 at 13 25 42

Here we are in the process of opening a PR and something like "Our preferred way of titling the PR's is ..." seems to be pretty straightforward, especially that we are linking the github docs there which contain images of PR's title field etc. let me know if you like this idea better, or maybe we should put in both places if @gatsbyjs/learning don't consider the previous change to be too much.

marcysutton:
There’s already a line in there about structuring commits that points to more details in the Managing PRs doc, I personally would say this change is overkill for new contributors given the squash and title change workflow for core. It doesn’t matter as much for docs https://github.com/gatsbyjs/gatsby/pull/18313/files#diff-e0d3555b671ad96068e6a0c25ee4136dR49

sidharthachatterjee:
Thanks for opening this @lekterable 

Like @marcysutton and @pieh mentioned, this doc ought to be as minimal as possible so we don't overwhelm new contributors. As such, I'm closing this PR for now but thank you once again for taking the time to open and discuss this! 

Cheers

PR #18323, opened 2019-10-08
Docs: Add SASS/SCSS recipe

1qk1:
## Changes
- Added Sass recipe

## Related Issues
none

### DIFF
--- a/docs/docs/recipes.md
+++ b/docs/docs/recipes.md
@@ -518,7 +518,7 @@ import "./styles.scss"
 import "./styles.sass"
 ```
 
-_Note: You can use Sass/SCSS files as modules too, like mentioned above, with the difference that instead of .css the extensions have to be .scss or .sass_
+_Note: You can use Sass/SCSS files as modules too, like mentioned in the previous recipe about CSS modules, with the difference that instead of .css the extensions have to be .scss or .sass_
 
 #### Additional resources
 
--- a/docs/docs/recipes.md
+++ b/docs/docs/recipes.md
@@ -468,6 +468,64 @@ Notice that the file extension is `.module.css` instead of `.css`, which tells G
 - More on [Using CSS Modules](/tutorial/part-two/#css-modules)
 - [Live example on Using CSS modules](https://github.com/gatsbyjs/gatsby/blob/master/examples/using-css-modules)
 
+### Using SCSS/SASS
+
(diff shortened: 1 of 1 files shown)


docs/docs/recipes.md
```diff
@@ -468,6 +468,64 @@ Notice that the file extension is `.module.css` instead of `.css`, which tells G
 - More on [Using CSS Modules](/tutorial/part-two/#css-modules)
 - [Live example on Using CSS modules](https://github.com/gatsbyjs/gatsby/blob/master/examples/using-css-modules)
 
+### Using SCSS/SASS
```
gillkyle:
```suggestion
### Using SCSS/Sass
```

docs/docs/recipes.md
```diff
@@ -468,6 +468,64 @@ Notice that the file extension is `.module.css` instead of `.css`, which tells G
...
+  font: 100% $font-stack
+  color: $primary-color
+```
+
+```javascript
+import "./styles.scss"
+import "./styles.sass"
+```
+
+_Note: You can use SASS/SCSS files as modules too, like mentioned above, with the difference that instead of .css the extensions have to be .scss or .sass_
```
gillkyle:
```suggestion
_Note: You can use Sass/SCSS files as modules too, like mentioned above, with the difference that instead of .css the extensions have to be .scss or .sass_
```

docs/docs/recipes.md
```diff
@@ -468,6 +468,64 @@ Notice that the file extension is `.module.css` instead of `.css`, which tells G
...
+
+#### Additional resources
+
+- [Difference between .sass and .scss](https://responsivedesign.is/articles/difference-between-sass-and-scss/)
+- [Sass guide from the official Sass website](https://sass-lang.com/guide)
+- [A most complete installation tutorial on Sass with some more explanations and more resources](https://www.gatsbyjs.org/docs/sass/)
```
gillkyle:
```suggestion
- [A more complete installation tutorial on Sass with some more explanations and more resources](https://www.gatsbyjs.org/docs/sass/)
```

1qk1:
Thanks, I wasn't sure how to write it because SCSS is all uppercase but in Sass only the first one is uppercase, and I wanted them to look the same when written together haha. I will correct the order I write them too (Sass/SCSS and SCSS/Sass)

1qk1:
Updated, check again @gillkyle!

Issue #18319, opened 2019-10-08
[gatsby-plugin-netlify] Error: The "headers" option to gatsby-plugin-netlify is in the wrong shape


ronanlevesque:
## Description

I can't run `gatsby build`. The build fails before the end with the following error message: "Error: The "headers" option to gatsby-plugin-netlify is in the wrong shape. You should pass in a object with string keys (representing the paths) and an array of strings as the value (representing the headers). Check your gatsby-config.js."

### Steps to reproduce

1) Start a new Gatsby project
2) Install `gatsby-plugin-netlify`
3) Update `gatsby-config.js` with the following:

```
{
  resolve: "gatsby-plugin-netlify",
  options: {
    mergeSecurityHeaders: false,
    headers: {
      "X-Frame-Options": ["sameorigin"],
      "/**/*.html": ["cache-control: public, max-age=300"],
      "/static/*": ["cache-control: public, max-age=31536000, immutable"],
      "/*.js": ["cache-control: public, max-age=31536000, immutable"],
      "/*.css": ["cache-control: public, max-age=31536000, immutable"],
      "/sw.js": ["cache-control: public, max-age=0, must-revalidate"],
    },
  },
},
```

### Expected result

The build should pass, the configuration above has been used for months without any issue.

### Actual result

Build is failing (see error message above)

### Environment

```
❯ npx gatsby info --clipboard

  System:
    OS: Linux 4.4 Ubuntu 14.04.4 LTS, Trusty Tahr  
    CPU: (6) x64 Intel(R) Core(TM) i5-8600K CPU @ 3.60GHz
    Shell: 4.3.11 - /bin/bash
  Binaries:
    Node: 10.7.0 - /usr/local/bin/node
    Yarn: 1.9.4 - /usr/local/bin/yarn
    npm: 6.4.1 - /usr/local/bin/npm
  Languages:
    Python: 2.7.6 - /usr/bin/python
  npmPackages:
    gatsby: ^2.15.31 => 2.15.31
    gatsby-image: ^2.2.24 => 2.2.24
    gatsby-plugin-manifest: ^2.2.20 => 2.2.20      
    gatsby-plugin-netlify: ^2.1.17 => 2.1.17       
    gatsby-plugin-offline: ^3.0.12 => 3.0.12       
    gatsby-plugin-react-helmet: ^3.1.10 => 3.1.10  
    gatsby-plugin-sharp: ^2.2.28 => 2.2.28
[...]

PR #18326, opened 2019-10-08
fix(gatsby-plugin-netlify): add warning instead for invalid headers

wardpeet:
## Description
Fixes a regression in #17538

Instead of failing we show a warning
image

You can mimic previous behaviour with:
```
{
  resolve: "gatsby-plugin-netlify",
  options: {
    mergeSecurityHeaders: false,
    headers: {
      "X-Frame-Options": ["sameorigin"],
    },
  },
},
```

## Related Issues
PR introducing regression https://github.com/gatsbyjs/gatsby/pull/17538
Fixes https://github.com/gatsbyjs/gatsby/issues/18319

### DIFF
--- a/packages/gatsby-plugin-netlify/src/build-headers-program.js
+++ b/packages/gatsby-plugin-netlify/src/build-headers-program.js
@@ -17,7 +17,7 @@ function getHeaderName(header) {
   return matches && matches[1]
 }
 
-function validHeaders(headers) {
+function validHeaders(headers, reporter) {
   if (!headers || !_.isObject(headers)) {
     return false
   }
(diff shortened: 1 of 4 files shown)


packages/gatsby-plugin-netlify/src/build-headers-program.js
```diff
@@ -26,10 +26,19 @@ function validHeaders(headers) {
...
     (headersList, path) =>
       _.isArray(headersList) &&
-      _.every(
-        headersList,
-        header => _.isString(header) && getHeaderName(header)
-      )
+      _.every(headersList, header => {
+        if (_.isString(header)) {
+          if (!getHeaderName(header)) {
+            reporter.warn(
```
pieh: Should this have more severe level? I.e. fail build completely to not deploy potentially broken build?

packages/gatsby-plugin-netlify/src/build-headers-program.js
```diff
@@ -26,10 +26,19 @@ function validHeaders(headers) {
...
+              `${path} contains an invalid header (${header}). Please check your plugin configuration`
```
pieh: In the terminal it doesn't say that this came from `gatsby-plugin-netlify`, so it be hard for users to figure out what's going on - let's mention plugin name in this message

wardpeet: oh yeah, let me do that

PR #18328, opened 2019-10-08
fix(gatsby-remark-responsive-iframe): iframe ratio is NaN if dimensions use px unit

d4rekanguok:
## Description

Hi folks!

I noticed a few `padding-bottom: NaN%` in gatsby-remark-responsive-iframe's test snapshot  <sup>[source](https://github.com/gatsbyjs/gatsby/blob/16e4e0e8cfe210f4e47ed6eb88d63b779cc08c42/packages/gatsby-remark-responsive-iframe/src/__tests__/__snapshots__/index.js.snap#L65)</sup>.

This was caused by the pixel unit not being stripped off when calculating the iframe ratio <sup>[source](https://github.com/gatsbyjs/gatsby/blob/16e4e0e8cfe210f4e47ed6eb88d63b779cc08c42/packages/gatsby-remark-responsive-iframe/src/index.js#L68)</sup>.

I think it'd be simpler to try & convert the dimensions into number first then check if it's finite, that way we won't have to check whether the dimensions are px or raw number twice.

I also took the liberty of modernizing the code a little bit:
- use native promise + async/await, remove bluebird dependency
- check for negative condition instead of positive one, so we don't have so many nested code block


d4rekanguok:
@wardpeet I'd appreciate your input on this one!

yarn.lock
```diff
@@ -21119,7 +21119,7 @@ wrappy@1:
   version "1.0.2"
   resolved "https://registry.yarnpkg.com/wrappy/-/wrappy-1.0.2.tgz#b5243d8f3ec1aa35f1364605bc0d1036e30ab69f"
 
-write-file-atomic@2.4.1, write-file-atomic@^2.0.0, write-file-atomic@^2.3.0:
+write-file-atomic@2.4.1, write-file-atomic@^2.0.0, write-file-atomic@^2.3.0, write-file-atomic@^2.4.2, write-file-atomic@^3.0.0:
   version "2.4.1"
```
pieh: Hmm, this part looks pretty weird - how `write-file-atomic@^2.4.2, write-file-atomic@^3.0.0` resolve to 2.4.1?

d4rekanguok: It is weird indeed, this is the result of running `yarn workspace gatsby-remark-responsive-iframe remove bluebird`, which doesnt have any dependencies and therefore shouldnt update yarn.lock at all?

I think I should be able to restore yarn.lock without affecting anything

d4rekanguok: So I previously added a new package to this plugin (in #18266), which moved write-file-atom 2.4.2 and 3.0.0 to their own entries in yarn.lock

This change here actually reverses that, restoring this line to what it was before the dependencies changes.

I'm... confused

pieh: Don't worry about this too much - just something that caught my eye. I'll pull this locally and see if I get same behaviour with lock file

pieh: Seems like we have yarn resolutions in our main package.json which would explain this - https://github.com/gatsbyjs/gatsby/blob/0ba9890f9e92b9eeb644e325825be263462064c6/package.json#L112-L114

But this was added ~2 months ago, so have no idea why lock didn't consolidate those before.

In any case - we should be good here

packages/gatsby-remark-responsive-iframe/src/index.js
```diff
@@ -1,85 +1,83 @@
...
 
-const isUnitlessOrPixelNumber = n =>
-  n && (isUnitlessNumber(n) || isPixelNumber(n))
+/**
+ * Convert anything to number, except for % value
+ * @param {*} n something to be converted to number
+ * @returns {number}
+ */
+const convert = n =>
+  typeof n === `string` && n.endsWith(`%`) ? NaN : parseInt(n, 10)
```
pieh: On first reading of this I thought it would be problematic if unit other than `%` or `px` was used - but seems like iframe width/height don't understand other units (like `em`, `rem`, `vw`, etc) and just treats them as `px` - worth adding comment about this, that we don't have to check units other than `%` because of that?

yarn.lock
d4rekanguok: Ah good to know, thank you. I've just looked at the previous PR, and the lock file changes was introduced randomly in a test snapshot update commit, so it's likely that I goofed something up.

I will pay more attention to unnecessary file changes in my commits.

pieh: Don't sweat it - we are all good here! :)

packages/gatsby-remark-responsive-iframe/src/index.js
d4rekanguok: That's a good idea, I should probably add that. 

Browser engines parse `width` & `height` the same way as `parseInt(n, 10)`, so actually even `<iframe width="500 this is not invalid">` will return an iframe with 500px width! If it was an invalid string, it'll fallback to the default 300px.

d4rekanguok: Hhhm I goofed here, `n.endsWith(`%`)` doesn't account for `width="100% "`. Will trim that off

pieh: Nice catch with adding `.trim()`!

d4rekanguok:
@pieh thank you for the speedy review!

PR #18332, opened 2019-10-08
feat(gatsby-cli): Add instructions after new

anikethsaha:
## Description
In this PR, I have added an after setup instruction for the `gatbsy-cli new` command.
Which looks something like this..

image

The above log will be printed when the installation and project setup is getting successful 

## Related Issues
N/A

### DIFF
--- a/packages/gatsby-cli/src/init-starter.js
+++ b/packages/gatsby-cli/src/init-starter.js
@@ -228,7 +228,7 @@ type InitOptions = {
   rootPath?: string,
 }
 
-const endInstruction = Path => {
+const successMessage = path => {
   report.info(`
 Your new Gatsby site has been successfully bootstrapped. Start developing it by running:
   $ cd ${Path}
--- a/packages/gatsby-cli/src/init-starter.js
+++ b/packages/gatsby-cli/src/init-starter.js
@@ -232,7 +232,7 @@ const successMessage = path => {
   report.info(`
 Your new Gatsby site has been successfully bootstrapped. Start developing it by running:
   $ cd ${path}
-  $ ${getPackageManager()} run develop
+  $ gatsby develop
 `)
 }
 
--- a/packages/gatsby-cli/src/init-starter.js
+++ b/packages/gatsby-cli/src/init-starter.js
@@ -228,6 +228,15 @@ type InitOptions = {
   rootPath?: string,
 }
 
+const endInstruction = Path => {
+  report.info(`
+Successfully Added Gatsby site !!
+start developing your site now by running the following instructions
+  $ cd ${Path}
+  $ ${getPackageManager()} run develop
+`)
+}
+
 /**
  * Main function that clones or copies the starter.
  */
@@ -286,5 +295,6 @@ module.exports = async (starter: string, options: InitOptions = {}) => {
   })
   if (hostedInfo) await clone(hostedInfo, rootPath)
   else await copy(starterPath, rootPath)
+  await endInstruction(rootPath)
   trackCli(`NEW_PROJECT_END`)
 }
--- a/packages/gatsby-cli/src/init-starter.js
+++ b/packages/gatsby-cli/src/init-starter.js
@@ -294,6 +294,6 @@ module.exports = async (starter: string, options: InitOptions = {}) => {
   })
   if (hostedInfo) await clone(hostedInfo, rootPath)
   else await copy(starterPath, rootPath)
-  await endInstruction(rootPath)
+  successMessage(rootPath)
   trackCli(`NEW_PROJECT_END`)
 }
--- a/packages/gatsby-cli/src/init-starter.js
+++ b/packages/gatsby-cli/src/init-starter.js
@@ -230,7 +230,7 @@ type InitOptions = {
 
 const endInstruction = Path => {
   report.info(`
-Successfully Added Gatsby site !!
+Your new Gatsby site has been successfully bootstrapped. Start developing it by running:
 start developing your site now by running the following instructions
   $ cd ${Path}
   $ ${getPackageManager()} run develop
--- a/packages/gatsby-cli/src/init-starter.js
+++ b/packages/gatsby-cli/src/init-starter.js
@@ -231,7 +231,6 @@ type InitOptions = {
 const endInstruction = Path => {
   report.info(`
 Your new Gatsby site has been successfully bootstrapped. Start developing it by running:
-start developing your site now by running the following instructions
   $ cd ${Path}
   $ ${getPackageManager()} run develop
 `)
--- a/packages/gatsby-cli/src/init-starter.js
+++ b/packages/gatsby-cli/src/init-starter.js
@@ -231,7 +231,7 @@ type InitOptions = {
 const successMessage = path => {
   report.info(`
 Your new Gatsby site has been successfully bootstrapped. Start developing it by running:
-  $ cd ${Path}
+  $ cd ${path}
   $ ${getPackageManager()} run develop
 `)
 }


packages/gatsby-cli/src/init-starter.js
```diff
@@ -228,6 +228,15 @@ type InitOptions = {
   rootPath?: string,
 }
 
+const endInstruction = Path => {
+  report.info(`
+Successfully Added Gatsby site !!
+start developing your site now by running the following instructions
+  $ cd ${Path}
+  $ ${getPackageManager()} run develop
```
pieh: This might not always work (it depends if starter is using `develop` script or not)

sidharthachatterjee: We could replace this with `gatsby develop` since we assume that the user has `gatsby-cli` globally installed.

packages/gatsby-cli/src/init-starter.js
```diff
@@ -228,6 +228,15 @@ type InitOptions = {
   rootPath?: string,
 }
 
+const endInstruction = Path => {
+  report.info(`
+Successfully Added Gatsby site !!
```
sidharthachatterjee:
```suggestion
Your new Gatsby site has been successfully bootstrapped. Start developing it by running:
```

packages/gatsby-cli/src/init-starter.js
```diff
@@ -228,6 +228,15 @@ type InitOptions = {
...
+start developing your site now by running the following instructions
```
sidharthachatterjee:
```suggestion

```

packages/gatsby-cli/src/init-starter.js
```diff
@@ -228,6 +228,15 @@ type InitOptions = {
   rootPath?: string,
 }
 
+const endInstruction = Path => {
```
sidharthachatterjee:
```suggestion
const successMessage = path => {
```

packages/gatsby-cli/src/init-starter.js
```diff
@@ -286,5 +295,6 @@ module.exports = async (starter: string, options: InitOptions = {}) => {
   })
   if (hostedInfo) await clone(hostedInfo, rootPath)
   else await copy(starterPath, rootPath)
+  await endInstruction(rootPath)
```
sidharthachatterjee:
```suggestion
  successMessage(rootPath)
```

packages/gatsby-cli/src/init-starter.js
```diff
@@ -228,6 +228,15 @@ type InitOptions = {
   rootPath?: string,
 }
 
+const endInstruction = Path => {
+  report.info(`
+Successfully Added Gatsby site !!
+start developing your site now by running the following instructions
+  $ cd ${Path}
```
sidharthachatterjee:
```suggestion
  $ cd ${path}
```

packages/gatsby-cli/src/init-starter.js
sidharthachatterjee:
```suggestion
  $ gatsby develop
```

packages/gatsby-cli/src/init-starter.js
anikethsaha: Sorry my bad, this is the method so I guess no point of making it async. 
Will change it soon

packages/gatsby-cli/src/init-starter.js
anikethsaha: What do you think about `npx`. I guess it will fit good here

sidharthachatterjee: `npx` is okay but since the rest of the docs says `gatsby develop`, might be nice to keep it consistent. 

Also, they're very likely to have `gatsby` globally installed since they just ran `gatsby new` 😅

packages/gatsby-cli/src/init-starter.js
sidharthachatterjee: No worries, you should be able to commit all the suggestions 🙂

packages/gatsby-cli/src/init-starter.js
wardpeet: What about, checking if gatsby-cli is installed, if not, show npx gatsby-cli develop? 🤷‍♂

sidharthachatterjee: That's a good idea too! Since the user could've ran `npx gatsby new` 😂

wardpeet: npx comes with node 8 so we should be good to go, unsure if it's npx gatsby or npx gatsby-cli

sidharthachatterjee: Probably `npx gatsby-cli` since npx takes package name

anikethsaha: Checking `gatsby-cli` in local Or in global? 
Cause I guess for global, it will need workaround as path wont be same for diff os. Right?

sidharthachatterjee: Yeah, I think let's just set it to `gatsby develop` for now since we encourage folks to install `gatsby-cli` globally in all documentation. We can revisit this later if necessary.

anikethsaha: Yeah using gatsby seems good to me. 
I will try to create a better package manager locating method later if required

anikethsaha: I changed it to `gatsby` 😃

anikethsaha:
I added those suggestions 👍

packages/gatsby-cli/src/init-starter.js
```diff
@@ -228,6 +228,14 @@ type InitOptions = {
   rootPath?: string,
 }
 
+const successMessage = path => {
+  report.info(`
+Your new Gatsby site has been successfully bootstrapped. Start developing it by running:
+  $ cd ${path}
```
muescha: should we include a backticks in case there are spaces in the whole path? or it`s on the response of the user if he use spaces here?


```suggestion
  $ cd \`${path}\`
```

anikethsaha: How would this show as an output? Like for example if the path is `my-gatsby-starter` ?

muescha: sorry on command line we use this tick

```suggestion
  $ cd '${path}'
```

muescha: or replace spaces ` ` with `\ `

```bash
gatsby new 'site with space'
info Creating new site from git: https://github.com/gatsbyjs/gatsby-starter-default.git                                                                                                                 
                                                                                                                                                                                                        
Cloning into 'site with space'...                                                                                                                                                                       
remote: Enumerating objects: 17, done.                                                                                                                                                                  
remote: Counting objects: 100% (17/17), done.                                                                                                                                                           
remote: Compressing objects: 100% (13/13), done.                                                                                                                                                        
remote: Total 2245 (delta 9), reused 12 (delta 4), pack-reused 2228                                                                                                                                     
Receiving objects: 100% (2245/2245), 11.83 MiB | 769.00 KiB/s, done.                                                                                                                                    
success Created starter directory layout                                                                                                                                                                
info Installing packages...                                                                                                                                                                             
                                                                                                                                                                                                        
yarn install v1.17.3                                                                                                                                                                                    
[1/4] 🔍  Resolving packages...                                                                                                                                                                         
[2/4] 🚚  Fetching packages...                                                                                                                                                                          
[3/4] 🔗  Linking dependencies...                                                                                                                                                                       
warning "gatsby > eslint-config-react-app@4.0.1" has incorrect peer dependency "eslint-plugin-flowtype@2.x".                                                                                            
warning "gatsby > @typescript-eslint/eslint-plugin > tsutils@3.17.1" has unmet peer dependency "typescript@>=2.8.0 || >= 3.2.0-dev || >= 3.3.0-dev || >= 3.4.0-dev || >= 3.5.0-dev || >= 3.6.0-dev || >=
 3.6.0-beta || >= 3.7.0-dev || >= 3.7.0-beta".                                                                                                                                                          
[4/4] 🔨  Building fresh packages...                                                                                                                                                                    
success Saved lockfile.                                                                                                                                                                                 
info Initialising git in site with space                                                                                                                                                                
                                                                                                                                                                                                        
info Create initial git commit in site with space   
```

then both works

```
cd site\ with\ space
cd 'site with space'
```

anikethsaha: But wouldn't it lead to confusion to the user they may actually write those `'` along with the path when following those instructions. Right?

muescha: I don't know what's the best. Less confusion is with spaces less names is replacing spaces with `\ ` and then technicaly it is still correct if some one do it with space

Let's wait what the core team decides

anikethsaha: Yeah sure! lets wait for their answer!

anikethsaha: cc @sidharthachatterjee , @wardpeet

PR #18027, opened 2019-10-01
feat(www): Add gatsby-plugin-theme-ui, dark theme MVP

fk:
Haaaaaacktooooobeeeer!  
🍂⚗️📐🔦🧰🧹🔨🙃

This PR ~(draft)~ puts `www` on `gatsby-plugin-theme-ui` 🔌, adding a "dark mode" 🌙 along the way. It's ~fairly~ functional ~already~ (and builds ;-)), but not without a few rough edges ~—both visible, and~ below the surface.

image

image

image

image

**Basic questions (IMO ;-)) …**

- _Do we even want .org on `theme-ui`?_ 😛
- If so, do we want to ship the initial version with or without dark mode?  
  The current state of the dark theme is usable, but MVP.

**… and a few 🙄 notes on where this is at:**

- [x] add [`gatsby-plugin-theme-ui`](https://www.gatsbyjs.org/docs/theme-ui/#using-theme-ui-in-gatsby) using the [`sx` prop](https://theme-ui.com/sx-prop) along the [custom pragma comment](https://theme-ui.com/jsx-pragma)
  - [x] ditch `utils/presets` in favor of `src/gatsby-plugin-theme-ui`, which now is our single (and currently pretty ugly) [shadowed](https://theme-ui.com/gatsby-plugin) source of truth
  - [x] no need for `wrap-root-element` anymore in [`www/gatsby-ssr.js`](https://github.com/gatsbyjs/gatsby/pull/18027/files#diff-fcd07cc82c362910f0734270336e9b30)—`src/gatsby-plugin-theme-ui/components` instead
  - [x] refactor a lot of CSS to use `sx`, so most token imports (from the removed `utils/presets`) are gone
    - still importing a couple tokens here and there, now from `src/gatsby-plugin-theme-ui`—(mostly) `mediaQueries`
    - didn't opt in to the `theme-ui` responsive styles array for now; not against it, just a bit scared ;-)
- [x] move global styles out of `utils/typography` to `utils/styles/global` and break those into chunks (Prism tokens, code block styles and language badges)—I'm not attached to the new location though; both things simply helped refactoring
- [x] add a first shot at a dark theme
  - [x] adjust styles and assets for dark mode
    - [x] layer model component
    - [x] algolia docsearch dropdown
    - [x] `homepage-logo-banner` logo colors
    - [x] feedback widget
    - [x] homepage section
    - [x] newsletter components
      - [x] default
      - [x] homepage
    - [x] Showcase and Starters
      - [x] modal
[...]

PR #18335, opened 2019-10-08
fix(www): Update README to include theme-ui

fk:
Following up on #18027, this adjusts the `www` README to contain rudimentary documentation regarding `theme-ui`. Should have been in that original PR … :/ 🙏


fk:
🙏 <3

PR #18341, opened 2019-10-08
Change 'delete' to undefined assignment

jrflga:
Changed to an undefined assignment because it's faster than delete operation.

JS perf. comparison:
https://jsbench.me/b1k1i9jiy7/1

### DIFF
--- a/starters/blog/src/utils/typography.js
+++ b/starters/blog/src/utils/typography.js
@@ -9,7 +9,7 @@ Wordpress2016.overrideThemeStyles = () => {
   }
 }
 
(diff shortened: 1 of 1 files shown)


sidharthachatterjee:
Thanks for opening this @jrflga but we won't be accepting this because in our opinion, it's not a very valuable change. We'd like to avoid unnecessary churn and this looks like a micro optimisation that is very unlikely to affect real world users and _reduces_ readability.

Nonetheless, thank you so much once again for taking the time to open this! Cheers

PR #18345, opened 2019-10-08
feat(gatsby-source-drupal): secrets and delete functionality

grantglidewell:
## Updates to Drupal Source Plugin

Adds delete functionality and option to use a secret defined in the drupal gatsby module.


grantglidewell:
Ive created this as a draft PR until Ive completed testing, but this is likely close to what we want to implement.

packages/gatsby-source-drupal/src/gatsby-node.js
```diff
@@ -176,13 +177,20 @@ exports.onCreateDevServer = (
...
       type: `application/json`,
     }),
-    async (req, res) => {
-      // we are missing handling of node deletion
-      const nodeToUpdate = JSON.parse(JSON.parse(req.body)).data
-
-      await handleWebhookUpdate(
+    async (req, _res) => {
+      const requestBody = JSON.parse(JSON.parse(req.body))
+      const { data, secret, action, id } = requestBody
```
pieh: Nice! It seems like this is backward compatible with previous version of drupal's gatsby module?

grantglidewell: it would be yes, however deleteNode is not accessible, that I believe would be a change that would have to happen in `onCreateDevServer`

pieh: `deleteNode` is an action, so you should be able to use `actions.deleteNode`

grantglidewell:
From basic testing the secret works as expected. Ill need some support in getting the `deleteNode` method accessible from `onCreateDevServer`

grantglidewell:
We potentially may want some different language here:
Screen Shot 2019-10-08 at 2 26 15 PM

packages/gatsby-source-drupal/src/gatsby-node.js
grantglidewell: great, thank you

grantglidewell:
@pieh this is now working in my tests, thank for your help! let me know if anything needs to change.

packages/gatsby-source-drupal/README.md
```diff
@@ -156,6 +156,7 @@ module.exports = {
         baseUrl: `https://live-contentacms.pantheonsite.io/`,
         apiBase: `api`, // optional, defaults to `jsonapi`
         concurrentFileRequests: 60, // optional, defaults to `20`
+        secret: `xxxx-xxxx-xxxxx`, // optional, must match Drupal instance preview secret
```
marcysutton: Should this exist in an environment variable rather than committed in code for security purposes?

grantglidewell: totally agree, Ill update the documentation to reflect that.

grantglidewell: Thanks, I think the documentation is much clearer and consistent with the rest of the readme file now.

sidharthachatterjee:
@grantglidewell Looks like we have a couple of failing tests!

packages/gatsby-source-drupal/src/gatsby-node.js
```diff
@@ -176,13 +176,21 @@ exports.onCreateDevServer = (
...
+      if (pluginOptions.secret && pluginOptions.secret !== secret) {
+        return reporter.warn(
+          `The secret in this request did not match your plugin options secret.`
+        )
+      }
+      if (action === `delete`) {
+        actions.deleteNode({ node: getNode(createNodeId(id)) })
```
pieh: I do have question here:
What happens if given node is being referenced (relationship) by other node? Does drupal automatically update other posts that reference deleted content and remove those? Or are those not touched and then there is content that still references deleted content?

We definitely will need to add some code here to handle back references (I can work on that), question is if we need to handle normal references as well or drupal actually updates those automatically

grantglidewell: That is a great question. Ive asked for some insight into this on drupal, but I think a modified/extended version of what we already use to update relationships should be able to handle this.

Is this something `deleteNode` itself should/could handle?

pieh: This is something that gatsby core potentially could handle, but doesn't do it right now (it's up to plugins to keep track of relationships - mostly because it's usually more efficient than any generic solution we could have in core)

grantglidewell:
> @grantglidewell Looks like we have a couple of failing tests!

hmmm.. these seem specific to node8 on windows. looking into it.

packages/gatsby-source-drupal/src/gatsby-node.js
grantglidewell: It seems from my conversation with someone who knows Drupal, the relationship are not necessarily updated on delete and it depends on how content and relationships are set up as to when and how that update or removal may happen. ugh.

grantglidewell: here is an issue discussing the problem https://www.drupal.org/project/drupal/issues/2723323
Some of the fixes purely deal with Drupal view which would solve for JSON API problems.

grantglidewell:
@sidharthachatterjee The failures I seea re in `normalize` which this PR hasnt modified. Not sure how to handle this.

grantglidewell:
@sidharthachatterjee all tests are passing :)

grantglidewell:
@DSchau @sidharthachatterjee looks like tests have stalled, and something else was merged creating a conflict.

DSchau:
@grantglidewell want to resolve the merge conflict and then we can get this merged?

grantglidewell:
this push should resolve it

grantglidewell:
@DSchau all tests passing. that linting test could be added as a pre-commit hook. lol could have saved me two commits here. possibly something I could add if interested. I do remember at one point pre-commit hooks running.

DSchau:
Published as gatsby-source-drupal@3.3.0.

Thanks again Grant!

PR #18349, opened 2019-10-08
fix(www): Main nav breaks layout on screens <~400px

fk:
https://twitter.com/replaymag_de/status/1181677564101828610 :/ 🙏 

image

(fully responsive now, holding up fine also on 320px wide viewports)


fk:
(Fix regression introduced in #18027)

PR #18350, opened 2019-10-08
Installation recommendation

PHironaka:
I tried installing Typography.js on a pre existing site and ran into an issue. The <style> tag with Typography.js settings were loading fine, but the fonts weren't changing. That was because I had a few font attribute stylings associated within styled components in my Layout.js component.

I thought adding a brief note would be helpful to users who were looking to install Typography.js on a pre-exisiting Gatsby site.


docs/docs/typography-js.md
```diff
@@ -64,6 +64,8 @@ const typography = new Typography({
 export default typography
 ```
 
+If you're installing Typography.js into an existing Gatsby Project you've started, you will need to delete all css font styles from your codebase in favor of your new Typography.js settings.
```
marcysutton:
```suggestion
If you're installing Typography.js into an existing Gatsby project you've started, you will need to delete all conflicting CSS font styles from your codebase in favor of your new Typography.js settings.
```
I'd recommend making it a little more specific! Let me know what you think.

PHironaka: @marcysutton  perfect! I like your version.

PR #18354, opened 2019-10-09
fix(blog): removed large margin-left from blog post layout

gillkyle:
## Description

The blog layout got a large amount of margin with the dark mode updates in #18027.

Looks like this:

Screen Shot 2019-10-08 at 6 33 43 PM
 
Removing that margin returns it back to this:

Screen Shot 2019-10-08 at 6 33 57 PM

### DIFF
--- a/www/src/templates/template-blog-post.js
+++ b/www/src/templates/template-blog-post.js
@@ -59,7 +59,9 @@ class BlogPostTemplate extends React.Component {
               <title>{post.frontmatter.title}</title>
               <link
                 rel="author"
-                href={`https://gatsbyjs.org${post.frontmatter.author.fields.slug}`}
+                href={`https://gatsbyjs.org${
+                  post.frontmatter.author.fields.slug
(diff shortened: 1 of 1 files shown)


www/src/templates/template-blog-post.js
```diff
@@ -116,9 +116,6 @@ class BlogPostTemplate extends React.Component {
                     mt: 3,
                     mb: 9,
                   },
-                  [mediaQueries.lg]: {
-                    ml: `-8em`,
```
marcysutton: Where did this originate, out of curiosity?

gillkyle: It wasn't in the previous commit. I wonder if it got introduced in a weird merge or if @fk was working on layouts and didn't notice it at the large screen size while reworking that page with the sx prop. I didn't poke too much through the commit log but in the history for the file it wasn't in the previous commit but was in the latest adding dark mode.

PR #18357, opened 2019-10-09
fix(www): Dark mode button for older Mobile Safari

fk:
Mobile Safari on iOS 12.4.2 won't render the dark mode toggle button correctly unless we help a little :/


danhannigan:
This was impacting current Mobile Safari as well (13) - was just about to file an issue and kick up a PR, but you got it!

fk:
👋 🙏 Thank you Dan, also for letting me know!
I unfortunately don't have the iOS emulator installed and was testing with a real(ly old) device 😅—glad to hear that this payed off for the latest versions, too.

PR #18358, opened 2019-10-09
docs: Update "Images, Files & Video" & link titles in sidebar

andersonnguyen:
## Description

Updated sidebar titles and breadcrumb titles of items listed under the "Images, Files & Video" section of the docs to adhere to the [Gatsby Style Guide](https://www.gatsbyjs.org/contributing/gatsby-style-guide/).

Changes include:

* Revising sidebar titles to use title case
* Revising breadcrumb titles to use title case
* Adding suggestions for breadcrumb titles where breadcrumb titles weren't already provided

## Related Issues

Related to #18284

### DIFF
--- a/www/src/data/sidebars/doc-links.yaml
+++ b/www/src/data/sidebars/doc-links.yaml
@@ -93,26 +93,32 @@
               breadcrumbTitle: ESLint
             - title: Proxying API Requests
               link: /docs/api-proxy/
-        - title: Images, Files, and Video
+        - title: Images, Files & Video
           link: /docs/images-and-files/
           items:
-            - title: Importing Assets Directly Into Files
+            - title: Importing Assets Directly into Files
               link: /docs/importing-assets-into-files/
-              breadcrumbTitle: Importing Assets Into Files
+              breadcrumbTitle: Importing Assets into Files
             - title: Using the Static Folder
               link: /docs/static-folder/
+              breadcrumbTitle: Static Folder
             - title: Using gatsby-image
               link: /docs/using-gatsby-image/
+              breadcrumbTitle: gatsby-image
             - title: Working with Images in Gatsby
               link: /docs/working-with-images/
-            - title: Preoptimizing your Images
+              breadcrumbTitle: Images in Gatsby
+            - title: Preoptimizing Your Images
               link: /docs/preoptimizing-images/
-            - title: Working With Video
+              breadcrumbTitle: Preoptimizing Images
+            - title: Working with Video
               link: /docs/working-with-video/
+              breadcrumbTitle: Video
             - title: Importing Media Content
               link: /docs/importing-media-content/
             - title: Working with GIFs
               link: /docs/working-with-gifs/
+              breadcrumbTitle: GIFs
             - title: Working with Images in Markdown
               link: /docs/working-with-images-in-markdown/
               breadcrumbTitle: Images in Markdown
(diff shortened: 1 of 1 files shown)


www/src/data/sidebars/doc-links.yaml
```diff
@@ -93,26 +93,32 @@
...
               link: /docs/working-with-images/
-            - title: Preoptimizing your Images
+              breadcrumbTitle: Images in Gatsby
+            - title: Preoptimizing Your Images
               link: /docs/preoptimizing-images/
-            - title: Working With Video
+              breadcrumbTitle: Preoptimizing Images
+            - title: Working with Video
               link: /docs/working-with-video/
+              breadcrumbTitle: Video
```
wardpeet:
```suggestion
              breadcrumbTitle: Video's in Gatsby
```

www/src/data/sidebars/doc-links.yaml
```diff
@@ -93,26 +93,32 @@
...
             - title: Importing Media Content
               link: /docs/importing-media-content/
             - title: Working with GIFs
               link: /docs/working-with-gifs/
+              breadcrumbTitle: GIFs
```
wardpeet:
```suggestion
              breadcrumbTitle: GIFs in Gatsby
```

www/src/data/sidebars/doc-links.yaml
```diff
@@ -93,26 +93,32 @@
...
+            - title: Importing Assets Directly into Files
               link: /docs/importing-assets-into-files/
-              breadcrumbTitle: Importing Assets Into Files
+              breadcrumbTitle: Importing Assets into Files
             - title: Using the Static Folder
               link: /docs/static-folder/
+              breadcrumbTitle: Static Folder
             - title: Using gatsby-image
               link: /docs/using-gatsby-image/
+              breadcrumbTitle: gatsby-image
```
wardpeet: Using the default title "Using gatsby-image" makes sense here in my opinion as gatsby-image is a bit bland.

```suggestion
```

www/src/data/sidebars/doc-links.yaml
andersonnguyen: Shouldn't the breadcrumb titles just be shorter names for the actual article title? If "Video" is not suitable as a breadcrumb title then I think we should just leave it to inherit the default title.

www/src/data/sidebars/doc-links.yaml
andersonnguyen: I was conflicted on this as well but there are other breadcrumb titles that are pretty short and bland, so I thought we should follow suit here. For example, "Customizing html.js" under the "Custom Configuration" section has a breadcrumb title of "html.js"

marcysutton: To chime in here, the reason for omitting words is to make them less redundant. “Html.js” in the case of the breadcrumb comes after “Custom configuration”, so the phrases work together. Hopefully that can act as a guide for how to work with breadcrumb titles.

andersonnguyen: Thanks for chiming in! That makes a lot more sense when you consider the entire breadcrumb chain. I'll revert it back to using the default title then as @wardpeet suggested.

www/src/data/sidebars/doc-links.yaml
andersonnguyen: Disregard the previous comment! I just realized the "in Gatsby" suggestion was to match the previous item's breadcrumb title (Images in Gatsby). I'll revise both the "Videos" and "GIFs" breadcrumb in a new commit shortly.

PR #18375, opened 2019-10-09
fix(docs): Update title casing

ziel-ar:
## Description

I updated all titles in 'docs' directory using Title Case.
There were cases where some titles had quotation marks, i removed them for unification.

## Related Issues
Related to #18284


docs/docs/add-custom-webpack-config.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: "Adding a Custom webpack Config"
+title: Adding a Custom Webpack Config
```
marcysutton: webpack as a formal noun isn't capitalized, so I think this one can be left as-is. The quotes don't actually change anything.

ziel-ar: webpack - all clear
quotes - yeah.. you know, perfectionist's hell ;)
one more thing that

PR #18386, opened 2019-10-09
feat(gatsby): adding ssr hooks for socket io() function

SamIngram96:
## Description

Adding an SSR hook to inject parameters into the socket io() initialization. This will allow users to control the socket configuration if they need to. original functionality is still intact and tests pass


SamIngram96:
I don't understand why these tests are failing, "'gatsby-transformer-remark' doesn't seem to be installed. Restart gatsby-dev to publish it" is not something I touched, yarn test completes without any errors

SamIngram96:
gonna try this again, integration tests are killing me

PR #18389, opened 2019-10-09
fix(www): Change of slider text to support light and dark mode

ruijdacd:
## Description

The color of the slider text was fixed on black, so the text would be impossible to read on dark mode.
This PR changes the color to `inherit`, so it can adapt to background changes alongside the text _"Need"_


ruijdacd:
Sure thing! Next time I’ll make sure I attach the screenshots

fk:
Thanks @ruijdacd and @m-allanson! 🙏

PR #18402, opened 2019-10-09
Docs/fix sourcing from prismic

jonniebigodes:
## Description
While triaging and trying to help a person with a issue related with gatsby and prismic, I followed the existing documentation for sourcing from prismic located [here](https://www.gatsbyjs.org/docs/sourcing-from-prismic/) reading up on it it came to my attention that is leading people into installing `dotenv`, but upon testing and checking the changes made to gatsby recently, the package is already present and it doesn't need to be added as extra dependency to the site.

With that i changed the wording in the documentation to reflect the change and removed code snippet for relevant to the `dotenv` package installation.


LekoArts:
Hi, thanks for opening the PR!
It doesn't hurt to tell people to install `dotenv` as it makes things more explicit (one might ask if `dotenv` would only be available in this starter?). We also do that in other guides through out the docs.

Thanks for the work you've done so far on issues/PRs, really appreciate it!

PR #18403, opened 2019-10-09
chore(showcase): Add tfe.energy

marcfehr:
## Description

Adding the freshly launched Gatsby website tfe.energy to the docs


docs/sites.yml
```diff
@@ -7670,3 +7670,17 @@
   built_by: Phil Tietjen
   built_by_url: "https://github.com/Phizzard"
   featured: false
+- title: TFE Energy
+  main_url: https://tfe.energy
+  url: https://tfe.energy
+  source_url: https://gitlab.com/marcfehrmedia/2019-07-03-tfe-energy
```
wardpeet: Seems like this URL isn't publicly available, can you remove it or make it public?

marcfehr: Oops, it's now public. Thanks.

marcfehr: > Seems like this URL isn't publicly available, can you remove it or make it public?

@wardpeet
Everything alright now?
Thanks :)

PR #18408, opened 2019-10-09
Title case, as per #18284

pixelflowxyz:
## Description

Changed title to the correct style

## Related Issues

#18284


pixelflowxyz:
AH! I just noticed that it's already all been done. Ouch!

PR #18410, opened 2019-10-09
Title case fix! #18284

pixelflowxyz:
Description
Changed title to the correct style - as per https://www.gatsbyjs.org/contributing/gatsby-style-guide/#format-titles-and-headers

Related Issues
#18284


pixelflowxyz:
AH! I just noticed that it's already all been done. Ouch!

PR #18411, opened 2019-10-09
Remove overflow-y style of the html in starter

oliviertassinari:
## Description

This change is meant to start a discussion around the need to add overflow-y on the `<html>` element. It's the first time I see this approach, and it breaks the scroll lock accessibility feature of https://material-ui.com/components/modal/. Is this style needed?

## Related Issues

Discovered in https://github.com/mui-org/material-ui/issues/17812.


KyleAMathews:
Hey! Thanks for raising this issue. This article explains why we added this https://css-tricks.com/snippets/css/force-vertical-scrollbar/

Vertical scrollbars appearing and disappearing is a very common source of layout "glitch" in SPAs including Gatsby sites which is why we add this. It's not a great solution as you point out as it sometimes breaks other things — but AFAIK, it's the only solution we have right now.

Feel free to remove it from your own sites but we'll leave it in the starter now as we think not having it will cause more problems than having it there.

oliviertassinari:
Ok, thanks for the update. I will see what we can do about it in Material-UI directly.

oliviertassinari:
A patch was released in Material-UI v4.5.2.

muescha:
with this patch in Material UI something need to change in gatsby or an css hack can be removed?

oliviertassinari:
@muescha What do you mean? Gatsby encourages overflow-y on the html element to avoid jumps, Material-UI's Modal set overflow: hidden on the html element when it opens.

muescha:
since here is mentioned [[Modal] Improve Gatsby support #17972](https://github.com/mui-org/material-ui/pull/17972)

PR #6245, opened 2018-07-01
First draft w/ alphabetical list

shannonbux:
no plugins category
how-to guides alphabetical

@calcsam @fk @m-allanson @KyleAMathews

Issue #18281, opened 2019-10-07
[docs] new doc on Winning Over Content Creators with Gatsby


marcysutton:
## Summary

We have some great docs on [Using Gatsby Professionally](https://www.gatsbyjs.org/docs/using-gatsby-professionally/), including [Winning Over (various) Stakeholders](https://www.gatsbyjs.org/docs/winning-over-stakeholders/). 

Through feedback, it was uncovered that the docs would benefit from an additional resource on Winning Over Content Creators. Given the potential for collaboration in using Gatsby with a Headless CMS, this seems like a great idea. A doc of this nature should talk about the benefits of using the Content Mesh with Gatsby and what a collaborative workflow using modern tools could look like.

For anyone interested in picking this up, please read through the existing Winning over Stakeholders docs as well as these resources to ensure your PR is accepted:

- [Gatsby Style Guide](https://www.gatsbyjs.org/contributing/gatsby-style-guide/)
- [Docs contributions](https://www.gatsbyjs.org/contributing/docs-contributions/), including updating the docs sidebar
- [Reference Guide template](https://www.gatsbyjs.org/contributing/docs-templates/#reference-guides)

DaniTulp:
Hi @marcysutton,

I'm working on an outline for this topic. Hopefully I will have a pull request up by tomorrow. If you have any suggestions please let me know!

Happy  #Hacktoberfest

marcysutton:
Closed with #18388.

PR #18421, opened 2019-10-09
docs: add issues to image heavy site stub

aditsachde:
## Description

There are a couple stubs in in documents labeled `building`, without issues. I'm not sure what the purpose of them were, but they were added in #6245. That PR does not mention a specific issue that prompted the addition of the stubs, but it seems that the build system relies on the issue field to check if a doc is a stub.

They could be removed, but the topics do seem useful to have a guide on.

## Related Issues

Related to #18284


docs/docs/building-a-blog.md
```diff
@@ -1,5 +1,6 @@
 ---
 title: Building a blog
+issue: https://github.com/gatsbyjs/gatsby/pull/6245
```
marcysutton: Stubs don't link to closed PRs, since there isn't a clear process for what a contributor should do with it. A better approach would be to file individual issues for each of these stubs so the scope of each one is limited to a single doc and people can pick them up one-by-one. If you have time to do that, it would be amazing and here's an example of it! https://github.com/gatsbyjs/gatsby/issues/18281 

Outside of that, I don't think we should merge this as-is with links to an old PR. So let us know what you have time to do!

aditsachde:
I completely agree with that, new issues would make things much more clear. I changed this PR to address just `building-an-image-heavy-site.md` and opened #18423 for it. I'll probably have more time to look over these and research them over the weekend, but I went ahead and did that page as I knew some of the things to mention.

I'm alright if you would like open this up to the community as I'm not sure I'll be able to get to them for a bit.

aditsachde:
Added #18485 and #18484 last night

PR #18428, opened 2019-10-10
chore(docs): Update pronouns on "Behind the scenes"

LuisPinedaJR:
## Description

Updated docs/docs/query-extraction.md

## Related Issues

 Gatsby Style Guide blitz for those new to open source #18284


docs/docs/query-behind-the-scenes.md
```diff
@@ -2,9 +2,9 @@
 title: How Queries Work
 ---
 
-We're talking about GraphQL queries here. These can be tagged graphql expressions at the bottom of your component source file (e.g. [query for Gatsby frontpage](https://github.com/gatsbyjs/gatsby/blob/master/www/src/pages/index.js#L165)), StaticQueries within your components (e.g. [showcase site details](https://github.com/gatsbyjs/gatsby/blob/master/www/src/components/showcase-details.js#L103)), or fragments created by plugins (e.g. [gatsby-source-contentful](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-source-contentful/src/fragments.js)).
+You're talking about GraphQL queries here. These can be tagged graphql expressions at the bottom of your component source file (e.g. [query for Gatsby frontpage](https://github.com/gatsbyjs/gatsby/blob/master/www/src/pages/index.js#L165)), StaticQueries within your components (e.g. [showcase site details](https://github.com/gatsbyjs/gatsby/blob/master/www/src/components/showcase-details.js#L103)), or fragments created by plugins (e.g. [gatsby-source-contentful](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-source-contentful/src/fragments.js)).
```
sidharthachatterjee: This reads weird. Maybe we need an exception in the style guide? 🤷‍♂ 

@shannonbux @marcysutton

marcysutton: It's somewhat subjective how to structure sentences, but I agree that this doesn't quite work here (the language used was odd to begin with). I'd recommend restructuring this and the following paragraphs to not talk about "talking", but rather present the information more directly. In some cases "we" means Gatsby's code, so it would be more appropriate to talk about the code as a subject since it's doing the action (not the user/developer).

Here's a recommendation:

```suggestion
As part of Gatsby's data layer, GraphQL queries can be specified as tagged graphql expressions at the bottom of your component source file (e.g. [query for Gatsby frontpage](https://github.com/gatsbyjs/gatsby/blob/master/www/src/pages/index.js#L165)), StaticQueries within your components (e.g. [showcase site details](https://github.com/gatsbyjs/gatsby/blob/master/www/src/components/showcase-details.js#L103)), or fragments created by plugins (e.g. [gatsby-source-contentful](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-source-contentful/src/fragments.js)).
```

docs/docs/query-behind-the-scenes.md
```diff
@@ -2,9 +2,9 @@
...
 
-Note that we are NOT talking about queries involved in the creation of your pages, which is usually performed in your site's gatsby-node.js (e.g. [Gatsby's website](https://github.com/gatsbyjs/gatsby/blob/master/www/gatsby-node.js#L165)). We're only talking about queries that are tied to particular pages or templates.
+Note that you are NOT talking about queries involved in the creation of your pages, which is usually performed in your site's gatsby-node.js (e.g. [Gatsby's website](https://github.com/gatsbyjs/gatsby/blob/master/www/gatsby-node.js#L165)). You're only talking about queries that are tied to particular pages or templates.
```
marcysutton:
```suggestion
Note that this process only applies to queries that are specified directly in components or templates. It doesn't apply to queries involved in the creation of dynamic pages through your site's `gatsby-node.js` file (e.g. on [Gatsby's website](https://github.com/gatsbyjs/gatsby/blob/master/www/gatsby-node.js#L165)).
```

docs/docs/query-execution.md
```diff
@@ -87,15 +87,15 @@ All queries queued after being extracted (from `query-watcher.js`).
...
+All queries whose component path isn't listed in `componentDataDependencies`. As a recap, in [Schema Generation](/docs/schema-generation/), you've seen that all Type resolvers record a dependency between the page whose query you're running and any nodes that were successfully resolved. So, If a component is declared in the `components` redux namespace (occurs during [Page Creation](/docs/page-creation/)), but is _not_ contained in `componentDataDependencies`, then by definition, the query has not been run. Therefore you need to run it. Checkout [Page -> Node Dependencies](/docs/page-node-dependencies/) for more info. The code for this step is in [findIdsWithoutDataDependencies](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/page-query-runner.js#L96).
```
marcysutton:
```suggestion
All queries whose component path isn't listed in `componentDataDependencies`. In [Schema Generation](/docs/schema-generation/), all Type resolvers record a dependency between the page whose query is running and any nodes that were successfully resolved. So, If a component is declared in the `components` redux namespace (occurs during [Page Creation](/docs/page-creation/)), but is _not_ contained in `componentDataDependencies`, then by definition, the query has not been run. Therefore it needs to be run. Checkout [Page -> Node Dependencies](/docs/page-node-dependencies/) for more info. The code for this step is in [findIdsWithoutDataDependencies](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/page-query-runner.js#L96).
```

docs/docs/query-execution.md
```diff
@@ -87,15 +87,15 @@ All queries queued after being extracted (from `query-watcher.js`).
...
+In `develop` mode, every time a node is created, or is updated (e.g. via editing a markdown file), you add that node to the [enqueuedDirtyActions](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/page-query-runner.js#L61) collection. When you execute your queries, you can lookup all nodes in this collection and map them to pages that depend on them (as described above). These pages' queries must also be executed. In addition, this step also handles dirty `connections` (see [Schema Connections](/docs/schema-connections/)). Connections depend on a node's type. So if a node is dirty, you mark all connection nodes of that type dirty as well. The code for this step is in [findDirtyIds](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/page-query-runner.js#L171). _Note: dirty ids is really talking about dirty paths_.
```
marcysutton: The user isn't doing the action here as far as I know....
```suggestion
In `develop` mode, every time a node is created, or is updated (e.g. via editing a markdown file), that node needs to be dynamically added to the [enqueuedDirtyActions](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/page-query-runner.js#L61) collection. When your queries are executed, the code will look up all nodes in this collection and map them to pages that depend on them (as described above). These pages' queries must also be executed. In addition, this step also handles dirty `connections` (see [Schema Connections](/docs/schema-connections/)). Connections depend on a node's type. So if a node is dirty, the code marks all connection nodes of that type dirty as well. The code for this step is in [findDirtyIds](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/page-query-runner.js#L171). _Note: dirty ids is really talking about dirty paths_.
```

docs/docs/query-execution.md
```diff
@@ -87,15 +87,15 @@ All queries queued after being extracted (from `query-watcher.js`).
...
 
 #### Queue Queries for Execution
 
-We now have the list of all pages that need to be executed (linked to their Query information). Let's queue them for execution (for realz this time). A call to [runQueriesForPathnames](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/page-query-runner.js#L127) kicks off this step. For each page or static query, we create a Query Job that looks something like:
+You now have the list of all pages that need to be executed (linked to their Query information). Let's queue them for execution (for realz this time). A call to [runQueriesForPathnames](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/page-query-runner.js#L127) kicks off this step. For each page or static query, you create a Query Job that looks something like:
```
marcysutton:
```suggestion
There is now a list of all pages that need to be executed (linked to their Query information). Gatsby will queue them for execution (for realz this time). A call to [runQueriesForPathnames](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/page-query-runner.js#L127) kicks off this step. For each page or static query, Gatsby creates a Query Job that looks something like:
```

docs/docs/query-execution.md
```diff
@@ -113,17 +113,17 @@ We now have the list of all pages that need to be executed (linked to their Quer
 }
 ```
 
-This Query Job contains everything we need to execute the query (and do things like recording dependencies between pages and nodes). So, we push it onto the queue in [query-queue.js](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/query-queue.js) and then wait for the queue to empty. Let's see how `query-queue` works.
+This Query Job contains everything you need to execute the query (and do things like recording dependencies between pages and nodes). So, you push it onto the queue in [query-queue.js](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/query-queue.js) and then wait for the queue to empty. Let's see how `query-queue` works.
```
marcysutton:
```suggestion
This Query Job contains everything it needs to execute the query (and do things like recording dependencies between pages and nodes). It gets pushed onto the queue in [query-queue.js](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/query-queue.js) and then waits for the queue to empty. Next, this doc will cover how `query-queue` works.
```

docs/docs/query-execution.md
```diff
@@ -113,17 +113,17 @@ We now have the list of all pages that need to be executed (linked to their Quer
...
 
-[query-queue.js](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/query-queue.js) creates a [better-queue](https://www.npmjs.com/package/better-queue) queue that offers advanced features like parallel execution, which is handy since queries do not depend on each other so we can take advantage of this. Every time an item is consumed from the queue, we call [query-runner.js](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/query-runner.js) where we finally actually execute the query!
+[query-queue.js](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/query-queue.js) creates a [better-queue](https://www.npmjs.com/package/better-queue) queue that offers advanced features like parallel execution, which is handy since queries do not depend on each other so you can take advantage of this. Every time an item is consumed from the queue, you call [query-runner.js](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/query-runner.js) where you finally actually execute the query!
```
marcysutton:
```suggestion
[query-queue.js](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/query-queue.js) creates a [better-queue](https://www.npmjs.com/package/better-queue) queue that offers advanced features like parallel execution, which is handy since queries do not depend on each other so Gatsby can take advantage of this. Every time an item is consumed from the queue, it calls [query-runner.js](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/query-runner.js) where it can finally execute the query!
```

docs/docs/query-execution.md
```diff
@@ -113,17 +113,17 @@ We now have the list of all pages that need to be executed (linked to their Quer
...
 
 Query execution involves calling the [graphql-js](https://graphql.org/graphql-js/) library with 3 pieces of information:
 
 1. The Gatsby schema that was inferred during [Schema Generation](/docs/schema-generation/).
 1. The raw query text. Obtained from the Query Job.
-1. The Context, also from the Query Job. Has the page's `path` amongst other things so that we can record [Page -> Node Dependencies](/docs/page-node-dependencies/).
+1. The Context, also from the Query Job. Has the page's `path` amongst other things so that you can record [Page -> Node Dependencies](/docs/page-node-dependencies/).
```
marcysutton:
```suggestion
1. The Context, also from the Query Job. Has the page's `path` amongst other things so that Gatsby can record [Page -> Node Dependencies](/docs/page-node-dependencies/).
```

docs/docs/query-execution.md
```diff
@@ -137,6 +137,6 @@ As queries are consumed from the queue and executed, their results are saved to
 /public/static/d/621/path---blog-2018-07-17-announcing-gatsby-preview-995-a74-dwfQIanOJGe2gi27a9CLKHjamc.json
 ```
 
-For static queries, instead of using the page's jsonName, we just use a hash of the query.
+For static queries, instead of using the page's jsonName, you just use a hash of the query.
```
marcysutton:
```suggestion
For static queries, instead of using the page's jsonName, Gatsby uses a hash of the query.
```

docs/docs/query-execution.md
```diff
@@ -137,6 +137,6 @@ As queries are consumed from the queue and executed, their results are saved to
...
 
-Now we need to store the association of the page -> the query result in redux so we can recall it later. This is accomplished via the [json-data-paths](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/redux/reducers/json-data-paths.js) reducer which we invoke by creating a `SET_JSON_DATA_PATH` action with the page's jsonName and the saved dataPath.
+Now you need to store the association of the page -> the query result in redux so we can recall it later. This is accomplished via the [json-data-paths](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/redux/reducers/json-data-paths.js) reducer which you invoke by creating a `SET_JSON_DATA_PATH` action with the page's jsonName and the saved dataPath.
```
marcysutton:
```suggestion
Now Gatsby needs to store the association of the page -> the query result in redux so it can be recalled later. This is accomplished via the [json-data-paths](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/redux/reducers/json-data-paths.js) reducer which is invoked by creating a `SET_JSON_DATA_PATH` action with the page's jsonName and the saved dataPath.
```

docs/docs/query-extraction.md
```diff
@@ -13,7 +13,7 @@ title: Query Extraction
...
+Up until now, you have [sourced all nodes](/docs/node-creation/) into redux, [inferred a schema](/docs/schema-generation/) from them, and [created all pages](/docs/page-creation/). The next step is to extract and compile all graphql queries from your source files. The entrypoint to this phase is [query-watcher extractQueries()](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/query-watcher.js), which immediately compiles all graphql queries by calling into [query-compiler.js](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/query-compiler.js).
```
marcysutton:
```suggestion
Up until now, Gatsby has [sourced all nodes](/docs/node-creation/) into redux, [inferred a schema](/docs/schema-generation/) from them, and [created all pages](/docs/page-creation/). The next step is to extract and compile all graphql queries from your source files. The entrypoint to this phase is [query-watcher extractQueries()](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/query-watcher.js), which immediately compiles all graphql queries by calling into [query-compiler.js](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/query-compiler.js).
```

docs/docs/query-extraction.md
```diff
@@ -22,7 +22,7 @@ The first thing it does is use [babylon-traverse](https://babeljs.io/docs/en/nex
 1. It informs us of any malformed queries, which are promptly reported back to the user.
 1. It builds a tree of queries and fragments they depend on. And outputs a single optimized query string with the fragments.
 
-After this step, we will have a map of file paths (of site files with queries in them) to Query Objects, which contain the raw optimized query text, as well as other metadata such as the component path and page `jsonName`. The following diagram shows the flow involved during query compilation
+After this step, you will have a map of file paths (of site files with queries in them) to Query Objects, which contain the raw optimized query text, as well as other metadata such as the component path and page `jsonName`. The following diagram shows the flow involved during query compilation
```
marcysutton:
```suggestion
After this step, Gatsby will have a map of file paths (of site files with queries in them) to Query Objects, which contain the raw optimized query text, as well as other metadata such as the component path and page `jsonName`. The following diagram shows the flow involved during query compilation
```

docs/docs/query-extraction.md
```diff
@@ -59,11 +59,11 @@ digraph {
 
 #### Store Queries in Redux
 
-We're now in the [handleQuery](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/query-watcher.js#L68) function.
+You're now in the [handleQuery](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/query-watcher.js#L68) function.
```
marcysutton:
```suggestion
Gatsby is now in the [handleQuery](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/query-watcher.js#L68) function.
```

docs/docs/query-extraction.md
```diff
@@ -59,11 +59,11 @@ digraph {
...
 
-If the query is a `StaticQuery`, we call the `replaceStaticQuery` action to save it to to the `staticQueryComponents` namespace which is a mapping from a component's path to an object that contains the raw GraphQL Query amongst other things. More details in [Static Queries](/docs/static-vs-normal-queries/). We also remove component's `jsonName` from the `components` Redux namespace. See [Page -> Node Dependencies](/docs/page-node-dependencies/).
+If the query is a `StaticQuery`, you call the `replaceStaticQuery` action to save it to to the `staticQueryComponents` namespace which is a mapping from a component's path to an object that contains the raw GraphQL Query amongst other things. More details in [Static Queries](/docs/static-vs-normal-queries/). We also remove component's `jsonName` from the `components` Redux namespace. See [Page -> Node Dependencies](/docs/page-node-dependencies/).
```
marcysutton:
```suggestion
If the query is a `StaticQuery`, Gatsby will call the `replaceStaticQuery` action to save it to the `staticQueryComponents` namespace which is a mapping from a component's path to an object that contains the raw GraphQL Query amongst other things. More details can be found in the doc on [Static Queries](/docs/static-vs-normal-queries/). Gatsby also removes a component's `jsonName` from the `components` Redux namespace. See [Page -> Node Dependencies](/docs/page-node-dependencies/).
```

docs/docs/query-extraction.md
```diff
@@ -59,11 +59,11 @@ digraph {
...
 
-If the query is just a normal every day query (not StaticQuery), then we update its component's `query` in the redux `components` namespace via the [replaceComponentQuery](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/redux/actions.js#L827) action.
+If the query is just a normal every day query (not StaticQuery), then you update its component's `query` in the redux `components` namespace via the [replaceComponentQuery](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/redux/actions.js#L827) action.
```
marcysutton:
```suggestion
If the query is just a normal every-day query (not StaticQuery), then Gatsby updates its component's `query` in the redux `components` namespace via the [replaceComponentQuery](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/redux/actions.js#L827) action.
```

docs/docs/query-extraction.md
```diff
@@ -91,7 +91,7 @@ digraph {
 
 #### Queue for execution
 
-Now that we've saved our query, we're ready to queue it for execution. Query execution is mainly handled by [page-query-runner.js](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/page-query-runner.js), so we accomplish this by passing the component's path to `queueQueryForPathname` function.
+Now that you've saved your query, you're ready to queue it for execution. Query execution is mainly handled by [page-query-runner.js](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/page-query-runner.js), so we accomplish this by passing the component's path to `queueQueryForPathname` function.
```
marcysutton:
```suggestion
Now that Gatsby has saved your query, it's ready to queue for execution. Query execution is mainly handled by [page-query-runner.js](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/page-query-runner.js), so it accomplishes this by passing the component's path to `queueQueryForPathname` function.
```

marcysutton:
looks like there are still a few comments that need to be addressed, as an FYI!

docs/docs/query-execution.md
```diff
@@ -113,17 +113,17 @@ We now have the list of all pages that need to be executed (linked to their Quer
 }
 ```
 
-This Query Job contains everything we need to execute the query (and do things like recording dependencies between pages and nodes). So, we push it onto the queue in [query-queue.js](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/query-queue.js) and then wait for the queue to empty. Let's see how `query-queue` works.
+This Query Job contains everything it needs to execute the query (and do things like recording dependencies between pages and nodes). It gets pushed onto the queue in [query-queue.js](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/query-runner/query-queue.js) and then waits for the queue to empty. Next, this doc will cover how `query-queue` works.
```
shannonbux: nice changes!

docs/docs/query-behind-the-scenes.md
shannonbux: oh yes, I can see how it would be easy to make this mistake since the style guide says "change 'we're' to 'you're.'" This is a good example of when it's not necessary to mention either we or you at all; it's more accurate and efficient to just talk about Gatsby.

Issue #17943, opened 2019-09-27
[gatsby-source-wordpress] add config option to manipulate list of endpoint we fetch


pieh:
There is need for users to be able to manipulate list of endpoints we get from wp rest introspection endpoint. This includes setting query params on routes and adding/removing routes.

Use cases for it are:
 - adding support for WPML translations (see #10915, #3263, #1648 ) (setting `lang` query param on post types endpoints + hitting them for each language)
 - being able to pull in draft / private posts ( https://github.com/gatsbyjs/gatsby/pull/10942#issuecomment-534182374 )

There were few attempts to implement this ( https://github.com/gatsbyjs/gatsby/pull/10942 , https://github.com/gatsbyjs/gatsby/pull/10335 ), but those implementations didn't allow for general purpose, robust system (they were mostly geared toward fixing particular use case)

I'd like to propose solution that is similar to wordpress filters, where user will be able to programatically manipulate list of endpoints with config option (and is similar to how `normalizer` config option works), where it's optional function that get list of endpoints, do some processing on it and return modified list to wordpress plugin:

```
{
      resolve: "gatsby-source-wordpress",
      options: {
        // [...] rest of config
        routesFilter: async ({ routes }) => {
         // do some work on routes array
          return routes
        }
      }
```

`routesFilter` is not final name of the config option - please comment on better name for it.

`routes` would be array of objects with following fields:
```js
{
  // required fields
  "url", "http://example.com/wp-json/wp/v2/pages", // url to endpoint - possibly this wouldn't be full absolute url and just "/wp/v2/pages" so users don't have to have special cases for wordpress.com hosting which have a bit different endpoints than user hosted wordpress instance
  "type": "wordpress__PAGE", // type used to generated gatsby data nodes
  // optional fields
  "queryParams": {
    "lang": "es"
  }, // optional object with params that will be used when fetching
  "normalizer": ({ entities }) => {
    // manipulate entites that were created after fetching this route - useful to set additional meta fields (like lang)
    return entities
  }, // optional function
}
```

Ultimate goal here is to create API that later on can be packaged into reusable `gatsby-source-wordpress` subplugins (see https://github.com/gatsbyjs/rfcs/pull/23 ). Creating subplugin system wouldn't be in scope for initial implementation. This features is just first step on path to be able to do this.

## Implementation:

[...]

PR #18436, opened 2019-10-10
feat(gatsby-source-wordpress): get posts by all statuses when authenticated with JWT

kepi0809:
Needed to get draft post from WP too, so added status=any to the URL in the source WP plugin.

the hardcoded endpoint may be not the best solution if you have a proposal fixing that I'll change it.

## Description

Get posts with all status (even custom ones)

### DIFF
--- a/packages/gatsby-source-wordpress/src/fetch.js
+++ b/packages/gatsby-source-wordpress/src/fetch.js
@@ -349,11 +349,13 @@ async function getPages(
     let result = []
 
     const getOptions = page => {
+      const hasPrivilage = _accessToken && url.includes("wp/v2/posts")
       let o = {
         method: `get`,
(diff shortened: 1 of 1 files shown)


wardpeet:
Hi,

Thanks for opening a pull request but could you tell us your use-case and reasoning for wanting to get all statuses?

pieh:
Hey, please check https://github.com/gatsbyjs/gatsby/issues/17943 for way we want to enable use cases like being able to pull drafts. We won't be merging this one. 

This will also be breaking change, because people would need to add `status` filters in their production builds if they need authentication - so we would need to bump major version, and it's not the way we want to solve this problem so that would mean we would have a lot of major version bumps while we get to point where we want to be. Giving users ability to set their own query params (and additional endpoints plugin hit) is most robust solution here.

PR #18441, opened 2019-10-10
feat(gatsby-plugin-mdx): src/pages shadowing

### LINKED ISSUE #16845: [gatsby-plugin-mdx]: default queries in src/pages
## Summary

gatsby-plugin-mdx has a `defaultLayouts` option that is used to inject MDX `wrapper`s into `src/pages`. In addition to that, we should enable a `defaultQuery` so that people can define a query to be associated with a wrapper.

### Basic example

One possible, backwards compatible, way forward on this is:

```js
{
  resolve: `gatsby-plugin-mdx`,
  options: {
    defaultLayouts: {
      default: {
        component: path.resolve('./src/components/whatever.js'),
        query: `query { mdx { ... } }`
      }
    }
  }
}
```

### Motivation

For the same reason people like the default `src/pages` support layout defaults across a set of pages, people want to build table of contents' without needing content authors to worry about including the query.

---

This also brings to light a potential bug where pageQueries could be added to require'd .mdx files.

ported from gatsby-mdx repo: https://github.com/ChristopherBiscardi/gatsby-mdx/issues/213

### LINKED ISSUE #16848: [gatsby-plugin-mdx] defaultLayouts should be defaultWrappers
The `defaultLayouts` option is no longer accurate naming for MDX, which merged `layouts` and `wrappers` into the same thing in 1.0.

`defaultLayouts` should become `defaultWrappers` so as to align with mdx core and also avoid confusion with Gatsby page layouts.

We might also want take this opportunity to look at the gatsby-plugin-mdx options and re-evaluate if the API is still serving us well. Renaming would be a breaking change (even though we could have some backwards compat).

ChristopherBiscardi:
## Description

Get rid of old `defaultLayouts` API in favor of taking advantage of shadowing API. This removes a "special" API in favor of using a core Gatsby API.

## Related Issues

Fixes #16848
Fixes #16845

TODO: 

* new docs, remove old docs
* `page-creator` seems to override `createPages`. How can we re-order that?

image


ChristopherBiscardi:
The main thing blocking this PR is that the default set of page-creator-plugin's added to src/pages overrides the createPages calls. It's possible this can be fixed with a clever onCreatePage call?

pieh:
> The main thing blocking this PR is that the default set of page-creator-plugin's added to src/pages overrides the createPages calls. It's possible this can be fixed with a clever onCreatePage call?

It feels like plugins (and users) should have easy way of disabling/overriding settings that default page-creator instance.

Recently there was added code to enable overriding default one for Site ( https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/bootstrap/load-plugins/load.js#L222-L240 ), it feels like we should do the same for those added to plugins - possibly having `disable` option or something to not even add it to plugin list, but overriding with `ignore` pattern that will match everything would work here (downside is that `gatsby-plugin-mdx` will need to bump peerDeps of gatsby version probably)

ChristopherBiscardi:
@pieh yeah, if there was an ability to tell the page-creator plugin to opt-out of handling .mdx files that would be ideal I think. We need the `resolvableExtensions` to include `.mdx` so that people can `require('./thing')` where `./thing.mdx` exists, since it compiles to a .js file and should be usable as such.

> It feels like plugins (and users) should have easy way of disabling/overriding settings that default page-creator instance.

This feels related to the themes need for plugin option modification. `onPluginOptions` could be a lifecycle that works on individual plugins and we could avoid solving the plugin-merging problem for now. Would that be an acceptable path to pursue? Seems like it could be called in bootstrap when plugins are originally required.

pieh:
`onPluginOptions` sound interesting and less single-purpose feature, I'm just not sure how this should work - could you comment how do you see this working?

We should also keep in mind all the extra inertia that comes with designing new APIs/extension points can make this quite a long process (and just not feasible for now)

ChristopherBiscardi:
@pieh yeah I understand the intertia here and that this would cause this PR to sit open while that happened. I'm going to look into the potential onCreatePage solution here too as a short-term solution.

## onPluginOptions

`onPluginOptions` would get called by each plugin in `flattenedPlugins` if defined. This means the results would have to be idempotent. Themes/plugins/user-site would have the opportunity to accept the options for a plugin and modify them and only them. So it's single options object in, single options object out.

```js
exports.onPluginOptions = ({ options, plugin }, pluginOptions) => {
  if (
    plugin === `gatsby-plugin-page-creator` &&
    options.path.endsWith("src/pages")
  ) {
    return {
      ...options,
      ignore: { patterns: [`**/*.mdx`] }
    };
  }
};
```

I recognize that this is an informal proposal and hasn't had a lot of the edge cases thought through like they would need to be if this became an RFC. This illustrates the general idea though. I think the RFC would have to address which "plugin merging" capabilities are being excluded from the proposal and why so as to address long-term compatibility.

pieh:
Also consider not using `resolvableExtensions` extension point and maybe trying to add those extensions directly via `onCreateWebpackConfig` - this should solve the problem of `gatsby-plugin-page-creator` picking up .md / .mdx files while still allowing to do those imports?

ChristopherBiscardi:
oh right, that would work and because themes grabs the extensions via the webpack config directly I think it would disable them for page-creator while keeping them enabled for theming. thanks!

johno:
I'm going to go ahead and close this for now, and revive when we begin working towards a gatsby-plugin-mdx v2.

💜💜💜

PR #18445, opened 2019-10-10
chore(docs): title casing and grammar fixes

FissionCat:
## Description

Style guide-related fixes for a handful of docs files.

## Related Issues

Addresses #18284

### DIFF
--- a/docs/docs/static-query.md
+++ b/docs/docs/static-query.md
@@ -1,10 +1,10 @@
 ---
-title: Querying data in components using StaticQuery
+title: Querying Data in Components using StaticQuery
 ---
 
-Gatsby v2 introduces `StaticQuery`, a new API that allows components to retrieve data via GraphQL query.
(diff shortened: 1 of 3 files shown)


docs/docs/static-query.md
```diff
@@ -41,11 +41,11 @@ export default () => (
 )
 ```
 
-Using `StaticQuery`, you can colocate a component with its data. No longer is it required to, say, pass data down from `Layout` to `Header`.
+By using `StaticQuery`, you can colocate a component with its data. It is no longer required to, say, pass data down from `Layout` to `Header`.
 
 ### useStaticQuery
 
-There's also a hooks version of StaticQuery called `useStaticQuery`. Check out its documentation [here!](/docs/use-static-query/)
+There's also a hooks version of StaticQuery called `useStaticQuery`. [Check out its documentation here!](/docs/use-static-query/)
```
marcysutton: Here's a suggestion!
```suggestion
There's also a React hooks version of StaticQuery: check out the documentation on [`useStaticQuery`](/docs/use-static-query/)
```

PR #18465, opened 2019-10-10
Support passing onClick to gatsby-image

### LINKED ISSUE #4736: Adding event handlers to gatsby-image <Img /> elements
## Description

Is it possible to add React event handler functionality via props to the  element? I referenced the doc on gatsby-image as well as the sharp plugin and no event handlers are discussed or listed in the props table.

Is the current default implementation to wrap the  element inside another element (i.e. div) and add the event handler to the wrapper? This worked fine for me but I wanted to clarify if it was possible to directly add event props to  or if it were coming in a future release.
 
i.e. 

### Steps to reproduce

I tried inlining an alert (as above) as well as writing a simple event handler to be called but neither works. I wrapped the  in a div and added the click handler to it and this worked as expected.

### Expected result

The React event would fire as normal when clicking on the .

### Actual result

Nada.

### Environment

Added the images and handler code inside the about.js file of the 4th tutorial project. Not attempting in actual project at the moment.

### File contents (if changed):

None

KyleAMathews:
Fix #4736

### DIFF
--- a/packages/gatsby-image/README.md
+++ b/packages/gatsby-image/README.md
@@ -414,6 +414,7 @@ While you could achieve a similar effect with plain CSS media queries, `gatsby-i
 | `fadeIn`               | `bool`              | Defaults to fading in the image on load                                                                                                       |
 | `durationFadeIn`       | `number`            | fading duration is set up to 500ms by default                                                                                                 |
 | `title`                | `string`            | Passed to the `img` element                                                                                                                   |
+| `onClick`              | `func`              | Passed to the `img` element                                                                                                                   |
 | `alt`                  | `string`            | Passed to the `img` element. Defaults to an empty string `alt=""`                                                                             |
 | `crossOrigin`          | `string`            | Passed to the `img` element                                                                                                                   |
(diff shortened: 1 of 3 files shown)


wardpeet:
Can you update [typings file](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-image/index.d.ts) and [README.md](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-image/README.md) to reflect the onClick prop.

KyleAMathews:
@wardpeet on it 👍

gillkyle:
Realizing we duplicate that props table in the [Gatsby Image API](https://www.gatsbyjs.org/docs/gatsby-image/#gatsby-image-props) doc as well. This might be a good place to split that table out into a separate MDX file and import it in these 2 files but as a quick solution I put up a Good First Issue to document this in that doc as well here: #18468

marcysutton:
Did this end up getting reverted @KyleAMathews @gillkyle ?

gillkyle:
It doesn't look like it was @marcysutton, I still see it in the history 🤔 : https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-image/src/index.js#L256

Issue #17980, opened 2019-09-28
[bug] rendering 404 instead of `/` or localized `/en/` or `/ru/` pages after updating gatsby from 2.15.22 to 2.15.23 or newer


JustFly1984:
## Description

We have an application with internationalization support. All pages created with gatsby-node createPages api. With garsby@2.15.22 and older there is normal behavior, but after update to 2.15.23 or newer (2.15.28), root page and localized versions of the root page displaying 404 template content. 

All other pages displayed correctly. If I roll back to  2.15.22 root page and localized versions displayed properly.

This incorrect behavior is consistent in development and production.

I do not have steps to reproduce, but expect other people to have same issue.

### Environment

System:
    OS: macOS 10.14.6
    CPU: (8) x64 Intel(R) Core(TM) i7-4960HQ CPU @ 2.60GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 12.9.0 - ~/.nvm/versions/node/v12.9.0/bin/node
    Yarn: 1.17.3 - /usr/local/bin/yarn
    npm: 6.10.2 - ~/.nvm/versions/node/v12.9.0/bin/npm
  Languages:
    Python: 2.7.15 - /usr/local/bin/python
  Browsers:
    Chrome: 77.0.3865.90
    Firefox: 69.0
    Safari: 13.0.1

JustFly1984:
currently bug persists in 2.15.28

sever1an:
Do you have a GitHub of the affected project?

JustFly1984:
I have project in private repo under NDA (

I have found commit which supposedly breaks my dev and build

this is screenshots of 2.15.22 build and 2.15.23:
<img width="713" alt="Screenshot 2019-09-29 02 43 28" src="https://user-images.githubusercontent.com/2255993/65823668-3e842c80-e263-11e9-9381-2508d4b7b2c9.png">
<img width="716" alt="Screenshot 2019-09-29 02 43 40" src="https://user-images.githubusercontent.com/2255993/65823669-3e842c80-e263-11e9-80ab-423568f74f2d.png">

They both shows that 404 and index.html build correctly. if I run gatsby serve, I see correct content for a second, and it rerenders ad 404 content after a moment.

<img width="784" alt="Screenshot 2019-09-29 02 53 10" src="https://user-images.githubusercontent.com/2255993/65823717-4e504080-e264-11e9-896c-c509f44589c0.png">

[...]

PR #18478, opened 2019-10-10
fix(gatsby): Fix matchpath ordering

sever1an:
## Description
Path matching should always chose the most specific match, such that paths like `/app/static-page` should be chosen over `/app/*`. There is presently a regression where `/app/*`  is chosen over `/app/<index.js>`, which is actually just `/app/`.

In this commit ( https://github.com/gatsbyjs/gatsby/commit/6c64948edf547b59bf68133956f6be55954f7735#diff-ac5196f8e8e35e4455b30b9164e3c45b ), the way static pages are handled within match paths was changed, such that any static page within the domain of a glob (`*`) match path is given its own entry in `.cache/match-paths.json`. To allow for static pages to take precedence the over glob paths, the order of this file is important... more specific routes must be tested before less specific ones. 

The order of entries in `.cache/match-paths.json` is ultimately determined by `packages/gatsby/src/bootstrap/requires-writer.js`.

Paths are assigned a score based on how specific they are by basically counting slashes, but the trailing slash is removed first. For an index page the trailing slash is significant, as it implies /index. Globs like `/*` and `/app/*` initially score one point higher than `/` and `/app/` then are equal after a bonus for no `*`. Yes, this might even present undefined behavior. Removing a trailing `*` along with the trailing `/` puts glob paths at the equivalent level of an index page with removed slash instead of a child page during the initial scoring, then the index gains a point for not having a `*`. I believe that this always places glob paths after static pages that they might shadow. It should also maintain the relative order of specificity among different glob paths. For example `/app/stuff/*` should be chosen over `/app/*` and `/*`.

## Related Issues

I believe that this a regression which will break existing apps when `gatsby` is updated. The author of 17980 discovered it with CI testing.

Fixes https://github.com/gatsbyjs/gatsby/issues/17980


wardpeet:
Hey, codewise this looks good. I couldn't reproduce the bug. Do you have a reproduction?

sever1an:
I have created two reproductions

This one shows an error on the path `/app/` https://github.com/sever1an/match-path-error/tree/folder-error

This one shows an  error on `/` and generates the wrong `match-paths.json` , but for some reason the `/*` route matches everything with `gatsby serve`
https://github.com/sever1an/match-path-error/tree/root-error

```javascript
[
    {
        "path": "/404.html",
        "matchPath": "/404.html"
    },
    {
        "path": "/zapp-glob/",
        "matchPath": "/app/*"
    },
    {
        "path": "/app/",
        "matchPath": "/app/"
    },
    {
        "path": "/404/",
        "matchPath": "/404/"
    },
    {
        "path": "/root-glob/",
        "matchPath": "/*"
    },
    {
        "path": "/",
        "matchPath": "/"
    }
]
```

sever1an:
Having the `/*` match everything in the root-error branch is bothering me a bit, any ideas why this would happen?

JustFly1984:
Is there any movement?

sever1an:
@wardpeet Were you able to use the reproduction?

JustFly1984:
Can you please merge this PR and publish?

JustFly1984:
I'm sorry to buzz, but 5 of my projects having bugs, which will be fixed with this commit

sever1an:
Have you tested this branch on your codebase?

sever1an:
@JustFly1984 Have you pulled and tested this branch with your projects?

JustFly1984:
@sever1an how can I test this branch in my project?

JustFly1984:
is there any movement?

pieh:
Published `gatsby@2.17.3`

JustFly1984:
Thank you, it solved my issues with 404

PR #18480, opened 2019-10-10
Removed outline on toggle

AkSachin:
Removed outline on toggling between dark and light mode.


marcysutton:
I don't think we should merge this -- focus outlines are there for keyboard users. If we're going to remove the default outline, it needs some kind of visual indicator on focus, and the links next to it use the default focus outline.

AkSachin:
I see. Makes sense.

LekoArts:
@AkSachin Would you be up for implementing these changes? If not we can close this PR :)

AkSachin:
You can close PR as of now. I'll request for new PR whenever I make proper changes that are keyboard-users friendly.

LekoArts:
Cool!

Issue #18143, opened 2019-10-04
Hook to control Socket.io configuration


JoeWrenn:
## Summary
I am proposing that a hook should exist for controlling the socket.io connection that is used when rendering a page from gatsby develop. This would allow multiple gatsby develop sites to work behind a proxy.

I am attempting to stitch together several gatsby develop instances behind a proxy environment. There is one final hurdle before it works in that the socket.io connection has a hardcoded connection to a relative URL path of "/socketio" by nature of the default constructor call here: https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/cache-dir/socketIo.js#L20

I would propose exposing the options passed to the socketio constructor in a similar manner to other gatsby hooks. This would allow for an instance of gatsby develop to be served from behind a proxy server while allowing the socket connection to go directly back to the instance instead of through the proxy. 

I think my general approach to this would mirror the SSR hooks where we provide functions to get and set the socketio configuration object.

SamIngram96:
Bumping for exposure

ANDIKKY:
This would allow multiple gatsby develop sites to work behind a proxy.

sidharthachatterjee:
@JoeWrenn Thank you for opening this!

This sounds like a pretty _niche_ use case and adding a hook for this doesn't seem like it would be worth the long term maintenance burden. 

It is currently possible to run multiple instances of `gatsby develop` on different ports using `gatsby develop -p XXXX` and that should suffice for the use case you describe, I think? 

Closing this for the moment but please feel free to reopen or comment in case I'm missing something. Thank you so much once again!

hughevans:
Just flagging that this is required in order to develop Gatsby sites within Github codespaces. Will have a play and see if I can pull together a PR.

PR #18509, opened 2019-10-11
feat(gatsby): adding socket hook

SamIngram96:
## Description

Adding an ssr hook to pass parameters into the socket when it's initialized

## Related Issues

https://github.com/gatsbyjs/gatsby/issues/18143

### DIFF
--- a/packages/gatsby/cache-dir/api-ssr-docs.js
+++ b/packages/gatsby/cache-dir/api-ssr-docs.js
@@ -196,3 +196,23 @@ exports.wrapPageElement = true
  * }
  */
 exports.wrapRootElement = true
+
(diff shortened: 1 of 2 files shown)


sidharthachatterjee:
@SamIngram96 Thank you taking the time to write this and for opening this PR! 

We won't be merging this in unfortunately because we think this is a pretty niche use case and isn't worth the long term maintenance burden. It is currently possible to run multiple instances of `gatsby develop` on different ports using `gatsby develop -p XXXX` and that should suffice for the use case the author of the issue describes. 

Nonetheless, thank you once again for taking the time to try this! We really appreciate it 💪 💜

PR #18517, opened 2019-10-11
fix(www): Fix `theme-ui` theme `breakpoints` (dark mode aftercare, episode 4)

fk:
The first breakpoint we get from `gatsby-design-tokens` breakpoints object is `xxs: 0`.
While this did make sense in the context of `styled-system`, it creates an unnecessary breakpoint in the context of `theme-ui`, which provides this "default" breakpoint for us already.

Switch a couple of `mediaQuery` usages to theme-ui's array syntax.

(Ref. #18027)

### DIFF
--- a/www/src/components/card.js
+++ b/www/src/components/card.js
@@ -29,10 +29,9 @@ const Card = ({ children }) => (
   >
     <div
       sx={{
-        p: 6,
-        pb: 0,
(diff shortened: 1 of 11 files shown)


fk:
https://www.youtube.com/watch?v=eUIzfBCyMWI <3 ;)

PR #18519, opened 2019-10-11
docs: Fix docs setup instructions broken link

jose4125:


lannonbr:
Thanks for opening this! this link is actually correct since that works on https://www.gatsbyjs.org

A good rule of thumb is that urls should work on gatsbyjs.org even though they might not work on GitHub since gatsbyjs.org is where docs are meant to be read.

I hope that makes sense. I'll be closing this PR for now but I urge you to go take a look at other open issues.

PR #18533, opened 2019-10-11
fix(gatsby-cli): properly update chalk color level

### LINKED ISSUE #18514: NO_COLOR does not work with Ink
Ink uses chalk internally which does a bunch of magic to figure out whether to render colours or not. 

In /packages/gatsby-cli/src/reporter/index.js we try to set `FORCE_COLOR` to 0 but that is too late because chalk has already initialised and set colours to true.

Mrigank11:
## Description
Update `chalk` color level on `reporterInstance.setColors`

## Related Issues
Fixes #18514

### DIFF
--- a/packages/gatsby-cli/src/reporter/index.js
+++ b/packages/gatsby-cli/src/reporter/index.js
@@ -119,6 +119,9 @@ const reporter: Reporter = {
     //  - ansi-colors: see https://github.com/doowb/ansi-colors/blob/8024126c7115a0efb25a9a0e87bc5e29fd66831f/index.js#L5-L7
     if (isNoColor) {
       process.env.FORCE_COLOR = `0`
+      // chalk determines color level at import time. Before we reach this point,
+      // chalk was already imported, so we need to retroactively adjust level
+      chalk.level = 0
     }
   },
   /**


packages/gatsby-cli/src/reporter/index.js
```diff
@@ -120,6 +120,13 @@ const reporter: Reporter = {
     if (isNoColor) {
       process.env.FORCE_COLOR = `0`
     }
+
+    // Change chalk color level
+    // 0	All colors disabled
+    // 1	Basic color support (16 colors)
+    // 2	256 color support
+    // 3	Truecolor support (16 million colors)
+    this.format.level = isNoColor ? 0 : 3
```
pieh: Could we move this into same if block as setting `FORCE_COLOR` env var, and only set it to 0 in that case? (to also not set Truecolor level in case we use colors).

And also add note like this:
```diff
    // disables colors in popular terminal output coloring packages
    //  - chalk: see https://www.npmjs.com/package/chalk#chalksupportscolor
    //  - ansi-colors: see https://github.com/doowb/ansi-colors/blob/8024126c7115a0efb25a9a0e87bc5e29fd66831f/index.js#L5-L7
    if (isNoColor) {
      process.env.FORCE_COLOR = `0`
+     // chalk determines color level at import time. Before we reach this point, 
+     // chalk was already imported, so we need to retroactively adjust level
+     chalk.level = 0
    }

```

sidharthachatterjee:
Yay, thanks so much for addressing this @Mrigank11 🥇

sidharthachatterjee:
Published in `gatsby-cli@2.8.4`

PR #18539, opened 2019-10-12
Remove title case from the heading to match the sidebar menu

jesCodingHere:
## Description

The sidebar menu doesn't have steps denoted by title case, but the page itself still has the title with the title case capitalization. This edit makes the two match up.


jesCodingHere:
Just found the style guide! Looks like the title case is appropriate for this article heading; instead, title casing needs to be added to the sidebar. I'm closing the pull request.

Issue #4038, opened 2018-02-14
[gatsby-transformer-remark] Improve opaque need for slug field


lourd:
## Description

The gatsby-transformer-remark has an [opaque dependency](https://github.com/gatsbyjs/gatsby/blob/ee5aa897d60e024a25108b69b4fca943768d3063/packages/gatsby-transformer-remark/src/extend-node-type.js#L219-L227) on having a `slug` field. It's needed if the `tableOfContents` property is requested as part of a GraphQL query. You have to use the `createNodeField` action on markdown nodes to add a field called `'slug'`, or else the query fails and it's not clear why.

### Environment

Gatsby version: 1.9.197

### Actual result

Given this query

```graphql
query AllPostsQuery {
  allMarkdownRemark {
    edges {
      node {
        tableOfContents
        frontmatter {
          title
        }
      }
    }
  }
}
```

It will have an error for the `tableOfContents` field

```
{
  "errors": [
    {
      "message": "Cannot read property 'slug' of undefined",
      "locations": [
        {
          "line": 13,
          "column": 6
        }
      ],
      "path": [
        "markdownRemark",
        "tableOfContents"
      ]
    }
  ],
  "data" : ...
}
```

### Expected behavior

The error should at least describe the need for a `slug` field and a link to where the person can read more about how to add it.

KyleAMathews:
:-( didn't realize this was happening. Yeah, this is very wrong. 

Perhaps I don't really understand what the tableOfContents is doing — why does it need to construct the full URL? Shouldn't it just do a relative link since it's all on the same page?

lourd:
That's what I thought. I found this PR (https://github.com/gatsbyjs/gatsby/pull/2856) about a special case that added the full URLs by default. I'm planning on adding an option to the plugin to specify whether or not you want to use full URL vs a relative hashes.

KyleAMathews:
@lourd yeah... ok — yeah slug shouldn't have been hard coded. Make it a full URL should be optional and if you select that, you should have to include a function which pulls the path off the MarkdownRemark node.

@Swizec how's that sound?

Swizec:
I like the idea of adding an option to enable full URLs. Didn't realize my PR would break things like that.

The original reason I added that was so you can print the tableOfContents on pages other than the page itself. Like on an index for your entire site. Useful for books/workshops/docs where it's convenient to have a full index on the frontpage.

[...]

PR #18543, opened 2019-10-12
feat(gatsby-transformer-remark): allow generating relative path in `tableOfContents`

Austaras:
## Description

Currently, gatsby generate absolute path in table of content which is not always necessary. After some digging I find the reason [here](https://github.com/gatsbyjs/gatsby/issues/4038) but somehow the option for generate relative path is missing.

However, there is various ways to achieve this:
1. add a relative flag. When true, generate relative path. This is the easiest solution but feels like somewhat wrong
2. add a absolute flag. When true, generate absolute path. IMO this is the right way but it will introduce a break change
3. generate relative path when `pathToSlugField` is empty. This will be a break change too.

@KyleAMathews What's your opinion?

## Related Issues
Related to #4038, #19193


Austaras:
the new commit is for route 3

Austaras:
may someone review this?

packages/gatsby-transformer-remark/src/extend-node-type.js
```diff
@@ -599,10 +601,8 @@ module.exports = (
       tableOfContents: {
         type: `String`,
         args: {
-          pathToSlugField: {
-            type: `String`,
-            defaultValue: `fields.slug`,
-          },
+          relative: `Boolean`,
```
pieh: I would go with second option here and use `absolute`, but set its default value to `true` for now (to avoid forcing breaking change) and keep default value for `pathToSlugField` as it is right now.

I agree with you that proposed behaviour is probably better default, but in my opinion this change is not big enough to warrant releasing new major version of this plugin, but we should track it and make the change for default values (make `absolute` default to false and `pathToSlugField` default to null/empty string) so we apply those changes when we really have something that would warrant major version bump.

Austaras: That make sense. Change as requested.

Austaras:
any chance get this one merged?

LekoArts:
Can you also add the new option to the GraphQL example?

Austaras:
added. Checks failure is due to a vulnerable dependency

pieh:
Published in `gatsby-transformer-remark@2.6.42`

Issue #18525, opened 2019-10-11
[gatsby-theme-blog-core] Would it be useful to have support for a featured image in the frontmatter?


ehowey:
## Summary

I have a use case for having a featured image supported in the front matter for a blog. My preference would be for this support to come directly from the core theme. I imagine this would also be a frequent use case for others as well.  I would be happy to take a go at building this functionality into `gatsby-theme-blog-core` but don't want to start on this project unless others also feel it would be useful?  Does this fit with the core theme vision or is it outside the mandate of the core blog theme?

I would imagine implementing something almost identical to what is discussed here:
https://www.gatsbyjs.org/docs/working-with-images-in-markdown/

I might need some help/careful review with the work in `gatsby-node.js` but this would be a good chance for me to contribute meaningfully to gatsby and also a good way for me to learn.

Issue #18271, opened 2019-10-07
How to use optional images in mdx frontmatter?


dcastil:
## Summary

I'd like to use optional images in frontmatter of markdown files processed with `gatsby-plugin-mdx`. I have multiple markdown files which correspond to article pages which use the same template and same page query. Some of those articles have a hero image specified in frontmatter and some don't.

How can I tell GraphQL that the `heroImage` frontmatter key can be an image but doesn't need to be?

Here is the current GraphQL behaviour when specifying a key in frontmatter if the image at `../assets/cms/hero-image.png` exists:

* `heroImage: ../assets/cms/hero-image.png` → works fine
* `heroImage: ''` → Field "heroImage" must not have a selection since type "String" has no subfields
* `# heroImage` → Unknown field 'heroImage' on type 'MdxFrontmatter'.

I'm a beginner to Gatsby and GraphQL but I think what I need is to make the `heroImage` field in GraphQL of type `File` instead of type `File!`.

## Relevant information

Here is a scenario on how it should work

```markdown
---
# my-markdown-file.md

heroImageOne: ../assets/cms/hero-image.png
heroImageTwo: ''
# heroImageThree
---
```

```jsx
// my-template.js

import React from 'react'
import { graphql } from 'gatsby'

export const pageQuery = graphql`
  query MyMdxQuery($fileAbsolutePath: String!) {
    mdx(fileAbsolutePath: { eq: $fileAbsolutePath }) {
      frontmatter {
        heroImageOne {
          childImageSharp {
            fluid {
              ...GatsbyImageSharpFluid
            }
          }
        }
        heroImageTwo {
          childImageSharp {
            fluid {
              ...GatsbyImageSharpFluid
            }
          }
        }
        heroImageThree {
          childImageSharp {
            fluid {
              ...GatsbyImageSharpFluid
            }
          }
        }
      }
    }
  }
`

export default function MyTemplateComponent({ data }) {
  const {
    heroImageOne, // ← { src, srcSet, … }
    heroImageTwo, // ← null
    heroImageThree // ← null
  } = data.mdx.frontmatter

  return <JSX />
}
```

> Note that I'm using the hero image as an example but I'm also generally interested in how to create optional frontmatter fields.
>
> The markdown files will be generated by netlifycms and the content editor should be able to use non-required fields. That would also enable adding frontmatter fields to the template without having to add the necessary key to every markdown file.

### Environment

```
  System:
    OS: macOS 10.14.6
[...]

PR #18545, opened 2019-10-12
Feat(gatsby-theme-blog-core)Add optional featured image support to frontmatter

ehowey:
## Description

This is a WIP pull request to look at adding in optional featured images by default to the core blog theme.  I have reached a point where it is working however the problem, and it is a big one, is that I can't figure out the best way to handle `null` query results. In other words to make having images optional.  As this PR is right now it works fine as long as _every_ blog post has an image, but as soon as one is missing an image it fails.  I am hoping that someone with more experience could point me in the right direction to make featured images an optional frontmatter type, similar to how the core theme is currently handling tags and keywords.

The frontmatter on posts could look like this:

```
---
title: Hello World (example)
date: 2019-04-15
featuredImage: ../assets/featured.jpg
---
```

I also recognize that this may not be something needed or wanted in the core blog theme - I will defer to folks with more knowledge and experience than me about this but I think this would be a common enough use case that optionally including it in the core theme like tags would be beneficial.

## Related Issues

Related to https://github.com/gatsbyjs/gatsby/issues/18525 and https://github.com/gatsbyjs/gatsby/issues/18271


ehowey:
I was able to add this support into my own work without modifying the core theme using: https://www.christopherbiscardi.com/post/accessing-frontmatter-through-interfaces/

Maybe what Chris suggests is actually a better solution because I don't think sharp can handle a null result? At least from what I know anyways.  If this is the case then feel free to close this PR.

pieh:
I think `gatsby-theme-blog-core` should have as little opinions as possible, and thing like that could be added via parent themes (or even regular plugins). It's pretty common to have support for `featuredImage`, but also not that uncommon not to have need for it, so personally I would vote not to add it here

ehowey:
I'll close this out because after doing some more thinking and digging I would tend to agree as well. Chris's solution is pretty easy to implement and provide a good workaround.  My vision was that the core theme would come with a "core" set of features that could be used or not used based on downstream choices by other themes/starters. I don't think there is an easy way to make images optional because of how sharp handles them, and if it isn't optional then it shouldn't be in here.

Issue #10203, opened 2018-11-29
Per-format quality settings for Gatsby Image (transformer/sharp)


coreyward:
## Summary

I'd like to see Gatsby Image support different quality settings for each image format available.

### Basic example

Here's what this could look like: 

```graphql
  query Example {
    resource {
      coverImage {
        fluid(maxWidth: 100, webp_quality: 90, jpg_quality: 50, png_quality: 60) {
          ...GatsbyImageSharpFluid_withWebp
        }
      }
    }
  }
```

### Motivation

Quality settings for JPG and PNG roughly align, but quality settings for WebP are _very_ different. Currently there's no way to specify that you'd like 90-quality WebP images and 50-quality JPG images in a Gatsby image query. 

A fallback could be established such that the default quality remains `50`, the `quality` setting overrides, and any undeclared format-specific quality settings fall back to whatever `quality` is set at. Setting `quality: 70, webp_quality: 90` would have the effect of setting the JPG and PNG quality to `70` while WebP quality is `90`.

coreyward:
I did a dive into this around the time I create the issue and found that doing this was a bit convoluted due to the way the code associated with this is structured and the different approach needed for `gatsby-image-contentful`. 

I'm still interested in making this happen, but don't have bandwidth right now.

baobabKoodaa:
I would love to see this happen!

Devin-Chaves:
Me too!

isaacschwab:
Has there been any progress on this issue? I'm having issues where my jpgs look fine with 50%, however, webp is getting too crunched. It would be really helpful to have this enhancement.

coreyward:
@isaacschwab Not as far as I'm aware, no. 

I'm using [Imgix](https://www.imgix.com/) on a number of projects, so I have the benefit of being able to generate multiple sizes and qualities by changing URL parameters. Instead of using the sharp pipeline stuff I've opted to build my own components that will generate a `srcset` for me using those URLs, which sidesteps this need.

isaacschwab:
@coreyward thanks for the response, and creating the initial issue. I have used similar tools to Imgix, they work well. For my current project I'm unable to go that route.  I think your proposed enhancement would be great for the flexibility and functionality of Gatsby, so hopefully this gets some more traction.

coreyward:
[...]

PR #18548, opened 2019-10-12
adding the ability to select different qualities for png, jpeg and we…

FailedSitcom:
…bp images

## Description

This gives the ability to set different qualities for different image formats:

```
query Example {
    resource {
      coverImage {
        fluid(maxWidth: 100, webp_quality: 90, jpeg_quality: 50, png_quality: 60) {
          ...GatsbyImageSharpFluid_withWebp
        }
      }
    }
  }
```

If none of the format specific qualities are passed, it will fallback to the `quality` argument if passed, otherwise the default quality will be used.

Before this is merged I would also like to add documentation explaining how it works, but as this is my first experience in contributing to Gatsby, I wanted to see if I've gone about this in the right way and if there are any concerns that I'm missing.

## Related Issues

Fixes https://github.com/gatsbyjs/gatsby/issues/10203


coreyward:
At a glance this looks pretty good, but I think it's important to also update the official Contentful-source plugin that integrates with Gatsby Image with these changes as well. 

https://github.com/gatsbyjs/gatsby/blob/2f4c4608df62e867f646881a00a5141860b7f5e0/packages/gatsby-source-contentful/src/extend-node-type.js#L62-L79

packages/gatsby-plugin-sharp/src/process-file.js
```diff
@@ -211,10 +217,12 @@ const compressPng = (pipeline, outputPath, options) =>
...
-            quality: `${options.quality}-${Math.min(
-              options.quality + 25,
-              100
-            )}`, // e.g. 40-65
+            quality: options.png_quality
+              ? `${options.png_quality}-${Math.min(
+                  options.png_quality + 25,
+                  100
+                )}`
+              : `${options.quality}-${Math.min(options.quality + 25, 100)}`, // e.g. 40-65
```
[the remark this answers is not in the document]
FailedSitcom: @coreyward nice call

FailedSitcom:
Thank you for the feedback @coreyward! I saw your mention of the Contentful integration on the original issue, and wanted to look into it before requesting a further review (but am very pleased you were able to point me to the relevant part of the code).

If I implemented it this week, would you be interested in taking a look?

coreyward:
I'd be happy to help review, but I'm not actually that familiar with that part of the codebase. I'm confident that one of the members of the core team will review and provide additional guidance as well—they do a great job with new contributors.

FailedSitcom:
Well I appreciate the time you've already taken to look over it, and will definitely check-in with the core team when I've addressed the points you raised.

Thanks again!

packages/gatsby-plugin-sharp/README.md
```diff
@@ -67,6 +70,9 @@ Automatically create sizes for different resolutions — we do 1x, 1.5x, and 2x.
 - `width` (int, default: 400)
 - `height` (int)
 - `quality` (int, default: 50)
+- `jpeg_quality` (int)
+- `png_quality` (int)
+- `webp_quality` (int)
```
vladar: Maybe replace underscores with camelCase as this looks inconsistent now (i.e. `png_quality` but `pngCompressionLevel`)?

vladar: We also need a note in docs about this feature. I guess the original issue description is a good start for it.

packages/gatsby-plugin-sharp/src/plugin-options.js
```diff
@@ -11,6 +11,9 @@ const pluginDefaults = {
 
 const generalArgs = {
   quality: 50,
+  jpeg_quality: ``,
+  png_quality: ``,
+  webp_quality: ``,
```
vladar: Having those args as empty strings seem a bit weird. Maybe we could change them to `null` (or `undefined`) and make sure they are handled properly in the `healOptions` function (with `options.quality` as a fallback as suggested by @coreyward below).

vladar:
@FailedSitcom thanks for the PR - it will be beneficial for the community once merged.

Please let me know if I can help you move this forward?

This could be a good case for a [pair programming](https://www.gatsbyjs.org/contributing/pair-programming/) session. Or, if you don't have time, you could provide us push-access to your branch so that we could do those tweaks and merge?

FailedSitcom:
@vladar thank you for your helpful suggestions and encouragement. I'd sadly been too busy over the last couple of weeks to implement the changes you suggested, but would be more than happy to this weekend.

If something comes up and I'm unable to do so, I'd be more than happy to provide push-access.

Thank you (and the awesome Gatsby team) again!

vladar:
Sounds good! Thanks for doing this!

PR #18556, opened 2019-10-12
fix(www): Add bottom spacing to center the close button

imagentleman:
## Description

The x in the close button (when you hover), on the modal that shows up when you click on a color in the guidelines page (https://www.gatsbyjs.org/guidelines/color/) is not vertically centered.

Screen Shot 2019-10-12 at 2 37 09 PM

The reason is that, that "x" is a character (lower case one), and that character includes some top whitespace.

When you inspect it, you can see that the letter element is actually centered:

Screen Shot 2019-10-12 at 2 36 59 PM

This pr adds some bottom padding to counter the letter top spacing (only to that specific button, so it doesn't break any current and future buttons made with that same component, via a prop that the styled component receives) and makes it look like this.

Screen Shot 2019-10-12 at 2 37 49 PM

### DIFF
--- a/www/src/components/guidelines/color/modal.js
+++ b/www/src/components/guidelines/color/modal.js
@@ -196,6 +196,7 @@ const ColorModal = ({ palette, color, handleModalClose }) => {
         <CloseButton
           onClick={handleModalClose}
           aria-label={`Close “${palette[color].name}” modal`}
+          paddingBottom="6"
         >
           &times;
         </CloseButton>


imagentleman:
> Hello again José, and thank you for another excellent PR description! 👋 🙏
> 
> Also "ugh" — this is probably a very good reason why this shouldn't have been a `&times;` in the first place … 🙄 — an icon and flexbox to center it probably would have been better, what do you think?

🤔 Yeah, an svg icon and flexbox would definitely be cleaner.

PR #18568, opened 2019-10-13
fix(gatsby-source-drupal): add error handling for empty body

timhagn:
## Description

While I was testing the preview functionality of the development server, it completely broke when an empty body was sent. With this PR it now will rather return a `500` Server Error.

## Related Issues

I reckoned this too small a change to open an issue for it ; ).

### DIFF
--- a/packages/gatsby-source-drupal/src/gatsby-node.js
+++ b/packages/gatsby-source-drupal/src/gatsby-node.js
@@ -195,7 +195,7 @@ exports.onCreateDevServer = (
           pluginOptions
         )
       } else {
-        res.status(500).send(`Received body was empty!`)
+        res.status(400).send(`Received body was empty!`)
       }
     }
   )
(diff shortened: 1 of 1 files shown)


packages/gatsby-source-drupal/src/gatsby-node.js
```diff
@@ -177,22 +177,26 @@ exports.onCreateDevServer = (
...
+            createNodeId,
+            createContentDigest,
+            getNode,
+            reporter,
+            store,
+          },
+          pluginOptions
+        )
+      } else {
+        res.status(500).send(`Received body was empty!`)
```
pieh: I think it should be code `400`:
> General error when fulfilling the request would cause an invalid state. Domain validation errors, missing data, etc. are some examples.

```suggestion
        res.status(400).send(`Received body was empty!`)
```

timhagn: Yup, thanks, sounds legit, committing : ).

timhagn:
You're more than welcome, thanks for reviewing and merging this such quickly : )!

PR #18572, opened 2019-10-13
chore(showcase): Add sagarhani.in to showcase

sagarhani:
## Description

Add sagarhani.in site built using Gatsby to showcase


sagarhani:
> Danger run resulted in 1 warning; to find out more, see the [checks page](https://github.com/gatsbyjs/gatsby/runs/258321165).
> 
> Generated by 🚫 [dangerJS](https://github.com/danger/danger-js/)

Seems like this warning is not because of my changes.

PR #18573, opened 2019-10-13
Update @typescript-eslint packages

mrseanbaines:
## Description

Updates `@typescript-eslint` packages, as the currently installed versions are pretty outdated and are missing some rules that are being referenced by the latest version of [`eslint-config-react-app`](@typescript-eslint/consistent-type-assertions) so was throwing an error when linting (the rule that was causing me issues is `@typescript-eslint/consistent-type-assertions`).

Tested locally and seems to have fixed the issue.


pieh:
@mrseanbaines Gatsby right now uses `eslint-config-react-app@^4.0.1`:
https://github.com/gatsbyjs/gatsby/blob/aafabc65b17a6b71f13fd06504700a7cc5d42f5e/packages/gatsby/package.json#L57

Which seems to set `@typescript-eslint/*` package versions to `1.x` ( https://unpkg.com/browse/eslint-config-react-app@4.0.1/package.json ). We might need to update that package as well (and any other peerDependencies that might need to be updated)?

mrseanbaines:
@pieh Think that's all the relevant `peerDependencides` updated

PR #18580, opened 2019-10-13
chore(starters): add gatsby-starter-leaflet

colbyfayock:
## Description
* Adds [gatsby-starter-leaflet](https://github.com/colbyfayock/gatsby-starter-leaflet) to showcase.
* Adds new category Mapping:Leaflet - the goal is to both provide the top level category of Mapping to encourage more map submissions with a subset of Leaflet, as there are other popular Mapping frameworks available.


docs/categories.yml
```diff
@@ -31,6 +31,7 @@ starter:
   - Language:Other
   - Language:TypeScript
   - Linting
+  - Mapping:Leaflet
```
wardpeet:
```suggestion
```

docs/starters.yml
```diff
@@ -4061,3 +4061,17 @@
     - Google Analytics
     - Simple design
     - Markdown support
+- url: https://gatsby-starter-leaflet.netlify.com/
+  repo: https://github.com/colbyfayock/gatsby-starter-leaflet
+  description: A Gatsby starter with Leafet!
+  tags:
+    - Landing Page
+    - Linting
+    - Mapping:Leaflet
```
wardpeet:
```suggestion
```

colbyfayock:
@wardpeet sure thing, updated it to just `Mapping`

wardpeet:
We don't want to add Mapping either 😛. Sorry about this but we don't want to add to many categories to our list. We also need to remove some that shouldn't be there.

colbyfayock:
fair enough, though I 🤞maybe it'll appear some day. it's an increasingly growing space with a lot of interesting technical work. we're building https://www.element84.com/dr/ on Gatsby for instance :)

removed from both files

PR #18587, opened 2019-10-14
feat(gatsby-theme-blog-core): Add theme option to customize excerpt length w/o graphQL

ehowey:
## Description

I was looking at the `gatsby-node.js` file and found myself wondering whether it would be possible to have a theme option in `gatsby-config.js`, similar to contentPath to control the length of the post excerpt.  I took a go at, and got it working.  The main benefit I see here is ease of use for the end user who does not have to go into graphQL to adjust the `pruneLength`.  I originally had the option called 'pruneLength' but went with `excerptLength` because I felt that was clearer to understand for the end user, this could be easily changed back.

I tested this to the extent that I was able and couldn't find any problems but given this is a more significant change I would want some good oversight from a more experienced developer in here. I don't want to break everyone's blog by accident.


freiksenet:
Sorry, we generally aren't going to merge PRs to themes that add more options. The preferred way to modify things in themes is either shadowing or, in this case, using schema customization in your site's gatsby-node.

Issue #18615, opened 2019-10-14
Set a sane schedule on Renovate 


sidharthachatterjee:
Renovate is fantastic but because of our config at the moment, we get several updates in a day. 

We should probably do two things

- batch all minor updates in a single PR
- schedule them at a reasonable time (twice a week, perhaps at 10 am GMT)

This should go in config at https://github.com/gatsbyjs/gatsby/blob/master/renovate.json 

Documentation for options is at https://docs.renovatebot.com/configuration-options/

FailedSitcom:
I'd like to take care of this 🙂

sidharthachatterjee:
@FailedSitcom It looks like @fboechats opened a PR doing the second point. Do you want to open another PR doing the first?

sidharthachatterjee:
Will likely need something like this 

https://github.com/gatsbyjs/gatsby/commit/c087a89aaac5a8281a7742d60fc883d0ae554ca7#diff-76e62915fba4048ac877a875e2e3b8d2

FailedSitcom:
@sidharthachatterjee I'd be glad to! Will submit a PR later today

FailedSitcom:
@sidharthachatterjee would you like a rule like the one before to exclude the same/equivalent patterns ("^react", "^@babel/", "^eslint")? Or are absolutely all minor versions to be grouped in this rule?

muescha:
Can we name also the PR something like the following, I always trapped into this PR because the `by bot` info is very small on the second line:

`[Renovatebot] chore: update ...`

sidharthachatterjee:
@FailedSitcom That was a setting I ended up setting while I was playing with Renovate and while it made sense in the beginning (for visibility into a vast set of updates when we first enabled it for packages), it seems less necessary now. 

@muescha I'm not sure if adding a label or changing the title to include that it is made by a bot adds any value since one can filter by user to see these PRs (or commits)

muescha:
yes i can add a search term ` -author:app/renovate` to the search box everytime but github not remember this setting.

filter by eye is not that easy because the second line with username is very small to see it :( and more smaller on my mobile phone

there are this 2 kinds of PR:

`chore: update ... to ...`
`fix: update dependency ...` 

this lines often get used by others users also for some PRs 


if we add an marker, then it's more clear its from a bot: 

`[bot] chore: update ... to ...`
`[bot] fix: update dependency ...` 

or just:
`[bot] update ... to ...`
[...]

PR #18620, opened 2019-10-14
Renovate Schedule

fboechats:
## Description

Added a timezone and schedule config to Renovate

## Related Issues

refs #18615


fboechats:
In the docs it says they do this by default. "By default, Renovate won't distinguish between "patch" (e.g. 1.0.x) and "minor" (e.g. 1.x.0) releases - it groups them together."

PR #18621, opened 2019-10-14
chore(www): Referrer policy

### LINKED ISSUE #18583: Showcase: Referrer not available on external link visits
## Description
I've noticed within GA traffic that the showcase I had submitted isn't showing up even 1 time in the traffic source. Now this could definitely simply be no clicks, but out of curiosity, I tried clicking myself while reviewing a GA debugger, looking at `document.referrer`, and GA itself and sure enough, the referrer was blank and the traffic source was defaulted to direct/none.

I tried looking around about it and unless it's intentionally being stripped, it seems like other issues could be security, https => http, but the links are both https.

### Steps to reproduce
* Visit showcase
* Click on Visit Website for one of the sites
* Open your web console and look at `document.referrer`
* It's blank

If you're running a debugger like Observepoint or the GA debugger, typically you can additionally see the traffic source on the new page, which is not available coming from Gatsby.

### Expected result

Clicking an external link from Gatsby should retain the `document.referrer` and additionally be available with Google Analytics or other traffic analytics tools.

### Actual result

```document.referrer=""```

### Environment

https://www.gatsbyjs.org/showcase/

### PR DESCRIPTION
**Pull request by colbyfayock**
## Description

The docs site currently includes `same-origin` as a referrer policy which is default per `gatsby-plugin-netlify`. The issue here, is it strips the referrer for outbound traffic, leaving sites to not be able to tell where the traffic came from, particularly in in Google Analytics, shows this traffic as `direct/none`.

This updates the `gatsby-config` to manually set the headers, allowing us to set the Referrer Policy to `strict-origin-when-cross-origin`, which will only show the origin for cross-origin traffic, but still allow internal traffic to capture the entire referring URI.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy

Confirmed this fixes the header in the `_headers` output:
```
/*
  X-Frame-Options: DENY
  X-XSS-Protection: 1; mode=block
  X-Content-Type-Options: nosniff
  Referrer-Policy: strict-origin-when-cross-origin
```

The only thing that's kind of a bummer here is the fact that we have to manually include all headers. Per the plugin docs, it seems like that's required, where intuitively, I would think that the `merge` option set to `false` would overwrite instead. That said, it could be nice to include an option within that plugin that would allow you to provide that functionality, where you can include a single header without stripping all of the defaults.

## Related Issues

Fixes #18583

### CODE REVIEW

File: www/gatsby-config.js [hunk 1]
```diff
@@ -306,7 +306,20 @@ module.exports = {
         ],
       },
     },
-    `gatsby-plugin-netlify`,
+    {
+      resolve: `gatsby-plugin-netlify`,
+      options: {
+        mergeSecurityHeaders: false,
```
wardpeet: Setting this to true (or omitting this because default is true) should merge headers so not everything needs to be duplicated. If this is not the case that's a bug in gatsby-plugin-netlify :joy:.
```suggestion
        mergeSecurityHeaders: false,
```

colbyfayock: thats strange. im having a hard time getting build to run locally due to some mdx errors elsewhere, but i tested this in the browser and it seems to be running as expected, not sure what i was seeing before 🤔made the change
image

local issues im running into. im not seeing this in the codebase temp remove it. post `yarn install` fwiw if that matters
image

colbyfayock:
caught the branch up with master

wardpeet:
Works like a charm. Thanks for fixing.

PR #18622, opened 2019-10-14
Updated babel.md to adhere to gatsby style guide

adibfirman:
# Description
Mainly change `we` to `you`

# Related Issue
#18284


LekoArts:
Thanks for the PR! However, we is referring to "Gatsby" here and hence the current form is correct.

PR #18624, opened 2019-10-14
Update from we to you

YzahFe:
for gatsby style guide


LekoArts:
Hi, thanks for the PR! But we refers to "Gatsby" here and changing these to "you" would be misleading. Hence I'm closing this PR.

PR #18634, opened 2019-10-14
chore(docs): Update we to you

dosden:


LekoArts:
@dosden Please bundle these "we" => "you" PRs in the future into one PR. It gets really noisy if you create a bunch of PRs changing one word instead of one where you change multiple words. I'd highly appreciate that. Thanks!

dosden:
This is well noted.
Much appreciated.

PR #18641, opened 2019-10-14
Update gatsby-config.js

aldrinpvincent:
Changing broken link on 'gatsby-plugin-offline' documentation


sidharthachatterjee:
This shouldn't be necessary. http://gatsby.dev seems to be down which we're fixing at the moment. 

Nonetheless, thank you for taking the time to open this!

PR #18644, opened 2019-10-14
grouping together minor updates in renovate.json rule

FailedSitcom:
## Description

Adding Renovate grouping rule to batch all minor updates in a single PR

## Related Issues

Fixes https://github.com/gatsbyjs/gatsby/issues/18615

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -20,6 +20,10 @@
     {
       "groupName": "minor updates in packages",
       "updateTypes": ["minor"]
+    },
+    {
+      "groupName": "patch updates in packages",
+      "updateTypes": ["patch"]
     }
   ],
   "timezone": "GMT",
--- a/renovate.json
+++ b/renovate.json
@@ -16,6 +16,12 @@
   "bumpVersion": null,
   "semanticCommitScope": null,
   "prHourlyLimit": 0,
+  "packageRules": [
+    {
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -16,6 +16,12 @@
   "bumpVersion": null,
   "semanticCommitScope": null,
   "prHourlyLimit": 0,
+  "packageRules": [
```
sidharthachatterjee: Can we add one more for patches perhaps?

sidharthachatterjee: Or maybe we only need to group patches and keep minors separate? What do you think

FailedSitcom: Thanks for the review @sidharthachatterjee! I'll take a look at both options and see which might be more appropriate.

FailedSitcom: @sidharthachatterjee I reviewed recent package updates, and whilst the vast majority of updates are patches (so definitely more important to batch), grouping together minor updates also feels like it would clear up a fair amount of noise with any major risks.

Let me know if you have any other thoughts, and thank you so much for such a helpful, welcoming introduction to contributing to Gatsby. 😄

sidharthachatterjee: I completely agree with you and this looks like a good approach! 

Thank you so much for taking the time to research this and to make a thought out decision and of course for contributing back this change!

Issue #18579, opened 2019-10-13
Automatic page dependency registration not working as described in the docs (ie hot reloading not always working when it should)


nickyfahey:
## Description

When you create a resolver that uses `context.nodeModel` a page dependency is not automatically added unless a path is specified, or you manually call `trackPageDependencies`.

The [schema customisation docs](https://www.gatsbyjs.org/docs/schema-customization/#taking-care-of-hot-reloading) say:

> When you retrieve nodes from the store with `context.nodeModel` methods, it is usually not necessary to do anything manually, because Gatsby will register dependencies for the query results automatically. The exception is `getAllNodes` which will not register data dependencies by default.

It also implies that specifying a connection type in `gatAllNodes` should be enough to add a page dependency. Except hot reloading does not work unless you either specify a path or call `context.nodeModel.trackPageDependencies` in the resolver.

```js
    // This does not add a page dependency:
    const things = context.nodeModel.getAllNodes(
      { type: "BlogPost" },
      { connectionType: "BlogPost" }
    )
    
    // Neither does this:
    const things = await context.nodeModel.runQuery({
    	type: "BlogPost",
    	query: { filter: { title: { eq: "A Blog Post"} } },
    })
    
    // This does work, but specifying the path manually should not be necessary:
    const things = context.nodeModel.getAllNodes(
    	{ type: "BlogPost" },
    	{ connectionType: "BlogPost", path: context.path }
    )
    
    // This will always work:
    const results = await context.nodeModel.runQuery(...)
    context.nodeModel.trackPageDependencies(results, { connectionType: "BlogPost" })
```

### Steps to reproduce

1. Download [this demo project](https://github.com/nickyfahey/gatsby-page-dependency-bug) and run `gatsby develop`
2. With the development site open in a web browser make a change to src/data/thing.yaml

### Expected result

Hot reloading should work. Changes made to the yaml file should automatically be reflected on the page displayed in the web browser.

### Actual result

Nothing happens

### Environment

  System:
    OS: Windows 10
    CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
  Binaries:
    Yarn: 1.19.1 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.11.3 - C:\Program Files\nodejs\npm.CMD
  Languages:
    Python: 2.7.16
  Browsers:
    Edge: 44.18362.387.0
  npmPackages:
[...]

PR #18650, opened 2019-10-14
fix(gatsby): create page dependencies from contextual node model methods even if no path is passed

nickyfahey:
## Description

Makes sure ContextualNodeModel methods that can create page dependencies get the path from the context if it is not passed in the pageDependencies arg.

## Related Issues

#18579


pieh:
Failing test is not related to your changes and was fixed in master sometime yesterday. When you have a chance, please merge master in again to resolve that

PR #18685, opened 2019-10-15
chore(showcase): New submissions for site showcase

flowen:
## Description

2 submissions for the site showcase


docs/sites.yml
```diff
@@ -7910,3 +7911,40 @@
     - Web Development
   built_by: IBM
   featured: true
+- title: Mevish Aslam, business coach
+  url: "https://mevishaslam.com/"
+  main_url: "https://mevishaslam.com/"
+  source_url: ""
```
pieh: Please remove this field entirely if you don't want to provide url to source code / project is private

docs/sites.yml
```diff
@@ -7910,3 +7911,40 @@
...
+    - Freelance
+    - Marketing
+    - Portfolio
+  built_by: Rou Hun Fan
+  built_by_url: "https://flowen.me"
+  featured: false
+- title: Principles of wealth
+  url: "https://principlesofwealth.net"
+  main_url: "https://principlesofwealth.net"
+  source_url: ""
```
pieh: Same as above

docs/sites.yml
flowen: I see, thanks for checking!

docs/sites.yml
flowen: should be good now

Issue #10692, opened 2018-12-27
Relationships between 3rd party schemas in gatsby-source-graphql


michaelhayman:
I have two GraphQL schemas.  I have them integrated as so in my `gatsby-config.js` file:

````
  plugins: [
    {
      resolve: 'gatsby-source-graphql',
      options: {
        typeName: 'GCMS',
        fieldName: 'gcms',
        url:
          'https://api-apeast.graphcms.com/v1/cjm2g4capq01bn21shzkcx/master',
      },
    },
    {
      resolve: 'gatsby-source-graphql',
      options: {
        typeName: 'PSQL',
        fieldName: 'psql',
        url:
          'https://my-server.herokuapp.com/v1alpha1/graphql',
      },
    },
  ]
````

Say `rooms` are in the first schema, and `availability` is in the second schema, how would I get all of a room's availability?  According to this configuration, the schemas are only merged at a superficial level - i.e. there's no relationships between the two schemas.  I think I'd first have to query for rooms, then query for availability, which is a) inconvenient for a developer and b) would result in an N+1 problem.

Looking at [the schema stitching guide](https://www.apollographql.com/docs/graphql-tools/schema-stitching.html), I'd need to use resolvers, but I don't see an option for that in `gatsby-source-graphql`.

Is there an API I'm missing?  The `gatsby-source-graphql` docs seem to lack this kind of documentation.  I saw some mention of relationships between nodes elsewhere but I'm not sure if this applies to this scenario.

The other option is setting up another graphql server to stitch the two schemas together, and then just simply pass one schema to gatsby, is that what's recommended if one wants to relate 2 schemas together?

Cheers,

pieh:
Hey @michaelhayman!

We currently don't support passing `resolvers` option to `graphql-tools`'s `mergeSchemas` that happens in gatsby core:
https://github.com/gatsbyjs/gatsby/blob/672e2b3087f22f27a0f2763da328c8592e7d69c6/packages/gatsby/src/schema/index.js#L37-L39

This needs a lot of thought to do it properly, because we want to add "cross-schema" resolvers between schema generated by gatsby core and 3rd party and we don't want to create API that will limit us.

But you still should be able to do that - just need some manual intermediatte schema stitching first - in `gatsby-node.js` file you could do:

```js
const { mergeSchemas } = require(`graphql-tools`)

exports.sourceNodes = ({ actions }) => {
[...]

PR #18691, opened 2019-10-15
Ability to transform external graphql schemas client-side

ariadne-github:
## Description

This adds a callback to transform client-side an external graphql schema, to add for example custom resolvers when the schema doesn't natively support this.

Example:
```
const customTransformSchema = (schema) => {

  const linkTypeDefs = `
    extend type StructuredContentLink {
      content: StructuredContent
    }
  `;
  
  const customRemoteSchema = mergeSchemas({
    schemas: [schema, linkTypeDefs ],
    resolvers: {
      StructuredContentLink: {
        content: {
          resolve(source, _args, context, info) {
            return info.mergeInfo.delegateToSchema({
              schema: schema,
              operation: 'query',
              fieldName: 'structuredContent',
              args: {
                structuredContentId: source.id,
              },
              context,
              info,
            });
          },
        },
      },
    },
  });

  return customRemoteSchema;
}
```

In this example schema, the type StructuredContentLink contains a reference to a StructuredContent by id, but the schema does not allow to query inside the relation directly. The transform function extends the schema and adds a resolver that queries the linked node.

With this option it will be possible to create new relations in native graphql schemas, and it's entirely optional.

It's not possible to achieve the same result by using the existing createSchema option, since the modified schema would be linked to the remote source, and that would result in validation errors server-side.

## Related Issues

This is very similar, but about a relation between different third-party schemas: #10692.

### DIFF
--- a/packages/gatsby-source-graphql/src/gatsby-node.js
+++ b/packages/gatsby-source-graphql/src/gatsby-node.js
@@ -96,7 +96,7 @@ exports.sourceNodes = async (
     })
     return {}
   }
-  
+
   let resultSchema = remoteSchema
   if (customSchemaTransformer) {
     resultSchema = customSchemaTransformer(remoteSchema)
--- a/packages/gatsby-source-graphql/src/gatsby-node.js
+++ b/packages/gatsby-source-graphql/src/gatsby-node.js
@@ -26,10 +26,10 @@ exports.sourceNodes = async (
     fieldName,
     headers = {},
     fetchOptions = {},
-    customSchemaTransfomer,
     createLink,
     createSchema,
     refetchInterval,
+    customSchemaTransformer,
   } = options
 
   invariant(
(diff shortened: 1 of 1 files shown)


freiksenet:
You can use [`createResolvers`](https://www.gatsbyjs.org/docs/node-apis/#createResolvers) API to do that. Look at [this guide to schema customization](https://www.gatsbyjs.org/docs/schema-customization/#extending-third-party-types) for an example.

I'll close this issue, if you have any more questions, feel free to reopen or create a new one. Thank you for using Gatsby! :purple_heart:

ariadne-github:
@freiksenet I've seen that API, but I've found that inside the resolver I'm only able to query on the internal Gatsby schema, and not on the thid-party ones, so it's impossible to create a relation between third-party schema's fields. 
If you can explain me how to do that with existing APIs I'll be very happy.
I can't find a button to re-open the PR, do I need to open an issue instead?

freiksenet:
You can actually delegate to the Gatsy schema when you are doing create resolver. You can also apply transforms manually if you need it in this case. I've created [a codesandox with an example](https://codesandbox.io/s/gatsby-starter-default-0st39?fontsize=14).

ariadne-github:
Thanks @freiksenet, I've solved it. My use case was a little more complicated since I needed to delegate to a non-root field, but I managed to do it by using a transform with WrapQuery, like in this example: https://gist.github.com/elliottsj/fa7042c0588ea6df1bc2f02f583d1100.

PR #18694, opened 2019-10-15
Topics/free headless cms

hashimwarren:
This blog post answers a popular question - Which  free headless CMS's are best to use?

free-headless-cms-1

free-headless-cms-2

free-headless-cms-3

free-headless-cms-4

free-headless-cms-5


packages/gatsby-plugin-sharp/src/__tests__/__snapshots__/index.js.snap
```diff
@@ -23,6 +23,19 @@ Object {
 }
 `;
 
+exports[`gatsby-plugin-sharp duotone fixed 2`] = `
```
gillkyle: Not sure how this got added but could it be removed? It doesn't seem related to me 🤔

Issue #18571, opened 2019-10-13
Extra comma and " | Gatsby" in showcase details title


rosslh:
## Description

Showcase details pages have an extra `,` and ` | Gatsby` in the document `<title/>`

### Steps to reproduce

- Visit https://www.gatsbyjs.org/showcase/www.metaprojection.ca
- Visit https://www.gatsbyjs.org/showcase/reactjs.org

### Expected result

- The title should be `MetaProjection: Showcase | GatsbyJS`
- The title should be `ReactJS: Showcase | GatsbyJS`

### Actual result

- The title is `MetaProjection,: Showcase | GatsbyJS | GatsbyJS`
- The title is `ReactJS,: Showcase | GatsbyJS | GatsbyJS`

### Environment

Reproduced on:
- OS
  - Kubuntu 16.04
- Browsers
  - Firefox 69.0.02
  - Chrome 77.0.3865.90

fboechats:
@rosslh How can I reproduce it locally??

mrkutly:
It looks like the happens when you link directly to a showcase page. If you navigate to any of them from the showcase index, the title is fine. The extra comma only gets added if you go directly to a show page.

mrkutly:
Digging around I found at least 2 react Helmet components that were tryin to set the title for each page. Setting the prop `titleTemplate=""` for the Helmet in the showcase-details.js component did the trick.

PR #18705, opened 2019-10-16
Fix issue 18571  - showcase details page titles

mrkutly:
## Description

Adds  `titleTemplate=""` prop to the `Helmet` component in components/showcase-details.js. Also adds `defer={false}` prop to the `Helmet` in views/showcase/index.js. There are multiple `Helmet` components being rendered at the same time, which was causing strange behavior for when the details page was a modal versus the full page.

For the React details page, you would sometimes get the title:
"React.js,: Showcase | GatsbyJS | GatsbyJS"

This PR fixes it so that it is now whether rendered as a modal or a full page:
"React.js: Showcase | GatsbyJS"

 The `titleTemplate="%s | GatsbyJS"` prop from the top-level `Helmet` in site-metadata.js was interfering with the title rendering correctly in the showcase-details. Once that was fixed by overwriting it in the showcase-details.js `Helmet`, the `defer={false}` prop was needed to prevent the `Helmet` in views/showcase/index.js from flickering with the bad version of the title while transitioning between modal open and closed state.

## Related Issues

Fixes [18571](https://github.com/gatsbyjs/gatsby/issues/18571)

### DIFF
--- a/www/src/components/showcase-details.js
+++ b/www/src/components/showcase-details.js
@@ -279,7 +279,7 @@ const ShowcaseDetails = ({ parent, data, isModal, categories }) => (
             }}
           >
             <div css={{ width: `100%` }}>
-              <Helmet>
+              <Helmet titleTemplate="">
                 <title>{data.sitesYaml.title}: Showcase | GatsbyJS</title>
                 <meta
                   property="og:image"
--- a/www/src/views/showcase/index.js
+++ b/www/src/views/showcase/index.js
@@ -50,7 +50,7 @@ class ShowcaseView extends Component {
 
     return (
       <Layout location={location}>
-        <Helmet>
+        <Helmet defer={false}>
           <title>Showcase</title>
           <meta
             name="description"


muescha:
Is your PR is more a workaround?

mrkutly:
Yes I suppose it is more a workaround than a good solution. I will try to find a better way to solve this that does not involve overriding the `titleTemplate` prop.

mrkutly:
@muescha I found what I believe is a much better solution and will open a PR for it. It is still a bit of a workaround, but it fixes the issue much more cleanly.

There is [this](https://github.com/nfl/react-helmet/issues/408) open issue for React Helmet about the extra commas being added into the titles. The current workaround for this is to wrap the title text in a string inside `{}`. This removes the comma.

I'm setting the `titleTemplate` prop on the `Helmet` component in `showcase-details.js` to be `"%s | GatsbyJS"`. This is because sometimes this is the only `Helmet` component on the page, so it needs the template. Before, this was being worked around by hardcoding `" | GatsbyJS"` into the title, which was causing the double "| GatsbyJS". By using titleTemplate, this will only render once.

I'll open another PR shortly.

muescha:
Should we ask the @gatsbyjs/core team what's the best here to avoid double helmet

mrkutly:
@muescha I don't think having multiple helmets is a problem as it is supported by the react-helmet library. We will need multiple helmets to update the title as the user navigates around the page.

The issue was that the `"| GatsbyJS" was hardcoded in one `title` tag but also set as the `titleTemplate` prop for a parent `Helmet`. 

According to the react-helmet docs:

Nested or latter components will override duplicate changes:
```javascript
<Parent>
    <Helmet>
        <title>My Title</title>
        <meta name="description" content="Helmet application" />
    </Helmet>

    <Child>
        <Helmet>
            <title>Nested Title</title>
            <meta name="description" content="Nested component" />
        </Helmet>
    </Child>
</Parent>
outputs:

<head>
    <title>Nested Title</title>
    <meta name="description" content="Nested component">
</head>
```

So we just need to make sure that we use `titleTemplate` for the top-level `Helmet` components (whichever those are at any time) and that we do not hardcode the template into the title itself.

PR #18719, opened 2019-10-16
Highlight empty lines without breaking line numbering or hiding them

ethernal:
Highlight empty lines without breaking line numbering or hiding them. See #17849

## Description

Allows to highlight empty lines without breaking line numbering and prevents highlighted line from being hidden. Currently that is broken if empty line is highlighted. See #17849 for screenshots and description.

## Related Issues

Fixes: #17849

### DIFF
--- a/packages/gatsby-remark-prismjs/src/directives.js
+++ b/packages/gatsby-remark-prismjs/src/directives.js
@@ -154,7 +154,12 @@ module.exports = function highlightLineRange(code, highlights = []) {
       })
       .map(line => {
         if (line.highlight) {
-          line.code = highlightWrap(line.code)
+          
+          // use non-breakable space to retain empty line in source listing 
+          // as it is important when both line highlight and line numbering are used.
+          line.code
+            ? (line.code = highlightWrap(line.code))
+            : (line.code = highlightWrap("&nbsp;"))
         }
         return line
       })
--- a/packages/gatsby-remark-prismjs/src/directives.js
+++ b/packages/gatsby-remark-prismjs/src/directives.js
@@ -154,9 +154,6 @@ module.exports = function highlightLineRange(code, highlights = []) {
       })
       .map(line => {
         if (line.highlight) {
-          
-          // use non-breakable space to retain empty line in source listing 
-          // as it is important when both line highlight and line numbering are used.
           line.code
             ? (line.code = highlightWrap(line.code))
             : (line.code = highlightWrap("&nbsp;"))


packages/gatsby-remark-prismjs/src/directives.js
```diff
@@ -154,7 +154,9 @@ module.exports = function highlightLineRange(code, highlights = []) {
       })
       .map(line => {
         if (line.highlight) {
-          line.code = highlightWrap(line.code)
+          line.code
```
muescha: better check for empty string or null? Otherwise a line with number `0` and text `false` would be also exchanged?

ethernal: Prism highlights from 1 and up. You cannot highlight line 0 as there is no such line.

ethernal: Snapshot tests are failing as I am injecting non breakable space into empty lines (&nbsp;) so these tests should just get the snapshots updated.

wardpeet:
Can you move the if check inside https://github.com/gatsbyjs/gatsby/pull/18719/files#diff-5fd2ab3304a334d4977bdcb27e492e3eR54

LekoArts:
Hi @ethernal 👋 Do you have time to finish this PR?

ethernal:
@LekoArts I hope to finish it this weekend. I would be grateful for guidance as my experience with GitHub is limited. I know I need to fix the tests as they fail due to introduction of `&nbsp` character.

Is there a way I can test all of it locally?

LekoArts:
@ethernal Hi, do you have time to wrap up this PR? You can read up on setting up the repository here: https://www.gatsbyjs.org/contributing/setting-up-your-local-dev-environment/

pvdz:
Looks like this was fixed in #21711 and as such and seeing how there's no updates in this PR in a while I'm going to close this PR. If this is incorrect please re-open and update it. Thanks :)

andrew-chang-dewitt:
> 
> 
> Looks like this was fixed in #21711 and as such and seeing how there's no updates in this PR in a while I'm going to close this PR. If this is incorrect please re-open and update it. Thanks :)

I'm not sure this was fixed in #21711. I'm still experiencing the issue in #17849 & manually editing `gatsby-remark-prismjs/src/directives.js` in my `node_modules` as this pull request would fixed it for me. 

I think my problem is because #21711 helps style the empty lines better, but it seem to actually insert the empty line like this PR's edit does, but I'm not sure

PR #16168, opened 2019-07-28
feat(gatsby): Bumps eslint-config-react-app 4.x

ThewBear:
## Description

This PR introduces eslint-config-react-app `4.0.1` which comes with

- Typescript linting supports
- `eslint-config-react-hooks`

## Related Issues & PRs

#15114 
#15976

PR #15976, opened 2019-07-22
feat(gatsby): Includes ts/tsx files to eslint rules

ashr81:
## Description

Includes ts/tsx files extensions to check for the eslint rules.

Fixes #15114

PR #18721, opened 2019-10-16
feat(gatsby-plugin-typescript): Supports linting

ThewBear:
## Description

Splitted from https://github.com/gatsbyjs/gatsby/pull/16168#issuecomment-515952470

`eslint-config-react-app` has supported linting Typescript since version [3.0.0](https://github.com/facebook/create-react-app/releases/tag/v3.0.0) (facebook/create-react-app#6513).

It would be great if Gatsby also supports this.

## Related Issues

#15976
#16168


pieh:
In core we do check if user has their own eslint config before adding eslint rule (which is opinionated, so this is a way for users to disable builtin one if they have their own config) - https://github.com/gatsbyjs/gatsby/blob/3c8623c44fba78b6ecfdbd2aa555440d4284be28/packages/gatsby/src/utils/webpack.config.js#L305-L307

And addition to that we only apply this loader in `gatsby develop` when building browser bundle.

In current implementation in this PR this part is missing, and it's not obvious how we could re-implement it (as `hasLocalEslint` function is internal one and not public one), but maybe we could examine config that gatsby build so far, and see if there is rule for eslint-loader to determine if typescript plugin should add config as well? This should cover both reservations?

Some quick code that could be clean up to check it:
```js
const hasBuiltInEslintRule = webpackConfig => {
  try {
    return webpackConfig.module.rules.some(rule => {
      if (rule.enforce === `pre`) {
        return rule.use.some(use => {
          return /eslint-loader/.test(use.loader)
        })
      }
      return false
    })
  } catch {
    return false
  }
}

exports.onCreateWebpackConfig = ({ stage, getConfig }) => {
  console.log(`stage ${stage}: ${hasBuiltInEslintRule(getConfig())}`)
}
```

will output:
 - in `gatsby develop` if there *is no* eslint config in root of directory
   ```
   stage develop-html: false
   stage develop: true
   ```
 - in `gatsby develop` if there *is* eslint config in root of directory
   ```
   stage develop-html: false
   stage develop: false
   ```
 - in `gatsby build` (regardless of existance of eslint config file):
   ```
   stage build-javascript: false
   stage build: true
   ```

So this is something you could use to determine wether to add eslint rule for .tsx? files or not

pieh:
All right, I was testing it and few notes:

We do need to add `typescript` as peerDependency to `gatsby-plugin-typescript`, update plugin readme so users can copy install command to install it too and in code do some conditional check if typescript is installed before adding the rule (potentially warn users that `typescript` is not installed and builtin eslint won't be working on typescript files).

Pretty confusingly, you can use `gatsby-plugin-typescript` without `typescript` being installed right now (as the way it's implemented it's using babel plugin and not actual typescript to compile the code).

LekoArts:
@ThewBear Hey 👋 Do you have time to address the requested changes?

LekoArts:
@ThewBear Hey 👋 Kind reminder of the above ping 😊

ThewBear:
@LekoArts Will address this weekend.

ThewBear:
@pieh I make the linting optional by checking if the user has `typescript` installed.

packages/gatsby-plugin-typescript/src/gatsby-node.js
```diff
@@ -33,6 +39,38 @@ function onCreateWebpackConfig({ actions, loaders }) {
       ],
     },
   })
+
+  if (stage === `develop`) {
+    let ts
```
wardpeet: Please use a more descriptive name
```suggestion
    let isTypescriptDepAvailable
```

packages/gatsby-plugin-typescript/src/gatsby-node.js
```diff
@@ -33,6 +39,38 @@ function onCreateWebpackConfig({ actions, loaders }) {
...
+    try {
+      ts = require.resolve(`typescript`)
+    } catch (e) {
+      reporter.warn(
+        `"typescript" is not installed. Builtin ESLint won't be working on typescript files.`
+      )
+    }
+
+    if (ts) {
+      const builtInEslintRule = getConfig().module.rules.find(rule => {
```
wardpeet:
```suggestion
      const hasBuiltInEslintRule = getConfig().module.rules.find(rule => {
```

ThewBear: Since `builtInEslintRule` is used to create `typescriptEslintRule` not only for the conditional check. I prefer to stick with `builtInEslintRule`.

wardpeet: Oh you're right :) builtInEslintRule makes more sense, I missed that part  👍

wardpeet:
Fixed in `gatsby-plugin-typescript@2.2.0`

pvdz:
Looks like in some cases (at least one) user configs are not considered after all --> #21913

PR #18726, opened 2019-10-16
fix(gatsby-plugin-netlify): cast `status` to string

visualfanatic:
## Description

If you pass the `statusCode` like in the example from the docs:
```js
createRedirect({
  fromPath: "/url_that_is/not_pretty",
  toPath: "/pretty/url",
  statusCode: 200,
})
```

... and try to add the `force` param then you are going to get the following error:
> status.concat is not a function

This is because `statusCode` is a `Numer`, and this PR makes a tiny change which casts the `status` to a `String` when appending the exclamation point.


TylerBarnes:
This is published in `gatsby-plugin-netlify@2.1.22`

Issue #17469, opened 2019-09-07
Image loading (via graphql) from front matter fails on Hot Reload when editing MDX while theme development via yarn workspaces


aTechGuide:
## Description
Hot reload of the post written in MDX (i.e. when a post is edited) fails when we pull featured image data inside graphql query with following error

```
{
  errors: [
    TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type undefined
        at validateString (internal/validators.js:112:11)
        at Object.resolve (path.js:981:7)
        at findLinkedFileNode (/Users/bornshrewd/code/blog/theme-test/node_modules/gatsby/dist/schema/resolvers.js:207:47)
        at resolveValue (/Users/bornshrewd/code/blog/theme-test/node_modules/gatsby/dist/schema/resolvers.js:222:108)
        at /Users/bornshrewd/code/blog/theme-test/node_modules/gatsby/dist/schema/resolvers.js:219:10
        at processTicksAndRejections (internal/process/task_queues.js:85:5)
        at async Promise.all (index 2)
        at async Promise.all (index 2)
        at async Promise.all (index 0)
        at async Promise.all (index 0)
        at async Promise.all (index 0) {
      message: 'The "path" argument must be of type string. Received type undefined',
      locations: [Array],
      path: [Array]
    }
  ],
  data: [Object: null prototype] {
    allMdx: [Object: null prototype] { edges: [Array] }
  }
}

 ERROR #11321  PLUGIN

"gatsby-theme-blog-starter" threw an error while running the createPages lifecycle:

The "path" argument must be of type string. Received type undefined

GraphQL request:13:13
12 |             slug
13 |             image {
   |             ^
14 |               publicURL
```

### Steps to reproduce
I'm using MDX inside gatsby themes and my post front matter looks like
```
---
title: Title
description: Description
pagetitle: Page Title
summary: Page Description
date: '2017-01-28'
update_date: '2017-01-29'
tags:
  - technology
label:
  - tech
slug: theme
published: true
image: ./atech-guide.png
---

Page Text

```

Pages for posts are dynamically created using the following query
```
{
    allMdx(
      filter: {frontmatter: {published: {eq: true}}}
    ) {
      edges {
        node {
          fileAbsolutePath
          id
          frontmatter {
            tags
            slug
            image {
              publicURL
            }
          }
        }
      }
    }
  }
```
[...]

PR #18743, opened 2019-10-16
fix(gatsby): Fixed error message when path is undefined

### LINKED ISSUE #18661: Hot reloading error: "path" argument must be of type string
Hot reloading in `gatsby develop` gives this error in my project. Hot reloading sometimes works despite the error (as in, I see the error in the console window where I ran `gatsby develop`, but the website in browser appears to hot reload and work just fine). This error appeared after I updated some (mainly Gatsby) npm dependencies. Stack trace is giving me no ideas on how to debug this. Is it a Gatsby bug?

Note that this error only comes up on hot reload. So I can run `gatsby develop` just fine as long as I don't make any changes to any files. I can also build just fine.

```
[ { TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type undefined
      at assertPath (path.js:39:11)
      at Object.resolve (path.js:1085:7)
      at findLinkedFileNode (/.../node_modules/gatsby/dist/schema/resolvers.js:208:47)
      at resolveValue (/.../node_modules/gatsby/dist/schema/resolvers.js:223:108)
      at /.../node_modules/gatsby/dist/schema/resolvers.js:220:10
      at process._tickCallback (internal/process/next_tick.js:68:7)
    message:
     'The "path" argument must be of type string. Received type undefined',
    locations: [ [Object] ],
    path:
     [ 'allMarkdownRemark', 'edges', 0, 'node', 'frontmatter', 'cover' ] } ]

```

drakeloud:
## Description

PR to fix the issue with the findLinkedFileNode assuming that 'dir' attribute exists on the object. Adds null and undefined check so that Path.resolve is not passed erroneous data. 

## Related Issues

Link to the issue that is fixed by this PR (if there is one)
Fixes #18661

### DIFF
--- a/packages/gatsby/src/schema/resolvers.js
+++ b/packages/gatsby/src/schema/resolvers.js
@@ -199,6 +199,9 @@ const fileByPath = (options = {}, fieldConfig) => async (
   const findLinkedFileNode = relativePath => {
     // Use the parent File node to create the absolute path to
     // the linked file.
+    
+    if (parentFileNode.dir === null || parentFileNode.dir === undefined) return null;
+
     const fileLinkPath = normalize(
       systemPath.resolve(parentFileNode.dir, relativePath)
     )
--- a/packages/gatsby/src/schema/resolvers.js
+++ b/packages/gatsby/src/schema/resolvers.js
@@ -199,8 +199,10 @@ const fileByPath = (options = {}, fieldConfig) => async (
   const findLinkedFileNode = relativePath => {
     // Use the parent File node to create the absolute path to
     // the linked file.
-    
-    if (parentFileNode.dir === null || parentFileNode.dir === undefined) return null;
+
+    if (parentFileNode.dir === null || parentFileNode.dir === undefined) {
+        return null;
+    }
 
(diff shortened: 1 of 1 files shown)


Jc2k:
I believe this also fixes #17469. There is another user hitting this on [StackOverflow](https://stackoverflow.com/questions/58358559/gatsby-develop-crashes-every-time-a-blog-entry-is-changed/58540703#58540703) too.

pieh:
See my comment in the issue - https://github.com/gatsbyjs/gatsby/issues/18661#issuecomment-545912376 - this just silence the error and is not actually fixing it. I'm not yet sure why this happens (we can't find File node that is parent of remark/mdx node)

drakeloud:
Hmm. I'm not exactly sure why either... I can keep looking around to see if I can find the root issue. I'm not super familiar with the Gatsby codebase yet, but I am happy to dig a little deeper.

pieh:
Let's close this PR as this is not correct fix. I'll add some context to the issue about how this should work (and which parts seems not to work anymore).

PR #18607, opened 2019-10-14
fix(gatsby-cli): work around react-redux hidden dependency on react-dom

pieh:
This is super weird - `react-redux` his hidden dependency on `react-dom`, and we didn't caught that because we weren't installing cli globally (it worked because `react-dom` was available in monorepo node_modules or in project directory)

fixes #18603

Issue #18603, opened 2019-10-14
Install react and react-dom globally? When installing the cli.


rob-bar:
When I clean install the cli I get this weird prompts on the absence of react and react-dom.
When I add them globally it works. Are they set as peer dependencies?

Is this normal? Or am I missing something...

```
npm install -g gatsby-cli
/usr/local/bin/gatsby -> /usr/local/lib/node_modules/gatsby-cli/lib/index.js

> core-js@2.6.10 postinstall /usr/local/lib/node_modules/gatsby-cli/node_modules/core-js
> node postinstall || echo "ignore"

Thank you for using core-js ( https://github.com/zloirock/core-js ) for polyfilling JavaScript standard library!

The project needs your help! Please consider supporting of core-js on Open Collective or Patreon:
> https://opencollective.com/core-js
> https://www.patreon.com/zloirock

Also, the author of core-js ( https://github.com/zloirock ) is looking for a good job -)


> gatsby-telemetry@1.1.32 postinstall /usr/local/lib/node_modules/gatsby-cli/node_modules/gatsby-telemetry
> node src/postinstall.js


> gatsby-cli@2.8.0 postinstall /usr/local/lib/node_modules/gatsby-cli
> node scripts/postinstall.js

+ gatsby-cli@2.8.0
added 350 packages from 181 contributors in 22.172s


gatsby new myPersonalBlog
/usr/local/lib/node_modules/gatsby-cli/node_modules/yoga-layout-prebuilt/yoga-layout/build/Release/nbind.js:53
        throw ex;
        ^

Error: Cannot find module 'react-dom'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
    at Function.Module._load (internal/modules/cjs/loader.js:562:25)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Object.<anonymous> (/usr/local/lib/node_modules/gatsby-cli/node_modules/react-redux/lib/utils/reactBatchedUpdates.js:6:17)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)

npm install -g react-dom
[...]

PR #18752, opened 2019-10-17
refactor(gatsby-cli): replace react-redux with React context

vladar:
## Description

This change removes `react-redux` and its hidden `react-dom` dependency from `gatsby-cli`. Instead, it uses React context to propagate redux store state to CLI components

## Related Issues

Re-addresses #18603 

Follow up after #18607


vladar:
Published in gatsby@2.16.4

PR #18753, opened 2019-10-17
Add config for gatsby-transformer-remark

dcgoodwin2112:
It may be helpful to include sample configurations for gatsby-transformer-remark as well as gatsby-plugin-mdx. If a user were to accidentally try to use the gatsbyRemarkPlugins option on the gatsby-trasformer-remark plugin, then the image loading process on markdown files would fail silently without any warnings or errors.

### DIFF
--- a/docs/docs/working-with-images-in-markdown.md
+++ b/docs/docs/working-with-images-in-markdown.md
@@ -138,7 +138,11 @@ npm install --save gatsby-remark-images gatsby-plugin-sharp
 
 Also make sure that `gatsby-source-filesystem` is installed and points at the directory where your images are located.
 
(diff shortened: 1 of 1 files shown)


docs/docs/working-with-images-in-markdown.md
```diff
@@ -167,7 +167,36 @@ module.exports = {
 }
 ```
 
-With this configuration, you can use the default Markdown syntax for images. They will be processed by Sharp and appear as if you placed them in a `gatsby-image` component.
+Here is a similar example using the `gatsby-transformer-remark` plugin instead of `gatsby-plugin-mdx`.
```
marcysutton: I can definitely see this being useful; it would help to break the sections into two subheadings. Can you add an h3 (`###`) for each section for the different inline image configurations?

dcgoodwin2112: No problem! The headings will definitely help to break up the sections better.

dcgoodwin2112:
This patch has been updated. Thank you for the feedback and for reviewing!

laurieontech:
Awesome! Looks good.

PR #18760, opened 2019-10-17
fix(gatsby-plugin-offline): Improve reliability of JS detection

davidbailey00:
## Description

Uses a similar but slightly different approach to JS disabled detection, which doesn't rely on preload working. Fixes issue with the API not being called on Firefox.


packages/gatsby-plugin-offline/src/sw-append.js
```diff
@@ -35,36 +98,4 @@ const navigationRoute = new NavigationRoute(async ({ event }) => {
...
-
-workbox.routing.registerRoute(/\/.gatsby-plugin-offline:.+/, ({ event }) => {
-  const { pathname } = new URL(event.request.url)
-
-  const api = pathname.match(/:(.+)/)[1]
-  MessageAPI[api]()
-
-  return new Response()
-})
+workbox.routing.registerRoute(/\/.gatsby-plugin-offline:.+/, handleAPIRequest)
```
wardpeet: Could you add a comment on when this route is hit?

packages/gatsby-plugin-offline/src/sw-append.js
```diff
@@ -3,13 +3,76 @@
...
+  return new Response(null, {
+    status: 302,
+    headers: {
+      Location: lastNavigationRequest,
+    },
+  })
+}
+
 const navigationRoute = new NavigationRoute(async ({ event }) => {
+  if (event.request.url.match(/\/.gatsby-plugin-offline:.+/)) {
```
wardpeet: Could you add a comment that this check should always be at the top and why :)

Issue #15080, opened 2019-06-24
Why are page-data.json files not content hashed?


mikheevm:
Hi gatsby team :) We've just spent few hours on debugging an issue with content not appearing on the website and turned out that it was because of page-data.json cached.

Why does it not use content hash? Should cache be handled another way?

marco-th:
@mikheevm The Gatsby Team describes [in this blog post](https://www.gatsbyjs.org/blog/2019-06-12-performance-improvements-for-large-sites/) why they are not hashing the page-data.json files anymore. 

I am also running into some issues because of this and would like an option to enable the hashing again.

gaplo917:
> There are only two hard things in Computer Science: cache invalidation and naming things.
-- Phil Karlton

If the new version .md file is incompatible to previous version (i.e. frontmatter schema)
OR
If the new React Component is incompatible to previous page-data.json


Which lead to:
* NEW html getting old page-data json will throw error (i.e. XXX is undefined)
* OLD html getting new page-data json also will hv the same issue

Suggested that gatsby should have a config to allow user toggle  page-data.json hash feature for cache friendly environment. Until someday the website development is stable(only content update, no React code changes), then user can switch it the feature off.

Currently, I have to downgrade to 2.8.8 for rapid agile development

KyleAMathews:
@gaplo917 how do you know you're seeing this error? We include a build hash in each page-data.json so if a client loads a json file and sees it came from a newer build, it'll force a full refresh to get the new component code.

If that's not working that's a bug we need to fix. More details about what's happening and reproduction instructions would be great!

KyleAMathews:
Also in general, you should only be caching files that are in the static folder and .js files as described in the caching docs https://www.gatsbyjs.org/docs/caching/#static-files

The new json files shouldn't be cached. If you're running into troubles, check out your cache-control settings.

gaplo917:
> Also in general, you should only be caching files that are in the static folder and .js files as described in the caching docs https://www.gatsbyjs.org/docs/caching/#static-files
> 
> The new json files shouldn't be cached. If you're running into troubles, check out your cache-control settings.

Thanks for your advice! I have assumed that everything can be cached.

Because gatsby 2.8.8 will generate hash for all page-data.json. This guarantee a new/old html will load new/old/same json

Will try your advice and see the result.

leonfs:
Thank-you everybody for bringing this problem into notice. 

[...]

PR #18763, opened 2019-10-17
Correct caching warning footnote about 'no-cache'

nihakue:
Footnote was incorrectly warning against using 'no-cache', and could use a little additional clarification.

[1]: https://tools.ietf.org/html/rfc7234#section-5.2.2.1
[2]: https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching#no-cache_and_no-store

## Description

Footnote was incorrectly warning against using 'no-cache', and could use a little additional clarification.

## Related Issues

Related to #15080

### DIFF
--- a/docs/docs/caching.md
+++ b/docs/docs/caching.md
@@ -45,7 +45,7 @@ When deploying with Now, follow the instructions in the [Now documentation](http
 
 ---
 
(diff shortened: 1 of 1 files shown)


nihakue:
@jaredsilver

PR #18771, opened 2019-10-17
asyncbits blog added to showcase

princeiiit:
It is a blog containing technical articles

### DIFF
--- a/docs/sites.yml
+++ b/docs/sites.yml
@@ -8022,3 +8022,11 @@
     - Blog
     - Education
   featured: false
+- title: asyncbits
+ built_by: HariRVA
+  built_by_url: "https://github.com/gatsbyjs/gatsby/edit/master/docs/sites.yml"
+  main_url: https://asyncbits.com/
+  url: https://asyncbits.com/
+ description: >
+    It is a blog containing technical articles
+ featured: false


pieh:
Hey @princeiiit! I'm a bit confused about the site - seems like there is something very custom about it - i.e. on top of having regular separate js chunks (that are being preloaded and prefetch), you seem to force single js file - this end up in user actually having to download both:
 - separate js chunks which are later not used - see warnings in dev tools:
   Screenshot 2019-10-21 at 13 48 37
 - your custom single-file bundle (which seems to contain all those chunks in single file)

Additional performance improvement would be to make use of `gatsby-image` for your images - especially on index page where you have thumbnails (i.e. 1344x742 image is being used for element that is at most ~600px x 200px), so there is a lot of wasted bandwidth

docs/sites.yml
```diff
@@ -8022,3 +8022,11 @@
     - Blog
     - Education
   featured: false
+- title: asyncbits
+ built_by: HariRVA
+  built_by_url: "https://github.com/gatsbyjs/gatsby/edit/master/docs/sites.yml"
```
sidharthachatterjee: This needs to be a url for the author, for HariRVA in this case (GitHub profile url is fine too)

sidharthachatterjee:
Closing this for now since it seems stale and the site doesn't fit guidelines that we'd like to prescribe (see comments from @pieh)

Thank you nonetheless for using Gatsby and for taking the time to open this! We'd be happy to see this reopened once cleaned up 🙂

PR #18790, opened 2019-10-18
[docs] added code blocks to highlight code specific to steps 4 and 5 in Using Gatsby Image doc

### LINKED ISSUE #18775: [docs] add more specific file instructions for Using Gatsby Image doc
## Description

Several comments on the [Using Gatsby Image](https://www.gatsbyjs.org/docs/using-gatsby-image/) doc mentioned their difficulty understanding where the GraphQL query should be added in step 4 of the instructions. Currently, the code block example in step 4 just shows a GraphQL query, that query is what would be used in the same code block in step 5, but the instructions don't make that super clear.

image

What the code titles look like:

Screen Shot 2019-10-17 at 10 01 36 AM

## What to Do

I propose a good course of action would be updating the code blocks for step 4 and 5 with a title (like in [other code blocks in the same doc](https://github.com/gatsbyjs/gatsby/blob/master/docs/docs/using-gatsby-image.md#L60)) to state that the code for each step would be in the same file (maybe something like `src/pages/my-page.js`). 

Ideally, we could update the code blocks in both places to show all the code that would be in that file (both the query and the react component), and then add code line highlighting (example [here](https://github.com/gatsbyjs/gatsby/blob/master/docs/docs/using-gatsby-image.md#L65-L72)) to show what code is added at the corresponding step, either step 4 or step 5.

sazhang:
## Description
In using-gatsby-image.md, 
1. under Step 4 - replaced the graphql fragment block with a jsx code block that contains the graphql fragment
2. under Step 5 - added the Img component to the jsx code block from Step 4

## Related Issues
Fixes #18775

### DIFF
--- a/docs/docs/using-gatsby-image.md
+++ b/docs/docs/using-gatsby-image.md
@@ -83,31 +83,30 @@ module.exports = {
 
 ```jsx:title=src/pages/my-dogs.js
 import React from "react"
-import { graphql } from "gatsby"
+import { useStaticQuery, graphql } from "gatsby" // highlight-line
 import Layout from "../components/layout"
 
-export default ({ data }) => {
+export default () => {
+  // highlight-start
+  const data = useStaticQuery(graphql`
+    query MyQuery {
+      file(relativePath: { eq: "images/corgi.jpg" }) {
+        childImageSharp {
+          # Specify the image processing specifications right in the query.
+          fluid {
+            ...GatsbyImageSharpFluid
+          }
+        }
+      }
+    }
+  `)
+  // highlight-end
   return (
     <Layout>
       <h1>I love my corgi!</h1>
     </Layout>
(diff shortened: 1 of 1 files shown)


docs/docs/using-gatsby-image.md
```diff
@@ -81,15 +81,33 @@ module.exports = {
...
+export default ({ data }) => {
+  return (
+    <Layout>
+      <h1>I love my corgi!</h1>
+    </Layout>
+  )
+}
+
+// highlight-start
+export const query = graphql`
```
marcysutton: The screencast embeds use the useStaticQuery hook. It seems like these should match if being expanded for more context.

sazhang: Updated :) Thanks!

PR #18792, opened 2019-10-18
chore(showcase): Add aprendegatsby.com

elrumordelaluz:


pieh:
@horacioh Hey, are you ok with adding this site to showcase? (seems like you are author of it)

horacioh:
hey @pieh! thanks for letting me know!

wow, thanks @elrumordelaluz for adding this!. While I feel flattered than someone else wanted to add this to the showcase, right now the project is pretty simple (just one page).

Eventually it will have more info and videos and stuff, but now, maybe is to simple for the showcase?

I dunno, @pieh if you feel it's ok to add it now, then I'm ok with that! ;)

LekoArts:
> wow, thanks @elrumordelaluz for adding this!. While I feel flattered than someone else wanted to add this to the showcase, right now the project is pretty simple (just one page).

> Eventually it will have more info and videos and stuff, but now, maybe is to simple for the showcase?

I agree with this sentiment and would argue that it's better for both sides to add the page when it's ready. We've also haven't accepted such pages (which are under construction and only have a newsletter signup) in the past. Let's close it for now and @horacioh can add it when it's done! :)

LekoArts:
No need to be sorry, thanks for being so enthusiastic about Gatsby 🎉

elrumordelaluz:
_Gatsby_ makes enthusiastic people / community!

PR #18793, opened 2019-10-18
fix(gatsby-plugin-feed): Remove wrong overriding in plugin-feed

freiksenet:


sidharthachatterjee:
Published in `gatsby-plugin-feed@2.3.19`

Issue #18728, opened 2019-10-16
Gatsby 2.13.50 installing incompatible gatsby-cli


aharris:
## Description

Gatsby 2.13.50 uses a `^` to include `gatsby-cli`. This lead to in npm install causing us to pull in an incompatible version of the gatsby cli.

![image](https://user-images.githubusercontent.com/713398/66927900-9b6f3900-effe-11e9-8fde-fb9f6904ac5a.png)

This version causes the error:

```
The above error occurred in the <ReduxStoreProvider> component:
    in ReduxStoreProvider
    in App
    
React will try to recreate this component tree from scratch using the error boundary you provided, App.
```

### Steps to reproduce

In an existing project running gatsby@2.13.50
- remove package.lock.json
- npm install
- npx gatsby develop

Clear steps describing how to reproduce the issue. Please please please link to a demo project if possible, this makes your issue _much_ easier to diagnose (seriously).

How to Make a Minimal Reproduction: https://www.gatsbyjs.org/contributing/how-to-make-a-reproducible-test-case/

### Expected result

The app builds.

### Actual result

```
The above error occurred in the <ReduxStoreProvider> component:
    in ReduxStoreProvider
    in App
    
React will try to recreate this component tree from scratch using the error boundary you provided, App.
```

### Environment

node v10.16.3
npm 6.9.0

package.json:

```
    "dependencies": {
        "axios": "0.19.0",
        "gatsby": "2.13.50",
        "gatsby-image": "2.2.8",
        "gatsby-plugin-env-variables": "1.0.1",
        "gatsby-plugin-google-analytics": "2.1.21",
        "gatsby-plugin-google-tagmanager": "2.1.13",
        "gatsby-plugin-intl": "0.2.8",
        "gatsby-plugin-manifest": "2.2.4",
        "gatsby-plugin-offline": "2.2.4",
        "gatsby-plugin-postcss": "2.1.4",
        "gatsby-plugin-react-helmet": "3.1.3",
        "gatsby-plugin-react-svg": "2.1.2",
        "gatsby-plugin-sass": "2.1.8",
        "gatsby-plugin-sharp": "2.2.9",
        "gatsby-plugin-typescript": "2.1.2",
        "gatsby-source-filesystem": "2.1.8",
        "gatsby-transformer-sharp": "2.2.5",
        "graphql": "14.4.2",
        "node-sass": "4.12.0",
        "prop-types": "15.7.2",
        "query-string": "5.1.1",
[...]

PR #18797, opened 2019-10-18
fix(gatsby-cli): re-add reporter._setStage as no-op function

pieh:
gatsby has dependency on gatsby-cli using caret version selector, which means that in case of using npm lock file or pinning gatsby version in project's dependency and not using any lock files, it will always install latest gatsby-cli

This means we need to preserve interface for backward compatibility.

Related issue/comment: https://github.com/gatsbyjs/gatsby/issues/18728#issuecomment-542777122


sidharthachatterjee:
Published in `gatsby-cli@2.8.3`

PR #18807, opened 2019-10-18
chore(starters): add iceberg-gatsby-multilang

diogorodrigues:


diogorodrigues:
I added prettier to fix all the JS syntaxe.

diogorodrigues:
> Awesome! Thanks for sharing @diogorodrigues!

Thanks @pieh , It's a pleasure for me to have my starter in Gatsby and be part of Gatsby's maintainers.

pieh:
There is nothing wrong on your side - it seems like we had some problems with taking screenshots (and to not fail entire build, we skip entries that we failed to pull screenshots for) - so this is on us to figure out

PR #18840, opened 2019-10-19
feat(site-showcase-validator): check 'source_url' for private repos

### LINKED ISSUE #18805: [site-showcase-validator] check source_url in entries are public
## Summary

Have a way for the site showcase validator to notify us if an entry has a `source_url` pointing to a private repo.

### Basic example

Say I have a site at gatsby-site.dev and in the entry I have the `source_url` at https://github.com/lannonbr/gatsby-site, but that repo is private. then we should contact the site maintainer to either make the repo public or remove the `source_url` field from the entry.

This could likely be done in the matter of when we go through each entry, if the entry has a source_url, take that URL, send a fetch request to it, and if it doesn't return a 200 response, log it.

lekterable:
## Description

I've added a check for private repositories in `source_url` entry. For non-public repos there's a notice printed and the summary message contains a total number of them. Except for that, I've also refactored the code a little to make it easier to read, moved the `sitesVisited` increment to the end of the loop and made constants use `const` for clarity.

Example run:
Zrzut ekranu 2019-10-19 o 20 26 20

## Related Issues

Closes #18805

### DIFF
--- a/.github/actions/gatsby-site-showcase-validator/index.js
+++ b/.github/actions/gatsby-site-showcase-validator/index.js
@@ -86,8 +86,9 @@ async function run() {
     )
   )
 
-  // If there are any non Gatsby sites, fail (non-zero exit code)
-  process.exit(nonGatsbySiteCount > 0 ? 1 : 0)
+  // If there are any non Gatsby sites or their `source_url` is private, fail (non-zero exit code)
+  const exitCode = nonGatsbySiteCount > 0 || nonPublicRepoCount > 0 ? 1 : 0
+  process.exit(exitCode)
 }
 
 run()
--- a/.github/actions/gatsby-site-showcase-validator/index.js
+++ b/.github/actions/gatsby-site-showcase-validator/index.js
@@ -6,7 +6,7 @@ const chalk = require("chalk")
 
 async function run() {
   // Grab down sites.yml
-  let url =
+  const url =
     "https://raw.githubusercontent.com/gatsbyjs/gatsby/master/docs/sites.yml"
 
   let yamlStr
@@ -19,17 +19,18 @@ async function run() {
   }
 
   // Parse YAML
-  let parsedYaml = yaml.safeLoad(yamlStr, "utf8")
+  const parsedYaml = yaml.safeLoad(yamlStr, "utf8")
 
   let sitesVisited = 0
   let nonGatsbySiteCount = 0
+  let nonPublicRepoCount = 0
   let erroredOut = 0
-  let totalSitesCount = parsedYaml.length
+  const totalSitesCount = parsedYaml.length
 
   // Loop over each site
   for (let site of parsedYaml) {
-    let siteUrl = site.main_url
-
+    const siteUrl = site.main_url
+    const sourceUrl = site.source_url
     let siteHtml
 
     // Fetch site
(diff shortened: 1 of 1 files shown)


lekterable:
btw. do we want to return `non-zero exit code` for the process when there are some `nonPublicRepos` just as we do for `nonGatsbySites`?

lannonbr:
> 
> 
> btw. do we want to return `non-zero exit code` for the process when there are some `nonPublicRepos` just as we do for `nonGatsbySites`?

Yeah I would say so. Given they're listing something that is invalid saying there's a GitHub or other git repo link but it is private is a case I think should be properly managed so yeah return a non-zero exit code.

lekterable:
done, thanks for the answer @lannonbr

.github/actions/gatsby-site-showcase-validator/index.js
```diff
@@ -47,34 +48,47 @@ async function run() {
...
+
+    // Check if provided repository is public
+    if (sourceUrl) {
+      const status = await fetch(sourceUrl).then(({ status }) => status)
+
+      if (status !== 200) {
+        console.log(
+          `${chalk.yellow(`[Notice]`)}: ${
+            site.title
+          } (${siteUrl}) provided a 'source_url', but it's repository is private (${sourceUrl})`
```
muescha: it can be private or a wrong url ...

.github/actions/gatsby-site-showcase-validator/index.js
```diff
@@ -47,34 +48,47 @@ async function run() {
...
     chalk.green(
-      `We visited ${sitesVisited}/${totalSitesCount} sites. Out of them, ${nonGatsbySiteCount} sites were not a Gatsby site and ${erroredOut} errored out when visiting it.`
+      `We visited ${sitesVisited}/${totalSitesCount} sites. Out of them, ${nonGatsbySiteCount} sites were not a Gatsby site, ${nonPublicRepoCount} provided private repositories, and ${erroredOut} errored out when visiting it.`
     )
   )
 
-  // If there are any non Gatsby sites, fail (non-zero exit code)
-  process.exit(nonGatsbySiteCount > 0 ? 1 : 0)
+  // If there are any non Gatsby sites or their `source_url` is private, fail (non-zero exit code)
+  const exitCode = nonGatsbySiteCount > 0 || nonPublicRepoCount > 0 ? 1 : 0
```
muescha: i don't know exactly the execution order but this is clearer: 
```suggestion
  const exitCode = (nonGatsbySiteCount > 0 || nonPublicRepoCount > 0) ? 1 : 0
```

lekterable: execution order is from left to right and there's only one operator so this seems redundant to me

.github/actions/gatsby-site-showcase-validator/index.js
lekterable: should we add an additional check, or modify the message somehow for this case? @lannonbr

lekterable:
Hey @lannonbr, I've tested it locally and I cannot find a way to tell if the repo is private or if a link is just broken as Github returns 404 for both cases (which is probably smart). Do we want to modify the warning message or should we assume that person who is adding the entry will make sure that it works?

.github/actions/gatsby-site-showcase-validator/index.js
muescha: I personally would go for more easy debug for the next one then for redundant chars.

muescha: But that's my personal taste

lannonbr:
Yeah given both return the same error code, I would say maybe say something like "the repo was inaccessible" or something of that matter so then we can then follow up with the maintainer to figure out if the repo was moved or is just private.

.github/actions/gatsby-site-showcase-validator/index.js
lekterable: message is now more generic, thanks for the suggestion

lekterable:
alrighty, that should do it, let me know what you think @lannonbr

PR #18845, opened 2019-10-19
fix: fix bottom navbar on Iphone X

### LINKED ISSUE #18836: mobile fixed bottom nav has wrong height on mobile in safari
## Description

The mobile fixed bottom nav has the wrong height on mobile in Safari.
<br>

### Steps to reproduce

Open the website gatsbyjs.org in the Safari browser and start to scroll to the bottom that the bottom nav is docked to the bottom of the screen.

### Expected result

The nav should have the correct height, the white background should be higher.

### Actual result

The mobile fixed bottom nav has the wrong height on mobile in Safari.

### Environment

Safari Mobile 13.0 running iOS 13.1.3 on an iPhone X.

lekterable:
## Description

Height of the mobile bottom navbar now includes its padding.

How it looks on iPhone X:
IMG_0110

How it looks on iPhone 8:
IMG_0929

## Related Issues

Fixes #18836

### DIFF
--- a/www/src/components/navigation-mobile.js
+++ b/www/src/components/navigation-mobile.js
@@ -51,7 +51,7 @@ const MobileNavigation = () => (
       bottom: 0,
       display: `flex`,
       fontFamily: `header`,
-      height: `headerHeight`,
+      height: `headerHeight + env(safe-area-inset-bottom)`,
       justifyContent: `space-around`,
       left: 0,
       paddingBottom: `env(safe-area-inset-bottom)`,
--- a/www/src/components/navigation-mobile.js
+++ b/www/src/components/navigation-mobile.js
@@ -51,7 +51,6 @@ const MobileNavigation = () => (
       bottom: 0,
       display: `flex`,
       fontFamily: `header`,
-      height: `headerHeight + env(safe-area-inset-bottom)`,
       justifyContent: `space-around`,
       left: 0,
       paddingBottom: `env(safe-area-inset-bottom)`,


www/src/components/navigation-mobile.js
```diff
@@ -51,7 +51,7 @@ const MobileNavigation = () => (
       bottom: 0,
       display: `flex`,
       fontFamily: `header`,
-      height: `headerHeight`,
+      height: `headerHeight + env(safe-area-inset-bottom)`,
```
fk: Awww thanks for taking care of this @lekterable! 🤗 🙏 
As I'm about to give this a shot on localhost — we don't need to `calc` here, and/or a `@supports` media query?

fk: Ah yeah, _looks_ like this fixes things because with this change, we now supply an invalid `height` — not setting it at all anymore:

image

This is fine though — we are explicitly setting the height of the navigation items/hrefs to `headerHeight` anyway 🤦‍♂ 😅 

image

fk: So, I'm assuming we can just remove the whole `height` declaration and things should work fine for iPhone X  — may I ask you to test that @lekterable, and update the PR if this works? Thank you! 🙏

lekterable: OMG 🤦‍♂ I knew it, that was my first thought! I was like "Oh yeah it's really weird that it works without any calc needed, but at the same, it's super cool, maybe they are using some super cool css-in-js library for it or something, so let's just go with it"

lekterable: I'll test with height removed right now!

fk: > maybe they are using some awesome css-in-js library for it or something, so let's just go with it

We _are_ using the awesome https://theme-ui.com/ ((on top of [Emotion](https://emotion.sh/docs/introduction)) which e.g. grabs `headerHeight` from our theme's `sizes` object for us), but it doesn't magically work irt `env` ;-).

> I'll test with height removed right now!

Thank you so much! 🤗

lekterable: yup, on real devices it works as expected with height removed 😅
iPhone X:
IMG_0114
iPhone 8:
IMG_0930 2

fk: Sweet deal! Let's get this in!

PR #18849, opened 2019-10-20
Resolve package.json of plugins using node resolution algorithm

### LINKED ISSUE #18848: Out of spec resolution of plugins
## Description

I've written some plugins, our company policy is to place source files in `./src` and built library files in `./lib`. Our `main` entry in our package.json points at our `./lib/index.js` file.

This is all within spec of the node resolution algorithm and is a pretty common practice (shared with babel, redux, blueprint, etc).

However Gatsby's plugin resolution incorrectly assumes the package.json file is located at the concatenation of the directory of the entry file and `package.json`.

```
Correct (but naively)
node_modules/gatsby-plugin-sharp -> node_modules/gatsby-plugin-sharp/package.json

Incorrect (due to this behaviour)
node_modules/gatsby-remark-videos/lib -> node_modules/gatsby-remark-videos/lib/package.json
```

This behaviour is due to these code paths, here:

https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/bootstrap/load-plugins/load.js#L94

and here:

https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/src/internal-plugins/internal-data-bridge/gatsby-node.js#L69

This behavior is not described in the docs regarding creating a new plugin.

https://www.gatsbyjs.org/docs/files-gatsby-looks-for-in-a-plugin/

Although I suppose the language of:

> ...main is the **name** of the file that will be loaded when your module...

Could be interpreted stating this subfolder concept isn't supported, but that's not very clear. 

I understand that the vast majority (I guess, literally all) of the plugins for Gatsby have followed an unspoken convention of having their entrypoint in the root, but this is at odds with the rest of the node ecosystem. I think the behaviour of Gatsby should be changed to match the wider ecosystem instead of a documentation change to specify this requirement.

This behaviour is "not a problem" when loading plugins from the local folder, given it resolves the package.json from the root of the plugin. This results in a plugin that works in the local plugins folder, but doesn't when open-sourced and installed via npm. 

### Steps to reproduce

To reproduce, have a plugin with an entrypoint that is in a subfolder. Below is a gatsby-starter-project with one such plugin

https://github.com/Mike-Dax/gatsby-plugin-resolution

### Expected result

Plugins should be resolved correctly.

### Actual result

An error stating `Error: Unable to find plugin "<plugin>". Perhaps you need to install its package?`

### Environment

```
  System:
    OS: macOS 10.14.6
    CPU: (12) x64 Intel(R) Core(TM) i7-8700B CPU @ 3.20GHz
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 12.2.0 - /var/folders/dm/pj8g0t455gg16xy2614p_2q40000gn/T/yarn--1571542290497-0.4449111789579836/node
    Yarn: 1.16.0 - /var/folders/dm/pj8g0t455gg16xy2614p_2q40000gn/T/yarn--1571542290497-0.4449111789579836/yarn
    npm: 6.9.0 - ~/.nvm/versions/node/v12.2.0/bin/npm
  Languages:
  Browsers:
    Chrome: 77.0.3865.120
    Firefox: 69.0.3
    Safari: 13.0.2
  npmPackages:
    gatsby: 2.16.1 => 2.16.1
    gatsby-image: 2.2.29 => 2.2.29
    gatsby-plugin-ffmpeg: 0.3.1 => 0.3.1
    gatsby-plugin-manifest: 2.2.23 => 2.2.23
    gatsby-plugin-offline: 3.0.16 => 3.0.16
    gatsby-plugin-page-creator: ^2.1.27 => 2.1.27
    gatsby-plugin-react-helmet: 3.1.13 => 3.1.13
    gatsby-plugin-sharp: 2.2.31 => 2.2.31
    gatsby-remark-copy-linked-files: 2.1.28 => 2.1.28
    gatsby-remark-videos: 0.3.5 => 0.3.5
    gatsby-source-filesystem: ^2.1.33 => 2.1.33
    gatsby-transformer-remark: 2.6.30 => 2.6.30
    gatsby-transformer-sharp: 2.2.23 => 2.2.23
```

## Relevant PRs

https://github.com/gatsbyjs/gatsby/pull/18849

Mike-Dax:
## Description

In short, plugin package.jsons are resolved in an out-of-spec manner in regards to the wider node ecosystem, this PR attempts to address this.

```
Correct (but naively)
node_modules/gatsby-plugin-sharp -> node_modules/gatsby-plugin-sharp/package.json

Incorrect (due to this behaviour)
node_modules/gatsby-remark-videos/lib -> node_modules/gatsby-remark-videos/lib/package.json

```

With this change:

```
node_modules/gatsby-plugin-sharp -> node_modules/gatsby-plugin-sharp/package.json
node_modules/gatsby-remark-videos/lib -> node_modules/gatsby-remark-videos/package.json
```

The symptom of this problem found in the community is an `Error: Unable to find plugin "<plugin>". Perhaps you need to install its package?` error.

I'm not familiar enough with Gatsby internals to know if this is a complete solution.

It should be non-breaking.

## Related Issues

Fixes #18848

## Tests

Windows tests fail due to a mismatch between `<PROJECT_ROOT>` and `C:/Users/circleci/project` in the snapshots. I'm not sure how to resolve this.

```
-     "packageJSONPath": "<PROJECT_ROOT>/packages/gatsby-plugin-page-creator/package.json",
+     "packageJSONPath": "C:/Users/circleci/project/packages/gatsby-plugin-page-creator/package.json",
```


packages/gatsby/src/bootstrap/load-plugins/load.js
```diff
@@ -89,17 +89,20 @@ function resolvePlugin(pluginName, rootDir) {
         ? createRequireFromPath(`${rootDir}/:internal:`)
         : require
     const resolvedPath = slash(path.dirname(requireSource.resolve(pluginName)))
+    const pluginNameIsPath = path.isAbsolute(pluginName)
+    const packageJSONPath = pluginNameIsPath
+      ? `${resolvedPath}/package.json`
+      : requireSource.resolve(`${pluginName}/package.json`)
```
vladar: I suspect the right fix would be to make sure `resolvedPath` points to the base package directory (containing `package.json`) and not to the directory containing the main `index.js` file.

Then we don't need `packageJSONPath` (and don't need to modify `internal-data-bridge`).

pieh: Few notes from our discussion about this:

> @Vladar:
> With current PR it tries loading gatsby-node from <PLUGIN_ROOT>/lib/gatsby-node. With my suggestion - searches  in <PLUGIN_ROOT>/gatsby-node so yeah - it's about what we consider the right thing

This is the question we need to answer. In my opinion it's not very intuitive that for us to use `<PLUGIN_ROOT>/lib/gatsby-node` - we are setting `main`, but `gatsby-node` is not really `main`.

I would lean toward using @vladar's suggestion for now. This will also match docs we have about plugin structure: https://www.gatsbyjs.org/docs/files-gatsby-looks-for-in-a-plugin/ (we look for gatsby-X files in the root of package).

If there will be need to support `gatsby-node` in different location we should provide proper way of doing that (maybe custom package.json fields similar to `main` for `gatsby-X` files)

Thoughts?

vladar: Makes sense to me

vladar:
@Mike-Dax Do you have time to continue working on this PR? If not, can you provide me access to your fork's branch so that we could proceed with this?

Mike-Dax:
@vlader Yeah I'll have some time this weekend.

PR #18871, opened 2019-10-21
refactor(gatsby-transformer-sharp): Use explicit graphql type definition vs inferring

### LINKED ISSUE #17531: [gatsby schema] Convert gatsby-transformer-sharp to use schema customization
Currently `gatsby-transformer-sharp` uses inference and old Gatsby schema customization (`extend-node-type`) for adding fields. There are several obscure errors that happen because of this.

It would be great to move sharp to use new schema customization. We should then also bump the version of transformer and make it rely on latest gatsby that has all the schema customization features.

If there are some features in schema customization that are missing for it to be possible, we should add that to schema customization along the way.

vladar:
## Description

Convert gatsby-transformer-sharp to use schema customization API (vs. relying on type inference)

Fixes #17531

### DIFF
--- a/packages/gatsby-transformer-sharp/src/customize-schema.js
+++ b/packages/gatsby-transformer-sharp/src/customize-schema.js
@@ -51,7 +51,6 @@ const getTracedSVG = async ({ file, image, fieldArgs, cache, reporter }) =>
   })
 
 const fixedNodeType = ({
-  type,
   pathPrefix,
   getNodeAndSavePathDependency,
   reporter,
@@ -200,7 +199,6 @@ const fixedNodeType = ({
 }
 
 const fluidNodeType = ({
-  type,
   pathPrefix,
   getNodeAndSavePathDependency,
   reporter,
@@ -357,19 +355,13 @@ const fluidNodeType = ({
   }
 }
 
-module.exports = ({
-  type,
+const createFields = ({
   pathPrefix,
   getNodeAndSavePathDependency,
   reporter,
   cache,
 }) => {
-  if (type.name !== `ImageSharp`) {
-    return {}
-  }
-
   const nodeOptions = {
-    type,
     pathPrefix,
     getNodeAndSavePathDependency,
     reporter,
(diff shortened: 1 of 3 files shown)


packages/gatsby-transformer-sharp/src/customize-schema.js
```diff
@@ -546,3 +538,35 @@ module.exports = ({
...
+    name: `ImageSharp`,
+    fields: createFields({
+      pathPrefix,
+      getNodeAndSavePathDependency,
+      reporter,
+      cache,
+    }),
+    interfaces: [`Node`],
+    extensions: {
+      infer: true,
```
freiksenet: Let's actually try to get it to `infer: false`. In addition to fields that we add in `createFields`, could you look at sourceNodes code for this plugin and see what fields are created to be inferred?

vladar: There was a discussion that we should keep it for BC but also add the `deprecated` version of `infer` which would warn when new node fields were added without corresponding schema extension.

freiksenet: Ok, we don't have to block this based on this. Could you add all the existing fields we add in the plugin and then we can merge it with infer: true, with a note that we'll deprecate it when deprecated infer lands.

vladar: @freiksenet I compared schemas generated with `infer: false` and `infer: true` and they are already identical, so I assume this plugin already describes all the fields

freiksenet: OK, then I was confused about this :) Sorry about it. This is good to be merged.

packages/gatsby-transformer-sharp/src/gatsby-node.js
```diff
@@ -1,32 +1,14 @@
...
 
-  // Check if there are any ImageSharp nodes. If so add fragments for ImageSharp.
-  // The fragment will cause an error if there are no ImageSharp nodes.
-  if (getNodesByType(`ImageSharp`).length == 0) {
-    return
-  }
-
-  // We have ImageSharp nodes so let's add our fragments to .cache/fragments.
+  // Add fragments for ImageSharp to .cache/fragments.
   await fs.copy(
```
sidharthachatterjee: This is no longer necessary as we run babel on plugin code now. We should be able to remove the copy operation.

packages/gatsby-transformer-sharp/src/gatsby-node.js
```diff
@@ -1,32 +1,14 @@
 const fs = require(`fs-extra`)
 
 exports.onCreateNode = require(`./on-node-create`)
-exports.setFieldsOnGraphQLNodeType = require(`./extend-node-type`)
+exports.createSchemaCustomization = require(`./customize-schema`)
```
sidharthachatterjee: Need to bump dependency on `gatsby` since older versions won't have this API.

vladar: Added in https://github.com/gatsbyjs/gatsby/pull/18871/commits/2fb1c2061f52ef0b934c8fd6c5e3b7fb52c05406

sidharthachatterjee:
Published in `gatsby-transformer-sharp@2.3.0`

PR #18883, opened 2019-10-21
feat(gatsby-remark-images-contentful): add native lazy loading support to contentful images

jdcataldo:
## Description
`gatsby-remark-images-contentful` currently doesn't support lazy loading of images when it generates image tags for Contentful images from markdown. This adds a new option called `loading` to allow for native lazy loading. This uses the same logic from `gatsby-remark-images`.


freiksenet:
Thank you!

PR #18885, opened 2019-10-21
fix: (gatsby-plugin-mdx) - update readme - String reference is supported

aravindballa:
## Description

Update readme as string references are also supported.


aravindballa:
https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-mdx/utils/default-options.js#L28

Issue #18383, opened 2019-10-09
Map errors in internal plugins to activities and set status correctly


sidharthachatterjee:
When a plugin throws an Error in `gatsby-node.js`, the `status` in the redux store `gatsby-cli` is still set to _Success_. This is because 
- don't catch errors in plugins and associate them correctly with an activity
- We don't create an activities for plugin API runners

We should be able to fix this 

- [x] Create activities for API runners  
- [x] Pass in reporter instance that delegates those function calls to the activity
- [x] Listen to _all_ thrown Errors within this scope and make the activity fail

PR #18887, opened 2019-10-21
feat(gatsby): Ensure status is set to Failed for thrown errors or panicOnBuild in plugins

madalynrose:
## Description

This will set status to `Failed` for errored extension calls in `gatsby develop`

Screen Shot 2019-10-21 at 12 18 54 PM

~This only handles if exceptions are thrown in plugins. Because plugins can call `panic` and `panicOnBuild`, we need to extend this functionality to override the reporter passed into the plugins to call them on the activity instead of the reporter.~

This handles thrown Errors and calls to panicOnBuild for API runners

## Related Issues

Addresses #18383

### DIFF
--- a/integration-tests/structured-logging/gatsby-node.js
+++ b/integration-tests/structured-logging/gatsby-node.js
@@ -5,7 +5,7 @@ const sleep = timeout =>
     setTimeout(resolve, timeout)
   })
 
-exports.createPages = async ({ reporter }) => {
+exports.createPages = async ({ actions, reporter }) => {
   const successfulActivity = reporter.createProgress(
     `Successful activity`,
     100,
@@ -33,6 +33,9 @@ exports.createPages = async ({ reporter }) => {
     await sleep(500)
     unsuccessfulActivity.tick(75)
     unsuccessfulActivity.panicOnBuild(`Your car is on fire`)
+    unsuccessfulActivity.done()
+
+    reporter.panicOnBuild(`Your car is on fire`)
   }
 
   if (process.env.PANIC_ON_BUILD) {
--- a/packages/gatsby/src/redux/plugin-runner.js
+++ b/packages/gatsby/src/redux/plugin-runner.js
@@ -8,6 +8,6 @@ emitter.on(`CREATE_PAGE`, action => {
   apiRunnerNode(
     `onCreatePage`,
     { page, traceId: action.traceId, parentSpan: action.parentSpan },
-    { pluginSource: action.plugin.name }
+    { pluginSource: action.plugin.name, activity: action.activity }
   )
 })
--- a/packages/gatsby/src/redux/plugin-runner.js
+++ b/packages/gatsby/src/redux/plugin-runner.js
@@ -8,6 +8,6 @@ emitter.on(`CREATE_PAGE`, action => {
   apiRunnerNode(
     `onCreatePage`,
     { page, traceId: action.traceId, parentSpan: action.parentSpan },
-    action.plugin.name
+    { pluginSource: action.plugin.name }
   )
 })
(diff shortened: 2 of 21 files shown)


pieh:
Hey @sidharthachatterjee, we feel like this is good to review. We didn't implement activity mapping for `sourceNodes` to not increase the scope of PR and this can be done in follow-up

integration-tests/structured-logging/gatsby-node.js
```diff
@@ -33,6 +33,9 @@ exports.createPages = async ({ reporter }) => {
     await sleep(500)
     unsuccessfulActivity.tick(75)
     unsuccessfulActivity.panicOnBuild(`Your car is on fire`)
+    unsuccessfulActivity.done()
+
+    reporter.panicOnBuild(`Your car is on fire`)
```
sidharthachatterjee: 😂

packages/gatsby/src/utils/api-runner-node.js
```diff
@@ -326,21 +329,33 @@ module.exports = async (api, args = {}, pluginSource) =>
...
-          id: `11321`,
-          context: {
-            pluginName,
-            api,
-            message: err instanceof Error ? err.message : err,
-          },
-          error: err instanceof Error ? err : undefined,
-        })
+        if (activity) {
+          activity.panicOnBuild({
```
sidharthachatterjee: Nit pick: Extract this Object out to a constant to prevent this duplication

packages/gatsby/src/utils/api-runner-node.js
```diff
@@ -161,6 +161,9 @@ const runAPI = (plugin, api, args) => {
         },
       }
     }
+    let localReporter = activity
+      ? { ...reporter, panicOnBuild: activity.panicOnBuild.bind(activity) }
```
sidharthachatterjee: We should bind other methods too (in a follow up of course)

packages/gatsby/src/redux/plugin-runner.js
```diff
@@ -8,6 +8,6 @@ emitter.on(`CREATE_PAGE`, action => {
   apiRunnerNode(
     `onCreatePage`,
     { page, traceId: action.traceId, parentSpan: action.parentSpan },
-    action.plugin.name
+    { pluginSource: action.plugin.name, activity: action.activity }
```
sidharthachatterjee: Where are we using `pluginSource`?

packages/gatsby/src/utils/api-runner-node.js
pieh: I wasn't quite sure if we needed to bind `panicOnBuild`, this should be only needed if `this` is used there (which seems like we don't use in `panicOnBuild`, but we do use in other methods - so worth doing check if we didn't break things)

packages/gatsby/src/redux/plugin-runner.js
pieh: We changed signature of `apiRunnerNode` from https://github.com/gatsbyjs/gatsby/blob/366d915ae8db3ba96b60e8cb4deb920c08adc4e7/packages/gatsby/src/utils/api-runner-node.js#L235
so 3rd argument accept object with optional fields instead of single `pluginSource` arg.

This is later used in api runner to not call `onCreatePage` for the plugin that created given page: https://github.com/gatsbyjs/gatsby/blob/366d915ae8db3ba96b60e8cb4deb920c08adc4e7/packages/gatsby/src/utils/api-runner-node.js#L248-L255

packages/gatsby/src/utils/api-runner-node.js
pieh: Oh, yeah - we should be able to just use `localReporter.panicOnBuild` we created in later commits - nice catch!

sidharthachatterjee:
Published in `gatsby@2.17.2` and `gatsby-cli@2.8.5`

PR #18890, opened 2019-10-21
Fixed "we - you" and "our -your

connectrv:
Fixed "we - you" and "our -your" in the file


LekoArts:
Hi, thanks for the PR! However, we want to leave blogposts out of the styleguide in order to not mess with the original intentions of the author and his blogpost. Thanks!

PR #18892, opened 2019-10-21
fix(www): Eliminate horizontal scroll on mobile devices

qmixi:
## Description
 The purpose of this PR is to eliminate horizontal scroll on mobile devices cased by `egghead iframe videos` with hardcoded `600px` width attribute. This PR solves this issue and furthermore, makes egghead iframe videos responsive. 
The maximum width - 600px remains at it was before :) 

Screenshots below present the problem with a horizontal scroll and how it looks with the dynamic height/width calculation.

Without RWD support:
Screenshot 2019-10-21 at 22 34 24
Screenshot 2019-10-21 at 22 34 08
Screenshot 2019-10-21 at 22 33 47

With RWD support:
Screenshot 2019-10-21 at 22 24 11
Screenshot 2019-10-21 at 22 23 55
Screenshot 2019-10-21 at 22 23 45
Screenshot 2019-10-21 at 22 23 35
Screenshot 2019-10-21 at 22 23 26

I verified other pages that use the `egghead-embed.js` component and they work properly with the changes from PR as well.


marcysutton:
Thanks for this! We need to test it on some real devices too, if you get a chance.

qmixi:
> Thanks for this! We need to test it on some real devices too, if you get a chance.

Thank you! :) I run it on my iphone X and there is no longer horizontal scroll. Everything is working properly.
IMG_5800
IMG_5801

www/src/components/shared/egghead-embed.js
```diff
@@ -1,25 +1,42 @@
...
-      </em>
-      .
-    </p>
-  </Fragment>
-)
+const EggheadEmbed = ({ lessonLink, lessonTitle }) => {
+  const [iframeWidth, setIframeWidth] = useState(0)
+  const iframeRef = useRef()
+
+  const handleResize = () => setIframeWidth(iframeRef.current.clientWidth)
```
igogo5yo: Hey, how about adding `useCallback` here?

igogo5yo: Thanks

www/src/components/shared/egghead-embed.js
```diff
@@ -1,25 +1,45 @@
...
+    }
+  }, [iframeRef.current])
+
+  return (
+    <>
+      <iframe
+        ref={iframeRef}
+        className="egghead-video"
+        width={600}
+        height={iframeWidth * 0.57}
```
sidharthachatterjee: Why `0.57` specifically? 🤔 

I assume that's the aspect ratio of the videos generally? What if that changes?

qmixi: @sidharthachatterjee yes, it is the ratio of the height to the width of the video element inside the iframe, which is fixed. I think that it won't change as it's a native ratio of the video. Maybe it's a good idea to create a `constant` variable on the top of this component, so other users will understand what this number stands for? What do you think about this?

sidharthachatterjee: Yup, could definitely be a constant as opposed to a magic number

qmixi: @sidharthachatterjee commit with this improvement added :)

sidharthachatterjee: @qmixi Could also just be `0.5625` or `9/16` :)

www/src/components/shared/egghead-embed.js
```diff
@@ -1,25 +1,47 @@
...
-
-    <p>
-      <em>
-        Video hosted on <a href={lessonLink}>egghead.io</a>
-      </em>
-      .
-    </p>
-  </Fragment>
-)
+const VIDEO_RATIO = 0.57
```
sidharthachatterjee:
```suggestion
const VIDEO_RATIO = 9/16
```

fk:
Thank you @qmixi and @sidharthachatterjee! 🙏 💜 🤗

Issue #10138, opened 2018-11-26
[gatsby-source-wordpress] TypeError: Cannot read property 'id' of undefined 


hensmith:
## Summary

I'm trying to do a `gatsby build` from a Wordpress blog using `gatsby-source-wordpress`. The build dies with a TypeError. Commenting out the line producing the error results in a seemingly working build. I'm hoping there's some common plugin or config change needed that will be immediately apparent to Gatsby veterans 🤞😇

```
  TypeError: Cannot read property 'id' of undefined

  - normalize.js:256 entities.map.e
    [gatsby-source-wordpress]/normalize.js:256:74
```

## Relevant information

The error means that the `.find()` call in the following block of code from [`normalize.js`](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-source-wordpress/src/normalize.js#L261) is returning `undefined`.

```javascript
// TODO generalize this for all taxonomy types.
exports.mapTagsCategoriesToTaxonomies = entities =>
  entities.map(e => {
    // Where should api_menus stuff link to?
    if (e.taxonomy && e.__type !== `wordpress__wp_api_menus_menus`) {
      // Replace taxonomy with a link to the taxonomy node.
      e.taxonomy___NODE = entities.find(t => t.wordpress_id === e.taxonomy).id
      delete e.taxonomy
    }
    return e
  })
```

Here are some of the relevant plugins present in the Wordpress installation. Because the code near the error mentions the menus API, I've tried a few menu API plugins.

* ACF to REST API
* Advanced Custom Fields PRO
* WP-REST-API V2 Menus (also tried WP REST API Menus, and again with neither installed)
* WPML REST API

### Environment (if relevant)

```
  System:
    OS: macOS 10.14.1
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 11.1.0 - ~/.nvm/versions/node/v11.1.0/bin/node
    Yarn: 1.12.3 - /usr/local/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v11.1.0/bin/npm
  Browsers:
    Chrome: 70.0.3538.102
    Safari: 12.0.1
  npmPackages:
    gatsby: ^2.0.19 => 2.0.55
    gatsby-source-wordpress: ^3.0.15 => 3.0.15
  npmGlobalPackages:
[...]

PR #18895, opened 2019-10-21
fix(gatsby-source-wordpress): add undefined check to avoid mapPostsToTagsCategories error

ryanfiller:
This is addressing null checking in the `mapPostsToTagsCategories` function in `normalize.js` for the plugin `gatsby-source-wordpress`.

Some of my posts seem to have categories, others do not, so I was recieving `TypeError - Cannot read property 'id' of undefined` on some of them. This code adds checks.

This is a similar issue as was solved here: https://github.com/gatsbyjs/gatsby/issues/10138

### DIFF
--- a/packages/gatsby-source-wordpress/src/normalize.js
+++ b/packages/gatsby-source-wordpress/src/normalize.js
@@ -240,26 +240,38 @@ exports.mapPostsToTagsCategories = entities => {
 
     let entityHasTags = e.tags && Array.isArray(e.tags) && e.tags.length
     if (tags.length && entityHasTags) {
-      e.tags___NODE = e.tags.map(
-        t =>
-          tags.find(
+      e.tags___NODE = e.tags
+        .map(t => {
+          const tagNode = tags.find(
             tObj =>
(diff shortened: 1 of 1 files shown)


pieh:
This is looking good from sanity perspective, so I'll probably go ahead and merge it.

But this raises the question on why those tags/categories can't be found in tags/categories arrays. Could you maybe add some `console.log('not found tag/category', t)`s in places where you return `undefined` in this PR + one console.log at the top of `mapPostsToTagsCategories` function that lists existing tags and categories (`console.log('tags and categories', { tags, categories })`) to get some insight into this? (this is pretty much follow up, so I'll probably merge PR before this, because it's valid, I'm just not sure what exactly could cause the need for those checks as data should be consistent, and if something reference a tag/category, then this tag/category should exists)

ryanfiller:
@pieh In regards to why those tags/categories can't be found in those arrays, our issue was that some posts don't have neither tags nor categories. I'll update with those console.logs for sure though.

pieh:
Published `gatsby-source-wordpress@3.1.44`

PR #18916, opened 2019-10-22
chore(gatsby-image): Add missing type definitions for GatsbyImageProps

dcastil:
## Description

Some properties defined in the [Gatsby Image prop types](https://github.com/gatsbyjs/gatsby/blob/3776d39e15368fdd1468d4424c93c8b4ffa3415e/packages/gatsby-image/src/index.js#L660-L685) are missing in the TypeScript type definitions.

This PR adds those type definitions and keeps the order of the TypeScript type definitions the same as in the prop types.


dcastil:
@qmixi I think it would make sense to add a comment to the `Image.propTypes` definition to not forget the TypeScript types.

```js
// If you modify these propTypes, please don't forget to change the TypeScript types in ../index.d.ts
```

What do you think? Should I add it to this PR?

pieh:
> I think it would make sense to add a comment to the `Image.propTypes` definition to not forget the TypeScript types.

Yeah, this makes a lot of sense - we have few places where prop types are being referenced (TS types that you updated here and documentation - both https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-image/README.md and https://github.com/gatsbyjs/gatsby/blob/master/docs/docs/gatsby-image.md ) - would be good to mention all of those places

dcastil:
I added the references as GitHub links to the propTypes of both `Image` and `ImageWithIEPolyfill`. Feel free to review the code again. 😊

dcastil:
Wow! https://github.com/gatsbyjs/gatsby/pull/18916#issuecomment-546963025 That's hilarious! Can't think of a better way to make people excited to contribute. 😄

PR #18923, opened 2019-10-22
adding --clean option to gatsby build

2x2xplz:
I previously commented https://github.com/gatsbyjs/gatsby/pull/16887#issuecomment-544351195 that while I don't think `clean` should be part of the default `gatsby build`, it should be an option. This is highly useful on continuous deployment scenarios which offer a single command to run upon each new commit. Thus, `gatsby clean` followed by `gatsby build` is not possible. This PR would make it possible.

To be very clear: THIS PR LIKELY DOES NOT WORK AS-IS. I am a gatsby user, compiling a forked CLI would mess up my main gatsby CLI which I rely upon to keep my live sites updated. I went through the code and edited the CLI command instructions (these probably work) and then the `build.js` file and make a best guess at where the clean command should be inserted, and how to call it. I will not be offended :) if major changes are necessary to the code. I will be disappointed :( if the PR (and the idea behind it) is outright rejected. 

Thank you!


m-allanson:
Thanks for opening this @2x2xplz, I commented on the linked issue on how you can achieve this effect without needing to add a new option to the build command. 

I think that adding this as an option on the build command is likely to be more confusing than helpful, for that reason I'm going to close this PR.

That said - I do appreciate the time taken to dig into this and I'm looking forward to seeing future PRs from you. Thank you!

PR #18925, opened 2019-10-22
fix(gatsby-plugin-netlify): don't add `/undefined` in generated `_headers` file

jonathonlui:
[Gatsby 2.9 removes pages-manifest](https://www.gatsbyjs.org/blog/2019-06-12-performance-improvements-for-large-sites/) which causes `gatsby-plugin-netlify` to generate a `_headers` that includes these lines:

```
...
/undefined
  Cache-Control: public, max-age=31536000, immutable
...
```

This change prevents generating the above /undefined lines.

### Minimum example to reproduce:
Create 2 files:

package.json

```json
{
  "dependencies": {
    "gatsby": "2.9",
    "gatsby-plugin-netlify": "2.1.22",
    "react": "16.11.0",
    "react-dom": "16.11.0"
  }
}
```

gatsby-config.js
```js

module.exports = {
  plugins: [
    'gatsby-plugin-netlify',
  ],
};
```

Then run:

```sh
yarn && yarn gatsby build && cat public/_headers
```

Which outputs `/undefined` :

```
## Created with gatsby-plugin-netlify

/*
  X-Frame-Options: DENY
  X-XSS-Protection: 1; mode=block
  X-Content-Type-Options: nosniff
  Referrer-Policy: same-origin
/undefined
  Cache-Control: public, max-age=31536000, immutable
/webpack-runtime-17f36917792917c25d15.js
  Cache-Control: public, max-age=31536000, immutable
/app-3b4e42cbdf110996b743.js
  Cache-Control: public, max-age=31536000, immutable
/static/*
  Cache-Control: public, max-age=31536000, immutable
/sw.js
  Cache-Control: no-cache
```


packages/gatsby-plugin-netlify/src/__tests__/build-headers-program.js
```diff
@@ -173,6 +173,27 @@ describe(`build-headers-program`, () => {
...
+    }
+
+    await buildHeadersProgram(pluginData, pluginOptions, reporter)
+
+    expect(reporter.warn).not.toHaveBeenCalled()
+    const output = await fs.readFile(
+      pluginData.publicFolder(`_headers`),
+      `utf8`
+    )
+    expect(output).toMatchSnapshot()
```
pieh: Can we add more specific assertion that checks directly if `/undefined` is not in the output?

jonathonlui: Assertion added.

PR #18930, opened 2019-10-23
chore(docs): explain import vs require

### LINKED ISSUE #18927: Improve docs on import vs require
## Summary
As someone with a pretty low understanding of Javascript coming into the project, I don't know much about the difference between `import` and `require` and when to use one over the other. Better documentation on that could be nice. Specifically in [part 7 of the tutorial](https://www.gatsbyjs.org/tutorial/part-seven/#creating-slugs-for-pages) near the text that says `function for creating slugs`

## Relevant information

Currently the only information I've seen about it is in step 2 [here](https://www.gatsbyjs.org/tutorial/part-two/#-include-the-stylesheet-in-gatsby-browserjs), where it says that usually we'll use `import`. However, when I got to part 7, I'm told to use `require` to get the `createFilePath` function from the `gatsby-source-filesystem` package with no explanation why it's using `require`. I thought maybe I could change that to use `import` instead, but that seems to break when I restart the server.

Because this is a different behavior from the rest of the tutorial, an explanation on why it needs to happen this way would be awesome. It could be like some of the "side quest" tutorials that are optional reading and not part of the critical path.

`gatsby-config.js`: N/A 
`package.json`: N/A 
`gatsby-node.js`: 
This works
```javascript
const { createFilePath } = require(`gatsby-source-filesystem`)

exports.onCreateNode = ({ node, getNode }) => {
  if (node.internal.type === "MarkdownRemark") {
    console.log(createFilePath({ node, getNode, basePath: `pages` }))
  }
}
```

But this doesn't
```javascript
import { createFilePath } from `gatsby-source-filesystem`

exports.onCreateNode = ({ node, getNode }) => {
  if (node.internal.type === "MarkdownRemark") {
    console.log(createFilePath({ node, getNode, basePath: `pages` }))
  }
}
```
`gatsby-browser.js`: N/A 
`gatsby-ssr.js`: N/A

dnsbty:
## Description

Adds a little more explanation around `import` vs `require` to clarify why you might choose one over the other. I'm pretty new to this, so any changes or improvements to this would be very welcome.

## Related Issues

Fixes #18927


docs/tutorial/part-two/index.md
```diff
@@ -114,7 +114,7 @@ import "./src/styles/global.css"
 // require('./src/styles/global.css')
 ```
 
-> Note: Both CommonJS (`require`) and ES Module (`import`) syntax work here. If you’re not sure which to choose, we use `import` most of the time.
+> Note: Both CommonJS (`require`) and ES Module (`import`) syntax work here. If you’re not sure which to choose, `import` will usually be the most common. When working with files that will only be run in the node env like `gatsby-node.js`, `require` will need to be used.
```
omonk: I think this should be expanded on a bit more, mainly because to a new-comer to JS and also Gatsby might find it difficult to understand both the difference between import and require and also which environment `gatsby-browser.js` file  is used in. 

In reality it's both, which leads to more questions like, `how does import work here when I'm using a version node that doesn't support esm`.

pieh: While this is valid concern, I'm not sure if diving into this topic in part 2 of tutorial is proper place for it?

pieh: We might need more dedicate resource to explain this in depth (just not in the tutorial)

omonk: Good point! This commons/esm topic here has a lot of other info around it and not the best place for it. Maybe a reference to another resource that would be helpful

sidharthachatterjee:
```suggestion
> Note: Both CommonJS (`require`) and ES Module (`import`) syntax work here. If you’re not sure which to choose, `import` is usually a good default. When working with files that are only run in a Node.js environment however (like `gatsby-node.js`), `require` will need to be used.
```

PR #18934, opened 2019-10-23
chore(gatsby-plugin-sharp): move image processing to a worker

wardpeet:
## Description
Starting implementing new jobs api. Move logic to separate worker file so it can run in parallel.

One caveat:
The queuing part is moved to the worker so multiple jobs with the same inputpath aren't merged anymore. This might make sharp a little bit slower but I don't think we should care.

- [x] add tests


packages/gatsby-plugin-sharp/src/worker.js
```diff
@@ -0,0 +1,48 @@
...
+        return {
+          outputPath: transform.outputPath,
+          args: transform.transforms,
+        }
+      }),
+      job.pluginOptions
+    )
+  )
+    .then(() => callback())
+    .catch(err => callback(err))
```
sidharthachatterjee: Let's add a test for failures to ensure that errors are propagated correctly from the worker to the calling file

packages/gatsby-plugin-sharp/src/scheduler.js
```diff
@@ -137,44 +130,41 @@ function runJobs(
...
-    })
+        boundActionCreators.setJob(
+          {
+            id: jobId,
+            imagesFinished,
+          },
+          { name: `gatsby-plugin-sharp` }
+        )
+      })
+      .catch(() => {})
```
sidharthachatterjee: This should probably reject with the previous message: `Failed to process image ${job.inputPath}`

This is also wrapped in a try catch block so perhaps we can do all error handling there?

DevSide: I guess you have to reject all job deferred here too

pieh:
Should `worker` be the module that handle job queue? Or should it only be responsible for executing the job? I was under the impression that worker should only need to know how to execute the job and wouldn't need to care about scheduling/queueing

packages/gatsby-plugin-sharp/src/scheduler.js
```diff
@@ -92,31 +82,34 @@ const scheduleJob = async (
...
+        // only show progress on build
+        jobs.forEach(() => {
+          if (reportStatus) {
+            bar.tick()
+          }
+        })
+
+        if (totalJobs === jobsFinished) {
+          bar.done()
+          totalJobs = 0
```
pieh: This is not regression, but shouldn't we also set `jobsFinished` to `0` as well here if we are resetting that state?

packages/gatsby-plugin-sharp/src/scheduler.js
wardpeet: I rewrote the pipeline :)

packages/gatsby-plugin-sharp/src/scheduler.js
```diff
@@ -137,44 +130,41 @@ function runJobs(
...
+      transforms: jobs.map(child => {
+        return {
+          outputPath: child.job.outputPath,
+          transforms: child.job.args,
+        }
+      }),
+      pluginOptions,
+    })
+      .then(() => {
+        jobs.map(job => {
```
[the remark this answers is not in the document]
wardpeet: Thanks, I rewrote the pipeline to have a better approach

wardpeet:
I've cleaned up a lot of things.
1) I'm throttling all jobs so I can merge all transforms of the same file together as we did before when the queue was created inside scheduler.js. It shouldn't slow down scheduling too much and give great perf with my previous tactic (it should be as fast as the original implementation).
2) Added tests for the scheduler, we didn't have them before so :tada:

Happy to hear your thoughts

packages/gatsby-plugin-sharp/src/scheduler.js
```diff
@@ -48,31 +62,70 @@ const scheduleJob = async (
...
+
+    // update the job
+    boundActionCreators.setJob(
+      {
+        id: registeredJob.id,
+        imagesCount: transforms.length,
+      },
+      { name: `gatsby-plugin-sharp` }
+    )
+    bar.total++
```
muescha: this value is later overwritten in Line 139 - so this is obsolete here?

```js
  imagesToProcess++
  bar.total = imagesToProcess
```

packages/gatsby-plugin-sharp/src/scheduler.js
```diff
@@ -1,88 +1,134 @@
...
+        })
+          .then(() => {
+            job.deferred.resolve()
+          })
+          .catch(job.deferred.reject)
+          .finally(() => {
+            if (bar) {
+              job.args.operations.forEach(() => {
+                bar.tick()
+              })
```
muescha: `progress.tick` can take an amount, if also the reporter.createProgressbar has the same tick(value) then we can do:
```js
bar.tick(job.args.operations.length)
```

packages/gatsby-plugin-sharp/src/worker.js
```diff
@@ -0,0 +1,48 @@
...
+}, 1)
+
+/**
+ * @param {WorkerInput} job
+ */
+module.exports = job =>
```
KyleAMathews: Why does the worker have a queue? My understanding of what we're building is that workers are ephemeral (cloud functions can stop working at any time) so they can die on errors, etc. so state shouldn't be stored in the worker.

wardpeet: Sadly we can't move the queue outside of a worker as it needs to know how many jobs it can execute at once. If we want to move it outside of the worker, we have to add concurrency to our jobs. This is hard because concurrency might vary on cloud vs local. We could say concurrency for a job is only used locally, and Infinity is used when doing cloud. Than it's up to cloud's pubsub to handle it.
```suggestion
module.exports = job =>
```
```suggestion
module.exports = job =>
```
```suggestion
module.exports = job =>
```

I'll give an example with images. Sharp is currently set to only process 1 inputFile and as many transforms of that file as it can for performance reasons. On cloud, we technically have an infinity amount of concurrency. When we put the concurrency in the worker we fix this problem as the worker knows how many it can execute at once. For local this means we only run one file at the time. For cloud, one function can run one input file at the time, which still gives us an infinity amount of parallel execution.

If we want to move it outside of the worker we'll probably want to change the api to something like
```
const queue = createJobQueue('IMAGE_PROCESSING', { concurrency: 1 })
queue .enqueue(args)
```
As in my opinion `enqueueJob('IMAGE_PROCESSING', args, { concurrency: 1})` is weird as each job potentially could change concurrency.

Does this make sense?

wardpeet: Sadly we can't move the queue outside of a worker as it needs to know how many jobs it can execute at once. If we want to move it outside of the worker, we have to add concurrency to our jobs. This is hard because concurrency might vary on cloud vs local. We could say concurrency for a job is only used locally, and Infinity is used when doing cloud. Than it's up to cloud's pubsub to handle it.

I'll give an example with images. Sharp is currently set to only process 1 inputFile and as many transforms of that file as it can for performance reasons. On cloud, we technically have an infinity amount of concurrency. When we put the concurrency in the worker we fix this problem as the worker knows how many it can execute at once. For local this means we only run one file at the time. For cloud, one function can run one input file at the time, which still gives us an infinity amount of parallel execution.

If we want to move it outside of the worker we'll probably want to change the api to something like
```
const queue = createJobQueue('IMAGE_PROCESSING', { concurrency: 1 })
queue.enqueue(args)
```
As in my opinion `enqueueJob('IMAGE_PROCESSING', args, { concurrency: 1})` is weird as each job potentially could change concurrency.

Does this make sense?

packages/gatsby-plugin-sharp/src/scheduler.js
```diff
@@ -1,88 +1,132 @@
...
+  } else {
     const jobId = uuidv4()
+    const deferred = pDefer()
+    scheduledPromise = deferred.promise
+
+    // Save a job in our queue
+    toProcess.set(job.inputPath, {
+      id: jobId,
+      args: {
+        inputPath: job.inputPath,
```
sidharthachatterjee: We need to update this to conform to the new spec

packages/gatsby-plugin-sharp/src/worker.js
sidharthachatterjee: Resolving this for now, we'll revisit this later if necessary

packages/gatsby-plugin-sharp/src/index.js
```diff
@@ -53,15 +53,12 @@ function queueImageResizing({ file, args = {}, reporter }) {
...
-    `static`,
+  const outputDir = path.join(process.cwd(), `public`, `static`)
+  const outputFilePath = path.join(
     file.internal.contentDigest,
-    argsDigestShort
+    argsDigestShort,
+    imgSrc
   )
-  const filePath = path.join(dirPath, imgSrc)
-  fs.ensureDirSync(dirPath)
```
sidharthachatterjee: We need to add this back (for a sanity check)

sidharthachatterjee:
```suggestion
  fs.ensureDirSync(outputDir)
```

packages/gatsby-plugin-sharp/src/scheduler.js
```diff
@@ -1,88 +1,140 @@
...
-
-  return { isQueued }
+  boundActionCreators.endJob({ id: job.id }, { name: `gatsby-plugin-sharp` })
 }
 
+const executeJobs = _.throttle(
+  (pluginOptions, boundActionCreators) => {
+    toProcess.forEach(job => {
+      const { task } = job
+      toProcess.delete(task.inputPaths[0])
```
sidharthachatterjee: Let's leave a comment for failure here?

```suggestion
      // We don't add this back even if it fails at the moment 
      toProcess.delete(task.inputPaths[0])
```

packages/gatsby-plugin-sharp/src/scheduler.js
```diff
@@ -1,88 +1,140 @@
...
+  if (reportStatus && !bar) {
     bar = createProgress(`Generating image thumbnails`, reporter)
     bar.start()
   }
-  totalJobs += 1
 
-  if (!isQueued) {
-    // Create image job
+  // Check if the output file already exists so we don't redo work.
+  if (existsSync(job.outputPath)) {
```
sidharthachatterjee: Add a TODO to remove once we hoist the check for if a job needs to be run or not to the job manager

```suggestion
  // TODO: Remove this once we hoist the check for if a job needs to be run or not to the job manager
  if (existsSync(job.outputPath)) {
```

packages/gatsby-plugin-sharp/src/scheduler.js
```diff
@@ -1,88 +1,140 @@
...
+const cleanupJob = (job, boundActionCreators) => {
+  if (bar) {
+    bar.tick(job.task.args.operations.length)
   }
 
-  // Check if the output file already exists so we don't redo work.
-  if (existsSync(job.outputPath)) {
-    return { existingPromise: Promise.resolve(job) }
-  }
+  imagesFinished += job.task.args.operations.length
```
sidharthachatterjee: Nit: Maybe rename `imagesFinished` to `imageCount`?

packages/gatsby-plugin-sharp/src/__tests__/scheduler.js
```diff
@@ -1,20 +1,217 @@
...
+    const workerMock = require(`../worker`).IMAGE_PROCESSING
+    workerMock.mockReturnValue(Promise.resolve())
+
+    const throttledFn = jest
+      .fn()
+      .mockImplementationOnce(() => {})
+      .mockImplementationOnce((cb, args) => cb(...args))
+    jest.doMock(`lodash`, () => {
+      return {
+        throttle: cb => (...args) => throttledFn(cb, args),
```
sidharthachatterjee: hacky boi

sidharthachatterjee:
Published in `gatsby-plugin-sharp@2.3.0`

PR #18935, opened 2019-10-23
fix(gatsby): ensure `@childOf` adds convenience fields before root field arguments

### LINKED ISSUE #18872: Schema customization: `@childOf` directive doesn't cover an edge case with convenience argument for the root field
## Description

Schema customization API and a new `@childOf` directive won't add a convenience argument for the root field (only happens when there are no nodes of the type). It does add a convenience _field_ but not an _argument_.

Follow up after #18871
Related to #15305

### Steps to reproduce

- create a site with `gatsby-transform-sharp` plugin (e.g. default starter)
- delete all static images (so that there was no single node of `ImageSharp` after sourcing)
- check graphql schema

### Expected result

```graphql
type Query {
  file:(childImageSharp: ImageSharpFilterInput) File
}

type File {
  childImageSharp: ImageSharp
}
```

What should happen?

### Actual result

```graphql
type Query {
  # Note the searchable argument is missing even though convenience 
  # field `childImageSharp` is added in the `File`
  # This doesn't happen when there is at least one ImageSharp node
  # (in this case argument exists).
  file: File
}

type File {
  childImageSharp: ImageSharp
}
```

### Notes
Looks like the call sequence during schema building causes root field arguments to be added before convenience child fields.
I think it's `addImplicitConvenienceChildrenFields` which causes this as it relies on nodes presence.

vladar:
# Description

This fixes an edge case situation when `@childOf` directive is applied after root field arguments are created (and thus convenience arguments are missing in the schema).

Since we use a schema composer, changing the ordering of calls seems to be enough. Tests are green and nothing seems to be affected. But I may be easily missing something here.

Fixes #18872

### DIFF
--- a/packages/gatsby/src/schema/extensions/__tests__/child-relations.js
+++ b/packages/gatsby/src/schema/extensions/__tests__/child-relations.js
@@ -872,6 +872,29 @@ describe(`Define parent-child relationships with field extensions`, () => {
     )
     await expect(buildSchema()).resolves.toEqual(expect.any(Object))
   })
+
+  it(`adds root field arguments for searchable child fields (when there are no data nodes)`, async () => {
+    dispatch(
+      createTypes(`
+        type Parent implements Node {
+          id: ID!
+        }
+        type ChildWithoutSourceNodes implements Node @childOf(types: ["Parent"]) {
+          name: String
+        }
+      `)
+    )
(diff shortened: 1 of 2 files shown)


sidharthachatterjee:
Published in `gatsby@2.17.2`

PR #18940, opened 2019-10-23
fix(gatsby): use hashes as module identifiers for webpack caching

wardpeet:
# Description
Use hashes instead of module ids to specify modules. This makes our builds idempotent.


wardpeet:
cc @freiksenet  please test before merging :joy:

Issue #6453, opened 2018-07-14
[v2] Enable SplitChunksPlugin during development


KyleAMathews:
This is just as fast and reduces the amount of JS that needs loaded in development + Tobias said it speeds up HMR.

TylerBarnes:
There's a PR open for this at #18949. 
You can test it by running `npm install gatsby@splitchunksdev`.

sidharthachatterjee:
Closing this as a duplicate of https://github.com/gatsbyjs/gatsby/issues/8342

PR #18949, opened 2019-10-23
feat(gatsby): Enable splitChunks for develop

TylerBarnes:
## Description

We enabled page chunking for develop so it's the same as in production. We also enabled splitChunks but we're not sure if it's actually more performant. It looks like the chunks are still big for the common and vendor chunks.
Rebuilding these big chunks might still make the build slow.
We tested this on `gatsby-starter-default` and it seemed fine but we're not sure if this would be slow on a larger site.

You can test it by running `npm install gatsby@splitchunksdev`.

## Related Issues

https://github.com/gatsbyjs/gatsby/issues/6453


TylerBarnes:
We did some quick tests using reactjs.org and there didn't seem to be a noticeable difference between using splitChunks vs not using it. This is probably still a good change as this is more similar to production.

packages/gatsby/cache-dir/page-renderer.js
```diff
@@ -16,9 +17,13 @@ class PageRenderer extends React.Component {
       loader: publicLoader,
     })
 
+    const AsyncPageComponent = loadable(() =>
+      this.props.pageResources.component()
```
muescha: you are sure it's a function and not the value itself? in line 26 it was before not a function
```suggestion
      this.props.pageResources.component
```

TylerBarnes: @muescha I believe it became a function due to another change in this PR. @wardpeet and I were working on this together and we were surprised that it was a function as well.

muescha: so then it's ok

packages/gatsby/src/utils/webpack.config.js
```diff
@@ -271,6 +271,13 @@ module.exports = async (
         use: [{
           loader: require.resolve(`./reach-router-add-basecontext-export-loader`),
         }],
+      },
+      {
+        test: (input) => {
```
wardpeet: @TylerBarnes I think we only want to add this in develop mode so when `stage === 'develop'`

pvdz:
Can you check this on the https://github.com/thechicagoreporter/govbook/blob/master/gatsby-node.js site? No images, 24k pages (3x8k). I'm curious about possible regressions at scale.

packages/gatsby/src/utils/webpack.config.js
TylerBarnes: updated this, thanks @wardpeet !

packages/gatsby/cache-dir/page-renderer.js
```diff
@@ -16,9 +17,17 @@ class PageRenderer extends React.Component {
       loader: publicLoader,
     })
 
+    let PageComponent
+
+    if (process.env.NODE_ENV === `production`) {
+      PageComponent = this.props.pageResources.component
+    } else {
+      PageComponent = loadable(() => this.props.pageResources.component())
+    }
```
TylerBarnes: Not sure if anyone has an opinion on this. page-renderer.js is being used for both development and production but because we're using `react-hot-loader-loader` in development only (in webpack.config.js), loadable needs to be used here in development as well.

TylerBarnes:
@wardpeet is this still valuable? We should try to get this in soon or close this otherwise. I've been so busy with other things I completely forgot about this. But I'm also not really sure what the next steps are for it

wardpeet:
i'll take over :)

TylerBarnes:
Thanks @wardpeet 😁much appreciated

Issue #6723, opened 2018-07-24
[www/showcase] Make site detail modal responsive


fk:
Currently the details modal shown for window widths 750px and larger is hardcoded to that same width of 750px: https://github.com/gatsbyjs/gatsby/blob/49c52f5e78d56083a0cb45e8cd9306fcd0b92de4/www/src/components/layout.js#L74-L94

This means when reducing the window width below 750px, the modal content won't adapt to the screen width, resulting in first links to the next and previous sites to be cut off, and finally the modal content itself.

Let's fix that!

ryanditjia:
At narrower viewport, we could place the prev and next sites _under_ the featured site. I can take this. Are we drawing the line at flexbox or can we use grid?

fk:
We still want to do this! ✌️

riencoertjens:
🤔hello, is this thing on? can I have a go at this? is there a design? for it?

gillkyle:
Hey @riencoertjens! I'm not aware of any design, if you'd like to work on it feel free to make a PR perhaps implementing the suggestion above to add the next/prev links below the site. In response to @ryanditjia's question about flexbox/grid we're happy to use either or whatever makes sense!

I'll mark this as Hacktoberfest - Claimed, if you are unsure of what you've got, feel free to post a screenshot or a link to your PR or branch and someone can take a look at it too to keep all the design consistent with other elements on the site 🙂

Italox:
@riencoertjens Are you working on this? Need some help?

riencoertjens:
I'm sorry I got a sudden flood of work right before I wanted to tackle this... Hoping to do it this weekend

riencoertjens:
@Italox if you want you can go ahead and do this, unfortunately I won't have time for it.

brockduncan:
Hi, I saw this hanging out here and thought I'd take a stab at it.

fk:
Fixed in #18956 by @brockduncan!
Thanks everybody! 🙏 

(Oh … and hey @riencoertjens :( 🙏)

PR #18956, opened 2019-10-23
Update make details modal responsive

brockduncan:
## Description

Adds a flex wrapper around the modal previous/next buttons and modal content. Changes width to maxWidth and increases it to 1050px to include the width of the buttons. Changes the order of those flex items at the medium breakpoint.

## Related Issues

#6723


fk:
PS: If you'd like to, you could create an issue and/or a follow-up PR that takes care of removing the rotation of the previous and next items on small screens (and any other idea you might irt layout)?

image

Let's get this in now though!

brockduncan:
Cool, thanks @fk !!
Yep, I can create an issue for that.

PR #18969, opened 2019-10-24
update highlight to focus on changed piece

dnsbty:
## Description
It was a little hard to find what I was supposed to be changing here since the whole thing was highlighted. This updates the highlights to focus on the change in the `Event` component in this code block in the tutorial.

## Related Issues
No issue. Just something I saw while I was going through the tutorial.


laurieontech:
Thanks for that attention to detail while you worked through the tutorial and for submitting a PR! Going to merge this now.

PR #18973, opened 2019-10-24
fix(gatsby): fix hashes used in webpack for output

wardpeet:
## Description
Made sure webpack build results are the same when source hasn't changed. I've changed a few config changes to make this work.

1. extract styles.js again, it was overridden by the new splitchunks
2. disable HMR for minicssextract loader in production. This plugin adds a date.now() to the code which means every build is different.
https://github.com/webpack-contrib/mini-css-extract-plugin/blob/50434b55c9df9cd3c26f2243e23df83e2964c720/src/loader.js#L24-L38
3. use moduleIds hashed to enable long term caching as ids are hashed instead of the default behavior "resolving order by default".

I've tested this on the default starter and builds aren't changing 🎉


wardpeet:
Seems like changing directories can change this behavior.

packages/gatsby/src/utils/webpack.config.js
```diff
@@ -497,6 +501,8 @@ module.exports = async (
             test: /\.(css|scss|sass|less|styl)$/,
             chunks: `all`,
             enforce: true,
+            // this rule trumps all other rules because of the priority.
+            priority: 10,
```
sidharthachatterjee: The z-index of webpack builds

packages/gatsby/src/utils/webpack.config.js
```diff
@@ -469,6 +469,10 @@ module.exports = async (
       runtimeChunk: {
         name: `webpack-runtime`,
       },
+      // use hashes instead of ids for module identifiers
+      // TODO update to deterministic in webpack 5 (hashed is deprecated)
+      // @see https://webpack.js.org/guides/caching/#module-identifiers
+      moduleIds: `hashed`,
```
sidharthachatterjee: Why is it deprecated though? Is the recommendation to use something else (that is deterministic as well)?

wardpeet: The value is "deterministic" not "hashed". It's a new value in webpack 5

sidharthachatterjee: Oh alrighty

sidharthachatterjee:
Tests failing because CircleCI can't clone cause port 22 is timing out on GitHub 🤕

PR #18985, opened 2019-10-24
feat(gatsby-source-shopify): add SEO field to articles

stoufa88:
Shopify storefront api now has a graphql field for article SEO (title and description)
https://help.shopify.com/en/api/storefront-api/reference/object/article/seo?api[version]=2019-10


pieh:
Checked this locally, and works nicely. I'm not quite sure how shopify API versioning works, with my test store I was getting `2019-07` api version where it was working. I'll merge this in as I don't have a way to test this with older API versions.

On https://help.shopify.com/en/api/versioning it says:
> If your app calls a stable version that is no longer supported, then Shopify falls forward and responds to your request with the same behaviour as the oldest supported stable version. For example, in April 2020, API requests that call version 2019-04 will be served version 2019-07, because that will be the oldest supported stable version.

And `2019-07` seems like oldest supported API, so I think we should be good here

PR #19000, opened 2019-10-25
docs: fix code snippet in working-with-video.md

dorelljames:
## Description

Fixes the `Video` component code. Following it blindly like copy & pasting the code, say for beginners will cause an error when used in `404 page` as per the example in the guide.

I have also corrected the path of where the file should be created and its syntax highlighting which is JSX.

## Related Issues

Related to #18284


dorelljames:
Pleasure's mine @pieh 😊

PR #19004, opened 2019-10-25
feat(gatsby): Delete unused page-data.json files and .html files

### LINKED ISSUE #17402: page-data.json files not deleted with pages
## Description

I left a comment in #10844 but realized this is actually a separate issue.
When creating a page from a node and then deleting that node in a subsequent build, the page is deleted but it's `page-data.json` remains. The result of this is that pages aren't fully deleted. The html for the page is removed but the pages data remains. If you visit a page that should be deleted you'll see the 404 page, and then the deleted page will load afterwards.

This is related to the global page manifest updates in Gatsby 2.9.0 (https://www.gatsbyjs.org/blog/2019-06-12-performance-improvements-for-large-sites/), as the issue isn't present in v2.8.8  (the last version before 2.9.0) since there is a `manifest.json` file instead of individual `page-data.json` files.

If I manually go into the `public/page-data` directory and delete the corresponding `page-data.json` file, the issue goes away (although Gatsby still tries to fetch this file).

### Steps to reproduce

- Create a page from a node in a GraphQL query in `gatsby-node.js` using the `createPages` api
- Run `gatsby build`
- Make sure the node that created that page will be fully deleted or not recreated in the next build
- Run `gatsby build && gatsby serve`
- Visit the deleted page to see it load in after the 404 page

### Expected result

The corresponding `page-data.json` file should be deleted and the deleted page shouldn't load.

### Actual result

The deleted page loads.

### Environment

```

  System:
    OS: macOS 10.14.5
    CPU: (8) x64 Intel(R) Core(TM) i5-8259U CPU @ 2.30GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 10.15.3 - ~/.nvm/versions/node/v10.15.3/bin/node
    Yarn: 1.16.0 - /usr/local/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v10.15.3/bin/npm
  Languages:
  Browsers:
    Chrome: 76.0.3809.132
    Firefox: 68.0.1
    Safari: 12.1.1
  npmGlobalPackages:
    gatsby-cli: 2.7.44

```

For now I'm just using Gatsby@2.8.8 but this seems like an important one! I noticed this affecting a couple of my projects already as well as a barebones test repo I made with `gatsby-source-wordpress`.

TylerBarnes:
## Description
This PR removes unused page-data.json files to prevent 404 pages from loading them when the corresponding html file is deleted.

## Related Issues
fixes #17402


wardpeet:
I would revert https://github.com/gatsbyjs/gatsby/pull/19004/commits/dff4d3ba43d204c38963c4b3b651c579a6c36e13 I don't see much point doing this right now. Selectively removing html only doesn't fix the issue, we also need to remove other assets if we want to be consistent. I wouldn't do that in this PR.

e2e-tests/production-runtime/package.json
```diff
@@ -5,8 +5,8 @@
   "author": "Kyle Mathews <mathews.kyle@gmail.com>",
   "dependencies": {
     "cypress": "3.4.1",
-    "gatsby": "^2.0.118",
-    "gatsby-plugin-manifest": "^2.0.17",
+    "gatsby": "2.17.5-dev-1572285531830",
+    "gatsby-plugin-manifest": "2.2.24-dev-1572285531830",
```
pieh: ops ;) gatsby-dev side effect - we should revert this

.circleci/config.yml
```diff
@@ -198,7 +198,10 @@ jobs:
     steps:
       - e2e-test:
           test_path: e2e-tests/production-runtime
-          test_command: CYPRESS_PROJECT_ID=is8aoq CYPRESS_RECORD_KEY=cb4708d2-1578-4665-9a07-c59f8db28d91 yarn test && CYPRESS_PROJECT_ID=htpvkv CYPRESS_RECORD_KEY=0d734841-c613-41d2-86e5-df0b5968f93f yarn test:offline
+          test_command: CYPRESS_PROJECT_ID=is8aoq CYPRESS_RECORD_KEY=cb4708d2-1578-4665-9a07-c59f8db28d91 yarn test &&
+          yarn test:removed-page &&
+          CYPRESS_PROJECT_ID=htpvkv CYPRESS_RECORD_KEY=0d734841-c613-41d2-86e5-df0b5968f93f yarn test:offline
+
```
pieh: I think we need to keep test_command in single line or lookup how to split lines in circleCI config

e2e-tests/production-runtime/package.json
TylerBarnes: Oops :p, thanks! Removed it

.circleci/config.yml
TylerBarnes: Fixed this with a multiline syntax

packages/gatsby/src/utils/page-data.js
```diff
@@ -22,7 +25,71 @@ const write = async ({ publicDir }, page, result) => {
...
+  })
+
+  const unusedPageDataFiles = allPageDataFiles.filter(pageDataPath => {
+    // string replaces allow us to match page-data paths against public html paths
+    const normalizedPageDataPath = getPageDataHtmlPath(pageDataPath)
+
+    // if this page-data.json file doesn't have a matching html file,
+    // it's not being used
+    const thisPageDataIsUnused =
+      !availablePagePaths.includes(normalizedPageDataPath) &&
```
me4502: This will be super slow when you have a large page count. Can this be a Set?

me4502: This also is occurring for /every/ page twice, so it's searching every page twice, for every page. A Set is definitely a very good idea here :P

TylerBarnes: @pieh and I manually tested this with a 25000 page test site and found that when timing calculating which pages needed to be deleted vs timing actually deleting them, it actually didn't take much time to calculate but took much more time deleting. I can make that change though. I've never used Set's before so seems like a good opportunity to learn about them.

TylerBarnes: I've updated this to use Set.has() instead

TylerBarnes:
@pieh I think this actually has a kind of long-term memory leak. Because it's expensive to delete empty folders after deleting the page-data.json inside, I'm not deleting empty folders right now. I can't delete the folder by default with the page data file, as the folder may also include the page data of nested pages. 
The problem is that these empty folders will stick around and cause globbing to take more time. As more pages are deleted, this will slowly take longer and longer until the cache is cleared.
I think this is still better than deleted content showing up on a live site, but I'm not entirely sure what the best path forward is.
I'll keep thinking on this and try some things out but I'd love to hear your thoughts if you have any ideas.

pieh:
> @pieh I think this actually has a kind of long-term memory leak. Because it's expensive to delete empty folders after deleting the page-data.json inside, I'm not deleting empty folders right now. I can't delete the folder by default with the page data file, as the folder may also include the page data of nested pages.
> The problem is that these empty folders will stick around and cause globbing to take more time. As more pages are deleted, this will slowly take longer and longer until the cache is cleared.
> I think this is still better than deleted content showing up on a live site, but I'm not entirely sure what the best path forward is.
> I'll keep thinking on this and try some things out but I'd love to hear your thoughts if you have any ideas.

If current implementation globs for all `page-data` files, we can figure out which directories will become empty based on list of `page-data` files and pages that are no longer there:

i.e. given this directory structure
```
page-data [dir]
 - foo [dir]
    - page-data.json
 - bar [dir]
    - page-data.json
    - baz [dir]
       - page-data.json
```

let's say we want to delete page-data for `/foo` and `/bar`
having list of page-data.json files:
```
foo/page-data.json
bar/page-data.json
bar/baz/page-data.json
```

we can calculate current directories:
```
foo
bar
bar/baz
```

figure out which files remain:
```
bar/baz/page-data.json
```

calculate directories we expect to be there:
```
bar
bar/baz
```

do diff on before and after and know that we need to delete `foo` directory as well

pieh:
I'm still not sure if we should do glob here at all, and if we shouldn't track state between builds/develop session to know which things need to be removed without the need of expensive fs globbing

TylerBarnes:
I think the problem with doing a diff is that it's likely also expensive to do that. Maybe we should merge this as-is since it's better than not being able to delete pages, and then open another issue and follow up with a PR that tracks state between builds and uses that instead.

packages/gatsby/src/utils/page-data.js
```diff
@@ -22,7 +25,75 @@ const write = async ({ publicDir }, page, result) => {
...
+ * @argument {object}
+ * {object} availablePagePaths: an array of paths to pages
+ * {object} directory: the root of the Gatsby site
+ * @returns {array}
+ */
+const getUnusedPageDataPaths = ({ availablePagePaths, directory }) => {
+  const allPageDataFiles = glob.sync(`./public/page-data/**/page-data.json`, {
+    cwd: directory,
+  })
+  const unusedPageDataFiles = allPageDataFiles.filter(pageDataPath => {
```
pvdz: This might be a problem at scale ...

Rather than explicitly looping through all pages, wouldn't it be more efficient to schedule a delete and cleanup in that step?

Consider a site with 100k pages ... This search will not be fun.

PR #19006, opened 2019-10-25
docs: Update storybook 5.* configuration

selrond:
The config can be simplified as seen in https://storybook.js.org/docs/guides/guide-react/#step-3-create-the-config-file


pieh:
Does moving `configure` call above our mocks/overrides have any impact?

selrond:
@pieh I don't think so. I've been using it in a site I'm building without any issues. Not technically sure though - there's no mention about order of stuff in config file in storybook docs.

PR #19007, opened 2019-10-25
Remove 'shortcut' link type before 'icon'

lewislbr:
## Description

According to [this MDN page](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types), referring to `<link rel="shortcut icon" ...>`:

>The shortcut link type is often seen before icon, but this link type is non-conforming, ignored and web authors must not use it anymore.

So I've removed it from all files where it was present, resulting in `<link rel="icon" ...>`.


lewislbr:
Nice find! Interesting to see where the legacy comes from. And to know that we could omit the icon declaration altogether to defer loading it 🤔.

PR #19012, opened 2019-10-25
feat(netlify): removes content hashes for assets

afenton90:
## Description

contenthashs are removed for assets generated by Gatsby when using `gatsby-plugin-netlify`. Netlify
handles cache invalidation in different ways that do not require contenthash'd files. See https://www.netlify.com/blog/2017/02/23/better-living-through-caching/ for more details. This also reduces the amount of files changed between deploys in Netlify improving deployment times.

## Related Issues

N/A


KyleAMathews:
Netlify doesn't actually handle cache invalidation in different ways. They do etag validation like everyone. They just find it simpler given their architecture to encourage everyone to not upload new versions of files.

Also I'm pretty sure this will break things because a new version of the site would overwrite the old files and then if someone has a site open, they would then attempt to load in the overwritten js bundle which then wouldn't work.

afenton90:
Thanks for taking a look @KyleAMathews. I've done some tests on local sites and if the Gatsby site is updated the `app-data.json` file will have an updated `webpackCompilationHash` which will trigger a full page reload on nav for anyone currently browsing a site.

On a Netlify hosted site that would trigger a re-fetch of resources that have been updated.
I can pair with one of the core team to test out a few scenarios?

KyleAMathews:
How often do we check for the new app-data.json?

Assuming this does work (and your logic seems plausible), this should be opt-in because it does make your site slower.

afenton90:
`app-data.json` is checked on the same sequence as `page-data.json` files. AKA each page pre-load.
On performing tests on larger sites gatsby.org & reactjs.org no deterioration in perf was seen.

Given that this logic is in place to perform a hard re-load when the browser has an older version of the site compared to any origin, are we cool with this feature in the Netlify plugin to go in?

Happy to work with a maintainer to do any testing required.

KyleAMathews:
Doing etag checks is slower than setting files as immutable (which is what the plugin currently does).

I get wanting to prioritize reducing upload times but this is slower for users so shouldn't be the default.

afenton90:
Ok so you would accept this as an option for the  Netlify plugin?

KyleAMathews:
Can you figure out some numbers on the impact? E.g. the same change to a site with this option enabled drops deploy time by x seconds? Some sort of ball park number would be helpful for anyone considering this

afenton90:
Yeah sure. I'll put some examples together using `reactjs.org` and report back here.
I'll go through the examples I mentioned before of a small change and big change.

afenton90:
Would you be able to publish a tag of this branch so I can install it in Netlify builds? Just so I can run the tests giving improved times.

packages/gatsby-plugin-netlify/README.md
```diff
@@ -34,6 +34,7 @@ plugins: [
       mergeCachingHeaders: true, // boolean to turn off the default caching headers
       transformHeaders: (headers, path) => headers, // optional transform for manipulating headers under each path (e.g.sorting), etc.
       generateMatchPathRewrites: true, // boolean to turn off automatic creation of redirect rules for client only paths
+      contentHashes: false, // boolean to remove content hashes for assets generated
```
muescha: - should not be in this first example the default here: `true`, 
(because later it is explained when to change it to `false`) 
- and also add an `optional`-notation to it?

afenton90: Have changed these to your recommendations. Wasn't aware that was the convention.

afenton90:
@KyleAMathews or @sidharthachatterjee could this be released under a tag so I can run some tests in Netlify on the `reactjs.org` site? That way I can make the logs public and link back here too.

packages/gatsby-plugin-netlify/README.md
muescha: See new comment

packages/gatsby-plugin-netlify/README.md
```diff
@@ -34,6 +34,7 @@ plugins: [
       mergeCachingHeaders: true, // boolean to turn off the default caching headers
       transformHeaders: (headers, path) => headers, // optional transform for manipulating headers under each path (e.g.sorting), etc.
       generateMatchPathRewrites: true, // boolean to turn off automatic creation of redirect rules for client only paths
+      contentHashes: true, // optional boolean to remove content hashes for assets generated
```
muescha: Mhhh. Now I read it again. But I think it is a little bit confusing now that readme part with the `true` option after my suggested change

- Now the comment not match the function of this option as true

- Maybe should some more or different notes here - (default is true but for netlify suggested to change to false)

I am a little unsure here. 

I think the default is true so that your PR is not a breaking change?

cc @marcysutton pls review for better wording

afenton90: Correct the default is `true` so it is not a breaking change. The explanation further down should explain the appeal of setting this to `false` for Netlify users.

KyleAMathews:
@afenton90 you can publish to your own npm package temporarily – been traveling so won't have time.

afenton90:
I've done some tests with the change to compare complete deploy times in Netlify before and after. As anticipated on very large sites it has a big impact on smaller sites not so much. I tested both sites by only changing the 404 page and leaving the rest the same:

- 3000 page site:
     - Before:  17 mins.
     - After: 7 mins.
- reactjs.org:
     - Before: 3 min 16sec
     - After: 3 min 10sec

jmymy:
Any update on moving this forward?

jdesherlia:
@afenton90 Did this ever get merged? I was hoping to test this out before we deploy our new site at the end of feb, as it seemed to be exactly what I was originally hoping `gatsby-plugin-netlify` already did.

afenton90:
@jdesherlia unfortunately not. @muescha any idea when this could be re-reviewed?

muescha:
Sorry. I am not from the @gatsbyjs/core team, I do not review approvals

afenton90:
@gatsby/core could this have a re-review?

wardpeet:
Hey @afenton90,

Removing content hashes means we cannot mark these files as immutable anymore. Immutable tells the browser to never fetch the resource again so users will always be able to fetch it from the cache without revalidating it (checkout https://engineering.fb.com/web/this-browser-tweak-saved-60-of-requests-to-facebook/). This is "slower" and can be problematic on bad connections.

All webpack assets from webpack are (at least should) be deterministic so the pages should only change when necessary (when a global file is added). I'll be tweaking these configurations in the next weeks so we put less stuff in the common chunks, this might already fix your issue.

Accepting this change would make user' performance slower which is something we care deeply about at Gatsby and I know you folks at Netlify do too.

Happy to chat more about this or setup a call to find out better solutions to the uploading problem.

jdesherlia:
@wardpeet I believe the proposal here is to provide a choice in this particular plugin to the user, based on how content hashes affect their config on Netlify. Just my $0.02... I know this would positively impact our build times, and I as a user, would prefer to have the option so that I can continually tweak my config to provide the best experience both for build and delivery.

afenton90:
Hey @wardpeet 

Appreciate you looking at this. The intention was to give site authors the easy opt-in to rely on Netlify's CDN and eTags for cache invalidation. https://www.netlify.com/blog/2017/02/23/better-living-through-caching/

If that is better suited to their particular use case to save on upload times in Netlify but still give the user a fast experience. The option is opt-in for this reason.

afenton90:
> All webpack assets from webpack are (at least should) be deterministic so the pages should only change when necessary (when a global file is added).

At the moment changing any file in the project will result in HTML output being touched due to the content hashes on the webpack manifest file. This is one of the main reasons for the PR.

muescha:
@wardpeet i understand your note right: 

- in the case if the hashes are removed, then the `cache-control` header needs also adjusted?

wardpeet:
I did a bunch of webpagetests with the old version and the new one to show the regression users will get. I explicitly tested this on a Nexus 4x and slow 4G to show you what the difference can be. I've attached all test runs that I'm using in the data below (I did more than the one listed to make sure they were consistent).

From this data, I can roughly see that must-revalidate is slower on bad connections. Here is a summary of every site I've tested
* gatsbyjs.org has little to no difference because the phone is busy parsing html & parsing js, so there is time enough to do the ETag check on the server without losing speed
* gatsby-starter-blog shows us an average delay of 250s of starting our javascript execution and leads up to a 470ms delay on replacing the image
* Gatsbygram shows the same as gatsby-starter-blog, a 250s start delay, and a 300ms complete render difference. The last pained hero shifted 800ms, which is massive.
Because of this vast delay, I do not want to put this burden on the users, even if it makes build times slower. I'll be working on bundling, so I'll see if we can make sure we don't have to upload every html each time. I hope this response is enough to convince you to not proceed with this PR.

I'll leave this PR open until Wednesday next week for comments. All feedback is appreciated.

---

**gatsbyjs.org 4G nexus (https://gatsbyjs-dev.netlify.com/)**
[Webpagetest immutable](https://webpagetest.org/result/200221_DF_1bae0512db3b478db2637f9b280a7bf8/)

<details>
<summary>DATA</summary>

|  | Run1 | Run 2 | Run 3 | Run 4 | Run 5 | Run 6 | AVG | 90th percentile | Avg from first-byte | 90th percentile from first-byte |
|-------------------|-------|-------|-------|-------|-------|-------|--------|-----------------|---------------------|---------------------------------|
| First Byte | 0,976 | 1,029 | 0,232 | 1,290 | 0,239 | 0,902 | 0,7780 | 1,1595 | 0,0000 | 0,0000 |
| Start Render | 1,783 | 1,803 | 0,943 | 1,781 | 0,976 | 1,631 | 1,4862 | 1,7930 | 0,7082 | 0,6335 |
| start-js | 2,566 | 2,380 | 1,670 | 2,921 | 1,782 | 2,240 | 2,2598 | 2,7435 | 1,4818 | 1,5840 |
| render-complete | 3,538 | 3,537 | 2,750 | 3,927 | 2,859 | 3,434 | 3,3408 | 3,7325 | 2,5628 | 2,5730 |
| Last Painted Hero | 2,380 | 2,643 | 1,377 | 2,571 | 1,507 | 2,491 | 2,1615 | 2,6070 | 1,3835 | 1,4475 |

</details>

[Webpagetest must-revalidate](https://webpagetest.org/result/200221_N1_ab09519ded92b499d5653930012de078/)

<details>
<summary>DATA</summary>

|  | Run1 | Run 2 | Run 3 | Run 4 | Run 5 | Run 6 | AVG | 90th percentile | Avg from first-byte | 90th percentile from first-byte |
|-------------------|-------|-------|-------|-------|-------|-------|--------|-----------------|---------------------|---------------------------------|
| First Byte | 0,916 | 1,377 | 1,024 | 2,527 | 0,915 | 0,902 | 1,2768 | 1,9520 | 0,0000 | 0,0000 |
| Start Render | 1,61 | 2,144 | 1,743 | 3,213 | 1,802 | 1,631 | 2,0238 | 2,6785 | 0,7470 | 0,7265 |
| start-js | 2,338 | 3,213 | 2,362 | 3,816 | 2,673 | 2,24 | 2,7737 | 3,5145 | 1,4968 | 1,5625 |
| render-complete | 3,382 | 4,359 | 2,698 | 5,086 | 3,755 | 3,434 | 3,7857 | 4,7225 | 2,5088 | 2,7705 |
| Last Painted Hero | 2,103 | 2,891 | 2,914 | 4,323 | 2,393 | 2,491 | 2,8525 | 3,6185 | 1,5757 | 1,6665 |

</details>

<ins>Differences</ins>

<details>
<summary>Diff avg from first-byte</summary>

|   | Start Render | start-js | render-complete | Last Painted Hero |
|-----------------|--------------|----------|-----------------|-------------------|
| immutable | 0,7082 | 1,4818 | 2,5628 | 1,3835 |
| must-revalidate | 0,747 | 1,4968 | 2,5088 | 1,5757 |
| diff | 0,0388 | 0,015 | -0,054 | 0,1922 |

</details>

**gatsby-starter-blog 4G nexus (https://trusting-mclean-d4b23c.netlify.com/)**
> This test was super flaky on must-revalidate, it didn't always end with a 304 and sometimes had a super stable 4G connection.


[Webpagetest immutable](https://webpagetest.org/result/200221_GR_04346eadce12402c3cbc0f255c3eacef/)

<details>
<summary>DATA</summary>

|  | Run1 | Run 2 | Run 3 | Run 4 | Run 5 | Run 6 | AVG | 90th percentile | Avg from first-byte | 90th percentile from first-byte |
|-------------------|-------|-------|-------|-------|-------|-------|----------|-----------------|---------------------|---------------------------------|
| First Byte | 1,033 | 1,012 | 1,357 | 0,955 | 1,318 | 1,002 | 1,112833 | 1,3375 | 0 | 0 |
| Start Render | 1,445 | 1,497 | 1,803 | 1,426 | 1,838 | 1,438 | 1,5745 | 2,6785 | 0,461666667 | 1,341 |
| start-js | 1,583 | 1,657 | 1,996 | 1,601 | 1,958 | 1,63 | 1,7375 | 3,5145 | 0,624666667 | 2,177 |
| render-complete | 1,749 | 1,872 | 2,16 | 1,775 | 2,083 | 1,798 | 1,906167 | 4,7225 | 0,793333333 | 3,385 |
| Last Painted Hero | 2,039 | 2,195 | 2,472 | 2,081 | 2,412 | 2,11 | 2,218167 | 3,6185 | 1,105333333 | 2,281 |

</details>

[Webpagetest must-revalidate](https://webpagetest.org/result/200221_KS_b39509eba1f22e3885d33a332589228d/)

<details>
<summary>DATA</summary>

|  | Run1 | Run 2 | Run 3 | Run 4 | Run 5 | Run 6 | AVG | 90th percentile | Avg from first-byte | 90th percentile from first-byte |
|-------------------|-------|-------|-------|-------|-------|-------|--------|-----------------|---------------------|---------------------------------|
| First Byte | 0,938 | 0,984 | 0,895 | 0,244 | 0,238 | 1,02 | 0,7198 | 1,0020 | 0,0000 | 0,0000 |
| Start Render | 1,329 | 1,433 | 1,337 | 0,685 | 0,678 | 1,412 | 1,1457 | 1,4225 | 0,4258 | 0,4205 |
| start-js | 2,208 | 2,139 | 1,537 | 0,915 | 1,007 | 1,678 | 1,5807 | 2,1735 | 0,8608 | 1,1715 |
| render-complete | 2,352 | 2,292 | 1,818 | 1,06 | 1,161 | 1,907 | 1,7650 | 2,3220 | 1,0452 | 1,3200 |
| Last Painted Hero | 2,904 | 2,929 | 2,403 | 1,449 | 1,594 | 2,487 | 2,2943 | 2,9165 | 1,5745 | 1,9145 |

</details>

<ins>Differences</ins>

<details>
<summary>Diff avg from first-byte</summary>

|   | Start Render | start-js | render-complete | Last Painted Hero |
|-----------------|--------------|----------|-----------------|-------------------|
| immutable | 0,4617 | 0,6247 | 0,7933 | 1,1053 |
| must-revalidate | 0,4258 | 0,8608 | 1,0452 | 1,5745 |
| diff | -0,0359 | 0,2361 | 0,2519 | 0,4692 |

</details>

**gatsbygram 4G nexus (https://quirky-golick-1f07d1.netlify.com/)**


[Webpagetest immutable](https://webpagetest.org/result/200221_VK_2b8682ca9168f09a4731801d81dd7cc2/)

<details>
<summaryAverages</summary>

|  | Run1 | Run 2 | Run 3 | Run 4 | Run 5 | Run 6 | AVG | 90th percentile | Avg from first-byte | 90th percentile from first-byte |
|-------------------|-------|-------|-------|-------|-------|-------|--------|-----------------|---------------------|---------------------------------|
| First Byte | 0,123 | 1,862 | 1,048 | 0,236 | 0,297 | 0,279 | 0,6408 | 1,4550 | 0,0000 | 0,0000 |
| Start Render | 0,782 | 2,493 | 1,597 | 0,787 | 0,839 | 0,832 | 1,2217 | 2,0450 | 0,5808 | 0,5900 |
| start-js | 0,854 | 2,727 | 1,721 | 0,911 | 0,988 | 0,964 | 1,3608 | 2,2240 | 0,7200 | 0,7690 |
| render-complete | 1,43 | 3,333 | 2,057 | 1,275 | 1,329 | 1,288 | 1,7853 | 2,6950 | 1,1445 | 1,2400 |
| Last Painted Hero | 2,402 | 4,22 | 3,014 | 2,408 | 2,339 | 2,318 | 2,7835 | 3,6170 | 2,1427 | 2,1620 |

</details>

[Webpagetest must-revalidate](https://webpagetest.org/result/200221_Q7_3ba766c2e7cb770fd6c5282043b0d47d/)

<details>
<summaryAverages</summary>

|  | Run1 | Run 2 | Run 3 | Run 4 | Run 5 | Run 6 | AVG | 90th percentile | Avg from first-byte | 90th percentile from first-byte |
|-------------------|-------|-------|-------|-------|-------|-------|--------|-----------------|---------------------|---------------------------------|
| First Byte | 0,924 | 0,955 | 0,906 | 0,909 | 0,143 | 1,366 | 0,8672 | 1,1605 | 0,0000 | 0,0000 |
| Start Render | 1,49 | 1,578 | 1,424 | 1,507 | 0,748 | 1,776 | 1,4205 | 1,6770 | 0,5533 | 0,5165 |
| start-js | 1,79 | 1,838 | 1,817 | 1,804 | 0,894 | 2,98 | 1,8538 | 2,4090 | 0,9867 | 1,2485 |
| render-complete | 2,179 | 2,277 | 2,218 | 2,164 | 1,391 | 3,67 | 2,3165 | 2,9735 | 1,4493 | 1,8130 |
| Last Painted Hero | 3,611 | 3,918 | 3,304 | 3,516 | 3,209 | 5,617 | 3,8625 | 4,7675 | 2,9953 | 3,6070 |

</details>


<ins>Differences</ins>

<details>
<summary>Diff avg from first-byte</summary>

|   | Start Render | start-js | render-complete | Last Painted Hero |
|-----------------|--------------|----------|-----------------|-------------------|
| immutable | 0,5808 | 0,72 | 1,1445 | 2,1427 |
| must-revalidate | 0,5533 | 0,9867 | 1,4493 | 2,9953 |
| diff | -0,0275 | 0,2667 | 0,3048 | 0,8526 |

</details>

jdesherlia:
@wardpeet It seems like these test results are all over the map. Also, I think the fundamental thing that is missing in your analysis is that this would be a choice. I'm finding it hard to understand why the gatsby team is fighting so hard to make an option available to developers that will give them more control over how they build and deploy their gatsby site.

I'm sure I am not alone here, but my site is more closely related to gatsbyjs.org, in terms of volume and content type, and the improved build times combined with the negligable difference in performance would be a far more preferable config. I can find 250ms to save elsewhere... I may not be able to find 2 minutes to shave off of builds after a certain point.

muescha:
I guess the prefered way of the Gatsby team is not to overload the core (default) packages with options and give the user the best solution (fast delivery)

In your case I think is the better solution to extract the `gatsby-plugin-netlify` into a own new package (`gatsby-Plugin-netlify-nohashes` with your changes. And you can also publish this package so that every user with the same need can use it. 

that's the benefit of the open plug-in structure of gatsby

wardpeet:
> I guess the preferred way of the Gatsby team is not to overload the core (default) packages with options and give the user the best solution (fast delivery)

Thanks, @muesha, this is precisely what I'm aiming for.

> @wardpeet It seems like these test results are all over the map. Also, I think the fundamental thing that is missing in your analysis is that this would be a choice. I'm finding it hard to understand why the gatsby team is fighting so hard to make an option available to developers that will give them more control over how they build and deploy their gatsby site.
> 
> I'm sure I am not alone here, but my site is more closely related to gatsbyjs.org, in terms of volume and content type, and the improved build times combined with the negligible difference in performance would be a far more preferable config. I can find 250ms to save elsewhere... I may not be able to find 2 minutes to shave off of builds after a certain point.

@jdesherlia, why would we want to add a feature into a plugin that slows down the user experience? Sometimes it's better to give site builders no option than an option that might increase something on the developer's side but decrease on the user's side. Performance is hard, and people don't always think it through what impact it can have on the user.

In the following weeks, I'll be looking into webpack bundling fixes, which should lead to fewer changes to the common & app bundle. In the end, Gatsbyjs.org ships smaller bundles, and it will have about the same results as we see with gatsbygram, ....

If you want this feature as of this moment, I would suggest creating a new plugin, as @muescha suggested. 

Your point about "these test results are all over the map" is valid. I did multiple tests, and they always came into a similar result. These tests show you what bad networking looks like, so if you can improve not going to the network, you should ;) That's why service workers are so important.

I hope this helps

jdesherlia:
@wardpeet @muescha a separate plugin can work, and I will pursue that. It just seems like a lot of overhead for supporting a simple toggle. I am just getting the sense that there is some other motivation for the gatsby team to be so insistent that this will degrade performance in ALL scenarios, when there are clearly tests in the comments above that show this assessment to be false. Not all websites are created equally, and certainly not all Gatsby sites.

Gatsby as an enterprise product ( which seems to be the aim https://www.gatsbyjs.com/pricing/ ) will need to accept that performance is not solely scoped to the end user, and different organizations have different criteria as to what they consider to be acceptable user, maintainer, and developer performance. I strongly believe that supporting an option like the one proposed here will enable more developers to tune gatsby + netlify to work for their specific business needs, allowing more businesses to use this combination of tools effectively.

@afenton90 care to collab on a publishing this fork to the gatsby plugin repo? Im sure folks other than us will find it useful.

KyleAMathews:
@jdesherlia no need to spin conspiracy theories ;-)

We often don't accept complicated changes we don't agree with. Remember we're maintaining lots of stuff which takes a ton of time and it serves everyone's interest to keep the oss team's overhead low so they can work on core features relevant to everyone. We often ask people to move things to plugins.

Also as Ward mentioned, this might not be an issue for long both for the reason he mentioned plus I've heard from Netlify folks that they're working on improvements to dramatically speed up deployments.

KyleAMathews:
Let's close this in favor of creating a separate plugin so y'all can unblock using this. Hopefully it won't be needed for very long.

afenton90:
@jdesherlia as much as it is dissapointing this didn't go in, I don't think creating a fork of this plugin to add the option & configuration would be a worth while addition to the plugin ecosystem, given the time commitment that comes with.
As an alternative, I'll look to write up a blog post on how to do this in Gatsby config files so contenthashes can be disabled per site.

jdesherlia:
@afenton90 were you ever able to write that blog post on how to make this work? Very much looking forward to reading it.

afenton90:
@jdesherlia afraid I never got around to it. I have my blog ready to go soon, so will upload it to that.
Alternatively, if the maintainers think it would be a worthwhile addition to the Gatsby blog I can put it there? I honestly won't be offended if the answer is no though.

Issue #18967, opened 2019-10-24
.gitignore in default starter does not work for .env.development, .env.production, etc. files


kellenmace:
## Description

When I create a new Gasby site using the Default Starter, then add environment variable files other than `.env`, such as `.env.deveolopment` or `.env.production`, they are not git ignored.

### Steps to reproduce

Create a new Gasby site using the Default Starter and add a environment variables file other than `.env`, such as `.env.deveolopment` or `.env.production`, then run `git status` to confirm that it is not being git ignored.

### Expected result

Environment variable files such as `.env.deveolopment` and `.env.production` should be git ignored.

### Actual result

Environment variable files such as `.env.deveolopment` and `.env.production` are not git ignored.

### Environment

System:
    OS: macOS 10.14.6
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 12.3.1 - /usr/local/bin/node
    npm: 6.9.0 - /usr/local/bin/npm
  Languages:
    Python: 2.7.10 - /usr/bin/python
  Browsers:
    Chrome: 78.0.3904.70
    Firefox: 69.0.3
    Safari: 13.0.2
  npmGlobalPackages:
    gatsby-cli: 2.8.5
    gatsby: 2.10.0

kellenmace:
Hey @bvlktech - Thanks for your reply! Yes, I can manually change the `.env` line in my `.gitignore` to `.env*` in my projects so that git starts ignoring the `.env.development`, `.env.production`, etc. files.

I was just thinking that that line should be changed in the repo though, so that anybody who uses the starters doesn't have to add the `*` themselves in order to ignore those files – it'll already be there for them. That's what my PR should accomplish: https://github.com/gatsbyjs/gatsby/pull/18968

bvlktech:
what starter are u using the default one?

kellenmace:
@bvlktech Yes, I typically use the default starter on my projects. Although on that PR, @pieh asked me to update the  `.gitignore` for all the starters so that they all get the update.

bvlktech:
well the default one is made / maintaind by gatsby itsself where as the other starters are by other devs i can change the .gitignore in the def starter but the others are up to those devs if they want to implament it ill also ref this as a discussion so this gets known about ☺️

PR #19015, opened 2019-10-25
feat(starter): expand .gitignore in default starter to ignore .env, .env.production, .env.development etc

bvlktech:
## Description

This is in ref to case [#18967](https://github.com/gatsbyjs/gatsby/issues/18967) in the issues tab.


bvlktech:
srry just saw as the first thing in the list and made a PR my bad

PR #19016, opened 2019-10-25
feat(gatsby): Jobs

sidharthachatterjee:
Heavily work in progress


packages/gatsby-cli/src/reporter/index.js
```diff
@@ -399,9 +399,9 @@ const reporter: Reporter = {
   _setStage() {},
 }
 
-console.log = (...args) => reporter.log(util.format(...args))
-console.warn = (...args) => reporter.warn(util.format(...args))
-console.info = (...args) => reporter.info(util.format(...args))
-console.error = (...args) => reporter.error(util.format(...args))
+// console.log = (...args) => reporter.log(util.format(...args))
```
sidharthachatterjee: These make debugging reducers _so painful_ 😒(you can't dispatch in reducers and structured logging makes console.log dispatch)

packages/gatsby/src/redux/reducers/tasks.js
```diff
@@ -0,0 +1,61 @@
+// TODO: Ensure this is not persisted to disk since tasks are ephemeral
+const initialState = {
+  tasks: new Map(),
```
sidharthachatterjee: `jobs` is taken 😅 

Will figure out how to migrate the name to that later

packages/gatsby/src/redux/reducers/tasks.js
```diff
@@ -0,0 +1,61 @@
...
+}
+
+const runTask = task => {
```
sidharthachatterjee: Still unclear if queue needs to be in core or workers, so no queue at the moment

packages/gatsby/src/redux/reducers/tasks.js
```diff
@@ -0,0 +1,61 @@
...
+        name,
+        args,
+        plugin,
+      }
+      tasks.set(id, payload)
+      //   process.stdout.write(payload)
+      runTask({
+        ...payload,
+        deferred,
+      })
```
pieh: I feel like instead of doing it here, we should have
``emitter.on(`CREATE_TASK`, taskAction => {})`` (reducers shouldn't handle side-effects - and also would fix console.log in reducer headache :P)

pieh: maybe even action creator itself just do this

packages/gatsby/src/redux/reducers/tasks.js
pieh: I initially was thinking of queue in core, but came around - core won't have enough context to know what are optimal params for queue for given task type (concurrency etc), unless this will be part of jobs/tasks api (which would complicate things without much gain)

packages/gatsby/src/redux/reducers/tasks.js
sidharthachatterjee: Oh yeah, of course. This will go into a separate folder for sure.

packages/gatsby/src/redux/actions/public.js
```diff
@@ -1300,4 +1302,35 @@ actions.createPageDependency = (
...
+    args,
+    plugin,
+    deferred,
+  })
+  return deferred.promise.finally(() => {
+    dispatch({
+      id,
+      type: `END_TASK`,
+    })
+  })
```
pieh: For the future - let's make sure we don't rely on promises to dispatch `END_TASK`.

Currently in master/latest we can hit scenario in `gatsby develop` that some images won't be processed: 
If, after initial bootstrap, we update data, which will run queries that will queue image processing and we terminate process after query running and before image processing complete, Gatsby will not finish image processing on next run because it doesn't track jobs between runs (it's also impossible to re-run them with current jobs, as those are used only to tell gatsby there is long running work, but not how to run this)

We would need to persist tasks to our .cache so gatsby knows that it needs to complete them. To persist we need to make sure we only use serializable primitives (and promises are not serializable).

This also will enable us to finally implement image processing on demand in `gatsby develop` which will significantly improve DX (as we can queue jobs, and only ones that users need when navigating development site will be executed).

All of this doesn't need to happen in initial implementation, as this is not one of main goals of it, but let's make sure we are ready to do so in the future.

packages/gatsby/src/redux/reducers/tasks.js
KyleAMathews: maybe job2 :-D

packages/gatsby/src/redux/reducers/tasks.js
wardpeet: the problem here that concurrency depends on the consumer, for gatsby-cloud this can be infinity and for local this can be 4

sidharthachatterjee:
It appears that we might have to update some test mocks

packages/gatsby/src/redux/reducers/tasks.js
KyleAMathews: concurrency also depends on external systems — we'll need to add something so workers can manage backpressure from external systems to not overwhelm them e.g. downloading from a small wp instance or uploading to an API with rate limits.

packages/gatsby/src/redux/actions/public.js
KyleAMathews: Also when jobs are serializable — we can send jobs to the cloud to do

sidharthachatterjee: Yup, they’re serializable now and are persisted between processes and re triggered if incomplete

sidharthachatterjee:
TODO: Figure out how to interrupt job workers on cloud

packages/gatsby/src/redux/reducers/tasks.js
wardpeet: that's something that the worker needs to do or the plugin. job creation is fully async and no way to communicate back to the process during a job. As it will make it a lot harder to run a worker in a sandbox environment

wardpeet:
Closing this PR in favor of #19858. With having more info about cloud & persisting jobs, the approach is a bit different now. It's also a bit easier to start over.

Sorry @sidharthachatterjee, thanks for your work! You inspired me for the changes I'm making. :heart:

Issue #19010, opened 2019-10-25
Extract stack-trace-utils.js to gatsby core utils


sidharthachatterjee:
We currently have duplicate code in `gatsby-cli` and code added in https://github.com/gatsbyjs/gatsby/pull/19009

We should be able to extract that common stack trace sanitisation code into a shared function in `gatsby-core-utils`

v-hal:
I'd like to help with this one.

v-hal:
Pull request here: #19017

PR #19017, opened 2019-10-25
chore(gatsby): Move duplicate node path check to gatsby-core-utils

v-hal:
## Description

Moved duplicate code under the gatsby-core-utils.

## Related Issues
Addresses #19010

### DIFF
--- a/packages/gatsby-core-utils/src/__tests__/path.js
+++ b/packages/gatsby-core-utils/src/__tests__/path.js
@@ -1,4 +1,4 @@
-const { joinPath, isNodePath } = require(`../path`)
+const { joinPath, isNodeInternalModulePath } = require(`../path`)
 const os = require(`os`)
 
 describe(`paths`, () => {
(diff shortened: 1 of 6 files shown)


packages/gatsby-core-utils/src/__tests__/path.js
```diff
@@ -21,4 +21,11 @@ describe(`paths`, () => {
       })
     }
   })
+  describe(`isNodePath`, () => {
+    it(`Matches common node files`, () => {
+      expect(isNodePath(`console.js`)).toBe(true)
+      expect(isNodePath(`http.js`)).toBe(true)
+      expect(isNodePath(`internal/foo`)).toBe(true)
```
pieh: Could we add at least one test path that we expect *not* to be `true`? Something like `/Users/username/dev/project-1/node_modules/packaga-name/index.js` maybe

v-hal: > That was very quick!
> 
> We named this poorly in our internals (`nodePaths`) but it was kinda acceptable, given that it wasn't exported anywhere. Given that we are making this more public API we should take more care in naming this - I'd suggest something like `IsNodeInternalPath` or `IsNodeInternalModulePath` to more clearly convey what this function check

Thanks! I was struggling a bit with the naming. I'll change it to `isNodeInternalModulePath`

v-hal: > Could we add at least one test path that we expect _not_ to be `true`? Something like `/Users/username/dev/project-1/node_modules/packaga-name/index.js` maybe

Added an unhappy path to the unit test.

PR #19019, opened 2019-10-25
Updated  we to you

connectrv:
Updated  we to you


connectrv:
> I don't like the entire structure of the sentence. Something like
> `` During the schema generation phase, a `ProcessedNodeType` is generated in Gatsby. ``
> sounds better to me

Updated.

marcysutton:
"We" means Gatsby is the subject here, not the reader. Can you update it to reflect the purpose of the document? https://www.gatsbyjs.org/contributing/gatsby-style-guide/#use-you-as-the-pronoun

LekoArts:
Let's close this PR for now. If you're up for changing the files according to the above mentioned link, feel free to create a new PR or reply here.

PR #19023, opened 2019-10-25
Update docs unit testing typescript

soosgyul:

### DIFF
--- a/docs/docs/unit-testing.md
+++ b/docs/docs/unit-testing.md
@@ -217,7 +217,7 @@ by running `npm test -- -u`.
 
 ## Using TypeScript
 
-If you are using TypeScript, you need to make two small changes to your
+If you are using TypeScript, you need to make two changes to your
 config.
 
 Update the transform in `jest.config.js`, like so:
--- a/docs/docs/unit-testing.md
+++ b/docs/docs/unit-testing.md
@@ -220,7 +220,7 @@ by running `npm test -- -u`.
 If you are using TypeScript, you need to make two changes to your
 config.
 
-Update the transform in `jest.config.js` to run `jest-preprocess` on files in your project's root directory, where `<rootDir>` is replaced with the correct relative path:
+Update the transform in `jest.config.js` to run `jest-preprocess` on files in your project's root directory, where `<rootDir>` is kept intact as a literal string to dynamically pass the relative path to Jest:
 
 ```js:title=jest.config.js
     "^.+\\.[jt]sx?$": "<rootDir>/jest-preprocess.js",
--- a/docs/docs/unit-testing.md
+++ b/docs/docs/unit-testing.md
@@ -254,9 +254,7 @@ const babelOptions = {
 }
 ```
 
-You may notice that two other options, `testRegex` and `moduleFileExtensions`,
-have been added. Option `testRegex` is the pattern telling Jest which files
-contain tests. The pattern above matches any `.js`, `.jsx`, `.ts` or `.tsx`
+The pattern above matches any `.js`, `.jsx`, `.ts` or `.tsx`
 file inside a `__tests__` directory, or any file elsewhere with the extension
 `.test.js`, `.test.jsx`, `.test.ts`, `.test.tsx`, or `.spec.js`, `.spec.jsx`,
 `.spec.ts`, `.spec.tsx`.
--- a/docs/docs/unit-testing.md
+++ b/docs/docs/unit-testing.md
@@ -217,33 +217,13 @@ by running `npm test -- -u`.
 
 ## Using TypeScript
 
-If you are using TypeScript, you need to make a couple of small changes to your
-config. First install `ts-jest`:
+If you are using TypeScript, you need to make two small changes to your
+config.
 
-```shell
-npm install --save-dev ts-jest
-```
-
-Then update the configuration in `jest.config.js`, like so:
+Update the transform in `jest.config.js`, like so:
 
 ```js:title=jest.config.js
-module.exports = {
-  transform: {
     "^.+\\.[jt]sx?$": "<rootDir>/jest-preprocess.js",
-  },
-  moduleNameMapper: {
-    ".+\\.(css|styl|less|sass|scss)$": "identity-obj-proxy",
-    ".+\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$":
-      "<rootDir>/__mocks__/file-mock.js",
-  },
-  testPathIgnorePatterns: ["node_modules", ".cache", "public"],
-  transformIgnorePatterns: ["node_modules/(?!(gatsby)/)"],
-  globals: {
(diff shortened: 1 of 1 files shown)


docs/docs/unit-testing.md
```diff
@@ -229,16 +229,13 @@ Then update the configuration in `jest.config.js`, like so:
...
-    "^.+\\.jsx?$": "<rootDir>/jest-preprocess.js",
+    "^.+\\.[jt]sx?$": "<rootDir>/jest-preprocess.js",
   },
-  testRegex: "(/__tests__/.*|(\\.|/)(test|spec))\\.([tj]sx?)$",
   moduleNameMapper: {
     ".+\\.(css|styl|less|sass|scss)$": "identity-obj-proxy",
     ".+\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$":
       "<rootDir>/__mocks__/file-mock.js",
   },
-  moduleFileExtensions: ["ts", "tsx", "js", "jsx", "json", "node"],
```
pieh: I see that `jest` have our current/previous `testRegex`/`moduleFileExtensions` settings as default now:

https://jestjs.io/docs/en/configuration.html#modulefileextensions-array-string
> moduleFileExtensions [array]
> Default: ["js", "json", "jsx", "ts", "tsx", "node"]

https://jestjs.io/docs/en/configuration.html#testregex-string-array-string
> testRegex [string | array]
> Default: (/__tests__/.*|(\\.|/)(test|spec))\\.[jt]sx?$

Below section you added about updating `jest.preprocess.js` there are 2 paragraphs explaining those setting fields which now seem redundant, so maybe we should remove them?

soosgyul: Good spot. I've removed those two paragraphs and pushed the changes.

docs/docs/unit-testing.md
```diff
@@ -249,9 +246,15 @@ module.exports = {
...
-contain tests. The pattern above matches any `.js`, `.jsx`, `.ts` or `.tsx`
+Also update `jest.preprocess.js` with the following babel preset to look like this:
+
+```js:title=jest-preprocess.js
+const babelOptions = {
+  presets: ["babel-preset-gatsby", "@babel/preset-typescript"],
+}
+```
+
+The pattern above matches any `.js`, `.jsx`, `.ts` or `.tsx`
```
pieh: We don't have "pattern above" anymore, so if we chose to keep information what tests will be picked up we should mention that `jest` by default will use those instead and link to `jest` config docs

soosgyul: I went ahead and did a bigger cleanup and removed more redundant references.

pieh:
This looks so much cleaner! One thing that I didn't think about before is that this assumes that `@babel/preset-typescript` is installed (which is dependency of `gatsby-plugin-typescript`, so if user use that, they will have this package installed). Given that this babel preset is being used in configuration probably worth leaving a note about this somewhere near the beginning of typescript section.

docs/docs/unit-testing.md
```diff
@@ -217,50 +217,24 @@ by running `npm test -- -u`.
...
+If you are using TypeScript, you need to make two small changes to your
+config.
 
-```shell
-npm install --save-dev ts-jest
+Update the transform in `jest.config.js`, like so:
+
+```js:title=jest.config.js
+    "^.+\\.[jt]sx?$": "<rootDir>/jest-preprocess.js",
 ```
```
marcysutton: I'd also add an example here to remove any ambiguity:
````suggestion

For example, if your test utilities are centrally located in a `lib/` directory in your Gatsby site:

```js:title=jest.config.js
    "^.+\\.[jt]sx?$": "lib/jest-preprocess.js",
```
````

File: docs/docs/unit-testing.md [hunk 4]
```diff
@@ -217,50 +217,24 @@ by running `npm test -- -u`.
...
 
-Then update the configuration in `jest.config.js`, like so:
+Also update `jest.preprocess.js` with the following babel preset to look like this:
```
marcysutton:
```suggestion
Also update `jest.preprocess.js` with the following Babel preset to look like this:
```

docs/docs/unit-testing.md
soosgyul: The test files don't have to live inside the lib folder. In my project where I've applied the changes all my components are in a separate folder, and all tests live inside this folder...

ascorbic: I think there's a misunderstanding here. The user isn;t supposed to replace `<rootDir>` with their own path: the jest config expects the literal string `<rootDir>`. which is replaced *by Jest* with the root directory of the project. e.g. see https://github.com/gatsbyjs/gatsby/blob/master/jest.config.js#L37

marcysutton: Ok, that is definitely confusing as angle brackets often denote "replace this" in docs. You can disregard my example, but we should clarify that the literal string `<rootDir>` needs to be included.

docs/docs/unit-testing.md
```diff
@@ -217,50 +217,24 @@ by running `npm test -- -u`.
...
 ## Using TypeScript
 
-If you are using TypeScript, you need to make a couple of small changes to your
-config. First install `ts-jest`:
+If you are using TypeScript, you need to make two changes to your
+config.
 
-```shell
-npm install --save-dev ts-jest
+Update the transform in `jest.config.js` to run `jest-preprocess` on files in your project's root directory, where `<rootDir>` is replaced with the correct relative path:
```
marcysutton:
```suggestion
Update the transform in `jest.config.js` to run `jest-preprocess` on files in your project's root directory, where `<rootDir>` is kept intact as a literal string to dynamically pass the relative path to Jest:
```
Is this accurate to communicate how the `<rootDir>` string is used here? (this was already updated once, based on my own confusion)

soosgyul: Yes, I think this is correct. Sorry I meant to update this all week, never got to it. :/

soosgyul:
I'm doubtful that my update to the documentation made the windows unit tests fail.

docs/docs/unit-testing.md
```diff
@@ -217,50 +217,24 @@ by running `npm test -- -u`.
...
 ## Using TypeScript
 
-If you are using TypeScript, you need to make a couple of small changes to your
-config. First install `ts-jest`:
+If you are using TypeScript, you need to make two changes to your
+config.
 
-```shell
-npm install --save-dev ts-jest
+Update the transform in `jest.config.js` to run `jest-preprocess` on files in your project's root directory, where `<rootDir>` is kept intact as a literal string to dynamically pass the relative path to Jest:
```
muescha: i suggest more easy english:

```suggestion
Update the transform in `jest.config.js` to run `jest-preprocess` on files in your project's root directory. 

**Note:** `<rootDir>` is replaced by Jest with the root directory of the project. Don't change it.
```

laurieontech:
This looks great! Thanks for the changes.

PR #19028, opened 2019-10-25
chore: Remove Flow

sidharthachatterjee:
It's spring cleanup time. We haven't used Flow in years. This PR removes:

- Type definitions for dependencies
- config
- script to check Flow types
- npm script to run above script
- flow-bin dependency in workspace root


yarn.lock
```diff
@@ -9248,11 +9247,6 @@ flatten@^1.0.2:
   version "1.0.2"
   resolved "https://registry.yarnpkg.com/flatten/-/flatten-1.0.2.tgz#dae46a9d78fbe25292258cc1e780a41d95c03782"
 
-flow-bin@^0.110.1:
-  version "0.110.1"
-  resolved "https://registry.yarnpkg.com/flow-bin/-/flow-bin-0.110.1.tgz#24ac70bf0871a5d6bc181ba99801ded4d5e3b442"
-  integrity sha512-6FhvNKNvPQ523mx7sqNxTQvI/HgAWa/pbIsQuCst53qRqs387EFfYqgm4I3Zae5HLaVFacBwgWKmjKd92vf19w==
-
 flow-parser@0.*:
```
maecapozzi: I wonder if this `flow-parser` package should also be removed?

sidharthachatterjee:
Yup, great point! That’s  why I left the flow types in.

yarn.lock
sidharthachatterjee: I think this is a dep of the eslint flow plugin. I wasn’t sure if I should remove that or keep it 🤔 

Wasn’t sure what was using that

PR #19031, opened 2019-10-26
feat(site-showcase-validator): Add user-agent for HTTP requests

lannonbr:
## Description

Just to make things a little more cleaner when hitting the various hundreds of links we ping, I am adding a custom user agent so if the site admin goes into the logs and sees a daily ping around the same time of day, they can know what it is coming from.

## Related Issues

N/A

### DIFF
--- a/.github/actions/gatsby-site-showcase-validator/index.js
+++ b/.github/actions/gatsby-site-showcase-validator/index.js
@@ -12,7 +12,12 @@ async function run() {
   let yamlStr
 
   try {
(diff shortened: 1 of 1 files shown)


pieh:
Maybe worth creating helper function to not have to set headers in all those requests?

```js
const fetchAsSiteValidator = url => fetch(url, { headers: { <insert_user_agent> } })
```

Alternatively maybe it make sense to set user agent just for fetch that checks the user site (as rest of fetches hit github or other git provider, where it shouldn't matter much?)

PR #18990, opened 2019-10-24
chore(docs): Update schema doc to second person reference

sagarnasit:
issue: #18284

PR #19038, opened 2019-10-26
Updated we to You to make it user interactive.

connectrv:
Updated we to You to make it user interactive.

### DIFF
--- a/docs/docs/schema-generation.md
+++ b/docs/docs/schema-generation.md
@@ -60,8 +60,8 @@ This step creates GraphQL input filters for each field so the objects can be que
 
 #### 4. ProcessedTypeNode creation with resolve implementation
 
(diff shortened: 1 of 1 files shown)


connectrv:
and our to your.

LekoArts:
Hi, thanks for the PR! But there is already this PR touching this file: https://github.com/gatsbyjs/gatsby/pull/18990

Furthermore, please read this carefully as it's important:

> When updating a doc to adhere to this part of the Gatsby Style Guide, one exception in English is when “we” refers to Gatsby’s core processes. The subject is the code in this case, rather than a teacher/reader connotation, and should be rewritten or restructured to not confuse the reader about what they are responsible for doing when something is happening automatically behind the scenes.

Issue #19018, opened 2019-10-25
refactor: extract custom ci detection code to gatsby core utils


pieh:
We currently have some custom CI detection code that augments `ci-info` with support for ZEIT's Now and heroku detection in `https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-telemetry/src/ci.js`

This is currently used for telemetry events, but this can be moved to `gatsby-core-utils` package so both `gatsby-telemetry` and `gatsby-cli` ( https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-cli/src/util/is-tty.js and https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-cli/src/reporter/index.js ) could make use of it.

## Task:

Move
 - https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-telemetry/src/ci.js
 - https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-telemetry/src/__tests__/ci.js

to `gatsby-core-utils`. Make sure that `isCI` and `getCIName` function are exported from that package. And update `gatsby-telemetry` and `gatsby-cli` to use those functions (need to add `gatsby-core-utils` to dependencies)

RobinCsl:
I'd like to help with it :)

pieh:
Awesome! Thanks @RobinCsl!

PR #19039, opened 2019-10-26
feat(gatsby-core-utils): Add isCI and getCIName

RobinCsl:
…emetry

Fixes https://github.com/gatsbyjs/gatsby/issues/19018


RobinCsl:
@pieh Thanks for your review! All your comments should be addressed now.
I've also added mock files for `gatsby-core-utils` for the tests impacted to pass.

packages/gatsby-cli/src/util/__mocks__/gatsby-core-utils.js
```diff
@@ -0,0 +1 @@
+module.exports = {}
```
pieh: I think this might be causing lots of failing tests right now (didn't have time to investigate - just looking at CI checks on the PR)

packages/gatsby-core-utils/README.md
```diff
@@ -63,3 +63,30 @@ const pathname = "./gatsby/is/awesome"
 const url = joinPath(BASEPATH, pathname)
 // ...
 ```
+
+### isCI
```
sidharthachatterjee: We love it when folks update documentation! 🥇 for that! 🤗

packages/__mocks__/gatsby-core-utils.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+    isCI: () => false,
```
sidharthachatterjee: Not sure if this file should be in this directory since the `packages` directory typically only has package folders?

packages/gatsby-core-utils/README.md
RobinCsl: My pleasure :)

packages/__mocks__/gatsby-core-utils.js
RobinCsl: Shall I move it to the root of the repo then? Otherwise, without mocking it, when I run 
```
yarn jest is-tty.js
```
for example, I was getting an error that the module could not be found.

But I think I found the problem: if I don't build the packages with `yarn lerna-prepare`, then I get this kind of errors and they disappeared afterwards.

I'll remove the mock file entirely!

packages/gatsby-cli/src/util/__mocks__/gatsby-core-utils.js
RobinCsl: I just deleted these mocks, I thought they were necessary but found out that running `yarn lerna-prepare` helped: the modules were missing because they were not built! :smile:

packages/__mocks__/gatsby-core-utils.js
sidharthachatterjee: Good idea!

RobinCsl:
Thanks Sidhartha! 
Is there something I still need to do to fix the pipeline? It looks like moving about `ci-info` caused some unhandled error to be thrown.

sidharthachatterjee:
@RobinCsl That was a bug in `gatsby-dev-cli` which we fixed in https://github.com/gatsbyjs/gatsby/pull/19085

Should be good now

